PR #948, opened 2020-02-20
Temporarily constraint to `xgboost<1.0.0`.

hvy:
Temporarily constraint to `xgboost<1.0.0` since `1.0.0` seems to contain accidental f-strings which cause our Python 3.5 CI builds to fail.

Should be reverted after https://github.com/dmlc/xgboost/issues/5328 is addressed.

PR #954, opened 2020-02-24
Revert "Temporarily constraint to `xgboost<1.0.0`".

hvy:
This reverts commit bb48c474756ce42796fe79aed45cffedbfc8db27 (https://github.com/optuna/optuna/pull/948) as xgboost has been fixed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -83,9 +83,7 @@ def get_extras_require():
             'scikit-learn',
             'torch',
             'torchvision>=0.5.0',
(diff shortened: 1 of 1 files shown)


hvy:
Regarding the note, we could also update the example to always return a `dict` with `list`s using the `as_pandas` parameter https://github.com/dmlc/xgboost/blob/cfae247231e4977a78b255810565369c934fe1b5/python-package/xgboost/training.py#L518.

PR #958, opened 2020-02-24
Add examples for `doctest` to `optuna/exceptions.py`.

nuka137:
This PR add doctest example to `optuna/exceptions.py` and `optuna/structs.py`.

Related Issue: #734

PR #809, opened 2019-12-18
Add `HyperbandPruner`

crcrpar:
~~This depends on #808~~
The algorithm would behave differently from the paper if the sampler is `TPESampler` because there is no way to filter out the trials of brackets other than that of the running trial.

---

~~Although #785 and #805 are for Hyperband, this PR follows another design policy.
The core design of #785 is to realize filtering by adding an attribute that represents pruner to Trial.
It will make `Study` more complex while it can keep the current loosely connected relationship between `Sampler` and `Pruner`.~~

~~On the other hand, the policy of this PR does not implement any filtering of trials.
This means that we will need to add some interaction between `HyperbandPruner` and `Sampler`, especially `TPESampler`.~~

EDIT:
In this implementation, trials are filtered by `bracket_id` that is computed by `HyperbandPruner._get_bracket_id` and this filtering is enabled by implementing `_BracketStudy` which just wraps `Study` for the `HyperbandPruner` use only. The idea is suggested in https://github.com/optuna/optuna/pull/809#discussion_r359688632.

Issue #514, opened 2019-09-13
Separate standard library imports and third party imports.


sile:
PEP-8 recommends separating standard library imports and third party imports (see below).

> Imports should be grouped in the following order:
> 
> 1. Standard library imports.
> 2. Related third party imports.
> 3. Local application/library specific imports.
>
> You should put a blank line between each group of imports.
> 
> From: https://www.python.org/dev/peps/pep-0008/#imports

Optuna code should follow this rule.

sile:
Resolved by #777 .

PR #963, opened 2020-02-27
Implement ThresholdPruner

himkt:
This PR introduces a simple threshold base pruner.
A threshold-based pruner is useful for people who want to
prune a trial whose metric is extremely bad.

```python
from optuna import create_study
from optuna.pruners import ThresholdPruner


def objective(trial):
    # something

pruner = optuna.pruners.ThresholdPruner(lower=0, upper=1)
# A trial will be pruned if a metric exceeds 1 or falls behind 0.
study = create_study(pruner=pruner)
study.optimize(objective)
```

- [x] warm_up
- [x] doctest
- [x] nan checking

### DIFF
--- a/optuna/pruners/threshold.py
+++ b/optuna/pruners/threshold.py
@@ -73,11 +73,11 @@ class ThresholdPruner(BasePruner):
     ) -> None:
 
         if isinstance(lower, bool) or isinstance(upper, bool):
-            raise ValueError("lower and upper should be either None, integers or floating points")
-        if (lower is not None) and (not isinstance(lower, int)) and (not isinstance(lower, float)):
-            raise ValueError("lower should be either None, integer or floating point.")
-        if (upper is not None) and (not isinstance(upper, int)) and (not isinstance(upper, float)):
-            raise ValueError("upper should be either None, integer or floating point.")
+            raise TypeError("lower and upper should be either None, integers or floating points")
+        if lower is not None and not isinstance(lower, (int, float)):
+            raise TypeError("lower should be either None, integer or floating point.")
+        if upper is not None and not isinstance(upper, (int, float)):
+            raise TypeError("upper should be either None, integer or floating point.")
         if lower is None and upper is None:
             raise ValueError("Either lower or upper must be specified.")
         if lower is not None and upper is not None and lower > upper:
--- a/optuna/pruners/threshold.py
+++ b/optuna/pruners/threshold.py
@@ -15,14 +15,39 @@ class ThresholdPruner(BasePruner):
     falls behind lower bound threshold.
 
     Example:
+        .. testcode::
 
-        .. code::
+            from optuna import create_study
+            from optuna.exceptions import TrialPruned
+            from optuna.pruners import ThresholdPruner
 
-            >>> from optuna import create_study
-            >>> from optuna.pruners import ThresholdPruner
-            >>>
-            >>> study = create_study(pruner=ThresholdPruner(upper=500))
-            >>> study.optimize(objective)
+
+            def objective_1(trial):
+                for step in range(n_trial_step):
+                    trial.report(ys_for_upper[step], step)
+
+                    if trial.should_prune():
+                        raise TrialPruned()
+
+
+            def objective_2(trial):
+                for step in range(n_trial_step):
+                    trial.report(ys_for_lower[step], step)
+
+                    if trial.should_prune():
+                        raise TrialPruned()
+
+
+            ys_for_upper = [0.0, 0.1, 0.2, 0.5, 1.2]
+            ys_for_lower = [100.0, 90.0, 0.1, 0.0, -1]
+            n_trial_step = 5
+
+            study = create_study(pruner=ThresholdPruner(upper=1.0))
+            study.optimize(objective_1, n_trials=10)
+
+            study = create_study(pruner=ThresholdPruner(lower=0.0))
+            study.optimize(objective_2, n_trials=10)
+
 
     Args
         lower:
--- a/optuna/pruners/threshold.py
+++ b/optuna/pruners/threshold.py
@@ -9,8 +9,8 @@ from optuna.pruners.percentile import _is_first_in_interval_step
 class ThresholdPruner(BasePruner):
     """Pruner to detect outlying metrics of the trials.
 
-    Prune if a metric exceeds upper bound threshold or
-    falls behind lower bound threshold.
+    Prune if a metric exceeds upper threshold,
+    falls behind lower threshold or reaches ``nan``.
 
     Example:
         .. testcode::
--- a/optuna/pruners/threshold.py
+++ b/optuna/pruners/threshold.py
@@ -1,6 +1,5 @@
 import math
 from typing import Optional
-import warnings
 
 import optuna
 from optuna.pruners import BasePruner
--- a/optuna/pruners/threshold.py
+++ b/optuna/pruners/threshold.py
@@ -3,8 +3,6 @@ from typing import Optional
 
 from optuna.pruners import BasePruner
 from optuna.pruners.percentile import _is_first_in_interval_step
-from optuna import structs
-from optuna.structs import FrozenTrial
 
 import optuna
 
@@ -72,16 +70,12 @@ class ThresholdPruner(BasePruner):
             self,
             lower: Optional[float] = None,
             upper: Optional[float] = None,
-            n_startup_trials: int = 5,
             n_warmup_steps: int = 0,
             interval_steps: int = 1
     ) -> None:
 
         if isinstance(lower, float) and isinstance(upper, float) and lower > upper:
             raise ValueError('lower should be smaller than upper')
-        if n_startup_trials < 0:
-            raise ValueError(
-                'Number of startup trials cannot be negative but got {}.'.format(n_startup_trials))
         if n_warmup_steps < 0:
             raise ValueError(
                 'Number of warmup steps cannot be negative but got {}.'.format(n_warmup_steps))
@@ -91,17 +85,10 @@ class ThresholdPruner(BasePruner):
 
         self.lower = lower
         self.upper = upper
-        self._n_startup_trials = n_startup_trials
         self._n_warmup_steps = n_warmup_steps
         self._interval_steps = interval_steps
 
-    def prune(self, study: 'optuna.study.Study', trial: FrozenTrial) -> bool:
-
-        all_trials = study.get_trials(deepcopy=False)
-        n_trials = len([t for t in all_trials if t.state == structs.TrialState.COMPLETE])
-
-        if n_trials < self._n_startup_trials:
-            return False
+    def prune(self, study: 'optuna.study.Study', trial: optuna.structs.FrozenTrial) -> bool:
 
         step = trial.last_step
         if step is None:
--- a/optuna/pruners/threshold.py
+++ b/optuna/pruners/threshold.py
@@ -124,17 +124,3 @@ class ThresholdPruner(BasePruner):
             return True
 
         return False
-
-    @property
-    def lower(self) -> Optional[float]:
-
-        if self._lower == -float("inf"):
-            return None
-        return self._lower
-
-    @property
-    def upper(self) -> Optional[float]:
-
-        if self._upper == float("inf"):
-            return None
-        return self._upper
--- a/optuna/pruners/threshold.py
+++ b/optuna/pruners/threshold.py
@@ -26,20 +26,20 @@ class ThresholdPruner(BasePruner):
             >>> study.optimize(objective)
 
     Args
-        lower_bound:
+        lower:
             minimum value which determines whether pruner prunes or not
-            (If value is smaller than lower_bound, it prunes)
-        upper_bound:
+            (If value is smaller than lower, it prunes)
+        upper:
             maximum value which determines whether pruner prunes or not
-            (If value is larger than upper_bound, it prunes)
+            (If value is larger than upper, it prunes)
 
     """
 
-    def __init__(self, lower_bound=None, upper_bound=None):
+    def __init__(self, lower=None, upper=None):
         # type: (Optional[float], Optional[float]) -> None
 
-        self.lower_bound = lower_bound
-        self.upper_bound = upper_bound
+        self.lower = lower
+        self.upper = upper
 
     def prune(self, study, trial):
         # type: (Study, FrozenTrial) -> bool
@@ -50,10 +50,10 @@ class ThresholdPruner(BasePruner):
 
         latest_value = trial.intermediate_values[last_step]
 
-        if self.lower_bound is not None and latest_value < self.lower_bound:
+        if self.lower is not None and latest_value < self.lower:
             return True
 
-        if self.upper_bound is not None and latest_value > self.upper_bound:
+        if self.upper is not None and latest_value > self.upper:
             return True
 
         return False
--- a/optuna/pruners/threshold.py
+++ b/optuna/pruners/threshold.py
@@ -21,16 +21,16 @@ class ThresholdPruner(BasePruner):
 
 
             def objective_1(trial):
-                for step in range(n_trial_step):
-                    trial.report(ys_for_upper[step], step)
+                for step, y in enumerate(ys_for_upper):
+                    trial.report(y, step)
 
                     if trial.should_prune():
                         raise TrialPruned()
 
 
             def objective_2(trial):
-                for step in range(n_trial_step):
-                    trial.report(ys_for_lower[step], step)
+                for step, y in enumerate(ys_for_lower):
+                    trial.report(y, step)
 
                     if trial.should_prune():
                         raise TrialPruned()
--- a/optuna/pruners/threshold.py
+++ b/optuna/pruners/threshold.py
@@ -51,8 +51,8 @@ class ThresholdPruner(BasePruner):
 
     Args
         lower:
-            minimum value which determines whether pruner prunes or not
-            (If value is smaller than lower, it prunes)
+            A minimum value which determines whether pruner prunes or not.
+            If an intermediate value is smaller than lower, it prunes.
         upper:
             maximum value which determines whether pruner prunes or not
             (If value is larger than upper, it prunes)
--- a/optuna/pruners/threshold.py
+++ b/optuna/pruners/threshold.py
@@ -81,7 +81,7 @@ class ThresholdPruner(BasePruner):
         if (upper is not None) and (not isinstance(upper, int)) and (not isinstance(upper, float)):
             raise ValueError('upper should be either None, integer or floating point.')
         if lower is None and upper is None:
-            warnings.warn('Both lower and upper are not specified.')
+            raise ValueError('Either lower or upper must be specified.')
         if lower is not None and upper is not None and lower > upper:
             raise ValueError('lower should be smaller than upper.')
         if n_warmup_steps < 0:
--- a/optuna/pruners/threshold.py
+++ b/optuna/pruners/threshold.py
@@ -67,36 +67,38 @@ class ThresholdPruner(BasePruner):
     """
 
     def __init__(
-            self,
-            lower: Optional[float] = None,
-            upper: Optional[float] = None,
-            n_warmup_steps: int = 0,
-            interval_steps: int = 1
+        self,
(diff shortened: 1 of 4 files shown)


himkt:
Could someone please review. :bow:

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,59 @@
...
+        lower_bound:
+            minimum value which determines whether pruner prunes or not
+            (If value is smaller than lower_bound, it prunes)
+        upper_bound:
+            maximum value which determines whether pruner prunes or not
+            (If value is larger than upper_bound, it prunes)
+
+    """
+
+    def __init__(self, lower_bound=None, upper_bound=None):
```
hvy: I'm wondering if we shouldn't provide a `n_warmup_steps` (https://github.com/optuna/optuna/blob/master/optuna/pruners/percentile.py#L127-L133) as users might want to give the model a few iterations before assessing it's performance. The user could check this him-/herself but this warmup parameter would be consistent with the median and percentile pruners.

hvy: Could be a matter of taste, but what do you think of shorter names? I personally prefer these as they still communicate the intent well enough.
```suggestion
    def __init__(self, lower=None, upper=None):
```

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,59 @@
...
+
+    from optuna.structs import FrozenTrial
+    from optuna.study import Study
+
+
+class ThresholdPruner(BasePruner):
+    """Pruner to detect abnormal metrics of the trials.
+
+    Prune if a metric exceeds upper bound threshold or
+    falls behind lower bound threshold which users specify.
```
hvy: A nit, but I think it's slightly verbose.
```suggestion
    falls behind lower bound threshold.
```

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Example:
+
+        .. code::
```
hvy: We could address this in a follow up but would you find it possible to rewrite this example as a doctest similar to https://github.com/optuna/optuna/pull/958/files?

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,59 @@
...
+            >>> from optuna import create_study
+            >>> from optuna.pruners import ThresholdPruner
+            >>>
+            >>> study = create_study(pruner=ThresholdPruner(upper=500))
+            >>> study.optimize(objective)
+
+    Args
+        lower_bound:
+            minimum value which determines whether pruner prunes or not
+            (If value is smaller than lower_bound, it prunes)
```
hvy: Nits.
```suggestion
            Minimum value which determines whether pruner prunes or not
            (prunes if value is smaller than lower_bound).
```

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,59 @@
...
+        upper_bound:
+            maximum value which determines whether pruner prunes or not
+            (If value is larger than upper_bound, it prunes)
```
hvy: Ditto.

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,59 @@
+from optuna.pruners import BasePruner
+
```
hvy: Just wondering but does the style checker require this blank line?

optuna/pruners/threshold.py
hvy: How about raising a `ValueError` in `__init__` if `lower_bound` is larger than `upper_bound`?

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,59 @@
...
+if TYPE_CHECKING:
+    from typing import Optional
+
+    from optuna.structs import FrozenTrial
+    from optuna.study import Study
+
+
+class ThresholdPruner(BasePruner):
+    """Pruner to detect abnormal metrics of the trials.
```
hvy: What do you think of using a different term than "abnormal". One might for instance want to use this pruner to prune trials that have reached a sufficiently high accuracy, in which case "abnormal" might sound misleading. A suggestion could be
```suggestion
    """Pruner to detect outlying metrics of the trials.
```

himkt: Thank you for the suggestion. :memo:

optuna/pruners/threshold.py
himkt: Sorry, it is not needed. (removed)

optuna/pruners/threshold.py
himkt: I tried to create a doctest in 868a857. (but `make doctest` seems to ignore this...)

hvy: Ah, you have to add the `ThresholdPruner` to https://github.com/optuna/optuna/blob/master/docs/source/reference/pruners.rst.

hvy:
FYI, the CI failure with sklearn's update is not related to this PR and should be fixed by now (if we'd rerun the same CI build).

optuna/pruners/threshold.py
himkt: Thank you, I memorized. :memo:

optuna/pruners/threshold.py
himkt: 8199eef
22dcb01
7c6944c

I introduced three members `n_startup_trials`, `n_warmup_steps` and `interval_steps`.

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,132 @@
...
+        self.upper = upper
+        self._n_startup_trials = n_startup_trials
+        self._n_warmup_steps = n_warmup_steps
+        self._interval_steps = interval_steps
+
+    def prune(self, study, trial):
+        # type: (Study, FrozenTrial) -> bool
+
+        all_trials = study.get_trials(deepcopy=False)
+        n_trials = len([t for t in all_trials if t.state == structs.TrialState.COMPLETE])
```
himkt: I'm not sure if I have to check `n_trials` in `ThresholdPruner`.
(like https://github.com/optuna/optuna/blob/master/optuna/pruners/percentile.py#L148-L149)

himkt:
> the CI failure with sklearn's update is not related to this PR and should be fixed by now

I got it! Thank you for letting me know. :pray:

I revised my PR.
Could you please take a look? :bow:

optuna/pruners/threshold.py
```diff
@@ -94,7 +95,7 @@ def __init__(
         self._n_warmup_steps = n_warmup_steps
         self._interval_steps = interval_steps
 
-    def prune(self, study: Study, trial: FrozenTrial) -> bool:
+    def prune(self, study: 'optuna.study.Study', trial: FrozenTrial) -> bool:
```
himkt: To use type hint, it needs `optuna.study.Study`.
However, `from optuna.study import Study` is a circular import and
`import optuna; optuna.study.Study` fails with a message `AttributeError: module 'optuna' has no attribute 'study'.`. (same as https://github.com/optuna/optuna/pull/809#discussion_r361407060)

As a workaround, I use a string literal.
(I want some suggestions if it is not good...)

optuna/pruners/threshold.py
hvy: The logic here is different from the percentile pruner in that you're not comparing reported values across trials. So I don't think you need this check, nor the `n_startup_trials` parameter.

himkt: It makes sense and I removed in 58ea2f82.

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,116 @@
+import math
+from typing import Optional
+
+from optuna.pruners import BasePruner
+from optuna.pruners.percentile import _is_first_in_interval_step
+
+import optuna
```
c-bata: Please reorder the import lines based on [this rule](https://github.com/optuna/optuna/issues/514)?

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,116 @@
...
+    def __init__(
+            self,
+            lower: Optional[float] = None,
+            upper: Optional[float] = None,
+            n_warmup_steps: int = 0,
+            interval_steps: int = 1
+    ) -> None:
+
+        if isinstance(lower, float) and isinstance(upper, float) and lower > upper:
+            raise ValueError('lower should be smaller than upper')
```
c-bata: I think you must not check the types of lower and upper.

```console
$ python
Python 3.8.1 (default, Jan 17 2020, 14:59:07) 
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def foo(lower: float, upper: float):
...     if isinstance(lower, float) and isinstance(upper, float) and lower > upper:
...         raise ValueError('lower should be smaller than upper')
... 
>>> foo(3, 2)
>>>
```

How about the following?

```suggestion
        if lower is not None and upper is not None and lower > upper:
            raise ValueError('lower should be smaller than upper')
```

optuna/pruners/threshold.py
himkt: Sorry for violating the rule. :bow:
1ba33f6 I run `isort` to fix the import order.

optuna/pruners/threshold.py
himkt: Thank you for catching this! I missed it.

optuna/pruners/threshold.py
c-bata: No. No problem 👍 Thank you for your swift fix.

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Args
+        lower:
+            minimum value which determines whether pruner prunes or not
+            (If value is smaller than lower, it prunes)
+        upper:
+            maximum value which determines whether pruner prunes or not
+            (If value is larger than upper, it prunes)
+        n_startup_trials:
+            Pruning is disabled until the given number of trials finish in the same study.
```
c-bata: It looks this parameter does not exist.

```suggestion
```

c-bata:
Thank you for your swift fix. I'll approve this pull request after fixed docstring 👍

optuna/pruners/threshold.py
himkt: You're right. Thanks! 🙇

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,113 @@
...
+
+    Example:
+        .. testcode::
+
+            from optuna import create_study
+            from optuna.exceptions import TrialPruned
+            from optuna.pruners import ThresholdPruner
+
+
+            def objective_1(trial):
```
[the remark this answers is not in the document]
himkt: Thank you for the suggestion, I totally agree with you.

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,113 @@
...
+            study.optimize(objective_1, n_trials=10)
+
+            study = create_study(pruner=ThresholdPruner(lower=0.0))
+            study.optimize(objective_2, n_trials=10)
+
+
+    Args
+        lower:
+            minimum value which determines whether pruner prunes or not
+            (If value is smaller than lower, it prunes)
```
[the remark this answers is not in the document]
himkt: It makes sense, thank you! (sorry for missing the period)

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,113 @@
...
+
+    """
+
+    def __init__(
+            self,
+            lower: Optional[float] = None,
+            upper: Optional[float] = None,
+            n_warmup_steps: int = 0,
+            interval_steps: int = 1
+    ) -> None:
```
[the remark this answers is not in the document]
himkt: In the case `lower == upper == None`, `ThresholdPruner` should raise an Exception since

1. we can create an instance of `ThresholdPruner` doesn't prune anything by setting `lower = -inf, upper = inf`
2. we can create a pruner that never prune a training using `NopPruner`.

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,125 @@
...
+            return False
+
+        if not _is_first_in_interval_step(
+            step, trial.intermediate_values.keys(), n_warmup_steps, self._interval_steps
+        ):
+            return False
+
+        latest_value = trial.intermediate_values[step]
+        if math.isnan(latest_value):
+            return True
```
hvy: I find this behavior to be useful to the majority of the users (although some might want it the other way) but how about documenting it?

hvy: How about covering this path in unit tests?

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,125 @@
...
+    def __init__(
+        self,
+        lower: Optional[float] = None,
+        upper: Optional[float] = None,
+        n_warmup_steps: int = 0,
+        interval_steps: int = 1,
+    ) -> None:
+
+        if isinstance(lower, bool) or isinstance(upper, bool):
+            raise ValueError("lower and upper should be either None, integers or floating points")
```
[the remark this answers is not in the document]
himkt: Thank you for the suggestion, I applied your suggestion (partially). b516491b

> Does the bool type need a separate check?

I check the `bool` separately since `isinstance(True, int)` returns `True`.
I'm not sure if it's enough to check `(int, float)`.
What do you think?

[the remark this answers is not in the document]
himkt: You're right.
Now I think it's reasonable to allow only `float` as an input.
(Allowing only `float` also makes it possible to remove `bool` specific checking.)

optuna/pruners/threshold.py
himkt: 0ddb2be added the test for nan.

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,125 @@
...
+        if (lower is not None) and (not isinstance(lower, int)) and (not isinstance(lower, float)):
+            raise ValueError("lower should be either None, integer or floating point.")
+        if (upper is not None) and (not isinstance(upper, int)) and (not isinstance(upper, float)):
+            raise ValueError("upper should be either None, integer or floating point.")
+        if lower is None and upper is None:
+            raise ValueError("Either lower or upper must be specified.")
+        if lower is not None and upper is not None and lower > upper:
```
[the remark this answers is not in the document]
himkt: aff308a applied your suggestion and introduced the interface for `_upper` and `_lower`.

optuna/pruners/threshold.py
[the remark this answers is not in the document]
himkt: b516491 :bow:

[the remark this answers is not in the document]
himkt: 9822b18 prohibits integers.

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,136 @@
...
+        n_warmup_steps: int = 0,
+        interval_steps: int = 1,
+    ) -> None:
+
+        if lower is not None and not isinstance(lower, (float)):
+            raise TypeError("lower should be either None or a floating point.")
+        if upper is not None and not isinstance(upper, (float)):
+            raise TypeError("upper should be either None or a floating point.")
+        if lower is None and upper is None:
+            raise ValueError("Either lower or upper must be specified.")
```
himkt: I leave it as it is since `lower`/`upper` could be `None`.

tests/pruners_tests/test_threshold.py
```diff
@@ -0,0 +1,58 @@
+import optuna
+
+
```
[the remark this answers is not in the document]
himkt: Sorry for the late reply.
I check the unacceptable values in https://github.com/optuna/optuna/pull/963/files#diff-6a0dcdbfaaa30632e519f5ee76996141R56-R77.

himkt:
@toshihikoyanase @hvy 
Thank you very much for your patient reviews.
I revised my PR, please take a look again.

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,140 @@
...
+            ys_for_lower = [100.0, 90.0, 0.1, 0.0, -1]
+            n_trial_step = 5
+
+            study = create_study(pruner=ThresholdPruner(upper=1.0))
+            study.optimize(objective_for_upper, n_trials=10)
+
+            study = create_study(pruner=ThresholdPruner(lower=0.0))
+            study.optimize(objective_for_lower, n_trials=10)
+
+
```
hvy: A nit.
```suggestion
```

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,140 @@
...
+        if self._lower == -float("inf"):
+            return None
+        return self._lower
+
+    @property
+    def upper(self) -> Optional[float]:
+
+        if self._upper == float("inf"):
+            return None
+        return self._upper
```
hvy: These properties will return `None` when the user passes `float("inf")` or `-float("inf")` to `__init__` which might be confusing. I bet it's safer to simply hide these properties as a first implementation. What do you think?

optuna/pruners/threshold.py
```diff
@@ -0,0 +1,140 @@
...
+        n_warmup_steps: int = 0,
+        interval_steps: int = 1,
+    ) -> None:
+
+        if lower is not None and not isinstance(lower, float):
+            raise TypeError("lower should be either None or a floating point.")
+        if upper is not None and not isinstance(upper, float):
+            raise TypeError("upper should be either None or a floating point.")
+        if lower is None and upper is None:
+            raise ValueError("Either lower or upper must be specified.")
```
hvy: The validation looks a lot simpler now. However, I'd like to align the semantics with https://github.com/optuna/optuna/blob/master/optuna/trial.py#L507-L514 in favor of usability as arguments have similar meanings, i.e. try casting to `float` (which would also mean "allowing" `bool`s ). Do you think that's reasonable? If yes, but if it's too much of dragging out, we could address that in a separate PR as it won't break existing code.

optuna/pruners/threshold.py
himkt: Sorry for missing.

optuna/pruners/threshold.py
himkt: It makes sense and I didn't consider the case user passes them directly.
Thank you for the suggestion! (hired in 6f710802)

optuna/pruners/threshold.py
himkt: Thank you, I think it's reasonable.
I revised my PR.

- introduce `_check_value` which is the same as https://github.com/optuna/optuna/blob/master/optuna/trial.py#L507-L514
- use it to check `lower` and `upper`

I'll create another PR to move `_check_value` to a suitable place (if you have a good idea, please let me know) and use it in `trial.py`.
What do you think? @hvy

optuna/pruners/threshold.py
```diff
@@ -71,12 +83,12 @@ def __init__(
...
-            raise TypeError("lower should be either None or a floating point.")
-        if upper is not None and not isinstance(upper, float):
-            raise TypeError("upper should be either None or a floating point.")
         if lower is None and upper is None:
-            raise ValueError("Either lower or upper must be specified.")
+            raise TypeError("Either lower or upper must be specified.")
+        if lower is not None:
+            _check_value(lower)
+        if upper is not None:
+            _check_value(upper)
```
hvy: How about returning and reassigning the casted value?

himkt: 713eda8 :+1:

tests/pruners_tests/test_threshold.py
```diff
@@ -56,10 +56,10 @@ def test_threshold_pruner_with_two_side() -> None:
 def test_threshold_pruner_with_invalid_inputs() -> None:
 
     with pytest.raises(TypeError):
-        optuna.pruners.ThresholdPruner(lower="0", upper=1.0)  # type: ignore
```
himkt: I have noticed `float()` can take `"0"`.

hvy: Thanks for the catch, you are correct. Accepting `"0"` seems reasonable and consistent.

optuna/pruners/threshold.py
hvy: Good question. I'd personally like to create a new lower level submodule (e.g. instead of `study.py` avoiding having several modules depend on `study.py`). It could be private. But I would assume others have other opinions.

Issue #734, opened 2019-11-25
Apply doctest to existing documents.


g-votte:
Though we've introduced doctest in #702, many existing examples are still incompatible with doctest. This issue is to ~list up those examples and~ apply doctest to them.

Below are files to be fixed, listed up in https://github.com/optuna/optuna/issues/734#issuecomment-586678423. (Thanks @keisuke-umezawa for your consolidating the information!)

- [x] optuna/visualization/parallel_coordinate
- [x] optuna/visualization/slice
- [x] optuna/visualization/intermediate_values
- [x] optuna/visualization/optimization_history
- [x] optuna/visualization/contour
- [x] optuna/logging
- [x] optuna/storages/rdb/storage
- [x] optuna/trial
- [x] optuna/pruners/successive_halving
- [x] optuna/pruners/nop
- [x] optuna/pruners/median
- [x] optuna/pruners/percentile
- [x] optuna/integration/cma
- [x] optuna/integration/lightgbm
- [x] optuna/integration/mxnet
- [x] optuna/integration/chainermn
- [x] optuna/integration/fastai
- [x] optuna/integration/keras
- [x] optuna/integration/pytorch_ignite
- [x] optuna/integration/tensorflow
- [x] optuna/integration/sklearn
- [x] optuna/integration/skopt
- [x] optuna/integration/pytorch_lightning
- [x] optuna/integration/xgboost
- [x] optuna/integration/tfkeras
- [x] optuna/integration/chainer
- [x] ~optuna/structs~
  - see also [the discussion about `structs.py`](https://github.com/optuna/optuna/pull/958#discussion_r383844795)
- [x] optuna/samplers/tpe/sampler
- [x] optuna/samplers/random
- [x] optuna/exceptions

nuka137:
@g-votte 

I'm interested in tackling this issue.
I think [optuna/trial.py](https://github.com/optuna/optuna/blob/master/optuna/trial.py#L161-L166) is one of the candidate.
Does this meet your intention?

g-votte:
@nuka137 
Of course, and your contribution is greatly welcomed!

nuka137:
@g-votte 

Created PR #882 .

P.S.
I also noticed `make doctest` raises below error in newest scikit-learn (0.22.1).
I think `_index_param_value` is removed in newest scikit-learn. (https://github.com/scikit-learn/scikit-learn/pull/15863/files)

```
[...]

PR #964, opened 2020-02-28
Add examples for `doctest` to `optuna/prunes/*.py`

nuka137:
This PR add doctest example to below files.

* optuna/pruners/successive_halving
* optuna/pruners/nop
* optuna/pruners/median
* optuna/pruners/percentile

Related Issue: #734

### DIFF
--- a/optuna/pruners/nop.py
+++ b/optuna/pruners/nop.py
@@ -11,16 +11,42 @@ class NopPruner(BasePruner):
 
     Example:
 
-        .. code::
+        .. testsetup::
 
-            >>> from optuna import create_study
-            >>> from optuna.pruners import NopPruner
-            >>>
-            >>> def objective(trial):
-            >>>     ...
-            >>>
-            >>> study = create_study(pruner=NopPruner())
-            >>> study.optimize(objective)
+            import numpy as np
+            from sklearn.model_selection import train_test_split
+
+            np.random.seed(seed=0)
+            X = np.random.randn(200).reshape(-1, 1)
+            y = np.where(X[:, 0] < 0.5, 0, 1)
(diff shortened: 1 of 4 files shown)


hvy:
Thanks @nuka137 for your constant contributions. We'll review it as soon as possible.

optuna/pruners/nop.py
```diff
@@ -11,16 +11,42 @@ class NopPruner(BasePruner):
...
+                n_train_iter = 100
+
+                for step in range(n_train_iter):
+                    clf.partial_fit(X_train, y_train, classes=classes)
+
+                    intermediate_value = clf.score(X_test, y_test)
+                    trial.report(intermediate_value, step)
+
+                    if trial.should_prune():
+                        print("Never reach here")
```
hvy: How about a slightly stricter doctest?
```suggestion
                    assert False, "should_prune() should always return False with this pruner."
```

nuka137:
@hvy 

Thanks for your review!
It's a good idea to apply strict doctest for `NopPruner`.
I fixed it.

hvy:
If duplication is a concern, we could also not have any example code at all. Instead, we could in the documentation for each pruner just refer to the example in the `TrialPruned` exception with a `.. seealso::`. In addition, we could improve the `TrialPruned` exception example to specify a particular pruner instead of relying on the default (https://github.com/nuka137/optuna/blob/master/optuna/exceptions.py#L48) to make it clearer.

toshihikoyanase:
@hvy Thank you for your suggestion. It sounds good to me, but it still has some cons (e.g., users need one more click to see the example).

So, how about accepting this PR for now and create a new PR if we work on it?

PR #967, opened 2020-02-28
Re-seed RNG before each hyperparameter sampling.

sile:
This PR resolves a reproducibility problem of samplers reported by #914.

The problem was introduced by the commits https://github.com/optuna/optuna/commit/631a2a2bfab2f76b86981acf00ff5b5f4bb0330a and https://github.com/optuna/optuna/commit/3f01ca234d91dd5281e784f9c0f5f124f897f275. Those defined custom `__setstate__` and `__getstate__` methods of `RandomSampler` where the state of RNG is reset.
If we simply reverted the change, however, it would cause another problem relating to parallel optimization. That is, all the parallel workers might sample the same hyperparameters for different trials (see https://github.com/optuna/optuna/pull/692#issuecomment-555869535 for the detail).
To avoid the duplicate work, this PR re-seeds RNG by using trial number information before each hyperparameter sampling, so even if all the workers have the same initial RNG, they will sample different hyperparamers for different trials.

Note that this change would incur extra computation cost due to the additional re-seeding. I'll take a benchmark to estimate the cost.

Close #914

Issue #914, opened 2020-02-11
Unable to reproduce results before and after a study is deepcopied


Y-oHr-N:
The behavior is different between 0.19.0 and 1.x.x.

## Environment

- Python version: 3.6.5
- OS: Windows 10

## Examples

```python
import copy

import optuna


def objective(trial):
    x = trial.suggest_uniform('x', -10, 10)

    return (x - 2) ** 2


study = optuna.create_study(sampler=optuna.samplers.TPESampler(seed=0))
study.optimize(objective, n_trials=1)

study = optuna.create_study(sampler=optuna.samplers.TPESampler(seed=0))
study = copy.deepcopy(study)
study.optimize(objective, n_trials=1)
```

### v0.19.0

```
[I 2020-02-12 02:37:41,933] Finished trial#0 resulted in value: 1.0480229520791995. Current best value is 1.0480229520791995 with parameters: {'x': 0.9762700785464951}.
[I 2020-02-12 02:37:41,990] Finished trial#0 resulted in value: 1.0480229520791995. Current best value is 1.0480229520791995 with parameters: {'x': 0.9762700785464951}.
```

### v1.0.0

```
[I 2020-02-12 02:38:44,010] Finished trial#0 resulted in value: 1.0480229520791995. Current best value is 1.0480229520791995 with parameters: {'x': 0.9762700785464951}.
[I 2020-02-12 02:38:44,067] Finished trial#0 resulted in value: 46.90638356377688. Current best value is 46.90638356377688 with parameters: {'x': 8.848823516763801}.
```

### v1.1.0

```
[I 2020-02-12 02:39:39,168] Finished trial#0 resulted in value: 1.0480229520791995. Current best value is 1.0480229520791995 with parameters: {'x': 0.9762700785464951}.
[I 2020-02-12 02:39:39,226] Finished trial#0 resulted in value: 2.756825098011473. Current best value is 2.756825098011473 with parameters: {'x': 3.6603689644207016}.
```

c-bata:
It seems a cause of this problem is the following commit.
https://github.com/optuna/optuna/commit/3f01ca234d91dd5281e784f9c0f5f124f897f275

Y-oHr-N:
I have read https://github.com/optuna/optuna/pull/692#issuecomment-558510300.

> But I don't think the approach to initialize the RandomSampler is the main topic of this PR, we can keep the current implementation and discuss it after introducing the joblib.

[...]

PR #968, opened 2020-02-28
Reseed sampler's random seed generator in Study.

toshihikoyanase:
This PR also addresses #914, but a different approach is used from #967.
Reseeding of the random seed generator in `RandomSampler.__setstate__` is introduced for the parallelization using joblib. I think the reseeding should be applied only in the parallel computation.
In this PR, I created a simple wrapper of `Study._optimize_sequential`, which reseeds the random sampler in advance. And I used it only when `n_jobs>1`. 
`RandomSampler.__setstate__` or `RandomSampler.__getstate__` are not necessary anymore, and deepcopy or pickle dump/load will keep the state of the random number generator.

### DIFF
--- a/optuna/samplers/base.py
+++ b/optuna/samplers/base.py
@@ -120,3 +120,16 @@ class BaseSampler(object, metaclass=abc.ABCMeta):
         """
 
         raise NotImplementedError
+
+    @abc.abstractmethod
+    def reseed_rng(self) -> None:
+        """Reseed sampler's random number generator.
+
+        This method is called by the :class:`~optuna.study.Study` class if trials are executed in
+        parallel with the option `n_jobs>1`. In that case, the sampler instance will be replicated
+        including the state of the random number generator, and they may suggest the same values.
+        To prevent this issue, this method will assign different seeds to each random number
+        generator.
+        """
+
+        raise NotImplementedError
--- a/optuna/samplers/base.py
+++ b/optuna/samplers/base.py
@@ -125,11 +125,11 @@ class BaseSampler(object, metaclass=abc.ABCMeta):
     def reseed_rng(self) -> None:
         """Reseed sampler's random number generator.
 
-        This method is called by the :class:`~optuna.study.Study` class if trials are executed in
-        parallel with the option `n_jobs>1`. In that case, the sampler instance will be replicated
-        including the state of the random number generator, and they may suggest the same values.
-        To prevent this issue, this method will assign different seeds to each random number
-        generator.
+        This method is called by the :class:`~optuna.study.Study` instance if trials are executed
+        in parallel with the option ``n_jobs>1``. In that case, the sampler instance will be
+        replicated including the state of the random number generator, and they may suggest the
+        same values. To prevent this issue, this method assigns a different seed to each random
+        number generator.
         """
 
         raise NotImplementedError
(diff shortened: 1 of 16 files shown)


optuna/study.py
```diff
@@ -566,6 +567,30 @@ def _append_trial(
...
+            n_trials,  # type: Optional[int]
+            timeout,  # type: Optional[float]
+            catch,  # type: Union[Tuple[()], Tuple[Type[Exception]]]
+            callbacks,  # type: Optional[List[Callable[[Study, structs.FrozenTrial], None]]]
+            gc_after_trial,  # type: bool
+            time_start  # type: Optional[datetime.datetime]
+    ):
+        # type: (...) -> None
+
+        _rng = getattr(self.sampler, "_rng", None)
```
sile: I think that this code cannot handle nested samplers (e.g., RNG field of `RandomSampler` owned by `TPESampler`).
How about adding an abstract method to `BaseSampler` such as `BaseSampler.reseed_rng` for this purpose?

optuna/samplers/base.py
```diff
@@ -120,3 +120,16 @@ def sample_independent(self, study, trial, param_name, param_distribution):
...
+        """Reseed sampler's random number generator.
+
+        This method is called by the :class:`~optuna.study.Study` instance if trials are executed
+        in parallel with the option ``n_jobs>1``. In that case, the sampler instance will be
+        replicated including the state of the random number generator, and they may suggest the
+        same values. To prevent this issue, this method assigns a different seed to each random
+        number generator.
+        """
+
+        raise NotImplementedError
```
g-votte, line 135: This always requires an additional step of overriding the method when the user want to make a new sampler, which I think a bit burdensome. Can't we just do `pass` instead of `NotImplementedSampler`?

toshihikoyanase: Thank you for your suggestion. That makes sense.
The current implementation will not work with existing user-defined samplers.
I fixed it in https://github.com/optuna/optuna/pull/968/commits/bcdcf9d2944d67db42546f9b1b6f42e8c512dcb5.

optuna/samplers/grid.py
```diff
@@ -95,6 +95,10 @@ def __init__(self, search_space):
         self._param_names = sorted(search_space.keys())
         self._n_min_trials = len(self._all_grids)
 
+    def reseed_rng(self) -> None:
```
sile, line 98: [nits] I think we can remove this code since it is the same behavior as the default implementation.

optuna/testing/sampler.py
```diff
@@ -45,6 +45,10 @@ def sample_independent(self, study, trial, param_name, param_distribution):
 
         return param_value
 
+    def reseed_rng(self) -> None:
```
sile, line 48: Ditto.

tests/samplers_tests/test_samplers.py
```diff
@@ -208,6 +215,10 @@ def sample_independent(self, study, trial, param_name, param_distribution):
 
         return self.unknown_param_value
 
+    def reseed_rng(self) -> None:
```
sile, line 218: Ditto.

optuna/samplers/grid.py
toshihikoyanase: You're absolutely right. Fixed: https://github.com/optuna/optuna/pull/968/commits/359c9ae5fad5c6bf13b8cb38cf206f41d894d71c.

PR #979, opened 2020-03-04
Fix overlap of labels in parallel coordinate plots

### LINKED ISSUE #612: Axis labels of parallel coordinate plots overlap if the parameter names are long.
As you can see in the following image, axis labels overlap when parameter names are too long. I think we can solve this issue if we limit the maximum length of the labels and/or tilt the labels.

newplot (39)

**Conditions**
- Optuna version: 0.17.1
- Python version: 3.7.2
- OS: macOS 10.13

**Code to reproduce**

```python
import optuna

def objective(trial):
    x = trial.suggest_uniform('x' * 100, -10, 10)
    y = trial.suggest_uniform('y' * 100, -10, 10)
    return x - y

study = optuna.create_study()

optuna.visualization.plot_parallel_coordinate(study)
```

VladSkripniuk:
Tilt and limit the maximum length of very long parameter names in parallel coordinate plot (fix #612)
Below is an example plot showing tilted parameter names
image


optuna/visualization/parallel_coordinate.py
```diff
@@ -110,6 +116,7 @@ def _get_parallel_coordinate_plot(study, params=None):
     traces = [
         go.Parcoords(
             dimensions=dims,
+            labelangle=labelangle,
```
hvy: How about avoiding the overlapping with the title by placing the labels at the bottom? https://plot.ly/python/reference/#parcoords-labelside

optuna/visualization/parallel_coordinate.py
```diff
@@ -80,6 +80,12 @@ def _get_parallel_coordinate_plot(study, params=None):
         all_params = set(params)
     sorted_params = sorted(list(all_params))
 
+    labelangle = None
+    for p_name in sorted_params:
+        if len(p_name) > 10:
+            labelangle = 30
+            break
```
hvy: Maybe a bit more Pythonic.
```python
if any(len(p_name) > 10 for p_name in sorted_params):
    labelangle = 30  
else: 
    labelangle = None
```
You can one-line it too but I'm skeptical it aids readability.

himkt:
@VladSkripniuk 
Could you please merge the latest master branch to resolve conflicts?
(might be the cause: #1030)

keisuke-umezawa:
@VladSkripniuk 
FYI
Thank you for mergeing the latest master. 
But could you install `black` and run the following commands?

```
black <path to Python file/directory> --line-length 99
```

Because we changed formatter from autpep8 to black in #1030 , we need to run the above command.
Sorry for inconvenience 🙏

VladSkripniuk:
@keisuke-umezawa thank you for your help with this PR.

optuna/visualization/parallel_coordinate.py
```diff
@@ -86,13 +86,21 @@ def _get_parallel_coordinate_plot(study, params=None):
         all_params = set(params)
     sorted_params = sorted(list(all_params))
 
+    if any(len(p_name) > 10 for p_name in sorted_params):
+        labelangle = 30
+        labelside = "bottom"
+    else:
+        labelangle = 0
+        labelside = "top"
```
hvy: How about always setting the angle to 30 and side to "bottom", no matter the params. The reasons is that the labels will overlap with this logic if there are too many hyperparameters that are all shorter than 11 characters.

hvy: If you are busy with https://github.com/optuna/optuna/pull/1037, maybe we could merge this PR as is and I could create a follow up. If this change sounds reasonable that is.

VladSkripniuk: @hvy I'm fine with always setting the angle to 30 and side to "bottom", thanks for suggestion.

hvy:
Let me modify the title of this PR to match the format of our release notes.

Issue #966, opened 2020-02-28
Exception/Method to exit a study.


g-votte:
## Motivation

The motivation of this issue is to enable the user to customize the exit criterion of a study. We've already had similar request in #850 where the user want to give a performance threshold. In addition, this feature would improve the `GridSampler` so that the user do not have to specify `n_trials`, as discussed in https://github.com/optuna/optuna/pull/665#discussion_r361424127.

## Description
This feature can be implemented as `optuna.exceptions.StopStudy`, which is caught and handled by the study object so that it stop running even if the number of finished trials does not meet `n_trials`.

The following snippet shows one of use-cases. The exception can be raised either in an objective function or a callback. 

```python
import ...

def objective(trial):
    ...

def callback(study, trial):
    if trial.value < EXIT_THRESHOLD:
        raise StopStudy

study = optuna.create_study()
study.optimize(objective, callbacks=[callback])
```

To help `GridSampler` (and the other samplers) stop the study, we'd need to support raising the exception inside `infer_relative_search_space` and `sample_relative`, which we may need some more design discussion.

## Alternatives
Another option is to provide some method/function to stop a study. In my opinion, providing an exception is more consistent with the current Optuna design, since we already have `optuna.exceptions.TrialPruned`.

VladSkripniuk:
Could you help me to understand what has to be done here? It looks like the code provided in the description solves the task. Here is an example:
```
import optuna
from optuna.exceptions import StopStudy

def objective(trial):
    x = trial.suggest_uniform('x', 0, 1)
    return x

EXIT_THRESHOLD = 0.5

def callback(study, trial):
    if trial.value < EXIT_THRESHOLD:
        raise StopStudy

study = optuna.create_study()
study.optimize(objective, callbacks=[callback])
```
which makes the study stop once the `EXIT_THRESHOLD = 0.5` is achieved:
```
[I 2020-03-05 21:12:08,942] Finished trial#0 resulted in value: 0.5258333026159131. Current best value is 0.5258333026159131 with parameters: {'x': 0.5258333026159131}.
[I 2020-03-05 21:12:08,991] Finished trial#1 resulted in value: 0.13087708031224587. Current best value is 0.13087708031224587 with parameters: {'x': 0.13087708031224587}.
Traceback (most recent call last):
  File "example_stop_study.py", line 15, in <module>
    study.optimize(objective, callbacks=[callback])
  File "/home/user/optuna_env/optuna/optuna/study.py", line 319, in optimize
[...]

PR #665, opened 2019-11-05
Add a sampler class based on grid search.

g-votte:
Resolves #656

PR #986, opened 2020-03-05
Exception to exit a study. #966

VladSkripniuk:
Add `optuna.exceptions.StopStudy` for stopping a study when a threshold is passed. #966

### DIFF
--- a/optuna/exceptions.py
+++ b/optuna/exceptions.py
@@ -80,10 +80,11 @@ class DuplicatedStudyError(OptunaError):
 
 
 class StopStudy(OptunaError):
-    """Exception for stopping a study, once a threshold is achieved.
+    """Exception for stopping a study.
 
-    This error is raised when a study achieves the same or better value of objective function
-    as threshold.
+    This error can be raised to stop a study once some exit criterion is met, e.g. once
+    required value of an objective function is achieved or all grid points are processed
+    in :class:`~optuna.samplers.GridSampler`.
     """
 
     pass
--- a/optuna/exceptions.py
+++ b/optuna/exceptions.py
@@ -79,6 +79,16 @@ class DuplicatedStudyError(OptunaError):
     pass
 
 
+class StopStudy(OptunaError):
+    """Exception for stopping a study, once a threshold is achieved.
+
+    This error is raised when a study achieves the same or better value of objective function
+    as threshold.
+    """
+
+    pass
+
+
 class ExperimentalWarning(Warning):
     """Experimental Warning class.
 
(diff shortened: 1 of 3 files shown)


optuna/exceptions.py
```diff
@@ -79,6 +79,16 @@ class DuplicatedStudyError(OptunaError):
     pass
 
 
+class StopStudy(OptunaError):
+    """Exception for stopping a study, once a threshold is achieved.
```
g-votte: Please note that the criteria would not only be exceeding a threshold but for more general purposes. For example, the exception may be raised in `GridSampler` to stop the study after all grid points are evaluated. (See also https://github.com/optuna/optuna/pull/665#discussion_r361424127 for more details.)

Would you mind modifying the docstring to cover more general scenarios?

optuna/study.py
```diff
@@ -594,7 +594,12 @@ def _optimize_sequential(
                 if elapsed_seconds >= timeout:
                     break
 
-            self._run_trial_and_callbacks(func, catch, callbacks, gc_after_trial)
+            try:
+                self._run_trial_and_callbacks(func, catch, callbacks, gc_after_trial)
+            except exceptions.StopStudy:
```
g-votte: Current implementation seems to catch `StopStudy` only when it’s raised in the callback. How about also covering those raised in the objective function?

optuna/exceptions.py
VladSkripniuk: Thank you, I extended the docstring to include more general use cases.

optuna/study.py
VladSkripniuk: What status should we assign to a trial, when `StopStudy` is raised in the objective function? I think `optuna.structs.TrialState.COMPLETE` is a reasonable choice, but there is no value we can assign to the trial in this case. I think this is an issue, so may be we should avoid raising `StopStudy` inside objective function.

g-votte: Good catch. The expected state is not trivial.

On the other hand, it might be a burden for the user to understand that "a callback is acceptable but an objective function is not."

g-votte:
Hi @VladSkripniuk 

I've actually felt that we need more careful discussion before fixing the API. The expected behavior is not trivial around the objective function and callbacks.

I've created a PR to propose another option, which I think makes the behavior more clear: #1025 

Please kindly let me know if you have any thought.

g-votte:
Although this PR better reflects the original design described in #966, we’ll go with #1025 to safely stop the optimization, and close this PR.

Thanks again for your contribution!

PR #988, opened 2020-03-06
Apply black to all files without string normalization.

keisuke-umezawa:


keisuke-umezawa:
Reopen PR in https://github.com/optuna/optuna/pull/1020.

Issue #711, opened 2019-11-18
Use function annotation syntax for Type Hints.


c-bata:
After dropping Python 2.7 support at #710, we can define type hints with function annotation syntax. 
~~Do you have a plan to update the coding style guideline?~~
https://github.com/optuna/optuna/wiki/Coding-Style-Conventions

## Progress

- [x] `optuna/integration/sklearn.py` (#1735)
- [x] `optuna/study.py` - assigned to harpy

## Note to the questioner

We still cannot use variable annotation syntax introduced by [PEP 526](https://www.python.org/dev/peps/pep-0526/) because we supports Python 3.5.

hvy:
Thanks for bringing this up. There is a tracking issue now so let me just leave a reference. https://github.com/optuna/optuna/issues/713. I believe you are correct in that the coding style guideline should be updated.

keisuke-umezawa:
@c-bata 
List of files grepped by `# type:`.

- [ ] setup
- [ ] optuna/visualization/parallel_coordinate
- [ ] optuna/visualization/slice
- [ ] optuna/visualization/intermediate_values
- [ ] optuna/visualization/optimization_history
- [ ] optuna/visualization/utils
- [ ] optuna/visualization/contour
- [ ] optuna/logging
- [ ] optuna/storages/rdb/models
- [ ] optuna/storages/rdb/storage
- [ ] optuna/storages/__init__
- [ ] optuna/storages/base
- [ ] optuna/storages/in_memory
- [ ] optuna/trial
- [ ] optuna/pruners/hyperband
- [ ] optuna/pruners/successive_halving
- [ ] optuna/pruners/nop
- [ ] optuna/pruners/median
- [ ] optuna/pruners/base
- [ ] optuna/pruners/percentile
- [ ] optuna/integration/cma
- [ ] optuna/integration/lightgbm
- [ ] optuna/integration/mxnet
- [ ] optuna/integration/chainermn
- [ ] optuna/integration/lightgbm_tuner/optimize
- [ ] optuna/integration/lightgbm_tuner/alias
- [ ] optuna/integration/lightgbm_tuner/__init__
- [ ] optuna/integration/lightgbm_tuner/sklearn
- [ ] optuna/integration/fastai
- [ ] optuna/integration/keras
- [ ] optuna/integration/__init__
- [ ] optuna/integration/pytorch_ignite
- [ ] optuna/integration/tensorflow
- [ ] optuna/integration/skopt
[...]

PR #989, opened 2020-03-06
use function annotation syntax for type hints

bigbird555:
This PR is related to issue #711


setup.py
```diff
@@ -10,8 +10,7 @@
 from typing import Optional  # NOQA
```
c-bata: I guess you can remove `# NOQA` comments.

bigbird555: Okay

PR #993, opened 2020-03-08
use function annotation syntax for type hints

bigbird555:
This PR is related to issue #711


optuna/visualization/parallel_coordinate.py
```diff
@@ -7,21 +7,18 @@
 from optuna.visualization.utils import _check_plotly_availability
 from optuna.visualization.utils import is_available
 
-if type_checking.TYPE_CHECKING:
-    from typing import DefaultDict  # NOQA
-    from typing import List  # NOQA
-    from typing import Optional  # NOQA
-
-    from optuna.study import Study  # NOQA
+from typing import DefaultDict  # NOQA
```
himkt: You can remove `# NOQA`, I think.

optuna/visualization/parallel_coordinate.py
```diff
@@ -7,21 +7,18 @@
 from optuna.visualization.utils import _check_plotly_availability
 from optuna.visualization.utils import is_available
 
-if type_checking.TYPE_CHECKING:
```
himkt: You should also remove `from optuna import type_checking` in line:6.

optuna/visualization/parallel_coordinate.py
```diff
@@ -7,21 +7,18 @@
...
 
 if is_available():
     from optuna.visualization.plotly_imports import go
 
 logger = get_logger(__name__)
 
 
-def plot_parallel_coordinate(study, params=None):
-    # type: (Study, Optional[List[str]]) -> go.Figure
+def plot_parallel_coordinate(study: Study, params: Optional[List[str]] = None) -> 'go.Figure':
```
himkt: I encountered the problem of importing `Study` class.
So, I wonder if you should use a forward reference here. What do you think about this?
https://github.com/optuna/optuna/issues/711#issuecomment-597088465

bigbird555: I think the revision above can not pass flake8 CI test....

himkt: Ah, you're right.
You have to import `optuna` or `optuna.study` at the head of `parallel_coordinate.py`.

himkt: And I found it doesn't need to use forward reference since there is no circular importing here.
You can use `from optuna.study import Sutdy` and `Study` here, really sorry for confusing you. :bow:

himkt:
Basically LGTM!

`black` was introduced in https://github.com/optuna/optuna/pull/1030.
I sent you the PR to apply black, please check https://github.com/bigbird555/optuna/pull/1.

himkt:
@bigbird555 

Thank you for revising PR, it looks basically good to me.
Could you please update your CircleCI configuration to our latest master branch?
(same as https://github.com/optuna/optuna/pull/996#issuecomment-600191625)

bigbird555:
@himkt I have updated CircleCI configuration in the above commit. Please check it out !

hvy:
Let me merge this PR as it has 2 approvals. 

I have two minor comments though, one on the import order of `typing` symbols and the second on the logic that's most likely unintentionally changed in `_get_parallel_coordinate_plot`. I'll create a follow up PR to address these points.

Thanks for your contribution!

hvy:
Let me modify the title of this PR to match the format of our release notes.

Issue #733, opened 2019-11-25
Warn of too many trials in visualization features.


g-votte:
The motivation is similar to #732. When the number of trials is too many, outputting some warning would be more user friendly.

michiboo:
Hi can I have a go at this?

g-votte:
@michiboo 
Sure! Please go for it.

michiboo:
@g-votte Hi I created a PR for this issue https://github.com/optuna/optuna/pull/994

hvy:
```python
import time
import optuna


def objective(trial, n_params):
    return sum(trial.suggest_float(f"{i}", 0, 1) for i in range(n_params))


def run_benchmark_plot_function(plot_func, study):
    start = time.time()
    fig = plot_func(study)
    end = time.time()
    print(f"{plot_func}: {end - start:.2f} s")
    fig.show()


if __name__ == "__main__":
    n_params = 32
    n_trials = 1000

    study = optuna.create_study(storage="sqlite:///994.db", study_name="994", load_if_exists=True)
    study.optimize(lambda trial: objective(trial, n_params=n_params), n_trials=n_trials)

    print(f"Trials {len(study.trials)}")

    # run_benchmark_plot_function(optuna.visualization.plot_contour, study)  # Too slow.
    run_benchmark_plot_function(optuna.visualization.plot_intermediate_values, study)
    run_benchmark_plot_function(optuna.visualization.plot_optimization_history, study)
    run_benchmark_plot_function(optuna.visualization.plot_parallel_coordinate, study)
    run_benchmark_plot_function(optuna.visualization.plot_param_importances, study)
    run_benchmark_plot_function(optuna.visualization.plot_slice, study)
```

Just leaving a record and some code for verifying this issue. (Took a quick look as the PR addressing this issue was being stale).
The problem seem to mostly derive from `plot_contour` which basically hangs (several minutes without returning). The other plot functions are more or less instantaneous, except for `plot_param_importance` which take a couple of seconds to finish.

I assume plotly is somewhat slow computing the contours. As for warnings, I believe we only need to raise one for this function. Maybe we can even improve the speed in some what by extracting a subset of trials, but I have not tried this yet.

hvy:
I believe this is still an open issue as #994 unfortunately seems stale.

blazespinnaker:
Can we reopen this?  Adding a parameter to sample randomly would be a good idea.  It bugs out quite easily.  At the very least a doc warning would be helpful

PR #994, opened 2020-03-08
add warning if n_trial if > 1000 in optmize methods

michiboo:
This PR is created to address the issue for #733


optuna/integration/chainermn.py
```diff
@@ -124,6 +124,11 @@ def optimize(
         the absence of ``n_jobs`` argument.
         """
 
+        if n_trials:
+            if n_trials > 1000:
+                warnings.warn("visualization features might not be stable "
+                              "n_trials is more than 1000")
```
g-votte: Is the number always around 1,000?

In case it's not true, how about just mentioning, for example: `Plotting {} trials, which may cause unstable behavior of the visualization feature with too many trials.`

optuna/integration/chainermn.py
```diff
@@ -124,6 +124,11 @@ def optimize(
         the absence of ``n_jobs`` argument.
         """
 
+        if n_trials:
```
g-votte: I think we’ve better put the warning logs in the visualization functions rather than in the optimize method.  Otherwise, it impacts even the user not using visualization features.

michiboo: hi but even plot_intermediate_values don't have n_trials parameter anymore

sile: Agree with @g-votte. I think that, within visualization functions, we can get the number of trials in a study via `len(study.trials)`.

toshihikoyanase:
@michiboo I found that this PR is labeled as `stale`. Please let me know if you have any problems.

hvy:
This should be relevant https://github.com/optuna/optuna/issues/733#issuecomment-655334709.

hvy:
Just leaving a note but I believe it's sufficient to warn for the contour plot given https://github.com/optuna/optuna/issues/733#issuecomment-655334709.

michiboo:
> Just leaving a note but I believe it's sufficient to warn for the contour plot given [#733 (comment)](https://github.com/optuna/optuna/issues/733#issuecomment-655334709).

Ok, I just solved the conflicts and remove other warnings except for the contour plot

optuna/visualization/_contour.py
```diff
@@ -61,6 +62,11 @@ def objective(trial):
         A :class:`plotly.graph_objs.Figure` object.
     """
 
+    if len(study.trials) > 1000:
+        warnings.warn(
+            "Plotting {} trials, which may cause unstable behavior"
+            " of the visualization feature with too many trials."
```
hvy, line 68: Should the number for actual trials be included in the string considering there's a placeholder?

hvy:
Thanks for your quick action and I'm sorry that the reply took time.

The change personally LGTM except for the comment above. I'd ideally place the warning condition based on number of not all trials but completed ones (see `_get_contour_plot`) but as it wasn't pointed out in previous review comments, feel free to ignore it.

optuna/visualization/_contour.py
michiboo: @hvy  Hi I changed it to actual number of 1000 and also the number are now based on completed trials

PR #1003, opened 2020-03-10
Embed plotly figures in documentation

harupy:
This PR proposes to embed Plotly figures that `visualization` module generates in the documentation ([discussion](https://gitter.im/optuna/optuna?at=5e660e04cc10be22a074ec11) with @hvy on Gitter).

demo

```bash
# Commands to generate the doc above.
cd docs/source/scripts
python plot_contour.py
cd ../..
make html
```


optuna/visualization/contour.py
```diff
@@ -50,6 +50,9 @@ def objective(trial):
 
             optuna.visualization.plot_contour(study, params=['x', 'y'])
 
+        .. raw:: html
+            :file: ../../docs/source/plotly_figures/plot_contour.html
```
harupy: `:file:` represents relative path from `contour.py`.

harupy:
The doc build in the CI failed with:

```
InputError: [Errno 2] No such file or directory: '../venv/lib/python3.8/site-packages/docs/source/plotly_figures/plot_contour.html'.
```

This is probably because Sphinx uses `optuna` installed under `venv`, not the one in the repo.

harupy:
The doc seems to be build properly.

https://138-220454778-gh.circle-artifacts.com/0/docs/build/html/reference/visualization.html

harupy:
A few things I'd like to discuss:
1. Does it make sense to move the example code and figure after the `Returns:` section?
2. Should we make the result deterministic?

.circleci/config.yml
```diff
@@ -97,6 +97,10 @@ jobs:
             cd docs
             make html
 
+      - store_artifacts:
+          name: store documentation
+          path: ./docs/build
```
hvy: This is very useful. I didn't know that we could preview documentations on CircleCI. This is not a strong opinion but this change could actually be separated into it's own PR.

hvy:
> Does it make sense to move the example code and figure after the Returns: section?

I see your point. I personally don't have a very strong opinion about this but yes it might be better as related information (function signature and argument/return value docs.) get closer.

> Should we make the result deterministic?

Right, I think we should if possible to make future diffs deterministic.

.circleci/config.yml
harupy: Got it. I'll revert this change and open a separate RP for this change.

docs/source/scripts/plot_contour.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def main():
+    def objective(trial):
+        x = trial.suggest_uniform('x', -100, 100)
+        y = trial.suggest_categorical('y', [-1, 0, 1])
+        return x ** 2 + y
+
+    study = optuna.create_study()
+    study.optimize(objective, n_trials=10)
```
harupy:
```suggestion
    sampler = TPESampler(seed=10)
    study = optuna.create_study(sampler=sampler)
    study.optimize(objective, n_trials=10)
```
@hvy Does this change make the result deterministic?

hvy: Yes, that should be enough (besides the fact that you need to import the `TPESampler`).

docs/Makefile
```diff
@@ -8,6 +8,8 @@ SPHINXPROJ    = Optuna
 SOURCEDIR     = source
 BUILDDIR      = build
 
+export PYTHONPATH := $(realpath ..):$(PYTHONPATH)  
+
```
harupy: Should this change also be separated?

docs/source/plotly_figures/plot_contour.html
```diff
@@ -0,0 +1,24 @@
+<html>
+<head><meta charset="utf-8" /></head>
+<body>
+    <div>
+        
+        
+            <div id="c7a563a1-e56b-4419-93c4-3354a268588a" class="plotly-graph-div" style="height:100%; width:100%;"></div>
```
harupy: `id` here is generated by `uuid.uuid4()`. Plotly doesn't provide an option to make this deterministic.

https://github.com/plotly/plotly.py/blob/65625d49226632abcf932d530aa9c90ba1bc59c1/packages/python/plotly/plotly/io/_html.py#L134

harupy: We could define the function below and replace the id value with some deterministic value. 

```python
import re

def replace_id(html, new_id):
    old_id = re.search(r'<div id="([0-9a-z-]+)"', html).group(1)
    return html.replace(old_id, new_id)
```

harupy:
The CI  failed with:

```
#!/bin/bash -eo pipefail
. venv/bin/activate
mypy .
optuna/visualization/parallel_coordinate.py:108: error: Argument 1 to "append" of "list" has incompatible type "Dict[str, Sequence[Any]]"; expected "Dict[str, object]"
optuna/visualization/parallel_coordinate.py:108: note: "Dict" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
optuna/visualization/parallel_coordinate.py:108: note: Consider using "Mapping" instead, which is covariant in the value type
optuna/integration/sklearn.py:385: error: Incompatible types in assignment (expression has type "float", variable has type "Union[<subclass of "float" and "Number">, <subclass of "str" and "Number">]")
optuna/integration/sklearn.py:389: error: Incompatible types in assignment (expression has type "float", variable has type "Union[<subclass of "float" and "Number">, <subclass of "str" and "Number">]")
optuna/integration/sklearn.py:396: error: Incompatible return value type (got "List[Union[float, <subclass of "str" and "Number">]]", expected "List[float]")
Found 4 errors in 2 files (checked 121 source files)

Exited with code exit status 1
```

@hvy @toshihikoyanase 
Did the new version of `mypy` (released 2 hours ago) introduce this issue?
https://pypi.org/project/mypy/0.770/#history

hvy:
Seems so. I created a quick fix here https://github.com/optuna/optuna/pull/1009.

docs/source/plotly_figures/plot_contour.html
hvy: Ah good catch. I think we can allow this kind of diff though. What do you think?

harupy: Agree on allowing this diff.

docs/Makefile
hvy: I'm wondering if you cannot append `'plotly_figures/plot_contour.html'` to `html_static_path ` https://github.com/optuna/optuna/blob/master/docs/source/conf.py#L101 instead of modifying the `PYTHONPATH`.

docs/source/scripts/plot_contour.py
```diff
@@ -0,0 +1,23 @@
...
+        return x ** 2 + y
+
+    sampler = optuna.samplers.TPESampler(seed=10)
+    study = optuna.create_study(sampler=sampler)
+    study.optimize(objective, n_trials=10)
+
+    fig = optuna.visualization.plot_contour(study, params=['x', 'y'])
+    fig_div = plotly.offline.plot(fig, output_type='div', include_plotlyjs=False, auto_open=False)
+
+    with open('../plotly_figures/plot_contour.html', 'w') as f:
```
hvy: A detail but how about creating the `plotly_figures` directory if it doesn't exist, instead of raising an error. It's acceptable "overengineering" I think in favor of usability.

harupy: You're right. I'll fix this.

docs/Makefile
harupy: We can use store the HTML files in the `_static` folder and embed them in `iframe` as follows. In this way, we don't need to modify `PYTHONPATH`.

```diff
        .. raw:: html
-            :file: ../../docs/source/plotly_figures/plot_contour.html
+            <iframe src="../_static/plot_contour.html" width="100%" height="500px" frameborder="0">
+            </iframe>
```

hvy: Great. I am just wondering what the benefits of using an iframe are. Is it cleaner in some sense (mostly thinking about style properties that you included, since they're not required with `:file:`) or can the file not be found otherwise?

harupy: > or can the file not be found otherwise?

Correct.

harupy: As you pointed out, we need to set some attributes of an iframe to render the plot properly.

harupy: @hvy Which way would you prefer (modifying `PYTHONPATH` or using an iframe)?

hvy: Sorry for the delayed response. I think it's better not to modify `PYTHONPATH`. 

Just curious what error you get when trying to read the raw HTML file without loading it in an iframe. It seems to work locally at least.

harupy:
```
.. raw:: html
    :file: _static/plot_contour.html
```

gives:

```
InputError: [Errno 2] No such file or directory: '../optuna/visualization/_static/plot_contour.html'.
```

hvy: I see. As I wrote above, it worked for me locally. But let's proceed with an iframe if that works as well. Could you go ahead and remove the `PYTHONPATH` and introduce the changes discussed in this thread?

harupy: > it worked for me locally.

Was the plot rendered properly?

> Could you go ahead and remove the PYTHONPATH and introduce the changes discussed in this thread?

Sure.

harupy:
The doctest for `plot_intermediate_values` returns an empty figure.

Screen Shot 2020-03-14 at 17 14 16


https://31149-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/visualization.html#optuna.visualization.plot_intermediate_values

hvy:
Ah, the example for the intermediate value plot should be fixed. It doesn't report any intermediate values at all, that's why it's empty. We need to change the example to utilize an objective function that contains a for loop ("partial fit") calling `Trial.report` in each loop iteration. That's something that we can address outside this PR. Thanks for the catch.

docs/source/scripts/plot_contour.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+import optuna
+import plotly
```
toshihikoyanase: Please separate the third party library and the local library (i.e., optuna). (c.f., https://github.com/optuna/optuna/issues/514)

docs/source/scripts/plot_contour.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def main():
+    def objective(trial):
+        x = trial.suggest_uniform('x', -100, 100)
+        y = trial.suggest_categorical('y', [-1, 0, 1])
+        return x ** 2 + y
```
toshihikoyanase: Is it possible to move the definition of `objective` to the outside of `main` to reduce the nest level?

harupy:
@toshihikoyanase Thanks! I'll fix them in this PR.

toshihikoyanase:
@harupy Thank you for your swift response. 

Similar changes can also be applied to the scripts for the other plots. So, please check them too.
- docs/source/scripts/plot_intermediate_values.py,
- docs/source/scripts/plot_optimization_history.py,  
- docs/source/scripts/plot_parallel_coordinate.py, and 
- docs/source/scripts/plot_slice.py.

harupy:
Do I need to apply Black?

toshihikoyanase:
> Do I need to apply Black?

Ah, yes. `black` was introduced by https://github.com/optuna/optuna/pull/1030. If the change is bothering you, please let me merge this branch first and fix the code style in a follow-up PR.

PR #1009, opened 2020-03-11
Fix `mypy==0.770` errors.

hvy:
Fixes `mypy` errors arising from https://github.com/python/mypy/releases/tag/v0.770.

### DIFF
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -2,6 +2,7 @@ from logging import DEBUG
 from logging import INFO
 from logging import WARNING
 from numbers import Integral
+from numbers import Number
 from time import time
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


optuna/integration/sklearn.py
```diff
@@ -367,7 +366,7 @@ def _partial_fit_and_score(
             if self.error_score == 'raise':
                 raise e
 
-            elif isinstance(self.error_score, Number):
+            elif isinstance(self.error_score, float):
```
toshihikoyanase: The current type hint for `self.error_score` is `Union[str, float]`, but the correspondence of `RandomSearchCV` in scikit-learn seems to be `str, Number` (c.f., https://github.com/scikit-learn/scikit-learn/blob/95d4f0841/sklearn/model_selection/_search.py#L1307-L1311).  So, I think keeping the `Number` here would avoid issues in real use cases.

hvy: You are right. I relaxed the type hints to allow `numbers.Number`s.

PR #1013, opened 2020-03-11
Update PyTorch Lightning example for 0.7.1 version.

### LINKED ISSUE #997: `optuna.integration.PyTorchLightningPruningCallback` does not work with `pytorch_lightning==0.7.1`.
## Environment

- Optuna version: 1.2.0
- Python version: 3.7.2
- OS: macOS 10.13.6
- (Optional) Other libraries and their versions: pytorch_lightning==0.7.1

Command
```console
pytest tests/integration_tests/test_pytorch_lightning.py
```

Error message:
```console
self = <optuna.integration.pytorch_lightning.PyTorchLightningPruningCallback object at 0x7f2194069a20>
logs = {'accuracy': 0.0}

    def check_metrics(self, logs):
        monitor_val = logs.get(self.monitor)
        error_msg = (f'Early stopping conditioned on metric `{self.monitor}`'
                     f' which is not available. Available metrics are:'
                     f' `{"`, `".join(list(logs.keys()))}`')
    
        if monitor_val is None:
            if self.strict:
>               raise RuntimeError(error_msg)
E               RuntimeError: Early stopping conditioned on metric `val_loss` which is not available. Available metrics are: `accuracy`

venv/lib/python3.6/site-packages/pytorch_lightning/callbacks/early_stopping.py:81: RuntimeError

```

## Additional context

The base class of `PyTorchLightningPruningCallback` (i.e., `EarlyStopping`) is updated by 0.7.x.
For example,
1. the `check_metrics` method is newly added, and
1. the arguments of `on_epoch_end` are changed.

For further details, please refer to the following code:
- https://github.com/PyTorchLightning/pytorch-lightning/blob/0.7.1/pytorch_lightning/callbacks/early_stopping.py
- https://github.com/PyTorchLightning/pytorch-lightning/blob/0.5.3.3/pytorch_lightning/callbacks/pt_callbacks.py

To resolve this issue, we may need to consider backward compatibility. Do we support both v0.7.x and 0.5.x?

festeh:
Fixes #997

Introduces backwards incompatible changes as there are too many in the 0.7.1 version. Suggestions are welcome.

### DIFF
--- a/examples/pytorch_lightning_simple.py
+++ b/examples/pytorch_lightning_simple.py
@@ -136,7 +136,7 @@ class LightningNet(pl.LightningModule):
 
 
 def objective(trial):
-    # filenames for each trial must be made unique in order to access each checkpoint
+    # Filenames for each trial must be made unique in order to access each checkpoint.
     checkpoint_callback = pl.callbacks.ModelCheckpoint(
         os.path.join(MODEL_DIR, "trial_{}".format(trial.number), "{epoch}"), monitor="val_acc"
     )
--- a/examples/pytorch_lightning_simple.py
+++ b/examples/pytorch_lightning_simple.py
@@ -23,7 +23,7 @@ import pkg_resources
 import shutil
 
 import pytorch_lightning as pl
-from pytorch_lightning.logging import LightningLoggerBase
+from pytorch_lightning import Callback
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
@@ -35,8 +35,8 @@ from torchvision import transforms
 import optuna
 from optuna.integration import PyTorchLightningPruningCallback
 
-if pkg_resources.parse_version(pl.__version__) < pkg_resources.parse_version("0.6.0"):
-    raise RuntimeError("PyTorch Lightning>=0.6.0 is required for this example.")
+if pkg_resources.parse_version(pl.__version__) < pkg_resources.parse_version('0.7.1'):
+    raise RuntimeError('PyTorch Lightning>=0.7.1 is required for this example.')
 
 PERCENT_VALID_EXAMPLES = 0.1
 BATCHSIZE = 128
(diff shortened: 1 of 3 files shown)


festeh:
Uh, it seems that 0.7.1 is incompatible with Python 3.5. I'm still thinking that modifying the example for 0.7.1 is a right thing to do, but not so sure now - we could discuss alternative solutions.

toshihikoyanase:
Hi @festeh! Thank you for your PR. 

> Uh, it seems that 0.7.1 is incompatible with Python 3.5.

The [pypi page](https://pypi.org/project/pytorch-lightning/) of pytorch-lightning says that it requires Python>=3.6. IMO, PyTorch Lightning is still actively developed, and it is difficult for Optuna developers to keep using the development resource to provide backward compatibility.
 So, I think we can keep your modification and just skip the CI for PyTorch Lightning on Python 3.5.

keisuke-umezawa:
@festeh 
FYI
If you change this part, you can select lightning version for python3.5.

https://github.com/optuna/optuna/blob/master/setup.py#L132-L133

hvy:
If you're busy, someone could probably take over this PR. Just let us know.

festeh:
Okay, I'll try to update the PR in next few days, sorry for the delay

festeh:
@hvy I think it's now ready for review. Thanks for your patience.

examples/pytorch_lightning_simple.py
```diff
@@ -146,31 +136,29 @@ def val_dataloader(self):
 
 
 def objective(trial):
-    # PyTorch Lightning will try to restore model parameters from previous trials if checkpoint
-    # filenames match. Therefore, the filenames for each trial must be made unique.
+    # filenames for each trial must be made unique in order to access each checkpoint
```
hvy, line 139:
```suggestion
    # Filenames for each trial must be made unique in order to access each checkpoint.
```

examples/pytorch_lightning_simple.py
```diff
@@ -46,20 +46,15 @@
 MODEL_DIR = os.path.join(DIR, "result")
 
 
-class DictLogger(LightningLoggerBase):
-    """PyTorch Lightning `dict` logger."""
+class MetricsCallback(Callback):
+    """PyTorch Lightning metric callback"""
```
hvy, line 50:
```suggestion
    """PyTorch Lightning metric callback."""
```

festeh:
Thanks, fixed!

setup.py
```diff
@@ -86,9 +86,7 @@ def get_extras_require() -> Dict[str, List[str]]:
                 "dask[dataframe]",
                 "dask-ml",
                 "keras",
-                # TODO(toshihikoyanase): Remove the version constraint after resolving the issue
-                # https://github.com/optuna/optuna/issues/997.
-                "pytorch-lightning<0.7.0",
+                "pytorch-lightning>=0.7.1",
```
toshihikoyanase, line 89: We may not need to install `pytorch-lightning` for Python 3.5 since the supported version of PyTorch Lightning is 3.6 or newer. But, this is a trivial issue and I'll create a followup if necessary.

toshihikoyanase: I submitted a PR (https://github.com/optuna/optuna/pull/1193). If you have any comments, please let me know.

Issue #534, opened 2019-09-17
OptGBM was released


Y-oHr-N:
Hi there,

Today, I have released OptGBM.
https://github.com/Y-oHr-N/OptGBM

This package provides a scikit-learn compatible estimator that tunes the hyperparameters of LightGBM with Optuna.

If you like this package, please consider merging into Optuna. I hope this package will contribute to the development of Optuna.

Finally, #507 implements similar features, so I might be able to help you.

Thanks.

smly:
Thank you for sharing us your nice package.
It seems well written and easy to use :+1:.
I think there are many things to consider: time-vs-performance, applicability of pruning, etc.
Let us consider how we can collaborate after releasing #507.

smly:
Sorry for my late response.
As you mentioned, some features are duplicated with LightGBM Tuner. And merging the entire code of OptGBM might be confusing to users.

May I introduce your library as an Optuna usage example in our `examples/README.md`?
Thanks again for your contributions to the Optuna community!

Y-oHr-N:
Sorry for the late reply.

> May I introduce your library as an Optuna usage example in our `examples/README.md`?

That is a good news for me.

We hope to discuss merging at the Optuna development sprint event on March 1. Current differences between LightGBMTuner and OptGBM that I know are as follows.
https://speakerdeck.com/yohrn/8th-place-solution-of-autowsl-2019?slide=14

smly:
@Y-oHr-N, thanks for your comments! I'm looking forward to discussing OptGBM with you at the event.
Now that https://github.com/optuna/optuna/pull/896 is merged and OptGBM is listed in `examples/README.md`. Let me close this issue as resolved. Thanks again for your great work!

Y-oHr-N:
Hi @smly,

I heard disappointing news that the event will be canceled.

So there are two things I want to check here.

1. Does the Optuna development team agree to merge OptGBM into Optuna?
2. If you agree, what are the topics we need to discuss, that is, what can be the obstacles to merging?

smly:
@Y-oHr-N, I'm sorry that I cannot meet you at the dev sprint event.
Today, Optuna development members including me discussed integrating OptGBM into Optuna. In order to answer your question, we need to discuss many things with you, such as remaining development tasks, API, maintenance, and existing issues in OptGBM. If it's okay with you, let's schedule a video conference to discuss these things with us.

Y-oHr-N:
@smly , OK. How do we communicate?

[...]

PR #1018, opened 2020-03-11
Merge OptGBM into Optuna

Y-oHr-N:
#### Reference Issues/PRs

This PR is related to #534.

#### What does this implement/fix? Explain your changes.

This PR proposes to merge [OptGBM](https://github.com/Y-oHr-N/OptGBM) that is a third-party library into Optuna.


Y-oHr-N:
We should discuss the following:

- search space
- default values of some parameters
    - n_estimators
    - early_stopping_rounds
- handling duplicate parameters
    - n_jobs
    - callbacks
- way to save boosters in trials
- necessity of weights in fold averaging
- etc.

smly:
Thanks a lot for your contribution!
Let me review the code after discussing the topics you have mentioned above.

**search space**

The search space you defined looks good to me. Do you have any special concerns about it?

**default values**

As for the default values, I have one suggestion:

* n_estimators=1000: LGBM
* n_jobs=1: LGTM
* early_stopping_rounds=10. LGTM
* refit=False: I'm not sure `refit=False` is better, but I think it is more stable than `refit=True`.
* n_tirals=20: I guess `n_trials=20` is too small to tune. How is 30 or 40?

smly:
**handling duplicate parameters**

What does it mean?

**way to save boosters in trials**
**necessity of weights in fold averaging**

I guess the above two can be seen as optional features. correct? If yes, it might be good to develop in a different PR for this.

Y-oHr-N:
#### search space

LightGBMTuner and OptGBM models should have the same search space to avoid user confusion.

#### default values of some parameters

I agree with setting `refit` to True.

I'm a little worried that the difference between the default values may confuse users. The default values of parameters of related classes are as follows.

- `lightgbm.LGBMModel`
    - n_estimators=100
    - n_jobs=-1
    - early_stopping_rounds=None
- `optuna.integration.LightGBMTuner`
    - n_estimators=1000
    - n_jobs=0
    - early_stopping_rounds=None
- `optuna.integration.OptunaSearchCV`
    - n_trials=10
    - refit=True

BTW, LightGBMTuner automatically determined `early_stopping_rounds` until version 0.19. Why was this fixed?
https://github.com/optuna/optuna/blob/v0.19.0/optuna/integration/lightgbm_tuner/optimize.py#L345-L350

#### handling duplicate parameters

I'm sorry for lack of explanation.

`callbacks` is a parameter passed to `optimize` method on `Study`. And it is also a parameter passed to `cv` function in LightGBM. OptGBM models pass `callbacks` to LightGBM, so users cannot use callback functions for Optuna.

The same can be said of `n_jobs`.

#### way to save boosters in trials

Sure. I'll consider it.

#### necessity of weights in fold averaging

It can be implemented immediately. I'm struggling with the name of the parameter that controls this behavior. How is `enable_weighted_averaging`?

Y-oHr-N:
#### support for older versions of LightGBM

The current implementation fails tests with LightGBM 2.2.1. From which version do you need to guarantee operation?

smly:
Sorry for my late response.

**Support for older versions of LightGBM**

Supporting many versions requires a lot of workarounds and maintenance costs.
How about supporting the latest minor version, v2.3.0 or later?
It seems reasonable in terms of release frequency.

(Optional) There is no guarantee that all users will use the latest version.
It is useful if you can partially support an older version with minimal effort.

smly:
**search space**

> LightGBMTuner and OptGBM models should have the same search space to avoid user confusion.

Partially agree. LightGBM Tuner and OptGBM do not use the exact same algorithm so far.
So I don't think it needs to match the search space exactly. Of course, the same is better.

**default values**

> LightGBMTuner automatically determined early_stopping_rounds until version 0.19. Why was this fixed?

I removed them to make the problem simpler. I believe that `learning_rate`, `n_estimators`, and `early_stopping_rounds` are appropriate to be solved as a different problem to reconcile the trade-off between execution time and loss.

> I'm a little worried that the difference between the default values may confuse users.

Good catch. `n_estimators=100` in LightGBM Tuner is better to avoid confusion.
As for `n_jobs=0`, I think it's not a problem. `n_jobs=0` is also the default parameter in LightGBM. https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_threads

smly:
Note for reviewers:

This PR includes the addition of version constraint, `lightgbm>=2.2.2`. I think it is reasonable.
This is required to support pandas categorical datatype. Pandas categorical datatype is useful and widely used to save memory usage.

optuna/integration/lightgbm_tuner/sklearn.py
```diff
@@ -1,58 +1,1309 @@
...
+                Predicted values.
+        """
+        X = check_X(X, accept_sparse=True, estimator=self, force_all_finite=False)
+
+        return self.booster_.predict(X, num_iteration=num_iteration)
+
+
+# alias classes
+OGBMClassifier = LGBMClassifier
+OGBMRegressor = LGBMRegressor
```
smly: For simplification, how about removing aliases? Without reading the source code, users have no way of knowing the difference between OGBMClassifier and LGBMClassifier.

optuna/integration/lightgbm_tuner/sklearn.py
```diff
@@ -1,58 +1,1309 @@
...
+        return np.average(results, axis=0, weights=self.weights)
+
+    def predict(self, X: TwoDimArrayLikeType, **kwargs: Any) -> np.ndarray:
+        results = [b.predict(X, **kwargs) for b in self.boosters]
+
+        return np.average(results, axis=0, weights=self.weights)
+
+
+class LGBMModel(lgb.LGBMModel):
+    """Base class for models in OptGBM."""
```
smly:
```suggestion
    """Base class for models."""
```

setup.py
```diff
@@ -88,7 +88,8 @@ def get_extras_require() -> Dict[str, List[str]]:
             "bokeh<2.0.0",
             "chainer>=5.0.0",
             "cma",
-            "lightgbm",
+            "lightgbm>=2.2.2",
+            "matplotlib",
```
smly: Is this required? If yes, could you tell me why it is necessary?

```suggestion
```

tests/integration_tests/lightgbm_tuner_tests/test_sklearn.py
```diff
@@ -0,0 +1,332 @@
...
+
+    assert isinstance(X, np.ndarray)
+    assert isinstance(y, np.ndarray)
+    assert isinstance(sample_weight, np.ndarray)
+
+
+def test_ogbm_classifier() -> None:
+    pytest.importorskip("sklearn", minversion="0.20.0")
+
+    # from sklearn.utils.estimator_checks import check_estimator
```
smly:
```suggestion
```

tests/integration_tests/lightgbm_tuner_tests/test_sklearn.py
```diff
@@ -0,0 +1,332 @@
...
+    from sklearn.utils.estimator_checks import check_estimators_pickle
+    from sklearn.utils.estimator_checks import check_set_params
+
+    clf = OGBMClassifier()
+    name = clf.__class__.__name__
+
+    # check_estimator(clf)
```
smly:
```suggestion
```

tests/integration_tests/lightgbm_tuner_tests/test_sklearn.py
```diff
@@ -0,0 +1,332 @@
...
+
+    check_estimators_pickle(name, clf)
+    check_set_params(name, clf)
+
+
+def test_ogbm_regressor() -> None:
+    pytest.importorskip("sklearn", minversion="0.20.0")
+
+    # from sklearn.utils.estimator_checks import check_estimator
```
smly:
```suggestion
```

tests/integration_tests/lightgbm_tuner_tests/test_sklearn.py
```diff
@@ -0,0 +1,332 @@
...
+    from sklearn.utils.estimator_checks import check_estimators_pickle
+    from sklearn.utils.estimator_checks import check_set_params
+
+    reg = OGBMRegressor()
+    name = reg.__class__.__name__
+
+    # check_estimator(reg)
```
smly: It's preferable to delete, not comment out.

```suggestion
```

setup.py
```diff
@@ -88,7 +88,8 @@ def get_extras_require() -> Dict[str, List[str]]:
             "bokeh<2.0.0",
             "chainer>=5.0.0",
             "cma",
-            "lightgbm",
+            "lightgbm>=2.2.2",
```
smly: I think this constraint is reasonable. This is required to support pandas categorical datatype. Pandas categorical datatype is useful and widely used to save memory usage.

setup.py
Y-oHr-N: This is required for `test_plot_importance`.

Y-oHr-N:
#### search space

> So I don't think it needs to match the search space exactly. Of course, the same is better.

Then, if there is a problem, let's fix it with another PR.

#### default values

Based on the previous discussion, I would modify the default values as follows:

- before
  - n_estimators=1000
  - n_jobs=1
  - n_trials=20
  - refit=False
  - early_stopping_rounds=10
- after
  - n_estimators=100
  - n_jobs=-1
  - n_trials=40
  - refit=True
  - early_stopping_rounds=10

#### support for older versions of LightGBM

> How about supporting the latest minor version, v2.3.0 or later?

No problem. OptGBM works fine with v2.2.2 or later.

optuna/integration/lightgbm_tuner/sklearn.py
```diff
@@ -1,58 +1,1294 @@
...
+
+        if is_best_trial:
+            boosters = callbacks[0]._boosters  # type: ignore
+            representations = []  # type: List[str]
+
+            for b in boosters:
+                b.free_dataset()
+                representations.append(b.model_to_string())
+
+            trial.study.set_user_attr("representations", representations)
```
smly: This implementation stores boosting models in optuna's storage.
SQLite can store up to 1 billion bytes in a row of string fields by default[1].
In RDB (MySQL), on the other hand, the maximum row field size of `TrialUserAttributeModel` is 2048 bytes[2].
I think this is too small to save the boosting model. What do you think about this?
Is it possible to do parallel execution in a practical case?

----
[1] https://www.sqlite.org/limits.html
[2] https://github.com/optuna/optuna/blob/master/optuna/storages/rdb/models.py#L276-L282

smly:
Thanks for your changes and responses! LGTM about these topics. > search space, default values, support for older version of LightGBM

optuna/integration/lightgbm_tuner/sklearn.py
Y-oHr-N: Thank you for pointing this out. I did not consider at all.

In the example below, boosters consume more than 10 MB of memory.
I can't think of a solution other than to save boosters to files instead.

```python
def total_size(o, handlers={}, verbose=False):
    """ Returns the approximate memory footprint an object and all of its contents.

    Automatically finds the contents of the following builtin containers and
    their subclasses:  tuple, list, deque, dict, set and frozenset.
    To search other containers, add handlers to iterate over their contents:

        handlers = {SomeContainerClass: iter,
                    OtherContainerClass: OtherContainerClass.get_elements}

    """
    # See https://code.activestate.com/recipes/577504/
    ...


import optuna.integration.lightgbm as lgb
from sklearn.datasets import load_digits

X, y = load_digits(return_X_y=True)

clf = lgb.LGBMClassifier(n_estimators=100, random_state=0)
clf.fit(X, y, early_stopping_rounds=None)
total_size(clf.study_.user_attrs["representations"])  # 10743344 bytes

clf = lgb.LGBMClassifier(n_estimators=1000, random_state=0)
clf.fit(X, y, early_stopping_rounds=None)
total_size(clf.study_.user_attrs["representations"])  # 19605949 bytes
```

smly: @Y-oHr-N, thank you for continuing to look at this PR!
I tested the almost same code with using RDB storage (MySQL). The error message is here: https://gist.github.com/smly/ec26cff6dec17568d97006a573dda5f8
As for supporting parallel execution, I think the discussion on #1076 will help resolve the issue. It is ongoing PR, but IMO (1) and (3) are preferable on practical usage. Your comments or insights for this problem are welcome!

Y-oHr-N: I fixed in Y-oHr-N/OptGBM#74 to save the booster as a file. If you like it, I'll fix it in a similar way.

smly: Thank you for the updates, I will check the changes.

smly: In parallel execution, different processes may write to the same file name at the same time in `dump()` method. This will corrupt the output file.

For example, if another process updates the best score while a process is outputting a file, multiple processes will output the same file. Is my understanding correct?

Y-oHr-N: You are right. Here are my solutions to address this issue:

(1) Lock a file while writting. (Exclusive control can be achieved using [fcntl](https://docs.python.org/3/library/fcntl.html), but it is not available on Windows)
(2) Separate directories by the trial number. (In this case, boosters are not overwitten)

There are also two patterns for saving boosters.

(a) Save boosters at the end of each trial
(b) Save boosters only if the best score is updated

#1076 employs (2) (trial number in the file name) and (a). These result in increased writing time and file size.

smly: It is quite reasonable not to save all the boosters.
IMO, `fcntl` is a good way to implement exclusive write even though Windows is not supported. For supporting Windows, naive file locking is also an option to implement it.

Let me have some time to think about this. I would like to hear ideas/opinions from other committers.

smly: Sorry for my late review. I will resume the review process from today. I also asked one of the other committers to review. Especially about parallelization. Thanks.

ytsmiling: Using `fcntl` is not a good option, especially when using this Tuner with NFS, as NFS is notorious for broken lock implementations. If storage consumption is a serious issue, you can add a storage compression mechanism via study callbacks, but I think it should be addressed in another PR.

Y-oHr-N: Thanks, @ytsmiling. Based on your review, I think it would be preferable to implement it in the way of (2)(a) and fix it in another PR.

@smily, do you have any suggestions?

smly: @ytsmiling, thanks for your review! Totally agree with your suggestion. I think your idea of using callback for storage compression is great.

@Y-oHr-N, it looks good to me. Thank you for continuing to look at this PR.

hvy:
Thanks @Y-oHr-N for your great effort in creating this PR!

I suspect that @smly wants to fill in here but let me quickly share some information from the maintainer team. As this PR requires (has required) quite some time to review, we think it's requires focused resources. At the same time, @smly has been busy in my understanding. The proper review might happen after the release of v2.0, so I'd just like to let you know. I'm sorry it's dragging out even longer but let's keep it going.

Y-oHr-N:
@smly, I am very sorry for the delay.
The fix has been completed. Please resume your review.

optuna/integration/_lightgbm_tuner/sklearn.py
```diff
@@ -1,49 +1,887 @@
...
+    "rmse": "l2",
+    "huber": "huber",
+    "fair": "fair",
+    "poisson": "poisson",
+    "quantile": "quantile",
+    "mean_absolute_percentage_error": "mape",
+    "mape": "mape",
+    "gamma": "gamma",
+    "tweedie": "tweedie",
+}
```
toshihikoyanase, line 74: IMO, we can reuse `_ALIAS_METRIC_LIST` in [`optuna/integration/_lightgbm_tuner/alias.py`](https://github.com/optuna/optuna/blob/0e9557a0930755127b96613d1aacfc652eeb37f3/optuna/integration/_lightgbm_tuner/alias.py#L43-L104). What do you think?

optuna/integration/_lightgbm_tuner/sklearn.py
```diff
@@ -1,49 +1,887 @@
...
+        self._best_score = self._Booster.best_score
+        self._evals_result = evals_result if evals_result else None
+        self._objective = params["objective"]
+
+        self._Booster.free_dataset()
+
+        return self
+
+
+class LGBMClassifier(LGBMModel, ClassifierMixin):
```
toshihikoyanase, line 492: Could you add this class to `docs/source/reference/integration.rst` to create the reference?

optuna/integration/_lightgbm_tuner/sklearn.py
```diff
@@ -1,49 +1,887 @@
...
+
+        if self._n_classes > 2 or raw_score or pred_leaf or pred_contrib:
+            return result
+
+        preds = result.reshape(-1, 1)
+
+        return np.concatenate([1.0 - preds, preds], axis=1)
+
+
+class LGBMRegressor(LGBMModel, RegressorMixin):
```
toshihikoyanase, line 768: ditto.

optuna/integration/_lightgbm_tuner/sklearn.py
```diff
@@ -1,49 +1,887 @@
...
+        colsample_bytree: float = 1.0,
+        reg_alpha: float = 0.0,
+        reg_lambda: float = 0.0,
+        random_state: Optional[RandomStateType] = None,
+        n_jobs: int = -1,
+        silent: bool = True,
+        importance_type: str = "split",
+        study: Optional[study_module.Study] = None,
+        timeout: Optional[int] = None,
+        train_dir: Union[pathlib.Path, str] = "optgbm_info",
```
toshihikoyanase, line 198: I think we have two discussion points on this line.

1. How about using the same variable name as `LightGBMTuner`?
This variable is passed to `model_dir` of `LightGBMTuner`, so I guess we can rename this variable to `model_dir`.

2. The default value

The default value of `LightGBMTuner`'s `model_dir` is `None` because it can be executed without this directory. How about using the same default value?

toshihikoyanase, line 198: Not a strong opinion, but how about using the keyword-only argument for these Optuna specific arguments. We would like to add some new features including the customizable step feature, and they require more arguments.

optuna/integration/_lightgbm_tuner/sklearn.py
```diff
@@ -1,49 +1,887 @@
...
+        eval_sample_weight: Optional[List[OneDimArrayLikeType]] = None,
+        eval_class_weight: Optional[List[Union[Dict[str, float], str]]] = None,
+        eval_init_score: Optional[List[OneDimArrayLikeType]] = None,
+        eval_group: Optional[List[OneDimArrayLikeType]] = None,
+        eval_metric: Optional[Union[Callable, List[str], str]] = None,
+        early_stopping_rounds: Optional[int] = 10,
+        verbose: bool = True,
+        feature_name: Union[List[str], str] = "auto",
+        categorical_feature: Union[List[int], List[str], str] = "auto",
+        callbacks: Optional[List[Callable]] = None,
```
toshihikoyanase, line 261: Could you tell us the reason why the `callbacks` argument is passed here while `timeout` is set at `__init__`? I think we can set all of them at `fit`.

toshihikoyanase (changes requested):
Thank you for your great PR. I added some early comments. Please take a look.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: Fixed in 41e76fb.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: Fixed in 41e76fb.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: Fixed in a5db770.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: Fixed in ac6dc5b.

Y-oHr-N: Like `LGBMModel`, `OptunaSearchCV` has `study` and `timeout`.
Changing either class may confuse the user.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: According to the scikit-learn documentation, it is recommended not to add too many parameters to `fit_params`.

https://scikit-learn.org/stable/developers/develop.html#fitting
> Depending on the nature of the algorithm, fit can sometimes also accept additional keywords arguments. However, any parameter that can have a value assigned prior to having access to the data should be an \_\_init\_\_ keyword argument. **fit parameters should be restricted to directly data dependent variables**.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: https://github.com/optuna/optuna/pull/1018/checks?check_run_id=1230177914
Sorry, I don't know how to fix the CI error.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: https://github.com/optuna/optuna/pull/1018/checks?check_run_id=1230177914
Sorry, I don't know how to fix the CI error.

fc0712:
What is the status?

optuna/integration/_lightgbm_tuner/sklearn.py
toshihikoyanase: Thank you!

optuna/integration/_lightgbm_tuner/sklearn.py
toshihikoyanase: Thank you for pointing it out. I think we can update `OptunaSearchCV` as well since it is still experimental.
But, we can work on it in a follow-up PR, so please keep the arguments as it is.

optuna/integration/_lightgbm_tuner/sklearn.py
toshihikoyanase: Thank you for your comment. Then, can we move the `callbacks` to `__init__`? I don't think it is data-dependent, and I feel a bit inconsistent.

optuna/integration/_lightgbm_tuner/sklearn.py
toshihikoyanase: I'm sorry, but the log has been deleted. Could you rebase the branch and push it to kick the CI?

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMModel.html#lightgbm.LGBMModel.fit
`callbacks` is an argument that is passed to fit on the original `LGBMModel`.
If we move `callbacks` to `__init__`, it will not be compatible.

toshihikoyanase: Ah, I misunderstood. I thought this `callbacks` is Optuna's callback function, but it is actually the LightGBM's callback.
Thank you for your explanation.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: https://github.com/optuna/optuna/pull/1018/checks?check_run_id=1790196768
Kicked the CI again. Please check.

toshihikoyanase: According to the log of `doctest`, something is wrong with the code example of `LGBMClassifier`:

```console
**********************************************************************
File "../../../../../../../opt/hostedtoolcache/Python/3.8.7/x64/lib/python3.8/site-packages/optuna/integration/_lightgbm_tuner/sklearn.py", line ?, in default
Failed example:
    clf.fit(X, y, eval_set=[(X_valid, y_valid)])
Expected:
    LGBMClassifier(...)
Got:

```
https://github.com/optuna/optuna/pull/1018/checks?check_run_id=1790196559

Do you have any idea about it?

optuna/integration/_lightgbm_tuner/sklearn.py
```diff
@@ -1,49 +1,854 @@
...
+            trials until it receives a termination signal such as Ctrl+C or
+            SIGTERM. This trades off runtime vs quality of the solution.
+
+        model_dir:
+            Directory for storing the files generated during training.
+
+        **kwargs:
+            Other parameters for the model. See
+            http:/lightgbm.readthedocs.io/en/latest/Parameters.html for more
+            parameters. Note, that **kwargs is not supported in sklearn, so it
```
toshihikoyanase, line 567: Please escape `**kwargs` to avoid syntax error.

```suggestion
            parameters. Note, that ``**kwargs`` is not supported in sklearn, so it
```

optuna/integration/_lightgbm_tuner/sklearn.py
```diff
@@ -1,49 +1,854 @@
...
```
toshihikoyanase, line 842: ditto.
```suggestion
            parameters. Note, that ``**kwargs`` is not supported in sklearn, so it
```

optuna/integration/_lightgbm_tuner/sklearn.py
```diff
@@ -1,49 +1,854 @@
...
+    Examples:
+        >>> import optuna.integration.lightgbm as lgb
+        >>> from sklearn.datasets import load_iris
+        >>> from sklearn.model_selection import train_test_split
+        >>> clf = lgb.LGBMClassifier(random_state=0)
+        >>> X, y = load_iris(return_X_y=True)
+        >>> X, X_valid, y, y_valid = train_test_split(X, y, random_state=0)
+        >>> clf.fit(X, y, eval_set=[(X_valid, y_valid)])
+        LGBMClassifier(...)
+        >>> y_pred = clf.predict(X)
```
toshihikoyanase, line 583: Please use the [`doctest` directive](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-doctest) for a doc-test style code block. In addition, could you set `verbose=False`? Currently, this example produces so many log messages.

```suggestion
    Examples:

        .. doctest::

            >>> import optuna.integration.lightgbm as lgb
            >>> from sklearn.datasets import load_iris
            >>> from sklearn.model_selection import train_test_split
            >>> clf = lgb.LGBMClassifier(random_state=0)
            >>> X, y = load_iris(return_X_y=True)
            >>> X, X_valid, y, y_valid = train_test_split(X, y, random_state=0)
            >>> clf.fit(X, y, eval_set=[(X_valid, y_valid)], verbose=False)
            LGBMClassifier(...)
            >>> y_pred = clf.predict(X)
```

In addition, this example causes an error due to the unexpected warning message (i.e., ` [LightGBM] [Warning] min_data_in_leaf is set=20, min_child_samples=20 will be ignored. Current value: min_data_in_leaf=20` ). Do you have any idea to suppress it?

```console
Warning, treated as error:
**********************************************************************
File "../../optuna/integration/_lightgbm_tuner/sklearn.py", line ?, in default
Failed example:
    clf.fit(X, y, eval_set=[(X_valid, y_valid)], verbose=False)
Expected:
    LGBMClassifier(...)
Got:
    [LightGBM] [Warning] min_data_in_leaf is set=20, min_child_samples=20 will be ignored. Current value: min_data_in_leaf=20
    LGBMClassifier(boosting_type='gbdt', class_weight=None, colsample_bytree=1.0,
                   importance_type='split', learning_rate=0.1, max_depth=-1,
                   min_child_samples=20, min_child_weight=0.001, min_split_gain=0.0,
                   model_dir=None, n_estimators=100, n_jobs=-1, num_leaves=31,
                   objective=None, random_state=0, reg_alpha=0.0, reg_lambda=0.0,
                   silent=True, study=None, subsample=1.0, subsample_for_bin=200000,
                   subsample_freq=0, timeout=None)

make: *** [Makefile:25: doctest] Error 2
```

optuna/integration/_lightgbm_tuner/sklearn.py
```diff
@@ -1,49 +1,854 @@
...
+
+        >>> import optuna.integration.lightgbm as lgb
+        >>> from sklearn.datasets import load_boston
+        >>> from sklearn.model_selection import train_test_split
+        >>> reg = lgb.LGBMRegressor(random_state=0)
+        >>> X, y = load_boston(return_X_y=True)
+        >>> X, X_valid, y, y_valid = train_test_split(X, y, random_state=0)
+        >>> reg.fit(X, y, eval_set=[(X_valid, y_valid)])
+        LGBMRegressor(...)
+        >>> y_pred = reg.predict(X)
```
toshihikoyanase, line 853: ditto.
```suggestion
    Examples:

        .. doctest::

            >>> import optuna.integration.lightgbm as lgb
            >>> from sklearn.datasets import load_boston
            >>> from sklearn.model_selection import train_test_split
            >>> reg = lgb.LGBMRegressor(random_state=0)
            >>> X, y = load_boston(return_X_y=True)
            >>> X, X_valid, y, y_valid = train_test_split(X, y, random_state=0)
            >>> reg.fit(X, y, eval_set=[(X_valid, y_valid)], verbose=False)
            LGBMRegressor(...)
            >>> y_pred = reg.predict(X)
```

toshihikoyanase (changes requested):
I investigated the CI error reported in https://github.com/optuna/optuna/pull/1018#discussion_r566755553.

I think this is due to the warning message produced by `lightgbm`. 
During the investigation, I found some syntax errors in the docstrings. 
Could you take a look?

optuna/integration/_lightgbm_tuner/sklearn.py
```diff
@@ -1,49 +1,854 @@
...
+                    X_valid,
+                    label=y_valid,
+                    weight=valid_weight,
+                    group=valid_group,
+                    init_score=valid_init_score,
+                    feature_name=feature_name,
+                    categorical_feature=categorical_feature,
+                )
+
+            valid_sets.append(valid_set)
```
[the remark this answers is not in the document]
Y-oHr-N: Fixed in e4f6240.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: Fixed in 9351d0f.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: Fixed in 9351d0f.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: Fixed in 9351d0f.

optuna/integration/_lightgbm_tuner/sklearn.py
Y-oHr-N: Fixed in 9351d0f.

> In addition, this example causes an error due to the unexpected warning message (i.e., `[LightGBM] [Warning] min_data_in_leaf is set=20, min_child_samples=20 will be ignored. Current value: min_data_in_leaf=20` ). Do you have any idea to suppress it?

https://github.com/optuna/optuna/blob/master/optuna/integration/_lightgbm_tuner/alias.py#L9-L12
`min_data_in_leaf`, `min_data_per_leaf`, `min_data`, and `min_child_samples` will be renamed to `min_data_in_leaf` by `_handling_alias_parameters`.

https://github.com/optuna/optuna/blob/master/optuna/integration/_lightgbm_tuner/optimize.py#L594-L599
On the other hand, `LightGBMTuner` will try to tune `min_child_samples`.

In this case, `min_data_in_leaf` conflicts with `min_child_samples`,  and `min_chile_samples`, which should be tuned, will be ignored. I think we can solve this problem by renaming all `min_child_samples` that appear in `optimize.py` to `min_data_in_leaf`.

optuna/integration/_lightgbm_tuner/sklearn.py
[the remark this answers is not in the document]
toshihikoyanase: Thanks!

optuna/integration/_lightgbm_tuner/sklearn.py
toshihikoyanase: Thanks!

optuna/integration/_lightgbm_tuner/sklearn.py
toshihikoyanase: Thanks!

optuna/integration/_lightgbm_tuner/sklearn.py
toshihikoyanase: Thanks!

PR #1019, opened 2020-03-11
Fix docstring on optuna/integration/tensorflow.py

nuka137:
This PR fixes docstring on optuna/integration/tensorflow.py.

From https://github.com/optuna/optuna/issues/734#issuecomment-597459320 , I removed original docstring and put a reference to examples/pruning/tensorflow_estimator_integration.py.
I think examples/pruning/tensorflow_estimator_integration.py is simple enough to understand a usage.

Related Issue: #734

### DIFF
--- a/optuna/integration/tensorflow.py
+++ b/optuna/integration/tensorflow.py
@@ -21,8 +21,9 @@ class TensorFlowPruningHook(SessionRunHook):
 
     Example:
 
-        See example on GitHub for adding a pruning SessionRunHook for a TensorFlow's Estimator.
-        https://github.com/optuna/optuna/blob/master/examples/tensorflow_estimator_simple.py
+        See `the example <https://github.com/optuna/optuna/blob/30dba8814d332ff4
+        0d9c7465ee660a8ecf499547/examples/tensorflow_estimator_simple.py>`_
+        if you want to add a pruning SessionRunHook for TensorFlow's Estimator.
 
     Args:
         trial:
--- a/optuna/integration/tensorflow.py
+++ b/optuna/integration/tensorflow.py
@@ -21,23 +21,9 @@ class TensorFlowPruningHook(SessionRunHook):
 
     Example:
 
-        Add a pruning SessionRunHook for a TensorFlow's Estimator.
+        See example on GitHub for adding a pruning SessionRunHook for a TensorFlow's Estimator.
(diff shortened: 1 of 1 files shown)


optuna/integration/tensorflow.py
```diff
@@ -20,23 +20,9 @@ class TensorFlowPruningHook(SessionRunHook):
...
-                    run_every_steps=10,
-                )
-                hooks = [pruning_hook]
-                tf.estimator.train_and_evaluate(
-                    clf,
-                    tf.estimator.TrainSpec(input_fn=train_input_fn, max_steps=500, hooks=hooks),
-                    eval_spec
-                )
+        See example on GitHub for adding a pruning SessionRunHook for a TensorFlow's Estimator.
+        https://github.com/optuna/optuna/blob/master/examples/tensorflow_estimator_simple.py
```
toshihikoyanase: How about using the link instead of showing the URL?

image

Integration modules for XGBoost and Scikit-Optimize employ the links, so we can follow the convention.

image

image


And also, I think we can simplify the sentence because the words like TensorFlow and SessionRunHook are shown in the previous sentence.

> See [an example on GitHub](https://github.com/optuna/optuna/blob/master/examples/tensorflow_estimator_simple.py) to check the usage of the pruning hook.

himkt: I think it would be good to include the commit hash in the link since
a path to an example may change in the future.
(e.g. `https://github.com/optuna/optuna/blob/30dba8814d332ff40d9c7465ee660a8ecf499547/examples/tensorflow_estimator_simple.py`)

For me, it also looks good not to simplify the words. 👍 
If we hold these words, it's reasonable for me to avoid repeatedly use `for`.
Let me put an example as well:

> See [the example](https://github.com/optuna/optuna/blob/30dba8814d332ff40d9c7465ee660a8ecf499547/examples/tensorflow_estimator_simple.py) if you want to add a pruning SessionRunHook for TensorFlow's Estimator.

nuka137:
@toshihikoyanase @himkt 

Thanks!
I fixed the issues you mentioned.

Fixed code will generate below HTML.

optuna_doc

nuka137:
@toshihikoyanase 

Rebased now.
I think the CI testing seems to be passed now.

toshihikoyanase:
@himkt Thanks! Merged.

Issue #510, opened 2019-09-07
Update Suggest API


cthoyt:
There is already a function in the `Trial` class for suggesting a discrete uniformly distributed float ([`Trial.suggest_discrete_uniform`](https://optuna.readthedocs.io/en/latest/reference/trial.html#optuna.trial.Trial.suggest_discrete_uniform)). It would be nice to have the same option that returns integers instead of floats.

My use case is for network representation learning - I want to suggest the embedding dimension in intervals of 50 between 50-350 (e.g. 50, 100, 150, ... , 300, 350).\

Currently, I can accomplish this by dividing out my interval then suggesting an integer and multiplying it like:

```python
embeddings_dim = 50 * trial.suggest_int('embeddings_dim', 1, 7)
```

But this loses the power of tracking the value within `optuna`. Alternatively I could use the following, but the the data type is tracked incorrectly.

```python
embeddings_dim = int(trial.suggest_discrete_uniform('embeddings_dim', 50, 350, q=7))
```

sile:
Sorry for the late reply.
we think your proposal is reasonable.
However, instead of adding `suggest_discrete_int` method, we're planning to reorganize Suggest API as follows:
- (a) **[NEW_METHOD]** `Trial.suggest_float(name: str, low: float, high: float, log: bool, q: Optional[float])` 
   - `sugegst_float(name, low, high, log=False, q=None)` is equivalent to `suggest_uniform(name, low, high)`
   - `suggest_float(name, low, high, log=True, q=None)` is equivalent to `suggest_loguniform(name, low, high)`
   - `suggest_float(name, low, high, log=False, q=0.1)` is equivalent to `suggest_discrete_uniform(name, low, high, q=0.1)`
- (b) **[MODIFIED]** `Trial.suggest_int(name: str, low: int, high: int, q: Optional[int])` (
   - `suggest_int(name, low, high, q=10)` is equivalent to `Trial.suggest_discret_int(name, low, high, q=10)`
- (c) **[NO_CHANGE]** `Trial.suggest_categorical(name: str, choices: List[Union[str, float]])` 
- Other existing suggest methods will be deprecated.

Note that the details of the API design is still under consideration (e.g., `q` may be renamed to a more descriptive name).

If you have any thought about the above new API design (draft), please let us know.

cthoyt:
Looks good to me! Maybe rather than giving log as a Boolean, you might use a keyword argument called “distribution” in case you’d ever like to use something more exotic

[...]

PR #1021, opened 2020-03-12
Add initial implementation of `suggest_float`.

himkt:
related #510 

This PR adds a new method `suggest_float`, which is a wrapper of `suggest_uniform` and `suggest_loguniform`.

### DIFF
--- a/optuna/trial.py
+++ b/optuna/trial.py
@@ -165,7 +165,7 @@ class Trial(BaseTrial):
         # type: (str, float, float, bool) -> float
         """Wrapper method for ``suggest_uniform`` and ``suggest_loguniform``.
 
-        .. versionadded:: 1.2.0
+        .. versionadded:: 1.3.0
 
         .. seealso::
             Please see also :func:`~optuna.trial.Trial.suggest_uniform` and
--- a/optuna/trial.py
+++ b/optuna/trial.py
@@ -160,7 +160,10 @@ class Trial(BaseTrial):
 
     def suggest_float(self, name, low, high, *, log=False):
         # type: (str, float, float, bool) -> float
-        """Wrapper method for ``suggest_uniform`` and ``suggest_loguniform``.
+        """Suggest a value for the floating point parameter.
+
+        Note that this is a wrapper method for :func:`~optuna.trial.Trial.suggest_uniform`
+        and :func:`~optuna.trial.Trial.suggest_loguniform`.
 
         .. versionadded:: 1.3.0
 
--- a/optuna/trial.py
+++ b/optuna/trial.py
@@ -30,7 +30,8 @@ class BaseTrial(object, metaclass=abc.ABCMeta):
     Note that this class is not supposed to be directly accessed by library users.
     """
 
-    def suggest_float(self, name, low, high, log):
+    @experimental("1.2.0")
+    def suggest_float(self, name, low, high, *, log=False):
         # type: (str, float, float, bool) -> float
 
         raise NotImplementedError
@@ -160,7 +161,7 @@ class Trial(BaseTrial):
         )
 
     @experimental("1.2.0")
-    def suggest_float(self, name, low, high, log=False):
+    def suggest_float(self, name, low, high, *, log=False):
         # type: (str, float, float, bool) -> float
         """Wrapper method for ``suggest_uniform`` and ``suggest_loguniform``.
 
@@ -943,7 +944,8 @@ class FixedTrial(BaseTrial):
         self._system_attrs = {}  # type: Dict[str, Any]
         self._datetime_start = datetime.now()
 
-    def suggest_float(self, name, low, high, log=False):
+    @experimental("1.2.0")
+    def suggest_float(self, name, low, high, *, log=False):
         # type: (str, float, float, bool) -> float
 
         if log:
--- a/optuna/trial.py
+++ b/optuna/trial.py
@@ -208,6 +208,11 @@ class Trial(BaseTrial):
             high:
                 Upper endpoint of the range of suggested values. ``high`` is excluded from the
                 range.
+            log:
+                A flag to sample the value from the log domain or not.
+                If ``log`` is true, the value is sampled from the range in the log domain.
+                Otherwise, the value is sampled from the range in the linear domain.
+                See also :func:`suggest_uniform` and :func:`suggest_loguniform`.
 
         Returns:
             A suggested float value.
--- a/optuna/trial.py
+++ b/optuna/trial.py
@@ -12,10 +12,16 @@ if type_checking.TYPE_CHECKING:
     from typing import Dict  # NOQA
     from typing import Optional  # NOQA
     from typing import Sequence  # NOQA
+    from typing import Union  # NOQA
 
     from optuna.distributions import BaseDistribution  # NOQA
     from optuna.distributions import CategoricalChoiceType  # NOQA
     from optuna.study import Study  # NOQA
+
+    FloatingPointDistributionType = Union[
+        distributions.UniformDistribution,
+        distributions.LogUniformDistribution
+    ]
 
 
 class BaseTrial(object, metaclass=abc.ABCMeta):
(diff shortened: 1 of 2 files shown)


tests/test_trial.py
```diff
@@ -25,6 +25,25 @@
     [storages.InMemoryStorage, lambda: storages.RDBStorage('sqlite:///:memory:')])
 
 
+@parametrize_storage
+def test_check_distribution_suggest_float(storage_init_func):
```
himkt: I added the test to check if:

- `suggest_float(...)` and `suggest_uniform(...)` are the same
- `suggest_float(..., log=True)` and `suggest_loguniform(...)` are the same

himkt:
Now I check if `suggest_float` and `suggest_uniform` (`suggest_loguniform`) are the same by following:

```
x1 = trial.suggest_float('x1', 10, 20)
x2 = trial.suggest_uniform('x1', 10, 20)  # looks like the result of suggest_float is cached

assert x1 == x2
```

I'm not sure this is good for testing this PR and I would really appreciate any feedback.

optuna/trial.py
```diff
@@ -146,6 +157,72 @@ def _init_relative_params(self):
         self.relative_params = self.study.sampler.sample_relative(self.study, trial,
                                                                   self.relative_search_space)
 
+    def suggest_float(self, name, low, high, log=False):
```
toshihikoyanase: Please add a description for `log` in docstring.

tests/test_trial.py
toshihikoyanase: I think checking the compatibility with the existing interface is enough for the first implementation. 

And also, I think we need to check the case `low==high`.

optuna/trial.py
```diff
@@ -146,6 +157,72 @@ def _init_relative_params(self):
...
+                    learning_rate_init = trial.suggest_float('learning_rate_init',
+                                                             1e-5, 1e-3, log=True)
+                    clf = MLPClassifier(hidden_layer_sizes=(100, 50), momentum=momentum,
+                                        learning_rate_init=learning_rate_init,
+                                        solver='sgd', random_state=0)
+                    clf.fit(X_train, y_train)
+
+                    return clf.score(X_test, y_test)
+
+                study = optuna.create_study()
```
toshihikoyanase: Maybe this objective function should be maximized. Please set `direction='maximize'`.

optuna/trial.py
himkt: Added in df6a469.

tests/test_trial.py
himkt: 34d60b1
I added the test for checking the case `low==high`.

optuna/trial.py
himkt: :pray: 4a4e8d2
I'll send another PR to fix the same mistakes in `suggest_uniform` and `suggest_loguniform`.

himkt:
@toshihikoyanase I revised my PR, please take a look.

optuna/trial.py
```diff
@@ -148,6 +159,82 @@ def _init_relative_params(self):
             self.study, trial, self.relative_search_space
         )
 
+    @experimental("1.2.0")
+    def suggest_float(self, name, low, high, log=False):
+        # type: (str, float, float, bool) -> float
+        """Wrapper method for ``suggest_uniform`` and ``suggest_loguniform``.
+
+        .. versionadded:: 1.2.0
```
sile: The next release version will be `v1.3.0`.
```suggestion
        .. versionadded:: 1.3.0
```

optuna/trial.py
```diff
@@ -12,18 +13,28 @@
...
+    ]
+
 
 class BaseTrial(object, metaclass=abc.ABCMeta):
     """Base class for trials.
 
     Note that this class is not supposed to be directly accessed by library users.
     """
 
+    def suggest_float(self, name, low, high, log):
```
sile: How about specifying the default value of `log` argument?
```suggestion
    def suggest_float(self, name, low, high, log=False):
```

optuna/trial.py
```diff
@@ -148,6 +160,82 @@ def _init_relative_params(self):
             self.study, trial, self.relative_search_space
         )
 
+    @experimental("1.2.0")
+    def suggest_float(self, name, low, high, *, log=False):
+        # type: (str, float, float, bool) -> float
+        """Wrapper method for ``suggest_uniform`` and ``suggest_loguniform``.
```
sile: Personally, at the first line, I prefer writing a sentence that briefly describes the behavior of the target method (see below as an example): 
```
Suggest a value for the floating point parameter.

Note that this is a wrapper method for :func:`~optuna.trial.Trial.suggest_uniform` and  :func:`~optuna.trial.Trial.suggest_loguniform`.
```
However, this is not a strong opinion and the current doc seems good as well.

optuna/trial.py
```diff
@@ -148,6 +160,82 @@ def _init_relative_params(self):
...
+            )  # type: FloatingPointDistributionType
+        else:
+            distribution = distributions.UniformDistribution(low=low, high=high)
+
+        self._check_distribution(name, distribution)
+
+        if low == high:
+            return self._set_new_param_or_get_existing(name, low, distribution)
+
+        return self._suggest(name, distribution)
```
sile: I might have understood incorrectly, but it seems that this code block can be simplified as follows:
```suggestion
        if log:
            return self.suggest_loguniform(low, high)
        else:
            return self.suggest_uniform(low, high)
```

optuna/trial.py
himkt: Thank you for the suggestion and I agree with your idea.
Let me apply your feedback. (6ff1d40)

optuna/trial.py
himkt: You're right, thank you.

optuna/trial.py
himkt: The default values were introduced in 988bdf9.

himkt:
@sile 
Thank you for the reviewing and I'm sorry for the late reply.
I update my PR and apply some feedbacks.
Could you please take a look?

optuna/trial.py
```diff
@@ -148,6 +158,84 @@ def _init_relative_params(self):
...
+                from sklearn.model_selection import train_test_split
+                from sklearn.neural_network import MLPClassifier
+
+                np.random.seed(seed=0)
+                X = np.random.randn(200).reshape(-1, 1)
+                y = np.random.randint(0, 2, 200)
+                X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=0)
+
+
+            ..testcode::
```
sile: Maybe a typo:
```suggestion
            .. testcode::
```

optuna/trial.py
```diff
@@ -148,6 +158,84 @@ def _init_relative_params(self):
...
+                A flag to sample the value from the log domain or not.
+                If ``log`` is true, the value is sampled from the range in the log domain.
+                Otherwise, the value is sampled from the range in the linear domain.
+                See also :func:`suggest_uniform` and :func:`suggest_loguniform`.
+
+        Returns:
+            A suggested float value.
+        """
+
+        if log:
```
sile: The comment https://github.com/optuna/optuna/pull/1021/files#r396901171 was marked as resolved but it seems the suggestion hasn't been applied.

optuna/trial.py
himkt: Thank you for catching!

optuna/trial.py
himkt: It seems I have resolved this unintentionally, sorry.
Applied in 4669c1b.

sile: Thanks!

crcrpar:
> So, please add `@experimental` annotation to `suggest_float`.

Do we really have `suggest_float` decorated with `@experimental`?

toshihikoyanase:
@crcrpar Thank you for pointing it out.
The current `suggest_float` employes keyword-only arguments (https://www.python.org/dev/peps/pep-3102/), and users have a little/no troubles even if we add new arguments. So, I changed my mind and declined my suggestion about `@experimental`.
I should have commented. I'm sorry for bothering you.

crcrpar:
Gotcha. Thanks for the explanation. That sounds modern and great way. 🙂

PR #1022, opened 2020-03-12
Fix a docstring of RDBStorage

c-bata:

### DIFF
--- a/optuna/storages/rdb/storage.py
+++ b/optuna/storages/rdb/storage.py
@@ -71,6 +71,8 @@ class RDBStorage(BaseStorage):
         engine_kwargs:
             A dictionary of keyword arguments that is passed to
             `sqlalchemy.engine.create_engine`_ function.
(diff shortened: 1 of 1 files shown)


optuna/storages/rdb/storage.py
```diff
@@ -71,6 +71,8 @@ class RDBStorage(BaseStorage):
         engine_kwargs:
             A dictionary of keyword arguments that is passed to
             `sqlalchemy.engine.create_engine`_ function.
+        skip_compatibility_check:
+            Skip to check the schema compatibility if True.
```
hvy: A nit on the grammar.
```suggestion
            Flag to skip schema compatibility check if set to True.
```

c-bata: Thank you!

hvy:
Let me just update the title of this PR to match the format of our release notes.

PR #1025, opened 2020-03-13
Add a method to stop `study.optimize`.

g-votte:
While the exception to stop the running optimization, `StopStudy`, is proposed in #966 and #986, this PR proposes a different approach: the `study.stop` method safely stops the optimization after the current trial finishes.

### Motivation 1: safely stopping the optimization
While `StopStudy` immediately stops the optimization, this method does it after the current trial finishes, which makes the expected behavior more trivial.

The expected `TrialState` might depends on the user's scenario. Some users want the state to be `COMPLETE` and others do `FAIL`. `Study.stop` just delegates this handling to the native logic of `study.optimize`, whereas we'd need careful design and implementation if the optimization is immediately stopped by the `StopStudy`.

This is also the case in the behavior around callbacks. If `StopStudy` is raised in a callback, how should we handle following callbacks: ignore or continue? It might depend on the user. On the other hand, we just need to run all callbacks with `study.stop`.

### Motivation 2: a method instead of an exception
How to trigger the API is more carefully controlled with a method than an exception. We can easily handle the user's calling `study.stop` outside the optimization. It's rather hard with `StopStudy`.

The downside might be that the API style is somewhat inconsistent with `TrialPruned`.


optuna/study.py
```diff
@@ -523,6 +529,17 @@ def _create_record_and_aggregate_column(trial):
 
         return df
 
+    @experimental("1.4.0")
+    def stop(self) -> None:
+        """Stop the optimization after the current trial finishes."""
```
ytsmiling, line 534: I think this method needs a more detailed description. For example,
```suggestion

        """Exit from the current optimization loop after the running trials finish.

        This method lets the running `optimize` method return immediately after
        all trials which the `optimize` method spawned finishes.
        This method does not affect any behaviors of parallel or successive
        study processes.
        
        Raises:
            RuntimeError:
                If this method is called outside an objective function or callback.
        """
```

toshihikoyanase: I agree. This method stops only a single study process if users run multiple processes. I think this behavior fits the behavior of  `study.optimize`, but some users may expect to stop all processes. So, this description is useful for such users.
I feel it may be more helpful for users if we show how to stop all processes (e.g., setting a flag to study.user_attr). It optional and we can separate PRs.

A net. `optimize` seems to be linked to the reference:
```
:meth:`~optuna.study.Study.optimize`
```

optuna/study.py
```diff
@@ -545,6 +545,7 @@ def stop(self) -> None:
         """
 
         if self._optimize_lock.acquire(False):
+            self._optimize_lock.release()
```
toshihikoyanase, line 548: :+1:

g-votte: I put `self._optimize_lock.release()` here; otherwise, [this test](https://github.com/optuna/optuna/pull/1025/commits/7651075652e0eb67a0b61efc0f3c770d7a71c878#diff-48005dfcd962e57ca1be2ec86e03144eR751-R753) fails, where the method raises an error even inside an objective function or callback.

@toshihikoyanase 
Thanks for your additional review!

g-votte:
There seems to be an extra indent after the second paragraph of the docs. (See also the figure in https://github.com/optuna/optuna/pull/1025#pullrequestreview-403131623)

I suspect that the cause is in the `experimental` decorator, and raised an issue about that. https://github.com/optuna/optuna/issues/1213

PR #1026, opened 2020-03-13
Support automatic stop of `GridSampler`.

g-votte:
This PR depends on #1025 and supports automatically stoping grid search after all grids are evaluated.

Note that there will be a breaking change where `GridSampler` currently raises an exception when a trial is triggered after evaluating all grids.


g-votte:
Although we still need to discuss how to solve #966 , I’ve created this PR to evaluate the concept of the `study.stop` method (#1025) .

g-votte:
Because #1025 has been merged, I think we can now go on with this PR.

optuna/samplers/grid.py
```diff
@@ -108,22 +112,29 @@ def sample_relative(self, study, trial, search_space):
...
-                "optimization by properly setting `n_trials` in `study.optimize`."
+        if len(target_grids) == 0:
+            _logger.warning(
+                "`GridSampler` is evaluating a duplicated point because all grids "
+                "have been evaluated. This may happen due to a timing issue during "
+                "distributed optimization or an unnecessary number of `n_trials`."
             )
 
+            study.stop()
+            target_grids = list(range(len(self._all_grids)))
```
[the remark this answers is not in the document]
g-votte: Thanks for your pointing it out. I updated the document. https://github.com/optuna/optuna/pull/1026/commits/e4182c179d712daa5520572b37b6a58cd7a59158

optuna/samplers/grid.py
```diff
@@ -48,10 +52,9 @@ def objective(trial):
 
     Note:
 
-        :class:`~optuna.samplers.GridSampler` raises an error if all combinations in the passed
-        ``search_space`` has already been evaluated. Please make sure that unnecessary trials do
-        not run during optimization by properly setting ``n_trials`` in the
-        :func:`~optuna.study.Study.optimize` method.
+        :class:`~optuna.samplers.GridSampler` automatically stops the optimization if all
+        combinations in the passed ``search_space`` has already been evaluated, internally
```
toshihikoyanase, line 56: Thank you for your update!

[nit] I guess `has` is a typo for `have`.
```suggestion
        combinations in the passed ``search_space`` have already been evaluated, internally
```

g-votte: Thanks! I merged the change.

optuna/samplers/grid.py
```diff
@@ -108,22 +111,29 @@ def sample_relative(self, study, trial, search_space):
...
         # In distributed optimization, multiple workers may simultaneously pick up the same grid.
         # To make the conflict less frequent, the grid is chosen randomly.
-        grid_id = random.choice(unvisited_grids)
+        grid_id = random.choice(target_grids)
 
         study._storage.set_trial_system_attr(trial._trial_id, "search_space", self._search_space)
         study._storage.set_trial_system_attr(trial._trial_id, "grid_id", grid_id)
 
+        # Once all grids are picked up, optimization stops after the current trial finishes.
+        if len(self._get_unvisited_grid_ids(study, include_unfinished=True)) == 0:
```
hvy, line 134: I'm a bit worried that it's not immediately clear why `_get_unvisited_grid_ids` is called twice in this method and with different arguments. At least it isn't to me. Would you mind leaving a comment about this logic?

g-votte: Thanks for your pointing this out. As you mention, the duplicated call of this method seems unnecessary. I just simplified logic.

optuna/samplers/_grid.py
```diff
@@ -108,18 +111,33 @@ def sample_relative(self, study, trial, search_space):
...
-                "optimization by properly setting `n_trials` in `study.optimize`."
+        if len(target_grids) == 0:
+            # This case may occur with distributed optimization or trial queue. If there is no
+            # target grid, `GridSampler` evaluates a visited, duplicated point with the current
+            # trial. After that, the optimization stops.
+
+            _logger.warning(
+                "`GridSampler` is evaluating a duplicated point because all grids  have been "
+                "evaluated. This may happen due to a timing issue during distributed optimization "
+                "or an unnecessary number of `n_trials`."
```
hvy, line 124: > ... an unnecessary number of `n_trials`

Isn't this what this PR addresses, or does this refer to the case of rerunning the optimization after a study has finished (in which case a single trial per worker will start)?

optuna/samplers/_grid.py
```diff
@@ -108,18 +111,33 @@ def sample_relative(self, study, trial, search_space):
...
-        unvisited_grids = self._get_unvisited_grid_ids(study)
+        target_grids = self._get_unvisited_grid_ids(study)
 
-        if len(unvisited_grids) == 0:
-            raise ValueError(
-                "All grids have been evaluated. If you want to avoid this error, "
-                "please make sure that unnecessary trials do not run during "
-                "optimization by properly setting `n_trials` in `study.optimize`."
+        if len(target_grids) == 0:
+            # This case may occur with distributed optimization or trial queue. If there is no
```
hvy, line 117: This is more of a note regarding the "trial queue", but this sampler might not work that well with it. That is, the `grid_id` system attribute is set on trials even when they are popped from the trial queue and will be interpreted by the sampler as if they consumed an entry from the grid. It might be a problem caused by the fact that `sample_relative` is called from queued trials. It's probably beyond the scope of this PR though.

tests/samplers_tests/test_grid.py
```diff
@@ -96,6 +92,25 @@ def sorted_values(d):
...
 
+def test_study_optimize_with_exceeding_number_of_trials():
+    # type: () -> None
+
+    def objective(trial):
+        # type: (Trial) -> float
+
+        return trial.suggest_int("a", 0, 100)
+
+    # When `n_trials` is `None`, the optimization stops just after all grids are evaluated.
```
hvy, line 103: 👍 , this should be the recommended way to use the grid sampler.

optuna/samplers/_grid.py
```diff
@@ -108,18 +111,33 @@ def sample_relative(self, study, trial, search_space):
...
+            # target grid, `GridSampler` evaluates a visited, duplicated point with the current
+            # trial. After that, the optimization stops.
+
+            _logger.warning(
+                "`GridSampler` is re-evaluating a configuration because the grid has been "
+                "exhausted. This may happen due to a timing issue during distributed optimization "
+                "or an unnecessary number of `n_trials`."
```
hvy, line 124: How about the following?
```suggestion
                "`GridSampler` is re-evaluating a configuration because the grid has been "
                "exhausted. This may happen due to a timing issue during distributed optimization "
                "or when re-running optimizations on already finished studies."
```

g-votte: Thanks! I merged the change.

hvy:
@g-votte sorry could you merge/rebase the master branch? I think we need the latest `setup.py` for the CI to pass.

g-votte:
@hvy 
Thanks for your reminding me of that..! I merged the master.

PR #1014, opened 2020-03-11
Use SELECT FOR UPDATE while updating trial state

c-bata:
## Summary

Fix a concurrency problem of `set_trial_state` by using `SELECT ... FOR UPDATE`.

## Check script

```sh
#!/bin/sh

set -ex

docker pull mysql:8.0

docker run \
  -d \
  --rm \
  -p 3306:3306 \
  --mount type=volume,src=mysql,dst=/etc/mysql/conf.d \
  -e MYSQL_USER=optuna \
  -e MYSQL_DATABASE=optuna \
  -e MYSQL_PASSWORD=password \
  -e MYSQL_ALLOW_EMPTY_PASSWORD=yes \
  --name optuna-mysql \
  mysql:8.0

sleep 20

python update_trial_state.py

docker stop optuna-mysql
```

```python
import optuna


def objective(trial):
    x1 = trial.suggest_uniform("x1", -10, 10)
    print("x1", x1)
    return (x1 - 5) ** 2


def main():
    optuna.logging.set_verbosity(optuna.logging.DEBUG)
    storage = optuna.storages.RDBStorage("mysql+mysqldb://optuna:password@127.0.0.1:3306/optuna")
    study = optuna.create_study(storage=storage)
    study.optimize(objective, n_trials=1)

    for i in range(10):
        study.enqueue_trial({"x1": i})

    study.optimize(objective, n_jobs=10, n_trials=10)


if __name__ == '__main__':
    main()
```
 
## Before

```
$ python update_trial_state.py 
[D 2020-03-12 01:55:32,027] pool_pre_ping=True was set to engine_kwargs to prevent connection timeout.
[I 2020-03-12 01:55:32,260] A new study created with name: no-name-eb48caa4-24a2-4310-8022-1c7240dbc8f9
x1 -7.25102819034571
[I 2020-03-12 01:55:32,562] Finished trial#0 resulted in value: 150.0876917206453. Current best value is 150.088 with parameters: {'x1': -7.25103}.
/Users/a14737/src/github.com/pfnet/optuna/optuna/_experimental.py:83: ExperimentalWarning: enqueue_trial is experimental (supported from v1.2.0). The interface can change in the future.
  warnings.warn(
[D 2020-03-12 01:55:33,299] Trial#1 is popped from the trial queue.
[D 2020-03-12 01:55:33,306] Trial#1 is popped from the trial queue.
[D 2020-03-12 01:55:33,361] Trial#2 is popped from the trial queue.
[...]

PR #1027, opened 2020-03-13
Add a todo comment in `GridSampler`.

c-bata:
GridSampler may evaluate some duplicated parameters like:
https://gist.github.com/c-bata/f759f64becb24eea2040f4b2e3afce8f

I think that the approach of #1014 might be the solution, but it's still not a perfect solution and a lot of API changes are required for this purpose. So I put a todo comment for now.


g-votte:
As @c-bata mentions, we still want to come up with solution to avoid duplicated evaluation. Thanks for clarifying the issue!

Issue #1017, opened 2020-03-11
MLflow integration


PhilipMay:
I am thinking of an [MLflow](https://mlflow.org) integration. What do you think about this? I would contribute but need someone who discusses with me on where to plug it in.

I saw the MLflow example here: https://github.com/optuna/optuna/tree/master/examples/mlflow

It should be able to automatically log everything that is written to the optuna DB also to MLflow.

What do you think?

g-votte:
@PhilipMay 
I love this idea. It should make Optuna more visualizable especially when the user is familiar with MLFlow.

A possible design might be to provide a callback function/class under `optuna.integration.mlflow` like what is shown in [the example](https://github.com/optuna/optuna/tree/master/examples/mlflow), while I think there're some other ways to integrate the library.

What kind of implementation are you thinking of?

PhilipMay:
I think I will build a callback implementation first.

PhilipMay:
> What kind of implementation are you thinking of?

Well - see the PR here: #1028 Feedback welcome. :-)

PhilipMay:
PR is merged. Closing this.

PR #1028, opened 2020-03-14
Add Mlflow integration callback. 

PhilipMay:
This is the PR for a [Mlflow](https://mlflow.org/docs/latest/index.html) integration callback. Also see the discussion here: #1017

## Todo
- [x] type annotations
- [x] extend `setup.py` 
- [x] tests
- [x] docstring

### DIFF
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -82,7 +82,7 @@ class MLflowCallback(object):
               Databricks CLI.
               `profile <https://github.com/databricks/databricks-cli#installation>`_,
               ``databricks://<profileName>``.
-        experiment:
+        experiment_name:
             Name of MLflow experiment to be activated. If not set ``study.study_name``
             will be taken. Either ``experiment`` or ``study.study_name`` must be set.
         metric_name:
@@ -95,7 +95,7 @@ class MLflowCallback(object):
         _check_mlflow_availability()
 
         self._tracking_uri = tracking_uri
-        self._experiment = experiment
+        self._experiment_name = experiment_name
         self._metric_name = metric_name
 
     def __call__(self, study, trial):
@@ -106,8 +106,8 @@ class MLflowCallback(object):
             mlflow.set_tracking_uri(self._tracking_uri)
 
         # This sets the experiment of MLflow.
-        if self._experiment is not None:
-            mlflow.set_experiment(self._experiment)
+        if self._experiment_name is not None:
+            mlflow.set_experiment(self._experiment_name)
         elif (
             study.study_name is not None
             and study.study_name != "no-name-00000000-0000-0000-0000-000000000000"
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -65,6 +65,12 @@ class MlflowCallback(object):
             import shutil
             shutil.rmtree(tempdir)
 
+        .. testoutput::
+            :hide:
+            :options: +NORMALIZE_WHITESPACE
+
+            INFO: 'my_study' does not exist. Creating a new experiment
+
     Args:
         tracking_uri:
             Set the tracking server URI.
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -59,23 +59,23 @@ class MlflowCallback(object):
     def __call__(self, study, trial):
         # type: (optuna.study.Study, optuna.structs.FrozenTrial) -> None
 
-        # set experiment
-        if self._experiment is None:
-            if (
-                study.study_name is None
-                or study.study_name == "no-name-00000000-0000-0000-0000-000000000000"
-            ):
-                raise ValueError("Either 'experiment' or 'study.study_name' must be set!")
-            else:
-                mlflow.set_experiment(study.study_name)
-        else:
-            mlflow.set_experiment(self._experiment)
-
         # set tracking_uri
         if self._tracking_uri is not None:
             mlflow.set_tracking_uri(self._tracking_uri)
 
         with mlflow.start_run(run_name=trial.number):
+            # set experiment
+            if self._experiment is None:
+                if (
+                    study.study_name is None
+                    or study.study_name == "no-name-00000000-0000-0000-0000-000000000000"
+                ):
+                    raise ValueError("Either 'experiment' or 'study.study_name' must be set!")
+                else:
+                    mlflow.set_experiment(study.study_name)
+            else:
+                mlflow.set_experiment(self._experiment)
+
             # set metric
             trial_value = trial.value if trial.value is not None else float("nan")
             metric_name = self._metric_name if self._metric_name is not None else "trial_value"
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -36,19 +36,29 @@ class MlflowCallback(object):
 
         Add Mlflow callback to optuna optimization.
 
-        .. code::
+        .. testsetup::
+            import pathlib
+            import tempfile
+            tempdir = tempfile.mkdtemp()
+            YOUR_TRACKING_URI = pathlib.Path(tempdir).as_uri()
+        .. testcode::
+            import optuna
+            from optuna.integration.mlflow import MlflowCallback
 
             def objective(trial):
                 x = trial.suggest_uniform('x', -10, 10)
                 return (x - 2) ** 2
 
             mlflc = MlflowCallback(
-                tracking_uri='https://my-tracking-server:5000',
+                tracking_uri=YOUR_TRACKING_URI,
                 metric_name='my metric score',
-                )
+            )
 
             study = optuna.create_study(study_name='my_study')
             study.optimize(objective, n_trials=10, callbacks=[mlflc])
+        .. testcleanup::
+            import shutil
+            shutil.rmtree(tempdir)
 
     Args:
         tracking_uri:
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -10,7 +10,7 @@ if type_checking.TYPE_CHECKING:
 class MlflowCallback(object):
     """Callback to track optuna trials with Mlflow.
 
-    This callback adds all relevant information that is
+    This callback adds relevant information that is
     tracked by Optuna to Mlflow.
 
     Example:
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -17,7 +17,7 @@ class MlflowCallback(object):
 
         Add Mlflow callback to optuna optimization.
 
-        .. testcode::
+        .. code::
 
             def objective(trial):
                 x = trial.suggest_uniform('x', -10, 10)
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -1,4 +1,5 @@
 import optuna
+from optuna import structs
 from optuna import type_checking
 from optuna._experimental import experimental
 
@@ -129,7 +130,13 @@ class MLflowCallback(object):
             tags["trial_number"] = str(trial.number)
             tags["trial_datetime_start"] = str(trial.datetime_start)
             tags["trial_datetime_complete"] = str(trial.datetime_complete)
-            tags["trial_state"] = str(trial.state)
+
+            # Set TrialState and convert it to str and remove the common prefix.
+            trial_state = trial.state
+            if isinstance(trial_state, structs.TrialState):
+                trial_state = str(trial_state).split(".")[-1]
+                tags["trial_state"] = str(trial_state)
+
             tags["study_direction"] = str(study.direction)
             tags.update(trial.user_attrs)
             distributions = {
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -17,7 +17,7 @@ class MlflowCallback(object):
 
         Add Mlflow callback to optuna optimization.
 
-        .. code::
+        .. testcode::
 
             def objective(trial):
                 x = trial.suggest_uniform('x', -10, 10)
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -88,7 +88,7 @@ class MLflowCallback(object):
     """
 
     def __init__(self, tracking_uri=None, metric_name="value"):
-        # type: (Optional[str], Optional[str]) -> None
+        # type: (Optional[str], str) -> None
 
         _check_mlflow_availability()
 
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -116,7 +116,7 @@ class MLflowCallback(object):
         else:
             raise ValueError("Either 'experiment' or 'study.study_name' must be set!")
 
-        with mlflow.start_run(run_name=trial.number):
+        with mlflow.start_run(run_name=str(trial.number)):
 
             # This sets the metric for MLflow.
             trial_value = trial.value if trial.value is not None else float("nan")
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -69,8 +69,8 @@ class MlflowCallback(object):
         tracking_uri:
             Set the tracking server URI.
 
-            - An empty string, or a local file path, prefixed with ``file:/``. Data is stored
-              locally at the provided file (or ``./mlruns`` if empty).
+            - An empty string, or a local directory path, prefixed with ``file:``. Data is stored
+              locally in the provided directory (or ``./mlruns`` if empty).
             - An HTTP URI like ``https://my-tracking-server:5000``.
             - A Databricks workspace, provided as the string ``databricks`` or, to use a
               Databricks CLI.
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -120,12 +120,16 @@ class MLflowCallback(object):
             tags["datetime_start"] = str(trial.datetime_start)
             tags["datetime_complete"] = str(trial.datetime_complete)
 
-            # Set TrialState and convert it to str and remove the common prefix.
+            # Set state and convert it to str and remove the common prefix.
             trial_state = trial.state
             if isinstance(trial_state, structs.TrialState):
                 tags["state"] = str(trial_state).split(".")[-1]
 
-            tags["direction"] = str(study.direction)
+            # Set direction and convert it to str and remove the common prefix.
+            study_direction = study.direction
+            if isinstance(study_direction, structs.StudyDirection):
+                tags["direction"] = str(study_direction).split(".")[-1]
+
             tags.update(trial.user_attrs)
             distributions = {
                 (k + "_distribution"): str(v) for (k, v) in trial.distributions.items()
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -89,7 +89,7 @@ class MLflowCallback(object):
     """
 
     def __init__(self, tracking_uri=None, experiment_name=None, metric_name=None):
-        # type: (str, str, str) -> None
+        # type: (Optional[str], Optional[str], Optional[str]) -> None
 
         _check_mlflow_availability()
 
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -29,7 +29,7 @@ def _check_mlflow_availability():
         )
 
 
-@experimental("1.3.0")
+@experimental("1.4.0")
 class MLflowCallback(object):
     """Callback to track optuna trials with MLflow.
 
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -1,5 +1,6 @@
 import optuna
 from optuna import type_checking
+from optuna._experimental import experimental
 
 if type_checking.TYPE_CHECKING:
     from typing import Dict  # NOQA
@@ -26,6 +27,7 @@ def _check_mlflow_availability():
         )
 
 
+@experimental("1.3.0")
 class MLflowCallback(object):
     """Callback to track optuna trials with MLflow.
 
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -110,7 +110,8 @@ class MLflowCallback(object):
             mlflow.set_experiment(study.study_name)
         else:
             raise ValueError(
-                "You must set 'study.study_name' inside Optuna! This is needed to set the MLflow experiment name."
+                "You must set 'study.study_name' inside Optuna! This "
+                "is needed to set the MLflow experiment name."
             )
 
         with mlflow.start_run(run_name=str(trial.number)):
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -84,7 +84,9 @@ class MLflowCallback(object):
             <https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.set_tracking_uri>`_
             for more details.
         metric_name:
-            Name of the metric.
+            Name of the metric. Since the metric itself is just a number,
+            `metric_name` can be used to give it a name. So you know later
+            if it was roc-auc or accuracy.
     """
 
     def __init__(self, tracking_uri=None, metric_name="value"):
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -31,7 +31,7 @@ def _check_mlflow_availability():
 
 @experimental("1.4.0")
 class MLflowCallback(object):
-    """Callback to track optuna trials with MLflow.
+    """Callback to track Optuna trials with MLflow.
 
     This callback adds relevant information that is
     tracked by Optuna to MLflow. The MLflow experiment
@@ -39,7 +39,7 @@ class MLflowCallback(object):
 
     Example:
 
-        Add MLflow callback to optuna optimization.
+        Add MLflow callback to Optuna optimization.
 
         .. testsetup::
 
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -37,11 +37,14 @@ class MlflowCallback(object):
         Add Mlflow callback to optuna optimization.
 
         .. testsetup::
+
             import pathlib
             import tempfile
             tempdir = tempfile.mkdtemp()
             YOUR_TRACKING_URI = pathlib.Path(tempdir).as_uri()
+
         .. testcode::
+
             import optuna
             from optuna.integration.mlflow import MlflowCallback
 
@@ -56,7 +59,9 @@ class MlflowCallback(object):
 
             study = optuna.create_study(study_name='my_study')
             study.optimize(objective, n_trials=10, callbacks=[mlflc])
+
         .. testcleanup::
+
             import shutil
             shutil.rmtree(tempdir)
 
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -1,10 +1,11 @@
 import optuna
+from optuna._experimental import experimental
 from optuna import structs
 from optuna import type_checking
-from optuna._experimental import experimental
 
 if type_checking.TYPE_CHECKING:
     from typing import Dict  # NOQA
+from typing import Optional
 
 try:
     import mlflow
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -119,7 +119,7 @@ class MLflowCallback(object):
 
             # This sets the metric for MLflow.
             trial_value = trial.value if trial.value is not None else float("nan")
-            metric_name = self._metric_name if self._metric_name is not None else "trial_value"
+            metric_name = self._metric_name if self._metric_name is not None else "value"
             mlflow.log_metric(metric_name, trial_value)
 
             # This sets the params for MLflow.
@@ -127,17 +127,17 @@ class MLflowCallback(object):
 
             # This sets the tags for MLflow.
             tags = {}  # type: Dict[str, str]
-            tags["trial_number"] = str(trial.number)
-            tags["trial_datetime_start"] = str(trial.datetime_start)
-            tags["trial_datetime_complete"] = str(trial.datetime_complete)
+            tags["number"] = str(trial.number)
+            tags["datetime_start"] = str(trial.datetime_start)
+            tags["datetime_complete"] = str(trial.datetime_complete)
 
             # Set TrialState and convert it to str and remove the common prefix.
             trial_state = trial.state
             if isinstance(trial_state, structs.TrialState):
                 trial_state = str(trial_state).split(".")[-1]
-                tags["trial_state"] = str(trial_state)
+                tags["state"] = str(trial_state)
 
-            tags["study_direction"] = str(study.direction)
+            tags["direction"] = str(study.direction)
             tags.update(trial.user_attrs)
             distributions = {
                 (k + "_distribution"): str(v) for (k, v) in trial.distributions.items()
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -66,7 +66,10 @@ class MlflowCallback(object):
         # This sets the experiment of Mlflow.
         if self._experiment is not None:
             mlflow.set_experiment(self._experiment)
-        elif study.study_name is not None:
+        elif (
+            study.study_name is not None
+            and study.study_name != "no-name-00000000-0000-0000-0000-000000000000"
+        ):
             mlflow.set_experiment(study.study_name)
 
         with mlflow.start_run(run_name=trial.number):
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -6,6 +6,7 @@ if type_checking.TYPE_CHECKING:
 
 try:
     import mlflow
+
     _available = True
 except ImportError as e:
     _import_error = e
@@ -21,9 +22,7 @@ def _check_mlflow_availability():
             "Mlflow is not available. Please install Mlflow to use this "
             "feature. It can be installed by executing `$ pip install "
             "mlflow`. For further information, please refer to the installation guide "
-            "of Mlflow. (The actual import error is as follows: "
-            + str(_import_error)
-            + ")"
+            "of Mlflow. (The actual import error is as follows: " + str(_import_error) + ")"
         )
 
 
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -59,12 +59,13 @@ class MlflowCallback(object):
     def __call__(self, study, trial):
         # type: (optuna.study.Study, optuna.structs.FrozenTrial) -> None
 
-        # set tracking_uri
+        # This sets the tracking_uri for Mlflow.
         if self._tracking_uri is not None:
             mlflow.set_tracking_uri(self._tracking_uri)
 
         with mlflow.start_run(run_name=trial.number):
-            # set experiment
+
+            # This sets the experiment of Mlflow.
             if self._experiment is None:
                 if (
                     study.study_name is None
@@ -76,15 +77,15 @@ class MlflowCallback(object):
             else:
                 mlflow.set_experiment(self._experiment)
 
-            # set metric
+            # This sets the metric for Mlflow.
             trial_value = trial.value if trial.value is not None else float("nan")
             metric_name = self._metric_name if self._metric_name is not None else "trial_value"
             mlflow.log_metric(metric_name, trial_value)
 
-            # set params
+            # This sets the params for Mlflow.
             mlflow.log_params(trial.params)
 
-            # set tags
+            # This sets the tags for Mlflow.
             tags = {}  # type: Dict[str, str]
             tags["trial_number"] = str(trial.number)
             tags["trial_datetime_start"] = str(trial.datetime_start)
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -84,10 +84,10 @@ class MLflowCallback(object):
             <https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.set_tracking_uri>`_
             for more details.
         metric_name:
-            Name of the metric. If not provided this will be called ``trial_value``.
+            Name of the metric. If not provided this will be called ``value``.
     """
 
-    def __init__(self, tracking_uri=None, metric_name=None):
+    def __init__(self, tracking_uri=None, metric_name="value"):
         # type: (Optional[str], Optional[str]) -> None
 
         _check_mlflow_availability()
@@ -109,8 +109,7 @@ class MLflowCallback(object):
 
             # This sets the metric for MLflow.
             trial_value = trial.value if trial.value is not None else float("nan")
-            metric_name = self._metric_name if self._metric_name is not None else "value"
-            mlflow.log_metric(metric_name, trial_value)
+            mlflow.log_metric(self._metric_name, trial_value)
 
             # This sets the params for MLflow.
             mlflow.log_params(trial.params)
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -89,7 +89,7 @@ class MLflowCallback(object):
             Name of the metric. If not provided this will be called ``trial_value``.
     """
 
-    def __init__(self, tracking_uri=None, experiment=None, metric_name=None):
+    def __init__(self, tracking_uri=None, experiment_name=None, metric_name=None):
         # type: (str, str, str) -> None
 
         _check_mlflow_availability()
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -34,8 +34,8 @@ class MLflowCallback(object):
     """Callback to track optuna trials with MLflow.
 
     This callback adds relevant information that is
-    tracked by Optuna to MLflow. You must set 'study.study_name' inside
-    Optuna because it is needed to set the MLflow experiment name.
+    tracked by Optuna to MLflow. The MLflow experiment
+    will be named after the Optuna study name.
 
     Example:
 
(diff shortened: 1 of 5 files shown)


PhilipMay:
@sile @toshihikoyanase @g-votte 
Can someone of you please help me fix this "error" from the ci:
``` bash
#!/bin/bash -eo pipefail
. venv/bin/activate
mypy .

optuna/integration/mlflow.py:49: error: "FixedTrial" has no attribute "number"
optuna/integration/mlflow.py:51: error: "FixedTrial" has no attribute "value"
optuna/integration/mlflow.py:60: error: "FixedTrial" has no attribute "number"
optuna/integration/mlflow.py:62: error: "FixedTrial" has no attribute "datetime_complete"
optuna/integration/mlflow.py:63: error: "FixedTrial" has no attribute "state"
Found 5 errors in 1 file (checked 122 source files)

Exited with code exit status 1
```

g-votte:
@PhilipMay 
Thanks for your PR! Let us take some time for reviewing.

> Can someone of you please help me fix this "error" from the ci:

I think it's because the expected type is not `FixedTrial` but `FrozenTrial`.

PhilipMay:
@g-votte thanks. Now I have this issue:
```
#!/bin/bash -eo pipefail
. venv/bin/activate
autopep8 . -r --diff --exit-code

/bin/bash: line 1: autopep8: command not found

Exited with code exit status 127

```

g-votte:
@PhilipMay 
The following PR might be the cause, where autopep8 has just been replaced with black.
Could you try merging and resolving the latest master?

https://github.com/optuna/optuna/pull/1030

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,97 @@
...
+from optuna import type_checking
+
+if type_checking.TYPE_CHECKING:
+    from typing import Dict  # NOQA
+
+
+class MlflowCallback(object):
+    """Callback to track optuna trials with Mlflow.
+
+    This callback adds all relevant information that is
```
g-votte: How about removing `all` from this sentence? This is because there's missing information in current implementation such as `intermediate_values`. Even if those pieces of information will be added in the future, we still might want to exclude some information such as `system_attrs`.

PhilipMay: fixed

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,97 @@
...
+    tracked by Optuna to Mlflow.
+
+    Example:
+
+        Add Mlflow callback to optuna optimization.
+
+        .. code::
```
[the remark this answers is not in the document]
PhilipMay: fixed

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,97 @@
...
+        self._metric_name = metric_name
+
+    def __call__(self, study, trial):
+        # type: (optuna.study.Study, optuna.structs.FrozenTrial) -> None
+
+        # set experiment
+        if self._experiment is None:
+            if (
+                study.study_name is None
+                or study.study_name == "no-name-00000000-0000-0000-0000-000000000000"
```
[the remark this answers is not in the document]
PhilipMay: I personaly would avoid logging to the default but you are right - it should be possible.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,97 @@
...
+        # type: (str, str, str) -> None
+
+        self._tracking_uri = tracking_uri
+        self._experiment = experiment
+        self._metric_name = metric_name
+
+    def __call__(self, study, trial):
+        # type: (optuna.study.Study, optuna.structs.FrozenTrial) -> None
+
+        # set experiment
```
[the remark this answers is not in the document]
PhilipMay: fixed

optuna/integration/mlflow.py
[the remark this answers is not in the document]
PhilipMay: changed

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,97 @@
...
+                or study.study_name == "no-name-00000000-0000-0000-0000-000000000000"
+            ):
+                raise ValueError("Either 'experiment' or 'study.study_name' must be set!")
+            else:
+                mlflow.set_experiment(study.study_name)
+        else:
+            mlflow.set_experiment(self._experiment)
+
+        # set tracking_uri
+        if self._tracking_uri is not None:
```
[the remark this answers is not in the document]
PhilipMay: yes. good point. I fixed this by setting tracking_uri first.

optuna/integration/mlflow.py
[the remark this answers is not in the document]
PhilipMay: I am back to `code` instead of `testcode` because  I can not provide a valid `tracking_uri`. I have not http URL and do not want to write to a directory. What do you think?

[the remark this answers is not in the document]
PhilipMay: unresolving again

tests/integration_tests/test_mlflow.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+from unittest.mock import patch
+
+import optuna
+from optuna.integration.mlflow import MlflowCallback
+from ..test_structs import _create_frozen_trial
```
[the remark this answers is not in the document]
PhilipMay: @g-votte how do you think I could avoid the relative import? Should I refactor `_create_frozen_trial` into `optuna.testing.integration.py`?

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,97 @@
+import mlflow
```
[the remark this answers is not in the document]
PhilipMay: removed the hard dependency

g-votte:
@PhilipMay 
Regarding [the doctest problem](https://github.com/optuna/optuna/pull/1028#discussion_r396016459), how about using `tempfile.mkdtemp` and cleaning up after testing? For example:

```python
        .. testsetup::
            import pathlib
            import tempfile
            tempdir = tempfile.mkdtemp()
            YOUR_TRACKING_URI = pathlib.Path(tempdir).as_uri()
        .. testcode::
            import optuna
            from optuna.integration.mlflow import MlflowCallback

            def objective(trial):
                x = trial.suggest_uniform('x', -10, 10)
                return (x - 2) ** 2

            mlflc = MlflowCallback(
                tracking_uri=YOUR_TRACKING_URI,
                metric_name='my metric score',
            )

            study = optuna.create_study(study_name='my_study')
            study.optimize(objective, n_trials=10, callbacks=[mlflc])
        .. testcleanup::
            import shutil
            shutil.rmtree(tempdir)
```

With `testsetup` and `testcleanup`, you can keep the code readability by hiding the setup and cleanup procedures. The picture below is generated docs.

image

optuna/integration/mlflow.py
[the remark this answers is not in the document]
g-votte: @PhilipMay 
Regarding [the doctest problem](https://github.com/optuna/optuna/pull/1028#discussion_r396016459), how about using `tempfile.mkdtemp` and cleaning up after testing? For example:

```python
        .. testsetup::
            import pathlib
            import tempfile
            tempdir = tempfile.mkdtemp()
            YOUR_TRACKING_URI = pathlib.Path(tempdir).as_uri()
        .. testcode::
            import optuna
            from optuna.integration.mlflow import MlflowCallback

            def objective(trial):
                x = trial.suggest_uniform('x', -10, 10)
                return (x - 2) ** 2

            mlflc = MlflowCallback(
                tracking_uri=YOUR_TRACKING_URI,
                metric_name='my metric score',
            )

            study = optuna.create_study(study_name='my_study')
            study.optimize(objective, n_trials=10, callbacks=[mlflc])
        .. testcleanup::
            import shutil
            shutil.rmtree(tempdir)
```

With `testsetup` and `testcleanup`, you can keep the code readability by hiding the setup and cleanup procedures. The picture below is generated docs.

image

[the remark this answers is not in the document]
PhilipMay: many thanks - I made the change

PhilipMay:
@g-votte what is the reason for this error? can you please help me? https://circleci.com/gh/optuna/optuna/32401?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PhilipMay:
@g-votte how do you think I could avoid the relative import? Should I refactor `_create_frozen_trial` into `optuna.testing.integration.py`?

tests/integration_tests/test_mlflow.py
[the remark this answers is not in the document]
PhilipMay: fixed

PhilipMay:
All tests are green. I am done with this PR. Can you do a final review please?

tests/integration_tests/test_mlflow.py
```diff
@@ -0,0 +1,98 @@
...
+    # type: (py.path.local) -> None
+
+    db_file_name = "file:{}".format(tmpdir)
+
+    mlflc = MlflowCallback(
+        tracking_uri=db_file_name, metric_name="my_metric", experiment="my_experiment"
+    )
+    study = optuna.create_study()
+    ft = _create_frozen_trial()
+    mlflc(study, ft)
```
g-votte: In addition to unit testing just for the function, how about adding more end-to-end cases where the callback is invoked via the `optimize` method? I think it improves coverages in terms of practical scenarios.

One of the simplest cases would be just like what we test in the doctest.

```python
def objective(trial):
    x = trial.suggest_uniform('x', -10, 10)
    return (x - 2) ** 2

tempdir = tempfile.mkdtemp()
uri = pathlib.Path().as_uri()

mlflc = MlflowCallback(tracking_uri=uri)
study = optuna.create_study(study_name='my_study')
study.optimize(objective, n_trials=10, callbacks=[mlflc])

# Assert mlflow data.
...
```

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,137 @@
...
+    if not _available:
+        raise ImportError(
+            "Mlflow is not available. Please install Mlflow to use this "
+            "feature. It can be installed by executing `$ pip install "
+            "mlflow`. For further information, please refer to the installation guide "
+            "of Mlflow. (The actual import error is as follows: " + str(_import_error) + ")"
+        )
+
+
+class MlflowCallback(object):
```
harupy: @PhilipMay @g-votte 
Should `Ml` be uppercase?

```suggestion
class MLflowCallback(object):
```

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,126 @@
...
+    _available = False
+    mlflow = object
+
+
+def _check_mlflow_availability():
+    # type: () -> None
+
+    if not _available:
+        raise ImportError(
+            "Mlflow is not available. Please install Mlflow to use this "
```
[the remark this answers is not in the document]
PhilipMay: I changed that.

setup.py
```diff
@@ -51,8 +51,15 @@ def get_extras_require() -> Dict[str, List[str]]:
...
-        "document": ["sphinx", "sphinx_rtd_theme",],
+        "doctest": [
+            "cma",
+            "pandas",
+            "plotly>=4.0.0",
+            "scikit-learn>=0.19.0",
+            "scikit-optimize",
+            "mlflow",
+        ],
+        "document": ["sphinx", "sphinx_rtd_theme", "mlflow",],
```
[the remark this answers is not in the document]
PhilipMay: removed mlflow from document

optuna/integration/mlflow.py
PhilipMay: I fixed that. Thanks.

tests/integration_tests/test_mlflow.py
PhilipMay: I implemented an end to end test.

PhilipMay:
Thanks for your comments and recommendations. I incorporated everything. Are there any more issues you see?

tests/integration_tests/test_mlflow.py
```diff
@@ -0,0 +1,130 @@
+import py
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+import optuna
+from optuna.integration.mlflow import MLflowCallback
+from tests.test_structs import _create_frozen_trial
+from tests.test_study import func
```
g-votte: I don’t think we should reuse those functions, because they seems not to be designed for usages outside `test_structs` and `test_study`. If we need similar logic, there would be two alternative options:
- (a) defining similar functions in `test_mlflow`, or
- (b) extracting the functions to more common places to somewhere under `optuna.testing`.

I personally much prefer (a), because I think `optuna.testing` should be the place for more common utilities than those functions. 

Even though we'd make some small duplicated logic, how about just taking option (a)?

PhilipMay: Ok. I refactored the functions into `test_mlflow`.

PhilipMay:
Merged and rebased master again.

tests/integration_tests/test_mlflow.py
```diff
@@ -0,0 +1,160 @@
...
+    mlflc(study, ft)
+
+    assert log_metric.called
+    assert log_metric.call_count == 1
+    call_args = log_metric.call_args_list[0][0]
+    assert call_args[0] == "trial_value"
+    assert call_args[1] == 0.2
+
+
+@patch("mlflow.log_metric")
```
g-votte: I think we no longer need mock for those tests, because we're currently able to put the data to the temporary directory instead of an external MLflow service. So, removing mock would make the test cases more practical.

What do you think?

PhilipMay: Yes that's right. I removed the mocks and replaced them by loading the experiments and runs with the MLFlow client.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,137 @@
...
+        # This sets the tracking_uri for MLflow.
+        if self._tracking_uri is not None:
+            mlflow.set_tracking_uri(self._tracking_uri)
+
+        # This sets the experiment of MLflow.
+        if self._experiment is not None:
+            mlflow.set_experiment(self._experiment)
+        elif (
+            study.study_name is not None
+            and study.study_name != "no-name-00000000-0000-0000-0000-000000000000"
```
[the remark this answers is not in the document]
PhilipMay: I fixed that and wrote a test for each if branch.

tests/integration_tests/test_mlflow.py
```diff
@@ -0,0 +1,160 @@
...
+@patch("mlflow.log_params")
+@patch("mlflow.set_tags")
+def test_end_to_end(set_tags, log_params, log_metric, tmpdir):
```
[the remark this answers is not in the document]
PhilipMay: Improved naming of the test functions.

tests/integration_tests/test_mlflow.py
```diff
@@ -0,0 +1,160 @@
...
+    # type: (optuna.trial.Trial, float) -> float
+
+    x = trial.suggest_uniform("x", -x_max, x_max)
+    y = trial.suggest_loguniform("y", 20, 30)
+    z = trial.suggest_categorical("z", (-1.0, 1.0))
+    assert isinstance(z, float)
+    return (x - 2) ** 2 + (y - 25) ** 2 + z
+
+
+def test_happy_case(tmpdir):
```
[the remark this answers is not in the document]
PhilipMay: Improved naming of the test functions.

PhilipMay:
Resolved all review items and merged + rebased master again.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,141 @@
...
+
+    if not _available:
+        raise ImportError(
+            "MLflow is not available. Please install MLflow to use this "
+            "feature. It can be installed by executing `$ pip install "
+            "mlflow`. For further information, please refer to the installation guide "
+            "of MLflow. (The actual import error is as follows: " + str(_import_error) + ")"
+        )
+
+
```
g-votte, line 28: Please add the `experimental` decorator like [this](https://github.com/optuna/optuna/blob/2d5f24b06eed56ece72b8dfa878135bb4bb63779/optuna/integration/lightgbm_tuner/__init__.py#L23). I think we've better release this feature as an experimental version at first.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,141 @@
...
+            shutil.rmtree(tempdir)
+
+        .. testoutput::
+            :hide:
+            :options: +NORMALIZE_WHITESPACE
+
+            INFO: 'my_study' does not exist. Creating a new experiment
+
+    Args:
+        tracking_uri:
```
g-votte, line 75: The description of this argument seems to be copied from the [MLflow document](https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.set_tracking_uri). Can't we just reference the original document of MLflow? For example:

```
The URI of the MLflow tracking server.

Please refer to `mlflow.set_tracking_uri
<https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.set_tracking_uri>`_
for more details.
```

Otherwise, we have to change those sentences whenever there's any change in the original document.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,141 @@
...
+            # This sets the metric for MLflow.
+            trial_value = trial.value if trial.value is not None else float("nan")
+            metric_name = self._metric_name if self._metric_name is not None else "trial_value"
+            mlflow.log_metric(metric_name, trial_value)
+
+            # This sets the params for MLflow.
+            mlflow.log_params(trial.params)
+
+            # This sets the tags for MLflow.
+            tags = {}  # type: Dict[str, str]
```
g-votte, line 130: Those pieces of information are similar to what we can get with [Study.trials_dataframe](https://optuna.readthedocs.io/en/latest/reference/study.html#optuna.study.Study.trials_dataframe). We may want to be consistent about the naming convention of their keys.

How about renaming keys as:
- `number` instead of `trial_number`
- `datetime_start` instead of `trial_datetime_start`
- `datetime_complete` instead of `trial_datetime_complete`
- `state` instead of `trial_state`

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,141 @@
...
+            tags["trial_datetime_start"] = str(trial.datetime_start)
+            tags["trial_datetime_complete"] = str(trial.datetime_complete)
+            tags["trial_state"] = str(trial.state)
+            tags["study_direction"] = str(study.direction)
+            tags.update(trial.user_attrs)
+            distributions = {
+                (k + "_distribution"): str(v) for (k, v) in trial.distributions.items()
+            }
+            tags.update(distributions)
+            mlflow.set_tags(tags)
```
g-votte, line 141: How about adding the information of [`user_attrs`](https://optuna.readthedocs.io/en/latest/reference/trial.html#optuna.trial.Trial.user_attrs) as well?

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,141 @@
...
+
+            # This sets the params for MLflow.
+            mlflow.log_params(trial.params)
+
+            # This sets the tags for MLflow.
+            tags = {}  # type: Dict[str, str]
+            tags["trial_number"] = str(trial.number)
+            tags["trial_datetime_start"] = str(trial.datetime_start)
+            tags["trial_datetime_complete"] = str(trial.datetime_complete)
+            tags["trial_state"] = str(trial.state)
```
g-votte, line 134: In [`Study.trials_dataframe`](https://optuna.readthedocs.io/en/latest/reference/study.html#optuna.study.Study.trials_dataframe), the common prefix, `TrialState.`, is removed by [those lines](https://github.com/optuna/optuna/blob/034cfdf1ff9cddcd9c0e5301f9a476181c5f2b41/optuna/study.py#L492-L494). To make the output more human-readable, how about taking the same approach here?

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,141 @@
...
+              ``databricks://<profileName>``.
+        experiment:
+            Name of MLflow experiment to be activated. If not set ``study.study_name``
+            will be taken. Either ``experiment`` or ``study.study_name`` must be set.
+        metric_name:
+            Name of the metric. If not provided this will be called ``trial_value``.
+    """
+
+    def __init__(self, tracking_uri=None, experiment=None, metric_name=None):
+        # type: (str, str, str) -> None
```
g-votte, line 93: The arguments seems to be optional. Please replace `str` with `Optional[str]`.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,141 @@
...
+              `profile <https://github.com/databricks/databricks-cli#installation>`_,
+              ``databricks://<profileName>``.
+        experiment:
+            Name of MLflow experiment to be activated. If not set ``study.study_name``
+            will be taken. Either ``experiment`` or ``study.study_name`` must be set.
+        metric_name:
+            Name of the metric. If not provided this will be called ``trial_value``.
+    """
+
+    def __init__(self, tracking_uri=None, experiment=None, metric_name=None):
```
hvy, line 92: Just curious but in what cases would you like to specify a different experiment name than the one implied by the study (name)?

PhilipMay: Since Optuna does not force the user to provide a study name I do not want to force the user to provide one when he is using this callback. So I introduced the experiment name parameter and enforce the user to provide either this param or a study name.

@hvy Do you think I should enforce the user to provide a study name?

optuna/integration/mlflow.py
PhilipMay: I think naming the metric would be a very useful feature for me. I have cases where I do different experiments and change the metric. From F1 to average precision or from accuracy to ROC_AUC. Giving the metric an optional explicit name is very useful in these cases.

optuna/integration/mlflow.py
PhilipMay: > 
> 
> How about adding the information of [`user_attrs`](https://optuna.readthedocs.io/en/latest/reference/trial.html#optuna.trial.Trial.user_attrs) as well?

@g-votte I did add them. See this line: `tags.update(trial.user_attrs)` Or did you mean anything else?

optuna/integration/mlflow.py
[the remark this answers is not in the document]
hvy: Sorry if it wasn't clear enough but you must also change the other occurrences of this variable.

hvy: I see, thanks for clarifying your thoughts on this. I think that if the user can specify the experiment name, he/she could as well specify the study name, and those would be consistent. I think this parameter makes it look like they should be treated independently (which I now understand is not the case). How about omitting this parameter and writing in the docs how the experiment name is determined?

optuna/integration/mlflow.py
g-votte: Sorry, I just overlooked that. Please just ignore this comment.

optuna/integration/mlflow.py
PhilipMay: @hvy What do you suggest happens when the user does not provide a study name? Raise an exception? I do not like to use the default experiment name because that is the place where "garbage" is collected when the user forgets to provide a study name.

optuna/integration/mlflow.py
PhilipMay: Ok. Added it for version 1.3.0

optuna/integration/mlflow.py
PhilipMay: Good idea. I changed that.

optuna/integration/mlflow.py
PhilipMay: Changed that.

optuna/integration/mlflow.py
PhilipMay: Good point. Changed this.

optuna/integration/mlflow.py
PhilipMay: > Those pieces of information are similar to what we can get with [Study.trials_dataframe](https://optuna.readthedocs.io/en/latest/reference/study.html#optuna.study.Study.trials_dataframe). We may want to be consistent about the naming convention of their keys.
> 
> How about renaming keys as:
> 
>     * `number` instead of `trial_number`
> 
>     * `datetime_start` instead of `trial_datetime_start`
> 
>     * `datetime_complete` instead of `trial_datetime_complete`
> 
>     * `state` instead of `trial_state`

Removed the prefixes.

optuna/integration/mlflow.py
PhilipMay: Since 1.3.0 is release today - should I change this to 1.4.0?

g-votte: Right; please update this to 1.4.0.

optuna/integration/mlflow.py
g-votte: > that is the place where "garbage" is collected

Hmm, I’m not sure if this is a general assumption. 

Even if it is the case, we can later change the behavior depending on users' feedback, since this is an `experimental` feature. Thus, I think we should adopt the simpler interface so far, where we just omit the parameter as @hvy mentions and just take the study name.

optuna/integration/mlflow.py
g-votte: > Removed the prefixes.

Thanks for your quick change!

> I think naming the metric would be a very useful feature for me.

This feature might be worth, but I think we should at first take the simpler choice of omitting the parameter given the maintenance cost. When a different metric name is necessary, the user has an alternative way of adding another metric outside this callback.

optuna/integration/mlflow.py
hvy: It might help if you could provide some concrete scenario. It sounds to me like if the user would forget to specify the study name (--> "garbage"), he/she is most likely to forget specifying the experiment name as well. Additionally, having to recall that there are two interfaces to specify the experiment name incurs additional load. I otherwise agree with @g-votte what we should start simple and expand the API as needed.

optuna/integration/mlflow.py
PhilipMay: Changed experimental version number to 1.4.0

optuna/integration/mlflow.py
PhilipMay: Ok. I made a suggestion. Now the MLflow expermiment name is taken from the Optuna study_name and from nowhere else. But I enforce the study_name to be set or an exception is raised.

What do you think?

optuna/integration/mlflow.py
PhilipMay: Well, honestly. For me it is very importat to be able to change the metric name to something else then just "value". Here is a suggestion that reduces comeplexity (maintainance cost) but still makes it possible to change the metric name after object creation.

``` python
    def __init__(self, tracking_uri=None):
        # type: (Optional[str]) -> None

        _check_mlflow_availability()

        self._tracking_uri = tracking_uri
        self._METRIC_NAME = "value"

    def __call__(self, study, trial):
        # type: (optuna.study.Study, optuna.structs.FrozenTrial) -> None

        # This sets the tracking_uri for MLflow.
        if self._tracking_uri is not None:
            mlflow.set_tracking_uri(self._tracking_uri)

        # This sets the experiment of MLflow.
        if (
            study.study_name is not None
            and study.study_name != "no-name-00000000-0000-0000-0000-000000000000"
        ):
            mlflow.set_experiment(study.study_name)
        else:
            raise ValueError(
                "You must set 'study.study_name' inside Optuna! This is needed to set the MLflow experiment name."
            )

        with mlflow.start_run(run_name=str(trial.number)):

            # This sets the metric for MLflow.
            trial_value = trial.value if trial.value is not None else float("nan")
            metric_name = self._metric_name if self._metric_name is not None else "value"
            mlflow.log_metric(self._METRIC_NAME, trial_value)

```

What do you think?

PhilipMay:
@hvy @g-votte 
I made all requested changes. Two items are still open:

- "experiment naming": I made a suggestion here: https://github.com/optuna/optuna/pull/1028#discussion_r403291546 and in code here: https://github.com/optuna/optuna/pull/1028/commits/c9eb7f070163442a44f855e7ab19c54ae3f3d97c
- "metric naming": I made a suggestion here: https://github.com/optuna/optuna/pull/1028#discussion_r403298014

What do you think?

optuna/integration/mlflow.py
hvy: I don't think a private attribute is a good compromise as it clearly communicates to the user that it shouldn't be accessed/modified. If this feature is requested by MLflow users, I believe that we could provide it with a proper API, as in the current implementation.

optuna/integration/mlflow.py
hvy: Thanks @PhilipMay for your changes. On the exception however, I don't think raising one here speaks for good design. First, I'm not sure the general user would find that intuitive, and second the name is currently hardcoded in the MLflow integration module and will only work for the in memory storage and seems fragile.

hvy:
Thanks as always for your changes. I see that the discussion is dragging out a bit but I tried to reply to your latest comments if you could take a look.

optuna/integration/mlflow.py
PhilipMay: Ok, thanks. So the implementation stays as it is.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,144 @@
...
+        # type: (optuna.study.Study, optuna.structs.FrozenTrial) -> None
+
+        # This sets the tracking_uri for MLflow.
+        if self._tracking_uri is not None:
+            mlflow.set_tracking_uri(self._tracking_uri)
+
+        # This sets the experiment of MLflow.
+        if (
+            study.study_name is not None
+            and study.study_name != "no-name-00000000-0000-0000-0000-000000000000"
```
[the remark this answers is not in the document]
PhilipMay: Ok. Noticed that when using a sql study the default name is like this `no-name-4d948f5e-3e89-4b7c-9207-d77837363042`. I changed the implementation to just use study_name.

optuna/integration/mlflow.py
PhilipMay: Changed it to just use study_name.

PhilipMay:
I incorporated everything and merged + rebased master again.
@hvy and @g-votte do you see any more issues?

PhilipMay:
For some reason the github test has an error. Is there an easy fix?

PhilipMay:
merged + rebased master again, all tests are green.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,135 @@
...
+            The URI of the MLflow tracking server.
+
+            Please refer to `mlflow.set_tracking_uri
+            <https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.set_tracking_uri>`_
+            for more details.
+        metric_name:
+            Name of the metric. If not provided this will be called ``trial_value``.
+    """
+
+    def __init__(self, tracking_uri=None, metric_name=None):
```
hvy, line 90: You don't have to hide the default argument value behind and a `None` and could use the actual `"value"`. Then you could simplify the documentation of this parameter omitting the explanation about the default value, the type hint and also the logic of checking whether it's a `None`.
```suggestion
    def __init__(self, tracking_uri=None, metric_name="value"):
```

tests/integration_tests/test_mlflow.py
```diff
@@ -0,0 +1,109 @@
...
+
+    x = trial.suggest_uniform("x", -x_max, x_max)
+    y = trial.suggest_loguniform("y", 20, 30)
+    z = trial.suggest_categorical("z", (-1.0, 1.0))
+    assert isinstance(z, float)
+    trial.set_user_attr("my_user_attr", "my_user_attr_value")
+    return (x - 2) ** 2 + (y - 25) ** 2 + z
+
+
+def test_study_name(tmpdir):
```
hvy, line 19: I'm a bit skeptical about using `py` which is in maintenance mode but it might be fine as we're using a `pytest` version that's meant to support it.

PhilipMay: Well, we need a temp directory to test without mocks. The type of this temp dir is in ˋpyˋ module. Nothing we can do here.

optuna/integration/mlflow.py
PhilipMay: Ok. Changed this.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,135 @@
...
+
+    Args:
+        tracking_uri:
+            The URI of the MLflow tracking server.
+
+            Please refer to `mlflow.set_tracking_uri
+            <https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.set_tracking_uri>`_
+            for more details.
+        metric_name:
+            Name of the metric. If not provided this will be called ``trial_value``.
```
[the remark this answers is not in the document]
PhilipMay: Ok. Changed.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+
+
+@experimental("1.4.0")
+class MLflowCallback(object):
+    """Callback to track optuna trials with MLflow.
+
+    This callback adds relevant information that is
+    tracked by Optuna to MLflow. You must set 'study.study_name' inside
+    Optuna because it is needed to set the MLflow experiment name.
```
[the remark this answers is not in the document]
PhilipMay: Docstring updated.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,134 @@
...
+
+            Please refer to `mlflow.set_tracking_uri
+            <https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.set_tracking_uri>`_
+            for more details.
+        metric_name:
+            Name of the metric. If not provided this will be called ``value``.
+    """
+
+    def __init__(self, tracking_uri=None, metric_name="value"):
+        # type: (Optional[str], Optional[str]) -> None
```
hvy, line 91: A detail but could you align the type hint to the new default value?
```suggestion
        # type: (Optional[str], str) -> None
```

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Args:
+        tracking_uri:
+            The URI of the MLflow tracking server.
+
+            Please refer to `mlflow.set_tracking_uri
+            <https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.set_tracking_uri>`_
+            for more details.
+        metric_name:
+            Name of the metric. If not provided this will be called ``value``.
```
hvy, line 87: Ditto.
```suggestion
            Name of the metric.
```

tests/integration_tests/test_mlflow.py
```diff
@@ -0,0 +1,109 @@
+from mlflow.tracking import MlflowClient
+import py
+
+import optuna
+from optuna.integration.mlflow import MLflowCallback
+
+
+def _objective_func(trial, x_max=1.0):
```
hvy, line 8: A nitpick, if you have time, we could remove and hard code the `x_max` argument. It's not passed any non-default value anyway.

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,134 @@
...
+            "MLflow is not available. Please install MLflow to use this "
+            "feature. It can be installed by executing `$ pip install "
+            "mlflow`. For further information, please refer to the installation guide "
+            "of MLflow. (The actual import error is as follows: " + str(_import_error) + ")"
+        )
+
+
+@experimental("1.4.0")
+class MLflowCallback(object):
+    """Callback to track optuna trials with MLflow.
```
g-votte, line 34: Please be consistent about the use of upper/lower case of "Optuna," starting with the upper case.

tests/integration_tests/test_mlflow.py
g-votte, line 8: To improve readability, we may want to simplify this objective function:
- as @hvy mentions, we can get rid of the `x_max` argument;
- `assert isinstance(z, float)` would not be necessary in this test scenario.

tests/integration_tests/test_mlflow.py
```diff
@@ -0,0 +1,109 @@
...
+        == "LogUniformDistribution(high=30, low=20)"
+    )
+    assert (
+        first_run_dict["data"]["tags"]["z_distribution"]
+        == "CategoricalDistribution(choices=(-1.0, 1.0))"
+    )
+    assert first_run_dict["data"]["tags"]["my_user_attr"] == "my_user_attr_value"
+
+
+def test_metric_name(tmpdir):
```
g-votte, line 65: Most lines of this function are duplicated with `test_study_name`.  How about using `parametrize` to avoid that? See also [this example](https://github.com/optuna/optuna/blob/94b56917359d71a34d02e5c048e048caaef6d9ea/tests/samplers_tests/test_samplers.py#L39).

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,134 @@
...
+            tags["number"] = str(trial.number)
+            tags["datetime_start"] = str(trial.datetime_start)
+            tags["datetime_complete"] = str(trial.datetime_complete)
+
+            # Set TrialState and convert it to str and remove the common prefix.
+            trial_state = trial.state
+            if isinstance(trial_state, structs.TrialState):
+                tags["state"] = str(trial_state).split(".")[-1]
+
+            tags["direction"] = str(study.direction)
```
g-votte, line 128: To be consistent with the trial state, how about trimming the header of the direction as well?

optuna/integration/mlflow.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Args:
+        tracking_uri:
+            The URI of the MLflow tracking server.
+
+            Please refer to `mlflow.set_tracking_uri
+            <https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.set_tracking_uri>`_
+            for more details.
+        metric_name:
+            Name of the metric.
```
g-votte, line 87: It'd be difficult for the user to immediately understand what the `metric_name` argument is like. How about describing its effect and advantage in the docs?

tests/integration_tests/test_mlflow.py
PhilipMay: changed that

optuna/integration/mlflow.py
PhilipMay: Fixed.

tests/integration_tests/test_mlflow.py
PhilipMay: ok - removed the assert

optuna/integration/mlflow.py
PhilipMay: good catch. changed that.

g-votte:
@harupy  
Hi, do you have any comment on the current design and implementation?

Because the reviewers are not so familiar with MLflow, we'd greatly appreciate any feedback from an MLflow expert.

tests/integration_tests/test_mlflow.py
PhilipMay: added it again or I get this error from CI tests:

```
#!/bin/bash -eo pipefail
. venv/bin/activate
mypy .

tests/integration_tests/test_mlflow.py:16: error: Unsupported operand types for + ("float" and "None")
tests/integration_tests/test_mlflow.py:16: error: Unsupported operand types for + ("float" and "str")
tests/integration_tests/test_mlflow.py:16: note: Right operand is of type "Union[None, bool, int, float, str]"
Found 2 errors in 1 file (checked 133 source files)

Exited with code exit status 1
```

tests/integration_tests/test_mlflow.py
PhilipMay: I think you are right. The two test functions are not dry. But I do not think that parametrization is the right thing to do since I would have to introduce if statements inside the test functions to distinguish between the tests. 

I could refactor both test functions and move the common calls to a single function. What do you think about this?

optuna/integration/mlflow.py
PhilipMay: I improved the docstring here.

harupy:
@g-votte @PhilipMay 

What do you think about using nested runs?

mlflow-demo

In the attached example, I use two types of runs.

- parent:
  - stores the summary of optimization (plots, best params, best value).
- child
  - nested under the parent run.
  - stores params and value of each tral.

PhilipMay:
> @g-votte @PhilipMay
> 
> What do you think about using nested runs?
> In the attached example, I use two types of runs.
> 
>     * parent:
>       * stores the summary of optimization (e.g. plots, best params, best metric).
>     * child
>       * nested under the parent run.
>       * stores params and metric of each tral.

Thanks for this idea @harupy .
I think that it is valid and a possible way to add diagrams and so on. 

Several times we has a discussion about simplicity during this PR. One issue that needs to be solved is that when you cancel the Optuna training with "KeyboardInterrupt" you have to catch this, generate plots, write to parent and then throw again. An other issue is that you have to update the parent when optuna training is resumed. And the most important question is: When you do distributed runs with multiple computers who is the last client that writes the summary? Or do you generate and write the summary after each trial?

To be honest this is something I do not want to implement. Because I do not need it, because I wanted to contribute a lightwight MLflow adapter and I also want to focus other topics.

My suggestion is: Lets try to bring this PR to a good 80 % solution and we can add more features later when someone wants to take time to code them.
@g-votte @hvy what do you think?

tests/integration_tests/test_mlflow.py
PhilipMay: Ok. Resolved by your code change.

tests/integration_tests/test_mlflow.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    assert (
+        first_run_dict["data"]["tags"]["y_distribution"]
+        == "LogUniformDistribution(high=30, low=20)"
+    )
+    assert (
+        first_run_dict["data"]["tags"]["z_distribution"]
+        == "CategoricalDistribution(choices=(-1.0, 1.0))"
+    )
+    assert first_run_dict["data"]["tags"]["my_user_attr"] == "my_user_attr_value"
```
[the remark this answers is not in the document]
PhilipMay: Accepted the suggested change.

PhilipMay:
Merged and rebased master again.

harupy:
@PhilipMay 

Thanks for the comment.

>  Lets try to bring this PR to a good 80 % solution and we can add more features later when someone wants to take time to code them.

You're right. I agree😄

PhilipMay:
All remarks are handled, code changes accepted and tests are green.

PhilipMay:
> @PhilipMay
> Thanks a lot for your long-running contribution!

Yes - thanks. Was fun with you implementing and discussing this stuff. Although I initialy thought that it would be easier and quicker I learned a lot from your comments and remarks.

Issue #821, opened 2019-12-27
add trials to study / build study from trials


louisabraham:
I would like to add trials to studies or artificially build a study from trials.

The second method might solve https://github.com/optuna/optuna/issues/685 as well

## Motivation

In some workflow, you might want to optimize different metrics that are computed from the same results.

For example, in classification problems, people hesitate to maximize the cross-entropy, the accuracy, the F1 score, etc…
Similarly, when doing k-fold cross-validation, people might want to optimize the average or the minimum of the fold scores.

In all these cases, it is very cheap to compute all the scores instead of just one, and this data could be provided to avoid cold start in other runs.

## Description

I think that providing an argument `trials` in `create_study` would be the most practical solution, as it avoids having to reindex a database.
It also solves https://github.com/optuna/optuna/issues/685 without having to actually delete trials because users will be able to recreate a study object with a clean list of trials.

louisabraham:
I saw the related issue #745 that has the same motivation. Sorry for not searching more.

This feature could be implemented with `Study._append_trial`. I think that such a workaround would work for #685 as well.

sile:
Thank you for your proposal!

Providing a feature to add trials to studies or build a study from trials sounds useful and, as you said, it could be implemented on top of `Study._append_trial()`.  Let me label this issue as a contribution welcomed.
I think that there are a lot of design choices to implement this feature (e.g., What is the best API to provide this feature? [a] "Adding `trials` option to `optuna.create_study` method", [b] "Creating a new function such as `optuna.build_study`", [c] "Adding a method such as `import_{trials|study}` to `Study` class", etc), but we could discuss the design on a draft (WIP) PR if someone creates it.

hvy:
@louisabraham Two new experimental APIs were introduced in #1335, one for adding trials to a study and for creating arbitrary trials (this is a rather low-level API). They should address this issue. Have you taken a look at it? I'm wondering if we cannot close this issue as resolved.

louisabraham:
Sure, it looks very much resolved. I'll play with them later. Thank you for your hard work!

PR #1029, opened 2020-03-15
[WIP] add build study

tohmae:
This PR is proposal for supporting import study from previous study.
This PR addresses issue #821 .

I do not have confidence that I implement according to feature in #821.
If I misunderstand, I am sorry.

The usage is as follows.
```python
import optuna
import optuna.distributions as distributions

def objective(trial):
    x = trial.suggest_int('x', 0, 10)
    y = trial.suggest_uniform('y', 0, 10)
    z = trial.suggest_categorical('z',[0,1])
    w = trial.suggest_discrete_uniform('w', 0.1, 1.0, 0.1)
    return x ** 2 + (y - 5)** 2 + z + (w - 0.2) ** 2


study = optuna.create_study()
study.optimize(objective, n_trials=50, n_jobs=1)

# add parameters and modify distributions in objective
def new_objective(trial):
    x = trial.suggest_int('x', 0, 10)
    y = trial.suggest_uniform('y', 4, 6)
    z = trial.suggest_categorical('z',[0])
    w = trial.suggest_discrete_uniform('w', -2.0, 2.0, 0.2)
    add_param1 = trial.suggest_uniform('add_param1', -2, 2)
    add_param2 = trial.suggest_categorical('add_param2',[False, True])

    # if suggest end, add set_suggest_end function
    trial.set_suggest_end()

    if add_param2 is True:
        return x ** 2 + (y - 5)** 2 + z + (w - 0.2) ** 2 + add_param1
    return x ** 2 + (y - 5)** 2 + z + (w - 0.2) ** 2

# default value in add parameters
add_default_values = {
   'add_param1':0 ,
   'add_param2':False,
}

# usage
# import trials which params in modify_distributions and add params of new_objective.
new_study = optuna.study.import_study(study, new_objective, add_default_values)
new_study.optimize(new_objective, n_trials=20, n_jobs=1)

### CODE REVIEW

**Comment by tohmae** [comment]
I think that optuna is better than other hypterparameter optimization frameworks becase we can define search space and define objective function together in objective function (Define-by-Run).
So I want to import trials from study only using new objective function, not defining new distributions. If you set trial.set_suggest_end(), you can only define new search space(trial.suggest) in new objective function.

**Comment by HideakiImamura** [comment]
@tohmae Thank you for your proposal! I'm interested in this direction to address the issue discussed in Issue #821. Would you mind updating your changes to fit the latest version of the master branch? I would like to have discussions after the update.

**Comment by tohmae** [comment]
@HideakiImamura Thank you for your comments. I have rebased this branch.

**Comment by HideakiImamura** [comment]
@tohmae Thank you so much! I would like to discuss motivation and interface design more.

In my opinion, the motivation for introducing this `import_study` function is as follows. Suppose we want to optimize a new objective function instead of the objective function that was optimized in one study. The optimization history of the existing objective function may be available because the new objective function is "somewhat related" to the existing objective function. This is the motivation to introduce the `import_study` function.

Here, we can think of the following cases for "some relation" between the existing objective function f and the new objective function g.
(1) The arguments of f and g are the same, but the outputs are different.
(2) The arguments of f and g are different.

In the case of (1), this is what was at the heart of the discussion in Issue #821 and Issue #745. This can be done by the user creating a new objective function history from the `user_attr` of each trial that was saved in the optimization of the existing objective function, as mentioned in the comment of Issue #745.To do this, you can set a function like `converter` as an argument of the `import_study` function.  This `converter` function takes a trial and creates a new trial corresponding to the new objective function based on the `user_attr` of the given trial.

In the case of (2), as noted in the comment on Issue #821, it would have required addition to `trial.params` or, although not discussed, a deletion. Thank you for your thoughts and implementation on the addition to `trial.params`. To make the specification of the `import_study` function more complete, we also need to consider how to remove a particular `param` from `trial.params`. To realize such a function, you can set a function like `converter` as an argument of the `import_study` function.  This `converter` is a function that takes a trial and returns a new trial that has been modified with the trial.params.

From the above considerations, I think that the `import_study` function should have the following interface, what do you think?
`import_study(study: Study, converter: Callable[Trial, Trial])`
The user of this function needs to figure out how to create a new trial for each trial in an existing study.

I apologize for the length of the explanation. I look forward to hearing from you.

p.s. I feel that the name of the function should be `build_study`, rather than `import_study`.

**Comment by tohmae** [comment]
@HideakiImamura Thank you for your comment. I update source codes according to your comment. I add the function
build_study(study: Study, converter: Callable[[Trial], Trial]).
and new convert function
1. new_metric_conveter       -> case (1)
2. new_objective_conveter    -> case (2) .
If my understanding is mistaken, please tell me.

I do not understand the case that a particular param is removed from trial.params. I think the case a paricular param is fixed value. For example in "new_objecive(trial)"  z is fixed value 0. "z = trial.suggest_categorical('z',[0])"
Please tell me the sample case that a paricular param is removed from trial.params.

case (1)
```python
import optuna

def objective(trial, metric_name):
    x = trial.suggest_uniform('x', -10 , 10)
    y = trial.suggest_uniform('y', 3, 30)

    metrics = {
        'plus': x + y,
        'minus': x - y
    }
    trial.set_user_attr('metrics', metrics)
   
    return metrics[metric_name]


# Run the first optimization that uses 'plus' metric.
study_a  = optuna.create_study()
study_a.optimize(lambda trial: objective(trial, 'plus'), n_trials=10)

converter = optuna.study.new_metric_converter(metric_name = 'minus')
study_b = optuna.study.build_study(study_a, converter)

assert len(study_b.trials) == 10

# Run the second optimization that uses 'minus' metric.
study_b.optimize(lambda trial: objective(trial, 'minus'), n_trials=10)
assert len(study_b.trials) == 20

```

case (2)
```python
import optuna

def objective(trial):
    x = trial.suggest_int('x', 0, 10)
    y = trial.suggest_uniform('y', 0, 10)
    z = trial.suggest_categorical('z',[0,1])
    w = trial.suggest_discrete_uniform('w', 0.1, 1.0, 0.1)
    return x ** 2 + (y - 5)** 2 + z + (w - 0.2) ** 2


study = optuna.create_study()
study.optimize(objective, n_trials=50, n_jobs=1)


# add parameters and modify distributions in objective
def new_objective(trial):
    x = trial.suggest_int('x', 0, 10)
    y = trial.suggest_uniform('y', 4, 6)
    z = trial.suggest_categorical('z',[0])
    w = trial.suggest_discrete_uniform('w', -2.0, 2.0, 0.2)
    add_param1 = trial.suggest_uniform('add_param1', -2, 2)
    add_param2 = trial.suggest_categorical('add_param2',[False, True])

    # if suggest end, add set_suggest_end function
    trial.set_suggest_end()

    if add_param2 is True:
        return x ** 2 + (y - 5)** 2 + z + (w - 0.2) ** 2 + add_param1
    return x ** 2 + (y - 5)** 2 + z + (w - 0.2) ** 2

# add default value at new params
add_default_values = {
   'add_param1':0 ,
   'add_param2':False,
}

# usage
# import trials which params in modify_distributions and add params of new_objective.
converter = optuna.study.new_objective_converter(new_objective, add_default_values)
new_study = optuna.study.build_study(study, converter)
new_study.optimize(new_objective, n_trials=20, n_jobs=1)
```

hvy:
There seems to be a couple of different cases as described above and am wondering if we cannot provide APIs for dealing with all them in a less ad-hoc (e.g. relying on the users having set `user_attrs`) manner. For instance, although I haven't implemented and tested this yet, some more flexible API as follows. This might be too low level and difficult to use. Very open for comments and criticism.

```python
# Original study.
study1 = optuna.create_study()
study1.optimize(objective, n_trials=10)

# New study.
study2 = optuna.create_study()

# Filter.
study2.import_trials(filter(lambda t: ..., study1.trials))

# Map.
study2.import_trials(map(lambda t: optuna.create_trial(params=t.params, value=..., status=...), study1.trials))

# Enqueue.
study_2.import_trials([optuna.create_trial(params=..., status='waiting')])

# `Study.add_trials` might be shorter and more approachable
# `optuna.create_trial` must be designed as well
```

hvy:
Implemented a very naive version of the API above in https://github.com/optuna/optuna/pull/1335. Might spark some ideas.

tohmae:
Thank you for your comments.
I think that your suggested API is easy to understand and develp.
In paticular classifying into three patterns(filter, map, enqueue) is very good.

I want to develop converter functions as bellow.
(1) study2.import_trials(filter(lambda t: new_objective_filter(t), study1.trials))
(2) study2.import_trials(map(lambda t: new_metric_map(t), study1.trials))

Please let me know what you think.

hvy:
Sorry @tohmae for the delayed reply. Your use case with converters can probably help other developers too. I'm not sure though if it should be incorporated directly as an API of Optuna. You don't want to for instance create a tutorial or an example using the interfaces introduced in #1335? Then we could expose them to the user that way.

tohmae:
@hvy Thank you for your comment. I will create an example by google colaboratory.
Please wait a moment.

tohmae:
I create an example in the [notebook](https://colab.research.google.com/drive/1JGSPSb34vbBr0uFJra-IFH_ldWZYikq6).

hvy:
Sorry for the late response, and thanks a lot for your notebook demonstrating the examples. I'm still a bit concerned with the changes as they are. Let me elaborate on my previous comment (https://github.com/optuna/optuna/pull/1029#issuecomment-652767837). The converters that you introduce here are quite use-case specific, e.g. requiring user attributes with specific keys, so I'm a bit concerned that the maintenance cost is rather high for a utility that can be implemented by the users (using currently existing APIs, `Study.add_trial`). 

I really like the practicalities of the examples in the notebook though. You don't want to provide them without relying on the converters under `optuna.study`? Then we could merge them as examples, for instance.

tohmae:
Thank you for comments.  I delete converters under optuna.study, and it move to [examples/add_trial_sample.ipynb](https://github.com/optuna/optuna/pull/1029/files#diff-83779045cd4e1c364acac055e1385ab5).  
I create a new example in the [notebook](https://colab.research.google.com/drive/1_X6Ghqiq3cqZJaNwik51TSRbDqngsotG#scrollTo=-h1jMnS-v5FC)

optuna/distributions.py
```diff
@@ -121,6 +127,13 @@ def single(self) -> bool:
 
         return self.low == self.high
 
+    def isinclude(self, value):
+        # type: (Any) -> bool
+        if self.low == self.high:
+            return value == self.low
+        else:
+            return self.low <= value and value < self.high
```
nzw0301, line 135: You can use shorter expression: `self.low <= value < self.high`

examples/add_trial_sample.ipynb
```diff
@@ -0,0 +1,423 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def objective(trial):\n",
+    "    x = trial.suggest_int('x', 0, 10)\n",
+    "    y = trial.suggest_uniform('y', 0, 10)\n",
+    "    z = trial.suggest_categorical('z',[0,1])\n",
```
nzw0301, line 23: `z = trial.suggest_categorical('z', [0, 1])`

examples/add_trial_sample.ipynb
```diff
@@ -0,0 +1,423 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#1. modify distributions\n",
+    "modify_distributions = {\n",
+    "    \"y\": optuna.distributions.UniformDistribution(4,6),\n",
```
nzw0301, line 40: ditto

examples/add_trial_sample.ipynb
```diff
@@ -0,0 +1,423 @@
...
+    "    \"z\": optuna.distributions.CategoricalDistribution([0]),\n",
+    "    \"w\": optuna.distributions.DiscreteUniformDistribution(-2.0,2.0,0.2),\n",
```
nzw0301, line 42: ditto

examples/add_trial_sample.ipynb
```diff
@@ -0,0 +1,423 @@
...
+    "}\n",
+    "\n",
+    "def new_objective(trial):\n",
+    "    x = trial.suggest_int('x', 0, 10)\n",
+    "    y = trial.suggest_uniform('y', 4, 6)\n",
+    "    z = trial.suggest_categorical('z',[0])\n",
```
nzw0301, line 48: ditto

examples/add_trial_sample.ipynb
```diff
@@ -0,0 +1,423 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#2. add parameters\n",
+    "add_distributions_params = {\n",
+    "    \"add_param1\": [optuna.distributions.UniformDistribution(-2,2),0],\n",
```
nzw0301, line 140: ditto

examples/add_trial_sample.ipynb
```diff
@@ -0,0 +1,423 @@
...
+    "    \"add_param2\": [optuna.distributions.CategoricalDistribution([False,True]),False],\n",
```
nzw0301, line 141: ditto

examples/add_trial_sample.ipynb
```diff
@@ -0,0 +1,423 @@
...
+    "}\n",
+    "\n",
+    "\n",
+    "def new_objective(trial):\n",
+    "    x = trial.suggest_int('x', 0, 10)\n",
+    "    y = trial.suggest_uniform('y', 4, 6)\n",
+    "    z = trial.suggest_categorical('z',[0])\n",
+    "    w = trial.suggest_discrete_uniform('w', -2.0, 2.0, 0.2)\n",
+    "    add_param1 = trial.suggest_uniform('add_param1', -2, 2)\n",
+    "    add_param2 = trial.suggest_categorical('add_param2',[False, True])\n",
```
nzw0301, line 151: ditto

examples/add_trial_sample.ipynb
```diff
@@ -0,0 +1,423 @@
...
+   "outputs": [],
+   "source": [
+    "# add parameters and modify distributions in objective\n",
+    "def new_objective(trial):\n",
+    "    x = trial.suggest_int('x', 0, 10)\n",
+    "    y = trial.suggest_uniform('y', 4, 6)\n",
+    "    z = trial.suggest_categorical('z',[0])\n",
+    "    w = trial.suggest_discrete_uniform('w', -2.0, 2.0, 0.2)\n",
+    "    add_param1 = trial.suggest_uniform('add_param1', -2, 2)\n",
+    "    add_param2 = trial.suggest_categorical('add_param2',[False, True])\n",
```
nzw0301, line 224: ditto

PR #1032, opened 2020-03-16
Add `study` argument to `optuna.integration.lightgbm.LightGBMTuner`.

toshihikoyanase:
This PR enables users to save LightGBMTuer's results in a `Study` instance. This is just a prototypical implementation and we need a discussion about the design and interface.

## Update March 27, 2020.

- `optuna.integration.lightgbm.LightGBMTuner` will be exposed to advanced users who want to use the pruning, resume, parallelization features.
- The following arguments of `optuna.integration.lightgbm.train` will be deprecated:
  - best_params
  - tuning_history

  (The uses will use `optuna.integration.lightgbm.LightGBMTuner` directly if they want to access such information. )

## Topics (updated March 19, 2020.)
- Can we use a single `Study` instance for all steps?
  - Yes, but a mechanism to distinguish steps is required in order to avoid contamination of trials.
  - In this PR, I add `StepwiseStudy` for this purpose.
- `LightGBMTuner` has two steps for `feature_fraction` tuning. Can we reuse the results in the first step?
  - The current step-wise algorithm employs the grid search for both steps, and the search results do not depend on the search history.
  - On the other hand, we need to separate them if we plan to change the search algorithms.
  - In this PR, I separate the search history using `StepwiseStudy`. It is a subclass of `Study` and overrides `Study.trials` and `Study.best_trial`. In those methods. `StepwiseStudy` chooses the target trials based on `step_name` in advance. `Samplers` and `Pruners` basically use the methods to access search history, so we do not have to modify them to use them with `LightGBMTuner`.
  - If you use the same `step_name` in both of the steps, you can share the search history in the second step.
- In this PR, users create a `Study` instance and pass it to `lgb.train`. In this case, the users need to set study direction while the current `LightGBM Tuner` sets the direction automatically. Is it acceptable?
  - I think it's acceptable if we check the inconsistency of the direction.
  - I add the check to the `LightGBMTuner.__init__`.

## Findings through https://github.com/optuna/optuna/pull/1039

I'd like to enable for `LightGBMTuner` to support the following features of Optuna in the future:

- Pruning
- Resume study using RDBStorage
- Parallelization

I create a prototype at https://github.com/optuna/optuna/pull/1039 based on the implementation of this PR (i.e., `StepwiseStudy`). In summary, I think we can support those features based on this PR with reasonable development costs.

### Pruning

I have a few changes because `StepwiseStudy` hide the trials in the other steps. I just add `LightGBMPruningCallback` to the `callbacks` argument of `lgb.train` and simply pass a pruner as an argument of `Study.__init__` as usual.

### Resume study using RDBStorage

I need to store state in `LightGBMTuner` instances and `OptunaObjective` instances to the storage. Among them, I want to discuss how to store the LightGBM boosters. Currently, I dumped them into files. If you have any ideas, please let me know.

In addition, I replaced `_GridSamplerUniform1D` with `optuna.samplers.GridSampler`. `_GridSamplerUniform1D` does not persistent the status and will start from scratch if we resume the study.

Remaining issues:
- GridSampler seems to ignore pruned trials, and we need to update it.
- Maybe users want to tune the pruner for each step.

### Parallelization

I think we have some choice to parallelize `LightGBMTuner`. Among them, I investigated just a simple case. I created a study instance and executed `lgb.train` with it in two processes (see [example](https://gist.github.com/toshihikoyanase/6253fe69f055e69d299ea5d4042331bd#file-lightgbm_tuner_parallel-py)).

In this case, I just ignored `ValueError` raised by `GridSampler` when it visited all candidates, and then it worked. But I still have some issues as follows:
- NFS is required to share the trained boosters,
- joblib's serializer cannot dump LightGBM boosters, and I have to create a wrapper function that does not return the boosters. In this case, it is not intuitive to obtain the best booster.
- I didn't check what happens if users use the `n_jobs` argument of `Study`.

TODO
- [x] Use `step_name` instead of `action`
- [x] Add tests for `StepwiseStudy`


hvy:
How about updating the docs of `lightgbm_tuner.train`, regarding the `study` parameter?

toshihikoyanase:
> How about updating the docs of lightgbm_tuner.train, regarding the study parameter?

@hvy Thank you for you suggestion. I agree with you.
I added a description of the `study` argument. This `train` method also has the other argument like `best_params` and `tuning_history`, but they will be removed in future PRs. So, I only added the `study` argument.

image

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -290,6 +291,10 @@ def __call__(self, trial):
             )
         )
 
+        trial.set_user_attr("elapsed_secs", elapsed_secs)
+        trial.set_user_attr("average_iteration_time", average_iteration_time)
```
hvy: Just wondering but are the above two attributes used somewhere, or are they there for the user? If so, how about documenting them?

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -276,12 +277,12 @@ def __call__(self, trial):
...
 
         if self.pbar is not None:
-            self.pbar.set_description(pbar_fmt.format(self.action, self.best_score))
+            self.pbar.set_description(pbar_fmt.format(self.step_name, self.best_score))
             self.pbar.update(1)
 
         self.report.append(
             dict(
-                action=self.action,
+                action=self.step_name,
```
hvy: How about leaving a comment that the key name differs from a historical reason?

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -533,10 +567,61 @@ def tune_params(self, target_param_names, n_trials, sampler):
...
+            updated_params = {p: best_trial.params[p] for p in target_param_names}
             self.lgbm_params.update(updated_params)
             self.best_params.update(updated_params)
+
+    def _create_stepwise_study(
+        self, study: "optuna.study.Study", step_name: str
+    ) -> "optuna.study.Study":
+
+        # This class is assumed to be passed to a sampler and a pruner corresponding to the step.
+        class _StepwiseStudy(optuna.study.Study):
```
hvy: I'm wondering whether a user wouldn't want to use this class (publicly) to diagnose the study. 🤔

hvy: On a second thought, maybe it's wiser to keep it private reducing complexity.

toshihikoyanase: > reducing complexity

I agree. I feel `_StepwiseStudy` is too specific for most of the users, and want to keep it private like `_BracketStudy` in HyperbandPruner.
https://github.com/optuna/optuna/blob/master/optuna/pruners/hyperband.py#L144

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: That makes sense. I added the description about them in `LightGBMTuner` docstring. To add the `LightGBMTuner` to the document,  the dependency checking in `optuna.integration.lightgbm` seems to be required. So, I'd like to separate the PRs to add the `LightGBMTuner` entry in the reference.

image

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: I add it. Thanks. Please note that I plan to remove the `self.report` and use `study` to retrieve the results in the future PRs, and then the comment will be removed too.
https://github.com/optuna/optuna/pull/1032/commits/ef51176e701105e99c6068f6d28daeb6dfc5d953#diff-0d8b1d930818af2b13939666bb2df3eaR285-R286

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -290,13 +293,37 @@ def __call__(self, trial):
...
+            A dictionary to store the best parameters.
+        tuning_history:
+            A List to store the history of parameter tuning.
+        study:
+            A :class:`~optuna.study.Study` instance to store optimization results. The
+            :class:`~optuna.trial.Trial` instances in it has the following user attributes:
+            ``elapsed_secs`` is the elapsed time since the optimization starts.
+            ``average_iteration_time`` is the average time of iteration to train the booster
+            model in the trial.
+    """
```
hvy: How about documenting `__init__` in `__init__`, following this recent docstring rule https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#docstrings ?

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -319,8 +346,11 @@ def __init__(
         sample_size=None,  # type: Optional[int]
         best_params=None,  # type: Optional[Dict[str, Any]]
         tuning_history=None,  # type: Optional[List[Dict[str, Any]]]
```
hvy: Do you think it's too much to in this PR raise a deprecation warning if this argument is passed, recommending the usage of the new `study` parameter?

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -319,8 +346,11 @@ def __init__(
         sample_size=None,  # type: Optional[int]
         best_params=None,  # type: Optional[Dict[str, Any]]
```
hvy: Not a strong opinion but maybe we can deprecate this as well. You can seemingly get the params via `lightgbm.basic.Booster.params` (key-value dictionary).

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -355,6 +385,28 @@ def __init__(
...
+            )
+        else:
+            self.study = study
+
+        if self.higher_is_better():
+            if self.study.direction != optuna.structs.StudyDirection.MAXIMIZE:
+                metric_name = self.lgbm_params.get("metric", "binary_logloss")
+                raise ValueError(
+                    "Study direction is inconsistent with the metric {}. "
+                    "Please set 'maximize' to the direction.".format(metric_name)
```
hvy: A nit.
```suggestion
                    "Please set 'maximize' as the direction.".format(metric_name)
```

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -355,6 +385,28 @@ def __init__(
...
+                )
+        else:
+            if self.study.direction != optuna.structs.StudyDirection.MINIMIZE:
+                metric_name = self.lgbm_params.get("metric", "binary_logloss")
+                raise ValueError(
+                    "Study direction is inconsistent with the metric {}. "
+                    "Please set 'minimize' to the direction.".format(metric_name)
```
hvy: Ditto
```suggestion
                    "Please set 'minimize' as the direction.".format(metric_name)
```

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -533,10 +591,61 @@ def tune_params(self, target_param_names, n_trials, sampler):
...
+        assert best_trial.value is not None
+
+        if self.compare_validation_metrics(best_trial.value, self.best_score):
+            self.best_score = best_trial.value
             self.best_booster = objective.best_booster
 
-            updated_params = {p: study.best_trial.params[p] for p in target_param_names}
+            updated_params = {p: best_trial.params[p] for p in target_param_names}
             self.lgbm_params.update(updated_params)
             self.best_params.update(updated_params)
```
hvy: Can we simplify this logic inside this `if`, making some of these `self.*` attribute `@property`s querying `self.study` (and `self.best_booster` for the `self.best_params`)? Not a strong opinion as as I still think we need this check of the booster is to be updated. But I otherwise believe it reduces code complexity.

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: Thank you for your suggestion. I  added the deprecation warning in https://github.com/optuna/optuna/pull/1032/commits/894a666c774e2b27f1811efad1e853c6add77cb3.

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: I also deprecated the `tuning_history` argument in https://github.com/optuna/optuna/pull/1032/commits/894a666c774e2b27f1811efad1e853c6add77cb3. Thanks!

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: Thanks! I fixed it in https://github.com/optuna/optuna/pull/1032/commits/894a666c774e2b27f1811efad1e853c6add77cb3.

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: I also fixed it in 894a666. Thanks!

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: Thank you for your suggestion. I introduced `self.best_score` and `self.best_params` as properties, and simplified the logic.

The changes are as follows:
- I found that `objective.best_booster` is not `None` only if the trial outperformed the past best trial, and check it instead of calling `self.compare_validation_metrics`. I'd like to update this logic when LightGBMTuner supports the parallelization feature because the best booster can be updated by the other workers.
- The update of `self.lgbm_params` is omitted because it is updated in`self.tune_params` right before the training.
- `self.best_params` was renamed to `self._best_params`. The main logic is in the new `self.best_params` property and `self._best_params` is only used to keep backward compatibility.

toshihikoyanase:
@hvy Thank you for your careful review. PTAL.

optuna/integration/lightgbm_tuner/optimize.py
hvy: Thanks!

toshihikoyanase:
I'm sorry but the unit tests for `LightGBMTuner.best_score` and `LightTuner.best_params` were missing. I added them in https://github.com/optuna/optuna/pull/1032/commits/77641967b29190d6a8a072d8e1aff38f6a2f93a6 and https://github.com/optuna/optuna/pull/1032/commits/7e484362adca1d84db32a39117d0bd026c854282.

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -290,13 +295,48 @@ def __call__(self, trial):
             )
         )
 
+        trial.set_user_attr("elapsed_secs", elapsed_secs)
+        trial.set_user_attr("average_iteration_time", average_iteration_time)
+        trial.set_user_attr("step_name", self.step_name)
+        trial.set_user_attr("lgbm_params", json.dumps(self.lgbm_params))
```
hvy, line 301: Very sorry for my late comment. Should we prefix the keys similar to the CMA-ES sampler (`cma:some_key`) as some make little sense without the context of this LightGBMTuner and are quite general English phrases. What do you think?

hvy:
I guess deprecation warnings will be raised from examples/lightgbm_

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: Thank you for letting me know. I'm not aware of it.
I'll like to add a prefix like `lightgbm_tuner:some_key`. If you have any concerns, please let me know.

toshihikoyanase: Fixed: https://github.com/optuna/optuna/pull/1032/commits/f904eec30e6cff1b29cc619e6f0ce9f8797707b6

toshihikoyanase:
> I guess deprecation warnings will be raised from examples/lightgbm_tuner_simple.

Thank you for your careful review. I just removed deprecated arguments to inhibit warnings because we'll make the history private for the users of `lightgbm_tuner.train`.
Commit: https://github.com/optuna/optuna/pull/1032/commits/c5fbd295217d730a7eb9faa55b17bbf90eb58379

optuna/integration/lightgbm_tuner/optimize.py
hvy: Thanks!!

tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
```diff
@@ -1,8 +1,13 @@
 import contextlib
-
 import mock
```
hvy, line 2: Ops, a tiny.
```suggestion

import mock
```

tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
```diff
@@ -1,8 +1,13 @@
...
 import numpy as np
 import pytest
+from typing import Any
+from typing import Dict
+from typing import Generator
+from typing import List
+from typing import Optional
+import warnings
```
hvy, line 10: Ditto regarding import order.

toshihikoyanase: Thanks! I fixed it in https://github.com/optuna/optuna/pull/1032/commits/7f0adbf47d26ba28422cd2e5e41024473d67efe7.

tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
toshihikoyanase: Thanks! Applied it in https://github.com/optuna/optuna/pull/1032/commits/1ee06dc02d118860795f9d9eb57d2539daa6f88b.

PR #1037, opened 2020-03-17
[WIP] Add simple gaussian process based sampler.

VladSkripniuk:
This is my initial attempt on implementing simple Bayesian Optimization based on Gaussian Process for GSoC 2020. I'll sincerely appreciate your feedback.

I ran the following code for comparison with ```optuna.samplers.tpe.TPESampler```.
```
import numpy as np
from matplotlib import pyplot as plt
from scipy.special import sinc
from sklearn.gaussian_process import GaussianProcessRegressor

import optuna
from optuna.samplers.gpbo import GPBO
from optuna.samplers.tpe import TPESampler

def f(x):
    x = x * 10
    x = x - 2
    return sinc(x) + 0.5*sinc(x-3)

X = np.asarray(np.arange(0, 1, 0.001))
y = f(X)

plt.plot(X, y)
plt.ylim(-2,2)
plt.savefig('debug/original.jpg')

def objective(trial):
    x = trial.suggest_uniform('x', 0, 1)

    return f(x)

sampler = GPBO()
study = optuna.create_study(sampler=sampler, direction='minimize')

sampler = TPESampler()
study = optuna.create_study(sampler=sampler, direction='minimize')
```
```optuna.samplers.gpbo.GPBO``` achieves score ```-0.319180173``` and ```optuna.samplers.tpe.TPESampler``` - ```-0.319178161```.

Objective function looks as follows
image
and here is the progress of GPBO
myimage

### DIFF
--- a/optuna/samplers/gpbo/__init__.py
+++ b/optuna/samplers/gpbo/__init__.py
@@ -1,0 +1,1 @@
+from .sampler import GPBO
--- a/optuna/samplers/gpbo/sampler.py
+++ b/optuna/samplers/gpbo/sampler.py
@@ -1,0 +1,164 @@
+import numpy as np
+from matplotlib import pyplot as plt
+from scipy.stats import norm
+from sklearn.gaussian_process import GaussianProcessRegressor, kernels
+
+import optuna
+from optuna import distributions
+from optuna.samplers.base import BaseSampler
+from optuna.samplers import random
+from optuna import samplers
+from optuna import type_checking
+
+if type_checking.TYPE_CHECKING:
+    from typing import Any  # NOQA
+    from typing import Dict  # NOQA
+    from typing import Optional  # NOQA
+
+    from optuna.distributions import BaseDistribution  # NOQA
+    from optuna.structs import FrozenTrial  # NOQA
+    from optuna.study import Study  # NOQA
+
+from optuna import structs
+from optuna.structs import StudyDirection
+
+class GPBO(BaseSampler):
+    """Simple Bayesian Optimization based on Gaussian Process sampler.
+
+    Example:
+
+        .. code::
+
+            >>> study = optuna.create_study(sampler=GPBO())
+            >>> study.optimize(objective, direction='minimize')
+    """
+
+    def __init__(self):
+        # type: () -> None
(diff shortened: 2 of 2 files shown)


Crissman:
Thanks for the PR! We're looking forward to your application for GSoC.

We're discussing this PR internally and will get back to you soon.

g-votte:
@VladSkripniuk 
Thanks for your sharing the idea! I think this is a good starting point to discuss the new sampler.

If we'll go with this implementation, I think one of action items is to get rid of the dependency to sklearn, since our target is to provide a GP-BO sampler without heavy dependencies. 

Also, we'd need the following discussion:
- how to support non-uniform distributions;
- how to support customizable kernels and acquisition functions.

If you have any idea, we'd very much like to discuss some specific implementation plans.

VladSkripniuk:
@Crissman and @g-votte thank you for getting back to me!

After reading a number of papers, including the ones mentioned on wiki, I think ["Bayesian Optimization for Categorical and Category-Specific Continuous Inputs"](https://arxiv.org/pdf/1911.12473.pdf) is the most interesting for me. Proposed Bandit-BO method shows superior performance on several problems with continuous and categorical variables. Authors deal with categorical variables by training separate GPs for each category. Let me know if this paper is relevant for optuna.

To support non-uniform distributions we can create a function similar to [`optuna.samplers.tpe.TPESampler._sample_numerical`](https://github.com/optuna/optuna/blob/3075a1cf6648b3a8f061f904177734a08bb3a3c3/optuna/samplers/tpe/sampler.py#L226). Integer parameters will be supported by rounding and loguniform distributed parameters will be supported by applying log. Categorical parameters are handled by the method proposed in the paper.

I plan to pass customizable kernels and acquisition functions as an argument in the `__init__` method of the new sampler.

I'm working on the proposal, which I wanted to send to optuna@preferred.jp to get some feedback and discuss schedule of deliverables.

g-votte:
Closing the PR, since our motivation is to provide a sampler without heavy dependencies. 

However, I think this sklearn wrapper is a good example for ppl to implement a user-defined sampler.

Thanks a lot for your proposal!

PR #951, opened 2020-02-21
Add a n_startup_trials option to SkoptSampler.

c-bata:
Provide a `n_startup_trials` option like CmaEsSampler.

PR #1043, opened 2020-03-20
Fix SkoptSampler's n_initial_points problem

grassking100:
The n_initial_points in SkoptSampler's skopt_kwargs could cause trouble when n_starup_trials is smaller than n_initial_points  default value. So I enforce the n_initial_points be set to zero and fix some tests. The issue is mentioned at https://github.com/optuna/optuna/pull/951#issuecomment-601832976.

### DIFF
--- a/optuna/integration/skopt.py
+++ b/optuna/integration/skopt.py
@@ -105,8 +105,9 @@ class SkoptSampler(BaseSampler):
             n_initial_points = self._skopt_kwargs["n_initial_points"]
             if n_initial_points > n_startup_trials:
                 raise ValueError(
-                    "The n_initial_points should smaller or equal to n_startup_trials to get expect"
-                    "behavior, but the value is {}.".format(self._skopt_kwargs["n_initial_points"])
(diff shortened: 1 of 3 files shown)


hvy:
Let me share some consensus from a discussion involving @c-bata. 

The consensus was to keep the existing logic but document this in `SkoptSampler` as a potential pitfall. The reason being that the existing API is quite clear about how `skopt(.Optimizer)` is used internally. Thus, it's better to leave it to the user to correctly configure this sampler and thus also `skopt`, instead of silently modify arguments/raise. What do you think?

grassking100:
Hi @hvy:
    Thank you for your reply. Yes, I agree with you that we should document it instead of modifying it directly.

hvy:
@grassking100 Should I close this PR and address the documentation in a separate PR or would you like to tackle the documentation, either here in in a new PR?

grassking100:
I would close the PR and add create a new PR to address the documentation.

PR #1057, opened 2020-03-28
[DO NOT MERGE] Add darglint check in circleci

keisuke-umezawa:
This PR for introcuding [darglint](https://github.com/terrencepreilly/darglint) to check that docstrings are compatible with function definitions.

In the current configuration, it will check as follows:

> - 000: Syntax, formatting, and style
> - 100: Args section
> - 200: Returns section
> - 500: Variables section

Also, we can choose strictness, and current strictness is the most mild.

> long: One-line descriptions and descriptions without arguments/returns/yields/etc. sections will be allowed. Anything more, and the docstring will be fully checked.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,12 @@ jobs:
             . venv/bin/activate
             mypy .
 
+      - run:
+          name: darglint
+          command: |
+            . venv/bin/activate
+            darglint optuna
+
   document:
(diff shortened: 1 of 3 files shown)


keisuke-umezawa:
Sorry, I tried to introduce `darglint`, but it is unstable. Every time I run the command, results are changed. So, I gave it up.

PR #1068, opened 2020-03-30
GitHub Actions to automatically label PRs.

hvy:
Introduces GitHub Actions to automatically label PRs. This hopefully makes the list of PRs somewhat more comprehensible. Note that new GitHub labels are suggested in this PR which are also open for comments.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,17 @@
+optuna.integration:
+  - optuna/integration/**/*
+
+optuna.pruners:
+  - optuna/pruners/**/*
+
+optuna.samplers:
+  - optuna/samplers/**/*
+
+optuna.storages:
+  - optuna/storages/**/*
+
+optuna.testing:
+  - optuna/testing/**/*
+
+optuna.visualization:
+  - optuna/visualization/**/*
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -15,3 +15,9 @@ optuna.testing:
 
 optuna.visualization:
   - optuna/visualization/**/*
+
+optuna.study:
+  - optuna/study.py
+
+optuna.trial:
+  - optuna/trial.py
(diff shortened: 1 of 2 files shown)


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,12 @@
+name: labeler
```
sile: I think that it's good to leave a link to https://github.com/actions/labeler as a comment for future readers who don't know much about GitHub Actions.

hvy: Thanks for the tip. Added!

.github/labeler.yml
```diff
@@ -0,0 +1,17 @@
...
+optuna.samplers:
+  - optuna/samplers/**/*
+
+optuna.storages:
+  - optuna/storages/**/*
+
+optuna.testing:
+  - optuna/testing/**/*
+
+optuna.visualization:
```
sile: How about adding study.py and trial.py as the labeling targets?

hvy: Fixed!

hvy:
PTAL.

hvy:
Created required labels (I'll fill in their descriptions once this PR is merged).

.github/labeler.yml
```diff
@@ -0,0 +1,23 @@
...
+  - optuna/visualization/**/*
+
+optuna.study:
+  - optuna/study.py
+
+optuna.trial:
+  - optuna/trial.py
```
toshihikoyanase, line 23: How about adding `optuna.importance` label for the hyperparameter importance feature introduced by #946?

hvy: Thanks for your inputs. I considered that as well but omitted it for now as its experimental, but it's not a strong opinion, let me add it.

hvy: Added!

hvy:
PTAL.

hvy:
Restart CI as `master` was fixed.

PR #952, opened 2020-02-22
Implement some methods almost compatible with Scikit-learn private methods

himkt:
This PR is follow-up for #881.

I implement some methods to reduce dependencies on scikit-learn private methods.
I basically define methods to be compatible with scikit-learn's, but some points are different.

(The name of this branch should be `sklearn-privates`...:innocent:)

PR #935, opened 2020-02-17
TensorFlow estimator integration example with only v2 APIs

nuka137:
This PR fixes TensorFlow estimator integration example to use v2 APIs. (c.f. #924 )
This PR also fixes name of variables which mentioned in [here](https://github.com/optuna/optuna/pull/924#pullrequestreview-358708888).

Related Issue: #930

PR #1070, opened 2020-03-31
Fix docstring on optuna/integration/*.py

nuka137:
This PR fixes docstring on optuna/integration/*.py. (c.f. https://github.com/optuna/optuna/issues/734#issuecomment-603645615)

Related Issue: #734

### DIFF
--- a/optuna/integration/chainer.py
+++ b/optuna/integration/chainer.py
@@ -28,7 +28,7 @@ class ChainerPruningExtension(Extension):
 
     See `the example <https://github.com/optuna/optuna/blob/master/
     examples/pruning/chainer_integration.py>`__
-    if you want to add a pruning extension which the observes validation
+    if you want to add a pruning extension which observes validation
     accuracy of a `Chainer Trainer <https://docs.chainer.org/en/stable/
     reference/generated/chainer.training.Trainer.html>`_.
 
--- a/optuna/integration/chainer.py
+++ b/optuna/integration/chainer.py
@@ -28,8 +28,8 @@ class ChainerPruningExtension(Extension):
 
     Example:
 
-        See `the example <https://github.com/optuna/optuna/blob/30dba8814d332ff4
-        0d9c7465ee660a8ecf499547/examples/pruning/chainer_integration.py>`__
+        See `the example <https://github.com/optuna/optuna/blob/master/
+        examples/pruning/chainer_integration.py>`__
         if you want to add a pruning extension which observes validation accuracy to
         `Chainer Trainer <https://docs.chainer.org/en/stable/reference/generated/
         chainer.training.Trainer.html>`_.
--- a/optuna/integration/chainer.py
+++ b/optuna/integration/chainer.py
@@ -26,13 +26,11 @@ if type_checking.TYPE_CHECKING:
 class ChainerPruningExtension(Extension):
     """Chainer extension to prune unpromising trials.
 
-    Example:
-
-        See `the example <https://github.com/optuna/optuna/blob/master/
-        examples/pruning/chainer_integration.py>`__
-        if you want to add a pruning extension which observes validation accuracy to
-        `Chainer Trainer <https://docs.chainer.org/en/stable/reference/generated/
-        chainer.training.Trainer.html>`_.
+    See `the example <https://github.com/optuna/optuna/blob/master/
+    examples/pruning/chainer_integration.py>`__
+    if you want to add a pruning extension which the observes validation
+    accuracy of a `Chainer Trainer <https://docs.chainer.org/en/stable/
+    reference/generated/chainer.training.Trainer.html>`_.
 
     Args:
         trial:
--- a/optuna/integration/chainer.py
+++ b/optuna/integration/chainer.py
@@ -29,7 +29,7 @@ class ChainerPruningExtension(Extension):
     Example:
 
         See `the example <https://github.com/optuna/optuna/blob/30dba8814d332ff4
-        0d9c7465ee660a8ecf499547/examples/pruning/chainer_integration.py>`_
+        0d9c7465ee660a8ecf499547/examples/pruning/chainer_integration.py>`__
         if you want to add a pruning extension which observes validation accuracy to
         `Chainer Trainer <https://docs.chainer.org/en/stable/reference/generated/
         chainer.training.Trainer.html>`_.
--- a/optuna/integration/chainer.py
+++ b/optuna/integration/chainer.py
@@ -28,14 +28,11 @@ class ChainerPruningExtension(Extension):
 
     Example:
 
-        Add a pruning extension which observes validation losses to
+        See `the example <https://github.com/optuna/optuna/blob/30dba8814d332ff4
+        0d9c7465ee660a8ecf499547/examples/pruning/chainer_integration.py>`_
+        if you want to add a pruning extension which observes validation accuracy to
         `Chainer Trainer <https://docs.chainer.org/en/stable/reference/generated/
         chainer.training.Trainer.html>`_.
-
-        .. code::
-
-            trainer.extend(
-                ChainerPruningExtension(trial, 'validation/main/loss', (1, 'epoch')))
 
     Args:
         trial:
--- a/optuna/integration/fastai.py
+++ b/optuna/integration/fastai.py
@@ -23,8 +23,7 @@ class FastAIPruningCallback(TrackerCallback):
 
     See `the example <https://github.com/optuna/optuna/blob/master/
     examples/fastai_simple.py>`__
-    if you want to add a pruning callback which monitors validation loss directly
-    to ``Learner``.
+    if you want to add a pruning callback which monitors validation loss of a ``Learner``.
 
     Example:
 
--- a/optuna/integration/fastai.py
+++ b/optuna/integration/fastai.py
@@ -23,8 +23,8 @@ class FastAIPruningCallback(TrackerCallback):
 
     Example:
 
-        See `the example <https://github.com/optuna/optuna/blob/30dba8814d332ff4
-        0d9c7465ee660a8ecf499547/examples/fastai_simple.py>`__
+        See `the example <https://github.com/optuna/optuna/blob/master/
+        examples/fastai_simple.py>`__
         if you want to add a pruning callback which monitors validation loss directly
         to ``Learner``.
 
--- a/optuna/integration/fastai.py
+++ b/optuna/integration/fastai.py
@@ -23,16 +23,10 @@ class FastAIPruningCallback(TrackerCallback):
 
     Example:
 
-        Add a pruning callback which monitors validation loss directly to ``Learner``.
-
-        .. code::
-
-            # If registering this callback in construction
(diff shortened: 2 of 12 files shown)


optuna/integration/sklearn.py
```diff
@@ -511,21 +511,24 @@ class OptunaSearchCV(BaseEstimator):
...
+                clf = SVC(gamma='auto')
+                param_distributions = {
+                    'C': optuna.distributions.LogUniformDistribution(1e-10, 1e+10)
+                }
+                optuna_search = optuna.integration.OptunaSearchCV(
+                    clf,
+                    param_distributions
+                )
+                X, y = load_iris(return_X_y=True)
+                optuna_search.fit(X, y) # doctest: +ELLIPSIS
```
himkt, line 530: Could you please tell me why `+ELLIPSIS` is needed here?

toshihikoyanase:
Thank you for your constant contribution!
It basically looks good to me except [this question](https://github.com/optuna/optuna/pull/1070#discussion_r403814167).
FYI: The rendered document can be found [here](https://33772-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/integration.html).

optuna/integration/sklearn.py
nuka137: Thanks.
This `+ELLIPSIS` option comes from original code for doctest.
I think we can delete it because this option does not make sense any more in our test.
i.e. We don't check output of `optuna_search.fit` in this test.

himkt: It makes sense, thank you for clarification. :+1:

optuna/integration/chainer.py
```diff
@@ -28,15 +28,12 @@ class ChainerPruningExtension(Extension):
 
     Example:
 
-        Add a pruning extension which observes validation losses to
+        See `the example <https://github.com/optuna/optuna/blob/30dba8814d332ff4
+        0d9c7465ee660a8ecf499547/examples/pruning/chainer_integration.py>`__
```
hvy, line 32: I believe it's more user friendly to reference the latest example and not a particular snapshot. This documentation must be maintained in accordance to the example scrip in any way. This is not a strong opinion though as I see that this is consistent with earlier similar changes, https://github.com/optuna/optuna/pull/1019.

nuka137: I agree.
I think it will be much more maintenance costs if we use specific commit in the URL.
We can avoid this by using master branch. (For example, we can see this usage in [TensorFlow Document](https://github.com/tensorflow/docs/blob/master/site/en/guide/profiler.md#install-the-profiler-and-gpu-prerequisites) .)
However, it also has the disadvantage described [here](https://github.com/optuna/optuna/pull/1019#discussion_r395010625).

I think using master branch in example URL is better because of simple and low maintenance cost.

Could you give me any advises about this?

cc: @himkt

himkt: I wrote the comment based on [the review which I received previously](https://github.com/optuna/optuna/pull/952#discussion_r384866934).
In #952, we have to refer to the implementation which is removed in the master branch.
However, in this PR, it would be better to refer to the implementation of the master branch.
So I agree with @hvy and @nuka137 to remove commit id and refer to the master branch.

@nuka137 I'm really sorry for confusing you.

hvy: Sorry for reraising a sort of settled issue. I believe however that the consensus to here refer to the master is reasonable. @nuka137 could you update the changes in this PR?

nuka137: Thanks!
OK, I will fix them (include examples/pruning/tensorflow_estimator_integration.py as well).

nuka137: @hvy @himkt 

Thanks for your advice!
I fixed all example links to refer master branch.

optuna/integration/tensorflow.py
```diff
@@ -21,8 +21,8 @@ class TensorFlowPruningHook(SessionRunHook):
 
     Example:
 
-        See `the example <https://github.com/optuna/optuna/blob/30dba8814d332ff4
-        0d9c7465ee660a8ecf499547/examples/tensorflow_estimator_simple.py>`_
```
hvy, line 25: Seems like wrong example. Should it not be https://github.com/optuna/optuna/blob/master/examples/pruning/tensorflow_estimator_integration.py ? Might be beyond the scope of this PR but I think we might as well fix it here as it's highly related.

optuna/integration/chainer.py
```diff
@@ -28,15 +28,12 @@ class ChainerPruningExtension(Extension):
 
     Example:
```
hvy, line 29: I think this `Example:` is mostly just noisy and that it can be removed. What do you think?

optuna/integration/chainer.py
```diff
@@ -28,15 +28,12 @@ class ChainerPruningExtension(Extension):
...
 
-        Add a pruning extension which observes validation losses to
+        See `the example <https://github.com/optuna/optuna/blob/master/
+        examples/pruning/chainer_integration.py>`__
+        if you want to add a pruning extension which observes validation accuracy to
         `Chainer Trainer <https://docs.chainer.org/en/stable/reference/generated/
         chainer.training.Trainer.html>`_.
```
hvy, line 35: The grammar can be improved a bit.
```
See the example if you want to add a pruning extension which the observes validation accuracy of a Chainer Trainer.
```
Not a strong opinion but I also believe simplifying it as follows is fine too. You can then be consistent across all integrations.
```
See example.
```
where `example` is a hyperlink.

optuna/integration/chainer.py
nuka137: I removed `Example:` from docstring, applied same thing to other files.

optuna/integration/tensorflow.py
nuka137: Nice catch! Thanks.
I fixed to point to correct link.

optuna/integration/chainer.py
nuka137: Thanks! I fixed it.

> Not a strong opinion but I also believe simplifying it as follows is fine too

It seems that same discussion was held in https://github.com/optuna/optuna/pull/1019#discussion_r394304394 .
I think it is good for users to show what the example actually does.

hvy: @nuka137, thanks for your fixed. What do you think of my latter suggestion above? Just writing "See example". I think it' less to maintain and there are some grammatical changes I'd like to make otherwise.

hvy: I just saw your reply. Okay, let's keep it as is. Let me leave some comments on the grammar.

optuna/integration/chainer.py
```diff
@@ -26,16 +26,11 @@
...
-        `Chainer Trainer <https://docs.chainer.org/en/stable/reference/generated/
-        chainer.training.Trainer.html>`_.
-
-        .. code::
-
-            trainer.extend(
-                ChainerPruningExtension(trial, 'validation/main/loss', (1, 'epoch')))
+    See `the example <https://github.com/optuna/optuna/blob/master/
+    examples/pruning/chainer_integration.py>`__
+    if you want to add a pruning extension which the observes validation
```
hvy, line 31:
```suggestion
    if you want to add a pruning extension which observes validation
```

optuna/integration/fastai.py
```diff
@@ -21,18 +21,10 @@ class FastAIPruningCallback(TrackerCallback):
...
-            # If registering this callback in construction
-            from functools import partial
-
-            learn = Learner(
-                data, model,
-                callback_fns=[partial(FastAIPruningCallback, trial=trial, monitor='valid_loss')])
+    See `the example <https://github.com/optuna/optuna/blob/master/
+    examples/fastai_simple.py>`__
+    if you want to add a pruning callback which monitors validation loss directly
+    to ``Learner``.
```
hvy, line 27:
```suggestion
    if you want to add a pruning callback which monitors validation loss of a ``Learner``.
```

optuna/integration/xgboost.py
```diff
@@ -29,16 +29,10 @@ def _get_callback_context(env):
...
-        .. code::
-
-                pruning_callback = XGBoostPruningCallback(trial, 'validation-error')
-                bst = xgb.train(param, dtrain, evals=[(dtest, 'validation')],
-                                callbacks=[pruning_callback])
-
+    See `the example <https://github.com/optuna/optuna/blob/master/
+    examples/pruning/xgboost_integration.py>`__
+    if you want to add a pruning callback which observes validation AUC to
+    training of an XGBoost model.
```
hvy, line 35:
```suggestion
    if you want to add a pruning callback which observes validation AUC of
    a XGBoost model.
```

optuna/integration/tfkeras.py
```diff
@@ -22,13 +22,9 @@ class TFKerasPruningCallback(Callback):
...
-    Example:
-
-        Add a pruning callback which observes validation losses.
-
-        .. code::
-
-            model.fit(x, y, callbacks=[TFKerasPruningCallback(trial, 'val_loss')])
+    See `the example <https://github.com/optuna/optuna/blob/master/
+    examples/pruning/tfkeras_integration.py>`__
+    if you want to add a pruning callback which observes validation accuracy.
```
hvy, line 27:
```suggestion
    if you want to add a pruning callback which observes the validation accuracy.
```

optuna/integration/tensorflow.py
```diff
@@ -19,11 +19,9 @@
...
     """TensorFlow SessionRunHook to prune unpromising trials.
 
-    Example:
-
-        See `the example <https://github.com/optuna/optuna/blob/30dba8814d332ff4
-        0d9c7465ee660a8ecf499547/examples/tensorflow_estimator_simple.py>`_
-        if you want to add a pruning SessionRunHook for TensorFlow's Estimator.
+    See `the example <https://github.com/optuna/optuna/blob/master/examples/
+    pruning/tensorflow_estimator_integration.py>`_
+    if you want to add a pruning SessionRunHook for TensorFlow's Estimator.
```
hvy, line 24:
```suggestion
    if you want to add a pruning hook to TensorFlow's estimator.
```
C.f. https://github.com/optuna/optuna/pull/935

optuna/integration/lightgbm.py
```diff
@@ -39,15 +39,10 @@
...
-        Add a pruning callback which observes validation scores to training of a LightGBM model.
-
-        .. code::
-
-                param = {'objective': 'binary', 'metric': 'binary_error'}
-                pruning_callback = LightGBMPruningCallback(trial, 'binary_error')
-                gbm = lgb.train(param, dtrain, valid_sets=[dtest], callbacks=[pruning_callback])
+    See `the example <https://github.com/optuna/optuna/blob/master/
+    examples/pruning/lightgbm_integration.py>`__
+    if you want to add a pruning callback which observes AUC to training
```
hvy, line 44:
```suggestion
    if you want to add a pruning callback which observes AUC
```

nuka137:
@hvy 

Thanks for your suggestions! Looks good to me.
I applied all suggested commits.

PR #1074, opened 2020-04-02
Apply lazy import for `optuna.dashboard`.

### LINKED ISSUE #589: Apply lazy import for `optuna.dashboard`.
Optuna always imports the dependencies of `optuna.dashboard` (e.g., `bokeh`), which makes unnecessary overhead in many use cases. Similar to #334, we can apply lazy import for them.

ytsmiling:
This PR apply lazy-import for `optuna.dashboard`.

This PR resolves #589.

### DIFF
--- a/optuna/__init__.py
+++ b/optuna/__init__.py
@@ -40,7 +40,7 @@ else:
         access and the imported submodule is directly accessed from the second access.
 
         Args:
-            name: name of module to apply lazy import
+            name: Name of module to apply lazy import.
         """
 
         def __init__(self, name: str) -> None:
--- a/optuna/__init__.py
+++ b/optuna/__init__.py
@@ -40,6 +40,7 @@ else:
         Args:
             name: name of module to apply lazy import
         """
+
         def __init__(self, name: str) -> None:
             super(_LazyImport, self).__init__(name)
             self._name = name
--- a/optuna/__init__.py
+++ b/optuna/__init__.py
@@ -36,6 +36,8 @@ else:
         This class wrap specified module and lazily import it when they are actually accessed.
         Otherwise, `import optuna` becomes slower because it imports all submodules and
         their dependencies (e.g., bokeh) all at once.
+        Within this project's usage, importlib override this module's attribute on the first
+        access and the imported submodule is directly accessed from the second access.
 
         Args:
             name: name of module to apply lazy import
@@ -47,7 +49,6 @@ else:
 
         def _load(self) -> types.ModuleType:
             module = importlib.import_module(self._name)
-            globals()[self._name] = module
             self.__dict__.update(module.__dict__)
             return module
 
--- a/optuna/__init__.py
+++ b/optuna/__init__.py
@@ -28,7 +28,7 @@ from optuna.type_checking import TYPE_CHECKING  # NOQA
 if TYPE_CHECKING:
     from optuna import dashboard  # NOQA
 else:
-    from typing import Any  # NOQA
+    from typing import Any
 
     class _LazyImport(types.ModuleType):
         """Module Wrapper for lazy import.
--- a/optuna/__init__.py
+++ b/optuna/__init__.py
@@ -31,7 +31,7 @@ else:
     from typing import Any
 
     class _LazyImport(types.ModuleType):
-        """Module Wrapper for lazy import.
+        """Module wrapper for lazy import.
 
         This class wraps specified module and lazily import it when they are actually accessed.
         Otherwise, `import optuna` becomes slower because it imports all submodules and
--- a/optuna/__init__.py
+++ b/optuna/__init__.py
@@ -1,4 +1,6 @@
-from optuna import dashboard  # NOQA
+import importlib
+import types
+
 from optuna import distributions  # NOQA
 from optuna import exceptions  # NOQA
 from optuna import importance  # NOQA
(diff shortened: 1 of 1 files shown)


optuna/__init__.py
```diff
@@ -20,3 +22,29 @@
...
+
+
+if TYPE_CHECKING:
+    from optuna import dashboard  # NOQA
+    from typing import Any  # NOQA
+else:
+
+    class _LazyImport(types.ModuleType):
+        def __init__(self, name):
+            # type: (str) -> None
```
g-votte, line 35: As discussed in #711, we've been using the function annotation syntax after dropping Python 2.

ytsmiling: Thank you for your review.
On the way to addressing the problem, I found that this class is not subject to type-checking.
Additionally, I had to import `typing` module even when `optuna.type_checking.TYPE_CHECKING` is False. The same will apply for `optuna.integration`.
Do you have any idea for a workaround, or are we okay with them?

sile: IMO, it's okay.
However, if anyone is concerned about the above points, we might be able to create, for example, `optuna._lazy_import` module and put the `_LazyImport` class under the module.

optuna/__init__.py
```diff
@@ -20,3 +22,29 @@
...
+            super(_LazyImport, self).__init__(name)
+            self._name = name
+
+        def _load(self):
+            # type: () -> types.ModuleType
+            module = importlib.import_module(self._name)
+            globals()[self._name] = module
```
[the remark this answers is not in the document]
ytsmiling: It avoids unnecessary indirect access to the loader class.

Sample Code:
```python
class DummyModule:
    attr = 1


class IndirectLoader:
    def _load(self):
        return DummyModule()

    def __getattr__(self, item):
        print("indirect access 1")
        return getattr(self._load(), item)


class DirectLoader:
    def _load(self):
        m = DummyModule()
        globals()['module2'] = m
        return m

    def __getattr__(self, item):
        print("indirect access 2")
        return getattr(self._load(), item)


print("test indirect")
module1 = IndirectLoader()
module1.attr
module1.attr

print("test direct")
module2 = DirectLoader()
module2.attr
module2.attr
```

Ouput:
```
test indirect
indirect access 1
indirect access 1
test direct
indirect access 2
```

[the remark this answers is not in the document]
sile: Thank you for the explanation. I understood that reason.
But, personally, I prefer to use `setattr` function for that purpose (such as [the lazy loading of integration modules](https://github.com/optuna/optuna/blob/master/optuna/integration/__init__.py#L76)) because it only affects the local state of the loader class (I would like to avoid mutating global states unless it's really necessary).
```python
# An example usage of `setattr`.
class DirectLoader2:
    def _load(self):
        return DummyModule()

    def __getattr__(self, item):
        print("indirect access 3")
        value = getattr(self._load(), item)
        setattr(self, item, value)
        return value

# print("test direct2")
# module3 = DirectLoader2()
# module3.attr
# module3.attr
# 
# Output:
#   test direct2
#   indirect access 3
```

optuna/__init__.py
```diff
@@ -20,3 +22,37 @@
 from optuna.study import Study  # NOQA
 from optuna.trial import Trial  # NOQA
 from optuna.version import __version__  # NOQA
+from optuna.type_checking import TYPE_CHECKING  # NOQA
+
+
+if TYPE_CHECKING:
+    from optuna import dashboard  # NOQA
+else:
+    from typing import Any  # NOQA
```
sile, line 31: I think that `# NOQA` comment isn't required here since `Any` is used at line 55.
```suggestion
    from typing import Any
```

optuna/__init__.py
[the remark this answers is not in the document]
sile: We discussed this offline and decided to remove the use of `globals` function.

optuna/__init__.py
```diff
@@ -20,3 +22,37 @@
...
+
+if TYPE_CHECKING:
+    from optuna import dashboard  # NOQA
+else:
+    from typing import Any
+
+    class _LazyImport(types.ModuleType):
+        """Module Wrapper for lazy import.
+
+        This class wrap specified module and lazily import it when they are actually accessed.
```
g-votte, line 36:
```suggestion
        This class wraps specified module and lazily import it when they are actually accessed.
```

optuna/__init__.py
```diff
@@ -20,3 +22,37 @@
...
+        Otherwise, `import optuna` becomes slower because it imports all submodules and
+        their dependencies (e.g., bokeh) all at once.
+        Within this project's usage, importlib override this module's attribute on the first
+        access and the imported submodule is directly accessed from the second access.
+
+        Args:
+            name: name of module to apply lazy import
```
g-votte, line 43: Just to be consistent with other classes in Optuna, please start the description with a capital and end with a period.
```suggestion
            name: Name of module to apply lazy import.
```

optuna/__init__.py
```diff
@@ -20,3 +22,37 @@
...
+from optuna.type_checking import TYPE_CHECKING  # NOQA
+
+
+if TYPE_CHECKING:
+    from optuna import dashboard  # NOQA
+else:
+    from typing import Any
+
+    class _LazyImport(types.ModuleType):
+        """Module Wrapper for lazy import.
```
g-votte, line 34:
```suggestion
        """Module wrapper for lazy import.
```

PR #1076, opened 2020-04-02
Support parallel execution of `LightGBMTuner`.

toshihikoyanase:
This is a draft PR of the parallelization of LightGBMTuner. In this PR, I also work on the resume feature of LightGBMTuner.

Here's [an example code](https://gist.github.com/toshihikoyanase/83f47966bef0f5d9d147d013742a3e85) to execute LightGBMTuner in parallel.
Please download it and execute `lightgbm_tuner_parallel.py` in two command prompts.

```
python lightgbm_tuner_parallel.py &
python lightgbm_tuner_parallel.py 
```

The result can be dumped as a CSV file using `lightgbm_tuner_dump_result.py` at the same directory.

```
python lightgbm_tuner_dump_result.py
```

## Design choice of `LightGBMTuner.best_booster`.

Currently, a booster corresponding to each trial is not shared with workers. This is because booster can be too large to store in Optuna's storage. I think we have several choices when users call `LightGBMTuner.best_booster` in a worker which does not have the best booster instance.

1. Raise `ValueError` with a message (e.g., In a parallel environment, best_booster is not always available. Please retrain it using the best parameters).
2. Retrain the best booster using the best parameters. This branch employs this approach.
3. Share booster instances using shared file system.

Which one do you prefer?

## Known issue

- `GridSampler` may sample the same parameter values at the same time. If so, it samples all combinations.
- `GridSampler` may leave running trials if it does not have any candidates.
- The parallel execution of LightGBMTuner is not tested.  Do you have any ideas to test it in a parallel environment?

[Done] ~~Please merge after https://github.com/optuna/optuna/pull/1032.~~

### DIFF
--- a/optuna/integration/lightgbm_tuner/__init__.py
+++ b/optuna/integration/lightgbm_tuner/__init__.py
@@ -32,7 +32,7 @@ def train(*args: Any, **kwargs: Any) -> Any:
 
     auto_booster = LightGBMTuner(*args, **kwargs)
     auto_booster.run()
-    return auto_booster.best_booster
+    return auto_booster.get_best_booster()
 
 
 def _check_lightgbm_availability():
--- a/optuna/integration/lightgbm_tuner/__init__.py
+++ b/optuna/integration/lightgbm_tuner/__init__.py
@@ -31,8 +31,8 @@ def train(*args: Any, **kwargs: Any) -> Any:
     _check_lightgbm_availability()
 
     auto_booster = LightGBMTuner(*args, **kwargs)
-    booster = auto_booster.run()
-    return booster
+    auto_booster.run()
+    return auto_booster.best_booster
 
 
 def _check_lightgbm_availability():
--- a/optuna/integration/lightgbm_tuner/optimize.py
+++ b/optuna/integration/lightgbm_tuner/optimize.py
@@ -1,6 +1,8 @@
 import contextlib
 import copy
 import json
+import os
+import pickle
 import time
 from typing import Any
 from typing import Dict
@@ -170,6 +172,7 @@ class OptunaObjective(BaseTuner):
         lgbm_kwargs,  # type: Dict[str, Any]
         best_score,  # type: float
         step_name,  # type: str
+        model_dir,  # type: Optional[str]
         pbar=None,  # type: Optional[tqdm.tqdm]
     ):
 
@@ -184,6 +187,7 @@ class OptunaObjective(BaseTuner):
         self.best_score = best_score
         self.best_booster_with_trial_number = None  # type: Optional[Tuple[lgb.Booster, int]]
         self.step_name = step_name
+        self.model_dir = model_dir
 
         self._check_target_names_supported()
 
@@ -243,6 +247,12 @@ class OptunaObjective(BaseTuner):
         val_score = self._get_booster_best_score(booster)
         elapsed_secs = t.elapsed_secs()
         average_iteration_time = elapsed_secs / booster.current_iteration()
+
+        if self.model_dir is not None:
+            path = os.path.join(self.model_dir, "{}.pkl".format(trial.number))
+            with open(path, "wb") as fout:
+                pickle.dump(booster, fout)
+
         if self.compare_validation_metrics(val_score, self.best_score):
             self.best_score = val_score
             self.best_booster_with_trial_number = (booster, trial.number)
@@ -305,6 +315,12 @@ class LightGBMTuner(BaseTuner):
             ``average_iteration_time`` is the average time of iteration to train the booster
             model in the trial. ``lgbm_params`` is a JSON-serialized dictionary of LightGBM
             parameters used in the trial.
+
+        model_dir:
+            A directory to save boosters. By default, it is set to :obj:`None` and no boosters are
+            saved. Please set shared directory (e.g., directories on NFS) if you want to access
+            :meth:`~optuna.integration.LightGBMTuner.best_booster` in distributed environments.
+            Otherwise, it may raise :obj:`ValueError`.
     """
 
     def __init__(
@@ -329,6 +345,7 @@ class LightGBMTuner(BaseTuner):
         best_params=None,  # type: Optional[Dict[str, Any]]
         tuning_history=None,  # type: Optional[List[Dict[str, Any]]]
         study=None,  # type: Optional[Study]
+        model_dir=None,  # type: Optional[str]
         verbosity=1,  # type: Optional[int]
     ):
         # type: (...) -> None
@@ -359,6 +376,10 @@ class LightGBMTuner(BaseTuner):
         )  # type: Dict[str, Any]
         self._parse_args(*args, **kwargs)
         self._best_booster_with_trial_number = None  # type: Optional[Tuple[lgb.Booster, int]]
+        self._model_dir = model_dir
+
+        if self._model_dir is not None and not os.path.exists(self._model_dir):
+            os.mkdir(self._model_dir)
 
         if best_params is not None:
             warnings.warn(
@@ -432,27 +453,31 @@ class LightGBMTuner(BaseTuner):
         if self._best_booster_with_trial_number is not None:
             if self._best_booster_with_trial_number[1] == self.study.best_trial.number:
                 return self._best_booster_with_trial_number[0]
+        else:
+            if len(self.study.trials) == 0:
+                raise ValueError("The best booster is not available because no trials completed.")
 
-        # No booster is available because the optimization is resumed or run in parallel.
-        # Recreate the best booster using the best parameters.
-        lgbm_params = copy.deepcopy(self.lgbm_params)
-        lgbm_params.update(self.best_params)
+        if self._model_dir is None:
+            raise ValueError(
+                "The best booster cannot be found. It may be found in the other processes due to "
+                "resuming or distributed computing. Please set the `model_dir` argument of "
+                "`LightGBMTuner.__init__` and make sure that boosters are shared with all "
+                "processes."
+            )
 
-        train_set = self.train_set
-        if self.train_subset is not None:
-            train_set = self.train_subset
+        best_trial = self.study.best_trial
+        path = os.path.join(self._model_dir, "{}.pkl".format(best_trial.number))
+        if not os.path.exists(path):
+            raise ValueError(
+                "The best booster cannot be found in {}. If you execute `LightGBMTuner` in "
+                "distributed environment, please use network file system (e.g., NFS) to share "
+                "models with multiple workers.".format(self._model_dir)
+            )
 
-        objective = OptunaObjective(
-            [],
-            lgbm_params,
-            train_set,
-            self.lgbm_kwargs,
-            -np.inf if self.higher_is_better() else np.inf,
-            step_name="best_booster",
-        )
-        optuna.create_study().optimize(objective, n_trials=1)
-        assert objective.best_booster_with_trial_number is not None
-        return objective.best_booster_with_trial_number[0]
+        with open(path, "rb") as fin:
+            booster = pickle.load(fin)
+
+        return booster
 
     def _get_params(self):
         # type: () -> Dict[str, Any]
@@ -485,15 +510,8 @@ class LightGBMTuner(BaseTuner):
         self.train_subset = None  # Use for sampling.
         self.lgbm_kwargs = kwargs
 
-    def run(self):
-        # type: () -> lgb.Booster
-        """Perform the hyperparameter-tuning with given parameters.
-
-        Returns:
-
-            lightgbm.Booster : Booster
-                The trained booster model.
-        """
+    def run(self) -> None:
+        """Perform the hyperparameter-tuning with given parameters."""
         # Surpress log messages.
         if self.auto_options["verbosity"] == 0:
             optuna.logging.disable_default_handler()
@@ -514,29 +532,27 @@ class LightGBMTuner(BaseTuner):
         with _timer() as t:
             self.tune_feature_fraction()
             if time_budget is not None and time_budget < t.elapsed_secs():
-                return self.best_booster
+                return
 
             self.tune_num_leaves()
             if time_budget is not None and time_budget < t.elapsed_secs():
-                return self.best_booster
+                return
 
             self.tune_bagging()
             if time_budget is not None and time_budget < t.elapsed_secs():
-                return self.best_booster
+                return
 
             self.tune_feature_fraction_stage2()
             if time_budget is not None and time_budget < t.elapsed_secs():
-                return self.best_booster
+                return
 
             self.tune_regularization_factors()
             if time_budget is not None and time_budget < t.elapsed_secs():
-                return self.best_booster
+                return
 
             self.tune_min_data_in_leaf()
             if time_budget is not None and time_budget < t.elapsed_secs():
-                return self.best_booster
-
-        return self.best_booster
+                return
 
     def sample_train_set(self):
         # type: () -> None
@@ -621,6 +637,7 @@ class LightGBMTuner(BaseTuner):
             self.lgbm_kwargs,
             self.best_score,
             step_name=step_name,
+            model_dir=self._model_dir,
             pbar=pbar,
         )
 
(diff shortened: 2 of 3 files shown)


optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -598,7 +598,7 @@ def tune_min_data_in_leaf(self):
         # type: () -> None
 
         param_name = "min_child_samples"
-        param_values = [5, 10, 25, 50, 100]
+        param_values = [5, 10, 25, 50, 100]  # type: List[Union[str, float, int, bool, None]]
```
toshihikoyanase, line 601: The type of candidate points of `optuna.samplers.GridSampler` is `List[Union[str, float, int, bool, None]]`, and `mypy` says `List[int]` is not compatible with it. I think `int` is included by `Union[str, float, int, bool, None]`, and I'm not sure why the error occurs. Please let me know if you have any ideas.

toshihikoyanase: I submitted the PR https://github.com/optuna/optuna/pull/1102 to resolve this issue.

toshihikoyanase: The change was reverted in https://github.com/optuna/optuna/pull/1076/commits/56a4ea4b00888f0dd228142fb7bc05f55b9ab45c because #1102 resolved the issue.

toshihikoyanase:
Thank you, @smly! I updated the `LightGBMTuner.best_booster` according to the discussion.
Regarding the example code to demonstrate how to get `best_booster` without model_dir in parallel environments, I felt it too detailed to explain in the reference.
How about adding it as an FAQ entry?

smly:
Thanks for the updates!! Let me review this PR.

> How about adding it as an FAQ entry?

Sounds good to me. :+1:

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -274,9 +247,15 @@ def __call__(self, trial):
         val_score = self._get_booster_best_score(booster)
         elapsed_secs = t.elapsed_secs()
         average_iteration_time = elapsed_secs / booster.current_iteration()
+
+        if self.model_dir is not None:
+            path = os.path.join(self.model_dir, "{}.pkl".format(trial.number))
+            with open(path, "wb") as fout:
+                pickle.dump(booster, fout)
```
smly, line 254: (Comment only): I support the use of `pickle.dump` instead of `booster.save_model`.
`pickle.load` may cause problems when the versions of LightGBM in the distributed and development env are different. At least, in my trials, there have been no major problems with the last two versions.
Also, `booster.save_model` does not save many attributes of the booster. It leads to some inconvenience in practical usage.

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -336,6 +315,12 @@ class LightGBMTuner(BaseTuner):
             ``average_iteration_time`` is the average time of iteration to train the booster
             model in the trial. ``lgbm_params`` is a JSON-serialized dictionary of LightGBM
             parameters used in the trial.
+
+        model_dir:
+            A directory to save boosters. By default, it is set to :obj:`None` and no boosters are
+            saved. Please set shared directory (e.g., directories on NFS) if you want to access
+            :meth:`~optuna.integration.LightGBMTuner.best_booster` in distributed environments.
+            Otherwise, it may raise :obj:`ValueError`.
```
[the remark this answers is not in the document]
toshihikoyanase: Thank you for your suggestion. Applied: https://github.com/optuna/optuna/pull/1076/commits/8c90f6c6b4f2e747b5908f77dd1d5efcd642818a.

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: > `booster.save_model` does not save many attributes of the booster. It leads to some inconvenience in practical usage.

Thank you for your comment. I also had a problem when I write the [test case](https://github.com/optuna/optuna/pull/1076/files#diff-20171b34cc0c35c6756d064ab0ccf033R659) because `booster.save_model` does not persist `booster.params`. I think it is the expected behavior of [`lightgbm.Booster`](https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/basic.py#L1831).

I think `pickle` is a reasonable choice since the [official example](https://github.com/microsoft/LightGBM/blob/master/examples/python-guide/advanced_example.py) also uses it.

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -624,22 +644,34 @@ def tune_params(self, target_param_names, n_trials, sampler, step_name):
...
+            t
+            for t in study.trials
+            if t.state in (optuna.trial.TrialState.COMPLETE, optuna.trial.TrialState.PRUNED)
+        ]
+        _n_trials = n_trials - len(complete_trials)
+        if _n_trials > 0:
+            try:
+                study.optimize(objective, n_trials=_n_trials, catch=())
+            except ValueError:
+                # ValueError is rased by GridSampler when all combinations were examined.
```
toshihikoyanase, line 664:
```suggestion
                # ValueError is raised by GridSampler when all combinations were examined.
                # This error handling will be removed after Study.stop is implemented.
```

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -457,6 +448,38 @@ def best_params(self) -> Dict[str, Any]:
...
 
+    @property
+    def best_booster(self) -> lgb.Booster:
+        """Return the best booster."""
+        if self._best_booster_with_trial_number is not None:
+            if self._best_booster_with_trial_number[1] == self.study.best_trial.number:
+                return self._best_booster_with_trial_number[0]
+        else:
+            if len(self.study.trials) == 0:
+                raise ValueError("The best booster is not available because no trials completed.")
```
toshihikoyanase, line 459: `else:` is not necessary.
```suggestion
        if len(self.study.trials) == 0:
            raise ValueError("The best booster is not available because no trials completed.")
```

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -457,6 +448,38 @@ def best_params(self) -> Dict[str, Any]:
             params.update(self.lgbm_params)
             return params
 
+    @property
+    def best_booster(self) -> lgb.Booster:
+        """Return the best booster."""
```
toshihikoyanase, line 453:
```suggestion
        """Return the best booster.

        If the best booster cannot be found, :class:`ValueError` will be raised. To prevent the
        errors, please save boosters by specifying the `model_dir` arguments of
        :meth:`~optuna.integration.lightgbm.LightGBMTuner.__init__` when you resume tuning
        or you run tuning in parallel.
        """
```

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -457,6 +448,38 @@ def best_params(self) -> Dict[str, Any]:
...
+        if self._best_booster_with_trial_number is not None:
+            if self._best_booster_with_trial_number[1] == self.study.best_trial.number:
+                return self._best_booster_with_trial_number[0]
+        else:
+            if len(self.study.trials) == 0:
+                raise ValueError("The best booster is not available because no trials completed.")
+
+        if self._model_dir is None:
```
toshihikoyanase, line 461:
```suggestion
        # The best booster exists, but this instance does not have it.
        # This may be due to resuming or parallelization.
        if self._model_dir is None:
```

toshihikoyanase:
Thank you for joining the mob-review, @smly !
I applied your suggestions in https://github.com/optuna/optuna/pull/1076/commits/f43ba1c342532a9ac6821ec51b3ffe2ca432ac49.
If I have overlooked your comments, please let me know.

smly:
When using RDB (MySQL), LightGBMTuner doesn't work.
It seems that `sample_independent` on `GridSampler` returns `np.float64` instead of `float`.

* Python version: 3.7.3
* Commit: daaa75f279f42f313ea20d5a6213ec32c2210f02
* Error log and script: https://gist.github.com/smly/50f78c526cca5c0673d277f0a5be0924

smly:
> I applied your suggestions in f43ba1c.

Thanks for the updates! Great!

toshihikoyanase:
> When using RDB (MySQL), LightGBMTuner doesn't work.
> It seems that sample_independent on GridSampler returns np.float64 instead of float.

@smly Thank you for your careful review.
Currently, `list(np.ndarray)` is used to create `param_values` which is an argument of `GridSampler`, and I find that `np.ndarray.tolist` convert the element of `np.ndarray` into Python built-in types. (See https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.tolist.html for further details.)

Let me share the quick fix https://github.com/optuna/optuna/pull/1076/commits/38d86003622b1a49ce4cafed4df5ec9157fcc509.  I'll check it with MySQL.

smly:
Thanks for your quick update! It resolved after applying 38d8600.

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -336,6 +315,13 @@ class LightGBMTuner(BaseTuner):
             ``average_iteration_time`` is the average time of iteration to train the booster
             model in the trial. ``lgbm_params`` is a JSON-serialized dictionary of LightGBM
             parameters used in the trial.
+
+        model_dir:
+            A directory to save boosters. By default, it is set to :obj:`None` and no boosters are
+            saved. Please set shared directory (e.g., directories on NFS) if you want to access
+            :meth:`~optuna.integration.LightGBMTuner.best_booster` in distributed environments.
+            Otherwise, it may raise :obj:`ValueError`. If the directory does not exist, it will be
+            created.
```
hvy, line 324: How about describing how the file names are determined?

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -457,6 +448,45 @@ def best_params(self) -> Dict[str, Any]:
             params.update(self.lgbm_params)
             return params
 
+    @property
+    def best_booster(self) -> lgb.Booster:
```
hvy, line 452: Looking at the new logic, this shouldn't really be a `@property` as operations aren't trivial. It's probably good to keep backwards compatibility but I'd personally deprecate this one and replace it with a method.

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -559,8 +580,10 @@ def tune_feature_fraction(self, n_trials=7):
         # type: (int) -> None
 
         param_name = "feature_fraction"
-        param_values = list(np.linspace(0.4, 1.0, n_trials))
-        sampler = _GridSamplerUniform1D(param_name, param_values)
+        param_values = np.linspace(0.4, 1.0, n_trials).tolist()
+        with warnings.catch_warnings():
```
hvy, line 584: Let's leave a TODO comment to remove this context.

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -580,11 +603,13 @@ def tune_feature_fraction_stage2(self, n_trials=6):
...
         best_feature_fraction = self.best_params[param_name]
-        param_values = list(
-            np.linspace(best_feature_fraction - 0.08, best_feature_fraction + 0.08, n_trials)
-        )
+        param_values = np.linspace(
+            best_feature_fraction - 0.08, best_feature_fraction + 0.08, n_trials
+        ).tolist()
         param_values = [val for val in param_values if val >= 0.4 and val <= 1.0]
-        sampler = _GridSamplerUniform1D(param_name, param_values)
+        with warnings.catch_warnings():
```
hvy, line 610: Ditto.

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -602,7 +627,9 @@ def tune_min_data_in_leaf(self):
 
         param_name = "min_child_samples"
         param_values = [5, 10, 25, 50, 100]
-        sampler = _GridSamplerUniform1D(param_name, param_values)
+        with warnings.catch_warnings():
```
hvy, line 630: Ditto.

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -49,37 +51,6 @@
 }
 
 
-class _GridSamplerUniform1D(optuna.samplers.BaseSampler):
```
hvy, line 52: Great that we can reuse the new grid sampler!

optuna/integration/lightgbm_tuner/optimize.py
hvy, line 254: How about logging this operation, including the file path?

optuna/integration/lightgbm_tuner/optimize.py
hvy: I think it's okay to discuss it later as we already have an approval for the current changes.

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: Thank you for your suggestion. I added it as follows:
> The filenames of the boosters will be ``{model_dir}/{trial_number}.pkl`` (e.g., ``./boosters/0.pkl``).

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: > Looking at the new logic, this shouldn't really be a @property as operations aren't trivial. 

I agree with you. This behavior is too complicated as a property.
So, I'll create a `get_best_booster` method and deprecate the `best_booster` property in this PR.
@smly What do you think?

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: Thanks! I added it.

optuna/integration/lightgbm_tuner/optimize.py
smly: Totally agree with your proposed change. 👍

toshihikoyanase: Thank you for your quick response. I'll apply changes.

toshihikoyanase: I added the `LightGBMTuner.get_best_booster` method and deprecated `LightGBMTuner.best_booster` in https://github.com/optuna/optuna/pull/1076/commits/bc6d1a1e94be9627b520f30b334d153facddcb77.

toshihikoyanase:
@smly
> How about adding the example code?  It must be helpful to understand the usage. I think it's also a good option to create a contribution-welcome issue instead of resolving it in this PR.

Sounds good! I think we can use the same scenario with `examples/lightgbm_tuner_simple.py` and replace `lightgbm_tuner.train` with `lightgbm_tuner.LightGBMTuner`. If it's OK, I'll create a contribution-welcome issue.

> IMO, in the example code of distributed LightGBM Tuner, it might be better to remove the code that uses best_booster to do inference (predict) and just output the best parameter.

I think it's reasonable. I'll add it to the description when I create the issue.

smly:
> I think we can use the same scenario with examples/lightgbm_tuner_simple.py and replace lightgbm_tuner.train with lightgbm_tuner.LightGBMTuner.

LGTM. Thanks a lot!

optuna/integration/lightgbm_tuner/__init__.py
```diff
@@ -31,8 +31,8 @@ def train(*args: Any, **kwargs: Any) -> Any:
     _check_lightgbm_availability()
 
     auto_booster = LightGBMTuner(*args, **kwargs)
-    booster = auto_booster.run()
-    return booster
+    auto_booster.run()
+    return auto_booster.best_booster
```
hvy, line 35: Do we need the following change to avoid the deprecation warning?
```suggestion
    return auto_booster.get_best_booster()
```

toshihikoyanase: You're absolutely right. I applied your suggestion. Thank you!
Commit: https://github.com/optuna/optuna/pull/1076/commits/9d3b6b4abeb2626c2fc32113588392d6830687c9

optuna/integration/lightgbm_tuner/optimize.py
hvy: Thanks for the discussion and the changes!

PR #1077, opened 2020-04-02
Add `number` property to `FixedTrial` and `BaseTrial`.

### LINKED ISSUE #426: "number" with FixedTrial might be useful
When testing optuna objective, `FixedTrial` are suggested to be used.
And, when saving machine learning models, `Trial.number` are suggested to be used to identify the trial.
However, `FixedTrial` don't have `number`/`trial_id` feature which prevents testing of objective.
Thus adding mock function of `number` might be helpful for testing.

ytsmiling:
This PR adds `number` property to FixedTrial and BaseTrial.
The constructor of FixedTrial is changed to accept the `number` argument to make testings easier.

This PR resolves #426.


optuna/trial.py
```diff
@@ -109,6 +109,12 @@ def datetime_start(self):
 
         raise NotImplementedError
 
+    @property
+    def number(self):
+        # type: () -> int
```
toshihikoyanase, line 114: Optuna dropped Python2 support and we can use a function annotation now. Please use it instead of the type comment. (See https://github.com/optuna/optuna/issues/711)
```suggestion
    def number(self) -> int:
```

optuna/trial.py
```diff
@@ -1050,6 +1057,12 @@ def datetime_start(self):
 
         return self._datetime_start
 
+    @property
+    def number(self):
+        # type: () -> int
```
toshihikoyanase, line 1062: ditto.
```suggestion
    def number(self) -> int:
```

tests/test_trial.py
```diff
@@ -421,6 +421,14 @@ def test_fixed_trial_datetime_start():
     assert trial.datetime_start is not None
 
 
+def test_fixed_trial_number():
+    # type: () -> None
```
toshihikoyanase, line 425: ditto.

```suggestion
def test_fixed_trial_number() -> None:
```

optuna/trial.py
ytsmiling: Thank you for pointing it out. I resolved it.

optuna/trial.py
```diff
@@ -927,15 +933,16 @@ def objective(trial):
 
     """
 
-    def __init__(self, params):
-        # type: (Dict[str, Any]) -> None
+    def __init__(self, params, number=0):
```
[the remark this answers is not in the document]
ytsmiling: Thank you for the review.
I added it to the class doc. Please re-review its sufficiency and format.

sile:
The CI jobs `doctest` and `document` were failed, but the cause seems to be resolved by https://github.com/optuna/optuna/pull/1093. Could you merge `master` branch to confirm it?

toshihikoyanase:
Let me slightly modify the PR title to quote the class/property names.

PR #1081, opened 2020-04-02
Support DiscreteUniformDistribution in `suggest_float`.

### LINKED ISSUE #1069: Support discrete parameters by `Trial.suggest_float()`.
This is a followup of #1021 , and it focuses on  https://github.com/optuna/optuna/pull/1021#pullrequestreview-376794576. The full discussion of `suggest_float` can be found the original issue  https://github.com/optuna/optuna/issues/510.

## Description

Please add `step` parameter to `Trial.suggest_float` in order to suggest discrete parameter like [`Trial.suggest_discrete_uniform()`](https://optuna.readthedocs.io/en/latest/reference/trial.html#optuna.trial.Trial.suggest_discrete_uniform).

I think we have two  points to be discussed as follows:
- parameter name
- `log=True` and ~~`step!=1`~~ `step is not None`

### Parameter name

I think `step` is the first choice to be consistent with [`Trial.suggest_int`](https://optuna.readthedocs.io/en/latest/reference/trial.html#optuna.trial.Trial.suggest_int). The `Trial.suggest_int` employed `step` in https://github.com/optuna/optuna/pull/910.

### `log=True` and ~~`step!=1`~~ `step is not None`

In https://github.com/optuna/optuna/issues/510, I think the design discussion about this case has not been finished yet. So, I think it would be reasonable to raise `NotImplemented` for now.

himkt:
## Description

This PR introduces `step` parameter to `Trial.suggest_float`.

Close #1069.

### DIFF
--- a/optuna/trial.py
+++ b/optuna/trial.py
@@ -1,7 +1,6 @@
 import abc
 from datetime import datetime
 import decimal
-import math
 import warnings
 
 from optuna import distributions
--- a/optuna/trial.py
+++ b/optuna/trial.py
@@ -250,7 +250,7 @@ class Trial(BaseTrial):
 
         if step is not None:
             if log:
-                raise NotImplementedError("The parameter step is not supported when log=True.")
+                raise NotImplementedError("The parameter `step` is not supported when `log` is True.")
             else:
                 return self.suggest_discrete_uniform(name, low, high, step)
         else:
@@ -977,7 +977,7 @@ class FixedTrial(BaseTrial):
 
         if step is not None:
             if log:
-                raise NotImplementedError("The parameter step is not supported when log=True.")
+                raise NotImplementedError("The parameter `step` is not supported when `log` is True.")
             else:
                 return self._suggest(
                     name, distributions.DiscreteUniformDistribution(low=low, high=high, q=step)
--- a/optuna/trial.py
+++ b/optuna/trial.py
@@ -182,14 +182,16 @@ class Trial(BaseTrial):
         # type: (str, float, float, bool, Optional[float]) -> float
         """Suggest a value for the floating point parameter.
 
-        Note that this is a wrapper method for :func:`~optuna.trial.Trial.suggest_uniform`
-        and :func:`~optuna.trial.Trial.suggest_loguniform`.
+        Note that this is a wrapper method for :func:`~optuna.trial.Trial.suggest_uniform`,
+        :func:`~optuna.trial.Trial.suggest_loguniform` and
+        :func:`~optuna.trial.Trial.suggest_discrete_uniform`.
 
         .. versionadded:: 1.3.0
 
         .. seealso::
-            Please see also :func:`~optuna.trial.Trial.suggest_uniform` and
-            :func:`~optuna.trial.Trial.suggest_loguniform`.
+            Please see also :func:`~optuna.trial.Trial.suggest_uniform`,
+            :func:`~optuna.trial.Trial.suggest_loguniform` and
+            :func:`~optuna.trial.Trial.suggest_discrete_uniform`.
 
         Example:
 
--- a/optuna/trial.py
+++ b/optuna/trial.py
@@ -30,8 +30,8 @@ class BaseTrial(object, metaclass=abc.ABCMeta):
     """
 
     @abc.abstractmethod
-    def suggest_float(self, name, low, high, *, log=False):
-        # type: (str, float, float, bool) -> float
+    def suggest_float(self, name, low, high, *, log=False, step=None):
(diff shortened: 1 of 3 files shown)


himkt:
I made this PR ready for review since I want to discuss this PR with members.

In #1069, it seems that raising `NotImplementedError` is expected only when `log=True and step!=1`.
However, `step` would be negative when `0<step<1` and be undefined when `step<=0`.
When `log=True` and `step<1`, what behavior is expected?

When `log=True` and `1<step`, I implemented `suggest_float` to call `suggest_discrite_uniform` with `low=low`, `high=high` and `q=math.log(step)`. I'm not sure whether this implementation is correct.

Could anyone take a look at this PR, please?

toshihikoyanase:
> In #1069, it seems that raising `NotImplementedError` is expected only when `log=True and step!=1`.

I'm really sorry but I made a mistake in #1069.  The condition should be `log=True and step is not None`.
I think we may need to add a new distribution like `LogDiscreteFloatDistribution` and its corresponding calculation to support the condition. This PR can focus on the API design if we separate the PRs.

himkt:
@toshihikoyanase Thank you for clarification, I revised my PR to raise `NotImplementedError` when
`log=True` and `step is not None` in 8a76e87.

himkt:
@hvy I'm sorry and thank you for catching my violating conventions.
(In the thread of #1087, I couldn't see the reference to this PR...:thinking:)

himkt:
Now I can see.

Screen Shot 2020-04-08 at 17 18 08

hvy:
Regarding how `step` may behave when `log=True`:
1. SMAC3 (`ConfigSpace`) treats the "step" as given in the log space when `log=True`. https://github.com/automl/ConfigSpace/blob/master/ConfigSpace/hyperparameters.pyx#L383 This seems a bit unintuitive to me at first glance as the bounds are not, by I might be missing something. 
2. Hyperopt is clear by its API design that similar to SMAC3, "step" is given in the log space. https://github.com/hyperopt/hyperopt/blob/d87309463a1496cb66d83827afe204d0700e4aef/docs/templates/getting-started/search_spaces.md#parameter-expressions
So, it seems like treating the "step" in the log space is consistent with other libraries, but again, it's likely open for discussion.

optuna/trial.py
```diff
@@ -158,8 +158,8 @@ def _init_relative_params(self):
             self.study, trial, self.relative_search_space
         )
 
-    def suggest_float(self, name, low, high, *, log=False):
-        # type: (str, float, float, bool) -> float
+    def suggest_float(self, name, low, high, *, log=False, step=None):
+        # type: (str, float, float, bool, Optional[float]) -> float
         """Suggest a value for the floating point parameter.
 
         Note that this is a wrapper method for :func:`~optuna.trial.Trial.suggest_uniform`
```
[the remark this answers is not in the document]
himkt: Thank you for catching, I added in 69680e5.

optuna/trial.py
```diff
@@ -237,15 +241,23 @@ def objective(trial):
...
 
         Returns:
             A suggested float value.
         """
 
-        if log:
-            return self.suggest_loguniform(name, low, high)
+        if step is not None:
+            if log:
+                raise NotImplementedError("The parameter step is not supported when log=True.")
```
toshihikoyanase, line 253: How about adding quotes to the argument names as can be seen in [distribution.py](https://github.com/optuna/optuna/blob/master/optuna/distributions.py#L313)?
 
```suggestion
                raise NotImplementedError("The parameter `step` is not supported when `log` is True.")
```

optuna/trial.py
```diff
@@ -960,13 +972,23 @@ def __init__(self, params, number=0):
...
-    def suggest_float(self, name, low, high, *, log=False):
-        # type: (str, float, float, bool) -> float
+    def suggest_float(self, name, low, high, *, log=False, step=None):
+        # type: (str, float, float, bool, Optional[float]) -> float
 
-        if log:
-            return self._suggest(name, distributions.LogUniformDistribution(low=low, high=high))
+        if step is not None:
+            if log:
+                raise NotImplementedError("The parameter step is not supported when log=True.")
```
toshihikoyanase, line 980: ditto.
```suggestion
                raise NotImplementedError("The parameter `step` is not supported when `log` is True.")
```

optuna/trial.py
himkt: Thank you for catching, I revised my PR.

himkt:
@toshihikoyanase Sorry for not replying, I revised my PR. PTAL.

himkt:
rel. #510

Issue #1063, opened 2020-03-29
Delete the argument `is_higher_better` from `TensorFlowPruningHook`.


keisuke-umezawa:
## Motivation

Because the argument [`is_higher_better` is obsolete](https://github.com/optuna/optuna/blob/master/optuna/integration/tensorflow.py#L56-L61) in `TensorFlowPruningHook`, we can delete `is_higher_better`.

## Description

Same as above.

himkt:
@keisuke-umezawa 
I think this issue has been resolved in #1083.

hvy:
Thanks for pointing that out. You're absolutely right, let me close this ticket.

PR #1083, opened 2020-04-02
Delete the argument `is_higher_better` from `TensorFlowPruningHook`

nuka137:
Related Issue: #1063 

Note: Test for the argument `is_higher_better` is also deleted.

### DIFF
--- a/optuna/integration/tensorflow.py
+++ b/optuna/integration/tensorflow.py
@@ -38,7 +38,7 @@ class TensorFlowPruningHook(SessionRunHook):
     """
 
     def __init__(self, trial, estimator, metric, run_every_steps):
-        # type: (optuna.trial.Trial, tf.estimator.Estimator, str, int, Optional[bool]) -> None
+        # type: (optuna.trial.Trial, tf.estimator.Estimator, str, int) -> None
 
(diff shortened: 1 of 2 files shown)


optuna/integration/tensorflow.py
```diff
@@ -35,13 +35,9 @@ class TensorFlowPruningHook(SessionRunHook):
...
            An interval to watch the summary file.
-        is_higher_better:
-           Please do not use this argument because this class refers to
-           :class:`~optuna.structs.StudyDirection` to check whether the current study is
-           ``minimize`` or ``maximize``.
     """
 
-    def __init__(self, trial, estimator, metric, run_every_steps, is_higher_better=None):
+    def __init__(self, trial, estimator, metric, run_every_steps):
         # type: (optuna.trial.Trial, tf.estimator.Estimator, str, int, Optional[bool]) -> None
```
toshihikoyanase, line 41: Please update the type hinting.
```suggestion
        # type: (optuna.trial.Trial, tf.estimator.Estimator, str, int) -> None
```

nuka137: Thanks!
I applied your suggested change.

PR #1085, opened 2020-04-03
Add visualization example script.

### LINKED ISSUE #605: Visualization examples require Jupyter Notebook
Visualization examples require Jupyter Notebook which is in contrast to other examples. Some users might prefer running the visualization examples as Python scripts (`.py`) as with the other examples. 

For the record, when plotting a plotly figure from a Python script, a local server is started and a browser will be opened to render it interactively, the content being similar to Jupyter Notebook.

ytsmiling:
This PR adds visualization example without jupyter notebook and resolve #605.

### DIFF
--- a/examples/visualization/plot_study.py
+++ b/examples/visualization/plot_study.py
@@ -5,7 +5,7 @@ In this example, we optimize the validation accuracy of object recognition using
 scikit-learn and Fashion-MNIST. We optimize a neural network. As it is too time
 consuming to use the whole Fashion-MNIST dataset, we here use a small subset of it.
 
-We have can execute this example as follows.
+We can execute this example as follows.
 
     $ python plot_study.py
 
--- a/examples/visualization/plot_study.py
+++ b/examples/visualization/plot_study.py
@@ -63,26 +63,26 @@ if __name__ == "__main__":
     study = optuna.create_study(direction="maximize", pruner=optuna.pruners.MedianPruner())
     study.optimize(objective, n_trials=100, timeout=600)
 
-    # Visualize the Optimization History
+    # Visualize the optimization history.
     plot_optimization_history(study).show()
 
-    # Visualize the Learning Curves of the Trials
+    # Visualize the learning curves of the trials.
     plot_intermediate_values(study).show()
 
-    # Visualize High-dimensional Parameter Relationships
+    # Visualize high-dimensional parameter relationships.
     plot_parallel_coordinate(study).show()
 
-    # Select Parameters to Visualize
+    # Select parameters to visualize.
     plot_parallel_coordinate(study, params=["lr_init", "n_units_l0"]).show()
 
-    # Visualize Hyperparameter Relationships
+    # Visualize hyperparameter relationships.
     plot_contour(study).show()
 
-    # Select Parameters to Visualize
+    # Select parameters to visualize.
     plot_contour(study, params=["n_units_l0", "n_units_l1"]).show()
 
-    # Visualize Individual Hyperparameters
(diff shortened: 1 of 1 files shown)


examples/visualization/plot_study.py
```diff
@@ -0,0 +1,92 @@
...
+    data = fmnist.data[:n_samples]
+    target = fmnist.target[:n_samples]
+
+    x_train, x_test, y_train, y_test = train_test_split(data, target)
+
+    def objective(trial):
+
+        clf = MLPClassifier(
+            hidden_layer_sizes=tuple(
+                [trial.suggest_int("n_units_l{}".format(i), 32, 64) for i in range(3)]
```
crcrpar, line 43: [nit] We can also optimize the number of layers, but I'd also not optimize it in this example since the objective is to make the visualization feature in use.

examples/visualization/plot_study.py
```diff
@@ -0,0 +1,92 @@
+"""
+Optuna example that visualizes the optimization result of multi-layer perceptrons.
+
+In this example, we optimize the validation accuracy of object recognition using
+scikit-learn and Fashion-MNIST. We optimize a neural network. As it is too time
+consuming to use the whole Fashion-MNIST dataset, we here use a small subset of it.
+
+We have can execute this example as follows.
```
crcrpar, line 8:
```suggestion
We can execute this example as follows.
```

examples/visualization/plot_study.py
```diff
@@ -0,0 +1,92 @@
...
+            value = clf.score(x_test, y_test)
+
+            # Report intermediate objective value.
+            trial.report(value, step)
+
+            # Handle pruning based on the intermediate value.
+            if trial.should_prune(step):
+                raise optuna.exceptions.TrialPruned()
+
+        return value
```
crcrpar, line 59: [question] Is `value` still alive here? I maybe am missing something...

ytsmiling: Yes, it's alive.

However, future change might make the code access to the `value` before the declaration (e.g., break from the loop before the first assignment), and some IDEs might warn the usage of `value`.
I'd like to change the code to declare `value` before the loop if you prefer.

crcrpar: No need to do that, thanks 🙂

examples/visualization/plot_study.py
```diff
@@ -0,0 +1,92 @@
...
+def build_objective():
+    fmnist = fetch_openml(name="Fashion-MNIST", version=1)
+    classes = list(set(fmnist.target))
+
+    # For demonstrational purpose, only use a subset of the dataset.
+    n_samples = 4000
+    data = fmnist.data[:n_samples]
+    target = fmnist.target[:n_samples]
+
+    x_train, x_test, y_train, y_test = train_test_split(data, target)
```
sile, line 37: I think that we can move this part into the beginning of the `objective` function and remove the slightly tricky `build_objective` function (c.f. [sklearn_simple.py](https://github.com/optuna/optuna/blob/master/examples/sklearn_simple.py#L32)).

ytsmiling: I initially thought that preparing an in-memory dataset outside the objective to avoid unnecessary disk I/O is beneficial, but I decided to change the code as suggested and follow the optuna's convention.

sile: Thank you for your understanding.
Optuna examples prefer simpler code to efficient but a bit complex one because there would be many users who are not Python experts.
(honestly, immediately after joining Optuna development, I had the same thoughts with you)

Issue #1084, opened 2020-04-03
Apply `@abc.abstractmethod` decorator to the abstract methods of `BaseTrial`.


sile:
Currently, the abstract methods of the `BaseTrial` class aren't decorated by `@abc.abstractmethod`, but I think that it is not intentional. We should apply the decorator to the methods since it is a good Python convention to follow.

PR #1087, opened 2020-04-04
Applied `@abc.abstractmethod` decorator to the abstract methods of `BaseTrial`

gorogoroumaru:
issue #1084

This PR applied `@abc.abstractmethod` decorator to the abstract methods of `BaseTrial`

### DIFF
--- a/optuna/integration/chainermn.py
+++ b/optuna/integration/chainermn.py
@@ -196,7 +196,7 @@ class ChainerMNTrial(BaseTrial):
             # type: () -> float
 
             assert self.delegate is not None
-            return self.delegate.suggest_float(name, low, high)
+            return self.delegate.suggest_float(name, low, high, log)
 
         return self._call_with_mpi(func)
 
--- a/optuna/integration/chainermn.py
+++ b/optuna/integration/chainermn.py
@@ -196,7 +196,7 @@ class ChainerMNTrial(BaseTrial):
             # type: () -> float
 
             assert self.delegate is not None
-            return self.delegate.suggest_float(name, low, high, log)
+            return self.delegate.suggest_float(name, low, high, log=log)
 
         return self._call_with_mpi(func)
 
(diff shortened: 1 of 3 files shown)


optuna/integration/chainermn.py
```diff
@@ -189,6 +189,17 @@ def __init__(self, trial, comm):
...
         self.comm = comm
 
+    def suggest_float(self, name, low, high, *, log=False):
+        # type: (str, float, float, bool) -> float
+
+        def func():
+            # type: () -> float
+
+            assert self.delegate is not None
+            return self.delegate.suggest_float(name, low, high)
```
crcrpar, line 199: It seems like the keyword argument of `log` is ignored.

Isn't it better to pass `log` to `self.delegate.suggest_float`?

crcrpar, line 199: It seems like the keyword argument of `log` is ignored.

Isn't it better to pass `log` to `self.delegate.suggest_float`?

gorogoroumaru: Thank you!
I will fix it.

crcrpar:
Thanks for the swift update. In my opinion, the newly implemented `suggest_float` should also be tested (ref: https://github.com/optuna/optuna/pull/1021) so that now I think this PR has better to be separated.

However, again, I'm not sure, so you don't need to hurry to separate this PR. 🙂

tests/integration_tests/test_chainermn.py
```diff
@@ -304,6 +304,37 @@ def test_init(storage_mode, comm):
 
             assert mn_trial.number == trial.number
 
+    @staticmethod
+    @pytest.mark.parametrize("storage_mode", STORAGE_MODES)
+    def test_suggest_float(storage_mode, comm):
```
hvy, line 309: How about also checking that exceptions are raised as excepted similar to other tests below?

gorogoroumaru: Thanks!
I applied your suggested change.

hvy: Thanks for your quick action.

hvy:
Allow me to update the title to somewhat better reflect the changes.

PR #1088, opened 2020-04-04
Add notebook verification for visualization examples.

### LINKED ISSUE #683: [RFC] CI for examples/visualization
It'll be better if we can validate that all the visualization examples run without failures with CI.

I guess https://github.com/computationalmodelling/nbval/ might be useful.

ytsmiling:
This PR makes CI checks whether notebooks run without failure. This PR resolves #683.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -54,7 +54,10 @@ jobs:
             python $file --pruning > /dev/null
           fi
         done
-
+      env:
+        OMP_NUM_THREADS: 1
+    - name: Run examples
+      run: |
         for file in `find examples -name '*.ipynb'`
         do
           echo $file
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -32,6 +32,9 @@ jobs:
 
         # Install all dependencies needed for examples.
         pip install --progress-bar off $(ls dist/*.tar.gz)[example]
+
+        # Install additional test-specific dependencies
+        pip install --progress-bar off nbval
     - name: Run examples
       run: |
         if [ ${{ matrix.python-version }} = 3.5 ]; then
@@ -50,6 +53,12 @@ jobs:
             echo $file --pruning
             python $file --pruning > /dev/null
           fi
+        done
+
+        for file in `find examples -name '*.ipynb'`
+        do
+          echo $file
+          pytest --nbval-lax $file > /dev/null
         done
       env:
         OMP_NUM_THREADS: 1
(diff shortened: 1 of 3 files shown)


.github/workflows/examples.yml
```diff
@@ -32,6 +32,9 @@ jobs:
 
         # Install all dependencies needed for examples.
         pip install --progress-bar off $(ls dist/*.tar.gz)[example]
+
+        # Install additional test-specific dependencies
+        pip install --progress-bar off nbval
```
ytsmiling, line 37: Is it okay to add test-specific dependencies here, or shall I move to examples' requirements?

.github/workflows/examples.yml
```diff
@@ -51,6 +54,12 @@ jobs:
             python $file --pruning > /dev/null
           fi
         done
+
+        for file in `find examples -name '*.ipynb'`
```
ytsmiling, line 58: Should I split this test case like `Run multi-node examples`?

.github/workflows/examples.yml
sile: I think that it's not a bad idea to install `nbval` here, but for future maintainability, it seems better to add it to the "example" dependency list defined in [setup.py](https://github.com/optuna/optuna/blob/master/setup.py#L57).

.github/workflows/examples.yml
sile: I prefer to split.

examples/visualization/plot_study.ipynb
```diff
@@ -136,7 +136,7 @@
         "def objective(trial):\n",
         "    \n",
         "    clf = MLPClassifier(\n",
-        "        hidden_layer_sizes=tuple([trial.suggest_int(f'n_units_l{i}', 32, 64) for i in range(3)]),\n",
+        "        hidden_layer_sizes=tuple([trial.suggest_int('n_units_l{}'.format(i), 32, 64) for i in range(3)]),\n",
```
sile, line 139: Nice catch 👍

setup.py
```diff
@@ -61,6 +61,7 @@ def get_extras_require() -> Dict[str, List[str]]:
             "mlflow",
             "mpi4py",
             "mxnet",
+            "nbeval",
```
crcrpar, line 64: Isn't this a typo? 

ref: https://github.com/computationalmodelling/nbval#installation

```suggestion
            "nbval",
```

ytsmiling: Thank you!

ytsmiling: It is! Thank you!

PR #1094, opened 2020-04-06
Revert "GitHub Actions to automatically label PRs."

hvy:
Reverts optuna/optuna#1068


hvy:
As stated in the README of https://github.com/actions/labeler, this actions does seemingly not work with forks (as they do not have write permissions to the upstream repository and there's nothing we can do).

PR #1099, opened 2020-04-08
Input validation for `LogUniformDistribution` for negative domains.

hvy:
Possibly a bug. Passing a bound to `LogUniformDistribution` that's not strictly positive could lead to unexpected behaviors (e.g. divide by zero error in the `RandomSampler`) as it's not  defined. This PR validates the input to for a friendlier error message. It also adds tests for `suggest_loguniform`.

#### Reproduction

```python
import optuna


def objective(trial):
    # Lower bound is not strictly positive so this will yield a divide by zero error (in the master branch).
    return trial.suggest_loguniform("x", 0.0, 10)


study = optuna.create_study()
```

```bash
/Users/hvy/repos/optuna/optuna/samplers/random.py:75: RuntimeWarning:

divide by zero encountered in log

[W 2020-04-08 09:39:39,747] Setting status of trial#0 as TrialState.FAIL because of the following error: OverflowError('Range exceeds valid bounds')
  ...
  File "/Users/hvy/repos/optuna/optuna/samplers/random.py", line 77, in sample_independent
    return float(numpy.exp(self._rng.uniform(log_low, log_high)))
  File "mtrand.pyx", line 1091, in numpy.random.mtrand.RandomState.uniform
OverflowError: Range exceeds valid bounds
```


tests/test_trial.py
```diff
@@ -133,6 +133,32 @@ def test_suggest_uniform(storage_init_func):
...
+        distributions.LogUniformDistribution(low=0.0, high=0.9)
+
+    mock = Mock()
+    mock.side_effect = [1.0, 2.0, 3.0]
+    sampler = samplers.RandomSampler()
+
+    with patch.object(sampler, "sample_independent", mock) as mock_object:
+        study = create_study(storage_init_func(), sampler=sampler)
+        trial = Trial(study, study._storage.create_new_trial(study._study_id))
+        distribution = distributions.LogUniformDistribution(low=0.1, high=3.0)
```
toshihikoyanase, line 153: It may be a nit-picky comment, but I don't think `LogUniformDistribution` includes the `high` value.
https://optuna.readthedocs.io/en/stable/reference/trial.html#optuna.trial.Trial.suggest_loguniform
```suggestion
        distribution = distributions.LogUniformDistribution(low=0.1, high=4.0)
```

hvy: Thanks for the catch. I applied your suggestion.

hvy:
PTAL.

hvy:
Resolved conflicts, PTAL.

PR #1102, opened 2020-04-08
Update type hinting of `GridSampler.__init__`.

toshihikoyanase:
The current type hinting of `GridSampler.__init__` is `Dict[str, List[Union[str, float, int, bool, None]
]`. As mentioned in https://github.com/optuna/optuna/pull/1076#discussion_r405252822, I found that `List[int]` was not compatible with `List[Union[str, float, int, bool, None]`. 

In this PR, `List` is replaced with `Sequence` for mypy to infer the type properly. The [typing reference](https://docs.python.org/3/library/typing.html#typing.List) recommend `Sequence` or `Iterable` instead of `List`.

> Generic version of list. Useful for annotating return types. To annotate arguments it is preferred to use an abstract collection type such as Sequence or Iterable.

`Dict` is also replaced with `Mapping` for the same reason.

### DIFF
--- a/optuna/samplers/grid.py
+++ b/optuna/samplers/grid.py
@@ -10,6 +10,8 @@ if type_checking.TYPE_CHECKING:
     from typing import Any  # NOQA
     from typing import Dict  # NOQA
     from typing import List  # NOQA
+    from typing import Mapping  # NOQA
+    from typing import Sequence  # NOQA
     from typing import Union
 
     from optuna.distributions import BaseDistribution  # NOQA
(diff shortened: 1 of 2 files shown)


hvy:
Do you by the way think that we should enforce this to other argument type hints as suggested in the official documentation?

himkt:
Personally, +1 for promoting.

hvy:
Two approvals, let me merge this PR.

hvy:
I personally find more constrained and concrete types clearer and allows the implementation to stay simple, but if `mypy` and other tools depend on this recommendation, maybe we should.

toshihikoyanase:
> Do you by the way think that we should enforce/promote this to other argument type hints as suggested in the official documentation?

I think that depends on the case. I feel that the modification in this PR is a kind of exceptional cases where the type hint has relatively large ambiguity. IMO, we can usually use concrete types before we face the mypy errors.

PR #1106, opened 2020-04-08
Implement fANOVA using `sklearn` instead of `fanova`.

hvy:
This PR continues on the experimental hyperparameter importance feature https://github.com/optuna/optuna/pull/946. It attempts to reimplements the same algorithm using `sklearn` instead of `fanova` to allow a wider range of users to use this feature.

#### Comparisons with `fanova` 

_See https://github.com/optuna/optuna/pull/1106#issuecomment-614544840 for updated plots where categorical distributions have been fixed._

Comparison between `Fanova` (blue, this implementation) and `fANOVA` (orange, original `fanova` implementation). Both single and pairwise importances are compared. The results are averaged over 10 different runs with different random seeds, for each implementation. Numerical distributions seems close while categorical may slightly differ.

Note that the current importance evaluation API in Optuna only allows single importances. To compute pairwise, you'll need to call the lower level, currently private, interfaces. Those were used to generate the following plots.

<details>
  <summary>LightGBM, 100 trials, `RandomSampler`</summary>
  
</details>
<details>
  <summary>PyTorch, 100 trials, `RandomSampler`</summary>
  
</details>
<details>
  <summary>ffmpeg, 150 trials, `RandomSampler`</summary>
  
</details>

#### TODO

- [x] Importance/performance comparisons with `fanova`.
- [x] Type hints.
- [x] Unit tests. Not sure to what extent we should test private interfaces, if at all. -> Wrote minimal tests for some of the private, lower level features.
- [x] Document.
- [x] Clean up code.


sile:
Found a typo in the title: s/fANVOVA/fANOVA/

hvy:
A silly mistake. Thanks.

HideakiImamura:
May I ask about the wording of the documentation rather than the contents of the code? 

Why should we use the `RandomSampler` with `FanovaImportanceEvaluator` as recommended in the document: "`It is recommended that trials are not skewed, that is, preferably belonging to a study that was optimized using the `RandomSampler`"? 
Please correct me if I'm wrong but the importance of hyperparameters can be judged by only using observed trial values according to Hutter et al. 2014 (Section 3.3). I think that how hyperparameters are selected is not matter.

hvy:
You might be right as if samples are concentrated near a point, the leafs around this point should be proportionally weighted to their domains https://github.com/optuna/optuna/pull/1106/files#diff-34d5b18389a9c4d35ba97492b7dbc5e6R130-R133. At least, we don't have to mention this in the documentation as the logic mimics the paper. However, I would still expect some degradation as the forest will be less accurate when predicting the leaf midpoints where data points are scarce.

optuna/importance/_fanova/_evaluator.py
```diff
@@ -0,0 +1,123 @@
...
+            search_spaces_is_categorical=search_spaces_is_categorical,
+        )
+
+        individual_importances = {}
+        for i, name in enumerate(distributions.keys()):
+            importance = evaluator.quantify_importance((i,))
+            individual_importance = importance[(i,)]["individual importance"]
+            individual_importances[name] = individual_importance
+
+        tot_importance = sum(v for v in individual_importances.values())
```
HideakiImamura, line 111:
```suggestion
        total_importance = sum(v for v in individual_importances.values())
```
How about renaming the variable like this for clarity?

optuna/importance/_fanova/_evaluator.py
```diff
@@ -0,0 +1,123 @@
...
+        for name in individual_importances.keys():
+            individual_importances[name] /= tot_importance
```
HideakiImamura, line 113:
```suggestion
            individual_importances[name] /= total_importance
```

optuna/importance/_fanova/_stats.py
```diff
@@ -0,0 +1,94 @@
...
+
+        self.weight_stat.push(weight)
+
+        self.avg += delta * weight / self.weight_stat.sum()
+        self.sdm += delta * weight * (x - self.avg)
+
+    def mean(self) -> float:
+        if self.weight_stat.sum() > 0.0:
+            return self.avg
+        return float("nan")
```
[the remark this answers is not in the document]
hvy: How about removing the outermost parenthesis as they're not needed?

optuna/importance/_fanova/_evaluator.py
hvy: I agree that "total" is better. Thanks!

optuna/importance/_fanova/_evaluator.py
```diff
@@ -0,0 +1,121 @@
...
+    .. note::
+
+        Requires the `sklearn <https://github.com/scikit-learn/scikit-learn>`_ Python package.
+
+    .. note::
+
+        Pairwise and higher order importances are not supported through this class. They can be
+        computed using :class:`~optuna.importance._fanova._fanova._Fanova` directly but is not
+        recommended as interfaces may change without prior notice.
+
```
HideakiImamura, line 42: @hvy Thank you for the insightful explanation. I understood from your reply that if the sampled hyperparameters were biased, the random forest would have poor predictive performance, so the sampling method should be changed. I agree with you. Certainly, I think it would be user-friendly to have explained in the documentation. What about adding more explanation as the note like below?

```suggestion
    .. note::
        
        The performance of fANOVA depends on the prediction performance of the underlying 
        random forest model. In order to obtain high prediction performance, it is necessary to 
        cover a wide range of the hyperparameter search space. It is recommended to use an 
        exploration-oriented sampler such as :class:`~optuna.samplers.RandomSampler`.
```

hvy: Thanks, looks good. I applied your suggestion.

optuna/importance/_fanova/_stats.py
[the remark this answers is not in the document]
HideakiImamura: Yes, I agree with you. Thank you for your suggestion.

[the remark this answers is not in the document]
hvy: Would you like to update the suggestion? I could then commit it as is.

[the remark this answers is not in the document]
HideakiImamura: I think either one is fine. Sorry for the confusion.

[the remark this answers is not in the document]
HideakiImamura: OK. I have updated it. Thank you for the notification.

optuna/importance/_fanova/_tree.py
```diff
@@ -0,0 +1,331 @@
...
+        # TODO(hvy): Keep objects `numpy.ndarray`s.
+        prod_midpoints = itertools.product(*midpoints)
+        prod_sizes = itertools.product(*sizes)
+
+        stat = _WeightedRunningStats()
+
+        # For each midpoint along the given dimensions, traverse this tree to compute the
+        # marginal predictions.
+        sample = numpy.full(self._n_features, numpy.nan, dtype=numpy.float64)
+        for i, (midpoints, sizes) in enumerate(zip(prod_midpoints, prod_sizes)):
```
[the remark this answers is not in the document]
hvy: Thanks, let me apply your suggestion.

hvy:
Noticed and fixed a bug with categorical midpoints traversals. Now the `_FanovaTree` doesn't care whether features are categorical or not. Features are one-hot encoded in the forest and the `_FanovaTree` will process the columns as if they were any other numerical column. 

<details>
  <summary>LightGBM, 100 trials, `RandomSampler`</summary>
  
</details>
<details>
  <summary>PyTorch, 100 trials, `RandomSampler`</summary>
  
</details>
<details>
  <summary>ffmpeg, 150 trials, `RandomSampler`</summary>
  
</details>

optuna/importance/_fanova/_fanova.py
```diff
@@ -0,0 +1,262 @@
...
+                for tree_index, tree in enumerate(self._trees):
+                    tree_variance = tree.variance
+                    if tree_variance == 0:
+                        continue
+                    fractions_total.append(
+                        self._V_U_total[sub_features][tree_index] / tree_variance
+                    )
+                    fractions_individual.append(
+                        self._V_U_individual[sub_features][tree_index] / tree_variance
+                    )
```
HideakiImamura, line 101: Would you like to switch the order of `fractions_total` and `fractions_individual`? The other orders are `fractions_individual` -> `fractions_total`.
```suggestion
                    fractions_individual.append(
                        self._V_U_individual[sub_features][tree_index] / tree_variance
                    )
                    fractions_total.append(
                        self._V_U_total[sub_features][tree_index] / tree_variance
                    )
```

HideakiImamura:
Thank you for updating. I checked the changed codes and understood the behavior of`_FanovaTree`, which does not care whether features are categorical or not. LGMTM.

optuna/importance/_fanova/_fanova.py
hvy: Makes sense. I added your suggestion.

optuna/importance/_fanova/_stats.py
```diff
@@ -0,0 +1,92 @@
...
+    def multiply_weights_by(self, a: float) -> "_WeightedRunningStats":
+        return _WeightedRunningStats(self.avg, a * self.sdm, self.weight_stat * a)
+
+    def sum_of_weights(self) -> float:
+        return self.weight_stat.sum()
+
+    def variance_population(self) -> float:
+        return self.divide_sdm_by(self.weight_stat.sum(), 0.0)
+
+    def divide_sdm_by(self, fraction: float, min_weight: float) -> float:
```
ytsmiling, line 89: What is the purpose of the `min_weght` argument? Is that the same with at-least-as-good-as-the-default on the page 5 of the original paper?

optuna/importance/_fanova/_stats.py
```diff
@@ -0,0 +1,92 @@
...
+        return (
+            max(0.0, self.sdm / fraction) if self.weight_stat.sum() > min_weight else float("nan")
```
ytsmiling, line 91: Are there any possibilities that self.sdm or fraction is smaller than zero?
I also do not understand why this function might return `nan`.

optuna/importance/_fanova/_evaluator.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+
+        individual_importances = {}
+        for i, name in enumerate(distributions.keys()):
+            importance = evaluator.quantify_importance((i,))
+            individual_importance = importance[(i,)]["individual importance"]
+            individual_importances[name] = individual_importance
+
+        total_importance = sum(v for v in individual_importances.values())
+        for name in individual_importances.keys():
```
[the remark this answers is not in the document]
hvy: I see. It might be a matter of preference but I personally prefer `key in dict` to check for existence and `key in dict.keys()` for `for` loops.

optuna/importance/_fanova/_fanova.py
```diff
@@ -0,0 +1,262 @@
...
+    from sklearn.preprocessing import OneHotEncoder
+
+    _available = True
+except ImportError as e:
+    _import_error = e
+    _available = False
+
+
+class _Fanova(object):
+    # Implements the fANOVA algorithm in https://github.com/automl/fanova using scikit-learn.
```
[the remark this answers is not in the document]
hvy: In my understanding, documented (docstring) interfaces are public interfaces, and `_Fanova` is private. However, this is not really documented for Optuna and might be open for discussion.

optuna/importance/_fanova/_stats.py
hvy: It might be better just to remove this method.

ytsmiling: Yes, I think that removing this method or the argument helps to keep the code simpler, but I'd like to leave this decision to you.

optuna/importance/_fanova/_fanova.py
```diff
@@ -0,0 +1,262 @@
...
+        self._V_U_total[features] = numpy.empty(n_trees, dtype=numpy.float64)
+
+        raw_features = numpy.concatenate([self._features_to_raw_features[f] for f in features])
+
+        for tree_index, tree in enumerate(self._trees):
+            stat = tree.marginalized_prediction_stat_for_features(raw_features)
+
+            if stat.sum_of_weights() > 0:
+                variance_population = stat.variance_population()
+                V_U_total = variance_population
```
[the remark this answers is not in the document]
ytsmiling: I'd like to hear opinions from other committers on these namings.

optuna/importance/_fanova/_evaluator.py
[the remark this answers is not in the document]
ytsmiling: Understood. If this is intentional, it's totally fine.

optuna/importance/_fanova/_evaluator.py
```diff
@@ -0,0 +1,133 @@
...
+            search_spaces_is_categorical=search_spaces_is_categorical,
+        )
+
+        individual_importances = {}
+        for i, name in enumerate(distributions.keys()):
+            importance = evaluator.quantify_importance((i,))
+            individual_importance = importance[(i,)]["individual importance"]
+            individual_importances[name] = individual_importance
+
+        total_importance = sum(v for v in individual_importances.values())
```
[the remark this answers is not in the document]
hvy: Fixed.

optuna/importance/_fanova/_fanova.py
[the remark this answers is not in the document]
hvy: Changes to lower case but yes there might be better names for these dictionaries.

optuna/importance/_fanova/_stats.py
```diff
@@ -0,0 +1,92 @@
+class _RunningStats(object):
+    def __init__(self, N: int = 0, avg: float = 0.0, sdm: float = 0.0) -> None:
```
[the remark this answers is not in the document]
hvy: Added a comment.

optuna/importance/_fanova/_stats.py
hvy: Simplified the code based on how it's used internally.

optuna/importance/_fanova/_stats.py
hvy: As it's used now, you're right. I removed this method and simplified `variance_population`.

hvy:
Thanks for the comments, they should be addressed now.

hvy:
The last commit https://github.com/optuna/optuna/pull/1106/commits/abed879915a5c2d8be24f1c107593b7362f0edbf has `_WeightedRunningStats` removed and uses `numpy.ndarray`s instead . 

Evaluating single and pairwise importances with my PyTorch example (including several categorical features which are more computationally demanding in Optuna's implementation due to the one-hot sparse encoding) is now roughly twice as fast. 

#### Before 

```
_Fanova: 1.581s 
fANOVA: 0.1682s 
```
_Average time taken per feature(s) (single and pairwise). 
`_Fanova` is Optuna's, `fANOVA` is AutoML's._

#### After 
```
_Fanova: 0.8128s
fANOVA: 0.1648s 
```

_Verified that evaluated importances are the same as before (https://github.com/optuna/optuna/pull/1106#issuecomment-614544840)._

#### Note

I think further performance improvements require e.g. changes to the marginal prediction tree traversal as that's now the most significant bottleneck associated with the Python runtime overhead. Also cleaned up some interfaces and implementation details.

optuna/importance/_fanova/_fanova.py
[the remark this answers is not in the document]
hvy: Removed "V_U_total" altogether as "V_U_individual" suffices. Also renamed the latter to "variances".

optuna/importance/_fanova/_stats.py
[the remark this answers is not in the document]
hvy: Removed this file and belonging classes in https://github.com/optuna/optuna/pull/1106/commits/5ca00b6ff8c6f2573ca30b94007e8451cc74f78d as they were really inefficient and creating a bottleneck.

optuna/importance/_fanova/_stats.py
hvy: Deleted this file.

optuna/importance/_fanova/_stats.py
hvy: Deleted this file.

optuna/importance/_fanova/_tree.py
```diff
@@ -0,0 +1,313 @@
...
+            sizes.append(size)
+
+        return midpoints, sizes
+
+    def _compute_features_split_values(self) -> List[numpy.ndarray]:
+        all_split_values = [set() for _ in range(self._n_features)]  # type: List[Set[float]]
+
+        for node_index in range(self._n_nodes):
+            feature = self._get_node_split_feature(node_index)
+            if feature >= 0:  # Not leaf.
```
ytsmiling, line 219: nit: Ditto.
```suggestion
            if not self._is_node_leaf(node_index):
                feature = self._get_node_split_feature(node_index)
```

optuna/importance/_fanova/_tree.py
```diff
@@ -0,0 +1,313 @@
...
+
+        node_indices = []
+        active_features_cardinalities = []
+
+        while len(active_nodes) > 0:
+            node_index = active_nodes.pop()
+            search_spaces = active_search_spaces.pop()
+
+            feature = self._get_node_split_feature(node_index)
+            if feature >= 0:  # Not leaf.
```
ytsmiling, line 109: nit: Using `_is_node_leaf` will slightly improve readability.
```suggestion
            if not self._is_node_leaf(node_index):
                feature = self._get_node_split_feature(node_index)
```

hvy: That's a perfectly sensible comment. This is actually in favor of performance. It contributes to around 10% of the gain in https://github.com/optuna/optuna/pull/1106#issuecomment-619388187.

hvy: Added comments about intention https://github.com/optuna/optuna/pull/1106/commits/19fc7b4c834121416f84d5d1537a5e4f9583cd8f. What do you think?

hvy:
Resolved conflicts and added references to the original work. PTAL.

PR #1108, opened 2020-04-09
Use latest `sphinx` version on RTD.

hvy:
An attempt to render parameter types in the documentation by using a newer version of `sphinx` on Read the Docs. I followed the note in https://docs.readthedocs.io/en/stable/builds.html#understanding-what-s-going-on but not really sure if this'll work.


hvy:
The RTD settings must then be updated https://docs.readthedocs.io/en/stable/guides/specifying-dependencies.html#specifying-a-requirements-file.

docs/requirements.txt
```diff
@@ -0,0 +1 @@
+sphinx==3.0.0
```
sile, line 1: How about allowing v3.0.0 or later?
```suggestion
sphinx>=3.0.0
```

hvy: That makes sense, thanks.

toshihikoyanase:
The parameter types are rendered successfully :tada:
image

hvy:
Just leaving a note but we didn't have to update the path in the RTD dashboard (as mentioned here https://github.com/optuna/optuna/pull/1108#issuecomment-611342241). Maybe because the default path is being used.

PR #1111, opened 2020-04-10
`single()` returns True when `step` or `q` is greater than `high-low`.

c-bata:
When step or q is greater than `high-low`, we can consider it as a single distribution.

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -217,7 +217,11 @@ class DiscreteUniformDistribution(BaseDistribution):
     def single(self):
         # type: () -> bool
 
-        return self.low == self.high
+        if self.low == self.high:
+            return True
+        if (self.high - self.low) < self.q:
+            return True
+        return False
 
     def _contains(self, param_value_in_internal_repr):
         # type: (float) -> bool
@@ -271,7 +275,11 @@ class IntUniformDistribution(BaseDistribution):
     def single(self):
         # type: () -> bool
 
-        return self.low == self.high
+        if self.low == self.high:
+            return True
+        if (self.high - self.low) < self.step:
+            return True
+        return False
 
     def _contains(self, param_value_in_internal_repr):
         # type: (float) -> bool
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -1,4 +1,5 @@
 import abc
+import decimal
 import json
 import warnings
 
@@ -219,7 +220,10 @@ class DiscreteUniformDistribution(BaseDistribution):
 
         if self.low == self.high:
             return True
-        if (self.high - self.low) < self.q:
+        high = decimal.Decimal(str(self.high))
+        low = decimal.Decimal(str(self.low))
+        q = decimal.Decimal(str(self.q))
+        if (high - low) < q:
             return True
         return False
 
--- a/tests/test_distributions.py
+++ b/tests/test_distributions.py
@@ -187,7 +187,9 @@ def test_single():
         distributions.UniformDistribution(low=1.0, high=1.0),
         distributions.LogUniformDistribution(low=7.3, high=7.3),
         distributions.DiscreteUniformDistribution(low=2.22, high=2.22, q=0.1),
+        distributions.DiscreteUniformDistribution(low=2.22, high=2.24, q=0.3),
         distributions.IntUniformDistribution(low=-123, high=-123),
+        distributions.IntUniformDistribution(low=-123, high=-120, step=4),
         distributions.CategoricalDistribution(choices=("foo",)),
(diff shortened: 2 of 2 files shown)


optuna/distributions.py
```diff
@@ -217,7 +217,11 @@ def __init__(self, low, high, q):
     def single(self):
         # type: () -> bool
 
-        return self.low == self.high
+        if self.low == self.high:
+            return True
+        if (self.high - self.low) < self.q:
+            return True
+        return False
```
c-bata, line 224: Now I'm considering to close this pull request if we should support the following case.

```python
>>> single(0.2, 0.1, 0.1)
False
>>> single(0.3, 0.2, 0.1)
True
>>> 0.3 - 0.2
0.09999999999999998
```

c-bata: In the discussion of core-devs, we decided to use decimal module.

c-bata:
It might be enough just adds `1e-10` or something small number:

```python
>>> def single(high, low, step):
...     return (high - low + 1e-10) < step
... 
>>> single(0.2, 0.1, 0.1)
False
>>> single(0.3, 0., 0.1)
False
```

hvy:
A drop by comment but I'm wondering if existing samplers really work under these settings (e.g. `count` in the `SkoptSampler` https://github.com/optuna/optuna/blob/master/optuna/integration/skopt.py#L187) 🤔

hvy:
I think `Decimal` is acceptable in terms of performance as it's already used [here](https://github.com/optuna/optuna/blob/master/optuna/trial.py#L1324).

c-bata:
> A drop by comment but I'm wondering if existing samplers really work under these settings (e.g. `count` in the `SkoptSampler` https://github.com/optuna/optuna/blob/master/optuna/integration/skopt.py#L187) 🤔

Yeah, I didn't check it. This PR might be `bug-fix`.

tests/test_distributions.py
```diff
@@ -197,7 +199,11 @@ def test_single():
         distributions.UniformDistribution(low=1.0, high=1.001),
         distributions.LogUniformDistribution(low=7.3, high=10),
         distributions.DiscreteUniformDistribution(low=-30, high=-20, q=2),
+        distributions.DiscreteUniformDistribution(low=-30, high=-20, q=10),
+        # In Python, "0.3 - 0.2 != 0.1" is True.
+        distributions.DiscreteUniformDistribution(low=0.2, high=0.3, q=0.1),
```
hvy, line 204: We could add one more case. But it might be overkill, please feel free to ignore it.
```suggestion
        distributions.DiscreteUniformDistribution(low=0.2, high=0.3, q=0.1),
        distributions.DiscreteUniformDistribution(low=0.7, high=0.8, q=0.1),
```
since
```
>>> 0.8 - 0.7
0.10000000000000009
>>> 0.3 - 0.2
0.09999999999999998
```

c-bata, line 204: @hvy How about this?

```suggestion
        # In Python, "0.8 - 0.7 == 0.1` is False but "0.3 - 0.2 != 0.1" is True.
        distributions.DiscreteUniformDistribution(low=0.7, high=0.8, q=0.1),
        distributions.DiscreteUniformDistribution(low=0.2, high=0.3, q=0.1),
```

hvy: Should it not be "and" instead of "but"? And a detail but I think you places a backtick by mistake 🙂

hvy:
> > A drop by comment but I'm wondering if existing samplers really work under these settings (e.g. count in the SkoptSampler https://github.com/optuna/optuna/blob/master/optuna/integration/skopt.py#L187) 🤔

> Yeah, I didn't check it. This PR might be bug-fix.

I'm sorry, it's probably not in the scope of this PR as it's the responsibility of the sampler.

tests/test_distributions.py
c-bata: Thank you for pointing it out. I updated a comment.

c-bata:
> Let us work on the sampler's issue (#1111 (comment)) in a new PR.

After merged this pull request, https://github.com/optuna/optuna/blob/master/optuna/integration/skopt.py#L187 isn't called by these settings because single distribution is skipped at `infer_relative_searchspace()`.

toshihikoyanase:
@c-bata Thank you for your reply.

I thought @hvy was also concerned about non-single cases. For example, we may need to fix the following cases.

```python
>>> int((0.3 - 0.1) // 0.1)  # current implementation
1
>>> int((Decimal("0.3") - Decimal("0.1")) // Decimal("0.1"))  # to be ?
2
```

c-bata:
Oh, I see! Thank you.

PR #1114, opened 2020-04-12
Return trial_id at `study._append_trial()`.

c-bata:
It is sometimes useful if `_append_trial()` returns `trial_id` (e.g. https://github.com/c-bata/optuna-memorydump).

### DIFF
--- a/tests/test_study.py
+++ b/tests/test_study.py
@@ -713,8 +713,7 @@ def test_append_trial(storage_mode):
         assert len(study.trials) == 0
 
         trial_id = study._append_trial(value=0.8)
-        if isinstance(study._storage, optuna.storages.InMemoryStorage):
-            assert trial_id == 0
+        assert study.trials[0]._trial_id == trial_id
         assert len(study.trials) == 1
         assert study.best_value == 0.8
 
--- a/tests/test_study.py
+++ b/tests/test_study.py
@@ -712,7 +712,7 @@ def test_append_trial(storage_mode):
         study = optuna.create_study(storage=storage)
         assert len(study.trials) == 0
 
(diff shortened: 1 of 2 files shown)


c-bata:
@crcrpar Exactly! thank you for your quick review 👍

tests/test_study.py
```diff
@@ -712,7 +712,9 @@ def test_append_trial(storage_mode):
         study = optuna.create_study(storage=storage)
         assert len(study.trials) == 0
 
-        study._append_trial(value=0.8)
+        trial_id = study._append_trial(value=0.8) == 0
+        if isinstance(study._storage, optuna.storages.InMemoryStorage):
+            assert trial_id == 0
```
c-bata, line 717: We can also check a trial id on RDBStorage like the following:

```python
        if isinstance(study._storage, optuna.storages.InMemoryStorage):
            assert study._append_trial(value=0.8) == 0
        else:
            assert study._append_trial(value=0.8) == 1
```

But I only checked trial_id when using InMemoryStorage. As far as I know, it is an anti-pattern to depend on the primary key (e.g. Django's default test case doesn't reset the sequences for some performance reasons. [refs django.test.TransactionTestCase.reset_sequences](https://docs.djangoproject.com/en/3.0/topics/testing/advanced/#django.test.TransactionTestCase.reset_sequences)).

c-bata:
PTAL.

tests/test_study.py
sile: I think that since there is only one trial we can check the correctness of the trial id as follows:
```python
assert study.trials[0]._trial_id == trial_id
```

c-bata: It's a good idea 👍

PR #1121, opened 2020-04-15
Replace `mock` with `unittest.mock`.

hvy:
`mock` is in the standard library starting from Python 3.3 meaning we should prefer that `mock` over the third-party `mock`. C.f. https://mock.readthedocs.io/en/latest/.


hvy:
Thanks. Yes maybe we'll have to put an upper bound on the version if it starts to break, until we also drop 3.5.

PR #1125, opened 2020-04-16
Add the pruning feature to `LightGBMTuner`.

toshihikoyanase:
This PR enables `LightGBMTuner` to prune unpromising trials using `optuna.pruners`.
Currently, pruning is activated if `LightGBMTuner.study.pruner` is not `NopPruner`. The default pruner created in the `LightGBMTuner.__init__` is `NopPruner`, and the pruning is disabled by default.

I have a concern about user-given studies. If users create LightGBMTuner with `LightGBMTuner(study=optuna.create_study())`, the `MedianPruner` will be used. I feel a bit tricky because the pruning feature is disabled by default in other usages. If you have any comments or suggestions, please let me know.

Please merge this branch after #1076.

### DIFF
--- a/optuna/integration/lightgbm_tuner/optimize.py
+++ b/optuna/integration/lightgbm_tuner/optimize.py
@@ -98,6 +98,13 @@ class BaseTuner(object):
             metric = list(metric)[-1]
         else:
             raise NotImplementedError
+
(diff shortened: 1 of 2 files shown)


toshihikoyanase:
Let me close this PR to revise the interface of the pruning.

PR #1132, opened 2020-04-17
Run `actions/stale` on weekday mornings Tokyo time.

hvy:
Limited the bot to run on weekdays only. Also, changed the time from midnight to morning (8:00, Tokyo). We might not want to have issues and PR "states" being changed by a bot during weekends when there's less work.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -2,7 +2,7 @@ name: stale
 
 on:
   schedule:
-    - cron: '0 23 * * 7,1,2,3,4'
+    - cron: '0 23 * * SUN-THU'
 
 jobs:
   stale:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -2,7 +2,7 @@ name: stale
 
 on:
   schedule:
-    - cron: '0 15 * * *'
+    - cron: '0 23 * * 7,1,2,3,4'
```
keisuke-umezawa, line 5: FYI
It also works.

```
'0 23 * * SUN,MON,TUE,WED,THU'
```

```
'0 23 * * SUN-THU'
```

hvy: I applied your second suggestion. Thank you!

hvy:
PTAL.

PR #1135, opened 2020-04-18
Remove unnecessary deep-copies.

ytsmiling:
This PR removes unnecessary deep-copies. This PR significantly speeds up in-memory execution, which is the default setting of optuna.

Microbench with 1000 trials and 30 params (CPU: Intel Core i7-6700K, python: 3.5.2):
```
Master:
         1546802373 function calls (1310307516 primitive calls) in 627.117 seconds
This PR:
         234867048 function calls (211661688 primitive calls) in 170.741 seconds
```

Code:
```python
import argparse
import cProfile
import math

import optuna


def objective(trial):
    return sum(
        math.sin(trial.suggest_float('param-{}'.format(i), 0, math.pi * 2))
        for i in range(30)
    )


if __name__ == '__main__':
    args = parser.parse_args()
    sampler = optuna.samplers.TPESampler(consider_prior=False)
    cProfile.run(
        "study.optimize(objective, n_trials=args.n_trial, gc_after_trial=False)",
        sort="cumtime"
    )
```

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -465,7 +465,7 @@ class Study(BaseStudy):
 
         _check_pandas_availability()
 
-        trials = self.trials
+        trials = self.get_trials(deepcopy=False)
 
         # If no trials, return an empty dataframe.
         if not len(trials):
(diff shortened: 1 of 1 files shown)


ytsmiling:
Related PR: #1139.

Issue #622, opened 2019-10-18
Add an FAQ page about `MySQL lost connection error`.


toshihikoyanase:
MySQL lost connection errors was fixed by the #504, but the instruction about `RDBStorage` argument (e.g., `{'pool_recycle': 30}`) is not documented. Please create an FAQ page about it. I think it is better to be explained somewhere in tutorials (e.g., [Saving/Resuming Study with RDB Backend](https://optuna.readthedocs.io/en/stable/tutorial/rdb.html)) as well.

toshihikoyanase:
This issue will be resolved by #806, and we don't need to add the document mentioned in the above comment. But, it requires time to examine performance degradation due to the option introduced by #806.

For meanwhile, please provide the option to avoid connection errors here.

## Problem

When I use the MySQL server as the backend of the storage, I got `pymysql.err.OperationalError: (2013, 'Lost connection to MySQL server during query')`. 

**Code**

```python
import optuna
import time

url = 'mysql+pymysql://...'

def objective(trial):
    return trial.suggest_int('x', -10, 10)

study = optuna.create_study(storage=url)
study.optimize(objective, n_trials=1)
print('Best value: {} (params: {})\n'.format(study.best_value, study.best_params))

time.sleep(70)
study.optimize(objective, n_trials=1)  # The connection will be lost here.
```

## Workaround

Please set `engine_kwargs={'pool_pre_ping': True}` to `RDBStorage.__init__`. When we enable the `pool_pre_ping` option, SQL clients send `ping` to database servers to check connections are still viable before sending actual queries. If the connections are lost, the clients automatically reconnect to the servers. For further details of `pool_pre_ping` option, please consult [this SQLAlchemy document](https://docs.sqlalchemy.org/en/13/core/pooling.html#disconnect-handling-pessimistic).

**Code**

```python
import optuna
import time

url = 'mysql+pymysql://...'

def objective(trial):
    return trial.suggest_int('x', -10, 10)

storage = optuna.storages.RDBStorage(url, {'pool_pre_ping': True})  # Enable pre_ping here.
study = optuna.create_study(storage=storage)
study.optimize(objective, n_trials=1)
print('Best value: {} (params: {})\n'.format(study.best_value, study.best_params))

time.sleep(70)
study.optimize(objective, n_trials=1)  # This will successfully complete.
```

toshihikoyanase:
#806 resolved this issue. Let me close it.

PR #1140, opened 2020-04-20
Add storage cache

ytsmiling:
## Motivation
Flushing updates of trials to persistent storages on every update (including parameter suggests) can be a significant bottleneck. This PR mitigates the bottleneck by introducing a wrapper class that wraps the `RDBStorage/ class and providing the lazy-sync capability.

## Description of the changes
- Add `CachedStorage`, which is a wrapper of `RDBStorage` and provide the lazy-sync capability.
    - While `CachedStorage` only supports `RDBStorage`, it is straightforward to support other classes such as `RedisStorage`. Thus, `CachedStorage` file is placed directly under the `optuna/storages` class.
- Introduce two additional public methods to `RDBStorage`.
    - One returns `FrozenTrial` when creating a new trial. Note, `create_new_trial` in `BaseStorage` only returns the trial-id.
    - The other receives a `FrozenTrial` and flush the updates to persistent storages.

## Current Status
- [x] Implement `CachedStorage` class.
- [x] Add tests.
- [x] Add docs where necessary.

## Microbench
(DB: official mysql docker image (tag:8.0.19), storage: SSD, CPU: Intel Core i7-6700K, python: 3.5.2)
- perf:optimize-study: time taken for a single optimization

### This PR
```
perf:optimize-study                               0:00:10.169289
perf:optimize-study                               0:00:47.566834
```

### Current master
```
perf:optimize-study                               0:00:22.499407
perf:optimize-study                               0:02:33.151298
```

Benchmark code:
```python
import argparse
from datetime import datetime
import math

import sqlalchemy
import optuna

profile_result = defaultdict(lambda: datetime.now() - datetime.now())


class Profile:
    def __init__(self, name):
        self.name = name

    def __enter__(self):
        self.start = datetime.now()

        profile_result[self.name] += datetime.now() - self.start


def print_profile():
    for key, value in sorted(profile_result.items(), key=lambda i: i[1],
                             reverse=True):
        print(key.ljust(50) + '{}'.format(value))


    def objective(trial):
        return sum([
            math.sin(trial.suggest_float('param-{}'.format(i), 0, math.pi * 2))
            for i in range(n_param)
        ])

    return objective


def define_flags(parser):
    parser.add_argument('mysql_user', type=str)
    parser.add_argument('mysql_password', type=str)
    parser.add_argument('mysql_host', type=str)
    parser.add_argument('n_study', type=int)
    parser.add_argument('n_param', type=int)
    return parser


if __name__ == '__main__':
    parser = define_flags(argparse.ArgumentParser())
    args = parser.parse_args()
    print(vars(args))

    storage_str = 'mysql+pymysql://{}:{}@{}/'.format(
        args.mysql_user,
        args.mysql_password,
        args.mysql_host,
    )
    engine = sqlalchemy.create_engine(storage_str)
    sampler = optuna.samplers.TPESampler()
    conn = engine.connect()
    conn.execute("commit")
    database_str = 'profile_storage_s{}_t{}_p{}'.format(
        args.n_study, args.n_trial, args.n_param)

    try:
        conn.execute(
            "drop database {}".format(database_str))
    except:
        pass
    conn.execute("create database {}".format(
        database_str
    ))
    conn.close()

    for i in range(args.n_study):
        storage = optuna.storages.get_storage(storage_str + database_str)
        study = optuna.create_study(sampler=sampler, storage=storage)
        study_id = study.study_id
        with Profile('perf:optimize-study'):
            study.optimize(build_objective_fun(args.n_param), n_trials=args.n_trial, gc_after_trial=False)

    print_profile()

```

### DIFF
--- a/optuna/storages/__init__.py
+++ b/optuna/storages/__init__.py
@@ -3,8 +3,8 @@ from typing import Union  # NOQA
 from optuna.storages.base import BaseStorage  # NOQA
 from optuna.storages.cached_storage import _CachedStorage
 from optuna.storages.in_memory import InMemoryStorage
-from optuna.storages.redis import RedisStorage
 from optuna.storages.rdb.storage import RDBStorage
+from optuna.storages.redis import RedisStorage
 
 
 def get_storage(storage):
--- a/optuna/storages/__init__.py
+++ b/optuna/storages/__init__.py
@@ -1,9 +1,10 @@
-from optuna.storages.base import BaseStorage  # NOQA
-from optuna.storages.in_memory import InMemoryStorage  # NOQA
-from optuna.storages.redis import RedisStorage  # NOQA
-from optuna.storages.rdb.storage import RDBStorage  # NOQA
-
 from typing import Union  # NOQA
+
+from optuna.storages.base import BaseStorage  # NOQA
+from optuna.storages.cached_storage import _CachedStorage
+from optuna.storages.in_memory import InMemoryStorage
+from optuna.storages.redis import RedisStorage
+from optuna.storages.rdb.storage import RDBStorage
 
 
 def get_storage(storage):
@@ -15,6 +16,8 @@ def get_storage(storage):
         if storage.startswith("redis"):
             return RedisStorage(storage)
         else:
-            return RDBStorage(storage)
+            return _CachedStorage(RDBStorage(storage))
+    elif isinstance(storage, RDBStorage):
+        return _CachedStorage(storage)
     else:
         return storage
--- a/optuna/storages/cached_storage.py
+++ b/optuna/storages/cached_storage.py
@@ -15,7 +15,7 @@ from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
 
 
-class _TrialUpdates:
+class _TrialUpdate:
     def __init__(self) -> None:
         self.state = None  # type: Optional[TrialState]
         self.value = None  # type: Optional[float]
@@ -29,7 +29,7 @@ class _TrialUpdates:
 class _StudyInfo:
     def __init__(self) -> None:
         self.trials = dict()  # type: Dict[int, FrozenTrial]
-        self.updates = dict()  # type: Dict[int, _TrialUpdates]
+        self.updates = dict()  # type: Dict[int, _TrialUpdate]
         self.param_distribution = {}  # type: Dict[str, distributions.BaseDistribution]
 
 
@@ -267,12 +267,12 @@ class _CachedStorage(base.BaseStorage):
         study_id, number = self._trial_id_to_study_id_and_number[trial_id]
         return self._studies[study_id].trials.get(number, None)
 
-    def _get_updates(self, trial_id: int) -> _TrialUpdates:
+    def _get_updates(self, trial_id: int) -> _TrialUpdate:
         study_id, number = self._trial_id_to_study_id_and_number[trial_id]
         updates = self._studies[study_id].updates.get(number, None)
         if updates is not None:
             return updates
-        updates = _TrialUpdates()
+        updates = _TrialUpdate()
         self._studies[study_id].updates[number] = updates
         return updates
 
--- a/optuna/storages/cached_storage.py
+++ b/optuna/storages/cached_storage.py
@@ -174,10 +174,14 @@ class _CachedStorage(base.BaseStorage):
                     self._studies[study_id].param_distribution[param_name] = distribution
                 if param_name in cached_trial.params:
                     return False
-                cached_trial.params[param_name] = distribution.to_external_repr(
+                params = copy.copy(cached_trial.params)
+                params[param_name] = distribution.to_external_repr(
                     param_value_internal
                 )
-                cached_trial.distributions[param_name] = distribution
+                cached_trial.params = params
+                dists = copy.copy(cached_trial.distributions)
+                dists[param_name] = distribution
+                cached_trial.distributions[param_name] = dists
                 updates.params[param_name] = param_value_internal
                 updates.distributions[param_name] = distribution
                 return True
@@ -222,7 +226,9 @@ class _CachedStorage(base.BaseStorage):
                 updates = self._get_updates(trial_id)
                 if step in cached_trial.intermediate_values:
                     return False
-                cached_trial.intermediate_values[step] = intermediate_value
+                values = copy.copy(cached_trial.intermediate_values)
+                values[step] = intermediate_value
+                cached_trial.intermediate_values = values
                 updates.values[step] = intermediate_value
                 self._flush_trial(trial_id)
                 return True
@@ -235,7 +241,9 @@ class _CachedStorage(base.BaseStorage):
             cached_trial = self._get_cached_trial(trial_id)
             if cached_trial is not None:
                 updates = self._get_updates(trial_id)
-                cached_trial.user_attrs[key] = value
+                attrs = copy.copy(cached_trial.user_attrs)
+                attrs[key] = value
+                cached_trial.user_attrs = attrs
                 updates.user_attrs[key] = value
                 self._flush_trial(trial_id)
                 return
@@ -248,7 +256,9 @@ class _CachedStorage(base.BaseStorage):
             cached_trial = self._get_cached_trial(trial_id)
             if cached_trial is not None:
                 updates = self._get_updates(trial_id)
-                cached_trial.system_attrs[key] = value
+                attrs = copy.copy(cached_trial.system_attrs)
+                attrs[key] = value
+                cached_trial.system_attrs = attrs
                 updates.system_attrs[key] = value
                 self._flush_trial(trial_id)
                 return
--- a/optuna/storages/cached_storage.py
+++ b/optuna/storages/cached_storage.py
@@ -50,14 +50,12 @@ class _CachedStorage(base.BaseStorage):
         self._trial_id_to_study_id_and_number = dict()  # type: Dict[int, Tuple[int, int]]
         self._lock = threading.RLock()
 
-    def __getstate__(self):
-        # type: () -> Dict[Any, Any]
+    def __getstate__(self) -> Dict[Any, Any]:
         state = self.__dict__.copy()
         del state["_lock"]
         return state
 
-    def __setstate__(self, state):
-        # type: (Dict[Any, Any]) -> None
+    def __setstate__(self, state: Dict[Any, Any]) -> None:
         self.__dict__.update(state)
         self._lock = threading.RLock()
 
--- a/optuna/storages/cached_storage.py
+++ b/optuna/storages/cached_storage.py
@@ -298,11 +298,6 @@ class _CachedStorage(base.BaseStorage):
 
         return self._backend.get_n_trials(study_id, state)
 
-    def flush(self) -> None:
-        for study in self._studies.values():
-            for trial_id in study.updates:
-                self._flush_trial(trial_id)
-
     def _flush_trial(self, trial_id: int) -> bool:
         if trial_id not in self._trial_id_to_study_id_and_number:
             # The trial has not been managed by this class.
--- a/optuna/storages/cached_storage.py
+++ b/optuna/storages/cached_storage.py
@@ -1,0 +1,318 @@
+import copy
+import threading
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
+
+from optuna import distributions
+from optuna.storages import base
+from optuna.storages.rdb.storage import RDBStorage
+from optuna.study import StudyDirection
+from optuna.study import StudySummary
+from optuna.trial import FrozenTrial
+from optuna.trial import TrialState
+
+
+class _TrialUpdates:
+    def __init__(self) -> None:
+        self.state = None  # type: Optional[TrialState]
+        self.value = None  # type: Optional[float]
+        self.values = dict()  # type: Dict[int, float]
+        self.user_attrs = dict()  # type: Dict[str, Any]
+        self.system_attrs = dict()  # type: Dict[str, Any]
+        self.params = dict()  # type: Dict[str, Any]
+        self.distributions = dict()  # type: Dict[str, distributions.BaseDistribution]
+
+
+class _StudyInfo:
+    def __init__(self) -> None:
+        self.trials = dict()  # type: Dict[int, FrozenTrial]
+        self.updates = dict()  # type: Dict[int, _TrialUpdates]
+        self.param_distribution = {}  # type: Dict[str, distributions.BaseDistribution]
+
+
+class _CachedStorage(base.BaseStorage):
+    """A wrapper class of storage backends.
+
+    This class is used in :func:`~optuna.get_storage` function and automatically
+    wraps :class:`~optuna.storages.RDBStorage` class.
+
+    Args:
+        storage:
+            BaseStorage class instance to wrap.
+    """
+
+    def __init__(self, backend: RDBStorage) -> None:
+        self._backend = backend
+        self._studies = {}  # type: Dict[int, _StudyInfo]
+        self._trial_id_to_study_id_and_number = dict()  # type: Dict[int, Tuple[int, int]]
+        self._lock = threading.RLock()
+
+    def __getstate__(self):
+        # type: () -> Dict[Any, Any]
+        state = self.__dict__.copy()
+        del state["_lock"]
+        return state
+
+    def __setstate__(self, state):
+        # type: (Dict[Any, Any]) -> None
+        self.__dict__.update(state)
+        self._lock = threading.RLock()
+
+    def create_new_study(self, study_name: Optional[str] = None) -> int:
+
+        study_id = self._backend.create_new_study(study_name)
+        self._studies[study_id] = _StudyInfo()
+        return study_id
+
+    def delete_study(self, study_id: int) -> None:
+
+        with self._lock:
+            if study_id in self._studies:
+                for trial_id in self._studies[study_id].trials:
+                    trial_ids_to_delete = []
+                    if trial_id in self._trial_id_to_study_id_and_number:
+                        trial_ids_to_delete.append(trial_id)
+                    for trial_id in trial_ids_to_delete:
+                        del self._trial_id_to_study_id_and_number[trial_id]
+                del self._studies[study_id]
+
+        self._backend.delete_study(study_id)
+
+    def set_study_direction(self, study_id: int, direction: StudyDirection) -> None:
+
+        self._backend.set_study_direction(study_id, direction)
+
+    def set_study_user_attr(self, study_id: int, key: str, value: Any) -> None:
+
+        self._backend.set_study_user_attr(study_id, key, value)
+
+    def set_study_system_attr(self, study_id: int, key: str, value: Any) -> None:
+
+        self._backend.set_study_system_attr(study_id, key, value)
+
+    def get_study_id_from_name(self, study_name: str) -> int:
+
+        return self._backend.get_study_id_from_name(study_name)
+
+    def get_study_id_from_trial_id(self, trial_id: int) -> int:
+
+        return self._backend.get_study_id_from_trial_id(trial_id)
+
+    def get_study_name_from_id(self, study_id: int) -> str:
+
+        return self._backend.get_study_name_from_id(study_id)
+
+    def get_study_direction(self, study_id: int) -> StudyDirection:
+
+        return self._backend.get_study_direction(study_id)
+
+    def get_study_user_attrs(self, study_id: int) -> Dict[str, Any]:
+
+        return self._backend.get_study_user_attrs(study_id)
+
+    def get_study_system_attrs(self, study_id: int) -> Dict[str, Any]:
+
+        return self._backend.get_study_system_attrs(study_id)
+
+    def get_all_study_summaries(self) -> List[StudySummary]:
+
+        return self._backend.get_all_study_summaries()
+
+    def create_new_trial(self, study_id: int, template_trial: Optional[FrozenTrial] = None) -> int:
+
+        trial_id, frozen_trial = self._backend._create_new_trial_with_trial(
+            study_id, template_trial
+        )
+        with self._lock:
+            if study_id not in self._studies:
+                self._studies[study_id] = _StudyInfo()
+            self._trial_id_to_study_id_and_number[trial_id] = study_id, frozen_trial.number
+            if frozen_trial.state == TrialState.RUNNING:
+                self._studies[study_id].trials[frozen_trial.number] = frozen_trial
+            return trial_id
+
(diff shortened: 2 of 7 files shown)


ytsmiling:
After fixing some bugs, I updated the microbenchmark.
```
perf:optimize-study                               0:00:45.771852
perf:optimize-study                               0:00:13.391903
```

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,287 @@
...
+        param_name: str,
+        param_value_internal: float,
+        distribution: distributions.BaseDistribution,
+    ) -> bool:
+
+        with self._lock:
+            if trial_id in self._hot_cache:
+                cached_trial, updates = self._hot_cache[trial_id]
+                study_id = self._trial_id_to_study_id[trial_id]
+                cached_dist = self._param_distributions[study_id].get(param_name, None)
```
sile, line 144: Since `RDBStorage` performs the distribution compatibility check only at trial-level (see below), it seems that `CachedStorage` also can do it the same level and remove the `_param_distributions` field.
https://github.com/optuna/optuna/blob/5c07d0afffeb506866860fa2ca21a71a4199c075/optuna/storages/rdb/storage.py#L547-L549

```suggestion
                cached_dist = cached_trial.distributions.get(param_name, None)
```

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def set_trial_system_attr(self, trial_id: int, key: str, value: Any) -> None:
+
+        with self._lock:
+            if trial_id in self._hot_cache:
+                cached_trial, updates = self._hot_cache[trial_id]
+                self._dirty_trials.add(trial_id)
+                cached_trial.system_attrs[key] = value
+                updates.system_attrs[key] = value
+                return
```
sile, line 239: Since `SuccessiveHalvingPruner` accesses to the system attributes of other trials (including RUNNING ones), it seems needed to delete this block to keep compatibility with the existing behavior.
https://github.com/optuna/optuna/blob/5c07d0afffeb506866860fa2ca21a71a4199c075/optuna/pruners/successive_halving.py#L221

optuna/storages/rdb/storage.py
```diff
@@ -437,6 +435,25 @@ def get_all_study_summaries(self) -> List[StudySummary]:
     def create_new_trial(self, study_id, template_trial=None):
         # type: (int, Optional[FrozenTrial]) -> int
 
+        return self.create_new_trial_with_trial(study_id, template_trial)[0]
+
+    def create_new_trial_with_trial(
```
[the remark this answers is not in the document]
ytsmiling: In that sense, all methods should have `_` prefix.
If this class turned out to be successful, I think it's beneficial to add this method to `BaseStorage` class interface and make this cache class available for all other storage backends.
Btw, I think this method should have a better name, which I'll address later.

optuna/storages/rdb/storage.py
```diff
@@ -437,6 +435,25 @@ def get_all_study_summaries(self) -> List[StudySummary]:
...
+        self, study_id: int, template_trial: Optional[FrozenTrial] = None
+    ) -> Tuple[int, FrozenTrial]:
```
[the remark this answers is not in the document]
ytsmiling: I'd like to avoid accessing the private attribute here.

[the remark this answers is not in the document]
sile: Okay. I'll mark this conversation as resolved.

optuna/storages/rdb/storage.py
```diff
@@ -488,11 +505,122 @@ def create_new_trial(self, study_id, template_trial=None):
...
+            )
+
         session.add(trial)
 
         self._commit(session)
 
-        return trial.trial_id
+        return trial.trial_id, frozen
+
+    def update_trial(
```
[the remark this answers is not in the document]
ytsmiling: Do you think it's better to make this method private this time and make it public when it becomes a member of the `BaseStorage` interface? If not, I'll leave this as is.

optuna/storages/cached_storage.py
ytsmiling: The model class checks the distribution compatibility in the study level. (And both of current PR and your suggestion are incompatible with the current implementation.)

optuna/storages/rdb/storage.py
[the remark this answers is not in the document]
sile: > all methods should have `_` prefix.

Yes, I basically think so about the methods of `RDBStorage` (it's not supposed to be accessed by the library users directly). An exception is the methods defined by `BaseStorage`, so if this method is added to the interface of the base class, we will be able to remove the `_` prefix at that timing.
We may need to update [the Coding Style Conventions](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#add-_-prefix-to-the-names-of-private-methods-functions-fields-and-classes) doc as it doesn't mention base class methods.

> I think this method should have a better name, which I'll address later.

👍

optuna/storages/cached_storage.py
ytsmiling: [This line](https://github.com/optuna/optuna/blob/5c07d0afffeb506866860fa2ca21a71a4199c075/optuna/storages/rdb/models.py#L400) checks the study-level distribution compatibility in RDB storage.

ytsmiling: I fixed this PR so that it checks distribution compatibility across workers.

optuna/storages/cached_storage.py
ytsmiling: I fixed this issue.

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,287 @@
...
+
+        with self._lock:
+            if trial_id in self._hot_cache:
+                cached_trial, updates = self._hot_cache[trial_id]
+                cached_trial.state = state
+                updates.state = state
+                if trial_id in self._dirty_trials:
+                    self._dirty_trials.remove(trial_id)
+                ret = self._update_storage(trial_id)
+                if cached_trial.state.is_finished():
```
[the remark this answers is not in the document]
ytsmiling: Applied your suggestion.

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def create_new_trial(self, study_id: int, template_trial: Optional[FrozenTrial] = None) -> int:
+
+        trial_id, frozen_trial = self._storage.create_new_trial_with_trial(
+            study_id, template_trial
+        )
+        with self._lock:
+            self._study_id_to_trial_id[study_id].append(trial_id)
+            self._trial_id_to_study_id[trial_id] = study_id
+            self._hot_cache[trial_id] = frozen_trial, _TrialUpdates()
```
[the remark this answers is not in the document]
ytsmiling: Applied your suggestion.

optuna/storages/rdb/storage.py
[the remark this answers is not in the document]
ytsmiling: Added `_` prefix, which I hope to be removed in the near future.

optuna/storages/rdb/storage.py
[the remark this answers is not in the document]
ytsmiling: Added the prefix.

optuna/storages/__init__.py
```diff
@@ -1,20 +1,22 @@
...
-from optuna.storages.redis import RedisStorage  # NOQA
-from optuna.storages.rdb.storage import RDBStorage  # NOQA
+from optuna.storages.base import BaseStorage
+from optuna.storages.cached_storage import CachedStorage as _Cache
+from optuna.storages.in_memory import InMemoryStorage
+from optuna.storages.rdb.storage import RDBStorage
+from optuna.storages.redis import RedisStorage
 
-from typing import Union  # NOQA
+from typing import Union
```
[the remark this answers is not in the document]
ytsmiling: Resolved this issue.

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,287 @@
...
+from optuna import distributions
+from optuna.storages import base
+from optuna.storages.rdb.storage import RDBStorage
+from optuna.study import StudyDirection
+from optuna.study import StudySummary
+from optuna.trial import FrozenTrial
+from optuna.trial import TrialState
+
+
+class CachedStorage(base.BaseStorage):
```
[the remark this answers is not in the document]
ytsmiling: Resolved this issue.

optuna/storages/cached_storage.py
sile: I see. Thank you for the explanation!

optuna/storages/rdb/storage.py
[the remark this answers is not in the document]
sile: > Do you think it's better to make this method private this time and make it public when it becomes a member of the `BaseStorage` interface?

Yes, I think so.

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,291 @@
...
+                else:
+                    self._param_distributions[study_id][param_name] = distribution
+                if param_name in cached_trial.params:
+                    return False
+                self._dirty_trials.add(trial_id)
+                cached_trial.params[param_name] = distribution.to_external_repr(
+                    param_value_internal
+                )
+                cached_trial.distributions[param_name] = distribution
+                if cached_dist:
```
grafi-tt, line 157: I guess the line

```
self._dirty_trials.add(trial_id)
```

should be moved under `if cached_dist:` .

The logic here is a bit difficult too. Some comments, like "If cached_dist isn't available, we have to access the storage in order to check distribution compatibility." would help readers.

ytsmiling: Thank you for your comment. Indeed, the line should be moved, and the whole logic requires more readability. I'll address them.

ytsmiling:
This PR should not be merged before #1191.

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,291 @@
...
+from optuna import distributions
+from optuna.storages import base
+from optuna.storages.rdb.storage import RDBStorage
+from optuna.study import StudyDirection
+from optuna.study import StudySummary
+from optuna.trial import FrozenTrial
+from optuna.trial import TrialState
+
+
+class _CachedStorage(base.BaseStorage):
```
toshihikoyanase, line 21: [Opinion Wanted] How about making this class public?
Currently, the `RDBStorage` instance created via the `get_storage` function will be wrapped by `_CachedStorage`, but users can create `RDBStorage` instance directly. It sometimes required to pads DB options such as `pool_pre_ping` and `wait_timeout` for MySQL (c.f., #622) . In such cases, users may want to apply cache to the `RDBStorage` instance by themselves.

toshihikoyanase: Ah, `get_storage` wraps `RDBStorage` and `get_storage` is called by `create_study`. So, `RDBStorage` will be wrapped by `_CachedStorage` and we can keep it private.

ytsmiling:
I rearranged commits and now this PR passed the updated tests #1191.
I addressed review comments and would like to ask re-reviews.

Some implementation has been changed from the latest reviews:
- CachedStorage uses `_StudyInfo`, which is similar to the current implementation of InMemoryStorage.

@sile @toshihikoyanase Please review this PR when you have time.

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,328 @@
...
+from optuna import distributions
+from optuna.storages import base
+from optuna.storages.rdb.storage import RDBStorage
+from optuna.study import StudyDirection
+from optuna.study import StudySummary
+from optuna.trial import FrozenTrial
+from optuna.trial import TrialState
+
+
+class _TrialUpdates:
```
toshihikoyanase, line 18: [Question] Could you tell me why this method name is plural? IMO, it can be `_TrialUpdate`.

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,328 @@
...
+        self._lock = threading.RLock()
+
+    def __getstate__(self):
+        # type: () -> Dict[Any, Any]
+        state = self.__dict__.copy()
+        del state["_lock"]
+        return state
+
+    def __setstate__(self, state):
+        # type: (Dict[Any, Any]) -> None
```
sile, line 60: Ditto:
```suggestion
    def __setstate__(self, state: Dict[Any, Any]) -> None:
```

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,328 @@
...
+    """
+
+    def __init__(self, backend: RDBStorage) -> None:
+        self._backend = backend
+        self._studies = {}  # type: Dict[int, _StudyInfo]
+        self._trial_id_to_study_id_and_number = dict()  # type: Dict[int, Tuple[int, int]]
+        self._lock = threading.RLock()
+
+    def __getstate__(self):
+        # type: () -> Dict[Any, Any]
```
sile, line 54: A nitpicking but let me suggest the following change:
```suggestion
    def __getstate__(self) -> Dict[Any, Any]:
```

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,328 @@
...
+            if study_id in self._studies:
+                for key, trial in self._studies[study_id].trials.items():
+                    trials[trial.number] = trial
+            return copy.deepcopy(trials) if deepcopy else trials
+
+    def get_n_trials(self, study_id: int, state: Optional[TrialState] = None) -> int:
+
+        return self._backend.get_n_trials(study_id, state)
+
+    def flush(self) -> None:
```
sile, line 303: It seems that this method is not used in this PR. So how about removing it at the moment?
Or, if we keep this as a public method, I think that it's needed to add `with self._lock:` at the head of the body.

optuna/storages/cached_storage.py
```diff
@@ -0,0 +1,328 @@
...
+        self.values = dict()  # type: Dict[int, float]
+        self.user_attrs = dict()  # type: Dict[str, Any]
+        self.system_attrs = dict()  # type: Dict[str, Any]
+        self.params = dict()  # type: Dict[str, Any]
+        self.distributions = dict()  # type: Dict[str, distributions.BaseDistribution]
+
+
+class _StudyInfo:
+    def __init__(self) -> None:
+        self.trials = dict()  # type: Dict[int, FrozenTrial]
```
sile, line 31: This is not a strong opinion but I like `running_trials` as the field name.
(you can ignore this comment)

ytsmiling: Thank you for the comment. I agree with that, but I'm planning to include finished trials into _StudyInfo in the next PR. So let me keep the name.

optuna/storages/cached_storage.py
ytsmiling: I'll replace the name with `_TrialUpdate`.

optuna/storages/cached_storage.py
sile: I see. Thank you for your response.

optuna/storages/cached_storage.py
ytsmiling: Thank you. I removed the method.

optuna/storages/rdb/storage.py
```diff
@@ -450,9 +448,28 @@ def get_all_study_summaries(self) -> List[StudySummary]:
     def create_new_trial(self, study_id, template_trial=None):
         # type: (int, Optional[FrozenTrial]) -> int
 
+        return self._create_new_trial_with_trial(study_id, template_trial)[0]
+
+    def _create_new_trial_with_trial(
+        self, study_id: int, template_trial: Optional[FrozenTrial] = None
+    ) -> Tuple[int, FrozenTrial]:
+        """Create a new trial and returns trial_id and the trial converted into a FrozenTrial.
```
toshihikoyanase, line 456: It is a bit difficult for me to parse this sentence. If possible, could you rephrase it? The following is just an example.
```suggestion
        """Create a new trial and return trial_id with the trial converted into a :class:`~optuna.trial.FrozenTrial`.
```

optuna/storages/rdb/storage.py
```diff
@@ -450,9 +448,28 @@ def get_all_study_summaries(self) -> List[StudySummary]:
...
+
+        Args:
+            study_id:
+                Study id.
+            template_trial:
+                A FrozenTrial with default trial attributes.
```
toshihikoyanase, line 462: [nit] Please ignore if not relevant. Is the following sentence too verbose?
```suggestion
                A :class:`~optuna.trial.FrozenTrial` with default values for trial attributes.
```

optuna/storages/rdb/storage.py
```diff
@@ -450,9 +448,28 @@ def get_all_study_summaries(self) -> List[StudySummary]:
...
+
+        Returns:
+            A tuple of the trial_id of the created trial and the trial converted
+            into a FrozenTrial instance.
```
toshihikoyanase, line 466: [nit] Please add an internal link to `FrozenTrial`.
```suggestion
            into a :class:`~optuna.trial.FrozenTrial` instance.
```

tests/storages_tests/test_cached_storage.py
```diff
@@ -0,0 +1,152 @@
...
+
+                    storage.set_trial_state(trial_id, state)
+                    assert update_mock.call_count == 1
+                    assert set_mock.call_count == 0
+                    assert get_mock.call_count == 0
+
+                    storage.get_trial(trial_id)
+                    assert update_mock.call_count == 1
+                    assert set_mock.call_count == 0
+                    assert get_mock.call_count == 1
```
toshihikoyanase, line 133: I think that this nest is a bit deep. How about concatenating with-statements by commas? (c.f., https://docs.python.org/3/reference/compound_stmts.html#the-with-statement)
If it's OK, please apply similar changes to other test cases.

```suggestion
        with patch.object(base_storage, "_update_trial", return_value=True) as update_mock, \
            patch.object(base_storage, "set_trial_state", return_value=True) as set_mock, \
            patch.object(base_storage, "get_trial", return_value=True) as get_mock:

            storage.get_trial(trial_id)
            assert update_mock.call_count == 0
            assert set_mock.call_count == 0
            assert get_mock.call_count == 0

            storage.set_trial_state(trial_id, state)
            assert update_mock.call_count == 1
            assert set_mock.call_count == 0
            assert get_mock.call_count == 0

            storage.get_trial(trial_id)
            assert update_mock.call_count == 1
            assert set_mock.call_count == 0
            assert get_mock.call_count == 1
```

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -215,3 +217,103 @@ def setup_trials(storage, study_id):
...
+            {"user_attrs": {"attrA": 2.3, "attrB": "foo"}},
+            {"user_attrs": {"attrA": 2.3, "attrB": "foo"}},
+        ),
+        (
+            {"system_attrs": {"attrC": 2.3, "attrB": "bar"}},
+            {"system_attrs": {"attrC": 2.3, "attrB": "bar"}},
+        ),
+    ]  # type: List[Tuple[Dict[str, Any], Dict[str, Any]]]
+
+    for fields_to_modify, kwargs in test_sets:
```
toshihikoyanase, line 262: IMO, we can make the logic shallow if you use `@pytest.mark.parametrize`. What do you think?

```python
@pytest.mark.parametrize(
    "fields_to_modify, kwargs",
    [
        (
            {"state": TrialState.COMPLETE, "datetime_complete": None},
            {"state": TrialState.COMPLETE},
        ),
        ...
    ],
)
def test_update_trial(fields_to_modify: Dict[str, Any], kwargs: Dict[str, Any]) -> None:
    storage = create_test_storage()
    study_id = storage.create_new_study()

    trial_id = storage.create_new_trial(study_id)
    trial_before_update = storage.get_trial(trial_id)
    ....
```

ytsmiling: It looks nice and I'll modify the code accordingly.

tests/storages_tests/test_cached_storage.py
ytsmiling: Thank you for the suggestion. I'll apply the change to all test cases.

optuna/storages/rdb/storage.py
```diff
@@ -508,7 +525,165 @@ def create_new_trial(self, study_id, template_trial=None):
...
+        return trial.trial_id, frozen
+
+    def _update_trial(
+        self,
+        trial_id: int,
+        state: Optional[TrialState] = None,
+        value: Optional[float] = None,
+        values: Optional[Dict[int, float]] = None,
+        params: Optional[Dict[str, Any]] = None,
+        dists: Optional[Dict[str, distributions.BaseDistribution]] = None,
```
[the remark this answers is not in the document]
ytsmiling: I was just too lazy, and I'll replace it with `distributions_`.
Note: The name `distributions` conflicts, and thus, I'll add the underscore.

optuna/storages/rdb/storage.py
ytsmiling: I think we can replace it with
```Create a new trial and return trial_id with a :class:`~optuna.trial.FrozenTrial`.```

ytsmiling:
@toshihikoyanase Thank you for reviewing this PR.
> 1. FrozenTrial is not frozen anymore. The attribute values are updated to keep the cache up-to-date. Not now, but we may rename it in the future.

Actually, the mutation of `FrozenTrial` was not introduced in this PR, but it has been a long history in the `InMemoryStorage` class.

> 2. This PR directly manipulates SQLAlchemy queries while the previous implementation tried to capslate them in optuna/rdb/models.py. It may be inevitable to speed up storages.

Yes, I think we should remove methods in the `models` package and directly manipulate `sqlalchemy`. It can result in a significant performance gain (e.g. get_all_study_summaries method). From a pure performance perspective, it's even better to stop using orm and directly use SQL, but it requires much more code changes and I think we do not need to switch to SQL for now.

optuna/storages/rdb/storage.py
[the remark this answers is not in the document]
toshihikoyanase: Thanks! I got it.

tests/storages_tests/test_cached_storage.py
toshihikoyanase: Thank you for your update!

tests/storages_tests/rdb_tests/test_storage.py
toshihikoyanase: Thanks! I confirmed the change https://github.com/optuna/optuna/pull/1140/commits/740c5961116aa1a6ae8a1637c8a87b9c62a1bc55.

optuna/storages/rdb/storage.py
toshihikoyanase: I confirmed the change in  https://github.com/optuna/optuna/pull/1140/commits/5ae39047691fcb01e1a6ac8956049ab76beb5a13.
It looks simpler and easier to understand. Thanks!

PR #1145, opened 2020-04-20
Remove unnecessary shape-conversion and a loop from TPE sampler.

ytsmiling:
## Motivation
The current default sampler (TPE) involves unnecessary shape conversions (with copy) and unnecessary loops. Removing them significantly improves performance especially for discreate-uniform and int-uniform distributions.

## Description of the changes
Remove a shape-conversion (with copy) and a loop.

## Benchmark

(CPU: Intel Core i7-6700K, python: 3.5.2)

distribution: DiscreateUniformDistribution

perf:optimize-study: time taken for a single optimization

### This PR
```
{'n_study': 1, 'n_trial': 1000, 'n_param': 10}
perf:optimize-study                               0:00:34.897372
```

### Current master
```
{'n_study': 1, 'n_trial': 1000, 'n_param': 10}
perf:optimize-study                               0:02:59.235532
```

### Code
```python
import argparse
from datetime import datetime
import math

import optuna

profile_result = defaultdict(lambda: datetime.now() - datetime.now())


class Profile:
    def __init__(self, name):
        self.name = name

    def __enter__(self):
        self.start = datetime.now()

        profile_result[self.name] += datetime.now() - self.start


def print_profile():
    for key, value in sorted(profile_result.items(), key=lambda i: i[1],
                             reverse=True):
        print(key.ljust(50) + '{}'.format(value))


    def objective(trial):
        return sum([
            math.sin(trial.suggest_discrete_uniform('param-{}'.format(i), 0, math.pi * 2, math.pi * 0.1))
            for i in range(n_param)
        ])

    return objective


def define_flags(parser):
    parser.add_argument('n_study', type=int)
    parser.add_argument('n_param', type=int)
    return parser


if __name__ == '__main__':
    parser = define_flags(argparse.ArgumentParser())
    args = parser.parse_args()
    print(vars(args))


    for i in range(args.n_study):
        study = optuna.create_study()
        with Profile('perf:optimize-study'):
            study.optimize(build_objective_fun(args.n_param), n_trials=args.n_trial, gc_after_trial=False)

    print_profile()
```

### DIFF
--- a/optuna/samplers/tpe/sampler.py
+++ b/optuna/samplers/tpe/sampler.py
@@ -371,8 +371,6 @@ class TPESampler(base.BaseSampler):
             raise ValueError(
                 "The 'sigmas' should be 2-dimension. " "But sigmas.shape = {}".format(sigmas.shape)
             )
-        _samples = samples
-        samples = _samples.flatten()
 
         p_accept = np.sum(
             weights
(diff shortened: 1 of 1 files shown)


optuna/samplers/tpe/sampler.py
```diff
@@ -371,8 +371,6 @@ def _gmm_log_pdf(
             raise ValueError(
                 "The 'sigmas' should be 2-dimension. " "But sigmas.shape = {}".format(sigmas.shape)
             )
-        _samples = samples
```
hvy, line 374: Asserting `samples.ndim == 1` might aid readability but it's absolutely not a strong opinion (and applies to the above checks raising errors as well).

ytsmiling: Thank you for reviewing.
On that point, I think the other assertions also need changes (`xyz.ndim != 1` &rarr; `xyz.ndim==2`).
I'll create another PR for the refactoring.

Issue #1129, opened 2020-04-16
Implement a decorator to Indicate the date for deprecation.


keisuke-umezawa:
## Motivation

We would like to implement a decorator to Indicate the date for deprecation to show when classes and functions are deprecated to users.

## Description

Implement a decorator to Indicate the date for deprecation, and annotate classes and functions which are deprecated.

scouvreur:
Is anyone working on this - could I take a look at it ?

keisuke-umezawa:
@scouvreur Of course! If you have any ambiguity, please tell about it.

HideakiImamura:
Hello everyone, PR #1382, which introduces deprecation decorator is merged. So let me close this Issue. If there is any discussion, I will open this Issue again.

PR #1146, opened 2020-04-20
Deprecation decorator

scouvreur:
## Motivation

As mentioned in issue #1129, we would like to implement a decorator to Indicate the date for deprecation to show when classes and functions are deprecated to users.

Implement a decorator to Indicate the date for deprecation, and annotate classes and functions which are deprecated.

## Description of the changes

Added the deprecation module to the optuna source code. Reference to original code from scikit-learn [here](https://github.com/scikit-learn/scikit-learn/blob/95d4f0841/sklearn/utils/deprecation.py#L9).

Functions which were deprecated :
- [ ] A
- [ ] B
- [ ] C


hvy:
Thanks for tackling the issue! I see it has been implemented in scikit-learn. I believe we should try to consider the needs in Optuna first however before just borrowing it. For instance, there are unused, and likely unnecessary private functions included. Also from a license perspective, I'm wondering if we shouldn't implement it from scratch, a minimum working decorator for our needs. Please let me know what you think.

scouvreur:
I agree completely @hvy ! I put this code from scikit-learn to get an idea of what a deprecation decorator would look like, and we can simplify it to get a minimum working example.
From what I had checked, as scikit-learn is under a 3-Clause BSD license, we can reuse the code and modify, but if we use exactly the same code we need to include their copyright.

What do you think would be the minimum features we would need for this PR ?

hvy:
I see, thanks for the clarification. I my understanding, we have to include the copyright in case of modification as well, c.f. https://github.com/scikit-learn/scikit-learn/blob/master/COPYING. That's why I think we should start writing from scratch, but please correct me if I'm wrong.

As for the scope, I assume being able to decorate classes and functions could be a first step, raising warnings and possibly appending a note to the documentation. I'm wondering if we can't just "reuse" our experimental-decorator (https://github.com/optuna/optuna/blob/master/optuna/_experimental.py). It also happens that we deprecate arguments but I guess this isn't really straight forward.

hvy:
Is this PR ready for review? Sorry, I don't mean to stress, but as it wasn't clear.

HideakiImamura:
Thanks for your contribution! I created a separate PR based on the experimental decorator's implementation. The approach seems to be the same as the existing decorator and easy to maintain, so I'll go ahead with this one. Please ask any questions or comments.

HideakiImamura:
Hello everyone, PR #1382, which introduces deprecation decorator is merged. So let me close this PR. If there is any discussion, I will open this PR again.

PR #1149, opened 2020-04-21
Bundle allennlp test data in the repository.

himkt:
This PR adds data for testing to the repository.
`test_alllennlp.py` uses datasets on cloud storage (Amazon S3), which causes network traffic.
To avoid this network I/O, I create small datasets and put them on the test directory.


crcrpar:
@toshihikoyanase Could you take a look as you assigned yourself?

PR #1155, opened 2020-04-23
[RFC] Define specs of storage classes, make implementations consistent, and add tests.

ytsmiling:
## Motivation
The current BaseStorage class does not have a concrete specification of its interface. As a result, each backend implementation does not offer the same checks nor handle errors in the same way. This also means that storage classes are not fully tested, and we even cannot add tests without specifications. 

Additionally, we need to specify the consistency model of storage classes for more efficient processing. Otherwise, storages need to offer strong-consistency, which is a current bottleneck of optimization with RDB backend.

## Description of the changes
This PR adds docstrings to the `BaseStorage` class and also adds `sync` and `update_trial` interface to the `BaseStorage` class. Current docstrings have descriptions for both current implementation and proposed specifications. I'd like to hear committers' opinion on the changes via comments or reviews.

After the specification reaches an agreement, I plan to make implementations of all backends consistent within this PR.

The following is the proposed class doc of `BaseStorage`:
```python
"""Base class for storages.

    This class is not supposed to be directly accessed by library users.

    Storage classes abstract a backend database and provide library internal interfaces to
    read/write history of studies and trials.

    Storage class can assume that the returned values are read-only. Additionally, storage
    classes can assume that a single process has at most one storage. However, the storage
    might be shared from multiple threads and storage classes must be thread-safe.

    Storage classes must support monotonic-reads consistency model, that is, if a
    process reads a data `X`, any successive reads on data `X` does not return
    older values.
    They must virtually support read-your-writes, that is, if a process writes to
    data `X`, any successive reads on data `X` from the same process must read
    the written value or one of more recent values.
    If `sync` method is called, any successive reads on `state` attribute
    of `Trial` and any attributes of `Study` must be equal or more recent than the value
    at the time that the `sync` method is called.
    If `sync` method is called, any successive reads on any attributes of `Trial` `T`
    must return the same or more recent value than any preceding writes to `T` via
    `update_trial` method.
    Writes on any attributes of `Trial` must be completed before any
    successive writes on the `state` attribute of the `Trial` with the same trial ID.

    Under assumptions that
    1. there are no process failures,
    2. every trial eventually terminates, and
    3. each trial is updated from only one process,
    storage classes must support eventual-consistency, monotonic-reads,
    monotonic-writes, and read-your-writes consistency models.

    Raises:
        KeyError:
            On reads and writes on non-existing studies and trials,
            storage classes always raise KeyError.
"""
```

An example of added class-doc in the initial state of this PR is as follows.
```python
    @abc.abstractmethod
    def get_trial(self, trial_id: int) -> FrozenTrial:

        """Read a trial using a trial ID.

        Args:
            trial_id:
                Trial ID of a trial to read.

        Returns:
            Trial with a matching trial ID.

        Raises (current implementation):
            in-memory:
                IndexError:
                    If no trial with the matching `trial_id` exists.
            rdb:
                ValueError:
                    If no trial with the matching `trial_id` exists.
            redis:
                AssertionError:
                    If no trial with the matching `trial_id` exists.

        Raises (proposal):
            KeyError:
                If no trial with the matching `trial_id` exists.
        """
```

## Other open discussions

There are many other open problems to discuss, even though they do not necessarily need to be treated in this PR.
The followings are some of them.

### Whether storages should treat `intermediate_values` in the same way as `state` attribute?

Some pruners or samplers might access to intermediate values of other running trials.

### Whether storages should support the bounded-staleness consistency class?

If we ensure that updates are written to persistent storage at least once every N seconds, the above problem on `intermediate_values` might be solved.

### Should the `BaseStorage` class provide more default implementations?

If each backends support at least `update_trial` and `get_trial`, BaseStorage can provide default implementations of many methods such as `get_trial_number`. This will greatly simplify the code and reduce maintenance costs.

### Should storages check whether there are running trials when they delete a study?

### DIFF
--- a/optuna/storages/base.py
+++ b/optuna/storages/base.py
@@ -24,162 +24,848 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
 
     Storage classes abstract a backend database and provide library internal interfaces to
     read/write history of studies and trials.
+
+    Storage class can assume that the returned values are read-only. Additionally, storage
+    classes can assume that a single process has at most one storage. However, the storage
+    might be shared from multiple threads and storage classes must be thread-safe.
+
+    Storage classes must support monotonic-reads consistency model, that is, if a
+    process reads a data `X`, any successive reads on data `X` does not return
+    older values.
+    They must virtually support read-your-writes, that is, if a process writes to
+    data `X`, any successive reads on data `X` from the same process must read
+    the written value or one of more recent values.
+    If `sync` method is called, any successive reads on `state` attribute
+    of `Trial` and any attributes of `Study` must be equal or more recent than the value
+    at the time that the `sync` method is called.
+    If `sync` method is called, any successive reads on any attributes of `Trial` `T`
+    must return the same or more recent value than any preceding writes to `T` via
+    `update_trial` method.
+    Writes on any attributes of `Trial` must be completed before any
+    successive writes on the `state` attribute of the `Trial` with the same trial ID.
+
+    Under assumptions that
+    1. there are no process failures,
+    2. every trial eventually terminates, and
+    3. each trial is updated from only one process,
+    storage classes must support eventual-consistency, monotonic-reads,
+    monotonic-writes, and read-your-writes consistency models.
+
+    Raises:
+        KeyError:
+            On reads and writes on non-existing studies and trials,
+            storage classes always raise KeyError.
     """
 
     # Basic study manipulation
 
     @abc.abstractmethod
-    def create_new_study(self, study_name=None):
-        # type: (Optional[str]) -> int
+    def create_new_study(self, study_name: Optional[str] = None) -> int:
+
+        """Creates a new study with a given name.
+
+        When no name is specified, storage class auto-generates the name.
+        Study ID is unique among all current and deleted studies.
+
+        Args:
+            study_name:
+                Name of a new study to create.
+
+        Returns:
+            Study ID, which is an unique id among studies, of the created study.
+
+        Raises (current implementation):
+            rdb:
+                optuna.exceptions.DuplicatedStudyError:
+                    If a study with the same `study_name` already exists.
+            redis:
+                optuna.exceptions.DuplicatedStudyError:
+                    If a study with the same `study_name` already exists.
+
+        Raises (proposal):
+            optuna.exceptions.DuplicatedStudyError:
+                If a study with the same `study_name` already exists.
+        """
 
         raise NotImplementedError
 
     @abc.abstractmethod
-    def delete_study(self, study_id):
-        # type: (int) -> None
+    def delete_study(self, study_id: int) -> None:
+
+        """Deletes a study specified by the study ID.
+
+        Trial ID is unique among all current and deleted trials.
+
+        Args:
+            study_id:
+                Study ID of a study to delete.
+
+        Raises (current implementation):
+            in-memory:
+                ValueError:
+                    If no study with the matching `study_id` exists.
+            rdb:
+                ValueError:
+                    If no study with the matching `study_id` exists.
+            redis:
+                ValueError:
+                    If no study with the matching `study_id` exists.
+
+        Raises (proposal):
+            KeyError:
+                If no study with the matching `study_id` exists.
+        """
 
         raise NotImplementedError
 
     @abc.abstractmethod
-    def set_study_user_attr(self, study_id, key, value):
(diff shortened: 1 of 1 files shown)


optuna/storages/base.py
```diff
@@ -24,162 +24,848 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+                Trial ID of a trial to add a parameter.
+            param_name:
+                Name of a parameter to add.
+            param_value_internal:
+                Internal representation of a value a parameter to add.
+            distribution:
+                Sampled distribution of a parameter to add.
+
+        Returns:
+            Return False when the parameter is already set to the trial.
```
ytsmiling, line 481: This only holds when each trial is updated from a single process.
We might explicitly add write-from-single-process as a requirement of this class.

optuna/storages/base.py
```diff
@@ -194,28 +880,159 @@ def get_best_trial(self, study_id):
...
+                New parameter dictionary. None when there are no updates.
+            dists:
+                New parameter distributions. None when there are no updates.
+            user_attrs:
+                New user_attr. None when there are no updates.
+            system_attrs:
+                New system_attr. None when there are no updates.
+
+        Returns:
+            True when success.
```
ytsmiling, line 1010: TODO: define what successes and failures are.

optuna/storages/base.py
```diff
@@ -24,162 +24,848 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+                AssertionError:
+                    If no trial with the matching `trial_id` exists.
+                RuntimeError:
+                    If the trial is already finished.
+
+        Raises (proposal):
+            KeyError:
+                If no trial with the matching `trial_id` exists.
+            RuntimeError:
+                If the trial is already finished.
```
c-bata, line 460: FYI, `study._pop_waiting_trial_id()` calls `set_trial_state()` to get `WAITING` trials. To avoid that multiple workers pop the same trial id, Optuna checked its returned value.

* `True`: This means that the `WAITING` state is properly changed to `RUNNING`.
* `False`: This means that the state has been already `RUNNING` (another worker pop this trial).

This is originally implemented by me. But I'm sorry that it's not clear.
We need to consider another interfaces like:

1. Raise an exception instead of return `False`.
2. Add new storage API for `enqueue_trial()`.
3. Keep the current design (not changed).

c-bata:
Great work! This is much needed.

> ### Should the BaseStorage class provide more default implementations?
> If each backends support at least update_trial and get_trial, BaseStorage can provide default implementations of many methods such as get_trial_number. This will greatly simplify the code and reduce maintenance costs.

In my opinion, we can remove methods if we can provide default implementations. In terms of `get_trial_number(trial_id)`, `get_trial(trial_id).number` is enough simple.

optuna/storages/base.py
ytsmiling: Thank you for catching the missing doc on the return value, that's my mistake.
I think we need further discussions on the return values of this and other methods.

If getting waiting trials is the only usage of this method, I think it'll better to add `states` argument to `get_all_trials` method.

ytsmiling:
Even though there are possibilities that some backends can process queries faster when they know which attributes are accessed, I think deprecating get_trial_* methods and use get_trial(trial_id).* is a good idea. The performance will not be a problem when there are appropriate caches.
For set_trial_* methods, I'd like to leave the interface so that the storage can control when to flush the updates to the storage (e.g., #1140 ).

optuna/storages/base.py
```diff
@@ -24,162 +24,848 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+    the written value or one of more recent values.
+    If `sync` method is called, any successive reads on `state` attribute
+    of `Trial` and any attributes of `Study` must be equal or more recent than the value
+    at the time that the `sync` method is called.
+    If `sync` method is called, any successive reads on any attributes of `Trial` `T`
+    must return the same or more recent value than any preceding writes to `T` via
+    `update_trial` method.
+    Writes on any attributes of `Trial` must be completed before any
+    successive writes on the `state` attribute of the `Trial` with the same trial ID.
+
```
ytsmiling, line 46: It might be better to explicitly require that any finished trials should not be updated. Otherwise, storage classes must read all trials on every `sync` calls.

ytsmiling:
I'd like to close this issue and split them into smaller PRs.
Related: #1170

PR #1158, opened 2020-04-23
Support Optuna callback functions at LightGBM Tuner.

c-bata:
## Motivation

To integrate Optuna LightGBM Tuner with MLFlow, I added `optuna_callbacks` argument in LightGBMTuner.

## Description of the changes

* Add `optuna_callbacks` argument.
* Fix typo.
* Remove unused import statement.
* ~~Add type hints for `start_time`.~~ Remove `start_time` attribute.


optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -432,6 +439,9 @@ def __init__(
         if valid_sets is None:
             raise ValueError("`valid_sets` is required.")
 
+        self.optuna_callbacks = optuna_callbacks
+        self.start_time = None  # type: Optional[float]
```
smly, line 443: [IMO] How about removing `self.start_time`? This is un-documented and not used.

c-bata: Agree. I'll remove `start_time` in this PR.

c-bata:
> LightGBM Tuner doesn't use a normal study. This makes the columns of many parameters in the mlflow ui appear to be empty. study.trials_dataframe() also output empty. How about we keep this PR open until this issue is resolved?

OK. I see. Do you have a plan to resolve this issue?
It looks enough just report default `lbgm_params` here.
https://github.com/optuna/optuna/blob/bc6d1a1e94be9627b520f30b334d153facddcb77/optuna/integration/lightgbm_tuner/optimize.py#L220-L244

c-bata:
If I added a patch like following, it looks default parameters are reported.

```diff
(venv) $ git diff --cached
diff --git a/optuna/integration/lightgbm_tuner/optimize.py b/optuna/integration/lightgbm_tuner/optimize.py
index 30cdf86d..fe88e4d1 100644
--- a/optuna/integration/lightgbm_tuner/optimize.py
+++ b/optuna/integration/lightgbm_tuner/optimize.py
@@ -242,6 +242,20 @@ class OptunaObjective(BaseTuner):
             param_value = int(trial.suggest_uniform("min_child_samples", 5, 100 + EPS))
             self.lgbm_params["min_child_samples"] = param_value
 
+        # report default parameters
+        for name in DEFAULT_LIGHTGBM_PARAMETERS:
+            if name in ["lambda_l1", "lambda_l2"]:
+                param = DEFAULT_LIGHTGBM_PARAMETERS[name]
+                # Use suggest_uniform instead of suggest_log_uniform
+                # because default parameter is 0.
+                trial.suggest_uniform(name, param, param)
+            elif name in ["num_leaves", "bagging_freq"]:
+                param = DEFAULT_LIGHTGBM_PARAMETERS[name]
+                trial.suggest_int(name, param, param)
+            elif name in ["feature_fraction", "bagging_fraction", "min_child_samples"]:
+                param = DEFAULT_LIGHTGBM_PARAMETERS[name]
+                trial.suggest_uniform(name, param, param)
+
         with _timer() as t:
             booster = lgb.train(self.lgbm_params, self.train_set, **self.lgbm_kwargs)
```

ScreenShot 2020-04-24 17 31 39

smly:
Thank you for the quick response.

However, your patch changes the original algorithm. It doesn't resolve the issue completely. The parameters other than `target_param_names` in `tune_params()` are not always the default values. Actually, the values of the parameters tuned in the previous step should be used.

For example, when `tune_num_leaves()` is called, the parameter `feature_fraction` is the tuned value in the previous step, not the default. `OptunaObjective` object keeps the results of the tuning in the previous step.

smly:
Oops, one mistake. `LightGBMTuner` object keeps the best params. Not `OptunaObjective` object.

c-bata:
Thank you for your comment.

I see. It looks working fine just select from `self.lgbm_params` instead of `DEFAULT_LIGHTGBM_PARAETTERS`. But I'm not confident so waiting for your fix.

smly:
@c-bata, I will approve this PR after removing `start_time`.

The issue of displaying empty parameters in mlflow UI is not a problem because users using the callback can control it. Also, the missing values correctly represent the behavior of the sampler. I am sorry to confuse you regarding this point.

Besides, the version of Optuna that I'm using for my review was outdated and I pointed problems incorrectly. When I checked out HEAD of this PR, there was no problem.

c-bata:
I see. Thank you for your review. I've just pushed a commit which removes `start_time` attribute.

optuna/integration/lightgbm_tuner/optimize.py
smly: Thanks!

smly:
LGTM. Thanks a lot for improving LightGBM Tuner!

toshihikoyanase:
Thank you for your PR. It's quite useful to visualize/analyze LGBM Tuner's results by mlflow.

> This makes the most of parameter columns in the mlflow ui appear to be empty.

I updated @smly's example to show all parameter values using `lightgbm_tuner:lgbm_params` in use attributes. Does it resolve the issue?

https://gist.github.com/toshihikoyanase/f962969e2df70276ce76aa5a40ef5e2d

image

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -351,6 +357,7 @@ def __init__(
         best_params=None,  # type: Optional[Dict[str, Any]]
         tuning_history=None,  # type: Optional[List[Dict[str, Any]]]
         study=None,  # type: Optional[Study]
+        optuna_callbacks=None,  # type: Optional[List[Callable[[Study, FrozenTrial], None]]]
```
toshihikoyanase, line 360: Could you add a test case to check if a given callback is actually called?

smly:
> I updated @smly's example to show all parameter values using lightgbm_tuner:lgbm_params in use attributes. Does it resolve the issue?

Yes, it's perfect. This is very helpful in understanding how LightGBM tuner works. We can easily confirm that the parameter value tuned from the previous step is used for the next step.
tuner_debug

optuna/integration/lightgbm_tuner/optimize.py
c-bata: Thank you for your review. I added a test at dd790e6.

tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
```diff
@@ -1,4 +1,5 @@
 import contextlib
+from mock import MagicMock
```
toshihikoyanase, line 2: Please use `unittest.mock` instead of `mock` (c.f., https://github.com/optuna/optuna/pull/1121)

c-bata: Thank you. CI is succeeded at 8fb0c69.

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -432,6 +439,8 @@ def __init__(
         if valid_sets is None:
             raise ValueError("`valid_sets` is required.")
 
+        self.optuna_callbacks = optuna_callbacks
```
toshihikoyanase, line 442: Do you think the `self.optuna_callbacks` can be a private attribute (i.e., `self._optuna_callbacks`)?
IMO, we do not expect users to update it during optimization, and it would be safe if it is private.

c-bata: Agree 👍 I pushed 945ac04a.

PR #1142, opened 2020-04-20
Add `IntersectionSearchSpace` to speed up the search space calculation.

sile:
This PR introduces `optuna.samplers.IntersectionSearchSpace` class that is faster than the existing `optuna.samplers.intersection_search_space` function.

This takes over https://github.com/optuna/optuna/pull/885. 

## Benchmark script

```python
import argparse
import logging
import optuna

parser = argparse.ArgumentParser()
parser.add_argument("--params", type=int, default=100)
parser.add_argument("--trials", type=int, default=1000)
args = parser.parse_args()


def objective(trial: optuna.Trial):
    val = 0
    for i in range(args.params):
        xi = trial.suggest_uniform(str(i), -4, 4)
        val += (xi - 2) ** 2
    return val


def main():
    logging.disable(level=logging.INFO)
    sampler = optuna.samplers.CmaEsSampler(seed=0)
    study = optuna.create_study(sampler=sampler)
    study.optimize(objective, n_trials=args.trials, gc_after_trial=False)
    print(study.best_params)


if __name__ == "__main__":
    main()
```

The benchmark script was run on my laptop. It seems to be 10 times faster than before.

**before(ae0c8b483132e70c2ee30ad10f9156cfd4366bba)**

```
$ time -p python3 examples/bench.py --params 10 --trials 1000
{'0': 1.9531119708170848, '1': 2.086781403839324, '2': 2.0622082517919336, '3': 1.9866026063257591, '4': 2.058500085804301, '5': 2.020942555633829, '6': 1.9894401312961876, '7': 2.043186636871422, '8': 1.954916778325024, '9': 2.093560500426215}

real 96.11
user 161.95
sys 87.45
```

**after**

```
$ time -p python3 examples/bench.py --params 10 --trials 1000
{'0': 1.9531119708170848, '1': 2.086781403839324, '2': 2.0622082517919336, '3': 1.9866026063257591, '4': 2.058500085804301, '5': 2.020942555633829, '6': 1.9894401312961876, '7': 2.043186636871422, '8': 1.954916778325024, '9': 2.093560500426215}

real 7.30
user 11.10
sys 7.65
```

PR #1160, opened 2020-04-23
Add predefined_search_space option to `create_study()`.

c-bata:
## Motivation

Add predefined_search_space option. We can sample parameters by `sample_relative` from the first trial.

See https://github.com/c-bata/goptuna/pull/99.

## Description of the changes

* Add predefined_search_space option to `create_study()`.


c-bata:
**Quick recap of this pull request.**

`predefined_search_space` option is added to Goptuna for Kubeflow/Katib integration.
Then I created this patch because this also seems to be useful in Optuna.

The main motivation is to trigger relative samplers from the first trial though there are some by-products that are performance improvements and Define-and-Run interface support.

Current concerns are as follows:

* It's hard to specify `Distribution` directly. The interface which accepts Hyperopt's search space might be better for users (`predefined_search_space` might be enough as a low-level interface).
* Due to #1142, the benefits has been lower in terms of the performance improvements.

It doesn't need to merge this pull request soon. We will discuss this pull request again if we start to implement a Hyperopt-compatible Define-and-Run interface.

c-bata:
ScreenShot 2020-07-27 19 37 57

CI seems to be failed at `tests/integration_tests/test_tensorboard.py` which is not relevant to this changes.

Issue #1136, opened 2020-04-19
Enable to skip pruning by specified interval for every pruning class.


keisuke-umezawa:
## Motivation

We want to specify interval to skip pruning, so that we can fasten iterations.
In most cases,  we do not need to run pruning after each iteration, but currently we can not avoid it.

## Description

Enable to skip pruning by specified interval for the certain PruningCallback classes.

- PruningCallback
    - [x] [LightGBMPruningCallback](https://github.com/optuna/optuna/blob/master/optuna/integration/lightgbm.py)
    - [ ] [XGBoostPruningCallback](https://github.com/optuna/optuna/blob/master/optuna/integration/xgboost.py)

VladSkripniuk:
`TensorFlowPruningHook` skips pruning based on `run_every_steps` argument. `ChainerPruningExtension` skips pruning based on `pruner_trigger` argument.
I think we need a discussion to decide whether we need to reformat these two classes to add `interval` argument.

toshihikoyanase:
I think we have an alternative approach to skip pruning. How about adding a method like `BasePruner.is_target_step(step: int) -> bool` so that the callbacks can query the timing of pruning to the pruners?

This is because the pruners have logic to skip pruning. If we add the `interval` argument to the callbacks, users may need to take care of the relationship between the logic and the `interval` argument.

For example, `MedianPruner` has the `n_interval_steps` argument to specify the interval in number of steps between the pruning checks. And `SuccessiveHalvingPruner` determines the pruning steps based on the arguments.

It may be confusing if users use both the `interval` argument and `SuccessiveHalvingPruner` because the pruning interval of `SuccessiveHalvingPruner` increases an exponential manner. So, IMO, the `BasePruner.is_target_step` method will improve usability.

I'm not fully sure if it covers actual use cases. So, if you have any comments and questions, please let me know.

c.f., https://github.com/optuna/optuna/pull/1161#issuecomment-625328649

keisuke-umezawa:
> This is because the pruners have logic to skip pruning. If we add the interval argument to the callbacks, users may need to take care of the relationship between the logic and the interval argument.

That is true. The `interval` should be managed by pruning class.

keisuke-umezawa:
@VladSkripniuk cc: @toshihikoyanase @hvy 

Sorry for being late for updates.

I discussed it with the other developers, and we concluded it as follows:

1. The intervals for reporting and pruning should be configured separately.
[...]

PR #1154, opened 2020-04-22
Enable to skip pruning by specified interval for PyTorchLightningPruningCallback.

VladSkripniuk:
## Motivation

fix #1136 
Work in progress:
- [ ] ChainerPruningExtension
- [ ] FastAIPruningCallback
- [ ] FastAIPruningCallback
- [ ] KerasPruningCallback
- [ ] LightGBMPruningCallback
- [ ] MXNetPruningCallback
- [ ] PyTorchIgnitePruningHandler
- [x] PyTorchLightningPruningCallback
- [ ] TensorFlowPruningHook
- [ ] TFKerasPruningCallback
- [ ] XGBoostPruningCallback

PR #1161, opened 2020-04-23
Add pruning interval for KerasPruningCallback

VladSkripniuk:
Address #1136.

### DIFF
--- a/optuna/integration/keras.py
+++ b/optuna/integration/keras.py
@@ -31,10 +31,14 @@ class KerasPruningCallback(Callback):
             An evaluation metric for pruning, e.g., ``val_loss`` and
             ``val_acc``. Please refer to `keras.Callback reference
             <https://keras.io/callbacks/#callback>`_ for further details.
-    """
+        interval:
+            Check if trial should be pruned every n-th epoch. By default `interval=1` and
+            pruning is performed after every epoch. Increase `interval` to run several
+            epochs faster before applying pruning.
+     """
 
-    def __init__(self, trial, monitor):
-        # type: (optuna.trial.Trial, str) -> None
+    def __init__(self, trial, monitor, interval=1):
+        # type: (optuna.trial.Trial, str, int) -> None
 
         super(KerasPruningCallback, self).__init__()
 
@@ -42,9 +46,13 @@ class KerasPruningCallback(Callback):
 
         self._trial = trial
         self._monitor = monitor
+        self._interval = interval
 
     def on_epoch_end(self, epoch, logs=None):
         # type: (int, Dict[str, float]) -> None
+
+        if (epoch + 1) % self._interval != 0:
+            return
 
         logs = logs or {}
         current_score = logs.get(self._monitor)
--- a/optuna/integration/keras.py
+++ b/optuna/integration/keras.py
@@ -32,8 +32,8 @@ class KerasPruningCallback(Callback):
             ``val_acc``. Please refer to `keras.Callback reference
             <https://keras.io/callbacks/#callback>`_ for further details.
         interval:
-            Check if trial should be pruned every n-th epoch. By default `interval=1` and
-            pruning is performed after every epoch. Increase `interval` to run several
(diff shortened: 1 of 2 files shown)


optuna/integration/keras.py
```diff
@@ -31,21 +31,29 @@ class KerasPruningCallback(Callback):
             An evaluation metric for pruning, e.g., ``val_loss`` and
             ``val_acc``. Please refer to `keras.Callback reference
             <https://keras.io/callbacks/#callback>`_ for further details.
-    """
+        interval:
+            Check if trial should be pruned every n-th epoch. By default `interval=1` and
+            pruning is performed after every epoch. Increase `interval` to run several
```
hvy, line 36:
```suggestion
            Check if trial should be pruned every n-th epoch. By default ``interval=1`` and
            pruning is performed after every epoch. Increase ``interval`` to run several
```

tests/integration_tests/test_keras.py
```diff
@@ -9,8 +9,9 @@
 from optuna.testing.integration import DeterministicPruner
 
 
-def test_keras_pruning_callback():
-    # type: () -> None
+@pytest.mark.parametrize("interval, epochs", [(1, 1), (2, 1), (2, 2)])
+def test_keras_pruning_callback(interval, epochs):
```
keisuke-umezawa, line 13: Just a nit, and we can fix it in another follow up PR.

For me, it is easier to understand what they are testing if we divide `test_keras_pruning_callback` and `test_keras_pruning_callback_with_interval`.

c.f. 
https://github.com/optuna/optuna/pull/1154/files#diff-58bac36a299f885bfb5a18600c4b6cb1R110

tests/integration_tests/test_keras.py
```diff
@@ -22,16 +23,19 @@ def objective(trial):
...
 
         return 1.0
 
     study = optuna.create_study(pruner=DeterministicPruner(True))
     study.optimize(objective, n_trials=1)
-    assert study.trials[0].state == optuna.trial.TrialState.PRUNED
+    if interval <= epochs:
+        assert study.trials[0].state == optuna.trial.TrialState.PRUNED
+    else:
+        assert study.trials[0].state == optuna.trial.TrialState.COMPLETE
```
keisuke-umezawa, line 38: Just a nit, and we can fix it in another follow up PR.

How about adding this line below?
```
assert study.trials[0].value == 1.0
```

VladSkripniuk: Right, I'll open a follow up PR once this PR is merged.

hvy:
@toshihikoyanase sorry for pinging but did you have any comments on including the 0-th step or any other details?

toshihikoyanase:
@hvy Thank you for reminding me.

I notice that users can control pruning interval in two ways if we add `interval` to pruning callbacks. For example, [`MedianPruner`](https://optuna.readthedocs.io/en/stable/reference/pruners.html#optuna.pruners.MedianPruner) has `n_warmup_steps` and `interval_steps` to control the steps to check the intermediate values. Users can configure the pruning interval by both `interval` of `KerasPruningCallback` and `interval_steps` of `MedianPruner`.

I'm thinking if we can unify these to make the interface simple.

This is just an example, but how about adding a method to check the pruning interval to `BasePruner` (e.g., the `is_target_step` method of the following snippet)?

```python

    def on_epoch_end(self, epoch, logs=None):
        # type: (int, Dict[str, float]) -> None

        if not self._trial.study.pruner.is_target_step(epoch):
            return

```

IMO, users usually use the `interval` of `KerasPruningCallback` to stop reporting intermediate values which are not referred by pruners. If so, it seems to be natural to calculate pruning interval in pruners.

What do you think?

toshihikoyanase:
@hvy
I think it takes time to design/implement `Pruner.is_target_step`, so I'd like to create a new PR if I work on it. Please merge this PR.
Thank you for your kind consideration

hvy:
Thanks @VladSkripniuk for the long running PR, and @toshihikoyanase for you careful insight. For reasons stated above, I've merged this PR.

PR #1162, opened 2020-04-23
Add pruning interval for XGBoostPruningCallback

VladSkripniuk:
Address #1136.

### DIFF
--- a/optuna/integration/xgboost.py
+++ b/optuna/integration/xgboost.py
@@ -44,9 +44,9 @@ class XGBoostPruningCallback(object):
             `XGBoost reference <https://xgboost.readthedocs.io/en/latest/parameter.html>`_
             for further details.
         interval:
-            Check if trial should be pruned every n-th epoch. By default `interval=1` and
-            pruning is performed after every epoch. Increase `interval` to run several
-            epochs faster before applying pruning.
+            Check if trial should be pruned every n-th iteration. By default `interval=1` and
+            pruning is performed after every iteration. Increase `interval` to run several
+            iterations faster before applying pruning.
     """
 
     def __init__(self, trial, observation_key, interval=1):
@@ -63,7 +63,6 @@ class XGBoostPruningCallback(object):
 
         context = _get_callback_context(env)
 
-        # Iterations start with 1.
         if (env.iteration + 1) % self._interval != 0:
             return
 
--- a/optuna/integration/xgboost.py
+++ b/optuna/integration/xgboost.py
@@ -63,7 +63,7 @@ class XGBoostPruningCallback(object):
 
         context = _get_callback_context(env)
 
-        if (env.iteration + 1) % self._interval != 0:
+        if env.iteration % self._interval != 0:
             return
 
         evaluation_result_list = env.evaluation_result_list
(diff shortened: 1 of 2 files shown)


optuna/integration/xgboost.py
```diff
@@ -43,20 +43,30 @@ class XGBoostPruningCallback(object):
...
         self._trial = trial
         self._observation_key = observation_key
+        self._interval = interval
 
     def __call__(self, env):
         # type: (xgb.core.CallbackEnv) -> None
 
         context = _get_callback_context(env)
+
+        # Iterations start with 1.
```
hvy, line 66: Are you sure this always holds? It seems to start from 0 occasionally, running existing tests. I'm not sure what controls it, though.

optuna/integration/xgboost.py
```diff
@@ -43,20 +43,30 @@ class XGBoostPruningCallback(object):
...
+        if (env.iteration + 1) % self._interval != 0:
```
hvy, line 67: Are XGBoost users more familiar with logic that include iteration 0? E.g. https://github.com/dmlc/xgboost/blob/6a169cd41a5ea23b4371bfe2c1b39c6687bfebcd/python-package/xgboost/callback.py#L52-L53.

optuna/integration/xgboost.py
```diff
@@ -43,20 +43,30 @@ class XGBoostPruningCallback(object):
             ``validation-merror``. Please refer to ``eval_metric`` in
             `XGBoost reference <https://xgboost.readthedocs.io/en/latest/parameter.html>`_
             for further details.
+        interval:
+            Check if trial should be pruned every n-th epoch. By default `interval=1` and
+            pruning is performed after every epoch. Increase `interval` to run several
+            epochs faster before applying pruning.
```
hvy, line 49: I guess it should be iteration and not epoch.

optuna/integration/xgboost.py
VladSkripniuk: Actually iterations start from 0, in one of existing tests iterations start from 1 because `env.iteration` is artificially set to 1 ([link](https://github.com/optuna/optuna/blob/3fe0ab712096eaf434e521b1b0b4f0fab4dc8740/tests/integration_tests/test_xgboost.py#L20)). Comment `# Iterations start with 1.` was wrong.

optuna/integration/xgboost.py
VladSkripniuk: Right, I fixed this.

optuna/integration/xgboost.py
VladSkripniuk: Indeed, it is more reasonable to include iteration 0 to prune unpromising trial earlier. I changed it to `env.iteration % self._interval`.

keisuke-umezawa:
@VladSkripniuk Sorry for being late. As I commented in https://github.com/optuna/optuna/issues/1136#issuecomment-817081729 , we concluded how we will resolve the issue. If you are still interested in it, we will be very happy that you will work on it. Thanks!

Crissman:
@VladSkripniuk This PR is stale and has conflicts with master, so closing for now. If there's further interest, we may want to take a look at #1161 (comment)

PR #1166, opened 2020-04-25
Fix split_observations for conditional parameters

y0z:
This PR fixes split_observations for conditional parameters to calculate EI values.

## Motivation
The current implementation of the split_observations in the TPESampler is as follows:
1. The sampler discards the trials that do not contain the parameter named ``param_name`` when it gets observation pairs. This causes different len(config_vals) for conditional parameters.
  https://github.com/optuna/optuna/blob/f614d69875cbbd24f989fc8c75f9b18f3bb7c541/optuna/samplers/tpe/sampler.py#L562
2. Then, the sampler calculates the y* value based on the gamma = min(int(np.ceil(0.1 * len(config_vals))), 25)
  https://github.com/optuna/optuna/blob/f614d69875cbbd24f989fc8c75f9b18f3bb7c541/optuna/samplers/tpe/sampler.py#L183
3. Finally, the sampler splits config_vals into the ones with loss_vals < y* (i.e., the observations for l(x)) and the remaining ones (i.e., the observations for g(x)).

Based on my understanding, y* which determines the integration interval for the EI calculation should be the same value for all parameters (ref. [the TPE paper](https://papers.nips.cc/paper/4443-algorithms-for-hyper-parameter-optimization.pdf)).
However, the current strategy may set different y* values for conditional parameters.
[Here](https://gist.github.com/y0z/d85e5b1c1cb0d608bc93a02f89e1f8d4) is an example of this situation (note: this example is maximization).
In this example, y* for params_classifier is 0.9466666666666667, whereas y* for params_svc_c 0.32.
This seems to cause a curious EI calculation, i.e., a different integration interval is set for each parameter.

## Description of the changes
This PR changes the split_observation procedure as follows:
1. The sampler does not discard the trials that do not contain the parameter named ``param_name`` when it gets observation.
2. Then, the sampler calculates the y* value based on the gamma = min(int(np.ceil(0.1 * len(config_vals))), 25). This y* is consistent among all hyperparameters.
3. After that, the sampler splits config_vals into the observations for l(x) and the observations for g(x).
4. Finally, the sampler discards the config_vals with the observations that do not contain the parameter named ``param_name`` from the observations for l(x) and g(x).

I have also confirmed that the split_observation procedure in the original TPE implementation in HyperOpt is like this.
https://github.com/hyperopt/hyperopt/blob/master/hyperopt/tpe.py#L625
In this implementation, ``o_vals`` is the set of config_vals that contain the parameter named ``param_name`` only, whereas ``l_vals`` is the set of all of the observed losses so far.

My apologies if I am misunderstanding the algorithm and please close this PR.

### DIFF
--- a/optuna/samplers/tpe/sampler.py
+++ b/optuna/samplers/tpe/sampler.py
@@ -137,14 +137,6 @@ class TPESampler(base.BaseSampler):
             )
         below_param_values, above_param_values = self._split_observation_pairs(values, scores)
 
-        # Parzen estimator construction requires at least one observation or a priror.
-        n_below = len(below_param_values)
-        n_above = len(above_param_values)
-        if not self._parzen_estimator_parameters.consider_prior and (n_below == 0 or n_above == 0):
-            return self._random_sampler.sample_independent(
-                study, trial, param_name, param_distribution
-            )
-
         if isinstance(param_distribution, distributions.UniformDistribution):
             return self._sample_uniform(param_distribution, below_param_values, above_param_values)
         elif isinstance(param_distribution, distributions.LogUniformDistribution):
--- a/optuna/samplers/tpe/sampler.py
+++ b/optuna/samplers/tpe/sampler.py
@@ -190,9 +190,9 @@ class TPESampler(base.BaseSampler):
         n_below = self._gamma(len(config_vals))
         loss_ascending = np.argsort(loss_vals)
         below = config_vals[np.sort(loss_ascending[:n_below])]
-        below = below[below != None].astype(float)
+        below = np.asarray([v for v in below if v is not None], dtype=float)
         above = config_vals[np.sort(loss_ascending[n_below:])]
-        above = above[above != None].astype(float)
+        above = np.asarray([v for v in above if v is not None], dtype=float)
         return below, above
 
     def _sample_uniform(self, distribution, below, above):
--- a/optuna/samplers/tpe/sampler.py
+++ b/optuna/samplers/tpe/sampler.py
@@ -135,8 +135,15 @@ class TPESampler(base.BaseSampler):
             return self._random_sampler.sample_independent(
                 study, trial, param_name, param_distribution
             )
-
         below_param_values, above_param_values = self._split_observation_pairs(values, scores)
+
+        # Parzen estimator construction requires at least one observation or a priror.
+        n_below = len(below_param_values)
+        n_above = len(above_param_values)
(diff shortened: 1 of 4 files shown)


optuna/samplers/tpe/sampler.py
```diff
@@ -135,9 +135,16 @@ def sample_independent(self, study, trial, param_name, param_distribution):
...
         below_param_values, above_param_values = self._split_observation_pairs(values, scores)
 
+        # Parzen estimator construction requires at least one observation or a priror.
+        n_below = len(below_param_values)
+        n_above = len(above_param_values)
+        if not self._parzen_estimator_parameters.consider_prior and (n_below == 0 or n_above == 0):
+            return self._random_sampler.sample_independent(
+                study, trial, param_name, param_distribution
+            )
+
```
ytsmiling, line 147: Is this fall-back consistent with the hyperopt implementation? cc: @HideakiImamura

ytsmiling:
When there are multiple workers, a similar problem will remain even after this PR is merged. Another PR should address this issue. Related issue: #1170.

optuna/samplers/tpe/sampler.py
y0z: I have looked through the HyperOpt code.
It seems that HyperOpt does not have an option to disable to use a prior.
Therefore, HyperOpt has no fall-back for a situation like this.

The current fall-back is the same for the case: n < self._n_startup_trials.
https://github.com/optuna/optuna/blob/master/optuna/samplers/tpe/sampler.py#L139

Another idea is simply ignoring self._parzen_estimator_parameters.consider_prior flag and enabling a prior in this situation.

HideakiImamura: I agree with @y0z. The hyperopt does not have any option to disable a prior. 
I think ignoring `consider_prior` flag is good idea in this situation, since it is consistent with normal situation.

y0z:
Thank you for your feedback.
I have changed the fall-back procedure.
Now, when the number of observations is zero, the Parzen estimator ignores the ``consider_prior`` flag and utilizes a prior.

HideakiImamura:
@y0z How about conducting some benchmark comparisons between existing TPE and proposed version?

HideakiImamura:
@y0z I think we need some benchmark comparison between existing TPE and proposed one to quantify the improvement, but it may take quite a lot of time. I think it's good to leave a comment in the code without doing any benchmark experiments like this.

```
Note: When the number of observations is zero, the Parzen estimator ignores the consider_prior flag and utilizes a prior. Validation of this approach is future work.
```

Issue #1107, opened 2020-04-08
Rename test datasets in examples.


toshihikoyanase:
Rename test datasets in examples.

As mentioned in https://github.com/optuna/optuna/issues/1091#issuecomment-610953375, some of the Optuna examples use 'test' datasets to tune hyperparameters, but they are 'validation' datasets actually. This naming convention may be confusing to beginners, and this issue proposes to rename them.

## List of target examples

The following examples use terms like `test dataset`  or variables like `x_test`. Please rename them, for example, `validation dataset` or `x_valid`, respectively.

- [ ] [examples/mlflow/keras_mlflow.py](https://github.com/optuna/optuna/blob/master/examples/mlflow/keras_mlflow.py)
- [ ] [examples/pruning/chainer_integration.py](https://github.com/optuna/optuna/blob/master/examples/pruning/chainer_integration.py)
- [ ] [examples/pruning/chainermn_integration.py](https://github.com/optuna/optuna/blob/master/examples/pruning/chainermn_integration.py)
- [ ] [examples/pruning/keras_integration.py](https://github.com/optuna/optuna/blob/master/examples/pruning/keras_integration.py)
- [ ] [examples/pruning/lightgbm_integration.py](https://github.com/optuna/optuna/blob/master/examples/pruning/lightgbm_integration.py)
- [ ] [examples/pruning/mxnet_integration.py](https://github.com/optuna/optuna/blob/master/examples/pruning/mxnet_integration.py)
- [ ] [examples/pruning/simple.py](https://github.com/optuna/optuna/blob/master/examples/pruning/simple.py)
- [ ] [examples/pruning/tensorflow_estimator_integration.py](https://github.com/optuna/optuna/blob/master/examples/pruning/tensorflow_estimator_integration.py)
- [ ] [examples/pruning/xgboost_integration.py](https://github.com/optuna/optuna/blob/master/examples/pruning/xgboost_integration.py)
- [ ] [examples/visualization/plot_study.py](https://github.com/optuna/optuna/blob/master/examples/visualization/plot_study.py)
- [ ] [examples/catboost_simple.py](https://github.com/optuna/optuna/blob/master/examples/catboost_simple.py)
- [ ] [examples/chainer_simple.py](https://github.com/optuna/optuna/blob/master/examples/chainer_simple.py)
- [ ] [examples/chainermn_simple.py](https://github.com/optuna/optuna/blob/master/examples/chainermn_simple.py)
[...]

PR #1167, opened 2020-04-26
Rename `test_` to `valid_` in docs and docstring.

himkt:
Related to #1107 .

I noticed `test` prefix (suffix) appear in documentation.
In 67009ad, I only rename `test` appeared in docstring but they also appear in the code.
(e.g. https://github.com/himkt/optuna/blob/rename-test-in-docs/optuna/integration/sklearn.py#L339)
I left them since I think it is better to focus on documentation fix in this PR.
However, I'm not sure whether it should also rename in code.
(I also wondering mixing `test` and `validation` could be confusing for users)

I want developers to advise me.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -152,7 +152,7 @@ For example, you can save SVM models trained in the objective function as follow
         # Save a trained model to a file.
         with open('{}.pickle'.format(trial.number), 'wb') as fout:
             pickle.dump(clf, fout)
-        return 1.0 - accuracy_score(y_test, clf.predict(X_test))
+        return 1.0 - accuracy_score(y_valid, clf.predict(X_valid))
 
 
     study = optuna.create_study()
@@ -161,7 +161,7 @@ For example, you can save SVM models trained in the objective function as follow
     # Load the best model.
     with open('{}.pickle'.format(study.best_trial.number), 'rb') as fin:
         best_clf = pickle.load(fin)
-    print(accuracy_score(y_test, best_clf.predict(X_test)))
+    print(accuracy_score(y_valid, best_clf.predict(X_valid)))
 
 
 How can I obtain reproducible optimization results?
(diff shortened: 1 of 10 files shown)


toshihikoyanase:
> However, I'm not sure whether it should also rename in code.
> (I also wondering mixing test and validation could be confusing for users)

I agree. If we use the same term in both documentation and code, it will be helpful for users/developers. The scikit-learn library uses `test` in some methods like `train_test_split`, but we can still use the `valid` in Optuna code because we hardly use `test data` during hyperparameter optimization.

himkt:
@toshihikoyanase Apologies I didn't respond.

> Do you include the code fix in this PR, or create a new PR

I'd like to create follow-up PR since code-fix isn't limited to `sklearn.py`.

himkt:
Thank you and apologize I missed it, fixed it.

himkt:
I found some docstrings having `test` prefix/suffix during extracting a list of files for follow-up PR.
I add `[WIP]` on the title of this PR since I'm working on extra fixes.

himkt:
f4b7034 fixes additional docstrings (`exceptions`, `pruners` and `trial`).

- `test_xx`

```bash
git grep -e "test_" --and --not -e "def test_" --and --not -e "create_test_storage" \
              --and --not -e "train_test_split" --and --not -e "test_create_new_study" \
              --and --not -e "test_get_study_id" --and --not -e "latest" \
              optuna examples tests
```

- `xx_test`

```bash
git grep -e "_test" --and --not -e "def test_" --and --not -e "create_test_storage" \
              --and --not -e "train_test_split" --and --not -e "test_create_new_study" \
              --and --not -e "test_get_study_id" --and --not -e "latest" \
              optuna examples tests
```

himkt:
Apologize adding the commit after being reviewed, I think I complete to fix all docstrings.

hvy:
@himkt, sorry but could you fix the conflict?

himkt:
@hvy PTAL.

hvy:
Thanks, let's merge this PR once CI agrees with us.

hvy:
Thank you!

Issue #1170, opened 2020-04-26
Major refactoring of storage classes.


ytsmiling:
## Motivation

Current storage class implementations have several issues listed in the following.
- Inconsistent error handlings between backends.
- Significant bottlenecks in the storage class on training with / querying to RDB.
- O(trial^2) running time even with random-sampler and CMA-ES.
- Undefined consistency models for multi-worker optimizations.

Some of these limitations come from the optuna's origin, that is, optuna was initially
developed for deep-learning applications with ~1000 trials.
Given increasing applications optuna, it will be a good time to reconsider the design
and implementations of storage classes.

This issue is meant to be used for open discussions and PR tracking.

## Roadmaps

Some of the roadmaps come from committers' discussions, but they are open for discussions. Comments are appreciated.

### Documentation and implementation consistency

- [ ] Document consistency models for storage classes.
- [ ] Document each method, especially concerning error-handling.
- [ ] Add tests and fix implementation on error-handling.
- [ ] Log when new studies are created in all backends.
- [ ] Define the default study name.

### Performance improvements

- [ ] Add cache for unfinished trials.
- [ ] Support bulk updates of trial attributes.
- [ ] Support bulk read of DB contents, and make storages sync only once per trial.
- [ ] Support incremental load of DB contents.
- [ ] Support indexing of finished trials.

### Refactoring

- [ ] Remove unnecessary return values.
- [ ] Add more default implementations in `BaseStorage` class.
- [ ] Removed unnecessary backend-specific method implementations.

### Miscellaneous issues

- [ ] Support heart-beat of trials.
- [ ] Support multi-studies in in-memory storage.
- [ ] Fix race-condition in waiting queue.
- [ ] Nan-handling in best-trial determination.
- [ ] Raise errors when a single trial is modified from multiple workers at the same time.
- [ ] Remove multi-threading support and add multi-processing support.

ytsmiling:
I'd like to summarize recent changes.

## Major Changes
One of the major changes is that trials are only loaded only when `storage.read_from_remote_storage` in RDB backend. This solved a bug that trial looks different set of trials under multi-worker training (#1298). Additionally, reduced communication cost improved overall performance. Another large change is that we add multi-worker tests, which has long been missing (#1497).

## Bulk update
[...]

PR #1174, opened 2020-04-27
Add storage specification to `BaseStorage` class doc.

ytsmiling:
## Motivation

As noted in #1170, defining consistency models let storage implementations to remove their bottlenecks.

## Description of the changes.

This PR adds a class doc to `BaseStorage` class which defines consistency models which storage classes should guarantee. Additionally, it also notes on thread-safety and data-persistence.

This PR introduces `load(study_id)` method. It is not supported by the current implementation and will be added by successive PRs. The method is not only required to reduce the sync cost with a remote database, but also necessary to alleviate the problem described in #1166 in multi-worker settings.

### DIFF
--- a/optuna/storages/base.py
+++ b/optuna/storages/base.py
@@ -31,6 +31,10 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
     When users modify return values of storage classes, it might break the internal states
     of storage classes, which will result in undefined behaviors.
 
+    Storage class can assume that each RUNNING trial is modified from only one process.
+    When users modify a RUNNING trial from multiple processes, it might lead to
+    an inconsistent internal state, which will result in undefined behaviors.
+
     Storage classes must support monotonic-reads consistency model, that is, if a
     process reads a data `X`, any successive reads on data `X` does not return
     older values.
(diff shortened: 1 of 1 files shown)


ytsmiling:
Thank you for checking the doc.
- I forgot to add sections and will address the layout issue.
- I agree that we should add a TODO comment. (Note: Current implementations do not require the load method, and we only need to add an empty method.)
- I intended to fix this spec (at least within #1170) and expand it only when we need more supports from storage for future development (e.g. for new dashboard features). This spec is intended to provide a set of conditions that ensures that storage implementations won't cause the system down even in multi-worker settings. If this spec is not sufficient to ensure/implement safe storages, it is ideal to address the ambiguity or corner cases of this doc within this PR.

optuna/storages/base.py
```diff
@@ -24,6 +24,69 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+    Storage classes must support monotonic-reads consistency model, that is, if a
+    process reads a data `X`, any successive reads on data `X` does not return
+    older values.
+    They must support read-your-writes, that is, if a process writes to data `X`,
+    any successive reads on data `X` from the same process must read the written
+    value or one of more recent values.
+
+    **Stronger consistency requirements for special data**
+
+    TODO(ytsmiling) Add load method to storage class implementations.
```
sile, line 61: [NOTE] I'd like to discuss the name when this method is implemented (`load` sounds too general to me).

ytsmiling: I agree with this comment. I added a TODO comment in #1170.

PR #1175, opened 2020-04-28
Add docstring to `BaseStorage` method interfaces.

ytsmiling:
## Motivation
Storage class implementations for each backend handles error in a different manner. Additionally, undocumented return values force users to read implementations of each method. This PR addresses these issues as a step in #1170. Any inconsistent behaviors with this docstring will be fixed in subsequent PRs.

## Description of the changes
Add docstrings to methods in `BaseStorage` class.


ytsmiling:
Current doc and implementation conflicts with #1174 and allows to change the attributes of WAITING trials, which should be fixed.

optuna/storages/base.py
```diff
@@ -29,157 +28,539 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+        """
 
         raise NotImplementedError
 
     @abc.abstractmethod
-    def set_trial_state(self, trial_id, state):
-        # type: (int, TrialState) -> bool
+    def set_trial_state(self, trial_id: int, state: TrialState) -> bool:
+
+        """Update a state of a specified trial.
```
ytsmiling, line 284: I don't think we should allow the "RUNNING → WAITING" state change. (Both this doc and current implementation allows it now.)

ytsmiling: This PR won't address this issue.

optuna/storages/base.py
```diff
@@ -194,31 +529,79 @@ def get_best_trial(self, study_id):
...
+            :exc:`RuntimeError`:
+                If the trial is already finished.
+        """
         if trial_state.is_finished():
             trial = self.get_trial(trial_id)
             raise RuntimeError(
                 "Trial#{} has already finished and can not be updated.".format(trial.number)
             )
+
+    def load(self, study_id: int) -> None:
```
sile, line 605: Do we need to introduce this new method in this PR?
I think that the responsibility of this PR should be simply adding the doc to **the existing methods**. So, IMO, it feels better to separate this part into another PR (and it could be easier to review if the implementation and usage code of this method are also included in the PR).

ytsmiling: I left this method because an internal committee discussion reached an agreement that we can introduce the method here, but my opinion is neutral.
I generally agree the opinion and also don't think it's a problem to change the prior decision.
I'll remove the method later.

sile: > an internal committer discussion reached an agreement that we can introduce the method here.
> ...
> I'll remove the method later.

Oh, sorry for my comment disregarding the previous decision, but I would appreciate it if you could do it.

ytsmiling: Removed the method in https://github.com/optuna/optuna/pull/1175/commits/cfd6f2128ced8a5644b66ce4040133b0e460b687.

sile: Thanks!

optuna/storages/base.py
```diff
@@ -29,158 +28,494 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+
+        Args:
+            trial_id:
+                Trial ID of a trial to update its state.
+            state:
+                New state of the trial.
+
+        Returns:
+            :py:obj:`True` if the state is successfully updated.
+            :py:obj:`False` if the state is kept the same.
```
hvy, line 273: A detail but I don't think you need the `:py` part?

optuna/storages/base.py
```diff
@@ -29,158 +28,494 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
         raise NotImplementedError
 
     @abc.abstractmethod
-    def get_all_study_summaries(self):
-        # type: () -> List[study.StudySummary]
+    def get_all_study_summaries(self) -> List[study.StudySummary]:
+        """Read a list of `study.StudySummary` objects.
+
+        Returns:
+            A list of `study.StudySummary` objects.
```
hvy, line 228: Why not make it a proper reference?

optuna/storages/base.py
```diff
@@ -29,158 +28,494 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+            trial_id:
+                Trial ID of a trial to update its state.
+            state:
+                New state of the trial.
+
+        Returns:
+            :py:obj:`True` if the state is successfully updated.
+            :py:obj:`False` if the state is kept the same.
+            The latter happens when this method tries to update the state of
+            RUNNING trial to RUNNING.
```
hvy, line 275: Let's make those proper references as well.

optuna/storages/base.py
```diff
@@ -29,158 +28,494 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
-    def get_n_trials(self, study_id, state=None):
-        # type: (int, Optional[TrialState]) -> int
+    def get_n_trials(self, study_id: int, state: Optional[TrialState] = None) -> int:
+        """Count the number of trials in a specified study.
+
+        Args:
+            study_id:
+                Study ID of a study to count trials.
+            state:
+
```
hvy, line 492: Ops, missing explanation?

optuna/storages/base.py
```diff
@@ -29,158 +28,494 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
-    def delete_study(self, study_id):
-        # type: (int) -> None
+        Args:
+            study_name:
+                Name of a new study to create.
 
+        Returns:
+            Study ID, which is an unique id among studies, of the created study.
+
+        Raises:
```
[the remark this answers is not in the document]
hvy: We (or at least I) haven't really paid much attention to documenting exceptions, but we definitely should.

ytsmiling:
@hvy Thank you for your review. I appreciate your follow-up PR suggestion, and I'd like to see it.

optuna/storages/base.py
```diff
@@ -29,158 +28,494 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+            :exc:`optuna.exceptions.DuplicatedStudyError`:
+                If a study with the same `study_name` already exists.
```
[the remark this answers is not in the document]
ytsmiling: Thank you. I fixed all of them I could found.

optuna/storages/base.py
```diff
@@ -29,158 +28,494 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+            study_id:
+                Study ID of a study to update.
+            direction:
+                Either StudyDirection.MAXIMIZE or StudyDirection.MINIMIZE.
+
+        Raises:
+            :exc:`KeyError`:
+                If no study with the matching `study_id` exists.
+            :exc:`ValueError`:
+                If `direction` is already set and the passed `direction` conflicts with it.
```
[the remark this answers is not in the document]
ytsmiling: Thank you. I made the point explicit.

optuna/storages/base.py
```diff
@@ -29,158 +28,494 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+        """
+        raise NotImplementedError
 
+    @abc.abstractmethod
+    def set_study_system_attr(self, study_id: int, key: str, value: Any) -> None:
+        """Register an ontuna-internal attribute to a study.
```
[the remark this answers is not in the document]
ytsmiling: Thank you. Nich cache.

optuna/storages/base.py
```diff
@@ -29,158 +28,494 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
-    def set_study_system_attr(self, study_id, key, value):
-        # type: (int, str, Any) -> None
+    def set_study_direction(self, study_id: int, direction: study.StudyDirection) -> None:
+        """Register a direction of optimization problem to a study.
+
+        Args:
+            study_id:
+                Study ID of a study to update.
+            direction:
+                Either StudyDirection.MAXIMIZE or StudyDirection.MINIMIZE.
```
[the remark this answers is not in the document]
ytsmiling: Thank you. I fixed the references.

optuna/storages/base.py
ytsmiling: Removed.

optuna/storages/base.py
ytsmiling: Thank you. I fixed them.

optuna/storages/base.py
ytsmiling: Thank you. I added references.

optuna/storages/base.py
ytsmiling: Thank you. I added the missing doc.

optuna/storages/base.py
```diff
@@ -29,158 +28,494 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+        Raises:
+            :exc:`KeyError`:
+                If no study with the matching `study_id` exists.
+            :exc:`ValueError`:
+                If `direction` is already set and the passed `direction` conflicts with it.
+        """
+        raise NotImplementedError
 
+    @abc.abstractmethod
+    def set_study_system_attr(self, study_id: int, key: str, value: Any) -> None:
```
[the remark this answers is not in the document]
ytsmiling: I agree with that and changed the order.

optuna/storages/base.py
```diff
@@ -29,158 +28,497 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
-        # type: (int, Optional[TrialState]) -> int
+    def get_all_trials(self, study_id: int, deepcopy: bool = True) -> List["FrozenTrial"]:
+        """Read all trials in a specified study.
+
+        Args:
+            study_id:
+                Study ID of a study to read trials from.
+            deepcopy:
+                Whether copy the list of trials before returning.
+                Set :py:obj:`True` when you might update the list or elements of the list.
```
hvy, line 475:
```suggestion
                Set :obj:`True` when you might update the list or elements of the list.
```

optuna/storages/base.py
```diff
@@ -29,158 +28,497 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+        Args:
+            trial_id:
+                Trial ID of a trial to set intermediate value.
+            step:
+                Step of the trial (e.g., Epoch of neural network training).
+            intermediate_value:
+                Reported value within an evaluation of an objective function.
+
+        Returns:
+            Return :py:obj:`False` when the intermediate of the step already exists.
```
hvy, line 393:
```suggestion
            Return :obj:`False` when the intermediate of the step already exists.
```

optuna/storages/base.py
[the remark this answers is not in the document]
hvy: Thanks!

optuna/storages/base.py
```diff
@@ -29,158 +28,497 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+                Study ID of a study to update.
+            direction:
+                Either :obj:`~optuna.study.StudyDirection.MAXIMIZE` or
+                :obj:`~optuna.study.StudyDirection.MINIMIZE`.
+
+        Raises:
+            :exc:`KeyError`:
+                If no study with the matching ``study_id`` exists.
+            :exc:`ValueError`:
+                If `direction` is already set and the passed ``direction`` is the opposite
```
hvy, line 120: Might be missing something but should this not be plain text?
```suggestion
                If direction is already set and the passed ``direction`` is the opposite
```

optuna/storages/base.py
```diff
@@ -194,29 +532,73 @@ def get_best_trial(self, study_id):
...
+        return self.get_trial(trial_id).system_attrs
+
+    def remove_session(self) -> None:
+        """Clean up all connections to a database."""
         pass
 
-    def check_trial_is_updatable(self, trial_id, trial_state):
-        # type: (int, TrialState) -> None
+    def check_trial_is_updatable(self, trial_id: int, trial_state: TrialState) -> None:
+        """Check whether a `trial_state` is updatable.
```
hvy, line 589: How about two backticks, or as plain text
```suggestion
        """Check whether a trial state is updatable.
```

PR #1177, opened 2020-04-28
Use `Trial.system_attrs` to store `LightGBMTuner`'s results.

toshihikoyanase:
## Motivation

Currently, `LightGBMTuner` uses `user_attrs` to store some results such as `lightgbm_tuner:elapsed_secs` and `lightgbm_tuner:lgbm_params`. I think the `LightGBMTuner` is a part of Optuna system and I think `system_attrs` is preferable to store them.

`optuna.samplers.CmaEsSampler` also use `system_attrs` to save the information (c.f., https://github.com/optuna/optuna/blob/master/optuna/samplers/cmaes.py#L204).

## Description of the changes

This PR simply replace `user_attrs` with `system_attrs` in `LightGBMTuner`.

### DIFF
--- a/optuna/integration/lightgbm_tuner/optimize.py
+++ b/optuna/integration/lightgbm_tuner/optimize.py
@@ -277,10 +277,10 @@ class OptunaObjective(BaseTuner):
             )
         )
 
-        trial.set_user_attr("lightgbm_tuner:elapsed_secs", elapsed_secs)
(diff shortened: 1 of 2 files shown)


hvy:
We should preferably get this change merged in 1.4.0 as they were first introduced in 1.4.0 (https://github.com/optuna/optuna/pull/1032)

PR #1179, opened 2020-04-28
Remove `_TimeKeeper` and use `timeout` of `Study.optimize`.

toshihikoyanase:
## Motivation

Currently, `LightGBMTuner` uses `_TimeKeeper` to provide `time_budget` feature. The study inside `LightGBMTuner` also has a similar feature (i.e., `timeout`), and I'd like to unify them.

## Description of the changes

This PR replaces `_TimeKeeper` with `timeout` of `Study.optimize`. This update brings the following behavior changes:

- Current master: Optimization will be only stopped at the end of a step and never stopped during a step.
- This PR: Optimization will be stopped at the end of a trial. It will be stopped during a step.

Both implementations may exceed given `time_budget`s, but the time difference of this PR is usually smaller than that of the master.

### DIFF
--- a/optuna/integration/lightgbm_tuner/optimize.py
+++ b/optuna/integration/lightgbm_tuner/optimize.py
@@ -1,4 +1,3 @@
-import contextlib
 import copy
 import json
 import os
@@ -51,26 +50,6 @@ DEFAULT_LIGHTGBM_PARAMETERS = {
 }
 
 _logger = optuna.logging.get_logger(__name__)
-
-
-class _TimeKeeper(object):
-    def __init__(self):
-        # type: () -> None
-
-        self.time = time.time()
-
-    def elapsed_secs(self):
-        # type: () -> float
-
-        return time.time() - self.time
-
-
-@contextlib.contextmanager
-def _timer():
-    # type: () -> Generator[_TimeKeeper, None, None]
-
-    timekeeper = _TimeKeeper()
-    yield timekeeper
 
 
 class BaseTuner(object):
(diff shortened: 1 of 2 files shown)


optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -382,6 +361,7 @@ def __init__(
         )  # type: Dict[str, Any]
         self._parse_args(*args, **kwargs)
         self._best_booster_with_trial_number = None  # type: Optional[Tuple[lgb.Booster, int]]
+        self.start_time = None  # type: Optional[float]
```
smly, line 364: [IMO] how about making this attribute private?

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -689,9 +647,17 @@ def tune_params(self, target_param_names, n_trials, sampler, step_name):
             if t.state in (optuna.trial.TrialState.COMPLETE, optuna.trial.TrialState.PRUNED)
         ]
         _n_trials = n_trials - len(complete_trials)
+
+        if self.start_time is None:
+            self.start_time = time.time()
```
smly, line 652: [question] Is there any reason to initialize `self.start_time` in `tune_params()`?
I think it's simpler to initialize it in `__init__()`.

toshihikoyanase: Thank you for your question.
I think users may have a gap of `time_budget`. For example, users can add time-consuming computation between `__init__` and `run()`. In that case, the actual time limitation will be shorter than the value of `time_budget`. So, I initialize `self.start_time` right before the `study.optimize`.

But it makes implementation complicated, and I think we can ignore the gap.

optuna/integration/lightgbm_tuner/optimize.py
toshihikoyanase: That makes sense. I'll fix it.

smly: Thanks! Resolved by 3e09fc4.

optuna/integration/lightgbm_tuner/optimize.py
smly: Thanks for your explanation. It makes sense. Looks good to me on this part.

PR #1182, opened 2020-04-29
Prefer `math.log` with base argument in `HyperbandPruner`.

hvy:
## Motivation

Simple is better than complex. (It's also faster as it's implemented in C, https://github.com/python/cpython/blob/master/Modules/mathmodule.c#L2347.)

## Description of the changes

Make use of the second argument in `math.log` for a non-trivial base.


hvy:
Let me close this PR as it most likely will be included in https://github.com/optuna/optuna/pull/1188.

PR #1184, opened 2020-04-29
Add pruning interval to TFKerasPruningCallback.

VladSkripniuk:
Address #1136 .


brethvoice:
If we implement the following feature request, this pull request might be obsolete?  What do you think @VladSkripniuk? 
https://github.com/optuna/optuna/issues/1936

PR #1192, opened 2020-04-30
Define key names of `system_attrs` as variables in `LightGBMTuner`.

toshihikoyanase:
## Motivation
`LightGBMTuner` uses a string literals for the keys of `system_attrs`, but it may cause mistakes in future development. 

## Description of the changes
This PR defines the key names as variables, similarly to the #1163 ([here](https://github.com/optuna/optuna/pull/1163/files#diff-f8d15e6875544e214162a5701f52fd61R19)).

### DIFF
--- a/optuna/integration/lightgbm_tuner/optimize.py
+++ b/optuna/integration/lightgbm_tuner/optimize.py
@@ -33,6 +33,11 @@ if type_checking.TYPE_CHECKING:
     VALID_SET_TYPE = Union[List[lgb.Dataset], Tuple[lgb.Dataset, ...], lgb.Dataset]
 
 
+ELAPSED_SECS_KEY = "lightgbm_tuner:elapsed_secs"
+AVERAGE_ITERATION_TIME_KEY = "lightgbm_tuner:average_iteration_time"
+STEP_NAME_KEY = "lightgbm_tuner:step_name"
+LGBM_PARAMS_KEY = "lightgbm_tuner:lgbm_params"
+
 # EPS is used to ensure that a sampled parameter value is in pre-defined value range.
 EPS = 1e-12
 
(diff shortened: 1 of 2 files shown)


optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -33,6 +33,11 @@
     VALID_SET_TYPE = Union[List[lgb.Dataset], Tuple[lgb.Dataset, ...], lgb.Dataset]
 
 
+ELAPSED_SECS_KEY = "lightgbm_tuner:elapsed_secs"
+AVERAGE_ITERATION_TIME_KEY = "lightgbm_tuner:average_iteration_time"
+STEP_NAME_KEY = "lightgbm_tuner:step_name"
+LGBM_PARAMS_KEY = "lightgbm_tuner:lgbm_params"
```
hvy, line 39: I think these can be prefixed with underscores since they aren't public. What do you think?

hvy, line 39: Do you think the following is too verbose? I personally find it easier to comprehend as the intended use is clear, but it's not a strong opinion.
```suggestion
LGBM_PARAMS_SYSTEM_ATTR_KEY = "lightgbm_tuner:lgbm_params"
```

toshihikoyanase: You are right. Users do not use these variables and we can make them private.
Fixed: https://github.com/optuna/optuna/pull/1192/commits/7d606c815b9ac8a2c5c3b468f5aa1c730938aa90

toshihikoyanase: IMO, the following line is a bit complicated because `system_attr` appears two times (if we ignore the case):

```python
 trial.set_system_attr(_LGBM_PARAMS_SYSTEM_ATTR_KEY, json.dumps(self.lgbm_params))
```

So, I personally prefer `LGBM_PARAMS_KEY`. Instead I add a comment when I define variables:

```python
# Define key names of `Trial.system_attrs`.
_ELAPSED_SECS_KEY = "lightgbm_tuner:elapsed_secs"
...
```

What do you think of these changes?

hvy: Thanks, the comments indeed help. LGTM!

PR #1193, opened 2020-04-30
Stop installing PyTorch Lightning if Python version is 3.5.

toshihikoyanase:
## Motivation

This is a follow-up PR for #1013 . PyTorch Lightning does not support Python 3.5 but is installed via `get_extras_require()` in `setup.py`.

## Description of the changes

This PR stops installing PyTorch Lightning if Python version is 3.5.x. CI jobs skip PyTorch Lightning, so we do not update the [`.github/workflows/examples.yml`](https://github.com/optuna/optuna/blob/master/.github/workflows/examples.yml#L39).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,7 +86,7 @@ def get_extras_require() -> Dict[str, List[str]]:
             ["llvmlite<=0.31.0"] if (3, 5) == sys.version_info[:2] else []
         )  # Newer `llvmlite` is not distributed with wheels for Python 3.5.
         + (
-            ["dask[dataframe]", "dask-ml", "keras", "tensorflow>=2.0.0", "tensorflow-datasets",]
+            ["dask[dataframe]", "dask-ml", "keras", "tensorflow>=2.0.0", "tensorflow-datasets"]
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -77,19 +77,16 @@ def get_extras_require() -> Dict[str, List[str]]:
...
         + (
-            [
-                "dask[dataframe]",
-                "dask-ml",
-                "keras",
-                "pytorch-lightning>=0.7.1",
-                "tensorflow>=2.0.0",
-                "tensorflow-datasets",
-            ]
+            ["dask[dataframe]", "dask-ml", "keras", "tensorflow>=2.0.0", "tensorflow-datasets",]
```
hvy, line 89:
```suggestion
            ["dask[dataframe]", "dask-ml", "keras", "tensorflow>=2.0.0", "tensorflow-datasets"]
```

toshihikoyanase: Thank you! I fixed it.

hvy:
Thanks for the quick fix, let me merge this PR.

PR #1194, opened 2020-04-30
Add pruning interval to MXNetPruningCallback.

VladSkripniuk:
Address #1136 .

### DIFF
--- a/optuna/integration/mxnet.py
+++ b/optuna/integration/mxnet.py
@@ -26,18 +26,26 @@ class MXNetPruningCallback(object):
             default metric name. For custom metrics, use the metric_name provided to
             constructor. Please refer to `mxnet.metrics reference
             <https://mxnet.apache.org/api/python/metric/metric.html>`_ for further details.
+        interval:
+            Check if trial should be pruned every n-th epoch. By default ``interval=1`` and
+            pruning is performed after every epoch. Increase ``interval`` to run several
+            epochs faster before applying pruning.
     """
 
-    def __init__(self, trial, eval_metric):
-        # type: (optuna.trial.Trial, str) -> None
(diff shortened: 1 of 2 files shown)


optuna/integration/mxnet.py
```diff
@@ -26,19 +26,27 @@ class MXNetPruningCallback(object):
...
         _check_mxnet_availability()
 
         self._trial = trial
         self._eval_metric = eval_metric
+        self._interval = interval
 
     def __call__(self, param):
         # type: (mx.model.BatchEndParams,) -> None
 
+        if param.epoch % self._interval != 0:
```
keisuke-umezawa, line 47: Same with https://github.com/optuna/optuna/pull/1195#discussion_r419275815 .

VladSkripniuk: MXNet uses 0-based indexing for epochs so `if param.epoch % self._interval != 0:` would try to prune training after the very first trial ([link](https://github.com/apache/incubator-mxnet/blob/0580200562a394060520d85621f32fbae20d11c4/python/mxnet/module/base_module.py#L568)). Interval between successive pruning attempts still equals `self._interval`.

MXNet does not support intervals for callbacks, but in other discussion it was [mentioned](https://github.com/optuna/optuna/pull/1162#discussion_r416424619) that XGBoost includes iteration 0 for callbacks with intervals. Probably it's better to follow this logic and call pruning after epochs 0, n, 2n...

Crissman:
@VladSkripniuk This PR is stale and has conflicts with master, so closing for now. If there's further interest, we may want to take a look at https://github.com/optuna/optuna/pull/1161#issuecomment-625598685

PR #1195, opened 2020-04-30
Add pruning interval to PyTorchIgnitePruningHandler.

VladSkripniuk:
Address #1136 .

### DIFF
--- a/optuna/integration/pytorch_ignite.py
+++ b/optuna/integration/pytorch_ignite.py
@@ -30,17 +30,25 @@ class PyTorchIgnitePruningHandler(object):
         trainer:
             A trainer engine of PyTorch Ignite. Please refer to `ignite.engine.Engine reference
             <https://pytorch.org/ignite/engine.html#ignite.engine.Engine>`_ for further details.
+        interval:
+            Check if trial should be pruned every n-th epoch. By default `interval=1` and
+            pruning is performed after every epoch. Increase `interval` to run several
+            epochs faster before applying pruning.
     """
 
-    def __init__(self, trial, metric, trainer):
-        # type: (Trial, str, Engine) -> None
+    def __init__(self, trial, metric, trainer, interval=1):
+        # type: (Trial, str, Engine, int) -> None
 
         self._trial = trial
         self._metric = metric
         self._trainer = trainer
+        self._interval = interval
 
     def __call__(self, engine):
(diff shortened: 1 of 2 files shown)


optuna/integration/pytorch_ignite.py
```diff
@@ -30,18 +30,26 @@ class PyTorchIgnitePruningHandler(object):
...
 
         self._trial = trial
         self._metric = metric
         self._trainer = trainer
+        self._interval = interval
 
     def __call__(self, engine):
         # type: (Engine) -> None
 
+        if self._trainer.state.epoch % self._interval != 0:
```
keisuke-umezawa, line 50: Just a question:
Why it is not `if (self._trainer.state.epoch + 1) % self._interval != 0:`? The reason is that  `__call__` is called after `self._trainer.state.epoch` is incremented?

In that case, `message = "Trial was pruned at {} epoch.".format(self._trainer.state.epoch)` should be below?

```
message = "Trial was pruned at {} epoch.".format(self._trainer.state.epoch - 1)
```

VladSkripniuk: Pytorch Ignite uses 1-based counting for epochs and iterations ([link](https://github.com/pytorch/ignite/blob/1eddc7176264c03607511467550114a69cb502b4/ignite/engine/events.py#L275)). So `if (self._trainer.state.epoch + 1) % self._interval != 0:` would call pruning after `self._interval-1` epochs. `if self._trainer.state.epoch % self._interval != 0:` would call pruning callback after running first `self._interval` epochs, which is more reasonable in my opinion. 

`self._trainer.state.epoch` is incremented at the very beginning of an epoch ([link](https://github.com/pytorch/ignite/blob/1eddc7176264c03607511467550114a69cb502b4/ignite/engine/engine.py#L675)) so I think it's not necessary to alter the `message`.

keisuke-umezawa: Thank you for the explanation!
Could you add this information as a comment? ``Pytorch Ignite uses 1-based counting for epochs and iterations (link)``.

Crissman:
@VladSkripniuk This PR is stale and has conflicts with master, so closing for now. If there's further interest, we may want to take a look at https://github.com/optuna/optuna/pull/1161#issuecomment-625598685

PR #1197, opened 2020-05-01
Update pytorch_lightning_simple.py

eecshope:
Wrong parameter for 'gpu': should be '0', a string, rather than 0, an integer

## Motivation

Bug Fixed. The former edition could not start a GPU training job because of the wrong parameter 'gpu'.
## Description of the changes

For the trainer, the value for parameter 'gpu' has been changed from an integer 0 to  a string '0'

### DIFF
--- a/examples/pytorch_lightning_simple.py
+++ b/examples/pytorch_lightning_simple.py
@@ -150,7 +150,7 @@ def objective(trial):
         val_percent_check=PERCENT_VALID_EXAMPLES,
         checkpoint_callback=checkpoint_callback,
         max_epochs=EPOCHS,
-        gpus=0 if torch.cuda.is_available() else None,
+        gpus='0' if torch.cuda.is_available() else None,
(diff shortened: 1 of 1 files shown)


himkt:
Thank you for your PR!

I read the documentation for PyTorch Lightning.
Integer and string are both accepted for a parameter `gpus`.
(ref. https://pytorch-lightning.readthedocs.io/en/latest/trainer.html#gpus)

I also went to see the implementation, `gpus` seems to be `Optional[Union[List[int], str, int]]`.
So I think we can use both `int` and `str`.
(ref. https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/trainer.py#L90)

sile:
Thank you for your contribution, but looking at https://github.com/optuna/optuna/pull/1197#issuecomment-623047874, the current code seems not a bug. So please let me close this PR.

PR #910, opened 2020-02-10
Add step parameter on IntUniformDistribution

hayata-yamamoto:
This PR is derived from #902. resolve #902

## Details of Implementations

* Add the q parameter on IntUniformDistribution 
* Modify suggest_int function on Trial
* Fixed codes this change influenced

## Questions

~1. IntUniformDistribution._contains~

~If it takes a value which is in [low, high] but not sampled, should it return False?~

~Talking of the following example, this distribution includes 2 in [low, high], but does not contain sampled candidate.~

```python 
iu = IntUniformDistribution(1, 10, 2) 
iu._contains(2) # True or False?
```

~2. default value of q parameter~ 

~I wonder whether q has to be optional argument or not. Since DiscreteUniformDistribution's q is a required parameter, I implemented this also as the required.~

~However, in terms of back-compatibility, it's seemed that an optional parameter is better because of raising errors after updating due to "Too few arguments".~

PR #1201, opened 2020-05-03
Implement log argument for suggest_int

nzw0301:
## Motivation

Solve (b) of https://github.com/optuna/optuna/issues/510 

## Description of the changes

- Add `log` argument to `suggest_int` as a counterpart of `suggest_float`
  - except for integrations 
- Implement `LogIntUniformDistribution` class
- Documentation of `LogIntUniformDistribution`

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -295,7 +295,7 @@ class IntUniformDistribution(BaseDistribution):
         return self.low <= value <= self.high
 
 
-class IntLogUniformDistribution(BaseDistribution):
+class IntLogUniformDistribution(IntUniformDistribution):
     """A uniform distribution on integers in the log domain.
 
     This object is instantiated by :func:`~optuna.trial.Trial.suggest_int`, and passed to
@@ -313,20 +313,12 @@ class IntLogUniformDistribution(BaseDistribution):
     def __init__(self, low, high, step):
         # type: (int, int, int) -> None
 
-        if low > high:
-            raise ValueError(
-                "The `low` value must be smaller than or equal to the `high` value "
-                "(low={}, high={}).".format(low, high)
-            )
+        super(IntLogUniformDistribution, self).__init__(low, high, step)
+
         if low <= 0.0:
             raise ValueError(
                 "The `low` value must be larger than 0 for a log distribution "
                 "(low={}, high={}).".format(low, high)
-            )
-
-        if step <= 0:
-            raise ValueError(
-                "The `step` value must be non-zero positive value, but step={}.".format(step)
             )
 
         if (high - low) % step != 0:
@@ -335,30 +327,10 @@ class IntLogUniformDistribution(BaseDistribution):
                 "low={}, high={}, step={}.".format(low, high, step)
             )
 
-        self.low = low
-        self.high = high
-        self.step = step
-
-    def to_external_repr(self, param_value_in_internal_repr):
-        # type: (float) -> int
-
-        return int(param_value_in_internal_repr)
-
-    def to_internal_repr(self, param_value_in_external_repr):
-        # type: (int) -> float
-
-        return float(param_value_in_external_repr)
-
     def single(self):
         # type: () -> bool
 
         return (self.high - self.low) % self.step == 0
-
-    def _contains(self, param_value_in_internal_repr):
-        # type: (float) -> bool
-
-        value = param_value_in_internal_repr
-        return self.low <= value < self.high
 
 
 class CategoricalDistribution(BaseDistribution):
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -306,10 +306,12 @@ class IntLogUniformDistribution(BaseDistribution):
             Lower endpoint of the range of the distribution. ``low`` is included in the range.
         high:
             Upper endpoint of the range of the distribution. ``high`` is included in the range.
+        step:
+            A step for spacing between values.
     """
 
-    def __init__(self, low, high):
-        # type: (float, float) -> None
+    def __init__(self, low, high, step):
+        # type: (int, int, int) -> None
 
         if low > high:
             raise ValueError(
@@ -322,8 +324,14 @@ class IntLogUniformDistribution(BaseDistribution):
                 "(low={}, high={}).".format(low, high)
             )
 
+        if step <= 0:
+            raise ValueError(
+                "The `step` value must be non-zero positive value, but step={}.".format(step)
+            )
+
         self.low = low
         self.high = high
+        self.step = step
 
     def to_external_repr(self, param_value_in_internal_repr):
         # type: (float) -> int
@@ -338,7 +346,9 @@ class IntLogUniformDistribution(BaseDistribution):
     def single(self):
         # type: () -> bool
 
-        return self.low == self.high
+        if self.low == self.high:
+            return True
+        return (self.high - self.low) < self.step
 
     def _contains(self, param_value_in_internal_repr):
         # type: (float) -> bool
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -332,6 +332,12 @@ class IntLogUniformDistribution(IntUniformDistribution):
 
         return self.low == self.high
 
+    def _contains(self, param_value_in_internal_repr):
+        # type: (float) -> bool
+
+        value = param_value_in_internal_repr
+        return (self.low <= value <= self.high) & (value % self.step == 0)
+
 
 class CategoricalDistribution(BaseDistribution):
     """A categorical distribution.
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -308,6 +308,11 @@ class IntLogUniformDistribution(BaseDistribution):
             Upper endpoint of the range of the distribution. ``high`` is included in the range.
         step:
             A step for spacing between values.
+
+    .. note::
+
+        ``high - low`` must be a multiple of ``step``. In other words, the
+        following equation must be satisfied for some integer ``k``: ``high = low + step * k``.
     """
 
     def __init__(self, low, high, step=1):
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -295,6 +295,58 @@ class IntUniformDistribution(BaseDistribution):
         return self.low <= value <= self.high
 
 
+class IntLogUniformDistribution(BaseDistribution):
+    """A uniform distribution on integers in the log domain.
+
+    This object is instantiated by :func:`~optuna.trial.Trial.suggest_int`, and passed to
+    :mod:`~optuna.samplers` in general.
+
+    Attributes:
+        low:
+            Lower endpoint of the range of the distribution. ``low`` is included in the range.
+        high:
+            Upper endpoint of the range of the distribution. ``high`` is included in the range.
+    """
+
+    def __init__(self, low, high):
+        # type: (float, float) -> None
+
+        if low > high:
+            raise ValueError(
+                "The `low` value must be smaller than or equal to the `high` value "
+                "(low={}, high={}).".format(low, high)
+            )
+        if low <= 0.0:
+            raise ValueError(
+                "The `low` value must be larger than 0 for a log distribution "
+                "(low={}, high={}).".format(low, high)
+            )
+
+        self.low = low
+        self.high = high
+
+    def to_external_repr(self, param_value_in_internal_repr):
+        # type: (float) -> int
+
+        return int(param_value_in_internal_repr)
+
+    def to_internal_repr(self, param_value_in_external_repr):
+        # type: (int) -> float
+
+        return float(param_value_in_external_repr)
+
+    def single(self):
+        # type: () -> bool
+
+        return self.low == self.high
+
+    def _contains(self, param_value_in_internal_repr):
+        # type: (float) -> bool
+
+        value = param_value_in_internal_repr
+        return self.low <= value < self.high
+
+
 class CategoricalDistribution(BaseDistribution):
     """A categorical distribution.
 
@@ -367,6 +419,7 @@ DISTRIBUTION_CLASSES = (
     LogUniformDistribution,
     DiscreteUniformDistribution,
     IntUniformDistribution,
+    IntLogUniformDistribution,
     CategoricalDistribution,
 )
 
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -330,7 +330,7 @@ class IntLogUniformDistribution(IntUniformDistribution):
     def single(self):
         # type: () -> bool
 
-        return (self.high - self.low) % self.step == 0
+        return self.low == self.high
 
 
 class CategoricalDistribution(BaseDistribution):
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -352,9 +352,7 @@ class IntLogUniformDistribution(BaseDistribution):
     def single(self):
         # type: () -> bool
 
-        if self.low == self.high:
-            return True
-        return (self.high - self.low) < self.step
+        return (self.high - self.low) % self.step == 0
 
     def _contains(self, param_value_in_internal_repr):
         # type: (float) -> bool
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -310,7 +310,7 @@ class IntLogUniformDistribution(IntUniformDistribution):
             A step for spacing between values.
     """
 
-    def __init__(self, low, high, step):
+    def __init__(self, low, high, step=1):
         # type: (int, int, int) -> None
 
         super(IntLogUniformDistribution, self).__init__(low, high, step)
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -308,11 +308,6 @@ class IntLogUniformDistribution(BaseDistribution):
             Upper endpoint of the range of the distribution. ``high`` is included in the range.
         step:
             A step for spacing between values.
-
-    .. note::
-
-        ``high - low`` must be a multiple of ``step``. In other words, the
-        following equation must be satisfied for some integer ``k``: ``high = low + step * k``.
     """
 
     def __init__(self, low, high, step=1):
@@ -334,12 +329,6 @@ class IntLogUniformDistribution(BaseDistribution):
                 "(low={}, high={}).".format(low, high)
             )
 
-        if (high - low) % step != 0:
-            raise ValueError(
-                "The `high` - `low` must be a multiple of `step`, but "
-                "low={}, high={}, step={}.".format(low, high, step)
-            )
-
         self.low = low
         self.high = high
         self.step = step
@@ -357,13 +346,15 @@ class IntLogUniformDistribution(BaseDistribution):
     def single(self):
         # type: () -> bool
 
-        return self.low == self.high
+        if self.low == self.high:
+            return True
+        return (self.high - self.low) < self.step
 
     def _contains(self, param_value_in_internal_repr):
         # type: (float) -> bool
 
         value = param_value_in_internal_repr
-        return (self.low <= value <= self.high) & (value % self.step == 0)
+        return self.low <= value <= self.high
 
 
 class CategoricalDistribution(BaseDistribution):
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -295,7 +295,7 @@ class IntUniformDistribution(BaseDistribution):
         return self.low <= value <= self.high
 
 
-class IntLogUniformDistribution(IntUniformDistribution):
+class IntLogUniformDistribution(BaseDistribution):
     """A uniform distribution on integers in the log domain.
 
     This object is instantiated by :func:`~optuna.trial.Trial.suggest_int`, and passed to
@@ -313,7 +313,15 @@ class IntLogUniformDistribution(IntUniformDistribution):
     def __init__(self, low, high, step=1):
         # type: (int, int, int) -> None
 
-        super(IntLogUniformDistribution, self).__init__(low, high, step)
+        if low > high:
+            raise ValueError(
+                "The `low` value must be smaller than or equal to the `high` value "
+                "(low={}, high={}).".format(low, high)
+            )
+        if step <= 0:
+            raise ValueError(
+                "The `step` value must be non-zero positive value, but step={}.".format(step)
+            )
 
         if low <= 0.0:
             raise ValueError(
@@ -326,6 +334,20 @@ class IntLogUniformDistribution(IntUniformDistribution):
                 "The `high` - `low` must be a multiple of `step`, but "
                 "low={}, high={}, step={}.".format(low, high, step)
             )
+
+        self.low = low
+        self.high = high
+        self.step = step
+
+    def to_external_repr(self, param_value_in_internal_repr):
+        # type: (float) -> int
+
+        return int(param_value_in_internal_repr)
+
+    def to_internal_repr(self, param_value_in_external_repr):
+        # type: (int) -> float
+
+        return float(param_value_in_external_repr)
 
     def single(self):
         # type: () -> bool
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -329,6 +329,12 @@ class IntLogUniformDistribution(BaseDistribution):
                 "The `step` value must be non-zero positive value, but step={}.".format(step)
             )
 
+        if (high - low) % step != 0:
+            raise ValueError(
+                "The `high` - `low` must be a multiple of `step`, but "
+                "low={}, high={}, step={}.".format(low, high, step)
+            )
+
         self.low = low
         self.high = high
         self.step = step
--- a/optuna/importance/_fanova.py
+++ b/optuna/importance/_fanova.py
@@ -101,10 +101,10 @@ def _distribution_to_hyperparameter(name: str, distribution: BaseDistribution) -
         hp = UniformFloatHyperparameter(name, lower=d.low, upper=d.high, log=True)
     elif isinstance(d, DiscreteUniformDistribution):
         hp = UniformFloatHyperparameter(name, lower=d.low, upper=d.high, q=d.q)
+    elif isinstance(d, IntLogUniformDistribution):
+        hp = UniformIntegerHyperparameter(name, lower=d.low, upper=d.high, q=d.step, log=True)
     elif isinstance(d, IntUniformDistribution):
         hp = UniformIntegerHyperparameter(name, lower=d.low, upper=d.high, q=d.step)
-    elif isinstance(d, IntLogUniformDistribution):
-        hp = UniformIntegerHyperparameter(name, lower=d.low, upper=d.high, log=True)
     elif isinstance(d, CategoricalDistribution):
         hp = CategoricalHyperparameter(name, choices=[d.to_internal_repr(c) for c in d.choices])
     else:
--- a/optuna/importance/_fanova.py
+++ b/optuna/importance/_fanova.py
@@ -8,6 +8,7 @@ from optuna.distributions import BaseDistribution
 from optuna.distributions import CategoricalDistribution
 from optuna.distributions import DiscreteUniformDistribution
 from optuna.distributions import IntUniformDistribution
+from optuna.distributions import IntLogUniformDistribution
 from optuna.distributions import LogUniformDistribution
 from optuna.distributions import UniformDistribution
 from optuna.importance._base import _get_distributions
@@ -102,6 +103,8 @@ def _distribution_to_hyperparameter(name: str, distribution: BaseDistribution) -
         hp = UniformFloatHyperparameter(name, lower=d.low, upper=d.high, q=d.q)
     elif isinstance(d, IntUniformDistribution):
         hp = UniformIntegerHyperparameter(name, lower=d.low, upper=d.high, q=d.step)
+    elif isinstance(d, IntLogUniformDistribution):
+        hp = UniformIntegerHyperparameter(name, lower=d.low, upper=d.high, log=True)
     elif isinstance(d, CategoricalDistribution):
         hp = CategoricalHyperparameter(name, choices=[d.to_internal_repr(c) for c in d.choices])
     else:
@@ -110,6 +113,7 @@ def _distribution_to_hyperparameter(name: str, distribution: BaseDistribution) -
             LogUniformDistribution.__name__,
             DiscreteUniformDistribution.__name__,
             IntUniformDistribution.__name__,
+            IntLogUniformDistribution.__name__,
             CategoricalDistribution.__name__,
         ]
         raise NotImplementedError(
--- a/optuna/importance/_fanova.py
+++ b/optuna/importance/_fanova.py
@@ -7,8 +7,8 @@ from optuna._experimental import experimental
 from optuna.distributions import BaseDistribution
 from optuna.distributions import CategoricalDistribution
 from optuna.distributions import DiscreteUniformDistribution
-from optuna.distributions import IntUniformDistribution
 from optuna.distributions import IntLogUniformDistribution
+from optuna.distributions import IntUniformDistribution
 from optuna.distributions import LogUniformDistribution
 from optuna.distributions import UniformDistribution
 from optuna.importance._base import _get_distributions
--- a/optuna/importance/_fanova.py
+++ b/optuna/importance/_fanova.py
@@ -101,10 +101,10 @@ def _distribution_to_hyperparameter(name: str, distribution: BaseDistribution) -
         hp = UniformFloatHyperparameter(name, lower=d.low, upper=d.high, log=True)
     elif isinstance(d, DiscreteUniformDistribution):
         hp = UniformFloatHyperparameter(name, lower=d.low, upper=d.high, q=d.q)
-    elif isinstance(d, IntLogUniformDistribution):
-        hp = UniformIntegerHyperparameter(name, lower=d.low, upper=d.high, q=d.step, log=True)
     elif isinstance(d, IntUniformDistribution):
         hp = UniformIntegerHyperparameter(name, lower=d.low, upper=d.high, q=d.step)
+    elif isinstance(d, IntLogUniformDistribution):
+        hp = UniformIntegerHyperparameter(name, lower=d.low, upper=d.high, q=d.step, log=True)
     elif isinstance(d, CategoricalDistribution):
         hp = CategoricalHyperparameter(name, choices=[d.to_internal_repr(c) for c in d.choices])
     else:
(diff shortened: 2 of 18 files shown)


optuna/importance/_fanova.py
```diff
@@ -101,7 +102,9 @@ def _distribution_to_hyperparameter(name: str, distribution: BaseDistribution) -
     elif isinstance(d, DiscreteUniformDistribution):
         hp = UniformFloatHyperparameter(name, lower=d.low, upper=d.high, q=d.q)
     elif isinstance(d, IntUniformDistribution):
-        hp = UniformIntegerHyperparameter(name, lower=d.low, upper=d.high)
+        hp = UniformIntegerHyperparameter(name, lower=d.low, upper=d.high, q=d.step)
```
nzw0301, line 105: Is it okay to pass `q` to `d.step`?

optuna/multi_objective/trial.py
```diff
@@ -42,14 +42,16 @@ def __init__(self, trial: Trial):
         # See also: https://github.com/optuna/optuna/pull/1054/files#r407982636
         self._n_objectives = multi_objective.study.MultiObjectiveStudy(trial.study).n_objectives
 
-    def suggest_float(self, name: str, low: float, high: float, *, log: bool = False) -> float:
+    def suggest_float(
+            self, name: str, low: float, high: float, *, log: bool = False, step: Optional[float] = None
```
nzw0301, line 46: @himkt Is it okay?

nzw0301:
Sorry, it seems a little bit complicated for me to add tests and integrations of `skopt` and `cma`. Please give me a little more time. 

In addition, I should implement `step` argument when `log=True`, which is available in as in [`automl/ConfigSpace`](https://github.com/automl/ConfigSpace/blob/master/ConfigSpace/hyperparameters.pyx#L764).

optuna/trial.py
```diff
@@ -1253,11 +1260,18 @@ def suggest_discrete_uniform(self, name, low, high, q):
         discrete = distributions.DiscreteUniformDistribution(low=low, high=high, q=q)
         return self._suggest(name, discrete)
 
-    def suggest_int(self, name, low, high, step=1):
-        # type: (str, int, int, int) -> int
-        sample = self._suggest(
-            name, distributions.IntUniformDistribution(low=low, high=high, step=step)
-        )
+    def suggest_int(self, name, low, high, log=False, step=1):
```
crcrpar, line 1263: This is __NOT__ backward-compatible. We need to discuss the order with users and developers. I know this order aligns with `suggest_float` and makes sense BUT personally the backward compatibility is prioritized.
https://optuna.readthedocs.io/en/latest/reference/trial.html#optuna.trial.Trial.suggest_float

himkt: How about introducing `log` after the `step` to keep the backward compatibility and showing the warning `log` and `step` will be keyword-only arguments? I think it makes users have time to migrate from positional arguments to keyword-only arguments with keeping the compatibility. (it's just my opinion)

optuna/distributions.py
```diff
@@ -144,7 +144,7 @@ def _contains(self, param_value_in_internal_repr):
         if self.low == self.high:
             return value == self.low
         else:
-            return self.low <= value and value < self.high
```
himkt, line 147: It's not a strong opinion but how about creating another PR to fix code-style? (83f11e5 3943436 53ac93c)

optuna/trial.py
himkt: How about introducing `log` after the `step` to keep the backward compatibility and showing the warning `log` and `step` will be keyword-only arguments? I think it makes users have time to migrate from positional arguments to keyword-only arguments with keeping the compatibility. (it's just my opinion)

optuna/multi_objective/trial.py
himkt: It looks good to me.

nzw0301: Thanks! I should have clarified the modification. I have added `step: Optional[float] = None` as argument of `suggest_float`.

optuna/distributions.py
nzw0301: Thank you for your feedback. Sure

nzw0301: Thank you for your feedback. Sure. I have created https://github.com/optuna/optuna/pull/1203

optuna/distributions.py
```diff
@@ -297,6 +295,72 @@ def _contains(self, param_value_in_internal_repr):
         return self.low <= value <= self.high
 
 
+class IntLogUniformDistribution(BaseDistribution):
```
sile, line 298: I personally prefer extending the existing `IntDistribution` class instead of creating a new class (c.f. the `step` field added by https://github.com/optuna/optuna/pull/910).
What do you think of it?

himkt: Ah, I'm really sorry that I suggested implementing `IntLogUniformDistribution` [here](https://github.com/optuna/optuna/issues/510#issuecomment-622399411) without understanding [the historical context](https://github.com/optuna/optuna/issues/510#issuecomment-622867398). I support @sile since I'm not confident about the internal API design.

Really sorry @nzw0301 for confusing you.

sile: I see. Thank you for your explanation.

nzw0301: @sile  Thank you for your comment! It sounds great. I will update it.

@himkt Never mind :)

optuna/trial.py
nzw0301: Thank you both @crcrpar and @himkt for your comment. I agree with maintaining backward-compatibility. But I don't know how to do that. @himkt's suggestion sounds nice.

nzw0301:
I'll start to write tests.

optuna/trial.py
crcrpar: Yeah, keyword only argument sounds good.

nzw0301: I have changed them with keyword-only arguments.

nzw0301:
Sorry, I'm not familiar with skopt and cma. Could someone help me create tests for them?

nzw0301:
I have added all codes related to `suggest_int` with `log` argument except for tests for integrations.

optuna/trial.py
```diff
@@ -732,14 +732,21 @@ def objective(trial):
...
 
         Returns:
             A suggested integer value.
         """
 
-        distribution = distributions.IntUniformDistribution(low=low, high=high, step=step)
+        if log:
+            distribution = distributions.IntLogUniformDistribution(low=low, high=high, step=step)
+        else:
+            distribution = distributions.IntUniformDistribution(low=low, high=high, step=step)
```
nzw0301, line 749: To avoid mypy error, @himkt and I come up with two solutions:

**1: Annotate `IntUniformDistribution`**

```python
if log:
    distribution: IntUniformDistribution = distributions.IntLogUniformDistribution(low=low, high=high, step=step)
else:
    distribution = distributions.IntUniformDistribution(low=low, high=high, step=step)
```

**2: Change the order of inside**
```python
if not log:
    distribution = distributions.IntUniformDistribution(low=low, high=high, step=step)
else:
    distribution = distributions.IntLogUniformDistribution(low=low, high=high, step=step)
```

But we aren't sure these ways are fine...

crcrpar:
> 
> 
> Sorry, I'm not familiar with`skopt` and `cma` integrations. Could someone help me create tests for them?

I guess @sile, @toshihikoyanase, or @c-bata have written most part of tests for the two integrations.
Could you guys help @nzw0301?

toshihikoyanase:
@crcrpar Thank you for letting me know. I'll check the tests for `cma` integrations.

optuna/distributions.py
```diff
@@ -295,6 +295,50 @@ def _contains(self, param_value_in_internal_repr):
...
+            raise ValueError(
+                "The `low` value must be larger than 0 for a log distribution "
+                "(low={}, high={}).".format(low, high)
+            )
+
+        if (high - low) % step != 0:
+            raise ValueError(
+                "The `high` - `low` must be a multiple of `step`, but "
+                "low={}, high={}, step={}.".format(low, high, step)
+            )
```
nzw0301, line 328: This part comes from [automl's ConfigSpace](https://github.com/automl/ConfigSpace/blob/d36a6b275bdffb1382cc0820c602668d66396e5b/ConfigSpace/hyperparameters.pyx#L818-L822) that causes the same error. But it may be removed for consistency to `DiscreteUniformDistribution` in Optuna.

nzw0301:
As @crcrpar suggests to me privately, since this PR becomes too long, let me split this PR into three integrations (chainermn, skopt, and cma) and the rest. This PR aims to introduce suggest_int with log except for integrations.

toshihikoyanase:
> let me split this PR into three integrations (`chainermn`, `skopt`, and `cma`) and the rest. 

Sorry for the delayed response about `cma`. I think it's reasonable. I'm looking forward to your update.

nzw0301:
I have removed several commits related to`skopt`, `cma`, and `chainermn` integrations.

nzw0301:
If this PR is still too large for reviewers, please let me know.

himkt:
Let me check the test for `chainermn`.
I want someone to help a check for `skopt`, which seems to require a knowledge about its internal API.

nzw0301:
@himkt Thanks! I create PRs for tests if this PR is merged.

optuna/distributions.py
```diff
@@ -295,6 +295,50 @@ def _contains(self, param_value_in_internal_repr):
...
+    def single(self):
+        # type: () -> bool
+
+        return self.low == self.high
+
+    def _contains(self, param_value_in_internal_repr):
+        # type: (float) -> bool
+
+        value = param_value_in_internal_repr
+        return (self.low <= value <= self.high) & (value % self.step == 0)
```
nzw0301, line 339: `_contains` is also the same as https://github.com/optuna/optuna/pull/1201/files#r424421651

HideakiImamura:
@nzw0301 Thank you for the significant PR! I will cover the check for `skopt` integration. I'm looking forward to your PR.

c-bata:
Hi, there are three problems on `optuna.samplers.CmaEsSampler`.

1. IntLogUniformDistribution is skipped at `infer_relative_search_space`. So `IntLogUniformDistribution` will be sampled by independent sampler.
    1. https://github.com/optuna/optuna/pull/1201/files#diff-907fdf1824eea9f63ff37741a3416e60R129-R140
1. The parameters of IntLogUniformDistribution is sampled from uniform distribution because LogUniformDistribution is not converted.
    1. https://github.com/optuna/optuna/pull/1201/files#diff-907fdf1824eea9f63ff37741a3416e60R289-R310
1. It looks that you need to fix boundary constraints like #1243.

But I guess it's hard to support `optuna.samplers.CmaEsSampler` in this pull request. How about reverting a changes of CmaEsSampler and add a TODO comment?

nzw0301:
@c-bata Thank you for your comments and reviewing `optuna.samplers.CmaEsSampler`!

Sorry for my incorrect codes. It makes sense for me to do that.

optuna/samplers/cmaes.py
```diff
@@ -324,6 +324,10 @@ def _initialize_x0(search_space: Dict[str, BaseDistribution]) -> Dict[str, np.nd
             log_high = math.log(distribution.high)
```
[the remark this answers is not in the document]
HideakiImamura: Sorry for the confusion. If you raise `NotImplementedError`, my comments on `CmaEsSampler` can be ignored. (But, it can be resolved by my comments.)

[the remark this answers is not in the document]
c-bata: @HideakiImamura Actually, I proposed @nzw0301 to put a todo comment because it seems to be hard to support CmaEsSampler in this PR (https://github.com/optuna/optuna/pull/1201#issuecomment-629960251). What do you think?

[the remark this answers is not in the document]
c-bata: [IMO] I think it's enough just put a todo comment because it will be just fallbacked to independent sampler with warning messages.

optuna/integration/chainermn.py
```diff
@@ -227,14 +227,14 @@ def func():
 
         return self._call_with_mpi(func)
 
-    def suggest_int(self, name, low, high, step=1):
-        # type: (str, int, int, int) -> int
+    def suggest_int(self, name, low, high, *, log=False, step=1):
```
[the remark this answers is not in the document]
nzw0301: To maintain backward-compatible, we need `*` that is called [Keyword-only argument](https://www.python.org/dev/peps/pep-3102/). The discussion is found at  https://github.com/optuna/optuna/pull/1201#pullrequestreview-404664661.

optuna/samplers/cmaes.py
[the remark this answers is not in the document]
HideakiImamura: I agree with c-bata-san's opinion. For the `CmaEsSampler`, we do not need to raise `NotImplementedErroe`. Thank you for your pointing.

optuna/integration/chainermn.py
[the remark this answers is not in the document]
HideakiImamura: Thank you so much for sharing this information! I've understood it.

nzw0301:
Thank you @HideakiImamura for your great feedback!

Note that the current implementation is almost the same as automl's [UniformIntegerHyperparameter](https://automl.github.io/ConfigSpace/master/API-Doc.html#ConfigSpace.hyperparameters.UniformIntegerHyperparameter). Please also check an example in https://github.com/optuna/optuna/issues/510#issuecomment-587557701.
Thus, BEFORE convention makes sense. (But if optuna community needs your proposal feature rather than automl's sampling scheme, I will update this PR.)

I feel `suggest_int` with `log` argument is confusing one. We should carefully document it.

I also agree with your proposal. Several people expect such behaviour of this function (I also did).
As future work, introducing additional argument seems great for users. `suggest_int` with such argument (like `base`?) will sample integer `i` in uniform or log-uniform space then calculate `retuned_value = base**i` or by using your proposed way.

optuna/samplers/cmaes.py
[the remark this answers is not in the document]
nzw0301: I keep the changes and update them regarding @HideakiImamura 's comments.

optuna/samplers/random.py
```diff
@@ -76,6 +76,16 @@ def sample_independent(self, study, trial, param_name, param_distribution):
...
+        elif isinstance(param_distribution, distributions.IntLogUniformDistribution):
+            log_low = numpy.log(param_distribution.low - 0.5)
+            log_high = numpy.log(param_distribution.high + 0.5)
+            s = numpy.exp(self._rng.uniform(log_low, log_high))
+            v = (
+                numpy.round((s - param_distribution.low) / param_distribution.step)
+                * param_distribution.step
+                + param_distribution.low
+            )
+            return int(min(max(v, param_distribution.low), param_distribution.high))
```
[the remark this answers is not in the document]
nzw0301: Unfortunately, we cannot change the order. `IntLogUniformDistribuion` is a subclass of `IntUniformDistribuion` (https://github.com/optuna/optuna/pull/1201#discussion_r421198628), so `isinstance(param_distribution, IntLogUniformDistribution)` should be checked earler than `isinstance(param_distribution, IntUniformDistribution)`. Otherwise, `isinstance(param_distribution, IntLogUniformDistribution)` is never checked.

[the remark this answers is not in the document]
HideakiImamura: Thank you for your reply. I feel that the `IntLogUniformDistribution` should inherit `BaseDistribution` directly as other distributions do. Doesn't it seem simpler that way? The logic is a bit redundant, but it seems to be easier to maintain.

optuna/trial/_trial.py
```diff
@@ -384,14 +384,22 @@ def objective(trial):
...
+                Otherwise, the value is sampled from the range in the linear domain.
             step:
                 A step of spacing between values.
 
         Returns:
             A suggested integer value.
         """
 
-        distribution = distributions.IntUniformDistribution(low=low, high=high, step=step)
+        distribution: distributions.IntUniformDistribution
```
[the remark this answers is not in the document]
nzw0301: Looks great! But these syntax is only supported in Python 3.6 and greater. I'm sorry. I will change it.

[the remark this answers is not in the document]
nzw0301: I changed from it to

```python
if log:
    distribution = distributions.IntLogUniformDistribution(
        low=low, high=high, step=step
    )  # type: Union[distributions.IntUniformDistribution, distributions.IntLogUniformDistribution]
else:
    distribution = distributions.IntUniformDistribution(low=low, high=high, step=step)
```
Ref: https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#explicit-types-for-variables

optuna/samplers/random.py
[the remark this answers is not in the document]
nzw0301: I agree with you. That is easier to maintain for now. `if` blocks could cause an unexpected error by inheriting `IntUniformDistribution`. Even if `IntLogUniformDistribution` inherits `BaseDistribtuion`, less than 10 lines added to the code. That seems acceptable.

optuna/trial/_trial.py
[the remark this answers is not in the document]
nzw0301: Oh, the line is too long according to `flake8`. Sorry, I will fix it.

nzw0301:
Thank you @c-bata and @HideakiImamura again for reviewing this PR.

My update summary:

- Regarding suggestions, I updated `cmaes.py`. Do not raise `NotImplementedError` in `cmaes.py` and raise it in `integrations/cma.py`.
- `IntLogUniformDistribution` inherits `BaseDistribution`.


We need to decide the behaviour of `suggest_int` when `log=True` as mentioned by @HideakiImamura in https://github.com/optuna/optuna/pull/1201#pullrequestreview-413257824. Note the current implementation is like https://github.com/optuna/optuna/issues/510#issuecomment-587557701 and it is also the same as Automl's [ConfigSpace.hyperparameters.UniformIntegerHyperparameter](https://automl.github.io/ConfigSpace/master/API-Doc.html#ConfigSpace.hyperparameters.UniformIntegerHyperparameter).

nzw0301:
I put codes to show examples and comparison with automl's `ConfigSpace`.

```python
## Comparison between ConfigSpace's integer sampling in `log` space and new Optuna's `suggest_int(log=True)`

import numpy as np
import ConfigSpace
import ConfigSpace as CS
import ConfigSpace.hyperparameters as CSH

low = 2
high = 64
step = 2
num_samples = 10_000

cs = CS.ConfigurationSpace(seed=1)
uniform_integer_hp = CSH.UniformIntegerHyperparameter(name='x', lower=low, upper=high, log=True, q=step)
cs.add_hyperparameter(uniform_integer_hp)

samples = [cs.sample_configuration().get('x') for _ in range(num_samples)]
samples = np.array(samples)

plt.hist(samples, bins=8);

assert np.sum(samples % 2) == 0 # no odd numebrs.

# 2, 4, 6, ..., 64 generated
np.bincount(samples)

## Optuna

import optuna
from optuna import samplers
from optuna.study import create_study
from optuna.trial import Trial
from optuna.distributions import IntLogUniformDistribution

sampler = samplers.RandomSampler(seed=1)
distribuion = IntLogUniformDistribution(low=low, high=high, step=step)

samples = []
for i in range(num_samples):
    trial = Trial(study, study._storage.create_new_trial(study._study_id))    
    
    samples.append(trial._suggest("x", distribuion))

samples = np.array(samples)

plt.hist(samples, bins=8);

assert np.sum(samples % 2) == 0 # no odd numebrs.

# 2, 4, 6, ..., 64 generated
np.bincount(samples)
```

Both ConfigSpace's outputs and optuna's outputs are: 

## Histgram 

download

## 0 origin bincount

```python
array([   0,    0, 1826,    0, 1428,    0,  830,    0,  645,    0,  557,
          0,  493,    0,  388,    0,  318,    0,  295,    0,  285,    0,
        235,    0,  225,    0,  212,    0,  181,    0,  174,    0,  172,
          0,  137,    0,  153,    0,  138,    0,  144,    0,  139,    0,
        101,    0,  114,    0,  113,    0,  107,    0,   90,    0,   95,
          0,   75,    0,   90,    0,   99,    0,   87,    0,   54])
```

Jupyter notebooks is found at https://gist.github.com/nzw0301/f9eed6b38f6932b70ec9d7a3a5eee0dc

optuna/trial/_trial.py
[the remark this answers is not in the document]
nzw0301: Could I add the following statements in `optuna/trial/_trial.py` to shorten the typing line?

```python
from optuna.distribution import IntUniformDistribution
from optuna.distribution import IntLogUniformDistribution
``` 

[CircleCI](https://circleci.com/gh/optuna/optuna/41431?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) fails due to this line now.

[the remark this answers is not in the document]
HideakiImamura: I think we can import them from not `optuna.distribution` but `optuna.distributions`.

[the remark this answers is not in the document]
nzw0301: Thank you for the quick response. Sorry, it was a typo. I will import them.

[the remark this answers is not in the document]
nzw0301: Done

HideakiImamura:
@nzw0301 Thanks for the quick update! I understand the specification of `suggest_int(log=True)` as it is currently implemented, based on the implementation of Automl. Besides, the specification of `suggest_int(log=True)` remains questionable (especially about the set of sampled candidate points is large). However, it's beyond the scope of this PR because Automl implements it this way, and I think any change to the internal logic would also require benchmark comparison experiments. Therefore, I would like to move towards a merge with the current implementation.

optuna/trial/_trial.py
```diff
@@ -384,14 +388,23 @@ def objective(trial):
...
             A suggested integer value.
         """
 
-        distribution = distributions.IntUniformDistribution(low=low, high=high, step=step)
+        if log:
+            distribution = IntLogUniformDistribution(
+                low=low, high=high, step=step
+            )  # type: Union[IntUniformDistribution, IntLogUniformDistribution]
+        else:
+            distribution = IntUniformDistribution(low=low, high=high, step=step)
```
HideakiImamura, line 407: I think we should also add type hinting here. 
```suggestion
            distribution = IntUniformDistribution(
                low=low, high=high, step=step
            ) # type: Union[IntUniformDistribution, IntLogUniformDistribution]
```

optuna/trial/_trial.py
```diff
@@ -384,14 +388,23 @@ def objective(trial):
                 Lower endpoint of the range of suggested values. ``low`` is included in the range.
             high:
                 Upper endpoint of the range of suggested values. ``high`` is included in the range.
+            log:
+                A flag to sample the value from the log domain or not.
+                If ``log`` is true, the value is sampled from the range in the log domain.
+                Otherwise, the value is sampled from the range in the linear domain.
```
HideakiImamura, line 394: As I mentioned in https://github.com/optuna/optuna/pull/1201/files#r426442884, we need more explanation about the specification of `trial.suggest_int(... log=True ...)`. Could you write the documentation in more detail? (How the domain space `[low, high]` is log converted. How the domain space is divided by `step`. How the distribution of sampled points looks like. How the set of possible sampled points looks like.) 

If it needs a lot of work, I will do it in the separate PR! :)

optuna/trial/_trial.py
nzw0301: It may not be possible.  When applying your suggestion, mypy causes the following error:

```
optuna/trial/_trial.py:407: error: Name 'distribution' already defined on line 403
```

HideakiImamura: Sorry for the confusion. I think the following would be better because it is easy to understand why it has a type of `Union[IntUniformDistribution, IntLogUniformDistribution]`.
```suggestion
        distribution = IntUniformDistribution(
            low=low, high=high, step=step
        ) # type: Union[IntUniformDistribution, IntLogUniformDistribution]
        if log:
            distribution = IntLogUniformDistribution(low=low, high=high, step=step)
```

optuna/trial/_trial.py
nzw0301: Thank you for your comment! I update the document, please see https://github.com/optuna/optuna/pull/1201/commits/73b5033bba9bbadf5b629d876da4225696f40c89#diff-ba717fa8dc7e3ae6816dd57373267dc4R393-R402 



How do you think?

optuna/trial/_trial.py
nzw0301: sorry for my misunderstanding and thank you for your clarification. Looks nice! I merge your suggestion.

optuna/trial/_trial.py
```diff
@@ -384,14 +388,26 @@ def objective(trial):
...
-        Returns:
-            A suggested integer value.
+            log:
+                A flag to sample the value from the log domain or not.
+                If ``log`` is true,
+                the range of suggested values is converted to the log domain:
+                ``[log(low-0.5), log(high+0.5)]``,
+                and sample a float value ``u`` uniformly distributed in the range.
+                Then it is reversed and transform to the original domain:
+                ``round((exp(u) - low) / step) * step + low``.
```
HideakiImamura, line 398: How about clarifying the orders of "dividing by `step`" and "converting log domain" as follows?
```suggestion
                If ``log`` is true, at first, the range of suggested values is divided into grid 
                points of width ``step``. The range of suggested values is then converted to a log 
                domain, from which a value is uniformly sampled. The uniformly sampled value is
                re-converted to the original domain and rounded to the nearest grid point that we 
                just split, and the suggested value is determined.
```

optuna/trial/_trial.py
```diff
@@ -384,14 +388,26 @@ def objective(trial):
...
+                For example,
+                if `low = 2`, `high = 8` and `step = 2`,
+                then the suggested value is in `[2, 4, 6, 8]`
```
HideakiImamura, line 401: It might be good to clarify how the range of suggested values is divided.
```suggestion
                then the range of suggested values is divided by ``step`` as `[2, 4, 6, 8]`
```

optuna/trial/_trial.py
HideakiImamura: Thanks! I put some minor suggestions below!

optuna/trial/_trial.py
nzw0301: Great! Thanks!

HideakiImamura:
@nzw0301 Oh... Ci failed because of invalid whitespaces. Could you fix them?

nzw0301:
Oops... I will fix it soon

optuna/distributions.py
toshihikoyanase: > But it may be removed for consistency to DiscreteUniformDistribution in Optuna.

I agree with you. IMO, `IntLogUniformDistribution` can also changes `high` if `(high - low) % step != 0`.
Do you work on it within this PR or create new one?

nzw0301: Hi, Thank you for your comment! I work on it within this PR.

nzw0301: Done

optuna/distributions.py
sile: Sorry for the late comment, but please let me confirm that did you decide to keep `IntLogUniformDistribution` class? (instead of adding `log` field to the existing `IntDistribution` as I suggested).

nzw0301: @sile I apologise that I completely misunderstood your comment. I will fix it.

sile: @nzw0301 Thank you for your reply. Honestly, now I'm not confident that it's definitely better to extend the existing class...
I'd like to hear opinions from other dev members. What do you think of it? > @HideakiImamura @toshihikoyanase

nzw0301: To be honest, me neither. Note that If we extend `IntUniformDistribution`, the same problem pointed at  https://github.com/optuna/optuna/pull/1201#discussion_r419170507 happens. As a result, I will make `log` and `step` keyword-only arguments or add `log ` argument after `step` argument but it is not consistent with `suggest_int` function.

optuna/trial/_base.py
```diff
@@ -54,8 +54,8 @@ def suggest_discrete_uniform(self, name, low, high, q):
         raise NotImplementedError
 
     @abc.abstractmethod
-    def suggest_int(self, name, low, high, step=1):
-        # type: (str, int, int, int) -> int
+    def suggest_int(self, name, low, high, *, log=False, step=1):
```
hvy, line 57: Please correct me if this has been discussed before but this seems like a breaking change for users that are calling `suggest_int` while specifying `step` as a positional argument. How about keeping `step` in the same position, add `log` afterwards, and
a. raise a deprecation warning that arguments orders will change, then change the order in the future, or
b. rearrange the order in `suggest_float` to match `suggest_int`?

hvy: I now see that a discussed was held here https://github.com/optuna/optuna/pull/1201#discussion_r419170507, but it still seems like an issue to me.

optuna/distributions.py
hvy: Again, sorry for joining in the middle of the conversation but I think there's room for rethinking what concrete `BaseDistributions` we should provide and what parameters they should involve, now that the suggest APIs are being changed. I'm wondering if we shouldn't do that first before discussing this particular class and leave this PR as is. C.f. https://github.com/optuna/optuna/issues/510

sile: @nzw0301 @hvy Thank you for your feedback! Now, I think that it's better to keep this PR as is, so let me resolve this conversation.

optuna/trial/_base.py
HideakiImamura: I agree with @hvy's idea (b). It might be better to change the order of arguments in `suggest_int` to `suggest_int(low, high, step, log)`. Changing the argument order of `suggest_float` can be a future work (another PR).

nzw0301: Thank you for your suggestion. Personally, option b seems better because both arguments are already keyword-only arguments in `suggest_float`. So its order can be changed with keeping the backward compatibility,

tests/test_distributions.py
```diff
@@ -150,6 +157,26 @@ def test_contains():
...
+    ilu = distributions.IntUniformDistribution(low=2, high=12)
+    assert not ilu._contains(0.9)
+    assert ilu._contains(2)
+    assert ilu._contains(4)
+    assert ilu._contains(6)
+    assert ilu._contains(12)
+    assert not ilu._contains(12.1)
+    assert not ilu._contains(13)
+
+    # IntLogUniformDistribution with a 'q' parameter.
```
sile, line 169: Maybe a typo:
```suggestion
    # IntLogUniformDistribution with a 'step' parameter.
```

optuna/samplers/random.py
```diff
@@ -84,15 +84,23 @@ def sample_independent(self, study, trial, param_name, param_distribution):
...
             return int(v)
         elif isinstance(param_distribution, distributions.IntLogUniformDistribution):
+            # adjust highest value
+            high = int(
+                numpy.floor(
+                    (param_distribution.high - param_distribution.low) / param_distribution.step
+                )
+                * param_distribution.step
+                + param_distribution.low
+            )
```
toshihikoyanase, line 94: (This is a reply to https://github.com/optuna/optuna/pull/1201#discussion_r427713669.)

Thank you for your swift action!

I have two additional comments about this logic.

1. Instead of adjusting the high value in the sampler, how about changing it in the `Trial.suggest_int` similarly to the `Trial.suggest_discrete_uniform`? It calls `_adjust_discrete_uniform_high` in `optuna/trial/_util.py`, and I think we can employ this approach.

2. I think implement this logic without `numpy` as follows:

`(param_distribution.high - param_distribution.low) // param_distribution.step * param_distribution.step + param_distribution.low`

Does it have problems?

tests/test_distributions.py
nzw0301: Thanks!  Indeed, it is copied from https://github.com/optuna/optuna/blob/2fd7fc693d12ec0ed93af5dce760da2a5671d6dd/tests/test_distributions.py#L136 Should I fix both?

optuna/samplers/random.py
nzw0301: Thank you for your comments!

**Comment. 1**: Great! It should be done in a similar way to`Trial.suggest_discrete_uniform`.

**Comment. 2**: I think they output different results. 

```python
import numpy

low = 2
high = 10
step = 2
samples = []
log_low = numpy.log(low - 0.5)
log_high = numpy.log(high + 0.5)

# current logic
rnd = numpy.random.RandomState(8)
for _ in range(1000):
    s = numpy.exp(rnd.uniform(log_low, log_high))
    v = (
        numpy.round((s - low) / step)
        * step
        + low
    )
    samples.append(int(min(max(v, low), high)))

# print(numpy.bincount(samples))
# array([  0,   0, 345,   0, 275,   0, 183,   0, 116,   0,  81])

# new logic without numpy
rnd = numpy.random.RandomState(8)
for _ in range(1000):
    s = numpy.exp(rnd.uniform(log_low, log_high))
    v = (
        (s - low) // step
        * step
        + low
    )
    samples.append(int(min(max(v, low), high)))

# print(numpy.bincount(samples))
# array([  0,   0, 842,   0, 502,   0, 321,   0, 224,   0, 111])
```

optuna/trial/_base.py
hvy: Thanks for the swift discussion. Sounds good. Let's deal with `suggest_float` after this PR (it'd mostly be a cosmetic fix).

tests/test_distributions.py
sile: Oh, it's great to fix the existing comment too if you can.

optuna/samplers/random.py
nzw0301: Again,  I misunderstood your comment. I'm sorry if my previous comment confused you.

toshihikoyanase: Thank you for your update. The commit https://github.com/optuna/optuna/pull/1201/commits/ad5d3faf02da2a0fed6aac25fc2bb3b8a79bdbf2 seems good to me.

nzw0301:
@toshihikoyanase Thank you for summarising the follow-up PRs. Looks great! 
I think implementation of log argument for `optuna.integration.ChainerMNTrial.suggest_int` is also necessary for future development.

nzw0301:
I would like to thank you @toshihikoyanase @himkt @c-bata @crcrpar @HideakiImamura @sile @hvy for reviewing this PR!

toshihikoyanase:
> I think implementation of log argument for optuna.integration.ChainerMNTrial.suggest_int is also necessary for future development.

@nzw0301 You're right. Thank you for pointing it out!

PR #1204, opened 2020-05-04
Import `optuna.exceptions.TrialPruned` in `__init__.py`.

keisuke-umezawa:
## Motivation

Although `optuna.exceptions.TrialPruned` is exception, but it is used by users to prune trials. So, it is better that users can use it by `optuna.TrialPruned`.

## Description of the changes

- Import TrialPruned in `__init__`.
- Use `optuna.TrialPruned` in unittests and integration.

## Discussion topics
1. Do we have any ideas to implement it in a better way?
2. Is it okay to use `exception` as normal control flow?

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -190,7 +190,7 @@ How are exceptions from trials handled?
 
 Trials that raise exceptions without catching them will be treated as failures, i.e. with the :obj:`~optuna.trial.TrialState.FAIL` status.
 
-By default, all exceptions except :class:`~optuna.exceptions.TrialPruned` raised in objective functions are propagated to the caller of :func:`~optuna.study.Study.optimize`.
+By default, all exceptions except :class:`~optuna.TrialPruned` raised in objective functions are propagated to the caller of :func:`~optuna.study.Study.optimize`.
 In other words, studies are aborted when such exceptions are raised.
 It might be desirable to continue a study with the remaining trials.
 To do so, you can specify in :func:`~optuna.study.Study.optimize` which exception types to catch using the ``catch`` argument.
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -190,7 +190,7 @@ How are exceptions from trials handled?
 
 Trials that raise exceptions without catching them will be treated as failures, i.e. with the :obj:`~optuna.trial.TrialState.FAIL` status.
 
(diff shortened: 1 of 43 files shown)


hvy:
Should we also update the Hyperband related following lines?
- https://github.com/optuna/optuna/blob/master/tests/pruners_tests/test_hyperband.py#L119
- https://github.com/optuna/optuna/blob/master/optuna/pruners/hyperband.py#L96

hvy:
I believe the discussions points are completely valid am wondering if we for instance should not return constant-like objects or enums instead of raising. However, I think it's fine to proceed with this PR and discuss these points outside.

keisuke-umezawa:
@hvy 
Thank you for reviewing. I fixed following.

> Should we also update the Hyperband related following lines?
> 
> https://github.com/optuna/optuna/blob/master/tests/pruners_tests/test_hyperband.py#L119
> https://github.com/optuna/optuna/blob/master/optuna/pruners/hyperband.py#L96

hvy:
Thanks for your fixes. Apologies for my late comments but I have some follow up questions. Since we don't have API references linking to `optuna.TrialPruned` (or other symbols directly under `optuna`) how about appending to the reference of `optuna.exceptions.TrialPruned` that it's also available as `optuna.TrialPruned` (just hit me that we have little to no references to `optuna.exceptions.TrialPruned` in the first place; not from the pruner page, nor from the pruning tutorial)? I'm also wondering whether we want to recommend using any (just an alias), or either (might deprecate the other) of them. Do you have any comments on this?

keisuke-umezawa:
@hvy 
> how about appending to the reference of optuna.exceptions.TrialPruned that it's also available as optuna.TrialPruned

Here, I just followed `Study` and `Trial`, which do not mention about alias. But I can add the information for `TrialPruned`.

> I'm also wondering whether we want to recommend using any (just an alias), or either (might deprecate the other) of them.

I just understood this issue as "users want to use TrialPruned by following code".

```
import optuna

...
raise optuna.TrialPruned()
```

But, if we want to stop users from using `optuna.exceptions.TrialPruned`, we need to choose another implementation.

hvy:
Note: The documentation is addressed here https://github.com/optuna/optuna/pull/1278.

toshihikoyanase:
Thank you for your PR. It basically looks good to me.
Could you update ChainerMN integration as well?

https://github.com/optuna/optuna/blob/master/optuna/integration/chainermn.py#L4

PR #1208, opened 2020-05-06
Raise `ValueError` that is not raised

harupy:
## Motivation

Just instantiating `ValueError` does nothing. It needs to be raised.

## Description of the changes

- Raise `ValueError`.
- Fix the test which currently catches a different `ValueError`.

### DIFF
--- a/tests/visualization_tests/test_parallel_coordinate.py
+++ b/tests/visualization_tests/test_parallel_coordinate.py
@@ -44,7 +44,7 @@ def test_plot_parallel_coordinate():
     assert figure.data[0]["dimensions"][1]["values"] == (1.0, 2.5)
 
     # Test with wrong params that do not exist in trials
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="Parameter optuna does not exist in your study."):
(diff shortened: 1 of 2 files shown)


tests/visualization_tests/test_parallel_coordinate.py
```diff
@@ -44,7 +44,7 @@ def test_plot_parallel_coordinate():
     assert figure.data[0]["dimensions"][1]["values"] == (1.0, 2.5)
 
     # Test with wrong params that do not exist in trials
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="Parameter optuna does not exist in your study."):
         plot_parallel_coordinate(study, params=["optuna", "optuna"])
```
harupy, line 48: Before the fix, `plot_parallel_coordinate(study, params=["optuna", "optuna"])` raised the following error.

```
File "/.../optuna/optuna/visualization/parallel_coordinate.py", line 105, in _get_parallel_coordinate_plot
    "range": (min(values), max(values)),
ValueError: min() arg is an empty sequence
```

PR #1210, opened 2020-05-06
Dissect `trial.py`.

### LINKED ISSUE #1072: Separate `trial.py` into `trials/xxx_trial.py`.
The line number of `trial.py` becomes large. (1071L)
How about creating a directory `trials` and separating `trial.py` into `trials/base_trial.py`, `trials/trial.py` and `trials/fixed_trial`? (file names are just examples.)

himkt:
Close #1072.

## Motivation
This PR separates `trial.py` into `trials/*.py` (since its codebase is becoming huge) such that it keeps the API unchanged.

## Description of the changes
- created `optuna/trials`
- changed `optuna/trial.py`: now it only imports modules in `optuna/trials`
- introduced `util.py`: contain `_adjust_discrete_uniform_high`. I think it needs a discussion about naming and placing

## Concerns
- `util.py`: naming for the module and placing
- script structure

## Another topic
In the source issue, we also discussed about changing API.
It's reasonable for me to make another PR about rethinking API since this PR becomes relatively huge.

## Memo

- 3e3efa7


himkt:
I think there is no discussion yet, I'd appreciate it if developers give me some comments.
@hvy Could you please give me some advice?

hvy:
I sort of started one here https://github.com/optuna/optuna/issues/1072#issuecomment-608195720. To add, introducing `trials` is a quite big change and if we are to do so, we should probably at least keep aliases for backward compatibility. However, in practice, I image that many end-users actually won't be affected as they don't have to import `optuna.trial`. For users extending Optuna, there's an entire different story and hence the backward compatibility.

HideakiImamura:
Thank you for your proposal! I totally agree with your idea to diet `optuna.trial.py`. But, I worry about extending namespaces `optuna.trials`. In my personal opinion, we should not extend so and let developers just use all trial classes as `optuna.trial.Trial`. I believe it is developer-friendly.

himkt:
> To add, introducing trials is a quite big change and if we are to do so, we should probably at least keep aliases for backward compatibility

I thought the concern is only about backward compatibility and added a trick to keep them [here](https://github.com/optuna/optuna/pull/1210/files#diff-815bd38671f30e415efa796835df638fR1-R5), which means `optuna.trial` is available, I think.

However, the availability of both `optuna.trial` and `optuna.trials` is confusing and
I agree with @HideakiImamura that we shouldn't extend namespace now.

I'll make the change to keep backward compatibility without extending namespace.
I want to discuss about rethinking API in another issue or PR.
Thanks, @hvy @HideakiImamura!

PR #1216, opened 2020-05-08
Do not label `contribution-welcome` and `bug` issues as stale.

hvy:
## Motivation

Issues labeled with `contribution-welcome`, as with any other issue or PR, are currently labeled stale if no updates occur for a certain period of time. Usually, these issues are still valid after this period.

## Description of the changes

Let the stale bot exclude issues labeled with `contribution-welcome`. See https://github.com/actions/stale/blob/v2.0.0/action.yml.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -20,3 +20,4 @@ jobs:
         days-before-close: -1
         stale-issue-label: 'stale'
         stale-pr-label: 'stale'
+        exempt-issue-labels: 'contribution-welcome'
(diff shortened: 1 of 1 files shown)


hvy:
Sorry @smly for my late change but I added the `bug` label as well for similar reason.

hvy:
PTAL.

hvy:
This issue might be open for discussion since it depends on how we want to address stale tickets.

smly:
@hvy, thanks for the change. LGTM :+1:

PR #1218, opened 2020-05-08
Fix the metric name in `KerasPruningCallback` example.

toshihikoyanase:
## Motivation

`examples/pruning/keras_integration.py` does not prune any trials. This is because the metric name has been changed from `val_acc` to `val_accuracy`.

## Changes

This PR simply updates the metric name, but it may be better to show warning messages if the metric cannot be found in the callback. If you have any comments, please let me know.

### DIFF
--- a/examples/pruning/keras_integration.py
+++ b/examples/pruning/keras_integration.py
@@ -74,7 +74,7 @@ def objective(trial):
         x_train,
         y_train,
         batch_size=BATCHSIZE,
-        callbacks=[KerasPruningCallback(trial, "val_acc")],
+        callbacks=[KerasPruningCallback(trial, "val_accuracy")],
         epochs=EPOCHS,
         validation_data=(x_valid, y_valid),
         verbose=1,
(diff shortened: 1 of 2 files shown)


keisuke-umezawa:
FYI
It seems that we can run the example if we write `metrics=["acc"],` here.

https://github.com/optuna/optuna/blob/b5f6a262084f42a01054d80bf91dbf76e60c9258/examples/pruning/keras_integration.py#L49

keisuke-umezawa:
Sorry, just a nit.
If possible, could you update this part from ``val_acc`` to ``val_accuracy``?

https://github.com/optuna/optuna/blob/master/optuna/integration/keras.py#L32

toshihikoyanase:
> If possible, could you update this part from val_acc to val_accuracy?
> https://github.com/optuna/optuna/blob/master/optuna/integration/keras.py#L32

@keisuke-umezawa Good catch! I fixed it in https://github.com/optuna/optuna/pull/1218/commits/ddb4cc3bf224489858f183530bf1002078277bc3. Thank you.

PR #1223, opened 2020-05-08
Fix cma's high  for categorical dist.

nzw0301:
## Motivation

The initialisation of `_Optimizer` for `DiscreteUniformDistribution` in `optuna/integration/cma.py` seems wrong. But I'm not sure.

## Description of the changes

Replace it with the correct value.


nzw0301:
I'm wrong. Sorry...

PR #1150, opened 2020-04-21
[AllenNLP] Implement config exporter to save training config with best_params in study.

himkt:
Depends on #1149.

This PR introduces `optuna.integration.allennlp.save_best_config`, which enables users to generate a jsonnet config file filled masked values with optimized ones in the `study.optimize()`.

A generated jsonnet file is executable on `allennlp train`.
I put a typical use case below:

```bash
# optimize hyperparameters using optuna
python optimize_hyperparams.py --input config.jsonnet --output optimized.config.jsonnet

# train a model using hyperparameters optimized in optimized_hyperparameters.py
allennlp train -s xxx optimized.config.jsonnet
```

PR #1225, opened 2020-05-10
[AllenNLP] Use `dump_best_config` in example.

himkt:
## Motivation
Show the usage of `dump_best_config` introduced in #1150.

## Description of the changes
`dump_best_config` was introduced in #1150 to dump the configuration file
which had achieved the best performance in the study.
This PR updates `allennlp_jsonnet.py` to use this feature.

### DIFF
--- a/examples/allennlp/allennlp_jsonnet.py
+++ b/examples/allennlp/allennlp_jsonnet.py
@@ -33,7 +33,7 @@ from optuna.integration import AllenNLPExecutor
 EXAMPLE_DIR = os.path.dirname(os.path.abspath(__file__))
 CONFIG_PATH = os.path.join(EXAMPLE_DIR, "classifier.jsonnet")
 MODEL_DIR = "result"
-BEST_CONFIG_PATH = "best.classifier.json"
+BEST_CONFIG_PATH = "best_classifier.json"
 
 
 def objective(trial):
--- a/examples/allennlp/allennlp_jsonnet.py
+++ b/examples/allennlp/allennlp_jsonnet.py
@@ -20,7 +20,6 @@ We have the following two ways to execute this example:
 
 """
 
-import json
 import os.path
 import shutil
 
@@ -34,7 +33,7 @@ from optuna.integration import AllenNLPExecutor
 EXAMPLE_DIR = os.path.dirname(os.path.abspath(__file__))
 CONFIG_PATH = os.path.join(EXAMPLE_DIR, "classifier.jsonnet")
 MODEL_DIR = "result"
-BEST_CONFIG_PATH = os.path.join(MODEL_DIR, "best.classifier.json")
+BEST_CONFIG_PATH = "best.classifier.json"
 
 
 def objective(trial):
@@ -64,10 +63,6 @@ if __name__ == "__main__":
         print("    {}: {}".format(key, value))
 
     dump_best_config(CONFIG_PATH, BEST_CONFIG_PATH, study)
-
-    print("Best AllenNLP config:")
-    best_config = json.load(open(BEST_CONFIG_PATH))
-    for key, value in best_config.items():
-        print("    {}: {}".format(key, value))
+    print("\nCreated optimized AllenNLP config to `{}`".format(BEST_CONFIG_PATH))
 
     shutil.rmtree(MODEL_DIR)
--- a/examples/allennlp/allennlp_jsonnet.py
+++ b/examples/allennlp/allennlp_jsonnet.py
@@ -20,17 +20,21 @@ We have the following two ways to execute this example:
 
 """
 
+import json
 import os.path
 import shutil
 
 import optuna
+from optuna.integration.allennlp import dump_best_config
 from optuna.integration import AllenNLPExecutor
 
 
 # This path trick is used since this example is also
 # run from the root of this repository by CI.
 EXAMPLE_DIR = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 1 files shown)


examples/allennlp/allennlp_jsonnet.py
```diff
@@ -20,17 +20,21 @@
...
+from optuna.integration.allennlp import dump_best_config
 from optuna.integration import AllenNLPExecutor
 
 
 # This path trick is used since this example is also
 # run from the root of this repository by CI.
 EXAMPLE_DIR = os.path.dirname(os.path.abspath(__file__))
+CONFIG_PATH = os.path.join(EXAMPLE_DIR, "classifier.jsonnet")
 MODEL_DIR = "result"
+BEST_CONFIG_PATH = os.path.join(MODEL_DIR, "best.classifier.json")
```
toshihikoyanase, line 37: This example shows the content of `best.classifier.json` in the stdout because the `best.classifier.json` will be removed with its parent directory. But, I feel it would be easy to understand the usage of `dump_best_config` if it keeps the file and shows the file path.
How about placing it in the current directory?

himkt: I agree with you, I revised PR to create `best.classifier.json` on current directory in 2400ca3.

```
%python allennlp_json.py
....
....
[I 2020-05-14 23:57:53,253] Finished trial#2 with value: 0.56328 with parameters: {'DROPOUT': 0.38850725212419135, 'EMBEDDING_DIM': 46, 'MAX_FILTER_SIZE': 3, 'NUM_FILTERS': 32, 'HIDDEN_SIZE': 21}. Best is trial#2 with valu
e: 0.56328.
Number of finished trials:  3
Best trial:
  Value:  0.56328
  Params:
    DROPOUT: 0.38850725212419135
    EMBEDDING_DIM: 46
    MAX_FILTER_SIZE: 3
    NUM_FILTERS: 32
    HIDDEN_SIZE: 21

Created optimized AllenNLP config to `best.classifier.json`
```

examples/allennlp/allennlp_jsonnet.py
```diff
@@ -59,4 +62,7 @@ def objective(trial):
     for key, value in trial.params.items():
         print("    {}: {}".format(key, value))
 
+    dump_best_config(CONFIG_PATH, BEST_CONFIG_PATH, study)
+    print("\nCreated optimized AllenNLP config to `{}`".format(BEST_CONFIG_PATH))
```
toshihikoyanase, line 66: [nit] Please add a period at the end of the sentence.
```suggestion
    print("\nCreated optimized AllenNLP config to `{}`.".format(BEST_CONFIG_PATH))
```

himkt: Thanks for the catch. :bow:

examples/allennlp/allennlp_jsonnet.py
```diff
@@ -24,13 +24,16 @@
...
+from optuna.integration.allennlp import dump_best_config
 from optuna.integration import AllenNLPExecutor
 
 
 # This path trick is used since this example is also
 # run from the root of this repository by CI.
 EXAMPLE_DIR = os.path.dirname(os.path.abspath(__file__))
+CONFIG_PATH = os.path.join(EXAMPLE_DIR, "classifier.jsonnet")
 MODEL_DIR = "result"
+BEST_CONFIG_PATH = "best.classifier.json"
```
hvy, line 36: Oh the output doesn't have to be a jsonnet (c.f. https://github.com/optuna/optuna/blob/master/tests/integration_tests/allennlp_tests/test_allennlp.py#L127)?

hvy, line 36: I think it's a better practice not to include punctuations in the file besides for separating between file extensions.

himkt: Good question. 😅

In fact, all the variables defined in the jsonnet file are rewritten with the values found in the trial and embedded in the body, so the output file is valid as both jsonnet and json.
Do you thinks I should update [jsonnet](https://github.com/optuna/optuna/blob/master/tests/integration_tests/allennlp_tests/test_allennlp.py#L127) to be `result.json`?

himkt: Thank you for pointing out. I fixed it in 5ab58d9.

hvy: So the input and output file types are different? I'm not familiar with jsonnet but should we document that in [`dump_best_config`](https://github.com/optuna/optuna/blob/master/optuna/integration/allennlp.py#L24)? I'm fine with either if they are aligned.

himkt: You're right. `dump_best_config` accepts both Json and Jsonnet files and dumps `config.json`, which is consistent with `allennlp train`. ([train](https://github.com/allenai/allennlp/blob/7d71398b44636c52e811467a684c3099c64009ec/allennlp/commands/train.py#L175-L238), [dump](https://github.com/allenai/allennlp/blob/7d71398b44636c52e811467a684c3099c64009ec/allennlp/models/archival.py#L84-L127))

I'll make another PR to modify the test and put a note that `dump_best_config` create a JSON file. I'm sorry for confusing you.

hvy: Sounds good. And thanks for the explanation, it helps since I'm not familiar with how AllenNLP works.

PR #1191, opened 2020-04-30
Increase test coverage of storage tests for single worker cases.

ytsmiling:
## Motivation
Many storage features/specs are not tested. This low test coverage is a direct cause of the inconsistent implementations among storages. This PR aims to increase the test coverage and, at the same time, remove found bugs/inconsistent behaviors. This PR inevitably relies on #1175 and #1176. While we should test specs related to multi-worker settings, it will make this PR too complicated and it will be separated into another PR.

## Description of the changes
- This PR increases test coverage by adding more cases in tests/test_storage.py. Additionally, this PR resolves many inconsistent behaviors and bugs in the current implementation.
- This PR slows down the test, which should be resolved by another PR by removing duplicated/redundant methods in the storage class.
- This PR includes multi-study support from the in-memory storage.
- This PR changes the interface of testing/StorageSuppliers and related tests. Importantly, this PR removes tests using "common sqlite's db file." Tests related to behaviors in multi-worker settings should be added in other ways by other PRs.

## The following space will be used to list all fixed/modified behaviors of storage implementations.

- [x] Study_id duplication in In-memory DB.
- [x] Study-name duplication is not handled in in-memory DB.
- [x] In-meomry DB raises ValueError when deleting non-existent study (expected: KeyError).
- [x] RDB backend raises ValueError when deleting non-existent study (expected: KeyError).
- [x] Redis backend raises ValueError when deleting non-existent study (expected: KeyError).
- [x] In-memory DB raises ValueError when study_id does not exist in get_study_name_from_id (expected: KeyError).
- [x] RDB backend raises ValueError when study_id does not exist in get_study_name_from_id (expected: KeyError).
- [x] Redis backend raises ValueError when study_id does not exist in get_study_name_from_id (expected: KeyError).
- [x] In-memory DB raises ValueError when study_id does not exist in get_study_direction (expected: KeyError).
- [x] RDB backend raises ValueError when study_id does not exist in get_study_direction (expected: KeyError).
- [x] Redis backend raises AssertionError when study_id does not exist in get_study_direction (expected: KeyError).
- [x] In-memory DB does not raise when study_id does not exist in get_study_user_attrs (expected: KeyError).
- [x] RDB backend raises ValueError when study_id does not exist in get_study_user_attrs (expected: KeyError).
- [x] Redis backend raises AssertionError when study_id does not exist in get_study_user_attrs (expected: KeyError).
- [x] In-memory DB does not raise when study_id does not exist in get_study_system_attrs (expected: KeyError).
[...]

PR #1228, opened 2020-05-11
Support multiple studies in `InMemoryStorage`.

ytsmiling:
## Motivation
This PR adds multi-study support to in-memory storage.
This partially addresses #1170 and relates to #1191.

## Description of the changes
Add multi-study support to in-memory db.

## Other notes
I checked that this passes tests in #1191 in python3.5.

Update:
- In this PR, in-memory storage raises ValueError instead of KeyError to pass deprecated tests, which will be fixed in #1191.
- This PR removes some deprecated tests which will be completely removed or replaced by #1191.


tests/storages_tests/test_storages.py
```diff
@@ -426,10 +426,7 @@ def test_set_and_get_trial_param(storage_init_func):
 
     # Setup trial_3: setting new params (to different study from trial_1).
     if isinstance(storage, InMemoryStorage):
-        with pytest.raises(ValueError):
-            # InMemoryStorage shares the same study if create_new_study is additionally invoked.
-            # Thus, the following line should fail due to distribution incompatibility.
-            storage.set_trial_param(trial_id_3, "y", 1, distribution_y_2)
+        pass
```
ytsmiling, line 429: These "tests" will be completely removed by #1191.

tests/integration_tests/test_chainermn.py
```diff
@@ -151,10 +151,7 @@ def test_init_with_multiple_study_names(storage_mode, comm):
     def test_init_with_incompatible_storage(comm):
         # type: (CommunicatorBase) -> None
 
-        study = TestChainerMNStudy._create_shared_study(InMemoryStorage(), comm)
-
-        with pytest.raises(ValueError):
-            ChainerMNStudy(study, comm)
+        pass
```
ytsmiling, line 154: Ditto.

sile:
@ytsmiling Thank you for your PR. I'll review this PR today.
By the way, could you run a benchmark to measure the execution speed of the new implementation?

ytsmiling:
Result of the same benchmark with #1135:
Current master:
       214310942 function calls (192531182 primitive calls) in 130.340 seconds
This PR:
         81599936 function calls (81348176 primitive calls) in 82.710 seconds

So, rewriting the in-memory db codes resulted in over 30% speedup.

sile:
Oh, great! Honestly, I was a bit concern that this PR might introduce some overhead.

ytsmiling:
I guess that the speed up comes from rewriting the `get_trials` method and other methods have increased latency.
But anyway, the overhead seems not that problem.

optuna/storages/in_memory.py
```diff
@@ -205,165 +230,213 @@ def set_trial_state(self, trial_id, state):
...
+        self._studies[study_id].trials[trial_number] = trial
 
     def get_all_trials(self, study_id, deepcopy=True):
         # type: (int, bool) -> List[FrozenTrial]
 
         self._check_study_id(study_id)
         with self._lock:
             if deepcopy:
-                return copy.deepcopy(self.trials)
+                return [copy.deepcopy(trial) for trial in self._studies[study_id].trials]
```
sile, line 405: I'm just curious: are there any reasons you prefer the current code to `copy.deepcopy(self._studies[study_id].trials)`?

optuna/storages/in_memory.py
```diff
@@ -53,108 +51,131 @@ def __setstate__(self, state):
...
+            self._studies[study_id] = _StudyInfo(study_name)
+            self._study_name_to_id[study_name] = study_id
 
-        return IN_MEMORY_STORAGE_STUDY_ID  # TODO(akiba)
+        return study_id
 
     def delete_study(self, study_id):
         # type: (int) -> None
 
         self._check_study_id(study_id)
```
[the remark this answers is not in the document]
ytsmiling: Yes, indeed, we should add locking. (Ideally, we should remove the multi-thread support in the future. It yields no speed gain in general use cases, and it complicates codes here and there and increases overheads in many ways.)

optuna/storages/in_memory.py
```diff
@@ -167,11 +188,12 @@ def create_new_trial(self, study_id, template_trial=None):
             trial = copy.deepcopy(template_trial)
 
         with self._lock:
-            trial_id = len(self.trials)
-            trial.number = trial_id
+            trial_id = len(self._trial_id_to_study_id_and_number)
```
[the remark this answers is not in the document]
ytsmiling: Yes, we definitely should replace the code. Thanks for the catch.

optuna/storages/in_memory.py
ytsmiling: Nothing. I remember that I intentionally used the verbose style because the original code did it, but it seems it's just my misunderstanding.

optuna/storages/in_memory.py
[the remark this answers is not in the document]
ytsmiling: #1232

optuna/storages/in_memory.py
```diff
@@ -205,165 +230,213 @@ def set_trial_state(self, trial_id, state):
...
 
         self._check_study_id(study_id)
         if state is None:
-            return len(self.trials)
+            return len(self._studies[study_id].trials)
 
-        return len([t for t in self.trials if t.state == state])
+        return len(
+            [0 for trial in self.get_all_trials(study_id, deepcopy=False) if trial.state == state]
+        )
```
hvy: How about using `sum`?
```python
sum(trial.state == state for trial in self.get_all_trials(study_id, deepcopy=False))
```

ytsmiling: Thank you. I adopted the change.

optuna/storages/in_memory.py
```diff
@@ -53,108 +51,131 @@ def __setstate__(self, state):
...
-            self.study_name = study_name
+            if study_name in self._study_name_to_id:
+                raise DuplicatedStudyError
+            self._studies[study_id] = _StudyInfo(study_name)
+            self._study_name_to_id[study_name] = study_id
+        else:
+            study_uuid = str(uuid.uuid4())
+            study_name = DEFAULT_STUDY_NAME_PREFIX + study_uuid
+            self._studies[study_id] = _StudyInfo(study_name)
+            self._study_name_to_id[study_name] = study_id
```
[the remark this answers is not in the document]
ytsmiling: I removed the duplicated code fragment. Thank you for pointing it out.

optuna/storages/in_memory.py
```diff
@@ -53,108 +51,131 @@ def __setstate__(self, state):
...
+        self._check_study_id(study_id)
+        return self._studies[study_id].direction
 
     def get_study_user_attrs(self, study_id):
         # type: (int) -> Dict[str, Any]
 
+        self._check_study_id(study_id)
         with self._lock:
-            return copy.deepcopy(self.study_user_attrs)
+            return self._studies[study_id].user_attrs
```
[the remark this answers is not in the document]
ytsmiling: Yes, it can be addressed into another PR.
However, the current implementation of this PR relies on this copying behavior of get_trial, and modifying the line requires a large change of codes.

I'll create another PR to change the copying behavior and I'd like to merge it before this PR.

optuna/storages/in_memory.py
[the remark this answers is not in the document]
ytsmiling: I added the guard.

optuna/storages/in_memory.py
```diff
@@ -53,108 +51,131 @@ def __setstate__(self, state):
...
+                and self._studies[study_id].direction != direction
+            ):
                 raise ValueError(
                     "Cannot overwrite study direction from {} to {}.".format(
-                        self.direction, direction
+                        self._studies[study_id].direction, direction
                     )
                 )
-            self.direction = direction
+            self._studies[study_id].direction = direction
```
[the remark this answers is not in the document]
ytsmiling: I adopted the change. Thank you.

ytsmiling:
https://github.com/optuna/optuna/pull/1228#discussion_r423461923
This will be a reason for the speed gain by this PR.

optuna/storages/in_memory.py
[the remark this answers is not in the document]
ytsmiling: After examining the code, I decided not to create separate PR and keep the original copying behavior in this PR.

ytsmiling:
After reverting the deep-copy behavior in https://github.com/optuna/optuna/pull/1228#discussion_r423461923, the benchmark became as follows:
```
         209306939 function calls (188385179 primitive calls) in 129.212 seconds
```
This is almost the same as the current master implementation.

optuna/storages/in_memory.py
```diff
@@ -196,174 +229,241 @@ def set_trial_state(self, trial_id, state):
...
     def _check_study_id(self, study_id):
         # type: (int) -> None
 
-        if study_id != IN_MEMORY_STORAGE_STUDY_ID:
-            raise ValueError(
-                "study_id is supposed to be {} in {}.".format(
-                    IN_MEMORY_STORAGE_STUDY_ID, self.__class__.__name__
-                )
-            )
+        with self._lock:
```
hvy, line 450: Can we skip this lock, assuming that it's acquired by the caller? (Maybe we could do that with `_check_trial_id` as well.)

optuna/storages/in_memory.py
```diff
@@ -53,126 +52,160 @@ def __setstate__(self, state):
...
-            self.trials = []
-            self.best_trial_id = None
-            self.param_distribution = {}
-            self.direction = StudyDirection.NOT_SET
-            self.study_user_attrs = {}
-            self.study_system_attrs = {}
-            self.study_name = DEFAULT_STUDY_NAME_PREFIX + IN_MEMORY_STORAGE_STUDY_UUID
+            self._check_study_id(study_id)
+
+            with self._lock:
```
hvy, line 76: We should not nest a lock acquisition here. There are a couple of similar occurrences further down.

optuna/storages/in_memory.py
```diff
@@ -196,174 +229,241 @@ def set_trial_state(self, trial_id, state):
...
-        if self.best_trial_id is None:
-            raise ValueError("No trials are completed yet.")
-        return self.get_trial(self.best_trial_id)
+        with self._lock:
+            self._check_study_id(study_id)
+
+            best_trial_id = self._studies[study_id].best_trial_id
+            if best_trial_id is None:
+                raise ValueError("No trials are completed yet.")
+            return self._get_trial(best_trial_id)
```
hvy, line 303: Should this be a deep copy if we are to keep the same behavior?

optuna/storages/in_memory.py
ytsmiling: Yes, we should not acquire locks in private methods. I'll fix this.

optuna/storages/in_memory.py
ytsmiling: Yes, this should be self.get_trial(best_trial_id). I'll fix this.

optuna/storages/in_memory.py
ytsmiling: Exactly. I wonder why these did not break the test. (These methods might not be covered in old tests.)

optuna/storages/in_memory.py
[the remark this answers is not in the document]
sile: Thank you for creating the issue. I also want to remove the multi-thread support in the future.

PR #1234, opened 2020-05-12
Add interval to PercentilePruner.

VladSkripniuk:
Address #1136 .

## Description of the changes

1. Add `is_target_step` to `BasePruner` and implement `is_target_step` in `PercentilePruner`.
2. Use `pruner.is_target_step(epoch)` in `KerasPruningCallback` to skip pruning by interval.

### DIFF
--- a/optuna/pruners/percentile.py
+++ b/optuna/pruners/percentile.py
@@ -192,3 +192,10 @@ class PercentilePruner(BasePruner):
         if direction == StudyDirection.MAXIMIZE:
             return best_intermediate_result < p
         return best_intermediate_result > p
+
+    def is_target_step(self, step):
+        # type: (int) -> bool
+        if step < self._n_warmup_steps:
+            return False
+
+        return (step - self._n_warmup_steps) % self._interval_steps == 0
--- a/optuna/integration/keras.py
+++ b/optuna/integration/keras.py
@@ -31,14 +31,10 @@ class KerasPruningCallback(Callback):
             An evaluation metric for pruning, e.g., ``val_loss`` and
             ``val_accuracy``. Please refer to `keras.Callback reference
             <https://keras.io/callbacks/#callback>`_ for further details.
-        interval:
-            Check if trial should be pruned every n-th epoch. By default ``interval=1`` and
-            pruning is performed after every epoch. Increase ``interval`` to run several
-            epochs faster before applying pruning.
-     """
+    """
 
-    def __init__(self, trial, monitor, interval=1):
-        # type: (optuna.trial.Trial, str, int) -> None
+    def __init__(self, trial, monitor):
+        # type: (optuna.trial.Trial, str) -> None
 
         super(KerasPruningCallback, self).__init__()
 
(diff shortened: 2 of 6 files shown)


optuna/pruners/percentile.py
```diff
@@ -192,3 +192,10 @@ def prune(self, study, trial):
         if direction == StudyDirection.MAXIMIZE:
             return best_intermediate_result < p
         return best_intermediate_result > p
+
+    def is_target_step(self, step):
+        # type: (int) -> bool
+        if step < self._n_warmup_steps:
+            return False
+
+        return (step - self._n_warmup_steps) % self._interval_steps == 0
```
toshihikoyanase, line 201: Please add a test case to confirm if the target steps determined by this function matches with those determined by the `prune` method? I think the test cases can be written using the [`DeteministicPruner(True)`](https://github.com/optuna/optuna/blob/master/optuna/testing/integration.py#L4-L13).

IMO, the current logic does not match with `prune` in some cases. For example, the check corresponding to L198 seems to be `if step <= self._nwarmup_steps` in the `prune` method.

toshihikoyanase:
@keisuke-umezawa Could you give us your feedback about https://github.com/optuna/optuna/pull/1234#pullrequestreview-411385804? I think it is controversial.

> I think we need to ensure that the target steps determined by `is_target_step` are actual targets of the `prune` method.

keisuke-umezawa:
@toshihikoyanase @VladSkripniuk 

> I think we need to ensure that the target steps determined by is_target_step are actual targets of the prune method.

I agree with the idea. Somehow, we should guarantee it.

One idea is:

```
    def is_target_step(self):
        # type: (int) -> bool

        all_trials = study.get_trials(deepcopy=False)
        n_trials = len([t for t in all_trials if t.state == TrialState.COMPLETE])

        if n_trials == 0:
            return False

        if n_trials < self._n_startup_trials:
            return False

        step = trial.last_step
        if step is None:
            return False

        n_warmup_steps = self._n_warmup_steps
        if step <= n_warmup_steps:
            return False

        if not _is_first_in_interval_step(
            step, trial.intermediate_values.keys(), n_warmup_steps, self._interval_steps
        ):
            return False

        return (step - self._n_warmup_steps) % self._interval_steps == 0

    def prune(self, study, trial):
        # type: (Study, FrozenTrial) -> bool
        
        if not is_target_step(epoch):
            return False

        direction = study.direction
        best_intermediate_result = _get_best_intermediate_result_over_steps(trial, direction)
        ...
```

toshihikoyanase:
@keisuke-umezawa 

Thank you for your proposal. It seems to be simple and straightforward, but I'm a bit concerned about a performance issue.

If `is_target_step` invokes `study.get_trials`, it takes communication time to access the storage. Besides, the number of `study.get_trials` calls will be doubled (i.e., in `prune` and `is_target_step`) in the worst case. So, I'd like to detect the target steps without accessing the storage if possible. Does it make sense?

keisuke-umezawa:
> So, I'd like to detect the target steps without accessing the storage if possible. Does it make sense?

I totally agree with it.

toshihikoyanase:
@keisuke-umezawa Thank you for your proposal! It sounds good to me.

toshihikoyanase:
@VladSkripniuk What do you think of @keisuke-umezawa 's suggestion? I think it's viable, but do you have any concerns?

toshihikoyanase:
@keisuke-umezawa Are you interested in a new PR with your proposed code?
This PR has been stale for a while and I'm thinking of the restart of the code review with it.

Crissman:
@VladSkripniuk This PR is stale and has conflicts with master, so closing for now. If there's further interest, we may want to take a look at #1161 (comment)

PR #1247, opened 2020-05-15
Add GitHub action that posts link to documentation

### LINKED ISSUE #1097: Add a description about document location stored in CircleCI artifact.
Thanks to #1008, documents of the submitted PRs are automatically built in CircleCI jobs and are stored in the artifact. But, some developers say that they cannot access the artifact tabs.

So, how about documenting the URL to the built document?
For example, a URL of the document in the artifact is as follows;

https://34609-122299416-gh.circle-artifacts.com/0/docs/build/html/index.html

Here, `34609` matched with the CircleCI's job ID. So, I think the document can be accessed with the following URL format and the job ID.

`https://{ciecleci-job-id}-122299416-gh.circle-artifacts.com/0/docs/build/html/index.html`

It would be more helpful if we add a bot that shows the document URLs to the PRs automatically, but it may be overengineering.

harupy:
## Motivation

Fixes #1097

## Description of the changes

Add a GitHub action that posts the link to the documentation stored in CircleCI.

(I captured this on my repo.)
image

### DIFF
--- a/.github/workflows/doc-link.yml
+++ b/.github/workflows/doc-link.yml
@@ -1,0 +1,13 @@
+name: Add Documentation Link
+
+on: status
+
+jobs:
+  doc-link:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: larsoner/circleci-artifacts-redirector-action@master
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          artifact-path: 0/docs/build/html/index.html
+          circleci-jobs: document
--- a/.github/workflows/doc-link.yml
+++ b/.github/workflows/doc-link.yml
@@ -6,8 +6,8 @@ jobs:
   doc-link:
     runs-on: ubuntu-latest
     steps:
-      - uses: larsoner/circleci-artifacts-redirector-action@master
-        with:
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


harupy:
https://help.github.com/en/actions/reference/events-that-trigger-workflows#status-event-status

> Note: This event will only trigger a workflow run if the workflow file is on the master or default branch.

We need to merge this PR to enable the action.

harupy:
I verified that this action works on my repo which has a similar CircleCI configuration to Optuna.

https://github.com/harupy/mlflow-extend/pull/68

hvy:
It seems like this job is executed quite often as seen in the log https://github.com/optuna/optuna/actions?query=workflow%3A%22Add+Documentation+Link%22. I'm wondering if we can't change the `on:` trigger to e.g. `push` instead.

harupy:
@hvy

> It seems like this job is executed quite often as seen in the log

There are currently 16 CI jobs and each job has two status update events (pending and success). This means `doc-link` is triggered 32 (16 x 2) times for one commit.

> I'm wondering if we can't change the on: trigger to e.g. push instead.

I don't think this is possible because `larsoner/circleci-artifacts-redirector-action` is called on the status event.

hvy:
Ah, I now understand it's waiting for the `status` from CircleCI to parse the artifact URL. Thanks for the explanation.

PR #1248, opened 2020-05-16
Fix experimental decorator

### LINKED ISSUE #1213: `experimental` decorator adds extra indents to docs
An extra indent appears in the docs when the [`experimental`](https://github.com/optuna/optuna/blob/2d5f24b06eed56ece72b8dfa878135bb4bb63779/optuna/_experimental.py#L60) decorator is added to a method of a class. This does not happen with a top level (i.e., not nested) class or function.

## Steps to reproduce

1. add an `experimental` decorator to a method of a class
2. make the document

## Reproducible examples
The following figure shows the docs of [`Study.enqueue_trial`](https://optuna.readthedocs.io/en/latest/reference/study.html#optuna.study.Study.enqueue_trial). An extra indent appears after the second paragraph.

image

## Expected behavior
no extra indent is shown after the second paragraph

For example, the docs of [`Study.get_trials`](https://optuna.readthedocs.io/en/latest/reference/study.html#optuna.study.Study.get_trials), which does not have the `experimental` decorator, appears as expected.

image

The following figure shows another example. The docs of the [`train`](https://optuna.readthedocs.io/en/latest/reference/integration.html#optuna.integration.lightgbm.train) function of LightGBMTuner has no extra indent. Although the function has an `experimental` decorator, it belongs to the top level classes.

image

harupy:
## Motivation

Fixes #1213 

## Description of the changes

Fix [`experimental`](https://github.com/optuna/optuna/blob/2d5f24b06eed56ece72b8dfa878135bb4bb63779/optuna/_experimental.py#L60) to decorate a class method properly.


harupy:
@crcrpar Thanks for the review! One question. We currently attach the experimental note to a function and class that do not have a docstring. For these functions and classes, I think just attaching a warning should be enough. What do you think?

crcrpar:
That completely makes sense🙂
Though I’m a bit negative about supporting that case if the implementation
gets complicated.

harupy:
@harupy The implementation will be like this:

```python
...

@functools.wraps(func)
def new_func(*args, **kwargs):
    warnings.warn(...)
    return func(*args, **kwargs)

if new_func.__doc__ is None:
    return new_func

new_func.__doc__ = ...

return new_func
```

crcrpar:
That sounds simple enough. It's up to you whether do that in this PR or a following PR. 😄

harupy:
@crcrpar I'll open a follow-up!

optuna/_experimental.py
```diff
@@ -72,10 +77,12 @@ def _experimental_wrapper(f: Any) -> Any:
...
-            docstring = _EXPERIMENTAL_DOCSTRING_TEMPLATE.format(ver=version)
             if func.__doc__ is None:
                 func.__doc__ = ""
-            func.__doc__ += docstring
+
+            docstring = _EXPERIMENTAL_DOCSTRING_TEMPLATE.format(ver=version)
+            indent = _get_indent(func.__doc__)
+            func.__doc__ = func.__doc__.strip() + textwrap.indent(docstring, indent) + indent
 
             # TODO(crcrpar): Annotate this correctly.
```
harupy, line 87: @crcrpar What does this comment mean?

crcrpar: I thought `def new_func(*args: Any, **kwargs: Any) -> Any`, `return func(*args, **kwargs)  # type: ignore
`, etc. were so lazy that they should be improved in the future.

tests/test_experimental.py
```diff
@@ -35,53 +53,59 @@ def test_str() -> None:
...
-            == "__init__(a, b, c)\n\n    "
-            + _experimental._EXPERIMENTAL_DOCSTRING_TEMPLATE.format(ver=version)
-        )
-
-        with pytest.warns(ExperimentalWarning):
-            decorated_sample("a", "b", "c")
+@pytest.mark.parametrize("version", ["1.1", 100, None])
+def test_experimental_raises_error_for_invalid_version(version: Any) -> None:
+    with pytest.raises(ValueError):
+        _experimental.experimental(version)
```
harupy, line 59: Separated the error test.

tests/test_experimental.py
```diff
@@ -35,53 +53,59 @@ def test_str() -> None:
...
+
+
+def test_experimental_decorator() -> None:
+    version = "1.1.0"
+    decorator_experimental = _experimental.experimental(version)
+    assert callable(decorator_experimental)
+    assert decorator_experimental.__name__ == "_experimental_wrapper"
```
harupy, line 66: Wonder if this assertion is necessary.

crcrpar: I don't think it necessary.

crcrpar:
@harupy Could you fix the `checks` failure?

harupy:
@crcrpar Thank you for the review :)

harupy:
@toshihikoyanase @hvy Could you please review this PR?

optuna/_experimental.py
```diff
@@ -28,6 +26,10 @@ def _validate_version(version: str) -> None:
         )
 
 
+def _get_docstring_indent(docstring: str) -> str:
+    return docstring.split("\n")[-1] if "\n" in docstring else ""
```
toshihikoyanase, line 30: [Question] This function seems to get the last line of the given docstring. Is it always corresponding to an indent?

harupy: > Is it always corresponding to an indent?

Not always. There are cases where this function does not work.

```python
# the last line contains characters
def f():
    """summary

    detail"""
    pass

# the last line has a different indent level
def f():
    """summary

    detail
  """
    pass

def f():
    """summary

    detail
      """
    pass
```

hvy: A not-so-strong drop by comment but I think it's okay assuming https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#docstrings. We could add warnings in case some of these assumptions above are broken but it also sounds like overengineering.

toshihikoyanase: Thank you for your explanation. Although `flake8` and `black` do not check such inconsistent indent levels, I think we can accept the current implementation as @hvy says.

PR #1259, opened 2020-05-19
Fix for O(N) queries being produced if even a single trial is cached

zzorba:
## Motivation
While debugging some performance issues when running with a remote database, we noticed that calling the following two methods in a different order produced WILDY different runtime behavior.
```python
study.best_params
study.trials_dataframe()
```

If `best_params` is called first, the call to `trials_dataframe` would take 5 minutes. If we inverted the calls, both methods would complete in a matter of seconds.

Digging in it turned out that `best_params` causes a single 'trial' to be cached in the system, which causes `trials_dataframe` to go from making a single bulk-database query to making N individual queries for each trial in the study (in our case, 700). This was probably especially noticeable in our case because the database was remote, but I expect even a local database would see improved performance from batch fetches.

## Description of the changes
This change modifies the cache code to prefer the bulk database fetch, which will almost always result in lower latency than making sequential queries. The cache is only used if less than 5 elements are missing (or not finished).

### DIFF
--- a/optuna/storages/rdb/storage.py
+++ b/optuna/storages/rdb/storage.py
@@ -731,16 +731,21 @@ class RDBStorage(BaseStorage):
     def get_all_trials(self, study_id, deepcopy=True):
         # type: (int, bool) -> List[FrozenTrial]
 
-        if self._finished_trials_cache.is_empty():
-            trials = self._get_all_trials_without_cache(study_id)
-            for trial in trials:
-                self._finished_trials_cache.cache_trial_if_finished(trial)
+        if not self._finished_trials_cache.is_empty():
+            trial_ids = self._get_all_trial_ids(study_id)
+            # Check if no more than 5 trials are missing from the cache.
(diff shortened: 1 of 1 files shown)


optuna/storages/rdb/storage.py
```diff
@@ -731,16 +731,21 @@ def _get_and_cache_trial(self, trial_id, deepcopy=True):
...
-            for trial in trials:
-                self._finished_trials_cache.cache_trial_if_finished(trial)
-
-            return copy.deepcopy(trials) if deepcopy else trials
+        if not self._finished_trials_cache.is_empty():
+            trial_ids = self._get_all_trial_ids(study_id)
+            # Check if no more than 5 trials are missing from the cache.
+            # This prevents a single item in the cache from resulting in O(N) database lookups when
+            # the bulk fetch below would be significantly faster.
+            if sum(not self._finished_trials_cache.get_cached_trial(trial_id) for trial_id in trial_ids) < 5:
```
ytsmiling, line 739: Please reformat code using black.

Issue #1237, opened 2020-05-13
XGBRegressor eval_metric error


CMobley7:
To Whom It May Concern,

I'm trying to use `XGBRegressor` with a pruning callback, similar to your example [here](https://github.com/optuna/optuna/blob/master/examples/pruning/xgboost_integration.py.). However,
I get either `XGBoostError: [05:03:32] /workspace/src/metric/metric.cc:31: Unknown metric function validation-rmse` with `eval_metric="validation-rmse"` or 

```
/usr/local/lib/python3.6/dist-packages/optuna/integration/xgboost.py in __call__(self, env)
     62             # Remove a third element: the stddev of the metric across the cross-valdation folds.
     63             evaluation_result_list = [(key, metric) for key, metric, _ in evaluation_result_list]
---> 64         current_score = dict(evaluation_result_list)[self._observation_key]
     65         self._trial.report(current_score, step=env.iteration)
     66         if self._trial.should_prune():

KeyError: 'validation-rmse'
``` 

with `eval_metric="rmse"` in `xgb_model.fit()`. My full code is below. If possible, can you advise on the correction configuration? Thanks in advance.

```
import optuna
import xgboost as xgb
from loguru import logger
from preprocess import build_train_test_data
from xgboost import XGBRegressor


def optuna_xgboost_stock_objective(
    trial,
    max_depth_low,
    max_depth_high,
    n_estimators,
):
    param = {
        "verbosity": 1,
        "objective": "reg:squarederror",
        "eval_metric": "rmse",
        "tree_method": "auto",
        "n_estimators": int(n_estimators),
        "booster": trial.suggest_categorical("booster", ["gbtree", "dart"]),
        "reg_lambda": trial.suggest_loguniform("reg_lambda", 1e-8, 1.0),
        "reg_alpha": trial.suggest_loguniform("reg_alpha", 1e-8, 1.0),
        "max_depth": int(trial.suggest_int("max_depth", max_depth_low, max_depth_high)),
        "learning_rate": trial.suggest_loguniform("learning_rate", 1e-8, 1.0),
        "gamma": trial.suggest_loguniform("gamma", 1e-8, 1.0),
        "grow_policy": trial.suggest_categorical(
            "grow_policy", ["depthwise", "lossguide"]
        ),
    }

    if param["booster"] == "dart":
        param["sample_type"] = trial.suggest_categorical(
            "sample_type", ["uniform", "weighted"]
        )
[...]

PR #1260, opened 2020-05-20
Add description of `observation_key` in `XGBoostPruningCallback`

smly:
## Motivation

This PR adds a description of `observation_key` in `XGBoostPruningCallback`.
Related issue: https://github.com/optuna/optuna/issues/1237.

XGBRegressor and XGBClassifier always uses a metric name like `validation_(index)-(metric)`, not `validation-(metric)`. This PR adds the description to notice this point.

## Description of the changes

This change adds the following description:

```
Note that when using the sklearn API callback,
the index number of `eval_set` must be included in the `observation_key`.
e.g., ``validation_0-error`` and ``validation_0-merror``.
```

### DIFF
--- a/optuna/integration/xgboost.py
+++ b/optuna/integration/xgboost.py
@@ -43,6 +43,9 @@ class XGBoostPruningCallback(object):
             ``validation-merror``. Please refer to ``eval_metric`` in
             `XGBoost reference <https://xgboost.readthedocs.io/en/latest/parameter.html>`_
             for further details.
+            Note that when using the sklearn API callback, the index number of `eval_set`
+            must be included in the `observation_key`, e.g., ``validation_0-error`` and
+            ``validation_0-merror``.
     """
 
     def __init__(self, trial, observation_key):
--- a/optuna/integration/xgboost.py
+++ b/optuna/integration/xgboost.py
@@ -44,7 +44,7 @@ class XGBoostPruningCallback(object):
             `XGBoost reference <https://xgboost.readthedocs.io/en/latest/parameter.html>`_
             for further details.
             Note that when using the sklearn API callback, the index number of `eval_set`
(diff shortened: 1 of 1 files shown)


optuna/integration/xgboost.py
```diff
@@ -43,6 +43,9 @@ class XGBoostPruningCallback(object):
             ``validation-merror``. Please refer to ``eval_metric`` in
             `XGBoost reference <https://xgboost.readthedocs.io/en/latest/parameter.html>`_
             for further details.
+            Note that when using the sklearn API callback, the index number of `eval_set`
```
toshihikoyanase, line 46: I have four comments on this line.

1. How about inserting this sentence before the previous line (e.g., `validation-merror`. When using the Scikit-Learn API, the index number of ...)? IMO, it will be easier to read if the explanation of ``observation_key`` is not separated. 
2. could you replace "sklearn API callback" with "Scikit-Learn API" since the [xgboost reference](https://xgboost.readthedocs.io/en/latest/python/python_api.html#module-xgboost.sklearn) calls it "Scikit-Learn" API?
3. How about adding a link to the reference? 
4. `eval_set` needs one more pair of backticks.

image

smly: Thanks for your review. Totally agree with your suggestion of (1), (2), (4).
As for (3), it is not documented that `observation_key` is passed to the callback function in a format like `validation_0-merror`. So I don't think adding [references](https://xgboost.readthedocs.io/en/latest/python/python_api.html#module-xgboost.sklearn) is helpful. How do you think about it?

toshihikoyanase: Thank you for your update. 

> As for (3), it is not documented that `observation_key` is passed to the callback function in a format like `validation_0-merror`. 

That makes sense. I agree to your opinion.

PR #1263, opened 2020-05-21
Move caching mechanism from `RDBStorage` to `_CachedStorage`.

ytsmiling:
## Motivation
There are two motivations.
### 1. Refactoring
There have been two caches in a single storage instance and this PR merges them into one.
### 2. Performance improvements
This PR improves the cache lookup mechanism and improves overall performance.

## Description of the changes
Move the cache mechanism in `RDBStorage` class into `_CachedStorage` class.

## Benchmark
The benchmark used the same code with #1140.
### Master
```
perf:optimize-study                               0:00:12.942167
perf:optimize-study                               0:00:45.675741
```
### This PR
```
perf:optimize-study                               0:00:12.357350
perf:optimize-study                               0:00:35.848832
```

### Updated (this PR)
```
perf:optimize-study                               0:00:24.644970
```
Also, this PR slows down get_all_trials without cache by 4x.


optuna/storages/rdb/storage.py
```diff
@@ -912,61 +914,169 @@ def get_trial_number_from_id(self, trial_id):
...
 
-        session = self.scoped_session()
+        frozen_trial = self._get_trials_from_trial_models(session, [trial_model])[0]
 
-        trial = models.TrialModel.find_or_raise_by_id(trial_id, session)
-        params = models.TrialParamModel.where_trial(trial, session)
-        values = models.TrialValueModel.where_trial(trial, session)
-        user_attributes = models.TrialUserAttributeModel.where_trial(trial, session)
-        system_attributes = models.TrialSystemAttributeModel.where_trial(trial, session)
+        return copy.deepcopy(frozen_trial)
```
sile, line 938: Just out of curiosity, why is this deep copy needed here?

optuna/storages/rdb/storage.py
```diff
@@ -912,61 +914,169 @@ def get_trial_number_from_id(self, trial_id):
...
+                models.TrialModel.datetime_complete,
+                models.TrialModel.number,
+            )
+            .filter(
+                ~models.TrialModel.trial_id.in_(cached_trial_ids),
+                models.TrialModel.study_id == study_id,
+            )
+            .all()
+        )
+        # On single-worker training, ``trial_models`` is always an empty list.
```
sile, line 969: Let me suggest the change below because "optimization" sounds a bit more Optuna-like term to me.
```suggestion
        # On single-worker optimization, ``trial_models`` is always an empty list.
```

optuna/storages/rdb/storage.py
ytsmiling: It will be unnecessary after #1274 and I'll remove the deep-copy as #1274 will be merged soon.

sile: I see. Thank you for the explanation!

optuna/storages/cached_storage.py
```diff
@@ -280,18 +323,26 @@ def get_trial(self, trial_id: int) -> FrozenTrial:
...
-                    trials[trial.number] = trial
-            return copy.deepcopy(trials) if deepcopy else trials
+            # The cache update will be moved into another method in the future.
+            if study_id not in self._studies:
+                self._studies[study_id] = _StudyInfo()
+            study = self._studies[study_id]
+            trials = self._backend._get_uncached_trials(study_id, study.cached_trial_ids)
+            if trials:
+                self._add_trials_to_cache(study_id, trials)
+                # ``self._study[study_id].trials`` should not have any dummy tirals.
```
hvy, line 340:
```suggestion
                # ``self._study[study_id].trials`` should not have any dummy trials.
```

optuna/storages/cached_storage.py
```diff
@@ -280,18 +323,26 @@ def get_trial(self, trial_id: int) -> FrozenTrial:
         with self._lock:
             trial = self._get_cached_trial(trial_id)
             if trial is not None:
-                return copy.deepcopy(trial)
+                return trial
```
hvy, line 326: Is performance degraded if we cache in this method? Could help leaving a comment.

ytsmiling: Since there is no method to know ``trial_id`` except for ``get_all_trials`` method and ``create_new_trial`` method, and trials are cached in those methods, the ``if trial is not None`` should be always `True` excluding some exceptional cases. Thus, adding codes that caches trials on cache miss won't affect the performance.
However, adding the code will indeed increase the consistency of implementation and reduce the confusion of readers. Thus, I'll the code, thank you.

optuna/storages/cached_storage.py
```diff
@@ -317,4 +368,29 @@ def _flush_trial(self, trial_id: int) -> bool:
...
+            datetime_complete=updates.datetime_complete,
         )
+
+    def _add_trials_to_cache(self, study_id: int, trials: List[FrozenTrial]) -> None:
+        study = self._studies[study_id]
+        max_trial_number = max(trial.number for trial in trials)
+        if len(study.trials) <= max_trial_number:
+            study.trials.extend(
+                [
+                    self._create_dummy_trial(i)
```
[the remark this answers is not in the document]
ytsmiling: I remember that I had some trouble to bypass `mypy`, but at least the current implementation seems to be free from the problem.
I tested several implementations including `List[FrozenTrial]`, `List[Optional[FrozenTrial]]`, and `Dict[int, FrozenTrial]`, and took benchmarks. It seems that their performances are comparable and I decided to replace the implementation with a dictionary. Sorry for the radical change after an initial review but I think this improves readability. Thank you for catching the ugly codes before this PR is merged.

optuna/storages/rdb/storage.py
```diff
@@ -555,6 +555,7 @@ def _update_trial(
         distributions_: Optional[Dict[str, distributions.BaseDistribution]] = None,
         user_attrs: Optional[Dict[str, Any]] = None,
         system_attrs: Optional[Dict[str, Any]] = None,
+        datetime_complete: Optional[datetime] = None,
```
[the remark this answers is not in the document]
ytsmiling: I'll add the doc. Thank you.

optuna/storages/rdb/storage.py
```diff
@@ -912,61 +914,169 @@ def get_trial_number_from_id(self, trial_id):
...
+        # On single-worker optimization, ``trial_models`` is always an empty list.
+        trials = self._get_trials_from_trial_models(session, trial_models) if trial_models else []
 
-        return copy.deepcopy(trials) if deepcopy else trials
+        self._commit(session)
+
+        return trials
+
+    @staticmethod
+    def _reduce_fn(
```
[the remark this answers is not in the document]
ytsmiling: Actually, taking advantage of ORM features can significantly reduce code complexity. I replaced the code from SQL-oriented implementation with ORM-oriented implementation.
The uncached `get_all_trials` method became over 4x slower to load 100 trials with 30 parameters, and hence `load_study` operations became slow as well, but the normal training time did not increase thanks to the cache mechanism (the training time might become longer in multi-worker settings, but I expect that the effect will still be negligible especially after we implement `load` method).
It's not easy to decide which implementations are better as `load_study` and subsequent uncached `get_all_trials` are frequently used for analyzing past training, the code reduction is significant and I'd like to adopt the ORM-oriented implementation this time. Sorry for bothering you with a large code change after the first review.

optuna/storages/rdb/storage.py
```diff
@@ -912,61 +914,169 @@ def get_trial_number_from_id(self, trial_id):
...
+                trial.state,
+                trial.value,
+                trial.datetime_start,
+                trial.datetime_complete,
+                {k: v[1].to_external_repr(v[0]) for k, v in params_dict[trial.trial_id].items()},
+                {k: v[1] for k, v in params_dict[trial.trial_id].items()},
+                user_attrs_dict[trial.trial_id],
+                system_attrs_dict[trial.trial_id],
+                intermediates_dict[trial.trial_id],
+                trial.trial_id,
```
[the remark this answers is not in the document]
ytsmiling: Thank you. I replaced with keyword arguments.

optuna/storages/cached_storage.py
ytsmiling: Thank you for the catch!

optuna/storages/rdb/storage.py
```diff
@@ -912,61 +914,169 @@ def get_trial_number_from_id(self, trial_id):
...
 
-        self._finished_trials_cache.cache_trial_if_finished(frozen_trial)
+        trials = self._get_uncached_trials(study_id, set())
 
-        # Terminate transaction explicitly to avoid connection timeout during transaction.
-        self._commit(session)
+        return copy.deepcopy(trials) if deepcopy else trials
 
-        return copy.deepcopy(frozen_trial) if deepcopy else frozen_trial
+    def _get_uncached_trials(self, study_id: int, cached_trial_ids: Set[int]) -> List[FrozenTrial]:
```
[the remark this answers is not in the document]
ytsmiling: I agree that the naming relies on `_CachedStorage` and is not appropriate. I replaced it with `_get_trials(..., excluded_trial_ids: ...)`. (I realized that your suggestion was `_get_all_trials` after I updated this PR. I'm neutral but if you prefer `_get_all_trials`, I'll re-update the name.)

optuna/storages/rdb/storage.py
```diff
@@ -912,61 +914,169 @@ def get_trial_number_from_id(self, trial_id):
...
 
-        session = self.scoped_session()
+        frozen_trial = self._get_trials_from_trial_models(session, [trial_model])[0]
 
-        trial = models.TrialModel.find_or_raise_by_id(trial_id, session)
-        params = models.TrialParamModel.where_trial(trial, session)
-        values = models.TrialValueModel.where_trial(trial, session)
-        user_attributes = models.TrialUserAttributeModel.where_trial(trial, session)
-        system_attributes = models.TrialSystemAttributeModel.where_trial(trial, session)
+        return frozen_trial
```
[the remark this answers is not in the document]
ytsmiling: It's not on purpose. Thank you for catching it. I'll wrap session using contextmanager not to forget committing/rollbacking changes in another PR.

ytsmiling:
@hvy Thank you for a detailed review. I made several large changes to address the comments and hope it improved the code quality.
@sile Thank you for the review. I addressed the review comments from @hvy and it involved non-trivial code changes. Sorry for bothering you but would you review this PR again?

sile:
Sure!

optuna/storages/cached_storage.py
ytsmiling: I remembered that it's non-trivial to get `study_id` from `trial_id`. `study_id` is required for the cache mechanism in this class and fetching the information requires a large chank of code. I'd like to keep the current implementation and not try to cache in this method.

optuna/storages/rdb/storage.py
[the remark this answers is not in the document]
hvy: It's not a strong opinion, let's leave it as is. Thanks!

optuna/storages/cached_storage.py
```diff
@@ -48,7 +54,7 @@ def __init__(self, backend: RDBStorage) -> None:
         self._backend = backend
         self._studies = {}  # type: Dict[int, _StudyInfo]
         self._trial_id_to_study_id_and_number = dict()  # type: Dict[int, Tuple[int, int]]
-        self._lock = threading.RLock()
+        self._lock = threading.Lock()
```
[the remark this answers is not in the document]
hvy: Ideally we should, but it's probably okay considering the cost trade off. Let's leave it as is.

> I think `Lock` is preferable for reducing bugs and also it runs much faster than `RLock`.

Absolutely right. Wondering why it was using `RLock` in the first place.

optuna/storages/cached_storage.py
```diff
@@ -120,33 +127,63 @@ def get_all_study_summaries(self) -> List[StudySummary]:
...
         with self._lock:
             if study_id not in self._studies:
                 self._studies[study_id] = _StudyInfo()
+            study = self._studies[study_id]
             self._trial_id_to_study_id_and_number[trial_id] = study_id, frozen_trial.number
-            if frozen_trial.state == TrialState.RUNNING:
-                self._studies[study_id].trials[frozen_trial.number] = frozen_trial
-            return trial_id
+            self._add_trials_to_cache(study_id, [frozen_trial])
+            # We cannot cache a WAITING trial.
```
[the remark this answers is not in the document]
hvy: Looks good.

optuna/storages/rdb/storage.py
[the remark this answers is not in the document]
hvy: Thanks for the very clear explanation. I agree, considering the vast reduction of code complexity (https://github.com/optuna/optuna/pull/1263/commits/5bde0c9707150af923e63864b85b4e84b8336235).

optuna/storages/rdb/storage.py
```diff
@@ -575,6 +572,9 @@ def _update_trial(
                 New user_attr. None when there are no updates.
             system_attrs:
                 New system_attr. None when there are no updates.
+            datetime_complete:
+                Completion time of the trial. Set only if and only if this method
```
sile, line 576: I'm not sure but it seems there is an extra "only".
```suggestion
                Completion time of the trial. Set if and only if this method
```

optuna/storages/cached_storage.py
```diff
@@ -24,12 +26,18 @@ def __init__(self) -> None:
...
+        self.datetime_complete = None  # type: Optional[datetime.datetime]
 
 
 class _StudyInfo:
     def __init__(self) -> None:
-        self.trials = dict()  # type: Dict[int, FrozenTrial]
+        # Trial number to corresponding FrozenTrial.
+        self.trials = {}  # type: Dict[int, FrozenTrial]
+        # A list of trials which do not require storage access to read latest attributes.
+        self.cached_trial_ids = set()  # type: Set[int]
```
hvy, line 37: Do you think it's preferable keeping this bookkeeping `set` even after changing the data structure of `self.trials` from a `list` to a `dict`?

optuna/storages/cached_storage.py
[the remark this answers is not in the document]
hvy: Got it. Again, code looks a bit simpler now, thank you. 

I was wondering whether we still need `cached_trial_ids` . I could create a separate comment on that.

optuna/storages/cached_storage.py
hvy: Ah you are right. That might be obvious from the code. A comment in the code might aid but it's definitely not a strong opinion.

optuna/storages/cached_storage.py
ytsmiling: The functionality is necessary even after we use `dict` instead of `list`. I think we should change the name of `cached_trial_ids` to something like `owned_or_immutable_trial_ids`, which keeps track of `RUNNING` trials created by this worker or finished trials created by any workers.
`self.trials` caches trials at least as-latest-as the time the `load` method (though it's still not implemented). However, some of the trials might be modified after they are added to the cache. In `get_all_trials` method, the `_CachedStorage` class caches trials into `self.trials` without checking trials' state. Thus, it can contain `RUNNING` trials owned by other workers. For such trials, we need to update the cache via `RDBStorage._get_trials(..., exclude_tiral_ids=...)` method on every `get_all_trials` method call.
(Note: After we implement an incremental-load feature, I think we can remove the necessity to track the `cached_trial_ids`. But it will be after v1.5 and might be after v2.0.)

optuna/storages/rdb/storage.py
ytsmiling: Yes, it is. I fixed the typo. Thank you.

optuna/storages/cached_storage.py
ytsmiling: I changed the name of `cached_trial_ids` to `owned_or_finished_trial_ids`.

hvy: Thank you for clarifying. 🙇

PR #1264, opened 2020-05-21
Cache study-related info in `_CachedStorage`.

ytsmiling:
## Motivation
In every `suggest` function, the default optuna sampler lookup `StudyDirection`. This PR removes redundant lookups and improve storage performance. 
This improvement is orthogonal with #1263.

## Description of the changes
Cache study name and `StudyDirection` in `_CachedStorage`.

## Microbench
The same configuration with #1140 was used.
### Master
```
perf:optimize-study                               0:00:12.942167
perf:optimize-study                               0:00:45.675741
```
### This PR
```
perf:optimize-study                               0:00:12.132461
perf:optimize-study                               0:00:40.967229
```

Improvements are larger when users use a larger number of parameters per trial.

### DIFF
--- a/optuna/storages/cached_storage.py
+++ b/optuna/storages/cached_storage.py
@@ -31,6 +31,8 @@ class _StudyInfo:
         self.trials = dict()  # type: Dict[int, FrozenTrial]
         self.updates = dict()  # type: Dict[int, _TrialUpdate]
         self.param_distribution = {}  # type: Dict[str, distributions.BaseDistribution]
+        self.direction = StudyDirection.NOT_SET  # type: StudyDirection
+        self.name = None  # type: Optional[str]
 
 
 class _CachedStorage(base.BaseStorage):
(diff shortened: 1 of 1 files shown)


optuna/storages/cached_storage.py
```diff
@@ -98,14 +112,30 @@ def get_study_id_from_name(self, study_name: str) -> int:
...
+                    return name
+
         return self._backend.get_study_name_from_id(study_id)
 
     def get_study_direction(self, study_id: int) -> StudyDirection:
 
+        with self._lock:
+            if study_id in self._studies:
+                direction = self._studies[study_id].direction
+                if direction != StudyDirection.NOT_SET:
```
sile, line 136: [MEMO] This condition can hold if a loaded study (i.e., created via `load_study()`) is used.

sile: It's not a strong opinion but I'm wondering that it might be good to cache the direction after retrieving it from the backend storage (line 137) to support the above `load_study()` case.
What do you think of it? > @ytsmiling

ytsmiling: I think we definitely cache that. Thank you for the catch and I'll add it.

PR #1265, opened 2020-05-21
Modify documentation and fix file extension in the test for AllenNLP integration.

himkt:
This PR modifies the documentation for AllenNLP integration.
And I also fix a small problem of incorrect specification for the file extension in the AllenNLP test.
(background: https://github.com/optuna/optuna/pull/1225#discussion_r427352698)

@toshihikoyanase @hvy 
Sorry for asking a review by name but could you please take a look?

### DIFF
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -26,11 +26,13 @@ def dump_best_config(input_config_file: str, output_config_file: str, study: opt
 
     Args:
         input_config_file:
-            Input configuration file used with :class:`~optuna.integration.AllenNLPExecutor`.
+            Input configuration file (Jsonnet) specified in
+            :class:`~optuna.integration.AllenNLPExecutor`.
         output_config_file:
-            Output configuration file.
+            Output configuration (Json) file.
         study:
-            An optimized study (``study.best_trial`` does not raise an error).
+            Instance of :class:`~optuna.study.Study`.
+            Note that :func:`~optuna.study.Study.optimize` mast have already called.
 
     """
     best_params = study.best_params
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -29,10 +29,10 @@ def dump_best_config(input_config_file: str, output_config_file: str, study: opt
             Input configuration file (Jsonnet) specified in
             :class:`~optuna.integration.AllenNLPExecutor`.
         output_config_file:
-            Output configuration (Json) file.
+            Output configuration (JSON) file.
         study:
             Instance of :class:`~optuna.study.Study`.
-            Note that :func:`~optuna.study.Study.optimize` mast have already called.
+            Note that :func:`~optuna.study.Study.optimize` must have been called.
 
     """
     best_params = study.best_params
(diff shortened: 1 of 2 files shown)


optuna/integration/allennlp.py
```diff
@@ -26,11 +26,13 @@ def dump_best_config(input_config_file: str, output_config_file: str, study: opt
 
     Args:
         input_config_file:
-            Input configuration file used with :class:`~optuna.integration.AllenNLPExecutor`.
+            Input configuration file (Jsonnet) specified in
+            :class:`~optuna.integration.AllenNLPExecutor`.
         output_config_file:
-            Output configuration file.
+            Output configuration (Json) file.
```
hvy, line 32: A nit but I think this is more common.
```suggestion
            Output configuration (JSON) file.
```

hvy: I'd personally remove the parenthesis here, mostly a matter of style.

optuna/integration/allennlp.py
```diff
@@ -26,11 +26,13 @@ def dump_best_config(input_config_file: str, output_config_file: str, study: opt
...
         study:
-            An optimized study (``study.best_trial`` does not raise an error).
+            Instance of :class:`~optuna.study.Study`.
+            Note that :func:`~optuna.study.Study.optimize` mast have already called.
```
hvy, line 35: Nits.
```suggestion
            Note that :func:`~optuna.study.Study.optimize` must have been called.
```

optuna/integration/allennlp.py
```diff
@@ -26,11 +26,13 @@ def dump_best_config(input_config_file: str, output_config_file: str, study: opt
 
     Args:
         input_config_file:
-            Input configuration file used with :class:`~optuna.integration.AllenNLPExecutor`.
+            Input configuration file (Jsonnet) specified in
+            :class:`~optuna.integration.AllenNLPExecutor`.
```
hvy, line 30: I think the previous "used with" sounds better.

hvy:
I labeled this PR `document` rather than `test` since the former changes seem more significant.

optuna/integration/allennlp.py
himkt: You're right, thank you.

optuna/integration/allennlp.py
himkt: OK, I reverted this change.

himkt:
@toshihikoyanase Thank you for the catch, I corrected it.

himkt:
I have noticed [the documentation](https://42449-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/integration.html?highlight=allennlp#optuna.integration.AllenNLPExecutor) and [the implementation](https://github.com/himkt/optuna/blob/modify-allennlp/optuna/integration/allennlp.py#L80-L88) do not match.
(The argument `include_package` only appears in the parameter section)

If someone knows what's wrong, please let me know.



Screen Shot 2020-05-24 at 16 59 49

hvy:
I reproduced it in my environment too. 🤔

hvy:
The issue seems to be caused by the experimental decorator. You can remove the decorate to observe that the documentation is rendered correctly.

PR #1274, opened 2020-05-22
Move deep-copies for optimization speed improvement.

ytsmiling:
## Motivation
The latest storage spec assumes that storage users won't directly modify return values of storage methods.
However, some methods did not take a copy before passing the return values to library users, which can result in undefined behavior. This PR fixes the issue.
Additionally, this PR removes unnecessary deep-copies and improves overall performance. 

## Description of the changes
- Copy `FrozenTrial`s of their attributes before passing them to users.
- Remove redundant deep-copies.

## Microbench
I used the same configuration with #1135 (1000 trials with 30 params w/ InMemoryDB).
```
This PR:
         117956150 function calls (112496390 primitive calls) in 98.936 seconds
Master:
         214822150 function calls (193042390 primitive calls) in 132.830 seconds
```

## Notes
I locally checked that any pruner method calls and sampler method calls under test/storage_test and tests/pruner_tests do not mutate `FrozenTrial` in their arguments.

### DIFF
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -635,7 +635,7 @@ class Trial(BaseTrial):
         # type: (str, BaseDistribution) -> Any
 
         if self._is_fixed_param(name, distribution):
-            param_value = self.system_attrs["fixed_params"][name]
+            param_value = self.storage.get_trial_system_attrs(self._trial_id)["fixed_params"][name]
         elif self._is_relative_param(name, distribution):
             param_value = self.relative_params[name]
         else:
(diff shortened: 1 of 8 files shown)


optuna/trial/_trial.py
```diff
@@ -768,7 +769,7 @@ def distributions(self):
             A dictionary containing all distributions.
         """
 
-        return self.storage.get_trial(self._trial_id).distributions
+        return copy.deepcopy(self.storage.get_trial(self._trial_id).distributions)
```
hvy, line 772: Should the `user_attrs` and `system_attrs` custom getters perform copied in a similar manner to this custom getter for `distributions`?

ytsmiling:
The updated PR achieved `89.966 seconds` in the same micro-benchmark, which is even better than the original PR.

optuna/trial/_trial.py
ytsmiling: Yes, we need deep-copies in user_attrs and system_attrs, and params as well. Thank you for pointing them out.

PR #1277, opened 2020-05-22
`log` argument for `suggest_int` of skopt integration

nzw0301:
## Motivation

One of the follow-up PRs related to #1201

## Description of the changes

Introduce`log` argument for `skopt` integration and add a test for it.

I'm not sure my implementation is correct: we can consider using `Integer(low, high, prior="log-uniform")` and then a sample is converted instead of using `Real`.

Ref: [Skopt's `Space.Integer`](https://scikit-optimize.github.io/stable/modules/generated/skopt.space.space.Integer.html#skopt.space.space.Integer)

### DIFF
--- a/optuna/integration/skopt.py
+++ b/optuna/integration/skopt.py
@@ -237,7 +237,7 @@ class _Optimizer(object):
                     np.round((value - distribution.low) / distribution.step) * distribution.step
                     + distribution.low
                 )
-                value = int(min(max(value, distribution.low), distribution.high))
+                value = min(max(value, distribution.low), distribution.high)
 
             params[name] = value
 
--- a/optuna/integration/skopt.py
+++ b/optuna/integration/skopt.py
@@ -237,6 +237,7 @@ class _Optimizer(object):
                     np.round((value - distribution.low) / distribution.step) * distribution.step
                     + distribution.low
                 )
+                value = int(min(max(value, distribution.low), distribution.high))
 
             params[name] = value
 
--- a/optuna/integration/skopt.py
+++ b/optuna/integration/skopt.py
@@ -177,7 +177,6 @@ class _Optimizer(object):
 
         dimensions = []
         for name, distribution in sorted(self._search_space.items()):
-            # TODO(nzw0301) support IntLogUniform
             if isinstance(distribution, distributions.UniformDistribution):
                 # Convert the upper bound from exclusive (optuna) to inclusive (skopt).
                 high = np.nextafter(distribution.high, float("-inf"))
@@ -189,6 +188,10 @@ class _Optimizer(object):
             elif isinstance(distribution, distributions.IntUniformDistribution):
                 count = (distribution.high - distribution.low) // distribution.step
                 dimension = space.Integer(0, count)
+            elif isinstance(distribution, distributions.IntLogUniformDistribution):
+                low = distribution.low - 0.5
+                high = distribution.high + 0.5
+                dimension = space.Real(low, high, prior="log-uniform")
             elif isinstance(distribution, distributions.DiscreteUniformDistribution):
                 count = int((distribution.high - distribution.low) // distribution.q)
                 dimension = space.Integer(0, count)
@@ -229,6 +232,11 @@ class _Optimizer(object):
                 value = value * distribution.q + distribution.low
             if isinstance(distribution, distributions.IntUniformDistribution):
                 value = value * distribution.step + distribution.low
+            if isinstance(distribution, distributions.IntLogUniformDistribution):
+                value = int(
+                    np.round((value - distribution.low) / distribution.step) * distribution.step
+                    + distribution.low
+                )
 
             params[name] = value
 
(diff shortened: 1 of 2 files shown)


optuna/integration/skopt.py
```diff
@@ -266,6 +274,8 @@ def _complete_trial_to_skopt_observation(self, study, trial):
                 param_value = (param_value - distribution.low) // distribution.q
             if isinstance(distribution, distributions.IntUniformDistribution):
                 param_value = (param_value - distribution.low) // distribution.step
+            if isinstance(distribution, distributions.IntLogUniformDistribution):
+                param_value = np.log(param_value)
```
nzw0301, line 278: Actually, this causes an error when `np.log` takes `param_value=1` in the following example.

```python
import optuna

def objective(trial):
    x = trial.suggest_uniform('x', -10, 10)
    y = trial.suggest_int('y', 1, 9, 2, True)
    return x**2 + y

sampler = optuna.integration.SkoptSampler()
```

nzw0301: Oops,  these lines are unnecessary.

HideakiImamura:
Thanks for your contribution! In my opinion, using `skopt.space.space.Integer` might be better because it matches other cases in Optuna such as `IntUniformDistribution` and `DiscreteUniformDistribution`. Its usage is almost the same as `IntUniformDistribution`:
```python
count = (distributions.high - distributions.low) // distributions.step
dimension = space.Integer(0, count, prior=`log-uniform`)
```

nzw0301:
@HideakiImamura Thank you for your comment! I note that `low` should be bigger than `0` when `prior='log-uniform'`:

```python
count = (distributions.high - distributions.low) // distributions.step + 1
dimension = space.Integer(1, count, prior=`log-uniform`)
```

Before I update this PR, I'm curious about the difference between skopt's sampling and optuna's one: should they return (almost) same samples or not?

When I implemented this PR by using `space.Integer` as follows, generated samples are not same as optuna's samples.

```python
import numpy as np
from skopt.space import Integer

low = 1
high = 3
step = 1
num_samples = 10000

count = (high - low) // step + 1
dimension = Integer(low=1, high=count, prior='log-uniform')
samples = ((dimension.rvs(n_samples=num_samples, random_state=1)) - 1 ) * step + low
print(np.bincount(samples))
```

The outputs is `array([   0, 3702, 4655, 1643])`. In other words,

| sample | freq | 
|----:|-----:|
| 1   | 3702|
| 2  | 4655|
| 3   | 1643|

On the other hand, optuna's `suggest_int(log=True)` generates 

| sample | freq | 
|----:|-----:|
| 1   | 5682 |
| 2  | 2622 |
| 3   | 1696 |

Optuna's code is as follows:

```python
import numpy as np
from skopt.space import Integer

low = 1
high = 3
step = 1
num_samples = 10000

sampler = samplers.RandomSampler(seed=1)
distribuion = IntLogUniformDistribution(low=low, high=high, step=step)

samples = []
for i in range(num_samples):
    trial = Trial(study, study._storage.create_new_trial(study._study_id))    
    
    samples.append(trial._suggest("x", distribuion))

samples = np.array(samples)
print(np.bincount(samples))
```

HideakiImamura:
@nzw0301 Thanks for your kind survey. Hmm... It is strange. Actually, `skopt.space.space.Integer` seems to have some kind of bug. But, if I increase `high` from 3 to 20, the generated samples look nice except `sample = 1` as follows.

スクリーンショット 2020-05-25 19 59 42

However, I think this can be a fatal problem when the `high-low` is small. Therefore, I think your implementation is better.

nzw0301:
@HideakiImamura Thanks! So, could you review this PR?

optuna/integration/skopt.py
```diff
@@ -229,6 +232,11 @@ def ask(self):
                 value = value * distribution.q + distribution.low
             if isinstance(distribution, distributions.IntUniformDistribution):
                 value = value * distribution.step + distribution.low
+            if isinstance(distribution, distributions.IntLogUniformDistribution):
+                value = int(
+                    np.round((value - distribution.low) / distribution.step) * distribution.step
+                    + distribution.low
+                )
```
HideakiImamura, line 239: I think we need to truncate so that its value is included in `[low ,high]`.

nzw0301: Thanks! I have updated it.

tests/integration_tests/test_skopt.py
```diff
@@ -154,8 +158,10 @@ def _objective(trial):
     p8 = trial.suggest_discrete_uniform("p8", 2.2, 2.2, 0.5)
     p9 = trial.suggest_categorical("p9", ["9", "3", "0", "8"])
     assert isinstance(p9, str)
+    p10 = trial.suggest_int("p10", 1, 8, log=True)
+    p11 = trial.suggest_int("p11", 3, 9, step=2, log=True)
```
HideakiImamura, line 162: I think it would be better to put `suggest_int(log=True)` just below `suggest_int` to make it look better.

nzw0301: Thanks! I updated it.

HideakiImamura:
I have noticed that in the line 192-193 we should expand `low` and `high` by not `0.5` but `0.5 * step`. But, if the recent PR #1329 is merged, it is prohibited to set `step >= 2` and `log = True`. Therefore, we don't have to do any action on this point now.

nzw0301:
@HideakiImamura @hvy Thank you for your reviews and comments!

hvy:
Thank you for your contribution, as always. 🙂

PR #1278, opened 2020-05-23
Create a documentation for the root namespace `optuna`.

keisuke-umezawa:
### Background
Currently, several classes and functions are imported in `optuna/__init__.py`, but we do not have the documentation for `optuna/__init__.py`.

e.g. `create_study`
- `create_study` is defined in `optuna/study.py`.
- `create_study` is imported in `optuna/__init__.py`.
- `create_study` description is written in `API Reference / Study` doc.
- No documentation for `optuna/__init__.py`

### Purpose
I think that we should create the documentation for the root namespace, and describe the classes and functions imported in `optuna/__init__.py` there.

## Description of the changes

To show what I want to do, I just modified the documentation for `create_study`.

## Discussion points
I checked the way of current branch implementation.

### What classes and functions should be imported in `optuna/__init__.py`?

From [this reddit discussion](https://www.reddit.com/r/Python/comments/1bbbwk/whats_your_opinion_on_what_to_include_in_init_py/), there are three type of policies, and optuna follows third one. 

- [ ] 1. Leave the __init__.py blank. This enforces explicit imports and thus clear namespaces. I've read Alex Martelli post in favor of this option on various questions at Stack Overflow. The cons are, the user of the package has to import seperate modules and call them with the dot notation.
- [ ] 2. Import all modules in __init__.py. The user doesn't have to do multiple imports. The cons are explicit vs implicit, and also as Martelli puts it (paraphrased), "if that's how your package works, maybe it should all go in a single module anyway".
- [x] 3. Import key functions from various modules directly into the package namespace. If you restructure modules, you still have the option to keep the same API for end users. Cons, it dirties the namespace, and very implicit / hacky.

### What is a good title for the documentation of API?
- [x] 1. `Public API`, or `Key API`. (Currently, in oputuna, documentation titles are named from functionality.)
- [ ] 2. `optuna` (Other libraries s.t. pytorch use module names for titles of API documentation. If we choose it, we need to change all titles to their module names.)

### Where to write the documentation?

e.g. `create_study`
- [ ] 1. It should be documented only in `optuna/__init__.py`. (numpy and pytorch follow this one.)
- [ ] 2. It should be documented only in `optuna/study/__init__.py`.
- [x] 3. It should be documented  in both `optuna/study/__init__.py` and `optuna/__init__.py`.

### DIFF
--- a/docs/source/reference/core.rst
+++ b/docs/source/reference/core.rst
@@ -10,3 +10,6 @@ Core
 .. autofunction:: get_all_study_summaries
 
 .. autofunction:: load_study
+
+.. autoclass:: TrialPruned
+    :members:
--- a/docs/source/reference/core.rst
+++ b/docs/source/reference/core.rst
@@ -6,25 +6,25 @@ Core
 .. autofunction:: create_study
 
 .. seealso::
-    :func:`optuna.study.create_study`.
+    This is an alias for :func:`optuna.study.create_study`.
 
 .. autofunction:: load_study
 
 .. seealso::
-    :func:`optuna.study.load_study`.
+    This is an alias for :func:`optuna.study.load_study`.
 
 .. autofunction:: delete_study
 
 .. seealso::
-    :func:`optuna.study.delete_study`.
+    This is an alias for :func:`optuna.study.delete_study`.
 
 .. autofunction:: get_all_study_summaries
 
 .. seealso::
-    :func:`optuna.study.get_all_study_summaries`.
+    This is an alias for :func:`optuna.study.get_all_study_summaries`.
 
 .. autoclass:: TrialPruned
     :members:
 
 .. seealso::
-    :class:`optuna.exceptions.TrialPruned`.
+    This is an alias for :class:`optuna.exceptions.TrialPruned`.
--- a/docs/source/reference/core.rst
+++ b/docs/source/reference/core.rst
@@ -5,11 +5,26 @@ Core
 
 .. autofunction:: create_study
 
-.. autofunction:: delete_study
-
-.. autofunction:: get_all_study_summaries
+.. seealso::
+    :func:`optuna.study.create_study`.
 
 .. autofunction:: load_study
 
+.. seealso::
+    :func:`optuna.study.load_study`.
+
+.. autofunction:: delete_study
+
+.. seealso::
+    :func:`optuna.study.delete_study`.
+
+.. autofunction:: get_all_study_summaries
+
+.. seealso::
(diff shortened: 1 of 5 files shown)


docs/source/reference/public.rst
```diff
@@ -0,0 +1,6 @@
+.. module:: optuna
+
+Public API
```
hvy, line 3: Hm, cannot come up with a definite idea and just naming a few (without really motivating them) but how about "Core", "Basic", "Common"? The fact that it's "public" is already implied and could be slightly confusing as one might think that the other parts are non-public, which is not the case. This is a minor but "API" is a bit redundant as it's mentioned in the section header "API reference". 🤔

hvy: Using module names might more straight forward but maybe beyond the scope of this PR.

HideakiImamura:
Thank you for the PR that makes a huge improvement in documentation! Some modules, classes and functions are now placed in `__init__.py` for convenience. I think the descriptions of functions and classes (e.g. `create_study` and `Study`) can be used in the original files, but the descriptions of each module (e.g. `study`) should be created anew?

HideakiImamura:
I realized after posting the above discussion that a brief description of each module (e.g. `study`) would be outside the scope of this PR. Anyway, I think the idea of creating a description of a function or class in `optuna/__init__.py` is a great idea! Let's keep it going!

docs/source/reference/public.rst
keisuke-umezawa: > Using module names might more straight forward but maybe beyond the scope of this PR.
 
I agree with this idea, but it will change all structure of reference documentation. I will hold a discussion about it later.

docs/source/reference/core.rst
```diff
@@ -0,0 +1,30 @@
...
+.. autofunction:: get_all_study_summaries
+
+.. seealso::
+    :func:`optuna.study.get_all_study_summaries`.
+
+.. autoclass:: TrialPruned
+    :members:
+
+.. seealso::
+    :class:`optuna.exceptions.TrialPruned`.
```
hvy, line 30: Not a strong opinion but how about clarifying the relationship, that "This is an alias for :class:`optuna.exceptions.TrialPruned`" since the purpose isn't obvious otherwise?

keisuke-umezawa:
@hvy @HideakiImamura Thank you for reviewing!

> Using module names might more straight forward but maybe beyond the scope of this PR.

btw, how do you think of it?
Should I create a issue to discuss it?

PR #1280, opened 2020-05-25
Avoid mutable default argument in `AllenNLPExecutor.__init__`.

hvy:
## Motivation

Prevent unintended bugs caused by a mutable default arguments in `AllenNLPExecutor.__init__`.

## Description of the changes

Changes the default value of previously mutable argument to `AllenNLPExecutor.__init__`.


hvy:
@himkt could you take a look this PR?

hvy:
It was @ytsmiling how found it by the way. Thanks.

PR #1283, opened 2020-05-25
Add a workflow to create distribution packages.

toshihikoyanase:
## Motivation

This PR will automate tasks to release Python packages at TestPyPI and PyPI using [`pypi-publish`](https://github.com/marketplace/actions/pypi-publish).

## Description of the changes

It creates a source tarball, checks it with `twine`, and uploads it to TestPyPI and PyPI.
It will be triggered when the `v*` tags (e.g., `v1.5.0`) are pushed. Note that we need to push tags manually if we want to release the package prior to release notes.

## References

- [pypi-publish config in AllenNLP](https://github.com/allenai/allennlp/blob/master/.github/workflows/master.yml)
- [pypi-publish](https://github.com/marketplace/actions/pypi-publish)
- [Publishing package distribution releases using GitHub Actions CI/CD workflows](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -4,7 +4,7 @@ name: Publish distributions to TestPyPI and PyPI
 on:
   push:
     tags:
-    - v*
+      - v*
   # TODO(toshihikoyanase): Remove pull_request when you enable PyPI upload.
   pull_request:
     branches:
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -5,6 +5,10 @@ on:
   push:
     tags:
     - v*
+  # TODO(toshihikoyanase): Remove pull_request when you enable PyPI upload.
+  pull_request:
+    branches:
+      - master
 
 jobs:
   build-n-publish:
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  build-n-publish:
+    name: Build and publish Python distributions to TestPyPI and PyPI
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
```
HideakiImamura, line 18: I'm not sure, but is it reasonable to always use python3.7 here?

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    - name: Build a tar ball
+      run: >-
+        python setup.py sdist
+
+    - name: Verify the distributions
+      run: twine check dist/*
+
+    - name: Publish distribution to Test PyPI
+      # Do not upload the package from forked repository.
```
HideakiImamura, line 32: Is it correct that the Github repository constrains to be `optuna/optuna`, so we can't automatically upload from forked repositories? If so, I think it would be appropriate to make this comment less imperative, as follows.
```suggestion
      # The following upload action cannot be executed in the forked repository.
```

.github/workflows/pypi-publish.yml
toshihikoyanase: Thank you for your comment. I just followed [the user guide](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#checking-out-the-project-and-building-distributions), and I noticed that I didn't check whether we could build the package on GitHub actions.

What do you think if I divide this PR into two?
1. Create distribution packages and store them as artifacts.
2. Publish the package via TestPyPI and PyPI.

.github/workflows/pypi-publish.yml
toshihikoyanase: Thank you for your suggestion. That makes sense. I'll also fix the same comment in `Publish distribution to PyPI`.

.github/workflows/pypi-publish.yml
HideakiImamura: I agree with your idea of splitting the PR into two parts. I think the two tasks are separated semantically.

hvy:
Diffs between publishing on PyPI vs Test PyPI seems rather small (just 2 lines?). Wondering if we can first run it on Test PyPI to make sure it works as intended.

toshihikoyanase:
@hvy Thank you for your comment. As you mentioned, we can use TestPyPI to check if this workflow correctly works. I'll divide this PR into three:
1. Create distribution packages and store them as artifacts.
2. Publish the package via TestPyPI
3. Publish the package via PyPI

toshihikoyanase:
@hvy, @HideakiImamura 
I removed steps to upload the distribution packages to TestPyPI and PyPI and add a step to upload it to the artifact. It successfully worked as can be seen in [this job](https://github.com/optuna/optuna/pull/1283/checks?check_run_id=708963932). I also submitted an another PR about TestPyPI (#1291).
Please take another look.

PR #1285, opened 2020-05-26
Fix `experimental` decorator to decorate a class properly

### LINKED ISSUE #1281: `experimental` decorator breaks class documentation.
The `experimental` decorator used on classes break documentation. This could be one manifestation but there is an issue with how the documentation including type hints are propagated to the decorated class. This does not apply for free functions.

See https://github.com/optuna/optuna/pull/1265#issuecomment-633195955 for how it may break.

## Expected behavior

Class documentation should not be altered by applying the experimental decorator.

## Steps to reproduce

1. Apply the experimental decorator to a class.
1. Build the document (`cd docs && make html`) 
1. Open the rendered documentation and note that the class signatures is broken.

- An issue regarding the indentation https://github.com/optuna/optuna/issues/1213.

harupy:
## Motivation

Fix #1281

`experimental` decorator has two issues when decorating a class:
1. It drops type annotations.
2. It drops keyword-only arguments.

This PR aims to fix them.

## Description of the changes

- Decorate `wrapped_init` by `functools.wraps` to preserve an original class signature.
- Remove `_make_func_spec_str` because it's no longer used.
- Fix the tests.

### DIFF
--- a/tests/test_experimental.py
+++ b/tests/test_experimental.py
@@ -11,28 +11,9 @@ def _sample_func(_: Any) -> int:
     return 10
 
 
-def _f() -> None:
-    pass
-
-
-def _g(a: Any, b: Any = None) -> None:
-    pass
-
-
-def _h(a: Any = None, b: int = 10) -> None:
-    pass
-
-
 class _Sample(object):
     def __init__(self, a: Any, b: Any, c: Any) -> None:
         pass
-
-
-def test_str() -> None:
-
-    assert _experimental._make_func_spec_str(_f) == "_f()\n\n    "
-    assert _experimental._make_func_spec_str(_g) == "_g(a, b=None)\n\n    "
-    assert _experimental._make_func_spec_str(_h) == "_h(a=None, b=10)\n\n    "
 
 
 @pytest.mark.parametrize("version", ["1.1.0", "1.1", 100, None])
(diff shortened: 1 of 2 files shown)


tests/test_experimental.py
```diff
@@ -74,10 +55,9 @@ def test_experimental_class_decorator(version: str) -> None:
 
         decorated_sample = decorator_experimental(_Sample)
         assert decorated_sample.__name__ == "_Sample"
-        assert (
-            decorated_sample.__doc__
-            == "__init__(a, b, c)\n\n    "
-            + _experimental._EXPERIMENTAL_DOCSTRING_TEMPLATE.format(ver=version)
+        assert decorated_sample.__init__.__name__ == "__init__"
```
harupy, line 58: Added an assertion here to ensure that `__init__.__name__` is not overriden.

harupy:
### Before:

Screen Shot 2020-05-26 at 12 32 24

https://42629-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/integration.html#optuna.integration.AllenNLPExecutor

### After:

Screen Shot 2020-05-26 at 12 30 59

https://42705-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/integration.html#optuna.integration.AllenNLPExecutor

PR #1291, opened 2020-05-26
Add TestPyPI release to workflow.

toshihikoyanase:
## Motivation
This PR depends on #1283. An additional PR to add the PyPI release will follow this PR.

## Description of the changes

This PR adds a step to upload distribution packages to TestPyPI. Please review #1283 in advance.

## Reference

- [pypi-publish](https://github.com/marketplace/actions/pypi-publish)
- [Publishing package distribution releases using GitHub Actions CI/CD workflows](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -1,10 +1,10 @@
 name: Publish distributions to TestPyPI and PyPI
 
-# Only activate if the v* tag is pushed.
+# Only activate if the release is published.
 on:
-  push:
-    tags:
-      - v*
+  release:
+    types:
+      - published
 
 jobs:
   build-n-publish:
@@ -30,7 +30,7 @@ jobs:
 
     - name: Publish distribution to Test PyPI
       # The following upload action cannot be executed in the forked repository.
-      if: startsWith(github.event.ref, 'refs/tags') && github.repository == 'optuna/optuna'
+      if: github.repository == 'optuna/optuna'
       uses: pypa/gh-action-pypi-publish@v1.1.0
       with:
         user: __token__
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -5,10 +5,6 @@ on:
   push:
     tags:
       - v*
-  # TODO(toshihikoyanase): Remove pull_request when you enable PyPI upload.
-  pull_request:
-    branches:
-      - master
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi-publish.yml
```diff
@@ -32,11 +28,13 @@ jobs:
...
-      uses: actions/upload-artifact@v1
+    - name: Publish distribution to Test PyPI
+      # The following upload action cannot be executed in the forked repository.
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') && github.repository == 'optuna/optuna'
+      uses: pypa/gh-action-pypi-publish@v1.1.0
       with:
-        name: optuna-dist
-        path: dist
+        user: __token__
+        password: ${{ secrets.test_pypi_password }}
```
hvy, line 37: I guess adding the secret to this organization/repository is a follow up task?

.github/workflows/pypi-publish.yml
```diff
@@ -32,11 +28,13 @@ jobs:
     - name: Verify the distributions
       run: twine check dist/*
 
-    # Save the package to the GitHub Actions' artifact. It will be deleted after 90 days.
-    - name: Save dist directory
-      uses: actions/upload-artifact@v1
+    - name: Publish distribution to Test PyPI
+      # The following upload action cannot be executed in the forked repository.
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') && github.repository == 'optuna/optuna'
```
hvy, line 33: Just wondering but we need the condition on `event_name` (or is it better to have in some sense)?

toshihikoyanase: I think it is no longer necessary because we remove `pull_request` from the trigger. I'll remove it. Thank you!

.github/workflows/pypi-publish.yml
toshihikoyanase: Yes, it is. I added it. Thanks!

.github/workflows/pypi-publish.yml
hvy: I agree, thanks!

.github/workflows/pypi-publish.yml
hvy: Thanks for the quick action.

PR #1292, opened 2020-05-26
Reorder arguments of `Trial.suggest_float`.

toshihikoyanase:
## Motivation
This is a followup PR for #1201. It addresses [this comment](https://github.com/optuna/optuna/pull/1201#discussion_r427863114), and makes the order of the `Trial.suggest_float` arguments consistent with the `Trial.suggest_int` arguments. I think it is helpful for users if we include this change into v1.5.0.

## Description of the changes

It reorders the argument of `Trial.suggest_float`. More concretely, it swaps `log` for `step`. This change does not affect user code because both of them are keyword-only arguments.

### DIFF
--- a/optuna/integration/chainermn.py
+++ b/optuna/integration/chainermn.py
@@ -1,4 +1,5 @@
 import gc
+from typing import Optional
 import warnings
 
 from optuna.logging import get_logger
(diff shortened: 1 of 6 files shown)


toshihikoyanase:
@hvy Thank you for your careful review. I reordered the arguments in the docstring for `Trial.suggest_float`. Fixed in https://github.com/optuna/optuna/pull/1292/commits/58ec8b508d693e854a9e9d5c59cb9f727e219dd3.

PR #901, opened 2020-02-06
Create dockerimage.yml

crcrpar:
rel: https://github.com/optuna/optuna/issues/842

PR #874, opened 2020-01-29
Add `Dockerfile`, GitHub Actions, VSCode `.devcontainer`.

crcrpar:
Resolves #842 

With this PR, Optuna repository will build docker images when master is updated or new version is released and then push those images to DockerHub and GitHub Packages.

Ref: [repo](https://github.com/crcrpar/gh-actions-optuna) & [dockerhub](https://hub.docker.com/repository/docker/crcrpar/gh-actions-optuna)

__motivation__
Providing official docker images and VSCode .devcontainer config for new users to try & develop Optuna with ease.

Also, I wanted to enable VSCode's jupyter integration, however, it does not seem to work according to https://github.com/microsoft/vscode-remote-release/issues/1930.

__TODO__

- [ ] update `dockerimage.yml` to handle `release` pattern.

PR #1295, opened 2020-05-27
Add a step to push images to Docker Hub.

toshihikoyanase:
## Motivation
This PR addresses a TODO item in #901. #901 creates docker images and this PR uploads them to Docker Hub.

## Description of the changes

- Add a step to upload images to Docker Hub. I extracted the code in #874 by @crcrpar.
- [Opinion Wanted] I wanted to use the access token of the `optuna` organization, but DockerHub does not allow such tokens. Instead, it provides a personal token. So, I parameterized the user name in addition to the access token and read them from GitHub Secret. If you have any concerns, please let me know.

Currently, the upload step will be skipped because I didn't set docker hub token to the GitHub Secret.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -46,7 +46,7 @@ jobs:
 
     - name: Login & Push to Docker Hub
       # The following push action cannot be executed in the forked repository.
-      if: github.repository == 'optuna/optuna'
+      if: github.repository == 'optuna/optuna' && env.DOCKER_HUB_USER != ''
       env:
         DOCKER_HUB_USER: ${{ secrets.DOCKER_HUB_USER }}
         DOCKER_HUB_TOKEN: ${{ secrets.DOCKER_HUB_TOKEN }}
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -25,6 +25,7 @@ jobs:
         python_version: ['3.5', '3.6', '3.7', '3.8']
         build_type: ['', 'dev']  # "dev" installs all the dependencies including pytest.
 
+    if: github.repository == 'optuna/optuna'
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


crcrpar:
>     * [Opinion Wanted] I wanted to use the access token of the `optuna` organization, but DockerHub does not allow such tokens. Instead, it provides a personal token. So, I parameterized the user name in addition to the access token and read them from GitHub Secret. If you have any concerns, please let me know.

Username doesn't matter if we can pull those docker images from DockerHub Optuna organization.

toshihikoyanase:
@crcrpar Thank you for your comment.

> Username doesn't matter if we can pull those docker images from DockerHub Optuna organization.

I think we don't need to use username to pull images, but it is required when we push the images. So, I use `docker login` with the username. If my understanding is wrong, please point it out to me.

crcrpar:
So the concern seems to be that there exists Optuna organization account on both GitHub and DockerHub, we're going to provide official docker images that are built and updated automatically by CI, which are pretty cool, but, unfortunately, those images are probably pushed to DockerHub by some personal DockerHub account, for example, you and me while personally those should be credited to Optuna organization. Then I suppose it's reasonable to create a DockerHub (and possibly GitHub) account named like `optuna-bot` for that purpose.

toshihikoyanase:
> while personally those should be credited to Optuna organization.

That makes sense. I created the `optunabot` account for this purpose. Sorry but dockerhub does not accept the hyphen in the account name.

HideakiImamura:
Thanks for the PR. The update is LGTM. 

Let me ask a question. How do I check for validity? Do we have to check it out in a production environment after merging? I think if it can pass CI by removing `on: pull_request`, we should do this before the merge.

.github/workflows/dockerimage.yml
```diff
@@ -43,3 +44,11 @@ jobs:
     - name: Build the Docker image
       run: |
         docker build . --build-arg PYTHON_VERSION=${{ matrix.python_version }} --build-arg BUILD_TYPE=${{ matrix.build_type }} --file Dockerfile --tag "${HUB_TAG}"
+
+    - name: Login & Push to Docker Hub
```
crcrpar, line 48: Personally, this step should only be triggered by master push and release.

toshihikoyanase:
@HideakiImamura Thank you for your review!

> Do we have to check it out in a production environment after merging? I think if it can pass CI by removing on: pull_request, we should do this before the merge.

What do you think if I test the action with my personal account? It cannot check some issues like the values of GitHub Secret, but I think we can confirm the logic of `dockerimage.yml`.

.github/workflows/dockerimage.yml
toshihikoyanase: That makes sense. We can keep the build step for testing purposes, and we'll be aware of the failure of docker build. Fixed in https://github.com/optuna/optuna/pull/1295/commits/dd16c2aeba3cfe1f15b2cadc6cbcec9bf1c5229c.
Thank you for your suggestion.

toshihikoyanase:
@HideakiImamura 

I tested the login & push step using my forked repository. The PR is https://github.com/toshihikoyanase/optuna/pull/1, and the images were pushed successfully as listed in https://hub.docker.com/r/toshihikoyanase/optuna/tags.

I also confirmed that the docker container can be created as expected.

```console
$ docker run toshihikoyanase/optuna:py3.7 -it bash
# python
>>> import optuna
>>> optuna.__version__
'1.5.0'
```

PTAL.

PR #1298, opened 2020-05-27
Add `read_trials_from_remote_storage` method to Storage implementations.

ytsmiling:
## Motivation
This PR implements `read_from_remote_storage`, which has been a TODO comment in the `BaseStorage` class doc (the method has been referred to be `load` method, and the name is subject for discussion).

This PR
1. improves optimization speed with RDB backend, and
2. fix trial-inconsistency bug between parameter sampling.

## Bugfix
Optuna's samplers have synced storage immediately, and thus, "past trials" might differ between `trial.suggest_*` method calls in a single trial. This can introduce unintentional behavior similar to #1166 in multi-worker settings. This PR mitigates the problem by loading remote storage's content only at the very beginning of each trial (and each prune call). Note, however, this PR cannot fix the behavior w/ multi-thread or w/ `RedisStorage`. The problem with `RedisStorage` will be resolved when we extend `_CachedStorage` and wrap `RedisStorage` as well.

## Description of the changes
This PR add implementations of `read_from_remote_storage` method.
Additionally, this PR modifies samplers' and study's implementation so that they avoid unintentional storage accesses.

## Microbenchmark
This PR added almost no overhead on optimization with `InMemoryStorage`.
I used the same setting with #1140 to measure performance improvement for optimization w/ RDB backend.

### Master
```
{'n_study': 1, 'n_trial': 500, 'n_param': 30}
perf:optimize-study                               0:03:29.107132
```
### This PR
```
{'n_study': 1, 'n_trial': 500, 'n_param': 30}
perf:optimize-study                               0:01:11.178871
```

### DIFF
--- a/optuna/storages/base.py
+++ b/optuna/storages/base.py
@@ -56,7 +56,7 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
     of a study, not necessarily for the attributes of a `Trial`.
     However, if the `read_from_remote_storage(study_id)` method is called, any successive reads on
     the `state` attribute of a `Trial` are guaranteed to return the same or more recent values than
-    the value at the time of call to the `load` method.
+    the value at the time of the call to the `read_from_remote_storage(study_id)` method.
     Let `T` be a `Trial`.
     Let `P` be the process that last updated the `state` attribute of `T`.
     Then, any reads on any attributes of `T` are guaranteed to return the same or
--- a/optuna/storages/base.py
+++ b/optuna/storages/base.py
@@ -52,13 +52,11 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
 
     **Stronger consistency requirements for special data**
 
-    TODO(ytsmiling) Add load method to storage class implementations.
-
     Under a multi-worker setting, a storage class must return the latest values of any attributes
     of a study, not necessarily for the attributes of a `Trial`.
-    However, if the `load(study_id)` method is called, any successive reads on the `state`
-    attribute of a `Trial` are guaranteed to return the same or more recent values than the value
-    at the time of call to the `load` method.
+    However, if the `read_from_remote_storage(study_id)` method is called, any successive reads on
+    the `state` attribute of a `Trial` are guaranteed to return the same or more recent values than
+    the value at the time of call to the `load` method.
     Let `T` be a `Trial`.
     Let `P` be the process that last updated the `state` attribute of `T`.
     Then, any reads on any attributes of `T` are guaranteed to return the same or
@@ -644,6 +642,19 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
                 If no trial with the matching ``trial_id`` exists.
         """
         return self.get_trial(trial_id).system_attrs
+
+    def read_from_remote_storage(self, study_id: int) -> None:
+        """Make an internal cache of trials up-to-date.
+
+        Args:
+            study_id:
+                ID of the study.
+
+        Raises:
+            :exc:`KeyError`:
+                If no study with the matching ``study_id`` exists.
+        """
+        raise NotImplementedError
 
     def remove_session(self) -> None:
         """Clean up all connections to a database."""
--- a/optuna/storages/base.py
+++ b/optuna/storages/base.py
@@ -54,9 +54,10 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
 
     Under a multi-worker setting, a storage class must return the latest values of any attributes
     of a study, not necessarily for the attributes of a `Trial`.
-    However, if the `read_from_remote_storage(study_id)` method is called, any successive reads on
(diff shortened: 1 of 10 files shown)


optuna/storages/rdb/storage.py
```diff
@@ -1009,6 +1009,12 @@ def get_n_trials(self, study_id, state=None):
         self._commit(session)
         return n_trials
 
+    def read_from_remote_storage(self, study_id: int) -> None:
+        # Check that that study exists.
```
HideakiImamura, line 1013: How about changing this comment to the following?
```suggestion
        # Make sure that the given study exists.
```

optuna/storages/rdb/storage.py
```diff
@@ -1009,6 +1009,12 @@ def get_n_trials(self, study_id, state=None):
...
+        session = self.scoped_session()
+        models.StudyModel.find_or_raise_by_id(study_id, session)
```
HideakiImamura, line 1015: It might be helpful to note that `KeyError` is raised if the study for the given `study_id` does not exist as follows.
```suggestion
        # If the study for the ``study_id`` does not exist, `KeyError` is raised.
        models.StudyModel.find_or_raise_by_id(study_id, session)
```

ytsmiling: It's documented in the `BaseStorage` doc and I think we don't need the verbose comment here.

optuna/storages/base.py
```diff
@@ -52,13 +52,11 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
-    TODO(ytsmiling) Add load method to storage class implementations.
-
     Under a multi-worker setting, a storage class must return the latest values of any attributes
     of a study, not necessarily for the attributes of a `Trial`.
-    However, if the `load(study_id)` method is called, any successive reads on the `state`
-    attribute of a `Trial` are guaranteed to return the same or more recent values than the value
-    at the time of call to the `load` method.
+    However, if the `read_from_remote_storage(study_id)` method is called, any successive reads on
+    the `state` attribute of a `Trial` are guaranteed to return the same or more recent values than
+    the value at the time of call to the `load` method.
```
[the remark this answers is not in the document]
ytsmiling: Thanks for the catch!

optuna/storages/cached_storage.py
```diff
@@ -353,9 +353,22 @@ def get_trial(self, trial_id: int) -> FrozenTrial:
...
 
     def get_all_trials(self, study_id: int, deepcopy: bool = True) -> List[FrozenTrial]:
+        if study_id not in self._studies:
+            self.read_from_remote_storage(study_id)
 
         with self._lock:
-            # The cache update will be moved into another method in the future.
+            study = self._studies[study_id]
+            # We need to sort trials by their number because some samplers assume this behavior.
+            # The following two lines are latency-sensitive.
```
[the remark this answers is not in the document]
ytsmiling: It's the latency to execute the two lines, and yes, the latency of these two lines is critical for the overall optimization speed.

optuna/storages/rdb/storage.py
ytsmiling: I adopted the change.

optuna/storages/cached_storage.py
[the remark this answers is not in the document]
HideakiImamura: I got it. Thank you for your reply.

optuna/storages/rdb/storage.py
HideakiImamura: I see. Thanks for your correction.

optuna/storages/base.py
```diff
@@ -645,6 +643,19 @@ def get_trial_system_attrs(self, trial_id: int) -> Dict[str, Any]:
         """
         return self.get_trial(trial_id).system_attrs
 
+    def read_from_remote_storage(self, study_id: int) -> None:
```
sile, line 646: It's not a strong opinion but, personally, I prefer `read_trials_from_remote_storage` to the current name because it sounds a bit more self-descriptive to me.
What do you think of it?

ytsmiling: Thank you for the comment. I think it's a better method-name and I'll adopt the suggestion.

ytsmiling:
@sile Thank you for reviewing this PR. I think we also should change the storage access by pruning implementations and add an additional call of `read_trials_from_remote_storage` at the very beginning of the`should_prune` method call.
However, I' not sure whether it works for all cases, and it's also notable that some concurrent PR (e.g., #1271) might change the pruning implementations. It seems that changing pruning implementations will be a too large change for this PR, and I'd like to create a separate PR or create an issue.

optuna/storages/base.py
sile: Thanks!

sile:
@ytsmiling Thank you for your reply. I agree that the change could be too large to include this PR. Creating another PR or issue seems good!

PR #1299, opened 2020-05-28
Introducing `optuna.visualization.plot_param_importances`.

hvy:
## Motivation

Hyperparameter importances can be assessed via `optuna.importance.get_param_importances`. Users might want to plot these importances for visual inspection which could be simplified by providing a function through Optuna's visualization submodule `optuna.visualization`.

## Description of the changes

- Introduces `optuna.visualization.plot_param_importances` to plot hyperparameter importances as bar charts.
- Adds a script to generate an example figure for the API reference (similar to other visualization functions).
- Adds a generate example figure and renders it in the API reference.

## TODO

 (Done) ~See `TODO`s in the diff.~


tests/visualization_tests/test_param_importances.py
```diff
@@ -0,0 +1,44 @@
...
+
+import pytest
+
+from optuna.study import create_study
+from optuna.testing.visualization import prepare_study_with_trials
+from optuna.trial import Trial
+from optuna.visualization import plot_param_importances
+
+
+def test_plot_param_importances() -> None:
```
hvy, line 11: Basically follows other visualization function unit tests.

optuna/visualization/param_importances.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+
+    _check_plotly_availability()
+
+    layout = go.Layout(
+        title="Hyperparameter Importances",
+        xaxis={"title": "Feature"},
+        yaxis={"title": "Importance"},
+        showlegend=False,
+    )
```
hvy, line 65: Intentionally sticking to default style for consistency with other visualization functions (i.e. intermediate value plot).

optuna/visualization/param_importances.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Importances cannot be evaluated without completed trials.
+    # Return an empty figure for consistency with other visualization functions.
+    trials = [trial for trial in study.trials if trial.state == TrialState.COMPLETE]
+    if len(trials) == 0:
+        logger.warning("Study instance does not contain trials.")
```
HideakiImamura, line 74: It's a minor detail, but how about changing the text to look like this?

```suggestion
        logger.warning("Study instance does not contain completed trials.")
```

optuna/visualization/param_importances.py
```diff
@@ -0,0 +1,83 @@
...
+    _check_plotly_availability()
+
+    layout = go.Layout(
+        title="Hyperparameter Importances",
+        xaxis={"title": "Feature"},
+        yaxis={"title": "Importance"},
+        showlegend=False,
+    )
+
+    # Importances cannot be evaluated without completed trials.
```
HideakiImamura, line 70: This is not in the scope of this PR, but how about evaluating the importance of pruned trials? I think the last intermediate values can be used to evaluate the importance. (needs-discussion)

optuna/visualization/param_importances.py
hvy: Makes sense. Thanks for you suggestion, applied!

hvy:
Thanks for the review, PTAL. I added a reference to the plot function from `get_param_importances` as suggested.

Including pruned trials might be useful especially if trials are expensive to evaluate. Maybe it could be an option. We should to run proper experiments before that though. Actually, looping through all trials to evaluate them in the plot functions isn't the cleanest design, but I did so to keep the behavior consistent with other visualization functions.

optuna/visualization/param_importances.py
```diff
@@ -0,0 +1,83 @@
...
+    # Return an empty figure for consistency with other visualization functions.
+    trials = [trial for trial in study.trials if trial.state == TrialState.COMPLETE]
+    if len(trials) == 0:
+        logger.warning("Study instance does not contain completed trials.")
+        return go.Figure(data=[], layout=layout)
+
+    importances = optuna.importance.get_param_importances(study, params=params)
```
toshihikoyanase, line 77: I think it may be more useful if we can compare the different importance evaluators. Do you think it is complicated to add the `evaluator` argument to `plot_param_importances`?

docs/source/plotly_figures/plot_param_importances.html
```diff
@@ -0,0 +1,20 @@
+<div>
+        
+                <script type="text/javascript">window.PlotlyConfig = {MathJaxConfig: 'local'};</script>
```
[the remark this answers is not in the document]
hvy: Oh you are right. Might be an issue on plotly since this file was generated automatically by it. It's also consistent across other generated files https://github.com/optuna/optuna/tree/master/docs/source/plotly_figures. Do you still want me to modify it?

optuna/visualization/param_importances.py
hvy: Thank you for the suggestion. You are right and it's probably easier from a user perspective if `plot_param_importances` is identical to `get_param_importances`. Fixed!

hvy:
Thank you for the review. Changes are made so PTAL.

docs/source/plotly_figures/plot_param_importances.html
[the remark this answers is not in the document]
toshihikoyanase: > Might be an issue on plotly since this file was generated automatically by it.

Thank you for your explanation. I think we can leave as it is.

PR #1301, opened 2020-05-29
Add link to regression example in the header of `keras_integration.py`.

zishiwu123:
## Motivation
This PR adds a link to `keras_mlflow.py` within the header of the `keras_integration.py` example.
It saves new users some time by allowing them to easily find both a classification and a regression example using Keras within `keras_integration.py`.

## Description of the changes
Only the header documentation of `keras_integration.py` was changed.

### DIFF
--- a/examples/pruning/keras_integration.py
+++ b/examples/pruning/keras_integration.py
@@ -9,6 +9,10 @@ results and stops unpromising trials.
 You can run this example as follows:
     $ python keras_integration.py
 
+For a similar Optuna example that demonstrates Keras without a pruner on a regression dataset,
+see the following link:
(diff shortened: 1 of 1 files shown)


hvy:
Of course not.

It is up to you whether to delete the branch or not since it's your fork. We're fine with either. I'd say that many usually do delete them, but there are some pros/cons to it https://stackoverflow.com/questions/10765321/should-i-delete-a-branch-after-merging-it/10766808. 

Sorry about the confusion with the reference point. I wanted future readers of this PR to be able to navigate to the discussion on Gitter. It has nothing to do with creating feature branches.

PR #1302, opened 2020-05-29
Implement `log` argument for `suggest_int` of pycma integration.

toshihikoyanase:
## Motivation
This one of the follow-up PRs of #1201.

## Description of the changes
This PR adds support for `IntLogUniformDistribution` to `optuna.integration.CmaEsSampler`.

I have two concerns about this implementation:
- Although `optuna.samplers.CmaEsSampler` quantizes the suggested values by `step` in the log domain, this implementation quantizes them in the linear domain similarly to the `optuna.samplers.RandomSampler`. If this change is acceptable, I think I also update the `optuna.samplers.CmaEsSampler`.
- When `optuna.samplers.RandomSampler` suggests values from `DiscreteUniformDistribution`, it expands `low` and `high` by `0.5 * q` to make the bins of quantization equivalent to others. On the other hand, it expands `0.5` when it uses `IntLogUniformDistribution`. I think this is because the `low` can be negative depending on `step`. In the most cases, users will use `step=1`  and the calculation will be correct. So, I follow this definition in `optuna.integration.CmaEsSampler`, but I'm a bit concerned about performance degradation when users specify `step>1`.

### DIFF
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -234,7 +234,7 @@ class PyCmaSampler(BaseSampler):
             elif isinstance(distribution, IntUniformDistribution):
                 x0[name] = int(numpy.mean([distribution.high, distribution.low]))
             elif isinstance(distribution, LogUniformDistribution) or isinstance(
-                dist, IntLogUniformDistribution
+                distribution, IntLogUniformDistribution
             ):
                 log_high = math.log(distribution.high)
                 log_low = math.log(distribution.low)
@@ -261,7 +261,7 @@ class PyCmaSampler(BaseSampler):
             elif isinstance(distribution, IntUniformDistribution):
                 sigma0s.append((distribution.high - distribution.low) / 6)
             elif isinstance(distribution, LogUniformDistribution) or isinstance(
-                dist, IntLogUniformDistribution
+                distribution, IntLogUniformDistribution
             ):
                 log_high = math.log(distribution.high)
                 log_low = math.log(distribution.low)
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -456,8 +456,7 @@ class _Optimizer(object):
             return int(v)
         if isinstance(dist, IntLogUniformDistribution):
             exp_value = math.exp(cma_param_value)
-            r = numpy.round((exp_value - dist.low) / dist.step)
-            v = r * dist.step + dist.low
+            v = numpy.round(exp_value)
             return int(min(max(v, dist.low), dist.high))
 
         if isinstance(dist, CategoricalDistribution):
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -431,7 +431,7 @@ class _Optimizer(object):
         # type: (Dict[str, BaseDistribution], str, Any) -> float
 
         dist = search_space[param_name]
-        if isinstance(dist, LogUniformDistribution) or isinstance(dist, IntLogUniformDistribution):
+        if isinstance(dist, (LogUniformDistribution, IntLogUniformDistribution)):
             return math.log(optuna_param_value)
         elif isinstance(dist, DiscreteUniformDistribution):
             return optuna_param_value - dist.low
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -233,11 +233,7 @@ class PyCmaSampler(BaseSampler):
                 x0[name] = numpy.mean([distribution.high, distribution.low])
             elif isinstance(distribution, IntUniformDistribution):
                 x0[name] = int(numpy.mean([distribution.high, distribution.low]))
-            elif isinstance(distribution, IntLogUniformDistribution):
-                log_high = math.log(distribution.high)
-                log_low = math.log(distribution.low)
-                x0[name] = math.exp(numpy.mean([log_high, log_low]))
-            elif isinstance(distribution, LogUniformDistribution):
+            elif isinstance(distribution, LogUniformDistribution) or isinstance(dist, IntLogUniformDistribution):
                 log_high = math.log(distribution.high)
                 log_low = math.log(distribution.low)
                 x0[name] = math.exp(numpy.mean([log_high, log_low]))
@@ -262,11 +258,7 @@ class PyCmaSampler(BaseSampler):
                 sigma0s.append((distribution.high - distribution.low) / 6)
             elif isinstance(distribution, IntUniformDistribution):
                 sigma0s.append((distribution.high - distribution.low) / 6)
-            elif isinstance(distribution, IntLogUniformDistribution):
-                log_high = math.log(distribution.high)
-                log_low = math.log(distribution.low)
-                sigma0s.append((log_high - log_low) / 6)
-            elif isinstance(distribution, LogUniformDistribution):
+            elif isinstance(distribution, LogUniformDistribution) or isinstance(dist, IntLogUniformDistribution):
                 log_high = math.log(distribution.high)
                 log_low = math.log(distribution.low)
                 sigma0s.append((log_high - log_low) / 6)
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -233,9 +233,7 @@ class PyCmaSampler(BaseSampler):
                 x0[name] = numpy.mean([distribution.high, distribution.low])
             elif isinstance(distribution, IntUniformDistribution):
                 x0[name] = int(numpy.mean([distribution.high, distribution.low]))
-            elif isinstance(distribution, LogUniformDistribution) or isinstance(
-                distribution, IntLogUniformDistribution
-            ):
+            elif isinstance(distribution, (LogUniformDistribution, IntLogUniformDistribution)):
                 log_high = math.log(distribution.high)
                 log_low = math.log(distribution.low)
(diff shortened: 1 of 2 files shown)


optuna/integration/cma.py
```diff
@@ -312,8 +319,12 @@ def __init__(
...
                 highs.append(r + 0.5 * dist.q)
             elif isinstance(dist, IntUniformDistribution):
-                lows.append(dist.low - 0.5)
-                highs.append(dist.high + 0.5)
+                lows.append(dist.low - 0.5 * dist.step)
+                highs.append(dist.high + 0.5 * dist.step)
+            elif isinstance(dist, IntLogUniformDistribution):
+                # TODO(toshihikoyanase): Shiting 0.5 is not sufficient if step > 1.
+                lows.append(self._to_cma_params(search_space, param_name, dist.low - 0.5))
+                highs.append(self._to_cma_params(search_space, param_name, dist.high + 0.5))
```
HideakiImamura, line 327: Your second concern is reasonable: instead of expanding 0.5, we should expand 0.5*step. As you say, I think that `IntLogUniformDistrbituion` uses 0.5 to ensure `low` positive, but if we just want to ensure it, how about doing the following? I think the `optuna.samplers.CmaEsSampler` can be fixed as well.

```suggestion
                # TODO(toshihikoyanase): Shiting 0.5 is not sufficient if step > 1.
                lows.append(self._to_cma_params(search_space, param_name, max(0, dist.low - 0.5 * dist.step)))
                highs.append(self._to_cma_params(search_space, param_name, dist.high + 0.5 * dist.step))
```

HideakiImamura: This change still has the problem of bias in discretization when `low - 0.5 * step` is negative. A possible solution is to move `[low-0.5*step, high+0.5*step]` parallel to `[1, high - low + step + 1]` for sampling when `low - 0.5 * step` is negative, and then undo it later.  However, this is a change that has to be made to other samplers at the same time, and I think it's outside the scope of this PR.

optuna/integration/cma.py
```diff
@@ -442,7 +453,13 @@ def _to_optuna_params(search_space, param_name, cma_param_value):
         if isinstance(dist, IntUniformDistribution):
             r = numpy.round((cma_param_value - dist.low) / dist.step)
             v = r * dist.step + dist.low
-            return v
+            return int(v)
+        if isinstance(dist, IntLogUniformDistribution):
+            exp_value = math.exp(cma_param_value)
+            r = numpy.round((exp_value - dist.low) / dist.step)
+            v = r * dist.step + dist.low
+            return int(v)
```
HideakiImamura, line 461: How about limiting the value of `v` to the range of `[low, high]` as in `DiscreteUniformDistribution`?
```suggestion
            return int(min(max(v, dist.low), dist.high))
```

optuna/integration/cma.py
```diff
@@ -312,8 +319,12 @@ def __init__(
                 lows.append(0 - 0.5 * dist.q)
                 highs.append(r + 0.5 * dist.q)
             elif isinstance(dist, IntUniformDistribution):
-                lows.append(dist.low - 0.5)
-                highs.append(dist.high + 0.5)
+                lows.append(dist.low - 0.5 * dist.step)
+                highs.append(dist.high + 0.5 * dist.step)
+            elif isinstance(dist, IntLogUniformDistribution):
+                # TODO(toshihikoyanase): Shiting 0.5 is not sufficient if step > 1.
```
nzw0301, line 325: I suppose `shifting` rather than `shiting`?

toshihikoyanase: Oops! Thank you for pointing it out. Fixed: https://github.com/optuna/optuna/pull/1302/commits/68f32c549f21e361d79174cd7d5a12ea0550ac6b

optuna/integration/cma.py
toshihikoyanase: I'm sorry for the delayed response.
Your new PR (#1329) will prohibit the simultaneous use of `step` and `log`, and we can keep the `master`'s implementation. So, I'll revert the change.

optuna/integration/cma.py
toshihikoyanase: Thank you! Applied in https://github.com/optuna/optuna/pull/1302/commits/25f19df463dae7ae04f8989bf7039d60b1bd9a6c.

HideakiImamura:
Hi @toshihikoyanase! Is this PR ready for the review?

toshihikoyanase:
@HideakiImamura I'm sorry for the delayed response. Regarding the `step` of `IntLogUniformDistribution`, #1329 ensures that the `step` is always `1`, and I set the range to `[low-0.5, high+0.5]` similar to the other samplers like `SkoptSampler`.
Please take another look.

optuna/integration/cma.py
```diff
@@ -226,13 +227,16 @@ def _initialize_x0(search_space):
...
                 x0[name] = numpy.mean([distribution.high, distribution.low])
             elif isinstance(distribution, DiscreteUniformDistribution):
                 x0[name] = numpy.mean([distribution.high, distribution.low])
             elif isinstance(distribution, IntUniformDistribution):
                 x0[name] = int(numpy.mean([distribution.high, distribution.low]))
+            elif isinstance(distribution, IntLogUniformDistribution):
+                log_high = math.log(distribution.high)
+                log_low = math.log(distribution.low)
+                x0[name] = math.exp(numpy.mean([log_high, log_low]))
             elif isinstance(distribution, LogUniformDistribution):
```
HideakiImamura, line 240: This logic appears to be redundant with the `LogUniformDistribution` case. We can remove its redundancy as follows.

```suggestion
            elif isinstance(distribution, LogUniformDistribution) or isinstance(dist, IntLogUniformDistribution):
```

optuna/integration/cma.py
```diff
@@ -252,13 +256,16 @@ def _initialize_sigma0(search_space):
...
                 sigma0s.append((distribution.high - distribution.low) / 6)
             elif isinstance(distribution, DiscreteUniformDistribution):
                 sigma0s.append((distribution.high - distribution.low) / 6)
             elif isinstance(distribution, IntUniformDistribution):
                 sigma0s.append((distribution.high - distribution.low) / 6)
+            elif isinstance(distribution, IntLogUniformDistribution):
+                log_high = math.log(distribution.high)
+                log_low = math.log(distribution.low)
+                sigma0s.append((log_high - log_low) / 6)
             elif isinstance(distribution, LogUniformDistribution):
```
HideakiImamura, line 269: Ditto.

```suggestion
            elif isinstance(distribution, LogUniformDistribution) or isinstance(dist, IntLogUniformDistribution):
```

optuna/integration/cma.py
```diff
@@ -425,7 +431,7 @@ def _to_cma_params(search_space, param_name, optuna_param_value):
         # type: (Dict[str, BaseDistribution], str, Any) -> float
 
         dist = search_space[param_name]
-        if isinstance(dist, LogUniformDistribution):
+        if isinstance(dist, LogUniformDistribution) or isinstance(dist, IntLogUniformDistribution):
```
hvy, line 434: A nit.
```suggestion
        if isinstance(dist, (LogUniformDistribution, IntLogUniformDistribution)):
```

optuna/integration/cma.py
```diff
@@ -226,14 +227,15 @@ def _initialize_x0(search_space):
...
             if isinstance(distribution, UniformDistribution):
                 x0[name] = numpy.mean([distribution.high, distribution.low])
             elif isinstance(distribution, DiscreteUniformDistribution):
                 x0[name] = numpy.mean([distribution.high, distribution.low])
             elif isinstance(distribution, IntUniformDistribution):
                 x0[name] = int(numpy.mean([distribution.high, distribution.low]))
-            elif isinstance(distribution, LogUniformDistribution):
+            elif isinstance(distribution, LogUniformDistribution) or isinstance(
+                distribution, IntLogUniformDistribution
+            ):
```
hvy, line 238: Ditto.

toshihikoyanase:
I'm sorry but I pushed the wrong branch. I'm reverting the change.

optuna/integration/cma.py
```diff
@@ -317,8 +316,11 @@ def __init__(
                 lows.append(0 - 0.5 * dist.q)
                 highs.append(r + 0.5 * dist.q)
             elif isinstance(dist, IntUniformDistribution):
-                lows.append(dist.low - 0.5)
-                highs.append(dist.high + 0.5)
+                lows.append(dist.low - 0.5 * dist.step)
+                highs.append(dist.high + 0.5 * dist.step)
```
hvy, line 320: Just wondering but is this change somewhat orthogonal to the other?

toshihikoyanase: You're right. It's the change for `IntUniformDistribution`.  I reverted the change in https://github.com/optuna/optuna/pull/1302/commits/682e8f4fac1f5e7a22af99cb93957041681815e9. Thank you for your careful review.

hvy: Should we create a separate PR for that?

toshihikoyanase: Thank you for your suggestion.
I created #1456. Please review it after this PR.

hvy: Thanks, I'll have a look at it.

PR #1305, opened 2020-05-29
Add experimental decorator to `LightGBMTuner` and `LightGBMTunerCV`.

toshihikoyanase:
## Motivation
`LightGBMTuner` and `LightGBMTunerCV` will be public to users in `v1.5.0`, and I think it would be safe to mark them as `experimental` similarly to `optuna.integration.lightgbm_tuner.train`.

## Description of the changes
Add `experimental` decorator to `LightGBMTuner` and `LightGBMTunerCV`.

### DIFF
--- a/optuna/integration/lightgbm_tuner/optimize.py
+++ b/optuna/integration/lightgbm_tuner/optimize.py
@@ -891,7 +891,7 @@ class LightGBMTuner(LightGBMBaseTuner):
         )
 
 
-@experimental("1.15.0")
+@experimental("1.5.0")
 class LightGBMTunerCV(LightGBMBaseTuner):
     """Hyperparameter tuner for LightGBM with cross-validation.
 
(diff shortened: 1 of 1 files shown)


optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -660,6 +661,7 @@ def best_trial(self) -> optuna.trial.FrozenTrial:
         return _StepwiseStudy(study, step_name)
 
 
+@experimental("1.15.0")
```
hvy, line 664:
```suggestion
@experimental("1.5.0")
```

optuna/integration/lightgbm_tuner/optimize.py
```diff
@@ -889,6 +891,7 @@ def _create_objective(
         )
 
 
+@experimental("1.15.0")
```
hvy, line 894:
```suggestion
@experimental("1.5.0")
```

toshihikoyanase:
@hvy You're absolutely right. Thank you for your careful review.

sile:
Thanks for the catch 🙇  > @hvy

PR #1321, opened 2020-06-03
Exempt issues with `no-stale` labelel from stale bot.

crcrpar:
**Before**

Exempt: bug,contribution-welcome

**After**

Exempt: bug, no-stale

Currently stale bot marks release tracker issues that theoretically not marked because we don't have an appropriate label which is exempt from the bot. So I've created a new label of "no-stale" and this PR updates the workflow and removes `bug` and `contribution-welcome`.

 Upon this PR gets merged, I'd love developers with write access to handle `bug`/`contribution-welcome` issues as follows
1. Attach both `bug`/`contribution-welcome` and `no-stale`.
2. Once someone (or you) is decided to tackle it, remove `no-stale`. This is because `stale` bot would help you and contributor stay tuned.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -20,4 +20,4 @@ jobs:
         days-before-close: -1
         stale-issue-label: 'stale'
         stale-pr-label: 'stale'
-        exempt-issue-labels: 'bug,no-stale'
+        exempt-issue-labels: 'no-stale'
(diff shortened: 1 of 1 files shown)


crcrpar:
Thanks for the review.

Before the (re)labeling operation, we need to decide what to do if an issue has both `bug` and `contribution-welcome`. Will we label it `no-stale`?

crcrpar:
The consensus of the discussion with reviewers: we gonna remove `bug` label also. We believe this decision would make the new operation simpler.

hvy:
Let me merge this PR as it has two approvals. 

We went over all existing tickets labeled `contribution-welcome` and `bug` and put `no-stale` labels where they were suitable. From now on, we should in general consider this `no-stale` label for each ticket separately. For instance, issues that have ongoing PRs are candidates for `no-stale`.

Issue #1320, opened 2020-06-03
Dashboard to html option producing a bokeh error.


apiszcz:
Optuna 1.5.0, Python 3.7.7

`optuna dashboard  --storage "postgres://postgres@localhost:5444/optuna" --study-name test --out optunadashboard.html`

```
[W 2020-06-03 06:30:19,131] Optuna dashboard is still highly experimental. Please use with caution!
usage: bokeh [-h] [-v]
             {build,info,init,json,sampledata,secret,serve,static} ...
bokeh: error: invalid choice: 'html' (choose from 'build', 'info', 'init', 'json', 'sampledata', 'secret', 'serve', 'static')
```

toshihikoyanase:
I confirmed that this error also occurred with `bokeh>=2.0.0`. I think the cause is the same as https://github.com/optuna/optuna/issues/1319#issuecomment-638780661. Please downgrade `bokeh` to `1.4.0`.

toshihikoyanase:
#1326 has been merged, and let me close this issue.

Issue #1319, opened 2020-06-03
Optuna dashboard page shows up blank on localhost.


apiszcz:
Optuna 1.5.0, Python 3.7.7

The following command
`optuna dashboard  --storage "postgres://postgres@localhost:5444/optuna" --study-name test
`the http server starts up and the web page opens to a blank page. Refresh the page produces another blank page. Any ideas on where I should look to resolve the issue?

Thank you for a great package and documentation!

```
[W 2020-06-03 06:27:42,079] Optuna dashboard is still highly experimental. Please use with caution!
[I 2020-06-03 06:27:42,087] Starting Bokeh server version 2.0.2 (running on Tornado 6.0.4)
[I 2020-06-03 06:27:42,089] User authentication hooks NOT provided (default user enabled)
[I 2020-06-03 06:27:42,092] Bokeh app running at: http://localhost:5006/dashboard
[I 2020-06-03 06:27:42,093] Starting Bokeh server with process id: 31360
[I 2020-06-03 06:27:43,014] 200 GET /dashboard (::1) 117.69ms
[I 2020-06-03 06:27:43,670] 200 GET /static/js/bokeh.min.js?v=57d29d5936e494351385d736f792154c (::1) 612.41ms
[I 2020-06-03 06:27:43,684] 200 GET /static/js/bokeh-widgets.min.js?v=fdb1f42cceaade9062de57963cac654c (::1) 11.93ms
[I 2020-06-03 06:27:43,697] 200 GET /static/js/bokeh-tables.min.js?v=0a7f930d4b0805283dfba427bb3d0dae (::1) 24.89ms
[I 2020-06-03 06:27:43,702] 200 GET /static/js/bokeh-gl.min.js?v=c376f0736800259be44cb1aed39a6cc3 (::1) 28.89ms
[I 2020-06-03 06:27:43,787] 101 GET /dashboard/ws (::1) 0.00ms
[I 2020-06-03 06:27:43,788] WebSocket connection opened
[I 2020-06-03 06:27:43,790] ServerConnection created
[W 2020-06-03 06:27:43,798] 404 GET /favicon.ico (::1) 1.00ms
[I 2020-06-03 06:28:38,374] 200 GET /dashboard (::1) 3.96ms
[I 2020-06-03 06:28:38,381] WebSocket connection closed: code=1001, reason=None
[I 2020-06-03 06:28:38,510] 101 GET /dashboard/ws (::1) 1.00ms
[...]

PR #1315, opened 2020-06-02
Introduces `optuna._imports.try_import` to DRY optional imports.

hvy:
## Motivation

DRY. Optional imports are currently associated with code duplication.

## Description of the changes

Introduces `optuna._imports` and `optuna._imports.try_import` to DRY optional imports. Note that deferred error messages are somewhat simplified/generalized.

## Note

Other components such as `optuna._LazyImport` and `optuna.integration._IntegrationModule` could probably be moved into the introduced `optuna._imports` in follow up work.

PR #1326, opened 2020-06-04
Raise `ImportError` if `bokeh` version is 2.0.0 or newer.

toshihikoyanase:
## Motivation
This PR addresses #1319 and #1320 

The current dashboard is not compatible with `bokeh>=2.0.0`, and it may not work correctly without showing any error messages.

## Description of the changes

This PR adds version checking of `bokeh`. Note that this modification will conflict with #1315 , please merge it in advance.

### DIFF
--- a/optuna/dashboard.py
+++ b/optuna/dashboard.py
@@ -27,6 +27,20 @@ with try_import() as _imports:
     import bokeh.themes
     import tornado.gen
 
+if _imports.is_successful():
+    from distutils.version import StrictVersion
+
+    from bokeh import __version__ as bokeh_version
+
+    if StrictVersion(bokeh_version) >= StrictVersion("2.0.0"):
+        raise ImportError(
+            "Your version of bokeh is " + bokeh_version + " . "
+            "Please install bokeh version earlier than 2.0.0. "
+            "Bokeh can be installed by executing `$ pip install 'bokeh<2.0.0'`. "
+            "For further information, please refer to the installation guide of bokeh. "
+        )
+
+
 _mode = None  # type: Optional[str]
 _study = None  # type: Optional[optuna.study.Study]
 
--- a/optuna/dashboard.py
+++ b/optuna/dashboard.py
@@ -1,4 +1,5 @@
 import collections
+from distutils.version import StrictVersion
 import threading
 import time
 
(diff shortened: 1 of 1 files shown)


hvy:
Could you resolve the conflict now that `try_import` is merged? I guess we should add this check ad the bottom of https://github.com/optuna/optuna/blob/master/optuna/dashboard.py#L20, similar to how it's done in https://github.com/optuna/optuna/blob/master/optuna/visualization/plotly_imports.py.

optuna/dashboard.py
```diff
@@ -27,6 +27,20 @@
     import bokeh.themes
     import tornado.gen
 
+if _imports.is_successful():
```
hvy, line 30: Just wondering but how come we're deferring the `bokeh` imports above but not the version check here?

toshihikoyanase: That was because `ImportError.name` was `None`, and the error message was not shown as expected. I input the `name` argument and move the version check into with-try_import clause in https://github.com/optuna/optuna/pull/1326/commits/1b7dcab2046e7f00dcf335840c7c9f00d23a3f9a.

```console
% optuna dashboard --study-name parallel --storage sqlite:///lgbtuner.db
[E 2020-06-08 15:55:34,522] Tried to import 'None' but failed. Please make sure that the package is installed correctly to use this feature. Actual error: Your version of bokeh is 2.0.2 . Please install bokeh version earlier than 2.0.0. Bokeh can be installed by executing `$ pip install 'bokeh<2.0.0'`. For further information, please refer to the installation guide of bokeh. .
```

hvy: Ah, thanks for the explanation and your suggestion for fix. That totally slipped through my mind. Let me create a fix that for the similar `plotly` version check.

optuna/dashboard.py
```diff
@@ -18,6 +18,9 @@
     from typing import Optional  # NOQA
 
 with try_import() as _imports:
+    from distutils.version import StrictVersion
```
hvy, line 21: How about importing this at the top, at the top indentation level, to not by mistake catch errors related to `StrictVersion`?

toshihikoyanase: Thank you. Fixed in https://github.com/optuna/optuna/pull/1326/commits/92ad990ff420f7428b4792a2a38912aa55d669d2.

hvy:
Should we label this `enhancement` ? Seems like a non-trial change to the user considering the issues referenced in the PR description.

toshihikoyanase:
> Should we label this `enhancement` ?

I agree with you. The version check is newly introduced by this PR and it may be worth notifying users.

Issue #1308, opened 2020-05-31
Unnecessary return value of `BaseStorage.set_trial_param`.


ytsmiling:
## TL;DR
This issue requests to deprecate the return value of `BaseStorage.set_trial_param` and change the behavior to "always overwrite existing values".

## General background
[`BaseStorage.set_trial_param`](https://github.com/optuna/optuna/blob/24300dc57661bff7961499a2d52534fcd41ad258/optuna/storages/base.py#L120) method returns True and set a new parameter when the specified parameter-name does not already have a value. Otherwise, it returns False and does not overwrite the parameter with the new value.
This behavior was necessary to prevent race-conditioning. However, the current optuna doesn't allow users to directly manipulate the storage and optuna internals do not modify parameters from multiple workers/threads, and thus, the functionality is no more required.

## Proposal and motivation
This issue requests to deprecate the return value of `BaseStorage.set_trial_param` and change its behavior to "always overwrite existing values".
Instead of using the return value, we'd better check whether the parameter already exists first by ``param_name in storage.get_trial_params(trial_id)`` and then perform sampling operation in [this method in trial](https://github.com/optuna/optuna/blob/24300dc57661bff7961499a2d52534fcd41ad258/optuna/trial.py#L981)
This change has two benefits.
1. We can remove unnecessary computation for re-sampling when the parameter value already exists.
2. The storage method becomes more consistent with other methods such as `set_trial_user_attrs`, which always overwrite existing values. It also simplifies some storage implemetations.

PR #1327, opened 2020-06-05
`BaseStorage.set_trial_param` to return `None` instead of `bool`.

hvy:
## Motivation

See https://github.com/optuna/optuna/issues/1308.

## Description of the changes

Fixes https://github.com/optuna/optuna/issues/1308 and the following.
- The storage always sets, i.e the distribution compatibility check is removed. The check is performed by the trial.
- Simplified `Trial` by removing `_set_new_param_or_get_existing`.

### DIFF
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -215,9 +215,6 @@ class Trial(BaseTrial):
 
         self._check_distribution(name, distribution)
 
-        if distribution.low == distribution.high:
-            return self._set_new_param_or_get_existing(name, distribution.low, distribution)
-
         return self._suggest(name, distribution)
 
     def suggest_loguniform(self, name, low, high):
@@ -270,9 +267,6 @@ class Trial(BaseTrial):
         distribution = LogUniformDistribution(low=low, high=high)
 
         self._check_distribution(name, distribution)
-
-        if distribution.low == distribution.high:
-            return self._set_new_param_or_get_existing(name, distribution.low, distribution)
 
         return self._suggest(name, distribution)
 
@@ -333,9 +327,6 @@ class Trial(BaseTrial):
         distribution = DiscreteUniformDistribution(low=low, high=high, q=q)
 
         self._check_distribution(name, distribution)
-
-        if distribution.low == distribution.high:
-            return self._set_new_param_or_get_existing(name, distribution.low, distribution)
 
         return self._suggest(name, distribution)
 
@@ -408,9 +399,6 @@ class Trial(BaseTrial):
             distribution = IntUniformDistribution(low=low, high=high, step=step)
 
         self._check_distribution(name, distribution)
-
-        if distribution.low == distribution.high:
-            return self._set_new_param_or_get_existing(name, distribution.low, distribution)
 
         return int(self._suggest(name, distribution))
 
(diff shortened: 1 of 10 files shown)


tests/test_trial.py
```diff
@@ -210,37 +210,24 @@ def test_suggest_low_equals_high(storage_init_func):
...
+    assert trial.suggest_float("e", 1.0, 1.0) == 1.0  # Suggesting a param.
+    assert trial.suggest_float("e", 1.0, 1.0) == 1.0  # Suggesting the same param.
+    assert trial.suggest_float("f", 0.5, 0.5, log=True) == 0.5  # Suggesting a param.
+    assert trial.suggest_float("f", 0.5, 0.5, log=True) == 0.5  # Suggesting the same param.
+    assert trial.suggest_float("g", 0.5, 0.5, log=False) == 0.5  # Suggesting a param.
+    assert trial.suggest_float("g", 0.5, 0.5, log=False) == 0.5  # Suggesting the same param.
+    assert trial.suggest_float("h", 0.5, 0.5, step=1.0) == 0.5  # Suggesting a param.
+    assert trial.suggest_float("h", 0.5, 0.5, step=1.0) == 0.5  # Suggesting the same param.
+    assert trial.suggest_int("i", 1, 1, log=True) == 1  # Suggesting a param.
+    assert trial.suggest_int("i", 1, 1, log=True) == 1  # Suggesting the same param.
```
hvy, line 230: Maybe we can remove this test function altogether.

hvy:
You are right, we should probably not change this behavior. Do you think we should put the check back to the storage? IMO, we can move this check to the trial to simplify the storage logic since it's possible with the current storage API. However, it might be suboptimal and we'll probably want some other storage API to optimize this check in follow-ups.

hvy:
Let me fix the code so that the logic isn't altered for the case you described. I'll add tests on the trial layer as well.

hvy:
PTAL. I moved the distribution compatibility back to the storage from the trial. The reason being that we cannot keep the same behavior explained by @ytsmiling above otherwise. If we try to check for compatibility in the trial, we cannot guarantee the expected behavior in a distributed environment where all initial trials sample the same parameter (having the same name) but with different distributions as they're exposed to race condition when writing the sampled parameter back to the storage.

tests/test_trial.py
ytsmiling: I found it useful to make sure that `sampler.sample_independent` is not called. If you don't want to add the check, I think you can remove this test function.
If you remove this test, I'm not sure if there is a test that calls the `distributions._get_single_value ()` method. I want to make sure that tests for the method are added when you remove this test.

hvy: Right, thanks for you suggestion. I changed this test to check the number of calls to `distributions._get_single_value`. What do you think?

hvy:
Right, thanks for the suggestion about the test. I changes it to count the number of calls to `distributions._get_single_value`. What do you think? It's not the most beautiful code but I think it's acceptable for this test. 

PTAL.

optuna/trial/_trial.py
```diff
@@ -638,29 +626,30 @@ def set_system_attr(self, key, value):
...
-        # type: (str, Any, BaseDistribution) -> Any
+        if name in trial.distributions:
+            # No need to sample if already suggested.
+            distributions.check_distribution_compatibility(trial.distributions[name], distribution)
+            param_value = distribution.to_external_repr(storage.get_trial_param(trial_id, name))
+        else:
+            if self._is_fixed_param(name, distribution):
+                param_value = storage.get_trial_system_attrs(trial_id)["fixed_params"][name]
+            elif distribution.single():
+                param_value = distributions._get_single_value(distribution)
```
toshihikoyanase, line 642: This is optional, but how about adding a test case to raise warnings if single-value distributions are given with fixed params that are out of ranges? This code correctly handles such a case, and I propose it for future development.

hvy: Thank you for the suggestion. I added tests for `enqueue_trial` since it's relevant (although a bit stretched) to this PR.

hvy:
PTAL.

PR #1335, opened 2020-06-07
`optuna.create_trial` and `Study.add_trial` to create custom studies.

hvy:
## Motivation

Addresses https://github.com/optuna/optuna/issues/821. This is a naive but a quite flexible alternative to #1029, mainly to continue the discussion on #1029.

## Description of the changes

C.f. https://github.com/optuna/optuna/pull/1029#issuecomment-633846299. In short, introduces the following as experimental
- `optuna.create_trial` to allows users to create arbitrary trials
- `Study.add_study`, `Study._append_trial` made public, to add trials to an existing (most likely an not yet optimized) study

The two APIs might better be introduced in separate PRs but they make more sense together. The former is not needed if #821 is solely about importing trials from an existing study to another.


optuna/study.py
```diff
@@ -580,51 +581,67 @@ def objective(trial):
...
+                    x = trial.suggest_uniform('x', 0, 10)
+                    return x ** 2
+
+                study = optuna.create_study()
+                assert len(study.trials) == 0
+
+                trial = optuna.create_trial(
+                    params={"x": 1.0},
+                    distributions={"x": UniformDistribution(0, 10)},
+                    value=5.0,
```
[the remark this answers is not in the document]
hvy: Thanks, you're right. I changed to a somewhat less trivial example (with `value=4.0`)

optuna/study.py
```diff
@@ -580,51 +581,67 @@ def objective(trial):
...
-            intermediate_values=intermediate_values,
+        self.add_trial(
+            create_trial(state=TrialState.WAITING, system_attrs={"fixed_params": params})
         )
 
+    @experimental("2.0.0")
+    def add_trial(self, trial: FrozenTrial) -> None:
+        """Add trial to study.
+
+        The trial is validated before being added.
```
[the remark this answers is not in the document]
hvy: I see your point. On the other hand, it should be clear from the type hint and I think this way of putting it is consistent across Optuna. Please correct me otherwise.

optuna/trial/_frozen.py
```diff
@@ -229,3 +230,79 @@ def duration(self):
...
+        datetime_start=datetime_start,
+        datetime_complete=datetime_complete,
+        params=params,
+        distributions=distributions,
+        user_attrs=user_attrs,
+        system_attrs=system_attrs,
+        intermediate_values=intermediate_values,
+    )
+
+    trial._validate()
```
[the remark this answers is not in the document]
hvy: As both APIs are public and not necessarily called together, I think it's important to validate in both. For instance, a user could modify an existing trial and pass it to `add_trial`, in which case no validation would take place.

optuna/study.py
```diff
@@ -580,51 +581,67 @@ def objective(trial):
...
+            :exc:`ValueError`:
+                If trial is an invalid state.
+
+        """
+
         trial._validate()
 
-        trial_id = self._storage.create_new_trial(self._study_id, template_trial=trial)
-        return trial_id
+        self._storage.create_new_trial(self._study_id, template_trial=trial)
```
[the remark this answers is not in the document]
hvy: Hm right. I cannot recall if we tried to foresee some usage of this method where this return value was required (given it's not used for the time being). I think we can reintroduce this method if needed in the future however.

hvy:
Thanks @HideakiImamura for taking a look at this PR. I tried addressing your comments above so PTAL.

optuna/study.py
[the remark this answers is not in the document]
HideakiImamura: Thanks for the explanation. I've understood it. We don't need to note it. Thanks!

HideakiImamura:
Thanks for the update! LGTM, but is this PR WIP now? I think it's OK to merge this features.

hvy:
Thanks for your positive feedback. As it's an experimental feature, I also think it's okay to start with this interface and gather feedback for improvements. I'm a bit concerned with the usability of `create_trial` for a general user, since the parameters are low level. I'd like to add some tests however. Let me address that before merging.

tests/test_trial.py
```diff
@@ -890,3 +891,37 @@ def test_frozen_trial_repr():
     )
 
     assert trial == eval(repr(trial))
+
+
+# TODO(hvy): Write exhaustive test include invalid combinations when feature is no longer
+# experimental.
+@pytest.mark.parametrize("state", [None, TrialState.COMPLETE, TrialState.FAIL])
+def test_create_trial(state: TrialState) -> None:
+    state = TrialState.COMPLETE
```
HideakiImamura, line 900: I think this like should be removed.
```suggestion
```

hvy: Thanks, fixed the test.

hvy:
Thanks for the review, comments should be addressed. PTAL.

optuna/trial/_frozen.py
```diff
@@ -199,3 +199,95 @@ def duration(self):
...
+@experimental("2.0.0")
+def create_trial(
+    *,
+    state: Optional[TrialState] = None,
+    value: Optional[float] = None,
+    params: Optional[Dict[str, Any]] = None,
+    distributions: Optional[Dict[str, BaseDistribution]] = None,
+    intermediate_values: Optional[Dict[int, float]] = None,
+    user_attrs: Optional[Dict[str, Any]] = None,
+    system_attrs: Optional[Dict[str, Any]] = None
```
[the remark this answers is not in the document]
hvy: Makes sense. I applied your suggestion along with https://github.com/optuna/optuna/pull/1335#discussion_r448087566.

HideakiImamura:
@hvy Could you resolve a conflict?

hvy:
Thanks for the heads up. Fixed and PTAL.

toshihikoyanase:
I just started reviewing this PR. Let me share comments about the documentation.

1) Could you add `add_study` to the list of `Study` methods?

image

2) How about adding an internal link to `FrozenTrial` in the description of `create_trial`. I think it is consistent with `create_study`.

image

hvy:
Thanks for the review.

1. Do you mean `add_trial`? Maybe I'm missing your point but it's there as a method of the `Study` class https://47940-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/generated/optuna.study.Study.html#optuna.study.Study.add_trial.

2. Thanks for the suggestion. That makes sense as it's there in the same table. aligned with `create_study`.

PTAL.

c-bata:
I guess the logic of #1426 will be broken if this feature is accepted.

optuna/trial/_frozen.py
```diff
@@ -199,3 +199,95 @@ def duration(self):
             return self.datetime_complete - self.datetime_start
         else:
             return None
+
+
+@experimental("2.0.0")
+def create_trial(
```
c-bata, line 205: I prefer more concrete function name like `create_frozen_trial()` because someone might add `create_fixed_trial()` or something similar function in the future.

optuna/study.py
```diff
@@ -534,51 +535,67 @@ def objective(trial):
...
-            intermediate_values=intermediate_values,
+        self.add_trial(
+            create_trial(state=TrialState.WAITING, system_attrs={"fixed_params": params})
         )
 
+    @experimental("2.0.0")
+    def add_trial(self, trial: FrozenTrial) -> None:
+        """Add trial to study.
+
+        The trial is validated before being added.
```
c-bata, line 546: [IMO] `trial._validate()` is not enough for this feature (ex: this api accepts running frozen trial).

optuna/study.py
```diff
@@ -534,51 +535,67 @@ def objective(trial):
...
-            distributions=distributions,
-            user_attrs=user_attrs,
-            system_attrs=system_attrs,
-            intermediate_values=intermediate_values,
+        self.add_trial(
+            create_trial(state=TrialState.WAITING, system_attrs={"fixed_params": params})
         )
 
+    @experimental("2.0.0")
+    def add_trial(self, trial: FrozenTrial) -> None:
```
c-bata, line 543: Hmm... I'm wondering that most Optuna user cannot distinguish the difference between `enqueue_trial()` and `add_trial()`. But I have no better idea now.

toshihikoyanase:
@hvy Oops, I mentioned `add_trial` of `Study`. It is not listed in the table of the `Study` method.

https://48021-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/study.html

I guess we need to add the following line to `docs/source/reference/study.rst`.

```python
   optuna.study.add_trial
```

optuna/trial/_frozen.py
toshihikoyanase: I agree with you in terms of a different viewpoint. The names of `optuna.create_trial` and `optuna.create_study` seem to be symmetrical, but `optuna.create_trial` is rarely used compared with `optuna.create_study`. So, I feel `optuna.create_trial` might be too general, and I also prefer the name like `create_frozen_trial()`.

toshihikoyanase: I may be too anxious, but some users may have trouble when they use `create_trial` and `add_trial` as an alternative to `study.optimize`. 

After emerging this PR, we can write an iteration of `study.optimize` using public API as follows:

```python
trial = optuna.create_trial( ... )
trial.value = objective(trial)
study.add_trial(trial)
```

This approach is more flexible than `study.optimize` but users need to take care of exceptions including `TrialPruned` and the trial state management by themselves. So, it may have many potholes and I think we cannot recommend to novice users.

I think providing the `optuna.create_trial` and `Study.add_trial` provides many merits to users, but we may carefully guide the users.

The followings are just examples:

- Provide `create_trial` as a function of `optuna.trial` module: `optuna.trial.create_trial()`
- Add tutorial for `create_trial` and `add_trial`

optuna/study.py
hvy: You have a point. We added a `.. seelalso::` forwarding to the user to `enqueue_trial` explaining different use cases. What do you think?

optuna/study.py
hvy: We discussed this a bit and it's likely controversial and you point out. How about starting this experimental feature with little stress without raising errors, allowing users to basically take all trials from one study and inserting it to another one? Running trials may already occur with e.g. keyboard interrupts and shouldn't affect the optimization.

c-bata: OK. I make this conversation resolved.

optuna/trial/_frozen.py
hvy: Thanks for the important feedback. It's now explicitly mentioned that this is not just experimental but also a low-level API. And since it's not mirroring `create_study`, it's not moved to `optuna.trial` instead of `optuna`.

As for the name, it's kept as is. I believe we can be more specific of what type of trial we are talking about, when we really need to distinguish them. In current Optuna, we usually don't in the user facing API (e.g. "get_trial").

hvy:
Merged with master to fix CI failure.

hvy:
Thanks for the many careful reviews. We just had a mob programming session since this is a controversial introduction. I believe it's important that it's marked as experimental and that we can follow up this PR with improvements based on feedback from actual use cases. Details for now are left in the comments above.

optuna/study.py
c-bata: Sounds reasonable for now 👍

hvy:
Would `study.copy_trials(trials: List[FrozenTrial])` replace both the introduced function and the method? 👀

toshihikoyanase:
Let me merge this PR since it received three approvals.
Please use the corresponding issue #821, or create a new one for further discussion.

PR #1341, opened 2020-06-09
Add `MPIStudy` to optimize distributed neural network training with Horovod

toshihikoyanase:
## Motivation

This PR enables Horovod users to optimize their distributed training of neural networks. This feature (`MPIStudy`) is originally developed by @g-votte for the [ABCI hackathon](https://github.com/g-votte/abci-optuna-horovod-example), and extended for the [Tsubame hackathon](https://github.com/toshihikoyanase/tsubame-optuna-example/blob/master/README.tensorflow.md).

## Description of the changes

The code is almost the same as `ChainerMNStudy` except for a small part of communicator access (e.g., `comm.mpi_comm.bcast` or `comm.bcast`). Only the rank-0 node will communicate with storage, samplers, and pruners, and broadcast the results to other nodes.

[Opinion Wanted]
- Most of the code is duplicated with `ChainerMNStudy`. I think we can reduce the code by sharing the implementation (e.g., a base class).
- Horovod starts to support `gloo`, which is a communication library without requiring MPI. Do we also provide `GlooStudy`? Or, is it possible to remove `gloo` or `mpi4py` from the dependencies if we use the high-level API of Horovod?

[TODO]
- [x] Add unit tests.
- [x] Add `horovod` to the dependencies of testing and example.
- [x] Add the example to CI jobs.


toshihikoyanase:
I set up the CI configuration for `MPIStudy` and made this PR review-ready.

PR #1345, opened 2020-06-09
Simplify log messages

Crissman:
## Motivation
Log messages for Optuna are long and use `#` which is superfluous.

## Description of the changes
Shorten the log messages for pruned, successful, and failed trials.

### DIFF
--- a/docs/source/tutorial/pruning.rst
+++ b/docs/source/tutorial/pruning.rst
@@ -55,19 +55,19 @@ Executing the script above:
 .. code-block:: bash
 
     $ python prune.py
-    [I 2018-11-21 17:27:57,836] Finished trial#0 resulted in value: 0.052631578947368474. Current best value is 0.052631578947368474 with parameters: {'alpha': 0.011428158279113485}.
-    [I 2018-11-21 17:27:57,963] Finished trial#1 resulted in value: 0.02631578947368418. Current best value is 0.02631578947368418 with parameters: {'alpha': 0.01862693201743629}.
-    [I 2018-11-21 17:27:58,164] Finished trial#2 resulted in value: 0.21052631578947367. Current best value is 0.02631578947368418 with parameters: {'alpha': 0.01862693201743629}.
-    [I 2018-11-21 17:27:58,333] Finished trial#3 resulted in value: 0.02631578947368418. Current best value is 0.02631578947368418 with parameters: {'alpha': 0.01862693201743629}.
-    [I 2018-11-21 17:27:58,617] Finished trial#4 resulted in value: 0.23684210526315785. Current best value is 0.02631578947368418 with parameters: {'alpha': 0.01862693201743629}.
-    [I 2018-11-21 17:27:58,642] Setting status of trial#5 as TrialState.PRUNED.
-    [I 2018-11-21 17:27:58,666] Setting status of trial#6 as TrialState.PRUNED.
-    [I 2018-11-21 17:27:58,675] Setting status of trial#7 as TrialState.PRUNED.
-    [I 2018-11-21 17:27:59,183] Finished trial#8 resulted in value: 0.39473684210526316. Current best value is 0.02631578947368418 with parameters: {'alpha': 0.01862693201743629}.
-    [I 2018-11-21 17:27:59,202] Setting status of trial#9 as TrialState.PRUNED.
+    [I 2020-06-12 16:54:23,876] Trial 0 finished with value: 0.3157894736842105 and parameters: {'alpha': 0.00181467547181131}. Best is trial 0 with value: 0.3157894736842105.
+    [I 2020-06-12 16:54:23,981] Trial 1 finished with value: 0.07894736842105265 and parameters: {'alpha': 0.015378744419287613}. Best is trial 1 with value: 0.07894736842105265.
+    [I 2020-06-12 16:54:24,083] Trial 2 finished with value: 0.21052631578947367 and parameters: {'alpha': 0.04089428832878595}. Best is trial 1 with value: 0.07894736842105265.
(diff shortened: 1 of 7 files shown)


optuna/study.py
```diff
@@ -789,7 +789,7 @@ def _run_trial(
 
         if math.isnan(result):
             message = (
-                "Setting status of trial#{} as {} because the objective function "
+                "Trial {} set as {} because the objective function "
```
hvy, line 792: A detail but how about being consistent with "Trial {} pruned", i.e. "Trial {} failed because ..."?

Crissman: Good idea. Done!

tests/test_study.py
```diff
@@ -407,7 +404,7 @@ def func_none(_):
         frozen_trial = study._storage.get_trial(trial._trial_id)
 
         expected_message = (
-            "Setting status of trial#3 as TrialState.FAIL because the returned "
+            "Trial 3 failed, because the returned "
             "value from the objective function cannot be casted to float. "
```
HideakiImamura:
```suggestion
            "value from the objective function cannot be cast to float. "
```

HideakiImamura: Sorry for the confusion. My concern is already addressed in the commit just submitted.

docs/source/tutorial/pruning.rst
```diff
@@ -55,19 +55,19 @@ Executing the script above:
...
+    [I 2020-06-12 16:54:24,303] Trial 4 finished with value: 0.07894736842105265 and parameters: {'alpha': 2.805688697062864e-05}. Best is trial 3 with value: 0.052631578947368474.
+    [I 2020-06-12 16:54:24,315] Trial 5 pruned. 
+    [I 2020-06-12 16:54:24,355] Trial 6 pruned. 
+    [I 2020-06-12 16:54:24,511] Trial 7 finished with value: 0.052631578947368474 and parameters: {'alpha': 2.243775785299103e-05}. Best is trial 3 with value: 0.052631578947368474.
+    [I 2020-06-12 16:54:24,625] Trial 8 finished with value: 0.1842105263157895 and parameters: {'alpha': 0.007021209286214553}. Best is trial 3 with value: 0.052631578947368474.
+    [I 2020-06-12 16:54:24,629] Trial 9 pruned. 
     ...
 
-We can see ``Setting status of trial#{} as TrialState.PRUNED`` in the log messages.
+We can see ``Trial 5 pruned.`` in the log messages.
```
HideakiImamura, line 70:
```suggestion
We can see ``Trial {} pruned.`` in the log messages.
```

docs/source/tutorial/pruning.rst
```diff
@@ -55,19 +55,19 @@ Executing the script above:
...
+    [I 2020-06-12 16:54:24,083] Trial 2 finished with value: 0.21052631578947367 and parameters: {'alpha': 0.04089428832878595}. Best is trial 1 with value: 0.07894736842105265.
+    [I 2020-06-12 16:54:24,185] Trial 3 finished with value: 0.052631578947368474 and parameters: {'alpha': 0.004018735937374473}. Best is trial 3 with value: 0.052631578947368474.
+    [I 2020-06-12 16:54:24,303] Trial 4 finished with value: 0.07894736842105265 and parameters: {'alpha': 2.805688697062864e-05}. Best is trial 3 with value: 0.052631578947368474.
+    [I 2020-06-12 16:54:24,315] Trial 5 pruned. 
+    [I 2020-06-12 16:54:24,355] Trial 6 pruned. 
+    [I 2020-06-12 16:54:24,511] Trial 7 finished with value: 0.052631578947368474 and parameters: {'alpha': 2.243775785299103e-05}. Best is trial 3 with value: 0.052631578947368474.
+    [I 2020-06-12 16:54:24,625] Trial 8 finished with value: 0.1842105263157895 and parameters: {'alpha': 0.007021209286214553}. Best is trial 3 with value: 0.052631578947368474.
+    [I 2020-06-12 16:54:24,629] Trial 9 pruned. 
     ...
```
HideakiImamura, line 68: The generated log messages seem to be changed. I'm fine with either, but maybe we shouldn't put in extra changes to clarify the PR changes.

docs/source/tutorial/pruning.rst
hvy: It might be better having a concrete number here, but I think the following sentence could be improved (in that case especially).

optuna/study.py
```diff
@@ -778,19 +776,18 @@ def _run_trial(
...
             _logger.warning(message)
             self._storage.set_trial_system_attr(trial_id, "fail_reason", message)
             self._storage.set_trial_state(trial_id, TrialState.FAIL)
             return trial
 
         if math.isnan(result):
-            message = (
-                "Setting status of trial#{} as {} because the objective function "
-                "returned {}.".format(trial_number, TrialState.FAIL, result)
+            message = "Trial {} failed, because the objective function " "returned {}.".format(
```
hvy, line 789: Implicit string concat.
```suggestion
            message = "Trial {} failed, because the objective function returned {}.".format(
```

docs/source/tutorial/pruning.rst
Crissman: I don't understand. I updated the tutorial to match the current output. That seems required. What are the "extra changes" you mention?

HideakiImamura: I mean the changes in optimized values. But, it's fine to update them here. Thanks!

docs/source/tutorial/pruning.rst
HideakiImamura: I agree. So, I think the following sentence can be improved like
"This means the trial 5 is stopped before it finishes all iterations. Trial 6 and 9 are stopped as well"

Crissman: Cleaned up those sentences.

PR #1350, opened 2020-06-10
Set `python_requires` in setup.py to clarify supported Python version

harupy:
## Motivation

To clarify Optuna only supports certain Python versions (3.5 or newer currently).

## Description of the changes

- Add `python_requires=">=3.5"` in `setup.py`.

## References
- https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires
- https://packaging.python.org/guides/dropping-older-python-versions/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -164,6 +164,7 @@ setup(
             "storages/rdb/alembic/versions/*.*",
         ]
     },
+    python_requires=">=3.5",
     install_requires=get_install_requires(),
     tests_require=get_tests_require(),
     extras_require=get_extras_require(),


hvy:
Interesting that there are still Py2 download, thanks @toshihikoyanase for the reference. Good catch anyhow, I don't think that `python_requires` has been discussed before. But given the numbers above and the fact that it's better fail early, I'm also okay with this change, introducing it for v2.0.

harupy:
@toshihikoyanase Thank you for the comment.

> A few people still use Optuna in Python 2.7 and Python 3.4 somehow

I think they use the old version because the recent versions have type annotations which should raise an error in Python 2.7 and 3.4.

toshihikoyanase:
@harupy

> they use the old version 

That makes sense. If so, this change does not affect them. Thank you for your comment.

hvy:
Slightly off the subject but do we want to start specifying `classifiers` too? https://packaging.python.org/guides/distributing-packages-using-setuptools/#classifiers

harupy:
@hvy That'd be useful because shields.io provides a badge that detects python classifiers on a PyPI project page and shows them.

Screen Shot 2020-06-11 at 13 42 35

https://github.com/tensorflow/tensorflow

hvy:
Thanks, I wasn't aware of shields.io depending on it. I created an issue here https://github.com/optuna/optuna/issues/1356.

PR #1362, opened 2020-06-12
Rename private modules in `optuna.samplers`.

toshihikoyanase:
## Motivation

This PR renames private modules in `optuna.samplers` like  #1359.

## Description of the changes

This PR adds `_` prefix to the following files and dir names:

- optuna/samplers/tpe
- optuna/samplers/base.py
- optuna/samplers/cmaes.py
- optuna/samplers/grid.py
- optuna/samplers/random.py


optuna/samplers/_tpe/__init__.py
```diff
@@ -0,0 +1 @@
+from optuna.samplers._tpe.sampler import TPESampler  # NOQA
```
hvy, line 1: As it's only used from `samplers/__init__.py`, maybe it's simpler to specify the full path there instead of introducing another layer of indirection? Maybe outside the scope of this PR.

toshihikoyanase: Thank you for pointing it out. I removed it and used `optuna.samplers._tpe.sampler.TPESampler` in `samplers/__init__.py`.
Commit: https://github.com/optuna/optuna/pull/1362/commits/7122120536644dc6c66a89f8796440b0e4cb8d09

toshihikoyanase:
@HideakiImamura @hvy 
> As a minor suggestion, how about renaming samplers/_tpe/sampler.py and samplers/_tpe/parzen_estimator.py as well?

> These changes make it clear to the user that those are hidden.

Thank you for your comments. I agree with you. Your suggestions make the naming rule clear. I'll apply the change.

toshihikoyanase:
@HideakiImamura @hvy Hmm, it may be unusual to see multiple `_` in a module path (https://github.com/optuna/optuna/pull/1362/commits/510dba8bf5a46cfe89672523a85041dd54c36935).

I briefly checked other libraries including [Chainer](https://github.com/chainer/chainer) and [PyTorch](https://github.com/pytorch/pytorch).

Chainer basically does not use `_` in directory names, and only uses it in filenames.
PyTorch sometimes uses `_` in directory names but does not add `_` to the children as follows:
- https://github.com/pytorch/pytorch/tree/master/torch/utils/data/_utils
- https://github.com/pytorch/pytorch/tree/master/torch/autograd/_functions
- https://github.com/pytorch/pytorch/tree/master/torch/distributed/rpc/_testing

So, I think we have two choices
1. Add `_` prefixes only to file names.
1. Add `_` prefixes to directory names. If the prefix is added to the directory, we do not add `_` to the children.

What do you think?

toshihikoyanase:
> 2. Add `_` prefixes to directory names. If the prefix is added to the directory, we do not add _ to the children.

I'd like to employ the second approach and reverted the changes in https://github.com/optuna/optuna/pull/1362/commits/d7664e12b31f79783e3fa63a00c8cc3933c5a01e. I couldn't find the other project that uses the current approach. And PEP8 assumes that a private directory also has the `_` prefix.

c.f., [PEP8 - Public and Internal Interfaces](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces)

toshihikoyanase:
In addition, we can use `__all__` to show the public interface clearly. I think we can discuss it in a followup PR.

PR #1367, opened 2020-06-14
Add missing `__init__.py` files

harupy:
## Motivation

`mypy` ignores directories that don't contain `__init__.py`.

## Description of the changes

Add `__init__.py` files.

I ran the test below to make sure all the directories contain `__init__.py`

```python
import os


def test_init_exists_in_all_directories() -> None:
    dirs = []
    for top in ["optuna", "tests"]:
        for root, _, files in os.walk(top):
            if any(f.endswith(".py") for f in files) and "__init__.py" not in files:
                dirs.append(root)
    assert len(dirs) == 0, "Found directories that don't contain `__init__.py`: {}".format(dirs)

```

### DIFF
--- a/tests/multi_objective/test_trial.py
+++ b/tests/multi_objective/test_trial.py
@@ -149,7 +149,7 @@ def test_dominates() -> None:
     directions = [StudyDirection.MINIMIZE, StudyDirection.MAXIMIZE]
 
     def create_trial(
-        values: List[Optional[float]], state: TrialState = TrialState.COMPLETE,
+        values: List[float], state: TrialState = TrialState.COMPLETE,
     ) -> FrozenMultiObjectiveTrial:
         n_objectives = len(values)
         trial = optuna.trial.FrozenTrial(
--- a/tests/multi_objective/test_trial.py
+++ b/tests/multi_objective/test_trial.py
@@ -1,6 +1,7 @@
 import datetime
 from typing import List
 from typing import Optional
+from typing import Tuple
 
 import pytest
 
@@ -13,7 +14,7 @@ from optuna.trial import TrialState
 def test_suggest() -> None:
     study = optuna.multi_objective.create_study(["maximize", "maximize"])
 
-    def objective(trial: optuna.multi_objective.trial.MultiObjectiveTrial) -> List[float]:
+    def objective(trial: optuna.multi_objective.trial.MultiObjectiveTrial) -> Tuple[float, float]:
         p0 = trial.suggest_float("p0", -10, 10)
         p1 = trial.suggest_uniform("p1", 3, 5)
         p2 = trial.suggest_loguniform("p2", 0.00001, 0.1)
@@ -22,7 +23,10 @@ def test_suggest() -> None:
         p5 = trial.suggest_categorical("p5", [7, 1, 100])
         p6 = trial.suggest_float("p6", -10, 10, step=1.0)
         p7 = trial.suggest_int("p7", 1, 7, log=True)
-        return [p0 + p1 + p2, p3 + p4 + p5 + p6 + p7]
+        return (
+            p0 + p1 + p2,
+            p3 + p4 + p5 + p6 + p7,
+        )
 
     study.optimize(objective, n_trials=10)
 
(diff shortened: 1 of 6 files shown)


harupy:
`mypy` check fails:

```bash
tests/multi_objective/test_trial.py:25: error: Unsupported operand types for + ("float" and "None")
tests/multi_objective/test_trial.py:25: error: Unsupported operand types for + ("float" and "str")
tests/multi_objective/test_trial.py:25: note: Right operand is of type "Union[None, bool, int, float, str]"
tests/multi_objective/test_trial.py:35: error: Argument 1 to "report" of "MultiObjectiveTrial" has incompatible type "Tuple[int, int, int]"; expected "Tuple[float]"
tests/multi_objective/test_trial.py:36: error: Argument 1 to "report" of "MultiObjectiveTrial" has incompatible type "Tuple[int, int, int]"; expected "Tuple[float]"
tests/multi_objective/test_trial.py:150: error: Value expression in dictionary comprehension has incompatible type "Optional[float]"; expected type "float"
tests/multi_objective/samplers_tests/test_nsga2.py:41: error: Argument "population_size" to "NSGAIIMultiObjectiveSampler" has incompatible type "float"; expected "int"
optuna/storages/rdb/alembic/env.py:32: error: Function is missing a return type annotation
optuna/storages/rdb/alembic/env.py:32: note: Use "-> None" if function does not return a value
optuna/storages/rdb/alembic/env.py:51: error: Function is missing a return type annotation
optuna/storages/rdb/alembic/env.py:51: note: Use "-> None" if function does not return a value
tests/multi_objective/test_study.py:131: error: Incompatible return value type (got "Tuple[float, float]", expected "List[float]")
Found 9 errors in 4 files (checked 171 source files)
```

harupy:
It seems the type annotation of `values` in `multi_objective` should be fixed. It is annotated as `Tuple[float]` which represents a single-element tuple, but it should be `Tuple[float, ...]` which represents a variable-length tuple.

https://github.com/optuna/optuna/blob/e7afd401a254e5dc1cd4213ea58918afeb221788/optuna/multi_objective/trial.py#L111

hvy:
Good catch! We should address the failures in the same PR. Taking a brief look through the errors, they mostly seem straight forward to address (e.g. by introducing assertions). Do you want to do that here? I otherwise suggest we split up this PR into multiple parts.

toshihikoyanase:
Thank you for finding the missing type check.

I think `optuna/storages/rdb/alembic/` is a kind of settings for `alembic` rather than a part of Optuna's submodules (i.e., Optuna doesn't import it as a module.) It may be controversial to include the `alembic` submodule only for `mypy`.

What do you think?

harupy:
@hvy Got it. I'll fix all the errors in this PR.

@toshihikoyanase If so, I'll make `mypy` ignore files under `alembic`.

tests/multi_objective/test_trial.py
```diff
@@ -13,28 +14,33 @@
...
 
-    def objective(trial: optuna.multi_objective.trial.MultiObjectiveTrial) -> List[float]:
+    def objective(trial: optuna.multi_objective.trial.MultiObjectiveTrial) -> Tuple[float, float]:
         p0 = trial.suggest_float("p0", -10, 10)
         p1 = trial.suggest_uniform("p1", 3, 5)
         p2 = trial.suggest_loguniform("p2", 0.00001, 0.1)
         p3 = trial.suggest_discrete_uniform("p3", 100, 200, q=5)
         p4 = trial.suggest_int("p4", -20, -15)
-        p5 = trial.suggest_categorical("p5", [7, 1, 100])
+        p5 = cast(int, trial.suggest_categorical("p5", [7, 1, 100]))
```
harupy, line 23: https://stackoverflow.com/questions/44739764/how-to-cast-a-typing-union-to-one-of-its-subtypes-in-python

harupy:
The tests in `tests/test_cli.py` failed.

toshihikoyanase:
@harupy I found similar issues in other PRs (e.g., https://circleci.com/gh/optuna/optuna/45221?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) and daily CI (https://github.com/optuna/optuna/actions/runs/136123282). I'm sorry but I'm under investigation.

harupy:
That makes sense. I'll fix the type annotaions.

harupy:
@hvy I have fixed the type annotations to use `Sequence` instead of `Tuple`. Could you review the code again?

hvy:
@toshihikoyanase could you cont. for the second approval?

tests/multi_objective/test_trial.py
harupy: This is just an idea. It'd be nice if we could infer the return type of `suggest_categorical` in a generic way like "return int if int list given" instead of returning `Union` of python primitive types because this enforces users to call `cast` to a suggested value to resolve type errors when they use Optuna with a type-checking tool like `mypy`.

PR #1373, opened 2020-06-15
Rename private modules in `optuna.storages`.

toshihikoyanase:
## Description of the changes
This PR renames private modules in `optuna.storages` like #1359.


toshihikoyanase:
Thank you for your review. I fixed the conflicts. PTAL.

PR #1376, opened 2020-06-15
Rename private modules in `optuna.integration.lightgbm_tuner`.

toshihikoyanase:
This PR renames private modules in `optuna.integration.lightgbm_tuner` like #1359.

In addition, it also adds an alias from `optuna.integration.lightgbm_tuner.LightGBMTunerCV` to `optuna.integration.lightgbm.LightGBMTunerCV` to be consistent with the `LightGBMTuner` class.


toshihikoyanase:
@HideakiImamura 
Thank you for your approval, but I think we can make `optuna.integration.lightgbm_tuner` private instead of its children like https://github.com/optuna/optuna/pull/1362#issuecomment-644498605. Can I work on it in this PR?

HideakiImamura:
Yes, you are right. Thanks for pointing. It's good to work on this PR.

toshihikoyanase:
I created the PR #1378  to rename `optuna.integration.lightgbm_tuner` module.
This is mainly because I needed to discard all the changes applied in this PR, and it seems to be reasonable to create a new one. So, let me close this PR.

PR #1378, opened 2020-06-16
Make `optuna.integration.lightgbm_tuner` private.

toshihikoyanase:
## Motivation
#1376 tried to rename the files in `optuna.integration.lightgbm_tuner`, such an approach is not consistent with https://github.com/optuna/optuna/pull/1362#issuecomment-644498605.

Instead, this PR renames the module `optuna.integration.lightgbm_tuner`.
I'm concerned that we exposed `optuna.integration.lightgbm_tuner` for `LightGBMTuner` and `LightGBMTuner` in the [reference](https://optuna.readthedocs.io/en/stable/reference/integration.html#optuna.integration.lightgbm_tuner.LightGBMTuner). I'm really sorry for the existing users, but I'd like to change the module of them to `optuna.integration.lightgbm`.

## Description of the changes

It renames `optuna.integration.lightgbm_tuner` to `optuna.integration._lightgbm_tuner`.

In addition, it also adds an alias from optuna.integration.lightgbm_tuner.LightGBMTunerCV to optuna.integration.lightgbm.LightGBMTunerCV to be consistent with the LightGBMTuner class.

[Opinion wanted]
- Do we have to rename private classes in `optuna.integration.lightgbm_tuner` such as  `BaseTuner`, `OptunaObjective` and `OptunaObjectiveCV`? In my understanding, the classes under the private module are also private, and we don't have to add the `_` prefix.


HideakiImamura:
Thanks for the updated PR!

In my understanding, we should rename the private classes in `optuna.integration.lightgbm_tuner ` as the private classes in `optuna.samplers.tpe` such as `_ParzenEstimator` and `_ParzenEstimatorParameters`. What do you think?

toshihikoyanase:
> we should rename the private classes in `optuna.integration.lightgbm_tuner` as the private classes in `optuna.samplers.tpe` such as `_ParzenEstimator` and `_ParzenEstimatorParameters`. 

Thank you for pointing it out. That makes sense. `optuna/integration/_lightgbm_tuner/optimize.py` is long and your suggestion will make the code clear. I'd like to apply it.

hvy:
You might have to fix the type hint of the return value of `LightGBMTuner.tune_params` as well.

toshihikoyanase:
> By the way, just wondering but I guess this method is sort of private? Then it might be okay to return an object that's of private type.

Yes, it is. I think we can also rename it to `LightGBMTuner._tune_params`.

optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -42,13 +42,13 @@
 _LGBM_PARAMS_KEY = "lightgbm_tuner:lgbm_params"
 
 # EPS is used to ensure that a sampled parameter value is in pre-defined value range.
-EPS = 1e-12
+_EPS = 1e-12
```
HideakiImamura, line 45: Do we need to make this type of variables private? If so, we need to change other variables (e.g. `EPS` in `optuna.samplers.tpe.sampler.py` for consistency.

hvy: I agree that it's open for discussion. Maybe contradictory to how we implicitly propagate private-ness between packages and modules. How about leaving this PR as is since it covers not just `optuna.integration.lightgbm_tuner` but the entire code base, and continue this discussion elsewhere?

toshihikoyanase: Thank you for your comments. 

> How about leaving this PR as is since it covers not just optuna.integration.lightgbm_tuner but the entire code base, and continue this discussion elsewhere?

I agree with you. We can work on this issue in a new PR or issue.

Issue #1307, opened 2020-05-31
[RFC] Change the default argument of `gc_after_trial` in `Study.optimize` to `False`.


ytsmiling:
## Description
`Study.optimize` calls `gc.collect()` after every objective  evaluation. This change was added in #377. In #377 we added the gc invocation because ChainerMN example loads the full MNIST dataset in every objective call and it wasted too much memory. I understand that the gc invocation is useful for some users. However, the gc call can be a bottleneck in the whole `optuna` optimization loop.
I'd like to request for comments about changing the default behavior of `Study.optimize` to *don't* invoke gc.
### Pros
- We need the gc in only some special environments such as CircleCI (otherwise, python automatically invoke gc when necessary).
- Removing the gc invocation improves optimization speed.
### Cons
- Some user programs might suffer from OOM in certain environments.
- The current optuna optimization loop is much faster compared to `v1.3` and further speed up might be beneficial for only a small fraction of users (Please refer to the benchmarks in the next section). It might be sufficient to add a performance tips section to the official doc.

## Microbenchmark
The followings are some microbenchmark results. (I used a similar setup with #1135.)
### Without GC (master)
```
optimization with 2000 trials / 30 params: 259.5s
optimization with 2000 trials /  2 params: 16.2s
optimization with 1000 trials / 30 params: 72.6s
optimization with 1000 trials /  2 params: 4.9s
```
(Remark: Performance depends on which samplers to use.)
### With GC (master)
```
optimization with 2000 trials / 30 params: 363.7s
optimization with 2000 trials /  2 params: 105.6s
optimization with 1000 trials / 30 params: 122.8s
optimization with 1000 trials /  2 params: 48.5s
```
(Remark: GC performance also depends on the user program.)
### Without GC (v1.3)
As for reference, I add a benchmark result on `v1.3`.
```
optimization with 1000 trials / 30 params: 325.7s
optimization with 1000 trials /  2 params: 83.6s
```
(Remark: Compared to `v1.3`, the current master has additional functionality in `InMemoryStorage` (#1228)  and performs appropriate CoW handling (#1139), which incur overheads to master.)
### With GC (v1.3)
```
optimization with 1000 trials /  2 params: 235.6s
```

hvy:
A brief recap.

1. OOM issues encountered https://github.com/optuna/optuna/pull/325
1. Decided to automatically run GC after each trial https://github.com/optuna/optuna/pull/377
[...]

PR #325, opened 2019-02-27
Fix occasional OOM errors during execution of ChainerMN examples on CircleCI.

sile:
The execution of CircleCI's `examples-*` jobs sometimes fails because of OOM error due to ChainerMN example.

I guess the cause of this error as follows:
1. A VM of CircleCI is equipped with very large memory (about 70GB).
2. But only 4GB is available for normal users.
3. However, Python interpreter recognizes as if there is a lot of memory.
   - GCs are hardly triggered by the interpreter.
4. When the memory usage of a Python program exceeds the 4GB limit, the program will be suddenly killed.
   - If a GC were triggered, the program would have been able to continue execution.

In this PR, I added manual invocations of `gc.collect()` method to prevent the above scenario.

Memory Usage Comparison
-------------------------------

Without `gc.collect()` (the memory usage in each execution has varied between 2GB and 4GB):
```console
// Preparation.
$ ssh -p ${CIRCLE_CI_PORT} ${CIRCLE_CI_HOST}
$ cd project && . venv/bin/activate
$ STORAGE_URL=sqlite:///example.db
$ STUDY_NAME=`optuna create-study --storage $STORAGE_URL`
$ export OMP_NUM_THREADS=1

// Execute a ChainerMN example.
$ /usr/bin/time -v mpirun -n 2 -- python examples/chainermn_simple.py $STUDY_NAME $STORAGE_URL | grep 'Maximum resident'
        Maximum resident set size (kbytes): 2546996  // Memory Usage=2.5GB 
```

With `gc.collect()` (memory usage=0.9GB):
```console
// Preparation.
$ vim examples/chainermn_simple.py  // Add `gc.collect()`

// Execute a ChainerMN example.
$ /usr/bin/time -v mpirun -n 2 -- python examples/chainermn_simple.py $STUDY_NAME $STORAGE_URL | grep 'Maximum resident'
        Maximum resident set size (kbytes): 954336  // Memory Usage=0.9GB
```

PR #1380, opened 2020-06-16
Do not run the GC after every trial by default.

hvy:
## Motivation

Fixes https://github.com/optuna/optuna/issues/1307. See this issue for discussions.

## Description of the changes

Changes the default GC collection behavior of `Study.optimize` to not run a collection.

### DIFF
--- a/optuna/integration/chainermn.py
+++ b/optuna/integration/chainermn.py
@@ -1,4 +1,3 @@
-import gc
 from typing import Optional
 
 from optuna._imports import try_import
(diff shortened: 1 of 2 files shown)


hvy:
Is it not addressed by https://github.com/optuna/optuna/blob/master/optuna/integration/chainermn.py#L146? I was also wondering whether this line should be kept or controllable similar to the standard `Study`.

toshihikoyanase:
The line is executed by non-rank-0 nodes, but I think we need to call GC for rank-0 node. The rank-0 node calls `Study.optimize` without `gc_after_trial` option at https://github.com/optuna/optuna/blob/master/optuna/integration/chainermn.py#L122. And I think it may cause similar issues to #325.

> this line should be kept or controllable similar to the standard Study.

Depending on the user code and environments, GC calls can be omitted. I support the latter approach, but I think we can work on it in a new PR.

hvy:
Thanks for your explanation. I now see what you meant, that the behavior of `ChainerMNStudy` will be different for different workers, that only the first worker will not run the GC collection. However, as the underlying motivation behind this change is to leave the memory handling to the users, I suggest we stop running the GC in the `ChainerMNStudy` as well. Let me change it that way (discussed with @toshihikoyanase).

hvy:
PTAL.

PR #1384, opened 2020-06-17
Deprecate `optuna study optimize` command.

toshihikoyanase:
## Motivation
Optuna provides the CLI to invoke optimization from the shell, but it does not support some features like the pruning and the callback. This PR proposes to remove it in the future instead of adding the new feature support because of the following reasons:

- the command-line options will be complicated to manage the pruning and callback, and
- most of the users directly execute their Python scripts.

## Description of the changes
This PR adds a deprecation warning to `optuna study optimize` command. 

We need to update the code comment in the example files, but I leave them as they are to keep the diff small.

### DIFF
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -296,8 +296,8 @@ class _StudyOptimize(_BaseCommand):
         # type: (Namespace) -> int
 
         message = (
-            "The use of the `study optimize` command is deprecated. Please execute your Python script directly "
-            "instead."
+            "The use of the `study optimize` command is deprecated. Please execute your Python "
+            "script directly instead."
         )
         warnings.warn(message, DeprecationWarning)
         self.logger.warning(message)
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -296,7 +296,7 @@ class _StudyOptimize(_BaseCommand):
         # type: (Namespace) -> int
 
         message = (
-            "The use of `study optimize` command is deprecated. Please execute your Python script "
+            "The use of the `study optimize` command is deprecated. Please execute your Python script directly "
             "instead."
         )
         warnings.warn(message, DeprecationWarning)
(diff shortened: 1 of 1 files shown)


optuna/cli.py
```diff
@@ -295,6 +295,13 @@ def get_parser(self, prog_name):
     def take_action(self, parsed_args):
         # type: (Namespace) -> int
 
+        message = (
+            "The use of `study optimize` command is deprecated. Please execute your Python script "
```
hvy, line 299: Some minor suggestions. I'm not sure about the line length though.
```suggestion
            "The use of the `study optimize` command is deprecated. Please execute your Python script directly "
```

optuna/cli.py
```diff
@@ -253,7 +253,7 @@ def take_action(self, parsed_args):
 
 
 class _StudyOptimize(_BaseCommand):
-    """Start optimization of a study."""
+    """Start optimization of a study. Deprecated since version 2.0.0."""
```
hvy, line 256: This message might be a bit difficult to notice blending in, but it's not a strong opinion.

toshihikoyanase: I agree. This message is only shown in the list of commands. Do we remove it?

```console
$ optuna
...
Commands:
  complete       print bash completion command (cliff)
  create-study   Create a new study.
  dashboard      Launch web dashboard (beta).
  delete-study   Delete a specified study.
  help           print detailed help for another command (cliff)
  storage upgrade  Upgrade the schema of a storage.
  studies        Show a list of studies.
  study optimize  Start optimization of a study. Deprecated since version 2.0.0.
  study set-user-attr  Set a user attribute to a study.
```

HideakiImamura: I think we don't have to remove it. It's helpful for users.

PR #1385, opened 2020-06-18
Add FAQ section about OOM.

hvy:
## Motivation

Follow-up to https://github.com/optuna/optuna/pull/1380. 

Now that we don't do any GC collection sweep by default, OOM might become an issue for certain users (as described in tickets that can be tracked from the above PR). A section in the FAQ might alleviate some of the problem.

## Description of the changes

Introduces a section in the FAQ about OOM, and adds a reference to it from the API reference for `gc_after_trial`.


hvy:
Thanks for your feedback, I addressed both your points. PTAL.

PR #1387, opened 2020-06-18
Deprecate step argument in IntLogUniformDistribution

nzw0301:
## Motivation

Since https://github.com/optuna/optuna/pull/1329 disables `suggest_int` to use `step > 1` with `log=True`, we do not need to have `step` argument in `IntLogUniformDistribution`. 

## Description of the changes

This PR removes the `step` argument and refactors the sampling and tests codes.

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -320,7 +320,8 @@ class IntLogUniformDistribution(BaseDistribution):
             A step for spacing between values.
 
             .. deprecated:: 2.0.0
-                ``step`` argument will be removed in the future. The removal of this feature is currently scheduled for v3.0.0, but this schedule is subject to change.
+                ``step`` argument will be removed in the future. The removal of this feature is
+                currently scheduled for v4.0.0, but this schedule is subject to change.
     """
 
     def __init__(self, low: int, high: int, step: int = 1) -> None:
@@ -334,6 +335,12 @@ class IntLogUniformDistribution(BaseDistribution):
             raise ValueError(
                 "The `low` value must be equal to or greater than 1 for a log distribution "
                 "(low={}, high={}).".format(low, high)
+            )
+
+        if step != 1:
+            warnings.warn(
+                "`step` accepts only `1`, so `step` is replaced with `1`. "
+                "`step` argument is deprecated and will be removed in the future."
             )
 
         self.low = low
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -3,6 +3,7 @@ import decimal
 import json
 import warnings
 
+from optuna import logging
 from optuna import type_checking
 
 if type_checking.TYPE_CHECKING:
@@ -12,6 +13,9 @@ if type_checking.TYPE_CHECKING:
     from typing import Union  # NOQA
 
     CategoricalChoiceType = Union[None, bool, int, float, str]
+
+
+_logger = logging.get_logger(__name__)
 
 
 class BaseDistribution(object, metaclass=abc.ABCMeta):
@@ -341,7 +345,7 @@ class IntLogUniformDistribution(BaseDistribution):
             )
 
         if step != 1:
-            warnings.warn(
+            _logger.warning(
                 "`step` accepts only `1`, so `step` is replaced with `1`. "
                 "`step` argument is deprecated and will be removed in the future. "
                 "The removal of this feature is currently scheduled for v4.0.0, "
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -347,7 +347,7 @@ class IntLogUniformDistribution(BaseDistribution):
                 "`step` argument is deprecated and will be removed in the future. "
                 "The removal of this feature is currently scheduled for v4.0.0, "
                 "but this schedule is subject to change.",
-                DeprecationWarning,
+                FutureWarning,
             )
 
         self.low = low
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -3,7 +3,6 @@ import decimal
 import json
 import warnings
 
-from optuna import logging
 from optuna import type_checking
 
 if type_checking.TYPE_CHECKING:
@@ -13,9 +12,6 @@ if type_checking.TYPE_CHECKING:
     from typing import Union  # NOQA
 
     CategoricalChoiceType = Union[None, bool, int, float, str]
-
-
-_logger = logging.get_logger(__name__)
 
 
 class BaseDistribution(object, metaclass=abc.ABCMeta):
@@ -326,9 +322,10 @@ class IntLogUniformDistribution(BaseDistribution):
             .. note::
                 This value is valid for only 1. Otherwise, the value is replaced with 1.
 
-            .. deprecated:: 2.0.0
-                ``step`` argument will be removed in the future. The removal of this feature is
-                currently scheduled for v4.0.0, but this schedule is subject to change.
+            .. warning::
+                Deprecated in v2.0.0. ``step`` argument will be removed in the future.
+                The removal of this feature is currently scheduled for v4.0.0,
+                but this schedule is subject to change.
     """
 
     def __init__(self, low: int, high: int, step: int = 1) -> None:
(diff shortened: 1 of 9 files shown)


HideakiImamura:
Hi @nzw0301! I'm so sorry to comment after the approval. Could you take a look?

With careful consideration, I've noticed that the `step` argument of `IntLogUniformDistribution` has already appeared in v1.5.0. Therefore, removing the argument without any deprecation is unfriendly to users. Can we address this problem by not dropping the argument but deprecating it? Sorry for the confusion.

nzw0301:
@HideakiImamura Sure! Please don't worry about it. So, do I just add a warning message to the constructor of `IntLogUniformDistribution` and revert `step` argument of `IntLogUniformDistribution`?

HideakiImamura:
@nzw0301 Thanks for your swift reply! Yes, adding warning messages as the `optuna._deprecated.deprecated` decorator is a good approach. (The deprecation decorator cannot be applied to the argument. But the default deprecation note is given in `optuna._deprecated.py`.)

nzw0301:
@HideakiImamura Thank you for your answer!  I'm reading the codes and I think we cannot change the template of deprecated message defined in https://github.com/optuna/optuna/blob/0a2c099f6292945186455f6dcb88a411b13c20fa/optuna/_deprecated.py#L14-L20
So I think this deprecated decorator's message may confuse users for this issue as you mentioned. So I will add a warning message not using this decorator... How do you think?

nzw0301:
In addition, even though I added the decorator to the `IntLogUniformDistribution`, no warning message appears because 
`warnings.warn` with `DeprecationWarning` in https://github.com/optuna/optuna/blob/0a2c099f6292945186455f6dcb88a411b13c20fa/optuna/_deprecated.py#L87-L97 is ignored by default unless it is triggered by code in `__main__` (see [ref](https://docs.python.org/3/library/warnings.html#warning-categories)). I suppose this behaviour is undesirable to notify users of the deprecation...

HideakiImamura:
Thanks for your reply! Yes, this decorator can only be applied to the function or class, so we cannot apply the decorator to the argument. But, I think the contents of the deprecation massages should be same as that of the deprecation decorator. Do you think the ‘_DEFAULT_DEPRECATION_NOTE’ is not appropriate for the ‘step’ argument? (If we use it, the d_ver = deprecated version =2.0.0 and r_ver = removed version = 4.0.0.)

nzw0301:
@HideakiImamura Yes, it is. `_DEFAULT_DEPRECATION_NOTE` sounds like `IntlogUniformDistribution` will be remoed in the future...

HideakiImamura:
Oh! I see. How about giving the deprecation note not to whole of the class but only to the ‘step’ argument explanation like https://github.com/optuna/optuna/blob/master/optuna/pruners/_hyperband.py#L117? Then, users can understand that the ‘step’ argument will only be removed.

nzw0301:
I completely agree with you. I'm also thinking about your way 😎 .

nzw0301:
Related to https://github.com/optuna/optuna/pull/1329, I change the codes such that `step` is replaced with `1` because `suggest_int` does not accept `step != 1` when `log=True`. Internally, we do not need to hold `step` as an instance variable so just showing the warning message when `step != 1` in the constructor.

optuna/distributions.py
```diff
@@ -333,22 +334,22 @@ def __init__(self, low: int, high: int, step: int = 1) -> None:
...
             )
 
-        high = _adjust_int_uniform_high(low, high, step)
+        if step != 1:
+            warnings.warn(
+                "`step` accepts only `1`, so `step` is replaced with `1`. "
+                "`step` argument is deprecated and will be removed in the future. "
+                "The removal of this feature is currently scheduled for v4.0.0, "
+                "but this schedule is subject to change."
+            )
```
HideakiImamura, line 349: Please use the `optuna.logging` for logging as done in https://github.com/optuna/optuna/blob/master/optuna/pruners/_hyperband.py#L13.

nzw0301: Thank you for your comment! Like this? https://github.com/optuna/optuna/pull/1387/commits/0842bafaa6a2b001d3b49495bb7e69e0d1c8ffa5

optuna/distributions.py
```diff
@@ -311,19 +315,20 @@ class IntLogUniformDistribution(BaseDistribution):
...
             Upper endpoint of the range of the distribution. ``high`` is included in the range.
         step:
             A step for spacing between values.
+
+            .. note::
+                This value is valid for only 1. Otherwise, the value is replaced with 1.
+
+            .. deprecated:: 2.0.0
+                ``step`` argument will be removed in the future. The removal of this feature is
+                currently scheduled for v4.0.0, but this schedule is subject to change.
```
toshihikoyanase, line 331: Recently, #1407 introduced `.. warning::` instead of `.. deprecated::` to show the deprecation as follows:

image
([link](https://optuna.readthedocs.io/en/latest/reference/structs.html#optuna.structs.StudySummary))

Could you update the style of this message using `.. warning::`?

optuna/distributions.py
```diff
@@ -333,22 +338,22 @@ def __init__(self, low: int, high: int, step: int = 1) -> None:
...
             )
 
-        high = _adjust_int_uniform_high(low, high, step)
+        if step != 1:
+            _logger.warning(
+                "`step` accepts only `1`, so `step` is replaced with `1`. "
+                "`step` argument is deprecated and will be removed in the future. "
+                "The removal of this feature is currently scheduled for v4.0.0, "
+                "but this schedule is subject to change."
+            )
```
toshihikoyanase, line 353: Could you use the `warning.warn` to show the message? I think this warning can be fixed by users.
(c.f., [Logging HOWTO](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#logging-howto))

optuna/distributions.py
nzw0301: Thank you for your comment! Sure. The current part of `step` shows 
Screenshot 2020-06-25 at 0 24 36

optuna/distributions.py
nzw0301: Sure.  I should have use `warning.warn`.

optuna/distributions.py
```diff
@@ -333,22 +335,23 @@ def __init__(self, low: int, high: int, step: int = 1) -> None:
...
             )
 
-        high = _adjust_int_uniform_high(low, high, step)
+        if step != 1:
+            warnings.warn(
+                "`step` accepts only `1`, so `step` is replaced with `1`. "
+                "`step` argument is deprecated and will be removed in the future. "
+                "The removal of this feature is currently scheduled for v4.0.0, "
+                "but this schedule is subject to change.",
+                DeprecationWarning,
```
toshihikoyanase, line 350: I'm sorry for asking you a change again, but could you use `FutureWarning` instead of `DeprecationWarning`? #1428 introduced it to notify the warning to users.
```suggestion
                FutureWarning,
```

tests/test_distributions.py
```diff
@@ -370,4 +358,4 @@ def test_int_uniform_distribution_asdict():
 def test_int_log_uniform_distribution_asdict():
     # type: () -> None
 
-    assert EXAMPLE_DISTRIBUTIONS["ilu"]._asdict() == {"low": 2, "high": 12, "step": 2}
+    assert EXAMPLE_DISTRIBUTIONS["ilu"]._asdict() == {"low": 2, "high": 12}
```
toshihikoyanase, line 361: Could you add a test case to ensure that `IntLogUniform` raises `FutureWarning` when it receives the `step` argument?

```python

def test_int_log_uniform_distribution_deprecation() -> None:
    with pytest.warns(FutureWarning):
        distributions.IntLogUniformDistribution(low=1, high=100, step=2)
```

nzw0301: Thank you for suggesting codes! I have added the test.

hvy:
Sorry, I didn't mean to block. Thanks @nzw0301 for your contribution as always and let me merge this PR as it's already approved by two!

On a side note, the changes basically LGTM but I'm a bit worried that the deprecation is halfway in that the `step` attribute is actually removed and it'll break user-defined samplers. If you don't mind, I'd like to follow up this PR with putting it back with deprecation warnings raised.

nzw0301:
Thank your kindful reviews as usual! 

@hvy I appreciate your follow-up PR!

PR #1389, opened 2020-06-19
[RFC] Stop building Docker images in Pull Requests

crcrpar:
## Motivation

Docker build jobs on Pull Request pushes sometimes push down the CI job results which are more critical and required such as `tests-python37`.
Also, I think we can reckon that Docker build works if CI for Pull Requests pass as these two do almost exactly the same in the installation phase.

## Description of the changes

- Change the trigger of this job
- Remove unused environment variable
- Remove if statement

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -6,6 +6,10 @@ on:
       - master
   release:
     types: [published]
+  pull_request:
+    paths:
+      - Dockerfile
+      - setup.py
 
 env:
   DOCKER_HUB_BASE_NAME: optuna/optuna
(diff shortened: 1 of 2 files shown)


crcrpar:
First of all, much appreciate your comment.



> However, there may be some developers using the generated images, so I'm concerned that this might be an unfriendly change. Don't we need to investigate how the generated images are being used by developers?

As the workflow shows, currently, there is almost no way to pull docker images built for PRs' commits since we only push the docker images on push to master. So, I don't think we have to do it.

toshihikoyanase:
Thank you for your PR!
I'm a bit concerned about the review of `Dockerfile` and `setup.py`. Is it possible to run the docker build workflow if the PR contains the changes regarding `Dockerfile` and `setup.py`?

crcrpar:
> I'm a bit concerned about the review of `Dockerfile` and `setup.py`. Is it possible to run the docker build workflow if the PR contains the changes regarding `Dockerfile` and `setup.py`?

We should be able to according to https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-paths and gave it a try in 3ce0409.

.github/workflows/dockerimage.yml
```diff
@@ -7,11 +7,11 @@ on:
   release:
     types: [published]
   pull_request:
-    branches:
-      - master
+    paths:
+      - Dockerfile
+      - setup.py
```
toshihikoyanase, line 12: [Note] The trigger is not limited to the `master` branch, but I think it is acceptable because the built images are not uploaded to the DockerHub.

PR #1392, opened 2020-06-19
Fix for PyTorch Lightning v0.8.0.

crcrpar:
PyTorch Lightning v0.8.0 has some breaking changes:
1. `Trainer` module and the changes of `(min|max)_nb_epochs` to `(min|max)_epochss` affect our integration tests for it.
2. `EarlyStoppingCallback` is now triggered by `on_validation_end` instead of `epoch_end`.

## Description of the changes

- Fixes for the above two changes
- Support Python 3.8

__TODO__

- [x] enable test & example on Python 3.8

### DIFF
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -28,18 +28,15 @@ class PyTorchLightningPruningCallback(EarlyStopping):
             how this dictionary is formatted.
     """
 
-    def __init__(self, trial, monitor):
-        # type: (optuna.trial.Trial, str) -> None
+    def __init__(self, trial: optuna.trial.Trial, monitor: str) -> None:
 
         _imports.check()
 
         super(PyTorchLightningPruningCallback, self).__init__(monitor=monitor)
 
         self._trial = trial
-        self._monitor = monitor
 
-    def _process(self, trainer, pl_module):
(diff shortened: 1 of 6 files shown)


optuna/integration/pytorch_lightning.py
```diff
@@ -42,14 +42,24 @@ def __init__(self, trial, monitor):
         self._trial = trial
         self._monitor = monitor
 
-    def on_epoch_end(self, trainer, pl_module):
+    def _process(self, trainer, pl_module):
         # type: (Trainer, LightningModule) -> None
         logs = trainer.callback_metrics
         epoch = pl_module.current_epoch
-        current_score = logs.get(self._monitor)
+        current_score = logs.get(self.monitor)
```
toshihikoyanase, line 49: I think `self._monitor` is no longer used. Can we remove it?

crcrpar: I think so. I did this change without intention so I consulted PyTorch Lightning 0.7.1. I think we can remove `self._monitor`.
https://github.com/PyTorchLightning/pytorch-lightning/blob/0.7.1/pytorch_lightning/callbacks/early_stopping.py#L51

toshihikoyanase: Thank you for your investigation. Then, let's remove it to keep the code clean.

PR #1397, opened 2020-06-20
Fix broken link to example and add README for AllenNLP examples.

himkt:
This PR fixes the broken reference and adds an index page for AllenNLP examples.

- ef449ad Fix broken link and add a link to `allennlp_jsonnet.py`.
- 9f2178e7 Add README

### DIFF
--- a/examples/allennlp/README.md
+++ b/examples/allennlp/README.md
@@ -1,6 +1,6 @@
 # Hyperparameter optimization using Optuna
 
-This examples show two ways using Optuna for tuning AllenNLP models.
+These examples show two ways of using Optuna for tuning AllenNLP models.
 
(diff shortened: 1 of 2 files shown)


examples/allennlp/README.md
```diff
@@ -0,0 +1,6 @@
+# Hyperparameter optimization using Optuna
+
+This examples show two ways using Optuna for tuning AllenNLP models.
```
hvy, line 3: A nit
```suggestion
These examples show two ways of using Optuna for tuning AllenNLP models.
```

himkt: Thank you for correcting, I applied your suggestion. (a1abaa5)

PR #1333, opened 2020-06-06
Implement `AllenNLPPruningCallback`.

himkt:
close #1012.

## Motivation
This PR introduces a pruning callback for AllenNLP.

## Description of the changes
- add `AllenNLPPruningCallback`
- [WIP] bump up the version of AllenNLP
  - `AllenNLPPruningCallback` uses `epoch_callbacks` in the current implementation, which will be available on the next major version of `AllenNLP`.

## Memo
- [RFC] how to use `AllenNLPPruningCallback` in `AllenNLPExecutor`.

PR #1399, opened 2020-06-20
Implement PrunerCallback for AllenNLP.

### LINKED ISSUE #1012: [Proposal] Add allennlp integration
It would be nice to integrate with allennlp lib.

## Motivation

[Allennlp](https://github.com/allenai/allennlp) is a NLP library that is used in NLP researchers/practitioners including me.
There is [allentune](https://github.com/allenai/allentune), which is the library of parameter optimization for allennlp.
By using allentune, users can search hyperparameters by random search.
However, `optuna` provides some features which are not implemented in allentune (e.g. pruning).
I think it would be useful that allennlp users can use optuna to search for hyperparameters.

## Description

To support allennlp in optuna, I think we have to implement:

- Custom MetricTracker compatible with allennlp's Trainer
  - http://docs.allennlp.org/master/api/training/metric_tracker/#metrictracker
  - MetricTracker performs early stopping

- Custom Trainer that inherits allennlp's Trainer
  - latest allennlp doesn't provide a way to use custom callback (https://github.com/allenai/allennlp/issues/3519)
    - (this topic is under discussion, so it could change an approach
  - So we may implement custom trainer inherits `allennlp.training.Trainer`

I want to discuss this in detail.
Thanks,

himkt:
close #1012.
revive #1333.

## Motivation
This PR introduces a pruning callback for AllenNLP.

## Description of the changes
- Introduce `AllenNLPPruningCallback`
- Bump up the version of AllenNLP in `setup.py`
  - `AllenNLPPruningCallback` uses `epoch_callbacks` in the current implementation,
      which will be available on the next major version of `AllenNLP`.

## Version requirements of AllenNLP [added in 2020/06/25]

This PR changes allennlp_simple.py, which is an example of AllenNLP+Optuna writing a model in Python script. After being merged this PR, `allennlp_simple.py` and `test_allennlp.py` will require the latest major version of AllenNLP (`allennlp>=1.0.0`). This is because an example of AllenNLP will use `AllenNLPPruningCallback`, the wrapper of Optuna's pruner implementations, which enables you to prune unpromising trials using efficient algorithms such as Hyperband. `AllenNLPPruningCallback` requires `allennlp>=1.0.0`, as it uses `EpochCallback`, which is introduced in the major version of AllenNLP. If you don't have to use `AllenNLPPruningCallback`, you can use Optuna with both the latest and earlier versions of AllenNLP.

You may want to use Optuna with AllenNLP v0.9.x or older. `AllenNLPExecutor` would work with the latest and earlier versions of AllenNLP. So if you're going to use Optuna+AllenNLP with Jsonnet, you can use Optuna without concern.

## Memo
- [RFC] how to use `AllenNLPPruningCallback` in `AllenNLPExecutor`.

### DIFF
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -36,7 +36,7 @@ from optuna.integration import AllenNLPPruningCallback
 DEVICE = -1  # If you want to use GPU, use DEVICE = 0.
 MAX_DATA_SIZE = 3000
 MODEL_DIR = os.path.join(os.getcwd(), "result")
-TARGET_METRIC = "validation_accuracy"
+TARGET_METRIC = "accuracy"
 
 
 class SubsampledDatasetReader(allennlp.data.dataset_readers.TextClassificationJsonReader):
@@ -120,15 +120,15 @@ def objective(trial):
         optimizer=optimizer,
         data_loader=data_loader,
         validation_data_loader=validation_data_loader,
-        validation_metric="+accuracy",
+        validation_metric=f"+{TARGET_METRIC}",
         patience=None,
         num_epochs=50,
         cuda_device=DEVICE,
         serialization_dir=serialization_dir,
-        epoch_callbacks=[AllenNLPPruningCallback(trial, TARGET_METRIC)],
+        epoch_callbacks=[AllenNLPPruningCallback(trial, "validation_" + TARGET_METRIC)],
     )
     metrics = trainer.train()
-    return metrics["best_" + TARGET_METRIC]
+    return metrics["best_validation_" + TARGET_METRIC]
 
 
 if __name__ == "__main__":
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -121,7 +121,7 @@ def objective(trial):
         data_loader=data_loader,
         validation_data_loader=validation_data_loader,
         validation_metric="+" + TARGET_METRIC,
-        patience=None,
+        patience=None,  # `patience=None` since it could conflict with AllenNLPPruningCallback
         num_epochs=50,
         cuda_device=DEVICE,
         serialization_dir=serialization_dir,
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -140,7 +140,8 @@ if __name__ == "__main__":
     torch.manual_seed(41)
     numpy.random.seed(41)
 
-    study = optuna.create_study(direction="maximize")
+    pruner = optuna.pruners.HyperbandPruner()
+    study = optuna.create_study(direction="maximize", pruner=pruner)
     study.optimize(objective, n_trials=50, timeout=600)
 
     print("Number of finished trials: ", len(study.trials))
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -108,10 +108,10 @@ def objective(trial):
     optimizer = torch.optim.SGD(model.parameters(), lr=lr)
 
     data_loader = torch.utils.data.DataLoader(
-        train_dataset, batch_size=10, collate_fn=allennlp.data.allennlp_collate
+        train_dataset, batch_size=64, collate_fn=allennlp.data.allennlp_collate
     )
     validation_data_loader = torch.utils.data.DataLoader(
-        valid_dataset, batch_size=10, collate_fn=allennlp.data.allennlp_collate
+        valid_dataset, batch_size=64, collate_fn=allennlp.data.allennlp_collate
     )
 
     serialization_dir = os.path.join(MODEL_DIR, "trial_{}".format(trial.number))
@@ -137,7 +137,7 @@ if __name__ == "__main__":
     numpy.random.seed(41)
 
     study = optuna.create_study(direction="maximize")
-    study.optimize(objective, n_trials=80, timeout=600)
+    study.optimize(objective, n_trials=50, timeout=600)
 
     print("Number of finished trials: ", len(study.trials))
     print("Best trial:")
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -120,7 +120,7 @@ def objective(trial):
         optimizer=optimizer,
         data_loader=data_loader,
         validation_data_loader=validation_data_loader,
-        validation_metric=f"+{TARGET_METRIC}",
+        validation_metric="+" + TARGET_METRIC,
         patience=None,
         num_epochs=50,
         cuda_device=DEVICE,
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -104,7 +104,7 @@ def objective(trial):
     if DEVICE > -1:
         model.to(torch.device("cuda:{}".format(DEVICE)))
 
-    lr = trial.suggest_float("lr", 1e-1, 1e0, log=True)
+    lr = trial.suggest_float("lr", 1e-5, 1e-1, log=True)
     optimizer = torch.optim.SGD(model.parameters(), lr=lr)
 
     data_loader = torch.utils.data.DataLoader(
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -19,6 +19,7 @@ We have the following two ways to execute this example:
 """
 
 import os
+import pkg_resources
 import random
 import shutil
 
@@ -132,6 +133,9 @@ def objective(trial):
 
 
 if __name__ == "__main__":
+    if pkg_resources.parse_version(allennlp.__version__) < pkg_resources.parse_version("1.0.0"):
+        raise RuntimeError("AllenNLP>=1.0.0 is required for this example.")
+
     random.seed(41)
     torch.manual_seed(41)
     numpy.random.seed(41)
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -108,14 +108,10 @@ def objective(trial):
     optimizer = torch.optim.SGD(model.parameters(), lr=lr)
 
     data_loader = torch.utils.data.DataLoader(
-        train_dataset,
-        batch_size=10,
-        collate_fn=allennlp.data.allennlp_collate
+        train_dataset, batch_size=10, collate_fn=allennlp.data.allennlp_collate
     )
     validation_data_loader = torch.utils.data.DataLoader(
-        valid_dataset,
-        batch_size=10,
-        collate_fn=allennlp.data.allennlp_collate
+        valid_dataset, batch_size=10, collate_fn=allennlp.data.allennlp_collate
     )
 
     serialization_dir = os.path.join(MODEL_DIR, "trial_{}".format(trial.number))
@@ -124,12 +120,12 @@ def objective(trial):
         optimizer=optimizer,
         data_loader=data_loader,
         validation_data_loader=validation_data_loader,
-        validation_metric='+accuracy',
+        validation_metric="+accuracy",
         patience=None,
         num_epochs=50,
         cuda_device=DEVICE,
         serialization_dir=serialization_dir,
-        epoch_callbacks=[AllenNLPPruningCallback(trial, TARGET_METRIC)]
+        epoch_callbacks=[AllenNLPPruningCallback(trial, TARGET_METRIC)],
     )
     metrics = trainer.train()
     return metrics["best_" + TARGET_METRIC]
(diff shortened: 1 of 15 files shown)


himkt:
> TypeError: intercept_args() got an unexpected keyword argument 'multiprocessing_context'

CI still fails with the same message as https://github.com/optuna/optuna/pull/1333#issuecomment-645071219.
I can't reproduce this error on local and docker container and working on the investigation.

himkt:
I found a way to reproduce the result: testing allennlp with fastai.
Following command gives me the error.
`python -m pytest tests/integration_tests/test_fastai.py tests/integration_tests/allennlp/test_allennlp.py`

https://gist.github.com/himkt/938ece598b0b9cc7d3c29b84650edfee

himkt:
bc3c4e1  I wrote the workaround fix to restore original `DataLoader.__ini__`.

optuna/integration/allennlp.py
```diff
@@ -128,3 +130,45 @@ def run(self) -> float:
...
+
+    def __call__(
+        self,
+        trainer: "allennlp.training.GradientDescentTrainer",
+        metrics: Dict[str, Any],
+        epoch: int,
+        is_master: bool,
+    ) -> None:
+        value = metrics.get(self._monitor)
+        if not value:
```
crcrpar, line 169: Not sure, `value` can be `None`, right?
```suggestion
        if value is None:
```

examples/allennlp/allennlp_simple.py
```diff
@@ -19,59 +19,66 @@
...
 
-GLOVE_FILE_PATH = "https://s3-us-west-2.amazonaws.com/allennlp/datasets/glove/glove.6B.50d.txt.gz"
+class SubsampledDatasetReader(allennlp.data.dataset_readers.TextClassificationJsonReader):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def _read(self, datapath):
+        data = list(super()._read(datapath))
+        random.shuffle(data)
+        yield from data[:3000]
```
crcrpar, line 49:
```suggestion
        yield from data[:MAX_DATA_SIZE]
```

examples/allennlp/allennlp_simple.py
```diff
@@ -19,59 +19,66 @@
...
 MAX_DATA_SIZE = 3000
+MODEL_DIR = os.path.join(os.getcwd(), "result")
+TARGET_METRIC = "validation_accuracy"
 
-DIR = os.getcwd()
-MODEL_DIR = os.path.join(DIR, "result")
 
-GLOVE_FILE_PATH = "https://s3-us-west-2.amazonaws.com/allennlp/datasets/glove/glove.6B.50d.txt.gz"
+class SubsampledDatasetReader(allennlp.data.dataset_readers.TextClassificationJsonReader):
+    def __init__(self, **kwargs):
```
crcrpar, line 43: I'm not sure `**kwargs` is OK. Request for other reviewers comments.

examples/allennlp/allennlp_simple.py
```diff
@@ -19,59 +19,66 @@
...
 import torch
 
 import optuna
+from optuna.integration import AllenNLPPruningCallback
 
 
 DEVICE = -1  # If you want to use GPU, use DEVICE = 0.
 MAX_DATA_SIZE = 3000
+MODEL_DIR = os.path.join(os.getcwd(), "result")
+TARGET_METRIC = "validation_accuracy"
```
crcrpar, line 39: If you don't use this variable other than L.131, we don't need to define this IMO.

examples/allennlp/allennlp_simple.py
```diff
@@ -97,29 +104,38 @@ def objective(trial):
...
     )
     metrics = trainer.train()
-    return metrics["best_validation_accuracy"]
+    return metrics["best_" + TARGET_METRIC]
 
 
 if __name__ == "__main__":
+    random.seed(41)
+    torch.manual_seed(41)
+    numpy.random.seed(41)
```
crcrpar, line 137: Could you tell me what numpy features with randomness used in this example?

examples/allennlp/allennlp_simple.py
```diff
@@ -97,29 +104,38 @@ def objective(trial):
...
         optimizer=optimizer,
-        iterator=iterator,
-        train_dataset=train_dataset,
-        validation_dataset=valid_dataset,
-        patience=3,
-        num_epochs=6,
+        data_loader=data_loader,
+        validation_data_loader=validation_data_loader,
+        validation_metric="+accuracy",
+        patience=None,
```
crcrpar, line 124: [question] Why `3` to `None`?

setup.py
```diff
@@ -74,12 +74,12 @@ def get_extras_require() -> Dict[str, List[str]]:
...
             "thop",
-            "torch==1.4.0" if sys.platform == "darwin" else "torch==1.4.0+cpu",
-            "torchvision==0.5.0" if sys.platform == "darwin" else "torchvision==0.5.0+cpu",
+            "torch==1.5.1" if sys.platform == "darwin" else "torch==1.5.1+cpu",
+            "torchvision==0.6.1" if sys.platform == "darwin" else "torchvision==0.6.1+cpu",
             "xgboost",
         ]
         + (
-            ["allennlp<1", "fastai<2", "pytorch_lightning>=0.7.1"]
+            ["allennlp==1.0.0", "fastai<2", "pytorch-lightning>=0.7.1"]
```
crcrpar, line 82: What's the difference between "pytorch_lightning" and "pytorch-lightning"?

himkt: Thank you for pointing out, I unintentionally introduced this change. I removed it in 7143a3e.

BTW, I found `pytorch-lightning` and `pytorch_lightning` both appear in `setup.py`.
(I tested `pip install` them and confirmed it downloaded `pytorch_lightning` in both cases)
https://github.com/optuna/optuna/pull/1392/commits/f53cbe225032bed25d61a06db0f5250243d3d348#diff-2eeaed663bd0d25b7e608891384b7298R82-R88

According to PyPI, `pytorch-lightning` seems to be right.
https://pypi.org/project/pytorch-lightning/

optuna/integration/allennlp.py
himkt: You're right. Thank you!

examples/allennlp/allennlp_simple.py
himkt: `GradientDescentTrainer` works with early stopping by default.
I update a value of patience, as early stopping could conflict with `AllenNLPPruningCallback`.
(Thank you for the question, I put a comment since it should be mentioned in a script)

examples/allennlp/allennlp_simple.py
himkt: Good question.
Actually, I'm not sure whether I should set seed for NumPy,
but allennlp does in `allennlp train` command and I followed it.
https://github.com/allenai/allennlp/blob/master/allennlp/common/util.py#L218-L247

Do you think it could be removed?

examples/allennlp/allennlp_simple.py
himkt: It is also used in L128. (https://github.com/optuna/optuna/pull/1399/files#diff-68d489e90782fdeb5d3d9d0274162956R128)

3224ad7ec
And I revised the PR to use `TARGET_METRIC` in `trainer` as well.

examples/allennlp/allennlp_simple.py
crcrpar: If AllenNLP does, I think it's okay, we don't need to remove.

himkt:
Thank you for pointing out, I remove literal string interpolation in e0dc445.

himkt:
I revised my PR, PTAL again. 🙏

setup.py
crcrpar: I side for `pytorch-lightning`. Thanks for checking PyPI.

examples/allennlp/allennlp_simple.py
crcrpar: Much appreciate your clear explanation.

himkt:
I rebased the master branch. (crcrpar gave my PR the look until 355a402 and 7d0e6ae is the commit simply putting the comment)

himkt:
I merged the master branch.
@toshihikoyanase kindly ping. 🙋‍♂️

examples/allennlp/allennlp_simple.py
```diff
@@ -19,59 +19,66 @@
...
 
 import allennlp
 import allennlp.data
 import allennlp.models
 import allennlp.modules
+import numpy
 import torch
 
 import optuna
+from optuna.integration import AllenNLPPruningCallback
```
toshihikoyanase, line 35: I think we can add a version check here since this example raises errors with `AllenNLP < 1.0.0`.

examples/allennlp/allennlp_simple.py
```diff
@@ -100,26 +107,35 @@ def objective(trial):
...
-    return metrics["best_validation_accuracy"]
+    return metrics["best_validation_" + TARGET_METRIC]
 
 
 if __name__ == "__main__":
+    random.seed(41)
+    torch.manual_seed(41)
+    numpy.random.seed(41)
+
     study = optuna.create_study(direction="maximize")
```
toshihikoyanase, line 139: I encountered nan-value errors during the training of AllenNLP model. This seems to happen stochastically, so how about fixing the seed value of `TPESampler`?

```
File "examples/allennlp/allennlp_simple.py", line 130, in objective
metrics = trainer.train()
File "/Users/yanase/.pyenv/versions/optuna-dev/lib/python3.7/site-packages/allennlp/training/trainer.py", line 802, in train
train_metrics = self._train_epoch(epoch)
File "/Users/yanase/.pyenv/versions/optuna-dev/lib/python3.7/site-packages/allennlp/training/trainer.py", line 569, in _train_epoch
raise ValueError("nan loss encountered")
ValueError: nan loss encountered
```

```suggestion
    study = optuna.create_study(direction="maximize", sampler=optuna.samplers.TPESampler(seed=41))
```

examples/allennlp/allennlp_simple.py
toshihikoyanase: We can check the version, for example, as follows:
```python
import pkg_resources

if pkg_resources.parse_version(allennlp.__version__) < pkg_resources.parse_version("1.0.0"):
    raise RuntimeError("AllenNLP>=1.0.0 is required for this example.")

```

examples/allennlp/allennlp_simple.py
```diff
@@ -19,59 +19,66 @@
...
 
-GLOVE_FILE_PATH = "https://s3-us-west-2.amazonaws.com/allennlp/datasets/glove/glove.6B.50d.txt.gz"
+class SubsampledDatasetReader(allennlp.data.dataset_readers.TextClassificationJsonReader):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def _read(self, datapath):
+        data = list(super()._read(datapath))
+        random.shuffle(data)
+        yield from data[:MAX_DATA_SIZE]
```
toshihikoyanase, line 49: How about slicing the dataset before shuffling it? We can use the same dataset between epochs to assess validation accuracy.

```suggestion
        data = list(super()._read(datapath))[:MAX_DATA_SIZE]
        random.shuffle(data)
        yield from data
```

tests/integration_tests/allennlp_tests/example_with_include_package.jsonnet
```diff
@@ -49,8 +49,7 @@ local DROPOUT = std.extVar('DROPOUT');
       bidirectional: true,
     },
   },
-  iterator: {
-    type: 'basic',
+  data_loader: {
```
toshihikoyanase, line 52: `example_with_environment_variables.jsonnet` also has the `iterator` key. I think we should replace it with `data_loader` too.

optuna/integration/allennlp.py
```diff
@@ -147,3 +149,45 @@ def run(self) -> float:
...
+    if you want to add a proning callback which observes a metric.
+
+    Args:
+        trial:
+            A :class:`~optuna.trial.Trial` corresponding to the current evaluation of the
+            objective function.
+        monitor:
+            An evaluation metric for pruning, e.g. ``validation_loss`` or
+            ``validation_accuracy``.
+    """
```
toshihikoyanase, line 169: Could you add `AllenNLPPruningCallback` to the list of references (i.e., `docs/source/reference/integration.rst`)?

examples/allennlp/allennlp_simple.py
toshihikoyanase: I think it's acceptable to keep the example simple. We can revisit it if we have problems.

toshihikoyanase:
If I pass [`-s` option](https://docs.pytest.org/en/stable/capture.html#accessing-captured-output-from-a-test-function) to pytest to disable capturing, pytest works as expected (see [872406225 log](https://github.com/toshihikoyanase/optuna/runs/872406225?check_suite_focus=true)). It may affect `capsys` features, so how about separating the test process as follows:
```yaml
          # pytest cannot collect allennlp tests with allennlp==1.0.0.
          pytest tests/integration_tests --ignore integration_tests/allennlp_tests/test_allennlp.py
          pytest -s integration_tests/allennlp_tests/test_allennlp.py
```

PR #1403, opened 2020-06-22
Skip storage access if logging is disabled.

toshihikoyanase:
## Motivation

Currently, `optuna.study.Study._log_completed_trial` accesses storage to fetch the best trial even if the logger is disabled. This can be a performance limitation if users want to stop the logger to reduce the optimization time.

## Description of the changes

This PR skip the storage access if the logger is disabled.
I assume that the logger is disabled if the following conditions are satisfied:
- the logging level is higher than `optuna.logging.INFO`, or
- both the default handler and the propagation option are disabled.

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -755,8 +755,6 @@ class Study(BaseStudy):
 
         if not _logger.isEnabledFor(logging.INFO):
             return
-        if not (logging._is_enabled_default_handler() or logging._is_enabled_propagation()):
-            return
 
         _logger.info(
             "Trial {} finished with value: {} and parameters: {}. "
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -753,7 +753,7 @@ class Study(BaseStudy):
 
     def _log_completed_trial(self, trial: trial_module.Trial, result: float) -> None:
 
-        if logging.get_verbosity() > logging.INFO:
+        if not _logger.isEnabledFor(logging.INFO):
             return
         if not (logging._is_enabled_default_handler() or logging._is_enabled_propagation()):
             return
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -751,8 +751,12 @@ class Study(BaseStudy):
 
         return trial
 
-    def _log_completed_trial(self, trial, result):
-        # type: (trial_module.Trial, float) -> None
+    def _log_completed_trial(self, trial: trial_module.Trial, result: float) -> None:
+
+        if logging.get_verbosity() > logging.INFO:
+            return
+        if not (logging._is_enabled_default_handler() or logging._is_enabled_propagation()):
+            return
 
         _logger.info(
             "Trial {} finished with value: {} and parameters: {}. "
--- a/tests/test_study.py
+++ b/tests/test_study.py
@@ -1020,16 +1020,41 @@ def test_log_completed_trial(capsys: _pytest.capture.CaptureFixture) -> None:
     optuna.logging.set_verbosity(optuna.logging.INFO)
 
     study = optuna.create_study()
-    study.optimize(lambda t: 1.0, n_trials=1)
+    study.optimize(lambda _: 1.0, n_trials=1)
     _, err = capsys.readouterr()
     assert "Trial 0" in err
 
     optuna.logging.set_verbosity(optuna.logging.WARNING)
-    study.optimize(lambda t: 1.0, n_trials=1)
+    study.optimize(lambda _: 1.0, n_trials=1)
     _, err = capsys.readouterr()
     assert "Trial 1" not in err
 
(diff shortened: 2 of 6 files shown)


hvy:
Not very exhaustive but let me share some quick benchmarks.

**master**
`INFO`: 2.544203996658325 s
`WARNING`: 2.3177778720855713 s

**This PR**
`INFO`: 2.523770809173584 s
`WARNING`: 1.2488319873809814 s

```python
import time

import optuna
from optuna import logging
from optuna.samplers import RandomSampler
import sqlalchemy


    return sum(trial.suggest_float(f'x{i}', 0.0, 1.0) for i in range(n_params))


if __name__ == "__main__":
    n_params = 10
    n_trials = 100

    for verbosity in [logging.INFO, logging.WARNING]:
        storage="mysql://root@localhost/logging"
        database="logging"
        engine = sqlalchemy.create_engine(storage)
        conn = engine.connect()
        conn.execute("commit")
        try:
            conn.execute("drop database {}".format(database))
        except:
            pass
        conn.execute("create database {}".format(database))
        conn.close()

        study = optuna.create_study(
            sampler=RandomSampler(),
            study_name=database,
            storage=storage,
        )

        logging.set_verbosity(verbosity)

        start = time.time()
        end = time.time()

        print(verbosity, end - start)
```

optuna/study.py
```diff
@@ -751,8 +751,12 @@ def _run_trial(
 
         return trial
 
-    def _log_completed_trial(self, trial, result):
-        # type: (trial_module.Trial, float) -> None
+    def _log_completed_trial(self, trial: trial_module.Trial, result: float) -> None:
+
+        if logging.get_verbosity() > logging.INFO:
```
hvy, line 756: We could also use https://docs.python.org/3/library/logging.html#logging.Logger.isEnabledFor. It internally uses a cache (https://github.com/python/cpython/blob/26a1ad1c24717990265b71ed093d691500d6301c/Lib/logging/__init__.py#L1695).
```suggestion
        if not _logger.isEnabledFor(logging.INFO):
```

toshihikoyanase:
@hvy Thank you for sharing the benchmark results. This PR halved the execution time!

> This PR
> `INFO`: 2.523770809173584 s
> `WARNING`: 1.2488319873809814 s

optuna/study.py
hvy, line 756: How about encapsulating this entire logic in `optuna.logging._is_enabled_for(logger, level)`, or something, since it's used in the MO study as well?

tests/test_study.py
```diff
@@ -1010,3 +1012,58 @@ def test_study_summary_lt_le():
     summaries.sort()
     assert summaries[0] == summary_0
     assert summaries[1] == summary_1
+
+
+def test_log_completed_trial_verbosity(capsys: _pytest.capture.CaptureFixture) -> None:
```
hvy, line 1017: Thanks for adding those tests! They doesn't exactly verify the introduced changes to the logic but I think they are good to keep.

optuna/study.py
```diff
@@ -751,8 +751,12 @@ def _run_trial(
...
+            return
+        if not (logging._is_enabled_default_handler() or logging._is_enabled_propagation()):
+            return
```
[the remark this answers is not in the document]
toshihikoyanase: Thank you for your careful review. The progress bar was completely out of my mind.
Although the progress bar works with the following `optuna.logging._is_enabled_default_hander`, I think checking the handler/propagation settings makes logic too complicated. I'll only check the log level to skip the storage access, and remove the `optuna.logging._is_enabled_default_hander()` and `optuna.logging._is_enabled_propagation()`.

I think it is reasonable because the [FAQ](https://optuna.readthedocs.io/en/stable/faq.html#how-to-suppress-log-messages-of-optuna) suggests `optuna.logging.set_verbosity` to suppress the log messages.

```
def _is_enabled_default_handler() -> bool:
    _configure_library_root_logger()

    assert _default_handler is not None
    return len(_get_library_root_logger().handlers) > 0
```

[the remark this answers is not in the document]
hvy: I'm also for keeping it simple in this PR. It should be enough for the majority of users.

hvy:
Thanks for you quick fixes. Changes LGTM but could you just fix the style errors in the CI?

tests/test_study.py
toshihikoyanase: Thank you for your comment. I also added the logic which skips storage access in https://github.com/optuna/optuna/pull/1403/commits/22ccbff53f36292507c2084edad4f26222ab32b0. I noticed that `_log_completed_trial` accessed twice to get the best trial. I think we can reduce it by reusing the first one, but it may be out of the scope of this PR.

Issue #1383, opened 2020-06-17
Add `last-updated-time` to RDB schema.


ytsmiling:
This issue proposes to add a `last_updated_time` column to the RDB schema (`trials` table).

## Motivation

The current implementation of optuna RDB storage classes requires to perform O(trial) communication per each `get_all_trials` call. To reduce the communication cost to O(worker) on average (during the optimization process), each storage instance needs to know which trials are created/updated after the last `get_all_trials` call. Even though optuna's trial has `datetime_start` and `datetime_complete`, they are registered on the client-side and cannot be used for synchronization purposes.

## Description

Add `last_updated_time` column (naming may need discussion) to the `trials` table, and add the information to `alembic`. The `last_updated_time` needs to be updated using the server-side timestamp at least when the trial is created. Then, `get_all_trials` only need to fetch (running trials + trials created after the last `get_all_trials` call), which should be O(worker).

## Additional context

* This change requires schema change, and major release will be the right time for this change if optuna accepts this proposal.
* `#get_all_trials call` was reduced to O(1) from O(param) per trial by #1298, and thus the communication cost is not that problem compared to the past.
* The current O(trial) communications can be a problem when optuna support O(1) execution of random-sampler and cma-es sampler in the future, but currently they're not.

hvy:
Please see #1405 and #1426 for why this issue has become stale. Let me label this as `needs-discussion`.

hvy:
This issue has potential and I'd very much like to tackle it, but let me close it for now as it has become stale and I'm not confident I'll be able to tackle it in the near future. Let's reopen it as storage performance becomes relevant again.

PR #1405, opened 2020-06-22
Add last-update column to `TrialModel`.

hvy:
## Motivation

Fixes https://github.com/optuna/optuna/issues/1383.

## Description of the changes

Adds a last-update timestamp column to `TrialModel`. It's named `datetime_last_update` to align with `datetime_{start,complete}` but it might be controversial this this columns is updated with the server timestamp.


optuna/storages/_rdb/alembic/versions/v2.0.0.a.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = "v2.0.0.a"
+down_revision = "v1.3.0.a"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("trials", sa.Column("datetime_last_update", sa.DateTime(), nullable=True))
```
c-bata, line 21: How about setting `datetime.now` as a default value instead of `nullable=True`?

hvy: Thanks, that sounds reasonable. I used the server timestamp so that it's aligned to the on-UPDATE logic. PTAL.

optuna/storages/_rdb/alembic/versions/v2.0.0.a.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "trials",
+        sa.Column(
+            "datetime_last_update",
+            sa.DateTime(),
+            server_default=sa.text("(CURRENT_TIMESTAMP)"),
```
c-bata, line 26: In terms of portability, it is prefer to pass `sa.func.now()` at server_default. I guess `sa.text("(CURRENT_TIMESTAMP)")` doesn't properly work on some databases.

```python
from alembic import op
from sqlalchemy import Column, TIMESTAMP, func

# specify "DEFAULT NOW" along with the column add
op.add_column('account',
    Column('timestamp', TIMESTAMP, server_default=func.now())
)
```

https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.Operations.add_column

optuna/storages/_rdb/alembic/versions/v2.0.0.a.py
hvy: SQLite's `ALTER TABLE` doesn't seem to allow `CURRENT_TIMESTAMP` as a default value, so we'll have to modify `upgrade`. I'm not sure we can do that while keeping the same DDL.

```
E       sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) Cannot add a column with non-constant default
E       [SQL: ALTER TABLE trials ADD COLUMN datetime_last_update DATETIME DEFAULT (CURRENT_TIMESTAMP) NOT NULL]
E       (Background on this error at: http://sqlalche.me/e/e3q8)
```

optuna/storages/_rdb/alembic/versions/v2.0.0.a.py
hvy: Thanks for your suggestion. `CURRENT_TIMESTAMP` was automatically generated from the changes to the model using `func.now` but if we can use `func.now()` in the migration script we probably should. 

This problem remains however, https://github.com/optuna/optuna/pull/1405#discussion_r444093388 and I haven't found a proper way to address it. I'm wondering if we should allow `NULL` and populate the column manually in this migration script.

optuna/storages/_rdb/alembic/versions/v2.0.0.a.py
hvy: Addressed with batch updates. https://alembic.sqlalchemy.org/en/latest/batch.html

optuna/storages/_rdb/alembic/versions/v2.0.0.a.py
hvy: Addressed. See https://github.com/optuna/optuna/pull/1405#discussion_r444710090.

hvy:
PTAL.

hvy:
Thanks for the quick review! I'd like to verify the expected behavior with https://github.com/optuna/optuna/pull/1426 if possible. I think we can do that before v2.

c-bata:
@hvy Should I wait to merge this PR until #1426 's behavior is verified?

hvy:
Thanks for your quick review and verification. I'd like to wait a bit since I don't think that this schema change alone will allow incremental load. Let me leave a note in #1426 and mark this PR as draft.

> [memo] I think datetime_start and datetime_complete should be set at database-level timestamp.

Agree. I noticed however that `sqlalchemy.func.now` gets rid of the millisecond units, as you also noticed above in the `datetime_last_update` column. Maybe that's acceptable for these columns.

hvy:
Closing this issue for now.

PR #1410, opened 2020-06-23
Prevent `doc-link` from running on unrelated status update events

harupy:
## Motivation

Prevnet [`doc-link`](https://github.com/optuna/optuna/blob/master/.github/workflows/doc-link.yml) from running on unrelated status update events to save [GitHub Actions usage](https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits).

## Description of the changes

Add an `if` condition to prevent `doc-link` from running on unrelated events.

### DIFF
--- a/.github/workflows/doc-link.yml
+++ b/.github/workflows/doc-link.yml
@@ -5,6 +5,7 @@ on: status
 jobs:
   doc-link:
     runs-on: ubuntu-latest
+    if: contains(github.context.target_url, 'circle-artifacts.com')
     steps:
     - uses: larsoner/circleci-artifacts-redirector-action@master
       with:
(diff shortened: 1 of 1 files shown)


harupy:
I have tested this on my repo:

Screen Shot 2020-06-23 at 20 14 21

https://github.com/harupy/mlflow-extend/actions

harupy:
The current condition is wrong. I'll fix it.

.github/workflows/doc-link.yml
```diff
@@ -5,6 +5,7 @@ on: status
 jobs:
   doc-link:
     runs-on: ubuntu-latest
+    if: "github.event.context == 'ci/circleci: document'"
```
harupy, line 8: Wrapped the condition string with `""` otherwise GitHub raise a syntax error.

PR #1370, opened 2020-06-15
Update layout and color of docs.

toshihikoyanase:
## Motivation
Currently, Optuna docs employ the default `sphinx_rtd_theme`. I think it is quite different from the style of https://optuna.org/.

c.f. The current docs: https://optuna.readthedocs.io/en/stable/index.html

image

## Description of the changes

This PR changes the layout and color of Optuna docs.
- Add the logo at the top left of the page.
- Add the navigation bar at the top of the page.
- Use color pallet of https://optuna.org/

image

image

image

c.f. https://optuna.org

image

PR #1414, opened 2020-06-24
Always use Sphinx RTD theme.

hvy:
## Motivation

Follow-up to https://github.com/optuna/optuna/pull/1370. The `logo_only` should take effect but it isn't.

## Description of the changes

Changes to always, incl. building on RTD, apply the RTD Sphinx Theme. This makes sure that `logo_only` takes effect on the deployed RTD environment.

### Note
- The removed condition was initially introduced in https://github.com/optuna/optuna/pull/191 when changing the theme and I'm not aware of the underlying reason but it seems like we can remove it now.
- Preview https://hvy-optuna.readthedocs.io/en/rtd-sphinx-test/index.html

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -16,11 +16,9 @@
 # import sys
 # sys.path.insert(0, os.path.abspath('.'))
 
-import os
 import pkg_resources
 
 __version__ = pkg_resources.get_distribution('optuna').version
-on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 
 # -- Project information -----------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


hvy:
Oh, I wasn't aware of the fact that it was based on Chainer. I guess it was this change https://github.com/chainer/chainer/commit/4ffecaf1fa6ab05e437df7fdd258c1f6265e0a7a. Guessing from the commit message "Support deploying to Read the Docs", maybe it used to be required for RTD deployment, but no longer. 🤔

toshihikoyanase:
>  maybe it used to be required for RTD deployment, but no longer. 🤔

I think so too. In 2015, the version of `sphinx_rtd_theme` was 0.1.8, and its [README](https://github.com/readthedocs/sphinx_rtd_theme/tree/0.1.8) suggested to remove `sphinx_rtd_theme` in the RTD environment.

> Currently if you import sphinx_rtd_theme in your local sphinx build, then pass that same config to Read the Docs, it will fail, since RTD gets confused. If you want to run this theme locally and then also have it build on RTD, then you can add something like this to your config. Thanks to Daniel Oaks for this.

We cannot see such instructions in the [README](https://github.com/readthedocs/sphinx_rtd_theme/tree/0.5.0) of the current version (0.5.0).

hvy:
Thank you for your investigation, that explains it. 🙂

PR #1413, opened 2020-06-24
Apply deprecation decorator.

toshihikoyanase:
## Motivation
Followup #1382. 

## Description of the changes

This PR applies the deprecation decorator to the following classes:

- `optuna.integratoin.CmaEsSampler`
- `optuna.struct.FrozenTrial`
- `optuna.struct.StudySummary`
- `optuna.struct.TrialPruned`

PR #1418, opened 2020-06-24
Allow omitting the removal version in `deprecated`.

hvy:
## Motivation

Better to avoid unnecessary verbosity and redundancy. Noticed while reviewing https://github.com/optuna/optuna/pull/1413 that the deprecation decorator could be improved.

## Description of the changes

Allows omitting the removal version when applying the deprecation decorator.

### DIFF
--- a/tests/test_deprecated.py
+++ b/tests/test_deprecated.py
@@ -31,7 +31,7 @@ class _Sample(object):
 
 
 @pytest.mark.parametrize("deprecated_version", ["1.1", 100, None, "2.0.0"])
-@pytest.mark.parametrize("removed_version", ["1.1", 10, None, "1.0.0"])
+@pytest.mark.parametrize("removed_version", ["1.1", 10, "1.0.0"])
 def test_deprecation_raises_error_for_invalid_version(
     deprecated_version: Any, removed_version: Any
 ) -> None:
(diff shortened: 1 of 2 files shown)


tests/test_deprecated.py
```diff
@@ -155,3 +155,27 @@ def __init__(self, a: Any, b: Any, c: Any) -> None:
...
     assert decorated_class.__name__ == _Class.__name__
     assert decorated_class.__doc__ == expected_class_doc
+
+
+def test_deprecation_decorator_default_removed_version() -> None:
+    deprecated_version = "1.1.0"
+    decorator_deprecation = _deprecated.deprecated(deprecated_version)
+    assert callable(decorator_deprecation)
+
+    def _func(_: Any) -> int:
```
toshihikoyanase, line 165: Do you think the argument _ is required? It is not introduced by this PR but I'm just curious about it.

hvy: You're right. The argument is not relevant to the test. I think we can fix that in a follow up, along with the other tests in this file as well as `test_experimental.py`.

toshihikoyanase: Thank you for your reply.

> I think we can fix that in a follow up, along with the other tests in this file as well as `test_experimental.py`.

I agree. Let me merge this PR.

hvy: https://github.com/optuna/optuna/pull/1423

PR #1420, opened 2020-06-24
Add Stable-Baselines3 RL Example.

### LINKED ISSUE #1314: Stable-Baselines3 RL Example
Hello,

I was wondering if you would be interested in adding an example with Optuna + Stable-Baselines3 for hyperparameter tuning in an reinforcement learning context?

It has been used successfully in both v2 and v3 in the zoo repo:
- https://github.com/DLR-RM/rl-baselines3-zoo
- https://github.com/araffin/rl-baselines-zoo

Stable Baselines:
- https://github.com/hill-a/stable-baselines (v2, tensorflow)
- https://github.com/DLR-RM/stable-baselines3 (v3, pytorch edition)

## Motivation

Hyperparameter tuning is key to RL success and Stable Baselines (SB and SB3) is one of the most popular RL library.

araffin:
Note: the only thing missing now is to deactivate tests for python < 3.6 (I don't know where that should be changed)

## Motivation

closes #1314 

## Description of the changes

Add an hyperparameter tuning example in an reinforcement learning context using Stable-Baselines3.


araffin:
@HideakiImamura where is the flake8 config to check the style locally?

I've been using `flake8 .` at the root of the project but it does not show the errors that CircleCI shows...

examples/rl/sb3_simple.py
```diff
@@ -0,0 +1,165 @@
...
+    n_steps = trial.suggest_categorical("n_steps", [8, 16, 32, 64, 128, 256, 512, 1024])
+    learning_rate = trial.suggest_loguniform("lr", 1e-5, 1)
+    ent_coef = trial.suggest_loguniform("ent_coef", 0.00000001, 0.1)
+    ortho_init = trial.suggest_categorical("ortho_init", [False, True])
+    net_arch = trial.suggest_categorical("net_arch", ["tiny", "small"])
+    activation_fn = trial.suggest_categorical("activation_fn", ["tanh", "relu"])
+
+    net_arch = {"tiny": [dict(pi=[64], vf=[64])], "small": [dict(pi=[64, 64], vf=[64, 64])],}[
+        net_arch
+    ]
```
crcrpar, line 56: As to dict, literal is preferrable because for example (1) we can declare a variable named dict and (2) dict() func is a bit slower than literal. So I would write like
```suggestion
    net_arch = {"pi": [64], "vf": [64],} if net_arch == "tiny" else {"pi": [64, 64], "vf": [64, 64],}
```

examples/rl/sb3_simple.py
```diff
@@ -0,0 +1,165 @@
...
+    return {
+        "n_steps": n_steps,
+        "gamma": gamma,
+        "gae_lambda": gae_lambda,
+        "learning_rate": learning_rate,
+        "ent_coef": ent_coef,
+        "max_grad_norm": max_grad_norm,
+        "policy_kwargs": dict(
+            net_arch=net_arch, activation_fn=activation_fn, ortho_init=ortho_init
+        ),
```
crcrpar, line 69: Same here. I would use dict literal.
```suggestion
        "policy_kwargs": {
            "net_arch": net_arch,
            "activation_fn": activation_fn,
            "ortho_init": ortho_init,
        },
```

examples/rl/sb3_simple.py
```diff
@@ -0,0 +1,165 @@
...
+
+    sampler = TPESampler(n_startup_trials=N_STARTUP_TRIALS)
+    # Do not prune before 1/3 of the max budget is used
+    pruner = MedianPruner(n_startup_trials=N_STARTUP_TRIALS, n_warmup_steps=N_EVALUATIONS // 3)
+
+    study = optuna.create_study(sampler=sampler, pruner=pruner, direction="maximize")
+    try:
+        study.optimize(objective, n_trials=N_TRIALS, n_jobs=N_JOBS)
+    except KeyboardInterrupt:
+        pass
```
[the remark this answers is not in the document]
araffin: We can have both, no?
This is more intended to create the report even when the user kills the optimization early.
The timeout would be more for the tests, no?

araffin:
Thanks for your comments, I added your suggestions.
One last thing that I did not find: how do you add that example to CI tests?

In the past, there was a stage but it was removed recently apparently.

crcrpar:
> One last thing that I did not find: how do you add that example to CI tests?
> 
> In the past, there was a stage but it was removed recently apparently.

Good catch! Currently, example runs are daily and not checked in PR's CI.

https://github.com/optuna/optuna/blob/e249aa8e2587f57e6b11b9341cca4e1a4b0d964a/.github/workflows/examples.yml#L3-L5

So, I think what we need to do is to add your example to `IGNORES`.

https://github.com/optuna/optuna/blob/e249aa8e2587f57e6b11b9341cca4e1a4b0d964a/.github/workflows/examples.yml#L38-L39

examples/rl/sb3_simple.py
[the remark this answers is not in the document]
crcrpar: You are absolutely right and what you do, i.e., the use of `KeyboardInterrupt` is really cool.

That being said, it's embarrassing to say this though, could you set `timeout` for the faster CI runs like this?
https://github.com/optuna/optuna/blob/e249aa8e2587f57e6b11b9341cca4e1a4b0d964a/examples/pytorch_simple.py#L136

HideakiImamura:
@araffin Thanks for the PR! The introduced example looks very impressive. I would like to review this PR after you addressing above @hvy's reviews. Thanks!

examples/rl/sb3_simple.py
```diff
@@ -0,0 +1,165 @@
...
+        self.eval_idx = 0
+        self.is_pruned = False
+
+    def _on_step(self) -> bool:
+        if self.eval_freq > 0 and self.n_calls % self.eval_freq == 0:
+            super(TrialEvalCallback, self)._on_step()
+            self.eval_idx += 1
+            self.trial.report(self.last_mean_reward, self.eval_idx)
+            # Prune trial if need
+            if self.trial.should_prune(self.eval_idx):
```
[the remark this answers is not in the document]
araffin: thanks, I think I wrote that code one year and half ago... so things have changed a bit ;)

examples/rl/sb3_simple.py
```diff
@@ -0,0 +1,170 @@
...
+    "env": ENV_ID,
+}
+
+
+def sample_a2c_params(trial: optuna.Trial) -> Dict[str, Any]:
+    """Sampler for A2C hyperparameters."""
+    gamma = trial.suggest_categorical("gamma", [0.9, 0.95, 0.98, 0.99, 0.999, 0.9999])
+    max_grad_norm = trial.suggest_categorical("max_grad_norm", [0.3, 0.5, 0.7, 0.9, 1, 2, 5])
+    gae_lambda = trial.suggest_categorical("gae_lambda", [0.8, 0.9, 0.92, 0.95, 0.98, 0.99, 1.0])
+    n_steps = trial.suggest_categorical("n_steps", [8, 16, 32, 64, 128, 256, 512, 1024])
```
HideakiImamura, line 46: `suggest_categorical` does not consider the value of each choice, so the performance might degrade for real-valued parameters. On the other hand, these `suggest_categorical` calls can be converted to `suggest_float` or `suggest_int` as follows.

```suggestion
    gamma = 1. - trial.suggest_float("gamma", 0.0001, 0.1, log=True)
    max_grad_norm = trial.suggest_float("max_grad_norm", 0.3, 5.0, log=True)
    gae_lambda = 1. - trial.suggest_float("gae_lambda", 0.001, 0.2. log=True)
    n_steps = 2 ** trial.suggest_int("exponent_n_steps", 3, 10)
```

araffin: thanks, I did not know about those...

araffin: btw, why do you set `log=True` ?

araffin:
Is there an easy way to plot the user attributes?

Because for a RL researcher, it's not very intuitive to see "gamma=0.002" in the terminal...

examples/rl/sb3_simple.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def sample_a2c_params(trial: optuna.Trial) -> Dict[str, Any]:
+    """Sampler for A2C hyperparameters."""
+    gamma = 1.0 - trial.suggest_float("gamma", 0.0001, 0.1, log=True)
+    max_grad_norm = trial.suggest_float("max_grad_norm", 0.3, 5.0, log=True)
+    gae_lambda = 1.0 - trial.suggest_float("gae_lambda", 0.001, 0.2, log=True)
+    n_steps = 2 ** trial.suggest_int("exponent_n_steps", 3, 10)
+    learning_rate = trial.suggest_loguniform("lr", 1e-5, 1)
+    ent_coef = trial.suggest_loguniform("ent_coef", 0.00000001, 0.1)
```
HideakiImamura, line 48: These lines can be converted to use `suggest_float` as follows.
```suggestion
    learning_rate = trial.suggest_float("lr", 1e-5, 1, log=True)
    ent_coef = trial.suggest_float("ent_coef", 0.00000001, 0.1, log=True)
```

HideakiImamura:
IMO, the current visualization of the optimized parameters looks good. If we have a sufficient time budget, it would be a good idea to run the training again with the optimized parameters and evaluate the performance of the parameters.

examples/rl/sb3_simple.py
araffin: is that the recommended way now?
I personally find it less clear that you have to notice `log=True` to know what is happening.

HideakiImamura: Yes, it is. Currently, we recommends to use `suggest_float` rather than `suggest_uniform`, `suggest_loguniform, and `suggest_discrete`.

HideakiImamura:
Could you merge the master branch? It will resolve the CI failure.

PR #1428, opened 2020-06-25
Replace `DeprecationWarning` with `FutureWarning` in `@deprecated`.

toshihikoyanase:
## Motivation

By default, the message of `DeprecationWarning` is suppressed and it may be difficult for users to find the deprecated features. Since Python 3.7, the definition of `FutureWarning` was changed, and I think we can utilize it to notify the deprecated feature to users.

Since Python 3.7
- `DeprecationWarning`: warnings about deprecated features when those warnings are intended for other Python developers
- `FeatureWarning`: warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.

Notes
- `FutureWarning` is also used by `scikit-learn` for this purpose ([See](https://github.com/scikit-learn/scikit-learn/blob/95d4f0841/sklearn/utils/deprecation.py#L102)). 
- Do Python 3.5, 3.6 users have any troubles with this change?

## Description of the changes
This PR uses `FutureWarning` instead of `DeprecationWarning` in `@deprecated`.

### DIFF
--- a/optuna/_deprecated.py
+++ b/optuna/_deprecated.py
@@ -107,7 +107,7 @@ def deprecated(
                         deprecated_version,
                         removed_version,
                     ),
-                    DeprecationWarning,
+                    FutureWarning,
                     stacklevel=2,
                 )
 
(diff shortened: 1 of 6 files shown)


hvy:
Seems like existing tests are breaking.

PR #1434, opened 2020-06-25
Align `sphinx-rtd-theme` and Python versions used on Read the Docs to CircleCI.

harupy:
## Motivation

Currently, the version of `sphinx_rtd_theme` is not specified. This causes Read-the-docs (RTD) to use the pre-installed `sphinx_rtd_theme<0.5`.

https://readthedocs.org/projects/optuna/builds/11327552/

Screen Shot 2020-06-25 at 21 36 45

This PR adds a version specifier for `sphinx-rtd-theme` that enforces the update on RTD and aligns the version used on RTD and CircleCI.

## Difference

The screenshots below compare how the doc looks like on RTD and CricleCI. On RTD, the type hints are not rendered.

###  Read-the-Docs

Screen Shot 2020-06-25 at 21 28 46

https://optuna.readthedocs.io/en/stable/reference/study.html

### CircleCI

Screen Shot 2020-06-25 at 21 28 51

https://46950-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/study.html

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
       - run:
           name: install
           command: |
-            python -m venv venv || virtualenv venv --python=python3
+            python -m venv venv || virtualenv venv --python=python3.8
             . venv/bin/activate
             pip install -U pip
             pip install --progress-bar off .[document]
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -13,7 +13,7 @@ formats: all
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: 3.7
+  version: 3.8
   install:
     - method: pip
       path: .
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -13,6 +13,8 @@ formats: all
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
+  # Sphinx requires either Python >= 3.8 or `typed-ast` to reflect type comments 
+  # in the documentation. See: https://github.com/sphinx-doc/sphinx/pull/6984
   version: 3.8
   install:
     - method: pip
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -13,7 +13,7 @@ formats: all
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  # Sphinx requires either Python >= 3.8 or `typed-ast` to reflect type comments 
+  # `sphinx` requires either Python >= 3.8 or `typed-ast` to reflect type comments 
   # in the documentation. See: https://github.com/sphinx-doc/sphinx/pull/6984
   version: 3.8
   install:
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -97,6 +97,8 @@ html_favicon = '../image/favicon.ico'
 
 html_logo = '../image/optuna-logo.png'
 
+html4_writer = True
+
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -97,8 +97,6 @@ html_favicon = '../image/favicon.ico'
 
 html_logo = '../image/optuna-logo.png'
 
-html4_writer = True
-
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ def get_extras_require() -> Dict[str, List[str]]:
             "mlflow",
         ],
         # TODO(hvy): Unpin `sphinx` version after https://github.com/sphinx-doc/sphinx/issues/7807.
-        "document": ["sphinx>=3.0.0,!=3.1.0,!=3.1.1", "sphinx_rtd_theme"],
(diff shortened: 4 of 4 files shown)


harupy:
It turns out the root cause is not `sphinx-rtd-theme`, but Python version. `sphinx` uses`ast` ([ref](https://docs.python.org/3/library/ast.html)) to extract type comments. In Python 3.8 (CircleCI), `ast.parse` has an argument named `type_comments` ([ref](https://docs.python.org/3/library/ast.html#ast.parse)) which allows parsing type comments. In Python < 3.8, this argument doesn't exist and Sphinx uses `typed-ast` (not a build-in library) **if it's avaialble** (otherwise use `ast` and ignore type comments). Currently, Python 3.7 is used in RTD and `typed-ast` is not installed so type comments are ignored. There are a couple ways to address this issue:

1. Use Python 3.8 on RTD
2. Install `typed-ast` on RTD

harupy:
Verified using Python 3.8 fixes the issue on my forked repo.

https://harupy-optuna.readthedocs.io/en/latest/reference/study.html

.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ jobs:
       - run:
           name: install
           command: |
-            python -m venv venv || virtualenv venv --python=python3
+            python -m venv venv || virtualenv venv --python=python3.8
```
harupy, line 72: Pinned python version to make it obvious that 3.8 is used.

setup.py
```diff
@@ -61,7 +61,7 @@ def get_extras_require() -> Dict[str, List[str]]:
             "mlflow",
         ],
         # TODO(hvy): Unpin `sphinx` version after https://github.com/sphinx-doc/sphinx/issues/7807.
-        "document": ["sphinx>=3.0.0,!=3.1.0,!=3.1.1", "sphinx_rtd_theme"],
+        "document": ["sphinx>=3.0.0,!=3.1.0,!=3.1.1", "sphinx_rtd_theme>=0.5.0"],
```
toshihikoyanase, line 64: I have a question about this version restriction. Is it necessary for us to update the `sphinx_rtd_theme` version to show the type hints correctly? Or, is it optional?

I think it's acceptable to use the latest sphinx, but I'm a bit concerned about removing the version restriction set by ReadTheDocs. I briefly checked the issues in the `sphinx_rtd_theme` repository, and I found some report which claims `v0.5.0` unintended changes some layouts (e.g., https://github.com/readthedocs/sphinx_rtd_theme/issues/949). I don't think problems occur in Optuna reference as far as I checked https://harupy-optuna.readthedocs.io/en/latest/reference/study.html.

harupy: > Is it necessary for us to update the sphinx_rtd_theme version to show the type hints correctly?

I thought it **was**, but it actually **isn't**. So let's stick to `sphinx_rtd_theme<0.5`.

toshihikoyanase: Thank you for your reply. I support your decision to focus only on the Python version.
Could you update the PR title too?

.readthedocs.yml
```diff
@@ -13,7 +13,7 @@ formats: all
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: 3.7
+  version: 3.8
```
harupy, line 16: Should I insert a comment on why 3.8 is used?

setup.py
```diff
@@ -61,7 +61,7 @@ def get_extras_require() -> Dict[str, List[str]]:
             "mlflow",
         ],
         # TODO(hvy): Unpin `sphinx` version after https://github.com/sphinx-doc/sphinx/issues/7807.
-        "document": ["sphinx>=3.0.0,!=3.1.0,!=3.1.1", "sphinx_rtd_theme"],
+        "document": ["sphinx>=3.0.0,!=3.1.0,!=3.1.1", "sphinx_rtd_theme<0.5.0"],
```
harupy, line 64: `<0.5.0` is required to prevent installing the latest version `sphinx_rtd_theme` on CircleCI.

.readthedocs.yml
toshihikoyanase: Yes, please. In the future, we may want to update Python version to 3.9 or so, and such a comment will help us.

setup.py
toshihikoyanase: How about adding a comment here to show that the version of `sphinx_rtd_theme` is restricted due to the ReadTheDocs?

setup.py
harupy: Sure. I'll update the title too.

docs/source/conf.py
```diff
@@ -97,6 +97,8 @@
 
 html_logo = '../image/optuna-logo.png'
 
+html4_writer = True
```
harupy, line 100: ### Without `html4_writer = True`:

Screen Shot 2020-06-26 at 22 56 26

### With `html4_writer = True`:

Screen Shot 2020-06-26 at 22 55 39

One of the `sphinx_rtd_theme` core members says the second image is how it should look like.

https://github.com/readthedocs/sphinx_rtd_theme/issues/949#issuecomment-648942868

toshihikoyanase: Thank you for your investigation. As far as I understand, we do not need to introduce `html4_writer` as the following reasons:
- Unlike https://github.com/readthedocs/sphinx_rtd_theme/issues/949,  the descriptions of parameters do not exceed the width of the container.
- `html4_writer` will be deprecated as discussed in https://github.com/readthedocs/sphinx_rtd_theme/issues/938.

What do you think?

harupy: That makes sense. I have reverted the change.

harupy:
#1434 should be merged first because it will update the version of `sphinx-rtd-theme` used in CircleCI. We should review the changes after that.

Issue #1429, opened 2020-06-25
Use a different color for argument names to differentiate them from type hints


harupy:
## Motivation

Currently, argument names and type hints in the documentation have the same font color, which makes it difficult to differentiate them, especially when a function or class has lots of arguments.

Current:

![Screen Shot 2020-06-25 at 14 02 44](https://user-images.githubusercontent.com/17039389/85657434-ff4d1400-b6ec-11ea-82cf-160283d8d300.png)

Using a different font color for argument names can solve this issue.

Proposed:

![Screen Shot 2020-06-25 at 14 03 05](https://user-images.githubusercontent.com/17039389/85657334-ddec2800-b6ec-11ea-84ea-a0b8b211d28a.png)

I think changing a font-style or font-weight should work too.

## Description

The `span` tags that hold argument names can be selected by `.sig-param > span:nth-child(1)` so adding the css below does the trick. 

```css
.sig-param > span:nth-child(1) {
  color: <color>;
}
```

harupy:
I'll open a PR.

hvy:
That'd be great!

hvy:
@harupy should we close this issue now that https://github.com/optuna/optuna/pull/1436 is merged?

harupy:
Looks like I used a wrong closing-issue keyword in #1436.

PR #1436, opened 2020-06-26
Use a different font color for argument names

harupy:
## Motivation

Addresses #1429 

## Description of the changes

Added CSS to change the font color of argument names.


harupy:
`lightsgreen`

Screen Shot 2020-06-26 at 12 10 00

Screen Shot 2020-06-26 at 12 07 41

harupy:
`lightcoral`

Screen Shot 2020-06-26 at 12 17 11

harupy:
`darkgoldenrod`

Screen Shot 2020-06-26 at 12 25 51

toshihikoyanase:
Thank you for your PR! I believe that this PR will improve the readability of the reference.

Let me share my idea about the colors introduced in #1370. I'm not a web designer and I reused the color pallet of https://optuna.org, instead of using new colors. So, I'd like to use an existing color or a variation of existing colors.


Personally, I like `lightseagreen` (`#20B2AA`) and red color (`#E74C3C`). (The red color is used to highlight the literal/variable/constants like `sqlite:///example.db`.) I just created a color pallet to choose the color for the argument. A, B, C, and D are originally used in the reference, and E will be newly introduced as `lightseagreen`.

- A: `#03588C` from Optuna logo
- B: `#2980B9` is the color of type names in the class definition
- C: `#E74C3C` is the color of the literals/variables/constants (e.g., ) marked with double backticks
- D: `#27AE60` is the color of `[source]` link
- E: `#20B2AA` is lightseagreen.

image

I'm not familiar with design, but the following is my personal opinion about the color.

- The lightseagreen (`#20B2AA` ) is located in the middle of A and D, and the colors in the argument seem to be balanced. So I feel it natural. I think the disadvantage of lightseagreen(`#20B2AA` ) may be too close to other colors like B(`#2980B9`) and the link color(`#2BA9CD`)
- The red color (`#E74C3C`) is located on the opposite side of A and D, and it will be clearly distinguished from the other colors in the argument. In addition, we do not need to add a new color. On the other hand, it may highlight the argument too much.

The following is the screenshot with the red color (`#E74C3C`).
スクリーンショット 2020-06-26 15 53 43
スクリーンショット 2020-06-26 15 54 06

I would appreciate your feedback.

harupy:
@toshihikoyanase Thanks for sharing your awesome ideas! It perfectly makes sense to reuse an existing color instead of adding a new one. I'm leaning toward the red color. It seems against the current lightblue-based theme, but clearly highlights argument names on both class and function signatures.

toshihikoyanase:
@harupy I'm sorry for your consideration, but I talked with @hvy, and I changed my mind.
He suggested the following colors scheme:

スクリーンショット 2020-06-26 16 51 00

スクリーンショット 2020-06-26 16 52 31

This scheme moderately highlights the argument name, and it is simpler and cleaner than my previous proposal.
- This color scheme does not include new colors.
- Use gray for argument names
- Use light gray for the background of class definitions, which is the same color as the background of function definitions.
- Use light blue for the types, which is the same color for the border-top of class definition

What do you think?

harupy:
The new scheme looks much better.

hvy:
Sorry I didn't mean to interfere your discussion and I really like the initiate in following up the improvements of the style of the docs. What I think is important is that
- the parameter names are more emphasized than the types (red achieves this quite well, but it also gives an impression of "caution".)
- we don't introduce unnecessary colors

toshihikoyanase:
@harupy @hvy

Thank you for your feedback. I made a prototype of https://github.com/optuna/optuna/pull/1436#issuecomment-650041575. I found inconsistency of the color scheme, so I fixed the colors in https://github.com/optuna/optuna/pull/1436#issuecomment-650041575.

image


I found out that the style defined in `source/_templates/footer.html` overwrites `source/_templates/custom.css`. It makes this PR complicated, and I think we integrate the style definition into `source/_templates/custom.css` in advance.
I'm sorry for bothering you, but could you check the PR if I create a PR?

harupy:
Sure, let's finish that PR first.

toshihikoyanase:
@harupy Thank you! I created a PR #1439. Could you check it?

harupy:
Now it looks like this:

Screen Shot 2020-06-26 at 22 03 40

https://47207-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/study.html

harupy:
For built-in types (such as `str`), I use the same color as the `[source]`  link. What do you think?

toshihikoyanase:
@harupy Thank you for your update. That's great!

> For built-in types (such as `str`), I use the same color as the `[source]` link to indicate they are links. 

I think users are familiar with the built-in types, and they rarely follow such links. So, IMO, we can use the theme color, but it is not a strong opinion.

harupy:
It looks like `theme.css` overwrites some styles in `custom.css`.

harupy:
#1434 should be merged first because it will update the version of `sphinx-rtd-theme` used in CircleCI. We should review the changes after that.

hvy:
#1434 is now merged.

hvy:
Took a quick look but looks good. It'll also make noisy interfaces more readable.

[This PR's LightGBMTunerCV](https://47921-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/generated/optuna.integration.lightgbm.LightGBMTunerCV.html#optuna.integration.lightgbm.LightGBMTunerCV)
[master's LightGBMTunerCV (as of 20200701)](https://optuna.readthedocs.io/en/latest/reference/generated/optuna.integration.lightgbm.LightGBMTunerCV.html)

harupy:
Great! Do you have any preference for the position of the colored border?

The current style rule is:

- For classes or functions, color the border-top
- For methods, properties, parameters, and returns, color the border-left

Screen Shot 2020-07-01 at 15 07 59

hvy:
I think the current positioning is good. Sort of acting as a separator for classes and free functions, although they are all displayed on their own page now. Maybe you had something else in mind?

harupy:
I actually don't have any preference for this. Whatever works for you is fine with me :)

harupy:
One last thing. I have applied auto-formatting with prettier to make it look pretty and consistent.

toshihikoyanase:
@harupy Thank you for applying the code formatter, but it produces the many diff not related to the font color for argument. I think the change includes both logic change and refactoring is likely to cause issues, and it makes review difficult.
I'm really sorry, but could you separate the PRs to apply the formater?

harupy:
@toshihikoyanase Sorry, I have reverted the auto-formatting.

harupy:
@hvy @toshihikoyanase Thank you for the review!

PR #1437, opened 2020-06-26
Remove `optuna.structs` from MLflow integration.

crcrpar:
Switch from `optuna.structs` to appropriate modules.

### DIFF
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -4,8 +4,8 @@ import optuna
 from optuna._experimental import experimental
 from optuna._imports import try_import
 from optuna._study_direction import StudyDirection
-from optuna import type_checking
 from optuna.trial import TrialState
+from optuna import type_checking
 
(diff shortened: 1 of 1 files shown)


optuna/integration/mlflow.py
```diff
@@ -1,12 +1,14 @@
+from typing import Optional
+
 import optuna
 from optuna._experimental import experimental
 from optuna._imports import try_import
-from optuna import structs
+from optuna._study_direction import StudyDirection
```
c-bata, line 6: How about importing from `optuna.study.StudyDirection`?

crcrpar: Hmm, maybe just I'm stupid, but it doesn't sound good by staring at this.

https://github.com/optuna/optuna/blob/master/optuna/study.py#L15

crcrpar: Ah, the message recommends that. Gotcha.

PR #1438, opened 2020-06-26
Continue to allow using `IntLogUnioformDistribution.step` during deprecation.

hvy:
## Motivation

A minor follow-up to https://github.com/optuna/optuna/pull/1387 and specifically https://github.com/optuna/optuna/pull/1387#issuecomment-649981524.

## Description of the changes

Allow users to continue using the `step` attribute of `IntLogUniformDistribution` during deprecation.

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -326,8 +326,7 @@ class IntLogUniformDistribution(BaseDistribution):
                 but this schedule is subject to change.
 
                 Samplers and other components in Optuna relying on this distribution will ignore
-                this value and assume that
-                :attr:`~optuna.distributions.IntLogUniformDistribution.step` is always 1.
+                this value and assume that ``step`` is always 1.
                 User-defined samplers may continue to use other values besides 1 during the
                 deprecation.
     """
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -2,13 +2,13 @@ import abc
 import copy
 import decimal
 import json
+from typing import Dict
 import warnings
 
 from optuna import type_checking
 
 if type_checking.TYPE_CHECKING:
     from typing import Any  # NOQA
-    from typing import Dict  # NOQA
     from typing import Sequence  # NOQA
     from typing import Union  # NOQA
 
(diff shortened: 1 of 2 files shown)


optuna/distributions.py
```diff
@@ -319,13 +320,16 @@ class IntLogUniformDistribution(BaseDistribution):
...
-                This value is valid for only 1. Otherwise, the value is replaced with 1.
-
             .. warning::
                 Deprecated in v2.0.0. ``step`` argument will be removed in the future.
                 The removal of this feature is currently scheduled for v4.0.0,
                 but this schedule is subject to change.
+
+                Samplers and other components in Optuna relying on this distribution will ignore
+                this value and assume that
+                :attr:`~optuna.distributions.IntLogUniformDistribution.step` is always 1.
```
toshihikoyanase, line 330: [nit] `step` appears twice in this warning message, but the styles are not consistent. I think both of them are acceptable, but could you choose one of them?

image

optuna/distributions.py
```diff
@@ -342,16 +346,35 @@ def __init__(self, low: int, high: int, step: int = 1) -> None:
...
-                FutureWarning,
-            )
+            self._warn_step()
 
         self.low = low
         self.high = high
+        self._step = step
+
+    def __repr__(self):
+        # type: () -> str
```
toshihikoyanase, line 356: Could you tell me why you use type comments for `__repr__` and `_asdict`?

optuna/distributions.py
hvy: Aligned to the former. I guess it's obvious without hyperreferencing.

optuna/distributions.py
hvy: Aligned with `BaseDistribution`. `IntLogUniformDistribution` as well as this entire module is mixed with modern/comment style type hints, but I aligned the type hints for this class now to modern style. What do you think?

hvy:
Thanks a lot for the quick review. Comments should be address, PTAL.

optuna/distributions.py
toshihikoyanase: Thank you for your update. I think we migrate the type hinting gradually, and we can use the new style if we do not have specific reasons.

PR #1449, opened 2020-06-29
Load environment variables in AllenNLPExecutor

### LINKED ISSUE #1446: AllenNLP integration: Can't load configurations that expect ext_vars from environment variables
Optuna's AllenNLP integration evaluates the jsonnet config that you give it, passing in the trial parameters as the `ext_vars`. AllenNLP's default behavior when loading a configuration is a little different, in that it also passes through environment variables: 

https://github.com/allenai/allennlp/blob/4de68a42a0df89df01a999ee48f361f24c8c19d4/allennlp/common/params.py#L488

So if you rely on specifying e.g. a path to a file as an environment variable, that configuration won't load as expected when using the integration

## Expected behavior

Optuna should pass environment variables through when the jsonnet is interpreted. Ideally, it'd just use the AllenNLP config loading code directly so that all behavior is identical, just overriding whatever the trial needs to override.

## Environment

- Optuna version: 1.5.0
- Python version: 3.6
- OS: Ubuntu 18.04.4
- (Optional) Other libraries and their versions: AllenNLP 1.0

You'll get `RuntimeError: undefined external variable` during the interpretation of the jsonnet here:

https://github.com/optuna/optuna/blob/9f5c5cf5c36df1628567f0b4a9873892d148ad40/optuna/integration/allennlp.py#L108

## Steps to reproduce

1. Use a AllenNLP configuration that expects to be able to use an environment variable to fill an extVar

himkt:
## Motivation
This PR resolves #1446.

## Description of the changes
The current implementation of `AllenNLPExecutor` doesn't pass environment variables to `AllenNLP`.
It means users couldn't use environment variables for specifying some configs such as file path.
This PR changes `AllenNLPExecutor._build_params` to load environment variables and pass them to `AllenNLP`.

(I confirmed it works in https://gist.github.com/himkt/ecf2c1a0fb0bc3bec4eb59eed03a5561)

### DIFF
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -103,14 +103,22 @@ class AllenNLPExecutor(object):
         """Create a dict of params for AllenNLP."""
         # _build_params is based on allentune's train_func.
         # https://github.com/allenai/allentune/blob/master/allentune/modules/allennlp_runner.py#L34-L65
-        for key, value in self._params.items():
-            self._params[key] = str(value)
-        _params = json.loads(_jsonnet.evaluate_file(self._config_file, ext_vars=self._params))
+        params = self._environment_variables()
+        params.update({key: str(value) for key, value in self._params.items()})
 
-        # _params contains a list of string or string as value values.
+        allennlp_params = json.loads(_jsonnet.evaluate_file(self._config_file, ext_vars=params))
+        # allennlp_params contains a list of string or string as value values.
         # Some params couldn't be casted correctly and
         # infer_and_cast converts them into desired values.
-        return allennlp.common.params.infer_and_cast(_params)
+        return allennlp.common.params.infer_and_cast(allennlp_params)
+
+    @staticmethod
+    def _is_encodable(value: str) -> bool:
+        # https://github.com/allenai/allennlp/blob/master/allennlp/common/params.py#L77-L85
+        return (value == "") or (value.encode("utf-8", "ignore") != b"")
+
+    def _environment_variables(self) -> Dict[str, str]:
+        return {key: value for key, value in os.environ.items() if self._is_encodable(value)}
 
     def run(self) -> float:
         """Train a model using AllenNLP."""
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -59,8 +59,8 @@ class AllenNLPExecutor(object):
 
     .. note::
         In :class:`~optuna.integration.AllenNLPExecutor`,
-        you can pass parameter to AllenNLP by either defining a search space using
-        Optuna suggest methods or setting a environment variable just like AllenNLP cli.
+        you can pass parameters to AllenNLP by either defining a search space using
+        Optuna suggest methods or setting environment variables just like AllenNLP CLI.
         If a value is set in both a search space in Optuna and the environment variables,
         the executor will use the value specified in the search space in Optuna.
 
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -58,7 +58,9 @@ class AllenNLPExecutor(object):
     examples/allennlp/classifier.jsonnet>`_.
 
     .. note::
-        :class:`~optuna.integration.AllenNLPExecutor` also reads environment variables.
+        In :class:`~optuna.integration.AllenNLPExecutor`,
+        you can pass parameter to AllenNLP by either defining a search space using
+        Optuna suggest methods or setting a environment variable just like AllenNLP cli.
         If a value is set in both a search space in Optuna and the environment variables,
         the executor will use the value specified in the search space in Optuna.
 
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -57,6 +57,11 @@ class AllenNLPExecutor(object):
     `config file <https://github.com/optuna/optuna/blob/master/
     examples/allennlp/classifier.jsonnet>`_.
 
+    .. note::
+        :class:`~optuna.integration.AllenNLPExecutor` also reads environment variables.
+        If a value is set in both a search space in Optuna and the environment variables,
+        the executor will use the value specified in the search space in Optuna.
+
     Args:
         trial:
             A :class:`~optuna.trial.Trial` corresponding to the current evaluation
@@ -105,10 +110,6 @@ class AllenNLPExecutor(object):
         _build_params is based on allentune's train_func.
         For more detail, please refer to
         https://github.com/allenai/allentune/blob/master/allentune/modules/allennlp_runner.py#L34-L65
-
-        :class:`~optuna.integration.AllenNLPExecutor` also reads environment variables.
-        If a value is set in both a search space in Optuna and the environment variables,
-        the executor will use the value specified in the search space in Optuna.
 
         """
         params = self._environment_variables()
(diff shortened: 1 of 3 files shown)


optuna/integration/allennlp.py
```diff
@@ -103,14 +103,22 @@ def _build_params(self) -> Dict[str, Any]:
         """Create a dict of params for AllenNLP."""
         # _build_params is based on allentune's train_func.
         # https://github.com/allenai/allentune/blob/master/allentune/modules/allennlp_runner.py#L34-L65
-        for key, value in self._params.items():
-            self._params[key] = str(value)
-        _params = json.loads(_jsonnet.evaluate_file(self._config_file, ext_vars=self._params))
+        params = self._environment_variables()
```
toshihikoyanase, line 106: Could you add a test case about the environmental variables? This change affects the behavior of a public method, and I think it would prevent degradation during future development.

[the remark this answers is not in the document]
himkt: Added a small description in 964d605.
I'm wondering if we should be placed on the top of `AllenNLPExecutor`, as `_build_params` may not appear in the documentation. I'd like to hear some opinions.

himkt: Added in c5e96a9.

optuna/integration/allennlp.py
```diff
@@ -57,6 +57,11 @@ class AllenNLPExecutor(object):
     `config file <https://github.com/optuna/optuna/blob/master/
     examples/allennlp/classifier.jsonnet>`_.
 
+    .. note::
+        :class:`~optuna.integration.AllenNLPExecutor` also reads environment variables.
+        If a value is set in both a search space in Optuna and the environment variables,
+        the executor will use the value specified in the search space in Optuna.
```
toshihikoyanase, line 63: Thank you for moving the comment to the `AllenNLP.__init__`. The rendered result is as follows:

image

How about clarifying another object (i.e., a search space of Optuna) of the `read` instead of omitting by `also`? If we add it, the note will be self-contained and easier to read.

`AllenNLPExecutor` reads environment variables as well as a search space of Optuna. If a value is set in both of them, the executor ...

tests/integration_tests/allennlp_tests/test_allennlp.py
```diff
@@ -8,7 +8,7 @@
 import optuna
 
 
-def test__set_param() -> None:
+def test__build_params() -> None:
```
HideakiImamura, line 11: A nit.

```suggestion
def test_build_params() -> None:
```

tests/integration_tests/allennlp_tests/test_allennlp.py
```diff
@@ -23,6 +23,52 @@ def test__set_param() -> None:
     assert params["model"]["hidden_size"] == [100, 200, 300]
 
 
+def test__build_params_overwiting_environment_variable() -> None:
```
HideakiImamura, line 26: Ditto and typo.
```suggestion
def test_build_params_overwriting_environment_variable() -> None:
```

tests/integration_tests/allennlp_tests/test_allennlp.py
```diff
@@ -23,6 +23,52 @@ def test__set_param() -> None:
...
+    params = executor._build_params()
+    os.environ.pop("TRAIN_PATH")
+    os.environ.pop("VALID_PATH")
+    assert params["train_data_path"] == "tests/integration_tests/allennlp_tests/sentences.train"
+    assert (
+        params["validation_data_path"] == "tests/integration_tests/allennlp_tests/sentences.valid"
+    )
+
+
+def test__build_params_when_optuna_and_environment_variable_both_exist() -> None:
```
HideakiImamura, line 47: Ditto.
```suggestion
def test_build_params_when_optuna_and_environment_variable_both_exist() -> None:
```

optuna/integration/allennlp.py
himkt: 0c5a774 I tried updating my docstring so that it mentions AllenNLP.
How do you feel about the change?
If it doesn't make sense for you, I'll apply for your suggestion.
Thank you for the suggestion for improving English, it's really helpful for me as always.

tests/integration_tests/allennlp_tests/test_allennlp.py
himkt: Thank you for the comment.
I named the test `test__build_params` because these are the tests for private methods (`test_` + `_build_params`).
https://github.com/optuna/optuna/pull/1449/files#diff-194650369fb014d2ea8f428774c04276L102

Do you think it's okay to remove consecutive underscores?

tests/integration_tests/allennlp_tests/test_allennlp.py
himkt: Thank you for the catch, I fixed the typo. :pray:
I removed a consecutive underscore after confirming the question at https://github.com/optuna/optuna/pull/1449#discussion_r447619368.

tests/integration_tests/allennlp_tests/test_allennlp.py
toshihikoyanase: Hmm, we use both `test_` and `test` for the test functions of the private methods.

We have three test functions which use `test_` as prefix:
- `tests/integration_tests/allennlp_tests/test_allennlp.py:11:def test__set_param() -> None:`
- `tests/integration_tests/lightgbm_tuner_tests/test_alias.py:5:def test__handling_alias_parameters():`
- `tests/integration_tests/lightgbm_tuner_tests/test_optimize.py:337:    def test__parse_args_wrapper_args(self):`

Most of the test functions employ `test`, although `test_` seems to follow the naming convention.
@keisuke-umezawa @hvy Do you have any ideas about it?

optuna/integration/allennlp.py
```diff
@@ -57,6 +57,13 @@ class AllenNLPExecutor(object):
     `config file <https://github.com/optuna/optuna/blob/master/
     examples/allennlp/classifier.jsonnet>`_.
 
+    .. note::
+        In :class:`~optuna.integration.AllenNLPExecutor`,
+        you can pass parameter to AllenNLP by either defining a search space using
+        Optuna suggest methods or setting a environment variable just like AllenNLP cli.
```
toshihikoyanase, line 63: A nit. 
- I propose a plural form to show that we can use multiple parameters clearly.
- I usually capitalize `cli` because it's an abbreviation. If you have specific reasons, please ignore it.

```suggestion
        you can pass parameters to AllenNLP by either defining a search space using
        Optuna suggest methods or setting environment variables just like AllenNLP CLI.
```

optuna/integration/allennlp.py
toshihikoyanase: Thank you for your update. SGTM with a nit. Please take a look at https://github.com/optuna/optuna/pull/1449#pullrequestreview-440476843.

tests/integration_tests/allennlp_tests/test_allennlp.py
hvy: Hm, I think both makes sense but I'd also suggest removing the second underscore(s). A test function/method doesn't necessarily have to exactly match one particular function/method/etc. It could test multiple, for instance, and this rule of `test_my_function_name` wouldn't apply anyway. I haven't really caught up on the context here so it's not a strong opinion, but I generally find it more appealing when the name of the test makes sense semantically.

Ideally, we shouldn't have to write tests for private interfaces. 🙁

himkt: It's reasonable for me and I removed underscores.
@HideakiImamura @hvy @toshihikoyanase Thank you for the bunch of discussions.

PR #1454, opened 2020-06-29
Translate type comments into type annotations using `com2ann`.

### LINKED ISSUE #711: Use function annotation syntax for Type Hints.

### LINKED ISSUE #1445: Use `com2ann` to translate type comments into type annotations
Related to https://github.com/optuna/optuna/issues/711.

I found this python package named `com2ann` which translates type comments into type annotations:

https://github.com/ilevkivskyi/com2ann

I'm testing this package on my local and it seems to be working fine so far. One adjustment that I had to make was prevent `com2ann` from translating type comments on variables into  [variable annotaions](https://www.python.org/dev/peps/pep-0526/) (introduced in Python 3.6) because Optuna still supports Pyhton 3.5. Here's the fixed version which preserves type comments on variables.

https://github.com/harupy/com2ann/tree/disable-variable-annotation

Can I give this a try?

harupy:
## Motivation

Resolves #711
Resolves #1445 

## Description of the changes

- Applied `com2ann` to all the files
- ~Added `from __future__ import annotations` to some files to avoid lint errors~

### DIFF
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -401,4 +401,3 @@ def main() -> int:
 
     argv = sys.argv[1:] if len(sys.argv) > 1 else ["help"]
     return _OptunaApp().run(argv)
-
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -33,8 +33,7 @@ if type_checking.TYPE_CHECKING:
     from typing import Tuple  # NOQA
 
 
-def _check_storage_url(storage_url):
-    # type: (Optional[str]) -> str
+def _check_storage_url(storage_url: Optional[str]) -> str:
 
     if storage_url is None:
         raise CLIUsageError("Storage URL is not specified.")
@@ -42,8 +41,7 @@ def _check_storage_url(storage_url):
 
 
 class _BaseCommand(Command):
-    def __init__(self, *args, **kwargs):
-        # type: (List[Any], Dict[str, Any]) -> None
+    def __init__(self, *args: List[Any], **kwargs: Dict[str, Any]) -> None:
 
         super(_BaseCommand, self).__init__(*args, **kwargs)
         self.logger = optuna.logging.get_logger(__name__)
@@ -52,8 +50,7 @@ class _BaseCommand(Command):
 class _CreateStudy(_BaseCommand):
     """Create a new study."""
 
-    def get_parser(self, prog_name):
-        # type: (str) -> ArgumentParser
+    def get_parser(self, prog_name: str) -> ArgumentParser:
 
         parser = super(_CreateStudy, self).get_parser(prog_name)
         parser.add_argument(
@@ -78,8 +75,7 @@ class _CreateStudy(_BaseCommand):
         )
         return parser
 
-    def take_action(self, parsed_args):
-        # type: (Namespace) -> None
+    def take_action(self, parsed_args: Namespace) -> None:
 
         storage_url = _check_storage_url(self.app_args.storage)
         storage = optuna.storages.get_storage(storage_url)
@@ -95,8 +91,7 @@ class _CreateStudy(_BaseCommand):
 class _DeleteStudy(_BaseCommand):
     """Delete a specified study."""
 
-    def get_parser(self, prog_name):
-        # type: (str) -> ArgumentParser
+    def get_parser(self, prog_name: str) -> ArgumentParser:
 
         parser = super(_DeleteStudy, self).get_parser(prog_name)
         parser.add_argument(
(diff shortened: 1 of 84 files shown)


harupy:
The conditional imports from the `typing` module cause an error.

```python
if type_checking.TYPE_CHECKING:
    from typing import Any  # NOQA
    ...
```

bigbird555:
I don't know if this will help, but when I was doing it myself, I would either delete `if type_checking.TYPE_CHECKING:` or rewrite `if type_checking.TYPE_CHECKING:` to `with try_import() as _imports:`.

harupy:
@bigbird555 Thanks for the comment! It looks like there are a few imports that should be moved under `with try_import()` (e.g. `from _pytest.recwarn import WarningsRecorder` in `tests/test_study.py`)

harupy:
It seems like we still need to use `TYPE_CHECKING` to prevent circular import...

bigbird555:
Okay.... I'm sorry I couldn't point out that problem.

harupy:
Here are the issues that I found.

1. `TYPE_CHECKING` is still required in some modules to prevent circular import ([ref](https://mypy.readthedocs.io/en/stable/common_issues.html#import-cycles)).
1. The return type of a method that returns a class itself must be a string ([ref](https://stackoverflow.com/questions/33533148/how-do-i-specify-that-the-return-type-of-a-method-is-the-same-as-the-class-itsel)).

```python
class Position:
    ...
    def __add__(self, other: 'Position') -> Position:  # raises an error
    def __add__(self, other: 'Position') -> 'Position':  # works
       ...
```

3. Just removing `if type_checking.TYPE_CHECKING` causes the import-order error. A package like [`isort`](https://github.com/timothycrosley/isort) can fix this issue.

HideakiImamura:
Thanks for your contributions as always! I would appreciate your great efforts and careful considerations. I have several comments.

> 1. TYPE_CHECKING is still required in some modules to prevent circular import.

In my experience, we should make it "string" to prevent circular imports. For example, 
```python
# optuna/pruners/_base.py
...
@abc.abstractmethod
        self, study: "Study", trial: "FrozenTrial"
...
```

> 3. Just removing if type_checking.TYPE_CHECKING causes the import-order error. A package like isort can fix this issue.

 Thanks for your survey! It sounds great. Could you apply `isort` to order imports?

Additionally, I saw CI results and have noticed that some modules are not used (e.g. `optuna.type_checking` and `typing`). Don't we have to fix them by hand? :(

harupy:
@HideakiImamura Thanks for the comment!

> Thanks for your survey! It sounds great. Could you apply isort to order imports?

I have tried `isort`, but its sorting rules differ from `hacking` we currently use.

> Additionally, I saw CI results and have noticed that some modules are not used (e.g. optuna.type_checking and typing). Don't we have to fix them by hand?

We could do this programmatically using `flake8` output log that tells us which files and lines cause unused-import-error.

harupy:
@HideakiImamura I think we should split this PR into smaller ones (otherwise it could take forever).

HideakiImamura:
@harupy Thanks for your proposal! I agree with your idea, but how do we proceed? I think we have several items, but we need to do them simultaneously to avoid inconsistency.

harupy:
@HideakiImamura 

Does the one-by-one approach @bigbird555 has been doing still make sense?

Because:

- This doesn't seem to be an urgent task
- It's easier to review and debug
- We can minimize conflictions

HideakiImamura:
Hmm... IMO, the one-by-one approach is safe to avoid the inconsistency and any conflict with other PRs. But, using `com2ann` looks a good approach to fix type hinting. I think it is helpful for committers to introduce `com2ann` in PR #711.

harupy:
Yes, I'll write how to install `com2ann` and how to  use it on #711.

HideakiImamura:
Thanks for the comment. I totally agree with you! Could you write to them and close this PR?

harupy:
I briefly described how to install and use `com2ann` in https://github.com/optuna/optuna/issues/711#issuecomment-652414062

PR #1455, opened 2020-06-30
Make `to_internal_repr` and `to_external_repr` private

keisuke-umezawa:
## Motivation

Becaue developers do not intended that they are used by optuna users, so I make them private. Also, those documentation will be excluded automatically.


keisuke-umezawa:
We will rethink API of distributions, so I closed it.

PR #1456, opened 2020-06-30
Use `step` to calculate range of `IntUniformDistribution` in PyCmaSampler.

toshihikoyanase:
## Motivation
This PR derived from https://github.com/optuna/optuna/pull/1302#discussion_r447462314.

## Description of the changes
`PyCmaSampler` does not consider the `IntUniformDistribution.step` to calculate the range (i.e., `low`, `high`). I think it is inconsistent with the range calculation for `DiscreteUniformDistribution`. This PR fixes the range of `IntUniformDistribution`.

### DIFF
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -316,8 +316,8 @@ class _Optimizer(object):
                 lows.append(0 - 0.5 * dist.q)
                 highs.append(r + 0.5 * dist.q)
             elif isinstance(dist, IntUniformDistribution):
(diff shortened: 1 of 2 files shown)


toshihikoyanase:
Thank you for your review. I resolved the conflict.

> Should we label this as a bug?

Hmm, I think it is `bug` or `enhancement`.  IMO, `bug` may be more appropriate because this issue may cause performance degradation when the solution is `low` or `high`.

PR #1460, opened 2020-07-01
Suppress progress bar of `LightGBMTuner` if `verbosity` == 1.

toshihikoyanase:
## Motivation

`LightGBMTuner` has `verbosity` options, and it suppresses log messages of Optuna and LightGBM. However, the progress bar is still shown in the terminal.

## Description of the changes

This PR suppresses the progress bar if `verbosity == 1` .

## Comment

@smly 
During the modification, I found that LightGBM's seed is set to `111` in the [`run`](https://github.com/optuna/optuna/blob/master/optuna/integration/_lightgbm_tuner/optimize.py#L438) method. I think usual users do not expect such configuration change. Do we remove the line?

### DIFF
--- a/optuna/integration/_lightgbm_tuner/optimize.py
+++ b/optuna/integration/_lightgbm_tuner/optimize.py
@@ -511,7 +511,10 @@ class _LightGBMBaseTuner(_BaseTuner):
         sampler: optuna.samplers.BaseSampler,
         step_name: str,
     ) -> _OptunaObjective:
-        pbar = tqdm.tqdm(total=n_trials, ascii=True)
+        if self.auto_options["verbosity"] == 0:
+            pbar = None
+        else:
+            pbar = tqdm.tqdm(total=n_trials, ascii=True)
 
         # Set current best parameters.
         self.lgbm_params.update(self.best_params)
@@ -548,8 +551,9 @@ class _LightGBMBaseTuner(_BaseTuner):
                 callbacks=self._optuna_callbacks,
             )
 
-        pbar.close()
-        del pbar
+        if pbar:
+            pbar.close()
+            del pbar
 
         return objective
 
@@ -559,7 +563,7 @@ class _LightGBMBaseTuner(_BaseTuner):
         target_param_names: List[str],
         train_set: "lgb.Dataset",
         step_name: str,
-        pbar: tqdm.tqdm,
+        pbar: Optional[tqdm.tqdm],
     ) -> _OptunaObjective:
 
         raise NotImplementedError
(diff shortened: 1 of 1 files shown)


smly:
> I think usual users do not expect such configuration change. Do we remove the line?
Thank you for finding the issue. You're right. The code that fixes the seed is a remnant of the old benchmark code.

toshihikoyanase:
@smly Thank you for your confirmation and approval of #1461.

optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -512,7 +512,10 @@ def _tune_params(
         sampler: optuna.samplers.BaseSampler,
         step_name: str,
     ) -> _OptunaObjective:
-        pbar = tqdm.tqdm(total=n_trials, ascii=True)
+        if self.auto_options["verbosity"] == 0:
+            pbar = None
+        else:
+            pbar = tqdm.tqdm(total=n_trials, ascii=True)
```
HideakiImamura, line 518: This "verbosity" decision is far away from the existing verbosity decision in `_LightGBMBaseTuner.run()` method (Line 436-441). How about moving this logic into the `if` statement in `_LightGBMBaseTuner.run()` method, and add the argument of `pbar` for `_tune_params` methods? My proposed suggestion looks as follows.

```python
...
# In `_LightGBMBaseTuner.run()`
if self.auto_options["verbosity"] == 0:
    self.auto_options["pbar"] = None
    self.lgbm_params["verbose"] = -1
    self.lgbm_params["seed"] = 111
    self.lgbm_kwargs["verbose_eval"] = False
else:
     self.auto_options["pbar"] = tqdm.tqdm(total=n_trials, ascii=True)
...
# In `_LightGBMBaseTuner._tune_params()`
pbar = self.auto_options["pbar"]
...
```

toshihikoyanase: Thank you for your proposal. I think it improves the locality of `self.auto_options["verbosity"]`, but it loses the locality of `pbar`. Currently, the `auto_options` dictionary handles literals only, but it may increase the potential of issues if we mix the instance like `tqdm`. For example, I think we need to remove it before pickling the `LightGBMTuner`.
What do you think?

HideakiImamura: Thanks for the reply. I agree with your idea not to use `auto_options` for this purpose. How about introducing another attribute `_use_pbar` for the locality of both `self.auto_options["verbosity"]` and `pbar` as follows?

```python
# In `_LightGBMBaseTuner.__init__()`
self._use_pbar = True
...
# In `_LightGBMBaseTuner.run()`
if self.auto_options["verbosity"] == 0:
    self._use_pbar = False
    self.lgbm_params["verbose"] = -1
    self.lgbm_params["seed"] = 111
    self.lgbm_kwargs["verbose_eval"] = False
...
# In `_LightGBMBaseTuner._tune_params()`
pbar = tqdm.tqdm(total=n_trials, ascii=True) if self._use_pbar else None
```

toshihikoyanase: Thank you for your suggestion. That makes sense. I applied the change in https://github.com/optuna/optuna/pull/1460/commits/9895eafe88ca5629aca7e192c51cfb3f11867638.

PR #1462, opened 2020-07-01
Remove Swig installation from Dockerfile

crcrpar:
Now swig installation of Dockerfile is the same as that of CI.

I'm not pretty sure whether this completely fixes the issue that some Docker images on DockerHub fail to install Keras :sweat:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,16 +7,6 @@ ENV PIP_OPTIONS "--no-cache-dir --progress-bar off"
 RUN apt-get update \
     && apt-get -y install openmpi-bin libopenmpi-dev \
     && rm -rf /var/lib/apt/lists/* \
-    && wget -q http://prdownloads.sourceforge.net/swig/swig-3.0.12.tar.gz \
-    && tar -zxvf swig-3.0.12.tar.gz \
-    && cd swig-3.0.12 \
-    && ./configure \
-    && make --silent -j \
-    && make install --silent \
-    && cd .. \
-    && rm -rf swig-3.0.12 \
-    && rm swig-3.0.12.tar.gz \
-    && swig -version \
     && pip install --no-cache-dir -U pip \
     && pip install ${PIP_OPTIONS} -U setuptools \
     && pip install ${PIP_OPTIONS} Cython  # for automl/ConfigSpace
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,8 @@ ARG PYTHON_VERSION=3.7
 
 FROM python:${PYTHON_VERSION}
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -16,11 +26,11 @@ ARG BUILD_TYPE='dev'
 
 RUN if [ "${BUILD_TYPE}" = "dev" ]; then \
         if [ "${PYTHON_VERSION}" \< "3.6" ]; then \
-            pip install --no-cache-dir -e '.[doctest, document, example, testing]' -f https://download.pytorch.org/whl/torch_stable.html; \
+            pip install --no-cache-dir --progress-bar off -e '.[doctest, document, example, testing]' -f https://download.pytorch.org/whl/torch_stable.html; \
```
harupy, line 29: How about putting the common pip install options in a variable and reuse it to DRY the commands?

```
PIP_OPTIONS="--cache-dir --progress-bar off"

pip install $PIP_OPTIONS -e '.[checking, ...]'
pip install $PIP_OPTIONS -e .
pip install $PIP_OPTIONS -e jupyter notebook
```

crcrpar: Sounds good!

hvy:
Feeling a bit bad interrupting but swig can actually be removed altogether since it's no longer a requirement, now that  https://github.com/optuna/optuna/pull/1106 is merged. It seems like other changes are included in this PR though. Maybe we can just generalize the title a bit.

Dockerfile
```diff
@@ -2,12 +2,14 @@ ARG PYTHON_VERSION=3.7
...
+ENV PIP_OPTIONS "--no-cache-dir --progress-bar off"
+
 RUN apt-get update \
     && apt-get -y install openmpi-bin libopenmpi-dev \
-    && apt-get -y install swig  \
     && rm -rf /var/lib/apt/lists/* \
     && pip install --no-cache-dir -U pip \
-    && pip install --no-cache-dir --progress-bar off -U setuptools
+    && pip install ${PIP_OPTIONS} -U setuptools \
+    && pip install ${PIP_OPTIONS} Cython  # for automl/ConfigSpace
```
hvy, line 12: Is `ConfigSpace` still required? I know that it was used with `fanova` before.

crcrpar: Let me see whether we can remove ConfigSpace... I could not catch up recent active development enough :crying_cat_face:

hvy: No worries, of course (unless we can improve the situation somehow). To elaborate a bit, this is the recent change I was referring to https://github.com/optuna/optuna/pull/1106/files#diff-2eeaed663bd0d25b7e608891384b7298.

PR #1056, opened 2020-03-27
Add snippet of API for integration with Catalyst.

VladSkripniuk:
This PR addresses integration with Catalyst proposed in #1055

PR #1465, opened 2020-07-01
Mark `CatalystPruningCallback` integration as experimental.

hvy:
## Motivation

Follow up to https://github.com/optuna/optuna/pull/1056.

## Description of the changes

Marks `CatalystPruningCallback` as experimental as the Catalyst ecosystem is quite big and we might want to reserve ourselves for fast changes to the design. Also includes some minor cosmetics.

### DIFF
--- a/optuna/integration/catalyst.py
+++ b/optuna/integration/catalyst.py
@@ -2,6 +2,7 @@ from typing import Any  # NOQA
 
 import optuna
 
+from optuna._experimental import experimental
 from optuna._imports import try_import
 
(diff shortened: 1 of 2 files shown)


harupy:
It looks like the catalyst integration is missing in the documentation.

https://github.com/optuna/optuna/blob/master/docs/source/reference/integration.rst

hvy:
Sorry must've missed that. Thanks for the catch. Do you want to create a PR for that? If you're busy I could as well.

PR #1466, opened 2020-07-01
Update CONTRIBUTING.md

crcrpar:
This PR adds some more general instruction and writes down what actions we think are contributions.


hvy:
Modified the title a bit. It's actually mentioned in the contribution guidelines (maybe the part about filenames isn't clear) 😉

PR #1470, opened 2020-07-02
Replace `suggest_*uniform` in examples with `suggest_(int|float)`.

crcrpar:
Recently added `suggest_float` is more intuitive compared to `suggest_uniform` and we kind of recommend this; thus I replace `suggest_uniform` in examples with `suggest_float` in this PR.

__TODO__
- [x] replaces `suggest_loguniform` with `suggest_uniform(..., log=True)`
- [x] replaces `int(suggest_loguniform(...))` with `suggest_int(..., log=True)`


examples/samplers/simulated_annealing_sampler.py
```diff
@@ -5,7 +5,7 @@
 
 Note that this implementation isn't intended to be used for production purposes and
 has the following limitations:
-- The sampler only supports `UniformDistribution` (i.e., `Trial.suggest_uniform` method).
+- The sampler only supports `UniformDistribution` (i.e., `Trial.suggest_float` method).
```
HideakiImamura, line 8: This line should not be changed.

toshihikoyanase: I agree.  `Trial.suggest_float` corresponds to `LogUniformDistribution` in addition to `UniformDistribution`, and it may be confusing for readers. An alternative fix will be to add condition of the `log` argument, for example, as follows:

```
i.e., `Trial.suggest_float` method with `log=False`
or
i.e., `Trial.suggest_float` method without the `log` option
```

crcrpar: Okay, leave this line as it was :slightly_smiling_face:

Issue #1394, opened 2020-06-19
Remove description about `optuna study optimize` from examples and docs.


toshihikoyanase:
## Motivation

#1384 deprecated the `optuna study optimize` command, but many examples have descriptions about it at the top of the files. I think we can remove them for consistency.

## Description

Please remove the descriptions about `optuna study optimize` from example files.
- [ ] docs/source/faq.rst ([this section](https://optuna.readthedocs.io/en/latest/faq.html#how-can-i-use-two-gpus-for-evaluating-two-trials-simultaneously))
- [ ] docs/source/tutorial/cli.rst
- [ ] docs/source/tutorial/configurations.rst
- [x] examples/allennlp/allennlp_jsonnet.py
- [x] examples/allennlp/allennlp_simple.py
- [x] examples/catalyst_simple.py
- [x] examples/catboost_simple.py
- [x] examples/chainer_simple.py
- [x] examples/dask_ml_simple.py
- [x] examples/keras_simple.py
- [x] examples/lightgbm_simple.py
- [x] examples/lightgbm_tuner_cv.py
- [x] examples/lightgbm_tuner_simple.py
- [x] examples/mxnet_simple.py
- [x] examples/optuna_search_cv_simple.py
- [x] examples/pytorch_ignite_simple.py
- [x] examples/pytorch_lightning_simple.py
- [x] examples/pytorch_simple.py
- [x] examples/quadratic_simple.py
- [x] examples/skimage_lbp_simple.py
- [x] examples/sklearn_additional_args.py
- [x] examples/sklearn_simple.py
- [x] examples/tensorflow_eager_simple.py
- [x] examples/tensorflow_estimator_simple.py
- [x] examples/xgboost_simple.py

### Example

This is just an example of the fix. Please feel free to change it depending on the examples.

Before:
```python
We have the following two ways to execute this example:

(1) Execute this code directly.
    $ python quadratic_simple.py


(2) Execute through CLI.
    $ STUDY_NAME=`optuna create-study --storage sqlite:///example.db`
    $ optuna study optimize quadratic_simple.py objective --n-trials=100 --study-name $STUDY_NAME \
      --storage sqlite:///example.db

```

After:
```python
You can run this example as follows:
    $ python quadratic_simple.py

```

ritvik1512:
I would like to try to tackle this. Thanks!

ritvik1512:
[...]

PR #1472, opened 2020-07-02
Remove description about 'optuna study optimize' from docs.

bigbird555:
## Motivation

This PR is related to the Issue #1394 

## Description of the changes

As the above title says, I removed description about 'optuna study optimize' from docs.


bigbird555:
Hmm. I think we need to have a little more discussion about the Issue #1394.

Issue #1340, opened 2020-06-08
Avoid problem with length limit of 5000 in MLflow


PhilipMay:
## Motivation
When using the `MLflowCallback` it is possible that the user added `user_attrs` (like lists of float) that are longer then 5000 characters when converted to a `str`. This causes problems with MLflow which limits the length to 5000. See example below:

``` bash
[...]
  File "/home/smay/miniconda3/envs/py38/lib/python3.8/site-packages/mlflow/utils/validation.py", line 136, in _validate_length_limit
    raise MlflowException(
mlflow.exceptions.MlflowException: Tag value '[0.8562690322984875, 0.8544098885636596, 0.8544098885636596, 0.8544098885636596, 0.8544098885636596, 0.859181214773054, 0.86273086038245, 0.86273086038245, 0.86273086038245, 0.86273086038245, 0.86273086038245, 0.8562690322984875, 0.8544098885636596, ' had length 5276, which exceeded length limit of 5000
```

## Description
I suggest to check the strings if they are `len() > 5000` and cut them if needed. If this happens a warning should be printed.

I can provide a PR if wanted. Just give me feedback to my proposal please.

hvy:
I'm not familiar with how "tags" are used in `mlflow`, but would truncating the string representation mitigate your issue? If so, I think that's a reasonable approach. It'd be good to know if this number 5000 can be retrieved via some public API but I guess it's not a must. 

Back to the usage of "tags", would you mind briefly explaining why you need more than 5000 characters since I suppose is beyond what's expected by `mlflow`.

PhilipMay:
> It'd be good to know if this number 5000 can be retrieved via some public API but I guess it's not a must.
I will have a look into the MLflow code to retrieve the 5000 limit.

> would you mind briefly explaining why you need more than 5000 characters since I suppose is beyond what's expected by mlflow.
Sometimes I am logging a list of feature importances (SHAP values) to the user attr of optuna. Since I have 50000 features (bioinformatics) it is a list of 50000 floats. That is longer then 5000 characters when converted to a string. But there are also other less extreme cases where I want to log the individual results of a nested ccross validation.

PhilipMay:
I asked the MLflow guys if they want to fix this issue on their side: https://github.com/mlflow/mlflow/issues/2931

PhilipMay:
Yes - you are right @github-actions . @hvy what do you think about this?

hvy:
[...]

PR #1481, opened 2020-07-03
 Temporary fix for problem with length limit of 5000 in MLflow

PhilipMay:
PR for #1340

### DIFF
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -1,3 +1,4 @@
+import textwrap
 from typing import Optional
 
 import optuna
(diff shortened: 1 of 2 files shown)


optuna/integration/mlflow.py
```diff
@@ -121,4 +121,16 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
+            # This is a temporary fix on Optuna side. It avoids an error with user
+            # attributes that are too long. It should be fixed on MLflow side later.
+            # When it is fixed on MLflow side this codeblock can be removed.
+            # see https://github.com/optuna/optuna/issues/1340
+            # see https://github.com/mlflow/mlflow/issues/2931
+            max_mlflow_tag_length = 5000
+            for key, value in tags.items():
+                value = str(value)  # make sure it is a string
+                if len(value) > max_mlflow_tag_length:
+                    tags[key] = value[:max_mlflow_tag_length]
```
hvy, line 134: `textwrap.shorten` with `placeholder` could be used but it's a detail, https://docs.python.org/3/library/textwrap.html#textwrap.shorten.

PhilipMay: I will add that tonight.

PhilipMay: I made the change.

PhilipMay:
IMO this is ready to be merged.

PR #1463, opened 2020-07-01
Apply auto-formatting to `custom.css` to make it pretty and consistent.

harupy:
## Motivation

A follow-up for #1436 

## Description of the changes

Apply auto-formatting (with prettier) to `custom.css` to make it look pretty and consistent.

PR #1483, opened 2020-07-04
Wider main content container in the documentation.

hvy:
## Motivation

Rightmost space in the documentation is unused (for desktop). Content container can be wider aiding readability and reducing content height.

## Description of the changes

Increases the width of container of the main content in the docs. Tables for instance are more compact.

### DIFF
--- a/docs/source/_static/css/custom.css
+++ b/docs/source/_static/css/custom.css
@@ -41,6 +41,11 @@ em.sig-param > span:not(:nth-child(1)):not(:nth-child(2)) {
   border-left: solid 3px #6ab0de;
 }
 
+/* Main content */
+.wy-nav-content {
+    max-width: 1200px;
+}
+
 /* Sidebar header (and topbar for mobile) */
 .wy-side-nav-search,
 .wy-nav-top {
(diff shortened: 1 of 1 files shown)


HideakiImamura:
Thanks for the PR. In the current document for desktop, the navigation bar at the top of pages is in the middle of the page. I think it is better to move the navigation bar to align a wider content container.

スクリーンショット 2020-07-06 17 06 15

hvy:
Makes sense, fixes. Thanks for pointing that out. PTAL.

docs/source/_static/css/custom.css
```diff
@@ -41,6 +41,11 @@ em.sig-param > span:not(:nth-child(1)):not(:nth-child(2)) {
   border-left: solid 3px #6ab0de;
 }
 
+/* Main content */
+.wy-nav-content {
+    max-width: 1200px;
```
toshihikoyanase, line 46: A nit, but could you use two whitespaces for indent? The current code format is introduced by #1463 using https://prettier.io, and I think we need to discuss the formatting rule for HTML/CSS if we change it.
```suggestion
  max-width: 1200px;
```

hvy: Ops, should've noticed that. Thanks, applied.

hvy:
PTAL.

PR #1486, opened 2020-07-06
Add `TensorBoardCallback` to docs.

hvy:
## Motivation

`TensorBoardCallback` which is a public API should be documented.

## Description of the changes

Adds `TensorBoardCallback` to the documentation.

### DIFF
--- a/docs/source/reference/integration.rst
+++ b/docs/source/reference/integration.rst
@@ -6,7 +6,7 @@ optuna.integration
    :nosignatures:
 
    optuna.integration.ChainerPruningExtension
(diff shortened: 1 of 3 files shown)


docs/source/reference/integration.rst
```diff
@@ -6,7 +6,7 @@ optuna.integration
    :nosignatures:
 
    optuna.integration.ChainerPruningExtension
-   optuna.integration.ChainerMNStudy 
+   optuna.integration.ChainerMNStudy
```
hvy, line 9: Sorry, unrelated change. I believe it's acceptable but let me know if I should separate it out.

Issue #1488, opened 2020-07-06
Duplicate trial numbers on distributed optimization


nandnor93:
When I run multiple Optuna processes on cloud servers that have job scheduler, Optuna occasionally assigns duplicate trial numbers (`trial.number`) to multiple trials that started at almost the same time.

## Expected behavior

Each trial obtains a different number; as a result, the trial numbers among a study become sequential.

## Environment

- Optuna version: 1.5.0
- Python version: 3.8.3
- OS: A CentOS 7.7 container running on Singularity on a SUSE Linux Enterprise Server 12 SP4 host
- Other software versions: PostgreSQL 9.5.21, psycopg 2.8.5

## Error messages, stack traces, or logs

A simplified copy of the below-mentioned test program's output looks like:

```
[I 2020-07-07 00:07:24,134] Using an existing study with name 'test_optuna_collision_01' instead of creating a new one.

......


  #         x         y  NODENAME                       STARTED
  0:   -7.260   105.276    r6i5n2    2020-07-06 23:42:14.603784
  1:    8.803    33.677    r6i5n2    2020-07-06 23:42:14.776819
  2:   -1.653    21.654    r6i5n2    2020-07-06 23:42:14.887652
  3:    0.024     8.856    r6i5n2    2020-07-06 23:42:14.993142
  4:    6.834    14.701    r6i5n2    2020-07-06 23:42:15.106251
  5:   -4.047    49.663    r6i5n2    2020-07-06 23:42:15.215284
  6:    7.026    16.205    r6i5n2    2020-07-06 23:42:15.325813
  7:   -1.137    17.117    r6i5n2    2020-07-06 23:42:15.432296
  8:    0.887     4.466    r6i5n2    2020-07-06 23:42:15.539958
  9:   -9.392   153.554    r6i5n2    2020-07-06 23:42:15.692091
 10:    2.665     0.112    r6i5n2    2020-07-06 23:42:15.836753
 11:    3.273     0.074    r6i5n2    2020-07-06 23:42:15.986390
 12:    3.699     0.489    r6i5n2    2020-07-06 23:42:16.100333
 13:    3.481     0.232    r6i5n2    2020-07-06 23:42:16.216604
 14:    3.461     0.213    r6i5n2    2020-07-06 23:42:16.331074
 15:    5.481     6.154    r6i5n2    2020-07-06 23:42:16.445370
 16:    1.499     2.253    r6i5n2    2020-07-06 23:42:16.560353
[...]

PR #1490, opened 2020-07-07
Fix race condition for trial number computation.

hvy:
## Motivation

Fixes https://github.com/optuna/optuna/issues/1488.

## Description of the changes

Allows `count_past_trials` to circumvent repeatable read isolation level. See the comments in the code for details.

## Note

Verified the changes with MySQL, by inspection, by checking for unique trial numbers after running a distributed optimization. The bug is otherwise reproducible, simply running a lot of trials with.

```sql
SELECT COUNT(DISTINCT(number)) FROM trials;
```

```python
import optuna

    return sum(trial.suggest_float(f"x{i}", 0.0, 1.0) for i in range(n_params))

if __name__ == "__main__":
    n_params = 10
    n_trials = 200

    database = "duplicatenumber1488"
    storage = "mysql://root@localhost/duplicatenumber1488"
    study = optuna.create_study(sampler=TPESampler(), study_name=database, storage=storage, load_if_exists=True)

```

**About isolation levels in sqlalchemy for different dialect** 
https://docs.sqlalchemy.org/en/13/core/connections.html#sqlalchemy.engine.Connection.execution_options.params.isolation_level.

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -470,7 +470,7 @@ class RDBStorage(BaseStorage):
         try:
             # Locking within a study is necessary since the creation of a trial is not an atomic
             # operation. More precisely, the trial number computed in `_get_prepared_new_trial` is
-            # prone to race conditions within this lock.
+            # prone to race conditions without this lock.
             session.query(models.StudyModel).filter(
                 models.StudyModel.study_id == study_id
             ).with_for_update().one()
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -463,16 +463,45 @@ class RDBStorage(BaseStorage):
         """
 
         session = self.scoped_session()
+        try:
+            # Lock all trials belonging to this study. This might lead to a deadlock
+            # (`OperationalError`) in which case we will retry.
+            session.query(models.TrialModel).filter(
+                models.TrialModel.study_id == study_id
+            ).with_for_update().all()
 
-        # The following line is necessary because this method, i.e. the creation of a trial, is not
-        # an atomic operation. The default isolation level respects repeatable reads, meaning that
-        # multiple processes entering this code section (note that a session is already created)
-        # simultaneously will lead to multiple trials being created with same trial numbers (but
-        # different trial IDs). `count_past_trials` used to determine the trial number must be
-        # allowed to read uncommitted data in such cases to compute the correct trial number based
-        # the trials created by other simultaneous processes.
-        session.connection(execution_options={"isolation_level": "READ_UNCOMMITTED"})
+            trial = self._get_prepared_new_trial(study_id, template_trial, session)
+            self._commit(session)
+        except OperationalError:
+            # Retry after deadlock. This may happen in distributed environments.
+            session.rollback()
+            return self._create_new_trial(study_id, template_trial)
 
+        if template_trial:
+            frozen = copy.deepcopy(template_trial)
+            frozen.number = trial.number
+            frozen.datetime_start = trial.datetime_start
+            frozen._trial_id = trial.trial_id
+        else:
+            frozen = FrozenTrial(
+                number=trial.number,
+                state=trial.state,
+                value=None,
+                datetime_start=trial.datetime_start,
+                datetime_complete=None,
+                params={},
+                distributions={},
+                user_attrs={},
+                system_attrs={},
+                intermediate_values={},
+                trial_id=trial.trial_id,
+            )
+
+        return frozen
+
+    def _get_prepared_new_trial(
+        self, study_id: int, template_trial: Optional[FrozenTrial], session: orm.Session
+    ) -> models.TrialModel:
         # Ensure that that study exists.
         models.StudyModel.find_or_raise_by_id(study_id, session)
 
@@ -527,29 +556,7 @@ class RDBStorage(BaseStorage):
         trial.number = trial.count_past_trials(session)
         session.add(trial)
 
-        self._commit(session)
-
-        if template_trial:
-            frozen = copy.deepcopy(template_trial)
-            frozen.number = trial.number
-            frozen.datetime_start = trial.datetime_start
-            frozen._trial_id = trial.trial_id
-        else:
-            frozen = FrozenTrial(
-                number=trial.number,
-                state=trial.state,
-                value=None,
-                datetime_start=trial.datetime_start,
-                datetime_complete=None,
-                params={},
-                distributions={},
-                user_attrs={},
-                system_attrs={},
-                intermediate_values={},
-                trial_id=trial.trial_id,
-            )
-
-        return frozen
+        return trial
 
     def _update_trial(
         self,
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -462,22 +462,31 @@ class RDBStorage(BaseStorage):
 
         """
 
-        session = self.scoped_session()
+        # Retry a couple of times. Deadlocks may occur in distributed environments.
+        n_retries = 0
+        while True:
+            session = self.scoped_session()
 
-        try:
-            # Ensure that that study exists.
-            #
-            # Locking within a study is necessary since the creation of a trial is not an atomic
-            # operation. More precisely, the trial number computed in `_get_prepared_new_trial` is
-            # prone to race conditions without this lock.
-            models.StudyModel.find_or_raise_by_id(study_id, session, for_update=True)
+            try:
+                # Ensure that that study exists.
+                #
+                # Locking within a study is necessary since the creation of a trial is not an
+                # atomic operation. More precisely, the trial number computed in
+                # `_get_prepared_new_trial` is prone to race conditions without this lock.
+                models.StudyModel.find_or_raise_by_id(study_id, session, for_update=True)
 
-            trial = self._get_prepared_new_trial(study_id, template_trial, session)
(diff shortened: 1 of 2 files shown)


c-bata:
"Read Uncommitted" is lower isolation level than "Repeatable Read". It looks that this PR depends on "Dirty Reads", right?

> ScreenShot 2020-07-07 18 17 57
> https://en.wikipedia.org/wiki/Isolation_(database_systems)

Like this table said, "Dirty reads" **may occure** when using "Read Uncommitted". So it sounds that this fixes is not safe.

hvy:
Thanks @c-bata for you quick comment. 

I changed the logic to do row level locking on all trials for the given study, and retrying on failures, rather than relaxing the isolation level. Verified the logic with MySQL and PostgreSQL locally. Note that this has the downside of significantly making creation of trials slower. However, this cannot be helped given our "constraint" on the number that it must be unique, etc.

hvy:
<details><summary>Code</summary>
<p>

```python
import argparse
import math
import time

import optuna
import sqlalchemy


class Profile:
    def __enter__(self):
        self.start = time.time()
        return self

        self.end = time.time()

    def get(self):
        return self.end - self.start


    def objective(trial):
        return sum(
            [
                math.sin(trial.suggest_uniform("param-{}".format(i), 0, math.pi * 2))
                for i in range(n_param)
            ]
        )

    return objective


if __name__ == "__main__":
    parser.add_argument("mysql_user", type=str)
    parser.add_argument("mysql_host", type=str)
    args = parser.parse_args()

    storage_str = "mysql+pymysql://{}@{}/".format(args.mysql_user, args.mysql_host)

    print(f"| #params | #trials | time(sec) |")
    print(f"| ------- | ------- | --------- |")

    for n_param in [1, 2, 4, 8, 16, 32]:
        for n_trial in [1, 10, 100, 1000]:
            engine = sqlalchemy.create_engine(storage_str)
            conn = engine.connect()
            conn.execute("commit")
            database_str = "profile_storage_t{}_p{}".format(n_trial, n_param)
            try:
                conn.execute("drop database {}".format(database_str))
            except Exception:
                pass
            conn.execute("create database {}".format(database_str))
            conn.close()

            storage = optuna.storages.get_storage(storage_str + database_str)
            study = optuna.create_study(storage=storage, sampler=TPESampler())

            with Profile() as prof:
                study.optimize(
                    build_objective_fun(n_param), n_trials=n_trial, gc_after_trial=False,
                )

            print(f"| {n_param} | {n_trial} | {prof.get():.2f} |")
```

</p>
</details>

| #params | #trials | PR(sec) | `master`(sec) |
| ------- | ------- | ---- | ------------- |
| 1 | 1 | 0.02 | 0.01 |
| 1 | 10 | 0.12 | 0.11 |
| 1 | 100 | 1.60 | 1.18 |
| 1 | 1000 | 35.97 | 16.84 |
| 2 | 1 | 0.02 | 0.02 |
| 2 | 10 | 0.13 | 0.11 |
| 2 | 100 | 1.53 | 1.25 |
| 2 | 1000 | 38.03 | 20.17 |
| 4 | 1 | 0.02 | 0.02 |
| 4 | 10 | 0.14 | 0.12 |
| 4 | 100 | 1.83 | 1.53 |
| 4 | 1000 | 42.60 | 26.23 |
| 8 | 1 | 0.04 | 0.03 |
| 8 | 10 | 0.16 | 0.15 |
| 8 | 100 | 2.36 | 2.04 |
| 8 | 1000 | 52.53 | 36.11 |
| 16 | 1 | 0.08 | 0.06 |
| 16 | 10 | 0.20 | 0.20 |
| 16 | 100 | 3.21 | 2.98 |
| 16 | 1000 | 71.93 | 54.78 |
| 32 | 1 | 0.11 | 0.11 |
| 32 | 10 | 0.30 | 0.29 |
| 32 | 100 | 4.95 | 4.64 |
| 32 | 1000 | 110.76 | 93.52 |

optuna/storages/_rdb/storage.py
```diff
@@ -463,7 +463,45 @@ def _create_new_trial(
         """
 
         session = self.scoped_session()
+        try:
+            # Lock all trials belonging to this study. This might lead to a deadlock
+            # (`OperationalError`) in which case we will retry.
+            session.query(models.TrialModel).filter(
+                models.TrialModel.study_id == study_id
+            ).with_for_update().all()
```
ytsmiling, line 471: How about locking a entry in the `studies` table instead of locking the whole trials (though I haven't checked that it improves the latency)?

hvy: Ah, that might scale better with the number of trials. I'll try your suggestion.

hvy: Looks good.

| #params | #trials | time(sec) |
| ------- | ------- | --------- |
| 1 | 1 | 0.02 |
| 1 | 10 | 0.11 |
| 1 | 100 | 1.26 |
| 1 | 1000 | 17.41 |
| 2 | 1 | 0.02 |
| 2 | 10 | 0.12 |
| 2 | 100 | 1.32 |
| 2 | 1000 | 20.22 |
| 4 | 1 | 0.02 |
| 4 | 10 | 0.13 |
| 4 | 100 | 1.61 |
| 4 | 1000 | 25.82 |
| 8 | 1 | 0.04 |
| 8 | 10 | 0.16 |
| 8 | 100 | 2.07 |
| 8 | 1000 | 35.07 |
| 16 | 1 | 0.06 |
| 16 | 10 | 0.20 |
| 16 | 100 | 3.02 |
| 16 | 1000 | 55.23 |
| 32 | 1 | 0.12 |
| 32 | 10 | 0.30 |
| 32 | 100 | 4.76 |
| 32 | 1000 | 90.28 |

optuna/storages/_rdb/storage.py
```diff
@@ -463,7 +463,46 @@ def _create_new_trial(
...
+            session.query(models.StudyModel).filter(
+                models.StudyModel.study_id == study_id
+            ).with_for_update().one()
+
+            trial = self._get_prepared_new_trial(study_id, template_trial, session)
+            self._commit(session)
+        except OperationalError:
+            # Retry after deadlock. This may happen in distributed environments.
+            session.rollback()
+            return self._create_new_trial(study_id, template_trial)
```
c-bata, line 483: I think that we should not do retry after deadlock. If deadlock is happen, we should reconsider the logic. What do you think?

hvy: Thanks for you comment. Could you elaborate, also explain why that is? It seemed natural to me (https://dev.mysql.com/doc/refman/5.7/en/innodb-deadlocks-handling.html) but to be frank, it's not easily reproduced (encountered it a few times with MySQL).

c-bata: Hmm. I've ever thought it's unnatural like deadlock on multi-threading programming. But according to the MySQL documentation you shared, it seems to be wrong. Thank you for sharing useful link.

> but to be frank, it's not easily reproduced (encountered it a few times with MySQL).

BTW, I mind the reason why current implementation will raise deadlock such frequently.

hvy: > such frequently.

Just to clarify, it happens very **in**frequently.

c-bata: Oh, you are right. I misread your comment.

c-bata: Sounds no problem. Thank you for your reply.

hvy:
Thanks @c-bata as always. Regarding the performance hit, it's actually negligible now with @ytsmiling's suggestion of restricting the lock to a single row (at most), instead of all trials.

ytsmiling:
It's highly unlikely, but if trial_id is not assigned sequentially, the `count_past_trial` method can fail (duplicated number can be assigned). I'm okay about leaving the method as is, but if you'd like to change the implementation, I'll re-review this PR.

optuna/storages/_rdb/storage.py
```diff
@@ -464,9 +464,46 @@ def _create_new_trial(
 
         session = self.scoped_session()
 
-        # Ensure that that study exists.
-        models.StudyModel.find_or_raise_by_id(study_id, session)
+        try:
```
hvy, line 467: Memo: Try n (maybe 3) times and propagate the `OperationalError` in case they all fail. It should be easier to debug.

hvy:
Changed the logic to propagate sqlalchemy errors after 3 retries to reduce the risk of silencing unexpected errors and to aid debugging in case of what would previously have resulted in maxim recursion depth error.

hvy:
PTAL.

optuna/storages/_rdb/storage.py
```diff
@@ -462,11 +462,57 @@ def _create_new_trial(
 
         """
 
-        session = self.scoped_session()
+        # Retry a couple of times. Deadlocks may occur in distributed environments.
+        n_retries = 0
+        while True:
+            session = self.scoped_session()
```
toshihikoyanase, line 468: Just a clarification, but do we have to re-create session instances when we retry?

hvy: We can create one single session and reuse it over multiple transactions. I can change it if you prefer that. In my understand we don't have that well established norms for the lifetime of sessions.

toshihikoyanase: Thank you for your reply. 

>  we don't have that well established norms for the lifetime of sessions.

I see. This topic seems to be out of the scope of this PR, so please keep the implementation in this PR.

PR #1491, opened 2020-07-08
`FanovaImportanceEvaluator` as default importance evaluator.

hvy:
## Motivation

Now that both `FanovaImportanceEvaluator` and `MeanDecreaseImpurityImportanceEvaluator` both have the same dependencies, we can use the former, which is more sophisticated, as the default evaluator.

## Description of the changes

Changes the default importance evaluator.

## Note

It'd be preferable to have it included in v2.0 since it is the first major release with HPI.

### DIFF
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -5,7 +5,7 @@ from typing import Optional
 from optuna.importance._base import BaseImportanceEvaluator
 from optuna.importance._fanova import FanovaImportanceEvaluator
 from optuna.importance._mean_decrease_impurity import (  # NOQA
-    MeanDecreaseImpurityImportanceEvaluator,
+    MeanDecreaseImpurityImportanceEvaluator,  # NOQA
 )
 from optuna.study import Study
 
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -4,7 +4,6 @@ from typing import Optional
 
 from optuna.importance._base import BaseImportanceEvaluator
 from optuna.importance._fanova import FanovaImportanceEvaluator
-
 from optuna.importance._mean_decrease_impurity import (  # NOQA
     MeanDecreaseImpurityImportanceEvaluator,
 )
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -5,9 +5,9 @@ from typing import Optional
 from optuna.importance._base import BaseImportanceEvaluator
 from optuna.importance._fanova import FanovaImportanceEvaluator
 
-# fmt: off
-from optuna.importance._mean_decrease_impurity import MeanDecreaseImpurityImportanceEvaluator  # NOQA
-# fmt: on
+from optuna.importance._mean_decrease_impurity import (  # NOQA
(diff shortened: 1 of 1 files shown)


optuna/importance/__init__.py
```diff
@@ -3,8 +3,10 @@
 from typing import Optional
 
 from optuna.importance._base import BaseImportanceEvaluator
-from optuna.importance._fanova import FanovaImportanceEvaluator  # NOQA
-from optuna.importance._mean_decrease_impurity import MeanDecreaseImpurityImportanceEvaluator
+from optuna.importance._fanova import FanovaImportanceEvaluator
+
+# fmt: off
```
toshihikoyanase, line 8: I think we can remove these lines. Do you have specific reasons to place them?

hvy: It's for preventing black from creating line breaks. If the following import is broken up over several lines, then flake8 will instead complain on unused variables, since `# NOQA` isn't correctly parsed. I've not encountered this problem before, do you happen to know a better workaround?

toshihikoyanase: I see. L9 exceeds 100 character limit.
Then, how about placing `# fmt: on` after the L9 to activate the black again?

hvy: You're right it must be turned back on. Thanks, and fixed.

optuna/importance/__init__.py
```diff
@@ -3,8 +3,11 @@
...
+from optuna.importance._mean_decrease_impurity import MeanDecreaseImpurityImportanceEvaluator  # NOQA
+# fmt: on
```
c-bata, line 10: How about this?

```suggestion
from optuna.importance._mean_decrease_impurity import (  # NOQA
    MeanDecreaseImpurityImportanceEvaluator,
)
```

refs: https://gitlab.com/pycqa/flake8/-/issues/385

hvy: It really works... thanks a lot, including the reference. Let me apply this suggestion. @toshihikoyanase please let me know if you have any comments on this.

optuna/importance/__init__.py
hvy: Let me resolve this thread thanks to https://github.com/optuna/optuna/pull/1491#discussion_r451317817.

optuna/importance/__init__.py
toshihikoyanase: Cool! Thank you for the suggestion.

hvy:
Seems like a `flake8` version dependent matter. It passes with `flake8-3.7.9` but not with `flake8-3.8.3` (as seen in the [CI build](https://app.circleci.com/pipelines/github/optuna/optuna/1563/workflows/64cb6c76-3430-46f0-90da-cfdf978e9a73/jobs/48784/steps). The cause is the comma, appended by black.

hvy:
Added another `# NOQA` (E126, H301) to work around it. PTAL.

```shell
./optuna/importance/__init__.py:8:9: E126 continuation line over-indented for hanging indent
./optuna/importance/__init__.py:8:48: H301: one import per line
1     E126 continuation line over-indented for hanging indent
1     H301: one import per line
```

PR #1495, opened 2020-07-09
Use `circleci/python` for dev image and install RDB servers.

toshihikoyanase:
## Motivation
The current docker images do not contain CircleCI related files, so we cannot use it to execute CI jobs locally.
In addition, it lacks DB servers and we need to install them to test with such servers.

## Description of the changes
This PR will apply the following changes:
- Change base image from `python` to `circleci/python`.
- Add DBs, i.e., `default-mysql-server`, `default-mysql-client`, `postgresql`, and `postgresql-client`
- Python DB bindings

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -35,16 +35,19 @@ jobs:
         export TAG_NAME="py${{ matrix.python_version }}"
         if [ "${{ github.event_name }}" = 'release' ]; then
           export TAG_NAME="${{ github.event.release.tag_name }}-${TAG_NAME}"
+          export BASE_IMAGE="python"
         fi
         if [ "${{matrix.build_type}}" = 'dev' ]; then
           export TAG_NAME="${TAG_NAME}-dev"
+          export BASE_IMAGE="circleci/python"
         fi
         echo "::set-env name=HUB_TAG::${DOCKER_HUB_BASE_NAME}:${TAG_NAME}"
 
 
     - name: Build the Docker image
       run: |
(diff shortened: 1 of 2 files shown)


toshihikoyanase:
@crcrpar Thank you for your review.

> could you tell me how to use the images on DockerHub for local CircleCI execution?

Sorry for the lack of description.
I'll simply edit the `.circleci/config.yml` to replace the docker image. For example, we can use the DockerHub image to skip the installation process as follows:

```
  checks:
    docker:
      - image: toshihikoyanase/optuna:py3.7-dev
      # - image: circleci/python:3.7
    steps:
      - checkout

      - run: &checkout-merge-master
          name: checkout-merge-master
          command: |
            set -ex
            if [[ -n "${CIRCLE_PR_NUMBER}" ]]; then
                FETCH_REFS="${FETCH_REFS} +refs/pull/${CIRCLE_PR_NUMBER}/merge:pr/${CIRCLE_PR_NUMBER}/merge"
                git fetch -u origin ${FETCH_REFS}
                git checkout "pr/${CIRCLE_PR_NUMBER}/merge"
            fi

      # - run:
      #     name: install
      #     command: |
      #       python -m venv venv || virtualenv venv
      #       . venv/bin/activate
      #       pip install -U pip
      #       pip install --progress-bar off .[checking]

      - run:
          name: black
          command: |
            # . venv/bin/activate
            black . --line-length 99 --check --exclude="docs"
```

At first, I thought we could also use it in daily CI to skip the installation, but we can't actually. It will not work if developers change the dependencies in their PRs.

Issue #1496, opened 2020-07-09
Retry insertion after deadlock when using RDB backend.


ytsmiling:
## Background
Some DBs can cause deadlock even when you try to insert a single row (see e.g., [MySQL InnoDB](https://dev.mysql.com/doc/refman/5.7/en/innodb-deadlocks-handling.html)). The documentation notes as follows.
```
Always be prepared to re-issue a transaction if it fails due to deadlock.
```
It seems that `sqlalchemy` does not automatically retry, and we need to handle the dead-lock like [these lines](https://github.com/optuna/optuna/pull/1490/files#diff-9aa9e205e3a2038ae9077f72d10abac1R467).

## Request
Add retry-codes for every code block which performs insertion. In the current `optuna` codebase, it will be safe to always retry them.

hvy:
Might be a good idea to address this after https://github.com/optuna/optuna/pull/1628.

PR #1497, opened 2020-07-09
Install RDB servers and their bindings for storage tests.

toshihikoyanase:
## Motivation
This PR is related to the changes in #1495. Recently, we received some issues on concurrent access to the `RDBStorage` such as #1488 and #1496. However, we do not use DB servers in any test case. To detect such issues, we need to run tests of storages using various storage backends.

## Description of the changes
This PR introduces MySQL and PostgreSQL to cover a wider variety of user environments.

## Update on June 14

- Replace apt packages with docker images.
  - The RDB servers failed on startup.
  - The [official document](https://circleci.com/docs/2.0/postgres-config/) recommends docker containers to launch RDB servers.
- Add test cases
  - A test case to reproduce the issue #1488.
  - Another test case to run many trials.
  - Currently, they fail due to the issue on #1496, but it will be resolved by #1509 
  - They may fail due to another issue reported in #1508 after merging #1509
- The test cases can be run as follows: `circleci build --job tests-rdbstorage`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,9 +11,9 @@ workflows:
       - tests-python37
       - tests-python36
       - tests-python35
+      - tests-rdbstorage
 
 jobs:
-
   # Lint and static type checking
 
   checks:
@@ -123,8 +123,6 @@ jobs:
           command: |
             sudo apt-get update
             sudo apt-get -y install openmpi-bin libopenmpi-dev
-            sudo apt-get -y install default-mysql-server default-mysql-client
-            sudo apt-get -y install postgresql postgresql-client
 
             python -m venv venv || virtualenv venv
             . venv/bin/activate
@@ -138,9 +136,6 @@ jobs:
 
             # Install all dependencies needed for testing.
             pip install --progress-bar off $(ls dist/*.tar.gz)[testing] -f https://download.pytorch.org/whl/torch_stable.html
-
-            # Install RDB bindings for storage tests.
-            pip install --progress-bar off PyMySQL cryptography psycopg2-binary
 
       - run:
           name: install-codecov
@@ -236,3 +231,39 @@ jobs:
                 --ignore tests/integration_tests/test_catalyst.py
 
       - run: *tests-mn
+
+  tests-rdbstorage:
+    docker:
+      - image: circleci/python:3.7
+      - image: circleci/redis:5.0.7
+      - image: circleci/mysql:5.7
+        environment:
+          MYSQL_DATABASE: optunatest
+          MYSQL_USER: user
+          MYSQL_PASSWORD: test
+      - image: circleci/postgres:10.1-alpine
+        environment:
+          POSTGRES_USER: user
+          POSTGRES_DB: optunatest
+          POSTGRES_PASSWORD: test
+    steps:
+      - checkout
+
+      - run: *checkout-merge-master
+
+      - run: *install
+
+      - run: &tests-rdb
+          name: tests-rdb
+          command: |
+            . venv/bin/activate
+            # Install RDB bindings for storage tests.
+            pip install --progress-bar off PyMySQL cryptography psycopg2-binary
+
+            export TEST_DB_URL=mysql+pymysql://user:test@127.0.0.1/optunatest
+            pytest tests/storages_tests/rdb_tests/test_with_server.py
+
+            export TEST_DB_URL=postgres+psycopg2://user:test@127.0.0.1/optunatest
+            pytest tests/storages_tests/rdb_tests/test_with_server.py
+          environment:
+            OMP_NUM_THREADS: 1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -253,7 +253,7 @@ jobs:
       - run: *install
 
       - run: &install-rdb-bindings
-          name: tests-mysql
+          name: install-rdb-bindings
           command: |
             . venv/bin/activate
             pip install --progress-bar off PyMySQL cryptography psycopg2-binary
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -274,4 +274,4 @@ jobs:
             pytest tests/storages_tests/rdb_tests/test_with_server.py
           environment:
             OMP_NUM_THREADS: 1
-            TEST_DB_URL: postgres+psycopg2://user:test@127.0.0.1/optunatest
+            TEST_DB_URL: postgresql+psycopg2://user:test@127.0.0.1/optunatest
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -252,17 +252,26 @@ jobs:
 
       - run: *install
 
-      - run: &tests-rdb
-          name: tests-rdb
+      - run: &install-rdb-bindings
+          name: tests-mysql
           command: |
             . venv/bin/activate
-            # Install RDB bindings for storage tests.
             pip install --progress-bar off PyMySQL cryptography psycopg2-binary
 
-            export TEST_DB_URL=mysql+pymysql://user:test@127.0.0.1/optunatest
-            pytest tests/storages_tests/rdb_tests/test_with_server.py
-
-            export TEST_DB_URL=postgres+psycopg2://user:test@127.0.0.1/optunatest
+      - run: &tests-mysql
+          name: tests-mysql
+          command: |
+            . venv/bin/activate
             pytest tests/storages_tests/rdb_tests/test_with_server.py
           environment:
             OMP_NUM_THREADS: 1
+            TEST_DB_URL: mysql+pymysql://user:test@127.0.0.1/optunatest
+
+      - run: &tests-postgresql
+          name: tests-postgresql
+          command: |
+            . venv/bin/activate
+            pytest tests/storages_tests/rdb_tests/test_with_server.py
+          environment:
+            OMP_NUM_THREADS: 1
+            TEST_DB_URL: postgres+psycopg2://user:test@127.0.0.1/optunatest
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -235,7 +235,6 @@ jobs:
   tests-rdbstorage:
     docker:
       - image: circleci/python:3.7
-      - image: circleci/redis:5.0.7
       - image: circleci/mysql:5.7
         environment:
           MYSQL_DATABASE: optunatest
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,6 +123,8 @@ jobs:
           command: |
             sudo apt-get update
             sudo apt-get -y install openmpi-bin libopenmpi-dev
+            sudo apt-get -y install default-mysql-server default-mysql-client
+            sudo apt-get -y install postgresql postgresql-client
 
             python -m venv venv || virtualenv venv
             . venv/bin/activate
@@ -136,6 +138,9 @@ jobs:
 
             # Install all dependencies needed for testing.
             pip install --progress-bar off $(ls dist/*.tar.gz)[testing] -f https://download.pytorch.org/whl/torch_stable.html
+
+            # Install RDB bindings for storage tests.
+            pip install --progress-bar off PyMySQL cryptography psycopg2-binary
 
       - run:
           name: install-codecov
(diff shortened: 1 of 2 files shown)


crcrpar:
Changes look okay.

> However, we do not use DB servers in any test case. To detect such issues, we need to run tests of storages using various storage backends.

Do we have enough tests to run at this point, or do you mean we are going to write those kinds of tests?

ytsmiling:
I think we don't have enough tests and need to add them (in other PRs).

toshihikoyanase:
I added test cases and noticed that the servers installed via apt did not work correctly. So, I used docker containers instead of them. For further details, please refer to https://github.com/optuna/optuna/pull/1497#issue-446628834.

tests/storages_tests/rdb_tests/test_with_server.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+    # Check intermediate values.
+    assert all((len(trial.intermediate_values) == 2 for trial in trials))
+    assert all((trial.params["x"] == trial.intermediate_values[0] for trial in trials))
+    assert all((trial.params["y"] == trial.intermediate_values[1] for trial in trials))
+
+    # Check attrs.
+    assert all((trial.user_attrs["x"] == trial.params["x"] for trial in trials))
+    assert all((trial.system_attrs["y"] == trial.params["y"] for trial in trials))
```
ytsmiling, line 71: I think this should be checked using `np.isclose` because ``user_attrs`` and ``system_attrs`` are once converted into `json` and then converted back into `float`.

tests/storages_tests/rdb_tests/test_with_server.py
```diff
@@ -0,0 +1,98 @@
...
+    N_TRIALS = 1000
+    study = optuna.create_study(study_name=_STUDY_NAME, storage=storage,)
+    # Run optimization
+    study.optimize(objective, n_trials=N_TRIALS)
+
+    trials = study.trials
+    assert len(trials) == N_TRIALS
+
+    _check_trials(trials)
+
```
ytsmiling, line 84: I think we should test both `study.trials` and `optuna.create_study(study_name=_STUDY_NAME, storage=storage,).trials` to check the communication with DB.

tests/storages_tests/rdb_tests/test_with_server.py
```diff
@@ -0,0 +1,98 @@
...
+    return storage_url
+
+
+def _check_trials(trials: Sequence[optuna.trial.FrozenTrial]) -> None:
+    # Check trial states.
+    assert all((trial.state == optuna.trial.TrialState.COMPLETE for trial in trials))
+
+    # Check trial values and params.
+    assert all(("x" in trial.params for trial in trials))
+    assert all(("y" in trial.params for trial in trials))
```
hvy, line 55: I think you can skip the extra parenthesis(es).
```suggestion
    assert all("y" in trial.params for trial in trials)
```

tests/storages_tests/rdb_tests/test_with_server.py
```diff
@@ -0,0 +1,98 @@
...
+    assert all((len(trial.intermediate_values) == 2 for trial in trials))
+    assert all((trial.params["x"] == trial.intermediate_values[0] for trial in trials))
+    assert all((trial.params["y"] == trial.intermediate_values[1] for trial in trials))
+
+    # Check attrs.
+    assert all((trial.user_attrs["x"] == trial.params["x"] for trial in trials))
+    assert all((trial.system_attrs["y"] == trial.params["y"] for trial in trials))
+
+
+def test_many_trials(storage: str) -> None:
```
hvy, line 74: More of a question but what kinds of problem is this test case trying to find with the large number of trials (that's not covered by existing unit tests)? I first thought that `tests-rdbstorage` was intended for distributed/parallel optimization but I understand that's not the case.

.circleci/config.yml
```diff
@@ -231,3 +231,39 @@ jobs:
                 --ignore tests/integration_tests/test_catalyst.py
 
       - run: *tests-mn
+
+  tests-rdbstorage:
+    docker:
+      - image: circleci/python:3.7
+      - image: circleci/redis:5.0.7
```
hvy, line 238: Is redis still used?

.circleci/config.yml
```diff
@@ -231,3 +231,39 @@ jobs:
...
+          command: |
+            . venv/bin/activate
+            # Install RDB bindings for storage tests.
+            pip install --progress-bar off PyMySQL cryptography psycopg2-binary
+
+            export TEST_DB_URL=mysql+pymysql://user:test@127.0.0.1/optunatest
+            pytest tests/storages_tests/rdb_tests/test_with_server.py
+
+            export TEST_DB_URL=postgres+psycopg2://user:test@127.0.0.1/optunatest
+            pytest tests/storages_tests/rdb_tests/test_with_server.py
```
hvy, line 267: A detail but how about using separate `run`s for different DB backends since they're quite independent?

tests/storages_tests/rdb_tests/test_with_server.py
```diff
@@ -0,0 +1,98 @@
...
+    trials = study.trials
+    assert len(trials) == N_TRIALS
+
+    _check_trials(trials)
+
+
+def test_multiprocess(storage: str) -> None:
+    n_workers = 8
+    study_name = _STUDY_NAME
+    storage = storage
```
hvy, line 89: Sorry but what is the intent of this line?

tests/storages_tests/rdb_tests/test_with_server.py
```diff
@@ -0,0 +1,98 @@
...
+    study_name = args[0]
+    storage = args[1]
+    # Create a study
+    study = optuna.create_study(study_name=study_name, storage=storage, load_if_exists=True,)
+    # Run optimization
+    study.optimize(objective, n_trials=20)
+
+
+@pytest.fixture
+def storage() -> str:
```
hvy, line 38: Might be controversial so you can ignore it but I personally think that the following increases the readability.
```suggestion
def storage_url() -> str:
```

tests/storages_tests/rdb_tests/test_with_server.py
ytsmiling: I think this test becomes useful after [this suggestion](https://github.com/optuna/optuna/pull/1497/files#r454069756) is merged. However, I also agree with that we don't need such many trials to test storage's behavior.

tests/storages_tests/rdb_tests/test_with_server.py
toshihikoyanase: That makes sense. Applied in https://github.com/optuna/optuna/pull/1497/commits/b73d1d99f07cf4dafec90d28376a1b1fd5aa17be.

tests/storages_tests/rdb_tests/test_with_server.py
toshihikoyanase: Thank you for your suggestion. I added the check for loaded study in https://github.com/optuna/optuna/pull/1497/commits/776b45d90ea35ea44b5837da32fe00bde2d19cd6.

tests/storages_tests/rdb_tests/test_with_server.py
toshihikoyanase: I thought study with many trials causes some issues, but I may get confused with performance measurement with a wall-time clock. I renamed the method to `test_loaded_study` and reduce the number of trials in https://github.com/optuna/optuna/pull/1497/commits/c8a3cbf4fe4ca4cad6065286fbce51e486581734.

HideakiImamura:
I will re-review this PR after the above discussions.

.circleci/config.yml
toshihikoyanase: Oops. I removed it in https://github.com/optuna/optuna/pull/1497/commits/1a55a9515b34866330572c4f9e25c44f05e84d9f.

.circleci/config.yml
toshihikoyanase: Sounds good! I divided the step in https://github.com/optuna/optuna/pull/1497/commits/15356075ed94d2fccd2c74ba61cbb9395016e09a.

tests/storages_tests/rdb_tests/test_with_server.py
toshihikoyanase: Thank you for pointing it out. I removed in https://github.com/optuna/optuna/pull/1497/commits/6222418f6de3e50821f1f44786a8a9ba56be75bf.

.circleci/config.yml
hvy: Thanks!

tests/storages_tests/rdb_tests/test_with_server.py
toshihikoyanase: Thank you for your suggestion. Applied!

tests/storages_tests/rdb_tests/test_with_server.py
toshihikoyanase: Thank you for your suggestion. I also applied the same changes to the other assertions in https://github.com/optuna/optuna/pull/1497/commits/f9a2d2f74e2bf4e5bd204ce22794a6b264afb777.

tests/storages_tests/rdb_tests/test_with_server.py
```diff
@@ -0,0 +1,113 @@
...
+    # Run optimization
+    study.optimize(objective, n_trials=N_TRIALS)
+
+    trials = study.trials
+    assert len(trials) == N_TRIALS
+
+    _check_trials(trials)
+
+    # Create a new study to confirm the study can load trial properly.
+    loaded_study = optuna.create_study(study_name=_STUDY_NAME, storage=storage_url,)
```
hvy, line 98: Will this work without `load_if_exists=True`? I otherwise think we can use `load_study`.

tests/storages_tests/rdb_tests/test_with_server.py
hvy: I'm wondering if this isn't covered by `test_multiprocess` already where a study is re-created using `load_study`.

tests/storages_tests/rdb_tests/test_with_server.py
toshihikoyanase: Thanks. It was my careless mistake. I replace it with `load_study` in https://github.com/optuna/optuna/pull/1497/commits/11c14188381c559cdf8dc5e2354161f4d2dd41c5.

tests/storages_tests/rdb_tests/test_with_server.py
ytsmiling: I think we can distinguish whether test failures are caused by multi-worker related issue or worker-database communications when they occur.
Additionally, we anyway need to run a study with a single worker somewhere before `tes_multiprocess` or add a CLI command to .circleci, otherwise multi-worker test fails because multiple workers try to create tables at the same time.

hvy: Ah, that makes sense, thanks. 

As for the table creation, we might want to leave a comment to avoid future breakage (e.g. just placing the function definitions in a different order might break)?

ytsmiling: I agree that we'd better leave a comment.

toshihikoyanase: That sounds reasonable. I added a comment on table creation in https://github.com/optuna/optuna/pull/1497/commits/3d38a589305f394b13acd7813a3298960090e937 .
>     # Please create the tables by placing this function before the multi-process tests.

.circleci/config.yml
```diff
@@ -231,3 +231,47 @@ jobs:
...
+          POSTGRES_PASSWORD: test
+    steps:
+      - checkout
+
+      - run: *checkout-merge-master
+
+      - run: *install
+
+      - run: &install-rdb-bindings
+          name: tests-mysql
```
hvy, line 256: Ops, a small one.
```suggestion
          name: install-rdb-bindings
```

toshihikoyanase: Thanks! Applied.

hvy:
Maybe we should run more trials after all given https://github.com/optuna/optuna/pull/1514#discussion_r454128925, and/or simply repeat the tests (e.g. with https://pypi.org/project/pytest-repeat/)?

toshihikoyanase:
> Maybe we should run more trials after all given #1514 (comment), and/or simply repeat the tests (e.g. with https://pypi.org/project/pytest-repeat/)?

That sounds reasonable since the problems caused by parallel processing occurs stochastically. 
I'll make a followup PR for it. Thanks!

hvy:
Tests seemingly passed after rebase 😄

PR #1498, opened 2020-07-09
Fix `CachedStorage` skipping trial param row insertion on cache miss.

hvy:
## Motivation

Fixes https://github.com/optuna/optuna/issues/1488#issuecomment-655428975.

## Description of the changes

Makes sure parameters are flushed/committed even for the first trial that is a "cache miss". 

## Note

This is only an issue for the first trial in a distributed environment/continuing optimization on an already started/finished study.

### DIFF
--- a/optuna/storages/_cached_storage.py
+++ b/optuna/storages/_cached_storage.py
@@ -229,9 +229,7 @@ class _CachedStorage(BaseStorage):
                 if cached_dist:
                     distributions.check_distribution_compatibility(cached_dist, distribution)
                 else:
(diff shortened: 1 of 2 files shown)


ytsmiling:
Thank you for creating this PR.
I think the addition in the `check_or_set_param_distribution`. Otherwise, when many workers try to add a new parameter with conflicting distributions, they can pass the `check_param_distribution`.

hvy:
Ah, right. You mean that we must insert after checking for compatibility to let the other workers know. Thanks.

hvy:
PTAL.

Issue #1251, opened 2020-05-17
Enable `objective(study.best_trial)`


liori:
I find it useful to be able to re-run trials to investigate them further, e.g. with more debugging/data reporting enabled. I thought that it would be natural if I could manually run a trial from a study e.g. by:

    my_objective(my_study.best_trial, more_reporting=True)

but given that the objective function calls various `suggest_*` methods, I get `AttributeError: 'FrozenTrial' object has no attribute 'suggest_int'`. And instead I need to do:

    my_objective(optuna.trial.FixedTrial(my_study.best_trial.params, my_study.best_trial.number), more_reporting=True)

Making `my_objective(my_study.best_trial)` work would be a small, but useful improvement in my workflow.

HideakiImamura:
Thanks for your proposal! I totally agree with your idea.
To realize the feature, we need to add some `suggest_*` methods for `FrozenTrial`. Let me give this issue the contribution-welcome label.

HideakiImamura:
I think there is room for discussion on how to realize this feature.

1. Adding some `suggest_*` methods to `FrozenTrial`.
2. Internally, converting `FrozenTrial` to `FixedTrial`.

In my opinion, the first choice is preferable in terms of simplicity. Also, if we add the `suggest_*` methods to the `FrozenTrial`, I think that `FixedTrial` may be deprecated. I would like to discuss these points.

hvy:
1 might be most simple from the user perspective but I'm wondering if it isn't a bit overkill for something that can be achieved quite easily `objective(optuna.trial.FixedTrial(study.best_trial.params))`. But I agree `objective(study.best_trial)` is ideal 🤔

nzw0301:
I agree with @hvy. Option 1 is simpler but I suppose `FronzenTrial` and `FixedTrial` share many methods by adding `suggest_*` methods to `FronzenTrial`.  I also like the 2nd option, e.g.,`my_objective(best_trial.to_fixed_trial())`.

Let me put some my first impression of this proposal.  In my experience, it sounds a little bit similar to `refit` mode in [`sklearn.model_selection.GridSearchCV`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html), where the model with the best hyper-parameters fits on the training data after determining the best hyper-parameters by performing CV. But it may not be necessary to solve this issue.

nzw0301:
[...]

PR #1503, opened 2020-07-09
Add sampling functions to `FrozenTrial`

nzw0301:
## Motivation

Solve #1251

## Description of the changes

Add sampling functions to `FrozenTrial` by inheriting `BaseTrial` as in #1251


optuna/storages/_rdb/storage.py
```diff
@@ -671,7 +671,7 @@ def _update_trial(
                 .all()
             )
             param_keys = set(param.param_name for param in trial_param)
-            trial_model.params.extend(
+            trial_model._params.extend(
```
HideakiImamura, line 674: This `trial_model` is not an instance of `FrozenTrial` but that of `TrialModel`. It is good to revert this change.

optuna/trial/_frozen.py
```diff
@@ -200,6 +208,171 @@ def duration(self):
...
+    def datetime_start(self):
+        # type: () -> Optional[datetime.datetime]
+
+        return self._datetime_start
+
+    @property
+    def number(self) -> int:
+
+        return self._number
+
```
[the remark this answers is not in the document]
HideakiImamura: I think it is good to re-order these getters to align the order of the arguments of `FrozenTrial`.

nzw0301:
@HideakiImamura Thank you for kindful comments! (Hopefully) I'll update this PR in the weekend...

optuna/trial/_frozen.py
```diff
@@ -200,6 +208,171 @@ def duration(self):
...
+        if step is not None:
+            if log:
+                raise ValueError("The parameter `step` is not supported when `log` is True.")
+            else:
+                return self._suggest(name, DiscreteUniformDistribution(low=low, high=high, q=step))
+        else:
+            if log:
+                return self._suggest(name, LogUniformDistribution(low=low, high=high))
+            else:
+                return self._suggest(name, UniformDistribution(low=low, high=high))
```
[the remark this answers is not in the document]
nzw0301: Oops. Thanks!

optuna/trial/_frozen.py
```diff
@@ -200,6 +208,171 @@ def duration(self):
...
+
+    def set_param(self, params):
+        # type: (Dict[str, Any]) -> None
+
+        self._params = params
+
+    def set_distributions(self, distributions):
+        # type: (Dict[str, BaseDistribution]) -> None
+
+        self._distributions = distributions
```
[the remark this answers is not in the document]
nzw0301: I didn't such a decorator. Thanks! I will replace them as your suggestion.

optuna/storages/_rdb/storage.py
nzw0301: Indeed this part caused errors... Thanks a lot!

optuna/storages/_cached_storage.py
```diff
@@ -236,11 +236,11 @@ def set_trial_param(
 
                 params = copy.copy(cached_trial.params)
                 params[param_name] = distribution.to_external_repr(param_value_internal)
-                cached_trial.params = params
+                cached_trial.set_param(params)
```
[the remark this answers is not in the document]
nzw0301: I have reverted them. thanks!

optuna/trial/_frozen.py
```diff
@@ -200,6 +208,171 @@ def duration(self):
...
+                return self._suggest(name, DiscreteUniformDistribution(low=low, high=high, q=step))
+        else:
+            if log:
+                return self._suggest(name, LogUniformDistribution(low=low, high=high))
+            else:
+                return self._suggest(name, UniformDistribution(low=low, high=high))
+
+
+    def suggest_uniform(self, name, low, high):
+        # type: (str, float, float) -> float
```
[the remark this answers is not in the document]
nzw0301: Sure!

optuna/trial/_frozen.py
```diff
@@ -166,29 +170,60 @@ def _validate(self):
...
+    def number(self) -> int:
+        return self._number
+
+    @number.setter
+    def number(self, value: int) -> None:
+        self._number = value
+
+    @property
+    def datetime_start(self) -> Optional[datetime.datetime]:
+        return self._datetime_start
```
HideakiImamura, line 182: We need a setter to `datetime_start`.

optuna/trial/_frozen.py
```diff
@@ -166,29 +170,60 @@ def _validate(self):
...
 
     @property
-    def last_step(self):
-        # type: () -> Optional[int]
+    def user_attrs(self) -> Dict[str, Any]:
+        return self._user_attrs
+
+    @property
+    def system_attrs(self) -> Dict[str, Any]:
+        return self._system_attrs
```
HideakiImamura, line 216: Ditto.

nzw0301:
I plan to add the following test to `test_trial.py` for this new feature:

```python
@parametrize_storage
def test_frozen_trial_sampling(storage_init_func):
    # type: (Callable[[], storages.BaseStorage]) -> None

    def objective(trial):
        # type: (BaseTrial) -> float

        trial.suggest_uniform("a", 0, 10)
        trial.suggest_loguniform("b", 0.1, 10)
        trial.suggest_discrete_uniform("c", 0, 10, 1)
        trial.suggest_int("d", 0, 10)
        trial.suggest_categorical("e", ["foo", "bar", "baz"])
        trial.suggest_int("f", 1, 10, log=True)

        return 1.0

    study = create_study(storage_init_func())

    best_trial = study.best_trial
    objective(best_trial)

    assert best_trial._suggested_params == study.best_trial.params
```

nzw0301:
I'll look at the failing test cases in `storages_tests`. It may take a time to see it (I'm not familiar to the storage part...)

optuna/trial/_frozen.py
```diff
@@ -166,29 +170,72 @@ def _validate(self):
                 )
 
     @property
-    def distributions(self):
-        # type: () -> Dict[str, BaseDistribution]
+    def number(self) -> int:
+        return self._number
```
HideakiImamura, line 174: In the Optuna coding style, we should insert a blank line as follows. Could you insert blank lines to other places?
```suggestion
    def number(self) -> int:
   
        return self._number
```

optuna/trial/_frozen.py
```diff
@@ -166,29 +170,72 @@ def _validate(self):
...
+    def datetime_start(self, value: Optional[datetime.datetime]) -> None:
+        self._datetime_start = value
+
+    def set_user_attr(self, key: str, value: Any) -> None:
+
+        self._user_attrs[key] = value
+
+    def set_system_attr(self, key: str, value: Any) -> None:
+
+        self._system_attrs[key] = value
```
HideakiImamura, line 194: It might be helpful for code readers to move these lines after `should_prune`. The order of the functions should match with that of `optuna/trial/_trial.py`.

optuna/trial/_frozen.py
HideakiImamura, line 174: These all getters and setters should be placed after public functions (`set_system_attr`). The function order should match with that of `optuna/trial/_trial.py`.

optuna/trial/_frozen.py
nzw0301: Absolutely!

By the way, I found that In `fixedTrial.py` and `Trial.py`, `_suggest`'s position is not consistent:

https://github.com/optuna/optuna/blob/0ec8ddc1809f7d3f315b119c51a8090452364a0a/optuna/trial/_fixed.py#L130
`_suggest` function appears just after public `suggest_*`, then other public functions are defined after `_suggest`.

https://github.com/optuna/optuna/blob/0ec8ddc1809f7d3f315b119c51a8090452364a0a/optuna/trial/_trial.py#L660
`_suggest` function appears after all public function.

nzw0301:
The order of tests of `FrozenTrial` matches the order of definition in `FrozenTrial`.

optuna/trial/_frozen.py
HideakiImamura: Thanks for the good catch! I think it is better to align the `optuna/trial/_trial.py`. (The `optuna/trial/_fixed.py` should be fixed in another PR.)

tests/test_trial.py
```diff
@@ -764,9 +765,325 @@ def test_study_id():
...
+        trial.suggest_categorical("e", ["foo", "bar", "baz"])
+        trial.suggest_int("f", 1, 10, log=True)
+
+        return 1.0
+
+    study = create_study(storage_init_func())
+    study.optimize(objective, n_trials=1)
+
+    best_trial = study.best_trial
+    objective(best_trial)
```
HideakiImamura, line 865: How about checking the objective value is equal to the `best_trial.value`?
```suggestion
    v = objective(best_trial)
    assert v == best_trial.value
```

tests/test_trial.py
```diff
@@ -764,9 +765,325 @@ def test_study_id():
...
+    def objective(trial: BaseTrial) -> float:
+
+        trial.suggest_uniform("a", 0, 10)
+        trial.suggest_loguniform("b", 0.1, 10)
+        trial.suggest_discrete_uniform("c", 0, 10, 1)
+        trial.suggest_int("d", 0, 10)
+        trial.suggest_categorical("e", ["foo", "bar", "baz"])
+        trial.suggest_int("f", 1, 10, log=True)
+
+        return 1.0
```
HideakiImamura, line 859: It is more informative to make the objective dependent on the suggested params.
```suggestion
    def objective(trial: BaseTrial) -> float:

        a = trial.suggest_uniform("a", 0, 10)
        b = trial.suggest_loguniform("b", 0.1, 10)
        c = trial.suggest_discrete_uniform("c", 0, 10, 1)
        d = trial.suggest_int("d", 0, 10)
        e = trial.suggest_categorical("e", [0, 1, 2])
        f = trial.suggest_int("f", 1, 10, log=True)

        return a + b + c + d + e + f
```

tests/test_trial.py
```diff
@@ -764,9 +765,325 @@ def test_study_id():
...
+        intermediate_values={},
+    )
+
+    assert trial.suggest_float("x", 0.0, 1.0) == 0.2
+
+    with pytest.raises(ValueError):
+        trial.suggest_float("x", 0.0, 1.0, step=10, log=True)
+
+    with pytest.raises(ValueError):
+        trial.suggest_uniform("y", 0.0, 1.0)
```
HideakiImamura, line 892:
```suggestion
        trial.suggest_float("y", 0.0, 1.0)
```

tests/test_trial.py
nzw0301: Definitely! But the objective returns `1`, so I update `objective` such that it returns a value depending on sample values.

tests/test_trial.py
nzw0301: Definitely! Thanks!

tests/test_trial.py
```diff
@@ -764,9 +765,328 @@ def test_study_id():
...
+    def objective(trial: BaseTrial) -> float:
+
+        a = trial.suggest_uniform("a", 0, 10)
+        b = trial.suggest_loguniform("b", 0.1, 10)
+        c = trial.suggest_discrete_uniform("c", 0, 10, 1)
+        d = trial.suggest_int("d", 0, 10)
+        e = trial.suggest_categorical("e", [0, 1, 2])
+        f = trial.suggest_int("f", 1, 10, log=True)
+
+        return a + b + c + d + e + f
```
nzw0301, line 859: mypy`s error shows

```python
tests/test_trial.py:859: error: Unsupported operand types for + ("float" and "None")
tests/test_trial.py:859: error: Unsupported operand types for + ("float" and "str")
tests/test_trial.py:859: note: Right operand is of type "Union[None, bool, int, float, str]"
Found 2 errors in 1 file (checked 183 source files)
```

nzw0301: I cannot understand why this error happens...

HideakiImamura: This is because the type of returned value for `suggest_categorical` is `CategoricalChoiceType = Union[None, bool, int, float, str]`. How about asserting the `e` is `float` in Line 858 as follows.

```python
assert isinstance(e, float)
```

nzw0301: Sorry for my poor understanding of `mypy`. (I'm still confused this technique as your suggestion, would you mind letting me know some reference about it?)


The returned value is not `float`, so `assert isinstance(e, int)` works. Anyway, mypy's error does not appear.
@HideakiImamura  Many thanks for your quick response!

nzw0301: Okay, probably [this part](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-attribute-exists-in-each-union-item-union-attr)...

> You can often work around these errors by using assert isinstance(obj, ClassName) or assert obj is not None to tell mypy that you know that the type is more specific than what mypy thinks.

HideakiImamura: Thanks for the reply and sorry for my late response.

> Okay, probably this part...

Yes, you are right! We have to insert some assertions like above to pass `mypy` sometimes. :)

optuna/trial/_frozen.py
nzw0301: I created the PR :)
https://github.com/optuna/optuna/pull/1528

optuna/trial/_frozen.py
```diff
@@ -2,16 +2,28 @@
...
+from optuna.trial._base import BaseTrial
 from optuna.trial._state import TrialState
 
 _logger = logging.get_logger(__name__)
 
+CategoricalChoiceType = Union[None, bool, int, float, str]
 
-class FrozenTrial(object):
+
+class FrozenTrial(BaseTrial):
```
hvy, line 26: Should we update the `FrozenTrial` in the `structs` submodule as well? It's under deprecation now but not removed until v3.0. I'm thinking that it'd be easier to maintain the code if they're identical. Maybe we can DRY the implementation with something along the following?

```python
# optuna/structs.py
FrozenTrial = deprecated(
    "1.4.0",
    text=(
        "This class was moved to :mod:`~optuna.trial`. Please use "
        ":class:`~optuna.trial.FrozenTrial` instead."
    ),
)(optuna.trial.FrozenTrial)
```

optuna/trial/_frozen.py
```diff
@@ -37,31 +49,31 @@ class FrozenTrial(object):
...
-        self.params = params
-        self.user_attrs = user_attrs
-        self.system_attrs = system_attrs
+        self._params = params
+        self._user_attrs = user_attrs
+        self._system_attrs = system_attrs
         self.intermediate_values = intermediate_values
         self._distributions = distributions
         self._trial_id = trial_id
+        self._suggested_params = {}  # type: Dict[str, Any]
```
hvy, line 76: The introduction of this attribute is a breaking change for `__eq__`. In this PR, this attribute is never really read from and can actually be removed. (Note that it is used in `FixedTrial` as the return value of the `params` attribute, which differs from here)

tests/test_trial.py
```diff
@@ -764,9 +765,329 @@ def test_study_id():
...
+    )
+
+    assert trial == eval(repr(trial))
+
+
+@parametrize_storage
+def test_frozen_trial_sampling(storage_init_func):
+    # type: (Callable[[], storages.BaseStorage]) -> None
+
+    def objective(trial: BaseTrial) -> float:
```
hvy, line 850: Oh, the mypy type checking passed with `BaseTrial`? I thought we had to change it given [`Study.optimize`](https://github.com/optuna/optuna/blob/master/optuna/study.py#L213).

optuna/trial/_frozen.py
hvy: Some attributes by the way cannot be kept consistent with both `Trial` (lazily populated) and `FrozenTrial` (set at construction). It might be okay to favor the latter as internal implementation relies on it. Those include `params`, `{user,system}_attrs`, `distributions`, etc. What do you think? Similar to how pruning works, it might be the case that we just have to document it.

optuna/trial/_frozen.py
nzw0301: I certainly agree with you. Your suggested code looks great!

hvy: Oh, that was a quick fix, thanks. Just FYI, I didn't verify it locally so not 100% sure that it will work as expected (e.g. that it doesn't alter the original docstrings).

tests/test_trial.py
nzw0301: Yes, it is...

I'm not sure, but how about this?
```python
    def objective(trial):
        # type: (Union[Trial, FrozenTrial]) -> float
```

optuna/trial/_frozen.py
nzw0301: At least `test_structs.py` works fine, but another test (`test_study.test_optimize_parallel_storage_warning`) starts to fail... I will investigate it.

optuna/trial/_frozen.py
nzw0301: Just documenting it sounds simple at once. I think that it is great to clean the logic by making code consistent as another PR.

hvy: Do you mean you'll include the additional documentation in this PR? I'm fine with either that or creating another one afterwards. 🙂  As for making the logic "consistent", I don't think it's possible for the reason described above. `params` for instance is populated as the user calls the `suggest_*` methods, while it's populated for a `FrozenTrial` when it's first constructed. If this `FrozenTrial` is passed to the objective function and the user assumes that it should act as a regular `Trial`, he/she might encounter unexpected behavior. This however, hopefully, is not a major concern in practice. Hope that clarifies!

tests/test_trial.py
hvy: I think we should relax [`ObjectiveFuncType`](https://github.com/optuna/optuna/blob/master/optuna/study.py#L42) to accept `BaseTrial`s and not only `Trial`s since the public APIs for both types are identical (and are likely not going to diverge in a non-compatible way the near future).

Was mypy not complaining on the current (in this PR) test? If so, we could do that in a follow-up PR. Open for feedback.

optuna/trial/_frozen.py
hvy: It might be altering the original, having side-effects on other tests having `optuna.trial.FrozenTrial.__init__` raising warnings as well. Maybe this will do?

```python
@deprecated(
    "1.4.0",
    text=(
        "This class was moved to :mod:`~optuna.trial`. Please use "
        ":class:`~optuna.trial.FrozenTrial` instead."
    ),
)
class FrozenTrial(optuna.trial.FrozenTrial):
    pass
```

HideakiImamura:
@nzw0301 Thanks for your great effort! I will re-review this PR after finishing the above discussions.

tests/test_trial.py
nzw0301: Thank you for your clarification! Yes, mypy was passed in this PR: 
https://app.circleci.com/pipelines/github/optuna/optuna/1711/workflows/96a2a01e-6ee5-4812-a31d-5ef182218b57/jobs/49893/parallel-runs/0/steps/0-106
I agree with you.

optuna/trial/_frozen.py
nzw0301: Your suggestion works! I really thank you for your help!

optuna/trial/_frozen.py
```diff
@@ -121,8 +128,82 @@ def __repr__(self):
...
+
+        choices = tuple(choices)
+        return self._suggest(name, CategoricalDistribution(choices=choices))
+
+    def report(self, value: float, step: int) -> None:
+        pass
+
+    def should_prune(self) -> bool:
+
+        return False
```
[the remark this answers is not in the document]
nzw0301: Definitely! I added documents to `report`, `should_prune`, and `last_step`.

optuna/trial/_frozen.py
nzw0301: Yes, I do!  I added a note that explains `FrozenTrial`s attributes could be overwritten by re-evaluation. https://github.com/optuna/optuna/pull/1503/commits/dbbc1c343e1842eb790648d916e9f45fce249f45

hvy:
@nzw0301 just checking but is this PR ready for another review?

nzw0301:
@hvy @HideakiImamura  Yes! Could you review this PR again?

optuna/trial/_frozen.py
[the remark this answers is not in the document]
hvy: Looks nice, thanks!

tests/test_trial.py
hvy: Created an issue so that we don't forget, https://github.com/optuna/optuna/issues/1577.

nzw0301:
It's my pleasure! Thank you for reviewing the PR and giving helpful feedback!

PR #1504, opened 2020-07-10
Align `LightGBMTuner` verbosity level to the original LightGBM.

toshihikoyanase:
## Motivation
The current `LightGBMTuner` has only two levels of verbosity (i.e., `0` or `1`) while the original LightGBM has four levels (i.e., `<0`, `=0`, `=1`, `>1`. See this [link](https://lightgbm.readthedocs.io/en/latest/Parameters.html#verbosity)).

## Description of the changes
This PR aligns the verbosity level to LightGBM. I follow [this implementation](https://github.com/microsoft/LightGBM/blob/bca2da97284403b1b724a0e2966a76ecbb733fc9/src/io/config.cpp#L234-L242) to determine the logging level.

Changes
- Instead of disabling Optuna's handler, this PR changes the logging level of Optuna.
- The current implementation also changes the value of `verbose` option of LightGBM, this PR keeps the user-given value.
- The current implementation allows `None` in addition to `int` for `verbosity`, this PR only accept `int`.
- The progress bar is enabled if `verbosity >= 1`.

### DIFF
--- a/optuna/integration/_lightgbm_tuner/optimize.py
+++ b/optuna/integration/_lightgbm_tuner/optimize.py
@@ -13,6 +13,7 @@ from typing import List
 from typing import Optional
 from typing import Tuple
 from typing import Union
+import warnings
 
 import numpy as np
 import tqdm
@@ -337,7 +338,8 @@ class _LightGBMBaseTuner(_BaseTuner):
         sample_size: Optional[int] = None,
         study: Optional[optuna.study.Study] = None,
         optuna_callbacks: Optional[List[Callable[[Study, FrozenTrial], None]]] = None,
-        verbosity: int = 1,
+        verbosity: Optional[int] = None,
+        show_progress_bar: bool = True,
     ) -> None:
 
         _imports.check()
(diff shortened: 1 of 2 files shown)


toshihikoyanase:
I talked with @hvy and changed this PR as follows:
- Deprecate the `verbosity` argument. Users will use `optuna.logging.set_verbosity()` to control Optuna's logging level.
- The `verbosity` argument only controls Optuna's logging level.
- Add the `show_progress_bar` argument to control the progress bars.

optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -653,7 +672,26 @@ class LightGBMTuner(_LightGBMBaseTuner):
...
+
+            .. warning::
+                Deprecated in v2.0.0. ``verbosity`` argument will be removed in the future.
+                The removal of this feature is currently scheduled for v4.0.0,
+                but this schedule is subject to change.
+
+                Please use :func:`~optuna.logging.set_verbosity` instead.
+
+        show_progress_bar:
+            Flag to show progress bars or not. To disable progress bar, set this ``False``.
```
hvy, line 687: We should probably prefer `:obj:`False`` but it's a minor nit.

HideakiImamura:
I will merge this PR after the author addressing https://github.com/optuna/optuna/pull/1504#discussion_r453384307.

PR #1506, opened 2020-07-10
Fix `_CachedStorage` and `RDBStorage` distribution compatibility check race condition.

hvy:
_Depends on https://github.com/optuna/optuna/pull/1490, using study table query with `FOR UPDATE`._
_Should be merged after https://github.com/optuna/optuna/pull/1498, addressing another parameter issue where rows are not inserted._

## Motivation

Trial distribution compatibility check is prone to race condition. This is because table query --> check does not involve any locks across trials. For instance, two workers could enter the distribution check simultaneously and both pass the check since they both see an empty table. The first worker should acquire a lock, check and insert, to ensure that the compatibility check really takes place for the second trial.

## Description of the changes

Locks across trials by locking a row in the study table.

### DIFF
--- a/optuna/storages/_cached_storage.py
+++ b/optuna/storages/_cached_storage.py
@@ -234,7 +234,7 @@ class _CachedStorage(BaseStorage):
                     # ones. By INSERT, it is assumed that no previous entry has been persisted
                     # already.
                     self._backend._check_and_set_param_distribution(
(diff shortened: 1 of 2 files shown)


hvy:
Resolved conflicts and rebased on `master` since #1490 and #1498 were merged. The diff is clearer now. PTAL.

Issue #1499, opened 2020-07-09
Deadlock can occur when using MySQL backend.


ytsmiling:
## Expected behavior

Optuna handles MySQL errors.

## Environment

- Optuna version: 1.5.0
- Python version: 3.7
- OS: mac OS Catalina
- (Optional) Other libraries and their versions: MySQL + pymysql

## Error messages, stack traces, or logs

```
sqlalchemy.exc.InvalidRequestError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely)
(pymysql.err.OperationalError) (1213, 'Deadlock found when trying to get lock; try restarting transaction')
[SQL: INSERT INTO trial_user_attributes (trial_id, `key`, value_json) VALUES (%(trial_id)s, %(key)s, %(value_json)s)]
[parameters: {'trial_id': 2998, 'key': 'nodename', 'value_json': '"anoanymized"'}]
(Background on this error at: http://sqlalche.me/e/13/e3q8) (Background on this error at: http://sqlalche.me/e/13/7s2a)
```

## Steps to reproduce

1. Parallel optimization with MySQL backend.

## Reproducible examples (optional)

<details><summary>Code</summary>
<p>

```python
import os
import optuna
from multiprocessing import Pool


def f(x):
    return (x - 3) ** 2


# Objective function
def optuna_objective(trial):
    x = trial.suggest_uniform("x", -10, 10)
    y = f(x)

    nodename = os.uname().nodename
    trial.set_user_attr("nodename", nodename)

    print("x=%8.3f, y=%8.3f" % (x, y))

    return y


# Entry point
def main(args):
    rank, = args
    # Create a study
    ECHO=False
    storage = optuna.storages.RDBStorage(os.environ["OPTUNA_SQL"], engine_kwargs=dict(echo=ECHO))
    study = optuna.create_study(
        study_name="test_optuna_collision_01",
        storage=storage,
        load_if_exists=True,
        direction="minimize"
    )

    # Run optimization
    study.optimize(optuna_objective, n_trials=20)

    import time
    time.sleep(10)

    # Visualize the result
    if rank == 0:
        print("  #  %8s  %8s  %8s  %28s" % ("x", "y", "NODENAME", "STARTED"))
    for trial in study.trials:
        try:
[...]

PR #1509, opened 2020-07-13
Retry after insertion into RDB-storage.

ytsmiling:
This PR is a follow-up of #1498 and #1499.

## Motivation
See #1498 and #1499.

## Description of the changes
- Add retry loop so that we can re-issue operations on dead-lock.
- Split data insertion into different sessions.

Additionally, this PR fixes the following two bugs.
- Intermediate values are not correctly overwritten.
- Params are not correctly overwritten. See #1327 for the change of storage specification.

## Additional notes
We may replace `add_all` statements by bulk-insert to improve performance within this or in another PR.

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -1,3 +1,4 @@
+from contextlib import contextmanager
 import copy
 from datetime import datetime
 import json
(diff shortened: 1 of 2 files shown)


ytsmiling:
This PR resolves #1499. Other operations can cause the same problem with #1499, but this PR removes the most critical one, and I think we can defer the fixes of them to other PRs.

hvy:
This issue also seems to be mitigated by avoiding the conditional table locks of `TrialSystemAttributeModel`, `TrialValueMode`, etc., doing an eager JOIN in the `TrialModel` query in the beginning. Do you think it's less preferable?

Nice introduction of the context sensitive session by the way. I was also thinking of that.

ytsmiling:
#1514 seems a better fix and I'd like to close this pr.

PR #1514, opened 2020-07-14
Fix frequent deadlock caused by conditional locks.

hvy:
## Motivation

Alternative to https://github.com/optuna/optuna/pull/1509. See also https://github.com/optuna/optuna/pull/1509#issuecomment-657913671.

This fix does not contain any retries but as far as I could verify locally, deadlocks no longer occur. 

## Description of the changes

Only uses a single lock, over all trial-related tables.


hvy:
Thanks for you quick review. You're right that it's a somewhat naive implementation in terms of performance. I can add a TODO comment about that.

optuna/storages/_rdb/storage.py
```diff
@@ -607,9 +607,16 @@ def _update_trial(
...
+
+        # Lock all related tables with eager joins.
+        # TODO(hvy): Naive in terms of performance. Avoid joining with unncessary tables.
         trial_model = (
             session.query(models.TrialModel)
             .filter(models.TrialModel.trial_id == trial_id)
+            .options(orm.selectinload(models.TrialModel.params))
+            .options(orm.selectinload(models.TrialModel.values))
+            .options(orm.selectinload(models.TrialModel.user_attributes))
+            .options(orm.selectinload(models.TrialModel.system_attributes))
```
toshihikoyanase, line 619: I dumped the SQL queries related to this line. The queries are the same whether we add `selectinload` or not as follows: 

```sql
SELECT trials.trial_id AS trials_trial_id, trials.number AS trials_number, trials.study_id AS trials_study_id, trial
s.state AS trials_state, trials.value AS trials_value, trials.datetime_start AS trials_datetime_start, trials.datetime_complete AS trials_datetime_complete 
FROM trials 
WHERE trials.trial_id = %(trial_id_1)s FOR UPDATE
```

As can be seen in the example in the [reference](https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html#selectin-eager-loading), the change is can be seen in the following SELECT queries.

```sql
# without selectinload
SELECT trial_user_attributes.trial_user_attribute_id AS trial_user_attributes_trial_user_attribute_id, trial_user_attributes.trial_id AS trial_user_attributes_trial_id, trial_user_attributes.`key` AS trial_user_attributes_key, trial_user_attributes.value_json AS trial_user_attributes_value_json 
FROM trial_user_attributes 
WHERE trial_user_attributes.trial_id = %(trial_id_1)s

# with selectionload
SELECT trial_user_attributes.trial_id AS trial_user_attributes_trial_id, trial_user_attributes.trial_user_attribute_id AS trial_user_attributes_trial_user_attribute_id, trial_user_attributes.`key` AS trial_user_attributes_key, trial_user_attributes.value_json AS trial_user_attributes_value_json 
FROM trial_user_attributes 
WHERE trial_user_attributes.trial_id IN (%(primary_keys_1)s) ORDER BY trial_user_attributes.trial_id```
```

So, I guess the `selectinload` option does not affect the record lock, but I may misunderstand the lock mechanism. What do you think of it?

toshihikoyanase: I used the following code to dump SQL queries: https://gist.github.com/toshihikoyanase/f839e9ab9b2cbba23366b635f12b79da

hvy: Thanks, you're right. We probably need to do a proper "JOIN" (e.g. `orm.joinedload`) here.

hvy: Updated. I verified that this properly locks backref tables. Sorry but could you check it again?

Note that we cannot do an `INNER JOIN` ([`innerjoin=True`](https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html#sqlalchemy.orm.joinedload)) and must do an `LEFT OUTER JOIN` since backref tables don't necessarily have rows corresponding to the trial ID when we enter this section. Performance-wise, this change doesn't seem to have any noticeable impact, but I've only tested on a very trivial case.

hvy:
That the number of queries are reduced is actually not that important here. "Eager loading", or importantly `JOIN`s, will make sure that locks propagate to joined tables (not just `TrialModel` but also `TrialParamModel`, etc.). This should prevent race conditions. 

Details might vary between dialects but these references might shed some light.

- https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE
- https://dev.mysql.com/doc/refman/8.0/en/innodb-locks-set.html

optuna/storages/_rdb/storage.py
toshihikoyanase: The change seems good to me, but I sometimes encountered the deadlock as follows:

```console
sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (1213, 'Deadlock found when trying to get lock; try restarting transaction')
[SQL: INSERT INTO trial_user_attributes (trial_id, `key`, value_json) VALUES (%(trial_id)s, %(key)s, %(value_json)s)]
[parameters: {'trial_id': 23, 'key': 'x', 'value_json': '7.048336196015672'}]
(Background on this error at: http://sqlalche.me/e/13/e3q8)

optuna.exceptions.StorageInternalError: An exception is raised during the commit. This typically happens due to invalid data in the commit, e.g. exceeding max length. (The actual exception is as follows: OperationalError("(pymysql.err.OperationalError) (1213, 'Deadlock found when trying to get lock; try restarting transaction')"))
```

Do you think we need retry like #1509?

ytsmiling:
Let me clarify the cause of the original bugs and changes introduced by this PR.
I initially thought that the problem was caused by #1496. However, after this PR, discussions, and investigation of logs provided in https://github.com/optuna/optuna/pull/1514#discussion_r454106924, I'm rather skeptical about why the eager-loading resolves the problem. I think the biggest diff in this PR is that it removes `with_for_update` from several queries. This removes oddly nested `FOR UPDATE` queries and it's plausible that it mitigates the frequent dead-lock.
When you invest the log when we use `selectinload`, you can find that tables are locked individually.
After we switched to `joinedload`, we can see that the query locks multiple tables at the same time. Though I'm not 100% sure, but it's not that surprising that it can cause a dead-lock depending on DB implementations.

There'll be several methods to address deadlocks, including retry-clause, but I'm currently wondering a more fundamental question: why do we need locks for the first place. Currently, optuna's spec requires that RUNNING trials are updated from only a single worker, and I think we don't need any locks. After you remove `with_for_update` from the first clause, you'll won't get the frequent deadlock in https://github.com/optuna/optuna/pull/1514#discussion_r454128925.

hvy:
The locks were introduced here https://github.com/optuna/optuna/commit/9ba2cf6466aa02fe65fe7472f8f050f9b73f1b4a, but yes I agree that those shouldn't be needed given the storage specification that a trial shouldn't be updated from a different worker than its creator. I removed the additional locks in the latest commit.

optuna/storages/_rdb/storage.py
hvy: Thanks for the additional information, this time the query seems to align with the expectation. It's still not entirely clear to me why the deadlocks were occur however. As for this issue, I changed the logic to not lock at all as they actually shouldn't be needed.

toshihikoyanase: Thank you for your update. I read the comment https://github.com/optuna/optuna/pull/1514#issuecomment-658001961, and I think your latest change sounds good to me.

hvy:
A rather fruitful detour. Thank you all for the fast actions.

PR #1518, opened 2020-07-15
Overwrite parameters correctly.

ytsmiling:
## Motivation
This PR fixes the storage spec violation that `_CachedStorage` + `RDBStorage` does not overwrite parameters. Background of the spec: #1308.

## Additional context
This functionality is (currently) never used by users since `trial.suggest` immediately returns when already-suggested keys are passed to the method.


hvy:
Resolved conflicts.

PR #1521, opened 2020-07-15
Introduce actions/cache and bump up actions/setup-python

crcrpar:


.github/workflows/examples.yml
```diff
@@ -16,13 +16,24 @@ jobs:
...
+      uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install
+    - name: Setup cache
+      uses: actions/cache@v2
+      env:
+        cache-name: daily-example
+      with:
+        path: /opt/**/site-packages
```
crcrpar, line 27: I'm not quite sure that this path specification is cool or not at all.

Ref: https://github.com/actions/setup-python/#hosted-tool-cache

.github/workflows/tests-integration.yml
```diff
@@ -24,10 +24,20 @@ jobs:
...
+      uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Setup cache
+      uses: actions/cache@v2
+      env:
+        cache-name: test-integration
+      with:
+        path: /opt/**/site-packages
```
crcrpar, line 36: ditto.

.github/workflows/tests.yml
```diff
@@ -33,10 +33,19 @@ jobs:
...
+      uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Setup Cache
+      uses: actions/cache@v2
+      env:
+        cache-name: test
+      with:
+        path: /opt/**/site-packages
```
crcrpar, line 45: ditto

hvy:
Interesting PR. Do you have any benchmarks? I once tried to cache the downloaded files with pip but that didn't improve a lot. https://github.com/hvy/optuna/pull/7/commits/498a4734b84b4e61bbb93b3530dd81dfa5f8ce4e

crcrpar:
IIRC, when the cache is there, the python preparation step took around 20 seconds.

crcrpar:
I suppose your commit specifies a wrong path as the path to cache. With actions/setup-python, python cache path would not be that one. See https://github.com/actions/setup-python/#hosted-tool-cache.

hvy:
Oh I didn't know that. Though it seemed like it skipped the download as expected [here in the `install` step](https://github.com/hvy/optuna/runs/753334487?check_suite_focus=true) when a cache hit occurred.

hvy:
> the python preparation step took around 20 seconds.

Do you mean the `Install` step? It'd be nice to have a before/after comparison (preferably for each of the modified files).

.github/workflows/examples.yml
hvy: Maybe safer to get it from the runtime? 

```bash
python -c "import site; print(site.getsitepackages())"
```

.github/workflows/examples.yml
```diff
@@ -16,13 +16,24 @@ jobs:
...
+        key: ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}
+        restore-keys: |
+          ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-
```
hvy, line 30: You also need to match with the hash from `setup.py` here.

.github/workflows/examples.yml
```diff
@@ -16,13 +16,24 @@ jobs:
...
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install
+    - name: Setup cache
+      uses: actions/cache@v2
+      env:
+        cache-name: daily-example
+      with:
+        path: /opt/**/site-packages
+        key: ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}
```
hvy, line 28: Could be over-engineering but we'll probably sooner or later want to manually overwrite/delete caches. The following is one way to prepare for it.

```suggestion
        key: ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}-v1
```

.github/workflows/examples.yml
```diff
@@ -16,13 +16,24 @@ jobs:
...
-    - name: Install
+    - name: Setup cache
+      uses: actions/cache@v2
+      env:
+        cache-name: daily-example
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}-v1
+        restore-keys: |
+          ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ env.cache-name }}
```
hvy, line 30: Did you mean the following?
```suggestion
          ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}
```

c-bata:
> To detect the changes of dependencies, I rely on setup.py ( hashFiles('**/setup.py) ), but I'm not confident enought at this moment.

FYI: I'm not confident but I mind this cache will be broken due to the Python's patch version up on host image like [this Japanese tweet](https://twitter.com/atsuoishimoto/status/1288270742337904641).

crcrpar:
At this point this seems okay though, I'm closing this pull request for two reasons.
(1) `cache` could lead to some errors mentioned by @c-bata (I really appreciate your helpful comments). (2) `--no-cache-dir` option might resolve recent pip installation failures in CI.

Issue #1351, opened 2020-06-10
Error when using custom metrics in optuna.integration.lightgbm


worldblue0214:
## Expected behavior
I got an error when using custom metrics in optuna.integration.lightgbm.

## Environment

- Optuna version: 1.5.0
- Python version: 3.7.7
- OS: MacOS Catalina 10.15.5
- Anaconda
- (Optional) Other libraries and their versions:
[conda_list.txt](https://github.com/optuna/optuna/files/4757062/conda_list.txt)

## Error messages, stack traces, or logs
```
 Early stopping, best iteration is:
 [113]	training's custom_metrics: 0.73954	valid_1's custom_metrics: 2.25621

KeyError                                  Traceback (most recent call last)
<ipython-input-45-c87e652c308e> in <module>
      1 best_params = {}
      2 model = lgb_tuner.train(params, train_set, num_boost_round = 2500, early_stopping_rounds = 50,
----> 3                   valid_sets = [train_set, val_set], verbose_eval = 100, feval= custom_metrics)

~/opt/anaconda3/envs/dev-python37/lib/python3.7/site-packages/optuna/_experimental.py in new_func(*args, **kwargs)
     62                 )
     63 
---> 64                 return func(*args, **kwargs)  # type: ignore
     65 
     66             return new_func

~/opt/anaconda3/envs/dev-python37/lib/python3.7/site-packages/optuna/integration/lightgbm_tuner/__init__.py in train(*args, **kwargs)
     44 
     45     auto_booster = LightGBMTuner(*args, **kwargs)
---> 46     auto_booster.run()
     47     return auto_booster.get_best_booster()

~/opt/anaconda3/envs/dev-python37/lib/python3.7/site-packages/optuna/integration/lightgbm_tuner/optimize.py in run(self)
    483         self.sample_train_set()
    484 
--> 485         self.tune_feature_fraction()
    486         self.tune_num_leaves()
    487         self.tune_bagging()

~/opt/anaconda3/envs/dev-python37/lib/python3.7/site-packages/optuna/integration/lightgbm_tuner/optimize.py in tune_feature_fraction(self, n_trials)
    511             warnings.simplefilter("ignore", category=optuna.exceptions.ExperimentalWarning)
    512             sampler = optuna.samplers.GridSampler({param_name: param_values})
--> 513         self.tune_params([param_name], len(param_values), sampler, "feature_fraction")
    514 
    515     def tune_num_leaves(self, n_trials: int = 20) -> None:

[...]

PR #1522, opened 2020-07-15
Support custom metrics with LightGBM tuner.

hvy:
## Motivation

Fixes https://github.com/optuna/optuna/issues/1351.

## Description of the changes

See https://github.com/optuna/optuna/issues/1351#issuecomment-643873194.

### DIFF
--- a/optuna/integration/_lightgbm_tuner/optimize.py
+++ b/optuna/integration/_lightgbm_tuner/optimize.py
@@ -109,6 +109,12 @@ class _BaseTuner(object):
         else:
             raise NotImplementedError
 
+        if self.lgbm_params.get("metric") == "None":
+            if len(booster.best_score[valid_name].keys()) > 0:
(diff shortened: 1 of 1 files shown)


toshihikoyanase:
Thank you for your PR. I think we can also take care of the direction of studies.
Currently, `LightGBMTuner` checks the metric name to determine the direction with simple logic. If the metric name starts from one of `auc, ndcg, map`, the direction will be `maximize`, otherwise `minimize`. So, the current implementation always sets the direction to `minimize`.

We have several approaches to fix this issue. One of the simplest methods is to ask users to set the direction via `study` argument of `LightGBMTuner.run`. But it is controversial, and I think we need further discussion.

philip-khor:
Not sure if this is helpful: the third argument of custom metrics specified by feval should indicate the direction to minimize.

PR #1528, opened 2020-07-16
Match the order of definition in `trial`

nzw0301:
## Motivation

The order of definition of functions is slightly different between `_fixed.py` and `_trial.py` as pointed by https://github.com/optuna/optuna/pull/1503#discussion_r454770915. The order consistency is supposed to be important for contributors.

## Description of the changes

Put the definition of `_suggest` in `_fixed.py` after the definitions of public function.

### DIFF
--- a/optuna/trial/_fixed.py
+++ b/optuna/trial/_fixed.py
@@ -126,6 +126,25 @@ class FixedTrial(BaseTrial):
 
         return self._suggest(name, CategoricalDistribution(choices=choices))
 
+    def report(self, value, step):
+        # type: (float, int) -> None
+
+        pass
+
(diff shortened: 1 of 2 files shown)


hvy:
Good catch on a detail. The changes look good in terms of consistency. 

On a side note, I'm not convinced that the order in `_trial.py` is any better. As a rule of thumb, one may for instance place private methods after all the public interfaces, c.f. https://stackoverflow.com/questions/10289461/what-is-a-good-way-to-order-methods-in-a-python-class

nzw0301:
@HideakiImamura Thank you for pointing out. I fixed it.

PR #1536, opened 2020-07-21
Create study before multiple workers are launched in Kubernetes MLflow example.

toshihikoyanase:
## Motivation
When Optuna creates the first study in an RDB server, it internally creates tables to store trials and studies. The current Kubernetes-MLFflow example runs `optuna.create_study` from multiple workers, and it may cause errors about duplicate table creation.

## Description of the changes
This PR creates a study prior to the worker launch and uses `optuna.load_study`  instead of `optuna.create_study` in the workers.


toshihikoyanase:
@hvy Thank you for your review.

> we shouldn't use the same job since the parallelism differs.

I agree. I couldn't find a simple solution to reuse the `worker` jobs.

> --skip-if-exists

I just added it in case that some users may reuse the PostgreSQL server and I think it is optional.

toshihikoyanase:
@HideakiImamura Thank you for your review. The workers seem to be waiting for the study creation even if the `study-creator-l4kxn` job has completed. I think that something is wrong with the init containers of the workers.
Could you show me the log message of init containers by executing the following commands?

```console
$ kubectl logs worker-45trk -c wait-for-study
$ kubectl logs worker-ljdtx -c wait-for-study
```

c.f. https://kubernetes.io/docs/concepts/workloads/pods/init-containers/

HideakiImamura:
@toshihikoyanase Thanks for the reply. Executing `$ kubectl logs worker-<pod id> -c wait-for-study` result in the following output.
```
...
+ sh check_study.sh
sh: 0: Can't open check_study.sh
+ [ -eq 0 ]
/bin/sh: 1: [: -eq: unexpected operator
+ echo waiting for study
waiting for study
+ sleep 2
+ sh check_study.sh
sh: 0: Can't open check_study.sh
+ [ -eq 0 ]
/bin/sh: 1: [: -eq: unexpected operator
+ echo waiting for study
waiting for study
+ sleep 2
...
```
Each worker fails to execute `check_study.sh`

toshihikoyanase:
Thank you for sharing your log messages.

> sh: 0: Can't open check_study.sh

Hmm, I guess your docker image is not updated, and the `check_study.sh` script is not included in the image. Could you check if your docker image contains it?

PR #1542, opened 2020-07-21
Reduce search space for AllenNLP example.

himkt:
## Motivation
Current `allennlp_simple.py` may be computationally extensive and it sometimes causes a timeout.
https://github.com/optuna/optuna/actions/runs/175854473

## Description of the changes
In this PR, I reduce the hyperparameter search space for the model.
Additionally, I extracted the dataset reader from the script and created the module because I'll also use it in `allennlp_jsonnet.py`.

I tested this new example on GitHub Actions to confirm that it reduced the execution time for each trial.
https://github.com/himkt/optuna/runs/894041267

### DIFF
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -40,7 +40,7 @@ from subsample_dataset_reader import SubsampleDatasetReader  # NOQA
 
 DEVICE = -1  # If you want to use GPU, use DEVICE = 0.
 N_TRAIN_DATA_SIZE = 2000
-N_VALID_DATA_SIZE = 1000
+N_VALIDATION_DATA_SIZE = 1000
 MODEL_DIR = os.path.join(os.getcwd(), "result")
 TARGET_METRIC = "accuracy"
 
@@ -53,7 +53,7 @@ def prepare_data():
         token_indexers={"tokens": indexer},
         tokenizer=tokenizer,
         train_data_size=N_TRAIN_DATA_SIZE,
-        valid_data_size=N_VALID_DATA_SIZE,
+        validation_data_size=N_VALIDATION_DATA_SIZE,
     )
     train_dataset = reader.read(
         "https://s3-us-west-2.amazonaws.com/allennlp/datasets/imdb/train.jsonl"
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -5,17 +5,6 @@ This script is based on the example of allentune (https://github.com/allenai/all
 In this example, we optimize the validation accuracy of sentiment classification using AllenNLP.
 Since it is too time-consuming to use the entire dataset, we here use a small subset of it.
 
-We have the following two ways to execute this example:
-
-(1) Execute this code directly.
-    $ python allennlp_simple.py
-
-
-(2) Execute through CLI.
-    $ STUDY_NAME=`optuna create-study --direction maximize --storage sqlite:///example.db`
-    $ optuna study optimize allennlp_simple.py objective --n-trials=100 --study-name $STUDY_NAME \
-      --storage sqlite:///example.db
-
 """
 
 import os
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -1,6 +1,6 @@
 """
 Optuna example that optimizes a classifier configuration for IMDB movie review dataset.
-This script is based on the example of allentune (https://github.com/allenai/allentune).
+This script is based on the example of AllenTune (https://github.com/allenai/allentune).
 
 In this example, we optimize the validation accuracy of sentiment classification using AllenNLP.
 Since it is too time-consuming to use the entire dataset, we here use a small subset of it.
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -22,6 +22,7 @@ import os
 import pkg_resources
 import random
 import shutil
+import sys
 
 import allennlp
 import allennlp.data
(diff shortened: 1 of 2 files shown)


examples/allennlp/allennlp_simple.py
```diff
@@ -105,11 +100,11 @@ def objective(trial):
     if DEVICE > -1:
         model.to(torch.device("cuda:{}".format(DEVICE)))
 
-    lr = trial.suggest_float("lr", 1e-5, 1e-1, log=True)
+    lr = trial.suggest_float("lr", 1e-2, 1e-1, log=True)
```
himkt, line 103: I increased the lower bound of learning rate, as I found that model training didn't succeed at all.

examples/allennlp/allennlp_simple.py
```diff
@@ -123,7 +118,7 @@ def objective(trial):
         validation_data_loader=validation_data_loader,
         validation_metric="+" + TARGET_METRIC,
         patience=None,  # `patience=None` since it could conflict with AllenNLPPruningCallback
-        num_epochs=50,
+        num_epochs=30,
```
himkt, line 121: I decreased the number of epochs for reducing the execution time for each trial.

examples/allennlp/allennlp_simple.py
```diff
@@ -33,38 +34,33 @@
 import optuna
 from optuna.integration import AllenNLPPruningCallback
 
+sys.path.append(os.path.dirname(os.path.abspath(__file__)))
```
himkt, line 37: Trick for importing `subsample_dataset_reader` when this example is launched on the top of this repo.

toshihikoyanase: Hmm, this trick does not work when users use CLI. I got the following error:
```console
name '__file__' is not defined
  File "/Users/yanase/.pyenv/versions/optuna-dev/lib/python3.7/site-packages/cliff/app.py", line 400, in run_subcommand
    result = cmd.run(parsed_args)
  File "/Users/yanase/.pyenv/versions/optuna-dev/lib/python3.7/site-packages/cliff/command.py", line 187, in run
    return_code = self.take_action(parsed_args) or 0
  File "/Users/yanase/pfn/code/optuna/optuna/cli.py", line 325, in take_action
    loader.exec_module(target_module)
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "allennlp_simple.py", line 37, in <module>
    sys.path.append(os.path.dirname(os.path.abspath(__file__)))
NameError: name '__file__' is not defined
```

examples/allennlp/subsample_dataset_reader.py
```diff
@@ -0,0 +1,26 @@
+import random
+
+import allennlp
+from allennlp.data.dataset_readers.dataset_reader import DatasetReader
+from sklearn.model_selection import train_test_split
+
+
+@DatasetReader.register("subsample")
+class SubsampledDatasetReader(allennlp.data.dataset_readers.TextClassificationJsonReader):
```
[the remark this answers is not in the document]
himkt: Thank you for the review.
Initially, I create `SubSampledDataset`, which describes a reader for a subsampled dataset.
But I renamed it to `SubsampleDataset` since I could describe this class as a sub-sample dataset.
(And I thought it would be better to use only present tense words in the filenames if it was possible.)

I'm not confident for this convention, so I'd like to get your feedback.
Do you think `SubsampledDatasetReader` is better?

examples/allennlp/allennlp_simple.py
himkt: Oh, you're right. Based on #1394, I'm wondering if I could remove a description of `optuna CLI` here.
What do you think?

toshihikoyanase: That makes sense. I think we can remove the instruction about `optuna study optimize`.

examples/allennlp/subsample_dataset_reader.py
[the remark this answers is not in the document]
toshihikoyanase: I think either one is OK because I found both use cases.

- `SubsampledDatasetReader` can be seen in the following issue:
https://github.com/allenai/allennlp/issues/3099#issuecomment-515688316
(BTW, [`itertools.islice`](https://docs.python.org/ja/3/library/itertools.html#itertools.islice) was employed in the code example in the issue. It can be useful if the dataset is too large to fit in memory.)
- `subsample` is also used in the variable name [`batch_subsample_map`](https://github.com/allenai/allennlp/issues/1408#issuecomment-401143900).

I prefer `SubsampleDatasetReader` for its simplicity, but it is not a strong opinion.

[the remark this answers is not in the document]
himkt: OK, so let me leave the dataset reader name as is.

> `islice` ... It can be useful if the dataset is too large to fit in memory.
Thank you for pointing out, you're right!
I gave it try to use `islice` in 4437f48 and found the dataset is quite balanced.

```
train: Counter({'0': 1003, '1': 997})
validation: Counter({'1': 489, '0': 511})
```

I also ran HPO on GPU to confirm training could succeed.

```
python allennlp_simple.py
....
[I 2020-07-22 13:38:58,896] Trial 21 finished with value: 0.704 and parameters: {'dropout': 0.18415084463564746, 'embedding_dim': 33, 'output_dim': 69, 'max_filter_size': 4, 'num_filters': 55, 'lr': 0.09093981266227305}. Best is trial 12 with value: 0.725.
Number of finished trials:  22
Best trial:
  Value:  0.725
  Params:
    dropout: 0.31246648830863727
    embedding_dim: 22
    output_dim: 98
    max_filter_size: 4
    num_filters: 38
    lr: 0.09838055114180111
```

So I determined to remove `train_test_split` and simply use `islice`.

examples/allennlp/allennlp_simple.py
himkt: ac4a13c removed the document about `optuna study optimize`.
I also remove it about `python allennlp_simple.py` as it is obvious to run a script,
after removing the way to use optuna cli.

examples/allennlp/allennlp_simple.py
himkt: 74c8a23 1e-2 -> 1e-3

examples/allennlp/subsample_dataset_reader.py
```diff
@@ -0,0 +1,20 @@
...
+        self.train_data_size = train_data_size
+        self.validation_data_size = validation_data_size
+
+    def _read(self, datapath):
+        if datapath.endswith("train.jsonl"):
+            data_size = self.train_data_size
+        else:
+            data_size = self.validation_data_size
+
+        yield from itertools.islice(super()._read(datapath), data_size)
```
HideakiImamura, line 20: While the previous `SubsampledDatasetReader` shuffles data before yielding, the current `SubsampleDatasetReader` does not do so. Is it intended?

himkt: Yes, it's intended for taking advantage of a generator. 🙋‍♂️ 

I take a quick look at the dataset to find that it seems to be shuffled in advance.
(I put the screenshot of `train.jsonl` at the bottom of this comment.)
Additionally, the numbers of positive/negative labels are almost the same by reading data sequentially from the top. (https://github.com/optuna/optuna/pull/1542#discussion_r458805693)
Then, I think I could remove shuffling as the preprocessing.

train.jsonl

HideakiImamura: Thanks for the reply. I understood what you intended. Let me resolve this conversation.

examples/allennlp/allennlp_simple.py
toshihikoyanase: Sounds good 👍

examples/allennlp/allennlp_simple.py
toshihikoyanase: > I also remove it about python allennlp_simple.py as it is obvious to run a script

Sounds good! The change is aligned with #1566.

Issue #1535, opened 2020-07-20
Missing trial parameters with AllenNLP training


mateuszpieniak:
I encountered a weird outcome whilst optimizing my AllenNLP model, which might be tough to reproduce. I miss some rows in `trial_params`. Surprisingly the table `trials` contains results for the missing parameters which suggests that the runs were successful.

### Table `trials`
![Values](https://user-images.githubusercontent.com/31375424/87988567-d8fa7880-cae0-11ea-8738-43c703a5d12b.png)

### Table `trial_params`

![Params](https://user-images.githubusercontent.com/31375424/87988620-f0396600-cae0-11ea-859c-dcfb3ba8edf7.png)
Missing `trial_id` == 2, `trial_id` == 3, `trial_id` == 4. All other rows up to `trial_id` == 40 are fine.

## Expected behavior

A table `trial_params` is filled correctly.

## Environment

- Optuna version: 1.5.0 (https://github.com/mateuszpieniak/optuna)
I removed only `allennlp.common.params.infer_and_cast` in `AllenNLPExecutor`. It doesn't work with `null` values in jsonnet. In addition, I believe that such casting is not needed anyway since jsonnet was designed in such way that a user is responsible for types casting e.g. `parseInt`, `parseJson` (https://jsonnet.org/ref/stdlib.html)
- Python version: 3.6.9
- OS: Ubuntu 18.04.4 LTS
- AllenNLP 1.0.0 
I know it wasn't supported officially in Optuna 1.5.0 & experimental, but it should work anyway as there is no modification in subsequent versions of Optuna

## Steps to reproduce
I have 4 GPUs, thus I have 4 runs and I want them all to share my SQLite database.
1. Open 4 terminal tabs.
2. In each tab type `export CUDA_DEVICE=0`, `export CUDA_DEVICE=1`, and so on (https://optuna.readthedocs.io/en/stable/faq.html#how-can-i-use-two-gpus-for-evaluating-two-trials-simultaneously).
3. In each tab type `python optuna_code.py`

### optuna_code.py
```
from optuna import Trial, create_study
from optuna.integration.allennlp import AllenNLPExecutor


def objective(trial: Trial) -> float:
    # Requires to define CUDA_DEVICE & DEBUG env variable externally to support multi GPU
    trial.suggest_categorical("POOLING", ["mean", "cls"])
    trial.suggest_float("DROPOUT", 0.0, 0.8)
    trial.suggest_float("ALPHA", 0.0, 1.0)
[...]

PR #1544, opened 2020-07-22
Remove `allennlp.common.params.infer_and_cast` from AllenNLP integrations.

himkt:
Related issue: #1535

## Motivation

Current `AllenNLPExecutor` and `dump_best_config` call `allennlp.common.params.infer_and_cast` to cast string literals from environment variables into desired types (in almost case, int or float). By calling `infer_and_cast`, user can define search space in Jsonnet simply by writing `some_parameter: std.extVar('some_parameter')`.

However, it has a drawback; an executor doesn't work with `null` in jsonnet. It would be better to let users cast their parameters in Jsonnet (by using `std.parseInt` for integer and `std.parseJson` for floating-point). See #1535 for details.

## Description of changes

In this PR, I remove `infer_and_cast` from `AllenNLPExecutor` and `dump_best_config` and update a docstring to instruct users to use Jsonnet functionalities for casting parameters.

I also update an example and tests as well.

### DIFF
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -42,7 +42,6 @@ def dump_best_config(input_config_file: str, output_config_file: str, study: opt
     for key, value in best_params.items():
         best_params[key] = str(value)
     best_config = json.loads(_jsonnet.evaluate_file(input_config_file, ext_vars=best_params))
-    best_config = allennlp.common.params.infer_and_cast(best_config)
 
     with open(output_config_file, "w") as f:
         json.dump(best_config, f, indent=4)
@@ -56,8 +55,11 @@ class AllenNLPExecutor(object):
     The interface may change without prior notice to correspond to the update.
 
     See the examples of `objective function <https://github.com/optuna/optuna/blob/
-    master/examples/allennlp/allennlp_jsonnet.py>`_ and
-    `config file <https://github.com/optuna/optuna/blob/master/
+    master/examples/allennlp/allennlp_jsonnet.py>`_.
+
+    From Optuna v2.1.0, user has to cast their parameters by using methods in Jsonnet.
+    Call ``std.parseInt`` for integer, or ``std.parseJson`` for floating point.
+    Please see the example of `config file <https://github.com/optuna/optuna/blob/master/
     examples/allennlp/classifier.jsonnet>`_.
 
     .. note::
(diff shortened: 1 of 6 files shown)


optuna/integration/allennlp.py
```diff
@@ -56,8 +55,11 @@ class AllenNLPExecutor(object):
...
 
     See the examples of `objective function <https://github.com/optuna/optuna/blob/
-    master/examples/allennlp/allennlp_jsonnet.py>`_ and
-    `config file <https://github.com/optuna/optuna/blob/master/
+    master/examples/allennlp/allennlp_jsonnet.py>`_.
+
+    From Optuna v2.1.0, user has to cast their parameters by using methods in Jsonnet.
+    Call ``std.parseInt`` for integer, or ``std.parseJson`` for floating point.
+    Please see the example of `config file <https://github.com/optuna/optuna/blob/master/
     examples/allennlp/classifier.jsonnet>`_.
```
HideakiImamura, line 63: I think it's better to make these lines as "Note".
```suggestion
    .. note::
        From Optuna v2.1.0, user has to cast their parameters by using methods in Jsonnet.
        Call ``std.parseInt`` for integer, or ``std.parseJson`` for floating point.
        Please see the example of `config file <https://github.com/optuna/optuna/blob/master/
        examples/allennlp/classifier.jsonnet>`_.
```

himkt:
> The std.parseInt(str) returns the integer, but the std.parseJson(str) returns the dictionary according to the reference. Is it reasonable here?

Thank you for putting your eyes on!
First of all, AFAIK, single string or number is also acceptable for JSON, so `std.parseJson` could be non-dictionary value.

Although I agree with you that it might feel strange to use `parseJson` for parsing floating points,
this approach seems to be the only way for... 🤔  (discussion about `parseFloat`: https://github.com/google/jsonnet/issues/759)

HideakiImamura:
Thanks for the explanation! The changes look good to me. Is there anything you're worried about, so that's in the draft?

himkt:
Oh, you're right. I changed the status of PR. Thank you for pointing out!

optuna/integration/allennlp.py
```diff
@@ -56,9 +55,13 @@ class AllenNLPExecutor(object):
     The interface may change without prior notice to correspond to the update.
 
     See the examples of `objective function <https://github.com/optuna/optuna/blob/
-    master/examples/allennlp/allennlp_jsonnet.py>`_ and
-    `config file <https://github.com/optuna/optuna/blob/master/
-    examples/allennlp/classifier.jsonnet>`_.
+    master/examples/allennlp/allennlp_jsonnet.py>`_.
+
+    .. note::
+        From Optuna v2.1.0, user has to cast their parameters by using methods in Jsonnet.
```
toshihikoyanase, line 61: A nit, but `user` is countable. How about using the plural form?
```suggestion
        From Optuna v2.1.0, users have to cast their parameters by using methods in Jsonnet.
```

optuna/integration/allennlp.py
```diff
@@ -56,9 +55,13 @@ class AllenNLPExecutor(object):
...
+        Call ``std.parseInt`` for integer, or ``std.parseJson`` for floating point.
+        Please see the example of `config file <https://github.com/optuna/optuna/blob/master/
```
toshihikoyanase, line 63: I thin the file here is also countable.
```suggestion
        Please see the `example configuration <https://github.com/optuna/optuna/blob/master/
```

himkt:
Thanks a bunch! I reflected your reviews, PTAL.

himkt:
@toshihikoyanase 
Thanks a bunch! I reflected your reviews, PTAL.

toshihikoyanase:
@himkt Thank you for your swift action, but I found the change in https://github.com/optuna/optuna/commit/44b3d848c81c8963a0c0e45ef9d8d31cfa6bb52b was reverted by  https://github.com/optuna/optuna/commit/bdfd750bc8b062b56282efec7e9233417db34486  during the force push. Is it intentional?

himkt:
Ugh, I'm really sorry for bothering you.
I reverted the commits for reflecting your suggestions by merging the latest master branch.
6848bd8 reflects the changes you gave me.

himkt:
@toshihikoyanase 

Ugh, I'm really sorry for bothering you.
I reverted the commits for reflecting your suggestions by merging the latest master branch.
6848bd8 reflects the changes you gave me.

PR #1548, opened 2020-07-23
Add IPOP-CMA-ES support in `CmaEsSampler`.

c-bata:
## Motivation

As the benchmark of two-dimensional Rastrigin function shows, IPOP-CMA-ES outperforms non-restarting CMA-ES on multimodal functions.

rastrigin-dim-2-9855c695e5b1f00465f21e9fd2bebc2e4611a27a42ec8774548e78d4a7359a09

[Auger, A., Hansen, N.: A restart CMA evolution strategy with increasing population size. In: Proceedings of the 2005 IEEE Congress on Evolutionary Computation (CEC’2005), pp. 1769–1776 (2005a)](https://sci2s.ugr.es/sites/default/files/files/TematicWebSites/EAMHCO/contributionsCEC05/auger05ARCMA.pdf)

## Description of the changes

* Add `restart_strategy` argument. Currently accepts `ipop` or `None` only. I will implement BIPOP-CMA-ES in the following pull request.
* After each `optimizer.tell(solutions)` call, check whether `optimizer.should_stop()` method returns True or not.
    * If True, re-initialize CMA object with multiplied popsize by `inc_popsize`. `inc_popsize` is 2 by default. According to the paper, it reveal similar performance for factors between 2 and 3.
    * Sample initial starting point uniformly at each restart.

## TODO:

This PR is ready for review.

* [x] Benchmark.
    * Script: https://github.com/CyberAgent/cmaes/blob/master/benchmark/runner.sh
    * Command: `BUDGET=2500 REPEATS=10 DIM=2 ./benchmark/runner.sh rastrigin ./kurobako-report.json`
* [x] Add test case.
* [x] Merge #1558 into master.
* [x] Release the new version of cmaes to PyPI and conda-forge.
    * https://github.com/CyberAgent/cmaes/releases/tag/v0.6.0
* [x] Update `install_requires`.


optuna/samplers/_cmaes.py
```diff
@@ -131,22 +147,39 @@ def __init__(
...
         self._search_space = optuna.samplers.IntersectionSearchSpace()
         self._consider_pruned_trials = consider_pruned_trials
+        self._restart_strategy = restart_strategy
+        self._inc_popsize = inc_popsize
+        self._n_restarts = 0
 
         if self._consider_pruned_trials:
             self._raise_experimental_warning_for_consider_pruned_trials()
 
+        # TODO(c-bata): Support BIPOP-CMA-ES.
```
c-bata, line 157: Refs: https://github.com/CyberAgent/cmaes/blob/master/examples/bipop_cmaes.py

optuna/samplers/_cmaes.py
```diff
@@ -108,6 +109,20 @@ def objective(trial):
...
+
+        restart_strategy:
+            Strategy for restarting CMA-ES optimization when converges to a local minimum.
+            If given :obj:`None`, CMA-ES will not restart (default).
+            If given 'ipop', CMA-ES will restart with increasing population size.
+            Please see also ``inc_popsize`` parameter.
+
+            .. note::
+                Added in v2.1.0 as an experimental feature. The interface may change in newer
+                versions without prior notice.
```
c-bata, line 121: I marked this argument as an experimental. But it doesn't need to do because I'm honestly not planning to change the interface.

optuna/samplers/_cmaes.py
```diff
@@ -58,7 +58,8 @@ def objective(trial):
 
         x0:
             A dictionary of an initial parameter values for CMA-ES. By default, the mean of ``low``
-            and ``high`` for each distribution is used.
+            and ``high`` for each distribution is used. Note that x0 is sampled uniformly within
+            the search space domain for each restart if you set ``restart_strategy`` argument.
```
c-bata, line 62: memo: I comment the reason why I choose this design here.

In pycma's implementation, `x0` accepts not only the float values. It also accepts callable object (e.g. `np.random.rand`) and string (e.g. `"np.random.rand"` which is evaluated by `eval()`). But I guess it's enough for almost every usecases to sample `x0` uniformly at each restart.

optuna/samplers/_cmaes.py
c-bata: I removed experimental label at a2403fcc.

optuna/samplers/_cmaes.py
c-bata: I implemented IPOP-CMA-ES and BIPOP-CMA-ES on Goptuna. The benchmark results are below.

rastrigin-function-dim-2-1a14f4610cabf55e51b3b006bfe723dcadef3a57d9eb5c82a09c2c9fbb804c42

optuna/samplers/_cmaes.py
```diff
@@ -365,7 +401,7 @@ def _to_optuna_param(distribution: BaseDistribution, cma_param: float) -> Any:
     return cma_param
 
 
-def _initialize_x0(search_space: Dict[str, BaseDistribution]) -> Dict[str, np.ndarray]:
+def _initialize_x0(search_space: Dict[str, BaseDistribution]) -> Dict[str, Any]:
```
HideakiImamura, line 404: If `self._x0` is set in `_init_optimizer()`, its type is guaranteed as `Dict[str, np.ndarray]`.

optuna/samplers/_cmaes.py
```diff
@@ -242,30 +272,39 @@ def sample_relative(
...
                 continue
-            return pickle.loads(bytes.fromhex(serialized_optimizer))
+            n_restarts = trial.system_attrs.get("cma:n_restarts", 0)  # type: int
+            return pickle.loads(bytes.fromhex(serialized_optimizer)), n_restarts
+        return None, 0
 
-        # Init a CMA object.
-        if self._x0 is None:
-            self._x0 = _initialize_x0(search_space)
+    def _init_optimizer(
```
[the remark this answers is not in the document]
c-bata: Yes, either is fine.

optuna/samplers/_cmaes.py
c-bata: We defined `self._x0` like the following.

https://github.com/optuna/optuna/blob/79fadf4b7a221bc55656d881f45ded60402bf2fd/optuna/samplers/_cmaes.py#L126-L137

And `_initialize_x0` returns external representation (e.g. it returns integer object if given IntUniformDistribution).

https://github.com/optuna/optuna/blob/79fadf4b7a221bc55656d881f45ded60402bf2fd/optuna/samplers/_cmaes.py#L404-L412

So I guess `Dict[str, Any]` is correct. What do you think?

c-bata: Hmm... `DiscreteUniformDistribution` returns internal representation and `IntLogUniformDistribution` returns float value 🤔  These values looks like internal representation.

https://github.com/optuna/optuna/blob/79fadf4b7a221bc55656d881f45ded60402bf2fd/optuna/samplers/_cmaes.py#L413-L416

optuna/samplers/_cmaes.py
```diff
@@ -94,6 +95,16 @@ def objective(trial):
             Note that the parameters of the first trial in a study are always sampled
             via an independent sampler, so no warning messages are emitted in this case.
 
+        restart_strategy:
+            Strategy for restarting CMA-ES optimization when converges to a local minimum.
+            If given :obj:`None`, CMA-ES will not restart (default).
+            If given 'ipop', CMA-ES will restart with increasing population size.
+            Please see also ``inc_popsize`` parameter.
```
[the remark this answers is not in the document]
c-bata: Thank you for your suggestion. I pushed 983c8a8.

ScreenShot 2020-08-01 19 03 43

optuna/samplers/_cmaes.py
```diff
@@ -365,23 +418,59 @@ def _to_optuna_param(distribution: BaseDistribution, cma_param: float) -> Any:
...
-def _initialize_x0(search_space: Dict[str, BaseDistribution]) -> Dict[str, np.ndarray]:
+def _initialize_x0(search_space: Dict[str, BaseDistribution]) -> Dict[str, float]:
     x0 = {}
     for name, distribution in search_space.items():
-        if isinstance(distribution, optuna.distributions.UniformDistribution):
-            x0[name] = np.mean([distribution.high, distribution.low])
-        elif isinstance(distribution, optuna.distributions.DiscreteUniformDistribution):
-            x0[name] = np.mean([distribution.high, distribution.low])
-        elif isinstance(distribution, optuna.distributions.IntUniformDistribution):
-            x0[name] = int(np.mean([distribution.high, distribution.low]))
```
c-bata, line 376: I noticed that this returns as external representation. But it seems to return internal representation here because x0 should be `1.5` when given `IntUniformDistribution(low=1, high=3)`.

optuna/samplers/_cmaes.py
```diff
@@ -242,30 +283,46 @@ def sample_relative(
...
+            mean = np.array([x0[k] for k in ordered_keys], dtype=float)
+        elif self._x0 is None:
+            # `_initialize_x0` returns internal representations.
+            x0 = _initialize_x0(search_space)
+            mean = np.array([x0[k] for k in ordered_keys], dtype=float)
+        else:
+            # `self._x0` is external representations.
+            mean = np.array(
+                [_to_cma_param(search_space[k], self._x0[k]) for k in ordered_keys], dtype=float
+            )
```
c-bata, line 325: The users set external representations at `x0` argument of the constructor. So we need to convert it to cma_param. So previous logic seems to be broken...

`_initialize_x0` and `_initialize_x0_uniformly` returns its internal representations as I commented https://github.com/optuna/optuna/pull/1548/files#r463950014.

optuna/samplers/_cmaes.py
c-bata: Thank you for pointing it out. I found a bug of x0 initialization and fix it. Could you check my following comments?

* https://github.com/optuna/optuna/pull/1548#discussion_r463950014
* https://github.com/optuna/optuna/pull/1548#discussion_r463950471

optuna/samplers/_cmaes.py
```diff
@@ -94,6 +106,16 @@ def objective(trial):
             Note that the parameters of the first trial in a study are always sampled
             via an independent sampler, so no warning messages are emitted in this case.
 
+        restart_strategy:
+            Strategy for restarting CMA-ES optimization when converges to a local minimum.
+            If given :obj:`None`, CMA-ES will not restart (default).
+            If given 'ipop', CMA-ES will restart with increasing population size.
+            Please see also ``inc_popsize`` parameter.
```
HideakiImamura, line 113: As commented on https://github.com/optuna/optuna/pull/1548#discussion_r463354929, how about using the experimental decorator to this newly introduced feature? For the docs, it is appropriate to note it as same as `consider_pruned_trials`:
```suggestion
        restart_strategy:
            Strategy for restarting CMA-ES optimization when converges to a local minimum.
            If given :obj:`None`, CMA-ES will not restart (default).
            If given 'ipop', CMA-ES will restart with increasing population size.
            Please see also ``inc_popsize`` parameter.
            
            .. note::
                Added in v2.1.0 as an experimental feature. The interface may change in newer
                versions without prior notice. See
                https://github.com/optuna/optuna/releases/tag/v2.1.0.
```

optuna/samplers/_cmaes.py
```diff
@@ -131,10 +154,20 @@ def __init__(
...
             self._raise_experimental_warning_for_consider_pruned_trials()
 
+        # TODO(c-bata): Support BIPOP-CMA-ES.
+        if restart_strategy not in ("ipop", None,):
+            raise ValueError(
+                "restart_strategy={} is unsupported. Please specify: 'ipop' or None.".format(
+                    restart_strategy
+                )
+            )
+
```
HideakiImamura, line 170: We can use the experimental decorator for the argument as follows.
```suggestion

@experimental("2.1.0", name="`restart_strategy != None` in CmaEsSampler")
    def _raise_experimental_warning_for_restart_strategy(self) -> None:
        pass

```

optuna/samplers/_cmaes.py
```diff
@@ -131,10 +154,20 @@ def __init__(
         self._cma_rng = np.random.RandomState(seed)
         self._search_space = optuna.samplers.IntersectionSearchSpace()
         self._consider_pruned_trials = consider_pruned_trials
+        self._restart_strategy = restart_strategy
+        self._inc_popsize = inc_popsize
```
HideakiImamura, line 159: Then, we can use the defined function to raise an experimental warning when `restart_strategy != None`:
```suggestion

        if self._restart_strategy:
            self._raise_experimental_warning_for_restart_strategy()

```

optuna/samplers/_cmaes.py
```diff
@@ -118,10 +140,11 @@ def __init__(
         independent_sampler: Optional[BaseSampler] = None,
         warn_independent_sampling: bool = True,
         seed: Optional[int] = None,
+        restart_strategy: Optional[str] = None,
+        inc_popsize: int = 2,
         *,
         consider_pruned_trials: bool = False
```
c-bata, line 146: Made keyword only arguments.

```suggestion
        *,
        consider_pruned_trials: bool = False,
        restart_strategy: Optional[str] = None,
        inc_popsize: int = 2
```

c-bata:
We decided to set experimental decorator at our off-line discussion.
Thank you @HideakiImamura!

HideakiImamura (approved):
Thanks for the update! LGTM!

optuna/samplers/_cmaes.py
```diff
@@ -242,30 +295,46 @@ def sample_relative(
...
+    ) -> Tuple[Optional[CMA], int]:
         # Restore a previous CMA object.
         for trial in reversed(completed_trials):
             serialized_optimizer = trial.system_attrs.get(
                 "cma:optimizer", None
             )  # type: Optional[str]
             if serialized_optimizer is None:
                 continue
-            return pickle.loads(bytes.fromhex(serialized_optimizer))
+            n_restarts = trial.system_attrs.get("cma:n_restarts", 0)  # type: int
```
hvy, line 314: I think we can implement IPOP-CMA-ES without introducing any additional system attributes. Although this system attribute, is modified as a counter, it seems like it's only read as a "boolean flag" to determine whether x0 should be initialized uniformly or not, in `_init_optimizer`. The value of this "flag" can be derived from the context when calling `_init_optimizer`, i.e. when called when the optimizer should stop, then initialize x0 uniformly, otherwise, don't. A counter might be useful if we for instance want to cap the number of restarts but I otherwise think we can simplify the implementation.

hvy (commented):
Thanks a lot for this very interesting PR. 

The implementation is simple and the logic looks good for the most part. I left one question though regarding the system attribute if you could take a look.

hvy:
> We decided to set experimental decorator at our off-line discussion.
Thank you @HideakiImamura!

Might be good for future record to leave a comment on why. 🙂

optuna/samplers/_cmaes.py
c-bata: Thank you for your review!
I agree with your opinion. I removed `n_restarts` from system_attrs.

c-bata:
> Might be good for future record to leave a comment on why. 🙂

We set experimental decorator because there are some room for discussion about API (e.g. it might be better to make `inc_popsize` float). And we might find the room for improvements after dogfooding this feature.

hvy (commented):
Thanks for your update. I thought that it basically looked good but noticed that the current logic will when collecting previous trial history include trials from "previous restarts". This is because the generation count is reset on restart (since a `CMA` object is re-instantiated) and we simply collect all trials belonging to the "same generation" and don't care if they stem from different restarts. Is this intended? Maybe we have to include the restart count after all and filter trials based on it in addition to the generation?

c-bata:
Oh, thanks. You are completely correct. It is not intended behavior.
To fix the bug, we need to add `n_restarts` or something optimizer unique information into the key of system_attrs.

hvy:
I'm really sorry after having asked you to remove `n_restarts` but reverting https://github.com/optuna/optuna/pull/1548/commits/4947c97583c54f5ce8f75c202cb8690e8e6da705 and also filtering based on this count when constructing `solution_trials` could be a reasonable fix?

c-bata:
> I'm really sorry after having asked you to remove n_restarts but reverting 4947c97

No. It's not your fault. It's just my mistake.

> reverting 4947c97 and also filtering based on this count when constructing solution_trials could be a reasonable fix?

Yeah, sounds reasonable. I'll fix the problem. Thank you!

c-bata:
I fixed the bug at cd2e10c.

hvy (approved):
Thanks for the quick fix and the long running effort. Changes LGTM!

Also leaving the results from a small experiment run locally. It's clear that the restart makes it more explorative escaping a local minimum.

#### Vanilla

newplot (18)

#### IPOP

newplot (19)

```python
import optuna

# Himmelblau
def objective(trial):
    x = trial.suggest_float("x", -6, 6)
    y = trial.suggest_float("y", -6, 6)
    return (x ** 2 + y - 11) ** 2 + (x + y ** 2 - 7) ** 2


if __name__ == "__main__":
    study = optuna.create_study(sampler=optuna.samplers.CmaEsSampler(restart_strategy="ipop"))
    optuna.visualization.plot_slice(study).show()
```

hvy:
@HideakiImamura might want to have another look as well, so I'll leave the final word to him.

c-bata:
I re-execute Rastrigin (2D) benchmark function at d35a3dc.

rastrigin-dim-2-9855c695e5b1f00465f21e9fd2bebc2e4611a27a42ec8774548e78d4a7359a09

HideakiImamura (approved):
Thanks for all of your efforts. LGTM, again!

c-bata:
Thank you for your thorough review!

PR #1556, opened 2020-07-26
Add type hints to storage.

keisuke-umezawa:
## Motivation

Please refer to the Issue #711 .

## Description of the changes

I added type hints into optuna/storage

### DIFF
--- a/optuna/storages/_in_memory.py
+++ b/optuna/storages/_in_memory.py
@@ -1,13 +1,12 @@
 import copy
 from datetime import datetime
 import threading
-import uuid
-
 from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Tuple
+import uuid
 
 from optuna import distributions  # NOQA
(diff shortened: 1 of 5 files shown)


optuna/storages/_in_memory.py
```diff
@@ -3,6 +3,12 @@
 import threading
 import uuid
 
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
```
hvy, line 10: A nit but you can place the `typing` imports with the other std imports, without separating them with an empty line. Ditto for the other files in this PR 🙂 .

hvy:
The hash mismatch error on CircleCI seems unrelated, but I've never seen it before. 🤔

optuna/storages/_in_memory.py
keisuke-umezawa: Thank you! I fixed them.

PR #1558, opened 2020-07-26
Avoid to sample the value which equals to upper bound

### LINKED ISSUE #1557: Independent sampler is triggered even if the search space is not dynamic.
## Expected behavior

Sometimes, independent sampler is triggered though the search space is clearly not dynamic.

```python
import numpy as np
import optuna


class Rastrigin:
    """
    https://www.sfu.ca/~ssurjano/rastr.html
    """
    def __init__(self, n):
        self.n = n

    def __call__(self, trial):
        x = np.array([trial.suggest_uniform(f"x{i}", -5.12, 5.12) for i in range(self.n)])
        return 10 * self.n + np.sum(x ** 2 - 10 * np.cos(2 * np.pi * x))


if __name__ == "__main__":
    objective = Rastrigin(8)

    for seed in range(10):
        study = optuna.create_study(
            sampler=optuna.samplers.CmaEsSampler(seed=seed)
        )
        study.optimize(objective, n_trials=1500)
        print("Best value: {} (params: {})\n".format(study.best_value, study.best_params))
```

## Environment

- Optuna version: master (rev: 278034a9)
- Python version: 3.8.2
- OS:macOS

```
$ python examples/quadratic_simple.py 
[W 2020-07-26 22:51:19,360] The parameter 'x6' in trial#1328 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
Best value: 40.1899514495306 (params: {'x0': 0.995858149009119, 'x1': -1.1351238915698543, 'x2': -0.1801230004399368, 'x3': 1.9177514176898494, 'x4': 0.1605867098097168, 'x5': -2.0070995883976925, 'x6': 1.2348466691400497, 'x7': -1.1301172829126584})

Best value: 32.4631637787495 (params: {'x0': -0.1888686495274782, 'x1': -0.8994779770534349, 'x2': 0.1759863776573024, 'x3': 0.08718985952421743, 'x4': -0.21347527704332003, 'x5': 0.1533930220554347, 'x6': -0.15323426159592224, 'x7': -0.009436181675708921})

Best value: 31.91496288803713 (params: {'x0': 0.02690799189465823, 'x1': 0.9301003789280453, 'x2': -0.9063281146700595, 'x3': -0.11441781075028418, 'x4': 0.8908104409619761, 'x5': -0.55904514348641, 'x6': -1.0243546160884258, 'x7': 1.0191798909768646})

[W 2020-07-26 22:51:26,492] The parameter 'x3' in trial#1483 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:26,492] The parameter 'x4' in trial#1483 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:26,493] The parameter 'x5' in trial#1483 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:26,515] The parameter 'x1' in trial#1492 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:26,515] The parameter 'x3' in trial#1492 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:26,516] The parameter 'x6' in trial#1492 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:26,524] The parameter 'x2' in trial#1495 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:26,525] The parameter 'x3' in trial#1495 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
Best value: 42.284741734204616 (params: {'x0': 0.8611571722449447, 'x1': -1.0202710839736775, 'x2': -0.9161870092865121, 'x3': 1.1074296676201707, 'x4': 2.020475272523612, 'x5': 2.6341278204885925, 'x6': -0.961020444464636, 'x7': -1.075404689687082})

Best value: 35.837188844247734 (params: {'x0': 0.16491643650172522, 'x1': -0.9135554351547925, 'x2': -0.15711841139753724, 'x3': -0.7403737469975655, 'x4': 0.9699238273692735, 'x5': -0.9646192974710259, 'x6': -0.24383102905966858, 'x7': -0.9980549047856908})

[W 2020-07-26 22:51:30,744] The parameter 'x7' in trial#1187 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:31,066] The parameter 'x3' in trial#1324 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:31,067] The parameter 'x4' in trial#1324 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:31,067] The parameter 'x7' in trial#1324 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:31,398] The parameter 'x2' in trial#1478 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:31,435] The parameter 'x3' in trial#1492 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
[W 2020-07-26 22:51:31,435] The parameter 'x7' in trial#1492 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
Best value: 29.19879250598784 (params: {'x0': 2.0770115850339996, 'x1': 1.0571498004291093, 'x2': 1.0843035099445237, 'x3': 1.0745606517288824, 'x4': -1.8806912321660192, 'x5': -0.9559137545222565, 'x6': -1.193838173651328, 'x7': -0.09452242291271634})

Best value: 29.34525115041371 (params: {'x0': 1.0355545841547409, 'x1': -0.14212348444599865, 'x2': -0.09354311427862283, 'x3': -0.10414346874509228, 'x4': -1.1537474227287918, 'x5': 0.9559745956916987, 'x6': 0.21069559289098455, 'x7': -1.1597322728811668})

: (omit)
```

c-bata:
## Motivation

To fix #1557.

## Description of the changes

It is not prohibited to sample the value which equals to upper bound in CyberAgent/cmaes (and pycma). But Optuna's continuous distribution only accepts `[low, high)`. So I cap the upper bound with tiny value to fix the problem which I reported to #1557.

### DIFF
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -314,13 +314,13 @@ class _Optimizer(object):
             elif isinstance(dist, DiscreteUniformDistribution):
                 r = dist.high - dist.low
                 lows.append(0 - 0.5 * dist.q)
-                highs.append(r + 0.5 * dist.q - _EPS)
+                highs.append(r + 0.5 * dist.q)
             elif isinstance(dist, IntUniformDistribution):
                 lows.append(dist.low - 0.5 * dist.step)
-                highs.append(dist.high + 0.5 * dist.step - _EPS)
+                highs.append(dist.high + 0.5 * dist.step)
             elif isinstance(dist, IntLogUniformDistribution):
                 lows.append(self._to_cma_params(search_space, param_name, dist.low - 0.5))
-                highs.append(self._to_cma_params(search_space, param_name, dist.high + 0.5) - _EPS)
+                highs.append(self._to_cma_params(search_space, param_name, dist.high + 0.5))
             else:
                 raise NotImplementedError("The distribution {} is not implemented.".format(dist))
 
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -38,6 +38,9 @@ _logger = logging.get_logger(__name__)
 # Minimum value of sigma0 to avoid ZeroDivisionError in cma.CMAEvolutionStrategy.
 _MIN_SIGMA0 = 1e-10
 
+# Tiny value that avoids to sample upper bound of the search space.
+_UPPER_BOUND_CAP = 1e-10
+
 
 class PyCmaSampler(BaseSampler):
     """A Sampler using cma library as the backend.
@@ -310,17 +313,17 @@ class _Optimizer(object):
                 highs.append(len(dist.choices) - 0.5)
             elif isinstance(dist, UniformDistribution) or isinstance(dist, LogUniformDistribution):
                 lows.append(self._to_cma_params(search_space, param_name, dist.low))
-                highs.append(self._to_cma_params(search_space, param_name, dist.high))
+                highs.append(self._to_cma_params(search_space, param_name, dist.high) - _UPPER_BOUND_CAP)
             elif isinstance(dist, DiscreteUniformDistribution):
                 r = dist.high - dist.low
                 lows.append(0 - 0.5 * dist.q)
-                highs.append(r + 0.5 * dist.q)
+                highs.append(r + 0.5 * dist.q - _UPPER_BOUND_CAP)
             elif isinstance(dist, IntUniformDistribution):
                 lows.append(dist.low - 0.5 * dist.step)
-                highs.append(dist.high + 0.5 * dist.step)
+                highs.append(dist.high + 0.5 * dist.step - _UPPER_BOUND_CAP)
             elif isinstance(dist, IntLogUniformDistribution):
                 lows.append(self._to_cma_params(search_space, param_name, dist.low - 0.5))
-                highs.append(self._to_cma_params(search_space, param_name, dist.high + 0.5))
+                highs.append(self._to_cma_params(search_space, param_name, dist.high + 0.5 - _UPPER_BOUND_CAP))
             else:
                 raise NotImplementedError("The distribution {} is not implemented.".format(dist))
 
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -35,11 +35,7 @@ if type_checking.TYPE_CHECKING:
 
 _logger = logging.get_logger(__name__)
 
-# Minimum value of sigma0 to avoid ZeroDivisionError in cma.CMAEvolutionStrategy.
-_MIN_SIGMA0 = 1e-10
-
-# Tiny value that avoids to sample upper bound of the search space.
-_UPPER_BOUND_CAP = 1e-10
+_EPS = 1e-10
 
 
 class PyCmaSampler(BaseSampler):
@@ -217,7 +213,8 @@ class PyCmaSampler(BaseSampler):
             sigma0 = self._initialize_sigma0(search_space)
         else:
             sigma0 = self._sigma0
-        sigma0 = max(sigma0, _MIN_SIGMA0)
+        # Avoid ZeroDivisionError in cma.CMAEvolutionStrategy.
+        sigma0 = max(sigma0, _EPS)
 
         optimizer = _Optimizer(search_space, self._x0, sigma0, self._cma_stds, self._cma_opts)
         trials = study.trials
@@ -313,17 +310,17 @@ class _Optimizer(object):
                 highs.append(len(dist.choices) - 0.5)
             elif isinstance(dist, UniformDistribution) or isinstance(dist, LogUniformDistribution):
                 lows.append(self._to_cma_params(search_space, param_name, dist.low))
-                highs.append(self._to_cma_params(search_space, param_name, dist.high) - _UPPER_BOUND_CAP)
+                highs.append(self._to_cma_params(search_space, param_name, dist.high) - _EPS)
             elif isinstance(dist, DiscreteUniformDistribution):
                 r = dist.high - dist.low
                 lows.append(0 - 0.5 * dist.q)
-                highs.append(r + 0.5 * dist.q - _UPPER_BOUND_CAP)
+                highs.append(r + 0.5 * dist.q - _EPS)
             elif isinstance(dist, IntUniformDistribution):
                 lows.append(dist.low - 0.5 * dist.step)
-                highs.append(dist.high + 0.5 * dist.step - _UPPER_BOUND_CAP)
+                highs.append(dist.high + 0.5 * dist.step - _EPS)
             elif isinstance(dist, IntLogUniformDistribution):
                 lows.append(self._to_cma_params(search_space, param_name, dist.low - 0.5))
-                highs.append(self._to_cma_params(search_space, param_name, dist.high + 0.5 - _UPPER_BOUND_CAP))
+                highs.append(self._to_cma_params(search_space, param_name, dist.high + 0.5) - _EPS)
             else:
                 raise NotImplementedError("The distribution {} is not implemented.".format(dist))
 
--- a/tests/integration_tests/test_cma.py
+++ b/tests/integration_tests/test_cma.py
@@ -174,6 +174,9 @@ class TestOptimizer(object):
     @staticmethod
     def test_init(search_space: Dict[str, BaseDistribution], x0: Dict[str, Any]) -> None:
 
+        # TODO(c-bata): Avoid exact assertion checks
+        eps = 1e-10
+
         with patch("cma.CMAEvolutionStrategy") as mock_obj:
             optuna.integration.cma._Optimizer(
                 search_space, x0, 0.2, None, {"popsize": 5, "seed": 1}
@@ -184,8 +187,16 @@ class TestOptimizer(object):
                 {
                     "BoundaryHandler": cma.BoundTransform,
                     "bounds": [
-                        [-0.5, -1.0, -1.5, -2.0, math.log(1.5), math.log(0.001), -2,],
-                        [1.5, 11.0, 1.5, 4.0, math.log(16.5), math.log(0.1), 2],
+                        [-0.5, -1.0, -1.5, -2.0, math.log(1.5), math.log(0.001), -2],
+                        [
+                            1.5,
+                            11.0 - eps,
+                            1.5 - eps,
+                            4.0 - eps,
+                            math.log(16.5) - eps,
+                            math.log(0.1) - eps,
+                            2 - eps,
+                        ],
                     ],
                     "popsize": 5,
                     "seed": 1,
--- a/tests/integration_tests/test_cma.py
+++ b/tests/integration_tests/test_cma.py
@@ -188,15 +188,7 @@ class TestOptimizer(object):
                     "BoundaryHandler": cma.BoundTransform,
                     "bounds": [
                         [-0.5, -1.0, -1.5, -2.0, math.log(1.5), math.log(0.001), -2],
-                        [
-                            1.5,
-                            11.0 - eps,
-                            1.5 - eps,
-                            4.0 - eps,
-                            math.log(16.5) - eps,
-                            math.log(0.1) - eps,
-                            2 - eps,
-                        ],
+                        [1.5, 11.0, 1.5, 4.0, math.log(16.5), math.log(0.1) - eps, 2 - eps],
                     ],
                     "popsize": 5,
                     "seed": 1,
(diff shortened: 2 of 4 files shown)


c-bata:
I guess PyCmaSampler also has the same problem. If I checked the bug, I'll push the following patch.

```diff
diff --git a/optuna/integration/cma.py b/optuna/integration/cma.py
index ee61eafd..6d46ca8c 100644
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -38,6 +38,9 @@ _logger = logging.get_logger(__name__)
 # Minimum value of sigma0 to avoid ZeroDivisionError in cma.CMAEvolutionStrategy.
 _MIN_SIGMA0 = 1e-10
 
+# Tiny value that avoids to sample upper bound of the search space.
+_UPPER_BOUND_CAP = 1e-10
+
 
 class PyCmaSampler(BaseSampler):
     """A Sampler using cma library as the backend.
@@ -310,17 +313,17 @@ class _Optimizer(object):
                 highs.append(len(dist.choices) - 0.5)
             elif isinstance(dist, UniformDistribution) or isinstance(dist, LogUniformDistribution):
                 lows.append(self._to_cma_params(search_space, param_name, dist.low))
-                highs.append(self._to_cma_params(search_space, param_name, dist.high))
+                highs.append(self._to_cma_params(search_space, param_name, dist.high) - _UPPER_BOUND_CAP)
             elif isinstance(dist, DiscreteUniformDistribution):
                 r = dist.high - dist.low
                 lows.append(0 - 0.5 * dist.q)
-                highs.append(r + 0.5 * dist.q)
+                highs.append(r + 0.5 * dist.q - _UPPER_BOUND_CAP)
             elif isinstance(dist, IntUniformDistribution):
                 lows.append(dist.low - 0.5 * dist.step)
-                highs.append(dist.high + 0.5 * dist.step)
+                highs.append(dist.high + 0.5 * dist.step - _UPPER_BOUND_CAP)
             elif isinstance(dist, IntLogUniformDistribution):
                 lows.append(self._to_cma_params(search_space, param_name, dist.low - 0.5))
-                highs.append(self._to_cma_params(search_space, param_name, dist.high + 0.5))
+                highs.append(self._to_cma_params(search_space, param_name, dist.high + 0.5 - _UPPER_BOUND_CAP))
             else:
                 raise NotImplementedError("The distribution {} is not implemented.".format(dist))
```

c-bata:
I also successfully reproduce this problem on PyCmaSampler.

```python
import optuna


def objective(trial):
    return x + y


if __name__ == "__main__":
    study = optuna.create_study(
        direction="maximize",
        sampler=optuna.integration.PyCmaSampler())
    study.optimize(objective, timeout=20*60)
    print("Best value: {} (params: {})\n".format(study.best_value, study.best_params))
```

```
$ python examples/quadratic_simple.py 
[W 2020-07-27 03:59:46,045] The parameter 'y' in trial#394 is sampled independently by using `RandomSampler` instead of `PyCmaSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `PyCmaSampler`, if this independent sampling is intended behavior.
[W 2020-07-27 03:59:46,515] The parameter 'y' in trial#398 is sampled independently by using `RandomSampler` instead of `PyCmaSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `PyCmaSampler`, if this independent sampling is intended behavior.
[W 2020-07-27 03:59:48,669] The parameter 'x' in trial#415 is sampled independently by using `RandomSampler` instead of `PyCmaSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `PyCmaSampler`, if this independent sampling is intended behavior.
[W 2020-07-27 03:59:48,936] The parameter 'x' in trial#417 is sampled independently by using `RandomSampler` instead of `PyCmaSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `PyCmaSampler`, if this independent sampling is intended behavior.
...
```

optuna/integration/cma.py
```diff
@@ -310,17 +310,17 @@ def __init__(
                 highs.append(len(dist.choices) - 0.5)
             elif isinstance(dist, UniformDistribution) or isinstance(dist, LogUniformDistribution):
                 lows.append(self._to_cma_params(search_space, param_name, dist.low))
-                highs.append(self._to_cma_params(search_space, param_name, dist.high))
+                highs.append(self._to_cma_params(search_space, param_name, dist.high) - _EPS)
```
HideakiImamura, line 313: If the value of `low` and `high` is relatively the same order with `_EPS`, that is, `low ≒ high ≒ 1e-10`, then this subtraction may cause an invalid interval `[low, high)`. How about subtracting the relatively small values as follows?
```python
low = self._to_cma_params(search_space, param_name, dist.low)
high = self._to_cma_params(search_space, param_name, dist.high)
lows.append(low)
highs.append(high - (high - low) * _EPS)
```

tests/integration_tests/test_cma.py
```diff
@@ -184,8 +187,16 @@ def test_init(search_space: Dict[str, BaseDistribution], x0: Dict[str, Any]) ->
                 {
                     "BoundaryHandler": cma.BoundTransform,
                     "bounds": [
-                        [-0.5, -1.0, -1.5, -2.0, math.log(1.5), math.log(0.001), -2,],
-                        [1.5, 11.0, 1.5, 4.0, math.log(16.5), math.log(0.1), 2],
+                        [-0.5, -1.0, -1.5, -2.0, math.log(1.5), math.log(0.001), -2],
```
HideakiImamura, line 190: I'm either fine, but it might be good to keep as it is.
```suggestion
                        [-0.5, -1.0, -1.5, -2.0, math.log(1.5), math.log(0.001), -2,],
```

optuna/integration/cma.py
c-bata: We can easily reproduce the representation error which is grater than the situation that you assumed.

```python
>>> high = 1.1e-10
>>> low = 0.9e-10
>>> diff = (high - low) * 1e-10
>>> representation_error = abs(0.1 - (0.3-0.2))
>>> representation_ > diff
True
```

So I think we should ignore the edge case. What do you think?

tests/integration_tests/test_cma.py
c-bata: Thank you for the suggestion 👍  Honestly, I also either is fine but feels more natural that trimmed the comma when `list` is written in one line. So I make resolved this without applying the changes.

I guess the comma is added when `black` formatted from multiple lines into a single line like following.

```
[
    0.5,
    1.0,
]
↓ formatted
[0.5, 1.0,]
```

optuna/integration/cma.py
```diff
@@ -310,17 +310,17 @@ def __init__(
...
             elif isinstance(dist, DiscreteUniformDistribution):
                 r = dist.high - dist.low
                 lows.append(0 - 0.5 * dist.q)
-                highs.append(r + 0.5 * dist.q)
+                highs.append(r + 0.5 * dist.q - _EPS)
             elif isinstance(dist, IntUniformDistribution):
                 lows.append(dist.low - 0.5 * dist.step)
-                highs.append(dist.high + 0.5 * dist.step)
+                highs.append(dist.high + 0.5 * dist.step - _EPS)
```
hvy, line 320: Is it required for ints as well where the upper bound is inclusive?

c-bata:
> Great, thanks for finding this edge case. To somewhat clarify on the cause, I assume it's returning False unexpectedly when check whether a value is contained in a distribution or not in _is_relative_param?

Exactly. `distribution._contains(...)` returning False unexpectedly at `_is_relative_param()`.

I put a screenshot of pycharm debugger below.

ScreenShot 2020-07-26 23 19 08

CMA-ES samples `x6 = 5.12` though the search space domain of Rastrigin function is `[-5.12, 5.12]` (See https://www.sfu.ca/~ssurjano/rastr.html).  So `distribution._contains()` returns False at L725.
https://github.com/optuna/optuna/blob/dd9c4aa2a9cee42d11be77d856131ea3ac5995f7/optuna/trial/_trial.py#L725

optuna/integration/cma.py
c-bata: Thank you for your review! As you pointed, this patch is not required for integers (and discrete uniforms) because these distributions accepts upper bound. Just to be sure, I executed following example but found no issues.

```python
import optuna


def objective(trial):
    x = trial.suggest_int("x", 0, 2)
    y = trial.suggest_discrete_uniform("y", 0, 2, 0.5)
    return x + y


if __name__ == "__main__":
    study = optuna.create_study(
        direction="maximize",
        sampler=optuna.samplers.CmaEsSampler())
    study.optimize(objective, timeout=20*60)
    print("Best value: {} (params: {})\n".format(study.best_value, study.best_params))
```


I'll fix it 👍

c-bata: I pushed f355888. PTAL.

optuna/integration/cma.py
c-bata: I mean if we should consider such a situation, we should define `_EPS = 1e-11` or more tiny values.

HideakiImamura: Thanks for the clarification! I understood what you mean. I totally agree with you and resolve this conversation without any fixes.

PR #1562, opened 2020-07-27
Make sampling from trunc-norm efficient in TPE sampler.

ytsmiling:
## Motivation
This PR makes sampling from truncated normal distribution in TPE sampler efficient.

## Description of the changes
Samples from trunc-norm are resampled only when they violate the range condition.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -389,7 +389,7 @@ class TPESampler(BaseSampler):
         active = np.argmax(self._rng.multinomial(1, weights, size=size), axis=-1)
         trunc_low = (low - mus[active]) / sigmas[active]
         trunc_high = (high - mus[active]) / sigmas[active]
-        samples = np.full(size, fill_value=high + 1.0, dtype=np.float64)
+        samples = np.full((), fill_value=high + 1.0, dtype=np.float64)
         while (samples >= high).any():
             samples = np.where(
                 samples < high,
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -389,17 +389,20 @@ class TPESampler(BaseSampler):
         active = np.argmax(self._rng.multinomial(1, weights, size=size), axis=-1)
         trunc_low = (low - mus[active]) / sigmas[active]
         trunc_high = (high - mus[active]) / sigmas[active]
-        while True:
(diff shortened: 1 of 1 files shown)


optuna/samplers/_tpe/sampler.py
```diff
@@ -389,17 +389,20 @@ def _sample_from_gmm(
...
         trunc_high = (high - mus[active]) / sigmas[active]
-        while True:
-            samples = truncnorm.rvs(
-                trunc_low,
-                trunc_high,
-                size=size,
-                loc=mus[active],
-                scale=sigmas[active],
-                random_state=self._rng,
+        samples = np.ones(size) * (high + 1.0)
```
hvy, line 392: You can reduce memory allocations by using `numpy.full`.
```suggestion
        samples = np.full(size, fill_value=high + 1.0, dtype=np.float64)
```

ytsmiling: Thank you for the suggestion. I tested several implementations given this comment. The difference between them was small, but the suggested code was consistently faster (up to 8% in my environment) than the original code. The difference between the suggestion and other variants was negligible. So I'd like to use the suggested implementation.

optuna/samplers/_tpe/sampler.py
```diff
@@ -389,17 +389,20 @@ def _sample_from_gmm(
...
         trunc_high = (high - mus[active]) / sigmas[active]
-        while True:
-            samples = truncnorm.rvs(
-                trunc_low,
-                trunc_high,
-                size=size,
-                loc=mus[active],
-                scale=sigmas[active],
-                random_state=self._rng,
+        samples = np.full(size, fill_value=high + 1.0, dtype=np.float64)
```
hvy, line 392: Sorry for the late comment but I now recalled that [`numpy.where`](https://numpy.org/doc/stable/reference/generated/numpy.where.html) broadcasts. You don't have to allocate memory for `size` even. This should address the degradation for the general case w/o resampling.
```suggestion
        samples = np.full((), fill_value=high + 1.0, dtype=np.float64)
```

ytsmiling: Actually, I observed that the suggested code https://github.com/optuna/optuna/pull/1562#discussion_r462118803 imposes an only negligible overhead to the current master, and I think we don't need to care the degradation too much.
Nevertheless, the newer code seems clearly better and let me adopt the suggestion.

PR #1565, opened 2020-07-28
Refine the help of `study-name`

belldandyxtq:
This PR refines the help message of `study-name` in cli.

## Motivation
The same message was used for `create-study` and other, which can be
misleading as no new study will be created other than `create-study`, but
the explanation makes it sound like it will be.

## Description of the changes
This PR removes such misleading parts and makes things more clear.

### DIFF
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -198,7 +198,7 @@ class _Dashboard(_BaseCommand):
             "--study", default=None, help="This argument is deprecated. Use --study-name instead.",
         )
         parser.add_argument(
-            "--study-name", default=None, help="The name of the study to show in the dashboard.",
+            "--study-name", default=None, help="The name of the study to show on the dashboard.",
         )
         parser.add_argument(
             "--out",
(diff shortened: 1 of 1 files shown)


belldandyxtq:
Thank you for the advise, I applied the formatter

optuna/cli.py
```diff
@@ -126,9 +124,7 @@ def get_parser(self, prog_name):
             "--study", default=None, help="This argument is deprecated. Use --study-name instead.",
         )
         parser.add_argument(
-            "--study-name",
-            default=None,
-            help="A human-readable name of a study to distinguish it from others.",
+            "--study-name", default=None, help="The name of the study to set attribute to.",
```
toshihikoyanase, line 127: I’m not fully sure, but `attribute` seems to be countable. Do we need an article?
```suggestion
            "--study-name", default=None, help="The name of the study to set the attribute to.",
```

belldandyxtq: Done in 00dfd8f

toshihikoyanase:
@c-bata Let me merge this PR since all CI jobs successfully finished.

PR #1568, opened 2020-07-29
Add default implementation of get_n_trials.

keisuke-umezawa:
## Motivation

To reduce custom implementation of each storage class.
Related https://github.com/optuna/optuna/issues/1170.

## Description of the changes

- Implement default implementation for `get_n_trials`.

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -539,7 +539,6 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         """
         raise NotImplementedError
 
-    @abc.abstractmethod
     def get_n_trials(self, study_id: int, state: Optional[TrialState] = None) -> int:
         """Count the number of trials in a study.
 
@@ -556,7 +555,11 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
             :exc:`KeyError`:
                 If no study with the matching ``study_id`` exists.
         """
-        raise NotImplementedError
+        self.read_trials_from_remote_storage(study_id)
+        if state is None:
+            return len(self.get_all_trials(study_id, deepcopy=False))
+
+        return len([t for t in self.get_all_trials(study_id, deepcopy=False) if t.state == state])
 
     def get_best_trial(self, study_id: int) -> FrozenTrial:
         """Return the trial with the best value in a study.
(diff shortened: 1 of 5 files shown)


keisuke-umezawa:
@ytsmiling 
For `Add more default implementations in BaseStorage class.`, I create a draft PR. Is it something you meant by `Add more default implementations in BaseStorage class.`?

optuna/storages/_base.py
```diff
@@ -636,6 +639,7 @@ def get_trial_system_attrs(self, trial_id: int) -> Dict[str, Any]:
         """
         return self.get_trial(trial_id).system_attrs
 
+    @abc.abstractmethod
```
hvy, line 642: Not sure this line is needed but maybe you had something in mind? Wondering since it's a breaking change for (possibly user defined) storage classes.

optuna/storages/_base.py
```diff
@@ -556,7 +555,11 @@ def get_n_trials(self, study_id: int, state: Optional[TrialState] = None) -> int
             :exc:`KeyError`:
                 If no study with the matching ``study_id`` exists.
         """
-        raise NotImplementedError
+        self.read_trials_from_remote_storage(study_id)
+        if state is None:
+            return len(self.get_all_trials(study_id, deepcopy=False))
+
+        return len([t for t in self.get_all_trials(study_id, deepcopy=False) if t.state == state])
```
hvy, line 562: The logic looks good, but do you think we should add tests for the `state` parameter? Seems like it's not tested, going through `test_storages.py`.

optuna/storages/_base.py
```diff
@@ -556,7 +555,11 @@ def get_n_trials(self, study_id: int, state: Optional[TrialState] = None) -> int
             :exc:`KeyError`:
                 If no study with the matching ``study_id`` exists.
         """
-        raise NotImplementedError
+        self.read_trials_from_remote_storage(study_id)
```
hvy, line 558: Sorry for the delayed comment. `read_trials_from_remote_storage` is actually meant to be called from outside the storage, to sync only when necessary, and currently only from the `_CachedStorage`. Given that this `_CachesStorage` is private and not expected to have been used directly by the user, I think we can safely remove this line. 
```suggestion
```

keisuke-umezawa:
@hvy It is ready for reviewing.

hvy (approved):
LGTM! Thanks for the changes and sorry for the delay.

Again, I'd vote for deprecating this method as a follow up but it's open for discussion.

hvy:
Ah, @keisuke-umezawa, could you merge this the latest `master` to run the CI one last time?

HideakiImamura (approved):
LGTM!

PR #1572, opened 2020-07-30
Bump PyTorch to 1.6

crcrpar:
## Motivation
PyTorch 1.6 dropped Python 3.5 thus we need a bit of changes though AllenNLP does not support the latest PyTorch. So, I'll close this PR.


crcrpar:
Looks like everything works well with PyTorch 1.6.

setup.py
```diff
@@ -77,17 +77,19 @@ def get_extras_require() -> Dict[str, List[str]]:
...
-            "torch==1.5.1" if sys.platform == "darwin" else "torch==1.5.1+cpu",
-            "torchvision==0.6.1" if sys.platform == "darwin" else "torchvision==0.6.1+cpu",
             "xgboost",
             "keras",
             "tensorflow>=2.0.0",
             "tensorflow-datasets",
         ]
+        + (
+            ["torch==1.6.0" if sys.platform == "darwin" else "torch==1.6.0+cpu"]
+            + ["pytorch-ignite", "thop"]
```
himkt, line 89: [just a question]
Why are you adding `thop` only when the version of Python is greater than 3.5?
[`pytorch_simple.py`](https://github.com/optuna/optuna/blob/4a4d37a6645f841662e5fe57cafef03269b03a0c/examples/multi_objective/pytorch_simple.py#L13) doesn't work on Python3.5?

setup.py
```diff
@@ -81,17 +81,19 @@ def get_extras_require() -> Dict[str, List[str]]:
             "mpi4py",
             "mxnet",
             "nbval",
-            "pytorch-ignite",
             "scikit-image",
             "scikit-learn>=0.19.0,<0.23.0",  # optuna/visualization/param_importances.py.
-            "thop",
-            "torch==1.5.1" if sys.platform == "darwin" else "torch==1.5.1+cpu",
-            "torchvision==0.6.1" if sys.platform == "darwin" else "torchvision==0.6.1+cpu",
```
hvy, line 89: Oh, we don't need `torchvision`?

hvy:
Sorry but could you address the conflicts?

setup.py
hvy: Also curious about this change. Do you have any comments on it @crcrpar?

crcrpar: Because `thop` obviously depends on PyTorch https://github.com/Lyken17/pytorch-OpCounter/blob/master/setup.py#L14.

setup.py
```diff
@@ -81,17 +81,19 @@ def get_extras_require() -> Dict[str, List[str]]:
...
             "keras",
             "tensorflow>=2.0.0",
             "tensorflow-datasets",
         ]
+        + (
+            ["torch==1.6.0" if sys.platform == "darwin" else "torch==1.6.0+cpu"]
+            + ["pytorch-ignite", "thop"]
+            if (3, 5) < sys.version_info[:2]
+            else []
+        )
```
crcrpar, line 96:
```suggestion
        + (
            (
                ["torch==1.6.0", "torchvision==0.7.0"]
                if sys.platform == "darwin"
                else ["torch==1.6.0+cpu", "torchvision==0.7.0+cpu"]
            )
            + ["pytorch-ignite", "thop"]
            if (3, 5) < sys.version_info[:2]
            else []
        )
```

setup.py
himkt: Thank you for the explanation.

1. `thop` depends on `PyTorch` explicitly
2. `PyTorch` dropped the support of Python 3.5
3. If `thop` appears at the top of `setup.py`, CIs using Python 3.5 will fail (so we have to conditionally add into dependencies)

It makes sense for me. 👍

setup.py
hvy: The suggestion LGTM. I assume this addresses https://github.com/optuna/optuna/pull/1572#discussion_r464740767 but I haven't seen suggestions from PR authors before. Is this one waiting for feedback or a commit from a reviwer?

hvy:
I changes looked good first but I think we have to ignore the PyTorch examples from our daily runs in https://github.com/optuna/optuna/blob/master/.github/workflows/examples.yml#L51 ?

PR #1574, opened 2020-07-30
Improve the search space of XGBoost example

c-bata:
## Motivation

I made XGBoost example more practical. In this example, accuracy is improved like the following:

* Before: accuracy=0.9790209790209791
* After: accuracy=0.993006993006993

## Description of the changes

I referred [here](https://www.analyticsvidhya.com/blog/2016/03/complete-guide-parameter-tuning-xgboost-with-codes-python/), [here](https://www.youtube.com/watch?v=VC8Jc9_lNoY) and Japanese Kaggle book to learn best practices to optimize XGBoost parameters.

### DIFF
--- a/examples/xgboost_simple.py
+++ b/examples/xgboost_simple.py
@@ -38,17 +38,20 @@ def objective(trial):
 
     param = {
         "silent": 1,
+        "seed": 1,
         "objective": "binary:logistic",
-        "booster": trial.suggest_categorical("booster", ["gbtree", "gblinear", "dart"]),
+        "booster": trial.suggest_categorical("booster", ["gbtree", "dart"]),
+        "min_child_weight": trial.suggest_loguniform("min_child_weight", 1e-3, 1e3),
+        "subsample": trial.suggest_uniform("subsample", 0.5, 1.0),
+        "colsample_bytree": trial.suggest_uniform("colsample_bytree", 0.5, 1.0),
         "lambda": trial.suggest_float("lambda", 1e-8, 1.0, log=True),
         "alpha": trial.suggest_float("alpha", 1e-8, 1.0, log=True),
+        "max_depth": trial.suggest_int("max_depth", 3, 10),
+        "eta": trial.suggest_float("eta", 0.01, 0.2, log=True),
+        "gamma": trial.suggest_float("gamma", 1e-8, 1.0, log=True),
+        "grow_policy": trial.suggest_categorical("grow_policy", ["depthwise", "lossguide"]),
     }
 
-    if param["booster"] == "gbtree" or param["booster"] == "dart":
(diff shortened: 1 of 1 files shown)


hvy:
Thanks for the interesting PR. 👍 

I couldn't reproduce the same degree of improvement (`xgboost==1.0.2`), unfixing the seed, but your suggestion still seems better.

**Mean over 20 runs (unfixing the seed)**

| `master` | This PR | 
| - | - |
| 0.9842657342657345 | 0.9895104895104897

examples/xgboost_simple.py
```diff
@@ -38,17 +38,20 @@ def objective(trial):
 
     param = {
         "silent": 1,
+        "seed": 1,
         "objective": "binary:logistic",
-        "booster": trial.suggest_categorical("booster", ["gbtree", "gblinear", "dart"]),
+        "booster": trial.suggest_categorical("booster", ["gbtree", "dart"]),
+        "min_child_weight": trial.suggest_loguniform("min_child_weight", 1e-3, 1e3),
```
hvy, line 44: `1e3` seems quite large, even though it's in log scale. It usually ends up at around `0.3` when I run it locally and the [breast cancer dataset only contains around ~500 samples](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_breast_cancer.html). Is this number also based on your references?

HideakiImamura: I agree. In [reference,](https://www.analyticsvidhya.com/blog/2016/03/complete-guide-parameter-tuning-xgboost-with-codes-python/), it is tuned as an integer and ends up to `6`. The range of the tuning seems to be `{1, 3, 5} -> {4, 5, 6} -> {6, 8, 10, 12}` (some stepwise tuning).

examples/xgboost_simple.py
```diff
@@ -38,17 +38,20 @@ def objective(trial):
...
+        "subsample": trial.suggest_uniform("subsample", 0.5, 1.0),
+        "colsample_bytree": trial.suggest_uniform("colsample_bytree", 0.5, 1.0),
         "lambda": trial.suggest_float("lambda", 1e-8, 1.0, log=True),
         "alpha": trial.suggest_float("alpha", 1e-8, 1.0, log=True),
+        "max_depth": trial.suggest_int("max_depth", 3, 10),
```
HideakiImamura, line 49: In [reference](https://www.analyticsvidhya.com/blog/2016/03/complete-guide-parameter-tuning-xgboost-with-codes-python/), it is tuned in `{3, 5, 7, 9}`. How about setting `step = 2`?
```suggestion
        "max_depth": trial.suggest_int("max_depth", 3, 10, step=2),
```

PR #1575, opened 2020-07-30
Test `_CachedStorage` in `test_study.py`.

ytsmiling:
## Motivation
`_CachedStorage` class was not used in the `test_study.py`. Since the `test_study.py` serves as a kind of integration test, adding the storage into the `test_study.py` will increase the ability to catch bugs in the test.

## Description of the changes
Test `_CachedStorage` in `test_study.py`.

### DIFF
--- a/tests/test_study.py
+++ b/tests/test_study.py
@@ -33,6 +33,7 @@ if type_checking.TYPE_CHECKING:
 STORAGE_MODES = [
     "inmemory",
     "sqlite",
+    "cache",
     "redis",
 ]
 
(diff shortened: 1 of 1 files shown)


tests/test_study.py
```diff
@@ -29,10 +29,10 @@
 
     CallbackFuncType = Callable[[optuna.study.Study, optuna.trial.FrozenTrial], None]
 
-# TODO(ytsmiling) Add tests for multi-worker settings.
```
hvy, line 32: Just for the record, the TODO was addressed by https://github.com/optuna/optuna/blob/master/tests/storages_tests/rdb_tests/test_with_server.py.

ytsmiling:
I happened to notice that while creating #1571.  The `test_study.py` was useful to debug #1571, but I encountered no bugs by this change in the current master.

PR #1585, opened 2020-08-03
Skip entire `allennlp` example directory in CI

hvy:
## Motivation

Fixes CI example failure caused by unintentionally executed allennlp example. C.f. https://github.com/optuna/optuna/runs/937729336?check_suite_focus=true

## Description of the changes

Skips entire `allenlp` directory, in case additional files would be added.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -37,9 +37,9 @@ jobs:
       run: |
         # TODO(hvy): Stop ignoring allennlp_jsonnet.py when https://github.com/optuna/optuna/issues/1526 is resolved.
         if [ ${{ matrix.python-version }} = 3.5 ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/examples.yml
```diff
@@ -37,9 +37,9 @@ jobs:
       run: |
         # TODO(hvy): Stop ignoring allennlp_jsonnet.py when https://github.com/optuna/optuna/issues/1526 is resolved.
         if [ ${{ matrix.python-version }} = 3.5 ]; then
-          IGNORES='chainermn_.*|pytorch_lightning.*|fastai_.*|allennlp_.*|catalyst_.*|sb3_.*|allennlp_jsonnet.py'
+          IGNORES='chainermn_.*|pytorch_lightning.*|fastai_.*|allennlp|catalyst_.*|sb3_.*|allennlp_jsonnet.py'
```
himkt, line 40: [just a question]

We can remove `allennlp_jsonnet.py` instead? (same for L42)

hvy: Fixed!

PR #1594, opened 2020-08-05
Use setup python v2

crcrpar:


HideakiImamura:
Thanks for the PR. What does this PR intend? Is it a same purpose with #1595?

crcrpar:
Not that much, main intention is it's better to use up-to-date things.

PR #1595, opened 2020-08-05
[DEPENDS ON #1594] Add `cache` to GitHub Actions Workflows

crcrpar:
## Motivation
Though with the cache we have not seen any great performance gain, I send this hoping the cache will reduce the pip installation failures that we've seen last couple weeks.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -19,10 +19,21 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install
+    - name: Setup cache
+      uses: actions/cache@v2
+      env:
+        cache-name: daily-example
+      with:
+        path: /opt/**/site-packages
+        key: ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}
+        restore-keys: |
+          ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-
+    - name: Install (apt)
       run: |
         sudo apt-get update
         sudo apt-get -y install openmpi-bin libopenmpi-dev
+    - name: Install (Python)
+      run: |
         python -m pip install --upgrade pip
         pip install --progress-bar off -U setuptools
         python setup.py sdist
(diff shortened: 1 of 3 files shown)


crcrpar:
As `cache` would skip the download of pip installation which failed the last couple weeks.

.github/workflows/examples.yml
```diff
@@ -19,10 +19,21 @@ jobs:
...
-    - name: Install
+    - name: Setup cache
+      uses: actions/cache@v2
+      env:
+        cache-name: daily-example
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}-v1
+        restore-keys: |
+          ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ env.cache-name }}
```
hvy, line 30: Suppose you meant the following.
```suggestion
          ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}
```

.github/workflows/examples.yml
```diff
@@ -19,10 +19,21 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install
+    - name: Setup cache
+      uses: actions/cache@v2
+      env:
+        cache-name: daily-example
```
hvy, line 25: What happens if we don't distinguish by workflow, does it still work?

Assuming we have to, a bit curious whether you could use some of the [default envs](https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables), such as `GITHUB_ACTION`. Not a request but wanted to know if you've tried.

.github/workflows/tests.yml
```diff
@@ -37,6 +37,15 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Setup Cache
```
hvy, line 40: One huge nit, for consistency with your other changes.
```suggestion
    - name: Setup cache
```

.github/workflows/examples.yml
hvy: Ah, although `setup.py` is identical, different libraries are installed for different workflows.

PR #1603, opened 2020-08-06
Make TPE categorical sampling faster

c-bata:
## Motivation

The cardinality of categorical parameters tends to be lower than `n_ei_candidates` (default: 24).

## Description of the changes

Use `np.arange(len(choices))` instead of sampling from `l(x)` distribution.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -344,13 +344,16 @@ class TPESampler(BaseSampler):
         below = list(map(int, below))
         above = list(map(int, above))
         upper = len(choices)
-        size = (self._n_ei_candidates,)
 
         weights_below = self._weights(len(below))
         counts_below = np.bincount(below, minlength=upper, weights=weights_below)
         weighted_below = counts_below + self._prior_weight
         weighted_below /= weighted_below.sum()
-        samples_below = self._sample_from_categorical_dist(weighted_below, size)
+        if self._n_ei_candidates >= upper:
+            samples_below = np.arange(upper)
+        else:
+            size = (self._n_ei_candidates,)
+            samples_below = self._sample_from_categorical_dist(weighted_below, size)
         log_likelihoods_below = TPESampler._categorical_log_pdf(samples_below, weighted_below)
 
         weights_above = self._weights(len(above))


optuna/samplers/_tpe/sampler.py
```diff
@@ -344,13 +344,16 @@ def _sample_categorical_index(self, distribution, below, above):
...
         upper = len(choices)
-        size = (self._n_ei_candidates,)
 
         weights_below = self._weights(len(below))
         counts_below = np.bincount(below, minlength=upper, weights=weights_below)
         weighted_below = counts_below + self._prior_weight
         weighted_below /= weighted_below.sum()
-        samples_below = self._sample_from_categorical_dist(weighted_below, size)
+        if self._n_ei_candidates >= upper:
+            samples_below = np.arange(upper)
```
hvy, line 353: I think this changes the logic to guarantees that all the choices will be included. Should confirm the performance with benchmarks, or was it covered in the Goptuna benchmarks?

HideakiImamura:
Hi all. I conducted a benchmark experiment. Please take a look.

## Setup
- optuna: current master v.s. this PR (https://github.com/optuna/optuna/pull/1603/commits/75c422103ae24f407635af24fd43e35cf2113863)
- (sampler, pruner): (`TPESampler`, `NopPruner`)
- dataset: ("HPO-Bench-Naval", "HPO-Bench-Parkinson", "HPO-Bench-Protein", "HPO-Bench-Slice", "NAS-Bench") in `kurobako`
- \# of runs: 100 studies for each experiment. 80 budgets (trials) for HPO-Bench-* and 120 budgets for NAS-Bench.

## Results
hpo-bench-naval-77d6558d0ea51787e1ad60be3e50c393a3e370ad812d5e1fdb2f5ca5c69a644d
hpo-bench-parkinson-b25702872d95f2a0d9b2d38f9f90e5950cdc0efa2da7c2954294ee7b23efc278
hpo-bench-protein-bba52364e4b4ec03003330070d0755e637eabb669ee5864ed4eec521bee215c6
hpo-bench-slice-88a2191f9b676e7878d93c8905ffde539488937a81829a0297ff7ef833fa76ae
nasbench-a-9c7ea8bffeac222495d24726b4db8f790031db67987031175788d150893e53bb

## Discussion
According to the above experiments, the new TPE performs almost samely as the current `master`'s one for HPO-Bench. However, it performs relatively bad for NAS-Bench. The HPO-Bench includes 6 real-valued parameters and 2 categorical parameters (See the[ original paper](https://arxiv.org/abs/1905.04970) for the detail). The NAS-Bench includes only categorical parameters (See the [original paper](https://arxiv.org/abs/1902.09635) for the detail). The new TPE degrades the performance of the categorical sampling of TPE. Therefore, I'm leaning towards not merging this PR because it improves speed but worsens performance. 

From the theoretical point of view, this PR's categorical sampling always "exploits". In the original (current master) TPE takes the trade-off between exploration and exploitation by probabilistically determining the `below` and `above`. Even if the estimated ratio `l(x)/g(x)` is poor, such `x` can be explored with a small probability.

I'm not sure but we can set the option flag to use this speed up for TPESampler if the number of categorical parameters is small.

c-bata:
@HideakiImamura Thank you for checking the benchmark!
Hmm... May I have the repeat count? It looks incomprehensible result from theoretical point of view.

HideakiImamura:
Thanks for the quick reply!

The number of repetitions of the experiment is 100. I took the mean and variance over all experiments.

Yes, you are right. If `n_ei_candidates` goes to infinity, the set chosen by `_sample_from_categorical_dist` converges to the "full" set. I'm not sure this is intended but the `n_ei_candidates` control this probabilistically chosen set. From the name of `n_ei_candidates`, we usually think this number control how we approximate EI maximization. But, this seems wrong for the categorical sampling. I'm also bit confused. I think the number of sampling for categorical sampling should be considered separately from `n_ei_candidates` for real-value sampling.

c-bata:
I see. Thank you for your deep investigation! `100` sounds enough repeat counts.

> I'm not sure this is intended but the n_ei_candidates control this probabilistically chosen set.

Yeah, I agree with your thoughts. From this result, I cannot find any other reasons.

I close this pull request. Thanks again!

HideakiImamura:
Thanks for all discussions @c-bata! I think we should name `n_ei_candidates` for the real-value sampling and the categorical sampling differently. What do you think?

c-bata:
Good suggestion. Code comment for these discussion is also useful.

c-bata:
I opened https://github.com/optuna/optuna/pull/1637.

I couldn't find more suitable name than `n_ei_candidates`. So I put a code comment for now.

PR #1628, opened 2020-08-12
Use context manager to treat session correctly.

keisuke-umezawa:
## Motivation

Currently, the code lines related to session creation, commit and rollback is written elsewhere. Moreover, they are not consistent because some of them will catch errors, and others will not, etc. 

In order to
1. Use the same code for session creation, commit and rollback, and
2. Treat sessions in the same manner, 

I used a contest manager.

## Description of the changes

I defined `_session_scope`, and use it when creating a new session.

```
@contextmanager
def _session_scope(self) -> Generator[orm.Session, None, None]:
...
```


keisuke-umezawa:
I need to check an error in `tests-rdbstorage`. Maybe, due to the difference between `_commit_with_integrity_check` and `_commit`? 🤔

keisuke-umezawa:
@hvy Sorry for being late. All tests passed. Could you review it?

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -140,16 +141,15 @@ def test_pickle_storage() -> None:
...
 
     storage = create_test_storage()
-    session = storage.scoped_session()
 
     # This object violates the unique constraint of version_info_id.
     v = VersionInfoModel(version_info_id=1, schema_version=1, library_version="0.0.1")
-    session.add(v)
-    with pytest.raises(StorageInternalError):
-        storage._commit(session)
+    with pytest.raises(IntegrityError):
```
keisuke-umezawa, line 150: Due to the change of error handling in `_session_scope`, this flow is changed to throw `IntegrityError`.

keisuke-umezawa:
@hvy Sorry for the late notification. I finished it.

optuna/storages/_rdb/storage.py
```diff
@@ -39,6 +41,43 @@
...
+        return session
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[Exception]],
+        exc_value: Optional[Exception],
+        traceback: Optional[TracebackType],
+    ) -> None:
+        try:
+            session = self._scoped_session()
```
hvy, line 59: The `session` object here may or may not refer to the same object as the `session` returned from `__enter__`. E.g. if the caller releases associated connections (by calling `session.remove()`) before entering `__exit__`, the object will refer to a completely new `Session` instance here. Can we somehow make sure to use the same object here? Would holding the `session` as a member attribute work? We might want to instantiate a new `_ScopedSessionContextManager` object in each `with` in that case, instead of holding a single instance as a member.

hvy: C.f. https://docs.sqlalchemy.org/en/13/orm/contextual.html. By the ways, it is good to know that the session factory is thread safe, in that it creates thread local sessions.

optuna/storages/_rdb/storage.py
```diff
@@ -39,6 +41,43 @@
...
+        session = self._scoped_session()
+        return session
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[Exception]],
+        exc_value: Optional[Exception],
+        traceback: Optional[TracebackType],
+    ) -> None:
+        try:
```
hvy, line 58: If my understanding is correct, `__exit__` is called as soon as an exception is raised and propagated from within the context. In this case, we shouldn't try to commit but just roll back and then propagate the error further, i.e. return `None` here. How about adding a check in the beginning of this method and doing so?

hvy (changes requested):
Thanks @keisuke-umezawa for the PR, I left some early comments regarding the context manager.

optuna/storages/_rdb/storage.py
keisuke-umezawa: > Would holding the session as a member attribute work?
For the first step, I tried it and did not work, because sometimes the member attribute is created by one thread, and is accessed from the other thread.
https://github.com/optuna/optuna/pull/1628/commits/c06a25c2bcb174ca4d4688c8827c29323ff9523d#diff-9aa9e205e3a2038ae9077f72d10abac1R44-R77

keisuke-umezawa: > We might want to instantiate a new _ScopedSessionContextManager object in each with in that case, instead of holding a single instance as a member.

That is true. I will work on it.

keisuke-umezawa: To avoid such a situation, I used `@contextmanager` and input `scoped_session` as an argument.

https://github.com/optuna/optuna/pull/1628/commits/3f26b36ea4ee3099ef65a5ad9a54f55c96449490

optuna/storages/_rdb/storage.py
keisuke-umezawa: In the latest commit, I used `@contextmanager` instead of class.

hvy:
Could you check the CI failures?

optuna/storages/_rdb/storage.py
```diff
@@ -39,6 +41,30 @@
...
+        raise
+    except SQLAlchemyError as e:
+        session.rollback()
+        message = (
+            "An exception is raised during the commit. "
+            "This typically happens due to invalid data in the commit, "
+            "e.g. exceeding max length. "
+            "(The actual exception is as follows: {})".format(repr(e))
+        )
+        raise optuna.exceptions.StorageInternalError(message).with_traceback(sys.exc_info()[2])
```
hvy, line 65: How about explicitly catching and handling all other types of exceptions? I'm not so sure whether the indentations are correct, though.
```suggestion
        raise optuna.exceptions.StorageInternalError(message).with_traceback(sys.exc_info()[2])
    except:
        session.rollback()
        raise
    finally:
        session.close()
```

optuna/storages/_rdb/storage.py
```diff
@@ -168,12 +194,9 @@ def create_new_study(self, study_name: Optional[str] = None) -> int:
...
 
-        session = self.scoped_session()
-
-        study = models.StudyModel.find_or_raise_by_id(study_id, session)
-        session.delete(study)
-
-        self._commit_with_integrity_check(session)
+        with _scoped_session(self.scoped_session) as session:
+            study = models.StudyModel.find_or_raise_by_id(study_id, session)
+            session.delete(study)
```
hvy, line 199: We may need to try and catch/silence the `IntegrityError` to keep the behavior consistent.

optuna/storages/_rdb/storage.py
```diff
@@ -39,6 +41,30 @@
...
+    try:
+        yield session
+        session.commit()
+    except IntegrityError as e:
+        _logger.debug(
+            "Ignoring {}. This happens due to a timing issue among threads/processes/nodes. "
+            "Another one might have committed a record with the same key(s).".format(repr(e))
+        )
+        session.rollback()
+        raise
```
hvy, line 56: We're not "ignoring" the `IntegrityError` any longer here but instead propagating it. We have to handle this outside the caller of the `_scoped_session` with `try/catch` or introduce a parameter to this context manager whether it should be silenced or not. I personally prefer the former as the latter feel a bit ad-hoc. What do you think, are there perhaps better alternatives?

optuna/storages/_rdb/storage.py
hvy: Ditto for most occasions where we used to call `self._commit_with_integrity_check`.

optuna/storages/_rdb/storage.py
```diff
@@ -449,20 +441,16 @@ def _create_new_trial(
...
+                    #
+                    # Locking within a study is necessary since the creation of a trial is not an
+                    # atomic operation. More precisely, the trial number computed in
+                    # `_get_prepared_new_trial` is prone to race conditions without this lock.
+                    models.StudyModel.find_or_raise_by_id(study_id, session, for_update=True)
+
+                    trial = self._get_prepared_new_trial(study_id, template_trial, session)
                 break  # Successfully created trial.
             except OperationalError:
                 session.rollback()
```
hvy, line 456: We can omit this clean-up as it's already handled by the context manager.

optuna/storages/_rdb/storage.py
```diff
@@ -590,135 +578,136 @@ def _update_trial(
...
-            session.rollback()
-            return False
+        with _scoped_session(self.scoped_session) as session:
+            trial_model = (
+                session.query(models.TrialModel)
+                .filter(models.TrialModel.trial_id == trial_id)
+                .one_or_none()
+            )
+            if trial_model is None:
+                session.rollback()
```
hvy, line 588: Ditto, redundant clean-up.
```suggestion
```

optuna/storages/_rdb/storage.py
```diff
@@ -590,135 +578,136 @@ def _update_trial(
...
+                raise KeyError(models.NOT_FOUND_MSG)
+            if trial_model.state.is_finished():
+                session.rollback()
```
hvy, line 591: Ditto.
```suggestion
```

optuna/storages/_rdb/storage.py
```diff
@@ -590,135 +578,136 @@ def _update_trial(
...
+                raise RuntimeError("Cannot change attributes of finished trial.")
+            if (
+                state
+                and trial_model.state != state
+                and state == TrialState.RUNNING
+                and trial_model.state != TrialState.WAITING
+            ):
+                session.rollback()
```
hvy, line 599: Not sure but I think adding a handler for `finally` in the context manager should take care of this?
```suggestion
```

optuna/storages/_rdb/storage.py
```diff
@@ -590,135 +578,136 @@ def _update_trial(
...
-
-        self.check_trial_is_updatable(trial_id, trial.state)
-
-        if state == TrialState.RUNNING and trial.state != TrialState.WAITING:
-            session.rollback()
+        try:
+            with _scoped_session(self.scoped_session) as session:
+                trial = models.TrialModel.find_by_id(trial_id, session, for_update=True)
+                if trial is None:
+                    session.rollback()
```
hvy, line 696: Ditto.
```suggestion
```

optuna/storages/_rdb/storage.py
```diff
@@ -590,135 +578,136 @@ def _update_trial(
...
+                    raise KeyError(models.NOT_FOUND_MSG)
+
+                self.check_trial_is_updatable(trial_id, trial.state)
+
+                if state == TrialState.RUNNING and trial.state != TrialState.WAITING:
+                    session.rollback()
```
hvy, line 702: Ditto?
```suggestion
```

hvy (changes requested):

optuna/storages/_rdb/storage.py
keisuke-umezawa: Instead of close, I used remove. https://docs.sqlalchemy.org/en/13/orm/contextual.html#using-thread-local-scope-with-web-applications

keisuke-umezawa: When I add finally section, the tests fail. I will check the reason.

optuna/storages/_rdb/storage.py
keisuke-umezawa: I chose the latter one, because all algorithm related to session management can be defined by `_scoped_session`.
In my perspective, other people can understand the logic only checking the definition of `_scoped_session`.
https://github.com/optuna/optuna/pull/1628/commits/539d061df32554a29a22de63885605bb75d0df87

If I need to follow backward compatibility strictly, I need to ignore IntegrityError only for `self._commit_with_integrity_check`. What do you think of it?

keisuke-umezawa: > If I need to follow backward compatibility strictly, I need to ignore IntegrityError only for self._commit_with_integrity_check. What do you think of it?

To follow it strictly, I committed it. https://github.com/optuna/optuna/pull/1628/commits/5166bec9c1f1a4f9ef2f2ded142fb084edd38165

optuna/storages/_rdb/storage.py
keisuke-umezawa: I found the reason. If we add `close` here, the code try to access the DB object after closing session.

https://github.com/optuna/optuna/pull/1628/files#diff-6048e6b64855f2df6bcd963dee2c988e671a2971a5aad48ef5e88c5f6d49b3adL1262-L1272

The current code is not compatible with `close`, but if I will modify everything, I can add `close` here.
I think it should be done in another PR.

optuna/storages/_rdb/storage.py
keisuke-umezawa: Since there are no insert or update before this line, I think I can remove this line.

optuna/storages/_rdb/storage.py
keisuke-umezawa: Since there are no insert or update before this line, I think I can remove this line.

optuna/storages/_rdb/storage.py
hvy: Sorry I don't quite understand what you mean by modifying everything but do you mind explaining what sorts of changes would be required? Does it fail when trying to close after a rollback or commit?

optuna/storages/_rdb/storage.py
```diff
@@ -146,14 +188,14 @@ def __setstate__(self, state: Dict[Any, Any]) -> None:
...
+            with _scoped_session(self.scoped_session) as session:
+                if study_name is None:
+                    study_name = self._create_unique_study_name(session)
 
-        study = models.StudyModel(study_name=study_name, direction=StudyDirection.NOT_SET)
-        session.add(study)
-        if not self._commit_with_integrity_check(session):
+                study = models.StudyModel(study_name=study_name, direction=StudyDirection.NOT_SET)
+                session.add(study)
+        except StorageInternalError:
```
hvy, line 198: Should it be `IntegrityError` to keep the behavior unchanged?
```suggestion
        except IntegrityError:
```

optuna/storages/_rdb/storage.py
keisuke-umezawa: In the current implementation, I enclosed code from `session = self.scoped_session()` to `self._commit(session)`  by `with` statement. But some lines access db values after `self._commit(session)` e.g. https://github.com/optuna/optuna/pull/1628/files#diff-6048e6b64855f2df6bcd963dee2c988e671a2971a5aad48ef5e88c5f6d49b3adR274 .
If I add `session.close()` here, we will have errors.

After face2face discussion with you, we concluded that we will solve this problem in another PR.

optuna/storages/_rdb/storage.py
```diff
@@ -38,6 +42,44 @@
...
+                "Another one might have committed a record with the same key(s).".format(repr(e))
+            )
+        else:
+            message = (
+                "An exception is raised during the commit. "
+                "This typically happens due to invalid data in the commit, "
+                "e.g. exceeding max length. "
+                "(The actual exception is as follows: {})".format(repr(e))
+            )
+            raise optuna.exceptions.StorageInternalError(message).with_traceback(sys.exc_info()[2])
```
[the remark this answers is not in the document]
keisuke-umezawa: Fixed.

optuna/storages/_rdb/storage.py
keisuke-umezawa: I agree with you. If I throw `IntegrityError` in the context manager and catch it here, we can keep backward compatibility here. 

I modified the code as 
https://github.com/optuna/optuna/pull/1628/commits/31ac6bfc9c2503f1f703b8c4a40b7330c7ee2c8a .

One concern is that we may throw `IntegrityError` instead of `StorageInternalError` where `_commit` was used. So, users will receive  `IntegrityError` instead of `StorageInternalError`, but it should be minor problem. (Basically, there are no error handlings for both `StorageInternalError` and `OptunaError`, so users will receive errors directly.)

hvy (approved):
LGTM!

I would perhaps leave some TODO to simplify the `_scoped_session` for future readers since the logic of `ignore_integrity_error` is somewhat complicated and difficult to understand without git blame. Also perhaps a TODO to try adding a finally with a session close (by fixing the caller side to extract the necessary information from the models before exiting the context).

HideakiImamura (commented):
Thank you for the great PR. Basically, LGTM. I have 2 minor comments. PTAL.

keisuke-umezawa:
@HideakiImamura Thank you for reviewing! I followed your suggestions.

HideakiImamura (approved):
LGTM! Thanks for working on the long-running PR. Let me merge this PR.

PR #1630, opened 2020-08-13
refactor TPE's `sample_from_categorical_dist`

crcrpar:
## Motivation

Some of the code doesn't look cool to me so I changed it but I'm not sure these changes do __NOT__ break anything.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -470,20 +470,18 @@ class TPESampler(BaseSampler):
     def _sample_from_categorical_dist(self, probabilities, size):
         # type: (np.ndarray, Tuple[int]) -> np.ndarray
 
-        if probabilities.size == 1 and isinstance(probabilities[0], np.ndarray):
-            probabilities = probabilities[0]
-        probabilities = np.asarray(probabilities)
-
         if size == (0,):
             return np.asarray([], dtype=float)
         assert len(size)
+
+        if probabilities.size == 1 and isinstance(probabilities[0], np.ndarray):
+            probabilities = probabilities[0]
         assert probabilities.ndim == 1
 
-        n_draws = int(np.prod(size))
-        sample = self._rng.multinomial(n=1, pvals=probabilities, size=int(n_draws))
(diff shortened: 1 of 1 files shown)


optuna/samplers/_tpe/sampler.py
```diff
@@ -470,20 +470,18 @@ def _gmm_log_pdf(
     def _sample_from_categorical_dist(self, probabilities, size):
         # type: (np.ndarray, Tuple[int]) -> np.ndarray
 
-        if probabilities.size == 1 and isinstance(probabilities[0], np.ndarray):
-            probabilities = probabilities[0]
-        probabilities = np.asarray(probabilities)
```
HideakiImamura, line 475: Thanks for the good catch! I agree that this conversion of type is not necessary.

optuna/samplers/_tpe/sampler.py
```diff
@@ -470,20 +470,18 @@ def _gmm_log_pdf(
...
+        if probabilities.size == 1 and isinstance(probabilities[0], np.ndarray):
+            probabilities = probabilities[0]
         assert probabilities.ndim == 1
 
-        n_draws = int(np.prod(size))
-        sample = self._rng.multinomial(n=1, pvals=probabilities, size=int(n_draws))
-        assert sample.shape == size + (probabilities.size,)
-        return_val = np.dot(sample, np.arange(probabilities.size))
-        return_val.shape = size
+        n_draws = np.prod(size).item()
```
HideakiImamura, line 481: Does this line convert `float` to `int`? Why do you add the `item()` method?

crcrpar: Why "convert `float` to `int`"?

The return type of `np.prod` is `np.int64` as `size` is expected to be a tuple of integers. Thus I think `item` is enough. Furthermore, previously we applied `int()` to the result of `np.prod(size)` twice.

HideakiImamura: Thanks for the reply. I understood. It actually converts `np.int64` to `int`.

PR #1631, opened 2020-08-13
Fix error messages in `TPEsampler`

kstoneriv3:
## Description of the changes

Fixed the error messages so that it matches the error conditions.


hvy:
This is somewhat of a nit but we usually wrap code (in PR titles) with backticks https://github.com/optuna/optuna/blob/master/CONTRIBUTING.md#title, in this case the "TPESampler" part.

toshihikoyanase:
I'll merge this PR since it got two approvals. 

@kstoneriv3 Let me slightly modify the PR title according to the @hvy's comment.

PR #1633, opened 2020-08-13
Add `.dockerignore`.

harupy:
## Motivation

We don't have to copy everything in a docker image.

## Description of the changes

Add `.dockerignore`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,9 +1,9 @@
 # Ignore everything
 **
 
-!/setup.py
-!/README.md
-!/optuna
-!/examples
+!setup.py
+!README.md
+!optuna
+!examples
 
 **/__pycache__
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+# Ignore everything
+**
+
+!/setup.py
+!/README.md
+!/optuna
+!/examples
+
+**/__pycache__
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -9,7 +9,7 @@ on:
   pull_request:
     paths:
       - Dockerfile
-      - .dockerignoree
+      - .dockerignore
       - setup.py
 
 env:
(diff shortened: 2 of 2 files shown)


.github/workflows/dockerimage.yml
```diff
@@ -9,6 +9,7 @@ on:
   pull_request:
     paths:
       - Dockerfile
+      - .dockerignoree
```
harupy, line 12: Looks like adding this doesn't trigger the docker-image action.

harupy: We could do:

- Commit this change and an empty `.dockerignore`.
- Create a new PR and edit `.dockerignore` in that PR.

Whta do you think?

harupy: It was just a typo :)

.dockerignore
```diff
@@ -0,0 +1,9 @@
+# Ignore everything
+**
+
+!setup.py
+!README.md
+!optuna
+!examples
```
hvy, line 7: How about whitelisting unit tests and docs related directories as well? I believe the provided Dockerfile is used by developers as well as users.

harupy: I think we would mount the project directory to the container working directory during development.

harupy:
```
# fix optuna code
# fix test

# Run the test using the fixed code
docker run ... -v $(pwd):/workspaces <imane> pytest tests/...
```

harupy: This VSCode doc might be helpful: https://code.visualstudio.com/docs/remote/containers

harupy: I have excluded `docs` and `tests` because:
- We don't need them when running a python script that uses optuna.
- During development, we usually mount the project directory so the dev image doesn't need to contain them.

hvy: Thanks for the detailed explanation. Mounting a completely separate directory when developing makes sense.

PR #1635, opened 2020-08-14
Run example after docker build to ensure that built image is setup properly

harupy:
## Motivation

Just because the docker build succeeds doesn't mean the built image is setup properly.

## Description of the changes

Run `sklearn_simple.py` after the docker build to prevent PRs like #1633 from breaking the docker images.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -10,6 +10,7 @@ on:
     paths:
       - Dockerfile
       - setup.py
+      - dockerimage.yml
 
 env:
   DOCKER_HUB_BASE_NAME: optuna/optuna
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -47,9 +47,9 @@ jobs:
       run: |
         docker build . --build-arg PYTHON_VERSION=${{ matrix.python_version }} --build-arg BUILD_TYPE=${{ matrix.build_type }} --file Dockerfile --tag "${HUB_TAG}"
 
-    - name: Run an example using the built image
+    - name: Verify the built image
       run: |
-        docker run "${HUB_TAG}" python examples/sklearn_simple.py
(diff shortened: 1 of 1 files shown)


.github/workflows/dockerimage.yml
```diff
@@ -46,6 +47,10 @@ jobs:
       run: |
         docker build . --build-arg PYTHON_VERSION=${{ matrix.python_version }} --build-arg BUILD_TYPE=${{ matrix.build_type }} --file Dockerfile --tag "${HUB_TAG}"
 
+    - name: Run an example using the built image
+      run: |
+        docker run "${HUB_TAG}" python examples/sklearn_simple.py
```
crcrpar, line 52: How about checking whether libraries are correctly installed or not by checking `pip freeze` instead of running a concreate example as running example might be time consuming?

harupy: How about `tutorial/001_first.py` which should run quick?

toshihikoyanase: > How about tutorial/001_first.py which should run quick?

This approach seems reasonable to me. I think we may overlook runtime error if we only check with `pip freeze`.

harupy: Agree with @toshihikoyanase. @crcrpar What do you think?

crcrpar: Personally https://github.com/optuna/optuna/blob/master/examples/quadratic_simple.py looks better than `tutorial/001_first.py` :smiley:

harupy: @crcrpar `quadratic_simple.py` looks good.

toshihikoyanase (approved):
The [CI job](https://github.com/optuna/optuna/pull/1635/checks?check_run_id=1001469759) executed the example successfully. LGTM!

crcrpar (approved):
Thanks!

PR #1637, opened 2020-08-14
Add code comment about n_ei_candidates for categorical parameters.

c-bata:
## Motivation

See https://github.com/optuna/optuna/pull/1603#issuecomment-674001555.

@HideakiImamura suggests that naming `n_ei_candidates` to more suitable name. But I just put a code comment for now because I have no idea.

## Description of the changes

Add a code comment about the discussion of #1603.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -345,10 +345,11 @@ class TPESampler(BaseSampler):
         above = list(map(int, above))
         upper = len(choices)
 
-        # We can use np.arange(len(distribution.choices)) instead of sampling from l(x)
-        # when the cardinality of categorical parameter is lower than n_ei_candidates.
-        # Though it seems to be theoretically correct, it leads performance degration
-        # on NASBench. See https://github.com/optuna/optuna/pull/1603 for more details.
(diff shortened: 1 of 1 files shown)


optuna/samplers/_tpe/sampler.py
```diff
@@ -344,6 +344,11 @@ def _sample_categorical_index(self, distribution, below, above):
         below = list(map(int, below))
         above = list(map(int, above))
         upper = len(choices)
+
+        # We can use np.arange(len(distribution.choices)) instead of sampling from l(x)
+        # when the cardinality of categorical parameter is lower than n_ei_candidates.
+        # Though it seems to be theoretically correct, it leads performance degration
+        # on NASBench. See https://github.com/optuna/optuna/pull/1603 for more details.
```
HideakiImamura, line 351:
```suggestion
        # We can use `np.arange(len(distribution.choices))` instead of sampling from `l(x)`
        # when the cardinality of categorical parameters is lower than `n_ei_candidates`.
        # Though it seems to be theoretically correct, it leads to performance degradation
        # on the NAS benchmark experiment in https://arxiv.org/abs/1902.09635. 
        # See https://github.com/optuna/optuna/pull/1603 for more details.
```

PR #1639, opened 2020-08-14
Update dockerfile to prevent re-running the installation step on unrelated changes

harupy:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,13 @@ RUN apt-get update \
     && pip install ${PIP_OPTIONS} -U setuptools
 
 WORKDIR /workspaces
-COPY . .
+
+# Create (dummy) files required to run setup.py
+RUN mkdir optuna \
+    && touch README.md \
+    && echo '__version__ = "0.0.0"' > optuna/version.py
+
+COPY setup.py .
 
 ARG BUILD_TYPE='dev'
 
@@ -27,3 +33,5 @@ RUN if [ "${BUILD_TYPE}" = "dev" ]; then \
     && pip install ${PIP_OPTIONS} jupyter notebook
 
 ENV PIP_OPTIONS ""
+
+COPY . .
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -45,7 +45,7 @@ jobs:
     - name: Build the Docker image
       run: |
         docker build . --build-arg PYTHON_VERSION=${{ matrix.python_version }} --build-arg BUILD_TYPE=${{ matrix.build_type }} --file Dockerfile --tag "${HUB_TAG}"
-        docker run "${HUB_TAG}" python 'import optuna; print(optuna.__version__)'
+        docker run "${HUB_TAG}" python -c 'import optuna; print(optuna.__version__)'
 
     - name: Login & Push to Docker Hub
       if: ${{ github.event_name != 'pull_request' }}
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -11,7 +11,13 @@ RUN apt-get update \
...
 
 WORKDIR /workspaces
-COPY . .
+
+# Create (dummy) files required to run setup.py
+RUN mkdir optuna \
+    && touch README.md \
+    && echo '__version__ = "0.0.0"' > optuna/version.py
+
+COPY setup.py .
```
harupy, line 20: You can sonsider these files as a dummy optuna.

Dockerfile
```diff
@@ -27,3 +33,5 @@ RUN if [ "${BUILD_TYPE}" = "dev" ]; then \
     && pip install ${PIP_OPTIONS} jupyter notebook
 
 ENV PIP_OPTIONS ""
+
+COPY . .
```
harupy, line 37: In this step, the dummy optuna get replaced with the true optuna.

Dockerfile
harupy: This dummy optuna is installed in the *editable* mode so we can overwrite it later.

Dockerfile
harupy: This trick might have some side effects that I haven't noticed yet though

harupy: (This trick might have side effects that I haven't noticed yet though)

harupy: We can verify this trick works by running an example script after the image build.

Dockerfile
```diff
@@ -11,7 +11,13 @@ RUN apt-get update \
     && pip install ${PIP_OPTIONS} -U setuptools
 
 WORKDIR /workspaces
-COPY . .
+
+# Create (dummy) files required to run setup.py
+RUN mkdir optuna \
+    && touch README.md \
+    && echo '__version__ = "0.0.0"' > optuna/version.py
```
toshihikoyanase, line 18: This hack may affect application developers of Optuna.
`pip` recognized Optuna version as `0.0.0`, and it will be updated with the following `requirements.txt` unexpectedly.

```console
$ python -c 'import optuna; print(optuna.__version__)'
2.0.0

$ pip freeze | grep optuna
# Editable install with no version control (optuna==0.0.0)

$ echo "optuna>1.0.0" > requirements.txt
$ pip install -r requirements.txt
```

`master` branch
```console
Requirement already satisfied: optuna>1.0.0 in /workspaces (from -r requirements.txt (line 1)) (2.0.0)
...
```

PR 1639
```
...
Building wheels for collected packages: optuna
  Building wheel for optuna (setup.py) ... done
  Created wheel for optuna: filename=optuna-2.0.0-py3-none-any.whl size=312963 sha256=970776f84fbedcf5ba738587c7c206dd19468cf4743f2a89bbde2f965b183f39
  Stored in directory: /root/.cache/pip/wheels/22/8b/08/d32553e8cd416e1974ae704d41102b5a691c9612ad982b7991
Successfully built optuna
Installing collected packages: optuna
  Attempting uninstall: optuna
    Found existing installation: optuna 0.0.0
    Uninstalling optuna-0.0.0:
      Successfully uninstalled optuna-0.0.0
Successfully installed optuna-2.0.0
```

harupy: @toshihikoyanase Thanks for the investigation. Given these side effects, seems like I should close this issue 👍

PR #1655, opened 2020-08-19
Add type hints integration/_lightgbm_tuner

upura:
## Motivation
Please refer to the Issue #711 .

## Description of the changes
I added type hints into `integration/_lightgbm_tuner/__init__.py`, `integration/_lightgbm_tuner/alias.py`, and `integration/_lightgbm_tuner/sklearn.py`.

## Local test
I ran two tests according to the [article](https://tech.preferred.jp/ja/blog/optuna-wants-your-pull-requests/).
```
$ pytest tests
$ circleci build --job checks
```

### DIFF
--- a/optuna/integration/_lightgbm_tuner/sklearn.py
+++ b/optuna/integration/_lightgbm_tuner/sklearn.py
@@ -12,7 +12,7 @@ class LGBMModel(lgb.LGBMModel):
     See: `pydoc lightgbm.LGBMModel`
     """
 
-    def __init__(self, *args: List[Any], **kwargs: Dict[str, Any]) -> None:
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
 
         warnings.warn(
             "LightGBMTuner doesn't support sklearn API. "
@@ -27,7 +27,7 @@ class LGBMClassifier(lgb.LGBMClassifier):
     See: `pydoc lightgbm.LGBMClassifier`
     """
 
-    def __init__(self, *args: List[Any], **kwargs: Dict[str, Any]) -> None:
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
 
         warnings.warn(
             "LightGBMTuner doesn't support sklearn API. "
@@ -42,7 +42,7 @@ class LGBMRegressor(lgb.LGBMRegressor):
     See: `pydoc lightgbm.LGBMRegressor`
     """
 
-    def __init__(self, *args: List[Any], **kwargs: Dict[str, Any]) -> None:
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
 
         warnings.warn(
             "LightGBMTuner doesn't support sklearn API. "
--- a/optuna/integration/_lightgbm_tuner/sklearn.py
+++ b/optuna/integration/_lightgbm_tuner/sklearn.py
@@ -1,13 +1,9 @@
+from typing import Any  # NOQA
+from typing import Dict  # NOQA
+from typing import List  # NOQA
 import warnings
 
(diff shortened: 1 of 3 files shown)


crcrpar (approved):
Thank you for your nice work!

optuna/integration/_lightgbm_tuner/sklearn.py
```diff
@@ -1,23 +1,18 @@
...
 
 class LGBMModel(lgb.LGBMModel):
     """Proxy of lightgbm.LGBMModel.
 
     See: `pydoc lightgbm.LGBMModel`
     """
 
-    def __init__(self, *args, **kwargs):
-        # type: (List[Any], Dict[str, Any]) -> None
+    def __init__(self, *args: List[Any], **kwargs: Dict[str, Any]) -> None:
```
harupy, line 15: Should the type of `args` be `Tuple` instead of `List`?

```python
def f(*args, **kwargs):
    print(type(args), type(kwargs))


f(1, a="a")
```

outputs:

```
<class 'tuple'> <class 'dict'>
```

HideakiImamura: I agree with @harupy. How about changing the type annotation from `List[Any]` to `Tuple[Any]`?

HideakiImamura (approved):
Thank you for the PR! LGTM except for @harupy's comment!

optuna/integration/_lightgbm_tuner/sklearn.py
upura: In my understanding, type hints of `*args` can be a special type like `str` or `List`. After receiving them, the type of `args` is dedicated as `Tuple`, isn't it?
https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values

harupy: looks like we should use `Any` for both `args` and `kwargs`.

HideakiImamura: Thanks for providing useful information. According to the reference, 
> In the body of function foo, the type of variable args is deduced as Tuple[str, ...] and the type of variable kwds is Dict[str, int].

I'm not sure how `mypy` checks the type hinting in the function signature (in the signature or in the function body), but the current one does not occur any `mypy` errors. I think `List[Any]` in the function signature is OK because it is easy to understand and no error occurs in `mypy`'s check. What do you think?

harupy: @HideakiImamura Tested using `mypy` and vscode:

mypy-check

harupy: If we use `*args: List[Any]`, `LGBModel` only accepts `List[Any]` as positional arguments.

hvy: Since we're only forwarding the arguments to `lightgbm` without any assumptions whatsoever, `Any` for both `*args` and `**kwargs` as suggested by @harupy sounds robust and rather reasonable to me.

HideakiImamura: Thanks for the investigation. Actually, I agree that we should use `Any` instead of `List[Any]` here.

upura: Thank you! Really helpful. So, as a scope of this PR, is it OK to use `Any` for both `args` and `kwargs`?

HideakiImamura (approved):
Thank you for your swift actions! LGTM again!

Issue #1654, opened 2020-08-19
Please add supports to skorch


UTUnex:
Skorch is a wrapper of pytorch which makes it have a sklearn API format for easy use. Currently, it's a pity that I didn't find support of optuna for skorch. You know that there are still many users who just transfer from sklearn to neural network and still want to keep our familiar sklearn-style:)

Recently I have been searching for the ways to apply the optuna to my skorch program but I didn't find any solutions and examples and I have to turn to other tools which is very disappointing...

Anyway, please consider my request:)

HideakiImamura:
Thank you for the interesting suggestion! Do you have any skorch's code in which you want to use Optuna? It will help us how we integrate the skorch with Optuna.

UTUnex:
> Thank you for the interesting suggestion! Do you have any skorch's code in which you want to use Optuna? It will help us how we integrate the skorch with Optuna.

Thank you for your reply. The following is the code from skorch documenation:


# %%
import torch
from torch import nn
import torch.nn.functional as F

torch.manual_seed(0);

# %%
import numpy as np
from sklearn.datasets import make_classification

# %%
X, y = make_classification(1000, 20, n_informative=10, random_state=0)
X, y = X.astype(np.float32), y.astype(np.int64)

# %%
X.shape, y.shape, y.mean()

# %%
class ClassifierModule(nn.Module):
    def __init__(
            self,
            num_units=10,
            nonlin=F.relu,
            dropout=0.5,
    ):
        super(ClassifierModule, self).__init__()
        self.num_units = num_units
        self.nonlin = nonlin
        self.dropout = dropout

        self.dense0 = nn.Linear(20, num_units)
        self.nonlin = nonlin
        self.dropout = nn.Dropout(dropout)
        self.dense1 = nn.Linear(num_units, 10)
        self.output = nn.Linear(10, 2)

    def forward(self, X, **kwargs):
        X = self.nonlin(self.dense0(X))
        X = self.dropout(X)
        X = F.relu(self.dense1(X))
        X = F.softmax(self.output(X), dim=-1)
        return X

# %%
from skorch import NeuralNetClassifier

# %%
net = NeuralNetClassifier(
    ClassifierModule,
    max_epochs=20,
    lr=0.1,
#     device='cuda',  # uncomment this to train with CUDA
)

# %%
net.fit(X, y)

# %%
y_pred = net.predict(X[:5])
y_pred


# %%
y_proba = net.predict_proba(X[:5])
y_proba

# %%


[...]

PR #1668, opened 2020-08-21
Implement skorch pruning callback

crcrpar:
Integration module for skorch and an example for it.


optuna/integration/skorch.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Args:
+        trial:
+            A :class:`~optuna.trial.Trial` corresponding to the current evaluation of the
+            objective function.
+        monitor:
+            An evaluation metric for pruning, e.g. ``val_loss`` or
+            ``val_acc``. The metrics are obtained from the returned dictionaries,
+            i.e., ``net.histroy``. The names are thus depend on how this dictionary
+            is formatted.
```
HideakiImamura, line 25: When ``val_loss`` is used, the user should specify `direction=minimize` in `optuna.create_study`. On the other hand, when ``val_acc`` is used, the user should specify ``direction=maximize` in `optuna.create_study`. I think it is kind for users to provide this information here.

optuna/integration/skorch.py
```diff
@@ -0,0 +1,45 @@
...
+with optuna._imports.try_import() as _imports:
+    from skorch.callbacks import Callback
+    from skorch.net import NeuralNet
+
+if not _imports.is_successful():
+    Callback = object  # NOQA
+    NeuralNet = object  # NOQA
+
+
+class SkorchPruningCallback(Callback):
```
HideakiImamura, line 14: How do we make this class "experimental" using the `experimental` decorator?

HideakiImamura (commented):
Thanks for the PR. This PR seems to be related to issue #1654. I have some early comments. 

I would like to know how this PR proceeds. What is your plan to remove the "draft"? I think adding an example is reasonable.

optuna/integration/skorch.py
crcrpar: What do you mean by that?

optuna/integration/skorch.py
crcrpar: That sounds a bit too much to me, and I don't think that kind comment should be on some other place like https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html?highlight=direction#optuna.study.Study.direction

crcrpar:
Yeah, an example would be a required thing to mark this pull request as ready for review and that's my plan, thanks.

optuna/integration/skorch.py
HideakiImamura: I think the more helpful the documentation is, the better for users. I would like to know other reviewers' opinions.

optuna/integration/skorch.py
HideakiImamura: Sorry for my grammatical mistake. I mean: how about using the experimental decorator for this newly introduced class?
```suggestion
@experimental("2.1.0")
class SkorchPruningCallback(Callback):
```

crcrpar: As per our internal discussion which is now published in https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#experimental-feature, I don't want to say this feature as "experimental".

optuna/integration/skorch.py
crcrpar: Okay, if that's the case, we should file some issues for some pruning callbacks that doesn't convey that information and label them "contribution-welcome".

optuna/integration/skorch.py
HideakiImamura: I got it. So this class is stable and there is no room for the interface to change at all? I think it's OK that we do not mark this class as "experimental".

By the way, how about other integration pruning callback classes? For example, I wonder whether the experimental decorator for the `CatalystCallback` can be removed.

optuna/integration/skorch.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def __init__(self, trial: optuna.trial.Trial, monitor: str) -> None:
+
+        _imports.check()
+
+        super().__init__()
+        self._trial = trial
+        self._monitor = monitor
+
+    def on_epoch_end(self, net: NeuralNet, **kwargs: Any) -> None:
```
crcrpar, line 36: this should be
```suggestion
    def on_epoch_end(self, net: "skorch.NeuralNet", **kwargs: Any) -> None:
```
?

crcrpar:
@UTUnex Does this example look good to you? Any feedback is welcome 😃

UTUnex:
Thank you all for your great efforts on the work. I checked the new example but I just find another module that is closer to what I expect for. The modul is __'optuna.integration.OptunaSearchCV ()'__, yeah, honestly, this is what I really like, which means I can just put all the required parameters in this function based on the demands.  I've just read the documentation about the optunasearchCV( ) and have some questions :  
(1) I don't know whether the estimator here can be the NeuralNetClassifer( ) in skorch, I haven't tried it but I hope it's compatible;
(2) I didn't find the option of tuning algorithm in the searchCV(), such as 'TPE', ''bayes', 'grid' or ' 'random'; if possible, I hope this can be added;
(3) Also, I didn't find the early stopping option in the searchCV( ), such as 'hyperband', 'BOHB' or the other advanced schemes which are useful for DL task tuning on large scale datasets;
(4) And, is it possible to add the single and multiple GPU support in the searchCV(), like searchCV( GPU usage = True, GPU_num = 2)
(5) How to output the searchCV results to a logger like tensorboard, ML flow, wandb etc, so that I can monitor and analyze the tuning process and hyperparameters combinations.

Summary: if possible, I hope you can add more functions to the optunasearchCV() and the expected code structure should be like : 
%% Data loading and preprocessing
     ....
%% Net design using pytorch  
     Class XXnet(nn.module)
         ....  
%% define the training function using skorch  
      net = NeuralNetClassifier( ...)  
%% hyperparameter tuning and logging to a certain dashboard:  
      search = optunasearchCV(net, **kwargs)
      search.fit(X_train, y_train)
%% Test  
      ...

The above is my advice, please check it. Thank you.

crcrpar:
Thank you for your feedback!
Now that it seems that you are interested in OptunaSearchCV rather than this integration module which is almost out of scope of this pull request so that I (and maybe others) would like to support you in your original issue or gitter 😃

examples/skorch_simple.py
```diff
@@ -0,0 +1,108 @@
+"""
+Optuna example that optimizes multi-layer perceptrons using skorch.
+
+In this example, we optimize the validation accuracy of hand-written digit recognition using
+skorch, and MNIST. We optimize the neural network architecture. As it is too time
```
HideakiImamura, line 5: A nit.
```suggestion
skorch and MNIST. We optimize the neural network architecture. As it is too time
```

examples/skorch_simple.py
```diff
@@ -0,0 +1,108 @@
...
+argument.
+    $ python skorch_simple.py [--pruning]
+
+"""
+
+import argparse
+
+import numpy as np
+import optuna
+from optuna.integration import SkorchPruningCallback
```
HideakiImamura, line 18: Import statements of `optuna` modules should be placed after the other third-party libraries like [this](https://github.com/optuna/optuna/blob/master/examples/pytorch_simple.py).

examples/skorch_simple.py
```diff
@@ -0,0 +1,108 @@
...
+SUBSET_RATIO = 0.4
+
+mnist = fetch_openml("mnist_784", cache=False)
+
+X = mnist.data.astype("float32")
+y = mnist.target.astype("int64")
+indices = np.random.permutation(len(X))
+N = int(len(X) * SUBSET_RATIO)
+X = X[indices][:N]
+y = y[indices][:y]
```
HideakiImamura, line 37: I'm not sure, but
```suggestion
y = y[indices][:N]
```
?

examples/skorch_simple.py
```diff
@@ -0,0 +1,108 @@
...
+        callbacks=[SkorchPruningCallback(trial, "valid_acc")],
+    )
+
+    net.fit(X_train, y_train)
+
+    return accuracy_score(y_test, net.predict(X_test))
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="PyTorch Lightning example.")
```
HideakiImamura, line 84:
```suggestion
    parser = argparse.ArgumentParser(description="Skorch example.")
```

optuna/integration/skorch.py
HideakiImamura: Yes, I agree with you. As @hvy pointed in https://github.com/optuna/optuna/pull/1705#discussion_r476296682, this is rendered in the docs as `object`. Using `"skorch.NeuralNet"` instead `NeuralNet` leads to informative documentation.

HideakiImamura (commented):
Thank you for your confirmation. I have several comments. Could you take a look?

HideakiImamura (approved):

optuna/integration/skorch.py
```diff
@@ -0,0 +1,44 @@
...
+
+with optuna._imports.try_import() as _imports:
+    from skorch.callbacks import Callback
+    from skorch.net import NeuralNet
+
+if not _imports.is_successful():
+    Callback = object  # NOQA
+
+
+class SkorchPruningCallback(Callback):
```
crcrpar, line 13: TODO: Add `versionadded` 

ref: https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#write-versionadded-directive-to-the-doc-of-new-features

optuna/integration/skorch.py
```diff
@@ -0,0 +1,44 @@
...
+    """Skorch callback to prune unpromising trials.
+
+    Args:
+        trial:
+            A :class:`~optuna.trial.Trial` corresponding to the current evaluation of the
+            objective function.
+        monitor:
+            An evaluation metric for pruning, e.g. ``val_loss`` or
+            ``val_acc``. The metrics are obtained from the returned dictionaries,
+            i.e., ``net.histroy``. The names are thus depend on how this dictionary
```
toshihikoyanase, line 23: A nit. Or, it may be a typo for `The names are thus dependent on ...`.
```suggestion
            i.e., ``net.histroy``. The names thus depend on how this dictionary
```

tests/integration_tests/test_skorch.py
```diff
@@ -0,0 +1,47 @@
...
+    valid_ds = torch.utils.data.TensorDataset(torch.zeros(3, 4), torch.zeros(3, dtype=torch.long))
+
+    def objective(trial: optuna.trial.Trial) -> float:
+
+        net = skorch.NeuralNetClassifier(
+            ClassifierModule,
+            max_epochs=10,
+            lr=0.02,
+            callbacks=[SkorchPruningCallback(trial, "valid_acc")],
+            train_split=predefined_split(valid_ds),
```
toshihikoyanase, line 34: In my environment, an error was occurred with `skorch==0.9.0`.  It works with `skorch==0.8.0`, and the new version may break compatibility.

toshihikoyanase (changes requested):
Thank you for your PR. `skorch` seems to be developed actively, and I think the callback and example help its users. The change basically looks good to me, but I have some small comments. Please take a look.

optuna/integration/skorch.py
crcrpar: @toshihikoyanase what's your take here?

optuna/integration/skorch.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def on_epoch_end(self, net: "NeuralNet", **kwargs: Any) -> None:
+        history = net.history
+        if not history:
+            return
+        epoch = len(history) - 1
+        current_score = history[-1, self._monitor]
+        self._trial.report(current_score, epoch)
+        if self._trial.should_prune():
+            message = "Trial was pruned at epoch {}".format(epoch)
```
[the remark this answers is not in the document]
crcrpar: :+1:

tests/integration_tests/test_skorch.py
crcrpar: Fixed this in https://github.com/optuna/optuna/pull/1668/commits/ce004db0d68ff0190ae6e3f9d59b9e56e3bc32df.

optuna/integration/skorch.py
crcrpar: Thanks.

optuna/integration/skorch.py
toshihikoyanase: IMO, it is a bit verbose if we add the comments about `direction` here because the other pruning callbacks such as `KerasPruningCallback` and `XGBoostPruningCallback` also have the metrics similar to `monitor`.

So, I'd like to place such a comment in the reference of direction (`Study.direction` or `create_study`) or [tutorial](https://optuna.readthedocs.io/en/stable/tutorial/pruning.html#activating-pruners) of pruning.

toshihikoyanase (commented):
The code and example look good to me.
Could you add `SkorchPruningCallback` reference to `docs/source/reference/integration.rst`?

toshihikoyanase (approved):
I also confirmed that the CI job for the example worked in my local environment.

PR #1695, opened 2020-08-23
Introduce isort to automatically sort import statements.

### LINKED ISSUE #1694: Introduce isort to sort imports?
## Motivation

It's a pain for both contributors and maintainers to sort imports. It'd be nice if we could automate this process.

## Solution

https://github.com/timothycrosley/isort

harupy:
- Introduce isort to automatically sort import statements. 
- Fixes #1694


pyproject.toml
```diff
@@ -17,3 +17,10 @@ exclude = '''
   | tutorial  # Optuna specific
 )/
 '''
+
+[tool.isort]
+profile = 'black'
+src_paths = ['optuna', 'tests', 'docs']
+skip_glob = ['docs/source/conf.py', '**/alembic/versions/*.py']
```
harupy, line 24: Ignore sphinx `conf.py` and alembic version files

harupy:
@HideakiImamura I think it's ready for review.

setup.py
```diff
@@ -51,7 +50,7 @@ def get_tests_require() -> List[str]:
 def get_extras_require() -> Dict[str, List[str]]:
 
     requirements = {
-        "checking": ["black", "hacking", "mypy"],
+        "checking": ["black", "flake8", "isort", "mypy"],
```
harupy, line 53: Replaced `hacking` with `flake8`.

harupy: maybe we should pin `isort` because the release cycle of `isort` is really fast recently.

https://pypi.org/project/isort/#history

HideakiImamura:
@harupy Thanks for your swift actions. I will start the review.

examples/pytorch_lightning_simple.py
```diff
@@ -15,14 +15,14 @@
...
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-from torch.optim import Adam
 import torch.utils.data
+from packaging import version
+from pytorch_lightning import Callback
+from torch.optim import Adam
 from torchvision import datasets
 from torchvision import transforms
```
harupy, line 27: https://github.com/PyCQA/isort#controlling-how-isort-sections-from-imports says:

> By default isort places straight (import y) imports above from imports (from x import y):

```python
import b
from a import a  # This will always appear below because it is a from import.
```

> However, if you prefer to keep strict alphabetical sorting you can set force sort within sections to true. Resulting in:

```python
from a import a  # This will now appear at top because a appears in the alphabet before b
import b
```

harupy: I don't have a strong opnion about this, but the default import-comes-first sorting rule looks much cleaner and straightforward.

HideakiImamura: Totally agree with you. It makes the sorting rule simple. I'd like to know other reviewers' opinions.

HideakiImamura: I think it is kind for developers to add an explanation of the  "import-comes-first" rule in `CONTRIBUTING.md`.

optuna/study.py
```diff
@@ -13,26 +14,25 @@
...
 from optuna import progress_bar as pbar_module
 from optuna import pruners
 from optuna import samplers
 from optuna import storages
 from optuna import trial as trial_module
-from optuna.trial import create_trial
+from optuna._experimental import experimental
+from optuna._imports import try_import
+from optuna._study_direction import StudyDirection
+from optuna._study_summary import StudySummary  # NOQA
```
HideakiImamura, line 32: Because these imports are placed after the imports of pruners, samplers, storages, and trials, the circular imports occur. How about separating these imports and other optuna imports by a blank line? The specific change will resolve circular imports.

HideakiImamura (commented):
Thanks for the PR! I have several comments. 

I'm a bit concerned with how to resolve the circular imports for `optuna.study`. I think `StudyDirection` should be imported as `from optuna.study import StudyDirection` by developers. I'd like to discuss this point with @harupy and other reviewers.

optuna/importance/__init__.py
```diff
@@ -4,11 +4,16 @@
...
+from optuna.importance._mean_decrease_impurity import MeanDecreaseImpurityImportanceEvaluator
 from optuna.study import Study
 
+__all__ = [
+    "BaseImportanceEvaluator",
+    "FanovaImportanceEvaluator",
+    "MeanDecreaseImpurityImportanceEvaluator",
+    "get_param_importances",
+]
+
```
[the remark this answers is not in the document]
harupy: No. I should have put a comment for this. To pass `flake8`, `isort` and `black`, I had to do something like:

```
# fmt: off
from optuna.importance._mean_decrease_impurity import (  # NOQA, skip: isort
    MeanDecreaseImpurityImportanceEvaluator
)
```

optuna/integration/chainermn.py
```diff
@@ -6,8 +7,8 @@
 from typing import Sequence
 from typing import Tuple
 from typing import Type
-import warnings
 
+from optuna import TrialPruned
```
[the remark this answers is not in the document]
harupy: Found a option to turn this off: https://github.com/PyCQA/isort/blob/74d8d65aaaa375c4f7fb100840aad25265be1763/docs/configuration/options.md#order-by-type

optuna/integration/sklearn.py
```diff
@@ -17,16 +17,16 @@
 import scipy as sp
 from scipy.sparse import spmatrix
 
-from optuna._experimental import experimental
-from optuna._imports import try_import
+from optuna import TrialPruned
```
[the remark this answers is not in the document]
harupy: Found a option to turn this off: https://github.com/PyCQA/isort/blob/74d8d65aaaa375c4f7fb100840aad25265be1763/docs/configuration/options.md#order-by-type

optuna/study.py
harupy: Isort ignores a separation by a blank line.

isort

optuna/integration/chainermn.py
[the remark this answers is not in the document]
harupy: Looks like there is no option to change this rule: https://github.com/PyCQA/isort/blob/develop/docs/configuration/options.md

optuna/importance/__init__.py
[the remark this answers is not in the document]
harupy: `__all__` prevents `flake8` from raising F401 (unused-import).

examples/pytorch_lightning_simple.py
harupy: I'll update `CONTRIBUTING.md` in a follow-up.

examples/tfkeras_simple.py
```diff
@@ -17,7 +17,6 @@
 
 import optuna
 
-
```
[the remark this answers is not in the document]
harupy: Found an option to control this: https://github.com/PyCQA/isort/blob/develop/docs/configuration/options.md#lines-after-imports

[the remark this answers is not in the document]
HideakiImamura: Thanks for the survey and reporting. In this regard, I prefer to keep the existing codebase as unchanged as possible. What do you think?

optuna/importance/__init__.py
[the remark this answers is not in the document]
HideakiImamura: Thanks for the explanation! I understood. Let me resolve this conversation.

optuna/integration/chainermn.py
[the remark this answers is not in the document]
HideakiImamura: Thank you for the investigation. So in order to use `isort`, we have to allow this rule. I don't think there is a problem with this point. We can accept this rule.

optuna/integration/sklearn.py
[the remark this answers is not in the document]
HideakiImamura: In my opinion, sorting by type does not contribute anything to readability. I think we should prevent the rule, but I'd like to know other opinions. What do you think?

optuna/study.py
HideakiImamura: Hmm... Thanks for the explanation. So in order to use `isort` in Optuna, we have to allow this rule: "not `from optuna.study import StudyDirection` but `from optuna._study_direction import StudyDirection` in the Optuna codebase". But, int the `integration` module, we can use `from optuna.study import StudyDirection`?

We should note this rule in `CONTRIBUTING.md` as a follow-up task.

harupy: or we can use `isort: skip` like:

```
# don't sort these imports
import a  # isort: skip
import b  # isort: skip
import c  # isort: skip

import ...
import ...
```

HideakiImamura: Thanks for providing an important suggestion. Looks good. I think it's better to use `isort: skip` for `from optuna._study_direction import StudyDirection` and `from optuna._study_summary import StudySummary`.  What do you think?

HideakiImamura: I internally discussed with Optuna core-dev members. As a result, we can accept `from optuna._study_direction import StudyDirection` in Optuna codebase. This new convention leads to reduce the number of codes dependent on the order of imports.

As a follow-up task, we should note in `CONTRIBUTING.md` to use `from optuna._study_direction import StudyDirection`.

harupy: Makes sense.

optuna/integration/sklearn.py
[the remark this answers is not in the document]
harupy: Either way is fine for me. I'll turn off this rule.

hvy (commented):
There's definitely a point in automating the checks for the import statements. Thanks a lot for the PR. 

The current configuration although somewhat different from what we have looks reasonable. I have one comment regarding the removal of `hacking` (and thus the relaxation of the [OpenStack guidelines](https://docs.openstack.org/hacking/latest/user/hacking.html#styleguide)). Maybe we can keep `hacking` but just skip the checks involving import statements. I think that'd make this PR easier to merge.

setup.cfg
```diff
@@ -1,6 +1,6 @@
 # This section is for flake8.
 [flake8]
-ignore = E203,W503
+ignore = E203,W503,H3
```
hvy, line 3: A nit. How about leaving a comment that `H3` covers `hacking`s import statement guidelines and that they're ignored because of `isort` compatibility?

harupy: @hvy Done!

hvy (approved):
LGTM, thanks for the big effort! 

We should update contribution guidelines in a follow up. I guess no changes are required in the coding style wiki as we don't mention import conventions.

HideakiImamura (approved):
Thank you for the swift actions! LGTM!

PR #1703, opened 2020-08-23
Use `Optional` for keyword arguments that default to `None`.

harupy:
## Motivation

Use `Optional` for arguments that defalult to `None`.

## Description of the changes

Use `Optional` for arguments that defalult to `None`.

## Test

Ran this script (**note that this script doesn't work for functions annotated using type comments**).

<details>
 <summary>Expand</summary>

```python
import ast
import os
import sys

from typing import Iterable, Union, List, Optional


def read_file(path: str) -> str:
    """Read a file and return its content"""
    with open(path, encoding="utf-8") as f:
        return f.read()


def iter_python_scripts(directory: str) -> Iterable[str]:
    """Yield all python scripts in the given directory"""
    for root, _, files in os.walk(directory):
        for f in files:
            if f.endswith(".py"):
                yield os.path.join(root, f)


def iter_func_defs(node: ast.AST) -> Iterable[ast.FunctionDef]:
    """Yield all function definitions in the given AST node"""
    for n in ast.iter_child_nodes(node):
        if isinstance(n, ast.FunctionDef):
            yield n
        yield from iter_func_defs(n)


def test_iter_func_defs() -> None:
    code = """
def a():
    pass

def b():
    def c():
        pass
    return c

class Class:
    def d(self):
        pass

    def e(self):
        def f():
            pass
        return f
"""
    root = ast.parse(code, type_comments=True)
    func_defs = [fd.name for fd in iter_func_defs(root)]

    assert func_defs == ["a", "b", "c", "d", "e", "f"]


def is_none(n: ast.AST) -> bool:
    return isinstance(n, ast.Constant) and (n.value is None) and (n.kind is None)


def is_annotated_as_optional(n: Optional[ast.expr]) -> bool:
    return (
        isinstance(n, ast.Subscript)
        and isinstance(n.value, ast.Name)
        and (n.value.id == "Optional")
    )


def contains_kwargs_that_should_be_annotated_as_optional(
    args: List[ast.arg], defaults: List[ast.expr]
) -> bool:
    """
    Return True if the given `args` contains an argument that defaults to `None`, but is NOT annotated
    as `Optional`. Note that this function doesn't work for functions annotated with type comments.
    """
    for arg, default in zip(args, defaults):
        if (
            # Return True if this argument defaults to None, but is not annotated as Optional.
            is_none(default)
            and (arg.annotation is not None)
            and not is_annotated_as_optional(arg.annotation)
        ):
            return True

    return False


def main() -> None:
    if sys.version_info[:2] < (3, 8):
        raise Exception(
            "This script throws in Python < 3.8 "
            "because `ast.parse` doesn't support the `type_comments` argument in Python < 3.8 "
            "See: https://docs.python.org/3/library/ast.html#ast.parse"
        )

    # ensure that `iter_func_defs` works as expected
    test_iter_func_defs()

    for d in ["optuna", "tests"]:
        for path in iter_python_scripts(d):
            src = read_file(path)
            root = ast.parse(src, type_comments=True)

            for f in iter_func_defs(root):
                from pprint import pprint

                defaults = f.args.defaults
                kwargs = f.args.args

                if contains_kwargs_that_should_be_annotated_as_optional(
                    f.args.args[-len(defaults) :], f.args.defaults
                ) or contains_kwargs_that_should_be_annotated_as_optional(
                    f.args.kwonlyargs, f.args.kw_defaults
                ):

                    print("{}:{}:{}".format(path, f.lineno, f.col_offset))


if __name__ == "__main__":
    main()
```

</details>


optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -919,7 +921,7 @@ def __init__(
         categorical_feature: str = "auto",
         early_stopping_rounds: Optional[int] = None,
         fpreproc: Optional[Callable[..., Any]] = None,
-        verbose_eval: Optional[Union[bool, int]] = True,
+        verbose_eval: Union[bool, int] = True,
```
toshihikoyanase, line 924: According to the [reference](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.cv.html#lightgbm.cv), `verbose_eval` of `lightgbm.cv` accepts `None`. So could you revert this change?
(I noticed that the default value is not matched with `lightgbm.cv` although it is not in the scope of this PR.)
```suggestion
        verbose_eval: Optional[Union[bool, int]] = True,
```

tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
```diff
@@ -52,7 +52,7 @@ def current_iteration(self):
 
 
 @contextlib.contextmanager
-def turnoff_cv(metric: Optional[str] = "binary_logloss") -> Generator[None, None, None]:
+def turnoff_cv(metric: str = "binary_logloss") -> Generator[None, None, None]:
```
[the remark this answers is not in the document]
toshihikoyanase: I confirmed that the `metric` argument does not receive `None` in the test cases. SGTM.

toshihikoyanase (commented):
Thank you for fixing the type hints. I have a comment on the argument of `LightGBM.cv`. Please take a look.

optuna/integration/_lightgbm_tuner/optimize.py
harupy: Done!

toshihikoyanase (approved):

PR #1705, opened 2020-08-23
Fix type hint of `optuna/integration`

akihironitta:
## Motivation
See #711.

## Description of the changes
Fixed type hints of some of the files under `optuna/integration`.

### DIFF
--- a/optuna/integration/fastai.py
+++ b/optuna/integration/fastai.py
@@ -8,8 +8,8 @@ with try_import() as _imports:
     from fastai.callbacks import TrackerCallback
 
 if not _imports.is_successful():
+    Learner = object  # NOQA
     TrackerCallback = object  # NOQA
-
 
 class FastAIPruningCallback(TrackerCallback):
     """FastAI callback to prune unpromising trials for fastai.
--- a/optuna/integration/fastai.py
+++ b/optuna/integration/fastai.py
@@ -4,11 +4,10 @@ import optuna
 from optuna._imports import try_import
 
 with try_import() as _imports:
-    from fastai.basic_train import Learner  # NOQA
+    from fastai.basic_train import Learner
     from fastai.callbacks import TrackerCallback
 
 if not _imports.is_successful():
-    Learner = object  # NOQA
     TrackerCallback = object  # NOQA
 
 
@@ -47,7 +46,7 @@ class FastAIPruningCallback(TrackerCallback):
             details.
     """
 
-    def __init__(self, learn: Learner, trial: optuna.trial.Trial, monitor: str) -> None:
+    def __init__(self, learn: "Learner", trial: optuna.trial.Trial, monitor: str) -> None:
 
         super(FastAIPruningCallback, self).__init__(learn, monitor)
 
--- a/optuna/integration/fastai.py
+++ b/optuna/integration/fastai.py
@@ -11,6 +11,7 @@ if not _imports.is_successful():
     Learner = object  # NOQA
     TrackerCallback = object  # NOQA
 
+
 class FastAIPruningCallback(TrackerCallback):
     """FastAI callback to prune unpromising trials for fastai.
 
--- a/optuna/integration/fastai.py
+++ b/optuna/integration/fastai.py
@@ -1,9 +1,7 @@
+from typing import Any
+
 import optuna
 from optuna._imports import try_import
-from optuna import type_checking
-
-if type_checking.TYPE_CHECKING:
-    from typing import Any  # NOQA
 
 with try_import() as _imports:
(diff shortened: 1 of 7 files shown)


akihironitta:
```
Exception occurred:
  File "/home/docs/project/venv/lib/python3.8/site-packages/optuna/integration/fastai.py", line 49, in FastAIPruningCallback
    def __init__(self, learn: Learner, trial: optuna.trial.Trial, monitor: str) -> None:
NameError: name 'Learner' is not defined
```
It seems `from fastai.basic_train import Learner` failed (possibly because fastai recently updated to v2).
```
$ python -c "import fastai; print(fastai.__version__); from fastai.basic_train import Learner"
1.0.0
```
```
$ python -c "import fastai; print(fastai.__version__); from fastai.basic_train import Learner"
2.0.0
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'fastai.basic_train'
```

For this PR, should I add a line `Learner = object` as in `optuna/integration/pytorch_lightning.py` to pass the tests? https://github.com/optuna/optuna/blob/78777d92bec121bed25c25b2843fee7639262759/optuna/integration/pytorch_lightning.py#L8-L11

crcrpar:
The change you suggest is necessary in this case.

Just for your information, we've pinned the fastai version as I've expected fastai to introduce some breaking changes :wink: 

- https://github.com/optuna/optuna/blob/9b0081a212cba04e97a28ca510e5e72559b01a0d/setup.py#L103
- https://github.com/optuna/optuna/blob/9b0081a212cba04e97a28ca510e5e72559b01a0d/setup.py#L147
- https://github.com/optuna/optuna/blob/9b0081a212cba04e97a28ca510e5e72559b01a0d/setup.py#L189

akihironitta:
Thank you for your response and pointing out that it's already been handled.

HideakiImamura (approved):
Thanks for the PR! LGTM!

optuna/integration/fastai.py
```diff
@@ -1,15 +1,14 @@
...
-
-if type_checking.TYPE_CHECKING:
-    from typing import Any  # NOQA
 
 with try_import() as _imports:
     from fastai.basic_train import Learner  # NOQA
     from fastai.callbacks import TrackerCallback
 
 if not _imports.is_successful():
+    Learner = object  # NOQA
```
hvy, line 11: These will be [rendered as `object` in the docs](https://54001-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/generated/optuna.integration.FastAIPruningCallback.html#optuna.integration.FastAIPruningCallback) which is somewhat misleading. Do you think it's possible to omit these `object` assignments altogether and and modify the type hints to in this case `"Learner"` instead fo `Learner`? We do that in several other integration modules e.g. https://github.com/optuna/optuna/blob/master/optuna/integration/chainer.py#L49.

optuna/integration/tfkeras.py
```diff
@@ -40,8 +37,7 @@ def __init__(self, trial, monitor):
         self._trial = trial
         self._monitor = monitor
 
-    def on_epoch_end(self, epoch, logs=None):
-        # type: (int, Dict[str, Any]) -> None
+    def on_epoch_end(self, epoch: int, logs: Dict[str, Any] = None) -> None:
```
hvy, line 40: Stretches the scope of this PR so it's optional, but we could also correct the type hint here (`Optional` since it can take the value `None`). C.f. https://github.com/optuna/optuna/pull/1703
```suggestion
    def on_epoch_end(self, epoch: int, logs: Optional[Dict[str, Any]] = None) -> None:
```
Note that `typing.Optional` must be imported in addition to the suggestion.

hvy (changes requested):

optuna/integration/fastai.py
akihironitta: Sure. Thank you for reviewing and kindly showing an example.

akihironitta:
Reflecting @hvy's review, I made several changes to the code.

optuna/integration/lightgbm.py
```diff
@@ -34,6 +32,8 @@
     setattr(sys.modules[__name__], "LightGBMTuner", tuner.__dict__["LightGBMTuner"])
     setattr(sys.modules[__name__], "LightGBMTunerCV", tuner.__dict__["LightGBMTunerCV"])
 
+    CallbackEnv = object  # NOQA
```
hvy, line 35: Can we not omit this assignment and use the string notation for `CallbackEnv` as well?

akihironitta: Sure.

hvy (approved):
Thanks for your changes, LGTM!

Issue #1539, opened 2020-07-21
Matplotlib backend for visualization functions


g-votte:
This issue proposes to implement the [Matplotlib](https://matplotlib.org/) backend for the visualization functions.

## Motivation

While the current visualization backend, [Plotly](https://plotly.com/python/), provides rich and interactive interfaces, the setup is sometimes hard especially when exporting figures as static ones. Specifically, the installation steps of [Orca](https://github.com/plotly/orca) are complicated in some environments (see https://github.com/plotly/orca/issues/150 for example) or may restrict the way of package management as `plotly-orca` favors Conda installation.

Providing another backend such as Matpotlib may mitigate this problem, where the library does not require extra setup to output static figures and is pre-installed in some popular platforms such as [Anaconda](https://www.anaconda.com/).

## Alternatives

Plotly has supported [Kaleido backend](https://github.com/plotly/plotly.py#kaleido) for static export since Plotly 4.9. Thus, we could guide users to install this backend, although it restricts the Plotly version where Optuna supports `>= 4.0.0` currently.

ytknzw:
@toshihikoyanase Thank you for your feedback! My idea above is a rough draft, so I appreciate your detailed thoughts. I agree with both points you mentioned. I'll work on it today and put a PR, then let's discuss further.

hvy:
I'm wondering if we shouldn't start discussing the public part first (or sooner), including how to handle visualization functions that are missing implementations for some specific backends. Logic introduced is otherwise quite difficult to document and test.

BenjaminBossan:
If I may make a proposal regarding the public API of the matplotlib plotting functions:

It would be really nice if as a user, I were allowed to pass the `ax` to the plotting function. This allows me to have a greater control about the final plot. E.g., right now I cannot change the plot size or make the plot part of a larger figure with several subplots. Here is an example how I could use this feature:

```python
...
import matplotlib.pyplot as plt
fig, axes = plt.subplots(1, 2, figsize=(20, 10))

# plot results on a big figure with 2 subplots
optuna.visualization.matplotlib.plot_edf(study, ax=axes[0])
optuna.visualization.matplotlib.plot_optimization_history(study, ax=axes[1])
```

The default argument for `ax` could be None, in which case the functions do exactly the same thing as they do right now. Inside the different functions, you just need to change the code to:

```python

def _get_edf_plot(studies: List[Study], ax: Optional[AxesSubplot] = None) -> "Axes":
    ...
[...]

PR #1707, opened 2020-08-24
[WIP] Create matplotlib backend for plotting functions.

ytknzw:
This PR proposes to implement ``matplotlib`` backend for plotting functions.

## Motivation
See issue #1539 for the original motivation.

## Description of the changes
|Created file in ``optuna.visualization``|Implemented function|Corresponding function with Plotly backend|
|----|----|----|
|``_optimization_history_matplotlib.py``|``_get_optimization_history_plot_matplotlib``|``_get_optimization_history_plot``|
|``_intermediate_values_matplotlib.py``|``_get_intermediate_plot_matplotlib``|``_get_intermediate_plot``|
|``_edf_matplotlib.py``|``_get_edf_plot_matplotlib``|``_get_edf_plot``|

List of Plotly backend functions to be implemented
- [ ] ``_contour.plot_contour``
- [ ] ``_contour._get_contour_plot``
- [ ] ``_contour._generate_contour_subplot``
- [ ] ``_edf.plot_edf``
- [x] ``_edf._get_edf_plot``
- [ ] ``_intermediate_values.plot_intermediate_values``
- [x] ``_intermediate_values._get_intermediate_plot``
- [ ] ``_optimization_history.plot_optimization_history``
- [x] ``_optimization_history._get_optimization_history_plot``
- [ ] ``_parallel_coordinate.plot_parallel_coordinate``
- [ ] ``_parallel_coordinate._get_parallel_coordinate_plot``
- [ ] ``_param_importances.plot_param_importances``
- [ ] ``_param_importances._get_distribution``
- [ ] ``_param_importances._get_color``
- [ ] ~``_param_importances._make_hovertext``~
- [ ] ``_slice.plot_slice``
- [ ] ``_slice._get_slice_plot``
- [ ] ``_slice._generate_slice_subplot``

### DIFF
--- a/optuna/visualization/_optimization_history_matplotlib.py
+++ b/optuna/visualization/_optimization_history_matplotlib.py
@@ -1,5 +1,5 @@
-from matplotlib import pyplot as plt
 from matplotlib.figure import Figure
+from matplotlib import pyplot as plt
 
 from optuna.logging import get_logger
 from optuna.study import Study
--- a/optuna/visualization/_optimization_history_matplotlib.py
+++ b/optuna/visualization/_optimization_history_matplotlib.py
@@ -1,0 +1,65 @@
+from matplotlib import pyplot as plt
+from matplotlib.figure import Figure
+
+from optuna.logging import get_logger
+from optuna.study import Study
+from optuna.study import StudyDirection
+from optuna.trial import TrialState
+
+_logger = get_logger(__name__)
+
+
+def _get_optimization_history_plot_matplotlib(study: Study) -> Figure:
+    """Plot optimization history of all trials in a study with matplotlib.
+
+    Args:
+        study:
+            A :class:`~optuna.study.Study` object whose trials are plotted for their objective
+            values.
+
+    Returns:
+        A :class:`matplotlib.figure.Figure` object.
+    """
+
+    # Setup
+    plt.style.use("ggplot")  # Use ggplot style sheet for similar outputs to plotly.
+    fig, ax = plt.subplots()
+    ax.set_title("Optimization History Plot")
+    ax.set_xlabel("#Trials")
+    ax.set_ylabel("Objective Value")
+    cmap = plt.get_cmap("tab10")  # Use tab10 colormap for similar outputs to plotly.
+
+    # Prepare data for plotting
+    trials = [t for t in study.trials if t.state == TrialState.COMPLETE]
+
+    if len(trials) == 0:
+        _logger.warning("Study instance does not contain trials.")
+        return fig
+
+    best_values = [float("inf")] if study.direction == StudyDirection.MINIMIZE else [-float("inf")]
+    comp = min if study.direction == StudyDirection.MINIMIZE else max
+    for trial in trials:
+        trial_value = trial.value
+        assert trial_value is not None  # For mypy
+        best_values.append(comp(best_values[-1], trial_value))
+    best_values.pop(0)
+
+    # Draw graphs
+    ax.scatter(
+        x=[t.number for t in trials],
+        y=[t.value for t in trials],
+        color=cmap(0),
+        alpha=1,
+        label="Objective Value",
+    )
+    ax.plot(
+        [t.number for t in trials],
+        best_values,
+        marker="o",
+        color=cmap(3),
+        alpha=0.5,
+        label="Best Value",
+    )
+    ax.legend()
+
+    return fig
--- a/optuna/visualization/_optimization_history_matplotlib.py
+++ b/optuna/visualization/_optimization_history_matplotlib.py
@@ -21,7 +21,7 @@ def _get_optimization_history_plot_matplotlib(study: Study) -> Figure:
         A :class:`matplotlib.figure.Figure` object.
     """
 
-    # Setup
+    # Set up the graph style.
     plt.style.use("ggplot")  # Use ggplot style sheet for similar outputs to plotly.
     fig, ax = plt.subplots()
     ax.set_title("Optimization History Plot")
@@ -29,7 +29,7 @@ def _get_optimization_history_plot_matplotlib(study: Study) -> Figure:
     ax.set_ylabel("Objective Value")
     cmap = plt.get_cmap("tab10")  # Use tab10 colormap for similar outputs to plotly.
 
-    # Prepare data for plotting
+    # Prepare data for plotting.
     trials = [t for t in study.trials if t.state == TrialState.COMPLETE]
 
     if len(trials) == 0:
@@ -44,7 +44,7 @@ def _get_optimization_history_plot_matplotlib(study: Study) -> Figure:
         best_values.append(comp(best_values[-1], trial_value))
     best_values.pop(0)
 
-    # Draw graphs
+    # Draw a scatter plot and a line plot.
     ax.scatter(
         x=[t.number for t in trials],
         y=[t.value for t in trials],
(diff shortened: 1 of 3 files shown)


optuna/visualization/_optimization_history_matplotlib.py
```diff
@@ -0,0 +1,65 @@
...
+    Args:
+        study:
+            A :class:`~optuna.study.Study` object whose trials are plotted for their objective
+            values.
+
+    Returns:
+        A :class:`matplotlib.figure.Figure` object.
+    """
+
+    # Setup
```
toshihikoyanase, line 24: A nit, but could you use sentences instead of a word in the code comment?
```suggestion
    # Setup the graph style.
```

optuna/visualization/_optimization_history_matplotlib.py
```diff
@@ -0,0 +1,65 @@
...
+    plt.style.use("ggplot")  # Use ggplot style sheet for similar outputs to plotly.
+    fig, ax = plt.subplots()
+    ax.set_title("Optimization History Plot")
+    ax.set_xlabel("#Trials")
+    ax.set_ylabel("Objective Value")
+    cmap = plt.get_cmap("tab10")  # Use tab10 colormap for similar outputs to plotly.
+
+    # Prepare data for plotting
```
toshihikoyanase, line 32: Please add a period at the end of the sentence.
```suggestion
    # Prepare data for plotting.
```

optuna/visualization/_optimization_history_matplotlib.py
```diff
@@ -0,0 +1,65 @@
...
+
+    best_values = [float("inf")] if study.direction == StudyDirection.MINIMIZE else [-float("inf")]
+    comp = min if study.direction == StudyDirection.MINIMIZE else max
+    for trial in trials:
+        trial_value = trial.value
+        assert trial_value is not None  # For mypy
+        best_values.append(comp(best_values[-1], trial_value))
+    best_values.pop(0)
+
+    # Draw graphs
```
toshihikoyanase, line 47: ditto. This code seems to draw only a single graph, so `graphs` may be `a/the graph`.
```suggestion
    # Draw graphs.
```

toshihikoyanase (commented):
Thank you for your PR. It looks great!

I created a [notebook](https://colab.research.google.com/drive/1hhfe4KfsAR5pZIGqSOvIY4Ei2uaWRve-#scrollTo=4Sj7ZTZGsmYA) to check the generated graphs and they look nice.

Minimization
image

Maximization
image

A nit-picky, but I have some comments on the code comments. Please take a look.

ytknzw:
Thank you for your comments. Just updated the code comments.

Chillee:
I think this is quite nice - plotly has always been fairly difficult for me to install (particularly to get them to show up in jupyter).

hvy:
Great, you've already implemented several of them. I'm just curious what the scope of this PR is. I believe we should also discuss how to introduce the `matplotlib` dependency and how to expose these features to the user. The sooner the better?

ytknzw:
Yes, we should discuss how to bring features to users. Sooner is better.

The scope of this PR itself is to implement the ``matplotlib`` backend as internal functions. 
https://github.com/optuna/optuna/issues/1539#issuecomment-678752628

hvy:
Did you plan to cover all of them? I'm wondering if some visualization functions aren't somewhat tricky to implement, such as the parallel coordinate plot. As you've already done a great job implementing several of the existing functions, how about moving on to designing the public API? I think we can cover the logic for how Optuna should behave when a function is not implemented with either library there, and then work on the visualization functions one by one. Having separate PRs for each one would make reviewing easier as well, which presumably would benefit us all.

ytknzw:
Thank you for your comments.

1. Scope
>Did you plan to cover all of them?

Yes, I intend to implement all the functions with matplotlib backend in this PR.

2. Point of dev focus
>how about moving on to designing the public API? I think we can cover the logic for how Optuna should behave when a function is not implemented with either library there, and then work on the visualization functions one by one.

I'm confused. @toshihikoyanase commented in issue #1539 as follows:

>>how about just keeping the current API for now and adding a matplotlib version of get_optimization_history? It may be an internal function in the first PR, and then we can discuss how to make it public in the second PR.

On the other hand, in my understanding, your suggestion is to stop implementing matplotlib backend functions now and to start thinking about the modification of current APIs (e.g. ``plot_optimization_history``) first. Please correct if I misunderstand.

3. PR size
>Having separate PRs for each one would make reviewing easier as well, which presumably would benefit us all.

OK, I close this PR and will instead make separate PRs later.

hvy:
I’m sorry if my comment confused you, it was just an idea to open up a discussion. Let me try to make up for it.

First and foremost, you don’t have to close this PR. Instead, we could continue working on this one to discuss the public API.
Once that is fixed/this PR is merged. We can continue implementing the remaining visualization functions.

What do you think @toshihikoyanase, @ytknzw?

Some approaches for how we could go about designing the public API. 
1. Introduce a “backend” parameter to existing visualization functions.
    - User must consider the argument in each call, implementation becomes somewhat complicated, and the return value becomes a `Union`.
2. Define a global/thread local state. 
    - User only have to configure it once. Other cons remain. We introduce a global state.
3. Introduce `optuna.visualization.matplotlib` to define `plot_*`
    - User switched between backends by changing the top import statement(s). Pro that we don’t have to keep a global state, code is clearly separated. What functions are implemented in what backend becomes very clear. Return value is not a `Union`.

I would suggest 3., as you might have noticed. And if so, changes required from this PR are not that large and we could incorporate them here.

toshihikoyanase:
@ytknzw @hvy I'm sorry for confusing you. When I suggested separating the PRs, we didn't have implementation and I couldn't imagine the pros & cons of the Interface choices. Now, thanks to the hard work of @ytknzw , we have three visualization functions and IMO, we can discuss the interface now.

@hvy Thank you for your comments on the interface.  I agree with your suggestion to introduce `optuna.visualization.matplotlib`. It doesn't affect existing code, and we may be able to build the implementation based on the first and second approaches on top of it. In addition, I don't think it requires more code changes than the other approaches (I think (1) add public plot functions, (2) create the sub module directory, and move the files).

@ytknzw I'm really sorry for changing my mind, but I think we can keep this PR and add a few changes to create the minimum public interface. What do you think?

ytknzw:
@hvy @toshihikoyanase Thank you for clarifications and sorry for my slow reply! I agree with you two. To introduce ``optuna.visualization.matplotlib`` would be the best option.
Let me try it in this PR, or another PR is better for review?

hvy:
Not slow at all, thanks for taking your time.

> Let me try it in this PR, or another PR is better for review?

I'm fine with either. What do you feel more comfortable with?

It would be easier to focus on the implementation/review if we create different PRs, one for fixing the API, followed by PRs for implementing the graphics. It's not a strong opinion.

ytknzw:
Thanks. I agree with your points. Modifying API and implementing matplotlib backend are different tasks. I'll creat another PR for ``optuna.visualization.matplotlib``.

HideakiImamura:
Thanks, @ytknzw for the great work for plotting functions! I'm very excited to merge these functions into Optuna. I will check and review this and subsequent PRs.

ytknzw:
@HideakiImamura 
Thank you!
Unfortunately, the API structure in #1756 is now largely different from this PR, so I close this PR and instead crea . 1 PR for each graph style (function) based on #1756 .

Issue #1334, opened 2020-06-07
Update docstrings to match the current guideline.


keisuke-umezawa:
# Motivation
Recently, we have introduced the docstrings guideline, and it is written in [wiki](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#docstrings).

However, some docstrings do not follow it. So, here, I would like to make a list which we need to update.

# List
## Need to add examples
- [ ] optuna.study.create_study
- [ ] optuna.study.load_study
- [ ] optuna.study.delete_study
- [ ] optuna.study.get_all_study_summaries
- [ ] ...

## Need to add description
- [ ] ...

nzw0301:
Thank you for making the list. This issue seems to be a good first issue for someone who considers contributing optuna!  But IMO, a solution to this issue seems unclear. For example, `need to add see also`: we may not understand how to solve it by reading the [guideline](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions) because there is not `see also` related section in the page. Could you explain it a little bit more?

keisuke-umezawa:
Thank you for feedback. In my first idea, the reference of `optuna.study.Study.get_trials` and `optuna.study.Study.trials` should be connected by `See also`, because they are related each other.

But, as you said, it is not clear. I will write explanation.

keisuke-umezawa:
@nzw0301 
I added explanations for each sections. e.g. `Since optuna users will use these functions and methods, they should have an example to explain how to use it. So, we need to add an Example: section in the docstring.`.

JoshKoh:
I think we need examples for multiple objective optimization and study.stop as introduced recently in ver. 1.5.
Especially how to stop study once objective function achieves certain threshold. 
Also, example for pruning needs more explanation:
for step in range(100):   # what is this for? Why do we need range(100), how to determine a range?
        clf.partial_fit(train_x, train_y, classes=classes)

        # Report intermediate objective value.
        intermediate_value = 1.0 - clf.score(valid_x, valid_y)
        trial.report(intermediate_value, step)

HideakiImamura:
We still have a lot of tasks to work on for this Issue. @keisuke-umezawa's list contains the most important items that must be documented for the users. I'm sure there are many other tasks, but the items on the list should be documented first.

I think this is the number one first issue for users who are interested in contributing to Optuna in the future. I look forward to hearing your contributions!

hvy:
[...]

PR #1712, opened 2020-08-24
Add Example to optuna.study.load_study

bigbird555:
## Motivation

This PR is related to #1334

## Description of the changes

This PR updates the docstrings of optuna.study.load_study.
Testcode in Example can be run here https://colab.research.google.com/drive/1u4-hz1tF_O8bgzadgEJ5wwRw8MBJoBnf?usp=sharing

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -887,7 +887,7 @@ def load_study(
             import os
 
             if os.path.exists("example.db"):
-                raise RuntimeError("a database file already exists. Please remove the file named example.db")
+                raise RuntimeError("a database file already exists. Please remove example.db")
 
         .. testcode::
 
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -885,8 +885,9 @@ def load_study(
         .. testsetup::
 
             import os
+
             if os.path.exists("example.db"):
-                raise RuntimeError("DB file already exists. Please remove example.db")
+                raise RuntimeError("a database file already exists. Please remove the file named example.db")
 
         .. testcode::
 
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -880,6 +880,21 @@ def load_study(
     # type: (...) -> Study
     """Load the existing :class:`~optuna.study.Study` that has the specified name.
 
+    Example:
+
+        .. testcode::
+
+            import optuna
+
+            def objective(trial):
+                x = trial.suggest_float("x", 0, 10)
+                return x ** 2
+
+            study = optuna.create_study(storage="sqlite:///example.db", study_name="my_study")
+            study.optimize(objective, n_trials=3)
+
+            loaded_study = optuna.load_study(study_name="my_study", storage="sqlite:///example.db")
+
     Args:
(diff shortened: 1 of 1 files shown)


optuna/study.py
```diff
@@ -880,6 +880,21 @@ def load_study(
...
+            import optuna
+
+            def objective(trial):
+                x = trial.suggest_float("x", 0, 10)
+                return x ** 2
+
+            study = optuna.create_study(storage="sqlite:///example.db", study_name="my_study")
+            study.optimize(objective, n_trials=3)
+
+            loaded_study = optuna.load_study(study_name="my_study", storage="sqlite:///example.db")
```
toshihikoyanase, line 896: Could you clean up the `example.db` at the end of the execution? It may affect the execution of other test cases that use `example.db`, and it would be safe if we remove it. I think we can use the [`testcleanup` directive](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testcleanup) for this purpose.

toshihikoyanase (changes requested):
Thank you for your PR. It basically looks good to me except for the clean-up of the DB file.

optuna/study.py
hvy: Not a must, but if possible, we could check for its existence in `.. testsetup::` and exit with a more graceful message.

optuna/study.py
```diff
@@ -880,6 +880,21 @@ def load_study(
...
+
```
hvy, line 897: Just an idea but it might be pedagogical to demonstrate accessing some property of the loaded study. The example is otherwise quite long in contrast to the actual function call of interest, so this may help emphasizing it. E.g.
```python
assert len(loaded_study.trials) == len(study.trials)
```

optuna/study.py
```diff
@@ -880,6 +880,32 @@ def load_study(
     # type: (...) -> Study
     """Load the existing :class:`~optuna.study.Study` that has the specified name.
 
+    Example:
+
+        .. testsetup::
+
+            import os
+            if os.path.exists("example.db"):
+                raise RuntimeError("DB file already exists. Please remove example.db")
```
toshihikoyanase, line 889: A nit, but please add a blank line between the import line and the code body. And also, I found that we tend to use `database` instead of `DB`.
```suggestion
            import os

            if os.path.exists("example.db"):
                raise RuntimeError("The database file already exists. Please remove example.db.")
```

toshihikoyanase (commented):
Thank you for your update. It almost looks good to me, but I have a small comment.

toshihikoyanase (approved):
LGTM! Thank you for your update.
@hvy Please merge after your review.

hvy (approved):
LGTM!

Thanks @toshihikoyanase for the ping, and thanks a lot for the careful PR. I double checked it locally since the CI is broken (https://github.com/optuna/optuna/issues/1714) and it runs like a charm.

hvy:
Let me fix the title a bit.

Issue #1719, opened 2020-08-25
Make Optuna PEP 561 Compliant


MarioIshac:
## Motivation

Make `optuna` [PEP 561](https://www.python.org/dev/peps/pep-0561) compliant such that its code can be type checked when it is added as a dependency. 

## Description

Currently `optuna` has inline type hints such that running `mypy` on `optuna` itself reports success. However, when `optuna` is imported by another project, and `mypy` is run there, I'm met with the following error: 

```
error: Skipping analyzing 'optuna': found module but no type hints or library stubs
```

This is because `optuna` is not declared as PEP 561-compliant package. To quote from PEP 561:

> Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing.

To reproduce:

1) Create a new Python project with a `virtualenv`
2) Create file `sample.py` with contents:
```
import optuna
```
3) `pip install mypy`.
4) `pip install -e git+https://github.com/optuna/optuna#egg=optuna`
4) Observe that `mypy sample.py` results in 
```
error: Skipping analyzing 'optuna': found module but no type hints or library stubs
note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```
5) `pip install -e git+https://github.com/MarioIshac/optuna@pep-561-compliance#egg=optuna` (wiping the previous git clone)
5) Observe that `mypy sample.py` results in:
```
Success: no issues found in 1 source file
```

PR #1720, opened 2020-08-25
Make Optuna PEP 561 Compliant

MarioIshac:
Fixes [#1719](https://github.com/optuna/optuna/issues/1719) by adding a `py.typed` under package `optuna` and distributing it as part of package data in `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -222,6 +222,7 @@ setup(
             "storages/_rdb/alembic.ini",
             "storages/_rdb/alembic/*.*",
             "storages/_rdb/alembic/versions/*.*",
+            "py.typed",
         ]
     },
     python_requires=">=3.5",


HideakiImamura (approved):
Thank you for the PR! The introduced feature looks good to me. LGTM.

hvy (approved):
Great initiative, I wasn't aware of this PEP. LGTM!

Verified it locally creating/installing sdists. After this PR, types are checked as expected (using `mypy`).

```python
# foo.py
import optuna

def create_study_with_bad_type(study_name: int) -> float:
    return optuna.create_study(study_name=study_name)
```

```
$ mypy foo.py
foo.py:5: error: Incompatible return value type (got "Study", expected "float")
foo.py:5: error: Argument "study_name" to "create_study" has incompatible type "int"; expected "Optional[str]"
Found 2 errors in 1 file (checked 1 source file)
```

Just a note. If my understanding is correct, users of Optuna might suddenly see errors in their type hints if they used to ignore the error raised by the failed import of `optuna` and their type hints aren't complying to what's defined by Optuna.

@toshihikoyanase, do you mind taking a look at this change before merging as well?

MarioIshac:
> Just a note. If my understanding is correct, users of Optuna might suddenly see errors in their type hints if they used to ignore the error raised by the failed import of `optuna` and their type hints aren't complying to what's defined by Optuna.

Great point, currently `mypy foo.py --ignore-missing-imports` for your example would return exit code 0 for failed input (because this type of error is suppressed), but exit code 1 if this PR is accepted and the import is now successful (but then failed type hints are exposed).

hvy:
It might be a good idea to emphasize it in the release notes.

toshihikoyanase (approved):
Thank you for your PR. It will help the application and library developers of Optuna. 
The change looks good to me.

@hvy Thank you for letting me know.
As you mentioned this change will cause type checking errors in some applications. I found that [`xfeat`](https://github.com/pfnet-research/xfeat) actually showed an additional error [here](https://github.com/pfnet-research/xfeat/blob/master/xfeat/optuna_selector/_kbest_explorer.py#L69) as follows:

```console
$ mypy .
xfeat/optuna_selector/_kbest_explorer.py:69: error: Incompatible types in assignment (expression has type "FrozenTrial", variable has type "None")
```

So, I think we should notify this change to the Optuna users, for example, in the release note as you suggested.

toshihikoyanase:
@MarioIshac I merged your PR, and it will be included in `v2.1.0`.
Thank you for your great contribution.

PR #1723, opened 2020-08-26
Fix issue where doctests are not executed

### LINKED ISSUE #1714: doctest not excuting any tests
## Motivation

doctest (`make doctest`) is run on CircleCI ([config.yml](https://github.com/optuna/optuna/blob/master/.circleci/config.yml)) but no tests are actually found/executed. Errors are therefore left undetected.

It seems like this started to occur after https://github.com/optuna/optuna/pull/1400.

#### Example

https://app.circleci.com/pipelines/github/optuna/optuna/2220/workflows/46d795b4-52dd-4ff5-b60c-8596c86a9ba1/jobs/54009

```
Doctest summary
===============
    0 tests
    0 failures in tests
    0 failures in setup code
    0 failures in cleanup code
build succeeded.
```

## Description

doctest should find and execute actual tests.

## Alternatives (optional)

If `make doctest` is run twice (or `make html` followed by `make doctest`) tests are found/executed. Looking up what `sphinx-build doctest` actually does might help.

harupy:
## Motivation

Fixes #1714 

## Description of the changes

- Generate autosummary stub files before running `make doctest`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,6 +105,7 @@ jobs:
           command: |
             . venv/bin/activate
             cd docs
+            sphinx-autogen source/**/*.rst
             make doctest
 
   # Unit tests


harupy:
worked

https://app.circleci.com/pipelines/github/harupy/optuna/28/workflows/280b9cd4-8e7d-4609-aca9-3ea878131447/jobs/206

.circleci/config.yml
```diff
@@ -105,6 +105,7 @@ jobs:
           command: |
             . venv/bin/activate
             cd docs
+            sphinx-autogen source/**/*.rst
```
harupy, line 108: Ref.

https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#sphinx-autogen-generate-autodoc-stub-pages

hvy (approved):
Quick as always, thank you very much. LGTM!

By the way, perhaps in a follow-up, what do you think of creating a custom `doctest` target in [`docs/Makefile`](https://github.com/optuna/optuna/blob/master/docs/Makefile) that always runs this command first? That also address this problem for users running the doctest locally.

harupy:
@hvy Makes sense. So the target would look like:

```
doctest:
        sphinx-autogen source/**/*.rst
	@$(SPHINXBUILD) -M doctest "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```

HideakiImamura (approved):

hvy:
Thanks @HideakiImamura for the review!

@harupy yes exactly. I haven't verified the syntax but something along that line (I usually see the `-b` option so I'm not so confident about the history of this Makefile 🤔 ). Do you want to continue working on it or should I?

harupy:
@hvy Created #1732

PR #1644, opened 2020-08-17
[WIP] A prototype of `BatchTrial`

toshihikoyanase:
## Motivation
See #895.

## Description of the changes
This PR is currently work-in-progress.

PR #1729, opened 2020-08-26
Add batch evaluation of the objective function

toshihikoyanase:
## Motivation
This PR is an alternative approach of #1644.

The approach in #1644 seems to have some problems:

### Inconsistent interface with `Study`

The implementation in #1644 creates `BatchStudy` using its `__init__` method, and `BatchStudy` has `batch_optimize` method. Users may get confused about whether they should use `study` or `batch_study` to analyze the results.

```python
batch_study = BatchStudy(study, batch_size=4)
batch_study.batch_optimize(objective, n_batches=10)
```

### Requires modification of `Study`

It modifies `Study` class to implement `BatchStudy` and `BatchMultiObjectiveStudy`. This is because `BatchStudy` inherits `Study` to reuse most of the internal functions and rewrite a part of them.

## Description of the changes

This implementation provides `optuna.batch.create_study` to create the `BatchStudy`, and `BatchStudy` has `optimize` method.

```python
study = optuna.batch.create_study(direction="maximize", batch_size=4)
study.optimize(objective, n_batches=10)
```

In this PR, I replaced `Study._run_trial_and_callback` with the function that supports the batch optimization. 
 `types.MethodType` is used to replace the method, and a similar approach is employed in the `MultiObjectiveStudy` to override the logging function [here](https://github.com/optuna/optuna/blob/master/optuna/multi_objective/study.py#L232-L234).

## Known issues

- The current implementation shows the log.
  - Resolved by https://github.com/optuna/optuna/pull/1729/commits/615192aa9cca5722121171bf9d349ccb01bc1f5f and https://github.com/optuna/optuna/pull/1729/commits/e39d031f3fec21acc2be4406575867562b7fd82c.
- The pruning feature may not be required for batch optimization. I think most of the users do not train multiple machine learning models simultaneously. The behavior of pruners may not be intuitive for users. Currently, it stops the evaluation of the batch trial only if all the trials in the batch should be pruned at the same time. 
  - The pruning feature was removed in https://github.com/optuna/optuna/pull/1729/commits/e73516f00db74e57ef6bb4b0133231e90d22edec.
- Do `optuna.visualization.plot_xxx` functions accept `BatchStudy`?
  - Yes, `mypy==0.782` accept it. `BatchStudy` does not inherit `Study`, but it has all `Study`'s methods. The `mypy` seems to accept such classes as `Study`.


optuna/batch/trial.py
```diff
@@ -0,0 +1,154 @@
...
+        *,
+        step: Optional[float] = None,
+        log: bool = False
+    ) -> np.ndarray:
+        return np.array(
+            [t.suggest_float(name, low, high, step=step, log=log) for t in self._get_trials()]
+        )
+
+    def suggest_uniform(self, name: str, low: float, high: float) -> np.ndarray:
+        return np.array([t.suggest_uniform(name, low, high) for t in self._get_trials()])
```
hvy, line 39: Since trials are suggested sequentially, it's not clear from this PR how a sampler would go about implementing batched sampling. How about clarifying that, how samplers are supposed to handle batched sampling. Did you imagine that samplers should be distinguished by API or logic depending on if they're "batched samplers" or not somehow? I believe the sampler could query the batch size from the study and try to compute the batch number from the trial number, but it feels a bit complicated and might not work if trial numbers aren't in sequence within a batch (I left another comment addressing this).

We should also make sure that all trials in a batch are suggested under same circumstances. This might be guaranteed by how we call `read_trials_from_remote_storage` but I'm not sure.

hvy, line 39: In my understanding, one of the motivations behind batched sampling is performance gains under vectorized computations (with e.g. `numpy`). The objective function will be able to utilize this. However, samplers will not. Maybe it's something further down the roadmap but I just thought that a "batched sampler" would ideally be vectorized as well, allocating some continuous memory for the entire batch in a single allocation instead of using batch size number of allocations and a concatenation.

optuna/batch/study.py
```diff
@@ -0,0 +1,197 @@
...
+from optuna._experimental import experimental
+import optuna.logging
+
+ObjectiveFuncType = Callable[["optuna.batch.trial.BatchTrial"], np.ndarray]
+CallbackFuncType = Callable[["optuna.study.Study", "optuna.trial.FrozenTrial"], None]
+
+_logger = optuna.logging.get_logger(__name__)
+
+
+class BatchStudy(object):
```
hvy, line 26: With MO, we also marked classes experimental in addition to `create_study` and `load_study`. How about following the same convention?

optuna/batch/study.py
```diff
@@ -0,0 +1,197 @@
...
+) -> List["optuna.trial.Trial"]:
+
+    # Sync storage once at the beginning of the objective evaluation.
+    self._storage.read_trials_from_remote_storage(self._study_id)
+
+    trials = []
+    for _ in range(batch_size):
+        trial_id = self._pop_waiting_trial_id()
+        if trial_id is None:
+            trial_id = self._storage.create_new_trial(self._study_id)
```
hvy, line 108: Not really a race condition but trial numbers within a single batch will not necessarily follow a sequence under distributed optimization. This might be suboptimal when analyzing studies with e.g. visualization when you might want to clearly see the boundaries between batches. I'm wondering if we cannot creates trials in batches as well.

optuna/batch/study.py
```diff
@@ -0,0 +1,197 @@
...
+            Size of batches.
+        """
+
+        return self._batch_size
+
+    def optimize(
+        self,
+        objective: ObjectiveFuncType,
+        timeout: Optional[int] = None,
+        n_batches: Optional[int] = None,
```
hvy, line 48: Just an idea but existing users of the framework might be more familiar with `n_trials` (and the number of batches can be derived from the total number of trials). It'd also make the interface similar the existing one and thus less to memorize.

optuna/batch/study.py
```diff
@@ -0,0 +1,197 @@
...
+    ) -> None:
+
+        n_trials = math.ceil(n_batches / n_jobs) if n_batches is not None else None
+
+        self._study._run_trial_and_callbacks = types.MethodType(  # type: ignore
+            partial(_run_trial_and_callbacks, batch_func=objective, batch_size=self._batch_size),
+            self._study,
+        )
+        self._study.optimize(
+            lambda _: 0,
```
hvy, line 63: A comment might be helpful for future readers.

optuna/batch/study.py
```diff
@@ -0,0 +1,197 @@
...
+
+@experimental("2.1.0")
+def create_study(
+    storage: Optional[Union[str, "optuna.storages.BaseStorage"]] = None,
+    sampler: Optional["optuna.samplers.BaseSampler"] = None,
+    pruner: Optional["optuna.pruners.BasePruner"] = None,
+    study_name: Optional[str] = None,
+    direction: str = "minimize",
+    load_if_exists: bool = False,
+    batch_size: int = 1,
```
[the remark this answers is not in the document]
hvy: I was even thinking of an alternative interface to batched studies that might make it easier to use for existing users (without having to use `optuna.batch` for simple operations).

```python
study = optuna.create_study()  # This is a regular study.
study.optimize(obejective, n_trials=256, batch_size=32)
```

hvy (commented):
Thanks for the very interesting PR, introducing a completely new paradigm without a single deletion from the existing codebase. I haven't went through all the lower level details, so some of my comments might be flawed, but I left some early impressions on the more higher layers.

optuna/batch/study.py
```diff
@@ -0,0 +1,197 @@
...
+        self._storage.set_trial_state(trial_id, optuna.trial.TrialState.COMPLETE)
+        self._log_completed_trial(trial, result)
+    return trials
+
+
+@experimental("2.1.0")
+def create_study(
+    storage: Optional[Union[str, "optuna.storages.BaseStorage"]] = None,
+    sampler: Optional["optuna.samplers.BaseSampler"] = None,
+    pruner: Optional["optuna.pruners.BasePruner"] = None,
```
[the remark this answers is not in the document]
toshihikoyanase: You're right. I removed the `pruner` argument in https://github.com/optuna/optuna/pull/1729/commits/ecb7ebc775206f6bbabc5381d23647dfe3473afb. I think we can add it after we find a way to apply pruning to batch optimization.

optuna/batch/study.py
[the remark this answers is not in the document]
toshihikoyanase: > I'm wondering if we cannot give the user flexibility to change the batch size between calls to `optimize`

Sounds good. Some algorithms including `IPOP CMA-ES` actually changes the batch_size (or population size). I've moved the `batch_size` argument to `optimize` method.

As for the following interface, it seems to be quite natural and user-friendly. I think we can make it by
1) moving the `_run_trial` function in `optuna/batch/study.py` into `optuna/study` and
2) relaxing the type of return value of `objective` functions.

On the other hand, this approach increases the complexity of `optuna/study.py`, and the change may include bugs that affect existing users. I'd like to isolate the changes related to bach optimization, but it's not a strong opinion.

What do you think?

```python
study = optuna.create_study()  # This is a regular study.
study.optimize(obejective, n_trials=256, batch_size=32)
```

optuna/batch/study.py
toshihikoyanase: I added it in https://github.com/optuna/optuna/pull/1729/commits/6860311582ef170021ce26a7af2f3e77a6201a63. Thanks!

optuna/batch/study.py
toshihikoyanase: That makes sense. I replaced it with `n_trials` in https://github.com/optuna/optuna/pull/1729/commits/e4cdf6c7489a56c1c165ff54399163231adb6400.

optuna/batch/study.py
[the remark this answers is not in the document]
HideakiImamura: Sounds good! But, I think editing `optuna/study.py` should be done in another PR when integrating the regular `Study` with the `BatchStudy`. After this PR has been merged, how about creating an issue about the integration of the `Study` and `BatchStudy` and including this point in it?

optuna/batch/trial.py
HideakiImamura: There is a trade-off between complicating the `suggest_*` logic of the `BatchTrial` and making the API of the `BatchSampler` different from that of a regular sampler, or complicating the internal logic of the `BatchSampler`. I prefer the latter.

If we choose the former, there are simply more areas that need to be fixed. Furthermore, I think there will be difficulties in integrating the `BatchTrial` with the regular `Trial` in the future.

On the other hand, if we choose the latter, the only thing we need to work on in this regard is the `BatchSampler` we'll implement in the future. Furthermore, the integration of `BatchTrial` and `Trial` would be possible without having to think about `BatchSampler`.

Therefore, I think the current implementation of `suggest_*` of `BatchTrial` is fine as it is

HideakiImamura: Even with the current `suggest_*` logic, we can still achieve batch sampling. The `BatchSampler` samples the number of batches and stores them internally in a queue, and then it can pop them at every `suggest_*`.

HideakiImamura (commented):
Thanks for the great PR! I've checked `optuna/batch/study.py` and `optuna/batch/trial.py`. I have several comments as follows.

I'd like to discuss how we support "conditional parameters" with batch optimization. I think we can support the conditional parameters as follows.
```python
def objective(batch_trial):
    x = batch_trial.suggest_categorical("x", ["A", "B"])

    trials = batch_trial._get_trials()
    n = len(trials)
    ret = []
    for i in range(n):
        if x[i] == "A":
            y = trials[i].suggest_float("y", -1, 1)
            ret.append(y ** 2)
        else:  # x[i] == "B"
            z = trials[i].suggest_float("z", -1, 1)
            ret.append(z + 1)

    return np.array(ret)
```
I'm not sure this is the best way to support conditional parameters, but if we can recommend this way to users, we have to
- make `BatchTrial._get_trials()` a public API and
- inform users that `BatchTrial` is different from `Trial` and how to use it in this way.

All opinions are welcome.

optuna/batch/trial.py
toshihikoyanase: @hvy @HideakiImamura Thank you for your insightful discussion.

> How about clarifying that, how samplers are supposed to handle batched sampling.

> Even with the current `suggest_* logic`, we can still achieve batch sampling.

I agree. Indeed, CMA-ES is a kind of batch algorithm, but we have implemented it without `BatchSampler`. And I chose this design to minimize the code change.

> I believe the sampler could query the batch size from the study and try to compute the batch number from the trial number

In terms of specification of the batch size, I think users can set it when they initialize the samplers in typical usages. For example, the population size of the CMA-ES is set at the initialization. So, I think the batch sampling is supported by the current Optuna, and I'd like to focus on the batch evaluation in this PR.

> We should also make sure that all trials in a batch are suggested under same circumstances. This might be guaranteed by how we call `read_trials_from_remote_storage` but I'm not sure.

Currently, I call it right before the creation of the first trial in a batch. I don't call it when I create second and later trials, and I think the sampler can get the same observation regarding the trials not included in the batch.

As for the trials in the same batch, it depends on the samplers' algorithms. For example, TPE with the constant liar will tentatively set the constant objective values (lies) to the previous trials to prevent the sampler from suggesting the same/similar parameter values.

optuna/batch/study.py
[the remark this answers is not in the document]
toshihikoyanase: > After this PR has been merged, how about creating an issue about the integration of the Study and BatchStudy and including this point in it?

Thanks. Sound reasonable.

optuna/batch/trial.py
toshihikoyanase: Thank you for your suggestion, but I think the `batched sample` is not necessary to implement the vectorized computation of parameter suggestions according to the comment of https://github.com/optuna/optuna/pull/1729#discussion_r488433123. The current design of samplers enables users to suggest trials in a batched manner if they add a parameter queue in the samplers.

optuna/batch/study.py
toshihikoyanase: > This might be suboptimal when analyzing studies with e.g. visualization when you might want to clearly see the boundaries between batches.

Can we provide such features as a post-process?
For example, 
1. Set the batch id to the 'system_attr' of each `Trial` instance,
2. After the optimization, a function sort the trials based on the batch id in the system attr and create a new study for analysis

Assigning the consecutive numbers to the trials in a batch is ideal, but I feel it might be overengineering because it requires new storage API.

toshihikoyanase:
@HideakiImamura Thank you for your insightful comment.

> I'd like to discuss how we support "conditional parameters" with batch optimization. I think we can support the conditional parameters as follows.

IMO, it would be simpler if users use standard sequential optimization interface instead of batched interface if they want to change the behavior trial by trial. As you mentioned in https://github.com/optuna/optuna/pull/1729#discussion_r488433123, the current sampler interface enables developers to implement batched sampling algorithms, and users can utilize the batch algorithms without `BatchedTrial`.

HideakiImamura:
@toshihikoyanase Thanks for the confirmation. I agree with you. We This PR introduces the "batch evaluation API". If the user wants to define the conditional search space, he/she should not use this batch evaluation API, but he/she can use the "batch sampling" without this batch evaluation API on the standard Optuna evaluation API (`Study` or `Trial`).

```python
sampler = optuna.samplers.BatchedSampler(batch_size)
study = optuna.create_study(sampler=sampler) # not `BatchStudy` but `Study`

# This objective is NOT batch evaluation. 
def objective(trial):
    # trial is not `BatchTrial` but `Trial`.

    # This is batch sampling because the "BatchedSampler" is specified.
    # "BatchedSampler" simultaneously samples the `batch_size` number of points, stores them internally and pops it in each `suggest_*`.
    x = trial.suggest_categorical("x", ["A", "B"])

    if x == "A":
        # This is batch sampling because the "BatchedSampler" is specified.
        y = trial.suggest_float("y", -1, 1)
        return y ** 2
    else:  # x == "B"
        # This is batch sampling because the "BatchedSampler" is specified.
        z = trials.suggest_float("z", -1, 1)
        return z + 1

study.optimize(objective)
```

If the user wants to use "batch sampling", "batch evaluation", and conditional search space, he/she have to use the private method `BatchTrial._get_trials()` and use it as follows. But we do not recommend this way for users, correct?

```python
sampler = optuna.samplers.BatchedSampler(batch_size)
batch_study = optuna.batch.create_study(sampler=sampler) # `BatchStudy`

def objective(batch_trial):
    # batch_trial is "BatchTrial"

    # This is batch sampling because the "BatchedSampler" is specified.
    # With the batch evaluation API, that is `BatchTrial`, we can get the `batch_size` number of points simultaneously. 
    xs = batch_trial.suggest_categorical("x", ["A", "B"])

    trials = batch_trial._get_trials()
    n = len(trials)
    ys = []
    zs = []
    for i in range(n):
        if xs[i] == "A":
            # This is batch sampling because the "BatchedSampler" is specified.
            ys.append(trials[i].suggest_float("y", -1, 1))
        else:  # xs[i] == "B"
            # This is batch sampling because the "BatchedSampler" is specified.
            zs.append(trials[i].suggest_float("z", -1, 1))
    ys = np.array(ys)
    zs = np.array(zs)

    # This is batch evaluation.
    ret = np.zeros(xs.shape)
    ret[xs == "A"] = ys ** 2
    ret[xs == "B"] = zs + 1
    return np.array(ret)

batch_study.optimize(objective)

File: tests/batch_tests/multi_objective_tests/test_study.py [hunk 8]
```diff
@@ -0,0 +1,75 @@
...
+        assert created_study._study_id == loaded_study._study_id
+
+
+@pytest.mark.parametrize("n_objectives", [1, 2, 3])
+def test_optimize(n_objectives: int) -> None:
+    directions = ["minimize" for _ in range(n_objectives)]
+    study = optuna.batch.multi_objective.create_study(directions)
+
+    def objective(trial: BatchMultiObjectiveTrial) -> List[np.ndarray]:
+        return [trial.suggest_float("v{}".format(i), 0, 5) for i in range(n_objectives)]
```
HideakiImamura, line 52: I think we should note somewhere that the shape of the returned value of the `objective` for the batch multi-objective optimization is `(n_objectives, batch_size)`, and the type of the returned value is not 2-dimensional `np.ndarray` but `List` of 1-dimensional `np.ndarray`. It is kind to users.

HideakiImamura (commented):
I have checked the all of added codes, and have several comments. Could you take a look?

PR #1730, opened 2020-08-27
Apply `black==20.8b1`

toshihikoyanase:
## Motivation
[`black`](https://pypi.org/project/black/#history) v20.8b1 has been released today, and it requires the style update of the code in the `master` branch.

## Description of the changes
This PR simply apply `black==20.8b1` to the code. Please let me know if we should use the previous version as a workaround.

### DIFF
--- a/examples/allennlp/allennlp_simple.py
+++ b/examples/allennlp/allennlp_simple.py
@@ -64,7 +64,9 @@ def create_model(vocab, trial):
     num_filters = trial.suggest_int("num_filters", 32, 128)
 
     embedding = allennlp.modules.Embedding(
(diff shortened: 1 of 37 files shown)


hvy (commented):
Thanks for your swift action. This is not a strong opinion but the magic trailing commas were seemingly unintended. So it might be cleaner to remove them instead of telling `black` to break there. What do you think?

toshihikoyanase:
@hvy Thank you for pointing it out. I was not aware of that. I'll remove the trailing commas to minimize the changes.

HideakiImamura (approved):
Thanks for the quick fix! LGTM!

hvy (approved):
Nice that it also catches bad indentations. Thanks, LGTM!

PR #1731, opened 2020-08-27
Add test for hypervolume computation for solution sets with duplicate points

hvy:
## Motivation

A solution set with duplicated points can be considered a rare edge case for hypervolume computation. It should be covered by unit tests.

## Description of the changes

Adds a test for hypervolume computation for solution sets with duplicate points.


HideakiImamura (approved):
Thank you for the PR. LGTM.

toshihikoyanase (approved):
LGTM. Thanks!

PR #1741, opened 2020-08-28
Add example to `optuna.study.delete_study`

norihitoishida:
## Motivation
This PR is related to #1334

## Description of the changes
Update the docstring of `delete_study`

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -919,6 +919,32 @@ def delete_study(
 ) -> None:
     """Delete a :class:`~optuna.study.Study` object.
 
+    Example:
+
+        .. testsetup::
+
+            import os
+
+            if os.path.exists("example.db"):
+                raise RuntimeError("'example.db' already exists. Please remove it.")
+
+        .. testcode::
+
+            import optuna
+
+            def objective(trial):
+                x = trial.suggest_float("x", -10, 10)
+                return (x - 2) ** 2
+
+            study = optuna.create_study(study_name="example-study", storage="sqlite:///example.db")
+            study.optimize(objective, n_trials=3)
+
+            optuna.study.delete_study(study_name="example-study", storage="sqlite:///example.db")
+
+        .. testcleanup::
+
+            os.remove("example.db")
+
     Args:
         study_name:
             Study's name.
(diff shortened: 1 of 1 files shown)


toshihikoyanase (commented):
Thank you for your PR. It basically looks good to me, but I'd like to know the opinions of the other reviewers.

I'm wondering if we need `RDBStorage` since it is not necessary for us to save the study unlike the `load_study`'s example.

I came up with the following example, but it may not helpful for readers. This is because it uses `optuna.storage` functions, which are not listed in the reference. What do you think?

```python
# Create in-memory storage.
storage = optuna.storages.get_storage(None)
study = optuna.create_study(study_name="example-study", storage=storage)

optuna.study.delete_study(study_name="example-study", storage=storage)
# The storage has no studies.
assert len(storage.get_all_study_summaries()) == 0
```

hvy:
Thanks @toshihikoyanase for the swift review.

I personally find the current RDB example totally fine. I assume that it's more common wanting to delete a study from an RDB storage than an in-memory storage. (`storage.get_all_study_summaries()` without any arguments seems a bit odd from a user perspective, but that might be a different topic)

toshihikoyanase:
> I assume that it's more common wanting to delete a study from an RDB storage than an in-memory storage.

That makes sense. So, I'll accept this PR.

> (`storage.get_all_study_summaries()` without any arguments seems a bit odd from a user perspective, but that might be a different topic)

I think it has no argument since it returns all study summaries in the storage (c.f., https://github.com/optuna/optuna/blob/master/optuna/storages/_base.py#L287-L295).

toshihikoyanase (approved):

optuna/study.py
```diff
@@ -919,6 +919,32 @@ def delete_study(
...
+            import optuna
+
+            def objective(trial):
+                x = trial.suggest_float("x", -10, 10)
+                return (x - 2) ** 2
+
+            study = optuna.create_study(study_name="example-study", storage="sqlite:///example.db")
+            study.optimize(objective, n_trials=3)
+
+            optuna.study.delete_study(study_name="example-study", storage="sqlite:///example.db")
```
hvy, line 942: How about using the shorthand alias similar to the doctests for `create_study` and `load_study`?
```suggestion
            optuna.delete_study(study_name="example-study", storage="sqlite:///example.db")
```

norihitoishida: Thanks! 
Your code looks better for the sake of simplicity and consistency.

hvy: Thanks for the quick fix!

hvy (approved):
LGTM!

PR #1744, opened 2020-08-29
Add example to `optuna.study.Study.set_user_attr`

norihitoishida:
## Motivation
This PR is related to #1334

## Description of the changes
Update docstring of `set_user_attr` , referred to [Tutorial](https://optuna.readthedocs.io/en/stable/tutorial/attributes.html).

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -324,16 +324,14 @@ class Study(BaseStudy):
 
         Example:
 
-            import optuna
+            .. testcode::
 
-            def objective(trial):
-                x = trial.suggest_float("x", -10, 10)
-                return (x - 2) ** 2
+                import optuna
 
-            study = optuna.create_study()
+                study = optuna.create_study()
 
-            study.set_user_attr("contributors", ["Akiba", "Sano"])
-            study.set_user_attr("dataset", "MNIST")
+                study.set_user_attr("contributors", ["Akiba", "Sano"])
+                study.set_user_attr("dataset", "MNIST")
 
         Args:
             key: A key string of the attribute.
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -322,7 +322,7 @@ class Study(BaseStudy):
     def set_user_attr(self, key: str, value: Any) -> None:
         """Set a user attribute to the study.
 
-        .. seealso
+        .. seealso::
 
             See :attr:`~optuna.study.Study.user_attrs` for related attribute.
 
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -322,6 +322,8 @@ class Study(BaseStudy):
     def set_user_attr(self, key: str, value: Any) -> None:
         """Set a user attribute to the study.
 
+        Please see the example code of the  :py:obj:`~optuna.study.Study.user_attrs`.
+
         Example:
 
             .. testcode::
@@ -335,11 +337,15 @@ class Study(BaseStudy):
 
                 study = optuna.create_study()
 
-                study.set_user_attr("objective function", "quadratic fuction")
+                study.set_user_attr("objective function", "quadratic function")
                 study.set_user_attr("dimensions", 2)
                 study.set_user_attr("contributors", ["Akiba", "Sano"])
 
-                study.user_attrs
+                assert study.user_attrs == {
+                    "objective function": "quadratic function",
+                    "dimensions": 2,
+                    "contributors": ["Akiba", "Sano"]
+                }
 
         Args:
             key: A key string of the attribute.
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -328,10 +328,18 @@ class Study(BaseStudy):
 
                 import optuna
 
+                def objective(trial):
+                    x = trial.suggest_float("x", 0, 1)
+                    y = trial.suggest_float("y", 0, 1)
+                    return x ** 2 + y ** 2
+
                 study = optuna.create_study()
 
+                study.set_user_attr("objective function", "quadratic fuction")
+                study.set_user_attr("dimensions", 2)
                 study.set_user_attr("contributors", ["Akiba", "Sano"])
-                study.set_user_attr("dataset", "MNIST")
+
+                study.user_attrs
 
         Args:
             key: A key string of the attribute.
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -322,6 +322,20 @@ class Study(BaseStudy):
     def set_user_attr(self, key: str, value: Any) -> None:
         """Set a user attribute to the study.
 
+        Example:
+
+            import optuna
+
+            def objective(trial):
+                x = trial.suggest_float("x", -10, 10)
+                return (x - 2) ** 2
+
+            study = optuna.create_study()
+            study.optimize(objective, n_trials=3)
+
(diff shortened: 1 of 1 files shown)


optuna/study.py
```diff
@@ -322,6 +322,19 @@ def _should_stop() -> bool:
...
         """Set a user attribute to the study.
 
+        Example:
+
+            import optuna
+
+            def objective(trial):
+                x = trial.suggest_float("x", -10, 10)
+                return (x - 2) ** 2
+
```
HideakiImamura, line 332: This objective is not used. How about removing it?

optuna/study.py
```diff
@@ -322,6 +322,19 @@ def _should_stop() -> bool:
     def set_user_attr(self, key: str, value: Any) -> None:
         """Set a user attribute to the study.
 
+        Example:
```
HideakiImamura, line 325: How about marking this example as a doctest by adding `.. testcode::`?

HideakiImamura (commented):
Thanks for the PR. I have some minor comments.

optuna/study.py
norihitoishida: Thanks for swift review and nice idea!

optuna/study.py
norihitoishida: Thanks for swift review!
I agree with you!

HideakiImamura (approved):

optuna/study.py
```diff
@@ -322,6 +322,17 @@ def _should_stop() -> bool:
...
+
+            .. testcode::
+
+                import optuna
+
+                study = optuna.create_study()
+
+                study.set_user_attr("contributors", ["Akiba", "Sano"])
+                study.set_user_attr("dataset", "MNIST")
```
toshihikoyanase, line 334: This comment is originally posted to #1745.
---

I think these values are taken from [this tutorial](https://optuna.readthedocs.io/en/stable/tutorial/attributes.html), It may be difficult for readers to imagine the actual usage. I guess they describe the objective function, but this example doesn't have objective functions. So, for example, how about adding a simple objective function and explaining it with the `user_attrs`?

```python
def objective(trial):
    return x ** 2 + y ** 2

...

study.set_user_attr("objective function", "quadratic fuction")
study.set_user_attr("dimensions", 2)
study.set_user_attr("contributors", ["Akiba", "Sano"])
```

What do you think of it? @HideakiImamura @norihitoishida

toshihikoyanase (commented):
Sorry for my carelessness, but I added the comment to #1745. I just copied it to this PR.

optuna/study.py
norihitoishida: I agree with you. Thank you!

HideakiImamura (approved):
@toshihikoyanase Thanks for the suggestion. It is reasonable to me, and LGTM again!

optuna/study.py
```diff
@@ -322,6 +322,25 @@ def _should_stop() -> bool:
...
+                import optuna
+
+                def objective(trial):
+                    x = trial.suggest_float("x", 0, 1)
+                    y = trial.suggest_float("y", 0, 1)
+                    return x ** 2 + y ** 2
+
+                study = optuna.create_study()
+
+                study.set_user_attr("objective function", "quadratic fuction")
```
HideakiImamura, line 338: A typo: 
```suggestion
                study.set_user_attr("objective function", "quadratic function")
```

HideakiImamura (changes requested):
Sorry for the review after the approval. I found a typo. Could you take a look?

optuna/study.py
```diff
@@ -322,6 +322,25 @@ def _should_stop() -> bool:
...
+                study.set_user_attr("dimensions", 2)
+                study.set_user_attr("contributors", ["Akiba", "Sano"])
+
+                study.user_attrs
```
HideakiImamura, line 342: As the same as here: https://github.com/optuna/optuna/pull/1745#discussion_r479732808, how about checking the contents?

optuna/study.py
```diff
@@ -322,6 +322,25 @@ def _should_stop() -> bool:
     def set_user_attr(self, key: str, value: Any) -> None:
         """Set a user attribute to the study.
 
+        Example:
+
```
HideakiImamura, line 326: It might be kind for users to note that "Please see the example code of the `user_attrs` method." What do you think?

optuna/study.py
norihitoishida: Thanks!

optuna/study.py
norihitoishida: Nice idea!

toshihikoyanase: @HideakiImamura 
I'm sorry, but I couldn't get the point. If users follow this instruction, they will see the same example as the example in this PR. Do you intend to remove this duplicated PR and guide users to the example in the `user_attrs` section?

norihitoishida: Sorry. 
I will do this:
- Add detailed explanation of this function, referred to [Tutorial](https://optuna.readthedocs.io/en/latest/tutorial/006_attributes.html).
- Note that "Please see the example code of the `user_attrs` method" into `Study.user_attrs`(#1745).

HideakiImamura: Thanks for the suggestion, @norihitoishida! Actually, the content of the example of `Study.set_user_attr` and `Study.user_attr` are same. Therefore, I initially thought that relating these two examples is reasonable. However, we don't have to explain these two examples are related here. Now, I think there is nothing that we should do here. Sorry, fo the confusion.

norihitoishida: @HideakiImamura 
Thanks for detail, I got it!

optuna/study.py
```diff
@@ -322,6 +322,31 @@ def _should_stop() -> bool:
     def set_user_attr(self, key: str, value: Any) -> None:
         """Set a user attribute to the study.
 
+        Please see the example code of the  :py:obj:`~optuna.study.Study.user_attrs`.
```
HideakiImamura, line 325: How about using `.. seealso::` samely as other methods?
```suggestion
        .. seealso
        
            See :attr:`~optuna.study.Study.user_attrs` for related attribute.
```

HideakiImamura (commented):
@norihitoishida Thanks for the swift actions! I have a minor comment. Could you take a look?

optuna/study.py
norihitoishida: @HideakiImamura  Thanks for nice idea!

optuna/study.py
```diff
@@ -322,6 +322,33 @@ def _should_stop() -> bool:
     def set_user_attr(self, key: str, value: Any) -> None:
         """Set a user attribute to the study.
 
+        .. seealso
```
HideakiImamura, line 325: A nit.
```suggestion
        .. seealso::
```

HideakiImamura (commented):
Thanks for the swift action! I have a minor comment.

optuna/study.py
norihitoishida: @HideakiImamura Thanks! (I made same fault at #1745)

HideakiImamura (approved):

The docs looks like this.
スクリーンショット 2020-09-03 0 02 04

toshihikoyanase (approved):
LGTM. Thank you for your contribution!

PR #1746, opened 2020-08-29
Add example to `optuna.study.Study.get_trials` 

norihitoishida:
## Motivation
This PR is related to #1334

## Description of the changes
Update the docstring of `Study.get_trials()`

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -114,6 +114,19 @@ class BaseStudy(object):
         For library users, it's recommended to use more handy
         :attr:`~optuna.study.Study.trials` property to get the trials instead.
 
+        Example:
+            .. testcode::
+                import optuna
+
+                def objective(trial):
+                    x = trial.suggest_uniform("x", -1, 1)
+                    return x ** 2
+
+                study = optuna.create_study()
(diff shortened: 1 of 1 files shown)


optuna/study.py
```diff
@@ -114,6 +114,20 @@ def get_trials(self, deepcopy: bool = True) -> List[FrozenTrial]:
...
+
+                def objective(trial):
+                    x = trial.suggest_uniform("x", -1, 1)
+                    return x ** 2
+
+                study = optuna.create_study()
+                study.optimize(objective, n_trials=3)
+
+                trials = study.get_trials()
+
```
toshihikoyanase, line 130: How about adding an assertion like this? It shows that `get_trials` returns all trial clearly.
```suggestion
                trials = study.get_trials()
                assert len(trials) == 3
```

toshihikoyanase (commented):
Thank you for your PRs as always. It basically looks good to me, but I added a suggestion.

optuna/study.py
norihitoishida: Thanks a lot!

hvy (approved):
LGTM!

PR #1752, opened 2020-08-29
Add example to `optuna.study.Study.stop` 

Ruketa:
## Motivation
This PR is related to #1334

## Description of the changes
Add the example for optuna.study.Study.stop

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -481,6 +481,23 @@ class Study(BaseStudy):
         spawned finishes.
         This method does not affect any behaviors of parallel or successive study processes.
 
+        Example:
+
+            .. testcode::
+
+                import optuna
+
+                study = optuna.create_study()
+
+                def objective(trial):
+                    if trial.number == 4:
+                        study.stop()
+                    x = trial.suggest_uniform("x", 0, 10)
+                    return x ** 2
+
+                study.optimize(objective, n_trials=10)
+                assert len(study.trials) == 5
+
         Raises:
             RuntimeError:
                 If this method is called outside an objective function or callback.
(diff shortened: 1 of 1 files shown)


optuna/study.py
```diff
@@ -481,6 +481,23 @@ def stop(self) -> None:
...
+
+                study = optuna.create_study()
+
+                def objective(trial):
+                    if trial.number == 4:
+                        study.stop()
+                    x = trial.suggest_uniform("x", 0, 10)
+                    return x ** 2
+
+                study.optimize(objective, n_trials=10)
```
hvy, line 498: A nit. Not really a convention but for consistency with other code examples.
```suggestion
                def objective(trial):
                    if trial.number == 4:
                        study.stop()
                    x = trial.suggest_uniform("x", 0, 10)
                    return x ** 2
                    
                study = optuna.create_study()
                study.optimize(objective, n_trials=10)
```

optuna/study.py
```diff
@@ -481,6 +481,23 @@ def stop(self) -> None:
...
+                assert len(study.trials) == 5
```
hvy, line 499: Great that you added an assertion demonstrating the effects of `stop`. 👍

hvy (changes requested):
Thanks for the PR. The usage of `stop` isn't obvious so this'll absolutely help users out. I left one minor comment.

This is by the way just a comment and not a change request but it might be more interesting to stop on the optimization based on the objective value rather than the number of trials (since this can be specified in `optimize`).

HideakiImamura (approved):
Thanks for the PR! LGTM except for @hvy's comment.

optuna/study.py
Ruketa: Thank you for reviewing PR. I fixed it.

hvy (approved):
LGTM, thank you!

PR #704, opened 2019-11-14
Make plot functions return Plotly graph object.

c-bata:
## Summary

I want to plot intermediate values with log-scale yaxis. I once designed APIs of plotting utilities like `plot_intermediate_values(study, yaxis_logscale=True)` but this approach might need a lot of parameters if we want to provide more flexible options. In this PR, I added option to pass parameters into `update_layout` method.

## Usage

Prepare a dummy study.

```python
import optuna

study = optuna.create_study()

values = [
    [10000, 1000, 100, 10, 1, 0.1, 0.01],
    [7500, None],
    [5000, 500, 50, 5, 0.5, 0.05, 0.005],
    [6000, None],
]
    
for intermediate_values in values:
    trial_id = study._storage.create_new_trial(study.study_id)
    trial = optuna.trial.Trial(study, trial_id)

    for i, v in enumerate(intermediate_values):
        if v is None:
            study._storage.set_trial_state(trial_id, optuna.structs.TrialState.PRUNED)
            break
        trial.report(v, step=i)
    else:
        study._storage.set_trial_state(trial_id, optuna.structs.TrialState.COMPLETE)
```

Then plotting without params:

```python
from optuna.visualization import plot_intermediate_values

plot_intermediate_values(study)
```

newplot (2)

Then plotting in log-scale yaxis:

```python
from optuna.visualization import plot_intermediate_values

plot_intermediate_values(study, {'yaxis_type': 'log'})
```

newplot (1)

PR #1756, opened 2020-08-30
Create `optuna.visualization.matplotlib`.

ytknzw:
This PR proposes to implement ``matplotlib`` backend for plotting functions as ``optuna.visualization.matplotlib``.

## Motivation
See issue #1539 and PR #1707 for the original motivation.

## Description of the changes
Create ``optuna.visualization.matplotlib``. All the functions corresponding to the current Plotly backend are to be implemented in ~PR #1707~ the following PRs.


optuna/visualization/matplotlib/__init__.py
```diff
@@ -0,0 +1,8 @@
+from optuna.visualization.matplotlib._contour import plot_contour  # NOQA
```
HideakiImamura, line 1: I'm not sure, but how about not listing functions that are not yet implemented? When users look at the documentation, they may be confused by the lack of documentation or by the unnatural type of functions.

optuna/visualization/matplotlib/_utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+    .. note::
+
+        :mod:`~optuna.visualization` module depends on plotly version 4.0.0 or higher. If a
+        supported version of plotly isn't installed in your environment, this function will return
+        :obj:`False`. In such case, please execute ``$ pip install -U plotly>=4.0.0`` to install
+        plotly.
+
+    Returns:
+        :obj:`True` if visualization is available, :obj:`False` otherwise.
```
HideakiImamura, line 21: Could you update the docstring to match the `matplotlib`?
```suggestion
    """Returns whether visualization with `matplotlib` is available or not.

    .. note::

        :mod:`~optuna.visualization` module depends on matplotlib version 3.0.0 or higher. If a
        supported version of matplotlib isn't installed in your environment, this function will return
        :obj:`False`. In such a case, please execute ``$ pip install -U matplotlib>=3.0.0`` to install
        matplotlib.

    Returns:
        :obj:`True` if visualization with `matplotlib` is available, :obj:`False` otherwise.
```

HideakiImamura (commented):
Thanks for the PR! I have several comments. Could you take a look?

optuna/visualization/matplotlib/__init__.py
HideakiImamura: I would like to discuss this point with @ytknzw and other reviewers. Thank you!

c-bata: > I'm not sure, but how about not listing functions that are not yet implemented?

I agree with his suggestion.

c-bata:
How about retuning matplotlib's Axes object like `pandas.DataFrame.plot()` function. It makes more customizable (related discussion is https://github.com/optuna/optuna/pull/704).

https://pandas.pydata.org/pandas-docs/version/0.23.4/generated/pandas.DataFrame.plot.html

c-bata:
> Some approaches for how we could go about designing the public API. 
> 1. Introduce a “backend” parameter to existing visualization functions.
>     - User must consider the argument in each call, implementation becomes somewhat complicated, and the return value becomes a `Union`.
> 2. Define a global/thread local state. 
>     - User only have to configure it once. Other cons remain. We introduce a global state.
> 3. Introduce `optuna.visualization.matplotlib` to define `plot_*`
>     - User switched between backends by changing the top import statement(s). Pro that we don’t have to keep a global state, code is clearly separated. What functions are implemented in what backend becomes very clear. Return value is not a `Union`.
> 
> I would suggest 3., as you might have noticed. And if so, changes required from this PR are not that large and we could incorporate them here.

https://github.com/optuna/optuna/pull/1707#issuecomment-681335837

In terms of public API's design, I'm also +1 to introduce `optuna.visualization.matplotlib` like this pull request.

optuna/visualization/matplotlib/__init__.py
ytknzw: I agree with you two. I didn't think well that part, sorry.

ytknzw:
@HideakiImamura @c-bata 
Thank you for your reviews and comments!
I've made some changes to the codes.

ytknzw:
@c-bata 
https://github.com/optuna/optuna/pull/1756#issuecomment-683383976
> How about retuning matplotlib's Axes object like pandas.DataFrame.plot() function.

Thank you for your suggestion. I assume you mean 'return' not 'retune', right?
My current Matplotlib implementation returns Figure object. For compatibility with the Plotly implementation, I'll work on this in PR #1707 as well.

c-bata:
Sorry for my delayed response :pray:

> Thank you for your suggestion. I assume you mean 'return' not 'retune', right?

Exactly! It was just a typo.

> My current Matplotlib implementation returns Figure object. For compatibility with the Plotly implementation, I'll work on this in PR #1707 as well.

IMO, it's better to type returned value to `matplotlib.Axes` in this pull request to avoid interface breaking changes.
What do you think?

toshihikoyanase:
> How about retuning matplotlib's Axes object like `pandas.DataFrame.plot()` function.

Sounds good. When we released the current visualization functions, some users requested to customize the figures. IMO, it is reasonable to provide such a feature for matplotlib backend.

@ytknzw 
Could you change the return type of `plot_xxx` in this PR, too?

ytknzw:
@c-bata @toshihikoyanase 
Thank you for your suggestions. Yes, I agree with you.
Just changed the return value type to `Axes`.

HideakiImamura (approved):
LGTM!

c-bata (approved):
LGTM after passed CI 👍

optuna/visualization/matplotlib/_utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+from optuna.distributions import LogUniformDistribution
+from optuna.trial import FrozenTrial
+from optuna.visualization.matplotlib import _matplotlib_imports
+
+__all__ = ["is_available"]
+
+
+def is_available() -> bool:
+    """Returns whether visualization with `matplotlib` is available or not.
```
toshihikoyanase, line 11: A nit, but a single backtick corresponds to italic. Please use double backticks for consistency. 
```suggestion
    """Returns whether visualization with ``matplotlib`` is available or not.
```

optuna/visualization/matplotlib/_utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+    .. note::
+
+        :mod:`~optuna.visualization` module depends on Matplotlib version 3.0.0 or higher. If a
```
toshihikoyanase, line 15: `optuna.visualization.plot_xx` does not depend on Matplotlib. Could you update the module path?

```suggestion
        :mod:`~optuna.visualization.matplotlib` module depends on Matplotlib version 3.0.0 or higher. If a
```

optuna/visualization/matplotlib/_utils.py
```diff
@@ -0,0 +1,33 @@
...
+        supported version of Matplotlib isn't installed in your environment, this function will
+        return :obj:`False`. In such a case, please execute ``$ pip install -U matplotlib>=3.0.0``
+        to install Matplotlib.
+
+    Returns:
+        :obj:`True` if visualization with `matplotlib` is available, :obj:`False` otherwise.
```
toshihikoyanase, line 21: ditto.
```suggestion
        :obj:`True` if visualization with ``matplotlib`` is available, :obj:`False` otherwise.
```

toshihikoyanase (changes requested):
Thank you for your update. I added some small comments. I may be able to work on them as a follow-up task.

toshihikoyanase (approved):
Thank you for your swift action. LGTM. I'm waiting for the CI jobs to merge this PR.

PR #1757, opened 2020-08-30
Fix Binh and Korn function in MO example

hvy:
## Motivation

See description below.
 
## Description of the changes

Fixes the Bing and Korn function in one of our MO examples. C.f. [Wikipedia]( https://en.wikipedia.org/wiki/Test_functions_for_optimization#Test_functions_for_multi-objective_optimization).

### DIFF
--- a/examples/multi_objective/plot_pareto_front.py
+++ b/examples/multi_objective/plot_pareto_front.py
@@ -6,7 +6,7 @@ def objective(trial):
     x = trial.suggest_float("x", 0, 5)
     y = trial.suggest_float("y", 0, 3)
 
-    v0 = (4 * x) ** 2 + (4 * y) ** 2
+    v0 = 4 * x ** 2 + 4 * y ** 2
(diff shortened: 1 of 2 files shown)


sile:
Thank you for catching and fixing the bug!
I think that we need to fix [the example code of `optuna.multi_objective.visualization.plot_pareto_front()`](https://github.com/optuna/optuna/blob/master/optuna/multi_objective/visualization/_pareto_front.py#L35) too.

toshihikoyanase (approved):
Good catch! It is identical to the equation in the original paper by Binh and Korn.

image

hvy:
> Thank you for catching and fixing the bug!
I think that we need to fix the example code of optuna.multi_objective.visualization.plot_pareto_front() too. 

Should be fixed now as well. PTAL.

sile (approved):
Thanks!
Let me merge this PR once CI passes.

PR #1716, opened 2020-08-25
Add padding to contour plot

zchenry:
## Motivation
As mentioned in https://github.com/optuna/optuna/issues/1697, it is difficult to see when points appear at the borders.

## Description of the changes
I checked available parameters of `plotly.graph_objects.Contour` but failed to find anything useful to introduce a padding. Although `plotly.graph_objects.Figure` has a `margin` parameter, it is used for the whole figure, not the plot(s) inside it.

To this end, I choose to use an empirical padding ratio of `0.05` (not sure if it is appropriate) and manually calculate the corresponding paddings.

Plot for the following code is attached.

```python
import optuna

def objective(trial):
    x = trial.suggest_float("x", 0, 1)
    y = trial.suggest_int("y", -1, 1)
    z = trial.suggest_loguniform("z", 1e-2, 1e3)
    return (x - 0.5) ** 2 + y + z / 100

study = optuna.create_study()
study.optimize(objective, n_trials=200)

optuna.visualization.plot_contour(study).show()
```

padding

PR #1758, opened 2020-08-30
Re-generate contour plot example with padding

hvy:
## Motivation

Follow-up to https://github.com/optuna/optuna/pull/1716 updating the docs according to the new logic.

## Description of the changes

Re-generates the contour plot html with the new logic Also increases the number of trials to make it more interesting.


HideakiImamura (approved):
LGTM!

toshihikoyanase (approved):
LGTM!

Issue #1226, opened 2020-05-11
Bad monitoring metrics with integration callbacks are difficult to notice.


hvy:
## Motivation

Bad/wrong monitoring metrics represented by key strings with the following integration callbacks are difficult to notice. For instance, a simple typo in the metric would make the pruner a NOOP pruner. One might expect an error.

- https://github.com/optuna/optuna/blob/master/optuna/integration/keras.py (resolved by #1759)
- https://github.com/optuna/optuna/blob/master/optuna/integration/tfkeras.py (resolved by #2156)
- https://github.com/optuna/optuna/blob/master/optuna/integration/pytorch_lightning.py (resolved by #2157, ongoing)

## Description

Above callbacks are not doing dictionary lookups that raise on failure but uses `dict.get`. I'm wondering if we cannot raise instead as that'd clearly communicate to the user that the pruner isn't doing anything. However, I'm not too familiar with all of the above libraries and it might not be possible.

hvy:
This is still an issue.

toshihikoyanase:
Keras pruning callback was fixed by @bigbird555 in #1759.
The callbacks of [tfkeras](https://github.com/optuna/optuna/blob/master/optuna/integration/tfkeras.py) and [pytorch lightning](https://github.com/optuna/optuna/blob/master/optuna/integration/pytorch_lightning.py) are still waiting for your contribution.

nzw0301:
@bigbird555 Kudos to your all contributions to this issue! 

---


Do we have something to do related to this issue? Otherwise, I suppose that we can close this isseu.

hvy:
You're right, let me close it. Thanks @nzw0301 for the heads up and again, thanks @bigbird555 !

PR #1759, opened 2020-08-30
Add a warning message for the cases where an evaluation metric does n…

bigbird555:
…ot exist.

## Motivation

## Description of the changes


optuna/integration/keras.py
```diff
@@ -57,6 +57,12 @@ def on_epoch_end(self, epoch: int, logs: Optional[Dict[str, float]] = None) -> N
         logs = logs or {}
         current_score = logs.get(self._monitor)
         if current_score is None:
+            message = (
+                "Warning: there is not an evaluation metric for pruning named "
+                + self._monitor
+                + " ."
+            )
+            print(message)
```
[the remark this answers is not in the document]
HideakiImamura: If we use `warnings.warn()`, the warning message can be simplified as follows: "There is not an evaluation metric for pruning named {}"

HideakiImamura (changes requested):
Thanks for the PR! I totally agree with your idea to warn when the invalid metric is specified. The change looks basically good to me. I have some minor comments.

bigbird555:
Thank you for your always thoughtful comments.
I completely understood the matter around coding conventions.
I'm going to follow your advice and fix the code.

optuna/integration/keras.py
```diff
@@ -57,6 +57,10 @@ def on_epoch_end(self, epoch: int, logs: Optional[Dict[str, float]] = None) -> N
         logs = logs or {}
         current_score = logs.get(self._monitor)
         if current_score is None:
+            message = "There is not an evaluation metric for pruning named {}.".format(
+                self._monitor
+            )
+            warnings.warn(message)
```
toshihikoyanase, line 63: Please import `warnings` to use it.

@hvy  I think you expect an error here instead of `warnings` in https://github.com/optuna/optuna/issues/1226#issue-615516601. I guess raising the errors may break user code if the Keras returns `None` as expected values. But, I'm not sure if it occurs.
What do you think of it?

bigbird555: Thank you for your reply!
I just noticed this mistake and fixed it.

hvy: Thanks @toshihikoyanase for pinging. I think the current design to yield a warning sound good. You're right about the compatibility issue, and I also believe we have to be more careful if we are to raise errors, i.e. make sure that we understand the behavior of the integration framework.

toshihikoyanase: @hvy Thank you for your reply. That makes sense. I'll accept `warning`.

optuna/integration/keras.py
```diff
@@ -1,6 +1,8 @@
 from typing import Dict
 from typing import Optional
 
+import warnings
+
```
toshihikoyanase, line 5: Please remove the blank line since `warnings` is a standard library.
```suggestion
import warnings

```

optuna/integration/keras.py
```diff
@@ -57,6 +59,10 @@ def on_epoch_end(self, epoch: int, logs: Optional[Dict[str, float]] = None) -> N
         logs = logs or {}
         current_score = logs.get(self._monitor)
         if current_score is None:
+            message = "There is not an evaluation metric for pruning named {}.".format(
+                self._monitor
+            )
+            warnings.warn(message)
```
toshihikoyanase, line 65: Could you add a test case to check this warning to `tests/integration_tests/test_keras.py`?
We can use `with pytest.warns(UserWarning)` to assert the warning.

toshihikoyanase (changes requested):
Thank you for your swift action. I added two comments.

toshihikoyanase (commented):
Thank you for your update. I suggest an additional change. I'd like to hear your opinions about it.

optuna/integration/keras.py
```diff
@@ -57,6 +58,10 @@ def on_epoch_end(self, epoch: int, logs: Optional[Dict[str, float]] = None) -> N
         logs = logs or {}
         current_score = logs.get(self._monitor)
         if current_score is None:
+            message = "There is not an evaluation metric for pruning named {}.".format(
```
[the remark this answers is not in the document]
bigbird555: I've changed the message to follow the feedback. It would be helpful if you could check it out.

optuna/integration/keras.py
```diff
@@ -57,6 +58,12 @@ def on_epoch_end(self, epoch: int, logs: Optional[Dict[str, float]] = None) -> N
         logs = logs or {}
         current_score = logs.get(self._monitor)
         if current_score is None:
+            message = (
+                "The metric '{}' is not in the evaluation logs for pruning. "
+                "'{}' may not exist as evaluation metric for pruning. "
+                "Or '{}' may be a typo.".format(self._monitor, self._monitor, self._monitor)
```
toshihikoyanase, line 64: IMO, this message is a bit verbose because `self._monitor` appears three times in it. Can we simplify it, for example, as follows:

> "The metric '{}' is not in the evaluation logs for pruning. Please make sure you set the correct metric name."

bigbird555: If brevity is the priority, that seems better. I will revise the message that way.

HideakiImamura (approved):
Sorry for the late review. Thanks for the update! LGTM!

toshihikoyanase (approved):
I'm sorry for the delayed response. The change looks good to me. Thank you for your continuous contribution.

If you also apply similar changes to callbacks for tf.keras and PyTorch Lightning (#1226), it would be a great help.

PR #1761, opened 2020-08-30
Add the learning curve prediction method by Akshay

nabenabe0928:
## Motivation
This PR is motivated by the paper issued by Akshay Chandrashekaran named [Speeding up Hyper-parameter Optimization by Extrapolation of Learning Curves using Previous Builds](http://akshayc.com/publications/ecml2017_gelc_submitted.pdf).

As another learning curve prediction (LCP) method, the following methods are famous:
1. [LCP using Bayesian Neural Networks](https://openreview.net/forum?id=S11KBYclx): Aaron Klein et al.
2. [LCP using RankNet](https://proceedings.icml.cc/static/paper_files/icml/2020/2472-Paper.pdf): Martin Wistuba and Tejaswini Pedapati
3. [LCP using the linear combination of multiple types of non-linear functions](https://www.aaai.org/ocs/index.php/IJCAI/IJCAI15/paper/view/11468/11222): Tobias Domhan et al.

The main problem of the aforementioned methods is the time required for the inference at each iteration. Especially, the Bayesian neural networks spent so much time on Monte-Carlo sampling, which often exceeds the time needed for training 1 Epoch.

For this reason, I introduced the model proposed by Akshay. This method uses the affine transformation of the learning curves and predicts the performance at the final epoch using the mean of the transformed learning curves. Since this method only approximates the 2 parameters x the number of the previous builds (usually 5 builds), it takes much less time to suggest the judgment of the early-stopping.

## Description of the changes
For this implementation, I referred to the experimental codes for the paper. The codes I referred to are the followings:
1. [The inference part](https://github.com/akshayc11/Spearmint/blob/elc/spearmint/pylrpredictor/prevonlyterminationcriterion.py)
2. [The gradient descent part](https://github.com/akshayc11/Spearmint/blob/elc/spearmint/pylrpredictor/gradient_descent.py)

The main changes from the original code are that I refactored the codes and made it possible to infer the future performance. HOWEVER, since I have not tested the whole code yet, still some tests are required.   

## ToDo
- [x] Test the code by running on the local environment
- [ ] Visualize learning curves and validate the performance
- [ ] Check if the code works properly on Optuna


HideakiImamura:
@nabenabe0928 Thanks for the PR and the private discussion. What about the implementation after that? If you have some troubles, I would like to help you.

PR #1767, opened 2020-08-31
Add multivariate TPE sampler

kstoneriv3:
## Motivation

This PR introduces a multivariate TPE sampler. The multivariate TPE sampler captures dependencies among hyperparameters better than the current implementation. Since the current `TPESampler` treats each hyperparameter independently, it cannot handle the dependencies among hyperparameters. 

## Description of the changes

`sample_relative` method is added to `TPESampler`. Users just have to turn on the `multivariate` flag when defining the TPE sampler as following.
```python3
# Example: Optimization of a quadratic funtion
import optuna

def objective(trial):
    x = trial.suggest_uniform('x', -1, 1)
    y = trial.suggest_uniform('y', -1, 1)
    return x**2 + y**2

sampler = optuna.samplers.TPESampler(multivariate=True)  # use multivariate TPE sampler
```

Major changes to `optuna.samplers._tpe.sampler` are made at 
- `TPESampler.infer_relative_search_space` 
- `TPESampler.sample_relative`
- `_split_multivariate_observation_pairs`
- `_get_multivariate_observation_pairs`. 

Most of the functionalities are separated in a new file `optuna.samplers._tpe.multivariate_parzen_estimator`. This separated file contains `_MultivariateParzenEstimator` class, which supports sampling and log_prob computation for given multivariate inputs.

## (Updated) Benchmarks:
I compared the multivariate TPE sampler with the current independent sampler. The benchmark was taken on four problems with 100 experiments per sampler.

hpo-bench-naval-4b38e028c25f65733b05b737c70a0401664a4bab047ab08e1c5e2395ed1e9705
hpo-bench-parkinson-996baeee92a729fabb24c4d9b498bdb1477f2f88f9f85552a0b7981d7501c078
hpo-bench-protein-cbf8512fed27dedfec67a7e80c0921d6ef737a1e902569a7bac60b82e47ceab2
hpo-bench-slice-63956311448c8b8fafcdef5dcba69c838349546715298e00e22cdb8c7a48c655

## TODOs:

- [x] Take a benchmark to compare this with the current implementation.
- [x] Improve the bandwidth selection algorithm for a multivariate parzen estimator.
- [ ] Consider separating the test code `optuna.tests.samplers_tests.tpe_tests.test_sampler` for `TPESampler` as it became lengthy.


optuna/samplers/_tpe/sampler.py
```diff
@@ -233,6 +310,38 @@ def _split_observation_pairs(
...
+
+        # TODO(kstoneriv3): Change of the order of 1. exclusion of `None` and 2. splitting might
+        # be needed. Independent sampler in `TPESampler` first splits the observations and
+        # then exclude `None`.
+
+        # We exclude param_vals with `None`.
+        config_vals_matrix = np.array([v for v in config_vals.values()])
+        index_none = np.any(np.equal(config_vals_matrix, None), axis=0)
+        config_vals = {k: v[~index_none] for k, v in config_vals.items()}
+        loss_vals = loss_vals[~index_none]
```
HideakiImamura, line 330: Here, we exclude the trials which have a `None` among any parameters. I think that this means that we cannot sample based on the case studies. As a default, the search space is the `IntersectionSearchSpace`, so it is OK in this PR not to think the case-based sampling, but I think it is not a good idea to prohibit the case-based sampling in the revel of the sampler logic. (It should be in the revel of the search space logic.)  Improving this strategy can be an important future work. How about mentioning it as comments?
```suggestion
        # We exclude trials with `None` values among any parameters. 
        # This means that we cannot sample based on case studies.
        # Improving this strategy is a future work.
        config_vals_matrix = np.array([v for v in config_vals.values()])
        index_none = np.any(np.equal(config_vals_matrix, None), axis=0)
        config_vals = {k: v[~index_none] for k, v in config_vals.items()}
        loss_vals = loss_vals[~index_none]
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -233,6 +310,38 @@ def _split_observation_pairs(
...
+        config_vals: Dict[str, List[Optional[float]]],
+        loss_vals: List[Tuple[float, float]],
+    ) -> Tuple[Dict[str, np.ndarray], Dict[str, np.ndarray]]:
+
+        config_vals = {k: np.asarray(v, dtype=float) for k, v in config_vals.items()}
+        loss_vals = np.asarray(loss_vals, dtype=[("step", float), ("score", float)])
+
+        # TODO(kstoneriv3): Change of the order of 1. exclusion of `None` and 2. splitting might
+        # be needed. Independent sampler in `TPESampler` first splits the observations and
+        # then exclude `None`.
```
HideakiImamura, line 324: I don't know why the order of "excluding None" and "splitting" is changed. Could you explain the reason for this? The current one leads to the small estimation of `n_below`.

HideakiImamura (changes requested):
Thanks for the great PR! I'm very looking forward to the multivariate sampling of TPE in Optuna. I have several comments on `_tpe/sampler.py`. Could you take a look?

optuna/samplers/_tpe/sampler.py
```diff
@@ -119,6 +122,8 @@ def objective(trial):
             for more details.
         seed:
             Seed for random number generator.
+        multivariate:
+            Use multivariate TPE when suggesting parameters.
```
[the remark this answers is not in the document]
kstoneriv3: Thank you for your suggestions. That makes sense!

optuna/samplers/_tpe/sampler.py
```diff
@@ -532,6 +641,28 @@ def _compare(cls, samples: np.ndarray, log_l: np.ndarray, log_g: np.ndarray) ->
...
+            if sample_size != score.size:
+                raise ValueError(
+                    "The size of the 'samples' and that of the 'score' "
+                    "should be same. "
+                    "But (samples.size, score.size) = ({}, {})".format(sample_size, score.size)
+                )
+            best = np.argmax(score)
+            return {k: v[best] for k, v in multivariate_samples.items()}
+        else:
+            return {k: np.asarray([]) for k in multivariate_samples.keys()}
```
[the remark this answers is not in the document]
kstoneriv3: Sounds good. `_compare_multivariate` is always given input of length `_n_ei_candidates`, which is greater than 0.

optuna/samplers/_tpe/sampler.py
kstoneriv3: There are pros and cons to this implementation.

### Pros of exclusion of `None` -> splitting
- simpler code (exclusion for both `above` and `below` makes repeated code, so probably not good for maintenance)
- avoiding `below` to have zero samples (in the case all samples of `below` contains `None` we basically sample from prior)
### Cons of exclusion of `None` -> splitting
- lead to small estimation of n_below if the effective number of samples (samples that do not contain None) is small
- different from current implementations of univariate cases

I do not mean to insist on my implementation, but I would like to hear opinions on this to see which is more suitable.

hvy (commented):
Great work! Still early in the review as the change are quite large but I left some comments if you could take a look.

optuna/samplers/_tpe/sampler.py
```diff
@@ -132,6 +142,8 @@ def __init__(
         gamma: Callable[[int], int] = default_gamma,
         weights: Callable[[int], np.ndarray] = default_weights,
         seed: Optional[int] = None,
+        *,
+        multivariate: bool = False
```
hvy, line 146: Seems like a good interface for exposing this feature to the user, not introducing any new concepts besides an argument 👍

optuna/samplers/_tpe/sampler.py
HideakiImamura: Another idea of comments:
"We exclude trials with `None` values among any parameters. This means that we transfer the search space into the instance of `InterSectionSpace`."

optuna/samplers/_tpe/sampler.py
HideakiImamura: Thanks for the clarification. Your implementation is reasonable to me. By the way, currently, we only support the `IntersectionSearchSpace` as a built-in search space. So, this is not so big a matter. I think we can accept the current implementation.

optuna/samplers/_tpe/sampler.py
```diff
@@ -155,13 +172,73 @@ def infer_relative_search_space(
...
+        if not self._multivariate:
+            return {}
+
+        search_space = {}  # type: Dict[str, BaseDistribution]
+        for name, distribution in self._search_space.calculate(study).items():
+            if distribution.single():
+                # `TPESampler` cannot handle distributions that contain
+                # just a single value, so we skip them. Note that the parameter values
+                # for such distributions are sampled in `Trial`.
+                continue
```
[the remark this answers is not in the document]
kstoneriv3: The multivariate TPE probably works with that. I will check and remove it is possible.
It's there as I just copied the code of infer_relatie_search_space from _cmaes.py.

optuna/samplers/_tpe/sampler.py
```diff
@@ -233,6 +310,38 @@ def _split_observation_pairs(
...
+
+        config_vals = {k: np.asarray(v, dtype=float) for k, v in config_vals.items()}
+        loss_vals = np.asarray(loss_vals, dtype=[("step", float), ("score", float)])
+
+        # TODO(kstoneriv3): Change of the order of 1. exclusion of `None` and 2. splitting might
+        # be needed. Independent sampler in `TPESampler` first splits the observations and
+        # then exclude `None`.
+
+        # We exclude param_vals with `None`.
+        config_vals_matrix = np.array([v for v in config_vals.values()])
```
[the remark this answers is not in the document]
kstoneriv3: Thank you, it makes good sense. I will fix this as you advised.

optuna/samplers/_tpe/sampler.py
```diff
@@ -532,6 +641,28 @@ def _compare(cls, samples: np.ndarray, log_l: np.ndarray, log_g: np.ndarray) ->
         else:
             return np.asarray([])
 
+    @classmethod
+    def _compare_multivariate(
+        cls,
+        multivariate_samples: Dict[str, np.ndarray],
+        log_l: np.ndarray,
+        log_g: np.ndarray,
+    ) -> np.ndarray:
```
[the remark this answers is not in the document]
kstoneriv3: Yes, you are right. I will fix this.

optuna/samplers/_tpe/sampler.py
kstoneriv3: The first sentence sounds good. What do you think about using the following as the second sentence?
"This is equivalent to converting the search space into the instance of `InterSectionSpace`."

optuna/samplers/_tpe/sampler.py
kstoneriv3: OK, thank you for your response. Then I will leave this as it is.

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
kstoneriv3: I do not think the order of iteration matters here. 
However, I totally agree that I should be careful about the order of iteration in using dict.

optuna/samplers/_tpe/sampler.py
```diff
@@ -651,3 +782,44 @@ def _get_observation_pairs(
...
+        else:
+            continue
+        scores.append(score)
+
+        # We extract param_value from trial.
+        for param_name in param_names:
+            param_value = None
+            if param_name in trial.params:
+                distribution = trial.distributions[param_name]
+                param_value = distribution.to_internal_repr(trial.params[param_name])
```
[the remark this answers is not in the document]
kstoneriv3: I am not sure if we can make the code faster and clearer by using 2-d `np.ndarray` here. 

In the `_MultivariateParzenEstimator` running behind the sample_relative, sampling, likelihood computation, and transformation of variables are carried out by iterating over parameters. This means we have to use 1-d slices of `np.ndarray` in these methods, and thus I do not expect a much speed up by using 2-d `np.ndarray` here.

As for clarity of using 2-d `np.ndarray`, using `dict` makes it easier to handle samples inside `_MultivariateParzenEstimator` because most of the attributes are stored as `dist`s in `_MultivariateParzenEstimator`. If we use 2-d `np.ndarray`, we have to carefully manage the order of the parameters (for example, by using `OrderedDict` as you pointed out elsewhere).

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
kstoneriv3: Yes, I will add a reference to BOHB there.

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
kstoneriv3: I updated the code to raise ValueError in the cases of empty input. Thus, I will close this.

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
kstoneriv3: I updated the docstring but the wordings might not be appropriate. Please have a look at it.

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
kstoneriv3: Removed this unnecessary if statement.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,418 @@
...
+
+        if samples_size == 0:
+            consider_prior = True
+
+        if consider_prior:
+
+            prior_mu = 0.5 * (low + high)
+            prior_sigma = 1.0 * (high - low)
+
+            mus = np.zeros(samples_size + 1)
```
hvy, line 286: This is also a detail but we only have to allocate memory here and not iterate over (fill) it.
```suggestion
            mus = np.empty(samples_size + 1)
```

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,418 @@
...
+            mus[:samples_size] = samples
+            mus[samples_size] = prior_mu
+
+            sigmas = np.zeros(samples_size + 1)
```
hvy, line 290: Ditto.
```suggestion
            sigmas = np.empty(samples_size + 1)
```

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,418 @@
...
+
+    def _calculate_weights(self, multivariate_samples: Dict[str, np.ndarray]) -> np.ndarray:
+
+        # We decide the weights.
+        consider_prior = self._parameters.consider_prior
+        prior_weight = self._parameters.prior_weight
+        weights_func = self._parameters.weights
+        sample_size = next(iter(multivariate_samples.values())).size
+        if consider_prior:
+            weights = np.zeros(sample_size + 1)
```
hvy, line 74: Ditto.
```suggestion
            weights = np.empty(sample_size + 1)
```

hvy (commented):
Started looking at the `_MultivariateParzenEstimator` so let me leave some again early comments. I  by the way find the code to overall be very readable 👍

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
hvy: As for using a dense 2d ndarray, I now see that we're clearly looping over each dimension and conditioning the logic based on the distribution. Your implementation looks reasonable. We can however assume that we're always taking the intersection and that all parameters are "not None".

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,418 @@
...
+        elif len(rescaled_samples_list[0]) == 0:
+            return np.ones(1, dtype=float)
+
+        rescaled_samples = np.array(rescaled_samples_list).T
+
+        # compute distance matrix of samples
+        distances = np.linalg.norm(
+            rescaled_samples[:, None, :] - rescaled_samples[None, :, :], axis=2
+        )
+        distances += np.diag(np.ones(len(samples)) * np.infty)
```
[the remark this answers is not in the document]
kstoneriv3: Makes sense a lot.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
kstoneriv3: Thanks for nice revision!

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,418 @@
...
+        assert isinstance(distribution, distributions.CategoricalDistribution)
+        choices = distribution.choices
+        consider_prior = self._parameters.consider_prior
+        prior_weights = self._parameters.prior_weight
+        if consider_prior:
+            weights = prior_weights / samples.size * np.ones((samples.size + 1, len(choices)))
+            weights[np.arange(samples.size), samples] += 1
+        else:
+            weights = prior_weights / samples.size * np.ones((samples.size, len(choices)))
+            weights[np.arange(samples.size), samples] += 1
```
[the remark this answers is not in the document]
kstoneriv3: Cool!

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,418 @@
...
+
+        # Categorical parameters are not considered.
+        param_names = list(multivariate_samples.keys())
+        continuous_param_types = (
+            distributions.UniformDistribution,
+            distributions.LogUniformDistribution,
+            distributions.IntUniformDistribution,
+            distributions.IntLogUniformDistribution,
+            distributions.DiscreteUniformDistribution,
+        )
```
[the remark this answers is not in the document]
kstoneriv3: Sound like a good idea. I will use `_NUMERICAL_DISTRIBUTION_CLASSES`.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,418 @@
...
+        rescaled_samples_list = []
+        for param_name in param_names:
```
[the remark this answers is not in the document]
kstoneriv3: No, they are always same as long as `IntersectnionSearchSpace` is used to calculate relative search space.
For me, it seems we do not have to iterate over the values of the `self._search_space`. How about the following?
`for param_name in self._search_space.keys():`

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,418 @@
...
+                samples = multivariate_samples[param_name]
+                samples = (samples - low) / (high - low)
+                rescaled_samples_list.append(samples)
+            else:  # Categorical parameters are ignored.
+                continue
+
+        if len(rescaled_samples_list) == 0:
+            return None
+        elif len(rescaled_samples_list[0]) == 0:
+            return np.ones(1, dtype=float)
```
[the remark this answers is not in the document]
kstoneriv3: OK, I will add comments here.

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
kstoneriv3: OK, I agree that removing the if statement and using assertion here make the code readable.

optuna/samplers/_tpe/sampler.py
```diff
@@ -146,6 +162,16 @@ def __init__(
...
 
+        self._multivariate = multivariate
+        self._search_space = IntersectionSearchSpace()
+
+        if multivariate:
+            self._raise_experimental_warning_for_multivariate()
+
+    @experimental("2.1.0", name="`multivariate = True` in TPESampler")
+    def _raise_experimental_warning_for_multivariate(self) -> None:
+        pass
```
HideakiImamura, line 173: As discussed in https://github.com/optuna/optuna/pull/1777, we have changed our coding style convention for the experimental argument. Could apply the same change in https://github.com/optuna/optuna/pull/1777?

optuna/samplers/_tpe/sampler.py
```diff
@@ -155,13 +181,67 @@ def infer_relative_search_space(
...
+
+        # If the number of samples is insufficient, we run random trial.
+        n = len(scores)
+        if n < self._n_startup_trials:
+            ret = {}
+            for param_name, param_distribution in search_space.items():
+                ret[param_name] = self._random_sampler.sample_independent(
+                    study, trial, param_name, param_distribution
+                )
+            return ret
```
HideakiImamura, line 223: If we return `{}`, parameters are automatically sampled by `sample_independent`. To simplify the code, how about doing it?
```suggestion
        if n < self._n_startup_trials:
            return {}
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -233,6 +313,40 @@ def _split_observation_pairs(
...
+        config_vals: Dict[str, List[Optional[float]]],
+        loss_vals: List[Tuple[float, float]],
+    ) -> Tuple[Dict[str, np.ndarray], Dict[str, np.ndarray]]:
+
+        config_vals = {k: np.asarray(v, dtype=float) for k, v in config_vals.items()}
+        loss_vals = np.asarray(loss_vals, dtype=[("step", float), ("score", float)])
+
+        # TODO(kstoneriv3): Change of the order of 1. exclusion of `None` and 2. splitting might
+        # be needed. Independent sampler in `TPESampler` first splits the observations and
+        # then exclude `None`.
```
HideakiImamura, line 327: I think we can remove this TODO comment because the current implementation is reasonable. As another option, how about removing `TODO` and rewriting the comment as follows? "We firstly exclude `None` and then split observations. This order is the opposite of the `independent _sample`."

optuna/samplers/_tpe/sampler.py
```diff
@@ -233,6 +313,40 @@ def _split_observation_pairs(
...
+
+        # We exclude param_vals with `None`
```
HideakiImamura, line 329: It seems that this line is unnecessary. 
```suggestion
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -651,3 +790,43 @@ def _get_observation_pairs(
...
+    if study.direction == StudyDirection.MAXIMIZE:
+        sign = -1
+
+    scores = []
+    values = {
+        param_name: [] for param_name in param_names
+    }  # type: Dict[str, List[Optional[float]]]
+    for trial in study._storage.get_all_trials(study._study_id, deepcopy=False):
+
+        # We extract score from trial.
```
HideakiImamura, line 809: A nit.
```suggestion
        # We extract score from the trial.
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -651,3 +790,43 @@ def _get_observation_pairs(
...
+                    score = (-step, float("inf"))
+                else:
+                    score = (-step, sign * intermediate_value)
+            else:
+                score = (float("inf"), 0.0)
+        else:
+            continue
+        scores.append(score)
+
+        # We extract param_value from trial.
```
HideakiImamura, line 825: Ditto.
```suggestion
        # We extract param_value from the trial.
```

HideakiImamura (commented):
Thanks for the swift actions! I have several minor comments on `_tpe/sampler.py`. Could you take a look?

optuna/samplers/_tpe/sampler.py
kstoneriv3: Sure.

optuna/samplers/_tpe/sampler.py
kstoneriv3: I agree. I will change the comment as you suggested.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
kstoneriv3: Now, I understand your point. I fixed this as you suggested.

optuna/samplers/_tpe/sampler.py
kstoneriv3: Fixed as you suggested.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,422 @@
...
+
+        rescaled_samples = np.array(rescaled_samples_list).T
+
+        # compute distance matrix of samples
+        distances = np.linalg.norm(
+            rescaled_samples[:, None, :] - rescaled_samples[None, :, :], axis=2
+        )
+        distances[np.diag_indices_from(distances)] += np.inf
+
+        return np.min(distances, axis=1)
```
kstoneriv3, line 249: ### Suggestion of new bandwidth selection rule (Scott's rule)

The current implementation of bandwidth selection uses the distance from the nearest neighbor sample as the bandwidth. However, this rule requires O(d*n^2) complexity for n samples with d parameters.

Thus, I suggest the use of another rule called Scott's rule, which is of complexity O(1). This rule sets the bandwidth proportionally to n^{-1/(d+4)}, which is theoretically optimal order of the bandwidth. 

In my benchmark (shown in the attached file), there is no significant difference between current implementation and Scott's rule. Also in the BOHB paper, Scott's rule is used for bandwidth selection in combination with multivariate TPE.

Please let me know your opinion on this.
```suggestion
    def _precompute_sigmas0(self, multivariate_samples: Dict[str, np.ndarray]) -> float:
        sample_size = next(iter(multivariate_samples.values())).size
        param_size = len(multivariate_samples)
        return sample_size ** (-1.0 / (param_size + 4))
```
Also, we can get rid of the following at line 278.
```
assert sigmas0 is not None
```


hpo-bench-naval-40f361d00195ba642c8866791ccec1b37a5cb8f0e3b1cf3092f76d731d4f9033
hpo-bench-parkinson-e042118c285061b2da9d3a89fc6c1dbf4b7f2f8384e8db7cabae2f31a5c2842e
hpo-bench-protein-aad4c2d81b4340ee2dc4ee252cea7eea3a49a3478fdbdf6065c265f99cd522f0
hpo-bench-slice-8ab21f97e472c9f9120602a69877fad413f2bacdae10ec7d1b637303fb04cbcc

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,422 @@
...
+            if isinstance(distribution, distributions.UniformDistribution):
+                transformed[param_name] = samples
+            elif isinstance(distribution, distributions.LogUniformDistribution):
+                transformed[param_name] = np.exp(samples)
+            elif isinstance(distribution, distributions.DiscreteUniformDistribution):
+                q = self._q[param_name]
+                low = self._low[param_name]
+                high = self._high[param_name]
+                samples = (np.round((samples - low) / q - 0.5) + 0.5) * q + low
+                transformed[param_name] = np.clip(samples, low, high)
```
HideakiImamura, line 188: These `low` and `high` are transformed `low` and `hight` in `_calculate_parzen_bounds`. We should use the original `distribution.low` and `distribution.high`.
```suggestion
                transformed[param_name] = np.clip(samples, distribution.low, distribution.high)
```

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,422 @@
...
+            elif isinstance(distribution, distributions.IntUniformDistribution):
+                q = self._q[param_name]
+                samples = np.round(samples / q) * q
+                samples = np.clip(samples, self._low[param_name], self._high[param_name])
```
HideakiImamura, line 192: Ditto.
```suggestion
                samples = np.clip(samples, distribution.low, distribution.high)
```

HideakiImamura (changes requested):
Thanks for the update. I have several comments for some transformation methods in `_tpe/multivariate_parzen_estimator.py`.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
HideakiImamura: Looks great! Using Scott's rule seems reasonable.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,422 @@
...
+    distributions.LogUniformDistribution,
+    distributions.IntUniformDistribution,
+    distributions.IntLogUniformDistribution,
+    distributions.DiscreteUniformDistribution,
+)
+
+
+# TODO(kstoenriv3): We need to consider if we really need `_ParzenEstimatorParameters` class
+# as there are alreaky so many attributes in `_MultivariateParzenEstimator` class.
+# `_ParzenEstimatorParameters` class is not greatly recuding the complexity of the code.
```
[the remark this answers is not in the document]
kstoneriv3: OK, I see your point. I agree that separating hyperparameters from other arguments is reasonable.
I will just remove the comments.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,422 @@
...
+            if isinstance(dist, distributions.CategoricalDistribution):
+                mus = sigmas = None
+                categorical_weights = self._calculate_categorical_params(samples, param_name)
+            else:
+                mus, sigmas = self._calculate_parzen_est_params(samples, param_name)
+                categorical_weights = None
+            self._mus[param_name] = mus
+            self._sigmas[param_name] = sigmas
+            self._categorical_weights[param_name] = categorical_weights
+
```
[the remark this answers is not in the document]
kstoneriv3: Makes sense. I will swap the order of methods. 
Are there any other conventions in how the methods should be ordered?

optuna/samplers/_tpe/multivariate_parzen_estimator.py
kstoneriv3: Thank you for pointing this out! It was a critical bug.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
HideakiImamura: I think we don't have any other rules for the order of methods. @hvy Do you know any other rules?

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,422 @@
...
+            distribution = self._search_space[param_name]
+            if isinstance(distribution, distributions.UniformDistribution):
+                transformed[param_name] = samples
+            elif isinstance(distribution, distributions.LogUniformDistribution):
+                transformed[param_name] = np.exp(samples)
+            elif isinstance(distribution, distributions.DiscreteUniformDistribution):
+                q = self._q[param_name]
+                low = self._low[param_name]
+                high = self._high[param_name]
+                samples = (np.round((samples - low) / q - 0.5) + 0.5) * q + low
```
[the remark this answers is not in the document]
kstoneriv3: This is quite a tricky part. 
Since `low = distribution.low - 0.5 * q`, this code is OK. 
`samples` are taken from `UnifromDistribution(distribution.low - 0.5*q, distribution.high + 0.5*q)`, argument of `np.round`, `(samples - low) / q - 0.5`, is included in [-0.5, (distribution.high - distribution.low) / q + 0.5].

For clarification, I will add some comments here.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,422 @@
...
+            elif isinstance(distribution, distributions.LogUniformDistribution):
+                transformed[param_name] = np.log(samples)
+            elif isinstance(distribution, distributions.DiscreteUniformDistribution):
+                transformed[param_name] = samples
+            elif isinstance(distribution, distributions.IntUniformDistribution):
+                transformed[param_name] = samples
+            elif isinstance(distribution, distributions.IntLogUniformDistribution):
+                transformed[param_name] = np.log(samples)
+            elif isinstance(distribution, distributions.CategoricalDistribution):
+                transformed[param_name] = samples
```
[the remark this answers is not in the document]
HideakiImamura: Hmm... I'm not sure the current one helps the readability for developers. The newbies for this multivariate TPE don't know that the same conditional statements are used in `_transform_to_uniform`, `_transform_from_uniform` and `_calculate_parzen_bounds`. Therefore, to reduce the amount of information at first glance of newbies, I think my suggested simplification is useful. 

I would like to know @hvy's opinion. What do you think?

hvy:
https://github.com/optuna/optuna/pull/1767#discussion_r482703233 

No I don't think so. If you're curious, I usually try to follow https://stackoverflow.com/questions/10289461/what-is-a-good-way-to-order-methods-in-a-python-class.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
kstoneriv3: How about this?
```suggestion
                samples = np.round((samples - distribution.low) / q) * q + distribution.low
```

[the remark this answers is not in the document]
HideakiImamura: Hmm... For example, please consider the case of `low =1.2, high = 2.4, q = 0.3`. This case the search space will be `{1.2, 1.5, 1.8, 2.1, 2.4}`. Let `sample = 1.7`. 

My suggested code results in `round((s - low) / q) * q + low = round((1.7 - 1.2) / 0.3) * 0.3 + 1.2 = 1.8`. This `1.8` is in the search space.

The current code results in `(round((1.7 - 1.2) / 0.3 - 0.5) + 0.5) * 0.3 + 1.2 = 1.65`. This `1.65` is not in the search space.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
hvy: By the way, in the BOHB paper, they multiply the bandwidths with a constant factor to promote exploration (Section 4.1. in https://arxiv.org/pdf/1807.01774.pdf). Is this something that has been considered?

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
HideakiImamura: I see. The `low` and `high` are not the original `low` and `high`. Yo new suggested code looks easy to understand. Great!

optuna/samplers/_tpe/multivariate_parzen_estimator.py
kstoneriv3: OK, I will adopt Scott's rule, with some comments for future maintenance.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,422 @@
...
+            self._sigmas[param_name] = sigmas
+            self._categorical_weights[param_name] = categorical_weights
+
+    def _calculate_weights(self, multivariate_samples: Dict[str, np.ndarray]) -> np.ndarray:
+
+        # We decide the weights.
+        consider_prior = self._parameters.consider_prior
+        prior_weight = self._parameters.prior_weight
+        weights_func = self._parameters.weights
+        sample_size = next(iter(multivariate_samples.values())).size
```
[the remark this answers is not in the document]
kstoneriv3: I agree that it degrades the readability. 
However, `next(iter(...)).size` is used only twice and they have different values. Here, it represents the number of components in the parzen estimator. In the other (second) occurrence, it represents `n_ei_candidates` as follows, and I assume we cannot do away with `next(iter(...)).size` here anyways.
```python3
    def log_pdf(self, multivariate_samples: Dict[str, np.ndarray]) -> np.ndarray:

        multivariate_samples = self._transform_to_uniform(multivariate_samples)
        weight_size = len(self._weights)
        sample_size = next(iter(multivariate_samples.values())).size
 ```
So, even though this is not the world's most readable code, I guess leaving this as it is OK for now.

File: optuna/samplers/_tpe/multivariate_parzen_estimator.py (hunk 24 above)
Comment (kstoneriv3) [reply optuna/samplers/_tpe/multivariate_parzen_estimator.py:250]: I am sorry to announce that I made a mistake in the above benchmark. 
I will not adopt this at the moment and will close this thread for now. If it performs well in the fixed benchmark, I might reopen this thread.

Comment (kstoneriv3) [reply optuna/samplers/_tpe/multivariate_parzen_estimator.py:246]: > By the way, in the BOHB paper, they multiply the bandwidths with a constant factor to promote exploration (Section 4.1. in https://arxiv.org/pdf/1807.01774.pdf). Is this something that has been considered?

No, but I believe this is worth exploring.

File: optuna/samplers/_tpe/multivariate_parzen_estimator.py (hunk 30 above)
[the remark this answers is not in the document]
Comment (hvy) [reply optuna/samplers/_tpe/multivariate_parzen_estimator.py:149]: Purely speaking about style and readability, I understand @HideakiImamura 's suggestion is more consistent with what we already have, which might be good in itself. From a readability point of view, @kstoneriv3 's original code contains fewer lines and less indentation, so I personally find it quite pleasing. 

Another idea is the following. I think this makes the code very clear in what it does. 

```python
def _transform_to_uniform(
    self, multivariate_samples: Dict[str, np.ndarray]
) -> Dict[str, np.ndarray]:

    transformed = {}
    for param_name, samples in multivariate_samples.items():
        distribution = self._search_space[param_name]

        assert distribution in distributions.DISTRIBUTION_CLASSES
        if isinstance(distribution, (LogUniformDistribution, IntLogUniformDistribution)):
            samples = np.log(samples)

        transformed[param_name] = samples
    return transformed
```

As for the assertion, I suggest we check all the distribution types in `__init__` and assert in following usages.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
hvy: How about leaving a TODO?

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
hvy: Okay 👍

HideakiImamura (commented):
I have a minor comment.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
kstoneriv3: OK, I'll adopt @hvy 's idea.
As for assertion, your statements were a bit unclear for me. Did you mean to assert all distributions in the `search_space` is in `DISTRIBUTION_CLASSES` in `__init__` ?

optuna/samplers/_tpe/sampler.py
```diff
@@ -155,13 +182,62 @@ def infer_relative_search_space(
...
+                (
+                    distributions.UniformDistribution,
+                    distributions.LogUniformDistribution,
+                    distributions.DiscreteUniformDistribution,
+                    distributions.IntUniformDistribution,
+                    distributions.IntLogUniformDistribution,
+                    distributions.CategoricalDistribution,
+                ),
+            ):
+                continue
```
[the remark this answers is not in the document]
kstoneriv3: Sounds reasonable to use `distributions.DISTRIBUTION_CLASSES`. I will fix this as you suggested.

optuna/samplers/_tpe/sampler.py
```diff
@@ -233,6 +309,28 @@ def _split_observation_pairs(
...
+        n_below = self._gamma(len(loss_vals))
+        index_loss_ascending = np.argsort(loss_vals)
+        # `np.sort` is used to keep chronological order.
+        index_below = np.sort(index_loss_ascending[:n_below])
+        index_above = np.sort(index_loss_ascending[n_below:])
+        below = {}
+        above = {}
+        for param_name in config_vals.keys():
+            below[param_name] = config_vals[param_name][index_below]
+            above[param_name] = config_vals[param_name][index_above]
```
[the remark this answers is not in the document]
kstoneriv3: Yes, you are right. I will fix this.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,419 @@
...
+    distributions.IntUniformDistribution,
+    distributions.IntLogUniformDistribution,
+    distributions.DiscreteUniformDistribution,
+)
+
+
+class _MultivariateParzenEstimator:
+    def __init__(
+        self,
+        multivariate_samples: Dict[str, np.ndarray],
```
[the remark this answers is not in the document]
kstoneriv3: I agree. I will change `samples` to `observations` where appropriate.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
HideakiImamura: I think @hvy means that we should raise a `NotImplementedError` as done in `TPESampler.sample_independent`.

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
kstoneriv3: Just for clarification, can't we just skip unsupported distributions and pass them to the `sample_independent`?

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
kstoneriv3: As discussed [here](https://github.com/optuna/optuna/pull/1767#discussion_r482732838), I changed some of `samples` to `observations`. This has led to the use of `observation_size`, which I feel is a bit lengthy. 
So, I would like to hear your opinion on changing `*_size` to `n_*`.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
hvy: Thanks @HideakiImamura for the clarification. Yes, I meant to raise in `_MultivariateParzenEstimator.__init__` after checking. Then, we can assume in following usages within the class. But I now think that it might be better to explicitly list all distribution instead of using `DISTRIBUTION_CLASSES` as this container may change.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
hvy: I see, this is not a strong opinion but `n_*` is I think more common in Optuna so it might be better both in terms of length and consistency.

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
hvy: That's also fine. In that case, it might be a good idea to let users/us know when we're falling back to independent similar to [how it's done in the `CmaEsSampler`](https://github.com/optuna/optuna/blob/master/optuna/samplers/_cmaes.py#L374-L377) (note that there might be user-defined distributions).

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
kstoneriv3: OK, I will shift from `*_size` to `n_*`.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
kstoneriv3: @hvy @HideakiImamura  This kind of situation is happening in both sampler.py and multivariate_parzen_estimator.py. I would like to take a unified approach to them. Which of the following do you think is the best approach?
- use `DISTRIBUTION_CLASSES`
- use `_DISTRIBUTION_CLASSES` (private variable)
- use an explicit tuple or list of all possible distributions.

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
kstoneriv3: Thank you, raising a warning like `CmaEsSampler` seems more reasonable to me.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
hvy: Good question. If a new distribution is introduced, either by a user or by Optuna (appended to `DISTRIBUTION_CLASSES`), my understand is that it's more user friendly to fall back to random sampling with a warning rather than raising an error. In this case, the second (or third) option sounds reasonable creating a private variable and reusing it throughout, including in `infer_relative_search_space`.

optuna/samplers/_tpe/sampler.py
```diff
@@ -143,9 +162,21 @@ def __init__(
         self._gamma = gamma
         self._weights = weights
 
+        self._logger = get_logger(__name__)
```
hvy, line 165: Can you define the logger at module level? Sorry it might not be clear but it's a style adopted in Optuna. You can take a look at `_cmaes.py` for how it's done https://github.com/optuna/optuna/blob/ee08edccfb6b0c204ff7cd2c6a67baaa5166635c/optuna/samplers/_cmaes.py#L21.

kstoneriv3: Thank you for pointing out! I was having a problem with testing around logger.

kstoneriv3: It seems that _logger defined at module level is not used in the _log_independent_sampling in cma. `self._logger` is used there. Could you elaborate on why this is the case?

hvy: You're right. I think it's unintentional.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
kstoneriv3: OK, I will go for the second option.

optuna/samplers/_tpe/sampler.py
kstoneriv3: done and passed CI

optuna/samplers/_tpe/multivariate_parzen_estimator.py
kstoneriv3: I took the benchmark again, and fortunately, Scott's rule outperformed the current strategy.
So I modified `_precompute_sigmas0` and added comments on TODOs.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,371 @@
...
+
+        return transformed
+
+    def _precompute_sigmas0(self, multivariate_samples: Dict[str, np.ndarray]) -> np.ndarray:
+        # We use Scott's rule for bandwidth selection.
+        # This rule was used in the BOHB paper.
+        n_samples = next(iter(multivariate_samples.values())).size
+        n_samples = max(n_samples, 1)
+        n_params = len(multivariate_samples)
+        # TODO(kstoneriv3): The constant factor 0.2 might not be optimal.
```
HideakiImamura, line 288: I think it is good to make this "constant" tunable. That is, how about making a new input `sigma0_magnitude` for the `parameters` in `_MultivariateParzenEstimator.__init__`.? I mean
- [1] Create `_MultivariateParzenEstimatorParameters` with parameters of the normal Parzen estimator and `sigma0_magnitude`.
- [2] Make the input of parameters in `_MultivariateParzenEstimator.__init__` as `_MultivariateParzenEstimatorParameters`.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,371 @@
...
+
+EPS = 1e-12
+_DISTRIBUTION_CLASSES = (
+    distributions.UniformDistribution,
+    distributions.LogUniformDistribution,
+    distributions.DiscreteUniformDistribution,
+    distributions.IntUniformDistribution,
+    distributions.IntLogUniformDistribution,
+    distributions.CategoricalDistribution,
+)
```
HideakiImamura, line 21: We already have the `DISTRIBUTION_CLASSES` in `optuna/distributions.py`.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,371 @@
...
+                        ),
+                    )
+            multivariate_samples[param_name] = samples
+        multivariate_samples = self._transform_from_uniform(multivariate_samples)
+        return multivariate_samples
+
+    def log_pdf(self, multivariate_samples: Dict[str, np.ndarray]) -> np.ndarray:
+
+        multivariate_samples = self._transform_to_uniform(multivariate_samples)
+        n_weights = len(self._weights)
```
HideakiImamura, line 125: This is equal to the number of observations? How about using `n_observations`?
```suggestion
        n_observations = len(self._weights)
```

HideakiImamura (commented):
Thanks for the update! I have some minor comments.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,393 @@
...
+        n_samples = next(iter(multivariate_samples.values())).size
+        if n_samples == 0:
+            return np.asarray([], dtype=float)
+        # We compute log pdf (compoment_log_pdf)
```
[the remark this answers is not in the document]
kstoneriv3: Thanks!

optuna/samplers/_tpe/multivariate_parzen_estimator.py
hvy: I believe the partially shared consensus was to maintain a list of supported distributions by multivariate TPE separately, confined to multivariate TPE, without being affected by changes to `distributions.py`.

kstoneriv3: I chose to define this tuple at the module level, in case modifications in `optuna/distributions.py` are made in the future.
What do you think about this point?

optuna/samplers/_tpe/multivariate_parzen_estimator.py
kstoneriv3: Good idea!

optuna/samplers/_tpe/multivariate_parzen_estimator.py
kstoneriv3: Yes, it is. Still, I am a little afraid similarity of words samples and observations might cause confusion. On the other hand, `n_weights` and `n_observations` being the same might cause another confusion as you pointed out.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
HideakiImamura: Thanks for the clarification. I understood. It's fine as it is now.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
kstoneriv3: Probably, changing all `n_observations` to `n_weights` is a good idea

HideakiImamura: Hmm. I think `n_weights` is misleading. When I see this, I wonder what weight is for `n_weight`. I think we already have the naming convention of `sample` and `observation`. Therefore, I think `n_observations` is reasonable here.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
kstoneriv3: I feel that we should include `sigma0_magnitude` in `_ParzenEstimatorParameter` as `_MultivariateParzenEstimator`'s arguments are organized followingly. What do you think?
```
class _MultivariateParzenEstimator:

    def __init__(
        self,
        multivariate_observations: Dict[str, np.ndarray],
        search_space: Dict[str, BaseDistribution],
    ) -> None:
```

tests/samplers_tests/tpe_tests/test_multivariate_parzen_estomator.py
```diff
@@ -0,0 +1,148 @@
...
+        weights=lambda x: np.arange(x) + 1.0,
+    )
+    # Parzen estimator almost becomes mixture of Dirac measures.
+    with patch(_precompute_sigmas0, return_value=1e-8 * np.ones(1)):
+        mpe = _MultivariateParzenEstimator(multivariate_samples, search_space, parameters)
+
+    log_pdf = mpe.log_pdf(multivariate_samples)
+    output_multivariate_samples = mpe.sample(np.random.RandomState(0), 100)
+    output_log_pdf = mpe.log_pdf(output_multivariate_samples)
+    # The more data points, the smaller the value of ``log_pdf`` should be.
```
kstoneriv3, line 147: I am not sure about the second comment (sorry, it's probably because of my unclear code). As can be found in the first comment, the log-likelihoods of the random samples (`output_multivaraite_samples`) are w.h.p. less than the samples used for the construction of the Parzen estimator (`multivariate_samples`) because the Parzen estimator becomes almost becomes a mixture of Dirac measures. For such Dirac measures, we can expect that the log-likelihood of the points becomes almost zero except for centers of each Dirac measures.

HideakiImamura: Thanks for the comment. I understood the theoretical reason for that. I will change the comment as follow: "The likelihood of the previous observations is a positive value, and that of the points sampled by the Parzen estimator is zero with high probability." What do you think?

kstoneriv3: Yes, that sounds OK to me but a small suggestion; I guess "almost" zero is a more accurate expression.

HideakiImamura: Thanks for the comment. I will change so soon.

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -26,20 +27,336 @@ def test_hyperopt_parameters(use_hyperband: bool) -> None:
...
+
+
+def test_infer_relative_search_space() -> None:
+
     sampler = TPESampler()
+    search_space = {
+        "a": distributions.UniformDistribution(1.0, 100.0),
+        "b": distributions.LogUniformDistribution(1.0, 100.0),
+        "c": distributions.DiscreteUniformDistribution(1.0, 100.0, 3.0),
+        "d": distributions.IntUniformDistribution(1, 100),
```
toshihikoyanase, line 42: How about adding an int distribution with `step=3` (or something) to the search space? I think the tests don't contain `trial.suggest_int` with the step parameter.

toshihikoyanase (commented):
Thanks for the great PR. I just started the review of tests. I added an initial comment.

tests/samplers_tests/tpe_tests/test_sampler.py
HideakiImamura: I will add a test. Thank you for the comment.

HideakiImamura:
@hvy @toshihikoyanase PTAL.

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -5,11 +5,13 @@
 from typing import Union
 from unittest.mock import Mock
 from unittest.mock import patch
+import warnings
```
hvy, line 8: I don't think we explicitly silence `ExperimentalWarning`s in other tests. Mostly just curious but how come we should do it here? Also I think it's a good practice to make the filter scoped/temporary with `with warnings.catch_warnings():`

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -26,20 +28,372 @@ def test_hyperopt_parameters(use_hyperband: bool) -> None:
...
+        weights=lambda n: np.asarray([i ** 2 + 1 for i in range(n)]),
+        n_startup_trials=5,
+        seed=0,
+        multivariate=True,
+    )
+    with patch.object(study._storage, "get_all_trials", return_value=past_trials):
+        assert sampler.sample_relative(study, trial, {"param-a": dist}) != suggestion
+
+
+def test_sample_relative_uniform_distributions() -> None:
```
hvy, line 179: Are these introduced because `samplers_tests/test_sampler.py` only cover independent sampling? I'm wondering if we shouldn't work on `samplers_tests/test_sampler.py` instead. It could be a separate issue.

HideakiImamura: Thank you for the insightful comment. Yes, you are correct. I think we can just remove these `test_sample_relative_DISTRIBUTIONS()` without any decrease in test coverage.

HideakiImamura: Thank you for the insightful comment. Actually, we can test `sample_relative` not only for `TPESampler(multivariate=True)` but also for any distributions in `sampelrs_tests/test_samplers.py`. We can work on this in another PR as a followup. 

Here, how about leaving them as it is?

hvy (changes requested):

optuna/samplers/_tpe/sampler.py
```diff
@@ -119,6 +134,16 @@ def objective(trial):
             for more details.
         seed:
             Seed for random number generator.
+        multivariate:
+            If this is :obj:`True`, the multivariate TPE is used when suggesting parameters.
+            The multivariate TPE is reported to outperform the independent TPE. See `BOHB: Robust
+            and Efficient Hyperparameter Optimization at Scale
+            <http://proceedings.mlr.press/v80/falkner18a.html>`_ for more details.
+
+            .. versionadded:: 2.2.0
```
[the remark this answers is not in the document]
kstoneriv3: I updated the code as you suggested.

optuna/samplers/_tpe/sampler.py
```diff
@@ -132,6 +157,9 @@ def __init__(
         gamma: Callable[[int], int] = default_gamma,
         weights: Callable[[int], np.ndarray] = default_weights,
         seed: Optional[int] = None,
+        *,
+        multivariate: bool = False,
+        warn_independent_sampling: bool = True
```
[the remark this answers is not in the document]
kstoneriv3: I agree it should be commented on. I updated the code as you suggested.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,375 @@
...
+        distribution = self._search_space[param_name]
+        assert isinstance(distribution, distributions.CategoricalDistribution)
+        choices = distribution.choices
+        consider_prior = self._parameters.consider_prior
+        prior_weights = self._parameters.prior_weight
+        if consider_prior:
+            shape = (n_weights + 1, len(choices))
+        else:
+            shape = (n_weights, len(choices))
+        weights = np.full(shape, fill_value=prior_weights / n_weights)
```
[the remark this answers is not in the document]
kstoneriv3: Your suggestion sounds reasonable to me. I change the code to reflect your point but I am not sure if the code looks good. Please have a look at it.

tests/samplers_tests/tpe_tests/test_multivariate_parzen_estomator.py
```diff
@@ -0,0 +1,149 @@
...
+    q = {"a": None, "b": None, "c": 3.0, "d": 1.0, "e": None, "f": None}
+    low = {"a": 1.0, "b": np.log(1.0), "c": -0.5, "d": 0.5, "e": np.log(0.5), "f": None}
+    high = {"a": 100.0, "b": np.log(100.0), "c": 101.5, "d": 100.5, "e": np.log(100.5), "f": None}
+
+    assert np.all(mpe._weights == weights)
+    assert mpe._q == q
+    assert mpe._low == low
+    assert mpe._high == high
+
+    sigmas = {
```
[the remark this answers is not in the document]
kstoneriv3: Sounds good! I updated the code as you suggested.

tests/samplers_tests/tpe_tests/test_multivariate_parzen_estomator.py
```diff
@@ -0,0 +1,149 @@
...
+    for param_name, values in mpe._sigmas.items():
+        assert np.all(
+            np.equal(mpe._sigmas[param_name], sigmas[param_name])
+        ), 'parameter "{}"'.format(param_name)
+        assert np.all(np.equal(mpe._mus[param_name], mus[param_name])), "parameter: {}".format(
+            param_name
+        )
+        assert np.all(
+            np.equal(mpe._categorical_weights[param_name], categorical_weights[param_name])
+        ), "parameter: {}".format(param_name)
```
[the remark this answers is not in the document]
kstoneriv3: Thank you for your suggestion. Totally agree.

tests/samplers_tests/tpe_tests/test_multivariate_parzen_estomator.py
```diff
@@ -0,0 +1,149 @@
...
+multivariate_samples = {
+    "a": np.array([1.0]),
+    "b": np.array([1.0]),
+    "c": np.array([1.0]),
+    "d": np.array([1]),
+    "e": np.array([1]),
+    "f": np.array([1]),
+}
+
+_precompute_sigmas0 = (
```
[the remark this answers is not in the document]
kstoneriv3: Makes sense. I updated the code.

tests/samplers_tests/tpe_tests/test_multivariate_parzen_estomator.py
```diff
@@ -0,0 +1,149 @@
...
+
+    # Test the values of the output.
+    # As we set ``consider_prior`` = False and pre-computed sigma to be 1e-8,
+    # the samples almost equals to the input ``multivariate_samples``.
+    output_multivariate_samples = mpe.sample(np.random.RandomState(0), 1)
+    for param_name, samples in output_multivariate_samples.items():
+        if samples.dtype == str:
+            assert samples[0] == "y", "parameter {}".format(param_name)
+        else:
+            assert np.allclose(samples, multivariate_samples[param_name]), "parameter {}".format(
```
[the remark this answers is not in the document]
kstoneriv3: Thank you for pointing out! updated.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,375 @@
...
+        consider_prior = self._parameters.consider_prior
+        consider_magic_clip = self._parameters.consider_magic_clip
+        sigmas0 = self._sigmas0
+        low = self._low[param_name]
+        high = self._high[param_name]
+        assert low is not None
+        assert high is not None
+
+        if n_weights == 0:
+            consider_prior = True
```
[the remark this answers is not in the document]
kstoneriv3: I strongly agree that `_calculate_categorical_params` must be consistent with `_calculate_parzen_est_params` so I updated the code.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
```diff
@@ -0,0 +1,375 @@
...
+        if consider_prior:
+            shape = (n_weights + 1, len(choices))
+        else:
+            shape = (n_weights, len(choices))
+        weights = np.full(shape, fill_value=prior_weights / n_weights)
+        weights[np.arange(n_weights), observations] += 1
+        weights /= weights.sum(axis=1, keepdims=True)
+        return weights
+
+    def _calculate_parzen_est_params(
```
kstoneriv3, line 317: Shouldn't this be `_calculate_numerical_params` instead?

HideakiImamura: The '_calculate_numerical_params` is more reasonable to me than the `_calculate_parzen_est_params`.

optuna/samplers/_tpe/sampler.py
```diff
@@ -119,6 +134,19 @@ def objective(trial):
...
+            and Efficient Hyperparameter Optimization at Scale
+            <http://proceedings.mlr.press/v80/falkner18a.html>`_ for more details.
+
+            .. note::
+                Added in v2.2.0 as an experimental feature. The interface may change in newer
+                versions without prior notice. See
+                https://github.com/optuna/optuna/releases/tag/v2.2.0.
+        warn_independent_sampling:
+            If this is :obj:`True`, a warning message is emitted when
+            the value of a parameter is sampled by using an independent sampler.
```
hvy, line 149: We could clarify mentioning that this argument does nothing if `multivariate` is `False`.

optuna/samplers/_tpe/multivariate_parzen_estimator.py
hvy: +1

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -26,20 +28,391 @@ def test_hyperopt_parameters(use_hyperband: bool) -> None:
...
+    """Test samples from discrete have expected intervals."""
+
+    study = optuna.create_study()
+    disc_dist = optuna.distributions.DiscreteUniformDistribution(1.0, 100.0, 0.1)
+
+    def value_fn(idx: int) -> float:
+        random.seed(idx)
+        return int(random.random() * 1000) * 0.1
+
+    past_trials = [frozen_trial_factory(i, dist=disc_dist, value_fn=value_fn) for i in range(1, 8)]
```
toshihikoyanase, line 246: This type of `value_fn` is not originally introduced in this PR, but I think we can set it once right before executing the `value_fn`. Do you think we can work on it in this PR or should I create a new PR?
```suggestion
    random.seed(1)
    past_trials = [frozen_trial_factory(i, dist=disc_dist, value_fn=lambda _: int(random.random() * 1000) * 0.1) for i in range(1, 8)]
```

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -26,20 +28,391 @@ def test_hyperopt_parameters(use_hyperband: bool) -> None:
...
+    with patch.object(study._storage, "get_all_trials", return_value=past_trials):
+        discrete_uniform_suggestion = sampler.sample_relative(study, trial, {"param-a": disc_dist})
+    assert 1.0 <= discrete_uniform_suggestion["param-a"] <= 100.0
+    assert (
+        abs(
+            int(discrete_uniform_suggestion["param-a"] * 10)
+            - discrete_uniform_suggestion["param-a"] * 10
+        )
+        < 1e-3
+    )
```
toshihikoyanase, line 260: How about using `np.testing.assert_almost_equal` here?
```suggestion
    np.testing.assert_almost_equal(int(discrete_uniform_suggestion["param-a"] * 10), discrete_uniform_suggestion["param-a"] * 10)
```

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
HideakiImamura: In my understanding, `prior_weight` is uniformly added to the number of counts for each choice in the independent categorical sampling. However, in this relative sampling, it seems that `prior_weight / n_weights` (or `prior_weights / (n_weights + 1)`) is added to the number of counts. To be consistent with the independent case, how about making `value = prior_weight`?

toshihikoyanase (commented):
Thank you for your update. I added some small comments on the tests. Please take a look.

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -26,20 +28,391 @@ def test_hyperopt_parameters(use_hyperband: bool) -> None:
...
+        t.suggest_uniform("a", 1.0, 100.0)
+        t.suggest_loguniform("b", 1.0, 100.0)
+        t.suggest_discrete_uniform("c", 1.0, 100.0, 3.0)
+        t.suggest_int("d", 1, 100)
+        t.suggest_int("e", 0, 100, step=2)
+        t.suggest_int("f", 1, 100, log=True)
+        t.suggest_categorical("g", ["x", "y", "z"])
+        return 0.0
+
+    # We test on the empty input
```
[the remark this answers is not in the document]
HideakiImamura: Thanks for the comment. I will remove it.

tests/samplers_tests/tpe_tests/test_sampler.py
HideakiImamura: I think it should be done in another PR because it affects the other codes of this PR.

HideakiImamura:
@hvy @toshihikoyanase Thanks for the all of reviews! Execpt for one discussion point (https://github.com/optuna/optuna/pull/1767#discussion_r493255301), I think this PR is ready to be merged. Could you take another look?

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
kstoneriv3: (This is related to #1798.) 

I believe that current implementation is consistent with the independent case. For example, when there are 10 observations from 3 categories, `consider_prior==False`, `sample_from_categorical_dist` randomly picks one observation and sample from the Parzen window that place weights of `1 + 1 / 10` to the observed category and `1 / 10` to the other two categories. Thus, as a whole, we sample from the observed categories with a probability of `11 / 13` and from the unobserved categories (in other words, from prior) with probability of `2 / 13`. 

As far as I understand, if we set `value = prior_weight`, the Parzen window cannot be highly centered around observed categories. In the previous example, when `prior_weight=1.0` and `value=prior_weight`, the Parzen estimator places weights of 2.0 to the observed categories and 1.0 to the others, no matter how many observations we have.

However, I believe the currently implementation of the categorical Parzen window is too much centered around the observed categories. It is highly inclined towards exploitation rather than exploration. I believe this is an important issue to be discussed in  #1798.

hvy:
Thanks @HideakiImamura, @kstoneriv3 for your quick actions. It LGTM. I agree though that it might be better to finish discussing the weights for categorial parameters. I am not too confident on the consensus behind the logic of the independent TPE but maybe @kstoneriv3 has some comment on why the logic differs?

optuna/samplers/_tpe/multivariate_parzen_estimator.py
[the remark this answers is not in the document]
HideakiImamura: Thanks for the detailed explanation. In the independent sampling, the `prior_weight` is added after the bin count for each choice. On the other hand, in the categorical sampling, the `value = prior_weight / n_weight` is added before the bin count for each choice and each observation, where `n_weight = the number of observations`. Therefore, the current logic is consistent.

I will change the order of "adding `prior_weight`" and "bin counting" for readability. I will also rename `n_weights` to `n_observations` for readability.

HideakiImamura (approved):
@kstoneriv3 Thanks for the discussion. I think this PR is ready to be merged. Thanks so much for the long effort!

@hvy @toshihikoyanase Please merge this PR after your approvals.

HideakiImamura:
TODOs after this PR.
1. making `value_fn` inline in `tests/samplers_tests/tpe_tests/test_sampler.py` as suggested [here](https://github.com/optuna/optuna/pull/1767#discussion_r493249913).
2. simplifying the tests of `sample_relative` as suggested [here](https://github.com/optuna/optuna/pull/1767#discussion_r489290985).

optuna/samplers/_tpe/sampler.py
```diff
@@ -145,8 +145,9 @@ def objective(trial):
                 versions without prior notice. See
                 https://github.com/optuna/optuna/releases/tag/v2.2.0.
         warn_independent_sampling:
-            If this is :obj:`True`, a warning message is emitted when
+            If this is :obj:`True` and `multivariate=True`, a warning message is emitted when
```
toshihikoyanase, line 148: A nit, but a pair of single backticks corresponds to italic.
```suggestion
            If this is :obj:`True` and ``multivariate=True``, a warning message is emitted when
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -145,8 +145,9 @@ def objective(trial):
...
             the value of a parameter is sampled by using an independent sampler.
+            If `multivariate=False`, this flag has no effect.
```
toshihikoyanase, line 150: ditto.
```suggestion
            If ``multivariate=False``, this flag has no effect.
```

toshihikoyanase (commented):
Thank you for reflecting my comments. The change LGTM, but I added small comments.

toshihikoyanase (approved):
Thank you for your swift action. The change looks good to me.

hvy (approved):
LGTM!

kstoneriv3:
Thank you so much for your great efforts in the reviews and commits! I am excited to see the multivariate TPE on the next release!

Issue #1766, opened 2020-08-30
Dask / Optuna Integration


mrocklin:
Hello Optuna maintainers, 

I work on Dask and I would like to improve integration between Optuna and Dask.  I did a brief investigation a couple of days ago (see https://github.com/dask/dask/issues/6571) and now have some questions.  Hopefully this is a good place to post questions:

## Nice design

First, (not actually a question) Optuna's distribution system seems really simple.  Thank you for the thoughtful design.  

## Storage

Second, I'm curious about the storage backend.  It looks like I can call `optuna.Study.optimize` with the same function in many places and if all of those places point to the same storage then everything will work smoothly.  Great.  If the user has a redis or relational database that is globally accessible then I think that there is no more work to do, other than maybe some documentation or examples.

However, I think that for many data science users setting up a relational database that is visible from every node in a cluster might be difficult.  I think that it would be useful to create an equivalent of the local in-memory storage that worked across a Dask cluster.   We can do this on the Dask side easily, but I have a couple of questions  that might influence the design.

1.  How fast should this be?  Is a 5ms response time fast enough, or should this be closer to 500us?
2.  I think that this will receive one query and one insert per trial.  Is that correct?  Those queries are for the full history of hyperparameters and scores so far?

### Testing

Also, I'm curious about registering external storages with Optuna. If I make a Dask storage for Optuna I would like to also use the test suite in `tests/storages_tests/test_storages.py`.  We have a few options here:

1.  We make a new dask-optuna package and copy-paste Optuna's test suite into that package
2.  We convert Optuna's test suite into a `unittest.TestSuite` class that we can then import in the dask-optuna package (this is nicer because `dask-optuna` tracks changes in `optuna`).
3.  We build the Dask storage within the Optuna codebase and add the Dask storage to the list of storages to test if Dask is installed

Some of these changes affect the optuna codebase.  Generally, the question here is if you want Dask modules inside github.com/optuna/optuna

## Joblib

I tried wrapping `study.optimize` in a `joblib.parallel_backend("dask")` context manager and was surprised that Dask did not take over from the joblib threading backend.  I'm not yet sure why.  If anyone has any thoughts here I welcome them.

## Is there anything else?

Other than Storage, is there anything else that we should consider?

I do not have practical experience with Optuna.  Is this useful?

Thank you for your time

ytsmiling:
Optuna-dask integration sounds really nice. And I'm looking forward to trying it.

[...]

PR #1772, opened 2020-08-31
Support AllenNLPPruningCallback for AllenNLPExecutor

himkt:
## Motivation
This PR adds support for `AllenNLPPruningCallback` in `AllenNLPExecutor`.

## Description of the changes

- Implement `get_url` for storages
- Set some Optuna related information as environment variables in `AllenNLPExecutor`
- Load them in `AllenNLPPruningCallback`

I want to get feedback about (a) designing an interface, (b) adding internal APIs (`get_url`), and (c) usage of them.


optuna/storages/_redis.py
```diff
@@ -65,6 +65,7 @@ def __init__(self, url: str) -> None:
 
         _imports.check()
 
+        self._url = url
```
himkt, line 68: In `RDBStororage`, `url` is a public attribute. Therefore I'm also wondering I should make it public.
https://github.com/optuna/optuna/blob/20ad7140c1bf492c45cb056593eecf3af4fee77f/optuna/storages/_rdb/storage.py#L104

optuna/integration/allennlp.py
```diff
@@ -188,3 +224,11 @@ def __call__(
         self._trial.report(float(value), epoch)
         if self._trial.should_prune():
```
himkt, line 225: [TODO] The current implementation re-creates a pruner (`MedianPruner`, which is a default pruning algorithm in Optuna) even if we specify a pruner in Optuna side.

optuna/integration/allennlp.py
```diff
@@ -157,22 +174,41 @@ class AllenNLPPruningCallback(EpochCallback):
...
             ``validation_accuracy``.
+        trial:
+            A :class:`~optuna.trial.Trial` corresponding to the current evaluation of the
+            objective function.
     """
 
-    def __init__(self, trial: optuna.trial.Trial, monitor: str):
+    def __init__(
+            self,
+            trial: Optional[optuna.trial.Trial] = None,
```
himkt, line 187: I want to make arguments `optional`, as `AllenNLPPruningCallback` would be initialized with any kind of information if it is specified in AllenNLP jsonnet configuration.

You can see the example in my GitHub repo.
https://github.com/himkt/optuna-allennlp/compare/master...feat/allennlp-jsonnet-pruning

optuna/integration/_allennlp.py
```diff
@@ -0,0 +1,14 @@
+from typing import Any
+from typing import Callable
+
+
+class EpochCallback:
```
himkt, line 5: This is introduced for building documentation.

This PR adds a decorator `@EpochCallback.register` to `AllenNLPPruningCallback`,
which is undefined in document CI as it doesn't install `allennlp`.

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,33 @@
...
     import allennlp.common.util
-    from allennlp.training import EpochCallback
 
 if _imports.is_successful():
     import _jsonnet
+    from allennlp.training import EpochCallback
 else:
-    EpochCallback = object  # NOQA
+
+    class EpochCallback:  # type: ignore
```
crcrpar, line 27: We should explain why `# type: ignore` as I don't think it's evident at a glance.

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,33 @@
...
 
 with try_import() as _imports:
     import allennlp
     import allennlp.commands
     import allennlp.common.util
-    from allennlp.training import EpochCallback
 
 if _imports.is_successful():
     import _jsonnet
+    from allennlp.training import EpochCallback
```
crcrpar, line 24: We should explain why only `EpochCallback` is imported conditionally.

optuna/integration/allennlp.py
himkt: You're right. I added some notes in 5ef26c8.

optuna/integration/allennlp.py
himkt: ditto.

optuna/integration/allennlp.py
himkt: https://github.com/optuna/optuna/pull/1772#discussion_r480194645
I implemented the API to restore a pruner defined in the Optuna side.

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,127 @@
...
+    elif pruner_class == "MedianPruner":
+        pruner = optuna.pruners.MedianPruner
+    elif pruner_class == "PercentilePruner":
+        pruner = optuna.pruners.PercentilePruner
+    elif pruner_class == "ThresholdPruner":
+        pruner = optuna.pruners.ThresholdPruner
+    elif pruner_class == "SuccessiveHalvingPruner":
+        pruner = optuna.pruners.SuccessiveHalvingPruner
+    elif pruner_class == "NopPruner":
+        pruner = optuna.pruners.NopPruner
```
crcrpar, line 69: Doesn't this work because this logic looks verbose to me?

```suggestion
    pruner = getattr(optuna.pruners, pruner_class, None)
```

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,127 @@
...
+    pruner = trial.study.pruner
+    kwargs = {}  # type: Dict[str, Any]
+
+    if isinstance(pruner, optuna.pruners.HyperbandPruner):
+        kwargs["min_resource"] = pruner._min_resource
+        kwargs["max_resource"] = pruner._max_resource
+        kwargs["reduction_factor"] = pruner._reduction_factor
+    elif isinstance(pruner, optuna.pruners.MedianPruner) or isinstance(
+        pruner, optuna.pruners.PercentilePruner
+    ):
```
crcrpar, line 118: Does this work?
```suggestion
    elif isinstance(pruner, optuna.pruners.PercentilePruner):
```

Ref: https://github.com/optuna/optuna/blob/fe1e3e54f961897b871daa7fe23d92c4e1eaa408/optuna/pruners/_median.py#L4

crcrpar (commented):
Dropping some comments that I don't think must be accessed :slightly_smiling_face:

optuna/storages/_base.py
```diff
@@ -558,6 +558,15 @@ def get_n_trials(self, study_id: int, state: Optional[TrialState] = None) -> int
...
         raise NotImplementedError
 
+    @abc.abstractclassmethod
+    def get_url(self) -> Optional[str]:
+        """Get a path to a storage.
+
+        Returns:
+            A physical path to a storage.
+        """
+        raise NotImplementedError
```
toshihikoyanase, line 568: The design of this method seems reasonable because of the following reasons:
1. `storage` argument of `optuna.create_study` accepts `str`, and all storages can have the URL string.
1. Some storages may not have the URL string (i.e., in-memory storage for Dask integration discussed in https://github.com/optuna/optuna/issues/1766), but such storage can raise errors.
1. Preserving `None` for `InMemoryStorage` is consistent with the other part of Optuna.

`BaseStorage` may be able to have `url` property, but I prefer the current proposal because of the second reason. Raising errors in the property does not seem to be a good practice.

This change will affect the public interface of the storage, and I'd like to hear opinions of developers and other reviewers, so your comments are more than welcome. (cc: @keisuke-umezawa @hvy @ytsmiling)

optuna/storages/_cached_storage.py
```diff
@@ -165,6 +165,12 @@ def get_all_study_summaries(self) -> List[StudySummary]:
 
         return self._backend.get_all_study_summaries()
 
+    def get_url(self) -> Optional[str]:
+
+        if hasattr(self, "_backend"):
+            return self._backend.url
```
toshihikoyanase, line 171: A nit, but how about using the `get_url` for consistency?
```suggestion
            return self._backend.get_url()
```

optuna/storages/_redis.py
toshihikoyanase: If we provide the `get_url` method, I think `_url` can be private so that we can encourage users to use the method instead of the property.

toshihikoyanase (commented):
Thank you for your PR! I've heard that many AllenNLP users usually use config rather than Python script and this feature will be a great help of them.

In terms of the logic introduced in `AllenNLPExecutor`,  this PR employs a bit hacky way to use Optuna storages and pruners in AllenNLP processes, but it is acceptable because it mostly affects the code inside the integration module.

It also introduces a new `get_url` method to `optuna.storage.BaseStorage`, but I think it is simple and straightforward and it should have such interface as `RDBStorage` has the `url` property.

optuna/integration/allennlp.py
himkt: Amazing suggestion! Thank you.

optuna/storages/_cached_storage.py
himkt: Ah, thank you. You're right. 🙇

optuna/integration/allennlp.py
```diff
@@ -157,22 +311,47 @@ class AllenNLPPruningCallback(EpochCallback):
...
             ``validation_accuracy``.
+        trial:
+            A :class:`~optuna.trial.Trial` corresponding to the current evaluation of the
+            objective function.
     """
 
-    def __init__(self, trial: optuna.trial.Trial, monitor: str):
+    def __init__(
+        self,
+        trial: Optional[optuna.trial.Trial] = None,
```
himkt, line 324: Making `trial` and `monitor` keyword argument so that `AllenNLPPruningCallback` can be instantiated without them. This change is introduced for making it possible to create a callback with environment variables. (https://github.com/optuna/optuna/pull/1772/files#diff-194650369fb014d2ea8f428774c04276R336-R350)

optuna/integration/allennlp.py
```diff
@@ -157,22 +328,47 @@ class AllenNLPPruningCallback(EpochCallback):
...
     Args:
-        trial:
-            A :class:`~optuna.trial.Trial` corresponding to the current evaluation of the
-            objective function.
         monitor:
             An evaluation metric for pruning, e.g. ``validation_loss`` or
             ``validation_accuracy``.
+        trial:
+            A :class:`~optuna.trial.Trial` corresponding to the current evaluation of the
+            objective function.
```
crcrpar, line 336: `trial` should be put top of the `Args` block because the order of arguments of dunder init is trial and monitor.

himkt: You're right, thank you! v

optuna/integration/allennlp.py
```diff
@@ -111,6 +254,26 @@ def __init__(
         else:
             self._include_package = include_package
 
+        system_attrs = {
```
toshihikoyanase, line 257: As an alternative approach, we can get the storage URL without changing the storage API. How about separating the PR to implement `Storage.get_url()`?

```
storage = trial.study._storage
if isinstance(storage, RDBStorage):
    url = storage.url
elif isinstance(storage, RedisStorage):
    url = storage._url
elif isinstance(storage, _CachedStorage):
    assert isinstance(storage._backend, RDBStorage)
    url = storage._backend.url
else:
    raise ValueError("Storage must be RDBStorage or RedisStorage.")
```

himkt: > How about separating the PR to implement Storage.get_url()?

Thank you for the suggestion, I agree with you.
I revised my PR (262f868 , 9b200af)  to remove `get_url` from each storage.

I rename `_url` in `RedisStorage` with `url` for consistency with `RDBStorage`. After this PR, I'll work on making `url` private in `RDBStorage` and `RedisStorage` at the same time.

Also, let me note `url` would be an empty string when no `url` can be loaded. It is because `AllenNLPExecutor` itself works on InMemoryStorage.

hvy: Thanks a lot! Fewer files are now involved and it looks a bit cleaner. A small comment on the latest changes though, I think there's little benefit in aligning `RedisStorage`'s URL attribute to `RDBStorage` making it public. The `RDBStorage` has several of its attributes made public already and it's not just the `url`. We should probably limit the scope of the visibility matter to `RDBStorage`. It's always easier to make a private attribute public, than the other way around.

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,142 @@
...
+
+        @classmethod
+        def register(cls: Any, *args: Any, **kwargs: Any) -> Callable:
+            def wrapper(subclass: Any, *args: Any, **kwargs: Any) -> None:
+                return subclass
+
+            return wrapper
+
+
+PPID = os.getppid()
```
hvy, line 47: Is this constant intentionally defined public?

optuna/integration/allennlp.py
```diff
@@ -163,16 +346,42 @@ class AllenNLPPruningCallback(EpochCallback):
...
+                and trial_id is not None
+                and monitor is not None
+                and storage is not None
+            ):
+                study = load_study(study_name, storage, pruner=_create_pruner())
+                self._trial = Trial(study, int(trial_id))
+                self._monitor = monitor
+            else:
+                message = "Fail to load study.\n"
+                message += "AllenNLPPruningCallback works only with Optuna with RDB or Redis."
```
hvy, line 383: A nit. It might not be immediately obvious for the general users that "RDB" and "Redis" are alternatives to the `InMemoryStorage`.
```suggestion
                message += "AllenNLPPruningCallback works only with Optuna with RDB or Redis storages."
```
(Sorry I did not check the line length).

optuna/integration/allennlp.py
hvy, line 47: A comment will surely help future maintainers, especially as you're using the parent id and not the id of the current process. Could you explain how this variable and the environment variables are used to communicate between processes?

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,142 @@
...
+    elif isinstance(pruner, optuna.pruners.ThresholdPruner):
+        kwargs["lower"] = pruner._lower
+        kwargs["upper"] = pruner._upper
+        kwargs["n_warmup_steps"] = pruner._n_warmup_steps
+        kwargs["interval_steps"] = pruner._interval_steps
+    elif isinstance(pruner, optuna.pruners.SuccessiveHalvingPruner):
+        kwargs["min_resource"] = pruner._min_resource
+        kwargs["reduction_factor"] = pruner._reduction_factor
+        kwargs["min_early_stopping_rate"] = pruner._min_early_stopping_rate
+
```
hvy, line 147: I think we'll end up trying to re-instantiate a pruner with empty kwargs `{}` afterwards if an unrecognized pruner is given here. I don't think it's expected and we might want to raise a warning here in an `else`, if not an error.

hvy (changes requested):

optuna/integration/allennlp.py
himkt: Thank you for your suggestion. It makes sense. I made `url` private in `RedisStorage` at c537aa2.

optuna/integration/allennlp.py
himkt: I was not sure if I could put this on public space.
But now I made this private in 6af736b.

optuna/integration/allennlp.py
himkt: You're right. I added `else` in 0ddc47f.

optuna/integration/allennlp.py
himkt: I created `_add_ppid` and added some notes there. 6af736b :pray:

himkt:
@hvy @toshihikoyanase 

Thank you for the reviews. I revised PR based on comments and suggestions.
PTAL.

optuna/integration/allennlp.py
```diff
@@ -163,16 +362,42 @@ class AllenNLPPruningCallback(EpochCallback):
...
+                and trial_id is not None
+                and monitor is not None
+                and storage is not None
+            ):
+                study = load_study(study_name, storage, pruner=_create_pruner())
+                self._trial = Trial(study, int(trial_id))
+                self._monitor = monitor
+            else:
+                message = "Fail to load study.\n"
+                message += "AllenNLPPruningCallback works only with Optuna and RDB or Redis storages."  # NOQA
```
toshihikoyanase, line 399: Can we remove `# NOQA` as follows?
```suggestion
                message = (
                    "Fail to load study.\n"
                    "AllenNLPPruningCallback works only with Optuna and RDB or Redis storages."
                )
```

toshihikoyanase (commented):
Thank you for your update. I added some minor comments.
I'll check the coverage of test cases next.

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,158 @@
...
+def _get_environment_variables_for_trial() -> Dict[str, Optional[str]]:
+    return {
+        "study_name": os.getenv(_add_ppid("OPTUNA_ALLENNLP_STUDY_NAME")),
+        "trial_id": os.getenv(_add_ppid("OPTUNA_ALLENNLP_TRIAL_ID")),
+        "storage": os.getenv(_add_ppid("OPTUNA_ALLENNLP_STORAGE_NAME")),
+        "monitor": os.getenv(_add_ppid("OPTUNA_ALLENNLP_MONITOR")),
+    }
+
+
+def _get_environment_variables_for_pruner() -> Dict[str, Optional[str]]:
```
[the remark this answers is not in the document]
himkt: I'm sorry I didn't notice it. You're right.

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,156 @@
...
+
+    try:
+        return int(value)
+    except ValueError:
+        try:
+            return float(value)
+        except ValueError:
+            if value == "True":
+                return True
+            elif value == "False":
```
hvy, line 108: A nit. Since we're returning if the preceding `if`.
```suggestion
            if value == "False":
```

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,156 @@
...
+                return False
+
+    return value
+
+
+def _get_environment_variables_for_trial() -> Dict[str, Optional[str]]:
+    return {
+        "study_name": os.getenv(_add_ppid("OPTUNA_ALLENNLP_STUDY_NAME")),
```
hvy, line 116: How about defining (most of) these constant strings at global scope? That'd prevent silly bugs introduced by typos. You can actually have them concatenated with the parent pid when constructing the constants.

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,156 @@
...
+    elif isinstance(pruner, optuna.pruners.ThresholdPruner):
+        kwargs["lower"] = pruner._lower
+        kwargs["upper"] = pruner._upper
+        kwargs["n_warmup_steps"] = pruner._n_warmup_steps
+        kwargs["interval_steps"] = pruner._interval_steps
+    elif isinstance(pruner, optuna.pruners.SuccessiveHalvingPruner):
+        kwargs["min_resource"] = pruner._min_resource
+        kwargs["reduction_factor"] = pruner._reduction_factor
+        kwargs["min_early_stopping_rate"] = pruner._min_early_stopping_rate
+    else:
```
toshihikoyanase, line 159: `NopPruner` is not supported. Could you add it?
```suggestion
    elif isinstance(pruner, optuna.pruners.NopPruner):
        pass
    else:
```

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,156 @@
...
+
+
+def _fetch_pruner_config(trial: optuna.Trial) -> Dict[str, Any]:
+    pruner = trial.study.pruner
+    kwargs = {}  # type: Dict[str, Any]
+
+    if isinstance(pruner, optuna.pruners.HyperbandPruner):
+        kwargs["min_resource"] = pruner._min_resource
+        kwargs["max_resource"] = pruner._max_resource
+        kwargs["reduction_factor"] = pruner._reduction_factor
```
toshihikoyanase, line 144: `HyperbandPruner` has some other states including `_pruners` and `_total_trial_allocation_budget` (https://github.com/optuna/optuna/blob/master/optuna/pruners/_hyperband.py#L143-L146). I guess it initializes the successive halving pruner every trial. Do you think if it affects optimization performance?

tests/integration_tests/allennlp_tests/test_allennlp.py
```diff
@@ -238,3 +238,58 @@ def objective(trial: optuna.Trial) -> float:
...
+            pass
+
+        def prune(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> bool:
+            return False
+
+    study = optuna.create_study(pruner=SomeNewPruner())
+    trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
+
+    with pytest.raises(ValueError):
+        optuna.integration.allennlp._fetch_pruner_config(trial)
```
toshihikoyanase, line 295: Could you add a test case to check if `_fetch_pruner_config` does not raise errors when it receives supported pruners?

toshihikoyanase (changes requested):
Thank you for your update. I added some minor comments on the pruners.

tests/integration_tests/allennlp_tests/test_allennlp.py
hvy, line 295: How about testing this using public API, i.e. the executor?

optuna/integration/allennlp.py
```diff
@@ -163,16 +360,44 @@ class AllenNLPPruningCallback(EpochCallback):
...
+            ):
+                study = load_study(study_name, storage, pruner=_create_pruner())
+                self._trial = Trial(study, int(trial_id))
+                self._monitor = monitor
+            else:
+                message = (
+                    "Fail to load study.\n"
+                    "AllenNLPPruningCallback works only with Optuna and RDB or Redis storages."
+                )
+                raise Exception(message)
```
hvy, line 400: Can we test this logic?

hvy, line 400: Can we provide more information and finer grained error handling by using a more concrete type of error?
```suggestion
                raise RuntimeException(message)
```

optuna/integration/allennlp.py
hvy: We should probably explain in the documentation that these arguments are types optional but are mandatory.

optuna/integration/allennlp.py
```diff
@@ -163,16 +360,44 @@ class AllenNLPPruningCallback(EpochCallback):
...
-        self._monitor = monitor
+        if trial is not None and monitor is not None:
+            self._trial = trial
+            self._monitor = monitor
+        else:
+            _environment_variables = _get_environment_variables_for_trial()
+            study_name = _environment_variables["study_name"]
+            trial_id = _environment_variables["trial_id"]
+            monitor = _environment_variables["monitor"]
+            storage = _environment_variables["storage"] or None
```
hvy, line 384: A comment for `or None` would be helpful.

optuna/integration/allennlp.py
```diff
@@ -163,16 +360,44 @@ class AllenNLPPruningCallback(EpochCallback):
...
 
         if allennlp.__version__ < "1.0.0":
             raise Exception("AllenNLPPruningCallback requires `allennlp`>=1.0.0.")
 
-        self._trial = trial
-        self._monitor = monitor
+        if trial is not None and monitor is not None:
+            self._trial = trial
+            self._monitor = monitor
+        else:
```
hvy, line 379: I guess we can end up here when user don't pass `trial` or `monitor`. The current exception about storage types could be misleading. How about handling this case with a different exception such as `ValueError` ?

hvy (changes requested):
I left some additional comments. Could you take a look? A bit concerned that the logic is difficult to test but I guess it's difficult.

optuna/integration/allennlp.py
toshihikoyanase: For example, the estimation of `max_resource` cannot be estimated correctly if the value of `max_resource` is `auto` and the resource (epoch) of each trial is different.

But, maybe this is an issue of `HyperbandPruner` and should be resolved by the pruner.

optuna/integration/allennlp.py
himkt: Thank you for the helpful suggestion. I defined parameters at the global scope in 65195ad.

tests/integration_tests/allennlp_tests/test_allennlp.py
himkt: I added `test_allennlp_pruning_callback_with_invalid_executor` in 07a9de9.

himkt: In 07a9de9, I added a parametrized test. In this test, I found some bugs. Thank you so much for the suggestion.

optuna/integration/allennlp.py
himkt: I revised the PR to check if `storage` is `None` 8d227cd with a specific message.
What do you think about this change? Do you think it'd be better to add more comments?

optuna/integration/allennlp.py
himkt: I added some explanations in 603308d. I'd like your feedback. :bow:

optuna/integration/allennlp.py
himkt: I'm not sure if I understand your intention correctly, I added some comments in 603308d and 2987f7d.
I divided the error into the one for storage (https://github.com/optuna/optuna/pull/1772/files#diff-194650369fb014d2ea8f428774c04276R420-R425) and the other (https://github.com/optuna/optuna/pull/1772/files#diff-194650369fb014d2ea8f428774c04276R405-R413).

Please correct me know if I misunderstand your comment.

optuna/integration/allennlp.py
```diff
@@ -9,17 +10,156 @@
...
 else:
-    EpochCallback = object  # NOQA
+    # `allennlp.training.EpochCallback` is a subclass of `Registerable`
+    # https://docs.allennlp.org/master/api/training/trainer/#epochcallback
+    # On the other hand, `EpochCallback` defined here is not `Registerable`,
+    # which causes a mypy checking feailure.
+    class EpochCallback:  # type: ignore
+        """Stub for EpochCallback.
+
+        This class is introduced for documentation CI.
```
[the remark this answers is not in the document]
himkt: You're right, thank you. I removed the mention to `documentation` in c415b49.

optuna/integration/allennlp.py
hvy: Sorry if I wasn't clear but I meant in the user facing documentation.

optuna/integration/allennlp.py
hvy: Looks good, thanks!

optuna/integration/allennlp.py
```diff
@@ -15,12 +18,166 @@
...
+
+            """
+
+            def wrapper(subclass: Any, *args: Any, **kwargs: Any) -> None:
+                return subclass
+
+            return wrapper
+
+
+PPID = os.getppid()
```
hvy, line 50: How about making this constant and the ones below private as we did before, by prefixing them with underscores?

tests/integration_tests/allennlp_tests/test_allennlp.py
```diff
@@ -238,3 +240,163 @@ def objective(trial: optuna.Trial) -> float:
...
+            {"percentile": 50.0, "n_startup_trials": 10, "n_warmup_steps": 1, "interval_steps": 3},
+        ),
+        (
+            "successive_halving",
+            {"min_resource": 3, "reduction_factor": 5, "min_early_stopping_rate": 1},
+        ),
+        ("threshold", {"lower": 0.0, "upper": 1.0, "n_warmup_steps": 3, "interval_steps": 2}),
+    ],
+)
+def test_allennlp_pruning_callback_with_executor(
```
toshihikoyanase, line 262: Can we simplify this test case by providing the pruner classes instead of the class name?
The following is just an example. Please fix the type hint of `pruner_class` if you employ this approach.

```python
@pytest.mark.parametrize(
    "pruner_class,pruner_kwargs",
    [
        (optuna.pruners.HyperbandPruner, {"min_resource": 3, "max_resource": 10, "reduction_factor": 5}),
        (optuna.pruners.MedianPruner, {"n_startup_trials": 8, "n_warmup_steps": 1, "interval_steps": 3}),
        (optuna.pruners.NopPruner, {}),
        (
            optuna.pruners.PercentilePruner,
            {"percentile": 50.0, "n_startup_trials": 10, "n_warmup_steps": 1, "interval_steps": 3},
        ),
        (
            optuna.pruners.SuccessiveHalvingPruner,
            {"min_resource": 3, "reduction_factor": 5, "min_early_stopping_rate": 1},
        ),
        (optuna.pruners.ThresholdPruner, {"lower": 0.0, "upper": 1.0, "n_warmup_steps": 3, "interval_steps": 2}),
    ],
)
def test_allennlp_pruning_callback_with_executor(
    pruner_class: Callable, pruner_kwargs: Dict[str, Union[int, float]]
) -> None:
    input_config_file = (
        "tests/integration_tests/allennlp_tests/example_with_executor_and_pruner.jsonnet"
    )

    def run_allennlp_executor(pruner: optuna.pruners.BasePruner) -> None:
        study = optuna.create_study(
            direction="maximize",
            pruner=pruner,
            storage=storage,
        )
        trial.suggest_float("DROPOUT", 0.0, 0.5)
        executor = optuna.integration.AllenNLPExecutor(trial, input_config_file, serialization_dir)
        executor.run()

    with tempfile.TemporaryDirectory() as tmp_dir:
        pruner_name = pruner_class.__name__
        os.mkdir(os.path.join(tmp_dir, pruner_name))
        storage = "sqlite:///" + os.path.join(tmp_dir, pruner_name, "result.db")
        serialization_dir = os.path.join(tmp_dir, pruner_name, "allennlp")

        pruner = pruner_class(**pruner_kwargs)
        run_allennlp_executor(pruner)
        ret_pruner = optuna.integration.allennlp._create_pruner()

        assert isinstance(ret_pruner, pruner_class)
        for key, value in pruner_kwargs.items():
            assert getattr(ret_pruner, "_{}".format(key)) != value

```

optuna/integration/allennlp.py
himkt: 7b2c92f made `PPID` private. :pray:

tests/integration_tests/allennlp_tests/test_allennlp.py
himkt: It really makes sense! I applied your suggestion in 90f3eec, thank you.

optuna/integration/allennlp.py
```diff
@@ -15,12 +18,166 @@
...
+To avoid this hazard, we add ID of a parent process to each key of
+environment variables.
+"""
+PREFIX = "{}_OPTUNA_ALLENNLP".format(_PPID)
+MONITOR = "{}_MONITOR".format(PREFIX)
+PRUNER_CLASS = "{}_PRUNER_CLASS".format(PREFIX)
+PRUNER_KEYS = "{}_PRUNER_KEYS".format(PREFIX)
+STORAGE_NAME = "{}_STORAGE_NAME".format(PREFIX)
+STUDY_NAME = "{}_STUDY_NAME".format(PREFIX)
+TRIAL_ID = "{}_TRIAL_ID".format(PREFIX)
```
toshihikoyanase, line 68: Can we make these variables private similar to `_PPID`?
```suggestion
_PREFIX = "{}_OPTUNA_ALLENNLP".format(_PPID)
_MONITOR = "{}_MONITOR".format(PREFIX)
_PRUNER_CLASS = "{}_PRUNER_CLASS".format(PREFIX)
_PRUNER_KEYS = "{}_PRUNER_KEYS".format(PREFIX)
_STORAGE_NAME = "{}_STORAGE_NAME".format(PREFIX)
_STUDY_NAME = "{}_STUDY_NAME".format(PREFIX)
_TRIAL_ID = "{}_TRIAL_ID".format(PREFIX)
```

himkt: Sorry for not considering the suggestion https://github.com/optuna/optuna/pull/1772#discussion_r488378436 well. 🙇 
I made these constants private in 41e654f.

optuna/integration/allennlp.py
```diff
@@ -112,6 +286,37 @@ def __init__(
...
+
+        if isinstance(storage, optuna.storages.RDBStorage):
+            url = storage.url
+        elif isinstance(storage, optuna.storages.RedisStorage):
+            url = storage._url
+        elif isinstance(storage, optuna.storages._CachedStorage):
+            assert isinstance(storage._backend, optuna.storages.RDBStorage)
+            url = storage._backend.url
+        else:
+            url = ""
```
toshihikoyanase, line 299: When I use the in-memory storage, I got the following error due to this line:
```console
[W 2020-09-16 15:24:13,416] Trial 0 failed because of the following error: ArgumentError("Could not parse rfc1738 URL from string ''")
```
I think it is not easy for users to find the cause since they do not provide the storage URL by themselves. Is it possible to tell users to use the `RDBStorage` or `RedisStorage` when they use in-memory storage?

himkt: Thank you for the catch.
I had to check whether `storage` is an empty string or not.
In d3f8837, I update the PR for it.

Now the following message appears when users attempt to use in-memory storage.

```
RuntimeError: If you want to use AllenNLPExecutor and AllenNLPPruningCallback, you have to use RDB or Redis storage.
```

optuna/integration/allennlp.py
himkt: Thank you, it makes sense. :pray:
78822da I added a note in the docstring based on comments in the codebase.

hvy:
Will the usage of `optuna_pruner` be documented somewhere?

optuna/integration/allennlp.py
toshihikoyanase: Thank you for your update. I confirmed the error message.
SGTM although the error handling is getting a bit complicated by https://github.com/optuna/optuna/commit/d3f8837bd69f76410433525e42133ae888e5e7f2.

toshihikoyanase (approved):
LGTM.

@hvy Please merge after your review.

---
> Will the usage of `optuna_pruner` be documented somewhere?

I also think that we can add `optuna_pruner` to the configuration in `examples/allennlp/classifier.jsonnet`, but it's optional and not a strong opinion.

optuna/integration/allennlp.py
```diff
@@ -15,12 +18,166 @@
...
+        def register(cls: Any, *args: Any, **kwargs: Any) -> Callable:
+            """Stub method for `EpochCallback.register`.
+
+            This method has the same signature as
+            `Registrable.register <https://docs.allennlp.org/master/
+            api/common/registrable/#registrable>`_ in AllenNLP.
+
+            """
+
+            def wrapper(subclass: Any, *args: Any, **kwargs: Any) -> None:
```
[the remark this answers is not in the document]
himkt: You're right, thank you!

himkt:
d8eb667 I added `optuna_pruner` to `examples/allennlp/classifier.jsonnet` and update `examples/allennlp/allennlp_jsonnet.py` to run optimization with Hyperband.
And also I added the reference to allennlp-guide in docstring in e133e17.

hvy (approved):
LGTM! Thanks for the long running effort. I would just say that the logic is a bit tricky and not necessarily robust to (internal) changes. I believe we can discuss alternatives such as pickling if needed.

hvy:
Updated the title a bit, hope you're fine with that 🙂

himkt:
I'd really appreciate patient reviews from all reviewers. Thank you so much.
I agree that the current implementation is hacky and fragile, I'll work on improving mechanism.

Issue #1718, opened 2020-08-25
LightGBMTunerCV stepwise tuning got an error with lightgbm==3.0.0rc1


nyanhi:
`LightGBMTunerCV.run()` got LightGBMError after updating `lightgbm` version==`3.0.0rc1`. 
While it's performing stepwise tuning, `lightgbm.basic` raised below error message at the step of `min_data_in_leaf` and tuning is interrupted. 
A workaround seems to be adding `feature_pre_filter=false` like following as the error message suggests:
```
params = {
    "objective": "binary",
    "metric": "binary_logloss",
    "verbosity": -1,
    "boosting_type": "gbdt",
    "feature_pre_filter": False,
}
``` 
## Environment

- Optuna version: 2.0.0
- Python version: 3.7
- OS: linux/amd64 (Docker image: optuna/optuna:v2.0.0-py3.7)
- Lightgbm version: 3.0.0rc

## Error messages, stack traces, or logs

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/optuna/study.py", line 709, in _run_trial
    result = func(trial)
  File "/usr/local/lib/python3.6/dist-packages/optuna/integration/_lightgbm_tuner/optimize.py", line 302, in __call__
    cv_results = lgb.cv(self.lgbm_params, self.train_set, **self.lgbm_kwargs)
  File "/usr/local/lib/python3.6/dist-packages/lightgbm/engine.py", line 554, in cv
    train_set._update_params(params) \
  File "/usr/local/lib/python3.6/dist-packages/lightgbm/basic.py", line 1436, in _update_params
    raise LightGBMError(decode_string(_LIB.LGBM_GetLastError()))
lightgbm.basic.LightGBMError: Reducing `min_data_in_leaf` with `feature_pre_filter=true` may cause unexpected behaviour for features that were pre-filtered by the larger `min_data_in_leaf`.
You need to set `feature_pre_filter=false` to dynamically change the `min_data_in_leaf`.
```

## Steps to reproduce

1. Update lightgbm to version==`3.0.0rc1`
2. Execute the example code in https://github.com/optuna/optuna/blob/master/examples/lightgbm_tuner_cv.py
3. The error occurs at 
https://github.com/optuna/optuna/blob/4680f36a470ffb9ead89abf65dcc7e7533fd789f/examples/lightgbm_tuner_cv.py#L24-L28

toshihikoyanase:
[...]

PR #1774, opened 2020-09-01
Disable `feature_pre_filter` in `LightGBMTuner`

hvy:
## Motivation

Fixes https://github.com/optuna/optuna/issues/1718.

## Description of the changes

Always disables `feature_pre_filter`. See the above issue for details.


toshihikoyanase (approved):
Thank you for your PR. The change looks good to me. I also confirmed that this change does not affect `LightGBM==2.3.1` since it does not have the `feature_pre_filter` option and simply ignores it.

I think we can mention this behavior in the documentation, but it may be too specific.

toshihikoyanase:
@smly I think you are so busy, but please let me know if you have any comments on the `feature_pre_filter` argument.

HideakiImamura (approved):
Thanks, LGTM!

PR #1777, opened 2020-09-01
Avoid using experimental decorator on CmaEsSampler

c-bata:
## Motivation

I think that current usage of `experimental` decorator is not intuitive.

* It cannot update docstring for constructor options.
* We defined `_raise_experimental_warning_for_restart_strategy` and `_raise_experimental_warning_for_consider_pruned_trials` just for showing warning messages.
* It doesn't reduce the code size very much.

## Description of the changes

Avoid using experimental decorator.

Output:

ScreenShot 2020-09-02 0 28 54


c-bata:
I will also remove an experimental decorator from SkoptSampler if we reach the concensus.

hvy:
Thank you for the suggestion. You are right that the current usage is suboptimal. Maybe we can avoid using the decorator after all, as these arguments are quite few anyway. I'd like to hear opinions from other developers.

I believe it's a matter of DRY otherwise (for instance, say we'd want to start logging these warnings in the future).So there are pros to isolating the logic in `_experimental.py`.  What we could do as an alternative is to provide a different interface. A simple one would be a function (it'd keep the DRY and address your second and possible third points). Do you find this over-engineering?

```python
def foo(a, b=None):
    if b is not None:
        _experimental("b")  # In practice, we just need to pass the keyword. 
```

c-bata:
Thank you for your comment!

I once considered to introduce a function like you suggested but I cannot imagine such a situation (e.g. logging these warnings in the future). At least, It looks a common practice to call `warrnings.warn()` directly in some popular Python projects which I've ever contributed to (such as scikit-learn and Django).

I personally prefer to avoid using any wrapper functions. But I'd also like to hear opinions from other developers.

hvy:
Oh, so it might be an anti-pattern? Just to clarify, do you think we should still keep the decorator and use it for classes/functions/methods? To elaborate a bit on my previous comment, if we keep `_experimental.py` (note that there's also `_deprecated.py`), I would rather try to isolate the logic the best we can (i.e., everything besides the docs for arguments).

toshihikoyanase:
> We defined `_raise_experimental_warning_for_restart_strategy` and `_raise_experimental_warning_for_consider_pruned_trials` just for showing warning messages.

> there are pros to isolating the logic in `_experimental.py`.

I feel both opinions are reasonable. IMO, we can accept the suggested change in this PR as an exceptional case and discuss alternative way to keep the DRY later.

If we apply `@experimental` to classes or functions, we have almost no design choices (e.g., `warnings.warn` or `logging.warning`, the template of the message, the place we insert the notes in the docstring, etc). It will reduce development costs as well as review costs.
On the other hand, when we want to annotate arguments, the developers need to find the message template and insert the message by themselves, and we cannot reduce the cost so much.

> A simple one would be a function (it'd keep the DRY and address your second and possible third points).

Some OSS projects like `chainer` and `mlflow` also provide `_experimental`-like function, and I think it's a good idea to keep the convention to notify experimental information (`warnings.warn` with `ExperimentalWarning`) to users.

---
Just FYI, but I come up with an alternative approach that uses the property. 

<details>
<summary>Apply experimental decorator to property setter</summary>

If we add the `_consider_pruned_trials` property and apply `experimental` decorator to the setter of the property, the experimental warning will be raised at `self._consider_pruned_trials = consider_pruned_trials`.

```python
    def __init__(...):
        ....
        self._consider_pruned_trials = consider_pruned_trials  # type: ignore

    @property
    def _consider_pruned_trials(self) -> bool:
        return self.__consider_pruned_trials

    @_consider_pruned_trials.setter  # type: ignore
    @experimental("2.0.0")
    def _consider_pruned_trials(self, value: bool) -> None:
        self.__consider_pruned_trials = value
```

```console
>>> import optuna
>>> optuna.integration.SkoptSampler(consider_pruned_trials=True)
/Users/yanase/pfn/code/optuna/optuna/integration/skopt.py:118: ExperimentalWarning: _consider_pruned_trials is experimental (supported from v2.0.0). The interface can change in the future.
  self._consider_pruned_trials = consider_pruned_trials  # type: ignore
```

But, it increases the lines and we need `# type: ignore` as a workaround for `mypy` (see https://github.com/python/mypy/issues/1362). I don't think it's a good idea.

</details>

c-bata:
> Oh, so it might be an anti-pattern? Just to clarify, do you think we should still keep the decorator and use it for classes/functions/methods?

@hvy I'm not sure it's an anti-pattern, just I thought it isn't a common practice.
I personally prefer to avoid using experimental decorator even if it's a classes, functions and methods.
This is not a strong opinion, so I only suggests to avoid using for constructor options in this pull request.
I think the current implementation of an experimental decorator (and a deprecation decorator) to overwrite docstring is a complex.

But according to the @toshihikoyanase's comment, some famous Python projects (chainer and mlflow) introduce similar utility functions.
So my recognition might not be correct. Thank you for this information!

---

> I feel both opinions are reasonable. IMO, we can accept the suggested change in this PR as an exceptional case and discuss alternative way to keep the DRY later.

I agree with your suggestion.

> Just FYI, but I come up with an alternative approach that uses the property.

@toshihikoyanase I guess your property setter based approach shows warning messages even if not providing `consider_pruned_trials` argument like `optuna.integration.SkoptSampler()`.

HideakiImamura:
Thanks for the important suggestion! In fact, applying the experimental decorator to arguments is debatable. Personally, I　think we don't have to use the experimental decorator for arguments, as @c-bata suggested.

I think we should use an experimental decorator for functions and classes. As @toshihikoyanase said, it will reduce development costs as well as review costs. Additionally, I'd like to know what is the benefit of being able to modify docstrings to be rich. The current default docstring seems to provide enough information.

c-bata:
Thank you for your opinions.
As for constructor options, it seems that we almost reach the consensus.
I pushed b4548b28 for SkoptSampler.

HideakiImamura (approved):
Thanks for the swift actions! LGTM.

HideakiImamura:
As a followup task, we should update the [Coding Style Conventions](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions).

hvy:
Sorry the delayed response. Only speaking about arguments, as they are still quite few, we can probably avoid using the decorator in these places. I agree that having to define private methods for this purpose isn't very clever. Again, sorry for the delay but let me continue with the review given that and leave the decorator outside the scope of this PR in terms of how it can be applied to classes/methods/functions.

optuna/integration/skopt.py
```diff
@@ -81,8 +82,8 @@ def objective(trial):
         consider_pruned_trials:
             If this is :obj:`True`, the PRUNED trials are considered for sampling.
 
-            .. note::
-                Added in v2.0.0 as an experimental feature. The interface may change in newer
+            .. versionadded:: 2.0.0
```
hvy, line 85: This change seem somewhat orthogonal to the intention but could you comment on the motivation? I personally find `.. note::` to be easier to notice in the docs.

c-bata: I just followed following Optuna's style convention. But as you said, it seemed orthogonal with the template of experimental decorator.
https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#write-versionadded-directive-to-the-doc-of-new-features

I'm not sure which directives should be used in Optuna, but I reverted this change for now.
Thanks!

HideakiImamura (approved):
Thanks for the update. LGTM, again! The conflicts should be resolved by merging the master branch.

HideakiImamura:
The current master branch is broken due to `isort`. I will fix it soon. Sorry for the confusion.

hvy (approved):
LGTM!

Sorry for the late review, and thanks for the quick changes as always.

optuna/integration/skopt.py
hvy: Me neither. I haven't exactly understood when to use `versionadded`. It might be suitable for experimental features since the version number is a part of it, but at the same time, it seems like a very general features given the brief documentation by sphinx. 🤔

PR #1786, opened 2020-09-04
proposal of `MutualInformationImportanceEvaluator`

kstoneriv3:
## Motivation

I propose a new parameter importance evaluation method using mutual information between the objective function and each parameter. The current implementation is just a draft, so I would like to discuss its design here.

#### Negative estimate of mutual information
Since the default estimation method of the mutual information in the proposed implementation use sample average as following, the estimate of the mutual information can be negative. Here, the p^ stands for 1-d or 2-d parzen estimators. 

To avoid this issue, we can sample x_i and y_i from the estimator of the joint distribution, p^(x,y). This is equivalent to calculating mutual information between p^(x)p^(y) and p^(x,y) using the Monte Carlo method. 
This Monte Carlo estimation is supported in addition to the sample average estimate. It can be computed if `n_MC` is specified in the `__init__` of the `MutualInformationImportanceEvaluator`. However, at the moment, I am not sure if we should support the Monte Carlo estimation.

#### Credibility of estimated importance
On several benchmark problems, the `MutualInformationImportanceEvaluator` returns somewhat inconsistent outputs from the `FanovaImportanceEvaluator`. (Thanks to @hvy for pointing this out!)
Moreover, the estimated values tend to be either 
- unstable when using samples average (e.g. sometimes estimate of mutual information becomes negative).
- uninformative when using the Monte Carlo average (all estimates tend to be fairly positive (0 <<) and differences in outputted importance estimates becomes smaller)
So whether we should include this to Optuna should be discussed in the first place.

## Description of the changes

Several files are added to `optuna.importance`. Mutual information is estimated by using multivariate Parzen estimators, which is originally implemented for multivariate TPE sampler.

## TODOs
- [ ] discuss whether or not to support mutual information importance evaluator in Optuna by comparing it with existing evaluators.
- [ ] determine the range of functionalities it should support
- [ ] polish the design of the code
- [ ] implement test code

### DIFF
--- a/optuna/importance/_mutual_information/_evaluator.py
+++ b/optuna/importance/_mutual_information/_evaluator.py
@@ -17,10 +17,10 @@ from optuna.trial import TrialState
 
 class MutualInformationImportanceEvaluator(BaseImportanceEvaluator):
 
-    def __init__(self) -> None:
-        pass
+    def __init__(self, n_MC: Optional[int]=None) -> None:
+        self._n_MC = n_MC
 
-    def evaluate(self, study: Study, params: Optional[List[str]] = None) -> Dict[str, float]:
+    def evaluate(self, study: Study, params: Optional[List[str]] = None, n_MC=None) -> Dict[str, float]:
 
         search_space = study.best_trial.distributions
         if params is None:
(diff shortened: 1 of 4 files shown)


optuna/importance/_mutual_information/_evaluator.py
```diff
@@ -0,0 +1,120 @@
...
+from optuna.study import Study
+from optuna.trial import TrialState
+
+
+class MutualInformationImportanceEvaluator(BaseImportanceEvaluator):
+    def __init__(self, n_MC: Optional[int] = None) -> None:
+        self._n_MC = n_MC
+
+    def evaluate(
+        self, study: Study, params: Optional[List[str]] = None, n_MC: Optional[int] = None
```
crcrpar, line 23: Where do you use the argument of  `n_MC`?

optuna/importance/_mutual_information/_evaluator.py
```diff
@@ -0,0 +1,120 @@
...
+                rng = np.random.RandomState()
+                mc_samples = mpe_joint.sample(rng, self._n_MC)
+                # We calculate log_probs.
+                score_log_prob = mpe_score.log_pdf({"score": mc_samples["score"]})
+                param_log_prob = mpe_param.log_pdf({param_name: mc_samples[param_name]})
+                joint_log_prob = mpe_joint.log_pdf(mc_samples)
+
+                mi[param_name] = np.mean(joint_log_prob - score_log_prob - param_log_prob)
+
+        mi = {k: v for k, v in sorted(mi.items(), key=lambda item: item[1], reverse=True)}
```
crcrpar, line 83: If the order of `mi` is important, we should use `OrderedDict` as dicts of some old-ish Python do not keep insertion order.

optuna/importance/_mutual_information/_parzen_estimator.py
```diff
@@ -0,0 +1,390 @@
...
+from typing import NamedTuple
+from typing import Optional
+from typing import Tuple
+
+import numpy as np
+import scipy.special
+import scipy.stats
+
+from optuna import distributions
+from optuna.distributions import BaseDistribution
```
crcrpar, line 12: [question] why do you import only BaseDistribution while not the other concrete distribution classes?

crcrpar (commented):
first of all, great work. I do know this is marked "Draft", but dropping some trivial comments :wink:

optuna/importance/_mutual_information/_evaluator.py
kstoneriv3: I just forgot to delete it. Thank you!

optuna/importance/_mutual_information/_evaluator.py
kstoneriv3: I do not think the order of `mi` matters here, but I appreciate your caution. I might make such a mistake somewhere else without it.

optuna/importance/_mutual_information/_parzen_estimator.py
kstoneriv3: I did not do this intentionally. Probably, I imported `BaseDistribution` because it is used for type hints. Lengthy type hints are not easy to read. What do you think is the best practice in this situation? Should all concrete distribution classes be imported?

optuna/importance/_mutual_information/_evaluator.py
kstoneriv3: Ah, yes, I now understood your point. The order actually matters in the last line. I will fix it. Thank you!

kstoneriv3:
@crcrpar Thank you so much for leaving comments swiftly! 
At the moment, we should probably consider if we actually need this functionality in the first place. After making PR, I took a look at comparisons of the `MutualInformationImportanceEvaluator` and existing evaluator and found some unstableness and limitations with this implementation/approach. I updated my first comments by adding the section "Credibility of estimated importance", hoping to clarify this situation.

kstoneriv3:
As there are lots of problems with this functionality, I will close this PR and open an Issue instead.

PR #1788, opened 2020-09-04
Fix fANOVA for `IntLogUniformDistribution`

hvy:
## Motivation

See below.

## Description of the changes

Fixes a bug in fANOVA where `IntLogUniformDistribution` parameters used to raise an assertion error.


HideakiImamura (approved):
Good catch! LGTM!

sile (approved):
LGTM!

PR #1791, opened 2020-09-07
Disable `--cache-from` if trigger of docker image build is `release`

toshihikoyanase:
## Motivation

The docker image build for `v2.1.0` was failed due to the missing image cache as follows:

```console
Run docker pull "${HUB_TAG}"
  docker pull "${HUB_TAG}"
  shell: /bin/bash -e {0}
  env:
    DOCKER_HUB_BASE_NAME: optuna/optuna
    HUB_TAG: optuna/optuna:v2.1.0-py3.7
Error response from daemon: manifest for optuna/optuna:v2.1.0-py3.7 not found: manifest unknown: manifest unknown
```

See https://github.com/optuna/optuna/actions/runs/242435515 for details.

## Description of the changes
This PR disables `--cache-from` option of the `docker build` and build the images from scratch if the trigger is `release`.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -46,6 +46,7 @@ jobs:
 
     - name: Build the Docker image
       run: |
+        # If the event is release, cache is not available because the image tag includes the Optuna version.
         if [ "${{ github.event_name }}" != 'release' ]; then
           docker pull "${HUB_TAG}"
           CACHE_FROM="--cache-from=${HUB_TAG}"
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -47,12 +47,11 @@ jobs:
     - name: Build the Docker image
       run: |
         if [ "${{ github.event_name }}" = 'release' ]; then
-          export CACHE_FROM=""
+          docker build . --build-arg PYTHON_VERSION=${{ matrix.python_version }} --build-arg BUILD_TYPE=${{ matrix.build_type }} --file Dockerfile --tag "${HUB_TAG}"
         else
           docker pull "${HUB_TAG}"
-          export CACHE_FROM="--cache-from ${HUB_TAG}"
+          docker build --cache-from "${HUB_TAG}" . --build-arg PYTHON_VERSION=${{ matrix.python_version }} --build-arg BUILD_TYPE=${{ matrix.build_type }} --file Dockerfile --tag "${HUB_TAG}"
         fi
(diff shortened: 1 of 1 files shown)


.github/workflows/dockerimage.yml
```diff
@@ -44,13 +44,13 @@ jobs:
...
         echo "::set-env name=HUB_TAG::${DOCKER_HUB_BASE_NAME}:${TAG_NAME}"
 
-    - name: Pull the Docker image
-      run: |
-        docker pull "${HUB_TAG}"
-
     - name: Build the Docker image
       run: |
-        docker build --cache-from "${HUB_TAG}" . --build-arg PYTHON_VERSION=${{ matrix.python_version }} --build-arg BUILD_TYPE=${{ matrix.build_type }} --file Dockerfile --tag "${HUB_TAG}"
+        if [ "${{ github.event_name }}" != 'release' ]; then
```
hvy, line 49: How about including an explanatory comment about why this condition is required and that we'll get a "cache miss" on new releases?

.github/workflows/dockerimage.yml
```diff
@@ -44,13 +44,13 @@ jobs:
...
+          docker pull "${HUB_TAG}"
+          CACHE_FROM="--cache-from=${HUB_TAG}"
+        fi
+        docker build ${CACHE_FROM} . --build-arg PYTHON_VERSION=${{ matrix.python_version }} --build-arg BUILD_TYPE=${{ matrix.build_type }} --file Dockerfile --tag "${HUB_TAG}"
```
hvy, line 53: Do we have to define `CACHE_FROM`, for instance as an empty string, in the case of releases?

hvy (changes requested):
Thanks for the PR! The approach overall seem good to me. I left some comments and questions if you could take a look.

.github/workflows/dockerimage.yml
toshihikoyanase: That makes sense. I added a comment in https://github.com/optuna/optuna/pull/1791/commits/c3fc8db6477906dfc49a01e4c06a748e7c1307bd.

.github/workflows/dockerimage.yml
toshihikoyanase: I think it is optional since the shell simply ignores undefined variables by default, but I added the missing `else` clause as you mentioned.

hvy: You're right. Still, thanks for the change. I think it aids comprehensibility.

hvy (approved):
LGTM!

HideakiImamura (approved):
Thanks, LGTM!

Issue #1500, opened 2020-07-09
lgb.LightGBMTunerCV: TrialState.FAIL because the returned value from the objective function cannot be casted to float


konradsemsch:
The lgb.LightGBMTuner seems to run for a number of trials and then breaks unexpectedly. The strange thing is that it runs without problems for a number of steps before the error occurs.

## Expected behavior

This is my code piece:

```python
self._parameters_set_2 = {
                "objective": "regression",
                "metric": "rmse", 
                "verbosity": -1,
                "boosting_type": "gbdt"
}

data_train_lgb = lgb.Dataset(
                self._pipeline_prepare.fit_transform(self._data["training"]["X"]), 
                label = self._data["training"]["y"]
)

tuner_lgb = lgb.LightGBMTunerCV(
                    self._parameters_set_2, 
                    data_train_lgb,  
                    verbose_eval = 1,
                    num_boost_round = 100, 
                    early_stopping_rounds = 10, 
                    time_budget = 10,
                    folds = self._data["validation_time_cv"]
                )
```

## Environment

- Optuna version: 1.5.0  
- Python version: 3.7
- OS: MacOS 10.15.3 (19D76)

## Error messages, stack traces, or logs

```
[I 2020-07-09 11:33:03,956] Finished trial#29 with value: 0.07808463369078895 with parameters: {'bagging_fraction': 0.9527650536197096, 'bagging_freq': 2}. Best is trial#29 with value: 0.07808463369078895.
bagging, val_score: 0.078085:  30%|######################################1                                                                                        | 3/10 [00:01<00:02,  2.38it/s][1]     cv_agg's rmse: 0.0783785 + 0.039904
[2]     cv_agg's rmse: 0.0783836 + 0.0398658
[3]     cv_agg's rmse: 0.0783535 + 0.0397835
[4]     cv_agg's rmse: 0.0783261 + 0.0398097
[5]     cv_agg's rmse: 0.0783087 + 0.0398446
[6]     cv_agg's rmse: 0.0782265 + 0.0396464
[7]     cv_agg's rmse: 0.0783195 + 0.0393913
[8]     cv_agg's rmse: 0.0783578 + 0.0392715
[9]     cv_agg's rmse: 0.0783393 + 0.0391258
[10]    cv_agg's rmse: 0.0784091 + 0.0390127
[11]    cv_agg's rmse: 0.0785363 + 0.0389685
[12]    cv_agg's rmse: 0.0785213 + 0.0386974
[...]

PR #1805, opened 2020-09-10
Copy datasets before objective evaluation.

toshihikoyanase:
## Motivation

As discussed in #1718, LightGBM datasets have parameters and they will be tuned during the model training. I guess the dataset parameters can be the cause of #1500.

## Description of the changes

This PR simply copied the dataset before training the booster models. Currently, I used the shallow copy, and I confirmed that I could enable `feature_pre_filter` when I ran the  `examples/lightgbm_tuner_cv.py`.

### DIFF
--- a/optuna/integration/_lightgbm_tuner/optimize.py
+++ b/optuna/integration/_lightgbm_tuner/optimize.py
@@ -228,6 +228,13 @@ class _OptunaObjective(_BaseTuner):
             param_value = int(trial.suggest_uniform("min_child_samples", 5, 100 + _EPS))
             self.lgbm_params["min_child_samples"] = param_value
 
+    def _copy_valid_sets(self, valid_sets: "VALID_SET_TYPE") -> "VALID_SET_TYPE":
+        if isinstance(valid_sets, list):
+            return [copy.copy(d) for d in valid_sets]
+        if isinstance(valid_sets, tuple):
+            return tuple([copy.copy(d) for d in valid_sets])
+        return copy.copy(valid_sets)
+
     def __call__(self, trial: optuna.trial.Trial) -> float:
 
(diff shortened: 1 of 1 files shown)


toshihikoyanase:
I made this PR review-ready. Please take a look.

@Y-oHr-N I think this trick was also applied to your original OptGBM and may be related to #1018. Please feel to add comments.

Y-oHr-N:
I have faced a similar problem during the development of OptGBM.
This process is necessary for the tuning to work correctly.

toshihikoyanase:
@Y-oHr-N Thank you for your comment. Please feel free to add comments on implementation if you have any feedback.

Y-oHr-N:
There is no feedback. LGTM.

toshihikoyanase:
Thank you for your review comment!

HideakiImamura (approved):
@toshihikoyanase Thanks for the PR and @Y-oHr-N Thanks for the review comment!

The change looks good to me. In my understanding, `lgb.train` may change the `train_set` and `valid_sets` in kwargs, but `lgb.cv` does not.

toshihikoyanase:
@HideakiImamura 
Thank you for your review.

> `lgb.train` may change the `train_set` and `valid_sets` in `kwargs`, but `lgb.cv` does not.

In terms of `valid_sets`, `lgb.cv` does not have it as can be seen in [the reference](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.cv.html#lightgbm.cv). On the other hand, the parameter of `train_set` may be changed by `lgb.cv`, and `copy(train_set)` is added right before `lgb.cv`.

HideakiImamura:
Let me merge this PR since I and @Y-oHr-N approved.

PR #1530, opened 2020-07-18
Introduces Multiobjective TPE to optuna.multi_objective.samplers

y0z:
## Motivation
This pull request introduces a multiobjective TPE (MOTPE) sampler proposed in [the paper](https://dl.acm.org/doi/abs/10.1145/3377930.3389817) to optuna.multi_objective.samplers.
Although this PR is work in progress, I want to discuss its API, a proper design of the unit tests, and so on.

Demo ([demo script](https://gist.github.com/y0z/c328709c48e8317989c828d366b30179)):
- Cyan points are initial samples with RandomSampler
- Black points are samples with MOTPE
- Red points are the obtained Pareto front approximation.
demo_small

## Description of the changes
- Add `MOTPEMultiObjectiveSampler` to `optuna.multi_objective.samplers`
- Add unit tests for MOTPE
- Add docs for MOTPE

## TODO
- [x] Add unit tests
- [x] Add docs
- [x] Change hypervolume calculation function
- [x] Benchmarks (comparisons among multiobjective samplers in Optuna)

PR #1812, opened 2020-09-12
Fix WFG._compute()

y0z:
This PR fixes an unintended side effect that _compute() overwrites the original user-defined reference_point.

Reference: https://github.com/optuna/optuna/pull/1530#discussion_r487354710


HideakiImamura (approved):
Thanks for the good catch! LGTM!

sile (approved):
LGTM

Issue #1067, opened 2020-03-30
Drop Python 3.5 support


hvy:
Thread to discussion dropping Python 3.5 support.

See also [Drop Python 2 support. #713]( https://github.com/optuna/optuna/issues/713)

hvy:
`scipy` which is an install requirement is dropping support for Python 3.5 in its next minor version upgrade 1.5.0.

hvy:
Memo: `llvmlite` which is required by `numba` which is required by e.g.`dask_ml` in our examples has stopped providing wheels for Python 3.5 starting from 0.32.

hvy:
Note that scipy 1.5.0 is already released, dopping 3.5. https://github.com/scipy/scipy/releases/tag/v1.5.0
PyTorch also dropped support for 3.5 in its recent 1.6 https://github.com/pytorch/pytorch/releases

c-bata:
cmaes will drop Python 3.5 support from next minor release.

hvy:
C.f. https://github.com/optuna/optuna/issues/711#issuecomment-683521987, when rewriting type hints.

harupy:
FYI: https://docs.python.org/3/whatsnew/3.6.html

hvy:
From our CI.
`DEPRECATION: Python 3.5 reached the end of its life on September 13th, 2020. Please upgrade your Python as Python 3.5 is no longer maintained. pip 21.0 will drop support for Python 3.5 in January 2021. pip 21.0 will remove support for this functionality.`

We should probably update the CI matrix and docs accordingly.

hvy:
Python 3.5 dropped from CI in https://github.com/optuna/optuna/pull/1810

hvy:
I think we can get rid of [`type_checking.py`](https://github.com/optuna/optuna/blob/master/optuna/type_checking.py) now.

c-bata:
@hvy I open #2032 to deprecate `optuna.type_checking` module. Can I unpin this issue?

hvy:
Thanks for the PR, I was considering creating it myself. I believe we can unpin it, but it's probably simpler to keep it pinned until we actually close it. It's not like we're having too many pinned tickets. Maybe I'm missing your point.

c-bata:
> I believe we can unpin it, but it's probably simpler to keep it pinned until we actually close it. It's not like we're having too many pinned tickets.

OK, I see 👍

PR #1818, opened 2020-09-12
Update setup.py to drop python 3.5 support

harupy:
## Motivation

Update setup.py to drop python 3.5 support (#1067)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -141,7 +141,8 @@ def get_extras_require() -> Dict[str, List[str]]:
             ["torch==1.6.0", "torchvision==0.7.0"]
             if sys.platform == "darwin"
             else ["torch==1.6.0+cpu", "torchvision==0.7.0+cpu"]
-        ),
+        )
+        + (["allennlp==1.0.0", "fastai<2"] if sys.version_info[:2] < (3, 8) else []),
         "tests": ["fakeredis", "pytest"],
         "optional": [
             "bokeh<2.0.0",  # optuna/cli.py, optuna/dashboard.py.
--- a/setup.py
+++ b/setup.py
@@ -210,7 +210,7 @@ setup(
             "py.typed",
         ]
     },
-    python_requires=">=3.5",
+    python_requires=">=3.6",
     install_requires=get_install_requires(),
     tests_require=get_tests_require(),
     extras_require=get_extras_require(),
@@ -232,7 +232,6 @@ setup(
         "Intended Audience :: Developers",
         "License :: OSI Approved :: MIT License",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
--- a/setup.py
+++ b/setup.py
@@ -104,8 +104,6 @@ def get_extras_require() -> Dict[str, List[str]]:
                 "fastai<2",
                 "dask[dataframe]",
                 "dask-ml",
-                "allennlp==1.0.0",
-                "fastai<2",
             ]
             if sys.version_info[:2] < (3, 8)
             else []
--- a/setup.py
+++ b/setup.py
@@ -87,19 +87,30 @@ def get_extras_require() -> Dict[str, List[str]]:
             "keras",
             "tensorflow>=2.0.0",
             "tensorflow-datasets",
+            "pytorch-ignite",
+            "pytorch-lightning>=0.8.1",
+            "thop",
+            "skorch",
+            "stable-baselines3>=0.7.0",
+            "catalyst",
         ]
         + (
-            (
-                ["torch==1.6.0", "torchvision==0.7.0"]
-                if sys.platform == "darwin"
-                else ["torch==1.6.0+cpu", "torchvision==0.7.0+cpu"]
-            )
-            + ["pytorch-ignite", "pytorch-lightning>=0.8.1", "thop", "skorch"]
+            ["torch==1.6.0", "torchvision==0.7.0"]
+            if sys.platform == "darwin"
+            else ["torch==1.6.0+cpu", "torchvision==0.7.0+cpu"]
         )
-        + ["stable-baselines3>=0.7.0"]
-        + (["allennlp==1.0.0", "fastai<2"] if sys.version_info[:2] < (3, 8) else [])
-        + (["dask[dataframe]", "dask-ml"] if sys.version_info[:2] < (3, 8) else [])
-        + ["catalyst"],
+        + (
+            [
+                "allennlp==1.0.0",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -172,11 +164,9 @@ def get_extras_require() -> Dict[str, List[str]]:
                 else ["torch==1.6.0+cpu", "torchvision==0.7.0+cpu"]
             )
             + ["pytorch-ignite", "pytorch-lightning>=0.8.1", "skorch"]
-            if (3, 5) < sys.version_info[:2]
-            else []
         )
-        + (["allennlp==1.0.0", "fastai<2"] if (3, 5) < sys.version_info[:2] < (3, 8) else [])
-        + (["catalyst"] if (3, 5) < sys.version_info[:2] else []),
+        + (["allennlp==1.0.0", "fastai<2"] if sys.version_info[:2] < (3, 8) else [])
+        + (["catalyst"] if sys.version_info[:2] else []),
```
c-bata, line 169: ditto.

setup.py
```diff
@@ -95,17 +95,11 @@ def get_extras_require() -> Dict[str, List[str]]:
...
-        + (["allennlp==1.0.0", "fastai<2"] if (3, 5) < sys.version_info[:2] < (3, 8) else [])
-        + (
-            ["llvmlite<=0.31.0", "fsspec<0.8.0"] if (3, 5) == sys.version_info[:2] else []
-        )  # Newer `llvmlite` is not distributed with wheels for Python 3.5.
-        # Newer `fsspec` uses f-strings, which is not compatible with Python 3.5.
+        + ["stable-baselines3>=0.7.0"]
+        + (["allennlp==1.0.0", "fastai<2"] if sys.version_info[:2] < (3, 8) else [])
         + (["dask[dataframe]", "dask-ml"] if sys.version_info[:2] < (3, 8) else [])
-        + (["catalyst"] if (3, 5) < sys.version_info[:2] else []),
+        + ["catalyst"],
```
c-bata, line 102: You can list `"stable-baselines3>=0.7.0"` and `"catalyst"` directly in L76-L90.

setup.py
```diff
@@ -135,11 +129,9 @@ def get_extras_require() -> Dict[str, List[str]]:
                 else ["torch==1.6.0+cpu", "torchvision==0.7.0+cpu"]
             )
             + ["pytorch-ignite", "pytorch-lightning>=0.8.1", "skorch"]
-            if (3, 5) < sys.version_info[:2]
-            else []
         )
-        + (["allennlp==1.0.0", "fastai<2"] if (3, 5) < sys.version_info[:2] < (3, 8) else [])
-        + (["catalyst"] if (3, 5) < sys.version_info[:2] else []),
+        + (["allennlp==1.0.0", "fastai<2"] if sys.version_info[:2] < (3, 8) else [])
+        + (["catalyst"]),
```
c-bata, line 134: ditto.

c-bata (changes requested):

hvy:
How about updating `python_requires` and `classifiers` too? I think it makes sense in scoping the PR to "update setup.py".

harupy:
@hvy @c-bata Thanks for the comments. I have addressed them. Can you take another look?

setup.py
```diff
@@ -87,25 +87,30 @@ def get_extras_require() -> Dict[str, List[str]]:
...
-        )  # Newer `llvmlite` is not distributed with wheels for Python 3.5.
-        # Newer `fsspec` uses f-strings, which is not compatible with Python 3.5.
-        + (["dask[dataframe]", "dask-ml"] if sys.version_info[:2] < (3, 8) else [])
-        + (["catalyst"] if (3, 5) < sys.version_info[:2] else []),
+            [
+                "allennlp==1.0.0",
+                "fastai<2",
+                "dask[dataframe]",
+                "dask-ml",
+                "allennlp==1.0.0",
```
harupy, line 108: Found duplicates. I'll remove them.

setup.py
```diff
@@ -87,25 +87,30 @@ def get_extras_require() -> Dict[str, List[str]]:
...
+                "fastai<2",
```
harupy, line 109: Found duplicates. I'll remove them.

setup.py
```diff
@@ -164,19 +167,17 @@ def get_extras_require() -> Dict[str, List[str]]:
...
+            ["torch==1.6.0", "torchvision==0.7.0"]
+            if sys.platform == "darwin"
+            else ["torch==1.6.0+cpu", "torchvision==0.7.0+cpu"]
         )
-        + (["allennlp==1.0.0", "fastai<2"] if (3, 5) < sys.version_info[:2] < (3, 8) else [])
-        + (["catalyst"] if (3, 5) < sys.version_info[:2] else []),
+        + (["allennlp==1.0.0", "fastai<2"] if sys.version_info[:2] < (3, 8) else []),
     }
 
     return requirements
```
harupy, line 183: Inserting an assertion like this can prevent duplicating packages.

```python
def contains_duplicates(arr):
    return len(arr) == len(set(arr))

for packages in requirements.values():
    assert contains_duplicates(map(get_package_name, packages))

return requirements
```

setup.py
```diff
@@ -127,19 +132,17 @@ def get_extras_require() -> Dict[str, List[str]]:
...
-                ["torch==1.6.0", "torchvision==0.7.0"]
-                if sys.platform == "darwin"
-                else ["torch==1.6.0+cpu", "torchvision==0.7.0+cpu"]
-            )
-            + ["pytorch-ignite", "pytorch-lightning>=0.8.1", "skorch"]
-            if (3, 5) < sys.version_info[:2]
-            else []
+            ["torch==1.6.0", "torchvision==0.7.0"]
+            if sys.platform == "darwin"
+            else ["torch==1.6.0+cpu", "torchvision==0.7.0+cpu"]
```
harupy, line 143: A function below can DRY the code a bit (should be out of scope of this PR though).

```python
def get_torch():
    return (
        ["torch==1.6.0", "torchvision==0.7.0"]
        if sys.platform == "darwin"
        else ["torch==1.6.0+cpu", "torchvision==0.7.0+cpu"]
    )

```

setup.py
harupy: done

c-bata (approved):

hvy (approved):
Thanks a lot, LGTM! 

This simplifies the dependency definitions that has become somewhat of a mess. It's also important that we have the updated `python_requires` included in v2.2.

setup.py
hvy: Sure, that might reduce maintenance cost.

setup.py
hvy: That's one way to ensure it. Not a strong opinion, but I think we can still manage these duplicates in review and that it sounds a bit like over-engineering. Please correct me.

Issue #1803, opened 2020-09-09
plot_contour


z7ye:
## Environment

- Optuna version: 2.0.0
- Python version:3.6.9
- OS:LINUX
- (Optional) Other libraries and their versions:

## Error messages, stack traces, or logs
in version 2.0.0 the code for contour plot use "max_value - min_value", in the case of categorical features, this will give an error. 
```
study = <optuna.study.Study object at 0x7f7af7a4e0b8>, params = None

      def _get_contour_plot(study: Study, params: Optional[List[str]] = None) -> "go.Figure":

          layout = go.Layout(title="Contour Plot")
          trials = [trial for trial in study.trials if trial.state == TrialState.COMPLETE]
          if len(trials) == 0:
              _logger.warning("Your study does not have any completed trials.")
              return go.Figure(data=[], layout=layout)
          all_params = {p_name for t in trials for p_name in t.params.keys()}
          if params is None:
              sorted_params = sorted(list(all_params))
          elif len(params) <= 1:
              _logger.warning("The length of params must be greater than 1.")
              return go.Figure(data=[], layout=layout)
          else:
 >               padding = (max_value - min_value) * padding_ratio
  E               TypeError: unsupported operand type(s) for -: 'str' and 'str'
```
In version 1.9.0, 
```
param_values_range = dict()
    for p_name in sorted_params:
        values = [t.params[p_name] for t in trials if p_name in t.params]
        param_values_range[p_name] = (min(values), max(values))
```

ytknzw:
I think this can be fixed with any numerical conversion like in ``_plot_parallel_coordinate.py``:
```
        try:
            tuple(map(float, values))
        except (TypeError, ValueError):
            vocab = defaultdict(lambda: len(vocab))  # type: DefaultDict[str, int]
            values = [vocab[v] for v in values]
```

sile:
Fixed by #1819

PR #1819, opened 2020-09-12
Fix contour plot error for categorical distributions

zchenry:
## Motivation
As mentioned in https://github.com/optuna/optuna/issues/1803, error occurs when a categorical distribution of non-numerical values are used.

## Description of the changes
This pull request fixed the above error. 

For a categorical distribution, plotly behaves differently depending on the values.

- For values such as `[""sigmoid", "relu", "leakyrelu""]` which cannot be converted to numbers, plotly will automatically plot it as the category type. In this case, updating the axes type will cause undesirable plots (shown below).
- For values such as `["10", "11", "100"]` which can be converted to numbers, plotly will plot it as normal numbers. In this case, we need to update the axes type.

Therefore, I introduce a dictinary `update_category_axes` to store the above information.

Sample code and output:
```python
def objective(trial):
    x = trial.suggest_categorical("x", ["10", "11", "100"])
    y = trial.suggest_categorical("y", ["sigmoid", "relu", "leakyrelu"])
    z = trial.suggest_loguniform("z", 1e-1, 1e2)
    return int("0" in x) + abs(len(y) - 5) + (z/50) ** 2

study = optuna.create_study()

optuna.visualization.plot_contour(study).show()
```

good

For parameter `"y"`, if the axes type is manually updated, the plot becomes as follows.

bad

### DIFF
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -10,6 +10,7 @@ from optuna.study import StudyDirection
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
 from optuna.visualization._plotly_imports import _imports
+from optuna.visualization._utils import _is_categorical
 from optuna.visualization._utils import _is_log_scale
 
 if _imports.is_successful():
@@ -90,14 +91,21 @@ def _get_contour_plot(study: Study, params: Optional[List[str]] = None) -> "go.F
 
     padding_ratio = 0.05
     param_values_range = {}
+    update_category_axes = {}
     for p_name in sorted_params:
         values = [t.params[p_name] for t in trials if p_name in t.params]
-        max_value = max(values)
+
         min_value = min(values)
+        max_value = max(values)
+
         if _is_log_scale(trials, p_name):
             padding = (math.log10(max_value) - math.log10(min_value)) * padding_ratio
             min_value = math.pow(10, math.log10(min_value) - padding)
             max_value = math.pow(10, math.log10(max_value) + padding)
+
+        elif _is_categorical(trials, p_name):
+            update_category_axes[p_name] = any([str(v).isnumeric() for v in set(values)])
+
         else:
             padding = (max_value - min_value) * padding_ratio
             min_value = min_value - padding
@@ -113,6 +121,12 @@ def _get_contour_plot(study: Study, params: Optional[List[str]] = None) -> "go.F
         figure = go.Figure(data=sub_plots, layout=layout)
         figure.update_xaxes(title_text=x_param, range=param_values_range[x_param])
         figure.update_yaxes(title_text=y_param, range=param_values_range[y_param])
+
+        if _is_categorical(trials, x_param) and update_category_axes[x_param]:
+            figure.update_xaxes(type="category")
+        if _is_categorical(trials, y_param) and update_category_axes[y_param]:
+            figure.update_yaxes(type="category")
+
         if _is_log_scale(trials, x_param):
             log_range = [math.log10(p) for p in param_values_range[x_param]]
             figure.update_xaxes(range=log_range, type="log")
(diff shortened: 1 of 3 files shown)


optuna/visualization/_contour.py
```diff
@@ -113,6 +121,12 @@ def _get_contour_plot(study: Study, params: Optional[List[str]] = None) -> "go.F
         figure = go.Figure(data=sub_plots, layout=layout)
         figure.update_xaxes(title_text=x_param, range=param_values_range[x_param])
         figure.update_yaxes(title_text=y_param, range=param_values_range[y_param])
+
+        if _is_categorical(trials, x_param) and update_category_axes[x_param]:
```
sile, line 125: I think that the following line can be simplified (and optimized) as follows:
```suggestion
        if update_category_axes.get(x_param, False):
```

optuna/visualization/_contour.py
```diff
@@ -113,6 +121,12 @@ def _get_contour_plot(study: Study, params: Optional[List[str]] = None) -> "go.F
...
+            figure.update_xaxes(type="category")
+        if _is_categorical(trials, y_param) and update_category_axes[y_param]:
```
sile, line 127: Ditto
```suggestion
        if update_category_axes.get(y_param, False):
```

optuna/visualization/_contour.py
```diff
@@ -90,14 +91,21 @@ def _get_contour_plot(study: Study, params: Optional[List[str]] = None) -> "go.F
...
         min_value = min(values)
+        max_value = max(values)
+
         if _is_log_scale(trials, p_name):
             padding = (math.log10(max_value) - math.log10(min_value)) * padding_ratio
             min_value = math.pow(10, math.log10(min_value) - padding)
             max_value = math.pow(10, math.log10(max_value) + padding)
+
+        elif _is_categorical(trials, p_name):
+            update_category_axes[p_name] = any([str(v).isnumeric() for v in set(values)])
```
sile, line 107: For future readers, how about leaving a brief comment about why `update_category_axes` is needed if a categorical parameter contains numeric like values?

optuna/visualization/_contour.py
```diff
@@ -140,14 +154,22 @@ def _get_contour_plot(study: Study, params: Optional[List[str]] = None) -> "go.F
                         showscale = False
                     figure.add_trace(contour, row=y_i + 1, col=x_i + 1)
                     figure.add_trace(scatter, row=y_i + 1, col=x_i + 1)
+
                 figure.update_xaxes(range=param_values_range[x_param], row=y_i + 1, col=x_i + 1)
                 figure.update_yaxes(range=param_values_range[y_param], row=y_i + 1, col=x_i + 1)
+
+                if _is_categorical(trials, x_param) and update_category_axes[x_param]:
```
sile, line 161: Ditto
```suggestion
                if update_category_axes.get(x_param, False):
```

optuna/visualization/_contour.py
```diff
@@ -140,14 +154,22 @@ def _get_contour_plot(study: Study, params: Optional[List[str]] = None) -> "go.F
...
+                    figure.update_xaxes(type="category", row=y_i + 1, col=x_i + 1)
+                if _is_categorical(trials, y_param) and update_category_axes[y_param]:
```
sile, line 163: Ditto
```suggestion
                if update_category_axes.get(y_param, False):
```

sile (commented):
Thank you for your PR!
This RP seems nice but I left some review comments so please take a look.

zchenry:
@sile Thank you for your comments. I simplifed the code and added a comment. I was looking at `plotly.io` source code but didn't find any function I can use in this case.

HideakiImamura (approved):
Thanks for the PR! LGTM!

sile (approved):
LGTM!

Issue #1800, opened 2020-09-09
Hide `__init__`


crcrpar:
Since v2.0.0, the documentation displays dunder init though the explanation itself is not helpful at all rather it looks generated by Sphinx or whatever.

So I think hiding them would beautify the documentation.

![image](https://user-images.githubusercontent.com/16191443/92548133-31542800-f291-11ea-8930-84aca9e7a2f0.png)
(Capture from https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html#optuna.study.Study)

PR #1820, opened 2020-09-12
Hide __init__ from docs

upura:
## Motivation

Please refer to the Issue #1800.

## Description of the changes

Add `autosummary/class.rst` to apply a template. You can see the updated version [here](https://58105-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/generated/optuna.study.Study.html#optuna.study.Study).

Screen Shot 2020-09-13 at 16 37 32

### DIFF
--- a/docs/source/_templates/autosummary/class.rst
+++ b/docs/source/_templates/autosummary/class.rst
@@ -1,27 +1,14 @@
 {% extends "!autosummary/class.rst" %}
 
 {% block methods %}
-{% if methods %}
-   .. HACK -- the point here is that we don't want this to appear in the output, but the autosummary should still generate the pages.
-      .. autosummary::
-         :toctree:
-      {% for item in all_methods %}
-         {%- if not item.startswith('_') or item in ['__call__'] %}
-         {{ name }}.{{ item }}
-         {%- endif -%}
-      {%- endfor %}
-{% endif %}
-{% endblock %}
+   {% set methods = methods | select("ne", "__init__") | list %}
+   {% if methods %}
+   .. rubric:: Methods
 
-{% block attributes %}
-{% if attributes %}
-   .. HACK -- the point here is that we don't want this to appear in the output, but the autosummary should still generate the pages.
-      .. autosummary::
-         :toctree:
-      {% for item in all_attributes %}
-         {%- if not item.startswith('_') %}
-         {{ name }}.{{ item }}
-         {%- endif -%}
-      {%- endfor %}
-{% endif %}
+   .. autosummary::
+   {% for item in methods %}
+      ~{{ name }}.{{ item }}
+   {%- endfor %}
+   {% endif %}
+
 {% endblock %}
--- a/docs/source/_templates/autosummary/class.rst
+++ b/docs/source/_templates/autosummary/class.rst
@@ -1,27 +1,27 @@
-{{ fullname }}
-{{ underline }}
+{% extends "!autosummary/class.rst" %}
 
-.. currentmodule:: {{ module }}
-
-.. autoclass:: {{ objname }}
-
-   ..
-      Ordinary methods
-
-{% for item in methods %}
-{% if item not in ('__init__',) %}
-   .. automethod:: {{ item }}
+{% block methods %}
+{% if methods %}
+   .. HACK -- the point here is that we don't want this to appear in the output, but the autosummary should still generate the pages.
+      .. autosummary::
+         :toctree:
+      {% for item in all_methods %}
+         {%- if not item.startswith('_') or item in ['__call__'] %}
+         {{ name }}.{{ item }}
+         {%- endif -%}
+      {%- endfor %}
 {% endif %}
-{%- endfor %}
+{% endblock %}
 
-   ..
-      Attributes
-
-{% block attributes %} {% if attributes %}
-
-   .. rubric:: Attributes
-
-{% for item in attributes %}
-   .. autoattribute:: {{ item }}
-{%- endfor %}
-{% endif %} {% endblock %}
+{% block attributes %}
+{% if attributes %}
+   .. HACK -- the point here is that we don't want this to appear in the output, but the autosummary should still generate the pages.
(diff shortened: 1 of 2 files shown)


docs/source/_templates/autosummary/class.rst
```diff
@@ -0,0 +1,27 @@
...
+   .. HACK -- the point here is that we don't want this to appear in the output, but the autosummary should still generate the pages.
+      .. autosummary::
+         :toctree:
+      {% for item in all_attributes %}
+         {%- if not item.startswith('_') %}
+         {{ name }}.{{ item }}
+         {%- endif -%}
+      {%- endfor %}
+{% endif %}
+{% endblock %}
```
harupy, line 27: workaround:

```
{% extends "!autosummary/class.rst" %}

{% block methods %}
   {% set methods = methods | select("ne", "__init__") | list %}
   {% if methods %}
   .. rubric:: Methods

   .. autosummary::
   {% for item in methods %}
      ~{{ name }}.{{ item }}
   {%- endfor %}
   {% endif %}

{% endblock %}
```

harupy:
```
{% set methods = methods | select("ne", "__init__") | list %}
```

- `select("ne", "__init__")` removes `__init__` in `methods`.
-  if we didn't put `| list`, `methods` would become a generator (which is truthy)

docs/source/conf.py
```diff
@@ -56,7 +56,7 @@
 ]
 
 # Add any paths that contain templates here, relative to this directory.
-templates_path = ['_templates']
+templates_path = ['_templates', '_templates/autosummary']
```
harupy, line 59: `templates_path = ['_templates']` works fine in my environment.

docs/source/_templates/autosummary/class.rst
upura: Thank you for your comment. I'm afraid that I'm a little confused as to what should I do. I'd appreciate it if you can directory edit this branch.

harupy: You can just replace the current template with my template :)

harupy: Can you apply my suggestion following the instruction here: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes?

upura:
@harupy Thanks for your help! All tests have passed and the artifacts have changed as expected.

docs/source/_templates/autosummary/class.rst
```diff
@@ -0,0 +1,14 @@
+{% extends "!autosummary/class.rst" %}
+
```
harupy, line 2:
```suggestion

{# An autosummary template to exclude the __init__ method for a class #}

```

How about inserting a comment to clarify what this template does?

upura: Sounds great!

crcrpar (commented):
Thank you @upura for your great effort and @harupy for your help. Much appreciated.

crcrpar (approved):
Really sorry about this: could you update the branch again?

upura:
@crcrpar Successfully updated.

HideakiImamura (approved):
Thank you for the great PR to improve the appearance of documents! LGTM!

docs/source/_templates/autosummary/class.rst
```diff
@@ -0,0 +1,19 @@
...
+{#
+An autosummary template to exclude the class constructor (__init__)
+which doesn't contain any docstring in Optuna.
+#}
+
+{% block methods %}
+   {% set methods = methods | select("ne", "__init__") | list %}
+   {% if methods %}
+   .. rubric:: Methods
```
toshihikoyanase, line 11: Can I ask a question? 
I'm a bit curious if we don't need to use `{{ _('Methods') }}` as can be seen in the original template as follows:

```rst
.. autoclass:: {{ objname }}

   {% block methods %}
   .. automethod:: __init__

   {% if methods %}
   .. rubric:: {{ _('Methods') }}  {# Here #}

   .. autosummary::
   {% for item in methods %}
      ~{{ name }}.{{ item }}
   {%- endfor %}
   {% endif %}
   {% endblock %}
```

(c.f., https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autosummary/templates/autosummary/class.rst#L11)

toshihikoyanase: Ah, I found that this notation was related to translation. `_()` is equivalent to `gettext()`, which translates a single string. We may integrate the [Chinese docs](https://zh-cn.optuna.org/) into the English docs somehow, so could you use `gettext` here?
https://jinja.palletsprojects.com/en/2.11.x/templates/#i18n

toshihikoyanase: Currently, we do not use the `i18n` feature, so I think this modification is optional.
Please ignore my comment, if you have any troubles to introduce `gettext()`.

upura: I just replaced `Methods` to `{{ gettext('Methods') }}`, and encountered the following error.
```
    getattr(obj, "unsafe_callable", False) or getattr(obj, "alters_data", False)
jinja2.exceptions.UndefinedError: 'gettext' is undefined
```

Sphinx uses the [Jinja templating engine](https://www.sphinx-doc.org/en/master/templating.html) and we have to introduce an extension via Sphinx in some way. And unfortunately, I'm afraid that I can't find a way to do this.

How about making this modification an another issue, and thinking about it when we work on translations?

toshihikoyanase: @upura Thank you for your investigation. That makes sense. I'll work on it when we introduce multi-lingual documentation.

toshihikoyanase (approved):

toshihikoyanase:
Let me slightly modify the title for the release note.

PR #1821, opened 2020-09-12
Remove TODO

sfujiwara:
## Motivation

- Remove resolved TODO

## Description of the changes

The TODO removed in this PR is written by me when I contributed to Optuna before.
The problem is that TensorFlow's logger print duplicated log message.

As far as I investigated, It's not Optuna's problem but TensorFlow's one:
https://github.com/tensorflow/tensorflow/issues/8023

### DIFF
--- a/examples/pruning/tfkeras_integration.py
+++ b/examples/pruning/tfkeras_integration.py
@@ -103,10 +103,6 @@ def objective(trial):
         callbacks=callbacks,
     )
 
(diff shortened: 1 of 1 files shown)


crcrpar (approved):
Thank you for your fix and detailed explanation

toshihikoyanase (approved):
I confirmed that the duplicated log was resolved when I executed `examples/tfkeras_simple.py`. LGTM!

Issue #1806, opened 2020-09-11
Increase coverage of metric alias of `LightGBMTuner`


toshihikoyanase:
In #1602, the alias of `l2` has been supported by `LightGBMTuner`. However, the original `lightgbm` has more metric aliases as can be seen in [ParseMetricAlias](https://github.com/microsoft/LightGBM/blob/083b02af73b8fcf9a92af78612c6cca6243df752/include/LightGBM/config.h#L1111-L1112), and some of them are not supported by `LightGBMTuner`.

Please add the aliases of the following metrics to [`_ALIAS_METRIC_LIST`](https://github.com/optuna/optuna/blob/master/optuna/integration/_lightgbm_tuner/alias.py#L43):

- [x] `l2`
- [x] `rmse`
- [x] `l1`
- [x] `binary_logloss`
- [x] `ndcg`
- [x] `map`
- [x] `multi_logloss`
- [x] `cross_entropy`
- [x] `cross_entropy_lambda`
- [x] `kullback_leibler`
- [x] `mape`
- [x] `auc_mu`
- [x] `custom`

Please note that `custom` needs an additional fix as mentioned in https://github.com/optuna/optuna/issues/1351.

toshihikoyanase:
Let me close this issue since #1807 and #1822 resolved it.
Thank you for your contribution, @upura and @nyanhi !

PR #1822, opened 2020-09-12
Add metric alias for LightGBMTuner

nyanhi:
## Motivation
To address #1806 

## Description of the changes
Added the aliases of the remaining metrics to [_ALIAS_METRIC_LIST](https://github.com/optuna/optuna/blob/master/optuna/integration/_lightgbm_tuner/alias.py#L43)

### DIFF
--- a/optuna/integration/_lightgbm_tuner/alias.py
+++ b/optuna/integration/_lightgbm_tuner/alias.py
@@ -41,6 +41,7 @@ def _handling_alias_parameters(lgbm_params: Dict[str, Any]) -> None:
 
 
 _ALIAS_METRIC_LIST = [
+    # The list `alias_names` do not include the `metric_name` itself.
     {
         "metric_name": "ndcg",
         "alias_names": [
@@ -55,20 +56,19 @@ _ALIAS_METRIC_LIST = [
     {"metric_name": "map", "alias_names": ["mean_average_precision"]},
     {
         "metric_name": "l2",
-        "alias_names": ["regression", "regression_l2", "l2", "mean_squared_error", "mse"],
+        "alias_names": ["regression", "regression_l2", "mean_squared_error", "mse"],
     },
     {
         "metric_name": "l1",
-        "alias_names": ["regression_l1", "l1", "mean_absolute_error", "mae"],
+        "alias_names": ["regression_l1", "mean_absolute_error", "mae"],
     },
     {
         "metric_name": "binary_logloss",
-        "alias_names": ["binary_logloss", "binary"],
+        "alias_names": ["binary"],
     },
     {
         "metric_name": "multi_logloss",
         "alias_names": [
-            "multi_logloss",
             "multiclass",
             "softmax",
             "multiclassova",
@@ -79,27 +79,23 @@ _ALIAS_METRIC_LIST = [
     },
     {
         "metric_name": "cross_entropy",
-        "alias_names": ["cross_entropy", "xentropy"],
+        "alias_names": ["xentropy"],
     },
     {
         "metric_name": "cross_entropy_lambda",
-        "alias_names": ["cross_entropy_lambda", "xentlambda"],
+        "alias_names": ["xentlambda"],
     },
     {
         "metric_name": "kullback_leibler",
-        "alias_names": ["kullback_leibler", "kldiv"],
+        "alias_names": ["kldiv"],
     },
     {
         "metric_name": "mape",
-        "alias_names": ["mape", "mean_absolute_percentage_error"],
-    },
-    {
-        "metric_name": "auc_mu",
-        "alias_names": ["auc_mu"],
+        "alias_names": ["mean_absolute_percentage_error"],
     },
     {
         "metric_name": "custom",
-        "alias_names": ["none", "null", "custom", "na"],
+        "alias_names": ["none", "null", "na"],
     },
     {
         "metric_name": "rmse",
--- a/optuna/integration/_lightgbm_tuner/alias.py
+++ b/optuna/integration/_lightgbm_tuner/alias.py
@@ -57,6 +57,50 @@ _ALIAS_METRIC_LIST = [
         "metric_name": "l2",
         "alias_names": ["regression", "regression_l2", "l2", "mean_squared_error", "mse"],
     },
+    {
+        "metric_name": "l1",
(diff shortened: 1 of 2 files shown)


tests/integration_tests/lightgbm_tuner_tests/test_alias.py
```diff
@@ -70,6 +70,64 @@ def test_handling_alias_metrics() -> None:
     _handling_alias_metrics(lgbm_params)
     assert lgbm_params["metric"] == "auc"
 
+
+    lgbm_params = {"metric": "rmse"}
+    _handling_alias_metrics(lgbm_params)
+    assert lgbm_params["metric"] == "rmse"
```
toshihikoyanase, line 76: This test case seems to be duplicated.

tests/integration_tests/lightgbm_tuner_tests/test_alias.py
```diff
@@ -70,6 +70,64 @@ def test_handling_alias_metrics() -> None:
...
+
+    for alias in ["auc_mu"]:
+        lgbm_params = {"metric": alias}
+        _handling_alias_metrics(lgbm_params)
+        assert lgbm_params["metric"] == "auc_mu"
+
+    for alias in ["none", "null", "custom", "na"]:
+        lgbm_params = {"metric": alias}
+        _handling_alias_metrics(lgbm_params)
+        assert lgbm_params["metric"] == "custom"
```
toshihikoyanase, line 129: I found that the test logic for each alias is almost the same. I think we can simplify the tests using `pytest.mark.parametrize`. What do you think of it?

toshihikoyanase (changes requested):
Thank you for your PR. It basically looks good to me. I added two comments to improve the unit tests.
In addition, please apply `black` to fix the code format.

optuna/integration/_lightgbm_tuner/alias.py
```diff
@@ -57,6 +57,50 @@ def _handling_alias_parameters(lgbm_params: Dict[str, Any]) -> None:
...
+        "metric_name": "kullback_leibler",
+        "alias_names": ["kullback_leibler", "kldiv"],
+    },
+    {
+        "metric_name": "mape",
+        "alias_names": ["mape", "mean_absolute_percentage_error"],
+    },
+    {
+        "metric_name": "auc_mu",
+        "alias_names": ["auc_mu"],
```
upura, line 98: If I understand correctly, the value of `metric_name` itself should not be included in `alias_names`. The same can be said for the other metric such as "l2".

Here are the function that process alias in `LightGBMTuner `.
https://github.com/optuna/optuna/blob/master/optuna/integration/_lightgbm_tuner/alias.py#L67

nyanhi: Thank you for the comment! I agree that the `metric_name` itself can be (but not necessarily) removed from the list of `alias_names`.
For the `metric_name` "auc_mu" where `alias_names` is unique and identical, I felt it's more explicit that `_ALIAS_METRIC_LIST` does cover the metric, hence added those 4 lines (but do not have effect until other alias will be added in future, which may be your point). 
Please kindly let me know your thought.

upura: In my opinion, it's better not to write something that isn't effective at that time. However, I'm sure each organization has its own policies, and it seems best to follow them.

toshihikoyanase: @upura 
> the value of `metric_name` itself do not have to be included in `alias_names`. The same can be said for the other metric such as "l2".

Thank you for pointing it out. The original code such as `ndcg` and `map ` by @smly does not contain `metric_name` in `alias_names`. Maybe I overlooked the inclusion of unnecessary items during the review process. 

@nyanhi 
Is it possible to remove such metric names from `alias_names`? If you do not have time enough, I'll fix it in the follow-up PR.

> "auc_mu"

I think you kindly included this metric because it is included in the list in https://github.com/optuna/optuna/issues/1806. I'm sorry for bothering you, but could you remove it? I feel it is verbose since it does not have any aliases.

tests/integration_tests/lightgbm_tuner_tests/test_alias.py
```diff
@@ -37,40 +41,56 @@ def test_handling_alias_parameter() -> None:
...
+            [
+                "ndcg",
+                "lambdarank",
+                "rank_xendcg",
+                "xendcg",
+                "xe_ndcg",
+                "xe_ndcg_mart",
+                "xendcg_mart",
+            ],
+            "ndcg",
```
toshihikoyanase, line 57: Thank you for adding the expected metric name to the aliases. Now we can check the identical transformation :+1:

optuna/integration/_lightgbm_tuner/alias.py
nyanhi: @toshihikoyanase 
Understood your points and made the changes. I have added below comment to avoid shake in notation going forward.
Thank you for the comment.
```
# The list `alias_names` do not include the `metric_name` itself.
```

tests/integration_tests/lightgbm_tuner_tests/test_alias.py
```diff
@@ -37,40 +41,56 @@ def test_handling_alias_parameter() -> None:
...
+    if len(aliases) > 0:
+        for alias in aliases:
+            lgbm_params = {"metric": alias}
+            _handling_alias_metrics(lgbm_params)
+            assert lgbm_params["metric"] == expect
+    else:
+        lgbm_params = {}
         _handling_alias_metrics(lgbm_params)
-        assert lgbm_params["metric"] == "rmse"
+        print("null case")
```
toshihikoyanase, line 95: A nit, but could you remove this line? The message is short and it looks like code for debugging.

toshihikoyanase (commented):
Thank you for your update. I added a small comment. Please take a look.

tests/integration_tests/lightgbm_tuner_tests/test_alias.py
nyanhi: Yes, thank you for pointing out.

toshihikoyanase (approved):

HideakiImamura (approved):
Thanks for the PR! This change improves the usability. LGTM!

PR #1823, opened 2020-09-12
Implement `plot_slice` and `_get_slice_plot` with Matplotlib backend.

ytknzw:
This PR proposes to implement a ``matplotlib`` backend for plotting parameter importance based on PR #1756 .

## Motivation
See issue #1539 and PR #1707 for the original motivation.

## Description of the changes
Create  ``optuna.visualization.matplotlib._slice.plot_slice``, ``optuna.visualization.matplotlib._slice._get_slice_plot``, ``optuna.visualization.matplotlib._slice._generate_slice_subplot``, ``optuna.visualization.matplotlib._slice._calc_lim_with_padding``.


optuna/visualization/matplotlib/_slice.py
```diff
@@ -1,11 +1,152 @@
...
+    padding_ratio: float,
+    obj_values: List[Union[int, float]],
+) -> "PathCollection":
+    x_values = []
+    y_values = []
+    trial_numbers = []
+    for t in trials:
+        if param in t.params:
+            x_values.append(t.params[param])
+            y_values.append(obj_values[t.number])
```
crcrpar, line 123: If some trials are pruned, `t.number` and the index of `t` in the `trials` would be different.

```suggestion
    for i. t in enumerate(trials):
        if param in t.params:
            x_values.append(t.params[param])
            y_values.append(obj_values[i])
```

crcrpar (commented):
Sorry for having kept you waiting for so long 🙇 
sample
This output of the plotly implementation's example looks cool.

However, this implementation doesn't seem to support `Study`s with some pruned trials. I found this with https://github.com/optuna/optuna/blob/master/examples/visualization/plot_study.ipynb.

optuna/visualization/matplotlib/_slice.py
ytknzw: Thank you for reviewing! I agree with you.

crcrpar (approved):
really appreciate your swift response and great work.

p.s. I resolved the conflicts 🙂

ytknzw:
Thanks!

optuna/visualization/matplotlib/_slice.py
```diff
@@ -1,11 +1,152 @@
...
+    value_max = max(values)
+    value_min = min(values)
+    if is_log_scale:
+        padding = (math.log10(value_max) - math.log10(value_min)) * padding_ratio
+        return (
+            math.pow(10, math.log10(value_min) - padding),
+            math.pow(10, math.log10(value_max) + padding),
+        )
+    else:
+        padding = (value_max - value_min) * padding_ratio
```
himkt, line 151: I think we have to consider categorical parameters here. ☝️ 
ref. https://colab.research.google.com/drive/1jcUmtMTuv9USSHoxCM6deUm3ojAlooCR?usp=sharing

toshihikoyanase: I agree with @himkt. Categorical parameters accept `None, bool, int, float, str`, so we need to take care of them in this function or in `_generate_slice_subplot` (I'm not sure if `ax.scatter` can handle `None` or `str`).

https://github.com/optuna/optuna/blob/965a3f31a7984d83ca1246a3554a77e9cf7370ed/optuna/distributions.py#L12

ytknzw: Thank you for your suggestions, @himkt and @toshihikoyanase!
I'll work on that change this weekend.

ytknzw: @himkt @toshihikoyanase 
Could you review the codes I added for handling categorical parameters?

optuna/visualization/matplotlib/_slice.py
```diff
@@ -1,11 +1,159 @@
...
+    xlim = _calc_lim_with_padding(x_values, padding_ratio, scale)
+    ax.set_xlim(xlim[0], xlim[1])
+    sc = ax.scatter(x_values, y_values, c=trial_numbers, cmap=cmap, edgecolors="grey")
+    ax.label_outer()
+
+    return sc
+
+
+def _calc_lim_with_padding(
+    values: List[Union[int, float]], padding_ratio: float, scale: str = None
```
himkt, line 143:
```suggestion
    values: List[Union[int, float]], padding_ratio: float, scale: Optional[str] = None
```

I think `scale` is `Optional[str]`. 👀

himkt (commented):
Thank you for the update! It basically looks good. 👍 
I dropped the minor comment, please take a look.

optuna/visualization/matplotlib/_slice.py
ytknzw: Yes, thanks!

himkt (approved):
LGTM. Thank you for your great contribution!

PR #1776, opened 2020-09-01
Add share_optimizer option on CmaEsSampler

c-bata:
## Motivation

fix #1775 

## Description of the changes

* Add `share_optimizer` option which enables to store CMA object in system_attrs (default: False).
* Add `randomize_x0` option to sample initial starting points randomly.

###  Which ones should be set as a default argument, `share_optimizer=False` or `share_optimizer=True`?

* Set `share_optimizer=False` from the next release: Existing users may be confused by this breaking change. From the next release, CmaEsSampler cannot resume optimizations.
* Set `share_optimizer=True` until v3.0.0 release: It keeps current behavior. New users may be confused because this option does not work with PostgreSQL and MySQL.

PR #1833, opened 2020-09-15
Store CMAES optimizer after splitting into substrings.

### LINKED ISSUE #1775: Cannot store CMA object on PostgreSQL and MySQL.
## Problem details

This bug is reported by @maartenpants on gitter. Original report is here:

> Hi there, has anyone else experienced this issue with CmaEsSampler and the PostgreSQL storage backend:
> ```
> (psycopg2.errors.StringDataRightTruncation) value too long for type character varying(2048)
> [SQL: INSERT INTO trial_system_attributes (trial_id, key, value_json) VALUES (%(trial_id)s, %(key)s, %(value_json)s) RETURNING trial_system_attributes.trial_system_attribute_id]
> ```
> It seems the serialized version of the CMA class is too large to fit in the 2048 character column. This code is the culprit:
> ```
> optimizer_str = pickle.dumps(optimizer).hex()
> study._storage.set_trial_system_attr(trial._trial_id, "cma:optimizer", optimizer_str)
> ```

https://gitter.im/optuna/optuna?at=5f484dea36e6f709fd08a5fa

I checked the length of serialized CMA object.

```python
import pickle

import numpy as np
from cmaes import CMA


def get_cma_object_size(dim: int) -> int:
    optimizer = CMA(mean=np.zeros(dim), sigma=1.3)
    optimizer_str = pickle.dumps(optimizer).hex()
    return len(optimizer_str)


if __name__ == '__main__':
    for i in range(2, 32, 2):
        print(f"dim={i} size={get_cma_object_size(i)}")
```

```
$ python examples/cma_size.py 
dim=2 size=2516
dim=4 size=2916
dim=6 size=3364
dim=8 size=3850
dim=10 size=4442
dim=12 size=5018
dim=14 size=5738
dim=16 size=6378
dim=18 size=7194
dim=20 size=8074
dim=22 size=8906
dim=24 size=9916
dim=26 size=10956
dim=28 size=12092
dim=30 size=13148
```

This result means that the bug is raised even on low-dimensional search space.

I've ever tested on SQLite3 but never bumped into this bug. The reason why is:

> (9) What is the maximum size of a VARCHAR in SQLite?
> SQLite does not enforce the length of a VARCHAR. You can declare a VARCHAR(10) and SQLite will be happy to store a 500-million character string there. And it will keep all 500-million characters intact. Your content is never truncated. SQLite understands the column type of "VARCHAR(N)" to be the same as "TEXT", regardless of the value of N.
> https://www.sqlite.org/faq.html#:~:text=(9)%20What%20is%20the%20maximum,all%20500%2Dmillion%20characters%20intact.

c-bata:
## Motivation

fix #1775

## Description of the changes

Split an optimizer string, puts these substrings into system_attrs.

### DIFF
--- a/optuna/samplers/_cmaes.py
+++ b/optuna/samplers/_cmaes.py
@@ -314,13 +314,9 @@ class CmaEsSampler(BaseSampler):
 
     def _store_optimizer(self, storage: BaseStorage, trial_id: int, optimizer: CMA) -> None:
         optimizer_str = pickle.dumps(optimizer).hex()
-        optimizer_len = len(optimizer_str)
-
-        for i in range(optimizer_len % _SYSTEM_ATTR_MAX_LENGTH):
-            start = i * _SYSTEM_ATTR_MAX_LENGTH
-            end = min((i + 1) * _SYSTEM_ATTR_MAX_LENGTH, optimizer_len)
-            key = "cma:optimizer:{}".format(i)
-            storage.set_trial_system_attr(trial_id, key, optimizer_str[start:end])
+        optimizer_attrs = _split_optimizer_str(optimizer_str)
+        for key in optimizer_attrs:
+            storage.set_trial_system_attr(trial_id, key, optimizer_attrs[key])
 
     def _restore_optimizer(
         self,
@@ -328,20 +324,18 @@ class CmaEsSampler(BaseSampler):
     ) -> Tuple[Optional[CMA], int]:
         # Restore a previous CMA object.
         for trial in reversed(completed_trials):
-            len_optimizer_keys = len(
-                [key for key in trial.system_attrs if key.startswith("cma:optimizer")]
-            )
-
-            if len_optimizer_keys == 0:
+            optimizer_attrs = {
+                key: value
+                for key, value in trial.system_attrs.items()
+                if key.startswith("cma:optimizer")
+            }
+            if len(optimizer_attrs) == 0:
                 continue
 
-            # "cma:optimizer" is used until v2.2.0.
-            serialized_optimizer = trial.system_attrs.get("cma:optimizer", None)
+            # Check "cma:optimizer" key for backward compatibility.
+            serialized_optimizer = optimizer_attrs.get("cma:optimizer", None)
             if serialized_optimizer is None:
-                serialized_optimizer = ""
-                for i in range(len_optimizer_keys):
-                    key = "cma:optimizer:{}".format(i)
-                    serialized_optimizer += trial.system_attrs[key]
+                serialized_optimizer = _concat_optimizer_str(optimizer_attrs)
 
             n_restarts = trial.system_attrs.get("cma:n_restarts", 0)  # type: int
             return pickle.loads(bytes.fromhex(serialized_optimizer)), n_restarts
@@ -432,6 +426,24 @@ class CmaEsSampler(BaseSampler):
                 copied_t.value = value
                 complete_trials.append(copied_t)
         return complete_trials
+
+
+def _split_optimizer_str(optimizer_str: str) -> Dict[str, str]:
+    optimizer_len = len(optimizer_str)
+    attrs = {}
+    for i in range(math.ceil(optimizer_len / _SYSTEM_ATTR_MAX_LENGTH)):
+        start = i * _SYSTEM_ATTR_MAX_LENGTH
+        end = min((i + 1) * _SYSTEM_ATTR_MAX_LENGTH, optimizer_len)
+        attrs["cma:optimizer:{}".format(i)] = optimizer_str[start:end]
+    return attrs
+
+
+def _concat_optimizer_str(attrs: Dict[str, str]) -> str:
+    serialized_optimizer = ""
+    for i in range(len(attrs)):
+        key = "cma:optimizer:{}".format(i)
+        serialized_optimizer += attrs[key]
+    return serialized_optimizer
 
 
 def _to_cma_param(distribution: BaseDistribution, optuna_param: Any) -> float:
--- a/optuna/samplers/_cmaes.py
+++ b/optuna/samplers/_cmaes.py
@@ -437,11 +437,9 @@ def _split_optimizer_str(optimizer_str: str) -> Dict[str, str]:
 
 
 def _concat_optimizer_attrs(optimizer_attrs: Dict[str, str]) -> str:
-    optimizer_str = ""
-    for i in range(len(optimizer_attrs)):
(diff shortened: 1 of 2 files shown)


optuna/samplers/_cmaes.py
```diff
@@ -16,13 +16,15 @@
...
 from optuna.samplers import BaseSampler
+from optuna.storages import BaseStorage
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
 
 
 _logger = logging.get_logger(__name__)
 
 _EPS = 1e-10
+_SYSTEM_ATTR_MAX_LENGTH = 2048
```
c-bata, line 27: This should be less than 2046 characters because optimizer_str is stored after `json.dumps()`.

```
sqlalchemy.exc.DataError: (MySQLdb._exceptions.DataError) (1406, "Data too long for column 'value_json' at row 1")
[SQL: INSERT INTO trial_system_attributes (trial_id, `key`, value_json) VALUES (%s, %s, %s)]
[parameters: (46, 'cma:optimizer:0', '"80049528050000000000008c0a636d6165732e5f636d61948c03434d419493942981947d94288c065f6e5f64696d944b028c085f706f7073697a65944b068c035f6d75944b038c075f6d ... (1751 characters truncated) ... 16c75657394682f68314b008594683387945294284b014b288594681189424001000000000000000000500000000000000050997218541cbe26c051d339ed0fa99440010000000000000"')]
(Background on this error at: http://sqlalche.me/e/9h9h)
```

c-bata:
I prepare a script to reproduce the bug and confirmed that this PR fixed it.
https://github.com/optuna/optuna/issues/1775

Before:

```
$ python examples/quadratic_simple.py
...
sqlalchemy.exc.DataError: (MySQLdb._exceptions.DataError) (1406, "Data too long for column 'value_json' at row 1")
[SQL: INSERT INTO trial_system_attributes (trial_id, `key`, value_json) VALUES (%s, %s, %s)]
[parameters: (38, 'cma:optimizer', '"80049528050000000000008c0a636d6165732e5f636d61948c03434d419493942981947d94288c065f6e5f64696d944b028c085f706f7073697a65944b068c035f6d75944b038c075f6d ... (2366 characters truncated) ... 00990201000000f0379026010000003037902601000000702a291301000000c0fe962601000000b025291301001400947494628c085f657073696c6f6e94473e45798ee2308c3a75622e"')]
(Background on this error at: http://sqlalche.me/e/9h9h)
...
optuna.exceptions.StorageInternalError: An exception is raised during the commit. This typically happens due to invalid data in the commit, e.g. exceeding max length.
```


After:

```
$ python examples/quadratic_simple.py 
[I 2020-09-15 18:32:52,087] A new study created in RDB with name: no-name-565356a3-0fcb-4249-a3b9-1b0067645202
...
[I 2020-09-15 18:32:57,672] Trial 29 finished with value: 54.187818214829825 and parameters: {'x': 1.4922967516260823, 'y': 51.96086861991607}. Best is trial 4 with value: 4.09659.
Best value: 4.09659 (params: {'x': -6.36791, 'y': -36.4537})
```

And I also confirmed that this change has no effect to optimization results. See https://gist.github.com/c-bata/28a6334c37f340a2ce8e54c4dee1e2bf#gistcomment-3454630 for details.

HideakiImamura (approved):
Thanks for the PR! This is an amazing alternative to PR #1776. The changes look great to me.

optuna/samplers/_cmaes.py
```diff
@@ -23,6 +23,8 @@
 _logger = logging.get_logger(__name__)
 
 _EPS = 1e-10
+# The value of system_attrs must be less than 2046 characters on RDBStorage.
+_SYSTEM_ATTR_MAX_LENGTH = 2045
```
hvy, line 27: Are there special characters involved? In the schema definition and the error messages say 2048.

optuna/samplers/_cmaes.py
```diff
@@ -413,6 +426,24 @@ def _get_trials(self, study: "optuna.Study") -> List[FrozenTrial]:
...
+        end = min((i + 1) * _SYSTEM_ATTR_MAX_LENGTH, optimizer_len)
+        attrs["cma:optimizer:{}".format(i)] = optimizer_str[start:end]
+    return attrs
+
+
+def _concat_optimizer_attrs(optimizer_attrs: Dict[str, str]) -> str:
+    optimizer_str = ""
+    for i in range(len(optimizer_attrs)):
+        key = "cma:optimizer:{}".format(i)
+        optimizer_str += optimizer_attrs[key]
```
hvy, line 443: I would generally use `str.join` on a list or with a list comprehension for repeated string concatenations (for performance), but I guess it doesn't make a lot of difference for just 2 or 3 iterations. Just a comment.

hvy (commented):
Thanks for the interesting PR. Basically LGTM! I left some questions rather than request if you don't mind.

hvy:
By the way, I first a moment though about providing this splitting under hood of the `Trial`'s and `Study`'s attr manipulation methods to avoid the storage error being propagated to the user. But that might be controversial. On the other hand, we can probably catch and reraise the storage error with a more friendly one.

optuna/samplers/_cmaes.py
c-bata: The value is double-quoted by `json.dumps()`. So `abc` is converted to `"abc"`.
https://github.com/optuna/optuna/pull/1833#discussion_r488526833

I think we can also set `_SYSTEM_ATTR_MAX_LENGTH = 2046`, but I put a margin of 1 character.

optuna/samplers/_cmaes.py
c-bata: Thank you for your comment.
I pushed https://github.com/optuna/optuna/pull/1833/commits/62c3996f01ee7acf80c3c1a42e2d6c96d8dc8b36.

c-bata:
FYI, 2048 character limits is just an implementation detail on RDBStorage, not a specification of BaseStorage. So I think we have two options:

1. Update a specification of BaseStorage, then exporting warning messages if the length of value exceeds 2048 characters limit.
1. Add Trial's and Study's attr manipulation methods in RDBStorage without updating a specification.

I prefer the former approach, but it contains a breaking change (though it's an undefined behavior).

hvy (approved):
LGTM!

hvy:
Let's discuss https://github.com/optuna/optuna/pull/1833#issuecomment-694617161 further. I'll create a separate issue for it.

Issue #1618, opened 2020-08-11
Document exceptions


hvy:
Exceptions that are part of public APIs could be documented using the `Raises:` syntax of Sphinx. For reference, [some methods of the `BaseStorage`](https://github.com/optuna/optuna/blob/master/optuna/storages/_base.py) are documented this way.

I believe this could be a tracking issue until we've covered a decent proportion of the public APIs where one could just leave a note when addressing some particular file/class/submodule/etc.

### _Edit: 2022-02-25 Obsolete list. Issue paused until updated._

{{Crissman edit -- changed to lower list generated by @keisuke-umezawa to keep one list for management.}}

- [x] ./optuna/visualization/_pareto_front.py:plot_pareto_front:22: DAR401: -r ValueError
- [x] ./optuna/storages/_base.py:create_new_study:96: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:delete_study:117: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:set_study_user_attr:131: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:set_study_system_attr:151: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:set_study_directions:171: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:get_study_id_from_name:194: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:get_study_id_from_trial_id:211: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:get_study_name_from_id:228: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:get_study_directions:245: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:get_study_user_attrs:262: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:get_study_system_attrs:279: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:get_all_study_summaries:296: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:create_new_trial:308: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:set_trial_state:330: DAR401: -r NotImplementedError
- [x] ./optuna/storages/_base.py:set_trial_param:360: DAR401: -r NotImplementedError
[...]

PR #1857, opened 2020-09-24
Add sphinx doc comments regarding exceptions in the optimize method

yuk1ty:
## Motivation

See issue #1618 for the original motivation. This issue is good for me to look through Optuna.

## Description of the changes

I'm trying to work on the issue #1618. This PR is a partial fulfillment of adding `Raises:` comments to docs. To begin with, I added a comment to `optimize` method.

Could anyone tell me about the following questions?
- How much comment should I write, for instance, if additional exceptions are thrown from inside of `optimize` method? 
- Is the comment in my PR sufficient?

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -304,6 +304,10 @@ class Study(BaseStudy):
                 Flag to show progress bars or not. To disable progress bar, set this ``False``.
                 Currently, progress bar is experimental feature and disabled
                 when ``n_jobs`` :math:`\\ne 1`.
+
+        Raises:
+            RuntimeError:
+                If nested invocation of this method occurs.
         """
 
         if not isinstance(catch, tuple):


optuna/study.py
```diff
@@ -304,6 +304,12 @@ def objective(trial):
                 Flag to show progress bars or not. To disable progress bar, set this ``False``.
                 Currently, progress bar is experimental feature and disabled
                 when ``n_jobs`` :math:`\\ne 1`.
+        
+        Raises:
+            TypeError:
+                If the catch argument is not a tuple.
```
crcrpar, line 310: [nit-picking]
```suggestion
                If the ``catch`` argument is not a tuple.
```

yuk1ty: Ah, thank you! I fixed it.

optuna/study.py
```diff
@@ -304,6 +304,12 @@ def objective(trial):
                 Flag to show progress bars or not. To disable progress bar, set this ``False``.
                 Currently, progress bar is experimental feature and disabled
                 when ``n_jobs`` :math:`\\ne 1`.
+
```
HideakiImamura, line 307: You can pass the CI by removing white spaces.
```suggestion

```

HideakiImamura (commented):
Thanks for the PR! I have a minor comment.

optuna/study.py
yuk1ty: I missed it. Thank you! Fixed.

hvy:
Thanks a lot for the PR, the changed looks good overall. I'm wondering though if we should skip documenting the `TypeError` though.

Python is a dynamically typed language and while we do check the types in certain APIs, we also don't in many (nor cover in unit tests). The expected types are communicated via the type hints and we sort of trust the user to follow these types. Unfortunately, the behavior is not really consistent and maintaining the documentation might not be worth it, at least for the time being.

yuk1ty:
> The expected types are communicated via the type hints and we sort of trust the user to follow these types. Unfortunately, the behavior is not really consistent and maintaining the documentation might not be worth it, at least for the time being.

I understood & I agree with that point so let me remove the `TypeError` comment.

HideakiImamura (approved):
LGTM!

optuna/study.py
```diff
@@ -304,6 +304,10 @@ def objective(trial):
...
+        Raises:
+            RuntimeError:
+                If nested invocation of this method occurred.
```
hvy, line 310: Sorry for the late review but could you just fix the tense to be consistent with e.g. https://github.com/optuna/optuna/blob/master/optuna/storages/_base.py#L105 ?
```suggestion
                If nested invocation of this method occurs.
```

yuk1ty: Sure!

hvy (approved):
Thanks for the swift response, LGTM!

hvy:
As for how to document exceptions in general, I think it's difficult to say. If we want to make it a part of the interface, then it should be documented. I guess we have to deal with this case by case since documenting every possible exception is infeasible.

PR #1861, opened 2020-09-24
Avoid global study in `Study.stop` testcode

hvy:
## Motivation

Python variable scoping isn't necessarily trivial, especially the testcode in `Study.stop`. The testcode should be as pedagogical as possible.

## Description of the changes

Avoids accessing the study object from a global in the objective function of the testcode in `Study.stop`. Instead, accesses the study object via the local trial.

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -558,7 +558,7 @@ class Study(BaseStudy):
 
                 def objective(trial):
                     if trial.number == 4:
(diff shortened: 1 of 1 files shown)


HideakiImamura (approved):
Thanks for the update! LGTM!

toshihikoyanase (approved):
Good catch! LGTM.

PR #1864, opened 2020-09-26
Move away from .scoring imports

norihitoishida:
## Motivation
`FutureWarning` occured in `OptunaSearchCV`.

## Description of the changes
The `sklearn.metrics.scorer` module is  deprecated in version 0.22 and will be removed in version 0.24. 
The corresponding classes / functions should instead be imported from `sklearn.metrics`.

## Same problems in skorch
https://github.com/skorch-dev/skorch/issues/573
https://github.com/skorch-dev/skorch/pull/575


crcrpar (approved):
nice work, thank you.

HideakiImamura (approved):
Thanks for the good catch! LGTM.

Issue #1865, opened 2020-09-27
n_trials not respected when using database


Madman600:
Running a study with rdb backend does not respect n_trials parameter.

For example if I modify "First Optimization" code to use rdb backend like so:

```
import optuna


def objective(trial):
    x = trial.suggest_uniform('x', -10, 10)
    return (x - 2) ** 2


storage = "mysql://optuna@192.168.10.3/optuna"

study = optuna.create_study(study_name="test", storage=storage, load_if_exists=True)
study.optimize(objective, n_trials=100)

print(study.best_params)

study.best_params

study.best_value

study.best_trial

study.trials

len(study.trials)

study.optimize(objective, n_trials=100)

len(study.trials)

```
The second call to study.optimize should not execute anymore trials because we have already performed 100 trials. Instead it performs another hundred trials.

Madman600:
https://github.com/optuna/optuna/pull/1866

hvy:
This might not be intuitive at first but it's by design. `n_trials` specifies how many trials `Study.optimize` should run, not the total number of trials for the study. If you do distributed training, you should make sure that this argument takes into consideration the number of workers, i.e. `n_trials = n_trials_total // n_workers`, if that makes any sense.

Madman600:
You're right I am confused. Say I'm running 100 studies and I want to run 5000 trials for each study. I kick off the studies, each one with a unique name in the database. I run 10 studies at a time. After some time I need to stop for some reason. 20 studies have had 5000 trials run but studies 21-30 have had 2000 trials run. Now I want to resume where I left off. I start the whole process again. How I think things should work is the database is queried again and we know that 5000 trials have already been run for the first 20 studies so we move to the 21-30 batch where 2000 trials have been run and resume at trial 2001. Pulling n_trials from the database would allow this. The way things are now even though the study is in the database and 5000 trials have been run, the code will act as if no trials have been run.

hvy:
Thank you for the concrete example, that makes it easier to discuss. 

> the code will act as if no trials have been run.

is not really true, if you resume a study. The sampling algorithm will retrieve the previous 2000 trials from your database in order to suggest the 2001th one. It's true however the that `Study.optimize` will run for another `n_trials` trials. 

[...]

PR #1866, opened 2020-09-27
Load i_trial from storage

Madman600:
## Motivation

If i'm executing a distributed study and expecting it to "resume", n_trials should respect the number of trials already executed.

## Description of the changes

Loads i_trial from storage.


hvy:
Can we close this PR as #1865 has been resolved?

PR #1882, opened 2020-10-01
Removed with_trace method from docs

### LINKED ISSUE #1615: Remove the document for `with_traceback` method of Optuna's exception classes
Currently, Optuna's exception classes have the documentations of `with_traceback` method, which is inherited from `Exception`. I don't think it is informative for readers and it can be removed from the reference.

image

The following `Exception` has the `with_traceback` method.
- [ ] `optuna.exceptions.CLIUsageError`
- [ ] `optuna.exceptions.OptunaError`
- [ ] `optuna.exceptions.TrialPruned`
- [ ] `optuna.exceptions.CLIUsageError`
- [ ] `optuna.exceptions.StorageInternalError`
- [ ] `optuna.exceptions.DuplicatedStudyError`

CC @keisuke-umezawa Please let me know if you have any comments.

i-am-jeetu:
## Motivation
Fixes #1615:  Optuna's exception classes have the documentations of with_traceback method, which is inherited from Exception. It is not informative for readers.

## Description of the changes
Configured sphinx autodoc to ignore all ``with_traceback`` methods while generating documentations.

Closes: #1615

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -172,6 +172,7 @@ autosummary_generate = True
 autodoc_default_options = {
     'members': True,
     'inherited-members': True,
+    'exclude-members': 'with_traceback',
 }
 
 sphinx_gallery_conf = {


toshihikoyanase:
Thank you for your PR!

A possible alternative would be to use the `:exclude-members:` directive at each class. We removed such directives in the past PR, but IMHO, we can use it here as an exception.
I'd like to know the other reviewers' opinions.
CC @keisuke-umezawa @crcrpar

crcrpar:
That sounds reasonable

i-am-jeetu:
Ok @toshihikoyanase, I will amend my PR. Using ``:exclude-members:`` directive at each class is much cleaner.

toshihikoyanase:
@crcrpar Thank you for your reply, but I'm not quite sure if you agree with the change in this PR or the alternative approach. If you are for this PR, could you approve it?

crcrpar (approved):
Either approach is fine for me.

toshihikoyanase (approved):
@crcrpar Thank you for your reply. I agree.

@i-am-jeetu
I prefer the simple approach of this PR, so let me merge it.

toshihikoyanase:
Let me slightly update the title for the release note.

PR #1891, opened 2020-10-02
Run documentation build and doctest in GitHub Actions.

crcrpar:
We want to migrate all the required checks to github actions as we saw some weird connection problems with Circle CI (at least twice).

It may or may not be better to use [the official images of readthedocs](https://hub.docker.com/r/readthedocs/build).

### DIFF
--- a/.github/workflows/sphinx-build.yml
+++ b/.github/workflows/sphinx-build.yml
@@ -17,7 +17,7 @@ jobs:
 
     - uses: actions/setup-python@v2
       with:
-        python_version: 3.8
+        python-version: 3.8
 
     - name: Install Dependencies
       run: |
@@ -45,7 +45,7 @@ jobs:
 
     - uses: actions/setup-python@v2
       with:
-        python_version: 3.8
+        python-version: 3.8
 
     - name: Install Dependencies
       run: |
--- a/.github/workflows/sphinx-build.yml
+++ b/.github/workflows/sphinx-build.yml
@@ -1,0 +1,59 @@
+name: Sphinx
+
+on:
+  push:
+    branches:
+      - master
+  pull_request: {}
+
+jobs:
+  documentation:
+
+    runs-on: ubuntu-latest
+
+    steps:
+
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-python@v2
+      with:
+        python_version: 3.8
+
+    - name: Install Dependencies
+      run: |
+        python -m pip install -U pip
+        pip install --progress-bar off -U .[document]
+
+    - name: Build Document
+      run: |
+        cd docs
+        make html -j
+        cd ../
+
+    - uses: actions/upload-artifact@v2
+      with:
+        name: built-document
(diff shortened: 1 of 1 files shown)


.github/workflows/sphinx-build.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    - name: Install Dependencies
+      run: |
+        python -m pip install -U pip
+        pip install --progress-bar off -U .[document]
+
+    - name: Build Document
+      run: |
+        cd docs
+        make html -j
```
toshihikoyanase, line 30: Why did you add `-j` option?
It was ignored in the `Makefile` when I dumped the following command:

https://github.com/optuna/optuna/blob/dc032afdda0e0248bb70831ee47ca10aeca3bd28/docs/Makefile#L20

The command dump:

```console
$ make html -j
echo @sphinx-build -M html "source" "build" -W --keep-going 
@sphinx-build -M html source build -W --keep-going
```

toshihikoyanase (commented):
Thank you for your PR as always.

> It may or may not be better to use the official images of readthedocs.

I prefer the official image since we can reduce environment-dependent issues. On the other hand, this [document](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on) says that we have only five supported runners, and I guess we cannot use the custom images.

.github/workflows/sphinx-build.yml
crcrpar: It's just my habit. Removed.

crcrpar:
We can set a container with `container` argument. Ref: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer

Though, after some investigation, I found that I can't make that happen at the moment because
- `actions/checkout` requires git 2.18 or higher
- the version of `readthedocs/build` images' git is 2.17
- sudo / su not available in `readthedocs/build`

.github/workflows/sphinx-build.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-python@v2
+      with:
+        python_version: 3.8
```
HideakiImamura, line 20: A nit. 
```suggestion
        python-version: 3.8
```

.github/workflows/sphinx-build.yml
```diff
@@ -0,0 +1,59 @@
...
```
HideakiImamura, line 48: ditto.
```suggestion
        python-version: 3.8
```

HideakiImamura (commented):
Thanks for the PR! The added workflow almost LGTM. I have minor comments. Could you take a look?

HideakiImamura (approved):

toshihikoyanase (approved):
LGTM. Thank you!

PR #1895, opened 2020-10-04
Remove unnecessary options of `black` and `isort`

akihironitta:
## Motivation
Not sure if this fix is even necessary, but let me point out that there is a redundant option for both `isort` and `black`. `--diff` option itself includes `--check` option as well, so I believe `--check` is unnecessary.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
           name: black
           command: |
             . venv/bin/activate
-            black . --check --diff
+            black . --diff
 
       - run:
           name: flake8
@@ -55,7 +55,7 @@ jobs:
           name: isort
           command: |
             . venv/bin/activate
-            isort . --check --diff
+            isort . --diff
 
       - run:
           name: mypy


hvy:
Thanks for your PR. Are you sure about the inclusion though? In my local runs, both with isort and black, the exit code depended on whether `--check` was explicitly set or not. Specifically, the codes were 0 when executed without `--check`, even when changes would were required. With `--check` the codes were 1.

akihironitta:
@hvy Thank you for your quick response.

Just checked them locally and realised that there is actually a difference in whether `--check` is given as you mentioned, so let's keep it as is.

I am so sorry for taking your time and confusing with my false statement.

hvy:
No worries, rather thanks for the consideration. It was also fruitful going through the specs of isort/black. Let me close this PR for now as these changes seemingly won't be needed.

hvy:
Maybe I should've waited with closing. Were you fine with that, or did you have had something else on your mind?

akihironitta:
I am totally fine with this PR to be closed. Thank you again!

Issue #1852, opened 2020-09-21
example for a custom wrapper on lightgbm.fit or .train with pruning


mirekphd:
I try to integrate Optuna pruners with our custom cross-validation function (a wrapper on `lightgbm.fit()`) to which I try to pass an `optuna.integration.LightGBMPruningCallback` instance via its `callbacks` argument, but then I get this error:

```
/opt/conda/lib/python3.8/site-packages/optuna/integration/lightgbm.py in __call__(self, env)
    118             return None
    119 
--> 120         raise ValueError(
    121             'The entry associated with the validation name "{}" and the metric name "{}" '
    122             "is not found in the evaluation result list {}.".format(

ValueError: The entry associated with the validation name "valid_0" and the metric name "auc" is not found in the evaluation result list [('oof', 'auc', 0.9398809523809524, True)].
```

The call sequence is: 
```
study.optimize(objective_instance) -> Objective_custom.__call__(self, trial) -> custom_wrapper(.. , callbacks=[lgbm_pruning_callback]) -> lightgbm.fit(callbacks=callbacks)
```

with `lgbm_pruning_callback = optuna.integration.LightGBMPruningCallback(trial, self.metric)` created inside `Objective_custom.__call__()` before the call to `custom_wrapper` .

Note it works fine without our wrapper (when passing callbacks directly to standard `lightgbm.cv`).

toshihikoyanase:
Hi @mirekphd 

The following message suggests that you use the custom `valid_name` (i.e., `oof`).

> ValueError: The entry associated with the validation name "valid_0" and the metric name "auc" is not found in the evaluation result list [('oof', 'auc', 0.9398809523809524, True)].

I think the pruner can find the metric value if you specify it at `LightGBMPruningCallback.__init__`  as follows:

```python
lgbm_pruning_callback = LightGBMPruningCallback(trial, metric="auc", valid_name="oof")
```

mirekphd:
Thank you so much for the answer! I went back to our custom objective function and added `valid_name` argument to the existing line `lgbm_pruning_callback = LightGBMPruningCallback(trial, metric=self.metric)`.

I set `valid_name` to the value reported in the above error message (i.e. `LightGBMPruningCallback(.., valid_name="oof"`) and then pruning ( `HyperbandPruner` to be more specific) started co-operating correctly with our custom cross-validation function!:) 

[...]

PR #1119, opened 2020-04-15
Add pruning to PyTorch example

Crissman:
Our PyTorch example is one of our most popular, but does not show how to use pruning with PyTorch, which is an advantage of Optuna over other hyperparameter optimization libraries. I've added pruning and would like to link to this version for my blog post on PyTorch with Optuna.

PR #1896, opened 2020-10-04
add lgbm pruning example w/ custom cv fun

mirekphd:
## Motivation
As a follow-up to #1852 (solved with the kind help of @toshihikoyanase ), and possibly contributing also some ideas towards solving #1119 (as noted by @bs1119 in #1852) , this PR is aimed to demonstrate how Optuna pruners can be combined with **custom cross-validation functions**, both for classification and regression models, with early stopping and without. 

In this example, we use Optuna to optimize a classification or regression model, with a custom CV function `custom_cv_fun()` applying LightGBM's `.fit()` method to each CV fold separately, instead of using standard `lgbm.cv()`. This modification aims to improve out-of-fold (CV) model accuracy, by using and individual, data-derived number of iterations for each fold model, which in theory should reduce under- and overfitting, thus improving accuracy of the standard `lightgbm.cv` implementation that uses a single common number of iterations for each fold (even when early stopping is enabled). Deterioration in computational performance is traded-off for improved predictive performance.

Throughout training, an Optuna pruner observes model metrics returned by the custom CV function and - if necessary - stops early unpromising trials via a callback, resulting in noticeably faster execution times with unaffected accuracy.

Special attention is paid to model reproducibility, with precise control over all relevant seed values, to let users arrive at the same results over repeated studies. Execution time comparisons between pruning and no-pruning setups required adding controls for the numbers of threads and workers used by both Optuna and the underlying `lightgbm` methods (but subject to reproducibility constraints which unfortunately reduce Optuna's multiprocessing performance).

## Description of the changes
A single independent python 3 script conforming to existing conventions is added to the folder dedicated for pruning examples.
The script was developed and tested under `lightgbm==3.0.0` with `optuna==2.1.0` in python 3.8, and reformatted with `black`.

### DIFF
--- a/examples/pruning/lightgbm_with_custom_cv_function.py
+++ b/examples/pruning/lightgbm_with_custom_cv_function.py
@@ -12,11 +12,6 @@ The example emphasizes reproducibility, giving the user precise control over all
 You can run this example using python 3 interpreter, e.g. as follows:
     $ python lightgbm_with_custom_cv_function.py
     
-Notes: 
-- OBJECTIVE can switch between "binary" / "classification" (default) and "regression" examples,
-- USE_PRUNING can turn off pruning if changed from default True to False,
-- USE_STOPPING can turn off early stopping if changed from default True to False,    
-    
 """
 
 # Import packages
@@ -192,18 +187,12 @@ class Objective_custom:
     def __call__(self, trial):
 
         params = {
-            "bagging_fraction": float(
-                trial.suggest_float("bagging_fraction", 0.1, 1, log=False)
-            ),
+            "bagging_fraction": float(trial.suggest_float("bagging_fraction", 0.1, 1, log=False)),
             "bagging_freq": trial.suggest_int("bagging_freq", 1, 100, log=True),
-            "feature_fraction": float(
-                trial.suggest_float("feature_fraction", 0.1, 1, log=False)
-            ),
+            "feature_fraction": float(trial.suggest_float("feature_fraction", 0.1, 1, log=False)),
             "lambda_l1": float(trial.suggest_float("lambda_l1", 1e-06, 100, log=True)),
             "lambda_l2": float(trial.suggest_float("lambda_l2", 1e-06, 100, log=True)),
-            "learning_rate": float(
-                trial.suggest_float("learning_rate", 0.0001, 0.3, log=True)
-            ),
+            "learning_rate": float(trial.suggest_float("learning_rate", 0.0001, 0.3, log=True)),
             # caution: num_boost_round should be used as argument, not passed among hyperparameters to params
             # 'num_boost_round': trial.suggest_int('num_boost_round', 100, 2000, log=False),
             "num_leaves": trial.suggest_int("num_leaves", 3, 100, log=True),
@@ -225,9 +214,7 @@ class Objective_custom:
             # random boosting and stopping rounds
             # will be passed as arguments
             num_boost_round = trial.suggest_int("num_boost_round", 100, 2000, log=False)
-            early_stopping_rounds = trial.suggest_int(
-                "early_stopping_rounds", 2, 500, log=True
-            )
+            early_stopping_rounds = trial.suggest_int("early_stopping_rounds", 2, 500, log=True)
         else:
             num_boost_round = self.fixed_rounds
             early_stopping_rounds = None
@@ -250,9 +237,7 @@ class Objective_custom:
 
         # get mean CV metric from the appropriate key
         # of the dict returned by the custom CV function
-        eval_mean_metric = cv_results_dict["folds_mean_data"][
-            "eval_mean-" + self.eval_metric
-        ]
+        eval_mean_metric = cv_results_dict["folds_mean_data"]["eval_mean-" + self.eval_metric]
 
         print("Mean CV metric: %.5f" % eval_mean_metric)
 
@@ -366,9 +351,7 @@ def main():
         verbosity=VERBOSE_LGBM,
     )
 
-    study = optuna.create_study(
-        sampler=OPTUNA_SAMPLER, pruner=OPTUNA_PRUNER, direction=OPT_DIR
-    )
+    study = optuna.create_study(sampler=OPTUNA_SAMPLER, pruner=OPTUNA_PRUNER, direction=OPT_DIR)
 
     start_time = datetime.now()
 
@@ -378,11 +361,7 @@ def main():
     optim_time_custom = datetime.now() - start_time
     print(
         "\nOptuna+custom fun. study with %d trials took %.2f s. Time per trial: %.2f s."
-        % (
-            TRIALS,
-            optim_time_custom.total_seconds(),
-            optim_time_custom.total_seconds() / TRIALS,
-        )
+        % (TRIALS, optim_time_custom.total_seconds(), optim_time_custom.total_seconds() / TRIALS)
     )
 
     optuna_best_metrics["custom"] = study.best_value
--- a/examples/pruning/lightgbm_with_custom_cv_function.py
+++ b/examples/pruning/lightgbm_with_custom_cv_function.py
@@ -1,20 +1,20 @@
 # Functions
 
 """
-This configurable Optuna example demonstrates the use of pruning when using 
+This configurable Optuna example demonstrates the use of pruning when using
 custom cross-validation functions based on LightGBM classifiers or regressors.
 
-In this example, we optimize cross-validation accuracy of either a classification model 
-of breast cancer probability or a regression model of Boston housing prices, 
-with a custom CV function custom_cv_fun() applying LightGBM's .fit() method 
-to each cross-validation fold separately, instead of using the standard 
+In this example, we optimize cross-validation accuracy of either a classification model
+of breast cancer probability or a regression model of Boston housing prices,
+with a custom CV function custom_cv_fun() applying LightGBM's .fit() method
+to each cross-validation fold separately, instead of using the standard
 built-in .cv() function.
 
-Throughout training, an Optuna pruner observes intermediate results and if necessary 
+Throughout training, an Optuna pruner observes intermediate results and if necessary
 stops early unpromising trials using a callback, resulting in noticeably faster execution
 times with unaffected accuracy.
 
-The example emphasizes reproducibility, giving the user precise control over all seed 
+The example emphasizes reproducibility, giving the user precise control over all seed
 values that are being used to let her arrive at the same results over repeated studies.
 
 You can run this example using python 3 interpreter, e.g. as follows:
--- a/examples/pruning/lightgbm_with_custom_cv_function.py
+++ b/examples/pruning/lightgbm_with_custom_cv_function.py
@@ -1,0 +1,415 @@
+# Functions
+
+"""
+This configurable Optuna example demonstrates the use of pruning when using custom cross-validation functions based on LightGBM classifiers or regressors.
+
+In this example, we optimize cross-validation accuracy of either a classification model of breast cancer probability or a regression model of Boston housing prices, with a custom CV function custom_cv_fun() applying LightGBM's .fit() method to each cross-validation fold separately, instead of using the standard built-in .cv() function.
+
+Throughout training, an Optuna pruner observes intermediate results and if necessary stops early unpromising trials using a callback, resulting in noticeably faster execution times with unaffected accuracy.
+
+The example emphasizes reproducibility, giving the user precise control over all seed values that are being used to let her arrive at the same results over repeated studies.
+
+You can run this example using python 3 interpreter, e.g. as follows:
+    $ python lightgbm_with_custom_cv_function.py
+    
+Notes: 
+- OBJECTIVE can switch between "binary" / "classification" (default) and "regression" examples,
+- USE_PRUNING can turn off pruning if changed from default True to False,
+- USE_STOPPING can turn off early stopping if changed from default True to False,    
+    
+"""
+
+# Import packages
+
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


mirekphd:
The funny thing is that the formatting "problems" that `black` is now finding in my file causing "Checks / check" to fail (see [Details](https://github.com/optuna/optuna/pull/1896/checks?check_run_id=1206183006)) were introduced... by `black` itself...:) with an apparently narrower maximum line width on my side than on yours. I can undo those black's style changes manually on my side to match your version, but they will re-emerge next time I use it, so we had better find the reason behind them and file it as an Issue in `black` to prevent similar future problems.

HideakiImamura:
What is the version of `black` in your local environment? The latest `black` is `20.8b1` and you have to the latest version.

mirekphd:
> 
> 
> What is the version of `black` in your local environment? The latest `black` is `20.8b1` and you have to the latest version.

Looks like we have the same version:
```
$ black --version
black, version 20.8b1

$ pip freeze | grep black
black==20.8b1
```

I wander: is black perhaps unstable over multiple runs...? I did run it definitely more than once on this file and initially these problematic lines looked OK - just like "your" black wants them to look...

HideakiImamura:
> I wonder: is black perhaps unstable over multiple runs...?

Hmm... I don't think so. I usually run `black` multiple times in my local environment, but I have no bad effects. 

Do you run `black .` at the optuna directory (the cloned directory from https://github.com/optuna/optuna)? In addition, you have to use Optuna's [pyproject.toml](https://github.com/optuna/optuna/blob/master/pyproject.toml), which includes the setting of the `black`. (It is contained in the optuna directory as a default if you cloned https://github.com/optuna/optuna.)

toshihikoyanase:
@mirekphd Maybe this is not related to your issue, but I encountered a similar issue when I use `black` extension of vscode. I couldn't fix the issue and I stop using the extension now. If you use IDEs, could you run `black` in your terminal?

(Please ignore my comment, if you don't use IDE.)

mirekphd:
I believe `black`'s default max line length is 88 characters (see [docs](https://test-black.readthedocs.io/en/latest/the_black_code_style.html#line-length)), and you bumped it to a more permissive 99 chars here:
https://github.com/optuna/optuna/blob/3163606048374becbec7f238026e8d18c9ef8310/pyproject.toml#L2
which would explain nicely why my "blackened" code looks narrower than that expected by CICD's `black`.

Later in the evening I will revert the code to pre-`black` version and re-run `black` with `--line-length 99` to see if I can replicate your results.

@toshihikoyanase : you can perhaps try the same method if the VSC extension can be parametrized in such a way.

mirekphd:
> I believe `black`'s default max line length is 88 characters (see [docs](https://test-black.readthedocs.io/en/latest/the_black_code_style.html#line-length)), and you bumped it to a more permissive 99 chars here:
> 
> https://github.com/optuna/optuna/blob/3163606048374becbec7f238026e8d18c9ef8310/pyproject.toml#L2
> 
> which would explain nicely why my "blackened" code looks narrower than that expected by CICD's `black`.

Yes, different `black` line length settings are the reason behind the problem with the `Checks/check black` test failure we have above.

As a proof, compare output of those two "black diffs":

`optuna/optuna` settings:
```
$ black --diff --color --line-length 99 lightgbm_with_custom_cv_function.py
```

vs. `black`'s default:
```
$ black --diff --color --line-length 88 lightgbm_with_custom_cv_function.py
```

applied to the original file: 
[lightgbm_with_custom_cv_function.zip](https://github.com/optuna/optuna/files/5354607/lightgbm_with_custom_cv_function.zip)

IMHO we could easily prevent this confusion from repeating itself for other long-line coders in the future by:
- documenting (e.g. [here](https://github.com/optuna/optuna/blob/master/CONTRIBUTING.md)) the non-standard line length, 
- passing the custom `--line-length=99` explicitly as an argument of `black` in the CICD command line, because it currently has no arguments, which suggests that `black`'s defaults can be used to reproduce the expected formatting.

mirekphd:
`flake8` has `--max-line-length` defaulting to 99, but nevertheless it raised 4 "line too long" errors even after passing `black --line-length 99` check, because unlike `black`, `flake8` checks also line length within the docstring...:
```
./examples/pruning/lightgbm_with_custom_cv_function.py:4:100: E501 line too long (154 > 99 characters)
./examples/pruning/lightgbm_with_custom_cv_function.py:6:100: E501 line too long (337 > 99 characters)
./examples/pruning/lightgbm_with_custom_cv_function.py:8:100: E501 line too long (206 > 99 characters)
./examples/pruning/lightgbm_with_custom_cv_function.py:10:100: E501 line too long (173 > 99 characters)
```

mirekphd:
Now that automated checks have been finally passed, I suppose we are ready for the proper code review?:)

toshihikoyanase:
Sorry for the delayed response. I think the code will help LGBM users, and I'll review the code in a day or two.

examples/pruning/lightgbm_with_custom_cv_function.py
```diff
@@ -0,0 +1,404 @@
+# Functions
+
```
toshihikoyanase, line 2: I think this comment can be removed.
```suggestion
```

examples/pruning/lightgbm_with_custom_cv_function.py
```diff
@@ -0,0 +1,404 @@
...
+    # get feature names
+    feature_names = list(X.columns)
+
+    # split data into k folds
+    kf = KFold(n_splits=nfold, shuffle=True, random_state=random_state)
+    splits = kf.split(X, y)
+
+    # iterate over folds
+    for n, (train_index, valid_index) in enumerate(splits):
```
toshihikoyanase, line 72: `n` is not used and we can remove it.
```suggestion
    for train_index, valid_index in splits:
```

examples/pruning/lightgbm_with_custom_cv_function.py
```diff
@@ -0,0 +1,404 @@
...
+        seed=SEED,
+        verbosity=VERBOSE_LGBM,
+    )
+
+    study = optuna.create_study(sampler=OPTUNA_SAMPLER, pruner=OPTUNA_PRUNER, direction=OPT_DIR)
+
+    start_time = datetime.now()
+
+    # run Optuna optimization over the specified number of trials
+    study.optimize(objective, n_trials=TRIALS, n_jobs=WORKERS_NUM)
```
toshihikoyanase, line 369: IMO, `n_jobs>1` will hardly improve the performance due to the GIL. How about stop using it?
Even so, `n_jobs` of LightGBM will utilize the multi-core processors effectively.

```suggestion
    study.optimize(objective, n_trials=TRIALS)
```

examples/pruning/lightgbm_with_custom_cv_function.py
```diff
@@ -0,0 +1,404 @@
...
+        "seed": SEED,
+        "verbosity": VERBOSE_LGBM,
+    }
+    all_best_params = {**optimized_best_params, **static_params}
+    optuna_best_params["custom"] = all_best_params
+
+    print(
+        "\nBest %s metric for optuna+custom fun. (CV models mean reported by Optuna): %.5f\n"
+        % (METRIC, optuna_best_metrics["custom"])
+    )
```
toshihikoyanase, line 397: Please use `format` instead of `%`.
```suggestion
    print(
        "\nBest {} metric for optuna+custom fun. (CV models mean reported by Optuna): {:.5f}\n".format(
            METRIC, optuna_best_metrics["custom"]
        )
    )
```

examples/pruning/lightgbm_with_custom_cv_function.py
```diff
@@ -0,0 +1,404 @@
...
+
+        # collect current fold data
+        fold_best_iterations.append(model.best_iteration_)
+        fold_best_scores.append(model.best_score_[eval_name])
+
+    # average folds iterations numbers
+    folds_data = {}
+    folds_data["best_iterations_mean"] = -1
+    if fold_best_iterations[0] is not None:
+        folds_data["best_iterations_mean"] = int(np.mean(fold_best_iterations))
```
toshihikoyanase, line 111: [Question] Could you tell me when the `model.best_iteration_` returns `None`?

examples/pruning/lightgbm_with_custom_cv_function.py
```diff
@@ -0,0 +1,404 @@
...
+
+        valid_y_df = pd.DataFrame({"y": valid_y})
+
+        return train_x_df, valid_x_df, train_y_df, valid_y_df
+
+    else:
+        return train_x, valid_x, train_y, valid_y
+
+
+class Objective_custom:
```
toshihikoyanase, line 167: Please use camel case (c.f., [coding style convention](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#example))
```suggestion
class ObjectiveCustom:
```

toshihikoyanase (changes requested):
Thank you for your PR. This example has many code comments, and therefore it is quite kind for readers. On the other hand, it is a bit longer than the other examples. If we omit some unnecessary parts, I think users can focus more on the core part of this example (i.e., `cusotm_cv_fun` and `Objective_custom`). What do you think?

examples/pruning/lightgbm_with_custom_cv_function.py
mirekphd: Here I have to kindly disagree:) From my experience on multi-CPU machines (in docker containers), muli-threading built into LGBM has its limits (notably, by Guolin Ke's own admission in one discussion on Githhub) LGBM internal parallelism does not work beyond certain point (specifically, when the user tries to access threads from more than 1 physical CPU socket). In practice, the optimum is around 8-16 threads, depending on data shape. You can see some of my public results trying to establish such optimum [here](https://github.com/mirekphd/gbdt-algos-performance-tests/blob/master/20200812/analyze-results-of-gpu-vs-cpu-tests-multiple-versions-20200812.ipynb).

Enter multi-processing (tested either with `loky` or `multiprocessing` backend, as `dask` does not work), which allows us to bypass this limitation of multithreading built into LightGBM, because processes can be run on separate CPU sockets. If you wish, I can supply results of additional performance tests of this PR script, e.g. time taken on average by an Optuna trial) for the multiprocessing vs. multithreading case?

examples/pruning/lightgbm_with_custom_cv_function.py
mirekphd: Apologies - here the comment would be probably in order. This occurs when the user switches off early stopping (by setting `USE_STOPPING=False`, and then the `fold_best_iterations` list is populated with `None` elements (as returned by `model.best_iteration_` in such cases).

examples/pruning/lightgbm_with_custom_cv_function.py
mirekphd: For consistency this change is also required in [#L349](https://github.com/mirekphd/optuna/blob/lgbm-w-cust-cv-fun/examples/pruning/lightgbm_with_custom_cv_function.py#L349)

examples/pruning/lightgbm_with_custom_cv_function.py
mirekphd: The most likely reason why multiprocessing did not seem to work "out-of-the-box" was this reproducibility enforcement in my example attained by... switching off multiprocessing by setting `WORKERS_NUM` back from 2 to 1 (in line [#L311](https://github.com/mirekphd/optuna/blob/lgbm-w-cust-cv-fun/examples/pruning/lightgbm_with_custom_cv_function.py#L311)).

We might change the current setting of `MAKE_REPRODUCIBLE` from `True` to `False` or set `WORKERS_NUM` to 1 instead of 2 (to require action that makes it more likely to notice the role of `MAKE_REPRODUCIBLE`) if you think others will also make the same conclusions about GIL being the culprit here, rather than the desire to obtain the same result at each Optuna run.

examples/pruning/lightgbm_with_custom_cv_function.py
mirekphd: I replaced `%`-style formatting with `{}`-style formatting also in [#L373](https://github.com/mirekphd/optuna/blob/lgbm-w-cust-cv-fun/examples/pruning/lightgbm_with_custom_cv_function.py#L373).

examples/pruning/lightgbm_with_custom_cv_function.py
toshihikoyanase: Thank you for your detailed comments. 

> `MAKE_REPRODUCIBLE `

I agree. To make the optimization reproducible, we need to set `n_jobs=1`. So, I think it's reasonable to set `WORKERS_NUM` to `1` by default.

>  switching off multiprocessing by setting `WORKERS_NUM` back from 2 to 1

A nit, but Optuna utilizes the multi-threading, not multiprocessing when we set `n_jobs>1`.

examples/pruning/lightgbm_with_custom_cv_function.py
toshihikoyanase: Thank you for adding the comment to L107 :+1:

toshihikoyanase (commented):
Thank you for your update.

I'm still wondering if we can make this example short. The lengths of the other examples are less than 200 lines, but this example has more than 400 lines.

I think this is due to the following reasons:
- it includes optional scenarios which can be executed if users edit the file, and
- it defines many variables to extract experimental settings from the execution of the experiment.

If we focus on the main scenario and directly input the variable, I think we can halve the code length. I created a prototype, so please take a look at it.
https://github.com/toshihikoyanase/optuna/blob/simplified-lgbm-w-cust-cv-fun/examples/pruning/lightgbm_with_custom_cv_function.py

PR #1904, opened 2020-10-06
Refactor `Study.optimize`

hvy:
# Motivation

`Study` and `study.py` are becoming quite large and difficult to read and maintain. This PR aims to address the problem by refactoring `Study.optimize`.

# Description 

- Refactors `Study.optimize` by e.g. moving the main logic out to a new file
- Introduces private "ask and tell" interfaces (which could be made public in the future)
- Public API isn't affected

## TODO

- ~I have basically only moved the code out to separate location and would like to go over the necessities of the involved functions and their interfaces~
- ~Ensure unit tests pass~

PR #1907, opened 2020-10-06
Refactor `Study.trials_dataframe`

hvy:
## Motivation

See https://github.com/optuna/optuna/pull/1904. Basically the same, but for `Study.trials_dataframe` instead of `Study.optimize`. 

I could see reasons for not merging this since it adds another level of indirection and a new file partially to slim down `study.py`. However, we can isolate the dependency of `pandas` to a more confined scope, namely `_dataframe.py`, rather than `study.py`. Additionally, the logic is suited as a free function since it didn't used to rely on `self` besides from a single public API call, to obtain a list of all trials.

## Description of the changes

See above.

### DIFF
--- a/optuna/_dataframe.py
+++ b/optuna/_dataframe.py
@@ -1,0 +1,83 @@
+import collections
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Set
+from typing import Tuple
+
+import optuna
+from optuna._imports import try_import
+from optuna.trial._state import TrialState
+
+
+with try_import() as _imports:
+    # `trials_dataframe` is disabled if pandas is not available.
+    import pandas as pd  # NOQA
+
+# Required for type annotation in `Study.trials_dataframe`.
+if not _imports.is_successful():
+    pd = object  # type: ignore # NOQA
+
+
+def _trials_dataframe(
+    study: "optuna.Study",
+    attrs: Tuple[str, ...],
+    multi_index: bool,
+) -> "pd.DataFrame":
+    _imports.check()
+
+    trials = study.get_trials(deepcopy=False)
+
+    # If no trials, return an empty dataframe.
+    if not len(trials):
+        return pd.DataFrame()
+
+    attrs_to_df_columns = collections.OrderedDict()  # type: Dict[str, str]
+    for attr in attrs:
+        if attr.startswith("_"):
+            # Python conventional underscores are omitted in the dataframe.
+            df_column = attr[1:]
+        else:
+            df_column = attr
+        attrs_to_df_columns[attr] = df_column
+
+    # column_agg is an aggregator of column names.
+    # Keys of column agg are attributes of `FrozenTrial` such as 'trial_id' and 'params'.
+    # Values are dataframe columns such as ('trial_id', '') and ('params', 'n_layers').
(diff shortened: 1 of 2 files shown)


optuna/_dataframe.py
```diff
@@ -0,0 +1,83 @@
...
+with try_import() as _imports:
+    # `trials_dataframe` is disabled if pandas is not available.
+    import pandas as pd  # NOQA
+
+# Required for type annotation in `Study.trials_dataframe`.
+if not _imports.is_successful():
+    pd = object  # type: ignore # NOQA
+
+
+def _trials_dataframe(
```
hvy, line 22: The implementation of this function is just a copy from `Study.trials_dataframe`.

optuna/_dataframe.py
```diff
@@ -0,0 +1,83 @@
...
+from typing import Set
+from typing import Tuple
+
+import optuna
+from optuna._imports import try_import
+from optuna.trial._state import TrialState
+
+
+with try_import() as _imports:
+    # `trials_dataframe` is disabled if pandas is not available.
```
toshihikoyanase, line 14: A nit, but `trials_dataframe` seems to be either `Study.trials_dataframe` or `_trials_dataframe`.
```suggestion
    # `Study.trials_dataframe` is disabled if pandas is not available.
```

optuna/_dataframe.py
```diff
@@ -0,0 +1,83 @@
...
+# Required for type annotation in `Study.trials_dataframe`.
+if not _imports.is_successful():
+    pd = object  # type: ignore # NOQA
+
+
+def _trials_dataframe(
+    study: "optuna.Study",
+    attrs: Tuple[str, ...],
+    multi_index: bool,
+) -> "pd.DataFrame":
```
toshihikoyanase, line 26: Do we need the trailing comma? The number of arguments is not so large, and I'm not sure if it is needed here.
```suggestion
    study: "optuna.Study", attrs: Tuple[str, ...], multi_index: bool
) -> "pd.DataFrame":
```

toshihikoyanase (commented):
Thank you for your PR. It basically looks good to me, but I have some small comments.

optuna/_dataframe.py
hvy: You're right, let's remove the comma.

hvy:
Thanks for the review! Comments should be addressed now, PTAL.

toshihikoyanase (approved):

optuna/_dataframe.py
```diff
@@ -0,0 +1,81 @@
...
+from typing import Tuple
+
+import optuna
+from optuna._imports import try_import
+from optuna.trial._state import TrialState
+
+
+with try_import() as _imports:
+    # `Study.trials_dataframe` is disabled if pandas is not available.
+    import pandas as pd  # NOQA
```
himkt, line 15: [question]

Is `# NOQA` still needed here? (I'm wondering if we can remove it, as there is no variable assignment here.)

himkt (approved):
LGTM with a nits question.

optuna/_dataframe.py
hvy: I think you're right. Removed it.

hvy:
Removed the `#NOQA` as pointed out in review, PTAL.

PR #1910, opened 2020-10-10
Objective type

### LINKED ISSUE #1577: Objective function type annotation relaxation to `BaseTrial` from `Trial`
## Motivation

The objective function type hint says that its parameter should be of type `Trial`. However, https://github.com/optuna/optuna/pull/1503 will allow `FrozenTrial`s (which will derive from `BaseTrial` but not `Trial`) to be passed to the objective function. Type hints should reflect this change.

More precisely, the type annotation of [`ObjectiveFuncType`](https://github.com/optuna/optuna/blob/master/optuna/study.py#L42) should be updated and all definitions of objective functions in e.g. tests or examples.

~It's open for discussion given the fact that user objective function type hints will break.~ (Edited 2020-08-19)

pchelaa:
## Motivation

Closes #1577.

## Description of the changes

The type annotation of `ObjectiveFuncType` was updated to support `BaseTrial` and all related definitions were changed

### DIFF
--- a/examples/skorch_simple.py
+++ b/examples/skorch_simple.py
@@ -44,7 +44,7 @@ device = "cuda" if torch.cuda.is_available() else "cpu"
 
 
 class ClassifierModule(nn.Module):
-    def __init__(self, trial: optuna.Trial) -> None:
+    def __init__(self, trial: optuna.trial.BaseTrial) -> None:
         super().__init__()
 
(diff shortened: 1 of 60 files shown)


HideakiImamura:
Hi @pchelaa. This PR is draft one. What are the TODOs of this PR?

PR #1913, opened 2020-10-11
Pin `mypy==0.782`

himkt:
## Motivation

CI seems to fail with the latest `mypy` (v0.790).
(I tested the behavior on https://github.com/himkt/optuna/pull/4)

The following issue may be related to the CI failures.
https://github.com/python/mypy/issues/9122#issuecomment-691285527

## Description of the changes
This PR pins the version of `mypy==0.782`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ def get_tests_require() -> List[str]:
 def get_extras_require() -> Dict[str, List[str]]:
 
     requirements = {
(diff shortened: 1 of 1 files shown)


toshihikoyanase (approved):
Thank you. I reproduced the error and I think this change is required as a tentative fix.
LGTM.

hvy (approved):
LGTM!

PR #1918, opened 2020-10-12
Improve MLflow callback functionality: allow nesting, and attached study attrs.

drobison00:
This patch provides two optional flags for the MLflow integration callback.
- Allow users to specify whether or not to nest mlflow runs.
- Allow users to indicate that the Optuna's `study.user_attrs` key value pairs should be added as MLflow tags. 

## Motivation
This patch is meant to add additional flexibility/usability for users of the mlflow callback, without changing any of the current default behaviour. 

## Description of the changes
Trial nesting: It often makes sense to group related trial runs within the context of a given experiment, without having to create an entirely new experiment. mlflow's `nested` argument facilitates this, and improves readability in the mlflow ui. This patch adds an optional flag, which defaults to False, to the mlflow integration callback that indicates trial runs should be nested.

Study attrs: For a given study, we may have a number of user defined attributes that we'd like to attach as metadata (tags) to our MLflow trials. This patch adds an optional flag, which defaults to False, to the MLflow integration callback which indicates whether or not to attach the `study.user_attrs` as tags on the MLflow trial.

### DIFF
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -79,9 +79,9 @@ class MLflowCallback(object):
             Flag indicating whether or not trials should be logged as
             nested runs. This is often helpful for aggregating trials
             to a particular study, under a given experiment.
-        attach_study_attrs:
-            Flag indicating whether or not to attach the study's user attrs
-            to the mlflow trial.
+        tag_study_user_attrs:
+            Flag indicating whether or not to add the study's user attrs
+            to the mlflow trial as tags.
     """
 
     def __init__(
@@ -89,7 +89,7 @@ class MLflowCallback(object):
         tracking_uri: Optional[str] = None,
         metric_name: str = "value",
         nest_trials: Optional[bool] = False,
-        attach_study_attrs: Optional[bool] = False,
+        tag_study_user_attrs: Optional[bool] = False,
     ) -> None:
 
         _imports.check()
@@ -97,7 +97,7 @@ class MLflowCallback(object):
         self._tracking_uri = tracking_uri
         self._metric_name = metric_name
         self._nest_trials = nest_trials
-        self._attach_study_attrs = attach_study_attrs
+        self._tag_study_user_attrs = tag_study_user_attrs
 
     def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
 
@@ -133,7 +133,7 @@ class MLflowCallback(object):
             if isinstance(study_direction, StudyDirection):
                 tags["direction"] = str(study_direction).split(".")[-1]
 
-            if self._attach_study_attrs:
+            if self._tag_study_user_attrs:
                 tags.update(study.user_attrs)
             tags.update(trial.user_attrs)
             distributions = {
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -75,14 +75,29 @@ class MLflowCallback(object):
             Name of the metric. Since the metric itself is just a number,
             `metric_name` can be used to give it a name. So you know later
             if it was roc-auc or accuracy.
+        nest_trials:
+            Flag indicating whether or not trials should be logged as
+            nested runs. This is often helpful for aggregating trials
+            to a particular study, under a given experiment.
+        attach_study_attrs:
+            Flag indicating whether or not to attach the study's user attrs
+            to the mlflow trial.
     """
 
-    def __init__(self, tracking_uri: Optional[str] = None, metric_name: str = "value") -> None:
+    def __init__(
+        self,
+        tracking_uri: Optional[str] = None,
+        metric_name: str = "value",
+        nest_trials: Optional[bool] = False,
+        attach_study_attrs: Optional[bool] = False,
+    ) -> None:
 
         _imports.check()
 
         self._tracking_uri = tracking_uri
         self._metric_name = metric_name
+        self._nest_trials = nest_trials
+        self._attach_study_attrs = attach_study_attrs
 
     def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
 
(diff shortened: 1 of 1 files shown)


hvy:
Thanks for the interesting PR.

@harupy do you have any comments on this change?

The motivations sound reasonable but I'm wondering how this change suffices to handle nested runs. In the callback itself, the runs are never nested, is this something that the caller does? Sorry for my very shallow understanding of mlflow.

optuna/integration/mlflow.py
```diff
@@ -75,14 +75,29 @@ def objective(trial):
...
+            to the mlflow trial.
     """
 
-    def __init__(self, tracking_uri: Optional[str] = None, metric_name: str = "value") -> None:
+    def __init__(
+        self,
+        tracking_uri: Optional[str] = None,
+        metric_name: str = "value",
+        nest_trials: Optional[bool] = False,
+        attach_study_attrs: Optional[bool] = False,
```
hvy, line 92: More if a question but would this better communicate the intent (considering `mlflow.set_tags`)?
```suggestion
        tag_study_attrs: Optional[bool] = False,
```

drobison00: I don't have any issue updating to the tag prefix if that makes more sense.

With respect to nesting the runs, setting `nested=True` in the callback's `mlflow.start_run`, causes it to be nested within a parent run, so if you did something like the snippet below, all the Optuna trials would be nested under the 'Parent Run'.

```python
mlfcb = RAPIDSMLflowCallback(...., nest_trials=True)
study = optuna.create_study(....)
with mlflow.start_run(run_name=f"Parent Run"):
    study.optimize(...., callbacks=[mlfcb])
```

hvy: As for the nesting, makes a lot of sense. Thanks for the code snippet.

The parameter name is also a matter of preference. Unless "attach" is commonly used in mlflow, I'd stick to this suggestion. However, I'm very open to suggestions from especially mlflow users.

drobison00:
@hvy Updated to the suggesting naming convention.

optuna/integration/mlflow.py
```diff
@@ -75,14 +75,29 @@ def objective(trial):
...
+            to the mlflow trial as tags.
     """
 
-    def __init__(self, tracking_uri: Optional[str] = None, metric_name: str = "value") -> None:
+    def __init__(
+        self,
+        tracking_uri: Optional[str] = None,
+        metric_name: str = "value",
+        nest_trials: Optional[bool] = False,
+        tag_study_user_attrs: Optional[bool] = False,
```
hvy, line 92: We're not handling `None` as it seems now. Actually, it can be a mandatory parameter, just that it has a default value.
```suggestion
        tag_study_user_attrs: bool = False,
```

optuna/integration/mlflow.py
```diff
@@ -75,14 +75,29 @@ def objective(trial):
...
+            Flag indicating whether or not to add the study's user attrs
+            to the mlflow trial as tags.
     """
 
-    def __init__(self, tracking_uri: Optional[str] = None, metric_name: str = "value") -> None:
+    def __init__(
+        self,
+        tracking_uri: Optional[str] = None,
+        metric_name: str = "value",
+        nest_trials: Optional[bool] = False,
```
hvy, line 91: Ditto.
```suggestion
        nest_trials: bool = False,
```

hvy (changes requested):
Thanks for your quick actions. LGTM besides those last two comments.

I by the way noticed that we should handle cases, e.g. raise warnings or prefix keys, when there are duplicates among the keys in the `tags` dictionary. This might be outside the scope of this PR though.

harupy:
I'm wondering which approach makes more senes.

- Option 1: the current approach where `MlflowCallback` doesn't create a parent run.
- Option 2: `MlflowCallback` automatically creates a parent run if a user set `nest_trials` to `True`.

drobison00:
@harupy I definitely would prefer to not have Optuna create the parent run or be responsible for it.

As an example, if I want to embed my Optuna runs into a broader process that creates and registers a model under the parent, based on Optuna's results, that doesn't seem like a concern that is relevant to Optuna to manage, and ends up over-complicating whats going on.

drobison00:
@hvy 

While its true that someone could cause themselves grief by setting user_attrs that clobber some existing tag, at that point, they've set the clobbering attribute on the study, and a flag that indicates they want those values to be tagged on the run. What do you think about adding a note to the argument description that says something to the effect of 'tags in study.user_attrs will supersede existing tags'?

harupy:
> As an example, if I want to embed my Optuna runs into a broader process that creates and registers a model under the parent, based on Optuna's results, that doesn't seem like a concern that is relevant to Optuna to manage, and ends up over-complicating whats going on.

@drobison00 That sounds reasonable :)

hvy:
> What do you think about adding a note to the argument description that says something to the effect of 'tags in study.user_attrs will supersede existing tags'?

Sounds reasonable as well given the fact that this is an "opt-in" feature as you say. Would you mind adding this note in this PR?

hvy:
Oh, and @harupy thanks a bunch your feedback.

optuna/integration/mlflow.py
```diff
@@ -75,14 +75,30 @@ def objective(trial):
...
             `metric_name` can be used to give it a name. So you know later
             if it was roc-auc or accuracy.
+        nest_trials:
+            Flag indicating whether or not trials should be logged as
+            nested runs. This is often helpful for aggregating trials
+            to a particular study, under a given experiment.
+        tag_study_user_attrs:
+            Flag indicating whether or not to add the study's user attrs
+            to the mlflow trial as tags. Please note that when this flag is
+            set, key value pairs in study.user_attrs will supersede existing tags.
```
hvy, line 85: Okay, so the logic is actually that the study user attributes are superseded too, so the explanation is somewhat misleading. I think we should either update the docs, or the logic (I'm fine with either).

hvy (approved):
LGTM!

harupy (approved):
LGTM :)

hvy:
Again, thanks @harupy for the review and @drobison00 for the improvement 🙂

PR #1919, opened 2020-10-13
Fix RDB test to avoid deadlock when creating study

toshihikoyanase:
## Motivation
CI sometimes failed at the following line:

https://github.com/optuna/optuna/blob/3fae91c3c9f48f9111c590143da472bd3883c887/tests/storages_tests/rdb_tests/test_with_server.py#L109

In this line, multiple workers try to create studies, and the deadlock occurs depending on the timing.

https://github.com/optuna/optuna/blob/3fae91c3c9f48f9111c590143da472bd3883c887/tests/storages_tests/rdb_tests/test_with_server.py#L33

## Description of the changes

I think we can fix `optuna.create_study` to avoid deadlock, but this PR only updates the test case.
It creates a study in a single (parent) process and shares it with workers.

### DIFF
--- a/tests/storages_tests/rdb_tests/test_with_server.py
+++ b/tests/storages_tests/rdb_tests/test_with_server.py
@@ -30,7 +30,7 @@ def run_optimize(args: Tuple[str, str]) -> None:
     study_name = args[0]
     storage_url = args[1]
     # Create a study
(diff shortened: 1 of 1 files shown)


hvy (approved):
Thanks, LGTM!!

himkt (approved):
LGTM!

hvy:
2 approvals so let me merge the changes.

PR #1781, opened 2020-09-03
Fix exception chaining all over the codebase

akihironitta:
## Motivation
Fixes #1779.

## Description of the changes
I used either:
- `raise .. from e` to chain exceptions where the first exception `e` has valuable information and seems useful for users.
- `raise .. from None` to unchain exceptions where the first exception has no valuable information and seems noisy and redundant.

## List of lines to check
I wrote [a simple script](https://gist.github.com/akihironitta/4223c1b32404b36c1b349d70c4c93b4d) using ast, a Python standard library, to list all the lines containing `raise` in `except` clause. (It doesn't necessarily mean that all these lines have to be changed.)

- [x] https://github.com/optuna/optuna/blob/fe1e3e54f961897b871daa7fe23d92c4e1eaa408/optuna/progress_bar.py#L21
- [x] https://github.com/optuna/optuna/blob/fe1e3e54f961897b871daa7fe23d92c4e1eaa408/optuna/integration/sklearn.py#L119
- [x] https://github.com/optuna/optuna/blob/fe1e3e54f961897b871daa7fe23d92c4e1eaa408/optuna/integration/sklearn.py#L315
- [x] https://github.com/optuna/optuna/blob/fe1e3e54f961897b871daa7fe23d92c4e1eaa408/optuna/integration/sklearn.py#L326
- [x] https://github.com/optuna/optuna/blob/fe1e3e54f961897b871daa7fe23d92c4e1eaa408/optuna/integration/chainer.py#L79
- [x] https://github.com/optuna/optuna/blob/fe1e3e54f961897b871daa7fe23d92c4e1eaa408/optuna/integration/chainermn.py#L321
- [x] https://github.com/optuna/optuna/blob/fe1e3e54f961897b871daa7fe23d92c4e1eaa408/optuna/storages/_rdb/storage.py#L112
- [x] https://github.com/optuna/optuna/blob/fe1e3e54f961897b871daa7fe23d92c4e1eaa408/optuna/storages/_rdb/storage.py#L140
- [x] https://github.com/optuna/optuna/blob/fe1e3e54f961897b871daa7fe23d92c4e1eaa408/optuna/storages/_rdb/storage.py#L474
- [x] https://github.com/optuna/optuna/blob/fe1e3e54f961897b871daa7fe23d92c4e1eaa408/optuna/storages/_rdb/storage.py#L1113
[...]

PR #1921, opened 2020-10-13
Fix duplicate stack traces

akihironitta:
## Motivation
There is currently a duplicate traceback in the error message when `StorageInternalError` is raised at: https://github.com/optuna/optuna/blob/e857b09ff65e8ba32b21023318345d08f4c4c830/optuna/storages/_rdb/storage.py#L1099

<details><summary>Example</summary><p>
Below is a brief version of the above. Run:

```python
import sys
from sqlalchemy.exc import SQLAlchemyError
try:
    raise SQLAlchemyError("Unexpected SQLAlchemyError occurs here!")
except SQLAlchemyError as e:
    message = (
        "An exception is raised during the commit. "
        "This typically happens due to invalid data in the commit, "
        "e.g. exceeding max length."
        )
    raise Exception(message).with_traceback(
        sys.exc_info()[2]
        ) from e
```
, and you'll get:
```shell
  File "/tmp/test.py", line 4, in <module>
    raise SQLAlchemyError("Unexpected SQLAlchemyError occurs here!")
sqlalchemy.exc.SQLAlchemyError: Unexpected SQLAlchemyError occurs here!

The above exception was the direct cause of the following exception:

  File "/tmp/test.py", line 13, in <module>
    ) from e
  File "/tmp/test.py", line 4, in <module>
    raise SQLAlchemyError("Unexpected SQLAlchemyError occurs here!")
Exception: An exception is raised during the commit. This typically happens due to invalid data in the commit, e.g. exceeding max length.
```

Note that there are duplicate lines in the error message:
```shell
  File "/tmp/test.py", line 4, in <module>
    raise SQLAlchemyError("Unexpected SQLAlchemyError occurs here!")
```

</p></details>

## Description of the changes
Removes `with_traceback()` to get rid of the duplicate.

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -3,7 +3,6 @@ from datetime import datetime
 import json
 import logging
 import os
-import sys
 from typing import Any
 from typing import Dict
 from typing import List
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -1096,9 +1096,7 @@ class RDBStorage(BaseStorage):
                 "This typically happens due to invalid data in the commit, "
                 "e.g. exceeding max length."
             )
-            raise optuna.exceptions.StorageInternalError(message).with_traceback(
-                sys.exc_info()[2]
-            ) from e
+            raise optuna.exceptions.StorageInternalError(message) from e
 
     def remove_session(self) -> None:
         """Removes the current session.


hvy (approved):
Thanks a lot, I should've noticed it when going through https://github.com/optuna/optuna/pull/1781. LGTM!

akihironitta:
@hvy No worries, **I** should've been more careful. Thank you for the review!

PR #1927, opened 2020-10-14
Fixed 'Mean of empty slice' warning

carefree0910:
## Motivation

When using `optuna`'s pruning strategy with `n_jobs > 1`, the following warning happens to occur frequently:

```python
...\site-packages\numpy\lib\nanfunctions.py:1366: RuntimeWarning:

Mean of empty slice
```

## Description of the changes

It's easy to be fixed with a simple condition statement. I've realized that you are using `black` to format your codes, so I've tried my best to follow it 😉

> BTW: I'm not sure whether you prefer returning `nan` of `inf` in such condition. I've chosen `inf` but feel free to change it to `nan` if you like 😆

### DIFF
--- a/optuna/pruners/_percentile.py
+++ b/optuna/pruners/_percentile.py
@@ -33,19 +33,22 @@ def _get_percentile_intermediate_result_over_trials(
     if len(completed_trials) == 0:
         raise ValueError("No trials have been completed.")
 
-    if direction == StudyDirection.MAXIMIZE:
+    intermediate_values = [
+        t.intermediate_values[step]
+        for t in completed_trials
+        if step in t.intermediate_values
+    ]
+    is_maximize = direction == StudyDirection.MAXIMIZE
+
+    if not intermediate_values:
+        return -math.inf if is_maximize else math.inf
+
+    if is_maximize:
         percentile = 100 - percentile
 
     return float(
         np.nanpercentile(
-            np.array(
-                [
-                    t.intermediate_values[step]
-                    for t in completed_trials
-                    if step in t.intermediate_values
-                ],
-                np.float,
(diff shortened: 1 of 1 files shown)


hvy:
Thanks a lot for the PR, this indeed seems like a bug. 

We cannot in my understanding determine if the trial should be pruned if records are missing, so we should return `False` from `should_prune`. I think doing an early return with nan will simplify the code/make it less error prone.

carefree0910:
Not sure if it requires me to edit it but I've made a commit anyway 🤣

optuna/pruners/_percentile.py
```diff
@@ -33,19 +33,20 @@ def _get_percentile_intermediate_result_over_trials(
     if len(completed_trials) == 0:
         raise ValueError("No trials have been completed.")
 
-    if direction == StudyDirection.MAXIMIZE:
+    intermediate_values = [
+        t.intermediate_values[step] for t in completed_trials if step in t.intermediate_values
+    ]
+    is_maximize = direction == StudyDirection.MAXIMIZE
```
hvy, line 39: This is a small detail but this local variable doesn't really improve the code quality and it's somewhat unrelated to the fix. Might it be sensible to revert this change?

hvy (commented):
Thanks again, the changes basically LGTM. I just left one nitpick.

optuna/pruners/_percentile.py
carefree0910: Sure! I forgot to revert this after I returned `math.nan` 🤣

hvy (approved):
LGTM!

HideakiImamura (approved):

PR #1929, opened 2020-10-14
Fixed `plot_contour`

carefree0910:
## Motivation
When mixture types were used as `categorical`, they could not compare to each other, which would raise errors.

## Description of the changes
I fixed it by converting them to `str` but this solution is nowhere near elegant 😕

### DIFF
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -72,7 +72,7 @@ def plot_contour(study: Study, params: Optional[List[str]] = None) -> "go.Figure
     return _get_contour_plot(study, params)
 
 
-def _get_param_values(trials: List[FrozenTrial], p_name: str) -> List[Union[str, Any]]:
+def _get_param_values(trials: List[FrozenTrial], p_name: str) -> List[Union[str, float, Any]]:
     values = [t.params[p_name] for t in trials if p_name in t.params]
     if not _is_categorical(trials, p_name):
         return values
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -3,6 +3,8 @@ from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Tuple
+from typing import Union
+from typing import Any
 
 from optuna._study_direction import StudyDirection
 from optuna.logging import get_logger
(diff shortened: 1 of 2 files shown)


crcrpar:
Could you provide us with minimal reproducible code or a test case?

carefree0910:
> Could you provide us with minimal reproducible code or a test case?

Sure!

```python
import optuna


def objective(trial: optuna.trial.Trial) -> float:
    foo = trial.suggest_categorical("foo", [None, 0])
    if foo is None:
        return 0.0
    return 1.0


study = optuna.create_study(study_name="foo")
optuna.visualization.plot_contour(study)
```

crcrpar:
TL; DR: Ah, that pattern may or may not be a corner case though should be supported in the short term.

Could you fix the mypy error?

---

Some personal take on this pull request.

The documentation says the type of `choices` is `Sequence[Union[None, bool, int, float, str]]` as you can see in https://optuna.readthedocs.io/en/stable/reference/generated/optuna.trial.Trial.html?highlight=support_categorical#optuna.trial.Trial.suggest_categorical.

In my understanding, this means that all the elements of `choices` should have the same type.
BUT, as the problem you encounter shows, we have not confirmed this assumption.

So, your patch completely makes sense, though alternatively, we can narrow the supported type of `choices` in runtime in order to avoid the problem you saw. By this, I mean, add the change to `suggest_categorical` to check that `choices` have the same type.

Anyway, my take is as follows. I reckon that idealy we should add the type check of `choices` in runtime to `suggest_categorical`. But, since this change might NOT be backward compatible, I'd like to move this pull request forward.

carefree0910:
Thanks for the detailed reply! I'll try to fix the mypy error, but I'm not familiar with it so I may need to struggle a little bit 🤣

---

And here are some of my thoughts:
1) Narrow down the supported type might dramatically reduce the flexibility of `optuna` 😢.
2) In this case it's weird to see `None` to be one of the supported types, since there could only have one `None` after all 🤣.

crcrpar (approved):

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM. A nit, but should we add the test case when the categorical parameter is used? @carefree0910 What do you think?

carefree0910:
I definitely agree! Just wondering if the test case above is enough and where should I add it to (maybe [here](https://github.com/optuna/optuna/blob/master/tests/visualization_tests/test_contour.py)?) 🤣

HideakiImamura:
@carefree0910 Yes, I think `tests/visualization_tests/test_coutour.py` is fine. Could you add a test?

carefree0910:
Sure!

HideakiImamura (approved):
Thanks for the update! LGTM!

PR #1931, opened 2020-10-14
Fix suggest_int method when using a large number

s2terminal:
## Motivation
I want to use a large number, for example a factorial of 50, as the upper endpoint of the range of suggested value of the `suggest_int` method. But I got the following error.

```
optuna/samplers/_random.py in sample_independent(self, study, trial, param_name, param_distribution)
     86             r = (param_distribution.high - param_distribution.low) / param_distribution.step
     87             # numpy.random.randint includes low but excludes high.
---> 88             s = self._rng.randint(0, r + 1)
     89             v = s * param_distribution.step + param_distribution.low
     90             return int(v)

mtrand.pyx in numpy.random.mtrand.RandomState.randint()

_bounded_integers.pyx in numpy.random._bounded_integers._rand_int64()

ValueError: high is out of bounds for int64
```

## Description of the changes
- Changed standard library instead of numpy for random numbers.
- Changed the result of the division from a float to an int.
- Changed the precision of decimal.
- Add test.

I am not confident in this change. Sorry if I have made the wrong changes.

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -508,6 +508,7 @@ def _adjust_discrete_uniform_high(low: float, high: float, q: float) -> float:
     d_q = decimal.Decimal(str(q))
 
     d_r = d_high - d_low
+    decimal.getcontext().prec = 10000
 
     if d_r % d_q != decimal.Decimal("0"):
         old_high = high
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -508,9 +508,8 @@ def _adjust_discrete_uniform_high(low: float, high: float, q: float) -> float:
     d_q = decimal.Decimal(str(q))
 
     d_r = d_high - d_low
-    decimal.getcontext().prec = 10000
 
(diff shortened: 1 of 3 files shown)


nzw0301:
In my opinion, many machine learning libraries assume that their hyper-parameter is in either `int64` or `int32` because they use numpy internally. So this change sounds a little bit problem-specific. I suggest you sample a float value from uniform distribution at first, then you convert it to a large integer.

s2terminal:
>sample a float value from uniform distribution at first, then you convert it to a large integer.

Thanks for the suggestion. In my case, that method seems to work well.

HideakiImamura:
@s2terminal Thank you for the PR. @nzw0301 Thank you for providing a great solution. I think the method proposed by @nzw0301 works well, so this PR can be closed. @s2terminal What do you think?

s2terminal:
The idea given to me by @nzw0301 made it work. Thank you. I'm closing this issue.

PR #1932, opened 2020-10-15
Add a test to verify `nest_trials` for `MLflowCallback` works properly.

harupy:
## Motivation

A follow-up for #1918 

## Description of the changes

Add a test to verify `nest_trials` for `MLflowCallback` works properly.

### DIFF
--- a/tests/integration_tests/test_mlflow.py
+++ b/tests/integration_tests/test_mlflow.py
@@ -165,7 +165,9 @@ def test_nest_trials(tmpdir: py.path.local) -> None:
     assert all(set(r.data.metrics.keys()) == {"value"} for r in child_runs)
 
 
-def test_mlflow_callback_fails_when_nest_trials_is_false_and_active_run_exists(tmpdir):
+def test_mlflow_callback_fails_when_nest_trials_is_false_and_active_run_exists(
+    tmpdir: py.path.local,
+) -> None:
     tmp_tracking_uri = "file:{}".format(tmpdir)
 
     study_name = "my_study"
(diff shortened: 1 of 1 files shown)


tests/integration_tests/test_mlflow.py
```diff
@@ -132,3 +134,31 @@ def test_tag_truncation(tmpdir: py.path.local) -> None:
...
+
+
+def test_nest_trials(tmpdir: py.path.local) -> None:
+    tmp_tracking_uri = "file:{}".format(tmpdir)
+
+    study_name = "my_study"
+    mlflow.set_tracking_uri(tmp_tracking_uri)
+    mlflow.set_experiment(study_name)
+
+    mlflc = MLflowCallback(tracking_uri=tmp_tracking_uri, nest_trials=True)
```
hvy, line 146: We might want to check the negative case where the parameter is `False`.

hvy (commented):
Thanks as always. Again, left a nitpick.

Maybe @drobison00 would like to have to a look/review this PR?

tests/integration_tests/test_mlflow.py
harupy: If we set `nest_trials` to `False` and an active run exists, MLflow would raise the following error:

```
Exception: Run with UUID aa67b134da7c43158015648fc00ffdb4 is already active.
To start a new run, first end the current run with mlflow.end_run().
To start a nested run call start_run with nested=True
```

Should I add a test to verify this behavior?

hvy: Yup, that' what I meant. Sorry I was probably not clear enough.

hvy (approved):
LGTM!

drobison00:
@hvy Sorry, was out for a few days. LGTM.

hvy:
That's great, let me have this PR merged then!

hvy:
@harupy by the way, you can omit the punctuation from the PR title now. We slightly changed the rules in https://github.com/optuna/optuna/blob/master/CONTRIBUTING.md#title (it used to require a punctuation at the end of the title but not anymore)

Issue #1897, opened 2020-10-05
Use variable annotation syntax for Type Hints


c-bata:
## Motivation

After dropping Python 3.5 (#1067), we can use type hints with variable annotation syntax.

https://www.python.org/dev/peps/pep-0526/

## Description

Thank you @harupy for creating a script to detect type comments for variables!
https://github.com/optuna/optuna/issues/711#issuecomment-683521987

harupy:
I have tested [com2ann](https://github.com/ilevkivskyi/com2ann) (which translates type comments into type annotations) and it worked.

Usage:

```bash
# install python 3.8 (com2ann doesn't work in python < 3.8)
pip install git+https://github.com/ilevkivskyi/com2ann.git

com2ann .
```

hvy:
For the record, https://github.com/optuna/optuna/pull/1937/files#r509361516.

c-bata:
I confirmed that all comment-style variable annotation are replaced at #1950.

PR #1937, opened 2020-10-15
Add variable annotations to `optuna/*.py`

akihironitta:
## Motivation
See #1897.

## Description of the changes
Adds variable annoations to:
- optuna/_imports.py
- optuna/dashboard.py
- optuna/progress_bar.py
- optuna/study.py

### DIFF
--- a/optuna/dashboard.py
+++ b/optuna/dashboard.py
@@ -38,8 +38,8 @@ with try_import() as _imports:
         )
 
 
-_mode = None  # type: Optional[str]
-_study = None  # type: Optional[optuna.study.Study]
+_mode: Optional[str] = None
+_study: Optional[optuna.study.Study] = None
 
 _HEADER_FORMAT = """
 <style>
@@ -99,7 +99,7 @@ if _imports.is_successful():
 
         def update(self, new_trials: List[optuna.trial.FrozenTrial]) -> None:
 
-            stream_dict = collections.defaultdict(list)  # type: Dict[str, List[Any]]
+            stream_dict: Dict[str, List[Any]] = collections.defaultdict(list)
 
             for trial in new_trials:
                 if trial.state != optuna.trial.TrialState.COMPLETE:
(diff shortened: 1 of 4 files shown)


akihironitta:
@HideakiImamura Do we need type annotations for temporary local variables as well?

For example, `need_to_add_callback` in this case:
https://github.com/optuna/optuna/blob/5940fd75059c469a729ec66ee84925f8c2e12ed7/optuna/dashboard.py#L230-L232

crcrpar:
`need_to_add_callback` don't need to get annotated as it's const-ish variable and its type is explicit enough to me 😉

optuna/dashboard.py
```diff
@@ -57,7 +58,7 @@
 </p>
 """
 
-_DATETIME_FORMAT = "%Y-%m-%d %H:%M:%S"
+_DATETIME_FORMAT: str = "%Y-%m-%d %H:%M:%S"
```
hvy, line 61: I'm wondering if we should avoid introducing type hints to variables that didn't used to have comment style type hints. We usually only add type hints where mypy would complain assuming that it's clever to detect possible bug-prone code (it get a bit verbose and may hurt readability too adding type hints). It also seems a bit orthogonal to the motivation behind #1897 but sorry, this was probably not communicated well enough. What do you think?

akihironitta: (Although I have added a lot of annotations already,) I have to admit that I strongly agree with you that too many annotations lower readability, and it sounds reasonable to add variable annotations only to variables which already have comment style annotations. I'll roll back some commits if you're ok with it!

akihironitta: If we add anntations only to variables which already have comment style annotations, we can just use `com2ann` suggested by @harupy in the issue all at once!

hvy: Sure, that might be better using `com2ann`. Do you want to close this PR and create a new one?

akihironitta: Sure.

hvy: Looking forward, thanks!

PR #1943, opened 2020-10-18
Allow filtering trials from `Study` and `BaseStorage` based on `TrialState`

hvy:
## Motivation

It is common to filter trials based on state (`TrialState`). For instance, in several samplers, visualization functions, hyperparameter importance helper functions, we do that after querying all trials from `Study.get_trials` or `BaseStorage.get_all_trials`. Introducing a state parameter to these methods would simplify the code, make it less error prone, and allow for optimization on the storage layer. 

## Description of the changes

Introduces an optional state (`state`) parameter to `Study.get_trials` and `BaseStorage.get_all_trials`. Also, demonstrates how this parameter should be used by simplifying the code in the `TPESampler`.

## Note

- Changes might be controversial in the sense that user might want to filter on other attributes than the state as well
- `BaseStorage.get_n_trials` already had a `state` parameter
- Could break for users inheriting `Study` and/or `BaseStorage`

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -527,7 +527,7 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         self,
         study_id: int,
         deepcopy: bool = True,
-        state: Optional[Union[Tuple[TrialState, ...], TrialState]] = None,
+        states: Optional[Tuple[TrialState, ...]] = None,
     ) -> List[FrozenTrial]:
         """Read all trials in a study.
 
@@ -537,7 +537,7 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
             deepcopy:
                 Whether to copy the list of trials before returning.
                 Set to :obj:`True` if you intend to update the list or elements of the list.
-            state:
+            states:
                 Trial states to filter on. If :obj:`None`, include all states.
 
         Returns:
@@ -567,7 +567,11 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
             :exc:`KeyError`:
                 If no study with the matching ``study_id`` exists.
         """
-        return len(self.get_all_trials(study_id, deepcopy=False, state=state))
+        # TODO(hvy): Align the name and the behavior or the `state` parameter with
+        # `get_all_trials`'s `states`.
+        if isinstance(state, TrialState):
+            state = (state,)
+        return len(self.get_all_trials(study_id, deepcopy=False, states=state))
 
     def get_best_trial(self, study_id: int) -> FrozenTrial:
         """Return the trial with the best value in a study.
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -558,7 +558,7 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
             study_id:
                 ID of the study.
             state:
-                :class:`~optuna.trial.TrialState` to filter trials.
+                Trial states to filter on. If :obj:`None`, include all states.
 
         Returns:
             Number of trials in the study.
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -3,6 +3,8 @@ from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import Tuple
+from typing import Union
 
 from optuna._study_direction import StudyDirection
 from optuna._study_summary import StudySummary
@@ -521,7 +523,12 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         raise NotImplementedError
 
     @abc.abstractmethod
-    def get_all_trials(self, study_id: int, deepcopy: bool = True) -> List[FrozenTrial]:
+    def get_all_trials(
+        self,
+        study_id: int,
+        deepcopy: bool = True,
+        state: Optional[Union[Tuple[TrialState, ...], TrialState]] = None,
+    ) -> List[FrozenTrial]:
         """Read all trials in a study.
 
         Args:
@@ -530,6 +537,8 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
             deepcopy:
                 Whether to copy the list of trials before returning.
                 Set to :obj:`True` if you intend to update the list or elements of the list.
+            state:
+                Trial states to filter on. If :obj:`None`, include all states.
 
         Returns:
             List of trials in the study.
@@ -540,7 +549,9 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         """
         raise NotImplementedError
 
-    def get_n_trials(self, study_id: int, state: Optional[TrialState] = None) -> int:
+    def get_n_trials(
+        self, study_id: int, state: Optional[Union[Tuple[TrialState, ...], TrialState]] = None
+    ) -> int:
         """Count the number of trials in a study.
 
         Args:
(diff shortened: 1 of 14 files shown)


optuna/storages/_base.py
```diff
@@ -540,7 +549,9 @@ def get_all_trials(self, study_id: int, deepcopy: bool = True) -> List[FrozenTri
         """
         raise NotImplementedError
 
-    def get_n_trials(self, study_id: int, state: Optional[TrialState] = None) -> int:
+    def get_n_trials(
+        self, study_id: int, state: Optional[Union[Tuple[TrialState, ...], TrialState]] = None
+    ) -> int:
         """Count the number of trials in a study.
```
HideakiImamura, line 555: I think it is better to change the explanation of the `state` argument to match its type annotation `Optional[Union[Tuple[TrialState, ...], TrialState]]`.

optuna/storages/_in_memory.py
```diff
@@ -378,25 +380,31 @@ def _set_trial(self, trial_id: int, trial: FrozenTrial) -> None:
...
             self._check_study_id(study_id)
-            if deepcopy:
-                return copy.deepcopy(self._studies[study_id].trials)
-            else:
-                return self._studies[study_id].trials[:]
 
-    def get_n_trials(self, study_id: int, state: Optional[TrialState] = None) -> int:
+            trials = self._studies[
+                study_id
+            ].trials  # type: Union[List[FrozenTrial], Iterator[FrozenTrial]]
```
HideakiImamura, line 395: I'm not sure when we should use variable type annotation, but we can do here.

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -330,43 +330,46 @@ def int_value_fn(idx: int) -> float:
...
 ) -> None:
-    study = optuna.create_study()
     dist = optuna.distributions.UniformDistribution(1.0, 100.0)
 
     # Prepare sampling result for later tests.
-    past_trials = [frozen_trial_factory(i, dist=dist) for i in range(1, 100)]
+    study = optuna.create_study()
+    for i in range(1, 100):
+        trial = frozen_trial_factory(i, dist=dist)
+        study._storage.create_new_trial(study._study_id, template_trial=trial)
```
HideakiImamura, line 348: Looks very good to me to use `create_new_trial`. Thanks!

HideakiImamura (commented):
Thanks for the PR. The introduced feature looks good to me. I have some minor comments. Could you take a look?

optuna/storages/_base.py
hvy: You're right, I updated the docs. Thanks!

optuna/storages/_in_memory.py
hvy: Seemingly because `trials` is reassigned with a different type. I by the way don't think introducing new local variables will reduce the complexity here so I'd rather keep it as is.

tests/samplers_tests/tpe_tests/test_sampler.py
hvy: We can do the same thing for other tests in this file but maybe in a separate PR.

hvy:
Thanks @HideakiImamura for your review. The comments should be addressed so PTAL.

HideakiImamura (approved):
Thanks for the update! LGTM.

hvy:
Thanks for the review. Just noticed a conflict so I merged with master.

tests/storages_tests/test_storages.py
```diff
@@ -821,6 +821,51 @@ def test_get_all_trials_deepcopy_option(storage_mode: str) -> None:
...
+    with StorageSupplier(storage_mode) as storage:
+        study_id = storage.create_new_study()
+        storage.set_study_direction(study_id, StudyDirection.MAXIMIZE)
+        generator = random.Random(51)
+
+        states = [
+            TrialState.COMPLETE,
+            TrialState.COMPLETE,
+            TrialState.PRUNED,
+        ]
```
crcrpar, line 836: Super nit and needn't be addressed: Isn't this be better a tuple?

optuna/storages/_base.py
```diff
@@ -521,7 +523,12 @@ def get_trial(self, trial_id: int) -> FrozenTrial:
         raise NotImplementedError
 
     @abc.abstractmethod
-    def get_all_trials(self, study_id: int, deepcopy: bool = True) -> List[FrozenTrial]:
+    def get_all_trials(
+        self,
+        study_id: int,
+        deepcopy: bool = True,
+        state: Optional[Union[Tuple[TrialState, ...], TrialState]] = None,
```
crcrpar, line 530: [Question] Do we need to accept `TrialState`? Personally tuple of `TrialState`s is enough.

tests/storages_tests/test_storages.py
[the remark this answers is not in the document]
hvy: I don't think it matters here but maybe I'm missing something. How come it be?

optuna/storages/_base.py
hvy: Good point. Curious but do you think we should do that for `Study.get_trials` as well? I tried to make `Study.get_trials` as "user-friendly" as possible and propagated this design down to the storage. But this interface might be an anti-pattern in the first place. I'd go with a `Sequence` and not a `Tuple` to be exact in that case, and name the argument `states`.

tests/storages_tests/test_storages.py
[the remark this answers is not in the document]
crcrpar: Yes, it doesn't matter at all and this is why I said super nit and needn't be addressed. I just prefer tuple as I expected `states` to be fixed.

optuna/storages/_base.py
crcrpar: Elaboration: I believe that users of this new argument peruse our documentation and/or catch up with Optuna well; thus, I think it's acceptable to make `state` to be a container even if users pass only one `TrialState`. By restricting `state` to be a container, developers and others who read Optuna codebase will enjoy some benefits, including easier maintenance and clearer logic.

optuna/study.py
```diff
@@ -125,13 +129,15 @@ def objective(trial):
                 Note that if you set the flag to :obj:`False`, you shouldn't mutate
                 any fields of the returned trial. Otherwise the internal state of
                 the study may corrupt and unexpected behavior may happen.
+            state:
+                Trial states to filter on. If :obj:`None`, include all states.
```
crcrpar, line 133: Following the discussion about `state`'s type. I'd like you to update this to mention that `state` (may be `states`) is required to be a container.

hvy: Isn't it clear from the type?

tests/storages_tests/test_storages.py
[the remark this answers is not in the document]
hvy: Fixed!

optuna/storages/_base.py
hvy: Fixed!

hvy:
Now only allowing tuples of states, PTAL.

optuna/study.py
crcrpar: Absolutely.

crcrpar (commented):
Is [this line](https://github.com/optuna/optuna/pull/1943/files#diff-c9e05ff7d5e7f6e91cf177e165c727d8731cf219cc7f31a3bee9f4ff3c3d3de8R107-R108) up-to-date? I think now `get_trials` is handy enough thanks to `states`.

> For library users, it's recommended to use more handy :attr:`~optuna.study.Study.trials` property to get the trials instead.

hvy:
I agree that that sentence is a bit vague and confusing. Should we leave a seealso or something to the trials attribute instead, or just remove it completely?

crcrpar:
Removing sounds good to me as https://optuna.readthedocs.io/en/latest/reference/generated/optuna.study.Study.html?highlight=get_trials#optuna.study.Study.trials says "I'm alias of `get_trials(deepcopy=True)`".

hvy:
Removed the documentation, PTAL.

hvy:
PTAL. Noticed that the MO study were forgotten so fixed it as well. There were no tests as far as I could find so only changed the logic but let me know if you want me to add tests. It might be fine as is since it's experimental and we're undergoing some major refactoring of the MO submodule.

crcrpar (approved):
Thanks for updating details as well.

crcrpar:
Okay, let's do that after multiobjective refactoring.

Issue #1858, opened 2020-09-24
Metric MAPE direction should be reversed with reference to MAP


dileepbapat:
https://github.com/optuna/optuna/blob/9fc26c2277b347788d5554dad66dba81585716b2/optuna/integration/_lightgbm_tuner/optimize.py#L146

toshihikoyanase:
Thank you for your question. `MAPE` stands for the mean absolute percentage error, and I think it is to be minimized  as can be seen in the LightGBM code:

https://github.com/microsoft/LightGBM/blob/2870490073c537eeb3648893257be8692b63148c/R-package/R/metrics.R#L13

>.METRICS_HIGHER_BETTER <- function() {
> ...
>         , "mape" = FALSE

Please let me know if I miss your point.

di1:
Yes it is there in LGBM, however integration in optuna/optuna/integration/_lightgbm_tuner/optimize.py (146) has starts with in the if preventing mape based optimization. It tries to increase error when we optimize using tuner.

toshihikoyanase:
@di1 Sorry for the delayed response.

> It tries to increase error when we optimize using tuner.

Could you share the log messages and reproducible code? I think the `LightGBMTuner` tried to minimize the error, but it failed for some reason.

toshihikoyanase:
@dileepbapat @di1 
I'm sorry but I didn't get the point. Now, I find the cause and I submitted #1966 to fix this issue.
If you have any comments, please let me know.

Issue #1636, opened 2020-08-14
How to optimize on feval when using LightGBMTunerCV


rafmacalaba:
Hello,

I have a custom feval function for lightgbm which is basically f1_score, how do I optimize on that instead of the default binary_logloss metric? Thanks for this library!

toshihikoyanase:
Hello, @rafmacalaba
Thank you for your question. Currently, `LightGBMTunerCV` does not support custom metrics as reported in #1351. #1522 tried to resolve the issue, but it has been suspended. If you are interested in taking over the development, please let us know.

rafmacalaba:
@toshihikoyanase got it, I am referring to `feval` optimization not the `metric`. I have an idea how to get it done. Expect a PR from me regarding this. Thanks!

toshihikoyanase:
@rafmacalaba Thank you! I'm looking forward to your PR.

Issue #1632, opened 2020-08-13
[RFC] Customizable steps of `LightGBMTuner` and `LightGBMTunerCV`


toshihikoyanase:
## Motivation
The steps of `LightGBMTuner` are fixed, and users cannot add the parameters to tune. I saw users' demand to customize the steps.

For example,
- @Amyantis submitted a feature request to change the number of trials at each step in #1550, and
- @kdlin wants to tune additional parameters such as `max_depth` and `max_bin` in the [Gitter conversation](https://gitter.im/optuna/optuna?at=5f2c8857dca038052a51abbd).

I created this issue to aggregate such user needs, and I'd like to discuss the feasible and user-friendly implementation of customizable steps.

## Description

Currently, I do not have any ideas for API design, but let me share some existing attempts.

**Approach 1**
I tried to extend `LightGBMTuner` to add a step to tune `max_depth`. See https://colab.research.google.com/drive/1vBfktSi7qXc8T4X6x1ib3IeqIlqcOAOH?usp=sharing. 
It was a bit complicated because I needed to modify two classes, i.e., `LightGBMTuner` and `_OptunaObjective`. I don't think this is a convenient way to customize LightGBM tuner's step, and we may need major refactoring of `LightGBMTuner` and `_OptunaObjective`.

**Approach 2**
As mentioned in [this Gitter comment](https://gitter.im/optuna/optuna?at=5f2e78f2d4bc1a71bc88e05c), @jeffzi created a prototype of `StepwiseTuner`, which accepts a dictionary containing step information. It also enables us to create a stepwise tuner for other libraries such as XGBoost and CatBoost.

kmedved:
A customizable stepwise tuner would be amazing functionality. The approach works incredibly well for LightGBM, and would be an even bigger boon to slower tuning libraries like Catboost or Ngboost (new probabilistic boosting library from Andrew Ng's team).

jeffzi:
Some time ago I prototyped a `StepwiseTuner` class that accepts a list of named Steps in its init. 

It's essentially an abstraction of `LightGBMTuner` that needs to be specialized in a dedicated subclass. The shared logic of looping through the steps, along with keeping track of the global time and trial budgets across steps is implemented in `StepwiseTuner.tune()`.

This approach is flexible. The steps and their order can be customized, the time and trial budget per step as well.

Here is a draft of the base interfaces:
```python

from optuna import distributions, logging, pruners, samplers

class BaseStep(ABC):
    def __init__(
[...]

Issue #1550, opened 2020-07-24
Make LightGBM Tuners number of trials per step parametrizable


Amyantis:
Hello,

The provided LightGBM tuners allow us to optimize LightGBM hyperparameters in a stepwise manner. Each step, defined as a method of [_LightGBMBaseTuner](https://github.com/optuna/optuna/blob/4fc8e5cd143a1ad8e5d9bd89e5bdb48a4713d3c8/optuna/integration/_lightgbm_tuner/optimize.py#L317), has a constant number of trials. The number of trials for each step is defined through the default values of the corresponding methods and thus are buried in the code.

The steps methods are called here, using default arguments:
https://github.com/optuna/optuna/blob/4fc8e5cd143a1ad8e5d9bd89e5bdb48a4713d3c8/optuna/integration/_lightgbm_tuner/optimize.py#L463
The default number of trials are set on each step method definition. For instance, 7 trials for _tune_feature_fraction_
https://github.com/optuna/optuna/blob/4fc8e5cd143a1ad8e5d9bd89e5bdb48a4713d3c8/optuna/integration/_lightgbm_tuner/optimize.py#L483

So far, we have:

- 7 trials for tune_feature_fraction
- 20 trials for tune_num_leaves
- 10 trials for tune_bagging
- 6 trials for tune_feature_fraction_stage2
- 20 trials for tune_regularization_factors

for a total of 63 optuna trials.

I would like, at least, to be able to parametrize that n=63 trials so that the number of trials at each step increases proportionally.

Thank you.

toshihikoyanase:
Thank you for your feature request.

> I would like, at least, to be able to parametrize that n=63 trials so that the number of trials at each step increases proportionally.

Sounds good! I'm also thinking of the mechanism to customize the tuning steps of `LightGBMTuner`. It may be a good starting point to change the number of trials.

I think we have come choices to change the total number of trials.

1. Add an option to specify the total number of trials.
```python
LightGBMTuner(n_trials=63)  # the default
LightGBMTuner(n_trials=68)  # +1 trial / step
LightGBMTuner(n_trials=58)  # -1 trial / step

LightGBMTuner(n_trials=64)
```
This approach is straightforward, but it may not be obvious which step should acquire the additional step if uses specify `n_trials=64`.

2. Add an option to specify the number of additional trials per step.
```python
LightGBMTuner(n_additional_trials=0)  # the default
LightGBMTuner(n_additional_trials=1)  # +1 trial / step
[...]

PR #1947, opened 2020-10-21
Add StepwiseTuner and StepwiseLightGBMTuner

jeffzi:
## Motivation

This PR is related to the discussion in #1632. The steps of `LightGBMTuner` are fixed and its implementation
is tightly coupled to a sequence of pre-defined steps, and to LightGBM in general.

The objective of this PR is twofold:

1. Define a customizable stepwise approach that could be used for multiple integration targets.
2. Refactor the LightGBM tuners to conform to 1. and test that its design is relevant.

### 1. StepwiseTuner

Introduce new concepts:

- `Step`: A [Step](https://github.com/jeffzi/optuna/blob/b96f4077b9958e308fea9e88eb47cb68688145b6/optuna/stepwise.py#L112) encampsulates the specific parameters of one specific optimization task.
- `StepwiseTuner`: A [StepwiseTuner](https://github.com/jeffzi/optuna/blob/b96f4077b9958e308fea9e88eb47cb68688145b6/optuna/stepwise.py#L238) corresponds to a sequence of optimization tasks (steps).
- `Step Objective`: Similar to a regular study objective but takes an extra dictionary of parameters to optimize.

A step holds:

- sampler
- pruner
- n_trials
- timeout
- distributions: A dictionary whose key and value are a parameter name and the corresponding distributions (`BaseDistribution`) to suggest from.

`BaseDistribution` is not usually exposed to end-users elsewhere in Optuna but it allows to defer the "suggestion" to a later time.
Implementations should provide reasonable pre-defined steps.

I chose to replicate the `Study` interface to feel natural. `StepwiseTuner` has a method `optimize()` that accepts the same 
arguments as `Study.optimize`, except the objective `func`. The `timeout` and `n_trials` passed to `optimize` controls the global budgets for all steps. 

`StepwiseTuner` also adds a pre-step that runs a single trial with default parameters to establish a baseline. Some users reported that LightGBMTuner does not always improve upon defaults, this solves the issue.

### 2. StepwiseLightGBMTuner & StepwiseLightGBMTuner

They subclass `StepwiseTuner`. [Pre-defined steps](https://github.com/jeffzi/optuna/blob/stepwise/optuna/integration/_lightgbm_tuner/steps.py) are offered, with the [default sequence](https://github.com/jeffzi/optuna/blob/d7b756371e453aeb5797b17770ba5d55b2f7e48e/optuna/integration/_lightgbm_tuner/steps.py#L186) replicating `LightGBMTuner`'s steps.

In addition to customizable steps, this implementation is further streamlined thanks to:
- Inference of the study's direction from a LightGBM [callback](https://github.com/jeffzi/optuna/blob/b96f4077b9958e308fea9e88eb47cb68688145b6/optuna/integration/_lightgbm_tuner/stepwise.py#L87). It decouples the tuner greatly from LightGBM internals and makes the code more robust.
- Serialization of the model as a string via LightGBM's [model_to_string](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.Booster.html#lightgbm.Booster.model_to_string). Consequently, I've not implemented file serialization. This is probably up to debate.

`StepwiseLightGBMTuner` fixes:
- #1351, #1636: `feval` is verified to work in the tests.
- #1550: parametrizable n_trials per step
- #1858: MAPE metric should be minimized, not maximized.
- Accepts a list of `metric` params but warns that only the first one will be used for optimization.

Lastly, I did not implement `sample_size`. From the code, I understand that it takes the last 'sample_size' rows of the dataset.
In my opinion, this could be easily handled by the user, not the tuner.

Example:
```python
import lightgbm as lgb
import numpy as np
import sklearn.datasets
from sklearn.metrics import accuracy_score
from sklearn.model_selection import train_test_split

from optuna.integration.lightgbm import StepwiseLightGBMTuner, step_feature_fraction


data, target = sklearn.datasets.load_breast_cancer(return_X_y=True)
train_x, val_x, train_y, val_y = train_test_split(data, target, test_size=0.25, random_state=42)
dtrain = lgb.Dataset(train_x, label=train_y)
dval = lgb.Dataset(val_x, label=val_y)

params = {
    "objective": "binary",
    "metric": "binary_logloss",
    "boosting_type": "gbdt",
    "verbosity": -1,
}
tuner = StepwiseLightGBMTuner(
    params,
    dtrain,
    steps=[step_feature_fraction(n_trials=2)], # use a predefined step
    valid_sets=[dval],
    verbose_eval=False,
    num_boost_round=50,
    early_stopping_rounds=5,
)
tuner.optimize()

model = tuner.get_best_booster()
prediction = np.rint(model.predict(val_x, num_iteration=model.best_iteration))
accuracy = accuracy_score(val_y, prediction)

print("Best params:", tuner.best_params)
print("  Accuracy = {}".format(accuracy))
```
### Other considerations

I wrote tests and docstrings but did not add the documentation to sphinx (waiting for your opinion on the PR).

I realize it may not be ideal to break the `LightGBMTuner` API but I think the proposed API homogenizes better with the rest of Optuna. If the PR is accepted, I propose to start by keeping both `LightGBMTuner` and `StepwiseLightGBMTuner` and scheduling `LightGBMTuner`'s deprecation.

The interface to scikit-learn and the alias for `lgb.train` should also be refactored in the future.

That's all for my proposal. I hope that you like it and that it opens a good discussion. :tada:

### DIFF
--- a/optuna/integration/__init__.py
+++ b/optuna/integration/__init__.py
@@ -14,7 +14,13 @@ _import_structure = {
     "cma": ["CmaEsSampler", "PyCmaSampler"],
     "mlflow": ["MLflowCallback"],
     "keras": ["KerasPruningCallback"],
-    "lightgbm": ["LightGBMPruningCallback", "LightGBMTuner", "LightGBMTunerCV"],
+    "lightgbm": [
+        "LightGBMPruningCallback",
+        "LightGBMTuner",
+        "LightGBMTunerCV",
+        "StepwiseLightGBMTuner",
+        "StepwiseLightGBMTunerCV",
+    ],
     "pytorch_ignite": ["PyTorchIgnitePruningHandler"],
     "pytorch_lightning": ["PyTorchLightningPruningCallback"],
     "sklearn": ["OptunaSearchCV"],
@@ -45,6 +51,8 @@ if TYPE_CHECKING:
     from optuna.integration.lightgbm import LightGBMPruningCallback  # NOQA
     from optuna.integration.lightgbm import LightGBMTuner  # NOQA
     from optuna.integration.lightgbm import LightGBMTunerCV  # NOQA
+    from optuna.integration.lightgbm import StepwiseLightGBMTuner  # NOQA
+    from optuna.integration.lightgbm import StepwiseLightGBMTunerCV  # NOQA
     from optuna.integration.mlflow import MLflowCallback  # NOQA
     from optuna.integration.mxnet import MXNetPruningCallback  # NOQA
     from optuna.integration.pytorch_ignite import PyTorchIgnitePruningHandler  # NOQA
--- a/optuna/integration/_lightgbm_tuner/__init__.py
+++ b/optuna/integration/_lightgbm_tuner/__init__.py
@@ -3,8 +3,8 @@ from typing import Any
 from optuna.integration._lightgbm_tuner.optimize import _imports as _imports
 from optuna.integration._lightgbm_tuner.optimize import LightGBMTuner
 from optuna.integration._lightgbm_tuner.optimize import LightGBMTunerCV  # NOQA
-from optuna.integration._lightgbm_tuner.steps import *
-from optuna.integration._lightgbm_tuner.stepwise import StepwiseLightGBMTuner
+from optuna.integration._lightgbm_tuner.steps import *  # NOQA
+from optuna.integration._lightgbm_tuner.stepwise import StepwiseLightGBMTuner  # NOQA
 from optuna.integration._lightgbm_tuner.stepwise import StepwiseLightGBMTunerCV  # NOQA
 
 
--- a/optuna/integration/_lightgbm_tuner/__init__.py
+++ b/optuna/integration/_lightgbm_tuner/__init__.py
@@ -3,6 +3,9 @@ from typing import Any
 from optuna.integration._lightgbm_tuner.optimize import _imports as _imports
 from optuna.integration._lightgbm_tuner.optimize import LightGBMTuner
 from optuna.integration._lightgbm_tuner.optimize import LightGBMTunerCV  # NOQA
+from optuna.integration._lightgbm_tuner.steps import *
+from optuna.integration._lightgbm_tuner.stepwise import StepwiseLightGBMTuner
+from optuna.integration._lightgbm_tuner.stepwise import StepwiseLightGBMTunerCV  # NOQA
 
 
 if _imports.is_successful():
--- a/optuna/integration/_lightgbm_tuner/steps.py
+++ b/optuna/integration/_lightgbm_tuner/steps.py
@@ -1,0 +1,194 @@
+from typing import Any
+from typing import Dict
+from typing import Optional
+from typing import Tuple
+
+import numpy as np
+
+from optuna import distributions
+from optuna import pruners
+from optuna import samplers
+from optuna import stepwise
+from optuna.trial import Trial
+
+
+__all__ = [
+    "step_feature_fraction",
+    "step_num_leaves",
+    "step_bagging",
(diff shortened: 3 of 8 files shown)


toshihikoyanase (commented):
Thank you for your great PR.
It contains not only a customizable version of `LightGBM`, but also a general interface which enables step-wise optimization for any objective functions.

I think it takes time to check all the lines, so let me leave just an initial comment.

To understand the usage of `StepwiseLightGBMTuner`, I created an example based on your example in this PR description and `examples/lightgbm_tuner_simple.py`. 
The main change from your example is to change the argument of `StepwiseLightGBMTuner.__init__`. More concretely, I removed `steps` and `num_boost_rounds`, and I changed the values of `verbose_eval` and `early_stopping_rounds`.

```python
    model = lgb.StepwiseLightGBMTuner(
        params, dtrain, valid_sets=[dtrain, dval], verbose_eval=100, early_stopping_rounds=100
    )

    model.optimize()
```

Do you think the optimization steps (i.e., sampler, search space, and `n_trials`) are identical to the [original example](https://github.com/optuna/optuna/blob/master/examples/lightgbm_tuner_simple.py)?

jeffzi:
If you don't provide `steps`, `StepwiseLightGBMTuner` will fall back to the [default steps](https://github.com/jeffzi/optuna/blob/d7b756371e453aeb5797b17770ba5d55b2f7e48e/optuna/integration/_lightgbm_tuner/steps.py#L186) which replicate the original `LightGBMTuner`, including `n_trials`.

In my own tests, the results are close but not identical because:
* `StepwiseTuner` runs one `Trial` with default hyperparameters to avoid performance regression compared to non-optimized hyperparameters. See code [here](https://github.com/jeffzi/optuna/blob/d7b756371e453aeb5797b17770ba5d55b2f7e48e/optuna/stepwise.py#L366).
* None of the approaches set the samplers'seeds. I believe this is something `StepwiseTuner` could potentially do. The problems are that 1) samplers don't have a common method `set_seed` and 2) it would not work if `n_jobs` > 1.

toshihikoyanase:
Thank you for your clear and concise answer.

> `StepwiseLightGBMTuner` will fall back to the default steps which replicate the original `LightGBMTuner`, including n_trials, sampler and search space.
> - `StepwiseTuner` runs one `Trial` with default hyperparameters to avoid performance regression compared to non-optimized hyperparameters.

Cool. I feel the difference between `LightGBMTuner` and `StepwiseLightGBMTuner` is small, and the change to evaluate the baseline configuration is reasonable. As suggested in the description, I also feel `StepwiseLightGBMTuner` can replace the current `LightGBMTuner`. I'd like to check the feasibility of the plan.

> - None of the approaches set the samplers'seeds.

Good point! Users sometimes make optimization reproducible, and I think the proposal is reasonable. I also check if this suggestion can be supported in the new API.

jeffzi:
I found another difference with `LightGBMTuner`. 

`LightGBMTuner` always picks the last `valid_sets` for the optimization (#1984) whereas `StepwiseLightGBMTuner` always picked the first `valid_sets`. 

The reasoning for picking the first `valid_sets` is to follow `first_metric_only`, which is required by `StepwiseLightGBMTuner`. I found out that LightGBM automatically ignores training data in valid_sets. I've now implemented this behavior for `StepwiseLightGBMTuner`. It was a bit tricky because `lgb.train` identifies training sets in `valid_sets` with the `is` operator. I had [to make sure to keep the same reference](https://github.com/jeffzi/optuna/blob/9c064b53ad68904abda203122ed45b1c73261718/optuna/integration/_lightgbm_tuner/stepwise.py#L88) for all train sets when `valid_sets` is copied.

If multiple `valid_sets` are detected, a warning explains that the first non-training set will be used.

PR #1948, opened 2020-10-21
[RFC] Run Kurobako on GitHub Actions using iterative/cml

crcrpar:
discussion items

- [ ] trigger configuration
- [ ] tasks

## Motivation

I think it's desirable to run benchmarks and share results automaticaly using CI to detect the regression of Optuna and the superiority of new samplers and pruners.


crcrpar:
Let me close as #1969 seems to work as expected.

Issue #1860, opened 2020-09-24
[RFC] Storage system and user attribute length limits


hvy:
## Motivation

The [storage specification](https://github.com/optuna/optuna/blob/master/optuna/storages/_base.py) doesn't mention any limits on the lengths of the system and user attributes that can be stored. Nor does it specify how to handle such writes. The behaviors for storing such attributes are in other words undefined and storage specific. This may for instance lead to unexpected errors raised from the storage layer to the user, when the user/Optuna tries to store a string that's too long (c.f. https://github.com/optuna/optuna/pull/1833#discussion_r490637113). The [`RDBStorage` for instance defines a maximum length](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/models.py#L31) in the schema but will simply propagate errors raised from sqlalchemy on INSERT failure further up. The `InMemoryStorage` on the other hand is only limited by the memory that's available. This should be addressed.

## Description

See above.

## Alternatives (optional)

Please refer to two alternatives suggested by @c-bata in https://github.com/optuna/optuna/pull/1833#issuecomment-694617161.

hvy:
Let me close this issues since https://github.com/optuna/optuna/pull/2395 addressed the failures with the `RDBStorage`.

PR #1949, opened 2020-10-21
Safe decode of system/usr attributes for optuna studies

zzorba:
## Motivation
We encountered issues where very large callstacks were being serialized into remote storage (as part of the `fail_system` system attr).  Unfortunately, because the system_attr serializer converts things to json before storing them, the column truncation caused by the database caused these simple "string" attributes to no longer be decodeable as 'json'. This causes the trial to fail to load, rendering the study inoperable for future trials or later analysis.

## Description of the changes
The code here wraps the current usage of `json.loads` with a simple try/except block to catch bad JSON decoding. If bad JSON is encountered, it simply treats the value as a string (not ideal in all situations, but arguably preferable behavior to rendering the trial/study inoperable.  For very long string fields this seems okay, for objects it might not be ideal but it should still be quite apparent that they are broken (as opposed to today where it fails when reading them back).

An alternate approach could try to prevent very long strings from entering the system, but defensive reads seems worthwhile regardless.

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -283,7 +283,7 @@ class RDBStorage(BaseStorage):
         # Ensure that that study exists.
         models.StudyModel.find_or_raise_by_id(study_id, session)
         attributes = models.StudyUserAttributeModel.where_study_id(study_id, session)
-        user_attrs = {attr.key: json.loads(attr.value_json) for attr in attributes}
+        user_attrs = {attr.key: self._decode_json(attr.value_json) for attr in attributes}
         # Terminate transaction explicitly to avoid connection timeout during transaction.
         self._commit(session)
 
@@ -296,7 +296,7 @@ class RDBStorage(BaseStorage):
         # Ensure that that study exists.
         models.StudyModel.find_or_raise_by_id(study_id, session)
         attributes = models.StudySystemAttributeModel.where_study_id(study_id, session)
-        system_attrs = {attr.key: json.loads(attr.value_json) for attr in attributes}
+        system_attrs = {attr.key: self._deck_json(attr.value_json) for attr in attributes}
         # Terminate transaction explicitly to avoid connection timeout during transaction.
         self._commit(session)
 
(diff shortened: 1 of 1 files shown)


hvy:
Thanks for raising an important issue. You're right that this is manifested as a bug. It is actually addressed in this existing issue https://github.com/optuna/optuna/issues/1860 and I'm wondering if we shouldn't raise the priority of this one instead.

For this particular manifestation, I assume you are talking about the `fail_reason` here https://github.com/optuna/optuna/blob/master/optuna/_optimize.py#L206.

> as opposed to today where it fails when reading them back

One question. Optuna should fail to write (in the line referenced above) but not read so I haven't understood how this change will work around the problem. Could you provide reproduction steps or elaborate a bit?

zzorba:
Sorry for delayed response.

Yes you are correct, we ran into issues with `fail_reason` specifically -- basically it seems like in some situations it tries to store a complete stack, which is too large for the backing datastore.

Strangely enough, the data is making into the Database (we are using Aurora RDS (MySql) for our backend). But it seems like Aurora/the storage driver must have truncated the string that was too long, since we end up with something that has an opening `"` but no closing `"` causing it to fail when read by the `json.loads` call (since it is not a valid 'json' string anymore). 

I agree that preventing the 'too large' data from going into the database would be a more appropriate fix long term.

hvy:
Thanks for your explanation. So it might be specific to Aurora? I'm a bit skeptical about this change then since it's quite specific to the experiment environment. I'm wondering if you managed to work around it.

I by the way created a related PR #1964 on which you might some comment.

zzorba:
Right, to be clear the aurora specific part I'm hypothesizing is the fail on deserialize because it truncates on store rather than fail the write transaction (due to column width). Those seem to be the two reasonable behaviors a database would have
1) Accept as much as the data and truncate it (which would lead to json deserialization problems when we read it back as 'json').
2) Reject the transaction as malformed.

I don't know which is more common in the various SQL flavors, since I assume its not a specified part of the spec.

So far we've worked around this by using a forked branch with these changes, as a temporary stopgap.

Thank you for suggesting that parallel PR, I agree that dropping `fail_reason` is one outcome, also snipping it to a reasonable length (500/1000 characters) would also work.

PR #1951, opened 2020-10-22
Support sep-CMA-ES algorithm

c-bata:
## Motivation

sep-CMA-ES is an algorithm which constrains the covariance matrix to be diagonal.
Due to reduce the model complexity, the learning rate for the covariance matrix is reduced. Consequently, this algorithm outperforms CMA-ES on separable functions.

## Description of the changes

sep-CMA-ES is implemented in https://github.com/CyberAgent/cmaes and released at v0.7.0.
It provides the same interface with `cmaes.CMA` class.

In this PR, I introduces `use_separable_cma` option as an experimental feature.
If the option is True, CmaEsSampler uses `cmaes.SepCMA` optimizer.

## Benchmark

The benchmark is continuously run on CyberAgent/cmaes repo.


optuna/samplers/_cmaes.py
```diff
@@ -66,6 +71,9 @@ def objective(trial):
       size. In Proceedings of the IEEE Congress on Evolutionary Computation (CEC 2005),
       pages 1769–1776. IEEE Press, 2005.
       <http://www.cmap.polytechnique.fr/~nikolaus.hansen/cec2005ipopcmaes.pdf>`_
+    - `Raymond Ros, Nikolaus Hansen. A Simple Modification in CMA-ES Achieving Linear
+      Time and Space Complexity. Raport instytutowy, INRIA, 2008.
```
toshihikoyanase, line 75: I'm wondering where `Raport instytutowy` comes from. The PDF document asks us to cite this paper as follows, so how about revising it?

> Raymond Ros, Nikolaus Hansen. A Simple Modification in CMA-ES Achieving Linear Time and
Space Complexity. 10th International Conference on Parallel Problem Solving From Nature, Sep
2008, Dortmund, Germany. inria-00287367

optuna/samplers/_cmaes.py
```diff
@@ -142,6 +150,16 @@ def objective(trial):
                 to set this flag :obj:`True` when the :class:`~optuna.pruners.HyperbandPruner` is
                 used. Please see `the benchmark result
                 <https://github.com/optuna/optuna/pull/1229>`_ for the details.
+
+        use_separable_cma:
+            If this is :obj:`True`, the covariance matrix is constrained to be diagonal.
+            Due to reduce the model complexity, the learning rate for the covariance matrix
+            is reduced. Consequently, this algorithm outperforms CMA-ES on separable functions.
```
toshihikoyanase, line 157: The abstract of the [paper](https://hal.inria.fr/inria-00287367/document) says
> Because the model complexity is reduced, the learning rate for the covariance matrix can be increased.

I guess it is not consistent with the second sentence of this description.

optuna/samplers/_cmaes.py
```diff
@@ -373,6 +400,17 @@ def _init_optimizer(
...
+
+        if self._use_separable_cma:
+            return SepCMA(
+                mean=mean,
+                sigma=sigma0,
+                bounds=bounds,
+                seed=self._cma_rng.randint(1, 2 ** 31 - 2),
+                n_max_resampling=10 * n_dimension,
+                population_size=population_size,
+            )
```
toshihikoyanase, line 412: I don't think the current test cases cover this logic. How about updating the `test_init_cma_opts`, for example, as follows?

```python
@pytest.mark.parametrize("use_separable_cma, cma_class_str", 
    [(False, "optuna.samplers._cmaes.CMA"), (True, "optuna.samplers._cmaes.SepCMA")])
def test_init_cmaes_opts(use_separable_cma: bool, cma_class_str: str) -> None:
    sampler = optuna.samplers.CmaEsSampler(
        x0={"x": 0, "y": 0},
        sigma0=0.1,
        seed=1,
        n_startup_trials=1,
        independent_sampler=DeterministicRelativeSampler({}, {}),
        use_separable_cma=use_separable_cma
    )

    with patch(cma_class_str) as cma_class:
        cma_obj = MagicMock()
        cma_obj.ask.return_value = np.array((-1, -1))
        cma_obj.generation = 0
        cma_obj.population_size = 5
        cma_class.return_value = cma_obj
        study.optimize(
            lambda t: t.suggest_uniform("x", -1, 1) + t.suggest_uniform("y", -1, 1), n_trials=2
        )

        assert cma_class.call_count == 1

        _, actual_kwargs = cma_class.call_args
        assert np.array_equal(actual_kwargs["mean"], np.array([0, 0]))
        assert actual_kwargs["sigma"] == 0.1
        assert np.allclose(actual_kwargs["bounds"], np.array([(-1, 1), (-1, 1)]))
        assert actual_kwargs["seed"] == np.random.RandomState(1).randint(1, 2 ** 32)
        assert actual_kwargs["n_max_resampling"] == 10 * 2
        assert actual_kwargs["population_size"] is None
```

toshihikoyanase (commented):
Thank you for your PR. I have some small comments.

BTW, `sep-CMA-ES` is interesting because it can sample parameter values independently according to the [original paper](https://hal.inria.fr/inria-00287367/document).

>  Consequently, in contrast to the CMA-ES, dependencies between variables are not captured and coordinates are sampled independently.

Just out of curiosity, do you think if we can provide `SepCmaEsSampler` which implements only `sample_independent`?

c-bata:
> BTW, sep-CMA-ES is interesting because it can sample parameter values independently according to the original paper.

Sounds interesting. I asked your question to @nmasahiro who implements sep-CMA-ES in https://github.com/CyberAgent/cmaes and he said

toshihikoyanase:
> I asked your question to @nmasahiro who implements sep-CMA-ES in https://github.com/CyberAgent/cmaes and he said yes.

Amazing! I'm looking forward to it.

> I will open an another pull request to implement SepCmaEsSampler. 

Before implementing the sampler, I think we can discuss the API design a little bit more. A possible alternative approach is to implement sep-CMA-ES as `CmaEsSampler.sample_independent`. The multivariate TPE sampler employed this approach. What do you think of it?

c-bata:
> A possible alternative approach is to implement sep-CMA-ES as CmaEsSampler.sample_independent

In my opinion, adding a new sampler is a good approach to keep the simple implementation.
But to refer some utility functions are required from both samplers (e.g. `_to_optuna_param` , `_to_cma_param` and so on), I add `SepCmaEsSampler` into `samplers/_cmaes` module like this:

* `samplers/_cmaes/cmaes.py`  - Define `CmaEsSampler`
* `samplers/_cmaes/sep_cmaes.py`  - Define `SepCmaEsSampler`
* `samplers/_cmaes/space.py`  - Define `_to_optuna_param` , `_to_cma_param` and so on.

What do you think?

c-bata:
I open #1975. Let me discuss it at the issue.

toshihikoyanase:
Thank you for creating the issue. I'll check it.

c-bata:
I reopened this to implement sep-CMA-ES as the relative sampler (See https://github.com/optuna/optuna/issues/1975#issuecomment-727722277 for the reason).

HideakiImamura:
@c-bata Thanks for the discussion at #1975 and the implementation. Is this PR ready for review? I'd like to move this PR ahead.

c-bata:
@HideakiImamura Sorry, this PR is not ready for review. I'll update this next week :bow:

c-bata:
@toshihikoyanase @HideakiImamura Sorry for the late response. I applied the suggestions from @toshihikoyanase and made this PR to ready for review.

c-bata:
@HideakiImamura @toshihikoyanase Can you review this if you have time?

optuna/samplers/_cmaes.py
```diff
@@ -144,6 +153,16 @@ def objective(trial):
...
 
+        use_separable_cma:
+            If this is :obj:`True`, the covariance matrix is constrained to be diagonal.
+            Due to reduce the model complexity, the learning rate for the covariance matrix
+            is increased. Consequently, this algorithm outperforms CMA-ES on separable functions.
+
+            .. note::
+                Added in v2.3.0 as an experimental feature. The interface may change in newer
+                versions without prior notice. See
+                https://github.com/optuna/optuna/releases/tag/v2.3.0.
```
toshihikoyanase, line 164: Could you update the version number?
```suggestion
                Added in v2.6.0 as an experimental feature. The interface may change in newer
                versions without prior notice. See
                https://github.com/optuna/optuna/releases/tag/v2.6.0.
```

optuna/samplers/_cmaes.py
```diff
@@ -325,23 +353,31 @@ def sample_relative(
...
                 continue
 
-            # Check "cma:optimizer" key for backward compatibility.
-            optimizer_str = optimizer_attrs.get("cma:optimizer", None)
-            if optimizer_str is None:
+            if not self._use_separable_cma and "cma:optimizer" in optimizer_attrs:
+                # Check "cma:optimizer" key for backward compatibility.
+                optimizer_str = optimizer_attrs["cma:optimizer"]
+            else:
                 optimizer_str = _concat_optimizer_attrs(optimizer_attrs)
```
toshihikoyanase, line 378: [Question] Do you expect users to switch CMA-ES to sep CMA-ES in the middle of the optimization?
For example, the following code does not raise any errors or warnings, but I'm not sure if it works properly or not.

```python
import optuna

def objective(trial):

    return (x - 0.5) ** 2 + (y - 0.5) ** 2

study = optuna.create_study(sampler=optuna.samplers.CmaEsSampler())

study.sampler=optuna.samplers.CmaEsSampler(use_separable_cma=True)

```

toshihikoyanase (commented):
Sorry for the delayed response, but I added a comment and a question.

HideakiImamura (approved):
LGTM besides @toshihikoyanase's comments.

optuna/samplers/_cmaes.py
c-bata: Yes, I expected such a situation and this PR supports the case you commented.

Although CmaEsSampler stores its `optimizer` and the restart count in `trial_system_attr`, the different system_attr keys are used between (normal) CMA-ES and sep-CMA-ES. So the parameters of sep-CMA-ES is initialized with `mean=np.array([0.5, 0.5])` and `sigma=(1-0)/6`. It means sep-CMA-ES does not exploit the state of (normal) CMA-ES.

c-bata:
Thank you for your review! I applied your suggestion and replied to your comment.

optuna/samplers/_cmaes.py
toshihikoyanase: Thank you for your explanation. That makes sense.
Users may ask questions about this behavior, and I'll answer them with the reference to your comment.

toshihikoyanase (approved):
Looks great to me. Thank you for accomplishing the long-running PR.

PR #1953, opened 2020-10-23
Make code example of `MOTPEMultiObjectiveSampler` executable

toshihikoyanase:
## Motivation
This is a follow-up task of #1530. Fix https://github.com/optuna/optuna/pull/1530/files#r510636616.

## Description of the changes
- import `optuna` instead  of `numpy`
- use `suggest_float` instead of `suggest_uniform`
- define `study`


hvy (approved):
Thanks, LGTM!

PR #1954, opened 2020-10-23
Add Fastai V2 pruner callback

### LINKED ISSUE #1674: [RFC] Support fastai v2
We've supported fastai with `FastAIPruningCallback` though it has targeted fastai v1.
Recently, FastAI V2 is released; thus we'd love to support it.
Being that said, early stopping callback of v2 which will be the base class of `FastAIPruningCallback` differs from that of v1 in e.g., the `__init__`'s arguments.

- v2: https://github.com/fastai/fastai/blob/72590db2e66af6dd0eaa8c8874a80f11a4b8cbc2/fastai/callback/tracker.py#L47-L62
- v1: https://github.com/fastai/fastai1/blob/ff9d8edaca74114ce3045ea00183fa5cfa897904/fastai/callbacks/tracker.py#L53-L75

hal-314:
## Motivation

Fixes  #1674

## Description of the changes

1. Current `FastAIPruningCallback` is renamed to `FastAIV1PruningCallback`. `FastAIPruningCallback` is marked as deprecated.
1. Tweak current test to use `FastAIV1PruningCallback`
1. Implement `FastAIV2PruningCallback` that is a fast V2 (code + tests)

I don't know how to finish as:
1. ¿How to manage requirements? I locally pass fast v2 test as I have installed fastai v2. However, I don't how to automatize this to install fastai V1 for some tests and fastai v2 for another
1. As a results I didn't run circleci job locally.
1. I wrote docs for both new classes. However, they doesn't show under fast.ai docs. ¿Where do I need to add them?


crcrpar:
Hi @hal-314 

First of all, thank you for tackling this, and I'm sorry for not having the issue informative enough.

In my opinion, now `pip` and `conda` both install fastai V2 by default.
So, I think we can make `FastAIPruningCallback` as an alias for `FastAIV2PruningCallback`.
This may or may not be too drastic, but I would tolerate as we will still support FastAI v1 with FastAIV1PruningCallback.
Also, decorate FastAIV1PruningCallback with `deprecated` decorator.

>     1. ¿How to manage requirements? I locally pass fast v2 test as I have installed fastai v2. However, I don't how to automatize this to install fastai V1 for some tests and fastai v2 for another

What you need to do is to update "fastai<2" in `setup.py` to "fastai>=2", for example.
With this change, there's no way to run the tests for v1 concisely with CI, so I think we can remove the tests for v1.
This generally is not desirable though I think I would say okay.

Ref: https://github.com/hal-314/optuna/blob/fastaiv2_integration/setup.py#L105

>     3\. I wrote docs for both new classes. However, they doesn't show under fast.ai docs. ¿Where do I need to add them?

Basically FastAI section of https://github.com/hal-314/optuna/blob/fastaiv2_integration/docs/source/reference/integration.rst is what I would like you to update.

Also, could you update https://github.com/optuna/optuna/blob/master/optuna/integration/__init__.py ?

hal-314:
@crcrpar Could you help me with CI errors?

1. I don't how why black format fails. I run it locally (`black .`) and it doesn't want to reformat `setup.py` or `test_fastaiv2`:/ 
1. `test_fastaiv2.py` fails because `fastai` module isn't found. However, I added it in setup.py. I'm missing something?

I implemented your suggestions.

crcrpar:
>     1. I don't how why black format fails. I run it locally (`black .`) and it doesn't want to reformat `setup.py` or `test_fastaiv2`:/

The reason your local black did nothing might be the version of your black.
Could you follow the changes suggested by black which you can check on https://github.com/optuna/optuna/pull/1954/checks?check_run_id=1308581038?

Alternatively, if you have docker intalled on yor environment, `docker run -it -v $(pwd):/workspaces optuna/optuna:py3.7-dev /bin/bash -c "black --check ."`.

---

>     2\. `test_fastaiv2.py` fails because `fastai` module isn't found. However, I added it in setup.py. I'm missing something?

There are two solutions to solve this I think.

One is to rename `test_fastaiv2.py` to `test_fastai.py` because `test_fastai.py` is skipped in CI as in the following two links

- https://github.com/hal-314/optuna/blob/fastaiv2_integration/.circleci/config.yml#L190
- https://github.com/hal-314/optuna/blob/fastaiv2_integration/.github/workflows/tests-integration.yml#L63

The other solution is to support fastai in Python 3.8 environment (When I initirally implemented fastai integration module, I wan't sure fastai supported Python 3.8 or not but now it seems to support. ref: https://github.com/fastai/fastai/commit/5a102284e685511e69e93b9fde8ca8b1ea588c4e). If you like this way, what you need to do is to update the setup.py since fastai is not installed in Python 3.8 environment as follows

- https://github.com/hal-314/optuna/blob/fastaiv2_integration/setup.py#L104
- https://github.com/hal-314/optuna/blob/fastaiv2_integration/setup.py#L139
- https://github.com/hal-314/optuna/blob/fastaiv2_integration/setup.py#L175

hal-314:
I fixed style and tests issues. I'm deeply sorry to make 2 extra commits due to my own mistakes with `test_fastaiv2.py` changes :( . 

Now, fastai test also runs in python 3.8. I removed skip test related bits.

It's ready for review and merge.

crcrpar (commented):
No worry about the extra commits for style changes. I also frequently do the same, embarassingly.
Also, thank you for checking that fastai now supports 3.8 👍 

I forgot to say this 😅  Could you add `examples/fastaiv1_simple.py` to `IGNORES` https://github.com/hal-314/optuna/blob/d7bc1ae40081fec328228380232e6707ec850575/.github/workflows/examples.yml#L49-L53? 
This is because we run all the examples on CI daily, which means we need to skip fastaiv1 example.

hal-314:
I don't understand what I need to put in `IGNORES`. Is only fastaiv1 that needs to be ignored or fastaiv2 too? Also, with this PR, I changed `setup.py` so it install fastaiv2 instead of fastaiv1. So, no fastaiv1 example is going to work under any python version. Therefore, fastaiv1 example should be skipped in CI. 

My bash skills are quite limited. Is this the right change (lets run fastaiv2 example while avoiding any fastaiv1)?
```bash
        if [ ${{ matrix.python-version }} = 3.8 ]; then
          IGNORES='chainermn_.*|dask_ml_.*|keras_.*|tensorboard_.*|tensorflow_.*|tfkeras_.*|fastaiv1_.*|allennlp'
        else
          IGNORES='chainermn_.*|fastaiv1_.*'
        fi
```

crcrpar:
Sorry for my poor explanation. You are right. I meant to say fastaiv1 example should be skipped in CI and thus your change is correct.

crcrpar (approved):

crcrpar:
Sorry @hal-314, could you fix the error of blackdoc?

ref: https://github.com/optuna/optuna/pull/1958

hal-314:
@crcrpar There is an small conflict in setup.py due to allennlp has being updated. When it's ready to merge, I can fix it.

HideakiImamura:
Sorry for the delay in reviewing. I will now start looking this PR.

examples/fastaiv2_simple.py
```diff
@@ -0,0 +1,108 @@
...
+    n_channels.append(2)
+
+    model = SimpleCNN(n_channels)
+
+    learn = Learner(
+        data,
+        model,
+        metrics=[accuracy],
+        # You could as FastAIPruningCallback in the fit function
+        cbs=[FastAIPruningCallback(trial), CudaCallback],
```
HideakiImamura, line 72: Ditto. `CudaCallback` doesn't seem to be necessary.

examples/fastaiv2_simple.py
```diff
@@ -0,0 +1,108 @@
+"""
+Optuna example that optimizes convolutional neural network and data augmentation using FastAI V2.
```
HideakiImamura, line 2: A nit.
```suggestion
Optuna example that optimizes convolutional neural network and data augmentation using fastai V2.
```

examples/fastaiv2_simple.py
```diff
@@ -0,0 +1,108 @@
...
+    # to disable progress bar and logging info
+    with learn.no_bar():
+        with learn.no_logging():
+            learn.fit(EPOCHS)
+
+    return learn.validate()[-1]
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="FastAI V2 example.")
```
HideakiImamura, line 85: Ditto.
```suggestion
    parser = argparse.ArgumentParser(description="fastai V2 example.")
```

HideakiImamura (commented):
Thanks for the great work! Sorry for the late review. The changes basically LGTM. I have several minor comments. Could you take a look?

tests/integration_tests/test_fastaiv2.py
```diff
@@ -0,0 +1,57 @@
...
+    data = DataLoaders(train_loader, test_loader, path=tmpdir)
+
+    def objective(trial: optuna.trial.Trial) -> float:
+        model = nn.Sequential(nn.Linear(20, 1), nn.Sigmoid())
+        learn = Learner(
+            data,
+            model,
+            loss_func=F.nll_loss,
+            metrics=[accuracy],
+            cbs=[CudaCallback],
```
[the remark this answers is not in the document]
hal-314: Here, we need it as `_generate_dummy_dataset` doesn't return a cuda tensor. Normally, you don't need it as fastai dataloaders load into gpu. So, I used CudaCallback to make it explicit.

examples/fastaiv2_simple.py
hal-314: Here, you are right. It isn't necessary.

hal-314:
Now, this PR is ready to merge :)

optuna/integration/fastaiv2.py
```diff
@@ -0,0 +1,80 @@
...
+        # self.idx is set by TrackTrackerCallback
+        self.optuna_trial.report(self.recorder.final_record[self.idx], step=self.epoch)
+
+        if self.optuna_trial.should_prune():
+            raise CancelFitException()
+
+    def after_fit(self) -> None:
+        super().after_fit()
+        if self.optuna_trial.should_prune():
+            raise optuna.TrialPruned(f"Trial was pruned at epoch {self.epoch}.")
```
HideakiImamura, line 77: Could you replace `optuna_trial` as `trial`?
```suggestion
        self.trial.report(self.recorder.final_record[self.idx], step=self.epoch)

        if self.trial.should_prune():
            raise CancelFitException()

    def after_fit(self) -> None:
        super().after_fit()
        if self.trial.should_prune():
            raise optuna.TrialPruned(f"Trial was pruned at epoch {self.epoch}.")
```

examples/fastaiv2_simple.py
```diff
@@ -0,0 +1,108 @@
...
+You can run this example as follows, pruning can be turned on and off with the `--pruning`
+argument.
+    $ python fastaiv2_simple.py [--pruning]
+"""
+
+import argparse
+
+from fastai.vision.all import accuracy
+from fastai.vision.all import aug_transforms
+from fastai.vision.all import CudaCallback
```
HideakiImamura, line 21: Could you remove the unused `CudaCallback` to pass `flake8`?
```suggestion
```

tests/integration_tests/test_fastaiv2.py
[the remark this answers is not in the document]
HideakiImamura: Thank you for your confirmation. I understood what you intend.

HideakiImamura (commented):
Thanks for the update. I suggest 2 minor changes to pass the CI. Could you take a look?

HideakiImamura:
Hi @hal-314! Do you have any trouble with this PR? I'd like to help you.

hal-314:
Hi @HideakiImamura . Sorry for the long delay. I have some very busy days and I didn't have time to check my code. I fixed your suggestions. 

Also, as you suggested, I have changed `.. code::` to `.. testcode::` but Sphinx docs fails as `cnn_learner` and other variables aren't defined in the code snippet. I think that defining them would collude code example in docs. User can check fastai example inside example scripts folder. So, can I keep `.. code::` instead of `.. testcode::`?

crcrpar:
Hi @hal-314, Yes, I'd use `.. code::`. 

@HideakiImamura I don't think it reasonable to pay the cost of running this code in Sphinx doctest. This is because the current code requires `dls` which consists of training dataset and validation dataset and `resnet18`, a trained ResNet18 model.
So, to make this testcode, we need to import fastai (In fastai, start import is kind of common which I don't want let it appear in documentation), we need to define `dls` which should suit `resnet18` fine-tuning, for example, CIFAR-10/100, and we need to have CI download the trained ResNet18 parameters.

HideakiImamura:
@hal-314 @crcrpar I agree with you. Thanks for your comments. It's reasonable to use `.. code::` here.

HideakiImamura (approved):
Thanks for the swift action and sorry for the late review. LGTM!

PR #1959, opened 2020-10-25
Support options available on AllenNLP except to `node_rank` and `dry_run`

himkt:
## Motivation
This PR adds `AllenNLPExecutor` support options available on [`allennlp.commands.train`](https://docs.allennlp.org/master/api/commands/train/).

## Description of the changes
Add options ~`recover`~, `force`, and `file_friendly_logging` to `AllenNLPExecutor`.

### DIFF
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -411,7 +411,9 @@ class AllenNLPPruningCallback(EpochCallback):
     """
 
     def __init__(
-        self, trial: Optional[optuna.trial.Trial] = None, monitor: Optional[str] = None,
+        self,
+        trial: Optional[optuna.trial.Trial] = None,
+        monitor: Optional[str] = None,
     ):
         _imports.check()
 
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -261,6 +261,17 @@ class AllenNLPExecutor(object):
             A path which model weights and logs are saved.
         metrics:
             An evaluation metric for the result of ``objective``.
+        recover:
+            If recover is ``True``, an executor resumes training
+            from the state in ``serialization_dir``.
+            Note that recovering training in ``AllenNLPExecurot`` needs to fix random seed
+            in a sampler for obtaining reproducible optimization result.
+            Please see `the FAQ <https://optuna.readthedocs.io/en/stable/
+            faq.html#how-can-i-obtain-reproducible-optimization-results>`_ for more information.
+        force:
+            If True, an executor overwrite the output directory if it exists.
+        file_friendly_logging:
+            If True, tqdm status is printed on separate lines and slows tqdm refresh rate.
         include_package:
             Additional packages to include.
             For more information, please see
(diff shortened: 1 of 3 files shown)


optuna/integration/allennlp.py
```diff
@@ -274,6 +278,8 @@ def __init__(
         config_file: str,
         serialization_dir: str,
         metrics: str = "best_validation_accuracy",
+        force: bool = False,
+        file_friendly_logging: bool = False,
         *,
         include_package: Optional[Union[str, List[str]]] = None
```
HideakiImamura, line 284: The arguments of `force` and `file_friendly_logging` are placed before `include_package`. Changing the order of arguments affects user code. How about placing them after `include_package`?
```suggestion
        *,
        include_package: Optional[Union[str, List[str]]] = None,
        force: bool = False,
        file_friendly_logging: bool = False,
```

himkt: Thank you for the suggestion! (I think this change doesn't introduce breaking compatibility since `include_package` is a keyword-only argument. But I agree with you for placing them after `include_package`)

HideakiImamura (approved):
Thanks for the update! LGTM!

himkt:
@toshihikoyanase Let me kindly ping. :pray:

tests/integration_tests/allennlp_tests/test_allennlp.py
```diff
@@ -173,6 +173,25 @@ def test_allennlp_executor_with_include_package_arr() -> None:
...
+            trial,
+            "tests/integration_tests/allennlp_tests/example_with_include_package.jsonnet",
+            tmp_dir,
+            force=True,
+            file_friendly_logging=True,
+            include_package=["tests.integration_tests.allennlp_tests.tiny_single_id"],
+        )
+
+        assert executor._force
+        assert executor._file_friendly_logging
```
toshihikoyanase, line 192: I guess the test may be more solid if we check if these values are used in `executor.run`. How about using the `unittest.mock` as follows?

```python
        from unittest import mock
        with open(os.path.join(executor._serialization_dir, "metrics.json"), "w") as fout:
            json.dump({executor._metrics: 1.0}, fout)

        with mock.patch("allennlp.commands.train.train_model", return_value=None) as mock_obj:
            executor.run()
            assert mock_obj.call_args[1]["force"]
            assert mock_obj.call_args[1]["file_friendly_logging"]
```

toshihikoyanase (changes requested):
I'm sorry for the delayed response. I added small comments on documentation and tests.

tests/integration_tests/allennlp_tests/test_allennlp.py
```diff
@@ -236,7 +243,7 @@ def objective(trial: optuna.Trial) -> float:
                 text_field_embedder=embedder, seq2vec_encoder=encoder, vocab=vocab
             )
             optimizer = torch.optim.SGD(model.parameters(), lr=0.1)
-            data_loader = DataLoader(
+            data_loader = allennlp.data.dataloader.PyTorchDataLoader(
```
himkt, line 246: > error: Argument "data_loader" to "GradientDescentTrainer" has incompatible type "torch.utils.data.dataloader.DataLoader[Any]"; expected "allennlp.data.dataloader.DataLoader"

For `mypy` error.

himkt:
@toshihikoyanase Thank you for the suggestion. As you said, it's better to use `unittest.mock`. 0a77df4 changes the way to test . PTAL. :pray:

toshihikoyanase (approved):

toshihikoyanase:
@himkt Could you resolve conflict, please?

toshihikoyanase:
Thank you for your update. Merged!

PR #1962, opened 2020-10-26
Fix path for `store_artifacts` step in `document` CircleCI job

harupy:
## Motivation

The path for the `store_artifacts` step in the `document` CircleCI job should be `docs/build/html` not `docs/build`.

## Description of the changes

With this change, the document job should finish a bit faster.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,7 +87,7 @@ jobs:
             make html
 
       - store_artifacts:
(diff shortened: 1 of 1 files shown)


hvy (approved):
Right, we don't need the intermediate files and such. LGTM!

toshihikoyanase (approved):
LGTM!

PR #1964, opened 2020-10-27
[RFC] Stop storing error stack traces in `fail_reason` in trial `system_attr`

hvy:
## Motivation

Objective function errors are stored in the system attributes of trials under the `fail_reason` key. This is not documented. We should either document it or remove it if it isn't necessary (note that the error is logged too).

## Description of the changes

Stops storing the stack trace in the `fail_reason` trial system attribute.

## Alternative

Document `fail_reason`.

### Note

See also https://github.com/optuna/optuna/pull/1949.


zzorba:
Just to echo this PR, stack traces can be very large so if we do decide to store them, it would probably be wise to truncate them somewhat. We have encountered some issues were very large stack traces cause json deserialization of trials to start failing due to storage limits imposed by the database on column widths.

toshihikoyanase:
I think `fail_reason` in `system_attrs` is originally introduced to distinguish stochastic errors from deterministic ones. In the case of stochastic errors, we may be able to get trial values if we rerun the failed trials.
But I've never seen the actual users of this feature, and I support this PR.

BTW, how about adding an exception to the argument of callback functions?
If users can access exception instances, users can easily implement such a re-run mechanism compared with parsing the log message or system_attrs.

```python
def callback(study: Study, trial: FrozenTrial, exception: Optional[Exception] = None):
    if isinstance(exception, MemoryError):
        study.enqueue_trial(trial.params)
```

If we implement such a feature, we need to fix `_optimize.py` too.
https://github.com/optuna/optuna/blob/fe9909a04cb343fcb346a9f4cbcf4c1d9efc066f/optuna/_optimize.py#L155-L170

hvy:
> I think fail_reason in system_attrs is originally introduced to distinguish stochastic errors from deterministic ones.

Do you mean to do this systematically? I believe the point is that the stack traces are logged. So the question is whether it should be stored in the study in addition to that. Sorry, I feel like I'm missing your point though.

> BTW, how about adding an exception to the argument of callback functions?

Nice, this sounds like something we can work on in either case. Could be useful for doing cleanups and such.

toshihikoyanase:
> Do you mean to do this systematically?

Yes, I do.  If the error information is included in study in addition to the log, users may be able to automate postprocess such as rerun or cleanup more easily as follows: (see also this [example notebook](https://colab.research.google.com/drive/1rcP9ATrUG7ytzRYzAe9W7VGC6u99iTjh#scrollTo=Y5vkyKGAG-fD))

```python
for trial in study.trials:
    if trial.state != optuna.trial.TrialState.FAIL:
        continue
    if "MemoryError" in trial.system_attrs["fail_reason"]:
        study_rerun.enqueue_trial(trial.params)
```

HideakiImamura (approved):
I agree with the suggestion of this PR. LGTM.

himkt:
Although using `fail_reason` for catching specific exceptions feels somewhat useful, it would be better to add an exception to the argument of callback (suggested in https://github.com/optuna/optuna/pull/1964#issuecomment-717638884) for such cases, I think. We can work on it in another PR.

So, this PR looks good.
@hvy Could you please fix a conflict? After resolving a conflict, I'll merge it. 👍

hvy:
Thanks for the review, just resolved the conflicts, PTAL.

himkt (approved):
LGTM

PR #1965, opened 2020-10-27
Add Raises comments to samplers

yuk1ty:
## Motivation

- See issue #1618 for the original motivation.

## Description of the changes

Another addition of `Raises:` comments. I worked on `samplers` module as far as I found.

### DIFF
--- a/optuna/samplers/_cmaes.py
+++ b/optuna/samplers/_cmaes.py
@@ -142,6 +142,10 @@ class CmaEsSampler(BaseSampler):
                 to set this flag :obj:`True` when the :class:`~optuna.pruners.HyperbandPruner` is
                 used. Please see `the benchmark result
                 <https://github.com/optuna/optuna/pull/1229>`_ for the details.
+
+    Raises:
+        ValueError:
+            If ``restart_strategy`` is not 'ipop' or :obj:`None`.
     """
 
     def __init__(
--- a/optuna/samplers/_search_space.py
+++ b/optuna/samplers/_search_space.py
@@ -43,6 +43,10 @@ class IntersectionSearchSpace(object):
 
         Returns:
             A dictionary containing the parameter names and parameter's distributions.
+
+        Raises:
+            ValueError:
+                If different studies are passed into this method.
         """
 
         if self._study_id is None:


optuna/samplers/_search_space.py
```diff
@@ -43,6 +43,10 @@ def calculate(
 
         Returns:
             A dictionary containing the parameter names and parameter's distributions.
+
+        Raises:
+            ValueError:
+                If ``self._study_id`` is not equal to ``study._study_id``.
```
toshihikoyanase, line 49: I guess this sentence could be more intuitive. How about rephrasing it by using the expression in `ValueError` at L54? The following is just an example:

> If multiple studies are inputted into this method.

toshihikoyanase (changes requested):
Thank you for your PR as always. I have some small comments.

toshihikoyanase:
Let me replace the `code-fix` label with the `document` label since this PR only updates docstring.

optuna/samplers/_search_space.py
hvy: Definitely a nit but "different" might be clearer than "multiple". It's also what we use for describing this behavior in the docs of `__init__`.

> If different studies are passed ..

hvy:
Looks great to me besides the nit above and @toshihikoyanase's comments.

optuna/samplers/_search_space.py
toshihikoyanase: > If different studies are passed ..

@hvy Much better! Thank you for your suggestion.

hvy (approved):
LGTM!

optuna/samplers/_search_space.py
yuk1ty: More clear than before. Thank you!

optuna/samplers/_cmaes.py
```diff
@@ -142,6 +142,10 @@ def objective(trial):
                 to set this flag :obj:`True` when the :class:`~optuna.pruners.HyperbandPruner` is
                 used. Please see `the benchmark result
                 <https://github.com/optuna/optuna/pull/1229>`_ for the details.
+
+    Raises:
+        ValueError:
+            If ``restart_strategy`` is not 'ipop' or :obj: `None`.
```
toshihikoyanase, line 148: A nit, but could you remove a white space between `:obj:` and `None`?
```suggestion
            If ``restart_strategy`` is not 'ipop' or :obj:`None`.
```

image

toshihikoyanase (changes requested):
I found a minor error, so could you take a look?

optuna/samplers/_cmaes.py
yuk1ty: oh, sorry, thank you!

hvy: Sorry should've noticed as well. Thanks @toshihikoyanase and @yuk1ty .

toshihikoyanase (approved):

optuna/samplers/_cmaes.py
toshihikoyanase: No problem at all. We employ two-approval rule to help each other.

PR #1966, opened 2020-10-27
Fix `StudyDirection` of `mape` in `LightGBMTuner`

### LINKED ISSUE #1858: Metric MAPE direction should be reversed with reference to MAP

toshihikoyanase:
## Motivation
Fixes #1858. I found that `mape` `startswith` `map` when I read the code in #1947 . 

https://github.com/optuna/optuna/blob/9fc26c2277b347788d5554dad66dba81585716b2/optuna/integration/_lightgbm_tuner/optimize.py#L146

## Description of the changes
- Use `in` instead of `startswith` to check the metrics. I assumed that the metrics are converted by `_handling_alias_metrics`.
- Add `mape` to the test case.

### DIFF
--- a/optuna/integration/_lightgbm_tuner/optimize.py
+++ b/optuna/integration/_lightgbm_tuner/optimize.py
@@ -144,7 +144,7 @@ class _BaseTuner(object):
     def higher_is_better(self) -> bool:
 
         metric_name = self.lgbm_params.get("metric", "binary_logloss")
-        return metric_name.startswith(("auc", "ndcg", "map"))
+        return metric_name in ("auc", "ndcg", "map")
 
     def compare_validation_metrics(self, val_score: float, best_score: float) -> bool:
 
--- a/tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
+++ b/tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
@@ -150,7 +150,7 @@ class TestBaseTuner(object):
             tuner = _BaseTuner(lgbm_params={"metric": metric})
             assert tuner.higher_is_better()
 
-        for metric in ["rmsle", "rmse", "binary_logloss"]:
+        for metric in ["rmsle", "rmse", "binary_logloss", "mape"]:
             tuner = _BaseTuner(lgbm_params={"metric": metric})
             assert not tuner.higher_is_better()
 


hvy (approved):
Makes sense, LGTM!

optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -144,7 +144,7 @@ def _metric_with_eval_at(self, metric: str) -> str:
     def higher_is_better(self) -> bool:
 
         metric_name = self.lgbm_params.get("metric", "binary_logloss")
-        return metric_name.startswith(("auc", "ndcg", "map"))
+        return metric_name in ("auc", "ndcg", "map")
```
hvy, line 147: Orthogonal to this change, but a related comment. Should be add an assertion before this line to make sure we've in control of all possible values of `metric_name`, i.e. `assert metric_name in (...)`?

HideakiImamura: Looks good. I agree with @hvy.

HideakiImamura (approved):
Thanks for the PR. LGTM. I think we should follow @hvy's suggestion. I'd like to merge if the suggestion is reflated.

optuna/integration/_lightgbm_tuner/optimize.py
toshihikoyanase: Hmm, I think it is reasonable, but I feel it is out-of-scope of this PR.

LightGBM is being developed actively and we can see some new metrics like [`gamma` and `tweedie`](https://lightgbm.readthedocs.io/en/latest/Parameters.html#metric-parameters) since September (https://github.com/optuna/optuna/issues/1806#issuecomment-694243823). Such new metrics (`gamma` and `tweedie`) can be handled properly now since they are to be minimized, but the assertion errors will be raised if we apply your suggestion. It may be controversial and I'd like to discuss it in a new PR.

hvy: Sounds good to me.

HideakiImamura: Thanks for the explanation.

> It may be controversial and I'd like to discuss it in a new PR.

It's reasonable. Thanks.

HideakiImamura:
Let me merge this PR since the above point will be discussed in the separate PR.

PR #1978, opened 2020-10-30
Add RDB support to `MoTPEMultiObjectiveSampler`

toshihikoyanase:
## Motivation
This is a follow-up task of #1530. It fixes the error regarding MoTPE with RDB in this comment https://github.com/optuna/optuna/pull/1530/files#r503843954.

## Description of the changes

It make `trial.system_attrs` JSON serializable.

The example code is as follows:
```python
import optuna


def objective(trial):
    # Binh and Korn function.

    v0 = 4 * x ** 2 + 4 * y ** 2
    return v0, v1


study = optuna.multi_objective.create_study(
    ["minimize", "minimize"],
    sampler=optuna.multi_objective.samplers.MOTPEMultiObjectiveSampler(),
    storage="sqlite:///foo.db",
)

optuna.multi_objective.visualization.plot_pareto_front(study).show()
```

### DIFF
--- a/optuna/multi_objective/samplers/_motpe.py
+++ b/optuna/multi_objective/samplers/_motpe.py
@@ -402,13 +402,19 @@ class MOTPEMultiObjectiveSampler(TPESampler, BaseMultiObjectiveSampler):
             above = np.log(above)
 
         size = (self._n_ehvi_candidates,)
+
+        weights_below: Callable[[int], np.ndarray]
+
         if self._weights is _default_weights_above:
+
             weights_below = lambda _: np.asarray(
                 study._storage.get_trial(trial._trial_id).system_attrs[_WEIGHTS_BELOW_KEY],
                 dtype=float,
             )
+
         else:
             weights_below = self._weights
+
         parzen_estimator_parameters_below = _ParzenEstimatorParameters(
             self._parzen_estimator_parameters.consider_prior,
             self._parzen_estimator_parameters.prior_weight,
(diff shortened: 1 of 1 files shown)


optuna/multi_objective/samplers/_motpe.py
```diff
@@ -286,20 +286,21 @@ def _split_mo_observation_pairs(
...
 
-            attrs = {"indices_below": indices_below, "indices_above": indices_above}
+            attrs = {
+                "indices_below": indices_below.tolist(),
+                "indices_above": indices_above.tolist(),
+            }
             if self._weights is _default_weights_above:
                 weights_below = self._calculate_default_weights_below(lvals, indices_below)
-                attrs["weights_below"] = weights_below
+                attrs["weights_below"] = weights_below.tolist()
```
hvy, line 295: Should we cast these to lists prior to inserting them into the dict to make the types consistent with the case in which we get a cache hit?

hvy:
Noticed that this sampler breaks with a key error if we specify `weights` with some arbitrary function. It's outside the scope of this PR so we can probably tackle this later.

HideakiImamura:
> Noticed that this sampler breaks with a key error if we specify weights with some arbitrary function. It's outside the scope of this PR so we can probably tackle this later.

PR #1979 fixes this bug.

optuna/multi_objective/samplers/_motpe.py
toshihikoyanase: Thank you for your comment.
I think these casts are required since `ndarray` is not JSON serializable. I added type casting from `list` to `ndarray` when we deserialize the values from `system_attrs` to make the types consistent (see https://github.com/optuna/optuna/pull/1978/commits/9d0a82bb1e7dc9778995342f00b80945c18e2c93).

HideakiImamura (approved):
LGTM.

hvy (approved):
Thanks for your quick fix, LGTM!

Issue #1967, opened 2020-10-27
Strange contour plot is generated with latest `plotly`


HideakiImamura:
## Expected behavior
This bug is originally reported by @harupy in [here](https://github.com/optuna/optuna/pull/1944#issuecomment-716473083).

The example code in the documentation should generate the contour plot as follows.
<img width="1869" alt="スクリーンショット 2020-10-27 12 02 48" src="https://user-images.githubusercontent.com/38826298/97252000-581feb00-184c-11eb-948c-edae184e590b.png">

But, it generates as follows with the latest `plotly` of version 4.12.0.
<img width="1867" alt="スクリーンショット 2020-10-28 4 11 22" src="https://user-images.githubusercontent.com/38826298/97350100-aaf0b580-18d3-11eb-99be-a68c4e5eba17.png">

## Environment

- Optuna version: https://github.com/optuna/optuna/commit/fe9909a04cb343fcb346a9f4cbcf4c1d9efc066f
- Python version: 3.8
- Plotly version: 4.12.0

## Reproducible examples

```python
import optuna


def objective(trial):
    x = trial.suggest_uniform("x", -100, 100)
    y = trial.suggest_categorical("y", [-1, 0, 1])
    return x ** 2 + y


study = optuna.create_study()
study.optimize(objective, n_trials=30)

fig = optuna.visualization.plot_contour(study, params=["x", "y"])
print(fig)
fig.show()

```

## Additional context (optional)

With the `plotly` of version 4.9.0,  the above code generates the appropriate contour plot.

hvy:
Let me close this issue as fixed by #1983.

PR #1983, opened 2020-10-31
Fix the range of categorical values.

toshihikoyanase:
## Motivation
This PR addresses https://github.com/optuna/optuna/issues/1967.

## Description of the changes
This PR changes the plot range of categorical parameters for `plotly>=4.12.0`. As I mentioned in the comment, plotly seems to generate contours using indices of categorical values, but it determines plot ranges based on raw values. So, I updated the range based on the indices.

Code (from #1819)
```python
import optuna

def objective(trial):
    x = trial.suggest_categorical("x", ["10", "11", "100"])
    y = trial.suggest_categorical("y", ["sigmoid", "relu", "leakyrelu"])
    z = trial.suggest_loguniform("z", 1e-1, 1e2)
    return int("0" in x) + abs(len(y) - 5) + (z/50) ** 2

study = optuna.create_study()

optuna.visualization.plot_contour(study).show()
```

This PR:
image

The current master:
image

### DIFF
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -118,7 +118,7 @@ def _get_contour_plot(study: Study, params: Optional[List[str]] = None) -> "go.F
             # Plotly uses indices instead of raw values to generate contours, but the raw values
             # are used to calculate ranges since plotly==4.12.0.
             # See https://github.com/optuna/optuna/issues/1967.
-            # TODO (yanase): Remove this if-clause after resolving the error.
+            # TODO(toshihikoyanase): Remove this if-clause after resolving the error.
             if version.parse(plotly.__version__) >= version.parse("4.12.0"):
                 min_value = 0
                 max_value = len(set(values)) - 1
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -117,7 +117,7 @@ def _get_contour_plot(study: Study, params: Optional[List[str]] = None) -> "go.F
 
             # Plotly uses indices instead of raw values to generate contours, but the raw values
             # are used to calculate ranges since plotly==4.12.0.
-            # See https://github.com/optuna/optuna/pull/1944#issuecomment-716968285.
+            # See https://github.com/optuna/optuna/issues/1967.
             # TODO (yanase): Remove this if-clause after resolving the error.
             if version.parse(plotly.__version__) >= version.parse("4.12.0"):
                 min_value = 0
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -5,6 +5,8 @@ from typing import List
 from typing import Optional
 from typing import Tuple
 
+from packaging import version
+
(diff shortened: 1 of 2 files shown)


hvy (commented):
Thanks, basically LGTM. I checked locally with `4.12.0` and `<4.12.0`, with and without these  changes. Just wondering if it is difficult to keep the contours all the way to the edges as previously, [docs](https://optuna.readthedocs.io/en/v2.3.0/reference/visualization/generated/optuna.visualization.plot_contour.html)?

toshihikoyanase:
@hvy Sorry for the late reply.
I'm not sure why the new version does not render the edges. I personally feel we can work on it in a follow-up PR to fix the current document soon.

CC @harupy @ytknzw Please let me know if you have any comments on this issue.

optuna/visualization/_contour.py
```diff
@@ -109,15 +111,23 @@ def _get_contour_plot(study: Study, params: Optional[List[str]] = None) -> "go.F
...
             padding = (math.log10(max_value) - math.log10(min_value)) * padding_ratio
             min_value = math.pow(10, math.log10(min_value) - padding)
             max_value = math.pow(10, math.log10(max_value) + padding)
 
-        elif _is_categorical(trials, p_name):
-            # For numeric values, plotly does not automatically plot as "category" type.
-            update_category_axes[p_name] = any([str(v).isnumeric() for v in set(values)])
-
         else:
             padding = (max_value - min_value) * padding_ratio
```
hvy, line 132: This code is executed now with categoricals without numerical conversions for `plotly<=4.11.0`. I think it'll lead to an error.

toshihikoyanase: Thank you for pointing it out. I'll check it.

toshihikoyanase: As you mentioned it raised the error when I used `plotly==4.11.0`. I fixed the error in the commit https://github.com/optuna/optuna/pull/1983/commits/5ad0bbb803ab8557887bb0099eacf9b4a02c67f9.

toshihikoyanase:
@hvy @himkt I resolved the issue with `plotly<4.12.0`. PTAL!

hvy (approved):
Thanks for the fix. Quickly verified this with different plotly versions locally. LGTM.

It wasn't entire clear to me what the error is referring to in the TODO comment, though.

toshihikoyanase:
@hvy I read the comment again, and I don't think it's a TODO comment. So, I revised it as follows in https://github.com/optuna/optuna/pull/1983/commits/ddf7cfd9f2cd8a588ef3e818acdd62feef6aa2ff:

> Plotly>=4.12.0 draws contours using the indices of categorical variables instead of
> raw values and the range should be updated based on the cardinality of categorical
> variables. See https://github.com/optuna/optuna/issues/1967.

himkt (approved):
Thank you for the update, LGTM.

> I'm not sure why the new version does not render the edges.

Let's modify it when we find the cause.

PR #1987, opened 2020-11-03
Universal data transformer

hvy:
## Motivation

Optuna handles various types of parameters including categorical parameters and numerical values in log space. Conversions to uniform space is sometimes required such as in the CMA-ES samplers and in the HPI features. A universal transformer would be useful for this kind of data conversion.

## Description of the changes

Introduces a data transformer to "warp" Optuna parameters and search spaces to uniform space, and vice versa.

## Note

- ~This transformer assumes "relative sampling", i.e. all parameters must be known. We might be able to relax this assumptions to allow it to be used for "independent sampling" if that would be useful.~(edit: Addressed and `RandomSampler` can now be simplified as follows https://github.com/optuna/optuna/commit/947d12d3e341be1f5ca68974e0cfcd8080f4d511)
- ~Requires scikit-learn for one-hot encoding. This might make it controversial, or optional~ (edit: reimplemented without scikit-learn)
- `BoTorchSampler.sample_relative` in https://github.com/optuna/optuna/pull/1989 can be simplified with this transformer as well ~(botorch requires gpytorch which already requires scikit-learn)~.

## TODO

- [x] Iterate on API
- [x] Pass CI

## Updates

Whether to apply log transformation, discretization offsets to the bounds can be controlled now. This allows `_Transform` to be used from both samplers which require these transformations as well as HPI which don't.


hvy:
Ready for review. Updated the description.

hvy:
@c-bata would you like to have a look at this PR? Similar to how it can be used for wrapping BoTorch, I'm wondering if we cannot use it in the `CmaEsSampler` to support categorical distributions.

c-bata:
I see! I'll review this 👍

hvy:
Thanks. I left some documentation in the code as well since it's likely more difficult to understand otherwise.

tests/test_transform.py
```diff
@@ -0,0 +1,250 @@
...
+
+    trans_params = trans.transform({"x0": param})
+
+    for trans_param in trans_params:
+        assert trans_param in (0.0, 1.0)
+
+
+@pytest.mark.parametrize("transform_log", [True, False])
+@pytest.mark.parametrize("transform_step", [True, False])
+def test_search_space_transform_shapes_dtypes_values_categorical_with_other_distribution(
```
c-bata, line 143: It is not easy to understand what this test method checks. I found code smells of "Conditional Test Logic" anti-pattern which is introduced by xUnit test patterns.
http://xunitpatterns.com/Conditional%20Test%20Logic.html

Could you make this test case more simple?

tests/test_transform.py
```diff
@@ -0,0 +1,250 @@
...
+        expected_bounds_shape = (1, 2)
+        expected_params_shape = (1,)
+    assert trans.bounds.shape == expected_bounds_shape
+    assert trans.bounds.dtype == numpy.float64
+    assert trans_params.shape == expected_params_shape
+    assert trans_params.dtype == numpy.float64
+
+
+@pytest.mark.parametrize("transform_log", [True, False])
+@pytest.mark.parametrize("transform_step", [True, False])
```
c-bata, line 54: How about separating a test method that checks these parameters.
It makes the test cases simple even though it increases the test code lines.

tests/test_transform.py
```diff
@@ -0,0 +1,250 @@
...
+        assert bound[1] == 1.0
+
+    trans_params = trans.transform({"x0": param})
+
+    for trans_param in trans_params:
+        assert trans_param in (0.0, 1.0)
+
+
+@pytest.mark.parametrize("transform_log", [True, False])
+@pytest.mark.parametrize("transform_step", [True, False])
```
c-bata, line 142: ditto.

c-bata (changes requested):
Thank you for the updates 👍  
At a glance, the implementation is clear but the test codes are bit complicated. I put some comments to make the test cases simple.

tests/test_transform.py
hvy: I couldn't understand this comment but do you mean to write two separate test functions, one for `transform_log` and one for `transform_step`?

tests/test_transform.py
hvy: Sure, we could for instance unroll the loops and perhaps skip the parameterization. Or actually, we could remove this test altogether. In the initial implementation using scikit-learn, categorical one-hot encoding were always placed in the first rows of the bounds ndarray. I wanted to check that the bounds were still correct.

tests/test_transform.py
c-bata: Yes. This is not a strong opinion.
It makes the tests simple and error messages will be clear.

hvy: I actually think we should test them and their permutations together. 

On a side note, we could probably omit this parameterizations from the other test functions. That might make the entire test file more comprehensible. What do you think?

hvy: Pushed a change with that, PTAL.

hvy:
PTAL.

tests/test_transform.py
hvy: Removed the entire function.

optuna/_transform.py
```diff
@@ -0,0 +1,271 @@
...
+            param = params[name]
+
+            if isinstance(distribution, CategoricalDistribution):
+                choice_idx = distribution.to_internal_repr(param)
+                trans_params[bound_idx + choice_idx] = 1
+                bound_idx += len(distribution.choices)
+            else:
+                trans_params[bound_idx] = _transform_numerical_param(
+                    param, distribution, self._transform_log
+                )
```
HideakiImamura, line 109: Currently, the transformation logic of the search space and the parameter value is different. I'm not sure but I think we should unify them. However, the current logic of those transformations is different in `CmaEsSampler`. @hvy @c-bata What do you think?

HideakiImamura (commented):
Thanks for the PR. I have several minor comments and questions. Please take a look.

optuna/_transform.py
HideakiImamura: I noticed that the transformation of the search space of this PR is used in `CmaEsSampler`.

optuna/_transform.py
```diff
@@ -0,0 +1,271 @@
...
+                bds = (
+                    _transform_numerical_param(d.low, d, transform_log) - half_step,
+                    _transform_numerical_param(d.high, d, transform_log) + half_step,
+                )
+            elif isinstance(d, IntLogUniformDistribution):
+                half_step = 0.5 if transform_step else 0.0
+                bds = (
+                    _transform_numerical_param(d.low - half_step, d, transform_log),
+                    _transform_numerical_param(d.high + half_step, d, transform_log),
+                )
```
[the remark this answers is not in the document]
hvy: I think it's difficult given that the steps are applied differently depending on the distribution, more precisely `IntLogUniformDistribution`. Could you elaborate on how?

optuna/_transform.py
hvy: > the transformation logic of the search space and the parameter value is different. I'm not sure but I think we should unify them.

Sorry, I couldn't understand exactly in what way you meant that they are different. Could you explain how we could do the transformation instead?

hvy:
Thanks @HideakiImamura for your feedback. I applied your suggestion and tried answering the comments, PTAL.

optuna/_transform.py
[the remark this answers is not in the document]
HideakiImamura: I thought the following implementation of `_transform_numerical_param`. I'm suggesting this because I think it would be easier to read the code if the case for distribution is localized. However, this is my preference, so I think there is a room of discussion.
```python
def _transform_numerical_param(
    param: Union[int, float], 
    distribution: BaseDistribution, 
    transform_log: bool, 
    transform_step: bool, 
    position: str
) -> float:
    d = distribution
    assert position is in ("low", "high", "inside")
    
    ...
    if isinstance(d, DiscreteUniformDistribution):
        if position == "low"
            extended_step = -0.5 * d.q if transform_step else 0.0
        elif position == "high"
            extended_step = 0.5 * d.q if transform_step else 0.0
        else:
            extended__step = 0
        
        trans_param = float(param) + extended_step
    elif ....
```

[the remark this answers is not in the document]
hvy: Oh, thank you for the detailed code. I now see what you mean, passing the position as well. It might make `_transform_search_space` more readable but I'm not sure it contributes to the overall readability. `_transform_numerical_param` is obviously more complicated and moreover it's no longer symmetrical to `_untransform_numerical_param`.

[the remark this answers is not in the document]
HideakiImamura: OK, I see. Let me resolve this conversation without any fixes.

optuna/_transform.py
HideakiImamura: Sorry for the confusion. I was wrong about the difference. Let me resolve this conversation.

tests/test_transform.py
```diff
@@ -0,0 +1,157 @@
...
+        assert bound[1] == expected_high
+
+    trans_params = trans.transform({"x0": param})
+    assert trans_params.size == 1
+
+    if isinstance(distribution, (IntUniformDistribution, IntLogUniformDistribution)):
+        assert expected_low <= trans_params <= expected_high
+    else:
+        # TODO(hvy): Change second `<=` to `<` when `suggest_float` is fixed.
+        assert expected_low <= trans_params <= expected_high
```
HideakiImamura, line 104: [just a question] What should we do to address this TODO comment?

optuna/_transform.py
```diff
@@ -0,0 +1,271 @@
...
+            else:
+                assert False, "Should not reach. Unexpected distribution."
+
+            bounds[bound_idx] = bds
+            column_to_encoded_columns.append(numpy.atleast_1d(bound_idx))
+            bound_idx += 1
+        else:
+            assert False, "Should not reach. Unexpected distribution."
```
HideakiImamura, line 215: How about testing to reach this `False` assertion with an unexpected distribution?

optuna/_transform.py
```diff
@@ -0,0 +1,271 @@
...
+    elif isinstance(d, LogUniformDistribution):
+        trans_param = math.log(param) if transform_log else float(param)
+    elif isinstance(d, DiscreteUniformDistribution):
+        trans_param = float(param)
+    elif isinstance(d, IntUniformDistribution):
+        trans_param = float(param)
+    elif isinstance(d, IntLogUniformDistribution):
+        trans_param = math.log(param) if transform_log else float(param)
+    else:
+        assert False, "Should not reach. Unexpected distribution."
```
HideakiImamura, line 240: Ditto.

optuna/_transform.py
```diff
@@ -0,0 +1,271 @@
...
+        )
+    elif isinstance(d, IntUniformDistribution):
+        param = int(numpy.round((trans_param - d.low) / d.step) * d.step + d.low)
+    elif isinstance(d, IntLogUniformDistribution):
+        if transform_log:
+            param = int(min(max(numpy.round(math.exp(trans_param)), d.low), d.high))
+        else:
+            param = int(trans_param)
+    else:
+        assert False, "Should not reach. Unexpected distribution."
```
HideakiImamura, line 269: Ditto.

HideakiImamura (commented):
Thanks for the update. I have a question and minor comments around tests. PTAL.

optuna/_transform.py
```diff
@@ -0,0 +1,271 @@
...
+            if isinstance(d, UniformDistribution):
+                bds = (
+                    _transform_numerical_param(d.low, d, transform_log),
+                    _transform_numerical_param(d.high, d, transform_log),
+                )
+            elif isinstance(d, LogUniformDistribution):
+                bds = (
+                    _transform_numerical_param(d.low, d, transform_log),
+                    _transform_numerical_param(d.high, d, transform_log),
+                )
```
c-bata, line 189: Now I remembered https://github.com/optuna/optuna/pull/1558. UniformDistribution and LogUniformDistribution only accept `[low, high)`. But some optimizer (e.g. pycma and cmaes) samples the upper bound value, then it will be fallback to the independent sampler.

We have some solutions for this problem.

1. Add a `_SearchSpaceTransform`'s option to cap the upper bound with a tiny value
1. Always cap the upper bound in `_SearchSpaceTransform` (without an option).
1. Cap the upper bound in CmaEsSampler and PyCmaSampler.

I think the second solution is enough for all samplers. But it seems a controversial topic.
May I have your opinion?

tests/test_transform.py
c-bata: Thank you for the update. It looks current test case is enough simple 👍  Good refactoring.

tests/test_transform.py
hvy: Noticed it's already addressed in the sense that `trans_params` is hard-coded instead of being suggested by a study. Fixed the code and removed the comment. Thanks!

optuna/_transform.py
hvy: We'll be testing the internals of something that's already internal. Seems a bit overkill to be honest.

optuna/_transform.py
hvy: You're right. I'm wondering if we shouldn't go with 3 in this PR to keep it reasonably scoped. However, I agree that 2 sounds like a fair compromise otherwise. We'd be able to isolate this somewhat controversial logic to a single point.

optuna/_transform.py
HideakiImamura: I'm not sure about this idea, but shouldn't we add the test to increase test coverage?

hvy: Hm, yes but this is not just any code but code that should never be reached unless internals are broken 🤔

HideakiImamura: Yes, you are right. This code should never be reached. I'm not intended to insist on my opinion, so let me resolve this conversation.

HideakiImamura (approved):
Thanks for the confirmation and for addressing my comments. LGTM.

optuna/_transform.py
hvy: @c-bata what do you think?

c-bata: > You're right. I'm wondering if we shouldn't go with 3 in this PR to keep it reasonably scoped. However, I agree that 2 sounds like a fair compromise otherwise. We'd be able to isolate this somewhat controversial logic to a single point.

Agree. The only concern is that all upper bounds (including categorical, int, and discrete) will be capped if we choose 3. But I guess it does not affect the optimization performance. I made this resolved.

(Sorry, I'm busy today. So I'll review other changes next week.)

hvy: (Thanks a lot. Don't mean to stress but do you have any estimates on when? If you're busy, we could also invite another reviewer or if your time is limited maybe schedule a mob session to reduce the number of roundtrips.)

HideakiImamura:
This PR is a blocker for #1988 and #1989. Normally Optuna has a two-approval system, but since this is a high importance PR and all the points raised by @c-bata have been corrected, let me merge it with my 1 approval.

If you have any other concerns about the `optuna/_transform.py` added by this PR, please feel free to raise an issue or DM us. We rely heavily on your usual insights in reviewing PRs, and this action is only meant to help us move forward with urgent PRs.

Issue #1986, opened 2020-11-03
allow change of user attributes even when trial is finished


jtdeng:
In some blackbox optimization, we need to set trial related user attributes even when it's finished, e.g. trial 2 is sampled around trial 1 (already finished), and when trial 2 is finished, we need to evaluate how good trial 2 can improve the objective, thus we will set a score (user attr) to trial 1.

in current implementation, when a trial is finished, there is a check to prevent any change after trial is finished, it would be good that optuna can allow the change on user defined attrs.

hvy:
Thanks for bringing this up. I'm not entire sure I've understood your use case, but I assume it is not possible to store this information in the attribute of trial 2? If you have code, that might help. Just to clarify there are certain assumptions introduced around the behavior of how trials may be modified. E.g. in this case that a trial can only be modifies before it's completed, and that a trial is never modified from a different process that the one that first started it. It's mostly an implementation detail, though. To elaborate, Optuna internally uses a cached storage on top of the RDB storage. This cached storage caches trials that have been completed and assumes that those are fixed throughout. We'll have to reconsider this behavior as well if this is an issue.

jtdeng:
> Thanks for bringing this up. I'm not entire sure I've understood your use case, but I assume it is not possible to store this information in the attribute of trial 2? If you have code, that might help. Just to clarify, there are certain assumptions introduced around the behavior of how trials may be modified. E.g. that a trial can only be modified before it's completed, and that a trial is never modified from a different process than the one that first started it. It's mostly an implementation detail, though. To elaborate, Optuna internally uses a cached storage on top of the RDB storage to reduce RDB access. This cached storage caches trials that have been completed and assumes that those are fixed throughout. We'll have to reconsider this behavior if this is an issue.

Hi hvy, thanks for your quick reply.
I understand the current design that no change is allowed when trial is finished, [_check_trial_is_updatable](https://github.com/optuna/optuna/blob/750456823d72b6fe7cf66717f3e3107a54c73307/optuna/storages/_cached_storage.py#L417), while in our sampler algorithm (WIP), we need to pick a trial (trial 1 mentioned above) from all completed trials and sample a candidate (say trial 2) around it, after evaluation of trial 2, we need to set some user defined attrs related to trial 1, thus all other distributed optimization processes could reuse this information in db to judge which trial could be picked in next iteration. these user attributes shall be trial related and can only be set after trial is completed.

[...]

PR #1989, opened 2020-11-03
Introduce `BoTorchSampler`

hvy:
##  Motivation

BoTorch provides a very flexible interface for BO. This PR aims to provide a wrapper, allowing users of Optuna to utilize BoTorch for their HPO and vice versa.

## Description of the changes

Introduces a `BoTorchSampler` integration sampler.

## Note

Depends on https://github.com/optuna/optuna/pull/1987 in order to support all distributions in Optuna.

## TODO

- [x] Iterate on API
- [x] Check CI configurations for dependencies, examples and tests
- [x] Add tests
- [x] Documentation
- [x] Finish example
- [x] Clean up internals of sampler
- [x] Verify that optimization works as expected


HideakiImamura:
#1987 has been merged. Could you merge the master branch?

hvy:
Master merge done.

hvy:
I'll be uploading benchmarks (Jupyter notebooks) to my fork. With the last commit: https://github.com/hvy/optuna/commit/e7bcfa5b042e8bf02083059bda683d991e204df2.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,507 @@
...
+            larger than 0 means that a constraints is violated. A value equal to or smaller than 0
+            is considered feasible.
+
+            If omitted, no constraints will be passed to ``candidates_func`` nor taken into
+            account during suggestion if ``candidates_func`` is omitted.
+
+            .. note::
+                ``constraints_func`` is called once per trial for each trial on each worker during
+                distributed optimization. Therefore, during distributed optimization, this
+                function should be deterministic to ensure that all workers hold the same values.
```
hvy, line 323: We might want to introduce some sort of trial post-processing to allow this in the future. C.f. https://github.com/optuna/optuna/issues/1986

hvy:
PTAL.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,507 @@
...
+        sequential=True,
+    )
+
+    candidates = unnormalize(candidates.detach(), bounds=bounds)
+
+    return candidates
+
+
+def _get_default_candidates_func(
+    n_objectives: int, n_objective_constraints: int
```
HideakiImamura, line 246: How about `n_objective_constraints` -> `n_constraints`?

HideakiImamura (commented):
Thanks for the great PR! I have several comments around `candidate_func`. Please take a look.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,507 @@
...
+    from botorch.optim import optimize_acqf
+    from botorch.sampling.samplers import SobolQMCNormalSampler
+    from botorch.utils.multi_objective.box_decomposition import NondominatedPartitioning
+    from botorch.utils.transforms import unnormalize
+    from gpytorch.mlls import ExactMarginalLogLikelihood
+    import torch
+
+
+@experimental("2.4.0")
+def ucb_candidates_func(
```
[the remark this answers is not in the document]
hvy: Sounds reasonable, thanks for the suggestion. I'll remove the UCB altogether.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,507 @@
...
+
+        additional_qehvi_kwargs = {}
+
+    model = SingleTaskGP(train_x, train_y, outcome_transform=Standardize(m=train_y.shape[-1]))
+    mll = ExactMarginalLogLikelihood(model.likelihood, model)
+    fit_gpytorch_model(mll)
+
+    partitioning = NondominatedPartitioning(num_outcomes=n_outcomes, Y=partitioning_y)
+
+    ref_point = train_obj.amin(dim=0) - 1e-8
```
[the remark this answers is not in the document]
hvy: Fixed to use `torch.Tensor.min`.

examples/multi_objective/botorch_simple.py
```diff
@@ -0,0 +1,48 @@
+from botorch.test_functions.multi_objective import C2DTLZ2
```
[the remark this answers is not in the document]
hvy: Makes sense but I think it's better to do it with the other MO examples.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,507 @@
...
+    Args:
+        train_x:
+            Previous parameter configurations. A ``torch.Tensor`` of shape
+            ``(n_trials, n_params)``. ``n_trials`` is the number of already observed trials
+            and ``n_params`` is the number of parameters. ``n_params`` may be larger than the
+            actual number of parameters if categorical parameters are included in the search
+            space, since these parameters are one-hot encoded.
+            Values are not normalized.
+        train_obj:
+            Objectives. A ``torch.Tensor`` of shape ``(n_trials, n_objectives)``. ``n_trials``
```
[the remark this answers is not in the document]
hvy: Thanks, fixed.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,507 @@
...
+    with single-objective optimization without objective constraints.
+
+    Args:
+        train_x:
+            Previous parameter configurations. A ``torch.Tensor`` of shape
+            ``(n_trials, n_params)``. ``n_trials`` is the number of already observed trials
+            and ``n_params`` is the number of parameters. ``n_params`` may be larger than the
+            actual number of parameters if categorical parameters are included in the search
+            space, since these parameters are one-hot encoded.
+            Values are not normalized.
```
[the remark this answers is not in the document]
hvy: Well, it's up to the callback implementation. BoTorch generally assumes normalized inputs. I added calls to `normalize` in the default callbacks by the way.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,507 @@
...
+            Objectives. A ``torch.Tensor`` of shape ``(n_trials, n_objectives)``. ``n_trials``
+            is identical to that of ``train_x``. ``n_objectives`` is the number of objectives.
+            Observations are not normalized.
+        train_con:
+            Objective constraints. A ``torch.Tensor`` of shape ``(n_trials, n_constraints)``.
+            ``n_trials`` is identical to that of ``train_x``. ``n_constraints`` is the number of
+            constraints. A constraint is violated if strictly larger than 0. If no constraints are
+            involved in the optimization, this argument will be :obj:`None`. In particular for
+            :func:`~optuna.integration.botorch.ucb_candidates_func`, it should always be
+            :obj:`None` since UCB does not support constrained optimization.
```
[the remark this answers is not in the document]
hvy: Removed the default UCB callback.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,507 @@
...
+
+
+@experimental("2.4.0")
+def qei_candidates_func(
+    train_x: "torch.Tensor",
+    train_obj: "torch.Tensor",
+    train_con: Optional["torch.Tensor"],
+    bounds: "torch.Tensor",
+) -> "torch.Tensor":
+    """Expected Improvement (qEI).
```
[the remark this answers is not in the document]
hvy: Thanks, fixed this line as well as the corresponding one for qEHVI.

optuna/integration/botorch.py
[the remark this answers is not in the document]
hvy: Removed, PTAL.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,507 @@
...
+
+    .. seealso::
+        :func:`~optuna.integration.botorch.ucb_candidates_func` for argument and return value
+        descriptions.
+    """
+
+    if train_obj.size(-1) != 1:
+        raise ValueError("Objective may only contain single values with qEI.")
+    if train_con is None:
+        raise ValueError("Constraints must be used with qEI.")
```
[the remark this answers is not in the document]
hvy: Fixed, and removed UCB.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,507 @@
...
+
+
+@experimental("2.4.0")
+def qehvi_candidates_func(
+    train_x: "torch.Tensor",
+    train_obj: "torch.Tensor",
+    train_con: Optional["torch.Tensor"],
+    bounds: "torch.Tensor",
+) -> "torch.Tensor":
+    """Expected Hypervolume Improvement (qEHVI).
```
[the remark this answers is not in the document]
hvy: Fixed.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,507 @@
...
+
+    Returns:
+        Next set of candidates. Usually the return value of BoTorch's ``optimize_acqf``.
+
+    """
+
+    if train_con is not None:
+        warnings.warn("Constraints are given but will be ignored.")
+
+    model = SingleTaskGP(train_x, train_obj)
```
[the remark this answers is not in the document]
hvy: Fixed by calling `normalize` on `train_x`.

optuna/integration/botorch.py
hvy: Removed this parameter since it was no longer necessary, now that we only condition on the number of objectives and choose between qEI and qEHVI.

hvy:
Thanks @HideakiImamura for your comments. They should be addressed now, PTAL.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,468 @@
...
+        n_trials = len(trials)
+        if n_trials < self._n_startup_trials:
+            return {}
+
+        trans = _SearchSpaceTransform(search_space)
+
+        values = numpy.empty((n_trials, study.n_objectives), dtype=numpy.float64)
+        params = numpy.empty((n_trials, trans.bounds.shape[0]), dtype=numpy.float64)
+        if self._constraints_func is not None:
+            n_objective_constraints = len(next(iter(self._trial_constraints.values())))
```
HideakiImamura, line 390: I preferred `n_constraints` to `n_objective_constraints` since it is shorter and the constraints are also for the input.

HideakiImamura (commented):
Thanks for the update. I have other comments. PTAL.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,468 @@
...
+            Previous parameter configurations. A ``torch.Tensor`` of shape
+            ``(n_trials, n_params)``. ``n_trials`` is the number of already observed trials
+            and ``n_params`` is the number of parameters. ``n_params`` may be larger than the
+            actual number of parameters if categorical parameters are included in the search
+            space, since these parameters are one-hot encoded.
+            Values are not normalized.
+        train_obj:
+            Previously observed objectives. A ``torch.Tensor`` of shape
+            ``(n_trials, n_objectives)``. ``n_trials`` is identical to that of ``train_x``.
+            ``n_objectives`` is the number of objectives. Observations are not normalized.
```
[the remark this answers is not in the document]
hvy: They are not normalized when given to the callback, it's up the implementation of the callback to normalize/standardize. We do this for the default provided callbacks. I think this is better since one familiar with BoTorch enough to implement his/her own callback should be aware of how to do this, and this documentation just aids along the way.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,468 @@
...
+            Hypervolume Improvement (qEHVI) is used.
+
+            The function should assume *maximization* of the objective.
+
+            .. seealso::
+                See :func:`optuna.integration.botorch.qei_candidates_func` for an example.
+        constraints_func:
+            An optional function that computes the objective constraints. It must take a
+            :class:`~optuna.multi_objective.study.MultiObjectiveStudy`, a
+            :class:`~optuna.multi_objective.trial.FrozenMultiObjectiveTrial` and return the
```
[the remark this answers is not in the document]
hvy: I aligned it with the callback to `study.optimize`. I'm not sure if it is needed. I do think it's easier to remember the APIs if these callbacks are aligned, but it's open for discussion.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,468 @@
...
+
+    def sample_relative(
+        self,
+        study: "multi_objective.study.MultiObjectiveStudy",
+        trial: "multi_objective.trial.FrozenMultiObjectiveTrial",
+        search_space: Dict[str, BaseDistribution],
+    ) -> Dict[str, Any]:
+        assert isinstance(search_space, OrderedDict)
+
+        trials = [t for t in study.get_trials(deepcopy=False) if t.state == TrialState.COMPLETE]
```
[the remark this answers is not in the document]
hvy: Good point. How about doing that in a follow up? I think we should accompany that with some experimental results.

optuna/integration/botorch.py
hvy: Ah, thanks, fixed.

hvy:
PTAL.

optuna/integration/botorch.py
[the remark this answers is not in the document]
HideakiImamura: I understood. Let me resolve this conversation.

optuna/integration/botorch.py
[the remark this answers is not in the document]
HideakiImamura: Let me resolve this conversation since I found it reasonable from https://github.com/optuna/optuna/pull/1989#discussion_r539112185.

optuna/integration/botorch.py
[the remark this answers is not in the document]
HideakiImamura: Hmm... Since `study.optimize` represents the entire optimization, I think it's natural for that callback to have `study`. On the other hand, since `BoTorchSampler.sample_relative` is a method that is only called in one trial, it seems unnatural for the callback to have `study`. I'd like to hear from other reviewers' opinions as well.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,467 @@
...
+
+        additional_qehvi_kwargs = {}
+
+    train_x = normalize(train_x, bounds=bounds)
+
+    model = SingleTaskGP(train_x, train_y, outcome_transform=Standardize(m=train_y.shape[-1]))
+    mll = ExactMarginalLogLikelihood(model.likelihood, model)
+    fit_gpytorch_model(mll)
+
+    partitioning = NondominatedPartitioning(num_outcomes=n_outcomes, Y=partitioning_y)
```
HideakiImamura, line 184: According to the [document](https://botorch.org/api/utils.html#botorch.utils.multi_objective.box_decomposition.NondominatedPartitioning),
> Note: it is only feasible to use this algorithm to compute an exact decomposition of the non-dominated space for m<5 objectives (alpha=0.0). 

How about adding a note: `The number of objectives should be smaller than 5 due to botorch.utils.multi_objective.box_decomposition.NondominatedPartitioning` or a TODO comment here?

optuna/integration/botorch.py
```diff
@@ -0,0 +1,467 @@
...
+    elif n_objectives > 1:
+        return qehvi_candidates_func
+    else:
+        assert False, "Should not reach."
+
+
+# TODO(hvy): Allow utilizing GPUs via some parameter, not having to rewrite the callback
+# functions.
+@experimental("2.4.0")
+class BoTorchSampler(BaseMultiObjectiveSampler):
```
HideakiImamura, line 237: Cannot be this sampler use with a single `Study` and `FrozenTrial`? It seems that this sampler is only available with the multi-objective module, but we will integrate the multi-objective module with Optuna itself, so this is not a problem after the integration.

HideakiImamura (commented):
Thanks for the update. I have other several comments. Please check them.

optuna/integration/botorch.py
hvy: Yes, we could wait with merging this PR until after https://github.com/optuna/optuna/pull/1994. Then, I could update this sampler to a standard sampler.

examples/multi_objective/botorch_simple.py
```diff
@@ -0,0 +1,54 @@
...
+from botorch.settings import validate_input_scaling
+from botorch.test_functions.multi_objective import C2DTLZ2
+import torch
+
+import optuna
+
+
+# C2-DTLZ2 minimization objective function with a single constraint.
+# It is negated to be a maximization problem since BoTorch otherwise assumes maximization.
+_OBJECTIVE = C2DTLZ2(dim=3, num_objectives=2, negate=True)
```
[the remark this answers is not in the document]
hvy: Ah, it's because you haven't change the `directions` when creating the study. I changed the example to automatically adjust the length of the argument.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,467 @@
...
+
+    The default value of ``candidates_func`` in :class:`~optuna.integration.BoTorchSampler`
+    with multi-objective optimization.
+
+    .. seealso::
+        :func:`~optuna.integration.botorch.qei_candidates_func` for argument and return value
+        descriptions.
+    """
+
+    n_outcomes = train_obj.size(-1)
```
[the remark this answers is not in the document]
hvy: Fixed.

optuna/integration/botorch.py
eytan: I would recommend defaulting to qParEGO instead of qEHVI for any more than 3 objectives, otherwise it can get very slow on a CPU (particularly if you aren't using approximate hypervolume).

sdaulton: We have a heuristic thatfor selecting the approximation level for the partitioning in Ax that may be of interest: https://github.com/facebook/Ax/blob/master/ax/models/torch/botorch_moo_defaults.py#L303 The heuristic is based on the optimization performance and wall time results in our paper (Sec F.4): https://arxiv.org/abs/2006.05078

eytan:
@hvy — it's awesome to see that you guys are working on BoTorch integration!  Please don't hesitate to reach out to the BoTorch / Ax team if you have any questions or could use some feedback.  I can't tell from the diff if this includes the whole `BoTorchSampler`, but we'd happy to provide a review when things are ready.

FYI, we have a good deal of logic in Ax to do various data transformations and algorithm selection. Without these transformations, BoTorch may not be as effective.  We can do the best that we can with providing some review here, but it might be worth considering interfacing with Ax rather than BoTorch directly.  This way you don't have to try to mimic all of the existing logic, and you will have automatic parity with the latest improvements (for which much of the logic exists at the Ax level).

optuna/integration/botorch.py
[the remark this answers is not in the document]
eytan: I am not super familiar with the optuna internals, but there are two things that might be worth considering.
1. If you have multiple concurrent processes and optuna is asked to generate a new HP config, it's important that you let BoTorch know that you have "pending points" that are awaiting computation.
2. Adding partial evaluations to the dataset without telling BoTorch that the data is from a partial evaluation will cause problems with the model.  If you wanted to include partial evaluation, you'd want to include another input dimension that tells you, for example, the % of the training that was completed before the results were returned, and probably use a different acquisition function, like [KG](https://github.com/facebook/Ax/blob/master/ax/models/torch/botorch_kg.py) or MES.

optuna/integration/botorch.py
hvy: The heuristic for alpha in Ax looks very practical. Thanks for the advice. As for qParEGO, I actually skipped it on purpose to keep this PR reasonably small, but given that it's pointed out by the author and @HideakiImamura, I should perhaps rethink that. Just have to run some more benchmarks to make sure it runs as expected, but let me work on those changes. Again, thanks a lot for your insightful comments.

hvy:
@eytan, @sdaulton thanks for the comments. Inputs from the authors really help and are definitely more than welcome. To give some background, this `BoTorchSampler` is an experimental feature to provide customizable/performant BO, while reserving ourselves for future changes such as introducing parameter constraints and/or multi fidelity optimization, etc. This PR in any case aims to cover the most common use cases. Let me know if you think that something is missing or looks less pretty.

As for using Ax, that might be a viable option as well. I understand that BoTorch has a lot of knobs to tune and leaving that to Ax, well has pros and cons. One benefit of wrapping BoTorch over Ax is that we maintainers first get to learn BoTorch. If the API of this sampler matures and use cases become more clear (you can probably help us here), whether we directly use BoTorch, or via Ax, would become more of an implementation detail and we could perhaps reconsider this choice.

hvy:
Bench including qParEGO, with the last commit.

Screen Shot 2020-12-14 at 23 13 39

(Notebook which is too large to open on GitHub https://github.com/optuna/optuna/commit/161e8bd867918e9014e006cf08b77c0586de690a)

eytan:
@hvy , for your latest plot, what are the comparisons to?  It looks like you need to do many more repeats to measure any type significant differences, although hopefully there are no differences between the optuna and base versions of each?

examples/multi_objective/botorch_simple.py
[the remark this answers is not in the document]
HideakiImamura: I confirmed that the example can be executed with various pairs of `dim` and `n_objectives`. Let me resolve this conversation.

optuna/integration/botorch.py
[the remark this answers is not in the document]
HideakiImamura: @eytan Thanks for your insightful comments.

1. Optuna will only tell BoTorch about trials that have already been completed, so BoTorch will have less information available than if it were completely sequential, but we think this is a compromise we should accept. Other samplers make the same compromise.
2. The idea of adding another dimension to the objective function to represent the progress of the optimization seems very reasonable to me. This is one of the so-called multi-fidelity optimization approaches, and I think it would be a similar device to the https://arxiv.org/abs/1605.07079. (It's "contextual optimization" as it's called in bandit problems and Bayesian optimization.)

In any case, I agree with the opinion that consideration of these problems should be a follow-up.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,550 @@
...
+            Callable[
+                [
+                    "MultiObjectiveStudy",
+                    "FrozenMultiObjectiveTrial",
+                ],
+                Sequence[float],
+            ]
+        ] = None,
+        n_startup_trials: int = 10,
+        independent_sampler: Optional[BaseMultiObjectiveSampler] = None,
```
HideakiImamura, line 401: For future extensibility, how about making these arguments "keyword only"? In the future, we may utilize more features of BoTorch and add more arguments, such as batch size for batch optimization, `fidelity_func` for multi-fidelity optimization, etc.

optuna/integration/botorch.py
```diff
@@ -0,0 +1,550 @@
...
+
+        candidates = candidates.numpy()
+
+        params = trans.untransform(candidates)
+
+        # Exclude upper bounds for parameters that should have their upper bounds excluded.
+        # TODO(hvy): Remove this exclusion logic when it is handled by the data transformer.
+        for name, param in params.items():
+            if isinstance(search_space[name], (UniformDistribution, LogUniformDistribution)):
+                params[name] = min(params[name], search_space[name].high - 1e-8)
```
toshihikoyanase, line 535: `1e-8` may be too large when users use `LogUniformDistribution` since users can specify `1e-8` as the value of `high`. How about shifting the value in the log-space, for example, as follows?

```suggestion
            if isinstance(search_space[name], UniformDistribution):
                params[name] = min(params[name], search_space[name].high - 1e-8)
            elif isinstance(search_space[name], LogUniformDistribution):
                shifted_high = math.exp(math.log(search_space[name].high) - 1e-8)
                params[name] = min(params[name], shifted_high)
```

optuna/integration/botorch.py
```diff
@@ -0,0 +1,467 @@
...
+        Next set of candidates. Usually the return value of BoTorch's ``optimize_acqf``.
+
+    """
+
+    if train_obj.size(-1) != 1:
+        raise ValueError("Objective may only contain single values with qEI.")
+    if train_con is not None:
+        train_y = torch.cat([train_obj, train_con], dim=-1)
+
+        best_f = (train_obj * (train_con <= 0)).max()
```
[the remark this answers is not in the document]
hvy: Fixed to check against `all`.

toshihikoyanase (commented):
Thank you for your great PR! I have not dived into the details of BoTorch-related code yet, but I left a comment and a question.

BTW, the current example may not be intuitive for novice users of `botorch` since the calculation of the objective values and the constraints is hidden by using botorch's implementation. How about using a simple function as can be seen in the `examples/multi_objective/plot_pareto_front.py`?

I used the following example code to examine the usability of this sampler:

```python
import optuna


def objective(trial):
    # Binh and Korn function.

    v0 = 4 * x ** 2 + 4 * y ** 2
    trial.set_user_attr("constraint", (20 - v0, ))
    return v0, v1

def constraints(study, trial):

sampler = optuna.integration.BoTorchSampler(
    constraints_func=constraints,
    n_startup_trials=10,
)

study = optuna.multi_objective.create_study(["minimize", "minimize"], sampler=sampler)

optuna.multi_objective.visualization.plot_pareto_front(study, include_dominated_trials=True).show()
```

optuna/integration/botorch.py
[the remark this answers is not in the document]
eytan: 1. I see, so this is a technical limitation of Optuna?  I'd be concerned that if this information about pending points weren't passed into BoTorch any type of parallel async evaluation will end up sampling the same set of points over and over again.
2. Yes, exactly.  The Knowledge Gradient implementation I linked to follows the [Practical Multi-fidelity Bayesian Optimization for Hyperparameter Tuning](http://proceedings.mlr.press/v115/wu20a/wu20a.pdf) paper, and we also have an [implementation](https://github.com/facebook/Ax/blob/9c546f7d5f15bb50b9eda4837a52a1d4f22d8f67/ax/models/torch/botorch_modular/mes.py) of [Multi-fidelity Max-value Entropy Search](http://proceedings.mlr.press/v119/takeno20a/takeno20a.pdf).  In BoTorch, we refer to the context variables as "fixed features" as they are not variables being optimized over.  We also support passing in multiple "trace observations" (samples from the learning curve), although this functionality hasn't been very thoroughly tested.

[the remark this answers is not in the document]
hvy: > ... any type of parallel async evaluation will end up sampling the same set of points over and over again

This is very useful information. Actually, we should be able to construct a tensor of "pending points" (since parameters that have been suggested but not evaluated are persisted and available to all workers) so this might be something that we can look into. In my understanding, we basically have to pass this information to the acquisition function. That'd probably require a change to the interface of the callback to the sampler but it should be acceptable given that this is currently marked experimental.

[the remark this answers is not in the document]
HideakiImamura: 1. Thank you for the useful information. This is not a technical limitation of Optuna, but rather a limitation of the current PR implementation of BoTorchSampler: Optuna can access parameters that have already been sampled and are being evaluated. If BoTorch can distinguish between parameters that are being evaluated and those that have already been evaluated, then we can work around this problem by devising a better implementation of BoTorchSampler. However, for this PR, I think it would be smart to handle this as a follow-up, without making that effort.
2. That's interesting information. The hyperparameter optimization framework Optuna might be a good practical example of multi-fidelity optimization! I'm looking forward to BoTorchSampler supporting multi-fidelity optimization in the near future as a follow-up task.

hvy:
> @hvy , for your latest plot, what are the comparisons to? It looks like you need to do many more repeats to measure any type significant differences, although hopefully there are no differences between the optuna and base versions of each?

@eytan, thanks for your comment. The plot was based on [BoTorch's tutorial notebook for qEHVI](https://botorch.org/tutorials/multi_objective_bo). [This is the modified notebook](https://github.com/optuna/optuna/blob/161e8bd867918e9014e006cf08b77c0586de690a/botorch_bench_local/constrained_multi_objective_bo.ipynb ) I used to create the plots (the file is quite large and might have to be downloaded). Just reran the optimizations with 10 trials and 60 observations. I attached the results below, which seems acceptable. But I believe we should prepare proper benchmark suites for other settings as well. Do you have any suites in Ax or BoTorch that we could perhaps look into?

Screen Shot 2020-12-16 at 15 40 24

optuna/integration/botorch.py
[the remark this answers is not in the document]
hvy: Please correct me if I'm wrong but in my understanding, [PyTorch only allows settings seeds at the global scope](https://pytorch.org/docs/stable/notes/randomness.html). If that's the case, then I don't we should.

hvy: Makes sense, fixed.

optuna/integration/botorch.py
hvy: Fixed.

hvy:
Thanks @toshihikoyanase for your comments. They should be addressed now, PTAL. I also changed the example to Binh and Korn as you suggested, expanded the search space and used the original constraints.

eytan:
> > @hvy , for your latest plot, what are the comparisons to? It looks like you need to do many more repeats to measure any type significant differences, although hopefully there are no differences between the optuna and base versions of each?
> 
> @eytan, thanks for your comment. The plot was based on [BoTorch's tutorial notebook for qEHVI](https://botorch.org/tutorials/multi_objective_bo). [This is the modified notebook](https://github.com/optuna/optuna/blob/161e8bd867918e9014e006cf08b77c0586de690a/botorch_bench_local/constrained_multi_objective_bo.ipynb) I used to create the plots (the file is quite large and might have to be downloaded). Just reran the optimizations with 10 trials and 60 observations. I attached the results below, which seems acceptable. But I believe we should prepare proper benchmark suites for other settings as well. Do you have any suites in Ax or BoTorch that we could perhaps look into?
> 
> Screen Shot 2020-12-16 at 15 40 24

This looks great.  Ax does have benchmarking functionality at https://github.com/facebook/Ax/tree/master/ax/benchmark I am not sure if there are any tutorials.

optuna/integration/botorch.py
eytan: When users specify these log distributions, are you doing any transformations of the inputs so that the models are fit on a log scale-inputs?  That would be really helpful for achieving good model performance.  One advantage of Ax is that a lot of the bookkeeping for these transforms are handled automatically (you just have to specify log_scale: True when defining your search space).

optuna/integration/botorch.py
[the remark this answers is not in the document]
HideakiImamura: Let me resolve this conversation since this is addressed.

HideakiImamura (approved):
Thanks for your update. LGTM!

optuna/integration/botorch.py
hvy: Yes, the data and the bounds are transformed accordingly before being passed to the functions that construct the models. So it should behave as you describe. There is a [helpers in Optuna that does the transformation](https://github.com/optuna/optuna/blob/master/optuna/_transform.py) such as one-hot encodings and discretizations. We could perhaps rely on Ax to do this but it's coupled to the data structures (distribution classes) in Optuna and it's a lower level utility to be used by several samplers. It might be good to be aware of inconsistencies if there are such.

(In Optuna, a users specifies this log distributions, discretizations, etc. in the [suggest methods](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.trial.Trial.html#optuna.trial.Trial.suggest_float))

tests/integration_tests/test_botorch.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+@pytest.mark.parametrize("n_objectives", [1, 2, 4])
+def test_botorch_candidates_func_none(n_objectives: int) -> None:
+    n_trials = 3
+    n_startup_trials = 2
+
+    sampler = BoTorchSampler(n_startup_trials=n_startup_trials)
+
+    study = multi_objective.create_study(["minimize"] * n_objectives, sampler=sampler)
```
toshihikoyanase, line 25: How about replacing the `multi_objective.create_study` with `optuna.create_study`? #1994 has been merged and you update the BoTorch example with it.
```suggestion
    study = optuna.create_study(directions=["minimize"] * n_objectives, sampler=sampler)
```

tests/integration_tests/test_botorch.py
```diff
@@ -0,0 +1,249 @@
...
+
+        return candidates
+
+    n_trials = 3
+    n_startup_trials = 1
+
+    sampler = BoTorchSampler(candidates_func=candidates_func, n_startup_trials=n_startup_trials)
+
+    study = multi_objective.create_study(["minimize"], sampler=sampler)
+    study.optimize(lambda t: [t.suggest_float("x0", 0, 1)], n_trials=n_trials)
```
toshihikoyanase, line 67: ditto.
```suggestion
    study = optuna.create_study(direction="minimize", sampler=sampler)
    study.optimize(lambda t: t.suggest_float("x0", 0, 1), n_trials=n_trials)
```

tests/integration_tests/test_botorch.py
```diff
@@ -0,0 +1,249 @@
...
+    ) -> torch.Tensor:
+        # Must be a `torch.Tensor`, not a list.
+        return torch.rand(1).tolist()  # type: ignore
+
+    sampler = BoTorchSampler(candidates_func=candidates_func, n_startup_trials=1)
+
+    study = multi_objective.create_study(["minimize"], sampler=sampler)
+
+    with pytest.raises(TypeError):
+        study.optimize(lambda t: [t.suggest_float("x0", 0, 1)], n_trials=3)
```
toshihikoyanase, line 88: ditto.
```suggestion
    study = optuna.create_study(direction="minimize", sampler=sampler)

    with pytest.raises(TypeError):
        study.optimize(lambda t: t.suggest_float("x0", 0, 1), n_trials=3)
```

tests/integration_tests/test_botorch.py
```diff
@@ -0,0 +1,249 @@
...
+        bounds: torch.Tensor,
+    ) -> torch.Tensor:
+        return torch.rand(2, 1)  # Must have the batch size one, not two.
+
+    sampler = BoTorchSampler(candidates_func=candidates_func, n_startup_trials=1)
+
+    study = multi_objective.create_study(["minimize"], sampler=sampler)
+
+    with pytest.raises(ValueError):
+        study.optimize(lambda t: [t.suggest_float("x0", 0, 1)], n_trials=3)
```
toshihikoyanase, line 105: ditto.
```suggestion
    study = optuna.create_study(direction="minimize", sampler=sampler)

    with pytest.raises(ValueError):
        study.optimize(lambda t: t.suggest_float("x0", 0, 1), n_trials=3)
```

tests/integration_tests/test_botorch.py
```diff
@@ -0,0 +1,249 @@
...
+        bounds: torch.Tensor,
+    ) -> torch.Tensor:
+        return torch.rand(1, 1, 1)  # Must have one or two dimensions, not three.
+
+    sampler = BoTorchSampler(candidates_func=candidates_func, n_startup_trials=1)
+
+    study = multi_objective.create_study(["minimize"], sampler=sampler)
+
+    with pytest.raises(ValueError):
+        study.optimize(lambda t: [t.suggest_float("x0", 0, 1)], n_trials=3)
```
toshihikoyanase, line 122: ditto.
```suggestion
    study = optuna.create_study(direction="minimize", sampler=sampler)

    with pytest.raises(ValueError):
        study.optimize(lambda t: t.suggest_float("x0", 0, 1), n_trials=3)
```

tests/integration_tests/test_botorch.py
```diff
@@ -0,0 +1,249 @@
...
+        train_x: torch.Tensor,
+        train_obj: torch.Tensor,
+        train_con: Optional[torch.Tensor],
+        bounds: torch.Tensor,
+    ) -> torch.Tensor:
+        return torch.rand(n_params - 1)  # Must return candidates for all parameters.
+
+    sampler = BoTorchSampler(candidates_func=candidates_func, n_startup_trials=1)
+
+    study = multi_objective.create_study(["minimize"] * n_params, sampler=sampler)
```
toshihikoyanase, line 138: ditto.
```suggestion
    study = optuna.create_study(directions=["minimize"] * n_params, sampler=sampler)
```

tests/integration_tests/test_botorch.py
```diff
@@ -0,0 +1,249 @@
...
+        constraints_func_call_count += 1
+
+        return (xs - 0.5,)
+
+    n_trials = 4
+    n_startup_trials = 2
+
+    sampler = BoTorchSampler(constraints_func=constraints_func, n_startup_trials=n_startup_trials)
+
+    study = multi_objective.create_study(["minimize"] * n_objectives, sampler=sampler)
```
toshihikoyanase, line 163: ditto.
```suggestion
    study = optuna.create_study(directions=["minimize"] * n_objectives, sampler=sampler)
```

tests/integration_tests/test_botorch.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def test_botorch_constraints_func_invalid_type() -> None:
+    def constraints_func(trial: FrozenTrial) -> Sequence[float]:
+        x0 = trial.params["x0"]
+        return x0 - 0.5  # Not a tuple, but it should be.
+
+    sampler = BoTorchSampler(constraints_func=constraints_func)
+
+    study = multi_objective.create_study(["minimize"], sampler=sampler)
```
toshihikoyanase, line 180: ditto.
```suggestion
    study = optuna.create_study(directions=["minimize"], sampler=sampler)
```

tests/integration_tests/test_botorch.py
```diff
@@ -0,0 +1,249 @@
...
+
+    with pytest.raises(TypeError):
+        study.optimize(lambda t: [t.suggest_float("x0", 0, 1)], n_trials=3)
+
+
+def test_botorch_n_startup_trials() -> None:
+    independent_sampler = RandomSampler()
+    sampler = BoTorchSampler(n_startup_trials=2, independent_sampler=independent_sampler)
+    study = multi_objective.create_study(["minimize", "maximize"], sampler=sampler)
```
toshihikoyanase, line 189: ditto.
```suggestion
    study = optuna.create_study(directions=["minimize", "maximize"], sampler=sampler)
```

tests/integration_tests/test_botorch.py
```diff
@@ -0,0 +1,249 @@
...
+        x2 = trial.suggest_float("x2", 0, 1, step=0.1)
+        x3 = trial.suggest_int("x3", 0, 2)
+        x4 = trial.suggest_int("x4", 2, 4, log=True)
+        x5 = trial.suggest_int("x5", 0, 4, step=2)
+        x6 = cast(float, trial.suggest_categorical("x6", [0.1, 0.2, 0.3]))
+        return [x0 + x1 + x2 + x3 + x4 + x5 + x6]
+
+    sampler = BoTorchSampler()
+
+    study = multi_objective.create_study(["minimize"], sampler=sampler)
```
toshihikoyanase, line 216: ditto.
```suggestion
    study = optuna.create_study(direction="minimize", sampler=sampler)
```

tests/integration_tests/test_botorch.py
```diff
@@ -0,0 +1,249 @@
...
+    storage = RDBStorage("sqlite:///:memory:")
+
+    sampler = BoTorchSampler()
+
+    study = multi_objective.create_study(["minimize"], sampler=sampler, storage=storage)
+    study.optimize(lambda t: [t.suggest_float("x0", 0, 1)], n_trials=3)
+
+    other_study = multi_objective.create_study(["minimize"], sampler=sampler, storage=storage)
+    with pytest.raises(RuntimeError):
+        other_study.optimize(lambda t: [t.suggest_float("x0", 0, 1)], n_trials=3)
```
toshihikoyanase, line 237: ditto.
```suggestion
    study = optuna.create_study(direction="minimize", sampler=sampler, storage=storage)
    study.optimize(lambda t: t.suggest_float("x0", 0, 1), n_trials=3)

    other_study = optuna.create_study(direction="minimize", sampler=sampler, storage=storage)
    with pytest.raises(RuntimeError):
        other_study.optimize(lambda t: t.suggest_float("x0", 0, 1), n_trials=3)
```

toshihikoyanase (commented):
Thank you for your update. It basically looks good to me, but I added suggestions.

Regarding the comment about random seeds (https://github.com/optuna/optuna/pull/1989#discussion_r544306401), [`SobolQMCNormalSampler`](https://botorch.org/v/0.1.2/api/sampling.html#sobolqmcnormalsampler) has its own `seed` argument, but it seems to follow the pytorch's global seed if we do not specify it.

setup.py
```diff
@@ -144,6 +144,7 @@ def get_extras_require() -> Dict[str, List[str]]:
             "torchvision==0.8.1+cpu ; sys_platform!='darwin'",
             "torchaudio==0.7.0",
             "allennlp==1.2.0 ; python_version<'3.8'",
+            "botorch ; python_version>'3.6'",
```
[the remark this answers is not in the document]
hvy: You're right, fixed!

examples/botorch_simple.py
```diff
@@ -0,0 +1,61 @@
...
+
+    sampler = optuna.integration.BoTorchSampler(
+        constraints_func=constraints,
+        n_startup_trials=10,
+    )
+    study = optuna.create_study(
+        directions=["minimize", "minimize"],
+        sampler=sampler,
+    )
+    study.optimize(objective, n_trials=32)
```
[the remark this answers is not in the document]
hvy: Good catch, fixed.

tests/integration_tests/test_botorch.py
hvy: Thanks, I should've noticed that. Fixed!

tests/integration_tests/test_botorch.py
hvy: Fixed.

tests/integration_tests/test_botorch.py
hvy: Fixed.

tests/integration_tests/test_botorch.py
hvy: Fixed.

tests/integration_tests/test_botorch.py
hvy: Fixed.

tests/integration_tests/test_botorch.py
hvy: Fixed.

tests/integration_tests/test_botorch.py
hvy: Fixed.

tests/integration_tests/test_botorch.py
hvy: Fixed.

tests/integration_tests/test_botorch.py
hvy: Fixed.

tests/integration_tests/test_botorch.py
hvy: Fixed.

tests/integration_tests/test_botorch.py
hvy: Fixed.

hvy:
Thanks @toshihikoyanase for your careful review, comments should be addressed. PTAL.

toshihikoyanase (approved):

PR #1787, opened 2020-09-04
Implement `plot_param_importances` and `_get_param_importance_plot` with Matplotlib backend.

ytknzw:
This PR proposes to implement a ``matplotlib`` backend for plotting parameter importance based on PR #1756 .

## Motivation
To provide another visualization option for users, who have only ``plotly`` now.
See issue #1539 and PR #1707 for the original motivation.

## Description of the changes
Create  ``optuna.visualization.matplotlib._param_importances.plot_param_importances``, ``optuna.visualization.matplotlib._param_importances._get_param_importance_plot``, ``optuna.visualization.matplotlib._param_importances._get_distribution``, ``optuna.visualization.matplotlib._param_importances._get_color``.

PR #1991, opened 2020-11-03
Change bar colours in the parameter importance plot.

ytknzw:
## Motivation
This PR is to change colours in the parameter importance plot to distinguishing ones for better recognition based on @himkt 's suggestion https://github.com/optuna/optuna/pull/1787#pullrequestreview-521346823 .

## Description of the changes
Change colour palettes used in both the Plotly and Matplotlib backends from blue-based ones to ones with various colours (`D3` and `tab10`). See PR #1756 for details about Matplotlib backends.

### DIFF
--- a/optuna/visualization/_param_importances.py
+++ b/optuna/visualization/_param_importances.py
@@ -22,7 +22,7 @@ if _imports.is_successful():
 
     from optuna.visualization._plotly_imports import go
 
-    D3 = plotly.colors.sequential.D3
+    D3 = plotly.colors.qualitative.D3
 
     _distribution_colors = {
         UniformDistribution: D3[0],
--- a/optuna/visualization/_param_importances.py
+++ b/optuna/visualization/_param_importances.py
@@ -22,15 +22,15 @@ if _imports.is_successful():
 
     from optuna.visualization._plotly_imports import go
 
-    Blues = plotly.colors.sequential.Blues
+    D3 = plotly.colors.sequential.D3
 
     _distribution_colors = {
-        UniformDistribution: Blues[-1],
-        LogUniformDistribution: Blues[-1],
-        DiscreteUniformDistribution: Blues[-1],
-        IntUniformDistribution: Blues[-2],
-        IntLogUniformDistribution: Blues[-2],
-        CategoricalDistribution: Blues[-4],
+        UniformDistribution: D3[0],
+        LogUniformDistribution: D3[1],
+        DiscreteUniformDistribution: D3[2],
+        IntUniformDistribution: D3[3],
+        IntLogUniformDistribution: D3[4],
+        CategoricalDistribution: D3[5],
     }
 
 logger = get_logger(__name__)
--- a/optuna/visualization/matplotlib/_param_importances.py
+++ b/optuna/visualization/matplotlib/_param_importances.py
@@ -27,14 +27,14 @@ if _imports.is_successful():
     from optuna.visualization.matplotlib._matplotlib_imports import Rectangle
 
     # Color definitions.
-    cmap = cm.get_cmap("tab20c")
+    cmap = cm.get_cmap("tab10")
     _distribution_colors = {
         UniformDistribution: cmap(0),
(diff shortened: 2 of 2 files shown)


hvy:
Thanks for this PR, let me take a look as I first introduced the colors for the plotly version. Could you just merge/rebase with the latest master now that #1787 is merged? That'd make reviewing easier.

ytknzw:
Thank you for your suggestion. I've just merged the latest.

optuna/visualization/_param_importances.py
```diff
@@ -22,15 +22,15 @@
...
-        DiscreteUniformDistribution: Blues[-1],
-        IntUniformDistribution: Blues[-2],
-        IntLogUniformDistribution: Blues[-2],
-        CategoricalDistribution: Blues[-4],
+        UniformDistribution: D3[0],
+        LogUniformDistribution: D3[1],
+        DiscreteUniformDistribution: D3[2],
+        IntUniformDistribution: D3[3],
+        IntLogUniformDistribution: D3[4],
+        CategoricalDistribution: D3[5],
```
hvy, line 33: Just want to clarify but do you think it's better to have different colors for different distributions? I am a bit skeptical since the distribution classes are not directly accessed by users and this change introduces a lot of information into the plots. Actually, colors are not the most important aspects in these figures.
newplot (27)

ytknzw: Thank you for your comment! I understand your points and think there're 3 options to improve the current implementation and this PR:

1. Use single colour (not colourmap) for all the distributions both in Plotly and Matplotlib backends.
- Replace the blue colourmap with a colour "blue" in current implementation.
- Close this PR.

2. Use single colour in Plotly backend while different colours with legends in Matplotlib backend.
- Replace the blue colourmap with a colour "blue" in current implementation of Plotly.
- Keep changes in this PR only for Matplotlib.

3. Use different colurs with legends both in Plotly and Matplolib backends.
- Keep all the changes in this PR.
- Add legends to Plotly.

If the distribution classes are not directly referenced by users, it would confuse them to use several colours (even in the same colourmap) without any legend.

hvy: Then, it might be better to just stick to a single color, no legends. For plotly, I believe we can still keep the mouse hover helpers, just as a bonus. Cc. @himkt

hvy: Seems like it's been labeled stale. Are there any further comments on the coloring?

ytknzw: So sorry for my slow reply! Let me check this weekend.

hvy: I haven't quite understood how hatches and more distinguishable colors would differ in practice. A concrete example might help. By the way, https://github.com/optuna/optuna/issues/2080 might help resolving this discussion.

nzw0301: I believe that colourblindness should be considered. [The new colour map in this PR](https://github.com/optuna/optuna/pull/1991#discussion_r517717525) couldn't be distinguished between red and green when I used [this simulation](https://www.color-blindness.com/coblis-color-blindness-simulator/). Please see also a suggestion in [the docs of matplotlib](https://matplotlib.org/stable/tutorials/colors/colormaps.html#color-vision-deficiencies) -- it recommends not using red and green at the same time.

I also test the colour maps in #2028 a little bit. They look well.

nzw0301:
@ytknzw we really appreciate your contribution to this pull request. As we discussed in this pull request, giving a different color to each distribution and legend is supposed to be unclear for optuna users since the users might not access a distribution class directly. In addition, single color among distributions is enough to visualize important hyperparameters when we think about the role of this plot -- the color does not matter. Thus let us close this pull request and summarize the follow-up work as follows:

1. use a single color map for both Plotly and Matplolib backends.
2. Remove legend from Matplolib backend.

Again, thank you for your pull request and concrete suggestions! The discussion in this pull request is definitely helpful to improve this plot function.

PR #1997, opened 2020-11-05
[DO NOT MERGE] A prototypical implementation of step-wise sampler

toshihikoyanase:
## Motivation

In #1632, we discuss the design of the customizable `LightGBMTuner`.  #1947 implements it on top of `optuna.stepwise.StepwiseTuner`, which is a newly introduced concept to invoke optimization in a step-wise manner.

Alternatively, I used a sampler (`StepwiseSampler`) to implement step-wise tuning in this PR. `StepwiseSampler` receives a list of steps, and a step consists of a search space, sampler, and `n_trials`.  Using `StepwiseSampler`, IMO, we can implement `LightGBMTuner` with the usual `Study` and `Sampler`. 

Please note that this is just a prototypical implementation to discuss the design of step tuning. Please let me know if you have any comments.

## Description of the changes

- Add `optuna.samplers.StepwiseSampler`
- Add an example to minimize a quadratic function using `StepwiseSampler`
- Add an example of `LightGBMTuner` using `StepwiseSampler`
- Add filter functions to `TPESampler` to exclude observations that are out of the target distribution. It is required by `LightGBMTuner`

## Comments on API design

- Currently, `Step.__init__` receives the generator functions of search spaces and samplers. It can be helpful if it can also handle the search spaces and samplers.
- We can also omit the distributions from the search spaces. For example, `Step({"x"}, TPESampler(), n_trials=10)` is simpler than `Step({"x": UniformDistribution(0, 100)}, TPESampler, n_trials=10)`.
- I sometimes want a callback mechanism which is called after a step finished.
- All trials include all parameters while the `LightGBMTuner` in the master branch includes only a part of parameters tuned at the step.
- If `steps` have `GridSampler`, the study will be stopped even if `StepwiseSampler` has remaining `steps`. To continue the tuning, `StepwiseSampler` turns off the `Study.stop_flag`, but such implementation may be ad-hoc.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -236,6 +236,7 @@ class TPESampler(BaseSampler):
 
         param_names = list(search_space.keys())
         values, scores = _get_multivariate_observation_pairs(study, param_names)
+        values, scores = _filter_out_multivariate_observation_pairs(values, scores, search_space)
 
         # If the number of samples is insufficient, we run random trial.
         n = len(scores)
@@ -272,6 +273,7 @@ class TPESampler(BaseSampler):
     ) -> Any:
 
         values, scores = _get_observation_pairs(study, param_name)
+        values, scores = _filter_out_observation_pairs(values, scores, param_distribution)
 
         n = len(values)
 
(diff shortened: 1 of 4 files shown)


jeffzi:
Sorry for the late reply. 

The idea of limiting complexity by reusing existing concepts is appealing :+1: . It works for regular objective function and doesn't force a special objective with base parameters. I also agree with accepting generators for search space and sampler in the `Step` constructor. It would indeed be nice to accept regular search spaces and samplers.

> For example, Step({"x"}, TPESampler(), n_trials=10) is simpler than Step({"x": UniformDistribution(0, 100)}, TPESampler, n_trials=10).

Do you mean that it would fall back to a default distribution?

However I think the Sampler approach has some limitations: 
1. Can't specify pruner per step
2. Does it support trial budget set on the `Study` when n_jobs > 1? Each thread will call [Step.update_n_trials()](https://github.com/toshihikoyanase/optuna/blob/191a88cb02b6f21ad7d68cd4b12105c64519300c/optuna/samplers/_stepwise.py#L93) in parallel. Therefore, I think the results will be inconsistent. 
3. Can't specify timeout per step: can be solved if I'm wrong about the above 2. 
4. We will still need a `Tuner` class to factorize the boilerplate when we implement stepwise for other libraries than LightGBM. For example, setting up the general flow (`run` function), or creating objective that accepts base parameters.
5. Minor point, how would you display the current step in the progress bar as LightGBMTuner does? To me, it looks like it should be done within the sampler since it knows which step is running.
6. Generally speaking, sampler has less visibility on the optimization process. For example, your idea (quoted below) would be hard to implement since the sampler does not know when a trial run ended.

> I sometimes want a callback mechanism which is called after a step finished.

In summary, I believe a full-featured implementation will require to add more and more to `StepwiseSampler` and, in my opinion, it feels out-of-scope for a sampler.

crcrpar:
friendly ping @toshihikoyanase for jeffzi's comment above

jeffzi:
@toshihikoyanase Is there still appetite for a step-wise mechanism (either sampler or #1947)?

toshihikoyanase:
@jeffzi I'm sorry to have kept you waiting for a long time. Some developers planned to take over this PR, but they did not have time enough to complete the work. I'll restart the implementation after the other PRs (e.g., #2205, #2221, and etc).

jeffzi:
No problem I totally understand. Feel free to ping me when you want to restart the discussion. I'm willing to help with the implementation.

optuna/samplers/_tpe/sampler.py
```diff
@@ -827,3 +829,40 @@ def _get_multivariate_observation_pairs(
...
+
+def _filter_out_observation_pairs(
+    values: List[Optional[float]],
+    scores: List[Tuple[float, float]],
+    distribution: BaseDistribution,
+) -> Tuple[List[Optional[float]], List[Tuple[float, float]]]:
+    ret_values = []
+    ret_scores = []
+    for value, score in zip(values, scores):
+        if value is None or distribution._contains(value):
```
toshihikoyanase, line 842: I think `value` should not be `None`. Otherwise, all the observations will be `None` and TPE algorithm will not work as expected.
```suggestion
        if value is not None or distribution._contains(value):
```

PR #1998, opened 2020-11-05
[Dependency Update] PyTorch and AllenNLP

crcrpar:
## Description of the changes

- PyTorch: 1.7.0
- vision: 0.8.1
- audio: 0.7.0 (following pytorch.org)
- Allennlp: 1.2.0

CC: @himkt

PR #1999, opened 2020-11-06
Update k8s docker

crcrpar:
Follow-up of #1998.

### DIFF
--- a/examples/kubernetes/simple/Dockerfile
+++ b/examples/kubernetes/simple/Dockerfile
@@ -1,8 +1,6 @@
-FROM python:3.5-slim-buster
+FROM optuna/optuna:py3.7-dev
 
 WORKDIR /usr/src/
-
-RUN pip install --no-cache-dir optuna scikit-learn psycopg2-binary
 
 COPY sklearn_distributed.py .
 COPY check_study.sh .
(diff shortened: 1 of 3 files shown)


hvy:
#1998 just merged. You can update the diff now 👍

examples/kubernetes/simple/Dockerfile
```diff
@@ -1,8 +1,6 @@
-FROM python:3.5-slim-buster
+FROM optuna/optuna:py3.7-dev
 
 WORKDIR /usr/src/
 
-RUN pip install --no-cache-dir optuna scikit-learn psycopg2-binary
```
hvy, line 5: What was `psycopg2-binary` used for?

crcrpar:
https://pypi.org/project/psycopg2-binary is for PostgreSQL. I thought we could remove that as I could ran the examples locally though we may have to install that.

hvy:
Hm, okay. It's also explicitly installed in our RDB storage tests on CircleCI https://github.com/optuna/optuna/blob/master/.circleci/config.yml#L220. I tried to run the Docker examples locally to verify but couldn't get the study creation pod to run. Did you also verify using the Dockerfile provided along with the shell script? Cc. @toshihikoyanase

crcrpar:
I'm reverting the changes to two Dockerfiles and only updates PyTorch version of mlflow/Dockerfile then I'm struggling.
Let me close this for now as I can't afford to make things great again now. 🙇

Issue #1856, opened 2020-09-23
`optuna.importance.get_param_importances` sometimes skips params


PhilipMay:
When doing `optuna.importance.get_param_importances(study)` some features are sometimes skipped.
Example:

![grafik](https://user-images.githubusercontent.com/229382/93984526-3383af80-fd84-11ea-9aff-b68beda33c6a.png)

While `plot_slice` shows this:

![grafik](https://user-images.githubusercontent.com/229382/93984674-63cb4e00-fd84-11ea-8be2-a0fb962ebbf2.png)

`get_param_importances` somehow skipped `learning_rate`. How does this come? Does optuna only display the most important?

hvy:
Hm I see. Do you have the code for reproduction? Conditional hyperparameters will be skipped (unless you specify which parameters to evaluate explicitly using the `params` parameter) but I would guess that `learning_rate` isn't conditioned?

PhilipMay:
Yes. learning_rate is not conditioned.

I switched to this sampler: `sampler = optuna.samplers.CmaEsSampler(restart_strategy='ipop')`

And get this warning:

```
[W 2020-09-23 09:39:55,691] The parameter 'learning_rate' in trial#53 is sampled independently by using `RandomSampler` instead of `CmaEsSampler` (optimization performance may be degraded). You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `CmaEsSampler`, if this independent sampling is intended behavior.
```

Is this warning in relation to the problem?

I can not print the full code here because it is business related. I am doing a cross validation and call this in a loop:

``` python
def train_on_split(silo_to_use, trial, max_seq_len, batch_size, class_weights, step, processor):
    dropout = trial.suggest_uniform('dropout', 0.0, 0.5)
    learning_rate = trial.suggest_uniform('learning_rate', 2e-7, 2e-4)  # 3e-5
    epochs = trial.suggest_int('epochs', 1, 5)  # 4
    warmup_proportion = trial.suggest_uniform('warmup_proportion', 0.1, 0.5)  # 0.2
```

PhilipMay:
@hvy is there anything I can help you with to debug this?

hvy:
[...]

PR #2003, opened 2020-11-09
Modify get_param_importance

tohmae:
This PR is proposal that get_param_importancees does not skip params when distributions of params are changed.
This PR addresses issue #1856.

### DIFF
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -2,16 +2,13 @@ from typing import Callable
 from typing import Dict
 from typing import List
 from typing import Optional
-from typing import TYPE_CHECKING
+
+import optuna
 
 from optuna.importance._base import BaseImportanceEvaluator
 from optuna.importance._fanova import FanovaImportanceEvaluator
 from optuna.importance._mean_decrease_impurity import MeanDecreaseImpurityImportanceEvaluator
 from optuna.trial import FrozenTrial
-
-
-if TYPE_CHECKING:
-    from optuna.study import Study
 
 
 __all__ = [
@@ -23,7 +20,7 @@ __all__ = [
 
 
 def get_param_importances(
-    study: "Study",
+    study: "optuna.Study",
     *,
     evaluator: Optional[BaseImportanceEvaluator] = None,
     params: Optional[List[str]] = None,
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -4,7 +4,6 @@ from typing import List
 from typing import Optional
 
 import optuna
-
 from optuna.importance._base import BaseImportanceEvaluator
 from optuna.importance._fanova import FanovaImportanceEvaluator
 from optuna.importance._mean_decrease_impurity import MeanDecreaseImpurityImportanceEvaluator
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -2,12 +2,16 @@ from typing import Callable
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import TYPE_CHECKING
 
 from optuna.importance._base import BaseImportanceEvaluator
 from optuna.importance._fanova import FanovaImportanceEvaluator
 from optuna.importance._mean_decrease_impurity import MeanDecreaseImpurityImportanceEvaluator
-from optuna.study import Study
-from optuna.trial import FrozenTrial
+
+
+if TYPE_CHECKING:
+    from optuna.study import Study
+    from optuna.trial import FrozenTrial
 
 
 __all__ = [
@@ -19,11 +23,11 @@ __all__ = [
 
 
 def get_param_importances(
-    study: Study,
+    study: "Study",
     *,
     evaluator: Optional[BaseImportanceEvaluator] = None,
     params: Optional[List[str]] = None,
-    target: Optional[Callable[[FrozenTrial], float]] = None,
+    target: Optional[Callable[["FrozenTrial"], float]] = None,
 ) -> Dict[str, float]:
     """Evaluate parameter importances based on completed trials in the given study.
 
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -2,8 +2,12 @@ from typing import Callable
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import TYPE_CHECKING
 
-import optuna
+
+if TYPE_CHECKING:
+    import optuna
+
 from optuna.importance._base import BaseImportanceEvaluator
 from optuna.importance._fanova import FanovaImportanceEvaluator
 from optuna.importance._mean_decrease_impurity import MeanDecreaseImpurityImportanceEvaluator
@@ -19,7 +23,7 @@ __all__ = [
 
 
 def get_param_importances(
-    study: "optuna.Study",
+    study: "optuna.study.Study",
     *,
     evaluator: Optional[BaseImportanceEvaluator] = None,
     params: Optional[List[str]] = None,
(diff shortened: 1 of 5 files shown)


hvy:
Seems like this has been stale for a while. Do you need any help in proceeding with it? 

By the way, I am wondering if it's not enough to alter the logic in `_get_distributions` to address this issue.

tohmae:
I am sorry I am late. I modify source code so that  get_param_importance function cat be passed with params argument and add test code.

hvy:
Sorry, I probably just missed this it but are you changes ready for review? 

Also, it'd be good if you could merge the latest master. We recently introduced RDB test requirements into our CI.

tohmae:
I merged the latest master. I would be grateful if you could review.

hvy:
Great, thanks for the updates and the heads up. I'll have a look at it later today!

hvy:
The `TYPE_CHECKING` changes, they seem unrelated at first. Do you think we should keep them? I would otherwise like to keep the PR minimal. What do you think?

tests/importance_tests/test_init.py
```diff
@@ -247,18 +247,76 @@ def objective(trial: Trial) -> float:
...
 
-    with pytest.raises(ValueError):
-        get_param_importances(study, evaluator=evaluator_init_func(), params=["x1"])
+    param_importance = get_param_importances(study, evaluator=evaluator_init_func())
+
+    assert isinstance(param_importance, OrderedDict)
+    assert len(param_importance) == 6
+    assert all(
+        param_name in param_importance for param_name in ["x1", "x2", "x3", "x4", "x5", "x6"]
+    )
```
hvy, line 279: An optional but it might be easier to create a set out of the keys and compare for equality.

optuna/importance/_base.py
```diff
@@ -63,43 +73,91 @@ def evaluate(
...
+        info_distribution = {
+            "is_categorical": False,
+            "low": distribution.low,
+            "high": distribution.high,
+        }
+
+    return info_distribution
+
+
+def _get_distributions(study: "Study", params: Optional[List[str]]) -> Dict[str, BaseDistribution]:
```
hvy, line 102: This function is becoming somewhat non-trivial. It's a private one but I'm wondering if we shouldn't write unit tests for it to be certain about keeping its behavior as expected.

optuna/importance/_base.py
```diff
@@ -63,43 +73,91 @@ def evaluate(
...
-
-    # New temporary required to pass mypy. Seems like a bug.
-    params_not_none = params
-    assert params_not_none is not None
-
-    # Compute the search space based on the subset of trials containing all parameters.
-    distributions = None
-    for trial in study.trials:
+    info_distributions = None
+    for trial in reversed(study._storage.get_all_trials(study._study_id, deepcopy=False)):
```
hvy, line 105: You might have reversed the iteration based on `IntersectionSearchSpace` but we don't a cursor here and don't think we can perform any  early breaks. Can we iterate over it in sequence for simplicity?

optuna/importance/_base.py
```diff
@@ -63,43 +73,91 @@ def evaluate(
...
-        if any(
-            trial_distributions[name] != distribution
-            for name, distribution in distributions.items()
-        ):
+        delete_list = []
+        for param_name, info_distribution in info_distributions.items():
+            if param_name not in trial.distributions:
+                delete_list.append(param_name)
+            else:
+                trial_info_distribution = _get_info_distribution(trial.distributions[param_name])
```
hvy, line 126: It feels a bit like over-engineering introducing a dictionary structure here. Actually, can we keep actual distribution objects instead? We could then perhaps keep updating the lows and highs of those objects during the iteration and return them as they are.

hvy (commented):
Thanks for the great update. I left some comments. Could you have a look at them when you have time?

tohmae:
>The TYPE_CHECKING changes, they seem unrelated at first. Do you think we should keep them? I would otherwise like to keep the PR minimal. What do you think?

I do not need intersection_search_space function, so I remove "from optuna.samplers import intersection_search_space" from _base.py. As a result when I import Study, ”ImportError: cannot import name 'Study' from partially initialized module 'optuna.study' (most likely due to a circular import)” message appears. So the TYPE_CHECKING changes. I can not find any other way. Please give me your advice if you have another good idea.

>Thanks for the great update. I left some comments. Could you have a look at them when you have time?

Thank you for your comments. I fixed the parts you pointed out. Could you please review again?

optuna/importance/_base.py
```diff
@@ -63,43 +67,47 @@ def evaluate(
...
-                "parameters are specified. Specified parameters: {}.".format(params)
-            )
+        delete_list = []
+        for param_name, param_distribution in distributions.items():
+            if param_name not in trial.distributions:
+                delete_list.append(param_name)
+            else:
+                trial_param_distribution = trial.distributions[param_name]
+                if isinstance(param_distribution, CategoricalDistribution):
+                    continue
```
hvy, line 96: Before this `continue`, how about asserting that choices are equal? Although it's unlikely to change in the near future, that'd introduce silent bugs.

tests/importance_tests/test_init.py
```diff
@@ -9,10 +9,19 @@
 import optuna
 from optuna import samplers
 from optuna import storages
+
+# from optuna.distributions import BaseDistribution
```
hvy, line 13: Let's delete the line altogether if the import isn't needed.

tests/importance_tests/test_init.py
```diff
@@ -27,6 +36,149 @@
...
+    study.optimize(objective, n_trials=3)
+
+    param_distributions = _get_distributions(study, None)
+    params = ["x1", "x2", "x3", "x4", "x5", "x6"]
+    assert isinstance(param_distributions, OrderedDict)
+    assert len(param_distributions) == len(params)
+    assert all(param in param_distributions for param in params)
+    for param_name, distribution in param_distributions.items():
+        assert isinstance(param_name, str)
+        #        assert isinstance(distribution, BaseDistribution)
```
hvy, line 137: I'm fine with either but let's remove this line or uncomment it.

optuna/importance/__init__.py
```diff
@@ -2,14 +2,18 @@
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import TYPE_CHECKING
```
hvy, line 5: Let's try avoiding having to rely on `TYPE_CHECKING` to be consistent. (I know you're working on it but leaving a comment for the record 🙇 )

hvy (changes requested):
Sorry for the delayed review and thanks for the updated. I left some minor comments but it otherwise LGTM!

Might be worth noting that the logic is a bit bug prone when `params` are specified demonstrated by the following test since it puts some assumptions on the caller that it'll exclude all trials that don't include the entire set of params. However, it shouldn't be an issue for now (it's also a private function).

```python
       if param_name == "x1":
            assert distribution.low == 0.1
            if "x4" in params:
                assert distribution.high == 18.0
            else:
                assert distribution.high == 19.0
```

optuna/importance/_base.py
```diff
@@ -63,43 +63,55 @@ def evaluate(
...
-                "parameters are specified. Specified parameters: {}.".format(params)
-            )
+        delete_list = []
+        for param_name, param_distribution in distributions.items():
+            if param_name not in trial.distributions:
+                delete_list.append(param_name)
+            else:
+                trial_param_distribution = trial.distributions[param_name]
+                if isinstance(param_distribution, CategoricalDistribution):
+                    if trial_param_distribution != param_distribution:
```
hvy, line 94: Can we assert this instead since `_get_distributions` is a private method and only used internally?

hvy:
Thanks, it basically LGTM although I left a single comment. The docs build seems to be failing so we should probably try to fix that as well. I'm not yet sure why that is.

PR #2009, opened 2020-11-11
Code quality improvements

srijan-deepsource:
## Motivation
This pull request fixes some of the quality issues raised by [DeepSource](https://deepsource.io) on my fork of this repository.
You can take a quick glance at all the issues detected [here](https://deepsource.io/gh/srijan-deepsource/optuna/).

## Description of the changes
- Remove redundant call
- Merge `isinstance` calls
- Remove unnecessary generator
- Iterate dictionary directly
- Remove unnecessary import alias
- Remove unnecessary comprehension
- Remove unnecessary use of comprehension
- Remove length check in favor of truthiness of the object
- Add .deepsource.toml


crcrpar (commented):
The changes look great. With this `.deepsource.toml`, will it be possible to have DeepSource send pull requests of this kind of improvement once we integrate DeepSource into this repository?

srijan-deepsource:
> The changes look great. With this `.deepsource.toml`, will it be possible to have DeepSource send pull requests of this kind of improvement once we integrate DeepSource into this repository?

Yes. 
You'd also need to Install DeepSource on your repository [here](https://deepsource.io/signup).
And then, activate analysis [here](https://deepsource.io/gh/optuna/optuna).

Most of the fixes in this PR have been generated automatically using DeepSource autofix.

I also noticed `optuna` is running `black` and `isort` as a part of CI Check.
DeepSource has a feature called `Transformers` (which I've included in the config) to take care of this too. It will automatically commit to the PR with the generated patch.

hvy:
Thanks @srijan-deepsource for this interesting PR.

The changed in the code generally look good (although I prefer looping over dicts with `keys()` to communicate the type of object being iterated over. Similarly doing a boolean check of lists with `len`). I'm however not confident in introducing DeepSource into our CI pipeline yet. We already have quite a few steps there and have been struggling for a while to simplify it so I'd like stronger motivations for introducing another one. Would you perhaps mind just merging the generated diffs?

srijan-deepsource:
> Thanks @srijan-deepsource for this interesting PR.
> 
> The changed in the code generally look good (although I prefer looping over dicts with `keys()` to communicate the type of object being iterated over. Similarly doing a boolean check of lists with `len`). I'm however not confident in introducing DeepSource into our CI pipeline yet. We already have quite a few steps there and have been struggling for a while to simplify it so I'd like stronger motivations for introducing another one. Would you perhaps mind just merging the generated diffs?

Few clarifications:
DeepSource won't be a part of the CI. It is a different service and the checks will run in parallel. In most (almost all) cases: these checks are gonna finish before the CI checks.
After integrating DeepSource, you can go ahead and remove all the linting checks from your CI (`black`, `isort`, `mypy` and `flake8`) as DeepSource internally takes care of that. This will make your CI pipeline simpler (and faster).

Having said that, let me know if you're still unsure about this.
I'll go ahead and remove the config file, and you can merge the rest of the changes.

hvy:
Thanks for your clarification. Just to respond to that, the parallel execution was actually assumed and the complexity was not about the cost of running but rather maintaining. Adding an abstraction on top of `black`, `isort`, etc. seems like one way to go about addressing this but I think we'll have to carefully choose what sort to introduce (considering local execution, etc.)

Again, having that said. It would be great if you could introduce the enhancements alone. Though, I'd like to leave the explicit `len` checks and `keys()`, unless there are strong reasons not to do so.

srijan-deepsource:
No problem. I'll remove the config file and make other changes once this comment gets resolved.

Here are my reasons behind the `len` and `keys` changes:

I removed the `len` check because it was being used to check if the sequence is empty or not.
Any empty sequence in Python returns a `Falsey` value, so `if sequence` is enough to check if it is empty or not.
On the other hand, when `len` is used: it must compute and find the length of the sequence and then evaluate if it is empty or not. Discarding `len` check will also give you a small performance improvement.
```bash
In [2]: x = list(range(10000))                                                                                                                                                                             
In [3]: y = [] 
In [4]: timeit.timeit(stmt="if len(x): pass", number=900000, globals=globals())                                                                                                                            
Out[4]: 0.06649999199999002

In [5]: timeit.timeit(stmt="if x: pass", number=900000, globals=globals())                                                                                                                                 
Out[5]: 0.02530318599997372
```

And the change for iterating dicts directly instead of using `.keys` --
Here's what [PEP 234](https://www.python.org/dev/peps/pep-0234/#dictionary-iterators) says:

> Dictionaries implement a tp_iter slot that returns an efficient iterator that iterates over the keys of the dictionary. During such an iteration, the dictionary should not be modified, except that setting the value for an existing key is allowed (deletions or additions are not, nor is the update() method). This means that we can write
> ```python
>for k in dict: ...
>```
>which is equivalent to, but much faster than
>
>```python
> for k in dict.keys(): ...
>```
>as long as the restriction on modifications to the dictionary (either by the loop or by another thread) are not violated.

Let me know if you want me to revert those two changes.

hvy:
Again, thanks for the explanation and the benchmark. I'm just a bit doubtful about those performance gains being significant for Optuna (without benchmarks with Optuna use cases that is), and feel like it's more about personal preference. If you could revert just those changes, I'd be happy to merge them, I would otherwise like some more feedback so that we have some consensus.

hvy:
Sorry for going back and forth. Discussed this with some other maintainers but it seems okay to proceed if you could keep everything besides the configuration file. Could you just remove the configuration file? The code changes that I addressed above are mostly preferences, as mentioned earlier, and since we are not checking them in the CI, let's have them merged.

srijan-deepsource:
Hey @hvy, am so sorry. This notification got lost in the pool of my existing notifications.

I can see there are some conflicts now, I'll take a look at the changes and update the PR.
Please expect some delay.

srijan-deepsource:
Okay. I think everything looks good now.

@hvy: I have removed the config as you've asked. Please take a final look?

Also, I would really appreciate it if you could drop any suggestion on what we can do better to onboard you (or Optuna) on DeepSource in the future. :)

hvy (approved):
No worries, and thanks for coming back to the PR. The current changes looks simple and LGTM. Let me merge them right away.

I find it great that DeepSource can check for various code smell and I think it'll be a candidate if we were to introduce a more sophisticated static analyzer in the future. It's just that we have not yet found enough needs for one. Personally, I'd love to see this becoming available as something you can run locally (maybe with some request cap). I think local reproduction is quite important for many developers here 🙂

PR #2013, opened 2020-11-11
Add `__all__` to reexport modules explicitly

toshihikoyanase:
## Motivation

The current code base employs implicit reexport in `**/__init__.py`, but it causes mypy errors when we specify `--strict` option. I encountered this error when I submitted a [PR](https://github.com/facebookresearch/hydra/pull/1132) which adds Optuna plugin to [`hydra`](https://hydra.cc/).
According to the [mypy reference](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport), we can avoid the error if we reexport modules explicitly. 

## Description of the changes

This PR defines `__all__` in `**/__init__.py`. Alternatively, we can use `from X import Y as Y` to export `X.Y` explicitly. I chose `__all__` because we can see `__all__` in some modules such as `optuna.importance` and `oputna.integration`.

### DIFF
--- a/optuna/trial/__init__.py
+++ b/optuna/trial/__init__.py
@@ -6,4 +6,11 @@ from optuna.trial._state import TrialState  # NOQA
 from optuna.trial._trial import Trial  # NOQA
 
 
-__all__ = ["BaseTrial", "FixedTrial", "create_trial", "FrozenTrial", "TrialState", "Trial"]
+__all__ = [
+    "BaseTrial",
+    "FixedTrial",
+    "FrozenTrial",
+    "Trial",
+    "TrialState",
+    "create_trial",
+]
--- a/optuna/trial/__init__.py
+++ b/optuna/trial/__init__.py
@@ -4,3 +4,6 @@ from optuna.trial._frozen import create_trial  # NOQA
 from optuna.trial._frozen import FrozenTrial  # NOQA
 from optuna.trial._state import TrialState  # NOQA
 from optuna.trial._trial import Trial  # NOQA
+
+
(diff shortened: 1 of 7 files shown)


hvy (approved):
LGTM!

Some comments on the cosmetics that we could potentially address.

- Add a trailing comma in the list in `trial/__init__.py` to force line breaks
- Order the elements in the lists lexicographically

toshihikoyanase:
@hvy Thank you for your review. I fixed the code styles according to your suggestion. PTAL.

hvy (approved):
LGTM!

hvy:
Ops, seems to be conflicting now.

optuna/trial/__init__.py
```diff
@@ -4,3 +4,13 @@
 from optuna.trial._frozen import FrozenTrial  # NOQA
 from optuna.trial._state import TrialState  # NOQA
 from optuna.trial._trial import Trial  # NOQA
+
```
HideakiImamura, line 7: `# NOQA`s can be removed samely as `optuna/__init__.py`.

HideakiImamura (commented):
Thanks for the PR. I have a minor comment.

optuna/trial/__init__.py
toshihikoyanase: Oops! I removed it in https://github.com/optuna/optuna/pull/2013/commits/671748794c76e1a12c7d330f20e323b98d71c702.

toshihikoyanase:
Thank you for your review. I applied your suggestions and resolved conflicts. PTAL.

HideakiImamura (approved):
LGTM.

Issue #2012, opened 2020-11-11
Wrap MLflow integration with try block to catch exceptions.


PhilipMay:
Hi,
when using the MLflow integration it might happen that the callback wants to log something to MLflow but can not connect by http. Then an exception is thrown which cancels the optimization.

My question is: Should the MLflow calls be wrapped in an try block to catch those exceptions and just log a warning without throwing it out?

What do you think?

PhilipMay:
Or even better: Maybe do do this change where Optuna is calling the callbacks and wrap them there.

PhilipMay:
@drobison00 what do you think?

PhilipMay:
I would apply the change here: https://github.com/optuna/optuna/blob/81840c2e08f452bd5ac959afaca0fee006bdb44e/optuna/_optimize.py#L170
add a try, catch exceptions and log them as a warning.

PhilipMay:
PR made in #2017

crcrpar:
@harupy Do you have any idea on this? I'm asking you because I think you are a committer of mlflow.

PhilipMay:
> @harupy Do you have any idea on this? I'm asking you because I think you are a committer of mlflow.

Yes - feedback would be nice. My solution (see PR) is not targeted at MLflow but a generic solution for all exceptions in Callbacks that might happen.

crcrpar:
Yes, your pull request would fix the error which this issue reports, but I have one comment which is now on your pull request.

Anamitr:
I could also use such feature.
Perhaps as configuration field "treat_mlflow_errors_as_warnings".

Exp. in case of a long experiment, if MLflow tracking server goes down, I want experiment to continue, without MLflow.

PR #2017, opened 2020-11-13
Wrap callbacks with try and log exceptions

PhilipMay:
This is the PR ragarding #2012

It wants to avoid a situation where a callback throws an exception and the training is stopped.
This might happen when MLflow Integration wants to use remote logging and can not connect to the tracking server (HTTP timeout) for example. This should not stop the training...

## TODO
- [ ] add docstring for `catch_callback_errors`

### DIFF
--- a/optuna/_optimize.py
+++ b/optuna/_optimize.py
@@ -36,6 +36,7 @@ def _optimize(
     callbacks: Optional[List[Callable[["optuna.Study", FrozenTrial], None]]] = None,
     gc_after_trial: bool = False,
     show_progress_bar: bool = False,
+    catch_callback_errors: bool = False,
 ) -> None:
     if not isinstance(catch, tuple):
         raise TypeError(
@@ -63,6 +64,7 @@ def _optimize(
                 reseed_sampler_rng=False,
                 time_start=None,
                 progress_bar=progress_bar,
+                catch_callback_errors=catch_callback_errors,
             )
         else:
             if show_progress_bar:
@@ -129,6 +131,7 @@ def _optimize_sequential(
     reseed_sampler_rng: bool,
     time_start: Optional[datetime.datetime],
     progress_bar: Optional[pbar_module._ProgressBar],
+    catch_callback_errors: bool = False,
 ) -> None:
     if reseed_sampler_rng:
         study.sampler.reseed_rng()
@@ -170,6 +173,8 @@ def _optimize_sequential(
                 try:
                     callback(study, frozen_trial)
                 except Exception as e:
+                    if not catch_callback_errors:
+                        raise
                     _logger.error(
                         "Exception raised while callback execution! "
                         f"Exception: {e} Callback: {callback}",
--- a/optuna/_optimize.py
+++ b/optuna/_optimize.py
@@ -167,7 +167,13 @@ def _optimize_sequential(
         if callbacks is not None:
             frozen_trial = copy.deepcopy(study._storage.get_trial(trial._trial_id))
             for callback in callbacks:
-                callback(study, frozen_trial)
+                try:
+                    callback(study, frozen_trial)
+                except Exception as e:
+                    _logger.error(
+                        f"Exception raised while callback execution! Exception: {e} Callback: {callback}",
+                        exc_info=True,
+                    )
 
         if progress_bar is not None:
             progress_bar.update((datetime.datetime.now() - time_start).total_seconds())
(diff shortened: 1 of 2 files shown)


PhilipMay:
Before I continue with this PR and fix the CI errors I would like to have feedback.
@toshihikoyanase @hvy Is this a good idea? Would you merge it?

crcrpar (commented):
I suspect the solution would let the errors that need fixes go through. Like, for example, it makes it harder to detect the errors due to the wrong arguments to the third-party library's functions. So I'd like the fix to only target MLflow. :)

harupy:
Can we make the proposed behavior optional like below?

```python
try:
    callback(...)
except Exception as e:
    if not ignore_callback_errors:
        raise
    _logger.error(...)
```

PhilipMay:
Made the change suggested by @harupy - still would like to get feedback if I should continue in this direction.

crcrpar:
@HideakiImamura @toshihikoyanase @hvy Could some of you give feedback on this?

hvy:
Not sure I've followed all the details but I would also suggest scoping this fix to the `MLflowCallback` if that is technically possible. If it's the case that the callback alone can't handle it, we could discuss changes to `Study.optimize`. However, this will likely take more time to wrap up since it's a very central API. 

Say we would discuss `Study.optimize`, I'm a bit skeptical about introducing a boolean. I'd suggest we at least try to keep it symmetrical to the `catch` parameter.  We could also consider making the `catch` parameter catch exceptions from callbacks (at all times) as well, instead. Moreover, I'm a bit worried that this'll make the API/logic unnecessary complicated. Again, I'd wait with this discussion if possible.

toshihikoyanase:
I agree with @hvy. The change of `Study.optimize` can affect many users, so I'd like to be a bit careful about changes.

IMO, we can keep the execution of callbacks as it is since it is simpler than the execution of objective functions.

*1. Results of callbacks are not stored in storage*

The Optuna need to save `TrialState` appropriately depending on `catch` of `Study.optimize`. (i.e., if an error is included in the `catch`, `TrialState` should be `COMPLETE`. Otherwise, it should be `FAIL`.) Users do not change the trial states directly, so I think it is reasonable to provide the `catch` argument. On the other hand, execution results of callbacks are not stored in storage, so users can handle errors in callbacks by themselves.

*2. Error handling depends on cases*

I'm not sure if users want to ignore errors. They may want more flexible error handling. In case of connection error like #2012, they want to retry the execution of callbacks. So, I personally feel a simple wrapper can be suitable for this purpose.

In the case of #2012, we can wrap the execution of `MLflowCallback` with try-except as follows:

```python
mlflc = MLflowCallback(
    tracking_uri="https://127.0.0.1",
    metric_name="my metric score"
)

def wrapped_callback(study, trial):
    try:
        mlflc(study, trial)
    except requests.exceptions.ConnectionError as e:
        _logger.error(f"The following error was raised in MLflowCallback: {e}")
```

<details>
<summary>Full code of wrapped callback approach</summary>

```python
import optuna

import requests

def objective(trial):
    return (x - 2) ** 2


mlflc = MLflowCallback(
    tracking_uri="https://127.0.0.1",
    metric_name="my metric score"
)

def wrapped_callback(study, trial):
    try:
        mlflc(study, trial)
    except requests.exceptions.ConnectionError as e:
        _logger.error(
            f"The following error was raised in MLflowCallback: {e}"
        )

study.optimize(objective, n_trials=10, callbacks=[wrapped_callback])
```
</details>

By extending the wrapper, we can easily retry the execution of `MLflowCallback` as follows:

```python
def wrapped_callback(study, trial):
    ex = None
    for i in range(0, 5):
        try:
            mlflc(study, trial)
            print("hoge")
            return
        except Exception as e:
            _logger.error(f"Retry {i}")
            ex = e
    _logger.error(f"The following error was raised in MLflowCallback: {ex}")
```

PhilipMay:
@hvy @toshihikoyanase could you please have a quick look at my implementation? IMO it changes nothing at the current behavior of code but simply adds the possibility to set a flag to catch errors...

himkt:
@PhilipMay Sorry for delayed response.

We want to keep `Study.optimize` simple (cf. https://github.com/optuna/optuna/pull/2017#issuecomment-728790184, https://github.com/optuna/optuna/pull/2017#issuecomment-728860004). This means we don't want to do stuffs such as handling exceptions in `Study.optimize`.

On the other hand, we agree with you that `Study.optimize` should not be stopped when some callbacks raises an exception. In particular, a callback invoking some APIs to send/receive data over the network could fail. For this case, I think a callback should handle exception (not `Study.optimize`).

So, could you please revise the PR to move the implementation of error handling to `MLFlowCallback.__call__`?

Issue #1928, opened 2020-10-14
Migration to GitHub Actions


crcrpar:
Upon #1755 gets merged, we will remove tests and checks from CircleCI but leave documentation workflow as is since we can swiftly check the newly build documentation with CircleCI.

crcrpar:
leave document and rdbstorage as they are and let's delete the other jobs.

jeromepatel:
@crcrpar I think this issue can be closed as both the tasks have been completed. If it needs any further changes/modification let me know, I would be happy to work on that.

PR #2027, opened 2020-11-16
Migrate tests and coverage to GitHub Actions

crcrpar:
## Motivation

Per title.

## Description of the changes

- remove tests-python?? from .circleci/config.yml
- add coverage job to github actions
- update CONTRIBUTING.md as per the above changes

ref: #1928


HideakiImamura (approved):
Thanks for the PR as always. I confirmed that the `Tests/coverage` job is appropriately executed and `tests-python??` jobs are removed. LGTM.

CONTRIBUTING.md
```diff
@@ -94,7 +94,11 @@ pytest tests/${TARGET_TEST_FILE_NAME}
 
 ## Continuous Integration and Local Verification
 
+Optuna repository uses GitHub Actions and CircleCI.
+
 CircleCI is used for continuous integration.
+Currently, we are migrating to GitHub Actions but still we use CirclCI for some tests listed below.
+`document` still exists because it makes it much easier to check built documentation and `tests-rdbstorage` is there because CircleCI better provides RDB storages.
```
hvy, line 101: How about explaining in what way CircleCI is "better", if not for the record here on GitHub?

.github/workflows/tests.yml
```diff
@@ -71,3 +68,65 @@ jobs:
     - name: Tests
       run: |
         pytest tests --ignore tests/integration_tests
+
+  coverage:
```
hvy, line 72: How about creating a separate workflow, e.g. `Tests (Coverage)` considering the fact that unit tests are already split depending on whether they're testing integration modules or not. Also, I think we can skip the cron scheduling for the code coverage.

hvy (commented):
Thanks for the your PR as always. Left some comments!

CONTRIBUTING.md
crcrpar: "CircleCI currently supports more various RDB storages than GitHub Actions" sounds better to you?

crcrpar:
Okay, CircleCI documentation build job somehow failed, but I think it doesn't matter a lot.

hvy:
Seemingly canceled somehow, let me rerun it.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,70 @@
+name: Coverage
+
+on:
+  push:
+    - master
```
hvy, line 5: I think you need to specify the branches.
```suggestion
    branches:
      - master
```

crcrpar:
Oh, somehow my push isn't reflected at all.

.github/workflows/tests.yml
```diff
@@ -46,9 +46,6 @@ jobs:
         key: ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}-v1
         restore-keys: |
           ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}
-    - name: Setup environment
-      run: |
-        sudo apt-get update
```
hvy, line 51: Is this change related to codecov?

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,71 @@
...
+      env:
+        OMP_NUM_THREADS: 1
+      run: |
+        pytest --cov=optuna --cov-report=xml tests
+
+    - name: Multi-node tests
+      env:
+        OMP_NUM_THREADS: 1
+      run: |
+        mpirun -n 2 -- pytest tests/integration_tests/test_chainermn.py
```
hvy, line 66: Probably outside the scope of this PR but I noticed that this pytest execution isn't recording the coverage?

.github/workflows/tests.yml
crcrpar: Negative. I thought this step should be removed as there's no apt install in this job. Am I stupid here?

.github/workflows/coverage.yml
crcrpar: Okay, then remove this. I copied things from Circle CI config, sorry.

.github/workflows/tests.yml
hvy: We could separated it into its own PR if that's the case. But, it's a small change so it's not really a strong opinion if you'd be busy with other things. I by the way also think it's for the ChainerMN tests (integration tests which are not executed here).

hvy (approved):
LGTM!

Let me merge this PR since the changes look good although there are minor comments left.

.github/workflows/coverage.yml
hvy: Created an issue at https://github.com/optuna/optuna/issues/2038.

PR #2035, opened 2020-11-18
Remove py35 from black config in pyproject.toml

crcrpar:
Changes

- remove py35 from black's targets
- add trailing commas to multiline arguments (ref: https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas)

ref: #1067


himkt (approved):
Thank you for the catch. LGTM

c-bata (approved):

PR #2037, opened 2020-11-18
Update logging in pytorch lightning example

pbmstrk:
## Motivation

I've been using optuna with Pytorch Lightning - been working great so far. I noticed that the pytorch lightning example is outdated, i.e. pytorch lightning has introduced a new API for logging and the old method of logging has been deprecated. For reference, the new logging [documentation](https://pytorch-lightning.readthedocs.io/en/latest/logging.html#) and the [mnist example](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pl_examples/basic_examples/mnist_classifier.py) from the lightning repo.

## Description of the changes

Logging by returning  `{log: dict}` from an `epoch_end` method is deprecated in pl and has been updated to use `self.log(...)`. Using `self.log(...)` also means that the `validation_epoch_end` method is no longer required as metrics are aggregated automatically.

The minimum pytorch lightning version to run the example would now be 1.0.2.

### DIFF
--- a/examples/pytorch_lightning_simple.py
+++ b/examples/pytorch_lightning_simple.py
@@ -30,8 +30,8 @@ import optuna
 from optuna.integration import PyTorchLightningPruningCallback
 
 
-if version.parse(pl.__version__) < version.parse("0.8.1"):
-    raise RuntimeError("PyTorch Lightning>=0.8.1 is required for this example.")
+if version.parse(pl.__version__) < version.parse("1.0.2"):
+    raise RuntimeError("PyTorch Lightning>=1.0.2 is required for this example.")
 
 PERCENT_VALID_EXAMPLES = 0.1
 BATCHSIZE = 128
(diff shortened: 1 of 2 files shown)


crcrpar (commented):
Oh, thank you for doing this. We have overlooked the deprecation warning messages in daily examples job 😅 

Could you also update the requirements of example, testing, and integration in setup.py?

reference for other reviewers: https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md#fixed-6 - why this example requires >1.0.2

pbmstrk:
@crcrpar ah yes i missed that - I've updated the version requirements

crcrpar (approved):
yup, confirmed this chagnes removes Deprecateion Warning locally, thank you.

crcrpar:
@pbmstrk In this example, `training_step` still returns `dict` but the v1.0.7 example https://github.com/PyTorchLightning/pytorch-lightning/blob/1.0.7/pl_examples/basic_examples/mnist_classifier.py returns just `Tensor`, don't we better follow this style?

toshihikoyanase (approved):
LGTM. Thank you for your contribution.

Notes:
- did not work with pytorch_lightning==1.0.1 as expected
- worked with pytorch_lightning==1.0.2
- worked with the latest pytorch_lightning (1.0.7)
- the official example has been moved to [here](https://github.com/PyTorchLightning/pytorch-lightning/blob/69b994919200581774de66ebacd510918e85bb1f/pl_examples/basic_examples/simple_image_classifier.py#L45-L49)

toshihikoyanase:
@crcrpar Sorry, but I'm not aware of your latest comment https://github.com/optuna/optuna/pull/2037#issuecomment-731514935. I created a follow-up PR #2043, so please add your comment there.

PR #2043, opened 2020-11-21
Change return type of `training_step` of PyTorch Lightning example

toshihikoyanase:
## Motivation
Please refer to https://github.com/optuna/optuna/pull/2037#issuecomment-731514935.

## Description of the changes
This PR changed the return type of `LightningNet.training_step` from `Dict` to `Tensor`.

CC @pbmstrk I'm not sure this change follows the lightning's convention. Please feel free to add your comments.

### DIFF
--- a/examples/pytorch_lightning_simple.py
+++ b/examples/pytorch_lightning_simple.py
@@ -101,7 +101,7 @@ class LightningNet(pl.LightningModule):
     def training_step(self, batch, batch_nb):
         data, target = batch
         output = self.forward(data)
-        return {"loss": F.nll_loss(output, target)}
+        return F.nll_loss(output, target)
 
     def validation_step(self, batch, batch_nb):
         data, target = batch


pbmstrk:
Since, PL 1.0.0 you can return either a tensor or dict from `training_step` (see [here](https://github.com/PyTorchLightning/pytorch-lightning/releases/tag/0.10.0)), so both methods should be fine. But as @crcrpar pointed out in #2037, the [mnist example](https://github.com/PyTorchLightning/pytorch-lightning/blob/1.0.7/pl_examples/basic_examples/mnist_classifier.py#L43-L47) and the [docs](https://pytorch-lightning.readthedocs.io/en/stable/converting.html#find-the-train-loop-meat) both only return a tensor, so it may be better to align the example with what the official PL examples do.

crcrpar (approved):
@pablete Thank you for the information 😃

toshihikoyanase:
@pbmstrk Thank you for your comment!

hvy (approved):
Thanks, changes LGTM!

Issue #2034, opened 2020-11-18
Support Python 3.9 with integration modules and optional dependencies


crcrpar:
Monitor below libraries to keep tests-integration and examples ci up-to-date

"locally confirmed" means `docker run --rm python:3.9-buster bash -c "pip install <library>"` did not raise exceptions. I didn't check the PyPI or conda page.
I have NOT checked the corresponding examples run successfully or not.

Check means the corresponding example and integration if exists work and CI gets updated.

- [x] scikit-learn: locally confirmed
- [ ] PyTorch: v1.8.0 has been released. Once the currently supported python versions get v1.8.0, let's do this. rel: https://github.com/optuna/optuna/issues/2454
- [ ] TensorFlow: https://github.com/tensorflow/tensorflow/issues/44485 (trending toward TF2.5?)
- [ ] mxnet: I know nothing.
- [ ] catboost: https://github.com/catboost/catboost/issues/1491#issuecomment-729521190
- [ ] dask: locally confirmed
- [ ] lightgbm: locally confirmed
- [ ] xgboost: locally confirmed
- [ ] mlflow: locally confirmed
- [ ] plotly: Not in their `.circleci/config.yml`

Optuna itself can be installed: #1908

hvy:
Updated the title to be a bit more specific.

0x41head:
Is anyone working on this issue currently ? If not I would like to pick it up.

crcrpar:
No.
It's much more than appreciated ☺️

0x41head:
@crcrpar I was confused as to what I should do with packages like xgboost, which are already updated to the latest version and are supported python 3.9. Do I need to make any changes anywhere ?

crcrpar:
In short, what we need to do is to update https://github.com/optuna/optuna/blob/master/.github/workflows/tests-integration.yml and https://github.com/optuna/optuna/blob/master/.github/workflows/examples.yml to assure that Optuna's integrations and examples work well on Python3.9

nzw0301:
I believe that this PR has been resolved. Let me close this.

PR #2051, opened 2020-11-27
Clean up `setup.py` requirements

crcrpar:
I think this is a good step before introducing Python3.9 check in setup.py (#2034).


crcrpar:
Confirmed locally.

```
~/ghq/github.com/optuna/optuna-1 setup/relax-py38-conditions*
py39 ❯ docker run -it -v $(pwd):/workspaces optuna/optuna:py3.8-dev /bin/bash
root@5d7377d6dcfe:/workspaces# pip install --progress-bar off -q -e '.[example]'
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

dask-ml 1.7.0 requires scikit-learn>=0.23, but you'll have scikit-learn 0.22.2.post1 which is incompatible.
root@5d7377d6dcfe:/workspaces# python examples/dask_ml_simple.py &> /dev/null
root@5d7377d6dcfe:/workspaces# cd examples/allennlp/
root@5d7377d6dcfe:/workspaces/examples/allennlp# python allennlp_jsonnet.py &> /dev/null
root@5d7377d6dcfe:/workspaces/examples/allennlp# python allennlp_simple.py &> /dev/null
root@5d7377d6dcfe:/workspaces/examples/allennlp# exit
```

crcrpar:
Requesting a review from @himkt as this pull request (unintentionally) includes the change to AllenNLP tests.

hvy:
Markers were recently introduced by https://github.com/optuna/optuna/pull/2105 . How do you think we should proceed with this PR? Perhaps resolve the conflicts and focus on the constraint relaxations?

crcrpar:
Yes, I'll do it that way.

tests/integration_tests/allennlp_tests/test_allennlp.py
```diff
@@ -230,7 +230,7 @@ def test_allennlp_pruning_callback() -> None:
 
         def objective(trial: optuna.Trial) -> float:
             reader = allennlp.data.dataset_readers.TextClassificationJsonReader(
-                tokenizer=allennlp.data.tokenizers.SpacyTokenizer()
```
himkt, line 233: [Note]

This change was introduced for avoiding the probabilistic error in loading a spacy model.
`test_allennlp_pruning_callback` sometimes fails with the following error.

```
    def load_model(name, **overrides):
        """Load a model from a shortcut link, package or data path.
        name (unicode): Package name, shortcut link or model path.
        **overrides: Specific overrides, like pipeline components to disable.
        RETURNS (Language): `Language` class with the loaded model.
        """
        data_path = get_data_path()
        if not data_path or not data_path.exists():
            raise IOError(Errors.E049.format(path=path2str(data_path)))
        if isinstance(name, basestring_):  # in data dir / shortcut
            if name.startswith("blank:"):  # shortcut for blank model
                return get_lang_class(name.replace("blank:", ""))()
            if name in set([d.name for d in data_path.iterdir()]):
                return load_model_from_link(name, **overrides)
            if is_package(name):  # installed as package
                return load_model_from_package(name, **overrides)
            if Path(name).exists():  # path to model data directory
                return load_model_from_path(Path(name), **overrides)
        elif hasattr(name, "exists"):  # Path or Path-like to model data
            return load_model_from_path(name, **overrides)
>       raise IOError(Errors.E050.format(name=name))
E       OSError: [E050] Can't find model 'en_core_web_sm'. It doesn't seem to be a shortcut link, a Python package or a valid path to a data directory.
/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/spacy/util.py:175: OSError
```

I didn't investigate this error in detail but it could use a simple whitespace tokenizer instead of SpacyTokenizer, which doesn't need any model resource. (Should we make another PR?)

himkt (approved):
LGTM.

@hvy Please take a look. :pray:

hvy (approved):
Sorry the having my review delayed. LGTM since examples have been verified as well. Libraries such as Dask officially support Python 3.8 (although not 3.9) too so seems reasonable.

PR #2054, opened 2020-11-29
Allow passing additional arguments when running tox

harupy:
Signed-off-by: harupy <17039389+harupy@users.noreply.github.com>

## Motivation

Allows passing additional argumetns when running tox like ([ref](https://tox.readthedocs.io/en/latest/example/general.html#interactively-passing-positional-arguments)):

```
tox -e test-py38 -- --verbose
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,7 +16,7 @@ extras =
 deps =
     scikit-optimize
     cma
-commands = pytest tests --ignore tests/integration_tests
+commands = pytest tests --ignore tests/integration_tests {posargs}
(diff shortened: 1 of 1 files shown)


toshihikoyanase (approved):
Thank you for your PR. I think this fix is reasonable because it enables users to show detailed logs or to exclude some test cases tentatively.
LGTM!

himkt (approved):
Thank you for the improvement! I also tested the patch on my environment. LGTM!

PR #2055, opened 2020-11-30
Added test and fix for relative sampling failure in multivariate TPE

alexrobomind:
## Motivation
Currently, on the master branch, the following minimal example fails:

```
import optuna

def mwe(trial):
	if trial.number == 0:
		trial.suggest_uniform("param1", 0, 1)
		raise optuna.exceptions.TrialPruned()
	
	if trial.number == 1:
		trial.suggest_uniform("param2", 0, 1)
		return 0
	
	return 0

sampler = optuna.samplers.TPESampler(n_startup_trials = 1, multivariate = True)
study = optuna.create_study(sampler = sampler)

study.optimize(mwe, 3)
```

This is because the multivariate TPE uses pruned and completed trials for its relative sampling, but the search space computation only looks at completed trials. Therefore, it can include parameters missing from the pruned trials, which the TPE can't handle. This leads to the assertion `assert param_name in trial.params` in sampler.py to fail.

## Description of the changes
This PR contains three modifications:
- Modifies the assertion param_name in trial.params to be a bit more explanatory
- Adds an optional argument `include_pruned` (default False) to IntersectionSearchSpace.calculate and intersection_search_space, and sets this to True in the TPESampler.
- Adds the above failing example as a test.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -838,11 +838,7 @@ def _get_multivariate_observation_pairs(
 
         # We extract param_value from the trial.
         for param_name in param_names:
-            assert (
-                param_name in trial.params
-            ), "Parameter {} not found in trial {}, which has parameters {}".format(
-                param_name, trial.number, list(trial.params)
-            )
+            assert param_name in trial.params
             distribution = trial.distributions[param_name]
             param_value = distribution.to_internal_repr(trial.params[param_name])
             values[param_name].append(param_value)
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -183,7 +183,7 @@ class TPESampler(BaseSampler):
         self._random_sampler = RandomSampler(seed=seed)
 
         self._multivariate = multivariate
-        self._search_space = IntersectionSearchSpace()
+        self._search_space = IntersectionSearchSpace(include_pruned=True)
 
         if multivariate:
             warnings.warn(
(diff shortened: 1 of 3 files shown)


alexrobomind:
One thing to consider: Maybe it would be better to have `include_pruned` as a parameter to the search space constructor, instead of the `calculate` method, as calling it inconsistently might yield some unexpected results.

HideakiImamura (commented):
Thanks for the PR! Basically LGTM. In my opinion, the `include_pruned` should be determined at the constructer of `IntersectionSearchSpace`. Could you change it?

HideakiImamura:
Hi @alexrobomind. Can I help you to proceed with this PR?

HideakiImamura:
@alexrobomind I can take over this PR if you are busy.

mhdadk:
@HideakiImamura I actually just got got an error because of what @alexrobomind mentioned in #2316. Does this mean that I should use the TPE sampler with `multivariate = False` until this is fixed?

alexrobomind:
I'm sorry for just leaving this lying around without notice. I had multiple urgent things coming up in series and forgot to finish this PR. If the checks pass successfully, this should meet the requirements now.

alexrobomind:
@HideakiImamura Also, I wanted to apologize for not answering your previous messages. My private mailbox was full, and I am still somewhat unfamiliar with the github interface, so I did not see the notification.

optuna/samplers/_tpe/sampler.py
```diff
@@ -838,7 +838,11 @@ def _get_multivariate_observation_pairs(
 
         # We extract param_value from the trial.
         for param_name in param_names:
-            assert param_name in trial.params
+            assert (
+                param_name in trial.params
+            ), "Parameter {} not found in trial {}, which has parameters {}".format(
+                param_name, trial.number, list(trial.params)
+            )
```
HideakiImamura, line 845: Looks like unrelated changes to this PR. Could you revert it?

HideakiImamura (commented):
Thanks for the update! I'd like to merge this until v2.6.0. PTAL.

HideakiImamura:
@not522 Could you review this PR if you have time? I think we should merge PR until v2.6.0, so it would be great if you could prioritize and review this PR.

optuna/samplers/_tpe/sampler.py
alexrobomind: I reverted it to the previous expression

not522 (approved):
LGTM!

HideakiImamura (approved):
Thanks! LGTM!

PR #2024, opened 2020-11-15
Add tox.ini

c-bata:
## Motivation

To close #2010.

tox makes it easier to run tests and lint checks for newcomers because it is a widely accepted as a standard Python tool.

## Description of the changes

Add tox.ini, the setting file of tox.

```
$ tox -e flake8 -e black -e blackdoc -e mypy
...
  flake8: commands succeeded
  black: commands succeeded
  blackdoc: commands succeeded
  mypy: commands succeeded
  congratulations :)
```

```
$ tox -e test-py38
...
  test-py38: commands succeeded
  congratulations :)
```

Issue #2010, opened 2020-11-11
Add tox.ini instead of checks.sh


c-bata:
## Motivation

I think tox is a better solution of [checks.sh](https://github.com/optuna/optuna/blob/master/checks.sh).
tox makes it easier to run tests and lint checks for newcomers because it is a widely accepted as a standard Python tool.

## Description

The setting file (tox.ini) of [optuna-dashboard](https://github.com/c-bata/optuna-dashboard) project is here:

```
[tox]
envlist =
    py36
    py37
    py38
    py39
    flake8
    black
    mypy

[testenv]
basepython = python3.9
commands = python -m unittest

[testenv:py36]
basepython = python3.6

[testenv:py37]
basepython = python3.7

[testenv:py38]
basepython = python3.8

[testenv:flake8]
deps = flake8
commands = flake8 .

[testenv:black]
deps = black
commands = black --check .

[testenv:mypy]
deps = mypy
commands = mypy .
```

Usage is like this:

```
$ pip install tox
$ tox -e ALL  # run all tests and lints
...
  py36: commands succeeded
  py37: commands succeeded
  py38: commands succeeded
  py39: commands succeeded
  flake8: commands succeeded
  black: commands succeeded
  mypy: commands succeeded
  congratulations :)
```

```
$ pip install tox
$ tox -e flake8 -e black -e mypy  # run lints
...
  flake8: commands succeeded
  black: commands succeeded
  mypy: commands succeeded
  congratulations :)
```

See the tox documentation for more details.

https://tox.readthedocs.io/en/latest/

crcrpar:
Hi @c-bata,

It sounds cool to me and I have one question.
`checks.sh` has an option of `-u` with which the script fixes on behalf of us. Does tox have any equivalent options or settings?

c-bata:
No, probably tox does not provide such an option. Code formatting is the out of responsibility of tox.

crcrpar:
`checks.sh` mitigates the burden of following Optuna's guideline by providing formatting as well as code check and installing dependencies, which is enough to provide rather than tox.ini.

c-bata:
OK. Thank you for your feedback.
I'm still not sure if we should still keeps `checks.sh` on repository root (In my opinion, it's enough just updates documentation about how to run auto-formatters). But for now, I opened #2024 without removing checks.sh.

crcrpar:
> In my opinion, it's enough just updates documentation about how to run auto-formatters

[...]

PR #2058, opened 2020-12-01
Add tox to dependencies for developers

himkt:
I opened this PR as I just noticed `tox` doesn't appear in `setup.py`.

@c-bata 
What do you think about adding the new package to dependencies for checking?
If you had something intention for not including this change in your PR (#2024), I'd like your opinion.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ def get_tests_require() -> List[str]:
 def get_extras_require() -> Dict[str, List[str]]:
 
     requirements = {
-        "checking": ["black", "hacking", "isort", "mypy==0.782", "blackdoc"],
+        "checking": ["black", "hacking", "isort", "mypy==0.782", "blackdoc", "tox"],
         "codecov": ["codecov", "pytest-cov"],
         "doctest": [
             "cma",


HideakiImamura:
@c-bata Let me assign you to this PR. I hope you will take a look at it when you have time.

crcrpar:
Personally, adding tox only to checking does not sound good because (a) the current tox.ini also supports tests including integration ones, and (b) we should first clarify a little bit more about how we expect users to utilize tox.ini.

c-bata:
Thank you for mentioning me.

As @crcrpar commented, the initial impression is that we should not add `tox` to checking section. It supports not only checks but also unit tests and integration tests. It seems that there are no appropriate section in extra_requires to add tox.

How about updating CONTRIBUTING.md like the following?

https://github.com/optuna/optuna/issues/2010#issuecomment-728668093

himkt:
@crcrpar @c-bata Thank you for the comments and clarifications!

> the current tox.ini also supports tests including integration ones

It makes sense. Let's make an update to CONTRIBUTING.md. :+1:

PR #1908, opened 2020-10-07
Partial support of python 3.9

crcrpar:
## Changes

- add Python 3.9 to tests except for integration as there are outstanding libraries for Python 3.9
- add Python 3.9 to Docker build workflow
- enable Docker BuildKit as it better deals `--cache-from`

## TODOs
- monitor integration libraries to keep setup.py and CI workflows up-to-date.

PR #2063, opened 2020-12-03
Add Python 3.9 to README badge

hvy:
## Motivation

Optuna works with Python 3.9, if integration modules are excluded. Integration modules are optional in Optuna and dependent on third party libraries so it might be reasonable to consider 3.9 as supported.

C.f. https://github.com/optuna/optuna/pull/1908 for the inclusion of 3.9 in the CI and https://github.com/optuna/optuna/issues/2034 for tracking integration module support.

## Description of the changes

Adds Python 3.9 to the list of supported versions in the badge in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # Optuna: A hyperparameter optimization framework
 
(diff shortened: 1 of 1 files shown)


toshihikoyanase (approved):
LGTM. Thank you for your PR!

image

PR #2065, opened 2020-12-04
Fix Read the Docs build failure caused by `pip install --find-links`

hvy:
## Motivation

Read the Docs seemingly does not support `pip install` with `-f, --find-links`, and causes the current build to fail. The reason `-f` is required is that the no-CUDA PyTorch is used for the tutorials.

## Description of the changes

Changes the requirements to the standard PyTorch with CUDA that does not require `-f`.

Note that this will increase the build time. It worked in my fork https://readthedocs.org/projects/hvy-optuna/builds/12484658/ even when the build time exceeded 30 minutes. However, I'm not sure this complies with the official requirements stated by Read the Docs https://docs.readthedocs.io/en/stable/builds.html.

## Alternative

As far as I know, we could

a. avoid using PyTorch in the tutorial
b. pre-build the tutorial

## References

- Read the Docs configuration in Optuna https://github.com/optuna/optuna/blob/master/.readthedocs.yml
- Read the Docs configuration documentations https://docs.readthedocs.io/en/stable/config-file/index.html

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -13,8 +13,11 @@ formats: all
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  # `sphinx` requires either Python >= 3.8 or `typed-ast` to reflect type comments
+  # `sphinx` requires either Python >= 3.8 or `typed-ast` to reflect type comments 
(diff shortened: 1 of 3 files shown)


crcrpar (approved):
thanks

HideakiImamura (approved):
LGTM.

Issue #1795, opened 2020-09-08
readthedocs pdf build is failing


hvy:
The pdf builds on readthedocs have been failing for the last couple of days. It was successfully built for v2.0.0 but is broken in `master` as well as in the latest v2.1.0 release. 

It might be worth trying to reproduce the error with the readthedocs' official Docker image https://docs.readthedocs.io/en/stable/builds.html#docker-images.

harupy:
[The failed build log](https://readthedocs.org/projects/optuna/builds/11765639) says:

```
! pdfTeX error (ext4): \pdfendlink ended up in different nesting level than \pd
fstartlink.
\AtBegShi@Output ...ipout \box \AtBeginShipoutBox 
                                                  \fi \fi 
l.4203 \subsubsection*{Methods}
                               
!  ==> Fatal error occurred, no output PDF file produced!
```

[This page](https://www.overleaf.com/learn/latex/Questions/What_does_%22%5Cpdfendlink_ended_up_in_different_nesting_level_than_%5Cpdfstartlink%22_mean%3F) says the error message means:

> The error message "\pdfendlink ended up on a different page" is usually due to a hyperlink (perhaps a citation, or a URL, or cross-reference) somewhere in your PDF, **that's breaking across two page (i.e. starting near the end of one page and end on the next). (Citations are possible culprits.) Unfortunately, pdfLaTeX can't deal with this**.

harupy:
The doc build has been failing since this commit: https://github.com/optuna/optuna/commit/05365bbf17dcff0bc0bf242ff23cc3b77677b932

hvy:
Thanks for your quick debugging. Do you know how we can address the issue, for instance if there's some configuration in sphinx that we can utilize? My understanding is that it could start to occur when we make a change to the documentation anywhere, and we additionally can't tell exactly where the error occurs.

harupy:
I have no idea why this error occurs lol. One possible reason is that the example added by https://github.com/optuna/optuna/commit/05365bbf17dcff0bc0bf242ff23cc3b77677b932 pushes donw the hyperlinks in `Args`, `Returns` and `See also` sections and one of them breaks across two pages.

toshihikoyanase:
[...]

PR #2068, opened 2020-12-04
Fix lt symbol

KoyamaSohei:
## Motivation

https://github.com/optuna/optuna/issues/1795#issuecomment-728605927

> 2. Syntax error of math expression

```
! Undefined control sequence.
l.2213 \lt
           \mathsf{high}\), where \(k\) is an integer.
The control sequence at the end of the top line
of your error message was never \def'ed. If you have
misspelled it (e.g., `\hobx'), type `I' and the correct
spelling (e.g., `I\hbox'). Otherwise just continue,
and I'll forget about whatever was undefined.
```
Fix this syntax error.

## Description of the changes

`\lt` means  __less than__  in ~~TeX~~ MathJax (not LaTeX) , so replace `\\lt` with `<`.

Related: 
http://docs.mathjax.org/en/latest/input/tex/html.html?highlight=lt#html-special-characters

Sphinx escape `<` to `&lt;`  on html building , so we need not using `\lt`.

## Preview

-  html build

Screenshot from 2020-12-04 15-29-11

- pdf build

Screenshot from 2020-12-04 15-38-30

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -190,7 +190,7 @@ class DiscreteUniformDistribution(BaseDistribution):
     .. note::
         If the range :math:`[\\mathsf{low}, \\mathsf{high}]` is not divisible by :math:`q`,
         :math:`\\mathsf{high}` will be replaced with the maximum of :math:`k q + \\mathsf{low}
-        \\lt \\mathsf{high}`, where :math:`k` is an integer.
+        < \\mathsf{high}`, where :math:`k` is an integer.
 
     Attributes:
         low:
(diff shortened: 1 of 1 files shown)


toshihikoyanase (commented):
Thank you for your investigation. The HTML file was rendered as expected and I'll check the PDF next.
image

toshihikoyanase (approved):
The PDF document was also built as expected and I checked the log did not contain the error message about the `<` symbol. LGTM!
Thank you for your contribution.

image

toshihikoyanase:
@chenghuzi I think you may be interested in #1795. Please feel free to add comments if you have any.

KoyamaSohei:
@toshihikoyanase 

Thank you for taking your time and review!

HideakiImamura (approved):
Thanks for the PR! LGTM!

PR #2069, opened 2020-12-05
Unpin mypy version

### LINKED ISSUE #1914: Fix type hints for `mypy == 0.790`
See https://github.com/optuna/optuna/pull/1913. 

Some type hints are not complying to the latest `mypy == 0.790` and should be addressed.

himkt:
Close #1914 

## Motivation
This PR unpins the mypy version.
For upgrading mypy to v0.790, we have to update some source codes for considering `_SupportLessThan` introduced in https://github.com/python/typeshed/pull/4192 (see also: https://github.com/python/typeshed/pull/4155).

I use `typing.cast`, which has no performance impact on run-time.
(ref. https://docs.python.org/3/library/typing.html#typing.cast)

## Description of the changes
- 0521520 Unpin mypy version
- 5496213 df51b19 AllenNLP specific changes
- 32c454c Explicitly casting values to `float`
- 2f80a57 Introduce `SortableParamValueSequenceType` and cast variables

### DIFF
--- a/optuna/multi_objective/samplers/_motpe.py
+++ b/optuna/multi_objective/samplers/_motpe.py
@@ -641,7 +641,7 @@ def _get_observation_pairs(
 
         # Convert all objectives to minimization
         score = [
-            cast(float, v) if d == StudyDirection.MINIMIZE else -v
+            cast(float, v) if d == StudyDirection.MINIMIZE else -cast(float, v)
             for d, v in zip(study.directions, trial.values)
         ]
 
--- a/optuna/multi_objective/samplers/_motpe.py
+++ b/optuna/multi_objective/samplers/_motpe.py
@@ -1,6 +1,7 @@
 import math
 from typing import Any
 from typing import Callable
+from typing import cast
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -640,7 +641,7 @@ def _get_observation_pairs(
 
         # Convert all objectives to minimization
         score = [
-            v if d == StudyDirection.MINIMIZE else -v  # type: ignore
+            cast(float, v) if d == StudyDirection.MINIMIZE else -v
             for d, v in zip(study.directions, trial.values)
         ]
 
--- a/optuna/samplers/_grid.py
+++ b/optuna/samplers/_grid.py
@@ -2,6 +2,7 @@ import collections
 import itertools
 import random
 from typing import Any
+from typing import cast
 from typing import Dict
 from typing import List
 from typing import Mapping
@@ -16,6 +17,7 @@ from optuna.trial import FrozenTrial
 
 
 GridValueType = Union[str, float, int, bool, None]
+SortableParamValueSequenceType = Union[List[str], List[float], List[int], List[bool]]
 
 
 _logger = get_logger(__name__)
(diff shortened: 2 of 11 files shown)


optuna/samplers/_grid.py
```diff
@@ -98,6 +100,8 @@ def __init__(self, search_space: Mapping[str, Sequence[GridValueType]]) -> None:
 
         self._search_space = collections.OrderedDict()
         for param_name, param_values in sorted(search_space.items(), key=lambda x: x[0]):
+            param_values = cast(SortableParamValueSequenceType, param_values)
```
himkt, line 103: I'm not sure if we have to check types of values in param_values explicitly.
I introduce `SortableParamValueSequenceType` without explicitly checking now.

toshihikoyanase:
Thank you for your PR.
@hvy If you have time, could you review this PR as an author of #1914?

optuna/samplers/_grid.py
hvy: So we need it for https://github.com/optuna/optuna/blob/master/optuna/samplers/_grid.py#L221-L223. We can probably replace it with a set comparison, do you want to do that in this PR? It might stretch the scope of this PR somewhat since we currently seemingly allow duplicate items in the grid. It also affect performance, although this in practice shouldn't be an issue.

hvy (approved):
Thanks for tackling the issue. I wasn't aware of the fact that we could use `cast` to address this.

LGTM, although I left an open question https://github.com/optuna/optuna/pull/2069#discussion_r537306639.

optuna/multi_objective/samplers/_motpe.py
```diff
@@ -640,7 +641,7 @@ def _get_observation_pairs(
 
         # Convert all objectives to minimization
         score = [
-            v if d == StudyDirection.MINIMIZE else -v  # type: ignore
+            cast(float, v) if d == StudyDirection.MINIMIZE else -v  # type: ignore
```
HideakiImamura, line 644: [Just a question] Is this `# type: ignore` necessary? We may avoid it by using `cast(float, -v)`.

setup.py
```diff
@@ -55,7 +55,7 @@ def get_tests_require() -> List[str]:
 def get_extras_require() -> Dict[str, List[str]]:
 
     requirements = {
-        "checking": ["black", "hacking", "isort", "mypy==0.782", "blackdoc"],
+        "checking": ["black", "hacking", "isort", "mypy", "blackdoc"],
```
HideakiImamura, line 58: `mypy` at `tox.ini` would also need to be unpinned.

HideakiImamura (commented):
Thanks for the PR! I have two minor comments. PTAL.

optuna/multi_objective/samplers/_motpe.py
himkt: Nice catch! I missed this as `# type: ignore` was placed here.
Mypy isn't happy with `cast(float, -v)`, because `Optional[float]` (`v`) doesn't support unary `-`.
51964b9  I changed here to use `-cast(float, v)` although it may feel strange.

setup.py
himkt: You're right. Thank you. dab902e

optuna/samplers/_grid.py
himkt: Thank you for the comment and sorry for the late reply. As you pointed out, I also think we can implement `_same_search_space` with a set comparison. Let me make another PR, as I think it would be better to focus only on mypy fixing in this PR.

himkt:
@HideakiImamura I'm sorry for not responding to you for a long while. Because many conflicting changes were introduced (due to my delay), I re-created the branch and restored changes with cherry-picking.

Please take a look again.

HideakiImamura (approved):
Thanks for the update! LGTM!

PR #2070, opened 2020-12-05
Change the order of arguments and default values of MultiObjectiveStudy.optimize

cafeal:
## Motivation
The order of arguments and default values are different between Study.optimize and MultiObjectiveStudy.optimize.j
It may sometimes cause an unexpected behavior.

## Description of the changes

Make it same with Study.optimize

### DIFF
--- a/optuna/multi_objective/study.py
+++ b/optuna/multi_objective/study.py
@@ -275,12 +275,12 @@ class MultiObjectiveStudy(object):
     def optimize(
         self,
         objective: ObjectiveFuncType,
-        timeout: Optional[int] = None,
         n_trials: Optional[int] = None,
+        timeout: Optional[int] = None,
         n_jobs: int = 1,
         catch: Tuple[Type[Exception], ...] = (),
         callbacks: Optional[List[CallbackFuncType]] = None,
-        gc_after_trial: bool = True,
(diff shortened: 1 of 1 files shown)


toshihikoyanase:
Thank you for your PR. IMO, the proposal is reasonable.

On the other hand, @HideakiImamura is working on the refactoring of the multi-objective optimization, e.g., #2030 and  #1994. In short, multi-objective study will be integrated into the `optuna.study.Study` and the `optuna.multi_objective` module will be deprecated. The new API will be released in the next release (v2.4.0), and I'm not sure if we should change the current API.

@HideakiImamura  What do you think of it?
CC @sile Do you have any comments as the original author of the `optuna.multi_objective` module?

cafeal:
> In short, multi-objective study will be integrated into the optuna.study.Study  

Sounds good! 

> The new API will be released in the next release (v2.4.0), and I'm not sure if we should change the current API.  

I can agree with you.
Feel free to close this PR if not in demand :)

HideakiImamura:
Thank you @cafeal for raising an important issue. As @toshihikoyanase said, the big refactoring of the multi-objective module is ongoing. Let me close this PR. Keep looking forward to the next release!

PR #2071, opened 2020-12-05
Fix parameter section of RandomSampler in docs

akihironitta:
I'd like to suggest a small fix in the docs. https://optuna.readthedocs.io/en/latest/reference/generated/optuna.samplers.RandomSampler.html#optuna.samplers.RandomSampler

## Description of the changes
*Before the change*

*After the change*


hvy (approved):
LGTM, good catch and thanks!

PR #2064, opened 2020-12-04
plot_optimization_history with target values other than objective value

not522:
## Motivation

Flexible plotting is sometimes useful, e.g., training duration or scores in other metrics.

## Description of the changes

Added `target` and `target_name` arguments to `plot_optimization_history`.

Issue #1916, opened 2020-10-12
[RFC] Preferred string formatting with Python 3.6+


hvy:
## Motivation

Part of https://github.com/optuna/optuna/issues/1067.

Since Python 3.5 has been dropped, "f-strings" ([PEP 498](https://www.python.org/dev/peps/pep-0498/)) can be used. We could agree on and document the preferred ways of formatting text in Optuna. 

Note that there's no such consensus for the time being.

If the preferred ways change, we should make sure that the code complies to it.

## Description

Discuss how text should be formatted and make sure that code complies to it.

crcrpar:
I would say `format` is preferrable as it's more explicit when a line gets long.

hvy:
I also find `format` to be preferable in certain cases where the string would become too long or when code/logic in f-strings are not particularly readable although this might be a matter of editor/IDE (a more exhaustive list of theoretical concerns are presented in the PEP). On the other hand, I don't see that this is enough to prohibit f-strings and I believe it's sensible to simply allow both.

crcrpar:
I don't say we should not use f-strings, nor should we migrate to them hurriedly. My attitude toward what to use and what to prohibit, including f-strings and whatever, is basically everything is OK unless linters don't warn.

hvy:
Agreed. Also checked with @c-bata, @keisuke-umezawa and other committers and it seemed okay allowing either. 

Let me close this issue as resolved. Thanks @crcrpar for the discussion.

PR #2075, opened 2020-12-08
plot_contour with target values other than objective value

not522:
## Motivation
This PR enables flexible plotting in `plot_contour`, same as #2064.
## Description of the changes
Added `target` and `target_name` arguments to `plot_contour`.

### DIFF
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -31,6 +31,7 @@ _logger = get_logger(__name__)
 def plot_contour(
     study: Study,
     params: Optional[List[str]] = None,
+    *,
     target: Optional[Callable[[FrozenTrial], float]] = None,
     target_name: str = "Objective Value",
 ) -> "go.Figure":
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -1,5 +1,6 @@
 import math
 from typing import Any
+from typing import Callable
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -27,7 +28,12 @@ if _imports.is_successful():
 _logger = get_logger(__name__)
 
 
-def plot_contour(study: Study, params: Optional[List[str]] = None) -> "go.Figure":
+def plot_contour(
+    study: Study,
+    params: Optional[List[str]] = None,
+    target: Optional[Callable[[FrozenTrial], float]] = None,
+    target_name: str = "Objective Value",
+) -> "go.Figure":
     """Plot the parameter relationship as contour plot in a study.
 
     Note that, If a parameter contains missing values, a trial with missing values is not plotted.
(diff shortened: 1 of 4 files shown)


optuna/visualization/_contour.py
```diff
@@ -55,17 +61,21 @@ def objective(trial):
...
-            A :class:`~optuna.study.Study` object whose trials are plotted for their objective
-            values.
+            A :class:`~optuna.study.Study` object whose trials are plotted for their target values.
         params:
             Parameter list to visualize. The default is all parameters.
+        target:
+            A function to specify the value to display. If it is :obj:`None`, the objective values
+            are plotted.
+        target_name:
+            Target's name to display on the axis label and the legend.
```
toshihikoyanase, line 71: A nit, but `target_name` is only shown in the [color bar](https://plotly.com/python/contour-plots/#color-bar-title) title in case of the contour plots.

image

```suggestion
            Target's name to display on the color bar.
```

optuna/visualization/matplotlib/_contour.py
```diff
@@ -44,10 +50,14 @@ def plot_contour(study: Study, params: Optional[List[str]] = None) -> "Axes":
...
-            A :class:`~optuna.study.Study` object whose trials are plotted for their objective
-            values.
+            A :class:`~optuna.study.Study` object whose trials are plotted for their target values.
         params:
             Parameter list to visualize. The default is all parameters.
+        target:
+            A function to specify the value to display. If it is :obj:`None`, the objective values
+            are plotted.
+        target_name:
+            Target's name to display on the axis label and the legend.
```
toshihikoyanase, line 60: ditto.
```suggestion
            Target's name to display on the color bar.
```

toshihikoyanase (commented):
Thank you for your PR. The change basically looks good to me, but I added some small comments.

hvy:
How about making the newly introduced parameter keyword only, as in https://github.com/optuna/optuna/pull/2082 ?

As for the CI failure, it should be fixed if you could merge the latest master branch.

not522:
Thanks, I updated PR following your comments.

optuna/visualization/_contour.py
```diff
@@ -242,13 +266,19 @@ def _generate_contour_subplot(
...
+            value = target(trial)
+
+        if isinstance(value, int):
+            value = float(value)
+        elif not isinstance(value, float):
             raise ValueError(
-                "Trial{} has COMPLETE state, but its value is non-numeric.".format(trial.number)
+                "Trial{} has COMPLETE state, but its target value is non-numeric.".format(
+                    trial.number
+                )
```
HideakiImamura, line 281: This is not directly related to this PR change, but what about using f-strings?
```suggestion
                f"Trial{trial.number} has COMPLETE state, but its target value is non-numeric."
```

HideakiImamura (commented):
Thanks for the PR! Basically, LGTM except for a very minor comment. PTAL.

optuna/visualization/_contour.py
not522: I also think it's better to use f-strings. Is it recommended in Optuna? Coding Style Conventions does not mention it.

HideakiImamura: According to the issue https://github.com/optuna/optuna/issues/1916#issuecomment-708158413, we'd like to use f-strings when the line doesn't get too long. This is not a strict rule, but now, we can use f-strings without a too long line.

toshihikoyanase (approved):
@HideakiImamura Please merge after your review.

HideakiImamura (approved):
LGTM!

PR #2076, opened 2020-12-08
Use lightgbm in visualization tutorial for faster documentation build

crcrpar:
This aims at fixing build in Read the Docs which recently failed due to
either timeout/memory consumption. At least, this tutorial build time is
much much much faster than PyTorch version as expected.

cc: @smly

### DIFF
--- a/tutorial/10_key_features/002_configurations.py
+++ b/tutorial/10_key_features/002_configurations.py
@@ -69,22 +69,25 @@ def objective(trial):
 
 ###################################################################################################
 # - Loops:
-import torch
-import torch.nn as nn
-
-
-def create_model(trial, in_size):
-    n_layers = trial.suggest_int("n_layers", 1, 3)
-
-    layers = []
-    for i in range(n_layers):
-        n_units = trial.suggest_int("n_units_l{}".format(i), 4, 128, log=True)
-        layers.append(nn.Linear(in_size, n_units))
-        layers.append(nn.ReLU())
-        in_size = n_units
-    layers.append(nn.Linear(in_size, 10))
-
-    return nn.Sequential(*layers)
+#
+# .. code-block:: python
+#
+#     import torch
+#     import torch.nn as nn
+#
+#
+#     def create_model(trial, in_size):
+#         n_layers = trial.suggest_int("n_layers", 1, 3)
+#
+#         layers = []
+#         for i in range(n_layers):
+#             n_units = trial.suggest_int("n_units_l{}".format(i), 4, 128, log=True)
+#             layers.append(nn.Linear(in_size, n_units))
+#             layers.append(nn.ReLU())
+#             in_size = n_units
+#         layers.append(nn.Linear(in_size, 10))
+#
+#         return nn.Sequential(*layers)
 
 
 ###################################################################################################
--- a/tutorial/10_key_features/005_visualization.py
+++ b/tutorial/10_key_features/005_visualization.py
@@ -36,10 +36,11 @@ def objective(trial):
     data, target = sklearn.datasets.load_breast_cancer(return_X_y=True)
     train_x, valid_x, train_y, valid_y = train_test_split(data, target, test_size=0.25)
     dtrain = lgb.Dataset(train_x, label=train_y)
+    dvalid = lgb.Dataset(valid_x, label=valid_y)
 
     param = {
         "objective": "binary",
-        "metric": "binary_logloss",
+        "metric": "auc",
         "verbosity": -1,
         "boosting_type": "gbdt",
         "lambda_l1": trial.suggest_float("lambda_l1", 1e-8, 10.0, log=True),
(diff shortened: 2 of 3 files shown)


crcrpar:
Ah, I forgot completely that I used PyTorch in configuration tutorial...

tutorial/10_key_features/002_configurations.py
```diff
@@ -69,22 +69,25 @@ def objective(trial):
...
+#
+#         layers = []
+#         for i in range(n_layers):
+#             n_units = trial.suggest_int("n_units_l{}".format(i), 4, 128, log=True)
+#             layers.append(nn.Linear(in_size, n_units))
+#             layers.append(nn.ReLU())
+#             in_size = n_units
+#         layers.append(nn.Linear(in_size, 10))
+#
+#         return nn.Sequential(*layers)
```
hvy, line 90: Let's use sklearn here as well https://scikit-learn.org/stable/modules/generated/sklearn.neural_network.MLPClassifier.html#sklearn-neural-network-mlpclassifier. It'll be less confusing to the reader.

tutorial/10_key_features/005_visualization.py
```diff
@@ -150,23 +88,23 @@ def objective(trial):
...
 plot_contour(study)
 
 ###################################################################################################
 # Select parameters to visualize.
-plot_contour(study, params=["n_units_l0", "n_units_l1"])
+plot_contour(study, params=["bagging_freq", "bagging_fraction"])
 
 ###################################################################################################
 # Visualize individual hyperparameters as slice plot. See :func:`~optuna.visualization.plot_slice` for the details.
 plot_slice(study)
```
hvy, line 103: How about specifying parameters here as well? The rendered plot looks a bit broken when it stretches outside the border https://64524-122299416-gh.circle-artifacts.com/0/docs/build/html/tutorial/10_key_features/005_visualization.html#sphx-glr-tutorial-10-key-features-005-visualization-py.

tutorial/10_key_features/005_visualization.py
```diff
@@ -28,109 +26,49 @@
...
+        "num_leaves": trial.suggest_int("num_leaves", 2, 256),
+        "feature_fraction": trial.suggest_float("feature_fraction", 0.4, 1.0),
+        "bagging_fraction": trial.suggest_float("bagging_fraction", 0.4, 1.0),
+        "bagging_freq": trial.suggest_int("bagging_freq", 1, 7),
+        "min_child_samples": trial.suggest_int("min_child_samples", 5, 100),
+    }
+
+    # Add a callback for pruning.
+    pruning_callback = optuna.integration.LightGBMPruningCallback(trial, "auc")
+    gbm = lgb.train(
```
hvy, line 57: Seems like it takes 10 minutes to run 4 trials. So it hasn't gotten particularly faster. It is because of the pruning? https://circleci.com/api/v1.1/project/github/optuna/optuna/64524/output/104/0?file=true&allocation-id=5fcf45d6dabe8b01b85d5353-0-build%2F69E17DC0 Can we reduce the number of boosting rounds or somehow shorten the time for each trial?

tutorial/10_key_features/005_visualization.py
```diff
@@ -28,109 +26,49 @@
...
+        "metric": "auc",
+        "verbosity": -1,
+        "boosting_type": "gbdt",
+        "lambda_l1": trial.suggest_float("lambda_l1", 1e-8, 10.0, log=True),
+        "lambda_l2": trial.suggest_float("lambda_l2", 1e-8, 10.0, log=True),
+        "num_leaves": trial.suggest_int("num_leaves", 2, 256),
+        "feature_fraction": trial.suggest_float("feature_fraction", 0.4, 1.0),
+        "bagging_fraction": trial.suggest_float("bagging_fraction", 0.4, 1.0),
+        "bagging_freq": trial.suggest_int("bagging_freq", 1, 7),
+        "min_child_samples": trial.suggest_int("min_child_samples", 5, 100),
```
hvy, line 52: Just a comment but if it turns out that we cannot run sufficiently many trials, we could reduce the number of hyperparameters. That should make the HPI results more convincing, for instance.

hvy (changes requested):
Thanks for your quick action. Left some early comments!

tutorial/10_key_features/002_configurations.py
hvy: Hm, how come? In my understanding, the code is rather simple and self documented. Something along the following.

```python
import sklearn.neural_network import MLPClassifier

def create_model(trial):
    n_layers = trial.suggest_int("n_layers", 3, 5)

    layers = []
    for i in range(n_layers - 2):
        n_units = trial.suggest_int("n_units_l{}".format(i), 4, 128, log=True)  # Irrelevant but I'm not sure if this should be log transformed. 
        layers.append(n_units)

    return MLPClassifier(layers)
```

crcrpar: I don't think sklearn MLPClassifier is so common enough to replace torch code.

keisuke-umezawa: In my understanding, pros/cons for PyTorch are as follows:

* pros
  * We can show a neural net example by using one of the most common library.
* cons
  * We can not detect whether the above code is broken or not by CI

crcrpar: Then the pros/cons sklearn are we are free from the ignorance of broken code by using not so popular feature of popular sklearn, right?

hvy: Well, the example is quite trivial and the main purpose is to demonstrate the usage of Optuna's API, so I think sklearn would do as well. Note that I'm also taking into account that sklearn is used in the previous example. On the other hand, this PR is a temporary fix so it's not a strong opinion. We have to deal with the runtime of the tutorial builds in a more sustainable matter (e.g. introducing new recipes would still quickly exceed the Read the Docs time limit). So, I'm fine either way.

crcrpar:
I won't do further work on this pull request given my situation and the release schedule. Sorry about this.

Issue #1347, opened 2020-06-09
Close connection after optimize


datasatanic:
## Expected behavior

I need to build many, fast-training models. For this i'm using multiprocessing.Pool:
```
with Pool(maxtasksperchild=10) as pool:
    res=pool.map(task,iter)
```
In each task i'm creating stydy with unique name  and run optimize. As storage is used Postgres 12.

After a few cycle, i get error from psycopg2: "too many client"
How I can manualy close connection for finished task and study. Or exist anotherway?
P.S. I know about the possibility to шncrease quantity of clients on Postgres, but It only test data. Real task is much more.

## Environment

- Optuna version: 1.5.0
- Python version: 3.7.7
- OS: Ubuntu 20.04
-  psycopg2-binary: 2.8.5
- Postgres version: 12

## Error messages, stack traces, or logs

```
# error messages, stack traces, or logs
The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/ilya/anaconda3/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/home/ilya/anaconda3/lib/python3.7/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "<ipython-input-11-68b9f5f269e0>", line 3, in task
    df:pd.DataFrame=pd.read_sql_query(select_query.format(pan=pan),connect_string)
  File "/home/ilya/anaconda3/lib/python3.7/site-packages/pandas/io/sql.py", line 332, in read_sql_query
    chunksize=chunksize,
  File "/home/ilya/anaconda3/lib/python3.7/site-packages/pandas/io/sql.py", line 1218, in read_query
    result = self.execute(*args)
  File "/home/ilya/anaconda3/lib/python3.7/site-packages/pandas/io/sql.py", line 1087, in execute
    return self.connectable.execute(*args, **kwargs)
  File "/home/ilya/anaconda3/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 2243, in execute
    connection = self._contextual_connect(close_with_result=True)
  File "/home/ilya/anaconda3/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 2311, in _contextual_connect
    self._wrap_pool_connect(self.pool.connect, None),
  File "/home/ilya/anaconda3/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 2349, in _wrap_pool_connect
    e, dialect, self
[...]

Issue #938, opened 2020-02-18
AttributeError is raised when using RDB storage.


c-bata:
## Reproducible examples (optional)

```python
import optuna

# Define a simple 2-dimensional objective function whose minimum value is -1 when (x, y) = (0, -1).
def objective(trial):
    x = trial.suggest_uniform('x', -100, 100)
    y = trial.suggest_categorical('y', [-1, 0, 1])
    return x**2 + y


if __name__ == '__main__':
    # Let us minimize the objective function above.
    print('Running 10 trials...')
    study = optuna.create_study(storage="sqlite:///db.sqlite3")
    study.optimize(objective, n_trials=10)
    print('Best value: {} (params: {})\n'.format(study.best_value, study.best_params))
```

## Expected behavior

tag: v1.1.0

```console
$ git co v1.1.0
$ python examples/quadratic_simple.py 
Running 10 trials...
[I 2020-02-18 14:05:57,875] A new study created with name: no-name-a2889f9d-79d7-40cc-aff4-f4021183b6aa
[I 2020-02-18 14:05:57,977] Finished trial#0 resulted in value: 78.25382960871015. Current best value is 78.25382960871015 with parameters: {'x': -8.78941577175128, 'y': 1}.
[I 2020-02-18 14:05:58,067] Finished trial#1 resulted in value: 1601.2080527969047. Current best value is 78.25382960871015 with parameters: {'x': -8.78941577175128, 'y': 1}.
[I 2020-02-18 14:05:58,159] Finished trial#2 resulted in value: 155.3355136572432. Current best value is 78.25382960871015 with parameters: {'x': -8.78941577175128, 'y': 1}.
[I 2020-02-18 14:05:58,252] Finished trial#3 resulted in value: 4928.445801453334. Current best value is 78.25382960871015 with parameters: {'x': -8.78941577175128, 'y': 1}.
[I 2020-02-18 14:05:58,349] Finished trial#4 resulted in value: 5494.046052713425. Current best value is 78.25382960871015 with parameters: {'x': -8.78941577175128, 'y': 1}.
[I 2020-02-18 14:05:58,451] Finished trial#5 resulted in value: 7.806334896764313. Current best value is 7.806334896764313 with parameters: {'x': -2.793981907021646, 'y': 0}.
[I 2020-02-18 14:05:58,540] Finished trial#6 resulted in value: 3581.7294062867. Current best value is 7.806334896764313 with parameters: {'x': -2.793981907021646, 'y': 0}.
[...]

PR #2079, opened 2020-12-08
Remove circular reference of study.

keisuke-umezawa:
## Motivation

To solve the issue 2 written in https://github.com/optuna/optuna/issues/1347#issuecomment-650852619.

Actually, in the previous code, study is never garbage collected and `_finalize` is never called, because weakref is used wrongly c.f. [note at the bottom](https://docs.python.org/3/library/weakref.html#weakref.finalize). So, in my understanding, `_finalize` is not needed. Also, [SQL alchemy documentation](https://docs.sqlalchemy.org/en/13/orm/contextual.html#thread-local-scope) says that we do not need to call scoped_session.remove().

> The scoped_session.remove() method, as always, removes the current Session associated with the thread, if any. However, one advantage of the threading.local() object is that if the application thread itself ends, the “storage” for that thread is also garbage collected. So it is in fact “safe” to use thread local scope with an application that spawns and tears down threads, without the need to call scoped_session.remove().

## Description of the changes

- [x] Remove weakref which make circular reference. c.f. [note at the bottom](https://docs.python.org/3/library/weakref.html#weakref.finalize)

## checks
```
import optuna

conn_str= "postgresql://localhost:5432/foo"

for i in range(101):  # 100 is the maximum allowed number of connections for postgresql by default.
    study = optuna.create_study(conn_str, study_name=f"test_{i}")
    study._storage._backend.delete_study(study._study_id)
```

The above code run correctly.

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -12,7 +12,6 @@ from typing import Optional
 from typing import Set
 from typing import Tuple
 import uuid
-import weakref
 
 import alembic.command
 import alembic.config
@@ -150,8 +149,6 @@ class RDBStorage(BaseStorage):
         self._version_manager = _VersionManager(self.url, self.engine, self.scoped_session)
         if not skip_compatibility_check:
             self._version_manager.check_table_schema_compatibility()
-
-        weakref.finalize(self, self._finalize)
 
     def __getstate__(self) -> Dict[Any, Any]:
 
@@ -1114,16 +1111,6 @@ class RDBStorage(BaseStorage):
         """
 
         self.scoped_session.remove()
-
-    def _finalize(self) -> None:
-
-        # This destructor calls remove_session to explicitly close the DB connection. We need this
-        # because DB connections created in SQLAlchemy are not automatically closed by reference
-        # counters, so it is not guaranteed that they are released by correct threads (for more
-        # information, please see the docstring of remove_session).
-
-        if hasattr(self, "scoped_session"):
-            self.remove_session()
 
     def upgrade(self) -> None:
         """Upgrade the storage schema."""


optuna/storages/_rdb/storage.py
```diff
@@ -1115,16 +1112,6 @@ def remove_session(self) -> None:
 
         self.scoped_session.remove()
 
-    def _finalize(self) -> None:
-
-        # This destructor calls remove_session to explicitly close the DB connection. We need this
-        # because DB connections created in SQLAlchemy are not automatically closed by reference
-        # counters, so it is not guaranteed that they are released by correct threads (for more
-        # information, please see the docstring of remove_session).
```
keisuke-umezawa, line 1123: However, this comment says that `_finalize` is needed 🤔 Even though, it is never called.

toshihikoyanase:
Thank you for your PR. As you and @hvy mentioned, `weakref.finalize`  should be fixed.

We have two issues about the finalization of `RDBStorage`, i.e., #938 and #1347. I roughly examined the reproducible code and this PR seems to resolve them. 

But, I'm not sure if we can just remove the finalization code. IMO, DB sessions should be closed somewhere before destroying the `RDBStorage` instances. Do you think the current implementation does not require it? 

Reproducible code for #938

```console
pip install jupyter
```

```python
import optuna

# Define a simple 2-dimensional objective function whose minimum value is -1 when (x, y) = (0, -1).
def objective(trial):
    x = trial.suggest_uniform('x', -100, 100)
    y = trial.suggest_categorical('y', [-1, 0, 1])
    return x**2 + y


if __name__ == '__main__':
    # Let us minimize the objective function above.
    print('Running 10 trials...')
    study = optuna.create_study(storage="sqlite:///db.sqlite3")
    print('Best value: {} (params: {})\n'.format(study.best_value, study.best_params))
```

Reproducible code for #1347

```console
docker run -it --rm --name postgres-test -e POSTGRES_PASSWORD=test -p 15432:5432 -d postgres
```

```python
from multiprocessing import Pool
import optuna
import random

conn_str = "postgres+psycopg2://postgres:test@localhost:15432/postgres"

def obj_func(trial:optuna.Trial,arg1:float):
    x=trial.suggest_float('x',0,100)
    return arg1*(x-3.3)**2

def task(arg1):
    study:optuna.Study=optuna.create_study(conn_str,study_name=f'test_{arg1}')
    study.optimize(lambda x:obj_func(x,arg1),n_trials=10)
    return study.best_trial

optuna.logging.set_verbosity(optuna.logging.ERROR)

# Create a dummy study to create DB tables.
optuna.create_study(conn_str)

task_ids = list(range(100))
random.shuffle(task_ids)

with Pool(maxtasksperchild=10) as pool:
    res=pool.map(task, task_ids)
```

keisuke-umezawa:
@toshihikoyanase Thank you for the comment. I understand the context of introducing weakref. I will check #938, and reply your comment later.

keisuke-umezawa:
I added the explanation for `remove` on the top comment.

[SQL alchemy documentation](https://docs.sqlalchemy.org/en/13/orm/contextual.html#thread-local-scope) says that we do not need to call scoped_session.remove().

> The scoped_session.remove() method, as always, removes the current Session associated with the thread, if any. However, one advantage of the threading.local() object is that if the application thread itself ends, the “storage” for that thread is also garbage collected. So it is in fact “safe” to use thread local scope with an application that spawns and tears down threads, without the need to call scoped_session.remove().

keisuke-umezawa:
sqlalchemy's changes related to scoped session:

no change in documentation recently: 
https://github.com/sqlalchemy/sqlalchemy/blame/50417cf758508e970f3fccc99af6a501ceac8eb2/doc/build/orm/contextual.rst

no significant change in code recently: 
https://github.com/sqlalchemy/sqlalchemy/blame/master/lib/sqlalchemy/orm/scoping.py
https://github.com/sqlalchemy/sqlalchemy/blame/d55971119aa48590416193b8b0e0f54aa0e97c82/lib/sqlalchemy/util/_collections.py#L1017

HideakiImamura (approved):
Thanks for the PR and the detailed explanation. The removal of the `_finalize` method seems reasonable to me. LGTM!

keisuke-umezawa:
@toshihikoyanase 
I surveyed materials related to this problem and summarized them on the above. So far, I can not reproduce the situation of b12b05d... Do you have any ideas to reproduce the situation?

toshihikoyanase:
@keisuke-umezawa Thank you for your investigation. 

I guess #1628 may resolve the root cause of the issue mentioned in the `RDBStorage.remove_session` comment. https://github.com/optuna/optuna/blob/1acac077b3cf9ea581319cd8b95d0f83f089ce23/optuna/storages/_rdb/storage.py#L1084-L1097
It ensures that each session calls `commit` or `rollback` at the end of the storage methods and keeps the transaction clean. So, threads can close the sessions without calling `rollback`.

If we don't call the `session.commit()` or `session.rollback()` intentionally, we can see `sqlite3.ProgrammingError` at the end of program execution as follows:

```python
import optuna

def error_callback(study, trial):
    session = study._storage._backend.scoped_session()
    if trial.number > 5:
        try:
            optuna.storages._rdb.models.StudyModel.find_or_raise_by_name("ho", session)
        finally:
            # session.commit()
            pass

def objective(trial):
    return trial.suggest_float("x", 0, 1)

study = optuna.create_study(storage="sqlite:///bar.db")
try:
    study.optimize(objective, n_trials=10, callbacks=[error_callback], n_jobs=4)
except:
    pass
```

```console
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 140345088288512 and this is thread id 140345489463104.
Exception closing connection <sqlite3.Connection object at 0x7fa4a3228030>
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/pool/base.py", line 697, in _finalize_fairy
    fairy._reset(pool)
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/pool/base.py", line 893, in _reset
    pool._dialect.do_rollback(self)
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 543, in do_rollback
    dbapi_connection.rollback()
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 140345088288512 and this is thread id 140345489463104.
```

In summary, I think we can accept this change.

toshihikoyanase (approved):
LGTM. Thank you for resolving the long-standing issue.

PR #2082, opened 2020-12-09
Make new visualization parameters keyword only

hvy:
## Motivation

What parameters to provide in the visualization functions is often open for discussion, e.g. what types of figure customization such as titles, legends, etc. Given that it might change in the future, it's safer to make new parameters keyword-only.

## Description of the changes

Makes not yet released new parameters introduced for visualization functions in v2.4.0 keyword-only

### DIFF
--- a/optuna/visualization/_optimization_history.py
+++ b/optuna/visualization/_optimization_history.py
@@ -17,6 +17,7 @@ _logger = get_logger(__name__)
 
 def plot_optimization_history(
     study: Study,
+    *,
     target: Optional[Callable[[FrozenTrial], float]] = None,
     target_name: str = "Objective Value",
 ) -> "go.Figure":
--- a/optuna/visualization/matplotlib/_optimization_history.py
+++ b/optuna/visualization/matplotlib/_optimization_history.py
@@ -20,6 +20,7 @@ _logger = get_logger(__name__)
 @experimental("2.2.0")
 def plot_optimization_history(
     study: Study,
+    *,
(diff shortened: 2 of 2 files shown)


ytsmiling:
You may have missed `multiobjective/visualiztion` (https://github.com/optuna/optuna/blob/190df03a53e4aa4a05091de328ccc28bf85d51fa/optuna/multi_objective/visualization/_pareto_front.py#L21-L26).

hvy:
I intentionally left it out since it felt somewhat unattractive to only cover `axis_order` but given that it's experimental, we could perhaps cover `names` and `include_domincated_trials` as well. What do you think?

ytsmiling:
I think it depends on whether we have plans to make the plot more customizable in the future. If the change was intentional, I'm fine to keep the `plot_pareto_front`'s interface as is.

hvy:
> I think it depends on whether we have plans to make the plot more customizable in the future. If the change was intentional, I'm fine to keep the plot_pareto_front's interface as is.

We don't have any concrete plans but it's not all that unlikely that that'll change. On a side note, it's experimental so we can change it at a later time for that matter.

toshihikoyanase:
@not522 If you have any comments as an original author of `target`-related keywords, please feel free to add review comments.

HideakiImamura (approved):
LGTM.

not522:
Thanks, LGTM!

HideakiImamura:
Let me merge this PR with 2 approvals.

PR #2078, opened 2020-12-08
Update pruning callback for xgboost 1.3.

trivialfis:
* Use new callback interface from xgboost 1.3.
* Remove the use of `silent` parameter.
* Update observation key from `error` to `logloss` in test.
* Unpin xgboost version on CI.

Close https://github.com/optuna/optuna/issues/2077 .

PR #2084, opened 2020-12-09
Pin xgboost<1.3

crcrpar:
## Motivation
Desirably I want to merge #2078 though it feels almost impossible to make that happen quick enough, keeping CI healthy as possible is preferred, in my humble opinion.


hvy (approved):
Makes sense as a temporary fix since the master is broken, LGTM!

PR #2087, opened 2020-12-10
Add note on the behavior of `suggest_float` with `step` argument.

crcrpar:
When `step` argument is given, `suggest_float` falls back to
`suggest_discrete_uniform` which includes both `low` and `high`.

Capture from https://64572-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/generated/optuna.trial.Trial.html#optuna.trial.Trial.suggest_float
^ Capture from https://64572-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/generated/optuna.trial.Trial.html#optuna.trial.Trial.suggest_float]

### DIFF
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -139,8 +139,7 @@ class Trial(BaseTrial):
 
                 .. note::
                     If ``step`` is specified, ``high`` is included as well as ``low`` because
-                    this method falls back to :func:`~optuna.trial.Trial.suggest_discrete_uniform`
-                    with ``step`` argument.
+                    this method falls back to :func:`~optuna.trial.Trial.suggest_discrete_uniform`.
 
             step:
                 A step of discretization.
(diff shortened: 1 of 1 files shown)


optuna/trial/_trial.py
```diff
@@ -136,6 +136,12 @@ def objective(trial):
             high:
                 Upper endpoint of the range of suggested values. ``high`` is excluded from the
                 range.
+
+                .. note::
+                    If ``step`` is specified, ``high`` is included as well as ``low`` because
+                    this method falls back to :func:`~optuna.trial.Trial.suggest_discrete_uniform`
+                    with ``step`` argument.
```
hvy, line 143: A nit but the following might be simpler and more concise (not having to be confused whether you didn't mean `q` or not). Sorry but I did not check the line length.
```suggestion
                    this method falls back to :func:`~optuna.trial.Trial.suggest_discrete_uniform`.
```

hvy (approved):
Good catch, I left a minor nit open for discussion but LGTM!

I must sat that the behavior is somewhat complicated...

toshihikoyanase:

himkt (approved):

PR #2089, opened 2020-12-11
plot_parallel_coordinate with target values other than objective value

not522:
## Motivation
This PR enables flexible plotting in `plot_parallel_coordinate`, same as #2064.

## Description of the changes
Added `target` and `target_name` arguments to `plot_parallel_coordinate`.


optuna/visualization/matplotlib/_parallel_coordinate.py
```diff
@@ -66,7 +84,14 @@ def _get_parallel_coordinate_plot(study: Study, params: Optional[List[str]] = No
...
 
-    obj_org = [t.value for t in trials if t.value is not None]
+    if target is None:
+
+        def _target(t: FrozenTrial) -> float:
+            return cast(float, t.value)
+
+        target = _target
+
+    obj_org = [target(t) for t in trials]
```
HideakiImamura, line 94: [Just a note] Before the change, it checked whether `t.value` was None or not, so the logic seems to be a little different after the change. However, the trials are filtered by `TrialState.COMPLETE`, `t.value` is guaranteed not to be None. There is no behavior change.
```

**Review by HideakiImamura — APPROVED** [review]
Thanks for the PR! LGTM!

File: optuna/visualization/_parallel_coordinate.py [hunk 2]
```diff
@@ -114,7 +139,7 @@ def _get_parallel_coordinate_plot(study: Study, params: Optional[List[str]] = No
             line={
                 "color": dims[0]["values"],
                 "colorscale": "blues",
-                "colorbar": {"title": "Objective Value"},
+                "colorbar": {"title": target_name},
                 "showscale": True,
                 "reversescale": study.direction == StudyDirection.MINIMIZE,
```
hvy, line 144: I'm wondering if it's not user friendlier to fix the color scale to either (any is fine) when `target` is specified. The direction is only associated with the value. As for the matplotlib variant, https://github.com/optuna/optuna/pull/2090 is related.

hvy (commented):
Thanks for the coming PRs. This is not a strong opinion but I left a minor comment if you could take a look.

optuna/visualization/_parallel_coordinate.py
not522: Good catch! I updated PR to display smaller values in darker color when `target` is specified.

hvy: Verified locally, thanks for the quick fix!

hvy (approved):
LGTM!

PR #2090, opened 2020-12-11
Fix flipped colormap in matplotlib backend plot_parallel_coordinate

not522:
## Motivation
The colormap in matplotlib backend `plot_parallel_coordinate` is flipped.
## Description of the changes
Use `Blues_r` if the study direction is `minimize`. This PR does not add unit tests because I don't know how to check it...
- plotly
plotly
- matplotlib (master)
master
- matplotlib (PR)
PR

### DIFF
--- a/optuna/visualization/matplotlib/_parallel_coordinate.py
+++ b/optuna/visualization/matplotlib/_parallel_coordinate.py
@@ -6,6 +6,7 @@ from typing import Optional
 import numpy as np
 
 from optuna._experimental import experimental
+from optuna._study_direction import StudyDirection
 from optuna.logging import get_logger
 from optuna.study import Study
 from optuna.trial import TrialState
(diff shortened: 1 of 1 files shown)


HideakiImamura (approved):
Looks solid to me. Checking a `cmap` of each object of the `Axes.collections` is difficult since the type of `matplotlib.Axes.collections[i].cmap` is `matplotlib.colors.ListedColormap`, but that of `matplotlib.pyplot.get_cmap("Blues_r")` is `matplotlib.colors.LinearSegmentedColormap`. It's OK to merge this PR without adding any tests.

hvy (approved):
Good catch and LGTM!

By the way, you can probably test this by checking the `ax.collections[0].cmap.name` string, but I think it's fine as is and I also checked the behavior locally. Let me merge this change.

PR #2088, opened 2020-12-11
Tune build time of #2076

toshihikoyanase:
## Motivation
Currently, the build of documents executes optimization code. In #2076 , the execution time of LightGBM took about ten minutes for only four trials ([log](https://circleci.com/api/v1.1/project/github/optuna/optuna/64524/output/104/0?file=true&allocation-id=5fcf45d6dabe8b01b85d5353-0-build%2F69E17DC0)). This PR investigates the cause and tune the build time.

## Description of the changes

Add an environmental variable `OMP_NUM_THREADS=1` like the execution of the unit tests.

PR #1722, opened 2020-08-26
Update tutorial

crcrpar:
New recipes are in https://github.com/crcrpar/optuna/tree/new-recipes.

This pull request restructures the current tutorial by separating contents into two categories named "Key Features" and "Recipes".
The former is aligned with the Key Features listed in README.md and the latter is the existing tutorial contents that explain the features not included in Key Features.

Ref: #1721

PR #2095, opened 2020-12-11
[CircleCI] Cache tutorial artifacts for faster documentation build

crcrpar:
This aims at making it faster to build documentation by cacheing tutotorial following GitHub Actions job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,6 +35,15 @@ jobs:
             pip install --progress-bar off .[document] -f https://download.pytorch.org/whl/torch_stable.html
 
       - run:
+          name: Calculate hash
+          command: |
+            find tutorial -type f -print0 | xargs -0 md5sum > /tmp/tutorial_files.md5
+
+      - restore_cache:
+          keys:
+            - tutorial-{{ checksum "/tmp/tutorial_files.md5" }}
+
+      - run:
           name: build
           command: |
             . venv/bin/activate
@@ -43,6 +52,12 @@ jobs:
 
       - store_artifacts:
           path: ./docs/build/html
+
+      - save_cache:
+          key: tutorial-{{ checksum "/tmp/tutorial_files.md5" }}
+          paths:
+            - tutorial/MNIST
+            - docs/source/tutorial
 
   # Unit tests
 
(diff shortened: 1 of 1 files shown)


hvy:
@himkt would you be interested in taking a look at this PR?

.circleci/config.yml
```diff
@@ -44,6 +53,12 @@ jobs:
       - store_artifacts:
           path: ./docs/build/html
 
+      - save_cache:
+          key: tutorial-{{ checksum "/tmp/tutorial_files.md5" }}
+          paths:
+            - tutorial/MNIST
```
himkt, line 59: Sorry but let me ask a question because I'm not familiar with the original PR and CircleCI: when `tutorial/MNIST` will be generated? I checked it appear in the original [PR](https://github.com/optuna/optuna/pull/1722/files#diff-e4697a2845e32e857d02b8540419a37459fa603b4db72167817367038b4f5a64R30) but I couldn't what step creates it...

himkt (commented):
I looked at the official [example](https://circleci.com/docs/2.0/sample-config/#sample-configuration-with-fan-infan-out-workflow) and the [documentation](https://circleci.com/docs/2.0/caching/) about caching and it looks solid.

I left a single question. Please take a look.

.circleci/config.yml
crcrpar: The visualization tutorial with PyTorch of #1722 downloaded MNIST. Then, it turned out to be not feasible on ReadTheDocs due to the longer build time and we migrated to lightgbm from PyTorch in #2088, which might be tentative though. So now, we don't need MNIST at this moment.

himkt, line 59:
```suggestion
```

Thank you for the explanation. So, we can remove it for now, right?

crcrpar: Absolutely.

himkt (approved):
LGTM

hvy:
Let me just clarify the motivation. Are we looking to extend the tutorial build time? I'm probably missing something but was looking at https://app.circleci.com/pipelines/github/optuna/optuna/4087/workflows/019bb499-ae45-42e4-953c-93a454b26746/jobs/65182, a build without this change, which already runs under 1 minute.

crcrpar:
okay let's close this then. I forgot that we don't use PyTorch in tutorial.

Issue #1834, opened 2020-09-15
`visualization.plot_contour` with `suggest_categorical` that contains `None`


nzw0301:
## Expected behavior

`visualization.plot_contour` shows an error if `params` contains a hyper-parameter that could be `None`.  But `choice` option of `suggest_categorical` can contain `None` as an element, so, I think users expect the plot functions perform in such a case.

I'm not sure that the other visualisation functions have a similar issue.

## Environment

- Optuna version: https://github.com/optuna/optuna/tree/a4237f087ec01bcd2c181d8a1a7d553d2292b7cf
- Python version: 3.8.3
- OS: MacOS 10.15.6

## Error messages, stack traces, or logs

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-c1f5a8612550> in <module>
     24 print(study.best_params, study.best_value)
     25 
---> 26 optuna.visualization.plot_contour(study, params=['l1_ratio', 'C'])

~/PycharmProjects/optuna/optuna/visualization/_contour.py in plot_contour(study, params)
     64 
     65     _imports.check()
---> 66     return _get_contour_plot(study, params)
     67 
     68 

~/PycharmProjects/optuna/optuna/visualization/_contour.py in _get_contour_plot(study, params)
     93     for p_name in sorted_params:
     94         values = [t.params[p_name] for t in trials if p_name in t.params]
---> 95         max_value = max(values)
     96         min_value = min(values)
     97         if _is_log_scale(trials, p_name):

TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
```

## Steps to reproduce

1. Run the following example

## Reproducible examples (optional)

```python
import numpy as np
from sklearn.datasets import load_iris
from sklearn.model_selection import train_test_split
from sklearn.linear_model import LogisticRegression

import optuna


rnd = np.random.RandomState(7)

X, y = load_iris(return_X_y=True)
X_train, X_valid, y_train, y_valid = train_test_split(X, y, random_state=rnd)

def objective(trial):
    l1_ratio = trial.suggest_categorical('l1_ratio', [None, 0.1])
    C = trial.suggest_float('C', 0., 7.)

    clf = LogisticRegression(l1_ratio=l1_ratio, C=C, random_state=0, solver="saga")
    clf.fit(X_train, y_train)
    return clf.score(X_valid, y_valid)

study = optuna.create_study(direction='maximize')
study.optimize(objective, n_trials=4)
[...]

PR #2096, opened 2020-12-11
Replace builtin `isnumerical` to capture float values in `plot_contour`.

nzw0301:
## Motivation

If the categorical candidates contain only `float` and `None`,  the `plot_contour` generate [a blank plot](https://user-images.githubusercontent.com/7121753/97346796-403d7b00-18cf-11eb-952e-06cd5b89e01e.png) with older `plotly`, which may be older than `4.12.0`. Note that the latest plotly is fine.

Please see #1834 for more details.

## Description of the changes

Just apply https://github.com/optuna/optuna/issues/1834#issuecomment-719941511.

### DIFF
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -129,7 +129,7 @@ def _get_contour_plot(
 
         elif _is_categorical(trials, p_name):
             # For numeric values, plotly does not automatically plot as "category" type.
-            update_category_axes[p_name] = any([str(v).isnumeric() for v in set(values)])
+            update_category_axes[p_name] = any([_is_numeric(str(v)) for v in set(values)])
 
             # Plotly>=4.12.0 draws contours using the indices of categorical variables instead of
             # raw values and the range should be updated based on the cardinality of categorical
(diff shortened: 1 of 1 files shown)


crcrpar (approved):

hvy:
@toshihikoyanase could you take a look at this change? I think you have followed the discussion so far the best.

optuna/visualization/_contour.py
```diff
@@ -129,7 +129,7 @@ def _get_contour_plot(
 
         elif _is_categorical(trials, p_name):
             # For numeric values, plotly does not automatically plot as "category" type.
-            update_category_axes[p_name] = any([str(v).isnumeric() for v in set(values)])
+            update_category_axes[p_name] = any([_is_numeric(str(v)) for v in set(values)])
```
toshihikoyanase, line 132: A nit-picky, but `[]` is not necessary for `any()`.
This is not introduced by this PR, but could you fix it to keep the code clean?

```suggestion
            update_category_axes[p_name] = any(_is_numeric(str(v)) for v in set(values))
```

toshihikoyanase (commented):
Thank you for your PR. The change looks good to me, but I think we can simplify the code a bit.

optuna/visualization/_contour.py
nzw0301: Make code cleaner

nzw0301:
@toshihikoyanase Thank you for your suggestion. I have updated the line mentioned by you.

toshihikoyanase (approved):

Issue #2091, opened 2020-12-11
Adding code & figure examples of `visualization.matplotlib` functions


HideakiImamura:
Providing example code and figure for every function in `visualization.matplotlib` is useful for users to know the difference between the figure output by the corresponding function in the plotly backend, and for developers to detect visual bugs that cannot be detected by unit tests. Currently, some functions in `visualization.matplotlib` do not have example code and figure, so it would be better to provide them.

## How to add a code example and figure
- Write a code example with the matplotlib plot directive.
- See the docstring of the [edf function](https://github.com/optuna/optuna/blob/master/optuna/visualization/matplotlib/_edf.py) for an example

## Current status
- [x] `optuna.visualization.matplotlib.plot_contour` #2100 
- [x] `optuna.visualization.matplotlib.plot_edf`
- [x] `optuna.visualization.matplotlib.plot_intermediate_values`
- [x] `optuna.visualization.matplotlib.plot_optimization_history`
- [x] `optuna.visualization.matplotlib.plot_parallel_coordinate` #2097 
- [x] `optuna.visualization.matplotlib.plot_param_importances` #2098 
- [ ] `optuna.visualization.matplotlib.plot_slice` #2099

keisuke-umezawa:
@nzw0301 Congrats for the last one mile!

PR #2097, opened 2020-12-11
Add `matplotlib.plot_parallel_coordinate` example

nzw0301:
## Motivation

See #2091

## Description of the changes

Add the example that is the same code as plotly's one.

### DIFF
--- a/optuna/visualization/matplotlib/_parallel_coordinate.py
+++ b/optuna/visualization/matplotlib/_parallel_coordinate.py
@@ -37,6 +37,26 @@ def plot_parallel_coordinate(
     .. seealso::
         Please refer to :func:`optuna.visualization.plot_parallel_coordinate` for an example.
 
+    Example:
+
+        The following code snippet shows how to plot the high-dimentional parameter relationships.
+
+        .. plot::
+
+            import optuna
+
+            def objective(trial):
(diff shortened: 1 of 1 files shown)


crcrpar (approved):
Looks good, thanks.

image

nzw0301:
@crcrpar  Thank you for your review! I'm a bit concern about the left-most xlabel's position. It is cut off, so sphinx or optuna's plot function may need to fix it.


Reference: [Why is my xlabel cut off in my matplotlib plot?](https://stackoverflow.com/questions/6774086/why-is-my-xlabel-cut-off-in-my-matplotlib-plot)

crcrpar:
Yeah, the truncation of x-labels is not cool.

But, I would not fix that by editing the current implementation like changing the configuration or calling `tight_layout`.
This is because those changes are unrelated to the logic of optuna.visualization functions.
If you kindly fix the truncation in the plot of the example code, I'd like you to do that in the example code with comments like
"getting ax and calling `tight_layout` to avoid the cut-off of x-labels in the figure".
Anyway, I don't think the fix necessary in this pull request.

hvy (approved):
Thank you for the PR as always. I agree with with @crcrpar that the label fix is outside the scope of this PR, and whether to use `tight_layout` might be an open discussion. LGTM!

PR #2098, opened 2020-12-11
Add `matplotlib.plot_param_importances` example

nzw0301:
## Motivation

See #2091

## Description of the changes

Add the example that is the same code as plotly's one.


crcrpar (approved):
The change itself solid, thanks.

However, I think the location of the legend has space for improvement, which should be accessed in a separate pull request.
image

hvy (approved):
Thanks for your steady contribution. LGTM!

The legend could be improved for sure. Maybe also aligned with the plotly version. At least it's transparent so it's not hiding crucial information 🙂

Issue #1739, opened 2020-08-28
ubuntu pip install error


Jie-Yuan:
![image](https://user-images.githubusercontent.com/20265321/91530606-fa4f4f80-e93d-11ea-9653-946a3c7e653e.png)

hvy:
Given my very shallow understanding, does it happen to be a duplicate of https://github.com/iterative/dvc/issues/1995 (https://github.com/iterative/dvc/issues/1995#issuecomment-540259645)?

Jie-Yuan:
Looking in indexes: https://pypi.tuna.tsinghua.edu.cn/simple
Collecting optuna
  Downloading https://pypi.tuna.tsinghua.edu.cn/packages/06/b0/9a6313c78bca92abfacc08a2ad8b27bfe845256f615786ee2b6452ae1978/optuna-2.0.0.tar.gz (226 kB)
     |████████████████████████████████| 226 kB 21.2 MB/s eta 0:00:01
Collecting alembic
  Downloading https://pypi.tuna.tsinghua.edu.cn/packages/60/1e/cabc75a189de0fbb2841d0975243e59bde8b7822bacbb95008ac6fe9ad47/alembic-1.4.2.tar.gz (1.1 MB)
     |████████████████████████████████| 1.1 MB 51.8 MB/s eta 0:00:01
  Installing build dependencies ... error
  ERROR: Command errored out with exit status 1:
   command: /usr/local/bin/python /usr/local/lib/python3.6/dist-packages/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-5ipz5ku3/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.tuna.tsinghua.edu.cn/simple -- 'setuptools>=40.8.0' wheel
       cwd: None
  Complete output (31 lines):
  Traceback (most recent call last):
    File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
      "__main__", mod_spec)
    File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
      exec(code, run_globals)
    File "/usr/local/lib/python3.6/dist-packages/pip/__main__.py", line 26, in <module>
      sys.exit(_main())
    File "/usr/local/lib/python3.6/dist-packages/pip/_internal/cli/main.py", line 73, in main
      command = create_command(cmd_name, isolated=("--isolated" in cmd_args))
[...]

PR #2105, opened 2020-12-14
Release wheel package via PyPI

toshihikoyanase:
## Motivation
This PR adds [Python wheel](https://pythonwheels.com/)-format package of Optuna to the [PyPI release](https://pypi.org/project/optuna/).

Currently, Optuna only provides source packages in [pypi](https://pypi.org/project/optuna/). When we install optuna from the source packages, the build of the wheel package is executed as follows:

```console
$ docker run -it -v $(pwd):/prj --rm -w /prj python:3.8 bash
$ pip install optuna
...
Building wheels for collected packages: optuna, PrettyTable, pyperclip, PyYAML
  Building wheel for optuna (PEP 517) ... done
  Created wheel for optuna: filename=optuna-2.3.0-py3-none-any.whl size=359705 sha256=cd8d7872a7d23de7923affeebcbe80ecd05a9a303c2051a1d0bb8f7fe876812d
```

The build process may fail depending on the user environments. For example, if users have installed `typing`, the installation will fail as reported in https://github.com/optuna/optuna/issues/1739. 

```
$ pip install typing
root@5c7aa57fc51b:/prj# pip install optuna
Collecting optuna
  Downloading optuna-2.3.0.tar.gz (258 kB)
     |████████████████████████████████| 258 kB 1.9 MB/s
  Installing build dependencies ... error
  ERROR: Command errored out with exit status 1:
   command: /usr/local/bin/python /usr/local/lib/python3.8/site-packages/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-auvtcqrn/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple -- 'setuptools>=40.8.0' wheel
       cwd: None
  Complete output (42 lines):
...
      self._abc_registry = extra._abc_registry
  AttributeError: type object 'Callable' has no attribute '_abc_registry'
```

We can avoid such errors by providing the wheel package in addition to the source package.

```console
$ pip install dist/optuna-2.3.0-py3-none-any.whl
Processing ./dist/optuna-2.3.0-py3-none-any.whl
...
Successfully installed Mako-1.1.3 MarkupSafe-1.1.1 PrettyTable-0.7.2 PyYAML-5.3.1 alembic-1.4.3 attrs-20.3.0 cliff-3.5.0 cmaes-0.7.0 cmd2-1.4.0 colorama-0.4.4 colorlog-4.6.2 joblib-0.17.0 numpy-1.19.4 optuna-2.3.0 packaging-20.8 pbr-5.5.1 pyparsing-2.4.7 pyperclip-1.8.1 python-dateutil-2.8.1 python-editor-1.0.4 scipy-1.5.4 six-1.15.0 sqlalchemy-1.3.20 stevedore-3.3.0 tqdm-4.54.1 wcwidth-0.2.5
```

## Description of the changes

This PR installs the `wheel` package and runs `python setup.py bdist_wheel`.
We may need to fix `setup.cfg` if we want to add LICENSE to the package (c.f., https://pythonwheels.com/).

Update 1: The license file has been included without adding `license_file` option to `setup.cfg`.

```console
...
adding license file "LICENSE" (matched pattern "LICEN[CS]E*")
```

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -19,11 +19,12 @@ jobs:
 
     - name: Install twine
       run: >-
-        python -m pip install -U twine
+        python -m pip install -U twine wheel
 
     - name: Build a tar ball
       run: >-
         python setup.py sdist
+        python setup.py bdist_wheel
 
     - name: Verify the distributions
       run: twine check dist/*
--- a/setup.py
+++ b/setup.py
@@ -107,22 +107,16 @@ def get_extras_require() -> Dict[str, List[str]]:
             "skorch",
             "stable-baselines3>=0.7.0",
             "catalyst",
-        ]
-        + (
-            ["torch==1.7.0", "torchvision==0.8.1", "torchaudio==0.7.0"]
-            if sys.platform == "darwin"
-            else ["torch==1.7.0+cpu", "torchvision==0.8.1+cpu", "torchaudio==0.7.0"]
-        )
-        + (
-            [
-                "allennlp==1.2.0",
-                "fastai",
-                "dask[dataframe]",
-                "dask-ml",
-            ]
-            if sys.version_info[:2] < (3, 8)
-            else ["fastai"]
-        ),
+            "torch==1.7.0 ; sys_platform=='darwin'",
+            "torch==1.7.0+cpu ; sys_platform!='darwin'",
+            "torchvision==0.8.1 ; sys_platform=='darwin'",
+            "torchvision==0.8.1+cpu ; sys_platform!='darwin'",
+            "torchaudio==0.7.0",
+            "allennlp==1.2.0 ; python_version<'3.8'",
+            "dask[dataframe] ; python_version<'3.8'",
+            "dask-ml ; python_version<'3.8'",
+            "fastai",
+        ],
         "experimental": ["redis"],
         "testing": [
             # TODO(toshihikoyanase): Remove the version constraint after resolving the issue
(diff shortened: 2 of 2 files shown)


HideakiImamura (approved):
Thanks for the PR. I confirmed that the error is reproduced and resolved by this fix.

hvy (commented):
Thanks, the change basically LGTM.

I wonder what we should do with environment dependent `extras_require` though. We have have to build two wheels to account for both paths when checking for macOS in https://github.com/optuna/optuna/blob/master/setup.py#L191. However, this is highly specific to our CI environment (albeit, it was pointed out as an issue. That's why the condition is there) skipping downloading PyTorch with CUDA to save time. I personally find it okay to overlook it here, but I assume it's worth mentioning (thanks @kmaehashi for pointing that out).

toshihikoyanase:
@hvy Thank you for pointing it out. I overlooked the platform-dependent code in `setup.py`. I think we can avoid the dependencies using the environment markers of [PEP-0508](https://www.python.org/dev/peps/pep-0508/#environment-markers) according to the [so article](https://stackoverflow.com/questions/59824042/create-python-wheel-based-on-platform). I'll apply to the `setup.py`.

toshihikoyanase:
@hvy I applied the environment markers to remove the platform-dependent code.
When I create the wheel package using the new `setup.py`, the platform checks can be found in `optuna-2.3.0.dist-info/METADATA` as follows:

```
Requires-Dist: torch (==1.7.0+cpu) ; (sys_platform != "darwin") and extra == 'example'
Requires-Dist: torchvision (==0.8.1+cpu) ; (sys_platform != "darwin") and extra == 'example'
Requires-Dist: torchaudio (==0.7.0) ; (sys_platform != "darwin") and extra == 'example'
Requires-Dist: torch (==1.7.0) ; (sys_platform == "darwin") and extra == 'example'
Requires-Dist: torchvision (==0.8.1) ; (sys_platform == "darwin") and extra == 'example'
Requires-Dist: torchaudio (==0.7.0) ; (sys_platform == "darwin") and extra == 'example'
```

It works as expected in the following environment:

- Linux: Python 3.8.6, pip 20.3
- Mac: Python 3.7.7, pip 20.3.1

PTAL.

kmaehashi:
Oh, I didn't know environment markers other than `python_version`. Switching based on `sys_platform ` sound nice 👍🏼

hvy (approved):
Cleans up setup.py a lot, looks good. LGTM!

Verified it locally using a container and it's working as expected.

Issue #2107, opened 2020-12-16
FastAI v2 example fails


crcrpar:
## Expected behavior

Runs in CI.

## Environment

Daily examples CI job.

## Error messages, stack traces, or logs

```
[W 2020-12-15 17:26:46,193] Trial 0 failed because of the following error: TypeError("no implementation found for 'torch.tensor.eq' on types that implement __torch_function__: [<class 'fastai.torch_core.TensorImage'>, <class 'fastai.torch_core.TensorCategory'>]")
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/optuna/_optimize.py", line 191, in _run_trial
    value_or_values = func(trial)
  File "examples/fastaiv2_simple.py", line 78, in objective
    learn.fit(EPOCHS)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/fastai/learner.py", line 206, in fit
    self._with_events(self._do_fit, 'fit', CancelFitException, self._end_cleanup)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/fastai/learner.py", line 155, in _with_events
    try:       self(f'before_{event_type}')       ;f()
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/fastai/learner.py", line 197, in _do_fit
    self._with_events(self._do_epoch, 'epoch', CancelEpochException)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/fastai/learner.py", line 155, in _with_events
    try:       self(f'before_{event_type}')       ;f()
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/fastai/learner.py", line 192, in _do_epoch
    self._do_epoch_validate()
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/fastai/learner.py", line 188, in _do_epoch_validate
    with torch.no_grad(): self._with_events(self.all_batches, 'validate', CancelValidException)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/fastai/learner.py", line 155, in _with_events
    try:       self(f'before_{event_type}')       ;f()
[...]

PR #2108, opened 2020-12-16
Skip FastAI v2 example in examples job.

crcrpar:
See: https://github.com/optuna/optuna/issues/2107.

Skip fastaiv2 because the good version of fastai is >2,<2.1 which requires pytorch1.6. That's incompatible with the other examples, unfortunately.


hvy (approved):
LGTM!

HideakiImamura (approved):
LGTM.

PR #2109, opened 2020-12-16
importance with target values other than objective value

not522:
## Motivation
This PR enables us to evaluate hyperparameter importance for customized target values, e.g., duration or scores in other metrics.
related to #2064.

## Description of the changes
Added `target` argument to `get_param_importances` and `evaluate` of evaluators.

### DIFF
--- a/tests/importance_tests/test_fanova.py
+++ b/tests/importance_tests/test_fanova.py
@@ -55,3 +55,20 @@ def test_fanova_importance_evaluator_seed() -> None:
     evaluator = FanovaImportanceEvaluator(seed=3)
     param_importance_different_seed = evaluator.evaluate(study)
     assert param_importance != param_importance_different_seed
+
+
+def test_fanova_importance_evaluator_with_target() -> None:
+    # Assumes that `seed` can be fixed to reproduce identical results.
+
+    study = create_study(sampler=RandomSampler(seed=0))
+    study.optimize(objective, n_trials=100)
+
+    evaluator = FanovaImportanceEvaluator(seed=0)
+    param_importance = evaluator.evaluate(study)
+    param_importance_with_target = evaluator.evaluate(
+        study,
+        target=lambda t: t.params["x1"] + t.params["x2"],
+    )
+
+    assert param_importance != param_importance_with_target
+    assert param_importance_with_target["x3"] == min(param_importance_with_target.values())
--- a/tests/importance_tests/test_fanova.py
+++ b/tests/importance_tests/test_fanova.py
@@ -61,7 +61,7 @@ def test_fanova_importance_evaluator_with_target() -> None:
     # Assumes that `seed` can be fixed to reproduce identical results.
 
     study = create_study(sampler=RandomSampler(seed=0))
-    study.optimize(objective, n_trials=100)
+    study.optimize(objective, n_trials=3)
 
     evaluator = FanovaImportanceEvaluator(seed=0)
     param_importance = evaluator.evaluate(study)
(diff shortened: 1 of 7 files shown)


tests/importance_tests/test_fanova.py
```diff
@@ -55,3 +55,20 @@ def test_fanova_importance_evaluator_seed() -> None:
     evaluator = FanovaImportanceEvaluator(seed=3)
     param_importance_different_seed = evaluator.evaluate(study)
     assert param_importance != param_importance_different_seed
+
+
+def test_fanova_importance_evaluator_with_target() -> None:
+    # Assumes that `seed` can be fixed to reproduce identical results.
+
+    study = create_study(sampler=RandomSampler(seed=0))
+    study.optimize(objective, n_trials=100)
```
HideakiImamura, line 64: `n_trials = 100` is too big for the unit test. How about setting `n_trials = 3` which is the same as above to reduce the execution time?

tests/importance_tests/test_init.py
```diff
@@ -103,6 +103,46 @@ def objective(trial: Trial) -> float:
...
+        if trial.number % 2 == 0:
+            x4 = trial.suggest_uniform("x4", 0.1, 3)
+
+        value = x1 ** 4 + x2 + x3
+        if trial.number % 2 == 0:
+            value += x4
+        return value
+
+    study = create_study(storage_init_func())
+    study.optimize(objective, n_trials=100)
```
HideakiImamura, line 125: Ditto.

tests/importance_tests/test_mean_decrease_impurity.py
```diff
@@ -55,3 +55,20 @@ def test_mean_decrease_impurity_importance_evaluator_seed() -> None:
     evaluator = MeanDecreaseImpurityImportanceEvaluator(seed=3)
     param_importance_different_seed = evaluator.evaluate(study)
     assert param_importance != param_importance_different_seed
+
+
+def test_mean_decrease_impurity_importance_evaluator_with_target() -> None:
+    # Assumes that `seed` can be fixed to reproduce identical results.
+
+    study = create_study(sampler=RandomSampler(seed=0))
+    study.optimize(objective, n_trials=100)
```
HideakiImamura, line 64: Ditto.

HideakiImamura (commented):
Thanks for the PR! I have some early comments. Could you take a look?

tests/importance_tests/test_init.py
```diff
@@ -103,6 +103,46 @@ def objective(trial: Trial) -> float:
...
+    assert isinstance(param_importance, OrderedDict)
+    assert len(param_importance) == 3
+    assert all(param_name in param_importance for param_name in ["x1", "x2", "x3"])
+    prev_importance = float("inf")
+    for param_name, importance in param_importance.items():
+        assert isinstance(param_name, str)
+        assert isinstance(importance, float)
+        assert importance <= prev_importance
+        prev_importance = importance
+    assert param_importance["x3"] == min(param_importance.values())
```
hvy, line 142: This is indeed expected but I'm not sure it's suitable as a unit test without fixing the seed. It's also more likely to fail if we reduce the number of trials as suggested by @HideakiImamura. Since the current interface of the evaluator makes it a bit difficult to do this, I think we can omit this line from this general HPI test.

hvy (changes requested):
Thanks for the PR, logic LGTM. Left a minor comment on one of the tests.

tests/importance_tests/test_init.py
not522: Yes, this assertion actually fails when we reduce `n_trials`, so I chose large `n_trials`. I think it is not suitable for a unit test. I agree to omit it.

not522:
I updated PR following your comments.

HideakiImamura (approved):
Thanks for the update! LGTM!

hvy (approved):
LGTM!

Please note that this will break existing user implemented evaluators. It should be mentioned in the release note.

PR #2119, opened 2020-12-18
plot_param_importances with target values other than objective value

not522:
## Motivation
This PR enables flexible plotting in `plot_param_importances`.
Related to #2064, #2109.
## Description of the changes
Added `target` argument to `plot_param_importances`.

### DIFF
--- a/optuna/visualization/_param_importances.py
+++ b/optuna/visualization/_param_importances.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from typing import Callable
 from typing import List
 from typing import Optional
 
@@ -13,6 +14,7 @@ from optuna.distributions import UniformDistribution
 from optuna.importance._base import BaseImportanceEvaluator
 from optuna.logging import get_logger
 from optuna.study import Study
+from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
 from optuna.visualization._plotly_imports import _imports
 
@@ -40,6 +42,8 @@ def plot_param_importances(
     study: Study,
     evaluator: Optional[BaseImportanceEvaluator] = None,
     params: Optional[List[str]] = None,
+    *,
+    target: Optional[Callable[[FrozenTrial], float]] = None,
 ) -> "go.Figure":
     """Plot hyperparameter importances.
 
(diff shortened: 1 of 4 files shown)


HideakiImamura (commented):
Thanks for the PR. 

I am concerned that `target` can be specified, but `target_name` cannot. Even if you switch the `target` and evaluate the importance and plot it, we cannot know from the figure what kind of target it is. 

For example, one solution would be to put the title `Importances for Objective Values` on the figure. I would like to hear other reviewers' opinions as well.

not522:
I updated PR following your comments.

optuna/visualization/_param_importances.py
```diff
@@ -81,6 +86,11 @@ def objective(trial):
             A list of names of parameters to assess.
             If :obj:`None`, all parameters that are present in all of the completed trials are
             assessed.
+        target:
+            A function to specify the value to evaluate importances. If it is :obj:`None`, the
+            objective values are used.
+        target_name:
+            Target's name to display on the axis label.
```
HideakiImamura, line 93: We set this name as not the axis label but the title.
```suggestion
            Target's name to display on the title. The title would be "Importance for ``target_name``".
```

optuna/visualization/matplotlib/_param_importances.py
```diff
@@ -93,26 +98,33 @@ def objective(trial):
             A list of names of parameters to assess.
             If :obj:`None`, all parameters that are present in all of the completed trials are
             assessed.
+        target:
+            A function to specify the value to evaluate importances. If it is :obj:`None`, the
+            objective values are used.
+        target_name:
+            Target's name to display on the axis label.
```
HideakiImamura, line 105: Ditto.
```suggestion
            Target's name to display on the title. The title would be "Importance for ``target_name``".
```

HideakiImamura (commented):
Thanks for the update! I have two minor comments. Please take a look.

HideakiImamura:

optuna/visualization/_param_importances.py
not522: It's the title of the xaxis. I think "axis label" makes sense.

HideakiImamura: Ops! Sorry for the confusion. Let me resolve this conversation without any changes.

HideakiImamura (approved):
LGTM!

toshihikoyanase (approved):
Both methods worked as expected. LGTM!

plotly
image

matplotlib
tmp

PR #2126, opened 2020-12-20
Make `create_trial`'s documentation and tests richer

crcrpar:
The motivation is to make the documentation more friendly to users and to add some tests.

### DIFF
--- a/optuna/trial/_frozen.py
+++ b/optuna/trial/_frozen.py
@@ -546,15 +546,23 @@ def create_trial(
         Please note that this is a low-level API. In general, trials that are passed to objective
         functions are created inside :func:`~optuna.study.Study.optimize`.
 
+    .. note::
+        When ``state`` is :obj:`None` or ``TrialState.COMPLETE``, the following parameters are
+        required:
+        * ``params``
+        * ``distributions``
+        * ``value`` or ``values``
+
     Args:
         state:
             Trial state.
         value:
-            Trial objective value. Must be specified if ``state`` is :class:`TrialState.COMPLETE`.
+            Trial objective value. Must be specified if ``state`` is ``None``
+            or :class:`TrialState.COMPLETE`.
         values:
             Sequence of the trial objective values. The length is greater than 1 if the problem is
             multi-objective optimization.
-            Must be specified if ``state`` is :class:`TrialState.COMPLETE`.
+            Must be specified if ``state`` is ``None`` or :class:`TrialState.COMPLETE`.
         params:
             Dictionary with suggested parameters of the trial.
         distributions:
--- a/tests/trial_tests/test_trial.py
+++ b/tests/trial_tests/test_trial.py
@@ -627,6 +627,18 @@ def test_create_trial(state: TrialState) -> None:
     assert trial.datetime_start is not None
     assert (trial.datetime_complete is not None) == (state is None or state.is_finished())
 
+    with pytest.raises(ValueError):
+        create_trial(state=state, value=value, values=(value,))
+
+    if state is None:
+        with pytest.raises(ValueError):
+            create_trial(state=state, value=None, values=None)
+        # Raise `ValueError` when either `params` or `distributions` is `None`.
+        with pytest.raises(ValueError):
+            create_trial(state=state, value=value, params=params, distributions=None)
+        with pytest.raises(ValueError):
+            create_trial(state=state, value=value, params=None, distributions=distributions)
+
 
 def test_suggest_with_multi_objectives() -> None:
     study = create_study(directions=["maximize", "maximize"])


HideakiImamura:
@hvy @keisuke-umezawa Could you review this PR if you have time?

optuna/trial/_frozen.py
```diff
@@ -546,6 +546,16 @@ def create_trial(
...
+        or ``TrialState.COMPLETE``.
+
+    .. note::
+
+        Both ``params`` and ``distributions`` are required when ``state``
+        is ``None`` or ``TrialState.COMPLETE``.
+
     Args:
         state:
             Trial state.
```
keisuke-umezawa, line 561: How about specifying the default value here, e.g. `The default value of this argument is ``TrialState.COMPLETE``.`, because we can easily understand why we need to follow some rules when ``state`` is ``None`` or ``TrialState.COMPLETE``.

keisuke-umezawa (commented):
Added nits comment. Others look good to me!

optuna/trial/_frozen.py
crcrpar: > because we can easily understand why we need to follow some rules when `state` is `None` or `TrialState.COMPLETE`.

Sorry, but "some rules" is too ambiguous to me.

Also I don't know what you mean by "the default value",  I guess you refer `state` from the words following "e.g.". If this is the case, what you say is wrong: the default value of `state` is `None`, and Optuna regard `state` of `None` as `TrialState.COMPLETE`.

Anyway, I respect your point, but I can't entirely agree with it because the default values are obvious (see the capture below) even if the type hints are kind of annoying to me.
image

optuna/trial/_frozen.py
```diff
@@ -546,6 +546,16 @@ def create_trial(
         Please note that this is a low-level API. In general, trials that are passed to objective
         functions are created inside :func:`~optuna.study.Study.optimize`.
 
+    .. note::
+
+        You cannot specify both ``value`` and ``values`` when ``state`` is ``None``
+        or ``TrialState.COMPLETE``.
```
hvy, line 552: Isn't this limitation true independent of the state (or, it should be at least)?

tests/test_trial.py
```diff
@@ -1271,6 +1271,19 @@ def test_create_trial(state: TrialState) -> None:
...
+        with pytest.raises(ValueError):
+            create_trial(state=state, value=None, values=None)
+        # Raise `ValueError` when both `value` and `values` are passed.
+        with pytest.raises(ValueError):
+            create_trial(state=state, value=value, values=[value, value])
+        # Raise `ValueError` when either `params` or `distributions` is `None`.
+        with pytest.raises(ValueError):
+            create_trial(state=state, value=value, params=params, distributions=None)
+        with pytest.raises(ValueError):
+            create_trial(state=state, value=value, params=None, distributions=distributions)
```
hvy, line 1285: How about parametrizing it and defining a separate function. It's quite independent of the above logic (`trial`) and would make test failures easier to comprehend.

hvy (changes requested):
Thanks for the PR. It makes perfect sense to improve the docs. I made it rather brief since it was an experimental low level API but it's seemingly used already.

tests/test_trial.py
crcrpar: I think it's a bit too early to separate the function as the API is still experimental, and the function should cover the whole bunch of  `create_trial` cases as it's named "create_trial". 

Also, separating functions is not always great as it increases the number of lines, and test_trial.py is already large enough to lose comprehensiveness.

hvy: It's not a strong opinion so let's then leave it as is.

optuna/trial/_frozen.py
crcrpar: yeah, that should be the case.

I added this note because the current latest documentation sounded like both `value` and `values` are required as per the caption below.

image

optuna/trial/_frozen.py
keisuke-umezawa: Oh, sorry, `TrialState.COMPLETE` is not the default value, actually. `None` is converted to `TrialState.COMPLETE`.

In that case, how about writing as follows?

```
        value:
            Trial objective value. Must be specified if ``state`` is :class:`TrialState.COMPLETE` or ``None``.
        values:
            Sequence of the trial objective values. The length is greater than 1 if the problem is
            multi-objective optimization.
            Must be specified if ``state`` is :class:`TrialState.COMPLETE` or ``None``.
```

crcrpar: I'm sorry, but which part are you mentioning?

The quote block is just a copy&paste of the part of `value` and `values` of the API documentation.

keisuke-umezawa: I meant replacing ":class:`TrialState.COMPLETE`" to ":class:`TrialState.COMPLETE` or ``None``".

crcrpar: I'm so sorry, but I still do not understand what you say.

This is a thread on the addition of `note` block about `params` and `distributions`. But now what you suggest looks like a change on the descriptions of `value` and `values`.

Do you want me to update the descriptions of `value` and `values` in the parameters section?
Do you think the addition of `note` which I authored is not necessary here?

keisuke-umezawa: > Do you want me to update the descriptions of value and values in the parameters section?

Yes. When I checked the note and the descriptions of `value` and `values`, it seems 
inconsistent for me, because the note says `Both ``params`` and ``distributions`` are required when ``state`` is None or TrialState.COMPLETE`, while the descriptions of `value` and `values` say `Must be specified if ``state`` is TrialState.COMPLETE`.

keisuke-umezawa: > Do you want me to update the descriptions of value and values in the parameters section?

Yes. When I checked the note and the descriptions of value and values, it seems
inconsistent for me, because the note says `You cannot specify both ``value`` and ``values`` when ``state`` is ``None`` or ``TrialState.COMPLETE`` ` by specifying ` ``None`` or ``TrialState.COMPLETE`` `, while the descriptions of value and values say `Must be specified if ``state`` is TrialState.COMPLETE ` by specifying only ` ``TrialState.COMPLETE`` ` .

optuna/trial/_frozen.py
hvy: Ah, I see. So, the same phrase and conditioning on the state is currently repeated quite often. I'm wondering if it would be easier to comprehend if put all these conditions under a single note, and simplify the notes, as well as the parameter documentation.

```
.. note::
    When ``state`` is ``TrialState.COMPLETE`` or :obj:`None`, the following parameters must not be :obj:`None`. Otherwise, errors are raised.

  - ``params``
  - ``distributions``
  - ``value`` or ``values``
```

That ``value`` and ``values`` cannot be specified at the same time is further elaborated under `Raises:`.

hvy (commented):
Sorry could you fix the conflict? Otherwise LGTM.

crcrpar:
Okay, resolved conflicts.

hvy (approved):
LGTM!

keisuke-umezawa (approved):
LGTM!

PR #2132, opened 2020-12-21
Additional docs about `optuna.multi_objective` deprecation

hvy:
## Motivation

Follow-up to https://github.com/optuna/optuna/pull/2124.

## Description of the changes

Adds additional explanatory docs about the deprecation of `optuna.multi_objective`. Format mimics https://github.com/optuna/optuna/blob/master/docs/source/reference/structs.rst.

## Note

It might be open for discussion whether this is the correct formulation since the submodule might come to include other MO features.

### DIFF
--- a/docs/source/reference/multi_objective/index.rst
+++ b/docs/source/reference/multi_objective/index.rst
@@ -3,6 +3,8 @@
 optuna.multi_objective
 ======================
 
+This module is deprecated, with former functionality moved to :class:`optuna.trial`, :class:`optuna.study`, :class:`optuna.samplers` and :class:`optuna.visualization`.
+
 .. toctree::
     :maxdepth: 1
 
--- a/docs/source/reference/multi_objective/index.rst
+++ b/docs/source/reference/multi_objective/index.rst
@@ -3,7 +3,7 @@
 optuna.multi_objective
 ======================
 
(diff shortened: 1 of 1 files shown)


HideakiImamura (approved):
LGTM.

HideakiImamura:

toshihikoyanase:
> It might be open for discussion whether this is the correct formulation since the submodule might come to include other MO features.

IMO, we may be able to place utilities such as hypervolume calculation and Pareto-front extraction. I think it may be too early to do so because they are only used in samplers currently. So, I agree with the deprecation of `optuna.multi_module`.

docs/source/reference/multi_objective/index.rst
```diff
@@ -3,6 +3,8 @@
 optuna.multi_objective
 ======================
 
+This module is deprecated, with former functionality moved to :class:`optuna.trial`, :class:`optuna.study`, :class:`optuna.samplers` and :class:`optuna.visualization`.
```
toshihikoyanase, line 6: A nit, but how about arranging the modules in the same order in the table?
In addition, we can use `:mod:` instead of `:class:` for the modules.

```suggestion
This module is deprecated, with former functionality moved to :mod:`optuna.samplers`, :mod:`optuna.study`, :mod:`optuna.trial` and :mod:`optuna.visualization`.
```

hvy:
Thanks for the review, I applied your suggestion. PTAL.

toshihikoyanase (approved):

PR #2133, opened 2020-12-21
`_run_trial` refactoring

hvy:
## Motivation

`_run_trial` responsible for the "ask and tell" has become quite complicated and is difficult to make changes to. More specifically, it contains many conditions with early returns, re-raises, several calls to tell, etc. 

## Description of the changes

Refactors `_run_trial` to have a single return, single re-raise, single call to tell.

### DIFF
--- a/optuna/_optimize.py
+++ b/optuna/_optimize.py
@@ -3,7 +3,9 @@ import datetime
 import gc
 import math
 import sys
+from typing import Any
 from typing import Callable
+from typing import cast
 from typing import List
 from typing import Optional
 from typing import Sequence
@@ -187,7 +189,10 @@ def _run_trial(
 
     state: Optional[TrialState] = None
     values: Optional[List[float]] = None
-    err: Optional[Exception] = None
+    func_err: Optional[Exception] = None
+    func_err_fail_exc_info: Optional[Any] = None
+    # Set to a string if `func` returns correctly but the return value violates assumptions.
+    values_conversion_failure_message: Optional[str] = None
 
     try:
         value_or_values = func(trial)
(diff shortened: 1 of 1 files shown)


HideakiImamura:
@crcrpar Would you mind reviewing this PR if you have time?

HideakiImamura (approved):
Great code fix. LGTM.

HideakiImamura:

optuna/_optimize.py
```diff
@@ -184,58 +185,69 @@ def _run_trial(
...
+    elif state == TrialState.FAIL:
+        if err is not None:
+            _logger.warning(
+                "Trial {} failed because of the following error: {}".format(
+                    trial.number, repr(err)
+                ),
+                exc_info=exc_info,
+            )
+        else:
+            _logger.warning(failure_message)
```
toshihikoyanase, line 229: These nested conditional branches are not intuitive for me. For example, it's a bit difficult for me to check if `failure_message` in L229 is not undefined.
Not a strong opinion, but how about moving the logging to the try-except block in L192-L211?

<details>
<summary>Alternative code which includes logging in the try-except block</summary>

``` python
    try:
        value_or_values = func(trial)
    except exceptions.TrialPruned as e:
        # TODO(mamu): Handle multi-objective cases.
        # Register the last intermediate value if present as the value of the trial.
        # TODO(hvy): Whether a pruned trials should have an actual value can be discussed.
        state = TrialState.PRUNED
        last_step = frozen_trial.last_step
        values = None if last_step is None else [frozen_trial.intermediate_values[last_step]]
        err = e
        _logger.info("Trial {} pruned. {}".format(trial.number, str(err)))
    except Exception as e:
        state = TrialState.FAIL
        values = None
        err = e
        exc_info = sys.exc_info()
        _logger.warning(
            "Trial {} failed because of the following error: {}".format(
                trial.number, repr(err)
            ),
            exc_info=exc_info,
        )
    else:
        state, values, failure_message = _check_and_convert_to_values(
            len(study.directions), value_or_values, trial
        )
        if state == TrialState.COMPLETE:
            assert values is not None
            study._log_completed_trial(trial, values)
        else:
            _logger.warning(failure_message)

    study._tell(trial, state, values)

    if state == TrialState.FAIL and err is not None and not isinstance(err, catch):
        raise err
    return trial
```

hvy: Thanks for your input. I understand your concern about the readability but I'm not sure about the suggested change and whether it's addressing the maintainability. One of the motivation behind this PR was to prepare fore https://github.com/optuna/optuna/pull/2134, where we're introducing trial post-processing. This will remain nontrivial, unless we reconsider the behavior and error handling in #2134. What do you think? This is also somewhat beyond the scope but generally speaking, we should log after a successful write to the storage.

hvy: By the way, I think we can introduce another variable/rename existing variables to address readability.

toshihikoyanase: > One of the motivation behind this PR was to prepare for #2134, where we're introducing trial post-processing. 

I see. #2134 will introduce another try-except block to call `_after_trial`, and then it's reasonable to extract the logging.

> I think we can introduce another variable/rename existing variables to address readability.

I agree. `exc_info` and `failure_message` may need refactoring because a linter (Pyright) shows unbound warnings as follows:

image

hvy: Thanks for the inputs and the linter warnings. I'll try to address them. 

By the way, are they from PyCharm? You don't happen to know if there are other alternatives out there?

toshihikoyanase: > are they from PyCharm?

They are from [pyright](https://github.com/microsoft/pyright) executed via [VSCode plugin](https://marketplace.visualstudio.com/items?itemName=ms-pyright.pyright). The [commadl-line tool of pyright](https://github.com/microsoft/pyright/blob/master/docs/command-line.md) is also available. I think the use of pyright is optional, though.

hvy: Noticed I don't even have npm installed in my current environment... I tried to address your comment though, PTAL.

toshihikoyanase (approved):

PR #2134, opened 2020-12-21
Trial post processing

hvy:
## Motivation

~Depends on https://github.com/optuna/optuna/pull/2133.~ (edit: merged)

Allow experimental trial post post-processing. As a consequence, the `BoTorchSampler` constraints callback assumptions are relaxed and they no longer have to be deterministic since they'll only be computed once per trial. The `GridSampler` can also make use of this change to stop the study.

## Description of the changes

Introduces `BaseSampler.after_trial` to allow trial post-processing by each sampler.

## TODO

- [x] Make sure API looks okay. Ensure compatibility with possible future ask-and-tell API.
- [x] Tests. Verify error handling and edge cases.

### DIFF
--- a/optuna/_optimize.py
+++ b/optuna/_optimize.py
@@ -224,8 +224,8 @@ def _run_trial(
     except Exception as e:
         state = TrialState.FAIL
         values = None
-        err = e
-        exc_info = sys.exc_info()
+        func_err = e
+        func_err_fail_exc_info = sys.exc_info()
 
     study._tell(trial, state, values)
 
--- a/optuna/_optimize.py
+++ b/optuna/_optimize.py
@@ -221,32 +221,29 @@ def _run_trial(
 
     try:
         trial._after_func(state, values)
-    except Exception as e:
-        state = TrialState.FAIL
-        values = None
-        func_err = e
-        func_err_fail_exc_info = sys.exc_info()
+    except Exception:
+        raise
+    finally:
+        study._tell(trial, state, values)
 
-    study._tell(trial, state, values)
-
-    if state == TrialState.COMPLETE:
-        study._log_completed_trial(trial, cast(List[float], values))
-    elif state == TrialState.PRUNED:
-        _logger.info("Trial {} pruned. {}".format(trial.number, str(func_err)))
-    elif state == TrialState.FAIL:
-        if func_err is not None:
-            _logger.warning(
-                "Trial {} failed because of the following error: {}".format(
-                    trial.number, repr(func_err)
-                ),
-                exc_info=func_err_fail_exc_info,
-            )
-        elif values_conversion_failure_message is not None:
-            _logger.warning(values_conversion_failure_message)
+        if state == TrialState.COMPLETE:
+            study._log_completed_trial(trial, cast(List[float], values))
+        elif state == TrialState.PRUNED:
+            _logger.info("Trial {} pruned. {}".format(trial.number, str(func_err)))
+        elif state == TrialState.FAIL:
+            if func_err is not None:
+                _logger.warning(
+                    "Trial {} failed because of the following error: {}".format(
+                        trial.number, repr(func_err)
+                    ),
+                    exc_info=func_err_fail_exc_info,
+                )
+            elif values_conversion_failure_message is not None:
+                _logger.warning(values_conversion_failure_message)
+            else:
+                assert False, "Should not reach."
         else:
             assert False, "Should not reach."
-    else:
-        assert False, "Should not reach."
 
     if state == TrialState.FAIL and func_err is not None and not isinstance(func_err, catch):
         raise func_err
--- a/optuna/_optimize.py
+++ b/optuna/_optimize.py
@@ -210,6 +210,14 @@ def _run_trial(
             len(study.directions), value_or_values, trial
         )
 
+    try:
+        trial._after_func(state, values)
+    except Exception as e:
+        state = TrialState.FAIL
+        values = None
+        err = e
+        exc_info = sys.exc_info()
+
     study._tell(trial, state, values)
 
     if state == TrialState.COMPLETE:
--- a/optuna/integration/botorch.py
+++ b/optuna/integration/botorch.py
@@ -463,14 +463,12 @@ class BoTorchSampler(BaseSampler):
                 constraints = study._storage.get_trial_system_attrs(trial._trial_id)[
                     "botorch:constraints"
                 ]
-                # `constraints` can be `None` if `constraints_func` raised an error. The best we
-                # can do is to fill with NaN.
-                if constraints is not None:
+                assert isinstance(constraints, tuple)
+                if con is None:
                     n_constraints = len(constraints)
-                    if con is None:
-                        con = numpy.full((n_trials, n_constraints), numpy.nan, dtype=numpy.float64)
+                    con = numpy.empty((n_trials, n_constraints), dtype=numpy.float64)
 
-                    con[trial_idx] = constraints
+                con[trial_idx] = constraints
 
         values = torch.from_numpy(values)
         params = torch.from_numpy(params)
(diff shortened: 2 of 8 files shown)


optuna/samplers/_base.py
```diff
@@ -137,6 +140,37 @@ def sample_independent(
 
         raise NotImplementedError
 
+    def after_trial(
+        self,
+        study: Study,
+        trial: FrozenTrial,
+        state: TrialState,
+        values: Optional[Sequence[float]],
```
hvy, line 148: Just accepting a `FrozenTrial` with its state and values set would speak for a cleaner API. This is however not possible since to the storage cache mechanism assumes that a finished trial is never modified.

hvy:
Exceptions raised from `after_trial` are not affected by the `catch` argument to `Study.optimize` and will always be propagated to the the optimize caller. This makes sense if we for instance would want to introduce a "before_trial" to e.g. check for SO/MO compatibility raising appropriate errors. These errors should likely not be affected by `catch` neither.

hvy:
Removed WIP, PTAL.

optuna/samplers/_base.py
```diff
@@ -137,6 +140,37 @@ def sample_independent(
...
+    ) -> None:
+        """Trial post-processing.
+
+        This method is called after the objective function returns and right before the trials is
+        finished and its state is stored.
+
+        .. note::
+            Added in v2.4.0 as an experimental feature. The interface may change in newer versions
+            without prior notice. See https://github.com/optuna/optuna/releases/tag/v2.4.0.
```
hvy, line 157: Avoided using the `@experimental` decorator since this method is called internally and we don't want actual warnings being propagated to the user. We can silence warning internally upon invocations but given that particular samplers could raise actual warnings, might not be very smart.

optuna/samplers/_grid.py
```diff
@@ -185,6 +179,23 @@ def sample_independent(
...
+        trial: FrozenTrial,
+        state: TrialState,
+        values: Optional[Sequence[float]],
+    ) -> None:
+        target_grids = self._get_unvisited_grid_ids(study)
+
+        if len(target_grids) == 0:
+            study.stop()
+        elif len(target_grids) == 1:
+            if state == TrialState.COMPLETE:
```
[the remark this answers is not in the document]
hvy: Thanks, makes sense. I also added a test for this case in the latest commit.

optuna/integration/botorch.py
```diff
@@ -488,8 +459,33 @@ def sample_relative(
...
+            if con is None:
+                warnings.warn(
+                    "`constraints_func` was given but no call to it correctly computed "
+                    "constraints. Constraints passed to `candidates_func` will be `None`."
+                )
+            elif numpy.isnan(con).any():
+                warnings.warn(
+                    "`constraints_func` was given but some calls to it did not correctly compute "
+                    "constraints. Constraints passed to `candidates_func` will contain NaN."
+                )
```
HideakiImamura, line 488: It seems that there is no test for this case. How about adding it?

optuna/integration/botorch.py
```diff
@@ -555,3 +551,31 @@ def sample_independent(
...
+
+            try:
+                con = self._constraints_func(trial)
+                if not isinstance(con, (tuple, list)):
+                    warnings.warn(
+                        f"Constraints should be a sequence of floats but got {constraints}."
+                    )
+                constraints = tuple(con)
+            except Exception:
+                raise
```
HideakiImamura, line 573: If calculated `con` cannot be cast to `tuple`, the error will be raised. How about testing it?

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM. I have some early comments on tests.

optuna/samplers/_grid.py
```diff
@@ -185,6 +179,23 @@ def sample_independent(
...
+    ) -> None:
+        target_grids = self._get_unvisited_grid_ids(study)
+
+        if len(target_grids) == 0:
+            study.stop()
+        elif len(target_grids) == 1:
+            if state.is_finished():
+                grid_id = study._storage.get_trial_system_attrs(trial._trial_id)["grid_id"]
+                if grid_id == target_grids[0]:
+                    study.stop()
```
g-votte, line 197:
```suggestion
            grid_id = study._storage.get_trial_system_attrs(trial._trial_id)["grid_id"]
            if grid_id == target_grids[0]:
                study.stop()
```

Thank your for applying my previous suggestion.

I've noticed that `if state.is_finished():` is now redundant because the state should be either of `COMPLETE`, `PRUNED`, or `FAIL` as long as `after_trial()` is triggered.

Sorry for suggesting for the same point twice.

g-votte (changes requested):

optuna/samplers/_grid.py
hvy: You're right. Thanks.

optuna/integration/botorch.py
hvy: Ah, it should be tested already by `test_botorch_constraints_func_invalid_type`.

optuna/integration/botorch.py
hvy: Sure, let me add tests to check the elements for NaN as well.

hvy: Added a test for the warning.

hvy:
Thanks for the review comments, PTAL.

optuna/_optimize.py
```diff
@@ -219,26 +219,31 @@ def _run_trial(
...
-                    trial.number, repr(func_err)
-                ),
-                exc_info=func_err_fail_exc_info,
-            )
-        elif values_conversion_failure_message is not None:
-            _logger.warning(values_conversion_failure_message)
+    try:
+        trial._after_func(state, values)
+    except Exception:
+        raise
```
HideakiImamura, line 225: It is more consistent to chain the exception like https://github.com/optuna/optuna/pull/1781.

hvy: Hm, #1781 fixes exceptions to not silence caught exceptions. Here, we're propagating and not silencing anything. Do you mean to construct an error here and chain with it? I think there are pros and cons but ss can be seen in #1781, we don't do that too often. In my understanding, it'll be more to document and understand from the user perspective.

HideakiImamura: Thanks for the detailed explanation. I understood. We should not construct an error here.

HideakiImamura (approved):
Thanks for the discussions. LGTM.

toshihikoyanase (commented):
It looks great to me. I add some small comments, but I think they are optional.

g-votte (approved):
I mainly reviewed the logic of the grid sampler, and it looks good to me.

@toshihikoyanase  
Please merge after your review.

optuna/integration/botorch.py
```diff
@@ -488,8 +459,33 @@ def sample_relative(
...
+                        )
+
+                    con[trial_idx] = constraints
+
+        if self._constraints_func is not None:
+            if con is None:
+                warnings.warn(
+                    "`constraints_func` was given but no call to it correctly computed "
+                    "constraints. Constraints passed to `candidates_func` will be `None`."
+                )
```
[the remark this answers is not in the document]
hvy: Added a test.

optuna/integration/botorch.py
```diff
@@ -488,8 +459,33 @@ def sample_relative(
                     value *= -1
                 values[trial_idx, obj_idx] = value
 
-            if con is not None:
-                con[trial_idx] = self._trial_constraints[trial_idx]
+            if self._constraints_func is not None:
+                constraints = study._storage.get_trial_system_attrs(trial._trial_id)[
+                    "botorch:constraints"
+                ]
```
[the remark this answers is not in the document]
hvy: Thanks, fixed the logic and added a test.

hvy:
Thanks for the review comments again, they should be addressed. PTAL.

The tests are quite involved now for the `BoTorchSampler` even though it really wasn't the main focus of this PR. Given that it's an experimental sampler, please don't hesitate to make changes to those tests as needed.

toshihikoyanase (approved):

Issue #2130, opened 2020-12-21
Implement `Study._has_multi_objective()` to check whether study has multiple objectives


HideakiImamura:
## Motivation
For developers, it is useful to have a method to determine if a study has more than one objective, which is easy to understand. See https://github.com/optuna/optuna/pull/2117#pullrequestreview-556116066 for the detail.

The name of the method can be improved.

himkt:
Resolved by #2142. (Thank you @nyanhi!)

PR #2142, opened 2020-12-22
Implement `Study._is_multi_objective()` to check whether study has multiple objectives

nyanhi:
## Motivation
To address the issue noted on https://github.com/optuna/optuna/issues/2130
I slightly changed the name from "_has_multi_objective" to "_is_multi_objective", but please let me know if "_has" sounds better.

## Description of the changes
Added simple method `Study._is_multi_objective()` which indicates if the study has more than 1 objectives.

Once I confirmed if this PR is directionally fine, I will also search and replace the code where using `len(study.directions) > 1` elsewhere e.g. https://github.com/optuna/optuna/blob/f65e66b9222c00e0d33e8f50398fef8045608f01/optuna/visualization/_contour.py#L87.


optuna/study.py
```diff
@@ -149,6 +149,15 @@ def trials(self) -> List[FrozenTrial]:
 
         return self.get_trials(deepcopy=True, states=None)
 
+    def _is_multi_objective(self) -> bool:
```
himkt, line 152: I'm not confident but I'm wondering whether this method should be public, as it could be invoked from outside the class (e.g. sampler, pruner). I want to hear other opinions.

optuna/study.py
```diff
@@ -149,6 +149,15 @@ def trials(self) -> List[FrozenTrial]:
...
+        """Return True if the study has multi objectives.
```
himkt, line 153:
```suggestion
        """Return :obj:`True` if the study has multi objectives.
```

We usually annotate an object with `:obj:`.

ref. https://github.com/optuna/optuna/blob/e2e2a5cd91f3df967838cea525b28c49993b6721/optuna/storages/_base.py#L338

himkt, line 153:
```suggestion
        """Return True if the study has multiple objectives.
```

I think it would be better to avoid an abbreviation.

optuna/study.py
```diff
@@ -149,6 +149,15 @@ def trials(self) -> List[FrozenTrial]:
...
+
+        Returns:
+            A boolean value indicates if the list `study.directions` has more than 1 element or not.
```
himkt, line 156:
```suggestion
            A boolean value indicates if `study.directions` has more than 1 element or not.
```

[IMO] We can remove `the list` here. What do you think?

optuna/study.py
```diff
@@ -149,6 +149,15 @@ def trials(self) -> List[FrozenTrial]:
...
+        """
+
+        return True if len(self.directions) > 1 else False
```
himkt, line 159:
```suggestion
        return len(self.directions) > 1
```

operator `>` returns a boolean value. So we can use the result directly.

tests/test_study.py
```diff
@@ -1033,9 +1033,11 @@ def test_log_completed_trial_skip_storage_access() -> None:
 def test_create_study_with_multi_objectives() -> None:
     study = optuna.create_study(directions=["maximize"])
     assert study.direction == StudyDirection.MAXIMIZE
+    assert study._is_multi_objective() is False
```
himkt, line 1036:
```suggestion
    assert not study._is_multi_objective()
```

Same as the previous comment. 👍

tests/test_study.py
```diff
@@ -1033,9 +1033,11 @@ def test_log_completed_trial_skip_storage_access() -> None:
...
 
     study = optuna.create_study(directions=["maximize", "minimize"])
     assert study.directions == [StudyDirection.MAXIMIZE, StudyDirection.MINIMIZE]
+    assert study._is_multi_objective() is True
```
himkt, line 1040:
```suggestion
    assert study._is_multi_objective()
```

Same here.

himkt (commented):
Hello @nyanhi, thank you for your contribution!
I left some comments, please take a look.

> I slightly changed the name from "_has_multi_objective" to "_is_multi_objective"

Thank you for you consideration! I'd like other developer's opinion, as I'm not confident which prefix `_has` or `_is` is good.
(It looks like there are some method having a prefix `_is` but no one having `_has` in Optuna. So `_is` would be better for the choice as you suggested.)

himkt:
CI failed with the following error.

```
self = <optuna.pruners._hyperband.HyperbandPruner._create_bracket_study.<locals>._BracketStudy object at 0x7fa287709750>
attr_name = '_is_multi_objective'

    def __getattribute__(self, attr_name):  # type: ignore
        if attr_name not in _BracketStudy._VALID_ATTRS:
            raise AttributeError(
>               "_BracketStudy does not have attribute of '{}'".format(attr_name)
            )
E           AttributeError: _BracketStudy does not have attribute of '_is_multi_objective'

optuna/pruners/_hyperband.py:306: AttributeError
```

https://github.com/optuna/optuna/runs/1595168444

I think we have to allow `study` to have an attribute `_is_multi_objective` in BracketStudy.
https://github.com/nyanhi/optuna/blob/3a4faa7e37676d23cf24199706f957e24bfabcbb/optuna/pruners/_hyperband.py#L271-L282

nyanhi:
Thank you for the detailed comments! I will read the code for BracketStudy and make the changes suggested.

optuna/study.py
HideakiImamura: IMO, this method is only invoked by developers, so it should be private to align other codes of Optuna (e.g. `Study._ask`, `Study._tell`, or `Study._log_completed_trials`)

optuna/study.py
HideakiImamura: I agree with you. The word "directions" is itself plural.

HideakiImamura (commented):
Thanks for the PR! LGTM besides @himkt's comments.

optuna/study.py
nyanhi: Thank you for the comments both. I will keep it private `_is...`.

nyanhi:
FYI - I'm replacing codes where using `len(Study.directions)` and expect to complete it by tomorrow morning.

nyanhi:
I noticed below code is using `len(study.directions)` to check if `study: "multi_objective.study.MultiObjectiveStudy"` is multi-objective or not. Given the class `MultiObjectiveStudy` is deprecated and already has a method `n_objective` which tells us `len(study._directions)`, I'm going to leave it as is. Please kindly let me know if any change is required for this.

https://github.com/optuna/optuna/blob/77aee02b935de1f7bb1e962aa84aedf8b254bd1d/optuna/multi_objective/samplers/_motpe.py#L168-L181

HideakiImamura (commented):
You may miss `optuna.visualization.{parallel_coordinate, param_importances, slice} and its matplot versions. Please apply your `Study._is_multi_objectives=()` to them.

nyanhi:
Thank you for pointing out @HideakiImamura! 
I merged the master branch and found some more `len(study.directions)` as well.

HideakiImamura (approved):

himkt (approved):
Thank you for your contribution! LGTM!

PR #2152, opened 2020-12-24
Fix wrong wording in distributed execution tutorial

crcrpar:


tutorial/10_key_features/004_distributed.py
```diff
@@ -4,7 +4,7 @@
 Easy Parallelization
 ====================
 
-It's straightforward to parallelize ``optuna.optimize()``.
+It's straightforward to parallelize :func:`optuna.study.Study.optimize`.
```
himkt, line 7: [question] Is a word such as `using` or `by` needed between `parallelize` and `:func:...`?

himkt (approved):
Nice catch!

LGTM ~with a question~.

HideakiImamura (approved):
Good catch! LGTM!

PR #2153, opened 2020-12-25
[Opinion Wanted] Set value and state to trial passed to `constraints_func` of `BotorchSampler`

toshihikoyanase:
## Motivation

In `constraints_func`, users cannot obtain trial values with the `trial` argument. In my understanding, this is because `after_trial` method is called before persisting the trial. Due to this limitation, we need to store objective values to `trial.user_attrs` to access them in `constraint_func`.

This design is quite consistent, but I guess it may not be intuitive for novice users of `BoTorchSampler`. They may want to access `trial.values` directly. Unlike `after_trial`, users cannot access the `study` object and I think they will not face any inconsistency of trial states even if we overwrite the `value` of `trial`.

## Description of the changes

Set `value` and `state` to `trial` before it is passed to the `constraints_func`.

### DIFF
--- a/optuna/integration/botorch.py
+++ b/optuna/integration/botorch.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+import copy
 import math
 from typing import Any
 from typing import Callable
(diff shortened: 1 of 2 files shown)


hvy:
Let me reopen this issue since it's still worth discussing.

PR #2156, opened 2020-12-27
`TFKerasPruningCallback` to warn when an evaluation metric does not exist

bigbird555:
## Motivation

## Description of the changes


hvy (approved):
Looks consistent and good, LGTM! Thanks as always.

PR #2157, opened 2020-12-27
`PyTorchLightningPruningCallback` to warn when an evaluation metric does not exist

bigbird555:
## Motivation

## Description of the changes


optuna/integration/pytorch_lightning.py
```diff
@@ -45,6 +47,11 @@ def on_validation_end(self, trainer: Trainer, pl_module: LightningModule) -> Non
         epoch = pl_module.current_epoch
         current_score = logs.get(self.monitor)
         if current_score is None:
+            message = (
+                "The metric '{}' is not in the evaluation logs for pruning. "
+                "Please make sure you set the correct metric name.".format(self._monitor)
```
hvy, line 52: Is the underscore redundant?
```suggestion
                "Please make sure you set the correct metric name.".format(self.monitor)
```

hvy (changes requested):
Basically LGTM but the CI seems to be failing. Tried to leave a comment on the code that looked suspicious.

tests/integration_tests/test_pytorch_lightning.py
```diff
@@ -87,3 +89,13 @@ def objective(trial: optuna.trial.Trial) -> float:
...
+
+
+def test_pytorch_lightning_pruning_callback_monitor_is_invalid() -> None:
+
+    study = optuna.create_study(pruner=DeterministicPruner(True))
+    trial = create_running_trial(study, 1.0)
+    callback = PyTorchLightningPruningCallback(trial, "InvalidMonitor")
+
+    with pytest.warns(UserWarning):
+        callback.on_epoch_end(0, {"loss": 1.0})
```
hvy, line 101: Is `on_epoch_end` the correct callback method? No warning is seemingly emitted, although the implementation of the integration module looks correct.

hvy:
Seems like tests are still failing. Could you take a look at the log from the CI?

tests/integration_tests/test_pytorch_lightning.py
```diff
@@ -87,3 +89,13 @@ def objective(trial: optuna.trial.Trial) -> float:
...
     assert study.trials[0].value == 1.0
+
+
+def test_pytorch_lightning_pruning_callback_monitor_is_invalid() -> None:
+
+    study = optuna.create_study(pruner=DeterministicPruner(True))
+    trial = create_running_trial(study, 1.0)
+    callback = PyTorchLightningPruningCallback(trial, "InvalidMonitor")
+
+    with pytest.warns(UserWarning):
```
nzw0301, line 100:
```suggestion
        trainer = pl.Trainer(
        min_epochs=0,  # Required to fire the callback after the first epoch.
        max_epochs=1,
        checkpoint_callback=False,
        callbacks=[callback],
    )
    model = Model()

    with pytest.warns(UserWarning):
        callback.on_validation_end(trainer, model)

```

tests/integration_tests/test_pytorch_lightning.py
```diff
@@ -87,3 +89,13 @@ def objective(trial: optuna.trial.Trial) -> float:
...
+        callback.on_validation_end(0, {"loss": 1.0})
```
nzw0301, line 101:
```suggestion

    trainer = pl.Trainer(
        min_epochs=0,  # Required to fire the callback after the first epoch.
        max_epochs=1,
        checkpoint_callback=False,
        callbacks=[callback],
    )
    model = Model()

    with pytest.warns(UserWarning):
        callback.on_validation_end(trainer, model)

```

nzw0301: I think the current `on_validation_end`'s arguments look incorrect. How about this? Locally, `tests/integration_tests/test_pytorch_lightning.py` is passed.

nzw0301:
I think it too late to leave a comment, but I would like to leave some comment.

Personally, I think it looks natural to be checked in `__init__` rather than `on_validation_end` or raise `RuntimeError` in `on_validation_end` by default for consistency.

According to [the official document](https://pytorch-lightning.readthedocs.io/en/stable/callbacks.html) of `pytorch-lightning`, 

> Directly calling methods (eg. on_validation_end) is strongly discouraged.

In addition, if we call `trainer.fit()` with an invalid `monitor`, PyTorch-Lightning raises `RuntimeError` by default [[ref](https://github.com/PyTorchLightning/pytorch-lightning/blob/4913cbb987a0516f8b33c016134b19c0588d107a/pytorch_lightning/callbacks/early_stopping.py#L136)]: when optuna shows the warning message, Pytorch-lightning also shows a similar message or even raises the error in practice.

bigbird555:
@hvy @nzw0301 
Thank you for your replies. Your advice is very helpful.
However, I've been a bit pressed for time lately, so I'll have to wait until I'm a bit more settled before I can consider the procedure to resolve this...

nzw0301:
@bigbird555 any update? :) If you are still busy, please feel free to ask us to take over this pull request.

bigbird555:
I'm sorry for making you wait so long. 
I'm not really sure what to do about this at the moment. 
So, if you don't have time for some reason, please feel free to take over the PR. 
Thank you for taking time out of your busy schedule to read this.

hvy (approved):
Thanks for the heads up. Personally, LGTM.

@nzw0301 could you have a look as well?  
I think we can follow it up with a different ticket if you think we should reconsider the behavior since the author seems busy but please let me know what you think. Also, just note that the pytorch-lightning's e.g. `EarlyStopping` seemingly provides a `strict: bool` argument to control whether it should result in an error or warning https://github.com/PyTorchLightning/pytorch-lightning/blob/4913cbb987a0516f8b33c016134b19c0588d107a/pytorch_lightning/callbacks/early_stopping.py#L136

nzw0301:
@bigbird555 Thank you for your quick response! Please do not worry about that. 

Indeed, when I read my previous comments, https://github.com/optuna/optuna/pull/2157#issuecomment-752264876, it seems weird -- this PR and the behaviour of PyTorch-lightning for invalid metric are different things. Sorry for confusing you. So the current changes look great to me.

nzw0301 (approved):

nzw0301:
@hvy Thank you for sharing the information. I realized that we do not need to create a follow-up change for this PR since pytorch-lightning cannot cause an error for the optuna callback with the invalid metric name.

PR #2158, opened 2020-12-28
Ask and Tell API (`Study.ask`, `Study.tell`)

hvy:
## Motivation

`Study.optimize` is a high-level API that takes care of trial creation, objective function evaluation, parallelism, etc. Sometimes, lower level operations are useful to gain finer control of the lifetime of a trial. Such cases involve arbitrary scheduling of trial suggestion, objective evaluation, for instance to have those two run on different processes.

## Description of the changes

Introduces an ask and tell API to provide a lower level alternative to `Study.optimize`.

~Depends on https://github.com/optuna/optuna/pull/2168.~ Edit: Merged.

### Example Usage

```python
# NOTE: Where `Trial`s are passed as arguments, trial numbers can be passed as well.

# Example 1: Simple case without pruning and no error handling.
trial = study.ask()
values = objective(trial)
study.tell(trial, values)

# Example 2: With error handling.
trial = study.ask()

values = None
try:
    values = objective(trial)
except TrialPruned:
    state = TrialState.PRUNED
except Exception:
    state = TrialState.FAIL
else:
    state = TrialState.COMPLETE
finally:
    study.tell(trial, values=values, state=state)

# Example 3: No `func` objective function.
trial = study.ask()

loss = None
state = TrialState.COMPLETE

lr = trial.suggest_float("lr", 1e-5, 1e-3, log=True)
n_layers = trial.suggest_int("n_layers", 3, 5)
# Evaluation follows. This can be scheduled arbitrarily by the user to other workers.
for epoch in range(epochs):
    loss = ...
    study.report(trial, loss, step=epoch)  # `Trial.report` also possible.
    if study.should_prune(trial):
        state = TrialState.PRUNED
        break

loss = ...  # Validation.

study.tell(trial, values=loss, state=state)
```

## TODO

- [x] Design
- [ ] ~Reporting (`Study.report`)~ Will be a separate PR
- [ ] ~Pruning (`Study.should_prune`)~ Will be a separate PR
- [x] Refactoring e.g. `_optimize.py`, delete usages of `Study._{ask,tell}`
- [x] Documentation
- [x] Tests
- [x] Example

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -20,6 +20,7 @@ from optuna._dataframe import _trials_dataframe
 from optuna._dataframe import pd
 from optuna._experimental import experimental
 from optuna._multi_objective import _get_pareto_front_trials
+from optuna._optimize import _check_and_convert_to_values
 from optuna._optimize import _optimize
 from optuna._study_direction import StudyDirection
 from optuna._study_summary import StudySummary  # NOQA
@@ -520,6 +521,7 @@ class Study(BaseStudy):
             raise ValueError(f"Cannot tell with state {state}.")
 
         if isinstance(trial, trial_module.Trial):
+            trial_number = trial.number
             trial_id = trial._trial_id
         elif isinstance(trial, int):
             trial_number = trial
@@ -536,20 +538,11 @@ class Study(BaseStudy):
             assert False, "Should not reach."
 
         if values is not None:
-            # TODO(hvy): Validate values.
-
-            if not isinstance(values, Sequence):
-                values = [values]
-
-            if len(values) != len(self.directions):
-                raise ValueError(
-                    "Trial {} could not be told because the number of values {} did not match "
-                    "the number of objectives {}.".format(
-                        trial.number if isinstance(trial, trial_module.Trial) else trial,
-                        len(values),
-                        len(self.directions),
-                    )
-                )
+            values, values_conversion_failure_message = _check_and_convert_to_values(
+                len(self.directions), values, trial_number
+            )
+            if values_conversion_failure_message is not None:
+                raise ValueError(values_conversion_failure_message)
 
         assert trial_id is not None
 
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -541,7 +541,10 @@ class Study(BaseStudy):
             values, values_conversion_failure_message = _check_and_convert_to_values(
                 len(self.directions), values, trial_number
             )
-            if values_conversion_failure_message is not None:
+            # When called from `Study.optimize` and `state` is pruned, the given `values` contains
+            # the intermediate value with the largest step so far. In this case, the value is
+            # allowed to be NaN and errors should not be raised.
+            if state != TrialState.PRUNED and values_conversion_failure_message is not None:
                 raise ValueError(values_conversion_failure_message)
 
         assert trial_id is not None
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -510,12 +510,11 @@ class Study(BaseStudy):
                 raise ValueError(
                     "No values were told. Values are required when state is TrialState.COMPLETE."
                 )
-        elif state == TrialState.PRUNED:
-            pass
-        elif state == TrialState.FAIL:
+        elif state in (TrialState.PRUNED, TrialState.FAIL):
             if values is not None:
                 raise ValueError(
-                    "Values were told. Values cannot be stored when state is TrialState.FAIL."
+                    "Values were told. Values cannot be specified when state is "
+                    "TrialState.PRUNED or TrialState.FAIL."
                 )
         else:
             raise ValueError(f"Cannot tell with state {state}.")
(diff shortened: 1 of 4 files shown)


hvy:
`Study.ask` and `Study.tell` done. Will start working on `Study.report` and `Study.should_prune`.

hvy:
Discussed with a couple of members including @g-votte and decided to continue with the remaining two methods in a separate PR to make reviewing smoother. Let me mark this PR ready for review. Cc. @not522

optuna/study.py
```diff
@@ -381,6 +382,186 @@ def objective(trial):
...
+        else:
+            assert False, "Should not reach."
+
+        if values is not None:
+            values, values_conversion_failure_message = _check_and_convert_to_values(
+                len(self.directions), values, trial_number
+            )
+            # When called from `Study.optimize` and `state` is pruned, the given `values` contains
+            # the intermediate value with the largest step so far. In this case, the value is
+            # allowed to be NaN and errors should not be raised.
```
hvy, line 546: This behavior leaves room for discussion but I leave it outside the scope of this PR trying to keep the behavior unchaned.

optuna/study.py
```diff
@@ -653,22 +834,14 @@ def _pop_waiting_trial_id(self) -> Optional[int]:
...
 
     def _ask(self) -> trial_module.Trial:
-        # Sync storage once at the beginning of the objective evaluation.
-        self._storage.read_trials_from_remote_storage(self._study_id)
-
-        trial_id = self._pop_waiting_trial_id()
-        if trial_id is None:
-            trial_id = self._storage.create_new_trial(self._study_id)
-        return trial_module.Trial(self, trial_id)
+        # TODO(hvy): Remove this method since `Study.ask` has been implemented.
```
hvy, line 837: I left these as TODOs since I know that we have users of those private previous ask and tell APIs, such as the Hydra Optuna sweeper. Cc. @toshihikoyanase . However, the logic is changed so it might still break. Maybe we can remove them at this point.

hvy:
Sorry for my last couple of commits, I think they are self explanatory. PTAL. Just assigned @sile as a second reviewer.

optuna/study.py
```diff
@@ -381,6 +382,192 @@ def objective(trial):
...
+        if not isinstance(trial, (trial_module.Trial, int)):
+            raise TypeError("Trial must be a trial object or trial number.")
+
+        if state == TrialState.COMPLETE:
+            if values is None:
+                raise ValueError(
+                    "No values were told. Values are required when state is TrialState.COMPLETE."
+                )
+        elif state == TrialState.PRUNED:
+            pass
```
sile, line 514: The existing logic (below) seems to set the last report value (or NaN) as the objective value if a trial is pruned.
https://github.com/optuna/optuna/blob/643c745b8aae70a2acc801819b870a685589f6b5/optuna/_optimize.py#L203-L207

Don't we need to force the behavior when users call the tell method with `state=TrialStat.PRUNED`?
(I know the existing logic is a controversial one though)

optuna/study.py
```diff
@@ -381,6 +382,192 @@ def objective(trial):
...
+            trial_number = trial.number
+            trial_id = trial._trial_id
+        elif isinstance(trial, int):
+            trial_number = trial
+            try:
+                trial_id = self._storage.get_trial_id_from_study_id_trial_number(
+                    self._study_id, trial_number
+                )
+            except NotImplementedError as e:
+                raise TypeError(
```
sile, line 533: It's just an idea but how about getting `trial_id` by using, for example, the following code in this case?
```python
warnings.warn("A message to warn a performance degradation")
trials = self.get_trials(deepcopy=False)

if len(trials) < trial_number:
    raise ...

trial_id = trials[trial_number]._trial_id
```

sile (commented):
Thank you for your PR. This feature is what I really wanted.
This PR already looks good, but I left some minor comments so please check them.

optuna/study.py
hvy: You're right. Fixed.

optuna/study.py
hvy: Ah, good idea. Added your suggestion in https://github.com/optuna/optuna/pull/2158/commits/d04c9a6f4eb30e3290ea056458ae680006ec6bd3.

hvy:
Thank you for the constructive comments. They should both have been addressed. PTAL.

sile (approved):
Thanks! LGTM!

hvy:
Merged with master since #2168 was merged.

not522 (commented):
Basically LGTM, just left minor comments.

optuna/study.py
```diff
@@ -657,22 +863,14 @@ def _pop_waiting_trial_id(self) -> Optional[int]:
         return None
 
     def _ask(self) -> trial_module.Trial:
```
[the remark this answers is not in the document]
hvy: Fixed.

optuna/study.py
```diff
@@ -657,22 +863,14 @@ def _pop_waiting_trial_id(self) -> Optional[int]:
...
-        self._storage.read_trials_from_remote_storage(self._study_id)
-
-        trial_id = self._pop_waiting_trial_id()
-        if trial_id is None:
-            trial_id = self._storage.create_new_trial(self._study_id)
-        return trial_module.Trial(self, trial_id)
+        # TODO(hvy): Remove this method since `Study.ask` has been implemented.
+        return self.ask()
 
     def _tell(
```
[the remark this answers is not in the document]
hvy: Fixed.

hvy:
PTAL.

Decorated `Study._{ask,tell}` as deprecated. They are private to start with so it's strictly speaking not needed, however, it's seemingly already used by by several.

not522 (approved):
LGTM!

PR #2159, opened 2020-12-29
Add description for tox to CONTRIBUTING.md

himkt:
- Related to #2058 and #2010

## Motivation
Add explanation for checking code using `tox`.

## Description of the changes
This PR update `CONTRIBUTING.md` to add an explanation of how to use `tox` in code checking.

@c-bata @crcrpar I'm sorry for the delay, please take a look.


crcrpar (commented):
I think we don't need to remove the information about `checks.sh`. Is there any reason/motivation behind the removal?

himkt:
You're right, thank you for pointing out.
75a4db0 restores the original text.

c-bata:
I think it's better to remove `checks.sh` from CONTRIBUTING.md as I commented at:

> I'm still not sure if we should still keeps checks.sh on repository root (In my opinion, it's enough just updates documentation about how to run auto-formatters).
> https://github.com/optuna/optuna/issues/2010#issuecomment-727580709

The file name of `checks.sh` does not imply the functionality for auto-formatting. It's hard to notice it unless the contributors will read `CONTRIBUTING.md`. So I cannot find the advantage to put `checks.sh` very much.

crcrpar:
I'm negative about removing `checks.sh` because it fixes style-related issues automatically with one option, which tox cannot provide.

Renaming and relocating the file sound reasonable to me though.

CONTRIBUTING.md
```diff
@@ -43,16 +43,26 @@ Coding style is checked with [flake8](http://flake8.pycqa.org) and [isort](https
 and additional conventions are described in the [Wiki](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions).
 Type hints, [PEP484](https://www.python.org/dev/peps/pep-0484/), are checked with [mypy](http://mypy-lang.org/).
 
-You can check the format, coding style, and type hint at the same time just by executing a script `checks.sh`.
-If your environment misses some dependencies such as black, flake8, isort or mypy, 
+You can check the format, coding style, and type hint at the same time just by executing a script `formats.sh`.
+If your environment misses some dependencies such as black, blackdoc, flake8, isort or mypy,
```
himkt, line 47: [minor change] Added a mention of `blackdoc`.

formats.sh
```diff
@@ -23,11 +25,11 @@ if [ ! ${#missing_dependencies[@]} -eq 0 ]; then
   pip install "${missing_dependencies[@]}"
 fi
 
-update=0
-while getopts "u" OPT
+update=1
+while getopts "n" OPT
```
himkt, line 29: Silently added a dry-run option. I'm not sure it is needed. If someone think it should be removed, I'll do.

himkt:
@c-bata @crcrpar Thank you for your suggestions!

I left `checks.sh`, as it turned out that the formatting feature of `checks.sh` is used by several developers.
(But I renamed it to `formats.sh`)

I'd like your reviews again. ⛩️

crcrpar (approved):

CONTRIBUTING.md
```diff
@@ -43,16 +43,26 @@ Coding style is checked with [flake8](http://flake8.pycqa.org) and [isort](https
...
-# Without any fixes.
-bash checks.sh
+# Install auto-formatters.
+$ pip install .[checking]
 
-# With fixes on the format by `black` and the import order by `isort`.
-bash checks.sh -u
+$ black .     # Fix the format by black.
+$ blackdoc .  # Fix the format of docstring by blackdoc.
+$ isort .     # Fix the import order by isort.
```
c-bata, line 65: Since we decided to leave the shell script, how about introducing `formats.sh` instead of these commands?

```suggestion
$ ./formats.sh 
```

c-bata (commented):
Thank you for your pull request. Basically LGTM!

CONTRIBUTING.md
himkt: You're right. Thank you.

c-bata (approved):
LGTM 👍

Issue #1980, opened 2020-10-30
Integrate multi-objective optimization module


HideakiImamura:
## Motivation
In the current Optuna's codebase, the `optuna.multi_objective` module is separated from other modules. However, to improve maintenance and ensure code consistency, Optuna's basic modules (e.g. study, trial, etc) should be designed with multi-objective optimization in mind. 

PR #1961 was submitted, but this work will be split into several small PRs. This issue tracks the integration of the multi-objective optimization module.

## TODOs
- [x] Multi-dimensional study directions and trial values: fix RDB schema #2030.
- [x] Multi-dimensional study directions and trial values: fix `study`, `frozen trial`, and `storages` #1994.
- [x] Pruning system: fix `trial.report` and `trial.should_prune` #2135.
- [x] Multi-objective samplers: migrate `NSGAIISampler` #2150.
- [x] Multi-objective samplers: migrate `MOTPESampler` #2167.
- [x] Visualization module: migrate `plot_pareto_front` #2110, #2122.
- [x] Visualization module: fix single-objective plot functions #2112, #2117, #2118, #2120, #2121, #2137.
- [x] Misc: fix multi-objective examples #2123.
- [x] Misc: fix other modules which assume the single-objective optimization #2136.
- [x] Misc: deprecate the previous multi-objective module #2124.

hvy:
Pinned this issue since I find myself coming back to it quite often and there was one slot left for the pins.

hvy:
@not522 is working on `Study.trials_dataframe` to support MO studies. I think we can close this issue as resolved after that.

hvy:
Let me close this issue as resolved now that https://github.com/optuna/optuna/pull/2181 addressing the above has been merged.

PR #2167, opened 2021-01-04
Migrate MOTPE sampler

not522:
## Motivation
Part of works for #1980. This PR migrates `MOTPESampler` to the standard samplers module.
## Description of the changes
- add `optuna.samplers.MOTPESampler` copied from `optuna.multi_objective.samplers.MOTPEMultiObjectiveSampler`
- add `optuna.samplers._motpe_hypervolume` copied from `optuna.multi_objective._hypervolume`
- remove "MultiObjective" from sampler's name
- use ordinal `Study` and `FrozenTrial` in `MOTPESampler`
- copy tests


HideakiImamura (commented):
Thanks for the PR! In my opinion, it is better to leave `optuna/multi_objective/_hypervolume` untouched and use it from `optuna/samplers/motpe.py`. It is the same for the test files.

As a follow-up, I think it's better to move and rename `optuna/_multi_objective.py` to `optuna/multi_objective/_utils.py`.

not522:
Thank you. You mean `optuna.multi_objective._hypervolume` will not be deprecated, is it right?

HideakiImamura:
Yes, it is because the `_hypervolume` module is private. I am saying the opposite of what I said before. Sorry for the confusion.

optuna/samplers/_motpe.py
```diff
@@ -0,0 +1,627 @@
...
+        study: optuna.study.Study,
+        trial: optuna.trial.FrozenTrial,
+        param_name: str,
+        param_distribution: BaseDistribution,
+    ) -> Any:
+        if len(study.directions) < 2:
+            raise ValueError(
+                "Number of objectives must be >= 2. "
+                "Please use optuna.samplers.TPESampler for single-objective optimization."
+            ) from None
```
hvy, line 172: Not introduced by this PR so a minor suggestion. I don't think we're in any error handling context so there's no exception to suppress?
```suggestion
            )
```

optuna/samplers/_motpe.py
```diff
@@ -0,0 +1,627 @@
...
+
+def default_gamma(x: int) -> int:
+    return int(np.floor(0.1 * x))
+
+
+def _default_weights_above(x: int) -> np.ndarray:
+    return np.ones(x)
+
+
+class MOTPESampler(TPESampler):
```
hvy, line 35: Would just like to verify but this sampler is stable and not experimental? C.f. https://github.com/optuna/optuna/commit/909324a09d8a45f07b78422284f04cc8c4ca2349

optuna/samplers/_motpe.py
```diff
@@ -0,0 +1,627 @@
...
+
+    This function collects observation pairs from the complete trials of the study.
+    Pruning is currently not supported.
+    The values for trials that don't contain the parameter named ``param_name`` are set to None.
+
+    Objective values are negated if their directions are maximization and all objectives are
+    treated as minimization in the MOTPE algorithm.
+    """
+
+    trials = study._storage.get_all_trials(study._study_id, deepcopy=False)
```
hvy, line 606: Can be a follow up but you can optimize/simplify this and the following code using the following.
```suggestion
    trials = study._storage.get_all_trials(study._study_id, deepcopy=False, states=(optuna.trial.TrialState.COMPLETE,))
```

optuna/samplers/_motpe.py
```diff
@@ -0,0 +1,627 @@
...
+                for i in range(1, num_variables + 1):
+                    x.append(trial.suggest_float(f"x{i}", 0.0, 2.0 * i))
+                return x
+
+
+            sampler = optuna.samplers.MOTPESampler(
+                n_startup_trials=n_startup_trials, n_ehvi_candidates=24, seed=seed
+            )
+            study = optuna.create_study(directions=["minimize"] * num_variables, sampler=sampler)
+            study.optimize(objective, n_trials=250)
```
hvy, line 111: Sounds like a of trials for a doctest example.
```suggestion
            study.optimize(objective, n_trials=n_startup_trials + 10)
```

optuna/samplers/_motpe.py
```diff
@@ -0,0 +1,627 @@
...
+        However, the current implementation only supports initialization with random sampling.
+
+    Example:
+
+        .. testcode::
+
+            import optuna
+
+            seed = 128
+            num_variables = 9
```
hvy, line 96: Related to previous comment about `n_trials`.
```suggestion
            num_variables = 2
```

hvy (changes requested):
Thanks a lot for the PR. Haven't gone through the unit tests yet but left early comments on everything else.

In addition to the comments, I'm wondering how come we don't put this sampler file inside the `_tpe` directory. It seems  related deriving the existing TPE sampler and reusing helpers inside `_tpe`.

optuna/samplers/_motpe.py
hvy: @HideakiImamura , @toshihikoyanase you probably know better than myself given #1530. Do you have any comments on this?

toshihikoyanase: I'm not pretty sure, but MoTPE may be able to be merged with TPE. Such a change may affect the public interface, and I think it is reasonable to keep the MoTPE experimental.

HideakiImamura: I agree with @toshihikoyanase. As a follow-up task, we can merge the multi-objective TPE with TPE after the v2.4.0 release.

HideakiImamura (approved):
Thanks for the update. LGTM.

optuna/samplers/_tpe/multi_objective_sampler.py
```diff
@@ -0,0 +1,628 @@
...
+
+def default_gamma(x: int) -> int:
+    return int(np.floor(0.1 * x))
+
+
+def _default_weights_above(x: int) -> np.ndarray:
+    return np.ones(x)
+
+
+@experimental("2.3.0")
```
hvy, line 36: This version should be bumped to match our upcoming release.
```suggestion
@experimental("2.4.0")
```

hvy (approved):
LGTM!

PR #2168, opened 2021-01-05
Storages to query trial IDs from numbers

hvy:
## Motivation

It is sometimes necessary in the internals (e.g. https://github.com/optuna/optuna/pull/2158) to be able to query for the trial ID given a trial number. The reason is that the latter is part of the public API while the ID is private and conversions are needed.

## Description of the changes

Introduces a method to the storage. This method return the trial ID given a study ID and a trial number.

### DIFF
--- a/tests/storages_tests/test_storages.py
+++ b/tests/storages_tests/test_storages.py
@@ -1084,10 +1084,8 @@ def test_get_trial_id_from_study_id_trial_number(storage_mode: str) -> None:
         # study. Create a second study within the same storage.
         study_id = storage.create_new_study()
 
-        storage.create_new_trial(study_id)
-        storage.create_new_trial(study_id)
         trial_id = storage.create_new_trial(study_id)
 
         assert trial_id == storage.get_trial_id_from_study_id_trial_number(
-            study_id, trial_number=2
+            study_id, trial_number=0
         )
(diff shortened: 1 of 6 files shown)


sile (approved):
LGTM 👍

tests/storages_tests/test_storages.py
```diff
@@ -1060,3 +1060,34 @@ def test_get_best_trial_for_multi_objective_optimization(storage_mode: str) -> N
...
+        # study. Create a second study within the same storage.
+        study_id = storage.create_new_study()
+
+        storage.create_new_trial(study_id)
+        storage.create_new_trial(study_id)
+        trial_id = storage.create_new_trial(study_id)
+
+        assert trial_id == storage.get_trial_id_from_study_id_trial_number(
+            study_id, trial_number=2
+        )
```
not522, line 1093: I feel it is unclear what this test checks.

not522 (commented):
Basically LGTM except one question.

tests/storages_tests/test_storages.py
hvy: It's testing the behavior of `get_trial_id_from_study_id_trial_number` when a storage contains more than one study. I.e., where the trial ID and the trial numbers are offset. I tried to explain it in the comments on line 1083, 1084. Can we improve this test function?

hvy: Simplified the test to avoid confusion. You're right it quite meaningless.

not522 (approved):
LGTM!

PR #2171, opened 2021-01-05
Fix unit test of median pruner.

keisuke-umezawa:
## Motivation

In the current master branch,  I think that there are some issues in `test_median_pruner_interval_steps` of `tests/pruners_tests/test_median.py`.

I explained the situation by using `"n_warmup_steps,interval_steps,report_steps,expected_prune_steps" = (1, 2, 1, [1, 3])` as an example.
* It only checks `i` from `0` to `2`, not including the last element of `expected_prune_steps`, 3, because `n_steps = max(expected_prune_steps)` and `i` was ranged by `range(n_steps)`. So, `expected_prune_steps` are not actually expected pruned steps here.
* Also, when someone sees this test case, they expected that it only passes only when pruning runs at at `i ==  1` and `i == 3`. However, the current test passes even when pruning does not run at `i ==  1` or `i == 3`, because it checks `i in expected_prune_steps`.

## Description of the changes

> * It only checks `i` from `0` to `2`, not including the last element of `expected_prune_steps` 3, because `n_steps = max(expected_prune_steps)` and `i` was ranged by `range(n_steps)`. So, `expected_prune_steps` are not actually expected pruned steps here.

To solve the above, I changed the code as https://github.com/optuna/optuna/pull/2171/files#diff-7945f3f629a92b7cd2958d1ee6b51816c7072f396447485a8db6704a6abbfd7aR125

> * Also, when someone sees this test case, they expected that it only passes only when pruning runs at at `i ==  1` and `i == 3`. However, the current test passes even when pruning does not run at `i ==  1` or `i == 3`, because it checks `i in expected_prune_steps`.

To solve the above, I changed the code as https://github.com/optuna/optuna/pull/2171/files#diff-7945f3f629a92b7cd2958d1ee6b51816c7072f396447485a8db6704a6abbfd7aR139.

### DIFF
--- a/tests/pruners_tests/test_median.py
+++ b/tests/pruners_tests/test_median.py
@@ -105,7 +105,9 @@ def test_median_pruner_n_warmup_steps() -> None:
 @pytest.mark.parametrize(
     "n_warmup_steps,interval_steps,report_steps,expected_prune_steps",
     [
-        (1, 2, 1, [1, 3]),
+        (0, 1, 1, [0, 1, 2, 3, 4, 5]),
+        (1, 1, 1, [1, 2, 3, 4, 5]),
+        (1, 2, 1, [1, 3, 5]),
         (0, 3, 10, list(range(29))),
         (2, 3, 10, list(range(10, 29))),
         (0, 10, 3, [0, 1, 2, 12, 13, 14, 21, 22, 23]),
@@ -120,16 +122,18 @@ def test_median_pruner_interval_steps(
     study = optuna.study.create_study()
 
     trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
-    n_steps = max(expected_prune_steps)
-    base_index = 0
-    for i in range(base_index, base_index + n_steps):
-        trial.report(base_index, i)
+    last_step = max(expected_prune_steps) + 1
+
+    for i in range(last_step):
+        trial.report(0, i)
     study._storage.set_trial_state(trial._trial_id, TrialState.COMPLETE)
 
     trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
-    for i in range(base_index, base_index + n_steps):
-        if (i - base_index) % report_steps == 0:
+
+    pruned = []
+    for i in range(last_step):
+        if i % report_steps == 0:
             trial.report(2, i)
-        assert pruner.prune(study=study, trial=study._storage.get_trial(trial._trial_id)) == (
-            i >= n_warmup_steps and i in expected_prune_steps
-        )
+        if pruner.prune(study=study, trial=study._storage.get_trial(trial._trial_id)):
+            pruned.append(i)
+    assert pruned == expected_prune_steps
--- a/tests/pruners_tests/test_median.py
+++ b/tests/pruners_tests/test_median.py
@@ -121,14 +121,14 @@ def test_median_pruner_interval_steps(
     pruner = optuna.pruners.MedianPruner(0, n_warmup_steps, interval_steps)
     study = optuna.study.create_study()
 
-    trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
+    trial = study.ask()
     last_step = max(expected_prune_steps) + 1
 
     for i in range(last_step):
         trial.report(0, i)
     study._storage.set_trial_state(trial._trial_id, TrialState.COMPLETE)
 
-    trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
+    trial = study.ask()
 
     pruned = []
     for i in range(last_step):


tests/pruners_tests/test_median.py
```diff
@@ -120,16 +122,18 @@ def test_median_pruner_interval_steps(
     study = optuna.study.create_study()
 
     trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
-    n_steps = max(expected_prune_steps)
-    base_index = 0
```
keisuke-umezawa, line 124: Removed unnecessary variable.

tests/pruners_tests/test_median.py
```diff
@@ -120,16 +122,18 @@ def test_median_pruner_interval_steps(
...
 
     trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
-    for i in range(base_index, base_index + n_steps):
-        if (i - base_index) % report_steps == 0:
+
+    pruned = []
+    for i in range(last_step):
+        if i % report_steps == 0:
             trial.report(2, i)
-        assert pruner.prune(study=study, trial=study._storage.get_trial(trial._trial_id)) == (
```
keisuke-umezawa, line 133: This test can pass wrongly even if `expected_prune_steps = [0, 1, 2, 3]` and `pruned = []`.

hvy:
Thanks for the PR @keisuke-umezawa. Curious what the context behind this change is. Could you perhaps fill in a brief description?

@crcrpar, would you mind jumping in as a reviewer?

hvy:
Assigned @not522 as well. Feel free to reassign it as needed.

tests/pruners_tests/test_median.py
crcrpar: How did you confirm the implementation is correct?

If some test case passes unexpectedly, then it'd be better to check the logic is correct or not.

keisuke-umezawa: Fortunately, it seems that the implementation is expected to be correct, because those test cases passed after this change. https://github.com/optuna/optuna/pull/2171/files#diff-7945f3f629a92b7cd2958d1ee6b51816c7072f396447485a8db6704a6abbfd7aR110-R113

Meanwhile, I will check the logic is correct or not as I need to understand it for another PR.

tests/pruners_tests/test_median.py
```diff
@@ -120,16 +122,18 @@ def test_median_pruner_interval_steps(
     study = optuna.study.create_study()
 
     trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
```
crcrpar, line 124: If I understand correctly, we can use `optuna.trial.create_trial` here, and I think we should use it instead as per the disclaimer of https://optuna.readthedocs.io/en/stable/reference/generated/optuna.trial.Trial.html#optuna.trial.Trial (quote follows).

> Note that the direct use of this constructor is not recommended. This object is seamlessly instantiated and passed to the objective function behind the optuna.study.Study.optimize() method; hence library users do not care about instantiation of this object.

This opinion is based on my take that we should not use some API if it is discouraged or deprecated.
So, what's your take?

tests/pruners_tests/test_median.py
```diff
@@ -120,16 +122,18 @@ def test_median_pruner_interval_steps(
...
-    base_index = 0
-    for i in range(base_index, base_index + n_steps):
-        trial.report(base_index, i)
+    last_step = max(expected_prune_steps) + 1
+
+    for i in range(last_step):
+        trial.report(0, i)
     study._storage.set_trial_state(trial._trial_id, TrialState.COMPLETE)
 
     trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
```
crcrpar, line 131: Repeat my comment above:

If I understand correctly, we can use `optuna.trial.create_trial` here, and I think we should use it instead as per the disclaimer of https://optuna.readthedocs.io/en/stable/reference/generated/optuna.trial.Trial.html#optuna.trial.Trial (quote follows).

> Note that the direct use of this constructor is not recommended. This object is seamlessly instantiated and passed to the objective function behind the optuna.study.Study.optimize() method; hence library users do not care about instantiation of this object.

This opinion is based on my take that we should not use some API if it is discouraged or deprecated.
So, what's your take?

tests/pruners_tests/test_median.py
keisuke-umezawa: I do not think we can replace the code by `optuna.trial.create_trial`, because it will create `FrozenTrial`, not `Trial`. Also, the above note is for library users, not for developers. So, imo, we can use `optuna.trial.Trial` here for unittest.

keisuke-umezawa: If we want to replace the code by something, it would be `trial = study._ask()`, although I do not think we should do it.

not522: `study.ask()` (introduced by #2158) is better, but I think replacement is optional.

not522 (approved):
LGTM!
It's just a minor comment; maybe the title of this PR should be "Fix" rather than "Improve."

tests/pruners_tests/test_median.py
keisuke-umezawa: Ok. I use `study.ask()` instead of using constructor.

crcrpar (approved):
Sorry for having kept you waiting.

PR #2174, opened 2021-01-06
Add constrained NSGA-II sampler

toshihikoyanase:
## Motivation
This PR extends NSGA-II sampler to handle constraints similarly to BoTorch sampler. The constrained NSGA-II is proposed by Deb et al. in the [original NSGA-II paper](https://ieeexplore.ieee.org/document/996017). It simply overwrites domination check as follows:

A trial `x` is said to constrained-dominate a trial `y`, if any of the following conditions is true:
1) Trial `x` is feasible and trial `y` is not.
2) Trial `x` and `y` are both infeasible, but solution `x` has a smaller overall constraint violation.
3) Trial `x` and `y` are feasible and trial `x` dominates trial `y`.

## Description of the changes
- Add `optuna.samplers.CNSGAIISampler`
- Make `optuna.samplers._nsga2._fast_non_dominated_sort` a  class method of `NSGAIISampler`
- Add a class method `NSGAIISampler._dominates`
- Add unit tests for `optuna.samplers.CNSGAIISampler`

### DIFF
--- a/docs/source/reference/samplers.rst
+++ b/docs/source/reference/samplers.rst
@@ -16,5 +16,6 @@ The :mod:`~optuna.samplers` module defines a base class for parameter sampling a
    optuna.samplers.CmaEsSampler
    optuna.samplers.PartialFixedSampler
    optuna.samplers.NSGAIISampler
+   optuna.samplers.CNSGAIISampler
    optuna.samplers.IntersectionSearchSpace
    optuna.samplers.intersection_search_space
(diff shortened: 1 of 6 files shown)


sile:
It's not a strong opinion but I think that it might be a possible option to add the constrained feature to the existing NSGA-II sampler instead of adding a new sampler.
What do you think of it?

toshihikoyanase:
@sile Thank you for your suggestion. I created a new PR #2175 according to your advice. I think the new API matches `BoTorchSampler` and it will be more user-friendly.

sile:
Great! Thank you for your swift response!

toshihikoyanase:
Let me close this PR since #2175 is more promising.

PR #2175, opened 2021-01-06
Add `constraints_func` argument to NSGA-II.

toshihikoyanase:
## Motivation
This is an alternative approach of #2174 suggested by @sile in https://github.com/optuna/optuna/pull/2174#issuecomment-755041088. Instead of adding a new sampler, this PR adds `constraints_func` to `NSGAIISampler` to handle constraints. The same algorithm is implemented as #2174.

This design follows `BoTorchSampler` and I think it provides better user experiences than #2174.

## Description of the changes
- Add `constraints_func` argument
- Make `_fast_non_dominated_sort` the method of `NSGAIISampler`
- Use `_constrained_dominates` if `constraints_func` is not `None`

## Example: solving multi-objective knapsack problem

See the following [notebook](https://colab.research.google.com/drive/1xwHwKRsyiM2DvG0ntlvneHGyYDhfopS_?usp=sharing).

## TODO

- [x] Add a simple comparison with NSGA-II

### DIFF
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -114,7 +114,7 @@ class NSGAIISampler(BaseSampler):
         if not (0.0 <= swapping_prob <= 1.0):
             raise ValueError("`swapping_prob` must be a float value within the range [0.0, 1.0].")
 
-        if constraints_func:
+        if constraints_func is not None:
             warnings.warn(
                 "The constraints_func option is an experimental feature."
                 " The interface can change in the future.",
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -353,7 +353,7 @@ class NSGAIISampler(BaseSampler):
                 con = self._constraints_func(trial)
                 if not isinstance(con, (tuple, list)):
                     warnings.warn(
-                        f"Constraints should be a sequence of floats but got {constraints}."
+                        f"Constraints should be a sequence of floats but got {type(con).__name__}."
                     )
                 constraints = tuple(con)
             except Exception:
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -424,15 +424,18 @@ def _constrained_dominates(
     if trial1.state != TrialState.COMPLETE:
         return True
 
-    if all(v <= 0 for v in constraints0) and all(v <= 0 for v in constraints1):
+    satisfy_constraints0 = all(v <= 0 for v in constraints0)
+    satisfy_constraints1 = all(v <= 0 for v in constraints1)
+
+    if satisfy_constraints0 and satisfy_constraints1:
         # Both trials satisfy the constraints.
         return _dominates(trial0, trial1, directions)
 
-    if all(v <= 0 for v in constraints0):
+    if satisfy_constraints0:
         # trial0 satisfies the constraints, but trial1 violates them.
         return True
 
-    if all(v <= 0 for v in constraints1):
+    if satisfy_constraints1:
         # trial1 satisfies the constraints, but trial0 violates them.
         return False
 
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -68,13 +68,15 @@ class NSGAIISampler(BaseSampler):
             :class:`~optuna.trial.FrozenTrial` and return the constraints. The return value must
             be a sequence of :obj:`float` s. A value strictly larger than 0 means that a
             constraints is violated. A value equal to or smaller than 0 is considered feasible.
+            If constraints_func returns more than one value for a trial, that trial is considered
+            feasible if and only if all values are equal to 0 or smaller.
 
             The constraints are handled by the constrained domination. A trial x is said to
             constrained-dominate a trial y, if any of the following conditions is true:
 
-            1. Trial x satisfies all constraints and trial y does not.
-            2. Trial x and y both violate constraints, but trial x has a smaller overall violation.
-            3. Trial x and y satisfies all constraints, and trial x dominates trial y.
+            1. Trial x is feasible and trial y is not.
+            2. Trial x and y are both infeasible, but trial x has a smaller overall violation.
+            3. Trial x and y are feasible and trial x dominates trial y.
 
             .. note::
                 Added in v2.5.0 as an experimental feature. The interface may change in newer
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -1,5 +1,4 @@
 from collections import defaultdict
-import copy
 import hashlib
 import itertools
 from typing import Any
@@ -327,11 +326,8 @@ class NSGAIISampler(BaseSampler):
     ) -> None:
         if self._constraints_func is not None:
             constraints = None
-            _trial = copy.copy(trial)
-            _trial.state = state
-            _trial.values = values
             try:
-                con = self._constraints_func(_trial)
+                con = self._constraints_func(trial)
                 if not isinstance(con, (tuple, list)):
                     warnings.warn(
                         f"Constraints should be a sequence of floats but got {constraints}."
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -72,9 +72,9 @@ class NSGAIISampler(BaseSampler):
             The constraints are handled by the constrained domination. A trial x is said to
             constrained-dominate a trial y, if any of the following conditions is true:
 
-            1. Trial x is feasible and trial y is not.
-            2. Trial x and y are both infeasible, but trial x has a smaller overall violation.
-            3. Trial x and y are feasible and trial x dominates trial y.
+            1. Trial x satisfies all constraints and trial y does not.
+            2. Trial x and y both violate constraints, but trial x has a smaller overall violation.
+            3. Trial x and y satisfies all constraints, and trial x dominates trial y.
 
             .. note::
                 Added in v2.5.0 as an experimental feature. The interface may change in newer
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -72,10 +72,9 @@ class NSGAIISampler(BaseSampler):
             The constraints are handled by the constrained domination. A trial x is said to
             constrained-dominate a trial y, if any of the following conditions is true:
 
-            1) Trial x is feasible and trial y is not.
-            2) Trial x and y are both infeasible, but solution x has a smaller overall constraint
-            violation.
-            3) Trial x and y are feasible and trial x dominates trial y.
+            1. Trial x is feasible and trial y is not.
+            2. Trial x and y are both infeasible, but trial x has a smaller overall violation.
+            3. Trial x and y are feasible and trial x dominates trial y.
 
             .. note::
                 Added in v2.5.0 as an experimental feature. The interface may change in newer
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -1,13 +1,17 @@
 from collections import defaultdict
+import copy
 import hashlib
 import itertools
 from typing import Any
+from typing import Callable
 from typing import cast
 from typing import DefaultDict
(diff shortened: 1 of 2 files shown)


hvy (commented):
Great PR, seems like a natural extension to NSGA-II.

Have only taken a quick glance but looks good in general. Left some early comments if you could take a look.

optuna/samplers/_nsga2.py
```diff
@@ -327,3 +376,49 @@ def _crowding_distance_sort(population: List[FrozenTrial]) -> None:
...
+    trial0: FrozenTrial, trial1: FrozenTrial, directions: Sequence[StudyDirection]
+) -> bool:
+    """Checks constrained-domination.
+
+    A trial x is said to constrained-dominate a trial y, if any of the following conditions is
+    true:
+    1) Trial x is feasible and trial y is not.
+    2) Trial x and y are both infeasible, but solution x has a smaller overall constraint
+    violation.
+    3) Trial x and y are feasible and trial x dominates trial y.
```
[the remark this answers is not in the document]
toshihikoyanase: That makes sense. I moved this description to the docstring for the `constraints_func` argument in the commit https://github.com/optuna/optuna/pull/2175/commits/eda1fd890680d175c8531280251a09d3d811f0e2. Thank you!

optuna/samplers/_nsga2.py
```diff
@@ -327,3 +376,49 @@ def _crowding_distance_sort(population: List[FrozenTrial]) -> None:
...
+    if len(constraints1) != len(constraints1):
+        raise ValueError("Trials with different numbers of constraints cannot be compared.")
+
+    if trial0.state != TrialState.COMPLETE:
+        return False
+
+    if trial1.state != TrialState.COMPLETE:
+        return True
+
+    if all(v <= 0 for v in constraints0) and all(v <= 0 for v in constraints1):
```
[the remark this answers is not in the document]
toshihikoyanase: Sounds reasonable. I extracted the violation checks to the local variable in the commit https://github.com/optuna/optuna/pull/2175/commits/3d77514658eea0a5345544fde016b997ca123592.

optuna/samplers/_nsga2.py
```diff
@@ -65,6 +78,7 @@ def __init__(
         crossover_prob: float = 0.9,
         swapping_prob: float = 0.5,
         seed: Optional[int] = None,
+        constraints_func: Optional[Callable[[FrozenTrial], Sequence[float]]] = None,
```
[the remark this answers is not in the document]
toshihikoyanase: Thanks! I added the experimental warnings in the commit https://github.com/optuna/optuna/pull/2175/commits/eda1fd890680d175c8531280251a09d3d811f0e2.

optuna/samplers/_nsga2.py
```diff
@@ -327,3 +393,52 @@ def _crowding_distance_sort(population: List[FrozenTrial]) -> None:
...
+    3) Trial x and y are feasible and trial x dominates trial y.
+    """
+
+    constraints0 = trial0.system_attrs[_CONSTRAINTS_KEY]
+    constraints1 = trial1.system_attrs[_CONSTRAINTS_KEY]
+
+    assert isinstance(constraints0, (list, tuple))
+    assert isinstance(constraints1, (list, tuple))
+
+    if len(constraints1) != len(constraints1):
```
hvy, line 417: Ops, seems like a typo.
```suggestion
    if len(constraints0) != len(constraints1):
```

toshihikoyanase: Thanks! I fixed it.

optuna/samplers/_nsga2.py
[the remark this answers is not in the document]
hvy: Sounds good. What I actually meant though was to describe when a trial is _feasible_. I.e. the following condition and that it's not `any`, although it might be clear. Not sure, what do you think?

```python
all(v <= 0 for v in constraints)
```

optuna/samplers/_nsga2.py
```diff
@@ -85,12 +112,20 @@ def __init__(
         if not (0.0 <= swapping_prob <= 1.0):
             raise ValueError("`swapping_prob` must be a float value within the range [0.0, 1.0].")
 
+        if constraints_func:
```
hvy, line 115: A nit, optional. We're using clear when checking for `None` to e.g. avoid evaluating arbitrary boolean castable objects.
```suggestion
        if constraints_func is not None:
```

optuna/samplers/_nsga2.py
```diff
@@ -253,51 +288,82 @@ def _select_parent(self, study: Study, population: List[FrozenTrial]) -> FrozenT
...
+        state: TrialState,
+        values: Optional[Sequence[float]],
+    ) -> None:
+        if self._constraints_func is not None:
+            constraints = None
+            try:
+                con = self._constraints_func(trial)
+                if not isinstance(con, (tuple, list)):
+                    warnings.warn(
+                        f"Constraints should be a sequence of floats but got {constraints}."
```
hvy, line 354: Sorry this logic probably comes from the `BoTorchSampler` which I wrote but the warning message is somewhat misleading. Could be a follow up but I just created https://github.com/optuna/optuna/pull/2197 for reference

hvy (changes requested):
Left some small comments. It LGTM otherwise and besides this ongoing discussion https://github.com/optuna/optuna/pull/2175/files#r556251445.

optuna/samplers/_nsga2.py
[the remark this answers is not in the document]
hvy: That is, something like 

> If `constraints_func` returns more than one value for a trial, that trial is considered feasible if and only if all values are equal to 0 or smaller.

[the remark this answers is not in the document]
toshihikoyanase: I'm sorry, but I completely misunderstood your comment. I added the suggested sentence to the definition of the feasibility of a single value in the commit https://github.com/optuna/optuna/pull/2175/commits/24734d72b401eea450c128bd1f6359fb5d1b0876.

optuna/samplers/_nsga2.py
toshihikoyanase: Ah, `constraints` will be `None` here, and we should show the type of `con`. I applied your suggestion in the commit https://github.com/optuna/optuna/pull/2175/commits/18a0c9e9074b19d7f1d02b71a3031502b4420037.

optuna/samplers/_nsga2.py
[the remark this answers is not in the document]
toshihikoyanase: When I read the description again, and I feel the definition of the constrained domination may be too specific. I'm afraid that it can be noise for the users of the original NSGA-II. What do you think?

image

optuna/samplers/_nsga2.py
hvy: Thanks for your quick move!

optuna/samplers/_nsga2.py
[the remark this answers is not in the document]
hvy: I was first fine with having it documented purely for maintainers, but after seeing your change, I now think that it won't hurts to have it documented publicly. It's well defined and informative to some. It's not a strong opinion so I'm fine either way.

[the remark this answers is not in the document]
toshihikoyanase: > I now think that it won't hurt to have it documented publicly. It's well defined and informative to some.

Thank you for your reply! Then, I'd like to keep the definition.

optuna/samplers/_nsga2.py
```diff
@@ -327,3 +395,52 @@ def _crowding_distance_sort(population: List[FrozenTrial]) -> None:
...
+    A trial x is said to constrained-dominate a trial y, if any of the following conditions is
+    true:
+    1) Trial x is feasible and trial y is not.
+    2) Trial x and y are both infeasible, but solution x has a smaller overall constraint
+    violation.
+    3) Trial x and y are feasible and trial x dominates trial y.
+    """
+
+    constraints0 = trial0.system_attrs[_CONSTRAINTS_KEY]
+    constraints1 = trial1.system_attrs[_CONSTRAINTS_KEY]
```
hvy, line 414: Could this dictionary lookup ever fail, for instance when a user would resume a once finished study? If so, we could raise a more user friendly error here. Could be a follow up.

optuna/samplers/_nsga2.py
```diff
@@ -327,3 +395,52 @@ def _crowding_distance_sort(population: List[FrozenTrial]) -> None:
...
+        return True
+
+    if satisfy_constraints1:
+        # trial1 satisfies the constraints, but trial0 violates them.
+        return False
+
+    # Both trials violate the constraints.
+    violation0 = sum(v for v in constraints0 if v > 0)
+    violation1 = sum(v for v in constraints1 if v > 0)
+    return violation0 < violation1
```
hvy, line 446: Would this be more aligned with how [`_dominates` is implemented](https://github.com/optuna/optuna/blob/master/optuna/_multi_objective.py#L58) ?
```suggestion
    return violation0 <= violation1
```

hvy (commented):
Sorry left two last comments/questions.

Also, just curious why you made some of the free functions methods. I'd personally keep them free (easier to comprehend, assuming they won't likely be overridden) but it's not a strong opinion, just want to make sure I understand why.

optuna/samplers/_nsga2.py
toshihikoyanase: Thank you for your suggestion, but I guess the current code follows the original `_dominates` function because both of them return `False` if x and y are identical.

https://github.com/optuna/optuna/blob/baf56206a357ffe0b1adb95289a6f4d3034c3ad3/optuna/_multi_objective.py#L55-L58

In addition,  the [original paper](https://www.iitk.ac.in/kangal/Deb_NSGA-II.pdf) excludes the case of `violation0 == violation1`:

> 2) Solutions and are both infeasible, but solution has a smaller overall constraint violation.

optuna/samplers/_nsga2.py
toshihikoyanase: Thank you for pointing it out. I think we can exclude the trials without constraint values from the population via the elitist selection if we add the following rule.

A trial without constraint values will be dominated by trials with constraint values:
1) Return false if neither `trial0` nor `trial1` does not have constraint values
2) Return false if `trial1` has constraint values, but `trial0` does
3) Return true if `trial0` has constraint values, but `trial1` does

This functionality can be reviewed independently, and I'll create a follow-up PR.

tests/samplers_tests/test_nsga2.py
```diff
@@ -190,8 +405,15 @@ def test_reseed_rng() -> None:
...
+def test_constraints_func_experimental_warning() -> None:
+    with pytest.warns(optuna.exceptions.ExperimentalWarning):
+        NSGAIISampler(constraints_func=lambda _: [0])
+
+
 # TODO(ohta): Consider to move this utility function to `optuna.testing` module.
-def _create_frozen_trial(number: int, values: List[float]) -> optuna.trial.FrozenTrial:
+def _create_frozen_trial(
+    number: int, values: List[float], constraints: Optional[List[float]] = None
+) -> optuna.trial.FrozenTrial:
```
HideakiImamura, line 416: This can be done in the follow-up, but how about re-defining `_create_frozen_trial` by using `optuna.trial.create_trial` since the direct instantiation of `FrozenTrial` is not a recommended usage?

HideakiImamura (commented):
Thanks for the great extension of NSGAII. Almost LGTM. I have a minor comment on the test.

optuna/samplers/_nsga2.py
hvy: Thanks, I must have missed the early return.

hvy (approved):
LGTM!

tests/samplers_tests/test_nsga2.py
toshihikoyanase: That makes sense. I update the `_create_frozen_trial` according to your suggestion in https://github.com/optuna/optuna/pull/2175/commits/1040b8444c9df998d74a1373723c1082e86271e0. PTAL.

HideakiImamura (approved):

PR #2176, opened 2021-01-06
Ask and Tell API (`Study.report`)

hvy:
## Motivation

See #2158.

## Description of the changes

~Depends on #2158.~ (Edit: merged)

Introduces `Study.report`. The logic of `Trial.report` is moved to the new method and `Trial.report` just indirects to `Study.report`.

## TODO

- [x] Implement
- [x] Document


hvy:
Ready for review.

hvy:
Merged with the latest master which includes `Study.{ask,tell}` and fixed some conflicts. PTAL.

hvy:
@sile, could you take a look at this one as well? It's a part of the ask-and-tell interface.

HideakiImamura (approved):
Looks perfect!

HideakiImamura:
Oh... I accidentally merged this PR with my single approval. Sorry.

@sile Could you leave comments on this PR if you have. I'd like to reflect on your reviews as a follow-up PR.

sile:
@hvy 

This PR itself looks good, but let me ask a question to just clarify the motivation of this PR.

At a glance, it seemed a bit redundant because `Trial` already has the `report` method.
I guess that the main motivation to introduce `Study.report` is allowing users to report intermediate values via trial numbers (not trial objects). Is it right?
If so, have you a plan to add other methods of `Trial` (e.g., `set_user_attr`, `suggest_xxx`) to `Study` for the same aim in the future?
(I'm a bit worried about it would scatter the `Study` API).

hvy:
I think that's valid concern. Not supporting pruning and reporting with trial numbers is also a viable approach (at least for now. Elaborated further down). I'd like to hear more comments from the user side here, when the `Trial` is at disposal and when it's not.

> I guess that the main motivation to introduce Study.report is allowing users to report intermediate values via trial numbers (not trial objects). Is it right?

That's correct. 

> If so, have you a plan to add other methods of Trial (e.g., set_user_attr, suggest_xxx) to Study for the same aim in the future?

As for additional methods, perhaps at most setters and getter for the user attributes? Suggest methods would require more work since the `Trial` is very stateful and we'd have to clearly define what suggest returns when called at different times (e.g. when there's a time delay between different hyperparameters or a delay between trial construction and suggestion). However, if this would be well defined, then we probably don't need these methods in the `Study` in the first place. We could allow something like `trial: Trial = study.ask(trial_number)` instead. But again, user feedback would be very appreciated here to clarify what use cases look like.

sile:
Thanks for the detailed response! 

> We could allow something like `trial: Trial = study.ask(trial_number)` instead.

Personally, I prefer this approach.

Issue #2179, opened 2021-01-06
Cannot run multi-objective optimization after upgrading storage from v2.3.0 to master.


toshihikoyanase:
## Expected behavior

When I upgrade the Optuna storage in MySQL 8.0 from v2.3.0 to the current master, I couldn't execute the multi-objective optimization due to the integrity error of `trial_id`. The single objective optimization can be executed even after the upgrade, so I think the unique constraint of `trial_id` is not set appropriately. 

Optimization should be executed after upgrading the storage.

## Environment

- Optuna version: v2.3.0 to master
- Python version: 3.7.7
- OS: macOS 10.15.7
- (Optional) Other libraries and their versions:
  - MySQL==8.0.22
  - PyMySQL==0.9.3

## Error messages, stack traces, or logs

```console
sqlalchemy.exc.IntegrityError: (pymysql.err.IntegrityError) (1062, "Duplicate entry '4' for key 'trial_values.trial_id'")
[SQL: INSERT INTO trial_values (trial_id, objective, value) VALUES (%(trial_id)s, %(objective)s, %(value)s)]
[parameters: {'trial_id': 4, 'objective': 1, 'value': 0.18631845259151747}]
```

## Steps to reproduce

1. Launch MySQL server
2. Run optimization with optuna==v2.3.0
3. Upgrade storage using optuna@master
4. Run multi-objective optimization with optuna@master

## Reproducible examples (optional)

1. Launch MySQL server
```console
$ docker run --name mysql -e MYSQL_ROOT_PASSWORD=test -p 3306:3306 -p 33060:33060 -d mysql:8.0.22
$ docker run --network host -it --rm mysql:8.0.22  mysql -h 127.0.0.1 -uroot -ptest -e "create database migration_optuna;"
```

2. Run optimization with optuna==v2.3.0
```python
import optuna

def objective(trial):
    return trial.suggest_float("x", 0, 1)

study = optuna.create_study(study_name="single", storage="mysql+pymysql://root:test@localhost:3306/migration_optuna", load_if_exists=True)
study.optimize(objective, n_trials=1)
```

3. Upgrade storage using optuna@master
```console
$ optuna storage upgrade --storage mysql+pymysql://root:test@localhost:3306/migration_optuna
```

4. Run multi-objective optimization with optuna@master
```python
import optuna

def objective(trial):
    return trial.suggest_float("x", 0, 1), trial.suggest_float("y", 0, 1)

[...]

PR #2180, opened 2021-01-06
Drop unnecessary constraint from upgraded `trial_values` table

toshihikoyanase:
## Motivation

Addresses #2179.

## Description of the changes

This PR removes the `trial_id` from the unique constraint of the `trial_values` table. It is not unique since v2.4.0 because of the support of multi-objective optimization. A pair of `trial_id` and `objective` is the unique key of `trial_values` now.

## TODO

- [x] Manual test with MySQL
- [x] Manual test with PostgreSQL
- [x] Manual test with SQLite

Test for SQLite

<details>
<summary>Python scripts for tests</summary>

single.py
```python
import sys
import optuna

def objective(trial):
    return trial.suggest_float("x", 0, 1)

study = optuna.create_study(study_name=sys.argv[1], storage=sys.argv[2], load_if_exists=True)
```

multi.py
```python
import sys
import optuna

def objective(trial):
    return trial.suggest_float("x", 0, 1), trial.suggest_float("y", 0, 1)

study = optuna.create_study(directions=["maximize", "minimize"], study_name=sys.argv[1], storage=sys.argv[2], load_if_exists=True)
```

</details>

Test for MySQL

```console
docker run --name mysql -e MYSQL_ROOT_PASSWORD=test -p 3306:3306 -p 33060:33060 -d mysql:8.0.22
docker run --network host -it --rm mysql:8.0.22  mysql -h 127.0.0.1 -uroot -ptest -e "create database test_optuna;"

pip uninstall optuna
pip install optuna
python single.py single mysql+pymysql://root:test@localhost:3306/test_optuna
pip install git+https://github.com/toshihikoyanase/optuna.git@fix-rdb-upgrade-v2.4.0
optuna storage upgrade --storage mysql+pymysql://root:test@localhost:3306/test_optuna
python multi.py multi mysql+pymysql://root:test@localhost:3306/test_optuna
```

Test for MySQL

```console
docker run -it --rm --name postgres-test -e POSTGRES_PASSWORD=test -p 15432:5432 -d postgres

pip uninstall optuna
pip install optuna
python single.py single postgres+psycopg2://postgres:test@localhost:15432/postgres
pip install git+https://github.com/toshihikoyanase/optuna.git@fix-rdb-upgrade-v2.4.0
optuna storage upgrade --storage postgres+psycopg2://postgres:test@localhost:15432/postgres
python multi.py multi postgres+psycopg2://postgres:test@localhost:15432/postgres
```

Test for SQLite

```console
pip uninstall optuna
pip install optuna
python single.py single sqlite:///example.db
pip install git+https://github.com/toshihikoyanase/optuna.git@fix-rdb-upgrade-v2.4.0
optuna storage upgrade --storage sqlite:///example.db
python multi.py multi sqlite:///example.db
```

### DIFF
--- a/optuna/storages/_rdb/alembic/versions/v2.4.0.a.py
+++ b/optuna/storages/_rdb/alembic/versions/v2.4.0.a.py
@@ -175,6 +175,7 @@ def upgrade():
                 batch_op.drop_constraint(c["name"], type_="unique")
             break
 
+
 # TODO(imamura): Implement downgrade
 def downgrade():
     pass
--- a/optuna/storages/_rdb/alembic/versions/v2.4.0.a.py
+++ b/optuna/storages/_rdb/alembic/versions/v2.4.0.a.py
@@ -144,7 +144,6 @@ def upgrade():
             batch_op.create_unique_constraint(
                 "uq_trial_values_trial_id_objective", ["trial_id", "objective"]
             )
-            batch_op.drop_constraint("trial_id", type_="unique")
 
         trials_records = session.query(TrialModel).all()
         objects = [
@@ -169,6 +168,12 @@ def upgrade():
     with op.batch_alter_table("trials", schema=None) as batch_op:
         batch_op.drop_column("value")
 
+    for c in Inspector.from_engine(bind).get_unique_constraints("trial_values"):
+        # MySQL changes the uniq constraint of (trial_id, step) to that of trial_id.
+        if c["column_names"] == ["trial_id"]:
+            with op.batch_alter_table("trial_values", schema=None) as batch_op:
+                batch_op.drop_constraint(c["name"], type_="unique")
+            break
 
 # TODO(imamura): Implement downgrade
 def downgrade():
--- a/optuna/storages/_rdb/alembic/versions/v2.4.0.a.py
+++ b/optuna/storages/_rdb/alembic/versions/v2.4.0.a.py
@@ -77,7 +77,8 @@ class TrialIntermediateValueModel(BaseModel):
 
 def upgrade():
     bind = op.get_bind()
-    tables = Inspector.from_engine(bind).get_table_names()
+    inspector = Inspector.from_engine(bind)
+    tables = inspector.get_table_names()
 
     if "study_directions" not in tables:
         op.create_table(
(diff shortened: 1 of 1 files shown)


toshihikoyanase:
When `step` is removed from `trial_values`, SQLite and PostgreSQL remove the unique constraint of (`trial_id`, `step`). MySQL, however, changes it to the unique constraint of `trial_id`. So, I add the existence of such an inappropriate constraint before removing it in the commit https://github.com/optuna/optuna/pull/2180/commits/0b9dd8e4e8cf142f5bd774705085dbcf0574caf4.

https://github.com/optuna/optuna/blob/add10e9fc5d734915c1cbe2d872db25c8297837c/optuna/storages/_rdb/alembic/versions/v2.4.0.a.py#L165-L166

hvy:
Thanks, and great description. I'm currently going through your changes and the reproduction steps.

I'm pretty sure this is mostly comes from how alembic does all the magic rather but indeed it seems like various databases behave differently. Pasted some perhaps related documentations (though I haven't enough time ensuring this is what's actually happening under the hood).

> Attempts to modify, rename, or drop a column that is used in a CHECK constraint, unless that constraint is also dropped in the same statement. Exception: If a CHECK constraint refers only to a single column, dropping the column automatically drops the constraint.

https://dev.mysql.com/doc/refman/8.0/en/alter-table.html

while PostgreSQL says

> This form drops a column from a table. Indexes and table constraints involving the column will be automatically dropped as well. 

https://www.postgresql.org/docs/9.1/sql-altertable.html

As for SQLite, they don't support dropping column in the first place.

hvy (approved):
LGTM!

(We could maybe stick a single `Inspector` object instead of creating a second one?, but it's really a minor and I also verified your solution to be working.)

toshihikoyanase:
@hvy
> We could maybe stick to a single Inspector object instead of creating a second one?

That makes sense. I applied your suggestion in https://github.com/optuna/optuna/pull/2180/commits/6a324196e933ce5041383a0ee209544575d98d73. PTAL.

hvy (approved):
Oh, thanks. Double checked the migration and LGTM!

Issue #2177, opened 2021-01-06
Ask and Tell API Request


hvy:
## Motivation

Please refer to the description in https://github.com/optuna/optuna/pull/2158. 

## Description

This is an issue to keep track of related PRs to Optuna's ask and tell API.

- [x] `Study.ask`, `Study.tell` #2158
- [x] ~`Study.report`, #2176~ Edit: Closed. See #2176.
- [x] ~`Study.should_prune` #2178~  Edit: Closed. See #2176.
- [x] ~Consider allowing `FrozenTrial`~ Edit: Closed. See #2176.
- [ ] Allow reporting and pruning by restoring Trials. #2213
- [x] Tutorial how to use Ask and tell
- [ ] Allow logging similar to `Study.optimize`

## Example

```python
# NOTE: Where `Trial`s are passed as arguments, trial numbers can be passed as well.

# Example 1: Simple case without pruning and no error handling.
trial = study.ask()
values = objective(trial)
study.tell(trial, values)

# Example 2: With error handling.
trial = study.ask()

values = None
try:
    values = objective(trial)
except TrialPruned:
    state = TrialState.PRUNED
except Exception:
    state = TrialState.FAIL
else:
    state = TrialState.COMPLETE
finally:
    study.tell(trial, values=values, state=state)

# Example 3: No `func` objective function.
trial = study.ask()

loss = None
state = TrialState.COMPLETE

lr = trial.suggest_float("lr", 1e-5, 1e-3, log=True)
n_layers = trial.suggest_int("n_layers", 3, 5)
# Evaluation follows. This can be scheduled arbitrarily by the user to other workers.
for epoch in range(epochs):
    loss = ...
    study.report(trial, loss, step=epoch)  # `Trial.report` also possible.
    if study.should_prune(trial):
        state = TrialState.PRUNED
        break

loss = ...  # Validation.

study.tell(trial, values=loss, state=state)
```

Shawn513:
HI, I would like to ask: 
If i've already have some observation points and would like to use these data to tell the study. 
So first I will need to 
Trail = study.ask() 
then how can I set the trail's parameters to the already known data points  ? Here I noticed that there's trial.suggest_int("n_layers", 3, 5),  but this will be generated a new data point , is there any func could directly feed the data point like in Skopt?

hvy:
[...]

PR #2196, opened 2021-01-13
Fix the wrong way to set `datetime_start`

### LINKED ISSUE #2036: Start time is wrong when calculating trials added by `enqueue_trial`.
I have a study with 30 trials. I read this study with all trials and use `enqueue_trial` to create a new study and add the best 10 trials. Then I start optuna to calculate the results on all 10 trials of the new study.

I observe that the logged "datetime_start" of all 10 trials is the same (which is wrong since the runs need a significant amount of time). After the 10 enqued runs the time is logged correctly for trial 11 and so on.

I am using optuna 2.3.0 from pip.

chenghuzi:
Fixes #2036.

## Motivation

The current mechanism to set the `datetime_start` of a given trial is to add a time value when creating that trial. However, this could be wrong if the created trial is not the one to be executed right after its creation. The `datetime_start` of these waiting trials won't be updated before and after their delayed execution. For specific example, see [#2306](https://github.com/optuna/optuna/issues/2036).

## Description of the changes

The mechanism to initialize `datetime_start` is changed. Now we set the time of trial start when its state changes to `RUNNING`. This ensures the `datetime_start ` to be the moment right before execution. This is implemented in `set_trial_state`.

Xnip2021-01-13_00-34-10
Trial 11-20 were appended to the study through `enqueue_trial ` after trial 10 but were executed several minutes later. After change the `datetime_start ` becomes the 2s duration set by the objective function:

```python
def objective(trial):
    time.sleep(2)
    return x**2
```

### DIFF
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -414,7 +414,7 @@ def _constrained_dominates(
     assert isinstance(constraints0, (list, tuple))
     assert isinstance(constraints1, (list, tuple))
 
-    if len(constraints1) != len(constraints1):
+    if len(constraints0) != len(constraints1):
         raise ValueError("Trials with different numbers of constraints cannot be compared.")
 
     if trial0.state != TrialState.COMPLETE:


HideakiImamura:
@c-bata @hvy Could you review this PR if you have time?

@PhilipMay @nzw0301 Please feel free to leave comments on this PR since you have discussed this problem in the original issue.

optuna/_optimize.py
```diff
@@ -193,6 +193,8 @@ def _run_trial(
     func_err_fail_exc_info: Optional[Any] = None
     # Set to a string if `func` returns correctly but the return value violates assumptions.
     values_conversion_failure_message: Optional[str] = None
+    # update the trial state to running and **change the datetime_start**
+    study._storage.set_trial_state(trial._trial_id, TrialState.RUNNING)
```
hvy, line 197: Might be better to do this inside `Study.ask`. With this logic, the timestamp will not be updated when the users uses `Study.ask` instead of `Study.optimize`. You might have to update the base branch to the latest master since this method was introduced very recently.

hvy (commented):
Thanks for tackling the long standing issue. Seems like a compact and nice solution. I have one high level question though (from the issue) and a change request.

- Should we not nullify the timestamp when enqueued?

optuna/_optimize.py
hvy: You could perhaps catch up on the ask and tell API from this issue https://github.com/optuna/optuna/issues/2177, but basically, it's an alternative to `Study.optimize` where the user directly creates and manages trial lifetimes.

chenghuzi:
> Thanks for tackling the long standing issue. Seems like a compact and nice solution. I have one high level question though (from the issue) and a change request.
> 
> * Should we not nullify the timestamp when enqueued?

For consistency I think it will be good: a trial should not be marked as started anytime before its execution. But I'm not sure if this will lead to unexpected behavior in other tests.

hvy:
Ops, seems like you got unrelated commits in this PR. Could you fix the history?

hvy:
It would unfortunately be a breaking change but my understanding is that the previous behavior wasn't clearly thought out. Some concrete breakages would include the constructions of the study summaries for each storage implementation, and probably `FrozenTrial._validate`. I think we can see this as a fix.

chenghuzi:
> It would unfortunately be a breaking change but my understanding is that the previous behavior wasn't clearly thought out. Some (internal) concrete breakages that would need to taken care of would include the constructions of the study summaries for each storage implementation, and probably `FrozenTrial._validate`. There might be others but I think we can see those as fixes.



> Ops, seems like you got unrelated commits in this PR. Could you fix the history?

Uh I'm not sure about this, you mean we can delete the previous commits in which I put `datetime_start` not in the `ask` api and directly use the last one?

hvy:
I meant to exclude the commits that were not authored by yourself. This PR currently comes with 66 commits of which many are authored by myself or others.

toshihikoyanase:
Sorry if I miss the point, but I guess your local `master` is not merged with `upstream/master` correctly. How about rebasing it?
The following is an example of rebase commands.
(I assume that you set `upstream` with `git remote add upstream git@github.com:optuna/optuna.git`.)

```console
$ git fetch upstream
$ git checkout -b master-new upstream/master
$ git rebase master-new
# Your `optuna/storages/_rdb/storage.py` conflicts with upstream/master's. Please fix it.
$ git add optuna/storages/_rdb/storage.py
$ git rebase --continue
```

chenghuzi:
Sorry it seems I got things messed up. After `git rebase` the commit history is still not clean, though the conflict is now resolved.

PR #2205, opened 2021-01-14
Warn if constraints are missing in constrained NSGA-II

toshihikoyanase:
## Motivation

Depends on #2175. Addresses https://github.com/optuna/optuna/pull/2175#discussion_r556374965.

### Adding custom rules to constrained domination

Add domination relationship between trials without constraint values:
1. If neither trial x nor y have constraint values, x does not dominate y and vise versa.
2. If trial x has constraint values but y doesn't, x dominates y.
3. If trial y has constraint value but x doesn't, x does not dominate y.

These rules penalize the trials without constraint values during the non-dominated sort. They will be moved to the tail of the sorted list. Please note that these rules are not described in the original NSGA-II paper.

~I thought we can modify the definition of constrained-domination, but found out that it may degrade the optimization performance implicitly. So, I reverted the change and chose to raise errors.~

## Alternative approach

We can simply raises `ValueError` or `RuntimeError` if trials' constraints are missing.

### DIFF
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -410,8 +410,20 @@ def _constrained_dominates(
     3) Trial x and y are feasible and trial x dominates trial y.
     """
 
-    constraints0 = trial0.system_attrs[_CONSTRAINTS_KEY]
-    constraints1 = trial1.system_attrs[_CONSTRAINTS_KEY]
+    constraints0 = trial0.system_attrs.get(_CONSTRAINTS_KEY)
+    constraints1 = trial1.system_attrs.get(_CONSTRAINTS_KEY)
+
+    if constraints0 is None and constraints1 is None:
+        # Neither Trial x nor y has constraints values
+        return _dominates(trial0, trial1, directions)
+
+    if constraints0 is not None and constraints1 is None:
+        # Trial x has constraint values, but y doesn't.
+        return True
+
+    if constraints0 is None and constraints1 is not None:
+        # If Trial y has constraint values, but x doesn't.
+        return False
 
     assert isinstance(constraints0, (list, tuple))
     assert isinstance(constraints1, (list, tuple))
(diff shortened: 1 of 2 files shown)


hvy (commented):
Thanks for your careful considerations. They sound reasonable. We could be more permissive as you say with heuristics (perhaps with warnings) but I also prefer preventing silent degradations (`BoTorchSampler` could maybe adopt this change as well since it's currently different 🤔 ).

Just two questions. Is `ValueError` the correct error type here? I was wondering what "value" would refer to. Would a more general `RuntimeError` be better?

HideakiImamura (approved):
Thanks for the PR. I think `ValueError` is fine since the values of constraints for some trials are missing. IMO, the `RuntimeError` seems to be ambiguous here.

toshihikoyanase:
@HideakiImamura Thank you for your review.
I had a conversation with potential users, and they said that they wanted to restart the study from trials without constraints. So, I'm thinking of adding the rule described as the alternative approach again. This time, I'd like to show warnings for users to be aware of the risk of performance degradation. I'll send a message after implementing it.

g-votte:
As @toshihikoyanase mentions, adding constraints after running non-constrained optimization is a realistic scenario.
I prefer just showing a warning unless the degradations always or very frequently happens.

hvy (approved):
LGTM!

I find the warning approach somewhat permissive and not necessarily user friendly to be honest. But I also assumed that those cases where users would change constraints "mid optimization" were rare. The logic is well defined for this sampler (although internal and not exposed to the user) so it's not a strong opinion.

hvy:
@HideakiImamura, could you do another review since the logic has been changed since you first approved it?

toshihikoyanase:
@HideakiImamura I fixed the CI problem. PTAL.

HideakiImamura (approved):
Sorry for the late review. LGTM.

PR #2214, opened 2021-01-15
Immediately persist suggested parameters with `_CachedStorage`

hvy:
## Motivation

When using the `_CachedStorage` (which by default wraps any `RDBStorage`), trial parameter suggestion are not immediately flushed to the DB (on cache hit). This is for performance reasons reducing DB access. However, this behavior may manifest like a bug when using the recent ask and tell API. 

```python
# Process A
storage = ...  # `RDBStorage`.
study = optuna.create_study(storage=storage, study_name=study_name)
assert isinstance(study._storage, _CachedStorage), "Automatically wrapped for performance reasons."
trial = study.ask()
x = trial.suggest_...  # Any parameter suggestion.
# Exit without flushing, e.g. just save the value of "x" to a file. `Study.tell` WOULD trigger a flush.

# Process B
storage = ... # Same storage.
study = optuna.load_study(storage=storage, study_name=study_name)
trial = study.trials[0]
trial.params["x"]  # master: KeyError: 'x', PR: pass
```

## Description of the changes

Changes `_CachedStorage. set_trial_param` to always flush to the underlying `RDBStorage`.

## Note

Performance degradation reports with MySQL are available [here](https://gist.github.com/hvy/67a279e843a43a1cfc04c2eb92ea605d#gistcomment-3594701) with reproduction code. There is a rough ~5ms addition per trial, per parameter.

### DIFF
--- a/tests/test_trial.py
+++ b/tests/test_trial.py
@@ -1,6 +1,7 @@
 import copy
 import datetime
 import math
+import tempfile
 from typing import Any
 from typing import Callable
 from typing import cast
@@ -26,6 +27,7 @@ from optuna.distributions import IntUniformDistribution
 from optuna.distributions import LogUniformDistribution
 from optuna.distributions import UniformDistribution
 from optuna.study import create_study
+from optuna.study import load_study
 from optuna.testing.integration import DeterministicPruner
 from optuna.testing.sampler import DeterministicRelativeSampler
 from optuna.trial import BaseTrial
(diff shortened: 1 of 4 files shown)


tests/test_trial.py
```diff
@@ -1341,3 +1343,22 @@ def objective(trial: Trial) -> Tuple[float, float]:
         return 1.0, 1.0
 
     study.optimize(objective, n_trials=1)
+
+
+def test_persisted_param() -> None:
```
hvy, line 1348: We may want to define similar tests for other set-like operations.

HideakiImamura (approved):
Thanks for the PR. LGTM.

hvy:
C.f. tracking issue for ask and tell #2177.

ytsmiling (approved):
Thanks for the quick fix! LGTM.

hvy:
Unless anyone wants another reviewer to jump in, I think we can merge this PR.

Issue #892, opened 2020-02-04
Parallelism and sequential algorithms


Crissman:
Many of the sampling algorithms in Optuna assume sequential sampling, which is violated when parallel trials are run. We should look at implementing an algorithm like [NNI's Constant Liar](https://github.com/microsoft/nni/blob/98f66f76d310b0e0679823d966fdaa6adafb66c2/docs/en_US/CommunitySharings/ParallelizingTpeSearch.md) to address this.

876arham:
I would like to work upon this problem!

Crissman:
@876arham Please give it a start!

Crissman:
No comment from @876arham . Still looking for contribution for this issue! Volunteers?

liori:
Some design question: would you want to have constant liar switchable (and if so, where would you put the switch?)? Or maybe always-on?

Crissman:
I think default on, which option to switch off would be best.

toshihikoyanase:
Hello @liori, I think you are working on this issue according to the [conversation](https://gitter.im/optuna/optuna?at=5efc782e3a0d3931fac59fa7) on gitter. Please feel free to ask us if you have any problem to create a PR.

liori:
Ah, indeed, thank you for remembering! I have some code that I think implements one of the variants of the constant liar for TPE; it's nothing complex. However, on a test problem I'm actually getting strictly worse results, so I wanted to spend some more time on investigation. But indeed maybe a PR will be a good idea. I'll try making one soon.

toshihikoyanase:
@liori Thank you for your quick response. Maybe I can help your investigation, and I'm looking forward to your PR!

PR #2217, opened 2021-01-17
[WIP] Support parallelizing a sequential algorithm TPE

tohmae:
This PR addresses issue #892.

The usage is as follows.

```python
import optuna
from optuna.samplers import CLSampler

def objective(trial):
    y = trial.suggest_uniform("y", -5, 5)
    z = trial.suggest_uniform("z", -2, 2)
    return x ** 2 + (y - 2) ** 2 + (z - 3) ** 2

study = optuna.create_study(sampler=CLSampler(param_names=["x","y","z"],strategy="min"))
study.optimize(objective, n_trials=50, n_jobs=5)

### DIFF
--- a/optuna/samplers/__init__.py
+++ b/optuna/samplers/__init__.py
@@ -8,6 +8,7 @@ from optuna.samplers._search_space import intersection_search_space
 from optuna.samplers._search_space import IntersectionSearchSpace
 from optuna.samplers._tpe.multi_objective_sampler import MOTPESampler
 from optuna.samplers._tpe.sampler import TPESampler
+from optuna.samplers._cl import CLSampler
 
 
 __all__ = [
@@ -21,4 +22,5 @@ __all__ = [
     "RandomSampler",
     "TPESampler",
     "intersection_search_space",
+    "CLSampler",
 ]
--- a/optuna/samplers/_cl.py
+++ b/optuna/samplers/_cl.py
@@ -1,0 +1,66 @@
+import math
+from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
+from typing import Union
+import warnings
+
+import numpy as np
+import scipy.special
+from scipy.stats import truncnorm
+
+from optuna import distributions
+from optuna._study_direction import StudyDirection
+from optuna.distributions import BaseDistribution
+from optuna.exceptions import ExperimentalWarning
+from optuna.logging import get_logger
+from optuna.samplers._base import BaseSampler
+from optuna.samplers._random import RandomSampler
+from optuna.samplers._search_space import IntersectionSearchSpace
+from optuna.samplers._tpe.multivariate_parzen_estimator import _MultivariateParzenEstimator
+from optuna.samplers._tpe.sampler import TPESampler
+from optuna.study import Study
+from optuna.trial import FrozenTrial
+from optuna.trial import TrialState
+
+
+class CLSampler(BaseSampler):
+    def __init__(
+        self,
+        *,
+        base_sampler: BaseSampler,
+        concurrency: int = 3,
+        strategy: str = "min",
+        param_names: List[str] = [],
+    ):
+        self.base_sampler = base_sampler
+        self.concurrency = concurrency
+        self.strategy = strategy
+        self.param_names = param_names
+
+
+    def sample_independent(
+        self,
+        study,
+        trial,
+        param_name,
+        param_distribution,
+    ):
+
+        sampler = self.base_sampler
+        if isinstance(sampler, TPESampler):
+            return sampler.clsamples_independent(study, trial, param_name, param_distribution, self.concurrency, self.strategy)
+
+    def infer_relative_search_space(
+        self, study: Study, trial: FrozenTrial
+    ) -> Dict[str, BaseDistribution]:
+        return self.base_sampler.infer_relative_search_space(study, trial)
+
+    def sample_relative(
+        self, study: Study, trial: FrozenTrial, search_space: Dict[str, BaseDistribution]
+    ) -> Dict[str, Any]:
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by HideakiImamura** [comment]
Thanks for the PR. This PR seems to be related to https://github.com/optuna/optuna/pull/1729.

IMO, changing `_optimize.py` is overkill. If we prepare the batch evaluation interface which will be introduced in #1729, we can use `CLSampler` in a simpler way. So, how about waiting for the #1729 to proceed with this PR?

From another point of view, `CLSampler` can be implemented by wrapping "any" sampler as `PartialFixedSampler` does. What do you think?

**Comment by HideakiImamura** [comment]
> From another point of view, CLSampler can be implemented by wrapping "any" sampler as PartialFixedSampler does. What do you think?

More concretely, I mean the following implementation is fine because the constant lier algorithm can be combined with any sampling algorithms.
```python
class CLSampler(BaseSampler):
    def __init__(param_names: List[str], strategy: str, base_sampler: BaseSampler):
        ...
```

tohmae:
Thank you for your prompt response.
As you pointed out, I also think that changing _optimize.py is overkill.
I want to wait for #1729.
CLSampler will be implemented by wrapping any sampler as you reviewd.

HideakiImamura:
Oops. Sorry for the confusion; the constant liar algorithm can be used not only in batch evaluation, as I imagined, but also in distributed optimization, as you implemented. I was mistaken. I think this PR can go ahead without waiting for #1729.

By the way, I don't think you need to modify _optimize.py when using it in distributed optimization. Why did you modify it? Could you elaborate more bit?

tohmae:
Thank you for your comment. I have not used distributed optimization much, so I implemented in batch evaluation. 
I tried to implement the constant liar algorithm in distributed optimization, but
it is hard to implement study.ask and trial._suggest_parallel in multi processes.
Now study.ask seemed to work correctly in multi processes by locking table, 
but trial._suggest_parallel dit not work correctly in multi processes.
If you have some time, it would be great to hear your thoughts.

The usage is as follows.
foo.py
```python
import optuna
from optuna.samplers import TPESampler,CLSampler

def objective(trial):
    y = trial.suggest_uniform("y", -5, 5)
    z = trial.suggest_uniform("z", -2, 2)
    return x ** 2 + (y - 2) ** 2 + (z - 3) ** 2

if __name__ == '__main__':
    clsampler=CLSampler(param_names=["x","y","z"],strategy="max",concurrency=5,base_sampler=TPESampler())
    study = optuna.load_study(study_name='distributed-example', storage='mysql://root@localhost/example', sampler=clsampler)
 ```
Process 1:
 ```sh
$ python foo.py
 ```
Process 2:
 ```sh
$ python foo.py
 ```

HideakiImamura:
Your code example looks great to me. I'll dive into your implementation from now.

HideakiImamura:
I have looked through the implementation details. Do we really need to work on `optuna.Study` or `optuna.Trial`? What `CLSampler` wants to do is something like this.
- Sample a predetermined number (`concurrency` or `n_batches`) of samples simultaneously.
- Each time it is requested by a `trial` object, it selects one of the sampled values and passes it on.

I'm wondering if these requirements can be achieved by simply implementing the `CLSampler` as follows. Keep a queue of length `n_batches` in `CLSampler`, and when `sample_independent` or `sample_relative` is called, pop one from the queue and output it. If the queue is empty, sample `n_batches` of samples with the constant liar algorithm.

Then, I don't think you need to modify the following files, which is done in the current implementation. `optuna/samplers/_tpe/sampler.py, optuna/storages/_cached_storage.py, optuna/storages/_rdb/models.py, optuna/storages/_rdb/storage.py , optuna/study.py, optuna/trial/_trial.py`.

HideakiImamura:
This algorithm is very important to Optuna, and it may be quicker to talk about it verbally to work out the implementation details. We can discuss it online if you like.

toshihikoyanase:
Thank you for working on #892.

Just FYI, but let me share my prototypical implementation of the Constant Liar for TPE. It simply replaces the observation values of `RUNNING` trials with a constant value given by users. I guess it simply follows the definition of the constant liar.

https://gist.github.com/toshihikoyanase/d2bbf883f4f4acde1793ccc7effee4b7#file-cltpe_sampler-py-L119-L120
(Please note that I created this about a half year ago, so it may not work with the current version of Optuna.)

Although the change is quite limited compared with this PR, I think my approach can work with  both of the situations you discussed:

1. Batch optimization

I think we can employ my approach for batch optimization. When we suggest four trials in a batch, we can repeat the sampling algorithm four times internally. Then, the constant liar can be used to avoid the values which are suggested in the same batch.

2. Parallel optimization

I think my approach also works in parallel/distributed settings. In each worker, the sampler executes the sampling algorithm independently, and the constant liar will supply the tentative objective values for the trials launched by the other workers.

In the following gist, I casually evaluated my implementation in a parallel setting, so please take a look:

https://gist.github.com/toshihikoyanase/d2bbf883f4f4acde1793ccc7effee4b7#gistcomment-3361888

Without constant liar
image

With constant liar
image

tohmae:
Thank you for your comments.

>I'm wondering if these requirements can be achieved by simply implementing the CLSampler as follows. Keep a queue of length n_batches in CLSampler, and when sample_independent or sample_relative is called, pop one from the queue and output it. If the queue is empty, sample n_batches of samples with the constant liar algorithm.

I am sorry, it is difficult for me.

>This algorithm is very important to Optuna, and it may be quicker to talk about it verbally to work out the implementation details.

Issue #892 was open for a long time, so I created the PR comfortably.
If it is very important to Optuna, I want to return #892 to you.　:bow:

jeromepatel:
@tohmae may I ask you that are you still working on this PR/Draft?

tohmae:
I am sorry for the late reply. I close this draft once.

PR #2230, opened 2021-01-19
Refactor `examples/multi_objective/pytorch_simple.py`

not522:
## Motivation
To fix the complicated code for sorting of `best_trials` .
## Description of the changes
- Simplify the sorting of `best_trials`.


hvy (approved):
Thanks, LGTM!

It would be great if we could fix `examples/botorch_simple.py` as well since it contains the exact same code.

toshihikoyanase (approved):
LGTM! Thank you.

PR #2236, opened 2021-01-19
Fix the wrong way to set `datetime_start` (clean)

### LINKED ISSUE #2036: Start time is wrong when calculating trials added by `enqueue_trial`.

chenghuzi:
> A clean version of [this PR](https://github.com/optuna/optuna/pull/2196) to change the way to set `datetime_start`, as that one is polluted and closed due to my mistaken push with `--force` option. Sorry for the repeated PR.

Fixes #2036.

## Motivation

The current mechanism to set the `datetime_start` of a given trial is to add a time value when creating that trial. However, this could be wrong if the created trial is not the one to be executed right after its creation. The `datetime_start` of these waiting trials won't be updated before and after their delayed execution. For specific example, see [#2306](https://github.com/optuna/optuna/issues/2036).

## Description of the changes

The mechanism to initialize `datetime_start` is changed. Now we set the time of trial start when its state changes to `RUNNING`. This ensures the `datetime_start ` to be the moment right before execution. This is implemented in `set_trial_state`.

Xnip2021-01-13_00-34-10
Trial 11-20 were appended to the study through `enqueue_trial ` after trial 10 but were executed several minutes later. After change the `datetime_start ` becomes the 2s duration set by the objective function:

```python
def objective(trial):
    time.sleep(2)
    return x**2
```


hvy (changes requested):
Thanks, the history looks better. Could you add unit tests to verify the change in logic? I'm by the way still wondering if we shouldn't set the datetime to `None` when trials are enqueued.

toshihikoyanase:
@c-bata Could you review this PR? The author of #2196 fixed the commit history and created this PR.

chenghuzi:
> Thanks, the history looks better. Could you add unit tests to verify the change in logic? I'm by the way still wondering if we shouldn't set the datetime to `None` when trials are enqueued.

Sure, I'm considering the unit test now. As for the second issue, there are two possible ways to implement it: 

1. Change the `FrozenTrial` and storage model since the column `datetime_start` has a default value `datetime.now`. 
2. Reset the `datetime_start` of a created trial to `None` in `enqueue_trial`.

1 involves many changes, so I'm not sure if it's ok. 2 requires modification in `_update_trial` and some other methods, but it's not very conceptually consistent as we 're canceling `datetime_start` after creation.

hvy:
I agree that 2 is less intrusive. Where were you thinking of canceling the `datetime_start`? You might have to change `RDBStorage._get_prepared_new_trial` to explicitly nullify the column after insertion. You'd also need to update `FrozenTrial._validate` to allow `datetime_start` to be `None` when the state is `TrialState.WAITING`, probably introduce a new parameter to `create_trial`.

chenghuzi:
I tried to nullify `datetime_start` in `RDBStorage._get_prepared_new_trial` today, but it seems not to work as sqlalchemy always adds a value to `datetime_start ` as long as `default` is set. So if we want a trial with null `datetime_start`, changes in trial model might be inevitable.

hvy:
Thanks for your quick response.

Curious but did you also try updating the row after insertion? I noticed as well (with MySQL) the unfortunate behavior that passing `None` for the column during insertion will fall back to the default value. However, updating the row afterwards with `None` successfully nullified the field.

Though this seems undocumented and somewhat risky to rely on, perhaps less controversial than altering the schema. If you've done a more exhaustive investigation I'd be glad to know how that turned out.

chenghuzi:
Yes that's feasible, we can `datetime_start` to `None` after the first `session.flush()`.

chenghuzi:
With changes in the 2nd stage processing in `_get_prepared_new_trial` and few methods in `FrozenTrial`, now the enqueued trials are initialized with null `datetime_start`, though the `datetime_start` set by `ask` is still not null, since there's no template trial and the modification of `datetime_start` is under that part. So we can either leave it as it is now, or create all trials with null `datetime_start`. To do that, I'll move the modification outside the template trial condition. How do you think about this?

hvy:
> the datetime_start set by ask is still not null,

I think this behavior is expected. A trial returned by `Study.ask` should be running (similar to a trial passed to the objective function callback).

chenghuzi:
> > the datetime_start set by ask is still not null,
> 
> I think this behavior is expected. A trial returned by `Study.ask` should be running (similar to a trial passed to the objective function callback).

We have `_storage.set_trial_state(trial_id, TrialState.RUNNING)` in the new `ask`, so there's a little redundancy, but that's ok as the effect is the same. So I'll push with the current changes then.

hvy:
> So I'll push with the current changes then.

Sounds great.

> We have _storage.set_trial_state(trial_id, TrialState.RUNNING) in the new ask

Are you referring to the call in `Study._pop_waiting_trial_id` ? I think it's needed just like in `Study.optimize` to update the state from `TrialState.WAITING`.

chenghuzi:
> Are you referring to the call in `Study._pop_waiting_trial_id` ? I think it's needed just like in `Study.optimize` to update the state from `TrialState.WAITING`.

Yes, so that should be ok.

c-bata:
> @c-bata Could you review this PR? The author of #2196 fixed the commit history and created this PR.

Sorry for the late reply. I'm busy today so that I'll review this PR tomorrow.

c-bata:
> Curious but did you also try updating the row after insertion? I noticed as well (with MySQL) the unfortunate behavior that passing `None` for the column during insertion will fall back to the default value. However, updating the row afterwards with `None` successfully nullified the field.
> 
> Though this seems undocumented and somewhat risky to rely on, perhaps less controversial than altering the schema. If you've done a more exhaustive investigation I'd be glad to know how that turned out.

Is this unfortunate behavior happened after removing the default of `datetime_start`?
https://github.com/optuna/optuna/blob/6f33c238356f4db766509cdaf08343717ba69977/optuna/storages/_rdb/models.py#L185

There are redundant SQL queries when updating the row after insertion. So I think it's better to change the model definition.

chenghuzi:
> There is one thing I would like to confirm, can we avoid this unfortunate behavior if we remove the default of `datetime_start`?
> 
> https://github.com/optuna/optuna/blob/6f33c238356f4db766509cdaf08343717ba69977/optuna/storages/_rdb/models.py#L185
> 
> There are redundant SQL queries when updating the row after insertion. So I think it's better to change the model definition.

Sure, removing default value is the most straightforward way. But this leads to errors in tests for `datetime_start`, such as `tests/integration_tests/test_chainermn` and `tests/storages_tests/rdb_tests/test_models.py`. I'm not sure if changing the logic or removing related test functions is ok, and that's why we turned to overwrite the initial `datetime_start`. For other tests removing that default value is ok.

hvy:
If we are to change the default value, we should make sure that trials constructed normally (not enqueue) do have a valid timestamp. I think the ChainerMN test should pass, by changing the trial construction logic and not the test.

chenghuzi:
I see your point. Previously my thought is that a trial should not be created with a non-null `datetime_start` if we remove the default value, because I assumed that all trials should be evaluated through `ask()` and that should be the place we set the `datetime_start`. With this logic, we may need to change ChainerMN test.

But if all normally constructed trials are expected to have valid `datetime_start`, we can simply set it in the instantiation of `FrozenTrial`. However, in this case the `test_default_datetime` still needs change by passing a datime_start parameter during init. Or alternatively. we can remove that test completely. Which one do you think is better?

chenghuzi:
@hvy if available, could you review the latest updates?

optuna/study.py
```diff
@@ -417,6 +417,10 @@ def ask(self) -> trial_module.Trial:
         trial_id = self._pop_waiting_trial_id()
         if trial_id is None:
             trial_id = self._storage.create_new_trial(self._study_id)
+
+        # update the trial state to running and **change the datetime_start**
+        self._storage.set_trial_state(trial_id, TrialState.RUNNING)
```
hvy, line 422: Does the trial already have the `RUNNING` state and a non-Null  `datetime_start` before reaching those lines?

optuna/trial/_frozen.py
```diff
@@ -317,8 +317,11 @@ def set_system_attr(self, key: str, value: Any) -> None:
 
     def _validate(self) -> None:
 
-        if self.datetime_start is None:
-            raise ValueError("`datetime_start` is supposed to be set.")
+        if self.state != TrialState.WAITING:
+            if self.datetime_start is None:
+                raise ValueError(
+                    "`datetime_start` is supposed to be set when the trial state is not waiting."
+                )
```
hvy, line 324: Can you add unit tests for this logic in `test_frozen.py`?

tests/storages_tests/rdb_tests/test_models.py
```diff
@@ -141,15 +141,12 @@ class TestTrialModel(object):
     @staticmethod
     def test_default_datetime(session: Session) -> None:
 
-        datetime_1 = datetime.now()
-
-        session.add(TrialModel(state=TrialState.RUNNING))
+        session.add(TrialModel(state=TrialState.WAITING))
```
hvy, line 144: Might be a good idea to add a comment that the state doesn't matter here.

tests/study_tests/test_study.py
```diff
@@ -990,3 +990,43 @@ def test_tell_pruned_values() -> None:
...
+
+    with StorageSupplier(storage_mode) as storage:
+        study = create_study(storage=storage)
+
+        def objective(trial: Trial) -> float:
+            time.sleep(1)
+            x = trial.suggest_int("x", -10, 10)
+            return x
+
+        study.enqueue_trial(params={"x": 1})
```
hvy, line 1006: How about adding an assertion that `datetime_start` is `None` after this line?

hvy (changes requested):
Thanks for the changes. Seems like we're getting things together. Besides the comment, I'm wondering if we should add tests for the `StudySummary` when there are only enqueued trials. It's somewhat of an edge case but very easy to reproduce.

optuna/study.py
chenghuzi: Yeah when the trial is created directly this leads to repeated changes to trial state. Already fixed in the latest commit. So now only those waiting trials will be updated.

optuna/trial/_frozen.py
chenghuzi: Added.

tests/study_tests/test_study.py
chenghuzi: Added.

chenghuzi:
> Thanks for the changes. Seems like we're getting things together. Besides the comment, I'm wondering if we should add tests for the `StudySummary` when there are only enqueued trials. It's somewhat of an edge case but very easy to reproduce.

Agreed and test added. Actually found an error if completed and waiting trials coexist in the same study. Have that fixed after eliminating waiting trials from `datetime_start` calculation for `StudySummary`.

Below is the test
https://github.com/optuna/optuna/blob/9ff90e20e1ea88aa41f69d84f4a34ef5ee374bb6/tests/study_tests/test_study.py#L1060

optuna/storages/_cached_storage.py
```diff
@@ -199,6 +200,11 @@ def set_trial_state(self, trial_id: int, state: TrialState) -> bool:
                 updates = self._get_updates(trial_id)
                 cached_trial.state = state
                 updates.state = state
+
+                if state == TrialState.RUNNING:
+                    updates.datetime_start = datetime.datetime.now()
+                    cached_trial.datetime_start = datetime.datetime.now()
```
hvy, line 206: If we set a running trial to running once again, it seems like `datetime_start` will be updated. I think it's inconsistent with the other storages. How about only updating the timestamp here if the state actually changes? I'm sorry for the lack of missing tests here.

optuna/storages/_in_memory.py
```diff
@@ -171,7 +171,9 @@ def _build_study_summary(self, study_id: int) -> StudySummary:
                 [
                     cast(datetime, trial.datetime_start)
                     for trial in self.get_all_trials(study_id, deepcopy=False)
-                ]
+                    if trial.datetime_start is not None
+                ],
+                default=None,
```
hvy, line 176: Didn't know about this argument 👍

optuna/study.py
```diff
@@ -457,6 +457,8 @@ def ask(
         trial_id = self._pop_waiting_trial_id()
         if trial_id is None:
             trial_id = self._storage.create_new_trial(self._study_id)
+        else:
+            self._storage.set_trial_state(trial_id, TrialState.RUNNING)
```
hvy, line 461: I think this state update is already performed in `_pop_waiting_trial_id` and isn't needed therefore.

tests/study_tests/test_study.py
```diff
@@ -1010,3 +1010,71 @@ def test_tell_pruned_values() -> None:
...
+
+        t2 = study.trials[2]
+        assert t2.datetime_start is not None
+        assert t2.datetime_complete is not None
+        duration_error2 = abs((t2.datetime_complete - t2.datetime_start).total_seconds() - 1)
+        assert duration_error2 < 0.1
+
+
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES)
+def test_studysummary_datetime_start_calculation(storage_mode: str) -> None:
```
hvy, line 1060: A nit.
```suggestion
def test_study_summary_datetime_start_calculation(storage_mode: str) -> None:
```

hvy (changes requested):

optuna/study.py
chenghuzi: You're right. Sorry forgot that

optuna/storages/_cached_storage.py
chenghuzi: Agreed. Will add that in next commit.

optuna/storages/_cached_storage.py
```diff
@@ -196,9 +197,15 @@ def set_trial_state(self, trial_id: int, state: TrialState) -> bool:
...
             if cached_trial is not None:
                 self._check_trial_is_updatable(cached_trial)
+                previous_state = copy.copy(cached_trial.state)
                 updates = self._get_updates(trial_id)
                 cached_trial.state = state
                 updates.state = state
+
+                if state == TrialState.RUNNING and previous_state == TrialState.WAITING:
+                    updates.datetime_start = datetime.datetime.now()
+                    cached_trial.datetime_start = datetime.datetime.now()
```
hvy, line 207: With the current cached storage logic, I don't think we'll enter this code since a waiting trial isn't cached. It might be worth adding an assertion. What do you think?

hvy (commented):
Sorry for the delay, and thanks for the updated. The changes basically LGTM. I left a minor questions if you could take a look.

hvy:
@c-bata could you take a look at these changes too? If you're busy, I could reassign it to someone else.

optuna/storages/_cached_storage.py
chenghuzi: I see. Previously added this partially because of logic in `test_set_trial_state`. Now fixed with assertion and modified tests.

c-bata:
Sorry for the delayed response. I'll review this later.

tests/study_tests/test_study.py
```diff
@@ -1010,3 +1010,71 @@ def test_tell_pruned_values() -> None:
...
+        assert t1.datetime_start is not None
+        assert t1.datetime_complete is not None
+        duration_error1 = abs((t1.datetime_complete - t1.datetime_start).total_seconds() - 1)
+        assert duration_error1 < 0.1
+
+        t2 = study.trials[2]
+        assert t2.datetime_start is not None
+        assert t2.datetime_complete is not None
+        duration_error2 = abs((t2.datetime_complete - t2.datetime_start).total_seconds() - 1)
+        assert duration_error2 < 0.1
```
c-bata, line 1056: How about using `pytest.approx()` here?
https://docs.pytest.org/en/stable/reference.html#pytest-approx

c-bata: On second thought, I concerned this test case makes the time for test execution longer (seemingly it consumes 5 seconds for each storage backends). We can also reduce the duration of `time.sleep`, but it makes tests fragile.

How about removing this test case? I think the following test case is enough for this purposes.

```python

            time.sleep(1)
            x = trial.suggest_int("x", -10, 10)
            return x

        study.enqueue_trial(params={"x": 1})
        assert study.trials[0].datetime_start is None

        assert study.trials[0].datetime_start is not None
```

c-bata (commented):
Thank you @chenghuzi for your contribution. The changes looks almost perfect.

tests/study_tests/test_study.py
chenghuzi: Thanks, didn't know `pytest.approx` before. As for the latter one, indeed it takes too long to finish. I agree that we can replace duration calculation with simple `is none` test as we've eliminated repeated `set_trial_state` to `RUNNING`. Will update later.

c-bata (approved):
LGTM!!

optuna/storages/_cached_storage.py
hvy: Ah, I meant to assert that a cached trial is never in the waiting state. A user may try to set the trial to running twice, in which case it should be a NOOP so the current assertion doesn't really hold in that case. I can follow this up in a PR.

hvy (approved):
Thanks for the quick review @c-bata, and the long awaited PR @chenghuzi , LGTM!

https://github.com/optuna/optuna/pull/2236#discussion_r579964281 is an issue I think but I'll create a follow up PR for this.

PR #2243, opened 2021-01-20
Ensure uniform and loguniform distributions less than high boundary

Crissman:
## Motivation

When using uniform or loguniform distributions, the high boundary should be excluded. Moving this logic to the search space transform module to centralize.

## Description of the changes

Use `numpy.nextafter` to subtract the numpy eps from the higher boundary of uniform and loguniform distributions.

### DIFF
--- a/optuna/_transform.py
+++ b/optuna/_transform.py
@@ -102,10 +102,7 @@ class _SearchSpaceTransform:
 
         Returns:
             A 1-dimensional ``numpy.ndarray`` holding the transformed parameters in the
-            configuration. For :class:`~optuna.distributions.UniformDistribution` and
-            :class:`~optuna.distributions.LogUniformDistribution`, the high bound of the
-            search space is reduced by the minimal amount possible, to ensure the high
-            boundary value is not suggested, as per the definitions of those distributions.
+            configuration.
 
         """
         trans_params = numpy.zeros(self._bounds.shape[0], dtype=numpy.float64)
@@ -199,12 +196,12 @@ def _transform_search_space(
             if isinstance(d, UniformDistribution):
                 bds = (
                     _transform_numerical_param(d.low, d, transform_log),
-                    _transform_numerical_param(d.high, d, transform_log, clip=True),
+                    _transform_numerical_param(d.high, d, transform_log),
                 )
             elif isinstance(d, LogUniformDistribution):
                 bds = (
                     _transform_numerical_param(d.low, d, transform_log),
-                    _transform_numerical_param(d.high, d, transform_log, clip=True),
+                    _transform_numerical_param(d.high, d, transform_log),
                 )
             elif isinstance(d, DiscreteUniformDistribution):
                 half_step = 0.5 * d.q if transform_step else 0.0
@@ -241,10 +238,7 @@ def _transform_search_space(
 
 
 def _transform_numerical_param(
-    param: Union[int, float],
-    distribution: BaseDistribution,
-    transform_log: bool,
-    clip: bool = False,
+    param: Union[int, float], distribution: BaseDistribution, transform_log: bool
 ) -> float:
     d = distribution
 
@@ -263,9 +257,6 @@ def _transform_numerical_param(
     else:
         assert False, "Should not reach. Unexpected distribution."
 
-    if clip:
-        trans_param = numpy.nextafter(trans_param, trans_param - 1)
-
     return trans_param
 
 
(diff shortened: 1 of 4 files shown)


hvy:
I think the previous logic was to do this processing at untransformation. Is there a particular reason why we're doing this at the transformation now?

Crissman:
Yes. Changing the boundaries is more accurate. Previous approach at the transformation resulted in possible double count of generation of parameters at the `high` boundary. Take a small float range, with 1000 possible values between or equal to the low and high boundaries. If it was equal to the high boundary, then fixing it at the transform range using `min` would reduce the top value to one less than the top value, in effect double-counting the chance of that value being generated.

Granted this is an extremely small probability, but changing the boundary itself equalizes the distribution by removing that chance.

toshihikoyanase:
@hvy Thank you for adding a comment. Could you continue the review, please?

@HideakiImamura If you have time, could you review this PR?

optuna/_transform.py
```diff
@@ -238,7 +238,10 @@ def _transform_search_space(
 
 
 def _transform_numerical_param(
-    param: Union[int, float], distribution: BaseDistribution, transform_log: bool
+    param: Union[int, float],
+    distribution: BaseDistribution,
+    transform_log: bool,
+    clip: bool = False,
```
HideakiImamura, line 244: For `clip`, I think it is better to specify it at each call instead of setting a default value, as it is less likely to lead to unintended behavior and is more consistent with other arguments.

HideakiImamura (commented):
Thanks for the PR! I have a minor comment. Please take a look.

tests/test_transform.py
```diff
@@ -95,6 +95,11 @@ def test_search_space_transform_numerical(
     expected_low = distribution.low  # type: ignore
     expected_high = distribution.high  # type: ignore
 
+    if isinstance(distribution, UniformDistribution) or isinstance(
+        distribution, LogUniformDistribution
+    ):
```
hvy, line 100: Mostly a cosmetic but you can shorten this.
```suggestion
    if isinstance(distribution, (UniformDistribution, LogUniformDistribution)):
```

optuna/_transform.py
Crissman: Hmm... I prefer less code. The other arguments are passed through from the calling function so they cannot use a default value. `clip` addresses a boundary case that is only used in two out of ten cases. Adding `clip=False` tells the reader two things: 1. there is a special case, and 2. you can ignore it this time. I think it's better to only flag the special case when used.

Is it ok to leave as is?

hvy (commented):
Thanks for the PR. Looks good overall. 

I would prefer separating the refactoring part and the change in logic since the scopes are different, although I agree that practical impacts are small. Not a strong opinion. 

For the record, this'll have unintended impacts on HPI, but again I think those are negligible. It might be a good idea to document this behavior under `transform`. I know this class is private but it's a quite important component and actually has docs already.

optuna/_transform.py
hvy: Just leaving a personal comment. I also often stumble upon the choice of whether to be explicit about arguments in those private APIs. In general, I agree with @HideakiImamura. An exceptions would be cases like this, where the default value is dominant in the vast majority of calls. The current decision by @Crissman is actually aligned with `_SearchSpaceTransform.__init__`'s `transform_{log,step}`, which also have their default values despite being private APIs.

Crissman:
Documentation of high behaviour documented in `transform`.

hvy:
Thanks for the documentation, looks good.

As for the following, briefly discussed with @g-votte. Let's keep this PR a pure refactoring PR. There are some controversies around the exclusion of the upper bound (in the first place, if we shouldn't make it inclusive instead) and there's little benefit in changing the behavior now before that's settled.

> Yes. Changing the boundaries is more accurate. Previous approach at the transformation resulted in possible double count of generation of parameters at the high boundary. Take a small float range, with 1000 possible values between or equal to the low and high boundaries. If it was equal to the high boundary, then fixing it at the transform range using min would reduce the top value to one less than the top value, in effect double-counting the chance of that value being generated.

> Granted this is an extremely small probability, but changing the boundary itself equalizes the distribution by removing that chance.

Crissman:
@hvy Understood. Reverted to refactor. 👍

hvy (approved):
Thanks @Crissman, looks quite straight forward now as a refactoring. Let me approve it and wait for @HideakiImamura's comments. LGTM!

HideakiImamura (approved):
Looks very good! LGTM.

PR #2244, opened 2021-01-20
Move botorch example to multi_objective directory

crcrpar:
I think botorch example is an example of multi objective so that
let it reside in multi objective directory.


g-votte (approved):
LGTM

PR #2249, opened 2021-01-21
Example to save PyTorch model

Crissman:
## Motivation

For machine learning, to avoid having a manual HPO step and training step, it's beneficial to use the Objective function also for training. Also, since the initial training with good parameters is found during the HPO trials, saving those trained parameters can give a jumpstart on the training.

## Description of the changes

Add a global switch to change from HPO to model training, and save or load models accordingly.


toshihikoyanase:
@crcrpar If you have time, could you review this PR, please? I think you are familiar with PyTorch.

Crissman:
Experimental testing of this actually showed decreased performance (regression to the mean?) so closing this PR.

PR #2257, opened 2021-01-24
Add init files in test directories.

keisuke-umezawa:
## Motivation

There are missing init files, so I added them by this PR.


hvy (changes requested):
Good catch and thanks, LGTM!

Could you just merge the latest master to fix the CI failure?

hvy (approved):
LGTM!

PR #2261, opened 2021-01-25
Add `Study.add_trials` to simplify creating customized study

not522:
## Motivation
`Study.add_trial` was introduced to create a custom study. But, we still need to write a bit long code in some cases, e.g., visualization with filtered trials.

```
study_x0 = create_study()
for trial in study.trials:
    if trial.params["x"] > 0:
        study_x0.add_trial(trial)
plot_slice(study_x0)
```

This PR adds `Study.add_trials` and enables to simplify the above code as follows.

```
study_x0 = create_study()
study_x0.add_trials(filter(lambda t: t.params["x"] > 0, study.trials))
plot_slice(study_x0)
```

## Description of the changes
- Add `Study.add_trials`.
- Add test.

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -848,6 +848,49 @@ class Study(BaseStudy):
 
         self._storage.create_new_trial(self._study_id, template_trial=trial)
 
+    @experimental("2.5.0")
+    def add_trials(self, trials: Sequence[FrozenTrial]) -> "Study":
+        """Add trials to study.
+
+        The trials are validated before being added.
+
+        Example:
+
+            .. testcode::
+
+                import optuna
+                from optuna.distributions import UniformDistribution
+
+
+                def objective(trial):
+                    x = trial.suggest_uniform("x", 0, 10)
+                    return x ** 2
+
+
+                study = optuna.create_study()
+                study.optimize(objective, n_trials=3)
+                assert len(study.trials) == 3
+
+                other_study = optuna.create_study().add_trials(study.trials)
+                assert len(other_study.trials) == len(study.trials)
+
+                other_study.optimize(objective, n_trials=2)
+                assert len(other_study.trials) == len(study.trials) + 2
+
+        .. seealso::
+
+            See :func:`~optuna.study.add_trial` for addition of each trial.
+
+        Args:
+            trials: Trials to add.
+
+        """
+
+        for trial in trials:
+            self.add_trial(trial)
+
+        return self
+
     def _pop_waiting_trial_id(self) -> Optional[int]:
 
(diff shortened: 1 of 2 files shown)


hvy:
@sile, @c-bata could you like to take a look at this PR?

optuna/study.py
```diff
@@ -848,6 +848,49 @@ def objective(trial):
...
+                other_study.optimize(objective, n_trials=2)
+                assert len(other_study.trials) == len(study.trials) + 2
+
+        .. seealso::
+
+            See :func:`~optuna.study.add_trial` for addition of each trial.
+
+        Args:
+            trials: Trials to add.
+
```
sile, line 886: How about adding `Returns:` and `Raises:` sections here?

not522: Thank you! I updated the PR following your comment.

sile (approved):
LGTM

optuna/study.py
```diff
@@ -848,6 +848,55 @@ def objective(trial):
 
         self._storage.create_new_trial(self._study_id, template_trial=trial)
 
+    @experimental("2.5.0")
+    def add_trials(self, trials: Sequence[FrozenTrial]) -> "Study":
```
c-bata, line 852: I think the minimum requirements of `trials` argument is `Iterable` since it is just iterated by for-loop. But the `Sequence` is an interface to ensure `Reversible` and `Collection` (`Sized`, `Iterable`, and `Container`).
https://docs.python.org/3/library/typing.html#typing.Sequence

How about using `Iterable[FrozenTrial]` here?

not522: Thank you. I changed it.

optuna/study.py
```diff
@@ -848,6 +849,55 @@ def objective(trial):
...
+
+        .. seealso::
+
+            See :func:`~optuna.study.Study.add_trial` for addition of each trial.
+
+        Args:
+            trials: Trials to add.
+
+        Returns:
+            Return self.
```
c-bata, line 889: @not522 @sile I prefer that `add_trials()` method returns nothing because of following reasons.

1. This method provides in-place operation.
1. `add_trial()` method does not return anything.

What do you think?

sile: I'm neutral about this point.
However, indeed, I also felt the return type of `add_trials()` was inconsistent with the one of `add_trial()` (so if we allow `add_trials()` returning `self`, it might be better to modify `add_trial()` as well).

not522: Thank you for your suggestion. To keep consistency with other methods, I changed the return value as None. PTAL.

c-bata (approved):
LGTM!

PR #2271, opened 2021-01-27
Add an option to the `Study.ask` method that allows define-and-run parameter suggestion.

g-votte:
This PR adds an option to the `Study.ask` method, where the user can specify the predefined search space so that the parameters in the space is automatically suggested.

## Motivation
The option allows the user to get suggestions in the define-and-run way. In the following example, the parameters are automatically suggested and accessible with `trial.params` just after the `ask` call, passing the previously defined search space as `fixed_distributions` option.

```python
study = optuna.create_study()

# For example, the distributions are previously defined when using `create_trial`.
distributions = {
    "optimizer": optuna.distributions.CategoricalDistribution(["adam", "sgd"]),
    "lr": optuna.distributions.LogUniformDistribution(0.0001, 0.1),
}
trial = optuna.trial.create_trial(
    params={"optimizer": "adam", "lr": 0.0001},
    distributions=distributions,
    value=0.5,
)
study.add_trial(trial)

# You can pass the distributions previously defined.
trial = study.ask(fixed_distributions=distributions)

# `optimizer` and `lr` are already suggested and accessible with `trial.params`.
print(trial.params)
```


g-votte:
This PR is ready for review. I also created #2304 to discuss supporting a similar option with the `optimize` method.

HideakiImamura:

optuna/study.py
```diff
@@ -409,17 +414,62 @@ def ask(self) -> trial_module.Trial:
...
+                distributions = {
+                    "optimizer": optuna.distributions.CategoricalDistribution(["adam", "sgd"]),
+                    "lr": optuna.distributions.LogUniformDistribution(0.0001, 0.1),
+                }
+                trial = optuna.trial.create_trial(
+                    params={"optimizer": "adam", "lr": 0.0001},
+                    distributions=distributions,
+                    value=0.5,
+                )
+                study.add_trial(trial)
```
HideakiImamura, line 439: I don't have a strong opinion, but these processes that `add_trial` to existing `study` are not directly related to `study.ask(fixed_distributions)`, so it is a bit strange to have them in the examples in this document. Since these processes are listed as concrete use cases in the PR, why not exclude them from the example?

optuna/study.py
```diff
@@ -409,17 +414,62 @@ def ask(self) -> trial_module.Trial:
...
+                # `optimizer` and `lr` are already suggested and accessible with `trial.params`.
+                assert "optimizer" in trial.params
+                assert "lr" in trial.params
+
+        Args:
+            fixed_distributions:
+                A dictionary containing the parameter names and parameter's distributions. Each
+                parameter in this dictionary is automatically suggested for the returned trial,
+                even when the suggest method is not explicitly invoked by the user. If this
+                argument is set to :obj:`None`, no parameter is automatically suggested.
```
HideakiImamura, line 453: Just curious, but wouldn't it be inconsistent with the arguments of the other methods in Optuna to only suggest to the specified ones when values are specified, and not suggest anything when it is `None`? For example, in `plot_param_importance`, if `params` is specified, it will calculate the importance only for the specified ones, and if not, it will calculate the importance for all parameters.

HideakiImamura (commented):
Thanks for the PR. This feature will help users avoid tedious redefinition of the search space and use Optuna efficiently. I have several comments. Please take a look.

optuna/study.py
g-votte: Hmm, do you have any idea of the alternative behavior? Being honest, I'm not sure if we should stick the consistency because: (1) this is not a visualization method and (2) "all parameters" are unknown for this method.

optuna/study.py
g-votte: Thanks for your comment. The motivation is to demonstrate a scenario in the document, where the search space have to be previously defined. But, I think we can just remove `add_trial` and `create_trial` like the following snippet.

```python
import optuna

study = optuna.create_study()

distributions = {
    "optimizer": optuna.distributions.CategoricalDistribution(["adam", "sgd"]),
    "lr": optuna.distributions.LogUniformDistribution(0.0001, 0.1),
}

# You can pass the distributions previously defined.
trial = study.ask(fixed_distributions=distributions)

# `optimizer` and `lr` are already suggested and accessible with `trial.params`.
assert "optimizer" in trial.params
assert "lr" in trial.params
```

HideakiImamura: Looks solid! It's simple and easy to understand how to use this feature.

optuna/study.py
HideakiImamura: > "all parameters" are unknown for this method.

It's true that we can't sample everything in advance, because we shouldn't know the whole search space at all when we ask for a trial. `visualization` and `study` methods are different, so we may not have to worry about that much. Let me resolve this conversation without any changes.

optuna/study.py
g-votte: Thanks. I removed the lines from the example.

HideakiImamura (approved):
Thanks for the update! LGTM!

optuna/study.py
```diff
@@ -386,7 +387,9 @@ def objective(trial):
             show_progress_bar=show_progress_bar,
         )
 
-    def ask(self) -> trial_module.Trial:
+    def ask(
+        self, fixed_distributions: Optional[Dict[str, BaseDistribution]] = None
```
toshihikoyanase, line 391: [Question] We sometimes call a dictionary of distributions `search_space` as can be seen in the arguments of [`Sampler.sample_relative`](https://github.com/optuna/optuna/blob/85f4d1c41baa00ea0c666fd6c1fc3924cca2d14e/optuna/samplers/_base.py#L75) . So this argument could be `fixed_search_space`.
I think you intentionally avoid the name because users may think that they cannot add parameters other than `fixed_search_space`. Actually, we can suggest the parameters which are not in the argument. Therefore, you chose `fixed_distributions` instead of `fixed_search_space`. Does it make sense?

```python
trial = study.ask({"x": UniformDistribution(1, 10)})

trial.suggest_float("y", 0, 1)  # Some users may think 
```

g-votte: Right; `search_space` might be rather confusing especially for ppl who are accustomed with the define-and-run style. In addition, we use the term of `distributions` in the argument of [`create_trial`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.trial.create_trial.html#optuna.trial.create_trial).

toshihikoyanase: Thank you for your answer. Now, I'm confident with the choice of `fixed_distribution`.

toshihikoyanase (approved):
LGTM!
I think this feature is suitable for config system and ML pipeline. I'll use it to simplify the implementation of [Hydra's Optuna plugin](https://github.com/facebookresearch/hydra/blob/2fddce8c7801a9c787e4db7860c9e9474e9f0036/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py#L190-L194). Thank you.

Issue #2247, opened 2021-01-21
`MOTPESampler` raises `ValueError` when it solves a simple knapsack problem


toshihikoyanase:
## Expected behavior

When I tried to solve a simplified multi-objective knapsack problem using `MOTPE`, I got `ValueError`. I think it should handle this problem or show a more user-friendly error message.

The problem is defined as follows:
```python
import numpy as np

import optuna


def create_knapsack(n_items, n_knapsack, seed):
    rng = np.random.RandomState(seed=seed)
    profits = rng.randint(10, 101, size=(n_knapsack, n_items))
    return profits


def knapsack(trial, profits):
    p0, p1 = 0, 0
    for i, p in enumerate(profits.tolist()):
        x = trial.suggest_categorical(f"x_{i}", (0, 1))
        p0 += p[0] * x
        p1 += p[1] * x

    return p0, p1

profits = create_knapsack(20, 2, 1)
objective = lambda t: knapsack(t, profits)
```

## Environment

- Optuna version: v2.4.0
- Python version: Python 3.6 (Google Colab)
- OS: N/A (Google Colab)

## Error messages, stack traces, or logs

```console
[W 2021-01-21 06:35:39,582] Trial 20 failed because of the following error: ValueError('pvals < 0, pvals > 1 or pvals contains NaNs',)
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/optuna/_optimize.py", line 198, in _run_trial
    value_or_values = func(trial)
  File "<ipython-input-3-7aaa3bf88873>", line 7, in <lambda>
    objective = lambda t: knapsack(t, profits)
  File "<ipython-input-2-d1a010db23b5>", line 15, in knapsack
    x = trial.suggest_categorical(f"x_{i}", (0, 1))
  File "/usr/local/lib/python3.6/dist-packages/optuna/trial/_trial.py", line 505, in suggest_categorical
    return self._suggest(name, CategoricalDistribution(choices=choices))
  File "/usr/local/lib/python3.6/dist-packages/optuna/trial/_trial.py", line 717, in _suggest
    study, trial, name, distribution
  File "/usr/local/lib/python3.6/dist-packages/optuna/samplers/_tpe/multi_objective_sampler.py", line 208, in sample_independent
    study, trial, param_distribution, below_param_values, above_param_values
[...]

PR #2274, opened 2021-01-27
Fix calculate_weights_below

y0z:
## Motivation
This PR fixes issue #2247.

## Description of the changes
- Fix `calculate_weights_below` (see #2247 for details)
- Add some test cases for `calculate_weights_below`

### DIFF
--- a/optuna/samplers/_tpe/multi_objective_sampler.py
+++ b/optuna/samplers/_tpe/multi_objective_sampler.py
@@ -557,9 +557,10 @@ class MOTPESampler(TPESampler):
                     - self._compute_hypervolume(
                         np.asarray(lvals_below[:i] + lvals_below[i + 1 :]), reference_point
                     )
-                    for i in range(len(lvals))
+                    for i in range(n_below)
                 ]
             )
+            contributions += EPS
             weights_below = np.clip(contributions / np.max(contributions), 0, 1)
             return weights_below
 
--- a/tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
+++ b/tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
@@ -567,6 +567,44 @@ def test_calculate_nondomination_rank() -> None:
     assert ranks == [0, 1, 0, 2, 0, 1]
 
 
+def test_calculate_weights_below() -> None:
+    sampler = MOTPESampler()
+
+    # Two samples
+    weights_below = sampler._calculate_weights_below(
+        np.array([[0.2, 0.5], [0.9, 0.4], [1, 1]]), np.array([0, 1])
+    )
+    assert len(weights_below) == 2
+    assert weights_below[0] > weights_below[1]
+    assert sum(weights_below) > 0
+
+    # Two equally contributed samples
+    weights_below = sampler._calculate_weights_below(
+        np.array([[0.2, 0.8], [0.8, 0.2], [1, 1]]), np.array([0, 1])
+    )
+    assert len(weights_below) == 2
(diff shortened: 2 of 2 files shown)


optuna/samplers/_tpe/multi_objective_sampler.py
```diff
@@ -557,9 +557,10 @@ def _calculate_weights_below(
                     - self._compute_hypervolume(
                         np.asarray(lvals_below[:i] + lvals_below[i + 1 :]), reference_point
                     )
-                    for i in range(len(lvals))
+                    for i in range(n_below)
```
toshihikoyanase, line 560: Thank you for fixing the length of `contributions`.
I'm wondering why we need to pad last `len(lvals) - n_below` contributions with the value of `contributions[n_below-1]`.
Was it a simple typo?

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
```diff
@@ -567,6 +567,44 @@ def test_calculate_nondomination_rank() -> None:
     assert ranks == [0, 1, 0, 2, 0, 1]
 
 
+def test_calculate_weights_below() -> None:
+    sampler = MOTPESampler()
+
+    # Two samples
```
toshihikoyanase, line 573: [nit] Could you add periods to the end of the code comments? We usually use full sentences as shown in the [contribution guide](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#block-and-inline-comments-are-supposed-to-be-sentences).

toshihikoyanase (commented):
Thank you for your PR. The change follows the discussion in #2247, and it looks great to me.

I have a small comment on the code-comment style and a question about the change that is not mentioned in #2247.

optuna/samplers/_tpe/multi_objective_sampler.py
y0z: `calculate_weights_below` should return weights of which length is equal to the number of below observations.
Actually, it returns `np.asarray([])` when `n_below == 0` and `np.asarray([1.0])` when `n_below == 1`.
Therefore, I think that `len(lvals)` is a simple typo for `len(lvals_below)`, that is, `n_below`.

y0z: `_calculate_weights_below` should return weights of which length is equal to the number of below observations.
Actually, it returns `np.asarray([])` when `n_below == 0` and `return np.asarray([1.0])` when `n_below == 1`.
Therefore, I think that `len(lvals)` is a simple typo for `len(lvals_below)`, that is, n_below.
The reason why we have not noticed this bug is that `zip(weights_below, below)` only loops `min(len(weights_below), len(below)) so that no runtime error occurs.
https://github.com/optuna/optuna/blob/391f2802a589b308821498021f81115a87ad9b4f/optuna/samplers/_tpe/multi_objective_sampler.py#L297

hvy:
@HideakiImamura do you want to take a look at this PR?

optuna/samplers/_tpe/multi_objective_sampler.py
HideakiImamura: Sounds reasonable to fix `len(lvals)` to `n_below`.

HideakiImamura (approved):
Thanks for the fix. LGTM!

optuna/samplers/_tpe/multi_objective_sampler.py
toshihikoyanase: Thank you for your explanation. That makes sense.

toshihikoyanase (approved):

Issue #1721, opened 2020-08-26
[Tracker] Improve Tutorial


crcrpar:
This issue is not labeled "contribution-welcome", though, if you are interested in, don't be shy to contact me 😃

What I'm going to do in the next months as follows:

1. Align the contents of tutorials with the README's Key Features
2. Add https://pytorch.org/tutorials/beginner/deep_learning_60min_blitz.html -ish thing
2. Add more contents as "Recipes". Some of the contents are
    - User-defined pruners
    - User-defined Optuna Callbacks
    - Human-in-the-loop optimization with Hyperparameter Importance
    - `enqueue_trial`
    - Introduction of multi-objective optimization

crcrpar:
Follow up tasks

- merge https://github.com/crcrpar/optuna/tree/new-recipes once 1722 is landed.
- move this section https://optuna.readthedocs.io/en/stable/tutorial/002_configurations.html#note-on-the-number-of-parameters to somewhere like new "new to optuna" tutorials that might be the Englsih translation of @HideakiImamura 's great tutorials https://colab.research.google.com/github/pfnet-research/optuna-hands-on/blob/master/ja/01_Optuna_Quick_Start_ja.ipynb, https://colab.research.google.com/github/pfnet-research/optuna-hands-on/blob/master/ja/02_Optuna_Advanced_Tutorial_ja.ipynb.

crcrpar:
Due to the unexpected ReadTheDocs build failure thanks to timeout or weird memory consumption, migrating to LightGbm https://github.com/optuna/optuna/pull/2076 tentatively but I wanna use some other platform such as GitHub Pages to host kind of "cool" PyTorch tutorial. The build time will not be problematic with GitHub Pages because we can make full use of cache in our GitHub Actions workflow.

crcrpar:
There's one thing to do: Multi-objective. But I don't have any idea of better contents, except for using PyTorch.

toshihikoyanase:
> There's one thing to do: Multi-objective.

I guess we can use LightGBM instead of PyTorch to mitigate the build time.

The following is just an example of multi-objective optimization for LightGBM. Some metrics like precision and recall show trade-off relationships. Then, we can apply multi-objective optimization to maximize both precision and recall simultaneously.
Practically, users may maximize the AUC and tune the threshold to valance the trade-off relationship, but I think it is simple and somewhat meaningful as an example.

What do you think?

[...]

PR #2278, opened 2021-01-28
Add Multi-objective to recipe (tutorial)

crcrpar:
4599a30 is based on https://github.com/optuna/optuna/issues/1721#issuecomment-767970082.

related #1721

### DIFF
--- a/tutorial/20_recipes/002_multi_objective.py
+++ b/tutorial/20_recipes/002_multi_objective.py
@@ -1,0 +1,71 @@
+"""
+.. _multi_objective_lgbm:
+
+Multi-objective optimization example with LightGBM
+==================================================
+
+This tutorial is an example of multi-objective optimization with Optuna which optimizes
+precision and recall while it would be more common to use F-measure and (ROC-)AUC instead.
+
+As of `v2.4.0 <https://github.com/optuna/optuna/releases/tag/v2.4.0>`_, :class:`~optuna.multi_objective.MultiObjectiveStudy` is merged into :class:`~optuna.study.Study`.
+"""
+
+
+###################################################################################################
+# First of all, import requirements.
+import lightgbm as lgb
+import numpy
+from sklearn import datasets
+from sklearn.metrics import precision_score, recall_score
+from sklearn.model_selection import train_test_split
+
+import optuna
+
+
+###################################################################################################
+# In the ``objective`` below, we train a model using Breast cancer dataset before
+# calculating precision and recall.
+def objective(trial):
+    data, target = datasets.load_breast_cancer(return_X_y=True)
+    train_x, valid_x, train_y, valid_y = train_test_split(data, target, test_size=0.25)
+    dtrain = lgb.Dataset(train_x, label=train_y)
+
+    param = {
+        "objective": "binary",
+        "metric": "binary_logloss",
+        "verbosity": -1,
+        "boosting_type": "gbdt",
+        "lambda_l1": trial.suggest_float("lambda_l1", 1e-8, 10.0, log=True),
+        "lambda_l2": trial.suggest_float("lambda_l2", 1e-8, 10.0, log=True),
+        "num_leaves": trial.suggest_int("num_leaves", 2, 256),
+        "feature_fraction": trial.suggest_float("feature_fraction", 0.4, 1.0),
+        "bagging_fraction": trial.suggest_float("bagging_fraction", 0.4, 1.0),
+        "bagging_freq": trial.suggest_int("bagging_freq", 1, 7),
+        "min_child_samples": trial.suggest_int("min_child_samples", 5, 100),
+    }
+
+    gbm = lgb.train(param, dtrain)
+    preds = gbm.predict(valid_x)
+    pred_labels = numpy.rint(preds)
+    precision = precision_score(valid_y, pred_labels)
+    recall = recall_score(valid_y, pred_labels)
+    return precision, recall
+
+
+###################################################################################################
+# To make :class:`~optuna.study.Study` compatible with multi-objective optimization,
+# what we need to do is only to pass the collection of ``"direction"``\\s.
+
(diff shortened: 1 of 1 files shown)


hvy:
@himkt could you jump in as a reviewer here?

tutorial/20_recipes/002_multi_objective.py
```diff
@@ -0,0 +1,71 @@
+"""
+.. _multi_objective_lgbm:
+
+Multi-objective optimization example with LightGBM
+==================================================
+
+This tutorial is an example of multi-objective optimization with Optuna which optimizes
+precision and recall while it would be more common to use F-measure and (ROC-)AUC instead.
+
+As of `v2.4.0 <https://github.com/optuna/optuna/releases/tag/v2.4.0>`_, :class:`~optuna.multi_objective.MultiObjectiveStudy` is merged into :class:`~optuna.study.Study`.
```
hvy, line 10:
```suggestion
Starting from `v2.4.0 <https://github.com/optuna/optuna/releases/tag/v2.4.0>`_, :class:`~optuna.multi_objective.MultiObjectiveStudy` is made available with :class:`~optuna.study.Study`.
```

tutorial/20_recipes/002_multi_objective.py
```diff
@@ -0,0 +1,71 @@
+"""
+.. _multi_objective_lgbm:
+
+Multi-objective optimization example with LightGBM
+==================================================
+
+This tutorial is an example of multi-objective optimization with Optuna which optimizes
+precision and recall while it would be more common to use F-measure and (ROC-)AUC instead.
```
hvy, line 8:
```suggestion
precision and recall. In practice, it is more common to use F-measure and (ROC-)AUC.
```

tutorial/20_recipes/002_multi_objective.py
```diff
@@ -0,0 +1,71 @@
...
+import numpy
+from sklearn import datasets
+from sklearn.metrics import precision_score, recall_score
+from sklearn.model_selection import train_test_split
+
+import optuna
+
+
+###################################################################################################
+# In the ``objective`` below, we train a model using Breast cancer dataset before
```
hvy, line 26:
```suggestion
# In the ``objective`` below, we train a model using the Breast cancer dataset before
```

tutorial/20_recipes/002_multi_objective.py
```diff
@@ -0,0 +1,71 @@
...
+    preds = gbm.predict(valid_x)
+    pred_labels = numpy.rint(preds)
+    precision = precision_score(valid_y, pred_labels)
+    recall = recall_score(valid_y, pred_labels)
+    return precision, recall
+
+
+###################################################################################################
+# To make :class:`~optuna.study.Study` compatible with multi-objective optimization,
+# what we need to do is only to pass the collection of ``"direction"``\\s.
```
hvy, line 57: Seems like the backslash isn't escaped properly. Could you have a look? https://65701-122299416-gh.circle-artifacts.com/0/docs/build/html/tutorial/20_recipes/002_multi_objective.html#sphx-glr-tutorial-20-recipes-002-multi-objective-py

tutorial/20_recipes/002_multi_objective.py
```diff
@@ -0,0 +1,71 @@
...
+
+optuna.logging.get_logger("optuna").addHandler(logging.StreamHandler(sys.stdout))
+
+study = optuna.create_study(directions=("maximize", "maximize"))
+study.optimize(objective, n_trials=100)
+
+
+###################################################################################################
+# Optuna provides a visualization feature unique to multi-objective: :func:`~optuna.visualization.plot_pareto_front`.
+optuna.visualization.plot_pareto_front(study)
```
hvy, line 71: Just an idea but how about passing `target_names` with "precision" and "recall"? That might be a tiny bit more pedagogical.

tutorial/20_recipes/002_multi_objective.py
```diff
@@ -0,0 +1,71 @@
+"""
+.. _multi_objective_lgbm:
+
+Multi-objective optimization example with LightGBM
```
hvy, line 4: A nit (for consistency).
```suggestion
Multi-objective Optimization Example with LightGBM
```

hvy (changes requested):
Thanks for the great PR. It's very simple and I think it'd help users to quickly get started with MO. The changes mostly LGTM so I left smaller comments, mostly on the English phrasings.

I by the way think that it could be even further improved it we'd be explicit with the sampler and explain that some support MO while some don't. Maybe it's beyond the scope of this tutorial? Anyhow, I think it can be left outside this PR.

toshihikoyanase:
Just FYI: I update the example notebook using the UCI Bank Marketing dataset. Please refer to https://github.com/optuna/optuna/issues/1721#issuecomment-769759092.

tutorial/20_recipes/002_multi_objective.py
HideakiImamura: I think we just need to remove the backslash here.
```suggestion
# what we need to do is only to pass the collection of ``"direction"``s.
```

HideakiImamura (commented):
Thanks for the PR. The content is interesting and easy to follow. How about using the UCI Bank dataset as @toshihikoyanase suggested?

toshihikoyanase:
@HideakiImamura Thank you for your suggestion. I talked about changing the target library from lightgbm to PyTorch with @crcrpar. The corresponding PR is #2305. Readers may understand the PyTorch's scenario that optimizes flops and accuracy more easily than the current precision-recall optimization.

It will double the build time (3min to 6min in GitHub Actions), but it seems to keep the build time limitation of ReadTheDocs.

HideakiImamura:
Thanks for the explanation. So, should we close this PR?

toshihikoyanase:
I think so, but I'd like to wait for the author's response.

PR #2280, opened 2021-01-29
Remove specification of `n_trials` from example of `GridSampler`

not522:
## Motivation
`GridSampler` automatically stops the optimization if all combinations have already been evaluated. Therefore, the specification of `n_trials` is not necessary.
## Description of the changes
Remove the specification of `n_trials` from the example of `GridSampler`.


hvy (approved):
Good catch, and makes sense especially given the elaborations further down. LGTM!

hvy:
@c-bata would you mind double checking this simplification?

c-bata:
LGTM!

We can also refactor the following test as a follow-up task.

https://github.com/optuna/optuna/blob/e4b846c0e219fa64cf992b4af68a9437237e4f8a/tests/samplers_tests/test_grid.py#L48

c-bata (approved):

PR #2294, opened 2021-02-01
Use `trainer.callback_metrics` in the Pytorch-Lightning exapmle

TezRomacH:
## Motivation

1. In examples you can remove one extra class `MetricsCallback`. Pytorch-Lightning itself saves metrics states in `trainer.callback_metrics`.
1. In the `PyTorchLightningPruningCallback` we can add another `init` parameters for supper class and make sure we save inner states properly 

## Description of the changes

1. I remover `MetricsCallback` and used the native PL functions to achieve the same result.
2. I added `**kwargs` in init and added `super().on_validation_end(trainer=trainer, pl_module=pl_module)` in validation step

### DIFF
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -1,5 +1,3 @@
-from typing import Any
-
 import optuna
 
 
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -29,18 +29,21 @@ class PyTorchLightningPruningCallback(EarlyStopping):
             ``pytorch_lightning.LightningModule.training_step`` or
             ``pytorch_lightning.LightningModule.validation_epoch_end`` and the names thus depend on
             how this dictionary is formatted.
+        **kwargs: Additional kwargs for ``pytorch_lightning.callbacks.EarlyStopping``
     """
 
-    def __init__(self, trial: optuna.trial.Trial, monitor: str) -> None:
+    def __init__(self, trial: optuna.trial.Trial, monitor: str, **kwargs) -> None:
 
         _imports.check()
 
-        super(PyTorchLightningPruningCallback, self).__init__(monitor=monitor)
+        super(PyTorchLightningPruningCallback, self).__init__(monitor=monitor, **kwargs)
 
         self._trial = trial
 
     def on_validation_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
 
+        # To correct check if need to stop TPU and save train metrics
+        super().on_validation_end(trainer=trainer, pl_module=pl_module)
         logs = trainer.callback_metrics
         epoch = pl_module.current_epoch
         current_score = logs.get(self.monitor)
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -40,7 +40,7 @@ class PyTorchLightningPruningCallback(EarlyStopping):
             training will stop when the quantity
             monitored has stopped decreasing; in ``max``
             mode it will stop when the quantity
-            monitored has stopped increasing.
+            monitored has stopped increasing. Default: ``min``.
         strict: whether to crash the training if ``monitor`` is
             not found in the validation metrics. Default: ``True``.
     """
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -29,14 +29,43 @@ class PyTorchLightningPruningCallback(EarlyStopping):
             ``pytorch_lightning.LightningModule.training_step`` or
             ``pytorch_lightning.LightningModule.validation_epoch_end`` and the names thus depend on
             how this dictionary is formatted.
-        **kwargs: Additional kwargs for ``pytorch_lightning.callbacks.EarlyStopping``
+        min_delta: minimum change in the monitored quantity
+            to qualify as an improvement, i.e. an absolute
+            change of less than ``min_delta``, will count as no
+            improvement. Default: ``0.0``.
+        patience: number of validation epochs with no improvement
+            after which training will be stopped. Default: ``3``.
+        verbose: verbosity mode. Default: ``False``.
+        mode: one of {``min``, ``max``}. In ``min`` mode,
+            training will stop when the quantity
+            monitored has stopped decreasing; in ``max``
+            mode it will stop when the quantity
+            monitored has stopped increasing.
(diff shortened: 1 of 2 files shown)


TezRomacH:
Mypy check crashes on files that I didn't even change 🤔

nzw0301:
I think you need to add a type for ` **kwargs`:  `**kwargs: Any`.

TezRomacH:
Didn't work :( 

Screenshot 2021-02-01 at 18 02 50

nzw0301:
Thank you for your reply. On my local machine, it works... Could you push your update?  I think your `mypy` errors do not contain pytorch-lightning's error anymore.

HideakiImamura:
@TezRomacH What is a version of `mypy`? Currently, we fix the version of `mypy` to `0.790` in `setup.py`. Please use `mypy=0.790`.

HideakiImamura:
CI is failing with unnatural error.... I think there is nothing wrong with the PR change, so I will re-run it.

optuna/integration/pytorch_lightning.py
```diff
@@ -29,18 +31,21 @@ class PyTorchLightningPruningCallback(EarlyStopping):
             ``pytorch_lightning.LightningModule.training_step`` or
             ``pytorch_lightning.LightningModule.validation_epoch_end`` and the names thus depend on
             how this dictionary is formatted.
+        **kwargs: Additional kwargs for ``pytorch_lightning.callbacks.EarlyStopping``
```
HideakiImamura, line 34: IMO, this `**kwargs` is ambiguous. In Optuna, we have no this type of arguments of classes. How about adding each argument of the superclass if you need it. Which arguments do you want to add?

HideakiImamura (commented):
Thanks for the PR. I have a comment about `**kwargs`. Please take a look.

TezRomacH:
Hi, guys! What's next? :)

optuna/integration/pytorch_lightning.py
```diff
@@ -29,18 +29,50 @@ class PyTorchLightningPruningCallback(EarlyStopping):
...
+            change of less than ``min_delta``, will count as no
+            improvement. Default: ``0.0``.
+        patience: number of validation epochs with no improvement
+            after which training will be stopped. Default: ``3``.
+        verbose: verbosity mode. Default: ``False``.
+        mode: one of {``min``, ``max``}. In ``min`` mode,
+            training will stop when the quantity
+            monitored has stopped decreasing; in ``max``
+            mode it will stop when the quantity
+            monitored has stopped increasing.
```
nzw0301, line 43: Could you add the default value?

nzw0301 (changes requested):

optuna/integration/pytorch_lightning.py
TezRomacH, line 43:
```suggestion
            monitored has stopped increasing. Default: ``min``.
```

TezRomacH: Done :)

nzw0301 (approved):
Awesome!

nzw0301:
@HideakiImamura Could you take a look this PR?

TezRomacH:
👀

HideakiImamura (commented):
Sorry for the late reviews. I have a comment. PTAL.

HideakiImamura:
Ops. My review comment disappeared. Let me rewrite it.

optuna/integration/pytorch_lightning.py
```diff
@@ -29,18 +29,50 @@ class PyTorchLightningPruningCallback(EarlyStopping):
...
+        patience: number of validation epochs with no improvement
+            after which training will be stopped. Default: ``3``.
+        verbose: verbosity mode. Default: ``False``.
+        mode: one of {``min``, ``max``}. In ``min`` mode,
+            training will stop when the quantity
+            monitored has stopped decreasing; in ``max``
+            mode it will stop when the quantity
+            monitored has stopped increasing. Default: ``min``.
+        strict: whether to crash the training if ``monitor`` is
+            not found in the validation metrics. Default: ``True``.
```
HideakiImamura, line 45: In the original [`pytorch_lightning.callbacks.EarlyStopping`](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/callbacks/early_stopping.py), the stopping rule at the end of validation is Pytorch Lightning's own. However, in this `optuna.integration.PyTorchLightningPruningCallback`, the stopping rule is Optuna's own, which is specified by the pruner object in the `optuna.pruners` module. The stopping rule is implemented in `EarlyStopping.on_validation_end`, but it is overwritten in `PyTorchLightningPruningCallback`.

On the other hand, the stopping rule at the end of epochs is implemented in `EarlyStopping` but isn't in `PytorchLightningPruningCallback`. So, the same rule is applied to both of them. However, IMO, the `PytorchLightningPruningCallback` is designed to use Optuna's pruning feature in PyTorch Lightning, so we should not use it for other purposes. I think if we want to perform the early stopping at the end of epochs, we should define another callback as follows.
```python
callbacks = [
    PyTorchLightningPruningCallback(trial, monitor="val_acc"),
    pl.callbacks.EarlyStopping(monitor="val_acc", mode="max"),
]
```

Therefore, I think we should not add these parameters to `PyTorchLightningPruningCallback`.

HideakiImamura:
PTAL.

HideakiImamura:
[Just a note] As a follow-up PR, we may change the inheritance of `PyTorchLightningPruningCallback`. If anything, it might be better to inherit its parent class, the abstract class `Callback`, rather than `EarlyStopping`. Wed need a further investigation about this.

TezRomacH:
Okay, I agree that EarlyStopping does slightly different things. Let's not inherit from it then, shall we? :)

TezRomacH:
Let me change the inheritance to Callback in this PR?

HideakiImamura:
I think that one PR should do one thing. We need to investigate if we really want to change the inheritance, and since that task will be done separately as a follow-up, why not just use `trainer.callback_metrics`, which is what we originally wanted to do in this PR?

TezRomacH:
Agreed

TezRomacH:
Removed everything that doesn't apply to the example. I will create an issue to discuss inheritance.

HideakiImamura (approved):

I will participate in the discussion on whether we should change the inheritance of `PyTorchLightningPruningCallbacl`. Thanks for raising an issue too!

HideakiImamura:
Let me slightly change the title of this PR to represent the content of the PR explicitly.

PR #2267, opened 2021-01-26
Adds missing lightgbm tuner metrics for the case of higher is better

mavillan:
## Motivation
Missing [higher_is_better] metrics in LightGBMTuner.

## Description of the changes

When instantiating a LightGBMTuner with metrics `auc_mu` or `average_precision` and an study with direction equal to `maximize` you get this error:

```
ValueError: Study direction is inconsistent with the metric auc_mu. Please set 'minimize' as the direction.
```

I just added the missing metrics to the `self.higher_is_better` method of `LightGBMTuner`

PR #2299, opened 2021-02-02
Apply code-fix to `LightGBMTuner` (Follow-up #2267)

toshihikoyanase:
## Motivation

This is a follow-up PR for #2267. It applies a simple code-fix to LightGBMTuner.

## Description of the changes

It removes a local variable definition and reduces 7 lines.

### DIFF
--- a/optuna/integration/_lightgbm_tuner/optimize.py
+++ b/optuna/integration/_lightgbm_tuner/optimize.py
@@ -146,14 +146,7 @@ class _BaseTuner(object):
     def higher_is_better(self) -> bool:
 
         metric_name = self.lgbm_params.get("metric", "binary_logloss")
-        supported_higher_is_better = [
-            "auc",
-            "auc_mu",
-            "ndcg",
-            "map",
-            "average_precision",
-        ]
-        return metric_name in supported_higher_is_better
+        return metric_name in ("auc", "auc_mu", "ndcg", "map", "average_precision")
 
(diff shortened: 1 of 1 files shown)


hvy (approved):
Thanks, seems reasonable. LGTM!

Just wondering why we have a seemingly non-trivial fallback value `"binary_logloss"`. Its value is irrelevant here. 🤔

HideakiImamura (approved):
LGTM.

HideakiImamura:
I'll merge this PR after resolving @hvy's question.

toshihikoyanase:
> Just wondering why we have a seemingly non-trivial fallback value "binary_logloss". Its value is irrelevant here. 🤔

Thank you for pointing it out. I guess `binary_logloss` was a default value of `metric`, but it's supposed to set the metric corresponding to `objective` according to the [reference](https://lightgbm.readthedocs.io/en/latest/Parameters.html#metric). This issue should be fixed in a follow-up PR.

> "" (empty string or not specified) means that metric corresponding to specified objective will be used (this is possible only for pre-defined objective functions, otherwise no evaluation metric will be added)

HideakiImamura:
@hvy Could you merge this PR after checking @toshihikoyanase's comment?

hvy:
Sure, thanks for the heads up.

I haven't completely understood the `binary_logloss` behavior and although it would seem like a local code fix here, I now see that `self.lgbm_params.get("metric", "binary_logloss")` is used is several places.

Issue #2286, opened 2021-01-29
Multi-GPUs with optuna


Uproar197:
Hello, now I meet a problem, the overview of my code is shown in the following link:
https://github.com/Uproar197/Code/blob/main/my_optuna_problem.py

I can use Optuna to conduct experiments, but I don’t know why when the training epoch is relatively small, eg. **5 epochs**,  the experiment can be performed according to the given trial value and output; but when the number of the epoch is relatively large, eg. **250 epochs**,  no matter how big **n_trial** is,  the experiment comes the fifth trial, it will **be stuck** and cannot continue.

Another question is that when I use 5 GPUs, it will start 5 trials during the experiment, but it seems that these 5 trials are only equivalent to one trial in n_trials. I'm not sure why?
![image](https://user-images.githubusercontent.com/46192007/106249841-33298280-6256-11eb-89ba-dd4691d4aa82.png)

In addition, I output each training process, but there is a problem after using optuna, that is, the number of outputs will increase as the number of trials increases. For example, in the **first trial**, **'Epoch: [{}/{}] Iter:[{}/{}], Time: {:.2f}, lr: {:.6f}, Loss: {:.6f}'** will output once; but when it comes to the **third trial**,  **'Epoch: [{}/{}] Iter:[{}/{}], Time: {:.2f}, lr: {:.6f}, Loss: {:.6f}'** will output three times. Do you know the reason?
![image](https://user-images.githubusercontent.com/46192007/106250444-fdd16480-6256-11eb-908c-43c74a4d7942.png)
![image](https://user-images.githubusercontent.com/46192007/106250534-1c376000-6257-11eb-9abb-3ffc5472ff2c.png)

Thanks in advance!

nzw0301:
I would answer one of your questions for sure.

> Another question is that when I use 5 GPUs, it will start 5 trials during the experiment, but it seems that these 5 trials are only equivalent to one trial in n_trials. I'm not sure why?

This is because you use PyTorch's distributed data-parallel.

I think other questions are also related to distributed data-parallel training.

Uproar197:
Sorry for my late response. Maybe it is the case you answered. And is there anyone who knows why it will be stuck?

nzw0301:
I think that the deadlock of Sqlite is the reason that your code stops. The other databases, such as MySQL or PostgreSQL, can prevent your issue. Please see details in https://optuna.readthedocs.io/en/latest/tutorial/10_key_features/004_distributed.html

[...]

PR #2302, opened 2021-02-03
Example folders

Crissman:
## Motivation
Responding to issue #1238.

## Description of the changes
Create folders for each platform that has more than one example to reduce the number of files in the examples folder.

Note, on Merge, optuna.org and Medium posts that reference these examples should be changed.

More updating of URLs will be necessary with this PR. Posting first as WIP to confirm ok to make the directory changes.

PR #2303, opened 2021-02-03
Add integration module for PyTorch Distributed Data Parallel

toshihikoyanase:
## Motivation

This PR adds an integration module, namely `TorchDistributedTrial`, to optimize hyperparameters of neural networks written with [PyTorch Distributed Data Parallel](https://pytorch.org/docs/stable/distributed.html). 

It is related to #1765 and #2286.

## Description of the changes

I employed the same approach as `ChainerMNTrial`, which broadcasts suggested values from rank-0 node to the other nodes. 

To use `torch.distributed.broadcast`, I need to convert messages to `torch.tensor`, so I encode them using `pickle`. We may be able to use `json` instead of `pickle`, but we need to add preprocessing/postprocessing to handle `Trial.datetime_start` since `datetime.datetime` is not JSON serializable.

### Future work

- Provide `TorchDistributedStudy` which wraps conditional branch regarding `Study.optimize`.


HideakiImamura:
@nzw0301 Would you be interested in this PR? Since you seem to be familiar with the PyTorch Distributed Data Parallel, your comments and reviews would be very useful to Optuna.

optuna/integration/pytorch_distributed.py
```diff
@@ -0,0 +1,250 @@
...
+        details.
+
+    See `the example <https://github.com/optuna/optuna/blob/master/
+    examples/pytorch_distributed_simple.py>`__
+    if you want to optimize an objective function that trains neural network
+    written with PyTorch distributed data parallel.
+
+    Args:
+        trial:
+            A :class:`~optuna.trial.Trial` object or :obj:`None`. Please set trial objece in
```
nzw0301, line 39:
```suggestion
            A :class:`~optuna.trial.Trial` object or :obj:`None`. Please set trial object in
```

toshihikoyanase: Thanks!

examples/pytorch_distributed_simple.py
```diff
@@ -0,0 +1,184 @@
+"""
+Optuna example that optimizes multi-layer perceptrons using PyTorch distributed.
+
+In this example, we optimize the validation accuracy of hand-written digit recognition using
+PyTorch distributed data parallel and MNIST. We optimize the neural network architecture as well
+as the optimizer configuration. As it is too time consuming to use the whole MNIST dataset, we
+here use a small subset of it.
+
+You can execute this example with mpirun command as follows:
+    $ mpirun -n 2 -- python pytorch_distributed_simple.py
```
nzw0301, line 10: I'm not familiar with `mpirun`, but is it necessary `--`? Since I ran this example on a GPU server, the command did not work with `--`: 

```bash
[mpiexec@a089] match_arg (../../utils/args/args.c:254): unrecognized argument
[mpiexec@a089] HYDU_parse_array (../../utils/args/args.c:269): argument matching returned error
[mpiexec@a089] parse_args (../../ui/mpich/utils.c:4770): error parsing input array
[mpiexec@a089] HYD_uii_mpx_get_parameters (../../ui/mpich/utils.c:5106): unable to parse user arguments

Usage: ./mpiexec [global opts] [exec1 local opts] : [exec2 local opts] : ...

```

toshihikoyanase: The difference may come from the implementation of MPI.
I guess you used [MPICH](https://www.mpich.org/), but I and the CI job use [Open MPI](https://www.open-mpi.org/). I'll check the behavior with MPICH too. Thank you for your feedback.

https://github.com/optuna/optuna/blob/b4450efc8239f1d180c023c9cee92cf675932bee/.github/workflows/examples.yml#L36

nzw0301: I got it. Thank you for your answer!

examples/pytorch_distributed_simple.py
```diff
@@ -0,0 +1,184 @@
+"""
+Optuna example that optimizes multi-layer perceptrons using PyTorch distributed.
+
+In this example, we optimize the validation accuracy of hand-written digit recognition using
+PyTorch distributed data parallel and MNIST. We optimize the neural network architecture as well
+as the optimizer configuration. As it is too time consuming to use the whole MNIST dataset, we
+here use a small subset of it.
+
+You can execute this example with mpirun command as follows:
```
nzw0301, line 9: If we modify this code for GPUs, we might need to install pytorch from source instead of binaries because binary version does not support MPI. So I would like to see some comment to mention it.

Reference: [`MPI Backend` part in this page](https://pytorch.org/tutorials/intermediate/dist_tuto.html)

examples/pytorch_distributed_simple.py
toshihikoyanase: I reproduced the error with MPICH and updated the example for it (commit: https://github.com/optuna/optuna/pull/2303/commits/a95f239b4eebb1406730d0c7879e88fa66fd2846).

Please let me know if it does not work in your environment.

examples/pytorch_distributed_simple.py
toshihikoyanase: Thank you for your suggestion. I added a note for the MPI backend in the commit https://github.com/optuna/optuna/pull/2303/commits/3f153b273789186b32e6d559e493ad0e5480e45c.

In my environment, the following error occurred if I switched the backend without GPUs. 

```console
RuntimeError: Distributed package doesn't have MPI built in. MPI is only included if you build PyTorch from source on a host that has MPI installed.
```

So, I guess the error does not depend on the usage of GPUs, but it only depends on the MPI backend.

nzw0301: Awesome! 

I have tested it on GPUs, but I think so.

examples/pytorch_distributed_simple.py
nzw0301: It works now 👍

examples/pytorch_distributed_simple.py
```diff
@@ -0,0 +1,197 @@
...
+        correct = 0
+        with torch.no_grad():
+            for data, target in valid_loader:
+                data, target = data.view(data.size(0), -1).to(DEVICE), target.to(DEVICE)
+                output = model(data)
+                # Get the index of the max log-probability.
+                pred = output.argmax(dim=1, keepdim=True)
+                correct += pred.eq(target.view_as(pred)).sum().item()
+
+        correct_tensor = torch.tensor([correct], dtype=torch.int)
```
nzw0301, line 130:
```suggestion
        correct_tensor = torch.tensor([correct], dtype=torch.int).to(DEVICE)
```

examples/pytorch_distributed_simple.py
```diff
@@ -0,0 +1,197 @@
...
+        dist.all_reduce(correct_tensor)
+        total_correct = correct_tensor.numpy().tolist()[0]
```
nzw0301, line 132:
```suggestion
        total_correct = correct_tensor.item()
```

examples/pytorch_distributed_simple.py
nzw0301: We need to send this tensor to a `local_rank`'s device, especially for GPUs.

examples/pytorch_distributed_simple.py
nzw0301: I think that `.item()` is enough to access the value.

examples/pytorch_distributed_simple.py
toshihikoyanase: You're absolutely right. Thank you.

examples/pytorch_distributed_simple.py
toshihikoyanase: Thank you for your suggestion. We can apply the same change to `_call_and_communicate` and `_call_and_communicate_obj`, so I'll update them.

toshihikoyanase: Updated in https://github.com/optuna/optuna/pull/2303/commits/496a1d0098d223475b288381687c7f2e0b5842eb. According to the [reference of `.tolist()`](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.tolist), data will be moved from GPU to cpu automatically. I removed `to("cpu").numpy()` because it is redundant.

nzw0301: Nice!

HideakiImamura:
@sile Could you review this PR if you have time? Please feel free to remove the assignment if you are busy.

sile:
It's okay! Let me review this PR next week.

optuna/integration/pytorch_distributed.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def _call_and_communicate(func: Callable, dtype: "torch.dtype") -> Any:
+    buffer = torch.empty(1, dtype=dtype)
+    rank = dist.get_rank()
+    if rank == 0:
+        result = func()
+        buffer[0] = result
+    if dist.get_backend() == "nccl":
+        buffer = buffer.cuda(torch.device(rank))
```
HideakiImamura, line 216: [Just a note] It seems that we only test with the `gool` backend, so we have not confirmed this one in the unit-tests. However, since GitHub Actions does not provide GPU environments, it cannot be confirmed by unit tests.

HideakiImamura (commented):
Thanks for the great PR. Almost LGTM. I have several comments. Please take a look.

examples/pytorch_distributed_simple.py
```diff
@@ -0,0 +1,197 @@
...
+        print("  Number of complete trials: ", len(complete_trials))
+
+        print("Best trial:")
+        trial = study.best_trial
+
+        print("  Value: ", trial.value)
+
+        print("  Params: ")
+        for key, value in trial.params.items():
+            print("    {}: {}".format(key, value))
```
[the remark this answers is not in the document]
toshihikoyanase: That makes sense. Thank you for your suggestion. I applied it in https://github.com/optuna/optuna/pull/2303/commits/1517e9d885a3f17e070607903a82fdde3828baa5.

optuna/integration/pytorch_distributed.py
```diff
@@ -0,0 +1,249 @@
...
+    Args:
+        trial:
+            A :class:`~optuna.trial.Trial` object or :obj:`None`. Please set trial object in
+            rank-0 node and set :obj:`None` in the other rank node.
+    """
+
+    def __init__(self, trial: Optional[optuna.trial.Trial]) -> None:
+
+        _imports.check()
+        self.delegate = trial
```
sile, line 45: Because it seems the `delegate` field should be private, how about adding `_` prefix to the name?
```suggestion
        self._delegate = trial
```

optuna/integration/pytorch_distributed.py
```diff
@@ -0,0 +1,249 @@
...
+
+    Args:
+        trial:
+            A :class:`~optuna.trial.Trial` object or :obj:`None`. Please set trial object in
+            rank-0 node and set :obj:`None` in the other rank node.
+    """
+
+    def __init__(self, trial: Optional[optuna.trial.Trial]) -> None:
+
+        _imports.check()
```
sile, line 44: I'm not sure it's acceptable, but how about adding a check like the following to detect wrong usage of this class earlier?
```suggestion
        _imports.check()
        
        if dist.get_rank() == 0:
            assert trial is not None, "error message"
        else:
            assert trial is None, "error message"
```

optuna/integration/pytorch_distributed.py
```diff
@@ -0,0 +1,249 @@
...
+    def number(self) -> int:
+        def func() -> int:
+
+            assert self.delegate is not None
+            return self.delegate.number
+
+        return _call_and_communicate(func, torch.int)
+
+    @property
+    def trial_id(self) -> int:
```
sile, line 145: Do we still need to provide this API?

optuna/integration/pytorch_distributed.py
```diff
@@ -0,0 +1,249 @@
...
+
+        warnings.warn(
+            "The use of `TorchDistributedTrial.trial_id` is deprecated. "
+            "Please use `TorchDistributedTrial.number` instead.",
+            FutureWarning,
+        )
+        return self._trial_id
+
+    @property
+    def _trial_id(self) -> int:
```
sile, line 155: Ditto.

optuna/integration/pytorch_distributed.py
```diff
@@ -0,0 +1,249 @@
...
+def _call_and_communicate_obj(func: Callable) -> Any:
+    buffer = None
+    size_buffer = torch.empty(1, dtype=torch.int)
+    rank = dist.get_rank()
+    if rank == 0:
+        result = func()
+        buffer = _to_tensor(result)
+        size_buffer[0] = buffer.shape[0]
+    if dist.get_backend() == "nccl":
+        size_buffer = size_buffer.cuda(torch.device(rank))
```
sile, line 230: Cann't we just call `buffer.cuda()` without the argument `torch.device(rank)`?

I saw that [the PyTorch official doc](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.cuda) says the default value of the first argument is "the current CUDA device".
(This is a curious question as I am not familiar with PyTorch)

sile (commented):
Thanks for your PR. It looks very useful.
I left some review comments, so please check them when you have time.

optuna/integration/pytorch_distributed.py
toshihikoyanase: Thank you for your question.
In my environment, processes stalled if I didn't specify the rank to `.cuda`. I guess DDP assumes that each process uses its own device, and the device for the communication is required to be set explicitly.

On the other hand, I saw the following warning message when I removed the `device_ids` argument from `DDP.__init__`. It may suggest we can remove the rank from `.cuda()` if we set `CUDA_VISIBLE_DEVICES`. For example, we can set `CUDA_VISIBLE_DEVICE=RANK`. In this case, setting the rank to `.cuda` may cause a problem because only the device 0 is visible to each process.

> Single-Process Multi-GPU is not the recommended mode for DDP. In this mode, each DDP instance operates on multiple devices and creates multiple module replicas within one process. The overhead of scatter/gather and GIL contention in every forward pass can slow down training. Please consider using one DDP instance per device or per module replica by explicitly setting device_ids or CUDA_VISIBLE_DEVICES.

I'll investigate it in detail.

optuna/integration/pytorch_distributed.py
toshihikoyanase: I thought it was required by `BaseTrial`, but it wasn't. I removed it in https://github.com/optuna/optuna/pull/2303/commits/ac0002c1d21a058b6390a07c6c1c66c3be36f826. Thank you.

optuna/integration/pytorch_distributed.py
toshihikoyanase: I removed it in https://github.com/optuna/optuna/pull/2303/commits/ac0002c1d21a058b6390a07c6c1c66c3be36f826. Thanks!

optuna/integration/pytorch_distributed.py
toshihikoyanase: I investigated this issue and found two problems to set `rank` to the device.  To resolve these problems, I think the device should be selected by users, and I added the `device` argument to `TorchDistributedTrial` in the commit https://github.com/optuna/optuna/pull/2303/commits/700ea5dcd7d955bde4308357502918a1af8f3720.

### Assing a GPU to each process using `CUDA_VISIBLE_DEVICES`

When I use `CUDA_VISIBLE_DEVICES` to assign a GPU to each process, only the `cuda:0` can be seen by the process. So, `size_buffer.cuda(torch.device(rank))` raises no such device error. In this case, all nodes should be set `0` instead of `rank`.

[Full example code](https://gist.github.com/toshihikoyanase/21dc5dd6d899e5173e54f28f41e194bc#file-only_device_zero_is_visible-py)

### Multi-worker setting

When I use multiple GPU nodes, I needed to set the local rank to the device instead of the world rank. Suppose we have two nodes, and each node has two GPUs. The rank will be 0, 1, 2 and 3, but each node has device 0 and device 1. `TorchDistributedTrial(trial, device=local_rank)` resolved this problem.

[Full example code](https://gist.github.com/toshihikoyanase/21dc5dd6d899e5173e54f28f41e194bc#file-multinode-py)

optuna/integration/pytorch_distributed.py
toshihikoyanase: Thank you for your suggestion. That makes sense. I think we can raise `ValueError` for the invalid input values, and I changed the error type in the commit https://github.com/optuna/optuna/pull/2303/commits/2d49ceb88dcce930f898f074d1beeea6d1af19e6.

optuna/integration/pytorch_distributed.py
toshihikoyanase: Thanks! I applied your suggestion in the commit https://github.com/optuna/optuna/pull/2303/commits/9a11d2228ebb2e9955b319b4ee03875e76280617 and https://github.com/optuna/optuna/pull/2303/commits/bb2c2d37ad8f35f1a9cf7ed89dc929a46600ca7d.

optuna/integration/pytorch_distributed.py
sile: Thank you for your detailed investigation!
The above explanation is very helpful for me to understand DDP's behavior!
Adding the `device` argument seems to be a good solution for this issue 👍

optuna/integration/pytorch_distributed.py
```diff
@@ -0,0 +1,250 @@
...
+        def func() -> Optional[datetime]:
+
+            assert self._delegate is not None
+            return self._delegate.datetime_start
+
+        return _call_and_communicate_obj(func, self._device)
+
+
+def _call_and_communicate(
+    func: Callable, dtype: "torch.dtype", device: Optional["torch.device"] = None
```
sile, line 209: Since this function receives `device` arg, which is always `self._device` on the caller side, how about changing it a method (and removing the `device` arg)?
 (if there is no reason you prefer function rather than method)

toshihikoyanase: You're absolutely right. I don't have a specific reason to keep it as a function. I'll make `_call_and_communicate` and `_call_and_cummunicate_obj` methods of `TorchDistributedTrial`.

toshihikoyanase: I fixed it in the commit https://github.com/optuna/optuna/pull/2303/commits/a03b430cacbd948ee9444c5e6f3adf4a72412cf3.

sile (approved):
Since all my comments have been addressed (thanks!) and it seems quite nice now, let me approve this PR.

optuna/integration/pytorch_distributed.py
```diff
@@ -0,0 +1,246 @@
...
+        # torch.bool seems to be the correct type, but the communication fails
+        # due to the RuntimeError.
+        return self._call_and_communicate(func, torch.uint8)
+
+    def set_user_attr(self, key: str, value: Any) -> None:
+
+        if dist.get_rank() == 0:
+            assert self._delegate is not None
+            self._delegate.set_user_attr(key, value)
+        dist.barrier()
```
not522, line 144: When only partial nodes call this method (e.g., calling from only the rank 0 node), execution stops silently. I recommend mentioning this behavior in the document. I think it is useful to add a `barrier` argument to control whether to synchronize.

tests/integration_tests/test_pytorch_distributed.py
```diff
@@ -0,0 +1,290 @@
...
+def test_suggest_float(storage_mode: str) -> None:
+    with StorageSupplier(storage_mode) as storage:
+        if dist.get_rank() == 0:
+            study = optuna.create_study(storage=storage)
+            trial = TorchDistributedTrial(study.ask())
+        else:
+            trial = TorchDistributedTrial(None)
+
+        x1 = trial.suggest_float("x", 0, 1)
+        assert 0 <= x1 < 1
```
HideakiImamura, line 61: Currently, the `UniformDistribution` contains the `high`, so whether or not the equality is valid at the right end depends on the sampler. To avoid future bugs, how about putting an equal at the right end? 
```suggestion
        assert 0 <= x1 <= 1
```

This is the same as the other assertions.

toshihikoyanase: Thanks! I applied the change in https://github.com/optuna/optuna/pull/2303/commits/f7b6678c82a5fb1414f8a192d4238a13316a30aa.

optuna/integration/pytorch_distributed.py
toshihikoyanase: Thank you for your suggestion. I added a note to call the methods from all workers in the commit https://github.com/optuna/optuna/pull/2303/commits/eea96c85f8f4321d64538cda32f028278a94c67d.

> I think it is useful to add a barrier argument to control whether to synchronize.

When I used `ChainerMNTrial`, I sometimes got errors without `barrier` due to the timing issue. So, I'm a bit conservative to add an option to switch `barrier`. I'd like to create a new PR if I work on it.

not522: I agree to use `barrier` as default behavior, but users may want to write code like below.
```python
if dist.get_rank() == 0:
    trial.set_user_attr(..., barrier=False)
dist.barrier()
```
In this case, the synchronization is done by users. I think it is reasonable to discuss this feature in another issue or PR.

optuna/integration/pytorch_distributed.py
```diff
@@ -0,0 +1,252 @@
...
+            A :class:`~optuna.trial.Trial` object or :obj:`None`. Please set trial object in
+            rank-0 node and set :obj:`None` in the other rank node.
+        device:
+            A `torch.device` to communicate with the other nodes. Please set a CUDA device
+            assigned to the current node if you use "nccl" as `torch.distributed` backend.
+
+    .. note::
+        The methods of :class:`~optuna.integration.TorchDistributedTrial` are expected to be
+        called by all workers at once. They invoke synchronous data transmission to share
+        processing results and synchronize timinig.
```
not522, line 46:
```suggestion
        processing results and synchronize timing.
```

not522 (approved):
LGTM except a typo.

optuna/integration/pytorch_distributed.py
toshihikoyanase: Oops! I applied the suggestion in https://github.com/optuna/optuna/pull/2303/commits/98ed9de6a948462fb65651ab43af1e57cc99706d. Thank you!

tests/integration_tests/test_pytorch_distributed.py
```diff
@@ -0,0 +1,290 @@
...
+def test_suggest_uniform(storage_mode: str) -> None:
+    with StorageSupplier(storage_mode) as storage:
+        if dist.get_rank() == 0:
+            study = optuna.create_study(storage=storage)
+            trial = TorchDistributedTrial(study.ask())
+        else:
+            trial = TorchDistributedTrial(None)
+
+        x1 = trial.suggest_uniform("x", 0, 1)
+        assert 0 <= x1 < 1
```
HideakiImamura, line 78: Same reason as above.
```suggestion
        assert 0 <= x1 <= 1
```

tests/integration_tests/test_pytorch_distributed.py
```diff
@@ -0,0 +1,290 @@
...
+def test_suggest_loguniform(storage_mode: str) -> None:
+    with StorageSupplier(storage_mode) as storage:
+        if dist.get_rank() == 0:
+            study = optuna.create_study(storage=storage)
+            trial = TorchDistributedTrial(study.ask())
+        else:
+            trial = TorchDistributedTrial(None)
+
+        x1 = trial.suggest_loguniform("x", 1e-7, 1)
+        assert 1e-7 <= x1 < 1
```
HideakiImamura, line 95: Ditto.
```suggestion
        assert 1e-7 <= x1 <= 1
```

tests/integration_tests/test_pytorch_distributed.py
```diff
@@ -0,0 +1,290 @@
...
+            trial = TorchDistributedTrial(study.ask())
+        else:
+            trial = TorchDistributedTrial(None)
+
+        trial.suggest_float("f", 0, 1)
+        trial.suggest_int("i", 0, 1)
+        trial.suggest_categorical("c", ("a", "b", "c"))
+
+        params = trial.params
+        assert 0 <= params["f"] < 1
```
HideakiImamura, line 262: Ditto.
```suggestion
        assert 0 <= params["f"] <= 1
```

HideakiImamura (commented):
I found several points to be fixed. Please take a look.

toshihikoyanase:
@HideakiImamura Thank you for your careful reviews. I applied your suggestions in https://github.com/optuna/optuna/pull/2303/commits/33a4673a476ca2d30bed4ab3d0290a8602b1c950.

HideakiImamura (approved):
Thanks for your quick fix! LGTM!

optuna/integration/pytorch_distributed.py
not522: Similar to the above comment, if an exception is raised in `self._delegate.set_user_attr`, execution silently stops. It is better to catch the exception and broadcast it to all nodes.

not522 (changes requested):

optuna/integration/pytorch_distributed.py
toshihikoyanase: Thanks. I added the code to broadcast exceptions to all nodes and unit tests in the commit https://github.com/optuna/optuna/pull/2303/commits/39070bef147e03ebc9f8c6e04a8662ac6c223309. PTAL.

nzw0301:
I confirm that this example with either 2 or 4 GPUs on one server with the `nccl` backend performs correctly. 🚀 

Modified code and imperfect benchmark are https://gist.github.com/nzw0301/9ba3c837a02539e8bde194f4f2c9d0e5

optuna/integration/pytorch_distributed.py
not522: LGTM!

not522 (approved):
LGTM! I confirmed the example works in my environment.

HideakiImamura:
@nzw0301 Thanks for the confirmation and reviews. Could you approve this PR if you agree with merging this PR?

not522:
#2302 introduced `examples/pytorch` folder.
@toshihikoyanase Could you move `pytorch_distributed_simple.py` to the folder?

toshihikoyanase:
@not522 Thanks! I moved the example to the `example/pytorch` directory.

optuna/integration/pytorch_distributed.py
```diff
@@ -0,0 +1,281 @@
...
+class TorchDistributedTrial(optuna.trial.BaseTrial):
+    """A wrapper of :class:`~optuna.trial.Trial` to incorporate Optuna with PyTorch distributed.
+
+    .. seealso::
+        :class:`~optuna.integration.TorchDistributedTrial` provides the same interface as
+        :class:`~optuna.trial.Trial`. Please refer to :class:`optuna.trial.Trial` for further
+        details.
+
+    See `the example <https://github.com/optuna/optuna/blob/master/
+    examples/pytorch_distributed_simple.py>`__
```
not522, line 31: Please update here, too.

toshihikoyanase: Oops! Thank you for your careful review. I fixed it in https://github.com/optuna/optuna/pull/2303/commits/6976083a8659327629046ae2b994e865be62cd03.

nzw0301 (approved):

I8dNLo:
Is Lightning compatibility planned?

nzw0301:
@I8dNLo 

> Is Lightning compatibility planned?

This PR was not for pytorch-lightning. PyTorch-lightning's counterpart is defined in https://github.com/optuna/optuna-integration/blob/main/optuna_integration/pytorch_lightning.py.

PR #2304, opened 2021-02-03
[RFC] Add an option to the `Study.optimize` method that allows define-and-run parameter suggestion. 

g-votte:
This PR depends on #2271; so, please review and merge after that.

## Motivation
Similar to #2271, this PR adds `fixed_distributions` option to the `optimize` method so that the user can specify the search space in the define-and-run style.

## Example
```python
import optuna

def objective(trial):
    # The user can access to the parameters just after the trial initialization.
    print(trial.params)  

    # Also, the user can suggest additional parameters in the define-by-run style.
    n_layers = trial.suggest_int("n_layers", 1, 4)

    loss = train_model(trial.params["optimizer"], trial.params["lr"], n_layers)

    return loss

study = optuna.create_study()

# You can pass the distributions previously defined.
distributions = {
    "optimizer": optuna.distributions.CategoricalDistribution(["adam", "sgd"]),
    "lr": optuna.distributions.LogUniformDistribution(0.0001, 0.1),
}
study.optimize(objective, fixed_distributions=distributions, n_trials=1)

# `optimizer`, `lr`, and `n_layers` are suggested in the trial.
print(study.trials[0].params)
```


g-votte:
Closing this PR so far. Let’s revisit this feature once we get enough feedback of #2271 or requests from users.

Issue #2315, opened 2021-02-05
Change inheritance of PytorchLightningPruningCallback


TezRomacH:
## Motivation

PyTorchLightningPruningCallback class is inherited from `pl.EarlyStopping`.
But EarlyStopping itself determines when to stop training, without using data from Trial. Is it necessary to keep such a construct or is it enough to inherit from `pl.Callback`?

## Additional context (optional)

Previous discussion was here: #2294

himkt:
Thank you @TezRomacH for reporting the problem (#2294) and working on fixing it (#2326).
Let me close this issue because the problem seems to be fixed now. (cc. @toshihikoyanase, @HideakiImamura)

PR #2326, opened 2021-02-07
PyTorchLightningPruningCallback from Callback

TezRomacH:
## Motivation

PoC for #2315
Now PyTorchLightningPruningCallback is just a regular Callback instance


toshihikoyanase:
@nzw0301 @HideakiImamura Could you review this PR since you reviewed #2294? If you don't have time enough, please feel free to remove the assignment.

nzw0301:
Sorry, i'm busy this week...

examples/pytorch_lightning_simple.py
```diff
@@ -81,58 +68,80 @@ def forward(self, data):
...
 
-    def forward(self, data):
+        # We optimize the number of layers, hidden units in each layer and dropouts.
+        n_layers = trial.suggest_int("n_layers", 1, 3)
+        dropout = trial.suggest_float("dropout", 0.2, 0.5)
+        output_dims = [
+            trial.suggest_int("n_units_l{}".format(i), 4, 128, log=True) for i in range(n_layers)
+        ]
+
+        self.model = Net(dropout, output_dims)
```
crcrpar, line 79: I think we can define `self.model` using `torch.nn.Sequential`, which will be more concise.
Could you kindly do it?

examples/pytorch_lightning_simple.py
```diff
@@ -81,58 +68,80 @@ def forward(self, data):
...
 
     trainer = pl.Trainer(
         logger=False,
         limit_val_batches=PERCENT_VALID_EXAMPLES,
-        checkpoint_callback=checkpoint_callback,
+        checkpoint_callback=False,
         max_epochs=EPOCHS,
-        gpus=1 if torch.cuda.is_available() else None,
-        callbacks=[PyTorchLightningPruningCallback(trial, monitor="val_acc", mode="max")],
+        gpus=-1 if torch.cuda.is_available() else None,
```
crcrpar, line 141: I think it's better to let this program use a GPU if available. We can disable the use of GPUs outside the program by using `CUDA_VISIBLE_DEVICES`.
```suggestion
        gpus=1 if torch.cuda.is_available() else None,
```

crcrpar (commented):
Thank you for the work.

It basically is cool 🙂

examples/pytorch_lightning_simple.py
TezRomacH: Setting `-1` in PL says that you use all available GPUs :)
https://pytorch-lightning.readthedocs.io/en/stable/trainer.html#gpus
Screenshot 2021-02-08 at 16 09 21

examples/pytorch_lightning_simple.py
TezRomacH: I thought for myself that it was better for the users to see that their model can be any `torch.nn.Module`. In Lightning we create model and experimenting with Trial.

Are you proposing to build a model inside Lightning?

HideakiImamura (commented):
Thanks for the PR. It looks like there are a lot of changes being made that have nothing to do with the original purpose of the PR; I think one PR should do one thing, so can you revert the unrelated changes?

TezRomacH:
> Thanks for the PR. It looks like there are a lot of changes being made that have nothing to do with the original purpose of the PR; I think one PR should do one thing, so can you revert the unrelated changes?

You're suggesting to separate updated example and updated callback into to two PRs?

HideakiImamura:
Yes, I think we should separate them. Furthermore, I doubt that the changes being made in this PR on the example are necessary. For example, Optuna's example does not have type hinting, nor is it checked by CI.

examples/pytorch_lightning_simple.py
crcrpar: Oh, thank you. Obviously I missed that.

HideakiImamura:
Sorry for the confusion. I had a discussion with a team member about type hinting in `examples`. It's OK that examples may or may not have type hinting. If you could make a separate PR for changes to examples and changes to integration, that would be much appreciated.

TezRomacH:
Done :)

TezRomacH:
Example is here https://github.com/optuna/optuna/pull/2332

HideakiImamura (approved):
Thanks for the simplification! LGTM. I'll review another PR.

crcrpar (approved):
looks great :)

PR #2332, opened 2021-02-09
Updated PL example with typing and DataModule

TezRomacH:
## Motivation

New example for #2326

## Description of the changes

- Added typing
- Removed unused logging
- Updated code with pytorch-lightning DataModules


examples/pytorch_lightning_simple.py
```diff
@@ -81,58 +68,80 @@ def forward(self, data):
...
 
-    def forward(self, data):
+        # We optimize the number of layers, hidden units in each layer and dropouts.
+        n_layers = trial.suggest_int("n_layers", 1, 3)
+        dropout = trial.suggest_float("dropout", 0.2, 0.5)
+        output_dims = [
+            trial.suggest_int("n_units_l{}".format(i), 4, 128, log=True) for i in range(n_layers)
+        ]
+
+        self.model = Net(dropout, output_dims)
```
crcrpar, line 79: > Are you proposing to build a model inside Lightning?
> (Quote from the thread of the previous pull request of https://github.com/optuna/optuna/pull/2326#discussion_r572037238)

No, I'm not and sorry for putting my comment incongruously. I just thought it'd be more concise to use `torch.nn.Sequential` likewise https://github.com/optuna/optuna/blob/master/examples/kubernetes/mlflow/pytorch_lightning_distributed.py#L34  instead of defining custom `torch.nn.Module`. However, this is completely optional.

TezRomacH: I've just updated `class Net` implementation, thanks! 

About distributed example:
I' haven't seen this lol. Should I simplify this as well?

TezRomacH:
Guys, is PR okay?

crcrpar (approved):

TezRomacH:
@HideakiImamura ?

examples/pytorch_lightning_simple.py
```diff
@@ -37,102 +38,106 @@
...
+def objective(trial: optuna.trial.Trial) -> float:
+
+    model = LightningNet(trial)
+    datamodule = MNISTDataModule(data_dir=DIR, batch_size=BATCHSIZE)
 
     trainer = pl.Trainer(
         logger=False,
         limit_val_batches=PERCENT_VALID_EXAMPLES,
-        checkpoint_callback=checkpoint_callback,
+        checkpoint_callback=False,
```
HideakiImamura, line 135: According to the [documentation](https://pytorch-lightning.readthedocs.io/en/stable/trainer.html#trainer-class-api), the method of passing a `ModelCheckpoint` instance to `checkpoint_callback` is deprecated, so passing a bool value seems to be a good idea. However, is it ok to use False here? Optuna creates a `pl.Trainer` for each trial, so if a user wants to use a checkpoint, it needs to be saved for each trial. To let users know this, shouldn't this value be set to True so that a different checkpoint is saved for each trial?

HideakiImamura (commented):
Sorry for the late review. Thanks for the great simplification! I have a question about checkpointintg. Could you take a look?

examples/pytorch_lightning_simple.py
TezRomacH: Hi, sorry for the late response.

I think is might be not such good but i used the same logic as in previous version.

```diff
- shutil.rmtree(MODEL_DIR)
```
https://github.com/optuna/optuna/pull/2332/files#diff-318c96a1dc252cea529ff2239e3001338de49f9f54e2a73193507c45aaa9b317L167

examples/pytorch_lightning_simple.py
TezRomacH: I thought it seemed like a really good idea to save the logs. I'll make a commit now.

examples/pytorch_lightning_simple.py
TezRomacH: I thought it seemed like a really good idea to save the logs. I'll make a commit now.

TezRomacH: Also I have enabled hyperparameter logging
Screenshot 2021-02-18 at 23 16 13

TezRomacH:
Hi! How is it going? :)

HideakiImamura:
Sorry for the late review. I'm checking your update :)

examples/pytorch_lightning_simple.py
HideakiImamura: Thanks for your response. I agree with you. Let me resolve this conversation.

HideakiImamura (approved):
LGTM!

Issue #2334, opened 2021-02-10
Tensorboard integration breaks when using suggest_float with step parameter


SammyRamone:
The tensorboard integration throws an error if the `trial.suggest_float()` method is used with a `step` parameter. If no step parameter is given, it works fine.

## Expected behavior
Should not throw an error.

## Environment

- Optuna version: 2.5.0
- Python version: 3.6
- OS: Ubuntu 18.04
- (Optional) Other libraries and their versions: tensorboard 2.4.1, tensorflow 2.4.1

## Error messages, stack traces, or logs

Traceback (most recent call last):
  File "test.py", line 43, in <module>
    study.optimize(objective, n_trials=10, timeout=600, callbacks=[tensorboard_callback])
  File "/home/marc/.local/lib/python3.6/site-packages/optuna/study.py", line 385, in optimize
    show_progress_bar=show_progress_bar,
  File "/home/marc/.local/lib/python3.6/site-packages/optuna/_optimize.py", line 73, in _optimize
    progress_bar=progress_bar,
  File "/home/marc/.local/lib/python3.6/site-packages/optuna/_optimize.py", line 178, in _optimize_sequential
    callback(study, frozen_trial)
  File "/home/marc/.local/lib/python3.6/site-packages/optuna/integration/tensorboard.py", line 41, in __call__
    self._initialization(study)
  File "/home/marc/.local/lib/python3.6/site-packages/optuna/integration/tensorboard.py", line 102, in _initialization
    self._add_distributions(trial.distributions)
  File "/home/marc/.local/lib/python3.6/site-packages/optuna/integration/tensorboard.py", line 70, in _add_distributions
    param_name, hp.Discrete(param_distribution.low, param_distribution.high)
  File "/home/marc/.local/lib/python3.6/site-packages/tensorboard/plugins/hparams/summary_v2.py", line 500, in __init__
    self._values = list(values)
TypeError: 'float' object is not iterable

## Steps to reproduce

1. Use example from here https://github.com/optuna/optuna/blob/master/examples/tensorboard_simple.py
2. Change line 33 to `dropout_rate = trial.suggest_float("DROPOUT_RATE", 0.1, 0.2, step=0.001)`
3. Run and error is thrown
4. Remove `step=0.001`
5. Run and no error is thrown

nzw0301:
[...]

PR #2335, opened 2021-02-11
Fix TensorBoard integration for `suggest_float`

nzw0301:
## Motivation

The current tensorboard integration does not work for hyper-parameters sampled by using `suggest_float` with `step` argument.

See https://github.com/optuna/optuna/issues/2334 for the reproducible code and the reported issue.

## Description of the changes 

- Replace `hp.Discrete` with `hp.RealInterval`.
  - Fix the argument of `hp.Discrete` is another option, but it takes an iterable objective as the argument, so I believe `RealInterval` is simpler.
- Add a test for `suggest_float`

### DIFF
--- a/optuna/integration/tensorboard.py
+++ b/optuna/integration/tensorboard.py
@@ -67,7 +67,7 @@ class TensorBoardCallback(object):
                 )
             elif isinstance(param_distribution, optuna.distributions.DiscreteUniformDistribution):
                 self._hp_params[param_name] = hp.HParam(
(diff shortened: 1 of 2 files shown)


himkt (approved):
Thank you @nzw0301 for sending a patch promptly.

The suggestion that it uses `ReadInterval` instead of `Discrete` with explicitly generating all candidates looks reasonable to me. I want other opinions because there could be some diverge concerns around here.

HideakiImamura (approved):
Looks perfect! LGTM.

nzw0301:
Thanks!

PR #2343, opened 2021-02-13
Add type conversions for upper/lower whose values are integers

himkt:
Resolve #2337.

## Motivation
This PR introduces type conversions for `distribution.high` and `distribution.low` where a distribution is either `UniformDistribution` or `LogUniformDistribution`. Without these conversions, `hp.RealInterval` raises an error reported in #2337.

I think this problem could also be solved by adding type conversions to (a) initializers of `UniformDistribution` and `LogUniformDistribution` or (b) `suggest_float`. I'd like other developer's opinions/suggestions for solving this problem.

https://github.com/optuna/optuna/blob/183bcf6d391325ac9574d7d7ff0d53689a0b9e26/optuna/distributions.py#L121-L122
https://github.com/optuna/optuna/blob/183bcf6d391325ac9574d7d7ff0d53689a0b9e26/optuna/distributions.py#L165-L166
https://github.com/optuna/optuna/blob/183bcf6d391325ac9574d7d7ff0d53689a0b9e26/optuna/trial/_trial.py#L169-L178)

## Description of the changes
- Added type conversion in `_add_distributions`
- Added the test

PR #2360, opened 2021-02-17
Convert low and high to float explicitly in distributions

keisuke-umezawa:
## Motivation

To follow up #2343.
Avoiding distributions from having int low and high.


HideakiImamura:
@himkt Could you review this PR since you are involved in #2343?

HideakiImamura (approved):
LGTM!

himkt (approved):

Issue #2356, opened 2021-02-16
examples/kubernetes `check_study.sh` should match whole words


twolffpiggott:
In the [Distributed Optimisation on Kubernetes](https://github.com/optuna/optuna/tree/master/examples/kubernetes) example, the `check_study` script is intended to ensure that a study exists before allowing the Optuna workers to start optimisation.

Since it uses `grep PATTERN` instead of `grep -w PATTERN`, it can succeed when a study that only has "k8s_mlflow" as a substring exists e.g. "foo_k8s_mlflow" but the study "k8s_mlflow" does not exist. This IMO is unexpected behaviour and can allow the workers to start without the expected study existing.

https://github.com/optuna/optuna/blob/7cd4fa3c764b5ef8c0e9d14096ebe2453ba3b651/examples/kubernetes/mlflow/check_study.sh#L2

I would suggest the very minor update of using `grep -w PATTERN` to ensure the expected study exists and prevent matching substrings.

hvy:
Makes sense, thanks a lot for the report. Would you like to send a PR? Otherwise, I could label this as `contribution-welcome`.

twolffpiggott:
Great, no problem- I'd love to make a PR :)

hvy:
Oh, that'd be great.

himkt:
Let me close this issue, as #2363 was merged. Thank you @twolffpiggott for your great contribution!

PR #2363, opened 2021-02-17
Update `examples/kubernetes/mlflow/check_study.sh` to match whole words

twolffpiggott:
## Motivation
Follows from issue #2356. The previous implementation of `check_study.sh` could succeed when another study with the desired study name as a substring existed but the desired study did not exist. This update matches on whole words and correctly ensures the existence of the desired study.

## Description of the changes
Minor change from the use of `grep PATTERN` in `check_study.sh` to `grep -w PATTERN`.

https://github.com/optuna/optuna/blob/7cd4fa3c764b5ef8c0e9d14096ebe2453ba3b651/examples/kubernetes/mlflow/check_study.sh#L2


hvy (approved):
Thanks for the quick action, that LGTM!

toshihikoyanase (approved):
Looks good to me. Thank you for your contribution.

Issue #2324, opened 2021-02-07
`model_dir` of `LightGBMTunerCV` is not generated.


nzw0301:
## Expected behavior 

According to [`LightGBMTuner` ](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.lightgbm.LightGBMTuner.html#optuna.integration.lightgbm.LightGBMTuner), if `model_dir` does not exist, optuna creates the directory, and saves boosters into the directory. I can confirm that I just add `model_dir=` (say `model_dir="./boosters"`) to the argument of `.train()` in [this example](https://github.com/optuna/optuna/blob/533f081d33817f2a9a37b21916eb5f03a0fc13c2/examples/lightgbm_tuner_simple.py#L30).

> `model_dir` – A directory to save boosters. By default, it is set to None and no boosters are saved. Please set shared directory (e.g., directories on NFS) if you want to access `get_best_booster()` in distributed environments. Otherwise, it may raise ValueError. If the directory does not exist, it will be created. The filenames of the boosters will be `{model_dir}/{trial_number}.pkl` (e.g., `./boosters/0.pkl`).

However, when I add the same argument to `LightGBMTunerCV` in [this example](https://github.com/optuna/optuna/blob/533f081d33817f2a9a37b21916eb5f03a0fc13c2/examples/lightgbm_tuner_cv.py#L25), the directory is not generated. We (at least I) do not expect this inconsistent behavior.

I also replace `train` with `run` and set `model_dir` as in `LightGBMTunerCV`'s case I described for `LightGBMTuner`, and it worked well:

```python
tuner = lgb.LightGBMTuner(params, dtrain, valid_sets=[dtrain, dval], verbose_eval=100, early_stopping_rounds=100,
                          model_dir="./best_boosters",)
tuner.run()
```

## Environment

- Optuna version: 533f081
- Python version: 3.8.6
- OS: MacOS Catalina
- (Optional) Other libraries and their versions:
    - lightgbm: 3.1.1

## Reproducible example

```python
import sklearn.datasets
from sklearn.model_selection import KFold

import optuna.integration.lightgbm as lgb


if __name__ == "__main__":
    data, target = sklearn.datasets.load_breast_cancer(return_X_y=True)
    dtrain = lgb.Dataset(data, label=target)

    params = {
        "objective": "binary",
        "metric": "binary_logloss",
        "verbosity": -1,
        "boosting_type": "gbdt",
    }

[...]

PR #2366, opened 2021-02-20
Fix - make model_dir by _LightGBMBaseTuner

nyanhi:
## Motivation
To address the issue #2324 .

## Description of the changes
Moved following code, which makes the model directory, to `_LightGBMBaseTuner.__init__()` so that it behaves consistently across the subclasses (i.e. `LightGBMTuner` and `LightGBMTunerCV`).
Also added descriptions of `model_dir` and `return_cvbooster`.

```
        if self._model_dir is not None and not os.path.exists(self._model_dir):
            os.mkdir(self._model_dir)
```


optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -955,9 +967,9 @@ def __init__(
         sample_size: Optional[int] = None,
         study: Optional[optuna.study.Study] = None,
         optuna_callbacks: Optional[List[Callable[[Study, FrozenTrial], None]]] = None,
+        model_dir: Optional[str] = None,
```
crcrpar, line 970: Could you tell me why `model_dir` is moved here?

crcrpar (commented):
Thank you for the patch!

Do you have any idea of how to check the `model_dir` is expectedly made in a unittest?

keisuke-umezawa:
@toshihikoyanase Could you also review this PR?

optuna/integration/_lightgbm_tuner/optimize.py
nyanhi: Sure, this cosmetic change was made so that arguments of `LightGBMTunerCV` is ordered similarly with `LightGBMTuner`. Please kindly let me know if any concern, happy to hear your opinion.

Let me think a bit for the unittest for making directory.

crcrpar: > this cosmetic change was made so that arguments of LightGBMTunerCV is ordered similarly with LightGBMTuner

I didn't notice. Thank you for pointing it out and addressing it.

> Please kindly let me know if any concern, happy to hear your opinion.

Meanwhile,  leave as it is.

My take: I found it a little bit too drastic because this change breaks backward compatibility.
However, as the number of arguments of `LightGBMTuner.__init__` is large, so maybe we can assume most users pass arguments with the corresponding keywords.
So, I'd say this change comes with a certain amount of risk which can be negligible.

nyanhi: Understood, I will cancel the order change. Thank you for explanation!

HideakiImamura:
@crcrpar Let me assign you since you already reviewed this PR.

crcrpar (approved):
Thank you for the work, including the test of mkdir :)

tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
```diff
@@ -647,6 +647,20 @@ def test_get_best_booster(self) -> None:
...
+        params: Dict = {"verbose": -1}
+        dataset = lgb.Dataset(np.zeros((10, 10)))
+
+        study = optuna.create_study()
+
+        with mock.patch("optuna.integration._lightgbm_tuner.optimize.os.mkdir") as m:
+            with mock.patch("os.path.exists", return_value=mock_exists):
+                LightGBMTuner(
+                    params, dataset, valid_sets=dataset, study=study, model_dir="./booster"
+                )
```
toshihikoyanase, line 661: I think we can remove `study` since it will be created inside `LightGBMTuner`.
```suggestion

        with mock.patch("optuna.integration._lightgbm_tuner.optimize.os.mkdir") as m:
            with mock.patch("os.path.exists", return_value=mock_exists):
                LightGBMTuner(
                    params, dataset, valid_sets=dataset, model_dir="./booster"
                )
```

tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
```diff
@@ -932,6 +946,19 @@ def test_optuna_callback(self) -> None:
...
+    def test_model_dir(self, dir_exists: bool, expected: bool) -> None:
+        unexpected_value = 20  # out of scope.
+
+        params: Dict = {"verbose": -1, "lambda_l1": unexpected_value}
+        dataset = lgb.Dataset(np.zeros((10, 10)))
+        study = optuna.create_study()
+
+        with mock.patch("os.mkdir") as m:
+            with mock.patch("os.path.exists", return_value=dir_exists):
+                LightGBMTunerCV(params, dataset, study=study, model_dir="./booster")
```
toshihikoyanase, line 959: ditto.
```suggestion

        with mock.patch("os.mkdir") as m:
            with mock.patch("os.path.exists", return_value=dir_exists):
                LightGBMTunerCV(params, dataset, model_dir="./booster")
```

tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
```diff
@@ -647,6 +647,20 @@ def test_get_best_booster(self) -> None:
         with pytest.raises(ValueError):
             tuner2.get_best_booster()
 
+    @pytest.mark.parametrize("mock_exists, expected", [(False, True), (True, False)])
+    def test_model_dir(self, mock_exists: bool, expected: bool) -> None:
```
toshihikoyanase, line 651: A nit, but how about using the same variable name in both `test_model_dir`?
```suggestion
    @pytest.mark.parametrize("dir_exists, expected", [(False, True), (True, False)])
    def test_model_dir(self, dir_exists: bool, expected: bool) -> None:
```

toshihikoyanase (changes requested):
Thank you for your PR. I confirmed that it resolves the issue #2324. I added some small comments on tests.

tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
nyanhi: Thank you for pointing out! I have removed `study` as you mentioned.

toshihikoyanase (approved):

PR #2373, opened 2021-02-22
Add PyTorch checkpoint example using `failed_trial_callback`

not522:
## Motivation
This PR adds a PyTorch checkpoint example. It describes how to restart optimization from previous saved checkpoint, even if the process where the trial is running is killed for some reason.
## Description of the changes
- Add `pytorch/pytorch_checkpoint.py` example.
- Modify example test to check the example can restart when the previous trial is killed.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -51,13 +51,13 @@ jobs:
     - name: Run examples
       run: |
         if [ ${{ matrix.python-version }} = 3.6 ]; then
-          IGNORES='chainermn_.*|fastai*|pytorch_distributed_.*|rapids_.*|botorch_.*'
+          IGNORES='chainermn_.*|fastai*|pytorch_distributed_.*|rapids_.*|botorch_.*|pytorch/pytorch_checkpoint*'
         elif [ ${{ matrix.python-version }} = 3.8 ]; then
-          IGNORES='chainermn_.*|dask_ml_.*|keras_.*|pytorch_distributed_.*|tensorboard_.*|tensorflow_.*|tfkeras_.*|fastai*|rapids_.*'
+          IGNORES='chainermn_.*|dask_ml_.*|keras_.*|pytorch_distributed_.*|tensorboard_.*|tensorflow_.*|tfkeras_.*|fastai*|rapids_.*|pytorch/pytorch_checkpoint*'
         elif [ ${{ matrix.python-version }} = 3.9 ]; then
-          IGNORES='allennlp_*|botorch_*|catalyst_*|catboost_*|chainer_*|chainermn_.*|dask_ml_.*|fastai*|gluon_*|haiku_*|hydra_*|keras_*|lightgbm_*|mxnet_*|optuna_search_cv_*|pytorch_distributed_.*|pytorch_ignite_*|pytorch_lightning_*|rapids_*|sb3_*|skimage_lbp_*|sklearn_*|skorch_*|tensorboard_*|tensorflow_*|tfkeras_*|xgboost_*|pruning/simple*|enqueue_trial*|visualization/plot_pareto_front*|visualization/plot_study*|multi_objective/pytorch_simple*'
+          IGNORES='allennlp_*|botorch_*|catalyst_*|catboost_*|chainer_*|chainermn_.*|dask_ml_.*|fastai*|gluon_*|haiku_*|hydra_*|keras_*|lightgbm_*|mxnet_*|optuna_search_cv_*|pytorch_distributed_.*|pytorch_ignite_*|pytorch_lightning_*|rapids_*|sb3_*|skimage_lbp_*|sklearn_*|skorch_*|tensorboard_*|tensorflow_*|tfkeras_*|xgboost_*|pruning/simple*|enqueue_trial*|visualization/plot_pareto_front*|visualization/plot_study*|multi_objective/pytorch_simple*|pytorch/pytorch_checkpoint*'
         else
-          IGNORES='chainermn_.*|fastai*|pytorch_distributed_.*|rapids_.*'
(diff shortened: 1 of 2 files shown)


examples/pytorch/pytorch_checkpoint.py
```diff
@@ -0,0 +1,194 @@
...
+    storage = optuna.storages.RDBStorage(
+        "sqlite:///example.db", heartbeat_interval=1, failed_trial_callback=restart_from_checkpoint
+    )
+    study = optuna.create_study(
+        storage=storage, study_name="pytorch_checkpoint", direction="maximize", load_if_exists=True
+    )
+    study.optimize(objective, n_trials=100, timeout=600)
+
+    pruned_trials = [t for t in study.trials if t.state == optuna.trial.TrialState.PRUNED]
+    complete_trials = [t for t in study.trials if t.state == optuna.trial.TrialState.COMPLETE]
```
crcrpar, line 180: Could you use [`Study.get_trials` with `states` argument](https://optuna.readthedocs.io/en/latest/reference/generated/optuna.study.Study.html?highlight=get_trials#optuna.study.Study.get_trials) instead?

examples/pytorch/pytorch_checkpoint.py
```diff
@@ -0,0 +1,194 @@
...
+    optimizer_name = trial.suggest_categorical("optimizer", ["Adam", "RMSprop", "SGD"])
+    lr = trial.suggest_float("lr", 1e-5, 1e-1, log=True)
+    optimizer = getattr(optim, optimizer_name)(model.parameters(), lr=lr)
+
+    if "checkpoint_path" in trial.user_attrs:
+        checkpoint = torch.load(trial.user_attrs["checkpoint_path"])
+        epoch_begin = checkpoint["epoch"] + 1
+        model.load_state_dict(checkpoint["model_state_dict"])
+        optimizer.load_state_dict(checkpoint["optimizer_state_dict"])
+        accuracy = checkpoint["accuracy"]
```
crcrpar, line 88: Where is this `accuracy` variable used?
If it's not used, I'd like you to remove this.

examples/pytorch/pytorch_checkpoint.py
```diff
@@ -0,0 +1,194 @@
...
+
+In this example, we optimize the validation accuracy of hand-written digit recognition using
+PyTorch and MNIST. We optimize the neural network architecture as well as the optimizer
+configuration. As it is too time consuming to use the whole MNIST dataset, we here use a small
+subset of it.
+
+Even if the process where the trial is running is killed for some reason, you can restart from
+previous saved checkpoint using heartbeat.
+
+"""
```
crcrpar, line 12: It'd be better to give example commands to run&kill this example first and restart hyperparameter optimization because this example is different from others in that this example will be once killed and resumed.
```suggestion
"""
Optuna example that optimizes multi-layer perceptrons using PyTorch with checkpoint.

In this example, we optimize the validation accuracy of hand-written digit recognition using
PyTorch and MNIST. We optimize the neural network architecture as well as the optimizer
configuration. As it is too time consuming to use the whole MNIST dataset, we here use a small
subset of it.

Even if the process where the trial is running is killed for some reason, you can restart from
previous saved checkpoint using heartbeat.

    $ timeout 20 python examples/pytorch/pytorch_checkpoint.py
    $ python examples/pytorch/pytorch_checkpoint.py
"""
```

toshihikoyanase:
@crcrpar @HideakiImamura Could you review this PR?
Let me set the milestone to `v2.6.0` since this is a good example of RDBStorage's `failed_trial_callback`.

examples/pytorch/pytorch_checkpoint.py
not522: If the process is killed between saving the last model and recording the objective value to the storage, the next trial skips the training loop and immediately returns `accuracy`. Saving `accuracy` is needed for this corner case.
I added a comment for this (tricky) implementation.

not522:
Thank you for the review. I updated the PR.

examples/pytorch/pytorch_checkpoint.py
crcrpar: Ah, that makes perfect sense to me. Thank you.

.github/workflows/examples.yml
```diff
@@ -101,3 +101,9 @@ jobs:
         done
       env:
         OMP_NUM_THREADS: 1
+    - name: Run PyTorch snapshot example
```
crcrpar, line 104: [nit-picking] how about checkpoint instead of snapshot for the consistency?
```suggestion
    - name: Run PyTorch checkpoint example
```

not522: Thanks! I fixed.

HideakiImamura (commented):
Thanks for the PR! The example looks good to me.

Why don't we output `plot_intermediate_values` at the end to make it easier for users who see this example to understand that resuming from a checkpoint is working? The following is an example of the output of the plot. The red line at the bottom starts in the middle (step=4), indicating that resuming from the checkpoint is working properly. (I set `n_trials` to 10 to make the plot easier to read.)

Unknown-4

crcrpar:
I'm not sure how to show users that figure because I don't know the standard-ish way of showing figures from python scripts.

crcrpar (approved):

not522:
I updated the PR to show intermediate values. The coding style follows `examples/visualization`.

HideakiImamura (approved):
Thanks for the update! LGTM!

HideakiImamura:
@not522 The CI is stacked and cannot be re-run... Could you merge the master branch to kick the CI?

not522:
OK, the CI is passed.

PR #2379, opened 2021-02-24
Pin botorch

crcrpar:
Related to #2381.

The arguments of [`NondominatedPartitioning`](https://botorch.org/api/utils.html?highlight=nondominatedpartitioning#botorch.utils.multi_objective.box_decompositions.non_dominated.NondominatedPartitioning) seems to have been changed during the transition to v0.4.0.


hvy:
Thanks for the quick fix. Could you just add a TODO comment briefly explaining how we can remove the pin? I think it should be enough to change the number of outcomes to the reference point (constructed below and passed to `qExpectedHypervolumeImprovement`) at https://github.com/optuna/optuna/blob/master/optuna/integration/botorch.py#L203-L205.

hvy (approved):
Thanks a bunch, LGTM!

toshihikoyanase (approved):
LGTM. Thank you!

PR #2398, opened 2021-02-27
Move study's tests for `optimize` to `test_optimize.py`

nzw0301:
## Motivation

Test functions related to `optimize` functions are in `test_study.py`, but they should be in `test_optimize.py`.

## Description of the changes

- Move the functions in `tests/multi_objective_tests/test_study.py` and `tests/study_tests/test_optimize.py` to `tests/study_tests/test_optimize.py`
- Rename function names due to the redundantancy.

### DIFF
--- a/tests/multi_objective_tests/test_study.py
+++ b/tests/multi_objective_tests/test_study.py
@@ -45,22 +45,6 @@ def test_load_study() -> None:
         assert created_study._study._study_id == loaded_study._study._study_id
 
 
-@pytest.mark.parametrize("n_objectives", [1, 2, 3])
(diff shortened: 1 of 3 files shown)


nzw0301:
We discuss this PR with @g-votte and @HideakiImamura. 

I moved inappropriate functions to `_optimize.py`. Ideally,  only `test_optimize_parallel` and `test_optimize_parallel_timeout` should be moved into `_optimize.py`. However, they depend on the `Func` class, and we didn't draw a good conclusion to move them, so let us keep them in `test_study.py` once.

Issue #1936, opened 2020-10-15
Upgrade TFKerasPruningCallback to work with TF 2.3


brethvoice:
## Description

According to the following link, "This callback is intend to be compatible for TensorFlow v1 and v2, but only tested with TensorFlow v1."

https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.TFKerasPruningCallback.html#optuna.integration.TFKerasPruningCallback

## Motivation

TF 2.3 is working well and worthy of recognition.  Currently, the hook is not working to prune trials.

## Description

Make the hook work with TF 2.3+.

## Alternatives (optional)

Fork the code and make it work myself.

## Additional context (optional)

For an example of how it does not work, run 'study_mnist_classifier_settings.py` at the following repo:
[https://github.com/brethvoice/optuna_demo_MNIST](https://github.com/brethvoice/optuna_demo_MNIST)

It will take about 30 minutes to run but you can observed how no trials are *ever* pruned.  Max trials = 10, max timeout is 30 minutes.

brethvoice:
Possibly need to approve/reject this pull request first?
https://github.com/optuna/optuna/pull/1184

toshihikoyanase:
@brethvoice Thank you for your report. I don't think this issue is directly related to #1184, and we can work on each issue independently.

@sfujiwara I thought `TFKerasPruningCallback` worked with the latest `tensorflow`. Do you have any comments on this issue?

sfujiwara:
> According to the following link, "This callback is intend to be compatible for TensorFlow v1 and v2, but only tested with TensorFlow v1."
> https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.TFKerasPruningCallback.html#optuna.integration.TFKerasPruningCallback

I guess the above statement is out dated and we have to fix it. When I implemented `TFKerasPruningCallback`, We tested the callback on v1 because TensorFlow v2 was experimental. But currently we use TensorFlow v2 on testing as below (currently `tensorflow` without specifying version would install v2):
https://github.com/optuna/optuna/blob/1375de019a824a705fa22682491385350ff283b1/setup.py#L113-L138

sfujiwara:
@brethvoice I tried your code. It's a little bit complicated to understand perfectly, but I found the metric name is wrong:

https://github.com/brethvoice/optuna_demo_MNIST/blob/db5544d919840a12647107700254bc60f2353257/study_mnist_classifier_settings.py#L114

[...]

PR #2399, opened 2021-02-27
fix outdated docstring of TFKerasPruningCallback

sfujiwara:
## Description of the changes

Currently, Optuna is tested with TensorFlow v2 and the statement `This callback is intend to be compatible for TensorFlow v1 and v2, but only tested with TensorFlow v1.` is outdated as mentioned here: https://github.com/optuna/optuna/issues/1936#issuecomment-710783063.

The packages used for test are below:
https://github.com/optuna/optuna/blob/ddc4af84e6eabce18f9dabdaee49fd4f1db28c3e/setup.py#L126-L146

When the version of TensorFlow is not specified, TensorFlow v2 is used.

### DIFF
--- a/optuna/integration/tfkeras.py
+++ b/optuna/integration/tfkeras.py
@@ -16,7 +16,7 @@ if not _imports.is_successful():
 class TFKerasPruningCallback(Callback):
     """tf.keras callback to prune unpromising trials.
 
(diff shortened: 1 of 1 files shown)


crcrpar (approved):
Tip hat for the good catch.

HideakiImamura (approved):
LGTM!

Issue #2331, opened 2021-02-09
MOTPESampler StorageInternalError after ~300 iterations


sehoff:
Running a multiobjective optimization using the MOTPESampler, an exception is raised after ~300 iterations when writing to a postgres DB. Apparently, the length of the insert into trial_system_attributes is too long.

- The optuna exceptions is: "optuna.exceptions.StorageInternalError: An exception is raised during the commit. This typically happens due to invalid data in the commit, e.g. exceeding max length."
- It occurs when inserting into trial_system_attributes and the key: 'motpe:splitcache'.

## Expected behavior

(i) Is there a chance to increase the limit (maybe no limit) of the column, such that this exception is not raised?
(ii) If not, can this be handled in another way ? I actually do not need this information myself, so maybe there is an option to not write it to the RDB

## Environment

- Optuna version: 2.5.0
- Python version: 3.8.5
- OS: Windows10

## Error messages, stack traces, or logs
- This is the error from the example provided below. The one I get is basically the same. Is use three objective functions but this should not matter here!

[I 2021-02-09 10:51:19,607] Trial 298 finished with values: [4.007762489974416, 38.003975116711004] and parameters: {'x': 0.3695368246169375, 'y': 0.9302597259613226}.
[I 2021-02-09 10:51:20,211] Trial 299 finished with values: [49.20740296652893, 12.833723674946304] and parameters: {'x': 2.291235054547539, 'y': 2.655577652121054}.
[I 2021-02-09 10:51:21,086] Trial 300 finished with values: [90.85183690981745, 12.874199880388066] and parameters: {'x': 4.5426754735860415, 'y': 1.4412004611205884}.
[I 2021-02-09 10:51:21,778] Trial 301 finished with values: [24.243756546890843, 23.964240426067477] and parameters: {'x': 2.279352836003099, 'y': 0.9303170350624244}.
[I 2021-02-09 10:51:22,231] Trial 303 finished with values: [35.489315997021954, 24.6078917994922] and parameters: {'x': 0.48805437834067955, 'y': 2.938389341635649}.
[I 2021-02-09 10:51:22,303] Trial 302 finished with values: [29.709922859084305, 19.10484199146676] and parameters: {'x': 2.121506619593528, 'y': 1.7107572527369035}.
[I 2021-02-09 10:51:23,111] Trial 304 finished with values: [37.020124677459904, 24.090399224150516] and parameters: {'x': 0.5188215621728609, 'y': 2.997641632348585}.
[...]

PR #2395, opened 2021-02-26
Mitigate string-length limitation of `RDBStorage`

toshihikoyanase:
## Motivation
This PR relates to #1860.

SQLAlchemy has multiple types for string data. Currently, [`sqlalchemy.types.String`](https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.String) is used for variable-length strings such as `Trial.system_attr` and `Trial.user_attr`. It has strict length limitation, i.e., 2048 characters.
On the other hand, [`sqlalchemy.types.Text`](https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.Text) may be suitable for such variable-length strings. It is corresponding to `TEXT` type of each database system, which is designed for large text objects.

Although `TEXT` is not a standard type of SQL, many database systems implement it.

| Database | Corresponding type | Length limitation|
| --- | --- | --- |
|SQLite| [TEXT](https://github.com/sqlalchemy/sqlalchemy/blob/2ce1f1d9a8011a4673428850f01b04196e070957/lib/sqlalchemy/dialects/sqlite/base.py#L2166-L2167) | [SQLITE_MAX_LENGTH](https://www.sqlite.org/limits.html#max_length) |
|PostgreSQL| [text](https://github.com/sqlalchemy/sqlalchemy/blob/2ce1f1d9a8011a4673428850f01b04196e070957/lib/sqlalchemy/dialects/postgresql/base.py#L2042) | [unlimited](https://www.postgresqltutorial.com/postgresql-char-varchar-text/)|
|MySQL| [text](https://github.com/sqlalchemy/sqlalchemy/blob/2ce1f1d9a8011a4673428850f01b04196e070957/lib/sqlalchemy/dialects/mysql/base.py#L1375) | [2^16 + 2](https://dev.mysql.com/doc/refman/8.0/en/blob.html) bytes |
|SQL Server| [text](https://github.com/sqlalchemy/sqlalchemy/blob/2ce1f1d9a8011a4673428850f01b04196e070957/lib/sqlalchemy/dialects/mssql/base.py#L1344) | [2^31-1](https://docs.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-ver15) bytes |
[...]

PR #2406, opened 2021-02-27
Change caching implementation of MOTPE

y0z:
## Motivation
This PR fixes issue #2331.

## Description of the changes
This PR changes the caching implementation of MOTPE.

Previously, MOTPE uses `trial.system_attrs` to cache its internal parameters. This causes `optuna.exceptions.StorageInternalError` after hundreds of iterations because of exceeding max length when the cache becomes large.

Now, MOTPE uses its member variables to cache its internal parameters and this fixes the issue mentioned above.

### DIFF
--- a/optuna/samplers/_tpe/multi_objective_sampler.py
+++ b/optuna/samplers/_tpe/multi_objective_sampler.py
@@ -293,9 +293,9 @@ class MOTPESampler(TPESampler):
             self._split_cache[trial._trial_id] = attrs
 
         below = cvals[indices_below]
-        self._weights_below[trial._trial_id] = np.asarray([
-            w for w, v in zip(weights_below, below) if v is not None
-        ], dtype=float)
+        self._weights_below[trial._trial_id] = np.asarray(
+            [w for w, v in zip(weights_below, below) if v is not None], dtype=float
+        )
         below = np.asarray([v for v in below if v is not None], dtype=float)
         above = cvals[indices_above]
         above = np.asarray([v for v in above if v is not None], dtype=float)
@@ -397,9 +397,8 @@ class MOTPESampler(TPESampler):
 
         size = (self._n_ehvi_candidates,)
 
-        array_weights_below = np.asarray(self._weights_below[trial._trial_id], dtype=float)
         weights_below: Callable[[int], np.ndarray]
-        weights_below = lambda _: array_weights_below  # NOQA
+        weights_below = lambda _: self._weights_below[trial._trial_id]  # NOQA
 
         parzen_estimator_parameters_below = _ParzenEstimatorParameters(
             self._parzen_estimator_parameters.consider_prior,
--- a/optuna/samplers/_tpe/multi_objective_sampler.py
+++ b/optuna/samplers/_tpe/multi_objective_sampler.py
@@ -225,19 +225,6 @@ class MOTPESampler(TPESampler):
                 )
             )
 
-    def after_trial(
-        self,
-        study: optuna.study.Study,
-        trial: optuna.trial.FrozenTrial,
-        state: optuna.trial.TrialState,
-        values: Optional[Sequence[float]],
-    ) -> None:
-        if trial._trial_id in self._split_cache:
-            del self._split_cache[trial._trial_id]
-        if trial._trial_id in self._weights_below:
-            del self._weights_below[trial._trial_id]
-        self._mo_random_sampler.after_trial(study, trial, state, values)
-
     def _split_mo_observation_pairs(
         self,
         study: optuna.study.Study,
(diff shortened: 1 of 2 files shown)


y0z:
ci/circleci: document and ci/circleci: document artifact seem to be stuck.

keisuke-umezawa:
@HideakiImamura 
If you have time, could you review it?

HideakiImamura:
@y0z I think the CI will pass if you merge the master branch. Could you merge the master branch?

optuna/samplers/_tpe/multi_objective_sampler.py
```diff
@@ -140,6 +140,10 @@ def __init__(
         )
         self._n_ehvi_candidates = n_ehvi_candidates
         self._mo_random_sampler = RandomSampler(seed=seed)
+        self._split_cache: Dict[int, Any] = {}
+        self._split_cache_lock = threading.Lock()
+        self._weights_below: Dict[int, Any] = {}
+        self._weights_below_lock = threading.Lock()
```
HideakiImamura, line 146: Is it necessary to take a lock to achieve this feature? In Optuna's parallelization scheme, one running trial is tied to only one process/one thread. It would be possible to remove all the code associated with the lock.

optuna/samplers/_tpe/multi_objective_sampler.py
```diff
@@ -247,8 +266,11 @@ def _split_mo_observation_pairs(
 
         # Solving HSSP for variables number of times is a waste of time.
         # We cache the result of splitting.
-        if _SPLITCACHE_KEY in trial.system_attrs:
-            split_cache = trial.system_attrs[_SPLITCACHE_KEY]
+        with self._split_cache_lock:
+            is_cached = trial._trial_id in self._split_cache
+        if is_cached:
+            with self._split_cache_lock:
+                split_cache = self._split_cache[trial._trial_id]
```
HideakiImamura, line 273: How about adding test cases to validate the `split_cache` is appropriately set?

HideakiImamura (commented):
Thanks for the update. Left some comments. PTAL.

optuna/samplers/_tpe/multi_objective_sampler.py
y0z: I have removed locks.

optuna/samplers/_tpe/multi_objective_sampler.py
y0z: I have added a simple test case for caches.

optuna/samplers/_tpe/multi_objective_sampler.py
```diff
@@ -289,14 +290,12 @@ def _split_mo_observation_pairs(
...
 
         below = cvals[indices_below]
-        study._storage.set_trial_system_attr(
-            trial._trial_id,
-            _WEIGHTS_BELOW_KEY,
-            [w for w, v in zip(weights_below, below) if v is not None],
-        )
+        self._weights_below[trial._trial_id] = [
+            w for w, v in zip(weights_below, below) if v is not None
+        ]
```
HideakiImamura, line 298: It's simpler to cast this as `np.ndarray` here. 
```suggestion
        self._weights_below[trial._trial_id] = np.asarray([
            w for w, v in zip(weights_below, below) if v is not None
        ], dtype=float)
```

y0z:
In ci/circleci: document, `urllib.error.HTTPError: HTTP Error 503: Service Unavailable` has occured though after merging #2459.

HideakiImamura:
@y0z Thanks for your swift actions. The CI failure has been fixed now by #2468. Could you merge the master branch again?

HideakiImamura:
@c-bata: @Crissman would like to review the harder PRs, so let me re-assign this PR to him.

HideakiImamura:
@Crissman Could you review this PR if you have time? It would be great if you could double-check if the problem reported in #2331 has been resolved.

Crissman:
This issue appears fixed by #2395 . Please confirm and close this PR and original issue in #2331. 🙇‍♂️

y0z:
I think that the memory-based cache (this PR) is more preferable to the DB-based cache (current implementation) because users generally do not need the cache data for MOTPE's internal implementation in their DB. Therefore, this PR is still useful and valuable.

Crissman:
@y0z Thanks for the PR. Discussed with the team, and we agree that this PR is still useful and valuable.

I've read over the code and tested it locally and it runs fine.

I'm not sure the additional test is required. I feel that if the cache was not working, we would see that result in the other multi-objective tests, so I think we can remove this test. What do you think?

Crissman:
After discussion, test worth doing. Merging!

Crissman (approved):

PR #2416, opened 2021-02-27
Eliminate duplicate `ExperimentalWarning` filters

HirokiNishimoto:
## Motivation
Thanks to # 2405, which changed the setting of `pyproject.toml`,` ExperimentalWarning` is no longer displayed when running pytest.
However, elsewhere in the code, there was some duplication left to prevent the experimental warning from appearing when running pytest. We don't need this anymore and would like to remove it.

## Description of the changes
Deleted all duplicate lines filtering `ExperimentalWarning` (specifically, the following two lines) from all python scripts included in `optuna/tests`
- `warnings.simplefilter("ignore", optuna.exceptions.ExperimentalWarning)`
- `@pytest.mark.filterwarnings("ignore::optuna.exceptions.ExperimentalWarning")`

### DIFF
--- a/tests/integration_tests/test_cma.py
+++ b/tests/integration_tests/test_cma.py
@@ -146,7 +146,6 @@ class TestPyCmaSampler(object):
     def test_call_after_trial_of_independent_sampler() -> None:
         independent_sampler = optuna.samplers.RandomSampler()
         with warnings.catch_warnings():
-            warnings.simplefilter("ignore", optuna.exceptions.ExperimentalWarning)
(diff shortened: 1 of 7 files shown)


g-votte:
Thanks for your PR!

We have had an issue regarding to the warning filter and temporarily reverted the config change in https://github.com/optuna/optuna/pull/2413

Let's revisit this PR after the issue is resolved.

g-votte:
Sorry that the review is pending for a long time. The coverage problem is still unresolved, which is tracked in https://github.com/optuna/optuna/issues/2517

Issue #2396, opened 2021-02-26
Don't use `numpy.append`


HideakiImamura:
In some parts of Optuna, `numpy.append` is used. However, dynamically changing `numpy.ndarray` is slow, so for performance, it is better to append as a list first, and then cast with `numpy.asarray`. We are looking for someone to list and fix the parts where `numpy.append` is used.

hvy:
I suppose you're referring to code that repeatedly calls `numpy.append` inside a loop. Allocating new arrays in each iteration is inefficient and I agree that such code can be optimized. Just leaving a note that this can be done in different ways. For instance, if you know the loop size, you could perform a single allocation prior to the loop.

jeromepatel:
I would like to work on this issue if you can guide me on this, it would be helpful.

nzw0301:
@HideakiImamura I think that this issue can be closed.

@jeromepatel Please do not worry that your kindful proposal was not accepted 😢

jeromepatel:
@nzw0301 I am happy with the progress and contribution I am making to this organization. So, thank you for all your support and guidance.

nzw0301:
Let me close this issue since the related pull request #2419 has been merged. Please feel free to re-open this issue if this issue has not been resolved yet.

PR #2419, opened 2021-02-27
Fix - Replacing numpy.append

nyanhi:
## Motivation
This is to address the issue #2396 for below. Please kindly note this is still _draft_ and I haven't confirmed if it will improve the performance yet.
`optuna.multi_objective.samplers._motpe.py`
`optuna.samplers._tpe.multi_objective_sampler.py`

## Description of the changes
As suggested in the discussion of #2396, this PR tries to allocate empty array with the size `n_below` and assign array inside of the loop.

### DIFF
--- a/optuna/multi_objective/samplers/_motpe.py
+++ b/optuna/multi_objective/samplers/_motpe.py
@@ -272,7 +272,7 @@ class MOTPEMultiObjectiveSampler(TPESampler, BaseMultiObjectiveSampler):
             last_idx = 0
             while len(indices_below) + sum(nondomination_ranks == i) <= n_below:
                 length = indices[nondomination_ranks == i].shape
-                indices_below[last_idx: last_idx+length] = indices[nondomination_ranks == i]
+                indices_below[last_idx : last_idx + length] = indices[nondomination_ranks == i]
                 last_idx += length
                 i += 1
 
--- a/optuna/multi_objective/samplers/_motpe.py
+++ b/optuna/multi_objective/samplers/_motpe.py
@@ -265,16 +265,19 @@ class MOTPEMultiObjectiveSampler(TPESampler, BaseMultiObjectiveSampler):
             assert 0 <= n_below <= len(lvals)
 
             indices = np.array(range(len(lvals)))
-            indices_below = np.array([], dtype=int)
+            indices_below = np.empty(n_below, dtype=int)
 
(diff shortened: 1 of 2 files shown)


HideakiImamura:

HideakiImamura (commented):
Thanks for the PR! I'd like to know the performance improvement. If you have trouble how to measure the performance, please tell us. We will help you as much as we can.

By the way, I have one comment. The `optuna/multi_objective` module is deprecated in 2.4.0, so we don't need to fix the module anymore. You can ignore the module when fix.

HideakiImamura:
@nyanhi Could you apply the formatter and merge the master branch to pass the CI?

nyanhi:
@HideakiImamura Thank you! Sure, I will cancel the change for `optuna/multi-objective` and apply formatter etc. by tomorrow morning.

HideakiImamura:
I confirmed that this PR's change does not affect the performance of `MOTPESampler`. So let me approve this PR.

<details>
<summary> profiling code </summary>

```python
import optuna

seed = 128


def objective(trial):
    x = []
    x.append(trial.suggest_float("x1", 0.0, 2.0))
    x.append(trial.suggest_float("x2", 0.0, 4.0))
    return x


sampler = optuna.samplers.MOTPESampler(
    n_startup_trials=20, n_ehvi_candidates=24, seed=seed
)
study = optuna.create_study(directions=["minimize"] * 2, sampler=sampler)
study.optimize(objective, n_trials=400
```
</details>

- current master
スクリーンショット 2021-03-03 17 40 55

- This PR
スクリーンショット 2021-03-03 17 41 29

HideakiImamura (approved):

nyanhi:
@HideakiImamura Apologies for the delay in response, I will also take a look this tomorrow. 
I could be wrong, but the profiling code seems not going inside the loop or size of the loop is pretty small.

nyanhi:
@HideakiImamura I found a bug in my previous commit and have fixed it. 

I agree that this PR would not change the performance much, given `np.append` occupies minor part of execution time in the first place (seems `n_below` tends to be reasonably small by definition? Not too sure).

himkt (approved):
@nyanhi 

I think ndarray operation in the profiling has a small part of the entire execution, as you said. But this PR looks solid that we can avoid changing the shape of ndarray dynamically.

Thank you for the contribution!

PR #2435, opened 2021-03-03
[RFC] Run examples if changed in pull request CI

crcrpar:
## Motivation

Make new/changed examples to be checked autonomically in CI.

## Description of the changes

- Create "tools" directory
- Move "formats.sh" to "tools"
- Add `tools/run_examples.py` which runs examples. We can narrow the scope with `--selected-files` argument. Also, there are some options to skip specific type of examples like jupyter notebook & multinode.

## What this enables
- run added/changed examples in pull request CI


HideakiImamura:
@himkt Could you review this PR if you have time?

tools/run_examples.py
```diff
@@ -0,0 +1,190 @@
...
+import sys
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
+
+
+EXAMPLES_DIR = Path(__file__).parents[1] / Path("examples")
+ALL_EXAMPLES: List[str] = [str(p) for p in EXAMPLES_DIR.glob("**/*.py")]
+# TODO (crcrpar): Allow fastaiv2 example once https://forums.fast.ai/t/ganlearner-error-no-implementation-found-on-types-that-implement-invisibletensor/83451/7 gets resolved.
```
himkt, line 14:
```suggestion
# TODO (crcrpar): Allow fastaiv2 example once https://forums.fast.ai/t/ganlearner-error-no-implementation-found-on-types-that-implement-invisibletensor/83451/7 gets resolved.  # NOQA
```

tools/run_examples.py
```diff
@@ -0,0 +1,190 @@
...
+    (3, 8): _DEFAULT_DENY_LIST + ["keras_", "tensorboard_", "tensorflow_", "tfkeras_"],
+}
+
+
+def _get_distributed_examples(files: List[str]) -> List[str]:
+    return [f for f in files if "distributed" in f]
+
+
+def _get_ipynb_examples(files: Optional[List[str]] = None) -> List[str]:
+    if files is None:
```
himkt, line 34: This else block is really needed? I don't understand when the lines are executed.

himkt (commented):
Thank you for tackling developer experience improvement. I personally agree with your idea of separating logic for executing scripts from the GitHub Actions configuration.

I'd like other members' opinions on it. @hvy @toshihikoyanase @nzw0301 What do you think about the changes? (I mention as you guys have changed `examples.yml` multiple times.)

tools/run_examples.py
```diff
@@ -0,0 +1,190 @@
...
+from pathlib import Path
+import subprocess
+import sys
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
+
+
+EXAMPLES_DIR = Path(__file__).parents[1] / Path("examples")
```
nzw0301, line 12: EXAMPLES_DIR = Path(__file__).parents[1] / "examples"

nzw0301, line 12: Simpler way: 

```suggestion
EXAMPLES_DIR = Path(__file__).parents[1] / Path("examples")
```

nzw0301:
@himkt I'm afraid that I checked the `examples.yml` once to add hydra's example. But I try to follow this PR; hopefully, I'll give some feedback...

tools/run_examples.py
crcrpar: Not needed and `files` should be `List[str]`, thanks

.github/workflows/examples.yml
```diff
@@ -47,55 +48,19 @@ jobs:
...
-          echo $file
-          STUDY_NAME=`optuna create-study --storage $STORAGE_URL`
-          mpirun -n 2 -- python $file $STUDY_NAME $STORAGE_URL > /dev/null
-        done
-        mpirun -n 2 -- python examples/pytorch/pytorch_distributed_simple.py
-      env:
-        OMP_NUM_THREADS: 1
-    - name: Run Hydra examples
+
+    - name: Run examples (Pull Request)
```
HideakiImamura, line 61: Typo?
```suggestion
    - name: Run examples (Schedule)
```

tools/run_examples.py
```diff
@@ -0,0 +1,190 @@
...
+        self.hydra_examples = _get_hydra_examples(all_examples)
+        self.vanilla_examples = list(
+            set(all_examples)
+            - set(self.distributed_examples)
+            - set(self.multinode_examples)
+            - set(self.hydra_examples)
+        )
+
+
+def parse_args():
```
HideakiImamura, line 72: Not a strong opinion, but the type hinting is missing, so how about adding for consistency? 
```suggestion
def parse_args() -> argparse.NameSpace:
```

The same is true for the other functions.

tools/run_examples.py
```diff
@@ -0,0 +1,190 @@
+import argparse
```
HideakiImamura, line 1: This may be beyond the scope of this PR, but I am curious how you plan to maintain this file. For example, there are several logics in this file, are there any plans to document them or provide tests for them?

HideakiImamura (commented):
Thanks for the PR. I personally agree with the idea to run examples if changed in PRs. In addition, I think the `tools/run_example.py` is easier to maintain than the current `.github/workflows/examples.yml`, so looks reasonable.

I have several early comments. Please take a look.

tools/run_examples.py
crcrpar: I think if we operate examples CI with this script, we ought to update this regularly, which I think enough.
Also, I basically think this script should be maintained in a more casual way compared to optuna source code because
the number of users & readers is limited, I expect them to be familiar with Optuna's CI, this script is ostensibly Python: shell script with convenient Python's argparse and string functions.

> For example, there are several logics in this file, are there any plans to document them or provide tests for them?

By document, do you mean we mention this script in Optuna's API documentation or somewhere else? Or do you mean I should write comments more in this script?
As to tests, do you have any concrete and reasonable test? If you want to check the selected tests are correct or not, we're going to maintain the lists of examples compatible with specific Python versions. If we do this, it's better to maintain that those lists in this script.

crcrpar:
Now that pytorch checkpoint, which requires the custom command to check, resides in examples and I'm not sure the current design is reasonable or not.

crcrpar:
thank you for reviews. I'm closing this since I'm going to leave this org and I don't have enough time to handle the growing number of exceptions in our example CI and resolve conflicts.

PR #2438, opened 2021-03-04
Hotfix for MNIST download in tutorial

hvy:
## Motivation

Hotfix for broken master since the v2.6 code freeze is near. This fix might not be needed when torchvision 0.9 is released. However, I'm not sure if we can immediately bump the dependency requirement without other incompatibilities.

Please note that examples are excluded since they're not in practice blocking the release.

## Description of the changes

Fixes a `HTTP Error 403: Forbidden` when trying to download MNIST from `torchvision` using the default `urllib` header.


toshihikoyanase (approved):
LGTM!

HideakiImamura (approved):
Thanks for your quick fix! LGTM!

PR #2375, opened 2021-02-23
Updated `Dockerfile` of Kubernetes simple example

0x41head:
## Motivation
#2050 

## Description of the changes
Changed the base image of the dockerfile to `optuna/optuna:py3.7-dev` and removed the `RUN` statement

PR #2440, opened 2021-03-04
Replace `sh` with `bash` in README of kubernetes examples

toshihikoyanase:
## Motivation

During the review of #2375, I found that the current instruction causes errors because of the inconsistency of shells. The instruction uses `sh`, but the shell script uses `bash` specific syntax (i.e., `function`).

## Description of the changes

This PR simply replaces `sh` in the README.md with `bash`.

Alternatively, we can use `sh` syntax in `run.sh`.

```diff
- function run {
+ run () {
```


HideakiImamura:
@crcrpar Could you review this PR if you have time?

HideakiImamura (approved):
Looks reasonable.

crcrpar (approved):

PR #2453, opened 2021-03-08
Apply #2438 to pytorch examples

crcrpar:
## Motivation

What was resolved in https://github.com/optuna/optuna/pull/2438 can happen in other PyTorch scripts if they use torchvision's MNIST downloader as follows:

image

Capture from https://github.com/optuna/optuna/runs/2050946963?check_suite_focus=true#step:7:6012

## Description of the changes

Copy & paste the fix of #2048 to PyTorch examples.

PR #2459, opened 2021-03-10
Apply urllib patch for MNIST download

crcrpar:
403 could occur anywhere we try to download MNIST from yann.lecun.com like https://github.com/optuna/optuna/runs/2067792885?check_suite_focus=true#step:7:8298.

Related: #2438 #2453

## Description of the changes

Apply the same patch with some comments changed to examples downloading MNIST however I'm not completely sure every single file downloads the dataset from yann.lecun.com.

### DIFF
--- a/examples/catalyst_simple.py
+++ b/examples/catalyst_simple.py
@@ -11,6 +11,7 @@ argument.
 
 import argparse
 import os
+import urllib
 
 from catalyst.dl import AccuracyCallback
 from catalyst.dl import SupervisedRunner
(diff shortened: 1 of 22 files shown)


crcrpar:
I frequently have seen the same error in chainermn_integration example: https://github.com/optuna/optuna/pull/2368/checks?check_run_id=2072768467#step:7:8848

hvy (approved):
LGTM!

I noticed the change is not necessary for some that use Fashion MNIST but it's a minor and I think this is a temporary fix.

toshihikoyanase (approved):
LGTM. Thank you!

PR #2466, opened 2021-03-11
Remove Cython from `install_requires`.

c-bata:
## Motivation

Cython is not a dependency of Optuna so I think it should not be listed here even though some optional dependencies don't provide Python 3.9 wheel. 
Refs https://github.com/conda-forge/optuna-feedstock/pull/24

## Description of the changes

Move Cython from `install_requires` to `extra_requires` for tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,4 @@
 import os
-import sys
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -121,6 +120,8 @@ def get_extras_require() -> Dict[str, List[str]]:
             "bokeh<2.0.0",
             "chainer>=5.0.0",
             "cma",
+            # TODO(c-bata): Remove Cython after removing version constraints of sklearn.
+            "Cython; python_version > '3.8'",
             "fakeredis",
             "lightgbm",
             "matplotlib>=3.0.0",
@@ -163,6 +164,8 @@ def get_extras_require() -> Dict[str, List[str]]:
             # https://github.com/optuna/optuna/issues/1000.
             "chainer>=5.0.0",
             "cma",
+            # TODO(c-bata): Remove Cython after removing version constraints of sklearn.
+            "Cython; python_version > '3.8'",
             "lightgbm",
             "mlflow",
             "mpi4py",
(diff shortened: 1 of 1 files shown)


toshihikoyanase:
@HideakiImamura Could you review this PR? The document jobs of CI fail time after time, so maybe we should resolve the CI issue, first.

```console
Unexpected failing examples:
/home/docs/project/tutorial/20_recipes/002_multi_objective.py failed leaving traceback:
  File "/home/docs/project/tutorial/20_recipes/002_multi_objective.py", line 123, in <module>
    study.optimize(objective, n_trials=30, timeout=300)
...
urllib.error.HTTPError: HTTP Error 503: Service Unavailable
```

setup.py
```diff
@@ -153,7 +150,13 @@ def get_extras_require() -> Dict[str, List[str]]:
             "botorch>=0.4.0 ; python_version>'3.6'",
             "fastai",
         ],
-        "tests": ["fakeredis", "pytest"],
+        "tests": [
+            # TODO(c-bata): Remove Cython after removing version constraints of sklearn
+            # from 'optional' in extra_requires.
+            "Cython; python_version > '3.8'",
```
toshihikoyanase, line 156: According to the comment, this line should be moved to `optional`. The installation of `optional` actually failed due to the following error:

```
ModuleNotFoundError: Please install Cython with a version >= 0.28.5 in order to build a scikit-learn from source.
```

toshihikoyanase (commented):
Thank you for your PR.

In my understanding, `Cython` is required by `scikit-learn` if Python version is 3.9. We can see `doctest`, `document`, `example` and `integration` but this PR doesn't add `Cython` to them. Could you update them, please?

setup.py
c-bata: Okay. I'm actually not sure we should force all Python 3.9 users to install Cython, but it's acceptable to add it in `optional` dependencies.

c-bata:
Regarding to the `document`, there is no constraints in scikit-learn so that Python 3.9 wheel is available. Thus, I omitted to add Cython in `document`.

toshihikoyanase (approved):

HideakiImamura (approved):
Thanks, LGTM!

PR #2477, opened 2021-03-13
Add example to `optuna.logging.disable_propagation`

jeromepatel:
## Motivation
With reference to issue #1334

## Description of the changes

Update docstring of `optuna.logging.disable_propagation` in `optuna/optuna/logging.py`. I have reference `optuna.logging.enable_propagation` example for simplicity and uniformity.
Let me know if it needs any modifications.


optuna/logging.py
```diff
@@ -175,6 +175,33 @@ def disable_propagation() -> None:
     """Disable propagation of the library log outputs.
 
     Note that log propagation is disabled by default.
```
keisuke-umezawa, line 177: How about modifying it to `Note that log propagation is disabled by default. You only need to use `disable_propagation` to stop log propagation when you use `enable_propagation`.

jeromepatel: Thank you for reviewing, seems more user friendly and readable. I'll change that.

keisuke-umezawa (approved):

optuna/logging.py
```diff
@@ -174,7 +174,35 @@ def enable_default_handler() -> None:
...
+                x = trial.suggest_float("x", -100, 100)
+                y = trial.suggest_categorical("y", [-1, 0, 1])
+                return x ** 2 + y
+
+        .. testcode::
+
+            import optuna
+
+            optuna.logging.disable_propagation()  # Stop propogating logs to the root logger.
+            optuna.logging.enable_default_handler()  # Start showing logs in sys.stderr.
```
toshihikoyanase, line 196: By default, the propagation is disabled and the default handler is enabled. So, I think these lines do not affect the behavior of logging. I guess we need to change the scenario of the code example.

For instance, how about the following?
1. disable the default hander
2. create a logger by `logging.getLogger` and set a hander.
3. enable the log propagation
4. execute `study.optimize` and confirm that log messages are shown
5. disable the log propagation
6. execute `study.optimize` and confirm that no log message are shown

toshihikoyanase (changes requested):
Thank you for your PR. I added a comment. Please let me know if I misunderstand your example.

optuna/logging.py
```diff
@@ -174,7 +174,35 @@ def enable_default_handler() -> None:
 def disable_propagation() -> None:
     """Disable propagation of the library log outputs.
 
-    Note that log propagation is disabled by default.
+    Note that log propagation is disabled by default. You only need to use `disable_propagation`
```
toshihikoyanase, line 177: `disable_propagation` is rendered in italic and it is not aligned with the other docstrings. How about replace it with 'this method'?
```suggestion
    Note that log propagation is disabled by default. You only need to use this function
```

image

jeromepatel:
@toshihikoyanase Thank you for your feedback and review. I agree with your proposed scenario and I have implemented it in the recent commit. Please correct me if I am wrong, as per what I understood here's the summary of the scenario:

1.  Disable the default handler (to stop propagating to `sys.stderr`)
2.  Create a new logger using `getLogger` and set a handler using `addHandler`
3.  Enable propagation to root logger and execute `study.optimize`
4.  Disable propagation to root logger and execute `study.optimize`
5. Assert the first logs are from the first optimize function call and there are no logs for the second optimize function call.

Thank you for your thoughtful comment.

optuna/logging.py
```diff
@@ -174,7 +174,50 @@ def enable_default_handler() -> None:
 def disable_propagation() -> None:
     """Disable propagation of the library log outputs.
 
-    Note that log propagation is disabled by default.
+    Note that log propagation is disabled by default. You only need to use this function
+    to stop log propagation when you use :func:`~optuna.logging.enable_propogation()`.
+
+    Example:
+
+        Stop propogating logs to the root logger and start logging to :obj:`sys.stderr`.
```
toshihikoyanase, line 182: A nit, but could you update the description of the example as well?
```suggestion
        Stop propagating logs to the root logger on the second optimize call.
```

toshihikoyanase (changes requested):
Thank you for your update. The new scenario sounds reasonable to me.
I added a small comment on the description.

toshihikoyanase (approved):

PR #2490, opened 2021-03-16
Add interval for lightgbm callback.

keisuke-umezawa:
## Motivation

Address #1136 . And to solve the problem instead of #1234.

### Assumption
- Users want to skip reporting and pruning with different intervals.
    - A user wants to report the value every 2 steps, and prune trials every 4 steps.

## Description of the changes

- Changed the lightgbm callback accordingly.


HideakiImamura:
@Crissman @toshihikoyanase Since the author has already requested reviews, let me assign this PR to you.

hvy:
Memo: Tests are missing.

optuna/integration/lightgbm.py
```diff
@@ -62,60 +62,72 @@ class LightGBMPruningCallback(object):
             Note that this argument will be ignored if you are calling
             `cv method <https://lightgbm.readthedocs.io/en/latest/Python-API.html#lightgbm.cv>`_
             instead of train method.
+        interval:
+            Check if trial should be reported for pruning every n-th epoch.
+            By default ``interval=1`` and reporting is performed after every epoch.
+            Notes that pruning itself is performed according to the interval definition
```
Crissman, line 68:
```suggestion
            Note that the pruning itself is performed according to the interval definition
```

optuna/integration/lightgbm.py
```diff
@@ -62,60 +62,72 @@ class LightGBMPruningCallback(object):
...
+            for evaluation_result in env.evaluation_result_list:
+                valid_name, metric, current_score, is_higher_better = evaluation_result[:4]
+                if valid_name != target_valid_name or metric != self._metric:
+                    continue
+
+                if is_higher_better:
+                    if self._trial.study.direction != optuna.study.StudyDirection.MAXIMIZE:
+                        raise ValueError(
+                            "The intermediate values are inconsistent with the objective values"
+                            "in terms of study directions. Please specify a metric to be minimized"
```
Crissman, line 109:
```suggestion
                            "in terms of the study directions. Please specify a metric to be minimized"
```

optuna/integration/lightgbm.py
```diff
@@ -62,60 +62,72 @@ class LightGBMPruningCallback(object):
...
+                            "for LightGBMPruningCallback."
+                        )
+                else:
+                    if self._trial.study.direction != optuna.study.StudyDirection.MINIMIZE:
+                        raise ValueError(
+                            "The intermediate values are inconsistent with the objective values"
+                            "in terms of study directions. Please specify a metric to be"
```
Crissman, line 116:
```suggestion
                            "in terms of the study directions. Please specify a metric to be"
```

Crissman (changes requested):
English review.

Note to self: review interactions between this interval and pruner intervals.

optuna/integration/lightgbm.py
```diff
@@ -62,60 +62,72 @@ class LightGBMPruningCallback(object):
             Note that this argument will be ignored if you are calling
             `cv method <https://lightgbm.readthedocs.io/en/latest/Python-API.html#lightgbm.cv>`_
             instead of train method.
+        interval:
+            Check if trial should be reported for pruning every n-th epoch.
+            By default ``interval=1`` and reporting is performed after every epoch.
```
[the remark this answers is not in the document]
keisuke-umezawa: In the new implementation, reporting not pruning happens every epoch.

keisuke-umezawa:
@toshihikoyanase I added a unittest. Could you review it?

Crissman:
Not clear how this implementation is different from #1183.

Is the intent to add a reporting interval that is distinct from the pruning interval? I don't understand how that would be helpful, since reporting should be a fairly quick action. Feels more important to control the pruning interval, which should be done at the pruner, not the integration.

optuna/integration/lightgbm.py
```diff
@@ -62,60 +63,85 @@ class LightGBMPruningCallback(object):
...
-        raise ValueError(
-            'The entry associated with the validation name "{}" and the metric name "{}" '
-            "is not found in the evaluation result list {}.".format(
-                target_valid_name, self._metric, str(env.evaluation_result_list)
-            )
-        )
+
+        if self._trial.should_prune():
+            message = "Trial was pruned at iteration {}.".format(env.iteration)
+            raise optuna.TrialPruned(message)
```
Crissman, line 145: Calling `should_prune` seems like it might also slow down performance, and without a `report`, might not have enough data to prune anyway. Did you consider putting this in scope of the if statement above?

Crissman:
Discussed with team, and there are significant performance costs to lightgbm (along with xgboost), so it is worthwhile to add an interval for the reporting. 👍

optuna/integration/lightgbm.py
keisuke-umezawa: I considered it, but there are pros/cons.

pros
- we can easily reduce the cost of `should_prune`

cons
- If I configured as `pruning_interval=7` and `reporting_interval=5`, the first time that `should_prune` will be called at 7 x 5 = 91 loops.

Some practical solutions will be:

1. Ask users to use the reasonable intervals for pruning and reporting e.g. `pruning_interval=8` and `reporting_interval=4`.
2. Refactor pruners to minimize the cost of `should_prune`
    - Move [these lines](https://github.com/optuna/optuna/blob/master/optuna/pruners/_percentile.py#L183-L186) to the beginning of `prune`.

Crissman: That feels weird having different intervals for pruning and reporting. Isn't the sole purpose of reporting to prune? What is the use of reporting if not to prune?

keisuke-umezawa: I heard that some users use report functionality to check intermediate results later.

keisuke-umezawa:
> Discussed with team, and there are significant performance costs to lightgbm (along with xgboost), so it is worthwhile to add an interval for the reporting. 👍

I will check that the current implementation work well by performance check.

keisuke-umezawa:
Perf with [this script](https://gist.github.com/keisuke-umezawa/4571551203eb02c1653c6cb36c571a4f) on this PR
| #trials | #interval | time(sec) | time/trial(sec) |
| ------- | --------- | --------- | --------------- |
| 100 | 0 | 4.83 | 0.048 |
| 100 | 1 | 2.71 | 0.027 |
| 100 | 2 | 3.00 | 0.030 |
| 100 | 4 | 3.47 | 0.035 |
| 100 | 8 | 2.88 | 0.029 |
| 100 | 16 | 3.01 | 0.030 |
| 100 | 32 | 3.36 | 0.034 |
| 500 | 0 | 25.24 | 0.050 |
| 500 | 1 | 16.30 | 0.033 |
| 500 | 2 | 16.01 | 0.032 |
| 500 | 4 | 18.33 | 0.037 |
| 500 | 8 | 18.83 | 0.038 |
| 500 | 16 | 25.91 | 0.052 |
| 500 | 32 | 28.91 | 0.058 |
| 1000 | 0 | 71.00 | 0.071 |
| 1000 | 1 | 44.69 | 0.045 |
| 1000 | 2 | 39.71 | 0.040 |
| 1000 | 4 | 44.15 | 0.044 |
| 1000 | 8 | 44.00 | 0.044 |
| 1000 | 16 | 46.87 | 0.047 |
| 1000 | 32 | 51.02 | 0.051 |

Crissman:
Thanks, @keisuke-umezawa -san. What's the plan from here? Check DB storage?

optuna/integration/lightgbm.py
keisuke-umezawa: > cons
> If I configured as pruning_interval=7 and reporting_interval=5, the first time that should_prune will be called at 7 x 5 = 35 loops.
> Some practical solutions will be:

Sorry, my previous explanation was wrong. For example, if pruning_interval=10 and reporting_interval=3, pruners will potentially prune the trial at 0, 12, not at the 30 loops.

image

So, in that sense, your suggestion will have few cons. I will choose your suggestion.

keisuke-umezawa:
@Crissman Thank you for your suggestion! I re-evaluated the performances with the more actual environment, and updates the comment. https://github.com/optuna/optuna/pull/2490#issuecomment-821781728

From the performance and the pros/cons for each approach (https://github.com/optuna/optuna/pull/2490#discussion_r624650956), I would like to choose your approach. Thank you.

optuna/integration/lightgbm.py
```diff
@@ -62,60 +63,85 @@ class LightGBMPruningCallback(object):
...
             if valid_name != target_valid_name or metric != self._metric:
                 continue
 
+            return evaluation_result
+
+        return None
+
+    def __call__(self, env: "CallbackEnv") -> None:
+
+        if (env.iteration + 1) % self._report_interval == 0:
```
keisuke-umezawa, line 103: A potential discussion point: `(env.iteration + 1) % self._report_interval` or `env.iteration % self._report_interval`

Crissman: I'd got with `+1` since that conforms to the expectation of every 5 trials, given the trials start at base 0.

Crissman (approved):
LGTM

toshihikoyanase:
@keisuke-umezawa Thank you for sharing the benchmarking results in https://github.com/optuna/optuna/pull/2490#issuecomment-821781728.

I guess we can see the clear performance difference if we use RDB as storage backend. What do you think of taking the benchmark using, for example, SQLite?

keisuke-umezawa:
## summary
TODO

## configuration
* with sqlite
* [this script](https://gist.githubusercontent.com/keisuke-umezawa/4571551203eb02c1653c6cb36c571a4f/raw/d89c39dc73ede7c709ea93410797c56f2907a0e1/perf_lightgbm_integration_sql.py)

## results
### Perf on the suggested idea (e2efac2)

### Perf on the previous approach (569bf6e)

### Perf on master

keisuke-umezawa:
@toshihikoyanase I tested with sqlite. Could you check it?

toshihikoyanase:
@keisuke-umezawa I'm sorry for the delayed response. 

Thank you for sharing the benchmark results, but could you increase the number of trials (e.g., `10` to `100`)?
In some of the settings, `#trials` is less than `#interval`.
https://github.com/optuna/optuna/pull/2490#issuecomment-877745401

keisuke-umezawa:
@toshihikoyanase 
I ran the experiment for the other configurations. Could you check it again?

Especially, the latest PR is faster than the one which does not use pruners. #trials=100, #interval=32, #num_iterations=1000 vs #trials=100, #interval=0, #num_iterations=1000. It means that the pruner is useful in the following use case.

| #trials | #interval | #num_iterations | time(sec) | time/trial(sec) |
| ------- | --------- | --------- | --------- | --------------- |
| 100 | 0 | 1000 | 147.60+-3.54 | 1.476 |
| 100 | 32 | 1000 | 98.98+-18.76 | 0.990 |

> Thank you for sharing the benchmark results, but could you increase the number of trials (e.g., 10 to 100)?
In some of the settings, #trials is less than #interval.
btw, we do not need to compare #trials and #interval, but #num_iterations and #interval. So, I added the column for #num_iterations.

optuna/integration/lightgbm.py
toshihikoyanase: > Did you consider putting this in scope of the if statement above?

At first, I was a bit suspicious of the approach, but later I found that the change didn't affect the behavior of the pruners. This is mainly because the pruners check `trial.last_step` to fetch the latest intermediate values, but they do not change if we call `should_prune` without `report`. 

PercentilePruner (MedianPruner): https://github.com/optuna/optuna/blob/7c889c4a8c46cf8c8adc5a967472c0b4900bb428/optuna/pruners/_percentile.py#L188

SuccesiveHalvingPruner (HyperbandPruner):
https://github.com/optuna/optuna/blob/7c889c4a8c46cf8c8adc5a967472c0b4900bb428/optuna/pruners/_successive_halving.py#L165

ThresholdPruner
https://github.com/optuna/optuna/blob/7c889c4a8c46cf8c8adc5a967472c0b4900bb428/optuna/pruners/_threshold.py#L116

In case of `PatientPruner`, the list of `steps` is not changed if we do not call `report`, and `should_prune` keeps the same return values as the last call.
https://github.com/optuna/optuna/blob/7c889c4a8c46cf8c8adc5a967472c0b4900bb428/optuna/pruners/_patient.py#L88




I checked the actual pruning timing of `HyperbandPruner` and `MedianPruner` in the following notebook:
https://colab.research.google.com/drive/14e4H2aWkPGFBQgnOHDq9xmjMjVhnZ-LS?usp=sharing

> Is an interval for pruning even implemented in lightgbm?

I don't think so. We use the callback feature of lightgbm, and it calls the callbacks every iteration.

toshihikoyanase:
@keisuke-umezawa I'm sorry for the delayed response.

> btw, we do not need to compare #trials and #interval, but #num_iterations and #interval. 

You're absolutely right. Thank you for pointing it out.

optuna/integration/lightgbm.py
```diff
@@ -62,60 +63,85 @@ class LightGBMPruningCallback(object):
             Note that this argument will be ignored if you are calling
             `cv method <https://lightgbm.readthedocs.io/en/latest/Python-API.html#lightgbm.cv>`_
             instead of train method.
+        report_interval:
+            Check if the trial should be reported for pruning every n-th epoch.
+            By default ``report_interval=1`` and reporting is performed after every epoch.
+            Note that the pruning itself is performed according to the interval definition
+            of the pruner.
```
toshihikoyanase, line 70: - I guess a trial reports the intermediate values, but the first sentence says that the trial is reported.
-  "boosting iteration" or "boosting round" is more suitable for lightgbm than "epoch".
https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.train.html#lightgbm-train

```suggestion
        report_interval:
            Check if the trial should report intermediate values for pruning every n-th boosting iteration.
            By default ``report_interval=1`` and reporting is performed after every iteration.
            Note that the pruning itself is performed according to the interval definition
            of the pruner.
```

toshihikoyanase (commented):
@keisuke-umezawa 
I'm sorry for the late reply. I appreciate the detailed discussion with @Crissman about the scope of `report_interval`. I agree with the current approach.

I went through the change and it almost looks good to me. I added some small comments.

keisuke-umezawa:
@toshihikoyanase 
Thank you for reviewing! I followed your comments.

optuna/integration/lightgbm.py
```diff
@@ -62,60 +64,83 @@ class LightGBMPruningCallback(object):
             Note that this argument will be ignored if you are calling
             `cv method <https://lightgbm.readthedocs.io/en/latest/Python-API.html#lightgbm.cv>`_
             instead of train method.
+        report_interval:
+            Check if the trial should report intermediate values for pruning every n-th boosting iteration.
```
hvy, line 68: Seems like this line is too long. https://github.com/optuna/optuna/pull/2490/checks?check_run_id=3775102398

toshihikoyanase (commented):
Thank you for your update. LGTM except for https://github.com/optuna/optuna/pull/2490#discussion_r720974134.

keisuke-umezawa:
@toshihikoyanase Sorry, I fixed flake8.

toshihikoyanase:
Let me slightly modify the title.

PR #2493, opened 2021-03-17
Revert from fashion-mnist to mnist

Crissman:
## Motivation

It appears fashion-mnist has been removed from the [OpenML list of datasets](https://www.openml.org/search?type=data), so reverting back to mnist.


toshihikoyanase (commented):
I can see `Fashion-MNIST` on the following page. In my environment, `plot_study` works properly. Do we really need to replace it?
https://www.openml.org/d/40996

Crissman:
Thanks. It doesn't show up on the list of data sets, but is there when searched for. I got an error from the call for it yesterday, but it tested ok today. 🤷‍♂️

Closing! 🙇‍♂️

PR #2494, opened 2021-03-17
Use whitespace tokenizer instead of spacy tokenizer

himkt:
This PR updates an AllenNLP example to use a simple rule-based tokenizer instead of a spaCy tokenizer.

## Motivation
CI for example failed due to the extremely large log file (https://github.com/optuna/optuna/actions/runs/658018631).

This problem happens with the recent update of AllenNLP.
https://github.com/allenai/allennlp/issues/5036

## Description of the changes
- ae92576 replace a spaCy tokenizer with a simple whitespace tokenizer

### DIFF
--- a/examples/allennlp/classifier.jsonnet
+++ b/examples/allennlp/classifier.jsonnet
@@ -30,7 +30,7 @@ local ENCODER = CNN_FIELDS(
   dataset_reader: {
     type: 'text_classification_json',
     tokenizer: {
-      type: 'spacy',
+      type: 'whitespace',
     },
     token_indexers: {
       tokens: {


HideakiImamura (commented):
Is this a hot-fix? It would be better to note that it will be undone after https://github.com/allenai/allennlp/issues/5036.

himkt:
> hot-fix

No, we can use a rule-based tokenizer from now on.
When I added this example, I didn't consider what tokenizer is suitable for the example.

A proper tokenizer could improve the model's performance.
But it is sufficient to use a simple tokenizer for the simple example, I think.

crcrpar (approved):
Confirmed in my local environment:
```
...
[I 2021-03-17 23:07:08,874] Trial 4 pruned.
Number of finished trials:  5
Best trial:
  Value:  0.6162
  Params:
    DROPOUT: 0.01258586420887281
    EMBEDDING_DIM: 46
    HIDDEN_SIZE: 16
    MAX_FILTER_SIZE: 3
    NUM_FILTERS: 25

Created optimized AllenNLP config to `best_classifier.json`.

```

HideakiImamura (approved):
> A proper tokenizer could improve the model's performance.
But it is sufficient to use a simple tokenizer for the simple example, I think.

Thanks for the confirmation. I understood that this fix is a permanent one. LGTM.

PR #2501, opened 2021-03-18
Fix Warm Starting CMA-ES with a maximize direction

c-bata:
## Motivation

Fix  a bug of Warm Starting CMA-ES with a maximize direction.

## Description of the changes

Multiply `-1` if the direction is maximize.

### DIFF
--- a/optuna/samplers/_cmaes.py
+++ b/optuna/samplers/_cmaes.py
@@ -2,6 +2,7 @@ import copy
 import math
 import pickle
 from typing import Any
+from typing import cast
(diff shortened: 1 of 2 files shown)


HideakiImamura:
@not522 @himkt Could you review this PR if you have time?

himkt (approved):
I'm not confident about the detail of CMA-ES but approve because the test logic is almost the same as `test_warm_starting_cmaes` and adding support for maximization looks solid.

not522 (approved):
LGTM!

PR #2503, opened 2021-03-20
Add log-linear algorithm for 2d Pareto front.

parsiad:
## Motivation
`study.best_trials` can be made faster when there are only two objectives.

## Description of the changes
Uses a log-linear algorithm algorithm to construct the Pareto frontier in the case of two objectives that works as follows:

1. Sort trials by their x coordinate.
2. Iterate through the sorted trials and drop any trial that does not have a better y coordinate than the current best y coordinate.

### DIFF
--- a/optuna/_multi_objective.py
+++ b/optuna/_multi_objective.py
@@ -34,7 +34,7 @@ def _get_pareto_front_trials_2d(study: "optuna.study.BaseStudy") -> List[FrozenT
         if curr_x != x:
             set_mask(width, hi=i)
             width = 0
-        if y > best_y:
+        if y > best_y or (y == best_y and width == 0):
             continue
         if y < best_y:
             width = 0
--- a/optuna/_multi_objective.py
+++ b/optuna/_multi_objective.py
@@ -15,26 +15,23 @@ def _get_pareto_front_trials_2d(study: "optuna.study.BaseStudy") -> List[FrozenT
     if n_trials == 0:
         return []
 
-    numbered_trials = sorted(
-        enumerate(trials),
-        key=lambda pair: (
-            _normalize_value(pair[1].values[0], study.directions[0]),
-            _normalize_value(pair[1].values[1], study.directions[1]),
+    trials.sort(
+        key=lambda trial: (
+            _normalize_value(trial.values[0], study.directions[0]),
+            _normalize_value(trial.values[1], study.directions[1]),
         ),
     )
 
-    mask = [False] * n_trials
-
-    index, last_nondominated_trial = numbered_trials[0]
-    mask[index] = True
+    last_nondominated_trial = trials[0]
+    pareto_front = [last_nondominated_trial]
     for i in range(1, n_trials):
-        index, trial = numbered_trials[i]
+        trial = trials[i]
         if _dominates(last_nondominated_trial, trial, study.directions):
             continue
-        mask[index] = True
+        pareto_front.append(trial)
         last_nondominated_trial = trial
 
-    pareto_front = [trial for trial, keep in zip(trials, mask) if keep]
+    pareto_front.sort(key=lambda trial: trial.number)
     return pareto_front
 
 
--- a/optuna/_multi_objective.py
+++ b/optuna/_multi_objective.py
@@ -8,7 +8,46 @@ from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
 
 
-def _get_pareto_front_trials(study: "optuna.study.BaseStudy") -> List[FrozenTrial]:
+def _get_pareto_front_trials_2d(study: "optuna.study.BaseStudy") -> List[FrozenTrial]:
+    trials = [trial for trial in study.trials if trial.state == TrialState.COMPLETE]
+    points = sorted(
+        (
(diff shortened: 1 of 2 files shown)


crcrpar:
Thank you for the great improvement.

Tests are okay, and outputs also look solid.

Stable: https://optuna.readthedocs.io/en/stable/reference/visualization/generated/optuna.visualization.plot_pareto_front.html?highlight=pareto

This pr: https://66655-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/visualization/generated/optuna.visualization.plot_pareto_front.html#optuna.visualization.plot_pareto_front

keisuke-umezawa:
@HideakiImamura 
Could you also review it?

parsiad:
I have pushed another commit because the algorithm was missing an edge case. For example, if the directions are `['minimize', 'maximize']` and the input points are `[1, 1], [2,2], [3, 2]`, the algorithm would have erroneously included all of them instead of dropping `[3, 2]`. In general, if the x coordinate gets worse (`2 -> 3`) but the y coordinate stays the same (`2 -> 2`), the algorithm was erroneously not classifying the new point as dominated.

I tried to run tests twice, but it looks like they are failing due to an unrelated issue:
```E   OSError: libopenblas.so.0: cannot open shared object file: No such file or directory```
For what it's worth, tests pass locally on my machine.

**Comment by crcrpar** [comment]
It seems like the latest mxnet caused that error, and I confirmed the tests on which this pull request has some effect pass: https://github.com/optuna/optuna/pull/2503/checks?check_run_id=2161074877#step:8:36.

Sorry, but `Tests (Integration)` is flakier than `Tests` because it has many requirements.

**Comment by parsiad** [comment]
> Sorry, but `Tests (Integration)` is flakier than `Tests` because it has many requirements.

No worries. I'll just leave it as is unless you need a passing :heavy_check_mark: to merge.

**Comment by Crissman** [comment]
PR #2508 fixed the MXNet problem. Please re-base with this PR. 🙇‍♂️

**Review by Crissman — APPROVED** [review]
LGTM. Thank you for the PR!

**Comment by parsiad** [comment]
> PR #2508 fixed the MXNet problem. Please re-base with this PR. bowing_man

Done. Tests pass now.

Thank you for your review!

File: optuna/_multi_objective.py [hunk 1]
```diff
@@ -8,7 +8,46 @@
 from optuna.trial import TrialState
 
 
-def _get_pareto_front_trials(study: "optuna.study.BaseStudy") -> List[FrozenTrial]:
+def _get_pareto_front_trials_2d(study: "optuna.study.BaseStudy") -> List[FrozenTrial]:
```
HideakiImamura, line 11: How about sorting the trials themself, and using `_dominates` function as done in `_get_pareto_front_trials_nd`?
```python
    pareto_front = []
    trials = [trial for trial in study.trials if trial.state == TrialState.COMPLETE]
    trials.sort(key=lambda t: _normalize_value(t.values[1], study.directions[1]))
    trials.sort(key=lambda t: _normalize_value(t.values[0], study.directions[0]))

    last_nondominated_trial: Optional[FrozenTrial] = None
    for i in range(len(trials)):
        if i == 0:
            pareto_front.append(trials[0])
            last_nondominated_trial = trials[0]
            continue

        assert last_nondominated_trial is not None
        if not _dominates(last_nondominated_trial, trials[i], study.directions):
            pareto_front.append(trials[i])
            last_nondominated_trial = trials[i]
```
This code does not preserve the order of all trials and pareto front trials. so it will be fixed.

HideakiImamura (commented):
Thanks for the PR! I have a comment to improve the code readability. Could take a look?

optuna/_multi_objective.py
parsiad: I like the idea of using `_dominates` to simplify the code. I did something similar to what you suggested but because, as you point out, the order of trials needs to be preserved when filtering, I am still using `mask`.

optuna/_multi_objective.py
```diff
@@ -26,6 +56,12 @@ def _get_pareto_front_trials(study: "optuna.study.BaseStudy") -> List[FrozenTria
     return pareto_front
 
 
+def _get_pareto_front_trials(study: "optuna.study.BaseStudy") -> List[FrozenTrial]:
+    if len(study.directions) == 2:
+        return _get_pareto_front_trials_2d(study)  # Log-linear in number of trials.
+    return _get_pareto_front_trials_nd(study)  # Quadratic in number of trials.
```
g-votte, line 62: The non-2d scenario seems not to be covered by any test case in `test_study.py`. We may want to add new ones, since it's risky to leave the general logic not tested.

parsiad: Added a test case for 3d.

optuna/_multi_objective.py
HideakiImamura: To preserve the order of trials, we only need to sort the Pareto front trials with its `trial.number`. Could you remove the `mask`?

parsiad: Nice catch. Done.

HideakiImamura (approved):
LGTM!

PR #2508, opened 2021-03-22
Install OpenBLAS for mxnet1.8.0

crcrpar:
Install OpenBLAS for mxnet1.8.0 to successfully dlopen libopenblas.so.0.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -33,7 +33,7 @@ jobs:
     - name: Install (apt)
       run: |
         sudo apt-get update
-        sudo apt-get -y install openmpi-bin libopenmpi-dev
+        sudo apt-get -y install openmpi-bin libopenmpi-dev libopenblas-dev
     - name: Install (Python)
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -33,7 +33,7 @@ jobs:
     - name: Install (apt)
       run: |
         sudo apt-get update
-        sudo apt-get -y install openmpi-bin libopenmpi-dev h
+        sudo apt-get -y install openmpi-bin libopenmpi-dev
     - name: Install (Python)
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


hvy:
Thanks for the quick PR. In case it works, I'd suggest we leave comments.
By the way, https://pypi.org/project/mxnet/ states that `libquadmath0` is required. Is this a different topic?

hvy:
Could you also update the [coverage.yml](https://github.com/optuna/optuna/blob/master/.github/workflows/coverage.yml) ?

crcrpar:
Updated examples.yml as well as coverage.yml.

.github/workflows/examples.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     - name: Install (apt)
       run: |
         sudo apt-get update
-        sudo apt-get -y install openmpi-bin libopenmpi-dev
+        sudo apt-get -y install openmpi-bin libopenmpi-dev h
```
hvy, line 36: Is this `h` intentional?
```suggestion
        sudo apt-get -y install openmpi-bin libopenmpi-dev
```

crcrpar: not, I don't know where this h comes from and what this h means.

hvy (approved):
LGTM!

HideakiImamura (commented):
`examples` includes MXnet exampels, but `.github/workflows/examples.yml` is not updated. Is it intentional?

Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM python:${PYTHON_VERSION}
 ENV PIP_OPTIONS "--no-cache-dir --progress-bar off"
 
 RUN apt-get update \
-    && apt-get -y install openmpi-bin libopenmpi-dev \
+    && apt-get -y install openmpi-bin libopenmpi-dev libopenblas-dev\
```
toshihikoyanase, line 8: How about adding a whitespace before `\` like the other lines?
```suggestion
    && apt-get -y install openmpi-bin libopenmpi-dev libopenblas-dev \
```

HideakiImamura (approved):

PR #2516, opened 2021-03-24
Support object representation of `StudyDirection` for `create_study` arguments

g-votte:
This PR enables to pass `StudyDirection` objects to the `direction`/`directions` argument of `optuna.create_study`.

## Motivation
Supporting `StudyDirection` objects sometimes simplifies the user’s code especially when the user already has the `StudyDirection` objects when creating the study, e.g., copying an existing study to new one as follows.

```python
study_src = optuna.load_study(study_name=study_name, storage=storage_url)
study_dist = optuna.create_study(directions=study_src.directions)

for t in study_src.trials:
    study_dist.add_trial(t)
```
Note that the code above raises an error in the second line without applying this PR, and the user has to convert the direction objects to the string representation.

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -1111,15 +1111,9 @@ def create_study(
     ):
         raise ValueError("Please set either 'minimize' or 'maximize' to direction.")
 
-    direction_objects = []
-    for d in directions:
-        if d == "minimize":
-            direction_objects.append(StudyDirection.MINIMIZE)
-        elif d == "maximize":
-            direction_objects.append(StudyDirection.MAXIMIZE)
-        else:
-            assert isinstance(d, StudyDirection)
-            direction_objects.append(d)
+    direction_objects = [
+        d if isinstance(d, StudyDirection) else StudyDirection[d.upper()] for d in directions
+    ]
 
     storage = storages.get_storage(storage)
     try:
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -1109,7 +1109,10 @@ def create_study(
         d not in ["minimize", "maximize", StudyDirection.MINIMIZE, StudyDirection.MAXIMIZE]
         for d in directions
     ):
-        raise ValueError("Please set either 'minimize' or 'maximize' to direction.")
+        raise ValueError(
+            "Please set either 'minimize' or 'maximize' to direction. You can also set the "
+            "corresponding `StudyDirection` member."
+        )
 
     direction_objects = [
         d if isinstance(d, StudyDirection) else StudyDirection[d.upper()] for d in directions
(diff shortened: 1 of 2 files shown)


optuna/study.py
```diff
@@ -1104,12 +1105,21 @@ def objective(trial):
...
-        StudyDirection.MINIMIZE if d == "minimize" else StudyDirection.MAXIMIZE for d in directions
-    ]
+    direction_objects = []
+    for d in directions:
+        if d == "minimize":
+            direction_objects.append(StudyDirection.MINIMIZE)
+        elif d == "maximize":
+            direction_objects.append(StudyDirection.MAXIMIZE)
+        else:
+            assert isinstance(d, StudyDirection)
```
g-votte, line 1121: I've added this assertion to pass the mypy check. Please let me know if there's much smarter way.

hvy: Not sure this addresses your comment but we could also write something like this.
```python
direction_objects = [d if isintance(d, StudyDirection) else StudyDirection[d.upper()] for d in directions]
```

hvy (approved):
Thanks, LGTM! Left a minor comment but it's mostly a cosmetic.

optuna/study.py
g-votte: Cool! I think this is much simpler and resolves the redundant assertion.

optuna/study.py
```diff
@@ -1104,11 +1105,14 @@ def objective(trial):
 
     if len(directions) < 1:
         raise ValueError("The number of objectives must be greater than 0.")
-    elif any(d != "minimize" and d != "maximize" for d in directions):
+    elif any(
+        d not in ["minimize", "maximize", StudyDirection.MINIMIZE, StudyDirection.MAXIMIZE]
+        for d in directions
+    ):
         raise ValueError("Please set either 'minimize' or 'maximize' to direction.")
```
toshihikoyanase, line 1112: A nit, but how about adding `StudyDirection` to the error message?
```suggestion
    ):
        raise ValueError(
            "Please set one of 'minimize', 'maximize' or a StudyDirection member to direction."
        )
```

toshihikoyanase (commented):
Thank you for your PR. This feature definitely improves the usability of `optuna.create_study`.
The change looks good to me, but I have a small comment on the error message.

optuna/study.py
g-votte: Thanks for your suggestion! I updated the error message.

toshihikoyanase: Thank you for your update. It's better than my proposal 👍

toshihikoyanase (approved):

PR #2525, opened 2021-03-26
Extract test cases regarding Pareto front to `test_multi_objective.py`

g-votte:
This PR extracts 2 test cases regarding Pareto front from `test_study.py` to `test_multi_objective.py`.

## Motivation
I believe this PR makes the codebase more organized and readable because of the following reasons:

- `_multi_objective.py` is mostly responsible to the internal logic of `best_trials`, and the 2d/nd condition highly depends on the internal implementation as well. Thus, those test cases should be put on the corresponding namespace of the filename.
- Extracting those scenarios reduces the number of test cases in `test_study.py`, which is already huge.

## Description of the changes
- Move the test cases of getting Pareto front from `test_study.py` to `test_multi_objective.py`
- Leave the minimal test scenario of `best_trials` in `test_study.py` to test the API call

### DIFF
--- a/tests/study_tests/test_study.py
+++ b/tests/study_tests/test_study.py
@@ -9,7 +9,6 @@ from typing import Callable
 from typing import Dict
 from typing import List
 from typing import Optional
-from typing import Tuple
 from unittest.mock import Mock  # NOQA
 from unittest.mock import patch
 import uuid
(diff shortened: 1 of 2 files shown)


toshihikoyanase:
@sile @himkt Could you review this PR?

himkt (approved):
LGTM. Thank you for cleaning `test_study.py`.

sile (approved):
LGTM!

Issue #2517, opened 2021-03-24
Adding pytest's `filterwarnings` causes 0% coverage


g-votte:
## Steps to reproduce

1. Apply the change of https://github.com/optuna/optuna/pull/2405 , which adds `filterwarnings` of pytest.
2. Run pytest with coverage reporting

## Current behavior

The coverage drastically downs and shows [0% in the CI](https://app.codecov.io/gh/optuna/optuna/compare/2405/overview).

## Expected behavior

The coverage does not change.

## Reproducible examples

You can quickly reproduce the drastic decrease by locally running the following command and checking the output report.

```console
pytest --cov=optuna --cov-report=html tests/study_tests/test_study.py
```

## Additional context (optional)

- https://github.com/optuna/optuna/pull/2405 was reverted in https://github.com/optuna/optuna/pull/2413 due to the coverage decrease.
- We want to re-apply the `filterwarning` after resolving this issue, which removes a lot of noises in the unit tests' log.
- This issue also blocks https://github.com/optuna/optuna/pull/2416.
- This issue was also investigated in https://github.com/optuna/optuna/pull/2417 . However, adding parentheses did not resolve the problem.

g-votte:
Some more attempts have been made in #2529. However, neither changing the `pytest` version nor adding the `--cov-append` option did not solve the problem.

jeromepatel:
Actually, I tried to solve the issue by trying different things but none of them did work. Here's the summary of what I have tried:
* Changed configuration to `tox.ini` 
* Added missing `__init.py__` file inside a test subdirectory 
* Removed coverage for `mpirun` (multi-node test)
* Changed `pytest` version and `pytest-cov` version (separately )  </br>      
Unfortunately, none of them resolved the current issue. I apologize for that.

PR #2529, opened 2021-03-27
[WIP] Fix filterwarnings issue by adding cov-append option

gasin:
## Motivation
fix this bug
https://github.com/optuna/optuna/issues/2517

## Description of the changes
Adding cov-append option may fix this problem.
At least, the coverage does not change in my environment.

Reference
https://github.com/pytest-dev/pytest-cov/issues/437


gasin:
"Coverage.py warning: No data was collected. (no-data-collected)"

This error causes this problem, and it seems hard for me.
I close this PR now, sorry.

g-votte:
Thanks for your contribution! Reported those results in the original issue https://github.com/optuna/optuna/issues/2517#issuecomment-809800444

PR #2534, opened 2021-03-29
Divide `examples` CI workflow

toshihikoyanase:
## Motivation

All examples in the `examples` directory are executed in the [examples workflow](https://github.com/optuna/optuna/actions/workflows/examples.yml) of GitHub Actions. Recently, we faced problems as follows:
- We can only execute the CI job once a day since it takes several hours,
- It failed for 20 days because of the update of the dependencies,
- `examples.py` has a bit complicated rules like `IGNORES`, 
- `setup.py` also has a bit complicated version constriants. Also, it cannot handle collision of requirements. For example, AllenNLP does not support PyTorch 1.8.0, but we'd like to test PyTorch examples with PyTorch 1.8.0.

To resolve the issues, this PR split workflows. It will mitigates the problems as follows:
- 👍 We can execute an individual example if contributors update the example.
- 👍 We can execute example CI jobs in parallel, and we can notice errors in multiple examples.
- 👍 It simplifies `.github/workflow/example.yml` and `setup.py`

It also has possible disadvantages.
- 🤔 The number of files increase (`examples/**/requirements.txt` and `.github/workflow/z_*.yml`).
- 🤔 Need to update multiple files if a dependent library is updated. For example, PyTorch appears in several requirements.txt such as `examples/pytorch/requirements.txt` and `examples/multi_objective/requirements.txt`

Please feel free to add your comments on the design.

## Description of the changes

- Create a workflow for each library
- Create `requirements.txt` in a directory of each library
- Remove the libraries in `examples/**/requirements.txt` from `setup.py['example']`


toshihikoyanase:
@himkt @HideakiImamura I resolved conflicts. Could you review this PR?

.github/workflows/z_fastai.yml
```diff
@@ -0,0 +1,42 @@
+name: z_fastai
```
crcrpar, line 1: If I recall correctly, fastaiv2 example didn't work as reported in https://github.com/optuna/optuna/issues/2107.
But now it seems working.

If this observation is correct, could you close the issue mentioned above once this pull request lands?

.github/workflows/z_fastai.yml
```diff
@@ -0,0 +1,42 @@
...
+
+        # Install minimal dependencies and confirm that `import optuna` is successful.
+        pip install --progress-bar off $(ls dist/*.tar.gz)
+        python -c 'import optuna'
+
+        pip install -r examples/fastai/requirements.txt
+    - name: Run examples
+      run: |
+        python examples/fastai/fastaiv2_simple.py 
+        python examples/fastai/fastaiv2_simple.py --pruning
```
HideakiImamura, line 40: The `examples/fastai/fastaiv1_simple.py` is removed. Is it intended?

.github/workflows/z_multi_objective.yml
```diff
@@ -0,0 +1,46 @@
...
+
+        # Install minimal dependencies and confirm that `import optuna` is successful.
+        pip install --progress-bar off $(ls dist/*.tar.gz)
+        python -c 'import optuna'
+
+        pip install -r examples/multi_objective/requirements.txt
+    - name: Run examples
+      run: |
+        python examples/multi_objective/botorch_simple.py
+        python examples/multi_objective/pytorch_simple.py
```
HideakiImamura, line 44: These examples are executed in `.github/workflows/examples.yaml`. I think we don't have to execute twice.

HideakiImamura: Again, I think we can run `botorch_simple.py` separately.

HideakiImamura (commented):
Thanks for the great improvement of examples CI! I basically agree with the idea of splitting the examples CI workflow. I have some comments. PTAL.

.github/workflows/examples.yml
```diff
@@ -48,63 +44,34 @@ jobs:
...
       run: |
-        STORAGE_URL=sqlite:///example.db
-        for file in `find examples -name 'chainermn_*.py'`
-        do
-          echo $file
-          STUDY_NAME=`optuna create-study --storage $STORAGE_URL`
-          mpirun -n 2 -- python $file $STUDY_NAME $STORAGE_URL > /dev/null
-        done
-        mpirun -n 2 -- python examples/pytorch/pytorch_distributed_simple.py
+        python examples/multi_objective/botorch_simple.py
```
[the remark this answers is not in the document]
himkt: Agree with @HideakiImamura. I think we can create `z_botorch.yml`. Furthermore, can we move `pytorch_simple.py` (should be renamed) to `pytorch` directory and run it in `z_pytorch.yml`?

himkt:
@toshihikoyanase 
We decided to separate examples from optuna/optuna. Can we close this PR now?

toshihikoyanase:
@himkt You're absolutely right. Thank you for letting me know.

Issue #2536, opened 2021-03-29
Add `__repr__` to sampler classes.


toshihikoyanase:
## Motivation

When we use experiment management frameworks like Hydra with Optuna, we sometimes want to log the Optuna's settings including sampler types. Currently, the sampler class does not have `__repr__` and the messages are less human-friendly if we simply print the sampler instantance:

```console
[HYDRA] Sampler: <optuna.samplers._tpe.sampler.TPESampler object at 0x7fd7901d6d30>
```

Adding the `__repr__`, we can simplify the logging in the experiment management. I think we can find similar requests during ML pipeline development.

c.f., https://github.com/facebookresearch/hydra/pull/1503

## Description

Add `__repr__` to `optuna.samplers` classes.
I guess the minimum requirement is the class name as can be seen in https://github.com/facebookresearch/hydra/pull/1503.

```python
class RandomSampler(BaseSampler):
    ...
    def __repr__(self):
        return self.__class__.__name__
```

## Additional context (optional)

See also @omry's comment: https://github.com/facebookresearch/hydra/pull/1503#discussion_r602631532

toshihikoyanase:
#2539 implemented `__str__` for the sampler classes. We have following follow-up tasks:

- [ ] Add unit test for `__str__`
- [ ] Add documentation for `__str__`

omry:
What I had in mind is something more informative.
For example, for a hypothetical FooSampler that takes:
```
seed: int, alpha: float
```
as parameter, a good sting representation can be:
```python
FooSampler(seed=123, alpha=0.9)
```

toshihikoyanase:
@omry Thank you for your feedback. @nzw0301 's suggestion (https://github.com/optuna/optuna/pull/2539#issuecomment-810095018) may fit your proposal, but I guess it may be too detailed to show all properties. What do you think of it?

```python
def __repr__(self) -> str:

    kwargs = ", ".join("{}={}".format(k, v) for k, v in sorted(self.__dict__.items()))
    return "{}({})".format(self.__class__.__name__, kwargs)

[...]

PR #2539, opened 2021-03-29
Add `__str__` for samplers

nzw0301:
## Motivation

https://github.com/optuna/optuna/issues/2536

## Description of the changes

Add `__str__` to `BaseSampler`.

Example: 

```python
import optuna


study = optuna.study.create_study(sampler=optuna.samplers.CmaEsSampler())

print(study.sampler)

# CmaEsSampler
```

### DIFF
--- a/optuna/samplers/_base.py
+++ b/optuna/samplers/_base.py
@@ -42,6 +42,9 @@ class BaseSampler(object, metaclass=abc.ABCMeta):
 
     """
 
+    def __str__(self) -> str:
+        return self.__class__.__name__
+
     @abc.abstractmethod
     def infer_relative_search_space(
         self, study: Study, trial: FrozenTrial


toshihikoyanase (commented):
Thank you for your swift action! The change is simple and straightforward, but let me add some discussions before starting reviews.

Please feel free to add comments on them.

### `__repr__` or `__str__`

#### [`__repr__`](https://docs.python.org/3/reference/datamodel.html?highlight=__repr__#object.__repr__)

> If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form `<...some useful description...>` should be returned.

I don't think we can use `__repr__` to recreate samplers since some samplers like `NSGAIISampler` has functions as the arguments (e.g., `constraints_func`). So, we'll provide useful description with this method.

#### [`__str__`](https://docs.python.org/3/reference/datamodel.html?highlight=__repr__#object.__str__)

> Called by `str(object)` and the built-in functions `format()` and `print()` to compute the “informal” or nicely printable string representation of an object. The return value must be a string object.

`__str__` can also be used to format `print(sampler)`, but I think we can simply use `__repr__` if we do not provide reproducible strings with `__repr__`.

### Module path

Currently, this method returns only class name. We can show a module path in addition to the class name, but I don't think it is mandatory since the sampler class names are identical now. (We have two `CmaEsSampler`s, but one of them was renamed `PyCmaEsSampler`.)

hvy:
I haven't properly followed the discussion but I'm also wondering if `__str__` won't suffice. It seems like we want to define a printable string representation rather than a unique/reconstructible representation of the objects?

toshihikoyanase:
@hvy Thank you for your comments.

> I'm also wondering if `__str__` won't suffice. It seems like we want to define a printable string representation rather than a unique/reconstructible representation of the objects?

Sounds reasonable. I missed the view point of uniqueness. Even though it may be difficult for us to provide reconstructable strings, we can still show unique strings by `__repr__`.

nzw0301:
Thank you for the fruitful discussion. By following `distrubtions.*`'s ``__repre__``, showing all arguments is supposed to be more informative.

```python
def __repr__(self) -> str:

    kwargs = ", ".join("{}={}".format(k, v) for k, v in sorted(self.__dict__.items()))
    return "{}({})".format(self.__class__.__name__, kwargs)
```

Results: 

```python
import optuna


study = optuna.study.create_study(sampler=optuna.samplers.CmaEsSampler())

print(study.sampler)

# CmaEsSampler(_cma_rng=RandomState(MT19937), _consider_pruned_trials=False, _inc_popsize=2, _independent_sampler=RandomSampler(_rng=RandomState(MT19937)), _n_startup_trials=1, _restart_strategy=None, _search_space=<optuna.samplers._search_space.IntersectionSearchSpace object at 0x7f92e1ab5640>, _sigma0=None, _source_trials=None, _use_separable_cma=False, _warn_independent_sampling=True, _x0=None)
```

hvy:
Assigned @toshihikoyanase and myself. Sounds fair to include the arguments. Again, I didn't quite understand why we prefer `__repr__` over `__str__`. Is there a reason we should not use the latter ?

nzw0301:
Thank you for your feedback. I misunderstood the previous comments. I agree with your suggestion: `__str__` is more appropriate than ``__repr__`.

hvy (approved):
LGTM!

Just a note that if we are to make this a proper public API, we should also add unit tests (and also document it). This looks good to me otherwise! I guess this naturally also raises the question whether we want to define similar behaviors for pruners and alike. But I don't see why we would rush.

toshihikoyanase (approved):
LGTM!

I reported the issue #2536 to create log messages about samplers intuitively, and this PR will changes, for example, Hydra Optuna plugin as follows:

[before](https://github.com/facebookresearch/hydra/blob/1cd3f5a72723216027d7e02663f06ed494726622/plugins/hydra_optuna_sweeper/hydra_plugins/hydra_optuna_sweeper/_impl.py#L191
): `log.info(f"Sampler: {type(self.sampler).__name__}")`
after: `log.info(f"Sampler: {self.sampler}")`

> define similar behaviors for pruners and alike.

I agree with you. We can make the code to generate the log message simpler if we implement `__str__` of pruners and storages. I'll update #2536.

PR #2541, opened 2021-03-30
Extract heartbeat-related functions for ask-and-tell interface

toshihikoyanase:
## Motivation

#2190 and #2347 added the heartbeat feature to `RDBStorage`. It is currently enabled when users start optimization using `Study.optimize`, and it is difficult for ask-and-tell API users to use the heatrbeat. This is mainly because the heartbeat implementation is embedded in a private function [`run_trial` in `optuna/_optimize.py`](https://github.com/optuna/optuna/blob/cda6136d5276a8144d5216a49a3f2f3c356bd77c/optuna/_optimize.py#L185).

## Description of the changes

This PR extracts the implementation of the heartbeat feature to the following functions:

- `run_failed_trial_callback`: execute the `failed_trial_callback` functions specified as the argument of `RDBStorage.__init__`. It should be called before `study.ask()`.
- `create_heartbeat_thread`: create and start a heartbeat thread. It is expected to call after `Study.ask`. It is called after `study.ask()`.
- `join_heartbeat_thread`: stop the heatbeat. It is expected to call after objective function evaluation. It is called after objective function evaluation.

Since these methods are not called in `Study.ask` or `Study.tell`, users have to call them by themselves.

```python
    for _ in range(N_TRIALS):
        optuna.study.run_failed_trial_callback(study)  # Here
        trial = study.ask()
        thread, stop_event = optuna.study.create_heartbeat_thread(study, trial)  # Here

        value = objective(trial)
        study.tell(trial, value)
        optuna.study.join_heartbeat_thread(study, thread, stop_event)  # Here
```

Alternatively, we might be able to call them in `Study.ask()` or `Study.tell()`, but the heartbeat threads may be not joined if users want to call ask and tell in the different process.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -99,12 +99,21 @@ jobs:
         done
       env:
         OMP_NUM_THREADS: 1
-    - name: Run PyTorch checkpoint example
+    - name: Run PyTorch checkpoint example (optimize)
       run: |
         # Unset errexit temporarily since timeout returns exit code 124.
         set +e
         timeout 20 python examples/pytorch/pytorch_checkpoint.py > /dev/null
         set -e
         python examples/pytorch/pytorch_checkpoint.py > /dev/null
+      env:
+        OMP_NUM_THREADS: 1
+    - name: Run PyTorch checkpoint example (ask-and-tell)
+      run: |
+        # Unset errexit temporarily since timeout returns exit code 124.
+        set +e
+        timeout 20 python examples/pytorch/pytorch_ask_and_tell_checkpoint.py > /dev/null
+        set -e
+        python examples/pytorch/pytorch_ask_and_tell_checkpoint.py > /dev/null
       env:
         OMP_NUM_THREADS: 1
--- a/examples/pytorch/pytorch_ask_and_tell_checkpoint.py
+++ b/examples/pytorch/pytorch_ask_and_tell_checkpoint.py
@@ -1,0 +1,228 @@
+"""
+Optuna example that optimizes multi-layer perceptrons using PyTorch with checkpoint.
+
+In this example, we optimize the validation accuracy of hand-written digit recognition using
+PyTorch and FashionMNIST. We optimize the neural network architecture as well as the optimizer
+configuration. As it is too time consuming to use the whole FashionMNIST dataset,
+we here use a small subset of it.
+
+Even if the process where the trial is running is killed for some reason, you can restart from
+previous saved checkpoint using heartbeat.
+
+    $ timeout 20 python examples/pytorch/pytorch_checkpoint.py
+    $ python examples/pytorch/pytorch_checkpoint.py
+"""
+
+import copy
+import datetime
+import os
+from threading import TIMEOUT_MAX
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+import torch.utils.data
+from torchvision import datasets
+from torchvision import transforms
+
+import optuna
+
+
+DEVICE = torch.device("cpu")
+BATCHSIZE = 128
+CLASSES = 10
+DIR = os.getcwd()
+EPOCHS = 10
+LOG_INTERVAL = 10
+N_TRAIN_EXAMPLES = BATCHSIZE * 30
+N_VALID_EXAMPLES = BATCHSIZE * 10
+
+
+def define_model(trial):
+    # We optimize the number of layers, hidden units and dropout ratio in each layer.
+    n_layers = trial.suggest_int("n_layers", 1, 3)
+    layers = []
+
+    in_features = 28 * 28
+    for i in range(n_layers):
+        out_features = trial.suggest_int("n_units_l{}".format(i), 4, 128)
+        layers.append(nn.Linear(in_features, out_features))
+        layers.append(nn.ReLU())
+        p = trial.suggest_float("dropout_l{}".format(i), 0.2, 0.5)
+        layers.append(nn.Dropout(p))
+
+        in_features = out_features
+    layers.append(nn.Linear(in_features, CLASSES))
+    layers.append(nn.LogSoftmax(dim=1))
+
+    return nn.Sequential(*layers)
+
+
+def get_mnist():
+    # Load FashionMNIST dataset.
+    train_loader = torch.utils.data.DataLoader(
+        datasets.FashionMNIST(DIR, train=True, download=True, transform=transforms.ToTensor()),
+        batch_size=BATCHSIZE,
+        shuffle=True,
+    )
+    valid_loader = torch.utils.data.DataLoader(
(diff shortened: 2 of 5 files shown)


toshihikoyanase:
The example works as expected. The trial 5 resumes training from 4th epoch in the following figure:
image

(Trial 4 was not plotted in the figure because the state is FAILED. I want an option to show intermediate values of FAILED trials in this case.)

toshihikoyanase:
@sile @keisuke-umezawa Could you review this PR, please?

toshihikoyanase:
@Crissman Could you check the docstring, please? Users need to call the heartbeat-related function at the right timing, and we should  explain it clearly. If you kindly check the logic too, it would be a great help of us.

examples/pytorch/pytorch_ask_and_tell_checkpoint.py
```diff
@@ -0,0 +1,227 @@
+"""
+Optuna example that optimizes multi-layer perceptrons using PyTorch with checkpoint.
+
+In this example, we optimize the validation accuracy of hand-written digit recognition using
+PyTorch and FashionMNIST. We optimize the neural network architecture as well as the optimizer
+configuration. As it is too time consuming to use the whole FashionMNIST dataset,
+we here use a small subset of it.
```
Crissman, line 7:
```suggestion
we use a small subset.
```

examples/pytorch/pytorch_ask_and_tell_checkpoint.py
```diff
@@ -0,0 +1,227 @@
...
+
+Even if the process where the trial is running is killed for some reason, you can restart from
```
Crissman, line 9:
```suggestion
Even if the process where the trial is running is killed for some reason, it can be restarted from
```

optuna/_optimize.py
```diff
@@ -338,3 +327,89 @@ def _record_heartbeat(trial_id: int, storage: storages.BaseStorage, stop_event:
...
+
+@experimental("2.7.0")
+def create_heartbeat_thread(
+    study: "optuna.study.Study", trial: "optuna.trial.Trial"
+) -> Optional[Tuple[Thread, Event]]:
+    """Create a thread for heartbeat.
+
+    This is a utility function for the heartbeat feature of :class:`~optuna.storage.RDBStorage`,
+    and it is expected to be used with :meth:`~optuna.study.Study.ask` and
+    :meth:`~optuna.study.Study.tell`. To observe heartbeat, please call this function after
```
Crissman, line 364:
```suggestion
    :meth:`~optuna.study.Study.tell`. To observe the heartbeat, please call this function after
```

optuna/_optimize.py
```diff
@@ -338,3 +327,89 @@ def _record_heartbeat(trial_id: int, storage: storages.BaseStorage, stop_event:
...
+
+@experimental("2.7.0")
+def join_heartbeat_thread(
+    study: "optuna.study.Study", thread: Optional[Thread], stop_event: Optional[Event]
+) -> None:
+    """Join a thread for heartbeat.
+
+    This is a utility function for the heartbeat feature of :class:`~optuna.storage.RDBStorage`,
+    and it is expected to be used with :meth:`~optuna.study.Study.ask` and
+    :meth:`~optuna.study.Study.tell`. To observe heartbeat, please call
```
Crissman, line 398:
```suggestion
    :meth:`~optuna.study.Study.tell`. To observe the heartbeat, please call
```

Crissman (changes requested):
Minor English edits. Other LGTM.

toshihikoyanase:
@Crissman Thank you for proofreading the docstrings. I applied your suggestions.
Could you also check the logic, please?

.github/workflows/examples.yml
```diff
@@ -108,3 +108,12 @@ jobs:
         python examples/pytorch/pytorch_checkpoint.py > /dev/null
       env:
         OMP_NUM_THREADS: 1
+    - name: Run PyTorch checkpoint example (ask-and-tell)
+      run: |
+        # Unset errexit temporarily since timeout returns exit code 124.
```
Crissman, line 113: Is this to download the FashionMNIST dataset?

Crissman:
Logic looks good to me. Tested by running locally, and also replicated the stop study midway, and recover from checkpoint. 👍

.github/workflows/examples.yml
toshihikoyanase: No, it isn't. It is for `timeout` command.
The `timeout` command returns non-zero exit code when the timeout happens, and the code stops the script execution without `set +e`.

Crissman (approved):
LGTM

examples/pytorch/pytorch_ask_and_tell_checkpoint.py
```diff
@@ -0,0 +1,227 @@
...
+            print(
+                f"Trial {trial.number} finished with value: {value} and parameters: "
+                f"{trial.params}. Best is trial {study.best_trial.number} with "
+                f"value: {study.best_value}."
+            )
+        except optuna.TrialPruned:
+            print(f"Trial {trial.number} was pruned.")
+            study.tell(trial, None, optuna.trial.TrialState.PRUNED)
+
+        optuna.study.join_heartbeat_thread(study, thread, stop_event)
```
sile, line 203: I think that this call should be placed within a `finally` block to prevent thread leaks.

sile: Furthermore, if possible, I prefer using python `Context` to create and stop those threads rather than bare function calls (as manual management seems more error-prone).

The code below is a pseudo code to share my thought:
```python
class TrialHeartbeat:
    def __init__(self, trial):
        self._trial = trial

    def __enter__(self):
        optuna.study.run_failed_trial_callback(self._trial.study)  # I think that it's okay to call this function if it's just after the trial creation.
        self._thread, sefl._stop_event = optuna.study.create_heartbeat_thread(self._trial.study, self._trial)
        return self

    def __exit__(self, ...):
       optuna.study.join_heartbeat_thread(sefl._trial.study, self._thread, self._stop_event)


# Usage
trial = study.ask()
with TrialHeartbeat(trial):
    ...
    study.tell(trial, objective_value)
```

sile (changes requested):

examples/pytorch/pytorch_ask_and_tell_checkpoint.py
toshihikoyanase: > Furthermore, if possible, I prefer using python `Context` to create and stop those threads rather than bare function calls (as manual management seems more error-prone).

Thank you for your suggestion. That sounds reasonable to me.

Users may sometimes calls `ask` and `tell` in different processes, but the current heartbeat related functions are supposed to be called in a single process.  `Context` will prevent users to call the functions in different processes.

I'll update the PR based on python `Context`.

Crissman:
Dismissed my review since the plan is to rewrite this based on a context manager implementation. 🙇  Ready to review again when ready!

PR #2547, opened 2021-03-31
[Tutorial] Mention the paper is in Japanese

crcrpar:
## Motivation

Some readers would assume the paper is written in English, but it's in Japanese, which might let them down slightly.
So, I want to clarify the fact.


toshihikoyanase:
@Crissman Could you review this PR? The change is limited, and you can merge it with a single approval.

Crissman (approved):
LGTM!

PR #2558, opened 2021-04-02
Concatenate the argument text after the deprecation warning

not522:
## Motivation
To display the additional text in the warning message is helpful for the migration from the deprecated feature.
## Description of the changes
Concatenate the argument text after the deprecation warning.

### DIFF
--- a/optuna/_deprecated.py
+++ b/optuna/_deprecated.py
@@ -21,6 +21,13 @@ _DEPRECATION_NOTE_TEMPLATE = """
 """
 
 
+_DEPRECATION_WARNING_TEMPLATE = (
+    "{name} has been deprecated in v{d_ver}. "
+    "This feature will be removed in v{r_ver}. "
+    "See https://github.com/optuna/optuna/releases/tag/v{d_ver}."
+)
+
+
 def _validate_two_version(old_version: str, new_version: str) -> None:
     if version.parse(old_version) > version.parse(new_version):
         raise ValueError(
@@ -101,14 +108,10 @@ def deprecated(
             # TODO(mamu): Annotate this correctly.
             @functools.wraps(func)
             def new_func(*args: Any, **kwargs: Any) -> Any:
-                message = (
-                    "{0} has been deprecated in v{1}. "
-                    "This feature will be removed in v{2}. "
-                    "See https://github.com/optuna/optuna/releases/tag/v{1}.".format(
-                        name if name is not None else func.__name__,
-                        deprecated_version,
-                        removed_version,
-                    )
+                message = _DEPRECATION_WARNING_TEMPLATE.format(
+                    name=(name if name is not None else func.__name__),
+                    d_ver=deprecated_version,
+                    r_ver=removed_version,
                 )
                 if text is not None:
                     message += " " + text
@@ -127,14 +130,10 @@ def deprecated(
 
             @functools.wraps(_original_init)
             def wrapped_init(self, *args, **kwargs) -> None:  # type: ignore
-                message = (
-                    "{0} has been deprecated in v{1}. "
-                    "This feature will be removed in v{2}. "
-                    "See https://github.com/optuna/optuna/releases/tag/v{1}.".format(
-                        name if name is not None else cls.__name__,
-                        deprecated_version,
-                        removed_version,
-                    )
+                message = _DEPRECATION_WARNING_TEMPLATE.format(
+                    name=(name if name is not None else cls.__name__),
+                    d_ver=deprecated_version,
+                    r_ver=removed_version,
                 )
                 if text is not None:
                     message += " " + text
--- a/optuna/_deprecated.py
+++ b/optuna/_deprecated.py
@@ -101,17 +101,18 @@ def deprecated(
             # TODO(mamu): Annotate this correctly.
             @functools.wraps(func)
             def new_func(*args: Any, **kwargs: Any) -> Any:
-                warnings.warn(
+                message = (
                     "{0} has been deprecated in v{1}. "
                     "This feature will be removed in v{2}. "
                     "See https://github.com/optuna/optuna/releases/tag/v{1}.".format(
                         name if name is not None else func.__name__,
                         deprecated_version,
                         removed_version,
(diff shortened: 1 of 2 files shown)


optuna/_deprecated.py
```diff
@@ -101,17 +101,18 @@ def _deprecated_func(func: Callable[[Any], Any]) -> Callable[[Any], Any]:
...
                         deprecated_version,
                         removed_version,
-                    ),
-                    FutureWarning,
-                    stacklevel=2,
+                    )
                 )
+                if text is not None:
+                    message += " " + text
+                warnings.warn(message, FutureWarning, stacklevel=2)
```
toshihikoyanase, line 115: Could you add a unit test to confirms that `text` is actually added to the end of the warning message?

```python
@pytest.mark.parametrize("text", [None, "", "test", "test" * 100])
def test_deprecation_text_specified(text: Optional[str]) -> None:

    ...
    with pytest.warns(FutureWarning) as record:
        decorated_func()

    assert isinstance(record.list[0].message, Warning)
    if text is not None:
        assert record.list[0].message.args[0].endswith(text)

```

optuna/_deprecated.py
```diff
@@ -126,14 +127,19 @@ def _deprecated_class(cls: Any) -> Any:
...
                         name if name is not None else cls.__name__,
                         deprecated_version,
                         removed_version,
-                    ),
+                    )
+                )
+                if text is not None:
+                    message += " " + text
+                warnings.warn(
+                    message,
```
toshihikoyanase, line 142: ditto.

toshihikoyanase (commented):
Thank you for your PR. The change basically looks good to me.
I added comments on the tests.

hvy (approved):
Thanks for the change, this should make the decorator more intuitive to use and the messages more friendly. 

It LGTM if @toshihikoyanase 's comment are addressed so let me approve the changes. (I'm feeling that some code could be DRY:ed, it's not a change request however)

toshihikoyanase:
The following is the actual warning messages with the `text` arguments. I think the message are informative for users.

```python
optuna/integration/_lightgbm_tuner/optimize.py

    tuner = lgb.LightGBMTuner(
        params, dtrain, valid_sets=[dtrain, dval], verbose_eval=100, early_stopping_rounds=100
    )

    tuner.best_booster()

examples/lightgbm/lightgbm_tuner_simple.py:33: FutureWarning: best_booster has been deprecated in v1.4.0. This feature will be removed in v3.0.0. See https://github.com/optuna/optuna/releases/tag/v1.4.0. Please get the best booster via :class:`~optuna.integration.lightgbm.LightGBMTuner.get_best_booster` instead.
```

```python
optuna/integration/cma.py
>>> optuna.integration.CmaEsSampler()
<stdin>:1: FutureWarning: CmaEsSampler has been deprecated in v2.0.0. This feature will be removed in v4.0.0. See https://github.com/optuna/optuna/releases/tag/v2.0.0. This class is renamed to :class:`~optuna.integration.PyCmaSampler`.
```

```python
optuna/integration/keras.py
>>> optuna.integration.KerasPruningCallback()
<stdin>:1: FutureWarning: KerasPruningCallback has been deprecated in v2.1.0. This feature will be removed in v4.0.0. See https://github.com/optuna/optuna/releases/tag/v2.1.0. Recent Keras release (2.4.0) simply redirects all APIs in the standalone keras package to point to tf.keras. There is now only one Keras: tf.keras. There may be some breaking changes for some workflows by upgrading to keras 2.4.0. Test before upgrading. REF:https://github.com/keras-team/keras/releases/tag/2.4.0
```

```python
optuna/structs.py
>>> optuna.structs.TrialPruned
/home/yanase/pfn/code/optuna-review/optuna/structs.py:18: FutureWarning: `structs` is deprecated. Classes have moved to the following modules. `structs.StudyDirection`->`study.StudyDirection`, `structs.StudySummary`->`study.StudySummary`, `structs.FrozenTrial`->`trial.FrozenTrial`, `structs.TrialState`->`trial.TrialState`, `structs.TrialPruned`->`exceptions.TrialPruned`.
  warnings.warn(_message, FutureWarning)
```
(Note that the other classes in `optuna/structs.py` also show the same `text` in the messages.)

not522:
Thank you for your reviews! I added `_DEPRECATION_WARNING_TEMPLATE` to remove duplication. PTAL.

tests/test_deprecated.py
```diff
@@ -137,14 +138,25 @@ def _func() -> int:
...
     assert decorated_func.__doc__ == expected_func_doc
 
+    with pytest.warns(FutureWarning) as record:
+        decorated_func()
+    assert isinstance(record.list[0].message, Warning)
+    expected_warning_message = _deprecated._DEPRECATION_WARNING_TEMPLATE.format(
+        name=(name if name is not None else "_func"), d_ver="1.1.0", r_ver="3.0.0"
+    )
+    if text is not None:
+        expected_warning_message += " " + text
```
toshihikoyanase, line 148: Hmm, maybe I'm too conservative, but this is a copy of message generation logic in `optuna/_deprecated.py`, and I'm not sure if this can test if the logic actually generates the expected strings.

How about simplifying the expected values using `actual.endswith(expected_keyword)`, `expected_keyword in actual`

not522: If this test is passed, `actual.endswith(expected_keyword)` and `expected_keyword in actual` are also passed regardless of `_deprecated.py`. Do you mean to relax the constraint of the test?

toshihikoyanase: > If this test is passed, actual.endswith(expected_keyword) and expected_keyword in actual are also passed

I agree. But, I thinking we can avoid/reduce the conditional branch during the expected message generation. How about removing the parametrization of `name` since this test is for the `test` argument.

not522: OK, I reverted the last commit.

toshihikoyanase (approved):

tests/test_deprecated.py
toshihikoyanase: Thanks!

Issue #2544, opened 2021-03-30
Is this normal contour plot with categorical value?


fsmosca:
The contour [plot seems weird](https://fsmosca.github.io/Mabigat/example.html). Is this normal?

The variable random_multi_pv in this case is categorical integer. If I change this to continuous, it is fine

Ref.:
optuna==2.6.0
plotly==4.14.3

I tried it with matplotlib, the same thing happened - not nice.
I tried it with optuna==2.5.0 / plotly, not nice too.

I use it in this [repo](https://github.com/fsmosca/Mabigat).

nzw0301:
Could you explain a little bit more about the wiredness of the plot?

fsmosca:
![](https://i.imgur.com/om0S2ur.png)

red marking: the plot does not exactly fit.
yellow: the number is not in sequence.

nzw0301:
Thank you for your helpful clarification! 

Let me put a minimum example.

```python
import optuna
import numpy as np


def objective(trial):
    x = trial.suggest_uniform('x', -100, 100)
    y = trial.suggest_categorical('y', np.arange(-10, 10, 1))
    z = trial.suggest_uniform('z', -100, 100)
    return x ** 2 + y - z

study = optuna.create_study()
study.optimize(objective, n_trials=30)

optuna.visualization.plot_contour(study, params=['x', 'y', 'z'])
```

nzw0301:
About yellow circles: I think a number of visualisation functions, such as `plot_contour` and `plot_parallel_coordinate`, do not sort the suggested values from categorical distribution when drawing a plot.

nzw0301:
Yellow circles: I was wrong. The suggested values from a categorical distribution are actually **sorted**, but they are soreted as `str`.

fsmosca:
Try this for the yellow.
```python
y = trial.suggest_categorical('y', [4, 5, 8, 11])
```

On red the plot fits as expected. I will check again my code why it does not fit well.

nzw0301:
Yes, I think the reason is same: categorical calues are sorted as `str`: 

```python
sorted(map(str, [4, 5, 8, 11]))
# ['11', '4', '5', '8']
```

This is why `10` or `11` appears in the left corner....

nzw0301:
If the suggested values only contain float or int, we can sort the values as a numerical value rather than `str`.
[...]

PR #2569, opened 2021-04-06
Sort categorical values on axis that contains only numerical values in `visualization.plot_contour`

nzw0301:
## Motivation

The contour plot's axis can be improved when a categorical distribution suggests only `int` or `float` values. 

See also a user request as in https://github.com/optuna/optuna/issues/2544.

### Example

```python
import optuna


def objective(trial):
    x = trial.suggest_uniform('x', -100, 100)
    y = trial.suggest_categorical('y', list(range(-10, 10)))
    z = trial.suggest_uniform('z', -100, 100)
    return x ** 2 + y - z

study = optuna.create_study()
study.optimize(objective, n_trials=200)

optuna.visualization.plot_contour(study, params=['y', 'z'])
```

Parameter `y`'s axis is sorted as `str`:

newplot (2)

## Description of the changes

 if suggested values contain only `int` or `float`, then the parameter is not treated as a categorical distribution.

---

### Improved plot

The generated plot on this pull request is as follows.
newplot (4)

## Note

I will send another PR for matplotlib backend.

I think there is room to improve the speed of `visualization.plot_contour`. `_is_categorical` or `_is_numerical` looks expensive if the number of trials is large. We can use a cache to store the evaluated value of `_is_numerical`.

In terms of this point, I think this speed-up should be separated from this pull request because it might be difficult for assignees to review both points.

### DIFF
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -137,8 +137,12 @@ def _get_contour_plot(
             min_value = math.pow(10, math.log10(min_value) - padding)
             max_value = math.pow(10, math.log10(max_value) + padding)
 
-        elif not _is_numerical(trials, p_name):
+        elif _is_numerical(trials, p_name):
+            padding = (max_value - min_value) * padding_ratio
+            min_value = min_value - padding
+            max_value = max_value + padding
 
+        else:
             # Plotly>=4.12.0 draws contours using the indices of categorical variables instead of
             # raw values and the range should be updated based on the cardinality of categorical
             # variables. See https://github.com/optuna/optuna/issues/1967.
@@ -148,10 +152,6 @@ def _get_contour_plot(
                 min_value = -padding
                 max_value = span + padding
 
-        else:
-            padding = (max_value - min_value) * padding_ratio
-            min_value = min_value - padding
-            max_value = max_value + padding
         param_values_range[p_name] = (min_value, max_value)
 
     if len(sorted_params) == 2:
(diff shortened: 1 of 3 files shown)


optuna/visualization/_contour.py
```diff
@@ -137,7 +137,7 @@ def _get_contour_plot(
             min_value = math.pow(10, math.log10(min_value) - padding)
             max_value = math.pow(10, math.log10(max_value) + padding)
 
-        elif _is_categorical(trials, p_name):
+        elif not _is_numerical(trials, p_name):
```
himkt, line 140: [nits]

Can we rewrite here like following? (if it's not, please leave it as is)

```python
        elif _is_numerical(trials, p_name):
            padding = (max_value - min_value) * padding_ratio
            min_value = min_value - padding
            max_value = max_value + padding

        else:
            # Plotly>=4.12.0 draws contours using the indices of categorical variables instead of
            # raw values and the range should be updated based on the cardinality of categorical
            # variables. See https://github.com/optuna/optuna/issues/1967.
            if version.parse(plotly.__version__) >= version.parse("4.12.0"):
                span = len(set(values)) - 1
                padding = span * padding_ratio
                min_value = -padding
                max_value = span + padding
```

himkt (approved):
It looks good, approved with a minor question.

HideakiImamura (approved):
LGTM.

optuna/visualization/_contour.py
nzw0301: Thanks! The modification should be included in this pull request, 1d62f3b.

PR #2592, opened 2021-04-15
[WIP] Serialize study and trials

hvy:
_Highly WIP and experimental. Do not merge._

## Motivation

TODO

## Description of the changes

Allows serializing studies, including trials.

## TODO

- [ ] Design (e.g. determine exactly what to serialize)
- [ ] Implement
- [ ] Document
- [ ] Test

### DIFF
--- a/optuna/serialization.py
+++ b/optuna/serialization.py
@@ -1,20 +1,43 @@
 import pickle
 
 import optuna
+from optuna._experimental import experimental
 
 
+@experimental("2.8.0")
 def save_study(study: optuna.study.Study, path: str) -> None:
-    study_in_memory = optuna.create_study()
-    for trial in study.trials:
-        study_in_memory.add_trial(trial)
+    tmp_study = optuna.create_study()
+
+    _copy_study(study, tmp_study)
 
     with open(path, "wb") as f:
-        pickle.dump(study_in_memory, f)
+        pickle.dump(tmp_study, f)
 
 
+@experimental("2.8.0")
 def load_study(study: optuna.study.Study, path: str) -> None:
     with open(path, "rb") as f:
-        study_in_memory = pickle.load(f)
+        tmp_study = pickle.load(f)
 
-    for trial in study_in_memory.trials:
-        study.add_trial(trial)
+    _copy_study(tmp_study, study)
+
+
+def _copy_study(src: optuna.study.Study, dst: optuna.study.Study) -> None:
+    _copy_study_system_attrs(src, dst)
+    _copy_study_user_attrs(src, dst)
+    _copy_study_trials(src, dst)
+
+
+def _copy_study_system_attrs(src: optuna.study.Study, dst: optuna.study.Study) -> None:
+    for key, value in src.system_attrs.items():
+        dst.set_system_attr(key, value)
+
+
+def _copy_study_user_attrs(src: optuna.study.Study, dst: optuna.study.Study) -> None:
+    for key, value in src.user_attrs.items():
+        dst.set_user_attr(key, value)
+
+
+def _copy_study_trials(src: optuna.study.Study, dst: optuna.study.Study) -> None:
+    for trial in src.trials:
+        dst.add_trial(trial)
--- a/optuna/serialization.py
+++ b/optuna/serialization.py
@@ -39,5 +39,4 @@ def _copy_study_user_attrs(src: optuna.study.Study, dst: optuna.study.Study) -> 
 
 
 def _copy_study_trials(src: optuna.study.Study, dst: optuna.study.Study) -> None:
-    for trial in src.trials:
-        dst.add_trial(trial)
+    dst.add_trials(src.trials)
--- a/optuna/serialization.py
+++ b/optuna/serialization.py
@@ -1,0 +1,20 @@
+import pickle
+
+import optuna
+
+
+def save_study(study: optuna.study.Study, path: str) -> None:
+    study_in_memory = optuna.create_study()
+    for trial in study.trials:
+        study_in_memory.add_trial(trial)
+
+    with open(path, "wb") as f:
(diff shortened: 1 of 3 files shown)


optuna/serialization.py
```diff
@@ -0,0 +1,20 @@
+import pickle
+
+import optuna
+
+
+def save_study(study: optuna.study.Study, path: str) -> None:
+    study_in_memory = optuna.create_study()
```
sile, line 7: I'm not confirmed, but maybe.
```suggestion
    study_in_memory = optuna.create_study(directions=study.directions)
```

optuna/serialization.py
```diff
@@ -0,0 +1,20 @@
+import pickle
+
+import optuna
+
+
+def save_study(study: optuna.study.Study, path: str) -> None:
+    study_in_memory = optuna.create_study(directions=study.directions)
+    for trial in study.trials:
```
sile, line 8: How about copying `study.user_attrs` and `study.system_attrs` too?

optuna/serialization.py
```diff
@@ -0,0 +1,20 @@
...
+        study_in_memory.add_trial(trial)
+
+    with open(path, "wb") as f:
+        pickle.dump(study_in_memory, f)
+
+
+def load_study(study: optuna.study.Study, path: str) -> None:
```
sile, line 15: [RFC] An alternative interface could be as follows (I'm not sure which one is better though):
```python
def load_study(path: str, *create_study_args, **create_study_kwargs) -> optuna.study.Study:
    with open(path, "rb") as f:
        study_in_memory = pickle.load(f)
        
    study = optuna.create_study(*create_study_args, directions=study_in_memory.directions, **create_study_kwargs)
    ...
```

sile: Another topic is that the name `load_study` sounds confusing as there's already a method that has the same name (`optuna.study.load_study`).

hvy: Thanks for the input. I was thinking about something like your alternative as well (even placing it next to `optuna.create_study`, `optuna.load_study`, an additional for instance `optuna.import_study` to somewhat align parameters to the former two, but with an additional path argument). I'm wondering though if it's not cleaner trying to symmetrize the interfaces between save and load and keep it separated from the study construction functions. I was imagining raising errors in case there are incompatibilities such as the number of objectives or mismatch in directions. What do you think?

As for the conflicting name of `load_study`, I agree that it can be improved. We could simply rename it to `import_study` (along with `export_study`). We could also make those functions `Study` methods `Study.{save,load}`.

sile: Thank you for your response.

>  to symmetrize the interfaces between save and load and keep it separated from the study construction functions.  I was imagining raising errors in case there are incompatibilities such as the number of objectives or mismatch in directions.

Sounds good!

> We could also make those functions Study methods `Study.{save,load}`.

Personally, I like this option (but want to hear opinions from other developers).

optuna/serialization.py
hvy: Makes sense to include in the serialization. Thanks, added!

optuna/serialization.py
hvy: Hm, I'm also beginning to think that the user loading the study would want to do that without prior knowledge about the study. More inputs from others would be fruitful.

sile: >  without prior knowledge about the study. 

It would be great if possible.
(I think that an appropriate validation (and error message) is also a practical solution though)

sile: Just an idea, but let me share another possible design.
```python
class Study:
  def save_to_sqlite(self, path):
      study = optuna.create_study(study_name=self.study_name, storage=f"sqlite:///{path}", directions=self.directions)
      ... copy trials, user-attrs, and system-attrs ...
```

To load the saved study, users can use the existing `optuna.study.load_study` function.

sile: Or,
```python
def copy_study(study_name, from_storage, to_storage):
  from_study = optuna.load_study(study_name=study_name, storage=from_storage)
  to_study = optuna.create_study(study_name=study_name, storage=to_storage directions=from_storage.directions)
      ... copy trials, user-attrs, and system-attrs ...
```

optuna/serialization.py
sile: Thanks!

optuna/serialization.py
hvy: Hm, I see. I was also imagining use cases where the storage information cannot be shared between the ones saving and loading. I'm not sure it's possible with this last design.

sile: If you specify an SQLite file path to `to_storage`, I think, you could handle the file in the same way with the file generated by `optuna.save_study(..)` in this PR.
(as a difference, users would have to know the target study name in addition to the file path)

hvy: Right, thanks for the explanation. I was imagining always directly writing to the storage of the receiving user (in the case of e.g. MySQL).

hvy: Leaving some conclusions from a discussion we had outside.

- Use SQLite as format. It will be less intrusive to the current API since we can reuse `load_study`. Additionally, behavior will be well defined in contrast to pickle
- Allow resolving conflicting study names at dump via argument. Could be a follow up

Possible follow-ups

- Functionality to convert between storages. E.g. via `copy_study` or arguments to `load_study`. It'll be useful for restoring studies and converting them to in-memory in favor of performance
- Allow omitting study name with `load_study` if there is only a single study in the storage

hvy: I'll update the PR according to the items above.

hvy:
Let me close this in favor of https://github.com/optuna/optuna/pull/2607. Introducing `copy_study` in favor of using pickle seems like a better defined and more robust solution.

PR #2597, opened 2021-04-17
Compile/upload tutorial docs and use it in readthedocs

himkt:
## Motivation
ReadTheDocs has an execution time limit for each build (5min).
Optuna uses sphinx-gallery and it means that some scripts are run on readthedocs, which takes a long time.
(ref. https://sphinx-gallery.github.io/stable/getting_started.html#create-simple-gallery)
Instead, GitHub Actions is relatively forgiving about execution time.
So, this PR changes a way to built a document: use GitHub Actions/Artifacts to running shinx-gallery to avoid running sphinx-gallery on readthedocs.

## Description of the changes
- 69424d9: edit `conf.py` to download artifacts when `GITHUB_TOKEN` is specified
- 9d44e53: add workflow to build document and upload pre-compiled tutorial to GitHub Artifacts

---

I confirmed it worked well in my personal repository:
- https://github.com/himkt/optuna-test-rtds/runs/2369939431
- https://readthedocs.org/projects/himktoptuna/builds/13536913/

### DIFF
--- a/.github/workflows/read-the-docs.yml
+++ b/.github/workflows/read-the-docs.yml
@@ -1,0 +1,41 @@
+name: Build/upload docs and trigger ReadTheDocs building
+
+on:
+  push:
+    branches:
+      - master
+  pull_request: {}
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+
+    - name: Install
+      run: |
+        python -m pip install -U pip
+        pip install --progress-bar off -U .[document]
+    - name: Build
+      run: |
+        cd docs
+        make html
+      env:
+        OMP_NUM_THREADS: 1
+
+    - uses: actions/upload-artifact@v2
+      with:
+        name: artifacts-${{ github.sha }}
+        path: docs/source/tutorial
+
+    - uses: dfm/rtds-action@v1
+      with:
+        webhook_url: ${{ secrets.RTDS_WEBHOOK_URL }}
+        webhook_token: ${{ secrets.RTDS_WEBHOOK_TOKEN }}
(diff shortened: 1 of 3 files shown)


HideakiImamura:
@toshihikoyanase Let me assign you since we've already discussed this PR.

toshihikoyanase:
> ReadTheDocs has an execution time limit for each build (5min).

A nit-picky, but the [RTD reference](https://docs.readthedocs.io/en/stable/builds.html) says that build time limitation is 15 minutes, not 5minites.

.github/workflows/read-the-docs.yml
```diff
@@ -0,0 +1,41 @@
...
+        OMP_NUM_THREADS: 1
+
+    - uses: actions/upload-artifact@v2
+      with:
+        name: artifacts-${{ github.sha }}
+        path: docs/source/tutorial
+
+    - uses: dfm/rtds-action@v1
+      with:
+        webhook_url: "aa"
```
toshihikoyanase, line 38: Is this value correct? Or, do you put a place-holder?

toshihikoyanase (commented):
Thank you for your PR. It basically looks great, but I have some questions.

1. How do we rebuild the RTD docs? Do we just re-run this GitHub Actions workflow?
2. If the artifact files are removed, can we rebuild the RTD docs?
3. If we can kick the RTD build only via GitHub Actions, we may want to [trigger the workflow manually](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow). What do you think of it?

himkt:
I close this PR, since we find this approach to be difficult to build a document when releasing a new version of Optuna. To launch a readthedocs job on GitHub release, readthedocs requires GitHub incoming webhook is successfully added. With GitHub incoming webhook, we can add an automation rule to activate a new release (tag) on readthedocs. However, we cannot find a way to launch a readthedocs job from GitHub Actions.

PR #2576, opened 2021-04-09
Improve `param_importances`

nzw0301:
## Motivation

The current `param_importances` add different colors depending on the distribution class that users may not access directly. In addition, single color among distributions is enough to visualize important hyperparameters when we think about the role of this plot -- the distribution does not matter.

See also https://github.com/optuna/optuna/pull/1991#issuecomment-816726700.

## Description of the changes

- use a single color map for both Plotly and Matplolib backends.
- Remove legend from Matplolib backend.

---

```python
import optuna


def objective(trial):
    x = trial.suggest_int("x", 0, 2)
    y = trial.suggest_float("y", -1.0, 1.0)
    z = trial.suggest_float("z", 0.0, 1.5)
    return x ** 2 + y ** 3 - z ** 4


sampler = optuna.samplers.RandomSampler(seed=10)
study = optuna.create_study(sampler=sampler)
study.optimize(objective, n_trials=10)

optuna.visualization.matplotlib.plot_param_importances(study)
```

## Plotly

newplot (1)

## Matplotlib

download

PR #2604, opened 2021-04-19
Remove `_get_distribution` from `visualization/matplotlib/_param_importances.py`

nzw0301:
## Motivation

The internal method, `_get_distribution`, has never been used in optuna since #2576.

## Description of the changes

Remove `_get_distribution`.


hvy (approved):
Good catch, LGTM!

PR #2603, opened 2021-04-19
Show legend of `optuna.visualization.matplotlib.plot_edf`

nzw0301:
## Motivation

As in [`optuna.visualization.matplotlib.plot_edf`](https://optuna.readthedocs.io/en/stable/reference/visualization/generated/optuna.visualization.matplotlib.plot_edf.html#optuna.visualization.matplotlib.plot_edf), we cannot know which `study` corresponds to each line plot since there is not `legend` on the generated plot.

## Description of the changes

Call `ax.legend()` explicitly.

If we have too many studies, users can remove `legend` by calling `ax.get_legend().remove()`, for example,

```python
ax = optuna.visualization.matplotlib.plot_edf([study0, study1, study2])
ax.get_legend().remove()
```

---

# Examples

## Reproducible code snippet

```python
import math

import optuna


def ackley(x, y):
    a = 20 * math.exp(-0.2 * math.sqrt(0.5 * (x ** 2 + y ** 2)))
    b = math.exp(0.5 * (math.cos(2 * math.pi * x) + math.cos(2 * math.pi * y)))
    return -a - b + math.e + 20


def objective(trial, low, high):
    x = trial.suggest_float("x", low, high)
    y = trial.suggest_float("y", low, high)
    return ackley(x, y)


sampler = optuna.samplers.RandomSampler(seed=10)

# Widest search space.
study0 = optuna.create_study(study_name="x=[0,5), y=[0,5)", sampler=sampler)
study0.optimize(lambda t: objective(t, 0, 5), n_trials=500)

# Narrower search space.
study1 = optuna.create_study(study_name="x=[0,4), y=[0,4)", sampler=sampler)
study1.optimize(lambda t: objective(t, 0, 4), n_trials=500)

# Narrowest search space but it doesn't include the global optimum point.
study2 = optuna.create_study(study_name="x=[1,3), y=[1,3)", sampler=sampler)
study2.optimize(lambda t: objective(t, 1, 3), n_trials=500)
```

## Master 
download (7)

## This PR
download (8)

PR #2606, opened 2021-04-19
Show legend of `optuna.visualization.matplotlib.plot_intermediate_values`

nzw0301:
## Motivation

As in [`optuna.visualization.matplotlib.plot_intermediate_values`](https://optuna.readthedocs.io/en/stable/reference/visualization/generated/optuna.visualization.matplotlib.plot_intermediate_values.html), we cannot know which trial corresponds to each line plot since there is no legend on the generated plot.

## Description of the changes

Call `ax.legend()` explicitly.

If we have too long legend, users can remove `legend` by calling `ax.get_legend().remove()`, for example,

```python
ax = optuna.visualization.matplotlib.plot_intermediate_values(study)
ax.get_legend().remove()
```

---

# Examples

## Reproducible code snippet

```python
import optuna


def f(x):
    return (x - 2) ** 2


def df(x):
    return 2 * x - 4


def objective(trial):
    lr = trial.suggest_float("lr", 1e-5, 1e-1, log=True)

    x = 3
    for step in range(128):
        y = f(x)

        trial.report(y, step=step)
        if trial.should_prune():
            raise optuna.TrialPruned()

        gy = df(x)
        x -= gy * lr

    return y


study.optimize(objective, n_trials=16)

optuna.visualization.matplotlib.plot_intermediate_values(study)
```

## Master 

download (11)

## This PR
download (10)

### DIFF
--- a/optuna/visualization/matplotlib/_intermediate_values.py
+++ b/optuna/visualization/matplotlib/_intermediate_values.py
@@ -81,7 +81,7 @@ def _get_intermediate_plot(study: Study) -> "Axes":
 
     # Set up the graph style.
     plt.style.use("ggplot")  # Use ggplot style sheet for similar outputs to plotly.
-    _, ax = plt.subplots()
+    _, ax = plt.subplots(tight_layout=True)
     ax.set_title("Intermediate Values Plot")
     ax.set_xlabel("Step")
     ax.set_ylabel("Intermediate Value")
--- a/optuna/visualization/matplotlib/_intermediate_values.py
+++ b/optuna/visualization/matplotlib/_intermediate_values.py
@@ -110,4 +110,7 @@ def _get_intermediate_plot(study: Study) -> "Axes":
         )
         return ax
 
+    if len(trials) >= 2:
+        ax.legend(bbox_to_anchor=(1.05, 1), loc="upper left", borderaxespad=0.0)
+
     return ax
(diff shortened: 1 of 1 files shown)


toshihikoyanase:
@himkt Could you review this PR as well? The change is similar to #2603.

himkt (approved):
LGTM

toshihikoyanase (commented):
Thank you for your PR. The change looks good to me, but I found that [the legend of the plot](https://67029-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/visualization/generated/optuna.visualization.matplotlib.plot_intermediate_values.html#optuna.visualization.matplotlib.plot_intermediate_values) in the reference was trimmed as follows:

image

Do we need to update the code example?

nzw0301:
@toshihikoyanase Good catch! thank you for pointing out it. I think so. 

In my experience, a common approach to fix it is `matplotlib.pyplot.tight_layout()`, but it does not work for our case. So, I added `tight_layout=True` to `plt.subplots` as in [this update](https://github.com/optuna/optuna/pull/2606/files#diff-1c45381f49d4dd81585f5a5a9f8308d17572ed8bbf7619f1899d2d2e89a75107R84), as a result, the plot looks fine:
https://67062-122299416-gh.circle-artifacts.com/0/docs/build/html/reference/visualization/generated/optuna.visualization.matplotlib.plot_intermediate_values.html#optuna.visualization.matplotlib.plot_intermediate_values

toshihikoyanase (approved):
Looks great! Thank you for your update.

PR #2607, opened 2021-04-20
Copy study

hvy:
## Motivation

See description below.

## Description of the changes

Alternative to https://github.com/optuna/optuna/pull/2592. See https://github.com/optuna/optuna/pull/2592#discussion_r613830437.

## Note

I first implemented a rollback but came to the conclusion that it's better handled by the called and reverted it in https://github.com/optuna/optuna/pull/2607/commits/39ccf298f61147d2ba072bf7e9d3e01fe3057ec0. It's a trivial operation (`delete_study`) since we never append to an existing study, and user might want to look into intermediate results in case of failures.

Issue #1128, opened 2020-04-16
Align the order of arguments of `optuna.study.create_study` and `optuna.study.load_study`.


keisuke-umezawa:
## Motivation

Currently, `optuna.study.create_study` and `optuna.study.load_study` have different orders of arguments, although their roles are similar. I want to align the order of arguments.

- [create_study](https://github.com/optuna/optuna/blob/master/optuna/study.py#L761-L768)
- [load_study](https://github.com/optuna/optuna/blob/master/optuna/study.py#L844-L849)

## Description

I want to align the order of arguments of `optuna.study.create_study` and `optuna.study.load_study`. But we should discuss following points:

1. Which order will we choose?
2. As much as possible we should avoid disturbing users because of changing orders.

hvy:
Thanks for bringing this up, it's definitely something that should be uprooted... The problem is that there doesn't seem to be a straight forward way to go through a deprecation process. Additionally, as both functions are likely already used by the majority of the users (at least `create_study`), it's better not to make any rushed changes before hearing the opinions from more users. Let me close this issue for now but allow reopening "as needed".

PR #2608, opened 2021-04-20
[RFC] Allow `None` as `study_name` when there is only a single study in `load_study`

hvy:
## Motivation

When loading an existing study from a storage, the name of the study is not necessarily known (see `copy_study` in https://github.com/optuna/optuna/pull/2607). E.g. user A dumps a study to a temporary storage and user B is given only access to that storage.

## Description of the changes

Allows passing `None` as `study_name` to `optuna.load_study` if there is only a single study stored in the storage.

## Notes

- Currently not aligned with `optuna.delete_study`
- Argument cannot be omitted due to the ordering (see also https://github.com/optuna/optuna/issues/1128)

### DIFF
--- a/optuna/study.py
+++ b/optuna/study.py
@@ -1160,7 +1160,7 @@ def create_study(
 
 
 def load_study(
-    study_name: str,
+    study_name: Optional[str],
     storage: Union[str, storages.BaseStorage],
     sampler: Optional["samplers.BaseSampler"] = None,
     pruner: Optional[pruners.BasePruner] = None,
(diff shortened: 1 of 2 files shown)


toshihikoyanase:
@nzw0301 Could you review this PR, please?

nzw0301 (approved):
Thank you for giving the context to review PR.  LGTM!

toshihikoyanase (approved):
Thank you for your PR. The change does not affect existing code and this feature is actually useful  when using `optuna.copy_study` as you mentioned.
LGTM!

PR #2609, opened 2021-04-20
Fix grammatical error in failure message

agarwalrounak:
## Motivation

I got the below error when I returned a string instead of a numeric value from an objective function:
` WARNING - optuna._optimize - Trial 1 failed, because the number of the values 7 is did not match the number of the objectives 1`
There is a slight grammatical error in the failure message above, i.e., the `is` word, which makes the sentence construction wrong.

## Description of the changes

Removed the `is` word from failure message.


nzw0301 (approved):
Thank you for sending the pull request to fix it!

PR #2612, opened 2021-04-21
Add MaxTrialsCallback class to enable stopping after fixed number of trials

Crissman:
## Motivation

Users may want to stop trials after a certain number of complete or incomplete trials.

## Description of the changes

Create an `optuna.study` class for call backs, which stops after the user set `n_trials` exceeds the number of trials according to the user set `TrialState` list. Default is `TrialState.COMPLETE`.

Todo:

- [x] Documentation
- [x] Test for failed
- [x] Create testing

### DIFF
--- a/examples/max_trials_callback.py
+++ b/examples/max_trials_callback.py
@@ -12,19 +12,8 @@ regardless of the number of workers/scripts running the Trials.
 from time import sleep
 
 import optuna
+from optuna.study import MaxTrialsCallback
 from optuna.trial import TrialState
-
-
-num_completed_trials = 10
-
-
-def max_trial_callback(study, trial):
-    # we consider all the running states and already completed states.
-    n_complete = len(
-        study.get_trials(deepcopy=False, states=[TrialState.COMPLETE, TrialState.RUNNING])
-    )
-    if n_complete >= num_completed_trials:
-        study.stop()
 
 
 def objective(trial):
@@ -40,7 +29,9 @@ if __name__ == "__main__":
         load_if_exists=True,
     )
 
-    study.optimize(objective, n_trials=50, callbacks=[max_trial_callback])
+    study.optimize(
+        objective, n_trials=50, callbacks=[MaxTrialsCallback(10, states=(TrialState.COMPLETE,))]
+    )
     trials = study.trials_dataframe()
     print("Number of completed trials: {}".format(len(trials[trials.state == "COMPLETE"])))
 
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -1,0 +1,59 @@
+from typing import Tuple
+
+import optuna
+from optuna._experimental import experimental
+from optuna.trial import FrozenTrial
+from optuna.trial import TrialState
+
+
+@experimental("2.8.0")
+class MaxTrialsCallback:
+    """Set a maximum number of trials before ending the study.
+
+    While the :obj:`n_trials` argument of :obj:`optuna.optimize` sets the number of trials that
+    will be run, you may want to continue running until you have a certain number of successfullly
+    completed trials or stop the study when you have a certain number of trials that fail.
+    This :obj:`MaxTrialsCallback` class allows you to set a maximum number of trials for a
+    particular :class:`~optuna.trial.TrialState` before stopping the study.
+
+    Example:
+
+        .. testcode::
+
+            import optuna
+            from optuna.study import MaxTrialsCallback
+            from optuna.trial import TrialState
+
+
+            def objective(trial):
+                x = trial.suggest_float("x", -1, 1)
+                return x ** 2
+
+
+            study = optuna.create_study()
+            study.optimize(
+                objective,
+                callbacks=[MaxTrialsCallback(10, states=(TrialState.COMPLETE,))],
+            )
+
+    Args:
+        n_trials:
+            The max number of trials. Must be set to an integer.
+        states:
+            Tuple of the :class:`~optuna.trial.TrialState` to be counted
+            towards the max trials limit. Default value is :obj:`(TrialState.COMPLETE,)`.
+    """
+
+    def __init__(
+        self, n_trials: int, states: Tuple[TrialState, ...] = (TrialState.COMPLETE,)
+    ) -> None:
+        self._n_trials = n_trials
+        self._states = states
+
+    def __call__(
+        self, study: "optuna.study.Study", trial: FrozenTrial
+    ) -> None:
+        trials = study.get_trials(deepcopy=False, states=self._states)
+        n_complete = len(trials)
+        if n_complete >= self._n_trials:
+            study.stop()
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -50,9 +50,7 @@ class MaxTrialsCallback:
         self._n_trials = n_trials
         self._states = states
 
-    def __call__(
-        self, study: "optuna.study.Study", trial: FrozenTrial
-    ) -> None:
+    def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
         trials = study.get_trials(deepcopy=False, states=self._states)
         n_complete = len(trials)
         if n_complete >= self._n_trials:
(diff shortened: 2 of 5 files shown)


hvy:
Thanks for the PR! This is a reoccurring pattern and something that'll definitely be useful. I'm wondering if we shouldn't mark it experimental however. I understand that there are two goals, one to provide a utility to stop a study, another to  introduce a new concept of "build-in callbacks". If the main goal is the former, introducing and adapting to a new parameter to `Study.optimize` is another if not a better option. E.g. `stop_if: Callable[[Study], bool]`, that can be invoked before the trial creation. This is probably beyond the scope of this PR but could potentially "conflict". If the latter, I think there are alternatives such as introducing a submodule such as `optuna.study.callbacks` (and it'd be great if we could list up actual other callbacks that might be relevant).

Crissman:
Thanks for the comment. I don't know which goal will be more important, so for now, I've marked it experimental, as suggested. How about we merge and wait for a while, and see which becomes more used?

hvy (commented):
Thanks for the test, I left one minor suggestion.

I think we can do that. We could consider introducing a new (experimental) submodule (i.e. `optuna.callbacks`) if/when we are to introduce more built in callback.

optuna/study.py
```diff
@@ -1411,3 +1411,55 @@ def objective(trial):
 
     storage = storages.get_storage(storage)
     return storage.get_all_study_summaries()
+
+
+@experimental("2.8.0")
+class MaxTrialsCallback:
```
hvy, line 1417: Note from "offline" discussion: Implement this class in a separate file (e.g. `optuna/_callbacks.py` since `optuna/study.py` is getting big.

examples/max_trials_callback.py
```diff
@@ -12,21 +12,10 @@
 from time import sleep
 
 import optuna
+from optuna._callbacks import MaxTrialsCallback
```
hvy, line 15: We could import `MaxTrialsCallback` at the top of `optuna/__init__.py` or `optuna/study.py` to make `MaxTrialsCallback` public.

hvy (changes requested):
Thanks for the migration to a separate file. That might be better since `study.py` is already huge. I noticed on the other hand that the callback is now hidden behind private API. Could you take a look at my comment?

Crissman:
@hvy PTAL

examples/max_trials_callback.py
hvy: Let's add the callback to the documentation (.rst) too. For instance https://github.com/optuna/optuna/blob/master/docs/source/reference/optuna.rst, if it's defined in `optuna/__init__.py`.

Crissman: `import optuna` didn't work, cuz then it couldn't find `optuna.study`. Stringed type hints worked. Didn't know you could do that!

Crissman:
Comments reflected, @hvy  PTAL

optuna/_callbacks.py
```diff
@@ -0,0 +1,57 @@
+from typing import Tuple
+
+from optuna._experimental import experimental
+from optuna._study_summary import StudySummary  # NOQA
```
hvy, line 4: I think this line is unnecessary.
```suggestion
```

Crissman: Good catch. The `NOQA` stopped it from being caught by the checkers. 😅

optuna/_callbacks.py
```diff
@@ -0,0 +1,56 @@
...
+            towards the max trials limit. Default value is :obj:`(TrialState.COMPLETE,)`.
+    """
+
+    def __init__(
+        self, n_trials: int, states: Tuple[TrialState, ...] = (TrialState.COMPLETE,)
+    ) -> None:
+        self._n_trials = n_trials
+        self._states = states
+
+    def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:  # type: ignore # noqa
```
hvy, line 52: More of a question but how come we need to ignore the type/style checks? mypy is for instance skipped now. I'm wondering if you cannot `import optuna` without circular imports, and thereby "fix" the type hint?

Crissman: My local environment has results different from CI 🤷‍♂️ , so will test again with CI.

Crissman: `# NOQA` required for flake8: https://github.com/optuna/optuna/pull/2612/checks?check_run_id=2572273720

Crissman: Confirmed `type: ignore` required for mypy test: https://github.com/optuna/optuna/pull/2612/checks?check_run_id=2572273720

Crissman: Error when using `import optuna` instead of `study: "optuna.study.Study"`:

```
➜  Optuna git:(max_callbacks) ✗ python examples/max_trials_callback.py 
  File "examples/max_trials_callback.py", line 14, in <module>
    import optuna
  File "/Users/Crissman/GDrive/Github/optuna/optuna/__init__.py", line 8, in <module>
    from optuna import importance
  File "/Users/Crissman/GDrive/Github/optuna/optuna/importance/__init__.py", line 6, in <module>
    from optuna.importance._base import BaseImportanceEvaluator
  File "/Users/Crissman/GDrive/Github/optuna/optuna/importance/_base.py", line 9, in <module>
    from optuna.samplers import intersection_search_space
  File "/Users/Crissman/GDrive/Github/optuna/optuna/samplers/__init__.py", line 1, in <module>
    from optuna.samplers._base import BaseSampler
  File "/Users/Crissman/GDrive/Github/optuna/optuna/samplers/_base.py", line 8, in <module>
  File "/Users/Crissman/GDrive/Github/optuna/optuna/study.py", line 21, in <module>
    from optuna._callbacks import MaxTrialsCallback  # NOQA
  File "/Users/Crissman/GDrive/Github/optuna/optuna/_callbacks.py", line 9, in <module>
    @experimental("2.8.0")
  File "/Users/Crissman/GDrive/Github/optuna/optuna/_callbacks.py", line 53, in MaxTrialsCallback
    def __call__(self, study: optuna.study.Study, trial: FrozenTrial) -> None:
AttributeError: module 'optuna' has no attribute 'study'
```

hvy: This seems to work https://github.com/hvy/optuna/commit/546cbe7d00bef14216b87430b40edf19a983f09f.

Crissman: Aha. Not *instead of* but *and*. Ok, did that. 👍

hvy: Sorry, should've been more clear. Thanks!

optuna/study.py
```diff
@@ -18,6 +18,7 @@
 from optuna import samplers
 from optuna import storages
 from optuna import trial as trial_module
+from optuna._callbacks import MaxTrialsCallback
```
hvy, line 21: To fix warning for unused declaration.
```suggestion
from optuna._callbacks import MaxTrialsCallback  # NOQA
```

Crissman: Thanks! That was what that was for... 😅

optuna/_callbacks.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(
+        self, n_trials: int, states: Tuple[TrialState, ...] = (TrialState.COMPLETE,)
+    ) -> None:
+        self._n_trials = n_trials
+        self._states = states
+
+    def __call__(
+        self, study: "optuna.study.Study", trial: FrozenTrial
+    ) -> None:  # NOQA # type: ignore
```
hvy, line 55: Again, we shouldn't need the exceptions here. https://github.com/optuna/optuna/pull/2612#discussion_r631557813

Crissman: Right. 😢

hvy: Thanks ❤️

hvy:
@not522 do you also want to take a look at this PR? If you're busy, we could reassign it.

hvy (approved):
LGTM!

Thanks for the long running effort. I think that the namespace is still controversial but this'll surely be useful for many and it sounds good to start providing this as an initial experimental feature under `optuna.study`.

not522 (approved):
Thank you. LGTM.

Issue #2620, opened 2021-04-23
`LightGBMTuner` may not be reproducible even if the `optuna_seed` argument is specified.


toshihikoyanase:
## Expected behavior

The `optuna_seed` argument was introduced to LightGBM tuners by #2431, but the unit tests about LightGBM tuner's reproducibility sometimes fail.

In the test, two lightgbm runs are executed with the same seed values, and they should return the same objective values.

## Environment

I saw this issue in the [CI jobs on macOS](https://github.com/optuna/optuna/runs/2408648740#step:7:462) and local windows machine.

Environment of window machine:
- Optuna version: the current master (commit f696b7d4d4afa680160d32ecc48db51d1a7743c1)
- Python version: 3.8.5
- OS: Ubuntu 20.04.2 LTS on Windows 10 WSL2
- (Optional) Other libraries and their versions: lightgbm==3.2.1

## Error messages, stack traces, or logs

The tests failed in the following CI job:
https://github.com/optuna/optuna/runs/2408648740#step:7:462

And I saw similar errors on the local windows machine.

```console
____________ TestLightGBMTuner.test_tune_best_score_reproducibility ____________

self = <tests.integration_tests.lightgbm_tuner_tests.test_optimize.TestLightGBMTuner object at 0x184611610>

...

>       assert best_score_second_try == best_score_first_try
E       assert 4.434944906944917 == 4.542293905553135

tests/integration_tests/lightgbm_tuner_tests/test_optimize.py:768: AssertionError

...

=========================== short test summary info ============================
FAILED tests/integration_tests/lightgbm_tuner_tests/test_optimize.py::TestLightGBMTuner::test_tune_best_score_reproducibility
===== 1 failed, 269 passed, 69 skipped, 142 warnings in 388.59s (0:06:28) ======
```

## Steps to reproduce

1. pip install `pytest-repeat`
2. Run the unit test several times

```console
$ pytest --count 10 tests/integration_tests/lightgbm_tuner_tests/test_optimize.py::TestLightGBMTuner::test_tune_best_score_reproducibility
$ pytest --count 10 tests/integration_tests/lightgbm_tuner_tests/test_optimize.py::TestLightGBMTunerCV::test_tune_best_score_reproducibility
```

## Possible Cause

[...]

PR #2621, opened 2021-04-23
Round objective values of `LightGBMTuner` and `LightGBMTunerCV` for reproducibility.

toshihikoyanase:
## Motivation

This PR is related to #2620. LightGBM sometimes returns slightly different scores, and `LightGBMTuner` loses the reproducibility due to the differences of objective value history.

## Description of the changes

This PR applies `round` to ignore such differences and keep the original scores in the system attrs.

## Opinion wanted

I'm not sure if this workaround works when the score is quite small like `1e-10`, but I believe that the usual scores are as large as 1e-2.


toshihikoyanase:
@tetsuoh0103 @smly @Y-oHr-N Please let me know is you have any comments.

toshihikoyanase:
I talked with the other committers and found that this approach is too ad-hoc.
I'd like to close this issue, and update the test cases to ensure that the samplers receive seed values.

PR #2423, opened 2021-02-28
Add QMC sampler

kstoneriv3:
## Motivation

As in #1797, Quasi-Monte Carlo (QMC) sampler should be supported as a good alternative to `RandomSampler`. I want to discuss the implementational details of the `QMCSampler` as there exist several design choices for ensuring that it works in a distributed environment.

## Description of the changes
 
`QMCSampler` is added at `optuna/sampler/_qmc.py`. Since `scipy` will introduce support of QMC in the 1.7.0 release, we use their implementation of QMC to generate several kinds of QMC sequences.

## Design Choices around Distributed Environment

To suggest QMC samples using distributed workers, we have to synchronize them. This is because QMC sequences are strictly ordered. They cannot be sampled independently; each worker must know exactly how many QMC samples were suggested so far. Since the storage of Optuna does not support atomic transactions, (as far as I understood) we have to compromise somewhere and thus there are some design choices. The possible design I considered are as follows:
- (Current PR) Use the system attributes of the study to count how many QMC samples have been suggested so far. The advantage of this implementation is that it is simple.
- In the system attributes of each trial, store the which sample of QMC sample (`qmc_id`) it is. Then, every time we suggest a new sample, we check all past trials to see how many QMC points have been suggested so far. (Above strategy should be faster than this because they access O(1) information while this access O(n_trials) information in each trial.
- Set `Trial._trial_id` or `Trial.number` equal to `qmc_id + const`. (This fails when there are different types of samplers running simultaneously.)
- Set the maximum number of samples of QMC and sample all points at the construction of `QMCSampler`. In each trial, we randomly pick a point that is not yet sampled from a pool of the pre-sampled points. This strategy is the same as the `GridSampler`. (This can be inefficient if we fail to sample early points of QMC sequences as the earlier points of the QMC sequences are more important than their later parts. So, it seems better to sample QMC in the ordered manner. )

## TODO:
- [x] Discuss design choices based on current design
- [x] Add docstring
- [x] Add test code
- [x] Remove unnecessary comments

PR #2625, opened 2021-04-24
A Proposal of `LatinHypercubeSampler`

kstoneriv3:
## Motivation

I would like to discuss whether it makes sense to include Latin hypercube sampling in optuna. Latin hypercube sampling is a stratified sampling method that usually performs better than random sampling in HPO.

## Discussion

I would like to hear opinions on the following points.
*  Whether it makes sense to add Latin hypercube sampling in optuna. 
Maybe `QMCSampler` is a good enough alternative. One advantage of LatinHypercube over QMC is that the Latin hypercube is more performant when the number of parameters is large. QMCs are known to degrade their performance when the dimension of the sample space is large. Theoretically, QMC excels in situations where there are several hyperparameters that all influence the objective value, which Latin hypercube is suited for situations where there are a lot of hyperparameters but only a few of them influence the objective value.
* (If we include it, ) whether it should be implement it as a new `qmc_type` of `QMCSampler` or introduce an `LHSampler`.
As Latin hypercube requires that the total sample size is pre-determined, I personally think its interface should be separated from `QMCSampler`. 
* ~~The algorithm requires that the total sample size is pre-determined. However, there is an online algorithm that does not require the information. It needs to be discussed whether we should use an online algorithm that does not require the total sample size to be pre-determined at the construction of sampler. (The algorithm can produce 2^k samples by adaptively doubling the total sample size.)~~ *(Edited)* Some of the advanced Latin hypercube algorithms supported in scipy cannot be samples this way. So I assume this is not a good sampling strategy.

## Benchmark
*(Edited)* The benchmark I took was inappropriate because it was comparing Latin hypercube with median pruner against random sampler without any pruner... . I will add new benchmark results later.

### DIFF
--- a/optuna/samplers/__init__.py
+++ b/optuna/samplers/__init__.py
@@ -1,6 +1,7 @@
 from optuna.samplers._base import BaseSampler
 from optuna.samplers._cmaes import CmaEsSampler
 from optuna.samplers._grid import GridSampler
+from optuna.samplers._lhs import LatinHypercubeSampler
 from optuna.samplers._nsga2 import NSGAIISampler
 from optuna.samplers._partial_fixed import PartialFixedSampler
 from optuna.samplers._random import RandomSampler
@@ -15,6 +16,7 @@ __all__ = [
     "CmaEsSampler",
     "GridSampler",
     "IntersectionSearchSpace",
+    "LatinHypercubeSampler",
     "MOTPESampler",
     "NSGAIISampler",
     "PartialFixedSampler",
--- a/optuna/samplers/_lhs.py
+++ b/optuna/samplers/_lhs.py
@@ -159,7 +159,6 @@ class LatinHypercubeSampler(BaseSampler):
         # also make sure that it is OrderedDict
         self._initial_search_space = search_space
         self._warn_incomplete_reseeding = warn_incomplete_reseeding
-        self._should_stop = False
 
     def reseed_rng(self) -> None:
 
@@ -244,9 +243,6 @@ class LatinHypercubeSampler(BaseSampler):
     ) -> None:
 
         self._independent_sampler.after_trial(study, trial, state, values)
-        if self._should_stop:  # TODO(kstoneriv3): Consider using information on `state`
-            study.stop()
-
 
     def _sample_lhs(
         self, study: Study, search_space: Dict[str, BaseDistribution]
@@ -261,7 +257,7 @@ class LatinHypercubeSampler(BaseSampler):
         # TODO(kstoneriv3): Consider using information on `state`
         # TODO(kstoneriv3): Not sure if it properly stops in distributed environment.
         if sample_id == self._n - 1:
-            self._should_stop = True
+            study.stop()
 
         # Use cached `lhs` (Latin hypercube samples) or construct a new one.
         if not self._is_lhs_cached(d, sample_id):
--- a/optuna/samplers/_lhs.py
+++ b/optuna/samplers/_lhs.py
@@ -1,4 +1,3 @@
-from collections import OrderedDict
 from typing import Any
 from typing import Dict
 from typing import Optional
@@ -21,7 +20,6 @@ from optuna.trial import TrialState
 _logger = logging.get_logger(__name__)
 
 _SUGGESTED_STATES = (TrialState.COMPLETE, TrialState.PRUNED)
-
 
 
 @experimental("2.x.0")  # TODO(kstoneriv3)
@@ -155,7 +153,7 @@ class LatinHypercubeSampler(BaseSampler):
         self._seed = seed or numpy.random.MT19937().random_raw()
         self._add_noise = add_noise
         self._independent_sampler = independent_sampler or optuna.samplers.RandomSampler(seed=seed)
-        self._cached_lhs = None
+        self._cached_lhs: Optional[numpy.ndarray] = None
         # TODO(kstoneriv3): make sure that search_space is either None or valid search space.
         # also make sure that it is OrderedDict
         self._initial_search_space = search_space
@@ -215,10 +213,6 @@ class LatinHypercubeSampler(BaseSampler):
         param_distribution: BaseDistribution,
     ) -> Any:
 
-        if self._initial_search_space is not None:
-            if self._warn_independent_sampling:
-                self._log_independent_sampling(trial, param_name)
-
         return self._independent_sampler.sample_independent(
             study, trial, param_name, param_distribution
         )
@@ -230,9 +224,9 @@ class LatinHypercubeSampler(BaseSampler):
         if search_space == {}:
             return {}
 
-        sample = self._sample_lhs(study, search_space)
+        sample_trans = self._sample_lhs(study, search_space)
         trans = _SearchSpaceTransform(search_space)
-        sample = trans.untransform(sample)
+        sample = trans.untransform(sample_trans)
         return sample
 
     def after_trial(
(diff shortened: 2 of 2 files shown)


kstoneriv3:
@HideakiImamura @hvy The current implementation is based on #2423 so I assume you should be able to review it smoothly if we decide to include it to optuna.

kstoneriv3:
An example of Latin hypercube sampling algorithm:
```python
# Sample n samples of Latin hypercube samples in d dimensional unit cube (i.e. [0, 1]^d).
def LatinHypercube(n, d):

    samples = np.zeros([n, d])
    for i in range(d):
        # Enforce a unique point per grid
        perm = np.random.permutation(n)
        noise = np.random.uniform(low=0, high=1, size=n)
        samples[:, i] = perm + noise

    # Scale them to fit [0, 1]^d
    samples = 1. / n * samples
    return samples
```
This way Latin hypercube makes sure that there is always one sample in each grid `[(k-1)/n, k/n]` for any `k = 1, ..., n`.

tupui:
> * Whether it makes sense to add Latin hypercube sampling in optuna.
>   Maybe `QMCSampler` is a good enough alternative. One advantage of LatinHypercube over QMC is that the Latin hypercube is more performant when the number of parameters is large. QMCs are known to degrade their performance when the dimension of the sample space is large.

This is not true. The only case where LHS is **as** performant as QMC (Sobol') is when it's heavily optimised. There are lot of literature on the topic I can provide if necessary (ex. Sergei Kucherenko). From a pure theoretical point of view, Sobol' satisfies more uniformity properties than LHS which only provides a guarantee on 1D sub-projections.

> Theoretically, QMC excels in situations where there are several hyperparameters that all influence the objective value, which Latin hypercube is suited for situations where there are a lot of hyperparameters but only a few of them influence the objective value.

LHS is not a weighted sampling strategy, hence this is not true.

> * ~The algorithm requires that the total sample size is pre-determined. However, there is an online algorithm that does not require the information. It needs to be discussed whether we should use an online algorithm that does not require the total sample size to be pre-determined at the construction of sampler. (The algorithm can produce 2^k samples by adaptively doubling the total sample size.)~ _(Edited)_ Some of the advanced Latin hypercube algorithms supported in scipy cannot be samples this way. So I assume this is not a good sampling strategy.

Indeed, SciPy does not provide any adaptive LHS strategy. But it's definitely something I would like to have if someone can help.

kstoneriv3:
> > * Whether it makes sense to add Latin hypercube sampling in optuna.
> >   Maybe `QMCSampler` is a good enough alternative. One advantage of LatinHypercube over QMC is that the Latin hypercube is more performant when the number of parameters is large. QMCs are known to degrade their performance when the dimension of the sample space is large.
> 
> This is not true. The only case where LHS is **as** performant as QMC (Sobol') is when it's heavily optimised. There are lot of literature on the topic I can provide if necessary (ex. Sergei Kucherenko). From a pure theoretical point of view, Sobol' satisfies more uniformity properties than LHS which only provides a guarantee on 1D sub-projections.
> 
> > Theoretically, QMC excels in situations where there are several hyperparameters that all influence the objective value, which Latin hypercube is suited for situations where there are a lot of hyperparameters but only a few of them influence the objective value.
> 
> LHS is not a weighted sampling strategy, hence this is not true.

I do not get your point here. Consider objective function f(x_1, ..., x_n) = g(x_n). Regardless of n, Latin hypercube with strength 1 has the same convergence rate. On the other hand, for large n, QMC's convergence rate degrades (at least, in practice,) as far as I understand. QMCs need a relatively low effective dimension of the problem (see PCA and Brownian bridge construction of the Brownian motion using QMC) while Latin hypercube does not have such constraints.

> > * ~The algorithm requires that the total sample size is pre-determined. However, there is an online algorithm that does not require the information. It needs to be discussed whether we should use an online algorithm that does not require the total sample size to be pre-determined at the construction of the sampler. (The algorithm can produce 2^k samples by adaptively doubling the total sample size.)~ _(Edited)_ Some of the advanced Latin hypercube algorithms supported in scipy cannot be sampled this way. So I assume this is not a good sampling strategy.
> 
> Indeed, SciPy does not provide any adaptive LHS strategy. But it's definitely something I would like to have if someone can help.

I believe it is possible to sample Latin hypercube with strength 1 in an online manner so that each 2^k sample becomes a Latin hypercube by the doubling method(, which is a well-known algorithm design technique). However, for Latin hypercube with a strength of more than 1, I do not believe such an online sampling is possible.

tupui:
> I do not get your point here. Consider objective function f(x_1, ..., x_n) = g(x_n). Regardless of n, Latin hypercube with strength 1 has the same convergence rate. On the other hand, for large n, QMC's convergence rate degrades (at least, in practice,) as far as I understand. QMCs need a relatively low effective dimension of the problem (see PCA and Brownian bridge construction of the Brownian motion using QMC) while Latin hypercube does not have such constraints.

The convergence rate depends on the effective dimensionality of the function. The convergence rate of LHS is by no mean guaranteed to be better than MC. It's only guaranteed to be at worst the same as MC. QMC has a better convergence rate than LHS even when the number of effective dimensions is high.

Have a look at the following articles

* https://www.researchgate.net/publication/276210905_Exploring_multi-dimensional_spaces_a_Comparison_of_Latin_Hypercube_and_Quasi_Monte_Carlo_Sampling_Techniques
* https://www.broda.co.uk/doc/Aplication_QMC_Finance.pdf
* https://www.researchgate.net/publication/354089921_Construction_and_comparison_of_high-dimensional_Sobol'_sequence_generators (I am a co-author)

> I believe it is possible to sample Latin hypercube with strength 1 in an online manner so that each 2^k sample becomes a Latin hypercube by the doubling method(, which is a well-known algorithm design technique). However, for Latin hypercube with a strength of more than 1, I do not believe such an online sampling is possible.

This should work yes. There is a lot of literature there, we would need to look at the most used/robust algorithm.

kstoneriv3:
> The convergence rate depends on the effective dimensionality of the function. The convergence rate of LHS is by no mean guaranteed to be better than MC. It's only guaranteed to be at worst the same as MC. QMC has a better convergence rate than LHS even when the number of effective dimensions is high.
> 
> Have a look at the following articles
> 
> * https://www.researchgate.net/publication/276210905_Exploring_multi-dimensional_spaces_a_Comparison_of_Latin_Hypercube_and_Quasi_Monte_Carlo_Sampling_Techniques
> * https://www.broda.co.uk/doc/Aplication_QMC_Finance.pdf
> * https://www.researchgate.net/publication/354089921_Construction_and_comparison_of_high-dimensional_Sobol'_sequence_generators (I am a co-author)

OK, but if we assume that QMC almost always has better convergence guarantees, I do not think it makes sense to introduce Latin hypercube to Optuna, as that will just increase the maintenance cost while offering almost no benefits to users. Do you think there is still any advantage of using Latin hypercube compared to Sobol' and Halton sequences?

tupui:
> OK, but if we assume that QMC almost always has better convergence guarantees, I do not think it makes sense to introduce Latin hypercube to Optuna, as that will just increase the maintenance cost while offering almost no benefits to users. Do you think there is still any advantage of using Latin hypercube compared to Sobol' and Halton sequences?

To me, it would mainly be to get the name and possibly new users 😅 So I think it's better to get it in the existing sampler you introduced.

Though, one clear advantage of LHS is when the number of sample is low, one can use the optimised version and get better results than QMC (with the constraints of fixed sample size OC). It can be costly to do but perform better than scrambled versions of QMC (I didn't read yet optimization of QMC, I will look).

kstoneriv3:
> > ... Do you think there is still any advantage of using Latin hypercube compared to Sobol' and Halton sequences?
> 
> To me, it would mainly be to get the name and possibly new users So I think it's better to get it in the existing sampler you introduced.

Well, I do not think the current implementation of `QMCSampler` is not compatible with `scipy.stats.qmc.LatinHypercube` unfortunately, as it samples the points of QMC one by one (not all at once). Thus, the introduction of LatinHypercube should make the code unnecessarily complicated IMO.

> Though, one clear advantage of LHS is when the number of sample is low, one can use the optimised version and get better results than QMC (with the constraints of fixed sample size OC). It can be costly to do but perform better than scrambled versions of QMC (I didn't read yet optimization of QMC, I will look).

Does it mean that as long as we use the unscrambled version of QMC, Latin hypercube is no better than QMC?

tupui:
> Does it mean that as long as we use the unscrambled version of QMC, Latin hypercube is no better than QMC?

Other way around. As long as you use the scrambled version of QMC, it _should_ be better than LHS. _Should_ as in, always problem dependent, but a general rule which work most of the time.

Issue #1447, opened 2020-06-28
Optuna prunes too aggressively, when objective is jittery (early stopping patience needed)


turian:
Optuna Pruners should have a parameter `early_stopping_patience` (or `checks_patience`), which defaults to 1. If the objective hasn't improved over the last `early_stopping_patience` checks, then (early stopping) pruning occurs.

## Motivation

My objective function is jittery. So Optuna is very aggressive and prunes trials when the objective increases slightly due to jitter.

This means Optuna is **worse** than hand tuning, because it prefers only optimization paths that are not jittery and monotonically decrease, which aren't always the ones that converge to the lowest optimum. In some problem settings, it is unusable.

## Description

What would be better is an optional parameter that says how many checks must occur without improvement, in order for early stopping pruning to happen.

Optuna Pruners should have a parameter `early_stopping_patience` (or `checks_patience`), which defaults to 1. If the objective hasn't improved over the last `early_stopping_patience` checks, then (early stopping) pruning occurs.

More specifically, the objective at a particular check for the purposes of pruning, reporting, etc. should be the min (or max) objective observed over the last `early_stopping_patience` checks.

## Alternatives (optional)

I have tried `interval_steps=10` but it doesn't help. Because if the tenth step is the one that has jitter, it gets pruned.

## Additional context (optional)

Pytorch early stopping has a `patience` parameter that is similar to the one I propose.

hvy:
Out of curiosity, what pruner(s) did you try? There are several and some are more aggressive than others. It might matter less in the case of jittery learning curves, though. The pruning timing logic is already quite complicated for many of the pruners so I'd personally like to be careful about introducing additional parameters to existing ones, but I also understand the need so would like to elaborate on details first.

As a workaround, you could simulate `patience` by conditionally calling `Trial.report` on your behalf. I also think that you could run several steps without reporting, average the result, then report the averaged value. Note that the `step` argument to `Trial.report` is just a metric/order used by the pruner, could be fictional and doesn't have to reflect actual iterations, epochs, etc.

turian:
I have just started with Optuna and am trying to beat my hand-tuned baseline hyperparameters, which I cannot yet :(

I have only tried MedianPruner, and tried adjusting the settings to:

```
optuna.pruners.MedianPruner(n_warmup_steps=5, interval_steps=10)
```

[...]

PR #2636, opened 2021-04-29
Implement `PatientPruner`

nzw0301:
## Motivation

Optuna's pruners sometimes prune a trial too aggressively. So, if the objective looks like a noisy loss curve over steps, this aggressive pruning is undesirable because the pruner prunes a promising trial.

For example

- the validation accuracy on neural networks especially, with stochastic procedures; stochastic gradient, dropout, data-augmentation, etc.
- online optimisation

To address this issue, this pull request introduces a wrapper class, `PatientPruner`, of Optuna's pruners based on the discussion in #1447.

`PatientPruner` calls the wrapped pruner the objective function does not improve for `patience` consecutive steps. 

Without `wrapped`, `PatientPruner` can be viewed as the early-stopping mechanism. So the default parameters follow the convention of neural network frameworks such as [`Pytorch ignite`](https://pytorch.org/ignite/master/generated/ignite.handlers.early_stopping.EarlyStopping.html) or [Keras](https://keras.io/api/callbacks/early_stopping/).

Please see #1447 for more details.

## Description of the changes

- [x] Implement PatientPruner
- [x] Documentation
- [x] Add test

### DIFF
--- a/optuna/pruners/_patient.py
+++ b/optuna/pruners/_patient.py
@@ -72,7 +72,7 @@ class PatientPruner(BasePruner):
         if patience < 0:
             raise ValueError("patience cannot be negative but got {}.".format(patience))
 
-        self.wrapped_pruner = wrapped_pruner
+        self._wrapped_pruner = wrapped_pruner
         self._patience = patience
         self._min_delta = min_delta
 
@@ -82,35 +82,38 @@ class PatientPruner(BasePruner):
 
         steps = np.asarray(list(intermediate_values.keys()))
 
+        # the first step is never pruned.
+        if steps.size <= 1:
+            return False
+
         # Do not prune if number of step to determine are insufficient.
         if steps.size < self._patience + 1:
             return False
 
         steps.sort()
         # This is the score patience steps ago
-        steps_before_patience = steps[: -self._patience]
+        steps_before_patience = steps[: -self._patience - 1]
         scores_before_patience = np.asarray(
             list(intermediate_values[step] for step in steps_before_patience)
         )
         # And these are the scores after that
-        steps_after_patience = steps[-self._patience :]
+        steps_after_patience = steps[-self._patience - 1 :]
         scores_after_patience = np.asarray(
             list(intermediate_values[step] for step in steps_after_patience)
         )
 
         direction = study.direction
         if direction == StudyDirection.MINIMIZE:
-            maybe_prune = np.min(scores_before_patience) - self._min_delta < np.min(
+            maybe_prune = np.nanmin(scores_before_patience) - self._min_delta < np.nanmin(
                 scores_after_patience
             )
         else:
-            maybe_prune = np.max(scores_before_patience) + self._min_delta > np.max(
+            maybe_prune = np.nanmax(scores_before_patience) + self._min_delta > np.nanmax(
                 scores_after_patience
             )
-
         if maybe_prune:
-            if self.wrapped_pruner is not None:
-                return self.wrapped_pruner.prune(study, trial)
+            if self._wrapped_pruner is not None:
+                return self._wrapped_pruner.prune(study, trial)
             else:
                 return True
         else:
--- a/optuna/pruners/_patient.py
+++ b/optuna/pruners/_patient.py
@@ -77,17 +77,15 @@ class PatientPruner(BasePruner):
         self._min_delta = min_delta
 
     def prune(self, study: "optuna.study.Study", trial: "optuna.trial.FrozenTrial") -> bool:
-
-        intermediate_values = trial.intermediate_values
-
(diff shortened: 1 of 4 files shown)


optuna/pruners/_patient.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+
+    def __init__(
+        self, wrapped_pruner: Optional[BasePruner], patience: int, min_delta: float = 0.0
+    ) -> None:
+
+        if patience < 0:
+            raise ValueError("patience cannot be negative but got {}.".format(patience))
+
+        self.wrapped_pruner = wrapped_pruner
```
hvy, line 75: How about documenting this attribute if made public using the `Attributes:` annotation?

optuna/pruners/_patient.py
```diff
@@ -0,0 +1,117 @@
...
+            list(intermediate_values[step] for step in steps_after_patience)
+        )
+
+        direction = study.direction
+        if direction == StudyDirection.MINIMIZE:
+            maybe_prune = np.min(scores_before_patience) - self._min_delta < np.min(
+                scores_after_patience
+            )
+        else:
+            maybe_prune = np.max(scores_before_patience) + self._min_delta > np.min(
```
hvy, line 107: Should we take the max here?
```suggestion
            maybe_prune = np.max(scores_before_patience) + self._min_delta > np.max(
```

nzw0301:
Thank you for the feedback! I'll update this pr this week including implements tests.

optuna/pruners/_patient.py
nzw0301: yes! Thank you for suggesting it.

optuna/pruners/_patient.py
nzw0301: Indeed, but I think `wrapped_pruned` can be private rather than public because users do not need to access the wrapped pruner.

hvy: Sounds good too. We can always make it public later on  if needed. Let's prefix the property with an underscore in this PR.

hvy:
@HideakiImamura , @himkt could you have a look at this PR? Let me know if you're busy with other tasks.

tests/pruners_tests/test_patient.py
```diff
@@ -0,0 +1,81 @@
+from typing import List
+
+import pytest
+
+import optuna
+
+
```
HideakiImamura, line 7: How about testing that `PatientPruner` raises `ExperimentalWarning`?

HideakiImamura (commented):
The code basically looks good to me. I have several comments. Could you take a look?

tests/pruners_tests/test_patient.py
nzw0301: Great!  78e5672

optuna/pruners/_patient.py
```diff
@@ -0,0 +1,119 @@
...
+            Wrapped pruner to perform pruning when ``PatientPruner`` allows a
+            trial to be pruned. If it is :obj:`None`, this pruner is equivalent to
+            early-stopping taken the intermediate values in the individual trial.
+        patience:
+            Pruning is disabled until the objective doesn't improve for
+            ``patience`` consecutive steps.
+        min_delta:
+            Tolerance value to check whether or not the objective improves.
+            This value should be non-negative for minimization,
+            and it should be non-positive for maximization.
```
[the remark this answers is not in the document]
nzw0301: Thank you for your suggestion. Absolutely. it looks more consistent with python's ML ecosystem. As in 884ff98, `min_delta` takes only non-negative value now.

HideakiImamura (approved):

nzw0301:
@himkt any update? :)

optuna/pruners/_patient.py
```diff
@@ -0,0 +1,122 @@
...
+        self._patience = patience
+        self._min_delta = min_delta
+
+    def prune(self, study: "optuna.study.Study", trial: "optuna.trial.FrozenTrial") -> bool:
+        step = trial.last_step
+        if step is None:
+            return False
+
+        intermediate_values = trial.intermediate_values
+        steps = np.asarray(list(intermediate_values.keys()))
```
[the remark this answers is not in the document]
nzw0301: but I think `numpy.sort` is faster than built-in `sort`.

himkt (approved):
Sorry for the late, LGTM.
(Let me put a note that sorting intermediates on each round may cause performance issue)

nzw0301:
@HideakiImamura @himkt  Thank you for reviewing the codes! 

I understand your concern. If we score the best intermediate value somehow instead of searching it every time we call `should_prune`, we can remove the logic. I'm not sure we can do that because optuna allows us to restart training...

PR #2644, opened 2021-05-05
Refactor a unittest in test_median.py

keisuke-umezawa:
## Motivation

`test_medain.py` can be simpler using should_prune and the ask-and-tell interface.

## Description of the changes

Use should_prune and the ask-and-tell interface.

### DIFF
--- a/tests/pruners_tests/test_median.py
+++ b/tests/pruners_tests/test_median.py
@@ -4,7 +4,6 @@ from typing import Tuple
 import pytest
 
 import optuna
-from optuna.trial import TrialState
 
 
 def test_median_pruner_with_one_trial() -> None:
--- a/tests/pruners_tests/test_median.py
+++ b/tests/pruners_tests/test_median.py
@@ -9,13 +9,13 @@ from optuna.trial import TrialState
 
 def test_median_pruner_with_one_trial() -> None:
 
-    study = optuna.study.create_study()
-    trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
-    trial.report(1, 1)
     pruner = optuna.pruners.MedianPruner(0, 0)
+    study = optuna.study.create_study(pruner=pruner)
+    trial = study.ask()
+    trial.report(1, 1)
 
     # A pruner is not activated at a first trial.
-    assert not pruner.prune(study=study, trial=study._storage.get_trial(trial._trial_id))
+    assert not trial.should_prune()
 
 
 @pytest.mark.parametrize("direction_value", [("minimize", 2), ("maximize", 0.5)])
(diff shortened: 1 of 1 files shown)


tests/pruners_tests/test_median.py
```diff
@@ -26,6 +26,7 @@ def test_median_pruner_intermediate_values(direction_value: Tuple[str, float]) -
     study = optuna.study.create_study(direction=direction)
 
     trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
+    trial = study.ask()
     trial.report(1, 1)
     study._storage.set_trial_state(trial._trial_id, TrialState.COMPLETE)
```
hvy, line 31: Is using tell outside the scope of this PR?

tests/pruners_tests/test_median.py
```diff
@@ -26,6 +26,7 @@ def test_median_pruner_intermediate_values(direction_value: Tuple[str, float]) -
     study = optuna.study.create_study(direction=direction)
 
     trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
```
hvy, line 28: This line is no longer needed because of the addition on line 29.
```suggestion
```

hvy (commented):
Thanks for the PR. The motivation sounds good. How about applying the same fixes to other test cases in `test_median.py`?

keisuke-umezawa:
@hvy Thank you for reviewing. I followed your comment.

> Thanks for the PR. The motivation sounds good. How about applying the same fixes to other test cases in test_median.py?

tests/pruners_tests/test_median.py
```diff
@@ -9,97 +9,97 @@
...
+    study = optuna.study.create_study(pruner=pruner)
 
-    trial = optuna.trial.Trial(study, study._storage.create_new_trial(study._study_id))
+    trial = study.ask()
     trial.report(float("nan"), 1)
     # A pruner is not activated if the study does not have any previous trials.
-    assert not pruner.prune(study=study, trial=study._storage.get_trial(trial._trial_id))
-    study._storage.set_trial_state(trial._trial_id, TrialState.COMPLETE)
+    assert not trial.should_prune()
+    study.tell(trial, -1)
```
keisuke-umezawa, line 50: What would be a good value here? `study.tell(trial, float("nan"))` rises ValueError.

hvy: In my understanding, the value doesn't matter for the purpose of these unit tests so I don't have a strong opinion but something that communicates this, like -1 (or 0) sounds reasonable to me. We could also leave a comment.

hvy:
@himkt could you also take a look at this PR?

tests/pruners_tests/test_median.py
himkt: It also makes sense for me to use `-1` here.

hvy:
Thanks for the updated @keisuke-umezawa, LGTM!

PR #2647, opened 2021-05-06
Add a queue-like data structure to manage updated trials

not522:
## Motivation
This PR introduces a queue-like data structure that manages trials with specified states. It helps developers code incremental algorithms with updated trials. Following function/method can be more efficient with `UpdatedTrialsQueue`.

- `_pop_waiting_trial_id` #2648 
- `IntersectionSearchSpace`
- `_GroupDecomposedSearchSpace.calculate` #2649 
- `TPESampler._get_observation_pairs`

## Description of the changes
- Introduces  `UpdatedTrialsQueue`
- Add tests
- Add the `__lt__` operator in `TrialState` to compare the trial's stages

### DIFF
--- a/optuna/_updated_trials_queue.py
+++ b/optuna/_updated_trials_queue.py
@@ -39,17 +39,6 @@ class UpdatedTrialsQueue(object):
 
         self._lock = threading.Lock()
 
-    def __getstate__(self) -> Dict[Any, Any]:
-
-        state = self.__dict__.copy()
-        del state["_lock"]
-        return state
-
-    def __setstate__(self, state: Dict[Any, Any]) -> None:
-
-        self.__dict__.update(state)
-        self._lock = threading.Lock()
-
     def _fetch_trials(self, deepcopy: bool) -> List["FrozenTrial"]:
         trials = self._study.get_trials(deepcopy=deepcopy)
         next_watching_trial_indices: List[int] = []
--- a/optuna/_updated_trials_queue.py
+++ b/optuna/_updated_trials_queue.py
@@ -1,0 +1,115 @@
+from collections import deque
+import threading
+from typing import Any
+from typing import Deque
+from typing import Dict
+from typing import List
+from typing import Tuple
+from typing import TYPE_CHECKING
+
+
+if TYPE_CHECKING:
+    from optuna import Study
+    from optuna.trial import FrozenTrial
+    from optuna.trial import TrialState
+
+
+class UpdatedTrialsQueue(object):
+    """A virtual queue of trials in the specified states.
+
+    This class imitates a queue that trial is added when it goes into specified states.
+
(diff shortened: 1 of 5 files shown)


optuna/trial/_state.py
```diff
@@ -29,3 +29,13 @@ def __repr__(self) -> str:
     def is_finished(self) -> bool:
 
         return self != TrialState.RUNNING and self != TrialState.WAITING
+
+    def __lt__(self, state: "TrialState") -> bool:
```
hvy, line 33: I think it's clearer to implement this as a method/function with a name. E.g.
```suggestion
    def is_promotable_to(self, state: "TrialState") -> bool:
```

optuna/_updated_trials_queue.py
```diff
@@ -0,0 +1,115 @@
...
+        del state["_lock"]
+        return state
+
+    def __setstate__(self, state: Dict[Any, Any]) -> None:
+
+        self.__dict__.update(state)
+        self._lock = threading.Lock()
+
+    def _fetch_trials(self, deepcopy: bool) -> List["FrozenTrial"]:
+        trials = self._study.get_trials(deepcopy=deepcopy)
```
hvy, line 54: Here, we're still constructing all trials. How about taking some benchmarks before going forward to understand in what practical use cases this feature still becomes beneficial?

optuna/_updated_trials_queue.py
```diff
@@ -0,0 +1,115 @@
...
+        self._study = study
+        self._states = states
+
+        self._queue: Deque[int] = deque()
+        self._watching_trial_indices: List[int] = []
+        self._next_min_trial_index = 0
+
+        self._lock = threading.Lock()
+
+    def __getstate__(self) -> Dict[Any, Any]:
```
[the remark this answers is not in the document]
not522: In #2648, to pickle a study requires `__{get,set}state__`. They should be implemented in that PR. I removed these methods from this PR.

optuna/_updated_trials_queue.py
not522: @hvy Yes, benchmarks are necessary. I will work on it with another PR.

@HideakiImamura I'm not sure, but it may be slow to check `self._watching_trial_indices` using `get_trial` one by one.

In specific cases, it has room to improve performance. For example, `_pop_waiting_trial_id` checks a trial is available after getting it. Therefore, if `self._queue` is not empty, we can return a trial in it without `self._fetch_trials`.

optuna/_updated_trials_queue.py
```diff
@@ -0,0 +1,104 @@
+from collections import deque
```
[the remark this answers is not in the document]
not522: This feature is mainly for developers, so I made it private.  We can use it with `from optuna._updated_trials_queue import UpdatedTrialsQueue`.

PR #2651, opened 2021-05-07
Simplify `add_distributions` in `_SearchSpaceGroup`

not522:
## Motivation
This is a follow-up PR of #2526. It replaces the recursive implementation of `_add_distributions` with a non-recursive one.

## Description of the changes
- Implement non-recursive `add_distributions`
- Remove the implementation detail from the document.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -158,33 +158,6 @@ class TPESampler(BaseSampler):
             is a maximal subset of the whole search space, which satisfies the following:
             for a trial in completed trials, the intersection of the subspace and the search space
             of the trial becomes subspace itself or an empty set.
-
-            The search space is decomposed based on the following recursive rules.
-
-            - Initialize the group of the search space with the empty set. The elements of the
-              group are the subset of the search space, and the type is the dictionary of
-              :class:`~optuna.distributions.BaseDistribution`.
-            - Update the group with the following procedure `ADD(trial)` by looking at past trials.
-
-            The procedure of `Add(trial)` is
-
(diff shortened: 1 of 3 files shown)


HideakiImamura:

optuna/samplers/_tpe/sampler.py
```diff
@@ -158,33 +158,6 @@ def objective(trial):
...
-            - If an element ``S`` of a group contains ``T``, remove ``S`` from the group and add
-              ``T`` and ``S-T`` to the group.
-            - If the intersection of an element ``S`` of the group and ``T`` is not empty, remove
-              ``S`` from the group and add ``S∩T``, ``S-T``, and ``T-S`` to the group.
-              We recursively add ``T-S`` to the group because the intersection of ``T-S`` and
-              some other elements of the group may not be empty.
-
-            The group of the search space recursively constructed based on the above rules are
-            disjoint and the union is the entire search space. We perform sampling from the joint
-            distribution for each element of this decomposed group of the search space.
```
HideakiImamura, line 186: How about leaving a note as follows?

```
The group of the search space is disjoint and the union is the entire search space. 
We perform sampling from the joint distribution for each element of this decomposed 
group of the search space.
```

HideakiImamura (commented):
Thanks for the simplification! Almost LGTM. I have a comment around the deleted docs. Please take a look.

hvy (approved):
Nice, LGTM!

optuna/samplers/_tpe/sampler.py
HideakiImamura: Ops. Thanks for the clarification. You are right. Let me resolve this conversation.

PR #2637, opened 2021-04-29
Install blackdoc in `formats.sh`

nzw0301:
## Motivation

In my understanding, `blackdoc` won't be installed in `formats.sh` if `blackdoc` isn't installed. In addition, even if `blackdoc` isn't installed, `formats.sh` says `blackdoc succeeded.`. It might confuse some contributors.

## Description of the changes

Install `blackdoc` in `formats.sh` if it is missing.

### Master branch

```zsh
$ pip uninstall isort mypy black blackdoc flake8

$ ./formats.sh
The following dependencies are missing: black hacking isort
Would you like to install the missing dependencies? (y/N): y
Collecting black
  Using cached black-21.4b2-py3-none-any.whl (130 kB)
Requirement already satisfied: hacking in /opt/homebrew/Caskroom/miniconda/base/envs/optuna/lib/python3.8/site-packages (4.1.0)
Collecting isort
  Using cached isort-5.8.0-py3-none-any.whl (103 kB)
Requirement already satisfied: mypy-extensions>=0.4.3 in /opt/homebrew/Caskroom/miniconda/base/envs/optuna/lib/python3.8/site-packages (from black) (0.4.3)
Requirement already satisfied: toml>=0.10.1 in /opt/homebrew/Caskroom/miniconda/base/envs/optuna/lib/python3.8/site-packages (from black) (0.10.2)
Requirement already satisfied: regex>=2020.1.8 in /opt/homebrew/Caskroom/miniconda/base/envs/optuna/lib/python3.8/site-packages (from black) (2021.4.4)
Requirement already satisfied: appdirs in /opt/homebrew/Caskroom/miniconda/base/envs/optuna/lib/python3.8/site-packages (from black) (1.4.4)
Requirement already satisfied: click>=7.1.2 in /opt/homebrew/Caskroom/miniconda/base/envs/optuna/lib/python3.8/site-packages (from black) (7.1.2)
Requirement already satisfied: pathspec<1,>=0.8.1 in /opt/homebrew/Caskroom/miniconda/base/envs/optuna/lib/python3.8/site-packages (from black) (0.8.1)
Collecting flake8<3.9.0,>=3.8.0
  Using cached flake8-3.8.4-py2.py3-none-any.whl (72 kB)
Requirement already satisfied: mccabe<0.7.0,>=0.6.0 in /opt/homebrew/Caskroom/miniconda/base/envs/optuna/lib/python3.8/site-packages (from flake8<3.9.0,>=3.8.0->hacking) (0.6.1)
[...]

PR #2653, opened 2021-05-07
Use `command` to check the existence of the libraries to avoid partially matching

nzw0301:
CC: @hvy 

## Motivation

As mentioned in https://github.com/optuna/optuna/pull/2637#issuecomment-833354539, `formats.sh` tries to install missing libraries by combining `pip freeze` and `grep`. However, `grep` can match another library name since it is partial matching. For example, `grep mypy` matches `mypy-extensions` even if `mypy` is not installed.

## Description of the changes

Use `command` to check the existence of libraries by following  https://github.com/optuna/optuna/pull/2637#issuecomment-833998737.

### DIFF
--- a/formats.sh
+++ b/formats.sh
@@ -9,19 +9,19 @@ command -v black &> /dev/null
 if [ $? -eq 1 ] ; then
   missing_dependencies+=(black)
 fi
-command -v blackdoc &> /dev/null;
+command -v blackdoc &> /dev/null
 if [ $? -eq 1 ] ; then
   missing_dependencies+=(blackdoc)
 fi
-command -v flake8 &> /dev/null;
+command -v flake8 &> /dev/null
 if [ $? -eq 1 ] ; then
   missing_dependencies+=(hacking)
 fi
-command -v isort &> /dev/null;
+command -v isort &> /dev/null
 if [ $? -eq 1 ] ; then
   missing_dependencies+=(isort)
 fi
-command -v mypy &> /dev/null;
+command -v mypy &> /dev/null
 if [ $? -eq 1 ] ; then
   # TODO(toshihikoyanase): Unpin mypy after resolving the following issue:
   # https://github.com/optuna/optuna/issues/2240.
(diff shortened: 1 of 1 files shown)


formats.sh
```diff
@@ -4,21 +4,25 @@
...
 
-res_pip_list=$(pip freeze)
 missing_dependencies=()
-if [ ! "$(echo $res_pip_list | grep black)" ] ; then
+command -v black &> /dev/null
+if [ $? -eq 1 ] ; then
   missing_dependencies+=(black)
 fi
-if [ ! "$(echo $res_pip_list | grep blackdoc)" ] ; then
+command -v blackdoc &> /dev/null;
```
HideakiImamura, line 12: As hvy said, we don't need a semicolon here.

formats.sh
```diff
@@ -4,21 +4,25 @@
...
+if [ $? -eq 1 ] ; then
   missing_dependencies+=(blackdoc)
 fi
-if [ ! "$(echo $res_pip_list | grep flake8)" ] ; then
+command -v flake8 &> /dev/null;
```
HideakiImamura, line 16: Ditto

formats.sh
```diff
@@ -4,21 +4,25 @@
...
+if [ $? -eq 1 ] ; then
   missing_dependencies+=(hacking)
 fi
-if [ ! "$(echo $res_pip_list | grep isort)" ] ; then
+command -v isort &> /dev/null;
```
HideakiImamura, line 20: Ditto

formats.sh
```diff
@@ -4,21 +4,25 @@
...
+if [ $? -eq 1 ] ; then
   missing_dependencies+=(isort)
 fi
-if [ ! "$(echo $res_pip_list | grep mypy)" ] ; then
+command -v mypy &> /dev/null;
```
HideakiImamura, line 24: Ditto

HideakiImamura (commented):
Thanks for the PR! I have minor comments. PTAL.

formats.sh
nzw0301: Oops, Indeed. Thank you for your comments on the lines.

HideakiImamura (approved):
LGTM!

PR #2657, opened 2021-05-09
Add constraint of Sphinx version

nzw0301:
## Motivation

Sphinx v4.0.0 has been released today, but it causes CI errors related to the documents, for example, https://github.com/optuna/optuna/runs/2537658469.

## Description of the changes

Add constraint not to use the latest sphinx.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,9 @@ def get_extras_require() -> Dict[str, List[str]]:
             "mlflow",
         ],
         "document": [
-            "sphinx",
+            # TODO(nzw): Remove the version constraint after resolving the issue
+            # https://github.com/optuna/optuna/issues/2658.
+            "sphinx<4.0.0",
             "sphinx_rtd_theme",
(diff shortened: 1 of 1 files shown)


hvy (approved):
LGTM, but do you know what the root cause is or how we can fix the error? Otherwise, we could leave a TODO to come back to the issue later.

nzw0301:
No, I don't. I could not build the document correctly even if I use an older sphinx.  I agree with you; either the issue or TODO comment is necessary to fix this problem.

nzw0301:
We can resolve the issue by waiting for a newer sphinx that will be released in the near future.
https://github.com/sphinx-doc/sphinx/issues/9193

toshihikoyanase (approved):
LGTM! Thank you for your fix.

hvy:
I see, the local build is likely due to the environment, e.g. installed libraries. Thanks for adding the comment and creating the issue.

PR #2652, opened 2021-05-07
Fix distribution's `_contains`

not522:
## Motivation
Fix #2557. The distribution's `_contains` should take account of the `step` argument.

## Description of the changes
- Fix distributions' `_contains`
- Fix test

PR #2663, opened 2021-05-11
Relax `GridSampler` allowing not-contained parameters during `suggest_*`

hvy:
## Motivation

See https://github.com/optuna/optuna/pull/2652#issuecomment-836384264.

## Description of the changes

Relaxes `GridSampler` to allow parameter values in its search space that are not contained within the domain defined by `suggest_*`.


nzw0301 (approved):
Thank you for your PR! I think this change makes sense, especially, when a user comparesgrid search and another sampler.

HideakiImamura (approved):
LGTM.

Issue #2600, opened 2021-04-19
How could I get the best hyperparameter when there is high stochasticity?


junqiqian:
The set of best hyperparameters I get is found to be unstable or "lucky". It is hard to be reproduced. The performance of the reproduction experiment with the same hyperparameters is much lower than the value as before. 
Now I use one trial for one set of hyperparameters. 

**Is it possible to avoid the previous issue without adding more trials for each set of hyperparameters (otherwise it would be expensive)?**
**Is there a way to have access to the mode of distribution of the TPE sampler so that I can know what is the best expected hyperparameters?**

HideakiImamura:
Thanks for the question. I have had a similar problem before and have considered ways to deal with it. I don't know if this will be helpful but please have a look.

> Is it possible to avoid the previous issue without adding more trials for each set of hyperparameters (otherwise it would be expensive)?

In principle, I don't think it is possible to reduce the noise on the objective function value of each hyperparameter without adding more trials. It would be difficult to put any effort into the optimization itself to deal with this problem. Instead, we can put some effort into analyzing the results after optimization as follows.

> Is there a way to have access to the mode of distribution of the TPE sampler so that I can know what is the best expected hyperparameters?

No, I think we cannot calculate the posterior mean of the distribution of TPE. In TPE, we model the probability `p(x | y < c, D_n)` where `x` is a hyperparameter, `y` is the objective value, `c` is a constant and `D_n` is data collected until `n` steps. We cannot calculate the posterior mean, which is `E[y | x, D_n]` for given `x`, in principle.

fsmosca:
> The set of best hyperparameters I get is found to be unstable or "lucky". It is hard to be reproduced. The performance of the reproduction experiment with the same hyperparameters is much lower than the value as before.
> Now I use one trial for one set of hyperparameters.

Could you describe you objective function?

junqiqian:
> Thanks for the question. I have had a similar problem before and have considered ways to deal with it. I don't know if this will be helpful but please have a look.
> 
> > Is it possible to avoid the previous issue without adding more trials for each set of hyperparameters (otherwise it would be expensive)?
> 
> In principle, I don't think it is possible to reduce the noise on the objective function value of each hyperparameter without adding more trials. It would be difficult to put any effort into the optimization itself to deal with this problem. Instead, we can put some effort into analyzing the results after optimization as follows.
> 
[...]

PR #2664, opened 2021-05-11
Constant liar for `TPESampler`

hvy:
## Motivation

Fixes https://github.com/optuna/optuna/issues/892.

See also https://github.com/optuna/optuna/pull/2217.

## Description of the changes

Introduces a `constant_liar: bool` parameter to `TPESampler` to allow turning constant liar on.

When this argument is `False` (default), the previous behavior is kept. If `True`, the values of the running states are penalized with `float("inf")` or `-float("inf")` depending on the direction. A more flexible `Union[bool, Callable[[Sequence[float]], float]]` would allow arbitrary values for penalization but I left it out to keep the behavior simple avoiding confusion with how the direction is impacting the final value.

### Benchmarks

https://colab.research.google.com/drive/1_JaQGl_sCbbujTdZbd9-_MQ9Bj8dROcF?usp=sharing

### Notes

_Note that this PR may conflict with the ongoing TPE refactoring (https://github.com/optuna/optuna/pull/2618)._

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -214,6 +214,21 @@ class TPESampler(BaseSampler):
             If this is :obj:`True` and ``multivariate=True``, a warning message is emitted when
             the value of a parameter is sampled by using an independent sampler.
             If ``multivariate=False``, this flag has no effect.
+        constant_liar:
+            If :obj:`True`, penalize running trials to avoid suggesting parameter configurations
+            nearby.
+
+            .. note::
+                It is recommended to set this value to :obj:`True` during distributed
+                optimization to avoid having multiple workers evaluating similar parameter
+                configurations. In particular, if each objective function evaluation is costly
+                and the durations of the running states are significant, and/or the number of
+                workers is high.
+
+            .. note::
+                Added in v2.8.0 as an experimental feature. The interface may change in newer
+                versions without prior notice. See
+                https://github.com/optuna/optuna/releases/tag/v2.8.0.
 
     Raises:
         ValueError:
@@ -235,6 +250,7 @@ class TPESampler(BaseSampler):
         multivariate: bool = False,
         group: bool = False,
         warn_independent_sampling: bool = True,
+        constant_liar: bool = False,
     ) -> None:
 
         self._parzen_estimator_parameters = _ParzenEstimatorParameters(
@@ -255,6 +271,7 @@ class TPESampler(BaseSampler):
         self._group_decomposed_search_space: Optional[_GroupDecomposedSearchSpace] = None
         self._search_space_group: Optional[_SearchSpaceGroup] = None
         self._search_space = IntersectionSearchSpace(include_pruned=True)
+        self._constant_liar = constant_liar
 
         if multivariate:
             warnings.warn(
@@ -274,6 +291,13 @@ class TPESampler(BaseSampler):
                 ExperimentalWarning,
             )
             self._group_decomposed_search_space = _GroupDecomposedSearchSpace(True)
+
+        if constant_liar:
+            warnings.warn(
+                "``constant_liar`` option is an experimental feature."
+                " The interface can change in the future.",
+                ExperimentalWarning,
+            )
 
     def reseed_rng(self) -> None:
 
@@ -350,7 +374,9 @@ class TPESampler(BaseSampler):
             return {}
 
         param_names = list(search_space.keys())
-        values, scores = _get_multivariate_observation_pairs(study, param_names)
+        values, scores = _get_multivariate_observation_pairs(
+            study, param_names, self._constant_liar
+        )
 
         # If the number of samples is insufficient, we run random trial.
         n = len(scores)
@@ -388,7 +414,7 @@ class TPESampler(BaseSampler):
 
         self._raise_error_if_multi_objective(study)
 
-        values, scores = _get_observation_pairs(study, param_name)
+        values, scores = _get_observation_pairs(study, param_name, self._constant_liar)
 
         n = len(values)
 
@@ -866,7 +892,9 @@ class TPESampler(BaseSampler):
 
 
 def _get_observation_pairs(
-    study: Study, param_name: str
+    study: Study,
+    param_name: str,
+    constant_liar: bool = False,  # TODO(hvy): Remove default value and fix unit tests.
 ) -> Tuple[List[Optional[float]], List[Tuple[float, float]]]:
     """Get observation pairs from the study.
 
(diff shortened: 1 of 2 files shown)


HideakiImamura:
@keisuke-umezawa Could you review this PR if you have time?

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -971,3 +971,42 @@ def test_invalid_multivariate_and_group() -> None:
 def test_group_experimental_warning() -> None:
     with pytest.warns(optuna.exceptions.ExperimentalWarning):
         _ = TPESampler(multivariate=True, group=True)
+
+
+@pytest.mark.parametrize("direction", ["minimize", "maximize"])
+def test_constant_liar_observation_pairs(direction: str) -> None:
```
hvy, line 977: Open for ideas if there's a better way to test. This test uses private APIs and doesn't test the propagation of the `constant_liar` argument from `__init__` to observation pair extraction.

optuna/samplers/_tpe/sampler.py
```diff
@@ -905,6 +939,9 @@ def _get_observation_pairs(
                     score = (-step, sign * intermediate_value)
             else:
                 score = (float("inf"), 0.0)
+        elif trial.state is TrialState.RUNNING:
+            assert constant_liar
+            score = (-float("inf"), sign * float("inf"))
```
HideakiImamura, line 944: The size of the constant is fixed to infinity, but it could be made adjustable (not a strong opinion).

optuna/samplers/_tpe/sampler.py
```diff
@@ -920,16 +957,24 @@ def _get_observation_pairs(
 
 
 def _get_multivariate_observation_pairs(
-    study: Study, param_names: List[str]
+    study: Study,
```
HideakiImamura, line 960: This function is merged in #2618. I think it's good to wait for #2618 to merge this PR.

tests/samplers_tests/tpe_tests/test_sampler.py
HideakiImamura: This might be out of the scope of this PR, but I have an idea not for the test but for verification of the algorithm itself. How about plotting the density of GMM like the following setting?

- Consider low-dimensional objective function such as the rosenblock function. 
- For `x: float`, the value of the density of GMM is calculated by `TPESampler._gmm_log_pdf(samples=[x], ...)`.

HideakiImamura (commented):
Thanks for the PR! I'm very excited to see this algorithm in the Optuna. I have several comments. PTAL.

optuna/samplers/_tpe/sampler.py
hvy: I agree. I tried to address that in the description.

tests/samplers_tests/tpe_tests/test_sampler.py
hvy: Right, we could perhaps do that to check the algorithm. But I'm wondering if that isn't already verifiable using the notebook in the description. Do you think we can write unit tests based on this?

optuna/samplers/_tpe/sampler.py
```diff
@@ -214,6 +214,21 @@ def objective(trial):
             If this is :obj:`True` and ``multivariate=True``, a warning message is emitted when
             the value of a parameter is sampled by using an independent sampler.
             If ``multivariate=False``, this flag has no effect.
+        constant_liar:
```
keisuke-umezawa, line 217: In nii, they can configure constant_liar by following arguments:

> classArgs requirements:

> optimize_mode (maximize or minimize, optional, default = maximize) - If 'maximize', tuners will try to maximize metrics. If 'minimize', tuner will try to minimize metrics.
>parallel_optimize (bool, optional, default = False) - If True, TPE will use the Constant Liar algorithm to optimize parallel hyperparameter tuning. Otherwise, TPE will not discriminate between sequential or parallel situations.
>constant_liar_type (min or max or mean, optional, default = min) - The type of constant liar to use, will logically be determined on the basis of the values taken by y at X. There are three possible values, min{Y}, max{Y}, and mean{Y}.

https://github.com/microsoft/nni/blob/e85f029bd4e4b1bdf3e679893fb6447e4d6b2c79/docs/en_US/Tuner/HyperoptTuner.rst

hvy: Do mean mean we should do that in Optuna as well? I tried to address that in the PR description.

tests/samplers_tests/tpe_tests/test_sampler.py
HideakiImamura: No, I don't think we need to go that far with unit tests. Indeed, as you said, we can see that various points are being evaluated at once, as shown in the colab above.

I thought this would be a good opportunity as the request to plot KDE values has been raised by users before, but this seems to be beyond the scope of this PR. (FYI: https://github.com/optuna/optuna/issues/2600#issuecomment-824663938)

optuna/samplers/_tpe/sampler.py
HideakiImamura: Looks reasonable. @hvy How about adding `constant_liar_type` argument?

hvy: Do you mean introducing another parameter in addition to `constant_liar` ? One should be enough since it includes the boolean information. However, again, I'm worried that it'll be more confusing than useful for many users.

HideakiImamura: > One should be enough since it includes the boolean information.

You are right. It's enough to include the type of constant liar information in `constant_liar`. 

> I'm worried that it'll be more confusing than useful for many users.

I see. It's OK for me to restrict the type of constant liar to `inf` in this PR. I think it will be enough if we work on it after the users request to change the type of constant liar.

HideakiImamura (approved):
LGTM! (All my concerns are resolved. Thanks, @hvy!)

HideakiImamura:
Note: we will resolve conflicts in #2618 after merging this PR.

optuna/samplers/_tpe/sampler.py
keisuke-umezawa: Sorry, it is just FYI.

And, especially, I noticed that they named the argument from the functionality `parallel_optimize`, not from the implementation `constant_liar`.
 
The argument `constant_liar` is a little bit difficult for users to recognize that the functionality is for distributed optimization, although they can easily understand that if they read docstrings.

In my opinion, we can use `constant_liar` here, because users will read the docstrings thoroughly if they want to use the functionality.

keisuke-umezawa (approved):
LGTM! Thank you for the implementation!

PR #2670, opened 2021-05-13
MLFlow decorator for optimization function

### LINKED ISSUE #2567: MLFlow logging multiple metrics
I want to log multiple metrics in MLFlow but under the hood, Optuna uses mlflow.log_metric in its MLFlowCallBack so there is no straigthforward way to do it.

## Motivation

When I am using MLFlow to log experiments, I might want to log additional metrics even though I am not optimizing specifically for them. Or I could be using their sum in my objective function and I might want to learn their individual contribution to the objective score.

## Description

MLFlowCallBack can get additional parameters (functional parameters, not hyperparameters) to log.

## Alternatives (optional)

--

--

lucafurrer:
## Motivation
With the callback integration of MLFlow it is not possible to add additional parameter, tags and metrics. Quite often you want to keep track of more information that actually available to Optuna. P.ex
- Additional metric which is not optimized
- Static parameters
- Evolution of metric over iterations in the algorithm
- Figures

fixes #2567 

## Description of the changes
- [x] Add a function usable as decorator to integration/mlflow
- [x] Restructured current code to avoid duplication
- [x] Unit-tests
- [x] Complete documentation


HideakiImamura:
@nzw0301 Could you review this PR if you have time?

HideakiImamura:
@lucafurrer To pass the CI, please run the formatter. In Optuna, we can run black, flake8m isort, and mypy by `bash formatter.sh`.

lucafurrer:
@nzw0301 If it is ok for you, I let you have a first look at it. If from architecture perspective it looks ok to you, I'll complete the documentation.

optuna/integration/mlflow.py
```diff
@@ -13,8 +14,79 @@
...
+                tags[key] = textwrap.shorten(value, max_mlflow_tag_length)
+
+        # This sets the tags for MLflow.
+        mlflow.set_tags(tags)
+
+    def log_metric(self, value: float) -> None:
+        trial_value = value if value is not None else float("nan")
+        mlflow.log_metric(self._metric_name, trial_value)
+
+    def log_params(self, trial: optuna.trial.BaseTrial) -> None:
```
nzw0301, line 84: Can we define this function as a static method: `add @staticmethod` just before the definition?

optuna/integration/mlflow.py
```diff
@@ -85,73 +157,42 @@ def objective(trial):
...
-        # This sets the experiment of MLflow.
-        mlflow.set_experiment(study.study_name)
+        self.set_experiment(study)
 
         with mlflow.start_run(run_name=str(trial.number), nested=self._nest_trials):
 
             # This sets the metric for MLflow.
-            trial_value = trial.value if trial.value is not None else float("nan")
-            mlflow.log_metric(self._metric_name, trial_value)
+            if trial.value:
```
nzw0301, line 167: I feel this breaks a (minor) backward compatibility. Is it necessary this `if` statement? The tests seem fine without this line.

optuna/integration/mlflow.py
```diff
@@ -85,73 +157,42 @@ def objective(trial):
...
-            # see https://github.com/optuna/optuna/issues/1340
-            # see https://github.com/mlflow/mlflow/issues/2931
-            max_mlflow_tag_length = 5000
-            for key, value in tags.items():
-                value = str(value)  # make sure it is a string
-                if len(value) > max_mlflow_tag_length:
-                    tags[key] = textwrap.shorten(value, max_mlflow_tag_length)
-
-            mlflow.set_tags(tags)
+def track_in_mlflow(optuna_mlflow: MLFlowIntegrator) -> Callable:
```
nzw0301, line 177: Could you add `@experimental("2.8.0")`?

tests/integration_tests/test_mlflow.py
```diff
@@ -210,3 +212,162 @@ def test_tag_study_user_attrs(tmpdir: py.path.local, tag_study_user_attrs: bool)
...
+
+    @track_in_mlflow(mlfli)
+    def _objective_func(trial: optuna.trial.Trial) -> float:
+
+        x = trial.suggest_float("x", -1.0, 1.0)
+        y = trial.suggest_float("y", 20, 30, log=True)
+        z = trial.suggest_categorical("z", (-1.0, 1.0))
+        assert isinstance(z, float)
+        trial.set_user_attr("my_user_attr", "my_user_attr_value")
+        mlflow.log_metric(metric_name, metric)
```
nzw0301, line 236: Thank you for including this test, which is helpful to understand the usage of the integrator.


Ideally, I would like to use only `mlfli`: an instance of `MLFlowIntegrator`. What do you think about the following changes?

If we update the definition of `MLFlowIntegrator.log_metric` as fllows,

```python
    def log_metric(self, value: float, metric_name=None) -> None:
        trial_value = value if value is not None else float("nan")
        if metric_name is None:
            metric_name = self._metric_name
        mlflow.log_metric(metric_name, trial_value)
```
the reported value is still stored as `_metric_name` by default.  When we store an additional metric, we can store it by calling `mlfli.log_metric(metric_name, metric)`.

The drawback of this update is that the order of argument is swapped as in the `mlflow.log_metric`, I'm completely happy with this. so I would like to hear the opinion from you and other developers.

nzw0301 (commented):
Hi, @lucafurrer Thank you for your PR to provide a better combination of optuna and mlflow! 
Your PR looks great. Unfortunately, I'm not much familiar with mlflow, so the other developers' opinions are really helpful. Anyway, let me give several comments on this PR.

optuna/integration/mlflow.py
```diff
@@ -13,8 +14,79 @@
...
+
+        # This sets the experiment of MLflow.
+        mlflow.set_experiment(study.study_name)
+
+    def calculate_tags(self, trial: optuna.trial.BaseTrial, study: optuna.study.Study) -> None:
+
+        tags: Dict[str, str] = {}
+        tags["number"] = str(trial.number)
+        tags["datetime_start"] = str(trial.datetime_start)
+        if isinstance(trial, optuna.trial.FrozenTrial):
```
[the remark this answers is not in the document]
lucafurrer: It is quite possible that I got something wrong. I have not been looking into the internals of Optuna before, but the property `datetime_complete` is not available on other types of trial after my observation.

optuna/integration/mlflow.py
lucafurrer: MLFlow does not like None as input for log_metric, if I remember right. I write a test for that (and if it is not necessary, remove the line)

tests/integration_tests/test_mlflow.py
lucafurrer: I would like to avoid to integrate all possible functionalities of logging of MLFlow into the MLFlowIntegrator. Just out of my mind there are
- log_metric(metric_name, metric_value, step=None) and log_metrics
- log_param and log_params
- log_figure
- log_text
- log_artifact
- log_model
I would hence rather provide a simple functionality to use the features of MLFlow in Optuna without wrapping them all into an other class.

I would actually rather go and hide the MLFlowIntegrator from the User than extend it more :wink: 

But I am happy to consider an other opinion.

lucafurrer:
> Hi, @lucafurrer Thank you for your PR to provide a better combination of optuna and mlflow!
> Your PR looks great. Unfortunately, I'm not much familiar with mlflow, so the other developers' opinions are really helpful. Anyway, let me give several comments on this PR.

Hi @nzw0301, I appreciate your review and comments. I will adapt those points in the near future and added some additional information. Of course any other review is welcome!

optuna/integration/mlflow.py
lucafurrer: I just checked, and the objective function actually should always return a float, hence we should be fine without this line. (Or actually a sequence of floats -> make it compatible with both.

lucafurrer: I updated the code of `log_metric` such that it can handle `float`, `Sequence[float]` and `None`

lucafurrer:
Hi @nzw0301 I am wondering about one thing. Would this be preferable for the user:
```diff
- def track_in_mlflow(optuna_mlflow: MLFlowIntegrator) -> Callable:
+ def track_in_mlflow(tracking_uri: Optional[str] = None,
+        metric_name: Union[str, Sequence[str]] = "value",
+       nest_trials: bool = False,
+        tag_study_user_attrs: bool = False,) -> Callable:
+   mlfi = MLFlowIntegrator(tracking_uri, metric_name, nest_trials, tag_study_user_attrs)
```
What do you think? This would avoid that the user of the functionality has to instantiate the MLFlowIntegrator.

nzw0301:
Hi, @lucafurrer Thank you for your swift updates! I think the new one looks much simpler for users.

optuna/integration/mlflow.py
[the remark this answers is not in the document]
nzw0301: Thank you for your clarification. You are right! Related to the comment, can I ask why the trial's type is relaxed from `FrozenTrial` to `baseTrial` in the argument of `set_tags`?

tests/integration_tests/test_mlflow.py
nzw0301: I agree with you! Optuna does not need to provide all `log` functions in MLflow.

optuna/integration/mlflow.py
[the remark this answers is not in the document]
lucafurrer: `set_tags` is called from the Callback and the decorator.
The callback is called after the result of the optimization function has been stored by Optuna and the `Trial` has been transformed to a `FrozenTrial`. However in the context of the decorator `set_tag` is called before returning back to the Study and hence it is still a `Trial`. I thought that the most proper way to handle this difference for this two cases using this one instance check. As always I am open for other ideas.


Just as information MLFlow also tracks the duration of an trial/run. However when using the Callback functionalty this value is not correct since for MLFlow the run only starts in the callback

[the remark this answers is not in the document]
lucafurrer: In the Callback we are working with a `FrozenTrial` (executed after the optimization function call has been handled by the study.
In the Decorator we did not return the to Study yet and hence still have a `Trial`.
That's why I had to relax to `BaseTrial` and introduce the instance check. 

If you have any better idea how to approach this, I am always open.

Just as a side note: MLFlow tracks to start and end of a run (=trial) as well. However when using the Callback functionality this information is not accurate since the run is only started after the trial has ended.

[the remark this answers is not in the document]
nzw0301: Thank you for your clarification. So `@track_in_mlflow` cannot save`TrialState` and `datetime_complete` as in the tag attributes.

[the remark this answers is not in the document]
nzw0301: I think one way to resolve it is to resume the run; `track_in_mlflow` provides only `with` block to enable users to call mlflow's logging function and stores the `run_id`, then we resume the run by using `run_id` in the callback to perform mlflow's logging for `FrozenTrial`.

https://github.com/nzw0301/optuna/commit/67943f1f2fd0f977043aaef5f96cf86314103e97

I believe there is room to improve the API, but what do you think of this?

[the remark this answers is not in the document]
lucafurrer: I like it. It is less a break and more an extension of the current behaviour than my proposition
Two points I think should be discussed
- What happens when somebody does not add the Callback but sets the track_in_mlflow -> Optuna does not log anything (I can life with that)
- Multi-threaded behavior: What happens if the `nb_jobs` is set in the `optimize`-call. I know it is deprecated but still available until v4.

I am not that sure if if an attribute is the right place to store the run-id for me it looks dangerous for potential multi-threading and exception handling. My guts tell me it would be better to save it on the trial. Could `system_attrs` be used for that?

HideakiImamura:
@harupy Do you have any opinions? Since you are a member of the MLFlow organization, your comments would be greatly appreciated.

HideakiImamura:
Hi @lucafurrer! I'm going to dive into your codes from now. By the way, could you apply the code formatter to pass the CI?

optuna/integration/mlflow.py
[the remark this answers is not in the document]
HideakiImamura: Thanks for fruitful discussions @nzw0301 and @lucafurrer. I agree with @nzw0301's idea. For the concerns by @lucafurrer, I think
- If users set `track_in_mlflow` but the callback, Optuna should not log anything. It is sufficient to document the behavior enough.
- If the `n_jobs` argument is set, `MLFlowIntegrator` should raise an ValueError until v4. It is also sufficient to document the behavior.

> I am not that sure if if an attribute is the right place to store the run-id for me it looks dangerous for potential multi-threading and exception handling.

If we decide not to support the multi-threading with this MLFlow integration, we don't have to consider this problem, right? And, I'm not sure about the "exception handling". Could you elaborate a little more about it?

optuna/integration/mlflow.py
```diff
@@ -1,20 +1,149 @@
...
+
+            # Set state and convert it to str and remove the common prefix.
+            trial_state = trial.state
+            if isinstance(trial_state, TrialState) and trial_state.is_finished():
+                tags["state"] = str(trial_state).split(".")[-1]
+
+        # Set direction and convert it to str and remove the common prefix.
+        study_direction = study.direction
+        if isinstance(study_direction, StudyDirection):
+            tags["direction"] = str(study_direction).split(".")[-1]
```
HideakiImamura, line 99: This code block is not provided in this PR, but this indicates that the MLFlowIntegrator does not support multi-objective optimization. If the study is multi-objective, the call of `study.direction` fails.

Now, it is not consistent here because the argument of `metric_name` can be used for multi-objective optimization. I think we have 2 choices.
- Supprot multi-objective optimization. We need to fix this code block and maybe other places. (IMO, it is out of scope of this PR. It's fine to leave this task as future work.)
- Keep the current behavior. We skip the work on multi-objective optimization and leave the task as future work.

@nzw0301 @lucafurrer What do you think?

HideakiImamura: In my understanding, "multi-metric logging" (target of this PR) and "multi-objective optimization" (out of scope of this PR) are different. Is this right?

HideakiImamura (commented):
Thanks for the PR! The added features basically LGTM. I have several comments and would like to discuss some points. Could you take a look?

lucafurrer:
> Hi @lucafurrer! I'm going to dive into your codes from now. By the way, could you apply the code formatter to pass the CI?

I forgot it again.... I'll amend the changes mentioned above and try not to forget. 

Btw. I wanted to respond earlier, but it was a busy week :wink:

optuna/integration/mlflow.py
lucafurrer: I started to include some support for multi-objective optimization on the fly. I think currently it is not supported at all by the MLFlow Callback, but it would be nice to have. Could be a good thing to do one thing at a time and include this in a other PR.

optuna/integration/mlflow.py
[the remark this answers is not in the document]
lucafurrer: I have not thought all cases through, but since the proposed flow is
1. Set id in callback class in decorator
2. Read out id when in callback function
3. Reset id in callback class
Whenever something happens between 1 and 3 the id would not be reset. But since we would overwrite it in 1 again there would not be much harm to be done. 

But I still would prefer to store the information on the trial than in a "persistent" variable.

optuna/integration/mlflow.py
```diff
@@ -1,20 +1,149 @@
...
+        self._nest_trials = nest_trials
+        self._tag_study_user_attrs = tag_study_user_attrs
+
+    def initialize_experiment(self, study: optuna.study.Study) -> None:
+        """Initialize a MLFlow experiment with the study name.
+
+        If a tracking uri has been provided, MLFlow will be initialized to use it.
+
+        Args:
+            study (optuna.study.Study): Study to be tracked in MLFlow
```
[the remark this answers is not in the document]
lucafurrer: Noted, thanks for the hint

optuna/integration/mlflow.py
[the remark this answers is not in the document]
lucafurrer: We @nzw0301 approach by saving the run_id in a persistent variable in the class we have the following flow
1. Set run_id
2. Read run_id
Unless 1 is run again, the id is never reset. So if out of some error 1 is not set properly for the next trial, we overwrite the existing trial.

Other problematic pattern could be
```python
mlflc = MLFlowCallback(...)

@mlflc.track_in_mlflow()
def objective1(trial):
   mlflow.log_metric('something', 1)
   ...
   return x

def objective2(trial)
   ...
   return x*x

study1 = optuna.create_study(study_name="my_study")
study1.optimize(objective1, n_trials=10, callbacks=[mlflc])

study2 = optuna.create_study(study_name="my_study")
study2.optimize(objective2, n_trials=10, callbacks=[mlflc])
```
This might be far from best practice, but nothing prevents somebody to do this and the seconde study will log everything into the last run from the previous study.

I think it is possible to create other problematic scenarios with saving the current run into a variable of the class. This problems could be avoided if the run_id is not stored in MLFlowCallback but on the trial p.ex in `system_attrs` or `user_attrs`

optuna/integration/mlflow.py
```diff
@@ -1,20 +1,149 @@
...
 
 with try_import() as _imports:
     import mlflow
 
 
+@experimental("2.8.0")
+class MLFlowIntegrator(object):
+    """Helper to integrate Optuna with MLFlow.
+
+    This class allows provides the functionalities to track Optuna trials in MLFlow
```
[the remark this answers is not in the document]
lucafurrer: With the change proposed by @nzw0301 `MLFlowIntegrator` is not needed anymore -> removed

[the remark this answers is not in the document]
lucafurrer: With the change proposed by @nzw0301 `MLFlowIntegrator` is not needed anymore -> removed

optuna/integration/mlflow.py
```diff
@@ -1,20 +1,149 @@
...
+
+        Args:
+            value (Union[float, Sequence[float], None]): Result of trial
+        """
+        if isinstance(value, Sequence):
+            for key, v in zip(self._metric_name, value):
+                mlflow.log_metric(key, v)
+        else:
+            trial_value = value if value is not None else float("nan")
+            mlflow.log_metric(self._metric_name, trial_value)
```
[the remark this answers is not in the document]
lucafurrer: As you mentioned later on, the multi objective support needs additional changes and I agree that it should be moved to an other PR. Since this part is only used for logging the objective value and all other metrics should be logged in the objective function itself, there is no need for having the sequence yet.

optuna/integration/mlflow.py
```diff
@@ -1,20 +1,149 @@
...
+from optuna.study import ObjectiveFuncType
 from optuna.trial import TrialState
 
 
 with try_import() as _imports:
     import mlflow
 
 
+@experimental("2.8.0")
+class MLFlowIntegrator(object):
```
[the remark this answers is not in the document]
lucafurrer: Removed class entierly

lucafurrer:
Hi @HideakiImamura and @nzw0301 
- I integrated the suggestion of @nzw0301 and this allowed me to drop MLFlowIntegrator. I still have some reservation about the usage of the `_last_run_id` and am interested in your opinion.
- I removed all traces of multi objective support as @HideakiImamura pointed out that it is not properly working for the moment anyway and I agree that this should be an other PR (I might have a look into it after t his one).

optuna/integration/mlflow.py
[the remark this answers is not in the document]
nzw0301: Thank you for your informative comments! As another scenario, we might want to resume optimization of the study like your example provided just before the comment of this comment with `optuna.study.load_study`. So personally, a re-usable callback sounds fine for some cases.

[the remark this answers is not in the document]
lucafurrer: Ok this should be fine with any of the implementations. I was also thinking about the fact that in any cases it might be useful to have a reference to the MLFlow-Run-ID stored in Optuna to allow jumping from one to an other. Still not sure if it should be `system_attrs` or `user_attrs`

nzw0301:
@lucafurrer Sorry for the inactive response to this PR -- I'm a little bit busy due to my main job until 7th June.

Could you resolve the `Spinx / doctest` error? 

Since the state of this PR is still draft, is there something to do for this PR? Otherwise, we can remove the draft status.

lucafurrer:
@nzw0301 May day job kept my quite busy lately as well. 
I'll try to fix the sphinx/doctest job. Last time I looked into it I did not see the problem, but I should find it.

Only open point otherwise would be to store the mlflow run id in the trial, but the rest is fine. Would still be functional without this.

lucafurrer:
Hi, how can I efficiently run `Sphinx / doctest` locally. `make doctest` takes extermely long and and I am think there is missing something in my setup to make `act -j` work

nzw0301:
> Sphinx / doctest locally.

How about replace [this line](https://github.com/optuna/optuna/blob/29d3ae2ec41fda4091395941aa9cdc953b06d3b1/docs/Makefile#L23) by `sphinx-autogen source/reference/integration.rst`?

nzw0301 (changes requested):
I'm still not sure that using either `user_attr` or `system_attr` to store `run_id` is a solution. It might break the existing code if a users access attributes. Personally, I still believe storing run_id as an instance variable is okay.

@HideakiImamura what do you think?

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,124 @@ def __init__(
...
+                with mlflow.start_run(run_name=str(trial.number), nested=self._nest_trials) as run:
+                    trial.set_user_attr(RUN_ID_ATTRIBUTE_KEY, run.info.run_id)
+
+                    return func(trial)
+
+            return wrapper
+
+        return decorator
+
+    def initialize_experiment(self, study: optuna.study.Study) -> None:
```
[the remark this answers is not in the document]
lucafurrer: Done

tests/integration_tests/test_mlflow.py
```diff
@@ -210,3 +211,169 @@ def test_tag_study_user_attrs(tmpdir: py.path.local, tag_study_user_attrs: bool)
...
+    study_name = "my_study"
+
+    param1_name = "my_param1"
+    param1_value = "a"
+    param2_name = "my_param2"
+    param2_value = 5
+
+    params = {param1_name: param1_value, param2_name: param2_value}
+
+    mlfli = MLflowCallback(tracking_uri=tracking_file_name, metric_name=metric_name)
```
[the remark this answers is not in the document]
lucafurrer: Missed this one

HideakiImamura (commented):
@lucafurrer @nzw0301 Sorry for the late review. The implementation almost looks good to me. I think we should not use `trial.user_attr` in the library. Please use `system_attr` instead.

@harupy Do you have any comments on this feature? I'm not sure if this feature is useful for MLFlow users or not, and I'm not sure if we are using standard MLFlow features; if it's feasible using other features of MLFlow, I'd appreciate it if you could let me know.

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,124 @@ def __init__(
...
+        # This is a temporary fix on Optuna side. It avoids an error with user
+        # attributes that are too long. It should be fixed on MLflow side later.
+        # When it is fixed on MLflow side this codeblock can be removed.
+        # see https://github.com/optuna/optuna/issues/1340
+        # see https://github.com/mlflow/mlflow/issues/2931
+        max_mlflow_tag_length = 5000
+        for key, value in tags.items():
+            value = str(value)  # make sure it is a string
+            if len(value) > max_mlflow_tag_length:
+                tags[key] = textwrap.shorten(value, max_mlflow_tag_length)
```
harupy, line 243:
```suggestion
        for key, value in tags.items():
            value = str(value)  # make sure it is a string
            if len(value) > max_mlflow_tag_length:
                tags[key] = textwrap.shorten(value, mlflow.utils.validation.MAX_TAG_VAL_LENGTH)
```

lucafurrer: thanks @harupy I integrated this suggestion.

lucafurrer:
Hi @HideakiImamura and @harupy  I liked and integrated you suggestions :smiley: Thanks a lot.

nzw0301 (approved):
LGTM. Thank you for your long run PR!

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+        mlflow.set_tags(tags)
+
+    def _log_metric(self, value: Union[float, Sequence[float], None]) -> None:
+        """Log the trial result as metric to MLFlow
+
+        Args:
+            value: Result of trial
+        """
+        trial_value = value if value is not None else float("nan")
+        mlflow.log_metric(self._metric_name, trial_value)
```
harupy, line 253: The current implementation doesn't seem to handle the case where `value` is an array.

tests/integration_tests/test_mlflow.py
```diff
@@ -210,3 +211,169 @@ def test_tag_study_user_attrs(tmpdir: py.path.local, tag_study_user_attrs: bool)
...
+    study = optuna.create_study(study_name=study_name)
+
+    mlflc._initialize_experiment(study)
+
+    mlfl_client = MlflowClient(tracking_file_name)
+    experiments = mlfl_client.list_experiments()
+    assert len(experiments) == 1
+
+    experiment = experiments[0]
+    assert experiment.name == study_name
```
harupy, line 274: This is not a big issue, but the code here can be simplified to: 

```suggestion
    mlfl_client = MlflowClient(tracking_file_name)
    assert mlfl_client.get_experiment_by_name(study_name) is None
```

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+                    trial.set_system_attr(RUN_ID_ATTRIBUTE_KEY, run.info.run_id)
+
+                    return func(trial)
+
+            return wrapper
+
+        return decorator
+
+    def _initialize_experiment(self, study: optuna.study.Study) -> None:
+        """Initialize a MLFlow experiment with the study name.
```
harupy, line 189:
```suggestion
        """Initialize an MLflow experiment with the study name.
```

nit

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+
+        # This is a temporary fix on Optuna side. It avoids an error with user
+        # attributes that are too long. It should be fixed on MLflow side later.
+        # When it is fixed on MLflow side this codeblock can be removed.
+        # see https://github.com/optuna/optuna/issues/1340
+        # see https://github.com/mlflow/mlflow/issues/2931
+        for key, value in tags.items():
+            value = str(value)  # make sure it is a string
+            if len(value) > mlflow.utils.validation.MAX_TAG_VAL_LENGTH:
+                tags[key] = textwrap.shorten(value, mlflow.utils.validation.MAX_TAG_VAL_LENGTH)
```
harupy, line 241: I've never used `textwrap.shorten` before so I checked how it works:

```python
>>> textwrap.shorten("0123456789", width=10)
'0123456789'
>>> textwrap.shorten("0123456789", width=9)
[...]
>>> textwrap.shorten("012 3456789", width=9)
012 [...]
```

Maybe a simpler truncation logic like below makes more sense here:

```python
def truncate(val, width, placeholder="..."):
    if len(val) > width:
        return val[: width - len(placeholder)] + placeholder
    return val

>>> truncate("0123456789", width=10)
'0123456789'
>>> truncate("0123456789", width=9)
012345...
>>> truncate("012 3456789", width=9)
0123 4...
```

tests/integration_tests/test_mlflow.py
harupy: Ah, other tests in this file use the current approach.

harupy:
@lucafurrer Left a few more minor comments, but otherwise looks good :)

optuna/integration/mlflow.py
lucafurrer: True this is a leftover from a half way attempt to make the code compatible with a sequence, as discussed with @HideakiImamura this would be the topic for an other PR.

optuna/integration/mlflow.py
lucafurrer: To be honest, I did just move the code around while refactoring. So changing this would change the behavior of the current implementation hence I am a little bit reluctant to change it even if I see your point.

nzw0301: @lucafurrer I agree with you. The refactoring of this logic is out of the scope of this PR. Of course, we can deal with this part as another PR later.

nzw0301: @lucafurrer I agree with you. The refactoring of this logic is out of the scope of this PR. As you said, this logic already exists in the MLFlow code of optuna.

@harupy Thank you for investigating it! Your suggestion looks better and more interpretable. I think that we should deal with this part as another PR!

harupy: +1 for addressing this in a follow-up PR

optuna/integration/mlflow.py
```diff
@@ -88,7 +127,7 @@ def objective(trial):
     def __init__(
         self,
         tracking_uri: Optional[str] = None,
-        metric_name: str = "value",
+        metric_name: Union[str] = "value",
```
HideakiImamura, line 130:
```suggestion
        metric_name: str = "value",
```

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+        with mlflow.start_run(
+            run_id=trial.system_attrs.get(RUN_ID_ATTRIBUTE_KEY),
+            run_name=str(trial.number),
+            nested=self._nest_trials,
+        ):
 
             # This sets the metric for MLflow.
-            trial_value = trial.value if trial.value is not None else float("nan")
-            mlflow.log_metric(self._metric_name, trial_value)
+            self._log_metric(trial.value)
```
HideakiImamura, line 153: We should add a comment like "As future work, we should support the multi-objective case" something like that.

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
 
             # This sets the params for MLflow.
-            mlflow.log_params(trial.params)
+            self._log_params(trial)
```
HideakiImamura, line 156: It is more consistent to have a dictionary instead of the instance of the Trial.
```suggestion
            self._log_params(trial.params)
```

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+        """Log the trial result as metric to MLFlow
+
+        Args:
+            value: Result of trial
+        """
+        trial_value = value if value is not None else float("nan")
+        mlflow.log_metric(self._metric_name, trial_value)
+
+    @staticmethod
+    def _log_params(trial: optuna.trial.BaseTrial) -> None:
```
HideakiImamura, line 256: Ditto.
```suggestion
    def _log_params(params: Dict[str, Any]) -> None:
```

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+
+    def _initialize_experiment(self, study: optuna.study.Study) -> None:
+        """Initialize an MLFlow experiment with the study name.
+
+        If a tracking uri has been provided, MLFlow will be initialized to use it.
+
+        Args:
+            study: Study to be tracked in MLFlow
+        """
+        # This sets the tracking_uri for MLflow.
```
HideakiImamura, line 196: It is more consistent with other Optuna's codebase to use a backtick for variables.
```suggestion
        # This sets the `tracking_uri` for MLflow.
```

optuna/integration/mlflow.py
HideakiImamura: +1 in a follow-up PR.

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+        if self._tracking_uri is not None:
+            mlflow.set_tracking_uri(self._tracking_uri)
+
+        # This sets the experiment of MLflow.
+        mlflow.set_experiment(study.study_name)
+
+    def _set_tags(self, trial: optuna.trial.FrozenTrial, study: optuna.study.Study) -> None:
+        """Sets the Optuna tags for the current MLFlow run
```
HideakiImamura, line 204: Should have a period.
```suggestion
        """Sets the Optuna tags for the current MLFlow run.
```

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+
+        Args:
+            trial: Trial to be tracked
+            study: Study to be tracked
```
HideakiImamura, line 208: Ditto.
```suggestion
            trial: Trial to be tracked.
            study: Study to be tracked.
```

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+    def track_in_mlflow(self) -> Callable:
+        """Decorator for using MLFlow logging in the objective function.
+
+        This decorator enables the extension of MLFlow logging provided by the callback.
+
+        All information logged in the decorated objective function will be added to the MLFlow
+        run for the trial created by the callback.
+
+        Returns:
+            ObjectiveFuncType: Objective function with tracking to MLFlow enabled
```
HideakiImamura, line 171: Ditto.
```suggestion
            ObjectiveFuncType: Objective function with tracking to MLFlow enabled.
```

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+                tags[key] = textwrap.shorten(value, mlflow.utils.validation.MAX_TAG_VAL_LENGTH)
+
+        # This sets the tags for MLflow.
+        mlflow.set_tags(tags)
+
+    def _log_metric(self, value: Union[float, None]) -> None:
+        """Log the trial result as metric to MLFlow
+
+        Args:
+            value: Result of trial
```
HideakiImamura, line 250: Ditto.
```suggestion
        """Log the trial result as metric to MLFlow.

        Args:
            value: Result of trial.
```

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+        """
+        trial_value = value if value is not None else float("nan")
+        mlflow.log_metric(self._metric_name, trial_value)
+
+    @staticmethod
+    def _log_params(trial: optuna.trial.BaseTrial) -> None:
+        """Log the parameters of the trial to MLFlow
+
+        Args:
+            trial: Trial to be tracked
```
HideakiImamura, line 260: Ditto.
```suggestion
        """Log the parameters of the trial to MLFlow.

        Args:
            trial: Trial to be tracked.
```

HideakiImamura (approved):
@lucafurrer Thanks for the long-running PR! LGTM! And, Thanks for your insightful reviews @nzw0301 @harupy! 

Let me leave some nit TODOs for the follow-up PR.

lucafurrer:
Thanks @HideakiImamura, @nzw0301 and @harupy for your time and review. It was a great experience to create my first PR for Optuna.

optuna/integration/mlflow.py
```diff
@@ -102,56 +141,122 @@ def __init__(
...
+
+        # Set state and convert it to str and remove the common prefix.
+        trial_state = trial.state
+        if isinstance(trial_state, TrialState) and trial_state.is_finished():
+            tags["state"] = str(trial_state).split(".")[-1]
+
+        # Set direction and convert it to str and remove the common prefix.
+        study_direction = study.direction
+        if isinstance(study_direction, StudyDirection):
+            tags["direction"] = str(study_direction).split(".")[-1]
```
nzw0301, line 224: Let me leave a comment for follow-up PR. This lien can be simplified like
```python
            tags["direction"] = direction = study.direction.name
```

Ref https://github.com/optuna/optuna/pull/2781#discussion_r665817263

PR #2673, opened 2021-05-14
Introduce speed benchmarks using asv

not522:
## Motivation
Benchmarking speed can help us implement speedups and detect slowdowns. This PR introduces benchmarks using [asv](https://asv.readthedocs.io/en/stable/).

## Description of the changes
- Add benchmarks
- Add speed benchmarks in GitHub Actions ([Result](https://github.com/not522/optuna/runs/2563402752))
- Configuration

## Discussion
The Redis benchmark use fakeredis. It may be misleading, but developers easily run benchmarks without setting up a Redis server.

### DIFF
--- a/.github/workflows/speed-benchmarks.yml
+++ b/.github/workflows/speed-benchmarks.yml
@@ -1,9 +1,6 @@
 name: Speed benchmarks
 
 on:
-  push:
-    branches:
-      - asv
   schedule:
     - cron: '0 23 * * SUN-THU'
 
@@ -12,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
 
     # Not intended for forks.
-    # if: github.repository == 'optuna/optuna'
+    if: github.repository == 'optuna/optuna'
 
     steps:
     - name: Checkout
--- a/.github/workflows/speed-benchmarks.yml
+++ b/.github/workflows/speed-benchmarks.yml
@@ -15,10 +15,10 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
-    - name: Setup Python3.7
+    - name: Setup Python3.9
       uses: actions/setup-python@v2
       with:
-        python-version: 3.7
+        python-version: 3.9
 
     - name: Setup cache
       uses: actions/cache@v2
@@ -26,9 +26,9 @@ jobs:
         cache-name: speed-benchmarks
       with:
         path: ~/.cache/pip
-        key: ${{ runner.os }}-3.7-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}-v1
+        key: ${{ runner.os }}-3.9-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}-v1
         restore-keys: |
-          ${{ runner.os }}-3.7-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}
+          ${{ runner.os }}-3.9-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}
 
     - name: Install
       run: |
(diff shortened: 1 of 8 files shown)


HideakiImamura:

benchmarks/optimize.py
```diff
@@ -0,0 +1,70 @@
...
+        return RandomSampler()
+    elif sampler_mode == "tpe":
+        return TPESampler()
+    elif sampler_mode == "cmaes":
+        return CmaEsSampler()
+    else:
+        assert False
+
+
+class TimeOptimize:
```
c-bata, line 41: I'd say naming classes like `XxxSuite` is more common.

https://github.com/scikit-image/scikit-image/tree/main/benchmarks
https://asv.readthedocs.io/en/stable/writing_benchmarks.html

```suggestion
class OptimizeSuite:
```

c-bata (commented):
Basically LGTM 👍 

I'm curious about how are we monitoring benchmark results? Are we supposed to prepare a web page like https://pandas.pydata.org/speed/pandas/?

c-bata (approved):

.github/workflows/speed-benchmarks.yml
```diff
@@ -0,0 +1,48 @@
...
+    if: github.repository == 'optuna/optuna'
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Setup Python3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
```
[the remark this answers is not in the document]
not522: Thank you for your suggestion! I updated to Python 3.9, and it works. https://github.com/not522/optuna/actions/runs/842345303

not522:
Currently, we have no concrete plan to deploy the benchmark web page. asv has a cool web interface, but we need to store the benchmark result somewhere.

benchmarks/optimize.py
```diff
@@ -0,0 +1,70 @@
...
+    params = (
+        "inmemory, random, 1000",
+        "inmemory, random, 10000",
+        "inmemory, tpe, 1000",
+        "inmemory, cmaes, 1000",
+        "sqlite, random, 1000",
+        "cache, random, 1000",
+        "redis, random, 1000",  # This benchmark uses fakeredis instead of Redis.
+    )
+    param_names = "storage, sampler, n_trials"
```
toshihikoyanase, line 69: According to the following result, the `param_name` was recognized as `s`. Do we need to put a list of strings here?

```suggestion
    param_names = ["storage, sampler, n_trials"]
```
```console
[100.00%] ··· ========================= ============                                                                                                                                                                 s                                                                                                                                                                              ------------------------- ------------                                                                                                                                                       inmemory, random, 1000    344±4ms                                                                                                                                                        inmemory, random, 10000   9.39±0.1s                                                                                                                                                         inmemory, tpe, 1000     6.27±0.2s                                                                                                                                                        inmemory, cmaes, 1000     987±10ms                                                                                                                                                         sqlite, random, 1000     27.4±1s                                                                                                                                                          cache, random, 1000     26.1±0.7s                                                                                                                                                         redis, random, 1000     1.67±0.01m                                                                                                                                                     ========================= ============ 
```

benchmarks/optimize.py
```diff
@@ -0,0 +1,70 @@
+from typing import cast
```
toshihikoyanase, line 1: Do we need to include the `benchmarks` module into the `optuna` distribution package? If not, could you exclude the `benchmark` directory?

When I built a dist file with `python setup.py sdist` and installed the tar ball, I can import the `benchmarks`. This is due to the sdist file includes the `benchmarks` directory.

toshihikoyanase (changes requested):
Thank you for your PR. The benchmark with `asv` looks great to me.
I added two small comments.

not522:
Thank you for your suggestions. I updated the PR.

toshihikoyanase (approved):

PR #2679, opened 2021-05-15
Treat `None` as special value in AllenNLP integration

himkt:
- Ref: https://github.com/himkt/allennlp-optuna/issues/39

## Motivation
Optuna pruning integration for AllenNLP creates a pruner using environment variables.
If a value None (None object) is specified in config, it would be parsed as "None" (string literal) when constructing a pruner. It causes the problem you encountered (referred from https://github.com/himkt/allennlp-optuna/issues/39#issuecomment-841614525).

This PR fixes a bug to parse the `None` object.

If `None` is passed to AllenNLP integration, Optuna rewrites it to the special symbol `_NONE` (`"<PYTHON_NONE_OBJECT>"`) and sets it to environment variable. When constructing a pruner from environment variables, the symbol is parsed as Python `None` object.

## Description of the changes
- 1d22815 updates a test to check if a pruner works well with `None`
- c410a0a adds a special logic to treat Python `None` object

---

before: https://github.com/optuna/optuna/runs/2589537538
after: https://github.com/optuna/optuna/runs/2589569845

### DIFF
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -111,7 +111,7 @@ def _infer_and_cast(value: Optional[str]) -> Optional[Union[str, int, float, boo
     to desired types.
 
     """
-    if value in (None, _NONE):
+    if value is None or value == _NONE:
         return None
 
     try:
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -72,7 +72,6 @@ _STUDY_NAME = "{}_STUDY_NAME".format(_PREFIX)
 _TRIAL_ID = "{}_TRIAL_ID".format(_PREFIX)
 
 
-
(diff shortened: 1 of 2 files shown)


optuna/integration/allennlp.py
```diff
@@ -104,7 +111,7 @@ def _infer_and_cast(value: Optional[str]) -> Optional[Union[str, int, float, boo
     to desired types.
 
     """
-    if value is None:
+    if value is None or value == _NONE:
```
himkt, line 114: 3b7e68f mypy doesn't accept `value in (None, _NONE)`.

himkt:
The problem occurs due to parameter storing logic.
Let me close this PR and work on fixing AllenNLP integration in #2682.

PR #2686, opened 2021-05-18
Do not wait for next heartbeat on study completion.

### LINKED ISSUE #2685: Waiting for Heartbeat
## Expected behavior

After `study.optimize` with a high heartbeat interval (e.g. 60s) finishes running trials, it should exit almost immediately. 

## Environment

- Optuna version: 2.7.0
- Python version: 3.9
- OS: Fedora 33
- (Optional) Other libraries and their versions: PostgreSQL

## Observed Behavior
Instead of exiting immediately, the main thread tries to join with the heartbeat thread, which only completes after the next scheduled hearbeat. This can lead to high delays where the program does basically nothing.

## Steps to reproduce

1. Create a RDB storage with a heartbeat interval of 60s.
2. Create a fast objective, a study and call `study.optimize(n_trials=1)`.
3. Measure runtime.

It might be better to use the [`wait(timeout=...)`](https://docs.python.org/3/library/threading.html#threading.Event.wait) method [here](https://github.com/optuna/optuna/blob/2fd68d0102a22e76b5a4a142cc87722d92dcf667/optuna/_optimize.py#L340).

Turakar:
Fixes #2685. This PR contains the proposed change exactly.

### DIFF
--- a/optuna/_optimize.py
+++ b/optuna/_optimize.py
@@ -11,7 +11,6 @@ import os
 import sys
 from threading import Event
 from threading import Thread
-import time
 from typing import Any
 from typing import Callable
 from typing import cast
@@ -335,6 +334,5 @@ def _record_heartbeat(trial_id: int, storage: storages.BaseStorage, stop_event: 
     assert heartbeat_interval is not None
     while True:
         storage.record_heartbeat(trial_id)
-        if stop_event.is_set():
+        if stop_event.wait(timeout=heartbeat_interval):
             return
-        time.sleep(heartbeat_interval)


hvy:
@Turakar, thanks you for the quick PR!

@HideakiImamura could you also have/ a look at this PR?

HideakiImamura (approved):
Looks solid! LGTM!

hvy (approved):
Sorry for the delayed response. Thanks a lot for the PR and LGTM!

It also seems like `Event.wait` is efficiently implemented in Python 3.

Turakar:
Just a quick query: When do you plan to release v2.8.0?

hvy:
Although not fixed, we are planning a release in July. Do you have some preference?

Turakar:
As early as possible, but I guess I will go for nightly or some custom patch then.

PR #2694, opened 2021-05-21
Add automatic retry callback

Crissman:
## Motivation

In some evironments, pre-emption or server disruption may be common, and it would be helpful to have Optuna able to automatically queue up trials that were interupted.

## Description of the changes

Add RetryFailedTrialCallback to the optuna/_callbacks.py for the storage trial_failed_callbacks.

Todo:

- [x] Convert to class to enable max retries
- [x] Add documentation
- [x] Add testing

### DIFF
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -86,9 +86,14 @@ class RetryFailedTrialCallback:
                 storage=storage,
             )
 
+    Args:
+        max_retry:
+            The max number of times a trial can be retried. Must be set to :obj:`None` or an
+            integer. If set to the default value of :obj:`None` will retry indefinitely.
+            If set to an integer, will only retry that many times.
     """
 
-    def __init__(self, max_retry: int) -> None:
+    def __init__(self, max_retry: Optional[int] = None) -> None:
         self._max_retry = max_retry
 
     def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
@@ -104,7 +109,7 @@ class RetryFailedTrialCallback:
             for s in study.trials
         )
 
-        if retries + 1 > self._max_retry:
+        if self._max_retry is not None and retries + 1 > self._max_retry:
             return
 
         study.add_trial(
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -69,21 +69,36 @@ def RetryFailedTrialCallback(study, trial):
 
         .. testcode::
 
-            study = optuna.create_study()
-            study.optimize(
-                objective,
-                callbacks=[MaxTrialsCallback(10, states=(TrialState.COMPLETE,))],
+            storage = optuna.storages.RDBStorage(
+                args.optuna_storage,
+                heartbeat_interval=60,
+                grace_period=120,
+                failed_trial_callback=RetryFailedTrialCallback(max_retry=3),
             )
 
+            study = optuna.create_study(
+                storage=storage,
+            )
     """
 
-    trial.user_attrs.update({"last_failed_trial": trial.number})
-    study.add_trial(
-        optuna.create_trial(
-            state=optuna.trial.TrialState.WAITING,
-            params=trial.params,
-            distributions=trial.distributions,
-            user_attrs=trial.user_attrs,
-            system_attrs=trial.system_attrs,
+    def __init__(self, max_retry: int) -> None:
+        self._max_retry = max_retry
+
+    def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
+        failed_trials = (
+            trial.user_attrs["failed_trials"] if "failed_trials" in trial.user_attrs else []
         )
-    )
+        failed_trials.append(trial.number)
+        if failed_trials.count(trial.number) > self._max_retry:
+            return
+
+        trial.user_attrs.update({"last_failed_trial": trial.number})
+        study.add_trial(
+            optuna.create_trial(
+                state=optuna.trial.TrialState.WAITING,
+                params=trial.params,
+                distributions=trial.distributions,
+                user_attrs=trial.user_attrs,
+                system_attrs=trial.system_attrs,
+            )
+        )
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -75,7 +75,7 @@ class RetryFailedTrialCallback:
             from optuna.storages import RetryFailedTrialCallback
 
             storage = optuna.storages.RDBStorage(
-                url='sqlite:///:memory:',
+                url="sqlite:///:memory:",
                 heartbeat_interval=60,
                 grace_period=120,
                 failed_trial_callback=RetryFailedTrialCallback(max_retry=3),
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -71,7 +71,7 @@ class RetryFailedTrialCallback:
 
         .. testcode::
 
-            from optuna.storage import RetryFailedTrialCallback
+            from optuna.storages import RetryFailedTrialCallback
 
             storage = optuna.storages.RDBStorage(
                 args.optuna_storage,
@@ -90,14 +90,17 @@ class RetryFailedTrialCallback:
         self._max_retry = max_retry
 
     def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
+        user_attrs = {"failed_trial": trial.number}
+        user_attrs.update(trial.user_attrs)
         failed_trials = (
-            trial.user_attrs["failed_trials"] if "failed_trials" in trial.user_attrs else []
+            study.user_attrs["failed_trials"] if "failed_trials" in study.user_attrs else []
         )
-        failed_trials.append(trial.number)
+        failed_trials.append(user_attrs["failed_trial"])
+        study.user_attrs.update({"failed_trials": failed_trials})
+        print(study.user_attrs, user_attrs, failed_trials)
         if failed_trials.count(trial.number) > self._max_retry:
             return
 
-        trial.user_attrs.update({"last_failed_trial": trial.number})
         study.add_trial(
             optuna.create_trial(
                 state=optuna.trial.TrialState.WAITING,
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -119,7 +119,7 @@ class RetryFailedTrialCallback:
 
     @staticmethod
     def retried_trial_number(trial: FrozenTrial) -> Optional[int]:
-        """Return the number of the trail being retried.
+        """Return the number of the trial being retried.
 
         Args:
             trial:
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -123,7 +123,7 @@ class RetryFailedTrialCallback:
 
         Args:
             trial:
-                The :obj:`trial` object.
+                The trial object.
 
         Returns:
             The number of the previous trial. If not retry of a previous trial,
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -105,8 +105,8 @@ class RetryFailedTrialCallback:
         system_attrs.update(trial.system_attrs)
 
         retries = sum(
-            ("failed_trial", system_attrs["failed_trial"]) in s.system_attrs.items()
-            for s in study.trials
+            ("failed_trial", system_attrs["failed_trial"]) in t.system_attrs.items()
+            for t in study.trials
         )
 
         if self._max_retry is not None and retries + 1 > self._max_retry:
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -118,7 +118,7 @@ class RetryFailedTrialCallback:
         )
 
     @staticmethod
-    def retried_trial_number(trial) -> Optional[int]:
+    def retried_trial_number(trial: FrozenTrial) -> Optional[int]:
         """Return the number of the trail being retried.
 
         Args:
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -59,7 +59,7 @@ class MaxTrialsCallback:
 
 @experimental("2.8.0")
 class RetryFailedTrialCallback:
-    """Retry a failed trial up to a maximum number of times
+    """Retry a failed trial up to a maximum number of times.
 
     When a trial fails, this callback can be used with the :class:`optuna.storage` class to
     recreate the trial in :obj:`TrialState.WAITING` to queue up the trial to be run again.
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -91,20 +91,22 @@ class RetryFailedTrialCallback:
         self._max_retry = max_retry
 
     def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
-        retries = sum(("failed_trial", trial.number) in s.user_attrs.items() for s in study.trials)
+        retries = sum(
+            ("failed_trial", trial.number) in s.system_attrs.items() for s in study.trials
+        )
 
         if retries + 1 > self._max_retry:
             return
 
-        user_attrs = {"failed_trial": trial.number}
-        user_attrs.update(trial.user_attrs)
+        system_attrs = {"failed_trial": trial.number}
+        system_attrs.update(trial.system_attrs)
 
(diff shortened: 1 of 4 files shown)


optuna/_callbacks.py
```diff
@@ -55,3 +55,58 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+        user_attrs = {"failed_trial": trial.number}
+        user_attrs.update(trial.user_attrs)
+        failed_trials = (
+            study.user_attrs["failed_trials"] if "failed_trials" in study.user_attrs else []
+        )
+        failed_trials.append(user_attrs["failed_trial"])
+        study.user_attrs.update({"failed_trials": failed_trials})
+        print(study.user_attrs, user_attrs, failed_trials)
+        if failed_trials.count(trial.number) > self._max_retry:
+            return
```
hvy, line 102: Is the user attribute of the study used for counting only? I'm wondering if we cannot get the count without the study user attribute, i.e. by looping over all trials in this method (it's not performance critical so it should be okay). This assumes though that retried trials are labeled in some way.

optuna/_callbacks.py
```diff
@@ -55,3 +55,58 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
         n_complete = len(trials)
         if n_complete >= self._n_trials:
             study.stop()
+
+
+@experimental("2.8.0")
+class RetryFailedTrialCallback:
+    """Retry a failed trial up to a maximum number of times
+
+    When a trail fails, this callback can be used with the :class:`optuna.storage` class to
```
hvy, line 64:
```suggestion
    When a trial fails, this callback can be used with the :class:`optuna.storage` class to
```

optuna/_callbacks.py
```diff
@@ -55,3 +55,58 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+            )
+
+    """
+
+    def __init__(self, max_retry: int) -> None:
+        self._max_retry = max_retry
+
+    def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
+        user_attrs = {"failed_trial": trial.number}
+        user_attrs.update(trial.user_attrs)
```
hvy, line 94: You need to use `set_user_attr` in order to persist the change in the storage. However, I don't think we can update the trial here since it's already been finished.

optuna/_callbacks.py
Crissman: Yes, this is possible, and an improvement. Will change the code accordingly. 🙇

optuna/_callbacks.py
Crissman: This isn't to update the current trial, but to provide a record for the retry of which trial it was a retry for.

optuna/_callbacks.py
Crissman: Done.

Crissman:
@hvy @HideakiImamura Tests passed and comments reflected. PTAL

optuna/_callbacks.py
```diff
@@ -55,3 +55,56 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+        self._max_retry = max_retry
+
+    def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
+        retries = sum(("failed_trial", trial.number) in s.user_attrs.items() for s in study.trials)
+
+        if retries + 1 > self._max_retry:
+            return
+
+        user_attrs = {"failed_trial": trial.number}
+        user_attrs.update(trial.user_attrs)
```
hvy, line 100: I'm somewhat skeptical about using user attributes to communicate with the users. How about at least adding some prefix to avoid shadowing, or maybe better use a system attribute instead? The latter is more robust but require a helper function to determine whether the trial was enqueued or not (if to be made public).

optuna/_callbacks.py
```diff
@@ -55,3 +55,56 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
         n_complete = len(trials)
         if n_complete >= self._n_trials:
             study.stop()
+
+
+@experimental("2.8.0")
+class RetryFailedTrialCallback:
+    """Retry a failed trial up to a maximum number of times
```
hvy, line 62: A huge nit.
```suggestion
    """Retry a failed trial up to a maximum number of times.
```

optuna/_callbacks.py
```diff
@@ -55,3 +55,56 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
         n_complete = len(trials)
         if n_complete >= self._n_trials:
             study.stop()
+
+
+@experimental("2.8.0")
+class RetryFailedTrialCallback:
```
hvy, line 61: For maintainers, it might be a bit confusing having defined this class in `optuna/_callbacks.py`, given its API namespace under `storages`. How about leaving a TODO comment to potentially move the implementation to `storages/` as well?

hvy (commented):
Thanks for the quick fixes! I left some additional comments if you could take a look.

optuna/_callbacks.py
Crissman: Done.

optuna/_callbacks.py
Crissman: Not sure about this one. I thought the idea of having an `optuna/_callbacks.py` file was to put all callbacks there. If this is better put in the `storages/` then I can move it there now. 🤷‍♂️

Crissman:
@hvy Updated. PTAL 🙇

optuna/_callbacks.py
hvy: It's an implementation details so it's not a strong opinion. Actually, `MaxTrialsCallback` is exposed via `optuna.study` and not `optuna` so it's already a bit confusing and maybe we can leave it as is. Was first thinking of something along `storages/_callbacks.py`.

optuna/_callbacks.py
```diff
@@ -55,3 +55,58 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+    def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
+        retries = sum(
+            ("failed_trial", trial.number) in s.system_attrs.items() for s in study.trials
+        )
+
+        if retries + 1 > self._max_retry:
+            return
+
+        system_attrs = {"failed_trial": trial.number}
+        system_attrs.update(trial.system_attrs)
```
hvy, line 102: Is the expectation that this line should update/overwrite the value of the `"failed_trial"` key with the very first trial number if this is not the first retry? Might be good to leave a comment since the logic isn't obvious.

optuna/_callbacks.py
```diff
@@ -55,3 +55,58 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+            )
+
+    """
+
+    def __init__(self, max_retry: int) -> None:
+        self._max_retry = max_retry
+
+    def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
+        retries = sum(
+            ("failed_trial", trial.number) in s.system_attrs.items() for s in study.trials
```
hvy, line 95: Don't we need the original trial number here instead of `trial.number` in case this is not the first retry?

optuna/_callbacks.py
hvy: Thanks! How about also providing a function like `def retried_trial_number(trial: FrozenTrial) -> Optional[int]` as well so that user can resume their objectives if necessary?

Crissman:
@hvy Comments reflected. PTAL

For now, I think it's better to have one `optuna/_callbacks.py` file, as opposed to have a `study/_callbacks.py` and `storage/_callbacks.py` each with one callback in them. Once we have more callbacks, then we can look at distributing.

hvy:
Maybe this comment was forgotten https://github.com/optuna/optuna/pull/2694#discussion_r638527571.

optuna/_callbacks.py
hvy: Looks great, thanks!

optuna/_callbacks.py
hvy: Thanks!

optuna/_callbacks.py
hvy:
```python
def retried_trial_number(trial):
    return trial.system_attr.get("failed_trial", None)

**Comment by Crissman** [comment]
@hvy new method implemented. PTAL

File: tests/storages_tests/rdb_tests/test_storage.py [hunk 9]
```diff
@@ -399,3 +400,34 @@ def failed_trial_callback(study: Study, trial: FrozenTrial) -> None:
             with patch.object(storage, "failed_trial_callback", wraps=failed_trial_callback) as m:
                 study.optimize(lambda _: 1.0, n_trials=1)
                 m.assert_called_once()
+
+
+def test_RetryFailedTrialCallback() -> None:
```
hvy, line 405: A cosmetic but let's stick to the snake case convention.
```suggestion
def test_retry_failed_trial_callback() -> None:
```

optuna/_callbacks.py
```diff
@@ -55,3 +56,78 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+                state=optuna.trial.TrialState.WAITING,
+                params=trial.params,
+                distributions=trial.distributions,
+                user_attrs=trial.user_attrs,
+                system_attrs=system_attrs,
+            )
+        )
+
+    @staticmethod
+    def retried_trial_number(trial: FrozenTrial) -> Optional[int]:
```
hvy, line 121: Should we make this static method experimental, or is a warning already raised?

optuna/_callbacks.py
```diff
@@ -55,3 +56,78 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+        """Return the number of the trail being retried.
```
hvy, line 122: Classic.
```suggestion
        """Return the number of the trial being retried.
```

optuna/_callbacks.py
```diff
@@ -55,3 +56,78 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+
+        Args:
+            trial:
+                The :obj:`trial` object.
```
hvy, line 126: Let's fix the reference (to `FrozenTrial`), or just make it plain text since the type is already annotated.
```suggestion
                The trial object.
```

hvy (changes requested):
Almost LGTM, thanks!

By the way, @ytsmiling, do you want to take a quick look at the provided functionality if it looks okay?

ytsmiling:
Thanks for mentioning. The callback looks great.
Minor comment: It might be useful if `retried_trial_number` is a sequence instead of a single integer, but this might be a matter of preference.

Crissman:
@ytsmiling Yes, it could be done as a sequence, but the intent behind the current interface is to check if the current trial might need special handling as a retry. Of course, the sequence can be found by checking all the trials in `study.trials`.  Let's leave as is for now. 🙇

optuna/_callbacks.py
Crissman: Added. 🙇

hvy:
It might be the case that we cannot put the experimental decorator after `@staticmethod`. https://github.com/optuna/optuna/runs/2690193999?check_suite_focus=true#step:8:152 Does changing their order help?

Crissman:
> It might be the case that we cannot put the experimental decorator after `@staticmethod`. https://github.com/optuna/optuna/runs/2690193999?check_suite_focus=true#step:8:152 Does changing their order help?

Lol. Just figured that out myself and pushed.

hvy:
Right, it might makes sense to keep and expose the entire history of retries. Just also hit me that the retry count is a required argument without a default value, and it does not have an option to retry indefinitely. We could add a default and/or allow `None` as a way to specify infinite retries. But I think those could be follow ups (latter won't break compatibility).

hvy (approved):
Thanks for the long running effort, LGTM!

Crissman:
Detailed review appreciated. 🙏

Crissman:
> ...We could add a default and/or allow `None` as a way to specify infinite retries. But I think those could be follow ups (latter won't break compatibility).

Or we could do that now. ;)

hvy:
Looks good still, or better! Thanks!

Crissman:
Still futzing. Added testing for the unlimited case and refactored the testing to parametrized version. 🙇

Crissman:
@hvy @HideakiImamura Ok, I think it's complete. PTAL

hvy (approved):
LGTM!

optuna/_callbacks.py
```diff
@@ -55,3 +56,84 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+
+@experimental("2.8.0")
+class RetryFailedTrialCallback:
+    """Retry a failed trial up to a maximum number of times.
+
+    When a trial fails, this callback can be used with the :class:`optuna.storage` class to
+    recreate the trial in :obj:`TrialState.WAITING` to queue up the trial to be run again.
+
+    This is helpful in environments where trials may fail due to external conditions, such as
+    being pre-empted by other processes.
```
HideakiImamura, line 69: A nit
```suggestion
    being preempted by other processes.
```

optuna/_callbacks.py
```diff
@@ -55,3 +56,84 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+        system_attrs = {"failed_trial": trial.number}
+
+        # Update the new object with the values in the trial.system_attrs.
+        # By doing this, if this failed try is already a rety, the 'failed_trial' value
+        # will be the first failed trial number.
+        system_attrs.update(trial.system_attrs)
+
+        retries = sum(
+            ("failed_trial", system_attrs["failed_trial"]) in s.system_attrs.items()
+            for s in study.trials
```
HideakiImamura, line 109: A nit, but in Optuna, we use `t` for the trial.
```suggestion
            ("failed_trial", system_attrs["failed_trial"]) in t.system_attrs.items()
            for t in study.trials
```

optuna/_callbacks.py
```diff
@@ -55,3 +56,84 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+                state=optuna.trial.TrialState.WAITING,
+                params=trial.params,
+                distributions=trial.distributions,
+                user_attrs=trial.user_attrs,
+                system_attrs=system_attrs,
+            )
+        )
+
+    @staticmethod
+    @experimental("2.8.0")
```
HideakiImamura, line 126: I think we can remove this experimental decorator since we already decorate the class itself.
```suggestion
```

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -399,3 +400,33 @@ def failed_trial_callback(study: Study, trial: FrozenTrial) -> None:
...
+        time.sleep(grace_period + 1)
+
+        # Exceptions raised in spawned threads are caught by `_TestableThread`.
+        with patch("optuna._optimize.Thread", _TestableThread):
+            study.optimize(lambda _: 1.0, n_trials=1)
+
+        # Test the last trial to see if it was a retry of the first trial or not
+        # Test max_rety=None to see if trial is retried
+        # Test max_rety=0 to see if no trials are retried
+        # Test max_rety=1 to see if trial is retried
```
HideakiImamura, line 429: A nit: period.
```suggestion
        # Test the last trial to see if it was a retry of the first trial or not.
        # Test max_rety=None to see if trial is retried.
        # Test max_rety=0 to see if no trials are retried.
        # Test max_rety=1 to see if trial is retried.
```

HideakiImamura (commented):
Thanks for the update! I have several minor comments. PTAL.

optuna/_callbacks.py
hvy: It was I who suggested this decorator so let me leave a note. The class constructor is decorated already, but this is a static method which technically can be invoked without warnings if this decorator isn't there.

HideakiImamura: Thanks for your comment. I understood. Let me resolve this conversation.

HideakiImamura (approved):
LGTM.

PR #2707, opened 2021-05-29
[RFC] Implement instance methods to return printable representation of pruners

himkt:
Apart from https://github.com/optuna/optuna/pull/2682.

## Motivation
The AllennLP integration restores a pruner using environment variables.
https://github.com/optuna/optuna/blob/bfb41abc4513f7beb73b2c3167d9e5728c56c202/optuna/integration/allennlp.py#L74-L95

However, this approach requires updating the integration module when
developers add a new pruner, which is not unfeasible.
https://github.com/optuna/optuna/blob/bfb41abc4513f7beb73b2c3167d9e5728c56c202/optuna/integration/allennlp.py#L148-L183

So, in this PR, I tried to implement `__repr__` method to each pruner, making a pruner responsible for reproducible object representation. If `__repr__` is introduced, restoring logic can be simplified as shown in d3a7d97.

I understand we have to be careful to introduce such changes, so I'd like other developers' opinions/suggestions.

## Description of the changes
- 18815da
- 67025e0

### DIFF
--- a/optuna/pruners/_base.py
+++ b/optuna/pruners/_base.py
@@ -7,6 +7,10 @@ class BasePruner(object, metaclass=abc.ABCMeta):
     """Base class for pruners."""
 
     @abc.abstractmethod
+    def __repr__(self) -> str:
+        raise NotImplementedError
+
+    @abc.abstractmethod
     def prune(self, study: "optuna.study.Study", trial: "optuna.trial.FrozenTrial") -> bool:
         """Judge whether the trial should be pruned based on the reported values.
 
--- a/optuna/pruners/_base.py
+++ b/optuna/pruners/_base.py
@@ -1,5 +1,7 @@
 import abc
+from typing import Any
 from typing import Dict
+import warnings
 
 import optuna
 
@@ -7,22 +9,10 @@ import optuna
 class BasePruner(object, metaclass=abc.ABCMeta):
     """Base class for pruners."""
 
-    SPECIAL_KEYWORDS: Dict[str, str] = {}
-
     def __repr__(self) -> str:
-        import inspect
-
         parameters = []
-        for keyword in inspect.signature(self.__class__.__init__).parameters:
-            if keyword == "self":
-                continue
-
-            access_keyword = "_{keyword}".format(keyword=keyword)
-            if keyword in self.SPECIAL_KEYWORDS:
-                access_keyword = self.SPECIAL_KEYWORDS[keyword]
-
-            if hasattr(self, access_keyword):
-                parameters.append("{}={}".format(keyword, repr(getattr(self, access_keyword))))
+        for key, value in self._get_init_arguments().items():
+            parameters.append("{}={}".format(key, repr(value)))
 
         return "{class_name}({parameters})".format(
             class_name=self.__class__.__name__,
@@ -49,3 +39,12 @@ class BasePruner(object, metaclass=abc.ABCMeta):
         """
 
         raise NotImplementedError
+
+    def _get_init_arguments(self) -> Dict[str, Any]:
+        warnings.warn(
+            "{} doesn't have a method `_get_init_arguments`."
+            " If you implement a pruner in your own, consider implementing it."
+            " For more information, please refer to Optuna GitHub PR"
+            " (https://github.com/optuna/optuna/pull/2707).".format(self.__class__.__name__)
+        )
+        return {}
--- a/optuna/pruners/_base.py
+++ b/optuna/pruners/_base.py
@@ -1,4 +1,5 @@
 import abc
+from typing import Dict
 
 import optuna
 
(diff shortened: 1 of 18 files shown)


nzw0301:
Let me leave a minor comment. [`inspect`](https://docs.python.org/3.8/library/inspect.html#introspecting-callables-with-the-signature-object) might make the logic simpler.

```python
import inspect

class Cat(object):
     def __init__(self, name, kwarg='default'):
         self.name = name
     def __repr__(self):
        return ",".join(
            [augment for augment in inspect.signature(self.__init__).parameters]
        )

print(Cat("tom"))
```

## output

```
name,kwarg
```

himkt:
Thank you for the awesome tip! I tried to use `inspect.signature` in dc04667.
For parameters that are transformed in `__ini__`, I declare a class variable `SPECIAL_KEYWORDS` to access these original values.

Does this update meet your suggestion? (Thank you for letting me know about the `inspect` module. It seems to be so useful)

Crissman:
@nzw0301 Can you take a look at this?

nzw0301:
@Crissman sure. But I'll review this pull request next week due to some deadlines related to my main job. If it is late, please feel free to unassign me from this PR.

optuna/pruners/_base.py
```diff
@@ -1,11 +1,34 @@
...
 import optuna
 
 
 class BasePruner(object, metaclass=abc.ABCMeta):
     """Base class for pruners."""
 
+    SPECIAL_KEYWORDS: Dict[str, str] = {}
+
+    def __repr__(self) -> str:
+        import inspect
```
nzw0301, line 13: maybe we can move this import line to the top?

optuna/pruners/_base.py
```diff
@@ -1,11 +1,34 @@
...
+
+        parameters = []
+        for keyword in inspect.signature(self.__class__.__init__).parameters:
+            if keyword == "self":
+                continue
```
nzw0301, line 18:
```suggestion
        for keyword in inspect.signature(self.__init__).parameters:
```

nzw0301 (changes requested):
I'll comment more later, but let me leave minor comments

nzw0301:
In my understanding, this PR aims to give a functionality to reconstruct a pruner with arguments from a string representation.
So a retuned value of `__repr__` is fixed at instantiation of the pruner. Thus we can create the retuned value at instantiation of the pruner and store it as an attribute for ``__repr__``. ``__repr__`` returns the created value at instantiation.

What do you think about it?

HideakiImamura:
@nzw0301 Thanks for the insightful comments. I and @himkt discussed privately and agreed to introduce the method like `_get_init_arguments` to pruners. As you suggested, this change will return the arguments at the instantiation and remove `SPECIAL_KEYWORDS`. Please wait @himkt to update the codes.

nzw0301:
@HideakiImamura Thank you for sharing it! It sounds great.

himkt:
@nzw0301 @HideakiImamura I'm sorry for the late response (currently, I'm struggling to find time for watching PRs).

As @HideakiImamura said, I'll implement `_get_init_arguments` to each pruner.
A concern of the change is that user-defined pruners affect the changes since when instantiating a pruner, `pruner.__repr__` would be invoked. For avoiding the error, I'd also implement `_get_init_arguments` to `BasePruner` like below.

```python
class BasePruner(object, metaclass=abc.ABCMeta):
    def _get_init_arguments(self):
        warning.warn("{} doesn't have a method `_get_init_arguments`. If you implement a pruner in your own, consider implementing it. For more information, please refer to Optuna GitHub PR (https://github.com/optuna/optuna/pull/2707).".format(self.__class__.__name__)
        return {}
```

optuna/pruners/_base.py
himkt: d7ceae7 removes the importing of `inspect`!

himkt:
@HideakiImamura @nzw0301 
I revised the PR to use `_get_init_argument`. PTAL.

nzw0301:
Thank you for letting me know the update! 

I should have given an example code for https://github.com/optuna/optuna/pull/2707#issuecomment-853252109. More concretely, I imagined something like this:

```python
import inspect

class CatPruner(object):
    def __init__(self, name, kwarg='default'):
        
        args = []
        for argument in inspect.signature(self.__init__).parameters:

            args.append("{}={}".format(argument, repr(locals()[argument])))
        self._repr_value = "{}({})".format(self.__class__.__name__, ",".join(args))

        self.name = name

        # do something more
        
    def __repr__(self):
        return self._repr_value


print(CatPruner("tom"))
# the output is `CatPruner(name='tom',kwarg='default')`
```

In my understanding, introduced `self._init_*` attributes are only for `__repr__` and they are not supposed to be overwritten. So the returned value of `__repr__` can be stored at instantiation of a pruner.

If I miss something important, please let me know!

himkt:
After discussion with @nzw0301, @HideakiImamura, @c-bata, and @g-votte, we decided to change the approach to fix this problem. We will make another PR to directly fix the AllenNLP integration. Let me close this PR.

nzw0301:
As a side note for https://github.com/optuna/optuna/issues/2536, we might implement `__str__` to give the printable format of a pruner. So, the approach discussed in this PR will be useful for that.

PR #2710, opened 2021-05-30
DOC: Updated command for installing auto-formatters

### LINKED ISSUE #2708: Autoformatting command's parameter change in CONTRIBUTING.md.
Issue with `CONTRIBUTING.md`

While running the command for installing autoformatters `$ pip install .[checking]` , I get the following error:
`no matches found: .[checking]`

When I run the following command:  `$ pip install -e ".[checking]"`, it works correctly.

## Environment
- Optuna version: 2.8.0.dev
- Python version: 3.8.5
- OS: Ubuntu 20.04

01-vyom:
## Motivation

While installing auto-formatters for optuna, I was getting an error the following error: no matches found: .[checking] . So, I added quotes to the command and it worked. The non-quote command works for bash but doesn't work in zsh.
## Description of the changes

Changed the command for installing auto-formatters, added " " to support installation in all shells.

closes #2708

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,7 +71,7 @@ If you catch format errors, you can automatically fix them by auto-formatters.
 
 ```bash
 # Install auto-formatters.
-$ pip install .[checking]
+$ pip install ".[checking]"
 
 $ ./formats.sh
 ```


Crissman (approved):
Thanks for the fix!

PR #2731, opened 2021-06-07
Use `repr` and `eval` to restore pruner parameters in AllenNLP integration

himkt:
The current implementation of AllenNLP integration doesn't handle `None` correctly.
When `None` is given as a parameter, it is saved as a string value of `"None"`.
This causes an error in `AllenNLPPruningCallback`.

In advance of this PR, we explored another approach to save a repr value of pruner and
restore it using eval. But this approach could induce some restriction to future implementation of the pruner.
Then we decided to fix the current implementation instead of changing the way of saving/restoring pruners.
For a detailed discussion, please see https://github.com/optuna/optuna/pull/2707.

## Motivation
Resolve https://github.com/himkt/allennlp-optuna/issues/39.

## Description of the changes
- a06cfae removes `infer_and_cast` and add `repr` and `eval` to restore pruner parameters

### DIFF
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -399,7 +399,8 @@ class AllenNLPPruningCallback(TrainerCallback):
         loads them to restore ``trial`` and ``monitor``.
 
     .. note::
-        Currently, build-in pruners are supported except for :class:`~optuna.pruners.PatientPruner`.
+        Currently, build-in pruners are supported except for
+        :class:`~optuna.pruners.PatientPruner`.
 
     Args:
         trial:
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -398,6 +398,9 @@ class AllenNLPPruningCallback(TrainerCallback):
         Then :class:`~optuna.integration.AllenNLPPruningCallback`
         loads them to restore ``trial`` and ``monitor``.
 
+    .. note::
+        Currently, build-in pruners are supported except to `~optuna.pruners.PatientPruner`.
+
     Args:
         trial:
             A :class:`~optuna.trial.Trial` corresponding to the current evaluation of the
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -399,7 +399,7 @@ class AllenNLPPruningCallback(TrainerCallback):
         loads them to restore ``trial`` and ``monitor``.
 
     .. note::
-        Currently, build-in pruners are supported except to `~optuna.pruners.PatientPruner`.
+        Currently, build-in pruners are supported except for :class:`~optuna.pruners.PatientPruner`.
 
     Args:
         trial:
--- a/optuna/integration/allennlp.py
+++ b/optuna/integration/allennlp.py
@@ -116,7 +116,8 @@ def _get_environment_variables_for_pruner() -> Dict[str, Optional[Union[str, int
     for key in keys.split(","):
         key_without_prefix = key.replace("{}_".format(_PREFIX), "")
         value = os.getenv(key)
-        assert value is not None
+        if value is None:
+            raise ValueError("{key} is not found in environment variables.")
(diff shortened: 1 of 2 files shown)


optuna/integration/allennlp.py
```diff
@@ -141,7 +115,9 @@ def _get_environment_variables_for_pruner() -> Dict[str, Optional[Union[str, int
     kwargs = {}
     for key in keys.split(","):
         key_without_prefix = key.replace("{}_".format(_PREFIX), "")
-        kwargs[key_without_prefix] = _infer_and_cast(os.getenv(key))
+        value = os.getenv(key)
+        assert value is not None
+        kwargs[key_without_prefix] = eval(value)
```
himkt, line 120: Of course, this implementation could be problematic when some new pruner having a parameter that is hard to restore using `eval`. However, the problems would only occur in this integration and we'll be exploring other solutions when we are in need.

optuna/integration/allennlp.py
```diff
@@ -141,7 +115,9 @@ def _get_environment_variables_for_pruner() -> Dict[str, Optional[Union[str, int
     kwargs = {}
     for key in keys.split(","):
         key_without_prefix = key.replace("{}_".format(_PREFIX), "")
-        kwargs[key_without_prefix] = _infer_and_cast(os.getenv(key))
+        value = os.getenv(key)
+        assert value is not None
```
nzw0301, line 119: How about using `ValueError` instead of `assert`? `assert` can be ignored by using `-O`.

himkt: Thanks. I adopted your suggestion in d4148c5.

hvy:
@nzw0301 , @HideakiImamura could you continue with the review here? Please feel free to mention other maintainers  as needed.

optuna/integration/allennlp.py
```diff
@@ -141,7 +115,10 @@ def _get_environment_variables_for_pruner() -> Dict[str, Optional[Union[str, int
     kwargs = {}
     for key in keys.split(","):
         key_without_prefix = key.replace("{}_".format(_PREFIX), "")
-        kwargs[key_without_prefix] = _infer_and_cast(os.getenv(key))
+        value = os.getenv(key)
+        if value is None:
+            raise ValueError("{key} is not found in environment variables.")
```
nzw0301, line 120: Maybe? 

```suggestion
            raise ValueError(f"{key} is not found in environment variables.")
```

nzw0301 (changes requested):

nzw0301:
@himkt Let me ask one question -- can we restore the `wrapped_pruner` of [`PatintPruner`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.pruners.PatientPruner.html#optuna.pruners.PatientPruner)?

himkt:
@nzw0301 
No. This PR only solves the existing bug and PatientPruner is not supported in AllenNLP integration for now.
(Later, I'd like to add support for it)

nzw0301:
@himkt thank you for your answer. So could you add a `TODO` or note section for it?

optuna/integration/allennlp.py
```diff
@@ -398,6 +398,9 @@ class AllenNLPPruningCallback(TrainerCallback):
         Then :class:`~optuna.integration.AllenNLPPruningCallback`
         loads them to restore ``trial`` and ``monitor``.
 
+    .. note::
+        Currently, build-in pruners are supported except to `~optuna.pruners.PatientPruner`.
```
nzw0301, line 402:
```suggestion
        Currently, build-in pruners are supported except for :class:`~optuna.pruners.PatientPruner`.
```

nzw0301 (changes requested):
Minor things

nzw0301 (approved):
👍

HideakiImamura (approved):
Thanks for the PR. Looks solid and I agree with the idea of supporting as many pruners as possible with as little implementation cost as possible. LGTM!

PR #2752, opened 2021-06-18
better logging module

SCUTJcfeng:
## Motivation
I want to disable the logging handler in optuna, so I took a closer look at the logging module and found there is a threading lock which looks weird, so I make this change to remove the lock.

## Description of the changes
- no function added or removed 
- _lock is removed and _logger is added

### DIFF
--- a/optuna/logging.py
+++ b/optuna/logging.py
@@ -6,14 +6,9 @@ from logging import FATAL  # NOQA
 from logging import INFO  # NOQA
 from logging import WARN  # NOQA
 from logging import WARNING  # NOQA
-import threading
 from typing import Optional
 
 import colorlog
-
-
-_lock: threading.Lock = threading.Lock()
-_default_handler: Optional[logging.Handler] = None
 
 
 def create_default_formatter() -> colorlog.ColoredFormatter:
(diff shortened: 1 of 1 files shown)


HideakiImamura:

SCUTJcfeng:
The change still break the tests, so I need to fix it.

hvy:
Thanks for your work! I haven't however understood the motivation. Could you elaborate a bit on the weirdness? The locking is needed in my understanding since the logger is (still) being modified (with the handlers, etc.) in a non-atomic way, possibly by different threads. 

As for the failing test, I'm not too familiar with how pytest does the capturing but it's at least caused by it. With pytest, the default stream `sys.stderr` of the `StreamHandler` is altered and hence the previous "resetting" logic that restored the stream to the stderr by re-instantiating the handler was needed. Since you're reusing the same handler, the test fails.

SCUTJcfeng:
In Python, logging module is thread-safe, so `addHandler` or  `removeHandler` or `setLevel ` is also thread-safe.

You are right about failing tests, so I add `_reset_default_handler`

HideakiImamura:
Thanks for the PR! The current implementation has a lock. Could you explain the motivation for this change in more detail?

SCUTJcfeng:
modify _default_handler is not thread-safe, logging functions are thread-safe

SCUTJcfeng:
I think last commit already fix flake8 H404/H405

Issue #2207, opened 2021-01-14
plot_param_importances is portraying parameter importances different from get_param_importances


XyDrKRulof:
(My first issue on an open source GitHub repository)
When I am using the function _plot_param_importances(study)_ the plot is displaying different values than the ordered dict received by the function _get_param_importances(study)_.

## Expected behavior

I would expect the function _plot_param_importances_ to display values close to those of _get_param_importances_.

## Environment

- Optuna version: 2.4.0
- Python version: 3.6.9
- OS: Ubuntu 18.04

## Error messages, stack traces, or logs

![importance_ddpg](https://user-images.githubusercontent.com/65008699/104577124-661a2680-5659-11eb-8252-fae6d9318756.png)
![Selection_048](https://user-images.githubusercontent.com/65008699/104577139-6a464400-5659-11eb-92fd-f00c5d598153.png)

![importance_td3](https://user-images.githubusercontent.com/65008699/104577161-703c2500-5659-11eb-9719-816dfbe95688.png)
![image](https://user-images.githubusercontent.com/65008699/104577748-2142bf80-565a-11eb-8137-e2836582115d.png)

## Steps to reproduce

1. Pretty Print optuna.importance.get_param_importances(study)
2. Plot optuna.visualization.plot_param_importances(study)
3. Compare

## Reproducible examples

```python
importance_dict = optuna.importance.get_param_importances(study)
pprint.pprint(importance_dict)
fig = optuna.visualization.plot_param_importances(study)
fig.write_image(f"images/importances.png")
```

XyDrKRulof:
Okay this is not an issue, this is expected behavior. The way optuna calculates the importances is through an random forest:

https://optuna.readthedocs.io/en/latest/reference/generated/optuna.importance.FanovaImportanceEvaluator.html?highlight=FanovaImportanceEvaluator

That means the results are not deterministic and can differ at each call.

hvy:
> That means the results are not deterministic and can differ at each call.

Exactly. If you really want to both visualize and get the dictionary with identical results, you'll have to create two evaluators you referred to above with the same seeds.

PR #2756, opened 2021-06-22
Hyperparameter importance made deterministic when no evaluator is specified

hvy:
## Motivation

Hyperparameter importance evaluations are non-deterministic by default (when no evaluator is specified). This may be confusing, for instance when you want to obtain the importances as a dictionary using `optuna.importance.get_param_importances(study)`, then visualize them via `optuna.visualization.plot_param_importances(study)`, just to observe that the computed importances are different.

## Description of the changes

Fixes the seed of the importance evaluator that is constructed when no evaluator is explicitly specified by the caller (default behavior).

## References

This has been raised as an issue before, e.g. https://github.com/optuna/optuna/issues/2207.

### DIFF
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -85,7 +85,7 @@ def get_param_importances(
             optimization.
     """
     if evaluator is None:
-        evaluator = FanovaImportanceEvaluator()
+        evaluator = FanovaImportanceEvaluator(seed=0)
 
     if not isinstance(evaluator, BaseImportanceEvaluator):
         raise TypeError("Evaluator must be a subclass of BaseImportanceEvaluator.")
--- a/tests/importance_tests/test_init.py
+++ b/tests/importance_tests/test_init.py
@@ -279,3 +279,16 @@ def test_get_param_importances_invalid_params_type(
 
     with pytest.raises(TypeError):
         get_param_importances(study, evaluator=evaluator_init_func(), params=[0])  # type: ignore
+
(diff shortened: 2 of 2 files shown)


g-votte:
Thanks for your PR. Being honest, I have concerns with this change.

FANOVA is random in nature because it uses Random Forest. If we fixed the random seed, we'd discard the random aspect of the algorithm.

hvy:
That's a valid point. We can alternatively try to work on the documentation, that the default algorithm is stochastic by nature and explain how the seed can be fixed in such cases. These pages might be relevant.

- https://optuna.readthedocs.io/en/v2.8.0/reference/importance.html
- https://optuna.readthedocs.io/en/v2.8.0/reference/generated/optuna.importance.FanovaImportanceEvaluator.html#optuna.importance.FanovaImportanceEvaluator
- https://optuna.readthedocs.io/en/v2.8.0/reference/generated/optuna.importance.MeanDecreaseImpurityImportanceEvaluator.html#optuna.importance.MeanDecreaseImpurityImportanceEvaluator
- https://optuna.readthedocs.io/en/v2.8.0/reference/visualization/generated/optuna.visualization.plot_param_importances.html#optuna.visualization.plot_param_importances

g-votte:
Cool! I agree that the stochastic behavior confuses some users, and prefer describing it in the document.

hvy:
Let me close this issue. I now think it's better to keep the current behavior.

PR #2759, opened 2021-06-25
[RFC] Multiprocess in-memory storage

hvy:
## Motivation

The `n_jobs` parameter in `Study.optimize` has been deprecated in favor of distributed optimization via different Python interpreter processes. However, there's a demand for `n_jobs`, and in addition, an in memory storage that work with both multithreading and multiprocessing.

## Description of the changes

- Reverts the deprecation of `n_jobs`
- Introduces a parameter to `Study.optimize` to switch between multithreading and multiprocessing
- Introduces a naive new multiprocess in memory storage

_These last change is independent of the other two and could be split._

Some notes on the current changes.

- Study is serialized to forked processes in each trial when `scheduler="processes"`. Objects belonging to a study, that is not managed by the multiprocessing manager such as the sampler/pruner is copied in each trial and cannot utilize cache/state
- Performance optimizations of the multirprocess in memory storage i.e. caching is left outside the scope of this PR
- Explicit serialization of objective functions is left outside the scope of this PR

## TODO

- [x] Implement
- [ ] Test

## Example usage

```python
import optuna as ot


# Note that the objective function must be serializable.
# Serialization of lambda objective functions is left outside the scope of this PR.
def obj(t):
    return t.suggest_float("x", 0, 1)


if __name__ == "__main__":
    study = ot.create_study(storage="multiprocess", study_name="foo")
    study.optimize(obj, n_trials=10, n_jobs=3, scheduler="threads")
    study.optimize(obj, n_trials=10, n_jobs=3, scheduler="processes")
    assert len(study.trials) == 20
```

### DIFF
--- a/optuna/storages/_in_memory.py
+++ b/optuna/storages/_in_memory.py
@@ -484,6 +484,9 @@ _Manager.register("InMemoryStorage", InMemoryStorage)
 class MultiprocessInMemoryStorage(BaseStorage):
     def __init__(self) -> None:
         self._manager = _Manager()
+        # A manager process is started at object construction. The started manager process is
+        # automatically shut down by the `multiprocessing` library, so explicit shutdown
+        # (`_Manager.shutdown`) is omitted.
         self._manager.start()
         self._manager_address = self._manager.address
         self._storage = self._manager.InMemoryStorage()  # type: ignore
--- a/optuna/storages/_in_memory.py
+++ b/optuna/storages/_in_memory.py
@@ -482,8 +482,6 @@ class _Manager(BaseManager):
 _Manager.register("InMemoryStorage", InMemoryStorage)
 
 
-# TODO(hvy): Either support multiple studies, either here or in the `InMemoryStorage`, or
-# hide `MultiprocessInMemoryStorage` from the user entirely.
 class MultiprocessInMemoryStorage(BaseStorage):
     def __init__(self) -> None:
         self._manager = _Manager()
--- a/optuna/storages/_in_memory.py
+++ b/optuna/storages/_in_memory.py
@@ -1,5 +1,7 @@
 import copy
 from datetime import datetime
+from multiprocessing.managers import BaseManager
+import os
 import threading
 from typing import Any
 from typing import cast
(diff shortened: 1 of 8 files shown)


optuna/storages/_in_memory.py
```diff
@@ -471,3 +472,106 @@ def __init__(self, name: str) -> None:
...
+
+
+class _Manager(BaseManager):
+    pass
+
+
+_Manager.register("InMemoryStorage", InMemoryStorage)
+
+
+class MultiprocessInMemoryStorage(BaseStorage):
```
sile, line 484: This is a very simple implementation 👍

optuna/storages/_in_memory.py
```diff
@@ -471,3 +472,106 @@ def __init__(self, name: str) -> None:
...
+    def __init__(self) -> None:
+        self._manager = _Manager()
+        self._manager.start()
```
sile, line 487: [QUESTION] Don't we need to explicitly call the `BaseManager.shutdown()` method to stop the server when the storage instance is dropped?

sile (commented):
Although this is still WIP, I took a quick look at this PR. The design seems good.

optuna/storages/_in_memory.py
hvy: That's a good point. I thought so too and actually implemented a finalizer in an earlier commit, but later reverted it https://github.com/optuna/optuna/pull/2759/commits/b59af0bb0e33f14ee39537ef21f945f40e12ebab since it was already handled by the library (but it's allowed to call `shutdown` multiple times).

> Manager processes will be shutdown as soon as they are garbage collected or their parent process exits. The manager classes are defined in the multiprocessing.managers module:

https://docs.python.org/3.6/library/multiprocessing.html

I think it's enough, but do you think we need to implement our own logic?

sile: Got it! Thanks for your explanation.
I also think that the current implementation is enough, but how about leaving a short comment about the behavior for future readers like me?

hvy: Sounds good, I added a comment here https://github.com/optuna/optuna/pull/2759/commits/47012a4d754f8d7c26821d85d8fcfc4239da1762.

sile: 👍

optuna/storages/_redis.py
```diff
@@ -73,6 +73,15 @@ def __init__(self, url: str) -> None:
...
         self._redis = redis.Redis.from_url(url)
 
+    def __getstate__(self) -> Dict[Any, Any]:
+        state = self.__dict__.copy()
+        del state["_redis"]
+        return state
+
+    def __setstate__(self, state: Dict[Any, Any]) -> None:
+        self.__dict__.update(state)
+        self._redis = redis.Redis.from_url(self._url)
```
hvy, line 83: This was to support `scheduler="processes"` with the Redis storage but I'm feeling that this multiprocessing scheduling requires more time and tests, at least for Redis. We can separate this PR into two, to the scheduler and multiprocess in memory storage. We could also leave the Redis storage outside the scope.

optuna/study/study.py
```diff
@@ -384,17 +381,25 @@ def objective(trial):
...
                 when ``n_jobs`` :math:`\\ne 1`.
+            scheduler:
+                Scheduler to determine how trials are executed in parallel when ``n_jobs`` is
+                larger than one. Can be either ``threads`` or ``processes`` for multithreading or
+                multiprocess, respectively. Ignored if ``n_jobs`` is one.
+
+                .. note::
+                    Added in v2.9.0 as an experimental feature. The interface may change in newer
+                    versions without prior notice. See
+                    https://github.com/optuna/optuna/releases/tag/v2.9.0.
```
c-bata, line 392: I am wondering if Optuna should provide such a feature.

I think that this feature is too much just for multiprocessing support. If you are planning to support another schedulers (e.g. a scheduler based on dask-distributed and/or Kubernetes jobs), I think we can use Joblib instead as in previous versions.

c-bata: One alternative approach is removing this option and guides users to use `ProcessPoolExecutor` in a documentation and a release note (like https://github.com/optuna/optuna/releases/tag/v2.7.0).

```python
storage = storages.MultiprocessInMemoryStorage()

with ProcessPoolExecutor(max_workers=5) as executor:
    for _ in range(5):
        executor.submit(study.optimize, objective, 100)
```

optuna/storages/_in_memory.py
c-bata: [nits] This is not a strong opinion, but `IPCStorage` or `InterProcessStorage` might be better.

Crissman:
@sile and @c-bata are you ok to continue looking at this PR?

sile:
Sure!

PR #2780, opened 2021-07-03
Redis Heartbeat

### LINKED ISSUE #2773: Redis Heartbeart
Add heartbeat support to the Redis storage backend similar to the RDB storage.

## Motivation

While the RDB storage backend supports the heartbeat functionality, the Redis storage does not. This prevents the Redis backend from being a drop-in replacement for the RDB backend.

## Description

Implement the methods of `BaseStorage` in `RedisStorage`.

## Alternatives

Move the heartbeat functionality out of the individual storage classes into `BaseStorage`, e.g. via trial system attributes. Although this provides some simplifications, I would not recommend this because deleting stale trials can be implemented more efficiently by the backend.

## Additional context

I would like to contribute this feature, if you think it fits into the library.

Turakar:
## Motivation
Fix #2773. Sorry for not waiting for your replies there, but I could not wait to start coding :smile: 

## Description of the changes
This PR adds heartbeat support to the Redis storage backend and unifies some of the heartbeat related tests for RDB and Redis backends. Detailed list of changes:

- Heartbeats are saved in a hash (i.e. dict) per study and not per trial. While this requires an additional lookup of the study ID in `set_trial_state()` and `record_heartbeat()`, it significantly improves the performance of `fail_stale_trials()` for studies with many non-running trials.
- Heartbeats are removed when a trial's state is updated on completion. As empty hashes are deleted in Redis automatically, there should be no permanent storage usage after finishing all running trials. However, a user can call `record_heartbeat` out of order, but as he/she is not supposed to call backend methods anyway and the RDB storage has the same problem, I ignored this. After all, looking up the trial state in `record_heartbeat()` would cause a performance cost.
- I moved the tests from the RDB storage tests to the general storage tests and added a new list `STORAGE_MODES_HEARTBEAT`. Additionally, `StorageSupplier` now supports passing `kwargs` to the backends to unify the setting of `heartbeat_interval`, `grace_period` and `failed_trial_callback`. In this process, I lost one single test which is redundant in my opinion, i.e. checking the return value of `_get_stale_trials()` in `test_fail_stale_trials()`.

### DIFF
--- a/optuna/storages/_redis.py
+++ b/optuna/storages/_redis.py
@@ -1,7 +1,6 @@
 import copy
 from datetime import datetime
 import pickle
-import time
 from typing import Any
 from typing import Callable
 from typing import cast
@@ -647,7 +646,9 @@ class RedisStorage(BaseStorage):
     def record_heartbeat(self, trial_id: int) -> None:
         study_id = self.get_study_id_from_trial_id(trial_id)
         self._redis.hset(
-            self._key_study_heartbeats(study_id), str(trial_id), pickle.dumps(time.time())
+            self._key_study_heartbeats(study_id),
+            str(trial_id),
+            pickle.dumps(self._get_redis_time()),
         )
 
     def fail_stale_trials(self, study_id: int) -> List[int]:
@@ -666,7 +667,7 @@ class RedisStorage(BaseStorage):
         else:
             grace_period = self._grace_period
 
-        current_time = time.time()
+        current_time = self._get_redis_time()
         heartbeats = self._redis.hgetall(self._key_study_heartbeats(study_id))
         stale = []
         for trial_id_raw, last_heartbeat_raw in heartbeats.items():
@@ -675,6 +676,10 @@ class RedisStorage(BaseStorage):
                 trial_id = int(trial_id_raw)
                 stale.append(trial_id)
         return stale
+
+    def _get_redis_time(self) -> float:
+        seconds, microseconds = self._redis.time()
+        return seconds + microseconds * 1e-6
 
     def _is_heartbeat_supported(self) -> bool:
         return True
--- a/optuna/storages/_redis.py
+++ b/optuna/storages/_redis.py
@@ -1,7 +1,9 @@
 import copy
 from datetime import datetime
 import pickle
+import time
 from typing import Any
+from typing import Callable
 from typing import cast
 from typing import Dict
 from typing import List
@@ -59,19 +61,50 @@ class RedisStorage(BaseStorage):
 
     Args:
         url: URL of the redis storage, password and db are optional. (ie: redis://localhost:6379)
+        heartbeat_interval:
+            Interval to record the heartbeat. It is recorded every ``interval`` seconds.
+        grace_period:
+            Grace period before a running trial is failed from the last heartbeat.
+            If it is :obj:`None`, the grace period will be `2 * heartbeat_interval`.
+        failed_trial_callback:
+            A callback function that is invoked after failing each stale trial.
+            The function must accept two parameters with the following types in this order:
+            :class:`~optuna.study.Study` and :class:`~optuna.FrozenTrial`.
+
+            .. note::
+                The procedure to fail existing stale trials is called just before asking the
+                study for a new trial.
 
     .. note::
         If you use plan to use Redis as a storage mechanism for optuna,
         make sure Redis in installed and running.
         Please execute ``$ pip install -U redis`` to install redis python library.
+
+    Raises:
+        :exc:`ValueError`:
+            If the given `heartbeat_interval` or `grace_period` is not a positive integer.
     """
 
-    def __init__(self, url: str) -> None:
+    def __init__(
+        self,
+        url: str,
+        heartbeat_interval: Optional[int] = None,
+        grace_period: Optional[int] = None,
+        failed_trial_callback: Optional[Callable[["optuna.Study", FrozenTrial], None]] = None,
+    ) -> None:
 
         _imports.check()
 
+        if heartbeat_interval is not None and heartbeat_interval <= 0:
+            raise ValueError("The value of `heartbeat_interval` should be a positive integer.")
+        if grace_period is not None and grace_period <= 0:
+            raise ValueError("The value of `grace_period` should be a positive integer.")
+
         self._url = url
         self._redis = redis.Redis.from_url(url)
+        self._heartbeat_interval = heartbeat_interval
+        self._grace_period = grace_period
+        self._failed_trial_callback = failed_trial_callback
 
     def create_new_study(self, study_name: Optional[str] = None) -> int:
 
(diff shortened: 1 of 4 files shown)


Turakar:
I am not sure about the CI failures, but I think it is not my fault, right?

Crissman:
@keisuke-umezawa Can you take a look at this?

not522 (commented):
Sorry for the late response. I left some initial comments. I'll take another look later.

not522 (commented):
Basically LGTM. Could you check my comments?

not522:
Could you merge master to fix the CI failure?

optuna/storages/_redis.py
```diff
@@ -365,6 +399,9 @@ def set_trial_state(self, trial_id: int, state: TrialState) -> bool:
             trial.datetime_complete = datetime.now()
             self._redis.set(self._key_trial(trial_id), pickle.dumps(trial))
             self._update_cache(trial_id)
+
+            study_id = self.get_study_id_from_trial_id(trial_id)
+            self._redis.hdel(self._key_study_heartbeats(study_id), str(trial_id))
```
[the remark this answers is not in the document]
Turakar: Yes, but it I wanted to make it explicit as I have to do the decoding manually. So while the Redis library automatically converts it to a string, it does not automatically convert it back. But this a matter of personal preference, so I will adjust to your wishes if you want.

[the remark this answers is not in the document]
Turakar: I just checked: Not doing the explicit conversion raises mypy errors.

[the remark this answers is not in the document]
not522: That makes sense. Let's keep explicit conversion.

not522 (approved):
Thank you! LGTM.

optuna/storages/_redis.py
```diff
@@ -59,19 +61,50 @@ def objective(trial):
...
         url: URL of the redis storage, password and db are optional. (ie: redis://localhost:6379)
+        heartbeat_interval:
+            Interval to record the heartbeat. It is recorded every ``interval`` seconds.
+        grace_period:
+            Grace period before a running trial is failed from the last heartbeat.
+            If it is :obj:`None`, the grace period will be `2 * heartbeat_interval`.
+        failed_trial_callback:
+            A callback function that is invoked after failing each stale trial.
+            The function must accept two parameters with the following types in this order:
+            :class:`~optuna.study.Study` and :class:`~optuna.FrozenTrial`.
```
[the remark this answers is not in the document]
keisuke-umezawa: Why did not merge the suggestion?

optuna/storages/_redis.py
```diff
@@ -607,3 +643,48 @@ def _check_trial_id(self, trial_id: int) -> None:
...
+
+    def _get_stale_trial_ids(self, study_id: int) -> List[int]:
+        assert self._heartbeat_interval is not None
+
+        if self._grace_period is None:
+            grace_period = 2 * self._heartbeat_interval
+        else:
+            grace_period = self._grace_period
+
+        current_time = time.time()
```
keisuke-umezawa, line 669: Can you use the current time of DB side, not client side?

optuna/storages/_redis.py
```diff
@@ -607,3 +643,48 @@ def _check_trial_id(self, trial_id: int) -> None:
 
         if not self._redis.exists(self._key_trial(trial_id)):
             raise KeyError("study_id {} does not exist.".format(trial_id))
+
+    def record_heartbeat(self, trial_id: int) -> None:
+        study_id = self.get_study_id_from_trial_id(trial_id)
+        self._redis.hset(
+            self._key_study_heartbeats(study_id), str(trial_id), pickle.dumps(time.time())
```
keisuke-umezawa, line 650: Can you use the current time of DB side, not client side? c.f. https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/storage.py#L1197

keisuke-umezawa (changes requested):
The current code uses the time of client. But, in this case, all the clients need to use the same clock. [The RDB](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/storage.py) implementation uses DB side clock. In this case, we do not use aligned clocks in the client side.

Could you try to use the DB side clock?

optuna/storages/_redis.py
Turakar: Thanks, I was not aware of this problem. Good catch!

optuna/storages/_redis.py
```diff
@@ -59,19 +61,50 @@ def objective(trial):
...
+    Raises:
+        :exc:`ValueError`:
+            If the given `heartbeat_interval` or `grace_period` is not a positive integer.
     """
 
-    def __init__(self, url: str) -> None:
+    def __init__(
+        self,
+        url: str,
+        heartbeat_interval: Optional[int] = None,
```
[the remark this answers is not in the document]
keisuke-umezawa: @Turakar Sorry for again. But do you have a reason you did not merge it?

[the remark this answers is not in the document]
Turakar: Oh, I think I lost this commit during a force push after the rebasing. That's my bad, sorry.

keisuke-umezawa (approved):

keisuke-umezawa:
LGTM! Thank you for your contribution!
In my local env, it also worked.

optuna/storages/_redis.py
```diff
@@ -365,6 +398,9 @@ def set_trial_state(self, trial_id: int, state: TrialState) -> bool:
             trial.datetime_complete = datetime.now()
             self._redis.set(self._key_trial(trial_id), pickle.dumps(trial))
             self._update_cache(trial_id)
+
```
keisuke-umezawa, line 401: Sorry, after adding LGTM, but it should be better to add:

```
if self.is_heartbeat_supported():
```

Turakar: I think you mean `is_heartbeat_enabled()`, right?

Turakar: Actually, I do not completely agree with your point. Suppose the following scenario:

1. A study is started with heartbeat enabled.
2. A trial is started.
3. The study aborts unexpectedly.
4. The study is reloaded from the storage manually on the console and the running trial is set to FAIL manually.

With the if-statement, the heartbeat would not be deleted, and as such the `_get_stale_trial_ids()` would report the trial, although it is already i state FAIL. I.e. with the if-statement you can go to some inconsistent state.

keisuke-umezawa: @Turakar Thank you. I understand.
In that case, could you add the following comment?
```
# To ensure that there are no failed trials with heartbeats in the DB under any circumstances
```

After that, I will merge it.

keisuke-umezawa (approved):

optuna/storages/_redis.py
Turakar: Added :slightly_smiling_face:

Turakar:
Is it ready to merge or should I do something else?

keisuke-umezawa:
@Turakar Sorry for being late! Meged it.

PR #2781, opened 2021-07-03
Introduce Weights & Biases integration

### LINKED ISSUE #2553: Weights and biases integration
NOTE: I haven't used any management tools for ML; so it is an idea to enlarge optuna community.

## Motivation

[weights and biases](https://wandb.ai/site) (W&B) is a tool for tracking machine learning experiments, visualize metrics, and share results like MLflow. As a counterpart of MLflow, it might be great to support W&B integration for W&B users.

## Description

Like optuna's mlflow integration, optuna might provide a callback for W&B and its example.

[A blog post to use optuna and W&B](https://www.h4pz.co/blog/2020/10/3/optuna-and-wandb)

xadrianzetx:
This PR introduces basic integration to Weights & Biases UI in a form of callback passed to `study.optimize()`.

This callback enables tracking of Optuna study in Weights & Biases. The study is tracked as a single experiment run, where all suggested hyperparameters and optimized metric are logged and plotted as a function of optimizer steps.

## Motivation
This PR resolves #2553, where proposition was initially raised.

## Description of the changes
* Implemented `WeightsAndBiasesCallback`.
* Add documentation and example for new callback.

## Example

Experiment from blog post mentioned in #2553 is now reduced to the following

```python
# [...]

black_box = Objective(seed=4444)
sampler = optuna.samplers.TPESampler(seed=4444)

wandb_kwargs = {
    "project": "optuna",
    "group": "summary",
    "job_type": "logging",
}

wandbc = WeightsAndBiasesCallback(
    metric_name="mse",
    wandb_kwargs=wandb_kwargs
)

study = optuna.create_study(direction="minimize", sampler=sampler)
study.optimize(black_box, n_trials=100, callbacks=[wandbc])

# [...]
```

and results in following visuals in Weights & Biases UI

image

image

### DIFF
--- a/optuna/integration/wandb.py
+++ b/optuna/integration/wandb.py
@@ -39,9 +39,7 @@ class WeightsAndBiasesCallback(object):
                 return (x - 2) ** 2
 
 
-            n_trials = 10
             wandbc = WeightsAndBiasesCallback(
-                n_trials=n_trials,
                 project_name="my-project-name",
                 group_name="my-group-name",
                 job_type="my-job-type",
@@ -49,11 +47,9 @@ class WeightsAndBiasesCallback(object):
 
 
             study = optuna.create_study(study_name="my_study")
-            study.optimize(objective, n_trials=n_trials, callbacks=[wandbc])
+            study.optimize(objective, n_trials=10, callbacks=[wandbc])
 
     Args:
-        n_trials:
-            The number of optimizer trials.
         metric_name:
             Name of the optimized metric. Since the metric itself is just a number,
             `metric_name` can be used to give it a name. So you know later
@@ -78,7 +74,6 @@ class WeightsAndBiasesCallback(object):
 
     def __init__(
         self,
-        n_trials: int,
         metric_name: Union[str] = "value",
         project_name: Optional[str] = None,
         group_name: Optional[str] = None,
@@ -89,7 +84,6 @@ class WeightsAndBiasesCallback(object):
 
         _imports.check()
 
-        self._n_trials = n_trials
         self._metric_name = metric_name
         self._project_name = project_name
         self._group_name = group_name
@@ -97,41 +91,17 @@ class WeightsAndBiasesCallback(object):
         self._run_name = run_name
         self._job_type = job_type
 
-    def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
+        self._initialize_run()
 
-        if trial.number == 0:
-            self._initialize_run(study, trial)
+    def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
 
         for key, value in trial.params.items():
             wandb.log({key: value}, step=trial.number)
 
         wandb.log({self._metric_name: trial.value}, step=trial.number)
 
-        # We need to know when study ends to allow
-        # Weights & Biases to finish run and sync up.
-        # This means that this callback can't be currently used
-        # for time-constrained or non-constrained studies.
-        if trial.number + 1 == self._n_trials:
-            datetime_complete = str(trial.datetime_complete)
-            wandb.config.update({"datetime_complete": datetime_complete})
-            wandb.finish()
-
-    def _initialize_run(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
-        """Initializes Weights & Biases run.
-
-        Args:
-            trial:
-                First trial in the study.
-            study:
-                Tracked study.
-        """
-
-        study_direction = str(study.direction).split(".")[-1]
-        config = {
-            "direction": study_direction,
-            "n_trials": self._n_trials,
-            "datetime_start": str(trial.datetime_start),
-        }
+    def _initialize_run(self) -> None:
+        """Initializes Weights & Biases run."""
 
         wandb.init(
             project=self._project_name,
@@ -139,6 +109,5 @@ class WeightsAndBiasesCallback(object):
             entity=self._entity_name,
             name=self._run_name,
             job_type=self._job_type,
-            config=config,
             reinit=True,
         )
--- a/optuna/integration/wandb.py
+++ b/optuna/integration/wandb.py
@@ -1,0 +1,142 @@
+from typing import Optional
+from typing import Union
+
+import optuna
+from optuna._imports import try_import
+
+
+with try_import() as _imports:
+    import wandb
+
+
+class WeightsAndBiasesCallback(object):
+    """Callback to track Optuna trials with Weights & Biases.
+
+    This callback enables tracking of Optuna study in
+    Weights & Biases. The study is tracked as a single experiment
+    run, where all suggested hyperparameters and optimized metric
+    are logged and plotted as a function of optimizer steps.
+
+    .. note::
+        User needs to be logged in to Weights & Biases before
+        using this callback. For more information, please
+        refer to `wandb setup <https://docs.wandb.ai/quickstart#1-set-up-wandb>`_.
+
+    Example:
+
+        Add Weights & Biases callback to Optuna optimization.
+
+        .. code::
+
+            import optuna
+            from optuna.integration.wandb import WeightsAndBiasesCallback
+
+
+            def objective(trial):
+                x = trial.suggest_float("x", -10, 10)
+                return (x - 2) ** 2
+
+
+            n_trials = 10
+            wandbc = WeightsAndBiasesCallback(
+                n_trials=n_trials,
+                project_name="my-project-name",
+                group_name="my-group-name",
+                job_type="my-job-type",
+            )
+
+
+            study = optuna.create_study(study_name="my_study")
+            study.optimize(objective, n_trials=n_trials, callbacks=[wandbc])
+
+    Args:
+        n_trials:
+            The number of optimizer trials.
+        metric_name:
+            Name of the optimized metric. Since the metric itself is just a number,
+            `metric_name` can be used to give it a name. So you know later
+            if it was roc-auc or accuracy.
+        project_name:
+            Name of the project under which run should be categorized.
+            If the project is not specified, the run is put in an "Uncategorized" project.
+        group_name:
+            Specifies a group under project into which run should be categorized.
+            Please refer to `Group Runs
+            <https://docs.wandb.ai/guides/track/advanced/grouping>`_ for more details.
+        entity_name:
+            Username or team name under which run should be logged. When set to `None`,
+            the run is logged under the current user.
+        run_name:
+            Specifies the display name in Weights & Biases UI for this run.
+            When set to `None`, random two word name is generated instead.
+        job_type:
+            Specifies type of the run, and is used as additional grouping
+            for the runs.
+    """
+
+    def __init__(
+        self,
+        n_trials: int,
+        metric_name: Union[str] = "value",
+        project_name: Optional[str] = None,
+        group_name: Optional[str] = None,
+        entity_name: Optional[str] = None,
+        run_name: Optional[str] = None,
+        job_type: Optional[str] = None,
+    ) -> None:
+
+        _imports.check()
+
+        self._n_trials = n_trials
+        self._metric_name = metric_name
+        self._project_name = project_name
+        self._group_name = group_name
+        self._entity_name = entity_name
+        self._run_name = run_name
+        self._job_type = job_type
+
+    def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
+
+        if trial.number == 0:
+            self._initialize_run(study, trial)
+
+        for key, value in trial.params.items():
+            wandb.log({key: value}, step=trial.number)
+
+        wandb.log({self._metric_name: trial.value}, step=trial.number)
+
+        # We need to know when study ends to allow
+        # Weights & Biases to finish run and sync up.
+        # This means that this callback can't be currently used
+        # for time-constrained or non-constrained studies.
+        if trial.number + 1 == self._n_trials:
+            datetime_complete = str(trial.datetime_complete)
+            wandb.config.update({"datetime_complete": datetime_complete})
+            wandb.finish()
+
+    def _initialize_run(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
+        """Initializes Weights & Biases run.
+
+        Args:
+            trial:
+                First trial in the study.
+            study:
+                Tracked study.
+        """
+
+        study_direction = str(study.direction).split(".")[-1]
+        config = {
+            "direction": study_direction,
+            "n_trials": self._n_trials,
+            "datetime_start": str(trial.datetime_start),
+        }
+
+        wandb.init(
+            project=self._project_name,
+            group=self._group_name,
+            entity=self._entity_name,
+            name=self._run_name,
+            job_type=self._job_type,
+            config=config,
+            reinit=True,
+        )
--- a/optuna/integration/wandb.py
+++ b/optuna/integration/wandb.py
@@ -1,3 +1,5 @@
+from typing import Any
+from typing import Dict
 from typing import Optional
 from typing import Union
 
@@ -39,11 +41,8 @@ class WeightsAndBiasesCallback(object):
                 return (x - 2) ** 2
 
 
-            wandbc = WeightsAndBiasesCallback(
-                project_name="my-project-name",
-                group_name="my-group-name",
-                job_type="my-job-type",
-            )
+            wandb_kwargs = {"project": "my-project"}
+            wandbc = WeightsAndBiasesCallback(wandb_kwargs=wandb_kwargs)
 
 
             study = optuna.create_study(study_name="my_study")
@@ -54,47 +53,20 @@ class WeightsAndBiasesCallback(object):
             Name of the optimized metric. Since the metric itself is just a number,
             ``metric_name`` can be used to give it a name. So you know later
             if it was roc-auc or accuracy.
-        project_name:
-            Name of the project under which run should be categorized.
-            If the project is not specified, the run is put in an "Uncategorized" project.
-        group_name:
-            Specifies a group under project into which run should be categorized.
-            Please refer to `Group Runs
-            <https://docs.wandb.ai/guides/track/advanced/grouping>`_ for more details.
-        entity_name:
-            Username or team name under which run should be logged. When set to :obj:`None`,
-            the run is logged under the current user.
-        run_name:
-            Specifies the display name in Weights & Biases UI for this run.
-            When set to :obj:`None`, random two word name is generated instead.
-        job_type:
-            Specifies type of the run, and is used as additional grouping
-            for the runs.
-        mode:
-            Specifies execution mode for Weights & Biases. One of
-            ``online``, ``offline`` or ``disabled``. Defaults to online.
+        wandb_kwargs:
+            Set of arguments passed when initializing Weights & Biases run.
+            Please refer to `Weights & Biases API documentation
+            <https://docs.wandb.ai/ref/python/init>`_ for more details.
     """
 
     def __init__(
-        self,
-        metric_name: Union[str] = "value",
-        project_name: Optional[str] = None,
-        group_name: Optional[str] = None,
-        entity_name: Optional[str] = None,
-        run_name: Optional[str] = None,
-        job_type: Optional[str] = None,
-        mode: Union[str] = "online",
+        self, metric_name: Union[str] = "value", wandb_kwargs: Optional[Dict[str, Any]] = None
     ) -> None:
 
         _imports.check()
 
         self._metric_name = metric_name
-        self._project_name = project_name
-        self._group_name = group_name
-        self._entity_name = entity_name
-        self._run_name = run_name
-        self._job_type = job_type
-        self._mode = mode
+        self._wandb_kwargs = wandb_kwargs or {}
 
         self._initialize_run()
 
@@ -116,12 +88,4 @@ class WeightsAndBiasesCallback(object):
     def _initialize_run(self) -> None:
         """Initializes Weights & Biases run."""
 
-        wandb.init(
-            project=self._project_name,
-            group=self._group_name,
-            entity=self._entity_name,
-            name=self._run_name,
-            job_type=self._job_type,
-            mode=self._mode,
-            reinit=True,
-        )
+        wandb.init(**self._wandb_kwargs)
--- a/optuna/integration/wandb.py
+++ b/optuna/integration/wandb.py
@@ -26,6 +26,17 @@ class WeightsAndBiasesCallback(object):
         using this callback in online mode. For more information, please
         refer to `wandb setup <https://docs.wandb.ai/quickstart#1-set-up-wandb>`_.
 
+    .. note::
+        Users who want to run multiple Optuna studies within the same process
+        should call ``wandb.finish()`` between subsequent calls to
+        ``study.optimize()``. Calling ``wandb.finish()`` is not necessary
+        if you are running one Optuna study per process.
+
+    .. note::
+        To ensure correct trial order in Weights & Biases, this callback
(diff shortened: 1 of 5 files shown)


optuna/integration/wandb.py
```diff
@@ -0,0 +1,142 @@
...
+
+import optuna
+from optuna._imports import try_import
+
+
+with try_import() as _imports:
+    import wandb
+
+
+class WeightsAndBiasesCallback(object):
```
xadrianzetx, line 12: Should I decorate it with `@experimental("2.9.0")`?

nzw0301: Yes! Thanks!

nzw0301:
Could you apply [`formats.sh`](https://github.com/optuna/optuna/blob/master/formats.sh) at the top of the optuna directory to fix the CI error for  `Checks`?

Note that the CI failures of `Converge` and `Tests (integration / tests-integration (3.7)` are not caused by this PR. We'll fix them soon (maybe Monday). I'm sorry for confusing you.

xadrianzetx:
Yeah, sorry, forgot to run `isort` on `__init__.py`

optuna/integration/wandb.py
```diff
@@ -0,0 +1,144 @@
...
+    """Callback to track Optuna trials with Weights & Biases.
+
+    This callback enables tracking of Optuna study in
+    Weights & Biases. The study is tracked as a single experiment
+    run, where all suggested hyperparameters and optimized metric
+    are logged and plotted as a function of optimizer steps.
+
+    .. note::
+        User needs to be logged in to Weights & Biases before
```
nzw0301, line 23: How about passing `mode` argument for `wandb.init`? Then we can use the W&B callback offline.

optuna/integration/wandb.py
```diff
@@ -0,0 +1,144 @@
...
+            wandb.log({key: value}, step=trial.number)
+
+        wandb.log({self._metric_name: trial.value}, step=trial.number)
+
+        # We need to know when study ends to allow
+        # Weights & Biases to finish run and sync up.
+        # This means that this callback can't be currently used
+        # for time-constrained or non-constrained studies.
+        if trial.number + 1 == self._n_trials:
+            datetime_complete = str(trial.datetime_complete)
```
nzw0301, line 115: I think `datetime_start` and `datetime_end` are stored at every trial since they are for trial, not study.

optuna/integration/wandb.py
```diff
@@ -0,0 +1,144 @@
...
+            wandb.config.update({"datetime_complete": datetime_complete})
+            wandb.finish()
+
+    def _initialize_run(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
```
nzw0301, line 119: I would like to call this method at `__init__` rather than the first epoch.
In addition, optuna might can `optimize` several times for the same study. it might be difficult to store `n_trials`.

optuna/integration/wandb.py
```diff
@@ -0,0 +1,144 @@
...
+
+        if trial.number == 0:
+            self._initialize_run(study, trial)
+
+        for key, value in trial.params.items():
+            wandb.log({key: value}, step=trial.number)
+
+        wandb.log({self._metric_name: trial.value}, step=trial.number)
+
+        # We need to know when study ends to allow
```
nzw0301, line 110: How about focusing on single process optimisation at this PR? If so, I think we do not need to call the following block because wandb automatically finished the logging.

ref: https://docs.wandb.ai/ref/python/run#finish

optuna/integration/wandb.py
```diff
@@ -0,0 +1,144 @@
...
+
+        self._n_trials = n_trials
+        self._metric_name = metric_name
+        self._project_name = project_name
+        self._group_name = group_name
+        self._entity_name = entity_name
+        self._run_name = run_name
+        self._job_type = job_type
+
+    def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
```
nzw0301, line 100: How about storing the same attributes as in [mlflow's callback](https://github.com/optuna/optuna/blob/cbfefe350d38aa47afff49094be0da1a60ee3d3f/optuna/integration/mlflow.py#L128-L141);
precisely, `distributions`, `direction` and `status`?

nzw0301 (commented):
Thank you for your PR! Basically, looks great! I leave several comments, but I'm not familiar with wandb, so if I was saying something wrong, please correct me.

Could you add a test for this callback?

optuna/integration/wandb.py
xadrianzetx: So if I understand you correctly, instead of logging entire Study as single run, we would log just a Trial in the same way `mlflow` integration does?

nzw0301: Sorry for my unclear comment. I mean the condition `if trial.number + 1 == self._n_trials`  and `wandb.finish()` might be unnecessary for single process optimisation. I suppose the tracked results are the same as the current code.

optuna/integration/wandb.py
xadrianzetx: Sure, thanks for the suggestion!

optuna/integration/wandb.py
xadrianzetx: Sure, although it would be nice in the future to have methods like `on_optimize_start()` and `on_optimize_end()` available for the callbacks to run this kind of cleanup in. Could i propose this in different issue after this PR?

optuna/integration/wandb.py
xadrianzetx: Sure I'll move it. And since we are focusing on single process support in this PR and letting `wandb.finish()` run after script exits, `n_trials` won't be needed as well.

optuna/integration/wandb.py
nzw0301: Of course!

optuna/integration/wandb.py
nzw0301: Sounds perfect! Thanks!

optuna/integration/wandb.py
xadrianzetx: Yeah, I'll remove that, since UI automatically adds those attributes anyway.

optuna/integration/wandb.py
xadrianzetx: Yes, I think logging trial related attributes could be solved with [a table with ](https://docs.wandb.ai/guides/data-vis/tables), and trial related ones can go to config as it is done now. Unfortunately tagging does not work in the same way as in `mlflow`.

Let me work on that some more.

optuna/integration/wandb.py
```diff
@@ -0,0 +1,144 @@
...
+
+            study = optuna.create_study(study_name="my_study")
+            study.optimize(objective, n_trials=n_trials, callbacks=[wandbc])
+
+    Args:
+        n_trials:
+            The number of optimizer trials.
+        metric_name:
+            Name of the optimized metric. Since the metric itself is just a number,
+            `metric_name` can be used to give it a name. So you know later
```
nzw0301, line 59:
```suggestion
            ``metric_name`` can be used to give it a name. So you know later
```

optuna/integration/wandb.py
```diff
@@ -0,0 +1,144 @@
...
+        group_name:
+            Specifies a group under project into which run should be categorized.
+            Please refer to `Group Runs
+            <https://docs.wandb.ai/guides/track/advanced/grouping>`_ for more details.
+        entity_name:
+            Username or team name under which run should be logged. When set to `None`,
+            the run is logged under the current user.
+        run_name:
+            Specifies the display name in Weights & Biases UI for this run.
+            When set to `None`, random two word name is generated instead.
```
nzw0301, line 73:
```suggestion
            When set to :obj:`None`, random two word name is generated instead.
```

optuna/integration/wandb.py
```diff
@@ -0,0 +1,144 @@
...
+            if it was roc-auc or accuracy.
+        project_name:
+            Name of the project under which run should be categorized.
+            If the project is not specified, the run is put in an "Uncategorized" project.
+        group_name:
+            Specifies a group under project into which run should be categorized.
+            Please refer to `Group Runs
+            <https://docs.wandb.ai/guides/track/advanced/grouping>`_ for more details.
+        entity_name:
+            Username or team name under which run should be logged. When set to `None`,
```
nzw0301, line 69:
```suggestion
            Username or team name under which run should be logged. When set to :obj:`None`,
```

nzw0301 (changes requested):
Let me leave minor comments related to the docstring.

xadrianzetx:
On testing - it seems that we can only check if experiment was created. Cannot really query local logs offline like in mlflow tests.

ref: https://github.com/wandb/client/issues/1308

nzw0301:
Thank you for your investigation! It sounds difficult to test like mlflow integration. Indeed, the callback currently a minimal one, so it sounds enough to check if the experiment was created correctly.

As a side note, we may learn a way to deal with the testing from other libraries that uses wandb such as 
- https://github.com/ray-project/ray/blob/master/python/ray/tune/tests/test_integration_wandb.py
- https://github.com/pytorch/ignite/blob/master/tests/ignite/contrib/handlers/test_wandb_logger.py
- https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/loggers/test_wandb.py

xadrianzetx:
Thanks, I will write some tests based on mockup of W&B API.

Crissman:
@not522 Can you take a look at this?

nzw0301:
Sorry, could you merge the master branch into your branch to fix the CI error?

optuna/integration/wandb.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __init__(
+        self,
+        metric_name: Union[str] = "value",
+        project_name: Optional[str] = None,
+        group_name: Optional[str] = None,
+        entity_name: Optional[str] = None,
+        run_name: Optional[str] = None,
+        job_type: Optional[str] = None,
+        mode: Union[str] = "online",
```
nzw0301, line 86: For more flexibility, how about passing the argument of `wandb.init` as a dictionary instance like `skopt_kwargs` in our [`SkoptSampler`](https://github.com/optuna/optuna/blob/master/optuna/integration/skopt.py#L107)?

```suggestion
        wandb_kwargs: Optional[Dict[str, Any]] = None,
```

Sorry, I should have noticed it more earlier.

optuna/integration/wandb.py
```diff
@@ -0,0 +1,127 @@
...
+
+        attributes = {
+            "direction": direction,
+            "trial_state": trial_state,
+            "distributions": trial.distributions,
+            **study.user_attrs,
+        }
+
+        wandb.config.update(attributes)
+        wandb.log({**trial.params, self._metric_name: trial.value}, step=trial.number)
```
nzw0301, line 114: Could you also store [`trial.user_attr`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.trial.FrozenTrial.html#optuna.trial.FrozenTrial.user_attrs)?

optuna/integration/wandb.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
+
+        trial_state = str(trial.state).split(".")[-1]
+        direction = str(study.direction).split(".")[-1]
+
+        attributes = {
+            "direction": direction,
+            "trial_state": trial_state,
+            "distributions": trial.distributions,
```
nzw0301, line 109: `trial.distributions` can be different among trials (I give an example below), so could you store `trial.distributions` by using `wandb.log`?

```python
import optuna
from optuna.integration.wandb import WeightsAndBiasesCallback


def objective(trial):
    if trial.number % 2:
    else:
        x = trial.suggest_float("x", -10, 20)
    return (x - 2) ** 2 + y


wandbc = WeightsAndBiasesCallback()
study.optimize(objective, callbacks=[wandbc], n_trials=10)

```

nzw0301 (changes requested):
Thank you for your quick update! I leave comments on the main logic.

tests/integration_tests/test_wandb.py
```diff
@@ -0,0 +1,112 @@
...
+    study = optuna.create_study(direction="minimize")
+    study.set_user_attr("contributors", ["Akiba", "Sano"])
+    study.set_user_attr("dataset", "MNIST")
+    study.optimize(_objective_func, n_trials=1, callbacks=[wandbc])
+
+    expected = {
+        "direction": "MINIMIZE",
+        "trial_state": "COMPLETE",
+        "distributions": {
+            "x": UniformDistribution(high=10.0, low=-10.0),
```
nzw0301, line 77: [nitpick] could you swap the order of `high` and `low`?

tests/integration_tests/test_wandb.py
```diff
@@ -0,0 +1,112 @@
...
+            "y": LogUniformDistribution(high=10.0, low=1.0),
```
nzw0301, line 78: ditto

optuna/integration/wandb.py
xadrianzetx: I think `wandb.log` is meant to be used with key-value pairs, where values are either scalars or one of [wandb artifacts](https://docs.wandb.ai/ref/python/data-types). If we pass dict with strings, it will not be accessible, instead rendering following information:

image

Proper way to store this, and other trial attributes seems to be an artifact [table](https://docs.wandb.ai/ref/python/data-types/table), where primary key is the trial number. But there is another problem - `wandb.log` does't really seem to handle well continuous logging of updated artifact. Consider following example:

If we roll out couple calls to our callback, including constant updates of our artifact, and giving more than enough time between `log` calls, we will end up looking with something like this

```python
import wandb
import time

wandb.init()

table = wandb.Table(columns=['foo', 'bar'])

table.add_data(*[0, 0])
time.sleep(1.)
wandb.log({"metric": 1, "my-table": table}, step=1)

table.add_data(*[1, 1])
time.sleep(1.)
wandb.log({"metric": 2, "my-table": table}, step=2)

table.add_data(*[2, 2])
time.sleep(1.)
wandb.log({"metric": 3, "my-table": table}, step=3)

table.add_data(*[3, 3])
time.sleep(1.)
wandb.log({"metric": 4, "my-table": table}, step=4)
```

which looks like that UI side

image

Note that we are missing some rows - I think `wandb` just does not manage to update artifact on time. This is even mentioned in `wandb.log` documentation:
> wandb.log is not intended to be called more than a few times per second. If you want to log more frequently than that it's better to aggregate the data on the client side or you may get degraded performance.

This can be fixed easily, by just building a table throughout the run, and commiting it at the end like so

```python
# [...]

table.add_data(*[0, 0])
time.sleep(1.)
wandb.log({"metric": 1}, step=1)

table.add_data(*[1, 1])
time.sleep(1.)
wandb.log({"metric": 2}, step=2)

table.add_data(*[2, 2])
time.sleep(1.)
wandb.log({"metric": 3}, step=3)

table.add_data(*[3, 3])
time.sleep(1.)
wandb.log({"metric": 4}, step=4)

wandb.log({"my-table": table})
``` 

but unfortunately, we don't really have a way to tell our callback that optimization is finished and it should upload artifacts (hence my earlier comments about allowing callbacks to use initialization and cleanup methods like e.g. [keras callback](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/Callback) `on_train_{begin | end}`).

So, due to all this, i think as a temporary solution, distributions should stay logged as part od config, so they can render properly in main run table, and to address your example - i can make callback track distributions used throughout the study, and just update config when anything new shows up. Please let me know if this is ok with you.

Also, for `trial.user_attr` same problems as above apply, but due to the fact that it can be any `dict`, even nested I'm not sure if  `config` workaround will work in this case, since attributes would get unrolled into main table, thus cluttering it. Can we not make it part of this PR, and wait for proper handling of artifacts? Sorry for bringing up `tral.user_attr` here, not in the original comment, but story is the same for all trial attributes.

optuna/integration/wandb.py
xadrianzetx: Sure! Good to have unified API :)

optuna/integration/wandb.py
nzw0301: Awesome! Thank you for sharing your investigation. Alternatively, we can ignore `distributions` at this time–do not track them by using wandb. `tral.user_attr` as well for the same reason as you provided.

xadrianzetx: Great, I will remove all trial related attributes for the time being (so `distributions`, `user_attr` and `state`). As I said previously I can propose and work on extending callback API if you guys don't mind such changes, and come back to improve this callback with proper implementation of artifact tracking.

nzw0301: Sounds perfect!

tests/integration_tests/test_wandb.py
xadrianzetx: Actually, those were removed in 5b17d4e0de5b085dbea8cc2d703857b9e835e76e, since we are not logging trial attributes at the moment.

optuna/integration/wandb.py
xadrianzetx: Addressed in 5b17d4e0de5b085dbea8cc2d703857b9e835e76e

optuna/integration/wandb.py
xadrianzetx: Addressed in 0c4a972720f8e46c3bfd124b167c4350c31df143

tests/integration_tests/test_wandb.py
xadrianzetx: Same as above.

tests/integration_tests/test_wandb.py
```diff
@@ -0,0 +1,77 @@
...
+
+    x = trial.suggest_uniform("x", low=-10, high=10)
+    y = trial.suggest_loguniform("y", low=1, high=10)
+    return (x - 2) ** 2 + (y - 25) ** 2
+
+
+@mock.patch("optuna.integration.wandb.wandb")
+def test_run_initialized(wandb: mock.MagicMock) -> None:
+
+    wandb.run = None
```
nzw0301, line 17: Please correct me if I asking a strange question, but is this line necessary?

nzw0301 (commented):
Basically, I really happy with the current code. Thank you for updating your PR immediately! I have a question about the test code.

tests/integration_tests/test_wandb.py
xadrianzetx: Huh, following pytorch tests, i thought it essentially disables [this](https://docs.wandb.ai/ref/python/run), but you are right - a this point `wandb` is mocked so there is nothing really to disable. Going to remove those.

nzw0301 (approved):
Brilliant! Looks great to me!

optuna/integration/wandb.py
```diff
@@ -0,0 +1,84 @@
...
+
+        self._initialize_run()
+
+    def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
+
+        direction = str(study.direction).split(".")[-1]
+        attributes = {"direction": direction}
+
+        wandb.config.update(attributes)
+        wandb.log({**trial.params, self._metric_name: trial.value}, step=trial.number)
```
vanpelt, line 79: `trial.number` must only increase to be able to use the `step=trial.number`.  If it's possible for `trial.number` to decrease I would suggest something like:

```
commit = self._last_trial_number != trial.number
wandb.log({**trial.params, self._metric_name: trial.value, "global_step": trial.number}, commit=commit)
self._last_trial_number = trial.number
```

optuna/integration/wandb.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def _initialize_run(self) -> None:
+        """Initializes Weights & Biases run."""
+
+        wandb.init(**self._wandb_kwargs)
```
vanpelt, line 84: If there's a way to hook into a run finishing, I would also add `wandb.finish()` in that hook.  We call this automatically when the process where `wandb.init` was called in exits, but if a user runs multiple trials in the same process you should call `wandb.finish()` before the next call to `wandb.init()`.

vanpelt (commented):
All and all looks good, just a couple comments to address.

optuna/integration/wandb.py
vanpelt: The addition of the `commit` logic is only needed if it's common for this function to be called many times for the same `trial.number`.  If not, you can just get rid of the commit logic.

optuna/integration/wandb.py
xadrianzetx: Yeah, i agree, but callback API needs to be modified a bit to allow pre/post optimization calls to callbacks. Out of scope for that PR, but I will work on this in near future.

optuna/integration/wandb.py
xadrianzetx: @nzw0301 please correct me if I'm wrong, but assuming sequential optimization, this should not be a problem, right?

nzw0301: Hi, @vanpelt, thank you for giving your feedback! In this PR, we can assume that `trial.number` must increase. However, in the future, we support such a situation when `trial.number` might decrease, especially, distributed optimisation by optuna. So your feedback is really helpful!

optuna/integration/wandb.py
nzw0301: Indeed, that is a really good point. Both scenarios might happen. Could you add a note section to the docstring? The note format is like this
https://github.com/optuna/optuna/blob/c2e528a3f2b8d67075cec117ecc964601f609cc3/optuna/integration/cma.py#L64

optuna/integration/wandb.py
nzw0301: Ah, so it would be great to mention it as a note as well. Optuna still supports the `n_jobs` argument of `study.optimize` function. Even though we use a single process optimization, it might be happen when `n_jobs != 1`.

xadrianzetx: I will add a note on this as well. Thanks @vanpelt for the feedback!

xadrianzetx:
@nzw0301 I've added missing notes to docstring. Sorry for small delay.

nzw0301:
@xadrianzetx Your update is always fast, I think :) I'm happy with the current code. 

Another dev-member will review this PR to merge this PR into the master branch, so please give us a little bit more time.

not522 (changes requested):

Could you add a note to the document that this feature only supports single-objective optimization, not multi-objective optimization?

And, I added small suggestions.

xadrianzetx:
Thanks @not522 for the review! I've addressed all suggestions in ed63ce78d7c7ae7fc9d55bdd0f69386d69c56c59. BTW multi-objective support should be pretty straightforward to implement. I will open PR for this soon.

optuna/integration/wandb.py
```diff
@@ -0,0 +1,95 @@
...
+        _imports.check()
+
+        self._metric_name = metric_name
+        self._wandb_kwargs = wandb_kwargs or {}
+
+        self._initialize_run()
+
+    def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
+
+        direction = str(study.direction).split(".")[-1]
```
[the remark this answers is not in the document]
xadrianzetx: Nice, didn't notice that `TrialState` inherits from `enum.Enum`.

not522 (approved):
Thank you! LGTM!
 
 @vanpelt Do you have any concerns about merging this PR?

not522:
Let me merge this PR.

@vanpelt Thank you for your review! If you have any comments, feel free to add them even after merging.

xadrianzetx:
Again, thank you for the reviews, @nzw0301, @not522, @vanpelt!

nzw0301:
@xadrianzetx kudos for your pull request! By the way, W&B team tries to contact you to tell you thanks and send you their shirt for this PR. Could you tell [me](mailto:k_nzw@klis.tsukuba.ac.jp) your contact info (maybe e-mail)? I'll forward your info to them.

Issue #852, opened 2020-01-22
Warn multiple intermediate value reports with the same step.


sile:
Currently, if `trial.report()` method is called multiple times with the same step in a trial, only the first reported value will be recorded into the storage and the others will be silently ignored.
Reporting multiple values with the same step is an unintended usage of the method, and it could cause bugs that are hard to notice (see #847 as an example of such bug).
Therefore, it seems better to add a check to detect the duplicate reports, and if it's detected, we should emit a warning message.

As an implementation note, this modification could be done without any additional DB costs (see https://github.com/optuna/optuna/pull/847#issuecomment-576512000 for the detail).

stakky:
@sile Hi. Can I take this issue?

sile:
Of course, please go for it!

stakky:
@sile 
I implemanted a prototype and tested it. I changed a pruning sample(optuna/examples/pruning/simple.py) to reproduce the issue and then warning message was emitted as intended.
https://github.com/stakky/optuna/commit/d42fc64fdfefbdede4ec3cdde623cd18daf80fe2
But this prototype emits the warning message too much. I think it should be emited per one trial. What do you think?

sile:
@stakky Thanks!

> But this prototype emits the warning message too much. I think it should be emited per one trial.

I agree with your opinion. I think that it's better to use [warnings.warn()](https://docs.python.org/3/library/warnings.html#warnings.warn
) instead of Optuna logger. `warnings.warn()` don't emit duplicate messages.
The detail has not been decided, but Optuna dev team is planning to use `warnings.warn()` for warnings related to implementation miss, and Optuna logger for warnings related to experimentation (e.g., trial failure report).

sile:
Anyway, if you create a PR, we can discuss it on the PR.

hvy:
Still seems to be an open issue.

norihitoishida:
Let me tackle this issue!

TakuyaInoue-github:
I'm sorry for interrupting but there is no activity on the issue. so i tuckle this issue and create PR.
Please check and comment.

PR #2782, opened 2021-07-04
add the warning if intermediate values is already reported

TakuyaInoue-github:
## Motivation
resolve the issue of https://github.com/optuna/optuna/issues/852.

## Description of the changes
add the warning if the step is already reported.

### DIFF
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -587,7 +587,7 @@ class Trial(BaseTrial):
 
         if step in intermediate_values:
             # Do nothing if already reported.
-            warnings.warn("this `step` {} is already repoted.".format(step))
+            warnings.warn("The reported value is ignored because this `step` {} is already reported.".format(step))
             return
 
         self.storage.set_trial_intermediate_value(self._trial_id, step, value)
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -587,9 +587,7 @@ class Trial(BaseTrial):
 
         if step in intermediate_values:
             # Do nothing if already reported.
-            # TODO(hvy): Consider raising a warning or an error.
-            # See https://github.com/optuna/optuna/issues/852.
-            warnings.warn("this 'step' {} is already repoted.".format(step))
+            warnings.warn("this `step` {} is already repoted.".format(step))
             return
 
         self.storage.set_trial_intermediate_value(self._trial_id, step, value)
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -588,8 +588,8 @@ class Trial(BaseTrial):
         if step in intermediate_values:
             # Do nothing if already reported.
             warnings.warn(
-                    "The reported value is ignored because this `step` {} is already reported.".format(
-                        step
+                "The reported value is ignored because this `step` {} is already reported.".format(
+                    step
                 )
             )
             return
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -587,7 +587,11 @@ class Trial(BaseTrial):
 
         if step in intermediate_values:
             # Do nothing if already reported.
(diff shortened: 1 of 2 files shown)


optuna/trial/_trial.py
```diff
@@ -589,6 +589,7 @@ def objective(trial):
             # Do nothing if already reported.
             # TODO(hvy): Consider raising a warning or an error.
             # See https://github.com/optuna/optuna/issues/852.
```
nzw0301, line 591: Could you remove these lines? I think this PR resolves this TODO task.

optuna/trial/_trial.py
```diff
@@ -589,6 +589,7 @@ def objective(trial):
...
+            warnings.warn("this 'step' {} is already repoted.".format(step))
```
nzw0301, line 592:
```suggestion
            warnings.warn("this `step` {} is already repoted.".format(step))
```

nzw0301 (changes requested):
Thank you for your PR! Could you add a test?

optuna/trial/_trial.py
nzw0301: We can improve the warning message by clarifying what happens for users, like

```python
"The reported value is ignored because this `step` {} is already reported.".format(step)
```

TakuyaInoue-github: >  Could you add a test?

Of course.
 I would like to confirm that the test is checking of return value is None at the multiple calling?

nzw0301: Sure! Probably no because the correct `report` also returns `None`. I think adding the following code after [`test_report`](https://github.com/optuna/optuna/blob/99484054fcec97eb530960a8612c3e6058ee5690/tests/trial_tests/test_trial.py#L573-L600) is enough.

```python
def test_report_warning() -> None:

    study = create_study()
    trial = study.ask()

    trial.report(1.23, 1)

        trial.report(1, 1)
```

tests/trial_tests/test_trial.py
```diff
@@ -600,6 +600,18 @@ def test_report() -> None:
...
 
 
+def test_report_warning() -> None:
+
+    study = create_study()
+    trial = study.ask()
+
+    trial.report(1.23, 1)
+
+    # Warn if multipe times call  report method at the same step
```
nzw0301, line 610:
```suggestion
    # Warn if multiple times call report method at the same step
```

nzw0301 (approved):
I'm happy with the PR. Thank you for a quick update. Please give us time to review this PR by another dev-member and fix the master branch's CI failure. We'll ask you to merge the master branch later when we fix it.

TakuyaInoue-github:
Thank you too for your appropriate points and comments.

nzw0301:
Hi, could you merge the master branch to fix the CI errors?

TakuyaInoue-github:
I merged the master blanch, so I would be happy if you could check.

Crissman (approved):
LGTM. Thanks!

nzw0301:
Let me rename the PR to fix the typo and specify the case when we see the warning message.

TakuyaInoue-github:
Sure, Please let me know, if i should do anything.

nvelden:
Can you set a time-out for this warning once it is displayed? The flooding of error messages caused my computer to crash.

nzw0301:
@mirekphd @nvelden  I'm wondering why the code reports intermediate value multiple times at the same step. I'm afraid that it sounds inaccurate usage of pruner or callback for me...

nzw0301:
@nvelden thank you for sharing the code! I suppose that the `model.fit` is called `k` times, where `k` is k-fold cross validation's `k`. I think that it works as we expected only in the first loop of cross-validation; the callback does nothing in the other loops because the pruner looked at the first cv's reported values that are not overwritten in other loops as too many warning messages said.

As a workaround, I suggest not using `XGBoostPruningCallback` except for the first cv loop. However, I'm not familiar with xgboost, so I'm not sure this is the best way for your case.

PR #2789, opened 2021-07-07
Fix the document of `RetryFailedTrialCallback.retried_trial_number`

not522:
## Motivation
The document of `RetryFailedTrialCallback.retried_trial_number` is unclear. This PR will clarify what to return.

For example,

- `trial1` is a retry of `trial0`
- `trial2` is a retry of `trial1`

In this case, `retried_trial_number(trial2)` should return `0`, but in the current document, it looks like to return `1`.

## Description of the changes
- Fix the document

### DIFF
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -132,7 +132,7 @@ class RetryFailedTrialCallback:
                 The trial object.
 
         Returns:
-            The number of the previous trial. If not retry of a previous trial,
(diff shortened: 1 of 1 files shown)


hvy (approved):
LGTM!

hvy:
For those wondering where in the code this is defined https://github.com/optuna/optuna/blob/master/optuna/_callbacks.py#L102-L105.

Crissman (approved):
LGTM

PR #2791, opened 2021-07-07
Avoid applying `constraint_fn` to non-`COMPLETE` trials in NSGAII-sampler.

### LINKED ISSUE #2790: `NSGAIISampler` tries to evaluate `constraint_func` even when trails fail.
## Expected behavior

`NSGAIISampler` does not evaluate `constraints_func` when trials did not complete.

## Environment

- Optuna version: `2.9.0.dev`
- Python version: 3.8.6
- OS: ubuntu (WSL2)

```
[I 2021-07-07 12:44:56,751] A new study created in memory with name: no-name-be57671e-58b7-4219-a290-93f93adccdcd
[I 2021-07-07 12:44:56,752] Trial 0 finished with values: [0.1, 0.2] and parameters: {}.
[I 2021-07-07 12:44:56,752] Trial 1 pruned.
  File "tmp.py", line 11, in <module>
    optuna.create_study(sampler=sampler, directions=["maximize", "minimize"]).optimize(objective, n_trials=200)
  File "path/to/optuna/optuna/study/study.py", line 400, in optimize
    _optimize(
  File "path/to/optuna/optuna/study/_optimize.py", line 65, in _optimize
    _optimize_sequential(
  File "path/to/optuna/optuna/study/_optimize.py", line 162, in _optimize_sequential
    trial = _run_trial(study, func, catch)
  File "path/to/optuna/optuna/study/_optimize.py", line 243, in _run_trial
    study.tell(trial, values=values, state=state)
  File "path/to/optuna/optuna/study/study.py", line 657, in tell
    self.sampler.after_trial(study, frozen_trial, state, values)
  File "path/to/optuna/optuna/samplers/_nsga2.py", line 353, in after_trial
    con = self._constraints_func(trial)
  File "tmp.py", line 10, in <lambda>
    sampler = optuna.samplers.NSGAIISampler(constraints_func=lambda t: [t.user_attrs["constraint"]])
KeyError: 'constraint'
```

## Steps to reproduce

1. Optimize an objective that failes w/ `NSGAIISampler` and `constraint_func`.

## Reproducible examples (optional)

```python
import optuna

def objective(trial):
    if trial.number % 2:
        raise optuna.TrialPruned
    trial.set_user_attr("constraint", 0.2 - 0.1)
    return 0.1, 0.2


sampler = optuna.samplers.NSGAIISampler(constraints_func=lambda t: [t.user_attrs["constraint"]])
optuna.create_study(sampler=sampler, directions=["maximize", "minimize"]).optimize(objective, n_trials=200)
```

ytsmiling:
## Motivation

This PR resolves #2790.

## Description of the changes

Stop evaluating `constraint_func` when trials' status are not `TrialState.COMPLETE`.

### DIFF
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -71,6 +71,10 @@ class NSGAIISampler(BaseSampler):
             If constraints_func returns more than one value for a trial, that trial is considered
             feasible if and only if all values are equal to 0 or smaller.
 
+            The constraint_func will be evaluated after each successful trial.
+            The function won't be called when trials fail or they are pruned, but this behavior is
+            subject to change in the future releases.
+
             The constraints are handled by the constrained domination. A trial x is said to
             constrained-dominate a trial y, if any of the following conditions is true:
 
--- a/optuna/samplers/_nsga2.py
+++ b/optuna/samplers/_nsga2.py
@@ -213,6 +213,7 @@ class NSGAIISampler(BaseSampler):
                     generation_to_runnings[generation].append(trial)
                 continue
 
+            # Do not use trials whose states are not COMPLETE, or `constraint` will be unavailable.
             generation_to_population[generation].append(trial)
 
         hasher = hashlib.sha256()
@@ -347,7 +348,8 @@ class NSGAIISampler(BaseSampler):
         state: TrialState,
         values: Optional[Sequence[float]],
     ) -> None:
-        if self._constraints_func is not None:
+        assert state in [TrialState.COMPLETE, TrialState.FAIL, TrialState.PRUNED]
+        if state == TrialState.COMPLETE and self._constraints_func is not None:
             constraints = None
             try:
                 con = self._constraints_func(trial)


hvy (approved):
Basically LGTM. My only concerns would be if different samplers would have similar constraint callback interfaces but some for instance would consider pruned trials for their suggestions. Then, the behavior would be different for different samplers which might be a bit confusing. Given it's an experimental feature, I think this change is reasonable.

I'd by the way prefer updating the docs about the guarantees of the trial and adding a test, but if you're busy we could follow that up in a separate PR.

ytsmiling:
Thanks for reviewing! 
> My only concerns would be if different samplers would have similar constraint callback interfaces but some for instance would consider pruned trials for their suggestions.

Make sense.
> I'd by the way prefer updating the docs about the guarantees of the trial

Let me update the doc! 

> and adding a test, but if you're busy we could follow that up in a separate PR.

I found there's a room for discussion, that is, we may add a test for this specific case or add more general tests for failing-trail handling. Anyway, it's very helpful if the core-dev team follow this up.

hvy:
@himkt a friendly ping (maybe it never got there in the first place) but are you able to review this PR? Otherwise, we can assign someone else.

himkt:
Sorry I missed it. :bow: I'll review it today.

himkt (approved):
Sorry for the late. LGTM.

Issue #1990, opened 2020-11-03
Support distributed training in AllenNLP integration


himkt:
Current `AllenNLPExecutor` doesn't consider distributed training.
It would be useful if an executor supports distributed training.

## Description

AllenNLP supports multi-device distributed training (https://medium.com/ai2-blog/c4d7c17eb6d6).
However, the current `AllenNLPExecutor` doesn't consider the multiple GPUs situation in distributed optimization.

In distributed optimization, multiple processes may try to allocate memories on multiple GPUs and it could cause a memory error. It would be needed to schedule GPUs for each process.

himkt:
Distributed training in AllenNLP spawns processes for training.
At this time, AllenNLPExecutor may fail to pass environment variables to them.
(https://github.com/himkt/allennlp-optuna/issues/20)

himkt:
#2977 introduces the support for distributed training. This feature will be available on the next release of Optuna.

Issue #297, opened 2019-01-07
View multiple studies in dashboard


MannyKayy:
Currently, you have to pass the study name to the dashboard cli when the dashboard is launched.
e.g. "example-study":
`optuna dashboard --storage "sqlite:///example.db" --study "example-study"`

It would be helpful if the study cli argument  was optional and we could select and view different studies within the database via the dashboard.

Thanks,

g-votte:
@MannyKayy  

Thanks a lot for your feedback!
I agree it’s more helpful if we can track multiple studies in a single dashboard process.

Actually, we’ve been working on dashboard enhancement and considering to add multi-study feature as part of that.
As it’ll still take some time before release, please be patient for a while.

sile:
@dwiel I think [`optuna.study.get_all_study_summaries`](https://optuna.readthedocs.io/en/stable/reference/study.html#optuna.study.get_all_study_summaries) function can be used for such purpose.

PR #2796, opened 2021-07-12
Support `AllenNLPPruningCallback` in distributed configuration

himkt:
This PR changes the way to pass the `trial` object to `AllenNLPPruningCallback`.

- https://github.com/optuna/optuna/issues/1990
- https://github.com/himkt/allennlp-optuna/issues/20

## Motivation
The current AllenNLP integration uses environment variables to store Optuna's trial information in environment variables and load them to re-construct the trial on the AllenNLP side.

- set

https://github.com/optuna/optuna/blob/2c16baa7bda8ea74ff2e1064684aa2d411a2a0fd/optuna/integration/allennlp/_executor.py#L161-L178

https://github.com/optuna/optuna/blob/2c16baa7bda8ea74ff2e1064684aa2d411a2a0fd/optuna/integration/allennlp/_executor.py#L213

https://github.com/optuna/optuna/blob/2c16baa7bda8ea74ff2e1064684aa2d411a2a0fd/optuna/integration/allennlp/_executor.py#L193-L195

- load

https://github.com/optuna/optuna/blob/2c16baa7bda8ea74ff2e1064684aa2d411a2a0fd/optuna/integration/allennlp/_pruner.py#L168-L198

This approach is problematic when users attempt to train a model in a distributed configuration.
In AllenNLP distributed training, `torch.multiprocessing` is used, which means training workers are spawned from the AllenNLP main training process. Spawned processes cannot find Optuna's process to fail to load trial's information. This is why we cannot support pruning callback in the distributed situation.

To solve this problem, in this PR I change the way trial is passed to the AllenNLP. I extend `allennlp.commands.train.train_model` to take `trial` as an input argument (632813c).
(This approach was originally taught by @matt-gardner. Thank you so much)
By being able to pass the trial object (python object) directly, we can remove the environment variable hacks and pruning callback is available in a distributed configuration.

Current Implementation | This PR's implementation
 --  |  --
optuna_before | optuna_after
Use environment variables to restore trial object in the construction of `AllenNLPPruningCallback` | Expand `allennlp.commands.train.train_model` so that it takes `trial` object as arguments and uses it to construct `AllenNLPPruningCallback`

## Description of the changes
- 632813c 23780ef change a way to pass a trial object
-  f27d0df updates tests
- 4691dfb reconstructs storage to get up-to-date information from remote storage
- fc93982 1734f42 minor changes

## Operation verifications
- [x] sqlite3
- [x] mysql
- [x] postgresql


optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,378 @@
...
+
+        if isinstance(trial.study._storage, InMemoryStorage):
+            message = "InMemoryStorage is not supported in distributed configuration."
+            message += " You have to use RDB or Redis to use Optuna with AllenNLP distributed."
+            raise ValueError(message)
+
+        elif isinstance(trial.study._storage, _CachedStorage):
+            # Reconstruct storage to purge a old cache.
+            if isinstance(trial.study._storage, _CachedStorage):
+                trial.study._storage = _CachedStorage(trial.study._storage._backend)
```
himkt, line 182: I don't know if I can explain it well, but I try to do.

trial_distributed

Worker processes are spawned from the main process. Each worker runs a training loop and reports a metric to Optuna. Optuna's trial object is copied to each worker.

When `trial.report` is invoked, an objective value is recorded to a storage but Optuna's `CachedStorage` mechanism caches a state in memory and avoid executing a query to the storage. Each worker terminates after training thus the changes in a worker's state are not propagated to the main process.

To get previous trials' information, I re-create a storage object (it fetches information from a remote storage) before starting training.

optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,378 @@
...
+                    world_size,
+                    device_ids,
+                    file_friendly_logging,
+                    include_in_archive,
+                    trial,
+                ),
+                nprocs=num_procs,
+            )
+        except Exception as e:
+            if "optuna.exceptions.TrialPruned" in str(e):
```
himkt, line 204: (I don't think it is a good way)

When a pruning happens, the worker terminates with `torch.multiprocessing.spawn.ProcessRaisedException`.
To prune a trial in Optuna, it has to raise `optuna.exceptions.TrialPruned` instead of `ProcessRaisedException`.
In the first implementation, I check a traceback and re-raise `TrialPruned` if it is needed.

tests/integration_tests/allennlp_tests/example_with_executor_and_pruner.jsonnet
```diff
@@ -60,6 +60,7 @@ local DROPOUT = std.parseJson(std.extVar('DROPOUT'));
     callbacks: [
       {
         type: 'optuna_pruner',
+        monitor: 'best_validation_loss',
```
himkt, line 63: After this PR is merged, `monitor` is required to specify in a config file.
I will document it.

himkt:
@hvy @HideakiImamura @toshihikoyanase
Sorry for mentioning but I want you to review this PR because you gave me helpful reviews and I think you have background knowledge.

optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,378 @@
...
+from typing import Union
+
+from allennlp.commands.train import TrainModel
+from allennlp.common import logging as common_logging
+from allennlp.common import Params
+from allennlp.common import util as common_util
+from allennlp.common.checks import check_for_gpu
+from allennlp.common.checks import ConfigurationError
+from allennlp.common.meta import Meta
+from allennlp.common.meta import META_NAME
```
himkt, line 20: `allennlp.common.meta` was introduced in AllenNLP v2.5.0 (https://github.com/allenai/allennlp/releases/tag/v2.5.0).
To support older versions of AllenNLP, we may have to conditionally import them.

HideakiImamura:

optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,406 @@
...
+from allennlp.data import Vocabulary
+from allennlp.models.archival import archive_model
+from allennlp.models.archival import CONFIG_NAME
+from allennlp.models.archival import verify_include_in_archive
+from allennlp.models.model import _DEFAULT_WEIGHTS
+from allennlp.models.model import Model
+from allennlp.training import util as training_util
+import torch
+import torch.distributed as dist
+import torch.multiprocessing as mp
```
HideakiImamura, line 31: Please move these imports inside with `try_import() as _imports:` block.

optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,406 @@
+"""
+This file is ported from AllenNLP original implementation. (Under Apache 2.0 license)
+https://github.com/allenai/allennlp/blob/main/allennlp/commands/train.py
+"""
+
+import logging
```
HideakiImamura, line 6: How about using Optuna's logging module?

optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,406 @@
...
+
+if not _imports.is_successful():
+    Meta = None  # type: ignore  # NOQA
+    META_NAME = None  # type: ignore  # NOQA
+
+
+logger = logging.getLogger(__name__)
+
+
+def train_model_with_optuna(
```
HideakiImamura, line 52: How about making this function private?
```suggestion
def _train_model_with_optuna(
```

optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,406 @@
...
+    dry_run: bool = False,
+    file_friendly_logging: bool = False,
+    trial: Trial = None,
+) -> Optional[Model]:
+    """Extended AllenNLP training utility.
+
+    # Parameters
+
+    params : `Params`
+        A parameter object specifying an AllenNLP Experiment.
```
HideakiImamura, line 68: How about adapting the docstring to Optuna's standard format? For example, 
```suggestion
    Args:
        params:
            A parameter object specifying an AllenNLP Experiment.
```

optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,406 @@
...
+    primary_port: int = 29500,
+    world_size: int = 1,
+    distributed_device_ids: List[int] = None,
+    file_friendly_logging: bool = False,
+    include_in_archive: List[str] = None,
+    trial: Trial = None,
+) -> Optional[Model]:
+    """Extended AllenNLP train_worker.
+
+    # Parameters
```
HideakiImamura, line 244: Ditto.

tests/integration_tests/allennlp_tests/example_with_executor_and_pruner.jsonnet
HideakiImamura: Is it documented anywhere?

HideakiImamura (commented):
Thanks for the PR and sorry for the late review. Basically, LGTM. I have several comments. Please take a look.

tests/integration_tests/allennlp_tests/example_with_executor_and_pruner.jsonnet
himkt: I found we can use a metric specified in `AllenNLPExecutor` in e1da97c.
This keeps compatibility and removed this line in 261b4ea.

optuna/integration/allennlp/_executor.py
```diff
@@ -141,41 +99,13 @@ def __init__(
...
-            _MONITOR: metrics,
-            _PRUNER_KEYS: ",".join(pruner_params.keys()),
-        }
-
-        if trial.study.pruner is not None:
-            system_attrs[_PRUNER_CLASS] = type(trial.study.pruner).__name__
-
-        system_attrs.update(pruner_params)
-        self._system_attrs = system_attrs
+        self._trial.set_system_attr("monitor", metrics)
```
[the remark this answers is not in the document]
himkt: It makes sense. I changed the key in 8f01851 and 4cefd1a.

optuna/integration/allennlp/_pruner.py
```diff
@@ -143,7 +81,7 @@ class AllenNLPPruningCallback(TrainerCallback):
 
     def __init__(
         self,
-        trial: Optional[optuna.trial.Trial] = None,
+        trial: optuna.trial.Trial,
```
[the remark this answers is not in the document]
himkt: It could leave it as is but not giving `trial` doesn't work anymore after this PR.

The current implementation restores a `trial` from environment variables so `trial` is optional. But this logic will be completely removed. The only when `trial` is not passed to a pruning callback is that users uses the integration in the Jsonnet style usage. This means it always happens internally and I think the change of this doesn't affect users.

So, I want to make `trial` required. If you don't understand what I mean (my explanation must not be well), please ask me.

optuna/integration/allennlp/_train.py
himkt: 44bb1b1 changes the logging module.

optuna/integration/allennlp/_train.py
himkt: 9f962c0 made it private.

optuna/integration/allennlp/_train.py
himkt: e4f4bd9 moves them.

optuna/integration/allennlp/_train.py
himkt: Thank you. You suggestion and b76a997 update docstrings.

optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,406 @@
+"""
+This file is ported from AllenNLP original implementation. (Under Apache 2.0 license)
+https://github.com/allenai/allennlp/blob/main/allennlp/commands/train.py
+"""
```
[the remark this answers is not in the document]
himkt: As I understand, Apache 2.0 allows us to modify and distribute a source code.
But it's an important point and we have to be carefully, we'll check in detail.

himkt:
Thank you for the review @HideakiImamura, I revised the PR. PTAL.

optuna/integration/allennlp/_train.py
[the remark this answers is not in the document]
himkt: Added copyright notice in b39790b.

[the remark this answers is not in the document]
himkt: Updated in 64c4f3c.

optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,438 @@
...
+    include_package: List[str] = None,
+    dry_run: bool = False,
+    node_rank: int = 0,
+    primary_addr: str = "127.0.0.1",
+    primary_port: int = 29500,
+    world_size: int = 1,
+    distributed_device_ids: List[int] = None,
+    file_friendly_logging: bool = False,
+    include_in_archive: List[str] = None,
+    trial: Trial = None,
```
HideakiImamura, line 278: Ditto.

optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,438 @@
+"""
```
HideakiImamura, line 1: In my understanding, most of this file is a transfer of AllenNLP code, and Optuna is only involved in the part that passes trial to `allennlp.commands.train.TrainModel`, restores `_CachedStorage`, and catches `TrialPruned`. As a suggestion, how about adding these Optuna-specific parts as an integration to the AllenNLP side? That is, put these changes together and send a PR to AllenNLP. Is this difficult to do?

HideakiImamura (commented):
Thanks for your swift actions. I added some comments. PTAL.

optuna/integration/allennlp/_train.py
himkt: I created the issue on the AllenNLP repo (https://github.com/allenai/allennlp/issues/5338).
I'd like to go on this PR for now and migrate if the feature request is approved. What do you think about the plan?

optuna/integration/allennlp/_train.py
```diff
@@ -0,0 +1,438 @@
...
+def _train_model_with_optuna(
+    params: Params,
+    serialization_dir: Union[str, PathLike],
+    recover: bool = False,
+    force: bool = False,
+    node_rank: int = 0,
+    include_package: List[str] = None,
+    dry_run: bool = False,
+    file_friendly_logging: bool = False,
+    trial: Trial = None,
```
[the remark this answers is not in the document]
himkt: Sorry and thank you, I fixed in ea79d88.

optuna/integration/allennlp/_train.py
HideakiImamura: Sounds reasonable. Thank you for creating the issue in AllenNLP. I agree with your idea. Let's proceed with this PR.

HideakiImamura: @toshihikoyanase Do you have any opinion on this? I'm very glad if you provide any insights.

toshihikoyanase: Sounds good!
After we merge this PR, we need to catch up with the changes of AllenNLP not to stale this integration.
But, I'm not sure if they include Optuna as a (mandatory) dependency, and I guess some plugin features may be preferable.
Anyway, thanks @himkt for creating the issue.

HideakiImamura (approved):
Sorry for the late review and thanks for the private discussion. The code looks good to me.

[Note] We decided to give up on putting any Optuna related code in the AllenNLP side. The reason for this is to avoid interdependencies between Optuna and AllenNLP.

himkt:
I decided to change my approach for supporting distributed training, because it is so hard to follow changes in AllenNLP training module. Instead, I'll work on the problem in another PR (https://github.com/optuna/optuna/pull/297).

@HideakiImamura @toshihikoyanase Sorry for closing the PR even you spent time to review. And, please, help/review a subsequent PR. :bow:

PR #2798, opened 2021-07-13
Match the case of `ID`

belldandyxtq:
## Description of the changes
Fix the case of id in the doc to match others https://github.com/optuna/optuna/blob/4cb27999446e97e7f194a76e090d7221771d047c/optuna/storages/_base.py#L101


hvy (approved):
LGTM!

I'll merge it once CI finishes.

PR #2811, opened 2021-07-16
Function for failing zombie trials and invoke their callbacks

hvy:
## Motivation

Currently, zombie trials are checked for and automatically failed in the beginning of each trial when running `Study.optimize` (https://github.com/optuna/optuna/blob/master/optuna/study/_optimize.py#L189-L195). However, users would like to fail these zombie trials without having to invoke `Study.optimize`.

_Zombie trials are trials that have been abnormally terminated and are stuck in the RUNNING state._

## Description of the changes

This PR introduces a function that allows users to sweep and fail zombie trials at any time.

## Note

The function is marked experimental since the Storage/heartbeat APIs in general are somewhat controversial and open for discussion affecting the design of this API. Note that we could make this a method of the `Study` but I'm not sure we should expose it to that degree, yet.

### DIFF
--- a/optuna/storages/_heartbeat.py
+++ b/optuna/storages/_heartbeat.py
@@ -5,7 +5,7 @@ from optuna._experimental import experimental
 
 
 @experimental("2.9.0")
-def fail_stale_trials_with_callback(study: "optuna.Study") -> None:
+def fail_stale_trials(study: "optuna.Study") -> None:
     """Fail stale trials and run their failure callbacks.
 
     The running trials whose heartbeat has not been updated for a long time will be failed,
--- a/optuna/storages/_heartbeat.py
+++ b/optuna/storages/_heartbeat.py
@@ -13,7 +13,7 @@ def fail_stale_trials_with_callback(study: "optuna.Study") -> None:
 
     .. seealso::
 
-       See :class:`~optuna.storages.RDBStorage`.
+        See :class:`~optuna.storages.RDBStorage`.
 
     Args:
         study:
--- a/optuna/storages/_heartbeat.py
+++ b/optuna/storages/_heartbeat.py
@@ -1,0 +1,32 @@
+import copy
+
+import optuna
+from optuna._experimental import experimental
+
+
+@experimental("2.9.0")
+def fail_stale_trials_with_callback(study: "optuna.Study") -> None:
+    """Fail stale trials and run their failure callbacks.
+
+    The running trials whose heartbeat has not been updated for a long time will be failed,
+    that is, those states will be changed to :obj:`~optuna.trial.TrialState.FAIL`.
+
+    .. seealso::
+
+       See :class:`~optuna.storages.RDBStorage`.
+
+    Args:
+        study:
+            Study holding the trials to check.
+    """
+    storage = study._storage
+
+    if not storage.is_heartbeat_enabled():
+        return
+
+    failed_trial_ids = storage.fail_stale_trials(study._study_id)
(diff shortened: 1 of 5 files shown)


hvy:
@not522 , @nzw0301 can you review this PR? 

I'm currently looking into the documentation build failure but the changes are otherwise ready for review.

optuna/study/_optimize.py
```diff
@@ -186,13 +186,7 @@ def _run_trial(
...
     catch: Tuple[Type[Exception], ...],
 ) -> trial_module.Trial:
-    if study._storage.is_heartbeat_enabled():
-        failed_trial_ids = study._storage.fail_stale_trials(study._study_id)
-        failed_trial_callback = study._storage.get_failed_trial_callback()
-        if failed_trial_callback is not None:
-            for trial_id in failed_trial_ids:
-                failed_trial = copy.deepcopy(study._storage.get_trial(trial_id))
-                failed_trial_callback(study, failed_trial)
+    optuna.storages.fail_stale_trials(study)
```
not522, line 189: It always raises `ExperimentalWarning`. I think it's better to suppress it, except the user explicitly uses `fail_stale_trials`.

hvy: You're right, I'll update the code.

hvy: Fixed!

hvy:
I'm not sure why the docs are failing. 🤔

hvy:
Fixed the above error by moving the function definitions to a separate file (and not an `__init__.py`). It might have been caused by the fact that I also imported `optuna` in the `__init__.py` before. Thanks @not522 for the debugging!

optuna/storages/_heartbeat.py
```diff
@@ -0,0 +1,32 @@
+import copy
+
+import optuna
+from optuna._experimental import experimental
+
+
+@experimental("2.9.0")
+def fail_stale_trials_with_callback(study: "optuna.Study") -> None:
```
hvy, line 8: Name is open for discussion. Just, `fail_stale_trials` might be simpler and better.

tests/storages_tests/test_storages.py
```diff
@@ -1087,3 +1088,31 @@ def test_get_trial_id_from_study_id_trial_number(storage_mode: str) -> None:
         assert trial_id == storage.get_trial_id_from_study_id_trial_number(
             study_id, trial_number=0
         )
+
+
+def test_fail_stail_trials_with_callback() -> None:
```
nzw0301, line 1093: typo
```suggestion
def test_fail_stale_trials_with_callback() -> None:
```

nzw0301 (changes requested):
Basically, looks great to me! I found a typo and minor indent issue. Could look at them?

hvy:
Let me resolve the conflicts...

hvy:
PTAL. 🙇

optuna/storages/_heartbeat.py
not522: I agree. With `with_callback`, it sounds like a callback is needed.

hvy: Thanks for the input. I remove the "with_callback" part again. I agree that this "with" preposition was rather confusing.

hvy:
Updated the function name given https://github.com/optuna/optuna/pull/2811#discussion_r677046846. PTAL.

nzw0301 (approved):

not522 (approved):
LGTM.

PR #2817, opened 2021-07-21
Optuna ask and tell CLI options

hvy:
## Motivation

It is sometimes useful to be able to utilize the ask and tell APIs of Optuna via the CLI, or without having to write any Python code.

This initial design is meant to be _experimental_. The search space definition format follows the JSON-`BaseDistribution` conversions defined in https://github.com/optuna/optuna/blob/master/optuna/distributions.py which is quite verbose, but intentionally avoids introducing any new concepts or formats. Similarly, the sampler specification is highly experimental. The current design only supports JSON serializable Python arguments. It does for instance not allow configuring which secondary sampler to fall back to with the `CmaEsSampler`, nor utilize constrained optimization (`constraints_func`) with the `NSGAIISampler`.

## Description of the changes

Introduces `optuna ask` and `optuna tell` CLI commands.

### `optuna ask`

```console
$ optuna ask --storage sqlite:///foo.db --study-name mystudy --sampler TPESampler --sampler-kwargs '{"multivariate": true}' --search-space '{"x": {"name": "UniformDistribution", "attributes": {"low": 0.0, "high": 1.0}}, "y": {"name": "UniformDistribution", "attributes": {"low": 1.0, "high": 100.0}}}' --direction minimize --out yaml > out.yaml
```

```console
$ yq e .trial.number out.yaml
0
$ yq e .trial.params out.yaml
x: 0.9792623438714788
y: 75.24708350450462
```

### `optuna tell`

```console
$ optuna tell --storage sqlite:///foo.db --study-name mystudy --trial-number 0 --values 1.0
```

## TODO

- [x] Address existing TODOs in the code
- [x] Add tests
- [x] Mark experimental

### DIFF
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -10,6 +10,7 @@ c.f. https://docs.openstack.org/cliff/latest/user/demoapp.html#setup-py
 from argparse import ArgumentParser  # NOQA
 from argparse import Namespace  # NOQA
 from importlib.machinery import SourceFileLoader
+import json
 import logging
 import sys
 import types
@@ -23,6 +24,7 @@ from cliff.app import App
 from cliff.command import Command
 from cliff.commandmanager import CommandManager
 from cliff.lister import Lister
+import yaml
 
 import optuna
 from optuna._deprecated import deprecated
@@ -367,6 +369,100 @@ class _StorageUpgrade(_BaseCommand):
                 "Please try updating optuna to the latest version by "
                 "`$ pip install -U optuna`."
             )
+
+
+class _Ask(_BaseCommand):
+    """Ask."""
+
+    def get_parser(self, prog_name: str) -> ArgumentParser:
+
+        # TODO(hvy): Support mulitple directions.
+        parser = super(_Ask, self).get_parser(prog_name)
+        parser.add_argument("--study-name", type=str)
+        parser.add_argument("--storage", type=str)
+        parser.add_argument("--sampler", type=str)
+        parser.add_argument("--direction", type=str, choices=("minimize", "maximize"))
+        parser.add_argument("--sampler-kwargs", type=str)
+        parser.add_argument("--search-space", type=str)
+        parser.add_argument("--out", type=str, choices=("json", "yaml"))
+        return parser
+
+    def take_action(self, parsed_args: Namespace) -> int:
+
+        # TODO(hvy): `parsed_args` input validation.
+        storage_url = _check_storage_url(self.app_args.storage)
+
+        if parsed_args.sampler is not None:
+            if parsed_args.sampler_kwargs is not None:
+                sampler_kwargs = json.loads(parsed_args.sampler_kwargs)
+            else:
+                sampler_kwargs = {}
+
+            sampler_cls = getattr(optuna.samplers, parsed_args.sampler)
+            sampler = sampler_cls(**sampler_kwargs)
+
+        search_space_json = json.loads(parsed_args.search_space)
+        search_space = {
+            name: optuna.distributions.json_to_distribution(json.dumps(j))
+            for name, j in search_space_json.items()
+        }
+
+        # TODO(hvy): Do not require direction.
+        if parsed_args.direction is None:
+            raise ValueError("Direction must be given.")
+
+        study = optuna.create_study(
+            storage=storage_url,
+            study_name=parsed_args.study_name,
+            load_if_exists=True,
+            sampler=sampler,
+            direction=parsed_args.direction,
+        )
+        trial = study.ask(fixed_distributions=search_space)
+
+        data = {
+            "number": trial.number,
+            "params": trial.params,
+        }
+
+        if parsed_args.out is None:
+            print(data)
+        elif parsed_args.out == "json":
+            print(json.dumps(data))
+        elif parsed_args.out == "yaml":
+            print(yaml.dump(data))
+        else:
+            assert False
+
+        return 0
+
+
+class _Tell(_BaseCommand):
+    """Tell."""
+
+    def get_parser(self, prog_name: str) -> ArgumentParser:
+
+        # TODO(hvy): Support multiple values.
+        # TODO(hvy): Support states other than `TrialState.COMPLETE`.
+        parser = super(_Tell, self).get_parser(prog_name)
+        parser.add_argument("--study-name", type=str)
+        parser.add_argument("--storage", type=str)
+        parser.add_argument("--number", type=int)
+        parser.add_argument("--value", type=float)
+        return parser
+
+    def take_action(self, parsed_args: Namespace) -> int:
+
+        # TODO(hvy): `parsed_args` input validation.
+        storage_url = _check_storage_url(self.app_args.storage)
+
+        study = optuna.load_study(
+            storage=storage_url,
+            study_name=parsed_args.study_name,
+        )
+        study.tell(trial=parsed_args.number, values=parsed_args.value)
+
+        return 0
 
 
 class _OptunaApp(App):
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -379,13 +379,37 @@ class _Ask(_BaseCommand):
     def get_parser(self, prog_name: str) -> ArgumentParser:
 
         parser = super(_Ask, self).get_parser(prog_name)
-        parser.add_argument("--study-name", type=str)
-        parser.add_argument("--direction", type=str, choices=("minimize", "maximize"))
-        parser.add_argument("--directions", type=str, nargs="+", choices=("minimize", "maximize"))
-        parser.add_argument("--sampler", type=str)
-        parser.add_argument("--sampler-kwargs", type=str)
-        parser.add_argument("--search-space", type=str)
-        parser.add_argument("--out", type=str, choices=("json", "yaml"), default="json")
+        parser.add_argument("--study-name", type=str, help="Name of study.")
+        parser.add_argument(
+            "--direction",
+            type=str,
+            choices=("minimize", "maximize"),
+            help="Direction of optimization.",
+        )
+        parser.add_argument(
+            "--directions",
+            type=str,
+            nargs="+",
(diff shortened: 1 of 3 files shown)


hvy:
Addressed the remaining TODOs. Removed the draft and made it ready for review. 🙇

Crissman:
@himkt Do you have time to review this week?

optuna/cli.py
```diff
@@ -23,11 +24,14 @@
 from cliff.command import Command
 from cliff.commandmanager import CommandManager
 from cliff.lister import Lister
+import yaml
```
hvy, line 27: I for some reasons though that `yaml` is part of the standard library, although it's not. Open for comments about this dependency and if we should leave it for future discussions, skipping the yaml support here.

Crissman:
How were you thinking of documenting this functionality? Do you want to do that in a separate PR?

Crissman (changes requested):
A few nits. Otherwise, LGTM!

optuna/cli.py
```diff
@@ -369,6 +373,122 @@ def take_action(self, parsed_args: Namespace) -> None:
             )
 
 
+class _Ask(_BaseCommand):
+    """Create a new trial and suggest parameters."""
+
+    def get_parser(self, prog_name: str) -> ArgumentParser:
+
+        parser = super(_Ask, self).get_parser(prog_name)
```
himkt, line 381: AFAIK, Optuna CLI doesn't provide help messages for subcommands.
(I think it may not be the problem because currently implemented commands are relatively easy to use)

However, arguments of `ask` and `tell` are complicated for me. I want a help message for them.
Please correct me if there is a way to show help in Optuna CLI.

<details>

```sh
> optuna --help
usage: optuna [--version] [-v | -q] [--log-file LOG_FILE] [-h] [--debug] [--storage STORAGE]

optional arguments:
  --version            show program's version number and exit
  -v, --verbose        Increase verbosity of output. Can be repeated.
  -q, --quiet          Suppress output except warnings and errors.
  --log-file LOG_FILE  Specify a file to log output. Disabled by default.
  -h, --help           Show help message and exit.
  --debug              Show tracebacks on errors.
  --storage STORAGE    DB URL. (e.g. sqlite:///example.db)

Commands:
  ask            Create a new trial and suggest parameters.
  complete       print bash completion command (cliff)
  create-study   Create a new study.
  dashboard      Launch web dashboard (beta).
  delete-study   Delete a specified study.
  help           print detailed help for another command (cliff)
  storage upgrade  Upgrade the schema of a storage.
  studies        Show a list of studies.
  study optimize  Start optimization of a study. Deprecated since version 2.0.0.
  study set-user-attr  Set a user attribute to a study.
  tell           Finish a trial created with the ask command.


(venv) 2021-07-27 11:17:09 [~/work/github.com/himkt/optuna] @ cli-ask-tell ●
> optuna ask --help
usage: optuna [--version] [-v | -q] [--log-file LOG_FILE] [-h] [--debug] [--storage STORAGE]

optional arguments:
  --version            show program's version number and exit
  -v, --verbose        Increase verbosity of output. Can be repeated.
  -q, --quiet          Suppress output except warnings and errors.
  --log-file LOG_FILE  Specify a file to log output. Disabled by default.
  -h, --help           Show help message and exit.
  --debug              Show tracebacks on errors.
  --storage STORAGE    DB URL. (e.g. sqlite:///example.db)

Commands:
  ask            Create a new trial and suggest parameters.
  complete       print bash completion command (cliff)
  create-study   Create a new study.
  dashboard      Launch web dashboard (beta).
  delete-study   Delete a specified study.
  help           print detailed help for another command (cliff)
  storage upgrade  Upgrade the schema of a storage.
  studies        Show a list of studies.
  study optimize  Start optimization of a study. Deprecated since version 2.0.0.
  study set-user-attr  Set a user attribute to a study.
  tell           Finish a trial created with the ask command.
```

</details>

optuna/cli.py
Crissman: This is an obscure usage of Optuna -- ask/tell from CLI. If possible, we should avoid adding a dependency for this. Is there a way to make this dependency checked and called only if they use the `--out=yaml` argument, and then tell them to install?

optuna/cli.py
```diff
@@ -369,6 +373,122 @@ def take_action(self, parsed_args: Namespace) -> None:
...
+        parser.add_argument("--study-name", type=str)
+        parser.add_argument("--direction", type=str, choices=("minimize", "maximize"))
+        parser.add_argument("--directions", type=str, nargs="+", choices=("minimize", "maximize"))
+        parser.add_argument("--sampler", type=str)
+        parser.add_argument("--sampler-kwargs", type=str)
+        parser.add_argument("--search-space", type=str)
```
himkt, line 387: I want to know why `search-space` is not required.

optuna/cli.py
himkt: I agree with you that `yaml` is almost the standard library.
But, I also think if it is used explicitly, it is added to `get_install_requires` in `setup.py`.

himkt (commented):
This feature is really interesting to me and I'm happy to see the PR!
Basically LGTM, but I left some comments.

> The current design only supports JSON serializable Python arguments. It does for instance not allow configuring which secondary sampler to fall back to with the CmaEsSampler, nor utilize constrained optimization (constraints_func) with the NSGAIISampler.

It makes sense. It would be informative to document a current limitation somewhere in the codebase. What do you think?

optuna/cli.py
hvy: Good point. The reason is that I tried to keep behaviors consistent. It's allowed for a trial not to be associated with any parameters. I.e. it's fine to not call `Trial.suggest_` in an objective function, and requiring a search space here I though would be inconsistent with this.

optuna/cli.py
hvy: > But, I also think if it is used explicitly, it is added to get_install_requires in setup.py.

Do you mean that we're depending on `yaml` already? I couldn't find it in https://github.com/optuna/optuna/blob/master/setup.py. Or did you mean that we could?

> Is there a way to make this dependency checked and called only if they use the `--out=yaml`

This sounds good. If you're okay with that, I'll update the code accordingly!

himkt: Sorry for the ambiguous comment. I was just wondering if we should add `yaml` to setup.py if we use it explicitly in cli.py.

> Is there a way to make this dependency checked and called only if they use the --out=yaml

But I also think this approach is good. 👍

optuna/cli.py
```diff
@@ -369,6 +373,122 @@ def take_action(self, parsed_args: Namespace) -> None:
...
+        parser.add_argument("--out", type=str, choices=("json", "yaml"), default="json")
+        return parser
+
+    def take_action(self, parsed_args: Namespace) -> int:
+
+        warnings.warn(
+            "ask is an experimental CLI command. The interface can change in the future.",
```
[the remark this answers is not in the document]
hvy: Thanks, fixed!

optuna/cli.py
```diff
@@ -369,6 +373,122 @@ def take_action(self, parsed_args: Namespace) -> None:
...
+        parser.add_argument("--study-name", type=str)
+        parser.add_argument("--storage", type=str)
+        parser.add_argument("--trial-number", type=int)
+        parser.add_argument("--values", type=float, nargs="+")
+        parser.add_argument("--state", type=str, default="complete")
+        return parser
+
+    def take_action(self, parsed_args: Namespace) -> int:
+        warnings.warn(
+            "tell is an experimental CLI command. The interface can change in the future.",
```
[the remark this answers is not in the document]
hvy: Fixed!

optuna/cli.py
hvy: Okay, so I just noticed that Optuna already depends on `yaml` (i.e. PyYAML) via cliff https://github.com/openstack/cliff/blob/117a1005dc208b55aba93652856c8001b28a48ce/requirements.txt. Optuna's CLI won't work without `yaml` independent of this PR. We can either include PyYAML in setup.py to be explicit, or perhaps just leave the code in this PR as is, maybe with a comment.

optuna/cli.py
himkt: Thank you for the clarifying. I undepstand and agree with you to keep a consistency.

optuna/cli.py
hvy: You're right about the current situation. I think this PR addresses the problem. Do you want to have a look at it? https://github.com/optuna/optuna/pull/2823

optuna/cli.py
hvy: By the way, we could require the argument, forcing the user to pass an empty dictionary to emulate the above situation, but I'm not sure if that'd make any users happy.

hvy:
> How were you thinking of documenting this functionality? Do you want to do that in a separate PR?

Thanks for pointing that out. I simply forgot to document the arguments. Should be fixed now.

hvy:
PTAL.

optuna/cli.py
himkt: Thank you so much for the investigation and patching! I approved the PR. 👍

optuna/cli.py
himkt: I also think it doesn't make users happy. The current implementation feels good to me.

optuna/cli.py
himkt: Thank you for digging!

Both are OK for me but I prefer adding it to setup.py explicitly, because dependency checking always passes (and it feels to be no worth checking) if clif (Optuna's core dependency) requires PyYAML.

hvy: Added to setup.py. We might want to be careful about using `yaml` elsewhere, outside the context of cliff/CLI but I think it's fine otherwise.

himkt (approved):
LGTM

Crissman (approved):
LGTM. Thanks!

hvy:
Thanks both of you for the prompt review!

PR #2818, opened 2021-07-21
Preinstall RDB binding Python libraries in Docker image

hvy:
## Motivation

As seen here https://github.com/optuna/optuna/blob/v2.8.0/.github/workflows/tests-rdbstorage.yml#L89, several Python libraries are necessary in order to utilize the `RDBStorage` in Optuna (with MySQL). It would be good to have these libraries pre-installed in our Docker images.

Motivation is related to https://github.com/optuna/optuna/pull/2817. Being able to just grab an image and with a single command obtain parameters is useful.

## Description of the changes

Pre-installs/includes Python libraries required for the `RDBStorage` and especially with MySQL.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,4 +26,7 @@ RUN if [ "${BUILD_TYPE}" = "dev" ]; then \
     fi \
     && pip install ${PIP_OPTIONS} jupyter notebook
 
+# Install RDB bindings.
+RUN pip install ${PIP_OPTIONS} PyMySQL cryptography psycopg2-binary
+
 ENV PIP_OPTIONS ""


Dockerfile
```diff
@@ -26,4 +26,7 @@ RUN if [ "${BUILD_TYPE}" = "dev" ]; then \
     fi \
     && pip install ${PIP_OPTIONS} jupyter notebook
 
+# Install RDB bindings.
+RUN pip install ${PIP_OPTIONS} PyMySQL cryptography psycopg2-binary
```
himkt, line 30: [Question]

Is `cryptography` required?
I know it is installed in `tests-rdbstorage`.

https://github.com/optuna/optuna/blob/e0fd830471cde878e2882c2ba1adaed307edd7c9/.github/workflows/tests-rdbstorage.yml#L89

But tests passed without the library.
https://github.com/himkt/optuna/pull/4/commits/504d24a8e2a3bc508b6d22a62b5c23beedb2a9db

hvy: Right, it's an optional dependency in pymysql for handling MySQL password (e.g. [setup.py](https://github.com/PyMySQL/PyMySQL/blob/master/setup.py#L21), [usage in code](https://github.com/PyMySQL/PyMySQL/blob/6ccbecc1a0dfd04065b081950d2d35b1dac0aaa8/pymysql/_auth.py#L137-L155)). It's suggested practice in my limited understanding. 

Cc. @toshihikoyanase do you know if we can or should remove it?

hvy: 📝  Asked @toshihikoyanase and @himkt. This library is apparently utilized automatically if installed to encrypt traffic. It seems fine to include it in our images.

toshihikoyanase: > This library is apparently utilized automatically if installed to encrypt traffic.

For example, `PyMySQL` uses it internally.
https://github.com/PyMySQL/PyMySQL/blob/6ccbecc1a0dfd04065b081950d2d35b1dac0aaa8/pymysql/_auth.py#L7-L14

hvy: It by the way seems like users will experience errors in case the dependency is missing (and `sha256_password` is specified).

himkt (approved):
LGTM! 😸

HideakiImamura:
@toshihikoyanase Could you continue to review this PR?

toshihikoyanase (approved):
LGTM! Thank you for your PR.

Issue #2455, opened 2021-03-08
Sort categorical values including numbers in `plot_parallel_coordinate`


nzw0301:
## Motivation

When the categorical values contain only numerical values such as int or float values, it would be great that the axis of `plot_parallel_coordinate` is sorted. (see also the example and plots below.)

I guess current both implementations don't sort the suggested values.

## Description

Treat the values as numerical and sort them.

## Additional context (optional)

```python
import optuna


def objective(trial):
    x = trial.suggest_uniform("x", -100, 100)
    y = trial.suggest_categorical("y", list(range(-1, 7)))
    return x ** 2 + y


sampler = optuna.samplers.TPESampler(seed=10)
study = optuna.create_study(sampler=sampler)
study.optimize(objective, n_trials=10)

optuna.visualization.plot_parallel_coordinate(study)
optuna.visualization.matplotlib.plot_parallel_coordinate(study)
```

The following both plots do not sort y-axis

## Plotly
![newplot (3)](https://user-images.githubusercontent.com/7121753/110315896-41d13980-804d-11eb-8d37-56b55986e92d.png)

## Matplotlib
![download (2)](https://user-images.githubusercontent.com/7121753/110315904-44cc2a00-804d-11eb-918c-1b06ca981a1c.png)

---

## Dependencies

- optuna: 2.6.0
- plotly: 4.14.3
- matplotlib 3.3.4

TakuyaInoue-github:
I tackled this issue(but I have to implement test.) I hope you will check out this PR and give me comments for fixes.

PR #2821, opened 2021-07-25
Sort values when the categorical values is numerical in `plot_parallel_coordinate`

TakuyaInoue-github:
## Motivation

For clarity, when every values of axis in `plot_parallel_coordinate` are numeric, it's better to have sorted.
And see also https://github.com/optuna/optuna/issues/2455.

## Description of the changes

When called 
When the method is called with categorical values of numeric, values for plotting is sorted.
After all the values for the plot are determined, change the order of the values according to the order of categorical values. 

## TODO
- [x] Implement logic.
- [x] Implement tests.

### DIFF
--- a/optuna/visualization/_parallel_coordinate.py
+++ b/optuna/visualization/_parallel_coordinate.py
@@ -183,13 +183,7 @@ def _get_parallel_coordinate_plot(
         dims.append(dim)
 
     if len(numeric_cat_params.keys()) != 0:
-        sorted_categorical_vals = (
-            pd.DataFrame()
-            .from_dict(numeric_cat_params)
-            .sort_values(by=list(numeric_cat_params.keys()))
-        )
-        idx = sorted_categorical_vals.index
-
+        idx = list(np.lexsort(list(numeric_cat_params.values()))[::-1])
         for dim in dims:
             dim.update({"values": tuple(np.array(dim["values"])[idx])})
 
--- a/optuna/visualization/_parallel_coordinate.py
+++ b/optuna/visualization/_parallel_coordinate.py
@@ -9,7 +9,6 @@ from typing import List
 from typing import Optional
 
 import numpy as np
-import pandas as pd
 
 from optuna.logging import get_logger
 from optuna.study import Study
--- a/optuna/visualization/_parallel_coordinate.py
+++ b/optuna/visualization/_parallel_coordinate.py
@@ -182,7 +182,7 @@ def _get_parallel_coordinate_plot(
         dims.append(dim)
 
     if len(numeric_cat_params.keys()) != 0:
-        idx = list(np.lexsort(list(numeric_cat_params.values()))[::-1])
+        idx = np.lexsort(list(numeric_cat_params.values())[::-1])
         for dim in dims:
             dim.update({"values": tuple(np.array(dim["values"])[idx])})
 
--- a/optuna/visualization/_parallel_coordinate.py
+++ b/optuna/visualization/_parallel_coordinate.py
@@ -8,6 +8,9 @@ from typing import Dict
 from typing import List
 from typing import Optional
 
+import numpy as np
+import pandas as pd
+
 from optuna.logging import get_logger
 from optuna.study import Study
 from optuna.study._study_direction import StudyDirection
@@ -126,6 +129,8 @@ def _get_parallel_coordinate_plot(
             "range": (min([target(t) for t in trials]), max([target(t) for t in trials])),
         }
     ]
+
+    numeric_cat_params: Dict[str, List[Any]] = {}
     for p_name in sorted_params:
         values = []
         for t in trials:
@@ -150,13 +155,22 @@ def _get_parallel_coordinate_plot(
             }
         elif _is_categorical(trials, p_name):
             vocab: DefaultDict[str, int] = defaultdict(lambda: len(vocab))
-            values = [vocab[v] for v in values]
+
+            if all([isinstance(v, (int, float)) for v in values]):
+                _ = [vocab[v] for v in sorted(values)]
+                values = [vocab[v] for v in values]
+                numeric_cat_params[p_name] = values
+                ticktext = list(sorted(vocab.keys()))
+            else:
+                values = [vocab[v] for v in values]
+                ticktext = list(sorted(vocab.keys(), key=lambda x: vocab[x]))
+
             dim = {
                 "label": p_name if len(p_name) < 20 else "{}...".format(p_name[:17]),
                 "values": tuple(values),
                 "range": (min(values), max(values)),
                 "tickvals": list(range(len(vocab))),
-                "ticktext": list(sorted(vocab.keys(), key=lambda x: vocab[x])),
+                "ticktext": ticktext,
             }
         else:
             dim = {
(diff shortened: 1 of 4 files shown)


optuna/visualization/_parallel_coordinate.py
```diff
@@ -150,13 +155,22 @@ def _target(t: FrozenTrial) -> float:
             }
         elif _is_categorical(trials, p_name):
             vocab: DefaultDict[str, int] = defaultdict(lambda: len(vocab))
-            values = [vocab[v] for v in values]
+
+            if all([isinstance(v, (int, float)) for v in values]):
```
nzw0301, line 159: FYI; `_is_numerical` function does a similar check. https://github.com/optuna/optuna/blob/master/optuna/visualization/_utils.py#L77-L83.

optuna/visualization/matplotlib/_parallel_coordinate.py
```diff
@@ -148,7 +153,14 @@ def _target(t: FrozenTrial) -> float:
             log_param_names.append(p_name)
         elif _is_categorical(trials, p_name):
             vocab = defaultdict(lambda: len(vocab))  # type: DefaultDict[str, int]
-            values = [vocab[v] for v in values]
+
+            if all([isinstance(v, (int, float)) for v in values]):
```
nzw0301, line 157: Ditto (`_is_numerical`).

optuna/visualization/_parallel_coordinate.py
```diff
@@ -167,6 +181,17 @@ def _target(t: FrozenTrial) -> float:
 
         dims.append(dim)
 
+    if len(numeric_cat_params.keys()) != 0:
+        sorted_categorical_vals = (
+            pd.DataFrame()
+            .from_dict(numeric_cat_params)
+            .sort_values(by=list(numeric_cat_params.keys()))
+        )
+        idx = sorted_categorical_vals.index
```
nzw0301, line 190: Could you implement this sort of logic without `pandas` because pandas is an optional dependency? 

By the way, if this sort of logic depends on pandas, then we need to `try_import`. For example, 
https://github.com/optuna/optuna/blob/c6c083d46adfaf626b4ea81f7254f5bc60305fa0/optuna/study/_dataframe.py#L14-L16

nzw0301 (commented):
Thank you for your consecutive pull request! This PR's status is still WIP, but let me give some feedback for future updates.

TakuyaInoue-github:
@nzw0301 Thank you for your feedback.I've addressed to your comments. Furthermore, I implemented some tests. Please check these changes.

optuna/visualization/_parallel_coordinate.py
```diff
@@ -167,6 +181,11 @@ def _target(t: FrozenTrial) -> float:
 
         dims.append(dim)
 
+    if len(numeric_cat_params.keys()) != 0:
+        idx = np.lexsort(list(numeric_cat_params.values())[::-1])
+        for dim in dims:
+            dim.update({"values": tuple(np.array(dim["values"])[idx])})
```
nzw0301, line 187: Could you add comments to explain this logic for other maintainers? I'm afraid that I still do not fully understand this sort of logic.

nzw0301 (changes requested):
Awesome! Basically, looks great! Could change your PR based on my comments?

optuna/visualization/_parallel_coordinate.py
TakuyaInoue-github: np.lexsort consumes the sort key in order from back to front. So I reversed the list to sort from head. 
(https://numpy.org/doc/stable/reference/generated/numpy.lexsort.html)
I'm going to write a comment as above, what do you think?

tests/visualization_tests/matplotlib_tests/test_parallel_coordinate.py
```diff
@@ -86,6 +86,45 @@ def test_plot_parallel_coordinate_categorical_params() -> None:
     assert len(figure.get_lines()) == 0
 
 
+def test_plot_parallel_coordinate_categorical_numeric_params() -> None:
+    # Test with categorical params that can be interpeted numeric.
```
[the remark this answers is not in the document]
TakuyaInoue-github: It looks great.  I'll fix the comment as you suggest.

optuna/visualization/_parallel_coordinate.py
nzw0301: Sorry, could you also explain `dim.update({"values": tuple(np.array(dim["values"])[idx])})`?

TakuyaInoue-github:
@nzw0301 I appreciate to your appropriate review. I address your comments. Please confirm this changes.

tests/visualization_tests/test_parallel_coordinate.py
```diff
@@ -129,6 +129,67 @@ def test_plot_parallel_coordinate_categorical_params() -> None:
...
+def test_plot_parallel_coordinate_categorical_numeric_params() -> None:
+    # Test with categorical params that can be interpreted as numeric params.
+    study_categorical_params = create_study()
+    study_categorical_params.add_trial(
+        create_trial(
+            value=0.0,
+            params={"category_a": 2, "category_b": 2},
+            distributions={
+                "category_a": CategoricalDistribution((1, 2, 10)),
+                "category_b": CategoricalDistribution((-1, 1, 2)),
```
nzw0301, line 141: Could you add another parameter sampled from another distribution except for categorical distribution? This is because the introduced [sort logic](https://github.com/optuna/optuna/pull/2821/files#diff-fb4c806a38d57b2092f804687882f64ad7ad51616bb06d183acdd72acee671d4R191-R193) affects such values.

nzw0301 (commented):
Brilliant! Sorry, could you look at my last comment on the test case?

tests/visualization_tests/test_parallel_coordinate.py
TakuyaInoue-github: Absolutely right. I implemented a test with sample from multiple distribution. Please check it right.

nzw0301: great! thanks!

nzw0301:
Thank you for your quick update and for pinging me. I'll review the update in a few days.

tests/visualization_tests/test_parallel_coordinate.py
nzw0301: Ah, I think the added test case, `test_plot_parallel_coordinate_sample_from_multiple_distribution` is enough. How about removing this `test_plot_parallel_coordinate_categorical_numeric_params` function and rename `test_plot_parallel_coordinate_sample_from_multiple_distribution` with `test_plot_parallel_coordinate_with_categorical_numeric_params`?

nzw0301: Ah, I think the added test case, `test_plot_parallel_coordinate_sample_from_multiple_distribution` is enough. How about removing this `test_plot_parallel_coordinate_categorical_numeric_params` function and rename `test_plot_parallel_coordinate_sample_from_multiple_distribution` with `test_plot_parallel_coordinate_with_categorical_numeric_params`?

nzw0301 (changes requested):
Thanks! The added test is supposed to be redundant. Let merge the two test cases.

nzw0301:
@HideakiImamura The CI of `document artifact` is not in the passed checks. Do you know why this happens?

tests/visualization_tests/test_parallel_coordinate.py
TakuyaInoue-github: Thanks your comment! I wonder if we need the case of sort that sort key are  multiple columns. If you need it , I will implement the case to the merged test. What do you think?

nzw0301: Yes!! The proposal sounds great. I appreciate you for implementing your additional test cases.

TakuyaInoue-github: Sorry for my late. I merged two test and added the case that proposed previous comment.

nzw0301 (approved):
LGTM! Thank you for contributing long-running PR!

HideakiImamura (approved):
Sorry for the late review. Looks perfect! LGTM!

PR #2830, opened 2021-07-30
Fix `bounds`' shape in the document

nzw0301:
## Motivation

I think that the current  `bounds`' shape is swapped.

After `bounds` is initialised by [here](https://github.com/optuna/optuna/blob/8ee931a77343f64e7ba270f0d37d7bb611dbe221/optuna/_transform.py#L172), it is [transposed](https://github.com/optuna/optuna/blob/8ee931a77343f64e7ba270f0d37d7bb611dbe221/optuna/integration/botorch.py#L498) in `sample_relative` in botorch sampler. Thus `bounds`' shape is `(2, n_params)` that is passed to `qei_candidates_func`. Indeed, according to the [botorch documentation](https://botorch.org/api/optim.html#module-botorch.optim.optimize), `bounds`' shape is supposed to be `(2, n_param)`.

## Description of the changes

Fix the shape and update the document.

### DIFF
--- a/optuna/integration/botorch.py
+++ b/optuna/integration/botorch.py
@@ -76,8 +76,8 @@ def qei_candidates_func(
             constraints. A constraint is violated if strictly larger than 0. If no constraints are
             involved in the optimization, this argument will be :obj:`None`.
         bounds:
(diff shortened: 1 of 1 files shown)


hvy (approved):
You're right. The bounds are transposed prior to the callback to match the BoTorch practices. Nice catch and LGTM!

PR #2833, opened 2021-08-02
Simplify documentation of `FrozenTrial`

hvy:
## Motivation

There was a report about a note in the documentation of `FrozenTrial` being a bit confusing. 
The note addressed the fact that the class is actually not "frozen" and can be mutated.
However, I expect it to be intuitive to most users given the dynamic nature of Python.

## Description of the changes

Removes the note from the docs of `FrozenTrial` that lead to the above report.


hvy:
Added a short description back on request, to address users being mislead by the name.

HideakiImamura (approved):
LGTM.

PR #2837, opened 2021-08-03
Remove storage URI from `ask` and `tell` CLI subcommands

hvy:
## Motivation

Storage URI may contain sensitive information which ideally isn't logged by `optuna {ask,tell}`.

## Description of the changes

Removes the storage URI as well as the study name from the log. The study name can be kept but was removed to make the log somewhat more consistent with the optimization log and the study name is already logged when loading the study.

### DIFF
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -457,10 +457,7 @@ class _Ask(_BaseCommand):
             }
         }
 
-        self.logger.info(
-            f"Asked trial {trial.number} with parameters {trial.params} in study "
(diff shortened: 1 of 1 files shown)


himkt (approved):
LGTM.

~Is milestone correct (2.10.1?)? I'm wondering if it would be v2.9.1.~ In the milestone page (https://github.com/optuna/optuna/milestone/38) I can see it is v2.9.1.

not522 (approved):
LGTM

PR #2846, opened 2021-08-05
Fix `datetime_complete` in `_CachedStorage`

not522:
## Motivation
The cached `datetime_complete` is inconsistent with the one in the RDB. This PR fixes the bug.

## Description of the changes
- Fix `datetime_complete` in `_CachedStorage`
- Add test

### DIFF
--- a/tests/storages_tests/test_cached_storage.py
+++ b/tests/storages_tests/test_cached_storage.py
@@ -36,7 +36,7 @@ def test_set_trial_state() -> None:
     storage.set_trial_state(trial_id, TrialState.COMPLETE)
 
     cached_trial = storage.get_trial(trial_id)
-    base_trial = storage._backend.get_trial(trial_id)
+    base_trial = base_storage.get_trial(trial_id)
 
     assert cached_trial == base_trial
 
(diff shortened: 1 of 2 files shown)


HideakiImamura:
@hvy @himkt Could you review this PR?

tests/storages_tests/test_cached_storage.py
```diff
@@ -28,6 +28,19 @@ def test_create_trial() -> None:
...
 
+def test_set_trial_state() -> None:
+    base_storage = RDBStorage("sqlite:///:memory:")
+    storage = _CachedStorage(base_storage)
+    study_id = storage.create_new_study("test-study")
+    trial_id = storage.create_new_trial(study_id)
+    storage.set_trial_state(trial_id, TrialState.COMPLETE)
+
+    cached_trial = storage.get_trial(trial_id)
+    base_trial = storage._backend.get_trial(trial_id)
```
hvy, line 39: This is definitely not a strong opinion but instead of accessing a private attribute, I think you can `load_study` to create a new study with a cleared cache.

hvy (approved):
Thanks for the fix, this does indeed look like a bug. LGTM!

tests/storages_tests/test_cached_storage.py
not522: It's not your suggested approach, but I updated the PR to use `base_storage.get_trial(trial_id)` for removing accessing to a private attribute.

himkt (approved):
Nice catch :+1: and LGTM!

Issue #2628, opened 2021-04-26
[RFC] CLI enhancement proposals


hvy:
## Motivation

The [`optuna` CLI](https://optuna.readthedocs.io/en/stable/reference/cli.html) provides various functionality such as checking the optuna version (`optuna --version`), operations on studies such as create/delete/list, etc. The CLI could be enriched. Below are ideas on how. They're all open, at least in terms of design, so please feel free to add ideas or criticize.

## Description

The CLI could be improved with the following functionalities.

### Listing trials

E.g. `optuna trials --csv -o mystudy.csv` to dump the trials from a study as a CSV file. This could also be a subcommand such as `optuna study trials ...`, `optuna study best-trials ...`.

### Showing a single trial

Could perhaps be covered by above.

### Additional operations on the study

E.g. `optuna study user-attrs` to show study user attributes. Note that the CLI currently allows setting user attributes, and get/set could potentially be made symmetric.

### Visualization

E.g. `optuna study visualization contour ...` or `optuna visualization contour ...` to generate a plot image.

## Alternatives (optional)

We write the above logic in Python.

hvy:
📝  Report CLI subcommand?

hvy:
Some minor improvements to the user experience https://github.com/optuna/optuna/pull/2823.

PR #2847, opened 2021-08-05
Introduce trials CLI

not522:
## Motivation
This PR introduces a way to get trial information using CLI. (proposed by #2628)

Current implementation outputs `study.trials_dataframe()`. I'm not sure which column we should show. Especially, I think `system_attrs` is redundant.

#2846 will fix the test failure.

## Description of the changes
- Add trials subcommands: `optuna trials`, `optuna best-trial`, `optuna best-trials`


HideakiImamura:

not522:
I updated the PR to remove the pandas dependency, which is the optional requirement.
- Stop the `system_attrs` output
- Modify `_dataframe.py`
- Add branched search space tests

HideakiImamura (commented):
Sorry for the late review. The changes basically LGTM. I have some minor comments. Could you take a look?

hvy:
Thanks a lot for this PR. Looks great and I think this'll make Optuna easier to use, especially with the Ask and tell CLI subcommands.

While it's convenient to have several options from cliff out of the box, I have some questions.
First, some options are redundant. For instance `--print-empty` doesn't really work for `best-trial`, Optuna will raise an error if the study is empty. This might not be such a big issue but I think it could be improved (e.g. option removed). 
Second, for formats such as JSON and YAML, it's be great if we can nest keys, especially parameters. E.g. get all parameters for a certain trial as a "dictionary" using `jq -e '.[0].params' mytrials.json`. Do you think this makes sense or is ever made possible with cliff's format?
I'm also wondering if we should make these subcommands experimental prior to having these discussions resolved. What do you think?

optuna/cli.py
```diff
@@ -187,6 +192,110 @@ def take_action(self, parsed_args: Namespace) -> Tuple[Tuple, Tuple[Tuple, ...]]
...
+def _format_trial_values(
+    record: Dict[Tuple[str, str], Any], columns: List[Tuple[str, str]]
+) -> List:
+    row: List = []
+    for column in columns:
+        value = record.get(column, np.nan)
+        if isinstance(value, (int, float)):
+            row.append(value)
+        elif isinstance(value, datetime.datetime):
+            row.append(value.strftime("%Y-%m-%d %H:%M:%S"))
```
hvy, line 204: A detail but how about making this format a module local (similar to `dashboard.py`)? The same format is also `_Studies`.

optuna/cli.py
```diff
@@ -187,6 +192,110 @@ def take_action(self, parsed_args: Namespace) -> Tuple[Tuple, Tuple[Tuple, ...]]
         return self._study_list_header, tuple(rows)
 
 
+def _format_trial_values(
+    record: Dict[Tuple[str, str], Any], columns: List[Tuple[str, str]]
+) -> List:
+    row: List = []
```
hvy, line 198: Is this type hint necessary for mypy? It doesn't seem to add any information.

optuna/study/_dataframe.py
```diff
@@ -49,16 +39,13 @@ def _trials_dataframe(
...
-
+    records = []
+    for trial in study.get_trials(deepcopy=False):
         record = {}
         for attr, df_column in attrs_to_df_columns.items():
             value = getattr(trial, attr)
             if isinstance(value, TrialState):
-                # Convert TrialState to str and remove the common prefix.
-                value = str(value).split(".")[-1]
+                value = value.name
```
hvy, line 48: 👍

optuna/study/_dataframe.py
```diff
@@ -77,20 +64,39 @@ def _create_record_and_aggregate_column(
...
+    _imports.check()
+
+    # If no trials, return an empty dataframe.
+    if len(study.get_trials(deepcopy=False)) == 0:
+        return pd.DataFrame()
+
+    if "value" in attrs and study._is_multi_objective():
+        attrs = tuple("values" if attr == "value" else attr for attr in attrs)
+
+    records, columns = _create_records_and_aggregate_column(study, attrs)
```
hvy, line 95: Good refactoring having this data creation logic independent of `pandas`.

not522:
Thank you for your reviews. I updated the PR to follow review comments.
I agree that the output format should be more flexible. I want to implement it in a follow-up PR to keep this PR simple.

HideakiImamura (approved):
Thanks for the update! LGTM!

hvy (approved):
> I agree that the output format should be more flexible. I want to implement it in a follow-up PR to keep this PR simple.

Sounds good and LGTM!
Let me merge this PR as it has two approvals and let's try to incorporate the follow up PR(s) in the same v2.10 release.

PR #2672, opened 2021-05-14
[WIP] TensorFlow with Python 3.9

hvy:
## Motivation

Relax Python version constraint with tensorflow related code. 

See https://github.com/tensorflow/tensorflow/releases/tag/v2.5.0

## Description of the changes

See above.

PR #2851, opened 2021-08-10
Avoid keras 2.6.0

nzw0301:
## Motivation

The current master branch fails in CI because of the latest release of keras 2.6.0. This PR is a hotfix.

Failed CI example: https://github.com/optuna/optuna/pull/2847/checks?check_run_id=3286012452

## Description of the changes

Introduce the version constraint of keras: `keras<2.6.0`.

PR #2530, opened 2021-03-27
Supporting Python 3.9 with integration modules and optional dependencies

0x41head:
## Motivation

#2034

PR #2852, opened 2021-08-12
Drop tensorflow constraints

nzw0301:
## Motivation

TensorFlow and Keras 2.6.0 have been released. Since this release, `tf.keras` uses [`keras-team/keras`](https://github.com/keras-team/keras) instead of the keras under the TensorFlow repo. 

## Description of the changes

Remove the constraints introduced by 

- https://github.com/optuna/optuna/pull/2851
- https://github.com/optuna/optuna/pull/2674
- https://github.com/optuna/optuna/pull/2530

### DIFF
--- a/.github/workflows/tests-integration.yml
+++ b/.github/workflows/tests-integration.yml
@@ -60,12 +60,6 @@ jobs:
         if [ ${{ matrix.python-version }} = 3.6 ]; then
           pytest tests/integration_tests \
             --ignore tests/integration_tests/test_botorch.py
-        elif  [ ${{ matrix.python-version }} = 3.9 ]; then
-          pytest tests/integration_tests \
-            --ignore tests/integration_tests/test_keras.py \
-            --ignore tests/integration_tests/test_tensorflow.py \
-            --ignore tests/integration_tests/test_tfkeras.py \
-            --ignore tests/integration_tests/test_tensorboard.py
         else
           pytest -s tests/integration_tests
         fi
--- a/setup.py
+++ b/setup.py
@@ -101,12 +101,7 @@ def get_extras_require() -> Dict[str, List[str]]:
             "scikit-learn>=0.24.2",
             "scikit-optimize",
             "xgboost",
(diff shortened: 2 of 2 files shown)


himkt (approved):
I confirmed that tests passed with `tensorflow==2.6.0`.

nzw0301:
I realised that https://github.com/optuna/optuna/pull/2672/ also resolves the same issue...

hvy:
Thanks for the heads up. 
Let me close https://github.com/optuna/optuna/pull/2672 which is somewhat stale in favor of this PR.

hvy:
Leaving some notes from a discussion I had with @nzw0301 .

- `import keras` should be avoided and one should instead import keras via `tensorflow`.
- Whether we should keep both `KerasPruningCallback` (currently deprecated) and `TFKerasPruningCallback` has been discussed before and is still an open discussion. It can be discussed outside the scope of this PR however.

hvy (commented):
The changes LGTM. However, with these changes, we'll basically start supporting the latest tensorflow with our callbacks. My only concern is that `KerasPruningCallback` may start to break given the above comment in a way that's difficult to fix without reintroducing these version constraints (in which case we'll need to test `KerasPruningCallback` and `TFKerasPruningCallback` against different versions of `keras` and `tensorflow`). What do you think @nzw0301 ?

nzw0301:
Thank you for sharing your perspective and review!

In my understanding, Keras has required TensorFlow explicitly since [Keras v2.4.0](https://github.com/keras-team/keras/releases/tag/2.4.0), which was released in June 2020. So personally, `TFKerasPruningCallback` is supposed to be enough and `KerasPruningCallback` might be removed in the next major version of Optuna (v3.0.0). Before that, we can fix the versions of Keras and TensorFlow 2.6.0 if a newer keras breaks instead of reintroducing tests for multiple keras versions.

hvy (approved):
Sounds good and LGTM!

Issue #2788, opened 2021-07-06
Follow-up of MLflowCallback


nzw0301:
## Motivation
Thanks to the great improvement by https://github.com/optuna/optuna/pull/2670, optuna provides a smoother connection to MLFLow via `MLflowCallback`. But we still have a TODO list to improve the codebase and its feature mentioned in the review of https://github.com/optuna/optuna/pull/2670.

## TODO tasks discussed in https://github.com/optuna/optuna/issues/2788#issuecomment-898890114 to close this issue

- [x] General code/docstring fixes mentioned in #2670 TODO list (https://github.com/optuna/optuna/pull/2855)
- [x] Introduction of multi-objective support  (https://github.com/optuna/optuna/pull/2863)
- [ ] Refactoring interface to be more similar to WeightsAndBiasesCallback (+ decoupling study name from mlflow experiment) (https://github.com/optuna/optuna/pull/2912)

Please see also the following `Description` for more detailed contexts.

---

## Description

Please see the discussion in the review. But we can separately work on this issue as follows:

- support multi-objective https://github.com/optuna/optuna/pull/2670#discussion_r633204507
- improve the docstring.
- improve the tag truncation https://github.com/optuna/optuna/pull/2670#discussion_r650478298

nzw0301:
Hi @rafarui thank you for your feedback! Indeed, when we introduce `WeightsAndBiasesCallback` recently, it has [`wandb_kwargs`](https://optuna.readthedocs.io/en/latest/reference/generated/optuna.integration.WeightsAndBiasesCallback.html#optuna.integration.WeightsAndBiasesCallback) that are used for the initialisation of wandb. So we can add a similar argument to `MLflowCallback` for a more flexible interface as you mentioned.

xadrianzetx:
Hi, I will work on that in following order (separate PRs):

* General code/docstring fixes mentioned in [#2670 TODO list](https://github.com/optuna/optuna/pull/2670#pullrequestreview-682465454)
* Introduction of multi-objective support
* Refactoring interface to be more similar to `WeightsAndBiasesCallback` (+ decoupling study name from mlflow experiment)

This order should keep things clear, let me know if you agree @nzw0301.

nzw0301:
Thank you for sharing the plan! It sounds perfect.

PR #2344, opened 2021-02-14
Consistently use `suggest_float`

crcrpar:
In the pursuit of the comprehensive use of  `suggest_float` instead of `suggest_( | log | discrete)uniform`.

PR #2863, opened 2021-08-19
Support multi-objective optimization in `MLflowCallback`

xadrianzetx:
## Motivation
This PR introduces support for tracking multi-objective optimization in MLflow UI by extending functionality of `MLflowCallback`. It's a second task in a list of improvements to this callback, as described in #2788.

## Description of the changes
* [x] Implement support for multi-objective optimization in `MLflowCallback`
* [ ] Update docstrings
* [ ] Write tests

### DIFF
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -1,6 +1,7 @@
 from typing import Any
 from typing import Callable
 from typing import Dict
+from typing import List
 from typing import Optional
 from typing import Sequence
 from typing import Union
@@ -125,7 +126,7 @@ class MLflowCallback(object):
     def __init__(
         self,
         tracking_uri: Optional[str] = None,
-        metric_name: str = "value",
+        metric_name: Union[str, List[str]] = "value",
         nest_trials: bool = False,
         tag_study_user_attrs: bool = False,
     ) -> None:
@@ -147,8 +148,8 @@ class MLflowCallback(object):
             nested=self._nest_trials,
         ):
 
-            # This sets the metric for MLflow.
-            self._log_metric(trial.value)
+            # This sets the metrics for MLflow.
+            self._log_metrics(trial.values)
 
             # This sets the params for MLflow.
             self._log_params(trial.params)
@@ -207,7 +208,7 @@ class MLflowCallback(object):
             study: Study to be tracked.
         """
 
-        tags: Dict[str, str] = {}
+        tags: Dict[str, Union[str, List[str]]] = {}
         tags["number"] = str(trial.number)
         tags["datetime_start"] = str(trial.datetime_start)
 
@@ -217,8 +218,9 @@ class MLflowCallback(object):
         if trial.state.is_finished():
             tags["state"] = trial.state.name
 
-        # Set study direction.
-        tags["direction"] = study.direction.name
+        # Set study directions.
+        directions = [d.name for d in study.directions]
+        tags["direction"] = directions if len(directions) != 1 else directions[0]
 
         tags.update(trial.user_attrs)
         distributions = {(k + "_distribution"): str(v) for (k, v) in trial.distributions.items()}
@@ -241,14 +243,36 @@ class MLflowCallback(object):
         # This sets the tags for MLflow.
         mlflow.set_tags(tags)
 
-    def _log_metric(self, value: Union[float, None]) -> None:
+    def _log_metrics(self, values: List[Union[float, None]]) -> None:
         """Log the trial result as metric to MLflow.
 
         Args:
             value: Result of trial.
         """
-        trial_value = value if value is not None else float("nan")
-        mlflow.log_metric(self._metric_name, trial_value)
+
+        if isinstance(self._metric_name, list):
+            if len(self._metric_name) != len(values):
+                raise RuntimeError(
+                    "Running multi-objective optimization "
+                    "with {} objective values, but {} names specified. "
+                    "Match objective values and names, or use default broadcasting.".format(
+                        len(values), len(self._metric_name)
+                    )
+                )
+
+            else:
+                names = self._metric_name
+
+        else:
+            if len(values) > 1:
+                # broadcast default name for multi-objective optimization
+                names = ["{}_{}".format(self._metric_name, i) for i in range(len(values))]
+
+            else:
+                names = [self._metric_name]
+
+        metrics = {name: val or float("nan") for name, val in zip(names, values)}
+        mlflow.log_metrics(metrics)
 
     @staticmethod
     def _log_params(params: Dict[str, Any]) -> None:
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -191,6 +191,7 @@ class MLflowCallback(object):
         Args:
             study: Study to be tracked in MLflow.
         """
+
         # This sets the `tracking_uri` for MLflow.
         if self._tracking_uri is not None:
             mlflow.set_tracking_uri(self._tracking_uri)
@@ -205,6 +206,7 @@ class MLflowCallback(object):
             trial: Trial to be tracked.
             study: Study to be tracked.
         """
+
         tags: Dict[str, str] = {}
         tags["number"] = str(trial.number)
         tags["datetime_start"] = str(trial.datetime_start)
@@ -255,4 +257,5 @@ class MLflowCallback(object):
         Args:
             params: Trial params.
         """
+
         mlflow.log_params(params)
--- a/tests/integration_tests/test_mlflow.py
+++ b/tests/integration_tests/test_mlflow.py
@@ -285,7 +285,7 @@ def test_log_metric(tmpdir: py.path.local) -> None:
     mlflc._initialize_experiment(study)
 
     with mlflow.start_run():
-        mlflc._log_metric(metric_value)
+        mlflc._log_metrics([metric_value])
 
     mlfl_client = MlflowClient(tracking_file_name)
     experiments = mlfl_client.list_experiments()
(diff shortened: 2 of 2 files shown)


nzw0301:
Thank you for your consecutive PR! I'll look at the PR in a few days.

nzw0301 (approved):
LGTM. Thanks!

tests/integration_tests/test_mlflow.py
```diff
@@ -19,6 +24,16 @@ def _objective_func(trial: optuna.trial.Trial) -> float:
     return (x - 2) ** 2 + (y - 25) ** 2 + z
 
 
+def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
+
+    x = trial.suggest_uniform("x", low=-10, high=10)
+    y = trial.suggest_loguniform("y", low=1, high=10)
```
toshihikoyanase, line 30: A nit, but please use `suggest_float` instead of `suggest_uniform` and `suggest_loguniform`. We prefer `suggest_float` as can be seen in https://github.com/optuna/optuna/pull/2344.
```suggestion
    x = trial.suggest_float("x", low=-10, high=10)
    y = trial.suggest_float("y", low=1, high=10, log=True)
```

optuna/integration/mlflow.py
```diff
@@ -239,14 +247,37 @@ def _set_tags(self, trial: optuna.trial.FrozenTrial, study: optuna.study.Study)
...
         Args:
-            value: Result of trial.
+            values: Results of a trial.
         """
-        trial_value = value if value is not None else float("nan")
-        mlflow.log_metric(self._metric_name, trial_value)
+
+        if isinstance(self._metric_name, list):
+            if len(self._metric_name) != len(values):
+                raise RuntimeError(
```
toshihikoyanase, line 259: `self._metric_name` is the correct type (i.e., list) and it is specified by users, so I think we can use more specific error than [`RuntimeError`](https://docs.python.org/3/library/exceptions.html#RuntimeError), such as [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError).

```suggestion
                raise ValueError(
```

optuna/integration/mlflow.py
```diff
@@ -239,14 +247,37 @@ def _set_tags(self, trial: optuna.trial.FrozenTrial, study: optuna.study.Study)
...
+        """Log the trial results as metrics to MLflow.
 
         Args:
-            value: Result of trial.
+            values: Results of a trial.
         """
-        trial_value = value if value is not None else float("nan")
-        mlflow.log_metric(self._metric_name, trial_value)
+
+        if isinstance(self._metric_name, list):
```
toshihikoyanase, line 257: When I set a tuple of strings, the following error occurred. In my understanding, `self._metric_name` is not necessarily a list. How about loose the type to `typing.Sequence`? See also https://github.com/optuna/optuna/blob/cbae80476c15b6d39e1d8851dc6a501c63c3ca92/optuna/study/study.py#L496 https://github.com/optuna/optuna/blob/cbae80476c15b6d39e1d8851dc6a501c63c3ca92/optuna/study/_optimize.py#L271

```console
mlflow.exceptions.MlflowException: Invalid metric name: '('foo', 'bar')_0'. Names] may only contain alphanumerics, underscores (_), dashes (-), periods (.), spaces ( ), and slashes (/).
```

optuna/integration/mlflow.py
```diff
@@ -239,14 +247,37 @@ def _set_tags(self, trial: optuna.trial.FrozenTrial, study: optuna.study.Study)
...
+                        len(values), len(self._metric_name)
+                    )
+                )
+
+            else:
+                names = self._metric_name
+
+        else:
+            if len(values) > 1:
+                # broadcast default name for multi-objective optimization
```
toshihikoyanase, line 272: A nit.
```suggestion
                # Broadcast default name for multi-objective optimization.
```

toshihikoyanase (changes requested):
Thank you for your great PR! I added a comment on the type of the `metric_name` argument.

optuna/integration/mlflow.py
xadrianzetx: I will relax type of `metric_name` in constructor, but internally we probably should keep checking for instance of `List` and `Tuple`, since checking just for `Sequence` will also pass strings.

tests/integration_tests/test_mlflow.py
xadrianzetx: Note: also TODO in `WeightsAndBiasesCallback`

optuna/integration/mlflow.py
xadrianzetx: Note: also TODO in `WeightsAndBiasesCallback`

optuna/integration/mlflow.py
xadrianzetx: Note: also TODO in `WeightsAndBiasesCallback`

optuna/integration/mlflow.py
xadrianzetx: Addressed in 3e6ada476ce08079eb93e812272a6fb4ad1536f9.

optuna/integration/mlflow.py
xadrianzetx: Addressed in ec628b84738cd8b9a84ad9f1bf7e5636aea8f718.

tests/integration_tests/test_mlflow.py
xadrianzetx: Addressed in 82b573ad6d32da113dcceb76fa3bee5fbd375325

optuna/integration/mlflow.py
xadrianzetx: Addressed in b3ee6bb69a30a97cb930462735b014c9f104cff1. At the end, constructor is relaxed to `Sequence`, and internally we are just checking for instance of `str`. This should enable user to ducktype any sequence safely. Also added test case covering error you have reported.

xadrianzetx:
Thanks for the review @toshihikoyanase! I've left some notes for myself, since changes you requested should also be mirrored in `WeightsAndBiasesCallback` as both APIs are pretty unified.

tests/integration_tests/test_mlflow.py
nzw0301: @toshihikoyanase I think we should mention such a preference at least in [coding style on Wiki](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions) or [`CONTRIBUTING.md`](https://github.com/optuna/optuna/blob/master/CONTRIBUTING.md).

toshihikoyanase: Thank you for your update!

> we can mention such a preference at least in our coding style on Wiki or CONTRIBUTING.md.

@nzw0301 Sounds good! I guess this topic is too specific to be mentioned in [`CONTRIBUTING.md`](https://github.com/optuna/optuna/blob/master/CONTRIBUTING.md), and I added [`Consistently use suggest_float` section](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#consistently-use-suggest_float) to the coding style wiki page. Please feel free to update it.

optuna/integration/mlflow.py
toshihikoyanase: Thanks!

optuna/integration/mlflow.py
toshihikoyanase: Thank you.

optuna/integration/mlflow.py
```diff
@@ -239,14 +247,37 @@ def _set_tags(self, trial: optuna.trial.FrozenTrial, study: optuna.study.Study)
...
+            else:
+                names = self._metric_name
+
+        else:
+            if len(values) > 1:
+                # Broadcast default name for multi-objective optimization.
+                names = ["{}_{}".format(self._metric_name, i) for i in range(len(values))]
+
+            else:
+                names = [self._metric_name]
```
toshihikoyanase, line 276: Thank you for relaxing the type of `metric_name`. 
I feel this `not isinstance(self._metric_name, str)` is a bit ambiguous. Non-string objects will be processed in the if clause, so a dictionary, for example, will also processed there. It raises the following error, but I guess we can show more user-friendly error message.

```
TypeError: expected string or bytes-like object
```

The following is just a suggestion of changes.

```suggestion
        if isinstance(self._metric_name, str):
            if len(values) > 1:
                # Broadcast default name for multi-objective optimization.
                names = ["{}_{}".format(self._metric_name, i) for i in range(len(values))]
            else:
                names = [self._metric_name]
        else:
            if not isinstance(self._metric_name, Sequence):
                raise TypeError(
                    "metric_name is supposed to be a string or a sequence of strings."
                )
            if len(self._metric_name) != len(values):
                raise ValueError(
                    "Running multi-objective optimization "
                    "with {} objective values, but {} names specified. "
                    "Match objective values and names, or use default broadcasting.".format(
                        len(values), len(self._metric_name)
                    )
                )
            else:
                names = self._metric_name
```

In addition, how about adding a test case corresponding to the type checking?

```python
def test_multiobjective_raises_on_type_mismatch(tmpdir: py.path.local) -> None:

    tracking_file_name = "file:{}".format(tmpdir)
    mlflc = MLflowCallback(tracking_uri=tracking_file_name, metric_name={0: "foo", 1: "bar"})
    study = optuna.create_study(study_name="my_study", directions=["minimize", "maximize"])

        study.optimize(_multiobjective_func, n_trials=1, callbacks=[mlflc])
```

optuna/integration/mlflow.py
toshihikoyanase: > constructor is relaxed to `Sequence`, and internally we are just checking for instance of `str`.

Related to this design, I added a suggestion in https://github.com/optuna/optuna/pull/2863#discussion_r695407397.

tests/integration_tests/test_mlflow.py
nzw0301: @toshihikoyanase I agree with you. It looks perfect! Thanks!

optuna/integration/mlflow.py
xadrianzetx: How do you feel about checking for `Sequence` at the beginning, before hitting this `if else` clause? Something like

```python
if not isinstance(self._metric_name, Sequence):
    raise ValueError("metric_name is supposed to be a string or a sequence of strings.")


if not isinstance(self._metric_name, str):
[...]
```

Issue #2741, opened 2021-06-10
[RFC] Improve sampler test coverage


hvy:
## Motivation

Prevent future bugs such as https://github.com/optuna/optuna/pull/2739 by adding more unit tests for samplers.

## Description

Sort out all scenarios samplers should be able to handle, i.e. conditional parameters. Add tests.

hvy:
Refactoring in general should not change the behavior or algorithms. It might be worth considering adding tests with fixed seeds that check/assert the values of `best_params`. That way, we could get rid of some of the tests that are testing private interfaces such as the current tests for the TPE sampler.

toshihikoyanase:
I'm interested in this issue.

> Sort out all scenarios samplers should be able to handle, i.e. conditional parameters. Add tests.

I came up with the following scenarios:
- A single parameter
- Conditional parameters
- Combination of different types of parameters
- Changing the range of `suggest_int` or `suggest_float`

I'd like to add test cases corresponding to them. Maybe I can parametrize the objective functions.

> It might be worth considering adding tests with fixed seeds that check/assert the values of `best_params`.

I guess such test cases have both pros and cons, and I'd like to exclude this from my PR.
It will increase the maintenance cost of the test. Actually, we needed to change the test case when we updated the Optuna version from v2.4.0 to v2.5.0 or later (https://github.com/facebookresearch/hydra/pull/1726/files#r673470375). This is due to the introduction of  the search space transform (https://github.com/optuna/optuna/commit/5db7b5bc0b1442d5429b7ba816e4f4271ef94b94#diff-98431f6c0563769a069980db67d10bde2830661626057a8e69ba3e46f8e98ed4), and I think it may be costly to find the cause of the test failure.

Alternatively, we can check the change of `best_values` / `best_params` using benchmark suites like kurobako. It is not an easy task to automate benchmarking as we tried it in https://github.com/optuna/optuna/pull/2543, but I feel it is a straight forward way to check the performance using the benchmark.

PR #2869, opened 2021-08-25
Add scenario tests for samplers.

toshihikoyanase:
## Motivation
This PR addresses a part of #2741.

## Description of the changes

This PR adds test cases which actually optimize objective functions containing the following parameters:
- A single parameter
- Conditional parameters
- Combination of different types of parameters
- Changing the range of `suggest_int` and `suggest_float`

### DIFF
--- a/tests/samplers_tests/test_samplers.py
+++ b/tests/samplers_tests/test_samplers.py
@@ -59,6 +59,9 @@ def parametrize_suggest_method(name: str) -> MarkDecorator:
             lambda t: t.suggest_float(name, 0, 10),
             lambda t: t.suggest_int(name, 0, 10),
             lambda t: cast(float, t.suggest_categorical(name, [0, 1, 2])),
+            lambda t: t.suggest_float(name, 0, 10, step=0.5),
+            lambda t: t.suggest_float(name, 1e-7, 10, log=True),
+            lambda t: t.suggest_int(name, 1, 10, log=True),
         ],
     )
 
--- a/tests/samplers_tests/test_samplers.py
+++ b/tests/samplers_tests/test_samplers.py
@@ -1,5 +1,6 @@
 import pickle
 from typing import Any
+from typing import cast
 from typing import Callable
 from typing import Dict
 from typing import Optional
@@ -46,6 +47,14 @@ parametrize_multi_objective_sampler = pytest.mark.parametrize(
     [
         optuna.samplers.NSGAIISampler,
         optuna.samplers.MOTPESampler,
+    ],
+)
+parametrize_suggest_method = lambda name: pytest.mark.parametrize(
+    f"suggest_method_{name}",
+    [
+        lambda t: t.suggest_float(name, 0, 10),
+        lambda t: t.suggest_int(name, 0, 10),
+        lambda t: cast(float, t.suggest_categorical(name, [0, 1, 2])),
     ],
 )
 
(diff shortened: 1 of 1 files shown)


toshihikoyanase:
@not522 @keisuke-umezawa Could you review this PR, please?

tests/samplers_tests/test_samplers.py
```diff
@@ -573,3 +586,82 @@ def test_sample_single_distribution(sampler_class: Callable[[], BaseSampler]) ->
...
+        v += trial.suggest_int("y", low, high)
+        return v
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", optuna.exceptions.ExperimentalWarning)
+        sampler = sampler_class()
+
+    study = optuna.study.create_study(sampler=sampler)
+    study.optimize(lambda t: objective(t, 0, 10), n_trials=10)
+    study.optimize(lambda t: objective(t, 0, 5), n_trials=10)
```
not522, line 664: How about adding other scenarios? For example, the range is expanding or different.

toshihikoyanase: That makes sense. I parametrized the range and added the suggested patterns in https://github.com/optuna/optuna/pull/2869/commits/862a435ec0fcc4c7da8661f62dd71d9026959622.

tests/samplers_tests/test_samplers.py
```diff
@@ -48,6 +50,17 @@
 )
 
 
+def parametrize_suggest_method(name: str) -> MarkDecorator:
+    return pytest.mark.parametrize(
+        f"suggest_method_{name}",
+        [
+            lambda t: t.suggest_float(name, 0, 10),
+            lambda t: t.suggest_int(name, 0, 10),
+            lambda t: cast(float, t.suggest_categorical(name, [0, 1, 2])),
```
keisuke-umezawa, line 59: [question] Why do you choose only those three suggest functions? Maybe, some samplers do not support other suggest functions?

keisuke-umezawa:
LGTM, other than one comment.

tests/samplers_tests/test_samplers.py
toshihikoyanase: I just covered the modern `suggest` methods without arguments. This is mainly because the test time. If we cover the all six distributions, the time of the tests increase as follows:

- `test_single_parameter_objective`: 9.70s -> 19.83s
- `test_combination_of_different_distributions_objective` : 34.17s -> 139.99s
(CPU: AMD Ryzen 5 3600, Memory: 64GB)

The change of test code:
```diff
--- a/tests/samplers_tests/test_samplers.py
+++ b/tests/samplers_tests/test_samplers.py
@@ -57,6 +57,9 @@ def parametrize_suggest_method(name: str) -> MarkDecorator:
             lambda t: t.suggest_float(name, 0, 10),
             lambda t: t.suggest_int(name, 0, 10),
             lambda t: cast(float, t.suggest_categorical(name, [0, 1, 2])),
+            lambda t: t.suggest_float(name, 0, 10, step=0.5),
+            lambda t: t.suggest_float(name, 1e-7, 10, log=True),
+            lambda t: t.suggest_int(name, 1, 10, log=True),
         ],
     )
```

I think it's a bit long to take more than two minutes for a single test function. Do you have any ideas to reduce the test time while increasing the test coverage?

not522: `SkoptSampler` costs most of the test time. How about using random sampling in `skopt`? It is 5 times faster in my environment.

```python
-        lambda: optuna.integration.SkoptSampler(skopt_kwargs={"n_initial_points": 1}),
+        lambda: optuna.integration.SkoptSampler(skopt_kwargs={"base_estimator": "dummy", "n_initial_points": 1}),
```

toshihikoyanase: @not522 Thank you for your investigation. When I applied `"base_estimator" : "dummy"` to `SkoptSampler`, the test time greatly decreased as follows:
- `test_single_parameter_objective`: 19.83s -> 2.0s
- `test_combination_of_different_distributions_objective` : 139.99s -> 6.18s

I guess it is actually [implementation of random sampling](https://github.com/scikit-optimize/scikit-optimize/blob/273db8f3d223c4d471761e6d9ad1bf183b1ee6e1/skopt/optimizer/dummy.py), and we can use it in the unit test. I'm wondering if I should include the change since it may be out-of-scope of this PR. What do you think?

not522: I agree to keep the number of functions smaller in this PR.

not522 (approved):
Thank you. LGTM.

tests/samplers_tests/test_samplers.py
toshihikoyanase: Since #2910 was merged into the master, I rebased the branch and introduced three additional suggest methods in https://github.com/optuna/optuna/pull/2869#discussion_r702512085. The test time is as follows:
- test_single_parameter_objective: 9.70s -> 19.83s -> 1.89s
- test_combination_of_different_distributions_objective : 34.17s -> 139.99s -> 6.03s

@keisuke-umezawa Please take another look.

keisuke-umezawa (approved):

PR #2883, opened 2021-08-30
Improve the docstring of `MLflowCallback` 

nzw0301:
## Motivation

I found a minor thing that `study.user_attrs` in the docstring can be linked to Optuna's documentation element.

## Description of the changes

Use sphinx's syntax to link to the corresponding part: [`study.user_attrs`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html#optuna.study.Study.user_attrs).

I think this PR is straightforward, so one approval is enough to be merged into the master branch.

PR #2879, opened 2021-08-27
Hyperband no longer assumes it is the only pruner

cowwoc:
Fixes #2878

PR #2885, opened 2021-08-30
Introduce version constraint of `alembic`

nzw0301:
## Motivation

[`alembic` 1.7.0](https://github.com/sqlalchemy/alembic/releases/tag/rel_1_7_0) has been released. However, the CI starts to fail due to `mypy`.

See also this CI error, https://github.com/optuna/optuna/pull/2883/checks?check_run_id=3463453849, as a concrete error.

## Description of the changes

As a hot-fix, this PR introduces a version constraint of `alembic<1.7.0` and TODO line.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,8 @@ def get_long_description() -> str:
 def get_install_requires() -> List[str]:
 
     requirements = [
-        "alembic",
+        # TODO(nzw) remove this version constraint after fixing mypy error
+        "alembic<1.7.0",
         "cliff",
         "cmaes>=0.8.2",
         "colorlog",


hvy:
Thanks for the quick fix. This sounds reasonable as a temporary fix.
Do you know if we need to be more careful than simply adding an assertion about the return value from `get_base` not being `None`? The behavior of `alembic` hasn't changed in my understanding and type hints were simply introduced starting `alembic==1.7.0`.

hvy:
@nzw0301 what do you think about https://github.com/optuna/optuna/pull/2887?

nzw0301:
Honestly, I do not know about the behaviour of `alembic` 1.7.0. This is why I created this hot-fix PR.

The proposed change in #2879 looks better than this hot-fix.

PR #2887, opened 2021-08-31
Fix recent `alembic` 1.7.0 type hint error

hvy:
## Motivation

Fixes the error described in https://github.com/optuna/optuna/pull/2885.

I think this fix is acceptable according to the description of https://alembic.sqlalchemy.org/en/latest/api/script.html#alembic.script.ScriptDirectory.get_base and that we always have 
precisely one version that doesn't have a `down_revision` in https://github.com/optuna/optuna/tree/master/optuna/storages/_rdb/alembic/versions.

## Description of the changes

Suppresses `None` return values from `ScriptDirectory.get_base`.

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -1346,7 +1346,9 @@ class _VersionManager(object):
     def _get_base_version(self) -> str:
 
         script = self._create_alembic_script()
-        return script.get_base()
+        base = script.get_base()
+        assert base is not None, "There should be exactly one base, i.e. v0.9.0.a."
+        return base
 
     def get_all_versions(self) -> List[str]:
 


nzw0301 (approved):
Thank you for your investigation and for fixing the root cause! The change looks great to me.

hvy:
Thanks @nzw0301 for the quick review!
@toshihikoyanase could you jump in as a second reviewer?

toshihikoyanase (approved):
Thank you for your quick fix. The change looks good to me. 

I confirmed that the change did not affect the behavior of the upgrade command. I successfully upgraded the schema from `v2.4.0` to `v2.6.0` (latest) as follows:

```console
# optuna create-study --storage sqlite:///foo.db --study-name foo --direction minimize
[E 2021-08-31 02:25:08,009] The runtime optuna version 2.10.0.dev is no longer compatible with the table schema (set up by optuna 2.4.0). Please execute `$ optuna storage upgrade --storage $STORAGE_URL` for upgrading the storage.
# optuna storage upgrade --storage sqlite:///foo.db
[I 2021-08-31 02:25:25,338] Upgrading the storage schema to the latest version.
[I 2021-08-31 02:25:25,393] Completed to upgrade the storage.
# optuna create-study --storage sqlite:///foo.db --study-name bar --direction minimize
[I 2021-08-31 02:25:28,691] A new study created in RDB with name: bar
bar
```

jieru-hu:
Thanks for fixing this! Is there any plan on releasing a new version soon? We have a dependency on optuna and our project's mypy is failing because of this. 

Thank you!

hvy:
Thanks for reaching out and sorry for the trouble. The initial release was planned towards the end of Sep., but I'm discussing it with the team. If it's urgent, I'd suggest avoiding the latest alembic release (i.e. introduce a change similar to https://github.com/optuna/optuna/pull/2885) or perhaps configure mypy meanwhile. Would that work for you?

jieru-hu:
> Thanks for reaching out and sorry for the trouble. The initial release was planned towards the end of Sep., but I'm discussing it with the team. If it's urgent, I'd suggest avoiding the latest alembic release (i.e. introduce a change similar to #2885) or perhaps configure mypy meanwhile. Would that work for you?

thanks for the response! I will pin the version of alembic for now then.

PR #2890, opened 2021-08-31
Add test cases for storage upgrade.

toshihikoyanase:
## Motivation

When we modify code related to the RDB schema, we usually check the behavior manually (e.g., #2180 and #2887). Such tests are cumbersome and the operators may make mistakes in the procedure.

## Description of the changes

This PR adds test cases for the storage upgrade.

- Adds sqlite3 database files created with different schema versions
- Executes `storage.upgrade` method in the test cases
- Start or resume studies after upgrading the schema

### Opinion wanted
- The sizes of database files are larger than 60KB. I included them in the repository, but what do you think of it?


hvy:
> The sizes of database files are larger than 60KB. I included them in the repository, but what do you think of it?

Thanks for the careful consideration. I think this is acceptable for now and for a reasonable amount of time forward given that they're excluded from the installation and it's still a small fraction of the entire repository (~2% in my understanding).

hvy:
@c-bata could you jump in as a reviewer? Let me know if you're busy.

tests/storages_tests/rdb_tests/alembic_tests/create-db.py
```diff
@@ -0,0 +1,43 @@
+import sys
+
+import optuna
+
+
+def objective(trial):
+    x = trial.suggest_uniform("x", -5, 5)  # optuna==0.9.0 does not have suggest_float.
```
hvy, line 7: Just curious but were you able to generate the database files for all versions using the same single script?

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -177,6 +183,71 @@ def test_upgrade() -> None:
...
+        x = trial.suggest_uniform("x", -5, 5)  # optuna==0.9.0 does not have suggest_float.
+        y = trial.suggest_int("y", 0, 10)
+        z = cast(float, trial.suggest_categorical("z", [-5, 0, 5]))
+        return x ** 2 + y ** 2 + z ** 2
+
+    def mo_objective(trial: Trial) -> Tuple[float, float]:
+        x = trial.suggest_float("x", -5, 5)
+        y = trial.suggest_int("y", 0, 10)
+        z = cast(float, trial.suggest_categorical("z", [-5, 0, 5]))
+        return x, x ** 2 + y ** 2 + z ** 2
```
hvy, line 198: How about leaving a comment that these objective functions must be similar to those in `create-db.py`?

tests/storages_tests/rdb_tests/alembic_tests/create-db.py
```diff
@@ -0,0 +1,43 @@
+import sys
+
+import optuna
+
+
+def objective(trial):
```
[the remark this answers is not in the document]
hvy: Are system/user attributed left out to keep the tests simpler/more maintainable? Checking those RDB tables, in addition to perhaps params and values might increase coverage (but also maintenance cost). My understanding is that we usually want to check those when we upgrade the schemas.

c-bata (commented):
> The sizes of database files are larger than 60KB. I included them in the repository, but what do you think of it?

How about creating database files before running each test suite like `alembic.command.upgrade(config, "v0.9.0.a")`?

https://github.com/optuna/optuna/blob/d4df5ab208ae902fb60cc8ced1539a22f257346a/optuna/storages/_rdb/storage.py#L1361

hvy (changes requested):
Again, thank you for the great work. Sounds good to keep the old database files.

I left some comments if you could take a look. My concerns is that it might be difficult to re-generate these database files if we'd want to make changes to them in the long-term, but I don't think it's urgent.

c-bata:
> > The sizes of database files are larger than 60KB. I included them in the repository, but what do you think of it?
> 
> How about creating database files before running each test suite like `alembic.command.upgrade(config, "v0.9.0.a")`?
> 
> https://github.com/optuna/optuna/blob/d4df5ab208ae902fb60cc8ced1539a22f257346a/optuna/storages/_rdb/storage.py#L1361

Ah, please ignore this comment :bow: We can create database files but we cannot run objective functions on the schema.

tests/storages_tests/rdb_tests/alembic_tests/create-db.py
```diff
@@ -0,0 +1,43 @@
+import sys
```
c-bata, line 1: How about renaming `create_db.py` ?

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -177,6 +183,71 @@ def test_upgrade() -> None:
...
+
+    src_db_file = os.path.join(os.path.dirname(__file__), "alembic_tests", f"{optuna_version}.db")
```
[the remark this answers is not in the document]
c-bata: Agree with hvy-san.

tests/storages_tests/rdb_tests/alembic_tests/create-db.py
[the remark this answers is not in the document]
toshihikoyanase: > Are type hints purposely omitted from this script?

Thank you for your suggestions. I added the type annotation in https://github.com/optuna/optuna/pull/2890/commits/08adb1a6cc54ac0d0ecc5d50a95673bc3d1cab7f, and I moved `create_db.py` to `tests/storage_tests/rdb_tests` that is the target of the formatter and the linter in https://github.com/optuna/optuna/pull/2890/commits/7aeb6e19fea9aa888af8104912604dd037c27708.

> Are system/user attributed left out to keep the tests simpler/more maintainable?

I just forgot to handle them. I added the corresponding test cases in https://github.com/optuna/optuna/pull/2890/commits/7aeb6e19fea9aa888af8104912604dd037c27708.

tests/storages_tests/rdb_tests/alembic_tests/create-db.py
```diff
@@ -0,0 +1,43 @@
...
+
+def mo_objective(trial):
+    x = trial.suggest_float("x", -5, 5)
+    y = trial.suggest_int("y", 0, 10)
+    z = trial.suggest_categorical("z", [-5, 0, 5])
+    return x, x ** 2 + y ** 2 + z ** 2
+
+
+if __name__ == "__main__":
+    storage_url = sys.argv[1]
```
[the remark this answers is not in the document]
toshihikoyanase: That makes sense. I replaced `sys.argv` with `argparse` in https://github.com/optuna/optuna/pull/2890/commits/dd250461d2d7bba3b083bdeda1c5b5242136134f. Also, I applied the suggested default value to the storage URL. It really helped the operation of database file creation. Thanks!

tests/storages_tests/rdb_tests/test_storage.py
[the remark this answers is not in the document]
toshihikoyanase: Sounds reasonable. I renamed `test_alembic` to `test_upgrade_assets` in https://github.com/optuna/optuna/pull/2890/commits/34d8ef962981ce947bb2c45406ef170b8bf76183.

tests/storages_tests/rdb_tests/alembic_tests/create-db.py
toshihikoyanase: Yes, we can generate the database files from 0.9.0 to the master with this script.

tests/storages_tests/rdb_tests/test_storage.py
toshihikoyanase: Thank you for your suggestion. I reused these objective functions in the test_storage.py in https://github.com/optuna/optuna/pull/2890/commits/34d8ef962981ce947bb2c45406ef170b8bf76183. I think we can reduce the maintenance cost, but what do you think?

tests/storages_tests/rdb_tests/alembic_tests/create-db.py
toshihikoyanase: Thanks! I renamed it in https://github.com/optuna/optuna/pull/2890/commits/34d8ef962981ce947bb2c45406ef170b8bf76183.

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -177,6 +184,72 @@ def test_upgrade() -> None:
...
+
+        # Resume single objective optimization.
+        study = load_study(storage=storage, study_name="single")
+        assert len(study.trials) == 1
+        study.optimize(objective_test_upgrade, n_trials=1)
+        assert len(study.trials) == 2
+        for trial in study.trials:
+            assert trial.system_attrs["a"] == 0
+            assert trial.user_attrs["b"] == 1
+            assert trial.intermediate_values[0] == 0.5
```
hvy, line 222: I'm wondering if we should check value(s), params, and such properties as well (covering all RDB tables).

tests/storages_tests/rdb_tests/alembic_tests/create-db.py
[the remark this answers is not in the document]
hvy: Thanks!

tests/storages_tests/rdb_tests/test_storage.py
hvy: Oh, that's even better. 👍

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -177,6 +184,82 @@ def test_upgrade() -> None:
...
+        for trial in study.trials:
+            assert trial.system_attrs["a"] == 0
+            assert trial.user_attrs["b"] == 1
+            assert -5 <= trial.params["x"] <= 5
+            assert 0 <= trial.params["y"] <= 10
+            assert trial.params["z"] in (-5, 0, 5)
+            assert -5 <= trial.values[0] < 5
+            assert 0 <= trial.values[1] <= 150
+        assert study.system_attrs["c"] == 2
+        assert study.user_attrs["d"] == 3
```
c-bata, line 260: How about splitting the test suite for these scenarios? It looks like a code smell of [conditional test logic](http://xunitpatterns.com/Conditional%20Test%20Logic.html).

c-bata (approved):
LGTM with nits.

tests/storages_tests/rdb_tests/test_storage.py
toshihikoyanase: Thank you for your suggestion. I added the assertions of params and values. At first, I used `enqueue_trial` to fix those values, but I couldn't use it in optuna==0.7.0, and I gave up.

tests/storages_tests/rdb_tests/test_storage.py
toshihikoyanase: That makes sense. I split `test_upgrade` into `test_upgrade_single_objective_optimization` and `test_upgrade_multi_objective_optimization` in https://github.com/optuna/optuna/pull/2890/commits/a36e0538b9067a4f542fc240621de58ad67d0171. Thank you!

tests/storages_tests/rdb_tests/test_storage.py
hvy: I think that's great, thanks!

hvy (approved):
Thanks, LGTM!

@c-bata do you want to take another look?

PR #2223, opened 2021-01-18
Include the upper bound of uniform/loguniform distributions.

c-bata:
## Motivation

In the current specification of UniformDistribution and LogUniformDistribution, the upper bounds are excluded.

```python
>>> import optuna
>>> d = optuna.distributions.UniformDistribution(low=0.0, high=1.0)
>>> d._contains(1.0)
False
>>> d = optuna.distributions.LogUniformDistribution(low=0.01, high=1.0)
>>> d._contains(1.0)
False
```

This behavior is based on the specification of [stdlib](https://docs.python.org/3/library/random.html#random.uniform) and/or [numpy](https://numpy.org/doc/stable/reference/random/generated/numpy.random.uniform.html) random modules. But it confuses the users and the developers in some situation.

* `enqueue_trial({'x': 1.0})` is not evaluated when `x` is defined as `x = suggest_float('x', 0, 1.0)`. I actually bumped this problem several times and it is difficult to notice this bug until it runs.
* https://github.com/optuna/optuna/pull/619
* https://github.com/optuna/optuna/pull/1558

In this PR, I propose to include the upper bound.

## Description of the changes

* Include the upper bound of uniform/loguniform distributions.

PR #2897, opened 2021-09-02
Add tests of `sample_relative` and fix type of return values of `SkoptSampler` and `PyCmaSampler`

toshihikoyanase:
## Motivation

Currently, tests of the `sample_relative` methods are missing in `tests/sampler_tests/test_sampler.py` while the `sample_independent` methods are tested in it. I think this is because only a few samplers implemented `sample_relative` when we introduced `sample_relative` and each sampler with `sample_relative` tested it in the individual test file (i.e., `tests/sampler_tests/test_cmaes.py`).

Now that multiple samplers implement `sample_relative`, I think we can introduce the test cases for `sample_relative` to `tests/sampler_tests/test_sampler.py`.

## Description of the changes

- Add test cases of `sample_relative` for `TPESampler(multivariate=True)`, `CmaEsSampler`, `SkoptSampler` and `PyCmaSampler`.
- Fix type of return values of `SkoptSampler` and `PyCmaSampler`.
  - This type checking can be seen in the test cases of `sample_independent`.

### DIFF
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -463,7 +463,7 @@ class _Optimizer(object):
             # Type of cma_param_value is np.floating, so cast it to Python's built-in float.
             return float(cma_param_value)
         if isinstance(dist, LogUniformDistribution):
-            return float(math.exp(cma_param_value))
+            return math.exp(cma_param_value)
         if isinstance(dist, DiscreteUniformDistribution):
             v = numpy.round(cma_param_value / dist.q) * dist.q + dist.low
             # v may slightly exceed range due to round-off errors.
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -459,8 +459,11 @@ class _Optimizer(object):
     ) -> Any:
 
         dist = search_space[param_name]
+        if isinstance(dist, UniformDistribution):
+            # Type of cma_param_value is np.floating, so cast it to Python's built-in float.
+            return float(cma_param_value)
         if isinstance(dist, LogUniformDistribution):
-            return math.exp(cma_param_value)
+            return float(math.exp(cma_param_value))
         if isinstance(dist, DiscreteUniformDistribution):
             v = numpy.round(cma_param_value / dist.q) * dist.q + dist.low
             # v may slightly exceed range due to round-off errors.
--- a/tests/samplers_tests/test_samplers.py
+++ b/tests/samplers_tests/test_samplers.py
@@ -6,6 +6,7 @@ from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Sequence
+from typing import Union
 import warnings
 
 import numpy as np
--- a/tests/samplers_tests/test_samplers.py
+++ b/tests/samplers_tests/test_samplers.py
@@ -48,7 +48,9 @@ parametrize_relative_sampler = pytest.mark.parametrize(
     [
         lambda: optuna.samplers.TPESampler(n_startup_trials=0, multivariate=True),
         lambda: optuna.samplers.CmaEsSampler(n_startup_trials=0),
-        lambda: optuna.integration.SkoptSampler(skopt_kwargs={"n_initial_points": 1}),
+        lambda: optuna.integration.SkoptSampler(
+            skopt_kwargs={"base_estimator": "dummy", "n_initial_points": 1}
+        ),
         lambda: optuna.integration.PyCmaSampler(n_startup_trials=0),
     ],
 )
@@ -276,7 +278,7 @@ def test_sample_relative_numerical(
         params = study.sampler.sample_relative(study, _create_new_trial(study), search_space)
         return [params[name] for name in search_space]
 
-    points = np.array([sample() for _ in range(2)])
+    points = np.array([sample() for _ in range(10)])
     for i, distribution in enumerate(search_space.values()):
         assert isinstance(
             distribution,
@@ -308,7 +310,7 @@ def test_sample_relative_categorical(relative_sampler_class: Callable[[], BaseSa
         params = study.sampler.sample_relative(study, _create_new_trial(study), search_space)
         return [params[name] for name in search_space]
 
-    points = np.array([sample() for _ in range(2)])
+    points = np.array([sample() for _ in range(10)])
     for i, distribution in enumerate(search_space.values()):
         assert isinstance(distribution, CategoricalDistribution)
         assert np.all([v in distribution.choices for v in points[:, i]])
@@ -342,7 +344,7 @@ def test_sample_relative_mixed(
         params = study.sampler.sample_relative(study, _create_new_trial(study), search_space)
         return [params[name] for name in search_space]
 
-    points = np.array([sample() for _ in range(2)])
+    points = np.array([sample() for _ in range(10)])
     assert isinstance(
         search_space["x"],
         (
--- a/tests/samplers_tests/test_samplers.py
+++ b/tests/samplers_tests/test_samplers.py
@@ -1,7 +1,9 @@
+from collections import OrderedDict
 import pickle
 from typing import Any
 from typing import Callable
 from typing import Dict
+from typing import List
 from typing import Optional
 from typing import Sequence
 import warnings
@@ -265,16 +267,17 @@ def test_sample_relative_numerical(
     y_distribution: BaseDistribution,
 ) -> None:
 
-    search_space = {"x": x_distribution, "y": y_distribution}
-    study = optuna.study.create_study(sampler=optuna.samplers.RandomSampler(seed=1))
-    for _ in range(4):
-        trial = study.ask(search_space)
-        study.tell(trial, sum(trial.params.values()))
-    study.sampler = relative_sampler_class()
+    search_space: Dict[str, BaseDistribution] = OrderedDict(x=x_distribution, y=y_distribution)
+    study = optuna.study.create_study(sampler=relative_sampler_class())
+    trial = study.ask(search_space)
+    study.tell(trial, sum(trial.params.values()))
 
-    values = study.sampler.sample_relative(study, _create_new_trial(study), search_space)
-    for name, distribution in search_space.items():
-        value = values[name]
+    def sample() -> List[float]:
+        params = study.sampler.sample_relative(study, _create_new_trial(study), search_space)
+        return [params[name] for name in search_space]
+
+    points = np.array([sample() for _ in range(2)])
+    for i, distribution in enumerate(search_space.values()):
         assert isinstance(
             distribution,
             (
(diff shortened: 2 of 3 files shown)


tests/samplers_tests/test_samplers.py
```diff
@@ -225,6 +234,120 @@ def sample() -> float:
...
+            (
+                UniformDistribution,
+                LogUniformDistribution,
+                DiscreteUniformDistribution,
+                IntUniformDistribution,
+                IntLogUniformDistribution,
+            ),
+        )
+        assert value >= distribution.low
+        assert value <= distribution.high
```
himkt, line 285:
```suggestion
        assert distribution.low <= value <= distribution.high
```

[nits]
It is really nits but it aligns with `test_sample_relative_mixed`.

https://github.com/optuna/optuna/pull/2897/files#diff-8b85ccf8a3534b2aa878bce8ece8ce930af410ab178ebc6d80031badbabb5052R345

toshihikoyanase: Good catch! Thank you!

hvy:
@nzw0301 could you review this PR? If you're busy, maybe @himkt could continue?

optuna/integration/cma.py
```diff
@@ -459,8 +459,11 @@ def _to_optuna_params(
     ) -> Any:
 
         dist = search_space[param_name]
+        if isinstance(dist, UniformDistribution):
+            # Type of cma_param_value is np.floating, so cast it to Python's built-in float.
+            return float(cma_param_value)
         if isinstance(dist, LogUniformDistribution):
-            return math.exp(cma_param_value)
+            return float(math.exp(cma_param_value))
```
nzw0301, line 466: Maybe I'm missing something, but `math.exp` returns `float` value if its argument is `numpy.float` value, so `float` cast can be removed:

```suggestion
            return math.exp(cma_param_value)
```

I locally confirmed this as follows and also checked the `tests/samplers_tests/test_samplers.py` was passed.
```bash
python -c "import numpy as np; import math; print(isinstance(math.exp(np.float64(0.)), np.floating))"
False
```

nzw0301 (commented):
I've left one minor comment. But I'll look at more in the next week. 

Have a good weekend 🍹

nzw0301:
Probably this is out-of-scope of this PR, but let me ask one thing: Why isn't BoTorchSampler in `tests/samplers_tests/test_samplers.py`?

I locally confirm that adding the sampler to 

- `parametrize_multi_objective_sampler`: no test failed errors.
- `parametrize_sampler`: one failed case related to nan's error handling
- `parametrize_relative_sampler`: many failed cases.

hvy (commented):
Thanks for the PR!
I left some comments and questions if you could take a look.

On a side note, it's a tiny bit worrisome that tests in `tests/samplers_tests/test_samplers.py` all quite varied in style (e.g. existing independent tests are using numpy arrays, variable naming conventions, etc.). Do you think we can or should address this perhaps in a different PR?

optuna/integration/cma.py
toshihikoyanase: You're absolutely right. I fixed it in https://github.com/optuna/optuna/pull/2897/commits/e812161c4047f58c782ce849b101fe71acf0d726. Thank you.

tests/samplers_tests/test_samplers.py
```diff
@@ -225,6 +234,119 @@ def sample() -> float:
...
+)
+def test_sample_relative_numerical(
+    relative_sampler_class: Callable[[], BaseSampler],
+    x_distribution: BaseDistribution,
+    y_distribution: BaseDistribution,
+) -> None:
+
+    search_space = {"x": x_distribution, "y": y_distribution}
+    study = optuna.study.create_study(sampler=optuna.samplers.RandomSampler(seed=1))
+    for _ in range(4):
```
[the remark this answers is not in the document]
toshihikoyanase: Thank you for pointing it out. I found that the number `1` is enough to determine the search space. I fixed it in https://github.com/optuna/optuna/pull/2897/commits/4c544e5e2d76a902710b7c4ebbe860eae33ceac6.

tests/samplers_tests/test_samplers.py
```diff
@@ -225,6 +234,119 @@ def sample() -> float:
...
+    ],
+)
+def test_sample_relative_numerical(
+    relative_sampler_class: Callable[[], BaseSampler],
+    x_distribution: BaseDistribution,
+    y_distribution: BaseDistribution,
+) -> None:
+
+    search_space = {"x": x_distribution, "y": y_distribution}
+    study = optuna.study.create_study(sampler=optuna.samplers.RandomSampler(seed=1))
```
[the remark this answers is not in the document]
toshihikoyanase: I should use the parametrized samplers. I fixed it in https://github.com/optuna/optuna/pull/2897/commits/4c544e5e2d76a902710b7c4ebbe860eae33ceac6. Thanks!

tests/samplers_tests/test_samplers.py
```diff
@@ -225,6 +234,119 @@ def sample() -> float:
...
+            (
+                UniformDistribution,
+                LogUniformDistribution,
+                DiscreteUniformDistribution,
+                IntUniformDistribution,
+                IntLogUniformDistribution,
+            ),
+        )
+        assert distribution.low <= value <= distribution.high
+        assert not isinstance(value, np.floating)
```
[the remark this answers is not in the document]
toshihikoyanase: I applied the suggestion in https://github.com/optuna/optuna/pull/2897/commits/b82c8c4d01b5ebae06b1dfc81eace38e71784652.

To distinguish `float` and `int`, I introduced conditional branches in the test case. If it makes test cases complicated, I'll split the test cases.

tests/samplers_tests/test_samplers.py
```diff
@@ -225,6 +234,119 @@ def sample() -> float:
...
+) -> None:
+
+    search_space = {"x": x_distribution, "y": y_distribution}
+    study = optuna.study.create_study(sampler=optuna.samplers.RandomSampler(seed=1))
+    for _ in range(4):
+        trial = study.ask(search_space)
+        study.tell(trial, sum(trial.params.values()))
+    study.sampler = relative_sampler_class()
+
+    values = study.sampler.sample_relative(study, _create_new_trial(study), search_space)
```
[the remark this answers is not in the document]
toshihikoyanase: Thank you for your comments. It's important to keep the test style to keep the maintainability.

> all quite varied in style (e.g. existing independent tests are using numpy arrays, variable naming conventions, etc.)

I also agree with the comment, so I restructured the sample_relative tests and employed a similar style to the sample_independent tests in https://github.com/optuna/optuna/pull/2897/commits/4c544e5e2d76a902710b7c4ebbe860eae33ceac6.

tests/samplers_tests/test_samplers.py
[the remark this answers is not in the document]
hvy: I think it's acceptable, thanks!

tests/samplers_tests/test_samplers.py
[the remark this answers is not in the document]
hvy: Thanks!

tests/samplers_tests/test_samplers.py
```diff
@@ -229,6 +242,142 @@ def sample() -> float:
...
+    x_distribution: BaseDistribution,
+    y_distribution: BaseDistribution,
+) -> None:
+
+    search_space: Dict[str, BaseDistribution] = OrderedDict(x=x_distribution, y=y_distribution)
+    study = optuna.study.create_study(sampler=relative_sampler_class())
+    trial = study.ask(search_space)
+    study.tell(trial, sum(trial.params.values()))
+
+    def sample() -> List[float]:
```
hvy, line 277: A nit but we're also parameterizing over int distributions.
```suggestion
    def sample() -> List[Union[int, float]]:
```

tests/samplers_tests/test_samplers.py
```diff
@@ -229,6 +242,142 @@ def sample() -> float:
...
+            (
+                UniformDistribution,
+                LogUniformDistribution,
+                DiscreteUniformDistribution,
+                IntUniformDistribution,
+                IntLogUniformDistribution,
+            ),
+        )
+        assert np.all(points[:, i] >= distribution.low)
+        assert np.all(points[:, i] <= distribution.high)
```
hvy, line 294: To be exact, we should change `<=` to `<` for uniform and loguniform, but I also think this is acceptable.

hvy (changes requested):
Thanks for the quick fixes, it looks more aligned with the existing tests now. It basically LGTM besides some minor comments and the comment from @nzw0301.

tests/samplers_tests/test_samplers.py
toshihikoyanase: Thank you for your comment. #2223 changed the range definitions of `UniformDistribution` and `LogUniformDistribution`, and I use `<=` for those. (Honestly speaking, the definitions are different from that of `suggest_float` and it is confusing. We may want to revise it in the version 3 discussion.)

I checked the distribution reference and the current specification of the ranges is as follows:

- [`UniformDistribution`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.distributions.UniformDistribution.html#optuna.distributions.UniformDistribution): `[low , high]` (Both included)
- [`LogUniformDistribution`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.distributions.LogUniformDistribution.html#optuna.distributions.LogUniformDistribution): `[low, high]`
- [`DiscreteUniformDistribution`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.distributions.DiscreteUniformDistribution.html#optuna.distributions.DiscreteUniformDistribution): `[low, high]`
- [`IntUniformDistribution`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.distributions.IntUniformDistribution.html#optuna.distributions.IntUniformDistribution): `[low, high]`
- [`IntLogUniformDistribution`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.distributions.IntLogUniformDistribution.html#optuna.distributions.IntLogUniformDistribution): `[low, high]`

hvy (approved):
LGTM!

I think that the tests could be improved a bit to not depend on numpy but it's open for discussion and probably outside the scope of this PR. Same goes for the BoTorch sampler parameterization.

toshihikoyanase:
> Probably this is out-of-scope of this PR, but let me ask one thing: Why isn't BoTorchSampler in `tests/samplers_tests/test_samplers.py`?
> 
> I locally confirm that adding the sampler to
> 
> * `parametrize_multi_objective_sampler`: no test failed errors.
> * `parametrize_sampler`: one failed case `test_sample_single_distribution`. `BotorchSampler` takes `nan` values
> * `parametrize_relative_sampler`: the introduced tests by this PR failed. This is because the tests uses `dict` to store distributions as a search space, but botorch assumes that `OrderedDict` in `sample_relative`: https://github.com/optuna/optuna/blob/20b3fa30750d96511cdd34a67f8a94a6127bb5b6/optuna/integration/botorch.py#L431

@nzw0301 Thank you for pointing it out and I'm sorry for the delayed response.
I needed to fix `BoTorchSampler` to resolve the error in the second item, so I create a new PR to fix it.
https://github.com/optuna/optuna/pull/2928

After merging it, I'll create a follow-up PR to fix the error in the third item.

nzw0301 (approved):
LGTM. Thanks!

hvy:
Seems like it's been stale although it's been approved. Let me merge it.

PR #2903, opened 2021-09-06
Add crossover operators for nsga2

yoshinobc:
## Motivation
- Until now, the only crossover operator in nsga2 was uniform crossover.
- The uniform crossover does not take into account the ordering relationship of variables, and is therefore weak in real number optimization.
- For this reason, we implemented several crossover operators used in real number optimization.

## Benchmark Results
I experimented with nasbench and ZDT on kurobako with 128 trial.
The following methods were compared.

- [BoTorch] BoTorch Sampler
- [NSGA-II-old] Previous NSGA-II Sampler
- [NSGA-II-kurobako] NSGA-II in kurobako
- [NSGA-II-uniform] New NSGA-II  Sampler
- [NSGA-II-blxalpha] New NSGA-II  Sampler with BLX-alpha crossover
- [NSGA-II-sbx] New NSGA-II  Sampler with SBX crossover
- [NSGA-II-vsbx] New NSGA-II  Sampler with vSBX crossover
- [NSGA-II-undx] New NSGA-II  Sampler with UNDX crossover
- [NSGA-II-spx] New NSGA-II  Sampler with SPX crossover

The detailed experimental commands are described at the end.
All images are arranged in the following order.
NSGA-II-blxalpha(upper left), NSGA-II-kurobako(upper center),  NSGA-II-old(upper right), NSGA-II-sbx(middle left), NSGA-II-spx(middle center), NSGA-II-undx(middle right), NSGA-II-uniform(lower left), NSGA-II-vsbx(lower center), BoTorch(lower right)

**NASBench (128 trial)**
nasbench-128

**ZDT1 (128 trial)**
zdt1-128

**ZDT2 (128 trial)**
zdt2 (1)

**ZDT3 (128 trial)**
zdt3-128

**ZDT4 (128 trial)**
zdt4-128

**ZDT5 (128 trial)**
zdt5-128

**ZDT6 (128 trial)**
zdt6-128

## Description of the changes
- Delete `optuna.samplers._nsga2.py` and create `optuna.samplers._nsga2.__init__.py` , `optuna.samplers._nsga2.crossover.py` , and `optuna.samplers._nsga2` . sampler.py.
- `sampler.py` implements the algorithm of `optuna.samplers._nsga2.py` .
- In `crossover.py` , the algorithm for each crossover operator are implemented.
    - The crossovers we implemented are `blxalpha`, `sbx`, `vsbx`, `undx`, and `spx`.
    - I'm trying to do parent selection in the crossover operator.

- This PR introduces a new `intersection search space` in NSGA2. This introduction allows the NSGAII sampler aligns other relative samplers. However, changes have been made from the previous behavior of NSGA2 during crossover.

Difference: Apply crossover to p0 and p1.
**before**
- If p0 and p1 both have the parameter that you want to crossover, then either p0 or p1 will be chosen.
- If p0 has the parameter to be crossover and p1 does not then a parameter generated from p0 or `random_sampler.sample_independent` is chosen.
-  If p1 has the parameter you want to cross and p0 does not, then p1 is chosen.
- if you don't have both parameters you want to cross, the parameter generated from {random_sampler.sample_independent{ will be chosen

**after**
-  If p0 and p1 both have the parameter that you want to cross, then either p0 or p1 will be chosen.
- Otherwise, the parameter generated from random_sampler's sample_independent will be chosen.

```
pip install git+https://github.com/yoshinobc/optuna.git@add-crossover-for-nsga2
pip install git+https://github.com/yoshinobc/kurobako-py.git@optuna-multi-objective
pip install botorch
kurobako --version
kurobako problem nasbench nasbench_full.bin --encoding C --metrics params accuracy > problems.json
kurobako problem-suite zdt >> problems.json
kurobako solver --name 'BoTorch' command python kurobakobotorch.py > solvers.json
kurobako solver --name 'NSGA-II-kurobako' nsga2 >> solvers.json
kurobako solver --name 'NSGA-II-uniform' command python nsga2uniform.py >> solvers.json
kurobako solver --name 'NSGA-II-blxalpha' command python nsga2blxalpha.py >> solvers.json
kurobako solver --name 'NSGA-II-sbx' command python nsga2sbx.py >> solvers.json
kurobako solver --name 'NSGA-II-vsbx' command python nsga2vsbx.py >> solvers.json
kurobako solver --name 'NSGA-II-undx' command python nsga2undx.py >> solvers.json
kurobako solver --name 'NSGA-II-spx' command python nsga2spx.py >> solvers.json

kurobako studies --problems $(cat problems.json) --solvers $(cat solvers.json) --repeats 1 --budget 128 | kurobako run -p 8 > result.json

pip uninstall optuna
pip install optuna
kurobako solver --name 'NSGA-II-old' command python nsga2uniform.py > solvers.json
kurobako studies --problems $(cat problems.json) --solvers $(cat solvers.json) --repeats 1 --budget 128 | kurobako run -p 8 > result-nsga2-old.json

cat result-nsga2-old.json >> result.json

cat result.json | grep NASBench | kurobako plot pareto-front --ymax  0.2 --xmax 25000000
cat result.json | grep ZDT1 | kurobako plot pareto-front --ymin 1 --ymax 7
cat result.json | grep ZDT2 | kurobako plot pareto-front --ymin 2 --ymax 7
cat result.json | grep ZDT3 | kurobako plot pareto-front --ymin 0 --ymax 7
cat result.json | grep ZDT4 | kurobako plot pareto-front --ymin 20 --ymax 250
cat result.json | grep ZDT5 | kurobako plot pareto-front --ymin 1 --ymax 6
cat result.json | grep ZDT6 | kurobako plot pareto-front --ymin 5 --ymax 10 --xmin 0.2
```

**nsga2uniform.py**
```
from kurobako import solver
from kurobako.solver.optuna_multi_objective import OptunaSolverFactory
import optuna

def create_study(n_objectives, seed):
    directions = ["minimize" for _ in range(n_objectives)]
    # In 'nsga2{crossover_name}.py', the newly implemented crossover method, we use crossover={crossover_name}.
    sampler = optuna.samplers.NSGAIISampler(population_size=8)
    return optuna.create_study(directions=directions, sampler=sampler)


if __name__ == "__main__":
    factory = OptunaSolverFactory(create_study)
    runner = solver.SolverRunner(factory)
    runner.run()

```


optuna/samplers/__init__.py
```diff
@@ -1,7 +1,7 @@
 from optuna.samplers._base import BaseSampler
 from optuna.samplers._cmaes import CmaEsSampler
 from optuna.samplers._grid import GridSampler
-from optuna.samplers._nsga2 import NSGAIISampler
+from optuna.samplers._nsga2.sampler import NSGAIISampler
```
HideakiImamura, line 4: How about adding this line in `optuna/samplers/_nsga2/__init__.py`? Then `NSGAIISampler` can be imported as before.

optuna/samplers/_nsga2/sampler.py
```diff
@@ -94,6 +96,7 @@ def __init__(
         *,
         population_size: int = 50,
         mutation_prob: Optional[float] = None,
+        crossover_name: str = "uniform",
```
HideakiImamura, line 99: Could you add a document of this argument like others? I think it's better for users to understand the crossover variants.

optuna/samplers/_nsga2/sampler.py
```diff
@@ -154,18 +162,36 @@ def sample_relative(
         generation = parent_generation + 1
         study._storage.set_trial_system_attr(trial_id, _GENERATION_KEY, generation)
 
+        dominates = _dominates if self._constraints_func is None else _constrained_dominates
         if parent_generation >= 0:
-            p0 = self._select_parent(study, parent_population)
+            # crossover
```
HideakiImamura, line 167: Please write down a sentence like `We choose a child based on the specified crossover method`.

optuna/samplers/_nsga2/sampler.py
```diff
@@ -154,18 +162,36 @@ def sample_relative(
...
+            else:
+                mutation_prob = self._mutation_prob
+
+            for param_name in child.keys():
+                param, param_distribution = child[param_name], search_space[param_name]
+                if param is None or self._rng.rand() < mutation_prob:
+                    child[param_name] = self._random_sampler.sample_independent(
+                        study, trial, param_name, param_distribution
+                    )
+            return child
```
HideakiImamura, line 194: `search_space.keys()` is a subset of `child.keys()` since we calculate the search space by `intersection_search_space`. The element of `child.keys()` might be missing in `search_space.keys()` so`search_space[param_name]` might occur the `KeyError`.

How about the following?
```suggestion
            params = {}
            for param_name in child.keys():
                if self._rng.rand() >= mutation_prob:
                    # With probability 1 - mutation_prob, we return the parameter value of the child.
                    params[param_name] = child[param_name]
            return params
```

tests/samplers_tests/test_nsga2.py
```diff
@@ -477,3 +477,302 @@ def test_call_after_trial_of_random_sampler() -> None:
...
+
+    assert len(study.trials) == n_trials
+
+    with pytest.raises(ValueError):
+        sampler = NSGAIISampler(population_size=2, crossover_name="spx")
+        study = optuna.create_study(directions=["minimize"] * n_objectives, sampler=sampler)
+        study.optimize(
+            lambda t: [t.suggest_float(f"x{i}", 0, 1) for i in range(n_objectives)],
+            n_trials=n_trials,
+        )
```
HideakiImamura, line 778: How about using `pytest.mark.parametrize` to reduce the number of codes?

HideakiImamura (commented):
Thanks for the great improvement of NSGA-II! I have several early comments. Could you take a look?

By the way, I think it's better to take some benchmark results. How about leaving results for WFG, ZDT, and kurobako?

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,397 @@
...
+    dominates: Callable[[FrozenTrial, FrozenTrial, Sequence[StudyDirection]], bool],
+) -> List[FrozenTrial]:
+    if crossover_name in ["uniform", "blxalpha", "sbx", "vsbx"]:
+        n_select = 2
+    elif crossover_name in ["undx", "spx"]:
+        n_select = 3
+    elif crossover_name in ["undxm"]:
+        n_select = 4
+    else:
+        raise ValueError(f"{crossover_name} is not exist in optuna.")
```
hvy, line 136: A nitpick. 
```suggestion
        raise ValueError(f"'{crossover_name}' is not a valid crossover name.")
```
It might be overkill but we could also include the list of acceptable crossover names in the message.

tests/samplers_tests/test_nsga2.py
hvy: Tests here are very general besides the negative exception checks (which are great). I think they're good to include in terms of code coverage but we could leave a TODO to move them to a more general test file such as `test_samplers.py` (there are actually two files named as the such, so one of them) to reduce test code.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,397 @@
...
+
+
+def _normalized_x1_to_x2(x1: np.ndarray, x2: np.ndarray) -> np.ndarray:
+    v_12 = x2 - x1
+    m_12 = np.linalg.norm(v_12, ord=2)
+    e_12 = v_12 / np.clip(m_12, 1e-10, None)
+    return e_12
+
+
+def _distance_from_x_to_PSL(x1: np.ndarray, x2: np.ndarray, x3: np.ndarray) -> np.ndarray:
```
hvy, line 308: For lower snake case convention.
```suggestion
def _distance_from_x_to_psl(x1: np.ndarray, x2: np.ndarray, x3: np.ndarray) -> np.ndarray:
```

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,397 @@
...
+    xks = [G + epsilon * (pk - G) for pk in xs]
+    ck = 0
+    for k in range(1, n + 1):
+        ck = rs[k - 1] * (xks[k - 1] - xks[k] + ck)
+    c = xks[-1] + ck
+    return c
+
+
+def _check_in_constraints(params: Dict, search_space: Dict) -> bool:
+    contains_flag = True
```
hvy, line 389: A huge nit but "flag" here seems a bit redundant (too).
```suggestion
    contains = True
```

optuna/samplers/_nsga2/crossover.py
hvy: I now noticed that we are performing a similar check in the sampler constructor. Can we simply omit raising an error here, maybe adding an assertion instead that the given crossover name is valid?

optuna/samplers/_nsga2/sampler.py
hvy: "name" here sounds a bit superfluous. I personally prefer just "crossover" since it's clear especially given the type.

hvy: Looks good to start with just introducing a single new parameter 👍 .

hvy: We're usually quite thoughtful when introducing new algorithms. Concretely, we often choose to raise experimental warnings before we're quite confident about either the API or the optimization performance. It communicates to the user that it should be used with some degree of caution. Thus, I'm wondering if we should do the same thing here, as for the first introduction. For reference, it'd be implemented similar to https://github.com/optuna/optuna/blob/master/optuna/samplers/_tpe/sampler.py#L262-L267. What do you think?

optuna/samplers/_nsga2/sampler.py
```diff
@@ -154,18 +162,36 @@ def sample_relative(
...
+                    dominates,
                 )
             else:
-                p1 = p0
+                child = parent_population[0].params
 
-            study._storage.set_trial_system_attr(
-                trial_id, _PARENTS_KEY, [p0._trial_id, p1._trial_id]
-            )
+            # mutation
```
hvy, line 181: As per convention. https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#block-and-inline-comments-are-supposed-to-be-sentences (it's a good read if you haven't already)
```suggestion
            # Mutation.
```

(Actually, the code is quite self explanatory so I even think you can omit this comment altogether)

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,397 @@
...
+_NUMERICAL_DISTRIBUTIONS = (
+    UniformDistribution,
+    LogUniformDistribution,
+    DiscreteUniformDistribution,
+    IntUniformDistribution,
+    IntLogUniformDistribution,
+)
+
+
+def crossover(
```
hvy, line 31: I'm a bit worried about the maintainability of this function and maybe more so the concrete crossover functions called from here. If you have time, it'd be great if you could add some explanatory comments to the algorithm operations, make variable names more readable, or elaborate/add on the references (I noticed some crossover algorithms were missing reference URLs.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,397 @@
...
+                / np.sqrt(_n - _m)
+                / np.sqrt(_m + 2)
+                / np.sqrt(2)
+            )
+            params_array = _undxm(xs, rng, sigma_xi, sigma_eta)
+        elif crossover_name == "spx":
+            epsilon = np.sqrt(len(xs[0]) + 2)
+            params_array = _spx(xs, rng, epsilon)
+        else:
+            raise ValueError(f"{crossover_name} is not exist in optuna.")
```
hvy, line 106: A ditto but we can assert that this code is never reached.
```suggestion
            assert False
```

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,397 @@
...
+    rs = [np.power(rng.uniform(0, 1), 1 / (k + 1)) for k in range(n)]
+    xks = [G + epsilon * (pk - G) for pk in xs]
+    ck = 0
+    for k in range(1, n + 1):
+        ck = rs[k - 1] * (xks[k - 1] - xks[k] + ck)
+    c = xks[-1] + ck
+    return c
+
+
+def _check_in_constraints(params: Dict, search_space: Dict) -> bool:
```
hvy, line 388: This is a nit given it's a private function, but what do you think about the following change? 
```suggestion
def _is_contained(params: Dict, search_space: Dict) -> bool:
```
"check" usually in Optuna communicates that it checks for something and raises if violated, while "is" communicates that the return is a bool.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,397 @@
...
+    crossover_name: str,
+    study: Study,
+    parent_population: Sequence[FrozenTrial],
+    search_space: Dict[str, BaseDistribution],
+    rng: np.random.RandomState,
+    swapping_prob: float,
+    dominates: Callable[[FrozenTrial, FrozenTrial, Sequence[StudyDirection]], bool],
+) -> Dict:
+
+    while True:
```
hvy, line 41: The containment check/loop break is quite far away from this line. The following may improve readability.
```suggestion
    while True:  # Repeat while parameters lie outside search space boundaries.
```

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,397 @@
...
+
+def crossover(
+    crossover_name: str,
+    study: Study,
+    parent_population: Sequence[FrozenTrial],
+    search_space: Dict[str, BaseDistribution],
+    rng: np.random.RandomState,
+    swapping_prob: float,
+    dominates: Callable[[FrozenTrial, FrozenTrial, Sequence[StudyDirection]], bool],
+) -> Dict:
```
hvy, line 39: Can we concretize the type? If so, it'd be preferable in terms of readability. Something like
```suggestion
) -> Dict[str, Any]:
```
(or the value may be `Union[int, float, CategoricalChoiceType]`)

hvy (commented):
Thanks a lot for the huge work, I hope this'll widen applications to multi-objective optimization even further. I haven't went through the details of the algorithms and their implementations in particular yet, but left some comments after a quick glance. Let me know what you think.

optuna/samplers/_nsga2/sampler.py
```diff
@@ -140,7 +146,9 @@ def reseed_rng(self) -> None:
     def infer_relative_search_space(
         self, study: Study, trial: FrozenTrial
     ) -> Dict[str, BaseDistribution]:
-        return {}
+
+        search_space = intersection_search_space(study)
```
hvy, line 150: It's worth highlighting that this change in introducing the intersection search space, while it aligns this sampler with other relative samplers, it's also introduces a change in behavior for conditional parameters. How about leaving a note about that the PR description, making the change easier to track in the future.

optuna/samplers/_nsga2/sampler.py
```diff
@@ -154,18 +162,36 @@ def sample_relative(
...
+                    study,
+                    parent_population,
+                    search_space,
+                    self._rng,
+                    self._swapping_prob,
+                    dominates,
                 )
             else:
-                p1 = p0
+                child = parent_population[0].params
```
sile, line 179: I'm worried that this new logic could decrease the diversity of children when crossover doesn't occur.
(To the best of my knowledge, `_collect_parent_population` is deterministic and the individuals in `parent_population` and the order of them are the same during the same generation. Thus, `child` seems to always have the same params during the same duration if the crossover is skipped. I'm not 100% sure though)

Could you explain the reason that the first individual in `parent_population` is always selected here?

optuna/samplers/_nsga2/sampler.py
```diff
@@ -154,18 +162,36 @@ def sample_relative(
         generation = parent_generation + 1
         study._storage.set_trial_system_attr(trial_id, _GENERATION_KEY, generation)
 
+        dominates = _dominates if self._constraints_func is None else _constrained_dominates
```
sile, line 165: How about renaming this variable with `dominates_func`?
(IMO, the change improves readability a bit)

optuna/samplers/_nsga2/sampler.py
```diff
@@ -124,9 +127,12 @@ def __init__(
                 " The interface can change in the future.",
                 ExperimentalWarning,
             )
+        if crossover_name not in ["uniform", "blxalpha", "sbx", "vsbx", "undx", "undxm", "spx"]:
+            raise ValueError(f"{crossover_name} is not exist in optuna.")
```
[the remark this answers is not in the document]
sile: Besides, IMO,
- it may be a bit more helpful to include the word "crossover" in the message
- the word "optuna" seems not to contain much information (so could be replaced with another word)
- => e.g., "Crossover '{crossover_name}' isn't supported by NSGAIISampler." seems more user friendly (IMO)

optuna/samplers/_nsga2/sampler.py
yoshinobc: I understand the cause of the `Key Error`.
With this fix, I think the individuals that are returned will have missing parameters due to the probability of `mutation_prob`.
In this case, is it correct that `sample_independent` will be called only for the missing parameter?

HideakiImamura: Yes, I think so. The missing parameter will be sampled by `sample_independent`, which is fallen back to the `_random_sampler.sample_independent`. This is a same logic as before.

optuna/samplers/_nsga2/sampler.py
yoshinobc: I forgot that `_collect_parent_population` is deterministic.
Do the following.
```suggestion
                parent_population_size = len(parent_population)
                child = parent_population[self._rng.choice(parent_population_size)].params
```

optuna/samplers/_nsga2/sampler.py
```diff
@@ -154,18 +177,35 @@ def sample_relative(
...
+                mutation_prob = 1.0 / max(1.0, params_len)
+            else:
+                mutation_prob = self._mutation_prob
+
+            params = {}
+            for param_name in child.keys():
+                if self._rng.rand() >= mutation_prob:
+                    params[param_name] = child[param_name]
+
+            return params
```
sile, line 208: Let me suggest the following change to improve code readability.
But, this is a matter of personal preference. So feel free to reject this change if you like the current code.
```suggestion
            for param_name in child.keys():
                if self._rng.rand() < mutation_prob:
                    # Do mutation.
                    # The value of this key will be newly sampled in the `sample_independent()` method.
                    del child[param_name]

            return child
```

yoshinobc:
### Benchmark Results
**NASBench**
nasbench-c-nsga-ii-old-2a51c25d6a7e91e322ed10b159a0f036117fcc1fdbe164c7d065533ecf461e6d
nasbench-c-nsga-ii-uniform-dba06ffeabefbe76233aaaf00e1b7682d5fd3f3b48708910b91ac17aa0d16ca9
nasbench-c-nsga-ii-blxalpha-859fe24ba5f95578dda01a7e0f593f6baa8a5b633a0d43e1047057cc83bf16ce
nasbench-c-nsga-ii-sbx-e8e1429592b15f73a109f3b68ae9df3b2295280db774454a684eebc5af4b5e97
nasbench-c-nsga-ii-vsbx-87b8f2dafdbbaf151a5378db4a0f1268c932364c72ce63a6a5ff7e4f93178efd
nasbench-c-nsga-ii-undx-ff4b08d415a33df7c01c819b77820da1750f1a4b2f2223949190cdb4eefdb4ad
nasbench-c-nsga-ii-undxm-5854363938c030a609d44d275dbdfe3884cc648cf63f6bb670686f0d84b8b886
nasbench-c-nsga-ii-spx-65c846af26282f28a1f763d8479d6dd74d278a74e3b164982c402ec1b393bb0b

yoshinobc:
**ZDT1**
128
zdzt1-128
1000
zdzt1-1000

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,411 @@
...
+    c = xks[-1] + ck  # equation (7)
+    return c
+
+
+def _is_constrained(params: Dict[str, Any], search_space: Dict[str, BaseDistribution]) -> bool:
+    contains = True
+    for param_name in params.keys():
+        param, param_distribution = params[param_name], search_space[param_name]
+        if isinstance(param_distribution, CategoricalDistribution):
+            continue
```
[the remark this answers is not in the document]
sile: Relating to the above question, I just remembered that `param` should be converted to the internal representation (by calling `Distribution.to_internal_repr()` method) before passing to `_contains()`.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+    search_space: Dict[str, BaseDistribution],
+    rng: np.random.RandomState,
+    swapping_prob: float,
+    dominates: Callable[[FrozenTrial, FrozenTrial, Sequence[StudyDirection]], bool],
+) -> Dict[str, Any]:
+
+    while True:  # Repeat while parameters lie outside search space boundaries.
+        parents = _selection(crossover_name, study, parent_population, rng, dominates)
+        child = {}
+        transes = []
```
HideakiImamura, line 44: I think we can define one instance of `_SearchSpaceTransform` here. It is better for code readability.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+        distributions = []
+        param_names = []
+        parents_not_categorical_params: List[List[np.float64]] = [[] for _ in range(len(parents))]
```
HideakiImamura, line 47: Please use not `numpy.float64` but `float`.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+        for param_name in search_space.keys():
+            param_distribution = search_space[param_name]
+            parents_param = [p.params[param_name] for p in parents]
```
HideakiImamura, line 50: It seems that the plural form is appropriate.
```suggestion
            parent_params = [p.params[param_name] for p in parents]
```

optuna/samplers/_nsga2/crossover.py
HideakiImamura: What is the meaning of this variable `parents_not_categorical_params`? It is difficult for me to understand the variable from its name.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+            params_array = _spx(xs, rng, epsilon)
+        else:
+            assert False
+
+        _params = [
+            trans.untransform(np.array([param])) for trans, param in zip(transes, params_array)
+        ]
+        for param in _params:
+            for param_name in param.keys():
+                child[param_name] = param[param_name]
```
HideakiImamura, line 115: I'm not sure but the length of `param` is 1, right? I think it is better for code readability to calculate `_params` as not list but dict. This refactoring is related to the above comment (instantiating one `_SearchSpaceTransform`).

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+        n_select = 3
+    elif crossover_name in ["undxm"]:
+        n_select = 4
+    else:
+        assert False
+    if len(parent_population) < n_select:
+        raise ValueError(
+            f"Using {crossover_name},"
+            f"the population size should be greater than or equal to {n_select}."
+        )
```
HideakiImamura, line 142: Is it documented that the `ValueError` occurs with an invalid `parent_population`?  In Optuna, all possible errors should be documented like [here](https://github.com/optuna/optuna/blob/master/optuna/samplers/_tpe/sampler.py#L214)

HideakiImamura: On a different note, do we need to raise the `ValueError` here? This function is called internally, so it seems to be sufficient that we assert the condition like `assert len(parent_population) >= n_select`.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+            for param_name in param.keys():
+                child[param_name] = param[param_name]
+
+        if _is_contained(child, search_space):
+            break
+
+    return child
+
+
+def _selection(
```
HideakiImamura, line 123: The name of the function is ambiguous. How about `_select_parents`?

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+    crossover_name: str,
+    study: Study,
+    parent_population: Sequence[FrozenTrial],
```
HideakiImamura, line 126: The name of this argument differs from that of `_select_parent` function. Is it intended? I think it is better for code readability to use the same name.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+        return candidate0
+    else:
+        return candidate1
+
+
+def _swap(p0_i: Any, p1_i: Any, rand: float, swapping_prob: float) -> Any:
+    if rand < swapping_prob:
+        return p1_i
+    else:
+        return p0_i
```
HideakiImamura, line 175: I think we don't have to define this utility function, since we can write this in one line like `p1_i if rand < swapping_prob else p0_i`.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+
+def _blxalpha(
+    x1: np.ndarray, x2: np.ndarray, rng: np.random.RandomState, alpha: float
+) -> np.ndarray:
+    # https://www.sciencedirect.com/science/article/abs/pii/B9780080948324500180
+
+    # http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.465.6900&rep=rep1&type=pdf
+    # Section 2 Crossover Operators for RCGA 2.1 Blend Crossover
+
+    assert x1.shape == x2.shape
```
HideakiImamura, line 199: Do we have any other assumptions on the shape of input? For example, `len(x1.shape) == 1`?

HideakiImamura: This comment will be applied in other functions.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+    for x1_i, x2_i in zip(x1, x2):
+        param = _swap(x1_i, x2_i, rng.rand(), swapping_prob)
+        child.append(param)
+    return np.array(child)
+
+
+def _blxalpha(
+    x1: np.ndarray, x2: np.ndarray, rng: np.random.RandomState, alpha: float
+) -> np.ndarray:
+    # https://www.sciencedirect.com/science/article/abs/pii/B9780080948324500180
```
HideakiImamura, line 194: Do we need this reference? Isn't the second one enough?

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+
+def _sbx(
+    x1: np.ndarray,
+    x2: np.ndarray,
+    rng: np.random.RandomState,
+    distributions: List[BaseDistribution],
+    eta: float,
+) -> np.ndarray:
+    # https://content.wolfram.com/uploads/sites/13/2018/02/09-2-2.pdf
+    # https://www.slideshare.net/paskorn/simulated-binary-crossover-presentation
```
HideakiImamura, line 221: Ditto.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+    rng: np.random.RandomState,
+    sigma_xi: float,
+    sigma_eta: float,
+) -> np.ndarray:
+    # https://ieeexplore.ieee.org/document/782672
+    # Section 2 Unimodal Normal Distribution Crossover
+
+    assert x1.shape == x2.shape == x3.shape
+
+    def _normalized_x1_to_x2(x1: np.ndarray, x2: np.ndarray) -> np.ndarray:
```
HideakiImamura, line 338: I think we should leave comments for these utility functions for code readability.

HideakiImamura: Especially, for "psl".

HideakiImamura (commented):
Thanks for the update! I have several comments on `crossover.py`. PTAL.

HideakiImamura:
And, could you resolve the conflicts? It will be resolved by merging the master branch and fixing the conflicts by hand.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,456 @@
...
+    dominates: Callable[[FrozenTrial, FrozenTrial, Sequence[StudyDirection]], bool],
+) -> List[FrozenTrial]:
+    if crossover_name in ["uniform", "blxalpha", "sbx", "vsbx"]:
+        n_select = 2
+    elif crossover_name in ["undx", "spx"]:
+        n_select = 3
+    elif crossover_name in ["undxm"]:
+        n_select = 4
+    else:
+        assert False
```
himkt, line 137: [nits] It would be user friendly to raise an exception with message.

optuna/samplers/_nsga2/crossover.py
yoshinobc: Thank you for your comment.
I have a question about here.
How can `assert` and `ValueError` be used in different ways?

`ValuError` : notifies the user of an expected bug and prompts him to change the hyperparameters.
`assert` : detects a bug that the developer did not anticipate (the developer assumes that the assert will not cause the program to terminate)

I used ValueError here because I thought it was above.

hvy (changes requested):

hvy:
It seems like the code gets stuck in an infinite loop for numerical/float distribution that is single. Seems to be a floating point issue. The bug is not present in `master`, by the way.

```python
import optuna as ot
from optuna.samplers import NSGAIISampler

samplers = [
    #NSGAIISampler(population_size=10),
    NSGAIISampler(population_size=10, crossover="uniform"),
    #NSGAIISampler(population_size=10, crossover="blxalpha"),
    #NSGAIISampler(population_size=10, crossover="sbx"),
    #NSGAIISampler(population_size=10, crossover="vsbx"),
    #NSGAIISampler(population_size=10, crossover="undx"),
    #NSGAIISampler(population_size=10, crossover="undxm"),
    #NSGAIISampler(population_size=10, crossover="spx"),
]


def objective(t):
    p = t.suggest_float("x", 1, 1)
    return p, p * 2


if __name__ == "__main__":
    #studies = []

    for sampler in samplers:
        print(sampler._crossover)
        study = ot.create_study(directions=["minimize", "minimize"], sampler=sampler)
        #studies.append(study)

    #for study in studies:
    #    print()
    #    for t in study.best_trials:
    #        print(t.values, t.params)
```

Output

```
uniform
[I 2021-09-21 16:26:53,238] A new study created in memory with name: no-name-4c9e28e4-5a86-4f0b-90f9-dd47a0fbca53
[I 2021-09-21 16:26:53,239] Trial 0 finished with values: [1.0, 2.0] and parameters: {'x': 1.0}.
[I 2021-09-21 16:26:53,239] Trial 1 finished with values: [1.0, 2.0] and parameters: {'x': 1.0}.
[I 2021-09-21 16:26:53,239] Trial 2 finished with values: [1.0, 2.0] and parameters: {'x': 1.0}.
[I 2021-09-21 16:26:53,239] Trial 3 finished with values: [1.0, 2.0] and parameters: {'x': 1.0}.
[I 2021-09-21 16:26:53,240] Trial 4 finished with values: [1.0, 2.0] and parameters: {'x': 1.0}.
[I 2021-09-21 16:26:53,240] Trial 5 finished with values: [1.0, 2.0] and parameters: {'x': 1.0}.
[I 2021-09-21 16:26:53,241] Trial 6 finished with values: [1.0, 2.0] and parameters: {'x': 1.0}.
[I 2021-09-21 16:26:53,241] Trial 7 finished with values: [1.0, 2.0] and parameters: {'x': 1.0}.
[I 2021-09-21 16:26:53,241] Trial 8 finished with values: [1.0, 2.0] and parameters: {'x': 1.0}.
[I 2021-09-21 16:26:53,242] Trial 9 finished with values: [1.0, 2.0] and parameters: {'x': 1.0}.
# Stuck here since `_is_contained` is always `False` with `child` being ` {'x': 0.9999999999999999}`.
```

optuna/samplers/_nsga2/crossover.py
yoshinobc: Thank you for your comment.
In sampler.py, around line 201, I am checking if the specified crossover is the one that is implemented.
Therefore, the assert in line 137 is written in assert since it is inherently unreachable.

optuna/samplers/_nsga2/__init__.py
```diff
@@ -0,0 +1,4 @@
+from optuna.samplers._nsga2.sampler import NSGAIISampler
+
+
+__all__ = ["NSGAIISampler"]
```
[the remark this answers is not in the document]
yoshinobc: Here, I used `__all__`, referring to `optuna.samplers.__init__.py`.
However, since `_tpe`, which is a private module as well, did not use `__all__` specifically, we want to leave `__init__.py` here empty.

tests/samplers_tests/test_nsga2.py
```diff
@@ -477,3 +478,117 @@ def test_call_after_trial_of_random_sampler() -> None:
...
+
+parametrize_nsga2_sampler = pytest.mark.parametrize(
+    "sampler_class",
+    [
+        NSGAIISampler(population_size=2, crossover="uniform"),
+        NSGAIISampler(population_size=2, crossover="blxalpha"),
+        NSGAIISampler(population_size=2, crossover="sbx"),
+        NSGAIISampler(population_size=2, crossover="vsbx"),
+        NSGAIISampler(population_size=3, crossover="undx"),
+        NSGAIISampler(population_size=3, crossover="spx"),
```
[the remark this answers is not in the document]
yoshinobc: Yes, `undxm` is intentionally left out.
This is because, unlike other crossover operators, `undxm` will cause errors depending on the number of dimensions of the search space. (In `undxm`, the number of parent individuals used for crossover is `m`, and the number of dimensions of the search space is `n`, where `n > m-2`).
Therefore, to test for `ValueError`, we remove `undxm` around line 500 and test it separately with `test_undxm_objective` and `test_undxm_dim` functions.

But
- Not mentioned in the document.
- Not enough testing.

So, we will address the above two points.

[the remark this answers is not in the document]
hvy: Sounds good. Also, I think you're asserting for the condition in the implementation but maybe it's better replaced with an error (e.g. `RuntimeError`) since it's not an implementation detail but something that should be clearly communicated to the library user.

tests/samplers_tests/test_nsga2.py
```diff
@@ -477,3 +478,117 @@ def test_call_after_trial_of_random_sampler() -> None:
...
+    sampler = NSGAIISampler(population_size=4, crossover="undxm")
+    study = optuna.create_study(directions=["minimize"] * n_objectives, sampler=sampler)
+    study.optimize(objective1_dim3, n_trials=n_trials)
+
+    assert len(study.trials) == n_trials
+
+    with pytest.raises(ValueError):
+        sampler = NSGAIISampler(population_size=2, crossover="undxm")
+        study = optuna.create_study(directions=["minimize"] * n_objectives, sampler=sampler)
+        study.optimize(objective1_dim2, n_trials=n_trials)
```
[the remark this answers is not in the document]
yoshinobc: Since positive ones have not been tested for undxm (The other crossovers are all positive `@pytest.mark.parametrize("n_params", [1, 2, 3])`.), we will separate them into negative and positive ones.

optuna/samplers/_nsga2/__init__.py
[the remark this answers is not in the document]
hvy: Seems like this comment isn't yet addressed.

optuna/samplers/_nsga2/__init__.py
```diff
@@ -0,0 +1 @@
+from optuna.samplers._nsga2.sampler import NSGAIISampler  # NOQA
```
hvy, line 1: How about keeping it blank as previously mentioned to keep indirections minimal.
```suggestion
```

optuna/samplers/_nsga2/sampler.py
```diff
@@ -140,7 +248,16 @@ def reseed_rng(self) -> None:
...
     ) -> Dict[str, BaseDistribution]:
-        return {}
+        search_space: Dict[str, BaseDistribution] = {}
+        for name, distribution in self._search_space.calculate(study).items():
+            if distribution.single():
+                # The `untransform` method of `optuna._transform._SearchSpaceTransform`
+                # does not assume a single value,
+                # so single value objects are not sampled with the `sample_relative` method,
+                # but with the `sample_ independent` method.
+                continue
```
hvy, line 258: Nice comment about _why_ (rather than _what_) those are skipped. 👍

optuna/samplers/_nsga2/__init__.py
yoshinobc: If it is kept blank, won't samplers that used to be importable with `from optuna.samplers._nsga2 import NSGAIISampler` become unimportable?

optuna/samplers/_nsga2/sampler.py
```diff
@@ -154,18 +271,39 @@ def sample_relative(
...
+                    dominates_func,
                 )
             else:
-                p1 = p0
+                parent_population_size = len(parent_population)
+                child_params = {}
+                for param_name in search_space.keys():
+                    child_params[param_name] = parent_population[
+                        self._rng.choice(parent_population_size)
+                    ].params[param_name]
```
hvy, line 294: It might be fine but did you intentionally change the logic to use different parents for different genes/params?

yoshinobc: With the previous code, it was possible for the child to have keys that were not in `search_space.keys()`, which would lead to bugs.
Therefore, only the keys in `search_space.keys()` are retrieved from the parent.

However, as you said, with this code, the child may have different parameters from different parent individuals, so I have to improve the code as follows.
```suggestion
                parent_params = parent_population[self._rng.choice(parent_population_size)].params
                for param_name in search_space.keys():
                    child_params[param_name] = parent_params[param_name]
```

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,503 @@
...
+                sigma_eta,
+            )
+        elif crossover_name == "undxm":
+            _m = len(parents_numerical_params_array) - 2
+            _n = len(parents_numerical_params_array[0])
+            if _n - _m <= 0:
+                raise RuntimeError(
+                    "undxm must be n > m-2,"
+                    " where m is the number of parental individuals to be selected"
+                    " at crossover (m=4) and n is the number of parameters of the individuals."
```
hvy, line 133: This error message is perhaps not that easy to comprehend for casual users of this sampler and I think it can be simplified a bit. 

```
f"'undxm' crossover requires that the number of non-categorical parameters in the intersection search space is at least 3. Actual: {_n}"
```

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,503 @@
...
+def _uniform(
+    x1: np.ndarray, x2: np.ndarray, rng: np.random.RandomState, swapping_prob: float
+) -> np.ndarray:
+    # https://www.researchgate.net/publication/201976488_Uniform_Crossover_in_Genetic_Algorithms
+    # Section 1 Introduction
+
+    assert x1.shape == x2.shape
+    assert len(x1.shape) == 1
+
+    child = []
```
hvy, line 219: Ditto. Also, it might be nice to keep return variable names consistent among the different crossover functions.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,503 @@
...
+
+
+def _uniform(
+    x1: np.ndarray, x2: np.ndarray, rng: np.random.RandomState, swapping_prob: float
+) -> np.ndarray:
+    # https://www.researchgate.net/publication/201976488_Uniform_Crossover_in_Genetic_Algorithms
+    # Section 1 Introduction
+
+    assert x1.shape == x2.shape
+    assert len(x1.shape) == 1
```
hvy, line 217: `ndim` might be convenient.
```suggestion
    assert x1.ndim == 1
```

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,503 @@
...
+    # TODO(ohta): Consider crowding distance.
+    if dominates(candidate0, candidate1, study.directions):
+        return candidate0
+    else:
+        return candidate1
+
+
+def _uniform(
+    x1: np.ndarray, x2: np.ndarray, rng: np.random.RandomState, swapping_prob: float
+) -> np.ndarray:
```
hvy, line 212: You can probably vectorize the logic in this function (stay in NumPy-world) instead of splitting up the ndarray, creating a Python list, and then "reconstructing" an ndarray, to make it faster. It can be left as future work. Ditto for other crossover functions.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,503 @@
...
+    assert x1.shape == x2.shape
+    assert len(x1.shape) == 1
+
+    xs = np.stack([x1, x2])
+
+    x_min = xs.min(axis=0)
+    x_max = xs.max(axis=0)
+    diff = alpha * (x_max - x_min)  # Equation (1)
+    low = x_min - diff  # Equation (1)
+    high = x_max + diff  # Equation (1)
```
hvy, line 241: Per convention.
```suggestion
    diff = alpha * (x_max - x_min)  # Equation (1).
    low = x_min - diff  # Equation (1).
    high = x_max + diff  # Equation (1).
```

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,503 @@
...
+    x2: np.ndarray,
+    rng: np.random.RandomState,
+    distributions: List[BaseDistribution],
+    eta: float,
+) -> np.ndarray:
+    # https://www.researchgate.net/profile/M-M-Raghuwanshi/publication/267198495_Simulated_Binary_Crossover_with_Lognormal_Distribution/links/5576c78408ae7536375205d7/Simulated-Binary-Crossover-with-Lognormal-Distribution.pdf
+    # Section 2 Simulated Binary Crossover (SBX)
+
+    # To avoid generating solutions that violate the box constraints,
+    # alpha1, alpha2, xls and xus is introduced, unlike the reference.
```
hvy, line 258:
```suggestion
    # alpha1, alpha2, xls and xus are introduced, unlike the reference.
```

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,503 @@
...
+    _xls = []
+    _xus = []
+    for distribution in distributions:
+        assert isinstance(distribution, _NUMERICAL_DISTRIBUTIONS)
+        _xls.append(distribution.low)
+        _xus.append(distribution.high)
```
hvy, line 264: Ditto about prefixed underscores. Could you either remove them or use different names if variable shadowing is an issue?

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,503 @@
...
+    c2 = 0.5 * ((xs_min + xs_max) + betaq2 * xs_diff)  # Equation (5)
+
+    # SBX applies crossover with establishment 0.5, and with probability 0.5,
+    # the gene of the parent individual is the gene of the child individual.
+    # The original SBX creates two child individuals,
+    # but optuna's implementation creates only one child individual.
+    # Therefore, when there is no crossover,
+    # the gene is selected with equal probability from the parent individuals x1 and x2.
+
+    child = []
```
hvy, line 301: Ditto about naming consistency.

optuna/samplers/_nsga2/crossover.py
hvy: Ditto for similar other occurrences.

hvy (changes requested):

optuna/samplers/_nsga2/sampler.py
hvy: The change looks correct. I think we should rerun the experiment although effects might be small (also given that the crossover probably used before was default 0.9/relative high?) though. Do you have time for this? Also, let me just correct my previous comment. The current logic in a sense does a crossover (although the code branch is for no crossover) so I think we should fix it.

optuna/samplers/_nsga2/__init__.py
hvy: Sorry, what samplers are you referring to? `_nsga2/` has always been private so it shouldn't be an issue. We'd have to update a single line in addition in `samplers/__init__.py`.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,503 @@
...
+            )
+        elif crossover_name == "undxm":
+            _m = len(parents_numerical_params_array) - 2
+            _n = len(parents_numerical_params_array[0])
+            if _n - _m <= 0:
+                raise RuntimeError(
+                    "undxm must be n > m-2,"
+                    " where m is the number of parental individuals to be selected"
+                    " at crossover (m=4) and n is the number of parameters of the individuals."
+                )
```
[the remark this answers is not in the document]
yoshinobc: Here, `m` is not `the number of parents`, but `the number of parents - 2`.
Therefore, to clarify the difference, we set `_m` (`_n` is the same as the usual the number of parameters).

However, since `n` and `m` are difficult to understand, we will change the names of these variables to `n_parents` and `n_params`.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,505 @@
...
+
+    def _normal(rng: np.random.RandomState, ds: List[np.ndarray]) -> np.ndarray:
+        # Compute an orthonormal basis by adding one appropriate vector to ds,
+        # and extract one vector from the orthonormal basis.
+
+        d = rng.normal(0, 1, size=ds[0].shape[0])
+        ds.append(d)
+        X = np.stack(ds)
+        Q, _ = np.linalg.qr(X.T)
+        return Q.T[-1]
```
HideakiImamura, line 437: Ditto. Could you give me some proofs or references of this algorithm?

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,505 @@
...
+
+def _is_contained(params: Dict[str, Any], search_space: Dict[str, BaseDistribution]) -> bool:
+    contains = True
+    for param_name in params.keys():
+        param, param_distribution = params[param_name], search_space[param_name]
+
+        if not param_distribution._contains(param_distribution.to_internal_repr(param)):
+            contains = False
+            break
+    return contains
```
HideakiImamura, line 505: The following is more simple, maybe.
```suggestion
def _is_contained(params: Dict[str, Any], search_space: Dict[str, BaseDistribution]) -> bool:
    for param_name in params.keys():
        param, param_distribution = params[param_name], search_space[param_name]

        if not param_distribution._contains(param_distribution.to_internal_repr(param)):
            return False
    return True
```

HideakiImamura (commented):
Thanks for the update. I have several comments around `crossover.py`. PTAL.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,505 @@
...
+                sigma_eta,
+            )
+        elif crossover_name == "undxm":
+            m = len(parents_numerical_params_array) - 2
+            n_params = len(parents_numerical_params_array[0])
+            if n_params - m <= 0:
+                raise RuntimeError(
+                    "'undxm' crossover requires that the number of non-categorical parameters"
+                    f" in the intersection search space is at least 3. Actual: {n_params}"
+                )
```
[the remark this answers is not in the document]
yoshinobc: Yes, `m` is the number of parents used in the crossover minus 2.
Similarly, yes, `n_params` is the number of non-categorical parameters.

The number of parents used in the crossover (`len(parents_numerical_params_array)`) is fixed to `4`, so `m` is fixed to `2`.

Therefore, if the non-categorical parameters (`n_params`) are more than `3`, I think `RuntimeError` will not be raised.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,505 @@
...
+
+        n = len(x1)
+        e_12 = _normalized_x1_to_x2(x1, x2)  # Normalized vector from x1 to x2.
+        basis_matrix = np.identity(n)
+        if np.count_nonzero(e_12) != 0:
+            v_01 = x1 - np.zeros(len(x1))
+            basis_matrix[0] = v_01 - np.dot(v_01, e_12) * e_12  # Vector orthogonal to e_12.
+        basis_matrix_t = basis_matrix.T
+        Q, _ = np.linalg.qr(basis_matrix_t)
+        return Q.T
```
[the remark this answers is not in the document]
yoshinobc: I'm sorry. I think this implementation contains a mistake (the definition of basis_matrix).

**Old**
The function calculates the orthonormal basis vectors of the subspace orthogonal to PSL (the line connecting `x1` and `x2`).
The first step is to find the subspace orthogonal to the PSL.
The function computes the normalized vector from `x1` to `x2`(`e_12`) on line 395 and the vector `v` orthogonal to the PSL on line 399.
In the current implementation, I first try to find the subspace orthogonal to the PSL.
The function computes the normalized vector from x1 to x2 on line 395 and the vector v orthogonal to the PSL on line 399.
The linearly independent vectors on `R^n` are called `basis_matrix`, and the first column of the `basis_matrix` is `v` (`basis_matrix = {v, e_2, e_3, ..., e_n}`).
The `basis_matrix` is assumed to be a subspace orthogonal to the PSL.
"However, in reality, the basis_matrix is not a subspace orthogonal to the PSL."

**New**
Therefore, we change the definition of `basis_matrix` as follows.
Let the linearly independent vectors on the `R^n` be `basis_matrixes`, and let the first column of the `basis_matrix` is `e_12` (the normalized vector from `x1` to `x2`). (`basis_matrix = {e_12, e_2, e_3, ..., e_n}`).
Then QR decompose the `basis_matrix` to get the orthogonal matrix `Q`.
When the `n` columns of the `basis_matrix` are linearly independent, the first `n` columns of `Q` will be the orthonormal basis of the column space of `basis_matrix` (https://en.wikipedia.org/w/index.php?title=QR_decomposition&oldid=983984255).
Therefore, if we remove the first column from `Q` and denote it as `Q'`, then `Q'` is the orthonormal basis of the subspace orthogonal to `e_12` (I tried to find a reference for this part, but couldn't...). 

```suggestion
    def _orthonormal_basis_vector_to_psl(x1: np.ndarray, x2: np.ndarray) -> np.ndarray:
        # Compute orthogonal basis vectors for the subspace orthogonal to psl.

        n = len(x1)
        e_12 = _normalized_x1_to_x2(x1, x2)  # Normalized vector from x1 to x2.
        basis_matrix = np.identity(n)
        if np.count_nonzero(e_12) != 0:
            basis_matrix[0] = e_12
        basis_matrix_t = basis_matrix.T
        Q, _ = np.linalg.qr(basis_matrix_t)
        return Q.T[1:]
```

yoshinobc:
Thank you very much, I will try to make undx-m into a PR again when I have some more lighting and so on.
Also, I pushed the code that removed undx-m the other day. How about this?

optuna/samplers/_nsga2/sampler.py
```diff
@@ -87,13 +152,18 @@ class NSGAIISampler(BaseSampler):
                 versions without prior notice. See
                 https://github.com/optuna/optuna/releases/tag/v2.5.0.
 
+    Raises:
+        ValueError:
+            If ``crossover`` is `undx` or `spx` and `population_size` is less than or equal to two
+            (To use three parental individuals during crossover applied).
```
HideakiImamura, line 158: I think this should be documented as follows.
- To be raised `ValueError` if ``crossover`` is not in `[uniform, blxalpha, sbx, vsbx, undx, spx]`.
- To be raised `ValueError` if ``population_size`` is smaller than ``n_parents``. The ``n_parents`` is determined by each crossover algorithm like ~~~.

HideakiImamura (commented):
Thanks for the update. Almost LGTM! I have a minor comment for the document.

optuna/samplers/_nsga2/sampler.py
yoshinobc: The documentation has been fixed.

HideakiImamura (approved):
Thanks for the long running PR! LGTM!

HideakiImamura:
Let me leave some TODOs as followups.

- Make crossover functions as Python classes for maintainability.
- Add UNDX-m implementation.
- Add tests to align other samplers' tests (if necessary)

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,431 @@
...
+
+    for key, value in search_space.items():
+        if isinstance(value, _NUMERICAL_DISTRIBUTIONS):
+            numerical_search_space[key] = value
+            numerical_distributions.append(value)
+
+    if len(numerical_distributions) != 0:
+        numerical_transform = _SearchSpaceTransform(numerical_search_space)
+
+    while True:  # Repeat while parameters lie outside search space boundaries.
```
sile, line 52: IMO, this while loop looks too large to grasp the overview and structure of the block without careful reading.
I think it's okay not to address that in this PR but this loop could be simplified by factoring out each sub-block into another function as follows:
```python
    while True:
        child_params = _categorical_params_crossover(search_space, swapping_prob)  # Replace L53..L64
        if len(numerical_distributions) == 0:
            if _is_contained(child_params, search_space):
                break
            continue

        # The following is applied only for numerical parameters.
        parents_numerical_params_array = _transform_numerical_search_space( # Replace L72..L82
           numerical_transform, parents
        )

         if crossover_name == "uniform":
             child_params_array = _uniform_crossover(  # Replace L84..L89
                 parents_numerical_params_array, rng, swapping_prob
             )
        elif crossover_name == "blxalpha":
            child_params_array = _blxalpha_crossover(  # Replace L91..L94
                parents_numerical_params_array, rng
             )
        elif crossover_name == "sbx":
            child_params_array = _sbx_crossover(  # Replace L96..L107
                parents_numerical_params_array, rng, study, numerical_distributions
             )
        elif crossover_name == "vsbx":
            child_params_array = _vsbx_crossover(  # Replace L108..L114
                parents_numerical_params_array, rng, study
             )
        elif crossover_name == "undx":
            child_params_array = _undx_crossover(  # Replace L116..L125
                parents_numerical_params_array, rng
             )
        elif crossover_name == "spx":
            child_params_array = _undx_crossover(  # Replace L127..128
                parents_numerical_params_array, rng
             )
        else:
            assert False

        child_numerical_params = numerical_transform.untransform(child_params_array)
        child_params.update(child_numerical_params)

        if _is_contained(child_params, search_space):
            break
```

sile (changes requested):
Thanks again for your great work.
I took another look at this PR and left minor comments. Please confirm them when you have time 🙏

optuna/samplers/_nsga2/crossover.py
hvy: The code is pretty straight forward so I personally think that the current code is fine, especially if it's going to be refactored into classes, but I do agree it's long and +1 for the change. If you don't want to introduce indirections for each algorithm, or change multiple places on the other hand, I think an alternative is to simply factor out the body of the loop into its own function. Definitely not strong opinions, mostly dropped by for a quick comment.

sile: @hvy Thank you for your comment. "to simply factor out the body of the loop" sounds reasonable.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,431 @@
...
+                    parents[0].params[param_name]
+                    if rng.rand() < swapping_prob
+                    else parents[-1].params[param_name]
+                )
+                child_params[param_name] = param
+
+        if len(numerical_distributions) == 0:
+            if _is_contained(child_params, search_space):
+                break
+            continue
```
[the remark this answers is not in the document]
yoshinobc: The intention of  L67 is that when the objective function has a categorical parameter in the dynamic search space, the parameter (categorical) with the parent individual may not be included in the corresponding param_distribution of the  current search space.
However, after checking past issues, it seems that there is no dynamic search space for categorical parameters, so L67 and L69 should be deleted.

optuna/samplers/_nsga2/crossover.py
yoshinobc: Thank you for your comment.

>> I think an alternative is to simply factor out the body of the loop into its own function. Definitely not strong opinions
Does this mean that each crossover function (e.g. `_uniform`, `_blxalpha`) will have a while loop process(L52 ... L136)?

optuna/samplers/_nsga2/crossover.py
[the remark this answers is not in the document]
sile: I see. Thank you for your explanation. 

> so L67 and L69 should be deleted.

👍

optuna/samplers/_nsga2/crossover.py
sile: > Does this mean ...

No, I had imagined the code like the following:
```python
# New function.
def try_crossover(...) -> Dict[str, Any]:
    ... # Copy code from L53 ~ L133 (but `break` should be replaced with `return child_params`)
    return child_params

def crossover(...) -> Dict[str, Any]:
    ...
    while True:
        child_params = try_crossover(...)
        if _is_contained(child_params, search_space):
            break
    ...
```

yoshinobc: Thank you!
I understand now. I've pushed the fixed version.

hvy: Thanks, looks good! Let me resolve this thread as addressed. 
@sile Please reopen if that'd be relevant.

optuna/samplers/_nsga2/crossover.py
```diff
@@ -0,0 +1,457 @@
...
+    numerical_search_space: Dict[str, BaseDistribution] = {}
+    numerical_distributions: List[BaseDistribution] = []
+
+    for key, value in search_space.items():
+        if isinstance(value, _NUMERICAL_DISTRIBUTIONS):
+            numerical_search_space[key] = value
+            numerical_distributions.append(value)
+
+    if len(numerical_distributions) != 0:
+        numerical_transform = _SearchSpaceTransform(numerical_search_space)
```
hvy, line 145: `numerical_transform` is created conditionally but is always referenced further down. We could ensure that this variable is defined perhaps as an Optional (i.e. possibly hold `None`).

optuna/samplers/_nsga2/sampler.py
```diff
@@ -140,7 +233,16 @@ def reseed_rng(self) -> None:
...
         self, study: Study, trial: FrozenTrial
     ) -> Dict[str, BaseDistribution]:
-        return {}
+        search_space: Dict[str, BaseDistribution] = {}
+        for name, distribution in self._search_space.calculate(study).items():
+            if distribution.single():
+                # The `untransform` method of `optuna._transform._SearchSpaceTransform`
+                # does not assume a single value,
+                # so single value objects are not sampled with the `sample_relative` method,
+                # but with the `sample_ independent` method.
```
hvy, line 242: A nit.
```suggestion
                # but with the `sample_independent` method.
```

hvy (changes requested):

optuna/samplers/_nsga2/crossover.py
yoshinobc: If `len(numerical_distributions)==0` in L144, the `numerical_transform` passed to the `try_crossover` function will be `None`. 
However, if `len(numerical_distributions)==0`, the function will return at L57 in the `try_crossover` function before using `numerical_transform`.

hvy: > If len(numerical_distributions)==0 in L144, the numerical_transform passed to the try_crossover function will be None.

Is this really true? My comment was that this indeed should be the logic, while it currently isn't.

yoshinobc: I'm sorry. I understand.
I need to add to assign `None` to `numerical_transform` if `len(numerical_distributions)==0`.

hvy: Exactly, and no worries, sorry if I wasn't clear.

optuna/samplers/_nsga2/sampler.py
```diff
@@ -124,14 +196,35 @@ def __init__(
...
+                "``crossover`` option is an experimental feature."
+                " The interface can change in the future.",
+                ExperimentalWarning,
+            )
+        n_parents = get_n_parents(crossover)
+        if population_size < n_parents:
+            raise ValueError(
+                f"Using {crossover},"
+                f" the population size should be greater than or equal to {n_parents}."
+                f"The specified `population_size` is {population_size}."
```
hvy, line 216: A nit.
```suggestion
                f" The specified `population_size` is {population_size}."
```

hvy (approved):
Just checking but is it correct that benchmarks have been covered already (c.f. https://github.com/optuna/optuna/pull/2903#issuecomment-964736529) including the relevant, if any, recent changes? Sorry if it's been discussed already. If so, LGTM! I left just a minor nit.

@sile do you want to take another look? Your last suggestions are seemingly addressed, by the way.

sile (approved):
Approved!
@hvy Thank you for the mention to remind me.

hvy:
Thanks @sile for the quick review, and sorry for leaving an ambiguous one. @HideakiImamura could you take a look at my previous review https://github.com/optuna/optuna/pull/2903#pullrequestreview-815477954 and merge if that seems reasonable? @yoshinobc, please feel free to comment as well.

yoshinobc:
Updated benchmarks and revised documentation.

HideakiImamura:
Thank you for your long running work! Let me merge this PR. Great work!!!

sile:
🎉

PR #2905, opened 2021-09-07
[RFC] Add abbreviated JSON formats of distributions

not522:
## Motivation
In the ask CLI, users write JSON formatted distributions. But, the format is complicated to code by hand. The purpose of this PR is to introduce JSON formats of distributions that are more human-readable. 
I think there are other possible formats. Please give me your opinion freely.

- master
```
$ optuna ask --storage sqlite:///mystorage.db --study-name mystudy \
    --search-space '{"x": {"name": "UniformDistribution", "attributes": {"low": 0.0, "high": 1.0}}, "y": {"name": "CategoricalDistribution", "attributes": {"choices": ["foo", "bar"]}}}'
```

- This PR (The master's format also works)
```
$ optuna ask --storage sqlite:///mystorage.db --study-name mystudy \
    --search-space '{"x": {"type": "float", "low": 0.0, "high": 1.0}, "y": {"type": "categorical", "choices": ["foo", "bar"]}}'
```
## Description of the changes
- Deserialize a distribution from an abbreviated format
- Add tests for abbreviated format

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -477,14 +477,52 @@ def json_to_distribution(json_str: str) -> BaseDistribution:
 
     json_dict = json.loads(json_str)
 
-    if json_dict["name"] == CategoricalDistribution.__name__:
-        json_dict["attributes"]["choices"] = tuple(json_dict["attributes"]["choices"])
+    if "name" in json_dict:
+        if json_dict["name"] == CategoricalDistribution.__name__:
+            json_dict["attributes"]["choices"] = tuple(json_dict["attributes"]["choices"])
 
-    for cls in DISTRIBUTION_CLASSES:
-        if json_dict["name"] == cls.__name__:
-            return cls(**json_dict["attributes"])
+        for cls in DISTRIBUTION_CLASSES:
+            if json_dict["name"] == cls.__name__:
+                return cls(**json_dict["attributes"])
 
-    raise ValueError("Unknown distribution class: {}".format(json_dict["name"]))
+        raise ValueError("Unknown distribution class: {}".format(json_dict["name"]))
+
+    else:
+        # Deserialize a distribution from an abbreviated format.
+        if json_dict["type"] == "categorical":
+            return CategoricalDistribution(json_dict["choices"])
+        elif json_dict["type"] in ("float", "int"):
+            low = json_dict["low"]
+            high = json_dict["high"]
+            step = json_dict.get("step")
+            log = json_dict.get("log")
+
+            if json_dict["type"] == "float":
+                if log:
+                    if step:
+                        raise ValueError(
+                            "The parameter `step` is not supported when `log` is true."
+                        )
+                    else:
+                        return LogUniformDistribution(low, high)
+                else:
+                    if step:
+                        return DiscreteUniformDistribution(low, high, step)
+                    else:
+                        return UniformDistribution(low, high)
+            else:
+                if log:
+                    if step:
+                        return IntLogUniformDistribution(low, high, step)
+                    else:
+                        return IntLogUniformDistribution(low, high)
+                else:
+                    if step:
+                        return IntUniformDistribution(low, high, step)
+                    else:
+                        return IntUniformDistribution(low, high)
+
+        raise ValueError("Unknown distribution class: {}".format(json_dict["type"]))
 
 
 def distribution_to_json(dist: BaseDistribution) -> str:
--- a/tests/test_distributions.py
+++ b/tests/test_distributions.py
@@ -34,10 +34,31 @@ EXAMPLE_JSONS = {
     '"attributes": {"low": 2, "high": 12, "step": 2}}',
 }
 
+EXAMPLE_ABBREVIATED_JSONS = {
+    "u": '{"type": "float", "low": 1.0, "high": 2.0}',
+    "l": '{"type": "float", "low": 0.001, "high": 100, "log": true}',
+    "du": '{"type": "float", "low": 1.0, "high": 9.0, "step": 2.0}',
+    "iu": '{"type": "int", "low": 1, "high": 9, "step": 2}',
+    "c1": '{"type": "categorical", "choices": [2.71, -Infinity]}',
+    "c2": '{"type": "categorical", "choices": ["Roppongi", "Azabu"]}',
+    "c3": '{"type": "categorical", "choices": ["Roppongi", "Azabu"]}',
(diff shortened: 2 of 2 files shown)


HideakiImamura:
@hvy @nzw0301 Could you review this PR?

tests/test_distributions.py
```diff
@@ -45,6 +56,16 @@ def test_json_to_distribution() -> None:
     pytest.raises(ValueError, lambda: distributions.json_to_distribution(unknown_json))
 
 
+def test_abbreviated_json_to_distribution() -> None:
+
+    for key in EXAMPLE_ABBREVIATED_JSONS:
+        distribution_actual = distributions.json_to_distribution(EXAMPLE_JSONS[key])
```
hvy, line 62: Ops, I think you meant the following (line might become too long).
```suggestion
        distribution_actual = distributions.json_to_distribution(EXAMPLE_ABBREVIATED_JSONS[key])
```

hvy (changes requested):
Great suggestion! I've only went over the logic which looks good but left a quick comment.

tests/test_distributions.py
not522: Exactly...

optuna/distributions.py
```diff
@@ -477,14 +477,52 @@ def json_to_distribution(json_str: str) -> BaseDistribution:
...
+                        return IntLogUniformDistribution(low, high, step)
+                    else:
+                        return IntLogUniformDistribution(low, high)
+                else:
+                    if step:
+                        return IntUniformDistribution(low, high, step)
+                    else:
+                        return IntUniformDistribution(low, high)
+
+        raise ValueError("Unknown distribution class: {}".format(json_dict["type"]))
```
nzw0301, line 525: Since `json_dict["type"]` is a distribution's type either `int`, `float`, or `categorical`, instead of Optuna's distribution names, how about the following message?

```suggestion
        raise ValueError("Unknown distribution type: {}".format(json_dict["type"]))
```

tests/test_distributions.py
```diff
@@ -45,6 +56,16 @@ def test_json_to_distribution() -> None:
     pytest.raises(ValueError, lambda: distributions.json_to_distribution(unknown_json))
 
 
+def test_abbreviated_json_to_distribution() -> None:
+
+    for key in EXAMPLE_ABBREVIATED_JSONS:
+        distribution_actual = distributions.json_to_distribution(EXAMPLE_ABBREVIATED_JSONS[key])
+        assert distribution_actual == EXAMPLE_DISTRIBUTIONS[key]
+
+    unknown_json = '{"type": "UnknownDistribution", "low": 1.0, "high": 2.0}'
```
nzw0301, line 65: [Nit] In this test,  a similar name to `int`, `float`, or `categorical` is supposed to be better than  
`*Distribution` since this test is for `type` not `name`. For example, 

```suggestion
    unknown_json = '{"type": "unknown", "low": 1.0, "high": 2.0}'
```

nzw0301 (approved):
LGMT! I've left minor comments for an error message and a test case.

not522:
Thank you for your suggestions. I updated the PR.

nzw0301:
Sorry, one more thing. How about adding a test for exception when `type="float"` with `step != 1` and `log=True` from [this line](https://github.com/optuna/optuna/pull/2905/files#diff-2ff77057715ff5d7b689aa1ba37562d8a47052b826e0267493bfdd13ac7206e3R503-R505)?

not522:
Thank you. I added a test. And I noticed that tests for distribution lack int distributions with default step. I sent another PR #2924 because it is related to the master's code.

optuna/distributions.py
```diff
@@ -477,14 +477,52 @@ def json_to_distribution(json_str: str) -> BaseDistribution:
...
+                if log:
+                    if step:
+                        return IntLogUniformDistribution(low, high, step)
+                    else:
+                        return IntLogUniformDistribution(low, high)
+                else:
+                    if step:
+                        return IntUniformDistribution(low, high, step)
+                    else:
+                        return IntUniformDistribution(low, high)
```
toshihikoyanase, line 523: I found that `step=0` is implicitly changed to `step=1`. 

```python
>>> optuna.distributions.json_to_distribution('{"type": "int", "low": 1, "high": 10, "step": 0}')
{'type': 'int', 'low': 1, 'high': 10, 'step': 0}
IntUniformDistribution(high=10, low=1, step=1)
```

If I tried to create `IntUniformDistribution` with `step=0`, the following error occurred.
How about raising `ValueError` if `step=0`?

```python
>>> optuna.distributions.IntUniformDistribution(low=1, high=10, step=0)
  File "/home/yanase/pfn/code/optuna-review/optuna/distributions.py", line 268, in __init__
    raise ValueError(
ValueError: The `step` value must be non-zero positive value, but step=0.
```

I found the same issue in `IntUniformDistribution` and `UniformDistribution`.

`IntLogUniformDistribution`:
```python
>>> optuna.distributions.json_to_distribution('{"type": "int", "low": 1, "high": 10, "step": 0, "log": true}')
{'type': 'int', 'low': 1, 'high': 10, 'step': 0, 'log': True}
IntLogUniformDistribution(high=10, low=1, step=1)
```

`UniformDistribution`:
```python
>>> optuna.distributions.json_to_distribution('{"type": "float", "low": 1, "high": 10, "step": 0}')
{'type': 'float', 'low': 1, 'high': 10, 'step': 0}
UniformDistribution(high=10.0, low=1.0)
```

toshihikoyanase (commented):
Thank you for your 

I summarized the type conversions of numerical distributions in the following table. I think The names of types are aligned with https://github.com/optuna/optuna/issues/2941, and they look reasonable.

| `Distribution` | `type` | `log` | `step` |
| --- | --- |  ---  | --- |
| `UniformDistribution` | `float` | None | None |
| `LogUniformDistribution` | `float` | `True` | None |
| `DiscreteUniformDistribution` | `float` | None | `step > 0` | 
| `IntUniformDistribution` | `int` | None | None or `step > 0` |
| `IntLogUniformDistribution` | `int` | `True` | None or `step > 0` |

In addition, hydra optuna plugin uses the same types (i.e., `float`, `int` and `categorical`) with the same attributes (i.e., `low`, `high`, `log`, `step` and `choices`) to describe distributions in config files.
https://hydra.cc/docs/plugins/optuna_sweeper/#configuring-through-config-file

I have some minor comments on the test cases and the behavior when `step=0`. Please take a look.

PR #2910, opened 2021-09-08
Speed-up sampler tests by using random sampling of skopt

toshihikoyanase:
## Motivation

Applies @not522 's suggestion in https://github.com/optuna/optuna/pull/2869#discussion_r702550677.

They found that `scikit-optimize` took a large portion of test time of `tests/sampler_tests/tests_samplers.py`, and we can reduce the test time by using the random sampling instead of Gaussian Process.

skopt-specific behavior is tested in `tests/integration_tests/test_skopt.py`, and I think it is reasonable to use the random sampling in `tests/integration_tests/test_skopt.py`.

## Description of the changes

- Set `"base_estimator": "dummy"` to `skopt_kwargs` of `SkoptSampler`. It reduced the test time from 12 sec to 8 sec (about 1.5 times faster).


HideakiImamura:

HideakiImamura (approved):
Thanks for the quick fix! LGTM!

not522 (approved):
LGTM.

PR #2912, opened 2021-09-08
Extend `MLflowCallback` interface

xadrianzetx:
## Motivation

Previously users had limited control over mlflow experiment and run initialization, while using `MLflowCallback`. This PR extends its functionality by allowing all `mlflow.start_run` arguments via `mlflow_kwargs` argument (similar to [#2781](https://github.com/optuna/optuna/pull/2781#discussion_r664086151)) and by allowing to decouple experiment name from study name (default behavior is not changed).

This is last task mentioned in #2788.

## Description of the changes

* [x] Add `mlflow_kwargs`argument 
* [x] Decouple experiment name from study name (conditionally)
* [x] ~~Remove `tag_study_user_attrs` argument~~ This has been moved outside of the scope of this PR.
* [x] Write tests
* [x] Update docstrings

### DIFF
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -248,7 +248,16 @@ class MLflowCallback(object):
         tags["direction"] = directions if len(directions) != 1 else directions[0]
 
         distributions = {(k + "_distribution"): str(v) for (k, v) in trial.distributions.items()}
-        tags.update({**distributions, **trial.user_attrs, **study.user_attrs})
+        study_attrs = study.user_attrs
+        trial_attrs = trial.user_attrs
+        collisions = study_attrs.keys() & trial_attrs.keys()
+
+        for key in collisions:
+            # make colliding trial and study user attributes unique
+            study_attrs["study_{}".format(key)] = study_attrs.pop(key)
+            trial_attrs["trial_{}".format(key)] = trial_attrs.pop(key)
+
+        tags.update({**distributions, **trial_attrs, **study_attrs})
 
         # This is a temporary fix on Optuna side. It avoids an error with user
         # attributes that are too long. It should be fixed on MLflow side later.
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -133,7 +133,6 @@ class MLflowCallback(object):
         tracking_uri: Optional[str] = None,
         metric_name: Union[str, Sequence[str]] = "value",
         nest_trials: bool = False,
-        tag_study_user_attrs: bool = False,
     ) -> None:
 
         _imports.check()
@@ -148,7 +147,6 @@ class MLflowCallback(object):
         self._tracking_uri = tracking_uri
         self._metric_name = metric_name
         self._nest_trials = nest_trials
-        self._tag_study_user_attrs = tag_study_user_attrs
 
     def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
 
@@ -234,12 +232,8 @@ class MLflowCallback(object):
         directions = [d.name for d in study.directions]
         tags["direction"] = directions if len(directions) != 1 else directions[0]
 
-        tags.update(trial.user_attrs)
         distributions = {(k + "_distribution"): str(v) for (k, v) in trial.distributions.items()}
-        tags.update(distributions)
-
-        if self._tag_study_user_attrs:
-            tags.update(study.user_attrs)
+        tags.update({**distributions, **trial.user_attrs, **study.user_attrs})
 
         # This is a temporary fix on Optuna side. It avoids an error with user
         # attributes that are too long. It should be fixed on MLflow side later.
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -137,8 +137,8 @@ class MLflowCallback(object):
             will supersede existing tags.
 
     .. note::
-        ``nest_trials`` argument is now a part of ``mlflow_kwargs``. Anyone using
-        ``nest_trials=True`` should migrate to ``mlflow_kwargs={"nested": True}``
+        ``nest_trials`` argument added in v2.3.0 is a part of ``mlflow_kwargs`` since v2.11.0.
+        Anyone using ``nest_trials=True`` should migrate to ``mlflow_kwargs={"nested": True}``
         to avoid raising :exc:`TypeError`.
 
     Raises:
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -117,13 +117,13 @@ class MLflowCallback(object):
             returned by objective function e.g. two objectives and default metric name
             will be logged as ``value_0`` and ``value_1``.
         create_experiment:
-            When ``True``, new MLflow experiment will be created for each optimization run,
-            named after the Optuna study. Setting this argument to ``False`` lets user run
+            When :obj:`True`, new MLflow experiment will be created for each optimization run,
+            named after the Optuna study. Setting this argument to :obj:`False` lets user run
             optimization under existing experiment, set via `mlflow.set_experiment
             <https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.get_tracking_uri>`_,
             by passing ``experiment_id`` as one of ``mlflow_kwargs`` or under default MLflow
             experiment, when no additional arguments are passed. Note that this argument
-            must be set to ``False`` when using Optuna with this callback within
+            must be set to :obj:`False` when using Optuna with this callback within
             Databricks Notebook.
         mlflow_kwargs:
             Set of arguments passed when initializing MLflow run.
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -132,6 +132,7 @@ class MLflowCallback(object):
         self,
         tracking_uri: Optional[str] = None,
         metric_name: Union[str, Sequence[str]] = "value",
+        create_experiment: bool = True,
         mlflow_kwargs: Optional[Dict[str, Any]] = None,
     ) -> None:
 
@@ -146,6 +147,7 @@ class MLflowCallback(object):
 
         self._tracking_uri = tracking_uri
         self._metric_name = metric_name
+        self._create_experiment = create_experiment
         self._mlflow_kwargs = mlflow_kwargs or {}
 
     def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
@@ -210,8 +212,8 @@ class MLflowCallback(object):
         if self._tracking_uri is not None:
             mlflow.set_tracking_uri(self._tracking_uri)
 
-        # This sets the experiment of MLflow.
-        mlflow.set_experiment(study.study_name)
+        if self._create_experiment:
+            mlflow.set_experiment(study.study_name)
 
     def _set_tags(self, trial: optuna.trial.FrozenTrial, study: optuna.study.Study) -> None:
         """Sets the Optuna tags for the current MLflow run.
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -132,7 +132,7 @@ class MLflowCallback(object):
         self,
         tracking_uri: Optional[str] = None,
         metric_name: Union[str, Sequence[str]] = "value",
-        nest_trials: bool = False,
+        mlflow_kwargs: Optional[Dict[str, Any]] = None,
     ) -> None:
 
         _imports.check()
@@ -146,7 +146,7 @@ class MLflowCallback(object):
 
         self._tracking_uri = tracking_uri
         self._metric_name = metric_name
-        self._nest_trials = nest_trials
+        self._mlflow_kwargs = mlflow_kwargs or {}
 
     def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
 
@@ -154,8 +154,10 @@ class MLflowCallback(object):
 
         with mlflow.start_run(
             run_id=trial.system_attrs.get(RUN_ID_ATTRIBUTE_KEY),
-            run_name=str(trial.number),
-            nested=self._nest_trials,
+            experiment_id=self._mlflow_kwargs.get("experiment_id"),
+            run_name=self._mlflow_kwargs.get("run_name") or str(trial.number),
+            nested=self._mlflow_kwargs.get("nested") or False,
+            tags=self._mlflow_kwargs.get("tags"),
         ):
 
             # This sets the metrics for MLflow.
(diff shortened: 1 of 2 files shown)


optuna/integration/mlflow.py
```diff
@@ -132,8 +132,8 @@ def __init__(
         self,
         tracking_uri: Optional[str] = None,
         metric_name: Union[str, Sequence[str]] = "value",
-        nest_trials: bool = False,
-        tag_study_user_attrs: bool = False,
```
xadrianzetx, line 136: After reading #1918 I'm proposing to remove this argument and log study user attributes by default. Reasons:
* Trial user attributes are always logged anyway. This introduces discrepancies in behavior.
* User must explicitly set study attributes via `study.set_user_attr` anyway. In my opinion this reduces chance for accidentally cluttering mlflow experiment.
* After manually setting attributes, it's reasonable for user to expect that they will be logged, without additional "are you sure?" question to answer in callback.
* Complicates API.
* Wasn't really tested.

Alternatively, we can make both trial and study attributes optional, to at least unify behavior. Let me know what you guys think.

xadrianzetx:
As a general note - tests for this module need to be refactored, as there are problems with code duplication and assertions exceeding what is suggested by test name. I can do it, but it's probably out of scope for this PR.

optuna/integration/mlflow.py
nzw0301: Thank you for your proposal. it makes sense. Let unify the logging *`.user_attrs` by removing the argument.

toshihikoyanase:
@nzw0301 Could you continue the review of this PR, please?

optuna/integration/mlflow.py
```diff
@@ -117,23 +116,34 @@ def objective(trial):
...
-            Flag indicating whether or not trials should be logged as
-            nested runs. This is often helpful for aggregating trials
-            to a particular study, under a given experiment.
-        tag_study_user_attrs:
-            Flag indicating whether or not to add the study's user attrs
-            to the mlflow trial as tags. Please note that when this flag is
-            set, key value pairs in :attr:`~optuna.study.Study.user_attrs`
-            will supersede existing tags.
+        create_experiment:
+            When ``True``, new MLflow experiment will be created for each optimization run,
```
nzw0301, line 120: [nit] For sphinx based documentation,

```suggestion
            When :obj:`True`, new MLflow experiment will be created for each optimization run,
```

optuna/integration/mlflow.py
```diff
@@ -117,23 +116,34 @@ def objective(trial):
...
+            named after the Optuna study. Setting this argument to ``False`` lets user run
```
nzw0301, line 121:
```suggestion
            named after the Optuna study. Setting this argument to :obj:`False` lets user run
```

optuna/integration/mlflow.py
```diff
@@ -117,23 +116,34 @@ def objective(trial):
...
+            optimization under existing experiment, set via `mlflow.set_experiment
+            <https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.get_tracking_uri>`_,
+            by passing ``experiment_id`` as one of ``mlflow_kwargs`` or under default MLflow
+            experiment, when no additional arguments are passed. Note that this argument
+            must be set to ``False`` when using Optuna with this callback within
```
nzw0301, line 126:
```suggestion
            must be set to :obj:`False` when using Optuna with this callback within
```

nzw0301 (commented):
I'll give further review in a few days, but let me leave minor comments for documentation.

optuna/integration/mlflow.py
xadrianzetx: Great catch, thanks!

optuna/integration/mlflow.py
```diff
@@ -234,12 +247,8 @@ def _set_tags(self, trial: optuna.trial.FrozenTrial, study: optuna.study.Study)
         directions = [d.name for d in study.directions]
         tags["direction"] = directions if len(directions) != 1 else directions[0]
 
-        tags.update(trial.user_attrs)
         distributions = {(k + "_distribution"): str(v) for (k, v) in trial.distributions.items()}
-        tags.update(distributions)
-
-        if self._tag_study_user_attrs:
-            tags.update(study.user_attrs)
+        tags.update({**distributions, **trial.user_attrs, **study.user_attrs})
```
nzw0301, line 251: In my understanding, if `trial.user_attrs` and `study.user_attrs` have the same key, i.e., `contributors` in the following example, `study.user_attrs`'s value overwrites `trial.user_attrs`'s value, `"contributors": "mamu"`, in MLflow's logging procedure.


```python
def objective(trial):
    trial.set_user_attr("contributors", ["akiba", "sano"])
    trial.study.set_user_attr("contributors", "mamu")
    return (x - 2) ** 2
```

I think there are a few ways to avoid this

- just adding a note section to the documentation because I think this is unlikely to happen in practice. 
- adding a warning message in this callback when the same key appears in `trial.user_attr` and `study.user_attr`
- adding prefix to each key: `trial.user_attr. contributors` and `study.user_attr. contributors` to make sure that keys are unique.

cc: @toshihikoyanase

nzw0301 (commented):
Basically, LGTM. I have one concern on duplicated key among `trial.user_attr` and `trial.study.user_attr`. What do you think about it?

optuna/integration/mlflow.py
xadrianzetx: Thank you for spotting this! I think we could prefix colliding keys to avoid this, since it will not complicate logic by too much.

nzw0301: Indeed. I agree with you!

xadrianzetx: Awesome! Patched in 40fafab38f0b9487d05dadc8cda7cc5ee8eaa293.

nzw0301 (approved):
LGTM! Thansk!

optuna/integration/mlflow.py
```diff
@@ -234,12 +247,17 @@ def _set_tags(self, trial: optuna.trial.FrozenTrial, study: optuna.study.Study)
...
         distributions = {(k + "_distribution"): str(v) for (k, v) in trial.distributions.items()}
-        tags.update(distributions)
+        study_attrs = study.user_attrs
+        trial_attrs = trial.user_attrs
+        collisions = study_attrs.keys() & trial_attrs.keys()
+
+        for key in collisions:
+            # make colliding trial and study user attributes unique
+            study_attrs["study_{}".format(key)] = study_attrs.pop(key)
+            trial_attrs["trial_{}".format(key)] = trial_attrs.pop(key)
```
toshihikoyanase, line 258: The duplicated attribute keys are implicitly renamed here. I'm concerned about the following two points:

- Users may not find the renamed attributes. This behavior is not documented.
- The renamed keys can collide with the other keys

For the second point, `study_foo` in trial's user attributes are overwritten by the value of `foo` in study's user attributes.

```python
import optuna


def objective(trial):
    trial.set_user_attr("foo", "baz")
    trial.set_user_attr("study_foo", "baz")
    return (x - 2) ** 2


mlflc = MLflowCallback(
    tracking_uri=None,
    metric_name="my metric score",
    mlflow_kwargs={}
)

study = optuna.create_study()
study.set_user_attr("foo", "bar")
study.optimize(objective, n_trials=10, callbacks=[mlflc])
```

I may be conservative, but the current master's behavior looks intuitive to me.

optuna/integration/mlflow.py
```diff
@@ -117,23 +116,34 @@ def objective(trial):
...
+            by passing ``experiment_id`` as one of ``mlflow_kwargs`` or under default MLflow
+            experiment, when no additional arguments are passed. Note that this argument
+            must be set to :obj:`False` when using Optuna with this callback within
+            Databricks Notebook.
+        mlflow_kwargs:
+            Set of arguments passed when initializing MLflow run.
+            Please refer to `MLflow API documentation
+            <https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.start_run>`_
+            for more details.
+
```
toshihikoyanase, line 133: I agree to remove the `nest_trials` argument since users can replace it with `mlflow_kwargs={"nested": True}`. How about adding a note to migrate the existing code with `nest_trials`?

toshihikoyanase (commented):
I'm sorry for the delayed response.

I have a comment on the removal of `tag_study_user_attrs` and a suggestion for migration. Please take a look.

optuna/integration/mlflow.py
xadrianzetx: My [reason to remove](https://github.com/optuna/optuna/pull/2912#discussion_r704658783) `tag_study_user_attrs` was based on different behavior when trial/study user attributes are logged (one conditionally, other one always), which I found counter intuitive. Collisions can happen in both configurations, but probably are rare cases, as @nzw0301 noted.

Anyway I'm fine to revert those changes to master.

toshihikoyanase: > different behavior when trial/study user attributes are logged (one conditionally, other one always)

Thank you for your explanation.
Hmm, if the mlflow has a target of tags corresponding to Optuna's study, the situation would be simple.

I guess users who add many user attributes to studies may want to turn off the record of studys' user attributes to keep the mlflow experiment clean. The `tag_study_user_attrs` flag will help such users.

> Collisions can happen in both configurations, but probably are rare cases, as @nzw0301 noted.

I agree with you in terms of the frequency. On the other hand, Optuna is getting more and more users, and such rare cases tend to be reported to the issues and gitter. Currently, the priority is written in the document, and users can resolve issues by themselves.

> Please note that when this flag is set, key value pairs in study.user_attrs will supersede existing tags.

https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.MLflowCallback.html#optuna.integration.MLflowCallback 

> Anyway I'm fine to revert those changes to master.

@nzw0301 What do you think of it?

nzw0301: Thank both of you for your comments. 

> The renamed keys can collide with the other keys
Indeed, I missed this point.

Users might also use `trial.user_attrs` for such a rate case. So if we do not save `study.user_attr` automatically, both `trial.user_attr` and `study.user_attr` should be conditional as suggested in https://github.com/optuna/optuna/pull/2912#discussion_r704658783.

toshihikoyanase: @nzw0301 
> So if we do not save study.user_attr automatically, both trial.user_attr and study.user_attr should be conditional as suggested in #2912 (comment).

Thank you for your comment. Sounds reasonable.

By the way, I think introducing `mlflow_kwargs` and handling of `trial.user_attrs` and `study.user_attrs` are separable development items. How about focusing on the `mlflow_kwargs` argument in this PR and creating a new PR for user attributes?

xadrianzetx: >By the way, I think introducing mlflow_kwargs and handling of trial.user_attrs and study.user_attrs are separable development items. How about focusing on the mlflow_kwargs argument in this PR and creating a new PR for user attributes?

Makes sense, let me revert changes to `*user_attrs` for now.

xadrianzetx: Done in b822511b391ffb8a1c580b1873fc94594c9f4c22.

optuna/integration/mlflow.py
xadrianzetx: Note to arguments added in 6a1aabd299900a88b0df944f6e5253a8c6f30ba6.

tests/integration_tests/test_mlflow.py
```diff
@@ -96,6 +96,37 @@ def test_study_name(tmpdir: py.path.local) -> None:
     assert first_run_dict["data"]["tags"]["my_user_attr"] == "my_user_attr_value"
 
 
+@pytest.mark.parametrize("name,expected", [(None, "Default"), ("foo", "foo")])
+def test_use_existing_or_default_experiment(
+    tmpdir: py.path.local, name: Optional[str], expected: str
+) -> None:
+
+    if name is not None:
+        tracking_file_name = "file:{}".format(tmpdir)
```
toshihikoyanase, line 105: [Note] This is out-of-scope of this PR, but the `tracking_file_name` is actually a URI, not a file name. I think you follow the naming convention of existing test cases, so I accept the change in this PR. Maybe we work on it in a follow-up PR.

optuna/integration/mlflow.py
```diff
@@ -117,22 +116,44 @@ def objective(trial):
...
+            <https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.start_run>`_
+            for more details.
         tag_study_user_attrs:
             Flag indicating whether or not to add the study's user attrs
             to the mlflow trial as tags. Please note that when this flag is
             set, key value pairs in :attr:`~optuna.study.Study.user_attrs`
             will supersede existing tags.
+
+    .. note::
+        ``nest_trials`` argument is now a part of ``mlflow_kwargs``. Anyone using
```
toshihikoyanase, line 140: Thank you for adding the note. A nit-picky, but could you show the Optuna versions which have the `nest_trials` argument? https://github.com/optuna/optuna/pull/1918 introduced the argument, so the version range will be from v2.3.0 to v2.10.0.

optuna/integration/mlflow.py
```diff
@@ -156,8 +178,10 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
 
         with mlflow.start_run(
             run_id=trial.system_attrs.get(RUN_ID_ATTRIBUTE_KEY),
-            run_name=str(trial.number),
-            nested=self._nest_trials,
+            experiment_id=self._mlflow_kwargs.get("experiment_id"),
```
toshihikoyanase, line 181: The current test cases do not specify `experiment_id` in the `mlflow_kwargs` argument. Is it possible to add a test case to set `experiment_id`?

toshihikoyanase (commented):
Thank you for your update!
It basically looks good to me, but I have two small comments.

tests/integration_tests/test_mlflow.py
xadrianzetx: I [commented](https://github.com/optuna/optuna/pull/2912#issuecomment-915462924) about the need to refactor some of the tests and I'm aiming to tackle it some time soon, so will have this comment in mind!

optuna/integration/mlflow.py
xadrianzetx: Done in ed0e78ddd908d607c0479a3f6c26f52ebf02b1af. I tried to follow style used in `deprecate` and `experimental` decorators. Hope it's readable.

optuna/integration/mlflow.py
xadrianzetx: Done in eed521106358dd5ce7fddae56501602959d63bb4.

toshihikoyanase (approved):
Thank you for your update! LGTM.

PR #2917, opened 2021-09-11
Fix an argument in docstring of `_CachedStorage`

nzw0301:
## Motivation

By chance, I found the argument in the docstring of `_CachedStorage` is not consistent with the actual value. 

## Description of the changes

Replace it with the actual argument.


himkt (approved):

toshihikoyanase (approved):
Good catch. Thanks!

PR #2918, opened 2021-09-13
Add feature to get sampler attributes

not522:
## Motivation
The current implementation of `after_trial` and `reseed_rng` has repeated code to call child samplers. This PR will add `get_child_samplers` to refactor these methods.

## Description of the changes
- Add `get_child_samplers`

### DIFF
--- a/optuna/integration/botorch.py
+++ b/optuna/integration/botorch.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 from typing import Any
 from typing import Callable
 from typing import Dict
+from typing import List
 from typing import Optional
 from typing import Sequence
 from typing import Union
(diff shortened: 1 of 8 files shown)


toshihikoyanase:
@Crissman @keisuke-umezawa Could you review this PR if you have time?

I'd like to know the following points:
-  existing user-written samplers work as expected after we merge this PR.
- `child` is the best term for  `get_child_samplers`. I think this is the first time to introduce the idea of parent-child to samplers.

Crissman (approved):
Thanks for the PR, @not522 !

> existing user-written samplers work as expected after we merge this PR.

Seems fine on a code review, since the new `get_child_samplers` is internal, and `reseed_rng` and `after_trial` are still available by inheritance. Both functions are covered by the current test cases.

I also ran our public SimulatedAnnealer on the new code, and it runs fine.

> child is the best term for get_child_samplers. I think this is the first time to introduce the idea of parent-child to samplers.

Yes, it is a new concept, but I think it makes sense from the objective inheritance point of view. Primarily used internally, so I don't think it would confuse users much.


LGTM.

not522:
Discussing with @g-votte , this feature makes samplers complicated. Let me close the PR. Thank you for your reviews.

PR #2922, opened 2021-09-14
Use `:obj:` for `True`, `False`, and `None` instead of inline code

nzw0301:
## Motivation

By chance, I found that a few `True`, `False`, and `None` are not used with `:obj:` in the docstring.

## Description of the changes

Use `:obj:` prefix instead of inline code syntax. 

~Note that the following codes do not use the prefix as well, but they are not shown in Optuna's document. If I should replace them as well, I'm happy to update this pull request!~

I locally confirmed no more such inline code in the Optuna codebase as follows.

### `True`
```bash
grep -i "\`\`True\`\`" -r optuna/
```

### `False`

```bash
grep -i "\`\`False\`\`" -r optuna/
```

### `None`
```bash
 grep -i "\`\`None\`\`" -r optuna/
```

### DIFF
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -146,7 +146,8 @@ class Trial(BaseTrial):
 
                 .. note::
                     The ``step`` and ``log`` arguments cannot be used at the same time. To set
-                    the ``step`` argument to a float number, set the ``log`` argument to ``False``.
+                    the ``step`` argument to a float number, set the ``log`` argument to
+                    :obj:`False`.
             log:
                 A flag to sample the value from the log domain or not.
                 If ``log`` is true, the value is sampled from the range in the log domain.
@@ -155,7 +156,7 @@ class Trial(BaseTrial):
 
                 .. note::
                     The ``step`` and ``log`` arguments cannot be used at the same time. To set
-                    the ``log`` argument to ``True``, set the ``step`` argument to ``None``.
+                    the ``log`` argument to :obj:`True`, set the ``step`` argument to :obj:`None`.
 
         Raises:
             :exc:`ValueError`:
@@ -407,7 +408,7 @@ class Trial(BaseTrial):
                 .. note::
                     The ``step != 1`` and ``log`` arguments cannot be used at the same time.
                     To set the ``step`` argument :math:`\\mathsf{step} \\ge 2`, set the
-                    ``log`` argument to ``False``.
+                    ``log`` argument to :obj:`False`.
             log:
                 A flag to sample the value from the log domain or not.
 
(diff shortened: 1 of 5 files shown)


optuna/trial/_trial.py
```diff
@@ -146,7 +146,8 @@ def objective(trial):
 
                 .. note::
                     The ``step`` and ``log`` arguments cannot be used at the same time. To set
-                    the ``step`` argument to a float number, set the ``log`` argument to ``False``.
+                    the ``step`` argument to a float number, set the ``log`` argument to
+                    :obj:`False`.
```
toshihikoyanase, line 150: 👍 
image

optuna/trial/_trial.py
```diff
@@ -155,7 +156,7 @@ def objective(trial):
 
                 .. note::
                     The ``step`` and ``log`` arguments cannot be used at the same time. To set
-                    the ``log`` argument to ``True``, set the ``step`` argument to ``None``.
+                    the ``log`` argument to :obj:`True`, set the ``step`` argument to :obj:`None`.
```
toshihikoyanase, line 159: 👍 
image

optuna/trial/_trial.py
```diff
@@ -407,7 +408,7 @@ def objective(trial):
                 .. note::
                     The ``step != 1`` and ``log`` arguments cannot be used at the same time.
                     To set the ``step`` argument :math:`\\mathsf{step} \\ge 2`, set the
-                    ``log`` argument to ``False``.
+                    ``log`` argument to :obj:`False`.
```
toshihikoyanase, line 411: 👍 
image

optuna/trial/_trial.py
```diff
@@ -423,7 +424,7 @@ def objective(trial):
 
                 .. note::
                     The ``step != 1`` and ``log`` arguments cannot be used at the same time.
-                    To set the ``log`` argument to ``True``, set the ``step`` argument to 1.
+                    To set the ``log`` argument to :obj:`True`, set the ``step`` argument to 1.
```
toshihikoyanase, line 427: 👍 
image

toshihikoyanase (approved):
Good catch! The change looks good to me. Thanks.

If you fix the rest of the items in `optuna//testing/visualization.py` and `optuna//progress_bar.py`, we can save the future review costs to publish the documents for the modules.

nzw0301:
Thank you for your quick check! I also replace them in `optuna//testing/visualization.py` and `optuna//progress_bar.py`. with minor update that replaces `Study.optimize` with ":func:`~optuna.study.Study.optimize`".

toshihikoyanase (approved):
Thank you for your update. LGTM again.

Issue #2929, opened 2021-09-16
`optuna.create_trial` generates infeasible trials if multiple values are set to `value`


toshihikoyanase:
## Expected behavior

I expect `optuna.create_trial` will raise errors if I set `value=(1, 2)` because I should use the `values` argument instead of `value`. I think the error will be `ValueError` since the input value is inappropriate.

## Environment

- Optuna version: 2.10.0.dev
- Python version: 3.8.10
- OS: Linux-4.19.128-microsoft-standard-x86_64-with-glibc2.29

## Error messages, stack traces, or logs

Tthe current optuna generates the infeasible trial whose `values` is `[(1, 2)]` as follows: 

```console
>>> import optuna
>>> optuna.create_trial(value=(1, 2))
FrozenTrial(number=-1, values=[(1, 2)], datetime_start=datetime.datetime(2021, 9, 16, 17, 19, 44, 238726), datetime_complete=datetime.datetime(2021, 9, 16, 17, 19, 44, 238726), params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=-1, state=TrialState.COMPLETE, value=None)
```

I think it should raise `ValueError`. Alternatively, the `values` should be `[1, 2]` if Optuna does not raise the error.

```console
>>> optuna.create_trial(values=(1, 2))
FrozenTrial(number=-1, values=[1, 2], datetime_start=datetime.datetime(2021, 9, 16, 17, 21, 31, 123218), datetime_complete=datetime.datetime(2021, 9, 16, 17, 21, 31, 123218), params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=-1, state=TrialState.COMPLETE, value=None)
```

## Steps to reproduce

1. Run `python -c 'import optuna; print(optuna.create_trial(value=(1,2)))'`

PR #2931, opened 2021-09-16
Validate `FrozenTrial.values`

toshihikoyanase:
## Motivation

This PR Addresses #2929.

## Description of the changes

Users use `optuna.create_trial` to create `FrozenTrial` instances and it calls `FrozenTrial._validate`. So, I added the value checking to `FrozenTrial._validate` instead of the setter of `FrozenTrial.values`.

- Check `FrozenTrial.values` in `FrozenTrial._validate`.
- Add test cases for `FrozenTrial.values`.

Alternatively, we can check then in . I think

### DIFF
--- a/optuna/trial/_frozen.py
+++ b/optuna/trial/_frozen.py
@@ -350,6 +350,22 @@ class FrozenTrial(BaseTrial):
                     "{}.".format(param_value, param_name, distribution)
                 )
 
+        if self.values is not None:
+            for i, v in enumerate(self.values):
+                # `value` can be None.
+                if v is None:
+                    continue
+                try:
+                    # Check if `values` element is float. This check accepts float-castable values
+                    # including numpy.floating and numpy.integer, which are sometimes returned
+                    # by objective functions.
+                    _ = v + 1
+                except TypeError:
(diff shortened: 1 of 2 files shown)


toshihikoyanase:
@not522 and @hvy suggested that we had some design choices about this change.

- `intermediate_values` should be validated in addition to `values`
- we can also check the trial values in Storage layer
- we should clarify the validation policy before diving into the details

I totally agree with them and make this PR draft.

optuna/trial/_frozen.py
```diff
@@ -350,6 +350,22 @@ def _validate(self) -> None:
...
+        if self.values is not None:
+            for i, v in enumerate(self.values):
+                # `value` can be None.
+                if v is None:
+                    continue
+                try:
+                    # Check if `values` element is float. This check accepts float-castable values
+                    # including numpy.floating and numpy.integer, which are sometimes returned
+                    # by objective functions.
+                    _ = v + 1
```
not522, line 362: `_ = v + 1` accepts complex numbers. How about using `isinstance(v, numbers.Real)` instead?

Issue #2945, opened 2021-09-21
`_SearchSpaceTransform` not expecting float distributions where `low == high`


hvy:
## Expected behavior

The higher bound offset logic introduced to the untransformation in `_SearchSpaceTransform` in https://github.com/optuna/optuna/pull/2243 may not be applied for float distributions where `low == high`, i.e. when those distributions are "single" (`BaseDistribution.single() == True`). 

## Steps to reproduce

1. Run the following example. Notice that an assertion error is raised.

## Reproducible examples (optional)

```python
    search_space = {
        "x0": UniformDistribution(1, 1),  # `low == high`
        "x1": LogUniformDistribution(2, 2),
    }

    params = {
        "x0": 1.0,
        "x1": 2.0,
    }

    trans = _SearchSpaceTransform(search_space)
    trans_params = trans.transform(params)
    untrans_params = trans.untransform(trans_params)


    for name in params.keys():
        assert untrans_params[name] == params[name]  # Fails, but shouldn't.
```

## Additional context (optional)

The most simple solution is to apply `nextafter` only if the distribution is not "single". Alternative is to prohibit "single" distributions in `_SearchSpaceTransform`.

hvy:
@Crissman what do you think, maybe there are better alternatives?

hvy:
Note that it's generally not an issue (in an obfuscated way) because parameter values computed by `sample_{relative,independent}` from distributions that are single are never really used because of this prior branching https://github.com/optuna/optuna/blob/master/optuna/trial/_trial.py#L713-L714.

hvy:
Fixed by https://github.com/optuna/optuna/pull/2947#event-5371218758.

PR #2947, opened 2021-09-22
Fix method `untransform` of `_SearchSpaceTransform` with `distribution.single() == True`

yoshinobc:
## Motivation
The current `untransform` method of `_SearchSpaceTransform` uses the higher bound offset logic, but it may return wrong values when `low==high` in `uniformDistribution` and `LogUniformDistribution`.
Details of this issue are reported in #2945 .

## Description of the changes
-  Do not apply `min()` when `d.single()==True`.
- Add some tests


hvy (approved):
Thanks for the quick, LGTM. The current logic was rather unintended and I think that this can be considered an improvement or even a bug fix. It shouldn't affect any built-in sampler in practice.

Crissman (approved):
LGTM. Merging!

PR #2953, opened 2021-09-24
Introduce the version constraint of scikit-learn

nzw0301:
## Motivation

Scikit-learn 1.0.0 is out. Unfortunately, scikit-optimize does not work with the latest scikit-learn.
CI error example: https://github.com/nzw0301/optuna/runs/3701260684?check_suite_focus=true
See also the reported issue to scikit-optimize repo: https://github.com/scikit-optimize/scikit-optimize/issues/1059

## Description of the changes

Introduce the version constraint of scikit-learn and TODO lines to the config and setup files that use scikit-optimize.

### DIFF
--- a/.github/workflows/mac-tests.yml
+++ b/.github/workflows/mac-tests.yml
@@ -46,11 +46,6 @@ jobs:
         optuna --version
         pip install --progress-bar off .[tests]
         pip install --progress-bar off .[optional]
-
-        # TODO(nzw0301) remove the constraint of sklearn `1.0.0`
-        # after skopt supports sklearn >= 1.0.0
-        pip install "scikit-learn<1.0.0"
-        
         # Required for samplers_tests/test_samplers.py.
         # TODO(hvy): Fix samplers_tests/test_samplers.py to not require optional depenendencies and remove these installs.
         pip install scikit-optimize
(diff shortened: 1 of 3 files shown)


.github/workflows/mac-tests.yml
```diff
@@ -46,6 +46,11 @@ jobs:
         optuna --version
         pip install --progress-bar off .[tests]
         pip install --progress-bar off .[optional]
+
+        # TODO(nzw0301) remove the constraint of sklearn `1.0.0`
+        # after skopt supports sklearn >= 1.0.0
+        pip install "scikit-learn<1.0.0"
```
hvy, line 52: Can or should we not update the `optional` extras requirements in `setup.py` instead?

hvy (commented):
Thanks for the quick fix. I left a single question if you could take a look.

.github/workflows/mac-tests.yml
nzw0301: Definitely. I've updated it. Thanks!

hvy (approved):
Thanks for the prompt response, LGTM!

not522 (changes requested):
Could you add constraints in document section's scikit-learn?

not522 (approved):
Thank you. LGTM!

PR #2961, opened 2021-09-27
Fixes `namedtuple` type name

sobolevn:
## Motivation

The best practice is to match `typename` with variable name.
For example, `mypy` reports any other usages as errors: https://github.com/python/mypy/pull/11206#issuecomment-927985235


hvy:
@toshihikoyanase , @keisuke-umezawa do you have time to review this change?

toshihikoyanase (approved):
Thank you for your PR! It will prevent future CI failure. LGTM!

- The [official Python document](https://docs.python.org/3/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields) also sets the same class name to the `typename` argument, and
- we don't have a particular reason to add `_` to the `typename`s.

keisuke-umezawa (approved):
LGTM!

PR #2849, opened 2021-08-07
Support PyTorch-lightning DDP training

tohmae:
This PR addresses issue #2699.
I corrected the previous PR(#2824), I create a new PR again.
If you have a time, could you review it ?

PR #2977, opened 2021-10-03
AllenNLP distributed pruning

himkt:
This PR changes the way to pass the `trial` object to `AllenNLPPruningCallback`.

- https://github.com/optuna/optuna/issues/1990
- https://github.com/himkt/allennlp-optuna/issues/20

This PR is the another try of https://github.com/optuna/optuna/pull/2796. It tries to support distributed training by porting AllenNLP training module and injecting Optuna stuff. However, the module is actively updated and it is difficult to follow these changes in one file. I had gave up on the approach and was looking for other solutions.

~This PR was heavily inspired by #2849 that adds a DDP support in PyTorch Lightning. A main idea is that a pruner raises `TrialPruned` in the end of training, instead of the end of epoch. A callback on the end of epoch is invoked in a forked process, which means we cannot raise an arbitrary exception because forked process terminates with an error of multiprocessing if it observes an exception. A callback on the end of training is invoked in a main process, which means we can raise any exception and Optuna can catch it correctly. Instad of raising `TrialPruned` in a callback on an epoch end, it sends a signal to make trainer stop a loop (ref. https://github.com/optuna/optuna/pull/2977#discussion_r720828528).~

This PR adds try-catch logic to `AllenNLPExecutor` to catch an exception raised from [PyTorch multiprocessing module](https://pytorch.org/docs/stable/multiprocessing.html#torch.multiprocessing.spawn) and raise `TrialPruned` in the executor. Thanks to this re-raising logic, AllenNLP integration can tell Optuna that a trial finishes to be pruned (Otherwise, `torch.multiprocessing.ProcessRaisedException` is propagated to Optuna optimize function and it causes interruption of optimization).

## Operation verifications
- [x] sqlite3
- [x] mysql
- [x] postgresql


optuna/integration/allennlp/_pruner.py
```diff
@@ -227,4 +245,16 @@ def on_epoch(
...
+    def on_end(
+        self,
+        trainer: "GradientDescentTrainer",
+        metrics: Dict[str, Any] = None,
+        epoch: int = None,
+        is_primary: bool = True,
+        **kwargs: Any,
+    ) -> None:
+        if self.pruned:
             raise optuna.TrialPruned()
```
himkt, line 260: This PR makes a pruning callback raise `TrialPruned` in `on_end`, instead of `on_epoch` (current implementation).
Note that this change is greatly inspired by #2849. Thank you so much for sending a awesome PR @tohmae.

optuna/integration/allennlp/_pruner.py
```diff
@@ -227,4 +245,16 @@ def on_epoch(
 
         self._trial.report(float(value), epoch)
         if self._trial.should_prune():
+            trainer._metric_tracker.should_stop_early = lambda: True  # type: ignore
```
himkt, line 248: To stop a training loop, I replace an AllenNLP built-in stopper with a function that always returns `True`.

optuna/integration/allennlp/_pruner.py
```diff
@@ -165,11 +162,22 @@ def __init__(
...
-            study_name = environment_variables["study_name"]
-            trial_id = environment_variables["trial_id"]
-            monitor = environment_variables["monitor"]
-            storage = environment_variables["storage"]
+            current_process = psutil.Process()
+
+            if os.getenv("OPTUNA_ALLENNLP_USE_DISTRIBUTED") == "1":
+                os.environ.pop("OPTUNA_ALLENNLP_USE_DISTRIBUTED")
+                parent_process = current_process.parent()
+                target_pid = parent_process.ppid()
```
himkt, line 170: Here, I get an id of a grandparent process. This is because pruner initialization is executed on child processes spawned by `torch.multiprocessing` in distributed training.

optuna/integration/allennlp/_executor.py
```diff
@@ -212,6 +210,19 @@ def run(self) -> float:
...
+
+            if "OPTUNA_ALLENNLP_USE_DISTRIBUTED" in os.environ:
+                warnings.warn(
+                    "Other process may already exists."
+                    " If you have some trouble, please unsed an environment"
+                    " variable `OPTUNA_ALLENNLP_USE_DISTRIBUTED`"
+                    " and try it again."
+                )
+
+            os.environ["OPTUNA_ALLENNLP_USE_DISTRIBUTED"] = "1"
```
himkt, line 224: One global variable is set to let a pruner know if it is initialized in distributed scenario.

optuna/integration/allennlp/_variables.py
```diff
@@ -1,22 +1,56 @@
...
-_MONITOR = "{}_MONITOR".format(_PREFIX)
-_PRUNER_CLASS = "{}_PRUNER_CLASS".format(_PREFIX)
-_PRUNER_KEYS = "{}_PRUNER_KEYS".format(_PREFIX)
-_STORAGE_NAME = "{}_STORAGE_NAME".format(_PREFIX)
-_STUDY_NAME = "{}_STUDY_NAME".format(_PREFIX)
-_TRIAL_ID = "{}_TRIAL_ID".format(_PREFIX)
+SPECIAL_DELIMITER = "[OPTUNA_ALLENNLP_INTEGRATION_DELIMITER]"
+
+
+class _VariableManager:
```
himkt, line 8: I introduce a class that manages environment variables specific to the AllenNLP integration.
It aims to extract an ownership of getting/setting environment variables from the executor and the pruner.

optuna/integration/allennlp/_variables.py
```diff
@@ -1,22 +1,97 @@
...
+        name_of_path = "optuna.integration.allennlp._variables._VariableManager.NAME_OF_KEY"
+        assert key is not None, f"{name} is not found in `{name_of_path}`."
+        key = key.format(self.target_pid)
+        value = os.environ.get(key)
+        assert value is not None, f"{key} is not found in environment variables."
+        return value
+
+    def acquire_lock(self, is_distributed: bool) -> None:
+        if "OPTUNA_ALLENNLP_USE_DISTRIBUTED" in os.environ:
+            raise EnvironmentError(
```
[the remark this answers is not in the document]
himkt:
```
os.putenv(key, value)
Set the environment variable named key to the string value. Such changes to the environment affect subprocesses started with os.system(), popen() or fork() and execv().
```

https://docs.python.org/3/library/os.html

In the implementation, `del` runs on forked processes so `OPTUNA_ALLENNLP_USE_DISTRIBUTED` remains even if it removes on a forked process.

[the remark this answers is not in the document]
himkt: I revised bc5bea8, because changing `os.environ` only affects inside a process, and other process doesn't know the change. (it means we don't have to implement any lock)

optuna/integration/allennlp/_executor.py
```diff
@@ -212,12 +213,30 @@ def run(self) -> float:
...
+        try:
+            allennlp.commands.train.train_model(
+                params=params,
+                serialization_dir=self._serialization_dir,
+                file_friendly_logging=self._file_friendly_logging,
+                force=self._force,
+                include_package=self._include_package,
+            )
+        except ProcessRaisedException as e:
+            if "raise TrialPruned()" in str(e):
```
himkt, line 238: Error massage looks like following:

```
-- Process 0 terminated with the following error:
  File "/path/to/optuna/venv/lib/python3.9/site-packages/torch/multiprocessing/spawn.py", line 59, in _wrap
    fn(i, *args)
  File "/path/to/optuna/venv/lib/python3.9/site-packages/allennlp/commands/train.py", line 504, in _train_worker
    metrics = train_loop.run()
  File "/path/to/optuna/venv/lib/python3.9/site-packages/allennlp/commands/train.py", line 577, in run
    return self.trainer.train()
  File "/path/to/optuna/venv/lib/python3.9/site-packages/allennlp/training/gradient_descent_trainer.py", line 750, in train
    metrics, epoch = self._try_train()
  File "/path/to/optuna/venv/lib/python3.9/site-packages/allennlp/training/gradient_descent_trainer.py", line 857, in _try_train
    callback.on_epoch(self, metrics=metrics, epoch=epoch, is_primary=self._primary)
  File "/path/to/optuna/optuna/integration/allennlp/_pruner.py", line 250, in on_epoch
    raise TrialPruned()
optuna.exceptions.TrialPruned
```

I inspect message and check if `"raise` TrialPruned()"` appears in it.

himkt:
I'm so sorry to repeatedly change the approach. As said in https://github.com/optuna/optuna/pull/2977#issuecomment-953927120, I found callback approach doesn't work. So I revised the PR to catch a torch.multiprocessing exception and raise TrialPruned manually. It is little bit dirty hack but I think this approach is most reasonable for the current situation. Please review it again.

himkt:
@toshihikoyanase @HideakiImamura 

I'm so sorry to repeatedly change the approach. As said in https://github.com/optuna/optuna/pull/2977#issuecomment-953927120, I found callback approach doesn't work. So I revised the PR to catch a torch.multiprocessing exception and raise TrialPruned manually. It is little bit dirty hack but I think this approach is most reasonable for the current situation. Please review it again.

optuna/integration/allennlp/_pruner.py
```diff
@@ -155,6 +152,8 @@ def __init__(
                 "please install Optuna v2.5.0 by executing `pip install 'optuna==2.5.0'`."
             )
 
+        self.pruned = False
```
HideakiImamura, line 155: We don't need this anymore?

HideakiImamura (commented):
Thanks for the update. The current logic is fine to me. I have a minor comment. Could you take a look?

optuna/integration/allennlp/_pruner.py
himkt: You're right, removed in 99c91bc. :pray:

HideakiImamura (approved):
LGTM again!

@toshihikoyanase Please merge this PR after your review.

himkt:
Let me kindly ping @toshihikoyanase. 🐱

optuna/integration/allennlp/_variables.py
```diff
@@ -1,22 +1,77 @@
...
+
+    def __init__(self, target_pid: int) -> None:
+        self.target_pid = target_pid
+
+    @property
+    def prefix(self) -> str:
+        return "{}_OPTUNA_ALLENNLP".format(self.target_pid)
+
+    def get_key(self, name: str) -> Optional[str]:
+        return self.NAME_OF_KEY.get(name)
```
toshihikoyanase, line 48: [Note] This class is only used by the developers of this integration, so I think the current design is acceptable. But, I have some comments on the implementation. We can address them in a follow-up if needed.
- The other files do not refer to this function. Can we prefix `_`?
- If we extract the key checking of `NAME_OF_KEY`, we can reduce the duplicated lines in `set_value` and `get_value`.
- Can we raise `KeyError` rather than `AssertionError`?
- If we use `@property`, the interface may be more Pythonic.

```suggestion
    def _get_key(self, name: str) -> Optional[str]:
        name_of_path = "optuna.integration.allennlp._variables._VariableManager.NAME_OF_KEY"
        if key is None:
            KeyError(f"{name} is not found in `{name_of_path}`.")
        return self.NAME_OF_KEY.get(name)
```

optuna/integration/allennlp/_pruner.py
```diff
@@ -129,7 +126,7 @@ class AllenNLPPruningCallback(TrainerCallback):
 
     .. note::
         Currently, build-in pruners are supported except for
-        :class:`~optuna.pruners.PatientPruner`.
+        :class:`~pruners.PatientPruner`.
```
toshihikoyanase, line 129: The `PatientPruner` is not linked correctly. Could you revert the change?

image

```suggestion
        :class:`~optuna.pruners.PatientPruner`.
```

toshihikoyanase (commented):
I'm sorry for the delayed response.
I have a change request and a comment.

optuna/integration/allennlp/_pruner.py
himkt: Right, thank you. :pray:

optuna/integration/allennlp/_variables.py
himkt: Thank you for the suggestion! All comments are reasonable for me.
I revised my PR to follow your comments (except to one).

- 69cc663 makes `_VariableManager.get_key` private
- 51212d9 extracts key checking procedure to `_VariableManager._get_key`
- 40cd34b raises `KeyError` instead of `AssertionError`

I agree with you that implementing getter using property but it would generate a bunch of copied code. So for this PR, I'd keep the current (non-pythonic) interface as is. What do you think? But if you strongly think it would be better to use property, I'll revised as well. 🐱

himkt:
@toshihikoyanase Thank you for the review, I revised PR based on your suggestions. 🐈

optuna/integration/allennlp/_variables.py
toshihikoyanase: Looks great! Thank you for your update.

himkt:
Sorry for changing the approach again and again. Thank you so much for the patient reviews! 🌟

PR #2983, opened 2021-10-05
Simplify the test of pytorch-lightning

nzw0301:
## Motivation

As pointed out in https://github.com/optuna/optuna/pull/2849#pullrequestreview-770976213, the model definition of distributed data-parallel training has many shared definitions to the simple model.

## Description of the changes

`ModelDDP` Inherits the method's definition from `Model`.

### DIFF
--- a/tests/integration_tests/test_pytorch_lightning.py
+++ b/tests/integration_tests/test_pytorch_lightning.py
@@ -45,6 +45,8 @@ class Model(pl.LightningModule):
         return {"validation_accuracy": accuracy}
 
     def validation_epoch_end(self, outputs: List[Dict[str, torch.Tensor]]) -> None:
+        if not len(outputs):
+            return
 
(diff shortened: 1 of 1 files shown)


tests/integration_tests/test_pytorch_lightning.py
```diff
@@ -45,6 +45,8 @@ def validation_step(  # type: ignore
         return {"validation_accuracy": accuracy}
 
     def validation_epoch_end(self, outputs: List[Dict[str, torch.Tensor]]) -> None:
+        if not len(outputs):
+            return
```
nzw0301, line 49: For DDP test, this value is empty. To avoid zero-division of the following `accuracy`'s line, I've introduced this condition.

HideakiImamura:

tests/integration_tests/test_pytorch_lightning.py
HideakiImamura: nice catch 👍

PR #2984, opened 2021-10-05
Add note for DDP training of pytorch-lightning

nzw0301:
## Motivation

This is a follow-up PR of https://github.com/optuna/optuna/pull/2849, concretely https://github.com/optuna/optuna/pull/2849#issuecomment-927324973.

PyTorch-Lightning's ddp training requires 

- pytorch-lightning >= 1.4.0
- instanciated study with `RDBStorage`

However, the documentation has not mentioned them at all.

## Description of the changes

Add a note section to address the requirements above.


optuna/integration/pytorch_lightning.py
```diff
@@ -40,6 +40,12 @@ class PyTorchLightningPruningCallback(Callback):
             ``pytorch_lightning.LightningModule.training_step`` or
             ``pytorch_lightning.LightningModule.validation_epoch_end`` and the names thus depend on
             how this dictionary is formatted.
+
+    .. note::
+        For the distributed data parallel training, ``accelerator="ddp"``,
```
toshihikoyanase, line 45: [Question] PyTorch Lightning supports multiple accelerators (e.g., `dp`, `ddp_spawn`, `ddp2`, and `horovod`) in addition to `ddp`. Does the callback support them?
https://pytorch-lightning.readthedocs.io/en/stable/advanced/multi_gpu.html#distributed-modes

optuna/integration/pytorch_lightning.py
```diff
@@ -40,6 +40,12 @@ class PyTorchLightningPruningCallback(Callback):
...
+        the version of PyTorchLightning needs to be higher than or equal to v1.4.0.
+        In addition, :class:`~optuna.study.Study` should be instantiated with
+        :class:`~optuna.storages.RDBStorage`.
```
HideakiImamura, line 48: Users can use `RDBStorage` without direct instantiation by specifying its URL, such as `study = optuna.create_study(storage="https://..."). How about not mentioning the concrete class name, and just saying "Study should be instantiated with RDB storage."?

HideakiImamura (commented):
Thanks for the followup. I have a minor comment. PTAL.

optuna/integration/pytorch_lightning.py
tohmae: I tested https://github.com/optuna/optuna-examples/pull/43 only in CPU environment and multi GPUs environment(2GPUs on one machine).
In multi GPUs envirioment, if accelerator is not specified in pl.Trainer, accelerator="ddp_spawn" is set.
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/accelerator_connector.py#L752-L757
So I tested only in ddp_spawn and ddp_cpu.

HideakiImamura:
@toshihikoyanase Could you give another review? Please merge this PR after your approval.

Issue #2995, opened 2021-10-09
CIs fail when testing LightGBM integration


himkt:
Currently, tests in LightGBM integration module fail.

```bash
[W 2021-10-09 13:53:50,068] Trial 0 failed because of the following error: TypeError('Data list can only be of ndarray or Sequence')
Traceback (most recent call last):
  File "/home/runner/work/optuna/optuna/optuna/study/_optimize.py", line 213, in _run_trial
    value_or_values = func(trial)
  File "/home/runner/work/optuna/optuna/tests/integration_tests/test_lightgbm.py", line 145, in objective
    callbacks=[pruning_callback],
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/lightgbm/engine.py", line 272, in train
    booster = Booster(params=params, train_set=train_set)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/lightgbm/basic.py", line 2605, in __init__
    train_set.construct()
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/lightgbm/basic.py", line 1819, in construct
    categorical_feature=self.categorical_feature, params=self.params)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/lightgbm/basic.py", line 1545, in _lazy_init
Error:     raise TypeError('Data list can only be of ndarray or Sequence')
TypeError: Data list can only be of ndarray or Sequence
```

https://github.com/optuna/optuna/runs/3846766083#step:7:483

PR #2996, opened 2021-10-09
Add constraint to LightGBM version

himkt:
ref. https://github.com/optuna/optuna/issues/2995

Tests fail when checking the LightGBM integration module.
CI installed `lightgbm==3.3.0`, which was released 12 hours ago.
https://github.com/microsoft/LightGBM/releases/tag/v3.3.0

We might have to fix our codebase, but, I send this PR for the first aid.


himkt:
This problem will be solved in #2997 instead.

PR #3023, opened 2021-10-19
fix: Avoid unnecessary warning

ghost:
The check self.step == 0 will always trigger _warn_step in IntLogUniformDistribution, even if it has the valid value of 1. Replacing the check with self._step == 0 removes this warning, which is anyway correctly raised in the constructor.


valkenburg:
Will do. Thanks for the fast review and feedback!

Issue #2902, opened 2021-09-05
Clean up deprecated features


himkt:
## Motivation

There are several methods/classes annotated with `@deprecated` in Optuna. It's time to consider whether they should be removed or when to remove them.

## Description

- [x] List deprecated features
- [x] Remove APIs whose `removed_version` are less than `v3.0.0` (deprecated in before `v2.0.0`)
- [x] Document `@deprecated` decorator (we remove `_get_removed_version_from_deprecated_version` and simplify it)

nuka137:
List of Classes/Methods with `@deprecated` is here.

|File|Method/Class|Deprecated Version|Remove Version|
|---|---|---|---|
|`integration/_lightgbm_tuner/optimize.py`|`LightGBMTuner::best_booster`|1.4.0|-|
|`integration/catalyst.py`|`CatalystPruningCallback`|2.7.0|-|
|`integration/chainermn.py`|`ChainerMNTrial::suggest_uniform`|2.11.0|6.0.0|
|`integration/chainermn.py`|`ChainerMNTrial::suggest_loguniform`|2.11.0|6.0.0|
|`integration/chainermn.py`|`ChainerMNTrial::suggest_discrete_uniform`|2.11.0|6.0.0|
|`integration/cma.py`|`CmaEsSampler`|2.0.0|-|
|`integration/fastaiv1.py`|`FastAIV1PruningCallback`|2.4.0|-|
|`integration/pytorch_distributed.py`|`TorchDistributedTrial::suggest_uniform`|2.11.0|6.0.0|
|`integration/pytorch_distributed.py`|`TorchDistributedTrial::suggest_loguniform`|2.11.0|6.0.0|
|`integration/pytorch_distributed.py`|`TorchDistributedTrial::suggest_discrete_uniform`|2.11.0|6.0.0|
|`multi_objective/samplers/_base.py`|`BaseMultiObjectiveSampler`|2.4.0|4.0.0|
|`multi_objective/samplers/_motpe.py`|`MOTPEMultiObjectiveSampler`|2.4.0|4.0.0|
|`multi_objective/samplers/_nsga2.py`|`NSGAIIMultiObjectiveSampler`|2.4.0|4.0.0|
|`multi_objective/samplers/_random.py`|`RandomMultiObjectiveSampler`|2.4.0|4.0.0|
|`multi_objective/study.py`|`create_study`|2.4.0|4.0.0|
|`multi_objective/study.py`|`load_study`|2.4.0|4.0.0|
[...]

PR #2559, opened 2021-04-02
Deprecate dashboard

not522:
This PR depends #2558.
## Motivation
The dashboard has not been updated in this half-year, and the new dashboard "[optuna-dashboard](https://github.com/optuna/optuna-dashboard)" has been rolled out. We should encourage migration.
## Description of the changes
Deprecate `optuna dashboard`.

PR #3000, opened 2021-10-12
Speed up `import optuna`

not522:
## Motivation
In the current implementation, `import optuna` slows down by importing libraries. This PR will speed up using lazy import.

```shell
$ time python -c "import optuna"
```

- master

```shell
real	0m2.022s
user	0m1.800s
sys	0m0.757s
```

- PR
```shell
real	0m1.078s
user	0m1.218s
sys	0m0.470s
```

### Profile using [tuna](https://github.com/nschloe/tuna)

```shell
$ python -X importtime <(echo import optuna) 2> import.log
$ tuna import.log
```

- master
image

- PR
image

## Description of the changes
- Move `_LazyImport`
- Lazy importing `optuna.importance`, `optuna.visualization`, `optuna.multi_objective.visualization`, and `optuna.study._dataframe`

PR #3058, opened 2021-10-31
Remove `optuna dashboard`

himkt:
~This PR updates the deprecation period of Optuna built-in dashboard command. `_Dashboard.take_action` is annotated to be removed in v3.0.0 but we're now on the migration to [`optuna-dashbaord`](https://github.com/optuna/optuna-dashboard). So, in this PR, I update `removed_version` to `v4.0.0`.~

~🔗  https://github.com/optuna/optuna/pull/2559#pullrequestreview-626783569~
~🔗  https://github.com/optuna/optuna/issues/2902~

In this PR, I remove `dashboard` subcommand based on its deprecation period.


not522 (approved):
In #2559, we choosed `v3.0.0` as `removed_version`, because this feature is experimental and we may use `optuna dashboard` as alias of `optuna-dashboard`. But this alias will cause circular dependency and we do not introduce it.
The default `removed_version` is the next next major version after the deprecated version, i.e., `v4.0.0`. I agree with this PR.
@g-votte, Do you have any comments on this change?

himkt:
I discussed with @g-votte and @not522 and decided that we remove `dashbaord` command. I revise this PR.

not522 (approved):
LGTM!

setup.py
```diff
@@ -86,7 +86,6 @@ def get_extras_require() -> Dict[str, List[str]]:
         "testing": [
             # TODO(toshihikoyanase): Remove the version constraint after resolving the issue
             # https://github.com/optuna/optuna/issues/1000.
```
nzw0301, line 88: Can we remove these lines since it issue refers to `bokeh` that will be removed in this PR?

himkt: Thank you. I removed in 80df813.

nzw0301: still in https://github.com/optuna/optuna/blob/80df8136168e283a7dab7d93ac6f5ad51bc2ded5/setup.py#L87-L88

himkt: :bow: 59d11ea

g-votte:
`bokeh` still remains here:
https://github.com/optuna/optuna/blob/master/optuna/__init__.py#L64

We may want to replace this with another optional dependency.

himkt:
@g-votte 

> We may want to replace this with another optional dependency.

`_LazyImport`  is not used only for `optuna.type_checking` and we don't have example of `_lazyImport` that imports third-party libraries. So I simply removed e.g. from the comment (f076e86). What do you think about it?

g-votte:
@himkt 
Thanks for your quick action. I agree that we can simply get rid of the specific library name from the comment.

Just FYI, lazy import is going to be applied again for third party modules with this PR:
https://github.com/optuna/optuna/pull/3000

g-votte (approved):
LGTM

@nzw0301 
Please merge if you don't have any other concern.

nzw0301:
Thanks! I confirmed that `bokeh` and unrelated `dashboard` are not in this repo anymore.

Issue #2941, opened 2021-09-20
Cleanup distributions: `FloatDistribution` and `IntDistribution`


himkt:
## Motivation

There are six classes that represent distributions in Optuna. For categorical parameters, `CategoricalDistribution` corresponds. For integer parameters, `IntUniformDistribution` and `IntLogUniformDistribution`. `UniformDistribution`, `LogUniformDistribution`, and `DiscreteUniformDistribution` for floating-point parameters.

`IntLoguniformDistribution` can be seen as a special case of `IntUniformDistribution`. As well, `LogUniformDistribution` and `DiscreteUniformDistribution` as of `UniformDistribution`.

In this issue, we will work on cleaning up distributions to three classes `CategoricalDistribution`, `IntDistribution`, and `FloatDistribution` by renaming and unifying the existing distributions. We think it improves the user experience because suggest APIs (`suggest_categorical`, `suggest_int`, `suggest_float`) would simply correspond to distribution names.

## Description

- [x] Introduce `FloatingDistribution` and `IntDistribution` #3063 
- [x] Add `IntDistribution` and `FloatDistribution` to the place where conditional branching is done by `UniformDistribution` etc #3111 #3126
- [x] Implement migration from current storage #3113 
- [x] Deprecate distributions mentioned in the description #2990 
- [x] Replace `UniformDistribution` etc to `IntDistribution` and `FloatDistribution` where instances of the distributions are created #3166 #3181
- [x] Update old distributions such that they inherit from `IntDistribution` and `FloatDistribution` #3244 
- [x] Automatically convert deprecated distribution if it is provided by user #3420 
- [x] Fix distribution compatibility for linear and logarithmic distribution #3444

keisuke-umezawa:
detail considerations for `Determine and implement migration from current storage`

1. Prepare SQL migration script for [distribution_json](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/models.py#L346). ref: [migration script](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/alembic/versions/v2.4.0.a.py)
2. Modify check compatibility such as [_check_compatibility_with_previous_trial_param_distributions](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/models.py#L357)

himkt:
@nyanhi @keisuke-umezawa
Thank you so much for working on this issue.
[...]

PR #3063, opened 2021-10-31
Add IntDistribution & FloatDistribution

nyanhi:
## Motivation
To address the issue #2941, the first task ("Introduce `FloatingDistribution` and `IntDistribution`). 

## Description of the changes
Added 2 classes (`FloatDistribution` and `IntDistribution`) in `optuna.distribution.py` and its tests in `tests.test_distribution.py`.

Test cases seems bit redundant at the moment, and will think if it can be simplified with pytest parametrize etc.


optuna/distributions.py
```diff
@@ -93,6 +95,87 @@ def __repr__(self) -> str:
...
+        if low > high:
+            raise ValueError(
+                "The `low` value must be smaller than or equal to the `high` value "
+                "(low={}, high={}).".format(low, high)
+            )
+
+        if log and step is not None:
+            raise ValueError("The parameter `step` is not supported when `log` is true.")
+
+        if log and low < 0.0:
```
keisuke-umezawa, line 135: Should be `<=`.

optuna/distributions.py
```diff
@@ -233,6 +316,61 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
...
+
+        if log and low < 1:
+            raise ValueError(
+                "The `low` value must be equal to or greater than 1 for a log distribution "
+                "(low={}, high={}).".format(low, high)
+            )
+
+        self.low = low
+        self.high = high
+        self.log = log
```
keisuke-umezawa, line 354: Better to cast them into `int`.

optuna/distributions.py
```diff
@@ -233,6 +316,61 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
...
+        log:
+            If ``log`` is True, this distribution is in log-scaled domain.
+
+    Raises:
+        ValueError:
+            If ``low`` value is larger than ``high`` value,
+            or ``low`` value is less than 1 when ``log`` is :obj:`True.
+    """
+
+    def __init__(self, low: int, high: int, log: bool = False) -> None:
```
keisuke-umezawa, line 339: We also need to add `step` argument here.

optuna/distributions.py
nyanhi: Thank you for prompt feedbacks! I have casted before passing them to `_adjust_int_uniform_high()`. 
Please kindly let me know if any concern.

nyanhi:
Thank you so much for reviewing this! I will try to clean up `tests.test_distribution.py` sometime this week and let you know when I'd like to get another look.

optuna/distributions.py
```diff
@@ -233,6 +316,114 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
...
+                "(low={}, high={}).".format(low, high)
+            )
+
+        if step <= 0:
+            raise ValueError(
+                "The `step` value must be non-zero positive value, but step={}.".format(step)
+            )
+
+        if log and step != 1:
+            self._warn_step()
```
keisuke-umezawa, line 366: Since we will create a new `IntDistribution` class, we can raise `ValueError` here with the comment `"Samplers and other components in Optuna only accept step is 1 when log argument is True"`.

Also, in this case, we can remove some adhoc code such as

- _warn_step
- _asdict

optuna/distributions.py
```diff
@@ -10,6 +10,8 @@
 
 
 CategoricalChoiceType = Union[None, bool, int, float, str]
+_int_distribution_deprecated_msg = "Use :class:`~optuna.distribution.IntDistribution` instead."
```
nzw0301, line 13: typo

```suggestion
_int_distribution_deprecated_msg = "Use :class:`~optuna.distributions.IntDistribution` instead."
```

optuna/distributions.py
```diff
@@ -10,6 +10,8 @@
...
+_float_distribution_deprecated_msg = "Use :class:`~optuna.distribution.FloatDistribution` instead."
```
nzw0301, line 14: ditto
```suggestion
_float_distribution_deprecated_msg = "Use :class:`~optuna.distributions.FloatDistribution` instead."
```

optuna/distributions.py
```diff
@@ -89,10 +91,91 @@ def __hash__(self) -> int:
 
     def __repr__(self) -> str:
 
-        kwargs = ", ".join("{}={}".format(k, v) for k, v in sorted(self.__dict__.items()))
+        kwargs = ", ".join("{}={}".format(k, v) for k, v in sorted(self._asdict().items()))
         return "{}({})".format(self.__class__.__name__, kwargs)
 
 
+class FloatDistribution(BaseDistribution):
+    """A continuous uniform distribution in a linear domain.
```
nzw0301, line 99: By specifying `step`, the distribution becomes discrete. In addition, when we use `log=True`, the domain is log-scale, not linear domain. How about this?

```suggestion
    """A distribution on floats.
```

I'm not sure this is the optimal explanation. 
What do you think about it?

optuna/distributions.py
```diff
@@ -89,10 +91,91 @@ def __hash__(self) -> int:
...
+
+    Attributes:
+        low:
+            Lower endpoint of the range of the distribution. ``low`` is included in the range.
+        high:
+            Upper endpoint of the range of the distribution. ``high`` is included from the range.
+        log:
+            If ``log`` is :obj:`True`, this distribution is in log-scaled domain.
+        step:
+            A discretization step. This parameter must be None
```
nzw0301, line 112:
```suggestion
            A discretization step. This parameter must be :obj:`None`
```

optuna/distributions.py
```diff
@@ -89,10 +91,91 @@ def __hash__(self) -> int:
...
+            when the parameter ``log`` is :obj:`True`.
+
+    Raises:
+        ValueError:
+            If ``low`` value is larger than ``high`` value.
+            If both ``log`` is :obj:`True` and ``step`` is not None.
```
nzw0301, line 118:
```suggestion
            If ``log`` is :obj:`True` and ``step`` is not :obj:`None`.
```

optuna/distributions.py
```diff
@@ -89,10 +91,91 @@ def __hash__(self) -> int:
...
+            return self.low == self.high
+        else:
+            if self.low == self.high:
+                return True
+            high = decimal.Decimal(str(self.high))
+            low = decimal.Decimal(str(self.low))
+            step = decimal.Decimal(str(self.step))
+            if (high - low) < step:
+                return True
+            return False
```
nzw0301, line 167: simplefied
```suggestion
            return (high - low) < step
```

optuna/distributions.py
```diff
@@ -233,6 +316,114 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
         return self.low <= value <= self.high and abs(k - round(k)) < 1.0e-8
 
 
+class IntDistribution(BaseDistribution):
+    """A discrete uniform distribution on integers.
```
nzw0301, line 320: Similary, 
```suggestion
    """A distribution on integers.
```

nzw0301 (dismissed):
Let me leave minor comments.

optuna/distributions.py
```diff
@@ -233,6 +316,114 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
...
+
+    This object is instantiated by :func:`~optuna.trial.Trial.suggest_float`, and passed to
```
nzw0301, line 322: Sorry, one more 

```suggestion
    This object is instantiated by :func:`~optuna.trial.Trial.suggest_int`, and passed to
```

optuna/distributions.py
nyanhi: Thank you for pointing out! I agree that the class is much more versatile than my explanation. Let me update as you suggested.

optuna/distributions.py
nyanhi: Thank you for pointing out!

nyanhi:
I have incorporated the changes suggested, and also tried to clean up the tests for the new distribution classes. However if you have any better idea, please kindly let me know so that I can help amending it.
Given the existing classes are expected to be deprecated soon, I tried not to make much changes.

optuna/distributions.py
```diff
@@ -89,10 +93,89 @@ def __hash__(self) -> int:
 
     def __repr__(self) -> str:
 
-        kwargs = ", ".join("{}={}".format(k, v) for k, v in sorted(self.__dict__.items()))
+        kwargs = ", ".join("{}={}".format(k, v) for k, v in sorted(self._asdict().items()))
```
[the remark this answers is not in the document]
keisuke-umezawa: Maybe, better to add a unittest for it.

keisuke-umezawa (commented):
Left some comments for the implementation code. Almost LGTM! I will check unittest code later.

optuna/distributions.py
```diff
@@ -233,6 +316,87 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
...
+    This object is instantiated by :func:`~optuna.trial.Trial.suggest_int`, and passed to
+    :mod:`~optuna.samplers` in general.
+
+    Attributes:
+        low:
+            Lower endpoint of the range of the distribution. ``low`` is included in the range.
+        high:
+            Upper endpoint of the range of the distribution. ``high`` is included from the range.
+        log:
+            If ``log`` is True, this distribution is in log-scaled domain.
```
nzw0301, line 331:
```suggestion
            If ``log`` is :obj:`True`, this distribution is in log-scaled domain.
```

optuna/distributions.py
```diff
@@ -233,6 +316,87 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
...
+        step:
+            A step for spacing between values. When ``log`` is True, ``step`` is assumed to be 1.
```
nzw0301, line 333:
```suggestion
            A step for spacing between values. When ``log`` is :obj:`True`, ``step`` is assumed to be 1.
```

nzw0301 (commented):
I found a few minor things.

optuna/distributions.py
nyanhi: Thank you for pointing out. I will think about a way to avoid such thing going forward.

optuna/distributions.py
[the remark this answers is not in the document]
nyanhi: Thank you for the suggestion, I have reverted the changes. 
I think `test_repr()` includes unittest for the ad-hoc `__repr__` override by `IntLogUniformDistribution` but kindly let me know if I'm not correctly following your point. Thank you!

tests/test_distributions.py
```diff
@@ -35,14 +51,14 @@
...
 EXAMPLE_ABBREVIATED_JSONS = {
-    "u": '{"type": "float", "low": 1.0, "high": 2.0}',
-    "l": '{"type": "float", "low": 0.001, "high": 100, "log": true}',
-    "du": '{"type": "float", "low": 1.0, "high": 9.0, "step": 2.0}',
-    "iu": '{"type": "int", "low": 1, "high": 9, "step": 2}',
+    "i": '{"type": "int", "low": 1, "high": 9, "log": false, "step": 2}',
+    "il": '{"type": "int", "low": 2, "high": 12, "log": true, "step": 1}',
+    "f": '{"type": "float", "low": 1.0, "high": 2.0, "log": false}',
+    "fl": '{"type": "float", "low": 0.001, "high": 100.0, "log": true}',
+    "fd": '{"type": "float", "low": 1.0, "high": 9.0, "step": 2.0, "log": false}',
```
keisuke-umezawa, line 58: Sorry, but we will work on this transformation in another PR. We can revert it.

tests/test_distributions.py
```diff
@@ -82,6 +98,17 @@ def test_abbreviated_json_to_distribution() -> None:
         pytest.raises(ValueError, lambda: distributions.json_to_distribution(distribution))
 
 
+def test_backward_compatibility_int_log_with_step() -> None:
```
keisuke-umezawa, line 101: Ditto.

tests/test_distributions.py
```diff
@@ -123,6 +157,21 @@ def test_check_distribution_compatibility() -> None:
...
+    )
+    distributions.check_distribution_compatibility(
+        EXAMPLE_DISTRIBUTIONS["f"], distributions.FloatDistribution(low=-3.0, high=-2.0)
+    )
+    distributions.check_distribution_compatibility(
+        EXAMPLE_DISTRIBUTIONS["fl"], distributions.FloatDistribution(low=0.1, high=1.0, log=True)
+    )
+    distributions.check_distribution_compatibility(
+        EXAMPLE_DISTRIBUTIONS["fd"], distributions.FloatDistribution(low=-1.0, high=11.0, step=0.5)
+    )
```
keisuke-umezawa, line 174: In the current implementation, i, il and id are compatible each other. To explicitly test it, you can check them here. Also, it can be applied to f, fl and fd.

tests/test_distributions.py
```diff
@@ -388,6 +594,24 @@ def test_int_log_uniform_distribution_asdict() -> None:
...
+
+def test_float_invalid_step() -> None:
+
+    with pytest.raises(ValueError):
+        distributions.FloatDistribution(low=1.0, high=10.0, step=0)
+
+    with pytest.raises(ValueError):
+        distributions.FloatDistribution(low=1.0, high=100.0, step=-1)
+
+
```
keisuke-umezawa, line 614: Could add unittests for ValueError for `FloatDistribution` and `IntDistribution`?

optuna/distributions.py
```diff
@@ -500,29 +667,16 @@ def json_to_distribution(json_str: str) -> BaseDistribution:
...
-                    else:
-                        return IntUniformDistribution(low, high)
+                if log and step != 1:
+                    step = 1
+                    warnings.warn(
+                        "Samplers and other components in Optuna will assume that `step` is 1 "
+                        "when `log` argument is True.",
+                        FutureWarning,
+                    )
+                return IntDistribution(low=low, high=high, log=log, step=step)
```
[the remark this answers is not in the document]
nyanhi: Understood, will include this change into a separate PR. Thank you for explanation.

optuna/distributions.py
```diff
@@ -233,6 +316,88 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
...
+        if low > high:
+            raise ValueError(
+                "The `low` value must be smaller than or equal to the `high` value "
+                "(low={}, high={}).".format(low, high)
+            )
+
+        # For backward compatibility, an abbreviated json for IntUniformDistribution
+        # may not have `log` attribute.
+        if log is None:
+            log = False
```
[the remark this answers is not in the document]
nyanhi: Thank you for the suggestion, that sounds better.

tests/test_distributions.py
nyanhi: Thank you very much for pointing out, added the tests.

optuna/distributions.py
```diff
@@ -233,6 +316,83 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
...
+
+    Raises:
+        ValueError:
+            If ``low`` value is larger than ``high`` value.
+            If ``low`` value is less than 1 when ``log`` is :obj:`True`.
+            If ``step`` is not positive value.
+            If ``log`` is :obj:`True` and ``step``!= 1.
+    """
+
+    def __init__(self, low: int, high: int, log: Union[None, bool] = False, step: int = 1) -> None:
```
keisuke-umezawa, line 344:
```suggestion
    def __init__(self, low: int, high: int, log: bool = False, step: int = 1) -> None:
```

keisuke-umezawa: Sorry for very nits.

keisuke-umezawa (commented):
Just one comment.

tests/test_distributions.py
```diff
@@ -333,6 +498,25 @@ def test_eq_ne_hash() -> None:
         assert not d != d_copy
         assert hash(d) == hash(d_copy)
 
+    # Different field values.
+    di0 = distributions.FloatDistribution(low=1, high=2)
+    di1 = distributions.FloatDistribution(low=1, high=3)
+    assert di0 != di1
+    assert not di0 == di1
```
himkt, line 505: [question] I'd like to know why it checks `not di0 == di1` in addition to `di0 != di1` (is it not equivalent?).

tests/test_distributions.py
```diff
@@ -333,6 +498,25 @@ def test_eq_ne_hash() -> None:
...
+    assert hash(di0) != hash(di1)
```
himkt, line 506: I think `di0 != di1` is sufficient to check equality of distributions because `hash()` encode a class name and parameters, which is checked in `__eq__`.

https://github.com/optuna/optuna/blob/d30263dc47f30015412dd76b8481d37a04cc70ee/optuna/distributions.py#L78-L84

https://github.com/optuna/optuna/blob/d30263dc47f30015412dd76b8481d37a04cc70ee/optuna/distributions.py#L86-L88

tests/test_distributions.py
```diff
@@ -333,6 +498,25 @@ def test_eq_ne_hash() -> None:
...
+
+    # Different distribution classes.
+    di2 = distributions.IntDistribution(low=1, high=2)
+    assert di0 != di2
+    assert not di0 == di2
+    assert hash(di0) != hash(di2)
+
+    # Different types.
+    assert di0 != 1
```
himkt, line 515: [question] Do we have to check a type here? It feels too much testing to me. 💭

optuna/distributions.py
```diff
@@ -10,6 +10,10 @@
 
 
 CategoricalChoiceType = Union[None, bool, int, float, str]
+_int_distribution_deprecated_msg = "Use :class:`~optuna.distributions.IntDistribution` instead."
```
himkt, line 13: [imo] I think variables should be introduced with references, so how about adding them in a subsequent pull request?

himkt (commented):
Thank you @nyanhi for your constant contribution, the change looks basically solid!
I left some minor comments.

optuna/distributions.py
nyanhi: Sure, just thought someone might be confused why similar classes are there but I agree that we can add it in a subsequent pull request.

tests/test_distributions.py
nyanhi: Thank you for the comments, sorry I mostly copied those from existing tests. I think those 2 statements are equivalent.

tests/test_distributions.py
nyanhi: My random guess is that these checks for existing classes might have tried to check the second condition of `__eq__()`. However, the check seems to be trivial and other classes do not have such check, so I think fine to remove it.

tests/test_distributions.py
nyanhi: Agreed, let me remove it.

tests/test_distributions.py
himkt: Sorry for pointing out something that wasn't your change. But I think it would be nice if we can cleanup tests as well.
Thank you for the quick updates as always!

nyanhi: No problem, reflected those changes to existing checks as well. Thank you.

keisuke-umezawa (approved):
LGTM! Thank you for the contribution!

himkt (approved):
LGTM. Thank you so much for the long run @nyanhi!

himkt:
@nzw0301 Please take a look at your requested changes because merging is blocked. :pray:

himkt:
Take care. 🏥

Merging.

nyanhi:
Thank you so much for all of the supports and idea! I hope nzw-san will get better soon.

Issue #2542, opened 2021-03-30
`ZeroDivisionError` when plotting parameter importance


dmar1n:
Plotting parameter importance of a multi-objective study causes a `ZeroDivisionError: Weights sum to zero, can't be normalized.` The error happens with a with two-objective study, which otherwise runs nicely and produces a Pareto plot without any problem.

## Expected behavior

Generate a plot of parameter importance based on a multi-objective study, as described in the documentation.

## Environment

- Optuna version: 2.6.0
- Python version: 3.8
- OS: Ubuntu 18.04

## Error messages, stack traces, or logs

```
ERROR Traceback (most recent call last):
  File "/opt/mt/anaconda3/envs/develop/lib/python3.8/site-packages/optuna/visualization/_param_importances.py", line 126, in plot_param_importances
    importances = optuna.importance.get_param_importances(
  File "/opt/mt/anaconda3/envs/develop/lib/python3.8/site-packages/optuna/importance/__init__.py", line 91, in get_param_importances
    return evaluator.evaluate(study, params=params, target=target)
  File "/opt/mt/anaconda3/envs/develop/lib/python3.8/site-packages/optuna/importance/_fanova/_evaluator.py", line 121, in evaluate
    evaluator.fit(
  File "/opt/mt/anaconda3/envs/develop/lib/python3.8/site-packages/optuna/importance/_fanova/_fanova.py", line 74, in fit
    self._trees = [_FanovaTree(e.tree_, search_spaces) for e in self._forest.estimators_]
  File "/opt/mt/anaconda3/envs/develop/lib/python3.8/site-packages/optuna/importance/_fanova/_fanova.py", line 74, in <listcomp>
    self._trees = [_FanovaTree(e.tree_, search_spaces) for e in self._forest.estimators_]
  File "/opt/mt/anaconda3/envs/develop/lib/python3.8/site-packages/optuna/importance/_fanova/_tree.py", line 23, in __init__
    statistics = self._precompute_statistics()
  File "/opt/mt/anaconda3/envs/develop/lib/python3.8/site-packages/optuna/importance/_fanova/_tree.py", line 185, in _precompute_statistics
    value = numpy.average(child_values, weights=child_weights)
  File "<__array_function__ internals>", line 5, in average
  File "/opt/mt/anaconda3/envs/develop/lib/python3.8/site-packages/numpy/lib/function_base.py", line 409, in average
    raise ZeroDivisionError(
[...]

PR #3074, opened 2021-11-01
[RFC] Handle constant feature vectors in `fANOVA` tree

xadrianzetx:
## Motivation

This PR attempts to fix underlying issue that caused `ZeroDivisionError` in #2542. 

The error occurs in `FanovaImportanceEvaluator` when one of hyperparameters has distribution set up such that `low = high`.  This setup causes suggestion for this hyperparameter be constant across all trials (this is acceptable in Optuna) essentially creating a constant feature vector for `fANOVA` tree to fit on. Since it's just a regular decision tree, it will never split a constant feature vector. This causes range of problems when calculating importances (e.g. empty arrays, strictly positive values equal to zero) that eventually accumulate to originally reported error.

## Description of the changes

* [x] Fix cardinality calculation in `fANOVA` tree (same as [original implementation](https://github.com/automl/random_forest_run/blob/cc2354ed2622c270d96b88a068ae87990e1da054/include/rfr/util.hpp#L51-L62)) to avoid strictly positive weights go to 0
* [x]  Improve numerical stability when calculating marginal variance
* [ ] Write tests

### DIFF
--- a/optuna/importance/_fanova/_tree.py
+++ b/optuna/importance/_fanova/_tree.py
@@ -71,7 +71,7 @@ class _FanovaTree(object):
             sample[features] = numpy.array(midpoints)
 
             value, weight = self._get_marginalized_statistics(sample)
-            weight *= float(numpy.prod(sizes))
+            weight *= float(numpy.prod(sizes)) + 1e-100  # Epsilon for numerical stability.
 
             values = numpy.append(values, value)
             weights = numpy.append(weights, weight)
@@ -299,7 +299,8 @@ class _FanovaTree(object):
 
 
 def _get_cardinality(search_spaces: numpy.ndarray) -> float:
-    return numpy.prod(search_spaces[:, 1] - search_spaces[:, 0])
+    c = numpy.prod(search_spaces[:, 1] - search_spaces[:, 0])
+    return c if c > 0 else numpy.prod(search_spaces.shape)
 
 
 def _get_subspaces(


xadrianzetx:
I've marked this as RFC since this solves edge case in which one of distributions has `low = high`. In this case importance for this hyperparameter calculated by `fANOVA` should always be 0, meaning equally viable solution that avoids `ZeroDivisionError` is just to set importance for this param to 0 (instead of calculating it). Would love to hear different opinions, esp @hvy. If this PR is a miss I'm ok to close it.

optuna/importance/_fanova/_tree.py
```diff
@@ -299,7 +299,8 @@ def _get_node_children_subspaces(
 
 
 def _get_cardinality(search_spaces: numpy.ndarray) -> float:
-    return numpy.prod(search_spaces[:, 1] - search_spaces[:, 0])
+    c = numpy.prod(search_spaces[:, 1] - search_spaces[:, 0])
+    return c if c > 0 else numpy.prod(search_spaces.shape)
```
hvy, line 303: Could you explain this logic taking the product of the shape?

hvy (commented):
Thanks a lot for the interesting fix. Fixing numerical instabilities sounds fair. Do you happen to know if this is how the original implementation addresses these edge cases as well? As you also mention, I also think that another reasonable approach is to in `_Fanova` assert that weights will be strictly positive, omitting the distributions where `low==high` (this can be checked using `BaseDistribution.single()`) already in `FanovaImportanceEvaluator.evaluate`. This'll keep the internals of `_Fanova` simple and well always get exactly 0 as the importance (and not a value close to it), although perhaps not the cleanest.

xadrianzetx:
> Do you happen to know if this is how the original implementation addresses these edge cases as well?

They set distribution bounds to [1, np.nan]
https://github.com/automl/fanova/blob/7b58914b1c38ee60a173b260fc4bb269aaf8acb1/fanova/fanova.py#L138-L140
Which then is handled as special case through the code, but generally ends up assigning exactly 0 importance to a parameter at particular tree. Of course, all trees have the same 0 importance for this param, so variance is also 0.
https://github.com/automl/fanova/blob/7b58914b1c38ee60a173b260fc4bb269aaf8acb1/fanova/fanova.py#L192-L201
We can't do exactly that, since there are small differences in how midpoints and sizes are calculated (only important in those special cases tbh).

optuna/importance/_fanova/_tree.py
xadrianzetx: It's a port of original implementation, which handles both categorical features and constants in this code path.
https://github.com/automl/random_forest_run/blob/cc2354ed2622c270d96b88a068ae87990e1da054/include/rfr/util.hpp#L49-L62
I think in case of constants, it's just important that this value is strictly positive (it causes `ZeroDivisionError` down the line). Didn't investigate code paths for categorical features.

hvy:
Great, thanks for elaborating and the pointers to the original implementation.

> On the second thought, it actually might be closer to original implementation if we pop single() distribution somewhere here

I have to say, I'm +1 for this approach. Code complexity will go up in either case and I personally find this solution to be easier to maintain, given that `single()` is a well-defined method already used throughout optuna, e.g. https://github.com/optuna/optuna/blob/v2.10.0/optuna/trial/_trial.py#L713-L714, to alleviate the burden (partially to be exact) of samplers to not have to handle these (low==high) cases.

optuna/importance/_fanova/_tree.py
hvy: I see, thanks! So the original implementation includes a code branch to for e.g. categorical paramters return the number of choices as the cardinality. In optuna, we're actually one-hot encoding all categorical parameters (e.g. ["foo", "bar", "baz"] gets converted into three numerical parameters internally with range (0, 1)). I'm not sure but do you think this code has the same effect? (In any case, it might be less relevant if we decide to go with your "second thought" approach)

xadrianzetx:
Yeah you're right, let's go with the alternative. Let me close this one and reimplement the fix.

PR #3077, opened 2021-11-03
Add upper version constraint of PyTorchLightning

nzw0301:
## Motivation

Pytorch-Lightning 1.5.0 has been released, but this release contains backwards-incompatible changes as in https://github.com/PyTorchLightning/pytorch-lightning/releases/tag/1.5.0#bc-changes. Due to this change, integration CI is falling now.

## Description of the changes

Add the version constraint of PyTorchLightning not to use `1.5.0` or higher version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,9 @@ def get_extras_require() -> Dict[str, List[str]]:
             "tensorflow<2.7.0",
             "tensorflow-datasets",
             "pytorch-ignite",
-            "pytorch-lightning>=1.0.2",
+            # TODO(nzw0301): remove the upper version constraint when the callback supports
+            # pytorch-lightning==1.5.0.
+            "pytorch-lightning>=1.0.2,<1.5.0",
             "skorch",
             "catalyst>=21.3",
             "torch==1.8.0 ; sys_platform=='darwin'",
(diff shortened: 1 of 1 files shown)


himkt (approved):
Thank you for the patch. Looks good as the first aid.

I confirmed `AttributeError: 'Trainer' object has no attribute 'accelerator_connector'` is raised in integration tests.

https://github.com/optuna/optuna/blob/d7e09dc9eb5bb6cc9e7ad83804d47254d7f586ac/optuna/integration/pytorch_lightning.py#L59

> backwards-incompatible changes as in https://github.com/PyTorchLightning/pytorch-lightning/releases/tag/1.5.0#bc-changes

It looks `Trainer.accelerator_connector` is renamed to `Trainer._accelerator_connector`. Is it possible to support the latest pytorch-lightning by switching reference to the connector attribute based on the version of lightning?

https://github.com/PyTorchLightning/pytorch-lightning/compare/1.4.9...1.5.0#diff-667a2513c158c2b01d73b479e4dea75587d8531f9cd286891d34570a2fd145dcR431

nzw0301:
Thank you for your comment and for sharing your investigation! 

> It looks `Trainer.accelerator_connector` is renamed to `Trainer._accelerator_connector`.

It sounds much simpler than expected change! Brilliant!

> Is it possible to support the latest PyTorch-lightning by switching reference to the connector attribute based on the version of lightning?

I suppose so. But I believe that we do not have enough resources to support multiple pytorch-lightning versions and follow frequent drastically break-backwards compatibility by this library. So I propose

- pin pytorch-lightning version in `setup.py` by default 
- pinned version is ideally the latest PyTorch-lightning

toshihikoyanase (approved):
LGTM. Thank you for your quick fix!

> It looks `Trainer.accelerator_connector` is renamed to `Trainer._accelerator_connector`. Is it possible to support the latest pytorch-lightning by switching reference to the connector attribute based on the version of lightning?

To support the multiple-version of `pytorch-lightning`, I think we need further discussion on the maintenance cost, testability, and etc. Also, we have another choice to pin the pytorch-lightning version by default as @nzw0301 mentioned. 

How about merging this PR and discussing it in a new PR? @himkt

toshihikoyanase:
Talked with @himkt offline. Let me merge this PR.

PR #3059, opened 2021-10-31
Pin `tensorflow` and `tensorflow-estimator` versions to `<2.7.0`.

toshihikoyanase:
## Motivation

As can be seen in the [CI log](https://github.com/optuna/optuna/runs/4058023388?check_suite_focus=true#step:7:19), `tests/integration_tests/test_tensorflow.py` failed due to the import error.

```
ImportError while importing test module '/home/runner/work/optuna/optuna/tests/integration_tests/test_tensorflow.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/integration_tests/test_tensorflow.py:10: in <module>
    from optuna.integration import TensorFlowPruningHook
E   ImportError: cannot import name 'TensorFlowPruningHook' from 'optuna.integration' (unknown location)
```

This may be related to the [release](https://github.com/tensorflow/estimator/releases/tag/v2.7.0) of `tensorflow-estimator==2.7.0`. I think It requires `tensorflow==2.7.0`, but it has not been released yet. These inconsistent versions seem to be the cause of the error.

CC @hvy @himkt 

## Description of the changes
This PR adds the version restriction to `tensorflow` and `tensorflow-estimator`.

PR #3078, opened 2021-11-04
Pin `keras` version to `<2.7.0`.

toshihikoyanase:
## Motivation

[CI jobs](https://github.com/optuna/optuna/runs/4100003344?check_suite_focus=true) failed due to the import error of `keras`. `keras==2.7.0` was released today, but the corresponding version of `tensorflow` has not been released yet. I guess the version inconsistency caused the issue similarly to #3059.

### Reproducible code

```
% docker run -it python:3.7 bash
# pip install 'tensorflow<2.7.0' 'tensorflow-estimator<2.7.0' 'tensorflow-datasets'
# pip freeze | grep keras
keras==2.7.0
# python
>>> import keras
  ...
  File "/usr/local/lib/python3.7/site-packages/tensorflow/python/eager/monitoring.py", line 135, in __init__
    self._metric = self._metric_methods[self._label_length].create(*args)
tensorflow.python.framework.errors_impl.AlreadyExistsError: Another metric with the same name already exists.
```

If I installed the release candidate of `tensorflow`, the keras can be imported successfully. So, I think we can simply remove the version constraint after `tensorflow==2.7.0` is released.

```
% docker run -it python:3.7 bash
# pip install --pre tensorflow
# python
>>> import keras
>>>
```

## Description of the changes

- Add version constraint of `keras`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,6 +99,7 @@ def get_extras_require() -> Dict[str, List[str]]:
             "scikit-optimize",
             "xgboost",
             # TODO(toshihikoyanase): Remove the constraints when tensorflow==2.7.0 is released.
+            "keras<2.7.0",
             "tensorflow-estimator<2.7.0",
             "tensorflow<2.7.0",
             "tensorflow-datasets",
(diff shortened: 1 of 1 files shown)


hvy (approved):
LGTM!

not522 (approved):
LGTM

PR #3080, opened 2021-11-04
Fix `parametrize_sampler`

not522:
## Motivation
Many samplers are accidentally excluded from `parametrize_sampler`.

## Description of the changes
Fix `parametrize_sampler`

### DIFF
--- a/tests/samplers_tests/test_samplers.py
+++ b/tests/samplers_tests/test_samplers.py
@@ -46,9 +46,11 @@ parametrize_sampler = pytest.mark.parametrize(
         lambda: optuna.integration.PyCmaSampler(n_startup_trials=0),
         optuna.samplers.NSGAIISampler,
     ]
-    + []
-    if sys.version_info < (3, 7, 0)
-    else [lambda: optuna.integration.BoTorchSampler(n_startup_trials=0)],
+    + (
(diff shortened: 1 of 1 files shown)


toshihikoyanase:
@not522 Good catch! I'm sorry but I happened to introduce this bug on the test 😅 in #2928. Let me review this PR.

toshihikoyanase:
Let me remove the `bug` label since it is a bug in the unit tests, not a bug in Optuna itself.

toshihikoyanase (approved):
LGTM! Thank you for your fix.

I understand that `[1] + [2] if ... else [3]` is equivalent to `([1] + [2]) if ... else [3]` as follows:
```pyhon
>>> [1] + [2] if True else [3]
[1, 2]
>>> [1] + [2] if False else [3]
[3]
```

```python
>>> [1] + ([2] if True else [3])
[1, 2]
>>> [1] + ([2] if False else [3])
[1, 3]
```
So, the current master tests only the `BoTorchSampler` if Python version is >=3.7.0.

The number of test cases is as follows:
- `master`: 216
- this branch: 694

himkt (approved):
It's difficult to notice. 😅 Really nice catch @not522!

Issue #3079, opened 2021-11-04
Suppress log messages of `tests/integration_tests/lightgbm_tuner_tests/test_optimize.py`


toshihikoyanase:
## Motivation

With `lightgbm==3.3.1`, the tests for `LightGBMTuner/LightGBMTunerCV` show more than 500,000 lines of the log messages as can be seen in https://github.com/optuna/optuna/runs/4090215476?check_suite_focus=true. Such a large number of messages prevents developers from troubleshooting. With the previous version (3.2.1), they showed about 2,000 lines of messages, and the log messages increased by 250 times.

I think we need to concentrate on informative messages and suppress the others.

### Logs of intermediate values

We can see the intermediate results in the log message as follows:

```console
  0%|          | 0/20 [00:00<?, ?it/s]
......[1]	valid_0's l2: 0
[2]	valid_0's l2: 0
[3]	valid_0's l2: 0
[4]	valid_0's l2: 0
[5]	valid_0's l2: 0
[6]	valid_0's l2: 0
[7]	valid_0's l2: 0
[8]	valid_0's l2: 0
[9]	valid_0's l2: 0
[10]	valid_0's l2: 0
[11]	valid_0's l2: 0
[12]	valid_0's l2: 0
[13]	valid_0's l2: 0
[14]	valid_0's l2: 0
```

I guess this is related to `verbose_eval` ~~and maybe we need to set `verbase_eval=False` to `LightGBMTuner.__init__` and `LightGBMTunerCV.__init__`~~.

Please note that `verbose_eval` was deprecated as mentioned in https://github.com/optuna/optuna/issues/3013. So, we might use the callbacks instead.

### Warnings from the `lightgbm` library

The lightgbm library shows warnings as follows:
```console
Warning: m...[LightGBM] [Warning] Auto-choosing row-wise multi-threading, the overhead of testing was 0.000017 seconds.
You can set `force_row_wise=true` to remove the overhead.
And if memory is not enough, you can set `force_col_wise=true`.
[LightGBM] [Info] Total Bins 25
[LightGBM] [Info] Number of data points in the train set: 10, number of used features: 5
[LightGBM] [Info] Start training from score 0.700000
Warning: ] [Warning] No further splits with positive gain, best gain: -inf
Warning: ] [Warning] Stopped training because there are no more leaves that meet the split requirements
```

I think this is mainly because a small configuration in terms of datasets and budget is used in the test cases, so I guess we can simply ignore such warnings.

PR #3049, opened 2021-10-27
Adapted early_stopping to callback

joseporiolayats:
## Motivation

To solve issue #3036 and start to commit to opensource projects in hacktoberfest!
## Description of the changes

I changed the lines containing the old call "early_stopping_rounds" into "callbacks=[early_stopping(val)]". I also looked if some other parameters affected by this callback rewrite from lightgbm were involved, but no one did. Hope this helps!

Issue #3036, opened 2021-10-25
Replace deprecated arguments with callback-ish way in lightgbm test


nzw0301:
## Motivation

Since lightgbm v3.3.0, train/cv's arguments have been deprecated as described in https://github.com/optuna/optuna/issues/3013. It would be great to replace such deprecated arguments with recommended way by lightgbm.

## Description

Replace the deprecated arguments with callback-ish way. The target lines are given in https://github.com/optuna/optuna/pull/3014#issuecomment-949362020.

See also https://github.com/optuna/optuna-examples/pull/64/files for the concrete replacement.

## Additional context (optional)

We do not need to care about the older lightgbm version.

PR #3014, opened 2021-10-17
Make default value of verbose_eval None for LightGBMTuner/LightGBMTunerCV to avoid conflict

chezou:
## Motivation

Close #3013

Since LightGBM 3.3.0, the following arguments are deprecated to use
callbacks instead:
- `verbose_eval`
- `early_stopping_rounds`
- `learning_rates`
- `eval_result`

However, the default value of `verbose_eval` for `LightGBMTuner` and
`LightGBMTunerCV` is `True` so that we need to write extra
`verbose_eval=None` to avoid blocking to use `log_evaluation` callback.

This patch avoids writing redundant `verbose_eval` arguments for
`log_evaluation` callback.

## Description of the changes

Make default value of `verbose_eval` as `None` for LightGBMTuner and LightGBMTunerCV

PR #3086, opened 2021-11-07
Fix verbosity for `tests/integration_tests/lightgbm_tuner_tests/test_optimize.py`

nyanhi:
## Motivation
To address the issue #3079. 

## Description of the changes
Specified `verbose_eval = False` to suppress the log of intermediate values.
Specified `verbose = -1` for `test_tune_num_leaves_negative_max_depth()` to suppress warning and info like below from LightGBM library:

```
Warning: m...[LightGBM] [Warning] Auto-choosing row-wise multi-threading, the overhead of testing was 0.000017 seconds.
You can set `force_row_wise=true` to remove the overhead.
And if memory is not enough, you can set `force_col_wise=true`.
[LightGBM] [Info] Total Bins 25
[LightGBM] [Info] Number of data points in the train set: 10, number of used features: 5
[LightGBM] [Info] Start training from score 0.700000
Warning: ] [Warning] No further splits with positive gain, best gain: -inf
Warning: ] [Warning] Stopped training because there are no more leaves that meet the split requirements
```

### DIFF
--- a/tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
+++ b/tests/integration_tests/lightgbm_tuner_tests/test_optimize.py
@@ -449,10 +449,7 @@ class TestLightGBMTuner(object):
 
     def test_tune_num_leaves_negative_max_depth(self) -> None:
 
-        params: Dict[str, Any] = {
-            "metric": "binary_logloss",
-            "max_depth": -1,
-        }
+        params: Dict[str, Any] = {"metric": "binary_logloss", "max_depth": -1, "verbose": -1}
         X_trn = np.random.uniform(10, size=(10, 5))
         y_trn = np.random.randint(2, size=10)
         train_dataset = lgb.Dataset(X_trn, label=y_trn)
@@ -464,6 +461,7 @@ class TestLightGBMTuner(object):
             num_boost_round=3,
             early_stopping_rounds=2,
             valid_sets=valid_dataset,
+            verbose_eval=False,
         )
         runner.tune_num_leaves()
         assert len(runner.study.trials) == 20
@@ -550,7 +548,7 @@ class TestLightGBMTuner(object):
         dataset = lgb.Dataset(np.zeros((10, 10)))
 
         study = optuna.create_study()
-        tuner = LightGBMTuner(params, dataset, valid_sets=dataset, study=study)
+        tuner = LightGBMTuner(params, dataset, valid_sets=dataset, study=study, verbose_eval=False)
 
         with mock.patch.object(_BaseTuner, "_get_booster_best_score", return_value=1.0):
             tuner.tune_regularization_factors()
@@ -591,6 +589,7 @@ class TestLightGBMTuner(object):
                 valid_sets=dataset,
                 study=study,
                 verbosity=verbosity,
+                verbose_eval=False,
                 time_budget=1,
             )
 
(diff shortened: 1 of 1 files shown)


nzw0301:
Thank you for your contribution! I'm worried about using `verbose_eval=False` because `verbose_eval` has been deprecated since lightgbm 3.3.0 and Optuna plans to use callback-ish way in its test as described in https://github.com/optuna/optuna/issues/3036 that will be addressed by this ongoing PR, https://github.com/optuna/optuna/pull/3049. So could you use the callback instead of `verbose_eval=False`?

nyanhi:
Thank you for the suggestion, that makes sense! Sorry I just realized the sequence of discussion.
I will try to replace it with the `log_evaluation()` callback.

toshihikoyanase:
@nzw0301 Thank you for your comment. I should have mentioned the deprecation of `verbose_eval`. Could you continue your review, please?

nzw0301:
Sorry, I could not install lightgbm >= 3.3.0 via conda due to https://github.com/microsoft/LightGBM/issues/4736 to locally check the tests. `pip install` never works on my environment (M1 mac), so let me assign myself once :(

toshihikoyanase:
@nzw0301 Thank you for letting me know. I got it.
@himkt Could you review this PR? If you don't have the test environment (due to https://github.com/microsoft/LightGBM/issues/4736), please feel free to unassign yourself.

himkt (commented):
Thank you for the patch! I confirmed that a tons of logs observed in https://github.com/optuna/optuna/issues/3079 disappeared.

I have one question: don't we need to add a version constraint for lightgbm (>=3.3.0)?
(`log_evaluation` seems to have been introduced in [v3.3.0](https://github.com/microsoft/LightGBM/releases/tag/v3.3.0), right?)

nyanhi:
Thank you. Per my understanding, #3036 and https://github.com/optuna/optuna/pull/3014#issuecomment-949480109 mentioned that only the latest version of LightGBM is supported in this module for the time being. (Please chime in if my understanding is not correct)

himkt:
Thank you for referring the discussion. I now understand we only support the latest version of LightGBM well. I was wondering we should update our version restriction like in the case of scikit-learn.
https://github.com/optuna/optuna/blob/d30263dc47f30015412dd76b8481d37a04cc70ee/setup.py#L142

However, Optuna doesn't always specify a minimum requirements. 💭  I'm wondering it would be safer to add it because LightGBM 3.2.x is relatively new and users may use it. @toshihikoyanase What do you think about it?

toshihikoyanase:
@nyanhi Thank you for updating the test code. It greatly reduced the log messages.

@himkt I'm sorry for the delayed response.

> I'm wondering it would be safer to add it because LightGBM 3.2.x is relatively new and users may use it.

This PR only changed the test code, and the users who utilize `lightgbm<3.3` can still use the `LightGBMTuner` and `LightGBMTunerCV`. So, I think we may not need to add the version restriction to `setup.py`.

himkt (approved):
Thank you @toshihikoyanase for clarifying the dependency policy. So, looks good to me!

toshihikoyanase (approved):
LGTM! This must improve the CI log. Thank you for your contribution.

@himkt Thank you for your kind ping.

Issue #2939, opened 2021-09-19
Make suggest API more consistent and improve documentation


himkt:
## Motivation

We want to make behaviors of suggesting API for integer parameters and floating-point parameters consistent.
Our focuses are summarized in the description.

### 1. Suggest APIs variant for floating-point

There are four methods available for the current suggest API for a floating-point parameter: `suggest_float`, `suggest_uniform`, `suggest_loguniform`, and `suggest_discrete_uniform`. On the other hand, there is one method for each integer and categorical parameter: `suggest_int` and `suggest_categorical`. We plan to unify methods for floating point parameter into `suggest_float` and deprecate `suggest_uniform`, `suggest_loguniform`, and `suggest_discrete_uniform`. Then, we can change the implementation of `suggest_uniform`, `suggest_loguniform`, and `suggest_discrete_uniform` so that they are wrapped by `suggest_float`.

### 2. Documentation about behavior about the upper endpoint of floating-point distribution

While suggest API for an integer parameter would return the upper endpoint of the parameter space, suggest API for a float parameter would **not** return the upper endpoint according to documentation. Although it is true for some situations (e.g. `TPESampler`), samplers could return an upper endpoint due to numerical error of a floating-point. We decided to update our documentation that an upper endpoint would be included as a result of suggest API for float.

### 3. Allowing `suggest_float` to add a parameter during optimization (to make consistent with `suggest_int`)

In the current implementation, behavior is different between integer and float when adding a parameter (e.g. `step`) to suggest API during an optimization process. An error occurs for float while it works in the integer. To make the behavior consistent, we plan to change `suggest_float` so that it works when adding an additional parameter during optimization.

xadrianzetx:
Regarding the first point - if I understand correctly, TODOs here are:
*  Make `suggest_uniform`, `suggest_loguniform,` and  `suggest_discrete_uniform` private.
*  Leave public APIs for those, but `@deprecate` them, and make them call `suggest_float`
*  Wrap private `suggest_uniform`, `suggest_loguniform,` and  `suggest_discrete_uniform` in `suggest_float`
*  Find and replace all occurrences of `suggest_uniform`, `suggest_loguniform,` and  `suggest_discrete_uniform` in the repo and replace with `suggest_float` (although this can be done separately, since `@deprecate` is used.

Please let me know if this sounds good, @himkt and I will open PR.

himkt:
@xadrianzetx 
Thank you so much for paying your attention to this issue!

[...]

PR #3091, opened 2021-11-08
Fix range description in `suggest_float` docstring

xadrianzetx:
## Motivation

Previously documentation stated that `high` distribution bound is not included in the range from which floating point value is suggested. This PR address that as per second point described in #2939.

## Description of the changes

* Fix docstring in `suggest_float` and methods calling `suggest_float`.

### DIFF
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -127,12 +127,7 @@ class Trial(BaseTrial):
             low:
                 Lower endpoint of the range of suggested values. ``low`` is included in the range.
             high:
-                Upper endpoint of the range of suggested values. ``high`` is excluded from the
-                range.
-
-                .. note::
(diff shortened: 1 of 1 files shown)


xadrianzetx:
The integration test fail is caused by one of Weights & Biases dependencies release - https://pypi.org/project/shortuuid/1.0.5/

toshihikoyanase:
@xadrianzetx Ah, you fixed `shortuuid`. Great!
The patched version seems to be released, so I re-runed the CI jobs. Thank you!

@himkt Could you review this PR since you are the author of #2939?

@HideakiImamura I guess this PR can be merged with a single approval, but if you have any concerns about the description to include the upper endpoints, please feel free to add comments.

himkt (approved):
Thank you for the contribution. This PR makes `suggest_float` consistent with `optuna.distributions`. LGTM!

PR #2810, opened 2021-07-15
Add Plotly-like interpolation algorithm to `optuna.visualization.matplotlib.plot_contour`

xadrianzetx:
This PR introduces new interpolation algorithm to be used when creating contour plots with `optuna.visualization.matplotlib.plot_contour()`, improving on linear interpolation offered by `scipy.griddata()`. 

The algorithm is a port of [Plotly.js](https://github.com/plotly/plotly.js/blob/master/src/traces/heatmap/interp2d.js#L30) version.

This addresses known issues in current implementation:

* contour lines which are not smooth
* blank areas on edges of the plot

Additionally, tweaks to `xrange` and `yrange`  are introduced in order to correctly visualize trial values at the edges of param grid.

## Motivation
This PR resolves #2738 and addresses [#2712 (review)](https://github.com/optuna/optuna/pull/2712#pullrequestreview-679629017).

## Description of the changes
- [x] Implement interpolation algorithm and helpers.
- [x] Switch to new algorithm in `optuna.visualization.matplotlib.plot_contour()`
- [x] Address `xrange` and `yrange` padding
- [x] Write tests

## Examples
All plots were generated using example from #2595 

Current algorithm
image

New algorithm
image

Plotly
image

And a very cool gif showing how this algorithm converges :^)
plot_iter2

Issue #2595, opened 2021-04-16
The objective range of Matplotlib contour plot has a different numerical range from Plotly backend


nzw0301:
## Expected behavior

The range of objective  in `optuna.visualization.matplotlib.plot_contour` should have a similar range to `optuna.visualization.plot_contour`.

## Environment

- Optuna version: 2.7.0
- Python version: 3.8.8
- OS: Mac
- (Optional) Other libraries and their versions:
    - matplotlib: 3.3.4
    - Plotly: 4.11.0

## Error messages, stack traces, or logs
The matplotlib range is from `-60 000` to `360 000` as in the following figure.

![download (2)](https://user-images.githubusercontent.com/7121753/114876635-79e04f00-9e39-11eb-97c6-3b7b16e89456.png)

On the other hand, Plotly's range is from `0` to `9 000`. That is supposed to be true.

![newplot (2)](https://user-images.githubusercontent.com/7121753/114876650-7baa1280-9e39-11eb-9c3e-554d16ed2496.png)

## Steps to reproduce

1. Please run a following example code blocks

## Reproducible examples (optional)

```python
import optuna


def objective(trial):
    x = trial.suggest_uniform('x', -100, 100)
    y = trial.suggest_categorical('y', list(range(-10, 10)))
    z = trial.suggest_uniform('z', -100, 100)
    return x ** 2 + y - z

study = optuna.create_study(sampler=optuna.samplers.TPESampler(seed=7))
study.optimize(objective, n_trials=200)

optuna.visualization.matplotlib.plot_contour(study, params=['y', 'z'])
optuna.visualization.plot_contour(study, params=['y', 'z']) # plotly
```

nzw0301:
I think the incorrect color bar comes from the interpolation of `z` values:
https://github.com/optuna/optuna/blob/ef0a7977a78a2ba7de20b652ce3c954828d461d1/optuna/visualization/matplotlib/_contour.py#L320-L325

If we set another interpolation method such as `nearest`, the colormap looks more accurate:
![download (5)](https://user-images.githubusercontent.com/7121753/115153212-b6b57b80-a0af-11eb-81d6-353788783a54.png)

nzw0301:
I would like to point out the different color scale between plotly and matplotlib:

- plotly: white (smaller objective) -- dark-blue (higher objective)
- matplotlib: white (higher objective) -- dark blue (smaller objective)

01-vyom:
I would like to work on this issue.

[...]

PR #3092, opened 2021-11-10
Speed up Matplotlib backend `plot_contour` using SciPy's `spsolve`

not522:
## Motivation
#2810 introduced Plotly's interpolation algorithm. But it needs a longer computation time. This PR will speed up `plot_contour` using Scipy's `spsolve`.

The output figure slightly changes because of convergence errors of algorithms.

## Description of the changes
- Speed up Matplotlib backend `plot_contour`
- Remove warning

## Examples

All plots were generated using example from #2595

- PR (0.234 sec)

PR

- master (2.988 sec)

master

- Plotly backend

plotly

### DIFF
--- a/optuna/visualization/matplotlib/_contour.py
+++ b/optuna/visualization/matplotlib/_contour.py
@@ -50,6 +50,11 @@ def plot_contour(
     .. seealso::
         Please refer to :func:`optuna.visualization.plot_contour` for an example.
 
+    Warnings:
+        Output figures of this Matplotlib-based
+        :func:`~optuna.visualization.matplotlib.plot_contour` function would be different from
+        those of the Plotly-based :func:`~optuna.visualization.plot_contour`.
+
     Example:
 
         The following code snippet shows how to plot the parameter relationship as contour plot.
@@ -96,6 +101,10 @@ def plot_contour(
 
     _imports.check()
     _check_plot_args(study, target, target_name)
+    _logger.warning(
+        "Output figures of this Matplotlib-based `plot_contour` function would be different from "
+        "those of the Plotly-based `plot_contour`."
+    )
     return _get_contour_plot(study, params, target, target_name)
 
 
--- a/optuna/visualization/matplotlib/_contour.py
+++ b/optuna/visualization/matplotlib/_contour.py
@@ -1,6 +1,5 @@
 from collections import defaultdict
 from typing import Callable
-from typing import cast
 from typing import DefaultDict
 from typing import Dict
 from typing import List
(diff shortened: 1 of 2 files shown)


toshihikoyanase:
@xadrianzetx Could you add your review comments as the author of #2810?

@HideakiImamura @himkt Could you review this PR?

optuna/visualization/matplotlib/_contour.py
```diff
@@ -51,11 +50,6 @@ def plot_contour(
     .. seealso::
         Please refer to :func:`optuna.visualization.plot_contour` for an example.
 
-    Warnings:
```
himkt, line 54: I think it would be user-friendly to leave this documentation because there is still difference between the plotly backend and this implementation. What do you think?

himkt (commented):
I'm amazed by the impressive speedup! I'm happy to see the series of improvement.
I put two superficial comments.

xadrianzetx (commented):
Basically, this implementation looks great to me. Setting up linear system, and then solving it with highly optimized solver is much better approach than current master! Let me leave a minor comment on some cosmetics.

optuna/visualization/matplotlib/_contour.py
not522: I think they are similar, and no need to show the warning. But it is not the main task of this PR. Let me revert the removal.

optuna/visualization/matplotlib/_contour.py
```diff
@@ -489,135 +476,34 @@ def _interpolate_zmap(zmap: Dict[complex, float], contour_plot_num: int) -> None
...
+                for dx, dy in ((-1, 0), (1, 0), (0, -1), (0, 1)):
+                    if 0 <= x + dx < contour_plot_num and 0 <= y + dy < contour_plot_num:
+                        a_data.append(1)
+                        a_row.append(grid_index)
+                        a_col.append(grid_index)
+                        a_data.append(-1)
+                        a_row.append(grid_index)
+                        a_col.append(grid_index + dy * contour_plot_num + dx)
+
+    z = scipy.sparse.linalg.spsolve(scipy.sparse.csc_matrix((a_data, (a_row, a_col))), b)
```
HideakiImamura, line 507: Is there any reason to use the CSC matrix instead of the CSR matrix?

HideakiImamura (commented):
Thanks for the PR. This is amazing improvement. Excellent!  I have a minor comment. Please take a look.

optuna/visualization/matplotlib/_contour.py
not522: I'm not sure which format is better to use. This PR follows `spsolve`'s example and implementation.
https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.spsolve.html
https://github.com/scipy/scipy/blob/v1.7.1/scipy/sparse/linalg/dsolve/linsolve.py#L139-L145

himkt: My simple benchmark shows a selection of data format doesn't have a significant impact.
code: https://colab.research.google.com/gist/himkt/947d7f2bb25d54dac8686fef42eb6521/naive-benchmark-csc-and-csr.ipynb
patch: https://github.com/not522/optuna/compare/fast-contour...himkt:oneshot/fast-contour?expand=1

himkt (approved):
LGTM!

HideakiImamura (approved):
Thanks for the benchmark @himkt. Let me approve this PR since my question was resolved.

not522:
Thank you for your reviews!
And thank you again, @xadrianzetx! Your great implementation and description were very helpful in writing the PR.

PR #3095, opened 2021-11-14
Set concurrency to github workflows

keisuke-umezawa:
## Motivation

If we have a new commit to a certain PR, we want to stop github actions for the previous commit in the PR.

## Description of the changes

Add a concurrency setting to the github actions which will run on the PR commits. ref: https://github.community/t/concurrecy-not-work-for-push/183068

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -8,11 +8,9 @@ on:
   schedule:
     - cron: '0 23 * * SUN-THU'
 
-
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
-
 
 jobs:
   checks:
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -8,6 +8,12 @@ on:
   schedule:
     - cron: '0 23 * * SUN-THU'
 
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+
 jobs:
   checks:
     runs-on: ubuntu-latest
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -9,7 +9,7 @@ on:
     - cron: '0 23 * * SUN-THU'
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -7,7 +7,7 @@ on:
   pull_request: {}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -6,6 +6,10 @@ on:
       - master
   pull_request: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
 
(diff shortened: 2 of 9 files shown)


keisuke-umezawa:
You can check that it runs correctly with #fc7ad59.

.github/workflows/checks.yml
```diff
@@ -8,6 +8,10 @@ on:
   schedule:
     - cron: '0 23 * * SUN-THU'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
keisuke-umezawa, line 12: It limits one concurrency for each PR and master branch.

keisuke-umezawa: If we want to limit one concurrency for each PR, but multiple concurrencies to master. We can use the following:

```
${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
```

HideakiImamura:

.github/workflows/checks.yml
HideakiImamura: I prefer the last one: one concurrency for each PR but multiple ones to master. This is because it is easier to identify the cause of CI failures when PRs are merged continuously in a short period of time.

keisuke-umezawa:
@HideakiImamura 
You can check that tests were killed if we have another push in the PR. 961b127

Also, with my fork, I confirmed that multiple tests work on master. (Sorry, the timing to take the screenshot was delayed for a moment, and one became done, but it was working.
スクリーンショット 2021-11-18 23 37 14
)

HideakiImamura (approved):
LGTM!

HideakiImamura:
@toshihikoyanase Please merge this PR after your review.

.github/workflows/checks.yml
toshihikoyanase: > It limits one concurrency for each PR and master branch.

Let me check my understanding. Since the `group` includes the [`github.workflow` context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context), this `concurrency` rule limits one concurrency *for each workflow*, right? That means we run at most 13 concurrent jobs per PR because we have 13 different workflows.

.github/workflows/checks.yml
```diff
@@ -8,6 +8,10 @@ on:
   schedule:
     - cron: '0 23 * * SUN-THU'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
```
toshihikoyanase, line 12: If [this comment](https://github.com/optuna/optuna/pull/3095#discussion_r756781164) is true, I don't think the concurrency rule is necessary because this workflow has only a single job.

.github/workflows/coverage.yml
```diff
@@ -6,6 +6,10 @@ on:
       - master
   pull_request: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
+  cancel-in-progress: true
+
```
toshihikoyanase, line 12: ditto.

toshihikoyanase (commented):
Sorry for the delayed response. Let me confirm my understanding of the number of concurrent jobs per PR as mentioned in https://github.com/optuna/optuna/pull/3095#discussion_r756781164.

.github/workflows/checks.yml
keisuke-umezawa: > Since the group includes the github.workflow context, this concurrency rule limits one concurrency for each workflow, right? That means we run at most 13 concurrent jobs per PR because we have 13 different workflows.

Yes. In my understanding, that will be what we want to do.

.github/workflows/checks.yml
keisuke-umezawa: Here, we want to limit the number of running jobs for each worin the PR as `1`. 

For example, in this PR case, the `group` would be something like `group: {{checks}}-{{refs/heads/feature/set-concurrency-to-ga}}`. And if we have two workflows which have the same name of group, the former will be killed.

The situation will happen if we push two commits on this PR in a short term. And, the check test for the first commit will be killed.

toshihikoyanase: > The situation will happen if we push two commits on this PR in a short term.

Ah, that makes sense. Then, we must include this concurrency setting here. Thank you for your explanation.

toshihikoyanase (approved):
LGTM! This change improves fairness for developers to submit many PR at once. Thank you.

Issue #3094, opened 2021-11-11
Does optuna support snowflake storage?


bholland:
Does optuna support snowflake for a connection? I get a lot of errors like "snowflake does not support indexes" when I attempt to create a study using a snowflake url. Do you have plans in the future to support it?

## Information for reproducibility (optional)

where the OptunaStorageUrl is a sqlalchemy URL object
study = optuna.create_study(storage=str(OptunaStorageUrl),
                                        study_name=f"{StudyName}",
                                        direction="minimize")

### Environment

<!--
You can get this information by typing the following:
```
python -c 'import optuna; print(optuna.__version__)': 
python -c 'import platform; print(platform.python_version())': 3.9.6
python -c 'import platform; print(platform.platform())': Windows-10-10.0.19042-SP0
-->

- Optuna version: 2.10.0
- Python version: 3.9.6
- OS: Windows-10-10.0.19042-SP0
- (Optional) Other libraries and their versions:

SQLAlchemy==1.4.26
snowflake-connector-python==2.7.0
snowflake-sqlalchemy==1.3.2

### Error messages, stack traces, or logs

```
# error messages, stack traces, or logs
```

### Steps to reproduce

1. Try and create a study using snowflake

### Reproducible examples

```python
# python code
```

## Note to the questioner

If you are more comfortable with [Stack Overflow], you may consider posting your question with the "optuna" tag there instead.
Alternatively, for issues that would benefit from more of an interactive session with the developers,
you may refer to the [optuna/optuna chat] on Gitter.

[Stack Overflow]: https://stackoverflow.com/questions/tagged/optuna
[optuna/optuna chat]: https://gitter.im/optuna/optuna

bholland:
I was able to get this working through modifying a few things. 

1. the model no longer has any indexes
2. the model no longer has any Check Constraints
3. all primary keys must declare an explicit sequence (e.g., trial_heartbeat_id = Column(Integer, primary_key=True) becomes trial_heartbeat_id = Column("trial_heartbeat_id", Integer, sa_sequence("trial_heartbeat_id_seq"), primary_key=True))
4. for_update = False must get set in the models.py file. select * for update is not supported on snowflake. TBH, I didn't quite get the need for this.
5. I had to create a file in my Libs/site-packages/alembic/ddl/ called snowflake.py populated with

from .impl import DefaultImpl

class SnowflakeImpl(DefaultImpl):
    __dialect__ = "snowflake"

[...]

PR #3101, opened 2021-11-15
snowflake connections

bholland:
**DO NOT APPROVE THIS PR** 

As per the discussion on how to build support for snowflake, these are the minimal changes I made to optuna to get it to work in my environment. I am looking at ways to get this to work within SQLAlchemy to keep a single model file and add bits that are available to specific platforms. More on that tomorrow. 

## Motivation
I needed snowflake support to start multiple processing jobs in AWS and our company demanded snowflake support. 

## Description of the changes
This breaks models. It will remove code that is not supported by snowflake to get snowflake to run. It also adds a class declaration if the snowflake engine is snowflake. That is there to avoid a key error within alembic since optuna uses the python alembic bindings rather than relying on the env.py.

### DIFF
--- a/optuna/storages/_rdb/models.py
+++ b/optuna/storages/_rdb/models.py
@@ -13,6 +13,7 @@ from sqlalchemy import ForeignKey
 from sqlalchemy import func
 from sqlalchemy import Integer
 from sqlalchemy import orm
+from sqlalchemy import Sequence as sa_sequence
 from sqlalchemy import String
 from sqlalchemy import Text
 from sqlalchemy import UniqueConstraint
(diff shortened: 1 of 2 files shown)


hvy:
@toshihikoyanase, @himkt could you check this PR as I understand that it's a follow up to https://github.com/optuna/optuna/issues/3094.

himkt (commented):
Thank you for the patch. As I said in the previous discussion, it is difficult to merge your change to Optuna directly. The reasons are summarized as following:

- we cannot maintain snowflake storage module (Optuna dev-team doesn't have access to Snowflake DB)
- adding dependencies for DWH solution is not reasonable. we focus on keeping the dependencies slim. For RDB, our targets are MySQL, PostgreSQL so far. We want users to use {MySQL,PostgreSQL}-compatible RDB solution on the cloud (GCP, AWS, ...)

By the way, you successfully implement the Snowflake based storage. How about making the codebase a third-party library and use it in Optuna?

```python
import optuna

from your_awesome_snowflake_study_module import RDBStorage


def objective(trial):
    ...
    accuracy = model.eval(x_valid, y_valid)
    return accuracy

storage = RDBStorage(snowflake_url)  # craete your snowflake storage
study = optuna.create_study(storage=storage)  # specify it in create_study
```

bholland:
I could do that. The second option is to enable a bunch of sqlalchemy
checks to build out indexes. Let me put that together and see if that is a
bit better. My goal there would be to have no impact if you don't have
snowflake and connect without an issue if you do.

Thanks
Ben

PR #3103, opened 2021-11-16
Check if texts in legend by order agnostic way

himkt:
CI failed after matplotlib v3.5.0 is available.
https://github.com/optuna/optuna/runs/4224429852?check_suite_focus=true

This is because the order of texts in legend has changed in the latest matplotlib.

3.4.3 | 3.5.0
-- | --
 | 

The order of texts is potentially undefined so we shouldn't test them based on it.
I updated the test to remove order specific checks.

### DIFF
--- a/tests/visualization_tests/matplotlib_tests/test_optimization_history.py
+++ b/tests/visualization_tests/matplotlib_tests/test_optimization_history.py
@@ -83,11 +83,17 @@ def test_plot_optimization_history_with_multiple_studies(direction: str) -> None
     figure = plot_optimization_history(studies)
     assert len(figure.get_lines()) == n_studies
 
+    legend_texts = []
+    expected_legend_texts = []
     for i, legend in enumerate(figure.legend().get_texts()):
+        legend_texts.append(legend.get_text())
+
         if i < n_studies:
-            assert legend.get_text() == f"Best Values of {studies[i].study_name}"
+            expected_legend_texts.append(f"Best Values of {studies[i].study_name}")
         else:
-            assert legend.get_text() == f"Objective Value of {studies[i-n_studies].study_name}"
+            expected_legend_texts.append(f"Objective Value of {studies[i-n_studies].study_name}")
+
+    assert sorted(legend_texts) == sorted(expected_legend_texts)
 
     # Test customized target.
     with pytest.warns(UserWarning):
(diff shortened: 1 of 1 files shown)


tests/visualization_tests/matplotlib_tests/test_optimization_history.py
```diff
@@ -96,12 +102,21 @@ def objective(trial: Trial) -> float:
     assert len(figure.get_legend().get_texts()) == n_studies
 
     # Test customized target name.
-    figure = plot_optimization_history(studies, target_name="Target Name")
-    assert (
-        figure.legend().get_texts()[n_studies].get_text()
```
himkt, line 101: What `figure.legend().get_texts()[n_studies]` returns is also potentially undefined.

HideakiImamura (approved):
Thanks for your quick fix. LGTM.

HideakiImamura:

tests/visualization_tests/matplotlib_tests/test_optimization_history.py
```diff
@@ -83,11 +83,17 @@ def objective(trial: Trial) -> float:
...
+        legend_texts.append(legend.get_text())
+
         if i < n_studies:
-            assert legend.get_text() == f"Best Values of {studies[i].study_name}"
+            expected_legend_texts.append(f"Best Values of {studies[i].study_name}")
         else:
-            assert legend.get_text() == f"Objective Value of {studies[i-n_studies].study_name}"
+            expected_legend_texts.append(f"Objective Value of {studies[i-n_studies].study_name}")
+
+    assert sorted(legend_texts) == sorted(expected_legend_texts)
```
toshihikoyanase, line 96: I don't think it is a good practice to create expected values using the actual values.
For example, if `figure.legend().get_texts()` returns smaller number of items than `n_studies`, I guess the assertion still passes.

How about preparing the expected values based on `n_studies`?

The following is just an example.
```suggestion
    expected_legend_texts = []
    for i in range(n_studies):
        expected_legend_texts.append(f"Best Values of {studies[i].study_name}")
        expected_legend_texts.append(f"Objective Value of {studies[i].study_name}")
    legend_texts = [legend.get_text() for legend in figure.legend().get_texts()]
    assert sorted(legend_texts) == sorted(expected_legend_texts)
```

tests/visualization_tests/matplotlib_tests/test_optimization_history.py
```diff
@@ -96,12 +102,23 @@ def objective(trial: Trial) -> float:
...
+
+        if i < n_studies:
+            expected_legend_texts.append(f"Best Values of {studies[i].study_name}")
+        else:
+            expected_legend_texts.append(
+                f"{custom_target_name} of {studies[i-n_studies].study_name}"
+            )
+
+    assert sorted(legend_texts) == sorted(expected_legend_texts)
+    assert figure.get_ylabel() == custom_target_name
```
toshihikoyanase, line 121: ditto.

```suggestion
    expected_legend_texts = []
    for i in range(n_studies):
        expected_legend_texts.append(f"Best Values of {studies[i].study_name}")
        expected_legend_texts.append(f"{custom_target_name} of {studies[i].study_name}")
    legend_texts = [legend.get_text() for legend in figure.legend().get_texts()]
    assert sorted(legend_texts) == sorted(expected_legend_texts)
    assert figure.get_ylabel() == custom_target_name
```

toshihikoyanase (commented):
Maybe I missed something, but I think we can make the test cases simpler and more robust. Please correct me if I'm wrong.

toshihikoyanase (approved):

PR #3105, opened 2021-11-19
Add regex filtering support to `get_all_study_summaries()`

Turakar:
## Motivation
Querying all study summaries just to select a subset of them from a large RDB database is awfully slow.

## Description of the changes
With this option, one can filter the results by a regex beforehand for all storage backends. This increased the query speed for a large remote database of mine from ~770 seconds to ~16 seconds by reducing the number of studies from ~3500 to ~50. I did not test for performance on the other storage backends, but I would expect performance increases there, too.

Note that I only execute the regex statements on the client side. The main reason for this is portability: While some SQL databases have a regex operator, the support varies between them. Also, Redis does not have such an operator.

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -294,8 +294,12 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         raise NotImplementedError
 
     @abc.abstractmethod
-    def get_all_study_summaries(self) -> List[StudySummary]:
+    def get_all_study_summaries(self, regex: str = None) -> List[StudySummary]:
         """Read a list of :class:`~optuna.study.StudySummary` objects.
+
+        Args:
+            regex:
+                A regular expression that is used to filter the studies by name.
 
         Returns:
             A list of :class:`~optuna.study.StudySummary` objects.
--- a/optuna/storages/_cached_storage.py
+++ b/optuna/storages/_cached_storage.py
@@ -169,9 +169,9 @@ class _CachedStorage(BaseStorage):
 
         return self._backend.get_study_system_attrs(study_id)
 
-    def get_all_study_summaries(self) -> List[StudySummary]:
+    def get_all_study_summaries(self, regex: str = None) -> List[StudySummary]:
 
-        return self._backend.get_all_study_summaries()
+        return self._backend.get_all_study_summaries(regex=regex)
 
     def create_new_trial(self, study_id: int, template_trial: Optional[FrozenTrial] = None) -> int:
 
--- a/optuna/storages/_in_memory.py
+++ b/optuna/storages/_in_memory.py
@@ -1,5 +1,6 @@
 import copy
 from datetime import datetime
+import re
 import threading
 from typing import Any
 from typing import cast
@@ -150,10 +151,19 @@ class InMemoryStorage(BaseStorage):
             self._check_study_id(study_id)
             return self._studies[study_id].system_attrs
 
-    def get_all_study_summaries(self) -> List[StudySummary]:
+    def get_all_study_summaries(self, regex: str = None) -> List[StudySummary]:
 
         with self._lock:
-            return [self._build_study_summary(study_id) for study_id in self._studies]
+            if regex is not None:
+                pattern = re.compile(regex)
+                studies = [
+                    study_id
+                    for study_id, study in self._studies.items()
+                    if pattern.match(study.name)
+                ]
+            else:
+                studies = list(self._studies.keys())
+            return [self._build_study_summary(study_id) for study_id in studies]
 
     def _build_study_summary(self, study_id: int) -> StudySummary:
         study = self._studies[study_id]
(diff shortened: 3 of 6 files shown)


toshihikoyanase:
If we introduce the filtering feature you proposed, we may also want additional filtering, for example, by dates or the number of trials. So, let me add `needs-discussion` label and discuss the design before diving into the details.

I thought not so many people use `get_all_study_summaries` function. Honestly speaking, I've hardly used it since I usually check the study names using `optuna studies` command and then load the target study. So, could you explain your use case in more detail?

Turakar:
Regarding additional filters, one should check that these are quick to evaluate. If they are, I can add them to this PR. However, I would advise against over-complicating this PR, as this will hinder a quick merge.

To my use case: I often create multiple similar studies, e.g. to have multiple iterations or compare various datasets and models. To collect the finished studies from the database, my analysis script uses `get_all_study_summaries()` to see a list of studies that it can filter for the number of trials and the associated user attributes (where the test dataset evaluation results are stored in my case). AFAIK, there is no other official way to collect all study names. With RegEx filtering, I can significantly improve the performance of this workflow as discussed above. I could use a manually generated list of study names, but this hinders me from dynamically reacting to newly created studies, e.g. when I increase the number of iterations to shrink the confidence intervals.

One note about `optuna studies`: This command uses `get_all_study_summaries()` internally and could therefore profit from the performance increase as well with an additional command line argument in some use cases (and is therefore equally slow at the moment). However, I do not use the CLI because I already set up a different toolset using SQL queries in PgAdmin - these are significantly faster for large databases. The same holds for the Optuna dashboard, which could provide a search function this way.

toshihikoyanase:
Thank you for sharing your use case. You usually create many variants of studies, and you filter them when you collect the optimization results.

>  it can filter for the number of trials and the associated user attributes (where the test dataset evaluation results are stored in my case).

So, you also filter studies by the number of trials and `study.user_attributes`, right?

> AFAIK, there is no other official way to collect all study names.

You're right. Users need to use `optuna.get_all_study_summaries` to fetch study names for now.

After reading your comment, I examined the implementation of `get_all_study_summaries` and found some room to speed up the processing. I observed that the queries in the `for study in study_summary:` loop significantly took the retrieval times and I tried to reduce them in #3108 and #3109. I think `include_best_trial=False` in #3109 shows competitive performance to this PR because this PR filter the studies on the client-side. Do you think you still need the `regex` option after speeding up the retrieval?

Turakar:
> Thank you for sharing your use case. You usually create many variants of studies, and you filter them when you collect the optimization results.

Correct.

> So, you also filter studies by the number of trials and `study.user_attributes`, right?

Correct. Filtering by these attributes, too, was not required to achieve a good enough performance, but would be still interesting. Filtering based on `study.user_attributes` would require custom filters in the form of callbacks or similar.

> You're right. Users need to use `optuna.get_all_study_summaries` to fetch study names for now.
> 
> After reading your comment, I examined the implementation of `get_all_study_summaries` and found some room to speed up the processing. I observed that the queries in the `for study in study_summary:` loop significantly took the retrieval times and I tried to reduce them in #3108 and #3109. I think `include_best_trial=False` in #3109 shows competitive performance to this PR because this PR filter the studies on the client-side. Do you think you still need the `regex` option after speeding up the retrieval?

Thank you for considering my use case and developing an alternative solution! I would propose to stall this PR until I had a chance to test your proposed changes with my database - unfortunately, I cannot share it with you. As the speed-up you achieved is 6x on your database, but the RegEx filtering achieved 48x on my database, I would first like to test this before redrawing this proposal.

On a high level, I agree that it would be nice to have a performance of `get_all_study_summaries()` that makes filtering obsolete and thereby reduces code complexity.

On a technical side: Do you plan a release in the near future or should I look into going nightly for testing your PRs? I won't be able to test your changes until next week.

Turakar:
In principle, it should be possible to query all study summaries in a single query using `JOIN` and `GROUP BY`. However, I only know SQL, and not SQLAlchemy's ORM formalism.

toshihikoyanase:
> For reference, I created the following SQL statement, which would be equivalent to #3109 with include_best_trial=False. This takes 9.826 seconds for completion on my database, and would therefore be quick enough.

Thank you for your investigation. Then, I'll ask the committers to review #3108 and #3109.

> The main slow-down comes from LEFT JOIN trials, by the way - from 0.229s to 9.826s. It might be worth considering to not doing this join if the user does not need it.

Interesting. Without `LEFT JOIN` we can get only `studies.study_name` and `studies.study_id`, and the method can be `optuna.get_all_study_names() -> List[str]`.

Turakar:
> Interesting. Without `LEFT JOIN` we can get only `studies.study_name` and `studies.study_id`, and the method can be `optuna.get_all_study_names() -> List[str]`.

One can also retrieve the user and system attributes, which would be of interest for my use case.

HideakiImamura:

keisuke-umezawa:
@Turakar I merged https://github.com/optuna/optuna/pull/3108 into master. You can check it if you install the master branch.

Turakar:
It seems to me like the `user_attrs` are always empty. Can you reproduce that?

PR #3108, opened 2021-11-22
Reduce number of queries to fetch `directions`, `user_attrs` and `system_attrs` of study summaries.

toshihikoyanase:
## Motivation

Related to #3105.
It takes a long time to fetch study summaries if we have many studies in the RDB storage. For example, I have about 100 studies in a MySQL server, and it took about 30 seconds to fetch all study summaries with `optuna studies` commend. I'd like to speed up this processing.

The current master fetches `directions`, `user_attrs` and `system_attrs` study-wise, but I think it finally fetches all of them when it exits the for-loop. So, if we can reduce the number of queries from O(n_studies) to O(1) if we fetch them at once.

With this change, I could reduce the retrieval time from 30 seconds to 13 seconds.

```console
# command
$ optuna studies --storage $OPTUNA_STORAGE
# master
1.56s user 1.30s system 9% cpu 31.713 total
# this PR
1.44s user 1.08s system 19% cpu 12.838 total
```

## Description of the changes

- Fetch `StudyDirectionModel`, `StudyUserAttributeModel` and `StudySystemAttributeModel` by `.all()`

Issue #2943, opened 2021-09-20
Reorganize storages API and make BaseStorage public


keisuke-umezawa:
blocked by #2938

## Motivation

Currently, `BaseStorage` is private to users, and it is not expected that users will inherit `BaseStorage` and implement new storage classes. Also, even if you actually try to implement it, there are many methods that must be implemented, and the requirements that each method must satisfy are unclear. So, there are high hurdles for general users to implement them. By reorganizing the storages API, we want to reduce the number of methods that must be implemented and clarify the requirements that each method must satisfy. After that, I want to make `BaseStorage` public if necessary.

## Description

The storage API will be reorganized from the following viewpoints.

* Reduce the number of methods that must be implemented
* Clarify the requirements that each method must satisfy

After that, considering the pros / cons of making BaseStorage public, and if pros is stronger, make it public.

## TODOs
- [ ] Redesine storages API to reduce their methods
- [ ] Reduce the number of methods that must be implemented
- [ ] Clarify the requirements that each method must satisfy
- [ ] Make `BaseStorage` public if needed

### Redesine storages API to reduce their methods
The following methods can be removed from the APIs:
- [x] set_trial_values and set_trial_state https://github.com/optuna/optuna/pull/3323
- [ ] Add deprecation to get_n_trials.

#### get_trial_xxx
We can provide the default implementation in `BaseStorage` using `get_trial`. c.f. https://github.com/optuna/optuna/pull/2994
- [x] get_trial_number_from_id https://github.com/optuna/optuna/pull/3338
- [x] get_trial_param https://github.com/optuna/optuna/pull/3338
- [x] get_trial_params

#### set_trial_xxx
Need discussion

#### get_study_xxx
Need discussion

#### set_study_xxx
Need discussion

not522:
I think we should merge `set_trial_state` and `set_trial_values`.

Storage classes must ensure that the trials are valid after calling their methods. If they break this assumption, invalid trials may be read from other threads or processes.

In the current implementation of the procedure of setting `state` and `values`, other threads or processes may get `RUNNING` trials that have `values`.

https://github.com/optuna/optuna/blob/c4791fc0a67d1eb03ef650e583bf05f06af41dc4/optuna/study/study.py#L661-L664

This situation cannot avoid while `set_trial_values` and `set_trial_state` are called sequentially. Therefore, `set_trial_state` and `set_trial_values` should be merged.

masap:
[...]

PR #3109, opened 2021-11-22
Add option to exclude best trials from study summaries

toshihikoyanase:
## Motivation

This PR is related to #3105 and depends on #3108.

The current master includes the best trials to study summaries. It needs O(n_studies) queries to fetch the best studies, and we can reduce the retrieval time if we skip the processing.

The `optuna studies` command does not use `StudySummary.best_trial` and #3105 suggested that some users do not use it when they fetch study summaries in their Python scripts.

This PR adds an option to exclude the best trials from study summaries.

It significantly reduces the retrieval time in my environment (100 studies in a MySQL server) as follows:

```console
# command
$ optuna studies --storage $OPTUNA_STORAGE
# master
1.56s user 1.30s system 9% cpu 31.713 total
# #3108
1.44s user 1.08s system 19% cpu 12.838 total
# this PR
1.25s user 1.30s system 51% cpu 4.963 total
```

## Description of the changes

- Add `include_best_trial: bool = True` option to `BaseStorage.get_all_study_summaries`
- Add `include_best_trial: bool = True` option to `optuna.get_all_study_summaries`
- Set `include_best_trial` to `False` in the `optuna studies` command
- Add test cases for the option.

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -294,7 +294,7 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         raise NotImplementedError
 
     @abc.abstractmethod
-    def get_all_study_summaries(self) -> List[StudySummary]:
+    def get_all_study_summaries(self, include_best_trial: bool = True) -> List[StudySummary]:
         """Read a list of :class:`~optuna.study.StudySummary` objects.
 
         Returns:
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -294,8 +294,13 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         raise NotImplementedError
 
     @abc.abstractmethod
-    def get_all_study_summaries(self, include_best_trial: bool = True) -> List[StudySummary]:
+    def get_all_study_summaries(self, include_best_trial: bool) -> List[StudySummary]:
         """Read a list of :class:`~optuna.study.StudySummary` objects.
+
+        Args:
+            include_best_trial:
+                If :obj:`True`, :obj:`~optuna.study.StudySummary` objects have the best trials in
+                the ``best_trial`` attribute. Otherwise, ``best_trial`` is :obj:`None`.
 
         Returns:
             A list of :class:`~optuna.study.StudySummary` objects.
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -370,7 +370,7 @@ class _Studies(_BaseCommand):
     def take_action(self, parsed_args: Namespace) -> None:
 
         storage_url = _check_storage_url(self.app_args.storage)
-        summaries = optuna.get_all_study_summaries(storage=storage_url)
+        summaries = optuna.get_all_study_summaries(storage_url, include_best_trial=False)
 
         records = []
         for s in summaries:
--- a/optuna/storages/_cached_storage.py
+++ b/optuna/storages/_cached_storage.py
@@ -156,9 +156,9 @@ class _CachedStorage(BaseStorage):
 
         return self._backend.get_study_system_attrs(study_id)
 
-    def get_all_study_summaries(self) -> List[StudySummary]:
+    def get_all_study_summaries(self, include_best_trial: bool = True) -> List[StudySummary]:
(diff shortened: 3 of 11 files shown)


toshihikoyanase:
I resolved the conflict and made this PR review-ready.
Additionally, I confirmed that removing the `directions`, `user_attrs`, `system_attrs`, `n_trials` and `datetime_start` did not affect the processing time of `optuna studies`.

The number of studies: 156
Metrics: the average of 3 times execution
DB servers: MySQL
This PR: 4.41 sec
Remove more attributes: 4.49 sec

toshihikoyanase:
@keisuke-umezawa This PR is also related to #2943 since it includes the API change (adding an argument). Could you check if it is consistent with the Storage API reorganization?

toshihikoyanase:
@himkt Could you join the review after @keisuke-umezawa confirms that this PR is consistent with #2943. Also, your comments on the API consistency discussion are more than welcome.

himkt:
@keisuke-umezawa I'd like to know the current status of #2943.

This PR changes `get_all_study_summaries` and `_get_study_summary`, which is not seemingly targeted to be changed or moved in #2943. Please correct me if some discussion is ongoing related to them.

keisuke-umezawa (approved):
LGTM!

keisuke-umezawa:
@himkt 
> @keisuke-umezawa I'd like to know the current status of https://github.com/optuna/optuna/issues/2943 (pointing out the reason why this PR could be revised after discussion is fairly enough 👍).
> 
> This PR changes get_all_study_summaries and _get_study_summary, which is not seemingly targeted to be changed or moved in https://github.com/optuna/optuna/issues/2943. Please correct me if some discussion is ongoing related to them.

Actually, we can deal with this PR in the separated discussion from #2943 . We can review and judge we can merge it only by this PR.

optuna/storages/_base.py
```diff
@@ -294,7 +294,7 @@ def get_study_system_attrs(self, study_id: int) -> Dict[str, Any]:
         raise NotImplementedError
 
     @abc.abstractmethod
-    def get_all_study_summaries(self) -> List[StudySummary]:
+    def get_all_study_summaries(self, include_best_trial: bool = True) -> List[StudySummary]:
```
himkt, line 297: [question]

The same default value (`True` for `include_best_trial`) are both added in `optuna.storages.BaseStorage.get_all_study_summaries` and `optuna.get_all_study_summaries`. I'm not sure if it is a good practice.

Is it impossible to remove the default value from storage `get_all_study_summaries` methods?This should be breaking change but it would be acceptable if users don't invoke this method directly but invoke `optuna.get_all_study_summaries`, I think.

himkt:
@toshihikoyanase This PR looks solid. Please check the single question.

optuna/storages/_base.py
toshihikoyanase: Thank you for your suggestion. That makes sense.
The `optuna.get_all_study_summaries` is public API, and users basically use it instead of  `optuna.storages.BaseStorage.get_all_study_summaries`.

`optuna.storages.RDBStorage` is listed in the reference page, but [the reference](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.storages.RedisStorage.html) does not encourage to call it.

> Note that library users can instantiate this class, but the attributes provided by this class are not supposed to be directly accessed by them.

I fixed it in https://github.com/optuna/optuna/pull/3109/commits/e891e941cf6ee4b2bc509bdb5eeb12a6c251ce0c and https://github.com/optuna/optuna/pull/3109/commits/a3cd1c7d5bbc14b28a8d0effa09965079533001a. In addition, the missing document about the `include_best_trial` is added in https://github.com/optuna/optuna/pull/3109/commits/e891e941cf6ee4b2bc509bdb5eeb12a6c251ce0c.

himkt: Thank you for your deep consideration! It looks fair enough.

himkt:
Let me add a label `compatibility`, as it changes the signature of `get_all_study_summaries`.

himkt (approved):

PR #3111, opened 2021-11-22
Support `FloatDistribution` across codebase

xadrianzetx:
## Motivation

This PR introduces support for `FloatDistribution` across all places where conditional branching is being done based on instance of distribution. Inside of this newly created branch, logic specific to `LogUniformDistribution`, `DiscreteUniformDistribution` and `UniformDistribution` is recreated. This is a piece of #2941.

Note - this code will be dead (although still picked up by coverage report) until new distributions are instantiated in `suggest_*` APIs and in test suite (additional parametrization), which is probably another work piece coming right after this.

## Description of the changes

Support `FloatDistribution` in:
* [x] `optuna/samplers`
* [x] `optuna/integration`
* [x] `optuna/testing`
* [x] `optuna/visualization`
* [x] `_SearchSpaceTransform` class
* [x] Test parametrization

### DIFF
--- a/optuna/_transform.py
+++ b/optuna/_transform.py
@@ -11,6 +11,7 @@ import numpy
 from optuna.distributions import BaseDistribution
 from optuna.distributions import CategoricalDistribution
 from optuna.distributions import DiscreteUniformDistribution
+from optuna.distributions import FloatDistribution
 from optuna.distributions import IntLogUniformDistribution
 from optuna.distributions import IntUniformDistribution
 from optuna.distributions import LogUniformDistribution
@@ -191,6 +192,7 @@ def _transform_search_space(
                 DiscreteUniformDistribution,
                 IntUniformDistribution,
                 IntLogUniformDistribution,
+                FloatDistribution,
             ),
         ):
             if isinstance(d, UniformDistribution):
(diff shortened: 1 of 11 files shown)


toshihikoyanase:
@himkt @HideakiImamura Could you review this PR, please?

himkt:
Sure!

@xadrianzetx Sorry for the late response. I will review your PR until this weekend. Please wait for a while. 🙏

himkt:
Thank you @xadrianzetx for the PR! I went through the diff and the change looks solid.
Could you please add tests for sampler and transformer?
https://github.com/optuna/optuna/pull/3126 would be a reference.

xadrianzetx:
@himkt, I've added `FloatDistribution` parametrization and support to sampler and transformer tests. This should cover most of the new distribution testing, but there is still room to improve imo, e.g. additional parametrization in TPE tests. And we probably should gradually switch instances of distributions to their new counterparts across testing suite, same as we did with `suggest` APIs, right? Probably topics for potential follow-ups.

himkt:
Thank you so much @xadrianzetx for the update! I think tests looks enough for this stage.

> And we probably should gradually switch instances of distributions to their new counterparts across testing suite, same as we did with suggest APIs, right?

Yes it's the same as I expect. :+1: There would be a large number of tests failing 💥 after changing distributions instantiated in suggest APIs. Let's update tests together with changing suggest APIs.

himkt (approved):
LGTM. Thank you for the careful work!

HideakiImamura (commented):
Thanks for the PR. Let me check one thing first. Am I correct in assuming that the definitions and appearances of older distributions, such as UniformDistribution, will be replaced by FloatDistribution in future PRs? For example, I think the UniformDistribution definition should be replaced by a FloatDistribution wrapper. In that case, I think it would be sufficient to test only FloatDistribution instead of UniformDistribution in unit tests.

xadrianzetx:
> Am I correct in assuming that the definitions and appearances of older distributions, such as UniformDistribution, will be replaced by FloatDistribution in future PRs?

That's right! And when it won't be possible to hit codepaths of older distributions, we will remove them from tests as well.

HideakiImamura (approved):
I'm so sorry for my late review. My concerns are resolved by the previous comment. Let me approve this PR. LGTM!

xadrianzetx:
Thanks for the reviews! I'll work on instantiating new distribution now.

PR #3113, opened 2021-11-24
Migrate to {Float,Int}Distribution using alembic

himkt:
🔗 #2941

This PR updates users storage where distributions for integer and floating-point are unified to `{Float,Int}Distribution`.


tests/storages_tests/rdb_tests/create_db.py
```diff
@@ -6,13 +6,16 @@
...
+    x2 = trial.suggest_float("x2", -5, 5, step=2)
+    x3 = trial.suggest_float("x3", 1e-5, 1e-5, log=True)
+    y1 = trial.suggest_int("y1", 0, 10)
+    y2 = trial.suggest_int("y2", 1, 10, log=True)
     z = cast(float, trial.suggest_categorical("z", [-5, 0, 5]))
     trial.set_system_attr("a", 0)
     trial.set_user_attr("b", 1)
     trial.report(0.5, step=0)
-    return x ** 2 + y ** 2 + z ** 2
+    return x1 ** 2 + y1 ** 2 + z ** 2 + x2 + x3 + y2
```
himkt, line 18: This change affects tests of older versions of storage. I'm wondering which approach to use: (a) entirely rewriting tests with updating test assets, and (b) create a new script (e.g. create_db_v3.py) and tests focusing on distribution migration.

himkt: What do you think about it @toshihikoyanase?

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -183,7 +192,7 @@ def test_upgrade_identity() -> None:
     assert old_version == new_version
 
 
-@pytest.mark.parametrize("optuna_version", ["0.9.0", "1.2.0", "1.3.0", "2.4.0", "2.6.0"])
+@pytest.mark.parametrize("optuna_version", ["0.9.0", "1.2.0", "1.3.0", "2.4.0"])
```
himkt, line 195: @toshihikoyanase 

I'm not confident but I think it would not change anything in the case where `optuna_version` is `2.6.0`.
(Before this PR, `storage.get_current_version()` and `storage.get_head_version()` are both `2.6.0`.

So I updated `2.6.0.db` to use in `test_upgrade_distributions` and `test_upgrade_{single,multi}_objective_optimization`.
dcf497deb

If I misunderstand something, please tell me.

HideakiImamura:
@xadrianzetx Please feel free to leave any comments if you have.

HideakiImamura:
@himkt Any TODOs on this? :)

himkt:
https://github.com/optuna/optuna/pull/3113/files#diff-42166922cf20058a0456b29a4eccb65307689d7aefed170b335539b5e3ed11ebR284-R330

https://github.com/optuna/optuna/pull/3113/files#diff-b180a12fdf953ed7d1b05902ae4554bb2db534c51f784ddb87d8bd2e1ce796e7R8-R19

For testing a migration, we run an optimization using `suggest_int` and `suggest_float`. So, this PR should be merged after changing the behaviors of them. I think this PR is ready for review itself (any TODO here).

himkt:
Also, PTAL @HideakiImamura @toshihikoyanase as this PR is now ready for review, I think. 🙋

tests/storages_tests/rdb_tests/README.md
```diff
@@ -0,0 +1,32 @@
+### Migration test guide
```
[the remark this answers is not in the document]
xadrianzetx: Small nit - maybe change title to something like `Guide to creating migration test assets`.

tests/storages_tests/rdb_tests/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+1. Install Optuna
+
+You can know the current stable schema version by checking `optuna/storages/_rdb/alembic/versions`.
```
xadrianzetx, line 5:
```suggestion
You can view the current stable schema version by checking `optuna/storages/_rdb/alembic/versions`.
```

tests/storages_tests/rdb_tests/create_db_v3.py
```diff
@@ -0,0 +1,66 @@
+from argparse import ArgumentParser
```
xadrianzetx, line 1: This is not a strong opinion, as those are utility scripts or I might have missed something, but I'm wondering if we could just add flag to `create_db.py` which would enable additional params in objective functions for `2.6.0.db` up. This would also reduce code in tests, since we could use existing parametrization.

xadrianzetx (commented):
It's great to see green CI for new distributions! Let me leave some minor comments on this PR.

tests/storages_tests/rdb_tests/create_db_v3.py
```diff
@@ -0,0 +1,66 @@
...
+import optuna
+
+
+def objective_test_upgrade_v3(trial: optuna.trial.Trial) -> float:
+    x1 = trial.suggest_float("x1", -5, 5)
+    x2 = trial.suggest_float("x2", -6, 6, step=2)
+    x3 = trial.suggest_float("x3", 1e-5, 1e-3, log=True)
+    y1 = trial.suggest_int("y1", 0, 10)
+    y2 = trial.suggest_int("y2", 1, 20, log=True)
+    y3 = trial.suggest_int("y3", 5, 15, step=3)
```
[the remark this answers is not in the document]
himkt: I agree, thank you!

tests/storages_tests/rdb_tests/create_db_v3.py
himkt: Thank you for the insightful suggestion. I'll try to make the script unify!

himkt: Unified in 3b314c1. :eyes:

xadrianzetx (approved):
LGTM! Feel free to resolve outstanding conversations, as I can't.

himkt:
Thank you @xadrianzetx for the review! ❤️

xadrianzetx:
Any time :+1:

tests/storages_tests/rdb_tests/test_storage.py
HideakiImamura: Certainly the current master does nothing with this test for `optuna_version="2.6.0"`. This test confirms that running the upgrade on 2.6.0 does not cause any errors at all. On the other hand, adding "2.6.0" here and creating a `test_upgrade_assets/2.6.0.db` with `optuna==2.6.0` will give us the following advantages when we need the storage migration in some Optuna version (e.g. 3.0.0).
- You can find out the version of Optuna that required migration in the past by looking at the code, without having to go through the PR on GitHub.
- You don't need to create a db file for the past optuna version, you can use it immediately from the code base.

Therefore, we should add "2.6.0" here in the past, and create db file with `optuna==2.6.0`.

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
```diff
@@ -0,0 +1,227 @@
...
+        session.bulk_save_objects(distributions)
+        session.commit()
+    except SQLAlchemyError as e:
+        session.rollback()
+        raise e
+    finally:
+        session.close()
+
+
+def downgrade() -> None:
```
HideakiImamura, line 207: 👍

HideakiImamura (commented):

By the way, the current unit tests only confirms the upgrade with SQLite. How about testing with MySQL and PostgreSQL, manually, like this: https://gist.github.com/toshihikoyanase/a398d53827fc7597265213b009e43316

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -267,6 +278,140 @@ def test_upgrade_multi_objective_optimization(optuna_version: str) -> None:
...
+            assert trial.params["z"] in (-5, 0, 5)
+            assert -25 <= trial.values[0] <= 25
+            assert 0 <= trial.values[1] <= 820
+
+        assert study.system_attrs["c"] == 2
+        assert study.user_attrs["d"] == 3
+
+
+@pytest.mark.parametrize("optuna_version", ["2.6.0"])
+def test_upgrade_distributions(optuna_version: str) -> None:
```
HideakiImamura, line 380: Isn't this distribution upgrade test something you want to do only with 2.6.0? I don't think parametrize is necessary.

tests/storages_tests/rdb_tests/create_db.py
```diff
@@ -15,42 +17,80 @@ def objective_test_upgrade(trial: optuna.trial.Trial) -> float:
...
+def mo_objective_test_upgrade_old(trial: optuna.trial.Trial) -> Tuple[float, float]:
+    x = trial.suggest_uniform("x", -5, 5)
     y = trial.suggest_int("y", 0, 10)
     z = cast(float, trial.suggest_categorical("z", [-5, 0, 5]))
     trial.set_system_attr("a", 0)
     trial.set_user_attr("b", 1)
     return x, x**2 + y**2 + z**2
 
 
+def objective_test_upgrade(trial: optuna.trial.Trial) -> float:
```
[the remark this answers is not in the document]
himkt: The motivation behind that is to cover all paths of `migrate_new_distribution`.
It'd be safer to check all distributions are converted correctly by the method.

https://github.com/optuna/optuna/pull/3113/files#diff-16230fabaf41ea46b09fd25f3169755b56b954694b602983a8d453867963230dR88-R135

tests/storages_tests/rdb_tests/test_storage.py
himkt: You're right. `objective_test_upgrade` contains a distribution that is not supported by old versions of Optuna to made me parametrize here. As you pointed out we should tests migration from other versions of Optuna so let me take time to improve.

himkt:
I added the following changes:

- f22681f...3f2fa18: split objective functions based on reasons (checking suggest values and storage migration)
- 3f2fa18...21cf1dd: add tests for v3.0.0

PTAL @HideakiImamura @toshihikoyanase.

tests/storages_tests/rdb_tests/test_storage.py
himkt: Sorry for not responding you, your comment totally make sense. I revised the PR to leave 2.6.0.db (albeit adding a new study) and add 3.0.0.db.

HideakiImamura (commented):
Sorry for the late. I leave several comments. Could you take a look?

himkt:
Thank you for the review @HideakiImamura. I revised my PR and added replies on review comments, PTAL.

tests/storages_tests/rdb_tests/create_db.py
```diff
@@ -24,6 +65,20 @@ def mo_objective_test_upgrade(trial: optuna.trial.Trial) -> Tuple[float, float]:
     return x, x**2 + y**2 + z**2
 
 
+def objective_test_upgrade_v3_distribution(trial: optuna.trial.Trial) -> float:
```
HideakiImamura, line 68: Thanks for the update of the name. I prefer to remove the `v3`.

> BTW, do you mean we will add a parametrization on test_upgrade_distributions after v3.0.0?

I'm not sure, but I think it is safe to test the later version of Optuna does not violate the instantiation rule of distributions. I think we can add the later version of Optuna in this parametrization.

tests/storages_tests/rdb_tests/create_db.py
```diff
@@ -24,6 +65,20 @@ def mo_objective_test_upgrade(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    x1 = trial.suggest_float("x1", -5, 5)
+    x2 = trial.suggest_float("x2", 1e-5, 1e-3, log=True)
+    x3 = trial.suggest_float("x3", -6, 6, step=2)
+    y1 = trial.suggest_int("y1", 0, 10)
+    y2 = trial.suggest_int("y2", 1, 20, log=True)
+    y3 = trial.suggest_int("y3", 5, 15, step=3)
+    z = cast(float, trial.suggest_categorical("z", [-5, 0, 5]))
+    trial.set_system_attr("a", 0)
+    trial.set_user_attr("b", 1)
+    trial.report(0.5, step=0)
```
HideakiImamura, line 78: We don't have set these attributes in this function judging from its use case.

tests/storages_tests/rdb_tests/create_db.py
himkt: Fair enough. 9ff961b revised the name of an objective as you originally suggested.

tests/storages_tests/rdb_tests/create_db.py
himkt: Oh, you're right... 01b9afb removes them. :bow:

tests/storages_tests/rdb_tests/create_db.py
```diff
@@ -24,6 +65,17 @@ def mo_objective_test_upgrade(trial: optuna.trial.Trial) -> Tuple[float, float]:
     return x, x**2 + y**2 + z**2
 
 
+def objective_test_upgrade_distribution(trial: optuna.trial.Trial) -> float:
```
HideakiImamura, line 68: A nit.
```suggestion
def objective_test_upgrade_distributions(trial: optuna.trial.Trial) -> float:
```

HideakiImamura (commented):
Thanks for the update! LGTM except a minor comment.

tests/storages_tests/rdb_tests/create_db.py
himkt: Applied in 1fffaa4. 👍

tests/storages_tests/rdb_tests/create_db.py
toshihikoyanase: I'm very sorry for the delayed response. The current code that adds a new objective function for recent versions of Optuna seems reasonable to me.

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
```diff
@@ -0,0 +1,224 @@
...
+    trial_id = Column(Integer, primary_key=True)
+    number = Column(Integer)
+    study_id = Column(Integer, ForeignKey("studies.study_id"))
+    state = Column(Enum(TrialState), nullable=False)
+    datetime_start = Column(DateTime)
+    datetime_complete = Column(DateTime)
+
+    study = orm.relationship(
+        StudyModel, backref=orm.backref("trials", cascade="all, delete-orphan")
+    )
```
toshihikoyanase, line 61: This attribute may be able to omitted since we use `bulk_save_objects` that does not check the relationship. Please refer to `v2.4.0.py`.

```suggestion
```

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
```diff
@@ -0,0 +1,224 @@
...
+    __table_args__: Any = (UniqueConstraint("trial_id", "param_name"),)
+    param_id = Column(Integer, primary_key=True)
+    trial_id = Column(Integer, ForeignKey("trials.trial_id"))
+    param_name = Column(String(MAX_INDEXED_STRING_LENGTH))
+    param_value = Column(Float)
+    distribution_json = Column(Text())
+
+    trial = orm.relationship(
+        TrialModel, backref=orm.backref("params", cascade="all, delete-orphan")
+    )
```
toshihikoyanase, line 75: ditto.
```suggestion
```

toshihikoyanase (commented):
I'm very sorry for the delayed response. I confirmed that I could migrate SQLite, MySQL and PostgreSQL, and the change basically looks great to me.
I added some small comments, so please take a look.

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -9,12 +9,21 @@
...
 from unittest.mock import patch
 
+from packaging import version
 import pytest
 from sqlalchemy.exc import IntegrityError
 
 from optuna import create_study
 from optuna import load_study
-from optuna import version
+from optuna import version as optuna_version
```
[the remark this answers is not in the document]
himkt: Fair enough, applied in 4dc2cf8.

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = "v3.0.0.a"
+down_revision = "v2.6.0.a"
+branch_labels = None
+depends_on = None
```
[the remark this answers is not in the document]
himkt: I see. I moved them in 3fc77a1. 🙋‍♂️

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
himkt: Thank you for the careful look. I removed relations in 3d45cbf.

HideakiImamura (approved):
Thanks for the long running PR! LGTM!

tests/storages_tests/rdb_tests/test_storage.py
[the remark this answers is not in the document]
toshihikoyanase: Thanks!

toshihikoyanase (commented):
Thank you for your update. I confirmed that the migration script successfully convert the distribution with MySQL, PostgreSQL, SQLite and MS SQL.

A nit-picky, but I added two small comments on blank lines.

toshihikoyanase:
This may be out-of-scope of this PR, but I'm curious if users are still able to add `UniformDistritubion` to storages by using `add_trial` or such lower-level APIs. If so, users cannot convert them after they execute `optuna storage upgrade` with this migration script.

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
```diff
@@ -0,0 +1,216 @@
...
+            log=False,
+            step=distribution.step,
+        )
+    elif isinstance(distribution, IntLogUniformDistribution):
+        new_distribution = IntDistribution(
+            low=distribution.low,
+            high=distribution.high,
+            log=True,
+            step=distribution.step,
+        )
```
toshihikoyanase, line 118: What happens when `IntLogUniformDistribution.log == True` and `IntLogUniformDistribution.step != 1`?

(@HideakiImamura memtioned that such inconsistent distributions  were not accepted by released versions of Optuna.)

toshihikoyanase: @HideakiImamura suggested that we can implicitly convert `step=1`. The minimum value of `step` is `1`, and the corresponding parameter values are supposed to be included by the converted distributions.

toshihikoyanase: @himkt suggested that it should be failed due to the invalid distribution arguments. The migration should be stopped, and users will delete the corresponding studies by themselves.

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
```diff
@@ -0,0 +1,216 @@
...
+
+    session = orm.Session(bind=bind)
+    try:
+        distributions = session.query(TrialParamModel).all()
+        for distribution in distributions:
+            distribution.distribution_json = migrate_new_distribution(
+                distribution.distribution_json,
+            )
+        session.bulk_save_objects(distributions)
+        session.commit()
```
toshihikoyanase, line 188: @himkt mentioned that we could divide batches for the users who have millions of trials.

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
```diff
@@ -0,0 +1,216 @@
...
+
+    session = orm.Session(bind=bind)
+    try:
+        distributions = session.query(TrialParamModel).all()
+        for distribution in distributions:
+            distribution.distribution_json = restore_old_distribution(
+                distribution.distribution_json,
+            )
+        session.bulk_save_objects(distributions)
+        session.commit()
```
toshihikoyanase, line 211: ditto.

toshihikoyanase (commented):
These are comments of mob-review with @HideakiImamura @himkt and me.

## RedisStorage

@himkt mentioned the data in `RedisStorage`.
The users of `RedisStorage` are supposed to convert the tables in a following steps:
1. convert RedisStorage to RDBStorage
2. migrate tables in RDBStorage
3. convert RDBStorage to RedisStorage

## `UniformDistribution` in `Study.ask` and `Study.add_trial`

@himkt mentioned that `UniformDistribution` in `Study.ask`.
`UniformDistribution` can be used in `Study.ask`. It will be persisted as `UniformDistribution`, not `FloatDisribution`. So, we'll need to convert them when we delete `UniformDisribution` in Optuna v6.0.0.

Alternatively, we can convert the old distributions to new ones in `Study.ask` and `Study.add_trial`.

The current approach of #3246 is the former one, but the latter one may be preferable since users can keep the storage consistent, and we may avoid the future storage migration. If so, we may extract `migrate_new_distribution` as a utility function for `optuna.distributions`.

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
toshihikoyanase: TODO: We'd like to check if we can re-run the migration script after removing the studies with invalid trials.

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
himkt: ✅  in 6c2df8c.

himkt:
00c300b makes migration to be performed in batches. PTAL.

himkt:
@HideakiImamura @toshihikoyanase 
324adf8 makes migration to be performed in batches. PTAL.

himkt:
Oh... sorry, it looks I failed to rebase...

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = "v3.0.0.a"
+down_revision = "v2.6.0.a"
+branch_labels = None
+depends_on = None
+
+MAX_INDEXED_STRING_LENGTH = 512
+BATCH_SIZE = 5000
```
HideakiImamura, line 46: Do you have any reasons for the value of 5000?

himkt: No strong evidence here. Documentation says for example 1000 for BATCH_SIZE empirically suite for large amount of records. So it may be sufficient to set 1000.

(I'm thinking that there could be users who have several thousands (1,000~3,000) of trial_params. For such case, refetching would happen on BATCH_SIZE=1,000. So I set 5,000 to aim to avoid refetching for most of use-cases.)

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
toshihikoyanase: I confirmed that we could re-run the migration script if it failed. So, users can re-run the migration after removing studies with invalid trials.

optuna/storages/_rdb/alembic/versions/v3.0.0.a.py
toshihikoyanase: Thanks!

HideakiImamura (approved):
Thanks for the update. LGTM.

toshihikoyanase (approved):
Thank you for your update. I confirmed the changes using [optuna-e2e](https://github.com/himkt/optuna-e2e/pull/1). 

LGTM!

PR #3114, opened 2021-11-27
Use `json.loads` to decode pruner configuration loaded from environment variables

himkt:
This PR changes the way to set/fetch pruner configuration to/from environment variables.

The current implementation of AllenNLP integration utilize `repr()` and `eval()` for storing and fetching parameters but it would be safer to avoid `eval()` invocation. For this purpose, I found JSON to be useful since all parameters to be stored are JSON decodable.

## Motivation
- Remove `eval()` usage from source code

## Description of the changes
- 200d1d changes the exception when a key is not found on environment variables (follow-up of https://github.com/optuna/optuna/pull/2977#discussion_r756009719)
- eb185c changes the way to store/fetch a pruner configuration

### DIFF
--- a/optuna/integration/allennlp/_executor.py
+++ b/optuna/integration/allennlp/_executor.py
@@ -171,8 +171,7 @@ class AllenNLPExecutor(object):
 
         if trial.study.pruner is not None:
             variable_manager.set_value("pruner_class", type(trial.study.pruner).__name__)
-            variable_manager.set_value("pruner_keys", list(pruner_params.keys()))
-            variable_manager.set_value("pruner_values", list(pruner_params.values()))
+            variable_manager.set_value("pruner_params", pruner_params)
 
     def _build_params(self) -> Dict[str, Any]:
         """Create a dict of params for AllenNLP.
--- a/optuna/integration/allennlp/_executor.py
+++ b/optuna/integration/allennlp/_executor.py
@@ -163,7 +163,7 @@ class AllenNLPExecutor(object):
         target_pid = psutil.Process().ppid()
         variable_manager = _VariableManager(target_pid)
 
-        pruner_params = _fetch_pruner_config(trial)
+        pruner_kwargs = _fetch_pruner_config(trial)
         variable_manager.set_value("study_name", trial.study.study_name)
         variable_manager.set_value("trial_id", trial._trial_id)
         variable_manager.set_value("storage_name", url)
@@ -171,7 +171,7 @@ class AllenNLPExecutor(object):
 
         if trial.study.pruner is not None:
             variable_manager.set_value("pruner_class", type(trial.study.pruner).__name__)
-            variable_manager.set_value("pruner_params", pruner_params)
+            variable_manager.set_value("pruner_kwargs", pruner_kwargs)
 
     def _build_params(self) -> Dict[str, Any]:
         """Create a dict of params for AllenNLP.
--- a/optuna/integration/allennlp/_executor.py
+++ b/optuna/integration/allennlp/_executor.py
@@ -14,7 +14,6 @@ from optuna._imports import try_import
 from optuna.integration.allennlp._environment import _environment_variables
 from optuna.integration.allennlp._variables import _VariableManager
 from optuna.integration.allennlp._variables import OPTUNA_ALLENNLP_DISTRIBUTED_FLAG
-from optuna.integration.allennlp._variables import SPECIAL_DELIMITER as DELIMITER
 
 
 with try_import() as _imports:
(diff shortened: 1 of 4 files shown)


optuna/integration/allennlp/_pruner.py
```diff
@@ -53,9 +52,9 @@ def wrapper(subclass: Any, *args: Any, **kwargs: Any) -> Any:
 
 
 def _create_pruner(
-    pruner_class: Optional[str],
```
himkt, line 56: Maked `pruner_*` non-optional because `_create_pruner` is called inside `AllenNLPPruningCallback` and it should raises an exception when `pruner_*` is not available.

https://github.com/optuna/optuna/blob/eb185c25232849e30f3e4ff218fd28ebb27c2f88/optuna/integration/allennlp/_pruner.py#L175-L179

https://github.com/optuna/optuna/blob/eb185c25232849e30f3e4ff218fd28ebb27c2f88/optuna/integration/allennlp/_variables.py#L63-L73

toshihikoyanase:
Sounds reasonable 👍 
Let me review this PR.

optuna/integration/allennlp/_executor.py
```diff
@@ -165,17 +164,15 @@ def __init__(
...
+        variable_manager.set_value("trial_id", trial._trial_id)
         variable_manager.set_value("storage_name", url)
         variable_manager.set_value("monitor", metrics)
 
         if trial.study.pruner is not None:
             variable_manager.set_value("pruner_class", type(trial.study.pruner).__name__)
-            variable_manager.set_value("pruner_keys", DELIMITER.join(pruner_params.keys()))
-            variable_manager.set_value("pruner_values", DELIMITER.join(pruner_params.values()))
+            variable_manager.set_value("pruner_keys", list(pruner_params.keys()))
+            variable_manager.set_value("pruner_values", list(pruner_params.values()))
```
HideakiImamura, line 175: How about just passing the `pruner_params` instead of dividing it into keys and values since they are re-constructed as `pruner_params` in the `_create_pruner`?

HideakiImamura (commented):
Thanks for the PR. The change is basically reasonable to me. Please take a look at my comment.

optuna/integration/allennlp/_variables.py
```diff
@@ -51,23 +50,23 @@ def _get_key(self, name: str) -> str:
...
         `get_value` is only called in `optuna.integration.allennlp.AllenNLPPruningCallback`.
 
         """
         key = self._get_key(name).format(self.target_pid)
         value = os.environ.get(key)
-        assert value is not None, f"{key} is not found in environment variables."
-
-        return value
+        if value is None:
+            raise KeyError(f"{key} is not found in environment variables.")
```
HideakiImamura, line 71: Since this `get_value` is only called inside the allennlp integration, we should assert the value is not None instead of rising an error.
```
assert value is not None
```

HideakiImamura (commented):
Thanks for the update! I have one more minor comment. PTAL.

optuna/integration/allennlp/_variables.py
himkt: Hmm, I think it conflicts with the previous discussion.
https://github.com/optuna/optuna/pull/2977#discussion_r756009719

> we should assert the value is not None instead of rising an error

`assert` statement also raises an error (`AssertionError`) so the difference is only the type of an exception (Both could raise an error). I'd like to know what you intend in more detail. 🐈

HideakiImamura: I missed that comment. Thanks for the information. I think we need to explain all errors that users may face except `AssertionError` in the document, but do not the current one. But, users must not face this error, it's OK to leave as this is. Let me resolve this comment.

HideakiImamura (approved):
LGTM.

himkt:
Let me kindly ping @toshihikoyanase. 🐈

optuna/integration/allennlp/_pruner.py
```diff
@@ -53,9 +51,8 @@ def wrapper(subclass: Any, *args: Any, **kwargs: Any) -> Any:
...
-    pruner_keys: Optional[str],
-    pruner_values: Optional[str],
+    pruner_class: str,
+    pruner_params: Dict[str, Any],
```
toshihikoyanase, line 55: `pruner_params` is used as kwargs like `pruner(**pruner_params)`, so I think `pruner_kwargs` may show the usage better.
I should mention it in the previous PR, though.

toshihikoyanase (commented):
I'm sorry for the delayed response. The change overall looks good to me, but I just add a small comment.

optuna/integration/allennlp/_pruner.py
himkt: It makes sense, thank you for the comment. Updated in c09f27f.

toshihikoyanase (approved):

optuna/integration/allennlp/_pruner.py
toshihikoyanase: Thanks!

Issue #2957, opened 2021-09-26
Make progress bar available with `n_jobs!=1`


nzw0301:
## Motivation

`study.optimize(show_progress_bar)` is only available when `n_jobs=1`.  However, it would be great to show the progress bar when `n_jobs != 1`.

I think we have two things to resolve this issue:

### 1. support `n_jobs` with a single process

For single progress, the current implementation with `n_jobs != 1` works as we expected.

code: https://github.com/nzw0301/optuna/commit/4df16b1372b10f053e54a78aef7bb861b04feaa0

https://user-images.githubusercontent.com/7121753/134806308-4931f0b2-c76a-4a75-8556-57fd3d644452.mov

The example code used in the screen capture is as follows

```python
import time
import optuna
​
​
def objective(trial):
    time.sleep(0.5)
​
    return 1.
​
​
study = optuna.create_study()
study.optimize(objective, n_trials=20, show_progress_bar=True, n_jobs=3)
```

### 2. support n_jobs with [mutlple processes](https://optuna.readthedocs.io/en/stable/tutorial/10_key_features/004_distributed.html#sphx-glr-tutorial-10-key-features-004-distributed-py): 

I've not tested the code above with multi prosses execution. I suppose that each process should show the same progress bar above.

### 1. single process version

- modify the existing optuna code by following https://github.com/nzw0301/optuna/commit/4df16b1372b10f053e54a78aef7bb861b04feaa0
- add test by following https://github.com/optuna/optuna/pull/2900

### 2. multi process

- confirm the single process code can be used for the multi-process case.
- add test if it is possible.

## Additional context (optional)

`n_jobs` is deprecated, but we might reintroduce `n_jobs` again on Optuna v3 roadmap. So this issue depends on the decision of https://github.com/optuna/optuna/issues/2937.

PR #3115, opened 2021-11-27
Show progress bar based on `timeout`

### LINKED ISSUE #2958: [RFC] Show progress bar based on `timeout`
## Motivation

The current [`study.optimize`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html#optuna.study.Study.optimize) shows the progress bar when we specify the number of trials. We can show the progress bar based on the `timeout` value as well.

Note that before implementing this feature, we need to decide the priority of `n_trials` and `timeout`. When we specify both values as not `None` values, there are two ways to show the progress bar. 

Moreover, we need to discuss the case when both `None`. I suppose not showing the progress bar makes sense when both are `None`.

## Description

- Show progress bar using `timeout` value.
- Add test by following https://github.com/optuna/optuna/pull/2900

We might need to support the `n_jobs!=1` case depending on https://github.com/optuna/optuna/issues/2957.

xadrianzetx:
## Motivation

This PR makes progress bar functional when optimization is being limited by `timeout` rather than `n_trials`.  In this case, bar represents percentage of allocated time already spend on optimization. Formatted elapsed time and timeout are also displayed. This implementation will prioritize `n_trials` when both `n_trials` and `timeout` are specified. Closes #2958

## Description of the changes

* Show progress bar based on `timeout`
* Write tests

### `timeout` specified

Peek 2021-11-28 11-55

### `n_trials` specified (same as current master)

Peek 2021-11-28 11-58

### Both `n_trials` and `timeout` specified (`n_trials` takes priority)

Peek 2021-11-28 12-00

### Neither `n_trials` nor `timeout` specified (progress bar automatically disabled)

Peek 2021-11-28 12-02

### DIFF
--- a/optuna/progress_bar.py
+++ b/optuna/progress_bar.py
@@ -87,7 +87,7 @@ class _ProgressBar(object):
         if self._is_valid:
             if self._n_trials is not None:
                 self._progress_bar.update(1)
-                if self._timeout is not None and elapsed_seconds is not None:
+                if self._timeout is not None:
                     self._progress_bar.set_postfix_str(
                         "{:.02f}/{} seconds".format(elapsed_seconds, self._timeout)
                     )
--- a/optuna/progress_bar.py
+++ b/optuna/progress_bar.py
@@ -1,5 +1,4 @@
 import logging
-import time
 from typing import Any
 from typing import Optional
 
@@ -65,7 +64,7 @@ class _ProgressBar(object):
             # to display formatted timeout, since postfix carries
             # extra comma space auto-format.
             # https://github.com/tqdm/tqdm/issues/712
-            total = time.strftime("%M:%S", time.gmtime(self._timeout))
+            total = tqdm.format_interval(self._timeout)
             self._progress_bar.set_description_str(total)
 
         global _tqdm_handler
--- a/optuna/progress_bar.py
+++ b/optuna/progress_bar.py
@@ -1,4 +1,5 @@
 import logging
+import time
 from typing import Any
 from typing import Optional
 
@@ -39,9 +40,11 @@ class _ProgressBar(object):
     def __init__(
         self, is_valid: bool, n_trials: Optional[int] = None, timeout: Optional[float] = None
     ) -> None:
-        self._is_valid = is_valid
+
+        self._is_valid = is_valid and (n_trials or timeout) is not None
         self._n_trials = n_trials
         self._timeout = timeout
+        self._last_elapsed_seconds = 0.0
 
         if self._is_valid:
             self._init_valid()
@@ -50,7 +53,21 @@ class _ProgressBar(object):
     # longer experimental.
     @experimental("1.2.0", name="Progress bar")
     def _init_valid(self) -> None:
-        self._progress_bar = tqdm(range(self._n_trials) if self._n_trials is not None else None)
+
+        if self._n_trials is not None:
+            self._progress_bar = tqdm(total=self._n_trials)
+
+        else:
+            fmt = "{percentage:3.0f}%|{bar}| {elapsed}/{desc}"
+            self._progress_bar = tqdm(total=self._timeout, bar_format=fmt)
+
+            # Using description string instead postfix string
+            # to display formatted timeout, since postfix carries
+            # extra comma space auto-format.
+            # https://github.com/tqdm/tqdm/issues/712
+            total = time.strftime("%M:%S", time.gmtime(self._timeout))
+            self._progress_bar.set_description_str(total)
+
         global _tqdm_handler
 
         _tqdm_handler = _TqdmLoggingHandler()
@@ -59,22 +76,35 @@ class _ProgressBar(object):
         optuna_logging.disable_default_handler()
         optuna_logging._get_library_root_logger().addHandler(_tqdm_handler)
 
-    def update(self, elapsed_seconds: Optional[float]) -> None:
+    def update(self, elapsed_seconds: float) -> None:
         """Update the progress bars if ``is_valid`` is :obj:`True`.
 
         Args:
             elapsed_seconds:
                 The time past since :func:`~optuna.study.Study.optimize` started.
         """
+
         if self._is_valid:
-            self._progress_bar.update(1)
-            if self._timeout is not None and elapsed_seconds is not None:
-                self._progress_bar.set_postfix_str(
-                    "{:.02f}/{} seconds".format(elapsed_seconds, self._timeout)
-                )
+            if self._n_trials is not None:
+                self._progress_bar.update(1)
+                if self._timeout is not None and elapsed_seconds is not None:
+                    self._progress_bar.set_postfix_str(
+                        "{:.02f}/{} seconds".format(elapsed_seconds, self._timeout)
+                    )
+
+            else:
+                assert self._timeout is not None
+                time_diff = elapsed_seconds - self._last_elapsed_seconds
+                if elapsed_seconds > self._timeout:
+                    # Clip elapsed time to avoid tqdm warnings.
+                    time_diff -= elapsed_seconds - self._timeout
+
+                self._progress_bar.update(time_diff)
+                self._last_elapsed_seconds = elapsed_seconds
 
     def close(self) -> None:
         """Close progress bars."""
+
         if self._is_valid:
             self._progress_bar.close()
             assert _tqdm_handler is not None
--- a/optuna/progress_bar.py
+++ b/optuna/progress_bar.py
@@ -92,8 +92,7 @@ class _ProgressBar(object):
                         "{:.02f}/{} seconds".format(elapsed_seconds, self._timeout)
                     )
 
-            else:
-                assert self._timeout is not None
+            elif self._timeout is not None:
                 time_diff = elapsed_seconds - self._last_elapsed_seconds
                 if elapsed_seconds > self._timeout:
                     # Clip elapsed time to avoid tqdm warnings.
@@ -101,6 +100,9 @@ class _ProgressBar(object):
 
                 self._progress_bar.update(time_diff)
                 self._last_elapsed_seconds = elapsed_seconds
+
+            else:
+                assert False
 
     def close(self) -> None:
         """Close progress bars."""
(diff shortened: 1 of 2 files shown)


HideakiImamura:

xadrianzetx:
Going through the code again, I've spotted a line that needed a fix. Sorry if somebody started a review already.

toshihikoyanase (commented):
Thank you for your great work. This may be out-of-scope of this PR, but I'm thinking of the relationship with #2957, i.e., multi-threading support for the progress bar.

cc. @nzw0301 Please feel free to add your comments as the author of #2957.

optuna/progress_bar.py
```diff
@@ -59,22 +76,35 @@ def _init_valid(self) -> None:
...
+
+            else:
+                assert self._timeout is not None
+                time_diff = elapsed_seconds - self._last_elapsed_seconds
+                if elapsed_seconds > self._timeout:
+                    # Clip elapsed time to avoid tqdm warnings.
+                    time_diff -= elapsed_seconds - self._timeout
+
+                self._progress_bar.update(time_diff)
+                self._last_elapsed_seconds = elapsed_seconds
```
toshihikoyanase, line 103: This might be out-of-scope of this PR, but I'm concerned that we need to rewrite this logic when we support multi-threading. As described in #2957, Optuna plans to enable the progress bar with `n_jobs!=1` and #3138 conducts on the change.

I found that this code did not work as expected.

1. The percentage can exceed 100%.
```console
[I 2021-12-01 18:16:01,496] Trial 78 finished with value: 0.00023993382411588578 and parameters: {'x': 0.00023993382411588578}. Best is trial 51 with value: 5.7434369678507294e-05.                                                                                    
 98%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▍     | 00:03/00:04time_diff=0.13739999999999997, elapsed_seconds=3.886893, last_elapsed_seconds=3.749493
/home/yanase/pfn/code/optuna-review/.venv/lib/python3.8/site-packages/tqdm/std.py:533: TqdmWarning: clamping frac to range [0, 1]
  full_bar = Bar(frac,
[I 2021-12-01 18:16:01,513] Trial 93 finished with value: 0.0693096165017975 and parameters: {'x': 0.0693096165017975}. Best is trial 51 with value: 5.7434369678507294e-05.                                                                                            
/home/yanase/pfn/code/optuna-review/.venv/lib/python3.8/site-packages/tqdm/std.py:533: TqdmWarning: clamping frac to range [0, 1]
  full_bar = Bar(frac,
101%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 00:03/00:04
```

2. The percentage can decrease after `elapsed_seconds` exceeds `timeout`.

```console
101%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 00:04/00:04time_diff=-0.22978699999999996, elapsed_seconds=4.236397, last_elapsed_seconds=4.229787
[I 2021-12-01 18:16:01,893] Trial 86 finished with value: 0.04476754994720062 and parameters: {'x': 0.04476754994720062}. Best is trial 51 with value: 5.7434369678507294e-05.                                                                                          
 95%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▋            | 00:04/00:04time_diff=-0.2363970000000002, elapsed_seconds=4.28378, last_elapsed_seconds=4.236397
[I 2021-12-01 18:16:01,909] Trial 89 finished with value: 0.0032808017066235364 and parameters: {'x': 0.0032808017066235364}. Best is trial 51 with value: 5.7434369678507294e-05.                                                                                      
 89%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏                          | 00:04/00:04time_diff=-0.28378000000000014, elapsed_seconds=4.300264, last_elapsed_seconds=4.28378
[I 2021-12-01 18:16:01,923] Trial 95 finished with value: 0.000660676902939195 and parameters: {'x': 0.000660676902939195}. Best is trial 51 with value: 5.7434369678507294e-05.                                                                                        
 82%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▋                                            | 00:04/00:04
```


Do you think we can extend this logic to support multi-threading?

optuna/progress_bar.py
```diff
@@ -59,22 +76,35 @@ def _init_valid(self) -> None:
         optuna_logging.disable_default_handler()
         optuna_logging._get_library_root_logger().addHandler(_tqdm_handler)
 
-    def update(self, elapsed_seconds: Optional[float]) -> None:
+    def update(self, elapsed_seconds: float) -> None:
```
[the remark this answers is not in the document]
xadrianzetx: That right!

optuna/progress_bar.py
xadrianzetx: I think the problem here is that each thread keeps its own timing, but references the same global progress bar. Without common source of time elapsed, it produces those artifacts above.

By looking at current implementation of `_optimize`, I think we can fix this by moving progress bar updates to executor context. In pseudocode:

```python
[...]
progress_bar = pbar_module._ProgressBar(show_progress_bar and n_jobs == 1, n_trials, timeout)
[...]

with ThreadPoolExecutor(max_workers=n_jobs) as executor:
    for n_submitted_trials in itertools.count():
        if study._stop_flag:
            break

        if (
            timeout is not None
            and (datetime.datetime.now() - time_start).total_seconds() > timeout
        ):
            break

        if n_trials is not None and n_submitted_trials >= n_trials:
            break

        if len(futures) >= n_jobs:
            completed, futures = wait(futures, return_when=FIRST_COMPLETED)
            for f in completed:
                # pbar update is handled in executor context, common for all tasks
                progress_bar.update((datetime.datetime.now() - time_start).total_seconds())
                f.result()

        futures.add(
            executor.submit(
                _optimize_sequential,
                study,
                func,
                1,
                timeout,
                catch,
                callbacks,
                gc_after_trial,
                True,
                time_start,
                None,  # we do not pass progress bar reference to task
            )
        )
```

optuna/progress_bar.py
```diff
@@ -59,22 +76,35 @@ def _init_valid(self) -> None:
...
-                )
+            if self._n_trials is not None:
+                self._progress_bar.update(1)
+                if self._timeout is not None:
+                    self._progress_bar.set_postfix_str(
+                        "{:.02f}/{} seconds".format(elapsed_seconds, self._timeout)
+                    )
+
+            else:
+                assert self._timeout is not None
```
HideakiImamura, line 96: We can call `study.optimize` without setting neither `n_trials` nor `timeout` as follows.
```
study.optimize(objective, callbacks=[optuna.study.MaxTrialsCallback(10)])
```
I think we should divide the cases as follows.
- `n_trials` is not None
- `timeout` is not None
- Otherwise (A progress bar cannot be enabled.)

HideakiImamura (commented):
Thanks for the PR! I am also wondering if it is possible to show the progress bar when `n_jobs!=1`, as @toshihikoyanase said. If you don't mind, can we discuss this a bit in this PR?

Other than that, I have one comment. Please check it out.

optuna/progress_bar.py
xadrianzetx: So I wrote a [quick and dirty](https://github.com/xadrianzetx/optuna/blob/3582b131f1d493d6cdf51265dfe2141c1385790d/optuna/study/_optimize.py#L84-L123) implementation of the above, which _appears_ to work. [Here's a collab notebook](https://colab.research.google.com/drive/1r88auxJBRqokHHaOh_RuWgZ6kcLsL_4s?usp=sharing) so you can try it out. The change could be introduced in #3138 and this PR is compatible with it, apart from small `elapsed` formatting issue.

optuna/progress_bar.py
xadrianzetx: Testing with callback you've provided, behavior is as expected - progress bar is disabled even with `show_progress_bar=True`. We are disabling progbar when neither `n_trial` nor `timeout` are specified by ensuring `is_valid=False` in constructor. This `assert` is there only to make mypy happy, and should never be actually raised. I see how this might be not the most readable solution. Alternatively, i could refactor into

```python
if self._n_trials is not None:
  [..]
elif self._timeout is not None:
  [..]
else:
  raise RuntimeError("Should not reach.")
```

I've seen similar codepath in optuna somewhere.

HideakiImamura: Thanks you for your explanation. I missed the flag of `self._is_valid`. It looks like easier to read as you suggested. Could you fix them?
```python
if self._n_trials is not None:
  [..]
elif self._timeout is not None:
  [..]
else:
  assert False
```

xadrianzetx: Done in 620c39be9b9bc8b93c0df11a1466a51a06b91aac!

HideakiImamura (approved):
Thanks for the update. Looks solid to me.

HideakiImamura:
@nzw0301 Could you give us another look on this?

nzw0301:
Looks great! Thank you for your contributions! 

One concern is that when `timeout` argument of `optimize` method takes more than or equal to 1 hour. For example

```python
import optuna
import time
​
def objectve(trial):
    time.sleep(10)
    return 1.
​
study = optuna.create_study()
study.optimize(objectve, show_progress_bar=True, timeout=60*60)

# tqdm's progress bar is something like as follows
# 0% ... 00:12/00:00
```

In this case, the hour part in `total` is omitted but we might expect `01:00:00`. What do you think about it?

xadrianzetx:
>In this case, the hour part in total is omitted but we might expect 01:00:00. What do you think about it?

You're right. I will look into making the formatting more robust.

xadrianzetx:
@nzw0301 it turns out `tqdm` has a nice public API for elapsed time formatting. So now we are handling timeouts > 60 min correctly and sure that elapsed and timeout formats will always be formatted in the same way. PTAL.

nzw0301 (approved):
Thank you for your swift follow-up updates! LGTM.

HideakiImamura (approved):
Thanks for the great improvement on the progress bar. LGTM!

Issue #3112, opened 2021-11-23
Remove flush update functionality from _CachedStorage


keisuke-umezawa:
## Motivation

_CachedStorage is designed to store multiple [updates](https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L44) and commit them at once by [_flush_trial](https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L417). However, due to the change of the design, _CachedStorage currently only stores some updates in the [set_trial_values](https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L292-L301), and does not in the other methods. It means that we no longer need the flush update functionality in _CachedStorage.

## Description

Remove flush update functionality from _CachedStorage:

- [self.updates](https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L44)
- Flush update implementation in each method e.g. https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L198-L215
- [_flush_trial](https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L417)
- [_get_updates](https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L362)
- [_update_trial](https://github.com/optuna/optuna/blob/c4ebc0394f3996720b8cd0829edfa8630d1b99c1/optuna/storages/_rdb/storage.py#L624)

## TODOs
- [ ] Remove flush update functionality from _CachedStorage
- [ ] Compare the speed benchmark between master and this branch

## Additional context (optional)

- By resolving the issue, 
    - we can reduce requirements for storages to be wrapped by _CachedStorage.
    - Also, we can get rid of some lines of code from  _CachedStorage.

HideakiImamura:
@keisuke-umezawa Is this issue completed by #3120?

keisuke-umezawa:
Yes!

PR #3120, opened 2021-11-28
Remove updates in cached storage

shu65:
## Motivation

This PR addresses https://github.com/optuna/optuna/issues/3112 .

## Description of the changes

I removed flush update functionality from _CachedStorage

- self.updates (https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L44)
- Flush update implementation in each method e.g. 
- _flush_trial (https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L417)
- _get_updates (https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L362)
-  _update_trial (https://github.com/optuna/optuna/blob/c4ebc0394f3996720b8cd0829edfa8630d1b99c1/optuna/storages/_rdb/storage.py#L624)

I compared the performance between master branch and the PR.
The benchmark code to measure the performance is following.
```
import optuna

def objective(trial):
    return x ** 2

study_name = 'example-study0'
study = optuna.create_study(study_name=study_name,
                            storage='sqlite:///./test.db',
                            load_if_exists=True)
%time study.optimize(objective, n_trials=100)
```

The computing time of master:
```
CPU times: user 6.79 s, sys: 832 ms, total: 7.62 s
Wall time: 11.7 s
```

The computing time of the PR:
```
CPU times: user 7.63 s, sys: 831 ms, total: 8.46 s
Wall time: 11.6 s
```

So, I confirmed that there is no problem with the performance of the PR.


hvy:
Thanks a lot for the PR @shu65. Let me go through this change today.
@ytsmiling do you also want to jump in and have a look?

optuna/storages/_cached_storage.py
```diff
@@ -204,15 +188,13 @@ def set_trial_state(self, trial_id: int, state: TrialState) -> bool:
...
-                cached_trial.state = state
-                updates.state = state
+                ret = self._backend.set_trial_state(trial_id, state)
 
+                cached_trial.state = state
                 if cached_trial.state.is_finished():
-                    now = datetime.datetime.now()
-                    updates.datetime_complete = now
-                    cached_trial.datetime_complete = now
-                return self._flush_trial(trial_id)
```
hvy, line 215: 📝 Last flush in `Study.tell`, which also in addition to the state used to include the trial value(s). After the change, the value is already "flushed" as this point.

ytsmiling:
Hi, thanks for the heads up. Let me check this PR!

tests/storages_tests/test_cached_storage.py
```diff
@@ -77,65 +54,44 @@ def test_uncached_set() -> None:
     storage = _CachedStorage(base_storage)
     study_id = storage.create_new_study("test-study")
 
+    trial_id = storage.create_new_trial(study_id)
```
hvy, line 57: Sorry, this comment doesn't apply to this line specifically but rather `test_uncached_set`, but the docs of this functions should be updated to reflect the new flushing behavior.

hvy (changes requested):
Looks good overall I think! Left a minor comment.

I'm wondering if we cannot get rid of `_CachedStorage._check_trial_is_updatable` as well, leaving the checking to the backend storage (and update the cache the last thing we do, after persisting). But the current changes are true to the description and well-scoped so it's not a strong opinion.

optuna/storages/_cached_storage.py
```diff
@@ -295,12 +276,11 @@ def set_trial_values(self, trial_id: int, values: Sequence[float]) -> None:
             cached_trial = self._get_cached_trial(trial_id)
             if cached_trial is not None:
                 self._check_trial_is_updatable(cached_trial)
-                updates = self._get_updates(trial_id)
                 cached_trial.values = values
-                updates.values = values
+                self._backend.set_trial_values(trial_id, values=values)
                 return
```
ytsmiling, line 281: This is not a strong opinion but I personally recommend to delete these two lines to eliminate code duplications. The same applies for `set_trial_intermediate_values`, `set_trial_user_attr`, `set_trial_system_attr`.

ytsmiling (commented):
Thank you for creating this PR. I found no obvious errors and I think this PR is almost ready to merge.

tests/storages_tests/test_cached_storage.py
shu65: Thank you for the comments. 

Is this fixed as you intended?
https://github.com/optuna/optuna/pull/3120/files#diff-4c5bf6b243e65dfd008619eaf78d0da4f88ee2c36948cd40fd74dc1a4ebc60e3R48-R50

hvy: Looks good. We might even write `The CachedStorage flushes any modification of trials to a persistent storage immediately.` which might be clearer. What do you think?

shu65: OK. I will modify it. Thank you

hvy: Thanks for the prompt fix, let me resolve this thread. 🙇

ytsmiling (approved):
Thank you for the prompt fix! LGTM!

hvy (approved):
Thanks for quick fixes, changes LGMT!
The implementation of this storage was simplified quite a bit 👍

hvy:
@HideakiImamura do you want to take a final look, and maybe merge?

HideakiImamura:
@hvy Ops, I assigned myself to the wrong one. Of course, you can merge without my approval.

Issue #3020, opened 2021-10-18
Implement `matplotlib.plot_optimization_history` with multiple studies 


nzw0301:
## Motivation
As a part of https://github.com/optuna/optuna/issues/2959.

By https://github.com/optuna/optuna/pull/2807,  
- [x] `plot_optimization_history` have accepted multiple studies as `study` arguments (https://github.com/optuna/optuna/pull/3062)
- [ ] introduced `error_bar` argument. 

For consistency with Plotly backend, we would like to use the same feature with the Matplotlib backend.

## Description

Implement similar logic introduced by https://github.com/optuna/optuna/pull/2807.

nzw0301:
Thanks to #3062, `matplotlib.plot_optimization_history` will be able to take multiple studies.

To resolve this issue, we need to introduce `error_bar` option by following https://github.com/optuna/optuna/pull/2807.

HideakiImamura:
Let me close this issue since #3122 has been merged.

PR #3122, opened 2021-11-28
Support error bar for matplotlib

TakuyaInoue-github:
## Motivation

For consistency,  We want to use error bars with matplotlib. Please check also https://github.com/optuna/optuna/issues/3020.

## Description of the changes

Implement 
I have modified the logic introduced by #2807 that can be used for matplotlib.

## Examples
```python
import optuna


def objective(trial):
    return trial.suggest_float("x", 0, 1) ** 2


n_studies = 5
studies = [optuna.create_study() for _ in range(n_studies)]
for study in studies:

plt.figure()
fig = optuna.visualization.matplotlib.plot_optimization_history(studies, error_bar=True)
plt.savefig("./error_bar.png", bbox_layout='tight')

```

download

### DIFF
--- a/optuna/visualization/matplotlib/_optimization_history.py
+++ b/optuna/visualization/matplotlib/_optimization_history.py
@@ -134,7 +134,8 @@ def _get_optimization_history_plot(
         ax = _get_optimization_histories_with_error_bar(studies, target, target_name, ax)
     else:
         ax = _get_optimization_histories(studies, target, target_name, ax)
-        plt.legend(bbox_to_anchor=(1.05, 1.0), loc="upper left")
+
+    plt.legend(bbox_to_anchor=(1.05, 1.0), loc="upper left")
     return ax
 
 
--- a/optuna/visualization/matplotlib/_optimization_history.py
+++ b/optuna/visualization/matplotlib/_optimization_history.py
@@ -172,9 +172,9 @@ def _get_optimization_histories_with_error_bar(
         for t in trials:
             target_values[t.number].append(_target(t))
 
-    mean_of_target_values = [np.mean(v) if len(v) > 0 else None for v in target_values]
-    std_of_target_values = [np.std(v) if len(v) > 0 else None for v in target_values]
-    trial_numbers = np.arange(max_trial_number + 2)[[v is not None for v in mean_of_target_values]]
+    mean_of_target_values = [np.mean(v) if len(v) > 0 else np.nan for v in target_values]
+    std_of_target_values = [np.std(v) if len(v) > 0 else np.nan for v in target_values]
+    trial_numbers = np.arange(max_trial_number + 2)[[v is not np.nan for v in mean_of_target_values]]
     means = np.asarray(mean_of_target_values)[trial_numbers]
     stds = np.asarray(std_of_target_values)[trial_numbers]
 
@@ -200,16 +200,16 @@ def _get_optimization_histories_with_error_bar(
             for i, t in enumerate(trials):
                 best_values[t.number].append(best_vs[i])
 
-        mean_of_best_values = [np.mean(v) if len(v) > 0 else None for v in best_values]
-        std_of_best_values = [np.std(v) if len(v) > 0 else None for v in best_values]
+        mean_of_best_values = [np.mean(v) if len(v) > 0 else np.nan for v in best_values]
+        std_of_best_values = [np.std(v) if len(v) > 0 else np.nan for v in best_values]
         means = np.asarray(mean_of_best_values)[trial_numbers]
         stds = np.asarray(std_of_best_values)[trial_numbers]
 
         ax.plot(trial_numbers, means, color="tab:red", label="Best Value")
         ax.fill_between(
             x=trial_numbers,
-            y1=np.array(means - stds, float),
-            y2=np.array(means + stds, float),
+            y1=means - stds,
+            y2=means + stds,
             color="tab:red",
             alpha=0.4,
         )
--- a/optuna/visualization/matplotlib/_optimization_history.py
+++ b/optuna/visualization/matplotlib/_optimization_history.py
@@ -174,7 +174,9 @@ def _get_optimization_histories_with_error_bar(
 
     mean_of_target_values = [np.mean(v) if len(v) > 0 else np.nan for v in target_values]
     std_of_target_values = [np.std(v) if len(v) > 0 else np.nan for v in target_values]
-    trial_numbers = np.arange(max_trial_number + 2)[[v is not np.nan for v in mean_of_target_values]]
+    trial_numbers = np.arange(max_trial_number + 2)[
+        [v is not np.nan for v in mean_of_target_values]
+    ]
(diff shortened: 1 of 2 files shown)


optuna/visualization/matplotlib/_optimization_history.py
```diff
@@ -128,11 +130,98 @@ def _get_optimization_history_plot(
...
+        yerr=stds,
+        capsize=5,
+        fmt="o",
+        color="tab:blue",
+    )
+    ax.scatter(trial_numbers, means, color="tab:blue", label=target_name)
+
+    best_values: List[List[float]] = [[] for _ in range(max_trial_number + 2)]
+
+    if target is None:
```
nzw0301, line 195: Since`best_values` is necessary when `target` is `None`, how about this?

```suggestion
    
    if target is None:
        best_values: List[List[float]] = [[] for _ in range(max_trial_number + 2)]
```

nzw0301 (changes requested):
Basically, LGTM. Brilliant! Could you slightly update the code?

nzw0301:
Sorry, could you add a docstring: https://github.com/optuna/optuna/blob/81d4db881ce5d65302d8b6cf60bc5275427ef831/optuna/visualization/_optimization_history.py#L69-L70?

optuna/visualization/matplotlib/_optimization_history.py
```diff
@@ -128,11 +130,93 @@ def _get_optimization_history_plot(
...
+        mean_of_best_values = [np.mean(v) if len(v) > 0 else None for v in best_values]
+        std_of_best_values = [np.std(v) if len(v) > 0 else None for v in best_values]
+        means = np.asarray(mean_of_best_values)[trial_numbers]
+        stds = np.asarray(std_of_best_values)[trial_numbers]
+
+        ax.plot(trial_numbers, means, color="tab:red", label="Best Value")
+        ax.fill_between(
+            x=trial_numbers,
+            y1=np.array(means - stds, float),
+            y2=np.array(means + stds, float),
```
nzw0301, line 210: I think `means` are `float` values, so we can drop the dtypes.

```suggestion
            y1=np.array(means - stds),
            y2=np.array(means + stds),
```

optuna/visualization/matplotlib/_optimization_history.py
```diff
@@ -128,11 +130,98 @@ def _get_optimization_history_plot(
...
+            for trial in study.get_trials(states=(TrialState.COMPLETE,))
+        ]
+    )
+
+    _target: Callable[
+        [
+            FrozenTrial,
+        ],
+        float,
+    ]
```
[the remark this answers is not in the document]
TakuyaInoue-github: I think this line break is supposed to be done by black. What do you think?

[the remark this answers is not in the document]
nzw0301: I suppose `_target: Callable[[FrozenTrial], float]` makes `black` happy.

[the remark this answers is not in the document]
TakuyaInoue-github: thanks, you're right

optuna/visualization/matplotlib/_optimization_history.py
TakuyaInoue-github: I think it seems to be right. but if case, I get an error like this `
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''`

HideakiImamura:
@himkt Could you review this PR if you have time?

optuna/visualization/matplotlib/_optimization_history.py
nzw0301: Thank you for investigating it! Indeed, it comes from filling `None` in `mean_of_target_values = [np.mean(v) if len(v) > 0 else None for v in target_values]`, for example.  So replacing `mean_of_target_values = [np.mean(v) if len(v) > 0 else None for v in target_values]` with  `mean_of_target_values = [np.mean(v) if len(v) > 0 else np.nan for v in target_values]` might be simpler. 

What do you think? cc: @himkt

tests/visualization_tests/matplotlib_tests/test_optimization_history.py
```diff
@@ -118,3 +118,54 @@ def fail_objective(_: Trial) -> float:
...
+
+    # Test customized target.
+    with pytest.warns(UserWarning):
+        figure = plot_optimization_history(studies, target=lambda t: t.number, error_bar=True)
+    assert len(figure.get_lines()) == 3
+
+    # Test customized target name.
+    figure = plot_optimization_history(studies, target_name="Target Name", error_bar=True)
+    legend_texts = [legend.get_text() for legend in figure.legend().get_texts()]
+    assert sorted(legend_texts) == sorted(["Best Value", "Target Name"])
```
nzw0301, line 160: simplified
```suggestion
    assert sorted(legend_texts) == ["Best Value", "Target Name"]
```

nzw0301 (changes requested):
let me leave a minor comment 🙇🏻

optuna/visualization/matplotlib/_optimization_history.py
```diff
@@ -128,11 +133,92 @@ def _get_optimization_history_plot(
...
+
+    target_values: List[List[float]] = [[] for _ in range(max_trial_number + 2)]
+    for study in studies:
+        trials = study.get_trials(states=(TrialState.COMPLETE,))
+        for t in trials:
+            target_values[t.number].append(_target(t))
+
+    mean_of_target_values = [np.mean(v) if len(v) > 0 else np.nan for v in target_values]
+    std_of_target_values = [np.std(v) if len(v) > 0 else np.nan for v in target_values]
+    trial_numbers = np.arange(max_trial_number + 2)[[v is not np.nan for v in mean_of_target_values]]
```
nzw0301, line 177: could you take a look my final suggestion?
```suggestion
    trial_numbers = np.arange(max_trial_number + 2)[
        [v is not np.nan for v in mean_of_target_values]
    ]
```

nzw0301: Indeed, it looks better.

himkt: It is different from the plotly's implementation. If you prefer to use the same logic, please ignore my suggestion and apply @nzw0301's one (or we can change the original implementation as a follow-up).

nzw0301: This is completely off-topic about this PR, but it is related to @himkt's comment, so let me leave my suggestion. 
When I review this PR, the logic to generate arguments of this plot function is really similar to ploty's one. So I'm wondering if we could merge the logic not depending on the backend of the packages: plotly/matplotlib.

nzw0301 (approved):
LGTM. Thank you for your consecutive pull requests on `matplotlib.plot_optimization_history`!

himkt (approved):
Thank you so much @TakuyaInoue-github for the awesome enhancement!

Issue #3116, opened 2021-11-28
Constraint-aware pareto-front plot


toshihikoyanase:
## Motivation

This is a sub-task of #2397.

## Description

Recently, Optuna enables us to add constraints to multi-objective optimization. The `plot_paret_front` function, however, does not consider whether the solution is feasible or not. So, I needed to modify it to visualize the constraint optimization.

For example, the following plot is an example of multi-objective knapsack problem. The objectives are to be maximized, but infeasible solutions are likely to be seen in the up-right region. In this case, I couldn't find the shape of Pareto front without removing the infeasible solutions.

![0199](https://user-images.githubusercontent.com/3255979/109284470-3cad0700-7863-11eb-82bc-63e48a530205.png)

We may have some design choices, but I think the first choice will be adding `constraints_func` to `plot_pareto_front`. This is because some samplers including [`NSGAIISampler`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.samplers.NSGAIISampler.html#optuna.samplers.NSGAIISampler) and [`BoTorchSampler`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.BoTorchSampler.html#optuna.integration.BoTorchSampler) have the same interfaces regarding constrained optimization.

```python
optuna.visualization.plot_pareto_front(study, *, ..., constraints_func: Optional[Callable[[FrozenTrial], Sequence[float]]] = None)
```

fukatani:
I started to work on this issue. thanks!

toshihikoyanase:
We still have some follow-up tasks.
See https://github.com/optuna/optuna/pull/3497#pullrequestreview-985660566

HideakiImamura:
Let me close this issue. We will work on the remaining followup task in the unifying different backends and tests.

PR #3125, opened 2021-11-28
Add `constraints_func` to `plot_pareto_front`

semiexp:
## Motivation
Related to #3116.

## Description of the changes
I added `constraints_func` property to `plot_parent_front` so that infeasible trials can be explicitly marked in case of training with objective constraints.

Here is an example of plots by this function:
test

Currently the implementation is repetitious, but I suppose refactoring is not the scope of this PR. Refactoring will take place in a separate PR.

### DIFF
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -1,7 +1,11 @@
 import json
 from typing import Any
+from typing import Callable
+from typing import Dict
 from typing import List
 from typing import Optional
+from typing import Sequence
+from typing import Tuple
 
 import optuna
 from optuna._experimental import experimental
@@ -24,6 +28,7 @@ def plot_pareto_front(
     target_names: Optional[List[str]] = None,
     include_dominated_trials: bool = True,
     axis_order: Optional[List[int]] = None,
+    constraints_func: Optional[Callable[[FrozenTrial], Sequence[float]]] = None,
 ) -> "go.Figure":
     """Plot the Pareto front of a study.
 
@@ -63,6 +68,16 @@ def plot_pareto_front(
         axis_order:
             A list of indices indicating the axis order. If :obj:`None` is specified,
             default order is used.
+        constraints_func:
+            An optional function that computes the objective constraints. It must take a
+            :class:`~optuna.trial.FrozenTrial` and return the constraints. The return value must
+            be a sequence of :obj:`float` s. A value strictly larger than 0 means that a
+            constraint is violated. A value equal to or smaller than 0 is considered feasible.
+            This specification is the same as in, for example,
+            :class:`~optuna.integration.BoTorchSampler`.
+
+            If given, trials are classified into three categories: feasible and best, feasible but
+            non-best, and infeasible. Categories are shown in different colors.
 
     Returns:
         A :class:`plotly.graph_objs.Figure` object.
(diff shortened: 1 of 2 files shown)


toshihikoyanase (commented):
Thank you for your PR. The change basically looks great!

I guess `constraints_func!=None` is not tested in the unit tests. Could you add test cases to https://github.com/optuna/optuna/blob/master/tests/visualization_tests/test_pareto_front.py?

semiexp:
Sorry, I forgot to add tests. Just added :)

toshihikoyanase:
Thanks! Let me review this PR!

optuna/visualization/_pareto_front.py
```diff
@@ -288,3 +408,65 @@ def _make_hovertext(trial: FrozenTrial) -> str:
...
+                    "x": 1.1 if include_dominated_trials else 1,
+                    "xpad": 40,
+                },
+            }
+    else:
+        assert color is not None
+        return {
+            "line": {"width": 0.5, "color": "Grey"},
+            "color": color,
+        }
```
toshihikoyanase, line 472: If `use_constraints_func==True`, `include_dominated_trials` seems to be ignored. I guess users may want to extract only the best feasible. What do you think of it?

semiexp: In this case, non pareto front trials are excluded in `_get_pareto_front_2d` or `_get_pareto_front_3d` (`_make_marker` need not care about `include_dominated_trials`).
However, pareto front but infeasible trials are still included. Do you mean this?

optuna/visualization/_pareto_front.py
```diff
@@ -133,42 +153,89 @@ def _get_pareto_front_2d(
...
+                mode="markers",
+                hovertemplate="%{text}<extra>Best Trial</extra>",
+                marker=_make_marker(
+                    trials[:n_best_trials],
+                    include_dominated_trials,
+                    use_constraints_func=False,
+                    dominated_samples=False,
+                ),
+            ),
+        ]
```
himkt, line 184: I'm just wondering if we can reduce lines here by iterating over pairs of range and label.
Just like:

```python
        data = [
            go.Scatter(
                x=[t.values[axis_order[0]] for t in trials[slice]],
                y=[t.values[axis_order[1]] for t in trials[slice]],
                text=[_make_hovertext(t) for t in trials[slice]],
                mode="markers",
                hovertemplate="%{text}<extra>" label "</extra>",
                marker=_make_marker(
                    trials[:n_best_trials],
                    include_dominated_trials,
                    use_constraints_func=False,
                    dominated_samples=False,
                ),
            )
            for slice, label in ((n_best_trials:, "Trial"), (:n_best_trials, "Best Trial"))
        ]
```

(I don't check if the code actually works mm)

himkt: I found this is addressed in https://github.com/optuna/optuna/pull/3128, please ignore.

toshihikoyanase (commented):
I found some differences on `Scatter()`.

optuna/visualization/_pareto_front.py
toshihikoyanase: Thank you for your explanation. When I wrote this comment, I thought some parts of the marker style should be changed. But I found that we don't need that since we don't use `colorscale` if `constraints_func` is specified.

> However, pareto front but infeasible trials are still included. Do you mean this?

As we talked in the dev sprint, we need to remove the infeasible solutions before extracting the Pareto solutions. Let's discuss this topic in #3128.

HideakiImamura:

semiexp:
Sorry, I have left this PR unattended...
At present `plot_pareto_front` still computes the pareto front based on all trials including infeasible ones. So we need to fix this first, but this requires computation of the pareto front of a `Study` ignoring infeasible trials.
Currently the best way to do so will be first creating another `Study` containing feasible trials only, but construction of a temporary `Study` would be tricky as it should be a storage as well. IMO the best option will be first changing the interface of `optuna.study._get_pareto_front_trials` so that it takes a sequence of `FrozenTrials`s as the argument, and using it in `plot_pareto_front` to compute the pareto front of feasible trials.

PR #3135, opened 2021-11-30
Avoid installing the latest MLfow to prevent doctests from failing

himkt:
## Motivation

Although I haven't investigated the cause yet, the doctest was failing with the latest MLflow.
https://github.com/optuna/optuna/runs/4368455449?check_suite_focus=true

I confirmed the problem doesn't occur with MLflow v1.21.0.
https://github.com/optuna/optuna/runs/4368582442?check_suite_focus=true
So, as a tentative patch, I introduce the version constraint.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ def get_extras_require() -> Dict[str, List[str]]:
             "plotly>=4.0.0",
             "scikit-learn>=0.24.2",
             "scikit-optimize",
-            "mlflow",
+            "mlflow==1.22.0",
         ],
         "document": [
             # TODO(nzw): Remove the version constraint after resolving the issue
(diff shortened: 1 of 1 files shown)


himkt:
https://github.com/optuna/optuna/blob/51b548324fb00e9b1cb4a60ac68bd0a438566bc4/optuna/integration/mlflow.py#L69

We expect callback to log a message by the above format.
MLflow changes a way of logging from `print` to `logging` from v1.22.0.
https://github.com/mlflow/mlflow/pull/5012/files#diff-b33420360cc29a5224c8d3080bbfcaa0ac1c8891416137e13be044fcdfdd26f9L84-R113

ref. https://github.com/mlflow/mlflow/pull/5012 pointed out by @nzw0301 

We have to follow the change in MLflow.

himkt:
We'll be able to update our doctest to follow the change in MLflow.
I'll look at the change and fix it later but I'd merge this PR for the first aid.
If someone is interested in, please comment here and work on! I'll be available after tomorrow. 🐱

himkt:
Sorry for the by-name mention but please take a look @nzw0301 @hvy.
I believe we share the context of this problem.

hvy (approved):
LGTM!

nzw0301 (approved):
LGTM. thanks!

Issue #3134, opened 2021-11-29
current status of n_jobs in v2.10?


rbelew:
### Question

I find a merge Jan 29 using `concurrent.futures` https://github.com/optuna/optuna/pull/2269, then 
I find  the `njobs` parameter was deprecated here https://github.com/optuna/optuna/pull/2560 on Apr 1.  And now there is discussion about reintroducing it here https://github.com/optuna/optuna/issues/2937 on Sept 18.  I also found an older work-arounds, using `joblib` https://gist.github.com/sile/2c329d0cb38f35e25a8c7bab44edbb63. 

I am wondering what the current recommended procedure for multi-core environments using v2.10.0 optuna.  What does the implementation currently do with `n_jobs` specified?  It seems to produce odd, non-montonic best values (see example below).

### Environment
```
>>> print(op.__version__) 
2.10.0
>>> print(platform.python_version()) 
3.8.5
>>> print(platform.platform()) 
macOS-10.15.7-x86_64-i386-64bit
```

### Error messages, stack traces, or logs

```
Trial 101 finished with best value: 3.051867903393622 and parameters: {'R0': 1.0712727856991768, 'alpha': 0.9121462925283631, 'gamma': 1.1823702204733835, 'mu': 0.036417631166185516}. 
Trial 188 finished with best value: 2.2365535937364998 and parameters: {'R0': 1.0694600337354254, 'alpha': 0.8999322001243446, 'gamma': 1.3864523757139084, 'mu': 0.02695870506276768}. 
Trial 277 finished with best value: 2.088390682306482 and parameters: {'R0': 1.054260021970593, 'alpha': 1.5693093881559335, 'gamma': 1.469040063506123, 'mu': 0.03048413180433311}. 
Trial 276 finished with best value: 15.34937870021244 and parameters: {'R0': 1.1377582001694242, 'alpha': 0.6932468664055964, 'gamma': 1.236927077386912, 'mu': 0.04018656386490023}. 
Trial 292 finished with best value: 1.3179174767632489 and parameters: {'R0': 1.0631375246172654, 'alpha': 1.5265556047812037, 'gamma': 1.3094584471794986, 'mu': 0.025065419768277475}. 
Trial 293 finished with best value: 1.6933340465788846 and parameters: {'R0': 1.0665697529499913, 'alpha': 1.5104281490808449, 'gamma': 1.4077384827255122, 'mu': 0.02765503785586052}. 

```
### Steps to reproduce

N/A

### Reproducible examples

[...]

PR #3143, opened 2021-12-03
Add a note logging_callback only works in single process situation

himkt:
🔗 https://github.com/optuna/optuna/issues/3134

## Motivation
This PR adds a note to FAQ that `logging_callback` doesn't work as expected with `n_jobs>1`.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -373,7 +373,7 @@ Here's an example:
     study = optuna.create_study()
     study.optimize(objective, n_trials=100, callbacks=[logging_callback])
 
-Note that this callback may show incorrect values when you try to optimize an objective function with `n_jobs!=1`
+Note that this callback may show incorrect values when you try to optimize an objective function with ``n_jobs!=1``
 (or other forms of distributed optimization) due to its reads and writes to storage that are prone to race conditions.
 
 How do I suggest variables which represent the proportion, that is, are in accordance with Dirichlet distribution?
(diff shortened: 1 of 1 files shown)


docs/source/faq.rst
```diff
@@ -373,6 +373,8 @@ Here's an example:
     study = optuna.create_study()
     study.optimize(objective, n_trials=100, callbacks=[logging_callback])
 
+Note that this callback would not work as expected when you try to optimize a objective function with `n_jobs>1`.
```
himkt, line 376: Not sure whether I should put a reason in detail here. 💭

himkt (commented):
@hvy I'd like your feedback. 🙋

docs/source/faq.rst
hvy: We could elaborate a bit perhaps. Something along

```suggestion
Note that this callback may show incorrect values when you try to optimize an objective function with `n_jobs>1` (or other forms of distributed optimization) due to its reads and writes to storage which are prone to race conditions.
```

hvy:
@keisuke-umezawa , could you also review this one?

docs/source/faq.rst
nzw0301: Nit
```suggestion
Note that this callback would not work as expected when you try to optimize a objective function with ``n_jobs!=1``.
```

himkt: Thank you @hvy @nzw0301 for the suggestions! I applied your changes in 463c890 (sorry I didn't commit your suggestions, suggestions conflicted each other to let me manual apply...🔥)

hvy (approved):
LGTM!

hvy:
@nzw0301 please feel free to review and/or merge if you'd like. 🙇

nzw0301 (approved):
LGTM

Issue #3132, opened 2021-11-29
`nan` handling of `optuna tell`.


ytsmiling:
When we perform `optuna tell` with `nan`, you'll get the following error message.

```text
$ optuna create-study --storage sqlite:///debug.db --study-name foo
[I 2021-11-29 15:22:23,775] A new study created in RDB with name: foo
foo
$ optuna ask --storage sqlite:///debug.db --study-name foo
.../python3.8/site-packages/optuna/cli.py:796: ExperimentalWarning: 'ask' is an experimental CLI command. The interface can change in the future.
warnings.warn(
[I 2021-11-29 15:22:31,214] Using an existing study with name 'foo' instead of creating a new one.
[I 2021-11-29 15:22:31,267] Asked trial 0 with parameters {}.
{"number": 0, "params": {}}
$ optuna tell --storage sqlite:///debug.db --study-name foo --trial-number 0 --values nan
.../python3.8/site-packages/optuna/cli.py:863: ExperimentalWarning: 'tell' is an experimental CLI command. The interface can change in the future.
warnings.warn(
[E 2021-11-29 15:22:43,074] Trial 0 failed, because the objective function returned nan.
```
However, the state of `trial 0` remains `RUNNING`.
```python
In [1]: import optuna
In [2]: optuna.load_study(storage="sqlite:///debug.db", study_name="foo").get_trials()[0]
Out[2]: FrozenTrial(number=0, values=None, datetime_start=datetime.datetime(2021, 11, 29, 15, 22, 31, 248437), datetime_complete=None, params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=1, state=TrialState.RUNNING, value=None)
```

## Expected behavior

I think there are several "correct" behavior. I personally prefer `pattern A` but this might not be optimal from a more general perspective.

### pattern A

`optuna tell` (more precisely, `study.tell`) set the study's state to `FAILED` and raise no errors just like `study.optimize`.

### pattern B

`optuna tell` (more precisely, `study.tell`) set the study's state to `FAILED` and raise an error.

### pattern C

Change the error message.

## Environment

- Optuna version: a3.0.0a1.dev
- Python version: 3.8.6
- OS: Linux-5.10.60.1-microsoft-standard-WSL2-x86_64-with-glibc2.29 
- (Optional) Other libraries and their versions:

## Additional context (optional)

None, null, and other nans:
```text
[...]

PR #3144, opened 2021-12-03
Move validation logic from `_run_trial` to `study.tell`

himkt:
🔗 https://github.com/optuna/optuna/issues/3132

This PR makes the behavior of `optuna tell` consistent with `study.optimize` when an objective value is `nan` or a list containing `nan`. Inside `study.optimize`, the case observing `nan` as an objective values is considered as a special case that Optuna doesn't raise an exception and make `trial.state` failed and show a log message. For Optuna CLI, I think it would be natural to behave as in `study.optimize`.

https://github.com/optuna/optuna/blob/cf33d05c6b6274c0ce24384af528d5fbc7dd0762/optuna/study/_optimize.py#L224-L230

https://github.com/optuna/optuna/blob/cf33d05c6b6274c0ce24384af528d5fbc7dd0762/optuna/study/_optimize.py#L256-L257

## Motivation
- Make `optuna tell` consistent with `study.optimize`
  - One may think `optuna tell` should only be consistent with `study.tell` . Any feedback is highly appreciated
    - If so, we should guide users to run `optuna tell ... --state fail` when they observe `nan`
- When `nan` is passed to `optuna tell`, Optuna makes a trial state FAILED

## Description of the changes
- abf53a655cd16509732000bb70009b633f447898 Add test case
- 84413f7e87231c1e07098801a0a044be1dd0b40d Invoke `_check_and_convert_to_values` in `_Tell.take_action`

## Behavior

```sh
> optuna create-study --storage sqlite:///example.db --study-name test
[I 2021-12-03 22:09:54,814] A new study created in RDB with name: test
test
```

```sh
> optuna ask --storage sqlite:///sample.db --study-name test
/home/himkt/work/github.com/himkt/optuna/optuna/cli.py:736: ExperimentalWarning: 'ask' is an experimental CLI command. The interface can change in the future.
  warnings.warn(
[I 2021-12-03 22:10:01,697] A new study created in RDB with name: test
[I 2021-12-03 22:10:01,715] Asked trial 0 with parameters {}.
{"number": 0, "params": {}}
```

```sh
> optuna tell --study-name test --storage sqlite:///sample.db --trial-number 0 --values nan
/home/himkt/work/github.com/himkt/optuna/optuna/cli.py:852: ExperimentalWarning: 'tell' is an experimental CLI command. The interface can change in the future.
  warnings.warn(
[I 2021-12-03 22:10:09,387] Told trial 0 with values None and state TrialState.FAIL.
[W 2021-12-03 22:10:09,387] Trial 0 failed, because the objective function returned nan.
```

```python
>>> import optuna
op>>> study = optuna.load_study(storage="sqlite:///sample.db", study_name="test")
>>> study.get_trials()
[FrozenTrial(number=0, values=None, datetime_start=datetime.datetime(2021, 12, 3, 22, 10, 1, 708719), datetime_complete=datetime.datetime(2021, 12, 3, 22, 10, 9, 371719), params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=1, state=TrialState.FAIL, value=None)]
>>>
```


himkt:
@g-votte and @hvy suggested to move the validation logic inside `Study.tell`. I read the code of `tell` and I'm not sure whether we should do or not. In the current logic of `Study.tell`, `state` is immutable and I feel this design is clean. But if we won't change this design, `Study.tell` and CLI `optuna  tell` would be different or we would force users to handle nan values on their own.

hvy:
Thanks a lot @himkt for quickly addressing the issue with an RFC. 

Just to clarify, do you mean that making `state` mutable (by defaulting to `None`), would make the behavior unnecessarily complicated, from a user-perspective, or from a code complexity point of view? 
I was imagining moving https://github.com/optuna/optuna/blob/release-v3.0.0-a0/optuna/study/_optimize.py#L223-L230 to `Study.tell` and having all the validation done there. `Study.optimize` would merely add logging on top of that. Would you like me to prototype on that or have you already tried perhaps?

My understanding is that the validation logic (currently only done in `Study.optimize`) isn't trivial and is costly for the ask-and-tell users to implement, especially via the CLI. I personally find it acceptable trying to move more logic to `Study.tell`, even though it'd make the API/behavior more involved. Perhaps starting with the CLI is a fine first approach, but I would argue that the interfaces are better designed kept consistent between the CLI and `Study.tell` at least for this topic.

himkt:
Sorry for the inactive @hvy.

> Just to clarify, do you mean that making state mutable (by defaulting to None), would make the behavior unnecessarily complicated, from a user-perspective, or from a code complexity point of view?

I said it by the point of code complexity. I was not sure how to tell a conversion failure to `study._optimize`.
But later I noticed we don't have to do so: it's enough to simply do logging in `tell`.

And I also found `_check_and_convert_to_values` is also called inside `study.tell`.
So, I simply tried to update it (many tests will fail. I'll check them) in c143c3a.
https://github.com/optuna/optuna/blob/10d1111f026d047ce42494b6f7c330a21c924c9c/optuna/study/study.py#L639-L647

hvy:
📝  Just had a discussion with @himkt outside this PR. This work is ready for review. 
📝  Make sure that `Study.tell`'s behavior is consistent (best-effort) with `Study.optimize`'s return value handling from the objective function.

optuna/study/study.py
```diff
@@ -651,13 +654,26 @@ def df(x):
 
         if values is not None:
             values, values_conversion_failure_message = _check_and_convert_to_values(
-                len(self.directions), values, trial_number
+                len(self.directions),
+                values,
+                trial_number,
```
HideakiImamura, line 659: Seems unnecessary. 
```suggestion
                len(self.directions), values, trial_number
```

HideakiImamura (commented):
Thanks for an important PR to resolve #3132. Also, I think this PR partially resolve #3205. I agree with the idea to make sure that Study.tell's behavior is consistent with Study.optimize's return value handling. 

I have a minor comment. Could you take a look?

HideakiImamura:
I confirmed that this PR makes the behavior of `Study.optimize` and `Study.tell` consistent as follows.

code: This is the same as `nan-tell.py` provided in #3205.
```python
import optuna
import numpy as np


# In-memory storage
study = optuna.create_study()
trial = study.ask()
try:
    study.tell(trial, float(np.nan))
except Exception as e:
    print(e)
finally:
    print(f"In-memory : Trial state is {study.trials[-1].state}")

# RDB storage (SQLite)
study = optuna.create_study(storage='sqlite:///../sqlite.db')
trial = study.ask()
try:
    study.tell(trial, float(np.nan))
except Exception as e:
    print(e)
finally:
    print(f"SQLite    : Trial state is {study.trials[-1].state}")

# RDB storage (MySQL)
study = optuna.create_study(storage="mysql+pymysql://root:test@localhost/mysql")
trial = study.ask()
try:
    study.tell(trial, float(np.nan))
except Exception as e:
    print(e)
finally:
    print(f"MySQL     : Trial state is {study.trials[-1].state}")

# RDB storage (PostgreSQL)
study = optuna.create_study(storage="postgresql+psycopg2://postgres:test@localhost:15432/postgres")
trial = study.ask()
try:
    study.tell(trial, float(np.nan))
except Exception as e:
    print(e)
finally:
    print(f"PostgreSQL: Trial state is {study.trials[-1].state}")
```

output:
```
(venv) mamu@HideakinoMacBook-puro 3205-3206 % python nan-tell.py
[I 2021-12-24 16:02:58,636] A new study created in memory with name: no-name-91d1fb55-6f9c-4c51-861a-42892b705424
[W 2021-12-24 16:02:58,637] Trial 0 failed, because the objective function returned nan.
In-memory : Trial state is TrialState.FAIL
[I 2021-12-24 16:02:58,709] A new study created in RDB with name: no-name-ccbae3aa-b661-4e01-913e-d2d489e4cb90
[W 2021-12-24 16:02:58,754] Trial 0 failed, because the objective function returned nan.
SQLite    : Trial state is TrialState.FAIL
[I 2021-12-24 16:02:58,907] A new study created in RDB with name: no-name-92237945-fb07-46b8-92cd-b0349805fb78
[W 2021-12-24 16:02:59,011] Trial 0 failed, because the objective function returned nan.
MySQL     : Trial state is TrialState.FAIL
[I 2021-12-24 16:02:59,322] A new study created in RDB with name: no-name-ffaba79b-3556-42da-b23a-29962316e788
[W 2021-12-24 16:02:59,439] Trial 0 failed, because the objective function returned nan.
PostgreSQL: Trial state is TrialState.FAIL
`

File: optuna/study/_optimize.py [hunk 2]
```diff
@@ -210,7 +208,10 @@ def _run_trial(
         thread.start()
 
     try:
-        value_or_values = func(trial)
+        value_or_values = cast(
+            Optional[Union[float, List[float]]],
+            func(trial),
+        )
```
hvy, line 214: Could you add a comment why the `cast` is needed?

optuna/study/_optimize.py
```diff
@@ -219,48 +220,35 @@ def _run_trial(
...
+        assert False, "Should not reach."
 
-    if state == TrialState.FAIL and func_err is not None and not isinstance(func_err, catch):
+    if (
+        frozen_trial.state == TrialState.FAIL
+        and func_err is not None
+        and not isinstance(func_err, catch)
+    ):
         raise func_err
     return trial
```
hvy, line 253: You can simplify the internal logic even further (at `_optimize_sequential`, the caller of `_run_trial`) now with the return value of `Study.tell`.
```suggestion
    return frozen_trial
```

optuna/study/_optimize.py
```diff
@@ -219,48 +220,35 @@ def _run_trial(
...
-            elif values_conversion_failure_message is not None:
-                _logger.warning(values_conversion_failure_message)
-            else:
-                assert False, "Should not reach."
-        else:
-            assert False, "Should not reach."
+    frozen_trial = study.tell(trial, values=value_or_values, state=state)
+
+    if frozen_trial.state == TrialState.COMPLETE:
+        study._log_completed_trial(trial, frozen_trial.values)
```
hvy, line 233: How about updating `_log_completed_trial` to take the frozen trial, simplifying the internals?
```suggestion
        study._log_completed_trial(frozen_trial)
```

hvy: Also, I think that the code would be cleaner if we avoid to use `trial` but instead `frozen_trial` from hereon.

optuna/study/_optimize.py
```diff
@@ -219,48 +220,35 @@ def _run_trial(
...
-        else:
-            state = TrialState.COMPLETE
 
     if study._storage.is_heartbeat_enabled():
         assert stop_event is not None
         assert thread is not None
         stop_event.set()
         thread.join()
 
-    # `Study.tell` may raise during trial post-processing.
```
hvy, line 238: Is changing this behavior (removing `finally` which ensures that results are always logged) also an intended part of this PR? Note that post-processing here refers to the after-trial sampler callback.

optuna/study/study.py
```diff
@@ -489,9 +489,9 @@ def tell(
         self,
         trial: Union[trial_module.Trial, int],
         values: Optional[Union[float, Sequence[float]]] = None,
-        state: TrialState = TrialState.COMPLETE,
+        state: Optional[TrialState] = None,
         skip_if_finished: bool = False,
-    ) -> None:
+    ) -> FrozenTrial:
```
hvy, line 494: 👍

hvy: By the way, how about updating the docs, about the return type, `state` parameter and the new failing behavior?

optuna/study/study.py
```diff
@@ -632,13 +630,18 @@ def df(x):
 
         frozen_trial = self._storage.get_trial(trial_id)
 
+        if state is None and values is None:
+            self._storage.set_trial_state(trial_id, TrialState.FAIL)
+            _logger.warning("You must specify either state or values.")
+            return self._storage.get_trial(trial_id)
```
hvy, line 636: Should this always be executed without taking into account `skip_if_finished` ? Wondering if something like the following can replace.
```suggestion
        if state is None and values is None:
            state = TrialState.FAIL
```

hvy: The logging message might be lacking information for the user, I think adding some context would help. (Besides from the comment above)

optuna/study/study.py
```diff
@@ -651,13 +654,26 @@ def df(x):
 
         if values is not None:
             values, values_conversion_failure_message = _check_and_convert_to_values(
```
hvy, line 656: Not a strong opinion and can be a separate PR but we can improve locality by moving the definition of `_check_and_convert_to_values` to this file now.

hvy (changes requested):
Looks good, the code becomes simpler with this change in my opinion. Left some comments still if you could take a look. 🙏

optuna/study/_optimize.py
hvy: Please note that the suggested change here is just a part of my suggestion in the comment.

optuna/study/_optimize.py
himkt: I tried to explain in 5c57e2b. Please give me a feedback. :pray:

optuna/study/_optimize.py
himkt: Thank you for the suggestion! It really makes sense to me.
aae3a54 aaf638d ad4e8b5 (ad4e8b5 6f413c6 for multi objective) updates the logic using `FrozenTrial`.

optuna/study/_optimize.py
himkt: same as https://github.com/optuna/optuna/pull/3144#discussion_r780758665

optuna/study/_optimize.py
himkt: Thank you for pointing out, this try-catch must not be removed.
03d80ea reverts the change, and 91d9424 fetches a frozen trial from the storage to avoid the unbounded error.

optuna/study/study.py
himkt: 9ef0d75 and 2cc402b update the documentation. ☝️

optuna/study/study.py
himkt: It makes sense. 137476c does it.

optuna/study/study.py
himkt: You suggestion is applied in adecbc4. I'll update message as well.

optuna/study/study.py
himkt: Done in 137476c. :+1:

optuna/study/study.py
himkt: I updated the doc in 6fa50b4. I'm not sure if it improves a usability so please give me a feedback. 🙏

himkt:
Thank you so much @hvy for the careful review. I revised my PR based on your comments/suggestions!

hvy:
Thanks a lot, I'll go through the changes today!

optuna/study/_optimize.py
```diff
@@ -237,93 +234,40 @@ def _run_trial(
...
+    except Exception as e:
+        tell_err = e
         raise
     finally:
-        if state == TrialState.COMPLETE:
-            study._log_completed_trial(trial, cast(List[float], values))
-        elif state == TrialState.PRUNED:
-            _logger.info("Trial {} pruned. {}".format(trial.number, str(func_err)))
-        elif state == TrialState.FAIL:
+        # NOTE (himkt):
```
HideakiImamura, line 242:
```suggestion
```

HideakiImamura (commented):
A minor comment

optuna/study/_optimize.py
hvy: Thanks, are we aligned that the error, which is raised by mypy is as follows, unless casted?

```
optuna/study/_optimize.py:216: error: Incompatible types in assignment (expression has type "Union[float, Sequence[float]]", variable has type "Union[float, List[float], None]")
```

It seems like you can change the type hint of `value_or_values` at its declaration (at line 195) from `List`, to `Sequence` as an alternative to the `cast`. Please correct me if I'm wrong.

optuna/study/_optimize.py
hvy: Thanks a lot, you can simplify it even further. There is a line in `_optimize_sequential` that obtains a frozen trial from the previously returned Trial. Now that we return a frozen trial, we can skip this line.

```
frozen_trial = copy.deepcopy(study._storage.get_trial(trial._trial_id))
```

optuna/study/_optimize.py
hvy: Looks great, thanks!

optuna/study/study.py
```diff
@@ -569,6 +633,9 @@ def df(x):
                 :class:`~optuna.trial.TrialState.PRUNED`.
                 ``trial`` is a trial number but no
                 trial exists with that number.
+
+        Returns:
+            A :class:`~optuna.trial.FrozenTrial`.
```
hvy, line 638: How about some elaboration.
```suggestion
            A :class:`~optuna.trial.FrozenTrial` representing the resulting trial.
```

optuna/study/study.py
```diff
@@ -543,9 +604,14 @@ def df(x):
...
             state:
-                State to be reported. Must be :class:`~optuna.trial.TrialState.COMPLETE`,
+                State to be reported. Must be :obj:`None`,
+                :class:`~optuna.trial.TrialState.COMPLETE`,
                 :class:`~optuna.trial.TrialState.FAIL` or
                 :class:`~optuna.trial.TrialState.PRUNED`.
+                If ``state`` is :obj:`None`,
+                it will be updated to :class:`~optuna.trial.TrialState.COMPLETE`
+                or :class:`~optuna.trial.TrialState.FAIL` depending on whether
+                a validation for ``values`` reported succeed or not.
```
hvy, line 614: Looks but, just a small nit.
```suggestion
                validation for ``values`` reported succeed or not.
```

optuna/study/_optimize.py
```diff
@@ -237,93 +234,39 @@ def _run_trial(
...
-        if state == TrialState.COMPLETE:
-            study._log_completed_trial(trial, cast(List[float], values))
-        elif state == TrialState.PRUNED:
-            _logger.info("Trial {} pruned. {}".format(trial.number, str(func_err)))
-        elif state == TrialState.FAIL:
+        # Even if an error happens in Study.tell, this logging section
+        # should be reached. Because `frozen_trial` would not be assigned when
+        # Study.tell fails, it fetches the trial from the storage.
+        if tell_err is not None:
+            frozen_trial = study._storage.get_trial(trial._trial_id)
```
hvy, line 246: Not a strong opinion but would the code become cleaner/no need for descriptive comment if we do this trial query inside the except clause/indent?

optuna/study/_optimize.py
hvy: Not a strong opinion but would the code become cleaner/no need for descriptive comment if we do this trial fetching inside the except clause/indent?

optuna/study/study.py
hvy: Thanks. Sorry for my vague comment but I meant that we could omit the warning altogether, if we get rid of the early return statement (which you did).

optuna/multi_objective/trial.py
```diff
@@ -382,6 +378,9 @@ def __hash__(self) -> int:
 
     # TODO(ohta): Implement `__repr__` method.
 
+    def _get_values(self) -> List[Optional[float]]:
+        return self.values
```
hvy, line 382: Do you still think we should keep this method? I know it's private and a small detail.

optuna/multi_objective/trial.py
```diff
@@ -276,7 +272,7 @@ def __init__(self, n_objectives: int, trial: FrozenTrial):
         self.n_objectives = n_objectives
         self._trial = trial
 
-        self.values = tuple(trial.intermediate_values.get(i) for i in range(n_objectives))
+        self.values = [trial.intermediate_values.get(i) for i in range(n_objectives)]
```
hvy, line 275: Could you elaborate a bit on this change?

tests/multi_objective_tests/test_trial.py
```diff
@@ -46,12 +46,12 @@ def objective(
 
     trial = study.trials[0]
     assert trial.intermediate_values == {1: (1, 2, 3), 2: (10, 20, 30)}
-    assert trial.values == (100, 200, 300)
+    assert trial.values == [100, 200, 300]
```
hvy, line 49: Ditto regarding the question why this change is needed.

optuna/study/_optimize.py
himkt: So sorry, you're totally  right... I changed the type of `value_or_values` using `Sequence`.

optuna/multi_objective/trial.py
himkt: During try-and-error I found `values` (and `_get_values`) of `FrozenTrial` is a list of float. My approach was to add `_get_values` to `FrozenMultiObjectiveTrial` so I updated the type hint for `self.values` from tuple to list.

Now I changed the approach to use self.values directly (also related. https://github.com/optuna/optuna/pull/3144#discussion_r782098716) and reverted the change in 2371b45.

optuna/study/study.py
```diff
@@ -569,6 +633,9 @@ def df(x):
                 :class:`~optuna.trial.TrialState.PRUNED`.
                 ``trial`` is a trial number but no
                 trial exists with that number.
+
+        Returns:
+            A :class:`~optuna.trial.FrozenTrial` representing the resulting trial.
```
hvy, line 638: Should we guarantee that this is a (deep) copy, and that the caller is free to modify it? It's probably important for the internal usage of `Study.tell` as well.

himkt:
> Maybe we can move all this logic to Study.tell if that'd make code easier to maintain?

`func_err` is not accesible inside Study.tell so I think at least we have to leave a log logic for failed trial (with `func_err`). Now `_run_trial` emits logs for the execution result of Study.tell and `Study.tell` emits logs for something happens during the execution. I have no idea for improving readability, please share an idea if you have. :pray:

optuna/study/study.py
hvy: Thanks!

himkt:
https://github.com/optuna/optuna/compare/6571c9419fb4b65468bdeebdc59479a6154456d3...5d5fcad615970bda82dc23d46ef47b1b0e6cc85a?expand=1

These commits cleanup warning/logging and update tests.

tests/multi_objective_tests/test_study.py
```diff
@@ -171,21 +171,20 @@ def test_log_completed_trial_skip_storage_access() -> None:
...
+    study._study._storage.create_new_trial(study._study._study_id)
+    frozen_trial = study.trials[0]
     storage = study._storage
 
     with patch.object(storage, "get_trial", wraps=storage.get_trial) as mock_object:
-        optuna.multi_objective.study._log_completed_trial(study, trial, [1.0])
-        # Trial.params and MultiObjectiveTrial._get_values access storage.
-        assert mock_object.call_count == 2
+        optuna.multi_objective.study._log_completed_trial(study, frozen_trial)
+        assert mock_object.call_count == 0
```
himkt, line 180: https://github.com/optuna/optuna/pull/3144/files#diff-07b9f79024da1a09ad276d497bf17b34debf8821f87907777a5942935d1816a8R483-R492

`_log_completed_trial` doesn't access to the storage any more.

optuna/cli.py
```diff
@@ -837,7 +837,7 @@ def get_parser(self, prog_name: str) -> ArgumentParser:
         parser.add_argument("--study-name", type=str, help="Name of study.")
         parser.add_argument("--trial-number", type=int, help="Trial number.")
         parser.add_argument("--values", type=float, nargs="+", help="Objective values.")
-        parser.add_argument("--state", type=str, default="complete", help="Trial state.")
+        parser.add_argument("--state", type=Optional[str], help="Trial state.")
```
hvy, line 840: Arguments are optional by default if they're not explicitly defined with `required=True`.
```suggestion
        parser.add_argument("--state", type=str, help="Trial state.")
```

optuna/cli.py
```diff
@@ -861,7 +861,11 @@ def take_action(self, parsed_args: Namespace) -> int:
             study_name=parsed_args.study_name,
         )
 
-        state = TrialState[parsed_args.state.upper()]
+        if parsed_args.state is not None:
+            state: Optional[TrialState] = TrialState[parsed_args.state.upper()]
+        else:
+            state = None
```
hvy, line 867:
```suggestion
        if parsed_args.state is not None:
            state: Optional[TrialState] = TrialState[parsed_args.state.upper()]
```

optuna/study/_optimize.py
```diff
@@ -237,93 +222,37 @@ def _run_trial(
 
     # `Study.tell` may raise during trial post-processing.
     try:
-        study.tell(trial, values=values, state=state)
+        with warnings.catch_warnings(record=True) as tell_warnings:
```
hvy, line 225: Do we have to enforce warnings with `simplefilter("always", ...)` in addition to this line?

optuna/study/_optimize.py
```diff
@@ -237,93 +222,37 @@ def _run_trial(
...
                     ),
                     exc_info=func_err_fail_exc_info,
                 )
-            elif values_conversion_failure_message is not None:
-                _logger.warning(values_conversion_failure_message)
-            else:
-                assert False, "Should not reach."
+            if tell_warnings is not None:
+                for tell_warning in tell_warnings:
+                    _logger.warning(tell_warning.message)
```
hvy, line 245: This is not a strong request change but more of a verification, but I though that the formats of the "post-trial" log messages were going to be unified in `_run_trial` (the formats of the warnings emitted by `Study.tell` don't necessarily have to be aligned with those). With the current logic they're split between `_run_trial` and in `Study.tell`. Do we share the same understanding?

tests/study_tests/test_study.py
```diff
@@ -1069,39 +1069,61 @@ def test_tell_values() -> None:
...
+        assert study.trials[-1].state == TrialState.FAIL
+        assert study.trials[-1].values is None
 
     # Missing values for completions.
     with pytest.raises(ValueError):
         study.tell(study.ask(), state=TrialState.COMPLETE)
 
-    # Default state is `TrialState.COMPLETE` for which values are required.
-    with pytest.raises(ValueError):
+    # Either state or values is required
```
hvy, line 1121: Nit.
```suggestion
    # Either state or values is required.
```

optuna/cli.py
himkt: Sorry but please explain it in detail. 🙇
Would not `state` be unbounded if `parsed_args.state` None, and an error happen in Study.tell?

optuna/study/_optimize.py
himkt: Does the unification you mention looks like following?

```diff
diff --git a/optuna/study/study.py b/optuna/study/study.py
index be175e4b..6d9a4d82 100644
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -702,6 +702,7 @@ class Study:
 
         if state is None and values is None:
             warnings.warn(
+                f"Trial {frozen_trial.number} failed. "
                 "Study.tell is called with `values` and `state` set to None but "
                 "you have to specify either `values` or `state` at least. "
                 "`values` is required if a trial finishes successfully or "
@@ -730,7 +731,9 @@ class Study:
                 # This path is reached when the objective returns invalid data
                 # including NaN or non-numerical points during Study.optimize.
                 else:
-                    warnings.warn(values_conversion_failure_message)
+                    warnings.warn(
+                        f"Trial {frozen_trial.number} failed. {values_conversion_failure_message}"
+                    )
                     state = TrialState.FAIL
 
             # When called from `Study.optimize` and `state` is pruned, the given `values` contains
```

optuna/study/_optimize.py
himkt: 0d5a674 adds a filter.

optuna/study/_optimize.py
```diff
@@ -237,93 +222,38 @@ def _run_trial(
 
     # `Study.tell` may raise during trial post-processing.
     try:
-        study.tell(trial, values=values, state=state)
+        with warnings.catch_warnings(record=True) as tell_warnings:
+            warnings.simplefilter("always", UserWarning)
```
himkt, line 226:
```
Once the context manager exits, the warnings filter is restored to its state when the context was entered.
```
https://docs.python.org/3/library/warnings.html#testing-warnings

optuna/cli.py
hvy: Sorry you're right. Please ignore this comment. 🙇

optuna/study/_optimize.py
```diff
@@ -237,93 +222,38 @@ def _run_trial(
...
-                        trial.number, repr(func_err)
+                        frozen_trial.number, repr(func_err)
                     ),
                     exc_info=func_err_fail_exc_info,
                 )
-            elif values_conversion_failure_message is not None:
-                _logger.warning(values_conversion_failure_message)
-            else:
-                assert False, "Should not reach."
+            if tell_warnings is not None:
```
hvy, line 244: Are these two `if` statements ever both evaluated to true? I'm curious since the first one is generally when the objective value raises, and the latter if tell fails due to e.g. bad return values from the objective.

optuna/study/_optimize.py
hvy: Sorry if my comment was unclear. Not really, what I meant was the following. The trial logs follow a certain convention, i.e. "Trial <some number> .... ", although implicit. At the same time, the logs messages are created (defined in code) at different places with not so good code locality. What I'm suggesting is to improve the code locality. Again, it might be okay to leave it as future work.

optuna/study/_optimize.py
himkt: As you pointed out, these rules should not be evaluated to be `True`: the first happens when an exception is raised in `func()`, which makes `state` to be `PRUNED` or `FAIL` and `value_or_values` to be `None`. For the case, the latter is always invoked with `state` not to be `None` (and `skip_if_finished` to be `False`).

Do you think these two if rules should be unified to one if-elif block?

optuna/study/_optimize.py
himkt: Thank you so much for explaining your intention in detail. ❤️ It may make sense to me.
So... you think we can create a logging helper for failed trials looks like 9e48748?

hvy: Thanks a slot, that's more or less exactly what I meant!
I think we can follow this up in a separate ticket about further improving the locality, e.g. make them all Study methods, introducing a new private submodule, or something.

optuna/study/_optimize.py
hvy: > Do you think these two if rules should be unified to one if-elif block?

Not a strong opinion but preferably yes. One reasons is, the latter statement doesn't have to be evaluated, if the first one is true. It'll also improve code readability in my opinion.

himkt: 👍. 884bdd3 unify them.

himkt:
644db66 fixes a log output.

### Before

```
[W 2022-02-02 11:26:58,758] Trial 0 failed because of the following error: Trial 0 failed, because the objective function returned nan.
[W 2022-02-02 11:27:21,888] Trial 0 failed because of the following error: Trial 0 failed, because the number of the values 2 did not match the number of the objectives 1.
```

### After

```
[W 2022-02-02 11:28:05,588] Trial 0 failed because of the following error: The objective function returned nan.
[W 2022-02-02 11:27:45,545] Trial 0 failed because of the following error: The number of the values 2 did not match the number of the objectives 1.
```

himkt:
@hvy I revised the PR to fix logging format. PTAL. :pray:

HideakiImamura:
Thank you for the update. I have confirmed that the current code resolves the problem of #3205.

<details>
<summary>code</summary>

```python
import optuna
import numpy as np


# In-memory storage
study = optuna.create_study()
trial = study.ask()
try:
    study.tell(trial, float(np.nan))
except Exception as e:
    print(e)
finally:
    print(f"In-memory : Trial state is {study.trials[-1].state}")
print(study.trials[0])

# RDB storage (SQLite)
study = optuna.create_study(storage='sqlite:///../sqlite.db')
trial = study.ask()
try:
    study.tell(trial, float(np.nan))
except Exception as e:
    print(e)
finally:
    print(f"SQLite    : Trial state is {study.trials[-1].state}")
print(study.trials[0])

# RDB storage (MySQL)
study = optuna.create_study(storage="mysql+pymysql://root:test@localhost/mysql")
trial = study.ask()
try:
    study.tell(trial, float(np.nan))
except Exception as e:
    print(e)
finally:
    print(f"MySQL     : Trial state is {study.trials[-1].state}")
print(study.trials[0])

# RDB storage (PostgreSQL)
study = optuna.create_study(storage="postgresql+psycopg2://postgres:test@localhost:15432/postgres")
trial = study.ask()
try:
    study.tell(trial, float(np.nan))
except Exception as e:
    print(e)
finally:
    print(f"PostgreSQL: Trial state is {study.trials[-1].state}")
print(study.trials[0])
```

</details>

<details>
<summary>master</summary>

```
(venv) mamu@HideakinoMacBook-puro optuna % python ../code_snippets_for_optuna/3205-3206/nan-tell.py
[I 2022-02-02 16:10:28,827] A new study created in memory with name: no-name-341d29a5-33bd-49cb-b76c-0ad2b54a5f5b
Trial 0 failed, because the objective function returned nan.
In-memory : Trial state is TrialState.RUNNING
FrozenTrial(number=0, values=None, datetime_start=datetime.datetime(2022, 2, 2, 16, 10, 28, 827850), datetime_complete=None, params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=0, state=TrialState.RUNNING, value=None)
[I 2022-02-02 16:10:28,896] A new study created in RDB with name: no-name-ebce68bc-c5b4-4a67-86d3-22339d3fbeb2
Trial 0 failed, because the objective function returned nan.
SQLite    : Trial state is TrialState.RUNNING
FrozenTrial(number=0, values=None, datetime_start=datetime.datetime(2022, 2, 2, 16, 10, 28, 917253), datetime_complete=None, params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=3, state=TrialState.RUNNING, value=None)
[I 2022-02-02 16:10:29,074] A new study created in RDB with name: no-name-acb4a0bd-0f02-4679-9803-343396e010d5
Trial 0 failed, because the objective function returned nan.
MySQL     : Trial state is TrialState.RUNNING
FrozenTrial(number=0, values=None, datetime_start=datetime.datetime(2022, 2, 2, 16, 10, 29, 158211), datetime_complete=None, params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=27, state=TrialState.RUNNING, value=None)
[I 2022-02-02 16:10:29,408] A new study created in RDB with name: no-name-8dcaeeec-7a37-4c26-84f6-4712a42f54d6
Trial 0 failed, because the objective function returned nan.
PostgreSQL: Trial state is TrialState.RUNNING
FrozenTrial(number=0, values=None, datetime_start=datetime.datetime(2022, 2, 2, 16, 10, 29, 457484), datetime_complete=None, params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=6, state=TrialState.RUNNING, value=None)
```

</details>

<details>
<summary>This PR</summary>

```
(venv) mamu@HideakinoMacBook-puro optuna % python ../code_snippets_for_optuna/3205-3206/nan-tell.py
[I 2022-02-02 16:14:28,964] A new study created in memory with name: no-name-fdac712d-4f1f-4e49-88f7-241a368c52b3
/Users/mamu/Dev/pfn/optuna/optuna/study/study.py:727: UserWarning: The objective function returned nan.
  warnings.warn(values_conversion_failure_message)
In-memory : Trial state is TrialState.FAIL
FrozenTrial(number=0, values=None, datetime_start=datetime.datetime(2022, 2, 2, 16, 14, 28, 965590), datetime_complete=datetime.datetime(2022, 2, 2, 16, 14, 28, 966415), params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=0, state=TrialState.FAIL, value=None)
[I 2022-02-02 16:14:29,047] A new study created in RDB with name: no-name-9aefc644-f205-4e49-97bd-d8d6e459c8b4
SQLite    : Trial state is TrialState.FAIL
FrozenTrial(number=0, values=None, datetime_start=datetime.datetime(2022, 2, 2, 16, 14, 29, 70134), datetime_complete=datetime.datetime(2022, 2, 2, 16, 14, 29, 79736), params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=4, state=TrialState.FAIL, value=None)
[I 2022-02-02 16:14:29,268] A new study created in RDB with name: no-name-93f53a78-a4d8-420e-b50f-e7ed203044f3
MySQL     : Trial state is TrialState.FAIL
FrozenTrial(number=0, values=None, datetime_start=datetime.datetime(2022, 2, 2, 16, 14, 29, 330636), datetime_complete=datetime.datetime(2022, 2, 2, 16, 14, 29), params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=28, state=TrialState.FAIL, value=None)
[I 2022-02-02 16:14:29,687] A new study created in RDB with name: no-name-744b4fd2-07ff-4f04-a963-f3aca4f36abf
PostgreSQL: Trial state is TrialState.FAIL
FrozenTrial(number=0, values=None, datetime_start=datetime.datetime(2022, 2, 2, 16, 14, 29, 735285), datetime_complete=datetime.datetime(2022, 2, 2, 16, 14, 29, 749762), params={}, distributions={}, user_attrs={}, system_attrs={}, intermediate_values={}, trial_id=7, state=TrialState.FAIL, value=None)
```

</details>

hvy:
Thanks for the updates and sorry for the delayed reply. Just ran some quick checks and looks good, let me have a look at the code this week. 

<details>
<summary>Check code</summary>

```python
import optuna as ot


def print_study(s):
    for t in s.trials:
        print(t.state, t.values)
    print()


if __name__ == "__main__":
    print("inf")
    s = ot.create_study()
    s.optimize(lambda _: float("inf"), n_trials=1)
    print_study(s)

    print("nan")
    s = ot.create_study()
    s.optimize(lambda _: float("nan"), n_trials=1)
    print_study(s)

    print("bad type")
    s = ot.create_study()
    s.optimize(lambda _: object(), n_trials=1)
    print_study(s)

    print("pruned")
    s = ot.create_study()
    def _f(t):
        raise ot.TrialPruned
    s.optimize(_f, n_trials=1)
    print_study(s)

    print("pruned, intermediate value")
    s = ot.create_study()
    def _f(t):
        t.report(step=1, value=1)
        raise ot.TrialPruned
    s.optimize(_f, n_trials=1)
    print_study(s)

    print("failed")
    s = ot.create_study()
    def _f(t):
        raise ValueError("Oh my error.")
    s.optimize(_f, n_trials=1, catch=(ValueError,))
    print_study(s)

    print("wrong n values")
    s = ot.create_study(directions=["minimize", "minimize"])
    s.optimize(lambda _: (1,), n_trials=1)
    print_study(s)

    print("ask, inf")
    s = ot.create_study()
    t = s.ask()
    s.tell(t, values=float("inf"))
    print_study(s)

    print("ask, nan")
    s = ot.create_study()
    t = s.ask()
    s.tell(t, values=float("nan"))
    print_study(s)

    print("ask, bad type")
    s = ot.create_study()
    t = s.ask()
    s.tell(t, values=object())
    print_study(s)

    print("ask, pruned")
    s = ot.create_study()
    t = s.ask()
    s.tell(t, state=ot.trial.TrialState.PRUNED)
    print_study(s)

    print("ask, pruned, intermediate value")
    s = ot.create_study()
    t = s.ask()
    t.report(step=1, value=1)
    s.tell(t, state=ot.trial.TrialState.PRUNED)
    print_study(s)

    print("ask, failed")
    s = ot.create_study()
    t = s.ask()
    s.tell(t, state=ot.trial.TrialState.FAIL)
    print_study(s)

    print("ask, skip if failed")
    s = ot.create_study()
    t = s.ask()
    s.tell(t, values=(1,))
    s.tell(t, values=(1,), skip_if_finished=True)
    print_study(s)
```

</details>

<details>
<summary>Results</summary>

```python
inf
[I 2022-02-07 23:11:18,853] A new study created in memory with name: no-name-c8eb1963-b82a-41cb-bec3-86b97f363584
[I 2022-02-07 23:11:18,853] Trial 0 finished with value: inf and parameters: {}. Best is trial 0 with value: inf.
TrialState.COMPLETE [inf]

nan
[I 2022-02-07 23:11:18,854] A new study created in memory with name: no-name-406a7147-c7c3-4901-bc7c-f51603a22f45
[W 2022-02-07 23:11:18,854] Trial 0 failed because of the following error: The objective function returned nan.
TrialState.FAIL None

bad type
[I 2022-02-07 23:11:18,854] A new study created in memory with name: no-name-db7da46e-3f5b-40ec-844f-9f936baaaa88
[W 2022-02-07 23:11:18,855] Trial 0 failed because of the following error: The value <object object at 0x120dff6d0> could not be  cast to float.
TrialState.FAIL None

pruned
[I 2022-02-07 23:11:18,855] A new study created in memory with name: no-name-07e9d812-5f8e-43a7-b169-481a216dbb85
[I 2022-02-07 23:11:18,855] Trial 0 pruned.
TrialState.PRUNED None

pruned, intermediate value
[I 2022-02-07 23:11:18,855] A new study created in memory with name: no-name-a59e3046-3b43-44fe-b2da-132031117dc2
[I 2022-02-07 23:11:18,856] Trial 0 pruned.
TrialState.PRUNED [1.0]

failed
[I 2022-02-07 23:11:18,856] A new study created in memory with name: no-name-e1dc3f41-c057-4813-a801-99d66a7eaa20
[W 2022-02-07 23:11:18,857] Trial 0 failed because of the following error: ValueError('Oh my error.')
  File "/Users/hvy/repos/optuna/optuna/study/_optimize.py", line 207, in _run_trial
    value_or_values = func(trial)
  File "pr_3144.py", line 47, in _f
    raise ValueError("Oh my error.")
ValueError: Oh my error.
TrialState.FAIL None

wrong n values
[I 2022-02-07 23:11:18,858] A new study created in memory with name: no-name-ffddcacc-a2ed-4a26-a8c7-4cf35a4007fb
[W 2022-02-07 23:11:18,858] Trial 0 failed because of the following error: The number of the values 1 did not match the number of the objectives 2.
TrialState.FAIL None

ask, inf
[I 2022-02-07 23:11:18,859] A new study created in memory with name: no-name-17bceeca-efb2-41fa-bf26-aac548b3908e
TrialState.COMPLETE [inf]

ask, nan
[I 2022-02-07 23:11:18,859] A new study created in memory with name: no-name-14915135-5476-4a01-b7a7-d6453089a1ac
/Users/hvy/repos/optuna/optuna/study/study.py:727: UserWarning: The objective function returned nan.
  warnings.warn(values_conversion_failure_message)
TrialState.FAIL None

ask, bad type
[I 2022-02-07 23:11:18,860] A new study created in memory with name: no-name-14074caa-6b38-4db8-9540-f2dcdd17c970
/Users/hvy/repos/optuna/optuna/study/study.py:727: UserWarning: The value <object object at 0x120dffaa0> could not be  cast to float.
  warnings.warn(values_conversion_failure_message)
TrialState.FAIL None

ask, pruned
[I 2022-02-07 23:11:18,861] A new study created in memory with name: no-name-4b79dff4-d0d0-464b-9437-2d8bdadb1ea1
TrialState.PRUNED None

ask, pruned, intermediate value
[I 2022-02-07 23:11:18,862] A new study created in memory with name: no-name-ba165729-d218-46f2-9c60-2d3b0c68cfbf
TrialState.PRUNED [1.0]

ask, failed
[I 2022-02-07 23:11:18,862] A new study created in memory with name: no-name-28047093-9806-4a9c-99bf-fa8dc25422bb
TrialState.FAIL None

ask, skip if failed
[I 2022-02-07 23:11:18,863] A new study created in memory with name: no-name-8fb308b9-90ca-4952-ae8d-eaa56543b106
[I 2022-02-07 23:11:18,863] Skipped telling trial 0 with values (1,) and state None since trial was already finished. Finished trial has values [1.0] and state TrialState.COMPLETE.
TrialState.COMPLETE [1.0]
```

</details>

optuna/study/_optimize.py
hvy: Should we try not to rely on this logic given https://bugs.python.org/issue29672 ?

himkt:
The mypy failure in d3dde71 makes me realize I had unintentionally changed the type what `_run_trial` returns.
Should I revert it? The code seemingly get frozen_trial for the given trial so I'm wondering if we can update to directly return frozen_trial in `_run_trial` and use it in `_optimize_sequential`.

https://github.com/optuna/optuna/blob/002d1e16d54ea7ee013cad291a1fd8e26ef42285/optuna/study/_optimize.py#L174

himkt:
I prototyped to change `_run_trial` based on https://github.com/optuna/optuna/pull/3144#issuecomment-1033896981.

himkt:
@hvy @HideakiImamura I revised the PR (https://github.com/optuna/optuna/compare/e9b044d..cefd65c) to work on:

- remove `catch_warnings`, which is based on https://github.com/optuna/optuna/pull/3301
- change the return type from `Trial` to `FrozenTrial` in `_run_trial`

optuna/study/study.py
```diff
@@ -482,9 +537,10 @@ def tell(
         self,
         trial: Union[trial_module.Trial, int],
         values: Optional[Union[float, Sequence[float]]] = None,
-        state: TrialState = TrialState.COMPLETE,
+        state: Optional[TrialState] = None,
         skip_if_finished: bool = False,
-    ) -> None:
+        suppress_warning: bool = False,
```
hvy, line 542: This is quite convenient. Also noticed you've added a test 👍  I'm wondering if we really want to expose this to the user however. It communicates complexity to an already somewhat complex method. Did you consider controlling this behavior by some internal method, e.g. `_tell` (although it's currently "used").

hvy (commented):
Thanks for the swift updates, great that we could get rid of the `catch_warnings`. I left some follow up questions if you could take a look. 🙇

optuna/study/study.py
```diff
@@ -664,6 +761,11 @@ def df(x):
 
             self._storage.set_trial_state(trial_id, state)
 
+        frozen_trial = copy.deepcopy(self._storage.get_trial(trial_id))
+        if warning_message is not None:
+            frozen_trial.set_user_attr("warning_message", warning_message)
```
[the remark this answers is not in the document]
himkt: You're completely right. :bow: I revised the PR to use system attributes in 2d7ae11. 🙋

optuna/study/study.py
himkt: It makes sense. I "restored" `_tell` (removing `@deprecated`) and moved the logic from `tell` to `_tell`.
https://github.com/optuna/optuna/compare/2d7ae11ef...b1b870f

hvy: Thanks, I'll have a look. 🙇

optuna/study/_optimize.py
himkt: I missed your comment. :bow: Yes, I think so.

himkt:
Let me kindly ping @hvy. :pray:

optuna/study/_optimize.py
```diff
@@ -217,111 +211,42 @@ def _run_trial(
...
         assert stop_event is not None
         assert thread is not None
         stop_event.set()
         thread.join()
 
-    # `Study.tell` may raise during trial post-processing.
+    # `Study._tell` may raise during trial post-processing.
     try:
-        study.tell(trial, values=values, state=state)
+        # Note: call `Study._tell` instead of `Study.tell` to suppress warning.
```
hvy, line 223:
```suggestion
        # Note: call `Study._tell_with_warning` instead of `Study.tell` to suppress warning.
```

hvy, line 223: Or, we can omit this line altogether. Code is self explanatory.
```suggestion
```

optuna/study/_optimize.py
```diff
@@ -217,111 +211,42 @@ def _run_trial(
...
-            state = TrialState.COMPLETE
 
     if study._storage.is_heartbeat_enabled():
         assert stop_event is not None
         assert thread is not None
         stop_event.set()
         thread.join()
 
-    # `Study.tell` may raise during trial post-processing.
+    # `Study._tell` may raise during trial post-processing.
```
hvy, line 221:
```suggestion
    # `Study._tell_with_warning` may raise during trial post-processing.
```

optuna/study/_optimize.py
```diff
@@ -217,111 +211,42 @@ def _run_trial(
...
-                    ),
-                    exc_info=func_err_fail_exc_info,
-                )
-            elif values_conversion_failure_message is not None:
-                _logger.warning(values_conversion_failure_message)
-            else:
-                assert False, "Should not reach."
+                _log_failed_trial(frozen_trial, repr(func_err), exc_info=func_err_fail_exc_info)
+            elif "study_tell_warning" in frozen_trial.system_attrs:
+                _log_failed_trial(frozen_trial, frozen_trial.system_attrs["study_tell_warning"])
```
hvy, line 239: Is an "else" meant to be logically possible here, i.e. a failed trial without any following warning log? If not, how about an `else: assert False, "Should not reach."` after these lines? On the other hand, if it's possible, a comment would be informative.

optuna/study/study.py
```diff
@@ -45,6 +45,61 @@
...
+    value = None
+    failure_message = None
+
+    try:
+        value = float(original_value)
+    except (
+        ValueError,
+        TypeError,
+    ):
+        failure_message = f"The value {repr(original_value)} could not be  cast to float."
```
hvy, line 94: Perhaps not introduced by this PR? 🤔 
```suggestion
        failure_message = f"The value {repr(original_value)} could not be cast to float."
```

optuna/study/study.py
```diff
@@ -1018,23 +992,169 @@ def _tell(
...
+            raise
+        finally:
+            if values is not None:
+                self._storage.set_trial_values(trial_id, values)
+
+            self._storage.set_trial_state(trial_id, state)
+
+        frozen_trial = copy.deepcopy(self._storage.get_trial(trial_id))
+        if warning_message is not None:
+            frozen_trial.set_system_attr("study_tell_warning", warning_message)
```
hvy, line 1139: Ah, so it it's not persisted in the storage. Not what I first expected but sounds fair.

hvy (commented):
Thanks for the updates again and sorry for the delayed review. I left some comments if you could take a look.

optuna/study/study.py
himkt: ✂️

optuna/study/_optimize.py
himkt: It is not expected that a trial fails without func error or warning message (from Study.tell). I added "else" in a21c7be. Thank you for the advise!

optuna/study/_optimize.py
himkt: > Code is self explanatory.

Agree after your comment. I removed the comment.

himkt:
Just rebased. Let me kindly ping (no rush). @hvy @HideakiImamura

HideakiImamura (commented):
Thanks for the update and sorry for the late reply. I have a comment. PTAL.

optuna/study/study.py
```diff
@@ -1019,23 +993,169 @@ def _tell(
     ) -> None:
         self.tell(trial, values, state)
 
-    def _log_completed_trial(self, trial: trial_module.Trial, values: Sequence[float]) -> None:
+    def _tell_with_warning(
```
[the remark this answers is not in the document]
himkt: Thank you, I revised the PR to extract `_tell_with_warning` into `_tell.py`.
https://github.com/optuna/optuna/compare/3ee7258...6a52b35

HideakiImamura (dismissed):
Thanks for the update! LGTM!

tests/multi_objective_tests/test_study.py
hvy: `test_log_completed_trial_skip_storage_access` can be removed in my opinion. It makes a lot less sense now without contextual background to readers. Not a strong opinion, though.

optuna/study/_tell.py
```diff
@@ -0,0 +1,221 @@
...
+        return copy.deepcopy(frozen_trial)
+
+    if state is None and values is None:
+        state = TrialState.FAIL
+        message_values_and_state_not_specified = (
+            "Study.tell is called with `values` and `state` set to None but "
+            "you have to specify either `values` or `state` at least. "
+            "`values` is required if a trial finishes successfully or "
+            "`state` is required if a trial is fails or pruned."
+        )
```
hvy, line 165: Not sure how to fix it but this message is displayed to users calling `Study.optimize` returning `None` from their objective functions, which might be less intuitive given the message is specific to `Study.tell`. In `master`, the message in this case is `Trial 0 failed, because the value None could not be cast to float.`. I think we want to keep this behavior. What do you think?

himkt: Thank you for your consideration, your feedback makes sense. I'll try to revise my PR.
I'm considering that it changes the error message on the value of `suppress_warning` assuming it be `True` only when `_tell_with_warning` is invoked by `study._run_trial`.

himkt: @hvy 
I tried to keep the current behavior as much as possible. b2aff2b drops a `raise` and checks a value even if it is `None`. After this change, `Study.optimize` make a trial whose value is `None` `TrialState.FAIL` with the message `Trial 0 failed, because the value None could not be cast to float.`. Although this patch changes the behavior when `Study.tell` is called directly, I think it doesn't change in the perspective of `Study.optimize` as `Study.optimize` already checked a value in `_run_trial` in the existing implementation. I'd like to hear your feedback again.

himkt:
from platform import python_branch


The output of the code provided in https://github.com/optuna/optuna/pull/3144#issuecomment-1031515142:

code:

<details>

```python
import optuna as ot

def print_study(s):
    for t in s.trials:
        print(t.state, t.values)
    print()

if __name__ == "__main__":
    print("inf")
    s = ot.create_study()
    s.optimize(lambda _: float("inf"), n_trials=1)
    print_study(s)

    print("nan")
    s = ot.create_study()
    s.optimize(lambda _: float("nan"), n_trials=1)
    print_study(s)

    print("None")
    s = ot.create_study()
    s.optimize(lambda _: None, n_trials=1)
    print_study(s)

    print("bad type")
    s = ot.create_study()
    s.optimize(lambda _: object(), n_trials=1)
    print_study(s)

    print("pruned")
    s = ot.create_study()
    def _f(t):
        raise ot.TrialPruned
    s.optimize(_f, n_trials=1)
    print_study(s)

    print("pruned, intermediate value")
    s = ot.create_study()
    def _f(t):
        t.report(step=1, value=1)
        raise ot.TrialPruned
    s.optimize(_f, n_trials=1)
    print_study(s)

    print("pruned, bad intermediate value")
    s = ot.create_study()
    def _f(t):
        t.report(step=1, value=float("nan"))
        raise ot.TrialPruned
    s.optimize(_f, n_trials=1)
    print_study(s)

    print("failed")
    s = ot.create_study()
    def _f(t):
        raise ValueError("Oh my error.")
    s.optimize(_f, n_trials=1, catch=(ValueError,))
    print_study(s)

    print("wrong n values")
    s = ot.create_study(directions=["minimize", "minimize"])
    s.optimize(lambda _: (1,), n_trials=1)
    print_study(s)

    print("ask, inf")
    s = ot.create_study()
    t = s.ask()
    s.tell(t, values=float("inf"))
    print_study(s)

    print("ask, pruned")
    s = ot.create_study()
    t = s.ask()
    s.tell(t, state=ot.trial.TrialState.PRUNED)
    print_study(s)

    print("ask, pruned, intermediate value")
    s = ot.create_study()
    t = s.ask()
    t.report(step=1, value=1)
    s.tell(t, state=ot.trial.TrialState.PRUNED)
    print_study(s)

    print("ask, pruned, bad intermediate value")
    s = ot.create_study()
    t = s.ask()
    t.report(step=1, value=float("nan"))
    s.tell(t, state=ot.trial.TrialState.PRUNED)
    print_study(s)

    print("ask, failed")
    s = ot.create_study()
    t = s.ask()
    s.tell(t, state=ot.trial.TrialState.FAIL)
    print_study(s)

    print("ask, skip if finished")
    s = ot.create_study()
    t = s.ask()
    s.tell(t, values=(1,))
    s.tell(t, values=(1,), skip_if_finished=True)
    print_study(s)

    print("ask, wrong n values")
    try:
        s = ot.create_study(directions=["minimize", "minimize"])
        t = s.ask()
        s.tell(t, values=(1,))
        print_study(s)
    except Exception as e:
        print("== stopped with exception ==")
        print(e)
        print()

    print("ask, nan")
    try:
        s = ot.create_study()
        t = s.ask()
        s.tell(t, values=float("nan"))
        print_study(s)
    except Exception as e:
        print("== stopped with exception ==")
        print(e)
        print()

    print("ask, None")
    try:
        s = ot.create_study()
        t = s.ask()
        s.tell(t, values=None)
        print_study(s)
    except Exception as e:
        print("== stopped with exception ==")
        print(e)
        print()

    print("ask, bad type")
    try:
        s = ot.create_study()
        t = s.ask()
        s.tell(t, values=object())
        print_study(s)
    except Exception as e:
        print("== stopped with exception ==")
        print(e)
        print()
```

</details>

- master:

<details>

```
> python compat.py                                                                                                                                  2022-03-29 16:07:01
inf
[I 2022-03-29 16:07:03,763] A new study created in memory with name: no-name-f17ae8bd-9587-4038-af1b-6c8674e3335a
[I 2022-03-29 16:07:03,764] Trial 0 finished with value: inf and parameters: {}. Best is trial 0 with value: inf.
TrialState.COMPLETE [inf]

nan
[I 2022-03-29 16:07:03,764] A new study created in memory with name: no-name-47b62ab6-e1f4-478f-8557-58f131a3f6f4
[W 2022-03-29 16:07:03,765] Trial 0 failed, because the objective function returned nan.
TrialState.FAIL None

None
[I 2022-03-29 16:07:03,765] A new study created in memory with name: no-name-327ed247-aa8e-4bb7-9a10-96233854a449
[W 2022-03-29 16:07:03,765] Trial 0 failed, because the value None could not be cast to float.
TrialState.FAIL None

bad type
[I 2022-03-29 16:07:03,765] A new study created in memory with name: no-name-920802e4-4a42-4184-addf-f6ea1a4e72bc
[W 2022-03-29 16:07:03,765] Trial 0 failed, because the value <object object at 0x148105370> could not be cast to float.
TrialState.FAIL None

pruned
[I 2022-03-29 16:07:03,766] A new study created in memory with name: no-name-33f0d958-531a-4e7d-b19d-7c71b531f152
[I 2022-03-29 16:07:03,766] Trial 0 pruned.
TrialState.PRUNED None

pruned, intermediate value
[I 2022-03-29 16:07:03,766] A new study created in memory with name: no-name-06559662-cc49-486c-8433-69c07436ad51
[I 2022-03-29 16:07:03,766] Trial 0 pruned.
TrialState.PRUNED [1.0]

pruned, bad intermediate value
[I 2022-03-29 16:07:03,767] A new study created in memory with name: no-name-91f56f8a-90eb-4675-afe3-2430da739a78
[I 2022-03-29 16:07:03,767] Trial 0 pruned.
TrialState.PRUNED None

failed
[I 2022-03-29 16:07:03,767] A new study created in memory with name: no-name-21642c44-499e-4aa4-8194-a6a9ad7c98bf
[W 2022-03-29 16:07:03,767] Trial 0 failed because of the following error: ValueError('Oh my error.')
  File "/Users/himkt/work/github.com/himkt/optuna/optuna/study/_optimize.py", line 211, in _run_trial
    value_or_values = func(trial)
  File "/Users/himkt/work/github.com/himkt/optuna/compat.py", line 55, in _f
    raise ValueError("Oh my error.")
ValueError: Oh my error.
TrialState.FAIL None

wrong n values
[I 2022-03-29 16:07:03,768] A new study created in memory with name: no-name-859eaa06-9128-41b2-906f-c580e4a14cbf
[W 2022-03-29 16:07:03,768] Trial 0 failed, because the number of the values 1 did not match the number of the objectives 2.
TrialState.FAIL None

ask, inf
[I 2022-03-29 16:07:03,768] A new study created in memory with name: no-name-950dadc8-2afa-41e9-a845-313af5103330
TrialState.COMPLETE [inf]

ask, pruned
[I 2022-03-29 16:07:03,769] A new study created in memory with name: no-name-3d4a5e7b-7bff-435d-8778-64f60015b944
TrialState.PRUNED None

ask, pruned, intermediate value
[I 2022-03-29 16:07:03,769] A new study created in memory with name: no-name-9580d088-69dd-4392-8d4c-d9353878fba6
TrialState.PRUNED [1.0]

ask, pruned, bad intermediate value
[I 2022-03-29 16:07:03,769] A new study created in memory with name: no-name-814a95a3-194b-47f6-b392-0d35785a5427
TrialState.PRUNED None

ask, failed
[I 2022-03-29 16:07:03,770] A new study created in memory with name: no-name-d98f3c94-a27c-4e1b-8f31-737c35083bf1
TrialState.FAIL None

ask, skip if finished
[I 2022-03-29 16:07:03,770] A new study created in memory with name: no-name-bcfa249b-50f7-4654-8cab-49b7cfeaa474
[I 2022-03-29 16:07:03,771] Skipped telling trial 0 with values (1,) and state TrialState.COMPLETE since trial was already finished. Finished trial has values [1.0] and state TrialState.COMPLETE.
TrialState.COMPLETE [1.0]

ask, wrong n values
[I 2022-03-29 16:07:03,771] A new study created in memory with name: no-name-3a65581a-4c59-4959-9ad0-53db92fbdbaa
== stopped with exception ==
Trial 0 failed, because the number of the values 1 did not match the number of the objectives 2.

ask, nan
[I 2022-03-29 16:07:03,771] A new study created in memory with name: no-name-a208474f-fd2d-4be5-9dd2-8ade38858a1d
== stopped with exception ==
Trial 0 failed, because the objective function returned nan.

ask, None
[I 2022-03-29 16:07:03,771] A new study created in memory with name: no-name-1da686dd-eaef-456e-97d5-0c835a8c6bcf
== stopped with exception ==
No values were told. Values are required when state is TrialState.COMPLETE.

ask, bad type
[I 2022-03-29 16:07:03,772] A new study created in memory with name: no-name-60f68a51-4366-4ea8-9f5d-341d76a59041
== stopped with exception ==
Trial 0 failed, because the value <object object at 0x1481053f0> could not be cast to float.
```

</details>


- this PR:

<details>

```
> python compat.py                                                                                                                                  2022-03-29 16:08:36
inf
[I 2022-03-29 16:08:39,583] A new study created in memory with name: no-name-b74514f3-c5d0-4cb2-8bf7-e3e9dee25a4b
[I 2022-03-29 16:08:39,584] Trial 0 finished with value: inf and parameters: {}. Best is trial 0 with value: inf.
TrialState.COMPLETE [inf]

nan
[I 2022-03-29 16:08:39,584] A new study created in memory with name: no-name-15d9b03c-c4a5-4ad7-9480-c29cc2833c40
[W 2022-03-29 16:08:39,585] Trial 0 failed because of the following error: The objective function returned nan.
TrialState.FAIL None

None
[I 2022-03-29 16:08:39,585] A new study created in memory with name: no-name-7900b634-5411-45c9-8ff2-aa867ab0f42a
[W 2022-03-29 16:08:39,585] Trial 0 failed because of the following error: The value None could not be cast to float.
TrialState.FAIL None

bad type
[I 2022-03-29 16:08:39,585] A new study created in memory with name: no-name-e0d263f8-e0bb-4e0f-98fb-c1ca770b08aa
[W 2022-03-29 16:08:39,586] Trial 0 failed because of the following error: The value <object object at 0x146765370> could not be cast to float.
TrialState.FAIL None

pruned
[I 2022-03-29 16:08:39,586] A new study created in memory with name: no-name-3a3c8896-e8e1-494d-8715-5643e0bce6b3
[I 2022-03-29 16:08:39,586] Trial 0 pruned.
TrialState.PRUNED None

pruned, intermediate value
[I 2022-03-29 16:08:39,586] A new study created in memory with name: no-name-f1e73178-225b-42b8-b660-e2deed256682
[I 2022-03-29 16:08:39,587] Trial 0 pruned.
TrialState.PRUNED [1.0]

pruned, bad intermediate value
[I 2022-03-29 16:08:39,587] A new study created in memory with name: no-name-73828a19-6911-43d5-b6e3-59a344fd1c80
[I 2022-03-29 16:08:39,587] Trial 0 pruned.
TrialState.PRUNED None

failed
[I 2022-03-29 16:08:39,587] A new study created in memory with name: no-name-d5d8e3f9-7cec-4e4a-8997-b6344b6851bc
[W 2022-03-29 16:08:39,588] Trial 0 failed because of the following error: ValueError('Oh my error.')
  File "/Users/himkt/work/github.com/himkt/optuna/optuna/study/_optimize.py", line 206, in _run_trial
    value_or_values = func(trial)
  File "/Users/himkt/work/github.com/himkt/optuna/compat.py", line 55, in _f
    raise ValueError("Oh my error.")
ValueError: Oh my error.
TrialState.FAIL None

wrong n values
[I 2022-03-29 16:08:39,588] A new study created in memory with name: no-name-b7d477b6-6f44-48e9-b40c-d151374dad6d
[W 2022-03-29 16:08:39,589] Trial 0 failed because of the following error: The number of the values 1 did not match the number of the objectives 2.
TrialState.FAIL None

ask, inf
[I 2022-03-29 16:08:39,589] A new study created in memory with name: no-name-6fca8a55-4c5a-4c8b-a9b1-0a122158ed9f
TrialState.COMPLETE [inf]

ask, pruned
[I 2022-03-29 16:08:39,589] A new study created in memory with name: no-name-50ab8eb1-b451-4340-9f83-b94eb38875bd
TrialState.PRUNED None

ask, pruned, intermediate value
[I 2022-03-29 16:08:39,589] A new study created in memory with name: no-name-1c4f0c4d-a477-4dc9-99b7-911fa21f9276
TrialState.PRUNED [1.0]

ask, pruned, bad intermediate value
[I 2022-03-29 16:08:39,590] A new study created in memory with name: no-name-087418c4-2a9d-461d-99b7-9eddb20b6316
TrialState.PRUNED None

ask, failed
[I 2022-03-29 16:08:39,590] A new study created in memory with name: no-name-6dc76d0a-4e8e-475f-895b-31f970288072
TrialState.FAIL None

ask, skip if finished
[I 2022-03-29 16:08:39,591] A new study created in memory with name: no-name-6951391f-99ea-4839-8c28-f1ae62236b61
[I 2022-03-29 16:08:39,591] Skipped telling trial 0 with values (1,) and state None since trial was already finished. Finished trial has values [1.0] and state TrialState.COMPLETE.
TrialState.COMPLETE [1.0]

ask, wrong n values
[I 2022-03-29 16:08:39,591] A new study created in memory with name: no-name-42d3b4f3-6747-48a8-b423-32fe11410347
/Users/himkt/work/github.com/himkt/optuna/optuna/study/_tell.py:177: UserWarning: The number of the values 1 did not match the number of the objectives 2.
  warnings.warn(values_conversion_failure_message)
TrialState.FAIL None

ask, nan
[I 2022-03-29 16:08:39,592] A new study created in memory with name: no-name-157b14b9-27e0-4743-a815-2c0a1e0a76be
/Users/himkt/work/github.com/himkt/optuna/optuna/study/_tell.py:177: UserWarning: The objective function returned nan.
  warnings.warn(values_conversion_failure_message)
TrialState.FAIL None

ask, None
[I 2022-03-29 16:08:39,593] A new study created in memory with name: no-name-ca8c50bb-8a79-43dd-b839-abc83174c3a8
/Users/himkt/work/github.com/himkt/optuna/optuna/study/_tell.py:177: UserWarning: The value None could not be cast to float.
  warnings.warn(values_conversion_failure_message)
TrialState.FAIL None

ask, bad type
[I 2022-03-29 16:08:39,593] A new study created in memory with name: no-name-94e558b9-4303-4494-a55a-e9d61e1d36de
/Users/himkt/work/github.com/himkt/optuna/optuna/study/_tell.py:177: UserWarning: The value <object object at 0x1467653f0> could not be cast to float.
  warnings.warn(values_conversion_failure_message)
TrialState.FAIL None
```

</details>

HideakiImamura:
Thanks for the update. The current behavior looks good to me. This is to be expected as this PR appears to have solved the problem with #3205, am I correct?

himkt:
I went through #3205.

I assume that `the problem with https://github.com/optuna/optuna/issues/3205` refers to:

```
If nan is passed in study.tell, mark the trial as FAILED regardless of the type of storage used. 
https://github.com/optuna/optuna/pull/3144
```

Is it right? If so, yes. The problem will be fixed in this PR.

himkt:
@hvy @HideakiImamura I update the test although diff becomes relatively huge... https://github.com/optuna/optuna/compare/e8beceb7c654eaa2e7dc08213c543c037467a4db..ff57599b055f5c6e363bd6a5aa56cb5c7277ce03?expand=1

Changes are divided to:
- create `test_tell.py` and move tests in `test_study.py` that tests behaviors of `Study.tell` to it
- move tests in `test_study.py` that tests behaviors of `Study.optimize` to `test_optimize.py`
- add cases that were provided in https://github.com/optuna/optuna/pull/3144#issuecomment-1031515142

himkt:
I talked with @HideakiImamura and decided to stop moving existing tests to `test_study.py`. There is need to discuss where to put tests for methods in Study (e.g. Study.optimize, Study.tell, ...) and it should be discussed in the another place (this PR is already huge).

himkt:
I talked with @HideakiImamura and decided to stop moving existing tests to `test_study.py`. There is need to discuss where to put tests for methods in Study (e.g. Study.optimize, Study.tell, ...) and it should be discussed in the another place (this PR is already huge). It is done in 474cae8 and bc2b690.

tests/study_tests/test_optimize.py
```diff
@@ -1,49 +1,54 @@
...
+        study = create_study(storage=storage)
+        with pytest.raises(ValueError):
+            _optimize._run_trial(study, func_value_error, ())
 
-        # Test trial with invalid objective value: nan
-        def func_nan(_: Trial) -> float:
 
-            return float("nan")
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES)
+def test_optimize_invoke_tell_with_suppressing_warning(storage_mode: str) -> None:
```
HideakiImamura, line 37: To match the other function names.
```suggestion
def test_run_trial_invoke_tell_with_suppressing_warning(storage_mode: str) -> None:
```

tests/study_tests/test_study.py
```diff
@@ -912,6 +913,66 @@ def objective(trial: Trial) -> float:
     assert states == []
 
 
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES)
+def test_run_trial(storage_mode: str) -> None:
```
HideakiImamura, line 917: It looks like this test should be placed in `test_optimize.py` since this does not call `Study.optimize` but `_run_trial`.

tests/study_tests/test_study.py
```diff
@@ -912,6 +913,66 @@ def objective(trial: Trial) -> float:
...
+    with StorageSupplier(storage_mode) as storage:
+        study = create_study(storage=storage)
+
+        frozen_trial = _optimize._run_trial(study, lambda _: float("inf"), catch=())
+        assert frozen_trial.state == TrialState.COMPLETE
+        assert frozen_trial.value == float("inf")
+
+
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES)
+def test_run_trial_automatically_fail(storage_mode: str) -> None:
```
HideakiImamura, line 928: Ditto.

tests/study_tests/test_study.py
```diff
@@ -912,6 +913,66 @@ def objective(trial: Trial) -> float:
...
+        assert frozen_trial.state == TrialState.FAIL
+        assert frozen_trial.value is None
+
+        frozen_trial = _optimize._run_trial(study, lambda _: [0, 1], catch=())  # type: ignore
+        assert frozen_trial.state == TrialState.FAIL
+        assert frozen_trial.value is None
+
+
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES)
+def test_run_trial_pruned(storage_mode: str) -> None:
```
HideakiImamura, line 951: Ditto.

HideakiImamura (commented):
Thanks for the update. I have several comments around tests. PTAL.

tests/study_tests/test_study.py
himkt: Totally right, sorry for my careless. :bow: 8d62268

himkt:
https://github.com/optuna/optuna/runs/5813743672?check_suite_focus=true#step:8:128
Tests passed in my local environment but CI isn't happy...

himkt:
[3bd7bd3](https://github.com/optuna/optuna/pull/3144/commits/3bd7bd32e1c24f304837be2aa95089a954be3a44) solves. @hvy @HideakiImamura PTAL. :pray:

tests/study_tests/test_optimize.py
```diff
@@ -1,49 +1,188 @@
...
+def test_run_trial_automatically_fail(storage_mode: str, caplog: LogCaptureFixture) -> None:
+    # We need to reconstruct our default handler to properly capture stderr.
+    logging._reset_library_root_logger()
+    logging.enable_default_handler()
+    logging.set_verbosity(logging.INFO)
+    logging.enable_propagation()
 
-        # Test trial with invalid objective value: None
-        def func_none(_: Trial) -> float:
+    logging.enable_propagation()
```
hvy, line 65: Line is repeated, a small careless.
```suggestion
```

tests/study_tests/test_optimize.py
```diff
@@ -1,49 +1,188 @@
...
 
 
 @pytest.mark.parametrize("storage_mode", STORAGE_MODES)
-def test_run_trial(storage_mode: str) -> None:
+def test_run_trial(storage_mode: str, caplog: LogCaptureFixture) -> None:
+    # We need to reconstruct our default handler to properly capture stderr.
+    logging._reset_library_root_logger()
+    logging.enable_default_handler()
+    logging.set_verbosity(logging.INFO)
+    logging.enable_propagation()
```
hvy, line 25: Can we do this, and the disabling of the propagation, in a `pytest.fixture` https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization ?

hvy:
I ran the local test script and it looked good too. 👍

tests/study_tests/test_optimize.py
himkt: Ah, thank you for letting me know! I extract this setup/teardown to the fixture!
5c199a7

HideakiImamura (approved):
I checked all of the test cases. Thanks for the long running work. LGTM!

hvy (approved):
LGTM!

Thank you very much for the effort. 🙇 
While code complexity is still somewhat high, it turned out a lot better than expected.
Tests look great in my opinion and should make future refactoring a lot easier.

There are some nits such as commenting about the in-memory system attribute for warning propagation, and maybe a request to check in unit tests that when `Study.tell` fails with e.g. `ValueError`, that a trial is not modified. These are not significant, or not something introduced by this PR.

himkt:
Thank you so much for informative reviews @hvy. ❤️ + Thank you for the patience!
I created the followup #3454 to apply your comments. 🩹

Issue #3021, opened 2021-10-18
[RFC] Drop Python 3.6 support


himkt:
We will gracefully stop the support for Python 3.6 after the EOL (23 Dec 2021).
https://devguide.python.org/#branchstatus

For now, we are not going to force users to upgrade Python (i.e. we currently don't have a plan to update `python_requires` to be `>=3.7` immediately after EOL). However, third-party libraries drop/will drop Python 3.6 before/after EOL and these may cause problems when using Optuna. We'll gradually stop dealing with such problems and will recommend users upgrade their Python.

not522:
We decide to stop Python 3.6 support for integrations and optuna-examples but keep supporting it in pure Optuna uses. It means that we will remove Python 3.6 tests from the CI of the integrations and the examples but remain in the other CIs.

The maintenance cost of testing Python 3.6 with third-party libraries is increasing because some of them have stopped supporting Python 3.6. However, some Optuna users still use Python 3.6, and then we do not update `python_requires` to be `>=3.7` now.

not522:
I think we have completed all tasks that need to be done now. When Python 3.6 will be the blocker to introduce new features or refactor codes, we will go back here and discuss dropping the Python 3.6 support.

not522:
We decide Optuna v3.0.3 (the latest release) is the last version supporting Python 3.6. Working on #4150.

PR #2442, opened 2021-03-05
Bump `torch` to 1.8.0

hvy:
## Motivation

Try bumping to latest torch releases.

## Description of the changes

See above.

PR #3156, opened 2021-12-09
Bump to `torch` related packages

nzw0301:
## Motivation

Optuna's PyTorch version looks a little bit old. The latest version is 1.10.0, but the used version is 1.8.0. Indeed, BoTorch drops PyTorch 1.8.* support since the recent version, `0.6.0`, according to its [release note](https://github.com/pytorch/botorch/releases/tag/v0.6.0).

## Description of the changes

- Bump to PyTorch related packages: `torch`, `torchvision`, and `torchaudio`.
- Remove upper constraint of Allennlp. 
- Avoid testing Allennlp with Python 3.6 due to incompatibility; Allennlp 2.8.0 does not support Python 3.6 anymore.

Related issue / context 
- https://github.com/optuna/optuna/issues/3021


nzw0301:
According to the CI, the stable Alllennlp versions do not support pytorch==1.10.0. I'll update this PR by following the previous bump by https://github.com/optuna/optuna/pull/2442#issuecomment-791055048.

nzw0301:
Ah, allennlp version is also restricted by optuna's `setup.py`.  I'll investigate the possibility to relax the constraint.

setup.py
```diff
@@ -113,12 +113,12 @@ def get_extras_require() -> Dict[str, List[str]]:
...
-            "torchvision==0.9.0+cpu ; sys_platform!='darwin'",
-            "torchaudio==0.8.0",
-            "allennlp>=2.2.0,<2.7.0",
+            "torch==1.10.0 ; sys_platform=='darwin'",
+            "torch==1.10.0+cpu ; sys_platform!='darwin'",
+            "torchvision==0.11.1 ; sys_platform=='darwin'",
+            "torchvision==0.11.1+cpu ; sys_platform!='darwin'",
+            "torchaudio==0.10.0",
+            "allennlp>=2.2.0 ; python_version>'3.6'",
             "botorch>=0.4.0 ; python_version>'3.6'",
```
himkt, line 122: [Question] Don't we have to update the lower constraint for botorch as well?

himkt (approved):
Thank you for the patch. LGTM in the perspective of AllenNLP.

setup.py
nzw0301: `pytest tests/integration_tests/test_botorch.py` doesn't show any error with botorch==0.4.0 and botorch==0.5.1.

himkt: Ah, so sorry, you are totally right. I forgot to delete this comment when submitting the review...

nzw0301: No worries. In fact, I did not check them. Fortunately, there was no error :)

hvy (commented):
Basically LGTM. I'm just wondering if we must also update the AllenNLP examples in `optuna-examples` to exclude Python 3.6 from the CI matrix https://github.com/optuna/optuna-examples/blob/main/.github/workflows/allennlp.yml. Do you have any insight on this?

himkt:
requirements.txt for AllenNLP example doesn't specify PyTorch so CI would run successfully, but it would be consistent to align depdendencies between `optuna` and `optuna-examples` as you pointed out. @hvy You think we should update optuna-examples, right?

hvy:
Ah, so it's not that the AllenNLP example will start to fail after this change (because of the upper constraint in https://github.com/optuna/optuna-examples/blob/v3.0.0-a0/allennlp/requirements.txt#L1 ?). In any case, yes I think we should keep the aligned, but if that's the case, I think we can go ahead and merge this PR first. Is this understanding correct?

himkt:
>  I think we can go ahead and merge this PR first. Is this understanding correct?

Yes, I think so. :+1:

hvy (approved):
LGTM!

Issue #3010, opened 2021-10-16
setup.py not compatible for setting up a development installation on a windows machine


avats-dev:
As per [setup guide](https://github.com/avats-dev/optuna/blob/master/CONTRIBUTING.md#setup-optuna) , `pip install -e .` fails to install optuna for development on a windows machine/environment.

## Expected behavior
It should install optuna with dev dependencies without any errors.

## Environment
- Windows 10/11
- python 3.9.1
- pip 21.2.4
- setuptools 58.2.0
- wheel 0.37.0
- optuna : current master

## Error messages, stack traces, or logs

```
Installing build dependencies ... done
Getting requirements to build wheel ... error
ERROR: Command errored out with exit status 1:
command: 'c:\python\python.exe' 'c:\python\lib\site-packages\pip\_vendor\pep517\in_process\_in_process.py' get_requires_for_build_wheel 'C:\Users\91728\AppData\Local\Temp\tmprb137rje'
      cwd: C:\Users\91728\Desktop\optuna\optuna
Complete output (22 lines):
  Traceback (most recent call last):
    File "c:\python\lib\site-packages\pip\_vendor\pep517\in_process\_in_process.py", line 349, in <module>
      main()
    File "c:\python\lib\site-packages\pip\_vendor\pep517\in_process\_in_process.py", line 331, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "c:\python\lib\site-packages\pip\_vendor\pep517\in_process\_in_process.py", line 117, in get_requires_for_build_wheel
      return hook(config_settings)
    File "C:\Users\91728\AppData\Local\Temp\pip-build-env-h54rpc43\overlay\Lib\site-packages\setuptools\build_meta.py", line 154, in get_requires_for_build_wheel
      return self._get_build_requires(
    File "C:\Users\91728\AppData\Local\Temp\pip-build-env-h54rpc43\overlay\Lib\site-packages\setuptools\build_meta.py", line 135, in _get_build_requires
      self.run_setup()
    File "C:\Users\91728\AppData\Local\Temp\pip-build-env-h54rpc43\overlay\Lib\site-packages\setuptools\build_meta.py", line 258, in run_setup
      super(_BuildMetaLegacyBackend,
    File "C:\Users\91728\AppData\Local\Temp\pip-build-env-h54rpc43\overlay\Lib\site-packages\setuptools\build_meta.py", line 150, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
[...]

PR #3158, opened 2021-12-09
Fixes 'UnicodeDecodeError: ...' issue on Windows

ysdede:
## Motivation
pip install from source (`pip install -e .`) throws `UnicodeDecodeError: 'charmap' codec can't decode byte` on Windows systems that use character encoding different than cp1252 ( in my case Turkish cp-1254).

## Description of the changes
 Encoding readme file with utf-8 fixes this issue.

 setup.py:
```python
 readme_filepath = os.path.join(os.path.dirname(__file__), "README.md")
    with open(readme_filepath, encoding="utf8") as f:
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ def get_version() -> str:
 def get_long_description() -> str:
 
     readme_filepath = os.path.join(os.path.dirname(__file__), "README.md")
-    with open(readme_filepath) as f:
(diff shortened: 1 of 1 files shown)


nzw0301:
Let me share a related issue. Note that as discussed in https://github.com/optuna/optuna/issues/3010#issuecomment-945422928, this modification is not an issue in practice for Optuna.

Issue #2911, opened 2021-09-08
Align arguments with `create_study` and `load_study`


himkt:
## Motivation

Currently, the orders of arguments are different between `create_study` and `load_study`, although users might think interfaces are the same. In this issue, we consider changing the interface of `create_study` or `load_study` to align them and make a change if we decide to do.

## Description

- Discuss if we should align interfaces of `create_study` and `load_study`
- Decide how to introduce the changes less burdensome for users (if we decide to change)
- Make PR(s) with appropriate messages to let users know about the change

himkt:
The current signature of `load_study` looks following:

```python
def load_study(
    study_name: Optional[str],
    storage: Union[str, storages.BaseStorage],
    sampler: Optional["samplers.BaseSampler"] = None,
    pruner: Optional[pruners.BasePruner] = None,
) -> Study:
```

Albeit `study_name` is optional, it is a positional, which means users have to specify `None` if they want to load a study without its name specified. I think it's reasonable to make `study_name` a keyword argument with default value `None` and switch the order from `study_name`, `storage`, `sampler`, and `pruner` to `storage`, `study_name`, `sampler`, and `pruner`. The order is still different between `create_study` and `load_study` (`study_name` will be second in `load_study` but is fourth in `create_study`) but it is not wired because `study_name` is more important in `load_study` compared to `create_study`.

If anyone has an opinion on this change, please feel free to comment!

HideakiImamura:
Note: https://github.com/optuna/optuna/pull/3159#issuecomment-995820038

HideakiImamura:
Let me close this issue, as #3270 is complete and further discussion is going on in #3324. Please feel free to reopen if you have any concern.

PR #3159, opened 2021-12-10
[RFC] Change order of load_study arguments

himkt:
🔗 #2911 

## Motivation
This PR aligns an orders of arguments between `create_study` and `load_study`.

### DIFF
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -784,9 +784,9 @@ class _Ask(_BaseCommand):
 
         try:
             study = optuna.load_study(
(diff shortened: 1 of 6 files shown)


not522:
How about making all arguments keyword-only? It will help users notice the API change.

himkt:
> making all arguments keyword-only

Nice suggestion, thank you. It feels reasonable to me. 👍

himkt:
[note]

We decided not to change arguments in v3 and guide users to use keyword arguments using warning message.
Let me close and reopen another PR.

PR #3163, opened 2021-12-13
Add tests for transformer with upper bound parameter

himkt:
This PR adds test cases and fixes a bug in tests for a search space transformer.

## Motivation
- 2aa08bd Add test cases for parameters are identical to upper bounds
- 0205e29 Remove a if-else branch


tests/test_transform.py
```diff
@@ -124,14 +130,7 @@ def test_search_space_transform_numerical(
...
     trans_params = trans.transform({"x0": param})
     assert trans_params.size == 1
-
-    if isinstance(
-        distribution,
-        (DiscreteUniformDistribution, IntUniformDistribution, IntLogUniformDistribution),
-    ):
-        assert expected_low <= trans_params <= expected_high
-    else:
-        assert expected_low <= trans_params < expected_high
```
himkt, line 134: Even if a `distribution` is a real distribution, a transformer doesn't subtract eps from a upper bound thus it would be the same as `expected_high`.

https://github.com/optuna/optuna/runs/4506565155?check_suite_focus=true

HideakiImamura (approved):
LGTM.

HideakiImamura:
@himkt Could you rebase on the master branch?

himkt:
@HideakiImamura Done!

hvy (approved):
LGTM!

Issue #3136, opened 2021-11-30
Update doctest to follow the latest MLflow


himkt:
MLflow released the new version (v1.22.0) 🎉 and it changed logging format.
The doctest in Optuna depends on the log format of MLflow so it fails with the latest MLflow.
(🔗 See also the tentative fix in https://github.com/optuna/optuna/pull/3135)

## Expected behavior

All doctests pass.

## Environment

- Optuna version: 2.10.0
- Python version: 3.9.6
- OS: Ubuntu 20.04 (WSL2)
- (Optional) Other libraries and their versions: MLflow 1.22.0

## Error messages, stack traces, or logs

```
Expected:
    INFO: 'my_other_study' does not exist. Creating a new experiment
Got nothing
```

https://github.com/optuna/optuna/runs/4368455449?check_suite_focus=true#step:6:6119

## Steps to reproduce

1. Install Optuna
2. Install the latest version of MLflow (`pip install -U mlflow`)
3. Run doctest

## Reproducible examples (optional)

None (see doctest)

https://github.com/optuna/optuna/blob/51b548324fb00e9b1cb4a60ac68bd0a438566bc4/optuna/integration/mlflow.py#L69

PR #3170, opened 2021-12-17
Remove testoutput from doctest of mlflow integration

himkt:
This PR solves https://github.com/optuna/optuna/issues/3136.

## Motivation
Currently, we check stdout of mlflow integration, which causes the CI error by the change of logging format in mlflow.
I first thought we should fix the tests but now I think we don't need to check this because it is specific to mlflow internal.

This PR removes testoutput in the mlflow test and the version constraint.

## Description of the changes
- ad8849c removes version constraint
- c2755ee removes testoutput from mlflow doctest

### DIFF
--- a/optuna/integration/mlflow.py
+++ b/optuna/integration/mlflow.py
@@ -62,12 +62,6 @@ class MLflowCallback(object):
 
             shutil.rmtree(tempdir)
 
-        .. testoutput::
-            :hide:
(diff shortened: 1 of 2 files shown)


nzw0301 (approved):
👍🏻

hvy (approved):
Sounds reasonable, LGTM! 👍

Issue #2959, opened 2021-09-26
Stabilise visualisation module


nzw0301:
## Motivation

Optuna has two visualisation backends: Plotly and Matplotlib. It would be great to minimise the difference between backends to provide consistent plots. Concretely, 

1. https://github.com/optuna/optuna/issues/3226
2. Unify appearance between plotly and matplotlib
3. https://github.com/optuna/optuna/issues/3225
4. handle with the outliner such as `nan`/`inf`.

In the same backend, we can improve more consistency among different plot functions. Concretely, 

5. Unify appearance among plot functions

### 1. Add more Matplotlib tests

See https://github.com/optuna/optuna/issues/3226

### 2. Unify appearance between Plotly and Matplotlib

I've already created related issues:

- https://github.com/optuna/optuna/issues/2893
- https://github.com/optuna/optuna/issues/2913
- https://github.com/optuna/optuna/issues/3197

But feel free to let me know the thing that should be addressed by this issue.

### 3. Add unimplemented plot functions to `optuna.visualization.matplotlib`

- https://github.com/optuna/optuna/issues/3225

### 4. handle with the outliner such as `nan`/`inf`.

- https://github.com/optuna/optuna/issues/3382

### 5. Unify appearance among plot functions in the same backend

- https://github.com/optuna/optuna/issues/3375

HideakiImamura:
@nzw0301 All enumerated items appear to be complete. Is it OK to close this issue?

nzw0301:
Sorry, not yet; we need to improve the matplotlib's tests. I only added file names that had already been resolved to the issue description above.

nzw0301:
I'll revise the issue description later. Thank you for kindly letting me know this!

nzw0301:
We might be able to add a unit test using objective values. For example, we access `plot_optimization_history`'s data as follows:

```python
import optuna

s = optuna.create_study()

def objective(t):
    
    return t.suggest_float("x", 0., 10.)

s.optimize(objective, n_trials=20)
fig = optuna.visualization.matplotlib.plot_optimization_history(s)
print(a.get_lines()[0].get_data())  # access best value
a.collections[0].get_offsets().data  # objective value
```

akawashiro:
I'm working on test_pareto_front.py.

belldandyxtq:
I will be working on `test_param_importances.py`

not522:
I found that the axes of `optuna.visualization.matplotlib.plot_contour` has bug.

```python
[...]

PR #3175, opened 2021-12-19
Add tests in visualization_tests/matplotlib_tests/test_slice.py

keisukefukuda:
## Motivation

This is a part of my Optuna hackathon(2021-12-19) outcomes. I'm working on #2959 `test_slice` part.

## Description of the changes

This PR adds more tests in `tests/visualization_tests/matplotlib_tests/test_slice.py` since Matplotlib tests have fewer tests than it's Plotly counterpart.

NOTE:
 * Currently it checks `len(figures.get_line()) == 0`, but it 's rather meaningless. In scatter plot, the plotted data is stored as PathCollection, which is available through `figure.findobj(PathCollection)` (actually `figure` is `Axes` object and maybe confusing). Although  `len(figures.get_line()) == 0` is harmless, I deleted them. Please let me know if I should put them back.

 * [x] test_plot_slice
 * [x] test_plot_slice_log_scale

### DIFF
--- a/tests/visualization_tests/matplotlib_tests/test_slice.py
+++ b/tests/visualization_tests/matplotlib_tests/test_slice.py
@@ -60,8 +60,6 @@ def test_plot_slice() -> None:
     # Test with a customized target name.
     figure = plot_slice(study, target_name="Target Name")
     assert len(figure) == 2
-    assert len(figure[0].get_lines()) == 0
-    assert len(figure[1].get_lines()) == 0
     assert len(figure[0].findobj(PathCollection)) == 1
     assert len(figure[1].findobj(PathCollection)) == 1
     assert figure[0].yaxis.label.get_text() == "Target Name"
--- a/tests/visualization_tests/matplotlib_tests/test_slice.py
+++ b/tests/visualization_tests/matplotlib_tests/test_slice.py
@@ -23,43 +23,47 @@ def test_plot_slice() -> None:
     # Test with no trial.
     study = prepare_study_with_trials(no_trials=True)
     figure = plot_slice(study)
-    assert len(figure.get_lines()) == 0
+    assert len(figure.findobj(PathCollection)) == 0
 
     study = prepare_study_with_trials(with_c_d=False)
 
     # Test with a trial.
     figure = plot_slice(study)
     assert len(figure) == 2
-    assert len(figure[0].get_lines()) == 0
-    assert len(figure[1].get_lines()) == 0
     assert len(figure[0].findobj(PathCollection)) == 1
     assert len(figure[1].findobj(PathCollection)) == 1
+    assert figure[0].yaxis.label.get_text() == "Objective Value"
 
+    # Scatter plot data is available as PathCollection
     data0 = figure[0].findobj(PathCollection)[0].get_offsets().data
     data1 = figure[1].findobj(PathCollection)[0].get_offsets().data
-    assert np.allclose(data0, [[1.0, 0.0],
-                               [2.5, 1.0]])
-    assert np.allclose(data1, [[2.0, 0.0],
-                               [0.0, 2.0],
-                               [1.0, 1.0]])
-    assert figure[0].yaxis.label.get_text() == "Objective Value"
+    assert np.allclose(data0, [[1.0, 0.0], [2.5, 1.0]])
+    assert np.allclose(data1, [[2.0, 0.0], [0.0, 2.0], [1.0, 1.0]])
 
     # Test with a trial to select parameter.
     figure = plot_slice(study, params=["param_a"])
-    assert len(figure.get_lines()) == 0
+    assert len(figure.findobj(PathCollection)) == 1
     assert figure.yaxis.label.get_text() == "Objective Value"
+
+    data0 = figure.findobj(PathCollection)[0].get_offsets().data
+    np.allclose(data0, [[1.0, 2.0], [2.5, 1.0]])
 
     # Test with a customized target value.
     with pytest.warns(UserWarning):
         figure = plot_slice(study, params=["param_a"], target=lambda t: t.params["param_b"])
-    assert len(figure.get_lines()) == 0
+    assert len(figure.findobj(PathCollection)) == 1
     assert figure.yaxis.label.get_text() == "Objective Value"
+
+    data0 = figure.findobj(PathCollection)[0].get_offsets().data
+    np.allclose(data0, [[1.0, 2.0], [2.5, 1.0]])
 
     # Test with a customized target name.
     figure = plot_slice(study, target_name="Target Name")
     assert len(figure) == 2
     assert len(figure[0].get_lines()) == 0
     assert len(figure[1].get_lines()) == 0
+    assert len(figure[0].findobj(PathCollection)) == 1
+    assert len(figure[1].findobj(PathCollection)) == 1
     assert figure[0].yaxis.label.get_text() == "Target Name"
 
     # Test with wrong parameters.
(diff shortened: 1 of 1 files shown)


tests/visualization_tests/matplotlib_tests/test_slice.py
```diff
@@ -21,33 +23,47 @@ def test_plot_slice() -> None:
...
+    assert np.allclose(data1, [[2.0, 0.0], [0.0, 2.0], [1.0, 1.0]])
+
     # Test with a trial to select parameter.
     figure = plot_slice(study, params=["param_a"])
-    assert len(figure.get_lines()) == 0
+    assert len(figure.findobj(PathCollection)) == 1
     assert figure.yaxis.label.get_text() == "Objective Value"
 
+    data0 = figure.findobj(PathCollection)[0].get_offsets().data
+    np.allclose(data0, [[1.0, 2.0], [2.5, 1.0]])
```
himkt, line 49:
```suggestion
    assert np.allclose(data0, [[1.0, 2.0], [2.5, 1.0]])
```

tests/visualization_tests/matplotlib_tests/test_slice.py
```diff
@@ -21,33 +23,47 @@ def test_plot_slice() -> None:
...
+
     # Test with a customized target value.
     with pytest.warns(UserWarning):
         figure = plot_slice(study, params=["param_a"], target=lambda t: t.params["param_b"])
-    assert len(figure.get_lines()) == 0
+    assert len(figure.findobj(PathCollection)) == 1
     assert figure.yaxis.label.get_text() == "Objective Value"
 
+    data0 = figure.findobj(PathCollection)[0].get_offsets().data
+    np.allclose(data0, [[1.0, 2.0], [2.5, 1.0]])
```
himkt, line 58:
```suggestion
    assert np.allclose(data0, [[1.0, 2.0], [2.5, 1.0]])
```

himkt (commented):
Let me put minor suggestions.

tests/visualization_tests/matplotlib_tests/test_slice.py
keisukefukuda: Fixed. 🙏

tests/visualization_tests/matplotlib_tests/test_slice.py
keisukefukuda: Fixed 🙏

tests/visualization_tests/matplotlib_tests/test_slice.py
```diff
@@ -21,33 +23,47 @@ def test_plot_slice() -> None:
...
+    assert np.allclose(data1, [[2.0, 0.0], [0.0, 2.0], [1.0, 1.0]])
+
     # Test with a trial to select parameter.
     figure = plot_slice(study, params=["param_a"])
-    assert len(figure.get_lines()) == 0
+    assert len(figure.findobj(PathCollection)) == 1
     assert figure.yaxis.label.get_text() == "Objective Value"
 
+    data0 = figure.findobj(PathCollection)[0].get_offsets().data
+    assert np.allclose(data0, [[1.0, 0.0], [2.5, 1.0]])
```
g-votte, line 49: This pattern appears very frequently in this PR. To avoid code duplication, how about defining a function like:

```python
def _assert_data_allclose(figure, values):
    data = figure.findobj(PathCollection)[0].get_offsets().data
    assert np.allclose(data, values)
```

tests/visualization_tests/matplotlib_tests/test_slice.py
```diff
@@ -21,33 +23,47 @@ def test_plot_slice() -> None:
...
     # Test with a trial.
     figure = plot_slice(study)
     assert len(figure) == 2
-    assert len(figure[0].get_lines()) == 0
-    assert len(figure[1].get_lines()) == 0
+    assert len(figure[0].findobj(PathCollection)) == 1
+    assert len(figure[1].findobj(PathCollection)) == 1
     assert figure[0].yaxis.label.get_text() == "Objective Value"
 
+    # Scatter plot data is available as PathCollection
```
g-votte, line 37: To be consistent with other comments, please put a period at the end of a sentence.
```suggestion
    # Scatter plot data is available as PathCollection.
```

g-votte (changes requested):

tests/visualization_tests/matplotlib_tests/test_slice.py
```diff
@@ -81,19 +98,19 @@ def test_plot_slice_log_scale() -> None:
 
     # Plot a parameter.
     figure = plot_slice(study, params=["y_log"])
-    assert len(figure.get_lines()) == 0
+    assert len(figure.findobj(PathCollection)) == 1
```
g-votte, line 101: In addition to the number of objects, we may want to assert the values using `allclose`, both for test accuracy and consistency with `test_slice_plot`.

g-votte (changes requested):

himkt (approved):
Let me put an approval because tests are now consistent with the counterpart of plotly.

tests/visualization_tests/matplotlib_tests/test_slice.py
```diff
@@ -21,33 +23,47 @@ def test_plot_slice() -> None:
...
     assert figure.yaxis.label.get_text() == "Objective Value"
 
+    data0 = figure.findobj(PathCollection)[0].get_offsets().data
+    assert np.allclose(data0, [[1.0, 2.0], [2.5, 1.0]])
+
     # Test with a customized target name.
     figure = plot_slice(study, target_name="Target Name")
     assert len(figure) == 2
     assert len(figure[0].get_lines()) == 0
     assert len(figure[1].get_lines()) == 0
```
HideakiImamura, line 64: In the above, these lines were removed, why not here?

HideakiImamura (commented):
Thanks for the PR! I agree with @g-votte's comments, and have a minor comment. Could you take a look?

HideakiImamura:
Let me ask another comment. Do you think we should apply the change in this PR to all other matplotlib visualization tests, that is, using `figure.findobj(PathCollection)` instead of `figure.get_lines()`?

tests/visualization_tests/matplotlib_tests/test_slice.py
himkt: I think we can leave them because tests for plotly also have the same point. If we introduce this helper, it would be nice to create another PR to update both tests.

tests/visualization_tests/matplotlib_tests/test_slice.py
keisukefukuda: The plotly `test_slice.py` does not have accuracy tests as well, so do you think we should modify both of plotly/matplotlib in this part?

tests/visualization_tests/matplotlib_tests/test_slice.py
keisukefukuda: This was a mistake, fixed. Thanks!

keisukefukuda:
> Let me ask another comment. Do you think we should apply the change in this PR to all other (scatter) matplotlib visualization tests, that is, using `figure.findobj(PathCollection)` instead of `figure.get_lines()`?

I think so. To make the two backend tests consistent, we need to check PathCollections of the Axes.
As long as scatter plot is used, checking `get_lines()` does not help (although harmless).

tests/visualization_tests/matplotlib_tests/test_slice.py
g-votte: Ah, sorry; just applying my comment causes inconsistency between the matplotlib and plotly tests.

As you mention, adding the same tests to plotly one should be the best way.

g-votte (approved):
LGTM. For the following comments, let's revisit them in future PRs. 
- https://github.com/optuna/optuna/pull/3175#discussion_r771903715
- https://github.com/optuna/optuna/pull/3175#discussion_r771905026

PR #3178, opened 2021-12-19
Add tests for generated plots of `matplotlib.plot_edf`

makinzm:
## Motivation

Add test in matplotlib test ``plot_edf``  of matplotlib.

## Description of the changes

This test confirms 
- that the values are monotonically non-decreasing and  
- that the values are in [0,1].
## Reference
- [Stabilise visualisation module #2959](https://github.com/optuna/optuna/issues/2959)  
- [Add tests for generated plots of plot_edf #3188](https://github.com/optuna/optuna/pull/3188)

### DIFF
--- a/tests/visualization_tests/matplotlib_tests/test_edf.py
+++ b/tests/visualization_tests/matplotlib_tests/test_edf.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Sequence
 
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


HideakiImamura (approved):
Thanks for the PR! LGTM!

tests/visualization_tests/matplotlib_tests/test_edf.py
```diff
@@ -11,6 +14,14 @@ def test_target_is_none_and_study_is_multi_obj() -> None:
         plot_edf(study)
 
 
+def _validate(y_values: List[float]) -> None:
```
toshihikoyanase, line 17: Maybe `Sequence` is preferrable if we can use it.
```suggestion
def _validate(y_values: Sequence[float]) -> None:
```

makinzm: Thank you for your advice. I will commit after modifying and checking.

toshihikoyanase (approved):
Thank you for your quick fix. LGTM!

PR #3179, opened 2021-12-19
[not for merge] Run Python 3.10 tests on CI

semiexp:
## Motivation
We would like to support Python 3.10.

## Description of the changes
This branch is to examine how current tests work on CI with Python 3.10.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


semiexp:
CI with Python 3.10 fails because of `ERROR: Could not find a version that satisfies the requirement torch==1.10.0+cpu (from versions: none)`. It looks like `torch` does not support Python 3.10 yet...

nzw0301:
Thank you for your examination! It was really helpful to create #3353.

PR #3187, opened 2021-12-19
Fix bug of nondeterministic behavior of TPESampler when `group=True`

### LINKED ISSUE #3137: Non-determinism in TPE sampler when `group=True`
TPE sampler is non-deterministic in different runs (with the same seed, no parallelization) when group decomposed search space is enabled.

## Expected behavior

Consider the script below:
```python

import optuna

sampler = optuna.samplers.TPESampler(n_startup_trials=1, seed=2, multivariate=True, group=True)

for _ in range(10):
    trial = study.ask()
    picked = [i for i in range(10) if trial.suggest_int(str(i), 0, 1) == 1]
    print(picked)
    study.tell(trial, len(picked))

```

Different runs of this script are expected to yield the same output (because the seed is fixed, the objective function is deterministic, and we are running in sequential mode), but they do not. For example, two consecutive runs of this script yielded the following outputs (your output might differ) :

First run:
```
[2, 7]
[2, 7]
[2, 7]
[5, 6, 7, 8, 9]
[1, 5]
[0, 2, 3, 4, 7]
[2, 3, 8]
[1, 2, 3, 6, 7]
[0, 1, 2, 4]
[3, 5, 7]
```

Second run:
```
[2, 7]
[2, 7, 9]
[2, 7]
[1, 6]
[2, 3, 6, 7, 8]
[0, 2, 3, 5, 7]
[1, 7, 8]
[0, 1, 2, 6, 7]
[3]
[3, 5]
```

Note that if we set `group=False` when initializing the TPE sampler, we get deterministic outputs.

## Environment

- Optuna version: 2.91
- Python version: 3.7.4
- OS: Linux

Not applicable.

## Steps to reproduce

Run the script above multiple times. Observe that the output changes everytime.

xuzijian629:
## Motivation
Fix #3137

Due to Python's nondeterministic string hashing, the iteration of `Dict[str, BaseDistribution]` is in nondeterministic order.

## Description of the changes
Sorted the keys to make the iteration deterministic.

I don't think that this PR is perfect to resolve the original issue because:
* It is very hard for users and developers to always keep in mind that iterating distributions is nondeterministic while development.
* Moreover, such situation may usually happen because `Dict[str, BaseDistribution]` appears in various positions in the codebase.

One of possible solutions may be creating `SearchSpace` class instead of the raw `Dict[str, BaseDistribution]` and manually handle hashing, but this may require a large-scale modification.

If you have any ideas for modification, please let me know.

### DIFF
--- a/tests/samplers_tests/tpe_tests/test_sampler.py
+++ b/tests/samplers_tests/tpe_tests/test_sampler.py
@@ -1060,7 +1060,9 @@ def run_tpe(k: int, sequence_dict: Dict[int, List[int]], hash_dict: Dict[int, in
     hash_dict[k] = hash("nondeterministic hash")
     sampler = TPESampler(n_startup_trials=1, seed=2, multivariate=True, group=True)
     study = create_study(sampler=sampler)
-    study.optimize(lambda t: np.sum(t.suggest_int(f"x{i}", 0, 10) for i in range(10)), n_trials=2)
+    study.optimize(
+        lambda t: np.sum([t.suggest_int(f"x{i}", 0, 10) for i in range(10)]), n_trials=2
+    )
     sequence_dict[k] = list(study.trials[-1].params.values())
 
 
--- a/tests/samplers_tests/tpe_tests/test_sampler.py
+++ b/tests/samplers_tests/tpe_tests/test_sampler.py
@@ -1,3 +1,5 @@
+from multiprocessing import Manager
+from multiprocessing import Process
 import random
 from typing import Callable
 from typing import Dict
@@ -1066,11 +1068,7 @@ def run_tpe(k: int, sequence_dict: Dict[int, List[int]], hash_dict: Dict[int, in
     sequence_dict[k] = list(study.trials[-1].params.values())
 
 
-# Test deterministic iteration as discussed in https://github.com/optuna/optuna/issues/3137
 def test_group_deterministic_iteration() -> None:
-    from multiprocessing import Manager
-    from multiprocessing import Process
-
     manager = Manager()
     sequence_dict: Dict[int, List[int]] = manager.dict()
     hash_dict: Dict[int, int] = manager.dict()
--- a/tests/samplers_tests/tpe_tests/test_sampler.py
+++ b/tests/samplers_tests/tpe_tests/test_sampler.py
@@ -1057,8 +1057,8 @@ def test_group_experimental_warning() -> None:
         _ = TPESampler(multivariate=True, group=True)
 
 
-# This function is used only in test_group_deterministic_iteration, but declared at top-level because
-# local function cannot be pickled, which occurs within multiprocessing.
+# This function is used only in test_group_deterministic_iteration, but declared at top-level
+# because local function cannot be pickled, which occurs within multiprocessing.
 def run_tpe(k: int, sequence_dict: Dict[int, List[int]], hash_dict: Dict[int, int]) -> None:
     hash_dict[k] = hash("nondeterministic hash")
     sampler = TPESampler(n_startup_trials=1, seed=2, multivariate=True, group=True)
--- a/tests/samplers_tests/tpe_tests/test_sampler.py
+++ b/tests/samplers_tests/tpe_tests/test_sampler.py
@@ -1,5 +1,4 @@
-from multiprocessing import Manager
-from multiprocessing import Process
+import multiprocessing
 import random
 from typing import Callable
 from typing import Dict
@@ -1058,6 +1057,8 @@ def test_group_experimental_warning() -> None:
         _ = TPESampler(multivariate=True, group=True)
 
 
+# This function is used only in test_group_deterministic_iteration, but declared at top-level because
+# local function cannot be pickled, which occurs within multiprocessing.
 def run_tpe(k: int, sequence_dict: Dict[int, List[int]], hash_dict: Dict[int, int]) -> None:
     hash_dict[k] = hash("nondeterministic hash")
     sampler = TPESampler(n_startup_trials=1, seed=2, multivariate=True, group=True)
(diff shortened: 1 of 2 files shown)


xuzijian629:
It seems that the first assertion of the following is not true for python 3.6...

https://github.com/optuna/optuna/blob/8afd18c8df43b80700ba9dc1f2dd544beff6830d/tests/samplers_tests/tpe_tests/test_sampler.py#L1084-L1087

Maybe the behavior of multiprocessing changed from some version after 3.6?

xuzijian629:
Well, I added a test but I'm not sure if that is what is expected. I wait for the review comments

nzw0301:
Thank you for sharing your investigation! In fact, Python 3.7 test will be failed because Converge CI, where python 3.7 was used, was also failed. I'll investigate it a little bit more.

nzw0301:
[The docs of `multiprocessing`](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) says child processes are initialised as a fork of parent process on Linux by default. Instead of `fork`, `spawn` option can create a child process as a fresh python process, so this might avoid the issue:


```python
    import multiprocessing
    from multiprocessing import Manager
    from multiprocessing import Process

    multiprocessing.set_start_method("spawn")
```

xuzijian629:
Thank you for your thorough investigation!

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -1054,6 +1056,37 @@ def test_group_experimental_warning() -> None:
...
+        picked = [i for i in range(10) if trial.suggest_int(str(i), 0, 1) == 1]
+        study.tell(trial, len(picked))
+        sequence.extend(picked)
+    sequence_dict[k] = sequence
+
+
+# Test deterministic iteration as discussed in https://github.com/optuna/optuna/issues/3137
+def test_group_deterministic_iteration() -> None:
+    from multiprocessing import Manager
+    from multiprocessing import Process
```
HideakiImamura, line 1075: How about moving these imports to the top of this file?

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -1054,6 +1056,37 @@ def test_group_experimental_warning() -> None:
...
+    sequence = []
+    for _ in range(10):
+        trial = study.ask()
+        picked = [i for i in range(10) if trial.suggest_int(str(i), 0, 1) == 1]
+        study.tell(trial, len(picked))
+        sequence.extend(picked)
+    sequence_dict[k] = sequence
+
+
+# Test deterministic iteration as discussed in https://github.com/optuna/optuna/issues/3137
```
HideakiImamura, line 1072: Looks like we don't have to note the issue here. I think the comments in the sampler.py are sufficient.

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -1054,6 +1056,37 @@ def test_group_experimental_warning() -> None:
         _ = TPESampler(multivariate=True, group=True)
 
 
+def run_tpe(k: int, sequence_dict: Dict[int, List[int]], hash_dict: Dict[int, int]) -> None:
```
HideakiImamura, line 1059: This function is only used in `test_group_deterministic_iteration`, so how about moving the definition inside `test_group_deterministic_iteration`? Then, this function does not need dictionaries as arguments.

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -1054,6 +1056,37 @@ def test_group_experimental_warning() -> None:
...
+    hash_dict[k] = hash("nondeterministic hash")
+    sampler = TPESampler(n_startup_trials=1, seed=2, multivariate=True, group=True)
+    study = create_study(sampler=sampler)
+    sequence = []
+    for _ in range(10):
+        trial = study.ask()
+        picked = [i for i in range(10) if trial.suggest_int(str(i), 0, 1) == 1]
+        study.tell(trial, len(picked))
+        sequence.extend(picked)
+    sequence_dict[k] = sequence
```
HideakiImamura, line 1069: How about simplifying the test using a standard `Study.optimize`? For example, I confirmed that the following code pass with this PR, but fails with the master.
```suggestion
    sampler = TPESampler(n_startup_trials=1, seed=2, multivariate=True, group=True)
    study = create_study(sampler=sampler)
    study.optimize(lambda t: np.sum(t.suggest_int(f"x{i}", 0, 10) for i in range(10)), n_trials=2)
    sequence_dict[k] = list(study.trials[-1].params.values())
```

HideakiImamura (commented):
Thanks for the PR! Your solution looks great to me. I have several comments around the test. Could you take a look?

tests/samplers_tests/tpe_tests/test_sampler.py
xuzijian629: To use this function in multiprocessing, the function must be pickled.
Since local functions cannot be pickled, I put this on global scope.

https://stackoverflow.com/questions/8804830/python-multiprocessing-picklingerror-cant-pickle-type-function

I update the code and added some comments.

xuzijian629:
Sorry for no response for a long time and thank you for your detailed review!!

I made some updates, based on https://github.com/optuna/optuna/pull/3187#issuecomment-997673037 and https://github.com/optuna/optuna/pull/3187#pullrequestreview-838168688

tests/samplers_tests/tpe_tests/test_sampler.py
HideakiImamura: Thanks you for the explanation. Sounds reasonable.

HideakiImamura (approved):
Thanks for the update! Looks great!

nzw0301 (approved):
LGTM. Thanks!

PR #3200, opened 2021-12-21
Import only train_model in AllenNLP integration

himkt:
After releasing `nltk` v3.6.6, CIs fail when loading wordnet stuff.
🔗 https://github.com/optuna/optuna/runs/4591822163?check_suite_focus=true

The error occurs in the checklist module of AllenNLP, which does not need to import in Optuna in nature.
(The current implementation of AllenNLP integration imports `allennlp.commands` which includes `checklist` module`)
🔗 https://github.com/allenai/allennlp/issues/5521

This PR changes the way to import a method `train_model`: from module-level import to only function import.

## Motivation
- Fix CI failure
- Reduce number of imported stuff

### DIFF
--- a/optuna/integration/allennlp/_executor.py
+++ b/optuna/integration/allennlp/_executor.py
@@ -18,8 +18,8 @@ from optuna.integration.allennlp._variables import SPECIAL_DELIMITER as DELIMITE
 
 
 with try_import() as _imports:
-    from allennlp.commands.train import train_model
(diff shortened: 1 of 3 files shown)


himkt:
Ah, `from allennlp.commands.train import train_model` still imports `__init__.py` on `allennlp/commands` and `checklist` appears there. This patch doesn't make sense. Closing.

Issue #2938, opened 2021-09-19
Determine the scope of application of CachedStorage


keisuke-umezawa:
## Motivation
Currently, `CachedStorage` provides inconsistent functionality due to the following situations:

* `CachedStorage` only supports `RDBStorage`, because it is necessary to implement several methods to be supported by `CachedStorage`, and they are only implemented in `RDBStorage`. However, it is possible to input other storages when constructing `CachedStorage`, and it will cause an error.
* When we pass an instance of `RDBStorage` to `Study`, it is forcibly wrapped in CachedStorage, and it is always CachedStorage internally.

## Description

The solution will be decided in consideration of the following points. Also, the changes have been merged.

* As `CachedStorage` is not compatible with other storages than `RDBStorage`, we will decide it will support other storages and how to support other storages.
* The specifications for `get_storage` to convert `RDBStorage` to `CachedStorage` are well defined.
* Based on the above decision, the implementation of `CachedStorage` has been decided.
    * If the scope of application is only RDB Storage, introduce a mechanism to limit the usage only for `RDBStorage`.
    * If the scope of application includes other storages, consider the implementation method (if the implementation method is the same as the current one, include the arrangement of APIs that the storage must satisfy)

## TODOs
- [x] https://github.com/optuna/optuna/issues/3112
- [x] Make a consensus for the solution in this issue that `CachedStorage` will support other storages and how to support other storages. #3204
- [x] The specifications for `get_storage` to convert `RDBStorage` to `CachedStorage` are well defined. [conclusion](https://github.com/optuna/optuna/issues/2938#issuecomment-1035986042)
- [x] Implementation

### Necessary methods to implement for `CachedStorage` 
- [_get_trials](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/storage.py#L870)
- [_create_new_trial](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/storage.py#L506)
- [_check_and_set_param_distribution](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/storage.py#L709)

keisuke-umezawa:
## Note for `Make a consensus for the solution in this issue that CachedStorage will support other storages and how to support other storages.`

In my opinion, if we just implement `_get_trials` in a storage, we can provide the cache functionality for the storage. As it is not so difficult, we can make it mandatory to implement in each storage class.

Current status:
- RDBStorage: Implemented
- RedisStorage: Will be implemented in #3204 
[...]

PR #3204, opened 2021-12-23
Wrap RedisStorage in CachedStorage

masap:
## Motivation
This PR fixes issue #2938.

Before this PR, ```study.optimize()``` required 305 seconds when ```n_trials=1000```, after this PR, it requires 128 seconds. This PR reduced the execution time about 58%.

## Description of the changes

Previously ```CachedStorage``` only wrap the ```RDBStorage```. This PR wrap RedisStorage in CachedStorage also.

### DIFF
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -4,6 +4,7 @@ from packaging import version
 
 import optuna
 from optuna.storages._cached_storage import _CachedStorage
+from optuna.storages._rdb.storage import RDBStorage
 
 
 # Define key names of `Trial.system_attrs`.
@@ -60,7 +61,10 @@ class PyTorchLightningPruningCallback(Callback):
         if self.is_ddp_backend:
             if version.parse(pl.__version__) < version.parse("1.5.0"):
                 raise ValueError("PyTorch Lightning>=1.5.0 is required in DDP.")
-            if not isinstance(self._trial.study._storage, _CachedStorage):
+            if not (
+                isinstance(self._trial.study._storage, _CachedStorage)
+                and isinstance(self._trial.study._storage._backend, RDBStorage)
+            ):
                 raise ValueError(
                     "optuna.integration.PyTorchLightningPruningCallback"
                     " supports only optuna.storages.RDBStorage in DDP."
--- a/optuna/storages/_redis.py
+++ b/optuna/storages/_redis.py
@@ -103,9 +103,9 @@ class RedisStorage(BaseStorage):
 
         self._url = url
         self._redis = redis.Redis.from_url(url)
-        self._heartbeat_interval = heartbeat_interval
-        self._grace_period = grace_period
-        self._failed_trial_callback = failed_trial_callback
+        self.heartbeat_interval = heartbeat_interval
+        self.grace_period = grace_period
+        self.failed_trial_callback = failed_trial_callback
 
     def create_new_study(self, study_name: Optional[str] = None) -> int:
 
@@ -705,12 +705,12 @@ class RedisStorage(BaseStorage):
         return confirmed
 
     def _get_stale_trial_ids(self, study_id: int) -> List[int]:
-        assert self._heartbeat_interval is not None
+        assert self.heartbeat_interval is not None
 
-        if self._grace_period is None:
-            grace_period = 2 * self._heartbeat_interval
+        if self.grace_period is None:
+            grace_period = 2 * self.heartbeat_interval
         else:
-            grace_period = self._grace_period
+            grace_period = self.grace_period
 
         current_time = self._get_redis_time()
         heartbeats = self._redis.hgetall(self._key_study_heartbeats(study_id))
@@ -730,11 +730,11 @@ class RedisStorage(BaseStorage):
         return True
 
     def get_heartbeat_interval(self) -> Optional[int]:
-        return self._heartbeat_interval
+        return self.heartbeat_interval
 
     @staticmethod
     def _key_study_heartbeats(study_id: int) -> str:
         return "study_id:{:010d}:heartbeats".format(study_id)
 
     def get_failed_trial_callback(self) -> Optional[Callable[["optuna.Study", FrozenTrial], None]]:
-        return self._failed_trial_callback
+        return self.failed_trial_callback
--- a/optuna/storages/_redis.py
+++ b/optuna/storages/_redis.py
@@ -8,6 +8,7 @@ from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Sequence
+from typing import Set
 from typing import Tuple
 
 import optuna
@@ -321,6 +322,24 @@ class RedisStorage(BaseStorage):
 
     def create_new_trial(self, study_id: int, template_trial: Optional[FrozenTrial] = None) -> int:
 
+        return self._create_new_trial(study_id, template_trial)._trial_id
+
+    def _create_new_trial(
+        self, study_id: int, template_trial: Optional[FrozenTrial] = None
+    ) -> FrozenTrial:
+        """Create a new trial and returns a :class:`~optuna.trial.FrozenTrial`.
+
+        Args:
+            study_id:
+                Study id.
+            template_trial:
+                A :class:`~optuna.trial.FrozenTrial` with default values for trial attributes.
+
+        Returns:
+            A :class:`~optuna.trial.FrozenTrial` instance.
+
+        """
+
         self._check_study_id(study_id)
 
         if template_trial is None:
@@ -361,7 +380,7 @@ class RedisStorage(BaseStorage):
         if trial.state.is_finished():
             self._update_cache(trial_id)
 
-        return trial_id
+        return trial
 
     @staticmethod
     def _create_running_trial() -> FrozenTrial:
@@ -505,6 +524,18 @@ class RedisStorage(BaseStorage):
             pipe.set(self._key_study_summary(study_id), pickle.dumps(study_summary))
             pipe.execute()
 
+    def _check_and_set_param_distribution(
+        self,
+        study_id: int,
+        trial_id: int,
+        param_name: str,
+        param_value_internal: float,
+        distribution: distributions.BaseDistribution,
+    ) -> None:
+
+        self._check_study_id(study_id)
+        self.set_trial_param(trial_id, param_name, param_value_internal, distribution)
+
     def get_trial_param(self, trial_id: int, param_name: str) -> float:
 
         distribution = self.get_trial(trial_id).distributions[param_name]
@@ -618,20 +649,31 @@ class RedisStorage(BaseStorage):
         states: Optional[Tuple[TrialState, ...]] = None,
     ) -> List[FrozenTrial]:
 
+        trials = self._get_trials(study_id, states, set())
+
+        if deepcopy:
+            return copy.deepcopy(trials)
+        else:
+            return trials
+
+    def _get_trials(
+        self,
+        study_id: int,
+        states: Optional[Tuple[TrialState, ...]],
+        excluded_trial_ids: Set[int],
+    ) -> List[FrozenTrial]:
+
         self._check_study_id(study_id)
 
         trials = []
-        trial_ids = self._get_study_trials(study_id)
+        trial_ids = set(self._get_study_trials(study_id)) - excluded_trial_ids
         for trial_id in trial_ids:
             frozen_trial = self.get_trial(trial_id)
 
             if states is None or frozen_trial.state in states:
                 trials.append(frozen_trial)
 
-        if deepcopy:
-            return copy.deepcopy(trials)
-        else:
-            return trials
+        return trials
 
     def read_trials_from_remote_storage(self, study_id: int) -> None:
         self._check_study_id(study_id)
--- a/tests/storages_tests/test_storages.py
+++ b/tests/storages_tests/test_storages.py
@@ -1229,6 +1229,23 @@ def test_fail_stale_trials(storage_mode: str) -> None:
         assert study.trials[0].state is TrialState.FAIL
 
 
+# This test is a workaround. This test should be removed when get_storage()
+# can return raw RDBStorage and RedisStorage. Because then test_fail_stale_trials
+# can test the fail_stale_trials().
+@pytest.mark.parametrize("storage_mode", ["sqlite", "redis"])
+def test_fail_stale_trials_raw(storage_mode: str) -> None:
+    heartbeat_interval = 1
+    grace_period = 2
+
+    with StorageSupplier(storage_mode) as storage:
+        assert isinstance(storage, (RDBStorage, RedisStorage))
+        storage.heartbeat_interval = heartbeat_interval
+        storage.grace_period = grace_period
+
+        study_id = storage.create_new_study()
+        assert storage.fail_stale_trials(study_id) == []
+
+
 @pytest.mark.parametrize("storage_mode", STORAGE_MODES)
 def test_read_trials_from_remote_storage(storage_mode: str) -> None:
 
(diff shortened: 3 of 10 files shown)


masap:
This is the code used for measurement.

```python
import optuna
import time

def objective(trial):
    x = trial.suggest_float('x', -10, 10)
    return (x - 2) ** 2

res = []
for i in range(10):
    storage = optuna.storages.RedisStorage(
        url='redis://localhost:6379',
    )

    start_time = time.monotonic()
    end_time = time.monotonic()

    res.append(end_time - start_time)

print(res)
```

HideakiImamura:
@keisuke-umezawa Could you review this PR since you are the author of #2938?

optuna/storages/_redis.py
```diff
@@ -321,6 +322,24 @@ def get_all_study_summaries(self) -> List[StudySummary]:
 
     def create_new_trial(self, study_id: int, template_trial: Optional[FrozenTrial] = None) -> int:
 
+        return self._create_new_trial(study_id, template_trial)._trial_id
+
+    def _create_new_trial(
+        self, study_id: int, template_trial: Optional[FrozenTrial] = None
+    ) -> FrozenTrial:
+        """Create a new trial and returns its trial_id and a :class:`~optuna.trial.FrozenTrial`.
```
HideakiImamura, line 330: For consistency.
```suggestion
        """Create a new trial and returns a :class:`~optuna.trial.FrozenTrial`.
```

optuna/storages/_redis.py
```diff
@@ -104,7 +105,7 @@ def __init__(
         self._redis = redis.Redis.from_url(url)
         self._heartbeat_interval = heartbeat_interval
         self._grace_period = grace_period
-        self._failed_trial_callback = failed_trial_callback
+        self.failed_trial_callback = failed_trial_callback
```
HideakiImamura, line 108: Why do you change the name of this attribute?

optuna/storages/_redis.py
```diff
@@ -505,6 +524,18 @@ def _set_best_trial(self, study_id: int, trial_id: int) -> None:
             pipe.set(self._key_study_summary(study_id), pickle.dumps(study_summary))
             pipe.execute()
 
+    def _check_and_set_param_distribution(
```
HideakiImamura, line 527: This method is used in `_CachedStorage`. I think we should define this method in the `BaseStorage` for the maintainability. What do you think?

HideakiImamura (commented):
Thanks for the PR! I have several early comments. Could you take a look?

optuna/storages/_redis.py
masap: Thanks! I also fixed same typo in ```RDBStorage```.

optuna/storages/_redis.py
masap: Because previously we used ```_failed_trial_callback``` only for ```RedisStorage```, but now it is also used by ```_CachedStorage```.

optuna/storages/_redis.py
masap: I do not think so...
Because ```BaseStorage``` is also inhedited by ```InMemoryStorage``` and it does not use ```_check_and_set_param_distribution```.
```
BaseStorage
 |
 | inherit
 |
 +-- _CachedStorage (calls _check_and_set_param_distribution of wrapping class)
 |    |
 |    | wrap
 |    |
 |    +--- RDBStorage
 |    |
 |    +--- RedisStorage
 |
 +-- InMemoryStorage (does not calls _check_and_set_param_distribution because InMemoryStorage does not wrapping class)
```

masap:
> Besides below comments, it looks like the test coverage decreases. Do you have any idea to improve it?

Ok, I will try to fix it.

masap:
When used with ```_CachedStorage```, these functions in ```RedisStorage``` are not used.
```
get_trial_number_from_id
read_trials_from_remote_storage
fail_stale_trials
```
These functions decrease test coverage. Of course we can add some unit tests for them to increase test coverage. However, since we will not use ```RedisStorage``` directly in the future, but will use it together with ```_CachedStorage```, there is no point in keeping these functions. So I changed these functions to raise ```NotImplementedError```. ```RDBStorage```, which has a similar problem, has also been fixed.

Any comments would be appreciated.

optuna/storages/_redis.py
masap: Now I modifyed ```get_failed_trial_callback``` of ```_CachedStorage``` to call ```get_failed_trial_callback``` of ```_backend``` instead. Because it is more reasonable and increases test coverage.

HideakiImamura: Sounds reasonable.

keisuke-umezawa:
@masap 
Thank you for the implementation!

> get_trial_number_from_id
> read_trials_from_remote_storage
> fail_stale_trials
> These functions decrease test coverage. Of course we can add some unit tests for them to increase test coverage. However, since we will not use RedisStorage directly in the future, but will use it together with _CachedStorage, there is no point in keeping these functions. So I changed these functions to raise NotImplementedError. RDBStorage, which has a similar problem, has also been fixed.

Basically, this is true. But, we have not concluded that we will not provide any chances for users to use RedisStorage and RDBStorage directly. Maybe, we will concluded it like the above. But, before that, could you leave those functions as they are?

masap:
@keisuke-umezawa Thank you for your review! Ok, I'll leave those functions as they were.

masap:
I reverted these functions, but that reduces the test coverage, so I added some tests.
But now the coverage is not good still, I will investigate it.

masap:
Now test coverage is fixed.
Any comments would be appreciated.

tests/storages_tests/test_storages.py
```diff
@@ -158,7 +158,6 @@ def test_get_study_id_from_name_and_get_study_name_from_id(storage_mode: str) ->
         # Generate unique study_name from the current function name and storage_mode.
         function_name = test_get_study_id_from_name_and_get_study_name_from_id.__name__
         study_name = function_name + "/" + storage_mode
-        storage = optuna.storages.get_storage(storage)
```
HideakiImamura, line 161: Why this line is deleted? Many other calls to get_storage have also been removed, why?

tests/storages_tests/test_storages.py
```diff
@@ -1232,3 +1227,28 @@ def failed_trial_callback(study: "optuna.Study", trial: FrozenTrial) -> None:
         optuna.storages.fail_stale_trials(study)
 
         assert study.trials[0].state is TrialState.FAIL
+
+
+@pytest.mark.parametrize("storage_mode", ["sqlite", "redis"])
+def test_fail_stale_trials_raw(storage_mode: str) -> None:
```
HideakiImamura, line 1233: What is the purpose of this test?

tests/storages_tests/test_storages.py
```diff
@@ -1232,3 +1227,28 @@ def failed_trial_callback(study: "optuna.Study", trial: FrozenTrial) -> None:
...
+    with StorageSupplier(storage_mode) as storage:
+        assert isinstance(storage, (RDBStorage, RedisStorage))
+        storage.heartbeat_interval = heartbeat_interval
+        storage.grace_period = grace_period
+
+        study_id = storage.create_new_study()
+        assert storage.fail_stale_trials(study_id) == []
+
+
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES_HEARTBEAT)
```
HideakiImamura, line 1246: Why the possible `storage_mode`s are `STORAGE_MODES_HEARTBEAT`? Might be `STORAGE_MODES`.

HideakiImamura (commented):
Thanks for the update. I have several comments around tests. Please take a look.

tests/storages_tests/test_storages.py
masap: 英語で正確に説明する自信が無かったので日本語にて失礼します。

この PR は ```RedisStorage``` を ```CachedStorage``` で wrap しており、この修正は test coverage を悪化させます。理由は、いくつかの関数（今まで直接 ```RedisStorage``` を使用していた場合は呼ばれていた関数）が呼ばれなくなるからです。

```RDBStorage``` についても同様に、```CachedStorage``` で wrap した際に、同じ関数が呼ばれなくなります。

最初の commit においては、 test coverage が悪化することは認識していたものの、 ```RDBStorage``` と同じレベルに悪化するので問題ないだろうと思い、そのままの状態で PR を送りました。

しかし、 test coverage を改善するようにとのご指摘があり、改善のために、 ```RDBStorage/RedisStorage``` の呼ばれない関数を obsolete にしてみました。

しかし、 ```RDBStorage/RedisStorage``` を ```CachedStorage``` で wrap せずに使う可能性があるので、それはやめて欲しいとのご指摘がありました。

そこで、 test coverage の改善のために、 ```RDBStorage/RedisStorage``` を直接使用するテストケースを追加しました。いくつかのテストにおいて ```get_storage()``` を削除しているのはそのためです。```get_storage()``` は ```CachedStorage``` で ```RDBStorage/RedisStorage``` を wrap してしまうので、それを避ける必要がありました。

File: tests/storages_tests/test_storages.py (hunk 5 above)
Comment (masap) [reply tests/storages_tests/test_storages.py:1233]: ```study``` を作成する際に内部で必ず ```get_storage()``` を呼んでしまうので ```fail_stale_trials()``` を生の ```RDBStorage/RedisStorage``` でテストすることは現状できません。

そこで、苦肉の策として生の ```RDBStorage/RedisStorage``` から ```fail_stale_trials()``` を呼んで、エラーにならないことを確認するレベルのテストがこれです。

「これだったら無い方がマシ」ということであれば削除します。

File: tests/storages_tests/test_storages.py (hunk 6 above)
Comment (masap) [reply tests/storages_tests/test_storages.py:1246]: ```STORAGE_MODES``` は ```inmemory``` を含みますが、```inmemory```はもちろんremote storageではないので```read_trials_from_remote_storage()```を持たず、エラーとなります。

そのため```STORAGE_MODES_HEARTBEAT```を使用しました。「hertbeat と無関係なテストでなぜ```STORAGE_MODES_HEARTBEAT```を使うのだ？」という違和感があるのは理解できます。私もそうだったので、当初は ```"sqlite", "cached_sqlite", "redis", "cached_redis"``` と並べて書こうかなとも思いましたが、結局「定数使え」と言われそうだなと思い、今の形にしました。

File: tests/storages_tests/test_storages.py (hunk 4 above)
Comment (masap) [reply tests/storages_tests/test_storages.py:161]: This PR wraps `RedisStorage` with `CachedStorage`, and this PR worsens the test coverage. The reason is that some functions (that would have been called if you were using `RedisStorage` directly) will no longer be called.

Similarly for `RDBStorage`, the same functions will no longer be called when wrapping with `CachedStorage`.

In the first commit, I was aware that the test coverage would deteriorate, but I thought it would be fine since it would deteriorate to the same level as `RDBStorage`, so I sent the PR as is.

However, I was asked to improve the test coverage, so I tried to obsolete the functions that are not called in `RDBStorage/RedisStorage` in order to improve it.

However, it was pointed out that `RDBStorage/RedisStorage` may be used without wrapping in `CachedStorage`, and they don't want that to happen.

So, to improve the test coverage, I have added test cases that use `RDBStorage/RedisStorage` directly. That's why I removed `get_storage()` in some tests, because `get_storage()` wraps `RDBStorage/RedisStorage` in `CachedStorage`, and we needed to avoid that.

File: tests/storages_tests/test_storages.py (hunk 5 above)
Comment (masap) [reply tests/storages_tests/test_storages.py:1233]: Currently, it is not possible to test `fail_stale_trials()` on raw `RDBStorage/RedisStorage` because `get_storage()` is always called internally when the `study` is created.

So, as a "last resort", here is a test that calls `fail_stale_trials()` from raw `RDBStorage/RedisStorage` to make sure it doesn't fail. If you think you're better off without it, I will remove it.

File: tests/storages_tests/test_storages.py (hunk 6 above)
Comment (masap) [reply tests/storages_tests/test_storages.py:1246]: `STORAGE_MODES` contains `inmemory` storage, but `inmemory` storage is of course not remote storage, so it does not have `read_trials_from_remote_storage()`, which will result in an error.

This is why we used `STORAGE_MODES_HEARTBEAT`. You may be wondering, "Why use `STORAGE_MODES_HEARTBEAT` for a test that has nothing to do with hertbeat?" I can understand why you might feel uncomfortable. At first, I thought about writing `"sqlite", "cached_sqlite", "redis", and "cached_redis"`, but then I thought that people would say "use constants", so I decided to use the current form.

File: tests/storages_tests/test_storages.py (hunk 4 above)
Comment (HideakiImamura) [reply tests/storages_tests/test_storages.py:161]: Thank you for the explanation. I agree with you.

File: tests/storages_tests/test_storages.py (hunk 5 above)
Comment (HideakiImamura) [reply tests/storages_tests/test_storages.py:1233]: Thank you for the explanation. How about adding the comment to note that we should remove this test if we have a option of the `get_storage` to enable a cache?

File: tests/storages_tests/test_storages.py (hunk 6 above)
Comment (HideakiImamura) [reply tests/storages_tests/test_storages.py:1246]: Thank yoo for the explanation. I agree with the idea that we should not test this with the in-memory storage. How about using `STORAGE_MODE` and early return the test if the `storage_mode` is the in-memory storage? It would be appriciate that you provide some useful comments.

File: tests/storages_tests/test_storages.py (hunk 5 above)
Comment (masap) [reply tests/storages_tests/test_storages.py:1236]: Ok, fixed.

File: tests/storages_tests/test_storages.py (hunk 6 above)
Comment (masap) [reply tests/storages_tests/test_storages.py]: Sorry. This is my mis-undestanding. `InMemoryStorage` has `read_trials_from_remote_storage()` and it can pass the test. So I fixed it to use just `STORAGE_MODES`.

I have modified `read_trials_from_remote_storage()` of `InMemoryStorage` to raise `NotImplemented` exception for investigation and wrongly used it for test...

**Review by HideakiImamura — APPROVED** [review]
Thanks for the fix! LGTM!

File: optuna/integration/pytorch_lightning.py [hunk 7]
```diff
@@ -60,7 +61,9 @@ def on_init_start(self, trainer: Trainer) -> None:
         if self.is_ddp_backend:
             if version.parse(pl.__version__) < version.parse("1.5.0"):
                 raise ValueError("PyTorch Lightning>=1.5.0 is required in DDP.")
-            if not isinstance(self._trial.study._storage, _CachedStorage):
+            if (not isinstance(self._trial.study._storage, _CachedStorage)) or (
+                not isinstance(self._trial.study._storage._backend, RDBStorage)
+            ):
```
himkt, line 66:
```suggestion
            if not (
                isinstance(self._trial.study._storage, _CachedStorage)
                and isinstance(self._trial.study._storage._backend, RDBStorage)
            ):
```

How about using transformed version?

masap: Thanks, it is more clear. Fixed.

optuna/storages/_redis.py
```diff
@@ -618,20 +649,31 @@ def get_all_trials(
...
     ) -> List[FrozenTrial]:
 
+        trials = self._get_trials(study_id, states, set())
+
+        if deepcopy:
+            return copy.deepcopy(trials)
+        else:
+            return trials
+
+    def _get_trials(
```
keisuke-umezawa, line 659: Could you add unit tests for this method? esp, for exlucded_trial_ids.

optuna/storages/_redis.py
keisuke-umezawa: Even though it will not be used by `InMemoryStorage`, we can define it in `BaseStorage`. Also, it will be better if we will define another wrapped storeges.
Could you move the definition to `BaseStorage`?

keisuke-umezawa (commented):
Almost LGTM. Could you check a few comments?

optuna/storages/_redis.py
masap: Ok, fixed. I reused the `test_get_trials_excluded_trial_ids()` of `RDBStorage` test.

optuna/storages/_redis.py
masap: That fix seems to be beyond the scope of this PR... So I will fix it in another PR after this PR is merged. Is it OK ?

masap:
I also fixed some existing `black` and `blackdoc` errors to avoid check failure.

masap:
I fixed `black` and `blackdoc` errors at #3260. Please re run test for this PR after #3260 is merged.

masap:
Rebased.

keisuke-umezawa (approved):
LGTM! Thank you for your contribution!

Issue #2993, opened 2021-10-08
Stabilize _callbacks.py


HideakiImamura:
## Motivation

There are currently two callback functions provided by Optuna, `MaxTrialsCallback` and `RetryFailedTrialsCallback`. In this development item, we will consider removing experimental labels from them.

In particular, there are the following concerns regarding `RetryFailedTrialsCallback`:
(1) The heartbeat function of `RDBStorage` and `_Cached Storage`, which is the premise of this function, is not sufficiently dog ​​fooded.
(2) When this function and `GridSampler` are used at the same time, the failed parameter is retried and the grid is consumed at the same time. That is, the failed parameter is preferentially sampled and its grid is no longer evaluated. This problem is similar to the problem with `enqueue_trial` described in "Stabilize study.py".
(3) The information of intermediate values ​​is not inherited in the trial that is retried by this function.

We will consider removing experimental labels after resolving these concerns.

[Update 2022/01/14]: We have decided there is no concern about `MaxTrialsCallback`, so we can remove the experimental label right now. On the other hand, the heartbeat functionality itself has not yet been fully exploited, so there are concerns about `RetryFailedTrialsCallback`. We need more document and improvement. See the below more details.

## What should we do

For `MaxTrialsCallback`
- [x] Remove the experimental label from `MaxTrialsCallback`. #3261

For `RetryFailedTrialsCallback`
- [x] Add FAQ entry for heatbeat. https://github.com/optuna/optuna/pull/3229 
- [x] ~Add a tutorial page for heartbeat.~ We have decided not to add a tutorial for heaetbeat since it is difficult to reproduce the situation where a trial is stale. We lead to the FAQ page and the example page instead of it. #3294
- [x] Fix the behavior of `RetryFailedTrialsCallback` so that it retry the latest failed trial. https://github.com/optuna/optuna/pull/3223
- [x] The behavior when `RetryFailedTrialsCallback` and `GridSampler` are used at the same time has been discussed and corrected appropriately.　https://github.com/optuna/optuna/pull/2946
- [x] Add an option to determine whether retry trials inherit intermediate values or not. #3269
- [x] ~After finishing all of above tasks, discuss whether the experimental label is removed or not at the end of April 2022.~ We discussed storages and concluded in #2943 that there should be a major refactoring of `BaseStorage`. Since the heartbeat functionality is also likely to undergo major changes, including interface changes, it was decided not to remove the `RetryFailedTrialsCallback`'s experimental in V3.

[...]

PR #3223, opened 2022-01-10
Add a history of retried trial numbers in `Trial.system_attrs`

belltailjp:
This PR originally proposed to change `Trial.system_attrs['failed_trial']` inserted by `RetryFailedTrialCallback` to be the trial number of the latest failure rather than the first, in order to help user to identify the latest situation of (repeatedly) failed trials, such as training snapshots in deep learning workloads.
Through the discussions, it turned into
- Keep `system_attrs['failed_trial']` as it is
- Add `system_attrs['retry_history']` which is a list of trial IDs in the ascending order, whose last item is identical to `'failed_trial'` (if exists)

Below is my _original_ proposal which is now obsolete but the motivation is still relevant.

-----

This PR proposes to introduce a breaking change for the experimental feature `RetryFailedTrialCallback`, in order to make retries always from the latest failure instead of the first one, to better support the case of repetitive failure and retry.

## Motivation

In the current implementation of `RetryFailedTrialCallback`, it always retries from *the first failed trial*, meaning that its `failed_trial` system attribute (`Trial.system_attrs['failed_trial']`) always points to the first one even in case of repetitive failure is occurred.
https://github.com/optuna/optuna/blob/v3.0.0-a0/optuna/_callbacks.py#L102-L105

The typical use of the `RetryFailedTrialCallback` in case of long-running workloads (such as deep learning) would be to recreate a new trial when failure, identify the failed trial using `retried_trial_number()` method, bring the intermediate results from the failed one (e.g., training snapshots) and run the new trial as a restarted process from them.
If the retried trials always point to the very first failed one, it means we can only restart from the time that failed first, no matter how many retries are made and how they progressed.

Here is an example of time series:
* trial=0 fails at step 5
    * `RetryFailedTrialCallback` adds a new trial=1 as the retry of the trial=0 (creates a new trial and set `system_attrs['failed_trial']=0`)
* trial=1 restarts from step 5, but failed again at step 15
    * The callback adds a new trial=2, *still as the retry of the trial=0*
* trial=2 restarts from step 5, rather than from step 10
* ...

and its brief diagram (`>` indicates when each trial has started, `*` is trial step in progress, `X` is failure).
```
-> step
trial 0 | >*****X
trial 1 |        >**********X
trial 2 |        >********...       # <---- trial 2 restarts from the time the trial=0 failed
```

As stated in the documentation, `RetryFailedTrialCallback` is particularly useful in case of the situation where preemption can happen. In that sense, we basically don't want to throw away any progress in the retried (but failed again) trials, i.e. step 6-15 in the trial=1 in the above example, so I'd say the current strategy is not efficient.
Moreover if there is a higher chance that the retried trial gets failed again before its completion (maybe due to too frequent preemption), we can basically expect no trials completed.
https://optuna.readthedocs.io/en/v3.0.0-a0/reference/generated/optuna.storages.RetryFailedTrialCallback.html

Therefore, I think it's better to change the behavior of the `RetryFailedTrialCallback` so it makes a retry from the latest failure instead of the first one, which would look like:
```
-> step
trial 0 | >*****X
trial 1 |        >**********X
trial 2 |                    >********...
```

## Description of the changes

* I modify `Trial.system_attrs['failed_trial']` to point to the last failed trial.
* Doing this will make it a bit difficult to identify how many retries originated from a certain trial are made (therefore difficult to realize the `max_retry` feature), I introduced a retry counter as `Trial.system_attrs['retry_count']` that increments every time `RetryFailedTrialCallback` makes a retry, which is simple and efficient.

Although I think the proposed behavior is more useful in practice as far as I can think of, if there would be a situation where the original one should still be better (or better to keep compatibility), I think it's possible to provide some sort of option.

### DIFF
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -143,7 +143,7 @@ class RetryFailedTrialCallback:
         return trial.system_attrs.get("failed_trial", None)
 
     @staticmethod
-    @experimental("2.11.0")
+    @experimental("3.0.0")
     def retry_history(trial: FrozenTrial) -> List[int]:
         """Return the list of trial IDs from the original trial to the specified one.
 
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -67,11 +67,11 @@ class RetryFailedTrialCallback:
     recreate the trial in :obj:`TrialState.WAITING` to queue up the trial to be run again.
 
     The failed trial can be identified by the
-    :func:`~optuna.storage.RetryFailedTrialCallback.retried_trial_number` function.
+    :func:`~optuna.storages.RetryFailedTrialCallback.retried_trial_number` function.
     Even if repetitive failure occurrs (a retried trial fails again),
     this method returns the number of the original trial.
     To get a full list including the numbers of the retried trials as well as their original trial,
-    call the :func:`~optuna.storage.RetryFailedTrialCallback.retry_history` function.
+    call the :func:`~optuna.storages.RetryFailedTrialCallback.retry_history` function.
 
     This callback is helpful in environments where trials may fail due to external conditions,
     such as being preempted by other processes.
@@ -158,7 +158,7 @@ class RetryFailedTrialCallback:
         Returns:
             A list of trial numbers in ascending order of the series of retried trials.
             The first item of the list indicates the original trial which is identical
-            to the :func:`~optuna.storage.RetryFailedTrialCallback.retried_trial_number`,
+            to the :func:`~optuna.storages.RetryFailedTrialCallback.retried_trial_number`,
             and the last item is the one right before the specified trial in the retry series.
             If the specified trial is not a retry of any trial, returns an empty list.
         """
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -64,6 +64,9 @@ class RetryFailedTrialCallback:
 
     When a trial fails, this callback can be used with the :class:`optuna.storage` class to
     recreate the trial in :obj:`TrialState.WAITING` to queue up the trial to be run again.
+    In case the retried trial fails again and is to be retried further,
+    the :func:`~optuna.storage.RetryFailedTrialCallback.retried_trial_number` of the next
+    recreated trial will point to the latest failed one.
 
     This is helpful in environments where trials may fail due to external conditions, such as
     being preempted by other processes.
@@ -136,7 +139,7 @@ class RetryFailedTrialCallback:
                 The trial object.
 
         Returns:
-            The number of the first failed trial. If not retry of a previous trial,
+            The number of the latest failed trial. If not retry of a previous trial,
             returns :obj:`None`.
         """
 
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -145,7 +145,7 @@ class RetryFailedTrialCallback:
     @staticmethod
     @experimental("3.0.0")
     def retry_history(trial: FrozenTrial) -> List[int]:
-        """Return the list of trial IDs from the original trial to the specified one.
+        """Return the list of retried trial IDs with respect to the specified trial.
 
         Args:
             trial:
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -106,14 +106,14 @@ class RetryFailedTrialCallback:
         self._inherit_intermediate_values = inherit_intermediate_values
 
     def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
-        retry_count = trial.system_attrs.get('retry_count', 0) + 1
+        retry_count = trial.system_attrs.get("retry_count", 0) + 1
         if self._max_retry is not None and retry_count > self._max_retry:
             return
 
         system_attrs = {
             **trial.system_attrs,
             "failed_trial": trial.number,
-            "retry_count": retry_count
+            "retry_count": retry_count,
         }
 
         study.add_trial(
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -68,7 +68,7 @@ class RetryFailedTrialCallback:
 
     The failed trial can be identified by the
     :func:`~optuna.storages.RetryFailedTrialCallback.retried_trial_number` function.
-    Even if repetitive failure occurrs (a retried trial fails again),
+    Even if repetitive failure occurs (a retried trial fails again),
     this method returns the number of the original trial.
     To get a full list including the numbers of the retried trials as well as their original trial,
     call the :func:`~optuna.storages.RetryFailedTrialCallback.retry_history` function.
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -103,20 +103,15 @@ class RetryFailedTrialCallback:
         self._inherit_intermediate_values = inherit_intermediate_values
 
     def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
-        system_attrs = {"failed_trial": trial.number}
-
-        # Update the new object with the values in the trial.system_attrs.
-        # By doing this, if this failed try is already a rety, the 'failed_trial' value
-        # will be the first failed trial number.
-        system_attrs.update(trial.system_attrs)
-
-        retries = sum(
-            ("failed_trial", system_attrs["failed_trial"]) in t.system_attrs.items()
-            for t in study.trials
-        )
-
-        if self._max_retry is not None and retries + 1 > self._max_retry:
+        retry_count = trial.system_attrs.get('retry_count', 0) + 1
+        if self._max_retry is not None and retry_count > self._max_retry:
             return
+
+        system_attrs = {
+            **trial.system_attrs,
+            "failed_trial": trial.number,
+            "retry_count": retry_count
+        }
 
         study.add_trial(
             optuna.create_trial(
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -1,4 +1,5 @@
 from typing import Container
+from typing import List
 from typing import Optional
 
 import optuna
@@ -106,15 +107,11 @@ class RetryFailedTrialCallback:
         self._inherit_intermediate_values = inherit_intermediate_values
 
     def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
-        retry_count = trial.system_attrs.get("retry_count", 0) + 1
-        if self._max_retry is not None and retry_count > self._max_retry:
-            return
-
-        system_attrs = {
-            **trial.system_attrs,
-            "failed_trial": trial.number,
-            "retry_count": retry_count,
-        }
+        system_attrs = {"failed_trial": trial.number, "retry_history": [], **trial.system_attrs}
+        system_attrs["retry_history"].append(trial.number)  # type: ignore
+        if self._max_retry is not None:
+            if self._max_retry < len(system_attrs["retry_history"]):  # type: ignore
+                return
 
         study.add_trial(
             optuna.create_trial(
@@ -132,7 +129,7 @@ class RetryFailedTrialCallback:
     @staticmethod
     @experimental("2.8.0")
     def retried_trial_number(trial: FrozenTrial) -> Optional[int]:
-        """Return the number of the trial being retried.
+        """Return the number of the original trial being retried.
 
         Args:
             trial:
@@ -144,3 +141,21 @@ class RetryFailedTrialCallback:
         """
 
         return trial.system_attrs.get("failed_trial", None)
+
+    @staticmethod
+    @experimental("2.11.0")
+    def retry_history(trial: FrozenTrial) -> List[int]:
+        """Return the list of trial IDs from the original trial to the specified one.
+
+        Args:
+            trial:
+                The trial object.
+
+        Returns:
+            A list of trial IDs in ascending order of the series of retried trials.
+            The first item of the list indicates the original trial which is identical
+            to the :func:`~optuna.storage.RetryFailedTrialCallback.retried_trial_number`,
+            and the last item is the one right before the specified trial in the retry series.
+            If the specified trial is not a retry of any trial, returns an empty list.
+        """
+        return trial.system_attrs.get("retry_history", [])
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -145,14 +145,14 @@ class RetryFailedTrialCallback:
     @staticmethod
     @experimental("3.0.0")
     def retry_history(trial: FrozenTrial) -> List[int]:
-        """Return the list of retried trial IDs with respect to the specified trial.
+        """Return the list of retried trial numbers with respect to the specified trial.
 
         Args:
             trial:
                 The trial object.
 
         Returns:
-            A list of trial IDs in ascending order of the series of retried trials.
+            A list of trial numbers in ascending order of the series of retried trials.
             The first item of the list indicates the original trial which is identical
             to the :func:`~optuna.storage.RetryFailedTrialCallback.retried_trial_number`,
             and the last item is the one right before the specified trial in the retry series.
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -136,7 +136,7 @@ class RetryFailedTrialCallback:
                 The trial object.
 
         Returns:
-            The number of the latest failed trial. If not retry of a previous trial,
+            The number of the first failed trial. If not retry of a previous trial,
             returns :obj:`None`.
         """
 
--- a/optuna/_callbacks.py
+++ b/optuna/_callbacks.py
@@ -65,12 +65,16 @@ class RetryFailedTrialCallback:
 
     When a trial fails, this callback can be used with the :class:`optuna.storage` class to
     recreate the trial in :obj:`TrialState.WAITING` to queue up the trial to be run again.
-    In case the retried trial fails again and is to be retried further,
-    the :func:`~optuna.storage.RetryFailedTrialCallback.retried_trial_number` of the next
-    recreated trial will point to the latest failed one.
 
-    This is helpful in environments where trials may fail due to external conditions, such as
-    being preempted by other processes.
+    The failed trial can be identified by the
+    :func:`~optuna.storage.RetryFailedTrialCallback.retried_trial_number` function.
+    Even if repetitive failure occurrs (a retried trial fails again),
+    this method returns the number of the original trial.
+    To get a full list including the numbers of the retried trials as well as their original trial,
+    call the :func:`~optuna.storage.RetryFailedTrialCallback.retry_history` function.
+
+    This callback is helpful in environments where trials may fail due to external conditions,
+    such as being preempted by other processes.
 
     Usage:
 
--- a/tests/storages_tests/test_storages.py
+++ b/tests/storages_tests/test_storages.py
@@ -1361,20 +1361,20 @@ def test_retry_failed_trial_callback_repetitive_failure(storage_mode: str) -> No
         assert len(trials) == n_trials + 1
 
         assert "failed_trial" not in trials[0].system_attrs
-        assert "retry_count" not in trials[0].system_attrs
+        assert "retry_history" not in trials[0].system_attrs
 
-        # trial 1-3 are retried ones of the trial 0
+        # trial 1-3 are retried ones originated from the trial 0
         assert trials[1].system_attrs["failed_trial"] == 0
-        assert trials[1].system_attrs["retry_count"] == 1
+        assert trials[1].system_attrs["retry_history"] == [0]
 
-        assert trials[2].system_attrs["failed_trial"] == 1
-        assert trials[2].system_attrs["retry_count"] == 2
+        assert trials[2].system_attrs["failed_trial"] == 0
+        assert trials[2].system_attrs["retry_history"] == [0, 1]
 
-        assert trials[3].system_attrs["failed_trial"] == 2
-        assert trials[3].system_attrs["retry_count"] == 3
+        assert trials[3].system_attrs["failed_trial"] == 0
+        assert trials[3].system_attrs["retry_history"] == [0, 1, 2]
 
         # trial 4~ are the newly started one and its retry after exceededing max_retry
         assert "failed_trial" not in trials[4].system_attrs
-        assert "retry_count" not in trials[4].system_attrs
+        assert "retry_history" not in trials[4].system_attrs
         assert trials[5].system_attrs["failed_trial"] == 4
-        assert trials[5].system_attrs["retry_count"] == 1
+        assert trials[5].system_attrs["retry_history"] == [4]
--- a/tests/storages_tests/test_storages.py
+++ b/tests/storages_tests/test_storages.py
@@ -1378,11 +1378,3 @@ def test_retry_failed_trial_callback_repetitive_failure(storage_mode: str) -> No
         assert 'retry_count' not in trials[4].system_attrs
         assert trials[5].system_attrs['failed_trial'] == 4
         assert trials[5].system_attrs['retry_count'] == 1
-
-        # Current behavior                                                                    
-        # assert 'failed_trial' not in trials[0].system_attrs                                 
-        # assert trials[1].system_attrs['failed_trial'] == 0                                  
-        # assert trials[2].system_attrs['failed_trial'] == 0                                  
-        # assert trials[3].system_attrs['failed_trial'] == 0 
-        # assert 'failed_trial' not in trials[4].system_attrs                                 
-        # assert trials[5].system_attrs['failed_trial'] == 4                                  
--- a/tests/storages_tests/test_storages.py
+++ b/tests/storages_tests/test_storages.py
@@ -1332,3 +1332,57 @@ def test_read_trials_from_remote_storage(storage_mode: str) -> None:
 
         study_id = storage.create_new_study()
         storage.read_trials_from_remote_storage(study_id)
+
+
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES_HEARTBEAT)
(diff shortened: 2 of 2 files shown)


not522:
Thank you for your suggestion!

The current implementation assumes to use `retried_trial_number` instead of `trial.number`. Please see the PyTorch example. https://github.com/optuna/optuna-examples/blob/29720558a2bac9e0bead3a237640f4013f4ee2bd/pytorch/pytorch_checkpoint.py#L92-L105

The objective function cannot read other trials information. Therefore it will simplify user code by gathering one optimization history to one directory.

Do you have some specific cases where the last trial number is better than the first one?

HideakiImamura:
Let me note that this PR might be related to #2993.

hvy:
Just for the record, similar discussions has been brought up before (with some alternative ideas, left open as TODOs). https://github.com/optuna/optuna/pull/2694#issuecomment-849323590

belltailjp:
I see what @not522 points out. I wasn't aware of that example...
This example assumes that the retried training results are saved into the same directory as the original trial.
My deep learning use-case, which is originally told by @toshihikoyanase, was to separate working directory for each trial even if it's a retried trial, which looks like:

```python
def objective(trial):
    out_dir = os.path.join(args.out, f"trial_{trial.number:03d}")

    if RetryFailedTrialCallback.retried_trial_number(trial) is not None:
        preempted_trial = RetryFailedTrialCallback.retried_trial_number(trial)
        os.makedirs(out_dir, exist_ok=True)
        try:
            failed_snapshot = os.path.join(args.out, f"trial_{preempted_trial:03d}", "snapshot")
            new_snapshot = os.path.join(out_dir, "snapshot")
            shutil.copy(failed_snapshot, new_snapshot)

    # Run training
    # and save results and snapshots into out_dir
    # Since there is a (copied) snapshot in out_dir, it resumes training from it.
```
Basically both approaches can realize the same thing.
Given that, the focus of discussion would be weather we want each trial to have the same working directory or separated.
To my use-case, honestly I'm getting to think whichever is fine. It may feel a little unnatural if the working directory of a previously failed and thus stopped trial continues to be updated. Since a failed trial side itself has no info indicating that this is retried by other trial(s), it may also possibly confuse me to see the inconsistency between DB status of the trial and the actual working directory, until finding out it comes from retrials. But it basically sounds to be a preference matter.

HideakiImamura:
Thank you for raising this important issue. Personally, I agree with your opinion. It is reasonable to keep snapshots of failed trials and snapshots of re-tried trials separate.

By the way, isn't it possible to manage snapshots of failed trials and snapshots of re-tried trials separately in the current master implementation? I am not sure how changing it to retry the last trial instead of the first trial would solve the problem.

belltailjp:
To deal with deep learning intermediate snapshots, maybe there are basically 2 approaches; The first one would be to save new snapshots every certain number of epochs/iterations during training as separeted files, and thus we need to find out which file is the latest snapshot when resuming training. The second one is to keep only the latest snapshot by typically just overwriting to the same file.
Both [my previous comment](https://github.com/optuna/optuna/pull/3223#issuecomment-1009708092) and [Optuna example](https://github.com/optuna/optuna-examples/blob/v3.0.0-a0/pytorch/pytorch_checkpoint.py#L92-L101) assume the former.
When using the Optuna example approach with the current `RetryFailedTrialCallback` behavior, retried trials have to re-use the working directory of the original trial (because each retry can only know the number to the original trial), so it overwrites to the same snapshot file, meaning that snapshots saved right before the failure will be immediately lost (unless we explicitly implement something like back-up logic).
In the proposed change of `RetryFailedTrialCallback` behavior in this PR, it can keep the snapshots saved right before failure in the working directory of each trial, because each retry can have its own independent working directory and **it knows what trial (which directory) has the very latest snapshot** by looking at `failed_trial` attribute.

So, if the question

> isn't it possible to manage snapshots of failed trials and snapshots of re-tried trials separately

means whether we can keep the last snapshots of failed trials without overwritten by retried trials with the current `RetryFailedTrialCallback` behavior, the answer would be that maybe we have to back-up the latest snapshot as separated file every time a retried trial starts, which would look like:

```
trial_000
- model_epoch_xxx
- ...
- model_epoch_yyy
- latest_snapshot                   <----- Retry from this. Continuously overwitten for each step (epoch)
- latest_snapshot.failed.epoch_005  <----- If the trial=0 fails at epoch 5, next retried trial=1 copies (backs-up) the current latest_snapshot to this file
- latest_snapshot.failed.epoch_010  <----- Similarly, if trial=1 fails at 10, retry=2 makes it from the latest_snapshot right after it starts.
```

With this PR, it can be like:
```
trial_000
- model_epoch_000
- ...
- model_epoch_005
- last_snapshot           <---- snapshot after epoch 5 (at failure). Won't be overwritten anymore
trial_001
- model_epoch_006
- ...
- model_epoch_010
- last_snapshot           <---- snapshot after epoch 10
...
```

belltailjp:
Since it sounds more like a matter of preference, how about providing features to support both types of users mentioned so far?

Current
- `RetryFailedTrialCallback.failed_trial` -> ID of the very first (original) trial

Suggestion: Add some new attributes to the `RetryFailedTrialCallback`.
- `RetryFailedTrialCallback.original_failed_trial` -> ID of the original trial (equivalent to the *current (master)* `failed_trial`)
- `RetryFailedTrialCallback.failed_trial` -> ID of the trial that failed right before this trial (equivalent to the `failed_trial` proposed in this PR)
- `RetryFailedTrialCallback.retry_count` -> Optional, but maybe a user finds it's sometimes useful

The naming can be anything. E.g., to keep compatibility with current codes, keep `failed_trial` as it is, and name `last_failed_trial` as the proposed one, though I kinda feel the naming in the above listing is more clear (not sure).

HideakiImamura:
Sorry for the late reply. Thank you for the detailed explanation. I understand that this PR changes the behavior of the currently provided PyTorch + Heartbeat example, but it is reasonable.

I think we don't have to add a new attribute like `original_failed_trial`. As you originally suggested, it is reasonable to change the behavior of `failed_trial` to return the latest failed trial.

optuna/_callbacks.py
```diff
@@ -97,21 +100,16 @@ def __init__(self, max_retry: Optional[int] = None) -> None:
...
+        retry_count = trial.system_attrs.get("retry_count", 0) + 1
+        if self._max_retry is not None and retry_count > self._max_retry:
             return
 
+        system_attrs = {
+            **trial.system_attrs,
+            "failed_trial": trial.number,
+            "retry_count": retry_count,
+        }
+
```
HideakiImamura, line 112: In the distributed case, if `retry_count` is referenced and updated in the following order, consistency may be lost.
Do you have any idea to resolve it? I'm not sure but introducing the lock will resolve it, or removing the new attribute `retru_count` will do.
```
process A: -----> R  --------------> U -------------->
process B: --------------> R --------------> U ------>
(R: reference the `retry_count` from RDB and U: update the `retry_count` in RDB)
```

belltailjp: In my understanding the retry process should be run only in the master process in DDP setup.
https://github.com/optuna/optuna-examples/blob/29720558a2bac9e0bead3a237640f4013f4ee2bd/pytorch/pytorch_distributed_simple.py#L179-L187

More specifically, `study.optimize` calls [`fail_stale_trials()`](https://github.com/optuna/optuna/blob/v3.0.0-a0/optuna/study/_optimize.py#L192) -> [_heartbeat](https://github.com/optuna/optuna/blob/v3.0.0-a0/optuna/storages/_heartbeat.py#L27-L32) -> eventually reaches to `RetryFailedTrialCallback`, but as in the example `optimize()` is called only from the master, so basically we don't have to concern about updating information in an RDB of the same trial from multiple processes.
(If we call `optimize` in each worker, they'll have different params suggested)

When referring these values (`retry_count`, `failed_trial` etc) at the beginning of each trial as follows,
https://github.com/optuna/optuna-examples/blob/main/pytorch/pytorch_checkpoint.py#L87
[`TorchDistributedTrial.system_attrs`](https://github.com/optuna/optuna/blob/v3.0.0-a0/optuna/integration/pytorch_distributed.py#L224) gathers them from the master process through MPI broadcasting instead of directly accessing to DB, and MPI automatically synchronizes all the workers when a communication occurs, so I think there's basically no risk of inconsistency no matter the timing of reference.

HideakiImamura: Thank you for the fruitful discussion. I understand the current code does not cause any inconsistency if we call the callback only in the master process. 

Should we add more documentation to the `RetryFailedTrialsCallback` which cannot be called in the multiple processes?

belltailjp: Well I don't think so, because it'll never happen unless we try to call `study.optimize()` from multiple worker processes (in my understanding this is the case in MPI setup generally, but would there be any exceptions?).
So my opinion would be that it's rather worth mentioning it in the documentation of `Study`, maybe like "do not call `study.optimize()` in child workers because it'll make each worker behave inconsistently (due to being suggested different params worker-wise)", maybe with a reference to the [official example pytorch_distributed_simple.py](https://github.com/optuna/optuna-examples/blob/main/pytorch/pytorch_distributed_simple.py).
https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html

For users I think this should be OK, but for developers who works on integrating a new tool that uses MPI stuff to Optuna, I guess we need to somehow make them sure to follow how `TorchDistributedTrial` / `ChainerMNTrial` are implemented especially around their `system_attrs`.

HideakiImamura: If you have multiple distributed processes processing a single trial, I don't think it will be a problem as you mentioned. On the other hand, if multiple distributed processes are processing different trials, there may be inconsistencies in referencing and updating the retry_count, right? Since it is often the case in Optuna that distributed processes handle separate trials that do not communicate with each other using MPI, etc., it is likely that this callback function will be called for a failed trial at the same time in some processes.

HideakiImamura: I'm still wondering if the retry_count inconsistency can occur under normal Optuna workloads. I'm sorry if I misunderstood something.

belltailjp: OK I see what you meant!
When running multiple trials in child worker processes separately without co-operating each other (maybe using MPI just as a bulk launcher, right? - this is equivalent to [Easy Parallelization](https://optuna.readthedocs.io/en/v3.0.0-a0/tutorial/10_key_features/004_distributed.html)), each process is in charge of an independent trial, meaning that each worker looks at the different DB entry, there will be no conflict.

not522:
I am conservative about this change because it is difficult to reuse the training history in safety. If the retrying trial terminated before saving the first snapshot, we lost all the training history and should restart from 0 epoch.

How about adding `failed_trials`, which returns all the failed trials? I think it is better than returning the last failed trial.

Anyway, it is necessary to ensure that the change will not break the PyTorch example.

belltailjp:
> If the retrying trial terminated before saving the first snapshot, we lost all the training history and should restart from 0 epoch.

I didn't have this viewpoint! Thanks for point it out.
In my [expected usage](https://github.com/optuna/optuna/pull/3223#issuecomment-1009708092), it'll first copy the latest snapshot(s) right after a retry trial has started. So unless the retried process is immediately killed again (say, in several seconds after it starts) during this initial copy, practically the concerned situation will less likely to happen. But right, there's still a possibility.

Having a full list of previous failed trial IDs in `failed_trials` seems to be good as it can cover any types of usage while making it relatively easier to implement a way to mitigate the problem you pointed out. I proposed `original_failed_trial` and `failed_trial` in my previous comment for such a flexibility but it sounds that your suggestion is much simpler for users.

> necessary to ensure that the change will not break the PyTorch example

Let me just clarify that you mean [this example](https://github.com/optuna/optuna-examples/blob/main/pytorch/pytorch_checkpoint.py) and the example should work exactly as it is with completely no change (if so maybe for compatibility with various versions of optuna?), or you just meant we need to keep it properly synced with the latest optuna after merging a PR?

HideakiImamura:
- Is the new attribute `failed_trials` the entire trials that failed in the current trial? If so, does this mean that it is up to the user to decide which trial should be re-try?
- I think we should `keep it properly synced with the latest optuna after merging a PR`

toshihikoyanase:
Thank you for your PR! I just read the comments and the following comment by @HideakiImamura sounds reasonable to me.

> Is the new attribute `failed_trials` the entire trials that failed in the current trial? If so, does this mean that it is up to the user to decide which trial should be re-try?

I examined this idea a bit, so please take a look.

(Not sure, but `failed_trials` may be inconsistent with `retried_trial_number`. If we follow the naming convention, it can be `failed_trial_numbers` or something, but it may be confusing with `retried_trial_number`. So, let me call such an attribute as `retry_history` here.)

**Design**

Add `retry_history` that keeps all trials retried for the original failed trial. It returns a list of trial numbers, and the first element is identical to the value of `retried_trial_number`

The example implementation (diff from the master) is as follows:

```diff
@experimental("2.8.0")
class RetryFailedTrialCallback:
    def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
        system_attrs = {"failed_trial": trial.number}
+        if "retry_history" in system_attrs:
+            system_attrs["retry_history"].append(trial.number)
+        else:
+            system_attrs["retry_history"] = [trial.number]

...

+    @staticmethod
+    def retry_history(trial: FrozenTrial) -> List[int]:
+        return trial.system_attrs.get("retry_history", [])
```

**Use cases**

The example in https://github.com/optuna/optuna/pull/3223#issuecomment-1009708092

```python
def objective(trial):
    out_dir = os.path.join(args.out, f"trial_{trial.number:03d}")

    if RetryFailedTrialCallback.retried_trial_number(trial) is not None:
        preempted_trial = RetryFailedTrialCallback.retry_history(trial)[-1]  # <-- here
        os.makedirs(out_dir, exist_ok=True)
        try:
            failed_snapshot = os.path.join(args.out, f"trial_{preempted_trial:03d}", "snapshot")
            new_snapshot = os.path.join(out_dir, "snapshot")
            shutil.copy(failed_snapshot, new_snapshot)

```


**Pros**
- Keep backward compatibility since it does not change `failed_trial` The existing users does not need to update their code.
- The last retried trial can be obtained by `retry_history(trial)[-1]`.
- We do not need to manage retry counter.

**Cons**
- Increase the number of the attributes of `RetryFailedTrialCallback`
- Naming of `retry_history`. Since the first element is the original trial, not retried trial. It may not be intuitive for users.
- Increase the database tables to preserve `system_attrs["retry_history"]`

HideakiImamura:
@toshihikoyanase's suggestion sounds reasonable to me. What do you think? @belltailjp @not522

not522:
I agree with @toshihikoyanase's idea. The interface has room to discuss; `retry_history` return ints or FrozenTrials?

not522:
> So unless the retried process is immediately killed again (say, in several seconds after it starts) during this initial copy, practically the concerned situation will less likely to happen.

I think this is not a rare case. For example, copying the latest snapshot will fail when storage trouble or shortage causes the termination of the previous trial.

belltailjp:
I'm also fine with the suggested approach.

> retry_history return ints or `FrozenTrials`?

I guess it can be simply a list of int from the viewpoints of consistency with the existing `failed_trial` as well as efficiency.

For naming of the interfaces to be introduced, I'll totally respect whatever you would choose. IMO having the initial (original) trial ID in an attribute named `retry_history` doesn't feel too strange in terms of naming, so it's fine to me if there's nothing better. Would appropriate documentation sufficiently help users understanding it?

---
>> So unless the retried process is immediately killed again (say, in several seconds after it starts) during this initial copy, practically the concerned situation will less likely to happen.
>
> I think this is not a rare case. For example, copying the latest snapshot will fail when storage trouble or shortage causes the termination of the previous trial.

Right, I also understand it can happen. I think dealing with it should be on the user's responsibility. Giving options can still be valuable (while those who uses it should accept or deal with the concerned situation).

not522:
> I think this is not a rare case. For example, copying the latest snapshot will fail when storage trouble or shortage causes the termination of the previous trial.

I realized this is unordinary. Storage trouble makes the trial `FAIL`, not stale. 🙇‍♂️

HideakiImamura:
Thanks for the discussions @belltailjp and @not522. I agree with using the name of `release history` with sufficient documents and its type as a list of ints.

optuna/_callbacks.py
```diff
@@ -124,15 +119,33 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
         Returns:
-            The number of the first failed trial. If not retry of a previous trial,
+            The number of the latest failed trial. If not retry of a previous trial,
             returns :obj:`None`.
         """
 
         return trial.system_attrs.get("failed_trial", None)
+
+    @staticmethod
+    @experimental("2.11.0")
```
HideakiImamura, line 136: A nit.
```suggestion
    @experimental("3.0.0")
```

optuna/_callbacks.py
```diff
@@ -124,15 +119,33 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
+    def retry_history(trial: FrozenTrial) -> List[int]:
+        """Return the list of trial IDs from the original trial to the specified one.
```
HideakiImamura, line 138: How about the following? I think it's easy to understand.
```suggestion
        """Return the list of retried trial IDs with respect to the specified trial.
```

optuna/_callbacks.py
```diff
@@ -124,15 +119,33 @@ def __call__(self, study: "optuna.study.Study", trial: FrozenTrial) -> None:
...
-        """Return the number of the trial being retried.
+        """Return the number of the original trial being retried.
 
         Args:
             trial:
                 The trial object.
 
         Returns:
-            The number of the first failed trial. If not retry of a previous trial,
+            The number of the latest failed trial. If not retry of a previous trial,
```
HideakiImamura, line 129: Should it be reverted?
```suggestion
            The number of the first failed trial. If not retry of a previous trial,
```

HideakiImamura (commented):
Thanks for the update! I have several minor comments. PTAL.

optuna/_callbacks.py
```diff
@@ -136,3 +131,21 @@ def retried_trial_number(trial: FrozenTrial) -> Optional[int]:
...
+    @experimental("3.0.0")
+    def retry_history(trial: FrozenTrial) -> List[int]:
+        """Return the list of retried trial IDs with respect to the specified trial.
+
+        Args:
+            trial:
+                The trial object.
+
+        Returns:
+            A list of trial IDs in ascending order of the series of retried trials.
```
not522, line 145:
```suggestion
            A list of trial numbers in ascending order of the series of retried trials.
```

optuna/_callbacks.py
```diff
@@ -136,3 +131,21 @@ def retried_trial_number(trial: FrozenTrial) -> Optional[int]:
         """
 
         return trial.system_attrs.get("failed_trial", None)
+
+    @staticmethod
+    @experimental("3.0.0")
+    def retry_history(trial: FrozenTrial) -> List[int]:
+        """Return the list of retried trial IDs with respect to the specified trial.
```
not522, line 138: Trial id and trial number are different conceptions in Optuna.

```suggestion
        """Return the list of retried trial numbers with respect to the specified trial.
```

not522 (commented):
Thank you for the update! I added a nitpick comment.

Note: This change may introduce some breaking changes for studies done by previous versions of Optuna. But, I think it is not a serious impact.

belltailjp:
I've fixed both!

> breaking changes for studies done by previous versions of Optuna
> it is not a serious impact

I agree. (Let me just clarify that) What will happen in this case would be that the number of previous trials is simply underestimated, so it won't basically break users' experiments.

HideakiImamura (approved):
Thank you for the long running PR! LGTM!

optuna/_callbacks.py
```diff
@@ -63,6 +64,9 @@ class RetryFailedTrialCallback:
 
     When a trial fails, this callback can be used with the :class:`optuna.storage` class to
     recreate the trial in :obj:`TrialState.WAITING` to queue up the trial to be run again.
+    In case the retried trial fails again and is to be retried further,
+    the :func:`~optuna.storage.RetryFailedTrialCallback.retried_trial_number` of the next
+    recreated trial will point to the latest failed one.
```
belltailjp, line 69: @HideakiImamura Sorry I've found this part is incorrect. Please wait a bit if you're gonna merge this PR until I fix it.

belltailjp: Fixed!!!

HideakiImamura (approved):
LGTM again!

optuna/_callbacks.py
```diff
@@ -64,8 +65,15 @@ class RetryFailedTrialCallback:
     When a trial fails, this callback can be used with the :class:`optuna.storage` class to
     recreate the trial in :obj:`TrialState.WAITING` to queue up the trial to be run again.
 
-    This is helpful in environments where trials may fail due to external conditions, such as
-    being preempted by other processes.
+    The failed trial can be identified by the
+    :func:`~optuna.storage.RetryFailedTrialCallback.retried_trial_number` function.
```
not522, line 69:
```suggestion
    :func:`~optuna.storages.RetryFailedTrialCallback.retried_trial_number` function.
```

optuna/_callbacks.py
```diff
@@ -64,8 +65,15 @@ class RetryFailedTrialCallback:
...
+    Even if repetitive failure occurrs (a retried trial fails again),
+    this method returns the number of the original trial.
+    To get a full list including the numbers of the retried trials as well as their original trial,
+    call the :func:`~optuna.storage.RetryFailedTrialCallback.retry_history` function.
```
not522, line 73:
```suggestion
    call the :func:`~optuna.storages.RetryFailedTrialCallback.retry_history` function.
```

not522 (commented):
Basically LGTM. Please fix small document errors.

belltailjp:
Thanks for pointing them out! I have updated.

not522 (approved):
LGTM. Thank you for your quick fix! @toshihikoyanase Please merge it if you approve this PR.

optuna/_callbacks.py
```diff
@@ -64,8 +65,15 @@ class RetryFailedTrialCallback:
     When a trial fails, this callback can be used with the :class:`optuna.storage` class to
     recreate the trial in :obj:`TrialState.WAITING` to queue up the trial to be run again.
 
-    This is helpful in environments where trials may fail due to external conditions, such as
-    being preempted by other processes.
+    The failed trial can be identified by the
+    :func:`~optuna.storages.RetryFailedTrialCallback.retried_trial_number` function.
+    Even if repetitive failure occurrs (a retried trial fails again),
```
toshihikoyanase, line 70: A nit. Let me commit by myself.
```suggestion
    Even if repetitive failure occurs (a retried trial fails again),
```

tests/storages_tests/test_storages.py
```diff
@@ -1274,3 +1274,49 @@ def test_read_trials_from_remote_storage(storage_mode: str) -> None:
...
+
+    with StorageSupplier(
+        storage_mode,
+        heartbeat_interval=heartbeat_interval,
+        grace_period=grace_period,
+        failed_trial_callback=RetryFailedTrialCallback(max_retry=max_retry),
+    ) as storage:
+        study = optuna.create_study(storage=storage)
+
+        # Make repeatedly failed and retried trials by heartbeat
```
toshihikoyanase, line 1294: A nit, but let me slightly fix the comment to follow the [code conventions](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#block-and-inline-comments-are-supposed-to-be-sentences).
```suggestion
        # Make repeatedly failed and retried trials by heartbeat.
```

tests/storages_tests/test_storages.py
```diff
@@ -1274,3 +1274,49 @@ def test_read_trials_from_remote_storage(storage_mode: str) -> None:
...
+            optuna.storages.fail_stale_trials(study)
+
+        trials = study.trials
+
+        assert len(trials) == n_trials + 1
+
+        assert "failed_trial" not in trials[0].system_attrs
+        assert "retry_history" not in trials[0].system_attrs
+
+        # trial 1-3 are retried ones originated from the trial 0
```
toshihikoyanase, line 1308: ditto.
```suggestion
        # Trial 1-3 are retried ones originated from the trial 0.
```

tests/storages_tests/test_storages.py
```diff
@@ -1274,3 +1274,49 @@ def test_read_trials_from_remote_storage(storage_mode: str) -> None:
...
+        assert trials[1].system_attrs["failed_trial"] == 0
+        assert trials[1].system_attrs["retry_history"] == [0]
+
+        assert trials[2].system_attrs["failed_trial"] == 0
+        assert trials[2].system_attrs["retry_history"] == [0, 1]
+
+        assert trials[3].system_attrs["failed_trial"] == 0
+        assert trials[3].system_attrs["retry_history"] == [0, 1, 2]
+
+        # trial 4~ are the newly started one and its retry after exceededing max_retry
```
toshihikoyanase, line 1318: Sorry for bothering you, but I'm not sure the phrase of `its retry after ...`. I suggested the update. So, could you check it, please?
```suggestion
        # Trial 4 and later are the newly started ones and they are retried after exceeding max_retry.
```

HideakiImamura:
The suggestions by @toshihikoyanase look good to me. Could you apply the suggestions @belltailjp?

not522:
LGTM again. Thank you!

HideakiImamura (approved):
Thanks for the update! LGTM again!

HideakiImamura:
@toshihikoyanase Please merge this PR after your approval.

HideakiImamura:
@belltailjp Could you fix the conflict? Sorry for the late response. I asked yanase-san and he said we can merge this PR with 2 approvals, so I'll merge this PR soon after you resolve the conflict.

HideakiImamura:
Thanks for the long running PR! Let me merge this PR.

PR #3233, opened 2022-01-17
Unify the validation function of edf value between visualization backends

nzw0301:
## Motivation

Test files of `plot_edf` define validation function for edf values. The implementations between Plotly and Matplotlib are identical, but the function name and its argument name are slightly different.

## Description of the changes

Make the function consistent by using Plotly's definition:

https://github.com/optuna/optuna/blob/acdf4b75d5475d17dae9752b5ad2885a8132894d/tests/visualization_tests/test_edf.py#L17

Another approach might be to import plotly's `_validate_edf_values` for `optuna/tests/visualization_tests/matplotlib_tests/test_edf.py`.

### DIFF
--- a/tests/visualization_tests/matplotlib_tests/test_edf.py
+++ b/tests/visualization_tests/matplotlib_tests/test_edf.py
@@ -14,8 +14,8 @@ def test_target_is_none_and_study_is_multi_obj() -> None:
         plot_edf(study)
 
 
-def _validate(y_values: Sequence[float]) -> None:
-    np_values = np.array(y_values)
+def _validate_edf_values(edf_values: Sequence[float]) -> None:
+    np_values = np.array(edf_values)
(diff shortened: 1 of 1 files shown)


himkt:
I'm wondering if it'd be good to extract these test utilities into `optuna.testing.visualization`.

nzw0301:
Could be? I'm also worried about enlarging optuna's main codebase; `optuna.testing.visualization`, especially, `_validate_edf_values`, is not necessary for users...

toshihikoyanase:
@himkt Could you continue the review, please?

himkt:
> I'm also worried about enlarging optuna's main codebase

OK. 👍

If we decided to introduce a new backend for visualization, it would be a good idea to rethink about creating a utility function, I think.

himkt (approved):

toshihikoyanase (approved):
LGTM!

PR #3244, opened 2022-01-21
Make existing int/float distributions wrapper of {Int,Float}Distribution

himkt:
This PR make existing distributions (except for categorical) subclasses of `FloatDistribution` and `IntDistribution`, which was introduced in https://github.com/optuna/optuna/pull/3063.

## Motivation
In v3.0.0, Optuna will aggregate distributions (except for `CategoricalDistribution`) into `FloatDistribution` and `IntDistribution`. Existing distributions such as `UniformDistribution` and `IntUniformDistribution` will be kept until v5.0.0 for backward compatibilities but these will be changed to be subclasses of `FloatDistribution` and `IntDistribution`.

An entire transition process will be done in https://github.com/optuna/optuna/pull/3246 but changing class architectures can be introduced in advance, which makes codebase simple that removes `if-then` rules for old distributions. For example, sampler implementation will change as following:

```diff
diff --git a/optuna/integration/skopt.py b/optuna/integration/skopt.py
index 81cc5bf3..c76a9ebe 100644
--- a/optuna/integration/skopt.py
+++ b/optuna/integration/skopt.py
@@ -243,22 +243,7 @@ class _Optimizer(object):

         dimensions = []
         for name, distribution in sorted(self._search_space.items()):
-            if isinstance(distribution, distributions.UniformDistribution):
-                # Convert the upper bound from exclusive (optuna) to inclusive (skopt).
-                high = np.nextafter(distribution.high, float("-inf"))
-                dimension = space.Real(distribution.low, high)
-            elif isinstance(distribution, distributions.LogUniformDistribution):
-                # Convert the upper bound from exclusive (optuna) to inclusive (skopt).
-                high = np.nextafter(distribution.high, float("-inf"))
-                dimension = space.Real(distribution.low, high, prior="log-uniform")
-            elif isinstance(distribution, distributions.IntUniformDistribution):
-                count = (distribution.high - distribution.low) // distribution.step
-                dimension = space.Integer(0, count)
-            elif isinstance(distribution, distributions.IntLogUniformDistribution):
-                low = distribution.low - 0.5
-                high = distribution.high + 0.5
-                dimension = space.Real(low, high, prior="log-uniform")
-            elif isinstance(distribution, distributions.IntDistribution):
+            if isinstance(distribution, distributions.IntDistribution):
                 if distribution.log:
                     low = distribution.low - 0.5
                     high = distribution.high + 0.5
@@ -266,9 +251,6 @@ class _Optimizer(object):
                 else:
                     count = (distribution.high - distribution.low) // distribution.step
                     dimension = space.Integer(0, count)
-            elif isinstance(distribution, distributions.DiscreteUniformDistribution):
-                count = int((distribution.high - distribution.low) // distribution.q)
-                dimension = space.Integer(0, count)
             elif isinstance(distribution, distributions.CategoricalDistribution):
                 dimension = space.Categorical(distribution.choices)
             elif isinstance(distribution, distributions.FloatDistribution):
```

## Description of the changes
- a9ddf6d update existing distributions to be wrappers of {Int,Float}Distribution
- 6cb654c remove tests for IntLogUniformDistribution with step larger than 1
- e06229e fix if-then rules to avoid multiple value transformation
- d3d61b8 update tests

#2941

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -15,7 +15,7 @@ CategoricalChoiceType = Union[None, bool, int, float, str]
 
 
 _float_distribution_deprecated_msg = "Use :class:`~optuna.distributions.FloatDistribution` instead."
-_int_distribution_deprecated_msg = "Use :class:`~optuna.distributions.IntDistribution` instead"
+_int_distribution_deprecated_msg = "Use :class:`~optuna.distributions.IntDistribution` instead."
 
 
 class BaseDistribution(object, metaclass=abc.ABCMeta):
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -178,7 +178,7 @@ class FloatDistribution(BaseDistribution):
             return self.low <= value <= self.high and abs(k - round(k)) < 1.0e-8
 
 
-@deprecated("3.0.0", "5.0.0", text=_float_distribution_deprecated_msg)
+@deprecated("3.0.0", "6.0.0", text=_float_distribution_deprecated_msg)
 class UniformDistribution(FloatDistribution):
     """A uniform distribution in the linear domain.
 
@@ -210,7 +210,7 @@ class UniformDistribution(FloatDistribution):
         return "{}({})".format(self.__class__.__name__, kwargs)
 
 
-@deprecated("3.0.0", "5.0.0", text=_float_distribution_deprecated_msg)
+@deprecated("3.0.0", "6.0.0", text=_float_distribution_deprecated_msg)
 class LogUniformDistribution(FloatDistribution):
     """A uniform distribution in the log domain.
 
@@ -243,7 +243,7 @@ class LogUniformDistribution(FloatDistribution):
         return "{}({})".format(self.__class__.__name__, kwargs)
 
 
-@deprecated("3.0.0", "5.0.0", text=_float_distribution_deprecated_msg)
+@deprecated("3.0.0", "6.0.0", text=_float_distribution_deprecated_msg)
 class DiscreteUniformDistribution(FloatDistribution):
     """A discretized uniform distribution in the linear domain.
 
@@ -363,7 +363,7 @@ class IntDistribution(BaseDistribution):
         return self.low <= value <= self.high and (value - self.low) % self.step == 0
 
 
-@deprecated("3.0.0", "5.0.0", text=_int_distribution_deprecated_msg)
+@deprecated("3.0.0", "6.0.0", text=_int_distribution_deprecated_msg)
 class IntUniformDistribution(IntDistribution):
     """A uniform distribution on integers.
 
@@ -403,7 +403,7 @@ class IntUniformDistribution(IntDistribution):
         return "{}({})".format(self.__class__.__name__, kwargs)
 
 
-@deprecated("3.0.0", "5.0.0", text=_int_distribution_deprecated_msg)
+@deprecated("3.0.0", "6.0.0", text=_int_distribution_deprecated_msg)
 class IntLogUniformDistribution(IntDistribution):
     """A uniform distribution on integers in the log domain.
 
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -14,7 +14,9 @@ from optuna._deprecated import deprecated
 CategoricalChoiceType = Union[None, bool, int, float, str]
 
 
-_float_distribution_deprecated_msg = "Use :class:`~optuna.distributions.FloatDistribution` instead."
+_float_distribution_deprecated_msg = (
+    "Use :class:`~optuna.distributions.FloatDistribution` instead."
+)
 _int_distribution_deprecated_msg = "Use :class:`~optuna.distributions.IntDistribution` instead."
 
 
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -14,7 +14,7 @@ from optuna._deprecated import deprecated
 CategoricalChoiceType = Union[None, bool, int, float, str]
 
 
-_float_distribution_deprecated_msg = "Use :class:`~optuna.distributions.FloatDistribution` instead"
+_float_distribution_deprecated_msg = "Use :class:`~optuna.distributions.FloatDistribution` instead."
 _int_distribution_deprecated_msg = "Use :class:`~optuna.distributions.IntDistribution` instead"
 
 
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -3,6 +3,7 @@ import copy
 import decimal
 import json
 from typing import Any
+from typing import cast
 from typing import Dict
 from typing import Sequence
 from typing import Union
@@ -288,7 +289,7 @@ class DiscreteUniformDistribution(FloatDistribution):
 
     @property
     def q(self) -> float:
-        return self.step
+        return cast(float, self.step)
 
 
 class IntDistribution(BaseDistribution):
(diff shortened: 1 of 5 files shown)


optuna/integration/skopt.py
```diff
@@ -323,15 +323,15 @@ def ask(self) -> Dict[str, Any]:
                 value = float(value)
             if isinstance(distribution, distributions.DiscreteUniformDistribution):
                 value = value * distribution.q + distribution.low
-            if isinstance(distribution, distributions.IntUniformDistribution):
```
himkt, line 326: A distribution that is an instance of `IntUniformDistribution` will be `IntUniformDistribution` and `IntDistribution` so we have to unify the if-then rules.

tests/test_distributions.py
```diff
@@ -337,17 +330,6 @@ def test_contains() -> None:
     assert not ilu._contains(12.1)
     assert not ilu._contains(13)
 
-    # `step` is ignored and assumed to be 1.
-    iluq = distributions.IntLogUniformDistribution(low=2, high=7, step=2)
```
himkt, line 341: `IntDistribution` doesn't allow `step` to be larger than `1`. The current `IntLogUniformDistribution` tolerates such step and replace it with `1` internally.

https://github.com/optuna/optuna/pull/3244/files#diff-2ff77057715ff5d7b689aa1ba37562d8a47052b826e0267493bfdd13ac7206e3L515-L522

This behavior is scheduled to be changed in v4.0.0 but we also stated that it is subject to change.
The impact of this change to users would be small and I think the current behavior is misleading.
I'd like to propose to change in this PR.

himkt:
I'm sure that @xadrianzetx are familiar with this change so I'd like your review! :D

optuna/distributions.py
```diff
@@ -280,36 +270,20 @@ class DiscreteUniformDistribution(BaseDistribution):
...
-
-        high = _adjust_discrete_uniform_high(low, high, q)
-
-        self.low = float(low)
-        self.high = float(high)
-        self.q = float(q)
-
-    def single(self) -> bool:
+        super().__init__(low=low, high=high, step=q)
+        self.q = q
```
xadrianzetx, line 274: How about making this a property?
```python
@property
def q(self) -> int:
    return self.step
```

tests/test_distributions.py
xadrianzetx: :+1:

optuna/integration/skopt.py
xadrianzetx: :+1:

Note - branching on old distributions will be removed in follow-ups.

xadrianzetx (commented):
Thanks for the PR and inviting me to do this review! I like the idea, but left some comments on implementation.

tests/trial_tests/test_trial.py
```diff
@@ -154,8 +154,10 @@ def test_check_distribution_suggest_int(storage_mode: str, enable_log: bool) ->
             trial.suggest_int("x", 10, 20, log=enable_log)
             trial.suggest_int("x", 10, 22, log=enable_log)
 
-        # We expect exactly one warning.
-        assert len(record) == 1
+        # We expect exactly four warnings.
+        # One is RuntimeWarning for inconsistent parameter specification.
+        # Three are FutureWarning for deprecated distributions.
+        assert len(record) == 4
```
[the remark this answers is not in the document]
himkt: It looks reasonable! Thank you. 📝

optuna/distributions.py
```diff
@@ -172,7 +178,8 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
             return self.low <= value <= self.high and abs(k - round(k)) < 1.0e-8
 
 
-class UniformDistribution(BaseDistribution):
+@deprecated("3.0.0", "5.0.0", text=_float_distribution_deprecated_msg)
```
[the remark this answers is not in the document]
himkt: It makes sense. 65aa381 changes deprecation periods for them. ☝️

optuna/distributions.py
himkt: It looks nice! I applied your suggestion in e4727cd. 👍

xadrianzetx (approved):
LGTM! Feel free to resolve conversations as I can't do it :neutral_face:

hvy:
@keisuke-umezawa do you mind jumping in as a reviewer for this change?

tests/test_distributions.py
HideakiImamura: 👍

optuna/distributions.py
```diff
@@ -190,27 +200,21 @@ class UniformDistribution(BaseDistribution):
...
+        d = copy.deepcopy(self.__dict__)
+        d.pop("log")
+        d.pop("step")
+        return d
 
-        value = param_value_in_internal_repr
-        return self.low <= value <= self.high
+    def __repr__(self) -> str:
+        kwargs = ", ".join("{}={}".format(k, v) for k, v in sorted(self._asdict().items()))
+        return "{}({})".format(self.__class__.__name__, kwargs)
```
HideakiImamura, line 213: I'm not sure but if we change the `BaseDistribution`'s `__repr__` to use `self._asdict`, would it not be necessary to implement `__repr__` for `UniformDistribution`, etc.?

optuna/distributions.py
```diff
@@ -416,43 +397,20 @@ class IntUniformDistribution(BaseDistribution):
...
 
-        value = param_value_in_internal_repr
-        return self.low <= value <= self.high and (value - self.low) % self.step == 0
+    def __repr__(self) -> str:
+        kwargs = ", ".join("{}={}".format(k, v) for k, v in sorted(self._asdict().items()))
+        return "{}({})".format(self.__class__.__name__, kwargs)
 
 
-class IntLogUniformDistribution(BaseDistribution):
+@deprecated("3.0.0", "6.0.0", text=_int_distribution_deprecated_msg)
```
HideakiImamura, line 412: There is no implementation of `__repr__` of `IntLogUniformDistribution`. Is it intended?

HideakiImamura (commented):
Thanks for the PR! Almost LGTM, but I have two minor comments. Please take a look.

optuna/distributions.py
himkt: I'm sorry but I don't understand what you mean here. I think `__repr__` is already implemented.

https://github.com/optuna/optuna/pull/3244/files#diff-2ff77057715ff5d7b689aa1ba37562d8a47052b826e0267493bfdd13ac7206e3R450-L508

optuna/distributions.py
himkt: It makes sense to me. I didn't do so because I focus only updating `__repr__` to fix failed tests but it would make codebase more simple. I revised the PR in 970d1c0.

@hvy 
Please give us a comment if you have. I mention you because you also have edited `__repr__` for `IntLogUniformDistribution`.
https://github.com/optuna/optuna/blob/bb0ad9a77e2c7715a7bddb82cc9db8ac321a39c7/optuna/distributions.py#L505-L509

hvy: Sorry, I cannot remember the exact details but I think your change in encapsulating the "formatting" in `_asdict` is what I meant. Comment was introduced in https://github.com/optuna/optuna/pull/1438. Just for the record, `_asdict` has an even longer history, https://github.com/optuna/optuna/pull/16.

HideakiImamura (approved):
Thanks for the update! LGTM!

HideakiImamura:
Let me merge this PR since we have two approvals of me and @xadrianzetx.

PR #3262, opened 2022-01-30
Make RedisStorage fast when running multiple trials

masap:
## Motivation
`RedisStorage` takes more time per trial as the number of trials increases.

## Description of the changes
This PR fixes this issue by combining multiple `get()` into a single `mget()`.
Even after applying this fix, `_get_trials()` is still O(N). This is because `mget()` is O(N) <https://redis.io/commands/mget>.

graph

### DIFF
--- a/optuna/storages/_redis.py
+++ b/optuna/storages/_redis.py
@@ -665,10 +665,16 @@ class RedisStorage(BaseStorage):
 
         self._check_study_id(study_id)
 
-        trials = []
+        queries = []
         trial_ids = set(self._get_study_trials(study_id)) - excluded_trial_ids
         for trial_id in trial_ids:
-            frozen_trial = self.get_trial(trial_id)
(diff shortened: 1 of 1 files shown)


HideakiImamura (approved):
Looks solid!

masap:
Thanks for reviewing !

himkt (commented):
The idea looks elegant for me!
Let me ask one question: is it possible to utilize `_get_trials` instead defining a logic here? (I think the idea behind this improvement is that reduces a number of queries to Redis, which looks the same as `_get_trials`)
https://github.com/optuna/optuna/blob/93057bf94b2eb9ac9a018b5e89822a70a26bf7f7/optuna/storages/_redis.py#L659-L682

himkt (approved):
Sorry, my comment is irrelevant. The changes looks good!

PR #3263, opened 2022-01-30
Add test for default grace period

masap:
## Motivation
The default `grace period` is not covered by any test. This PR add the test for it and improve coverage.

## Description of the changes
Add test for default `grace period`.

### DIFF
--- a/tests/storages_tests/test_storages.py
+++ b/tests/storages_tests/test_storages.py
@@ -1219,9 +1219,10 @@ def test_retry_failed_trial_callback(storage_mode: str, max_retry: Optional[int]
 
 
 @pytest.mark.parametrize("storage_mode", ["sqlite", "redis"])
-def test_fail_stale_trials(storage_mode: str) -> None:
+@pytest.mark.parametrize("grace_period", [None, 2])
+def test_fail_stale_trials(storage_mode: str, grace_period: Optional[int]) -> None:
     heartbeat_interval = 1
-    grace_period = 2
+    _grace_period = (heartbeat_interval * 2) if grace_period is None else grace_period
 
     def failed_trial_callback(study: "optuna.Study", trial: FrozenTrial) -> None:
         assert study.system_attrs["test"] == "A"
@@ -1238,7 +1239,7 @@ def test_fail_stale_trials(storage_mode: str) -> None:
         trial = study.ask()
         trial.set_system_attr("test", "B")
         storage.record_heartbeat(trial._trial_id)
-        time.sleep(grace_period + 1)
+        time.sleep(_grace_period + 1)
 
         assert study.trials[0].state is TrialState.RUNNING
 


tests/storages_tests/test_storages.py
```diff
@@ -1219,9 +1219,10 @@ def test_retry_failed_trial_callback(storage_mode: str, max_retry: Optional[int]
 
 
 @pytest.mark.parametrize("storage_mode", ["sqlite", "redis"])
-def test_fail_stale_trials(storage_mode: str) -> None:
+@pytest.mark.parametrize("grace_period", [None, 2])
+def test_fail_stale_trials(storage_mode: str, grace_period: int) -> None:
     heartbeat_interval = 1
-    grace_period = 2
+    _grace_period = 2
```
toshihikoyanase, line 1225: I guess this code is a bit fragile since it depends on the value of `heartbeat_interval` implicitly (i.e., `grace_period = 2 * heartbeat_interval` if `grace_period` is `None`).
How about specifying the value explicitly?

toshihikoyanase (commented):
Thank you for improving the test! I added some small comments.

tests/storages_tests/test_storages.py
```diff
@@ -1219,9 +1219,10 @@ def test_retry_failed_trial_callback(storage_mode: str, max_retry: Optional[int]
 
 
 @pytest.mark.parametrize("storage_mode", ["sqlite", "redis"])
-def test_fail_stale_trials(storage_mode: str) -> None:
+@pytest.mark.parametrize("grace_period", [None, 2])
+def test_fail_stale_trials(storage_mode: str, grace_period: int) -> None:
```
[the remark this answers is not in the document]
masap: Thanks for reviewing ! Ok, fixed.

tests/storages_tests/test_storages.py
masap: Ok, fixed.

tests/storages_tests/test_storages.py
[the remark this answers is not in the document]
toshihikoyanase: Thanks!

tests/storages_tests/test_storages.py
```diff
@@ -1238,7 +1238,7 @@ def failed_trial_callback(study: "optuna.Study", trial: FrozenTrial) -> None:
         trial = study.ask()
         trial.set_system_attr("test", "B")
         storage.record_heartbeat(trial._trial_id)
-        time.sleep(grace_period + 1)
+        time.sleep((heartbeat_interval * 2) + 1)
```
toshihikoyanase, line 1241: This line does not refer to the value of `grace_period`. So, if we add the case `grace_period=4`, the test will fail.

Could you use the value of `grace_period` if it is not `None`? The following is an example:
```python
_grace_period = (heartbeat_interval * 2) if grace_period is None else grace_period
time.sleep(_grace_period + 1)
```

toshihikoyanase (changes requested):
Thanks for your update. Could you check the comment on `time.sleep`?

tests/storages_tests/test_storages.py
masap: Ok, fixed.

toshihikoyanase (approved):
Looks great to me. Thank you for your swift actions!

nzw0301 (approved):
LGTM. Thanks!

PR #3268, opened 2022-01-30
Add missing cli methods to the list

nzw0301:
## Motivation

A few methods do not on the list on the tutorial page.

## Description of the changes

Add methods. Their descriptions come from their API pages.

### DIFF
--- a/tutorial/20_recipes/004_cli.py
+++ b/tutorial/20_recipes/004_cli.py
@@ -8,12 +8,17 @@ Command-Line Interface
    :header: Command, Description
    :widths: 20, 40
 
+    ask, Create a new trial and suggest parameters.
+    best-trial, Show the best trial.
+    best-trials, Show a list of trials located at the Pareto front.
     create-study, Create a new study.
     delete-study, Delete a specified study.
     storage upgrade, Upgrade the schema of a storage.
     studies, Show a list of studies.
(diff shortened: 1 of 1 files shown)


nzw0301:
By the way, can we provide method names with links on the first column of the table for usability?

tutorial/20_recipes/004_cli.py
```diff
@@ -17,7 +18,7 @@
     studies, Show a list of studies.
     study optimize, Start optimization of a study.
     study set-user-attr, Set a user attribute to a study.
-    tell, Finish a trial, which was created by the ask command.
+    tell, Finish a trial\\, which was created by the ask command.
```
nzw0301, line 21: Since `,` is used to create the table, I've added `\\` as an escape for the comma

toshihikoyanase (approved):
Good catch! Thank you!

toshihikoyanase:
> By the way, can we provide method names with links on the first column of the table for usability?

That's a good idea! Can we work on it in a follow-up PR to minimize the changes?

nzw0301:
thank you for your quick review! Adding links might not be simple. Let's address it as a followup.

toshihikoyanase:
That makes sense. Let me merge this PR.

PR #3272, opened 2022-01-31
Force implementation of `_check_and_set_param_distribution()` by `BaseStorage`

masap:
## Description of the changes
To make it easy to add new cacheable storage, this patch make the implementation of `_check_and_set_param_distribution()` mandatory.

This is requested by Hideaki Imamura and Keisuke Umezawa in this discussion.
https://github.com/optuna/optuna/pull/3204#pullrequestreview-866850847

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -10,6 +10,7 @@ from typing import Tuple
 from typing import Union
 
 import optuna
+from optuna import distributions
 from optuna.distributions import BaseDistribution
 from optuna.study._study_direction import StudyDirection
 from optuna.study._study_summary import StudySummary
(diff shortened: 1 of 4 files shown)


toshihikoyanase:
> This is requested by Hideaki Imamura and Keisuke Umezawa in this discussion.
#3204 (review)
@keisuke-umezawa @HideakiImamura Could you review this PR, please?

I'm not confident with the definition of `check_and_set_param_distribution` with `@abc.abstractmethod` since it breaks backward compatibility of user-defined storage classes.

masap:
> I'm not confident with the definition of check_and_set_param_distribution with
> @abc.abstractmethod since it breaks backward compatibility of user-defined storage classes.

@toshihikoyanase
Is this a comment to me ? If so, I would like to try to create the `user-defined storage class`.
Where is the documentation of it ? I could not find out it in the optuna source tree...

toshihikoyanase:
> @toshihikoyanase Is this a comment to me ? If so, I would like to try to create the `user-defined storage class`. 

No, it isn't. I just wanted to know the opinions of the reviewers. Sorry for bothering you.

> Where is the documentation of it ? I could not find out it in the optuna source tree...

I mean if some users has their own storage class in their code (not optuna code base), this change might break their storage classes because they do not have the `check_and_set_param_distribution` method.

masap:
> I mean if some users has their own storage class in their code (not optuna code base),
> this change might break their storage classes because they do not have the `check_and_set_param_distribution method`.

I fixed to leave the method name to `_check_and_set_param_distribution` for backward compatibility.

I'll wait for other reviewers to decide whether to add `abc.abstractmethod` or not.

HideakiImamura:
@masap Sorry for the late reply. We discussed whether we should make the `_CachedStorage` public and concluded that, in Optuna V3, we don't. Therefore, changes such as those proposed in this PR are no longer necessary at this stage. Sorry for the confusion.

Please see https://github.com/optuna/optuna/issues/2938#issuecomment-1035986042 for more details. Thank you for your continuing contribution to Optuna. We are still in the process of discussing the policy, but if you could help us with the implementation of 2943, we would be very happy.

masap:
OK, dropped.

Issue #3256, opened 2022-01-27
More control over how GridSampler prioritizes configurations after search space is exhausted


cschell:
### Motivation

I run jobs on a cluster performing trials using GridSampler. These jobs can be evicted at any time, which causes these trials to get the state "Fail" in the database. The jobs will restart eventually to pick up a new trial (I'm not attempting to restart previous runs/trails). I need to make sure, that configurations of "failed" trials are not "forgotten".

### Description

The current behaviour of `GridSampler` suggesting a configuration for a new trial is as follows:

1. [pick a configuration that has not been visited yet](https://github.com/optuna/optuna/blob/master/optuna/samplers/_grid.py#L137)
    1. get a list of all configurations that are "[finished](https://github.com/optuna/optuna/blob/854c44ff4ba7238059d69641e32b44f3dbd587e8/optuna/trial/_state.py#L31)" (i.e., "not running" and "not waiting") or "running"
    2. pick a configuration that is not in that list
2. if there are no such unvisited configurations left, ignore the state of each configuration and just pick one from [_all_ configurations](https://github.com/optuna/optuna/blob/master/optuna/samplers/_grid.py#L151); also, issue a warning that "_`GridSampler` is re-evaluating a configuration because the grid has been exhausted._"

It would be great for my purposes to tell `GridSampler` for the second step what configuration-states are allowed to be re-run. I propose a parameter `allow_rerun: List[TrialState]` that allows to specify that:

a.  `GridSampler(search_space={...}, allow_rerun=None)` => could be the default, which just does not change the current behaviour: if search space is exhausted, consider all configurations equally
b.  `GridSampler(search_space={...}, allow_rerun=[TrialState.FAIL])` => if search space is exhausted, only rerun configurations that are Failed and exit if none are left (so the search doesn't run endlessly)
c.  `GridSampler(search_space={...}, allow_rerun=[TrialState.FAIL, TrialState.PRUNED])` => if search space is exhausted, only rerun configurations that are either Failed or Pruned
d.  `GridSampler(search_space={...}, allow_rerun=[])` => if search space is exhausted, just exit

Altogether, this allows me to make sure that all configurations will be evaluated.

I can work on a PR for that, but first I would like to make sure this is something you approve of.

himkt:
Hello @cschell, thank you for raising the issue.

> These jobs can be evicted at any time, which causes these trials to get the state "Fail" in the database

[...]

PR #3273, opened 2022-01-31
Warn if heartbeat is used with ask-and-tell.

toshihikoyanase:
## Motivation

Related to #3256. Currently, the heartbeat feature of `RDBStorage` works with `Study.optimize`. In other words, the heartbeat signals are never sent if users use the ask-and-tell interface.

@himkt suggested the change to notify users that the heartbeat was disabled when they used ask-and-tell, and I created a prototypical implementation.

## Description of the changes

- Check the `Study._optimize_lock` to check if `Study.ask` is called in `Study.optimize`
- If not, check the storage is enabled the heartbeat feature
- If so, raise warnings 

Example code

```python
import time

import optuna


def objective(trial):
    time.sleep(3)
    return 1


if __name__ == "__main__":
        "sqlite:///:memory:",
        heartbeat_interval=1,
        grace_period=2,
    )
    for i in range(2):
        trial = study.ask()
        value = objective(trial)
        # study.tell(trial=trial, values=value)
    print(study.trials_dataframe())
```

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -463,6 +463,17 @@ class Study:
             A :class:`~optuna.trial.Trial`.
         """
 
+        if self._optimize_lock.acquire(False):
+            _msg = "Heartbeat of RDBStorage is supposed to be used with Study.optimize."
+            if isinstance(self._storage, storages.RDBStorage):
+                if self._storage.heartbeat_interval is not None:
+                    warnings.warn(_msg)
+            elif isinstance(self._storage, storages._CachedStorage) and isinstance(
+                self._storage._backend, storages.RDBStorage
+            ):
+                if self._storage._backend.heartbeat_interval is not None:
+                    warnings.warn(_msg)
+
         fixed_distributions = fixed_distributions or {}
 
         # Sync storage once every trial.
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -464,6 +464,7 @@ class Study:
         """
 
         if self._optimize_lock.acquire(False):
+            self._optimize_lock.release()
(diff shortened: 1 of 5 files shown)


optuna/study/study.py
```diff
@@ -462,6 +462,18 @@ def ask(
             A :class:`~optuna.trial.Trial`.
         """
 
+        if self._optimize_lock.acquire(False):
+            self._optimize_lock.release()
```
himkt, line 466: [Question] What is the difference between this and checking `self._optimize_lock.locked`?
https://docs.python.org/3/library/threading.html#threading.Lock.locked

toshihikoyanase: Thank you for your suggestion. That makes sense. We don't need to acquire lock here because we just wanted to know the status of the lock. I fixed it in https://github.com/optuna/optuna/pull/3273/commits/d70c2bb59d3650a1dc6b33eb325ff6269955c373.

himkt (approved):
LGTM!

nzw0301:
How about mentioning this behaviour in the docs? Precisely, 

- As a note section:
    1. https://optuna.readthedocs.io/en/latest/reference/generated/optuna.storages.RedisStorage.html
    2. https://optuna.readthedocs.io/en/latest/reference/generated/optuna.storages.RDBStorage.html
- As a note or a sentence:
    3. https://optuna.readthedocs.io/en/latest/faq.html#can-i-monitor-trials-and-make-them-failed-automatically-when-they-are-killed-unexpectedly

nzw0301 (approved):
Awesome. Thanks!

PR #3276, opened 2022-02-01
Carry over notes on `step` behavior to new distributions

xadrianzetx:
This PR ensures that documentation on `step` behavior in `DiscreteUniformDistribution` and `IntUniformDistribution` is carried over to `FloatDistribution` and `IntDistribution`.

## Description of the changes

* Fix documentation

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -108,6 +108,12 @@ class FloatDistribution(BaseDistribution):
     This object is instantiated by :func:`~optuna.trial.Trial.suggest_float`, and passed to
     :mod:`~optuna.samplers` in general.
 
+    .. note::
+        When ``step`` is not :obj:`None`, if the range :math:`[\\mathsf{low}, \\mathsf{high}]`
+        is not divisible by :math:`step`, :math:`\\mathsf{high}` will be replaced with
+        the maximum of :math:`k step + \\mathsf{low} < \\mathsf{high}`, where :math:`k`
+        is an integer.
+
     Attributes:
         low:
             Lower endpoint of the range of the distribution. ``low`` is included in the range.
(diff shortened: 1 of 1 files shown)


optuna/distributions.py
```diff
@@ -108,6 +108,12 @@ class FloatDistribution(BaseDistribution):
     This object is instantiated by :func:`~optuna.trial.Trial.suggest_float`, and passed to
     :mod:`~optuna.samplers` in general.
 
+    .. note::
+        When ``step`` is not :obj:`None`, if the range :math:`[\\mathsf{low}, \\mathsf{high}]`
+        is not divisible by :math:`step`, :math:`\\mathsf{high}` will be replaced with
+        the maximum of :math:`k step + \\mathsf{low} < \\mathsf{high}`, where :math:`k`
```
himkt, line 114:
```suggestion
        the maximum of :math:`k \\times step + \\mathsf{low} < \\mathsf{high}`, where :math:`k`
```

himkt, line 114: I didn't notice the formats for maximum are different between `IntDistribution` (`k \\times \\mathsf{step}`) and `DiscreteUniformDistribution` (`kd`). 📝  I think it would also be preferable to use `k \\times \\mathsf{step}` because `d` is replace with `step`, which is not one character variable.

himkt (commented):
Thank you for adding notes, looks basically solid. :+1:

toshihikoyanase:
@himkt Thank you for your review. I agree with your suggestion. Otherwise, LGTM.

optuna/distributions.py
xadrianzetx: Thanks for the suggestion! Applied in ee2f4cd1c3c1ad44f69987bbf2f9a3e1379612e1.

Screenshot from 2022-02-02 13-23-50

optuna/distributions.py
```diff
@@ -108,6 +108,12 @@ class FloatDistribution(BaseDistribution):
     This object is instantiated by :func:`~optuna.trial.Trial.suggest_float`, and passed to
     :mod:`~optuna.samplers` in general.
 
+    .. note::
+        When ``step`` is not :obj:`None`, if the range :math:`[\\mathsf{low}, \\mathsf{high}]`
+        is not divisible by :math:`step`, :math:`\\mathsf{high}` will be replaced with
+        the maximum of :math:`k \\times \\mathsf{step} + \\mathsf{low} < \\mathsf{high}`,
```
himkt, line 114:
```suggestion
        is not divisible by :math:`\\mathsf{step}`, :math:`\\mathsf{high}` will be replaced
        with the maximum of :math:`k \\times \\mathsf{step} + \\mathsf{low} < \\mathsf{high}`,
```

nitpick

xadrianzetx: Good catch, thank you! cb47e922b24e9217988ef7b84914d665b508a5ae.

himkt (approved):
Thank you for the quick revise, LGTM!

himkt:
@toshihikoyanase Please take a look again. 🙏

HideakiImamura (approved):
Thanks for the PR! LGTM!

PR #3277, opened 2022-02-01
Switched order of function definitions

BasLaa:
## Motivation
Test functions for progress bar in optimize were thought to be able to be simplified.
This would complicate the logic so we decided to leave it.
However, a change of order of functions puts similar tests next to each other
## Description of the changes
Change order of test functions

PR #3280, opened 2022-02-02
switch function order progbar tests

BasLaa:
## Motivation

Test functions for progress bar in optimize were thought to be able to be simplified.
This would complicate the logic so we decided to leave it.
However, a change of order of functions puts similar tests next to each other, making it more clear.

## Description of the changes

Change order of test functions


BasLaa:
This is the non-messy version of #3277

nzw0301 (approved):
LGTM. Thanks!

PR #3283, opened 2022-02-02
Add setter method for DiscreteUniformDistribution.q

himkt:
Follow up of #3244 

## Motivation
After #3244, `DiscreteUniformDistribution` is the subclass of `FloatDistribution`.
`DiscreteUniformDistribution` has `q` attribute as a discretization step on the other hand `FloatDistribution` has `step` attribute for the purpose. So I added the proxy method `q` to get `step` attribute but forgot to add a setter, which introduces a backward compatibility.

In this PR, I introduce the setter for `DiscreteUniformDistribution.q`.

## Description of the changes
- c76fd21f adds a setter

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -280,6 +280,10 @@ class DiscreteUniformDistribution(FloatDistribution):
     def q(self) -> float:
         return cast(float, self.step)
 
+    @q.setter
+    def q(self, v: float) -> float:
+        self.step = v
+
 
 class IntDistribution(BaseDistribution):
     """A distribution on integers.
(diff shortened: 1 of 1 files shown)


xadrianzetx (approved):
Great! Aliasing `step` as `q` ensures that situation where `q != step` is not possible. LGTM!

toshihikoyanase:
@xadrianzetx Thank you for your review!

@hvy Could you review this PR since it is a follow up for #3244?

hvy (approved):
LGTM!

Just for the record, `q` was not mutable prior to this change. We explicitly defined an accessor (`@property`) which prohibited writes. I guess unit tests could be worked on to cover this, if it's intended usage.

Issue #2955, opened 2021-09-25
Stabilize `study.py`


himkt:
## Motivation

In this issue, we consider to release a stable version of methods in `study.py`: `enqueue_trial`, `add_trial`, `add_trials`, and `copy_study`. For `enqueue_trial` and `copy_study`, we know we need to discuss/fix:

### Needs discussion

- [x] For `enqueue_trial`, we have to decide whether it allows to specify attributes other than `params` (e.g. `user_attrs`).
  - [updated:2021/11/28] we decide to add `user_attrs` parameter to `enqueue_trial` https://github.com/optuna/optuna/pull/3185
- [x] For `copy_study`, we need to clarify the specification when `copy_study` is invoked for a running study.
  - [updated:2021/11/28] we decide to document that `copy_study` copies a study regardless of its state (if study is in the optimization, it at the time of the call will be copied as is) #3295

### Needs fixing

- [x] An enqueued param is ignored when it is out of range of a parameter distribution #3298
  - We plan to change the behavior so that it prioritizes a enqueued param
    - current: show warning message and ignore enqueued parameter
    - desired: show warning message and use enqueued parameter
 - [x] When `enqueue_trial` and `GridSampler` are used simultaneously, parameters on a grid are overwritten by `enqueue_trial` (https://github.com/optuna/optuna/pull/2946)
   - We will change the behavior so that all parameters on a grid are used

## Description

- [x] Reach agreements about matters in `needs discussion`
- [x] Determine approaches how to fix problems in `needs fixing`
- [x] Merge PRs to reflect modifications for them
- [x] Decide if we release a stable version of each method
- [ ] Stabilize methods we decide to release

PR #3295, opened 2022-02-07
Add a note for copy_study: it craetes a copy regardless of its state

himkt:
🔗 https://github.com/optuna/optuna/issues/2955

This PR adds a note for `optuna.copy_study`.
It explicitly mentions that `copy_study` will create a copy regardless of its status.

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1323,6 +1323,10 @@ def copy_study(
     The direction(s) of the objective(s) in the study, trials, user attributes and system
     attributes are copied.
 
+    .. note::
+        :func:`optuna.copy_study` copies a study even if the optimization is working on.
+        It means users will get a copied study that contains a trial that is not finished.
+
     Example:
 
         .. testsetup::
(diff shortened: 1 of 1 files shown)


optuna/study/study.py
```diff
@@ -1323,6 +1323,10 @@ def copy_study(
     The direction(s) of the objective(s) in the study, trials, user attributes and system
     attributes are copied.
 
+    .. note::
+        :func:`optuna.copy_study` copies a study even if the optimization is working on.
```
HideakiImamura, line 1327: I'm not sure but I think the notation that omits the path is more commonly used in Optuna.
```suggestion
        :func:`~optuna.copy_study` copies a study even if the optimization is working on.
```

HideakiImamura (commented):
Thanks for the PR. I have a minor comment.

optuna/study/study.py
himkt: Thank you, I applied. And I found there are both styles in `study.py`. So, I'll open another PR to fix other place e.g. https://github.com/optuna/optuna/blob/15e9f38e4a448f2c1775e46e396ba7fc190df537/optuna/study/study.py#L1305.

HideakiImamura (approved):
LGTM.

PR #3305, opened 2022-02-09
Add description field for question-and-help-support

himkt:
I propose to add a field for describing the question to question-and-help-support.

Preview:

Screen Shot 2022-02-09 at 21 49 34


nzw0301 (approved):

toshihikoyanase (approved):
Good catch! Thank you.

[Note] Currently, the question form has the following three fields:
- Environment,
- Error messages, and
- Steps to reproduce.

It does not have fields to describe questions in natural languages, so the change looks reasonable.

PR #3314, opened 2022-02-15
Update README to inform v3.0.0a2

himkt:
Update README to inform users about v3.0.0a2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Optuna can dynamically construct the search spaces for the hyperparameters.
 
 ## News
 
-- **2022-02-14** Pre-releases of Optuna 3.0 are available! Early adopters may want to upgrade and provide feedback for a smoother transition to the coming full release. You can install a pre-release version by `pip install --pre optuna`. Find the latest one [here](https://github.com/optuna/optuna/releases)
+- **2022-02-14** Pre-releases of Optuna 3.0 are available! Early adopters may want to upgrade and provide feedback for a smoother transition to the coming full release. You can install a pre-release version by `pip install -U --pre optuna`. Find the latest one [here](https://github.com/optuna/optuna/releases)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ Optuna can dynamically construct the search spaces for the hyperparameters.
 
 ## News
 
-- **2021-12-06** First alpha version of Optuna 3.0 is released! Early adopters may want to upgrade and provide feedback for a smoother transition to the coming major release. Try `pip install optuna==3.0.0a0`.
+- **2022-02-14** Pre-releases of Optuna 3.0 are available! Early adopters may want to upgrade and provide feedback for a smoother transition to the coming full release. Find the latest one [here](https://github.com/optuna/optuna/releases). Try `pip install optuna==3.0.0a2`
```
himkt, line 26: > Try `pip install optuna==3.0.0a2`

Isn't it needed anymore? Our release note currently doesn't show the install command so I'm wondering if we can remove it here.

hvy: Definitely not a strong opinion, but I think we can remove it from here, if we're forwarding to the list of releases. Alternatively, we could show the command with the `--pre` option instead of specifying a concrete value. These ways, we'd not have to update the news after each pre-release (i.e. b0, rc, ..)

hvy (approved):
LGTM!

README.md
himkt: Thank you for the comments! Both (removing and showing the command with `pre` option) sound reasonable. I personally prefer the latter choice and revise the PR to show the command in 16a788c!

hvy (approved):
LGTM!

(Again, not a strong opinion but it might be friendly to include a `-U` (or `--upgrade`) in the example command, if we're targeting existing users)

himkt:
It makes sense! [ab99f8b](https://github.com/optuna/optuna/pull/3314/commits/ab99f8b879444e77bd913278f0e9d2c1cfd42e3c) add `-U`.

HideakiImamura (approved):
LGTM.

PR #3322, opened 2022-02-18
Fix layout in tutorial (with workaround)

himkt:
This PR patches the style of tutorial.

I'm not sure what is the root cause but DOMs are different between sphinx 3.x (current stable) and 4.x (current latest).
For example, each content in the tutorial is made with div on v2.10.0 but with section tag in v3.0.0-a2.

current stable v2.10.0 (sphinx 3.5.4) | current latest v3.0.0-a2 (sphinx 4.4.0)
-- | --
Screen Shot 2022-02-18 at 13 54 18 | Screen Shot 2022-02-18 at 13 53 26

I'm not sure if this difference introduces the layout difference but the size of component differs in the latest documentation. Note that the width is not explicitly set in both builds.

current stable | current latest
-- | --
Screen Shot 2022-02-18 at 14 11 38 | Screen Shot 2022-02-18 at 14 11 23

In this PR, I add the style to set `min-width` for `sphx-glr-thumbcontainer` to guarantee a component has the same width at least as one in stable build.

current stable | current latest | this PR
-- | -- | --
Screen Shot 2022-02-18 at 13 45 28 | Screen Shot 2022-02-18 at 13 45 50 | Screen Shot 2022-02-18 at 13 46 28

Please let me know if someone have a proper way to fix this problem.

### DIFF
--- a/docs/source/_static/css/custom.css
+++ b/docs/source/_static/css/custom.css
@@ -181,6 +181,7 @@ a:hover {
 }
 
 /* This is the patch fix for tutorial page */
+/* TODO (himkt): Remove here after resolving https://github.com/optuna/optuna/pull/3322 */
 .sphx-glr-thumbcontainer {
   min-width: 182px;
 }
--- a/docs/source/_static/css/custom.css
+++ b/docs/source/_static/css/custom.css
@@ -179,3 +179,8 @@ a:hover {
 .rst-content .admonition {
   background: #f1f3f4;
 }
+
+/* This is the patch fix for tutorial page */
+.sphx-glr-thumbcontainer {
+  min-width: 182px;
+}


docs/source/_static/css/custom.css
```diff
@@ -179,3 +179,9 @@ a:hover {
 .rst-content .admonition {
   background: #f1f3f4;
 }
+
+/* This is the patch fix for tutorial page */
+/* TODO (himkt): Remove here after resolving https://github.com/optuna/optuna/pull/3322 */
+.sphx-glr-thumbcontainer {
+  min-width: 182px;
```
nzw0301, line 186: Can I ask how to decide this value?

nzw0301 (approved):
Awesome.

toshihikoyanase (approved):
Good catch! Thank you!

PR #3258, opened 2022-01-29
Add integration test for `RedisStorage`

masap:
## Motivation
Add integration test for `RedisStorage`

PR #3323, opened 2022-02-19
Merge `set_trial_state()` and `set_trial_values()` into one  function

masap:
## Motivation
In `tell()` of `Study` class, there is trailing code.

```
    if values is not None:
        self._storage.set_trial_values(trial_id, values)

    self._storage.set_trial_state(trial_id, state)
```

Consider process A is running ask-and-tell, process B is getting trial information by `study.get_trials()`. There are 3 steps at this code.

At step 1 of process A, process B get state=`TrialState.RUNNING`, value=`None`.
At step 2 of process A, process B get state=`TrialState.RUNNING`, value!=`None`.
At step 3 of process A, process B get state=`TrialState.COMPLETE`, value!=`None`.

```
    #step 1

    if values is not None:
        self._storage.set_trial_values(trial_id, values)

    #step 2

    self._storage.set_trial_state(trial_id, state)

    #step 3
```

We can get state=`TrialState.RUNNING`, value!=`None` at step 2 unexpectedly. It breaks consistency of trials because we expects only these combinations.
   - state=`TrialState.RUNNING`, value=`None`
   - state=`TrialState.COMPLETE`, value!=`None`

This issue is reported by @not522 in https://github.com/optuna/optuna/issues/2943#issuecomment-997744058.

## Description of the changes
To resolve the issue, this patch merges `set_trial_state()` and `set_trial_values()` into one function. To keep backward compatibility, we leaved `set_trial_state()` and `set_trial_values()` as deprecated.

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -334,31 +334,6 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         raise NotImplementedError
 
     @abc.abstractmethod
-    def set_trial_state(self, trial_id: int, state: TrialState) -> bool:
-        """Update the state of a trial.
-
-        Args:
-            trial_id:
-                ID of the trial.
-            state:
-                New state of the trial.
-
-        Returns:
-            :obj:`True` if the state is successfully updated.
-            :obj:`False` if the state is kept the same.
-            The latter happens when this method tries to update the state of
-            :obj:`~optuna.trial.TrialState.RUNNING` trial to
-            :obj:`~optuna.trial.TrialState.RUNNING`.
-
-        Raises:
-            :exc:`KeyError`:
-                If no trial with the matching ``trial_id`` exists.
-            :exc:`RuntimeError`:
-                If the trial is already finished.
-        """
-        raise NotImplementedError
-
-    @abc.abstractmethod
     def set_trial_param(
         self,
         trial_id: int,
@@ -445,16 +420,28 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         return trial.distributions[param_name].to_internal_repr(trial.params[param_name])
 
     @abc.abstractmethod
-    def set_trial_values(self, trial_id: int, values: Sequence[float]) -> None:
-        """Set return values of an objective function.
+    def set_trial_state_values(
+        self, trial_id: int, state: TrialState, values: Optional[Sequence[float]] = None
+    ) -> bool:
+        """Update the state and values of a trial.
 
-        This method overwrites any existing trial values.
+        Set return values of an objective function to values argument.
+        If values argument is not :obj:`None`, this method overwrites any existing trial values.
 
         Args:
             trial_id:
                 ID of the trial.
+            state:
+                New state of the trial.
             values:
                 Values of the objective function.
+
+        Returns:
+            :obj:`True` if the state is successfully updated.
+            :obj:`False` if the state is kept the same.
+            The latter happens when this method tries to update the state of
+            :obj:`~optuna.trial.TrialState.RUNNING` trial to
+            :obj:`~optuna.trial.TrialState.RUNNING`.
 
         Raises:
             :exc:`KeyError`:
--- a/optuna/storages/_cached_storage.py
+++ b/optuna/storages/_cached_storage.py
@@ -181,38 +181,6 @@ class _CachedStorage(BaseStorage):
                 study.owned_or_finished_trial_ids.add(frozen_trial._trial_id)
         return trial_id
 
-    def set_trial_state(self, trial_id: int, state: TrialState) -> bool:
-
-        with self._lock:
-            cached_trial = self._get_cached_trial(trial_id)
-            if cached_trial is not None:
-                # When a waiting trial is updated to running, its `datetime_start` must be
-                # updated. However, a waiting trials is never cached so we do not have to account
-                # for this case.
-                assert cached_trial.state != TrialState.WAITING
-
-                self._check_trial_is_updatable(cached_trial)
-                ret = self._backend.set_trial_state(trial_id, state)
-
-                cached_trial.state = state
-                if cached_trial.state.is_finished():
-                    backend_trial = self._backend.get_trial(trial_id)
-                    cached_trial.datetime_complete = backend_trial.datetime_complete
-                return ret
-
-        ret = self._backend.set_trial_state(trial_id, state)
-        if (
-            ret
-            and state == TrialState.RUNNING
-            and trial_id in self._trial_id_to_study_id_and_number
-        ):
-            # Cache when the local thread pop WAITING trial and start evaluation.
-            with self._lock:
-                study_id, _ = self._trial_id_to_study_id_and_number[trial_id]
-                self._add_trials_to_cache(study_id, [self._backend.get_trial(trial_id)])
-                self._studies[study_id].owned_or_finished_trial_ids.add(trial_id)
-        return ret
-
     def set_trial_param(
         self,
         trial_id: int,
@@ -265,14 +233,41 @@ class _CachedStorage(BaseStorage):
 
         return self._backend.get_best_trial(study_id)
 
-    def set_trial_values(self, trial_id: int, values: Sequence[float]) -> None:
+    def set_trial_state_values(
+        self, trial_id: int, state: TrialState, values: Optional[Sequence[float]] = None
+    ) -> bool:
 
         with self._lock:
             cached_trial = self._get_cached_trial(trial_id)
             if cached_trial is not None:
+                # When a waiting trial is updated to running, its `datetime_start` must be
+                # updated. However, a waiting trials is never cached so we do not have to account
+                # for this case.
+                assert cached_trial.state != TrialState.WAITING
+
                 self._check_trial_is_updatable(cached_trial)
-                cached_trial.values = values
-        self._backend.set_trial_values(trial_id, values=values)
+                ret = self._backend.set_trial_state_values(trial_id, state=state, values=values)
+
+                if values is not None:
+                    cached_trial.values = values
+                cached_trial.state = state
+                if cached_trial.state.is_finished():
+                    backend_trial = self._backend.get_trial(trial_id)
+                    cached_trial.datetime_complete = backend_trial.datetime_complete
+                return ret
+
+        ret = self._backend.set_trial_state_values(trial_id, state=state, values=values)
+        if (
+            ret
+            and state == TrialState.RUNNING
+            and trial_id in self._trial_id_to_study_id_and_number
+        ):
+            # Cache when the local thread pop WAITING trial and start evaluation.
+            with self._lock:
+                study_id, _ = self._trial_id_to_study_id_and_number[trial_id]
+                self._add_trials_to_cache(study_id, [self._backend.get_trial(trial_id)])
+                self._studies[study_id].owned_or_finished_trial_ids.add(trial_id)
+        return ret
 
     def set_trial_intermediate_value(
         self, trial_id: int, step: int, intermediate_value: float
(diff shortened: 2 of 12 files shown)


masap:
Some tests fail because http://azure.archive.ubuntu.com/ubuntu does not response. I will re-run later...

masap:
Merge #3258 first, because of conflict. After #3258 is merged, I will fix the conflict.

toshihikoyanase:
@keisuke-umezawa Could you examine this change is consistent with #2943 before diving into the details?

@not522 Thank you for your interest in this PR!

masap:
Rebased.

masap:
Rebased to fix conflict caused by #3258.

keisuke-umezawa:
Before review, I would like to summarize the current spec of `set_trial_state_values`, 

1. `values` are set only when they are not `None` and trial is updatable.
2. `state` is set only when trial is updatable and state change is in the following pattern:
    - WAITING -> ANY
    - Others -> (COMPLETE, PRUNED, FAIL, WAITING)
3. The 1 and 2 are set in a atomic manner.

Also, the following things can be additional expected spec of `set_trial_state_values`, although we need more dicsussion.
1. `values` are set only when target state is `COMPLETE`.
2. etc.

masap:
Fixed conflict.

masap:
Rebased to retest.

optuna/storages/_cached_storage.py
```diff
@@ -180,6 +181,11 @@ def create_new_trial(self, study_id: int, template_trial: Optional[FrozenTrial]
                 study.owned_or_finished_trial_ids.add(frozen_trial._trial_id)
         return trial_id
 
+    @deprecated(
+        "3.0.0",
+        "4.0.0",
+        text="Use :func:`~optuna.storages._CachedStorage.set_trial_state_values` instead.",
+    )
```
keisuke-umezawa, line 188: Since we have not make `CachedStorage` and `InMemoryStorage` public, we can remove unnecessary methods. So, could you remove `set_trial_state()` and `set_trial_values()` from `CachedStorage` and `InMemoryStorage`?

optuna/storages/_in_memory.py
```diff
@@ -335,6 +346,40 @@ def set_trial_values(self, trial_id: int, values: Sequence[float]) -> None:
             trial.values = values
             self._set_trial(trial_id, trial)
 
+    def set_trial_state_values(
+        self, trial_id: int, state: TrialState, values: Optional[Sequence[float]] = None
+    ) -> bool:
+
+        with self._lock:
+            trial = self._get_trial(trial_id)
+            self.check_trial_is_updatable(trial_id, trial.state)
```
keisuke-umezawa, line 355: Maybe, we can remove it.

optuna/storages/_in_memory.py
```diff
@@ -335,6 +346,40 @@ def set_trial_values(self, trial_id: int, values: Sequence[float]) -> None:
...
+
+            trial = copy.copy(trial)
+            self.check_trial_is_updatable(trial_id, trial.state)
+
+            if values is not None:
+                trial.values = values
```
keisuke-umezawa, line 361: If values are not None, we always set values.
After this line, we can write `self._set_trial(trial_id, trial)`?

keisuke-umezawa (changes requested):
Add some comments. Could you check them?

optuna/storages/_base.py
```diff
@@ -458,6 +458,43 @@ def set_trial_values(self, trial_id: int, values: Sequence[float]) -> None:
         """
         raise NotImplementedError
 
+    # TODO(masap): When set_trial_state() and set_trial_values() are removed, make this function
+    # abstract and raise NotImplementedError.
+    def set_trial_state_values(
```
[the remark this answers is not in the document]
masap: Thanks for your review!

> So, could you change it as abstractmethod and remove set_trial_state() and set_trial_values() from BaseStorage?
> Also, in this case, we do not need to provide default implementation for this method.

According to comment from @toshihikoyanase, there are users who has user-defined storage classes.

 - https://github.com/optuna/optuna/pull/3272#issuecomment-1025398438
 - https://github.com/optuna/optuna/pull/3272#issuecomment-1025472081

Visible storage classes for users are `RDBStorage` and `RedisStorage`. So I guess users inherit them to create user-defined storage classes.

If we remove `set_trial_state()` and `set_trial_values()` and make `set_trial_state_values()` returns `NotImplementedError`, it breaks backward compatibility.

How do you think about this ?

optuna/storages/_cached_storage.py
masap: Ok, fixed. I also remove remove `abstract` from `set_trial_state()` and `set_trial_values()` of `BaseStorage` class.

optuna/storages/_in_memory.py
masap: Yes, it is duplicated. Removed.

optuna/storages/_in_memory.py
masap: It makes code more simple, but when updating state we call `self._set_trial(trial_id, trial)` twice. So I selected more effective way. Which do you like ?

masap:
Rebased for retest.

optuna/storages/_base.py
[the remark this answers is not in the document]
keisuke-umezawa: Thank you for the comment. We will discuss it and make decision for it.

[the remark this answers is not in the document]
keisuke-umezawa: cc: @toshihikoyanase 
After the discussion, we reached the following consensus.

Currently, as we have not make BaseStorage public yet, in our policy it should be allowed to add breaking changes into BaseStorage.

Could you follow the first comment by me?

> Since we have not make BaseStorage public, we can change which methods are abstract and remove unnecessary methods. So, could you change it as abstractmethod and remove set_trial_state() and set_trial_values() from BaseStorage?

optuna/storages/_in_memory.py
keisuke-umezawa: I understand. Leave this line as it is.

optuna/storages/_base.py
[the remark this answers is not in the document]
masap: > Currently, as we have not make BaseStorage public yet, in our policy it should be allowed to add breaking changes into BaseStorage.

Ok, It is reasonable. I removed the methods and test for user defined storage.

keisuke-umezawa (approved):
LGTM! Thank you for working on this complicated issue.

@not522 Could you check this PR?

optuna/storages/_in_memory.py
```diff
@@ -323,17 +296,38 @@ def get_trial_param(self, trial_id: int, param_name: str) -> float:
...
             self.check_trial_is_updatable(trial_id, trial.state)
 
-            trial.values = values
-            self._set_trial(trial_id, trial)
+            if values is not None:
+                trial.values = values
+
+            if state == TrialState.RUNNING and trial.state != TrialState.WAITING:
+                if values is not None:
+                    self._set_trial(trial_id, trial)
```
not522, line 314: When `state` is `RUNNING`, `values` should be `None`. These lines can be omitted.

optuna/storages/_rdb/storage.py
```diff
@@ -644,6 +645,11 @@ def _get_prepared_new_trial(
 
         return trial
 
+    @deprecated(
+        "3.0.0",
+        "4.0.0",
```
not522, line 650: As default, we set `removed_version` as the two later major version of `deprecated_version`.
```suggestion
        "5.0.0",
```

optuna/storages/_rdb/storage.py
```diff
@@ -764,6 +770,11 @@ def _lift_numerical_limit(value: float) -> float:
             return float(np.sign(value) * float("inf"))
         return value
 
+    @deprecated(
+        "3.0.0",
+        "4.0.0",
```
not522, line 775:
```suggestion
        "5.0.0",
```

optuna/storages/_redis.py
```diff
@@ -410,6 +411,11 @@ def _create_running_trial() -> FrozenTrial:
             datetime_complete=None,
         )
 
+    @deprecated(
+        "3.0.0",
+        "4.0.0",
```
not522, line 416:
```suggestion
        "5.0.0",
```

optuna/storages/_redis.py
```diff
@@ -541,6 +547,11 @@ def _check_and_set_param_distribution(
         self._check_study_id(study_id)
         self.set_trial_param(trial_id, param_name, param_value_internal, distribution)
 
+    @deprecated(
+        "3.0.0",
+        "4.0.0",
```
not522, line 552:
```suggestion
        "5.0.0",
```

optuna/storages/_redis.py
```diff
@@ -549,6 +560,40 @@ def set_trial_values(self, trial_id: int, values: Sequence[float]) -> None:
...
+
+        trial = self.get_trial(trial_id)
+        self.check_trial_is_updatable(trial_id, trial.state)
+
+        if values is not None:
+            trial.values = values
+
+        if state == TrialState.RUNNING and trial.state != TrialState.WAITING:
+            if values is not None:
+                self._redis.set(self._key_trial(trial_id), pickle.dumps(trial))
```
not522, line 575:
```suggestion
```

not522 (commented):
Thank you for your great PR! I added minor comments. Please check them.
I think that we can remove the tests for `set_trial_state` and `set_trial_values`.

optuna/storages/_base.py
```diff
@@ -439,17 +414,29 @@ def get_trial_param(self, trial_id: int, param_name: str) -> float:
...
-    def set_trial_values(self, trial_id: int, values: Sequence[float]) -> None:
-        """Set return values of an objective function.
+    def set_trial_state_values(
+        self, trial_id: int, state: TrialState, values: Optional[Sequence[float]] = None
+    ) -> bool:
+        """Update the state and values of a trial.
 
-        This method overwrites any existing trial values.
+        Set return values of an objective function to values argument.
+        If values argument is not None, this method overwrites any existing trial values.
```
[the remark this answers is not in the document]
masap: Thank you for your review ! Fixed.

optuna/storages/_in_memory.py
masap: Right ! Fixed. I also fixed related tests.

And I removed overwrite test from `test_set_trial_state_values_for_values` because I could not find out the way to update existing trial value. If it is possible please tell me.

```
        # Values can be overwritten.
        storage.set_trial_state_values(trial_id_1, state=trial_1.state, values=(0.2,))
        assert storage.get_trial(trial_id_1).value == 0.2
```

optuna/storages/_rdb/storage.py
masap: Ok, Fixed.

optuna/storages/_rdb/storage.py
masap: Fixed.

optuna/storages/_redis.py
masap: Fixed.

optuna/storages/_redis.py
masap: Fixed.

optuna/storages/_redis.py
masap: Fixed.

masap:
> I think that we can remove the tests for `set_trial_state` and `set_trial_values`.

How about remove the tests when the methods are removed ?

not522:
> How about remove the tests when the methods are removed ?

OK, let's keep them as it is.

not522:
Your PR looks good to me. But I realize my original idea is not smart -- `set_trial_state_values` requires complex if-statements. I will discuss the storages API design with other core developers this week. Please wait for other members' suggestions.

masap:
> I will discuss the storages API design with other core developers this week. Please wait for other members' suggestions.

Ok.

not522 (approved):
Thank you for waiting. Let me merge this PR.
The storage API design is still under discussion. We may change it in follow-up PRs, but core developers agreed that it is not a blocker to merge this PR.

masap:
Thanks!

Issue #3293, opened 2022-02-07
Supporting heartbeat in Ask-and-tell


himkt:
### Motivation

[Heartbeat](https://optuna.readthedocs.io/en/latest/faq.html#can-i-monitor-trials-and-make-them-failed-automatically-when-they-are-killed-unexpectedly) is currently available only on `Study.optimize`. In other words, we cannot utilize heartbeat functionalities even if the storage is created with `heartbeat_interval`. This behavior is [confusing](https://github.com/optuna/optuna/issues/3256) and we will show a warning message for such case.

Supporting heartbeat in ask-and-tell use-case would make distributed optimization easier so it would be beneficial to discuss about it.

### Description

Discussion should covers at least: (a) whether we do or not (and it is possible or not) (b) API design.

PR #3329, opened 2022-02-24
[PoC] Support heartbeat in ask-and-tell

himkt:
🔗 https://github.com/optuna/optuna/issues/3293

## Description of the changes

```python
import optuna


    return trial.suggest_float("x", 0, 10)


storage = optuna.storages.RDBStorage(url="sqlite:///sample.db", heartbeat_interval=60)


for _ in range(10):
    with study.ask_with_heartbeat() as trial:
        v = objective(trial)
        study.tell(trial, v)
```


himkt:
Let me close this PoC PR. We will revisit if it needs.

PR #3166, opened 2021-12-14
Switch to `FloatDistribution`

xadrianzetx:
## Motivation

This PR makes the switch from old `UniformDistribution` classes to unified  `FloatDistribution` by instantiating it in `suggest` APIs and few other minor places in codebase. Additionally, all instances of old distributions in test suite are replaced with new one, and parametrization over deprecated classes is removed following https://github.com/optuna/optuna/pull/3111#pullrequestreview-825894668. Part of #2941. This also solves last part of #2939 - parameters within new distribution can be changed freely.

## Description of the changes

* [x] Use `FloatDistribution` in `suggest` APIs and other minor places
* [x] Use instances of `FloatDistribution` in test suite
* [x] Remove *some* test parametrization for old distributions (w/o `pytest.parametrize` cases)
* [x] Fix tutorials

PR #3337, opened 2022-02-27
Follow up for {Float, Int} distributions

nyanhi:
## Motivation
As mentioned in below comment, this is follow up PR relating to #3166.
https://github.com/optuna/optuna/pull/3166#issuecomment-1048607416

## Description of the changes

- [x] In `tests/integration_tests/test_chainermn.py`, renamed the test of `test_suggest_discrete_uniform` to `test_suggest_float_with_step`
- [x] Remove `UniformDistribution` appearance from `test_multi_objective_sample_independent_uniform_distributions` in `tests/samplers_tests/tpe_tests/test_multi_objective.py`. Also, similar tests are parametrized.
- [x] Parametrizing `test_multi_objective_sample_independent_int_distributions` in `tests/samplers_tests/tpe_tests/test_multi_objective.py`. 
- [x] Add tests for `FloatDistribution(log=True)` and `IntDistribution(log=True)` to `test_empty_range_contains` in `tests/test_distributions.py`.
- [x] Given distribution class will handle the error case "log=True & step" (while it was handled in `suggest_{int, float}()` before creating distribution), re-ordered the error handling to be consistent with what used to be.

### DIFF
--- a/tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
+++ b/tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
@@ -359,6 +359,7 @@ def test_multi_objective_sample_independent_categorical_distributions() -> None:
     "log, step",
     [
         (False, 1),
+        (True, 1),
         (False, 2),
     ],
 )
@@ -369,7 +370,7 @@ def test_multi_objective_sample_int_distributions(log: bool, step: int) -> None:
     random.seed(128)
 
     def int_value_fn(idx: int) -> float:
-        return random.randint(0, 100)
+        return random.randint(1, 100)
 
     int_dist = optuna.distributions.IntDistribution(1, 100, log, step)
     past_trials = [
@@ -379,7 +380,7 @@ def test_multi_objective_sample_int_distributions(log: bool, step: int) -> None:
         for i in range(16)
     ]
 
-    trial = frozen_trial_factory(16, [0, 0])
+    trial = frozen_trial_factory(16, [1, 1])
     sampler = TPESampler(seed=0)
     attrs = MockSystemAttr()
     with patch.object(study._storage, "get_all_trials", return_value=past_trials), patch.object(
--- a/tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
+++ b/tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
@@ -378,7 +378,7 @@ def test_multi_objective_sample_int_distributions(log: bool, step: int) -> None:
         for i in range(16)
     ]
 
-    trial = frozen_trial_factory(16, [1, 1])
+    trial = frozen_trial_factory(16, [0, 0])
     sampler = TPESampler(seed=0)
     attrs = MockSystemAttr()
     with patch.object(study._storage, "get_all_trials", return_value=past_trials), patch.object(
--- a/tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
+++ b/tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
@@ -244,13 +244,61 @@ def test_multi_objective_sample_independent_misc_arguments() -> None:
         assert sampler.sample_independent(study, trial, "param-a", dist) != suggestion
 
 
-def test_multi_objective_sample_independent_uniform_distributions() -> None:
-    # Prepare sample from uniform distribution for cheking other distributions.
+@pytest.mark.parametrize(
+    "log, step",
+    [
+        (False, None),
+        (True, None),
+        (False, 0.1),
+    ],
+)
+def test_multi_objective_sample_independent_float_distributions(
+    log: bool, step: Optional[float]
+) -> None:
+    # Prepare sample from float distribution for checking other distributions.
     study = optuna.create_study(directions=["minimize", "maximize"])
     random.seed(128)
+    float_dist = optuna.distributions.FloatDistribution(1.0, 100.0, log=log, step=step)
+
+    if float_dist.step:
+
+        def value_fn(number: int) -> float:
+            return int(random.random() * 1000) * 0.1
+
+        past_trials = [
+            frozen_trial_factory(
+                i, [random.random(), random.random()], dist=float_dist, value_fn=value_fn
+            )
+            for i in range(16)
+        ]
+    else:
+        past_trials = [
+            frozen_trial_factory(i, [random.random(), random.random()], dist=float_dist)
+            for i in range(16)
+        ]
+
+    trial = frozen_trial_factory(16, [0, 0])
+    sampler = TPESampler(seed=0)
+    attrs = MockSystemAttr()
+    with patch.object(study._storage, "get_all_trials", return_value=past_trials), patch.object(
+        study._storage, "set_trial_system_attr", side_effect=attrs.set_trial_system_attr
+    ), patch.object(study._storage, "get_trial", return_value=trial), patch(
+        "optuna.trial.Trial.system_attrs", new_callable=PropertyMock
+    ) as mock1, patch(
+        "optuna.trial.FrozenTrial.system_attrs",
+        new_callable=PropertyMock,
+    ) as mock2:
+        mock1.return_value = attrs.value
+        mock2.return_value = attrs.value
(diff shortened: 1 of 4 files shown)


tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
```diff
@@ -244,58 +244,39 @@ def test_multi_objective_sample_independent_misc_arguments() -> None:
...
-        "optuna.trial.Trial.system_attrs", new_callable=PropertyMock
-    ) as mock1, patch(
-        "optuna.trial.FrozenTrial.system_attrs",
-        new_callable=PropertyMock,
-    ) as mock2:
-        mock1.return_value = attrs.value
-        mock2.return_value = attrs.value
-        uniform_suggestion = sampler.sample_independent(study, trial, "param-a", uni_dist)
+        past_trials = [
+            frozen_trial_factory(
```
nyanhi, line 269: Note - there was discussions (1) if `frozen_trial_factory()` can be simply replaced by `study.optimize()` to get past_trials and (2) `state_fn` argument in `frozen_trial_factory()` seems not used. Those points are to be assessed in a separate issue/PR.

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
```diff
@@ -345,11 +312,11 @@ def value_fn(idx: int) -> float:
     ) as mock2:
         mock1.return_value = attrs.value
         mock2.return_value = attrs.value
-        discrete_uniform_suggestion = sampler.sample_independent(
-            study, trial, "param-a", disc_dist
-        )
-    assert 1.0 <= discrete_uniform_suggestion <= 100.0
-    assert abs(int(discrete_uniform_suggestion * 10) - discrete_uniform_suggestion * 10) < 1e-3
+        suggestion = sampler.sample_independent(study, trial, "param-a", dist)
+    if float_dist.log or float_dist.step != 1.0:
```
himkt, line 316: I think this condition is always `True` given the parametrization.

Furthermore, I'm not sure `float_suggeestion` would be the same as `suggestion` even if the parameterization is expected. (the current code only checks if they are not identical (https://github.com/optuna/optuna/pull/3337/files#diff-e52a24e2c14ee4ff0d8b5e62f38c4e515ec3dad12446570f0771ef74d27689faL314)). Does `sample_independent` return the same suggestion only by re-initializing `TPESampler` with reseed?

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
```diff
@@ -244,58 +244,39 @@ def test_multi_objective_sample_independent_misc_arguments() -> None:
...
+            frozen_trial_factory(
+                i, [random.random(), random.random()], dist=float_dist, value_fn=value_fn
+            )
+            for i in range(16)
+        ]
+    else:
+        past_trials = [
+            frozen_trial_factory(i, [random.random(), random.random()], dist=float_dist)
+            for i in range(16)
+        ]
```
himkt, line 278: I'm wondering if we can simplify it based on the current approach like following:

```python
    if float_dist.step:
        value_fn: Optional[Callable[[int], int]] = lambda number: int(random.random() * 1000) * 0.1
    else:
        value_fn = None

    past_trials = [
        frozen_trial_factory(i, [random.random(), random.random()], dist=float_dist, value_fn=value_fn)
        for i in range(16)
    ]
```

What do you think?

hvy:
@HideakiImamura could you jump in as a second reviewer?
Cc. @xadrianzetx if you have any comments on the changes. 🙇

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
```diff
@@ -309,29 +290,15 @@ def test_multi_objective_sample_independent_log_uniform_distributions() -> None:
...
-        return int(random.random() * 1000) * 0.1
-
-    past_trials = [
-        frozen_trial_factory(
-            i, [random.random(), random.random()], dist=disc_dist, value_fn=value_fn
-        )
-        for i in range(16)
-    ]
+    if float_dist.step == 0.1:
+        assert abs(int(float_suggestion * 10) - float_suggestion * 10) < 1e-3
```
xadrianzetx, line 297: This test will be silently omitted if we change/add step parametrization.

xadrianzetx (commented):
Thanks for picking up those remaining tasks! I have few comments.

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
nyanhi: Agreed, I couldn't figure out how to pass mypy check and ended up like that. Thank you very much for letting me know the solution with type annotation!

I think `Optional[Callable[[int], float]]` should be used but please let me know if otherwise.

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
nyanhi: Thank you so much for pointing out, I have fixed the conditioning branch to 
`if float_dist.log or float_dist.step == 1.0`
and re-seeded.

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
nyanhi: Thank you for heads up! Given the step parametrization remains unchanged, I think this part is fine but kindly let me know if I missed anything.

xadrianzetx: I think we can just simply do
```suggestion
    if float_dist.step is not None:
        assert abs(int(float_suggestion * 10) - float_suggestion * 10) < 1e-3
```
Also, `int` always rounds down to the nearest integer, so i guess this test will always pass since `0.0 <= float_suggestion - int(float_suggestion) < 1.0` is always `True`

nyanhi: If I understood correctly, this test ensures that `float_suggestion` is at most one decimal place when the `FloatDistribution` given `step=0.1` (In other word, test fails let's say when `float_suggestion = 1.23`).
Hence, I think the condition branch should be explicitly `== 0.1`.

Please kindly let me know if I misunderstood any point.

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
himkt: > I think Optional[Callable[[int], float]] should be used but please let me know if otherwise.

You're totally right, my annotation was wrong. :bow:

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
xadrianzetx: Ok, let's leave it for now then and maybe revisit if parametrization changes. :+1:

xadrianzetx (approved):

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
```diff
@@ -388,24 +353,32 @@ def cat_value_fn(idx: int) -> float:
...
+    int_dist = optuna.distributions.IntDistribution(1, 100, log, step)
     past_trials = [
         frozen_trial_factory(
             i, [random.random(), random.random()], dist=int_dist, value_fn=int_value_fn
         )
         for i in range(16)
     ]
 
-    trial = frozen_trial_factory(16, [0, 0])
+    trial = frozen_trial_factory(16, [1, 1])
```
himkt, line 381:
```suggestion
    trial = frozen_trial_factory(16, [0, 0])
```

Please correct me if I missed something but I think we don't have to change here (it's values of an objective but parameters).

himkt (commented):
Thank you @nyanhi! Looks solid! I left the one minor suggestion, PTAL. :pray:

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
nyanhi: Yes, let me revert the change. Thank you.

himkt (approved):
Thank you @nyanhi for the PR, LGTM!

PR #3338, opened 2022-02-27
Move the get_trial_xxx abstract functions to base

belldandyxtq:
## Motivation
This PR is a part of https://github.com/optuna/optuna/issues/2943 to move the `get_trial_xxx` abstract functions to base class so that user do not need to implement them again

## Details
The following functions are removed from each subclass, and implemented in the `BaseStorage` class
```
get_trial_number_from_id
get_trial_param
get_trial_params
```

## Note
The following implementations are preserved as they are optimized for performance

- `get_trial_param` in `rdb`
- `get_trial_number_from_id` in `n_memory`

### DIFF
--- a/optuna/storages/_in_memory.py
+++ b/optuna/storages/_in_memory.py
@@ -315,6 +315,14 @@ class InMemoryStorage(BaseStorage):
                 )
             return self.get_trial(best_trial_id)
 
+    def get_trial_param(self, trial_id: int, param_name: str) -> float:
+
+        with self._lock:
+            trial = self._get_trial(trial_id)
+
+            distribution = trial.distributions[param_name]
+            return distribution.to_internal_repr(trial.params[param_name])
+
     def set_trial_values(self, trial_id: int, values: Sequence[float]) -> None:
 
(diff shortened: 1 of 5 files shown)


optuna/storages/_in_memory.py
```diff
@@ -315,14 +315,6 @@ def get_best_trial(self, study_id: int) -> FrozenTrial:
                 )
             return self.get_trial(best_trial_id)
 
-    def get_trial_param(self, trial_id: int, param_name: str) -> float:
-
-        with self._lock:
-            trial = self._get_trial(trial_id)
-
-            distribution = trial.distributions[param_name]
-            return distribution.to_internal_repr(trial.params[param_name])
```
HideakiImamura, line 324: These logics is in the lock. Do you intend to change the behavior of get_trial_param of InMemoryStorage in this PR?

belldandyxtq: Sorry, I have put them back

HideakiImamura (approved):
LGTM!

PR #3347, opened 2022-02-27
Use coverage directly

higucheese:
## Motivation

To fix https://github.com/optuna/optuna/issues/2517. This change affects the CI behavior, so I want to test it here.

## Description of the changes

- Filtering `ExperimentalWarning`, which was reverted previously
- Replace `pytest --cov` with `coverage run -m pytest`

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+concurrency = multiprocessing,thread
+source = optuna/integration
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,3 @@
 [run]
 concurrency = multiprocessing,thread
-source = optuna/integration
+source = optuna/
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -59,18 +59,18 @@ jobs:
       env:
         OMP_NUM_THREADS: 1
       run: |
-        coverage run --source=optuna -m pytest tests -m "not skip_coverage" \
+        coverage run --source=optuna -m pytest --collect-only tests -m "not skip_coverage" \
             --ignore tests/integration_tests/test_pytorch_lightning.py
         coverage combine
-        coverage xml
+        pytest --cov=optuna --cov-report=xml --cov-append tests -m "not skip_coverage"
 
     - name: Multi-node tests
       env:
         OMP_NUM_THREADS: 1
       run: |
-        mpirun -n 2 -- coverage run -m pytest tests/integration_tests/test_chainermn.py
+        mpirun -n 2 -- coverage run -m pytest --collect-only tests/integration_tests/test_chainermn.py
         coverage combine --append
-        coverage xml
+        mpirun -n 2 -- pytest --cov=optuna --cov-report=xml --cov-append tests/integration_tests/test_chainermn.py
 
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v1
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -61,13 +61,15 @@ jobs:
       run: |
         coverage run --source=optuna -m pytest tests -m "not skip_coverage" \
             --ignore tests/integration_tests/test_pytorch_lightning.py
+        coverage combine
         coverage xml
 
     - name: Multi-node tests
       env:
         OMP_NUM_THREADS: 1
       run: |
-        mpirun -n 2 -- coverage run --append --source=optuna -m pytest tests/integration_tests/test_chainermn.py
+        mpirun -n 2 -- coverage run -m pytest tests/integration_tests/test_chainermn.py
+        coverage combine --append
         coverage xml
 
     - name: Upload coverage to Codecov
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -60,9 +60,9 @@ jobs:
     - name: Tests
       env:
         OMP_NUM_THREADS: 1
-        PYTHONPATH: .
-        COVERAGE_PROCESS_START: .coveragerc
-        COVERAGE_COVERAGE: yes
+        PYTHONPATH: .  # To invoke sitecutomize.py
+        COVERAGE_PROCESS_START: .coveragerc  # https://coverage.readthedocs.io/en/6.4.1/subprocess.html
+        COVERAGE_COVERAGE: yes  # https://github.com/nedbat/coveragepy/blob/65bf33fc03209ffb01bbbc0d900017614645ee7a/coverage/control.py#L255-L261
       run: |
         coverage run --source=optuna -m pytest tests -m "not skip_coverage" \
             --ignore tests/integration_tests/test_pytorch_lightning.py
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -55,9 +55,14 @@ jobs:
         pip install --progress-bar off .[optional]
         pip install --progress-bar off .[integration] --extra-index-url https://download.pytorch.org/whl/cpu
 
+        echo 'import coverage; coverage.process_startup()' > sitecustomize.py
+
     - name: Tests
       env:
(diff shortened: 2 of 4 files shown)


higucheese:
The main problem I targeted seems to be fixed! But somehow the code coverage decreased a bit.

hvy:
Thank you for the PR. 🙇 
@himkt, could you jump in as an additional reviewer here?

higucheese:
The `coverage` cannot trace multi processes running concurrently by default, and this should be the cause of the degradation. To fix this, I have added `.coveragerc`, a configuration file for `coverage`, in order to specify `concurrency` option (somehow this cannot be specified in the command line). ref: https://coverage.readthedocs.io/en/6.3.2/config.html#run

higucheese:
There is still the remaining problem of not tracing `__setstate__` of storage classes. This is a command to reproduce.
```sh
$ coverage run -m pytest tests/integration_tests/test_pytorch_lightning.py::test_pytorch_lightning_pruning_callback_ddp_monitor
```
I'm going to investigate this furthermore.

himkt:
Hi @higucheese, I added tests to serialize/deserialize storage in https://github.com/optuna/optuna/pull/3407. Could you please try to merge/rebase the master branch to kick coverage again?

higucheese:
@himkt I'm sorry for missing the comment. I have kicked CI with the latest master. Really thanks for the improvement of tests.

contramundum53:
@higucheese Let me kindly ping you. Is there any progress on https://github.com/optuna/optuna/pull/3347#issuecomment-1057044883 ?

higucheese:
@contramundum53 Thank you for the remind. In the last optuna dev sprint, I investigated the difference of how `coverage` and `pytest-cov` take traces of multi-threaded executions using PyTorch or its family which provide an original multi-threading method. Although I still do not figure out where the coverage difference comes, I think invoking `coverage` in the beginning of all Python processes may solve the coverage regression problem. ref: https://coverage.readthedocs.io/en/6.3.2/subprocess.html#subprocess

I'm sorry for no activity in May. I am going to make time to work on in this weekend.

HideakiImamura:
@higucheese @himkt Let me remove my assignment until the WIP is removed.

higucheese:
### Why adding `filterwarnings` results in 0 % coverage

When filtering a user-defined `Warnings`, it is necessary to scan the source for the definition before starting pytest. In doing so, the entire source is read, resulting in very few lines being traced by pytest-cov. Note that the built-in `Warning` does not cause this problem because it does not need to look for the definition.

### Reasons to use `coverage` instead of `pytest-cov

Because the tracer is also invoked when looking for a user-defined `Warning` definition.

### Why `coverage` has less coverage than `pytest-cov` in normal use.

`coverage run` does not trace for child processes by default. Child processes created by standard libraries are supported by setting the `concurrency` option, but some of them may still be untraceable.

#### How `pytest-cov` avoids the weakness

It allocates tracers as global variables and implements a mechanism to follow child processes through environment variables.

### Why normal `subprocess` tracking is not enough

These are hypothesises.

- The PyTorch family does not track well against the multi-process libraries implemented by the PyTorch family.
- It does not track well with complex process trees that include callbacks.

### Why use `sitecustomize.py`?

By invoking tracers at the same time as any Python processes are invoked, we can trace all child processes.

### What kind of environment variable is `COVERAGE_COVERAGE`?

The name of the environment variable is whether or not to measure coverage for the `coverage` itself. By setting it, tracers that are not recognized by the main tracer of `coverage run` will also dump results.

setup.py
```diff
@@ -132,7 +132,6 @@ def get_extras_require() -> Dict[str, List[str]]:
             "fakeredis<=1.7.1; python_version<'3.8'",
             "fakeredis ; python_version>='3.8'",
             "pytest",
-            "pytest-cov",
```
himkt, line 135: https://github.com/codecov/codecov-python/blob/12c0dddd6350a1d511f22d363e89d9019b149164/setup.py#L45

[memo] `coverage` is installed as the dependency of `codecov`. Although it might beyond this PR's context, it might be safer to add `coverage` in `test` extras.

.coveragerc
```diff
@@ -0,0 +1,3 @@
+[run]
+concurrency = multiprocessing,thread
```
himkt, line 2: [memo] `concurrency` supports comma-separated string.

- https://github.com/nedbat/coveragepy/blob/8a5049e0fe6b717396f2af14b38d67555f9c51ba/coverage/config.py#L354
- https://github.com/nedbat/coveragepy/blob/8a5049e0fe6b717396f2af14b38d67555f9c51ba/coverage/config.py#L408-L419
- https://github.com/nedbat/coveragepy/blob/8a5049e0fe6b717396f2af14b38d67555f9c51ba/coverage/config.py#L92-L108

.github/workflows/coverage.yml
```diff
@@ -55,18 +55,30 @@ jobs:
...
             --ignore tests/integration_tests/test_pytorch_lightning.py
+        coverage combine
+        coverage xml
 
     - name: Multi-node tests
       env:
         OMP_NUM_THREADS: 1
+        PYTHONPATH: .
+        COVERAGE_PROCESS_START: .coveragerc
+        COVERAGE_COVERAGE: yes
```
himkt, line 77: > The purpose of the environment variable is whether or not to measure coverage for the coverage itself. By setting it, tracers that are not recognized by the main tracer of coverage run will also dump results.

Is this documented anywhere? The reference to the document will help for developers in the future. I only found the source code. 😇 

https://github.com/nedbat/coveragepy/blob/65bf33fc03209ffb01bbbc0d900017614645ee7a/coverage/env.py#L119-L120

himkt (approved):
Thank you so much @higucheese for the deep investigation. We really appreciate your effort and enabling coverage calculation with `filterwarnings` will improve our development experience.

> The PyTorch family does not track well against the multi-process libraries implemented by the PyTorch family.
> It does not track well with complex process trees that include callbacks.

Since unsupported parallel execution backends are typically used in integrations, I think it is acceptable to give up the 100% coverage. I'm little bit worry about `It does not track well with complex process trees that include callbacks` but the result doesn't show the significant degradation. I think it's OK as well.

.github/workflows/coverage.yml
higucheese: I cannot find any explanation of `COVERAGE_COVERAGE` in [the official doc](`https://coverage.readthedocs.io/en/6.4.1/index.html`) neither. 😇 I think this environment variable is for `coveragepy` developers, and its purpose is shortly explained in the source code here.

https://github.com/nedbat/coveragepy/blob/65bf33fc03209ffb01bbbc0d900017614645ee7a/coverage/control.py#L255-L261

.github/workflows/coverage.yml
```diff
@@ -55,18 +55,30 @@ jobs:
...
         pip install --progress-bar off .[integration] --extra-index-url https://download.pytorch.org/whl/cpu
 
+        echo 'import coverage; coverage.process_startup()' > sitecustomize.py
+
     - name: Tests
       env:
         OMP_NUM_THREADS: 1
+        PYTHONPATH: .
+        COVERAGE_PROCESS_START: .coveragerc
+        COVERAGE_COVERAGE: yes
```
himkt, line 65:
```suggestion
        COVERAGE_COVERAGE: yes  # https://github.com/nedbat/coveragepy/blob/65bf33fc03209ffb01bbbc0d900017614645ee7a/coverage/control.py#L255-L261
```

.github/workflows/coverage.yml
himkt: I see, thank you! So I'm wondering if it'd be better to leave the comment in the configuration (https://github.com/optuna/optuna/pull/3347/files#r912462028). :pray:

.github/workflows/coverage.yml
higucheese: Looks good! I added some comments to them in abe39b7 in addition to your suggestion.

Issue #3205, opened 2021-12-24
`nan` handling in Optuna


HideakiImamura:
A new version of https://github.com/optuna/optuna/issues/2169 adapted to the latest Optuna.
And related to https://github.com/optuna/optuna/issues/3132.

[note: fixed 2022/01/30]

## Expected behavior

~Regardless of the type of storage used, mark the corresponding trial as `FAILED` when an objective function returns `nan`, when `nan` is passed in `trial.report`, or when `nan` is passed in `study.tell`.~
Regardless of the type of storage used, mark the corresponding trial as `FAILED` when an objective function returns `nan` or when `nan` is passed in `study.tell`.
On the other hand, regardless of the type of storage used, accept the reported `nan` when `nan` is reported with `trial.report`.

We need to fix the following.
- [x] ~If `nan` is passed in `trial.report`, mark the trial as `FAILED` regardless of the type of storage used.~ If `nan` is passed in `trial.report`, accept the reported `nan` regardless of the type of storage used. #3348
- [x] ~Add document for workaround when passing `nan` in `trial.report`. In a trial, if the intermediate value is `nan` at the beginning of the calculation, but it reverts to the appropriate value in the middle, the user can check if it is not `nan` before `trial.report`.~
- [x] If `nan` is passed in `study.tell`, mark the trial as `FAILED` regardless of the type of storage used. https://github.com/optuna/optuna/pull/3144 

## Environment

- Optuna version: a70cbff8dbc454f137b61622633404bd4e4657fc
- Python version: 3.9.9
- OS: macOS Catalina 10.15.7
- (Optional) Other libraries and their versions:
psycopg2:    2.9.2
PyMySQL:     1.0.2

## Error messages, stack traces, or logs

If we run the script `nan-objective.py`, we will get the following output. The script is provided in the section of `Reproducible examples`. I think this output is ideal and does not need to be modified. We should mark the trial as `FAILED` for all storages.
<details>
<summary>output</summary>

```
(venv) mamu@HideakinoMacBook-puro test % python nan-objective.py 
[I 2021-12-24 13:36:27,998] A new study created in memory with name: no-name-769e5aec-62ec-4719-ac85-6a6fbf375bee
[W 2021-12-24 13:36:28,002] Trial 0 failed, because the objective function returned nan.
In-memory : Trial state is TrialState.FAIL
[...]

PR #3348, opened 2022-02-27
Accept `nan` in `trial.report`

belldandyxtq:
## Motivation
This is a part of https://github.com/optuna/optuna/issues/3205 to support `nan` in `trial.report`. As shown in the issue, `trial.report` fails when the backend is `MySQL`, since `MySQL` cannot store `nan`.

## Description of the changes
1. This PR changes the `nan` to `None` to store in the backend.
2. Add tests for `nan` in `report`.


optuna/storages/_rdb/storage.py
```diff
@@ -49,6 +50,17 @@
 _logger = optuna.logging.get_logger(__name__)
 
 
+def _ensure_not_nan(value):
+    # Ensure the value is not Nan, which is not supported by MySQL
+    # if Nan, change it the None
+    if isinstance(value, tuple) or isinstance(value, list):
```
nzw0301, line 56: This can be simplified:
```suggestion
    if isinstance(value, (tuple, list)):
```

belldandyxtq:
Should we change `None` back to `nan` in `get`?

hvy:
Thanks for the PR.
@himkt , let me know if you're busy, but could you otherwise review this change?

HideakiImamura:
@belldandyxtq Please apply the `formats.sh` to pass the CI.

HideakiImamura (approved):
Thanks for the mob review! LGTM.

[TODO] We need to add a migration script. I must be done until the next release, but it should be ASAP.

HideakiImamura:
I fixed the conflict. 

@himkt Please merge this PR after your approval.

himkt (approved):
LGTM.

Note that after this PR being merged, `trial.report(step=x, value=float("nan"))` does not raise an exception without writing an intermediate entry until a storage migration. The current behavior is that stops an optimization with raising an exception. Users who stops the optimization using `trial.report` + `value=nan` may observe this behavior change (but we think no one rely on this behavior, since pruners consider trials to be pruned when `nan` exists in intermediates).

Issue #2981, opened 2021-10-05
Support Python 3.10


hvy:
## Motivation

Python 3.10 has been released https://www.python.org/downloads/release/python-3100/.

Since 3.10 users can be expected, we should make sure that Optuna works with this version.

## Description

Add 3.10 to the unit test and [`optuna-example`](https://github.com/optuna/optuna-examples/) CI matrices. Make sure that all jobs complete successfully. Note that necessary Docker images and other tools may be missing, yet.

## Additional context (optional)

See also the following issues/PRs from 3.9 for reference.

- https://github.com/optuna/optuna/pull/1908
- https://github.com/optuna/optuna/issues/2034

toshihikoyanase:
We'd like to check this issue before releasing v3. Let me add the `v3` label.

semiexp:
I tested the compatibility with Python 3.10 in #3179, but `pip install torch` failed (`torch` does not support 3.10 yet): https://github.com/optuna/optuna/runs/4572899427?check_suite_focus=true

hugovk:
Here's the torch issue: https://github.com/pytorch/pytorch/issues/66424

nzw0301:
FYI: PyTorch 1.11.0 has been released! This version starts to [support Python 3.10](https://github.com/pytorch/pytorch/releases/tag/v1.11.0).

HideakiImamura:
We will work on PyTorchLightningCallback in [another issue](https://github.com/optuna/optuna/issues/3418), so we will remove it from the TODO of this issue.

PR #3353, opened 2022-03-01
Partially support Python 3.10

nzw0301:
## Motivation

Partially resolve https://github.com/optuna/optuna/issues/2981.

## Description of the changes

- Add Python 3.10 to `python_version ` of CIs files.
- Run tests except for PyTorch-related tests with python 3.10 
    - As tested by https://github.com/optuna/optuna/pull/3179, PyTorch does not support Python 3.10 yet.
- Update `tox.ini`: make integration test consistent with CIs for Python 3.6 and add Python 3.10


nzw0301:
Can I remove this part?
https://github.com/optuna/optuna/blob/58d8d844264dddf4599cadf08908fb0475a31894/.github/workflows/tests.yml#L51-L54

hvy:
@toshihikoyanase , @keisuke-umezawa could you review this PR? Also, cc. @himkt if you have any inputs.

.github/workflows/tests-integration.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
```
keisuke-umezawa, line 21: Better to use single quotation to align other lines

tests/samplers_tests/test_samplers.py
```diff
@@ -53,9 +53,15 @@
...
-            lambda: optuna.integration.BoTorchSampler(n_startup_trials=0),
             lambda: optuna.samplers.QMCSampler(),
         ]
+    )
+    # TODO(nzw0301): Remove version constraints if BoTorch supports Python 3.10
+    # or Optuna does not support Python 3.6.
+    + (
+        []
+        if sys.version_info >= (3, 10, 0) or sys.version_info < (3, 7, 0)
+        else [lambda: optuna.integration.BoTorchSampler(n_startup_trials=0)]
```
keisuke-umezawa, line 64: I am not against this solution, but how will it remind us later?

tests/samplers_tests/test_samplers.py
```diff
@@ -53,9 +53,15 @@
         []
         if sys.version_info < (3, 7, 0)
         else [
-            lambda: optuna.integration.BoTorchSampler(n_startup_trials=0),
             lambda: optuna.samplers.QMCSampler(),
         ]
+    )
+    # TODO(nzw0301): Remove version constraints if BoTorch supports Python 3.10
+    # or Optuna does not support Python 3.6.
```
keisuke-umezawa, line 60: I am not against this solution, but how will it remind us later?

keisuke-umezawa (commented):
Overall lgtm. I just left a few comments.

tests/samplers_tests/test_samplers.py
nzw0301: As you can see, this PR adds to-do lines to L59 -- L60 in this file. If you mean that the todo is not enough, I would like to add explicit tasks to https://github.com/optuna/optuna/issues/2981 for the followups.

.github/workflows/tests-integration.yml
nzw0301: Thank you for your suggestion. I've updated all `python-version` in this PR.

tests/samplers_tests/test_samplers.py
keisuke-umezawa: Thanks! That would be a good solution!

> I would like to add explicit tasks to https://github.com/optuna/optuna/issues/2981 for the followups.

keisuke-umezawa (approved):
LGTM!

hvy:
Just a comment but if we are to update `setup.py`, perhaps the README badge should be updated as well https://github.com/optuna/optuna/blob/master/README.md?plain=1#L5 . It could very much be a separate PR but I think we have to do it manually?

nzw0301:
Thank you for your comment! Good point. I suppose so. Here is the new link:
https://img.shields.io/badge/python-3.6%20%7C%203.7%20%7C%203.8%20%7C%203.9%20%7C%203.10-blue.

.github/workflows/dockerimage.yml
```diff
@@ -27,13 +27,13 @@ jobs:
...
+    # TODO (nzw0301): Remove `exclude` once integration works with Python 3.10.
     strategy:
       matrix:
-        python_version: ['3.6', '3.7', '3.8', '3.9']
+        python_version: ['3.6', '3.7', '3.8', '3.9', '3.10']
         build_type: ['', 'dev']  # "dev" installs all the dependencies including pytest.
         exclude:
-        - python_version: '3.9'
+        - python_version: '3.10'
           build_type: 'dev'
```
toshihikoyanase, line 37: Python 3.9-dev image will be newly published. I don't think it problematic because of CI results.

toshihikoyanase (approved):

Note that this PR includes the following changes in addition to the Python 3.10 support:
- release of Python 3.9 dev image
- remove Cython from setup of GitHub Actions,

Issue #3245, opened 2022-01-21
Introduce a rich benchmark environment in CI


HideakiImamura:
This is the prerequisite task for #2964, #2906 and #2951.

## Motivations

Recently, a benchmarking environment using [`kurobako`](https://github.com/optuna/kurobako) was introduced in Optuna for benchmarking the performance evaluation of sampling and pruning algorithms. The benchmarking feature itself is still experimental and we need to add more problems/features.

Please feel free to leave any comments/feedback for the benchmarking feature.

## What should we do

- [x] Support the multi-objective problem. In `kurobako`, there is some multi-objective problem such that ZDD and NAS-bench (C). We can just use those problems. In addition, visualization of benchmark results is different from single-objective problems. We can show the pareto-front plot and optimization-history plot for each objective coordinate. (We should discuss this point.) #3271
    - [ ] https://github.com/optuna/optuna/pull/3691 
- [x] Support problems other than `kurobako`. We plan to support [Bayesmark](https://github.com/uber/bayesmark). #3354. See https://github.com/optuna/optuna/issues/3464 for the remaining tasks.
- [x] Make NAS-Bench-201 and 301 available in benchmark. See https://arxiv.org/abs/2201.13396 and https://arxiv.org/abs/2008.09777.
    - [x] https://github.com/optuna/optuna/pull/3465
    - ~[ ] NAS-Bench-301~
- [x] Make WFG problems available in benchmark. See https://github.com/HideakiImamura/WFGtestSuite. #3349

## References
See https://github.com/optuna/optuna/tree/master/benchmarks for more details about the current benchmark environment.

HideakiImamura:
Note that @xadrianzetx is working on the second item in https://github.com/xadrianzetx/dev-optuna-bayesmark.

HideakiImamura:
Now, NAS-Bench-201 and 301 are proposed, but they are not available in kurobako. It would be great to make it available in kurobako. 
- https://arxiv.org/abs/2201.13396
- https://arxiv.org/abs/2008.09777

HideakiImamura:
Note that @kei-mo is working on the fourth item.

hvy:
For the record, let me add an additional reference https://github.com/optuna/optuna/pull/3155.

HideakiImamura:
Note that @contramundum53 will work on the third item.

divyanshugit:
@HideakiImamura  If anything is left in this issue then I would love to work on it.

[...]

PR #3354, opened 2022-03-01
Add performance benchmarks using `bayesmark`

xadrianzetx:
## Motivation

As a part of #3245, this PR extends CI benchmarking environment by adding support for [bayesmark](https://github.com/uber/bayesmark). New workflow allows to test optimizers in parallel by performing hyperparameter search for several ML models fitted to a selection of toy datasets, all provided in [scikit-learn](https://scikit-learn.org/stable/).

Dataset list:
* [Breast Cancer Wisconsin](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_breast_cancer.html#sklearn.datasets.load_breast_cancer)
* [Diabetes Data Set](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_diabetes.html#sklearn.datasets.load_diabetes)
* [Digits](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_digits.html#sklearn.datasets.load_digits)
* [Iris](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_iris.html#sklearn.datasets.load_iris)
* [Wine](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_wine.html#sklearn.datasets.load_wine)

Model list:
* [AdaBoost](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.AdaBoostClassifier.html)
* [Decistion Tree](https://scikit-learn.org/stable/modules/generated/sklearn.tree.DecisionTreeClassifier.html)
* [kNN](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html)
* [Linear Model](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html)
* [Multi-layer Perceptron](https://scikit-learn.org/stable/modules/generated/sklearn.neural_network.MLPClassifier.html)
* [Random Forest](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html)
* [SVM](https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html)

Benchmark generates markdown report similar in style to one provided by [kurobako](https://github.com/optuna/kurobako) as well as visualizations for each solved problem. The workflow can be parametrized and triggered manually similarly to kurobako workflow.

## Description of the changes

* [x] Add benchmark script using `bayesmark`
* [x] Add markdown report builder script and template
* [x] Implement CI workflow to run benchmark in distributed fashion
* [x] Updated `benchmarks/README.md`

## Examples
Example benchmark run can be found [here](https://github.com/xadrianzetx/optuna-bayesmark/actions/runs/1922672771). More on [bayesmark scoring](https://bayesmark.readthedocs.io/en/stable/scoring.html).


benchmarks/bayesmark/optuna_optimizer.py
```diff
@@ -0,0 +1,116 @@
...
+
+_PRUNERS = {
+    "NopPruner": pruners.NopPruner,
+    "MedianPruner": pruners.MedianPruner,
+    "PatientPruner": pruners.PatientPruner,
+    "PercentilePruner": pruners.PercentilePruner,
+    "SuccessiveHalvingPruner": pruners.SuccessiveHalvingPruner,
+    "HyperbandPruner": pruners.HyperbandPruner,
+    "ThresholdPruner": pruners.ThresholdPruner,
+}
```
xadrianzetx, line 31: Note - Currently all pruners are `nop` since vanilla bayesmark does not support partial fit. Leaving for possible future extensions.

benchmarks/run_bayesmark.py
```diff
@@ -0,0 +1,159 @@
...
+
+    with open(os.path.join("benchmarks", "bayesmark", "config.json"), "w") as file:
+        json.dump(config, file, indent=4)
+
+    samplers = " ".join(config.keys())
+    metric = "nll" if args.dataset in ["breast", "iris", "wine", "digits"] else "mse"
+    cmd = (
+        f"bayesmark-launch -n {args.budget} -r {args.repeat} "
+        f"-dir runs -b {_DB} "
+        f"-o RandomSearch {samplers} "
```
xadrianzetx, line 52: Note - Builtin random sampler is used here since it's automatically detected by bayesmark to be used as baseline for leaderboards. Otherwise we would have to specify one.

hvy:
@himkt could you jump in as a second reviewer?

benchmarks/bayesmark/optuna_optimizer.py
```diff
@@ -0,0 +1,116 @@
...
+import numpy as np
+
+from bayesmark.abstract_optimizer import AbstractOptimizer
+from bayesmark.experiment import experiment_main
+import optuna
+from optuna import pruners
+from optuna import samplers
+
+
+_SAMPLERS = {
```
HideakiImamura, line 15: I don't want to restrict the sampler in the current ones. I mean this benchmark should be feasible for any samplers we add to `optuna.samplers` in the future. How about instantiating the sampler class from its name as follows?

```python
if sampler_name in optuna.samplers.__dict__:
    try:
        if self.param.sampler in ["BoTorchSampler", "PyCmaSampler"]:
            sampler = getattr(optuna.integration, self.param.sampler)(**sampler_kwargs)
        else:
            sampler = getattr(optuna.samplers, self.param.sampler)(**sampler_kwargs)
        assert isinstance(sampler, optuna.samplers.BaseSampler)
    except TypeError:
        raise TypeError(
            f"__init__() of {self.param.sampler} got an unexpected keyword "
            f"argument. Given arguments are {sampler_kwargs}."
        )
```

benchmarks/bayesmark/optuna_optimizer.py
```diff
@@ -0,0 +1,116 @@
...
+    "GridSampler": samplers.GridSampler,
+    "RandomSampler": samplers.RandomSampler,
+    "TPESampler": samplers.TPESampler,
+    "CmaEsSampler": samplers.CmaEsSampler,
+    "NSGAIISampler": samplers.NSGAIISampler,
+}
+
+_PRUNERS = {
```
HideakiImamura, line 23: Ditto.

benchmarks/bayesmark/optuna_optimizer.py
```diff
@@ -0,0 +1,116 @@
...
+
+        suggestions: Suggestion = dict()
+        for name, config in self.api_config.items():
+            low, high = config["range"]
+            log = config["space"] == "log"
+
+            if config["space"] == "logit":
+                low = np.log(low / (1 - low))
+                high = np.log(high / (1 - high))
+
```
HideakiImamura, line 79: This is not a strong opinion, but how about validating the config here? E,g, if `config["space"] == "logit"`, we need `0 < low <= high < 1`.

benchmarks/bayesmark/optuna_optimizer.py
```diff
@@ -0,0 +1,116 @@
...
+            if config["type"] == "real":
+                param = trial.suggest_float(name, low, high, log=log)
+
+            elif config["type"] == "int":
+                param = trial.suggest_int(name, low, high, log=log)
+
+            else:
+                # TODO(xadrianzetx) Support `suggest_categorical` if benchmark is extended.
```
HideakiImamura, line 87: :+1:

HideakiImamura (commented):
Thanks for the PR and sorry for the late. I checked all of codes, and have several comments. Please take a look.

benchmarks/bayesmark/optuna_optimizer.py
xadrianzetx: Done in 569e1f1b51f5da664d4dde142b765e7df9c32abe.

benchmarks/run_bayesmark.py
xadrianzetx: This way `bayesmark` automatically derives baseline run. From the doc
> The bayesmark-anal command looks for a baseline.json file in [DB_ROOT]/[DBID]/derived, which states the best possible and random search performance. If no such file is present, bayesmark-anal automatically calls bayesmark-baseline to build it. The baselines are inferred from the random search performance in the logs. The baseline values are considered fixed (not random) quantities when bayesmark-anal builds confidence intervals. Therefore, we allow the user to leave them fixed and do not rebuild them when bayesmark-anal is called if a baselines file is already present.

Without baseline run benchmark [will fail](https://github.com/xadrianzetx/dev-optuna-bayesmark/runs/5566869587?check_suite_focus=true). We could use Optuna random or grid samplers instead, but that's more code for essentially the same effect.

benchmarks/bayesmark/optuna_optimizer.py
HideakiImamura: Sounds reasonable. I agree with the current design as you explained.

benchmarks/report_bayesmark.py
```diff
@@ -0,0 +1,92 @@
+from collections import Counter
+import io
+import json
+import os
+from typing import Dict
+
+
+_TABLE_HEADER = "\n|Ranking|Solver|Score|\n|---:|:---|---:|\n"
```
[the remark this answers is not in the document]
HideakiImamura: Hmm... I see. Then does that mean that the line in the output of plot does not correspond to this score? If they correspond, I was wondering if it would be possible to get the mean and standard deviation based on the plot.

HideakiImamura (commented):
Sorry foe the too late. I have several comments. Could you take a look?

benchmarks/report_bayesmark.py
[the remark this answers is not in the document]
xadrianzetx: It does correspond (score seems to be [`100 * (1 - normalized mean)`](https://github.com/uber/bayesmark/blob/master/docs/scoring.rst#mean-scores)), however error bars on that plot do not show standard deviation, but rather are based on  [t-statistic](https://github.com/uber/bayesmark/blob/master/docs/scoring.rst#error-bars). Also those numbers are kept as precalculated aggregates (mean, lower and upper bound for each optimizer at any step), hence my comment about re-writing the aggregation mechanism to get additional stats (which might not be a bad idea).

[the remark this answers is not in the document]
HideakiImamura: Thanks for the explanation. I think we can re-write the aggregation mechanism as a follow-up task.

HideakiImamura (approved):
Thanks, LGTM!

Note that we can re-write the aggregation mechanism of the results by our own as a follow-up task.

.github/workflows/performance-benchmarks-bayesmark.yml
```diff
@@ -0,0 +1,112 @@
...
+        python benchmarks/report_bayesmark.py
+    
+    - uses: actions/upload-artifact@v2
+      with:
+        name: benchmark-report
+        path: report
+    
+    - uses: geekyeggo/delete-artifact@v1
+      with:
+        name: partial-reports
```
himkt, line 112: Missing newline. 🍵

benchmarks/bayesmark/report_template.md
```diff
@@ -0,0 +1,36 @@
+# Benchmark Result Report
+
+* Number of Solvers: {num_solvers}
+* Number of Models: {num_models}
+* Number of Datasets: {num_datasets}
+* Number of Problems: {num_problems}
+
+Final score for each problem is calculated as `100 x (1-loss)`. Solver with lowest score in each problem wins. For more details visit [bayesmark docs.](https://bayesmark.readthedocs.io/en/stable/scoring.html)
```
himkt, line 8: Given a solver with loss = 0 (good) and other with loss = 1. Scores become 100 * (1-0) = 100 and 100 * (1-1) = 0 respectively. If `Solver with lowest score in each problem wins`, the latter solver (loss = 1) wins. But it doesn't make sense, I think.

benchmarks/bayesmark/optuna_optimizer.py
```diff
@@ -0,0 +1,117 @@
...
+from optuna import pruners
+from optuna import samplers
+
+
+_SAMPLERS = {
+    "GridSampler": samplers.GridSampler,
+    "RandomSampler": samplers.RandomSampler,
+    "TPESampler": samplers.TPESampler,
+    "CmaEsSampler": samplers.CmaEsSampler,
+    "NSGAIISampler": samplers.NSGAIISampler,
```
himkt, line 20:
```suggestion
    "NSGAIISampler": samplers.NSGAIISampler,
    "QMCSampler": samplers. QMCSampler,
```

Can we add QMCSampler? Or NO to support older versions of Optuna?

benchmarks/report_bayesmark.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def assemble_report(self) -> str:
+
+        num_datasets = len(self._datasets)
+        num_models = len(self._models)
+
+        with open(os.path.join("benchmarks", "bayesmark", "report_template.md")) as file:
+            report_template = file.read()
+
+        # TODO(xadrianzetx) Consider using proper templating engine.
```
himkt, line 56: :+1: for it. We may be able to use a template engine such as jinja. What concerns does it block the task (I think the current implementation is fine, just for curiosity)?

himkt (commented):
Thank you @xadrianzetx for the nice PR! Although I'm not familiar with the benchmark mechanism but I left some comments.

benchmarks/bayesmark/report_template.md
xadrianzetx: I think you're right. In our case `loss` is normed CV score returned by last objective function evaluation, averaged across runs. Since we are using MSE for regression and negative log-likelihood for classification, which are both minimized, behavior you're describing would be correct (highest score wins). CC @HideakiImamura

benchmarks/report_bayesmark.py
xadrianzetx: Yeah, I think we are good for now. Template engine would be nice if we wanted to add more fidelity to the report (like e.g. embedding plots or using different font colors) later on.

.github/workflows/performance-benchmarks-bayesmark.yml
xadrianzetx: Fixed in 3db2f4fd83b44532018fa4adfd05eaff0f35a77d. There was also extra line between sections, so I removed that. I wonder if it would be worth introducing some kind of linter for those workflow files...

benchmarks/bayesmark/report_template.md
xadrianzetx: Done in c08bcacd67e3c3cb22338f34aa688f29e8676712.

benchmarks/bayesmark/optuna_optimizer.py
xadrianzetx: Added QMCSampler as well as samplers from integration modules in 440d42a49df6053b92c323bceb62dea12f0481ce.

himkt:
Thank you @xadrianzetx for the fix. I ran the job on my forked repo (https://github.com/himkt/optuna-test-bayesmark/actions/workflows/performance-benchmarks-bayesmark.yml) and artifacts looks good.

himkt (approved):

PR #3362, opened 2022-03-09
Fix speed benchmarks

not522:
## Motivation
This PR will fix the failure of speed benchmarks due to the change of `STORAGE_MODES` (#3204).

## Description of the changes
- Fix speed benchmarks
- Add `--strict` option of `asv` to fail the CI when some cases failing (https://github.com/airspeed-velocity/asv/pull/865)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -167,7 +167,7 @@ def get_extras_require() -> Dict[str, List[str]]:
             "fastai",
         ],
         "benchmark": [
-            "asv",
+            "asv>=0.5.0",
             "virtualenv",
             "botorch",
         ],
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -167,7 +167,7 @@ def get_extras_require() -> Dict[str, List[str]]:
             "fastai",
         ],
         "benchmark": [
-            "asv",
+            "asv>=0.5.0",
```
not522, line 170: Note: `--strict` was introduced in asv v0.5.0

HideakiImamura:

HideakiImamura (approved):
LGTM.

himkt:
Confirmed that the job failed on strict mode (https://github.com/himkt/optuna-test-asv/runs/5563416229?check_suite_focus=true) and worked with patch (https://github.com/himkt/optuna-test-asv/runs/5563528557?check_suite_focus=true).

himkt (approved):
Thank you for the fix, LGTM!

PR #3368, opened 2022-03-10
Improve Matplotlib parallel coordinate test

nzw0301:
## Motivation

A part of https://github.com/optuna/optuna/issues/3226

## Description of the changes

Improve test functions by following Plotly's test.

- Implement similar tests to Plotly's test
- Refactor `matplotlib.plot_parallel_coodinate` a little bit

## TODO

- [x] Add simple tests to `test_plot_parallel_coordinate_unique_hyper_param`.
- [x] Add test to `test_plot_parallel_coordinate_only_missing_params` after https://github.com/optuna/optuna/pull/3415 is merged onto the master.

### Followup

- Add test for line collection's data for each line.

---

This PR is based on https://github.com/optuna/optuna/pull/3284

PR #3369, opened 2022-03-10
Fix objective scale in parallel coordinate of Matplotlib

nzw0301:
## Motivation

The motivation can be categorised into `2. Unify appearance between Plotly and Matplotlib` in https://github.com/optuna/optuna/issues/2959. 

The colourbar of `optuna.visualize.matplotlib.plot_parallel_coordinate` always includes `0` by default. As a result, the generated colourbar is inconsistent with Plotly's counterpart when zero is not in the range of objective values. For example, 

```python
import optuna
def objective(trial):
    return trial.suggest_int("x", 10, 20)


study = optuna.create_study()
optuna.visualization.matplotlib.plot_parallel_coordinate(study)
```

Clearly, `objective` does not return zero. However, the generated colour bar always includes `0` as follows:

Screenshot 2022-03-10 at 22 35 00

Plotly's counterpart does not use this trick (at least in the latest stable). So the plot looks like

Screenshot 2022-03-10 at 22 34 37

## Description of the changes

Remove the hard-coded zero. The generated plots will be like

Screenshot 2022-03-10 at 22 34 34

### DIFF
--- a/optuna/visualization/matplotlib/_parallel_coordinate.py
+++ b/optuna/visualization/matplotlib/_parallel_coordinate.py
@@ -197,7 +197,7 @@ def _get_parallel_coordinate_plot(
     xs = [range(len(sorted_params) + 1) for _ in range(len(dims_obj_base))]
     segments = [np.column_stack([x, y]) for x, y in zip(xs, dims_obj_base)]
     lc = LineCollection(segments, cmap=cmap)
-    lc.set_array(np.asarray([target(t) for t in trials] + [0]))
+    lc.set_array(np.asarray([target(t) for t in trials]))
     axcb = fig.colorbar(lc, pad=0.1)
     axcb.set_label(target_name)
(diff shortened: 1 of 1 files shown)


nzw0301:
Note that the test case will be covered by https://github.com/optuna/optuna/pull/3368, specifically, https://github.com/optuna/optuna/pull/3368/files#diff-93d6e2dd3730bf35c1856229ee8f0c6b56dbd2fa3218fa4bc4178eea20fb955aR36.

himkt (approved):
LGTM for this change.

> Note that the test case will be covered by https://github.com/optuna/optuna/pull/3368

I'm wondering if it needs to add a test case whose all values are not 0... But after this PR, the test will not make any sense so I think it's okay to do nothing.

HideakiImamura (approved):
LGTM.

Issue #3336, opened 2022-02-25
Keyboard interruption handling


CorentinNeovision:
### Motivation

At the moment, if a study is launched and then manually interrupted (with control-C), then the current trial is killed without updating the study database. This leaves a trial with the status RUNNING in the database which will never be cleaned.

I implemented a simple solution which would change that behaviour in the case of sequential execution (it doesn't change anything in case of parallel optimization, if n_jobs > 1), by setting the Trial to the FAIL state in database. I want to know if you would like that I open a pull request to add this change.

### Description

I implemented a simple handling which allow to handle a `KeyboardInterrupt` in `optuna.study._optimize._run_trial`. As a consequence, now if a study is interrupted by a `KeyboardInterrupt`, you get an additional warning log in addition to the `KeyboardInterrupt` stacktrace, which would look like this:
`[W 2022-02-25 11:05:17,337] Trial 15 failed because of the following error: KeyboardInterrupt()`
And the database stores properly that the Trial status is FAIL

himkt:
Hi @CorentinNeovision, I'd like to be careful to introduce as it would needs some discussion.
(the motivation behind that is to keep exception handling slim as possible)

Instead, how about catching logic in your objective function?

```python
import time

import optuna


def objective(trial: optuna.trial.Trial) -> float:
    v = trial.suggest_float("v", 0, 10)
    try:
        time.sleep(1)
    except KeyboardInterrupt:
        trial.set_user_attr("reason", "keyboard interrupt")  # if you need
        return float("nan")

    return v


study = optuna.create_study()
study.optimize(objective, n_trials=10)
print(study.trials_dataframe())
```

### log

```sh
> python main.py                                                                                                                                       2022-02-26 00:48:56
[I 2022-02-26 00:48:58,439] A new study created in memory with name: no-name-ac8f8396-21bd-4a16-82dd-0538ca7c7499
/Users/himkt/work/github.com/himkt/optuna/optuna/trial/_trial.py:169: FutureWarning: UniformDistribution has been deprecated in v3.0.0. This feature will be removed in v6.0.0. See https://github.com/optuna/optuna/releases/tag/v3.0.0. Use :class:`~optuna.distributions.FloatDistribution` instead.
  distribution = UniformDistribution(low=low, high=high)
[I 2022-02-26 00:48:59,446] Trial 0 finished with value: 1.6357764656636398 and parameters: {'v': 1.6357764656636398}. Best is trial 0 with value: 1.6357764656636398.
[...]

PR #3374, opened 2022-03-10
Improve Ctrl-C interruption handling

CorentinNeovision:
In case of a keyboard interruption during a non parallel optimization,
will set the proper FAIL log and status for the interrupted trial.

## Motivation

As mentioned in #3336

## Description of the changes

Add an except for `KeyboardInterrupt` during optimization. Also a minor tweaks to a type hint for `func_err`, which is now not always an `Exception` as `KeyboardInterrupt` is only a `BaseException`.

### DIFF
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -193,7 +193,7 @@ def _run_trial(
 
     state: Optional[TrialState] = None
     values: Optional[List[float]] = None
-    func_err: Optional[Exception] = None
+    func_err: Optional[BaseException] = None
     func_err_fail_exc_info: Optional[Any] = None
     # Set to a string if `func` returns correctly but the return value violates assumptions.
     values_conversion_failure_message: Optional[str] = None
@@ -212,6 +212,9 @@ def _run_trial(
     except exceptions.TrialPruned as e:
         # TODO(mamu): Handle multi-objective cases.
         state = TrialState.PRUNED
+        func_err = e
+    except KeyboardInterrupt as e:
+        state = TrialState.FAIL
         func_err = e
     except Exception as e:
         state = TrialState.FAIL
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -193,7 +193,7 @@ def _run_trial(
 
     state: Optional[TrialState] = None
     values: Optional[List[float]] = None
-    func_err: Optional[BaseException] = None
+    func_err: Optional[Union[Exception, KeyboardInterrupt]] = None
     func_err_fail_exc_info: Optional[Any] = None
     # Set to a string if `func` returns correctly but the return value violates assumptions.
     values_conversion_failure_message: Optional[str] = None
@@ -213,10 +213,7 @@ def _run_trial(
         # TODO(mamu): Handle multi-objective cases.
         state = TrialState.PRUNED
         func_err = e
-    except KeyboardInterrupt as e:
-        state = TrialState.FAIL
-        func_err = e
-    except Exception as e:
+    except (Exception, KeyboardInterrupt) as e:
         state = TrialState.FAIL
         func_err = e
         func_err_fail_exc_info = sys.exc_info()
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -307,6 +307,9 @@ class Study:
         default choice for the sampler is TPE.
         See also :class:`~optuna.samplers.TPESampler` for more details on 'TPE'.
 
+        An optimization will be stopped when receiving a termination signal such as SIGINT and SIGTERM.
+        Unlike other signals, trial is automatically failed when receiving SIGINT (ctrl+c).
+
         Example:
 
             .. testcode::
@@ -329,7 +332,7 @@ class Study:
                 The number of trials for each process. If this argument is set to :obj:`None`,
                 there is no limitation on the number of trials. If ``timeout`` is also set to
                 :obj:`None`, the study continues to create trials until it receives a termination
-                signal such as Ctrl+C or SIGTERM, in which case it also kills the current trial.
+                signal such as Ctrl+C or SIGTERM.
 
                 .. seealso::
                     :class:`optuna.study.MaxTrialsCallback` can ensure how many times trials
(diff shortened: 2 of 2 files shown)


optuna/study/_optimize.py
```diff
@@ -193,7 +193,7 @@ def _run_trial(
 
     state: Optional[TrialState] = None
     values: Optional[List[float]] = None
-    func_err: Optional[Exception] = None
+    func_err: Optional[BaseException] = None
```
himkt, line 196: I personally prefer `Union[Exception,KeyboardInterrupt]` over `BaseException`, because `BaseException` is little bit too general.

himkt (commented):
> I think the documentation could be improved, I tried to add a little more info but I couldn't find how to detail the status of interrupted Trials without being overly specific. If there is somewhere better in the doc than the n_trials and timeout doc to explain what kind of interruption leads to what status for the Trial, for sequential or parallel optimization, I'd be happy to do it.

Thank you for the consideration!
It's the just one idea but how about mentioning it in the description section?

```python
    def optimize(
        self,
        ...
        show_progress_bar: bool = False,
    ) -> None:
        """Optimize an objective function.

        Optimization is done by choosing a suitable set of hyperparameter values from a given
        ...
        See also :class:`~optuna.samplers.TPESampler` for more details on 'TPE'.
        +
        + An optimization will be stopped when receiving a termination signal such as SIGINT and SIGTERM.
        + Unlike other signals, trial is automatically failed when receiving SIGINT (ctrl+c).

        Example:
```

himkt:
I'd like @sile to be reviewer as well. If you are not available, please tell me and remove the assignment.

sile:
👍 
I think that I can review this PR within a week.

optuna/study/_optimize.py
```diff
@@ -213,6 +213,9 @@ def _run_trial(
         # TODO(mamu): Handle multi-objective cases.
         state = TrialState.PRUNED
         func_err = e
+    except KeyboardInterrupt as e:
+        state = TrialState.FAIL
+        func_err = e
```
sile, line 218: I'm not sure that we should do to that but it might be better to set the result of `sys.exc_info()` to `func_err_fail exc_info` here as the following `Exception` handling branch does.
What do you think of that? > @CorentinNeovision @himkt

optuna/study/study.py
```diff
@@ -329,7 +329,7 @@ def objective(trial):
                 The number of trials for each process. If this argument is set to :obj:`None`,
                 there is no limitation on the number of trials. If ``timeout`` is also set to
                 :obj:`None`, the study continues to create trials until it receives a termination
-                signal such as Ctrl+C or SIGTERM.
+                signal such as Ctrl+C or SIGTERM, in which case it also kills the current trial.
```
sile, line 332: The current phrase "it also kills the current trial" sounds a bit ambigious to me (some users would not understand what the word "kill" means exactly). 
How about rewriting this sentence or adding a phrase such as "i.e., set the state of the trail to TrialState.FAILED"?

sile: Besides, I think that this PR can handle Ctrl+C (SIGINT) but cannot SIGTERM. So, the current doc seems a bit misleading in this aspect. This is just an idea, but we can replace "in which case" with "if the signal is SIGINT (Ctrl+C)" to make the condition clear.

optuna/study/_optimize.py
himkt: Thank you for the catch. It makes sense. `except` can takes multiple exceptions so we can update here from

```python
    except KeyboardInterrupt as e:
        state = TrialState.FAIL
        func_err = e
    except Exception as e:
        state = TrialState.FAIL
        func_err = e
        func_err_fail_exc_info = sys.exc_info()
```

to

```python
    except (Exception, KeyboardInterrupt) as e:
        state = TrialState.FAIL
        func_err = e
        func_err_fail_exc_info = sys.exc_info()
```

:link: https://docs.python.org/3/tutorial/errors.html#handling-exceptions

optuna/study/study.py
CorentinNeovision: SIGTERM could probably also be handled, but it would require more work with signal handling, and furthermore probably wouldn't be compatible with windows. So that was not in the scope of my PR

sile:
> maybe it should also be specified that this only works when n_jobs=1 ?

+1

I agree that supportingparallel optimization is too hard to address within this PR. So adding a note document sounds sufficient and nice.

himkt:
> it doesn't change anything in case of parallel optimization, if n_jobs > 1

I thought the change will only affect a single process execution as you said in the original issue. It looks solid to me, although it would be better to document the behavior.

himkt:
Wow, sorry for the duplicated comment. I agree with @sile.

optuna/study/study.py
```diff
@@ -307,6 +307,9 @@ def optimize(
         default choice for the sampler is TPE.
         See also :class:`~optuna.samplers.TPESampler` for more details on 'TPE'.
 
+        An optimization will be stopped when receiving a termination signal such as SIGINT and SIGTERM.
```
himkt, line 310: https://github.com/optuna/optuna/runs/5562609091?check_suite_focus=true
This line is too long. Please fix and check if CI pass. :pray:

sile (approved):
LGTM 👍

sile:
To merge this PR, please wait for another review by @himkt

optuna/study/study.py
```diff
@@ -307,6 +307,11 @@ def optimize(
         default choice for the sampler is TPE.
         See also :class:`~optuna.samplers.TPESampler` for more details on 'TPE'.
 
+        An optimization will be stopped when receiving a termination signal such as SIGINT and
+        SIGTERM. Unlike other signals, trial is automatically and cleanly failed when receiving
```
himkt, line 311:
```suggestion
        Optimization will be stopped when receiving a termination signal such as SIGINT and
        SIGTERM. Unlike other signals, a trial is automatically and cleanly failed when receiving
```

himkt (commented):
@CorentinNeovision I'm sorry to miss your commits after the review and thank you @sile for your kindly ping. Your PR looks good except for minor suggestions. After my suggestions are applied (or not applied with explanations), I'll approve and merge.

optuna/study/study.py
CorentinNeovision: I'm ok with your suggestions

himkt (commented):
LGTM! Thank you @CorentinNeovision for your contribution!

PR #3378, opened 2022-03-12
Cleanup `MLflowCallback` tests

xadrianzetx:
## Motivation

A bit overdue cleanup of `MLflowCallback` tests as discussed in https://github.com/optuna/optuna/pull/2912#issuecomment-915462924 and https://github.com/optuna/optuna/pull/2912#discussion_r720187136.

## Description of the changes

* Reduced test duplication
* Fixed `tracking_uri` naming
* Refactored `test_log_*` to use only public APIs
* simplified `test_study_name`

### DIFF
--- a/tests/integration_tests/test_mlflow.py
+++ b/tests/integration_tests/test_mlflow.py
@@ -310,6 +310,24 @@ def test_mlflow_callback_fails_when_nest_trials_is_false_and_active_run_exists(
             study.optimize(_objective_func, n_trials=1, callbacks=[mlflc])
 
 
+def test_tag_always_logged(tmpdir: py.path.local) -> None:
+
+    tracking_uri = f"file:{tmpdir}"
+    study_name = "my_study"
+    n_trials = 3
+
+    mlflc = MLflowCallback(tracking_uri=tracking_uri)
+    study = optuna.create_study(study_name=study_name)
+    study.optimize(_objective_func, n_trials=n_trials, callbacks=[mlflc])
+
+    mlfl_client = MlflowClient(tracking_uri)
+    experiment = mlfl_client.list_experiments()[0]
+    runs = mlfl_client.search_runs([experiment.experiment_id])
+
+    assert all((r.data.tags["direction"] == "MINIMIZE") for r in runs)
+    assert all((r.data.tags["state"] == "COMPLETE") for r in runs)
+
+
 @pytest.mark.parametrize("tag_study_user_attrs", [True, False])
 def test_tag_study_user_attrs(tmpdir: py.path.local, tag_study_user_attrs: bool) -> None:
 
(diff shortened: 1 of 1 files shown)


himkt:
Thank you for the PR!

@keisuke-umezawa @nzw0301 Could you please take a look?
Also please comment @toshihikoyanase as the former reviewer of related PRs.

nzw0301:
Due to my main job, I'll be busy this week, so it might take time to review this PR 🙇🏻

tests/integration_tests/test_mlflow.py
```diff
@@ -52,47 +56,22 @@ def _objective_func_long_user_attr(trial: optuna.trial.Trial) -> float:
...
-    )
-    assert (
-        first_run_dict["data"]["tags"]["y_distribution"]
-        == "LogUniformDistribution(high=30.0, low=20.0)"
-    )
-    assert (
-        first_run_dict["data"]["tags"]["z_distribution"]
-        == "CategoricalDistribution(choices=(-1.0, 1.0))"
-    )
-    assert first_run_dict["data"]["tags"]["my_user_attr"] == "my_user_attr_value"
```
HideakiImamura, line 95: Could you explain why you removed these lines?

tests/integration_tests/test_mlflow.py
```diff
@@ -425,40 +423,28 @@ def _objective_func(trial: optuna.trial.Trial) -> float:
...
-    study = optuna.create_study(study_name=study_name)
-
-    mlflc._initialize_experiment(study)
-
-    mlfl_client = MlflowClient(tracking_file_name)
-    experiments = mlfl_client.list_experiments()
-    assert len(experiments) == 1
-
-    experiment = experiments[0]
-    assert experiment.name == study_name
```
HideakiImamura, line 443: Ditto.

HideakiImamura (commented):
Thanks for the PR. I have a naive question.

tests/integration_tests/test_mlflow.py
xadrianzetx: `test_study_name` covers this already, so I removed this duplication.

nzw0301: I think this test function aims to verify `mlflc._initialize_experiment(study)`; if we remove `mlflc._initialize_experiment` in this test, we need to rename the test function name or remove this test. So we might need the original test case. 

I suppose `test_log_metric`, `test_log_metric_none`, and `test_log_params` are also same; they aim to verify a primate method of the callaback.

xadrianzetx: I think the same can be archived by testing only using public APIs, regardless of callback implementation details. We test for specific callback behavior (is experiment created with correct name?) which in this case is covered by another test. The test will stay the same even if `mlflc._initialize_experiment` is removed. Same goes for `log_*`

tests/integration_tests/test_mlflow.py
HideakiImamura: Thanks for the explanation. I found `test_tag_trial_user_attrs` tests the user attrs. Looks good.

tests/integration_tests/test_mlflow.py
HideakiImamura: Sounds reasonable. Let me resolve this conversation.

HideakiImamura (approved):
Thanks for the explanations! LGTM!

nzw0301 (approved):
LGTM. Thanks!

PR #3381, opened 2022-03-14
Lazy import of alembic, sqlalchemy, and scipy

himkt:
Heavily inspired by #3000.

### Description

This PR makes alembic modules be imported on demand.
It drops the dependency on alembic in `import optuna`.
As a bonus, the speed of `import optuna` will be improved a little (3.3x faster):

Before | After
-- | --
0.584s (100%) | 0.175s (29%)
Screen Shot 2022-03-15 at 21 13 56 | Screen Shot 2022-03-15 at 21 14 01

### Other packages

As other dominant parts, there are some libraries: `numpy`, `redis`, and `tqdm`. `numpy` appears manywhere in our codebase (to completely make lazy, it needs to change many files: #3383). `redis` is already optional, and it is imported using `try_import()`. `try_import()` tries to import a module and suppress to raise an exception if it fails, which means it's impossible to make it lazy. I didn't make `tqdm` lazy because it is not so heavy.


himkt:
@hvy Could you please take a look at this PR? I assign you because you are the author importing module. If you're busy, please feel free to unassign.

optuna/storages/_rdb/storage.py
```diff
@@ -14,34 +14,49 @@
...
+    alembic_command = _LazyImport("alembic.command")
+    alembic_config = _LazyImport("alembic.config")
+    alembic_migration = _LazyImport("alembic.migration")
+    alembic_script = _LazyImport("alembic.script")
+
+    sqlalchemy = _LazyImport("sqlalchemy")
+    sqlalchemy_engine = _LazyImport("sqlalchemy.engine")
+    sqlalchemy_exc = _LazyImport("sqlalchemy.exc")
+    sqlalchemy_orm = _LazyImport("sqlalchemy.orm")
+    sqlalchemy_sql_functions = _LazyImport("sqlalchemy.sql.functions")
```
hvy, line 56: Can we omit some these imports to reduce the LOC and improve readability, i.e. just import the root `sqlalchemy` and access those as submodules at usages.

optuna/storages/_rdb/storage.py
```diff
@@ -14,34 +14,49 @@
...
+
+    optuna_rdb_models = _LazyImport("optuna.storages._rdb.models")
```
hvy, line 58: Not a strong opinion but how about both keeping the diff smaller and the code less verbose? I believe there are dittos elsewhere.
```suggestion
    models = _LazyImport("optuna.storages._rdb.models")
```

optuna/storages/_rdb/storage.py
```diff
@@ -14,34 +14,49 @@
...
+    import alembic.config as alembic_config
+    import alembic.migration as alembic_migration
+    import alembic.script as alembic_script
+    import sqlalchemy
+    import sqlalchemy.engine as sqlalchemy_engine
+    import sqlalchemy.exc as sqlalchemy_exc
+    import sqlalchemy.orm as sqlalchemy_orm
+    import sqlalchemy.sql.functions as sqlalchemy_sql_functions
+
+    from optuna.storages._rdb import models as optuna_rdb_models
```
hvy, line 45: Sorry, why was it that this conditioning was needed in the first place?

hvy (commented):
Thanks for the PR. Left some early comments in if you could take a look. I think we can reduce the code a bit.

optuna/storages/_rdb/storage.py
himkt: Nice question. `models.py` also imports sqlalchemy modules/classes (https://github.com/optuna/optuna/blob/a07a36e124d6523677d718819cad61628e8621e7/optuna/storages/_rdb/models.py#L5-L19).

Instead of making them lazy, I make the module itself lazy to avoid introducing `_LazyImport` to `models.py` as well. If you feel it unreasonable, I'm fine to stop this conditioning and make sqlalchemy imports in `models.py` lazy.

optuna/storages/_rdb/storage.py
himkt: `sqlalchemy.{exc,orm,sql.function}` are not accessible from `sqlalchemy`.

optuna/storages/_rdb/storage.py
hvy: Ah, my question was rather, why do we need to condition based on `TYPE_CHECKING` and why can we not always do a lazy import. I'm probably just missing something related to mypy?

himkt: As you expected, it's for static checker. Without additional imports, all modules are treated as `Any`. It could hurts the quality of Optuna codebase. Also, with `TYPE_CHECKING`, we are backed by IDE supports (e.g. pylance).

w/o TYPE_CHECKING | w/ TYPE_CHECKING
-- | --
Screen Shot 2022-04-24 at 19 50 11 | Screen Shot 2022-04-24 at 19 50 40

himkt:
I revised the PR. PTAL @hvy @not522. :pray:

hvy (approved):
LGTM!

not522 (approved):
LGTM

PR #3177, opened 2021-12-19
[WIP] Unify plot pareto front

semiexp:
NOTE: #3128 should be merged before this PR. Once #3128 is merged, I will rebase this feature branch upon `master`

## Motivation
It is desirable to support `constraints_func` in `matplotlib.plot_pareto_front` as in #3128. To avoid repetition of pareto-front computation codes, we would like to factor out the preprocessing in `plot_pareto_front`s for pyplot and matplotlib.

## Description of the changes
Preprocessing of `plot_pareto_front` (for pyplot) is factored out so that it can be used by `matplotlib.plot_pareto_front` as well. At the same time, both 2D and 3D plots are done in a single function as in #3128.

PR #3389, opened 2022-03-17
Unify pareto front

semiexp:
## Motivation
We have two separate implementations of `plot_pareto_front`s for pyplot and matplotlib. Here we have 2 issues:

- They should essentially do the same things, except how the plots are rendered using respective libraries. Therefore, we have redundant codes in two versions of `plot_pareto_front`.
- They are slightly different in supported features (`constraints_func` is supported only by pyplot, while `targets` is supported only by matplotlib).

## Description of the changes
"Backend" (logic for obtaining the necessary information to plot) is now factored out as `_get_pareto_front_info`.

At present, signatures of `plot_pareto_front`s as well as their supported features are unchanged. I'm not sure whether we should enhance features (`targets` to pyplot version, `constraints_func` to matplotlib version) in this PR or another PR.

### DIFF
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -229,10 +229,8 @@ def _get_pareto_front_info(
 
     _targets = targets
     if _targets is None:
-        if len(study.directions) == 2:
-            _targets = _targets_default_2d
-        elif len(study.directions) == 3:
-            _targets = _targets_default_3d
+        if len(study.directions) in (2, 3):
+            _targets = _targets_default
         else:
             raise ValueError(
                 "`plot_pareto_front` function only supports 2 or 3 objective"
@@ -315,12 +313,8 @@ def _get_pareto_front_info(
     )
 
 
-def _targets_default_2d(trial: FrozenTrial) -> Sequence[float]:
-    return trial.values[0], trial.values[1]
-
-
-def _targets_default_3d(trial: FrozenTrial) -> Sequence[float]:
-    return trial.values[0], trial.values[1], trial.values[2]
+def _targets_default(trial: FrozenTrial) -> Sequence[float]:
+    return trial.values
 
 
 def _get_non_pareto_front_trials(
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -29,7 +29,7 @@ _logger = optuna.logging.get_logger(__name__)
 class _ParetoFrontInfo(NamedTuple):
     n_targets: int
     target_names: Sequence[str]
-    best_trials_with_values: Sequence[Tuple[FrozenTrial, Sequence[float]]]
+    best_trials_with_values: Optional[Sequence[Tuple[FrozenTrial, Sequence[float]]]]
     non_best_trials_with_values: Optional[Sequence[Tuple[FrozenTrial, Sequence[float]]]]
     infeasible_trials_with_values: Optional[Sequence[Tuple[FrozenTrial, Sequence[float]]]]
     axis_order: Sequence[int]
@@ -308,7 +308,7 @@ def _get_pareto_front_info(
     return _ParetoFrontInfo(
         n_targets=n_targets,
         target_names=target_names,
-        best_trials_with_values=best_trials_with_values,  # type: ignore
+        best_trials_with_values=best_trials_with_values,
         non_best_trials_with_values=non_best_trials_with_values,
         infeasible_trials_with_values=infeasible_trials_with_values,
         axis_order=axis_order,
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -1,3 +1,4 @@
+import collections
 import json
 from typing import Any
 from typing import Callable
@@ -6,7 +7,9 @@ from typing import List
 from typing import NamedTuple
 from typing import Optional
 from typing import Sequence
+from typing import Tuple
 from typing import Union
+import warnings
 
 import optuna
 from optuna._experimental import experimental
@@ -24,11 +27,11 @@ _logger = optuna.logging.get_logger(__name__)
 
 
 class _ParetoFrontInfo(NamedTuple):
-    n_dim: int
+    n_targets: int
     target_names: List[str]
-    best_trials: List[FrozenTrial]
-    non_best_trials: Optional[List[FrozenTrial]]
-    infeasible_trials: Optional[List[FrozenTrial]]
+    best_trials_with_values: List[Tuple[FrozenTrial, Sequence[float]]]
+    non_best_trials_with_values: Optional[Sequence[Tuple[FrozenTrial, Sequence[float]]]]
+    infeasible_trials_with_values: Optional[Sequence[Tuple[FrozenTrial, Sequence[float]]]]
     axis_order: List[int]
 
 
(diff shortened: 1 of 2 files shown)


himkt:
@HideakiImamura Please take a look at the PR continue from #3177.

optuna/visualization/_pareto_front.py
```diff
@@ -171,53 +130,200 @@ def _make_scatter_object(
...
+                    " your `targets` returns {}".format(type(trials_with_values[0][1]))
+                )
+            return len(trials_with_values[0][1])
+        return None
+
+    n_targets = (
+        _infer_n_targets(best_trials_with_values)
+        or _infer_n_targets(non_best_trials_with_values)
+        or _infer_n_targets(infeasible_trials_with_values)
+    )
```
himkt, line 272: [question]

Do we have to invoke `_infer_n_targets` for `non_best_trials_with_values` here? I expect `_infer_n_targets` is invoked for each type of trials because `best_trials_with_values` would be `None` when no feasible trial could be available. 

But, I think `non_best_trials_with_values` would be also empty when `best_trials_with_values` is empty, and so it'd be enough to check `best_trials_with_values` and `infeasible_trials_with_values`. Please correct me if I don't understand your intention.

himkt, line 272: [question]

Do we have to invoke `_infer_n_targets` for `non_best_trials_with_values` here? I expect `_infer_n_targets` is invoked for `{best,non_best,_infeasible}_trials` because `best_trials_with_values` would be `None` when no feasible trial could be available. In such case, I think non_best_trials would be also empty so it'd be enough to check `best_trials_with_values` and `infeasible_trials_with_values`. Please correct me if I don't understand your intention.

optuna/visualization/_pareto_front.py
```diff
@@ -171,53 +130,200 @@ def _make_scatter_object(
...
+    constraints_func: Optional[Callable[[FrozenTrial], Sequence[float]]] = None,
+    targets: Optional[Callable[[FrozenTrial], Sequence[float]]] = None,
+) -> _ParetoFrontInfo:
+    if axis_order is not None:
+        warnings.warn(
+            "`axis_order` has been deprecated in v3.0.0. "
+            "This feature will be removed in v5.0.0. "
+            "See https://github.com/optuna/optuna/releases/tag/v3.0.0.",
+            DeprecationWarning,
+        )
```
himkt, line 194: [memo] After this PR is merged, `axis_order` is also deprecated in plotly backend.

himkt, line 194: [memo] After this PR is merged, `axis_order` is also deprecated in plotly backend.

optuna/visualization/_pareto_front.py
```diff
@@ -22,6 +26,15 @@
 _logger = optuna.logging.get_logger(__name__)
 
 
+class _ParetoFrontInfo(NamedTuple):
+    n_targets: int
+    target_names: List[str]
```
himkt, line 31: [question]

I found `List` and `Sequence` both appear in this file. Do you have any policy to distinguish uses of them?

himkt, line 31: [question]

`List` and `Sequence` both appear in this file. Do you have any policy to distinguish uses of them?

optuna/visualization/_pareto_front.py
```diff
@@ -171,53 +130,200 @@ def _make_scatter_object(
...
+        for trial in study.get_trials(states=(TrialState.COMPLETE,)):
+            if all(map(lambda x: x <= 0.0, constraints_func(trial))):
+                feasible_trials.append(trial)
+            else:
+                infeasible_trials.append(trial)
+        best_trials = _get_pareto_front_trials_by_trials(feasible_trials, study.directions)
+        if include_dominated_trials:
+            non_best_trials = _get_non_pareto_front_trials(feasible_trials, best_trials)
+        else:
+            non_best_trials = []
```
himkt, line 214:
```python
import optuna


def objective(trial):
    v0 = 4 * x ** 2 + 4 * y ** 2
    if x > 3:
        trial.set_user_attr("constraint", (1,))
    else:
        trial.set_user_attr("constraint", (0,))
    return v0, v1


def constraints(trial):


study = optuna.create_study(
    sampler=optuna.samplers.NSGAIISampler(constraints_func=constraints),
)
study.optimize(objective, n_trials=100, n_jobs=1)

optuna.visualization.plot_pareto_front(
    study,
    constraints_func=constraints,
    include_dominated_trials=False,
)
```

I ran the optimization and visualized the pareto front plot to get the following figure:

Screen Shot 2022-03-23 at 14 48 45

It looks like dominated trials are included in the figure.

I want ask you whether this is expected. I don't understand why they are plotted even `non_best_trials` is constructed from `feasible_trials`.

himkt (commented):
Thank you @semiexp for the awesome cleanup! I took your PR and left comments/questions. :memo:

optuna/visualization/matplotlib/_pareto_front.py
```diff
@@ -100,221 +99,84 @@ def objective(trial):
...
-    for trial in study.get_trials():
-        if trial.state == TrialState.COMPLETE and trial not in pareto_trials:
-            non_pareto_trials.append(trial)
-    return non_pareto_trials
-
-
-def _get_trials(study: Study, include_dominated_trials: bool) -> List[FrozenTrial]:
-    trials = study.best_trials
-    if len(trials) == 0:
-        _logger.warning("Your study does not have any completed trials.")
```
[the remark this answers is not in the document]
semiexp: I agree. Deleted `_logger` from matplotlib `_pareto_front.py`

optuna/visualization/_pareto_front.py
semiexp: I was unaware that we can skip check for `non_best_trials_with_values`. I think just removing the check may be a little confusing, so I will leave a short comment here.

optuna/visualization/_pareto_front.py
semiexp: I just had a vague policy that I use `List` if the container is created in this code and `Sequence` if the container is given by users. But now this distinction seems unnecessary. I will change `List` to `Sequence`.

optuna/visualization/_pareto_front.py
semiexp: Do you mean grey trials by "dominated trials"? In the current implementation, `infeasible_trials` are first excluded from all trials, then remaining ones are classified as `best` and `non_best`.
There will be some approaches for classifying trials:

- Exclude infeasible ones first, then classify feasible ones into best and non-best ones. (current approach)
- Exclude infeasible ones first, then classify feasible ones into best and non-best ones. Also, if `include_dominated_trials=False`, remove "dominated" infeasible trials.
- Exclude non-best ones first, then classify best ones by their feasibility. I think this is not a good idea, because feasible trials may be excluded just because they are dominated only by infeasible ones.

I understand that choice of the approach can be controversial.
Anyway, I think this PR itself should not change the behavior because this is for refactoring.

optuna/visualization/_pareto_front.py
```diff
@@ -22,6 +26,15 @@
 _logger = optuna.logging.get_logger(__name__)
 
 
+class _ParetoFrontInfo(NamedTuple):
+    n_targets: int
+    target_names: Sequence[str]
+    best_trials_with_values: Sequence[Tuple[FrozenTrial, Sequence[float]]]
```
HideakiImamura, line 32: The `best_trials_with_values` can be `None`. How about adding `Optional` to this? Then, we can remove `# type: ignore` from line 311.

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM. I have a minor comment. PTAL.

optuna/visualization/_pareto_front.py
himkt: Thank you for taking the time to explain, it totally makes sense. I assumed the third approach but I didn't care `feasible trials may be excluded just because they are dominated only by infeasible ones`.

I also agree with you that this PR should focus on the refactoring.

optuna/visualization/_pareto_front.py
himkt: Thank you.

> Some distinctions have been existing before this change

It's okay to leave them. After this visualization refactoring was done, we'll another PR to cleanup the signatures of utilities in visualization modules.

himkt (approved):
Thank you so much for greatly improving the readability+consistency. It looks good except https://github.com/optuna/optuna/pull/3389#discussion_r838338432.

himkt:
Thank you for the update. Additional changes also looks good.

HideakiImamura (approved):
Thanks for the update. LGTM!

PR #3391, opened 2022-03-20
Make `set_trial_param()` of `RedisStorage` faster

masap:
## Description of the changes
By using `mset()` instead of `pipeline()`, this patch improve performance of DB access related process in `set_trial_param()` from 228(usec) to 161(usec).

chart

### DIFF
--- a/optuna/storages/_redis.py
+++ b/optuna/storages/_redis.py
@@ -1,3 +1,4 @@
+from collections.abc import Mapping
 import copy
 from datetime import datetime
 import pickle
@@ -10,6 +11,7 @@ from typing import List
 from typing import Optional
 from typing import Sequence
 from typing import Set
+from typing import Union
 
 import optuna
 from optuna import distributions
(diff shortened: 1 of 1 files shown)


masap:
This PR conflicts with #3392 and #3393 each other. So please merge this PR first.

optuna/storages/_redis.py
```diff
@@ -463,19 +465,19 @@ def set_trial_param(
...
         trial = self.get_trial(trial_id)
 
-        with self._redis.pipeline() as pipe:
-            pipe.multi()
-            # Set study param distribution.
-            param_distribution[param_name] = distribution
-            pipe.set(
-                self._key_study_param_distribution(study_id), pickle.dumps(param_distribution)
-            )
+        queries: Mapping[Union[str, bytes], Union[bytes, float, int, str]]
```
toshihikoyanase, line 468: I have questions about type hinting.

1. Can the type of the keys be `str`? Both `_key_study_param_distribution` and `_key_trial` seems to return `str`.
2. Can the type of the values be `bytes`? The [pickle reference](https://docs.python.org/3/library/pickle.html#pickle.dumps) says that `pickle.dumps` returns a bytes object.

toshihikoyanase (commented):
Thank you for your PR! The use of the [mset](https://redis.io/commands/MSET) method seems reasonable to me.

I have a question about type hinting, so please take a look.

toshihikoyanase:

optuna/storages/_redis.py
masap: Thank you four your review! Indeed we only dealing with the types, but when we define like `queries: Mapping[str, bytes]`, `mypy` complains to it...

```
optuna/storages/_redis.py:480: error: Argument 1 to "mset" of "BasicKeyCommands" has incompatible type "Dict[str, bytes]"; expected "Mapping[Union[str, bytes], Union[bytes, float, int, str]]"
Found 1 error in 1 file (checked 275 source files)
```

himkt (approved):
Thank you for the PR! LGTM for the change.

I also want to avoid allowing bytes unnecessarily (https://github.com/optuna/optuna/pull/3391/files#r832737814), but it drives from the redis-py's specification and I think we don't have anything to do.

himkt:
If @toshihikoyanase is also fine for the change, I'll take a look at subsequent PRs.

optuna/storages/_redis.py
masap: @toshihikoyanase Is there any comment ?

masap:
@himkt Thank you !

himkt:
Let me kindly ping @toshihikoyanase. :pray:

toshihikoyanase (approved):
LGTM. I'm sorry for the delayed review, again.

optuna/storages/_redis.py
masap: Thanks !

masap:
Thanks !

PR #3394, opened 2022-03-20
Remove unused function from `RedisStorage`

masap:
## Description of the changes
This patch removes unused function from `RedisStorage`.

### DIFF
--- a/optuna/storages/_redis.py
+++ b/optuna/storages/_redis.py
@@ -629,14 +629,6 @@ class RedisStorage(BaseStorage):
 
         self._redis.set(self._key_trial(trial_id), pickle.dumps(trial))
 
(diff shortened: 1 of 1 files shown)


toshihikoyanase (approved):
Good catch! I confirmed that `RedisStorage._del_trial` is not used in the code base and tests. Thank you!

toshihikoyanase:
@keisuke-umezawa Could you review this PR, please? I don't think this PR breaks compatibility because it only removes the unused private method, but please feel free to add comments if you have any concerns.

nzw0301 (approved):
I think this was a follow-up task of https://github.com/optuna/optuna/pull/974#pullrequestreview-380914398. Nice catch!

masap:
Thanks !

Issue #3382, opened 2022-03-14
Handle with the outliner in visualization module


nzw0301:
### Motivation

Sub-issue of #2959

Some visualisation methods cannot deal with a trial including inf/nan as an objective or intermediate value. For `inf` value, we would like to remove such trials before showing the plot. In addition, we would like to print warning messages like `trial 0 is skipped because its objective value is inf`.

### Description

I think we can address this issue by the following steps:

## 1. Implement a filter utility function to remove trails whose objective values are `inf` in https://github.com/optuna/optuna/blob/master/optuna/visualization/_utils.py.

Resolved by https://github.com/optuna/optuna/pull/3395

### `optuna.visualization.plot_contour`

<img width="981" alt="Screenshot 2022-03-14 at 20 25 54" src="https://user-images.githubusercontent.com/7121753/158162815-c3a7184d-a2bd-4cce-997a-ff4779013343.png">

Fixed by https://github.com/optuna/optuna/pull/3451.

~Current: If the objective value is `inf`, the plot looks strange.~
~Expected: The trial should be removed from the plot.~

### `optuna.visualization.plot_edf`

Fixed by https://github.com/optuna/optuna/pull/3435.

~Current: No lines~
~Expected: edf plot should be shown. A trial whose objective is `inf` should be removed.~

### `optuna.visualization.plot_parallel_coordinate`

<img width="965" alt="Screenshot 2022-03-14 at 20 29 48" src="https://user-images.githubusercontent.com/7121753/158163336-2f9df5ee-2c7a-48f5-8ead-fa279a575610.png">

Current: Objective value's axis looks strange.
Expected: All lines should be within the plot.

### `optuna.visualization.plot_param_importances`

Fixed by https://github.com/optuna/optuna/pull/3500.

~Current: No bars~
~Expected: Importance plot should be shown.~

~Note that if this problem comes from the algorithm to compute importance score, it should be also fixed.~

### `optuna.visualization.matplotlib.plot_contour`

Fixed by https://github.com/optuna/optuna/pull/3451.

<img width="484" alt="Screenshot 2022-03-14 at 20 33 42" src="https://user-images.githubusercontent.com/7121753/158163820-67910037-4ed8-460b-84f4-0b622da9114b.png">

Current: No contour lines
Expected: contour lines should be shown correctly.

### `optuna.visualization.matplotlib.plot_edf`

Fixed by https://github.com/optuna/optuna/pull/3435.

~Current: No lines~
[...]

PR #3395, opened 2022-03-21
Add utility to filter trials with `inf` values from visualizations

xadrianzetx:
## Motivation

A neat utility function which should help us skip trials with `inf` values/intermediate values when making visualizations. Part of #3382.

## Description of the changes

* Implemented `_filter_inf_trials`
* Test new function


toshihikoyanase:
@nzw0301 Could you review this PR as the author of #3382?

xadrianzetx:
@nzw0301, @toshihikoyanase I found pretty stupid bug in my test code and fixed it in c85c662c4931f80d8ade0730c24ed4c757d520b0. Sorry if you already started reviews :facepalm:.

nzw0301:
Sure

nzw0301:
I'll take a look at this PR this week...

optuna/visualization/_utils.py
```diff
@@ -99,3 +103,26 @@ def _get_param_values(trials: List[FrozenTrial], p_name: str) -> List[Any]:
...
+            _logger.info(
+                f"Trial {trial.number} is omitted in visualization ",
+                "because its objective value is inf.",
+            )
+
+        elif any(np.isinf(intermediate_values)):
+            _logger.info(
+                f"Trial {trial.number} is omitted in visualization ",
+                "because one or more intermediate values are inf.",
+            )
```
nzw0301, line 123: In my understanding, `intermediate_values` is used only by `plot_intermediate_values` methods in `optuna.visualization.*.py` and current plot methods can deal inf/nan correctly. So we might not need this skip part for intermediate values. I confirm that by running the following code:

```python
import numpy as np
import optuna

# nan case
def objective(trial):

    for step in range(10):
        if step == trial.number:
            trial.report(np.nan, step)
        else:
            trial.report(trial.number, step)
    return 1.

study = optuna.create_study()
optuna.visualization.plot_intermediate_values(study)
optuna.visualization.matplotlib.plot_intermediate_values(study)

# inf case
def objective(trial):

    for step in range(10):
        if step == trial.number:
            trial.report(np.inf, step)
        else:
            trial.report(trial.number, step)
    return 1.

study = optuna.create_study()
study = optuna.create_study()
optuna.visualization.plot_intermediate_values(study)
optuna.visualization.matplotlib.plot_intermediate_values(study)
```

Definitely, my poor issue description asked you to implement this. Sorry for that...

tests/visualization_tests/test_utils.py
```diff
@@ -59,3 +62,57 @@ def test_check_plot_args() -> None:
 
     with pytest.warns(UserWarning):
         _check_plot_args(study, lambda t: cast(float, t.value), "Objective Value")
+
+
+@pytest.mark.parametrize(
+    "value, expected", [(float("inf"), 1), (-float("inf"), 1), (0.0, 2), (float("nan"), 2)]
```
nzw0301, line 68: Should the filter be applied to this case `(float("nan"), 2)`?

optuna/visualization/_utils.py
xadrianzetx: No worries, I missed how plotly/matplotlib backends handle this case as well. Double checked using the test case you provided and indeed both backends render plots with missing datapoints where inf/nan were, which i guess is as valid as removing such trials altogether. Removed unnecessary code path and test in fbef410d801df8281f072d8c200b624161df2c50.

tests/visualization_tests/test_utils.py
xadrianzetx: You're right, I'll add support for that case later today.

xadrianzetx: Done in 8951790097091d67f0a1abc5bcd1ec7236c4f837.

nzw0301 (approved):
Thank you for your updates! I didn't know `np.isfinite`! LGTM!

optuna/visualization/_utils.py
```diff
@@ -99,3 +103,19 @@ def _get_param_values(trials: List[FrozenTrial], p_name: str) -> List[Any]:
...
         return values
     return list(map(str, values))
+
+
+def _filter_nonfinite(trials: List[FrozenTrial]) -> List[FrozenTrial]:
+
+    filtered_trials: List[FrozenTrial] = []
+    for trial in trials:
+        # Not a Number, positive infinity and negative infinity are considered to be non-finite.
+        if trial.value is not None and not np.isfinite(trial.value):
```
HideakiImamura, line 113: It seems that this utility assumes that the trial is single-objective optimization, since `trial.value` only supports the single-objective optimization. How about extending this to multi-objective optimization?

HideakiImamura (commented):
Thanks for the PR. Is the application of this utility follow-up tasks?

xadrianzetx:
>Is the application of this utility follow-up tasks?

Yep. I'll pick them up (maybe one by one, maybe bulk, don't know yet) after this one is merged.

optuna/visualization/_utils.py
xadrianzetx: Done in a16bf53b83a3435af2d321c5121c039cd5087cbd.

optuna/visualization/_utils.py
```diff
@@ -99,3 +103,19 @@ def _get_param_values(trials: List[FrozenTrial], p_name: str) -> List[Any]:
     if _is_numerical(trials, p_name):
         return values
     return list(map(str, values))
+
+
+def _filter_nonfinite(trials: List[FrozenTrial]) -> List[FrozenTrial]:
```
nzw0301, line 108: I realised that we need to care about transformed value by `target` that is used in visualization methods. I think we can extend this function by adding `target` argument.


```suggestion
def _filter_nonfinite(trials: List[FrozenTrial], target: Optional[Callable[[FrozenTrial], float]] = None) -> List[FrozenTrial]:
```

xadrianzetx: Good catch! So i guess we should be doing transformation like
```python
if target is None:
  value = trial.value
else:
  value = target(trial)
```
before doing checks with `isfinite`.

nzw0301: I think so too!

xadrianzetx: Done in f0a8618d6137ea870f1477880ad7ae847651db38. CC @HideakiImamura, this patch removes explicit handling of multi-objective studies, since those are [required to specify target](https://github.com/optuna/optuna/blob/f072d2cd8d4e8cd35244cf65ab720fd7366aad7e/optuna/visualization/_utils.py#L57-L61) which selects one of objectives to plot. From our point of view, we are always dealing with single objective.

HideakiImamura: Sounds reasonable. Thanks for the update.

HideakiImamura: How about leaving a comment that states this utility only supports the single-objective optimization?

optuna/visualization/_utils.py
```diff
@@ -108,6 +113,31 @@ def _get_param_values(trials: List[FrozenTrial], p_name: str) -> List[Any]:
...
+) -> List[FrozenTrial]:
+
+    filtered_trials: List[FrozenTrial] = []
+    for trial in trials:
+        if target is None:
+
+            def _target(t: FrozenTrial) -> float:
+                return cast(float, t.value)
+
+            target = _target
```
nzw0301, line 127: We can move `target` part to the outside of `for` block:
```suggestion
    filtered_trials: List[FrozenTrial] = []
    if target is None:

        def _target(t: FrozenTrial) -> float:
            return cast(float, t.value)

        target = _target    

    for trial in trials:
```

xadrianzetx: Ehh, I don't know how to code :facepalm:. Thanks and done in 817cef4700471fc41fbd451075e91ed81989cede!

optuna/visualization/_utils.py
xadrianzetx: Comment added as part of 817cef4700471fc41fbd451075e91ed81989cede :^)

HideakiImamura (approved):
Thanks for the update. LGTM.

PR #3403, opened 2022-03-23
Making optuna compatible with wandb sweep panels

captain-pool:
## Motivation
The trial scores logged to W&B using the `WeightsandBiasesCallback` cannot be used to create panels W&B Sweeps like panels, For example: Parameter Importance, Parallel Coordinates, etc. This pull request, fixes this problem, by giving the user an option to log each trial as a different wandb run. This new PR also allows an user, to log their own metrics (for ex: images, 3D objects etc.), along side the score generated by the objective in the same run for each trial.

## Description of the changes
- Tagging runs with with study_name, helps in grouping / filtering runs in the `as_sweeps` mode.
- Added an optional keyword to switch to logging trial results to different runs instead of one run.
-  reusing user runs to log trial metrics. (especially useful while searching parameters for deep learning training pipelines).

## TODO:
- [x] Unit Tests.
- [ ] Updating Optuna Examples


nzw0301:
Oh, amazing!! @xadrianzetx are you interested in seeing this PR?

optuna/integration/wandb.py
```diff
@@ -126,10 +156,25 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
-        wandb.log({**trial.params, **metrics}, step=trial.number)
+        run_id = trial._user_attrs.get("run_id")
+
+        if not (self._as_sweeps and run_id):
+            run = wandb.run or self._initialize_run()
+            run.log({**trial.params, **metrics})
+
+        else:
+            # If user is already logging wandb metrics from inside a trial
+            # reuse that run to log these study metrics
```
nzw0301, line 167: We expect a comment is a full sentence (ref: https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#block-and-inline-comments-are-supposed-to-be-sentences):
```suggestion
            # reuse that run to log these study metrics.
```

xadrianzetx:
Sure! Happy to review it,

optuna/integration/wandb.py
```diff
@@ -126,10 +155,33 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
+
+    def _initialize_run(self) -> wandb.sdk.wandb_run.Run:
+        """Initializes Weights & Biases run."""
+        run = wandb.init(tags=self._tags, **self._wandb_kwargs)
+        if not isinstance(run, wandb.sdk.wandb_run.Run):
+            raise OSError(
+                "Cannot create a Run."
+                "Expected wandb.sdk.wandb_run.Run as a return."
+                "Got: %s" % (type(run))
+            )
```
xadrianzetx, line 186: I think simple assertion is ok here.
```suggestion
        assert isinstance(run, wandb.sdk.wandb_run.Run)
```

xadrianzetx (changes requested):
Great to see new functionality for this callback! Let me leave few initial comments. BTW can you make sure `Checks` CI workflow is passing? You can run it locally [like that](https://github.com/optuna/optuna/blob/master/CONTRIBUTING.md#checking-the-format-coding-style-and-type-hints), and those jobs with status `Expected` should be re-triggered on next push or after merging master to this branch (this seems to sometimes happen on GitHub for some reason).

optuna/integration/wandb.py
```diff
@@ -51,15 +57,17 @@ def objective(trial):
                 x = trial.suggest_float("x", -10, 10)
                 return (x - 2) ** 2
 
+            study = optuna.create_study(study_name="my_study")
```
nzw0301, line 60: ditto (https://github.com/optuna/optuna/pull/3403/files#r835770611)
```suggestion

            study = optuna.create_study(study_name="my_study")
```

optuna/integration/wandb.py
nzw0301: In my understanding, [`assert`](https://docs.python.org/3/using/cmdline.html) can be ignored by using `-O` option: `python -O sample.py`, so I prefer to `raise`.

optuna/integration/wandb.py
```diff
@@ -97,8 +110,24 @@ def __init__(
 
         self._metric_name = metric_name
         self._wandb_kwargs = wandb_kwargs or {}
+        self._study_name = study_name
+
+        iapi = wandb.InternalApi()
```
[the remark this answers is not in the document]
captain-pool: Sure.

optuna/integration/wandb.py
```diff
@@ -126,10 +155,33 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
-    def _initialize_run(self) -> None:
-        """Initializes Weights & Biases run."""
+        if not (self._as_sweeps and run_id):
+            run = wandb.run or self._initialize_run()
+            step = trial.number if wandb.run else None
+            run.log({**trial.params, **metrics}, step=step)
+
+        else:
+            # If user is already logging wandb metrics from inside a trial
+            # reuse that run to log these study metrics.
```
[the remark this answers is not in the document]
captain-pool: Yes. That way, we don't create extra runs for logging study metrics if there's already a run present which an user is using to log their own metrics. An user can ofc choose to not use this feature, by not specifying the `run_id` to reuse in  `trial._user_attrs`.

nzw0301:
Failed tests of integration should be fixed by https://github.com/optuna/optuna/pull/3412 🙇🏻

optuna/integration/wandb.py
```diff
@@ -126,10 +155,33 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
-        wandb.config.update(attributes)
-        wandb.log({**trial.params, **metrics}, step=trial.number)
+        run_id = trial._user_attrs.get("run_id")
 
-    def _initialize_run(self) -> None:
-        """Initializes Weights & Biases run."""
+        if not (self._as_sweeps and run_id):
+            run = wandb.run or self._initialize_run()
+            step = trial.number if wandb.run else None
+            run.log({**trial.params, **metrics}, step=step)
```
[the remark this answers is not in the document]
captain-pool: You're right. I should put `step = ...` before `run = ...`. The reason I'm not putting `step = trial.number` in all the runs, is because, everytime a new run is initialized, the result should be logged at `step = 0`, not at `step = trial.number`.

optuna/integration/wandb.py
```diff
@@ -51,15 +57,17 @@ def objective(trial):
...
-
+            wandbc = WeightsAndBiasesCallback(study_name=study.name, wandb_kwargs=wandb_kwargs)
 
-            study = optuna.create_study(study_name="my_study")
             study.optimize(objective, n_trials=10, callbacks=[wandbc])
 
     Args:
+        study_name:
+            Study name for which the Callback is being used. This is used for tagging
+            run(s) generated during the experiment.
```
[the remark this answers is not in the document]
captain-pool: Sure.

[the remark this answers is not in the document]
captain-pool: I kept it this way to maintain uniformity in tagging runs in both the scenarios. Also, since one can put multiple tags to a run, adding an extra tag, should cause no harm and just act as an extra piece of information in identifying runs which an user may or may not choose to use. Also, auto grouping works only when `group=` is set in `wandb.init(...)`. Setting an extra tag should be virtually harmless.

Setting this tag can also be useful in the off chance of an user creates and logs multiple studies (or a single study multiple times) to the same project. Since run names and run-ids are randomly generated, having no useful information to identify the optuna study that created the run, tags can helpful in such scenarios.

captain-pool:
The formatting tests are passing. However, I'm facing an error in Sphinx. It says `wandb` is not defined for `optuna/integration/wandb.py:  def _initialize_run(self) -> wandb.sdk.wandb_run.Run`. What can I do to avoid this?

Also, I tried building the documentation locally, `make html` is raising a segmentation fault for some reason during this step:
```
generating gallery...
Using Sphinx-Gallery to convert rst text blocks to markdown for .ipynb files.
```
after the trials finishes.

Finally, running tests with `tox` was failing for `mypy` and was complaining some types were missing. I tried installing the types locally and reexecuting the command. Didn't work. So, I had to slightly update the `tox.ini` file and add `commands_pre = mypy --install-types --non-interactive` to the `mypy` test. It worked perfectly after that. Should I add that change to the PR as well? I am using python 3.8.12

nzw0301 (changes requested):
I skimmed the code to fix the sphinx error

nzw0301:
I'm not familiar with `tox.ini` and it sounds out-of-scope of this PR's aim, so we really appreciate you sending another PR to fix it.

optuna/integration/wandb.py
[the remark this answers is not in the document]
nzw0301: I agree with @xadrianzetx for two reasons:

1. This change breaks the backward compatibility of the callback. Even though this callback is still experimental, ideally we would like to avoid such a change. 
2. We can do the same thing by using `wandb_kwargs`. Related to this, the default `study_name` of `create_study` is `None`, which means users do not give a name to a study.

optuna/integration/wandb.py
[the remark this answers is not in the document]
xadrianzetx: Makes sense. There's similar functionality in Optuna's [mlflow callback](https://github.com/optuna/optuna/blob/ab272511c4c6ef6da3a7ee4c48a600cd699ff942/optuna/integration/mlflow.py#L192-L219) available via decorator to use on objective function where additional metrics/artifacts are logged. I wonder if we could do the same for W&B?

optuna/integration/wandb.py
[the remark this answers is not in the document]
captain-pool: Sure. But doesn't optuna create a random uuid study name if `None` is provided? The idea is to tag the runs with that study name. How can I do that without asking for the study name as a parameter?

optuna/integration/wandb.py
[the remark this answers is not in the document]
captain-pool: Nice! Will check it out. Thanks for sharing.

optuna/integration/wandb.py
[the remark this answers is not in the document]
xadrianzetx: IMO behavior would be:

* With `as_sweeps` being `False` - tagging run with study name would be purely optional and done with entry to `wandb_kwargs` specified by user. The reason is that in this mode 1 study = 1 wandb run, so even if user runs multiple studies as a part of the same project, it's obvious that runs are separate entities and it's up to user to decide how they want to distinguish between them (as @nzw0301 said, default study name might be used, which IMO does not really help to pinpoint experiment. Something like date would be more useful).
* With `as_sweeps` being `True` - here multiple wandb runs will refer to the same study, and if user runs multiple studies within the same project, we will lose track of what's what. I think name tagging can be useful here, since it solves this problem (doesn't matter what the study name is, as long as we can filter runs by it). In this mode, we are initializing new wandb run on each callback call and we have `study` instance available there, so we can use it to derive study name automatically and pass it to `wandb.init`.

captain-pool:
The tests seem to be failing for some other integration.

nzw0301:
The error has been fixed by https://github.com/optuna/optuna/pull/3427. Could your merge the latest master branch into the working branch?

optuna/integration/wandb.py
```diff
@@ -72,6 +75,11 @@ def objective(trial):
             Set of arguments passed when initializing Weights & Biases run.
             Please refer to `Weights & Biases API documentation
             <https://docs.wandb.ai/ref/python/init>`_ for more details.
+        as_multirun:
+            Creates new runs for each trial. Useful for generating W&B Sweeps like
+            panels (for ex., parameter importance, parallel coordinates, etc).
+        multirun_tag:
+            Tag name to tag runs which are generated during the study.
```
nzw0301, line 82: Do you plan to implement this argument since only `as_multirun` has been added to the arguments of the callback?

optuna/integration/wandb.py
```diff
@@ -52,11 +55,11 @@ def objective(trial):
                 return (x - 2) ** 2
 
 
+            study = optuna.create_study(multirun_tag="my_study")
```
nzw0301, line 58: I'm not sure but it might be as follows:
```suggestion
            study = optuna.create_study(study_name="my_study")
```

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,161 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
-    wandbc = WeightsAndBiasesCallback()
     study = optuna.create_study(direction="minimize")
+    wandbc = WeightsAndBiasesCallback()
     study.optimize(_objective_func, n_trials=1, callbacks=[wandbc])
 
     expected = {"direction": ["MINIMIZE"]}
-    wandb.config.update.assert_called_once_with(expected)
+    wandb.run.config.update.assert_called_once_with(expected)
+
+    #
```
nzw0301, line 93:
```suggestion
```

nzw0301 (changes requested):
Let me minor comments before seeing the details.

optuna/integration/wandb.py
```diff
@@ -126,10 +147,43 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
+            run.finish()
+
+    def track_in_wandb(self, func: ObjectiveFuncType) -> ObjectiveFuncType:
+        @functools.wraps(func)
+        def wrapper(trial: optuna.trial.Trial) -> Union[float, Sequence[float]]:
+
+            run = wandb.run  # Uses global run when `as_multirun` is set to False.
+            if not run:
+                run = self._initialize_run()
+                run.name = "trial/{trial.number}/{run.name}"
```
nzw0301, line 174:
```suggestion
                run.name = f"trial/{trial.number}/{run.name}"
```

optuna/integration/wandb.py
```diff
@@ -126,10 +147,43 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
 
-        wandb.init(**self._wandb_kwargs)
+        run.log({**trial.params, **metrics}, step=step)
+
+        run.config.update({**attributes})
+
+        if self._as_multirun:
+            run.finish()
+
+    def track_in_wandb(self, func: ObjectiveFuncType) -> ObjectiveFuncType:
```
nzw0301, line 167: If we add this public API, we need `experimental` decorator.
```suggestion
    @experimental("3.0.0")
    def track_in_wandb(self, func: ObjectiveFuncType) -> ObjectiveFuncType:
```

optuna/integration/wandb.py
```diff
@@ -126,10 +147,43 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
 
-        wandb.config.update(attributes)
-        wandb.log({**trial.params, **metrics}, step=trial.number)
+        step = trial.number if wandb.run else None
+        run = wandb.run
 
-    def _initialize_run(self) -> None:
-        """Initializes Weights & Biases run."""
+        # When an user doesn't use `track_in_wandb` decorator.
+        # Might create an extra run if an user logs in wandb but doesn't use the decorator.
```
nzw0301, line 154: I'm afraid that I could not imagine this case. Could you share a minimal example code for this use case?

optuna/integration/wandb.py
captain-pool: Oops, forgot to fix that. Working on it.

optuna/integration/wandb.py
captain-pool: Nope. I added this initially, and forgot to update. I’ll fix these comment sections.

optuna/integration/wandb.py
```diff
@@ -95,10 +104,22 @@ def __init__(
...
         self._wandb_kwargs = wandb_kwargs or {}
+        tags = self._wandb_kwargs.get("tags")
+
+        if as_multirun and not (
+            tags and isinstance(tags, Sequence) and all(map(lambda x: isinstance(x, str), tags))
+        ):
+            raise RuntimeError(
+                "atleast one `tag` in `wandb_kwargs` is necessary "
+                "when `as_multirun` is set to `True`."
+            )
```
nzw0301, line 116:
```suggestion
```

optuna/integration/wandb.py
```diff
@@ -126,10 +147,43 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
+            run = self._initialize_run()
+            run.name = f"trial/{trial.number}/{run.name}"
 
-        wandb.init(**self._wandb_kwargs)
+        run.log({**trial.params, **metrics}, step=step)
+
+        run.config.update({**attributes})
+
+        if self._as_multirun:
+            run.finish()
```
nzw0301, line 165: How about this?
```suggestion
            run.tags = tuple(self._wandb_kwargs.get("tags", ())) + (study.study_name, )
            run.finish()
```

nzw0301 (changes requested):

optuna/integration/wandb.py
nzw0301: Thank you for your clarification.

nzw0301:
When `as_multirun=True`, shall we store `trial.number` with `wandb.log`? It would let users easily access the results on wandb after optimisation using `trial.number` because we currently only know the `trial.number` from the run name with additional strings: `trial/` prefix and wandb's run name. What do you think about this?

optuna/integration/wandb.py
nzw0301: Mmore context: We can remove this when we store the tag as in https://github.com/optuna/optuna/pull/3403#discussion_r840878666.

optuna/integration/wandb.py
captain-pool: Sure. For example an user can do something like this:

```python3
def objective(trial):
  with wandb.init() as run:
    run.log({"test": 0})
  return loss
```

Instead of doing the correct thing:

```python3

@track_in_wandb
def objective(trial):
  wandb.log({"test": 1})
```

If an user uses the first option, there's no way for us to reuse that run. Therefore, when we try to log the params, it would create another run to log it, instead of reusing the last one.

optuna/integration/wandb.py
captain-pool: That would have worked if he had access to the `study` object 🙂. Is there way to get access it without asking the user to pass it as a parameter?

nzw0301: One of the arguments of the callback is  `study`.

xadrianzetx: Agree with @nzw0301, this is the logic I had in mind for this mode.

optuna/integration/wandb.py
```diff
@@ -126,10 +147,44 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
+        # Might create an extra run if an user logs in wandb but doesn't use the decorator.
+
+        if not run:
+            run = self._initialize_run()
+            run.name = f"trial/{trial.number}/{run.name}"
 
-        wandb.init(**self._wandb_kwargs)
+        run.log({**trial.params, **metrics}, step=step)
+
+        run.config.update({**attributes})
```
xadrianzetx, line 162: I think we wanted to store `trial.params` here as well to enable param importance plot. https://github.com/optuna/optuna/pull/3403#discussion_r835780841. Why did you revert that change?

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,165 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    study = optuna.create_study(direction="minimize", study_name="test_study")
     wandbc = WeightsAndBiasesCallback()
-    study = optuna.create_study(direction="minimize")
     study.optimize(_objective_func, n_trials=1, callbacks=[wandbc])
 
     expected = {"direction": ["MINIMIZE"]}
-    wandb.config.update.assert_called_once_with(expected)
+    wandb.run.config.update.call_args[0][0].pop("x")
+    wandb.run.config.update.call_args[0][0].pop("y")
+    wandb.run.config.update.assert_called_once_with(expected)
```
xadrianzetx, line 93: I think we could use [enqueue_trial](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html#optuna.study.Study.enqueue_trial) here to control parameters suggested, so we can test if `run.config` contains all expected values.
```python
study = optuna.create_study(direction="minimize", study_name="test_study")
study.enqueue_trial({"x": 5.0, "y": 1.0})
study.optimize(_objective_func, n_trials=1, callbacks=[wandbc])
expected = {"direction": ["MINIMIZE"], "x": 5.0, "y": 1.0]}
wandb.run.config.update.assert_called_once_with(expected)
```

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,165 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    _wrapped_func = wandbc.track_in_wandb(_objective_func)
+    study.optimize(_wrapped_func, n_trials=1, callbacks=[wandbc])
+    wandb.run.config.update.call_args[0][0].pop("x")
+    wandb.run.config.update.call_args[0][0].pop("y")
+    wandb.run.config.update.assert_called_once_with(expected)
+
+    wandb.run = None
+    study.optimize(_objective_func, n_trials=1, callbacks=[wandbc])
+    wandb.init().config.update.call_args[0][0].pop("x")
+    wandb.init().config.update.call_args[0][0].pop("y")
```
xadrianzetx, line 106: Ditto.

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,165 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    wandb.run.config.update.call_args[0][0].pop("x")
+    wandb.run.config.update.call_args[0][0].pop("y")
+    wandb.run.config.update.assert_called_once_with(expected)
+
+    wandbc = WeightsAndBiasesCallback(as_multirun=True)
+    wandb.run = mock.MagicMock()
+    _wrapped_func = wandbc.track_in_wandb(_objective_func)
+    study.optimize(_wrapped_func, n_trials=1, callbacks=[wandbc])
+    wandb.run.config.update.call_args[0][0].pop("x")
+    wandb.run.config.update.call_args[0][0].pop("y")
```
xadrianzetx, line 100: Ditto.

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,165 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
 
-    wandbc = WeightsAndBiasesCallback()
     study = optuna.create_study(directions=["minimize", "maximize"])
+    wandbc = WeightsAndBiasesCallback()
     study.optimize(_multiobjective_func, n_trials=1, callbacks=[wandbc])
 
     expected = {"direction": ["MINIMIZE", "MAXIMIZE"]}
-    wandb.config.update.assert_called_once_with(expected)
+    wandb.run.config.update.call_args[0][0].pop("x")
+    wandb.run.config.update.call_args[0][0].pop("y")
```
xadrianzetx, line 121: Ditto.

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,165 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    _wrapped_func = wandbc.track_in_wandb(_multiobjective_func)
+    study.optimize(_wrapped_func, n_trials=1, callbacks=[wandbc])
+    wandb.run.config.update.call_args[0][0].pop("x")
+    wandb.run.config.update.call_args[0][0].pop("y")
+    wandb.run.config.update.assert_called_once_with(expected)
+
+    wandb.run = None
+    study.optimize(_multiobjective_func, n_trials=1, callbacks=[wandbc])
+    wandb.init().config.update.call_args[0][0].pop("x")
+    wandb.init().config.update.call_args[0][0].pop("y")
```
xadrianzetx, line 135: Ditto.

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,165 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    wandb.run.config.update.call_args[0][0].pop("x")
+    wandb.run.config.update.call_args[0][0].pop("y")
+    wandb.run.config.update.assert_called_once_with(expected)
+
+    wandbc = WeightsAndBiasesCallback(as_multirun=True)
+    wandb.run = mock.MagicMock()
+    _wrapped_func = wandbc.track_in_wandb(_multiobjective_func)
+    study.optimize(_wrapped_func, n_trials=1, callbacks=[wandbc])
+    wandb.run.config.update.call_args[0][0].pop("x")
+    wandb.run.config.update.call_args[0][0].pop("y")
```
xadrianzetx, line 129: Ditto.

xadrianzetx (changes requested):

tests/integration_tests/test_wandb.py
captain-pool: Fixed. Thanks for pointing it out.

xadrianzetx:
Sorry for small delay @captain-pool. I agree with @nzw0301 on the https://github.com/optuna/optuna/pull/3403#issuecomment-1086282560, what do you think about this?

captain-pool:
@nzw0301 @xadrianzetx added the changes you requested. Lmk if there's anything else.

nzw0301:
Thank you for your updates. I'm afraid I'm really busy this month, but I'll review this PR hopefully next week.

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,163 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+
+    wandb.init = mock.MagicMock()
+    study.optimize(_objective_func, n_trials=1, callbacks=[wandbc])
+    wandb.init.assert_called_once_with(
+        project="optuna", group="summary", job_type="logging", mode="offline", tags=["test-tag"]
     )
 
 
 @mock.patch("optuna.integration.wandb.wandb")
 def test_attributes_set_on_epoch(wandb: mock.MagicMock) -> None:
```
xadrianzetx, line 81: General comments to this test:
* We don't have to test `track_in_wandb` decorator, as it plays no part in setting the attributes.
* For the remaining part, I think we can keep logic from `master` (modified with `enqueue_trial` and mocking correct objects, as in your test ofc) and just parametrize `as_multirun`. Behavior doesn't change.

Something like:
```python
@mock.patch("optuna.integration.wandb.wandb")
@pytest.mark.parametrize("as_multirun", [True, False])
def test_attributes_set_on_epoch(wandb: mock.MagicMock, as_multirun: bool) -> None:

    # Vanilla update
    wandb.sdk.wandb_run.Run = mock.MagicMock

    wandbc = WeightsAndBiasesCallback(as_multirun=as_multirun)
    study = optuna.create_study(direction="minimize")
    study.enqueue_trial({"x": 1.1, "y": 2.2})
    study.optimize(_objective_func, n_trials=1, callbacks=[wandbc])

    expected = {"direction": ["MINIMIZE"], "x": 1.1, "y": 2.2}
    wandb.run.config.update.assert_called_once_with(expected)
```

*Didn't check if linters pass :^)*

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,163 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    wandb.run.config.update.assert_called_once_with(expected)
+
+    wandb.run = None
+    study.enqueue_trial(trial_params)
+    study.optimize(_objective_func, n_trials=1, callbacks=[wandbc])
+    wandb.init().config.update.assert_called_once_with(expected)
 
 
 @mock.patch("optuna.integration.wandb.wandb")
 def test_multiobjective_attributes_set_on_epoch(wandb: mock.MagicMock) -> None:
```
xadrianzetx, line 110: Same comments as in `test_attributes_set_on_epoch`.

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,163 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    study = optuna.create_study()
     wandbc = WeightsAndBiasesCallback()
     target_n_trials = 10
-    study = optuna.create_study()
     study.optimize(_objective_func, n_trials=target_n_trials, callbacks=[wandbc])
-    assert wandb.log.call_count == target_n_trials
+    assert wandb.run.log.call_count == target_n_trials
+
+    wandbc = WeightsAndBiasesCallback(as_multirun=True)
+    wandb.run = mock.MagicMock()
```
xadrianzetx, line 149:
```suggestion
    wandb.run.reset_mock()
```

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,163 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    _wrapped_func = wandbc.track_in_wandb(_objective_func)
+    study.optimize(_wrapped_func, n_trials=target_n_trials, callbacks=[wandbc])
```
xadrianzetx, line 151: If we want to actually test if behavior is correct when objective function is decorated, we should prepare one, call `wandb.log` in it, and then assert `call_count == 2 * target_n_trials` (as we should be calling in objective function and callback per each trial). Rest of the test looks good imo.

tests/integration_tests/test_wandb.py
```diff
@@ -31,79 +32,163 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    study.optimize(_objective_func, n_trials=target_n_trials, callbacks=[wandbc])
+    assert wandb.init().log.call_count == target_n_trials
 
 
 @pytest.mark.parametrize(
     "metric,expected", [("value", ["x", "y", "value"]), ("foo", ["x", "y", "foo"])]
 )
 @mock.patch("optuna.integration.wandb.wandb")
-def test_values_registered_on_epoch(wandb: mock.Mock, metric: str, expected: List[str]) -> None:
+def test_values_registered_on_epoch(
```
xadrianzetx, line 164: General comments here:
* How about moving decorator tests to separate case  with custom objective function calling `wandb.log`. We should be testing if all metrics (callback + decorator) are registered.
* Remaining logic in this test can be simplified by using `pytest.mark.parametrize` instead of deepcopying a study.

tests/integration_tests/test_wandb.py
```diff
@@ -116,26 +201,43 @@ def test_values_registered_on_epoch(wandb: mock.Mock, metric: str, expected: Lis
 )
 @mock.patch("optuna.integration.wandb.wandb")
 def test_multiobjective_values_registered_on_epoch(
```
xadrianzetx, line 203: Ditto.

xadrianzetx (changes requested):
Callback logic now looks good apart from small bug I've missed earlier (sorry!). Let me leave some comments on tests.,

optuna/integration/wandb.py
```diff
@@ -124,12 +134,51 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
+            run.name = f"trial/{trial.number}/{run.name}"
 
-        wandb.init(**self._wandb_kwargs)
+        run.log({**trial.params, **metrics}, step=step)
+
+        run.config.update({**attributes, **trial.params})
+
+        if self._as_multirun:
+            run.tags = tuple(self._wandb_kwargs.get("tags", ())) + (study.study_name,)
+            run.finish()
```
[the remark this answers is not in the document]
captain-pool: I think this is happening because  config.update tries to update the values of trial.params for a single run. Apparently that is forbidden by default, since config is meant to be static. In that case, for keeping consistency of logged data between two modes, we should either: remove `trial.params` in both the cases, or use `allow_value_change=True` when `as_multirun=False` while logging `trial.params`. IMO, either cases would work fine. But, just to give the user an extra option on how they should retrieve their parameters, maybe we should log `trial.params` in the config in both cases.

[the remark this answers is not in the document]
xadrianzetx: If config was meant to be static, let's keep it that way, especially since we are using it only to get param importance plot in W&B, which does not work with `as_multirun=False` anyway. For consistency, we are logging params with `wandb.log` in both modes, so I'm fine with logging them in config conditionally.

optuna/integration/wandb.py
```diff
@@ -124,12 +134,52 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
+
+        if self._as_multirun:
+            run.config.update({**attributes, **trial.params})
+            run.tags = tuple(self._wandb_kwargs.get("tags", ())) + (study.study_name,)
+            run.finish()
+        else:
+            run.config.update(attributes)
+
+    @experimental("3.0.0")
+    def track_in_wandb(self, func: ObjectiveFuncType) -> ObjectiveFuncType:
```
nzw0301, line 163: TODO: we need to add documentation here like https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.MLflowCallback.html#optuna.integration.MLflowCallback. But it is completely okay to add it as a follow-up.

tests/integration_tests/test_wandb.py
```diff
@@ -28,114 +30,267 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    assert wandb.init().finish.call_count == n_trials
+
 
 @mock.patch("optuna.integration.wandb.wandb")
-def test_attributes_set_on_epoch(wandb: mock.MagicMock) -> None:
+@pytest.mark.parametrize("as_multirun", [True, False])
+def test_attributes_set_on_epoch(wandb: mock.MagicMock, as_multirun: bool) -> None:
 
-    wandb.config.update = mock.MagicMock()
+    # Vanilla update
```
nzw0301, line 98: Let's follow Optuna's coding style.
```suggestion
    # Vanilla update.
```

tests/integration_tests/test_wandb.py
```diff
@@ -28,114 +30,267 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    wandb.sdk.wandb_run.Run = mock.MagicMock
+    expected = {"direction": ["MINIMIZE"]}
+    trial_params = {"x": 1.1, "y": 2.2}
+    expected_with_params = {"direction": ["MINIMIZE"], "x": 1.1, "y": 2.2}
```
nzw0301, line 102:
```suggestion
    expected_config_with_params = expected_config | trial_params
```

tests/integration_tests/test_wandb.py
```diff
@@ -28,114 +30,267 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
 
 @mock.patch("optuna.integration.wandb.wandb")
-def test_attributes_set_on_epoch(wandb: mock.MagicMock) -> None:
+@pytest.mark.parametrize("as_multirun", [True, False])
+def test_attributes_set_on_epoch(wandb: mock.MagicMock, as_multirun: bool) -> None:
 
-    wandb.config.update = mock.MagicMock()
+    # Vanilla update
+    wandb.sdk.wandb_run.Run = mock.MagicMock
+    expected = {"direction": ["MINIMIZE"]}
```
nzw0301, line 100:
```suggestion
    expected_config = {"direction": ["MINIMIZE"]}
```

tests/integration_tests/test_wandb.py
```diff
@@ -28,114 +30,267 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
 @mock.patch("optuna.integration.wandb.wandb")
-def test_multiobjective_attributes_set_on_epoch(wandb: mock.MagicMock) -> None:
+@pytest.mark.parametrize("as_multirun", [True, False])
+def test_multiobjective_attributes_set_on_epoch(wandb: mock.MagicMock, as_multirun: bool) -> None:
 
-    wandb.config.update = mock.MagicMock()
+    wandb.sdk.wandb_run.Run = mock.MagicMock
+    trial_params = {"x": 1.1, "y": 2.2}
+    expected = {"direction": ["MINIMIZE", "MAXIMIZE"]}
+    expected_with_params = {"direction": ["MINIMIZE", "MAXIMIZE"], "x": 1.1, "y": 2.2}
```
nzw0301, line 125: Unify the definition order and informative names.
```suggestion
    expected_config = {"direction": ["MINIMIZE", "MAXIMIZE"]}
    trial_params = {"x": 1.1, "y": 2.2}    
    expected_config_with_params = expected_config | trial_params
```

tests/integration_tests/test_wandb.py
```diff
@@ -28,114 +30,267 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    "metric,as_multirun,expected",
+    [("value", False, ["x", "y", "value"]), ("foo", True, ["x", "y", "foo", "trial_number"])],
 )
 @mock.patch("optuna.integration.wandb.wandb")
-def test_values_registered_on_epoch(wandb: mock.Mock, metric: str, expected: List[str]) -> None:
+def test_values_registered_on_epoch(
+    wandb: mock.MagicMock, metric: str, as_multirun: bool, expected: List[str]
+) -> None:
+    def assert_call_args(log_func: mock.MagicMock, regular: bool) -> None:
+        kall = log_func.call_args
```
nzw0301, line 177: Could you rename `kall` with a clearer name for Optuna's other maintainers? I could not understand the meaning of this variable.

tests/integration_tests/test_wandb.py
```diff
@@ -28,114 +30,267 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
 @mock.patch("optuna.integration.wandb.wandb")
 def test_multiobjective_values_registered_on_epoch(
-    wandb: mock.Mock, metrics: Union[str, Sequence[str]], expected: List[str]
+    wandb: mock.MagicMock,
+    metrics: Union[str, Sequence[str]],
+    as_multirun: bool,
+    expected: List[str],
 ) -> None:
+    def assert_call_args(log_func: mock.MagicMock, regular: bool) -> None:
+        kall = log_func.call_args
```
nzw0301, line 237: ditto

nzw0301 (changes requested):
The main part looks awesome. Thanks. Let me leave minor suggestions.

tests/integration_tests/test_wandb.py
nzw0301: oh, nice comment. I didn't know that.

captain-pool: Don’t think we can. MyPy complains if I do something like:
```
expected.update(trial_params)
```
or
```
expected_config_with_params = {**expected, **trial_params}
```
That’s because, all the entries of `expected` are strings, but the values of `trial_params` are floating point numbers, and mypy complains if we dynamically try to create a heterogeneous dictionary.

I thought maybe we could use `TypedDict` for this, but importing `TypedDict` is different for python 3.9 and other lower versions. Do you guys have any suggestions?

optuna/integration/wandb.py
nzw0301: Ah, i got it. Thank you for your explanaion.

optuna/integration/wandb.py
```diff
@@ -124,12 +134,52 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
...
 
-        wandb.config.update(attributes)
-        wandb.log({**trial.params, **metrics}, step=trial.number)
+        step = trial.number if wandb.run else None
+        run = wandb.run
 
-    def _initialize_run(self) -> None:
-        """Initializes Weights & Biases run."""
+        # When an user doesn't use `track_in_wandb` decorator.
+        # Might create an extra run if an user logs in wandb but doesn't use the decorator.
```
nzw0301, line 147: Typos
```suggestion
        # When n user doesn't use `track_in_wandb` decorator.
        # Might create an extra run if a user logs in wandb but doesn't use the decorator.
```

tests/integration_tests/test_wandb.py
xadrianzetx: How about declaring a type for `expected`? The test becomes:
```python
from typing import Any


@mock.patch("optuna.integration.wandb.wandb")
@pytest.mark.parametrize("as_multirun", [True, False])
def test_attributes_set_on_epoch(wandb: mock.MagicMock, as_multirun: bool) -> None:

    # Vanilla update.
    wandb.sdk.wandb_run.Run = mock.MagicMock
    expected: Dict[str, Any] = {"direction": ["MINIMIZE"]}
    trial_params = {"x": 1.1, "y": 2.2}

    study = optuna.create_study(direction="minimize")
    wandbc = WeightsAndBiasesCallback(as_multirun=as_multirun)
    study.enqueue_trial(trial_params)
    study.optimize(_objective_func, n_trials=1, callbacks=[wandbc])

    if as_multirun:
        wandb.run = None
        expected.update(trial_params)
        study.enqueue_trial(trial_params)
        study.optimize(_objective_func, n_trials=1, callbacks=[wandbc])
        wandb.init().config.update.assert_called_once_with(expected)
    else:
        wandb.run.config.update.assert_called_once_with(expected)
```

optuna/integration/wandb.py
```diff
@@ -160,18 +159,42 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
             run.config.update(attributes)
 
     @experimental("3.0.0")
-    def track_in_wandb(self, func: ObjectiveFuncType) -> ObjectiveFuncType:
-        @functools.wraps(func)
-        def wrapper(trial: optuna.trial.Trial) -> Union[float, Sequence[float]]:
+    def track_in_wandb(self) -> Callable:
+        """Decorator for using W&B for logging inside the objective function.
+
+        The run is initialized with the same `wandb_kwargs` that are passed to the callback.
```
nzw0301, line 165:
```suggestion
        The run is initialized with the same ``wandb_kwargs`` that are passed to the callback.
```

optuna/integration/wandb.py
```diff
@@ -143,8 +143,7 @@ def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -
         step = trial.number if wandb.run else None
         run = wandb.run
 
-        # When an user doesn't use `track_in_wandb` decorator.
-        # Might create an extra run if an user logs in wandb but doesn't use the decorator.
+        # Might create extra runs if an user logs in wandb but doesn't use the decorator.
```
nzw0301, line 146: typo:

```suggestion
        # Might create extra runs if a user logs in wandb but doesn't use the decorator.
```

nzw0301 (changes requested):

nzw0301:
Thank you for updating the PR and adding examples.
I suppose it would be great to show the same example as the second example code on https://optuna.readthedocs.io/en/latest/reference/generated/optuna.integration.MLflowCallback.html because it is minimal reproducible.

captain-pool:
@nzw0301 not quite sure where I made a mistake, but the doctest is failing. Can you please look into it?

optuna/integration/wandb.py
```diff
@@ -52,11 +56,52 @@ def objective(trial):
...
+                return (x - 2) ** 2
 
-            study = optuna.create_study(study_name="my_study")
+            study = optuna.create_study(as_multirun=True)
+            study.optimize(objective, n_trials=10, callbacks=[wandbc])
+
+
+        Add additional logging to Weights & Biases.
+
+        ..code::
```
nzw0301, line 88: We need space.
```suggestion
        .. code::
```

optuna/integration/wandb.py
```diff
@@ -52,11 +56,52 @@ def objective(trial):
...
+            wandbc = WeightsAndBiasesCallback(wandb_kwargs=wandb_kwargs)
+
+            study.optimize(objective, n_trials=10, callbacks=[wandbc])
+
+
+
+        Weights & Biases logging in multirun mode.
+
+        ..code::
+            import optuna
```
nzw0301, line 71:
```suggestion
        .. code::

            import optuna
```

nzw0301 (commented):
Hopefully, this fixed the CI error.

tests/integration_tests/test_wandb.py
```diff
@@ -28,114 +31,268 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+    wandb.sdk.wandb_run.Run = mock.MagicMock
+
+    study = optuna.create_study()
+    wandbc = WeightsAndBiasesCallback(metric_name=metric, as_multirun=True)
+    _wrapped_func = wandbc.track_in_wandb()(
+        functools.partial(logging_objective_func, log_func=wandb.run.log)
+    )
+
+    study.enqueue_trial({"x": 2, "y": 25})
+    study.optimize(_wrapped_func, n_trials=1, callbacks=[wandbc])
```
xadrianzetx, line 212: Not a strong opinion, but we call `logging_objective_func` only twice (in this test and multiobjective counterpart), so maybe allow for small repetition in favor of readability. What do you think?
```suggestion
    wandbc = WeightsAndBiasesCallback(metric_name=metric, as_multirun=True)

    @wandbc.track_in_wandb()
    def _decorated_objective(trial: optuna.trial.Trial) -> float:
        result = _objective_func(trial)
        wandb.run.log({"result": result})
        return result

    study.enqueue_trial({"x": 2, "y": 25})
    study.optimize(_decorated_objective, n_trials=1, callbacks=[wandbc])
```

tests/integration_tests/test_wandb.py
```diff
@@ -28,114 +31,268 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+
+    logged_metric = wandb.run.log.mock_calls[0][1][0]
+
+    call_args = wandb.run.log.call_args
+    assert list(call_args[0][0].keys()) == expected
+    assert call_args[1] == {"step": 0}
+    assert logged_metric == {"result": 0}
```
xadrianzetx, line 219: I think that we can be a bit clearer what is being tested here - `wandb.run.log` should be called twice (from decorator and from callback), so IMO we can test following:
* Number of calls
* Payload of the first call
* Payload of the second call

```suggestion
    logged_in_decorator = wandb.run.log.mock_calls[0][1][0]
    logged_in_callback = wandb.run.log.mock_calls[1][1][0]
    assert len(wandb.run.log.mock_calls) == 2
    assert list(logged_in_decorator) == ["result"]
    assert list(logged_in_callback) == expected

    call_args = wandb.run.log.call_args
    assert call_args[1] == {"step": 0}
```

Also tesing for metric + values would be nicer, but I guess we would need to remove parametrization. What do you think?

tests/integration_tests/test_wandb.py
```diff
@@ -28,114 +31,268 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+
+    wandbc = WeightsAndBiasesCallback(as_multirun=True, metric_name=metrics)
+    _wrapped_func = wandbc.track_in_wandb()(
+        functools.partial(logging_multiobjective_func, log_func=wandb.run.log)
+    )
+
+    study = optuna.create_study(directions=["minimize", "maximize"])
+    study.enqueue_trial({"x": 2, "y": 24})
+
+    study.optimize(_wrapped_func, n_trials=1, callbacks=[wandbc])
```
xadrianzetx, line 278: Ditto. Also, how about removing unnecessary spacing?

tests/integration_tests/test_wandb.py
```diff
@@ -28,114 +31,268 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
+
+    logged_metrics = wandb.run.log.mock_calls[0][1][0]
+
+    call_args = wandb.run.log.call_args
+    assert list(call_args[0][0].keys()) == expected
+    assert call_args[1] == {"step": 0}
+    assert logged_metrics == {"result0": 1, "result1": -1}
```
xadrianzetx, line 285: Ditto.

xadrianzetx (changes requested):
I have few remaining comments on tests.

tests/integration_tests/test_wandb.py
captain-pool: Can you please elaborate on what you mean by testing metric + values? An example would be very helpful.

captain-pool: Not sure what you are saying, can you elaborate it a bit?

xadrianzetx: At the current stage test asserts if metric names are correct:
```python
assert list(logged_in_decorator) == ["result"]
assert list(logged_in_callback) == expected
```
I think it's because we are passing `expected` as a list of metric names via `mark.parametrize` decorator. IMO we could pass a dict of metric names and respective values and assert something like following:
```python
assert logged_in_decorator == {"result": 0}
assert logged_in_callback == {"x": 1.0, "y": 0.0, "foo":...}
```
But I'm fine with leaving this as future work.

optuna/integration/wandb.py
```diff
@@ -72,6 +122,9 @@ def objective(trial):
             Set of arguments passed when initializing Weights & Biases run.
             Please refer to `Weights & Biases API documentation
             <https://docs.wandb.ai/ref/python/init>`_ for more details.
+        as_multirun:
+            Creates new runs for each trial. Useful for generating W&B Sweeps like
+            panels (for ex., parameter importance, parallel coordinates, etc).
```
xadrianzetx, line 127: Note - At the moment wandb attempts to upload data after every trial in this mode, which in some cases visibly affects performance. Possible solution is to cache all runs locally and then upload once when optimization is finished, but that might require introducing additional callback hooks in Optuna. Leaving for future work.

xadrianzetx (approved):
LGTM! New panels look amazing on W&B!
Screenshot from 2022-04-15 18-33-42
*Created with [this](https://gist.github.com/xadrianzetx/f246c2368be4d1c7592d69cdaeca4d91) script.*

captain-pool:
@nzw0301 it says you requested some changes, but for some reason, I don't seem to find them. Is this change request still open? If yes, can you please point it out?

tests/integration_tests/test_wandb.py
```diff
@@ -31,111 +32,265 @@ def _multiobjective_func(trial: optuna.trial.Trial) -> Tuple[float, float]:
...
-        (["foo", "bar"], ["x", "y", "foo", "bar"]),
-        (("foo", "bar"), ["x", "y", "foo", "bar"]),
+        ("value", False, ["x", "y", "value_0", "value_1"]),
+        ("value", True, ["x", "y", "value_0", "value_1", "trial_number"]),
+        (["foo", "bar"], False, ["x", "y", "foo", "bar"]),
+        (("foo", "bar"), True, ["x", "y", "foo", "bar", "trial_number"]),
     ],
 )
 @mock.patch("optuna.integration.wandb.wandb")
 def test_multiobjective_values_registered_on_epoch(
```
nzw0301, line 225: same as `test_values_registered_on_epoch`'s comment.

nzw0301 (changes requested):
Sorry for the delayed respose. Could you take a look at my suggestions?

nzw0301 (approved):
Brilliant! Thank you for improving Optuna WandBCallback and creating long-running PR!

PR #3406, opened 2022-03-24
Remove unnecessary string concatenation

himkt:
I found the unnecessary string concatenation.


not522 (approved):
LGTM

toshihikoyanase (approved):
Good catch! Thank you.

Issue #3418, opened 2022-03-30
Support latest PyTorch lightning


himkt:
### Expected behavior

Optuna works with the latest PyTorch-Lightning (1.6.0).

### Environment

- Optuna version: 3.0.0b0.dev
- Python version: 
- OS: Independent
- (Optional) Other libraries and their versions: PyTorch-Lightning 1.6.0

### Error messages, stack traces, or logs

```shell
tests/integration_tests/test_pytorch_lightning.py:158: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
optuna/study/study.py:399: in optimize
    _optimize(
optuna/study/_optimize.py:68: in _optimize
    _optimize_sequential(
optuna/study/_optimize.py:162: in _optimize_sequential
    trial = _run_trial(study, func, catch)
optuna/study/_optimize.py:262: in _run_trial
    raise func_err
optuna/study/_optimize.py:211: in _run_trial
    value_or_values = func(trial)
tests/integration_tests/test_pytorch_lightning.py:143: in objective
    trainer = pl.Trainer(
/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/pytorch_lightning/utilities/argparse.py:339: in insert_env_defaults
    return fn(self, **kwargs)
/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py:561: in __init__
    self._call_callback_hooks("on_init_start")
/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py:1617: in _call_callback_hooks
    fn(self, *args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <optuna.integration.pytorch_lightning.PyTorchLightningPruningCallback object at 0x7f892c1c5730>
trainer = <pytorch_lightning.trainer.trainer.Trainer object at 0x7f892c1c5040>

    def on_init_start(self, trainer: Trainer) -> None:
>       self.is_ddp_backend = trainer._accelerator_connector.distributed_backend is not None
[...]

PR #3431, opened 2022-04-01
Use latest pl

nzw0301:
## Motivation

investigation of https://github.com/optuna/optuna/issues/3418.

## Description of the changes

- Use PyTorch 1.11.0 and PyTorch-Lightning 1.6.0
- Fix pl callback


nzw0301:
I suppose `on_fit_end` is called at every (child) process, which causes the error.

nzw0301:
Let me ask pl community... https://github.com/PyTorchLightning/pytorch-lightning/discussions/12569

nzw0301:
I further investigated this a little bit. A similar library, [`ray-lightning`](https://github.com/ray-project/ray_lightning), uses `PlugIn` (or let's say `Strategy` in PyTorchLightning 1.6). I do not really understand what this library does, but using `Strategy` instead of a callback looks like an alternative approach to replace our DDP functionality for PL.

nzw0301:
As a simple solution, we run a logic defined in `on_fit_end` after `trainer.fit`. This approach is similar to [`CatboostCallback`](https://github.com/optuna/optuna/blob/2a208f045cdac0b8c88102fc73f02308f348e493/optuna/integration/catboost.py#L104-L107).

nzw0301:
Well, PyTorch's tests were okay, but catboost's tests failed with python 3.7.

PR #3435, opened 2022-04-02
Make EDF plots handle trials with nonfinite values

xadrianzetx:
## Motivation

Part of #3382

## Description of the changes

* Update EDF plot on both Plotly and Matplotlib backends
* Tests

## Example
Examples are generated using [this](https://gist.github.com/HideakiImamura/67a150dec5774a18732527875774a9fe) notebook. Current master yields empty plots for both backends. This PR:

* Plotly
Screenshot from 2022-04-02 18-08-38

* Matplotlib
Screenshot from 2022-04-02 18-09-01

### DIFF
--- a/tests/visualization_tests/matplotlib_tests/test_edf.py
+++ b/tests/visualization_tests/matplotlib_tests/test_edf.py
@@ -3,7 +3,9 @@ from typing import Sequence
 import numpy as np
 import pytest
 
+from optuna.distributions import FloatDistribution
 from optuna.study import create_study
+from optuna.trial import create_trial
 from optuna.visualization.matplotlib import plot_edf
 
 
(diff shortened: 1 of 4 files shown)


nzw0301:
Awesome. When I ran a test implemented in this PR, users see the same log message as follows:

```
[I 2022-04-03 17:05:05,749] A new study created in memory with name: no-name-bcbc6e99-72f4-4abe-9bb2-20d4b78d8feb
[I 2022-04-03 17:05:05,757] Trial 1 is omitted in visualization because its objective value is inf or nan.
[I 2022-04-03 17:05:05,758] Trial 1 is omitted in visualization because its objective value is inf or nan.
```

I suppose one message is enough. So I came up with two ways to avoid this:

1. Replace `_logger.info` with `warnings.warn`. By doing so, we can add more detailed assertions to tests implemented by https://github.com/optuna/optuna/pull/3395 like [this](https://github.com/optuna/optuna/blob/3f1e77ffb2e47762f14a54b86907a355148be4ee/tests/visualization_tests/test_utils.py#L64-L65).
2. Add a boolean argument to decide whether or not to show the log message. If this value is `Fase`, we remove the invalidate trials without logging messages to not repeat the same logging message.

What do you think about this?

xadrianzetx:
I would keep it a logger message as it seems more in line with convention in wiki. For testing, we can use [caplog fixture](https://docs.pytest.org/en/6.2.x/logging.html#caplog-fixture). As for the flag, good idea, let me put this PR back to draft for a moment and implement this feature in separate PR.

Also, I think EDF plot functions can be refactored to query the trials just once, I'll look into it more some other time.

tests/visualization_tests/matplotlib_tests/test_edf.py
```diff
@@ -76,3 +78,59 @@ def test_plot_optimization_history(direction: str) -> None:
...
+    study.add_trial(
+        create_trial(
+            value=value,
+            params={"x": 0.0},
+            distributions={"x": FloatDistribution(0.0, 1.0)},
+        )
+    )
+
+    figure = plot_edf(study)
+    assert all(np.isfinite(figure.get_lines()[0].get_xdata()))
```
nzw0301, line 103: This assertion was fine on the latest master branch as well, we might need to finer assertion. I tried `figure.findobj(PathCollection)` by following `test_slice.py`, but it returned empty.

Note that Plotly's tests look perfect.

nzw0301: I think this might be a difficult issue; we can work on this as a follow-up and keep the tests as is in this PR.

nzw0301: This was my mistake (actually I checked on this branch, not on the master). The test was perfect. Sorry for confusing you.

xadrianzetx: Sounds good to me!

HideakiImamura:
@xadrianzetx #3438 has been merged. Could you move this PR forward?

nzw0301 (approved):
LGTM. Thanks!

PR #3436, opened 2022-04-02
Fix log message formatting in `filter_nonfinite`

xadrianzetx:
## Motivation

Since commas were used to separate parts of log message, second part was passed as positional argument instead of being concatenated to first part, causing logger to raise warning.

## Description of the changes

* Fix log message


nzw0301 (approved):
Nice catch. Thanks!

PR #3437, opened 2022-04-03
Remove read_trials_from_remote_storage

keisuke-umezawa:
## Motivation
Currently, the method `read_trials_from_remote_storage` is implemented in the all storages classes. However, only the `CachedStorage` has meaningful implementation actually, because:

1. The `read_trials_from_remote_storage` in `RDBStorage` and `RedisStorage` is never called, because they are always wrapped by `CachedStorage`, and `CachedStorage` never calls `read_trials_from_remote_storage` in the wrapped storages.
2. The `read_trials_from_remote_storage` in `InMemoryStorage` just checks if study_id exists or not.

As a result, we can remove `read_trials_from_remote_storage` implementations other than `CachedStorage`.

To remove the implementations, we can move the calls of `read_trials_from_remote_storage` from study to `get_all_trials` in CachedStorage, because `read_trials_from_remote_storage` is always called before the calls of `get_all_trials`.

## Description of the changes

- [x] Remove the `read_trials_from_remote_storage` implementations other than `CachedStorage`
- [x] Move the calls of `read_trials_from_remote_storage` from study to `get_all_trials` in CachedStorage

### DIFF
--- a/optuna/storages/_cached_storage.py
+++ b/optuna/storages/_cached_storage.py
@@ -334,8 +334,8 @@ class _CachedStorage(BaseStorage):
         deepcopy: bool = True,
         states: Optional[Container[TrialState]] = None,
     ) -> List[FrozenTrial]:
-        if study_id not in self._studies:
-            self.read_trials_from_remote_storage(study_id)
+
+        self.read_trials_from_remote_storage(study_id)
 
         with self._lock:
             study = self._studies[study_id]
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -678,19 +678,6 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         """
         return self.get_trial(trial_id).system_attrs
 
-    def read_trials_from_remote_storage(self, study_id: int) -> None:
-        """Make an internal cache of trials up-to-date.
-
-        Args:
-            study_id:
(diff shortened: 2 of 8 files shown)


optuna/storages/_cached_storage.py
```diff
@@ -334,8 +334,8 @@ def get_all_trials(
         deepcopy: bool = True,
         states: Optional[Container[TrialState]] = None,
     ) -> List[FrozenTrial]:
-        if study_id not in self._studies:
-            self.read_trials_from_remote_storage(study_id)
+
+        self.read_trials_from_remote_storage(study_id)
```
keisuke-umezawa, line 338: It might be better to rename it as `_read_trials_from_remote_storage`.

HideakiImamura:

HideakiImamura (commented):
I am not confident that this function is erasable. This is because the `**Stronger consistency requirements for special data***` in the BaseStorage docstring describes the conventions that the storage must satisfy using this function. Do you have any explanation for this?

himkt:
I'm also not confident but I think RDBStorages already satisfy `Stronger consistency requirements`. I'm wondering if we can remove `read_from_remote_storage` from BaseStorage and move a document for stronger consistency requirements to CachedStorage.

PR #3442, opened 2022-04-04
Unify fail_stale_trials in each Storage impl

knshnb:
## Motivation

Currently, `fail_stale_trials` method is implemented in each Storage class (almost same functionality).

## Description of the changes

- added `_get_stale_trial_ids` method to `BaseStorage`
- unified `fail_stale_trials` and moved to `optuna/storages/_heartbeat.py`

I'm wondering if I should remove the first underscore of `_get_stale_trial_ids`.

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -730,6 +730,8 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         Args:
             study_id:
                 ID of the study.
+        Returns:
+            List of IDs of trials whose heartbeat has not been updated for a long time.
         """
         return []
 
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -724,20 +724,14 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         """
         pass
 
-    def fail_stale_trials(self, study_id: int) -> List[int]:
-        """Fail stale trials.
-
-        The running trials whose heartbeat has not been updated for a long time will be failed,
-        that is, those states will be changed to :obj:`~optuna.trial.TrialState.FAIL`.
-        The grace period is ``2 * heartbeat_interval``.
+    def _get_stale_trial_ids(self, study_id: int) -> List[int]:
+        """Get the stale trial ids of the study.
 
         Args:
             study_id:
-                ID of the related study.
-        Returns:
-            List of trial IDs of the failed trials.
+                ID of the study.
         """
-        pass
+        return []
 
     def is_heartbeat_enabled(self) -> bool:
         """Check whether the storage enables the heartbeat.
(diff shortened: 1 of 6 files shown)


HideakiImamura:
@not522 Could you review this PR if you have time?

not522 (commented):
Thank you for your PR!
Could you make `fail_stale_trials` of `RDBStorage` and `RedisStorage` deprecated instead of removal? Because they are public methods.
Note: `fail_stale_trials` of `BaseStorage` and `_CachedStorage` are private, therefore they can be removed.

knshnb:
Thanks for your comment! I reverted them with deprecated messages.

knshnb:
Thanks for your comment! I reverted them with deprecated messages.

g-votte:
@knshnb 
Thanks for your PR!

@not522 
I'm a little worried about leaving the unused method with the standard `@deprecated` policy.

- Though both `RDBStorage` and `RedisStorage` are public, we don't assume that the user directly call `fail_stale_trials` as written [here](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.storages.RDBStorage.html#optuna.storages.RDBStorage). Breaking the backward compatibility scarcely affects the library users.
- It would be harmful for maintainability. We'd have to take care of them for years, despite that the bloated storage classes are currently problematic.

> Thank you for your PR!
> Could you make fail_stale_trials of RDBStorage and RedisStorage deprecated instead of removal? Because they are public methods.

g-votte:
Hmm, since we have a precedent in #3323, we just may want to follow the standard `@deprecated` policy so far. We can just ignore [my previous comment](https://github.com/optuna/optuna/pull/3442#issuecomment-1089868760) at least in this PR, and discuss it later in another issue/PR.

HideakiImamura (approved):
Thanks for the PR. LGTM.

not522:
> I'm wondering if I should remove the first underscore of _get_stale_trial_ids.

That makes sense, but we need more discussion on the storage API design (#2943). I prefer keeping it private in this PR.

optuna/storages/_base.py
```diff
@@ -724,20 +724,14 @@ def record_heartbeat(self, trial_id: int) -> None:
...
+    def _get_stale_trial_ids(self, study_id: int) -> List[int]:
+        """Get the stale trial ids of the study.
 
         Args:
             study_id:
-                ID of the related study.
-        Returns:
-            List of trial IDs of the failed trials.
+                ID of the study.
         """
```
not522, line 733: Could you add the `Returns` section?

not522 (changes requested):

knshnb:
@not522 Thank you for the comment! PTAL. (Please ignore the merge commits, which I made by mistake...)

not522 (approved):
LGTM!

PR #3246, opened 2022-01-21
Enable `FloatDistribution` and `IntDistribution`

himkt:

Issue #2662, opened 2021-05-11
Inconsistent error handling in `suggest_int` and `suggest_float`


hvy:
## Expected behavior

More consistent error handling for `suggest_int` and `suggest_float`.

## Reproducible examples (optional)

```python
import optuna


# Suggest int. OK.
study = optuna.create_study()
trial = study.ask()
trial.suggest_int("x", 0, 10)
trial = study.ask() 
trial.suggest_int("x", 0, 10, step=2)  # OK.

# Suggest float. ValueError.
study = optuna.create_study()
trial = study.ask()
trial.suggest_float("x", 0, 10)
trial = study.ask()
trial.suggest_float("x", 0, 10, step=2)  # ValueError: Cannot set different distribution kind to the same parameter name.

# Note: Similar behavior reproducible without the second `study.ask()`s.
```

## Additional context (optional)

The inconsistency stem from historical reasons and in particular inconsistent distribution types and a consistency check based on these types https://github.com/optuna/optuna/blob/master/optuna/distributions.py#L521-L522. For instance, to suggest floats with steps you previously had to use `suggest_discrete_uniform`, similarly for log `suggest_loguniform`, until `suggest_float` was introduced which covers both cases. Internally, `suggest_float` simply calls `suggest_discrete_uniform`, etc. and hence the associated distribution types become different depending on arguments. At the same time, this interface is not symmetric for `suggest_int`. There is no `suggest_discrete_int` or `suggest_logint`. 

To properly fix this issue, we probably have to rethink what distribution types to provide and make sure that samplers and other components depending on these distributions are aware.

hvy:
This is still an issue.

hvy:
Let me reopen as this is still an issue.

himkt:
#3166 will fix the problem. (It has not been merged into the master branch yet)

contramundum53:
This issue has already been solved.

PR #3444, opened 2022-04-05
Fix distribution compatibility for linear and logarithmic distribution

himkt:
This PR fixes the regression of distribution compatibility checker.

The current `check_distribution_compatibility` treats a pair of the same type distribution compatible whose log configuration are different (e.g. `FloatDistribution(low=1, high=10, log=False)` and `FloatDistribution(low=1, high=10, log=True)`. This change was introduced in https://github.com/optuna/optuna/pull/3063.

This means sampler could observes a trial whose value doesn't exist in the search space. For example, consider the case that suggest_float("x", -10, 10) returns -1 at trial 1. In trial 2, user changes a search space by invoking `suggest_float("x", 1, 10, log=True)`. A sampler observe the past values of x as `[-1]`, while `-1` is not contained in `FloatDistribution(1, 10, log=True)`.

(This works well when a sampler samples a parameter independent on the past trials (e.g. QMCSampler, RandomSampler))


himkt:
Note: the following test was expected to fail but succeeded.

https://github.com/optuna/optuna/blob/19f46cc2d36f97d48cf684c16da81b7bc49651b1/tests/integration_tests/test_chainermn.py#L301-L302

himkt:
https://github.com/optuna/optuna/pull/3246/files#diff-d4bd9efbd95eea0d1e5e945856370ca28fb15d54755c765d1a15030f91f25ea8L589-L590

himkt:
PTAL @HideakiImamura @not522. :pray:

tests/integration_tests/test_chainermn.py
```diff
@@ -298,9 +298,6 @@ def test_suggest_float(storage_mode: str, comm: CommunicatorBase) -> None:
 
                 assert x1 == x2
 
-                # Adding a parameter should not cause any errors.
-                mn_trial.suggest_float("x1", low1, high1, log=True)
-
```
not522, line 303: How about testing for an error? It's changed by introducing `FloatDistribution`. https://github.com/optuna/optuna/commit/d5904abb2a6431e79140e3f1fbc36adc419b6986#diff-c61e26fb5e8d21683854b515c6ed466e629b7ea3d155e3846147d0c7ab786825

tests/integration_tests/test_chainermn.py
```diff
@@ -312,9 +309,6 @@ def test_suggest_float(storage_mode: str, comm: CommunicatorBase) -> None:
                 x4 = mn_trial.suggest_float("x2", low2, high2, log=True)
                 assert x3 == x4
 
-                # Neither should changing.
-                mn_trial.suggest_float("x2", low2, high2)
-
```
not522, line 317: ditto.

not522 (commented):
Basically LGTM. Please check comments for tests.

optuna/distributions.py
```diff
@@ -619,6 +619,13 @@ def check_distribution_compatibility(
     if dist_old.__class__ != dist_new.__class__:
         raise ValueError("Cannot set different distribution kind to the same parameter name.")
 
+    if isinstance(dist_old, (FloatDistribution, IntDistribution)):
+        # For mypy.
+        assert isinstance(dist_new, (FloatDistribution, IntDistribution))
+
+        if dist_old.log != dist_new.log:
```
HideakiImamura, line 626: I think it should be checked the change like `step = None -> step != None` and vice versa for `FloatDistribution`.

HideakiImamura: Whether or not to make an error in this case (`step=None -> step!=None`) is debatable, but I think it should be an error in order to have the same specifications as v2.10.

himkt: Hmm, I think it's the intended change because we will make `suggest_float` to make consistent with `suggest_int` https://github.com/optuna/optuna/issues/2662. Please correct me if I misunderstand something. :pray:

tests/integration_tests/test_chainermn.py
himkt: Thank you for the careful review, you're absolutely right.
I add(restore) tests in 3a2a5e8.

tests/integration_tests/test_chainermn.py
himkt: https://github.com/optuna/optuna/pull/3444#discussion_r859310246

not522 (approved):
LGTM

optuna/distributions.py
HideakiImamura: I'm fine of this specification change. Let me resolve this conversation.

PR #3470, opened 2022-04-13
Fix speed benchmarks CI

not522:
## Motivation
Speed benchmarks CI failed due to `ModuleNotFoundError: No module named 'matplotlib'`.

https://github.com/optuna/optuna/actions/runs/2145391889

This is because the `asv` benchmark and the `bayesmark` benchmark are in the same directory. This PR makes a separate `asv` directory.

## Description of the changes
- make `benchmarks/asv` directory


himkt (approved):
LGTM. I confirmed the speed benchmark worked with this patch.
Let me merge with one approval. Thank you @not522!

PR #3475, opened 2022-04-18
Separate heartbeat functionality from `BaseStorage`

g-votte:
This PR separates the heartbeat-related methods from `BaseStorage`.

## Background
Though the [original responsibility of storage classes](https://github.com/optuna/optuna/blob/master/optuna/storages/_base.py#L29-L30) is to provide read-and-write methods of study/trial history, the current `BaseStorage` interface is too much bloated so that it includes heartbeat functionality.

Separating the heartbeat-related methods from `BaseStorage` reduces the responsibility of storages with the following advantages:
- simplify the implementation process of a storage subclass for a new backend
- increase the cohesion of existing storage classes, improving their readability and maintainability

I believe this change also promotes [the exposure of `BaseStorage` to library users](https://github.com/optuna/optuna/issues/2943).

## Approach
To minimize the amount of changes on the current storage subclasses, this PR just separates the interface of `BaseStorage`. Specifically:
- introduce the `BaseHeartbeat` class, which defines the interface of heartbeat, and move `record_heartbeat`, `get_heartbeat_interval`, `get_failed_trial_callback`, and `_get_stale_trial_ids` to this new class
- replace the superclass of current heartbeat-supported storages (i.e., `RDBStorage` and `RedisStorage`) with multiple inheritance of `BaseStorage` and `BaseHeartbeat`


g-votte:
I removed WIP of this PR. Please proceed to code review.

## Followup tasks
There would be several followup tasks. I'd like to work on them in other PRs to keep the changes being small batches.
- Move the heartbeat-related tests from `test_storages.py` to a new file like `test_heartbeat.py`
- Get rid of `Hearbeat` inheritance from `CachedStorage`, since heartbeat and cache should be independent responsibilities from each other 
- Get rid of `is_heartbeat_enabled` from `BaseStorage` to make `BaseStorage` totally independent from the heartbeat class

HideakiImamura:
@c-bata @keisuke-umezawa Could you review this PR if you have time?

optuna/storages/_heartbeat.py
```diff
@@ -25,13 +101,22 @@ def fail_stale_trials(study: "optuna.Study") -> None:
     if not storage.is_heartbeat_enabled():
         return
 
+    heartbeat = get_heartbeat(storage)
```
c-bata, line 104: Not sure but how about using TypeGuard here and removing `NullHeartbeat` class? The code is like below:

```python
try:
    from typing import TypeGuard
except ImportError:
    from typing_extensions import TypeGuard


def is_heartbeat_enabled(storage: BaseStorage)->TypeGuard[HeartbeatStorage]:
    if isinstance(storage, BaseHeartbeat) and self.get_heartbeat_interval() is not None:
        return True
    return False
```

```python
def fail_stale_trials(study: Study) -> None:
    storage = study._storage
    if not is_heart_beat_enabled(storage):
        return

    failed_trial_ids = []
    for trial_id in storage._get_stale_trial_ids(study._study_id):
        ...
```

Links:

* https://docs.python.org/ja/3/library/typing.html#typing.TypeGuard
* https://github.com/python/typing/tree/master/typing_extensions

c-bata (commented):
Changes are basically LGTM. But I personally do not prefer introducing `NullHeartbeat` class. Can we remove the class and use `typing.TypeGuard` instead?

HideakiImamura (commented):
Thanks for the PR. I have a nit comment. PTAL.

c-bata (approved):
Looks great to me!

keisuke-umezawa (commented):
We also need to remove `_is_heartbeat_supported` of all storages.
Other than that, lgtm!

HideakiImamura (approved):
LGTM!

g-votte:
@keisuke-umezawa @HideakiImamura 
I removed `_is_heartbeat_supported` from the concrete classes. PTAL.
https://github.com/optuna/optuna/pull/3475/commits/52a42fa4f30ee6967da1658ab4dc5e82617411f5

PR #3476, opened 2022-04-19
[DO NOT MERGE] Test reviewdog

himkt:
Sorry for raising the meaningless PR. It tests reviewdog workflow.

### DIFF
--- a/optuna/logging.py
+++ b/optuna/logging.py
@@ -34,7 +34,7 @@ def _get_library_name() -> str:
 
 def _get_library_root_logger() -> logging.Logger:
 
-    return logging.getLogger(_get_library_name())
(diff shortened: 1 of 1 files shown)


himkt:
@kmaehashi told me that `GITHUB_TOKEN` attached to a workflow triggered by PR events outside from `optuna/optuna` has only `read` permission.  

🔗 https://github.com/optuna/optuna/blob/15438dae5f58a7258dfd1e12d353287e1163e2bb/.github/workflows/reviewdog.yml#L4

We need to use `pull_request_target` to pass a token of `optuna/optuna` or create a dedicated personal token to make this workflow work.

🔗 https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

PR #3477, opened 2022-04-20
Revise docstring in MLFflow and WandB callbacks

nzw0301:
## Motivation

I suppose we have room to improve the docstring in the callback pages as introduced by this PR. In addition, [the description of the `MLflowCallback` docstring](https://optuna.readthedocs.io/en/latest/reference/generated/optuna.integration.MLflowCallback.html#optuna.integration.MLflowCallback) is broken. A note section is added between `Retuned Types` and `Raises`, as a result, we see two- `Parameters` sections. See the following screenshot:

Screenshot 2022-04-20 at 20 16 37

## Description of the changes

- The decorator's examples can move to the more exact place: `track_in_*` method section.
- We would like to use an informative example in `track_in_wandb`. To do so, this PR uses the same objective definition in the example as MLFLow's `track_in_mlflow` example.
- Fix the note section in mlflow. As a result, the note section is shown as follows:

Screenshot 2022-04-20 at 20 23 48

### DIFF
--- a/optuna/integration/wandb.py
+++ b/optuna/integration/wandb.py
@@ -73,7 +73,7 @@ class WeightsAndBiasesCallback(object):
             from optuna.integration.wandb import WeightsAndBiasesCallback
 
             wandb_kwargs = {"project": "my-project"}
-            wandbc = WeightsAndBiasesCallback(wandb_kwargs=wandb_kwargs)
+            wandbc = WeightsAndBiasesCallback(wandb_kwargs=wandb_kwargs, as_multirun=True)
 
 
             @wandbc.track_in_wandb()
@@ -82,7 +82,7 @@ class WeightsAndBiasesCallback(object):
                 return (x - 2) ** 2
 
 
-            study = optuna.create_study(as_multirun=True)
+            study = optuna.create_study()
             study.optimize(objective, n_trials=10, callbacks=[wandbc])
 
 
--- a/optuna/integration/wandb.py
+++ b/optuna/integration/wandb.py
@@ -86,29 +86,6 @@ class WeightsAndBiasesCallback(object):
             study.optimize(objective, n_trials=10, callbacks=[wandbc])
 
 
-        Add additional logging to Weights & Biases.
-
-        .. code::
-
-            import optuna
-            from optuna.integration.wandb import WeightsAndBiasesCallback
-            import wandb
-
-            wandb_kwargs = {"project": "my-project"}
-            wandbc = WeightsAndBiasesCallback(wandb_kwargs=wandb_kwargs)
-
-
-            @wandbc.track_in_wandb()
-            def objective(trial):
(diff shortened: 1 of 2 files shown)


toshihikoyanase:
@contramundum53 Could you review this PR, please?
It only update the layout of the reference, so I think you can merge it with a single approval.

contramundum53:
Thank you for your PR, it looks good to me!

contramundum53 (dismissed):

optuna/integration/wandb.py
```diff
@@ -212,6 +189,33 @@ def track_in_wandb(self) -> Callable:
...
+
+                wandb_kwargs = {"project": "my-project"}
+                wandbc = WeightsAndBiasesCallback(wandb_kwargs=wandb_kwargs)
+
+
+                @wandbc.track_in_wandb()
+                def objective(trial):
+                    x = trial.suggest_float("x", -10, 10)
+                    mlflow.log_param("power", 2)
+                    mlflow.log_metric("base of metric", x - 2)
```
contramundum53, line 210: Sorry I cannot test this code because I am not a user of wandb, but is it possible to use `mlflow.log_param` or `mlflow.log_metric` in `wandbc.track_in_wandb`? If so, doesn't this example need `import mlflow`?

toshihikoyanase: Good catch! I guess the `mlflow` example happened to contaminate the wandb example.
We can simply move the example in this file.
```
            @wandbc.track_in_wandb()
            def objective(trial):
                loss = (x - 2) ** 2
                wandb.log({"loss": loss})
                return loss
```

nzw0301: Thank you for pointing out this! I intended to use a similar example code to mlflow because the original example was not informative; `loss` is stored as an objective value and `loss` is overwritten by each`wandb.log` with WeightsAndBiasesCallback(multi_run=False), which is the default.


I revised the example as in https://github.com/optuna/optuna/pull/3477/commits/b9a0f576975d53e0ff40787bc9900248f36954ef.

optuna/integration/wandb.py
```diff
@@ -73,7 +73,7 @@ def objective(trial):
             from optuna.integration.wandb import WeightsAndBiasesCallback
 
             wandb_kwargs = {"project": "my-project"}
-            wandbc = WeightsAndBiasesCallback(wandb_kwargs=wandb_kwargs)
+            wandbc = WeightsAndBiasesCallback(wandb_kwargs=wandb_kwargs, as_multirun=True)
```
nzw0301, line 76: Note suggested in https://github.com/optuna/optuna/pull/3513.

optuna/integration/wandb.py
contramundum53: Thanks, LGTM!

contramundum53 (approved):

PR #3270, opened 2022-01-30
Convert all positional arguments to keyword-only

higucheese:
## Motivation

This PR is based on https://github.com/optuna/optuna/issues/2911, but simply changing the order of positional arguments of `create_study` or `load_study` can confuse users. So this aims to encourage them to use keyword-only arguments first without any changes to the interface. Once users get used to keyword-only arguments, it gets much easier to align their arguments.

## Description of the changes

- Change all positional arguments of `{create,load,delete,copy}_study()` to keyword-only arguments.
- When a caller of `{create,load,delete,copy}_study()` sets values as positional arguments, the decorator `_convert_positional_args` converts them to keyword arguments according to the signature of the decorated function and sets a warning.

PR #3490, opened 2022-04-22
Remove torch.distributed calls from TorchDistributedTrial properties

### LINKED ISSUE #3484: TorchDistributedTrial properties can make torch.distributed hang
### Expected behavior

Reading TorchDistributedTrial properties should have no side effects, in particular it should not call torch.distributed methods. 

```
# ...
if rank == 0:
    print(hp_search_trial.number)

torch.distributed.barrier()
# execution continues
```

### Environment

- Optuna version: 2.10.0
- Python version: 3.9.12
- OS: Linux-4.18.0-240.el8.x86_64-x86_64-with-glibc2.28
- (Optional) Other libraries and their versions: Pytorch 1.11

Processes hang after a trial property is read on one node only.
In the example above, reading the trial number on only one node causes a call to a torch.distributed method on one node only. The next distributed functions will hang because it expects the other nodes to call the same method first.

### Steps to reproduce

```py
import torch
import optuna
from torch import multiprocessing


def train(rank, world_size, trial):
    print(f"process rank {rank} started.")

    torch.distributed.init_process_group(
        backend="gloo",
        init_method="tcp://127.0.0.1:29400",
        world_size=world_size,
        rank=rank)

    print(f"process rank {rank} initialized.")

    trial = optuna.integration.TorchDistributedTrial(trial if rank == 0 else None)

    torch.distributed.barrier()
    print(f"process rank {rank} ready.")

    if rank == 0:
        print(trial.number)  # <<<------ rank 0 makes distributed op alone

    torch.distributed.barrier()  # <<<------ will hang here because rank 1 is late by one distributed op
    print(f"process rank {rank} exiting.")


def run_trial(trial):
    multiprocessing.spawn(train, (2, trial), nprocs=2)
    return 0


def main():
    study = optuna.create_study()
    study.optimize(run_trial, n_trials=1)


if __name__ == "__main__":
    main()
```

torch.distributed essentially works by keeping a history of called functions. For instance on two nodes, it may look like this:

| index | rank0 | rank1 |
|--|--|--|
| 0 | dist.all_reduce | dist.all_reduce |
| 1 | dist.all_reduce | dist.all_reduce |
| 2 | dist.broadcast |  |

In this example rank1 is a bit late so rank0 will hang before executing the next distributed operation.

Back to optuna, a common use-case in a torch.distributed setup is to log or save a checkpoint on rank 0:

```
if torch.distributed.get_rank() == 0:
    print(hp_search_trial.number)

torch.distributed.barrier()
# execution continues
```

Before this code, all nodes participating in the distributed environment have issued the same distributed functions calls.
The current implementation of `hp_search_trial.number` causes a call to torch.distributed.broadcast **but only on rank 0**.
Upon reaching the next distributed operation (`torch.distributed.barrier()` in this example), pytorch hangs on rank0 until rank1 executes a broadcast to mirror rank0 operations but this never happens.

Optuna documentation already states that trial methods use the distributed api and thus must be called on all nodes, but properties are not mentioned.
I think it is inconvenient and non-pythonic that reading properties has a side effect.

nlgranger:
Fixes #3484

## Motivation

Removes all calls to torch.distributed from TorchDistributedTrial properties as they have undesired side-effects: namely they break the synchronization of distributed processes if one forgets to call them simultaneously on all distributed processes.

## Description of the changes

Prior to this PR, reading properties on TorchDistributedTrial triggered a call to torch.distributed.broadcast in order to retrieve the values, this was necessary because they are solely stored on the process with rank 0.
This PR instead broadcasts the values to all processes each time they are updated by calls to suggest_uniform, etc. This is not inconvenient because these functions are already called on all distributed processes anyways and Optuna documentation is clear about this.

### DIFF
--- a/optuna/integration/pytorch_distributed.py
+++ b/optuna/integration/pytorch_distributed.py
@@ -43,7 +43,9 @@ def updates_properties(f: Callable[_P, _T]) -> Callable[_P, _T]:
 
     @functools.wraps(f)
     def wrapped(*args: _P.args, **kwargs: _P.kwargs) -> _T:
-        self: TorchDistributedTrial = args[0]
+        # TODO(nlgranger): Remove type ignore after mypy includes
+        # https://github.com/python/mypy/pull/12668
+        self: TorchDistributedTrial = args[0]  # type: ignore
 
         def state() -> Sequence:
             assert self._delegate is not None
--- a/optuna/integration/pytorch_distributed.py
+++ b/optuna/integration/pytorch_distributed.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+import functools
 import pickle
 from typing import Any
 from typing import Callable
@@ -22,6 +23,35 @@ with try_import() as _imports:
 _suggest_deprecated_msg = (
     "Use :func:`~optuna.integration.TorchDistributedTrial.suggest_float` instead."
 )
+
+
+def updates_state(f: Callable) -> Any:
+    @functools.wraps(f)
+    def wrapped(self, *args, **kwargs) -> Any:
+        def state() -> Sequence:
+            assert self._delegate is not None
+            return (
+                self._delegate.number,
+                self._delegate.params,
+                self._delegate.distributions,
+                self._delegate.user_attrs,
+                self._delegate.system_attrs,
+                self._delegate.datetime_start,
+            )
+
+        try:
+            return f(self, *args, **kwargs)
+        finally:
+            (
+                self._number,
+                self._params,
+                self._distributions,
+                self._user_attrs,
+                self._system_attrs,
+                self._datetime_start,
+            ) = self._call_and_communicate_obj(state)
+
+    return wrapped
 
 
 @experimental("2.6.0")
@@ -73,7 +103,16 @@ class TorchDistributedTrial(optuna.trial.BaseTrial):
             assert trial is None, "error message"
         self._delegate = trial
         self._device = device
+        self._params = {}
 
+        self._number = self._broadcast(getattr(self._delegate, "number", None))
+        self._params = self._broadcast(getattr(self._delegate, "params", None))
+        self._distributions = self._broadcast(getattr(self._delegate, "distributions", None))
+        self._user_attrs = self._broadcast(getattr(self._delegate, "user_attrs", None))
+        self._system_attrs = self._broadcast(getattr(self._delegate, "system_attrs", None))
+        self._datetime_start = self._broadcast(getattr(self._delegate, "datetime_start", None))
+
+    @updates_state
     def suggest_float(
         self,
         name: str,
@@ -91,20 +130,24 @@ class TorchDistributedTrial(optuna.trial.BaseTrial):
         return self._call_and_communicate(func, torch.float)
 
     @deprecated("3.0.0", "6.0.0", text=_suggest_deprecated_msg)
+    @updates_state
     def suggest_uniform(self, name: str, low: float, high: float) -> float:
 
         return self.suggest_float(name, low, high)
 
     @deprecated("3.0.0", "6.0.0", text=_suggest_deprecated_msg)
+    @updates_state
     def suggest_loguniform(self, name: str, low: float, high: float) -> float:
 
         return self.suggest_float(name, low, high, log=True)
 
     @deprecated("3.0.0", "6.0.0", text=_suggest_deprecated_msg)
+    @updates_state
     def suggest_discrete_uniform(self, name: str, low: float, high: float, q: float) -> float:
 
         return self.suggest_float(name, low, high, step=q)
 
+    @updates_state
     def suggest_int(self, name: str, low: int, high: int, step: int = 1, log: bool = False) -> int:
         def func() -> float:
 
@@ -113,6 +156,7 @@ class TorchDistributedTrial(optuna.trial.BaseTrial):
 
         return self._call_and_communicate(func, torch.int)
 
+    @updates_state
     def suggest_categorical(self, name: str, choices: Sequence["CategoricalChoiceType"]) -> Any:
         def func() -> CategoricalChoiceType:
 
@@ -121,6 +165,7 @@ class TorchDistributedTrial(optuna.trial.BaseTrial):
 
         return self._call_and_communicate_obj(func)
 
+    @updates_state
     def report(self, value: float, step: int) -> None:
         err = None
         if dist.get_rank() == 0:
@@ -136,6 +181,7 @@ class TorchDistributedTrial(optuna.trial.BaseTrial):
         if err is not None:
             raise err
 
+    @updates_state
     def should_prune(self) -> bool:
         def func() -> bool:
 
@@ -147,6 +193,7 @@ class TorchDistributedTrial(optuna.trial.BaseTrial):
         # due to the RuntimeError.
         return self._call_and_communicate(func, torch.uint8)
 
+    @updates_state
     def set_user_attr(self, key: str, value: Any) -> None:
         err = None
         if dist.get_rank() == 0:
@@ -162,6 +209,7 @@ class TorchDistributedTrial(optuna.trial.BaseTrial):
         if err is not None:
             raise err
 
+    @updates_state
     def set_system_attr(self, key: str, value: Any) -> None:
         err = None
 
@@ -180,57 +228,27 @@ class TorchDistributedTrial(optuna.trial.BaseTrial):
 
     @property
     def number(self) -> int:
-        def func() -> int:
-
-            assert self._delegate is not None
-            return self._delegate.number
-
-        return self._call_and_communicate(func, torch.int)
+        return self._number
 
     @property
     def params(self) -> Dict[str, Any]:
-        def func() -> Dict[str, Any]:
-
-            assert self._delegate is not None
-            return self._delegate.params
-
-        return self._call_and_communicate_obj(func)
+        return self._params
 
     @property
     def distributions(self) -> Dict[str, BaseDistribution]:
-        def func() -> Dict[str, BaseDistribution]:
-
-            assert self._delegate is not None
-            return self._delegate.distributions
-
-        return self._call_and_communicate_obj(func)
+        return self._distributions
 
     @property
     def user_attrs(self) -> Dict[str, Any]:
-        def func() -> Dict[str, Any]:
-
-            assert self._delegate is not None
-            return self._delegate.user_attrs
-
-        return self._call_and_communicate_obj(func)
+        return self._user_attrs
 
     @property
     def system_attrs(self) -> Dict[str, Any]:
-        def func() -> Dict[str, Any]:
-
-            assert self._delegate is not None
-            return self._delegate.system_attrs
-
-        return self._call_and_communicate_obj(func)
+        return self._system_attrs
 
     @property
     def datetime_start(self) -> Optional[datetime]:
-        def func() -> Optional[datetime]:
-
-            assert self._delegate is not None
-            return self._delegate.datetime_start
-
-        return self._call_and_communicate_obj(func)
+        return self._datetime_start
 
     def _call_and_communicate(self, func: Callable, dtype: "torch.dtype") -> Any:
         buffer = torch.empty(1, dtype=dtype)
--- a/optuna/integration/pytorch_distributed.py
+++ b/optuna/integration/pytorch_distributed.py
@@ -30,7 +30,8 @@ def updates_properties(f: Callable[..., Any]) -> Callable[..., Any]:
 
     This decorator ensures trial properties (params, distributions, etc.) on all distributed
     processes are up-to-date with the wrapped trial stored on rank 0.
-    It should be applied to all :class:`~optuna.integration.TorchDistributedTrial` methods that update property values.
+    It should be applied to all :class:`~optuna.integration.TorchDistributedTrial`
+    methods that update property values.
     """
 
     @functools.wraps(f)
--- a/optuna/integration/pytorch_distributed.py
+++ b/optuna/integration/pytorch_distributed.py
@@ -6,6 +6,9 @@ from typing import Callable
 from typing import Dict
 from typing import Optional
 from typing import Sequence
+from typing import TypeVar
+
+from typing_extensions import ParamSpec
 
 import optuna
 from optuna._deprecated import deprecated
@@ -20,12 +23,16 @@ with try_import() as _imports:
     import torch.distributed as dist
 
 
+_T = TypeVar("_T")
+_P = ParamSpec("_P")
+
+
 _suggest_deprecated_msg = (
     "Use :func:`~optuna.integration.TorchDistributedTrial.suggest_float` instead."
 )
 
 
-def updates_properties(f: Callable[..., Any]) -> Callable[..., Any]:
+def updates_properties(f: Callable[_P, _T]) -> Callable[_P, _T]:
     """Method decorator to fetch updated trial properties from rank 0 after ``f`` is run.
 
     This decorator ensures trial properties (params, distributions, etc.) on all distributed
@@ -35,7 +42,7 @@ def updates_properties(f: Callable[..., Any]) -> Callable[..., Any]:
     """
 
     @functools.wraps(f)
-    def wrapped(*args: Any, **kwargs: Any) -> Any:
+    def wrapped(*args: _P.args, **kwargs: _P.kwargs) -> _T:
         self: TorchDistributedTrial = args[0]
 
         def state() -> Sequence:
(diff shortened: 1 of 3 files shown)


nlgranger:
Does this need a test? If so I'd like help on what should be tested.

toshihikoyanase:
> Does this need a test? 

Since this PR adds new functionality, I guess we need tests.
I'd like to check the following two points.

1. The rank-0 node gets property values without hanging processes.
2. The property values are properly updated after changing the values.

I'm think they can confirmed that by adding a following assertions. What do you think of it?

```
@pytest.mark.filterwarnings("ignore::optuna.exceptions.ExperimentalWarning")
def test_params(storage_mode: str) -> None:
        if dist.get_rank() == 0:
            trial = TorchDistributedTrial(study.ask())
        else:
            trial = TorchDistributedTrial(None)

        trial.suggest_float("f", 0, 1)
        trial.suggest_int("i", 0, 1)
        trial.suggest_categorical("c", ("a", "b", "c"))

        params = trial.params
        assert 0 <= params["f"] <= 1
        assert 0 <= params["i"] <= 1
        assert params["c"] in {"a", "b", "c"}
+
+        if dist.get_rank() == 0:
+            assert 0 <= trial.params["f"] <= 1
+            assert 0 <= params["i"] <= 1
+            assert params["c"] in {"a", "b", "c"}
```

toshihikoyanase:
In addition, the change will increase communication to update the property values.
Could you briefly check the difference of processing time of functions with the `@updates_state` annotation?

nlgranger:
> 1. The rank-0 node gets property values without hanging processes.
> 2. The property values are properly updated after changing the values.

I have added a new test to check that. It loops over all properties so that it doesn't miss one by accident in the future.
Since it checks for deadlocks, I have added a timeout to the test which in turns requires the pytest-timeout package dependency.



> Could you briefly check the difference of processing time of functions with the `@updates_state` annotation?

I have run the following code on my computer (CPU only experiment with gloo distributed backend)

```py
import os
import tempfile
import time

from torch import distributed as dist

import optuna
from optuna.integration import TorchDistributedTrial


def main():
    os.environ["WORLD_SIZE"] = os.environ["OMPI_COMM_WORLD_SIZE"]
    os.environ["RANK"] = os.environ["OMPI_COMM_WORLD_RANK"]
    os.environ["MASTER_ADDR"] = "127.0.0.1"
    os.environ["MASTER_PORT"] = "20000"

    dist.init_process_group("gloo")

    tempdir = tempfile.NamedTemporaryFile()
    url = "sqlite:///{}".format(tempdir.name)
        url,
        engine_kwargs={"connect_args": {"timeout": 10}}
    )

    if dist.get_rank() == 0:
        trial = TorchDistributedTrial(study.ask())
    else:
        trial = TorchDistributedTrial(None)

    dist.barrier()

    t1 = time.monotonic()

    for i in range(100):
        trial.suggest_float(str(i), 0., 1.)

    dist.barrier()

    t2 = time.monotonic()

    if dist.get_rank() == 0:
        print(t2 - t1)


if __name__ == "__main__":
    main()
```

The timing values are fairly consistent and the update_states causes an an overhead of 23% (0.66s -> 0.83s). Although significant, this is absolutely negligible in a typical PyTorch workflow.

nlgranger:
I'm gonna need help with mypy checks, I'm no expert in python typing hints.

optuna/integration/pytorch_distributed.py
```diff
@@ -24,6 +25,42 @@
...
+                self._delegate.number,
+                self._delegate.params,
+                self._delegate.distributions,
+                self._delegate.user_attrs,
+                self._delegate.system_attrs,
+                self._delegate.datetime_start,
+            )
+
+        try:
+            return f(self, *args, **kwargs)
```
toshihikoyanase, line 50: Mypy showed an error due to missing type annotation for the arguments.
Not sure, but the optuna code corresponding to this `wrapped` function takes `args` and `kwargs`, so I guess we can refactor this function as follows:

```suggestion
    def wrapped(*args: Any, **kwargs: Any) -> Any:
        def state() -> Sequence:
            self: TorchDistributedTrial = args[0]
            assert self._delegate is not None
            return (
                self._delegate.number,
                self._delegate.params,
                self._delegate.distributions,
                self._delegate.user_attrs,
                self._delegate.system_attrs,
                self._delegate.datetime_start,
            )

        try:
            return f(*args, **kwargs)
```

In addition, I guess we can use `properties` instead of `state`. What do you think of it?

optuna/integration/pytorch_distributed.py
```diff
@@ -24,6 +25,42 @@
 )
 
 
+def updates_state(f: Callable[..., Any]) -> Callable[..., Any]:
+    """Method decorator to fetch updated trial state from rank 0 after f is run.
```
toshihikoyanase, line 29: A nit.
```suggestion
    """Method decorator to fetch updated trial properties from rank 0 after ``f`` is run.
```

optuna/integration/pytorch_distributed.py
```diff
@@ -24,6 +25,42 @@
...
+
+    This decorator ensures trial properties (params, distributions, etc.) on all distributed
+    processes are up-to-date with the wrapped trial stored on rank 0.
+    It should be applied to all Trial methods that update property values.
```
toshihikoyanase, line 33:
```suggestion
    It should be applied to all :class:`~optuna.integration.TorchDistributedTrial` methods that update property values.
```

optuna/integration/pytorch_distributed.py
```diff
@@ -73,7 +110,16 @@ def __init__(
             assert trial is None, "error message"
         self._delegate = trial
         self._device = device
+        self._params = {}
```
toshihikoyanase, line 113: Can we remove this line?
```suggestion
```

tests/integration_tests/test_pytorch_distributed.py
```diff
@@ -334,3 +334,38 @@ def test_distributions(storage_mode: str) -> None:
         assert distributions["i"] == optuna.distributions.IntDistribution(0, 1)
         assert distributions["il"] == optuna.distributions.IntDistribution(1, 128, log=True)
         assert distributions["c"] == optuna.distributions.CategoricalDistribution(("a", "b", "c"))
+
+
+@pytest.mark.filterwarnings("ignore::optuna.exceptions.ExperimentalWarning")
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES)
+@pytest.mark.timeout(15, method="signal")
```
toshihikoyanase, line 341: In my environment, `pytest-timeout` did not terminate the MPI processes as expected.
Could you check if it terminate the processes if you add following lines to `test_updates_state`?

```
       if dist.get_rank() == 0:
            trial.suggest_float("f", 0, 1)
```

tests/integration_tests/test_pytorch_distributed.py
```diff
@@ -334,3 +334,38 @@ def test_distributions(storage_mode: str) -> None:
...
+
+        property_names = [
+            p
+            for p in dir(TorchDistributedTrial)
+            if isinstance(getattr(TorchDistributedTrial, p), property)
+        ]
+
+        # rank 0 can read properties without messing up distributed state
+        if dist.get_rank() == 0:
+            [getattr(trial, p) for p in property_names]
```
toshihikoyanase, line 363: [Note] This test does not confirms that the values of properties are correct. But it is reasonable since the value of properties are checked in the other test cases such as `test_params`.

tests/integration_tests/test_pytorch_distributed.py
```diff
@@ -334,3 +334,38 @@ def test_distributions(storage_mode: str) -> None:
...
+        trial.suggest_int("i", 0, 1)
+        trial.suggest_categorical("c", ("a", "b", "c"))
+
+        property_names = [
+            p
+            for p in dir(TorchDistributedTrial)
+            if isinstance(getattr(TorchDistributedTrial, p), property)
+        ]
+
+        # rank 0 can read properties without messing up distributed state
```
toshihikoyanase, line 361: Cosmetic fix to keep the [coding style](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#block-and-inline-comments-are-supposed-to-be-sentences). In addition, could you clarify what does the `messing up distributed state` mean? The following is just an example.
```suggestion
        # Rank 0 can read properties without deadlock.
```

tests/integration_tests/test_pytorch_distributed.py
```diff
@@ -334,3 +334,38 @@ def test_distributions(storage_mode: str) -> None:
...
+        if dist.get_rank() == 0:
+            [getattr(trial, p) for p in property_names]
+
+        dist.barrier()
+
+        # same with rank 1
```
toshihikoyanase, line 367: ditto. Just a cosmetic fix.
```suggestion
        # Same with rank 1.
```

toshihikoyanase (commented):
Thank you for your swift action!

> I'm gonna need help with mypy checks, I'm no expert in python typing hints.

I add suggestion to fix type hints.

> The timing values are fairly consistent and the update_states causes an an overhead of 23% (0.66s -> 0.83s). Although significant, this is absolutely negligible in a typical PyTorch workflow.

Thank you for sharing the speed benchmark. This change will add 1.7 msec per method call, and I think it is acceptable as you mentioned.

> Since it checks for deadlocks, I have added a timeout to the test which in turns requires the pytest-timeout package dependency.

`pytest-timeout` may not work with mpi jobs. Please check my comments below.

tests/integration_tests/test_pytorch_distributed.py
toshihikoyanase: The test finished after 30 minutes. The `torch.distributed` has the [timeout functionality](https://pytorch.org/docs/stable/distributed.html#torch.distributed.init_process_group), and we do not need to introduce `pytest-timeout`.

```console
tests/integration_tests/test_pytorch_distributed.py:347: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
optuna/_experimental.py:90: in wrapped_init
    _original_init(self, *args, **kwargs)
optuna/integration/pytorch_distributed.py:116: in __init__
    self._number = self._broadcast(getattr(self._delegate, "number", None))
optuna/integration/pytorch_distributed.py:286: in _broadcast
    dist.broadcast(size_buffer, src=0)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tensor = tensor([5], dtype=torch.int32), src = 0, group = None, async_op = False

    def broadcast(tensor, src, group=None, async_op=False):
        """
        Broadcasts the tensor to the whole group.
    
        ``tensor`` must have the same number of elements in all processes
        participating in the collective.
    
        Args:
            tensor (Tensor): Data to be sent if ``src`` is the rank of current
                process, and tensor to be used to save received data otherwise.
            src (int): Source rank.
            group (ProcessGroup, optional): The process group to work on. If None,
                the default process group will be used.
            async_op (bool, optional): Whether this op should be an async op
    
        Returns:
            Async work handle, if async_op is set to True.
            None, if not async_op or if not part of the group
    
        """
        _check_single_tensor(tensor, "tensor")
        if _rank_not_in_group(group):
            return
    
        opts = BroadcastOptions()
        opts.rootRank = src
        opts.rootTensor = 0
    
        if group is None or group is GroupMember.WORLD:
            default_pg = _get_default_group()
            work = default_pg.broadcast([tensor], opts)
        else:
            group_src_rank = _get_group_rank(group, src)
            opts.rootRank = group_src_rank
            work = group.broadcast([tensor], opts)
        if async_op:
            return work
        else:
>           work.wait()
E           RuntimeError: Application timeout caused pair closure

.venv/lib/python3.8/site-packages/torch/distributed/distributed_c10d.py:1167: RuntimeError
```

tests/integration_tests/test_pytorch_distributed.py
nlgranger: I mean the state of the distributed context remains synchronized between processes. ie: without deadlock indeed.

tests/integration_tests/test_pytorch_distributed.py
nlgranger: Fixed by simply lowering the distributed process group default timeout.

nlgranger:
Hi. Is there anything left to resolve?

optuna/integration/pytorch_distributed.py
```diff
@@ -24,6 +25,45 @@
 )
 
 
+def updates_properties(f: Callable[..., Any]) -> Callable[..., Any]:
```
toshihikoyanase, line 28: @himkt mentioned IDEs cannot infer argument types of the annotated methods. Could you use `ParamSpec` as can be seen in https://github.com/optuna/optuna/pull/3270?

toshihikoyanase (commented):
@nlgranger Thank you for your kind ping. I just came back from vacation.

@HideakiImamura The change almost looks great to me except for the type inference. Could you assign another reviewer, please?

nlgranger:
The mypy test will pass in the next release (https://github.com/python/mypy/pull/12668), I think I'd rather wait than find a workaround if that's ok for you.

optuna/integration/pytorch_distributed.py
```diff
@@ -19,11 +23,54 @@
...
+
+    This decorator ensures trial properties (params, distributions, etc.) on all distributed
+    processes are up-to-date with the wrapped trial stored on rank 0.
+    It should be applied to all :class:`~optuna.integration.TorchDistributedTrial`
+    methods that update property values.
+    """
+
+    @functools.wraps(f)
+    def wrapped(*args: _P.args, **kwargs: _P.kwargs) -> _T:
+        self: TorchDistributedTrial = args[0]
```
toshihikoyanase, line 46: Thank you for your update. Could you ignore the type checking tentitatively?
```suggestion
        # TODO(nlgranger): Remove type ignore after mypy includes
        # https://github.com/python/mypy/pull/12668
        self: TorchDistributedTrial = args[0]  # type: ignore
```

toshihikoyanase (approved):
Thanks for your update!
The suggestion successfully worked in my VS Code.
image

I'm sorry but it took some time to notice that `@experimental` prevents the type inference as well.

LGTM!

nzw0301:
I'll review this PR until the end of next week.

optuna/integration/pytorch_distributed.py
```diff
@@ -91,20 +148,24 @@ def func() -> float:
         return self._call_and_communicate(func, torch.float)
 
     @deprecated("3.0.0", "6.0.0", text=_suggest_deprecated_msg)
+    @updates_properties
```
nzw0301, line 151: Since this method calls `suggest_float` that is decorated by `updates_properties`, can we remove `updates_properties`?
```suggestion
```

I suppose the other deprecated `suggest_*` methods too.

nzw0301 (changes requested):
I locally confirmed that the properties were shared correctly among nodes by using `pytest`.  Basically, LGTM. 

I have one comment. Could you take a look at my comment?

optuna/integration/pytorch_distributed.py
nlgranger: Yes it is possible but I though if all methods are decorated one can change the implementation without having to remember about adding it.

nzw0301: Thanks. Hmm, but in my understanding, when we call one of these deprecated suggest methods, `updates_properties` is executed twice, which is supposed to be unnecessary communications among processes. 

@toshihikoyanase  What do you think about this?

toshihikoyanase: Good catch! `suggest_uniform` simply calls `suggest_float` and `suggest_float` actually update the properties. So, @nzw0301  suggestion sounds reasonable to me.

> one can change the implementation without having to remember about adding it.

That makes sense, but we will not update the deprecated methods so frequently.

nlgranger: I have removed the redundant decorators in the latest commit.

optuna/integration/pytorch_distributed.py
```diff
@@ -19,11 +23,56 @@
...
+    methods that update property values.
+    """
+
+    @functools.wraps(f)
+    def wrapped(*args: _P.args, **kwargs: _P.kwargs) -> _T:
+        # TODO(nlgranger): Remove type ignore after mypy includes
+        # https://github.com/python/mypy/pull/12668
+        self: TorchDistributedTrial = args[0]  # type: ignore
+
+        def state() -> Sequence:
```
nzw0301, line 50: How about replacing `state` with `fetch_properties` or something else because `state` sounds a little bit vague? 
```suggestion
        def fetch_properties() -> Sequence:
```

nzw0301 (changes requested):
Thanks. I have one more comment.

optuna/integration/pytorch_distributed.py
nlgranger: I updated the decorator name to match as well.

nzw0301: thanks!

nzw0301 (approved):
LGTM. Thanks!

nlgranger:
Thank you all for the review work!

PR #3493, opened 2022-04-24
Fix label of issue template

himkt:
Issue template for documentation was introduced in https://github.com/optuna/optuna/pull/3169.
`labels` specification is not correct, i.e. `labels` is `documentation` but the label in our GitHub issue is `document`.


HideakiImamura (approved):
Nice catch! LGTM.

toshihikoyanase (approved):
Good catch! I confirmed that the label is actually `document`.

toshihikoyanase:
The document label is assigned by default now. Thank you!
image

PR #3501, opened 2022-04-24
Fix torch installation

keisuke-umezawa:
## Motivation

To fix torch installation for mac.

## Description of the changes

Fix setup.py


nzw0301:
In my understanding, we can use the latest torch to resolve https://github.com/optuna/optuna/issues/2981, so I'm not sure that this temporal fix is necessary.

himkt:
@keisuke-umezawa will investigate the installation error on his environment.
I agree with @nzw0301 that we can use the latest version of PyTorch.

Note that I think it's nice to CPU version of PyTorch on document build as well.

himkt:
Sorry for the vague comment.

I think we now install Torch (+CUDA support).

- https://github.com/optuna/optuna/pull/3501/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L88
- https://github.com/optuna/optuna/runs/6186087410?check_suite_focus=true#step:5:46

And I think we can use CPU version instead.

nzw0301:
Thank you for your clarification. #3510 will use only the CPU version. So is it enough?

himkt:
Brilliant, I think it solves the problem!

nzw0301:
As discussed with @himkt in this PR, #3510 will resolve this issue. Let me close this PR once.

PR #3506, opened 2022-04-24
Add constraints option to TPESampler

knshnb:
## Motivation
I implemented a constrained optimization option in multi-objective `TPESampler` with the similar strategy as NSGA-II (#2175).
Interfaces are same as `NSGAIISampler`.

## Description of the changes
- Add constraints_func argument
- Make `_get_observation_pairs` also return constraints of trials
- When dividing into below and above in TPE, consider constraint values before HSSP

I'm considering if I should enable the constrained optimization in TPESampler, which can be naturally introduced when implementing constrained optimization in MOTPESampler.


himkt:
Could you please be a reviewer @HideakiImamura @toshihikoyanase?

knshnb:
I checked if it works in a toy example, which is modified a little from [Binh and Corn function with constraints](https://github.com/optuna/optuna-examples/blob/1b050c1cd37eadb7ea32216c5d1ba3a15c28b0cb/multi_objective/botorch_simple.py)

```python
from typing import Tuple

import optuna


def objective(trial: optuna.Trial) -> Tuple[float, float]:
    # Binh and Korn function with constraints.

    v0 = 4 * x**2 + 4 * y**2

    # Modified constraints function.
    trial.set_user_attr("constraint", (1000 - v0,))

    return v0, v1


def constraints(trial: optuna.Trial) -> Tuple[float]:


if __name__ == "__main__":
    samplers = {
        "nsgaii": optuna.samplers.NSGAIISampler,
        "motpe": optuna.samplers.TPESampler,
    }
    for use_const in [False, True]:
        for sampler_name in ["nsgaii", "motpe"]:
            sampler = samplers[sampler_name](
                seed=0, constraints_func=constraints if use_const else None
            )
            study = optuna.create_study(directions=["minimize", "minimize"], sampler=sampler)
            study.optimize(objective, n_trials=500)
            optuna.visualization.plot_pareto_front(
                study, target_names=["v0", "v1"], constraints_func=constraints
            ).write_image(f"result/{sampler_name}_{'constraints' if use_const else ''}_pareto.jpg")
```

knshnb:
As discussed privately with @HideakiImamura and @toshihikoyanase, I'll consider some other strategies of how to calculate weights for `_ParzenEstimator` by testing on a task with a smaller feasible region.

knshnb:
I tried another toy example that has a smaller feasible region, as suggested by @HideakiImamura.
```python
from typing import Tuple

import numpy as np

import optuna


def objective(trial: optuna.Trial) -> Tuple[float, float]:
    x = trial.suggest_float("x", 0, 6)
    y = trial.suggest_float("y", 0, 6)

    c = np.sin(x) * np.sin(y) + 0.95
    trial.set_user_attr("constraint", (c,))

    f1 = np.sin(x) - y
    f2 = np.cos(x) + y**2

    return f1, f2


def constraints(trial: optuna.Trial) -> Tuple[float]:


if __name__ == "__main__":
    samplers = {
        "nsgaii": optuna.samplers.NSGAIISampler,
        "motpe": optuna.samplers.TPESampler,
    }
    for use_const in [False, True]:
        for sampler_name in ["nsgaii", "motpe"]:
            sampler = samplers[sampler_name](
                seed=0, constraints_func=constraints if use_const else None
            )
            study = optuna.create_study(sampler=sampler, directions=["minimize", "minimize"])
            study.optimize(objective, n_trials=400)

            fig = optuna.visualization.plot_pareto_front(study, constraints_func=constraints)
            fig.write_image(
                f"result/small_feasible/{sampler_name}_{'constraints' if use_const else ''}_pareto.jpg"
            )
```

- MOTPE without constraints
image

- **MOTPE with constraints (this PR)**
image

- NSGA2 without constraints
image

- NSGA2 with constraints
image

The current implementation already works fine (seems slightly better than NSGA2 with constraints), but anyway I will try another strategy for the weight calculation.

knshnb:
I implemented another strategy: ignore all infeasible trials in hypervolume calculation for `ParzenEstimator`'s weight and set infeasible trials' weight `EPS`.

The experimental results are fine overall, although they are dependent on random seeds.
Seed 0:
image
Seed 1:
image
Seed 2:
image
Seed 3:
image
Seed 4:
image

knshnb:
As a side note, there is some room for improvement, such as
- using `EPS` as infeasible trials' weight might be too small
  - it depends on the scale of hypervolume, so we might need to consider normalization in output space
- we set the same weight for all infeasible trials in below, regardless of constraints values

However, I think the current implementation is enough as a straightforward extension of the method used in `NSGAIISampler`.

knshnb:
I supported `sample_independent` and tested on `multivariate=True` as well.
The results might be less dependent on random seeds when `multivatiate=True`.

Seed 0:
image
Seed 1:
image
Seed 2:
image
Seed 3:
image
Seed 4:
image

optuna/samplers/_tpe/sampler.py
```diff
@@ -635,13 +694,44 @@ def _get_observation_pairs(
                 param_value = None
             values[param_name].append(param_value)
 
-    return values, scores
+        if get_constraints:
+            assert constraints is not None
+            constraints.append(cast(Sequence[float], trial.system_attrs.get(_CONSTRAINTS_KEY)))
```
HideakiImamura, line 699: If `trial.system_attrs` misses the key of `_CONSTRAINTS_KEY`, the result of `cast(Sequence[float], trial.system_attrs.get(_CONSTRAINTS_KEY))` will be None. So, `constraints` can include `None`.

(1) The type of `constraints` should be `Optional[List[Optional[Sequence[float]]]]`.
(2) The handling of constraints after this function returns should be implemented with the consideration that there may be None inside. (I have not checked the handling after this function yet.)

optuna/samplers/_tpe/sampler.py
```diff
@@ -554,7 +607,12 @@ def _get_observation_pairs(
     param_names: List[str],
     multivariate: bool,
     constant_liar: bool = False,  # TODO(hvy): Remove default value and fix unit tests.
-) -> Tuple[Dict[str, List[Optional[float]]], List[Tuple[float, List[float]]]]:
+    get_constraints: bool = False,
```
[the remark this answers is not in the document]
HideakiImamura: On the other hand, do we need this argument? I think we can just return `List[Optional[Sequence[float]]]` always.

optuna/samplers/_tpe/sampler.py
```diff
@@ -635,13 +694,44 @@ def _get_observation_pairs(
...
+    constraints: Optional[List[Sequence[float]]],
 ) -> Tuple[np.ndarray, np.ndarray]:
+    # When constrains is not None, trials are split into below and above
+    # according to the following rules.
+    # 1. Feasible trials are better than infeasible trials.
+    # 2. Infeasible trials are sorted by sum of how much they violate each constraint.
+    # 3. Feasible trials are sorted by loss_vals.
+    if constraints is not None:
+        # 1-dimensional violation value (violation_1d==0 <=> feasible, >0 <=> infeasible).
+        violation_1d = np.maximum(np.array(constraints), 0).sum(1)
```
HideakiImamura, line 716: If `constraints` includes `None`, the type of each item in `np.array(constraints)` includes `None`. How about specifying the dtype like `np.array(constraints, dtype=float)`? It convert `None` to `np.nan`, so we need to filter `np.nan` before taking `np.maximum` to compare `violation_1d[idx[n_below]] > 0`.

HideakiImamura (commented):
Thanks for the great improvement of TPESampler! I have several comments. PTAL.

optuna/samplers/_tpe/sampler.py
```diff
@@ -554,7 +607,12 @@ def _get_observation_pairs(
     param_names: List[str],
     multivariate: bool,
     constant_liar: bool = False,  # TODO(hvy): Remove default value and fix unit tests.
-) -> Tuple[Dict[str, List[Optional[float]]], List[Tuple[float, List[float]]]]:
+    get_constraints: bool = False,
+) -> Tuple[
+    Dict[str, List[Optional[float]]],
+    List[Tuple[float, List[float]]],
+    Optional[List[Sequence[float]]],
```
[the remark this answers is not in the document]
knshnb: The return type of `constraints_func` is `Sequence[float]` following `NSGAIISampler`. Do you suggest to cast from `Sequence[float]` to `Tuple[float]` anywhere?
https://github.com/optuna/optuna/blob/0964da28442996931e9a6e967b586649bc357ce3/optuna/samplers/nsgaii/_sampler.py#L121

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
knshnb: I'm wondering if it is fine to return something like `[None, None, ...]`, which we do not use at all, when `constraints_func` is `None`.
But it might be probably better than the current implementation because the interface is consistent?

optuna/samplers/_tpe/sampler.py
knshnb: I did not consider the case `constraints` includes `None`. When `constraints` include `None`, it cannot be simply converted using `np.array` because `constraints` might be something like `[(1.0, 1.0), None, (2.0, 2.0)]` and the dimension does not match. I need to consider how to handle this.

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
knshnb: Also, I give the return value of this function to `_split_observation_pairs` and check if it is `None`. If we always return `List[Optional[Sequence[float]]]`, we might need to add an additional argument to `_split_observation_pairs` in order to specify if we use constrains.
https://github.com/optuna/optuna/blob/1e2bca1bd742935d9463206a30704f5b6c4bec96/optuna/samplers/_tpe/sampler.py#L714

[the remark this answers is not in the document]
knshnb: So, I think it's okay to return `Optional[List[Optional[Sequence[float]]]]` by this function, although a little bit complicated. What do you think? @HideakiImamura

HideakiImamura:
By the way, I think it is good idea to add some unit tests to validate the behavior.

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
HideakiImamura: Thanks for the explanation. It's OK to leave the type. Thanks.

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
HideakiImamura: Yes, I mean we can cast it.

optuna/samplers/_tpe/sampler.py
knshnb: I decided to handle `None` inside `_get_observation_pairs`. This makes `_get_observation_pairs` a bit more complex, but the whole logic and return type are simplified if we need to consider `None`. 
https://github.com/optuna/optuna/pull/3506/commits/2f1ff5a57ebe8f09013534cfe311978afc9ab14c#diff-2cc3a4088f95a10e806b68655d375f6f7f5fe2de161a0182c16168e378be8065R713-R715

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
knshnb: Now we do not need to cast by https://github.com/optuna/optuna/pull/3506/commits/2f1ff5a57ebe8f09013534cfe311978afc9ab14c.

optuna/samplers/_tpe/sampler.py
knshnb: Resolved by https://github.com/optuna/optuna/pull/3506/commits/2f1ff5a57ebe8f09013534cfe311978afc9ab14c.

optuna/samplers/_tpe/sampler.py
```diff
@@ -205,6 +207,17 @@ def objective(trial):
                 Added in v2.8.0 as an experimental feature. The interface may change in newer
                 versions without prior notice. See
                 https://github.com/optuna/optuna/releases/tag/v2.8.0.
+        constraints_func:
```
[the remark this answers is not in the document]
knshnb: I added a reference to v3.0.0-b1.
[Note] We need to explain this feature in the release note.

knshnb:
Thanks for a lot of useful comments! I addressed all the requests and will update tests after #3574 is merged.

knshnb:
@HideakiImamura @toshihikoyanase I finished updating the tests and now it's fully ready to review! PTAL🙏

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
HideakiImamura: Thanks for the update. We can just add a reference to v3.0.0 like https://github.com/optuna/optuna/blob/master/optuna/visualization/_pareto_front.py#L109

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
```diff
@@ -489,28 +489,45 @@ def test_multi_objective_sample_independent_ignored_states() -> None:
     assert len(set(suggestions)) == 1
 
 
-def test_multi_objective_get_observation_pairs() -> None:
+@pytest.mark.parametrize("constraints_enabled", [False, True])
+@pytest.mark.parametrize("constraint_value", [-2, 2])
+def test_multi_objective_get_observation_pairs(
```
HideakiImamura, line 494: I think we can just add another test function to improve the readability of unit tests. It slightly increase the code redundancy, but it can be accepted. What do you think?

knshnb: Umm, this is also not a strong opinion, but I think if there are multiple tests on the same function, the fixing cost is high when the interface of the original function changes. Which part of the current implementation do you think is hindering readability?
Also, do you have the same opinion on other unit tests on `test_sampler.py`?

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
knshnb: Thanks! I fixed the note according to that.

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
HideakiImamura: > Which part of the current implementation do you think is hindering readability?

My personal opinion is that constraint setup (e.g., [this one](https://github.com/optuna/optuna/pull/3506/files#diff-e52a24e2c14ee4ff0d8b5e62f38c4e515ec3dad12446570f0771ef74d27689faR502)) is required even for tests that have nothing to do with constrained optimization.

I think a test should be in a form that is immediately understandable in itself and should not have complex logic inside. For example, but what about cutting out the parts that can be commonly used by constrained optimization tests and existing tests as utilities and referencing them from separate test functions? (I know I said so myself, but this is going to be a major undertaking, so maybe we should tackle this in a separate PR.)

toshihikoyanase:
I have not fully checked yet, but the combination with `constant_liar` may cause `IndexError`

```
[I 2022-06-02 12:32:16,506] Trial 1 finished with values: [-1.251217679317449, 8.03118943229791] and parameters: {'x': -1.251217679317449, 'y': 8.03118943229791}. 
  File "pr3506.py", line 22, in <module>
    study.optimize(objective, n_trials=100, n_jobs=4)
  File "/home/yanase/pfn/code/optuna-review/optuna/study/study.py", line 405, in optimize
    _optimize(
  File "/home/yanase/pfn/code/optuna-review/optuna/study/_optimize.py", line 104, in _optimize
    f.result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 437, in result
    return self.__get_result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/yanase/pfn/code/optuna-review/optuna/study/_optimize.py", line 161, in _optimize_sequential
    frozen_trial = _run_trial(study, func, catch)
  File "/home/yanase/pfn/code/optuna-review/optuna/study/_optimize.py", line 190, in _run_trial
    trial = study.ask()
  File "/home/yanase/pfn/code/optuna-review/optuna/study/study.py", line 499, in ask
    trial = trial_module.Trial(self, trial_id)
  File "/home/yanase/pfn/code/optuna-review/optuna/trial/_trial.py", line 54, in __init__
    self._init_relative_params()
  File "/home/yanase/pfn/code/optuna-review/optuna/trial/_trial.py", line 63, in _init_relative_params
    self.relative_params = self.study.sampler.sample_relative(
  File "/home/yanase/pfn/code/optuna-review/optuna/samplers/_tpe/sampler.py", line 370, in sample_relative
    return self._sample_relative(study, trial, search_space)
  File "/home/yanase/pfn/code/optuna-review/optuna/samplers/_tpe/sampler.py", line 394, in _sample_relative
    indices_below, indices_above = _split_observation_pairs(scores, self._gamma(n), violations)
  File "/home/yanase/pfn/code/optuna-review/optuna/samplers/_tpe/sampler.py", line 734, in _split_observation_pairs
    if violation_1d[idx[n_below]] > 0:
IndexError: index 1 is out of bounds for axis 0 with size 1
```

Let me share the reproducible code.
```python
import random
import time

import optuna

def objective(trial):

    trial.set_user_attr("constraint", y)
    time.sleep(random.random())

    return x, y


def constraints_func(trial):
    return [trial.user_attrs["constraint"] - 5]


study = optuna.create_study(
    sampler=optuna.samplers.TPESampler(
        multivariate=True,
        constant_liar=True,
        constraints_func=constraints_func,
        n_startup_trials=1
    ),
    storage="sqlite:///pr3506.db",
    study_name="foo",
    load_if_exists=True
)
study.optimize(objective, n_trials=100, n_jobs=4)
for trial in study.best_trials:
    print(trial.values, trial.user_attrs["constraint"])
```

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -856,11 +889,30 @@ def test_split_observation_pairs() -> None:
...
+def test_split_observation_pairs_with_constraints() -> None:
+    indices_below, indices_above = _tpe.sampler._split_observation_pairs(
+        [
+            (-7, [-2]),  # PRUNED (with intermediate values)
+            (float("inf"), [0.0]),  # PRUNED (without intermediate values)
+            (
+                -3,
+                [float("inf")],
+            ),  # PRUNED (with a NaN intermediate value; it's treated as infinity)
+            (-float("inf"), [-5.0]),  # COMPLETE
```
HideakiImamura, line 907: The comments look strange. Could you update them as they are well-described?

HideakiImamura (commented):
Thanks for the update. I have several minor comments PTAL.

I have confirmed that the same problem reported by @toshihikoyanase occurs. Can you resolve the problem?

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
knshnb: I understand your point.

> what about cutting out the parts that can be commonly used by constrained optimization tests and existing tests as utilities and referencing them from separate test functions?

Then, could I work on the above suggestion after this PR is merged?

tests/samplers_tests/tpe_tests/test_sampler.py
knshnb: How are the comments strange? It's actually copy-pasted from the above function `test_split_observation_pairs()`.

knshnb:
@toshihikoyanase Thanks for pointing out the case that causes an error! It happens when `n_startup_trials` is 0 or 1 (not related to `constant_liar`). This was caused because I did not consider the case that `_split_observation_pars` is called for a very small number of trials (By default, `RandomSampler` is used in such cases).

I added a test for that case (https://github.com/optuna/optuna/pull/3506/commits/2290708c2a2ab86692d967ce88b428c34010c7dc) and fixed the code (https://github.com/optuna/optuna/pull/3506/commits/7a6ca0ba5ef00230e72fdb560a90641ee8d9041f).

@HideakiImamura Thanks for the comments! Please take a look again 🙏

toshihikoyanase:
@knshnb Thank you for your investigation! I confirmed that the current `TPESampler` works with `n_startup_trials=1`. I'll check the details of the code next!

optuna/samplers/_tpe/sampler.py
```diff
@@ -516,7 +550,29 @@ def after_trial(
         state: TrialState,
         values: Optional[Sequence[float]],
     ) -> None:
-
+        assert state in [TrialState.COMPLETE, TrialState.FAIL, TrialState.PRUNED]
```
toshihikoyanase, line 553: [Question] Could you tell me why this assertion is added? The following lines also checks the state in if-statement, and I guess it might not be necessary.

toshihikoyanase: Ah, it follows the implementation of NSGAIISampler.
https://github.com/optuna/optuna/blob/83c8022bcb3847b261e3f0b65aeaccf46dbb8cb6/optuna/samplers/nsgaii/_sampler.py#L391

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -856,11 +898,30 @@ def test_split_observation_pairs() -> None:
             (-float("inf"), [-5.0]),  # COMPLETE
         ],
         2,
+        None,
     )
     assert list(indices_below) == [0, 3]
     assert list(indices_above) == [1, 2]
 
 
+def test_split_observation_pairs_with_constraints() -> None:
```
HideakiImamura, line 907: Why not test both cases, ``if n_below >= len(idx) or violation_1d[idx[n_below]] > 0` or not then?

HideakiImamura (commented):

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
toshihikoyanase: I agree with @HideakiImamura . This test changes the expected values depending on the parametrized value (i.e., `constraints_enabled`), and I think it is a kind of [conditional test logic](http://xunitpatterns.com/Conditional%20Test%20Logic.html). As you mentioned, we can work on it in a follow-up PR.

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -753,27 +753,42 @@ def test_sample_independent_pruned_state() -> None:
     assert len(set(suggestions)) == 3
 
 
+def test_constrained_sample_independent_zero_startup() -> None:
+    """Tests TPESampler with constrained option works when n_startup_trials=0"""
```
toshihikoyanase, line 757: A nit.
```suggestion
    """Tests TPESampler with constrained option works when n_startup_trials=0."""
```

toshihikoyanase: Maybe this the function name explains this comment.

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -783,54 +798,80 @@ def objective(trial: Trial) -> float:
...
+    ) == (
         {"x": [5.0, 5.0, 5.0, 5.0]},
         scores,
+        violations,
     )
-    assert _tpe.sampler._get_observation_pairs(study, ["y"], True) == ({"y": []}, [])
-    assert _tpe.sampler._get_observation_pairs(study, ["z"], True) == (
+    assert _tpe.sampler._get_observation_pairs(
+        study, ["y"], True, constraints_enabled=constraints_enabled
+    ) == ({"y": []}, [], [] if constraints_enabled else None)
```
toshihikoyanase, line 832: ditto. This test function also changes the expected values depending on the parametrized value.

toshihikoyanase (commented):
The code basically looks good to me. Let me add some cosmetic comments.

optuna/samplers/_tpe/sampler.py
```diff
@@ -633,13 +699,54 @@ def _get_observation_pairs(
...
+            constraint = trial.system_attrs.get(_CONSTRAINTS_KEY)
+            if constraint is None:
+                warnings.warn(
+                    f"Trial {trial.number} does not have constraint values."
+                    " It will be treated as a lower priority than other trials."
+                )
+                violation = float("inf")
+            else:
+                # Violation values of infeasible dimensions are summed up.
+                violation = sum(map(lambda x: max(x, 0.0), constraint))
```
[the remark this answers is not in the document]
knshnb: Here, we use the sum of only positive constraint values (=infeasible axes), not all constraint values. For example, when `violation = 4`, we ignore feasible axes and `constraint = [-100, 1, 3]`.
Should I replace this with NSGAII's implementation, which might be easier to understand?
https://github.com/optuna/optuna/blob/83c8022bcb3847b261e3f0b65aeaccf46dbb8cb6/optuna/samplers/nsgaii/_sampler.py#L511

optuna/samplers/_tpe/sampler.py
knshnb: Yeah, I just followed NSGAIISampler without much thought. Do you think I should remove this line?

knshnb:
@HideakiImamura @toshihikoyanase I updated the code! PTAL.

optuna/samplers/_tpe/sampler.py
toshihikoyanase: I just wanted to know the reason. I think we can keep it. Thank you!

optuna/samplers/_tpe/sampler.py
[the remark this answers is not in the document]
knshnb: I updated the code to follow NSGAII's implementation.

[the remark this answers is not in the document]
HideakiImamura: Thanks for the update. I agree with the idea to align the NSGAII's implementation.

optuna/samplers/_tpe/sampler.py
```diff
@@ -633,13 +699,53 @@ def _get_observation_pairs(
...
+        violation_1d = np.array(violations, dtype=float)
+        idx = violation_1d.argsort()
+        if n_below >= len(idx) or violation_1d[idx[n_below]] > 0:
+            # Below is filled by all feasible trials and trials with smaller violation values.
+            indices_below = idx[:n_below]
+            indices_above = idx[n_below:]
+        else:
+            # All trials in below are feasible.
+            # Feasible trials with smaller loss_vals are selected.
+            (feasible_idx,) = (violation_1d == 0).nonzero()
```
HideakiImamura, line 739:
```suggestion
            (feasible_idx,) = (violation_1d <= 0).nonzero()
```

HideakiImamura (commented):
Thanks for the update. Almost LGTM. I have a nit comment. 

By the way, since we have slightly changed the implementation, could you take a benchmark again to confirm the no performance degradation?

knshnb:
Since there has been a large amount of code modification, I confirmed that the current code works fine on the toy problem https://github.com/optuna/optuna/pull/3506#issuecomment-1119418026.

seed=0
image
seed=1
image

optuna/samplers/_tpe/sampler.py
knshnb: As we privately discussed, I keep this line because `violation_1d` is guaranteed to be greater than or equal to 0.
https://github.com/optuna/optuna/pull/3506/files#diff-2cc3a4088f95a10e806b68655d375f6f7f5fe2de161a0182c16168e378be8065R712-R713

HideakiImamura (approved):
Thanks for the long running work! LGTM!

toshihikoyanase (approved):
LGTM. Thank you for accomplishing the long running PR!

[Notes] I checked the behavior of constrained MOTPE with the following simple code.

```python
import optuna

def objective(trial):
    trial.set_user_attr("constraint", [y - 5])
    return x, y

def constraints_func(trial):

study = optuna.create_study(
    directions=["maximize", "maximize"],
    sampler=optuna.samplers.TPESampler(
        multivariate=True,
        constraints_func=constraints_func,
    ),
)

optuna.visualization.plot_pareto_front(study, constraints_func=constraints_func).show()
optuna.visualization.plot_optimization_history(study, target=lambda t: t.values[0]).show()
optuna.visualization.plot_optimization_history(study, target=lambda t: t.values[1]).show()
```

The constrained MOTPE intensively search around `x=10, y=5` instead of `x=10, y=10` since `y>5` is infeasible.

image

By comparing optimization histories of `x` and `y`, we can see that the sampler gradually focused on `y=5` and I guess it can successfully choose the search points considering the constraint.

*Optimization history of `x`*
image

*Optimization history of `y`*
image

toshihikoyanase:
I found a minor bug of `TPESampler` during the review and reported in https://github.com/optuna/optuna/issues/3670.

okaikov:
is it possible to implement constraints for single objective TPE?  if so, are you planning to add this feature?

knshnb:
Actually, constrained optimization for single objective TPE itself is supported by this PR as well. However, we haven't developed the related features (such as log, visualization, etc.) and dogfooded well yet. Feel free to request if you encounter any problems!

```
import optuna


def objective(trial):
    trial.set_user_attr("constraint", (-x,))
    return x


def constraints_func(trial):


sampler = optuna.samplers.TPESampler(constraints_func=constraints_func)
```
Sorry, log of the best value does not take the constraint into account, but the optimization itself can work.

okaikov:
The reason I mentioned this is because I get the following warning on all trials when running single objective TPE with constraints:
optuna/samplers/_tpe/sampler.py:680: UserWarning: Trial 0 does not have constraint values. It will be treated as a lower priority than other trials.

knshnb:
Could you share the reproducible code?

okaikov:
this happens when constant_liar==True.

knshnb:
@okaikov Thanks for the report!
I created an [issue](https://github.com/optuna/optuna/issues/4221). It seems that the warning is a bug and you can ignore it for now.

okaikov:
Is it possible to provide a parameter to the constraints function? Example: _constraints_func(trial=trial, metrics=metrics)_
The use case is that I want to change the constraints during the study, so I don't want to set hardcoded values to the usr_attr and read it from there, instead I will compute the constraints dynamically.

knshnb:
Thanks for the comment! Could you create a new issue with a more detailed explanation of how you want to change the constraints during the study?

AntoineJarry:
Hi! I would like to understand exactly how constraints work for a single‐objective TPE. From what I’ve seen in the code, infeasible trials seem to be treated like other completed trials, and their weights do not appear to differ. Does that mean that, when I implement my own weighting function, I also need to explicitly handle infeasible trials to manage them correctly?

nabenabe0928:
@AntoineJarry 
Hey, although this doesn't answer your question and I don't know exactly why you would like to understand the detail, what about reading [the c-TPE paper](https://arxiv.org/abs/2211.14411)?

The implementation is available from the Optuna interface as well:)
https://hub.optuna.org/samplers/ctpe/

AntoineJarry:
@nabenabe0928 Thanks for your answer. I ask this because, this will affect the result of the optimization you know. And in the classical TPE Sampler the constraint_func is available for single objective function but I'm pretty sure she is not working as c-Tree sampler and even not really constraint the acquisition function.

nabenabe0928:
@AntoineJarry 
Ok, I actually didn't check your first comment well.

> From what I’ve seen in the code, infeasible trials seem to be treated like other completed trials.

Actually, we do treat infeasible trials differently.
Namely, feasible trials will fill up the trials_below first and infeasible trials might be appended to trials_below only if feasible trials are not sufficient to fill up trials_below.

> Their weights do not appear to differ. 

This is indeed correct.

> Does that mean that, when I implement my own weighting function, I also need to explicitly handle infeasible trials to manage them correctly?

Thus, you don't need to take care of the weighting func.

> in the classical TPE Sampler the constraint_func is available for single objective function but I'm pretty sure she is not working as c-Tree sampler

In fact, it does work as c-TPE partially.
They calculate feasible probability differently, but the primary concept is indeed identical.


> TPESampler even not really constraint the acquisition function.

I hope my explanation above helped you understand, but it does constrain the acquisition function by penalizing infeasible trials.

AntoineJarry:
@nabenabe0928 Thank you very much. I now understand and see the line that does this. So, infeasible trials go into the bad density. The thing that differs from c-TPE is that in c-TPE, we also multiply by the probability of satisfying the constraint, is that correct, or am I mistaken?

Issue #3448, opened 2022-04-06
Add Shapley importance evaluator


contramundum53:
### Motivation

It is reported that [using ANOVA for evaluating importance of parameters has conceptual problems when those parameters are dependent, and that Shapley values provide an elegant solution to it.](https://arxiv.org/pdf/1610.02080.pdf) We could provide Shapley importance evaluator as an alternative to `FanovaImportanceEvaluator` and possibly make that the default importance evaluator in the future.

### Description

* Add `ShapleyImportanceEvaluator` class deriving from `BaseImportanceEvaluator` that computes the Shapley values of fitted random forest.
  * [This repo](https://github.com/slundberg/shap) seems to provide a fast and exact algorithm for computing Shapley values for tree-ensemble models.
* Possibly make this the default importance evaluator (which is currently `FanovaImportanceEvaluator`)

liaison:
hey @contramundum53 this seems to be an interesting task, also I would very much like to have this function myself.
I can implement a `ShapleyImportanceEvaluator` which returns a list of `mean` SHAP value for each hyperparameter, and then using these mean values as the alternative feature importance, other than the default ANOVA one. 

On the other hand, the alternative plots to display the SHAP values would be the bar chart with confidence interval, or the scatter plot as the default `summary_plot` for the SHAP module. 

If you think the first option that I proposed is good enough, I can get down to the business and provide a PR soon.

Let me know what you think.

contramundum53:
Thank you for your interest! I think the first option is good. What do you think, @HideakiImamura?

HideakiImamura:
Looks good to me! I think this feature should added in `optuna.integrations` if we use https://github.com/slundberg/shap.

liaison:
@contramundum53 @HideakiImamura voila. Here is [the PR](https://github.com/optuna/optuna/pull/3507).

PR #3500, opened 2022-04-24
Filter trials with nonfinite value in `optuna.visualization.plot_param_importances` and `optuna.visualization.matplotlib.plot_param_importance` 

takoika:
## Motivation

This PR is a part of #3382. `optuna.visualization.plot_param_importances` and `optuna.visualization.matplotlib.plot_param_importance` will come to handle non-finite value properly.

## Description of the changes
Currently `plot_param_importances` accepts two evaluators: `MeanDecreaseImpurityImportanceEvaluator` and `MeanDecreaseImpurityImportanceEvaluator`. So this PR changes two evaluator to filter trials with an infinite value.

- [x] Filter trials with non-finite value
- [x] Add unit tests
- [x] Verify obtained graphs

Obtained graphs are below by using https://gist.github.com/HideakiImamura/67a150dec5774a18732527875774a9fe .

* wtih `FanovaImportanceEvaluator` with `Plotly`
plt_fnova

* with `FnovaImportanceEvaluator` with `Matplotlib`
mpl_fnova

* wtih `MeanDecreaseImpurityImportanceEvaluator` with `Plotly`
plt_mean

* wtih `MeanDecreaseImpurityImportanceEvaluator` with `Matplotlib`

mpl_mean

Note that I am not sure how filtering nonfinite values affect obtained importance scores.

PR #3507, opened 2022-04-24
calculate the feature importance with mean absolute SHAP values

liaison:
## Motivation
This PR adds the functionality to calculate the feature importance for the hyperparameters during the trial, based on the [SHAP values](https://shap.readthedocs.io/en/latest/index.html).  

## Description of the changes
Here is [the issue](https://github.com/optuna/optuna/issues/3448) that this PR is addressing. 
In order to calculate the SHAP values, we need a surrogate model, _e.g._ RandomForest. This PR uses the RandomForest model from the `MeanDecreaseImpurityImportanceEvaluator`.

### DIFF
--- a/optuna/importance/_mean_abs_shap.py
+++ b/optuna/importance/_mean_abs_shap.py
@@ -1,0 +1,79 @@
+from collections import OrderedDict
+from typing import Callable
+from typing import Dict
+from typing import List
+from typing import Optional
+
+import pandas as pd
+
+from optuna._imports import try_import
+from optuna.importance._mean_decrease_impurity import MeanDecreaseImpurityImportanceEvaluator
+from optuna.study import Study
+from optuna.trial import FrozenTrial
+
+
+with try_import() as _imports:
+    from shap import TreeExplainer
+
+
+class ShapleyImportanceEvaluator(MeanDecreaseImpurityImportanceEvaluator):
+    """Shapley (SHAP) parameter importance evaluator.
+
+    This evaluator fits a random forest that predicts objective values given hyperparameter
+    configurations. Feature importances are then computed using the mean absolute SHAP value for each feature.
+
+    .. note::
+
+        This evaluator requires the `sklean <https://scikit-learn.org/stable/>`_ Python package and
+        `SHAP <https://shap.readthedocs.io/en/latest/index.html>`_.
+        The model for the SHAP calculation is based on `sklearn.ensemble.RandomForestClassifier
+        <https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html>`_.
+
+    Args:
+        n_trees:
+            Number of trees in the random forest.
+        max_depth:
+            The maximum depth of each tree in the random forest.
+        seed:
+            Seed for the random forest.
+    """
+
+    def __init__(
+        self, *, n_trees: int = 64, max_depth: int = 64, seed: Optional[int] = None
+    ) -> None:
+        _imports.check()
+
+        MeanDecreaseImpurityImportanceEvaluator.__init__(
+            self, n_trees=n_trees, max_depth=max_depth, seed=seed
+        )
+        # Explainer from SHAP
+        self._explainer: TreeExplainer = None
+
+    def evaluate(
+        self,
+        study: Study,
+        params: Optional[List[str]] = None,
+        *,
+        target: Optional[Callable[[FrozenTrial], float]] = None,
+    ) -> Dict[str, float]:
+
+        # Train a RandomForest from the parent class
+        super().evaluate(study=study, params=params, target=target)
+
+        # Create Tree Explainer object that can calculate shap values
+        self._explainer = TreeExplainer(self._forest)
+
+        # Generate SHAP values for the parameters during the trials
+        shap_values = self._explainer.shap_values(self._trans_params)
+        df_shap = pd.DataFrame(shap_values, columns=self._param_names)
+
+        # Calculate the mean absolute SHAP value for each parameter
+        mean_abs_shap_values = []
+        for param in df_shap.columns:
+            mean_abs_shap_values.append((param, df_shap[param].abs().mean()))
+
+        # Use the mean absolute SHAP values as the feature importance
+        mean_abs_shap_values.sort(key=lambda t: t[1], reverse=True)
+        feature_importances = OrderedDict(mean_abs_shap_values)
+
+        return feature_importances
--- a/optuna/importance/_mean_decrease_impurity.py
+++ b/optuna/importance/_mean_decrease_impurity.py
@@ -52,6 +52,9 @@ class MeanDecreaseImpurityImportanceEvaluator(BaseImportanceEvaluator):
             min_samples_leaf=1,
             random_state=seed,
         )
+        self._trans_params = numpy.empty(0)
+        self._trans_values = numpy.empty(0)
+        self._param_names: List[str] = list()
 
     def evaluate(
         self,
@@ -82,27 +85,27 @@ class MeanDecreaseImpurityImportanceEvaluator(BaseImportanceEvaluator):
         trans = _SearchSpaceTransform(distributions, transform_log=False, transform_step=False)
 
         n_trials = len(trials)
-        trans_params = numpy.empty((n_trials, trans.bounds.shape[0]), dtype=numpy.float64)
-        trans_values = numpy.empty(n_trials, dtype=numpy.float64)
+        self._trans_params = numpy.empty((n_trials, trans.bounds.shape[0]), dtype=numpy.float64)
+        self._trans_values = numpy.empty(n_trials, dtype=numpy.float64)
 
         for trial_idx, trial in enumerate(trials):
-            trans_params[trial_idx] = trans.transform(trial.params)
-            trans_values[trial_idx] = trial.value if target is None else target(trial)
+            self._trans_params[trial_idx] = trans.transform(trial.params)
+            self._trans_values[trial_idx] = trial.value if target is None else target(trial)
 
         encoded_column_to_column = trans.encoded_column_to_column
 
-        if trans_params.size == 0:  # `params` were given but as an empty list.
+        if self._trans_params.size == 0:  # `params` were given but as an empty list.
             return OrderedDict()
 
         forest = self._forest
-        forest.fit(trans_params, trans_values)
+        forest.fit(self._trans_params, self._trans_values)
         feature_importances = forest.feature_importances_
         feature_importances_reduced = numpy.zeros(len(distributions))
         numpy.add.at(feature_importances_reduced, encoded_column_to_column, feature_importances)
 
         param_importances = OrderedDict()
-        param_names = list(distributions.keys())
+        self._param_names = list(distributions.keys())
         for i in feature_importances_reduced.argsort()[::-1]:
-            param_importances[param_names[i]] = feature_importances_reduced[i].item()
+            param_importances[self._param_names[i]] = feature_importances_reduced[i].item()
 
         return param_importances
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -5,6 +5,7 @@ from typing import Optional
 
 from optuna.importance._base import BaseImportanceEvaluator
 from optuna.importance._fanova import FanovaImportanceEvaluator
+from optuna.importance._mean_abs_shap import ShapleyImportanceEvaluator
 from optuna.importance._mean_decrease_impurity import MeanDecreaseImpurityImportanceEvaluator
 from optuna.study import Study
 from optuna.trial import FrozenTrial
@@ -14,6 +15,7 @@ __all__ = [
     "BaseImportanceEvaluator",
     "FanovaImportanceEvaluator",
     "MeanDecreaseImpurityImportanceEvaluator",
+    "ShapleyImportanceEvaluator",
     "get_param_importances",
 ]
 
--- a/tests/importance_tests/test_mean_abs_shap.py
+++ b/tests/importance_tests/test_mean_abs_shap.py
@@ -1,0 +1,73 @@
+from optuna import create_study
+from optuna import Trial
+from optuna.importance import ShapleyImportanceEvaluator
+from optuna.samplers import RandomSampler
+
+
+def objective(trial: Trial) -> float:
+    x1 = trial.suggest_float("x1", 0.1, 3)
+    x2 = trial.suggest_float("x2", 0.1, 3, log=True)
+    x3 = trial.suggest_float("x3", 2, 4, log=True)
+    return x1 + x2 * x3
+
+
+def test_mean_decrease_impurity_importance_evaluator_n_trees() -> None:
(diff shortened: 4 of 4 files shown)


optuna/importance/_mean_abs_shap.py
```diff
@@ -0,0 +1,79 @@
+from collections import OrderedDict
+from typing import Callable
+from typing import Dict
+from typing import List
+from typing import Optional
+
+import pandas as pd
```
nzw0301, line 7: pandas is an optional package in Optuna, could you move this import inside `try_import` block like this?
https://github.com/optuna/optuna/blob/c622be313074c0037fe3128ff5d20f9262c7af76/optuna/study/_dataframe.py#L14-L16

nzw0301:
As mentioned in https://github.com/optuna/optuna/issues/3448#issuecomment-1103648389, I also suggest adding this shape feature to `optuna.integration`, not `optuna.importance`.

himkt:
I'd like @contramundum53 to review the PR as another reviewer!

nzw0301:
To fix the CI error due to missing shap library, 
could you add `shap` as an element of https://github.com/optuna/optuna/blob/9d803682c5a72a56e12c5de9a6458ca2aaa503f6/setup.py#L94 and  https://github.com/optuna/optuna/blob/9d803682c5a72a56e12c5de9a6458ca2aaa503f6/setup.py#L141 in https://github.com/optuna/optuna/blob/master/setup.py.

optuna/importance/_mean_abs_shap.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __init__(
+        self, *, n_trees: int = 64, max_depth: int = 64, seed: Optional[int] = None
+    ) -> None:
+        _imports.check()
+
+        MeanDecreaseImpurityImportanceEvaluator.__init__(
+            self, n_trees=n_trees, max_depth=max_depth, seed=seed
+        )
+        # Explainer from SHAP
```
nzw0301, line 49:
```suggestion
        # Explainer from SHAP.
```
Optuna's coding style says we write a full sentence:
https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#block-and-inline-comments-are-supposed-to-be-sentences

optuna/importance/_mean_abs_shap.py
nzw0301: By the way, can we implement this feature without pandas? Ideally, we would not like to use an optional dependency for usability to make installation easier..

liaison: @nzw0301 Yep. I thought about that too. The usage of `pandas` is not mandatory, it is just that I'm more familiar with the module. I agree that it is better to keep the minimal dependency.

optuna/importance/_mean_abs_shap.py
liaison: Thanks for the info!

nzw0301:
Thank you for your update! I've skimmed the main logic, and it looks nice. I come up with one minor thing, so let me write here to remember it. 

I'm wondering if we could split the part of fitting random forest in  `optuna/importance/_mean_decrease_impurity.py` into a function because the current shap computes the importance score with the parent class: `MeanDecreaseImpurityImportanceEvaluator`, whose importance score calculation can be skipped when we evaluate shap based importance score. But it can be addressed in the follow-up PR.

optuna/importance/_mean_decrease_impurity.py
```diff
@@ -52,6 +52,9 @@ def __init__(
             min_samples_leaf=1,
             random_state=seed,
         )
+        self._trans_params = numpy.empty(0)
+        self._trans_values = numpy.empty(0)
+        self._param_names: List[str] = list()
```
nzw0301, line 57: My minor concern comes from this change. In my understanding, `self. _trans_params` and `self._param_names` are stored to be used in the `evaluate` method in the shap class after computing the parent's `evaluate` method. Similar to the comment on the decomposition of random forest fitting, this part also might be split as a method or static method rather than storing them as instance attributes.

g-votte:
In my opinion, the `ShapleyImportanceEvaluator` should not inherit `MeanDecreaseImpurityImportanceEvaluator` but hold it as "has-a" relationship.

As known as the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle), a sub class should not break any behavior of its superclass. In other words, as long as an evaluator is a (grandchild) instance of `MeanDecreaseImpurityImportanceEvaluator`, it should evaluate a study with the mean decrease impurity algorithm. In the current PR, however, `ShapleyImportanceEvaluator` breaks the behavior of `MeanDecreaseImpurityImportanceEvaluator` replacing it with a totally different algorithm.

### Alternative approach 1
We can implement `ShapleyImportanceEvaluator` as a direct subclass of `BaseImportanceEvaluator` and have it hold a `MeanDecreaseImpurityImportanceEvaluator` as a property.

```python
class ShapleyImportanceEvaluator(BaseImportanceEvaluator):
    def __init__(
        self, *, n_trees: int = 64, max_depth: int = 64, seed: Optional[int] = None
    ) -> None:
        _imports.check()

        self._backend_evaluator = MeanDecreaseImpurityImportanceEvaluator(
            n_trees=n_trees, max_depth=max_depth, seed=seed
        )
        self._explainer: TreeExplainer = None

    def evaluate(...) -> Dict[str, float]:
        # Train a RandomForest from the parent class.
        self._backend_evaluator.evaluate(study=study, params=params, target=target)

        # Get necessary properties of self._backend_evaluator in the following lines.
        ...
```

### Alternative approach 2
This is related to @nzw0301’s [previous comment](https://github.com/optuna/optuna/pull/3507#issuecomment-1109782959). We might be able to remove even the "has-a" relationship once we split the random forest fitting as a static function. Both `ShapleyImportanceEvaluator` and `MeanDecreaseImpurityImportanceEvaluator` just rely on the split function instead of being tightly-coupled each other.


I propose to rewrite the `ShapleyImportanceEvaluator` in either of the alternative approaches above. What do you think? @liaison @nzw0301 @himkt @contramundum53

nzw0301:
@g-votte Thank you for your suggestion with a detailed explanation! I agree with you.

liaison:
hi @g-votte @nzw0301 I quite agree with you both actually. To tell you a secret, it bothered me a bit to have `ShapleyImportanceEvaluator` inherited from `MeanDecreaseImpurityImportanceEvaluator`. 

Admittedly I took a little shortcut to have the least changes for the implementation. I thought about extracting the logic of retrieving trial parameters and fitting a RandomForest as a separate function. But the logic requires quite some context (tedious to pass everything as parameters). In addition, the exception handling in between complicates things further. Inheriting all this logic seems like an easy choice while minimizing the changes. Voila. So you can say it is a choice of "optimization" :) 

We could further refactor the classes should we have more evaluators in the future, or we want to have a different surrogate model (other than RandomForeset) for SHAP value calculation. Let me know what you think.

optuna/importance/_mean_decrease_impurity.py
```diff
@@ -82,27 +85,27 @@ def evaluate(
...
         numpy.add.at(feature_importances_reduced, encoded_column_to_column, feature_importances)
 
         param_importances = OrderedDict()
-        param_names = list(distributions.keys())
+        self._param_names = list(distributions.keys())
         for i in feature_importances_reduced.argsort()[::-1]:
-            param_importances[param_names[i]] = feature_importances_reduced[i].item()
+            param_importances[self._param_names[i]] = feature_importances_reduced[i].item()
 
         return param_importances
```
contramundum53, line 111: Thank you for your PR! 
As also mentioned by @nzw0301, we don't actually need to introduce these member fields. Instead of having `_trans_params`, `_trans_values` and `_param_names` as member fields, we could make another function `_evaluate` returning `(param_importances, trans_params, trans_values, param_names)` in `MeanDecreaseImpurityImportanceEvaluator`  and call that function from both `ShapleyImportanceEvaluator.evaluate()` and `MeanDecreaseImpurityImportanceEvaluator.evaluate()`.

nzw0301: I would like to suggest spliting the current evaluation methods into i) pre-processing part that returns `trans, trans_params, trans_values` and ii) the computing importance score part because the evaluator classes including this shap class call this parameter transformation. The first part might be the static method and the second part calls the first pre-processing method. 
I'm still not sure this approach is optimal.

liaison: hi @contramundum53 @nzw0301 I do understand your concerns about separating the concerns and removing the class members. The problem is the implementation of `MeanDecreaseImpurityImportanceEvaluator.evaluate()` is so intertwined with many states and variables. 

If we were to have a separate method to retrieve the parameters from the trials, then returning the tuple of `(trans, trans_params, trans_values)` isn't sufficient, we need also the `distribution` which is used at the end of the function. In addition, there are some exceptions raising and early returns in-between that need to be properly re-transformed. I'm not convinced that there is much to gain by doing so. 

Will the code be more modularized? probably. But one might argue that having the class encapsulate the state information is itself a good way to modularize the code, which is the current situation now.

nzw0301: Right, but then the pre-processing takes `distributions` as an argument to be reused after random forest fitting. This does not need to return an empty `OrderedDict`; we might return it before and after preprocessing.

What do you think @g-votte? I would like to know whether the first approach suggested in https://github.com/optuna/optuna/pull/3507#issuecomment-1110439225 is the current implementation or not. Specifically, can we introduce new private attributes to `MeanDecreaseImpurityImportanceEvaluator` for `ShapleyImportanceEvaluator`.

liaison: hi @nzw0301 @g-votte @contramundum53 I've given it another thought. 

First of all, I think it is a good thing to keep the class `MeanDecreaseImpurityImportanceEvaluator` **stateful**, rather than insisting on having a _stateless_ `evaluate()` function. One would like to look into the evaluator after the evaluation if there is anything wrong, or simply the state information could be useful for other purposes (e.g. pass it on to `ShapleyImportanceEvaluator`).

If you agree with the above point, then we can address the points that make you guys uneasy (including myself), namely 1). we access directly the attributes of `MeanDecreaseImpurityImportanceEvaluator` (although Python doesn't have any access control anyway). 2). it is not entirely necessary to evaluate `MeanDecreaseImpurityImportanceEvaluator` in `ShapleyImportanceEvaluator`. 

To address the above two concerns, here is what I propose (hopefully the code is self-explanatory): 

```
       # Retrieve the trial parameters and fit a RandomForest model out of them.
       model = self._backend_evaluator.fit(study=study, params=params, target=target)
       
       # Retrieve the trial parameters.
       (trans_params, trans_values, param_names)  = self._backend_evaluator.get_params()
```

PS: the above APIs are inspired by the `sklearn.BaseEstimator`.

liaison: hi @nzw0301 @g-votte @contramundum53 any feedback on the above proposal?

contramundum53: Sorry, we were on a vacation.
Thanks, I agree with your proposal.
One minor comment: I think it would be more readable and robust to make the return value of `get_params()` a `dict` instead of a `tuple`.

optuna/integration/shap.py
```diff
@@ -0,0 +1,86 @@
...
+from optuna.importance._base import BaseImportanceEvaluator
+from optuna.importance._mean_decrease_impurity import MeanDecreaseImpurityImportanceEvaluator
+from optuna.study import Study
+from optuna.trial import FrozenTrial
+
+
+with try_import() as _imports:
+    from shap import TreeExplainer
+
+
```
nzw0301, line 19: Could you add `experimental`? 
```suggestion

@experimental("3.0.0")
```

Note that we need to add `from optuna._experimental import experimental` to the import block

optuna/integration/shap.py
```diff
@@ -0,0 +1,86 @@
...
+        # Calculate the mean absolute SHAP value for each parameter.
+        mean_abs_shap_values = []
+        for param_index in range(shap_values.shape[1]):
+            # Add tuples of ("feature_name": mean_abs_shap_value).
+            mean_abs_shap_values.append(
+                (
+                    self._backend_evaluator._param_names[param_index],
+                    np.abs(shap_values[:, param_index]).mean(),
+                )
+            )
```
nzw0301, line 80: simplified:
```suggestion
        # Tuples of ("feature_name": mean_abs_shap_value).
        mean_abs_shap_values = tuple(
            zip(self._backend_evaluator._param_names, np.abs(shap_values).mean(axis=0))
        )
```

optuna/integration/shap.py
```diff
@@ -0,0 +1,86 @@
...
+class ShapleyImportanceEvaluator(BaseImportanceEvaluator):
+    """Shapley (SHAP) parameter importance evaluator.
+
+    This evaluator fits a random forest that predicts objective values given hyperparameter
+    configurations. Feature importances are then computed as the mean absolute SHAP values.
+
+    .. note::
+
+        This evaluator requires the `sklean <https://scikit-learn.org/stable/>`_ Python package and
+        `SHAP <https://shap.readthedocs.io/en/latest/index.html>`_.
```
nzw0301, line 29: Since users tend to the stable version, how about referring to the stable docs.
```suggestion
        `SHAP <https://shap.readthedocs.io/en/stable/index.html>`_.
```

nzw0301 (approved):
Sorry for the delayed response. I agree with your suggestions at this time: make this importance evaluator stateful in this PR. 
Since we notice that all importance evaluators have common logics thanks to this PR, we shall refactor them in the future, but I think this PR should not address such refactoring. Hence, the pull request looks great to me now.

Nevertheless, I've left a few comments that can be fixed by a follow-up pull request.


After creating this PR, https://github.com/optuna/optuna/pull/3500 has introduced two new tests:
`test_mean_decrease_impurity_importance_evaluator_with_infinite` and `test_multi_objective_mean_decrease_impurity_importance_evaluator_with_infinite` to the test of `MeanDecreaseImpurityImportanceEvaluator`. We'll add a similar test functions to shap's counterpart.

optuna/importance/_mean_decrease_impurity.py
g-votte: Sorry for coming late. I think the current PR reflects what I suggested in https://github.com/optuna/optuna/pull/3507#issuecomment-1110439225 .

As mentioned by @nzw0301 in https://github.com/optuna/optuna/pull/3507#pullrequestreview-970496123, we'd need some followup refactoring to extract the common logic among importance evaluators and to get rid of unnecessary calculation with the current approach. #3552 will shortly address those problems.

contramundum53:
I experimented with [this objective](https://optuna.readthedocs.io/en/stable/reference/visualization/generated/optuna.visualization.plot_param_importances.html?highlight=plot_param_importances) and `ShapleyImportanceEvaluator`. 

This is the result for `ShapleyImportanceEvaluator`.
スクリーンショット 2022-05-13 12 10 54

Just for reference, this is the result for the default fANOVA importance evaluator.
スクリーンショット 2022-05-13 12 11 01

The result looks good to me.

PR #3514, opened 2022-04-26
Correct `FronzenTrial`'s docstring

### LINKED ISSUE #3488: The document of `FrozenTrial` is wrong
### What is an issue?

The document says "it suggests **best** parameter values among performed trials." But it is not correct.

nzw0301:
## Motivation

Fix #3488.

## Description of the changes

- Correct the explanation of behavior of `suggest` methods.
- Minor update in docstring.

### DIFF
--- a/optuna/trial/_frozen.py
+++ b/optuna/trial/_frozen.py
@@ -29,7 +29,7 @@ class FrozenTrial(BaseTrial):
     """Status and results of a :class:`~optuna.trial.Trial`.
 
     This object has the same methods as :class:`~optuna.trial.Trial`, and it suggests the same
-    parameter values as in :attr:`params`; it does not sample any value from distribution.
+    parameter values as in :attr:`params`; it does not sample any value from a distribution.
     In contrast to :class:`~optuna.trial.Trial`,
     :class:`~optuna.trial.FrozenTrial` does not depend on :class:`~optuna.study.Study`, and it is
     useful for deploying optimization results.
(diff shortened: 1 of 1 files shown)


optuna/trial/_frozen.py
```diff
@@ -28,8 +28,9 @@
 class FrozenTrial(BaseTrial):
     """Status and results of a :class:`~optuna.trial.Trial`.
 
-    This object has the same methods as :class:`~optuna.trial.Trial`, and it suggests best
-    parameter values among performed trials. In contrast to :class:`~optuna.trial.Trial`,
+    This object has the same methods as :class:`~optuna.trial.Trial`, and it suggests the same
+    parameter values as in :attr:`params`; it does not suggest fresh parameter values.
```
himkt, line 32: I'm not sure `fresh parameter values` is well defined for users. How about:

```suggestion
    parameter values as in :attr:`params`; it does not sample any value from distribution.
```

optuna/trial/_frozen.py
```diff
@@ -28,8 +28,9 @@
 class FrozenTrial(BaseTrial):
     """Status and results of a :class:`~optuna.trial.Trial`.
 
-    This object has the same methods as :class:`~optuna.trial.Trial`, and it suggests best
-    parameter values among performed trials. In contrast to :class:`~optuna.trial.Trial`,
+    This object has the same methods as :class:`~optuna.trial.Trial`, and it suggests the same
+    parameter values as in :attr:`params`; it does not sample any value from distribution.
```
himkt, line 32:
```suggestion
    parameter values as in :attr:`params`; it does not sample any value from a distribution.
```

So sorry... :bow:

himkt (approved):

Issue #3485, opened 2022-04-22
`--verbose` and `--quiet` option in CLI is not working


not522:
### Expected behavior

`--verbose` and `--quiet` options in CLI should change the logging level of the Optuna's logger.

### Environment

- Optuna version: 3.0.0b1.dev
- Python version: 3.8.2
- OS: macOS-10.16-x86_64-i386-64bit
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
See steps to reproduce.
```

### Steps to reproduce

```sh
~/optuna$ optuna create-study --storage sqlite:///test.db --study-name test-study -q
[I 2022-04-22 20:36:11,192] A new study created in RDB with name: test-study
test-study
```

The INFO level logging message should be suppressed by `-q`.

hvy:
@nyanhi is working on debugging this issue.

PR #3532, opened 2022-05-04
Fix `--verbose` and `--quiet` options in CLI

nyanhi:
## Motivation
To address the issue #3485.

## Description of the changes
`_OptunaApp` will set verbosity with the same value specified by its super `App` in `cliff.app`. Also test code is added.
Ref. https://github.com/openstack/cliff/blob/master/cliff/app.py#L211-L219

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1560,7 +1560,8 @@ def test_configure_logging_verbosity(verbosity: str, expected: bool) -> None:
 
         # Create study.
         args = ["optuna", "create-study", "--storage", storage_url, verbosity]
-        # If `--verbose`, the INFO level log message should be generated.
+        # `--verbose` makes the log level DEBUG.
+        # `--quiet` makes the log level WARNING.
         result = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         error_message = result.stderr.decode()
         assert ("A new study created in RDB with name" in error_message) == expected
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1542,3 +1542,25 @@ def test_tell_with_nan() -> None:
         assert len(study.trials) == 1
         assert study.trials[0].state == TrialState.FAIL
         assert study.trials[0].values is None
+
+
+@pytest.mark.skip_coverage
+@pytest.mark.parametrize(
+    "verbosity, expected",
+    [
+        ("--verbose", True),
+        ("--quiet", False),
+    ],
+)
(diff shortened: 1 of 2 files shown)


not522 (commented):
Thank you for your PR! Please check my little comment for the test.

nyanhi:
Thank you for reviewing. Sorry if I'm missing some basic point, but may I check if the comment has saved? I cannot see it somehow.

not522:
Sorry, my comment was gone somewhere 🤔

tests/test_cli.py
```diff
@@ -1542,3 +1542,25 @@ def test_tell_with_nan() -> None:
...
+)
+def test_configure_logging_verbosity(verbosity: str, expected: bool) -> None:
+
+    with StorageSupplier("sqlite") as storage:
+        assert isinstance(storage, RDBStorage)
+        storage_url = str(storage.engine.url)
+
+        # Create study.
+        args = ["optuna", "create-study", "--storage", storage_url, verbosity]
+        # If `--verbose`, the INFO level log message should be generated.
```
not522, line 1563: Your unit tests seem OK, but could you clarify the logging level on the comment? The logging level with the `--verbose` option is `DEBUG`. I'm not sure it is easy to test `DEBUG` level logging messages because Optuna has a small number of debug messages.

nyanhi: This unit test confirms the necessary condition that the behavior changes depending on the value of the `--verbose` option. As you point out, it doesn't fully test for the `DEBUG` level - sorry I couldn't think of simple way to generate `DEBUG` logging message via CLI.

not522: > sorry I couldn't think of simple way to generate DEBUG logging message via CLI.

I agree it is not easy to test.

> If `--verbose`, the INFO level log message should be generated.

This comment seems confusing about the log level with the `--verbose` option. How about

```python
# `--verbose` makes the log level DEBUG.
# `--quiet` makes the log level WARNING.
```

nyanhi: Thank you so much for the suggestion, I have updated the comment. Kindly let me know if anything else.

not522 (approved):
Thank you! LGTM!

HideakiImamura (approved):
Thanks for the fix! LGTM.

I confirmed that the INFO message is suppressed with `--quiet`.
```
(venv) mamu@HideakinoMacBook-puro optuna % optuna create-study --storage sqlite:///test.db --study-name test-study -q
test-study
```

PR #3538, opened 2022-05-05
Remove `get_study_id_from_trial_id`

keisuke-umezawa:
## Motivation

I would like to remove get_study_id_from_trial_id because it is not used anywhere in optuna.

## Description of the changes

- [x] Add deprecation to get_study_id_from_trial_id.

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -11,7 +11,6 @@ from typing import Tuple
 from typing import Union
 
 import optuna
-from optuna._deprecated import deprecated
 from optuna.distributions import BaseDistribution
 from optuna.study._study_direction import StudyDirection
 from optuna.study._study_summary import StudySummary
@@ -207,23 +206,6 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         Raises:
             :exc:`KeyError`:
                 If no study with the matching ``study_name`` exists.
-        """
-        raise NotImplementedError
-
-    @deprecated("3.0.0", "4.0.0")
-    def get_study_id_from_trial_id(self, trial_id: int) -> int:
-        """Read the ID of a study to which a trial belongs.
-
-        Args:
-            trial_id:
-                ID of the trial.
-
-        Returns:
-            ID of the study.
-
-        Raises:
-            :exc:`KeyError`:
-                If no trial with the matching ``trial_id`` exists.
         """
         raise NotImplementedError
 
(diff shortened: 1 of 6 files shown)


HideakiImamura:

optuna/storages/_base.py
```diff
@@ -209,7 +210,7 @@ def get_study_id_from_name(self, study_name: str) -> int:
         """
         raise NotImplementedError
 
-    @abc.abstractmethod
+    @deprecated("2.9.0", "4.0.0")
```
toshihikoyanase, line 213: [Question] Could you tell me why you specify `2.9.0` instead of `3.0.0`?

c-bata (approved):
LGTM after fixed deprecated version as @toshihikoyanase commented.

optuna/storages/_base.py
keisuke-umezawa: Sorry, it should be 3.0.0! Thank you.

keisuke-umezawa:
@toshihikoyanase 
Thank you for your review. I updated this PR. Could you check it again?

optuna/storages/_base.py
toshihikoyanase: Thank you for your update. 

By the way, we usually remove deprecated methods after two major releases. For example, `optuna.integration.CmaEsSampler` was deprecated at 2.0.0 and will be removed at 4.0.0.

> optuna/integration/cma.py:@deprecated("2.0.0", "4.0.0", text=_cma_deprecated_msg)

Do you intentionally set the `removed_version` to 4.0.0?

IMO, it looks reasonable to remove this method at 4.0.0 since it is not a user-facing API and no one uses it.

toshihikoyanase:
@keisuke-umezawa Thank you for your update. I have an additional question about `removed_version` value.
https://github.com/optuna/optuna/pull/3538#discussion_r874706572

optuna/storages/_base.py
toshihikoyanase: @g-votte suggested that this method could be removed without deprecation.
This is mainly because it is no longer used and it is a private API.
What do you think of it @keisuke-umezawa ?

keisuke-umezawa: I agreed on it. I will remove those methods.

keisuke-umezawa:
@toshihikoyanase 
I removed "get_study_id_from_trial_id". Could you check it again?

toshihikoyanase (approved):

c-bata (approved):

PR #3543, opened 2022-05-06
Remove the internal logic that calculates the interaction of two or more variables in fANOVA

c-bata:
## Motivation

There are some unreachable lines in fANOVA module.

## Description of the changes

* Remove unreachable lines.
* Use `np.where()` instead of list comprehension syntax
* Use `numpy.clip(x, 0.0, None)` instead of `numpy.clip(x, 0.0, np.inf)`
* Use new style classes.


optuna/importance/_fanova/_fanova.py
```diff
@@ -73,58 +72,40 @@ def fit(
...
         self._column_to_encoded_columns = column_to_encoded_columns
-        self._variances = {}
 
         if all(tree.variance == 0 for tree in self._trees):
             # If all trees have 0 variance, we cannot assess any importances.
             # This could occur if for instance `X.shape[0] == 1`.
             raise RuntimeError("Encountered zero total variance in all trees.")
 
-    def get_importance(self, features: Tuple[int, ...]) -> Tuple[float, float]:
+    def get_importance(self, feature: int) -> Tuple[float, float]:
```
hvy, line 81: Seems fair 👍  Just for the record, we used to accept tuples in order to allow/check the computations for pairwise importances. But it's not available through public APIs for the time being and we can reintroduce the logic whenever needed.

HideakiImamura:
@hvy Could you review this PR?

@c-bata Could you fix the CI errors?

c-bata:
Thank you. I've fixed CI errors :bow:

optuna/importance/_fanova/_fanova.py
```diff
@@ -73,58 +72,41 @@ def fit(
 
         self._trees = [_FanovaTree(e.tree_, search_spaces) for e in self._forest.estimators_]
         self._column_to_encoded_columns = column_to_encoded_columns
-        self._variances = {}
```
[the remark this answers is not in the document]
c-bata: Thank you. That makes sense! I'll revert this line.

optuna/importance/_fanova/_tree.py
```diff
@@ -33,13 +33,7 @@ def __init__(self, tree: "sklearn.tree._tree.Tree", search_spaces: numpy.ndarray
...
     def variance(self) -> float:
         if self._variance is None:
-            leaf_node_indices = numpy.array(
-                [
-                    node_index
-                    for node_index in range(self._n_nodes)
-                    if self._is_node_leaf(node_index)
-                ]
-            )
+            leaf_node_indices = numpy.where(numpy.array(self._tree.feature) < 0)[0]
```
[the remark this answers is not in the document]
c-bata: Good catch! I confirmed that we can use `np.nonzero()` here.

```python
(Pdb) p numpy.where(numpy.array(self._tree.feature) < 0)[0]
array([1, 2])
(Pdb) p numpy.nonzero(numpy.array(self._tree.feature) < 0)[0]
array([1, 2])
```

The bottleneck of the FanovaImportanceEvaluator is not actually in the method. No performance gains can be obtained with this change.

optuna/importance/_fanova/_fanova.py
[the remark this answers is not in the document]
hvy: Great, thanks!

optuna/importance/_fanova/_fanova.py
```diff
@@ -80,51 +79,35 @@ def fit(
             # This could occur if for instance `X.shape[0] == 1`.
             raise RuntimeError("Encountered zero total variance in all trees.")
 
-    def get_importance(self, features: Tuple[int, ...]) -> Tuple[float, float]:
+    def get_importance(self, feature: int) -> Tuple[float, float]:
```
hvy, line 82: Perhaps we should update the documentation at the top of this file that mentions pairwise importances (which are no longer computable after this change).

HideakiImamura: +1. In addition, we need to change the documentation of `FanovaImportanceEvaluator` https://github.com/optuna/optuna/blob/master/optuna/importance/_fanova/_evaluator.py#L36.

HideakiImamura:
I think the title of this PR is less informative. How about changing it like "Remove the internal logic that calculates the interaction of two or more variables"?

optuna/importance/_fanova/_fanova.py
c-bata: Thanks! Good catch. I'll update the documentation.

c-bata:
@HideakiImamura @hvy Thank you for your review. I applied your suggestions :bow:

hvy:
> I think the title of this PR is less informative. How about changing it like "Remove the internal logic that calculates the interaction of two or more variables"?

How about ending with "in fANOVA" to make clear the domain?

optuna/importance/_fanova/_fanova.py
```diff
@@ -55,7 +51,7 @@ def __init__(
             random_state=seed,
         )
         self._trees: Optional[List[_FanovaTree]] = None
-        self._variances: Optional[Dict[Tuple[int, ...], numpy.ndarray]] = None
+        self._variances: Dict[int, numpy.ndarray] = {}
         self._column_to_encoded_columns: Optional[List[numpy.ndarray]] = None
```
hvy, line 55: `None` -> `{}` looks like a simplification. At the same time, having only one of the three attributes non-optional seems less clear than the previous logic, as if `self._variances` has a different behavior/semantic. What do you think?

optuna/importance/_fanova/_fanova.py
```diff
@@ -80,51 +76,35 @@ def fit(
...
             return
 
-        for k in range(1, len(features)):
-            for sub_features in itertools.combinations(features, k):
-                if sub_features not in self._variances:
-                    self._compute_variances(sub_features)
-
-        raw_features = numpy.concatenate([self._column_to_encoded_columns[f] for f in features])
-
+        raw_feature = self._column_to_encoded_columns[feature]
```
hvy, line 104: A nit but how about keeping it plural? I think it's more natural given that it's still an ndarray with possibly more than 1 element (for categorical with more than 1 choice). It's still also the name of the argument in `get_marginal_variance`.
```suggestion
        raw_features = self._column_to_encoded_columns[feature]
```

optuna/importance/_fanova/_fanova.py
```diff
@@ -12,14 +12,10 @@
 This stands in contrast to the original implementation which is partially written in C++.
 Since Python runtime overhead may become noticeable, included are instead several
 optimizations, e.g. vectorized NumPy functions to compute the marginals, instead of keeping all
-running statistics. Known cases include assessing higher order importances, e.g. pairwise
-importances, this is due to the fact that the number of partitions to visit grows exponentially,
-or when assessing categorical features with a larger number of choices since each choice is
-given a unique one-hot encoded raw feature.
```
hvy, line 18: We can keep the categorical part I think. It's still relevant without the pairwise importances.

c-bata: Oh, thanks. I reverted it at 72b564a

optuna/importance/_fanova/_fanova.py
c-bata: Thank you for your review. Yes, it may contain multiple values :bow: I renamed it at 
72b564a.

optuna/importance/_fanova/_fanova.py
c-bata: Thanks! For me either sounds fine. But I don't need to make it a non-optional attribute at least. I made `self._variances` optional at 72b564a.

c-bata:
> How about ending with "in fANOVA" to make clear the domain?

Thanks! I updated the title and the code.

HideakiImamura (approved):
LGTM.

optuna/importance/_fanova/_fanova.py
```diff
@@ -12,14 +12,10 @@
...
-
+running statistics. Known cases include when assessing categorical features with a larger
```
hvy, line 15: Nit.
```suggestion
running statistics. Known cases include assessing categorical features with a larger
```

c-bata: Thank you :bow: I applied it at e9a0955.

hvy (approved):
Thanks for the updates and for syncing with the latest master branch to fix the CI failure, LGTM!

PR #3544, opened 2022-05-07
Take the same search space as in `CategoricalDistribution` by `GridSearchSampler`

### LINKED ISSUE #3535: `GridSearchSampler` raises `TypeError` when `choices` of `CategoricalDistribution` are not comparable
### Expected behavior

The folliwing grid search based optimisation works without any error because default sampler works fine.
```python
import optuna


def objective(trial):
    y = trial.suggest_categorical("y", [None, 1])
    if y is None:
        y = -1
    return x ** 2 + y

# Grid search.
search_space = {"x": [-50, 0, 50], "y": [None, 1]}
study = optuna.create_study(sampler=optuna.samplers.GridSampler(search_space))


# Note Random search + TPE works well.
# study.optimize(objective, n_trials=20)
```

### Environment

- Optuna version: 3.0.0b1.dev
- Python version: 3.9.12
- OS: macOS-10.16-x86_64-i386-64bit

```shell
TypeError                                 Traceback (most recent call last)
Input In [15], in <cell line: 17>()
     14 study.optimize(objective, n_trials=20)
     16 # Grid search
---> 17 study = optuna.create_study(sampler=optuna.samplers.GridSampler(search_space))
     18 study.optimize(objective, n_trials=20)

File ~/Documents/optuna/optuna/samplers/_grid.py:112, in GridSampler.__init__(self, search_space)
    109 for param_name, param_values in sorted(search_space.items(), key=lambda x: x[0]):
    110     param_values = cast(SortableParamValueSequenceType, param_values)
--> 112     self._search_space[param_name] = sorted(param_values)
    114 self._all_grids = list(itertools.product(*self._search_space.values()))
    115 self._param_names = sorted(search_space.keys())

TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

### Steps to reproduce

1. Run the code above
2.
3.
```python
# python code
```

Since grid search sampler implementation sorts `choices`, the current implementation assumes the choices are sortable.

nzw0301:
## Motivation

Fix #3535.

## Description of the changes

Remove `sorted` part.

### DIFF
--- a/optuna/samplers/_grid.py
+++ b/optuna/samplers/_grid.py
@@ -20,7 +20,7 @@ from optuna.trial import TrialState
 
 
 GridValueType = Union[str, float, int, bool, None]
-SortableParamValueSequenceType = Union[List[str], List[float], List[int], List[bool]]
+SortableParamValueSequenceType = Sequence[GridValueType]
 
 
 _logger = get_logger(__name__)
@@ -106,10 +106,9 @@ class GridSampler(BaseSampler):
                 self._check_value(param_name, value)
 
         self._search_space = collections.OrderedDict()
-        for param_name, param_values in sorted(search_space.items(), key=lambda x: x[0]):
+        for param_name, param_values in sorted(search_space.items()):
             param_values = cast(SortableParamValueSequenceType, param_values)
-
-            self._search_space[param_name] = sorted(param_values)
+            self._search_space[param_name] = sorted(param_values, key=lambda x: str(x))
 
         self._all_grids = list(itertools.product(*self._search_space.values()))
         self._param_names = sorted(search_space.keys())
@@ -256,7 +255,7 @@ class GridSampler(BaseSampler):
                 return False
 
             param_values = cast(SortableParamValueSequenceType, search_space[param_name])
-            for i, param_value in enumerate(sorted(param_values)):
+            for i, param_value in enumerate(sorted(param_values, key=lambda x: str(x))):
                 if param_value != self._search_space[param_name][i]:
                     return False
 
--- a/optuna/samplers/_grid.py
+++ b/optuna/samplers/_grid.py
@@ -2,7 +2,6 @@ import collections
 import itertools
 import random
 from typing import Any
-from typing import cast
 from typing import Dict
 from typing import List
 from typing import Mapping
@@ -20,7 +19,6 @@ from optuna.trial import TrialState
 
 
 GridValueType = Union[str, float, int, bool, None]
-SortableParamValueSequenceType = Sequence[GridValueType]
 
 
 _logger = get_logger(__name__)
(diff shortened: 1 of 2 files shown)


optuna/samplers/_grid.py
```diff
@@ -20,7 +20,7 @@
 
 
 GridValueType = Union[str, float, int, bool, None]
-SortableParamValueSequenceType = Union[List[str], List[float], List[int], List[bool]]
+SortableParamValueSequenceType = Sequence[GridValueType]
```
nzw0301, line 23: Since `List` is not the same as the type of `choices` in `CateoricalDistriibution`, `Sequence` is supposed to be more accurate.

not522 (commented):
I'm not sure, but is the sorting needed? I think the order given by users is enough to calculate grid ids.

HideakiImamura:

optuna/samplers/_grid.py
```diff
@@ -106,10 +104,8 @@ def __init__(self, search_space: Mapping[str, Sequence[GridValueType]]) -> None:
                 self._check_value(param_name, value)
 
         self._search_space = collections.OrderedDict()
-        for param_name, param_values in sorted(search_space.items(), key=lambda x: x[0]):
-            param_values = cast(SortableParamValueSequenceType, param_values)
-
-            self._search_space[param_name] = sorted(param_values)
+        for param_name, param_values in sorted(search_space.items()):
```
contramundum53, line 107:
```suggestion
        for param_name, param_values in sorted(search_space.items(), key=lambda x: x[0]):
```
Is there some reason to sort by `param_values` also?

contramundum53 (commented):
Thanks for your PR!
I left a minor comment. Otherwise it looks good to me.

optuna/samplers/_grid.py
not522: This does not change the behavior because `param_name`s are distinct.

contramundum53: Yes if `x[1]` is of a sortable type, but if `x[1] is None`, wouldn't that be an error?

contramundum53: (I am not sure whether the lexicographical comparison in Python is strict or lazy, but just for mental ease :) )

not522 (approved):
LGTM!

optuna/samplers/_grid.py
contramundum53: `search_space.items()` is a sequence of tuples of `str` and `Sequence[GridValueType]`s, and a `Sequence[GridValueType]` may not be order-comparable (a generator would be an example). While it is true that there would never be overlapping keys and the `Sequence` element never participates in comparison (thus there would never be a `TypeError`), I think it would be more readable if you explicitly specified the sorting key to be the first element, or at least left some comment.

contramundum53: (It's just my mental type checker complaining about the code, and if you think it is more readable this way, I'll merge this as it is.)

contramundum53 (approved):

optuna/samplers/_grid.py
nzw0301: Thank you for your explanation! Hmm, I guess other parts in Optuna do a similar sort without explicit key specification. So the current code looks fine.

PR #3545, opened 2022-05-07
Replace `ValueError` with `warning` in `GridSearchSampler`

### LINKED ISSUE #3540: Show warning message by `GridSearchSampler` whose `CategoricalDistribution` contains unsupported typed values, not raising `ValueError`
### Motivation

The grid search sampler has a strict type rule for its search space for `CategoricalDistribution`, which is not consistent with the other samplers as reported in #3534.

### Description

`CategoriaclDistribution` shows a warning message when the `choices` argument contains an unsupported type.

When we run the following code
```python
import optuna

optuna.distributions.CategoricalDistribution([[1], [2]])
```

then we see the following warning message:

```bash
/Users/nzw/Documents/optuna/optuna/distributions.py:501: UserWarning: Choices for a categorical distribution should be a tuple of None, bool, int, float and str for persistent storage but contains [1] which is of type list.
  warnings.warn(message)
/Users/nzw/Documents/optuna/optuna/distributions.py:501: UserWarning: Choices for a categorical distribution should be a tuple of None, bool, int, float and str for persistent storage but contains [2] which is of type list.
  warnings.warn(message)
```

On the other hand, the grid search sampler raises `ValueError` when we use such unsupported typed value as an element of `choices` of the distribution.

### Alternatives (optional)

_No response_

_No response_

nzw0301:
## Motivation

Resolve #3540.
## Description of the changes

By following `CategoricalDistribution`, the grid sampler shows a warning message when search space contains unsupported types.

### DIFF
--- a/optuna/samplers/_grid.py
+++ b/optuna/samplers/_grid.py
@@ -211,11 +211,12 @@ class GridSampler(BaseSampler):
         if param_value is None or isinstance(param_value, (str, int, float, bool)):
             return
 
-        raise ValueError(
+        message = (
             "{} contains a value with the type of {}, which is not supported by "
             "`GridSampler`. Please make sure a value is `str`, `int`, `float`, `bool`"
-            " or `None`.".format(param_name, type(param_value))
+            " or `None` for persistent storage.".format(param_name, type(param_value))
         )
+        warnings.warn(message)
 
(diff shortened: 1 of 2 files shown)


not522 (approved):
LGTM.

HideakiImamura (approved):
LGTM.

PR #3546, opened 2022-05-07
Improve the tests of `optuna.visualization.plot_pareto_front`

### LINKED ISSUE #3502: Missing test case for `optuna.visualization.plot_pareto_front`
### Motivation

#3495 added the `target` option to `optuna.visualization.plot_pareto_front`, but the corresponding test case does not check invalid `targets` function.

For example, users may input the `target` functions that returns more than four values (i.e., https://github.com/optuna/optuna/pull/3495#discussion_r857078406). 

### Description

Please add test cases that inputs invalid `targets` functions and check errors.

The expected errors can be seen in the docstring, but it may be updated in https://github.com/optuna/optuna/pull/3429.

For example, 
- `RuntimeError` may be consistent with other component if the study with more than four objective (and `targets=None`).
- `ValueError` may be appropriate if the `targets` functions returns more than three values.

### Alternatives (optional)

_No response_

_No response_

nzw0301:
## Motivation

Resolve #3502.

## Description of the changes

- Add missing tests for errors of Plotly's `plot_pareto_front`
- Make docs more accurate in terns of the number of objectives
- Minor update of tests.

### DIFF
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -77,7 +77,7 @@ def plot_pareto_front(
     Args:
         study:
             A :class:`~optuna.study.Study` object whose trials are plotted for their objective
-            values. ``study.n_objectives`` must be eigher 2 or 3.
+            values. ``study.n_objectives`` must be either 2 or 3 when ``targets`` is :obj:`None`.
         target_names:
             Objective name list used as the axis titles. If :obj:`None` is specified,
             "Objective {objective_index}" is used instead. If ``targets`` is specified
@@ -103,7 +103,7 @@ def plot_pareto_front(
             A function that returns targets values to display.
             The argument to this function is :class:`~optuna.trial.FrozenTrial`.
             ``axis_order`` and ``targets`` cannot be used at the same time.
-            If your study has more than 4 objectives, ``targets`` must be specified.
+            If ``study.n_objectives`` is neither 2 nor 3, ``targets`` must be specified.
 
             .. note::
                 Added in v3.0.0 as an experimental feature. The interface may change in newer
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -342,7 +342,7 @@ def _make_json_compatible(value: Any) -> Any:
         json.dumps(value)
         return value
     except TypeError:
-        # the value can't be converted to JSON directly, so return a string representation
+        # The value can't be converted to JSON directly, so return a string representation.
         return str(value)
 
 
@@ -357,7 +357,6 @@ def _make_scatter_object(
 ) -> Union["go.Scatter", "go.Scatter3d"]:
     trials_with_values = trials_with_values or []
 
-    assert n_targets in (2, 3)
     marker = _make_marker(
         [trial for trial, _ in trials_with_values],
         include_dominated_trials,
@@ -374,8 +373,7 @@ def _make_scatter_object(
             marker=marker,
             showlegend=False,
         )
-    else:
-        assert n_targets == 3
+    elif n_targets == 3:
         return go.Scatter3d(
             x=[values[axis_order[0]] for _, values in trials_with_values],
             y=[values[axis_order[1]] for _, values in trials_with_values],
@@ -385,6 +383,11 @@ def _make_scatter_object(
             hovertemplate=hovertemplate,
             marker=marker,
             showlegend=False,
+        )
+    else:
+        raise ValueError(
+            "`plot_pareto_front` function only supports 2 or 3 targets."
+            " you used {} targets now.".format(n_targets)
         )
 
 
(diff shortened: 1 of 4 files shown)


HideakiImamura:

optuna/visualization/_pareto_front.py
```diff
@@ -386,6 +384,11 @@ def _make_scatter_object(
             marker=marker,
             showlegend=False,
         )
+    else:
+        raise ValueError(
+            "`plot_pareto_front` function only supports 2 or 3 targets."
+            " you used {} targets now.".format(n_targets)
+        )
```
c-bata, line 391: It looks that the number of targets are validated in `_get_pareto_front_info`. Thus these are unreachable lines. How about replacing these lines like the following?

```suggestion
    assert False, "Must not reach here"
```

tests/visualization_tests/matplotlib_tests/test_pareto_front.py
```diff
@@ -79,23 +79,25 @@ def test_plot_pareto_front_2d(
     plt.savefig(BytesIO())
 
     # Test with `target_names` argument.
-    with pytest.raises(ValueError):
+    error_message = "The length of `target_names` is supposed to be 2."
```
c-bata, line 82: I think tests should be passed even if error messages are changed. Error messages are just implementation details and it's out of specification of `plot_pareto_front()`. To reduce the maintainability of these tests, How about reverting these changes?

c-bata (changes requested):
Thank you for your pull request! I left some review comments.

tests/visualization_tests/test_pareto_front.py
```diff
@@ -370,6 +380,79 @@ def test_plot_pareto_front_invalid_axis_order(
...
+            targets=lambda t: (t.values[0], t.values[1], t.values[2]),
+        )
+
+
+@pytest.mark.parametrize(
+    "targets",
+    [
+        lambda t: t.values[0],
+    ],
+)
```
HideakiImamura, line 401: If we use just one `targets`, I think we don't have to parametrize it.

tests/visualization_tests/matplotlib_tests/test_pareto_front.py
HideakiImamura: I think there is value in maintaining the error messages that are displayed to users, as I have had users complain in the past when error messages were unexpectedly changed to something that was difficult to understand.

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM. I have some comments. PTAL.

optuna/visualization/_pareto_front.py
nzw0301: Alternatively, I'm wondering if we could move `ValueError` part into `_get_pareto_front_info` not to repeat the same error message in different files.

tests/visualization_tests/matplotlib_tests/test_pareto_front.py
nzw0301: Thank you for your discussion! I also cared about maintainability because this plot function has eight `ValueError` sources, so I found difficult it to which test case corresponds to which `ValueError` in `optuna/visualization/_pareto_front.py`.

optuna/visualization/_pareto_front.py
c-bata: Oh, I didn't notice that point :bow:

> Alternatively, I'm wondering if we could move ValueError part into _get_pareto_front_info not to repeat the same error message in different files.

Yes, I think that makes sense 👍

tests/visualization_tests/matplotlib_tests/test_pareto_front.py
c-bata: Thank you for your comments!

> as I have had users complain in the past when error messages were unexpectedly changed to something that was difficult to understand.

Hmm... In this case, the problem is that error messages are changed to something that was difficult to understand. It sounds that the change of error messages itself was not a problem.

> I found difficult it to which test case corresponds to which ValueError in optuna/visualization/_pareto_front.py. This is why I've (partially) added error message checks.

Thank you for sharing your opinion! I think we don't need to check whether each test case corresponds to which ValueError. It depends too much on the implementation details.

But anyway, as @HideakiImamura said, it might be a bit controversial. We should not deepen this discussion too much. I will approve this PR even if these changes won't be reverted :bow:

optuna/visualization/_pareto_front.py
nzw0301: Thanks! I've moved the value error part as in https://github.com/optuna/optuna/pull/3546/commits/a0a693351f7ba1f5c543d0fa9a922da30371b158.

c-bata (approved):
Thank you for the updates! LGTM!

HideakiImamura (approved):
Thanks for the update. LGTM.

PR #3552, opened 2022-05-12
Unify preprocessing of importance evaluators

contramundum53:
## Motivation
There has been duplicates in preprocessing code of `MeanDecreaseImpurityImportanceEvaluator` and `FanovaImportanceEvaluator`. We propose to refactor the code and remove the duplication.

## Description of the changes
* Refactor the preprocessing code into a function

Note that #3507 also gains advantage from this change.


g-votte:
Thanks for your PR.

Can't we rename [`_utils.py`](https://github.com/optuna/optuna/pull/3552/files#diff-38f08262bf8eeb4bdf67b52d395b4b13c593020321c6d0039399096df9e99a26) representing more specific responsibility? Once we named a module "utils," we'd put everything into utils, which confuses what the module does.

contramundum53:
Currently `_utils.py` contains a function for preprocessing and a function for postprocessing, which is supposed to be used in pair. Can you come up with a better name?

optuna/_transform.py
```diff
@@ -90,6 +90,14 @@ def column_to_encoded_columns(self) -> List[numpy.ndarray]:
     def encoded_column_to_column(self) -> numpy.ndarray:
         return self._encoded_column_to_column
 
+    @property
+    def num_params(self) -> int:
+        return len(self._column_to_encoded_columns)
+
+    @property
+    def num_features(self) -> int:
+        return len(self.encoded_column_to_column)
```
hvy, line 99: Seems like dead code. Are these lines left intentionally?

hvy:
Thanks for the updates. I'll try to go through it this week.
Could you just check the CI failures?

contramundum53:
The CI failures are about `shap.py`, which I haven't worked on yet. Let us first discuss whether this direction is viable, and after we have agreed I'll work on `shap.py`.

optuna/_transform.py
contramundum53: This property is added as the symmetric counterpart of `num_params`. Even though it isn't currently used in any implementations, it is likely that some `ImportanceEvaluator` will need it in the future.

hvy: Will it be used for `shap.py` perhaps? If it's not used by any of our evaluators, I'd say we should wait with this introduction until it become necessary (while I understand the symmetry). It's easier to add later. We could leave a TODO comment for future maintainers, if that'd be the case.

contramundum53: I don't think `shap.py` needs `num_features`.
I made it a TODO.

hvy:
I haven't dug into the details, but the overall approach looks great to me, moving validations to `get_param_importances`, leaving the evaluator interface intact and separating out logic to functions.
If we can fix `shap.py` too, it might become easier to see exactly what is DRY:ed and what is not.

optuna/_transform.py
hvy: Sorry for the nitpicking, but that looks good. Thanks. 🙇

contramundum53:
The changes have now been applied to `shap.py`.
It seems that `shap.py` previously did not support the cases where categorical parameters are present. I fixed the code to support it.

contramundum53:
Note that we have changed the interface of `BaseImportanceEvaluator.evaluate()` slightly such that they never take `None` for `params` and `target`. The default values for these arguments are filled in `get_param_importances()`.
This removes code duplication in `ImportanceEvaluator`s.

optuna/_transform.py
```diff
@@ -90,6 +90,12 @@ def column_to_encoded_columns(self) -> List[numpy.ndarray]:
     def encoded_column_to_column(self) -> numpy.ndarray:
         return self._encoded_column_to_column
 
+    @property
+    def num_params(self) -> int:
+        return len(self._column_to_encoded_columns)
```
c-bata, line 95: I have three suggestions.

1. [nits] `n_params` is more consistent with other variable names in Optuna (e.g. `n_trials` and `n_trees`).
2. `SearchSpaceTransform` transforms the search space between Optuna's search space representation and numpy-based representation. It is difficult to know `num_params` means which ones. `n_original_params` or `n_optuna_params` might be more clear. What do you think?
3. Not a strong opinion but `len(self._search_space)` is more intuitive I think.

```suggestion
    @property
    def n_original_params(self) -> int:
        return len(self._search_space)
```

Also could you update the Attributes section of docstring.

optuna/importance/_base.py
```diff
@@ -132,3 +137,51 @@ def _check_evaluate_args(completed_trials: List[FrozenTrial], params: Optional[L
...
+        if set(params) <= set(trial.params) and numpy.isfinite(float(target(trial)))
+    ]
+
+
+def _feature_importances_to_param_importances(
+    feature_importances: numpy.ndarray, trans: _SearchSpaceTransform
+) -> numpy.ndarray:
+    param_importances = numpy.zeros(trans.num_params)
+    numpy.add.at(param_importances, trans.encoded_column_to_column, feature_importances)
+    return param_importances
```
c-bata, line 168: I think this logic is not so common and a bit domain specific as it depends feature_importances representation of scikit-learn. We cannot make this logic DRY by moving this logic from `optuna/importance/_mean_decrease_impurity.py`.

How about moving this logic to `optuna/importance/_mean_decrease_impurity.py`?

contramundum53: I think this logic is as common as `_SearchSpaceTransform`, since what this function does is basically un-transform the importance values computed on the per-feature basis.

optuna/_transform.py
contramundum53: I thought the domain of `_SearchSpaceTransform` is called "parameters" (as they are the same as the user-specified parameters) and the range is called "features" (as in "one-hot features"). I guess adding a `n_features` property alongside with `n_params` would make it clear.

contramundum53: Another possible solution: make `self._search_space` public and remove `num_params` property.
What do you think?

optuna/importance/_base.py
c-bata: > I think this logic is as common as _SearchSpaceTransform

At least, this logic looks not duplicated now. This change would rather increase the lines of code than before. I think we should wait to try refactoring until we can gain the benefit of DRY. It helps us to avoid premature abstractions. What do you think?

optuna/_transform.py
c-bata: > I thought the domain of _SearchSpaceTransform is called "parameters" (as they are the same as the user-specified parameters) and the range is called "features" (as in "one-hot features").

I checked the docstring of `_SearchSpaceTransform` but I cannot find the term `features`. For example, the description of `column_to_encoded_columns` is `Constructed mapping from original parameter column index to encoded column indices.`. What do you think?

https://github.com/optuna/optuna/blob/93e23d6a7cd5ed126d669427884e409d15292196/optuna/_transform.py#L18-L57

c-bata: > Another possible solution: make self._search_space public and remove num_params property (which will be no more than len(trans.search_space)). What do you think?

We don't need to make `search_space` public since it is just referred at `n_original_params` (or `num_params`) property, that is defined in `_SearchSpaceTransform` class.

contramundum53: No, I mean instead of making a new `num_params` property, we can directly use `len(trans.search_space)` from `ImportanceEvaluator`s.

c-bata: > No, I mean instead of making a new num_params property, we can directly use len(trans.search_space) from ImportanceEvaluators.

Ah, that sounds make sense. I think we still don't need to make it public because `search_space` is a constructor argument.

optuna/importance/__init__.py
```diff
@@ -87,4 +89,29 @@ def get_param_importances(
...
+        params = list(intersection_search_space(study, ordered_dict=True).keys())
+
+    importances = evaluator.evaluate(study, params=params, target=target)
+
+    # Sort the importances in descending order.
+    return OrderedDict(
+        reversed(
+            sorted(importances.items(), key=lambda name_and_importance: name_and_importance[1])
+        )
+    )
```
nzw0301, line 117: It would be informative to address that the returned dictionary is sorted by its values descendingly in the `Retuned` part of the docstring based on this change.

optuna/importance/_base.py
```diff
@@ -19,9 +25,8 @@ class BaseImportanceEvaluator(object, metaclass=abc.ABCMeta):
     def evaluate(
         self,
         study: Study,
-        params: Optional[List[str]] = None,
-        *,
```
nzw0301, line 23: I suppose that this breaks the backward compatibility of user-defined importance evaluator. Could you revert this change: make `target` argument keyword-only?

I think the concrete evaluator class's `evaluate` method too.

optuna/importance/_base.py
```diff
@@ -132,3 +137,51 @@ def _check_evaluate_args(completed_trials: List[FrozenTrial], params: Optional[L
                     "Study must contain completed trials with all specified parameters. "
                     "Specified parameters: {}.".format(params)
                 )
+
+
+def _split_distributions(
```
nzw0301, line 142: Personaly, `_split_distributions` sounds vague. How about the following name?
```suggestion
def _split_distributions_to_single_and_nonsingle_distributions(
```
or `_split_distributions_to_single_and_nonsingle`.

nzw0301 (changes requested):
Basically, the approach looks great. I have a few comments on docstring and backward compatibility.

optuna/importance/_base.py
contramundum53: Making `target` keyword-only is okay, but since we don't want to have the default argument `None`, we are breaking the backward compatibility anyways.
How about making the interface clean for this time?

contramundum53: The backward compatibility of types will be broken, but the runtime backward compatibility is maintained, and the backward compatibility of types will be broken anyways even if we make `target` keyword-only.

contramundum53:
We reverted all breaking changes and split them into #3589.

contramundum53:
@c-bata @nzw0301 PTAL 🙇‍♂️

optuna/importance/_base.py
```diff
@@ -44,13 +50,8 @@ def evaluate(
...
                 If it is :obj:`None` and ``study`` is being used for single-objective optimization,
-                the objective values are used. Can also be used for other trial attributes, such as
-                the duration, like ``target=lambda t: t.duration.total_seconds()``.
-
-                .. note::
-                    Specify this argument if ``study`` is being used for multi-objective
-                    optimization. For example, to get the hyperparameter importance of the first
-                    objective, use ``target=lambda t: t.values[0]`` for the target parameter.
+                the objective values are used. ``target`` must be specified if ``study`` is being
+                used for multi-objective optimization.
```
nzw0301, line 54: if we delete the existing explanation of `target`, I think that we need to clarify that `target` returns a single float value.

optuna/importance/_mean_decrease_impurity.py
```diff
@@ -64,49 +62,36 @@ def evaluate(
...
         feature_importances = forest.feature_importances_
-        feature_importances_reduced = numpy.zeros(len(distributions))
-        numpy.add.at(feature_importances_reduced, encoded_column_to_column, feature_importances)
 
-        param_importances = OrderedDict()
-        self._param_names = list(distributions.keys())
-        for i in feature_importances_reduced.argsort()[::-1]:
-            param_importances[self._param_names[i]] = feature_importances_reduced[i].item()
+        # Untransform feature importances to param importances
+        #  by adding up relevant feature importances
```
nzw0301, line 93:
```suggestion
        # by adding up relevant feature importances.
```

optuna/samplers/_search_space/intersection.py
```diff
@@ -90,7 +96,14 @@ def calculate(self, study: Study, ordered_dict: bool = False) -> Dict[str, BaseD
...
         self._cursor = next_cursor
-        search_space = self._search_space or {}
+
+        if self._search_space is None:
+            if allow_empty_study:
+                search_space = {}
+            else:
+                raise ValueError("Cannot determine intersection search space for an empty study.")
+        else:
+            search_space = self._search_space
```
nzw0301, line 106: Can we simplify this part as follows?
```suggestion

        if self._search_space is None and not allow_empty_study:
            raise ValueError("Cannot determine intersection search space for an empty study.")

        search_space = self._search_space or {}
```

optuna/samplers/_search_space/intersection.py
```diff
@@ -33,7 +33,9 @@ def __init__(self, include_pruned: bool = False) -> None:
 
         self._include_pruned = include_pruned
 
-    def calculate(self, study: Study, ordered_dict: bool = False) -> Dict[str, BaseDistribution]:
+    def calculate(
+        self, study: Study, ordered_dict: bool = False, allow_empty_study: bool = True
```
nzw0301, line 37: I'm not sure why this new argument is necessary for this PR. Could you elaborate on that?

nzw0301 (changes requested):

optuna/samplers/_search_space/intersection.py
contramundum53: Sorry, I think I forgot to revert this part.

optuna/importance/_base.py
contramundum53: Sorry, I forgot to revert this part.

tests/importance_tests/test_init.py
```diff
@@ -263,24 +262,6 @@ def objective(trial: Trial) -> float:
...
-    study = create_study()
-    study.optimize(objective, n_trials=3)
-
-    with pytest.raises(TypeError):
-        get_param_importances(study, evaluator=evaluator_init_func(), params={})  # type: ignore
-
-    with pytest.raises(TypeError):
-        get_param_importances(study, evaluator=evaluator_init_func(), params=[0])  # type: ignore
-
-
```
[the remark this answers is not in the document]
contramundum53: In our new test policy, we do not test arguments for invalid types.

contramundum53:
Applied @nzw0301-san's comments.

contramundum53:
Applied @nzw0301 -san's comments.

tests/importance_tests/test_init.py
```diff
@@ -94,7 +97,6 @@ def objective(trial: Trial) -> float:
             assert isinstance(importance, float)
             assert importance <= prev_importance
             prev_importance = importance
-        assert math.isclose(1.0, sum(i for i in param_importance.values()), abs_tol=1e-5)
```
c-bata, line 97: Is it necessary to remove these assertions even though `parametrize_evaluator` does not contain ShapleyImportanceEvaluator`?

tests/importance_tests/test_mean_decrease_impurity.py
```diff
@@ -31,10 +32,10 @@ def test_mean_decrease_impurity_importance_evaluator_n_trees() -> None:
     study.optimize(objective, n_trials=3)
 
     evaluator = MeanDecreaseImpurityImportanceEvaluator(n_trees=10, seed=0)
-    param_importance = evaluator.evaluate(study)
```
c-bata, line 34: In this test file, I think calling `evaluator.evalute(study)` directly is good approach to focus the behavior of `MeanDecreaseImpurityImportanceEvaluator`. What do you think?

c-bata: Changes in `tests/integration_tests/test_shap.py` and `tests/importance_tests/test_fanova.py` are also.

tests/importance_tests/test_init.py
contramundum53: If you decide to remove the specification that importance values should add up to 1, I think we should also remove them from existing tests.

tests/importance_tests/test_mean_decrease_impurity.py
contramundum53: Hmm, I don't think we should test interfaces that are deemed private (not to be called directly by user) in the document.
But if you consider them public, then we should test that also I think.

contramundum53: Maybe we should also discuss that in #3589.

tests/importance_tests/test_init.py
contramundum53: But maybe this could be left for another PR.

tests/importance_tests/test_init.py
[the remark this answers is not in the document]
contramundum53: But I think in the current state, leaving this test would not break CI. We'll revisit this in subsequent PRs.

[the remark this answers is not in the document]
nzw0301: Ah. Sorry, I didn't know the pocily!

tests/importance_tests/test_mean_decrease_impurity.py
c-bata: > Hmm, I don't think we should test interfaces that are deemed private (not to be called directly by user) in the document. But if you consider them public, then we should test that also I think.

Hm, I don't think that. Given that logic, we cannot write unittests for storage as well. It doesn't important that the interface is public to end-users.

contramundum53:
We have remade this PR with minimum changes in #3597.

PR #3554, opened 2022-05-12
Reduce the number of `copy.deepcopy()` calls in `importance` module.

c-bata:
## Motivation

Now I'm trying to implement Cython optimized fANOVA implementation. The source code is still in private, but these lines are one of the bottlenecks as `copy.deepcopy()` is called twice internally.

## Description of the changes

* Use `study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))` instead of `study.trials`.


nzw0301 (approved):
Nice! LGTM.

not522 (approved):
LGTM!

Issue #3206, opened 2021-12-24
`inf` handling in Optuna


HideakiImamura:
## Expected behavior

Regardless of the type of storage used, the trial accepts `inf` for its value when an objective function returns `inf`, when `inf` is passed in `trial.report`, or when `inf` is passed in `study.tell`. Note that it is the responsibility of the sampler or pruner to decide how to use the `inf` values stored in the storage. For example, `TPESampler`, the default sampler in Optuna, is implemented to be able to use `inf`.

We need to fix the followings.
- [x] If `inf` is returned by objective function, the study accepts it regardless of the type of storage used.
- [x] If `inf` is passed in `trial.report`, the study accepts it regardless of the type of storage used.
- [x] If `inf` is passed in `study.tell`, the study accepts it regardless of the type of storage used.

## Environment

- Optuna version: a70cbff8dbc454f137b61622633404bd4e4657fc
- Python version: 3.9.9
- OS: macOS Catalina 10.15.7
- (Optional) Other libraries and their versions:
psycopg2:    2.9.2
PyMySQL:     1.0.2

## Error messages, stack traces, or logs

If we run the script `inf-objective.py`, we will get the following output. The script is provided in the section of `Reproducible examples`. I think we need to modify this behavior as explained in `Expected behavior`. Concretely, we should fix the behavior with MySQL to accept the `inf`.
<details>
<summary>output</summary>

```
(venv) mamu@HideakinoMacBook-puro test % python inf-objective.py 
[I 2021-12-24 13:58:54,771] A new study created in memory with name: no-name-f7f28497-2f1c-4739-b9c8-edca726274fc
[I 2021-12-24 13:58:54,772] Trial 0 finished with value: inf and parameters: {}. Best is trial 0 with value: inf.
In-memory : Trial state is TrialState.COMPLETE
[I 2021-12-24 13:58:54,843] A new study created in RDB with name: no-name-168f23c0-0ac5-4785-a6d6-bdfd31dda5ab
[I 2021-12-24 13:58:54,909] Trial 0 finished with value: inf and parameters: {}. Best is trial 0 with value: inf.
SQLite    : Trial state is TrialState.COMPLETE
[I 2021-12-24 13:58:55,052] A new study created in RDB with name: no-name-22068297-6685-4435-bf88-124621f797c1
Record does not exist.
MySQL     : Trial state is TrialState.RUNNING
[...]

PR #3559, opened 2022-05-13
Handling inf/-inf for `trial_values` table in RDB

contramundum53:
## Motivation
The value of INF is supported in optuna, but is not supported in some databases, and this causes error when using RDBStorage. Previously this was worked around by replacing inf by a very large value, but it is nice to be able to handle inf exactly. (See #3206)

## Description of the changes
* Add a column in database that specifies whether the contained value is finite or +inf or -inf.
* Add a migration script for this change.


contramundum53:
Currently the migration script is buggy for PostgreSQL. We are working on this issue.

HideakiImamura:
@contramundum53 Is this PR ready for review? I'm not sure but, we should merge the change of `intermediate_value` at first (the PR will be submitted by @c-bata), and then work on this PR, shouldn't we?

contramundum53:
#3603 is merged and this PR is ready for review. 

optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
```diff
@@ -0,0 +1,167 @@
...
+        session.bulk_update_mappings(TrialValueModel, mapping)
+        session.commit()
+    except SQLAlchemyError as e:
+        session.rollback()
+        raise e
+    finally:
+        session.close()
+
+    with op.batch_alter_table("trial_values", schema=None) as batch_op:
+        batch_op.drop_column("value_type")
```
c-bata, line 165: I think we should make `value` column `nullable=False` here.

c-bata (commented):
Changes almost looks good 👍  I'll check the migration script on optuna-e2e.

optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
contramundum53: Sorry, we didn't share the discussion with you. When `value_type` is not `FINITE`, we put `NULL` in `value` column. In discussion among me, @toshihikoyanase and @HideakiImamura, we agreed that that is more easy to understand than setting `value` as non-nullable and putting e.g. 0.0.

optuna/storages/_rdb/storage.py
```diff
@@ -786,29 +787,6 @@ def get_trial_param(self, trial_id: int, param_name: str) -> float:
...
-
-        # Floats can't be compared for equality because they are
-        # approximate and not stored as exact values.
-        # https://dev.mysql.com/doc/refman/8.0/en/problems-with-float.html
-        if value is None:
-            return float("nan")
-        elif np.isclose(value, _RDB_MIN_FLOAT) or np.isclose(value, _RDB_MAX_FLOAT):
-            return float(np.sign(value) * float("inf"))
-        return value
-
```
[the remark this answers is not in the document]
contramundum53: Thanks! I forgot to remove that. 
It is fixed now.

optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
toshihikoyanase: I may miss the point, but `value` does not accept `null` before we migrate the schema, so we should revert `nullable` of `value` in this `downgrade` function.

contramundum53: Sorry, I was misunderstood. Thanks for the clarification!

optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
```diff
@@ -0,0 +1,167 @@
...
+        records = session.query(TrialValueModel).all()
+        mapping = []
+        for r in records:
+            value: float
+            if np.isclose(r.value, RDB_MAX_FLOAT) or np.isposinf(r.value):
+                value = float("inf")
+            elif np.isclose(r.value, RDB_MIN_FLOAT) or np.isneginf(r.value):
+                value = float("-inf")
+            elif np.isnan(r.value):
+                value = float("nan")
```
toshihikoyanase, line 110: [Notes] Trial values do not accept `NaN`, but it may happen to be stored in the storage (e.g., during trial-and-error of optuna codebase). This code simply pass such NaN value as `FINITE` value type. I guess it is acceptable, but if we strictly reject `NaN` value, we should raise errors here.

toshihikoyanase: I talked with @contramundum53 and @HideakiImamura , and they proposed to remove this `nan` handling since `nan` value is not expected here.

contramundum53: Thank you for pointing that out! I forgot to remove those lines.
I fixed it now.

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -42,6 +42,7 @@ def test_init() -> None:
 
     assert storage.get_current_version() == storage.get_head_version()
     assert storage.get_all_versions() == [
+        "v3.0.0.d",
```
toshihikoyanase, line 45: I guess we need to update the following version lists:
- https://github.com/optuna/optuna/blob/ebbf70a608e3eab52f882befc3b65a7ba52d1711/tests/storages_tests/rdb_tests/test_storage.py#L176-L179
- https://github.com/optuna/optuna/blob/ebbf70a608e3eab52f882befc3b65a7ba52d1711/tests/storages_tests/rdb_tests/test_storage.py#L224
- https://github.com/optuna/optuna/blob/ebbf70a608e3eab52f882befc3b65a7ba52d1711/tests/storages_tests/rdb_tests/test_storage.py#L268

optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
toshihikoyanase: Thanks!

optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
```diff
@@ -0,0 +1,172 @@
...
+    finally:
+        session.close()
+
+    with op.batch_alter_table("trial_values", schema=None) as batch_op:
+        batch_op.drop_column("value_type")
+        batch_op.alter_column(
+            "value",
+            existing_type=sa.Float(precision=FLOAT_PRECISION),
+            existing_nullable=True,
+            nullable=False,
```
toshihikoyanase, line 169: `existing_nullable` is not required because we also set `nullable` in this function. c.f., https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.Operations.alter_column.params.existing_nullable

>  Optional; the existing nullability of the column. Required on MySQL **if the existing nullability is not being changed;** else MySQL sets this to NULL.
```suggestion
            nullable=False,
```

tests/storages_tests/rdb_tests/test_storage.py
contramundum53: Thanks! I added "3.0.0.d".

toshihikoyanase:
I checked newly created tables using optuna-e2e. I used the following objective function, and it worked as expected.

```python
import optuna

    x1 = trial.suggest_float("x1", 0, 10)
    x2 = trial.suggest_float("x2", 1, 10, log=True)
    x3 = trial.suggest_float("x3", 0, 10, step=2.0)
    y1 = trial.suggest_int("y1", 0, 10)
    y2 = trial.suggest_int("y2", 1, 10, log=True)
    y3 = trial.suggest_int("y3", 0, 10, step=2)
    c1 = trial.suggest_categorical("c1", [1, 10, 100])
    trial.report(value=1, step=1)

    if trial.number % 3 == 0:
        return float("inf")
    elif trial.number % 3 == 1:
        return float("-inf")

    return x1**y1 + x2**y2 + x3**y3 + c1

```

<details>
<summary>SQLite3</summary>

```console
echo .dump | sqlite3 ./data/sample.db
```

```sql
CREATE TABLE trial_values (
	trial_value_id INTEGER NOT NULL, 
	trial_id INTEGER NOT NULL, 
	objective INTEGER NOT NULL, 
	value FLOAT, 
	value_type VARCHAR(7) NOT NULL, 
	PRIMARY KEY (trial_value_id), 
	UNIQUE (trial_id, objective), 
	FOREIGN KEY(trial_id) REFERENCES trials (trial_id)
);
INSERT INTO trial_values VALUES(1,1,0,NULL,'INF_POS');
INSERT INTO trial_values VALUES(2,2,0,NULL,'INF_NEG');
INSERT INTO trial_values VALUES(3,3,0,101718548.10233116149,'FINITE');
```

</details>

<details>
<summary>MySQL</summary>

```console
docker compose run --rm mysql mysqldump --skip-column-statistics -u root -p -h mysql  -B optuna
```

```sql
CREATE TABLE `trial_values` (
  `trial_value_id` int NOT NULL AUTO_INCREMENT,
  `trial_id` int NOT NULL,
  `objective` int NOT NULL,
  `value` double DEFAULT NULL,
  `value_type` enum('FINITE','INF_POS','INF_NEG') NOT NULL,
  PRIMARY KEY (`trial_value_id`),
  UNIQUE KEY `trial_id` (`trial_id`,`objective`),
  CONSTRAINT `trial_values_ibfk_1` FOREIGN KEY (`trial_id`) REFERENCES `trials` (`trial_id`)
) ENGINE=InnoDB AUTO_INCREMENT=14 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;

INSERT INTO `trial_values` VALUES (1,1,0,NULL,'INF_POS'),(2,2,0,NULL,'INF_NEG'),(3,3,0,89.19247874796469,'FINITE'), ...
```

</details>


<details>
<summary>PostgreSQL</summary>

```console
docker compose run --rm postgresql pg_dump -h postgresql -U root --password optuna
```

```sql
CREATE TYPE public.trialvaluetype AS ENUM (
    'FINITE',
    'INF_POS',
    'INF_NEG'
);

CREATE TABLE public.trial_values (
    trial_value_id integer NOT NULL,
    trial_id integer NOT NULL,
    objective integer NOT NULL,
    value double precision,
    value_type public.trialvaluetype NOT NULL
);

COPY public.trial_values (trial_value_id, trial_id, objective, value, value_type) FROM stdin;
1       1       0       \N      INF_POS
2       2       0       \N      INF_NEG
3       3       0       222.32992747318895      FINITE
```

</details>

tests/storages_tests/rdb_tests/test_storage.py
toshihikoyanase: Thank you!

optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
```diff
@@ -0,0 +1,170 @@
...
+        batch_op.add_column(
+            sa.Column(
+                "value_type",
+                sa.Enum("FINITE", "INF_POS", "INF_NEG", name="trialvaluetype"),
+                nullable=False,
+                server_default="FINITE",
+            ),
+        )
+    with op.batch_alter_table("trial_values") as batch_op:
+        batch_op.alter_column("value_type", server_default=None)
```
toshihikoyanase, line 91: I thought MySQL requires `existing_type` and `existing_nullable`, but the current code worked as expected in my environment.
c.f., https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.Operations.alter_column.params.existing_type


I used the `mysql:8.0.25` docker image with [optuna-e2e](https://github.com/himkt/optuna-e2e).

```sql
CREATE TABLE public.trial_values (
    trial_value_id integer NOT NULL,
    trial_id integer NOT NULL,
    objective integer NOT NULL,
    value double precision,
    value_type public.trialvaluetype NOT NULL
);

```

toshihikoyanase (commented):
The change basically looks good to me except for omitted arguments of `alter_column`. I'd like to hear other reviewers'  comments on 
 https://github.com/optuna/optuna/pull/3559/files#r887718310.

optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
HideakiImamura: I also confirmed that the migration has passed, but I cannot understand the consistency with the documentation.
 🤔  

We have already missed `existing_nullable` in the migration script [here](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/alembic/versions/v2.6.0.a_.py#L23).

HideakiImamura (commented):
Thanks, basically LGTM. I would like to resolve the question by @toshihikoyanase.

optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
contramundum53: I also updated the migration code in v3.0.0.c, but left all previous code.

toshihikoyanase: Thank you for your update. Honestly, I prefer separated PRs to change the different migration scripts. But, it sounds reasonable for this case due to the release schedule. I checked both trial model and trial intermediate model and I'll share the result.

toshihikoyanase:
@HideakiImamura Please merge after your review.

HideakiImamura (approved):
Thanks for the long running work! LGTM!

PR #3517, opened 2022-04-27
Cleanup setuppy

himkt:
To cleanup `setup.py`, I propose the following refactoring:

- remove `experimental`. It's not used
- unify `document` and `doctest`. They always co-occur
- unify `tests` and `optional` and create `develop`
- unify `testing` and `integration`. A lot of overlap

## Motivation
🔗 https://github.com/optuna/optuna/issues/1576

Issue #2639, opened 2021-05-03
FloatingPointError due to underflow encountered in exp in TPE sampler since latest version


hugocool:
Since upgrading to the latest version optuna, an optimisation which used to work now doesn't even though it is the same dataset.
I get the same error on a different dataset and have included the code where i encounter the problem below.
The error does not happen in first ten trials, however after enough trials this error is encountered.

## Expected behavior

I expect the optuna TPE sampler to work, just like it did before.

## Environment

- Optuna version: 2.7.0
- Python version: 3.7
- OS: linux (google colab)
- xgboost version : 1.4.1

## Error messages, stack traces, or logs

```
[I 2021-05-03 20:06:30,229] Trial 41 finished with value: 58.96419669999999 and parameters: {'booster': 'gbtree', 'lambda': 0.10473082821492902, 'alpha': 0.0034424292875279952, 'subsample': 0.08381812656141696, 'max_depth': 1, 'min_child_weight': 15, 'eta': 0.33668240574729064, 'gamma': 0.003057081444431449, 'grow_policy': 'depthwise'}. Best is trial 9 with value: 55.697677600000006.
[I 2021-05-03 20:06:30,707] Trial 42 pruned. Trial was pruned at iteration 0.
[W 2021-05-03 20:06:30,716] Trial 43 failed because of the following error: FloatingPointError('underflow encountered in exp')
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/optuna/_optimize.py", line 217, in _run_trial
    value_or_values = func(trial)
  File "<ipython-input-22-46c8a6d0e2a1>", line 27, in objective
    "subsample": trial.suggest_float("subsample", 0, 1.0),
  File "/usr/local/lib/python3.7/dist-packages/optuna/trial/_trial.py", line 178, in suggest_float
    return self.suggest_uniform(name, low, high)
  File "/usr/local/lib/python3.7/dist-packages/optuna/trial/_trial.py", line 237, in suggest_uniform
    return self._suggest(name, distribution)
  File "/usr/local/lib/python3.7/dist-packages/optuna/trial/_trial.py", line 717, in _suggest
    study, trial, name, distribution
  File "/usr/local/lib/python3.7/dist-packages/optuna/samplers/_tpe/sampler.py", line 288, in sample_independent
    return self._sample_uniform(param_distribution, below_param_values, above_param_values)
[...]

PR #3560, opened 2022-05-13
Remove SciPy from the install requirement

### LINKED ISSUE #3035: Lightweight install option?
I'd like to use Optuna for a project, but the installed package with dependencies comes to a whopping 174MB. This is an issue for my application since I'd like to use it in AWS Lambda, which has a hard cap of 250MB for all code in a lambda, and some other packages I'm using consume 90MB (bringing me to 260MB).

Is there a lightweight install option for Optuna, a way to strip out any optional functionality? Or any other suggestions on how to get the size down?

Thanks!

not522:
## Motivation
SciPy requires relatively large storage and memory. It causes trouble in some constrained environments. E.g., This PR resolves #3035.

The refactoring of `setup.py` is going in #3517. I will update this PR after #3517 is merged.

## Description of the changes
Remove SciPy from the install requirement

### DIFF
--- a/optuna/samplers/_tpe/parzen_estimator.py
+++ b/optuna/samplers/_tpe/parzen_estimator.py
@@ -1,23 +1,14 @@
+import math
 from typing import Callable
 from typing import Dict
 from typing import NamedTuple
 from typing import Optional
 from typing import Tuple
-from typing import TYPE_CHECKING
 
 import numpy as np
 
 from optuna import distributions
-from optuna._imports import _LazyImport
 from optuna.distributions import BaseDistribution
-
-
-if TYPE_CHECKING:
-    import scipy.special as special
-    import scipy.stats as stats
-else:
-    special = _LazyImport("scipy.special")
-    stats = _LazyImport("scipy.stats")
 
 
 EPS = 1e-12
@@ -123,20 +114,15 @@ class _ParzenEstimator:
                 assert sigmas is not None
 
                 # We sample from truncnorm.
-                trunc_low = (low - mus[active]) / sigmas[active]
-                trunc_high = (high - mus[active]) / sigmas[active]
-                samples = np.full((), fill_value=high + 1.0, dtype=np.float64)
-                while (samples >= high).any():
+                samples = np.full(size, fill_value=high + 1.0, dtype=np.float64)
+                while not (low <= samples).all() or not (samples < high).all():
                     samples = np.where(
-                        samples < high,
+                        (low <= samples) & (samples < high),
                         samples,
-                        stats.truncnorm.rvs(
-                            trunc_low,
-                            trunc_high,
+                        rng.normal(
                             size=size,
                             loc=mus[active],
                             scale=sigmas[active],
-                            random_state=rng,
                         ),
                     )
             samples_dict[param_name] = samples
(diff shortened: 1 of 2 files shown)


c-bata:
@HideakiImamura @contramundum53 Could you review this PR?

not522:
My implementation of truncnorm is too slow in some cases, and it causes the CI to get stuck. I'm working to fix it.

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -448,7 +436,7 @@ def _normal_cdf(x: np.ndarray, mu: np.ndarray, sigma: np.ndarray) -> np.ndarray:
         denominator = x - mu
         numerator = np.maximum(np.sqrt(2) * sigma, EPS)
         z = denominator / numerator
-        return 0.5 * (1 + special.erf(z))
+        return 0.5 * (1 + np.vectorize(math.erf)(z))
```
contramundum53, line 439: FYI, I micro-benchmarked this line and got the following results:
```
In [10]: xs = np.linspace(-2, 2, 100000)

In [11]: %timeit np.vectorize(math.erf)(xs)
6.5 ms ± 35.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [12]: %timeit special.erf(xs)
631 µs ± 36.7 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

`np.vectorize(math.erf)` could be 10 times slower than `special.erf`. This could be or not be a problem.

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -200,8 +186,10 @@ def log_pdf(self, samples_dict: Dict[str, np.ndarray]) -> np.ndarray:
                     )
                     log_pdf = np.log(cdf + EPS) - np.log(p_accept + EPS)
             component_log_pdf += log_pdf
-        ret = special.logsumexp(component_log_pdf + np.log(self._weights), axis=1)
-        return ret
+        weighted_log_pdf = component_log_pdf + np.log(self._weights)
+        max_ = weighted_log_pdf.max(axis=1)
+        with np.errstate(divide="ignore"):
```
contramundum53, line 191: Why was this necessary, when there doesn't seem to be a case where division by zero is involved?

not522: It follows the SciPy's implementation. I will add a comment why `np.errstate` is necessary. https://github.com/scipy/scipy/blob/7076a90ff42a2349c621e1a081fbf5f715c421c0/scipy/special/_logsumexp.py#L112-L113

contramundum53: Is there a case where that would incur a division by zero?
(Except when `self._weights` are all zero, which would be invalid anyways)

not522: It introduced to resolve inf related bug. It never occurs in our situation, so we can remove `np.errstate`. Thank you for careful review!
https://github.com/scipy/scipy/issues/4014
https://github.com/scipy/scipy/pull/4016

not522: By the way, `np.errstate(under='ignore')` may be preferred. #2639

contramundum53:
Other than my comment and https://github.com/optuna/optuna/pull/3560#issuecomment-1127231048, this PR looks good to me. I will review this again after https://github.com/optuna/optuna/pull/3560#issuecomment-1127231048 is fixed.

contramundum53:
For https://github.com/optuna/optuna/pull/3560#issuecomment-1127231048, I think sampling from a uniform distribution and evaluating its inverse error function will suffice. However, I could not find an implementation of inverse error function without relying on `scipy.special.erfinv`, so I think we need to bring its implementation to optuna (which can be a little tough).

contramundum53:
I took a benchmark of the above code, compared with scipy.
```
l, r = -0.001, 0.001
sample_trunc_normal(l, r, size=1000, oversample_rate=2) 
# ==> 62.8 µs ± 279 ns
stats.truncnorm(l, r).rvs(size=1000)
# ==> 316 µs ± 19 µs
```

Issue #3556, opened 2022-05-13
`GridSampler` does not stop the optimization when NaN in the `search_space`


not522:
### Expected behavior

`GridSampler` should stop the optimization when all grids are evaluated.

### Environment

- Optuna version: 3.0.0b1.dev
- Python version: 3.8.6
- OS: macOS-10.16-x86_64-i386-64bit
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
See steps to reproduce.
```

### Steps to reproduce

In the following code, `optimize` should stop with 2 trials, but it infinitely creates trials.

```python
import optuna
sampler = optuna.samplers.GridSampler({"x": [0, float("nan")]})
study = optuna.create_study(sampler=sampler)
study.optimize(lambda _: 0)
```

### Additional context (optional)

The bug is coded here.

https://github.com/optuna/optuna/blob/c23e17c70bd77489e223f3d4c322d86ec5a6ec66/optuna/samplers/_grid.py#L270

If the `param_value` is NaN, this condition is wrong.

```python
float("nan") != float("nan")  # True
```

contramundum53:
I suspect that there are other similar bugs in optuna (related to nan not comparing equal to itself), and that this kind of bug needs a fundamental and comprehensive fix (if they are to be fixed). I think it is a good opportunity for us to thoroughly list up similar bugs before making this a good first issue.

contramundum53:
I listed up all (in)equality comparisons of `float` values (or objects containing `float` values).

```
15 results - 11 files

optuna/distributions.py:
  377          value = param_value_in_internal_repr
  378:         return self.low <= value <= self.high and (value - self.low) % self.step == 0 #!!
  379  

  614          return
  615:     if dist_old.choices != dist_new.choices: #!!
  616          raise ValueError(

  723  
  724:     if new_distribution != distribution and not suppress_warning: #!!
  725          message = (

optuna/importance/_base.py:
  93          if any(
  94:             trial_distributions[name] != distribution #!!
  95              for name, distribution in distributions.items()

optuna/importance/_fanova/_fanova.py:
  77  
  78:         if all(tree.variance == 0 for tree in self._trees): #!!
  79              # If all trees have 0 variance, we cannot assess any importances.

optuna/multi_objective/trial.py:
  353  
  354:         if values0 == values1: #!!
  355              return False

optuna/samplers/_grid.py:
  269              for i, param_value in enumerate(sorted(param_values)):
[...]

PR #3567, opened 2022-05-17
Fix `CategoricalDistribution` with NaN

not522:
## Motivation
Thanks to @contramundum53's investigation, I found that `CategoricalDistribution` is buggy with NaN.  https://github.com/optuna/optuna/issues/3556#issuecomment-1128247623

## Description of the changes
Fix `CategoricalDistribution`'s `__eq__` and `check_distribution_compatibility` to handle NaN.


tests/test_distributions.py
```diff
@@ -119,7 +122,7 @@ def test_check_distribution_compatibility() -> None:
     # test the same distribution
     for key in EXAMPLE_JSONS:
         distributions.check_distribution_compatibility(
-            EXAMPLE_DISTRIBUTIONS[key], EXAMPLE_DISTRIBUTIONS[key]
+            EXAMPLE_DISTRIBUTIONS[key], distributions.json_to_distribution(EXAMPLE_JSONS[key])
```
not522, line 125: We need to create new objects to compare NaNs.

>The built-in containers typically assume identical objects are equal to themselves.

https://docs.python.org/3/reference/expressions.html#value-comparisons

```python
>>> x = float("nan")
>>> y = float("nan")
>>> x == x
False
>>> x == y
False
>>> [x] == [x]
True
>>> [x] == [y]
False
```
Note that `copy.deepcopy` does not create new object because `float("nan")` is immutable.
```python
>>> [x] == [copy.deepcopy(x)]
True
```

tests/test_distributions.py
```diff
@@ -10,6 +10,9 @@
 from optuna import distributions
 
 
+_choices = (None, True, False, 0, 1, 0.0, 1.0, float("nan"), float("inf"), -float("inf"), "", "a")
```
not522, line 13: Note: `CategoricalChoiceType` is `Union[None, bool, int, float, str]`. `_choices` contains all types and their edge cases.

not522:
We should fix this line, too. I'm investigating how to fail the master's implementation in the unit test.
https://github.com/optuna/optuna/blob/b099d232181dfceb3d110342ce567a35708ee615/optuna/distributions.py#L497

not522:
Fixed. Now, this PR is ready for review.

c-bata:
@knshnb Could you review this PR if you have the time?

optuna/distributions.py
```diff
@@ -450,6 +453,12 @@ def _asdict(self) -> Dict:
         return d
 
 
+def _nanequal(value1: Any, value2: Any) -> bool:
```
c-bata, line 456: [nits] Not a strong opinion but how about naming this `_param_value_equal()`? `_nanequal` sounds like just checking both of two values are nan or not for me.

optuna/distributions.py
```diff
@@ -450,6 +453,12 @@ def _asdict(self) -> Dict:
...
+    value1_is_nan = isinstance(value1, Real) and np.isnan(float(value1))
+    value2_is_nan = isinstance(value2, Real) and np.isnan(float(value2))
```
c-bata, line 458: Can we write these lines like below?

```suggestion
    value1_is_nan = isinstance(value1, float) and np.isnan(value1)
    value2_is_nan = isinstance(value2, float) and np.isnan(value2)
```

optuna/distributions.py
```diff
@@ -509,6 +517,32 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
...
+                        return False
+            else:
+                if value != other_value:
+                    return False
+        return True
+
+    def __hash__(self) -> int:
+
+        return hash((self.__class__,) + tuple(sorted(self.__dict__.items())))
+
```
c-bata, line 545: I'm wondering if we can make these lines simple (and reduce the code sizes) by moving the logic of L526-L540 to `check_distribution_compatibility()` function. Do you think there is a benefit that overloading CategoricalDistribution's comparison operators?

c-bata (commented):
Thank you for your pull request. I left some review comments.

optuna/distributions.py
not522: `CategoricalDistribution`'s comparison operator is called from other codes. So, we need to overload `__eq__` to fix the comparison with NaN.

https://github.com/optuna/optuna/blob/bf8ec607095a4b1d320b1171fa61227e9bf1366a/optuna/distributions.py#L724
https://github.com/optuna/optuna/blob/bf8ec607095a4b1d320b1171fa61227e9bf1366a/optuna/importance/_base.py#L92
https://github.com/optuna/optuna/blob/bf8ec607095a4b1d320b1171fa61227e9bf1366a/optuna/samplers/_search_space/intersection.py#L86
https://github.com/optuna/optuna/blob/bf8ec607095a4b1d320b1171fa61227e9bf1366a/optuna/trial/_trial.py#L647

optuna/distributions.py
not522: I'm not sure if we should support `Real` types other than `float`. This PR can handle other NaNs like `np.float32("nan")`, but I think it is very rare to suggest such values. What do you think?

optuna/distributions.py
c-bata: Thank you for your reply! I see. It sounds make sense to overload the operator.

optuna/distributions.py
c-bata: Thanks! Such cases (e.g. `np.float32("nan")`) completely slipped my mind. Let me resolve this comment.

optuna/distributions.py
```diff
@@ -509,6 +517,32 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
...
+                    return False
+                for choice, other_choice in zip(value, other_value):
+                    if not _nanequal(choice, other_choice):
+                        return False
+            else:
+                if value != other_value:
+                    return False
+        return True
+
+    def __hash__(self) -> int:
```
knshnb, line 542: This is might be a silly question on Python, but when is this method called?

knshnb (commented):
Thanks for the PR!
I'm not sure I fully understand the background, but how about making `CategoricalChoiceType` a class and defining `__eq__()` for that? It is used in other places and calling `_nanequal()` if we want to compare them might not be clear. I think, this way, we can also make `CategoricalDistribution.__eq__()` simpler (basically just compare `self.__dict__` and `other.__dict__`).

knshnb:
I'm sorry, I had a misunderstanding, and please ignore the above suggestion.
It is possible to define a new class like `_CategoricalChoice` with `__eq__()` and let `CategoricalDistribution.choices` be `Sequence[_CategoricalChoice]` while keeping the original input of `Sequence[CategoricalChoiceType]`. However, `CategoricalDistribution.choices` doesn't seem to be touched from other places, so I think the current implementation is still fine.

not522:
Thank you for your reviews! I updated this PR. PTAL.

c-bata (commented):
Thanks for the update! Changes looks almost good to me but I leave two suggestions.

not522:
Thank you! I merged your suggestions.

c-bata:
Thanks for the update! LGTM after fixed CI issues :bow:

not522:
Now, the CI is back to normal.

c-bata (approved):
LGTM!

knshnb (approved):
Thanks for updating. LGTM!

Issue #3563, opened 2022-05-16
Fix random seed related tests


nzw0301:
In these files (under `/tests`), the `seed` method is used as a seed value:
`integration_tests/test_botorch.py`, `multi_objective_tests/samplers_tests/test_motpe.py`, `multi_objective_tests/samplers_tests/test_nsga2.py`, `multi_objective_tests/samplers_tests/test_samplers.py`, `samplers_tests/test_nsgaii.py`, `samplers_tests/test_partial_fixed.py`, `samplers_tests/test_samplers.py`, `samplers_tests/tpe_tests/test_sampler.py`

In these files (under `/optuna`), the `RandomState` object is overwritten instead of calling the `seed` method:
`multi_objective/samplers/_nsga2.py`, `samplers/_random.py`, `samplers/_tpe/sampler.py`, `samplers/nsgaii/_sampler.py`

I might overlook some files.

_Originally posted by @gasin in https://github.com/optuna/optuna/issues/3527#issuecomment-1120919233_

---
We can find detailed discussion: https://github.com/optuna/optuna/pull/3527#discussion_r867450297

PR #3569, opened 2022-05-17
Use reseeding method of `np.random.RandomState`  and fix `test_reseed_rng`

nzw0301:
## Motivation

As briefly summarised by #3563, we can reuse the same instance `np.random.RandomState` for reseeding by calling `seed()` method rather than instantiate a new `np.random.RandomState` instance. 

Related to this,  the current `test_reseed_rng`s do not test the expected behaviours of `reseed` methods.

## Description of the changes

[x]  Replace `np.random.RandomState()` with `seed()` method
[ ] Refactor `test_reseed_rng`

### DIFF
--- a/optuna/multi_objective/samplers/_nsga2.py
+++ b/optuna/multi_objective/samplers/_nsga2.py
@@ -94,7 +94,7 @@ class NSGAIIMultiObjectiveSampler(BaseMultiObjectiveSampler):
 
     def reseed_rng(self) -> None:
         self._random_sampler.reseed_rng()
-        self._rng = np.random.RandomState()
+        self._rng.seed()
 
     def infer_relative_search_space(
         self,
--- a/optuna/samplers/_random.py
+++ b/optuna/samplers/_random.py
@@ -44,7 +44,7 @@ class RandomSampler(BaseSampler):
 
     def reseed_rng(self) -> None:
 
-        self._rng = numpy.random.RandomState()
+        self._rng.seed()
 
     def infer_relative_search_space(
         self, study: Study, trial: FrozenTrial
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -280,7 +280,7 @@ class TPESampler(BaseSampler):
 
     def reseed_rng(self) -> None:
 
-        self._rng = np.random.RandomState()
+        self._rng.seed()
         self._random_sampler.reseed_rng()
 
     def infer_relative_search_space(
--- a/optuna/samplers/nsgaii/_sampler.py
+++ b/optuna/samplers/nsgaii/_sampler.py
@@ -174,7 +174,7 @@ class NSGAIISampler(BaseSampler):
 
     def reseed_rng(self) -> None:
         self._random_sampler.reseed_rng()
-        self._rng = np.random.RandomState()
+        self._rng.seed()
 
     def infer_relative_search_space(
         self, study: Study, trial: FrozenTrial


nzw0301:
I'm wondering if the `test_reseed_rng` function could be defined in `test_samplers.py` to reduce the duplicated implementation in each sampler test.

HideakiImamura:
@not522 Could you review this PR if you have time?

optuna/multi_objective/samplers/_nsga2.py
```diff
@@ -94,7 +94,7 @@ def __init__(
 
     def reseed_rng(self) -> None:
         self._random_sampler.reseed_rng()
-        self._rng = np.random.RandomState()
+        self._rng.seed()
```
HideakiImamura, line 97: This is not strong opinion, but `optuna/multi_objective` module is deprecated, so we don't have to fix this modules.  (In mu understanding, if a feature is deprecated, it is under the maintenance mode, so we don't have to fix it unless it has a critical bug.)

tests/samplers_tests/test_grid.py
```diff
@@ -215,10 +215,10 @@ def test_same_seed_trials() -> None:
 
 def test_reseed_rng() -> None:
```
HideakiImamura, line 216: I prefer to move all `test_reseed_rng` of samplers into `tests/samplers_tests/test_samplers.py` and use `parametrize`. Is it reasonable?

HideakiImamura (commented):
Thanks for the PR! I have two comments to simplify the changes. PTAL.

tests/samplers_tests/test_grid.py
nzw0301: Thanks. I'll create a parameterised version to compare which approach is better.

optuna/multi_objective/samplers/_nsga2.py
nzw0301: Thank you for your comment. I think you're right. Should I revert the change?

tests/samplers_tests/test_grid.py
nzw0301: What do you think about this unified implementation?
https://github.com/optuna/optuna/pull/3569/commits/53a903239267dbb3de6629805d7d3b22b20c8779

optuna/multi_objective/samplers/_nsga2.py
HideakiImamura: > I think you're right. Should I revert the change?

Yes! It would be nice to revert them!

> By the way, should we replace...

Nice catch! Yes, I think it should be replaced, but don't have to use `multivariate=True`. (TPESampler automatically use multi-objective TPE if the number of objectives >= 2.)

And, perhaps a test could be added to `tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py` to check that MOTPESampler is deprecated. It can be a followup.

HideakiImamura:
Thanks for the update! I will re-review soon.

tests/samplers_tests/test_samplers.py
```diff
@@ -84,6 +86,51 @@
...
+    lambda: PartialFixedSampler(
+        fixed_params={"x": 0}, base_sampler=optuna.samplers.RandomSampler()
+    )
+)
+
+
+@parametrize_reseedable_sampler
+def test_sampler_reseed_rng(sampler_class: Callable[[], BaseSampler]) -> None:
+    sampler = sampler_class()
+    has_rng = "_rng" in sampler.__dict__
```
HideakiImamura, line 100: This depends on the name of each sampler's private attribute, which I think is too much of a dependency on the implementation. How about checking each `attr` of `sampler.__dict__` is an instance of `np.random.RandomState`?

tests/samplers_tests/test_samplers.py
```diff
@@ -84,6 +86,51 @@
...
+    else:
+        sampler.reseed_rng()
+
+    if has_rng:
+        assert str(original_random_state) != str(sampler._rng.get_state())  # type: ignore
+
+    if has_another_sampler:
+        assert str(original_had_sampler_random_state) != str(
+            sampler.__dict__[had_sampler_name]._rng.get_state()  # type: ignore
+        )
```
HideakiImamura, line 133:
```suggestion
    if has_rng:
        original_random_state = sampler._rng.get_state()  # type: ignore
        sampler.reseed_rng();
        assert str(original_random_state) != str(sampler.__dict__[rng_name].get_state())  # type: ignore

    if has_another_sampler:
        had_sampler = sampler.__dict__[had_sampler_name]
        original_had_sampler_random_state = had_sampler._rng.get_state()  # type: ignore

        with patch.object(
            had_sampler,
            "reseed_rng",
            wraps=had_sampler.reseed_rng,
        ) as mock_object:
            sampler.reseed_rng()
            assert mock_object.call_count == 1

        assert str(original_had_sampler_random_state) != str(
            sampler.__dict__[had_sampler_name]._rng.get_state()  # type: ignore
        )
```

HideakiImamura (commented):
Thanks for the update. Basically, LGTM. I have several comments. PTAL.

By the way, since we have introduced the parametrized test, we can remove tests scattered in each test file.

optuna/multi_objective/samplers/_nsga2.py
nzw0301: # Revert the deprecated samplers under `optuna/multi_objective/samplers/`

I realised that these samplers somehow inherit `optuna/samplers`, so we need to fix the tests because the current tests for these deprecated samplers failed by introducing the change to `optuna/samplers` by this PR.

# Replace `MOTPESampler ` with `TPESampler `

Good to know that. Thank you for your comment.

tests/samplers_tests/test_samplers.py
```diff
@@ -84,6 +86,51 @@
         else [lambda: optuna.integration.BoTorchSampler(n_startup_trials=0)]
     ),
 )
+parametrize_reseedable_sampler = copy.deepcopy(parametrize_sampler)
+parametrize_reseedable_sampler.args[1].append(
+    lambda: PartialFixedSampler(
+        fixed_params={"x": 0}, base_sampler=optuna.samplers.RandomSampler()
+    )
+)
```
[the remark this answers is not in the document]
nzw0301: When I added the `PartialFixedSampler` above, a few tests failed as follows:

```
FAILED tests/samplers_tests/test_samplers.py::test_float[distribution3-<lambda>5] - assert False
FAILED tests/samplers_tests/test_samplers.py::test_float[distribution4-<lambda>5] - numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'subtract' output from dtype('float64') to dtype('int64') with cast...
FAILED tests/samplers_tests/test_samplers.py::test_float[distribution5-<lambda>5] - numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'subtract' output from dtype('float64') to dtype('int64') with cast...
FAILED tests/samplers_tests/test_samplers.py::test_int[distribution6-<lambda>5] - assert False
FAILED tests/samplers_tests/test_samplers.py::test_categorical[choices0-<lambda>5] - ValueError: '0' not in (1, 2, 3).
FAILED tests/samplers_tests/test_samplers.py::test_categorical[choices1-<lambda>5] - ValueError: '0' not in ('a', 'b', 'c').
FAILED tests/samplers_tests/test_samplers.py::test_categorical[choices2-<lambda>5] - ValueError: '0' not in (1, 'a').
FAILED tests/samplers_tests/test_samplers.py::test_conditional_sample_independent[<lambda>5] - ValueError: '0' not in ('a', 'b').
```

So adding `lambda: PartialFixedSampler(fixed_params={}, base_sampler=optuna.samplers.RandomSampler(),)` to `parametrized` works.

---

I realised that `GridSampler` might also be added to the list. But I suppose that adding `PartialFixedSampler` to the parametrized list is not simple since we need to define`search_space` depending on the tests.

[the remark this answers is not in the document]
nzw0301: > By the way, we have introduced reseed_rng to GridSampler, we can make BaseSampler.reseed_rng as an abstract method. What do you think?

Sounds nice. I'll update this PR to do so.

tests/samplers_tests/test_samplers.py
nzw0301: Thank you for your proposal. I agree with you.

tests/samplers_tests/test_samplers.py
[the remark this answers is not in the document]
not522: > By the way, we have introduced reseed_rng to GridSampler, we can make BaseSampler.reseed_rng as an abstract method. What do you think?

I think that we need to keep `BaseSampler.reseed_rng` as a non-abstract method because the change requires implementing `reseed_rng` in user-defined samplers.

tests/samplers_tests/test_samplers.py
nzw0301: done: https://github.com/optuna/optuna/pull/3569/commits/f1845b1356df40621ceca2a7fa127a70dabe424a

tests/samplers_tests/test_samplers.py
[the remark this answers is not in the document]
HideakiImamura: > I think that we need to keep BaseSampler.reseed_rng as a non-abstract method because the change requires implementing reseed_rng in user-defined samplers.

Sounds nice. Let me resolve this conversation.

HideakiImamura (commented):
Sorry for the delay. I have several comments. PTAL.

tests/samplers_tests/test_samplers.py
```diff
@@ -84,6 +85,68 @@
...
+        return None
+
+    sampler = sampler_class()
+    has_another_sampler = False
+
+    rng_name = _extract_attr_name_from_sampler_by_cls(sampler, np.random.RandomState)
+    has_rng = rng_name is not None
+
+    had_sampler_name = _extract_attr_name_from_sampler_by_cls(sampler, BaseSampler)
+    has_another_sampler = had_sampler_name is not None
```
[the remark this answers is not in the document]
nzw0301: Love it. I suppose that it makes the test more readable. 

Done: https://github.com/optuna/optuna/pull/3569/commits/35c6ce928d89864ae81659b60f020f14faa61c70

not522 (commented):
Basically LGTM. I'm thinking about reducing if-statement in `test_sampler_reseed_rng`...

not522:
We can use `.seed` here, too. https://github.com/optuna/optuna/blob/9618a93b28e206beabab4951df5e2adc0d64cd35/optuna/samplers/_cmaes.py#L378

tests/samplers_tests/test_samplers.py
```diff
@@ -86,6 +86,102 @@
...
+    rng_name = _extract_attr_name_from_sampler_by_cls(sampler, np.random.RandomState)
+    has_rng = rng_name is not None
+    assert expected_has_rng == has_rng
+
+    had_sampler_name = _extract_attr_name_from_sampler_by_cls(sampler, BaseSampler)
+    has_another_sampler = had_sampler_name is not None
+    assert expected_has_another_sampler == has_another_sampler
+
+    # CmaEsSampler has a RandomState that is not reseed by its reseed method.
+    if has_rng and not isinstance(sampler, optuna.samplers.CmaEsSampler):
```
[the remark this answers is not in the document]
nzw0301: Thanks! Indeed it looks nice!

HideakiImamura (approved):
Thanks for the update! LGTM!

PR #3576, opened 2022-05-21
Add tests for inf/nan objectives for `ShapleyImportanceEvaluator`

nzw0301:
## Motivation

Followup of https://github.com/optuna/optuna/pull/3507.

After sending [the PR](https://github.com/optuna/optuna/pull/3507), https://github.com/optuna/optuna/pull/3500 has introduced two new tests:
`test_mean_decrease_impurity_importance_evaluator_with_infinite` and `test_multi_objective_mean_decrease_impurity_importance_evaluator_with_infinite` to the test of `MeanDecreaseImpurityImportanceEvaluator`. The counterpart of `ShapleyImportanceEvaluator` is not implemented yet.

## Description of the changes

Implement tests.

### DIFF
--- a/tests/integration_tests/test_shap.py
+++ b/tests/integration_tests/test_shap.py
@@ -1,7 +1,13 @@
+from typing import Tuple
+
+import pytest
+
 from optuna import create_study
 from optuna import Trial
+from optuna.distributions import FloatDistribution
 from optuna.integration.shap import ShapleyImportanceEvaluator
 from optuna.samplers import RandomSampler
+from optuna.trial import create_trial
 
 
 def objective(trial: Trial) -> float:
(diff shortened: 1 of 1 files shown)


toshihikoyanase (approved):
Thank you for your PR. Looks solid.

I think this is out-of-scope of this PR, but the f-anova evaluator and the mean decrease impurity evaluator also have similar test cases. Maybe we can merge them, for example, in `tests/importance_tests/tests_evaluator.py`.

nzw0301:
Thank you for your quick review!

> I think this is out-of-scope of this PR, but the f-anova evaluator and the mean decrease impurity evaluator also have similar test cases. Maybe we can merge them, for example, in tests/importance_tests/tests_evaluator.py.

I completely agree with you.

tests/integration_tests/test_shap.py
```diff
@@ -71,3 +77,78 @@ def test_mean_abs_shap_importance_evaluator_with_target() -> None:
...
+            params={"x1": 1.0, "x2": 1.0, "x3": 3.0},
+            distributions={
+                "x1": FloatDistribution(low=0.1, high=3),
+                "x2": FloatDistribution(low=0.1, high=3, log=True),
+                "x3": FloatDistribution(low=2, high=4, log=True),
+            },
+        )
+    )
+    # Importance scores are calculated with a trial with an inf value.
+    param_importance_with_inf = evaluator.evaluate(study)
```
contramundum53, line 108:
```suggestion
    param_importance_with_inf = get_param_importances(study, evaluator=evaluator)
```
Maybe this is out of scope of this PR, but I think we should use public interfaces in tests whenever it is possible.

nzw0301: Thanks. How about applying the change as a follow-up when we unify the test as suggested in https://github.com/optuna/optuna/pull/3576#pullrequestreview-981549111 after we merge https://github.com/optuna/optuna/pull/3552?

nzw0301: Since as you said the suggested change is supposed to be out-of-scope in terms of the aim of this PR.

contramundum53 (approved):
Looks good to me. Thanks!

PR #3580, opened 2022-05-23
Install 3rd party libraries in CI for lint

not522:
## Motivation
In the checks workflow, we have not installed 3rd party libraries for visualization, integration, benchmarks, etc. It skips some checks, especially in mypy.

## Description of the changes
* Install 3rd party libraries in CI for lint.


not522:
`pytest.warns(None)` is deprecated. https://docs.pytest.org/en/7.0.x/deprecations.html#using-pytest-warns-none

not522:
Checks CI passed. Now, this PR is ready for review.

contramundum53:

contramundum53 (approved):
LGTM!

toshihikoyanase (commented):
Thank you for your PR. The type checking in Python is getting more and more popular, and it may be a good time to introduce more strict checking. On the other hand, I'm not confident with the change in terms of developers' experience.

The change will increase the duration of the checks jobs from 1 min to 15 min. I see some developers use the jobs instead of a local linter and their turn-around time will also increase. Also, the required CI jobs might be more fragile to the update of third-party libraries.

image

Not a strong opinion, but how about adding this strict `checks` job to daily CI?
Currently, the daily CI is mainly checked by the committers and it hardly affects external developers.

toshihikoyanase:
I talked with @not522 , and they will move this strict lint to daily CI. Keep this PR as it is, please.

toshihikoyanase:
If we install recent `xgboost`, `xgb.core.CallbackEnv` is not defined, and the following errors are raised. So we need `type: ignore` there.

```console
optuna/integration/xgboost.py:79: error: Name "xgb.core.CallbackEnv" is not defined
optuna/integration/xgboost.py:100: error: Name "xgb.core.CallbackEnv" is not defined
```

https://github.com/optuna/optuna/runs/6552734319?check_suite_focus=true

On the other hand, the errors will not be raised when we do not install `xgboost` since we ignore missing import.

```console
optuna/integration/xgboost.py:7[9](https://github.com/optuna/optuna/runs/6912328489?check_suite_focus=true#step:8:10): error: Unused "type: ignore" comment
optuna/integration/xgboost.py:100: error: Unused "type: ignore" comment
```

https://github.com/optuna/optuna/runs/6912328489?check_suite_focus=true

IMO, we can omit the type check of older xgboost, for example, by using `@typing.no_type_check`. (But I'm not sure if it is expected usage of the annotation). And, I'm also wondering why we keep the callback since the other integration modules only support the latest version of the target libraries.

not522:
I found following errors.

1. `xarray.Dataset.from_dict` is an untyped function. (It will be fixed in the next `xarray` release.)
1. `chainer`'s and `pytorch.distributed`'s functions are untyped.
1. The latest `PyTorchLightning` is not supported.
1. The latest `XGBoost` is not supported.
1. Incompatible subclassing in `test_pytorch_lightning.py`.
1. `pytest.warns(None)` is deprecated.

For 1, we need to wait.
2 is hard to resolve. I added `type: ignore` comments.
3 and 4 should be fixed in other PRs.
5 is hard to resolve. I recommend that we just ignore it.
6 should be fixed in this PR.

.github/workflows/checks-integration.yml
```diff
@@ -0,0 +1,50 @@
...
+
+    - name: Install
+      run: |
+        python -m pip install -U pip
+        pip install --progress-bar off -U .[checking]
+        pip install --progress-bar off -U .[optional]
+        pip install --progress-bar off -U .[integration] --extra-index-url https://download.pytorch.org/whl/cpu
+        pip install --progress-bar off -U .[benchmark]
+        pip install --progress-bar off -U bayesmark
+        pip install --progress-bar off -U kurobako
```
not522, line 39: Skip installing naslib because it requires specific library versions.

not522:
All checks are passed. PTAL.

setup.cfg
```diff
@@ -20,7 +20,6 @@ disallow_incomplete_defs = True
 check_untyped_defs = True
 no_implicit_optional = True
 warn_redundant_casts = True
-warn_unused_ignores = True
```
toshihikoyanase, line 23: [Notes] `.github/workflows/checks-integration.yml` checks with this option (`warn_unused_ignores`) and unnecessary `# type: ignore` will be detected there.

tests/integration_tests/test_pytorch_lightning.py
```diff
@@ -20,30 +23,42 @@ def __init__(self) -> None:
...
+            List[List[Union[torch.Tensor, Dict[str, Any]]]],
+        ],
+    ) -> None:
         if not len(outputs):
             return
 
-        accuracy = sum(x["validation_accuracy"] for x in outputs) / len(outputs)
+        accuracy = sum(
+            x["validation_accuracy"] for x in cast(List[Dict[str, torch.Tensor]], outputs)
+        ) / len(outputs)
```
toshihikoyanase, line 61: [Notes] `outputs` can be `List[torch.Tensor]`, but original code apparently expects `List[Dict[str, torch.Tensor]]` , so this fix is acceptable to me.

toshihikoyanase (approved):
Thank you for your update. The change looks great to me.
Some changes are not intuitive to me, and I left some notes just for reminder.

.github/workflows/checks-integration.yml
```diff
@@ -0,0 +1,50 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
```
toshihikoyanase, line 20: The scheduled run might not be intended to run in forks. But, we can work on it in a follow-up PR.

https://github.com/optuna/optuna/blob/fc5e262042a60bbcc888541ce7fa5e08769ac478/.github/workflows/speed-benchmarks.yml#L11-L12

toshihikoyanase:
@not522 requested another review from @contramundum53 . Please merge after your review.

contramundum53 (approved):
LGTM. Thanks!

PR #3564, opened 2022-05-16
Add intermediate_value_type column to represent inf/-inf on RDBStorage.

c-bata:
## Motivation

Refs #3206 
Follow-up https://github.com/optuna/optuna/pull/3238 and https://github.com/optuna/optuna/pull/3327

To store `inf` and `-inf` as an intermediate_value, #3228 introduces the logic that clips intermediate_value like `np.clip(intermediate_value, np.finfo(np.float32).min, np.finfo(np.float32).max)`. But the logic was broken at #3327 since we changed the floating point precision to double precision numbers.

## Description of the changes

This PR adds a new enum column that makes intermediate_value able to represent `inf` and `-inf`.

I confirmed the migration script by using himkt 's optuna-2e. The corresponding branch is:
https://github.com/c-bata/optuna-e2e/tree/optuna-3564

PR #3581, opened 2022-05-23
Add option to skip table creation to `RDBStorage`.

toshihikoyanase:
## Motivation

Currently, `RDBStorage` accidentally creates missing tables when users try to upgrade database schema. `op.create_table` in  the migration scripts are ignored.

https://github.com/optuna/optuna/blob/bf8ec607095a4b1d320b1171fa61227e9bf1366a/optuna/storages/_rdb/alembic/versions/v2.4.0.a.py#L83-L115

This is because [`sqlalchemy.schema.MetaData.create_all`](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.MetaData.create_all) is always called in `RDBStorage.__init__`.

In https://github.com/optuna/optuna/pull/3564, `v2.4.0.a.py` migration script requires this option not to create the latest `trial_intermediate_values` table.

## Description of the changes

- Add `skip_table_creation` option

### DIFF
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -657,7 +657,7 @@ class _StorageUpgrade(_BaseCommand):
         if storage_url.startswith("redis"):
             self.logger.info("This storage does not support upgrade yet.")
             return
-        storage = RDBStorage(storage_url, skip_compatibility_check=True)
+        storage = RDBStorage(storage_url, skip_compatibility_check=True, skip_table_creation=True)
         current_version = storage.get_current_version()
         head_version = storage.get_head_version()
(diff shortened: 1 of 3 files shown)


c-bata:
Thank you! I didn't notice that. At a glance, changes looks great to me. I will rebase #3564 into this branch and fix the migration script :bow:

c-bata (approved):
Thanks for your thorough review for #3564. I confirmed that #3564 works as expected without server_default and a tricky if statement.

contramundum53 (approved):
Looks good to me. Thanks!

Issue #3464, opened 2022-04-11
Make `bayesmark` benchmark consistent with `kurobako` benchmark


HideakiImamura:
### Motivation

Part of #3245.

Recently, the algorithm benchmark environment with [`bayesmark`](https://github.com/uber/bayesmark) has been introduced in #3354. We can easily evaluate the performance of the black-box optimization algorithm on the CI (GitHub Actions) for the hyperparameter optimization task of machine learning models with `sklearn`. The benchmark result is generated by the features of `bayesmark`, so it is slightly different from that of the benchmark environment with `kurobako`. 

To compare with the result fairly, it would be great to make the benchmark result of `bayesmark` consistent with that of `kurobako`. We want a help to do this.

## About `bayesmark` benchmark

The benchmark result with `bayesmark` consists of plots (.png files) and a report (.md file). 

The plot looks like the following.
![optuna-wine-RF-sumamry](https://user-images.githubusercontent.com/38826298/162708304-47cf711a-8151-4687-abb9-8dc61d7e5865.png)

<details><summary>The report looks like this.</summary><div>

```
# Benchmark Result Report

* Number of Solvers: 4
* Number of Models: 7
* Number of Datasets: 5
* Number of Problems: 35

Final score for each problem is calculated as `100 x (1-loss)`. Solver with lowest score in each problem wins. For more details visit [bayesmark docs.](https://bayesmark.readthedocs.io/en/stable/scoring.html)

## Table of Contents

1. [Problem Leaderboards](#problem-leaderboards)
2. [Datasets](#datasets)
3. [Models](#models)

## Problem Leaderboards

### (1) Problem: Wine-kNN

|Ranking|Solver|Score|
|---:|:---|---:|
|1|RandomSearch_0.0.8_a376313|98.43800|
|2|CmaEsSampler-NopPruner-Optuna_3.0.0b0.dev0_a376313|99.33737|
|3|RandomSampler-NopPruner-Optuna_3.0.0b0.dev0_a376313|99.72773|
|4|TPESampler-NopPruner-Optuna_3.0.0b0.dev0_a376313|100.00000|
### (2) Problem: Digits-RF

|Ranking|Solver|Score|
|---:|:---|---:|
|1|CmaEsSampler-NopPruner-Optuna_3.0.0b0.dev0_a376313|29.71156|
|2|RandomSearch_0.0.8_a376313|52.48060|
[...]

PR #3584, opened 2022-05-24
Make `bayesmark` benchmark results comparable to `kurobako`

xadrianzetx:
## Motivation

Previously `bayesmark` benchmark was using its own internal ranking system to score optimizers. This made comparing `kurobako` and `bayesmark` benchmark results impossible. This PR implements ranking strategy described in [A Strategy for Ranking Optimization Methods using Multiple Criteria](https://proceedings.mlr.press/v64/dewancker_strategy_2016.pdf), which is also used in `kurobako`, making two benchmarks directly comparable. Additionally, plot styling is adjusted to match more closely that of `kurobako`.

Example run with report and plots - https://github.com/xadrianzetx/optuna-bayesmark/actions/runs/2379168197

Closes https://github.com/optuna/optuna/issues/3464, and part of https://github.com/optuna/optuna/issues/3245.

~RFC tag due to ongoing discussion with @HideakiImamura on best approach to this benchmark - https://github.com/optuna/optuna/issues/3464#issuecomment-1129769341. Alternative would be to drop `bayesmark` and run sklearn problems with `kurobako` directly.~

## Description of the changes

* Re-implemented `partial_report` to use raw values bypassing ones aggregated by `bayesmark`
* Implemented kurobako-like ranking strategy and metrics
* Extended markdown report
* Adjusted plot styling
* Minor fixes to benchmark CI


HideakiImamura:
Thanks for submitting PR! I will check this implementation soon.

.github/workflows/performance-benchmarks-bayesmark.yml
```diff
@@ -90,22 +91,32 @@ jobs:
...
+      uses: actions/download-artifact@v2
       with:
         name: partial-reports
         path: partial
     
+    - name: Install Python libraries
+      run: |
+        python -m pip install --upgrade pip
+        pip install --progress-bar off -U setuptools wheel
+        pip install --progress-bar off numpy scipy pandas
```
HideakiImamura, line 107: How about merging this install section to that of Line 53?

HideakiImamura (commented):
Thanks for the amazing improvement of the bayesmark benchmark. I would like to merge this PR. Let me leave an early comment.

HideakiImamura:
I think we can remove the RFC from the title of this PR, since @xadrianzetx's solution is simple enough to merge.

.github/workflows/performance-benchmarks-bayesmark.yml
xadrianzetx: This is install section of `report` job, while on L53 we have installs for `benchmarks` job - both have separate environments and installs. AFAIK there is no way to define common job steps between jobs (like GitLab [.before_script](https://docs.gitlab.com/ee/ci/yaml/#before_script)), but I could be wrong here.

benchmarks/report_bayesmark.py
```diff
@@ -1,55 +1,255 @@
...
+        self._problems_body.write("".join([_HEADER_FORMAT, *metric_format, _LINE_BREAK]))
+
+        positions = np.abs(ranking.borda - (max(ranking.borda) + 1))
+        for pos, solver in zip(positions, ranking.solvers):
+            self._solvers.add(solver)
+            row_buffer = io.StringIO()
+            row_buffer.write(f"|{pos}|{solver}|")
+            for metric in metrics:
+                mean, variance = report.summarize_solver(solver, metric)
+                row_buffer.write(f"{mean:.5f} +- {np.sqrt(variance):.5f}|")
```
xadrianzetx, line 216: Oh, btw, currently each metric is rendered with fixed number of decimal places,which is not true to `kurobako`, where this changes depending on metric. I initially planned to implement that by adding another class attribute to each metric (besides name) which would make it very convenient to reference here but, well... I forgot :facepalm:. Can do it now, or save it for some neat follow-up, depends how do you feel about that @HideakiImamura.

.github/workflows/performance-benchmarks-bayesmark.yml
HideakiImamura: Thanks for the explanation. You're right.

HideakiImamura:
@sile Could you review this PR if you have time?

benchmarks/report_bayesmark.py
HideakiImamura: To simplify the change, I think it is good to say that this should be a follow-up.

benchmarks/report_bayesmark.py
```diff
@@ -1,55 +1,255 @@
...
+Samples = Dict[str, List[float]]
+
+
+class BaseMetric(ABC):
+    @property
+    def name(self) -> str:
+        """Metric name displayed in final report."""
+
+    def calculate(self, data: pd.DataFrame) -> List[float]:
+        """Calculates metric for each study in data frame."""
```
HideakiImamura, line 34: It is recommended to write it in the same way as Optuna's other abstract classes.
```suggestion
class BaseMetric(object, metaclass=abc.ABCMeta):
    @abc.abstractmethod
    @property
    def name(self) -> str:
        """Metric name displayed in final report."""

    @abc.abstractmethod
    def calculate(self, data: pd.DataFrame) -> List[float]:
        """Calculates metric for each study in data frame."""
```

HideakiImamura (commented):

sile:
@HideakiImamura Yep, I'll review this PR this or next weekend.

benchmarks/bayesmark/report_template.md
```diff
@@ -4,14 +4,22 @@
...
 
 ## Table of Contents
 
-1. [Problem Leaderboards](#problem-leaderboards)
-2. [Datasets](#datasets)
-3. [Models](#models)
+1. [Overall Results](#overall-results)
+2. [Problem Leaderboards](#problem-leaderboards)
+3. [Datasets](#datasets)
+4. [Models](#models)
```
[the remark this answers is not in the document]
xadrianzetx: > And, how about renaming `Problem Leaderboards` as `Individual Results` as `kurobako` does?

Done in 0da6fa1dfb3ba071c5036f415c23b99eedcb8c50.

> How about adding the "Solvers" and "Studies" sections as kurobako generated? 

Thanks for bringing this up. I was planning to emulate those sections as well (even provisioned bayesmark study IDs in partial reports), but figured it would blow up diff too much (required changes to `PartialReport` to retrieve and serve study IDs, `ReportBuilder` to render new sections and new templates for solver spec). Maybe it's better to leave this relatively big change as follow-up? I'll try my best to set some time aside for those outstanding tasks and actually finish them.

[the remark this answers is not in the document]
HideakiImamura: > Maybe it's better to leave this relatively big change as follow-up?

Sounds reasonable. I agree with you.

.github/workflows/performance-benchmarks-bayesmark.yml
```diff
@@ -59,8 +59,7 @@ jobs:
         python -c 'import optuna'
         optuna --version
 
-        pip install --progress-bar off bayesmark
-        pip install --progress-bar off matplotlib
+        pip install --progress-bar off bayesmark matplotlib pandas
```
HideakiImamura, line 62: I tried to run the benchmark on my fork. I found that we cannot use BoTorchSampler. https://github.com/HideakiImamura/optuna/runs/6682473415?check_suite_focus=true 

How about using `benchmarks` as follow?
```suggestion
        pip install --progress-bar off .[benchmark]
        pip install --progress-bar off bayesmark matplotlib pandas
```

xadrianzetx: Oh, great catch, thanks! I guess the same goes for `cma` (`PyCmaSampler`) and `scikit-optimize` (`SkoptSampler`), but those are defined in `integration` section of requirements. I don't want to pull all integration modules, nor repeat `cma` and `scikit-optimize` in `benchmark` section, so do you think installing them with `matplotlib` and `pandas` is ok here?

xadrianzetx: Fixed in 5ece8c771b87a07ebfe116b25b1051a98fbf48db and tested on [fork](https://github.com/xadrianzetx/optuna-bayesmark/actions/runs/2429401655).

benchmarks/bayesmark/report_bayesmark.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+class AUCMetric(BaseMetric):
+    name = "AUC"
+
+    def calculate(self, data: pd.DataFrame) -> List[float]:
+
+        aucs: List[float] = list()
+        for _, grp in data.groupby("uuid"):
+            auc = np.sum(grp.generalization.cummin() - grp.generalization.min())
```
sile, line 58: I'm not sure but it seems the value of `grp.generalization.min()` could differ between optimizations. If so, it would make it difficult to compare the meanings of multiple AUCs.
(e.g., if there is a sampler that always samples a constant parameter set, the value of `np.sum(grp.generalization.cummin() -  grp.generalization.min())` would always be 0 (the best AUC value))
 
How about using a common value (e.g, just a constant value `0`) here instead?
(please let me know if I misunderstand something 🙏 )

benchmarks/run_bayesmark.py
```diff
@@ -49,71 +47,76 @@ def run_benchmark(args: argparse.Namespace) -> None:
...
+    fig, ax = plt.subplots()
+    fig.set_size_inches(12, 8)
+    warmup = json.loads(args.plot_warmup)
+    metric = df.metric[0]
+    color_lookup = build_color_dict(df["opt"].unique())
 
-    for benchmark in summary.coords["function"].values:
-        for metric, ax in zip(["mean", "median"], axs):
-            make_plot(summary, ax, benchmark, metric, plot_warmup)
+    for optimizer, summary in summaries.groupby("opt"):
```
sile, line 74: It seems that the line color of each sampler could vary plot by plot. If it's possible, it would be better to keep using the same color for each sampler I think.
optuna-breast-linear-sumamry
optuna-diabetes-DT-sumamry
(images from https://github.com/xadrianzetx/optuna-bayesmark/actions/runs/2379168197)

xadrianzetx: Do you think we should produce color lookup table for each optimizer/pruner combo, or is it enough to make colors consistent throughoutt single benchmark run for subset of tested optimizers (so essentially making sure `build_color_dict(df["opt"].unique())` works on sorted values)?

sile: >  is it enough to make colors consistent throughout single benchmark run

Yes. I think this is a reasonable solution and consistent with the kurobako behavior.

benchmarks/bayesmark/report_bayesmark.py
sile: > Having fixed reference would leak absolute generalization performance into metric

I think this is an　expected AUC characteristic.
Let me quote a paragraph about AUC from [the paper](https://proceedings.mlr.press/v64/dewancker_strategy_2016.pdf):
image
It says "a specified lower bound f<sub>LB</sub> on the function" (i.e., a constant value for each function) is used to calculate AUC.

> AUC values could get very large since we are using MSE

Yeah, this is a subtle problem😅
This problem is due to an inherent property of MSE. So just accepting very large values in the result would be a ( compromised but) practical solution I think .

xadrianzetx: Thank you for the explanation, makes sense :+1: . Implemented in e2d85c4f4de85d8cfff1f9db4b477fc729471189.

benchmarks/run_bayesmark.py
xadrianzetx: Implemented in d81b200bdbbcfe7528e1bae766fefe5f4d59be2c.

xadrianzetx:
CI fail seems unrelated.

sile (approved):
LGTM 👍

HideakiImamura (approved):
LGTM.

xadrianzetx:
Thanks for reviews @HideakiImamura, @sile! There are some outstanding tasks - https://github.com/optuna/optuna/pull/3584#discussion_r883913394, https://github.com/optuna/optuna/pull/3584#discussion_r885926097, and README update. I'll work on them soon.

PR #3585, opened 2022-05-25
Restore virtualenv for benchmark extras

himkt:
## Motivation

I removed `virtualenv` from benchmark extras since I thought it was installed in Python by default. It was wrong.
https://github.com/optuna/optuna/actions/runs/2381070905

This PR restores `virtualenv`.


contramundum53 (approved):
Thanks for your PR!

not522 (approved):
LGTM.

PR #3589, opened 2022-05-26
Further refactor `ImportanceEvaluator`s

contramundum53:
## Motivation
This is a follow-up PR to #3552. By slightly breaking backward compatibility of the interface of `ImportanceEvaluator`s, we could make further refactoring.

## Description of the changes
We change the interface of `BaseImportanceEvaluator` such that the arguments `params` and `target` do not get `None` values. We fill the default `params` and `target` in `get_param_importances()` function, rather than in each `ImportanceEvaluator`. This makes it easier to implement new `ImportanceEvaluator`, as well as keeps the default value code in one place.

We also gather the code for sorting results in `get_param_importances()`, such that `ImportanceEvaluator`s do not need to return a sorted `OrderedDict`.

This breaks the interface of `ImportanceEvaluator.evaluate()` function if they are directly called from user code. However, the documentation explicitly states that users should not call that function directly, thus I don't think that is a problem.

This also breaks the type integrity of user-defined `ImportanceEvaluator`s. However, there won't be any runtime errors as long as they are called from `get_param_importances()`, since the domain of `ImportanceEvaluator.evaluate` will become smaller.


c-bata:
Should we make this draft until #3552 is merged?

contramundum53:
This PR is review-ready now. Could you take a look? @c-bata @nzw0301

optuna/importance/__init__.py
```diff
@@ -78,13 +81,45 @@ def get_param_importances(
 
     Returns:
         An :class:`collections.OrderedDict` where the keys are parameter names and the values are
-        assessed importances.
+        assessed importances. The importances will be sorted in a descending order.
 
     """
+
+    if len(study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))) == 0:
+        raise ValueError("The study does not contain completed trials.")
```
c-bata, line 89: I guess this check actually causes performance degradation of optuna-fast-fanova. Can we avoid to issue an additional SQL query here?

contramundum53: In my understanding, if we use `CachedStorage`, the results of `get_trials` are cached so that we won't need to fetch all data every time.

c-bata: Yes, it's true if we use CachedStorage.

contramundum53: We could change the first argument of `BaseImportanceEvaluator.evaluate` from `study` to `trials: List[FrozenTrial]`, and this would remove the duplicate of both SQL commands and the code. However, I think it is good to do that in a followup PR, since that is going to be another large change. What do you think? @c-bata

c-bata:

optuna/importance/__init__.py
contramundum53: #3664 Here is the followup PR :)

tests/integration_tests/test_shap.py
```diff
@@ -194,7 +211,7 @@ def objective(trial: Trial) -> Tuple[float, float]:
         study.optimize(objective, n_trials=3)
 
         with pytest.raises(ValueError):
-            ShapleyImportanceEvaluator().evaluate(study)
+            get_param_importances(study, evaluator=ShapleyImportanceEvaluator())
```
c-bata, line 214: I'd say it's better to focus the behavior of `ShapleyImportanceEvaluator().evaluate(study)` in this test file. So how about calling `ShapleyImportanceEvaluator().evaluate(study)` directly?

optuna/importance/__init__.py
```diff
@@ -68,8 +71,8 @@ def get_param_importances(
         target:
             A function to specify the value to evaluate importances.
             If it is :obj:`None` and ``study`` is being used for single-objective optimization,
-            the objective values are used. ``target`` must be specified if ``study`` is being
-            used for multi-objective optimization.
```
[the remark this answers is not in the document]
contramundum53: It is already written in the note right below, so I thought it would be a duplicate and removed it.

tests/integration_tests/test_shap.py
contramundum53: `ShapleyImportanceEvaluator().evaluate` does not handle errors anymore, so we cannot directly test it for invalid cases.
Maybe we do not need these tests because the error-handling behavior is already tested in `tests/importance_tests/test_init.py`.

tests/integration_tests/test_botorch.py
```diff
@@ -78,9 +79,9 @@ def candidates_func(
         train_obj: torch.Tensor,
         train_con: Optional[torch.Tensor],
         bounds: torch.Tensor,
-    ) -> torch.Tensor:
+    ) -> Any:
         # Must be a `torch.Tensor`, not a list.
-        return torch.rand(1).tolist()
+        return [torch.rand(1).tolist()]
```
toshihikoyanase, line 84: I guess this change is not related to refactoring of `ImportanceEvaluator`. Could you revert it, please?

toshihikoyanase (changes requested):
I'm trying to catch up with the series of the changes. Let me share my early comment.

tests/integration_tests/test_botorch.py
contramundum53: Without this change, mypy fails.

toshihikoyanase: We'll work on this change in https://github.com/optuna/optuna/pull/3679. So, could you revert the change in this PR?

optuna/importance/__init__.py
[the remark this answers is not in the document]
c-bata: Thanks! You are correct. I missed that :bow:

tests/integration_tests/test_shap.py
c-bata: Agree 👍  I think we can simply remove these tests.

tests/integration_tests/test_botorch.py
contramundum53: I reverted this change. PTAL.

nzw0301:
@contramundum53  Could you resolve conflicts?

contramundum53:
Fixed.

c-bata (approved):
Thanks for the update! LGTM.

tests/integration_tests/test_botorch.py
toshihikoyanase: Thanks.

toshihikoyanase:
I'm sorry for the delayed response. Could you resolve the conflict, please?

contramundum53:
Thanks, fixed.

toshihikoyanase:
The importance tests failed. Could you check them?
```
=========================== short test summary info ============================
FAILED tests/importance_tests/test_init.py::test_importance_evaluator_seed[MeanDecreaseImpurityImportanceEvaluator]
FAILED tests/importance_tests/test_init.py::test_importance_evaluator_seed[FanovaImportanceEvaluator]
FAILED tests/importance_tests/test_init.py::test_importance_evaluator_seed[ShapleyImportanceEvaluator]
FAILED tests/importance_tests/test_init.py::test_importance_evaluator_with_target[MeanDecreaseImpurityImportanceEvaluator]
FAILED tests/importance_tests/test_init.py::test_importance_evaluator_with_target[FanovaImportanceEvaluator]
FAILED tests/importance_tests/test_init.py::test_importance_evaluator_with_target[ShapleyImportanceEvaluator]
```

https://github.com/optuna/optuna/runs/7492253049?check_suite_focus=true

PR #3592, opened 2022-05-26
Fix nan comparison in grid sampler

### LINKED ISSUE #3556: `GridSampler` does not stop the optimization when NaN in the `search_space`

contramundum53:
## Motivation
Fix #3556


contramundum53:
@c-bata @knshnb Could you review this PR?

optuna/samplers/_grid.py
```diff
@@ -261,7 +268,9 @@ def _same_search_space(self, search_space: Mapping[str, Sequence[GridValueType]]
                 return False
 
             for i, param_value in enumerate(search_space[param_name]):
-                if param_value != self._search_space[param_name][i]:
+                if not GridSampler._grid_value_equal(
```
c-bata, line 271: nit: Can we use `self` instead of `GridSampler` here? Other staticmethods in this class is called via `self` object (e.g. `_check_value`).

```suggestion
                if not self._grid_value_equal(
```

contramundum53: Thanks, fixed.

tests/samplers_tests/test_grid.py
```diff
@@ -228,3 +228,10 @@ def test_enqueued_insufficient_trial() -> None:
 
     with pytest.raises(ValueError):
         study.optimize(lambda trial: trial.suggest_int("a", 0, 100))
+
+
+def test_nan() -> None:
+    sampler = optuna.samplers.GridSampler({"x": [0, float("nan")]})
+    study = optuna.create_study(sampler=sampler)
+    study.optimize(lambda _: 0)
```
c-bata, line 236: nit: How about sampling "x" as this test case is for sampler.

```suggestion
    study.optimize(lambda trial: trial.suggest_float("x", 0, 1))
```

c-bata (commented):
@contramundum53 Could you fix CI failures?

tests/samplers_tests/test_grid.py
contramundum53: You cannot return `nan` as a value.

contramundum53: Thanks, I'll fix it.

contramundum53:
```

self = CategoricalDistribution(choices=(0, nan)), param_value_in_external_repr = nan

    def to_internal_repr(self, param_value_in_external_repr: CategoricalChoiceType) -> float:

        try:
            return self.choices.index(param_value_in_external_repr)
        except ValueError as e:
>           raise ValueError(
                "'{}' not in {}.".format(param_value_in_external_repr, self.choices)
            ) from e
E           ValueError: 'nan' not in (0, nan).

optuna/distributions.py:499: ValueError
```
There was a further bug. I'm working on it.

contramundum53:
It seems that #3567 must be merged to pass the tests.

knshnb:
@contramundum53 Now, #3567 has been merged!

contramundum53:
@knshnb @c-bata This PR is ready for review now. PTAL.

c-bata (approved):
LGTM!

optuna/samplers/_grid.py
```diff
@@ -251,6 +252,12 @@ def _get_unvisited_grid_ids(self, study: Study) -> List[int]:
 
         return list(unvisited_grids)
 
+    @staticmethod
+    def _grid_value_equal(value1: GridValueType, value2: GridValueType) -> bool:
+        value1_is_nan = isinstance(value1, Real) and np.isnan(float(value1))
+        value2_is_nan = isinstance(value2, Real) and np.isnan(float(value2))
+        return (value1 == value2) or (value1_is_nan and value2_is_nan)
```
knshnb, line 259: I thought making `_grid_value_equal` a function instead of the static method might be more consistent with #3567, but this is not a strong opinion and I leave it to you.

knshnb (commented):
Thank you! I left a small comment.

optuna/samplers/_grid.py
contramundum53: I think it would be nicer to be a static method, because this method is never used from outside the class.

knshnb: Sure, then I'll merge this PR.

knshnb (approved):
LGTM!

PR #3594, opened 2022-05-30
Remove `language` from docs configuration

hvy:
## Motivation

See below description.

## Description of the changes

Tries to fix the CI failure observed in https://github.com/optuna/optuna/pull/3543, i.e. https://github.com/optuna/optuna/runs/6647370356?check_suite_focus=true.


toshihikoyanase (approved):
As for the hotfix, the change looks great. On the other hand, do we have any drawbacks?

docs/source/conf.py
```diff
@@ -72,13 +72,6 @@
 # The master toctree document.
 master_doc = "index"
 
-# The language for content autogenerated by Sphinx. Refer to documentation
-# for a list of supported languages.
-#
-# This is also used if you do content translation via gettext catalogs.
-# Usually you set "language" from the command line for these cases.
-language = None
```
toshihikoyanase, line 80: `WARNING: Invalid configuration value found: 'language = None'. Update your configuration to a valid langauge code. Falling back to 'en' (English).`

Can we replace `None` with `en`?

hvy: Yes, I think we can do that too. The current change is based on the following, and results will be identical in either case in my understanding. Do you want me to change?

- https://github.com/sphinx-doc/sphinx/pull/10481/files#diff-da24b355349b53e4c8047034767d008e4c2ce9ea6c8933a3acd50d0828f00528
- https://github.com/sphinx-doc/sphinx/issues/10474#issuecomment-1140389657

hvy:
> On the other hand, do we have any drawbacks?

I don't think so, please see https://github.com/optuna/optuna/pull/3594#discussion_r884419421.

c-bata (approved):
LGTM!

c-bata:
Let me merge this PR since this PR is a blocker of #3543. @toshihikoyanase Please open a follow-up issue if you have something concern about https://github.com/optuna/optuna/pull/3594#discussion_r884417529.

docs/source/conf.py
toshihikoyanase: Thank you for your explanation. In the above PR, the Sphinx developer recommended the removal, so the current change looks reasonable.

> So please remove the configuration from these projects.

toshihikoyanase:
@c-bata Thank you for merging this PR. As mentioned in https://github.com/optuna/optuna/pull/3594#discussion_r884432997, I don't have any concerns now.

PR #3595, opened 2022-05-30
Make `FloatTypeEnum` common

contramundum53:
## Motivation
Python's `float` type can represent finite values, `inf`, `-inf` and `nan`. However, only finite values can be reliably stored in RDBs (For example, MySQL cannot store `inf` or `nan` and SQLite cannot store `nan`).

In #3564, `IntermediateValue.FloatTypeEnum` was introduced to solve this problem specifically for `IntermediateValueModel`. However, `IntermediateValue.FloatTypeEnum` can be widely used to represent python's `float` values, thus we propose to take `FloatTypeEnum` out of `IntermediateValue`. This is used in #3559.

## Description of the changes
Let `FiniteFloat` be the set of finite float values. Python's `float` type (denoted as `PyFloat` here) can be described as
```
enum PyFloat{ 
    Finite(FiniteFloat), 
    PosInf, 
    NegInf, 
    NaN
}
```
in Rust notation. (We use Rust notation here because Python's type system does not have its strict counterpart.)

This type cannot be directly represented in RDBs. To represent this type in RDBs, there is a `FloatTypeEnum` defined as
```
enum FloatTypeEnum{ FiniteOrNaN, PosInf, NegInf }
```
and a `PyFloat` is represented as a tuple `(Option<FiniteFloat>, FloatTypeEnum)`. (Note that both `Option<FiniteFloat>` and `FloatTypeEnum` can be safely stored in RDBs. Namely, we use nullable real type in RDBs for `Option<FiniteFloat>`.)

More specifically, we convert `value: PyFloat` to `(Option<FiniteFloat>, FloatTypeEnum)` with the following logic:
```
match value {
    Finite(x) => (Some(x), FloatTypeEnum::FiniteOrNan),
    PosInf => (Some(0.0), FloatTypeEnum::PosInf),
    NegInf => (Some(0.0), FloatTypeEnum::NegInf),
    NaN => (None, FloatTypeEnum::FiniteOrNaN)
}
```

There are also some cases (e.g. `TrialValueModel.value` in #3559) that we want to represent a float value that can never be `nan`, i.e. 
```
enum PyFloatWithoutNaN{ 
    Finite(FiniteFloat), 
    PosInf, 
    NegInf
}
```
in Rust notation. This can be represented as `(FiniteFloat, FloatTypeEnum)` with the conversion logic being
```
match value {
    Finite(x) => (x, FloatTypeEnum::FiniteOrNan),
    PosInf => (0.0, FloatTypeEnum::PosInf),
    NegInf => (0.0, FloatTypeEnum::NegInf)
}
```

In this PR, we change #3564 slightly to conform to this semantics, and we plan to use this for #3559 subsequently.


toshihikoyanase:
@c-bata Could you review this PR, please? This is a follow-up PR for #3564. Please feel free to remove the assignment if you are busy.

optuna/storages/_rdb/storage.py
```diff
@@ -57,11 +58,52 @@
...
+        return (None, models.FloatTypeEnum.FINITE_OR_NAN)
+    else:
+        return _float_without_nan_to_pair_repr(value)
+
+
+def _pair_repr_to_float_without_nan(value: float, float_type: models.FloatTypeEnum) -> float:
+
+    if float_type == models.FloatTypeEnum.INF_POS:
+        assert value == 0.0
+        return np.Infinity
```
toshihikoyanase, line 89: Could you tell me why we user `np.Infinity` here instead of built-in `inf`?
```suggestion
        return float("inf")
```

optuna/storages/_rdb/storage.py
```diff
@@ -57,11 +58,52 @@
 
     models = _LazyImport("optuna.storages._rdb.models")
 
+
+_logger = optuna.logging.get_logger(__name__)
+
+
```
toshihikoyanase, line 64: Could you revert this change? It seems to be out-of-scope of this PR.

toshihikoyanase (commented):
Thank you for your PR. Let me share my early comments.

optuna/storages/_rdb/storage.py
contramundum53: For no particular reason. If `float("inf")` is preferrable, I'll fix it.

optuna/storages/_rdb/storage.py
contramundum53: It seems that I accidentally swapped this line with `_RDB_MAX_FLOAT`. Should I revert it?

optuna/storages/_rdb/storage.py
```diff
@@ -64,6 +65,46 @@
 _logger = optuna.logging.get_logger(__name__)
 
 
+def _float_without_nan_to_pair_repr(value: float) -> Tuple[float, models.FloatTypeEnum]:
```
HideakiImamura, line 68: I couldn't understand the meaning of word "pair_repr" from here. How about renaming it like "_float_without_nan_to_float_to_be_stored"?

HideakiImamura, line 68: These methods are only used in RDBStorage. How about making them as `staticmethod`s of RDBStorage?

tests/storages_tests/test_with_server.py
```diff
@@ -122,10 +119,15 @@ def test_store_infinite_values(input_value: float, expected: float, storage_url:
...
     study_id = storage.create_new_study()
     trial_id = storage.create_new_trial(study_id)
-    storage.set_trial_intermediate_value(trial_id, 1, input_value)
-    storage.set_trial_state_values(trial_id, state=TrialState.COMPLETE, values=(input_value,))
-    assert storage.get_trial(trial_id).value == expected
-    assert storage.get_trial(trial_id).intermediate_values[1] == expected
+    storage.set_trial_intermediate_value(trial_id, 1, value)
+
+    got_value = storage.get_trial(trial_id).intermediate_values[1]
+    assert got_value == value or (np.isnan(got_value) and np.isnan(value))
```
HideakiImamura, line 125: 👍

tests/storages_tests/test_with_server.py
```diff
@@ -122,10 +119,15 @@ def test_store_infinite_values(input_value: float, expected: float, storage_url:
...
+
+    if not np.isnan(value):
+        # nan is not acceptable as the final value.
```
HideakiImamura, line 128: This line should be moved above the line 127. In addition, the sentence should be started with an uppercase letter.
```suggestion
        # A `nan` is not acceptable as the final value.
```

optuna/storages/_rdb/storage.py
HideakiImamura, line 68: How about testing these methods directly to confirm the behavior as expected?

HideakiImamura (commented):

optuna/storages/_rdb/storage.py
contramundum53: What about "_float_without_nan_to_stored_repr"? Does that make sense?

contramundum53: :+1: I'll fix it. Thanks.

contramundum53:
@HideakiImamura Applied comments. PTAL 🙇

HideakiImamura:
Let me close this PR since we have proceed #3603.

PR #3553, opened 2022-05-12
Move heartbeat-related tests in `test_storages.py` to another file

g-votte:
Please review/merge this PR after #3475

## Motivation
This is a followup PR of #3475. Since storage and heartbeat are independent responsibilities after we merge #3475, this PR moves heartbeat-related tests in `test_storages.py`.

## Description of the changes
This PR moves all heartbeat-relates tests in `test_storages.py` to a new file named `test_heartbeat.py`. The test methods are moved as they are.

PR #3596, opened 2022-05-30
Move `is_heartbeat_enabled` from storage to heartbeat

g-votte:
Please review and merge after this PR after #3553 

## Motivation
This is one of followup PRs of the heartbeat refactoring (#3475).

This change has the following advantages:
- By removing the `is_heartbeat_enabled` method, `BaseStorage` becomes completely independent from the heartbeat mechanism.
- The cyclic dependency b/w `optuna/storages/_base.py` and `optuna/storages/_heartbeat.py` is resolved.

## Description of the changes
- Move `is_heartbeat_enabled` from `BaseStorage` class to `_heartbeat` module
- Replace `storage.is_heartbeat_enabled()` with `is_heartbeat_enabled(storage)` in all lines where `BaseStorage.is_heartbeat_enabled` was called


g-votte:
Since #3553 has been merged, this PR is ready for review. PTAL.

toshihikoyanase:
@contramundum53 @keisuke-umezawa  Could you review this PR? This seems to be a subsequent PR of https://github.com/optuna/optuna/pull/3475 and https://github.com/optuna/optuna/pull/3553. Please feel free to remove the assignment if you are busy.

contramundum53 (approved):
I did not follow the context but this change looks good.

optuna/storages/_heartbeat.py
```diff
@@ -100,3 +101,14 @@ def fail_stale_trials(study: "optuna.Study") -> None:
...
+
+def is_heartbeat_enabled(storage: BaseStorage) -> bool:
+    """Check whether the storage enables the heartbeat.
+
+    Returns:
+        :obj:`True` if the storage also inherits :class:`~optuna.storages._heartbeat.BaseHeartbeat`
+        and the return value of :meth:`~optuna.storages.BaseStorage.get_heartbeat_interval` is an
+        integer, otherwise :obj:`False`.
+    """
+    return isinstance(storage, BaseHeartbeat) and storage.get_heartbeat_interval() is not None
```
contramundum53, line 114:
```suggestion
def is_heartbeat_enabled(storage: BaseStorage) -> bool:
    """Check whether the storage enables the heartbeat.

    Returns:
        :obj:`True` if the storage also inherits :class:`~optuna.storages._heartbeat.BaseHeartbeat`
        and the return value of :meth:`~optuna.storages.BaseStorage.get_heartbeat_interval` is an
        integer, otherwise :obj:`False`.
    """
    return isinstance(storage, BaseHeartbeat) and storage.get_heartbeat_interval() is not None

BaseStorage.is_heartbeat_enabled = is_heartbeat_enabled
```

From users' perspective, I think `storage.is_heartbeat_enabled()` is more intuitive than `is_heartbeat_enabled(storage)`. So, what about patching `BaseStorage` class? This will remove the cyclic dependency while still keeping the syntax `storage.is_heartbeat_enabled()`.

Of course this is subject to discussion because it is not standard that a bound function is not defined in the class.

contramundum53, line 114:
```suggestion
    return isinstance(storage, BaseHeartbeat) and storage.get_heartbeat_interval() is not None

BaseStorage.is_heartbeat_enabled = is_heartbeat_enabled

```

From users' perspective, I think `storage.is_heartbeat_enabled()` is more intuitive than `is_heartbeat_enabled(storage)`. So, what about patching `BaseStorage` class? This will remove the cyclic dependency while still keeping the syntax `storage.is_heartbeat_enabled()`.

Of course this is subject to discussion because it is not standard that a bound function is not defined in the class.

contramundum53: I understand that there are arguments both ways, so I don't retract my approval.

keisuke-umezawa (approved):
Looks good to me!

keisuke-umezawa:
@contramundum53 You left one comment saying `I understand that there are arguments both ways, so I don't retract my approval.` If you are ok with PR, could you merge it? Or, could you discuss it with @g-votte ?

contramundum53:
@g-votte Do you like my approach? If not, I think I can merge it.

optuna/storages/_heartbeat.py
g-votte: Hmm, even if providing `BaseStorage.is_heartbeat_enabled()` is more intuitive, I think we need to discuss some design trade-off.
- Since `BaseStorage` currently is not open for library users, we can prioritize simplicity over intuitiveness.
- Patching `BaseStorage` in `_heartbeat.py` might be somewhat confusing in terms of the library structure.

Can we just leave those discussion out-of-scope of this PR so far?

g-votte:
@contramundum53 
Thanks for your confirmation. As I commented in https://github.com/optuna/optuna/pull/3596#discussion_r888635533,  I'd like to leave this out-of-scope for this PR.

optuna/storages/_heartbeat.py
contramundum53: I think that's a valid argument. I'll merge this PR.

PR #3609, opened 2022-06-01
Move heartbeat-related thread operation in `_optimize.py` to `_heartbeat.py`

g-votte:
Please review and merge this PR after #3596.

Although the heartbeat mechanism has been decoupled from storages in #3475 and #3596, it still has tight-coupling with other modules like `_optimize.py`. Therefore, this PR separates a part of heartbeat logic in that file, by encapsulating the thread operation for heartbeat.

## Motivation
Since the thread operation is always required when using the heartbeat mechanism, the handling logic should be in `_heartbeat.py` rather than another module.

The change causes other specific advantages:
- Simplifying the responsibility of `_run_trial`: the responsibility of `_run_trial` is much more than running a trial, where the method even handles starting/joining a thread and related conditional branches. Encapsulating those logics will decrease the internal complexity of `_run_trial`.
- Making the thread operation reusable: I believe this change also promotes #3293. By encapsulating the thread operation of heartbeat, it'd become much easier to run the same logic with ask-and-tell methods.

## Description of the changes
- Most heartbeat-related logics are moved from `_optimize.py` to `_hearbeat.py`.
- `BaseHeartbeatThread` and its child classes are introduced in `_heartbeat.py` to encapsulate the thread operation.
  - `HeatbeatThread` is responsible for starting and joining heartbeat threads when the storage supports heartbeat.
  - The null object, `NullHeartbeatThread`, does nothing when the storage does not support heartbeat.
  - The factory method, `get_heartbeat_thread`, determins which child class should be instantiated.


g-votte:
As #3596 has been merged, this PR is ready for review. PTAL.

HideakiImamura:

optuna/storages/_heartbeat.py
```diff
@@ -68,6 +70,59 @@ def get_failed_trial_callback(self) -> Optional[Callable[["optuna.Study", Frozen
         raise NotImplementedError()
 
 
+class BaseHeartbeatThread:
+    def start(self, trial_id: int) -> None:
+        raise NotImplementedError()
+
+    def join(self) -> None:
+        raise NotImplementedError()
```
toshihikoyanase, line 78: [Question] This class has only abstract methods. Could you explain why you avoid `ABCMeta` unlike the `BaseStorage` and `BaseSampler`?

https://github.com/optuna/optuna/blob/07a8193c34e5e389770c6ae549001bb574510e7a/optuna/storages/_base.py#L22

optuna/study/_optimize.py
```diff
@@ -194,15 +192,9 @@ def _run_trial(
...
-    thread: Optional[Thread] = None
 
-    if is_heartbeat_enabled(study._storage):
-        assert isinstance(study._storage, BaseHeartbeat)
-        heartbeat = study._storage
-        stop_event = Event()
-        thread = Thread(target=_record_heartbeat, args=(trial._trial_id, heartbeat, stop_event))
-        thread.start()
+    heartbeat_thread = get_heartbeat_thread(study._storage)
+    heartbeat_thread.start(trial._trial_id)
```
toshihikoyanase, line 197: [Question] Can `get_heartbeat_thread` create a context-manager?

I guess it enable us to use the `with` statement to ensure that `heartbeat_thread.join` is always called. I prototyped it and it seemed to work as far as I checked in unit tests (i.e., `tests/study_tests`).

https://github.com/g-votte/optuna/compare/refactor-heartbeat-thread...toshihikoyanase:heatbeat-context-manager?expand=1

toshihikoyanase (commented):
Thank you for your PR. It indeed simplifies `_optimize.py` and sounds reasonable to me.
I have two questions about the design of `HeartbeatThread`.

optuna/storages/_heartbeat.py
g-votte: Right; this definitely can be an ABCMeta class. I changed there.

optuna/study/_optimize.py
g-votte: Cool. I agree that supporting the `with` scope makes the resource management easier and safer.

I committed a change based on your proposal: https://github.com/optuna/optuna/pull/3609/commits/148993accf07a36cf4d90b921b051477d03757e2

There're some additional changes to remove `trial_id` from the `start` arguments, which make `__enter__` and `start` more symmetry.

toshihikoyanase (commented):
Thank you for your update. Removing `trial_id` from `start` sounds reasonable to me.
I confirmed that `get_heartbeat_thread` enables users to use the heartbeat feature with ask-and-tell interface. So, the change looks good in terms of the functionality.

```python
for _ in range(5):
    trial = study.ask()

    with get_heartbeat_thread(trial._trial_id, study._storage):
        x = trial.suggest_float("x", 0.0, 1.0)

        if random.random() < 0.5:
            continue

        study.tell(trial, f(x))

print(study.trials_dataframe())
```

BTW, `BaseHeartbeat` class has detailed docstrings, but `HeartbeatThread` -related classes don't. Do we intentionally omit the docstrings?


<details>
<summary>Full example code of Ask-and-Tell with heartbeat</summary>

```python
import random
import time

import optuna
import optuna.storages._heartbeat
from optuna.storages._heartbeat import get_heartbeat_thread


def f(x: float) -> float:
    return x

study = optuna.create_study(
    direction="maximize",
    storage=optuna.storages.RDBStorage(
        url="sqlite:///tmp.db",
        heartbeat_interval=1,
        grace_period=2,
    ),
)

for i in range(5):
    trial = study.ask()

    with get_heartbeat_thread(trial._trial_id, study._storage):
        x = trial.suggest_float("x", 0.0, 1.0)

        if random.random() < 0.5:
            continue

        study.tell(trial, f(x))

print(study.trials_dataframe())

time.sleep(4)

print(study.trials_dataframe())
```
</details>

optuna/storages/_heartbeat.py
toshihikoyanase: Thanks!

g-votte:
@toshihikoyanase 
Thanks for your comment.

I think we can add docstring later, after this API gets more stable. In this PR, let's just target to extract the internal logic just as an internal, prototype API.

toshihikoyanase:
> I think we can add docstring later, after this API gets more stable. We might need some more discussion to actually extend this API to ask-and-tell, e.g., how to hide `_trial_id` from the user.

Sounds reasonable. Thank you for your explanation.

toshihikoyanase (approved):
LGTM!

contramundum53 (commented):
Thanks, the code becomes clearer after refactoring. I have some minor comments.

optuna/study/_optimize.py
```diff
@@ -194,32 +192,18 @@ def _run_trial(
...
-    if is_heartbeat_enabled(study._storage):
-        assert stop_event is not None
-        assert thread is not None
-        stop_event.set()
-        thread.join()
+    with get_heartbeat_thread(trial._trial_id, study._storage):
+        try:
+            value_or_values = func(trial)
+        except exceptions.TrialPruned as e:
+            # TODO(mamu): Handle multi-objective cases.
```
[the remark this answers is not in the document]
g-votte: I think this is because pruning has not yet supported for multi-objective optimization.

optuna/storages/_heartbeat.py
```diff
@@ -68,6 +72,76 @@ def get_failed_trial_callback(self) -> Optional[Callable[["optuna.Study", Frozen
...
+        self._thread = Thread(
+            target=self._record_heartbeat, args=(self._trial_id, self._heartbeat, self._stop_event)
+        )
+        self._thread.start()
+
+    def join(self) -> None:
+        assert self._stop_event is not None
+        assert self._thread is not None
+        self._stop_event.set()
+        self._thread.join()
```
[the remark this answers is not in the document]
g-votte: After offline discussion with @toshihikoyanase and @contramundum53, we couldn't fully reach consensus: `stop`, `join`, `stop_and_join`, `exit`.

Since this choice is mostly a matter of taste, let's revisit it later, when we expose this API to the user. So far, I think we can just leave it `join`.

PR #3629, opened 2022-06-04
Add `skip_if_exists` argument to `enqueue_trial`

### LINKED ISSUE #3616: Add a `skip_if_exists` option to `Study.enqueue_trial`
### Motivation

Currently, `Study.enqueue_trial` does not support a `skip_if_exists` option, so it is possible to unintentionally add duplicate trials with the same parameters to the `Study`. For example, in a clustered environment, where processes are often preempted, it is often the case that the parameters of a previously running trial are enqueued back into the study when the process is restarted. In such cases, it is undesirable to repeatedly insert the same parameters.

This PR suggests to add a `skip_if_exists` option to `Study.enqueue_trial`.

### Description

Please add a `skip_if_exists` option to `Study.enqueue_trial`. This option has a value of `bool`. If `Study.enqueue_trial(params, skip_if_exists=True)` and the `Study` already includes the trial with given `params`, the trial is not enqueued (skipped).

Thr `np.isclose` may be used to determine the identity of `params` and parameters of existing trials.

### Alternatives (optional)

_No response_

_No response_

xadrianzetx:
## Motivation

Resolves #3616.

## Description of the changes

* Implemented `skip_if_exists` argument in `enqueue_trial`
* Added tests & updated docs


xadrianzetx:
I had to slightly change the approach to ensure categorical types are handled correctly. Sorry for pushing changes after reviewers have been assigned.

tests/study_tests/test_study.py
```diff
@@ -687,6 +687,98 @@ def objective(trial: Trial) -> float:
...
+        study.optimize(objective, n_trials=1)
+
+        unfixed_params = {"x", "y", "z"} - set(new_params)
+        t1 = study.trials[1]
+        assert all(t1.params[k] == new_params[k] for k in new_params)
+        assert all(-10 <= t1.params[k] <= 10 for k in unfixed_params)
+
+
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES)
+@pytest.mark.parametrize("param", ["foo", 1, 1.1])
```
c-bata, line 772: How about checking edge cases like the following?

* `float('inf')` / `float('-inf')` / `float('nan')`
* `1e17` / `1e-17`
* None

optuna/study/study.py
```diff
@@ -786,13 +789,18 @@ def objective(trial):
                 Parameter values to pass your objective function.
             user_attrs:
                 A dictionary of user-specific attributes other than ``params``.
+            skip_if_exists:
+                When :obj:`True`, prevents duplicate trials from being enqueued again.
```
c-bata, line 793: Not a strong opinion, but how about updating a docstring to describe the method may accept duplicated parameters if the same parameters are enqueued at the same time?

c-bata (commented):
Thank you for your pull request! I leave some review comments.

optuna/study/study.py
```diff
@@ -928,6 +936,18 @@ def _pop_waiting_trial_id(self) -> Optional[int]:
 
         return None
 
+    def _should_skip_enqueue(self, params: Dict[str, Any]) -> bool:
+
+        for trial in self.get_trials(deepcopy=False):
+            trial_params = trial.system_attrs.get("fixed_params", trial.params)
+            if trial_params == params:
```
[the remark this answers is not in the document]
xadrianzetx: That was my initial approach, but implementation turns out a bit complicated, since we need to check for types now, so I figured precision in cpython `dict_equal` is enough for this API. But I see where `json` encoding-decoding can introduce float representation related bugs, so if you don't mind slightly complicated  type checking, I'm fine with bringing back implementation from 56126a21e78957f6834193cbb4be490368fa2ba0 (with less bugs this time :^)).

optuna/study/study.py
xadrianzetx: 8a3bbc13da7e54592a1f7e18251df43c88ec2a4c :+1:

optuna/study/study.py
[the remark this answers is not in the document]
c-bata: Sounds good :+1: I think slightly complicated type checking is acceptable for this purpose.

Furthermore, `float('nan') == float('nan')` will be false actually. It caused a bug that is reported at #3556. To be honest, I'm not sure we should support the edge case. However, given that we've fixed #3556, it seems that we need to support the `float('nan')` input here.

keisuke-umezawa:
Seems good to me apart from the comments from @c-bata !

optuna/study/study.py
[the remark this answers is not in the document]
xadrianzetx: Done in cf6810e645ca24a21cd060017f1fc329fa5f38d5 (somehow awkwardly). PTAL @c-bata.

tests/study_tests/test_study.py
xadrianzetx: Done in f4f6b5fd83ea5daeda161482ec09f0de1f47a896.

optuna/study/study.py
```diff
@@ -928,6 +945,31 @@ def _pop_waiting_trial_id(self) -> Optional[int]:
...
+            for param_name, param_value in params.items():
+                existing_param = trial_params[param_name]
+                if not isinstance(param_value, type(existing_param)):
+                    # Enqueued param has distribution that does not match existing param
+                    # (e.g. trying to enqueue categorical to float param).
+                    # We are not doing anything about it here, since sanitization should
+                    # be handled regardless if `skip_if_exists` is `True`.
+                    continue
+
+                if isinstance(param_value, (float, np.floating)):
```
c-bata, line 965: In https://github.com/optuna/optuna/pull/3567/files#r879072605, we use `isinstance(..., Real)` to handle float-like types like `np.floating`. Can we use `numbers.Real` here as well?

optuna/study/study.py
```diff
@@ -928,6 +945,31 @@ def _pop_waiting_trial_id(self) -> Optional[int]:
...
+                    if np.isnan(param_value) or np.isclose(param_value, existing_param):
```
c-bata, line 966: Should we prefer `atol=0.0` as described in https://numpy.org/doc/stable/reference/generated/numpy.isclose.html?

```suggestion
                    if np.isnan(param_value) or np.isclose(param_value, existing_param, atol=0.0):
```

optuna/study/study.py
```diff
@@ -928,6 +945,31 @@ def _pop_waiting_trial_id(self) -> Optional[int]:
...
+                        return True
+                else:
+                    if param_value == existing_param:
+                        return True
```
c-bata, line 970: [nit] Can we reduce the nesting like the following?

```suggestion
                elif param_value == existing_param:
                    return True
```

c-bata (commented):
Thanks for the update. I leave some suggestions.

optuna/study/study.py
xadrianzetx: I'm afraid `mypy` is not happy with that change - `isclose` accepts just about any type, but not `Real`. :sweat_smile:

optuna/multi_objective/study.py
```diff
@@ -402,7 +402,7 @@ def enqueue_trial(self, params: Dict[str, Any]) -> None:
                 Parameter values to pass your objective function.
         """
 
-        self._study.enqueue_trial(params)
+        self._study.enqueue_trial(params, skip_if_exists=False)
```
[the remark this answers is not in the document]
c-bata: That makes sense! Thanks for the update.

optuna/study/study.py
c-bata: Yes, you are correct. In #3567, `Real` values are casted to `float` to call `np.isnan()`.

https://github.com/optuna/optuna/blob/b213753a816689a97b6433e1e780ca32c4293e07/optuna/distributions.py#L464-L465

Do you think it make sense to cast values with `float()` like:

```python
is_repeated = (
    np.isnan(float(param_value)) or np.isclose(float(param_value), float(existing_param), atol=0.0)
    if isinstance(param_value, Real)
    else param_value == existing_param
)
```

c-bata (commented):
Sorry for the late response. Changes looks almost good to me except for https://github.com/optuna/optuna/pull/3629#discussion_r899931682 👍

optuna/study/study.py
xadrianzetx: Sounds reasonable. Implemented in 87e3c27183533447a0de3561715db62d29d84526.

c-bata (approved):
LGTM!

optuna/study/study.py
```diff
@@ -928,6 +946,37 @@ def _pop_waiting_trial_id(self) -> Optional[int]:
...
+
+            repeated_params: List[bool] = []
+            for param_name, param_value in params.items():
+                existing_param = trial_params[param_name]
+                if not isinstance(param_value, type(existing_param)):
+                    # Enqueued param has distribution that does not match existing param
+                    # (e.g. trying to enqueue categorical to float param).
+                    # We are not doing anything about it here, since sanitization should
+                    # be handled regardless if `skip_if_exists` is `True`.
+                    continue
```
keisuke-umezawa, line 965: [nits] If you add `repeated_params.append(False)` here, you do not need to check `len(repeated_params) > 0`. And you can just return `all(repeated_params)` at the end.

keisuke-umezawa (commented):
I just left a nits comment. Could you check it?

optuna/study/study.py
```diff
@@ -753,7 +756,10 @@ def objective(trial):
         self._stop_flag = True
 
     def enqueue_trial(
-        self, params: Dict[str, Any], user_attrs: Optional[Dict[str, Any]] = None
+        self,
+        params: Dict[str, Any],
+        user_attrs: Optional[Dict[str, Any]] = None,
+        skip_if_exists: bool = True,
```
not522, line 762: I prefer that the default value of `skip_if_exists` is `False` because skipping an enqueued trial may not be intuitive for users.

optuna/study/study.py
```diff
@@ -786,13 +792,25 @@ def objective(trial):
...
         .. seealso::
 
             Please refer to :ref:`enqueue_trial_tutorial` for the tutorial of specifying
             hyperparameters manually.
         """
 
+        if skip_if_exists and self._should_skip_enqueue(params):
+            warnings.warn(
+                f"Trial with params {params} already exists. Skipping enqueue.", RuntimeWarning
+            )
```
not522, line 811: I think it should be `INFO` because it is expected behavior.

optuna/study/study.py
xadrianzetx: a82dbabc7d52b188c9272e8d8e1502f772c2a5da :+1:

optuna/study/study.py
xadrianzetx: Switched to `logging.info` in 37ea9702a3349110a82ec74c9ea7ee29717dec26 and made tests independent from logger messages in 221bd19ef70999cfed517759ff7a16aba30e5ea4.

optuna/study/study.py
xadrianzetx: Thanks for the suggestion. I've removed length check in 3b99b8436bcf2e95836369b1b2c4d96d28c55b13, but otherwise early breakout logic remains unchanged, otherwise we would always return after checking only first trial.

keisuke-umezawa (approved):
LGTM! Thank you for your quick updates.

PR #3642, opened 2022-06-07
Fix document build of v2.10.1

toshihikoyanase:
## Motivation

We'd like to build the reference of Optuna v2.10.x. I struggled in https://github.com/optuna/optuna/pull/3639 and I found that many libraries were supposed to be fixed to pass CIs.

Then, we decided to add minimum changes for the document build and ignore CI errors caused by the update of dependencies.

## Description of the changes

- Pin Sphinx and related libraries.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,13 +66,17 @@ def get_extras_require() -> Dict[str, List[str]]:
             "mlflow",
         ],
         "document": [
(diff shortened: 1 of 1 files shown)


contramundum53 (approved):
Thanks! LGTM!

PR #3643, opened 2022-06-07
Remove experimental from `plot_pareto_front`

### LINKED ISSUE #2973: Stabilize `plot_pareto_front`
## Motivation

Stable API of multi-objective optimization has been available since [Optuna v2.4.0](https://github.com/optuna/optuna/releases/tag/v2.4.0), but the visualization for multi-objective optimization, i.e., [`optuna.visualization.plot_pareto_front`](https://optuna.readthedocs.io/en/stable/reference/visualization/generated/optuna.visualization.plot_pareto_front.html#optuna.visualization.plot_pareto_front) and [`optuna.visualization.matplotlib.plot_pareto_front`](https://optuna.readthedocs.io/en/stable/reference/visualization/generated/optuna.visualization.matplotlib.plot_pareto_front.html#optuna.visualization.matplotlib.plot_pareto_front), is still experimental.

## Description

Like the other plot functions, we'd like to stabilize `optuna.visualization.plot_pareto_front` first, then `optuna.visualization.matplotlib.plot_pareto_front` will follow it.

We have the following points to discuss to fix the API of `plot_pareto_front`.

### Arguments

Currently, the function has three arguments: `target_names`, `include_domintated_trials` and `axis_order`.
We'd like to discuss the types and names of them. The following is just examples of design choices:
- `target_names` or `objective_names`
- The default value of `include_dominated_trials`
- `axis_order` or `objective_order`

### Extension

As proposed in https://github.com/optuna/optuna/issues/2397, we may want to extend the function in the future. For example, `NSGAIISampler` and `BoTorchSampler` support the constrained optimization, but we cannot distinguish feasible solutions from infeasible ones in the current Pareto front plots.

### Many objective optimization support

Currently, the function raises errors when studies have more than three objectives. We have some design choices and should be discussed before stabilizing the API.

Possible alternatives are as follows:
- Choose two or three objectives
- Tile 2D plot like the contour plot for more than three parameters.

### Environment

Currently, we check the function using the unit tests and the readthedocs documents. However, users often use the function in notebooks such as [Jupyter notebook](https://jupyter.org/), [JupyterLab](https://jupyter.org/), and [Google Colab](https://colab.research.google.com/). So, the function should be verified in such various environments. Note that the 3D plot is not shown in JupyterLab and should be fixed,

## Related issues

- Currently, all matplotlib-based visualization functions are experimental. See https://github.com/optuna/optuna/issues/2959.
- Possible extension of `plot_pateto_front` can be fond in https://github.com/optuna/optuna/issues/2397.

contramundum53:
## Motivation
Resolve #2973.

## Description of the changes
* Remove experimental from `plot_pareto_front` function.
* Add experimental note and warning for `constraint_func` argument, because it is newly introduced in `v3.0.0`.

### DIFF
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -13,6 +13,7 @@ import warnings
 
 import optuna
 from optuna._experimental import experimental
+from optuna.exceptions import ExperimentalWarning
 from optuna.study import Study
 from optuna.study._multi_objective import _get_pareto_front_trials_by_trials
 from optuna.trial import FrozenTrial
@@ -99,6 +100,11 @@ def plot_pareto_front(
             If given, trials are classified into three categories: feasible and best, feasible but
             non-best, and infeasible. Categories are shown in different colors. Here, whether a
             trial is best (on Pareto front) or not is determined ignoring all infeasible trials.
+
+            .. note::
+                Added in v3.0.0 as an experimental feature. The interface may change in newer
+                versions without prior notice.
+                See https://github.com/optuna/optuna/releases/tag/v3.0.0.
         targets:
             A function that returns targets values to display.
             The argument to this function is :class:`~optuna.trial.FrozenTrial`.
@@ -140,6 +146,12 @@ def plot_pareto_front(
             ),
         ]
     else:
+        warnings.warn(
+            "``constraint_func`` argument is an experimental feature."
+            " The interface can change in the future.",
+            ExperimentalWarning,
+        )
+
         data = [
             _make_scatter_object(
                 info.n_targets,
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -12,7 +12,6 @@ from typing import Union
 import warnings
 
 import optuna
-from optuna._experimental import experimental
 from optuna.exceptions import ExperimentalWarning
 from optuna.study import Study
 from optuna.study._multi_objective import _get_pareto_front_trials_by_trials
@@ -36,7 +35,6 @@ class _ParetoFrontInfo(NamedTuple):
     axis_order: Sequence[int]
 
 
-@experimental("2.4.0")
 def plot_pareto_front(
     study: Study,
     *,
--- a/tests/visualization_tests/matplotlib_tests/test_pareto_front.py
+++ b/tests/visualization_tests/matplotlib_tests/test_pareto_front.py
@@ -497,7 +497,7 @@ def test_constraints_func_experimental_warning() -> None:
     study = optuna.create_study(directions=["minimize", "minimize"])
 
     def constraints_func(t: FrozenTrial) -> Sequence[float]:
-        return [1.0] if t.params["x"] == 1 and t.params["y"] == 0 else [-1.0]
+        return [1.0]
 
     with pytest.warns(optuna.exceptions.ExperimentalWarning):
         plot_pareto_front(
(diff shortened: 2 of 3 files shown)


optuna/visualization/_pareto_front.py
```diff
@@ -140,6 +144,12 @@ def objective(trial):
             ),
         ]
     else:
+        warnings.warn(
+            "``constraint_func`` argument is an experimental feature."
```
toshihikoyanase, line 148: A nit.
```suggestion
            "``constraints_func`` argument is an experimental feature."
```

toshihikoyanase: BTW, warning messages are not rendered as ReST, so we may not need to quote with the double back-ticks. It follows the existing code base, though.

https://github.com/optuna/optuna/blob/07a8193c34e5e389770c6ae549001bb574510e7a/optuna/samplers/_tpe/sampler.py#L255-L260

optuna/visualization/_pareto_front.py
```diff
@@ -140,6 +144,12 @@ def objective(trial):
...
+            " The interface can change in the future.",
+            ExperimentalWarning,
```
toshihikoyanase, line 150: Should we check this warning in unit tests?
The `multivariate` of `TPESampler` has such a test case.

https://github.com/optuna/optuna/blob/07a8193c34e5e389770c6ae549001bb574510e7a/tests/samplers_tests/tpe_tests/test_sampler.py#L35-L37

optuna/visualization/_pareto_front.py
contramundum53: Thanks for pointing that out! I applied your suggestion.

optuna/visualization/_pareto_front.py
contramundum53: I added such a test in `test_pareto_front.py`.

HideakiImamura:

tests/visualization_tests/matplotlib_tests/test_pareto_front.py
```diff
@@ -491,3 +491,16 @@ def test_plot_pareto_front_using_axis_order_and_targets() -> None:
...
+    study = optuna.create_study(directions=["minimize", "minimize"])
+
+    def constraints_func(t: FrozenTrial) -> Sequence[float]:
+        return [1.0] if t.params["x"] == 1 and t.params["y"] == 0 else [-1.0]
+
+    with pytest.warns(optuna.exceptions.ExperimentalWarning):
+        plot_pareto_front(
+            study=study,
+            constraints_func=constraints_func,
+        )
```
toshihikoyanase, line 506: The `constraints_func` seems to be a bit complicated. It refers to `t.params["x"]` and `t.params["y"]` although the study is empty.

Developers might be able to focus on the experimental warnings if we use simple `constraints_func`. The following is just an example, but can we remove the `t.params` in `constraints_func`?

```
    with pytest.warns(optuna.exceptions.ExperimentalWarning):
        plot_pareto_front(study=study, constraints_func=lambda _: [1.0])
```

optuna/visualization/_pareto_front.py
toshihikoyanase: Thanks!

tests/visualization_tests/matplotlib_tests/test_pareto_front.py
contramundum53: Thanks, I applied your suggestion.

toshihikoyanase (approved):
Thank you for your updates. LGTM.

[Notes] I confirmed that the experimental warning was actually shown as follows:
image

knshnb (approved):
Thanks for the PR! LGTM.

PR #3658, opened 2022-06-11
[hotfix] Add onnx and version constrained protobuf to document dependencies

nzw0301:
## Motivation

The master branch's documentation builds are broken due to missing onnx, ex: https://github.com/optuna/optuna/runs/6841134629?check_suite_focus=true.
 
## Description of the changes

As a hotfix, 
- Add `onnx` to `document` of `setup.py` because `thop` doesn't install onnx explicitly.
- Add protobuf version constraint by following onnx's [`requirements.txt`](https://github.com/onnx/onnx/blob/64e79b78a410e69a804f4f198bbaf45193fb3c32/requirements.txt). Otherwise documentation builds are broken, ex. https://github.com/optuna/optuna/runs/6841513286?check_suite_focus=true.

PR #3659, opened 2022-06-11
Remove the implementation of `read_trials_from_remote_storage` in the all storages apart from CachedStorage

keisuke-umezawa:
## Motivation

The method `read_trials_from_remote_storage` is an abstract method of `BaseStorage`, and is expected to be called for the `Stronger consistency requirements for special data` as follows:

```
    **Stronger consistency requirements for special data**
    Under a multi-worker setting, a storage class must return the latest values of any attributes
    of a study, not necessarily for the attributes of a `Trial`.
    However, if the `read_trials_from_remote_storage(study_id)` method is called, any successive
    reads on the `state` attribute of a `Trial` are guaranteed to return the same or more recent
    values than the value at the time of the call to the
    `read_trials_from_remote_storage(study_id)` method.
```

However, currently, all the other storages apart from `_CachedStorage` naturally hold this `Stronger consistency requirements for special data` without calling `read_trials_from_remote_storage`, and it is only needed for CachedStorage.

It means that we can assume that this method is only for CachedStorage to sync data from remote to local. Therefore, we decided that we remove the implementation of read_trials_from_remote_storage in the other storages apart from CachedStorages.

## Description of the changes

In [the PR](https://github.com/optuna/optuna/pull/3437), I removed `read_trials_from_remote_storage` from all storages, and CachedStorage automatically sync from the remote storage when it calls `get_all_trials` . But, the solution has some concerns to degrade performance of the network traffic.

In this PR, I just removed the implementation of `read_trials_from_remote_storage` in the all storages apart from `CachedStorage`. And, make `read_trials_from_remote_storage` called only when the storage is `CachedStorage`.

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -60,8 +60,8 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
     of a study and a trial. Generally, typical storages naturally hold this requirement. However,
     :class:`~optuna.storages._CachedStorage` does not, so we introduce the
     `read_trials_from_remote_storage(study_id)` method in the class. The detailed explanation how
-    :class:`~optuna.storages._CachedStorage` aquires this requirement, you can check this
-    class docstrings.
+    :class:`~optuna.storages._CachedStorage` aquires this requirement, is available at
+    the docstring.
 
     .. note::
 
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -59,7 +59,7 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
     Under a multi-worker setting, a storage class must return the latest values of any attributes
     of a study and a trial. Generally, typical storages naturally hold this requirement. However,
     :class:`~optuna.storages._CachedStorage` does not, so we introduce the
-    `read_trials_from_remote_storage(study_id)` method in the class.The detailed explanation how
+    `read_trials_from_remote_storage(study_id)` method in the class. The detailed explanation how
     :class:`~optuna.storages._CachedStorage` aquires this requirement, you can check this
     class docstrings.
 
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -54,6 +54,25 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
     any successive reads on data `X` from the same process must read the written
     value or one of the more recent values.
 
+    **Stronger consistency requirements for special data**
+
+    Under a multi-worker setting, a storage class must return the latest values of any attributes
+    of a study, not necessarily for the attributes of a `Trial`.
+    However, if the `read_trials_from_remote_storage(study_id)` method is called, any successive
+    reads on the `state` attribute of a `Trial` are guaranteed to return the same or more recent
+    values than the value at the time of the call to the
+    `read_trials_from_remote_storage(study_id)` method.
+    Let `T` be a `Trial`.
+    Let `P` be the process that last updated the `state` attribute of `T`.
+    Then, any reads on any attributes of `T` are guaranteed to return the same or
+    more recent values than any writes by `P` on the attribute before `P` updated
+    the `state` attribute of `T`.
+    The same applies for `user_attrs', 'system_attrs' and 'intermediate_values` attributes.
+
+    .. note::
+
+        These attribute behaviors may become user customizable in the future.
+
     **Data persistence**
 
     A storage class does not guarantee that write operations are logged into a persistent
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -57,17 +57,11 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
     **Stronger consistency requirements for special data**
 
     Under a multi-worker setting, a storage class must return the latest values of any attributes
-    of a study, not necessarily for the attributes of a `Trial`.
-    However, if the `read_trials_from_remote_storage(study_id)` method is called, any successive
-    reads on the `state` attribute of a `Trial` are guaranteed to return the same or more recent
-    values than the value at the time of the call to the
-    `read_trials_from_remote_storage(study_id)` method.
-    Let `T` be a `Trial`.
-    Let `P` be the process that last updated the `state` attribute of `T`.
-    Then, any reads on any attributes of `T` are guaranteed to return the same or
-    more recent values than any writes by `P` on the attribute before `P` updated
-    the `state` attribute of `T`.
-    The same applies for `user_attrs', 'system_attrs' and 'intermediate_values` attributes.
+    of a study and a trial. Generally, typical storages naturally hold this requirement. However,
+    :class:`~optuna.storages._CachedStorage` does not, so we introduce the
+    `read_trials_from_remote_storage(study_id)` method in the class.The detailed explanation how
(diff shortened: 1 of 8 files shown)


keisuke-umezawa:
I will rebase to master after [this PR](https://github.com/optuna/optuna/pull/3658) is merged to pass CIs.

optuna/storages/_base.py
```diff
@@ -54,25 +54,6 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
     any successive reads on data `X` from the same process must read the written
     value or one of the more recent values.
 
-    **Stronger consistency requirements for special data**
```
keisuke-umezawa, line 57: Where should I put this comment? Can it be moved to `storages._cached_storage.py`?

c-bata:
@HideakiImamura @nzw0301 Could you review this PR?

@keisuke-umezawa I've just merged #3658. Could you merge or rebase master branch?

nzw0301:
Of course. But reviewing this PR might take time because I'm not familiar with storage logic.

c-bata:
Thank you for letting me know! No, problem. You may take your time.

nzw0301:
I suppose that we still need to discuss the point raised by https://github.com/optuna/optuna/pull/3437#pullrequestreview-930230216. Even if we remove the concrete implementations except for `CachedStorage`, I would like to keep the docstring in `BaseStorage` because it is necessary when Optuna decides to support a new storage class (graphQL? MongoDB? IDK).

If there were some discussion on this point, please kindly let me know where I should check.

optuna/storages/_base.py
HideakiImamura: I'm not sure, but I think we don't have to move this comment, since the BaseStorage and _CachedStorage are not public. We should consider the appropriate place of this comment when we make them public.

keisuke-umezawa:
Just rebased.

optuna/storages/_base.py
keisuke-umezawa: Ok. I reverted it.

c-bata:
@keisuke-umezawa Could you please respond to nozawa-san's question or provide something context (https://github.com/optuna/optuna/pull/3659#issuecomment-1153583813)?

keisuke-umezawa:
> I suppose that we still need to discuss the point raised by https://github.com/optuna/optuna/pull/3437#pullrequestreview-930230216. Even if we remove the concrete implementations except for CachedStorage, I would like to keep the docstring in BaseStorage because it is necessary when Optuna decides to support a new storage class (graphQL? MongoDB? IDK). If there were some discussion on this point, please kindly let me know where I should check.

Currently, all the storages apart from `CachedStorage` naturally support `Stronger consistency requirements for special data` without calling `read_trials_from_remote_storage`. Only `CachedStorage` needs `read_trials_from_remote_storage`.

So, in the base storage, should we say something as follows?

```
    **Stronger consistency requirements for special data**

    Under a multi-worker setting, a storage class must return the latest values of any attributes
    of a study and the attributes of a `Trial`.
    However, in the `CachedStorage`, it is guaranteed by `read_trials_from_remote_storage(study_id)` method. 
    If the `read_trials_from_remote_storage(study_id)` method is called, any successive
    reads on the `state` attribute of a `Trial` are guaranteed to return the same or more recent
    values than the value at the time of the call to the
    `read_trials_from_remote_storage(study_id)` method.
    Let `T` be a `Trial`.
    Let `P` be the process that last updated the `state` attribute of `T`.
    Then, any reads on any attributes of `T` are guaranteed to return the same or
    more recent values than any writes by `P` on the attribute before `P` updated
    the `state` attribute of `T`.
    The same applies for `user_attrs', 'system_attrs' and 'intermediate_values` attributes.
```

nzw0301:
> Currently, all the storages apart from `CachedStorage` naturally support Stronger consistency requirements for special data without calling `read_trials_from_remote_storage`. Only `CachedStorage` needs `read_trials_from_remote_storage`.

Thanks.  I'm still not sure why we refactor this method. [As I said](https://github.com/optuna/optuna/pull/3659#issuecomment-1153583813), if optuna adds another storage class, we might implement `read_trials_from_remote_storage` for consistency, as a result, the codebase becomes complicated depending on the storage backend. Therefore I would like to suggest 

1. keeping `read_trials_from_remote_storage` method, whose implementation is `pass`, not `raise NotImplementedError` if we don't really need checking to exist of the study given study id. Otherwise keeping the current implementation.
2. keeping its documentation base storage class, and adding requirements of `read_trials_from_remote_storage`: checking the existence of study id (if necessary) and guaranteeing consistency. 
3. add an explanation to each concrete implementation (for example, Since `RDBStorage` has the consistency, we only skip the guarantee of consistency.

What do you think?

contramundum53:
@HideakiImamura -san, let us de-assign you because you look busy.
@c-bata -san, could you please review this PR?

nzw0301:
Thank you for your explanation. Fair enough. So could you add more comments about consistency to `_CachedStorage.read_trials_from_remote_storage`?

nzw0301:
I think mentioning the following order might be informative for future committers:

1. the requirement as in the current docstring and then 
2. Usual storage classes don't require explicit implementation except `CachedStorage` (this part can be moved to another place if there is another right place)

keisuke-umezawa:
@nzw0301
Thank you for you comment!
I updated the docstrings in [3609fee](https://github.com/optuna/optuna/pull/3659/commits/3609feecc6c9e051fb53745f227d0020c880a943)
Could you check that?

optuna/storages/_base.py
```diff
@@ -57,17 +57,11 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
-    `read_trials_from_remote_storage(study_id)` method.
-    Let `T` be a `Trial`.
-    Let `P` be the process that last updated the `state` attribute of `T`.
-    Then, any reads on any attributes of `T` are guaranteed to return the same or
-    more recent values than any writes by `P` on the attribute before `P` updated
-    the `state` attribute of `T`.
-    The same applies for `user_attrs', 'system_attrs' and 'intermediate_values` attributes.
+    of a study and a trial. Generally, typical storages naturally hold this requirement. However,
+    :class:`~optuna.storages._CachedStorage` does not, so we introduce the
+    `read_trials_from_remote_storage(study_id)` method in the class.The detailed explanation how
```
nzw0301, line 62:
```suggestion
    `read_trials_from_remote_storage(study_id)` method in the class. The detailed explanation how
```

nzw0301 (changes requested):
Thanks. The docsting looks informative for future committers. But I still might miss the motivation or the context of this PR. So I'm happy to revert my suggestions to improve this PR if my comments were not appropriate.

optuna/storages/_base.py
```diff
@@ -57,17 +57,11 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
...
+    :class:`~optuna.storages._CachedStorage` aquires this requirement, you can check this
+    class docstrings.
```
c-bata, line 64:
```suggestion
    :class:`~optuna.storages._CachedStorage` aquires this requirement, is available at
    the docstring.
```

c-bata (commented):
LGTM with nits.

nzw0301:
I think this PR is almost done, so let me commit two typos suggested by reviewers (including myself).

c-bata (approved):
LGTM!

PR #3664, opened 2022-06-13
Remove duplicate calls to `study.get_trials` in `ImportanceEvaluator`s

contramundum53:
## Motivation
This is a followup PR from #3589. @c-bata had a performance concern in #3589. We resolve the problem by removing the duplicated calls to `study.get_trials`.

## Description of the changes
We change the interface of `BaseImportanceEvaluator.evaluate` to pass the list of `completed_trials`, rather than `study` itself. 

In addition to performance improvement, this also makes it easier to test the functions. However, refactoring the tests is out of scope of this PR.

### DIFF
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from typing import Any
 from typing import Callable
 from typing import Dict
 from typing import List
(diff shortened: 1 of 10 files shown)


contramundum53:
The dependent PR is of low priority and is taking time. Let me remove `stale` label.

Issue #3605, opened 2022-05-31
`RDBStorage` does not sort trails with postgres


pnkov:
### Expected behavior

`RDBStorage.get_all_trials()` should sort trials like `CachedStorage` do when using postgres.
Also, it would be great if `intermediate_values` were sorted too.

### Environment

- Optuna version: 2.10.0
- Python version: 3.8.13
- OS: Linux-3.10.0-1160.62.1.el7.x86_64-x86_64-with-glibc2.2.5

### Error messages, stack traces, or logs

```shell
See steps to reproduce.
```

### Steps to reproduce

1. Perform a study with `postgresql+psycopg2://...` storage.
2. Get trails from `CachedStorage` and `RDBStorage` by your study id.

```python
>>> import optuna

>>> storage = optuna.storages.get_storage('postgresql+psycopg2://...')
>>> [t.number for t in storage.get_all_trials(study_id=...)]
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, ...]
>>> storage.__class__
<class 'optuna.storages._cached_storage._CachedStorage'>

>>> storage = optuna.storages.RDBStorage('postgresql+psycopg2://...')
>>> [t.number for t in storage.get_all_trials(study_id=...)]
[33, 34, 28, 38, 5, 36, 1, 11, 10, 9, 8, 7, 14, 24, 42, 32, 17, 40, 43, 41, 49, 25, 54, 58, 48, ...]
>>> storage.__class__
<class 'optuna.storages._rdb.storage.RDBStorage'>
```

### Additional context (optional)

Found this while fixing plots https://github.com/optuna/optuna-dashboard/pull/235

https://www.postgresql.org/docs/current/sql-select.html says:
> If `ORDER BY` is not given, the rows are returned in whatever order the system finds fastest to produce.

c-bata:
@pnkov Thank you for reporting a bug! I executed a following script with PostgreSQL 14.3 server and Optuna 2.10.0.

```python
import optuna


def objective(trial):
    x = trial.suggest_float("x", -100, 100)
    y = trial.suggest_categorical("y", [-1, 0, 1])
    return x**2 + y


def main():
    storage = optuna.storages.RDBStorage("postgresql+psycopg2://root:root@localhost:5432/optuna")
    study = optuna.create_study(storage=storage)
[...]

PR #3666, opened 2022-06-14
Fix bug in `IntersectionSearchSpace`

contramundum53:
## Motivation
`IntersectionSearchSpace` has a bug that `self._cursor` is not updated when there are no running trials. We fix this bug.

## Description of the changes
Fix `self._cursor` such that it is updated to the last trial when there are no running trials.

### DIFF
--- a/optuna/samplers/_search_space/intersection.py
+++ b/optuna/samplers/_search_space/intersection.py
@@ -73,6 +73,7 @@ class IntersectionSearchSpace:
 
             if not trial.state.is_finished():
                 next_cursor = trial.number
+                continue
 
             if self._search_space is None:
                 self._search_space = copy.copy(trial.distributions)
--- a/optuna/samplers/_search_space/intersection.py
+++ b/optuna/samplers/_search_space/intersection.py
@@ -1,9 +1,6 @@
 from collections import OrderedDict
 import copy
-import itertools
-from typing import Any
 from typing import Dict
-from typing import Generator
 from typing import Optional
 
 import optuna
@@ -69,40 +66,29 @@ class IntersectionSearchSpace:
 
         trials = study.get_trials(deepcopy=False, states=states_of_interest)
 
-        current_cursor = self._cursor
+        next_cursor = trials[-1].number if len(trials) > 0 else -1
+        for trial in reversed(trials):
+            if self._cursor > trial.number:
+                break
 
-        def reversed_new_trials_iter(
-            finished: bool,
-        ) -> Generator[optuna.trial.FrozenTrial, None, None]:
-            nonlocal current_cursor
-            return (
-                trial
-                for trial in itertools.takewhile(
-                    lambda t: t.number < current_cursor, reversed(trials)
-                )
-                if trial.state.is_finished() == finished
-            )
+            if not trial.state.is_finished():
+                next_cursor = trial.number
 
-        dists: Any
-        dists = self._search_space.items() if self._search_space is not None else None
-        for trial in reversed_new_trials_iter(finished=True):
-            if dists is None:
-                dists = trial.distributions.items()
-            else:
-                # We use lambda to capture the value of `trial.distributions`.
-                dists = (
-                    lambda d: (
-                        (param, dist) for param, dist in dists if param in d and dist == d[param]
-                    )
-                )(trial.distributions)
+            if self._search_space is None:
+                self._search_space = copy.copy(trial.distributions)
+                continue
 
-        self._search_space = dict(dists) if dists is not None else None
+            delete_list = []
+            for param_name, param_distribution in self._search_space.items():
+                if param_name not in trial.distributions:
+                    delete_list.append(param_name)
+                elif trial.distributions[param_name] != param_distribution:
+                    delete_list.append(param_name)
 
-        self._cursor = max(
-            (trial.number for trial in reversed_new_trials_iter(finished=False)),
-            default=-1 if len(trials) == 0 else trials[-1].number,
-        )
+            for param_name in delete_list:
+                del self._search_space[param_name]
 
+        self._cursor = next_cursor
         search_space = self._search_space or {}
 
         if ordered_dict:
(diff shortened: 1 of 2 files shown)


c-bata:
@not522 @keisuke-umezawa Could you review this PR?

optuna/samplers/_search_space/intersection.py
```diff
@@ -59,20 +59,20 @@ def calculate(self, study: Study, ordered_dict: bool = False) -> Dict[str, BaseD
             if self._study_id != study._study_id:
                 raise ValueError("`IntersectionSearchSpace` cannot handle multiple studies.")
 
-        states_of_interest = [optuna.trial.TrialState.COMPLETE]
+        states_of_interest = [optuna.trial.TrialState.COMPLETE, optuna.trial.TrialState.RUNNING]
```
not522, line 62:
```suggestion
        states_of_interest = [optuna.trial.TrialState.COMPLETE, optuna.trial.TrialState.WAITING, optuna.trial.TrialState.RUNNING]
```

optuna/samplers/_search_space/intersection.py
```diff
@@ -59,20 +59,20 @@ def calculate(self, study: Study, ordered_dict: bool = False) -> Dict[str, BaseD
...
 
         if self._include_pruned:
             states_of_interest.append(optuna.trial.TrialState.PRUNED)
 
-        next_cursor = self._cursor
-        for trial in reversed(study.get_trials(deepcopy=False)):
+        trials = study.get_trials(deepcopy=False, states=states_of_interest)
+
+        next_cursor = trials[-1].number if len(trials) > 0 else -1
```
not522, line 69: I think that the cursor should point to the youngest unfinished trial.
```suggestion
        next_cursor = trials[-1].number + 1 if len(trials) > 0 else -1
```

not522 (commented):
Could you add tests for this change?

optuna/samplers/_search_space/intersection.py
contramundum53: Thanks, I updated the code.

optuna/samplers/_search_space/intersection.py
contramundum53: Thanks!

contramundum53:
> Could you add tests for this change?

Should I add tests for `_cursor`? I thought that would count as implementation detail.

not522:
I see, so this bug only slows the execution down but does not change the behavior.

not522 (approved):
LGTM

optuna/samplers/_search_space/intersection.py
```diff
@@ -59,20 +59,24 @@ def calculate(self, study: Study, ordered_dict: bool = False) -> Dict[str, BaseD
...
+        ]
 
         if self._include_pruned:
             states_of_interest.append(optuna.trial.TrialState.PRUNED)
 
-        next_cursor = self._cursor
-        for trial in reversed(study.get_trials(deepcopy=False)):
+        trials = study.get_trials(deepcopy=False, states=states_of_interest)
+
+        next_cursor = trials[-1].number + 1 if len(trials) > 0 else -1
```
keisuke-umezawa, line 73: [question] It relies on that the assumption that `get_trials` returns a list of ordered trials, but is it actually guaranteed in all the storages? e.g. [redis](https://github.com/optuna/optuna/blob/master/optuna/storages/_redis.py#L687-L699), [SQL](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/storage.py#L953-L962)

optuna/samplers/_search_space/intersection.py
```diff
@@ -59,20 +59,24 @@ def calculate(self, study: Study, ordered_dict: bool = False) -> Dict[str, BaseD
...
+        for trial in reversed(trials):
             if self._cursor > trial.number:
                 break
 
             if not trial.state.is_finished():
```
keisuke-umezawa, line 78: [question] This clause is continued in the PR, but not in the master branch. Is it ok?

optuna/samplers/_search_space/intersection.py
keisuke-umezawa: They are always wrapped by cached storage and trials are guaranteed to be ordered.
https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L344

optuna/samplers/_search_space/intersection.py
```diff
@@ -59,20 +59,24 @@ def calculate(self, study: Study, ordered_dict: bool = False) -> Dict[str, BaseD
...
                 next_cursor = trial.number
-
-            if trial.state not in states_of_interest:
                 continue
```
keisuke-umezawa, line 81: Could you add `assert trial.state == optuna.trial.TrialState.COMPLETE or trial.state == optuna.trial.TrialState.PRUNED` here, to understand this logic easily? Or, could you add unit tests which guarantee that intersections will not count RUNNING and WAITING?

optuna/samplers/_search_space/intersection.py
keisuke-umezawa: Could you add unit tests to catch this error? 

Currently, we do not check that `trials in the state of RUNNING, WAITING are not counted, but later they are counted after completion` in unit tets. 
Only we checked that FAIL and PRUNED are skipped here.
https://github.com/optuna/optuna/blob/master/tests/samplers_tests/search_space_tests/test_intersection.py#L54-L55

optuna/samplers/_search_space/intersection.py
contramundum53: Yes, we assume that `get_trials` do return the trials in chronological order. However, there is a known issue (#3605) that the returned trials can be unsorted. This issue is to be fixed.

optuna/samplers/_search_space/intersection.py
keisuke-umezawa: @contramundum53 Sorry, I added a comment on the resolved conversation. Could you check it
?

contramundum53: @keisuke-umezawa I added a test case for waiting trial. PTAL.

keisuke-umezawa (approved):
Thank you for adding a unittest to catch the error. LTGM.
I will merge it even though a non-critial ci has not passed.

PR #3668, opened 2022-06-14
Remove `trial_values` records whose values are None

c-bata:
## Motivation

I found an Optuna DB that contains `trial_values` records which `value` column is `NULL` although the `value` column of `trial_values` table is defined as `nullable=False`. It causes a migration error while executing a migration script `v3.0.0.b.py`.
https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/alembic/versions/v3.0.0.b.py#L34-L39

The `value` column was supposed to changed `nullable=False` at #2030. However, the schema migration to make the column nullable is not added in `v2.4.0.a.py` script. Furthermore, it inserts trial_values records for RUNNING, WAITING, and PRUNED trials here:
https://github.com/optuna/optuna/blob/c51af67bde7f67daf9a19227400bc80a51a47ab2/optuna/storages/_rdb/alembic/versions/v2.4.0.a.py#L149-L154

## Description of the changes

The cause of the problem is actually located in `v2.4.0.a.py`. However, about 2.5 years have passed since we released [Optuna v2.4.0](https://github.com/optuna/optuna/releases/tag/v2.4.0). In this PR, I made changes to `v3.0.0.b.py` to fix this issue in the migration script that will be run on the next release of Optuna..

### DIFF
--- a/optuna/storages/_rdb/alembic/versions/v3.0.0.b.py
+++ b/optuna/storages/_rdb/alembic/versions/v3.0.0.b.py
@@ -60,7 +60,7 @@ def upgrade():
         .filter(and_(TrialModel.state == TrialState.COMPLETE, TrialValueModel.value.is_(None)))
         .count()
     ) != 0:
-        raise ValueError("Found invalid trial_values' records (value=None and state='Complete')")
+        raise ValueError("Found invalid trial_values records (value=None and state='COMPLETE')")
     session.query(TrialValueModel).filter(TrialValueModel.value.is_(None)).delete()
 
     with op.batch_alter_table("trial_intermediate_values") as batch_op:
--- a/optuna/storages/_rdb/alembic/versions/v3.0.0.b.py
+++ b/optuna/storages/_rdb/alembic/versions/v3.0.0.b.py
@@ -79,7 +79,7 @@ def upgrade():
         batch_op.alter_column(
             "value",
             type_=Float(precision=FLOAT_PRECISION),
-            existing_nullable=False,
+            nullable=False,
         )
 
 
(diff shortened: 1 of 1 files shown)


c-bata:

HideakiImamura (approved):
LGTM.

optuna/storages/_rdb/alembic/versions/v3.0.0.b.py
```diff
@@ -15,10 +24,45 @@
...
+    bind = op.get_bind()
+    session = Session(bind=bind)
+
+    if (
+        session.query(TrialValueModel)
+        .join(TrialModel, TrialValueModel.trial_id == TrialModel.trial_id)
+        .filter(and_(TrialModel.state == TrialState.COMPLETE, TrialValueModel.value.is_(None)))
+        .count()
+    ) != 0:
+        raise ValueError("Found invalid trial_values' records (value=None and state='Complete')")
```
toshihikoyanase, line 63: A nit. Maybe the state is `COMPLETE` instead of `Complete` according to the definition of `TrialState`.

```suggestion
        raise ValueError("Found invalid trial_values records (value=None and state='Complete')")
```

toshihikoyanase (commented):
Thank you for your PR. I added my early comment.
The logic looks solid, and I'll check the script with optuna-e2e next.

optuna/storages/_rdb/alembic/versions/v3.0.0.b.py
c-bata, line 63: To apply yanase-san's suggestion (See https://github.com/optuna/optuna/pull/3668/files#r901603363).

```suggestion
        raise ValueError("Found invalid trial_values records (value=None and state='COMPLETE')")
```

c-bata: Thanks! I applied your suggestion at https://github.com/optuna/optuna/pull/3668/commits/3a66cf37c9f5f7c2c91e233571f85c59ad8088a8

contramundum53 (approved):

PR #3678, opened 2022-06-15
Add random tests for non-dominated sort

contramundum53:
## Motivation
The tests in `test_nsgaii.py` misses some important edge cases. The missed edge cases include
* `inf` and `-inf` in values/constraints
* duplicated values/constraints
* apparently different constraint values with same meaning (e.g. both are feasible)
etc.
We try to cover these cases by randomly generating test cases and automatically checking whether the output is correct in those cases.

## Description of the changes
* Add random test cases in `test_nsgaii.py`
* Remove existing duplicate tests


contramundum53:
We decided not to introduce randomized test for this. #3686 is a non-randomized version.

PR #3627, opened 2022-06-04
Enable `warn_unused_ignores` for `mypy`

harupy:
Signed-off-by: harupy <hkawamura0130@gmail.com>

## Motivation

#3579

## Description of the changes

Enable `warn_unused_ignores`.

PR #3679, opened 2022-06-15
Workaround `mypy` bug

contramundum53:
## Motivation
There seems to be a bug in `mypy` that does not correctly keep track of whether `# type: ignore` is used. The line reports error if `# type: ignore` is not there, but reports "unused type ignore" error if `# type: ignore` is there. We workaround this problem by using `typing.Any`.

## Description of the changes
* Use `typing.Any` instead of specifying `Tensor` as type.
* There is a bug that the function does not return a list but a scalar. We fix that bug.

### DIFF
--- a/tests/integration_tests/test_botorch.py
+++ b/tests/integration_tests/test_botorch.py
@@ -1,3 +1,4 @@
+from typing import Any
 from typing import cast
 from typing import Optional
 from typing import Sequence
@@ -78,9 +79,9 @@ def test_botorch_candidates_func_invalid_type() -> None:
         train_obj: torch.Tensor,
         train_con: Optional[torch.Tensor],
         bounds: torch.Tensor,
-    ) -> torch.Tensor:
+    ) -> Any:
         # Must be a `torch.Tensor`, not a list.
-        return torch.rand(1).tolist()
+        return [torch.rand(1).tolist()]
 
     sampler = BoTorchSampler(candidates_func=candidates_func, n_startup_trials=1)
 
--- a/tests/integration_tests/test_botorch.py
+++ b/tests/integration_tests/test_botorch.py
@@ -79,7 +79,7 @@ def test_botorch_candidates_func_invalid_type() -> None:
         train_obj: torch.Tensor,
         train_con: Optional[torch.Tensor],
         bounds: torch.Tensor,
-    ) -> Any:
+    ) -> Any:  # The return type is Any because we don't want mypy to check the type.
         # Must be a `torch.Tensor`, not a list.
         return [0.0]
 
--- a/tests/integration_tests/test_botorch.py
+++ b/tests/integration_tests/test_botorch.py
@@ -81,7 +81,7 @@ def test_botorch_candidates_func_invalid_type() -> None:
         bounds: torch.Tensor,
     ) -> Any:
         # Must be a `torch.Tensor`, not a list.
-        return [torch.rand(1).tolist()]
+        return [0.0]
 
     sampler = BoTorchSampler(candidates_func=candidates_func, n_startup_trials=1)
 


c-bata:
@HideakiImamura @nzw0301 Could you review this PR?

toshihikoyanase:
The `checks` CI job passed without this workaround (See the job corresponding to https://github.com/optuna/optuna/pull/3589/commits/138fc443b2b5168e177fab4e8572fd196db6bc5d). Do we still need it?

tests/integration_tests/test_botorch.py
```diff
@@ -78,9 +79,9 @@ def candidates_func(
         train_obj: torch.Tensor,
         train_con: Optional[torch.Tensor],
         bounds: torch.Tensor,
-    ) -> torch.Tensor:
+    ) -> Any:
         # Must be a `torch.Tensor`, not a list.
-        return torch.rand(1).tolist()
+        return [torch.rand(1).tolist()]
```
toshihikoyanase, line 84: In `checks` CI job, `torch` is not installed, and `mypy` cannot infer the type of this return value.
IMO, we can simply return a list of immediate values here. Or, do we have reason to stick `torch.rand`?

contramundum53: Good point, I think `[0.0]` would suffice.

HideakiImamura: Could you fix the line? Or, is it out of scope of this PR?

contramundum53: There are more than one places where `torch.rand` is used, and we might as well replace them all in another PR.

nzw0301: I think we can replace the line with `[0.0]`. As a result, we can create an issue with `contribution welcome and` good-first-issue`.

contramundum53: @nzw0301 Do you mean that I should fix this in this PR?

nzw0301: Sorry, I meant that this PR can fix it.

contramundum53: Fixed.

tests/integration_tests/test_botorch.py
```diff
@@ -78,9 +79,9 @@ def candidates_func(
         train_obj: torch.Tensor,
         train_con: Optional[torch.Tensor],
         bounds: torch.Tensor,
-    ) -> torch.Tensor:
+    ) -> Any:
```
nzw0301, line 82: `List[float]` sounds more accurate than `Any`. What do you think?
```suggestion
    ) -> List[float]:
```

tests/integration_tests/test_botorch.py
nzw0301: thanks!

tests/integration_tests/test_botorch.py
contramundum53: This is a test case for invalid type, and making it `List[float]` would have to introduce a `# type: ignore`.

nzw0301: Right. Somehow my local mypy was happy with `List[float]`. Thank you for your explanation!

nzw0301 (approved):
LGTM. Thanks!

contramundum53:
@HideakiImamura -san, let us unassign you because you look busy. 
@keisuke-umezawa -san, could you review this PR?

tests/integration_tests/test_botorch.py
keisuke-umezawa: Could you add a comment to explain why we put `Any` here? Apart from it, LGTM.

contramundum53: I added a comment. PTAL.

tests/integration_tests/test_botorch.py
```diff
@@ -78,9 +79,9 @@ def candidates_func(
         train_obj: torch.Tensor,
         train_con: Optional[torch.Tensor],
         bounds: torch.Tensor,
-    ) -> torch.Tensor:
+    ) -> Any:  # The return type is Any because we don't want mypy to check the type.
```
nzw0301, line 82: So this fix is a kinda hotfix of mypy bug according to the issue description, right? Should we add a TODO comment?

keisuke-umezawa (approved):
LGTM! But could you check a comment from @nzw0301 ?

tests/integration_tests/test_botorch.py
contramundum53: I'm not sure which bug of `mypy` we should watch for.

nzw0301: hmm, how about [this](https://github.com/optuna/optuna/pull/3627#discussion_r8895076520)? I suppose the situation is similar to this case.

PR #3683, opened 2022-06-16
Add tests for `_constrained_dominates`

contramundum53:
## Motivation
Currently no test is added for `_dominates` and `_constrained_dominates`. We add tests for those functions.

## Description of the changes
Add tests for `_dominates` and `_constrained_dominates`.

For `_dominates`, we test all pairs of trials with values of combinations of `-inf`, `-1`, `1`, `inf`, and all combinations of directions.

For `_constrained_dominates`, we split the test cases into three cases:
1. Feasible vs. feasible: In this case, we test that `_constrained_dominates` always return the same results as `_dominates`, regardless of the values of constraints. We use the same test cases as `_dominates`, with constraint values of either `-inf`, `-1` or `0`.
2. Infeasible vs. feasible: In this case, we test that feasible trials always dominate the infeasible ones. We use the test trials where each component of constraint values is from the list `[-inf, -1, 0, nan, 2, inf]`.
3. Infeasible vs. Infeasible: In this case, we test integral cases where the violation value is 0, 1, 2, 3 and `inf`.

Note that a `nan` value is treated as infeasible but currently the violation of that value is counted as 0. Possibly we should fix it in another PR.

PR #3686, opened 2022-06-17
Add tests for `_fast_non_dominated_sort`

contramundum53:
## Motivation
Currently, the tests for `_fast_non_dominated_sort` misses some important edge cases. The missed edge cases include
* `inf` and `-inf` in values/constraints
* `nan` in constraints
* duplicated values/constraints
* apparently different constraint values with same meaning (e.g. both are feasible)
etc.
We refactor the tests and cover these cases.

Random tests, suggested in #3678, is not included in this PR.

This is a follow up PR after #3683.

## Description of the changes
* Add test cases.
* Add a function to automatically judge whether the output is correct, i.e., all trials in the same rank do not dominate each other and all trials in rank `>0` are dominated by some trial in the rank above.


contramundum53:

tests/samplers_tests/test_nsgaii.py
```diff
@@ -304,266 +306,115 @@ def test_constrained_dominates_infeasible_vs_infeasible(direction: StudyDirectio
...
-        {0, 1},
-        {3, 4, 5},
-        {2},
-    ]
+    values = itertools.product(
+        [10, 20, 20, 30, float("inf"), float("inf"), -float("inf")], repeat=values_dim
+    )
+    for directions in itertools.product(
+        [StudyDirection.MINIMIZE, StudyDirection.MAXIMIZE], repeat=values_dim
+    ):
```
HideakiImamura, line 345: Can we use `pytest.mark.parametrize` for this?

tests/samplers_tests/test_nsgaii.py
```diff
@@ -304,266 +306,115 @@ def test_constrained_dominates_infeasible_vs_infeasible(direction: StudyDirectio
...
-        _create_frozen_trial(4, [15, 15], [4, 4]),
-    ]
-    population_per_rank = sampler._fast_non_dominated_sort(trials, directions)
-    assert [{t.number for t in population} for population in population_per_rank] == [
-        {0},
-        {1},
-        {2, 3},
-        {4},
-    ]
+@pytest.mark.parametrize("values_dim", [1, 2, 3])
```
HideakiImamura, line 336: How about splitting each dimension test into separate functions as we done for `_constrained_dominates`?

HideakiImamura (commented):
Thanks for the update. I have comments. PTAL.

tests/samplers_tests/test_nsgaii.py
contramundum53: Unlike `_constrained_dominates`, this does not make the test easier, since we need to get the whole list anyways.

contramundum53:
Sorry for late response. I foxed the test.

contramundum53:
@HideakiImamura @c-bata PTAL.

HideakiImamura:
Could you merge the master to fix the CI error?

contramundum53:
@HideakiImamura Fixed. PTAL.

contramundum53:
I simplified `test_fast_non_dominated_sort_with_constraints`.

tests/samplers_tests/test_nsgaii.py
```diff
@@ -336,266 +338,111 @@ def test_constrained_dominates_infeasible_vs_infeasible(direction: StudyDirectio
...
-        _create_frozen_trial(5, [0, 9, 9], [0]),
-    ]
-    population_per_rank = sampler._fast_non_dominated_sort(trials, directions)
-    assert [{t.number for t in population} for population in population_per_rank] == [
-        {2},
-        {0, 3, 5},
-        {1},
-        {4},
-    ]
+                    assert not _constrained_dominates(trial1, trial2, direction)
```
c-bata, line 357: I'm wondering if these lines can check any behavior of `_non_dominated_sort` since `trial1` and `trial2` are completely the same `FrozenTrial` objects. What do you think?

tests/samplers_tests/test_nsgaii.py
```diff
@@ -336,266 +338,111 @@ def test_constrained_dominates_infeasible_vs_infeasible(direction: StudyDirectio
...
 
+        # Check that each trial is dominated by some trial in the rank above.
+        for i in range(len(population_per_rank) - 1):
+            for trial2 in population_per_rank[i + 1]:
+                assert any(
+                    _constrained_dominates(trial1, trial2, direction)
+                    for trial1 in population_per_rank[i]
+                )
```
c-bata, line 365: Do you think we can write these logic like the following since we've already checked that trials in the same rank do not dominate each other?

```suggestion
        for i in range(len(population_per_rank) - 1):
            trial1 = population_per_rank[i][0]
            for trial2 in population_per_rank[i + 1]:
                assert _constrained_dominates(trial1, trial2, direction)
```

c-bata:
Thank you for your pull request. Changes basically looks good except for my previous comments.

tests/samplers_tests/test_nsgaii.py
contramundum53: No, you cannot write it this way because a trial in a lower rank may not dominate a particular trial in higher rank. 
An example is
* Rank 0: `(0, 5)`, `(5, 0)`
* Rank 1: `(7, 1)`
Here, `(7, 1)` is of rank 1 because `(7, 1)` is dominated by `(5, 0)`, but it is not dominated by `(0, 5)`.

c-bata: I see. Thank you for your explanation :bow:

c-bata (approved):
LGTM after passed CI.

HideakiImamura (approved):
LGTM.

PR #3689, opened 2022-06-17
Refactor tests of constrained TPE

knshnb:
## Motivation

Follow-up task of #3506 (review comment: https://github.com/optuna/optuna/pull/3506/files#r886353871). The logic of the original tests was made complex by parametrizing `constraints_enabled`.

## Description of the changes

~I split the tests for constrained TPE into different functions.~
According to the private discussion with @HideakiImamura, we decided to keep the parametrization in one function. Instead of branching inside the functions, they are defined by parametrizing tuple.

### DIFF
--- a/tests/samplers_tests/tpe_tests/test_sampler.py
+++ b/tests/samplers_tests/tpe_tests/test_sampler.py
@@ -4,6 +4,7 @@ from typing import Callable
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import Sequence
 from typing import Union
 from unittest.mock import Mock
 from unittest.mock import patch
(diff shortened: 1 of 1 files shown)


tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -767,115 +767,161 @@ def test_constrained_sample_independent_zero_startup() -> None:
     sampler.sample_independent(study, trial, "param-a", dist)
 
 
-@pytest.mark.parametrize("direction", ["minimize", "maximize"])
-@pytest.mark.parametrize("constraints_enabled", [False, True])
-def test_get_observation_pairs(direction: str, constraints_enabled: bool) -> None:
-    def objective(trial: Trial) -> float:
+def get_observation_pairs_objective(trial: Trial) -> float:
```
HideakiImamura, line 770: How about moving this function inside `test_get_obdervation_pairs`? And, I'm not sure, but we should define other function for `test_get_observation_pairs_constrained` to set the constraint values.

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -767,113 +768,130 @@ def test_constrained_sample_independent_zero_startup() -> None:
     sampler.sample_independent(study, trial, "param-a", dist)
 
 
-@pytest.mark.parametrize("direction", ["minimize", "maximize"])
-@pytest.mark.parametrize("constraints_enabled", [False, True])
-def test_get_observation_pairs(direction: str, constraints_enabled: bool) -> None:
+@pytest.mark.parametrize("direction, sign", [("minimize", 1), ("maximize", -1)])
```
HideakiImamura, line 771: Do we need the parametrization for `sign`? I think it is directly calculated in the test.

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -767,113 +768,130 @@ def test_constrained_sample_independent_zero_startup() -> None:
...
         scores,
-        violations,
+        expected_violations,
     )
 
 
-@pytest.mark.parametrize("direction", ["minimize", "maximize"])
-@pytest.mark.parametrize("constraints_enabled", [False, True])
-def test_get_observation_pairs_multi(direction: str, constraints_enabled: bool) -> None:
+@pytest.mark.parametrize("direction, sign", [("minimize", 1), ("maximize", -1)])
```
HideakiImamura, line 858: Ditto.

HideakiImamura (commented):
Thanks for the PR. I have minor comments. PTAL.

tests/samplers_tests/tpe_tests/test_sampler.py
knshnb: In the original code, there was a conditional branch by `direction` inside the test function (https://github.com/optuna/optuna/pull/3689/files#diff-b71dcddb8db142eb739b8f72a42c3d5784df2e8a7ca6608cb31adebe14ee62f0L799).
I thought eliminating it is better because conditional branch by parametrized values in tests reduces readability and is not a good practice in general.

HideakiImamura: Thanks. Could you remove it in this PR?

HideakiImamura: Oh, I see. IMO, since this is a one line conditional branch, we can accept it. If anything, I think the previous version is simpler and sufficiently readable. @toshihikoyanase What do you think?

toshihikoyanase: > IMO, since this is a one line conditional branch, we can accept it.

I agree with @HideakiImamura . In general, we try to remove conditional branches from test cases, but the following line is simple enough, and managing the pairs of direction and sign by hand might increase the maintenance cost.

```python
sign = 1 if direction == "minimize" else -1
```

knshnb: Then, I followed your comments! PTAL.

HideakiImamura (approved):
Thanks for the fix. LGTM.

toshihikoyanase (approved):
Looks solid! Thank you for your contribution.

This is out-of-scope of this PR, but I found that the current test case did not check if elements of violations were summed up or not. I think we can check it if we add a violation vector like (1, 2), i.e,, both of the elements are positive.

PR #3690, opened 2022-06-17
Add inf & nan tests for test_constraints_func

contramundum53:
## Motivation
`test_constraints_func` in `test_nsgaii.py` lacks test cases for `inf` or `nan` values. We add those test cases.

## Description of the changes
Add `inf` and `nan` cases for `test_constraints_func`.

### DIFF
--- a/tests/samplers_tests/test_nsgaii.py
+++ b/tests/samplers_tests/test_nsgaii.py
@@ -37,7 +37,7 @@ from optuna.trial import FrozenTrial
 def _nan_equal(a: Any, b: Any) -> bool:
     if isinstance(a, float) and isinstance(b, float) and np.isnan(a) and np.isnan(b):
         return True
-    
+
     return a == b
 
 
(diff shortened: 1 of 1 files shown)


contramundum53:
@not522 @knshnb Could you review this PR?

not522 (approved):
LGTM!

tests/samplers_tests/test_nsgaii.py
```diff
@@ -129,7 +138,7 @@ def constraints_func(trial: FrozenTrial) -> Sequence[float]:
         nonlocal constraints_func_call_count
         constraints_func_call_count += 1
 
-        return (trial.number,)
+        return (constraint_value,)
```
knshnb, line 141: In the current code, the case where `constraint_value` is different between trials is not tested, which was originally tested.
How about returning something like `(constraint_value + trial.number,)`?

knshnb (commented):
Thanks for enhancing the test cases. I left a small comment. PTAL.

tests/samplers_tests/test_nsgaii.py
contramundum53: Thanks for pointing that out! I fixed the code. PTAL.

knshnb (approved):
Thanks! LGTM.

PR #3693, opened 2022-06-18
Control metric decimal digits precision in `bayesmark` benchmark report

xadrianzetx:
## Motivation

This allows each metric to be formatted individually, depending on required precision. Defaults are set to mirror kurobako settings - 6 digits for `best value` metric, 3 for everything else. Follow up to #3584 (one of outstanding tasks mentioned in https://github.com/optuna/optuna/pull/3584#issuecomment-1149669967).

## Description of the changes

* Introduce `precision` property in `BaseMetric` class
* Set precision in all metrics to mirror kurobako


contramundum53:

contramundum53 (approved):
LGTM!

PR #3695, opened 2022-06-20
Fix calling storage API in study tests

wattlebirdaz:
## Motivation
`test_study.py` and `test_study_summary.py` should call `study.get_all_study_summaries` instead of `storage.get_all_study_summaries`

## Description of the changes
changed from `storage.get_all_study_summaries` to `get_all_study_summaries` which is an alias of `study.get_all_study_summaries` in `test_study.py` and `test_study_summary.py`


contramundum53:
@c-bata @toshihikoyanase Could you review this PR?

c-bata (approved):
Good catch 👍  LGTM.

toshihikoyanase (approved):
Good catch! Thank you.

PR #3696, opened 2022-06-20
DRY `test_frozen.py`

contramundum53:
## Motivation
DRY `test_frozen.py`.

## Description of the changes
Extract common test logic from `test_frozen.py` and `test_fixed.py`.


contramundum53:

tests/trial_tests/test_fixed_frozen.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+@parametrize_trial_type
+def test_datetime_start(trial_type: type) -> None:
+
+    trial = _create_trial(trial_type)
+    assert trial.datetime_start is not None
+    old_date_time_start = trial.datetime_start
+    trial.datetime_complete = datetime.datetime.now()  # type: ignore
+    assert trial.datetime_complete != old_date_time_start  # type: ignore
```
HideakiImamura, line 253: I think we can remove `# type: ignore` by asserting them not `None`.

tests/trial_tests/test_fixed_frozen.py
```diff
@@ -0,0 +1,253 @@
...
+    elif trial_type == FrozenTrial:
+        trial = create_trial(value=0.2, params=params, distributions=distributions)
+        trial.number = 0
+        return trial
+    else:
+        assert False
+
+
+@parametrize_trial_type
+def test_suggest_float(trial_type: type) -> None:
```
HideakiImamura, line 36: I think we can dry `test_suggest_*` by parametrizing the methods and its arguments.

HideakiImamura (commented):

tests/trial_tests/test_frozen.py
```diff
@@ -21,9 +19,17 @@
 from optuna.trial import TrialState
 
 
+def _create_default_trial() -> FrozenTrial:
```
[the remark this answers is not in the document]
contramundum53: I thought `_create_trial` is confusing because we use `create_trial` from `optuna.trial`.

tests/trial_tests/test_fixed_frozen.py
contramundum53: Hmm... Considering the amount of difference in each test function and considering that these tests test deprecated functions, I think it may be a little over-engineering to parametrize `test_suggest_*`.  I think we could reconsider it when `suggest_uniform` etc. are removed.

tests/trial_tests/test_frozen.py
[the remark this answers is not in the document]
HideakiImamura: > _create_trial is confusing

I don't think so. It is consistent with the `test_fixed_frozen.py`.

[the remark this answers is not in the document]
contramundum53: In `test_fixed_frozen.py`, `_create_trial` also takes arguments of `params` and `distributions`. We could do the same, i.e. add `params` and `distributions` arguments (with default values) to `_create_default_trial` here and rename it to `_create_trial`, and after that replace all direct usages of `optuna.trial.create_trial` by `_create_trial`.

tests/trial_tests/test_fixed_frozen.py
HideakiImamura: Hmm.. I see. Can we fix the type annotation as `Union[FixedTrial, FrozenTrial]` to remove the `# type: ignore`?

contramundum53: No, it didn't work. I think mypy is not smart enough to infer that.

tests/trial_tests/test_fixed_frozen.py
contramundum53: There is a subtle difference between `suggest_float` and `suggest_int` about the accepted value of `(step, log)` that prevents it from being DRYed.

tests/trial_tests/test_frozen.py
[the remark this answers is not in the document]
contramundum53: I fixed it. PTAL.

tests/trial_tests/test_fixed_frozen.py
HideakiImamura: It seems that the `FixedTrial` does not have `datetime_complete`. How about just asserting as follows?
```suggestion
    assert datetime.datetime.now() != old_date_time_start
```

tests/trial_tests/test_fixed_frozen.py
HideakiImamura: Sounds great. It would be fixed after removing `suggest_uniform` and etc.

tests/trial_tests/test_fixed_frozen.py
contramundum53: Thanks, I'll apply your suggestion.

HideakiImamura:
Looks very good except the file name. I would like to other opinions about that.

tests/trial_tests/test_fixed_frozen.py
```diff
@@ -0,0 +1,213 @@
...
+        (lambda self, *args: self.suggest_int(*args, step=2), IntDistribution(1, 10, step=2)),
+        (lambda self, *args: self.suggest_float(*args), FloatDistribution(1, 10)),
+        (
+            lambda self, *args: self.suggest_float(*args, log=True),
+            FloatDistribution(1, 10, log=True),
+        ),
+        (lambda self, *args: self.suggest_float(*args, step=1), FloatDistribution(1, 10, step=1)),
+    ],
+)
+def test_not_contained_param(
```
not522, line 163: We can test `Trial` here. It should be the same behavior as `FrozenTrial` and `FixedTrial`.

tests/trial_tests/test_fixed_frozen.py
```diff
@@ -0,0 +1,213 @@
...
+        trial_type=trial_type,
+        params={"x": 1.0},
+        distributions={"x": distribution},
+    )
+    with pytest.warns(UserWarning):
+        assert suggest_func(trial, "x", 10, 100) == 1
+
+
+@parametrize_trial_type
+def test_set_user_attrs(trial_type: type) -> None:
```
not522, line 176: ditto.

tests/trial_tests/test_fixed_frozen.py
```diff
@@ -0,0 +1,213 @@
...
+
+    trial = _create_trial(trial_type)
+    trial.set_user_attr("data", "MNIST")
+    assert trial.user_attrs["data"] == "MNIST"
+
+
+@parametrize_trial_type
+def test_set_system_attrs(trial_type: type) -> None:
```
not522, line 184: ditto.

tests/trial_tests/test_fixed_frozen.py
```diff
@@ -0,0 +1,253 @@
+import datetime
```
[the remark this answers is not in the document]
not522: I think we can test some features of `Trial` with this file. I recommend adding tests and renaming it to `test_trials.py`.

contramundum53:
@not522 I applied your comments. PTAL.

tests/trial_tests/test_trials.py
```diff
@@ -0,0 +1,233 @@
...
+from optuna.distributions import IntDistribution
+from optuna.study.study import create_study
+from optuna.trial import BaseTrial
+from optuna.trial import create_trial
+from optuna.trial import FixedTrial
+from optuna.trial import FrozenTrial
+from optuna.trial import Trial
+
+
+parametrize_trial_type = pytest.mark.parametrize("trial_type", [FixedTrial, FrozenTrial])
```
HideakiImamura, line 20: I feel uncomfortable that Trial is not included here, why not use parametrize_trial_type in tests where Trial cannot be included (e.g. test_suggest_float)?

tests/trial_tests/test_trials.py
```diff
@@ -0,0 +1,233 @@
...
+
+@pytest.mark.parametrize("trial_type", [FixedTrial, FrozenTrial, Trial])
+def test_set_system_attrs(trial_type: type) -> None:
+
+    trial = _create_trial(trial_type)
+    trial.set_system_attr("system_message", "test")
+    assert trial.system_attrs["system_message"] == "test"
+
+
+@parametrize_trial_type
```
HideakiImamura, line 211: `Trial` can be included.

tests/trial_tests/test_trials.py
```diff
@@ -0,0 +1,233 @@
...
+def test_should_prune(trial_type: type) -> None:
+
+    # Never prunes trials.
+    assert not _create_trial(trial_type).should_prune()
+
+
+@parametrize_trial_type
```
HideakiImamura, line 227: Ditto.

HideakiImamura (commented):
Thanks for the update. I have minor comments. PTAL.

tests/trial_tests/test_trials.py
contramundum53: Sorry, could you elaborate on that?

contramundum53:
@HideakiImamura Thanks, I applied your comments.

HideakiImamura (approved):
Thanks for the update. LGTM!

not522 (approved):
LGTM

PR #3697, opened 2022-06-20
Remove `storage.get_all_study_summaries(include_best_trial: bool)`

wattlebirdaz:
## Motivation
`StudySummary` is an entity that contains `best_trial`, `n_trials`, and `datetime_start` in addition to the basic attributes of the `Study` entity. Since these information requires aggregation of trials in the database, they make the implementation of storage API such as `storage.get_all_study_summaries` more complex and often slower. We try to solve this by introducing another entity called `FrozenStudy` which only contains the static information of `Study` and replacing `storage.get_all_study_summaries` with a simpler `storage.get_all_studies` that returns a list of `FrozenStudy`. 

## Description of the changes
- Introduced `FrozenStudy` entity.
  - This entity is invisible to the user.
  - There is room for discussion on the name of this entity.
- The internal behavior of `study.get_all_study_summaries` changes from calling `storage.get_all_study_summaries` to `storage.get_all_studies` and `stoage.get_all_trials`. There are no change in the user API itself.
- Removed `storage.get_all_study_summaries` from the storage interface and the implementations.


contramundum53:
@c-bata @toshihikoyanase  Could you review this PR?

c-bata (commented):
At a glance, changes looks good to me. However, we have to be careful about introducing FrozenStudy not to become a technological debt in the future. Please let us discuss the naming and the future direction at Optuna core-devs mtg.

optuna/storages/_redis.py
```diff
@@ -140,7 +141,7 @@ def create_new_study(self, study_name: Optional[str] = None) -> int:
                 "study_id:{:010d}:directions".format(study_id),
                 pickle.dumps([StudyDirection.NOT_SET]),
             )
-
+            # TODO(wattlebirdaz): Replace StudySummary with FrozenStudy
```
toshihikoyanase, line 144: A nit.
```suggestion
            # TODO(wattlebirdaz): Replace StudySummary with FrozenStudy.
```

I agree with you. If we simply replace `StudySummary` with `FrozenStudy`, existing users cannot read the studies from RedisStorage. So, we may need to provide migration code, which will make the PR complicated.

toshihikoyanase (commented):
Thank you for your PR. I talked with some committers including @c-bata and decided to continue the review. We couldn't determine the name of `FrozenStudy`, and we should keep it private so that we can rename it later.

Let me share my early comments.

optuna/study/_frozen.py
```diff
@@ -0,0 +1,92 @@
...
+        directions:
+            A sequence of :class:`~optuna.study.StudyDirection` objects.
+        user_attrs:
+            Dictionary that contains the attributes of the :class:`~optuna.study.Study` set with
+            :func:`optuna.study.Study.set_user_attr`.
+        system_attrs:
+            Dictionary that contains the attributes of the :class:`~optuna.study.Study` internally
+            set by Optuna.
+
+    """
```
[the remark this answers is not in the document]
wattlebirdaz: I agree that `FrozenStudy` sounds like it will contain a list of `FrozenTrial`. IMO, `StudyConfig` sounds like an entity user directly modify to change the configuration of `Study`, which is not true in our scenario. (Users change the config by calling some method in `Study` and `StudyConfig` is indirectly modified afterwards.) I think other than `StudySetting`, `StudyInfo`, `StudyEntry`, `StudyContext` also look like an option.

optuna/study/_frozen.py
```diff
@@ -0,0 +1,92 @@
...
+                This attribute is only available during single-objective optimization.
+        directions:
+            A sequence of :class:`~optuna.study.StudyDirection` objects.
+        user_attrs:
+            Dictionary that contains the attributes of the :class:`~optuna.study.Study` set with
+            :func:`optuna.study.Study.set_user_attr`.
+        system_attrs:
+            Dictionary that contains the attributes of the :class:`~optuna.study.Study` internally
+            set by Optuna.
+
```
c-bata, line 32: How about leaving a note to let other Optuna developers know that this class is private and not referenced by Optuna users?

optuna/study/study.py
```diff
@@ -1411,4 +1411,47 @@ def objective(trial):
...
+        n_trials = len(all_trials)
+
+        if len(s.directions) == 1:
+            direction = s.direction
+            directions = None
+            if include_best_trial and len(completed_trials) != 0:
+                if direction == StudyDirection.MAXIMIZE:
+                    best_trial = max(completed_trials, key=lambda t: cast(float, t.value))
+                else:
+                    best_trial = min(completed_trials, key=lambda t: cast(float, t.value))
```
c-bata, line 1431: Can we use `storage.get_best_trial()` here?

optuna/study/_frozen.py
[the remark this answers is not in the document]
c-bata: Not a strong opinion, but I'd say it is also important to give it a similar name since the concept is similar to FrozenTrial.

optuna/study/study.py
wattlebirdaz: `storage.get_all_trials()` is called in line L.1414, which loads all information of trials into memory. That said, current implementation is more performant than calling `storage.get_best_trial()` because it does not have to reach the storage again.

c-bata: Sounds reasonable 👍  Thanks for the clarification.

c-bata (approved):
Thanks for the update! Though changes looks good to me, let me continue to assign myself for the discussion of the naming.

optuna/study/_frozen.py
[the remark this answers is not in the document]
toshihikoyanase: Hmm, I don't think the current `FrozenStudy` is similar to `FrozenTrial` because `FrozenTrial` can be a snapshot of corresponding object (i.e., `Trial`) but `FrozenStudy` cannot. This is mainly because `Study` has the `trials` attribute and `FrozenStudy` doesn't.

We might need to reorganize `Study` since it contains two different information:
- Experimental settings (i.e., information in `FrozenStudy`)
- Experimental results (i.e., a list of `FrozenTrial`s)
I guess it is out-of-scope of this PR, and I'd like to discuss it in the next major release (v4?).

optuna/study/_frozen.py
```diff
@@ -0,0 +1,94 @@
...
+from optuna import logging
+from optuna.study._study_direction import StudyDirection
+
+
+_logger = logging.get_logger(__name__)
+
+
+class FrozenStudy:
+    """Basic attributes of a :class:`~optuna.study.Study`.
```
toshihikoyanase, line 14: Do you think if it is possible to clarify what kind of attributes of `Study` are basic?
For example, `FrozenTrial` shows that it contains status and results of a `Trial`.
https://optuna.readthedocs.io/en/stable/reference/generated/optuna.trial.FrozenTrial.html

optuna/study/_frozen.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if len(self._directions) > 1:
+            raise RuntimeError(
+                "This attribute is not available during multi-objective optimization."
+            )
+
+        return self._directions[0]
+
+    @property
+    def directions(self) -> Sequence[StudyDirection]:
```
toshihikoyanase, line 92: This method actually returns `list` and the concrete type hints for return values will help users.
https://docs.python.org/3/library/typing.html#typing.List
> Generic version of [list](https://docs.python.org/3/library/stdtypes.html#list). Useful for annotating return types. To annotate arguments it is preferred to use an abstract collection type such as [Sequence](https://docs.python.org/3/library/typing.html#typing.Sequence) or [Iterable](https://docs.python.org/3/library/typing.html#typing.Iterable).

Also, the return type of `Study.directions` is `List[StudyDirection]`.
https://github.com/optuna/optuna/blob/f0c4f380d9b7c3a86a1b15b952ee6e09b1451904/optuna/study/study.py#L192

```suggestion
    def directions(self) -> List[StudyDirection]:
```

tests/storages_tests/test_storages.py
```diff
@@ -862,32 +860,23 @@ def check_set_and_get(trial_id: int, key: str, value: Any) -> None:
...
-            assert summary.user_attrs == expected_summary.user_attrs
-            assert summary.system_attrs == expected_summary.system_attrs
-            if include_best_trial:
-                if expected_summary.best_trial is not None:
-                    assert summary.best_trial is not None
-                    assert summary.best_trial == expected_summary.best_trial
-            else:
-                assert summary.best_trial is None
+            assert frozen_study is not None
+            assert frozen_study.direction == expected_frozen_study.direction
```
toshihikoyanase, line 876: [Notes] This is out-of-scope of this PR, but this test case does not cover `FrozenStudy.directions`.

toshihikoyanase (commented):
The change basically looks great to me, but I added some minor comments.

optuna/study/_frozen.py
toshihikoyanase: This comment comes from `StudySummary`, and we'll keep this comment as it.
https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.StudySummary.html#optuna.study.StudySummary

optuna/study/_frozen.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def __init__(
+        self,
+        study_name: str,
+        direction: Optional[StudyDirection],
+        user_attrs: Dict[str, Any],
+        system_attrs: Dict[str, Any],
+        study_id: int,
+        *,
+        directions: Optional[List[StudyDirection]] = None,
```
toshihikoyanase, line 45: Sorry for the nit-picky comment, but we prefer `Sequence` when we specify the type of arguments. (c.f., https://github.com/optuna/optuna/blob/f0c4f380d9b7c3a86a1b15b952ee6e09b1451904/optuna/study/study.py#L1033)

I guess the code actually accept a sequence of `StudyDirection` since we apply `list` function to `directions` at L51.

```suggestion
        directions: Optional[Sequence[StudyDirection]] = None,
```

toshihikoyanase (approved):

PR #3698, opened 2022-06-20
Unify the logic of edf functions

nzw0301:
## Motivation

Since `optuan.visualization.plot_edf` and `optuan.visualization.matplotlib.plot_edf` have almost the same logic to generate figures, this PR aims to reduce the number of lines of duplicated lines in edf-related files for consistent tests, which are follow-up.

## Description of the changes

By taking the same approach as in https://github.com/optuna/optuna/blob/master/optuna/visualization/_pareto_front.py

- Define `_EDFInfo` thats stores minimal information for the plot not depending on the backend visualisation packages.
- Each concrete plot function calls `_get_edf_info` to gather information for plot.

### DIFF
--- a/optuna/visualization/_edf.py
+++ b/optuna/visualization/_edf.py
@@ -1,6 +1,7 @@
 from typing import Callable
 from typing import cast
 from typing import List
+from typing import NamedTuple
 from typing import Optional
 from typing import Sequence
 from typing import Union
@@ -20,6 +21,12 @@ if _imports.is_successful():
     from optuna.visualization._plotly_imports import go
 
 _logger = get_logger(__name__)
+
+
+class _EDFInfo(NamedTuple):
+    study_names: List[str]
+    x_values: np.ndarray
+    y_values_by_study: List[np.ndarray]
 
 
 def plot_edf(
@@ -102,6 +109,33 @@ def plot_edf(
 
     _imports.check()
 
+    layout = go.Layout(
+        title="Empirical Distribution Function Plot",
+        xaxis={"title": target_name},
+        yaxis={"title": "Cumulative Probability"},
+    )
+
+    info = _get_edf_info(study, target, target_name)
+
+    if len(info.study_names) == 0 or len(info.x_values) == 0:
+        return go.Figure(data=[], layout=layout)
+
+    traces = []
+    for y_values, study_name in zip(info.y_values_by_study, info.study_names):
+        traces.append(go.Scatter(x=info.x_values, y=y_values, name=study_name, mode="lines"))
+
+    figure = go.Figure(data=traces, layout=layout)
+    figure.update_yaxes(range=[0, 1])
+
+    return figure
+
+
+def _get_edf_info(
+    study: Union[Study, Sequence[Study]],
+    target: Optional[Callable[[FrozenTrial], float]] = None,
+    target_name: str = "Objective Value",
+) -> _EDFInfo:
+
     if isinstance(study, Study):
         studies = [study]
     else:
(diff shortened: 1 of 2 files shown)


contramundum53:
@xadrianzetx Could you review this PR?

optuna/visualization/_edf.py
```diff
@@ -102,30 +109,43 @@ def objective(trial, low, high):
...
     _imports.check()
 
-    if isinstance(study, Study):
-        studies = [study]
-    else:
-        studies = list(study)
+    layout = go.Layout(
+        title="Empirical Distribution Function Plot",
+        xaxis={"title": target_name},
+        yaxis={"title": "Cumulative Probability"},
```
nzw0301, line 115: Possibly `_EDFInfo` contains these values too.

optuna/visualization/_edf.py
```diff
@@ -22,6 +23,12 @@
 _logger = get_logger(__name__)
 
 
+class _EDFInfo(NamedTuple):
+    study_names: List[str]
+    x_values: np.ndarray
+    y_values_by_study: List[np.ndarray]
```
HideakiImamura, line 29: It is implicitly assumed that each member of this `_EDFInfo` has a corresponding name of `study_names` and each element of `y_values_by_study`. To make this explicit, how about the following?

```suggestion
class _EDFLineInfo(NamedTuple):
    study_name: str
    y_values_by_study: np.ndarray


class _EDFInfo(NamedTuple):
    lines: List[_EDFLineInfo]
    x_values: np.ndarray
```

contramundum53 (approved):
LGTM!

optuna/visualization/_edf.py
```diff
@@ -102,30 +113,44 @@ def objective(trial, low, high):
...
+        studies = [study]
+    else:
+        studies = list(study)
+
+    _check_plot_args(studies, target, target_name)
 
     if len(studies) == 0:
         _logger.warning("There are no studies.")
-        return go.Figure(data=[], layout=layout)
+        return _EDFInfo(lines=[], x_values=np.array([]))
```
xadrianzetx, line 153: This is a small nit, but maybe `lines` and `x_values` should be `Optional`? This would simplify checking for empty plot in concrete implementations and i guess reduce allocation a bit. What do you think @nzw0301?

optuna/visualization/_edf.py
xadrianzetx: Those are just literals, so imo it's fine to keep them as they are.

optuna/visualization/_edf.py
nzw0301: Thank you for your comment and concrete suggestions! I think the other visualisation parts (ex. #3682) might have a similar case; @HideakiImamura do you have a preference? 

Explicitly returning `lines=None, x_values=None` sounds easy to test the logic when plot data is invalid. One drawback might be to add a few lines to avoid mypy error.

HideakiImamura (approved):
LGTM. I would like to merge this PR after the review by @xadrianzetx.

optuna/visualization/_edf.py
xadrianzetx: Thank you for the opinion, let's keep it as it is then.

xadrianzetx (approved):
LGTM!

optuna/visualization/_edf.py
nzw0301: In fact, I realised that we might be able to simplify the initialisation by using the default values as in https://github.com/optuna/optuna/pull/3781.

PR #3700, opened 2022-06-20
Unify the logic of `plot_param_importances` functions

nzw0301:
## Motivation

Since `optuan.visualization.plot_param_importances` and `optuan.visualization.matplotlib.plot_param_importances` have almost the same logic to generate figures, this PR aims to reduce the number of lines of duplicated lines in `plot_param_importances`-related files for consistent tests, which are follow-up.

## Description of the changes

By taking the same approach as in https://github.com/optuna/optuna/blob/master/optuna/visualization/_pareto_front.py and #3698,

- Define `_ImportancesInfo` that stores minimal information for the plot not depending on the backend visualisation packages.
- Calls `_get_importances_info` to gather information for the plot from each concrete plot function.

### DIFF
--- a/optuna/visualization/_param_importances.py
+++ b/optuna/visualization/_param_importances.py
@@ -26,9 +26,6 @@ logger = get_logger(__name__)
 
 
 class _ImportancesInfo(NamedTuple):
-    title: str
-    x_axis_name: str
-    y_axis_name: str
     importance_values: List[float]
     param_names: List[str]
     importance_labels: List[str]
@@ -43,12 +40,6 @@ def _get_importances_info(
 ) -> _ImportancesInfo:
     _check_plot_args(study, target, target_name)
 
-    title = "Hyperparameter Importances"
-    x_axis_name = f"Importance for {target_name}"
-    y_axis_name = "Hyperparameter"
-
-    # Importances cannot be evaluated without completed trials.
-    # Return an empty figure for consistency with other visualization functions.
     trials = _filter_nonfinite(
         study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,)), target=target
     )
@@ -56,9 +47,6 @@ def _get_importances_info(
     if len(trials) == 0:
         logger.warning("Study instance does not contain completed trials.")
         return _ImportancesInfo(
-            title=title,
-            x_axis_name=x_axis_name,
-            y_axis_name=y_axis_name,
             importance_values=[],
             param_names=[],
             importance_labels=[],
@@ -74,9 +62,6 @@ def _get_importances_info(
     importance_labels = [f"{val:.2f}" if val >= 0.01 else "<0.01" for val in importance_values]
 
     return _ImportancesInfo(
-        title=title,
-        x_axis_name=x_axis_name,
-        y_axis_name=y_axis_name,
         importance_values=importance_values,
         param_names=param_names,
         importance_labels=importance_labels,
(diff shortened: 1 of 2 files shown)


contramundum53:
@xadrianzetx @knshnb Could you review this PR?

optuna/visualization/_param_importances.py
```diff
@@ -24,6 +25,64 @@
 logger = get_logger(__name__)
 
 
+class _ImportancesInfo(NamedTuple):
+    title: str
+    x_axis_name: str
+    y_axis_name: str
```
xadrianzetx, line 31: Similarly to `plot_edf`, I think those are simple enough to be passed directly as argument to function that creates figure in each of backends. No need to store as part of info class.

optuna/visualization/_param_importances.py
```diff
@@ -24,6 +25,64 @@
...
+    target_name: str,
+) -> _ImportancesInfo:
+    _check_plot_args(study, target, target_name)
+
+    title = "Hyperparameter Importances"
+    x_axis_name = f"Importance for {target_name}"
+    y_axis_name = "Hyperparameter"
+
+    # Importances cannot be evaluated without completed trials.
+    # Return an empty figure for consistency with other visualization functions.
```
xadrianzetx, line 51: A nit, but we're no longer dealing with figures after this function call has been moved, so this comment lies a bit. I think it can be safely removed, or rephrased.

optuna/visualization/_param_importances.py
```diff
@@ -24,6 +25,64 @@
...
+
+    if len(trials) == 0:
+        logger.warning("Study instance does not contain completed trials.")
+        return _ImportancesInfo(
+            title=title,
+            x_axis_name=x_axis_name,
+            y_axis_name=y_axis_name,
+            importance_values=[],
+            param_names=[],
+            importance_labels=[],
```
xadrianzetx, line 64: Similarly to EDF plot - maybe consider those attributes as `Optional`?

optuna/visualization/_param_importances.py
```diff
@@ -89,45 +148,36 @@ def objective(trial):
...
+    if len(importance_values) == 0:
+        return go.Figure(data=[], layout=layout)
 
-    importances = OrderedDict(reversed(list(importances.items())))
-    importance_values = list(importances.values())
-    param_names = list(importances.keys())
-    importance_labels = [f"{val:.2f}" if val >= 0.01 else "<0.01" for val in importance_values]
+    hovertemplate = [
+        _make_hovertext(param_name, importance, study)
+        for (param_name, importance) in zip(param_names, importance_values)
```
xadrianzetx, line 169:
```suggestion
        for param_name, importance in zip(param_names, importance_values)
```

optuna/visualization/_param_importances.py
nzw0301: Thank you for yoru coments. I agree with you; I will remove these values from the named tuple.

optuna/visualization/_param_importances.py
nzw0301: Nice catch. Indeed, this warning message explains this part, so let's remove this!

optuna/visualization/_param_importances.py
nzw0301: I suppose so.

xadrianzetx (approved):
Thank you for the updates, LGTM!

knshnb:
Thanks for the PR. The changes look very solid.
Could you kindly wait until #3721, to validate that it works correctly, which seems to be merged in a few days?

nzw0301:
Thank both of you for reviewing this PR! 

@knshnb Sure!

nzw0301:
Let's try visualisation regression test!

knshnb (approved):
It is working correctly! LGTM.
https://output.circle-artifacts.com/output/job/404fffa3-654b-4362-9083-a18a7cbac0d6/artifacts/0/optuna-visualization-regression-tests/public/plot_param_importances.html

PR #3705, opened 2022-06-21
Add test cases for crossovers of NSGAII

toshihikoyanase:
## Motivation

Recently, several PRs such as https://github.com/optuna/optuna/pull/3683 and https://github.com/optuna/optuna/pull/3706 improves the unit tests of `NSGAIISampler`. This PR is trying to improve the `crossover` tests.

### Test scenario

- The crossovers handle `parameters`. They are `float` values but they does not take `inf`, `-inf`, `nan` unlike trial values and trial intermediate values. So, I didn't add such test cases.
- The return values of crossovers should not be `nan` `inf` and the corresponding test cases exist. I'm not fully sure if it is enough to check a random input. I guess it is acceptable if the next item (edge case checking) can detect the errors of algorithm logic.
- The logic of the crossover algorithms are checked in `test_crossover_deterministic`. Previously, it checks only the center point (roughly speaking) of the crossover models and this PR adds edges of the crossover models.

### Discussions

- All points of `test_crossover_deterministic` test case are on a line (`y = x + 1`). They shrink the distributions of `SPX` and `UNDX` and I don't think it is a typical case.
- If we want to check the crossover algorithms in detail, we can create individual test files. On the other hand, we can check the algorithm implementation using benchmark, not unit tests.

## Description of the changes

- Use `np.isnan` and `np.isinf` instead of `np.nan in ...` and `np.inf in ...`
- Check the invalid population size of all crossovers.
- Check the end-point of crossover models in addition to the mean points.

### DIFF
--- a/tests/samplers_tests/test_nsgaii.py
+++ b/tests/samplers_tests/test_nsgaii.py
@@ -652,9 +652,10 @@ def test_crossover_duplicated_param_values(crossover: BaseCrossover) -> None:
         (VSBXCrossover(), 0.0, np.array([2.0, 3.0])),  # c1 = (p1 + p2) / 2.
         (VSBXCrossover(), 0.5, np.array([3.0, 4.0])),  # p2.
         (VSBXCrossover(), 1.0, np.array([3.0, 4.0])),  # p2.
+        # p1, p2 and p3 are on x + 1, and distance from child to PSL is 0.
+        (UNDXCrossover(), -0.5, np.array([3.0, 4.0])),  # [2, 3] + [-1, -1] + [0, 0].
         (UNDXCrossover(), 0.0, np.array([2.0, 3.0])),  # [2, 3] + [0, 0] + [0, 0].
-        (UNDXCrossover(), 0.5, np.array([1.0, 2.0])),  # [2, 3] + [-1, -1] + [1, 1].
-        (UNDXCrossover(), 1.0, np.array([0.0, 1.0])),  # [2, 3] + [-2, -2] + [0, 0].
+        (UNDXCrossover(), 0.5, np.array([1.0, 2.0])),  # [2, 3] + [-1, -1] + [0, 0].
     ],
 )
 def test_crossover_deterministic(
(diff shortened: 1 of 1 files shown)


contramundum53:

tests/samplers_tests/test_nsgaii.py
```diff
@@ -603,17 +605,61 @@ def test_crossover_inlined_categorical_distribution() -> None:
...
+        (SPXCrossover(), 0.5, np.array([2.75735931, 3.75735931])),  # rs = [0.5, 0.25].
+        (SPXCrossover(), 1.0, np.array([-1.0, 0.0])),  # rs = [1, 1], xks[0].
+        (SBXCrossover(), 0.0, np.array([2.0, 3.0])),  # c1 = (p1 + p2) / 2.
+        (SBXCrossover(), 0.5, np.array([3.0, 4.0])),  # p2.
+        (SBXCrossover(), 1.0, np.array([3.0, 4.0])),  # p2.
+        (VSBXCrossover(), 0.0, np.array([2.0, 3.0])),  # c1 = (p1 + p2) / 2.
+        (VSBXCrossover(), 0.5, np.array([3.0, 4.0])),  # p2.
+        (VSBXCrossover(), 1.0, np.array([3.0, 4.0])),  # p2.
+        (UNDXCrossover(), 0.0, np.array([2.0, 3.0])),  # [2, 3] + [0, 0] + [0, 0].
+        (UNDXCrossover(), 0.5, np.array([1.0, 2.0])),  # [2, 3] + [-1, -1] + [1, 1].
```
HideakiImamura, line 656: In this line, the value of `expected_params` and the comment are inconsistent?

HideakiImamura (commented):
Thanks for the PR. I have a minor comment. PTAL.

contramundum53 (approved):
LTGM with imamura-san's comment.

tests/samplers_tests/test_nsgaii.py
toshihikoyanase: You are absolutely right. Thank you for pointing it out. I fixed the comment in https://github.com/optuna/optuna/pull/3705/commits/6ef5216b8c20c8c1e4870d2d1b03b689d2c91623.
In fact, third element is always `0` since all parents are on a line (i.e., `x + 1`).

I also added the negative `rand_value`, so that `UNDXCrossover` could generate values from [2, 3] to [3, 4].

HideakiImamura (approved):
LGTM.

PR #3708, opened 2022-06-21
Use `get_trials` with `states` argument to filter trials depending on trial state

nzw0301:
## Motivation

`study.get_trials` method can fetch the trials depending on their trial state. By using this, we can remove several `if` filtering in optuna codebase.

## Description of the changes

As the title, use `get_trials` with explicit states values rather than get all trials then filter them by their state.

### DIFF
--- a/optuna/integration/cma.py
+++ b/optuna/integration/cma.py
@@ -183,7 +183,7 @@ class PyCmaSampler(BaseSampler):
         self._raise_error_if_multi_objective(study)
 
         if self._warn_independent_sampling:
-            complete_trials = [t for t in study.trials if t.state == TrialState.COMPLETE]
+            complete_trials = study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))
             if len(complete_trials) >= self._n_startup_trials:
                 self._log_independent_sampling(trial, param_name)
 
@@ -210,7 +210,7 @@ class PyCmaSampler(BaseSampler):
             self._warn_independent_sampling = False
             return {}
 
-        complete_trials = [t for t in study.trials if t.state == TrialState.COMPLETE]
+        complete_trials = study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))
         if len(complete_trials) < self._n_startup_trials:
             return {}
 
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -220,7 +220,7 @@ def _get_pareto_front_info(
     if constraints_func is not None:
         feasible_trials = []
         infeasible_trials = []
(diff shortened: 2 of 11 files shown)


contramundum53 (approved):
LGTM!

contramundum53:

knshnb (commented):
Thanks for the PR and it looks great overall!
Let me clarify one point. You converted `deepcopy=False` to `True` in several cases. Did you confirm it is okay? (I briefly checked they are fine in most cases though.)

optuna/integration/cma.py
```diff
@@ -183,7 +183,7 @@ def sample_independent(
         self._raise_error_if_multi_objective(study)
 
         if self._warn_independent_sampling:
-            complete_trials = [t for t in study.trials if t.state == TrialState.COMPLETE]
+            complete_trials = study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))
```
nzw0301, line 186: `complete_trials` is necessary to count the number of complete trials, so we can use `deepcopy=False`.

optuna/integration/cma.py
```diff
@@ -210,7 +210,7 @@ def sample_relative(
             self._warn_independent_sampling = False
             return {}
 
-        complete_trials = [t for t in study.trials if t.state == TrialState.COMPLETE]
+        complete_trials = study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))
```
nzw0301, line 213: ditto

optuna/visualization/_pareto_front.py
```diff
@@ -220,7 +220,7 @@ def _get_pareto_front_info(
     if constraints_func is not None:
         feasible_trials = []
         infeasible_trials = []
-        for trial in study.get_trials(states=(TrialState.COMPLETE,)):
+        for trial in study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,)):
```
nzw0301, line 223: In my understanding, `trial` is not modified in this file, so we an use `deepcopy=False`.

optuna/visualization/_slice.py
```diff
@@ -86,7 +86,7 @@ def _get_slice_plot(
 
     layout = go.Layout(title="Slice Plot")
 
-    trials = [trial for trial in study.trials if trial.state == TrialState.COMPLETE]
+    trials = study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))
```
nzw0301, line 89: ditto

nzw0301:
@knshnb thank you for your review and comments. I believe the introduced changes are supposed to be fine because the trials are not modified in the later part. If you have any concerns, let me know!

knshnb (approved):
Thanks for the detailed explanation! LGTM.

Issue #3681, opened 2022-06-16
matplotlib's plot_contour() method raise an error when given dynamic search space.


c-bata:
### Expected behavior

As plotly's plot_contour() can handle dynamic search spaces properly, matplotlib's plot_contour() method should be able to accept dynamic search space.

### Environment

- Optuna version:3.0.0rc0.dev
- Python version:3.9.11
- OS:macOS-12.3.1-x86_64-i386-64bit
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
$ python bug_matplotlib_plot_contour.py
[I 2022-06-16 11:43:25,236] A new study created in memory with name: A single-objective study with 3-dimensional dynamic search space
[I 2022-06-16 11:43:25,238] Trial 0 finished with value: -0.9723216524479252 and parameters: {'category': 'bar', 'x2': -5.986063716221181}. Best is trial 0 with value: -0.9723216524479252.
[I 2022-06-16 11:43:25,239] Trial 1 finished with value: 1.760435351638048 and parameters: {'category': 'foo', 'x1': 3.326813985318985}. Best is trial 1 with value: 1.760435351638048.
[I 2022-06-16 11:43:25,239] Trial 2 finished with value: -7.583657590003812 and parameters: {'category': 'bar', 'x2': -2.246155852266906}. Best is trial 1 with value: 1.760435351638048.
[I 2022-06-16 11:43:25,239] Trial 3 finished with value: 18.474087129547364 and parameters: {'category': 'foo', 'x1': 6.298149267946306}. Best is trial 3 with value: 18.474087129547364.
[I 2022-06-16 11:43:25,240] Trial 4 finished with value: -9.428683789636048 and parameters: {'category': 'bar', 'x2': -1.9293838094551692}. Best is trial 3 with value: 18.474087129547364.
[I 2022-06-16 11:43:25,240] Trial 5 finished with value: -15.02901036791338 and parameters: {'category': 'bar', 'x2': -8.876726759511609}. Best is trial 3 with value: 18.474087129547364.
[I 2022-06-16 11:43:25,240] Trial 6 finished with value: 35.83677106510998 and parameters: {'category': 'foo', 'x1': 7.9863821349050195}. Best is trial 6 with value: 35.83677106510998.
[...]

PR #3711, opened 2022-06-22
Add `visual_regression_tests` to find regression bugs of visualization module.

c-bata:
## Motivation

It is difficult to ensure the behavior of the visualization module with unit tests, which can easily lead to test case omissions and increased complexity. To solve this problem, I created https://github.com/c-bata/optuna-visualization-regression-tests. This tool prepares various studies and executes each function in Optuna's visualization module. The generated images can be viewed in HTML as shown below.

plot_slice

See https://c-bata.link/optuna-visualization-regression-tests/index.html for details.

This PR makes it easier for reviewers to notice regression bugs by running this tool on every new PR created in Optuna and uploading the results to the CircleCI Artifact.

FYI: This tool actually helped me find the following bug.
https://github.com/optuna/optuna/issues/3681

## Description of the changes

Add `visual_regression_tests` and execute it on CircleCI.

### DIFF
--- a/.github/workflows/doc-link.yml
+++ b/.github/workflows/doc-link.yml
@@ -14,10 +14,10 @@ jobs:
         circleci-jobs: document
   visualization-regression-tests-link:
     runs-on: ubuntu-latest
-    if: "github.event.context == 'ci/circleci: visualization-regression-tests'"
+    if: "github.event.context == 'ci/circleci: visual-regression-tests'"
     steps:
     - uses: larsoner/circleci-artifacts-redirector-action@master
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        artifact-path: 0/optuna-visualization-regression-tests/docs/index.html
-        circleci-jobs: visualization-regression-tests
+        artifact-path: 0/visual_regression_tests/public/index.html
+        circleci-jobs: visual-regression-tests
--- a/.github/workflows/doc-link.yml
+++ b/.github/workflows/doc-link.yml
@@ -12,3 +12,12 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         artifact-path: 0/docs/build/html/index.html
         circleci-jobs: document
+  visualization-regression-tests-link:
+    runs-on: ubuntu-latest
+    if: "github.event.context == 'ci/circleci: visualization-regression-tests'"
+    steps:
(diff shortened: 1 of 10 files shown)


.github/workflows/doc-link.yml
```diff
@@ -12,3 +12,12 @@ jobs:
...
         circleci-jobs: document
+  visualization-regression-tests-link:
+    runs-on: ubuntu-latest
+    if: "github.event.context == 'ci/circleci: visual-regression-tests'"
+    steps:
+    - uses: larsoner/circleci-artifacts-redirector-action@master
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        artifact-path: 0/visual_regression_tests/public/index.html
+        circleci-jobs: visual-regression-tests
```
c-bata, line 23: According to @toshihikoyanase, the link seems not to be listed on the status area of this PR until merged to master branch, since the trigger condition of this workflow is `on: status`.

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.
> https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#status

HideakiImamura:
Also, it may not be necessary to trigger this workflow in every PR. For example, would it be possible to trigger the workflow only when the file of the corresponding visualization function is modified? (This is possible with GitHub Actions, but I'm not sure if it's possible with CircleCI)

c-bata:
Thanks for your review.

> Wouldn't it be difficult to have to submit a PR to Optuna every time we want to add a test case study?

Correct. We need to submit a PR every time we want to add a new test case for visual regression tests.

> For example, would it be possible to leave the study definition file and the code that outputs the images here and just call it in Optuna to output html and attach the link?

Sorry, I don't see your point. Do you mean reviewers can manually run this tool instead of using CircleCI?

> Also, it may not be necessary to trigger this workflow in every PR. For example, would it be possible to trigger the workflow only when the file of the corresponding visualization function is modified? (This is possible with GitHub Actions, but I'm not sure if it's possible with CircleCI)

According to https://circleci.com/developer/orbs/orb/circleci/path-filtering, it seems to be possible. I'll try it 👍

c-bata:
We decided to transfer https://github.com/c-bata/optuna-visualization-regression-tests to Optuna organization. So we can clone it and run on CircleCI workflow instead. It makes easier to update the list of studies.

PR #3714, opened 2022-06-23
Return Python's native float values.

toshihikoyanase:
## Motivation

Fix the mypy error of the daily CI https://github.com/optuna/optuna/runs/7014092845?check_suite_focus=true#step:8:10.
@not522 mentioned that the error might be related to numpy 1.23.0 release.
https://pypi.org/project/numpy/1.23.0/

## Description of the changes

- Convert `numpy.floating` to Python's native `float`.

```python
>>> import numpy as np
>>> y = np.var(1)
>>> y
0.0
>>> type(y)
<class 'numpy.float64'>
>>> type(y.item())
<class 'float'>
```


HideakiImamura (approved):
LGTM.

PR #3725, opened 2022-06-26
Simplify `bayesmark` benchmark report rendering

xadrianzetx:
## Motivation

Previously, `BayesmarkReportBuilder` class was responsible for handling report data as well as rendering the benchmark report. This was accomplished somewhat awkwardly with multiple string buffers, which made it hard to read and difficult to extend as any additions to report structure (such as ones mentioned in https://github.com/optuna/optuna/pull/3584#discussion_r885216386) would significantly bump code complexity. In this PR, we are offloading all rendering related tasks to templating engine, leaving aforementioned class to deal only with data. This simplifies code by a fair margin and makes it a bit more SOLID and easy to extend.

## Description of the changes

* Remove all string buffers from `BayesmarkReportBuilder`
* Introduce simple abstractions over report data
* Offload report rendering to templating engine

### DIFF
--- a/.github/workflows/performance-benchmarks-bayesmark.yml
+++ b/.github/workflows/performance-benchmarks-bayesmark.yml
@@ -105,7 +105,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install --progress-bar off -U setuptools wheel
-        pip install --progress-bar off numpy scipy pandas
+        pip install --progress-bar off numpy scipy pandas Jinja2
 
     - name: Run benchmark report builder
       run: |
--- a/benchmarks/bayesmark/report_bayesmark.py
+++ b/benchmarks/bayesmark/report_bayesmark.py
@@ -189,18 +189,17 @@ class DewanckerRanker:
 class BayesmarkReportBuilder:
     def __init__(self) -> None:
 
-        self._solvers: Set[str] = set()
-        self._datasets: Set[str] = set()
-        self._models: Set[str] = set()
-        self._firsts: Dict[str, int] = defaultdict(int)
-        self._borda: Dict[str, int] = defaultdict(int)
-        self._metric_precedence = str()
-        self._problems_counter = 1
-        self._problems_body = io.StringIO()
+        self.solvers: Set[str] = set()
+        self.datasets: Set[str] = set()
+        self.models: Set[str] = set()
+        self.firsts: Dict[str, int] = defaultdict(int)
+        self.borda: Dict[str, int] = defaultdict(int)
+        self.metric_precedence = str()
+        self.problems: List[Problem] = list()
 
     def set_precedence(self, metrics: List[BaseMetric]) -> None:
 
-        self._metric_precedence = " -> ".join([m.name for m in metrics])
+        self.metric_precedence = " -> ".join([m.name for m in metrics])
 
     def add_problem(
         self,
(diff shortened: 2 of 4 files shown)


HideakiImamura (commented):
Thanks for the PR. The looks great to me. Can I have a example `report.md` generated from this PR?

xadrianzetx:
https://github.com/xadrianzetx/optuna-bayesmark/actions/runs/2563787777

HideakiImamura:
@c-bata Could you review this PR? This PR uses `Jinja2` and dynamically generates the report file using it. It would be great to have a comment from you.

.github/workflows/performance-benchmarks-bayesmark.yml
```diff
@@ -105,7 +105,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install --progress-bar off -U setuptools wheel
-        pip install --progress-bar off numpy scipy pandas
+        pip install --progress-bar off numpy scipy pandas Jinja2
```
c-bata, line 108: Should we update `How to run locally` section as well?
https://github.com/optuna/optuna/tree/master/benchmarks#how-to-run-locally-1

benchmarks/bayesmark/report_bayesmark.py
```diff
@@ -1,6 +1,6 @@
 import abc
 from collections import defaultdict
-import io
+from dataclasses import dataclass
```
c-bata, line 3: I have a question. Optuna still supports Python 3.6. Should we also support Python 3.6 for benchmark scripts? Or has it already been dropped?

benchmarks/bayesmark/report_bayesmark.py
```diff
@@ -186,21 +183,35 @@ def rank(self, report: PartialReport) -> None:
...
-        self._metric_precedence = str()
-        self._problems_counter = 1
-        self._problems_body = io.StringIO()
+        self.solvers: Set[str] = set()
+        self.datasets: Set[str] = set()
+        self.models: Set[str] = set()
+        self.firsts: Dict[str, int] = defaultdict(int)
+        self.borda: Dict[str, int] = defaultdict(int)
+        self.metric_precedence = str()
+        self.problems: List[Problem] = list()
```
c-bata, line 210: nits

```suggestion
        self.metric_precedence = ""
        self.problems: List[Problem] = []
```

benchmarks/bayesmark/report_bayesmark.py
HideakiImamura: I think we don't have to support Python 3.6 in the benchmark. In https://github.com/optuna/optuna/issues/3021, we have decided not to support Python 3.6 in integration modules and examples.

HideakiImamura (approved):
Looks solid! I checked the generated report file is fine.

benchmarks/bayesmark/report_bayesmark.py
c-bata: Thank you for letting me know!

.github/workflows/performance-benchmarks-bayesmark.yml
xadrianzetx: Thanks, updated in 50ca9f23a81078df1004d9ef496e09d9527a324a! On the related note - I've planned a bigger update of bayesmark README section as separate work piece to follow sometime soon since report and scoring logic evolved a bit.

c-bata (approved):
LGTM!

Issue #3586, opened 2022-05-26
Unify the visualization tests of different backends


HideakiImamura:
## Motivation

Currently, visualization tests have two test file for each visualization function, e.g. `tests/visualization_tests/test_contour.py` for plotly and `tests/visualization_tests/matplotlib_tests/test_contour.py` for matplotlib. There is a lot of overlap between those test codes, and they are not well-maintained tests. This issue aims to unify both files to reduce code redundancy and improve maintainability.

## How to do that

First, I will prototype a PR to unify the `test_contour.py`. Let's use that as a springboard for discussion.

HideakiImamura:
Let me close this PR since we have finished the prototyping.

PR #3736, opened 2022-06-29
Unify implementation of `plot_optimization_history` between plotly and matplotlib

knshnb:
## Motivation

Part of #3586.
This PR unifies a part of `plot_optimization_history` that can be commonly used in plotly and matplotlib backends. It also aims at unifying some logic in `_get_optimization_histories` and `_get_optimization_histories_with_error_bar`.

## Description of the changes

- ~Introduce `_OptimizationHistoryInfo` for `_get_optimization_histories` and `_OptimizationHistoryErrorBarInfo` for `_get_optimization_histories_with_error_bar`.~
- ~Reuse `_get_optimization_history_info_list` inside `_get_optimization_history_error_bar_info`~
- Introduced `_OptimizationHistoryInfo`, which is used for both plotting with and without error bars.
- Refactored the logic of calculating mean and std for error bars.

### DIFF
--- a/optuna/visualization/_optimization_history.py
+++ b/optuna/visualization/_optimization_history.py
@@ -89,7 +89,8 @@ def _get_optimization_history_info_list(
     # When error_bar=True, a list of 0 or 1 element is returned.
     if len(info_list) == 0:
         return []
-    max_num_trial = max(max(info.trial_numbers) for info in info_list) + 1
+    all_trial_numbers = [number for info in info_list for number in info.trial_numbers]
+    max_num_trial = max(all_trial_numbers) + 1
 
     def _aggregate(label_name: str, use_best_value: bool) -> Tuple[List[int], _ValuesInfo]:
         # Calculate mean and std of values for each trial number.
--- a/optuna/visualization/_optimization_history.py
+++ b/optuna/visualization/_optimization_history.py
@@ -23,12 +23,16 @@ if _imports.is_successful():
 _logger = get_logger(__name__)
 
 
+class _ValuesInfo(NamedTuple):
+    values: List[float]
+    stds: Optional[List[float]]
+    label_name: str
+
+
 class _OptimizationHistoryInfo(NamedTuple):
     trial_numbers: List[int]
-    values: List[float]
-    label_name: str
-    best_values: Optional[List[float]]
-    best_label_name: Optional[str]
+    values_info: _ValuesInfo
+    best_values_info: Optional[_ValuesInfo]
 
 
 def _get_optimization_history_info_list(
@@ -46,8 +50,6 @@ def _get_optimization_history_info_list(
     for study in studies:
         trials = study.get_trials(states=(TrialState.COMPLETE,))
         label_name = target_name if len(studies) == 1 else f"{target_name} of {study.study_name}"
-        best_values: Optional[List[float]]
-        best_label_name: Optional[str]
         if target is None:
             values = [cast(float, t.value) for t in trials]
             if study.direction == StudyDirection.MINIMIZE:
@@ -57,17 +59,15 @@ def _get_optimization_history_info_list(
             best_label_name = (
                 "Best Value" if len(studies) == 1 else f"Best Value of {study.study_name}"
             )
+            best_values_info = _ValuesInfo(best_values, None, best_label_name)
         else:
             values = [target(t) for t in trials]
-            best_values = None
-            best_label_name = None
+            best_values_info = None
         optimization_history_info_list.append(
             _OptimizationHistoryInfo(
                 trial_numbers=[t.number for t in trials],
-                values=values,
-                label_name=label_name,
-                best_values=best_values,
-                best_label_name=best_label_name,
+                values_info=_ValuesInfo(values, None, label_name),
+                best_values_info=best_values_info,
             )
         )
 
@@ -82,21 +82,11 @@ def _get_optimization_history_info_list(
     return optimization_history_info_list
 
 
-class _OptimizationHistoryErrorBarInfo(NamedTuple):
-    trial_numbers: List[int]
-    value_means: List[float]
-    value_stds: List[float]
-    label_name: str
-    best_value_means: Optional[List[float]]
-    best_value_stds: Optional[List[float]]
-    best_label_name: Optional[str]
-
-
 def _get_optimization_history_error_bar_info(
     study: Union[Study, Sequence[Study]],
     target: Optional[Callable[[FrozenTrial], float]],
     target_name: str,
-) -> Optional[_OptimizationHistoryErrorBarInfo]:
+) -> Optional[_OptimizationHistoryInfo]:
 
     info_list = _get_optimization_history_info_list(study, target, target_name)
     if info_list is None:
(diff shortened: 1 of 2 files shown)


knshnb:
I feel now there are a bit too many functions in `optuna/visualization/optimization_history.py`. Maybe we can move the code of `_get_optimization_history_plot` into `plot_optimization_history`?

HideakiImamura:

optuna/visualization/_optimization_history.py
```diff
@@ -23,6 +24,111 @@
...
+    mean_of_values = [np.mean(v) if len(v) > 0 else None for v in values]
+    std_of_values = [np.std(v) if len(v) > 0 else None for v in values]
+    trial_numbers = np.arange(max_trial_number + 2)[[v is not None for v in mean_of_values]]
+    value_means = np.asarray(mean_of_values)[trial_numbers]
+    value_stds = np.asarray(std_of_values)[trial_numbers]
+
+    best_value_means: Optional[List[float]]
+    best_value_stds: Optional[List[float]]
+    best_label_name: Optional[str]
+    if target is None:
```
contramundum53, line 104: I'm sorry, I didn't understand why you only calculate `best_values` when `target is None`. I thought that has nothing to do with `target`.

contramundum53:
The warnings are also common logic between plotly and matplotlib implementations. How about extracting them also?

contramundum53:
The visual regression test looks OK, but I'd like to see how it looks when there are more than one studies.

optuna/visualization/_optimization_history.py
```diff
@@ -123,79 +218,48 @@ def _get_optimization_histories_with_error_bar(
...
 
-    _target: Callable[[FrozenTrial], float]
-    if target is None:
-
-        def _target(t: FrozenTrial) -> float:
-            return cast(float, t.value)
-
-    else:
-        _target = target
+    eb_info = _get_optimization_history_error_bar_info(studies, target, target_name)
```
contramundum53, line 222: For testability, maybe this function could receive `_OptimizationHistoryErrorBarInfo` directly. This way, we could test `_get_optimization_history_error_bar_info` and `_get_optimization_histories_with_error_bar` separately, i.e., we do not need to make `studies` to test `_get_optimization_histories_with_error_bar`.

optuna/visualization/_optimization_history.py
```diff
@@ -215,42 +279,26 @@ def _get_optimization_histories(
     layout: "go.Layout",
 ) -> "go.Figure":
 
+    info_list = _get_optimization_history_info_list(studies, target, target_name)
```
contramundum53, line 282: Ditto.

optuna/visualization/matplotlib/_optimization_history.py
```diff
@@ -139,73 +114,35 @@ def _get_optimization_histories_with_error_bar(
...
-            target_values[t.number].append(_target(t))
-
-    mean_of_target_values = [np.mean(v) if len(v) > 0 else np.nan for v in target_values]
-    std_of_target_values = [np.std(v) if len(v) > 0 else np.nan for v in target_values]
-    trial_numbers = np.arange(max_trial_number + 2)[
-        [v is not np.nan for v in mean_of_target_values]
-    ]
-    means = np.asarray(mean_of_target_values)[trial_numbers]
-    stds = np.asarray(std_of_target_values)[trial_numbers]
+    eb_info = _get_optimization_history_error_bar_info(studies, target, target_name)
```
contramundum53, line 117: Ditto.

optuna/visualization/matplotlib/_optimization_history.py
```diff
@@ -217,40 +154,30 @@ def _get_optimization_histories(
     target_name: str,
     ax: "Axes",
 ) -> "Axes":
+
+    info_list = _get_optimization_history_info_list(studies, target, target_name)
```
contramundum53, line 158: Ditto.

contramundum53:
> I feel now there are a bit too many functions in `optuna/visualization/optimization_history.py`. Maybe we can move the code of `_get_optimization_history_plot` into `plot_optimization_history`?

Sorry for asking for a (partial) revert, but while writing #3745, I noticed that it would be easier to test if `_get_optimization_history_plot` had a single argument receiving the `Info` object, since we do not need to create `study` each time. How about this?

knshnb:
Thanks for the comment! I changed the code to pass `Info` directly to `_get_optimization_history_plot_with_error_bar` and `_get_optimization_history_plot`. PTAL. @contramundum53

HideakiImamura (commented):
Basically, LGTM. I would like to wait that we add the multiple studies setting in the visual regression test.

optuna/visualization/_optimization_history.py
```diff
@@ -74,187 +174,75 @@ def objective(trial):
...
-    return _get_optimization_history_plot(studies, target, target_name, error_bar)
-
-
-def _get_optimization_history_plot(
-    studies: List[Study],
-    target: Optional[Callable[[FrozenTrial], float]],
-    target_name: str,
-    error_bar: bool,
-) -> "go.Figure":
+    _check_plot_args(study, target, target_name)
```
HideakiImamura, line 177: This check of arguments can be done in the `_get_xxx_info`.

optuna/visualization/_optimization_history.py
```diff
@@ -74,187 +174,75 @@ def objective(trial):
...
 
     layout = go.Layout(
         title="Optimization History Plot",
         xaxis={"title": "Trial"},
         yaxis={"title": target_name},
     )
```
HideakiImamura, line 183: Should be moved in `_get_optimization_history_plot`.

optuna/visualization/_optimization_history.py
```diff
@@ -74,187 +174,75 @@ def objective(trial):
...
-        _logger.warning("There are no complete trials.")
-        return go.Figure(data=[], layout=layout)
-
     if error_bar:
-        return _get_optimization_histories_with_error_bar(studies, target, target_name, layout)
+        info = _get_optimization_history_error_bar_info(study, target, target_name)
+        info_list = None if info is None else [info]
     else:
-        return _get_optimization_histories(studies, target, target_name, layout)
+        info_list = _get_optimization_history_info_list(study, target, target_name)
```
HideakiImamura, line 189: This logic should be wrapped and removed from here. I means 
```python
def _get_optimization_history_info_list(...) -> List[_OptimizationHistoryInfo]:
    # Create info list
    
    # If error_bar=False or created info list is dummy, return the info list

    # Create error bar info, and return it


def plot_optimization_history(...) -> go.Figure:
    _imports.check()    
    info_list = _get_optimization_history_info_list(...)
    return _get_optimization_history_plot(info_list)
```

HideakiImamura (commented):
Thanks for the update. Overall, you have improved very well. I have several comments. Please take a look.

knshnb:
@HideakiImamura It was still in the middle of the update when you reviewed, but thank you for the comment anyway! I think some of the suggestions were resolved by my follow-up commits, but let me address the rest of them tomorrow.

HideakiImamura (commented):
Thanks for the update! Basically, LGTM. Could you apply other my suggestions?

optuna/visualization/_optimization_history.py
```diff
@@ -23,6 +23,106 @@
...
+
+def _get_optimization_history_error_bar_info(
+    study: Union[Study, Sequence[Study]],
+    target: Optional[Callable[[FrozenTrial], float]],
+    target_name: str,
+) -> Optional[_OptimizationHistoryInfo]:
+
+    info_list = _get_optimization_history_info_list(study, target, target_name)
+    if info_list is None:
+        return None
```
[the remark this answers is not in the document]
knshnb: I think `None` is more natural as a mean and std of an empty list and also, `None` is easier to check here: https://github.com/optuna/optuna/pull/3736/files#diff-09c6bccd75bec4a1a72b73b7917f5998892505af80949580ce2d1e4b8750d485R187.
Instead, I removed the `Optional` of `_get_optimization_history_info_list`'s return value.

optuna/visualization/_optimization_history.py
knshnb: Fixed.

optuna/visualization/_optimization_history.py
knshnb: Fixed.

optuna/visualization/_optimization_history.py
```diff
@@ -23,6 +23,106 @@
...
+    max_trial_number = max(max(info.trial_numbers) for info in info_list)
+
+    # Calculate mean and std of target values for each trial number.
+    values: List[List[float]] = [[] for _ in range(max_trial_number + 2)]
```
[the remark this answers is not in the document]
knshnb: Removed!

optuna/visualization/_optimization_history.py
```diff
@@ -23,6 +23,106 @@
...
+    # TODO(knshnb): Took over `+2` from the previous code, but not sure why it is necessary.
+    for trial_numbers, values_info, _ in info_list:
+        for trial_number, value in zip(trial_numbers, values_info.values):
+            values[trial_number].append(value)
+    mean_of_values = [np.mean(v) if len(v) > 0 else None for v in values]
+    std_of_values = [np.std(v) if len(v) > 0 else None for v in values]
+    eb_trial_numbers = np.arange(max_trial_number + 2)[[v is not None for v in mean_of_values]]
+    value_means = np.asarray(mean_of_values)[eb_trial_numbers]
+    value_stds = np.asarray(std_of_values)[eb_trial_numbers]
+    eb_values_info = _ValuesInfo(value_means, value_stds, target_name)
```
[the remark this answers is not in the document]
knshnb: We decided to keep this to avoid the conflict of variable names.

optuna/visualization/_optimization_history.py
knshnb: Fixed!

knshnb:
Let me summarize the major updates from the original PR.
- Changed `Info` class so that we can handle both with and without error bars by `_OptimizationHistoryInfo`
- Unified the plot logic both with and without error bars.
- Unified `_get_optimization_history_info_list` and `_get_optimization_history_error_bar_info` into one function.
- DRY calculation of mean and std for error bars.

@HideakiImamura @contramundum53 PTAL 🙏

optuna/visualization/_optimization_history.py
```diff
@@ -23,6 +24,95 @@
...
+                best_values_info=best_values_info,
+            )
+        )
+
+    if len(info_list) == 0:
+        _logger.warning("There are no studies.")
+
+    if sum(len(info.trial_numbers) for info in info_list) == 0:
+        _logger.warning("There are no complete trials.")
+        info_list.clear()
```
contramundum53, line 84: Currently, `info_list` contains n `_OptimizationHistoryInfo`s when there are n studies, except when all studies are empty.
Is there some reason for this (seemingly inconsistent) behavior?

knshnb: I followed the [original code](https://github.com/optuna/optuna/blob/191f4db7228eb205e409f9c44fa1f64595823194/optuna/visualization/_optimization_history.py#L110-L112).
I don't know why, but it is tested [here](https://github.com/optuna/optuna/blob/191f4db7228eb205e409f9c44fa1f64595823194/tests/visualization_tests/test_optimization_history.py#L146-L148), so it throws an error if we return a list of n empty lists instead of a single empty list.

HideakiImamura (approved):
Thanks for the update! LGTM!

optuna/visualization/_optimization_history.py
contramundum53: I see. I think fixing this behavior can be another PR.

contramundum53 (approved):
LGTM. Thanks!

PR #3738, opened 2022-06-30
Raise error on `nan` in `_constrained_dominates`

contramundum53:
## Motivation
Currently, NSGA-II behaves very counterintuitively when `nan` is given as constraint values. `nan` values are treated as infeasible, but it is currently treated as zero violation.

We resolve this issue by simply raising `ValueError` when a `nan` is found.

## Description of the changes
Raise `ValueError` in `_constrained_dominates` function when a `nan` constraint is encountered.

### DIFF
--- a/optuna/samplers/nsgaii/_sampler.py
+++ b/optuna/samplers/nsgaii/_sampler.py
@@ -485,6 +485,9 @@ def _constrained_dominates(
     assert isinstance(constraints0, (list, tuple))
     assert isinstance(constraints1, (list, tuple))
 
+    if np.any(np.isnan(constraints0)) or np.any(np.isnan(constraints1)):
+        raise ValueError("NaN is not acceptable as constraint value.")
+
     if len(constraints0) != len(constraints1):
         raise ValueError("Trials with different numbers of constraints cannot be compared.")
 
--- a/tests/samplers_tests/test_nsgaii.py
+++ b/tests/samplers_tests/test_nsgaii.py
@@ -153,7 +153,7 @@ def test_constraints_func(constraint_value: float) -> None:
     assert constraints_func_call_count == n_trials
     for trial in study.trials:
         for x, y in zip(trial.system_attrs[_CONSTRAINTS_KEY], (constraint_value + trial.number,)):
-            assert _nan_equal(x, y)
+            assert x == y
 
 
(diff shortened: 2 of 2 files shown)


contramundum53:
I made a simple tracking issue -> #3746

HideakiImamura (approved):
LGTM.

HideakiImamura:

optuna/samplers/nsgaii/_sampler.py
```diff
@@ -485,6 +485,9 @@ def _constrained_dominates(
     assert isinstance(constraints0, (list, tuple))
     assert isinstance(constraints1, (list, tuple))
 
+    if np.any(np.isnan(constraints0)) or np.any(np.isnan(constraints1)):
+        raise ValueError("NaN is not acceptable as constraint value.")
```
keisuke-umezawa, line 489: [question] For example, only constraints1 has nan, can we assume that constraints0 is feasible? like follows

```
    if constraints0 is not None and constraints1 is None:
        # Trial x has constraint values, but y doesn't.
        return True
```

contramundum53: I think having `nan` in constraints means there is a bug in the user code, and we should notify the user as early as possible. That's why I didn't consider `nan` constraint as an infeasible trial but reported error.

HideakiImamura:
@keisuke-umezawa is busy now, so could you review this PR? @nzw0301

nzw0301 (commented):
I think we don't need `_nan_equal` anymore because this PR drops `nan` from the parameterization for the constraint. Could you remove it for readability?
https://github.com/optuna/optuna/blob/6c643ea342ff9fb2dc85fc4c9e5f380871968eaa/tests/samplers_tests/test_nsgaii.py#L156

tests/samplers_tests/test_nsgaii.py
```diff
@@ -157,6 +156,15 @@ def constraints_func(trial: FrozenTrial) -> Sequence[float]:
             assert _nan_equal(x, y)
```
contramundum53, line 156:
```suggestion
            assert x == y
```

contramundum53:
Fixed. PTAL. @nzw0301

nzw0301 (approved):
Look great. Thanks!

Issue #3675, opened 2022-06-15
The `optuna.samplers._tpe.sampler._calculate_weights_below_for_multi_objective` should return `[]` when `indices=[]`


HideakiImamura:
### Expected behavior

The `optuna.samplers._tpe.sampler._calculate_weights_below_for_multi_objective` should return `[]` when `indices=[]`, but it raises an `IndexError`. 

### Environment

- Optuna version: current master https://github.com/optuna/optuna/commit/d3c53e053a18341405cdfc9587d4525fdce7494d
- Python version: 3.9
- OS: mac, linux
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
(venv) mamu@HideakinoMacBook-puro 3662 % python hoge.py
Traceback (most recent call last):
  File "/Users/mamu/Dev/pfn/code_snippets_for_optuna/3662/hoge.py", line 5, in <module>
    optuna.samplers._tpe.sampler._calculate_weights_below_for_multi_objective(
  File "/Users/mamu/Dev/pfn/optuna/optuna/samplers/_tpe/sampler.py", line 764, in _calculate_weights_below_for_multi_objective
    cvals = list(config_values.values())[0][indices]
IndexError: arrays used as indices must be of integer (or boolean) type
```

### Steps to reproduce

```python
import optuna
import numpy as np


optuna.samplers._tpe.sampler._calculate_weights_below_for_multi_objective(
    {"x": np.array([1.0, 2.0, 3.0], dtype=float)},
    [(0, [0.2, 0.5]), (0, [0.9, 0.4]), (0, [1, 1])],
    np.array([]),
)
```

contramundum53:
Specifying `np.array([], dtype=np.int64)` as indices will correctly return `[]`.

PR #3741, opened 2022-06-30
Fix `test_calculate_weights_below`

contramundum53:
## Motivation
Related issue: #3675 
We fix a bug in `test_calculate_weights_below` about empty input.

## Description of the changes
Specify `np.array([], dtype=np.int64)` as `indices` instead of `np.array([])`.


toshihikoyanase:
@HideakiImamura  Could you review this PR as an author of https://github.com/optuna/optuna/issues/3675? I think we can merge this PR with a single approval since it only fixes a test case and does not change library code.

PR #3742, opened 2022-06-30
Fix `_calculate_weights` such that it throws `ValueError` on invalid weights

### LINKED ISSUE #3687: An invalid `weights` argument of `TPESamper` does not raise any error
### Expected behavior

If we pass an invalid `weights` to `TPESamper`, it does not raise any error. We should raise an `ValueError` for those ivalid `weight`s. Examples of invalid `weight`s are as follows.

```python
lambda x: np.zeros(x),
lambda x: -np.ones(x),
lambda x: float("inf") * np.ones(x),
lambda x: -float("inf") * np.ones(x),
lambda x: np.asarray([float("nan") for _ in range(x)])
```

### Environment

- Optuna version: current master https://github.com/optuna/optuna/commit/b213753a816689a97b6433e1e780ca32c4293e07
- Python version: 3.9
- OS: mac

```shell
(venv) mamu@HideakinoMacBook-puro code_snippets_for_optuna % python tpe-tests/hoge.py 
/Users/mamu/Dev/pfn/optuna/optuna/samplers/_tpe/parzen_estimator.py:230: RuntimeWarning: invalid value encountered in true_divide
  weights /= weights.sum()
[nan]
[1.]
[nan]
[nan]
[nan]
```

### Steps to reproduce

```python
import optuna
from optuna.samplers._tpe.parzen_estimator import _ParzenEstimator
from optuna.samplers._tpe.parzen_estimator import _ParzenEstimatorParameters

import numpy as np

def test_invalid_weights(weights):
    parameters = _ParzenEstimatorParameters(
        prior_weight=1.0,
        consider_prior=False,
        consider_magic_clip=False,
        consider_endpoints=False,
        weights=weights,
        multivariate=False,
    )
    mpe = _ParzenEstimator(
        {"a": np.asarray([0.0])}, {"a": distributions.FloatDistribution(-1.0, 1.0)}, parameters
    )

    print(mpe._weights)

weights_funcs = [
	lambda x: np.zeros(x),
	lambda x: -np.ones(x),
	lambda x: float("inf") * np.ones(x),
	lambda x: -float("inf") * np.ones(x),
	lambda x: np.asarray([float("nan") for _ in range(x)]),
]

for weights in weights_funcs:
	test_invalid_weights(weights)

```

_No response_

contramundum53:
## Motivation
Resolve #3687.

## Description of the changes
Add checks for invalid weights in `_calculate_weights`.

### DIFF
--- a/optuna/samplers/_tpe/parzen_estimator.py
+++ b/optuna/samplers/_tpe/parzen_estimator.py
@@ -1,3 +1,4 @@
+from multiprocessing.sharedctypes import Value
 from typing import Callable
 from typing import Dict
 from typing import NamedTuple
@@ -54,7 +55,6 @@ class _ParzenEstimator:
         parameters: _ParzenEstimatorParameters,
         predetermined_weights: Optional[np.ndarray] = None,
     ) -> None:
-
         self._search_space = search_space
         self._parameters = parameters
         self._n_observations = next(iter(observations.values())).size
@@ -204,7 +204,6 @@ class _ParzenEstimator:
         return ret
 
     def _calculate_weights(self, predetermined_weights: Optional[np.ndarray]) -> np.ndarray:
-
         # We decide the weights.
         consider_prior = self._parameters.consider_prior
         prior_weight = self._parameters.prior_weight
@@ -216,6 +215,13 @@ class _ParzenEstimator:
 
         if predetermined_weights is None:
             w = weights_func(n_observations)[:n_observations]
+            if w is not None:
+                if np.any(w < 0):
+                    raise ValueError("Negative parameters.weights are not allowed.")
+                if len(w) > 0 and np.sum(w) <= 0:
+                    raise ValueError("parameters.weights of all zeros are not allowed")
+                if not np.all(np.isfinite(w)):
+                    raise ValueError("parameters.weights of infinite or NaN values are not allowed.")
         else:
             w = predetermined_weights[:n_observations]
 
--- a/optuna/samplers/_tpe/parzen_estimator.py
+++ b/optuna/samplers/_tpe/parzen_estimator.py
@@ -203,6 +203,7 @@ class _ParzenEstimator:
         return ret
 
     def _calculate_weights(self, predetermined_weights: Optional[np.ndarray]) -> np.ndarray:
+
         # We decide the weights.
         consider_prior = self._parameters.consider_prior
         prior_weight = self._parameters.prior_weight
--- a/optuna/samplers/_tpe/parzen_estimator.py
+++ b/optuna/samplers/_tpe/parzen_estimator.py
@@ -216,12 +216,12 @@ class _ParzenEstimator:
             w = weights_func(n_observations)[:n_observations]
             if w is not None:
                 if np.any(w < 0):
-                    raise ValueError("Negative parameters.weights are not allowed.")
+                    raise ValueError(f"The `weights` function is not allowed to return negative values {w}. The argument of the `weights` function is {n_observations}.")
                 if len(w) > 0 and np.sum(w) <= 0:
-                    raise ValueError("parameters.weights of all zeros are not allowed")
+                    raise ValueError(f"The `weight` function is not allowed to return all-zero values {w}. The argument of the `weights` function is {n_observations}.")
                 if not np.all(np.isfinite(w)):
                     raise ValueError(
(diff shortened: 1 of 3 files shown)


HideakiImamura:

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -54,7 +54,6 @@ def __init__(
         parameters: _ParzenEstimatorParameters,
         predetermined_weights: Optional[np.ndarray] = None,
     ) -> None:
-
```
HideakiImamura, line 57: Unnecessary removal.

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -204,7 +203,6 @@ def log_pdf(self, samples_dict: Dict[str, np.ndarray]) -> np.ndarray:
         return ret
 
     def _calculate_weights(self, predetermined_weights: Optional[np.ndarray]) -> np.ndarray:
-
```
HideakiImamura, line 207: Ditto.

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -216,6 +214,15 @@ def _calculate_weights(self, predetermined_weights: Optional[np.ndarray]) -> np.
...
             w = weights_func(n_observations)[:n_observations]
+            if w is not None:
+                if np.any(w < 0):
+                    raise ValueError("Negative parameters.weights are not allowed.")
+                if len(w) > 0 and np.sum(w) <= 0:
+                    raise ValueError("parameters.weights of all zeros are not allowed")
+                if not np.all(np.isfinite(w)):
+                    raise ValueError(
+                        "parameters.weights of infinite or NaN values are not allowed."
+                    )
```
HideakiImamura, line 225: How about returning more descriptive error message as follows?
```suggestion
                if np.any(w < 0):
                    raise ValueError(f"The `weights` function is not allowed to return negative values {w}. The argument of the `weights` function is {n_observations}.")
                if len(w) > 0 and np.sum(w) <= 0:
                    raise ValueError(f"The `weight` function is not allowed to return all-zero values {w}. The argument of the `weights` function is {n_observations}.")
                if not np.all(np.isfinite(w)):
                    raise ValueError(
                        f"The `weights`function is not allowed to return infinite or NaN values {w}. The argument of the `weights` function is {n_observations}."
                    )
```

HideakiImamura (commented):

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -204,7 +203,6 @@ def log_pdf(self, samples_dict: Dict[str, np.ndarray]) -> np.ndarray:
         return ret
 
     def _calculate_weights(self, predetermined_weights: Optional[np.ndarray]) -> np.ndarray:
-
         # We decide the weights.
```
contramundum53, line 206:
```suggestion

        # We decide the weights.
```
Revert the line.

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -54,7 +54,6 @@ def __init__(
         parameters: _ParzenEstimatorParameters,
         predetermined_weights: Optional[np.ndarray] = None,
     ) -> None:
```
contramundum53, line 56:
```suggestion
    ) -> None:

```
Revert the line.

contramundum53 (commented):
I reverted some changes as @HideakiImamura suggested.

HideakiImamura (approved):
LGTM.

not522 (commented):
Could you add tests for invalid weights?

HideakiImamura:
I think the current `test_invalid_weights` is the test for invalid weights.

HideakiImamura:
@contramundum53 By the way, the linter checks fails due to the line length over. Could you fix it?

contramundum53:
Fixed.

not522:
Let me clarify the conditions for throwing a `ValueError`. If `consider_prior` is `True`, `prior_weight` is added to the weights. Then I think `np.sum(w) == 0` should be allowed in this case.

contramundum53:
Is there any valid use cases where `np.sum(w) == 0`?

not522:
I understood that `np.sum(w) == 0` and `consider_prior` is `True` works but is not useful.

not522 (approved):
LGTM

PR #3743, opened 2022-06-30
Support `inf` values for crowding distance

contramundum53:
## Motivation
Currently, `_calc_crowding_distance` function returns wrong values when there are trials with `inf` values. We try to fix it.

## Description of the changes
We define the crowding distance as the summation of the crowding distance of each dimension of value calculated as follows:
* If all values in that dimension are the same, i.e., `[1, 1, 1]` or `[inf, inf]`, the crowding distances of all trials in that dimension are zero.
* Otherwise, the crowding distances of that dimension is the difference between two nearest values besides that value, one above and one below, divided by the difference between the maximal and minimal finite value of that dimension. Please note that:
  * the nearest value below the minimum is considered to be `-inf` and the nearest value above the maximum is considered to be `inf`, and
  * `inf - inf` and `(-inf) - (-inf)` is considered to be zero.


toshihikoyanase:
@knshnb @xadrianzetx Could you review this PR, please? If you are busy, please feel free to remove the assignment.

optuna/samplers/nsgaii/_sampler.py
```diff
@@ -410,29 +410,46 @@ def after_trial(
...
-        manhattan_distances[population[0].number] = float("inf")
-        manhattan_distances[population[-1].number] = float("inf")
+        vs = (
+            [-float("inf")]
+            + [cast(float, population[j].values[i]) for j in range(len(population))]
+            + [float("inf")]
+        )
+
+        # Smallest finite value.
+        v_min = next(x for x in vs if x != -float("inf"))
```
toshihikoyanase, line 432: I'm not sure, but could you tell me what happens when objective values of all individuals are `-inf`?

contramundum53: In that case, it will not reach this line because of L422.

toshihikoyanase: That makes sense. Thanks.

optuna/samplers/nsgaii/_sampler.py
```diff
@@ -410,29 +410,46 @@ def after_trial(
...
-        if width == 0:
+        # If population have the same values[i], ignore that value.
+        if population[0].values[i] == population[-1].values[i]:
             continue
 
-        manhattan_distances[population[0].number] = float("inf")
-        manhattan_distances[population[-1].number] = float("inf")
+        vs = (
+            [-float("inf")]
+            + [cast(float, population[j].values[i]) for j in range(len(population))]
```
xadrianzetx, line 427: Do we need to cast here? `FrozenTrial.values` should be already typed `List[float]`.

tests/samplers_tests/test_nsgaii.py
```diff
@@ -432,12 +432,17 @@ def test_fast_non_dominated_sort_missing_constraint_values() -> None:
...
             [[float("-inf")], [1], [2]],
-            [float("inf"), float("nan"), float("inf")],
-        ),  # TODO(knshnb): Decide expected behavior for this case.
+            [float("inf"), float("inf"), float("inf")],
+        ),
+        ([[float("inf")], [float("inf")], [float("inf")]], [0, 0, 0]),
+        (
+            [[-float("inf")], [-float("inf")], [-float("inf")], [0], [1], [2], [float("inf")]],
+            [0, 0, float("inf"), float("inf"), 1, float("inf"), float("inf")],
+        ),
```
xadrianzetx, line 445: How about adding following test cases:
```python
([[-float("inf")], [-float("inf")]], [0, 0]),
([[-float("inf")], [float("inf")]], [float("inf"), float("inf")]),
```

optuna/samplers/nsgaii/_sampler.py
```diff
@@ -410,29 +410,46 @@ def after_trial(
...
-            assert v_high is not None
-            assert v_low is not None
+        # Largest finite value.
+        v_max = next(x for x in reversed(vs) if x != float("inf"))
 
-            manhattan_distances[population[j].number] += (v_high - v_low) / width
+        width = v_max - v_min
+        if width <= 0:
+            # width == 0 or width == -inf
+            width = 1
```
xadrianzetx, line 440:
```suggestion
            width = 1.0
```

xadrianzetx (changes requested):
This is out of scope here, but I think by representing values from each dimension as a numpy array, we could simplify some operations and increase readability.

optuna/samplers/nsgaii/_sampler.py
contramundum53: Unfortunately, `FrozenTrial.values` has type `Optional[List[float]]`. (I'm not sure why `Optional` is necessary here.)
The code is still wrong, I'll change it to the following.
```suggestion
            + [cast(List[float], population[j].values)[i] for j in range(len(population))]
```

contramundum53:
@xadrianzetx I fixed the above points. PTAL.

optuna/samplers/nsgaii/_sampler.py
```diff
@@ -410,29 +410,46 @@ def after_trial(
...
-            assert v_high is not None
-            assert v_low is not None
+        # Largest finite value.
+        v_max = next(x for x in reversed(vs) if x != float("inf"))
 
-            manhattan_distances[population[j].number] += (v_high - v_low) / width
+        width = v_max - v_min
+        if width <= 0:
+            # width == 0 or width == -inf
+            width = 1.0
```
knshnb, line 440: Let me clarify my understanding. This condition is satisfied only when there are only 0 (then, `width==-inf`) or 1 (then, `width==0`) unique finite values. In those cases, only `-inf`, `0`, and `inf` can appear as the differences, so the value of `width` doesn't matter.

optuna/samplers/nsgaii/_sampler.py
```diff
@@ -410,29 +410,46 @@ def after_trial(
...
+        def subtract_nonfinite(x: float, y: float) -> float:
+            if x == y:
+                return 0
+            else:
+                return x - y
+
+        for j in range(len(population)):
+            manhattan_distances[population[j].number] += (
+                subtract_nonfinite(vs[j + 2], vs[j]) / width
+            )
```
knshnb, line 452: A nit, but I think there is no need to define this as a function since used only once.
```suggestion
        for j in range(len(population)):
            # inf - inf and (-inf) - (-inf) is considered to be zero.
            gap = 0.0 if vs[j] == vs[j + 2] else vs[j + 2] - vs[j]
            manhattan_distances[population[j].number] += gap / width
```

knshnb (commented):
Thanks for the careful improvement in the algorithm of crowding distance! It looks fine overall and I left some small comments.
Also, the code might become a little complex with this PR, so it might be better to add what you described in "Description of the changes" as a comment or docstring.

optuna/samplers/nsgaii/_sampler.py
contramundum53: Yes, you are correct.

optuna/samplers/nsgaii/_sampler.py
xadrianzetx: Seems like `typing.cast` does not cost much at runtime and keeps mypy happy, so I agree with the changes :+1:.

xadrianzetx (approved):
LGTM!

knshnb (approved):
Thanks for the update! LGTM.

Issue #3747, opened 2022-07-01
Add line to Pareto front for multi-objective optimisation


JAEarly:
### Motivation

Currently, for the Pareto plots, the dominating points are not connected, only shown in another colour. Adding a line connecting the dominating points would make it clearer which points are on the front and which are not. I find it especially hard to distinguish between dominating and non-dominating points for the extreme lights and darks of the marker colour gradients. 

### Description

This would be a relatively minor change to add a line connecting the points on the Pareto front.

I had a quick look at making these changes locally. Below are examples in 2D and 3D. I'm happy to open a PR if this is accepted as being a good idea.

![image](https://user-images.githubusercontent.com/7620667/176888141-2ed1e2c2-ad36-45ac-86d7-ae367c7855f8.png)

![image](https://user-images.githubusercontent.com/7620667/176888151-0cf13a99-bc35-4a25-9be3-a1ce4642ec1d.png)

nzw0301:
Thank you for proposing the feature and sharing the expected results! Personally, the proposed feature looks useful.

JAEarly:
I've created a PR: #3749

y0z:
Hi. Thank you for your useful proposal.
I leave a brief comment that may be helpful.
This is not mandatory but can be our future work.

In this case, I think that showing [an attainment surface](https://eden.dei.uc.pt/~cmfonsec/fonseca-ppsn1996-reprint.pdf) (i.e., the connection of the front of all dominated areas) is better than simple linear interpolation.
The former is a more popular way in the community and it never contains (imaginary) points that are not dominated by the incumbent nondominated objective vectors.

[2d example](https://www.researchgate.net/figure/Attainment-curve-and-hypervolume-in-a-bi-objective-space-with-both-objectives-being_fig4_271936836)
![Attainment-curve-and-hypervolume-in-a-bi-objective-space-with-both-objectives-being](https://user-images.githubusercontent.com/30489874/176998063-c45d8bf2-7b93-4f0c-a0e8-b7974d7a150e.png)
and [3d examples](https://www.cs.bham.ac.uk/~jdk/plot_attainments/).

The concept of the attainment surface can also be extended to multiple runs to assess an optimization algorithm performance (e.g., https://mlopez-ibanez.github.io/eaf/) though I'm not familiar with the implementation details of 3d-case and multiple runs.

PR #3749, opened 2022-07-01
Add line to Pareto front visualisation

JAEarly:
## Motivation
Currently, for the Pareto plots, the dominating points are not connected, only shown in another colour. Adding a line connecting the dominating points would make it clearer which points are on the front and which are not. I find it especially hard to distinguish between dominating and non-dominating points for the extreme lights and darks of the marker colour gradients.

See #3747.

## Description of the changes
The plotting function `_make_scatter_object` in `_pareto_front.py` now has an additional argument `mode` . This determines whether a line should be added. Other thing of note is that if a line is being added, the points have to be sorted. For efficiency, this sorting is only performed if the line is required (i.e., when plotting the points with the line, the points are not sorted).

### DIFF
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -11,6 +11,8 @@ from typing import Tuple
 from typing import Union
 import warnings
 
+import numpy as np
+
 import optuna
 from optuna.exceptions import ExperimentalWarning
 from optuna.study import Study
@@ -387,7 +389,9 @@ def _make_scatter_object(
         y = [values[axis_order[1]] for _, values in trials_with_values]
         # If a line is being added, the points need to be in order.
         if "lines" in mode:
-            x, y = zip(*sorted(zip(x, y), key=lambda p: p[0]))
+            idx_order = np.argsort(x)
+            x = [x[idx] for idx in idx_order]
+            y = [y[idx] for idx in idx_order]
         return go.Scatter(
             x=x,
             y=y,
@@ -400,10 +404,13 @@ def _make_scatter_object(
     elif n_targets == 3:
         x = [values[axis_order[0]] for _, values in trials_with_values]
         y = [values[axis_order[1]] for _, values in trials_with_values]
-        z = [values[axis_order[1]] for _, values in trials_with_values]
+        z = [values[axis_order[2]] for _, values in trials_with_values]
         # If a line is being added, the points need to be in order.
         if "lines" in mode:
-            x, y, z = zip(*sorted(zip(x, y, z), key=lambda p: p[0]))
+            idx_order = np.argsort(x)
+            x = [x[idx] for idx in idx_order]
+            y = [y[idx] for idx in idx_order]
+            z = [z[idx] for idx in idx_order]
         return go.Scatter3d(
             x=x,
             y=y,
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -17,6 +17,7 @@ import optuna
 from optuna.exceptions import ExperimentalWarning
 from optuna.study import Study
 from optuna.study._multi_objective import _get_pareto_front_trials_by_trials
+from optuna.study._study_direction import StudyDirection
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
 from optuna.visualization._plotly_imports import _imports
@@ -35,6 +36,7 @@ class _ParetoFrontInfo(NamedTuple):
     non_best_trials_with_values: Optional[Sequence[Tuple[FrozenTrial, Sequence[float]]]]
     infeasible_trials_with_values: Optional[Sequence[Tuple[FrozenTrial, Sequence[float]]]]
     axis_order: Sequence[int]
+    pareto_front_line_values: Optional[Tuple[Sequence[float], Sequence[float]]]
 
 
 def plot_pareto_front(
@@ -45,6 +47,7 @@ def plot_pareto_front(
     axis_order: Optional[List[int]] = None,
     constraints_func: Optional[Callable[[FrozenTrial], Sequence[float]]] = None,
     targets: Optional[Callable[[FrozenTrial], Sequence[float]]] = None,
+    pareto_front_line: Optional[str] = None,
 ) -> "go.Figure":
     """Plot the Pareto front of a study.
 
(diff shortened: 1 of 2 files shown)


HideakiImamura:
@nzw0301 Could you give any comments about this PR?

JAEarly:
Hi @nzw0301 and @HideakiImamura, thanks for the feedback. Please see my responses to each point below:

### 1. Implementation design
This makes sense. I'm happy to change my code such that the required information is provided in `_get_pareto_front_info`.

### 2. The style of line
I've just been looking in to changing how the line is plotted such that it uses attainment surfaces rather than interpolation. My intention was to add an optional argument `pareto_front_line` to the `plot_pareto_front` function with three options (None, 'linear', and 'attainment'). This would allow the user to select if they wanted a linear interpolation, an attainment surface, or no line at all (i.e., the current functionality).

### 3. 3D Use Case
I must admit I'm unfamiliar with Pareto fronts in 3D (and 3D plotting in general). Would it be appropriate to leave the linear/attainment options unimplemented in the 3D case? The Pareto front line would only be available in the 2D use case. The 3D use case could be left as an open issue for further work.

nzw0301:
@JAEarly thank you for your response! I'm glad to hear that you are interested in implementing the feature! Let's discuss a little bit more before moving forward with this PR. 

### 1.  Implementation design

Great! I suppose we have two approaches to creating Pareto front linear information: (a) adding the info to `_get_pareto_front_info` or (b) defining a new `_get_` method that might take `_ParetoFrontInfo`. 

### 2.  The style of line

Indeed, we also agreed that the introduced feature would be an optional argument rather than a default when we discussed it. I'm not sure we need to implement two ways to visualise the Pareto-front liner, so I would like to ask other dev members' opinions.

### 3. 3D Use Case

It sounds reasonable to leave 3d case as a follow-up!

JAEarly:
@nzw0301 I've made some changes based on your feedback.

The Pareto front line is now computed in `_get_pareto_front_info` and stored in `_ParetoFrontInfo` as the list of values that represent the line to be plotted: `pareto_front_line_values`. This can either be computed using linear interpolation or as an attainment surface. I've kept the linear interpolation as that is the most common way that I have seen Pareto fronts plotted.

I've added additional checks for the optional argument `pareto_front_line` to ensure it is None, 'linear', or 'attainment', and I also ensure that is has to be None if the plot is not 2D.

There are also new tests for a) input sanitisation, and b) testing the attainment surface generation to ensure the generated line is correct (depending on optimisation directions).

HideakiImamura:
I don't understand the need to provide this feature as one of Optuna's visualization features, since the scatter plot can adequately approximate the attainment surface when the number of trials is large. Also, I am not sure how useful an interpolated attainment surface would be for a small number of trials. Can you elaborate a bit more on the need for this function, with some evidence from the literature?

HideakiImamura:
@y0z Any comments are welcome.

y0z:
I agree that the scatter plot is usually enough to visualize the result of a single run of 2-dimensional optimization.
Therefore, the 2d empirical attainment function for a single run itself is not so important though its extendability is promising.

I also remark that, in more complex scenarios, another way of visualization can be useful such as comparing the performances of algorithms, understanding the average results of multiple runs, etc. (cf. [The GECCO 2020 tutorial on Visualization in multiobjective optimization](https://dis.ijs.si/tea/Publications/GECCO2020Tutorial.pdf)).
However, those elaborated visualization functionalities are often required by researchers who analyze algorithm behaviors in detail and write papers rather than practitioners who simply conduct hyperparameter optimization.

Indeed, enhancing `optuna.visualization` for multiobjective optimization is generally helpful, but at first, it would be better to discuss which functionalities are actually needed by Optuna users and/or developers rather than immediately implementing adhoc ideas.

In my opinion, visualization functionalities like
- plotting a graph of an achieved hypervolume-vs-# of trials
- supporting multiple objectives in a parallel coordinates plot

seem to be relatively reasonable because these are the natural extensions of useful visualization functionalities for single-objective optimization that `optuna.visualization` provides whereas functionalities like

- the 2d empirical attainment function for a single optimization run
- a visualization summarizing the results of multiple studies

should be carefully discussed with respect to their needs.
The latter functionalities may be too simple or too complex as a function of Optuna; therefore it could be better not to provide them or to provide them as functions of a separate toolkit besides Optuna.

JAEarly:
Thank you for the comments.

My original motivation for this contribution is that I have sometimes struggled to distinguish the dominating trials from the non-dominating ones, especially at the extreme trial numbers (i.e., very light or very dark points often look similar despite being from different colour ranges). Adding the linear interpolation (and later the attainment surface) seemed like a natural way to better distinguish the dominating trials. However, alternative solutions could solve my problem (e.g., different marker styles for the dominating/non-dominating trials). I am aware that the trials are distinguished when hovered over, although this is less intuitive that the other solutions.

To give a little more detail on my use case: I'm using a relatively small number of trials (100) as each trial is expensive to compute, and I often end up with a rather small Pareto front that can be difficult to identify with the current setup. I'm approaching this as a practitioner rather than a researcher, but, at least in my opinion and use case, there is a need to add _something_ that improves the visibility of the Pareto front. I can't really weigh in from the pure MOO research perspective as it is not my area of expertise.

Going forward, I'm happy for this to be integrated as part of a wider updated to Optuna's MOO visualisation, or, as an alternative, reworked such that it can be added to the FAQ section if appropriate.

y0z:
Based on my understanding, `include_dominated_trials=False` will resolve your problem.
Please refer to [the documentation](https://optuna.readthedocs.io/en/stable/reference/visualization/generated/optuna.visualization.plot_pareto_front.html#optuna-visualization-plot-pareto-front).

In addition, you can change the color of points by updating their `color` property as follows.

```python
import optuna

def objective(trial):

    v0 = 4 * x ** 2 + 4 * y ** 2
    return v0, v1

study = optuna.create_study(directions=["minimize", "minimize"])
fig = optuna.visualization.plot_pareto_front(study, include_dominated_trials=False)

for d in fig._data:
  d['marker']['color'] = 'black';

fig.show()
```

newplot

HideakiImamura:
Thank you for your explanation @JAEarly. 

> My original motivation for this contribution is that I have sometimes struggled to distinguish the dominating trials from the non-dominating ones, especially at the extreme trials. ... alternative solutions could solve my problem (e.g., different marker styles for the dominating/non-dominating trials).

I see. Then, rather than implementing this feature in Optuna, it may be more useful for users who may encounter the same problem to describe the problem you encountered and how to solve it in the FAQ. How about editing the FAQ?

Issue #3729, opened 2022-06-27
Make MPI test more stable


nzw0301:
### Motivation

Optuna has a few tests for parallel computing, concretely, [chainermn](https://optuna.readthedocs.io/en/latest/reference/generated/optuna.integration.ChainerMNStudy.html) and [ddp pytorch](https://optuna.readthedocs.io/en/latest/reference/generated/optuna.integration.TorchDistributedTrial.html) as in [`.github/workflows/tests-mpi.yml`](https://github.com/optuna/optuna/blob/96b7fe7b945c81f4b1262d2e71cc2e138f9dce18/.github/workflows/tests-mpi.yml) or [`.github/workflows/mac-tests.yml`](https://github.com/optuna/optuna/blob/bf8ec607095a4b1d320b1171fa61227e9bf1366a/.github/workflows/mac-tests.yml). 
 Even though we've separated MPI-related test from the integration test by https://github.com/optuna/optuna/pull/3606, the root issue has not been fixed: the MPI tests are not stable; it fails often, ex: https://github.com/optuna/optuna/runs/7067666490?check_suite_focus=true.

### Description

We would like to make the test more stable, but I don't know how...

PR #3758, opened 2022-07-06
set `OMPI_MCA_rmaps_base_oversubscribe=yes` before `mpirun`

nzw0301:
## Motivation

Resolve https://github.com/optuna/optuna/issues/3729

## Description of the changes

Following the suggestion by https://bitbucket.org/mpi4py/mpi4py/issues/87/multiple-test-errors-with-openmpi-30#comment-42045755.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -76,6 +76,7 @@ jobs:
         COVERAGE_PROCESS_START: .coveragerc
         COVERAGE_COVERAGE: yes
       run: |
+        export OMPI_MCA_rmaps_base_oversubscribe=yes
         mpirun -n 2 -- coverage run -m pytest tests/integration_tests/test_chainermn.py
         coverage combine --append
         coverage xml
--- a/.github/workflows/mac-tests.yml
+++ b/.github/workflows/mac-tests.yml
@@ -125,6 +125,7 @@ jobs:
 
     - name: Tests MPI
       run: |
+        export OMPI_MCA_rmaps_base_oversubscribe=yes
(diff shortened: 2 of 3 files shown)


nzw0301:
At least on my local machien (m1 mac), the it allows us to use more processes.

```bash
$ mpirun -n 9 pytest tests/integration_tests/test_pytorch_distributed.py
--------------------------------------------------------------------------
There are not enough slots available in the system to satisfy the 9
slots that were requested by the application:

  pytest

Either request fewer slots for your application, or make more slots
available for use.

A "slot" is the Open MPI term for an allocatable unit where we can
launch a process.  The number of slots available are defined by the
environment in which Open MPI processes are run:

  1. Hostfile, via "slots=N" clauses (N defaults to number of
     processor cores if not provided)
  2. The --host command line parameter, via a ":N" suffix on the
     hostname (N defaults to 1 if not provided)
  3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
  4. If none of a hostfile, the --host command line parameter, or an
     RM is present, Open MPI defaults to the number of processor cores

In all the above cases, if you want Open MPI to default to the number
of hardware threads instead of the number of processor cores, use the
--use-hwthread-cpus option.

Alternatively, you can use the --oversubscribe option to ignore the
number of available slots when deciding the number of processes to
launch.

# the following line is introduced by this PR
$ export OMPI_MCA_rmaps_base_oversubscribe=yes

$ mpirun -n 9 pytest tests/integration_tests/test_pytorch_distributed.py
# works fine.
...
```

contramundum53 (approved):
LGTM!

HideakiImamura (approved):
LGTM.

PR #3764, opened 2022-07-07
Add tests for `_dominates` function

contramundum53:
## Motivation
Since the discussion in #3683 has become very long, we separate the tests for `_dominates` only.

## Description of the changes
Add tests for `_dominates`. 
We test all pairs of trials with values of combinations of `-inf`, `-1`, `1`, `inf`, and all combinations of directions.

### DIFF
--- a/tests/study_tests/test_multi_objective.py
+++ b/tests/study_tests/test_multi_objective.py
@@ -2,23 +2,18 @@ from typing import Sequence
 
 import pytest
 
-import optuna
 from optuna.study import StudyDirection
 from optuna.study._multi_objective import _dominates
-from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
-
-
-def _create_trial(values: Sequence[float], state: TrialState = TrialState.COMPLETE) -> FrozenTrial:
-    return optuna.trial.create_trial(values=list(values), state=state)
+from optuna.trial import create_trial
 
 
 @pytest.mark.parametrize(
     ("v1", "v2"), [(-1, 1), (-float("inf"), 0), (0, float("inf")), (-float("inf"), float("inf"))]
 )
 def test_dominates_1d_not_equal(v1: float, v2: float) -> None:
-    t1 = _create_trial([v1])
-    t2 = _create_trial([v2])
+    t1 = create_trial(values=[v1])
+    t2 = create_trial(values=[v2])
 
     assert _dominates(t1, t2, [StudyDirection.MINIMIZE])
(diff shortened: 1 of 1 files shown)


HideakiImamura:

tests/study_tests/test_multi_objective.py
```diff
@@ -0,0 +1,121 @@
...
+    #
+    # In the following code, we check that for each position of "t" cell, the relation
+    # above holds.
+
+    # Generate the set of all possible indices.
+    all_indices = set((i, j) for i in range(len(vals)) for j in range(len(vals)))
+    for (t_i, t_j) in all_indices:
+        # Generate the set of all indices that dominates the current index.
+        dominating_indices = set(
+            (D_i, D_j) for D_i in range(t_i + 1) for D_j in range(t_j, len(vals))
```
toshihikoyanase, line 68: A nit, but could you keep the coding convention?
We may also update the table above.
https://peps.python.org/pep-0008/#function-and-variable-names
```suggestion
            (d_i, d_j) for d_i in range(t_i + 1) for d_j in range(t_j, len(vals))
```

tests/study_tests/test_multi_objective.py
```diff
@@ -0,0 +1,121 @@
...
+
+import optuna
+from optuna.study import StudyDirection
+from optuna.study._multi_objective import _dominates
+from optuna.trial import FrozenTrial
+from optuna.trial import TrialState
+
+
+def _create_trial(values: Sequence[float], state: TrialState = TrialState.COMPLETE) -> FrozenTrial:
+    return optuna.trial.create_trial(values=list(values), state=state)
```
toshihikoyanase, line 13: IMO, we can remove this helper function by `from optuna.trial import create_trial`. This is because the values pass to this functions are always lists, and the default value of `state` is `TrialState.COMPLETE`.

https://optuna.readthedocs.io/en/latest/reference/generated/optuna.trial.create_trial.html#optuna.trial.create_trial

toshihikoyanase (commented):
Sorry for the delayed response. I have two small comments.

tests/study_tests/test_multi_objective.py
c-bata: Agree. It looks enough to call `create_trial()` directly.

contramundum53:
Fixed. PTAL.

c-bata (approved):
LGTM!

toshihikoyanase (approved):

Issue #2524, opened 2021-03-25
Confusion Regarding plot_contour


osama-usuf:
**Hello everybody,**

So I am using Optuna for optimizing a network with multiple hyper-parameters (5, to be exact). While I'm finding the visualization tools really helpful, I have a small confusion about the **`optuna.visualization.plot_contour`** method. When I use this method over my problem, Optuna provides a total of 25 plots (a separate contour plot for each possible pair of hyperparameters).

What I am confused by is the following: when using GridSampler, it is possible for two hyperparameters to have the same exact values over multiple trials, where the difference in each of these trials might be in one (or more) of the other three hyperparameters. Now, in a given contour plot (one of the 25) - how exactly is the objective value being calculated? Is it being summed/averaged over all trials, ignoring the other hyperparameters (this is what I believe is happening after going through the underlying code), or is it calculating partial dependencies in some other way? If this is the case, then doesn't it make more sense to have the contour plots depict the best objective values only (maximum/minimum, depending on the study direction)?

I'd appreciate any clarifications, thanks a bunch!

fsmosca:
> What I am confused by is the following: when using GridSampler, it is possible for two hyperparameters to have the same exact values over multiple trials, where the difference in each of these trials might be in one (or more) of the other three hyperparameters. Now, in a given contour plot (one of the 25) - how exactly is the objective value being calculated?

Contour plot has 3 axis, the x, y and depth, like a bird's-eye view we are on top looking down. The x and y are the parameter values while the depth is the objective value shown in color because we are looking down we cannot see the depth difference but we can see which one is high or low by color. If x and y are the same for a given 2 trials on the same parameter pair for example then the location of the point is the same. 

For direction=maximize if the color around those points are darker the objective values are lower which means a bad parameter combination coming from two trials. If the colors are lighter that means the objective values of the 2 trials are both high - good. If the colors are in the middle not lighter and not darker that means one trial has a high objective value and one has a lower objective value. Perhaps a shade of blue to make it darker is in the formula in the calculation of objective color. As the objective value gets smaller, more shade is added.

Crissman:
Thank you @fsmosca for the answer. 🙇  I believe this address this question, so closing this issue. Please feel free to re-open if the answer is not clear.

BenjaminMidtvedt:
To be clear, if two trials have the same x and y values, is the color representative of the average objective value of those trials? A quick glance at the source suggested that the objective value of the last trial is used, but I may be mistaken.

PR #3765, opened 2022-07-07
Refactor importances plot

contramundum53:
## Motivation
For testability, we should separate `_get_importances_info` and `_get_importances_plot`.
See the discussions in #2524.

## Description of the changes
Separate `_get_importances_info` and `_get_importances_plot`. The latter takes a single `_ImportanceInfo` object and returns the plotted figure object.

### DIFF
--- a/optuna/visualization/_param_importances.py
+++ b/optuna/visualization/_param_importances.py
@@ -30,7 +30,6 @@ class _ImportancesInfo(NamedTuple):
     param_names: List[str]
     importance_labels: List[str]
     target_name: str
-    hover_template: List[str]
 
 
 def _get_importances_info(
@@ -53,7 +52,6 @@ def _get_importances_info(
             param_names=[],
             importance_labels=[],
             target_name=target_name,
-            hover_template=[],
         )
 
     importances = optuna.importance.get_param_importances(
@@ -64,17 +62,12 @@ def _get_importances_info(
     importance_values = list(importances.values())
     param_names = list(importances.keys())
     importance_labels = [f"{val:.2f}" if val >= 0.01 else "<0.01" for val in importance_values]
-    hover_template = [
-        _make_hovertext(param_name, importance, study)
-        for param_name, importance in zip(param_names, importance_values)
-    ]
 
     return _ImportancesInfo(
         importance_values=importance_values,
         param_names=param_names,
         importance_labels=importance_labels,
         target_name=target_name,
-        hover_template=hover_template,
     )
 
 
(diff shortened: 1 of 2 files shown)


HideakiImamura:

optuna/visualization/_param_importances.py
```diff
@@ -29,6 +29,8 @@ class _ImportancesInfo(NamedTuple):
     importance_values: List[float]
     param_names: List[str]
     importance_labels: List[str]
+    target_name: str
+    hover_template: List[str]
```
HideakiImamura, line 33: Since this is for plotly, why not calculate it in _get_importances_plot of plotly?

HideakiImamura (commented):
Thanks for the PR> I have a minor comments. PTAL.

nzw0301 (approved):
LGTM. I agree with @HideakiImamura: removing plotly specific logic from `_ImportancesInfo `.

HideakiImamura:
> Calculating hover_template needs the information about study.

Hmm. I see. Then, it is clear for readability to create another function generating `hover_template` like `_get_hover_template`, and to pass `_ImportancesInfo` and `hover_template` to `_get_importaces_plot`. The interface of plotly's `get_importances_plot` and matplotlib's `get_importances_plot` have difference, but it is acceptable, I think.

contramundum53:
Making `hover_template` cannot be done independently (it either has to use part of `_ImportancesInfo` or it has to invoke `ImportanceEvaluator` once more).
How about keeping `distributions` rather than `hover_template` in `_ImportancesInfo`, and use that information only in plotly implementation?

nzw0301:
I agree to https://github.com/optuna/optuna/pull/3765#issuecomment-1178574510 by @HideakiImamura. In my understanding, `_ImportancesInfo` will have minimal common values not depending on the visualisation packages, so separating hover template (or distributions too) that is only for plotly from `_ImportancesInfo` sounds appropriate.

The expected code looks like
```python
    ...
    importances_info = _get_importances_info(study, evaluator, params, target, target_name)
    hover_template = _get_hover_template(importances_info, study)
    return _get_importances_plot(importances_info, hover_template)


def _get_hover_template(importances_info: _ImportancesInfo, study: Study) -> List[str]:
    return [
        _make_hovertext(param_name, importance, study)
        for param_name, importance in zip(importances_info.param_names, importances_info.importance_values)
    ]

```
It looks acceptable to me.

contramundum53:
Sorry for being late. I applied your suggestion. PTAL. @nzw0301

nzw0301:
@contramundum53 Thanks! 

@HideakiImamura could you have any other comments or suggestions? Otherwise, let's merge this PR.

HideakiImamura (approved):
LGTM.

PR #3768, opened 2022-07-07
Raise `ValueError` for invalid returned type of `target` in `_filter_nonfinite`

nzw0301:
## Motivation

In my understanding, the removed part by this PR is not reachable because such case is detected before reaching the removed part as follows:

```python
import optuna


def objective(trial):
    y = trial.suggest_categorical("y", [-1])
    return x ** 2 + y


optuna.visualization.plot_contour(study, params=["x", "y"], target=lambda t: "optuna")
```

The output error is as follows
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [28], in <cell line: 14>()
     11 study = optuna.create_study(sampler=sampler)
     12 study.optimize(objective, n_trials=10)
---> 14 optuna.visualization.plot_contour(study, params=["x", "y"], target=lambda t: "optuna")

File ~/Documents/optuna/optuna/visualization/_contour.py:109, in plot_contour(study, params, target, target_name)
     64 """Plot the parameter relationship as contour plot in a study.
     65 
     66 Note that, if a parameter contains missing values, a trial with missing values is not plotted.
   (...)
    105     A :class:`plotly.graph_objs.Figure` object.
    106 """
    108 _imports.check()
--> 109 info = _get_contour_info(study, params, target, target_name)
    110 return _get_contour_plot(info)

File ~/Documents/optuna/optuna/visualization/_contour.py:249, in _get_contour_info(study, params, target, target_name)
    240 def _get_contour_info(
    241     study: Study,
    242     params: Optional[List[str]] = None,
    243     target: Optional[Callable[[FrozenTrial], float]] = None,
    244     target_name: str = "Objective Value",
    245 ) -> _ContourInfo:
    247     _check_plot_args(study, target, target_name)
--> 249     trials = _filter_nonfinite(
    250         study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,)), target=target
    251     )
    253     all_params = {p_name for t in trials for p_name in t.params.keys()}
    254     if len(trials) == 0:

File ~/Documents/optuna/optuna/visualization/_utils.py:160, in _filter_nonfinite(trials, target, with_message)
    157 filtered_trials: List[FrozenTrial] = []
    158 for trial in trials:
    159     # Not a Number, positive infinity and negative infinity are considered to be non-finite.
--> 160     if not np.isfinite(target(trial)):
    161         if with_message:
    162             _logger.warning(
    163                 f"Trial {trial.number} is omitted in visualization "
    164                 "because its objective value is inf or nan."
    165             )

TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

```

## Description of the changes

Move the unreachable part to `_filter_nonfinite` to provide a better error message.


HideakiImamura:

HideakiImamura:
@toshihikoyanase Let me change the assignment since @contramundum53 is busy now. Could you review this PR?

HideakiImamura (commented):
Agree with change this error handling, but it might be good to move this in `filter_nonfinite` or before it.

nzw0301:
Thanks. I've moved the value error part to `filter_nonfinite` function.

optuna/visualization/_utils.py
```diff
@@ -156,8 +156,15 @@ def _target(t: FrozenTrial) -> float:
 
     filtered_trials: List[FrozenTrial] = []
     for trial in trials:
+        value = target(trial)
+
+        if not isinstance(value, float):
+            raise ValueError(
+                f"Trial{trial.number} has COMPLETE state, but its target value is non-numeric."
```
toshihikoyanase, line 163: `_filter_nonfinite` function does not ensure that trials always have the COMPLETE state. The current implementation filters incomplete trials when calling this function, but we'll need to synchronize the code if we hard-code the state. How about using the trial state as follows?
```suggestion
                f"Trial{trial.number} has {trial.state.name} state, but its target value is non-numeric."
```

tests/visualization_tests/test_parallel_coordinate.py
```diff
@@ -348,13 +348,13 @@ def test_color_map(direction: str) -> None:
         assert not line["reversescale"]
 
     # When `target` is not `None`, `reversescale` is always `True`.
-    line = plot_parallel_coordinate(study, target=lambda t: t.number).data[0]["line"]
+    line = plot_parallel_coordinate(study, target=lambda t: float(t.number)).data[0]["line"]
```
toshihikoyanase, line 351: IMO, it is a bit too strict to reject `int` values although the type of the `target` requires `float`. How about accepting `int` as the original code intended?

c.f., `study.tell` accepts the float-castable values, and I'd like to make them consistent, personally.
https://github.com/optuna/optuna/blob/fc5e262042a60bbcc888541ce7fa5e08769ac478/optuna/study/_tell.py#L65-L71

toshihikoyanase (commented):
I'm sorry for the delayed response. The change basically looks great to  me. I added some small comments.

optuna/visualization/_utils.py
nzw0301: Thanks. I suppose the state information looks unnecessary to show the error message, so I've modified a message by considering https://github.com/optuna/optuna/pull/3768#discussion_r919724822.

tests/visualization_tests/test_parallel_coordinate.py
nzw0301: Agreed. Thank you for sharing a similar code. I've updated the using the same logic in `optuna/optuna/study/_tell.py`.

tests/visualization_tests/test_utils.py
```diff
@@ -173,3 +174,10 @@ def test_filter_inf_trials_message(caplog: LogCaptureFixture, with_message: bool
         assert n_filtered_as_inf == 1
     else:
         assert msg not in caplog.text
+
+
+def test_invalid_target_filter_nonfinite() -> None:
```
HideakiImamura, line 179: This is easier to understand.
```suggestion
def test_filter_nonfinite_with_invalid_target() -> None:
```

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM. I have a minor comment. PTAL.

optuna/visualization/_utils.py
toshihikoyanase: Thanks!

tests/visualization_tests/test_parallel_coordinate.py
toshihikoyanase: Thanks!

HideakiImamura (approved):
LGTM.

PR #3775, opened 2022-07-08
DRY constraints in `Sampler.after_trial`

contramundum53:
## Motivation
There is duplicate logic in `Sampler.after_trial` of samplers that support constraints, i.e., `TPESampler`, `NSGAIISampler`, and `BoTorchSampler`. We extract that logic in `process_constraints_after_trial`.

## Description of the changes
Extract common logic in `process_constraints_after_trial`.

### DIFF
--- a/optuna/samplers/_base.py
+++ b/optuna/samplers/_base.py
@@ -201,9 +201,9 @@ _CONSTRAINTS_KEY = "sampler:constraints"
 
 
 def _process_constraint_after_trial(
-    constraints_func: Callable[[FrozenTrial], Sequence[float]], study: Study, trial: FrozenTrial
+    constraints_func: Callable[[FrozenTrial], Sequence[float]], study: Study, trial: FrozenTrial, state: TrialState
 ) -> None:
-    if trial.state not in [TrialState.COMPLETE, TrialState.PRUNED]:
+    if state not in [TrialState.COMPLETE, TrialState.PRUNED]:
         return
 
     constraints = None
--- a/optuna/samplers/_base.py
+++ b/optuna/samplers/_base.py
@@ -200,7 +200,7 @@ class BaseSampler(object, metaclass=abc.ABCMeta):
 _CONSTRAINTS_KEY = "constraints"
 
 
-def _process_constraint_after_trial(
+def _process_constraints_after_trial(
     constraints_func: Callable[[FrozenTrial], Sequence[float]],
     study: Study,
     trial: FrozenTrial,
--- a/optuna/samplers/_base.py
+++ b/optuna/samplers/_base.py
@@ -197,7 +197,7 @@ class BaseSampler(object, metaclass=abc.ABCMeta):
             )
 
 
-_CONSTRAINTS_KEY = "sampler:constraints"
+_CONSTRAINTS_KEY = "constraints"
 
 
 def _process_constraint_after_trial(
--- a/optuna/samplers/_base.py
+++ b/optuna/samplers/_base.py
@@ -1,13 +1,15 @@
 import abc
-from typing import Any
+from typing import Any, Callable
 from typing import Dict
 from typing import Optional
 from typing import Sequence
+import warnings
 
 from optuna.distributions import BaseDistribution
 from optuna.study import Study
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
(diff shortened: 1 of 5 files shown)


nzw0301:
The tests are supposed to be failed. Could you fix them?

contramundum53:
Sorry, I fixed it. PTAL.

optuna/samplers/_tpe/sampler.py
```diff
@@ -29,7 +31,6 @@
 from optuna.trial import TrialState
 
 
-_CONSTRAINTS_KEY = "tpe:constraints"
```
c-bata, line 32: It looks that there is a breaking change here. What do you think?

contramundum53: Is this considered a breaking change?

contramundum53: One alternative approach is to keep the original `_CONSTRAINTS_KEY`. This would make it impossible to switch sampler for one study.

c-bata:
@keisuke-umezawa  Could you review this PR?

optuna/samplers/_tpe/sampler.py
c-bata: Sorry. This change looks acceptable since the new logic works as expected with storages that stores previous constraints keys.

nzw0301: This feature has been introduced by https://github.com/optuna/optuna/pull/3506 for (only?) about 30 days; we haven't released this as a part of the stable release. So it sounds okay to introduce this backward compatibility for this time because the effect is supposed to be limited.

c-bata: Thank you for your comment!
That sounds reasonable. We may omit data migration for this change.

contramundum53: How about `nsgaii:constraints` and `botorch:constraints`?

c-bata (approved):
LGTM!

optuna/samplers/_tpe/sampler.py
knshnb: We decided that this change is acceptable because both of constrained optimizations of NSGAII and BoTorch are currently experimental.

optuna/samplers/_base.py
```diff
@@ -193,3 +195,33 @@ def _raise_error_if_multi_objective(self, study: Study) -> None:
                 "If the study is being used for multi-objective optimization, "
                 f"{self.__class__.__name__} cannot be used."
             )
+
+
+_CONSTRAINTS_KEY = "sampler:constraints"
```
knshnb, line 200: Now, I feel the meaning of "sampler" is not clear. How about just using "constraints"?

```suggestion
_CONSTRAINTS_KEY = "constraints"
```

knshnb (commented):
Thanks for the PR! It looks almost fine. I left some small comments.

optuna/samplers/_base.py
```diff
@@ -193,3 +195,33 @@ def _raise_error_if_multi_objective(self, study: Study) -> None:
...
+
+
+def _process_constraint_after_trial(
```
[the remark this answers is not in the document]
contramundum53: Thanks for pointing that out. I'll fix it.

optuna/samplers/_base.py
contramundum53: This is because this key is only used in samplers (not pruners or importance evaluators.) What do you think?

contramundum53: I think it can be better to just say `constraints` in case we need the information of constraints in e.g. importance evaluators.

contramundum53: I'll apply your suggestion.

optuna/samplers/_base.py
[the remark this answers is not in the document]
contramundum53: Thanks, I applied it.

contramundum53:
@HideakiImamura @c-bata From the discussion in https://github.com/optuna/optuna/pull/3775#discussion_r920794811, I think naming the key `constraints` is better than `sampler:constraints`. What do you think? (I think this can be very easily changed so this duscussion can be done in subsequent PRs.)

knshnb (approved):
LGTM!

PR #3781, opened 2022-07-12
Use default value in named tuple for empty case

nzw0301:
## Motivation

https://github.com/optuna/optuna/pull/3698/files#r902636883

## Description of the changes

Use default value in named tuple for info classes.

### DIFF
--- a/optuna/visualization/_edf.py
+++ b/optuna/visualization/_edf.py
@@ -29,8 +29,8 @@ class _EDFLineInfo(NamedTuple):
 
 
 class _EDFInfo(NamedTuple):
-    lines: List[_EDFLineInfo]
-    x_values: np.ndarray
+    lines: List[_EDFLineInfo] = []
+    x_values: np.ndarray = np.array([])
 
 
 def plot_edf(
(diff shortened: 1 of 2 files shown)


contramundum53 (approved):
This PR makes the code more concise, thanks!

optuna/visualization/_edf.py
```diff
@@ -29,8 +29,8 @@ class _EDFLineInfo(NamedTuple):
 
 
 class _EDFInfo(NamedTuple):
-    lines: List[_EDFLineInfo]
-    x_values: np.ndarray
+    lines: List[_EDFLineInfo] = []
+    x_values: np.ndarray = np.array([])
```
c-bata, line 33: Not sure but I'm wondering if there is a problem to set mutable object in namedtuple.

nzw0301: Thanks. Indeed, it has the potential to induce a bug when we modify the attribute of the named tuple and reuse it. We don't modify the attribute after creating this info in the current code, so I believe that it doesn't matter for now. But we agree to use immutable objectives is much safer to avoid the unintended bugs.

c-bata: Agree. We may use dataclass or avoid using namedtuple's default value instead.

Issue #3677, opened 2022-06-15
The `optuna.samplers._tpe.sampler._solve_hssp` should consider `+/-inf`


HideakiImamura:
### Expected behavior

The `optuna.samplers._tpe.sampler._solve_hssp` should consider `+/-inf`.
As shown in [here](https://github.com/optuna/optuna/pull/3662/files#r895502801), currently it does not take into account +/- inf, so it behaves unnaturally. Ideally, one would expect the following in a two-dimensional setting.

```
# dimension is 2
test_case = np.asarray([[random.random() for _ in range(2)] for _ in range(8)])
test_case = np.vstack([test_case, [float("inf") for _ in range(2)]])
r = 1.1 * np.max(test_case, axis=0) 
indices = _tpe.sampler._solve_hssp(test_case, np.arange(len(test_case)), 5, r)
approx = _tpe.sampler._compute_hypervolume(test_case[indices], r)
# approx should be 0, but it is inf.

# dimension is 3
test_case = np.asarray([[random.random() for _ in range(3)] for _ in range(8)])
test_case = np.vstack([test_case, [float("inf") for _ in range(3)]])
r = 1.1 * np.max(test_case, axis=0) 
indices = _tpe.sampler._solve_hssp(test_case, np.arange(len(test_case)), 5, r)
approx = _tpe.sampler._compute_hypervolume(test_case[indices], r)
# approx should be 0, but it is nan.
```

The problem may lie in the module that calculates the hypervolume. `optuna.samplers._tpe.sampler.compute_hypervolume` or `optuna._hypervolume`.

### Environment

- Optuna version: current master https://github.com/optuna/optuna/commit/d3c53e053a18341405cdfc9587d4525fdce7494d
- Python version: 3.9
- OS: mac
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
(venv) mamu@HideakinoMacBook-puro 3662 % python piyo.py
/Users/mamu/Dev/pfn/optuna/optuna/_hypervolume/utils.py:14: RuntimeWarning: invalid value encountered in subtract
  return float(np.abs(np.prod(point1 - point2)))
/Users/mamu/Dev/pfn/optuna/optuna/_hypervolume/wfg.py:40: RuntimeWarning: invalid value encountered in double_scalars
  volume -= np.prod(intersection)
/Users/mamu/Dev/pfn/optuna/optuna/_hypervolume/wfg.py:39: RuntimeWarning: invalid value encountered in subtract
[...]

PR #3782, opened 2022-07-12
Support inf values in TPE

### LINKED ISSUE #3676: The `optuna.samplers._tpe.sampler._calculate_weights_below_for_multi_objective` should consider `+/-inf`
### Expected behavior

Should be resolved after #3674.

The `optuna.samplers._tpe.sampler._calculate_weights_below_for_multi_objective` should consider `+/-inf`.

Currently, the `nan`'s list is returned when the trials include `+/-inf` as objective values, but it should return the list of floats. This might be fixed in #3674.

### Environment

- Optuna version: current master [d3c53e0](https://github.com/optuna/optuna/commit/d3c53e053a18341405cdfc9587d4525fdce7494d)
- Python version: 3.9
- OS: mac, linux

```shell
(venv) mamu@HideakinoMacBook-puro 3662 % python fuga.py
/Users/mamu/Dev/pfn/optuna/optuna/_hypervolume/utils.py:14: RuntimeWarning: invalid value encountered in subtract
  return float(np.abs(np.prod(point1 - point2)))
[nan nan nan]
```

### Steps to reproduce

```python
import optuna
import numpy as np


# +/-inf objective values.
weights_below = optuna.samplers._tpe.sampler._calculate_weights_below_for_multi_objective(
    {"x": np.array([1.0, 2.0, 3.0, 4.0], dtype=float)},
    [
        (0, [-float("inf"), -float("inf")]),
        (0, [0.0, -float("inf")]),
        (0, [-float("inf"), 0.0]),
        (0, [float("inf"), float("inf")]),
    ],
    np.array([0, 1, 2]),
    None
)
print(weights_below)

```

_No response_

contramundum53:
## Motivation
Resolve #3676, #3677 
## Description of the changes
We add a function `_clamp_inf_and_calc_reference_point` and use that function in `_calculate_weights_below_for_multi_objective` and `_split_observation_pairs`.
In `_clamp_inf_and_calc_reference_point`, we
* make `reference_point` consider only finite values.
* replace `-inf` to be the minimum of `2 * best_value` and `0.5 * best_value`.
* replace `+inf` to be the `reference_point`, such that `+inf` makes zero contribution to the hypervolume.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -908,9 +908,7 @@ def _clamp_inf_and_calc_reference_point(lvals: np.ndarray) -> Tuple[np.ndarray, 
         ]
     )
     worst_point[worst_point == -np.inf] = 0.0
-    posinf_replacement = np.maximum(
-        np.maximum(1.1 * worst_point, 0.9 * worst_point), worst_point + EPS
-    )
+    posinf_replacement = np.maximum.reduce([1.1 * worst_point, 0.9 * worst_point, worst_point + EPS])
 
     best_point = np.array(
         [
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -773,7 +773,7 @@ def _split_observation_pairs(
         if subset_size > 0:
             rank_i_lvals = lvals[nondomination_ranks == i]
             rank_i_indices = indices[nondomination_ranks == i]
-            rank_i_lvals, reference_point = _normalize_and_calc_reference_point(rank_i_lvals)
+            rank_i_lvals, reference_point = _clamp_inf_and_calc_reference_point(rank_i_lvals)
             selected_indices = _solve_hssp(
                 rank_i_lvals, rank_i_indices, subset_size, reference_point
             )
@@ -872,7 +872,7 @@ def _calculate_weights_below_for_multi_objective(
         weights_below = np.asarray([1.0])
     else:
 
-        lvals, reference_point = _normalize_and_calc_reference_point(lvals)
+        lvals, reference_point = _clamp_inf_and_calc_reference_point(lvals)
 
         hv = _compute_hypervolume(lvals, reference_point)
         indices_mat = ~np.eye(n_below).astype(bool)
@@ -893,7 +893,14 @@ def _calculate_weights_below_for_multi_objective(
     return weights_below_all
 
 
-def _normalize_and_calc_reference_point(lvals: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+def _clamp_inf_and_calc_reference_point(lvals: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+    """Clamp inf values in lvals and calculate the reference point.
+    
+    This function satisfies the following axioms:
+    1. Clamped lvals only contains finite values.
+    2. Clamping preserves the original order of lvals.
+    3. Clamped lvals are always smaller than or equal to the calculated reference point.
+    """
     worst_point = np.array(
         [
             np.max(lvals[:, i][lvals[:, i] != np.inf], initial=-np.inf)
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -887,14 +887,11 @@ def _calculate_weights_below_for_multi_objective(
         )
         neginf_replacement = np.minimum(2 * best_point, 0.5 * best_point)
 
-        print(lvals)
         is_posinf = lvals == np.inf
         lvals[is_posinf] = np.broadcast_to(posinf_replacement, lvals.shape, subok=True)[is_posinf]
         is_neginf = lvals == -np.inf
         lvals[is_neginf] = np.broadcast_to(neginf_replacement, lvals.shape, subok=True)[is_neginf]
 
-        print(is_posinf)
-        print(lvals)
         hv = _compute_hypervolume(lvals, reference_point)
         indices_mat = ~np.eye(n_below).astype(bool)
         contributions = np.asarray(
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -897,7 +897,7 @@ def _clamp_inf_and_calc_reference_point(lvals: np.ndarray) -> Tuple[np.ndarray, 
     """Clamp inf values in lvals and calculate the reference point.
     
     This function satisfies the following axioms:
-    1. Clamped lvals and the calculated reference point only contains finite values.
+    1. Clamped lvals and the calculated reference point only contain finite values.
     2. Clamping preserves the original order of lvals.
     3. Clamped lvals are always smaller than or equal to the calculated reference point.
     """
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -895,7 +895,7 @@ def _calculate_weights_below_for_multi_objective(
 
 def _clamp_inf_and_calc_reference_point(lvals: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
     """Clamp inf values in lvals and calculate the reference point.
-    
+
     This function satisfies the following axioms:
     1. Clamped lvals and the calculated reference point only contain finite values.
     2. Clamping preserves the original order of lvals.
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -902,6 +902,7 @@ def _clamp_inf_and_calc_reference_point(lvals: np.ndarray) -> Tuple[np.ndarray, 
     3. Clamped lvals are always smaller than or equal to the calculated reference point.
     """
     worst_point = np.amax(lvals, axis=0, initial=-np.inf, where=~np.isposinf(lvals))
+    # When there are no finite values for a dimension, the reference point could be any finite value.
     worst_point[worst_point == -np.inf] = 0.0
     posinf_replacement = np.maximum.reduce([1.1 * worst_point, 0.9 * worst_point, worst_point + EPS])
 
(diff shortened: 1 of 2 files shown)


tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
```diff
@@ -575,33 +618,37 @@ def test_solve_hssp(dim: int) -> None:
...
-    test_case = np.vstack([test_case, [-float("inf") for _ in range(2)]])
-    truth, approx = _compute_hssp_truth_and_approx(test_case, subset_size)
-    assert np.isinf(truth)
-    assert np.isinf(approx)
-
-    test_case = np.asarray([[random.random() for _ in range(3)] for _ in range(8)])
-    test_case = np.vstack([test_case, [-float("inf") for _ in range(3)]])
-    truth, approx = _compute_hssp_truth_and_approx(test_case, subset_size)
-    assert np.isinf(truth)
-    assert np.isinf(approx)
```
contramundum53, line 604: We don't need these tests because `_solve_hssp` never gets `inf` arguments.

c-bata:
@HideakiImamura @toshihikoyanase  Could you review this PR if you have time?

optuna/samplers/_tpe/sampler.py
```diff
@@ -893,3 +891,41 @@ def _calculate_weights_below_for_multi_objective(
     weights_below_all[feasible_mask] = weights_below
     weights_below_all = weights_below_all[~np.isnan(cvals)]
     return weights_below_all
+
+
+def _clamp_inf_and_calc_reference_point(lvals: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
```
HideakiImamura, line 896: I'm not sure but "clip" instead of "clamp" is true?

optuna/samplers/_tpe/sampler.py
```diff
@@ -893,3 +891,41 @@ def _calculate_weights_below_for_multi_objective(
...
+    worst_point = np.array(
+        [
+            np.max(lvals[:, i][lvals[:, i] != np.inf], initial=-np.inf)
+            for i in range(lvals.shape[1])
+        ]
+    )
+    worst_point[worst_point == -np.inf] = 0.0
+    posinf_replacement = np.maximum(
+        np.maximum(1.1 * worst_point, 0.9 * worst_point), worst_point + EPS
+    )
```
HideakiImamura, line 913: A nit.
```suggestion
    posinf_replacement = np.maximum.reduce([1.1 * worst_point, 0.9 * worst_point, worst_point + EPS])
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -893,3 +891,41 @@ def _calculate_weights_below_for_multi_objective(
...
+
+    best_point = np.array(
+        [
+            np.min(lvals[:, i][lvals[:, i] != -np.inf], initial=np.inf)
+            for i in range(lvals.shape[1])
+        ]
+    )
+    best_point[best_point == np.inf] = 0.0
+    neginf_replacement = np.minimum(np.minimum(2 * best_point, 0.5 * best_point), best_point - EPS)
```
HideakiImamura, line 922: Ditto.
```suggestion
    neginf_replacement = np.minimum.reduce([2 * best_point, 0.5 * best_point, best_point - EPS])
```

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
```diff
@@ -515,7 +515,7 @@ def test_calculate_weights_below_for_multi_objective() -> None:
     assert np.isclose(weights_below[1], weights_below[2])
     assert sum(weights_below) > 0
 
-    # +/-inf objective values.
+    # Test -inf objective values.
```
HideakiImamura, line 518: The test includes `+inf`. Should we remove `[float("inf"), float("inf")]` to make it consistent with the comment?

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM. I have several minor comments. PTAL.

optuna/samplers/_tpe/sampler.py
toshihikoyanase: +1 with @HideakiImamura  since NumPy uses [`clip`](https://numpy.org/doc/stable/reference/generated/numpy.clip.html).

optuna/samplers/_tpe/sampler.py
```diff
@@ -893,3 +891,41 @@ def _calculate_weights_below_for_multi_objective(
...
+    1. Clamped lvals and the calculated reference point only contain finite values.
+    2. Clamping preserves the original order of lvals.
+    3. Clamped lvals are always smaller than or equal to the calculated reference point.
+    """
+    worst_point = np.array(
+        [
+            np.max(lvals[:, i][lvals[:, i] != np.inf], initial=-np.inf)
+            for i in range(lvals.shape[1])
+        ]
+    )
```
toshihikoyanase, line 909: A nit, but can we remove the for loop?
```suggestion
    worst_point = np.amax(lvals, axis=0, initial=-np.inf, where=~np.isposinf(lvals))
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -893,3 +891,41 @@ def _calculate_weights_below_for_multi_objective(
...
+    worst_point[worst_point == -np.inf] = 0.0
```
toshihikoyanase, line 910: Could you add a code comment to explain why this hack is needed?
I guess `worst_point` element can be `-np.inf` if all values are `-np.inf`, but do we have any other cases?

optuna/samplers/_tpe/sampler.py
toshihikoyanase: I want a code comment here as well. I guess the first element is for the positive `worst_point` values, and the second one is for negative ones, and third one is for zeros.

optuna/samplers/_tpe/sampler.py
```diff
@@ -893,3 +891,41 @@ def _calculate_weights_below_for_multi_objective(
...
+    posinf_replacement = np.maximum(
+        np.maximum(1.1 * worst_point, 0.9 * worst_point), worst_point + EPS
+    )
+
+    best_point = np.array(
+        [
+            np.min(lvals[:, i][lvals[:, i] != -np.inf], initial=np.inf)
+            for i in range(lvals.shape[1])
+        ]
+    )
```
toshihikoyanase, line 918: ditto.
```suggestion
    best_point = np.amin(lvals, axis=0, initial=np.inf, where=~np.isneginf(lvals))
```

toshihikoyanase (commented):
Let me share some comments on  `_clamp_inf_and_calc_reference_point`.

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
contramundum53: Sorry, I'll update the comments.
```suggestion
    # Test +/-inf objective values.
```

optuna/samplers/_tpe/sampler.py
contramundum53: Yes, `worst_point == -np.inf` happens whenever there are no finite values for a dimension. In that case, the reference point could be any value.
```suggestion
    # When there are no finite values for a dimension, the reference point could be any finite value.
    worst_point[worst_point == -np.inf] = 0.0
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -893,3 +891,30 @@ def _calculate_weights_below_for_multi_objective(
...
+
+    This function satisfies the following axioms:
+    1. Clamped lvals and the calculated reference point only contain finite values.
+    2. Clamping preserves the original order of lvals.
+    3. Clamped lvals are always smaller than or equal to the calculated reference point.
+    """
+    worst_point = np.amax(lvals, axis=0, initial=-np.inf, where=~np.isposinf(lvals))
+    # When there are no finite values for a dimension, the reference point could be any finite value.
+    worst_point[worst_point == -np.inf] = 0.0
+    posinf_replacement = np.maximum.reduce([1.1 * worst_point, 0.9 * worst_point, worst_point + EPS])
```
contramundum53, line 907:
```suggestion
    # For positive worst_point, we use 1.1 * worst_point as the reference point.
    # For negative worst_point, we use 0.9 * worst_point as the reference point.
    # For worst point that is almost zero, we use worst_point + EPS as the reference point.
    posinf_replacement = np.maximum.reduce([1.1 * worst_point, 0.9 * worst_point, worst_point + EPS])
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -893,3 +891,30 @@ def _calculate_weights_below_for_multi_objective(
...
+
+
+def _clamp_inf_and_calc_reference_point(lvals: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+    """Clamp inf values in lvals and calculate the reference point.
+
+    This function satisfies the following axioms:
+    1. Clamped lvals and the calculated reference point only contain finite values.
+    2. Clamping preserves the original order of lvals.
+    3. Clamped lvals are always smaller than or equal to the calculated reference point.
+    """
```
toshihikoyanase, line 903: Could you add explanation that the clipped values do not affect the TPE algorithm? In my understanding, single-objective TPE only consider the order of objective values. Is the situation the same for the multi-objective case?

toshihikoyanase: Does this clipping algorithm keeps the expected hypervolume improvement?
IMO, the order of the trials can be changed depending on the scale of each parameter.

For example, when we have three points such as (`-inf`, -900), (-100, -1000) and (-90, -1200) in a target rank and a reference point (0, 0), the clipped points will be (-110, -900), (-100, -1000) and (-90, -1200).

The changes of the hypervolumes will be as follows:
Before clipping: `inf` (`inf` * 900), 1000 (10 * 100), 18000 (90 * 200)
After clipping: 9000 (10 * 900), 1000 (10 * 100), 18000 (90 * 200)

So, the rank of first point and third point will be changed:
Before clipping: 0, 2, 1
After clipping: 1, 2, 0

contramundum53: Yes, it can change the order of the hypervolume improvement. This is intended, because we don't want all trials containing at least one `inf` to all have the same hypervolume improvement.

contramundum53: This should not be a breaking change, because having `inf` as hypervolumes would result in `nan`.

optuna/samplers/_tpe/sampler.py
contramundum53: Thanks, fixed.

optuna/samplers/_tpe/sampler.py
contramundum53: > In my understanding, single-objective TPE only consider the order of objective values. Is the situation the same for the multi-objective case?
No, as for multi-objective case, the weights depend on the hypervolume improvements. Thus, the clipped values DO affect the TPE algorithm.

PR #3784, opened 2022-07-12
Unify slice plot tests

nzw0301:
## Motivation

Same as other visualization test unification such as https://github.com/optuna/optuna/pull/3701.

## Description of the changes

- [x] Remove matplot version
- [x] Minor fix in the plot function
- [x] Unify the test logic by following #3701 as much as possible.


tests/visualization_tests/test_contour.py
```diff
@@ -516,7 +516,6 @@ def test_get_contour_info_nonfinite_multiobjective(objective: int, value: float)
         reverse_scale=True,
         target_name="Objective Value",
     )
-    assert info.sorted_params == ["param_b", "param_d"]
```
nzw0301, line 519: This is checked in `assert info == _ContourInfo(...` above too, so we can remove this.

tests/visualization_tests/test_slice.py
```diff
@@ -1,99 +1,373 @@
...
+    "params",
+    [
+        ["param_a"],
+        ["param_a", "param_b"],
+        ["param_a", "param_b", "param_c"],
+        ["param_a", "param_b", "param_c", "param_d"],
+        None,
+    ],
+)
+def test_get_slice_plot_info_params(params: Optional[List[str]]) -> None:
```
nzw0301, line 187: This is the counterpart of 
- `test_get_contour_info_too_short_params`
- `test_get_contour_info_2_params`
- `test_get_contour_info_more_than_2_params`

in `test_contour.py`. Unlike `plot_contour`, plot slice's info can handle only one param case and the shape of sub info linearly depends on the size of params, so I unify them.

nzw0301: For the same reason, this PR doesn't have the conterpart of `test_get_contour_info_log_scale_and_str_category_more_than_2_params`.

c-bata:

tests/visualization_tests/test_slice.py
```diff
@@ -1,99 +1,364 @@
...
+            _SliceSubplotInfo(
+                param_name="param_d",
+                x=[4.0, 2.0],
+                y=[2.0, 1.0],
+                trial_numbers=[1, 2],
+                is_log=False,
+                is_numerical=True,
+            ),
+        ],
+    )
```
contramundum53, line 354: I don't think we need this test, since this seems not to be increasing the coverage. 
```suggestion
```

tests/visualization_tests/test_slice.py
```diff
@@ -1,99 +1,364 @@
...
+                is_log=False,
+                is_numerical=False,
+            ),
+        ],
+    )
+
+
+def test_get_slice_plot_info_mixture_category_types() -> None:
+
+    study = _create_study_mixture_category_types()
```
contramundum53, line 268: How about simply inlining `_create_study_mixture_category_types` function?

tests/visualization_tests/test_slice.py
```diff
@@ -1,99 +1,364 @@
...
+                is_log=False,
+                is_numerical=True,
+            ),
+        ],
+    )
+
+
+def test_get_slice_plot_info_log_scale_and_str_category_2_params() -> None:
+
+    study = _create_study_with_log_scale_and_str_category_2d()
```
contramundum53, line 241: How about simply inlining `_create_study_with_log_scale_and_str_category_2d` function?

tests/visualization_tests/test_slice.py
```diff
@@ -1,99 +1,364 @@
...
+
+
+@pytest.mark.parametrize(
+    "params",
+    [
+        ["param_a", "param_b"],  # First column has 1 observation.
+        ["param_b", "param_a"],  # Second column has 1 observation
+    ],
+)
+def test_generate_contour_plot_for_few_observations(params: List[str]) -> None:
```
contramundum53, line 211:
```suggestion
def test_get_slice_plot_info(params: List[str]) -> None:
```
Typo

tests/visualization_tests/test_slice.py
```diff
@@ -1,99 +1,364 @@
...
+    [
+        ["param_a"],
+        ["param_a", "param_b"],
+        ["param_a", "param_b", "param_c"],
+        ["param_a", "param_b", "param_c", "param_d"],
+        None,
+    ],
+)
+def test_get_slice_plot_info_params(params: Optional[List[str]]) -> None:
+
```
contramundum53, line 179: I don't think we need so many test cases because there is no `if` statements that changes behavior based on the number of params. Actually the case where `params is not None` has already been tested in `test_generate_contour_plot_for_few_observations`, so we only need to test the case where params is default.

tests/visualization_tests/test_slice.py
```diff
@@ -1,99 +1,364 @@
...
+@pytest.mark.parametrize(
+    "params",
+    [
+        ["param_a", "param_b"],
+        ["param_a", "param_b", "param_c"],
+    ],
+)
+def test_get_slice_plot_info_customized_target(params: List[str]) -> None:
+
```
contramundum53, line 195:
```suggestion
def test_get_slice_plot_info_customized_target() -> None:
    params = ["param_a", "param_b"]
```
We don't need this parametrization.

tests/visualization_tests/test_slice.py
```diff
@@ -1,99 +1,364 @@
...
+    study = prepare_study_with_trials()
+    info = _get_slice_plot_info(
+        study, params=params, target=lambda t: t.params["param_d"], target_name="Objective Value"
+    )
+    n_params = len(params)
+    assert len(info.subplots) == n_params
```
contramundum53, line 201: Can we assert the content of `info` rather than just the size of `info.subplots`?

tests/visualization_tests/test_slice.py
```diff
@@ -1,99 +1,364 @@
...
+        [prepare_study_with_trials, []],
+        [prepare_study_with_trials, ["param_a"]],
+        [prepare_study_with_trials, ["param_a", "param_b"]],
+        [prepare_study_with_trials, ["param_a", "param_b", "param_c"]],
+        [prepare_study_with_trials, ["param_a", "param_b", "param_c", "param_d"]],
+        [prepare_study_with_trials, None],
+        [_create_study_with_log_scale_and_str_category_2d, None],
+        [_create_study_mixture_category_types, None],
+    ],
+)
```
contramundum53, line 115: We don't need so many test cases.
* `_get_slice_plot` function only distinguishes the cases of zero param, one param or `>=2` params, so we don't need the tests for three or more params.
* Both `_get_slice_plot_info(create_study())` and `_get_slice_plot_info(_create_study_with_failed_trial())` generate the same empty `_SlicePlotInfo`, so assuming that `_get_slice_plot_info` is well-tested, we only need to include one of them.
* Specifying `None` to `params` generates the same `_SlicePlotInfo` as `["param_a", ..., "param_d"]`, so assuming we have a correct `_get_slice_plot_info`, we don't need that test either.

contramundum53:
I think `plot_slice` is much simpler than `plot_contour` and thus we could make `test_slice.py` much simpler than `test_contour.py`.

tests/visualization_tests/test_slice.py
nzw0301: The other test files for the plot methods have the same test. Should we remove all of them?

tests/visualization_tests/test_slice.py
nzw0301: In fact, the slice plot code has `if` a block depending on the number of subplots, which depends on the number of params: https://github.com/optuna/optuna/blob/946b692c02aeaf17eba4b0966afdf500b9797dc7/optuna/visualization/_slice.py#L163-L165.

tests/visualization_tests/test_slice.py
nzw0301: `test_plot_slice` uses `_create_study_mixture_category_types`.

tests/visualization_tests/test_slice.py
nzw0301: `test_plot_slice` uses `_create_study_mixture_category_types`.

tests/visualization_tests/test_slice.py
nzw0301: Thanks. I've removed several test cases by following suggestions. Honestly, I'm not sure calling private two times makes this test simpler. Could you share examples?

tests/visualization_tests/test_slice.py
nzw0301: Sure.

tests/visualization_tests/test_slice.py
HideakiImamura: I think this test is necessary because it is not for coverage, but to verify behavior.

HideakiImamura (commented):
Thanks for the update. Basically, LGTM. I have several comments. PTAL.

tests/visualization_tests/test_slice.py
contramundum53: I think we have already tested the behavior of `inf` values in the single-objective case, and we use the same code for multi-objective case, so we don't need to check the multi-objective case again.

tests/visualization_tests/test_slice.py
contramundum53: Sorry, I missed that :bow:

tests/visualization_tests/test_slice.py
contramundum53: Sorry, I missed that again :bow:

tests/visualization_tests/test_slice.py
contramundum53: But this is a test for `_get_slice_plot_info`, not `_get_slice_plot`.
(I mean, this is the whole point of splitting the tests for `_get_slice_plot_info` and `_get_slice_plot`.)

tests/visualization_tests/test_slice.py
contramundum53: https://github.com/optuna/optuna/pull/3788/files#diff-c6413b01b35fb8fc6ac600bd332caa5482dde4702d3d08ef95b67faa35a3ac93R142-R195
I think these three cases are all we need to cover all branches of the function.

tests/visualization_tests/test_slice.py
```diff
@@ -1,99 +1,372 @@
...
+def test_get_slice_plot_info_params(params: Optional[List[str]]) -> None:
 
-def test_plot_slice_log_scale() -> None:
+    study = prepare_study_with_trials()
+    n_params = len(params) if params is not None else 4
 
-    study = create_study()
+    info = _get_slice_plot_info(study, params=params, target=None, target_name="Objective Value")
+    assert len(info.subplots) == n_params
```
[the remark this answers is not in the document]
nzw0301: Sure.

tests/visualization_tests/test_slice.py
nzw0301: Ah, right. 
@HideakiImamura what do you think about this? Should I keep only a few cases such as `params=None` or `params=["param_a"]`.

tests/visualization_tests/test_slice.py
contramundum53: OK, I think the current test is simple enough. Thanks.

tests/visualization_tests/test_slice.py
contramundum53: @HideakiImamura What do you think?

nzw0301:
Hi, I'm afraid I will not be able to work on this PR (more specifically, the whole Optuna development) until the end of this month at least. So please feel free to take over this PR if my response is not fast enough for the optuna v3 release schedule).

tests/visualization_tests/test_slice.py
HideakiImamura: Thanks for the comment, and sorry for the late reply. We have this type of tests in other visualization files, so I think we should remove it together if we decide to remove them. So, I think it could be outside the scope of this PR.

tests/visualization_tests/test_slice.py
HideakiImamura: This is a point where the discussion could get heated. A similar issue has been pointed out by @contramundum53  in PR #3800, but as pointed out by @g-votte , I think this PR could be merged, with the intention of putting it together later to be incorporated into the revision.

HideakiImamura (approved):
LGTM.

contramundum53 (approved):
Sorry for late reply. LGTM!

PR #3785, opened 2022-07-13
DRY importance tests

contramundum53:
## Motivation
Currently there are duplicate test code in `importance_tests/fanova.py` and `importance_tests/mean_decrease_impurity.py`. We refactor these tests.

## Description of the changes
DRY `test_importance_evaluator_seed` and `test_importance_evaluator_with_target`.


tests/importance_tests/test_init.py
```diff
@@ -267,3 +269,51 @@ def objective(trial: Trial) -> float:
...
+    x2 = trial.suggest_float("x2", 0.1, 3, log=True)
+    x3 = trial.suggest_float("x3", 2, 4, log=True)
+    return x1 + x2 * x3
+
+
+def multi_objective_function(trial: Trial) -> Tuple[float, float]:
+    x1 = trial.suggest_float("x1", 0.1, 3)
+    x2 = trial.suggest_float("x2", 0.1, 3, log=True)
+    x3 = trial.suggest_float("x3", 2, 4, log=True)
+    return x1, x2 * x3
```
c-bata, line 285: Is this objective function used?

contramundum53: Seems not. I'll remove it.

tests/importance_tests/test_init.py
```diff
@@ -267,3 +269,51 @@ def objective(trial: Trial) -> float:
...
+
+
```
contramundum53, line 287:
```suggestion
```
Remove unnecessary function.

c-bata:

tests/importance_tests/test_init.py
```diff
@@ -267,3 +269,44 @@ def objective(trial: Trial) -> float:
     assert all([param in param_importance for param in ["x", "y"]])
     assert param_importance["x"] == 1.0
     assert param_importance["y"] == 0.0
+
+
+def objective(trial: Trial) -> float:
+    x1 = trial.suggest_float("x1", 0.1, 3)
+    x2 = trial.suggest_float("x2", 0.1, 3, log=True)
+    x3 = trial.suggest_float("x3", 2, 4, log=True)
+    return x1 + x2 * x3
```
c-bata, line 278: How about renaming this function since it conflicts with inner function of `test_get_param_importances` and `test_get_param_importance_target_is_none_and_study_is_multi_obj`?

c-bata: Hmm... Good question 😄
I have no idea. We may define this function as an inner function like other test cases since these test cases checks different behavior.

c-bata (approved):
LGTM!

not522 (commented):
How about adding `ShapleyImportanceEvaluator` to `parametrize_evaluator`?

contramundum53:
To do that, we need to add `shap` to the dependency. Considering that there are some difficulty installing `shap` to some environments including M1 Mac, I think we should separate the test for `ShapleyImportanceEvaluator`.

not522:
It is similar to `test_samplers.py`. I think that we can add tests only when `shap` is installed using `optuna.integration.shap._imports.is_successful()`.

contramundum53:
Wouldn't that always be skipped in CI?

not522:
The dependencies are installed here. https://github.com/optuna/optuna/blob/124b8b1cde05700e56706e6716e3926d69907103/.github/workflows/tests.yml#L64-L67

contramundum53:
Thanks. I fixed it.

contramundum53:
Notes: Some tests assume that the sum of the importance values are normalized to 1. However, this is not a natural choice for `ShapleyImportanceEvaluator`. We instead replaced the sum tests with an assertion that the importance values are non-negative finite values.

not522 (approved):
LGTM. I think the sanity check should be divided per evaluator, but we can do it in a follow-up PR.

PR #3793, opened 2022-07-19
Skip tests if required libraries are not installed

not522:
## Motivation
`samplers_tests/test_samplers.py` and `importance_tests/test_init.py` require integration libraries. This PR skip tests if the libraries are not installed. It makes running tests easier in local environments.

## Description of the changes
- Add `pytest.skipif` for integration samplers and an importance evaluator.

### DIFF
--- a/tests/importance_tests/test_init.py
+++ b/tests/importance_tests/test_init.py
@@ -21,10 +21,16 @@ from optuna.testing.storages import StorageSupplier
 from optuna.trial import Trial
 
 
-evaluators = [MeanDecreaseImpurityImportanceEvaluator, FanovaImportanceEvaluator]
-
-if optuna.integration.shap._imports.is_successful():
-    evaluators += [optuna.integration.shap.ShapleyImportanceEvaluator]
+evaluators = [
+    MeanDecreaseImpurityImportanceEvaluator,
+    FanovaImportanceEvaluator,
+    pytest.param(
+        optuna.integration.shap.ShapleyImportanceEvaluator,
+        marks=pytest.mark.skipif(
+            not optuna.integration.shap._imports.is_successful(), reason="shap is not installed."
+        ),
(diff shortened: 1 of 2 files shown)


contramundum53:
@HideakiImamura @knshnb Could you review this PR?

HideakiImamura (approved):
LGTM.

not522:
I found an alternative way. Which do you prefer? https://github.com/optuna/optuna/compare/master...not522:optuna:skipif_integration_alternative

knshnb:
Thanks for the PR. The change looks great overall, but I'm afraid that, in the current PR, we might miss some import errors when we fail to build the CI environment correctly for some reason.

What do you think about the following strategies? (Cc: @HideakiImamura)
- Add import tests that run only on CI
- Do not skip tests on CI

HideakiImamura:
Sorry for the late comment. 

> What do you think about the following strategies?

I think the first one is fine. "Add separate import tests that run only on CI". It reduces the difficulty in local environments and keeps the robustness in CI.

HideakiImamura:
Let me confirm that the current situation of this PR. Is this ready for review, or do we need to decide something?

c-bata:
Let me reassign reviewer since Imamura-san will take a vacation.

not522:
I discussed this PR with @knshnb. We concluded that it would be better to make `test_samplers.py` work without integration, not skipping tests.

I. e., we will remove this part

https://github.com/optuna/optuna/blob/79f0329a3d41224b7f2f7a5620aa477e2130a9d9/.github/workflows/tests.yml#L64-L74

and the command `pytest tests/samplers_tests/test_samplers.py --integration` tests integration samplers.

Let me close this PR. I will create a new PR to resolve this issue.
@nzw0301 Sorry for the confusion. 🙇

PR #3803, opened 2022-07-21
Add note for deprecation of `plot_pareto_front`'s `axis_order`

not522:
## Motivation
`plot_pareto_front`'s `axis_order` is deprecated, but the document does not mention it.

## Description of the changes
Add note in the document.


contramundum53 (approved):
LGTM!

Issue #3724, opened 2022-06-26
Use test stubs of samplers where we can in unit tests using samplers


keisuke-umezawa:
### Motivation

To make unit tests using samplers deterministic , we would like to use test stubs instead of instances which have non-deterministic behaviors. In this issue, we would like to use test stubs for samplers (e.g. DeterministicRelativeSampler, FirstTrialOnlyRandomSampler) where we can so that we make unit tests using samplers deterministic.

Furthermore, to make unit tests using samplers deterministic, there are 5 ways used in Optuna. e.g.

- Use a fixed random seed to make RandomSampler deterministic
- Use a mock to fix suggested values
- Use a test stub to fix suggested values
- Use a single value to fix suggested values
- Test suggested values are in the expected range. e.g. 1 <= x <=2

We will align a way to write unit tests using samplers if needed.

#### Current direction
1. If we do not need to check suggested values in a unit test,
    1. we should `use a fixed random seed to make RandomSampler deterministic`.
1. If we need to check suggested values in a unit test,
    1. we should choose one of ways from the following:
        - Use a mock to fix suggested values
        - Use a test stub to fix suggested values
        - Use a single value to fix suggested values
        - Test suggested values are in the expected range. e.g. 1 <= x <=2
    1. And, the current direction is that `using a test stub to fix suggested values`.

However, we have not concluded that `using a test stub to fix suggested values` is always the best way when we need to check suggested values in a unit test, especially when `using a single value to fix suggested values` and `testing suggested values are in the expected range`. We will discuss it in this issue.

Note: `If the current test stubs of samplers are not enough to reproduce the current functions of the unit tests, we may introduce new test stubs for samplers.`

#### Todos
- [ ] Implement new test stubs if needed
    - [x] DeterministicSampler #3827 
- [ ] Make a consensus that which solution is appropriate when we need to check suggested values in a unit test
- [ ] Use test stubs for samplers in the following files:
    - [x] Unit tests `using a mock to fix suggested values` #3807
        - [x] https://github.com/optuna/optuna/blob/master/tests/trial_tests/test_trial.py
            - test_suggest_uniform
            - test_suggest_loguniform
            - test_suggest_discrete_uniform
            - test_suggest_int
            - test_suggest_int_log
    - [ ] Unit tests `using a single value to fix suggested values`
        - [ ] https://github.com/optuna/optuna/blob/master/tests/study_tests/test_study.py
            - test_callbacks
[...]

PR #3805, opened 2022-07-24
Implement FixedSeedRandomSampler and use it in test_trial

keisuke-umezawa:
## Motivation

For https://github.com/optuna/optuna/issues/3724, it partially solved the issue for test_trial.

## Description of the changes

- Implemented FixedSeedRandomSampler
- Used FixedSeedRandomSampler in test_trial

### DIFF
--- a/optuna/testing/samplers.py
+++ b/optuna/testing/samplers.py
@@ -76,3 +76,9 @@ class FirstTrialOnlyRandomSampler(optuna.samplers.RandomSampler):
         return super(FirstTrialOnlyRandomSampler, self).sample_independent(
             study, trial, param_name, param_distribution
         )
+
+
+class FixedSeedRandomSampler(optuna.samplers.RandomSampler):
+    def __init__(self) -> None:
+        seed = 42
+        super().__init__(seed)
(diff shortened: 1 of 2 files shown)


contramundum53:
Should we also use `FixedSeedRandomSampler` for the places where the default sampler (TPE sampler with no seed) is used?

contramundum53:
Do we really need a `FixedSeedRandomSampler`? How about directly specifying `seed=42` to `RandomSampler` where it is needed?

optuna/testing/samplers.py
```diff
@@ -76,3 +76,9 @@ def sample_independent(
         return super(FirstTrialOnlyRandomSampler, self).sample_independent(
             study, trial, param_name, param_distribution
         )
+
+
+class FixedSeedRandomSampler(optuna.samplers.RandomSampler):
+    def __init__(self) -> None:
+        seed = 42
```
g-votte, line 83: We may want to avoid hard-coding a magic number in the logic. How about extracting this as an argument or somewhere else?

toshihikoyanase:
@keisuke-umezawa will revise the PR according to the off-line discussion. I'll assign reviewers after that.

Issue #3815, opened 2022-07-27
Reduce warnings in unit tests


not522:
### Motivation

The Optuna's unit tests raise many warnings, which make it hard to recognize unexpected warning messages.

### Description

We can filter expected warnings using `@pytest.mark.filterwarnings` or `warnings.simplefilter("ignore")`. Some warnings may be unexpected or unsuitable. We should fix tests or Optuna's codes in such cases.

This issue is contribution-welcome. We welcome any fixes, even how trivial.

jmsykes83:
I will work on this.

nzw0301:
Hi @jmsykes83 thank you for letting us know your interest! I'm also working on this prioritised issue too since this issue affects many developers' experiences. Do you have specific modules you would like to resolve this issue? I'd like to avoid conflicts with your forthcoming pull request.

jmsykes83:
I will let you take the lead on this one.  Was there any direction or ideas you had in approaching this.

nzw0301:
Thank you for your reply. How about some tests in `tests/visualization_tests`, maybe `tests/visualization_tests/test_utils.py` is the easiest one? The expected change will be similar to changes introduced by the linked PRs with this PR such as adding `@pytest.mark.filterwarnings` decorator to a test file.

jmsykes83:
I know I can do this:
`
@pytest.mark.filterwarnings("ignore:Trial0")
def test_filter_nonfinite_with_invalid_target() -> None:
    study = prepare_study_with_trials()
    trials = study.get_trials(states=(TrialState.COMPLETE,))
    with pytest.raises(ValueError):
        _filter_nonfinite(trials, target=lambda t: "invalid target")  # type: ignore
`
But is this the solution you're looking for.  This seems like a solution that could become very cluttered in the long run.  This filters the only warning I am getting for pytest tests/visualization_tests/test_utils.py

nzw0301:
In my understanding, the expected change is something like this. If we use stronger `filterwarnings`, we might miss unintended warning messages, so such a minimal filter looks enough.

jmsykes83:
Did we need to filter the `user warnings` for `test_parallel_coordinate.py` as well?  If so, all of them or just the ones with the `  Attempting to set identical bottom == top == 29.999999999999996 results in singular transformations; automatically expanding?`

nzw0301:
[...]

PR #3818, opened 2022-07-27
Fix return value's type of `_get_node_value`

not522:
## Motivation
Part of #3815.
`self._tree.value[node_index]` is `numpy.ndarray` of size 1.

## Description of the changes
Cast the return value to `float` from `numpy.ndarray`.

### DIFF
--- a/optuna/importance/_fanova/_tree.py
+++ b/optuna/importance/_fanova/_tree.py
@@ -255,7 +255,7 @@ class _FanovaTree:
         return self._get_node_left_child(node_index), self._get_node_right_child(node_index)
 
     def _get_node_value(self, node_index: int) -> float:
-        return float(self._tree.value[node_index])
+        return self._tree.value[node_index].item()
 
     def _get_node_split_threshold(self, node_index: int) -> float:
         return self._tree.threshold[node_index]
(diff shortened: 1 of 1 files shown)


optuna/importance/_fanova/_tree.py
```diff
@@ -255,7 +255,7 @@ def _get_node_children(self, node_index: int) -> Tuple[int, int]:
         return self._get_node_left_child(node_index), self._get_node_right_child(node_index)
 
     def _get_node_value(self, node_index: int) -> float:
-        return self._tree.value[node_index]
+        return float(self._tree.value[node_index])
```
toshihikoyanase, line 258: I guess we can also extract a float value from `numpy.ndarray` of size 1 with `.item()`, but it can return the other type like `int` depending on the content. So, the current implementation seems reasonable to me. What do you think?

```suggestion
        return self._tree.value[node_index].item()
```

not522: Your suggestion depends on the `scikit-learn` implementation, but it looks good to me because it returns a float.

not522: I rethink your suggestion and revert the change because it's too implementation dependent, which assumes `self._tree.value`'s dimension is 2 or more.

toshihikoyanase: Thank you for your investigation. Sounds reasonable to me.

toshihikoyanase (approved):
LGTM!

toshihikoyanase:
Let me merge this PR with a single approval since the fix is obvious.

PR #3814, opened 2022-07-27
Raise `ValueError` when waiting trial is told

not522:
## Motivation
The told trial should be `RUNNING`, not `WAITING`.

## Description of the changes
Fix `_tell_with_warning` and add a test.

PR #3819, opened 2022-07-27
Fix `Study.tell` with invalid values

not522:
## Motivation
Fix unexpected behavior of `Study.tell` with invalid values and `state=TrialState.COMPLETE`.

```python
import optuna

study = optuna.create_study()

trial = study.ask()
study.tell(trial, values="invalid", state=TrialState.COMPLETE)
print(study.trials[-1].state, study.trials[-1].value)
```

* master
```
TrialState.COMPLETE None
```

* PR
```
ValueError: The number of the values 7 did not match the number of the objectives 1.
```

## Description of the changes
Fix `Study.tell` and add tests.

### DIFF
--- a/optuna/study/_tell.py
+++ b/optuna/study/_tell.py
@@ -172,6 +172,9 @@ def _tell_with_warning(
         len(study.directions), values, trial_number
     )
 
+    if state == TrialState.COMPLETE and values_conversion_failure_message is not None:
+        raise ValueError(values_conversion_failure_message)
+
     if state is None:
         if values_conversion_failure_message is None:
             state = TrialState.COMPLETE
--- a/tests/study_tests/test_study.py
+++ b/tests/study_tests/test_study.py
@@ -1417,6 +1417,22 @@ def test_tell_invalid() -> None:
     with pytest.raises(ValueError):
         study.tell(study.ask(), state=TrialState.COMPLETE)
 
+    # Invalid values for completions.
+    with pytest.raises(ValueError):
+        study.tell(study.ask(), "a", state=TrialState.COMPLETE)  # type: ignore
+
+    with pytest.raises(ValueError):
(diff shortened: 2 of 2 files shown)


toshihikoyanase:
@knshnb This PR also fix `Study.tell` similarly to #3814. Could you review this PR, please?

contramundum53 (approved):
LGTM!

optuna/study/_tell.py
```diff
@@ -172,6 +172,9 @@ def _tell_with_warning(
         len(study.directions), values, trial_number
     )
 
+    if state == TrialState.COMPLETE and values_conversion_failure_message is not None:
+        raise ValueError(values_conversion_failure_message)
+
     if state is None:
```
toshihikoyanase, line 178: Not a strong opinion, but how about connecting this if statement with the next one?
The latter one also checks the `state` and `value_conversion_failure_message`. This function has a bit complicated value checking, and I'd like to reorganize the structure if possible.
```suggestion
    if state == TrialState.COMPLETE:
        if values_conversion_failure_message is not None:
            raise ValueError(values_conversion_failure_message)
    elif state is None:
```

toshihikoyanase (commented):
The change looks good to me. I added a small suggestion.

optuna/study/_tell.py
not522: > This function has a bit complicated value checking, and I'd like to reorganize the structure if possible.

I agree, but may I work with another PR? I have a refactoring plan. It can be complicated to fix and refactor in same PR.

toshihikoyanase: That makes sense. Thank you.

toshihikoyanase (approved):
LGTM.

@not522 will refactor `Study.tell` as a follow up task.
@knshnb Let me merge this PR since it got two approvals.

PR #3807, opened 2022-07-25
Use DeterministicRelativeSampler in test_trial

keisuke-umezawa:
## Motivation

For https://github.com/optuna/optuna/issues/3724, it partially solved the issue for test_trial. Instead of using mock objects, the unit tests use DeterministicRelativeSampler to make them more readable.

## Description of the changes

- Used DeterministicRelativeSampler in test_trial

PR #380, opened 2019-04-15
New sampler interface.

sile:
This PR proposes a new interface of `BaseSampler`.

In addition to independent sampling algorithms (e.g., TPE), the new interface also supports implementing relative sampling algorithms such as GP, CMA-ES, etc.
It provides the following three abstract methods:
- `infer_relative_search_space()`:
   - Infers a relative search space for a trial.
   - The search space will be used by `sample_relative()` method.
- `sample_relative()`:
   - Samples all hyperparameters from a search space that inferred by `infer_relative_search_space()`.
   - The sampled hyperparameters will be used in the next trial.
       - If the trial requires hyperparameters not contained in the result of relative sampling, those will be independently sampled by using `sample_independent()` method.
- `sample_independent()`:
   - This is essentially the same as the current `BaseSampler.sample()` method.

See also: https://github.com/pfnet/optuna/pull/399

PR #3827, opened 2022-07-29
Introduce `DeterministicSampler` instead of `DeterministicRelativeSampler`

c-bata:
## Motivation

DeterministicRelativeSampler 
`DeterministicRelativeSampler` is a bit bothering to use since it requires `relative_search_space` just for sampling specific parameter values. In this PR, I introduce `DeterministicSampler` that does not require `relative_search_space` since it returns parameter values in `sample_independent`.

```python
class DeterministicSampler(optuna.samplers.BaseSampler):
    def __init__(self, params: Dict[str, Any]) -> None:
        self.params = params

        self, study: "optuna.study.Study", trial: "optuna.trial.FrozenTrial"
        return {}

    def sample_relative(
        self,
        study: "optuna.study.Study",
        trial: "optuna.trial.FrozenTrial",
        search_space: Dict[str, BaseDistribution],
    ) -> Dict[str, Any]:
        return {}

    def sample_independent(
        self,
        study: "optuna.study.Study",
        trial: "optuna.trial.FrozenTrial",
        param_name: str,
    ) -> Any:
        param_value = self.params[param_name]
        assert param_distribution._contains(param_distribution.to_internal_repr(param_value))
        return param_value
```

## Description of the changes

* Introduce `DeterministicSampler` and use it instead.
* Replace the use of `DeterministicRelativeSampler` with `RandomSampler` if it doesn't need to use it.

## TODO

* This PR conflicts with https://github.com/optuna/optuna/pull/3807. Please merge #3807 first, then I will push https://github.com/keisuke-umezawa/optuna/compare/feature/use-relative-sampler-test-trial...c-bata:opuna:use-deterministic-sampler?expand=1 and will make this PR ready for review.

### DIFF
--- a/tests/integration_tests/test_chainermn.py
+++ b/tests/integration_tests/test_chainermn.py
@@ -222,8 +222,8 @@ class TestChainerMNStudy:
     @pytest.mark.parametrize("storage_mode", STORAGE_MODES)
     def test_relative_sampling(storage_mode: str, comm: CommunicatorBase) -> None:
 
-        relative_params = {"x": 1.0, "y": 25.0, "z": -1.0}
-        sampler = DeterministicSampler(relative_params)
+        params = {"x": 1.0, "y": 25.0, "z": -1.0}
+        sampler = DeterministicSampler(params)
 
         with MultiNodeStorageSupplier(storage_mode, comm) as storage:
             study = TestChainerMNStudy._create_shared_study(storage, comm, sampler=sampler)
@@ -237,9 +237,9 @@ class TestChainerMNStudy:
             # Assert trial counts.
             assert len(mn_study.trials) == n_trials
 
-            # Assert the parameters in `relative_params` have been suggested among all nodes.
+            # Assert the parameters in `params` have been suggested among all nodes.
             for trial in mn_study.trials:
-                assert trial.params == relative_params
+                assert trial.params == params
 
     @staticmethod
     def _create_shared_study(
--- a/tests/integration_tests/test_chainermn.py
+++ b/tests/integration_tests/test_chainermn.py
@@ -22,7 +22,6 @@ from optuna.storages import BaseStorage
 from optuna.storages import InMemoryStorage
 from optuna.storages import RDBStorage
 from optuna.testing.pruners import DeterministicPruner
-from optuna.testing.samplers import DeterministicSampler
 from optuna.testing.storages import StorageSupplier
 from optuna.trial import Trial
 from optuna.trial import TrialState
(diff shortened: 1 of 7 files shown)


HideakiImamura:
@hvy @keisuke-umezawa Could you review this PR?

keisuke-umezawa (approved):
LGTM! Thank you for working on it.

tests/integration_tests/test_chainermn.py
```diff
@@ -222,13 +222,8 @@ def objective(_trial: ChainerMNTrial, _comm: bool) -> float:
...
 
-        relative_search_space = {
-            "x": distributions.FloatDistribution(low=-10, high=10),
-            "y": distributions.FloatDistribution(low=20, high=30, log=True),
-            "z": distributions.CategoricalDistribution(choices=(-1.0, 1.0)),
-        }
-        relative_params = {"x": 1.0, "y": 25.0, "z": -1.0}
-        sampler = DeterministicRelativeSampler(relative_search_space, relative_params)
+        params = {"x": 1.0, "y": 25.0, "z": -1.0}
+        sampler = DeterministicSampler(params)
```
hvy, line 226: I'm wondering if this `test_relative_sampling` is any meaningful at all. Do you think we should keep it? I'm neutral but the test is a bit confusing with "relative" in its name while using an "independent sampler".

Reference https://github.com/optuna/optuna/pull/380

c-bata: Thank you for your comment!
As you said, it looks that this test is not meaningful at all.

I don't think we should keep this test case since the test case is still passed although I changed to independent_sampler. I pushed https://github.com/optuna/optuna/pull/3827/commits/361874d5d0ba2a2ec198d3a59540c22ecb47c0d7.

PTAL.

hvy: Sounds good, thanks for the quick fix.

tests/integration_tests/test_chainermn.py
```diff
@@ -218,34 +217,6 @@ def objective(_trial: ChainerMNTrial, _comm: bool) -> float:
...
-
-            # Assert trial counts.
-            assert len(mn_study.trials) == n_trials
-
-            # Assert the parameters in `relative_params` have been suggested among all nodes.
-            for trial in mn_study.trials:
-                assert trial.params == relative_params
-
     @staticmethod
     def _create_shared_study(
```
hvy, line 221: Not a strong opinion but the optional `sampler` argument is no longer passed any called, with `test_relative_sampling` removed. How about removing this unused argument from the definition?

tests/integration_tests/test_cma.py
```diff
@@ -36,7 +35,7 @@ def test_init_cma_opts() -> None:
             cma_stds={"x": 1, "y": 1},
             seed=1,
             cma_opts={"popsize": 5},
-            independent_sampler=DeterministicRelativeSampler({}, {}),
+            independent_sampler=optuna.samplers.RandomSampler(),
```
hvy, line 38: Not sure but can we simplify by removing the argument altogether? (some ditto occurrences further down)
```suggestion
```

hvy (approved):
LGTM!

I left some optional comments if you have time to take a look at them. The master is broken (hopefully fixed by https://github.com/optuna/optuna/pull/3842) so I'll wait with merging anyway. 🙇

tests/integration_tests/test_chainermn.py
c-bata: Sounds reasonable! I removed `sampler` argument at f6f09ac.

tests/integration_tests/test_cma.py
hvy: I think there are several other places in the diff of this PR where this comment applies. Can we simplify them too?

hvy: - https://github.com/optuna/optuna/pull/3827/files#diff-59566bd9069c996def7505be23675a39fe165c302f2ddf6c2963dfc1281c7f1a
- https://github.com/optuna/optuna/pull/3827/files#diff-0bc6d8303a01e2fed31f1c1c6cd56fe355216944ce530c5a4fb121fc101f274a
- https://github.com/optuna/optuna/pull/3827/files#diff-0bc6d8303a01e2fed31f1c1c6cd56fe355216944ce530c5a4fb121fc101f274a 

Did you exclude those on purpose?

c-bata: Yes, I deliberately excluded them. In the following lines, mock objects are inserted into `independent_sampler` to check the `call_count`.

hvy: Ah I see, thanks. Let me resolve this thread.
(For a moment thought about having `independent_sampler` (and other arguments) a public attribute of e.g. `PyCmaSampler` but that's an entirely out-of-scope topic)

hvy (approved):
Thanks for the PR and the long running review, LGTM!

For the record, `optuna.testing` is not documented as public and this change is therefore not a compat breaking change.

PR #3828, opened 2022-07-29
Normalize importance values

contramundum53:
## Motivation
The documentation of `get_params_importance` states that the returned importance values should sum up to 1.0. However, `ShapleyImportanceEvaluator` currently does not normalize its output to 1.0.

It is more natural (and fits to many people's intuition) that the mean absolute SHAP value be *not* normalized to 1.0. We propose to add an option `normalize` to `get_params_importance` that controls whether the returned importance values should be normalized to 1.0. This option is defaulted to `True`, and the normalized version is displayed in `plot_params_importance`.

## Description of the changes
* Add `normalize` option to `get_params_importance`
* Add tests

## Alternative approach
There are possible alternative approaches to this problem.
* Specify `normalize` option in `ShapleyImportanceEvaluator`. This way, `plot_params_importance` can also display the unnormalized values. However, we may have to change the way `<0.01` is displayed in `plot_params_importance`.
* Force all importance values to be normalized/unnormalized.

### DIFF
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -25,17 +25,17 @@ def get_param_importances(
     evaluator: Optional[BaseImportanceEvaluator] = None,
     params: Optional[List[str]] = None,
     target: Optional[Callable[[FrozenTrial], float]] = None,
-    normalize_importances: bool = True,
+    normalize: bool = True,
 ) -> Dict[str, float]:
     """Evaluate parameter importances based on completed trials in the given study.
 
     The parameter importances are returned as a dictionary where the keys consist of parameter
     names and their values importances.
-    The importances are represented by floating point numbers that sum to 1.0 over the entire
-    dictionary.
-    The higher the value, the more important.
+    The importances are represented by non-negative floating point numbers, where higher values
+    mean that the parameters are more important.
     The returned dictionary is of type :class:`collections.OrderedDict` and is ordered by
     its values in a descending order.
+    If ``normalize`` is :obj:`True`, the sum of the importance values are normalized to 1.0.
 
     If ``params`` is :obj:`None`, all parameter that are present in all of the completed trials are
     assessed.
@@ -77,6 +77,10 @@ def get_param_importances(
                 Specify this argument if ``study`` is being used for multi-objective
                 optimization. For example, to get the hyperparameter importance of the first
                 objective, use ``target=lambda t: t.values[0]`` for the target parameter.
+        normalize:
+            A boolean option to specify whether the sum of the importance values should be
+            normalized to 1.0.
+            Defaults to :obj:`True`.
 
     Returns:
         An :class:`collections.OrderedDict` where the keys are parameter names and the values are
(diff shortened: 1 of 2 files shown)


optuna/importance/__init__.py
```diff
@@ -75,6 +77,10 @@ def get_param_importances(
                 Specify this argument if ``study`` is being used for multi-objective
                 optimization. For example, to get the hyperparameter importance of the first
                 objective, use ``target=lambda t: t.values[0]`` for the target parameter.
+        normalize:
+            A boolean option to specify whether the sum of the importance values should be
```
HideakiImamura, line 81: How about making this argument to "experimental"?

optuna/importance/__init__.py
```diff
@@ -24,16 +25,17 @@ def get_param_importances(
...
     The parameter importances are returned as a dictionary where the keys consist of parameter
     names and their values importances.
-    The importances are represented by floating point numbers that sum to 1.0 over the entire
-    dictionary.
-    The higher the value, the more important.
+    The importances are represented by non-negative floating point numbers, where higher values
+    mean that the parameters are more important.
     The returned dictionary is of type :class:`collections.OrderedDict` and is ordered by
     its values in a descending order.
+    If ``normalize`` is :obj:`True`, the sum of the importance values are normalized to 1.0.
```
HideakiImamura, line 38: How about the following?
```suggestion
    By default, the sum of the importance values are normalized to 1.0.
```

HideakiImamura (commented):
Thanks for the PR. I have some early comments. PTAL.

HideakiImamura:

@contramundum53 Could you merge the master to fix the CI error?

tests/importance_tests/test_init.py
```diff
@@ -103,15 +107,19 @@ def objective(trial: Trial) -> float:
 
         # Sanity check for param importances
         assert all(0 <= x < float("inf") for x in param_importance.values())
+        if normalize:
+            assert len(param_importance) == 0 or np.isclose(sum(param_importance.values()), 1.0)
```
nzw0301, line 111: I think the first condition is always false in this test function.
```suggestion
            assert np.isclose(sum(param_importance.values()), 1.0)
```

tests/importance_tests/test_init.py
```diff
@@ -184,6 +206,8 @@ def objective(trial: Trial) -> float:
...
```
nzw0301, line 210: ditto: the first condition is always false in this test function.
```suggestion
            assert np.isclose(sum(param_importance.values()), 1.0)
```

nzw0301 (changes requested):

nzw0301:
This might be out-of-scope in this PR, but let me leave one comment to improve docs. 

In my understanding, `get_param_importances`'s `normalize` flag doesn't affect MDI and fanova importance evaluators since they internally normalise importance score with sklearn (concretely, in `forest.feature_importances_`). So I'm wondering if we could clarify this behaviour on their pages.

contramundum53:
I think we might make `FanovaImportanceEvaluator` return unnormalized values in the future.

contramundum53:
Fixed. Thanks. @nzw0301

optuna/importance/__init__.py
contramundum53: Fixed. PTAL. @HideakiImamura

nzw0301 (approved):
Thanks.

HideakiImamura (approved):
LGTM.

PR #3740, opened 2022-06-30
Raise errors for NaN in constraint values

contramundum53:
## Motivation
Related to #3738, we should raise `ValueError` as early as possible when `nan` is detected in constraint values. We check that in `after_trial`.

## Description of the changes
Check for `nan` constraint values in `after_trial`.

Note that #3738 is still necessary because we cannot detect `nan` constraints in `add_trial`ed trials.

PR #3839, opened 2022-08-01
Fix infeasible case in NSGAII test

nzw0301:
## Motivation

Fix CI.

## Description of the changes

Remove `nan` as infeasible case since nsgaii raises value error when constraint contains nan value by https://github.com/optuna/optuna/pull/3740.


HideakiImamura (approved):
LGTM.

PR #3808, opened 2022-07-26
Fix check for number of objective values

not522:
## Motivation
The master's implementation incorrectly accepts single objective value for multi-objective optimization.

## Description of the changes
Fix `_check_and_convert_to_values` and add a test.

PR #3841, opened 2022-08-02
Refactor `_tell.py`

not522:
## Motivation
The implementation of `_tell.py` is complicated, which causes bugs (#3808, #3814, #3819). This PR adds private methods and simplifies `_tell_with_warning`.

## Description of the changes
* Add `_get_frozen_trial` and `_check_state_and_values` methods
* Change `_check_and_convert_to_values` to `_check_values`


optuna/study/_tell.py
```diff
@@ -97,106 +132,67 @@ def _tell_with_warning(
...
-            f"Skipped telling trial {trial_number} with values "
+            f"Skipped telling trial {frozen_trial.number} with values "
             f"{values} and state {state} since trial was already finished. "
             f"Finished trial has values {frozen_trial.values} and state {frozen_trial.state}."
         )
         return copy.deepcopy(frozen_trial)
-    elif frozen_trial.state == TrialState.WAITING:
-        raise ValueError("Cannot tell a waiting trial.")
+    elif frozen_trial.state != TrialState.RUNNING:
+        raise ValueError(f"Cannot tell a {frozen_trial.state.name} trial.")
```
not522, line 144: Originally, storage classes throw `RuntimeError` for finished trials. This PR changes to throw `ValueError` explicitly.

HideakiImamura:
@knshnb @contramundum53 Could you review this PR?

contramundum53:
@HideakiImamura Should this be merged before v3?

optuna/study/_tell.py
```diff
@@ -97,106 +132,70 @@ def _tell_with_warning(
...
         _logger.info(
-            f"Skipped telling trial {trial_number} with values "
+            f"Skipped telling trial {frozen_trial.number} with values "
             f"{values} and state {state} since trial was already finished. "
             f"Finished trial has values {frozen_trial.values} and state {frozen_trial.state}."
         )
         return copy.deepcopy(frozen_trial)
-    elif frozen_trial.state == TrialState.WAITING:
-        raise ValueError("Cannot tell a waiting trial.")
+    elif frozen_trial.state != TrialState.RUNNING:
```
knshnb, line 143: Let me clarify my understanding of this change. Here, you explicitly add assertation conditions of `frozen_trial.state` that were validated in `study._storage.set_trial_state_values(frozen_trial._trial_id, state, values)`?

knshnb (commented):
Thanks for the PR! I left some comments. PTAL.

optuna/study/_tell.py
not522: Exactly. I think it is clearer to validate it here like other variables.

optuna/study/_tell.py
```diff
@@ -97,106 +132,70 @@ def _tell_with_warning(
...
+        raise ValueError(f"Cannot tell a {frozen_trial.state.name} trial.")
+
+    _check_state_and_values(state, values)
```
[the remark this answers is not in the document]
not522: I rearranged the validation order according to the order of the arguments of this method.

optuna/study/_tell.py
```diff
@@ -22,59 +20,96 @@
...
+        if values is not None:
+            raise ValueError(
+                "Values were told. Values cannot be specified when state is "
+                "TrialState.PRUNED or TrialState.FAIL."
+            )
+    elif state is not None:
+        raise ValueError(f"Cannot tell with state {state}.")
+
+
+def _check_values(
```
knshnb, line 78: [Nit] I feel the roles of `_check_state_and_values` and `_check_values` are unclear at a glance. How about renaming this function to something like `_check_non_none_values`? Or do you come up with any better idea?

optuna/study/_tell.py
```diff
@@ -22,59 +20,96 @@
...
+        return f"The value {repr(value)} could not be cast to float."
 
-    if value is not None and math.isnan(value):
-        value = None
-        failure_message = f"The value {original_value} is not acceptable."
+    if math.isnan(value):
+        return f"The value {value} is not acceptable."
 
-    return value, failure_message
+    return None
```
knshnb, line 112: [Nit] Not a strong opinion, but I think this function is simple enough to be directly put inside `_check_values`, which would make the role of each function clearer. What do you think?

optuna/study/_tell.py
knshnb: Thanks for the clarification!

optuna/study/_tell.py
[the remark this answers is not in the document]
knshnb: That makes sense! Then, I think adding the comment at the beginning of each validation part will also make the code much more readable.

optuna/study/_tell.py
```diff
@@ -97,106 +132,70 @@ def _tell_with_warning(
...
                 warning_message = values_conversion_failure_message
 
     assert state is not None
 
+    if values is not None:
+        if isinstance(values, Sequence):
+            values = [float(value) for value in values]
+        else:
+            values = [float(values)]
+
```
knshnb, line 190: Can we simplify `_check_values` a little by doing this conversion first?

knshnb (commented):
I added some other small comments. PTAL.

contramundum53 (approved):
LGTM except for @knshnb 's comment!

optuna/study/_tell.py
not522: This conversion is not safe without value checking and requires error handling.  I think it is better to check first and convert later.

knshnb: Oh, sorry for not being clear. I meant the conversion without cast (https://github.com/optuna/optuna/pull/3841/files#diff-b54e13f298eb2cf1237096a8de699cf349e84774b96f8458ef8b8dc0bc03615aR81-R84) can be done first.

optuna/study/_tell.py
not522: How about `_check_values_are_feasible`?

knshnb: That sounds fine! (Or you can also ignore this comment because now I feel it's too minor a point...)

not522:
Thank you for the detailed review. I addressed your comments. PTAL.

knshnb:
Thanks for the update! Almost LGTM, but you dropped the cast to `float` of each element of `values` by [this commit](https://github.com/optuna/optuna/pull/3841/commits/48883f731d5704009b2a1f592dba307e48facfef)?

not522:
Thank you. Fixed.

knshnb (approved):
LGTM. Thanks for the great refactoring!

PR #3854, opened 2022-08-05
Add journal-style log storage

wattlebirdaz:
## Motivation
Introduce **JournalStorage** to optuna.

Journal storage writes a record of every operation to the database as it is executed and at the same time, keeps a latest snapshot of the database in-memory. If the database crashes for any reason, the storage can re-establish the contents in memory by replaying the operations stored from the beginning.

Journal storage has several benefits over the conventional value logging storages.
1. The number of IOs can be reduced because of larger granularity of logs.
2. Journal storage (`BaseJournalLogStorage`) has a simpler backend API than the conventional storage (`BaseStorage`), which makes it easier to plug-in other backend implementations.
3. Journal storage keeps a snapshot in-memory so it can avoid the use of internal cache, which reduces complications of code.

## Description of the changes
- `optuna/storages/_journal/storage.py`
  - Implementation of the `BaseStorage` named as `JournalStorage`.
- `optuna/storages/_journal/base.py`
  - Abstract class of `JournalStorage` backend named as `BaseJournalLogStorage`.
- `optuna/storages/_journal/file.py`
  - File storage backend that implements `BaseJournalLogStorage` named as `JournalFileLogStorage`. This file also contains implementations of locks needed for synchronizing processes.
- `tests/storages_tests/test_journal.py`
  - Test `JournalFileLogStorage`.

## References
- https://zenn.dev/sile/articles/335fe2213f9204


c-bata:

optuna/storages/_journal/storage.py
```diff
@@ -0,0 +1,604 @@
...
+
+    def _sync_with_backend(self) -> None:
+        logs = self._backend.read_logs(self._log_number_read)
+        for log in logs:
+            self._log_number_read += 1
+            self._apply_log(log)
+
+    def _create_unique_study_name(self) -> str:
+        self._study_name_suffix_num += 1
+        return "no-name-" + self._pid + "-" + str(self._study_name_suffix_num)
```
keisuke-umezawa, line 363: [question] The naming convention for a study is different from the [InMemoryStorage](https://github.com/optuna/optuna/blob/master/optuna/storages/_in_memory.py#L65). Did you do it on purpose?

keisuke-umezawa: Also, better to use `DEFAULT_STUDY_NAME_PREFIX `.

optuna/storages/_journal/storage.py
```diff
@@ -0,0 +1,604 @@
...
+
+    # Basic study manipulation
+
+    def create_new_study(self, study_name: Optional[str] = None) -> int:
+        with self._thread_lock:
+            study_name = study_name or self._create_unique_study_name()
+            self._write_log(JournalOperation.CREATE_STUDY, {"study_name": study_name})
```
keisuke-umezawa, line 370: [question] IMO, we should check that there are any duplicated study names like [this line](https://github.com/optuna/optuna/blob/master/optuna/storages/_in_memory.py#L62) before committing this log, otherwise we will have an error when syncing with backend. But, did you do it on purpose?

optuna/storages/_journal/storage.py
```diff
@@ -0,0 +1,604 @@
...
+            self._sync_with_backend()
+
```
keisuke-umezawa, line 372: [nits] You can add a log like [this](https://github.com/optuna/optuna/blob/master/optuna/storages/_in_memory.py#L69)

optuna/storages/_journal/storage.py
```diff
@@ -0,0 +1,604 @@
...
+            for frozen_study in self._studies.values():
+                if frozen_study.study_name == study_name:
+                    return frozen_study._study_id
+            assert False, "Should not reach."
+
+    def delete_study(self, study_id: int) -> None:
+        with self._thread_lock:
+            self._write_log(JournalOperation.DELETE_STUDY, {"study_id": study_id})
```
keisuke-umezawa, line 380: [question] IMO, we should check that the study exists like [this line](https://github.com/optuna/optuna/blob/master/optuna/storages/_in_memory.py#L76) before committing this log, otherwise we will have an error when syncing with backend. But, did you do it on purpose?

keisuke-umezawa (commented):
Quickly checked `Basic study manipulation` part in storage.py, and left some comments to understand the algorithm.

optuna/storages/_journal/storage.py
wattlebirdaz: Yes, on purpose. In the current design, erroneous logs can be flushed ("committed") to the backend. This is because replaying the unread logs already in the backend and appending the new logs to the backend cannot be done in one critical section. Whatever the log is, a worker needs to write the log first and then replay all the unread log in the backend (including the log written just now) to see if it raises an error.

optuna/storages/_journal/storage.py
wattlebirdaz: Same as: https://github.com/optuna/optuna/pull/3854#discussion_r941951828

optuna/storages/_journal/storage.py
wattlebirdaz: I had no special reason for this. Will push a fix.

optuna/storages/_journal/storage.py
```diff
@@ -0,0 +1,608 @@
...
+            return copy.deepcopy(self._replay_result.get_all_studies())
+
+    # Basic trial manipulation
+    def create_new_trial(self, study_id: int, template_trial: Optional[FrozenTrial] = None) -> int:
+        log: Dict[str, Any] = {
+            "study_id": study_id,
+            "datetime_start": datetime.datetime.now().isoformat(),
+        }
+
+        if template_trial:
```
keisuke-umezawa, line 173: [note] Difficult to understand the logic.

keisuke-umezawa (commented):
Went through the code of implementation. I will check unit tests later.

Some comments for the high level architecture.
1. If the file stores a lot of studies e.g. 1000 but users just want to read one of studies, can we do anything for it to reduce the reading time and memory usage?
2. Does this storage support multi-threading or multi-processing? It can be written in docstrings. If it supports both of them, I will check the code more deeply.

optuna/storages/_journal/storage.py
keisuke-umezawa: I understand. Thank you for explanation.

wattlebirdaz:
Thank you for the comment!

> * If the file stores a lot of studies e.g. 1000 but users just want to read one of studies, can we do anything for it to reduce the reading time and memory usage?

I think there are two possible ways to achieve better performance in reading the information of a specific study.

1. Implement a snapshot persisting feature.
2. Use a separate file for saving logs of different studies.

We may work on these in the near future but to keep it simple I think saving all logs into one file will do for now. What do you think? @c-bata 

> * Does this storage support multi-threading or multi-processing? It can be written in docstrings. If it supports both of them, I will check the code more deeply.

Yes it does support both multi-threading & multi-process executions. You can try them using the script [here](https://gist.github.com/wattlebirdaz/02a9afeca728001907dde7f8f2e027d4)!

optuna/storages/_journal/storage.py
```diff
@@ -0,0 +1,608 @@
...
+                    "Another study with name '{}' already exists. "
+                    "Please specify a different name, or reuse the existing one "
+                    "by setting `load_if_exists` (for Python API) or "
+                    "`--skip-if-exists` flag (for CLI).".format(study_name)
+                ),
+            )
+            return
+
+        study_id = self._next_study_id
+        self._next_study_id += 1
```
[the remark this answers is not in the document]
wattlebirdaz: This actually works. If a worker creates a new study, it will log the operation into the backend. (This is done by the [`_write_log`](https://github.com/wattlebirdaz/optuna/blob/a0cc8dc334a5e9bcea10cc4c6e82ffd4c4131280/optuna/storages/_journal/storage.py#L101) function.) If any worker reads this log from the backend, it will replay it by calling the [`apply_create_study`](https://github.com/wattlebirdaz/optuna/blob/a0cc8dc334a5e9bcea10cc4c6e82ffd4c4131280/optuna/storages/_journal/storage.py#L388) function. (This is done by the [`sync`](https://github.com/wattlebirdaz/optuna/blob/a0cc8dc334a5e9bcea10cc4c6e82ffd4c4131280/optuna/storages/_journal/storage.py#L102) function.) Since the logs are shared by the centralized file, the order of the logs is kept the same, therefore, the all the workers can issue the same study_id for a specific study.

c-bata:
> We may work on these in the near future but to keep it simple I think saving all logs into one file will do for now. What do you think? @c-bata

Agree 👍  For now, as a workaround for such a situation, users can save logs in separate files.

optuna/storages/_journal/storage.py
```diff
@@ -0,0 +1,608 @@
...
+    def __init__(self, pid: str) -> None:
+        self.log_number_read = 0
+        self._pid = pid
+        self._studies: Dict[int, FrozenStudy] = {}
+        self._trials: Dict[int, FrozenTrial] = {}
+
+        self._study_id_to_trial_ids: Dict[int, List[int]] = {}
+        self._trial_id_to_study_id: Dict[int, int] = {}
+        self._next_study_id: int = 0
+        self._trial_ids_owned_by_this_process: List[int] = []
```
keisuke-umezawa, line 321: We should rename `_trial_ids_owned_by_this_process` to `_trial_ids_created_by_this_process` because this list holds trials which will be just created by this process. The reason I emphasized this name is that, when we enqueue a trial, it is created by this process but not necessarily owned by it.

optuna/storages/_journal/storage.py
```diff
@@ -0,0 +1,608 @@
...
+
+        with self._thread_lock:
+            self._write_log(JournalOperation.SET_TRIAL_STATE_VALUES, log)
+            self._sync_with_backend()
+
+            if (
+                state == TrialState.RUNNING
+                and trial_id not in self._replay_result._trial_ids_owned_by_this_process
+            ):
+                return False
```
keisuke-umezawa, line 255: [question] What happens when the trial state becomes `WAITING` to `RUNNING` in the multi-processing situation?

I will pick up a concrete example like:
1. `The process A` enqueues `the trial 1` with `WAITING` state. In that case, `the process A` has `the trial 1` in `_trial_ids_owned_by_this_process`.
2. And, as it is in the multi-processing situation, `the process B` asks `the trial 1` coincidentally. In that case, `the process B` tries to `set_trial_state_values(trial._trial_id, state=TrialState.RUNNING)` into `the trial 1`.
3. However, `the process B` does not have `the trial 1` in the `_trial_ids_owned_by_this_process`. So, `set_trial_state_values` returns `False`. It is different behavior comparing to the current master, AFAIK.

It is a complicated behavior, and I might be wrong, but could you check it?

keisuke-umezawa: And, if it is a bug, we need to add a unit test for it. Also, we need to carefully check that we have enough unit tests around `WAITING` state behaviors.

optuna/storages/_journal/storage.py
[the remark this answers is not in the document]
keisuke-umezawa: Thank you for answering my weird question, I understand. It is in the applying phase.

optuna/storages/_journal/storage.py
wattlebirdaz: Thank you for reporting the case. The implementation was wrong. Fixed in [b67e0b5](https://github.com/optuna/optuna/pull/3854/commits/b67e0b53d4aa33ab0e0bc64a96942732deb1c0c6).

optuna/storages/_journal/storage.py
wattlebirdaz: Addressed in [b67e0b5](https://github.com/optuna/optuna/pull/3854/commits/b67e0b53d4aa33ab0e0bc64a96942732deb1c0c6).

wattlebirdaz:
[note] [c89613f](https://github.com/optuna/optuna/pull/3854/commits/c89613fd93f7613a95ffa88819c52286fbcb5ad9)

# Bench 10000 trials using 10 processes

## Benchcode
```python3
import optuna
from optuna import optuna
from optuna.logging import set_verbosity, ERROR

def objective(trial):
    x = trial.suggest_float("x", 0, 0)
    return x**2


num_executors = 10
num_executions = 10
trials_per_worker = 1000
study_name = "concurrent_optimization"


def test_concurrent_optimization(id):
    study = optuna.create_study(
        study_name=study_name,
        storage=optuna.storages.JournalStorage(
            optuna.storages.JournalFileStorage("./journal_logfile")
        ),
        load_if_exists=True,
    )
    study.optimize(objective, n_trials=trials_per_worker)
    return id

set_verbosity(ERROR)
with ProcessPoolExecutor(num_executors) as pool:
    result = pool.map(test_concurrent_optimization, [id for id in range(num_executions)])
    for id in result:
        print("{} done !".format(id))
```
## Use seek:
```python3
    def __init__(self, file_path: str, lock_obj: Optional[JournalFileBaseLock] = None) -> None:
        self._file_path: str = file_path
        self._lock = lock_obj or JournalFileSymlinkLock(self._file_path)
        open(self._file_path, "a").close()  # Create a file if it does not exist
        self._log_number_offset: Dict[int, int] = {0: 0}

    def read_logs(self, log_number_from: int) -> List[Dict[str, Any]]:
        with get_lock_file(self._lock):
            logs = []
            with open(self._file_path, "r") as f:
                log_number_start = 0
                if log_number_from in self._log_number_offset:
                    f.seek(self._log_number_offset[log_number_from])
                    log_number_start = log_number_from

                for log_number, line in enumerate(f, start=log_number_start):
                    if log_number + 1 not in self._log_number_offset:
                        byte_len = len(line.encode("utf-8"))
                        self._log_number_offset[log_number + 1] = (
                            self._log_number_offset[log_number] + byte_len
                        )
                    if log_number < log_number_from:
                        continue
                    logs.append(json.loads(line))

            return logs
```

real    1m36.268s
user    7m41.293s
sys     6m48.092s

## Use readlines:

```python3
    def __init__(self, file_path: str, lock_obj: Optional[JournalFileBaseLock] = None) -> None:
        self._file_path: str = file_path
        self._lock = lock_obj or JournalFileSymlinkLock(self._file_path)
        open(self._file_path, "a").close()  # Create a file if it does not exist

    def read_logs(self, log_number_from: int) -> List[Dict[str, Any]]:
        with get_lock_file(self._lock):
            logs = []
            with open(self._file_path, "r") as f:
                for lineno, line in enumerate(f):
                    if lineno < log_number_from:
                        continue
                    logs.append(json.loads(line))
            return logs
```
real    7m1.939s
user    38m26.430s
sys     29m20.125s

optuna/storages/_journal/base.py
```diff
@@ -0,0 +1,23 @@
...
+from typing import Any
+from typing import Dict
+from typing import List
+
+
+class BaseJournalLogStorage(abc.ABC):
+    """Base class for Journal storages.
+
+    Storage classes implementing this base class must guarantee process safety. Thread-safety
+    need not be guaranteed. Use :class:`~optuna.storages.JournalFileSymlinkLock` or
```
sile, line 11: I'm not sure that it's appropriate to mention `JournalFileSymlinkLock` and `JournalFileOpenLock` here as those classes seem too specific to `JournalFileStorage`. For example, let's assume that an implemention of this abstract class uses RDB (or Redis, etc) as the storage and users want to run distributed optimizations from multiple machines, `JournalFileSymlinkLock` and `JournalFileOpenLock` could not (and should not) be used for such purposes (instead there would be RDB specific locking mechanisms we could use).

optuna/storages/_journal/file.py
```diff
@@ -0,0 +1,175 @@
...
+        self._lockrenamefile = self._lockfile + str(uuid.uuid4()) + RENAME_FILE_SUFFIX
+
+    def acquire(self) -> bool:
+        while True:
+            try:
+                os.symlink(self._lock_target_file, self._lockfile)
+                return True
+            except OSError as err:
+                if err.errno == errno.EEXIST:
+                    continue
```
sile, line 56: I think that it's better to take a sleep (e.g., exponential backoff) here to reduce I/O load when contention is high.

optuna/storages/_journal/file.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def acquire(self) -> bool:
+        while True:
+            try:
+                open_flags = os.O_CREAT | os.O_EXCL | os.O_WRONLY
+                os.close(os.open(self._lockfile, open_flags))
+                return True
+            except OSError as err:
+                if err.errno == errno.EEXIST:
+                    continue
```
sile, line 100: Ditto

optuna/storages/_journal/file.py
```diff
@@ -0,0 +1,175 @@
...
+            Path of file to persist the log to.
+
+        lock_obj:
+            Lock object for process exclusivity.
+
+    """
+
+    def __init__(self, file_path: str, lock_obj: Optional[JournalFileBaseLock] = None) -> None:
+        self._file_path: str = file_path
+        self._lock = lock_obj or JournalFileSymlinkLock(self._file_path)
```
sile, line 142: It seems that if users specify `lock_obj`  they need to write the same path twice as follows:  
```python
storage = JournalFileStorage("/path/to/optuna-db", JournalFileSymlinkLock("/path/to/optuna-db"))
```
IMO, it's redundant and error prone, so how about refine the interface? (e.g., makes `lock_obj` a function that takes a filepath and returns an `JournalFileBaseLock` instance)

optuna/storages/_journal/file.py
```diff
@@ -0,0 +1,175 @@
...
+        open(self._file_path, "a").close()  # Create a file if it does not exist
+        self._log_number_offset: Dict[int, int] = {0: 0}
+
+    def read_logs(self, log_number_from: int) -> List[Dict[str, Any]]:
+        with get_lock_file(self._lock):
```
sile, line 147: Do we actually need to acquire a lock here?
Since `JournalFileStorage` just appends JSONs to a file and doesn't modify the existing part of the file, it would be safe to read the contents of the file without lock (we should take care a case that the JSON at the file tail could be temporarily broken depending on the file write / flush timing though).
Let me know if I misunderstand something :pray:

optuna/storages/_journal/file.py
wattlebirdaz: The assumption is that there are very few users who change `lock_obj` (most cases should be fine if symlink lock is used, and we only provide a path in case it is necessary to change it), so this case, IMO need not be considered that much. 

As for the interface that specifies a factory function to generate `lock_obj`, although duplicate specification of filepath is eliminated, it may be necessary to prepare a factory function for each lock class, and it is somewhat questionable whether this would really make the user happy.

optuna/storages/_journal/file.py
wattlebirdaz: Nice catch!

As you mentioned, optimistic reading of the data entails risks of reading collapsed data due to concurrent writers, so we will need to add some mechanism that the data it read is not broken. Do you have any suggestions on doing so?
I think the easiest way to do so is to add a checksum for each line of JSON and persist it together. And when reading data, make sure that the checksum matches with the checksum computed from the line of JSON.

optuna/storages/_journal/base.py
wattlebirdaz: My intention of this doc-string was to mention those `Journal...Locks` as an option for storages which do not internally have locking or multi-process exclusion mechanisms. That's why I wrote `Use ... if your storage
    does not natively support multi-process query executions.`. I'll fix the doc-string because it seems ambiguous.

optuna/storages/_journal/file.py
sile: >  it may be necessary to prepare a factory function for each lock class

Hmm, I thought that we could write the code as follows without any additional functions (because Python's constructors can be used as functions):
```python
storage = JournalFileStorage("/path/to/optuna-db", JournalFileSymlinkLock)
```

optuna/storages/_journal/file.py
sile: IMO, it's enough to just ignore JSON decode error of the last line as follows:

```python
                last_decode_error = None # Added
                for log_number, line in enumerate(f, start=log_number_start):
                    if last_decode_error is not None:
                        raise last_decode_error # Added (the error is not at the last line)
                    if log_number + 1 not in self._log_number_offset:
                        byte_len = len(line.encode("utf-8"))
                        self._log_number_offset[log_number + 1] = (
                            self._log_number_offset[log_number] + byte_len
                        )
                    if log_number < log_number_from:
                        continue
                    try:        
                        logs.append(json.loads(line))
                    except e:
                        last_decode_err = e # Added
```

optuna/storages/_journal/file.py
wattlebirdaz: I think this should be avoided because we don't define the constructor arguments of `JournalFileBaseLock`.  Users might implement their own file lock which does not have `file_path` as constructor argument. On such cases, we cannot use the passed `UserDefinedLock` as alternative to the default `JournalFileSymlinkLock`.

sile: > Users might implement their own file lock which does not have file_path as constructor argument

On such case, users can use `lambda` as follows I think:

```python
storage = JournalFileStorage("/path/to/optuna-db", lambda path: UserDefinedLock(...))
```

Honestly, I cannot imagine that there would be an implmentation class of `JournalFileBaseLock` that doesn't take the DB file path as a constructor argument (as it's a file based lock anyway). But, the design is a matter of taste and, as you said above, using a lock class other than  `JournalFileSymlinkLock` would be very rare case, so feel free to ignore my opinion and keep the implementation as-is.

wattlebirdaz: I was imagining something like `ThreadLock` in python. `ThreadLock` does not take the "resource to protect" as an constructor argument. 
Anyways, thank you for the comment!

optuna/storages/_journal/file.py
wattlebirdaz: Fixed in https://github.com/optuna/optuna/pull/3854/commits/d7869c0edb5f6b6a94d2f858ef65369563fb50ba

optuna/storages/_journal/file.py
wattlebirdaz: Fixed in https://github.com/optuna/optuna/pull/3854/commits/3917a29e3796673f3c5cdbc4e7fa7eb0b1aa7f4f.

optuna/storages/_journal/file.py
wattlebirdaz: Fixed in https://github.com/optuna/optuna/pull/3854/commits/3917a29e3796673f3c5cdbc4e7fa7eb0b1aa7f4f.

optuna/storages/_journal/base.py
wattlebirdaz: Fixed in https://github.com/optuna/optuna/pull/3854/commits/948684ae8fde0f0da4ee2d508dac383a360005ec.

optuna/storages/_journal/file.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+class JournalFileBaseLock(abc.ABC):
+    @abc.abstractmethod
+    def acquire(self) -> bool:
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def release(self) -> None:
+        raise NotImplementedError
```
c-bata, line 28: Could you add the docstring to each method in this class to fill in the description of the `acquire()` and `release()` methods in JournalFileSymlinkLock and JournalFileOpenLock?

Screenshot 2022-08-30 14 48 32

* https://output.circle-artifacts.com/output/job/4db460c6-d068-405d-a771-84ff61b3f0e9/artifacts/0/docs/build/html/reference/generated/optuna.storages.JournalFileSymlinkLock.html#optuna.storages.JournalFileSymlinkLock
* https://output.circle-artifacts.com/output/job/4db460c6-d068-405d-a771-84ff61b3f0e9/artifacts/0/docs/build/html/reference/generated/optuna.storages.JournalFileOpenLock.html#optuna.storages.JournalFileOpenLock

optuna/storages/_journal/file.py
```diff
@@ -0,0 +1,184 @@
...
+                    last_decode_error = err
+                    del self._log_number_offset[log_number + 1]
+
+            return logs
+
+    def append_logs(self, logs: List[Dict[str, Any]]) -> None:
+        with get_lock_file(self._lock):
+            what_to_write = ""
+            for log in logs:
+                what_to_write += json.dumps(log) + "\n"
```
c-bata, line 180: It looks that we can simply use `"\n".join()` here.

```suggestion
            what_to_write = "\n".join([json.dumps(log) for log in logs])
```

optuna/storages/_journal/storage.py
```diff
@@ -0,0 +1,612 @@
...
+            if "distributions" not in log
+            else {k: json_to_distribution(v) for k, v in log["distributions"].items()}
+        )
+        params = (
+            {}
+            if "params" not in log
+            else {
+                k: distributions[k].to_external_repr(param) for k, param in log["params"].items()
+            }
+        )
```
c-bata, line 476: [nits] Can we simplify these lines like below?

```suggestion
        distributions = {}
        if "distributions" in log
            distributions = {k: json_to_distribution(v) for k, v in log["distributions"].items()}
        params = {}
        if "params" in log
            params = {k: distributions[k].to_external_repr(p) for k, p in log["params"].items()}
```

optuna/storages/_journal/storage.py
```diff
@@ -0,0 +1,612 @@
...
+    2. Journal storage has simpler backend API than value logging storage.
+    3. Journal storage keeps a snapshot in-memory so no need to add more cache.
+
+    Example:
+
+        .. code::
+
+            storage = JournalStorage(JournalFileStorage("./log_file"))
+
+            study = optuna.create_study(storage=storage)
```
c-bata, line 76: [nits] How about adding `import` statements like the docstring of RedisStorage?
https://optuna.readthedocs.io/en/stable/reference/generated/optuna.storages.RedisStorage.html

c-bata (commented):
Thank you for your pull request. Changes almost looks good to me.
I wrote some minor review comments.

optuna/storages/_journal/storage.py
```diff
@@ -71,10 +71,17 @@ class JournalStorage(BaseStorage):
...
         .. code::
 
+            import optuna
+
+
+            def objective(trial):
+                ...
+
+
             storage = JournalStorage(JournalFileStorage("./log_file"))
```
c-bata, line 81: Should we import JournalStorage and JournalFileStorage here?

wattlebirdaz: I added the prefix like this.
https://github.com/optuna/optuna/pull/3854/commits/90a59b31766a0650f8055004e2a101c311111eeb
Which do you think is better, with `import optuna.storages.JournalStorage` and  `import optuna.storages.JournalFileStorage` or without them?

optuna/storages/_journal/file.py
wattlebirdaz: https://github.com/optuna/optuna/pull/3854/commits/9c987a7374bb5f79d17a2f789656cd7afc549a2c
https://github.com/optuna/optuna/pull/3854/commits/1e89cda03f34fadff657feecd8a3cf88144390db

optuna/storages/_journal/file.py
wattlebirdaz: https://github.com/optuna/optuna/pull/3854/commits/bcce9487ecfc1a6c14b187653fe523afe3006e4b

optuna/storages/_journal/storage.py
wattlebirdaz: https://github.com/optuna/optuna/pull/3854/commits/6b5b0af6d137435ed8ab313dbc5bd568cd0318a8
https://github.com/optuna/optuna/pull/3854/commits/90a59b31766a0650f8055004e2a101c311111eeb

c-bata (approved):
LGTM!

c-bata:
Please merge the master branch to fix CI failures after #3950 is merged.

sile (approved):
LGTM

PR #3861, opened 2022-08-08
Fix typo: Replace `trail` with `trial`

nzw0301:
## Motivation

I found a typo.

## Description of the changes

Use the correct spell `trials` instead of `trails`.


nzw0301:
I think a single approval is enough for this PR since the introduced change is trivial.

not522 (approved):
Good catch! LGTM.

Issue #3699, opened 2022-06-20
Type-Checking for the objective function


valentindoering:
### Motivation

<img width="889" alt="Screenshot 2022-06-20 at 15 52 58" src="https://user-images.githubusercontent.com/54007786/174617322-a3d59f44-3020-4af8-a35e-f63abcd81404.png">
This is the error, if you accidentally return a string in the  provided objective function. A type assertion would have saved me 2h hours of debugging.

(number of values - 18 - was the length of the string, the number was very confusing)

### Description

please add type, assertions, nice error descriptions to the optuna API.

nzw0301:
@valentindoering If you have any suggestions to improve the logging message (#3868) or another approach, please let us know!

PR #3857, opened 2022-08-07
Change the log message format for failed trials.

erentknn:
## Motivation
This PR fixes the issue #3779. 

## Description of the changes

I simply added a new callback to print the parameters and trial number of the current trial. I am not too familiar with the codebase so there might be some mistakes but I can create a fix quickly.

PR #3868, opened 2022-08-10
Make log message user-friendly when objective returns a sequence of unsupported values

nzw0301:
## Motivation

As reported by https://github.com/optuna/optuna/issues/3699, the logger's message might be unclear for users when the objective returns `str` (technically, a sequence of a value that cannot be converted into a float value). For example,

```python
import optuna


def objective(trial):
    y = trial.suggest_categorical("y", [-1, 0, 1])
    return "sample"



for num_objectives in [1, 2]:
    study = optuna.create_study(sampler=sampler, directions=["minimize"]*num_objectives)
    study.optimize(objective, n_trials=2)
```

The output is as follows:

```shell
# single objective
[I 2022-08-10 13:14:33,683] A new study created in memory with name: no-name-0a782794-38dd-4fcc-a9a0-9d624184d70d
[W 2022-08-10 13:14:33,686] Trial 0 failed because of the following error: The number of the values 6 did not match the number of the objectives 1.
[W 2022-08-10 13:14:33,686] Trial 1 failed because of the following error: The number of the values 6 did not match the number of the objectives 1.

# multi-objective
[I 2022-08-10 13:14:33,687] A new study created in memory with name: no-name-5d52d04d-1ccd-4649-91dc-2fe78ec4288b
[W 2022-08-10 13:14:33,688] Trial 0 failed because of the following error: The number of the values 6 did not match the number of the objectives 2.
[W 2022-08-10 13:14:33,689] Trial 1 failed because of the following error: The number of the values 6 did not match the number of the objectives 2.
```

By applying the change in this PR, the output will be as follows:

```shell
# single objective
[I 2022-08-10 13:09:45,755] A new study created in memory with name: no-name-f2888b23-e568-4d6e-b1e0-9f5310ad7960
[W 2022-08-10 13:09:45,757] Trial 0 failed because of the following error: The value 's' could not be cast to float.
[W 2022-08-10 13:09:45,759] Trial 1 failed because of the following error: The value 's' could not be cast to float.

# multi-objective
[I 2022-08-10 13:09:45,761] A new study created in memory with name: no-name-3437816d-e80d-4c6f-ba69-b6daa0ce834a
[W 2022-08-10 13:09:45,762] Trial 0 failed because of the following error: The value 's' could not be cast to float.
[W 2022-08-10 13:09:45,764] Trial 1 failed because of the following error: The value 's' could not be cast to float.
```

## Description of the changes

Move the logic that checks the consistency between the number of objectives and directions.

## Discussions

Note that this PR cannot detect the following case, where the sequence of `str` digit, to maintain backwards compatibility of the master brach:

```python
import optuna


def objective(trial):
    y = trial.suggest_categorical("y", [-1, 0, 1])
    return "12"


study = optuna.create_study(sampler=sampler, directions=["minimize"]*2)
study.optimize(objective, n_trials=2)
```

### DIFF
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -255,4 +255,4 @@ def _log_failed_trial(
         exc_info=exc_info,
     )
 
-    _logger.warning("Trial {} failed with value {}.".format(trial.number, value_or_values))
+    _logger.warning("Trial {} failed with value {}.".format(trial.number, repr(value_or_values)))
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -218,9 +218,18 @@ def _run_trial(
             _logger.info("Trial {} pruned. {}".format(frozen_trial.number, str(func_err)))
         elif frozen_trial.state == TrialState.FAIL:
             if func_err is not None:
-                _log_failed_trial(frozen_trial, repr(func_err), exc_info=func_err_fail_exc_info)
+                _log_failed_trial(
+                    frozen_trial,
+                    repr(func_err),
+                    exc_info=func_err_fail_exc_info,
+                    value_or_values=value_or_values,
+                )
             elif STUDY_TELL_WARNING_KEY in frozen_trial.system_attrs:
-                _log_failed_trial(frozen_trial, frozen_trial.system_attrs[STUDY_TELL_WARNING_KEY])
+                _log_failed_trial(
+                    frozen_trial,
+                    frozen_trial.system_attrs[STUDY_TELL_WARNING_KEY],
+                    value_or_values=value_or_values,
+                )
             else:
                 assert False, "Should not reach."
         else:
@@ -236,9 +245,14 @@ def _run_trial(
 
 
 def _log_failed_trial(
-    trial: FrozenTrial, message: Union[str, Warning], exc_info: Any = None
+    trial: FrozenTrial,
+    message: Union[str, Warning],
+    exc_info: Any = None,
+    value_or_values: Any = None,
 ) -> None:
     _logger.warning(
         "Trial {} failed because of the following error: {}".format(trial.number, message),
         exc_info=exc_info,
     )
+
+    _logger.warning("Trial {} failed with value {}.".format(trial.number, value_or_values))
--- a/optuna/study/_tell.py
+++ b/optuna/study/_tell.py
@@ -30,16 +30,6 @@ def _check_and_convert_to_values(
     else:
         _original_values = [original_value]
 
-    if n_objectives != len(_original_values):
-        return (
-            None,
-            (
-                f"The number of the values "
-                f"{len(_original_values)} did not match the number of the objectives "
-                f"{n_objectives}."
-            ),
-        )
-
     _checked_values = []
     for v in _original_values:
         checked_v, failure_message = _check_single_value(v, trial_number)
@@ -52,6 +42,16 @@ def _check_and_convert_to_values(
             _checked_values.append(checked_v)
         else:
             assert False
+
+    if n_objectives != len(_original_values):
+        return (
+            None,
+            (
+                f"The number of the values "
+                f"{len(_checked_values)} did not match the number of the objectives "
+                f"{n_objectives}."
+            ),
+        )
 
     return _checked_values, None
 


c-bata:
@not522 @HideakiImamura could you review this PR?

not522:
Sorry for late reply. How about outputting the same log as a `COMPLETE` trial? I recommend the following.

```
[...] Trial 0 finished with value: ['s', 'a', 'm', 'p', 'l', 'e'] and parameters: {'x': 54.26412865334919, 'y': 1}.
[...] Trial 0 failed because of the following error: ...
```

nzw0301:
I'd like to say that you don't need to say sorry at all since many dev members living in Japan were (are?) on holiday :)

Thank you for your recommendation. I have two concerns:
1. The failed frozen trial has `None` as `value` or `values` attributes instead of `['s', 'a', 'm', 'p', 'l', 'e']`. So using never-stored invalid value: `['s', 'a', 'm', 'p', 'l', 'e']`, might be confusing for users especially for debugging. But I might be too worried about this.
2. As concerned by @g-votte, increasing the number of logging lines is undesirable since optuan already shows many logging lines.

not522:
> The failed frozen trial has None as value or values attributes instead of ['s', 'a', 'm', 'p', 'l', 'e']. So using never-stored invalid value: ['s', 'a', 'm', 'p', 'l', 'e'], might be confusing for users especially for debugging. But I might be too worried about this.

I see. There may be room for improvement in the wording, but anyway, outputting returned values and parameters would be useful for debugging.

> As concerned by @g-votte, increasing the number of logging lines is undesirable since optuna already shows many logging lines.

I think that helpful error messages can be long. (I agree that INFO messages should be short and clear, as you mentioned.)

nzw0301:
> outputting returned values and parameters would be useful for debugging.
> I think that helpful error messages can be long. (I agree that INFO messages should be short and clear, as you mentioned.)

I completely agree with you. The retuned value (and params) should be in failed trial log. I've updated the PR to show the retuned value too. I believe that showing `params` in a failed trial log message will be addressed by a related PR, https://github.com/optuna/optuna/pull/3857.

The current log messages with the example code above is as follows:

```bash
[I 2022-08-22 14:38:26,369] A new study created in memory with name: no-name-c8713597-3eb6-4da4-9909-c24b3b16c175
[W 2022-08-22 14:38:26,370] Trial 0 failed because of the following error: The value 's' could not be cast to float.
NoneType: None
[W 2022-08-22 14:38:26,370] Trial 0 failed with value None.
[W 2022-08-22 14:38:26,371] Trial 1 failed because of the following error: The value 's' could not be cast to float.
NoneType: None
[W 2022-08-22 14:38:26,371] Trial 1 failed with value None.
[I 2022-08-22 14:38:26,372] A new study created in memory with name: no-name-170c9e03-6c61-4091-9284-b8485fcd6745
[W 2022-08-22 14:38:26,373] Trial 0 failed because of the following error: The value 's' could not be cast to float.
NoneType: None
[W 2022-08-22 14:38:26,373] Trial 0 failed with value None.
[W 2022-08-22 14:38:26,373] Trial 1 failed because of the following error: The value 's' could not be cast to float.
NoneType: None
[W 2022-08-22 14:38:26,373] Trial 1 failed with value None.
```

nzw0301:
The log message looks something wrong... I'll revise the change. Sorry for confusing you.

optuna/study/_optimize.py
```diff
@@ -236,9 +245,14 @@ def _run_trial(
...
+    message: Union[str, Warning],
+    exc_info: Any = None,
+    value_or_values: Any = None,
 ) -> None:
     _logger.warning(
         "Trial {} failed because of the following error: {}".format(trial.number, message),
         exc_info=exc_info,
     )
+
+    _logger.warning("Trial {} failed with value {}.".format(trial.number, value_or_values))
```
not522, line 258: I think the message should distinct float and string, e.g., 1.23 and '1.23'.

```suggestion
    _logger.warning("Trial {} failed with value {}.".format(trial.number, repr(value_or_values)))
```

nzw0301: Thank you for your suggestion!

HideakiImamura (approved):
Looks great.

not522 (approved):
LGTM!

PR #3872, opened 2022-08-10
Fix docstring in `FrozenTrial`

wattlebirdaz:
## Motivation
Fix docstring in `FrozenTrial`.

## Description of the changes
Added docstring for `distributions`, `system_attrs`, and `trial_id` for `FrozenTrial`.

### DIFF
--- a/optuna/trial/_frozen.py
+++ b/optuna/trial/_frozen.py
@@ -113,12 +113,18 @@ class FrozenTrial(BaseTrial):
             Datetime where the :class:`~optuna.trial.Trial` finished.
         params:
             Dictionary that contains suggested parameters.
+        distributions:
+            Dictionary that contains parameter distributions.
         user_attrs:
             Dictionary that contains the attributes of the :class:`~optuna.trial.Trial` set with
             :func:`optuna.trial.Trial.set_user_attr`.
+        system_attrs:
+            Dictionary that contains the attributes of the :class:`~optuna.trial.Trial` set with
+            :func:`optuna.trial.Trial.set_system_attr`.
         intermediate_values:
             Intermediate objective values set with :func:`optuna.trial.Trial.report`.
-
+        trial_id:
+            ID of the :class:`~optuna.trial.Trial`
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


optuna/trial/_frozen.py
```diff
@@ -113,12 +113,18 @@ def objective(trial):
...
             Dictionary that contains the attributes of the :class:`~optuna.trial.Trial` set with
             :func:`optuna.trial.Trial.set_user_attr`.
+        system_attrs:
+            Dictionary that contains the attributes of the :class:`~optuna.trial.Trial` set with
+            :func:`optuna.trial.Trial.set_system_attr`.
         intermediate_values:
             Intermediate objective values set with :func:`optuna.trial.Trial.report`.
-
+        trial_id:
+            ID of the :class:`~optuna.trial.Trial`
```
c-bata, line 127: FrozenTrial provides `_trial_id` attribute which is not public to users. So I'd say we don't need to list `trial_id` here. What do you think?

contramundum53:
@wattlebirdaz  Could you fix the following CI error?
```
/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/optuna/trial/_frozen.py:docstring of optuna.trial.FrozenTrial.distributions:1: WARNING: duplicate object description of optuna.trial.FrozenTrial.distributions, other instance in reference/generated/optuna.trial.FrozenTrial, use :noindex: for one of them
```

optuna/trial/_frozen.py
c-bata: Thank you for your input!
There is no `trial_id` attribute actually. It should not be listed in `Attributes` section since it's a constructor argument.

We may be able to document `trial_id` argument in `Args` section. However, I cannot find such an example that document `Args` section in docstring for Python classes. Do you think we should add `Args` section in this docstring?
https://google.github.io/styleguide/pyguide.html#384-classes

contramundum53: Sorry, I misunderstood. I agree that it should not be listed in `Attributes` section.

c-bata (approved):
LGTM!

c-bata:
Let me merge this PR with one approval since this PR just changes the docstring of FrozenTrial.

Issue #3565, opened 2022-05-16
`constant_liar=True` does not work in multi-objective `TPESampler`


knshnb:
### What is an issue?

In multi-objective `TPESampler`, running trials are skipped even when `constant_liar=True`. I think we should at least mention this in the document and should probably give a warning.
Also, it seems that constant_liar can be easily supported in the multi-objective case by setting `inf` or `-inf` to each objective.

- [x] Add to the document
- [ ] Raise a warning if `constant_liar=True` is set for multi-objective optimization
- [ ] Support `constant_liar=True` for multi-objective optimization
https://github.com/optuna/optuna/blob/595e731cd150f60a9ca9ea89f135dc95d35ac83b/optuna/samplers/_tpe/sampler.py#L617-L618

HideakiImamura:
We should add a document for this in this issue.

Making the `constant_liar` option available in the multi-objective optimization is a future work.

knshnb:
To support this feature, we might need to be careful with #3676 and #3677, which are inf-related issues with multi-objective `TPESampler`.

knshnb:
I updated the issue description. The second task is also contribution-welcome!

not522:
This issue has been resolved by #5021. Thank you for your contributions.

PR #3881, opened 2022-08-11
Add note for `constant_liar` with multi-objective function

nzw0301:
## Motivation

Currently, TPE sampler's `constant_liar` argument is ignored for multi-objective function as pointed out by https://github.com/optuna/optuna/issues/3565. This feature should be clarified in the documentation as suggested by https://github.com/optuna/optuna/issues/3565#issuecomment-1166913554.

## Description of the changes

- Add a note section to explain when we can use `constant_liar`.
- Minor document fix (https://github.com/optuna/optuna/pull/3881/commits/9970d20392a54ec5addc25536fb7cc77ff05388e)

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -206,6 +206,10 @@ class TPESampler(BaseSampler):
                 workers is high.
 
             .. note::
+                This feature can be used for only single-objective optimization; this argument is
+                ignored for multi-objective optimization.
+
+            .. note::
                 Added in v2.8.0 as an experimental feature. The interface may change in newer
                 versions without prior notice. See
                 https://github.com/optuna/optuna/releases/tag/v2.8.0.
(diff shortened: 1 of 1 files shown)


optuna/samplers/_tpe/sampler.py
```diff
@@ -205,6 +205,10 @@ def objective(trial):
                 and the durations of the running states are significant, and/or the number of
                 workers is high.
 
+            .. note::
+                This feature can be used for only single-objective optimization; this argument is
+                ignored for multi-objective optimization.
+
```
nzw0301, line 211: Alternatively, remove this note and update `constant_liar`'s explanation as follows since `constant_liar` already has three `note` section and added section looks short:

```
            If :obj:`True`, penalize running trials to avoid suggesting parameter configurations
            nearby for only single-objective optimization. This argument is ignored for multi-objective optimization.
```

contramundum53 (approved):
LGTM! Thanks!

PR #3882, opened 2022-08-11
Use `copybutton_prompt_text` not to copy the bash prompt

nzw0301:
## Motivation

When we click the copy button (concretely, see also the bottom right corner in the screenshot below) on the code snippet of a command line example, the prompt sign, `$`,  is also copied, which is not ideal because users would like to run the copied command without unnecessary prompt removal.

Screenshot 2022-08-11 at 23 00 07

Page: https://optuna.readthedocs.io/en/latest/installation.html
 
## Description of the changes

Use `copybutton_prompt_text` variable by `sphinx-copybutton` that says

> When this variable is set, sphinx-copybutton will remove the prompt from the beginning of any lines that start with the text you specify. In addition, only the lines that contain prompts will be copied if any are discovered. If no lines with prompts are found, then the full contents of the cell will be copied.

Ref: https://sphinx-copybutton.readthedocs.io/en/latest/use.html#strip-and-configure-input-prompts-for-code-cells

As a result, we copy the command line without prompt `$`.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -179,6 +179,9 @@ autodoc_default_options = {
     "exclude-members": "with_traceback",
 }
 
+# sphinx_copybutton option to not copy prompt.
+copybutton_prompt_text = "$ "
+
 # Sphinx Gallery
 pio.renderers.default = "sphinx_gallery"
 


nzw0301:
https://output.circle-artifacts.com/output/job/9126214a-52cd-4294-948d-d94c8df0b8a4/artifacts/0/docs/build/html/installation.html is a generated page, where we can copy a code snippet

contramundum53 (approved):
Good catch! LGTM!

PR #3887, opened 2022-08-12
Simplify IntersectionSearchSpace.calculate

knshnb:
## Motivation

I simplified the implementation of `IntersectionSearchSpace.calculate` a little.

## Description of the changes

Directly filter the search space dictionary instead of creating a delete list and deleting them one by one.

### DIFF
--- a/optuna/samplers/_search_space/intersection.py
+++ b/optuna/samplers/_search_space/intersection.py
@@ -83,15 +83,11 @@ class IntersectionSearchSpace:
                 self._search_space = copy.copy(trial.distributions)
                 continue
 
-            delete_list = []
-            for param_name, param_distribution in self._search_space.items():
-                if param_name not in trial.distributions:
-                    delete_list.append(param_name)
-                elif trial.distributions[param_name] != param_distribution:
-                    delete_list.append(param_name)
-
-            for param_name in delete_list:
-                del self._search_space[param_name]
+            self._search_space = {
+                name: distribution
+                for name, distribution in self._search_space.items()
+                if name in trial.distributions and trial.distributions[name] == distribution
+            }
 
         self._cursor = next_cursor
         search_space = self._search_space or {}
--- a/optuna/samplers/_search_space/intersection.py
+++ b/optuna/samplers/_search_space/intersection.py
@@ -86,7 +86,7 @@ class IntersectionSearchSpace:
             self._search_space = {
                 name: distribution
                 for name, distribution in self._search_space.items()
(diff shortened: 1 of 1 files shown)


contramundum53:
@knshnb This might cause a performance degradation, since the `set` object is created many times throughout the calculation. 
(Originally, `delete_list` is an empty list in most cases.)

contramundum53:
@not522 What do you think?

contramundum53:
I think it is possible to simplify this program using generators without affecting the performance, but the code might be trickier.

knshnb:
Thanks for pointing it out! I confirmed that when `delete_list` is empty, creating a new `dict` object takes longer than just iterating over the existing `dict` in the Google Colab environment.

```python
d = {x: 2 * x for x in range(1000)}
```
```python
%%timeit
for _ in range(10000):
  d1 = {k: v for k, v in d.items() if k >= 0}
```
1 loop, best of 5: 1.02 s per loop
```python
%%timeit
for _ in range(10000):
  delete_list = []
  for k, v in d.items():
    if k < 0:
      delete_list.append(k)
  for k in delete_list:
    del d[k]
```
1 loop, best of 5: 595 ms per loop

knshnb:
This part should probably not be a bottleneck, but I don't have a strong opinion and there might be no need to dare to include this change. What do you think?

contramundum53:
I see, it seems that this change won't make much difference.

optuna/samplers/_search_space/intersection.py
```diff
@@ -83,15 +83,11 @@ def calculate(self, study: Study, ordered_dict: bool = False) -> Dict[str, BaseD
...
-                    delete_list.append(param_name)
-                elif trial.distributions[param_name] != param_distribution:
-                    delete_list.append(param_name)
-
-            for param_name in delete_list:
-                del self._search_space[param_name]
+            self._search_space = {
+                name: distribution
+                for name, distribution in self._search_space.items()
+                if name in trial.distributions and trial.distributions[name] == distribution
```
not522, line 89:
```suggestion
                if trial.distributions.get(name) == distribution
```

not522 (commented):
Basically LGTM. I left one suggestion.

optuna/samplers/_search_space/intersection.py
knshnb: Thanks!

not522 (approved):

contramundum53:
LGTM!

PR #3890, opened 2022-08-14
Let `frozen_trial` be used in `study.ask(...)`

nickvazz:
## Motivation

When using the `study.ask`/`study.tell` interface, if passing a `FrozenTrial` to `study.tell`, it would throw an error although within the `optuna/study/_tell.py` file, it would go retrieve it. It seems like if you already have the `FrozenTrial` it should be able to `study.tell(FrozenTrial)`

## Description of the changes

- Don't `Assert` when checking if it is a trial and it is a frozen trial
- Grab the `trial.number` and `trial._trial_id` from the frozen trial as well
- Just re-use the frozen trial

### DIFF
--- a/optuna/study/_tell.py
+++ b/optuna/study/_tell.py
@@ -79,7 +79,7 @@ def _check_single_value(
 
 def _tell_with_warning(
     study: "optuna.Study",
-    trial: Union[trial_module.Trial, int],
+    trial: Union[trial_module.Trial, trial_module.FrozenTrial, int],
     values: Optional[Union[float, Sequence[float]]] = None,
     state: Optional[TrialState] = None,
     skip_if_finished: bool = False,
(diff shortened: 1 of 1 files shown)


contramundum53:
@not522 @c-bata Could you review this PR? It might involve some design discussion.

not522:
Thanks for the PR. I discussed the design with other developers, and we concluded `FrozenTrial` is unsuitable for the argument of `tell`. The document says, "`FrozenTrial` does not depend on `Study`". For example, when using `create_trial`, the `number` is -1.
Another concern is multi-processing. If another process changes the trial's state, the `FrozenTrial`'s state and the one saved in storage may be different. Therefore, we need to call `study._storage.get_trial` even though we have `FrozenTrial`.
Let me close this PR for the above reasons.

PR #3902, opened 2022-08-19
Update cli tutorial

### LINKED ISSUE #3032: Update CLI tutorial
The current CLI tutorial uses deprecated features. https://optuna.readthedocs.io/en/v2.10.0/tutorial/20_recipes/004_cli.html

New features of CLI have been introduced into master. I recommend rewriting the tutorial using the ask-and-tell interface.

nzw0301:
## Motivation

Resolve #3032.

## Description of the changes

As described in #3032, remove deprecated command examples and use ask and tell commands. However, since the `ask` and `tell` commands in CLI module doesn't look suitable for the current example code `foo.py`, I've created another scenario to use these commands.


optuna/cli.py
```diff
@@ -847,7 +847,7 @@ def get_parser(self, prog_name: str) -> ArgumentParser:
             "--skip-if-finished",
             default=False,
             action="store_true",
-            help="If specified, tell is skipped without any error when the trial is already"
+            help="If specified, tell is skipped without any error when the trial is already "
```
nzw0301, line 850: Currently, this part is printed as `alreadyfinished`.

tutorial/20_recipes/004_cli.py
```diff
@@ -42,33 +42,30 @@ def objective(trial):
...
-
-###################################################################################################
+# However, if we cannot write ``objective`` explicitly in Python code such as developing a new
+# drug in a lab, a more interactive way is suitable.
+# In Optuna CLI, :ref:`ask_and_tell` style commands provide such an interactive and flexible interface.
+#
+# Let us assume we minimize the objective value depending on a parameter ``x`` in :math:`[-10, 10]`
+# and objective value is calculated via some experiments by hand.
 # Even so, we can invoke the optimization as follows.
-# (Don't care about ``--storage sqlite:///example.db`` for now, which is described in :ref:`rdb`.)
```
nzw0301, line 56: I'd like to suggest removing `(` and `)`.

toshihikoyanase:
@keisuke-umezawa Could you review this PR, please. I think we can merge this PR with a single approval since it updates the tutorial only.

tutorial/20_recipes/004_cli.py
```diff
@@ -42,33 +42,30 @@ def objective(trial):
...
-#         return (x - 2) ** 2
-#
 #     $ STUDY_NAME=`optuna create-study --storage sqlite:///example.db`
-#     $ optuna study optimize foo.py objective --n-trials=100 --storage sqlite:///example.db --study-name $STUDY_NAME
-#     [I 2018-05-09 10:40:25,196] Finished a trial resulted in value: 54.353767789264026. Current best value is 54.353767789264026 with parameters: {'x': -5.372500782588228}.
-#     [I 2018-05-09 10:40:25,197] Finished a trial resulted in value: 15.784266965526376. Current best value is 15.784266965526376 with parameters: {'x': 5.972941852774387}.
-#     ...
-#     [I 2018-05-09 10:40:26,204] Finished a trial resulted in value: 14.704254135013741. Current best value is 2.280758099793617e-06 with parameters: {'x': 1.9984897821018828}.
+#     $ optuna ask --storage sqlite:///example.db --study-name $STUDY_NAME --sampler TPESampler \
+#          --search-space '{"x": {"name": "FloatDistribution", "attributes": {"step": null, "low": -10.0, "high": 10.0, "log": false}}}'
```
keisuke-umezawa, line 58: Do we actually need to specify `step` and `log` here? If this commend can run without these arguments, we can remove them. Some people might get stuck by them.

keisuke-umezawa (commented):
Thanks for updating it with a more concrete example. I left a comment. Could you check it?

tutorial/20_recipes/004_cli.py
nzw0301: No, we don't. However, I believe that we use https://optuna.readthedocs.io/en/stable/reference/generated/optuna.distributions.distribution_to_json.html to obtain the json formatted distribution, which includes `step` and `log` as seen above.

nzw0301: How about 

1. removing `log` and `step` key values from the example above for simplicity and readability.
2. Adding a sentence to mention https://optuna.readthedocs.io/en/stable/reference/generated/optuna.distributions.distribution_to_json.html to generate json formatted distribution?

keisuke-umezawa: I understand. In that case, to generate the json formatted distribution, we need to explain how we can create it. I agree on your idea.

> Adding a sentence to mention https://optuna.readthedocs.io/en/stable/reference/generated/optuna.distributions.distribution_to_json.html to generate json formatted distribution?

You can explain it by the above. But, users need to run some code to generate it. The other way is that you can explain it by some sentences like:

```
You can choose distributions from `FloatDistribution` and `IntDistribution`, and specify parameters by the key values of json with `low`, `high`, `log` and `step`. If you want to use log scale, you can set `"log": true`, otherwise `"log": false`. If you want to specify step e.g. 2, you can set `"step": 2`, otherwise `"step": null`.
```

The above sentences is just a quick job, so we need to refine it.

nzw0301: Thanks. I think the explanation of each argument of the distribution looks unnecessary for this tutorial, so I've added only links to https://optuna.readthedocs.io/en/stable/reference/generated/optuna.distributions.distribution_to_json.html t and distributions. For the consistency with the output of `distribution_to_json`, I've not removed `log` and `step` from the example code.

keisuke-umezawa (approved):
LGTM! Thank you for the updates.

PR #3665, opened 2022-06-14
hotfix: Add a version constraint of cached-path

c-bata:
## Motivation

Fix the tests of allennlp integration.
https://github.com/optuna/optuna/runs/6871089185?check_suite_focus=true

## Description of the changes

AllenNLP uses `cached_path.file_friendly_logging()` function [here](https://github.com/allenai/allennlp/blob/a6271a319cb92f73e798cf3a33ac49ee18cd7cc5/allennlp/common/file_utils.py#L135). However, [config-path v1.1.3](https://github.com/allenai/cached_path/releases/tag/v1.1.3) removed it.

In this PR, I add a version constraint as a hotfix.

PR #3904, opened 2022-08-20
Remove version constraint of `cached-path`

nzw0301:
## Motivation

Revert than change by https://github.com/optuna/optuna/pull/3665 because the latest stable allennlp doesn't require an older `cached-path` anymore according to https://github.com/allenai/allennlp/blob/2d8fe00904d4b1a249dab2784309c86618ffa1a0/requirements.txt#L6.

## Description of the changes

Remove the changes by https://github.com/optuna/optuna/pull/3665.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -92,8 +92,6 @@ def get_extras_require() -> Dict[str, List[str]]:
         ],
         "integration": [
             "allennlp>=2.2.0 ; python_version>'3.6'",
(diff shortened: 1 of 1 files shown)


nzw0301:
Hmm, by removing the line, the version of `protobuf` is inconsistent between allennlp and tensorflow. We need to wait for another update in tensorflow.

PR #3870, opened 2022-08-10
Attach abstractmethod decorator to `BaseStorage.get_trial_id_from_study_id_trial_number`

wattlebirdaz:
## Motivation
Attach `abc.abstractmethod` decorator to `BaseStorage.get_trial_id_from_study_id_trial_number(self, study_id: int, trial_number: int) -> int`

## Description of the changes
Attached `abc.abstractmethod` decorator to `BaseStorage.get_trial_id_from_study_id_trial_number(self, study_id: int, trial_number: int) -> int`

PR #3909, opened 2022-08-22
Remove abstractmethod decorator from `get_trial_id_from_study_id_trial_number`

c-bata:
## Motivation

Revert changes in #3870.

## Description of the changes

@knshnb told me that `get_trial_id_from_study_id_trial_number` is actually not an abstractmethod though it raises NotImplementedError. When `NotImplementedError` is raised, following lines handles an exception and returns trial_id.

https://github.com/optuna/optuna/blob/5a1bb75bcb92173c7ee99ebe8b1abf1335f85b3d/optuna/study/_tell.py#L122-L142

I think we should implement the default logic in BaseSampler. However, this PR revert #3870 for now, since we plan to release v3.0 next week.

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -331,7 +331,6 @@ class BaseStorage(object, metaclass=abc.ABCMeta):
         """
         raise NotImplementedError
 
-    @abc.abstractmethod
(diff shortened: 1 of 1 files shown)


knshnb (approved):
LGTM!

PR #3910, opened 2022-08-22
Move default logic of `get_trial_id_from_study_id_trial_number()` method to BaseStorage.

c-bata:
## Motivation

Follow-up #3909. Please see the description of #3909.

## Description of the changes

Implement default logic of `get_trial_id_from_study_id_trial_number` in BaseStorage.


c-bata:
@hvy @knshnb Could you review this PR if you have time?

c-bata:
@wattlebirdaz Could you review this PR since hvy-san will take a vacation.

wattlebirdaz:
Thanks for the pull request. LGTM.

wattlebirdaz (approved):
Thanks for the pull request. LGTM.

tests/study_tests/test_study.py
```diff
@@ -1485,22 +1485,12 @@ def test_tell_duplicate_tell() -> None:
...
-                study.tell(study.ask().number, 1.0)
-
-            with pytest.raises(ValueError):
-                study.tell(study.ask().number + 1, 1.0)
+        study = create_study(storage=storage)
+        study.tell(study.ask(), 1.0)
+        study.tell(study.ask().number, 1.0)
+
+        with pytest.raises(ValueError):
+            study.tell(study.ask().number + 1, 1.0)
```
knshnb, line 1493: With this, `InMemoryStorage.get_trial_id_from_study_id_trial_number` is tested instead of `BaseStorage.get_trial_id_from_study_id_trial_number`.

knshnb: I cannot think of a way to easily test `InMemoryStorage.get_trial_id_from_study_id_trial_number` (except for mocking all the abstract methods of `InMemoryStorage`), but do you come up with a good way...?

optuna/study/_tell.py
```diff
@@ -123,23 +123,6 @@ def _tell_with_warning(
...
                 study._study_id, trial_number
             )
-        except NotImplementedError as e:
-            warnings.warn(
-                "Study.tell may be slow because the trial was represented by its number but "
-                f"the storage {study._storage.__class__.__name__} does not implement the "
-                "method required to map numbers back. Please provide the trial object "
-                "to avoid performance degradation."
-            )
-
```
knshnb, line 133: Do you think it is okay to drop this error message? (In the current interface, keeping this warning seems to be a bit difficult.)

knshnb (commented):
Thanks for the PR! I think this PR is generally a good direction, but I left some concerns and questions. PTAL.

optuna/study/_tell.py
c-bata: Yes. Even in the current interface it is possible to emit a warning message in `BaseStorage.get_trial_id_from_study_id_trial_number()`, but I don't think we should do that.

BaseStorage already provides the default logic in some storage methods, which may cause a performance degradation (e.g. `get_trial_number_from_id`, `get_n_trials`, `get_best_trial`, etc.). However, they do not emit warning messages actually. `RDBStorage` can be made faster by overriding these methods, but if we use InMemoryStorage or CachedStorage, this default logic would not be a bottleneck.

tests/study_tests/test_study.py
c-bata: Good catch! It is a difficult problem how to test concrete methods in an abstract base class. I thought about it and decided it would be better to remove this test case. At least it should not be in `test_study.py`.

c-bata: It seems that `get_trial_id_from_study_id_trial_number()` is enough tested at:

https://github.com/optuna/optuna/blob/2c09c5d689411aae19a613970ade3a1b54b01ddb/tests/storages_tests/test_storages.py#L1070-L1096

optuna/study/_tell.py
knshnb: Thanks for elaborating. That makes sense. Let's drop this error message for consistency with the other default methods 👍

tests/study_tests/test_study.py
knshnb: I agree that this test should not be located in `test_study.py`.
Still, I feel the default logic of `BaseStorage.get_trial_id_from_study_id_trial_number` should be tested somewhere (not tested in `test_get_trial_id_from_study_id_trial_number` because all the concrete class override the method). We should probably discuss how to test default implementations in `BaseStorage` as another issue.

knshnb: So, let me merge this PR first. If you have any opinions, please leave a comment!

knshnb (approved):
Thanks for the update! LGTM. Please comment if you have any opinion on the point I mentioned.

c-bata:
Thank you for your thorough review!
In general it is difficult to test concrete methods in an abstract base class. I think the fundamental solution is to avoid implementing concrete methods in `RDBStorage`.

PR #3911, opened 2022-08-22
Suppress warnings in tests for `multi_objective` module

not522:
## Motivation
Part of https://github.com/optuna/optuna/issues/3815.
`multi_objective` module is deprecated. We can ignore `FutureWarning` in its tests.

## Description of the changes
Ignore `FutureWarning` in tests for `multi_objective` module.

### DIFF
--- a/tests/multi_objective_tests/samplers_tests/test_nsga2.py
+++ b/tests/multi_objective_tests/samplers_tests/test_nsga2.py
@@ -10,6 +10,9 @@ from optuna import multi_objective
 from optuna.study._study_direction import StudyDirection
 
 
+pytestmark = pytest.mark.filterwarnings("ignore::FutureWarning")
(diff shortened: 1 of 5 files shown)


nzw0301 (approved):
Looks great to me! Since all tests in `tests/multi_objective_tests/samplers_tests/test_motpe.py` are already wrapped by `@pytest.mark.filterwarnings("ignore:FutureWarning")`, we don't need `pytestmark` as in the files changed by this PR.

toshihikoyanase (approved):
LGTM. The change suppresses 880 warnings in total.

I've never know, but the `pytestmark` global variable can be used to apply marks at the module level. 
https://docs.pytest.org/en/7.1.x/example/markers.html#marking-whole-classes-or-modules

PR #3913, opened 2022-08-22
Replace `q` with `step` in private function and warning message

nzw0301:
## Motivation

A warning message in `FloatDistribution` uses the deprecated class / function's argument `q` when we run the following code:

```python
import optuna

optuna.distributions.FloatDistribution(1, 2, step=0.3)
```

```bash
/Users/nzw/Documents/optuna/optuna/distributions.py:668:
    UserWarning: The distribution is specified by [1, 2] and q=0.3, but the range is not divisible by `q`. It will be replaced by [1, 1.9].
  warnings.warn(
```

In the warming message, `q` isn't clear for users from the current API.

On the other hand, the deprecated class `DiscreteUniformDistribution` and method `suggest_discrete_uniform`  refer to `q` as "A discretization step. q must be larger than 0." and "A step of discretization.", respectively. So I suggest replacing `q` with `step` in the warning message since we can infer the meaning of `step` from the docs of deprecated class/function.

## Description of the changes

Replace `q` with `step` in the warning message and the private method of the distribution module.

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -655,20 +655,20 @@ def check_distribution_compatibility(
         )
 
 
-def _adjust_discrete_uniform_high(low: float, high: float, q: float) -> float:
+def _adjust_discrete_uniform_high(low: float, high: float, step: float) -> float:
     d_high = decimal.Decimal(str(high))
     d_low = decimal.Decimal(str(low))
-    d_q = decimal.Decimal(str(q))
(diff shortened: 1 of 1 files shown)


not522 (approved):
Good catch! LGTM.

toshihikoyanase (approved):
Sounds reasonable. Thank you!

PR #3942, opened 2022-08-29
Reduced warning messages from tests/visualization_tests/test_slice.py

jmsykes83:
## Motivation:

As part of #3815
---
## Description:

Suppressed user warnings for tests involving the _check_plot_args function
in the tests/visualization_tests/test_slice.py file.


nzw0301 (commented):
Thank you for your PR again! I'm wondering if we could change `target_name` from the default `target_name`: `"Objective Value"` to reduce the warning messages because it doesn't increase # lines of test files and we can avoid unintended filtering from another source.

jmsykes83:
> 

I will look into it.

Issue #3518, opened 2022-04-27
Gotchas with the MaxTrialsCallback


ernestum:
### What is an issue?

The documented go-to solution to do distributed optimization for `n` trials independent of the number of workers is to use the [MaxTrialsCallback](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.MaxTrialsCallback.html?highlight=MaxTrialsCallback). There are a number unexpected behaviors with this:

1. If you use the default `states` argument and run `m` workers on `n` trials, you will actually get `n + m` trials. This is because already running trials are not counted by default. This can be very confusing. Unfortunately just adding `TrialState.RUNNING` to the `states` list won't solve the issue for all cases: when you have stale trials you have to remove them by hand before adding more workers.

2. When you try to continue a study that already has the specified `n_trials`, each worker will add one more trial to the study even though it should not. This is due to the callback only being executed first _after_ the first trials has been executed.

3. If you enable pruning and don't include `TrialState.PRUNED` in the `states` argument, you might run into the issue that you workers keep running forever when all the new trials are being pruned. This should at least be documented.

HideakiImamura:
Thanks for the raising this issue. It would be great if you give us a PR to fix this.

nzw0301:
Hi in my understanding, actually a few problems in the issue description can be avoided by using Optuna's API or adding a few lines. Let me share them as follows.

> 1. If you use the default `states` argument and run `m` workers on `n` trials, you will actually get `n + m` trials. This is because already running trials are not counted by default. This can be very confusing. Unfortunately just adding `TrialState.RUNNING` to the `states` list won't solve the issue for all cases: when you have stale trials you have to remove them by hand before adding more workers.

By passing `None` as `states`, it would perform `m` trials in practice since this case callback cares all trial states including `RUNNING`. If we have stale trials in them, we can re-run them by using https://optuna.readthedocs.io/en/stable/reference/generated/optuna.storages.fail_stale_trials.html, which makes a trial state `FAIL` and https://optuna.readthedocs.io/en/stable/reference/generated/optuna.storages.RetryFailedTrialCallback.html, which retry the failed trial. So we don't need remove them by hand. 

> 2. When you try to continue a study that already has the specified `n_trials`, each worker will add one more trial to the study even though it should not. This is due to the callback only being executed first _after_ the first trials has been executed.

Yes, since this callback is called after the evaluation of the trial. Fortunately, this can be avoided by adding a few lines as follows:

```python
[...]

Issue #3938, opened 2022-08-28
N_trials on parallel optimization runs n_trials per pod and not in total


Ohadkaiser:
### Expected behavior

When running the optimize function in optuna on distributed pods in parallel, the expected behavior is that the pods will synchronize on the number of trial and the total amount will be the number set with the parameters n_trials.
Currently, n_trials define the number of trials per pod. 

### Environment

- Optuna version: 2.10.0
- Python version: 3.7
- OS: Linux 
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
.
```

### Steps to reproduce

.

Ohadkaiser:
The change that need be done is the following.

In _optimize_sequential function we need to check if study.storage.get_n_trials if smaller than n_trials instead of i_trial

nzw0301:
I suppose it is not the bug as in https://github.com/optuna/optuna/issues/3518.

Ohadkaiser:
> I suppose it is not the bug as in #3518.

That you for your response!
I will suggest a solution to #3518 as I see fit.

Issue #2907, opened 2021-09-07
Unexpected behavior using --n-trials flag with distributed optimization


jkterry1:
Per the original discussion [here](https://github.com/DLR-RM/rl-baselines3-zoo/issues/153), when running trials with distributed optimization, the n-trials flag only cuts off running once a single runner crosses it, not all runners. I don't believe this is what users would expect this do to and this was unexpected behavior to me and the maintainer of one of the larger open source projects depending on Optuna. Could this be changed? I also can't find this feature documented on the website after Googling.

nzw0301:
As a beginner of optuna, I think it depends on the usage of optuna and the current behaviour is fine. This is because we can increase the number of trials if we are not happy with the results of the first optimisation like

```python
study.optimize(objective, n_trials=100) # not good objective value yet
study.optimize(objective, n_trials=100) # perform additional 100 trials to get a better value
```


From a more practical perspective, I think [this callback](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.MaxTrialsCallback.html) is useful to set the maximum number of trials for distributed optimisation.

As a contributor, I think it would be great to mention this behaviour in [distributed optimisation's tutorial page](https://optuna.readthedocs.io/en/stable/tutorial/10_key_features/004_distributed.html) or study.optimize's `n_trials`. Thank you for raising the issue and using Optuna for your OSS!

jkterry1:
-Why would that second line not be possible with this functionality?
-If this is a documentation failure, it seems like a rather severe one that you really might want to address

not522:
It is by design, not by the limitation of library implementation. I think if nothing happens on the second line, users will be confused.

Optuna also has the feature to load study. In the following case, I believe users expect 10 trials will execute.

```python
study = optuna.study.load_study(...)
study.optimize(n_trials=10)
```

The document seems to have room for improvement. To mention `MaxTrialsCallback` in study.optimize's `n_trials` will help with distributed optimization.

jkterry1:
bump for github actions bot since this needs to be fixed in the docs

PR #3951, opened 2022-08-31
making sure that n_trials will run in parallel mode

Ohadkaiser:
## Motivation
Suggesting solution to #3518 and #3938.
In this way you will be able to make sure that the number of trials that is delivered in the parameter n_trials will be exact when running in parallel. 

## Description of the changes
Added the ability to get the amount of trials in the study and to stop after the number of trials is reached

### DIFF
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -149,7 +149,7 @@ def _optimize(
     callbacks: Optional[List[Callable[["optuna.Study", FrozenTrial], None]]] = None,
     gc_after_trial: bool = False,
     show_progress_bar: bool = False,
-    optimize_function: Optional[Callable] = _optimize_sequential,
+    optimize_function: Callable = _optimize_sequential,
 ) -> None:
     if not isinstance(catch, tuple):
         raise TypeError(
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -99,7 +99,7 @@ def _optimize_parallel(
     reseed_sampler_rng: bool,
     time_start: Optional[datetime.datetime],
     progress_bar: Optional[pbar_module._ProgressBar],
-):
+) -> None:
     if reseed_sampler_rng:
         study.sampler.reseed_rng()
 
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -147,7 +147,7 @@ def _optimize_sequential(
             break
 
         if n_trials is not None:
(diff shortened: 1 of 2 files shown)


not522:
Your suggestion seems to break the following code. (Original: https://github.com/optuna/optuna/issues/2907#issuecomment-914610505)

```python
study.optimize(objective, n_trials=100) # not good objective value yet
study.optimize(objective, n_trials=100) # perform additional 100 trials to get a better value
```

I recommend not changing the behavior of existing user codes. Instead, it is better to add a new feature.

HideakiImamura:

optuna/study/_optimize.py
```diff
@@ -122,60 +229,6 @@ def _optimize(
...
-        if callbacks is not None:
-            for callback in callbacks:
-                callback(study, frozen_trial)
-
-        if progress_bar is not None:
-            progress_bar.update((datetime.datetime.now() - time_start).total_seconds())
-
-    study._storage.remove_session()
-
-
```
nzw0301, line 178: Could you revert this change? It makes reviews harder.

Ohadkaiser: I changed the order so that I could use `_optimize_sequential` as a default parameter for `_optimize` function

nzw0301:
TBH, I'm not sure that introducing a new public method: `optimize_parallel` and its private method that shares many lines with the existing method for this scenario is the best approach.

Ohadkaiser:
> TBH, I'm not sure that introducing a new public method: `optimize_parallel` and its private method that shares many lines with the existing method for this scenario is the best approach.

Do you have any suggestions that won't break the current optimize method implementation?

nzw0301:
I might still misunderstand the situation, but for me, [`MaxTrialsCallback`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.MaxTrialsCallback.html) with `states=None` is supposed to be enough.

Ohadkaiser:
> I might still misunderstand the situation, but for me, [`MaxTrialsCallback`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.MaxTrialsCallback.html) with `states=None` is supposed to be enough.

I'm trying to fix #3518, So it won't be enough in my opinion

nzw0301:
I've left comments on the original issue as in https://github.com/optuna/optuna/issues/3518#issuecomment-1236119954. Could you give some examples that my suggestion cannot resolve?

Issue #3670, opened 2022-06-14
Infinite loop in `TPESampler`


toshihikoyanase:
### Expected behavior

`TPESampler` should work with `n_startup_trials=1` when we apply it to multi-objective optimization problem.

### Environment

- Optuna version: 3.0.0rc0.dev
- Python version: 3.8.10
- OS: ubuntu 20.04 (on WSL2)
- (Optional) Other libraries and their versions: N/A

### Error messages, stack traces, or logs

```shell
The optimization stalled during the second trial of a study as follows:


[I 2022-06-14 22:05:14,674] A new study created in memory with name: no-name-b71eae7e-a922-42bf-a232-e096a8dbdeb7
[I 2022-06-14 22:05:14,675] Trial 0 finished with values: [-0.2716466405358364, -1.5373243468897773] and parameters: {'x': -0.2716466405358364, 'y': -1.5373243468897773}. 
/home/yanase/pfn/code/optuna-review/optuna/samplers/_tpe/sampler.py:585: RuntimeWarning: invalid value encountered in true_divide
  vecs = (vecs - lb) / (ub - lb)


When I hit `Ctrl + C`, the script stopped at the nondomination rank-based selection of the TPESampler.

```console
Traceback (most recent call last):
  File "/home/yanase/pfn/code/optuna-review/optuna/study/_optimize.py", line 196, in _run_trial
    value_or_values = func(trial)
...
  File "/home/yanase/pfn/code/optuna-review/optuna/samplers/_tpe/sampler.py", line 447, in sample_independent
    indices_below, indices_above = _split_observation_pairs(scores, self._gamma(n), violations)
  File "/home/yanase/pfn/code/optuna-review/optuna/samplers/_tpe/sampler.py", line 778, in _split_observation_pairs
    length = indices[nondomination_ranks == i].shape[0]
```

The corresponding code is shown below:
https://github.com/optuna/optuna/blob/d3c53e053a18341405cdfc9587d4525fdce7494d/optuna/samplers/_tpe/sampler.py#L774-L781

When the study has only a single trial, the default gamma function returns `1`, and the `n_below=1`. So the condition of the while loop at L777 is always `True`. So, the optimization process was trapped in the while loop.

On the other hand, the single objective TPE can select `below` even if we set `n_startup_trials=1`. I'm not sure, but we can fix the multi-objective logic by referring to the single objective one.

### Steps to reproduce

1. Execute the following python script.

```python
import optuna

def objective(trial):
[...]

PR #3953, opened 2022-09-01
Fix infinite loop bug in `TPESampler`

gasin:
## Motivation
In some special cases, `TPESampler` did not halt.
https://github.com/optuna/optuna/issues/3670

## Description of the changes
- Fix the flaw of the logic
    - See https://github.com/optuna/optuna/issues/3670#issuecomment-1232963093
    - The behavior changes only when `last_idx == n_below && n_below == len(indices)`
        - the case which caused an infinite loop 
- Add tests for multi-objective cases of `_split_observation_pairs`
    -  One is a base case, and the other is a corner case that had caused an infinite loop.

## Check
In https://github.com/optuna/optuna/issues/3670, it is mentioned that the bug occurs when `n_startup_trials=0` or `n_startup_trials=1`.
I checked the code below works successfully.
```python
import optuna


def objective(trial):
    return x, y

study_0 = optuna.create_study(
    directions=["maximize", "maximize"],
    sampler=optuna.samplers.TPESampler(n_startup_trials=0),
)
    
study_1 = optuna.create_study(
    directions=["maximize", "maximize"],
    sampler=optuna.samplers.TPESampler(n_startup_trials=1),
)

study_0.optimize(objective, n_trials=10)
for trial_0 in study_0.best_trials:
    print(trial_0.values)

study_1.optimize(objective, n_trials=10)
for trial_1 in study_1.best_trials:
    print(trial_1.values)
```

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -748,7 +748,7 @@ def _split_observation_pairs(
         # Nondomination rank-based selection
         i = 0
         last_idx = 0
-        while last_idx + sum(nondomination_ranks == i) <= n_below:
+        while last_idx < n_below and last_idx + sum(nondomination_ranks == i) <= n_below:
             length = indices[nondomination_ranks == i].shape[0]
             indices_below[last_idx : last_idx + length] = indices[nondomination_ranks == i]
             last_idx += length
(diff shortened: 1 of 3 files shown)


not522:
Thank you! Your fix looks good to me. Could you move the tests to `tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py`?

HideakiImamura:

contramundum53 (approved):
Good catch! LGTM!

gasin:
> Thank you! Your fix looks good to me. Could you move the tests to `tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py`?

Thank you for the comments.
I moved the tests https://github.com/optuna/optuna/pull/3953/commits/a6954cfee8ecd07a05a9528957ef11a56ba3ac1c.

not522 (approved):
LGTM!

Issue #3836, opened 2022-07-31
Use the latest stable scipy


nzw0301:
### Motivation

The current optuna is not compatibly with the latest stable scipy as mentioned in https://github.com/optuna/optuna/pull/3834. 

### Description

The expected changes might be as follows:

- Remove the upper version constraint of scipy instrodcued by https://github.com/optuna/optuna/pull/3834  
- Fix tests (see https://github.com/optuna/optuna/runs/7597875494?check_suite_focus=true too).

nzw0301:
I think the minimal example as follows extracted from https://github.com/optuna/optuna/blob/5d19e5e1f5dd9b3f9a11c74d215bd2a9c7ff43d2/optuna/samplers/_qmc.py#L281-L298

```python
from scipy.stats import qmc

sampler = qmc.Sobol(d=2, scramble=False)
_ = sampler.fast_forward(0)
```

The error message is as follows:

```shell
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/Caskroom/miniconda/base/envs/py38/lib/python3.8/site-packages/scipy/stats/_qmc.py", line 1586, in fast_forward
    _fast_forward(
  File "_sobol.pyx", line 334, in scipy.stats._sobol.__pyx_fuse_0_fast_forward
OverflowError: can't convert negative value to npy_uint32
```

Note that this does not happen with `qmc.Halton` instead of `qmc.Sobol`.

ekungurov:
Is optuna compatible with SciPy 3.9.x? Both PyPi and Conda have builds with scipy<3.9 dependency

nzw0301:
@ekungurov it will be compatible with the forthcoming Optuna 3.1.0.

PR #3959, opened 2022-09-02
Use the latest stable scipy

gasin:
## Motivation
We want to use the latest stable scipy by removing CI errors.
https://github.com/optuna/optuna/issues/3836

## Description of the changes
As described in https://github.com/optuna/optuna/issues/3836#issuecomment-1211817560, errors occur when we try to use `fast_forward` method of `scipy.stats._qmc.Sobol`.
However, that happens only if the argument of the `fast_forward` method is `0`, which has no effect on `scipy.stats._qmc.Sobol`.
Therefore, simply skipping the `fast_forward` method when its argument is `0` will resolve the problem.

## CHECK
By CI.

### DIFF
--- a/optuna/samplers/_qmc.py
+++ b/optuna/samplers/_qmc.py
@@ -294,6 +294,8 @@ class QMCSampler(BaseSampler):
             raise ValueError("Invalid `qmc_type`")
 
         forward_size = sample_id  # `sample_id` starts from 0.
+        # Skip fast_forward with forward_size==0 because Sobol doesn't support the case,
+        # and fast_forward(0) doesn't affect sampling.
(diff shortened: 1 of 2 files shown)


nzw0301 (changes requested):
Brilliant! The changes look great to me. Let me leave one comment on this PR: how about adding a comment like 
`Skip fast_forward with forward_size==0 because Sobol doesn't support the case, and fast_forward(0) doesn't affect sampling.` since this condition is not trivial for others?

not522 (approved):
Thank you! LGTM.

nzw0301 (approved):
Awesome. Thanks!

PR #3926, opened 2022-08-24
Add version constraint for `typing_extensions` to use `ParamSpec`

nzw0301:
## Motivation

Optuna uses `typing_extensions.ParamSpec` that has been available since `typing_extensions` v 3.10.0.0 (specifically, https://github.com/python/typing_extensions/commit/b697a12f2793655db99dde660bb47458ad36aa55). When a user upgrades their oputna version to v3 with older `typing_extensions`: `typing_extensions<3.10.0.0`, the user cannot import optuna due to using an older `typing_extensions` as reported in https://github.com/optuna/optuna/issues/3558.

## Description of the changes

Add version specification to use appropriate `typing_extensions`.

PR #3980, opened 2022-09-09
Support older `typing_extension` versions

c-bata:
## Motivation

We set the version constraint of typing_extensions `typing_extensions>=3.10.0.0` at #3926. However, `typing_extension==3.10.0.0` is not too older version since it was released at May 2021. So it may causes the installation issues (e.g. We cannot install `tensorflow=2.5.0` with `typing_extensions>=3.10.0.0`).

## Description of the changes

If `typing_extensions.ParamSpec` is unavailable, use `typing.Any` instead.

### DIFF
--- a/optuna/_convert_positional_args.py
+++ b/optuna/_convert_positional_args.py
@@ -12,7 +12,7 @@ try:
 
     _P = ParamSpec("_P")
 except ImportError:
-    _P = Any
+    pass
 
 
 _T = TypeVar("_T")
@@ -22,7 +22,7 @@ def convert_positional_args(
     *,
     previous_positional_arg_names: Sequence[str],
     warning_stacklevel: int = 2,
-) -> Callable[[Callable[_P, _T]], Callable[_P, _T]]:
+) -> "Callable[[Callable[_P, _T]], Callable[_P, _T]]":
     """Convert positional arguments to keyword arguments.
 
     Args:
@@ -30,7 +30,7 @@ def convert_positional_args(
         warning_stacklevel: Level of the stack trace where decorated function locates.
     """
 
-    def converter_decorator(func: Callable[_P, _T]) -> Callable[_P, _T]:
+    def converter_decorator(func: "Callable[_P, _T]") -> "Callable[_P, _T]":
 
         assert set(previous_positional_arg_names).issubset(set(signature(func).parameters)), (
             f"{set(previous_positional_arg_names)} is not a subset of"
--- a/optuna/_convert_positional_args.py
+++ b/optuna/_convert_positional_args.py
@@ -6,11 +6,16 @@ from typing import Sequence
 from typing import TypeVar
 import warnings
 
-from typing_extensions import ParamSpec
+
+try:
+    from typing_extensions import ParamSpec
+
+    _P = ParamSpec("_P")
+except ImportError:
+    _P = Any
 
 
 _T = TypeVar("_T")
-_P = ParamSpec("_P")
 
 
 def convert_positional_args(
--- a/optuna/_deprecated.py
+++ b/optuna/_deprecated.py
@@ -7,14 +7,19 @@ from typing import TypeVar
 import warnings
 
 from packaging import version
-from typing_extensions import ParamSpec
 
 from optuna._experimental import _get_docstring_indent
 from optuna._experimental import _validate_version
 
 
+try:
+    from typing_extensions import ParamSpec
+
+    FP = ParamSpec("FP")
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -39,7 +39,7 @@ def get_install_requires() -> List[str]:
         "scipy!=1.4.0" if sys.version[:3] == "3.6" else "scipy>=1.7.0",
         "sqlalchemy>=1.1.0",
         "tqdm",
-        "typing_extensions>=3.10.0.0",
+        "typing_extensions",
```
c-bata, line 42: We should set the lower limit actually. I'll work on it at #3974.

c-bata:

optuna/_convert_positional_args.py
```diff
@@ -6,11 +6,16 @@
...
 import warnings
 
-from typing_extensions import ParamSpec
+
+try:
+    from typing_extensions import ParamSpec
+
+    _P = ParamSpec("_P")
+except ImportError:
+    _P = Any
```
toshihikoyanase, line 15: When I execute `import optuna` without `typing-extensions`, I got the following error:
```
  File "optuna/optuna/_experimental.py", line 48, in <module>
    ) -> Callable[[Callable[FP, FT]], Callable[FP, FT]]:
  File "/usr/lib/python3.8/typing.py", line 813, in __getitem__
    raise TypeError(f"Callable[args, result]: args must be a list."
TypeError: Callable[args, result]: args must be a list. Got typing.Any
```

I guess `Callable[args, result]` requirs a list as args and `Any` is not acceptable here.

```suggestion
    _P = [Any]
```

optuna/_deprecated.py
```diff
@@ -7,14 +7,19 @@
...
 from optuna._experimental import _get_docstring_indent
 from optuna._experimental import _validate_version
 
 
+try:
+    from typing_extensions import ParamSpec
+
+    FP = ParamSpec("FP")
+except ImportError:
+    FP = Any
```
toshihikoyanase, line 20: ditto
```suggestion
    FP = [Any]
```

optuna/_experimental.py
```diff
@@ -6,13 +6,18 @@
...
-
 from optuna.exceptions import ExperimentalWarning
 
 
+try:
+    from typing_extensions import ParamSpec
+
+    FP = ParamSpec("FP")
+except ImportError:
+    FP = Any
```
toshihikoyanase, line 17: ditto
```suggestion
    FP = [Any]
```

optuna/integration/pytorch_distributed.py
```diff
@@ -23,8 +21,15 @@
     import torch.distributed as dist
 
 
+try:
+    from typing_extensions import ParamSpec
+
+    _P = ParamSpec("_P")
+except ImportError:
+    _P = Any
```
toshihikoyanase, line 29: ditto
```suggestion
    _P = [Any]
```

toshihikoyanase (commented):
I got errors when I import `optuna` without typing-extensions. Could you take a look?

setup.py
contramundum53: [Comment] Do we really need the lower limit? I think this dependency is optional and we could leave the lower limit unspecified.

c-bata: Thank you for your comment. You are right. We can move typing_extensions to optional dependencies like #3990.

optuna/_convert_positional_args.py
c-bata: Thank you for your review :bow:
I pushed 6f3b807 to fix the problem.

c-bata:
@toshihikoyanase @contramundum53 Thank you for your review :bow:

To avoid problems at runtime, I specified the type hints as strings. However, on second thought, I don't think it is necessary to list `typing_extensions` in `install_requires` if these annotations are not evaluated. I've just opened #3990 as an alternative approach. What do you think?

toshihikoyanase:
> I don't think it is necessary to list `typing_extensions` in `install_requires` in this approach. I opened https://github.com/optuna/optuna/pull/3990 as an alternative approach. It would be better to close this PR and review https://github.com/optuna/optuna/pull/3990 perhaps. What do you think?

Sounds reasonable. I roughly checked the #3990 and it looks promising. I think `ParamSpec` was introduced mainly for code suggestion of IDE (i.e., VS Code and PyCharm), so I'll check the behavior of them tomorrow.

contramundum53:
I agree that #3990 is better. Let me close this PR first.

PR #3990, opened 2022-09-12
Make typing_extensions optional.

c-bata:
## Motivation

An alternative approach for #3980 to fix `typing_extensions` related issues.

## Description of the changes

Check the `typing.TYPE_CHECKING` flag to check if type checking is enabled.

### DIFF
--- a/optuna/_convert_positional_args.py
+++ b/optuna/_convert_positional_args.py
@@ -3,21 +3,23 @@ from inspect import signature
 from typing import Any
 from typing import Callable
 from typing import Sequence
+from typing import TYPE_CHECKING
 from typing import TypeVar
 import warnings
 
-from typing_extensions import ParamSpec
 
+if TYPE_CHECKING:
+    from typing_extensions import ParamSpec
 
-_T = TypeVar("_T")
-_P = ParamSpec("_P")
+    _P = ParamSpec("_P")
+    _T = TypeVar("_T")
 
 
 def convert_positional_args(
     *,
     previous_positional_arg_names: Sequence[str],
     warning_stacklevel: int = 2,
-) -> Callable[[Callable[_P, _T]], Callable[_P, _T]]:
+) -> "Callable[[Callable[_P, _T]], Callable[_P, _T]]":
     """Convert positional arguments to keyword arguments.
 
     Args:
@@ -25,7 +27,7 @@ def convert_positional_args(
         warning_stacklevel: Level of the stack trace where decorated function locates.
     """
 
-    def converter_decorator(func: Callable[_P, _T]) -> Callable[_P, _T]:
+    def converter_decorator(func: "Callable[_P, _T]") -> "Callable[_P, _T]":
 
         assert set(previous_positional_arg_names).issubset(set(signature(func).parameters)), (
             f"{set(previous_positional_arg_names)} is not a subset of"
@@ -33,7 +35,7 @@ def convert_positional_args(
         )
 
         @wraps(func)
-        def converter_wrapper(*args: Any, **kwargs: Any) -> _T:
+        def converter_wrapper(*args: Any, **kwargs: Any) -> "_T":
 
             if len(args) >= 1:
                 warnings.warn(
--- a/optuna/_deprecated.py
+++ b/optuna/_deprecated.py
@@ -3,19 +3,22 @@ import textwrap
 from typing import Any
 from typing import Callable
 from typing import Optional
+from typing import TYPE_CHECKING
 from typing import TypeVar
 import warnings
 
 from packaging import version
-from typing_extensions import ParamSpec
 
 from optuna._experimental import _get_docstring_indent
 from optuna._experimental import _validate_version
 
 
-FT = TypeVar("FT")
-FP = ParamSpec("FP")
-CT = TypeVar("CT")
+if TYPE_CHECKING:
+    from typing_extensions import ParamSpec
+
+    FT = TypeVar("FT")
+    FP = ParamSpec("FP")
+    CT = TypeVar("CT")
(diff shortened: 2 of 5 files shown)


contramundum53:
I removed `typing_extension` and the tests ran successfully.
(Except for `importance_tests` and `sampler_tests/test_samplers.py` which use `torch` internally.)
LGTM.

contramundum53 (approved):

toshihikoyanase:
I don't think this is a blocker of this PR, but I found that `typing-extensions` was installed along with `SQLAlchemy` in Python 3.6 and 3.7. Without `typing-extensions`, users cannot import optuna due to the following errors.

```console
$ python -c "import optuna"
  File "/usr/local/lib/python3.6/site-packages/importlib_metadata/_compat.py", line 9, in <module>
    from typing import Protocol
ImportError: cannot import name 'Protocol'

...
  File "/proj/optuna/storages/_rdb/storage.py", line 27, in <module>
    from optuna.storages._rdb.models import TrialValueModel
  File "/proj/optuna/storages/_rdb/models.py", line 8, in <module>
    from sqlalchemy import asc
...
  File "/usr/local/lib/python3.6/site-packages/importlib_metadata/_meta.py", line 1, in <module>
    from ._compat import Protocol
  File "/usr/local/lib/python3.6/site-packages/importlib_metadata/_compat.py", line 11, in <module>
    from typing_extensions import Protocol  # type: ignore
ModuleNotFoundError: No module named 'typing_extensions'
```

The reproducible code is as follows:

1. Launch a container of the specific version of the python docker image.
```console
$  git clone https://github.com/c-bata/optuna.git -b make-typing-extensions-optional
$ cd optuna
$ MINOR_VERSION=7 
$ docker run -it --rm -v $(pwd):/proj --workdir=/proj python:3.$MINOR_VERSION bash
```

2. Execute following commands
```console
$ python --version
$ pip install -U pip
$ pip install .
$ python -c "import optuna"
$ pip freeze | grep typing
$ pip uninstall typing-extensions
$ python -c "import optuna"
```

The results are shown in the following table:

| Python | typing-extensions | `import optuna` without `typing-extensions` | 
| --- | --- | --- |
|  3.6.15 |  4.1.1 | `ImportError` |
|  3.7.12 | 4.3.0 | `ImportError` |
| 3.8.12 | not installed | ✔️  |
| 3.9.13 | not installed | ✔️ |
| 3.10.4 | not installed | ✔️ |

toshihikoyanase:
>  (e.g. We cannot install tensorflow=2.5.0 with typing_extensions>=3.10.0.0).

With this PR, we can use `tensorflow=2.5.0` and optuna, if we install `tensorflow` in prior to `optuna`.

```
root@fee7c0b17cda:/proj# pip freeze | grep tensorflow
tensorflow==2.5.0
tensorflow-estimator==2.5.0
root@fee7c0b17cda:/proj# pip freeze | grep typing
typing-extensions==3.7.4.3
root@fee7c0b17cda:/proj# python -c "import optuna"
```

c-bata:
@toshihikoyanase Thank you for your review!
It looks like the issue of SQLAlchemy. I checked that typing-extensions are specified in `install_requires` for each version of SQLAlchemy. The results are as follows:

* In the latest release SQLAlchemy v1.4.41, `typing_extensions` is not listed in `install_requires`.
    * https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_4_41/setup.cfg
* In the `main` branch, `typing_extensions>=4.1.0` is listed in `install_requires`. So this issue will be resolved at the next release.
    * https://github.com/sqlalchemy/sqlalchemy/blob/main/setup.cfg#L40

Even though Optuna's RDBStorage users may bump into this problem, I don't think `typing-extensions` should be included in Optuna's `install_requires`. What do you think?

toshihikoyanase:
> Even though Optuna's RDBStorage users may bump into this problem, I don't think typing-extensions should be included in Optuna's install_requires. What do you think?

I agree with you. With this PR, the `typing_extensions` will not be a dependency of Optuna, so I don't think we have to keep it in the requirements. In addition, users have workaround to use libraries that requires older version of `typing_extensions`, so the original issue seems to be resolved. I'll approve this PR after checking the behavior of code suggestion features of IDE (i.e., VS Code).

toshihikoyanase:
> In the latest release SQLAlchemy v1.4.41, typing_extensions is not listed in install_requires.
> https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_4_41/setup.cfg

Actually, it indirectly depends on `typing_extensions` via `importlib-metadata`.
- https://github.com/sqlalchemy/sqlalchemy/blob/ac9a74ac0d60f5e95d9dbbce7f804f532bef2bea/setup.cfg#L42-L43
- https://github.com/python/importlib_metadata/blob/53e42d6d70bd2068cd5bf0d351df89aa3cc0af31/setup.cfg#L21

And also, `PrettyTable` required by `cliff` also depends on `importlib-metadata`. So, some other libraries might depends on `typing_extensions` and it may be difficult to remove `typing_extensions` from the indirect dependencies.

toshihikoyanase:
In my environment, the code suggestion of VS Code worked as expected. I cannot understand why, but it worked even if `typing_extensions` was not installed in the virtual environment.

The code suggestion for the deprecated method (i.e., `trial.suggest_uniform`) is shown as follows:
image

The code suggestion for a method with `@convert_positional_args` (i.e., `optuna.create_study`) is shown as follows:
image

The code suggestion for an experimental method (i.e., `optuna.storages.RetryFailedTrialCallback`) is shown as follows:
image

toshihikoyanase (approved):
LGTM.

PR #3993, opened 2022-09-12
Use `log_loss` instead of deprecated `log` since `sklearn` 1.1

nzw0301:
## Motivation

The `loss` argument of skelarn's [`SGDClassifier`](https://scikit-learn.org/stable/modules/classes.html#module-sklearn.linear_model) reads

> Deprecated since version 1.1: The loss ‘log’ was deprecated in v1.1 and will be removed in version 1.3. Use loss='log_loss' which is equivalent.

By this deprecation, the output of [a tutorial](https://optuna.readthedocs.io/en/stable/tutorial/20_recipes/006_user_defined_pruner.html#sphx-glr-tutorial-20-recipes-006-user-defined-pruner-py) looks a little bit messy as follows:

Screenshot 2022-09-12 at 21 41 20

## Description of the changes

By following sklearn, replace `"log"` with `"log_loss"` in the code of the tutorial.

### DIFF
--- a/tutorial/20_recipes/006_user_defined_pruner.py
+++ b/tutorial/20_recipes/006_user_defined_pruner.py
@@ -108,7 +108,7 @@ def objective(trial):
         iris.data, iris.target, train_size=100, test_size=50, random_state=0
     )
 
-    loss = trial.suggest_categorical("loss", ["hinge", "log", "perceptron"])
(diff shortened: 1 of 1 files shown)


contramundum53 (approved):
LGTM! Thanks!

PR #4016, opened 2022-09-26
Support CMA-ES with margin in `CmaEsSampler`

knshnb:
## Motivation

Support [CMA-ES with margin](https://arxiv.org/abs/2205.13482), which lower-bounds the marginal probability associated with each discrete space so that samples can avoid being fixed to a single point.

In Optuna, `IntDistribution` and `FloatDistribution` with `step` can be regarded as discrete search spaces. I supported the combination with the `source_trials` argument, but currently, it cannot be used with `use_separable_cma=True`.

In the benchmarks of the HPO bench, it outperformed the original CMA-ES when the number of trials is large (200-400+) ([link](https://github.com/CyberAgentAILab/cmaes/pull/121#issuecomment-1296691448)).

## Description of the changes

- [x] Add `with_margin` argument to `CmaEsSampler`.
- [x] Use `CMAwM` class, which was released in [v0.9.0 of cmaes](https://github.com/CyberAgentAILab/cmaes/releases/tag/v0.9.0).
- [x] Add tests.


knshnb:
Now it's ready for review!

optuna/samplers/_cmaes.py
```diff
@@ -177,6 +180,15 @@ def objective(trial):
                 versions without prior notice. See
                 https://github.com/optuna/optuna/releases/tag/v2.6.0.
 
+        with_margin:
+            If this is :obj:`True`, CMA-ES with margin is used.
```
c-bata, line 184: How about adding the paper of CMA-ES with Margin in the following list?

https://github.com/optuna/optuna/blob/23c874764b2be969ebe19c86135614a513ff8ae4/optuna/samplers/_cmaes.py#L77-L91

c-bata: Furthermore, how about adding the description about when Optuna users should use this option?

optuna/samplers/_cmaes.py
```diff
@@ -287,13 +314,7 @@ def infer_relative_search_space(
...
                 continue
 
-            if not isinstance(
-                distribution,
-                (
-                    optuna.distributions.FloatDistribution,
-                    optuna.distributions.IntDistribution,
-                ),
-            ):
+            if not isinstance(distribution, (FloatDistribution, IntDistribution)):
```
c-bata, line 317: Do you think we can leave a todo comment about the support of `CategoricalDistribution`, which contains only two choices?

optuna/samplers/_cmaes.py
```diff
@@ -177,6 +180,15 @@ def objective(trial):
                 versions without prior notice. See
                 https://github.com/optuna/optuna/releases/tag/v2.6.0.
 
+        with_margin:
+            If this is :obj:`True`, CMA-ES with margin is used.
+            Currently, this option cannot be used with ``use_separable_cma=True``.
```
c-bata, line 185: Also this option cannot be used with `source_trials`.

c-bata (commented):
Thank you for your pull request!
I took a quick look and left some comments. Let me review the implementation details later.

optuna/samplers/_cmaes.py
knshnb: Thanks! I updated the docstring according to your comment.

optuna/samplers/_cmaes.py
c-bata: I asked @nomuramasahir0 about this topic, and then he said "It's possible in theory. The thorough performance analysis has not been conducted though".

Therefore, seemingly, the use of the source_trials and with_margin options should not be prohibited on the interface. Let me resolve this comment :bow:

c-bata:
@nomuramasahir0 Could you review this PR if you have time?

nomuramasahir0:
Thank you for the PR about CMA-ES with Margin. LGTM!

optuna/samplers/_cmaes.py
```diff
@@ -484,6 +519,26 @@ def _init_optimizer(
                 population_size=population_size,
             )
 
+        if self._with_margin:
+            steps = []
+            for dist in trans._search_space.values():
+                assert isinstance(dist, (IntDistribution, FloatDistribution))
+                # Set step 0.0 for continuous search space.
+                steps.append(0.0 if dist.step is None else dist.step)
```
c-bata, line 527: nit:

```suggestion
            # Set step 0.0 for continuous search space.
            steps = [dist.step or 0.0 for dist in trans._search_space.values()]
```

optuna/samplers/_cmaes.py
```diff
@@ -484,6 +519,26 @@ def _init_optimizer(
...
+                return CMAwM(
+                    mean=mean,
+                    sigma=sigma0,
+                    bounds=trans.bounds,
+                    steps=np.array(steps, dtype=float),
+                    cov=cov,
+                    seed=self._cma_rng.randint(1, 2**31 - 2),
+                    n_max_resampling=10 * n_dimension,
+                    population_size=population_size,
+                )
```
c-bata, line 540: We can do early return like this:

```suggestion
        # If there is no discrete search space, we use `CMA` because CMAwM` throws an error.
        if self._with_margin and any(dist.step is not None for dist in trans._search_space.values()):
            # Set step 0.0 for continuous search space.
            steps = [dist.step or 0.0 for dist in trans._search_space.values()]
            return CMAwM(
                mean=mean,
                sigma=sigma0,
                bounds=trans.bounds,
                steps=np.array(steps, dtype=float),
                cov=cov,
                seed=self._cma_rng.randint(1, 2**31 - 2),
                n_max_resampling=10 * n_dimension,
                population_size=population_size,
            )
```

optuna/samplers/_cmaes.py
knshnb: This does not work because `BaseDistribution` does not have the "step" attribute (mypy throws an error).

c-bata: Ah, I see.

optuna/samplers/_cmaes.py
c-bata, line 540: nit: Perhaps either would be fine, but left one minor suggestion.

```suggestion
            steps = np.empty(len(trans._search_space), dtype=float)
            for i, dist in enumerate(trans._search_space.values()):
                assert isinstance(dist, (IntDistribution, FloatDistribution))
                # Set step 0.0 for continuous search space.
                steps[i] = dist.step or 0.0

            # If there is no discrete search space, we use `CMA` because CMAwM` throws an error.
            if np.any(steps > 0.0):
                return CMAwM(
                    mean=mean,
                    sigma=sigma0,
                    bounds=trans.bounds,
                    steps=steps,
                    cov=cov,
                    seed=self._cma_rng.randint(1, 2**31 - 2),
                    n_max_resampling=10 * n_dimension,
                    population_size=population_size,
                )
```

c-bata (approved):
LGTM with nits!

Although there might be the same issue discussed at https://github.com/optuna/optuna/pull/4142#discussion_r1023673421, we should merge this PR first and it should be addressed as a follow-up with the sep-CMA-ES issue.

PR #4015, opened 2022-09-26
Skip to add `intermediate_value_type` and `value_type` columns if exists

c-bata:
## Motivation

The data migration written in `v3.0.0.c.py` is sometimes failed if there are a lot of `trial_intermediate_values` records since  it takes long time. At that time, users cannot resume the database migration.

## Description of the changes

Skip to add `intermediate_value_type` and `value_type` columns if exists to make the migration idempotent.

PR #4020, opened 2022-09-27
Fix the data migration script for v3 release.

c-bata:
## Motivation

Follow-up https://github.com/optuna/optuna/pull/4015

The data migration written in v3.0.0.c.py takes a long time if there are a lot of `trial_intermediate_values` records.

## Description of the changes

Make the data migration scripts `v3.0.0.c.py` and `v3.0.0.d.py` faster.

### optuna-e2e scripts

I tested this change with the following optuna-e2e's scripts.
https://github.com/c-bata/optuna-e2e/tree/optuna-303-data-migration

### About `nan` support

We may omit `nan` support since `trial.report()` accepts `nan` value from [v3.0.0-b1](https://github.com/optuna/optuna/pull/3348) and `v3.0.0.c.py` script is also added at [v3.0.0-b1](https://github.com/optuna/optuna/pull/3564). The users do not store `nan` value in their databases unless Optuna is directly installed from GitHub, not PyPI. I added support for such cases in this PR though.

### DIFF
--- a/optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
+++ b/optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
@@ -95,6 +95,7 @@ def upgrade():
                 sa.or_(
                     IntermediateValueModel.intermediate_value > RDB_MAX_FLOAT - 1,
                     IntermediateValueModel.intermediate_value < RDB_MIN_FLOAT + 1,
+                    IntermediateValueModel.intermediate_value.is_(None),
                 )
             )
             .all()
@@ -102,7 +103,9 @@ def upgrade():
         mapping = []
         for r in records:
             value: float
-            if np.isclose(r.intermediate_value, RDB_MAX_FLOAT) or np.isposinf(
+            if r.intermediate_value is None or np.isnan(r.intermediate_value):
+                value = float("nan")
+            elif np.isclose(r.intermediate_value, RDB_MAX_FLOAT) or np.isposinf(
                 r.intermediate_value
             ):
                 value = float("inf")
@@ -110,8 +113,6 @@ def upgrade():
                 r.intermediate_value
             ):
                 value = float("-inf")
-            elif np.isnan(r.intermediate_value):
-                value = float("nan")
             else:
                 value = r.intermediate_value
             (
--- a/optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
+++ b/optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
@@ -89,7 +89,16 @@ def upgrade():
 
     session = orm.Session(bind=bind)
     try:
-        records = session.query(IntermediateValueModel).all()
+        records = (
+            session.query(IntermediateValueModel)
+            .filter(
+                sa.or_(
+                    IntermediateValueModel.intermediate_value > RDB_MAX_FLOAT - 1,
+                    IntermediateValueModel.intermediate_value < RDB_MIN_FLOAT + 1,
+                )
+            )
+            .all()
+        )
         mapping = []
         for r in records:
             value: float
--- a/optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
+++ b/optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
@@ -93,8 +93,8 @@ def upgrade():
             session.query(IntermediateValueModel)
             .filter(
                 sa.or_(
-                    IntermediateValueModel.intermediate_value > RDB_MAX_FLOAT - 1,
-                    IntermediateValueModel.intermediate_value < RDB_MIN_FLOAT + 1,
+                    IntermediateValueModel.intermediate_value > 1e16,
+                    IntermediateValueModel.intermediate_value < -1e16,
                     IntermediateValueModel.intermediate_value.is_(None),
                 )
             )
--- a/optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
+++ b/optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
@@ -102,7 +102,16 @@ def upgrade():
 
     session = orm.Session(bind=bind)
     try:
-        records = session.query(TrialValueModel).all()
+        records = (
+            session.query(TrialValueModel)
+            .filter(
+                sa.or_(
+                    TrialValueModel.value > RDB_MAX_FLOAT - 1,
+                    TrialValueModel.value < RDB_MIN_FLOAT + 1,
+                )
+            )
+            .all()
+        )
         mapping = []
         for r in records:
             value: float
--- a/optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
+++ b/optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
@@ -108,7 +108,6 @@ def upgrade():
                 sa.or_(
                     TrialValueModel.value > 1e16,
                     TrialValueModel.value < -1e16,
-                    TrialValueModel.value.is_(None),
                 )
             )
             .all()
--- a/optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
+++ b/optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
@@ -108,6 +108,7 @@ def upgrade():
                 sa.or_(
                     TrialValueModel.value > RDB_MAX_FLOAT - 1,
                     TrialValueModel.value < RDB_MIN_FLOAT + 1,
+                    TrialValueModel.value.is_(None),
                 )
             )
             .all()
--- a/optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
+++ b/optuna/storages/_rdb/alembic/versions/v3.0.0.d.py
@@ -106,8 +106,8 @@ def upgrade():
             session.query(TrialValueModel)
             .filter(
                 sa.or_(
-                    TrialValueModel.value > RDB_MAX_FLOAT - 1,
-                    TrialValueModel.value < RDB_MIN_FLOAT + 1,
+                    TrialValueModel.value > 1e16,
+                    TrialValueModel.value < -1e16,
                     TrialValueModel.value.is_(None),
                 )
             )


optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
```diff
@@ -89,20 +89,30 @@ def upgrade():
 
     session = orm.Session(bind=bind)
     try:
-        records = session.query(IntermediateValueModel).all()
+        records = (
+            session.query(IntermediateValueModel)
+            .filter(
+                sa.or_(
+                    IntermediateValueModel.intermediate_value > 1e16,
+                    IntermediateValueModel.intermediate_value < -1e16,
```
c-bata, line 97: `1e16` is a heuristic number to filter the target records.

Please note that I tried using `RDB_MAX_FLOAT - 1` instead of `1e16` here (299d2379bdebe9f77a5a64c7361e90c76cfdb617), but it did not work actually. So I use `1e16` as a heuristic threshold that can filter most of intermediate values.

optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
```diff
@@ -89,20 +89,30 @@ def upgrade():
...
+                    IntermediateValueModel.intermediate_value.is_(None),
```
c-bata, line 98: I confirmed that `nan` values are stored as NULL on `SQLite3`, `MySQL` and `PostgreSQL` with [optuna-e2e](https://github.com/c-bata/optuna-e2e/tree/optuna-303-data-migration).

<details>
<summary>SQLite3</summary>

```
$ sqlite3 ./data/sample.db
sqlite> .mode column
sqlite> .header on
sqlite> select * from trial_intermediate_values;
trial_intermediate_value_id  trial_id  step  intermediate_value
---------------------------  --------  ----  ---------------------
1                            1         1     3.40282346638529e+38
2                            1         2     -3.40282346638529e+38
3                            1         3
4                            2         1     3.40282346638529e+38
5                            2         2     -3.40282346638529e+38
6                            2         3
7                            3         1     3.40282346638529e+38
8                            3         2     -3.40282346638529e+38
9                            3         3
10                           4         1     3.40282346638529e+38
11                           4         2     -3.40282346638529e+38
12                           4         3
13                           5         1     3.40282346638529e+38
14                           5         2     -3.40282346638529e+38
15                           5         3
16                           6         1     3.40282346638529e+38
17                           6         2     -3.40282346638529e+38
18                           6         3
19                           7         1     3.40282346638529e+38
20                           7         2     -3.40282346638529e+38
21                           7         3
22                           8         1     3.40282346638529e+38
23                           8         2     -3.40282346638529e+38
24                           8         3
25                           9         1     3.40282346638529e+38
26                           9         2     -3.40282346638529e+38
27                           9         3
28                           10        1     3.40282346638529e+38
29                           10        2     -3.40282346638529e+38
30                           10        3

sqlite> select * from trial_intermediate_values where intermediate_value is null;
trial_intermediate_value_id  trial_id  step  intermediate_value
---------------------------  --------  ----  ------------------
3                            1         3
6                            2         3
9                            3         3
12                           4         3
15                           5         3
18                           6         3
21                           7         3
24                           8         3
27                           9         3
30                           10        3
```


</details>

<details>
<summary>MySQL</summary>

```
$ mysql --host 127.0.0.1 --port 3306 --user root -D optuna -proot

mysql> select * from trial_intermediate_values;
+-----------------------------+----------+------+--------------------+
| trial_intermediate_value_id | trial_id | step | intermediate_value |
+-----------------------------+----------+------+--------------------+
|                           1 |        1 |    1 |         3.40282e38 |
|                           2 |        1 |    2 |        -3.40282e38 |
|                           3 |        1 |    3 |               NULL |
|                           4 |        2 |    1 |         3.40282e38 |
|                           5 |        2 |    2 |        -3.40282e38 |
|                           6 |        2 |    3 |               NULL |
|                           7 |        3 |    1 |         3.40282e38 |
|                           8 |        3 |    2 |        -3.40282e38 |
|                           9 |        3 |    3 |               NULL |
|                          10 |        4 |    1 |         3.40282e38 |
|                          11 |        4 |    2 |        -3.40282e38 |
|                          12 |        4 |    3 |               NULL |
|                          13 |        5 |    1 |         3.40282e38 |
|                          14 |        5 |    2 |        -3.40282e38 |
|                          15 |        5 |    3 |               NULL |
|                          16 |        6 |    1 |         3.40282e38 |
|                          17 |        6 |    2 |        -3.40282e38 |
|                          18 |        6 |    3 |               NULL |
|                          19 |        7 |    1 |         3.40282e38 |
|                          20 |        7 |    2 |        -3.40282e38 |
|                          21 |        7 |    3 |               NULL |
|                          22 |        8 |    1 |         3.40282e38 |
|                          23 |        8 |    2 |        -3.40282e38 |
|                          24 |        8 |    3 |               NULL |
|                          25 |        9 |    1 |         3.40282e38 |
|                          26 |        9 |    2 |        -3.40282e38 |
|                          27 |        9 |    3 |               NULL |
|                          28 |       10 |    1 |         3.40282e38 |
|                          29 |       10 |    2 |        -3.40282e38 |
|                          30 |       10 |    3 |               NULL |
+-----------------------------+----------+------+--------------------+
30 rows in set (0.01 sec)
```

</details>

<details>
<summary>PostgreSQL</summary>

```
$ docker exec -it optuna-e2e-postgresql-1 /bin/bash
root@c7a20f532829:/# psql optuna
psql (12.10 (Debian 12.10-1.pgdg110+1))
Type "help" for help.

optuna=# select * from trial_intermediate_values;
 trial_intermediate_value_id | trial_id | step |   intermediate_value
-----------------------------+----------+------+-------------------------
                           1 |        1 |    1 |  3.4028234663852886e+38
                           2 |        1 |    2 | -3.4028234663852886e+38
                           3 |        1 |    3 |
                           4 |        2 |    1 |  3.4028234663852886e+38
                           5 |        2 |    2 | -3.4028234663852886e+38
                           6 |        2 |    3 |
                           7 |        3 |    1 |  3.4028234663852886e+38
                           8 |        3 |    2 | -3.4028234663852886e+38
                           9 |        3 |    3 |
                          10 |        4 |    1 |  3.4028234663852886e+38
                          11 |        4 |    2 | -3.4028234663852886e+38
                          12 |        4 |    3 |
                          13 |        5 |    1 |  3.4028234663852886e+38
                          14 |        5 |    2 | -3.4028234663852886e+38
                          15 |        5 |    3 |
                          16 |        6 |    1 |  3.4028234663852886e+38
                          17 |        6 |    2 | -3.4028234663852886e+38
                          18 |        6 |    3 |
                          19 |        7 |    1 |  3.4028234663852886e+38
                          20 |        7 |    2 | -3.4028234663852886e+38
                          21 |        7 |    3 |
                          22 |        8 |    1 |  3.4028234663852886e+38
                          23 |        8 |    2 | -3.4028234663852886e+38
                          24 |        8 |    3 |
                          25 |        9 |    1 |  3.4028234663852886e+38
                          26 |        9 |    2 | -3.4028234663852886e+38
                          27 |        9 |    3 |
                          28 |       10 |    1 |  3.4028234663852886e+38
                          29 |       10 |    2 | -3.4028234663852886e+38
                          30 |       10 |    3 |
(30 rows)

optuna=# select * from trial_intermediate_values where intermediate_value is NULL;
 trial_intermediate_value_id | trial_id | step | intermediate_value
-----------------------------+----------+------+--------------------
                           3 |        1 |    3 |
                           6 |        2 |    3 |
                           9 |        3 |    3 |
                          12 |        4 |    3 |
                          15 |        5 |    3 |
                          18 |        6 |    3 |
                          21 |        7 |    3 |
                          24 |        8 |    3 |
                          27 |        9 |    3 |
                          30 |       10 |    3 |
(10 rows)
```

</details>

optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
```diff
@@ -89,20 +89,30 @@ def upgrade():
...
+            elif np.isclose(r.intermediate_value, RDB_MAX_FLOAT) or np.isposinf(
                 r.intermediate_value
             ):
                 value = float("inf")
             elif np.isclose(r.intermediate_value, RDB_MIN_FLOAT) or np.isneginf(
                 r.intermediate_value
             ):
                 value = float("-inf")
-            elif np.isnan(r.intermediate_value):
-                value = float("nan")
```
c-bata, line 105: Seemingly this code is incorrect as summarized at https://github.com/optuna/optuna/pull/4020/files#r981124322. We may omit this case though.

optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
contramundum53: I think in this particular case we could use `>= RDB_MAX_FLOAT`. This would not introduce false positives.

optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
contramundum53: An already-converted `inf` values are also represented as `None`, so this would make the migration non-idempotent.

optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
c-bata: I don7t think we could use `>= RDB_MAX_FLOAT` since `> RDB_MAX_FLOAT - 1` didn't work. It may cause the representation error of floating point numbers.

contramundum53: `> RDB_MAX_FLOAT-1` doesn't work because `RDB_MAX_FLOAT - 1 == RDB_MAX_FLOAT` in double precision. `>= RDB_MAX_FLOAT` should work.

contramundum53:
I checked the script using optuna-e2e. The migration worked in the normal situation.

optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
toshihikoyanase: > I think in this particular case we could use `>= RDB_MAX_FLOAT`. This would not introduce false positives.

I  tried this approach, and I could migrate data as expected.

Note that some database including PostgreSQL and MSSQL do not accept numpy number type, so I needed to convert `RDB_MAX_FLOAT` to built-in float type with `RDB_MAX_FLOAT.item()`.

```console
[E 2022-10-04 07:55:35,159] (psycopg2.ProgrammingError) can't adapt type 'numpy.float32'
[SQL: SELECT trial_intermediate_values.trial_intermediate_value_id AS trial_intermediate_values_trial_intermediate_value_id, trial_intermediate_values.intermediate_value AS trial_intermediate_values_intermediate_value, trial_intermediate_values.intermediate_value_type AS trial_intermediate_values_intermediate_value_type 
FROM trial_intermediate_values 
WHERE trial_intermediate_values.intermediate_value >= %(intermediate_value_1)s OR trial_intermediate_values.intermediate_value <= %(intermediate_value_2)s OR trial_intermediate_values.intermediate_value IS NULL]
[parameters: {'intermediate_value_1': 3.4028235e+38, 'intermediate_value_2': -3.4028235e+38}]
(Background on this error at: https://sqlalche.me/e/14/f405)
```

Do you think if any databases make stored values smaller than original values, for example, due to numerical error or limitation of internal representation?

c-bata: Oh, I see. I'm sorry that my opinion seemed wrong. And thank you yanase-san for checking the approach.

I am not sure that this logic (`>= RDB_MAX_FLOAT`) will always work correctly in any database. While there may be cases where `>=RDB_MAX_FLOAT` could be used to further speed up the data migration, this heuristic `1e16` still seems reasonable. It doesn't matter if the value is sufficiently large to reduce the number of records in order to speed up data migration.

What do you think?

toshihikoyanase: Thank you for your comment. IMO, the current heuristic thresholds are acceptable since they just filter the record to accelerate the migration. The logic to convert `RDB_MAX_FLOAT` to `inf` is not changed in this PR.

contramundum53 (approved):
LGTM!

optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
```diff
@@ -94,20 +94,30 @@ def upgrade():
...
+                )
+            )
+            .all()
+        )
         mapping = []
         for r in records:
             value: float
-            if np.isclose(r.intermediate_value, RDB_MAX_FLOAT) or np.isposinf(
+            if r.intermediate_value is None or np.isnan(r.intermediate_value):
+                value = float("nan")
```
toshihikoyanase, line 112: @contramundum53 This line does not distinguish `None` and `nan`. I think this is reasonable according to [this comment](https://github.com/optuna/optuna/pull/4020/files#r981124322). I don't remember the discussion in ##3603, but do you have any concerns about this change?

toshihikoyanase (approved):
I also confirmed that the new migration script could migrate databases successfully. LGTM.

optuna/storages/_rdb/alembic/versions/v3.0.0.c.py
contramundum53: One concern is https://github.com/optuna/optuna/pull/4020/files#r981834790, where half-migrated `inf` values would be considered as `nan` values. However, we decided not to consider the case where the batch operation fails halfway, so I don't think that is a problem now.

toshihikoyanase: Thank you for your comment. That makes sense.

Issue #4028, opened 2022-09-29
Fix `test_reproducible_in_other_process` for fixed `PYTHONHASHSEED`


not522:
### Motivation

`test_reproducible_in_other_process` requires `PYTHONHASHSEED` is not set. If testing with the fixed hash seed, this test fails. For example, tox sets `PYTHONHASHSEED` https://tox.wiki/en/latest/example/basic.html#special-handling-of-pythonhashseed.

https://github.com/optuna/optuna/blob/e8abeec7d58b64b207bd7606056311d20de0241f/tests/samplers_tests/test_samplers.py#L997-L1037

### Description

If `PYTHONHASHSEED` is set, we should store the value, remove it from `os.environ` before calling the subprocess, and restore the original seed after the process.

g-votte:
@nzw0301 is working on this issue.

PR #4031, opened 2022-09-29
Unset `PYTHONHASHSEED` for the hash-depedenet test

nzw0301:
## Motivation

Resolve https://github.com/optuna/optuna/issues/4028, which describes the motivation quite well.

## Description of the changes

Unset the environmental variable `PYTHONHASHSEED`. I've conformed `tox` doesn't fail anymore by adding the following lines to `tox.ini`

```
[testenv:pytest]
extras =
    test
    optional
deps =
    scikit-optimize
    cma
    botorch
commands = pytest tests/samplers_tests/test_samplers.py::test_reproducible_in_other_process
```

and run the following command on the CLI: 

```bash
tox -e pytest
```

### DIFF
--- a/tests/samplers_tests/test_samplers.py
+++ b/tests/samplers_tests/test_samplers.py
@@ -1021,7 +1021,9 @@ def run_optimize(
 def test_reproducible_in_other_process(sampler_class_index: int) -> None:
     # This test should be tested without `PYTHONHASHSEED`. However, tox sets the
     # environmental variable "PYTHONHASHSEED" by default.
-    os.unsetenv("PYTHONHASHSEED")
+    hash_seed = os.getenv("PYTHONHASHSEED")
+    if hash_seed is not None:
+        del os.environ["PYTHONHASHSEED"]
 
     # Multiprocessing supports three way to start a process.
     # We use `spawn` option to create a child process as a fresh python process.
@@ -1040,3 +1042,6 @@ def test_reproducible_in_other_process(sampler_class_index: int) -> None:
     assert not (hash_dict[0] == hash_dict[1] == hash_dict[2])
     # But the sequences are expected to be the same.
     assert sequence_dict[0] == sequence_dict[1] == sequence_dict[2]
+
+    if hash_seed is not None:
+        os.environ["PYTHONHASHSEED"] = hash_seed
--- a/tests/samplers_tests/test_samplers.py
+++ b/tests/samplers_tests/test_samplers.py
@@ -1039,9 +1039,9 @@ def test_reproducible_in_other_process(sampler_class_index: int) -> None:
         p.start()
         p.join()
     # Hashes are expected to be different because string hashing is nondeterministic per process.
+    if hash_seed is not None:
+        os.environ["PYTHONHASHSEED"] = hash_seed
+
     assert not (hash_dict[0] == hash_dict[1] == hash_dict[2])
     # But the sequences are expected to be the same.
     assert sequence_dict[0] == sequence_dict[1] == sequence_dict[2]
-
-    if hash_seed is not None:
-        os.environ["PYTHONHASHSEED"] = hash_seed
--- a/tests/samplers_tests/test_samplers.py
+++ b/tests/samplers_tests/test_samplers.py
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 import multiprocessing
 from multiprocessing.managers import DictProxy
+import os
 import pickle
 import sys
 from typing import Any
@@ -1018,6 +1019,10 @@ def run_optimize(
 
 @pytest.mark.parametrize("sampler_class_index", range(len(sampler_class_with_seed)))
 def test_reproducible_in_other_process(sampler_class_index: int) -> None:
+    # This test should be tested without `PYTHONHASHSEED`. However, tox sets the
+    # environmental variable "PYTHONHASHSEED" by default.
+    os.unsetenv("PYTHONHASHSEED")
+
     # Multiprocessing supports three way to start a process.
     # We use `spawn` option to create a child process as a fresh python process.
     # For more detail, see https://github.com/optuna/optuna/pull/3187#issuecomment-997673037.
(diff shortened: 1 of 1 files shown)


not522:
Could you restore the original seed at the end of `test_reproducible_in_other_process`?

nzw0301:
@not522 Sure. I've updated the PR. 

Note that I replace `unsetenv` by following the recommendation of https://docs.python.org/3/library/os.html#os.unsetenv that reads

> Deletion of items in [os.environ](https://docs.python.org/3/library/os.html#os.environ) is automatically translated into a corresponding call to [unsetenv()](https://docs.python.org/3/library/os.html#os.unsetenv); however, calls to [unsetenv()](https://docs.python.org/3/library/os.html#os.unsetenv) don’t update [os.environ](https://docs.python.org/3/library/os.html#os.environ), so it is actually preferable to delete items of [os.environ](https://docs.python.org/3/library/os.html#os.environ).

not522 (approved):
Thank you! LGTM.

tests/samplers_tests/test_samplers.py
```diff
@@ -1035,3 +1042,6 @@ def test_reproducible_in_other_process(sampler_class_index: int) -> None:
     assert not (hash_dict[0] == hash_dict[1] == hash_dict[2])
     # But the sequences are expected to be the same.
     assert sequence_dict[0] == sequence_dict[1] == sequence_dict[2]
+
+    if hash_seed is not None:
+        os.environ["PYTHONHASHSEED"] = hash_seed
```
not522, line 1047: Could you move these lines before asserts? If the assertion fails, `PYTHONHASHSEED` does not back to the original.

```suggestion
    if hash_seed is not None:
        os.environ["PYTHONHASHSEED"] = hash_seed

    assert not (hash_dict[0] == hash_dict[1] == hash_dict[2])
    # But the sequences are expected to be the same.
    assert sequence_dict[0] == sequence_dict[1] == sequence_dict[2]
```

not522 (changes requested):

tests/samplers_tests/test_samplers.py
nzw0301: Thank you for your comment and suggestions!

not522: The comment (This test should be~) doesn't match the code. Sorry, it is my mistake.

nzw0301: Thanks. I might misunderstand your comment, but I've slightly updated the comments not to specify `tox` case. What do you think?

not522: LGTM!

not522 (approved):

toshihikoyanase:
I'm sorry for the delayed response. I did not investigate in detailed, but I'm a bit anxious about the failure of the test case might change `PYTHONHASHSEED` unexpectedly. Do you know if pytest has a mechanism like `try-except-finally` to ensure `PYTHONHASHSEED` can always be reverted?

nzw0301:
Possibly https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization can do that. I'll check it.

nzw0301:
When I ran the target test: `test_reproducible_in_other_process`, and the following test, `test_another`, in https://github.com/nzw0301/optuna/blob/demo-unset-python-seed/tests/samplers_tests/test_samplers.py#L1038-L1069, where I commented out the other lines for this local test, I confirmed the hashseed value was restored in `test_another` even though an error was raised in the target function as @toshihikoyanase concerned.

```bash
PYTHONHASHSEED=1 pytest tests/samplers_tests/test_samplers.py


------------------------------------------------------------------------------------ Captured stdout call -------------------------------------------------------------------------------------
seed: None
....
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=================================================================================== short test summary info ===================================================================================
FAILED tests/samplers_tests/test_samplers.py::test_reproducible_in_other_process[0] - ValueError
FAILED tests/samplers_tests/test_samplers.py::test_reproducible_in_other_process[1] - ValueError
FAILED tests/samplers_tests/test_samplers.py::test_another - AssertionError: assert '1' is None
```

As we can see in the last line that comes from `test_another`, the hash seed is restored. I also note that the seed value was correctly unset in `test_reproducible_in_other_process`, which was confirmed by  `seed: None` line in the output.

tests/samplers_tests/test_samplers.py
```diff
@@ -1016,8 +1018,29 @@ def objective(trial: Trial) -> float:
...
+    def restore_seed() -> None:
+        if hash_seed is not None:
+            os.environ["PYTHONHASHSEED"] = hash_seed
+
+    request.addfinalizer(restore_seed)
+
+
 @pytest.mark.parametrize("sampler_class_index", range(len(sampler_class_with_seed)))
-def test_reproducible_in_other_process(sampler_class_index: int) -> None:
+def test_reproducible_in_other_process(sampler_class_index: int, unset_seed_in_test: None) -> None:
```
toshihikoyanase, line 1039: I have two questions about type hint. `unset_seed_in_test` is not referred to in this test function, and I guess why `mypy` raise any error without any annotations. I know that it is a fixture and is called background.

And also, the type of `unset_seed_in_test` is set to `None`. Is it a return type of `unset_seed_in_test()`?

toshihikoyanase (commented):
Thank you for your investigation and update. The fixture for finalization sounds great to me. I have questions about type hints.

tests/samplers_tests/test_samplers.py
nzw0301: Yes, your understanding is correct.

toshihikoyanase: Thank you.

toshihikoyanase (approved):

toshihikoyanase:
@not522 I found that you self-requested a review. Please merge after your review.

not522 (approved):
LGTM

Issue #3982, opened 2022-09-09
unrobust optuna sampling behaviour


HarveyYan:
### Expected behavior

I have something like this in the code: `trial.suggest_categorical('somename', [False, True])`

and when it is running sometimes it gives an error like `ValueError: 'True' not in (False, True).`, but sometimes it samples just fine without this kind of error.

I looked into issue with pdb and at `optuna/samplers/nsgaii/_crossover.py(115)` you actually have a `child_params` dictionary which contains  sampled parameters and there is `{..., 'somename': "False"}`, although it would make much better sense to have that parameter sampled in `boolean` values.

Any ideas why that would be happening? Thank you in advance.

### Environment

- Optuna version: 3.0.1
- Python version: 3.8.13
- OS:Linux-4.15.0-1051-aws-x86_64-with-glibc2.17
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
File "/lib/python3.8/site-packages/optuna/study/study.py", line 419, in optimize
    _optimize(
  File "/lib/python3.8/site-packages/optuna/study/_optimize.py", line 66, in _optimize
    _optimize_sequential(
  File "/lib/python3.8/site-packages/optuna/study/_optimize.py", line 160, in _optimize_sequential
    frozen_trial = _run_trial(study, func, catch)
  File "/lib/python3.8/site-packages/optuna/study/_optimize.py", line 187, in _run_trial
    trial = study.ask()
  File "/lib/python3.8/site-packages/optuna/study/study.py", line 514, in ask
    trial = trial_module.Trial(self, trial_id)
  File "/lib/python3.8/site-packages/optuna/trial/_trial.py", line 54, in __init__
    self._init_relative_params()
  File "/lib/python3.8/site-packages/optuna/trial/_trial.py", line 63, in _init_relative_params
    self.relative_params = self.study.sampler.sample_relative(
  File "/lib/python3.8/site-packages/optuna/samplers/nsgaii/_sampler.py", line 211, in sample_relative
    child_params = perform_crossover(
  File "/lib/python3.8/site-packages/optuna/samplers/nsgaii/_crossover.py", line 115, in perform_crossover
    if _is_contained(child_params, search_space):
  File "/lib/python3.8/site-packages/optuna/samplers/nsgaii/_crossover.py", line 160, in _is_contained
    if not param_distribution._contains(param_distribution.to_internal_repr(param)):
[...]

PR #4035, opened 2022-09-30
Specify object type to numpy array init to avoid unintended str cast 

nzw0301:
## Motivation

Fix the bug reported in #3982, specifically, the code shared in https://github.com/optuna/optuna/issues/3982#issuecomment-1263163810.

## Description of the changes

As mentioned by https://github.com/optuna/optuna/issues/3982#issuecomment-1263194047, current code casts the categorical values to `str` even elements in `choice` are not str type when values contain a str element. Such a minimal case is as follows:

```python
import numpy as np

print(np.array(["hoge", True]))
# array(['hoge', 'True'], dtype='<U5')

print(np.array(["hoge", 1]))
# array(['hoge', '1'], dtype='<U21')
```

### DIFF
--- a/optuna/samplers/nsgaii/_crossover.py
+++ b/optuna/samplers/nsgaii/_crossover.py
@@ -41,7 +41,8 @@ def _try_crossover(
             [
                 [parent.params[p] for p in categorical_search_space]
                 for parent in [parents[0], parents[-1]]
-            ]
+            ],
+            dtype=object,
         )
 
         child_categorical_array = _inlined_categorical_uniform_crossover(


HideakiImamura:

not522 (commented):
Could you add a test for this bug?

HideakiImamura (approved):
I privately discussed with @not522 and concluded that we can merge this PR without tests, and we will add tests as a follow-up task.

LGTM!

not522 (approved):
I checked that the [reproducible code](https://github.com/optuna/optuna/issues/3982#issuecomment-1263163810) runs successfully. LGTM.

PR #491, opened 2019-08-21
Remove `InTrialStudy` class.

sile:
This PR contains the following changes:
- Remove `InTrialStudy` class (use `Study` instead of it)
- Add a lock to check nested invocations of `Study.optimize` method
  - An exception is raised if a nested invocation is detected

PR #4051, opened 2022-10-07
Allow users to call `study.optimize()` in multiple threads

### LINKED ISSUE #4049: Allow users to call `study.optimize()` in multiple threads.
### Motivation

For users who are familiar with `concurrent.futures`, I'd like to support the following usage that calls `study.optimize()` in multiple threads, although we can concurrently evaluate objective function in multiple threads via the `n_jobs` option of `study.optimize()`.

```python
import optuna
from concurrent.futures import ThreadPoolExecutor


def objective(trial):
    y = trial.suggest_float("y", -100, 100)
    return x**2 + y


if __name__ == "__main__":
    study = optuna.create_study()
    with ThreadPoolExecutor(10) as pool:
        for i in range(10):
            pool.submit(study.optimize, objective, n_trials=10)

    print(f"Number of trials: {len(study.get_trials())}")
    print(f"Best params: {study.best_params}")
```

Also I'd like to support the following usage of `dask.distributed` (refs https://github.com/optuna/optuna/pull/2023#issuecomment-1268792200).

```python
import optuna

from dask.distributed import Client, wait


def objective(trial):
    y = trial.suggest_float("y", -100, 100)
    return x**2 + y


if __name__ == "__main__":
    with Client("...") as client:
        print(f"Dask dashboard is available at {client.dashboard_link}")
        backend_storage = optuna.storages.InMemoryStorage()
        dask_storage = optuna.integration.DaskStorage(storage=backend_storage)
        study = optuna.create_study(direction="maximize", storage=dask_storage)

        futures = []
        for i in range(10):
            futures.append(client.submit(study.optimize, objective, n_trials=100, pure=False))
        wait(futures)

        print(f"Number of trials: {len(study.get_trials())}")
        print(f"Best params: {study.best_params}")
```

### Description

In the current implementation of Optuna, we can call `study.optimize()` in multiple processes like:

```python
import optuna


def objective(trial):
    y = trial.suggest_float("y", -100, 100)
    return x**2 + y


if __name__ == "__main__":
        "sqlite:///example-db.sqlite3",
        engine_kwargs={"connect_args": {"timeout": 15.0}}
    )
    with ProcessPoolExecutor() as pool:
        for i in range(10):
            pool.submit(study.optimize, objective, n_trials=10)

    print(len(study.get_trials()))
```

However, it will be actually broken if we replace `ProcessPoolExecutor` with `ThreadPoolExecutor` due to following checks.

https://github.com/optuna/optuna/blob/6fc5392c458101d5bfda12d3f97dd88e11d4fd5b/optuna/study/_optimize.py#L53-L54
https://github.com/optuna/optuna/blob/6fc5392c458101d5bfda12d3f97dd88e11d4fd5b/optuna/study/_optimize.py#L121

As discussed at https://github.com/optuna/optuna/pull/2023#issuecomment-1268792200, the above checks are added at #491. In my recognition, it is originally intended to prevent the illegal calls of `study.optimize()`.
For example, Optuna's Trial object holds the Study object like [this](https://github.com/optuna/optuna/blob/master/optuna/trial/_trial.py#L26-L48), so components such as Sampler, Pruner, and the objective function can access the study through trials. It allows illegal calls from within the objective function like:

```python
def objective(trial):
    trial.study.optimize(objective, n_trials=10)
```

At least, I don't think it was intended to make the above use cases impossible. The above use cases will be fixed by simply removing the above three lines, but it may not be that easy as some features such as `study.stop()` and progress_bar may break.

### Alternatives (optional)

_No response_

_No response_

c-bata:
## Motivation

Close #4049

## Description of the changes

Use a context variable instead of using `threading.Lock`.
https://docs.python.org/3/library/contextvars.html

In this PR, I declare a module-global variable `_in_optimize_loop` holding a `ContextVar(..., default=False)`, which is a thread-local object, and then calls `_optimize()` function inside a new context.

### DIFF
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -61,7 +61,6 @@ def _optimize(
 
     progress_bar = pbar_module._ProgressBar(show_progress_bar, n_trials, timeout)
 
-    _in_optimize_loop.set(True)
     study._stop_flag = False
 
     try:
@@ -137,6 +136,7 @@ def _optimize_sequential(
     time_start: Optional[datetime.datetime],
     progress_bar: Optional[pbar_module._ProgressBar],
 ) -> None:
+    _in_optimize_loop.set(True)
     if reseed_sampler_rng:
         study.sampler.reseed_rng()
 
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -2,6 +2,7 @@ from concurrent.futures import FIRST_COMPLETED
 from concurrent.futures import Future
 from concurrent.futures import ThreadPoolExecutor
 from concurrent.futures import wait
+from contextvars import ContextVar
 import datetime
 import gc
 import itertools
@@ -32,6 +33,7 @@ from optuna.trial import TrialState
 
 
 _logger = logging.get_logger(__name__)
+_in_optimize_loop = ContextVar("_in_optimize_loop", default=False)
 
 
 def _optimize(
@@ -50,7 +52,7 @@ def _optimize(
             "The catch argument is of type '{}' but must be a tuple.".format(type(catch).__name__)
         )
 
-    if not study._optimize_lock.acquire(False):
+    if _in_optimize_loop.get():
         raise RuntimeError("Nested invocation of `Study.optimize` method isn't allowed.")
 
     if show_progress_bar and n_trials is None and timeout is not None and n_jobs != 1:
@@ -59,6 +61,7 @@ def _optimize(
 
     progress_bar = pbar_module._ProgressBar(show_progress_bar, n_trials, timeout)
 
+    _in_optimize_loop.set(True)
     study._stop_flag = False
 
     try:
@@ -118,7 +121,7 @@ def _optimize(
                         )
                     )
     finally:
-        study._optimize_lock.release()
+        _in_optimize_loop.set(False)
         progress_bar.close()
 
 
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -52,9 +52,6 @@ def _optimize(
             "The catch argument is of type '{}' but must be a tuple.".format(type(catch).__name__)
         )
 
-    if _in_optimize_loop.get():
-        raise RuntimeError("Nested invocation of `Study.optimize` method isn't allowed.")
-
     if show_progress_bar and n_trials is None and timeout is not None and n_jobs != 1:
         warnings.warn("The timeout-based progress bar is not supported with n_jobs != 1.")
         show_progress_bar = False
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -399,10 +399,6 @@ class Study:
                 Currently, progress bar is experimental feature and disabled
                 when ``n_trials`` is :obj:`None`, ``timeout`` not is :obj:`None`, and
                 ``n_jobs`` :math:`\\ne 1`.
-
-        Raises:
-            RuntimeError:
-                If nested invocation of this method occurs.
         """
 
         ctx = copy_context()
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1,6 +1,6 @@
+from contextvars import copy_context
 import copy
 from numbers import Real
-import threading
 from typing import Any
 from typing import Callable
 from typing import cast
@@ -32,6 +32,7 @@ from optuna.distributions import BaseDistribution
 from optuna.storages._cached_storage import _CachedStorage
 from optuna.storages._heartbeat import is_heartbeat_enabled
 from optuna.study._multi_objective import _get_pareto_front_trials
+from optuna.study._optimize import _in_optimize_loop
 from optuna.study._optimize import _optimize
 from optuna.study._study_direction import StudyDirection
 from optuna.study._study_summary import StudySummary  # NOQA
@@ -82,19 +83,7 @@ class Study:
         self.sampler = sampler or samplers.TPESampler()
         self.pruner = pruner or pruners.MedianPruner()
 
-        self._optimize_lock = threading.Lock()
         self._stop_flag = False
-
-    def __getstate__(self) -> Dict[Any, Any]:
-
-        state = self.__dict__.copy()
-        del state["_optimize_lock"]
-        return state
-
-    def __setstate__(self, state: Dict[Any, Any]) -> None:
-
-        self.__dict__.update(state)
-        self._optimize_lock = threading.Lock()
 
     @property
     def best_params(self) -> Dict[str, Any]:
@@ -416,7 +405,9 @@ class Study:
                 If nested invocation of this method occurs.
         """
 
-        _optimize(
+        ctx = copy_context()
+        ctx.run(
+            _optimize,
             study=self,
             func=func,
             n_trials=n_trials,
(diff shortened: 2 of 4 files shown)


HideakiImamura:
@hvy @knshnb Could you review this PR?

c-bata:
@HideakiImamura Thank you for assign reviewers! Let me change an assignee to @sile since he is an author of #491 :bow: 

@sile Could you review this PR?

optuna/study/_optimize.py
```diff
@@ -134,6 +136,7 @@ def _optimize_sequential(
     time_start: Optional[datetime.datetime],
     progress_bar: Optional[pbar_module._ProgressBar],
 ) -> None:
+    _in_optimize_loop.set(True)
```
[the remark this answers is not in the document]
c-bata: Thank you for your review! I actually tried it once. However, `n_jobs` option was broken since `_in_optimize_loop` is a thread local object and `n_jobs` option spawns new threads.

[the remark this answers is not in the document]
sile: That makes sense. Thanks!

sile (approved):
LGTM 👍

hvy (commented):
Thanks for the quick and compact PR!
Looks good overall. I have some comments, questions and concerns though. 

The main points are in the `_optimize` function, excluding `_optimize_sequential`, since the latter is already designed with multi-threading in mind. As for the progress bar, it will break as you mention but I think it's also behaving as expected. (I  got deadlocks occasionally but those disappeared when upgrading to the latest version of tqdm).

My question is, if this should be documented behavior that users may depend on. I noticed you didn't update any documentation or add tests. I believe `n_jobs` will cover most of the cases.

My perhaps only concern is that there is now a single state across the thread. That is, if a single thread creates multiple studies/tries to do two optimizations at the same time, he/she will get an error. I don't think you can do the following after this PR.

```python
import optuna

study = optuna.create_study()

def func(trial):
    study_inner = optuna.create_study()
    study_inner.optimize(lambda t: t.suggest_float("x", 0, 1), n_trials=10)  # Works in master, raises nested invocations error after this PR
    return study_inner..best_trial.values

study.optimize(func, n_trials=10)
```

Do you think it's okay, or can we think of some solution, e.g. keeping a state per study somehow, either bound to the object or perhaps a thread local dict?

I also think removing this constraint altogether is another viable/simple option, documenting and leaving it up to the user to not nest invocations of `Study.optimize`. Cc. @sile

sile:
> My perhaps only concern is that there is now a single state across the thread. That is, if a single thread creates multiple studies/tries to do two optimizations at the same time, he/she will get an error.

@hvy Good catch! 
It sounds critical to me because there is a possibility that a user tries to optimize a third-party library that also uses Optuna internally (and the user would get an error when calling `Study.optimize` method).  So, we should allow multiple study instances in the same thread from calling `optimize` method I think.

sile:
> I also think removing this constraint altogether is another viable/simple option, documenting and leaving it up to the user to not nest invocations of `Study.optimize`.

I think this is a reasonable option.

c-bata:
Thank you for your review!

> My question is, if this should be documented behavior that users may depend on. I noticed you didn't update any documentation or add tests. I believe n_jobs will cover most of the cases.

Yes, we probably should add some test cases for following scenarios :bow:

1. Call `study.optimize()` in multiple threads
1. Check whether or not a nested `study.optimize()` call raises a `RuntimeError()`. (This scenario can be omitted if we just remove this constraint.)

> Do you think it's okay, or can we think of some solution, e.g. keeping a state per study somehow, either bound to the object or perhaps a thread local dict?
> I also think removing this constraint altogether is another viable/simple option, documenting and leaving it up to the user to not nest invocations of Study.optimize.

Good catch! I would prefer the solution that just removes the constraint and updates the documentation. "keeping a state per study" seems to be an another viable option. However, I'd say we should prioritize simplicity.

> Seems fine but wondering whether you first considered threading.local. I believe we only need a thread local boolean, but please correct me if I'm wrong.

Either one would be fine. I just chose `contextvars` since it is newly added to Python and is more sophisticated.
https://peps.python.org/pep-0567/#converting-code-that-uses-threading-local

setup.py
```diff
@@ -42,6 +42,7 @@ def get_install_requires() -> List[str]:
         "sqlalchemy>=1.3.0",
         "tqdm",
         "PyYAML",  # Only used in `optuna/cli.py`.
+        'contextvars;python_version<"3.7"',
```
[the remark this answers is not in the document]
c-bata: Either one would be fine. I just chose `contextvars` since it is newly added to Python and is more sophisticated.
https://peps.python.org/pep-0567/#converting-code-that-uses-threading-local

c-bata:
Apparently we reached consensus to remove the constraint, update the documentation, and add some test cases (please tell me if you have other concerns). I will update the PR later :pray:

hvy:
Sounds reasonable to me, thanks.

optuna/study/study.py
```diff
@@ -410,13 +399,11 @@ def objective(trial):
                 Currently, progress bar is experimental feature and disabled
                 when ``n_trials`` is :obj:`None`, ``timeout`` not is :obj:`None`, and
                 ``n_jobs`` :math:`\\ne 1`.
-
-        Raises:
-            RuntimeError:
-                If nested invocation of this method occurs.
```
c-bata, line 416: Screenshot 2022-10-11 17 36 23

I tried to update the docstring like above to let users know nested invocation of `study.optimize()` is prohibited. But on second thought, I'm not sure if it is worth noting in the documentation.

@hvy Do you have any ideas or can we keep the documentation as it is?

c-bata:
I removed the constraint and updated the test cases.

optuna/study/study.py
hvy: Good point. I actually think we can omit it from the documentation as well, as I can't see valid use cases for trying to do so, and the behavior after this change is rather expected. @sile might know better about why we should, or should not, document it, though.

optuna/study/study.py
```diff
@@ -760,8 +746,7 @@ def objective(trial):
 
         """
 
-        if self._optimize_lock.acquire(False):
-            self._optimize_lock.release()
+        if not _in_optimize_loop.get():
```
hvy, line 749: Do you think we need to keep this thread local context, again, that may still span multiple studies? 
(That is, you can still call `Study.stop` on an "inner study", even when it's not running `Study.optimize`, which may be non-intuitive).

What would happen by the way if we keep a local member variable bound to each study instead? Or, I think we actually can discuss this error too, whether we need to keep it. Do you see negative consequences in removing it?

setup.py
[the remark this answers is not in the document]
hvy: Definitely not a strong opinion but if the additional sophistication is not necessary, I'd prefer sticking to `threading.local` to keep dependencies simpler and possibly maintenance easier. The amount of code is basically the same I think. It might just be that I'm not very familiar with `contextvars`.

optuna/study/study.py
sile: > I actually think we can omit it from the documentation as well

Agreed😆

c-bata:
@hvy Thank you for your thorough review.
This PR seems to contain some controversial changes. To make it less controversial, we can start with https://github.com/optuna/optuna/pull/4068 that do not change any behavior, and then discuss others in following pull requests. Does it sound reasonable?

optuna/study/study.py
c-bata: Good point.
I think we should avoid removing it, at least in this PR, since it may be non-intuitive as you said. "keeping a local member variable bound to each study" sounds reasonable option for us. I've just opened #4068 that just replacing `threading.Lock` with `threading.local`, so that the above concern can be resolved.

hvy:
Sounds great, thanks for the swift PR. 🙇 
Do you want to me review the newly opened PR, or do you want others to have a look too? I'm open for either, so please let me know.

c-bata:
Thank you for the confirmation. Let me assign you and @sile as reviewers :pray:

Issue #4026, opened 2022-09-29
Improve tests for heartbeat


not522:
### Motivation

Current tests for heartbeat are inadequate. We need to add lacking patterns.

### Description

[test_heartbeat.py](https://github.com/optuna/optuna/blob/e8abeec7d58b64b207bd7606056311d20de0241f/tests/storages_tests/test_heartbeat.py) lacks the following patterns.

- Test that `fail_stale_trials` does not change the trial state if the trial is not stale.
- Test that `fail_stale_trials` does not change the trial state if `record_heartbeat` is never called.

cross32768:
Can I take this issue?

g-votte:
@cross32768
Please go ahead!

g-votte:
Closing as #4055 fixed this issue.

PR #4054, opened 2022-10-07
Improve test for heartbeat: Add test for the case that trial state should be kept running

cross32768:
## Motivation
Resolve Issue https://github.com/optuna/optuna/issues/4026

## Description of the changes
* Test that `fail_stale_trials` does not change the trial state if the trial is not stale.
* Test that `fail_stale_trials` does not change the trial state if `record_heartbeat` is never called.

### DIFF
--- a/tests/storages_tests/test_heartbeat.py
+++ b/tests/storages_tests/test_heartbeat.py
@@ -231,7 +231,6 @@ def test_fail_stale_trials(storage_mode: str, grace_period: Optional[int]) -> No
         check_keep_trial_state_in_running(study)
 
         storage.record_heartbeat(trial._trial_id)
-
         check_keep_trial_state_in_running(study)
 
         time.sleep(_grace_period + 1)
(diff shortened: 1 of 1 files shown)


cross32768:
Should I take some action to solve the fail of CircleCI Pipeline?

HideakiImamura:

@cross32768 Thanks for the PR! I have never encountered this problem. We cannot rerun the CI for this problem. Could you try to rebase on the master branch?

cross32768:
I Added a small commit to rerun CircleCI Pipeline, but it looks not working again...

cross32768:
I added some `time.sleep` for test (original code have some `time.sleep()`. I added one more `time.sleep()` sentence, thus it may double the running time of test code).
Could this cause CircleCI to dead?

HideakiImamura:
Oh... Thanks for the try.

I will investigate your change soon. Please wait a minuite.

HideakiImamura:
I checked your code, and I think this is reasonable and has nothing todo with the current error. It may be the problem of the branch based on this PR. Could you create another branch like `improve-heartbeat-test` and re-create the PR? Sorry for inconvenience.

PR #4055, opened 2022-10-07
Improve test for heartbeat: Add test for the case that trial state should be kept running

cross32768:
## Motivation
Resolve Issue https://github.com/optuna/optuna/issues/4026

## Description of the changes

* Test that `fail_stale_trials` does not change the trial state if the trial is not stale.
* Test that `fail_stale_trials` does not change the trial state if `record_heartbeat` is never called.

MEMO: This PR is re-create version of https://github.com/optuna/optuna/pull/4054 to tackle CircleCI issue.

### DIFF
--- a/tests/storages_tests/test_heartbeat.py
+++ b/tests/storages_tests/test_heartbeat.py
@@ -231,7 +231,6 @@ def test_fail_stale_trials(storage_mode: str, grace_period: Optional[int]) -> No
         check_keep_trial_state_in_running(study)
 
         storage.record_heartbeat(trial._trial_id)
-
         check_keep_trial_state_in_running(study)
 
         time.sleep(_grace_period + 1)
--- a/tests/storages_tests/test_heartbeat.py
+++ b/tests/storages_tests/test_heartbeat.py
@@ -205,6 +205,16 @@ def test_fail_stale_trials(storage_mode: str, grace_period: Optional[int]) -> No
         assert study.system_attrs["test"] == "A"
         assert trial.system_attrs["test"] == "B"
 
+    def check_change_trial_state_to_fail(study: "optuna.Study") -> None:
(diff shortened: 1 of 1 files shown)


cross32768:
Hmm...

It may be a timing issue, if there's anything I can do, let me know, I'll try.

HideakiImamura:
Thanks for the re-creation. 

I have an idea. Do you have an account of CircleCI? If not, could you try to create an account?

HideakiImamura:
Oh, it seems that the CircleCI jobs have started and completed. Could you add a dummy change and push it to re-run the CI (to remove the CircleCI Pipeline error)?

HideakiImamura:

tests/storages_tests/test_heartbeat.py
```diff
@@ -205,6 +205,16 @@ def failed_trial_callback(study: "optuna.Study", trial: FrozenTrial) -> None:
         assert study.system_attrs["test"] == "A"
         assert trial.system_attrs["test"] == "B"
 
+    def check_change_trial_state_to_fail(study: "optuna.Study") -> None:
+        assert study.trials[0].state is TrialState.RUNNING
+        optuna.storages.fail_stale_trials(study)
+        assert study.trials[0].state is TrialState.FAIL  # type: ignore [comparison-overlap]
```
HideakiImamura, line 211: Good catch 👍 

```suggestion
        assert study.trials[0].state is TrialState.FAIL  # type: ignore [comparison-overlap]
```

HideakiImamura (approved):
Thanks LGTM!

tests/storages_tests/test_heartbeat.py
HideakiImamura, line 211: Good catch 👍

toshihikoyanase (approved):

PR #4056, opened 2022-10-07
Make TPESampler reproducible

contramundum53:
## Motivation
It seems that currently `TPESampler` is not strictly reproducible because the results of `np.argsort` have randomness by default when there are multiple equal objective values. To make it reproducible, we need to specify `kind="stable"`. 

## Description of the changes
Specify `kind="stable"` to `np.argsort` in `split_observation_pairs`.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -702,7 +702,7 @@ def _split_observation_pairs(
     # 3. Feasible trials are sorted by loss_vals.
     if violations is not None:
         violation_1d = np.array(violations, dtype=float)
-        idx = violation_1d.argsort()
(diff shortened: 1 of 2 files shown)


c-bata:

contramundum53:
Note that there may be some performance change introduced by this PR.

HideakiImamura (commented):
Could you add a test to validate this change?

contramundum53:
@HideakiImamura I added a reproducibility test for constant objective.

HideakiImamura (approved):
Thanks for the update. LGTM.

HideakiImamura:
@not522 Please merge this PR after your approval.

not522 (approved):
LGTM

Issue #4058, opened 2022-10-08
plot_pareto_front crahses if a sampler is GridSampler


ilpoli:
### Expected behavior

plot_pareto_front should work with any sampler if there're multiple directions.

### Environment

- Optuna version: 3.0.2
- Python version: 3.7.14
- OS: Google Colab

### Error messages, stack traces, or logs

```shell
ValueError                                Traceback (most recent call last)
<ipython-input-67-dac422f460f1> in <module>
----> 9 fig = optuna.visualization.plot_pareto_front(study)
     10 fig.show()

3 frames
/usr/local/lib/python3.7/dist-packages/optuna/visualization/_pareto_front.py in plot_pareto_front(study, target_names, include_dominated_trials, axis_order, constraints_func, targets)
    129 
    130     info = _get_pareto_front_info(
--> 131         study, target_names, include_dominated_trials, axis_order, constraints_func, targets
    132     )
    133     return _get_pareto_front_plot(info)

/usr/local/lib/python3.7/dist-packages/optuna/visualization/_pareto_front.py in _get_pareto_front_info(study, target_names, include_dominated_trials, axis_order, constraints_func, targets)
    252         if include_dominated_trials:
    253             non_best_trials = _get_non_pareto_front_trials(
--> 254                 study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,)), best_trials
    255             )
    256         else:

/usr/local/lib/python3.7/dist-packages/optuna/visualization/_pareto_front.py in _get_non_pareto_front_trials(trials, pareto_trials)
    362     non_pareto_trials = []
    363     for trial in trials:
--> 364         if trial not in pareto_trials:
    365             non_pareto_trials.append(trial)
    366     return non_pareto_trials

/usr/local/lib/python3.7/dist-packages/optuna/trial/_frozen.py in __eq__(self, other)
    181         if not isinstance(other, FrozenTrial):
    182             return NotImplemented
--> 183         return other.__dict__ == self.__dict__
    184 
    185     def __lt__(self, other: Any) -> bool:
```

### Steps to reproduce

```python
def objective(trial):
    a = trial.suggest_int("a", 100, 200, 10)
    b = trial.suggest_float("b", 0.5, 1, step= 0.1)
    
    return a + b


search_space = {
    "a": np.linspace(100, 200, 10+1), 
    "b": np.linspace(0.5, 1.0, 5+1),
}
study = optuna.create_study(
    directions=["minimize", "minimize"], 
[...]

PR #4062, opened 2022-10-11
Convert `search_space` values of `GridSampler` explicitly

toshihikoyanase:
## Motivation

See #4058 .
In addition, I found that `study.optimize` raised errors when I use `RDBStorage` with search space defined with `numpy` functions such as `np.arange` and `np.linspace`.

```console
    def test_study_optimize_with_numpy_related_search_space() -> None:
    
            a = trial.suggest_float("a", 0, 100)
            b = trial.suggest_float("b", -0.1, 0.1)
            return a + b
    
        # Test that all combinations of the grid is sampled.
        search_space = {
            "a": np.linspace(0, 10, 11),
            "b": np.arange(-0.1, 0.1, 0.05),
        }
        with StorageSupplier("sqlite") as storage:
            study = optuna.create_study(  # type: ignore
                sampler=samplers.GridSampler(search_space),
                storage=storage,
            )
>           study.optimize(objective, n_trials=None)
...
self = <json.encoder.JSONEncoder object at 0x7f0cbdf50ca0>, o = array([ 0.,  1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.])

    def default(self, o):
        """Implement this method in a subclass such that it returns
        a serializable object for ``o``, or calls the base implementation
        (to raise a ``TypeError``).
    
        For example, to support arbitrary iterators, you could
        implement default like this::
    
            def default(self, o):
                try:
                    iterable = iter(o)
                except TypeError:
                    pass
                else:
                    return list(iterable)
                # Let the base class default method raise the TypeError
                return JSONEncoder.default(self, o)
    
        """
>       raise TypeError(f'Object of type {o.__class__.__name__} '
                        f'is not JSON serializable')
E       TypeError: Object of type ndarray is not JSON serializable

/usr/lib/python3.8/json/encoder.py:179: TypeError
```

This is due to `np.arange` and `np.linspace` return `np.ndarray`, which is not json-serializable.
Users should cast them to lists by themselves, but I guess `GridSampler` will be more user-friendly if we take care of this issue.

## Description of the changes

- Cast values of `search_space` mapping of `GridSampler.__init__` to `list`
- Add a test case

### DIFF
--- a/optuna/samplers/_grid.py
+++ b/optuna/samplers/_grid.py
@@ -109,7 +109,7 @@ class GridSampler(BaseSampler):
 
         self._search_space = collections.OrderedDict()
         for param_name, param_values in sorted(search_space.items()):
-            self._search_space[param_name] = param_values
+            self._search_space[param_name] = list(param_values)
 
         self._all_grids = list(itertools.product(*self._search_space.values()))
         self._param_names = sorted(search_space.keys())
--- a/tests/samplers_tests/test_grid.py
+++ b/tests/samplers_tests/test_grid.py
@@ -16,6 +16,7 @@ from optuna import samplers
 from optuna.samplers._grid import GridValueType
 from optuna.storages import RetryFailedTrialCallback
 from optuna.testing.objectives import pruned_objective
+from optuna.testing.storages import StorageSupplier
 from optuna.trial import Trial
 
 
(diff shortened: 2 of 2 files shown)


c-bata:
@keisuke-umezawa @contramundum53 Could you review this PR?

contramundum53:
How about defining a custom Json decoder class that can decode `numpy` array/scalars, rather than converting them to list/float everywhere?

toshihikoyanase:
> defining a custom Json encoder class that can encode numpy array/scalars

Thank you for your comment. I don't think such a encoder resolves the original issue https://github.com/optuna/optuna/issues/4058. We need to make trials comparable including the contents of `system_attrs`.

contramundum53:
Thanks for explanation!
In that case, I think it is too fragile to simply convert `ndarray`s to `list`s, since one `nan` value in the `ndarray` would make a false comparison results.
I think comparing trial ids would be more appropriate.

toshihikoyanase:
> In that case, I think it is too fragile to simply convert ndarrays to lists, since one nan value in the ndarray would make false comparison results.
> I think either manually comparing trials with is or comparing trial ids would be more appropriate.

@contramundum53 Thank you for your comment. That makes sense. We cannot solve all problems related to `ndarray`, `RDBStorage` and trial comparison in this PR.

How about focusing on the original issue in #4058 to accept the following code ?

```python
search_space = {
    "a": np.linspace(100, 200, 10+1), 
    "b": np.linspace(0.5, 1.0, 5+1),
}
```

I guess it is consistent with the float casting in `study.tell` that accepts `np.floating`.
https://github.com/optuna/optuna/blob/4bc3dfce1502a3bfc8c4f83b9c1a1ae8a16fb51b/optuna/study/_tell.py#L165

contramundum53:
@toshihikoyanase Let's make that a follow-up issue. LGTM!

contramundum53 (approved):

c-bata:
Let me unassign @keisuke-umezawa since he looks busy.

@cross32768 Could you review this PR?

cross32768:
OK! I'll try it.

cross32768:
I think this PR is good minimal change to resolve the original issue. LGTM!

cross32768 (approved):

c-bata:
Let me merge this PR since approved by @contramundum53 and @cross32768.

PR #4065, opened 2022-10-11
Change skip condition in `_get_observation_pairs` from multivariate to group

knshnb:


knshnb:
Let me close this issue because I decided to adopt https://github.com/optuna/optuna/pull/4079.

PR #4079, opened 2022-10-14
Use all trials in TPESampler even when multivariate=True

knshnb:
## Motivation

In the current TPE, the trials without parameters that we want to sample are skipped when `multivatiate=True` ([link](https://github.com/optuna/optuna/blob/8787df5c9e3d8eae0a29ffbed91e08237c767dbc/optuna/samplers/_tpe/sampler.py#L633-L637)). However, when `multivariate=False`, those trials are used effectively in the following two steps.
- Split the whole trials into `below` and `above`.
- Hypervolume calculation for multi-objective optimization.

This PR fixes the behavior of `multivariate=True` and makes it consistent with `multivariate=False`.

## Description of the changes

- [x] Use all trials when `multivariate=False`.
- [x] Remove the `multivariate` argument from `_get_observation_pairs`.

[Note] This PR slightly changes the current behavior of `TPESampler` in the following points.
- When `multivariate=True` and `group=True`, the trials that do not include suggesting parameters are also used. (When `multivariate=True` and `group=False`, this does not change the behavior because parameters in the intersection search space are included in all trials)
- ~Compare `n_startup_trials` with only the number of trials that include suggesting parameters.~


c-bata:

tests/samplers_tests/tpe_tests/test_sampler.py
```diff
@@ -764,103 +764,28 @@ def objective(trial: Trial) -> float:
...
 
-@pytest.mark.parametrize("direction", ["minimize", "maximize"])
-@pytest.mark.parametrize(
-    "constraints_enabled, constraints_func, expected_violations",
-    [
-        (False, None, None),
-        (True, lambda trial: [(-1, -1), (0, -1), (1, -1), (2, -1)][trial.number], [0, 0, 1, 2]),
-    ],
-)
-def test_get_observation_pairs_multi(
```
HideakiImamura, line 818: Can we leave this test case and just remove the `True` flag from the call of `get_observation_pairs`?

HideakiImamura (commented):
Thanks for the great simplification of `TPESampler`. I have a comment. PTAL.

tests/samplers_tests/tpe_tests/test_sampler.py
knshnb: Good catch! Restored that test: https://github.com/optuna/optuna/pull/4079/commits/b9a15da2b6a4a84fbd09ae6d097cdee0fc4767b7.

knshnb:
I resolved the conflict with https://github.com/optuna/optuna/pull/4073. PTAL 🙏

not522 (commented):
Could you update the document of `n_startup_trials`? It does not match the current behavior.

knshnb:
@not522 Thanks for the comments. As discussed privately, I decided to keep the original behavior of `n_startup_trials` and focus only on the change of `_get_observation_pairs` in this PR.

not522 (approved):
LGTM

PR #4090, opened 2022-10-21
Deprecate description-checked label

g-votte:
This PR deprecates `description-checked` label. Since we no longer use the label in the issue operation, let's remove it from `exempt-issue-labels` in `stale.yml`.


c-bata (approved):
LGTM. Let me merge this PR after passed CI since we Optuna core-dev member reached a consensus on this change.

Issue #3661, opened 2022-06-13
Add best score reporting into progress bar


Rexhaif:
### Motivation

It becomes quite inconvenient to scroll through hundreds of optuna run logs in jupyter notebook to find what is the current best score(s). At the same time, progress bar always stays on  top of the output cell, and it would be easy to track current best score via progress bar description line. Also, it is already implemented in LightGBM integration: https://github.com/optuna/optuna/blob/2d6c15ee68065132070862b66de14894f35294ad/optuna/integration/_lightgbm_tuner/optimize.py#L205

### Description

Add the current best score(s) reporting to the progress bar's description or postfix string, when the progress bar is used.

### Alternatives (optional)

1. Look at the run logs, provided by optuna.
2. Save optimization runs info into some persistent storage - SQLite, Redis, RDBMS and track best scores via optuna-dashboard
3. Track scores by using MLFlow/WandB integration

david1542:
Hi, I [created a PR](https://github.com/optuna/optuna/pull/4093) that tries to address this issue. It looks like this now:
<img width="1449" alt="image" src="https://user-images.githubusercontent.com/9879252/197252448-3adad1e0-1f92-4fd9-bb1c-27d25b4d7d2b.png">

At the moment, it is shown by default. Maybe we can add a new flag that controls this behaviour. Can someone from the maintainers take a look? 🙏

PR #4093, opened 2022-10-21
Add more stats (`best_trial`, `best_value`) to tqdm's progress bar

david1542:
## Motivation
This PR tries to address #3661. Namely, it adds `best_trial` and `best_value` (from the `study` object) to tqdm's progress bar, on each update.

## Description of the changes
Namely, it adds `best_trial` and `best_value` (from the `study` object) to tqdm's progress bar, on each update.

### DIFF
--- a/optuna/progress_bar.py
+++ b/optuna/progress_bar.py
@@ -89,14 +89,14 @@ class _ProgressBar:
             elapsed_seconds:
                 The time past since :func:`~optuna.study.Study.optimize` started.
             study:
-                The current study object
+                The current study object.
         """
 
-        if not study._is_multi_objective():
-            msg = f"Best trial: {study.best_trial.number}. Best value: {study.best_value}"
-            self._progress_bar.set_description(msg)
-
         if self._is_valid:
+            if not study._is_multi_objective():
+                msg = f"Best trial: {study.best_trial.number}. Best value: {study.best_value}"
+                self._progress_bar.set_description(msg)
+
             if self._n_trials is not None:
                 self._progress_bar.update(1)
                 if self._timeout is not None:
--- a/optuna/progress_bar.py
+++ b/optuna/progress_bar.py
@@ -75,15 +75,25 @@ class _ProgressBar:
         optuna_logging.disable_default_handler()
         optuna_logging._get_library_root_logger().addHandler(_tqdm_handler)
 
-    def update(self, elapsed_seconds: float) -> None:
+    def update(self, elapsed_seconds: float, best_trial: int, best_value: Optional[float]) -> None:
         """Update the progress bars if ``is_valid`` is :obj:`True`.
 
         Args:
             elapsed_seconds:
                 The time past since :func:`~optuna.study.Study.optimize` started.
+            best_trial:
+                The number of the current best trial.
+            best_value:
+                The score of the current best trial.
         """
 
         if self._is_valid:
+            if self._n_trials is None and self._timeout is None:
+                assert False
+
+            msg = f"Best trial: {best_trial}. Best value: {best_value}"
+            self._progress_bar.set_description(msg)
+
             if self._n_trials is not None:
                 self._progress_bar.update(1)
                 if self._timeout is not None:
@@ -99,9 +109,6 @@ class _ProgressBar:
 
                 self._progress_bar.update(time_diff)
                 self._last_elapsed_seconds = elapsed_seconds
-
-            else:
-                assert False
 
     def close(self) -> None:
         """Close progress bars."""
--- a/optuna/progress_bar.py
+++ b/optuna/progress_bar.py
@@ -8,6 +8,7 @@ from tqdm.auto import tqdm
 from optuna import logging as optuna_logging
 from optuna._experimental import experimental_func
 
+
 if TYPE_CHECKING:
     from optuna.study import Study
 
--- a/optuna/progress_bar.py
+++ b/optuna/progress_bar.py
@@ -1,12 +1,15 @@
 import logging
 from typing import Any
 from typing import Optional
+from typing import TYPE_CHECKING
 
 from tqdm.auto import tqdm
 
 from optuna import logging as optuna_logging
 from optuna._experimental import experimental_func
 
+if TYPE_CHECKING:
+    from optuna.study import Study
 
 _tqdm_handler: Optional["_TqdmLoggingHandler"] = None
 
@@ -37,7 +40,10 @@ class _ProgressBar:
     """
 
     def __init__(
-        self, is_valid: bool, n_trials: Optional[int] = None, timeout: Optional[float] = None
+        self,
+        is_valid: bool,
+        n_trials: Optional[int] = None,
+        timeout: Optional[float] = None,
     ) -> None:
 
         self._is_valid = is_valid and (n_trials or timeout) is not None
@@ -75,25 +81,21 @@ class _ProgressBar:
         optuna_logging.disable_default_handler()
         optuna_logging._get_library_root_logger().addHandler(_tqdm_handler)
 
-    def update(self, elapsed_seconds: float, best_trial: int, best_value: Optional[float]) -> None:
+    def update(self, elapsed_seconds: float, study: "Study") -> None:
         """Update the progress bars if ``is_valid`` is :obj:`True`.
 
         Args:
             elapsed_seconds:
                 The time past since :func:`~optuna.study.Study.optimize` started.
-            best_trial:
-                The number of the current best trial.
-            best_value:
-                The score of the current best trial.
+            study:
+                The current study object
         """
 
-        if self._is_valid:
-            if self._n_trials is None and self._timeout is None:
-                assert False
-
-            msg = f"Best trial: {best_trial}. Best value: {best_value}"
+        if not study._is_multi_objective():
+            msg = f"Best trial: {study.best_trial.number}. Best value: {study.best_value}"
             self._progress_bar.set_description(msg)
 
+        if self._is_valid:
             if self._n_trials is not None:
                 self._progress_bar.update(1)
                 if self._timeout is not None:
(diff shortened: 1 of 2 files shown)


nzw0301:
I will not be able to be a reviewer soon, but let me leave a comment. 

Could you fix the failed CIs? I think the root cause is that `best_value` and `best_trial` are not defined in multi-objective cases. So we might need to show a warning message or disable the introduced feature by this PR with multi-objectives. 

In addition, could you add a test when you implement a new feature by following [the contribution guide](https://github.com/optuna/optuna/blob/master/CONTRIBUTING.md)?

david1542:
Hi @nzw0301, thanks for the reply. What do you think about formatting a longer message for multi objective studies? The downside is that the message could potentially be very long. Maybe we can limit it.

nzw0301:
> What do you think about formatting a longer message for multi objective studies? The downside is that the message could potentially be very long. Maybe we can limit it.

I think your understanding is correct since we have multiple best values and trials. So we do not need to show message for multi-objective rather than limiting them because such truncated message sounds meaningless.

contramundum53:

toshihikoyanase:
> I think we do not need to show a message for multi-objectives rather than limiting them 

I agree with @nzw0301 . In multi-objective optimization, we can see hundreds or thousands of best trials, and it may be problem dependent to truncate informative trials.

As @nzw0301 will review this PR, let me change the assignment.

optuna/progress_bar.py
```diff
@@ -75,15 +75,25 @@ def _init_valid(self) -> None:
...
                 The time past since :func:`~optuna.study.Study.optimize` started.
+            best_trial:
+                The number of the current best trial.
+            best_value:
+                The score of the current best trial.
         """
 
         if self._is_valid:
+            if self._n_trials is None and self._timeout is None:
+                assert False
```
nzw0301, line 92: is this refactoring?

optuna/progress_bar.py
```diff
@@ -75,15 +75,25 @@ def _init_valid(self) -> None:
...
         optuna_logging._get_library_root_logger().addHandler(_tqdm_handler)
 
-    def update(self, elapsed_seconds: float) -> None:
+    def update(self, elapsed_seconds: float, best_trial: int, best_value: Optional[float]) -> None:
         """Update the progress bars if ``is_valid`` is :obj:`True`.
 
         Args:
             elapsed_seconds:
                 The time past since :func:`~optuna.study.Study.optimize` started.
+            best_trial:
```
nzw0301, line 84:
```suggestion
            best_trial_number:
```

optuna/progress_bar.py
```diff
@@ -75,15 +75,25 @@ def _init_valid(self) -> None:
...
+                The number of the current best trial.
+            best_value:
+                The score of the current best trial.
```
nzw0301, line 87: `value` would be more consist with the terminology in Optuna docs.
```suggestion
                The value of the current best trial.
```

optuna/progress_bar.py
```diff
@@ -75,15 +75,25 @@ def _init_valid(self) -> None:
         optuna_logging.disable_default_handler()
         optuna_logging._get_library_root_logger().addHandler(_tqdm_handler)
 
-    def update(self, elapsed_seconds: float) -> None:
+    def update(self, elapsed_seconds: float, best_trial: int, best_value: Optional[float]) -> None:
```
nzw0301, line 78: To avoid the error due to the multi-objective, how about taking `study`?
```suggestion
    def update(self, elapsed_seconds: float, study: Study) -> None:
```

optuna/study/_optimize.py
```diff
@@ -174,7 +174,11 @@ def _optimize_sequential(
                 callback(study, frozen_trial)
 
         if progress_bar is not None:
-            progress_bar.update((datetime.datetime.now() - time_start).total_seconds())
+            elapsed_seconds = (datetime.datetime.now() - time_start).total_seconds()
+            best_trial = study.best_trial.number
+            best_value = study.best_trial.value
+
+            progress_bar.update(elapsed_seconds, best_trial, best_value)
```
nzw0301, line 181:
```suggestion
            progress_bar.update(elapsed_seconds, study)
```

optuna/progress_bar.py
nzw0301: Note that we need to import `study` as follows to avoid circular import.

```python
if TYPE_CHECKING:
```

nzw0301 (changes requested):

optuna/progress_bar.py
david1542: Applied :)

optuna/progress_bar.py
david1542: Yeah. It's not necessary anymore, since I applied you suggestions and got rid of the custom description for now.

optuna/progress_bar.py
```diff
@@ -75,15 +75,25 @@ def _init_valid(self) -> None:
...
+            best_value:
+                The score of the current best trial.
         """
 
         if self._is_valid:
+            if self._n_trials is None and self._timeout is None:
+                assert False
+
+            msg = f"Best trial: {best_trial}. Best value: {best_value}"
+            self._progress_bar.set_description(msg)
```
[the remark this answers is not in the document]
david1542: Applied :)

optuna/study/_optimize.py
david1542: Done

david1542:
@nzw0301 Thanks for the review! I changed the code according to your suggestions. Can please you have a look? :)

optuna/progress_bar.py
```diff
@@ -75,14 +82,20 @@ def _init_valid(self) -> None:
...
 
-    def update(self, elapsed_seconds: float) -> None:
+    def update(self, elapsed_seconds: float, study: "Study") -> None:
         """Update the progress bars if ``is_valid`` is :obj:`True`.
 
         Args:
             elapsed_seconds:
                 The time past since :func:`~optuna.study.Study.optimize` started.
+            study:
+                The current study object
```
nzw0301, line 92: Let's follow Optuna coding guide.
```suggestion
                The current study object.
```

optuna/progress_bar.py
```diff
@@ -75,14 +82,20 @@ def _init_valid(self) -> None:
...
         """
 
+        if not study._is_multi_objective():
+            msg = f"Best trial: {study.best_trial.number}. Best value: {study.best_value}"
+            self._progress_bar.set_description(msg)
+
```
nzw0301, line 98: Could you move these lines to L100, specifically, just after `if self._is_valid:`? This will fix the failed CIs.

nzw0301 (commented):
Thanks! Could you fix a little bit more?

optuna/progress_bar.py
dudulasry: Fixed

optuna/progress_bar.py
dudulasry: Fixed

nzw0301: Thanks!

nzw0301:
I'll check the failed test case in a few hours.

contramundum53:
> When progress bar is shown based on `timeout` argument in `optimize`, optuna's code already uses `set_description_str` which is supposed to be a conflict with the changes introduced by this PR.
> 
> To address this part would be difficult for the first contributor. Thus I suppose the simplest solution is to support only when `n_trials` is not `None` for simplicity, a `timeout` part is a future task.
> 
> @contramundum53 what do you think?

@nzw0301 I couldn't find where `set_description_str` is used outside `progress_bar.py`. Could you tell me where a conflict would arise?

contramundum53:
@nzw0301 Thanks for your explanation.
I don't think it would be too difficult to remove the existing `set_description_str` usage.
https://github.com/optuna/optuna/pull/4093/files#diff-06c80a6776a808e2539b9cf426ccffa8e3f8dd11f2d51bb6bf2d61c781125b5aR67-R75
We could replace these lines with
```
            total = tqdm.format_interval(self._timeout)
            fmt = "{percentage:3.0f}%|{bar}| {elapsed}/" + total
            self._progress_bar = tqdm(total=self._timeout, bar_format=fmt)
```
How about this?

contramundum53:
@nzw0301 I'm sorry, the correct format seems to be 
```
fmt = "{desc}: {percentage:3.0f}%|{bar}| {elapsed}/" + total
```
according to the [documentation](https://tqdm.github.io/docs/tqdm/). Could you try that?

nzw0301:
@contramundum53  Thanks, but the progress bar stat wad not change. Did you change the other part too?

nzw0301:
@contramundum53 sorry, my local code was not the same as yours. I can see the output correctly now.

nzw0301:
@david1542 Could you replace https://github.com/optuna/optuna/blob/master/optuna/progress_bar.py#L60-L68 with 

```python
            total = tqdm.format_interval(self._timeout)
            fmt = "{desc} {percentage:3.0f}%|{bar}| {elapsed}/" + total
            self._progress_bar = tqdm(total=self._timeout, bar_format=fmt)
```

? 

Thanks @contramundum53 for your suggestions!

optuna/progress_bar.py
```diff
@@ -75,15 +82,21 @@ def _init_valid(self) -> None:
...
 
         Args:
             elapsed_seconds:
                 The time past since :func:`~optuna.study.Study.optimize` started.
+            study:
+                The current study object.
         """
 
         if self._is_valid:
+            if not study._is_multi_objective():
```
nzw0301, line 96: I realised that when a study has only non-complete trials such as a failed trial, the current code does not work as we expected. 
```suggestion
            if not study._is_multi_objective() and study.get_trials(states=[TrialState.COMPLETE]):
```

Note that we add the following import statement about L10.

```python
```

nzw0301:
I'll take over this PR.

PR #4097, opened 2022-10-26
Add a lock to `MLflowCallback`.

### LINKED ISSUE #4088: MLflowCallback error when using n_jobs > 1
### Expected behavior

be able to use more than one job and at the same time log it in mlflow

### Environment

- Optuna version: 3.0.3
- Python version: 3.9.7
- OS: Linux-4.14.35-1902.8.4.el7uek.x86_64-x86_64-with-glibc2.31
- MLFlow version:  1.29.0

```shell
Exception: Run with UUID 13a5f691281742d1a9d6b6f98c43203a is already active. To start a new run, first end the current run with mlflow.end_run(). To start a nested run, call start_run with nested=True
```

### Steps to reproduce

trying to use multiprocessing i get 

i know it is an mlflow issue, but there are some fixes to do when interacting with the mlflow api to support multiprocessing
so, some changes to MLflowCallback should be done

potential fixes
https://github.com/mlflow/mlflow/issues/3592
https://github.com/mlflow/mlflow/issues/1420

i tried even using:
`
    mlflow_kwargs={
        'nested': False
    }
`
but this raises duplication data and errors in mlflow

i also tried with
`
from joblib import parallel_backend
with parallel_backend('multiprocessing'):`
during study.optimize()

_No response_

gen740:
## Motivation
Fix #4088.

Optuna uses mlflow fluent API which is not threadsafe.
There is a way to use [Tracking API](https://mlflow.org/docs/latest/python_api/mlflow.client.html#module-mlflow.client) as mentioned in [this issue](https://github.com/mlflow/mlflow/issues/3592).
Instead of using Tracking API, I choose `threading.Lock()` to lock entire mlflow functions in order to avoid race-condition for simplicity.

## Description of the changes
Add a lock to `MLflowCallback` not to race-condition.

### DIFF
--- a/tests/integration_tests/test_mlflow.py
+++ b/tests/integration_tests/test_mlflow.py
@@ -279,6 +279,14 @@ def test_multiple_jobs(tmpdir: py.path.local, n_jobs: int) -> None:
     study = optuna.create_study(study_name=study_name)
     study.optimize(_objective_func, n_trials=n_trials, callbacks=[mlflc], n_jobs=n_jobs)
 
+    mlfl_client = MlflowClient(tracking_uri)
+    assert len(mlfl_client.list_experiments()) == 1
+
+    experiment = mlfl_client.list_experiments()[0]
+    runs = mlfl_client.list_run_infos(experiment.experiment_id)
+
+    assert len(runs) == n_trials
+
 
 def test_mlflow_callback_fails_when_nest_trials_is_false_and_active_run_exists(
     tmpdir: py.path.local,
--- a/tests/integration_tests/test_mlflow.py
+++ b/tests/integration_tests/test_mlflow.py
@@ -393,11 +393,12 @@ def test_log_mlflow_tags(tmpdir: py.path.local) -> None:
     assert all([tags[key] == str(value) for key, value in expected_tags.items()])
 
 
-def test_track_in_mlflow_decorator(tmpdir: py.path.local) -> None:
+@pytest.mark.parametrize("n_jobs", [1, 2, 4, 6])
+def test_track_in_mlflow_decorator(tmpdir: py.path.local, n_jobs: int) -> None:
 
     tracking_uri = f"file:{tmpdir}"
     study_name = "my_study"
-    n_trials = 3
+    n_trials = n_jobs * 2
 
     metric_name = "additional_metric"
     metric = 3.14
@@ -418,7 +419,7 @@ def test_track_in_mlflow_decorator(tmpdir: py.path.local) -> None:
     tracked_objective = mlflc.track_in_mlflow()(_objective_func)
 
     study = optuna.create_study(study_name=study_name)
-    study.optimize(tracked_objective, n_trials=n_trials, callbacks=[mlflc])
+    study.optimize(tracked_objective, n_trials=n_trials, callbacks=[mlflc], n_jobs=n_jobs)
 
     mlfl_client = MlflowClient(tracking_uri)
     experiments = mlfl_client.list_experiments()
--- a/tests/integration_tests/test_mlflow.py
+++ b/tests/integration_tests/test_mlflow.py
@@ -393,7 +393,7 @@ def test_log_mlflow_tags(tmpdir: py.path.local) -> None:
     assert all([tags[key] == str(value) for key, value in expected_tags.items()])
 
 
-@pytest.mark.parametrize("n_jobs", [1, 2, 4, 6])
+@pytest.mark.parametrize("n_jobs", [1, 2, 4])
 def test_track_in_mlflow_decorator(tmpdir: py.path.local, n_jobs: int) -> None:
 
     tracking_uri = f"file:{tmpdir}"
--- a/tests/integration_tests/test_mlflow.py
+++ b/tests/integration_tests/test_mlflow.py
@@ -268,6 +268,18 @@ def test_nest_trials(tmpdir: py.path.local) -> None:
     assert all(set(r.data.metrics.keys()) == {"value"} for r in child_runs)
 
 
+@pytest.mark.parametrize("n_jobs", [2, 4, 6])
+def test_multiple_jobs(tmpdir: py.path.local, n_jobs: int) -> None:
+    tracking_uri = f"file:{tmpdir}"
+    study_name = "my_study"
+    # The race-condition usually happens after first trial for each job
+    n_trials = n_jobs * 2
+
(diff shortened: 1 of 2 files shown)


toshihikoyanase:
[Question] Do we check the bugfix in unit tests? If we can write a simple test case, it would be helpful for the other developers.

gen740:
No.
I will do that.

tests/integration_tests/test_mlflow.py
```diff
@@ -268,6 +268,18 @@ def test_nest_trials(tmpdir: py.path.local) -> None:
...
+@pytest.mark.parametrize("n_jobs", [2, 4, 6])
+def test_multiple_jobs(tmpdir: py.path.local, n_jobs: int) -> None:
+    tracking_uri = f"file:{tmpdir}"
+    study_name = "my_study"
+    # The race-condition usually happens after first trial for each job
+    n_trials = n_jobs * 2
+
+    mlflc = MLflowCallback(tracking_uri=tracking_uri)
+    study = optuna.create_study(study_name=study_name)
+    study.optimize(_objective_func, n_trials=n_trials, callbacks=[mlflc], n_jobs=n_jobs)
```
toshihikoyanase, line 280: How about checking the number of runs?
This is because mlflow might omit records (runs) silently if they collide.

nzw0301 (commented):
Can we add a decorator's, `track_in_mlflow`, test because this PR also fixes the issue in the decorator in my understanding?

tests/integration_tests/test_mlflow.py
gen740: That's nice. Thank you!

david-thrower:
Hi,

I had 3 comments on this one. I don't claim to know everything, and please don't think I am.

One comment was a Feature Request:  I think it may eliminate a lot of confusion for many if the callback has a validation that detects and raises a warning if a user makes **this** mistake in setting up their run. So doing will cause a constellation of errors that appear to be an incidental race condition, when in reality was the result of the user's error (my) error ... and (user) will waste copious amounts of  time doing the typical race condition troubleshooting like adding locks, semaphores, and manually creating run ids from random numbers with a different seed for each worker  ... all which won't work.

```python3
import optuna
import mlflow

mlflc = MLflowCallback(
    metric_name="my metric score",
)

@mlflc.track_in_mlflow()
def objective(trial):
    with mlflow,start_run(): #   <----------------------------------  THIS MISTAKE
        mlflow.log_param("power", 2)
        mlflow.log_metric("base of metric", x - 2)

        return (x - 2) ** 2
# ...
```

The docs for MLflow show this as the preferred setup, and most posts on "how to fix MLflow race conditions will tell you to " tell you to do this to fix it (which pours gasoline on the fire in this case.).

Second comment: Documentation issue.

Third comment, have we considered the pros and cons of the file lock vs a semaphore. I am a little concerned that a lock may cause deadlocks, especially if a process fails mid-flight. I don't claim to be an expert at this area, but I have heard that semaphores can be a tad more robust against dead locks (at the expense of thread safety). Hopefully the lock it is at least being applied under a context handler.

HideakiImamura:
@nzw0301 Could you review this PR if you have time?

HideakiImamura (approved):
Looks very good.

HideakiImamura:
@david-thrower Thanks for the three very useful comments based on the user's point of view. As a reviewer, I am happy with the changes in the current PR and think your three suggestions are changes that should be made in another PR. (Keeping code changes small is important to speed up development and improve review quality.) How about elaborating on your three comments and creating a new issue? Since the three comments seem to be somewhat independent of each other, it would be helpful if you could create three issues for each of them, if possible.

david-thrower:
I fully agree, and by no means suggest holding the bugfix. Anything fixing this issue is great. I will look at this deeper and open another issue if needed. Sorry for my less than coordinated thinking.

nzw0301:
Unfortunately, I'm a little bit busy for a few weeks; let me unassign myself.

tests/integration_tests/test_mlflow.py
```diff
@@ -268,6 +268,26 @@ def test_nest_trials(tmpdir: py.path.local) -> None:
     assert all(set(r.data.metrics.keys()) == {"value"} for r in child_runs)
 
 
+@pytest.mark.parametrize("n_jobs", [2, 4, 6])
```
toshihikoyanase, line 271: Do we need all these values?
In my environment, we the current change increase the test time from 1.9 sec to 2.5 sec. If we remove `6` from `test_multiple_jobs` and `test_track_in_mlflow_decorator`, it the test time will be about 2.0 sec.

And I also confirmed that `n_jobs=2` and `n_jobs=4` failed with the current master.

```suggestion
@pytest.mark.parametrize("n_jobs", [2, 4])
```

tests/integration_tests/test_mlflow.py
```diff
@@ -373,11 +393,12 @@ def test_log_mlflow_tags(tmpdir: py.path.local) -> None:
     assert all([tags[key] == str(value) for key, value in expected_tags.items()])
 
 
-def test_track_in_mlflow_decorator(tmpdir: py.path.local) -> None:
+@pytest.mark.parametrize("n_jobs", [1, 2, 4, 6])
```
toshihikoyanase, line 396: ditto.
```suggestion
@pytest.mark.parametrize("n_jobs", [1, 2, 4])
```

toshihikoyanase (commented):
Sorry for the delayed response.
I confirmed the change resolved the problem in https://github.com/optuna/optuna/issues/4088#issuecomment-1287811133.
It looks good to me, but I have comments on tests.

toshihikoyanase:
I also confirmed that the decorator worked with this change.

```python
import optuna


mlflc = MLflowCallback(
    metric_name="my metric score",
)

@mlflc.track_in_mlflow()
def objective(trial):
    return (x - 2) ** 2

study = optuna.create_study()
study.optimize(objective, n_trials=10, callbacks=[mlflc], n_jobs=4)
```

tests/integration_tests/test_mlflow.py
gen740: I thought setting n_jobs to 6 is not such an expensive task. So, I added `6` just in case. To be honest, this test sometimes passes the test when n_jobs is 2 on my machine. But it would be enough to test only 2 and 4.

toshihikoyanase (approved):

Issue #4096, opened 2022-10-24
Support GPU in BoTorchSampler


knshnb:
### Motivation

Currently, BoTorchSampler only supports CPU computation, while the BoTorch backend supports cuda accelaration.

### Description

Add `device: optional[torch.device] = None` argument to `BoTorchSampler.__init__` that can be used in the following way.
```python
sampler = BoTorchSampler()  # CPU
sampler = BoTorchSampler(device=torch.device("cuda:0"))  # GPU
```

### Additional context (optional)

We don't require a test of the GPU case for now because it is difficult to run in the current CI.

PR #4101, opened 2022-10-27
Add device argument to BoTorchSampler

nzw0301:
## Motivation

Resolve https://github.com/optuna/optuna/issues/4096.

## Description of the changes

Add `device` argument to `BoTorchSampler` and send input data of BoTorch's candidate functions to the device.

Naive benchmark results on Google colab were 1 min (GPU) vs 2 min (CPU). The code I used as follows:

```python
import torch
import optuna
from optuna.integration import BoTorchSampler


def o(t):
    for i in range(30):
        t.suggest_int(f"x_{i}", -100, 100)
        t.suggest_categorical(f"y_{i}", list(range(-100, 100)))

    return t.suggest_float("z", 1e-4, 10000, log=True)

# GPU
s = optuna.create_study(sampler=BoTorchSampler(device=torch.device("cuda:0")))
s.optimize(o, n_trials=50) # about 1 min

# CPU
s = optuna.create_study(sampler=BoTorchSampler(device=torch.device("cpu")))
s.optimize(o, n_trials=50) # about 2 min
```

## [updated]

The colab notebook is available at https://colab.research.google.com/drive/1PMfHH2ZlZ_FL1aFez5V56XzUmAFM2eil?usp=sharing.

### DIFF
--- a/tests/integration_tests/test_botorch.py
+++ b/tests/integration_tests/test_botorch.py
@@ -410,3 +410,22 @@ def test_call_after_trial_of_independent_sampler() -> None:
     ) as mock_object:
         study.optimize(lambda _: 1.0, n_trials=1)
         assert mock_object.call_count == 1
+
+
+@pytest.mark.parametrize("device", [None, torch.device("cpu"), torch.device("cuda:0")])
+def test_device_argument(device: Optional[torch.device]) -> None:
+
+    sampler = BoTorchSampler(device=device)
+    if not torch.cuda.is_available() and sampler._device.type == "cuda":
+        pytest.skip(reason="GPU is unavailable.")
+
+    def objective(trial: Trial) -> float:
+        return trial.suggest_float("x", 0.0, 1.0)
+
+    def constraints_func(trial: FrozenTrial) -> Sequence[float]:
+        x0 = trial.params["x"]
+        return [x0 - 0.5]
+
+    sampler = BoTorchSampler(constraints_func=constraints_func, n_startup_trials=1)
+    study = optuna.create_study(sampler=sampler)
+    study.optimize(objective, n_trials=3)
--- a/optuna/integration/botorch.py
+++ b/optuna/integration/botorch.py
@@ -387,7 +387,8 @@ class BoTorchSampler(BaseSampler):
         seed:
             Seed for random number generator.
         device:
-            PyTorch's device instance to store input and output data of BoTorch.
+            A ``torch.device`` to store input and output data of BoTorch. Please set a CUDA device
+            if you fasten sampling.
     """
 
     def __init__(
@@ -408,7 +409,7 @@ class BoTorchSampler(BaseSampler):
         n_startup_trials: int = 10,
         independent_sampler: Optional[BaseSampler] = None,
         seed: Optional[int] = None,
-        device: "torch.device" = torch.device("cpu"),
+        device: Optional["torch.device"] = None,
     ):
         _imports.check()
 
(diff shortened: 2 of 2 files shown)


contramundum53:
LGTM! 

contramundum53 (approved):

knshnb:
Thanks for the PR!
Could you delete the [TODO comment](https://github.com/optuna/optuna/blob/6ad71fc4224e9652320195e37d22ebcd1f9dbf2f/optuna/integration/botorch.py#L331-L332)? Also, how about adding a test of the argument `device=torch.device("cpu")`.

tests/integration_tests/test_botorch.py
```diff
@@ -410,3 +410,22 @@ def test_call_after_trial_of_independent_sampler() -> None:
...
         study.optimize(lambda _: 1.0, n_trials=1)
         assert mock_object.call_count == 1
+
+
+@pytest.mark.parametrize("device", [None, torch.device("cpu"), torch.device("cuda:0")])
+def test_device_argument(device: Optional[torch.device]) -> None:
+
+    sampler = BoTorchSampler(device=device)
+    if not torch.cuda.is_available() and sampler._device.type == "cuda":
+        pytest.skip(reason="GPU is unavailable.")
```
nzw0301, line 420: The result when GPU is available looks fine too: https://colab.research.google.com/drive/1PMfHH2ZlZ_FL1aFez5V56XzUmAFM2eil#scrollTo=3_b1lgABqyyw&line=1&uniqifier=1

tests/integration_tests/test_botorch.py
```diff
@@ -410,3 +410,22 @@ def test_call_after_trial_of_independent_sampler() -> None:
...
+
+    def objective(trial: Trial) -> float:
+        return trial.suggest_float("x", 0.0, 1.0)
+
+    def constraints_func(trial: FrozenTrial) -> Sequence[float]:
+        x0 = trial.params["x"]
+        return [x0 - 0.5]
```
nzw0301, line 427: In my understanding, a constraint is also sent onto `device`, so I include this part in the test.

tests/integration_tests/test_botorch.py
```diff
@@ -410,3 +410,22 @@ def test_call_after_trial_of_independent_sampler() -> None:
     ) as mock_object:
         study.optimize(lambda _: 1.0, n_trials=1)
         assert mock_object.call_count == 1
+
+
+@pytest.mark.parametrize("device", [None, torch.device("cpu"), torch.device("cuda:0")])
```
HideakiImamura, line 415: The test of `device = torch.device("cuda:0")` seems to be skipped in CI now.
From https://github.com/optuna/optuna/actions/runs/3344000982/jobs/5537846173, 
```
tests/integration_tests/test_botorch.py .....................s
```

How about adding a TODO comment for future work to enable the GPU test on CI?

HideakiImamura (commented):
Thanks for the PR. I have two comments. Could you take a look?

tests/integration_tests/test_botorch.py
nzw0301: Yes, you are right. I created a skip case intendedly to support both GPU-available and GPU-unavailable environments because we cannot assume all developers have GPUs on their local machines to run the test locally. 
If we create such TODO, GitHub issue is supposed to be more appropriate than in this test function.

HideakiImamura (approved):
Sorry for the late. LGTM!

PR #3985, opened 2022-09-10
Use truncated norm cdf instead of norm cdf

nzw0301:
## Motivation

When we run the following optuna optimisation with dynamic search space, 

```python
import optuna

def objective1(trial):
    return trial.suggest_float("x", -1, 1)

def objective2(trial):
    return trial.suggest_float("x", -100, -90)

study = optuna.create_study(sampler=optuna.samplers.TPESampler(seed=2))
study.optimize(objective1, n_trials=10)
study.optimize(objective2, n_trials=10)
```

TPE sampler (specifically, Parzen Estimator) raises the `RuntimeWarning` as follows:

```
/Users/nzw/optuna/optuna/samplers/_tpe/parzen_estimator.py:193: RuntimeWarning: divide by zero encountered in divide
  coefficient = 1 / z / p_accept
/Users/nzw/optuna/optuna/samplers/_tpe/sampler.py:494: RuntimeWarning: invalid value encountered in subtract
```

One concern is to break the test `tests/samplers_tests/tpe_tests/test_sampler.py::test_sample_relative_handle_unsuccessful_states` that I'm not familiar with. So I really appreciate any feedback from others.

## Description of the changes

Replace `cdf` with truncated cdf.

PR #4105, opened 2022-10-31
Remove scipy dependencies from TPESampler

amylase:
## Motivation

Retry of #3560.

> SciPy requires relatively large storage and memory. It causes trouble in some constrained environments. E.g., This PR resolves https://github.com/optuna/optuna/issues/3035.

## Description of the changes

Now SciPy is only a test dependency. QMCSampler users may experience import error due to missing scipy after this change.

TPESampler time consumption is doubled. You can read the detailed results from the links below:
- this branch: https://github.com/amylase/optuna/actions/runs/3358589727
- master: https://github.com/amylase/optuna/actions/runs/3358590880

### DIFF
--- a/optuna/samplers/_tpe/_truncnorm.py
+++ b/optuna/samplers/_tpe/_truncnorm.py
@@ -1,0 +1,175 @@
+# This file contains the codes from SciPy project.
+#
+# Copyright (c) 2001-2002 Enthought, Inc. 2003-2022, SciPy Developers.
+# All rights reserved.
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+
+# 2. Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+
+# 3. Neither the name of the copyright holder nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import math
+import sys
+from typing import Callable
+
+import numpy as np
+
+
+_norm_pdf_C = math.sqrt(2 * math.pi)
+_norm_pdf_logC = math.log(_norm_pdf_C)
+
+
+def _log_sum(log_p: float, log_q: float) -> float:
+    if log_p > log_q:
+        log_p, log_q = log_q, log_p
+    return math.log1p(math.exp(log_p - log_q)) + log_q
+
+
+def _log_diff(log_p: float, log_q: float) -> float:
+    # returns log(q - p).
+    # assuming that log_q is always greater than log_q
+    return math.log1p(-math.exp(log_p - log_q)) + log_q
+
+
+def _ndtr(a: float) -> float:
+    x = a / 2**0.5
+    z = abs(x)
+
+    if z < 1 / 2**0.5:
+        y = 0.5 + 0.5 * math.erf(x)
+    else:
+        y = 0.5 * math.erfc(z)
+        if x > 0:
+            y = 1.0 - y
+
+    return y
+
+
+def _log_ndtr(a: float) -> float:
+    if a > 6:
+        return -_ndtr(-a)
+    if a > -20:
+        return math.log(_ndtr(a))
+
+    log_LHS = -0.5 * a**2 - math.log(-a) - 0.5 * math.log(2 * math.pi)
+    last_total = 0.0
+    right_hand_side = 1.0
+    numerator = 1.0
+    denom_factor = 1.0
+    denom_cons = 1 / a**2
+    sign = 1
+    i = 0
+
+    while abs(last_total - right_hand_side) > sys.float_info.epsilon:
+        i += 1
+        last_total = right_hand_side
+        sign = -sign
+        denom_factor *= denom_cons
+        numerator *= 2 * i - 1
+        right_hand_side += sign * numerator * denom_factor
+
+    return log_LHS + math.log(right_hand_side)
+
+
+def _norm_logpdf(x: float) -> float:
+    return -(x**2) / 2.0 - _norm_pdf_logC
+
+
+def _log_gauss_mass(a: float, b: float) -> float:
+    """Log of Gaussian probability mass within an interval"""
+
+    # Calculations in right tail are inaccurate, so we'll exploit the
+    # symmetry and work only in the left tail
+
+    def mass_case_left(a: float, b: float) -> float:
+        return _log_diff(_log_ndtr(a), _log_ndtr(b))
+
+    def mass_case_right(a: float, b: float) -> float:
+        return mass_case_left(-b, -a)
+
+    def mass_case_central(a: float, b: float) -> float:
+        # Previously, this was implemented as:
+        # left_mass = mass_case_left(a, 0)
+        # right_mass = mass_case_right(0, b)
+        # return _log_sum(left_mass, right_mass)
+        # Catastrophic cancellation occurs as np.exp(log_mass) approaches 1.
+        # Correct for this with an alternative formulation.
+        # We're not concerned with underflow here: if only one term
+        # underflows, it was insignificant; if both terms underflow,
+        # the result can't accurately be represented in logspace anyway
+        # because sc.log1p(x) ~ x for small x.
+        return math.log1p(-_ndtr(a) - _ndtr(-b))
+
+    if b <= 0:
+        return mass_case_left(a, b)
+    elif a > 0:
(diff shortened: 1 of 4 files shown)


c-bata:
@contramundum53 @not522 Could you review this PR?

optuna/samplers/_tpe/_truncnorm.py
```diff
@@ -0,0 +1,175 @@
...
+_norm_pdf_logC = math.log(_norm_pdf_C)
+
+
+def _log_sum(log_p: float, log_q: float) -> float:
+    if log_p > log_q:
+        log_p, log_q = log_q, log_p
+    return math.log1p(math.exp(log_p - log_q)) + log_q
+
+
+def _log_diff(log_p: float, log_q: float) -> float:
```
not522, line 51: I think it is more readable to return `log(p - q)`.

tests/samplers_tests/tpe_tests/test_truncnorm.py
```diff
@@ -0,0 +1,37 @@
...
+from scipy.stats import truncnorm as truncnorm_scipy
+
+import optuna.samplers._tpe._truncnorm as truncnorm_ours
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 8, 0), reason="SciPy 1.9.2 is not supported in Python 3.7"
+)
+@pytest.mark.parametrize(
+    "a,b", [(-np.inf, np.inf), (-10, +10), (-1e-3, +1e-3), (10, 100), (-100, -10)]
```
not522, line 14: Could you add smaller `a` and `b` values? For example, `(-1, 1)`.

not522 (commented):
Thank you for your PR! I added minor comments. I'll check more detail and the benchmark results.
Just to confirm, does this PR keep the behavior except for numerical error?

tests/samplers_tests/tpe_tests/test_truncnorm.py
amylase: added (-1, +1)

amylase:
Thank you for the review @not522! I have updated my code as per your comments. Please have a look.

> Just to confirm, does this PR keep the behavior except for numerical error?

Yes, my intention is to keep the TPESampler behavior unchanged.

amylase:
Speedtest benchmark results are as follows:

master
```
[  0.00%] · For Optuna commit a8799363 <master>:
[  0.00%] ·· Benchmarking virtualenv-py3.8-fakeredis
[ 50.00%] ··· Running (optimize.OptimizeSuite.time_optimize--).
[100.00%] ··· optimize.OptimizeSuite.time_optimize                                                                    ok
[100.00%] ··· ============================= ============
                storage, sampler, n_trials
              ----------------------------- ------------
                  inmemory, random, 1000      250±10ms
                 inmemory, random, 10000     5.00±0.01s
                   inmemory, tpe, 1000       3.17±0.2s
                  inmemory, cmaes, 1000       738±30ms
                   sqlite, random, 1000      42.2±0.2s
               cached_sqlite, random, 1000   42.7±0.3s
                   redis, random, 1000       6.74±0.05s
                cached_redis, random, 1000   6.70±0.02s
              ============================= ============
```

remove-scipy-dependency
```
[  0.00%] · For Optuna commit 84876b75 <remove-scipy-dependency>:
[  0.00%] ·· Benchmarking virtualenv-py3.8-fakeredis
[ 50.00%] ··· Running (optimize.OptimizeSuite.time_optimize--).
[100.00%] ··· optimize.OptimizeSuite.time_optimize                                                                    ok
[100.00%] ··· ============================= ============
                storage, sampler, n_trials
              ----------------------------- ------------
                  inmemory, random, 1000      234±3ms
                 inmemory, random, 10000     4.78±0.03s
                   inmemory, tpe, 1000       5.34±0.04s
                  inmemory, cmaes, 1000       691±20ms
                   sqlite, random, 1000      41.6±0.5s
               cached_sqlite, random, 1000   40.7±0.1s
                   redis, random, 1000       6.77±0.02s
                cached_redis, random, 1000   6.76±0.02s
              ============================= ============
```

contramundum53:
@not522 @amylase Should we also test the case where `trunc_below` and `trunc_above` is equal?

optuna/samplers/_tpe/_truncnorm.py
```diff
@@ -0,0 +1,175 @@
...
+# modification, are permitted provided that the following conditions
+# are met:
+
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+
+# 2. Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials provided
+#    with the distribution.
```
c-bata, line 16: Like the CuPy's documentation, could you add the license page in our documentation?
https://docs.cupy.dev/en/stable/license.html#scipy

amylase: added a new doc page.

amylase:
Updated the branch.
- Added a testcase suggested by @contramundum53. 
- Added a third-party license page
- Use scipy for better performance if available

not522:
It seems still slower than master even if `scipy` is installed.

amylase:
@not522 yeah. scipy implementation is actually slower than ours in my environment. So I decided to completely stop using scipy code in TPESampler.

amylase:
Updated the branch. The major updates are:

### Included the changes from #3985 (by @nzw0301 )
Since #3985 includes the dependency on SciPy, I cherry-picked these changes here. 
 
### Lazily import SciPy
In my previous change, SciPy was eagerly imported. I fixed it.

## Performance Notes
I used the following code and command to evaluate the performance impact of this PR.

#### mybench.py
```python
import optuna


    return x**2 + y**2


def main():
    study = optuna.create_study(sampler=TPESampler())


if __name__ == "__main__":
    main()
```

```shell
$ python -m timeit "import mybench; mybench.main()"
```

### Results

master: 4.04s
nzw0301:resolve-zero-divide-tpe (#3985): 9.18s
amylase:remove-scipy-dependency w/ scipy: 9.43s
amylase:remove-scipy-dependency w/o scipy: 91.9s

Diff between master and #3985 is from sp.stats.truncnorm.cdf which is 20x slower than sp.special.erf. I think this is inevitable.
```python
>>> timeit.timeit("sp.stats.truncnorm.cdf(0, a=-1, b=1)", setup="import scipy as sp; sp.stats.truncnorm.cdf(0, -1, 1)", number=10000)
0.9007200490004834
>>> timeit.timeit("sp.special.erf(0)", setup="import scipy as sp; sp.special.erf(0)", number=10000)
0.005753725999966264
```

#3985 change have TPESampler use our implementations of special functions much more times than it did. This is why the current code is **10x slower without SciPy**.

contramundum53:
I think we could call `sp.special.erf` three times to work around `truncnorm.cdf`.

amylase:
@contramundum53 the code actually does so after several nested function calls. the problem is that the number of function calls are huge (there were 141008396 `_ndtr` calls in benchmark code in [my previous comment](https://github.com/optuna/optuna/pull/4105#issuecomment-1319722308))

amylase:
Reduced redundant function calls. now the benchmark code takes *49.9sec*.

not522:
I tried to speed up `truncnorm`. In my environment, this branch is 3x faster than the PR. https://github.com/optuna/optuna/commit/327fe598b418d2ed31a3ebf32d73ea21b9c7e3c9

- Use `np.frompyfunc` instead of `np.vectorize`
- Use `np.ndarray` as arguments instead of  `float`
- Set `maxsize` of `lru_cache` is 1000 (the default is 128)

contramundum53:
How about implementing https://stackoverflow.com/questions/457408/is-there-an-easily-available-implementation-of-erf-for-python with numpy? Would that further increase speed?

amylase:
Added more optimizations.
* Included @not522's optimization
* Added our own `erf` implementation
  * Followed the @contramundum53's suggestion. The implementation they suggested was very fast but not accurate enough.
  * I decided to use the cite the implementation of FreeBSD's standard C library. It has very permissive license as stated in the top of `_erf.py`.

Now the benchmark code finishes in *15.6 sec*, while scipy version takes in *9.43 sec* as reported in my previous comment.

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -184,24 +169,34 @@ def log_pdf(self, samples_dict: Dict[str, np.ndarray]) -> np.ndarray:
...
+
                     upper_bound = np.minimum(samples + q / 2.0, high)
                     lower_bound = np.maximum(samples - q / 2.0, low)
-                    cdf = cdf_func(upper_bound[:, None], mus[None], sigmas[None]) - cdf_func(
-                        lower_bound[:, None], mus[None], sigmas[None]
-                    )
+                    cdf = cdf_func(
+                        upper_bound[:, None], mus[None], sigmas[None], low, high
+                    ) - cdf_func(lower_bound[:, None], mus[None], sigmas[None], low, high)
                     log_pdf = np.log(cdf + EPS) - np.log(p_accept + EPS)
```
not522, line 195: I noticed that it can be rewritten as follows. With this change, the speed is comparable to using SciPy in my environment.

```python
log_gauss_mass = truncnorm._log_gauss_mass((lower_bound[:, None] - mus) / sigmas, (upper_bound[:, None] - mus) / sigmas)
log_p_accept = truncnorm._log_gauss_mass((low - mus) / sigmas, (high - mus) / sigmas)
log_pdf = log_gauss_mass - log_p_accept
```

amylase: This is blasting fast! The speed was comparable to SciPy version in my environment too.

amylase:
Included @not522's optimization on `log_pdf` it is now comparable or slightly faster than scipy version.

not522:
Thank you for your long-term efforts. It's almost LGTM.

1. Please remove `_trunc_normal_cdf` and `truncnorm.cdf`. They are no longer needed.
2. `_bisect` has room for improvement, but I think it's OK to merge.

Please merge the master branch. I will check that this PR does not change the behavior.

docs/source/index.rst
```diff
@@ -97,6 +97,8 @@ License
 
 MIT License (see `LICENSE <https://github.com/optuna/optuna/blob/master/LICENSE>`__).
 
+Optuna uses the codes from SciPy project (see :doc:`Third-party License <license_thirdparty>`).
```
c-bata, line 100: nit:

```suggestion
Optuna uses the codes from SciPy and fdlibm projects (see :doc:`Third-party License <license_thirdparty>`).
```

optuna/samplers/_tpe/_truncnorm.py
```diff
@@ -0,0 +1,235 @@
...
+    return m
+
+
+def _ndtri_exp_single(y: float) -> float:
+    # TODO(amylase): Justify this constant
+    return _bisect(_log_ndtr_single, -100, +100, y)
+
+
+def _ndtri_exp(y: np.ndarray) -> np.ndarray:
+    return np.frompyfunc(_ndtri_exp_single, 1, 1)(y).astype(float)
```
contramundum53, line 170:
```suggestion
def _ndtri_exp(z0: np.ndarray):
    C = 2 * math.sqrt(math.pi * 2)
    large = z0 > math.log(0.5)
    y = np.sqrt(-z0)
    y[large] = np.sqrt(-np.log(1 - np.exp(z0[large])))
    x = np.full_like(y, -1)
    for i in range(4):
        w = np.sqrt(-log_ndtr(x))
        x += C * (w - y) * w * np.exp(0.5 * x ** 2 - w ** 2)
    x[large] = -x[large]
    return x
```
[Possible followup] An algorithm based on Newton's method is simpler and should be a *lot* faster.

not522 (approved):
LGTM!

contramundum53 (approved):
LGTM!
Thanks for the PR!

PR #4110, opened 2022-11-02
Enable mypy `--no-implicit-reexport` option.

### LINKED ISSUE #3579: Add mypy's `--strict` option
### Motivation

mypy's `--strict` option will make Optuna's code more robust. It checks such as unused `ignore`, returning `Any` but `float` is specified, etc.

### Description

`--strict` is a set of the following options. We can add them one by one.

- [x] `--warn-unused-configs`
- [ ] `--disallow-any-generics`
- ~[ ] `--disallow-subclassing-any`~
- [x] `--disallow-untyped-calls` #3704 
- [x] `--disallow-untyped-defs`
- [x] `--disallow-incomplete-defs`
- [x] `--check-untyped-defs`
- ~[ ] `--disallow-untyped-decorators`~
- [x] `--no-implicit-optional` #3599 
- [x] `--warn-redundant-casts` #3602 
- [x] `--warn-unused-ignores` #3627 
- [ ] `--warn-return-any`
- [x] `--no-implicit-reexport`
- [x] `--strict-equality` #3648 
- [x] `--strict-concatenate`

### Alternatives (optional)

_No response_

_No response_

gen740:
## Motivation

The motivation for this PR is the following four points.

- Enabling mypy `--diable-implicit-reexport` option. Partially resolve #3579.
- Remove some of `# NOQA` comment
- Good for another project which uses optuna.
- Give a nice user experience.

This change also allows any other project which uses optuna to use mypy with `--diable-implicit-reexport` option.

Enabling this option, mypy will not resolve the "implicit reexport" value. So if some project uses a package that doesn't have "explicit reexport" syntax (`__all__`). mypy will give a `not defined` error.

(Because chainer doesn't have `__all__` syntax,  I added `# type: ignore` type comment in this PR. `optuna/integration/chainer.py`, etc.)

Moreover, this change is not only good for non-mypy users but also `pylance` or `pyright` users.

For example, when you write the following code in VSCode and use pylance.

```python
import inspect
from optuna.integration import MLflowCallback

print(inspect.getmodule(MLflowCallback))
```

pylance would issue the following diagnosis. And completion doesn't work either.

> "MLflowCallback" is not exported from a module "optuna.integration" Import from "optuna.integration.mlflow" instead

This is because pyright does not resolve implicit reexported value.

And more, `__all__` syntax affects only `from .. import *` syntax. But in `optuna/integration/__init__.py` this `__all__` value has no effect.

Try the below code,

```python
from optuna.integration import *
MLflowCallback
```

and python interpreter emits `NameError: name 'MLflowCallback' is not defined` error.

## Description of the changes

- Add `__all__` syntax, to `__init__.py` which doesn't have it.
- Add `__all__` syntax, to `..._improts.py` which doesn't have it.
- Remove the `# NOQA` comment which is no longer needed.
- Add `type: ignore` comment, due to the used package having no `__all__` syntax.
- Change the import statement which accesses the value across another file.

I'm not confident about `optuna/study/_dataframe.py` change, please tell me if anyone has a better idea.

And even in this PR, `optuna.integration` is not allowed to import by using `import *` syntax.

```python
from optuna.integration import *
```

### DIFF
--- a/optuna/integration/_lightgbm_tuner/__init__.py
+++ b/optuna/integration/_lightgbm_tuner/__init__.py
@@ -2,13 +2,15 @@ from typing import Any
 
 from optuna.integration._lightgbm_tuner.optimize import _imports
 from optuna.integration._lightgbm_tuner.optimize import LightGBMTuner
-from optuna.integration._lightgbm_tuner.optimize import LightGBMTunerCV  # NOQA
+from optuna.integration._lightgbm_tuner.optimize import LightGBMTunerCV
 
 
 if _imports.is_successful():
-    from optuna.integration._lightgbm_tuner.sklearn import LGBMClassifier  # NOQA
-    from optuna.integration._lightgbm_tuner.sklearn import LGBMModel  # NOQA
-    from optuna.integration._lightgbm_tuner.sklearn import LGBMRegressor  # NOQA
+    from optuna.integration._lightgbm_tuner.sklearn import LGBMClassifier
+    from optuna.integration._lightgbm_tuner.sklearn import LGBMModel
+    from optuna.integration._lightgbm_tuner.sklearn import LGBMRegressor
+
+    __all__ = ["LightGBMTuner", "LightGBMTunerCV", "LGBMClassifier", "LGBMModel", "LGBMRegressor"]
 
 
 def train(*args: Any, **kwargs: Any) -> Any:
--- a/optuna/integration/_lightgbm_tuner/__init__.py
+++ b/optuna/integration/_lightgbm_tuner/__init__.py
@@ -10,7 +10,7 @@ if _imports.is_successful():
     from optuna.integration._lightgbm_tuner.sklearn import LGBMModel
     from optuna.integration._lightgbm_tuner.sklearn import LGBMRegressor
 
-    __all__ = ["LightGBMTuner", "LightGBMTunerCV", "LGBMClassifier", "LGBMModel", "LGBMRegressor"]
+__all__ = ["LightGBMTuner", "LightGBMTunerCV", "LGBMClassifier", "LGBMModel", "LGBMRegressor"]
 
 
 def train(*args: Any, **kwargs: Any) -> Any:
--- a/optuna/integration/lightgbm.py
+++ b/optuna/integration/lightgbm.py
@@ -19,8 +19,6 @@ if _imports.is_successful():
     from optuna.integration._lightgbm_tuner import LightGBMTuner
     from optuna.integration._lightgbm_tuner import LightGBMTunerCV
 
-__all__ = ["Dataset", "LightGBMTuner", "LightGBMTunerCV"]
-
     _names_from_tuners = ["train", "LGBMModel", "LGBMClassifier", "LGBMRegressor"]
 
     # API from lightgbm.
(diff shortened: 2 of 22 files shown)


optuna/integration/_lightgbm_tuner/__init__.py
```diff
@@ -2,13 +2,15 @@
...
 
 if _imports.is_successful():
-    from optuna.integration._lightgbm_tuner.sklearn import LGBMClassifier  # NOQA
-    from optuna.integration._lightgbm_tuner.sklearn import LGBMModel  # NOQA
-    from optuna.integration._lightgbm_tuner.sklearn import LGBMRegressor  # NOQA
+    from optuna.integration._lightgbm_tuner.sklearn import LGBMClassifier
+    from optuna.integration._lightgbm_tuner.sklearn import LGBMModel
+    from optuna.integration._lightgbm_tuner.sklearn import LGBMRegressor
+
+    __all__ = ["LightGBMTuner", "LightGBMTunerCV", "LGBMClassifier", "LGBMModel", "LGBMRegressor"]
```
not522, line 13: How about defining `__all__` at the top level?

```suggestion
__all__ = ["LightGBMTuner", "LightGBMTunerCV", "LGBMClassifier", "LGBMModel", "LGBMRegressor"]
```

optuna/integration/lightgbm.py
```diff
@@ -8,16 +8,18 @@
...
     # To pass tests/integration_tests/lightgbm_tuner_tests/test_optimize.py.
-    from lightgbm import Dataset  # NOQA
+    from lightgbm import Dataset
 
-    from optuna.integration._lightgbm_tuner import LightGBMTuner  # NOQA
-    from optuna.integration._lightgbm_tuner import LightGBMTunerCV  # NOQA
+    from optuna.integration._lightgbm_tuner import LightGBMTuner
+    from optuna.integration._lightgbm_tuner import LightGBMTunerCV
+
+    __all__ = ["Dataset", "LightGBMTuner", "LightGBMTunerCV"]
```
not522, line 22: ditto.

```suggestion
__all__ = ["Dataset", "LightGBMTuner", "LightGBMTunerCV"]
```

optuna/visualization/_plotly_imports.py
```diff
@@ -19,3 +19,5 @@
             "For further information, please refer to the installation guide of plotly. ",
             name="plotly",
         )
+    else:
+        __all__ = ["_imports", "plotly", "go", "Contour", "Scatter", "make_subplots"]
```
not522, line 23: ditto.

```suggestion

__all__ = ["_imports", "plotly", "go", "Contour", "Scatter", "make_subplots"]
```

optuna/visualization/matplotlib/_matplotlib_imports.py
```diff
@@ -24,3 +24,16 @@
...
+        "matplotlib",
+        "matplotlib_version",
+        "plt",
+        "Axes",
+        "LineCollection",
+        "PathCollection",
+        "Colormap",
+        "ContourSet",
+        "Rectangle",
+    ]
```
not522, line 39: ditto.

```suggestion
__all__ = [
    "_imports",
    "matplotlib",
    "matplotlib_version",
    "plt",
    "Axes",
    "LineCollection",
    "PathCollection",
    "Colormap",
    "ContourSet",
    "Rectangle",
]
```

not522 (commented):
Thank you for your great PR!

> I'm not confident about optuna/study/_dataframe.py change, please tell me if anyone has a better idea.

It is a bit weird, but type annotation in `study.py` needs it. I think the change is OK, and resolving this tricky import can be done in another PR.

optuna/integration/_lightgbm_tuner/__init__.py
not522: Maybe you missed this.

gen740: Oh, thanks!

not522 (approved):
Thank you. LGTM!

HideakiImamura (approved):
LGTM.

Issue #4084, opened 2022-10-18
RuntimeError: Cannot clone object OptunaSearchCV, as the constructor either does not set or modifies parameter param_distributions


RNarayan73:
### Expected behavior

Hello,
I recently upgraded from Optuna 2.10.1 to 3.0.3 and am using OptunaSearchCV to optimise a Scikit-Learn pipeline as follows.
![image](https://user-images.githubusercontent.com/82275999/196514857-a3612ef9-2544-4d8e-82c1-4bdb681f67db.png)

using an amended version of cross_validate(). However, I get the error above when it tries to clone the estimator.

Has anyone come across this and is there a resolution?

Thanks in advance,
Narayan

### Environment

- Optuna version: 3.0.3
- Python version: 3.8.13
- OS: Windows 11 Home
- Scikit-Learn: 1.1.2

### Error messages, stack traces, or logs

```shell
OptunaSearchCV is experimental (supported from v0.17.0). The interface can change in the future.
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Input In [77], in <cell line: 2>()
      1 start = time()
----> 2 sv_results_opt = cross_val_thresh(opt_pipe, X, y, cv_val[VAL], result_metrics, #scoring=score_metrics, 
      3                                   return_estimator=True, return_train_score=True, 
      4                                   thresh_split=SPLIT, 
      5                                  )
      6 end = time()

Input In [46], in cross_val_thresh(estimator, X, y, cv, result_metrics, return_estimator, return_train_score, thresh_split, *args, **kwargs)
     12 estimator_list = []
     14 for i, (train_idx, test_idx) in enumerate(cv.split(X, y)):
---> 16     est_i = clone(estimator)
     18     test_idx_test = test_idx[0::2] if thresh_split else test_idx
     19     #test_idx_thresh = test_idx[1::2]
     20     
     21     #X_train, X_test = X[train_idx, :], X[test_idx, :]    # more versatile than .loc, works with df and np

File C:\Anaconda3\envs\skl_py38\lib\site-packages\sklearn\base.py:96, in clone(estimator, safe)
     94     param2 = params_set[name]
     95     if param1 is not param2:
---> 96         raise RuntimeError(
     97             "Cannot clone object %s, as the constructor "
     98             "either does not set or modifies parameter %s" % (estimator, name)
     99         )
    100 return new_object

[...]

PR #4120, opened 2022-11-04
OptunaSearchCV rejects deprecated distributions in param_distributions

amylase:
## Motivation
OptunaSearchCV constructed with deprecated distributions in param_distributions throws an exception when it is cloned. [issue](https://github.com/optuna/optuna/issues/4084)

This change is to reject the deprecated distributions in OptunaSearchCV to prevent from this error.

## Description of the changes

The cause of the issue is that sklearn's `clone` method runs sanity check to ensure that the estimator's constructor argument is not edited. OptunaSearchCV's implementation [violates this sanity check](https://github.com/optuna/optuna/issues/4084#issuecomment-1283320431).

I compared several options that solve this issue
- Reject deprecated distributions (selected)
  - (+) Users will know how to fix their code because they will get better error messages.
  - (-) some of existing code will not work after this change. I guess it is okay because `OptunaSearchCV` is marked as experimental and deprecated distributions are marked as deprecated.
- Only edit values of dict object passed to `param_distributions` argument
  - (+) The reported issue will be fixed and the there is no backward incompatibility.
  - (-) it is weird if the estimator tweaks the parameter given to its constructor.

### DIFF
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -713,12 +713,16 @@ class OptunaSearchCV(BaseEstimator):
         if not isinstance(param_distributions, dict):
             raise TypeError("param_distributions must be a dictionary.")
 
-        # TODO(himkt): Remove this method with the deletion of deprecated distributions.
-        # https://github.com/optuna/optuna/issues/2941
-        param_distributions = {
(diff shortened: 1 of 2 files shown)


c-bata (approved):
LGTM 👍

not522 (approved):
LGTM

PR #4125, opened 2022-11-06
Add workflow to test Optuna with lower versions of constraints

himkt:
Takeover of https://github.com/optuna/optuna/pull/3974.

## Motivation

To guarantee that Optuna works with the oldest versions of dependencies,
this PR adds the workflow to run the test with them.


himkt:
For this PR, I'd like only to add a workflow.

However, ideally, I think it's better to avoid redundant controls for the lower constraints both in `setup.py` and `test-with-lower.yml`. If it's possible to construct the lower bounded dependencies from the requirements specifications, this redundancy could be skipped. Unfortunately, the dependencies are embedded in `setup.py`, which makes it little bit difficult to parse the dependencies (importing `get_install_requires` wouldn't work well due to the existence of the `setup` invocation).

https://github.com/optuna/optuna/blob/1429e0e9a2991401659d3cd8bb229d80e0666b94/.github/workflows/tests-with-lower.yml#L59

So, I'd like to propose for the migration to `pyproject.toml` for managing package metadata. If the dependencies are written in TOML, we can easily parse them and get lower bouned packages.

.github/workflows/tests-with-lower.yml
```diff
@@ -0,0 +1,84 @@
...
+          ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}
+
+    - name: Setup pip
+      run: |
+        python -m pip install --upgrade pip
+        pip install --progress-bar off -U setuptools
+
+    - name: Install dependencies with lower bound constraints
+      run: |
+        # Install dependencies whose have lower bound constraints.
```
HideakiImamura, line 58: A nit.
```suggestion
        # Install dependencies with lower bound constraints.
```

HideakiImamura (commented):
Basically, LGTM. I think we need add other lowest dependencies which are specified in `optional`. In addition, we need to check `botorch == 0.4.0`.

.github/workflows/tests-with-lower.yml
```diff
@@ -0,0 +1,85 @@
+name: Tests with the oldest dependencies
```
toshihikoyanase, line 1: The difference between this file and `tests.yaml` seems to be limited:

```diff
1c1
< name: Tests
---
> name: Tests with the oldest dependencies
51c51
<     - name: Install
---
>     - name: Setup pip
55a56,63
>     - name: Install dependencies with lower bound constraints
>       run: |
>         # Install dependencies with lower bound constraints.
>         pip install alembic==1.5.0 cmaes==0.8.2 packaging==20.0 sqlalchemy==1.3.0
>         pip install plotly==4.0.0 scikit-learn==0.24.2  # optional extras
> 
>     - name: Install
>       run: |
```

It may be costly to maintain both files. Can we merge these files?
I'm not sure, but we may be able to use `matrix` and `if` in `tests.yaml` as follows:


```diff
diff --git a/.github/workflows/tests.yml b/.github/workflows/tests.yml
index 00d346cd3..1b29b7d4d 100644
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,6 +19,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
+        lower-bound: [true, false]
 
     services:
       redis:
@@ -48,11 +49,20 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}
 
-    - name: Install
+    - name: Setup pip
       run: |
         python -m pip install --upgrade pip
         pip install --progress-bar off -U setuptools
 
+    - name: Install dependencies with lower bound constraints
+      run: |
+        # Install dependencies with lower bound constraints.
+        pip install alembic==1.5.0 cmaes==0.8.2 packaging==20.0 sqlalchemy==1.3.0
+        pip install plotly==4.0.0 scikit-learn==0.24.2  # optional extras
+      if: ${{ matrix.lower-version }}
+
+    - name: Install
+      run: |
         # Install minimal dependencies and confirm that `import optuna` is successful.
         pip install --progress-bar off .
         python -c 'import optuna'
```

HideakiImamura:
By the way, the CI fails due to the old `plotly` does not support `kaleido`. I think we need to update the lower version of `plotly` to 4.9 according to [here](https://github.com/plotly/Kaleido#use-kaleido-to-export-plotlypy-figures-as-static-images).

.github/workflows/tests-with-lower.yml
himkt: Thank you so much for the advice, it really makes sense. I updated the PR based on your suggestion.

himkt:
@toshihikoyanase @HideakiImamura 
Updating the lower bound for plotly and unifying workflows were done. PTAL again. 🙏

.github/workflows/tests.yml
```diff
@@ -48,11 +49,20 @@ jobs:
...
       run: |
         python -m pip install --upgrade pip
         pip install --progress-bar off -U setuptools
 
+    - name: Install dependencies with lower bound constraints
+      run: |
+        # Install dependencies with lower bound constraints.
+        pip install alembic==1.5.0 cmaes==0.8.2 packaging==20.0 sqlalchemy==1.3.0
+        pip install plotly==4.9.0 scikit-learn==0.24.2  # optional extras
+      if: ${{ matrix.use-lower-version }}
```
HideakiImamura, line 62: I'm not sure if I'm understanding this correctly, but even if a lower version of library A is installed first, if a newer version of A is internally required for the subsequent installation of another library B, won't the newer version of A be installed on its own?

HideakiImamura: This may be a limitation of `pip`. Perhaps something closer to what we want to do is to install a lower limit version after the normal installation?

himkt: I see, thank you for your insight. Actually it seemingly happened to upgrade `packaging` unexpectedly.
https://github.com/optuna/optuna/actions/runs/3404326039/jobs/5661473599#step:8:171

> Perhaps something closer to what we want to do is to install a lower limit version after the normal installation?

I see. There could be other approach that overwrites `setup.py` (ideally `pyproject.toml`) to pin the versions of packages whose are constrained using a lower-bound. Do you think it is enough to just downgrading after installation as you suggested?

himkt: 24109e1 tried to reinstall packages, it seemingly worked as expected.

.github/workflows/tests.yml
```diff
@@ -19,6 +19,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.7', '3.8', '3.9', '3.10']
+        use-lower-version: [true, false]
```
HideakiImamura, line 22: The CI job name will be `Tests (3.7, false)` something like that. It is difficult to see the actual job. How about using the string as follows?
```suggestion
        use-lower-version: ['', 'use-lower-version']
```

.github/workflows/tests.yml
```diff
@@ -48,11 +49,13 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}
```
HideakiImamura, line 50: Should we need to fix the cache key? Since the installed libraries are different `use-lower-version=true` and `use-lower-version= false`.

himkt: Thank you. BTW, do we need this `restore-keys` specification? In my understanding, `{{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}` only matches `{{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}-v1` in this workflow.
(This key representation also appears in other workflows, but, I think `env.cache-name` is different there)

When we update the `key` (e.g. `{{ runner.os }}-${{ matrix.python-version }}-${{ env.cache-name }}-${{ hashFiles('**/setup.py') }}-v2`, `restore-keys` will be used to fetch the v1 cache. But I think it's OK to expire the old caches when updating it. What do you think?

HideakiImamura: Thank you for your consideration. I thought about it some more myself and came to the conclusion that there is no need to modify this line. The reasons are as follows.

First, `restore-keys` is an optional key that is searched for when there is no cache matching `key`. Thus, the intent of my previous comment was that both `key` and `restore-keys` should be modified. However, in the current workflow, if the `third-party-package-installation-strategy` is `with-minimum-version`, the `Install dependencies with lower bound constraints` job after the `Install` job uninstalls and reinstalls `alembic` and others, so caching dependencies makes no sense.

I'm sorry for confusion, but could you undo the change for this?

HideakiImamura: [Note] We have noticed that we should change the cache key, since we should distinguish between `Install dependencies with minimum versions` step is used or not.

HideakiImamura (approved):
LGTM.

toshihikoyanase:
I'm sorry but I'm busy this week. So, let me remove the assignment.
@c-bata Could you assign a new reviewer, please?

c-bata:
Sure 👍  Let me assign @contramundum53 as a new reviewer.

contramundum53:
LGTM.

contramundum53 (approved):

contramundum53:
Nit: the name for "normal" could be something like "with latest version", which is more informative.

himkt:
```
ImportError: cannot import name 'CMAwM' from 'cmaes' (/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/site-packages/cmaes/__init__.py)
```

himkt:
Update the minimum version based on https://github.com/optuna/optuna/blob/c6d6f42e5785d26b6becf40a164a790d24659a69/setup.py#L30.

HideakiImamura:
Thanks for the update. Let me merge this PR since we have got 2 approvals.

Issue #3630, opened 2022-06-04
No input sanitization in `enqueue_trial`


xadrianzetx:
### Expected behavior

While working on #3629 I found that passing param value of different type than suggested in objective function to `enqueue_trial` causes undefined behavior, depending on type of value passed and expected distribution. So for example:

* Enqueuing `{"x": "foo"}` but using `FloatDistribution` or `IntDistribution` for `x` in objective function causes optimization to crash with cryptic tracebacks.
* Enqueuing `{"x": 1}` but using `FloatDistribution` for `x` results in silent cast to float.
* Enqueuing `{"x": 1.1}` but using `IntDistribution` for `x` results in warning message and cast to int.

I'd expect either one of following behaviors:
1. Optimization fails with nice message informing user that incorrect param values were enqueued (including expected and actual distributions). 
2. Optimization continues, but trial with incorrect params is marked as failed with warning message.

This might be a bit nit-picky, but `enqueue_trial` allows user to fix suggestions to just about anything, and in large enough codebase I'd say its fairly easy to mistakenly switch two differently typed parameters, in which case current tracebacks (if any) are a bit confusing.

### Environment

- Optuna version: 3.0.0b1.dev
- Python version: 3.8.10
- OS: Ubuntu 20.04.4 LTS
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
Traceback (most recent call last):
  File "uhoh.py", line 7, in <module>
    trial.suggest_float("x", 1, 10)  # uh-oh!
  File "/home/adrian/code/optuna/env/lib/python3.8/site-packages/optuna/trial/_trial.py", line 156, in suggest_float
    return self._suggest(name, distribution)
  File "/home/adrian/code/optuna/env/lib/python3.8/site-packages/optuna/trial/_trial.py", line 587, in _suggest
    if self._is_fixed_param(name, distribution):
  File "/home/adrian/code/optuna/env/lib/python3.8/site-packages/optuna/trial/_trial.py", line 616, in _is_fixed_param
    contained = distribution._contains(param_value_in_internal_repr)
  File "/home/adrian/code/optuna/env/lib/python3.8/site-packages/optuna/distributions.py", line 183, in _contains
    return self.low <= value <= self.high
TypeError: '<=' not supported between instances of 'float' and 'str'
```

### Steps to reproduce

```python
import optuna
study = optuna.create_study()
study.enqueue_trial({"x": "foo"})
# Some time later in objective function...
[...]

PR #4126, opened 2022-11-06
Validation enqueue trial

cross32768:
## Motivation
Resolve https://github.com/optuna/optuna/issues/3630

## Description of the changes
Add explicit cast to float in `to_internal_repr` method of FloatDistribution and IntDistribution, and add error message when this cast failed.

Note: some value which can be casted to float, but not so appropriate (i.e. `"2.2"`) can pass the current validation. I consider handling it next.

### DIFF
--- a/optuna/testing/distributions.py
+++ b/optuna/testing/distributions.py
@@ -18,4 +18,4 @@ class UnsupportedDistribution(BaseDistribution):
         return {}
 
     def to_internal_repr(self, param_value_in_external_repr: Any) -> float:
-        return param_value_in_external_repr
+        return float(param_value_in_external_repr)
--- a/optuna/testing/distributions.py
+++ b/optuna/testing/distributions.py
@@ -1,3 +1,4 @@
+from typing import Any
 from typing import Dict
 
 from optuna.distributions import BaseDistribution
@@ -15,3 +16,6 @@ class UnsupportedDistribution(BaseDistribution):
     def _asdict(self) -> Dict:
 
         return {}
+
+    def to_internal_repr(self, param_value_in_external_repr: Any) -> float:
+        return param_value_in_external_repr
--- a/tests/test_distributions.py
+++ b/tests/test_distributions.py
@@ -203,7 +203,7 @@ def test_check_distribution_compatibility() -> None:
 @pytest.mark.parametrize(
     "value", (0, 1, 4, 10, 11, 1.1, "1", "1.1", "-1.0", True, False, np.ones(1), np.array([1.1]))
 )
-def test_int_internal_representation(value: int) -> None:
+def test_int_internal_representation(value: Any) -> None:
     i = distributions.IntDistribution(low=1, high=10)
 
     if isinstance(value, int):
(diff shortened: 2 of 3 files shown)


cross32768:
An idea: when str is inputted, raise error for FloatDistribution and IntDistribution

HideakiImamura:
How about adding a test case to validate this change?

cross32768:
> An idea: when str is inputted, raise error for FloatDistribution and IntDistribution

After some discussion, I update code to allow all float-castable value

cross32768:
> How about adding a test case to validate this change?

Thank you. I'll try next!

cross32768:
Done: Add some test cases

optuna/testing/distributions.py
```diff
@@ -15,3 +16,6 @@ def _contains(self, param_value_in_internal_repr: float) -> bool:
     def _asdict(self) -> Dict:
 
         return {}
+
+    def to_internal_repr(self, param_value_in_external_repr: Any) -> float:
+        return param_value_in_external_repr
```
HideakiImamura, line 21: Doesn't `mypy` causes an error on this since it implicitly cast the `Any` value to `float`? It might be good to cast explicitly the value to `float`.

HideakiImamura (commented):
Thanks for the update! Almost LGTM! I have a naive comment. PTAL.

optuna/testing/distributions.py
cross32768: Thank you for your comment. I add cast to float in next commit.

tests/test_distributions.py
```diff
@@ -200,17 +200,45 @@ def test_check_distribution_compatibility() -> None:
     )
 
 
-@pytest.mark.parametrize("value", (0, 1, 4, 10, 11))
+@pytest.mark.parametrize(
+    "value", (0, 1, 4, 10, 11, 1.1, "1", "1.1", "-1.0", True, False, np.ones(1), np.array([1.1]))
+)
 def test_int_internal_representation(value: int) -> None:
```
knshnb, line 206:
```suggestion
def test_int_internal_representation(value: Any) -> None:
```

tests/test_distributions.py
```diff
@@ -200,17 +200,45 @@ def test_check_distribution_compatibility() -> None:
...
+    with pytest.raises(ValueError):
+        i.to_internal_repr(value)
 
 
-@pytest.mark.parametrize("value", (1.99, 2.0, 4.5, 7, 7.1))
+@pytest.mark.parametrize(
+    "value",
+    (1.99, 2.0, 4.5, 7, 7.1, 1, "1", "1.1", "-1.0", True, False, np.ones(1), np.array([1.1])),
+)
 def test_float_internal_representation(value: float) -> None:
```
knshnb, line 227:
```suggestion
def test_float_internal_representation(value: Any) -> None:
```

knshnb (commented):
Thanks for the PR. The code looks great! I left nit suggestions.

knshnb (approved):
LGTM!

HideakiImamura:
Hmm... The visual regression test fails. Could you try to resolve the error? (The CI queue is very crowded and something bad may be happening as a result.)

HideakiImamura:
Sorry for the confusion. The change of this PR does not affect the visual regression test. Let me merge this PR after passing the required CI.

HideakiImamura (approved):
LGTM.

PR #4127, opened 2022-11-06
Remove unused function: `find_any_distribution`

himkt:
This is the cleanup PR for `setup.py`

## Motivation
I think `find_any_distribution` is no more used.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,9 +2,7 @@ import os
 import sys
 from typing import Dict
 from typing import List
-from typing import Optional
 
-import pkg_resources
 from setuptools import find_packages
 from setuptools import setup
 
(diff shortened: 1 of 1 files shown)


toshihikoyanase (approved):
`find_any_distribution` was not used since #1474 (2 years ago).

Good catch! Thank you.

HideakiImamura (approved):
Nice catch! LGTM!

g-votte:
Thanks for PR and swift review!

Adding some historical backgrounds:
- `find_any_distribution` is introduced in https://github.com/optuna/optuna/pull/175 to detect the out-of-date `pfnopt` package. (`pfnopt` is the initial repo name before Optuna become public.)
- The detection logic is removed in https://github.com/optuna/optuna/pull/1474. Thus, `find_any_distribution` is unused now.

PR #4100, opened 2022-10-27
Migrate CLI from `cliff` to `argparse` 

cross32768:
## Motivation
To resolve problem that optuna CLI does not work for the latest importlib-metadata, an idea is to migrate CLI from `cliff` (`cliff` depends `stevedore`, which causes the above problem) to `argparse`.

## Description of the changes
Replace features depends on `cliff` by `argparse`.
~~This PR is currently WIP, but I create PR for discussion.~~

* Done
  * Rewrite all subcomands (like `optuna ask`, `optuna study optimize`, etc.) by `argparse`
  * Pass all tests in [tests/test_cli.py](https://github.com/optuna/optuna/blob/master/tests/test_cli.py)

* WIP
  * Some options implemented inside of cliff are dropped now (`--log-file`, `-h/--help`, and ` --debug`). I will implement these features next
  * `optuna help` returns help with different format. I will investigate the cause and implement format close to current style if I can

--- Original `optuna help` (depends on `cliff`) ---

```
usage: optuna [--version] [-v | -q] [--log-file LOG_FILE] [-h] [--debug]
              [--storage STORAGE]        

options:
  --version             show program's version number and exit
  -v, --verbose         Increase verbosity of output. Can be repeated.
  -q, --quiet           Suppress output except warnings and errors.
  --log-file LOG_FILE
                        Specify a file to log output. Disabled by default.
  -h, --help            Show help message and exit.
  --debug               Show tracebacks on errors.
  --storage STORAGE
                        DB URL. (e.g. sqlite:///example.db)

Commands:
  ask  Create a new trial and suggest parameters.
  best-trial  Show the best trial.
  best-trials  Show a list of trials located at the Pareto front.
  complete  print bash completion command (cliff)
  create-study  Create a new study.
  delete-study  Delete a specified study.
  help  print detailed help for another command (cliff)
  storage upgrade  Upgrade the schema of a storage.
  studies  Show a list of studies.
  study optimize  Start optimization of a study. Deprecated since version 2.0.0.
  study set-user-attr  Set a user attribute to a study.
  tell  Finish a trial, which was created by the ask command.
  trials  Show a list of trials.
```
--- Current `optuna help` (only depends on argparse) ---

```
usage: optuna [--version] {ask,best-trial,best-trials,create-study,delete-study,storage upgrade,studies,study optimize,study set-user-attr,tell,trials,help} ...

positional arguments:
[...]

PR #4129, opened 2022-11-06
Added `user_attrs` to print by optuna studies in `cli.py`

### LINKED ISSUE #3899: Print `study.user_attr` by `optuna studies` command
### Motivation

Using the cli to query studies for their user_attr would be quite useful.  There doesn't seem to be a way to do this, though there does seem to be a way to set a user_attr from the cli

### Description

optuna studies could have a longform option, with format json/yaml/table; or a new option, optuna study --study-name could be added.

### Alternatives (optional)

_No response_

_No response_

gonzaload:
## Motivation

Resolve #3899


toshihikoyanase (commented):
Thank you for your PR. Could you check the values of `user_attrs` in the unit tests, please?

I think we can check them in `test_studies_command` and `test_studies_command_flatten`.
https://github.com/optuna/optuna/blob/1ab1132ea45384a7b2dfbdadb877d8577c47899b/tests/test_cli.py#L273-L392

gonzaload:
It should now be ready

toshihikoyanase:
Thanks! I'll check it.

tests/test_cli.py
```diff
@@ -351,15 +351,17 @@ def test_studies_command_flatten(output_format: Optional[str]) -> None:
         if output_format is None or output_format == "table":
             expected_keys_1 = expected_keys_2 = [
                 "name",
+                "user_attrs",
```
toshihikoyanase, line 354: Thank you for fixing the test.

I think we can add assertions to check if the values of `user_attrs` are correct. For example, we keep `study_1` empty (no user_attrs) and add 
 a user attribute to `study_2`. Then, we check the values in a similar way to `studies[0]["n_trials"]` and `studies[0]["direction"]`.

What do you think?

optuna/cli.py
```diff
@@ -343,6 +343,7 @@ class _Studies(_BaseCommand):
 
     _study_list_header = [
         ("name", ""),
+        ("user_attrs", ""),
```
toshihikoyanase, line 346: According to https://github.com/optuna/optuna/issues/3899#issuecomment-1232431835, the purpose of this PR is to make `optuna studies` and `optuna trial` consistent in terms of `user_attrs`.
On the other hand, current implementation is inconsistent when the `user_attrs` are empty.

*optuna trial*
`user_attrs` column is hidden when no trials have `user_attrs`.

```python
import optuna

study = optuna.create_study(
    storage="sqlite:///4129.db",
    study_name="4129",
)
```

```console
% study.optimize(objective, n_trials=2)
+--------+-------+---------------------+---------------------+----------------+----------+
| number | value | datetime_start      | datetime_complete   | duration       | state    |
+--------+-------+---------------------+---------------------+----------------+----------+
|      0 |   1.0 | 2022-11-06 14:25:54 | 2022-11-06 14:25:54 | 0:00:00.003276 | COMPLETE |
+--------+-------+---------------------+---------------------+----------------+----------+
```

*optuna studies*
`user_attrs` is always shown.

```console
% optuna create-study --storage sqlite:///4129-empty.db --study-name empty 
%  optuna studies --storage sqlite:///4129-empty.db                                                                                                                   [~/pfn/code/optuna]
+-------+------------+---------------+----------+----------------+
| name  | user_attrs | direction     | n_trials | datetime_start |
+-------+------------+---------------+----------+----------------+
| empty | {}         | ('MINIMIZE',) |        0 |                |
+-------+------------+---------------+----------+----------------+
```

Do we align the behavior when the `user_attrs` values are empty?
@g-votte @nzw0301

g-votte: As @toshihikoyanase mentions, the goal is to make `optuna studies` and `optuna trials` consistent. It is the case even for the edge case where `user_attrs` is empty.

gonzaload: True I will fix this

toshihikoyanase:
@knshnb  This PR is also about CLI like #4100. Could you join the review, please?

gonzaload:
So I have manage to emulate the trials behaviour, but I would like to know the final scope of the PR.

So far I have done
- Modify `_Studies(_BaseCommand)` in `cli.py` to print `user_attrs` whenever they are present.
- Modify `test_studies_command` and `test_studies_command_flatten` to test new functionality

After looking at how the `optuna trials` actually work  I realised that the current solution is more of a workaround,
since the functionality to print `user_attrs` is actually implemented in `_convert_to_dict`.
Also I found that currently there is no test in `optuna trials` for checking the print out of `user_attrs` 

So now my question to @g-votte and @toshihikoyanase 
Should this PR also include:
- Delete the workaround and implement a print `user_attrs` functionality based on `_convert_to_dict`.
- Testing of `user_attrs` for `optuna_trials`.
or would the current work be enough and I should create a separate issue for such modifications?

toshihikoyanase:
@gonzaload Thank you for your update. The current workaround is simple, and it seems reasonable to me.
As you mentioned, the current code is a bit inconsistent with `optuna trials` command, but the tasks you suggested seems to be out-of-scope of this PR.

> Delete the workaround and implement a print `user_attrs` functionality based on `_convert_to_dict`: needs quite more time
> Testing of `user_attrs` for optuna_trials.

The first task is a refactoring task and it does not affect users. So, we may be work on it after merging this PR.
About the second task, we can just focus on `optuna studies` test to keep the PR simple.

By the way, I'm a bit busy this week, so let me remove the review assignment.
@c-bata Could you assign a new reviewer, please?

c-bata:
Sure 👍  Let me assign @contramundum53 as a second reviewer.

contramundum53 (approved):
LGTM!

knshnb (approved):
Sorry for the delayed response. Let's make the two tasks @toshihikoyanase mentioned follow-up work. LGTM.

knshnb:
@gonzaload Thank you for the contribution to Optuna! Let me merge this PR.

Issue #4087, opened 2022-10-19
Make storages pickalable


HideakiImamura:
### Motivation

Currently, some storages are pickalable, but others are not. Specifically, 
- `InMemoryStorage` and `RDBStorage` are pickalable.
- `RedisStorage` and `JournalStorage` are not pickalable.

As a result, the following code will not work in certain environments for `RedisStorage` and `JournalStorage`, but work for `RDBStorage`. This is a variant of the test of `test_multiprocess` in `tests/storage_tests/test_with_server.py`.

```python
import argparse
from concurrent.futures import ProcessPoolExecutor
from typing import Sequence, Tuple

import numpy as np
import redis

import optuna

SQLITE_URL = "sqlite:///test.db"
REDIS_URL = "redis://localhost:6379"
JOURNAL_FILE = "JOURNAL_FILE"


def f(x: float, y: float) -> float:
    return (x - 3) ** 2 + y


def objective(trial: optuna.Trial) -> float:
    x = trial.suggest_float("x", -10, 10)
    y = trial.suggest_float("y", -10, 10)
    trial.report(x, 0)
    trial.report(y, 1)
    trial.set_user_attr("x", x)
    trial.set_system_attr("y", y)
    return f(x, y)


def run_optimize(args: Tuple[str, str]) -> None:
    study_name = args[0]
    storage = args[1]
    # Create a study
    study = optuna.load_study(study_name=study_name, storage=storage)
    # Run optimization
    study.optimize(objective, n_trials=20)


def _check_trials(trials: Sequence[optuna.trial.FrozenTrial]) -> None:
    # Check trial states.
    assert all(trial.state == optuna.trial.TrialState.COMPLETE for trial in trials)

    # Check trial values and params.
    assert all("x" in trial.params for trial in trials)
    assert all("y" in trial.params for trial in trials)
    assert all(
        np.isclose(
            np.asarray([trial.value for trial in trials]),
            [f(trial.params["x"], trial.params["y"]) for trial in trials],
            atol=1e-4,
        ).tolist()
    )

    # Check intermediate values.
    assert all(len(trial.intermediate_values) == 2 for trial in trials)
    assert all(trial.params["x"] == trial.intermediate_values[0] for trial in trials)
    assert all(trial.params["y"] == trial.intermediate_values[1] for trial in trials)

    # Check attrs.
    assert all(
        np.isclose(
            [trial.user_attrs["x"] for trial in trials],
            [trial.params["x"] for trial in trials],
[...]

PR #4130, opened 2022-11-06
add getstate and setstate to journal storage

shu65:
## Motivation

This PR solves the journal storage part of this issue https://github.com/optuna/optuna/issues/4087

## Description of the changes

This PR adds `__setstate__` and `__getstate__` to `JournalStorage`

### DIFF
--- a/tests/storages_tests/test_journal.py
+++ b/tests/storages_tests/test_journal.py
@@ -1,7 +1,6 @@
 from concurrent.futures import as_completed
 from concurrent.futures import ProcessPoolExecutor
 from concurrent.futures import ThreadPoolExecutor
-import os
 import pickle
 import tempfile
 from types import TracebackType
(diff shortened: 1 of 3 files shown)


tests/storages_tests/test_journal.py
```diff
@@ -116,3 +118,21 @@ def test_pop_waiting_trial_multiprocess_safe() -> None:
                 if trial_id is not None:
                     trial_id_set.add(trial_id)
         assert len(trial_id_set) == num_enqueued
+
+
+def test_pickle_dump_and_load() -> None:
+    with tempfile.TemporaryDirectory() as workdir:
+        file_name = os.path.join(workdir, "journal.log")
+        file_storage = optuna.storages.JournalFileStorage(file_name)
```
HideakiImamura, line 126: How about parametrizing the log storage using `LOG_STORAGE`? Note that the current `JournalRedisStorage` is not pickalable, so we need to skip the test for it.

tests/storages_tests/test_journal.py
```diff
@@ -116,3 +118,21 @@ def test_pop_waiting_trial_multiprocess_safe() -> None:
...
+        num_enqueued = 10
+        for i in range(num_enqueued):
+            study.enqueue_trial({"i": i})
+        with open(os.path.join(workdir, "pickled.pkl"), "wb") as f:
+            pickle.dump(storage, f)
+        with open(os.path.join(workdir, "pickled.pkl"), "rb") as f:
+            loaded_storage = pickle.load(f)
+        study_id = loaded_storage.get_study_id_from_name(study.study_name)
+        assert len(loaded_storage.get_all_trials(study_id=study_id)) == num_enqueued
+
```
HideakiImamura: Could you apply a formatter?

tests/storages_tests/test_journal.py
```diff
@@ -116,3 +118,21 @@ def test_pop_waiting_trial_multiprocess_safe() -> None:
...
+        file_storage = optuna.storages.JournalFileStorage(file_name)
+        storage = optuna.storages.JournalStorage(file_storage)
+        study = optuna.create_study(storage=storage)
+        num_enqueued = 10
+        for i in range(num_enqueued):
+            study.enqueue_trial({"i": i})
+        with open(os.path.join(workdir, "pickled.pkl"), "wb") as f:
+            pickle.dump(storage, f)
+        with open(os.path.join(workdir, "pickled.pkl"), "rb") as f:
+            loaded_storage = pickle.load(f)
```
HideakiImamura, line 135: Do we need yo write down the pickeled object to file? It can be enough to pickle and unpickle the object.

tests/storages_tests/test_journal.py
```diff
@@ -116,3 +118,21 @@ def test_pop_waiting_trial_multiprocess_safe() -> None:
...
+        study_id = loaded_storage.get_study_id_from_name(study.study_name)
+        assert len(loaded_storage.get_all_trials(study_id=study_id)) == num_enqueued
```
HideakiImamura, line 137: In addition, it might be better to confirm that the enqueued parameter `i`s are correct.

HideakiImamura (commented):
Thanks for the PR! Basically, LGTM. I have some comments on the test. Please take a look.

tests/storages_tests/test_journal.py
```diff
@@ -116,3 +117,23 @@ def test_pop_waiting_trial_multiprocess_safe() -> None:
                 if trial_id is not None:
                     trial_id_set.add(trial_id)
         assert len(trial_id_set) == num_enqueued
+
+
+@pytest.mark.parametrize("log_storage_type", LOG_STORAGE)
+def test_pickle_dump_and_load(log_storage_type: str) -> None:
+    if log_storage_type.startswith("redis"):
+        pytest.skip("The `fakeredis` does not support multi process environments.")
```
HideakiImamura, line 125: I think the skip message should be changed.
```suggestion
        pytest.skip("The `JournalRedisStorage` is not pickalable.")
```

tests/storages_tests/test_journal.py
shu65: Thank you for the comment. I use `trial == loaded_trial` instead of confirming that the enqueued parameter is are correct. Because, we can confirm all properties of trial.

HideakiImamura (commented):
Thanks for the update. I have a minor comment on the skip message. PTAL.

HideakiImamura (approved):
LGTM!

HideakiImamura (approved):
Thanks for the update. LGTM with the local confirmation.

Issue #3083, opened 2021-11-05
Hyperband bracket assignment using hash in parallelization


donny8:
I think there could be a problem when using Hyperband pruner in parallelization.
In Hyperband, each trial is assigned to the bracket by computing hash("{}{}".format(study.studyname, trial.number).
However, the hash gives different value for the same input when executed on different terminal.
A trial n, for example, could be assigned to the bracket 1 at process 1 and the bracket 2 at process 2 , which ruins the benefits of using Hyperband pruner instead of the SuccesiveHalving pruner.

## Expected behavior

A trial needs to be assigned to the identical bracket among each process.
 

## Environment

- Optuna version: 2.10.0
- Python version: 3.6.9
- OS: Linux

## Error messages, stack traces, or logs

I attached an custom-made log that I'd added in the SuccessiveHalving code, which is a part of the Hyperband.
The trials 1, 2, 3, 4 were included in bracket 1 at both process 1 and 2
However, the trials 5, 7, 9 were included in the bracket only at process 2. 

**Trials for bracket 1 at process 1**
![image](https://user-images.githubusercontent.com/55653250/140302703-a7ea8e8a-0f9d-4be1-ae1c-246686535c17.png)

**Trials for the same bracket 1 at process 2**
![image](https://user-images.githubusercontent.com/55653250/140302720-94e346e1-0984-4278-a278-005534efe9cc.png)

## Steps to reproduce

1. Add the following codes at the beginning of the **def _get_competing_values( )** in the **_successive_halving.py**
```
for t in trials :
    if rung_key in t.system_attrs:
        compet = [t.system_attrs[rung_key]]
        print('   Prune SH Get_competing_values: '+ rung_key, ' trial ',t.number,'competing ',compet)
```
2. Check if the trials consisting of a bracket differs among each process

## Additional context (optional)

As you can see [here](https://docs.python.org/3.3/using/cmdline.html), it was due to the update that hash randomization is turned on after python version 3.3.
Of course, the randomization could be turned off by setting the PYTHONHASHSEED=0 when executing the python script.
Since I did not see any warning or suggestion about this problem, I would like to ask if my assertion was legit.

hvy:
I think you're right. It's clearly manifested when creating the bracket studies (a study which contains a subset of the trials) in https://github.com/optuna/optuna/blob/v2.10.0/optuna/pruners/_hyperband.py.

Let me share a somewhat more limited but self contained example that demonstrates the issue that you're describing.

[...]

PR #4131, opened 2022-11-06
Stop using hash function in _get_bracket_id in Hyperband

zaburo-ch:
## Motivation

Fix the bug reported in https://github.com/optuna/optuna/issues/3083  that _get_bracket_id of Hyperband returns different results depending on the process.

## Description of the changes

Simply stopped using `hash`

### DIFF
--- a/optuna/pruners/_hyperband.py
+++ b/optuna/pruners/_hyperband.py
@@ -250,10 +250,7 @@ class HyperbandPruner(BasePruner):
             return 0
 
         assert self._n_brackets is not None
-        n = (
-            hash("{}_{}".format(study.study_name, trial.number))
-            % self._total_trial_allocation_budget
-        )
+        n = trial.number % self._total_trial_allocation_budget
         for bracket_id in range(self._n_brackets):
             n -= self._trial_allocation_budgets[bracket_id]
             if n < 0:


zaburo-ch:
I am concerned that if n_trials is not sufficiently larger than total_trial_allocation_budget, the bracket_id will be biased towards smaller ones by this change. This happens when max_resource is large and n_trials is small, so I wonder if it would make a difference in results in NN hyperparameter optimization.

```

import optuna as ot


def objective(t):
    x = t.suggest_float("x", 0, 1)

    for step in range(100):
        t.report(step=step, value=x * step)
        if t.should_prune():
            raise ot.TrialPruned

    return x


if __name__ == "__main__":
    p = ot.pruners.HyperbandPruner()
    study = ot.create_study(
        study_name="test_hyperband_3083",
        pruner=p,  # You actually don't need to prune, for this demonstration.
        load_if_exists=True,
    )

    # Check bracket splits. Using private APIs, only use for testing.
    brackets_to_trial_numbers = defaultdict(list)
    for t in study.trials:
        brackets_to_trial_numbers[study.pruner._get_bracket_id(study, t)].append(t.number)
    print({k: len(v) for k, v in brackets_to_trial_numbers.items()})
    # {0: 81, 1: 19}
```

c-bata (approved):
~~LGTM~~

Sorry. On second thought, as you said, the bracket_id will be biased due to this change. Let me review again :bow:

zaburo-ch:
To avoid the bias discussed above, I modified the implementation to use crc32, which was mentioned in https://github.com/optuna/optuna/issues/3083.

The result of the above code after modification is `{3: 9, 1: 21, 0: 58, 4: 2, 2: 10}`

c-bata (approved):
Thanks for the update! LGTM.

not522 (commented):
Could you remove this note? https://github.com/optuna/optuna/blob/f60976d6a7fc78db883de1f399be81d80fbb4fde/optuna/pruners/_hyperband.py#L63-L68

zaburo-ch:
Thanks for the review. I simply deleted the note, is this ok?

not522 (approved):
Thank you! LGTM!

Issue #4070, opened 2022-10-12
ValueError: math domain error


jmrichardson:
### Expected behavior

Getting a math domain error on what appears to be valid parameter suggestions (ie no 0 or negative log)

### Environment

- Python version:3.8.13
- Optuna version:3.0.2
- OS:Linux-3.10.0-1127.el7.x86_64-x86_64-with-glibc2.17

### Error messages, stack traces, or logs

```shell
Error Message:

Traceback (most recent call last):
  File "eras.py", line 47, in <module>
    pipeline = pipeline.fit(X_train, y_train, x_test=X_test, y_test=y_test)
  File "/home/jmrichardson/fml/fml/pipelines/standard.py", line 236, in fit
    model.optimize(features, test=test_features)
  File "/home/jmrichardson/fml/fml/models/lgbm.py", line 92, in optimize
    study.optimize(partial(objective, train, test), n_trials=self.n_trials, timeout=self.timeout, show_progress_bar=True)
  File "/home/jmrichardson/anaconda3/envs/fml/lib/python3.8/site-packages/optuna/study/study.py", line 419, in optimize
    _optimize(
  File "/home/jmrichardson/anaconda3/envs/fml/lib/python3.8/site-packages/optuna/study/_optimize.py", line 66, in _optimize
    _optimize_sequential(
  File "/home/jmrichardson/anaconda3/envs/fml/lib/python3.8/site-packages/optuna/study/_optimize.py", line 160, in _optimize_sequential
    frozen_trial = _run_trial(study, func, catch)
  File "/home/jmrichardson/anaconda3/envs/fml/lib/python3.8/site-packages/optuna/study/_optimize.py", line 187, in _run_trial
    trial = study.ask()
  File "/home/jmrichardson/anaconda3/envs/fml/lib/python3.8/site-packages/optuna/study/study.py", line 514, in ask
    trial = trial_module.Trial(self, trial_id)
  File "/home/jmrichardson/anaconda3/envs/fml/lib/python3.8/site-packages/optuna/trial/_trial.py", line 54, in __init__
    self._init_relative_params()
  File "/home/jmrichardson/anaconda3/envs/fml/lib/python3.8/site-packages/optuna/trial/_trial.py", line 63, in _init_relative_params
    self.relative_params = self.study.sampler.sample_relative(
  File "/home/jmrichardson/anaconda3/envs/fml/lib/python3.8/site-packages/optuna/samplers/nsgaii/_sampler.py", line 211, in sample_relative
[...]

PR #4137, opened 2022-11-06
Validation for the parameter enqueued in to_internal_repr of FloatDistribution and IntDistribution

cross32768:
Note: this PR depends on PR https://github.com/optuna/optuna/pull/4126. Please review and merge it first.

## Motivation

Resolve https://github.com/optuna/optuna/issues/4070

## Description of the changes

Add validation to raise error when value <= 0.0 is enqueued with log=True.
In addition, I add to error handling when not a number is enqueued.
I also added test cases to validate this change.

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -523,7 +523,7 @@ class Study:
 
     def tell(
         self,
-        trial: Union[trial_module.Trial, int],f
+        trial: Union[trial_module.Trial, int],
         values: Optional[Union[float, Sequence[float]]] = None,
         state: Optional[TrialState] = None,
         skip_if_finished: bool = False,
@@ -808,10 +808,11 @@ class Study:
                 Parameter values to pass your objective function.
 
                 .. note::
-                    This method raises ValueError for invalid parameter values like not a number (e.g. `np.nan`),
+                    This method raises ValueError for invalid parameter values
+                    like not a number (e.g. `np.nan`),
                     non-positive value when parameter is log-scale,
                     value which cannot be casted to float for int/float distribution,
-                    categorical value which does not be included in choices for categorical distribution, etc.
+                    value which does not be included in choices for categorical distribution, etc.
 
             user_attrs:
                 A dictionary of user-specific attributes other than ``params``.
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -808,9 +808,9 @@ class Study:
                 Parameter values to pass your objective function.
 
                 .. note::
-                    This method raises ValueError for invalid parameter values like nan,
+                    This method raises ValueError for invalid parameter values like not a number (e.g. `np.nan`),
                     non-positive value when parameter is log-scale,
-                    values which cannot be casted to float for int/float distribution,
+                    value which cannot be casted to float for int/float distribution,
                     categorical value which does not be included in choices for categorical distribution, etc.
 
             user_attrs:
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -806,14 +806,6 @@ class Study:
         Args:
             params:
                 Parameter values to pass your objective function.
-
-                .. note::
-                    This method raises ``ValueError`` for invalid parameter values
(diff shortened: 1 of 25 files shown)


tests/test_distributions.py
```diff
@@ -213,9 +213,21 @@ def test_int_internal_representation(value: Any) -> None:
...
+    "value, kwargs",
+    [
+        ("foo", {}),
+        ((), {}),
+        ([], {}),
+        ({}, {}),
+        (set(), {}),
+        (np.ones(2), {}),
+        (np.nan, {}),
+        (0, dict(log=True)),
```
knshnb, line 226:
```suggestion
        (0, dict(log=True)),
        (-1, dict(log=True)),
```

tests/test_distributions.py
```diff
@@ -234,9 +246,21 @@ def test_float_internal_representation(value: Any) -> None:
...
```
knshnb, line 259:
```suggestion
        (0.0, dict(log=True)),
        (-1.0, dict(log=True)),
```

tests/test_distributions.py
knshnb: How about adding a test case of negative values?

tests/test_distributions.py
knshnb: Ditto.

optuna/trial/_trial.py
```diff
@@ -597,6 +597,7 @@ def _suggest(self, name: str, distribution: BaseDistribution) -> Any:
                     study, trial, name, distribution
                 )
 
+            # param_value is validated here (invalid value like np.nan raises ValueError)
```
knshnb, line 600:
```suggestion
            # `param_value` is validated here (invalid value like np.nan raises ValueError).
```
Please also refer to https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#optuna-specific-conventions.

knshnb (commented):
Thanks for the PR. It looks good overall! Please take a look at some small comments I left.

HideakiImamura (dismissed):
LGTM except for @knshnb's comments.

cross32768:
Could someone check the updated document? (source: study.py)

optuna/study/study.py
```diff
@@ -806,6 +806,14 @@ def objective(trial):
         Args:
             params:
                 Parameter values to pass your objective function.
+
+                .. note::
+                    This method raises ``ValueError`` for invalid parameter values
+                    like not a number (e.g. ``np.nan``),
+                    non-positive value when parameter is log-scale,
+                    value which cannot be casted to float for int/float distribution,
+                    value which does not be included in choices for categorical distribution, etc.
```
HideakiImamura, line 815: The error does not occur in the call of `enqueue_trial`. The error occurs in the next call of `suggest_***`. There is some discussions as to where such documentation should be written. I think it should be in `suggest_***` or `***Distribution`. (I prefer the later.)

By the way, we have [a rule to document the error](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#exceptions).  I think we should NOT document the error when the parameter value is invalid, since it is obvious that such parameters are not acceptable. However, the obviousness is open for discussion.

cross32768: Thank you for your comment.
I agree to write such a kind of documentation in ``***Distribution``.

And sorry I had insufficient understanding for a rule to document the error.
I think error caused by not a number or values which cannot be casted to float/int is obvious, but non-positive value with log=True option and values does not be included in choices may require more discussion.

cross32768:
For a moment, I remove the added document

optuna/study/study.py
knshnb: I basically agree with the discussion. IMO, only it should be documented in the `log` argument of `***Distribution` that non-positive values are not allowed because this PR changes the behavior.

cross32768: I updated the code in response to this comment.

knshnb (approved):
Thanks for the update. LGTM.

PR #4142, opened 2022-11-06
Support GroupDecomposedSearchSpace in CmaEsSampler

zaburo-ch:
## Motivation

Modify CmaEsSampler to support GroupDecomposedSearchSpace, which is already supported by TPESampler and performance improvements have been reported.

## Description of the changes

 - Each group has its own optimizer and n_restarts.
 - They are stored in system_attr with different attr_name for each group.
 - When the group configuration changes, the prefix of attr_name changes and each optimizer and n_restarts are initialized.
   - Assume that the number of groups increases monotonically when the group configuration changes

### DIFF
--- a/optuna/samplers/_cmaes.py
+++ b/optuna/samplers/_cmaes.py
@@ -22,6 +22,8 @@ from optuna._transform import _SearchSpaceTransform
 from optuna.distributions import BaseDistribution
 from optuna.exceptions import ExperimentalWarning
 from optuna.samplers import BaseSampler
+from optuna.samplers._search_space.group_decomposed import _GroupDecomposedSearchSpace
+from optuna.samplers._search_space.group_decomposed import _SearchSpaceGroup
 from optuna.study._study_direction import StudyDirection
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
@@ -188,7 +190,8 @@ class CmaEsSampler(BaseSampler):
                 Added in v2.6.0 as an experimental feature. The interface may change in newer
                 versions without prior notice. See
                 https://github.com/optuna/optuna/releases/tag/v2.6.0.
-
+        group:
+            # TODO
     """
 
     def __init__(
@@ -206,6 +209,7 @@ class CmaEsSampler(BaseSampler):
         inc_popsize: int = 2,
         use_separable_cma: bool = False,
         source_trials: Optional[List[FrozenTrial]] = None,
+        group: bool = False,
     ) -> None:
         self._x0 = x0
         self._sigma0 = sigma0
@@ -220,6 +224,9 @@ class CmaEsSampler(BaseSampler):
         self._inc_popsize = inc_popsize
         self._use_separable_cma = use_separable_cma
         self._source_trials = source_trials
+        self._group = group
+        self._group_decomposed_search_space: Optional[_GroupDecomposedSearchSpace] = None
+        self._search_space_group: Optional[_SearchSpaceGroup] = None
 
         if self._restart_strategy:
             warnings.warn(
@@ -272,6 +279,14 @@ class CmaEsSampler(BaseSampler):
                 )
             )
 
+        if group:
+            warnings.warn(
+                "``group`` option is an experimental feature."
+                " The interface can change in the future.",
+                ExperimentalWarning,
+            )
+            self._group_decomposed_search_space = _GroupDecomposedSearchSpace()
+
     def reseed_rng(self) -> None:
         # _cma_rng doesn't require reseeding because the relative sampling reseeds in each trial.
         self._independent_sampler.reseed_rng()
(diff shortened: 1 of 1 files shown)


zaburo-ch:
Benchmark (https://github.com/optuna/optuna/pull/2526#issuecomment-817509298) results

```python
import optuna
import numpy as np

optuna.logging.set_verbosity(optuna.logging.ERROR)

def objective(trial):
    c = trial.suggest_categorical("c", ["A", "B"])
    dim = 5

    if c == "A":
        x = np.zeros(dim)
        for i in range(dim):
            x[i] = trial.suggest_float(f"x{i}", -3, 3)
        return np.exp(-np.linalg.norm(x - 1)-np.linalg.norm(x+1))
    else:
        y = np.zeros(dim)
        for i in range(dim):
            y[i] = trial.suggest_float(f"y{i}", -10, 10)
        return np.exp(-np.linalg.norm(y))

n_trials = 300

study1_best_values = []
study2_best_values = []

for i in range(100):
    sampler = optuna.samplers.CmaEsSampler(group=True)
    study1 = optuna.create_study(sampler=sampler, direction="maximize")
    study1.optimize(objective, n_trials=n_trials)

    sampler = optuna.samplers.CmaEsSampler()
    study2 = optuna.create_study(sampler=sampler, direction="maximize")
    study2.optimize(objective, n_trials=n_trials)

    study1_best_values.append(study1.best_value)
    study2_best_values.append(study2.best_value)

print(f"Best value: {np.mean(study1_best_values)} +/- {np.std(study1_best_values)} with group = True")
print(f"Best value: {np.mean(study2_best_values)} +/- {np.std(study2_best_values)} with group = False")
```
```
/Users/kaizaburo/workspace/optuna/optuna/samplers/_cmaes.py:286: ExperimentalWarning: ``group`` option is an experimental feature. The interface can change in the future.
  ExperimentalWarning,
Best value: 0.41092519527439586 +/- 0.2622665954277653 with group = True
Best value: 0.021568580549167528 +/- 0.030629298161601325 with group = False
```

optuna/samplers/_cmaes.py
```diff
@@ -387,7 +442,7 @@ def sample_relative(
         study._storage.set_trial_system_attr(
             trial._trial_id, generation_attr_key, optimizer.generation
         )
-        study._storage.set_trial_system_attr(trial._trial_id, "cma:n_restarts", n_restarts)
```
zaburo-ch, line 390: In the implementation of the master branch, when `use_separable_cma` is True, the keys for reading and writing n_restarts do not seem to be the same. Is this expected behavior?

zaburo-ch (commented):
I would be grateful if you could review if the current direction seems to be a suitable one.

HideakiImamura:
@gen740 will review this PR.

c-bata:
Also let me assign @eukaryo as a reviewer.

optuna/samplers/_cmaes.py
gen740: I'm not confident about expected behavior, but it seems to be weird that the `get` and `set` key is different.

In this case, this `set_trial_system_attr` will set the storage's `trial.system_attr`, and `get` (`_cmaes.py:412`) is getting from the same `trial.system_attr`, and restore optimizer.

If using `SEPCMA` (by setting `use_separable_cma=False`), trials were saved to `system_attr` with a key starts with `cma:` and restored with `sepcma:` so it seems this code (`_cmaes.py L417-424`)
```
            if not self._use_separable_cma and "cma:optimizer" in optimizer_attrs:
                # Check "cma:optimizer" key for backward compatibility.
                optimizer_str = optimizer_attrs["cma:optimizer"]
            else:
                optimizer_str = _concat_optimizer_attrs(optimizer_attrs)

            n_restarts: int = trial.system_attrs.get(attr_key_n_restarts, 0)
            return pickle.loads(bytes.fromhex(optimizer_str)), n_restarts
```
never be executed with `use_separable_cma=False`

gen740:
As I reviewed, [this code](https://github.com/optuna/optuna/blob/master/optuna/samplers/_cmaes.py#L417-L424) would be never executed with `use_separable_cma=False`, and optimizer maybe not properly restored (this seems to be resolved by this PR).

Is this an expected behavior?

optuna/samplers/_cmaes.py
gen740: And also it seems to be natural to branch [this attribute](https://github.com/zaburo-ch/optuna/blob/cmaes-group-decomposition-search-space/optuna/samplers/_cmaes.py#L344-L347) corresponding to `SepCMA` and `CMA`.

zaburo-ch:
Is it ok to keep going in the current direction and modify it so that it passes testing after https://github.com/optuna/optuna/pull/4184 is merged?

c-bata:
@zaburo-ch We are sorry for the delay in replying :bow: We fixed the system_attr's problem in another PR since we prioritized the release of CMA-ES with Margin :bow: In terms of changes for GroupDecomposedSearchSpace, it looks suitable. If you are too busy to resolve conflicts and continue working on this PR, please let us know so we can take it over to another collaborators or committers :pray:

zaburo-ch:
@c-bata Sorry for the late reply. I had missed your reply. I checked and there is a refactor in _attr_keys, so it will take some time to understand this and re-implement it. I will try to work on it at a later date if you don't mind taking the time, but if it needs to be done sooner, feel free to take over.

c-bata:
@zaburo-ch Huge thanks for your effort :bow: It doesn't need to be done sooner, so you can work on this at your own pace. Please let us know if you have any problems.

PR #4143, opened 2022-11-06
suggest_categorical: Add overloads for type narrowing

ConnorBaker:
## Motivation

When sampling from a categorical distribution where all elements share the same type, we expect the sampled value to have that type. Currently, this is not the case.

Given the values `[1,2,3,4]`, the value returned from `suggest_categorical` is `CategoricalChoiceType = Union[None, bool, int, float, str]`. This loss of type specificity manifests in the test suite in the use of `cast` and `assert` in an attempt to regain it.

## Description of the changes

This change adds overloads for each inhabitant of `CategoricalChoiceType` and allows us to drop the use of `cast` and `assert` in several places.

### DIFF
--- a/optuna/trial/_base.py
+++ b/optuna/trial/_base.py
@@ -3,6 +3,7 @@ import datetime
 from typing import Any
 from typing import Dict
 from typing import Optional
+from typing import overload
 from typing import Sequence
 
 from optuna._deprecated import deprecated_func
@@ -51,6 +52,38 @@ class BaseTrial(abc.ABC):
     def suggest_int(self, name: str, low: int, high: int, step: int = 1, log: bool = False) -> int:
 
         raise NotImplementedError
+
+    @overload
+    @abc.abstractmethod
+    def suggest_categorical(self, name: str, choices: Sequence[None]) -> None:
+        ...
+
+    @overload
+    @abc.abstractmethod
+    def suggest_categorical(self, name: str, choices: Sequence[bool]) -> bool:
+        ...
+
+    @overload
+    @abc.abstractmethod
+    def suggest_categorical(self, name: str, choices: Sequence[int]) -> int:
+        ...
+
+    @overload
+    @abc.abstractmethod
+    def suggest_categorical(self, name: str, choices: Sequence[float]) -> float:
+        ...
+
+    @overload
+    @abc.abstractmethod
+    def suggest_categorical(self, name: str, choices: Sequence[str]) -> str:
+        ...
+
+    @overload
+    @abc.abstractmethod
+    def suggest_categorical(
+        self, name: str, choices: Sequence[CategoricalChoiceType]
+    ) -> CategoricalChoiceType:
+        ...
 
     @abc.abstractmethod
     def suggest_categorical(
--- a/optuna/multi_objective/trial.py
+++ b/optuna/multi_objective/trial.py
@@ -3,6 +3,7 @@ from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import overload
 from typing import Sequence
 from typing import Union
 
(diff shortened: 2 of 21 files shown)


toshihikoyanase:
Thank you for your PR. I think this PR has the same difficulty as https://github.com/optuna/optuna/pull/4144#pullrequestreview-1169960509.

contramundum53:
Maybe we could use `TypeVar` to make the signature simpler?

xadrianzetx:
We've explored this possibility and it depends if `chocies` were intended to be a homogeneous sequence. I also prefer `TypeVar`, but it seems like in this case it might be too limiting, as mypy will complain if choices are of multiple types, like `[1, "fish", 2, "fish"]`.

contramundum53:
Would overloading with `CategoricalChoiceType` also result in such warning?

xadrianzetx:
Yes. Behavior is like that:
* This PR
```python
trial.suggest_categorical("x", choices=[1, 2, 3])  # Mypy ok, returns int
trial.suggest_categorical("x", choices=[b'foo', b'bar', b'baz'])  # Mypy fails, type not in CategoricalChoiceType
trial.suggest_categorical("x", choices=[1, "fish", 2, "fish"])  # Mypy ok, returns CategoricalChoiceType (like current master)
```

* Optuna-distributed#46
```python
trial.suggest_categorical("x", choices=[1, 2, 3])  # Mypy ok, returns int
trial.suggest_categorical("x", choices=[b'foo', b'bar', b'baz'])  # Mypy fails, type not in CategoricalChoiceType
trial.suggest_categorical("x", choices=[1, "fish", 2, "fish"])  # Mypy fails, all choices should be of the same type like the first one (int)
```

contramundum53:
I see, thanks for the investigation!

contramundum53:
I think we can merge this PR. What do you think? @toshihikoyanase

contramundum53:
We decided that we don't need that test. Let me reopen this PR.

contramundum53:

toshihikoyanase:
@ConnorBaker Could you merge the current master to your feature branch, please? We need to resolve the conflicts before merging this PR.

ConnorBaker:
Sorry about that, just rebased it!

ConnorBaker:
@toshihikoyanase does it look like it'll merge cleanly now?

toshihikoyanase:
I'm sorry for the delayed response. I'm on new year's holiday, and I'll check it tomorrow.

optuna/trial/_base.py
```diff
@@ -52,6 +53,38 @@ def suggest_int(self, name: str, low: int, high: int, step: int = 1, log: bool =
...
+    @overload
+    @abc.abstractmethod
+    def suggest_categorical(
+        self, name: str, choices: Sequence[CategoricalChoiceType]
+    ) -> CategoricalChoiceType:
+        ...
+
     @abc.abstractmethod
     def suggest_categorical(
         self, name: str, choices: Sequence[CategoricalChoiceType]
```
toshihikoyanase, line 90: [Note] When I read [PEP 484](https://peps.python.org/pep-0484/#function-method-overloading), I thought we could remove the type hints here, but I confirmed that we needed them actually. The example code in the [mypy document](https://mypy.readthedocs.io/en/stable/more_types.html) also has the hints.

toshihikoyanase (approved):
We may be able to apply `@overload` to `ChainerMNTrial` and `TorchDistributedTrial` in a follow-up PR. The CircleCI job failed due to some auth errors, but I don't think this PR affects visual regression tests.

toshihikoyanase:
[Notes] I confirmed that the CircleCI jobs passed in a follow-up PR https://github.com/optuna/optuna/pull/4300, and the [CI error](https://app.circleci.com/pipelines/github/optuna/optuna/10605) is not a blocker of this PR.

contramundum53 (approved):
LGTM! Thanks for your PR!

PR #4144, opened 2022-11-06
pruners: Use overloads to catch improper configurations as type errors

ConnorBaker:
## Motivation

Currently, improper arguments given to the pruners are only caught at runtime. Given the addition of `@overload` and `Literal` to `typing` and `typing_extensions`, we can provide warning about some improper uses of the pruners through the type checker.

## Description of the changes

This change adds a few overloads for the HyperBand, Successive Halving, and Threshold pruners to catch some invalid configurations at type checking time.

As an example:

```python
h1 = HyperbandPruner(1, "auto", 3, 0) # Works
# Argument of type Literal['cats'] cannot be assigned to parameter max_resource of type 
# int in function __init__
#   Literal['cats'] is incompatible with int
h2 = HyperbandPruner(1, "cats", 3, 0) # Type error! Not int or "auto".
# Argument of type Literal['cats'] cannot be assigned to parameter max_resource of type 
# int in function __init__
#   Literal['auto'] is incompatible with int
h3 = HyperbandPruner(1, "auto", 3, 1) # Type error! bootstrap_count>0 means we no auto
h4 = HyperbandPruner(max_resource="auto") # Works
h5 = HyperbandPruner(max_resource=1) # Works
```

### DIFF
--- a/optuna/pruners/_hyperband.py
+++ b/optuna/pruners/_hyperband.py
@@ -3,6 +3,8 @@ import math
 from typing import Container
 from typing import List
 from typing import Optional
+from typing import overload
+from typing import TYPE_CHECKING
 from typing import Union
 
 import optuna
@@ -11,6 +13,9 @@ from optuna.pruners._base import BasePruner
 from optuna.pruners._successive_halving import SuccessiveHalvingPruner
 from optuna.trial._state import TrialState
 
+
+if TYPE_CHECKING:
+    from typing_extensions import Literal
 
 _logger = logging.get_logger(__name__)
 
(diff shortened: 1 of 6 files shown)


toshihikoyanase (commented):
Thank you for your PR. I'm afraid we cannot apply the changes that require a new version of `typing_extension` since some ML libraries including `tensorflow` need older versions of `typing_extension`. Could you check the discussion in #3980 and #3990, please?

ConnorBaker:
It seems like the older version of `typing_extensions` supports `Literal` so I changed it back.

Although I wonder if I need to guard the import with `if TYPE_CHECKING` because I moved it out of the main requirements?

optuna/pruners/_hyperband.py
```diff
@@ -136,10 +141,33 @@ def objective(trial):
...
+        self,
+        min_resource: int = 1,
+        # NOTE: We must provide a defualt value or else we cannot write the arguments in order.
+        # This value is not used; if the case a default is not provided the original init is called
+        # and max_resource is set to "auto".
+        max_resource: int = 100,
+        reduction_factor: int = 3,
+        bootstrap_count: int = 0,
+    ) -> None:
+        ...
```
contramundum53, line 165: I'm sorry I'm not sure about the behavior of the type checker, but do we need these overloads? In principle the type `Union[Literal['auto'], int]` should be enough, isn't it?

tests/pruners_tests/test_successive_halving.py
```diff
@@ -253,10 +256,15 @@ def test_successive_halving_pruner_min_early_stopping_rate_parameter() -> None:
 def test_successive_halving_pruner_bootstrap_parameter() -> None:
 
     with pytest.raises(ValueError):
-        optuna.pruners.SuccessiveHalvingPruner(bootstrap_count=-1)
+        # Because the default min_resource is "auto", the bootstrap_count must be zero.
+        # The type checker correctly detects this type error.
+        optuna.pruners.SuccessiveHalvingPruner(bootstrap_count=-1)  # type: ignore
```
contramundum53, line 261: Maybe after we are able to detect these errors by type-checking, we could remove these tests according to the test policy.

contramundum53: @ConnorBaker We concluded that we can remove these invalid tests in this PR if we can detect them by type-checking. How about that?

contramundum53:
I think we can take over.

ConnorBaker:
Apologies, I haven't had much time as of late. Please do!

PR #2023, opened 2020-11-14
Add Dask integration

jrbourbeau:
## Motivation

This PR adds initial native Dask support for Optuna. See the discussion over in https://github.com/optuna/optuna/issues/1766 for additional context.

## Description of the changes

This PR adds a new `DaskStorage` class, which is adapted from https://github.com/jrbourbeau/dask-optuna, to Optuna

Still TODO:

- [x] Add documentation on `DaskStorage`
- [ ] ...

Closes https://github.com/optuna/optuna/issues/1766

PR #4146, opened 2022-11-07
Cache `study.directions` to reduce the number of `get_study_directions()` calls.

c-bata:
## Motivation

As discussed at https://github.com/optuna/optuna/pull/2023#issuecomment-1303069347, `study.directions` property method is called many times during the optimization. `study.directions` is set when creating a study via the `optuna.create_study(..., directions=...)` function and never be modified, so that we can obviously use `cached_property` and reduce the calls of `storage.get_study_directions()`.

## Description of the changes

~~Use `functools.cached_property` instead of `property`.~~

`functools.cached_property` seemed to be introduced in Python 3.8. So I pushed [1e145ca](https://github.com/optuna/optuna/pull/4146/commits/1e145caeb825ac62eff8013650202920fa185739).

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -203,7 +203,7 @@ class Study:
             A list of :class:`~optuna.study.StudyDirection` objects.
         """
 
-        if not self._directions:
+        if self._directions is None:
             self._directions = self._storage.get_study_directions(self._study_id)
         return self._directions
 
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1170,9 +1170,8 @@ def create_study(
         sampler = samplers.NSGAIISampler()
 
     study_name = storage.get_study_name_from_id(study_id)
+    storage.set_study_directions(study_id, direction_objects)
     study = Study(study_name=study_name, storage=storage, sampler=sampler, pruner=pruner)
-
-    study._storage.set_study_directions(study_id, direction_objects)
 
     return study
 
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -82,7 +82,7 @@ class Study:
         study_id = storage.get_study_id_from_name(study_name)
         self._study_id = study_id
         self._storage = storage
-        self._directions: Optional[List[StudyDirection]] = None
+        self._directions = storage.get_study_directions(study_id)
 
         self.sampler = sampler or samplers.TPESampler()
         self.pruner = pruner or pruners.MedianPruner()
@@ -203,8 +203,6 @@ class Study:
             A list of :class:`~optuna.study.StudyDirection` objects.
         """
 
-        if self._directions is None:
-            self._directions = self._storage.get_study_directions(self._study_id)
         return self._directions
 
     @property
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1,4 +1,5 @@
 import copy
+from functools import cached_property
 from numbers import Real
 import threading
 from typing import Any
@@ -194,7 +195,7 @@ class Study:
 
         return self.directions[0]
 
-    @property
+    @cached_property
     def directions(self) -> List[StudyDirection]:
         """Return the directions of the study.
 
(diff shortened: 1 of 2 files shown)


not522:
I'm not confident, but what about using `lru_cache`?

c-bata:
Thank you for the suggestion!
It sounds a bit not intuitive since this is a property method and there are no arguments. What do you think?

not522:
Hmm, I agree `lru_cache` is not intuitive.
On second thought, how about making `directions` the `Study`'s attribute? We can set it in `__init__`.

c-bata:
Sounds reasonable! One possible concern is that users may try to rewrite `study.directions` like `study.directions = [...]`. Do you think it's acceptable?

If it's not acceptable, we can set `study._directions` in `__init__` and provide users via the property method.

not522:
> If it's not acceptable, perhaps we can set `study._directions` in `__init__` and provide users via the property method.

I favor this plan.

optuna/study/study.py
```diff
@@ -82,6 +82,7 @@ def __init__(
         study_id = storage.get_study_id_from_name(study_name)
         self._study_id = study_id
         self._storage = storage
+        self._directions: Optional[List[StudyDirection]] = None
```
c-bata, line 85:
```suggestion
        self._directions = storage.get_study_directions(study_id)
```

optuna/study/study.py
```diff
@@ -202,7 +203,9 @@ def directions(self) -> List[StudyDirection]:
             A list of :class:`~optuna.study.StudyDirection` objects.
         """
 
-        return self._storage.get_study_directions(self._study_id)
+        if self._directions is None:
+            self._directions = self._storage.get_study_directions(self._study_id)
+        return self._directions
```
c-bata, line 208:
```suggestion
        return self._directions
```

c-bata:
PTAL.

not522 (approved):
LGTM

Note: we can assume users don't directly create `Study`. The document says

> Note that the direct use of this constructor is not recommended.

toshihikoyanase:
@contramundum53 Could you review this PR, please?

optuna/study/study.py
```diff
@@ -1169,10 +1170,9 @@ def objective(trial):
         sampler = samplers.NSGAIISampler()
 
     study_name = storage.get_study_name_from_id(study_id)
+    storage.set_study_directions(study_id, direction_objects)
```
contramundum53, line 1173: Almost LGTM, but I'm not confident about the consequence of moving this line before creating `Study`. Could you explain the motivation behind this?

contramundum53:
There seems to be a slight risk of creating an inconsistent state when more than two processes create the study at the same time. I don't think we need to address that, but maybe we should mention that in the document?

optuna/study/study.py
c-bata: In order to apply [this change](https://github.com/optuna/optuna/pull/4146#issuecomment-1308045371), we must ensure that directions must be set before instantiating `Study` object. Is my answer what you need?

c-bata:
> There seems to be a slight risk of creating an inconsistent state when more than two processes create the study at the same time with different directions. I don't think we need to address that, but maybe we should mention that in the document?

Could you tell me more details on how to create inconsistent studies with more than two processes?

optuna/study/study.py
contramundum53: I see, thanks!

contramundum53:
A race condition is hit when
1. Process A creates a study
2. Process B creates the same study
3. Process A sets a direction in the storage
4. Process B sets a different direction in the storage.
I think the consequence is that two processes continues optimization for different directions, which is not what we expect. (We expect an error in this case.)

contramundum53:
Well, is there currently any protection against inconsistent directions at all?
If not, maybe we could add that in the future.

contramundum53:
It seems that `Storage` has protection against inconsistent directions in it.

contramundum53 (approved):
LGTM! Thanks for the PR!

PR #4132, opened 2022-11-06
Add `BruteForceSampler`

semiexp:
This PR adds `BruteForceSampler`, which performs exhaustive search on the defined search space.
The implementation is based on [@not522 -san's](https://github.com/not522/optuna/commit/210062d955e7b6db4c809a641974af6ae95c6064) and some improvements take place in this PR.

PR #4152, opened 2022-11-09
Add `BruteForceSampler` in the samplers' list

not522:
## Motivation
Follow up of #4132.

## Description of the changes
Add `BruteForceSampler` in the samplers' list to generate the its document.


toshihikoyanase (approved):
I confirmed that `BruteForceSampler` is shown in the list of the samplers.
Since the change is obvious, let me merge this PR with a single approval.

image

PR #4157, opened 2022-11-10
Add Identity() to primary key columns

dycw:
## Motivation
As per #3365, Oracle does not support auto-incrementing by default, and the [official](https://docs.sqlalchemy.org/en/14/dialects/oracle.html#auto-increment-behavior) `sqlalchemy` way to handle this is the `Identity` column.

## Description of the changes
Added the `Identity` construct to each of the PK columns, with the exception of `VersionInfoModel` which is explicitly `autoincrement=False`.

Note: reading the contributions page, I cannot seem to get testing working.

```
✦ ❯ pytest
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.9.13, pytest-7.2.0, pluggy-1.0.0
rootdir: /home/derek/work/optuna, configfile: pyproject.toml
plugins: xdist-3.0.2
collected 4672 items / 14 errors

================================================================================================== ERRORS ==================================================================================================
________________________________________________________________________ ERROR collecting tests/integration_tests/test_catboost.py _________________________________________________________________________
ImportError while importing test module '/home/derek/work/optuna/tests/integration_tests/test_catboost.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../.pyenv/versions/3.9.13/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/integration_tests/test_catboost.py:3: in <module>
    import catboost as cb
E   ModuleNotFoundError: No module named 'catboost'
_________________________________________________________________________ ERROR collecting tests/integration_tests/test_chainer.py _________________________________________________________________________
ImportError while importing test module '/home/derek/work/optuna/tests/integration_tests/test_chainer.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../.pyenv/versions/3.9.13/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/integration_tests/test_chainer.py:7: in <module>
    import chainer
E   ModuleNotFoundError: No module named 'chainer'
________________________________________________________________________ ERROR collecting tests/integration_tests/test_chainermn.py ________________________________________________________________________
tests/integration_tests/test_chainermn.py:40: in <module>
    class Func:
tests/integration_tests/test_chainermn.py:45: in Func
    def __call__(self, trial: ChainerMNTrial, comm: CommunicatorBase) -> float:
E   NameError: name 'CommunicatorBase' is not defined
[...]

Issue #4149, opened 2022-11-08
SQLAlchemy & maximum number of expressions reached


dycw:
### Expected behavior

After ~1,800 trials, the SQLAlchemy queries being emitted are now running into a "maximum number of expressions" error.

In my own work, I have solved this for myself by chunking the values of the `in` clause using [`chunked`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.chunked), and then chaining the results together.

In Oracle, this is 1000. In MSSQL, this is 2100. In Postgres this is 32767. In SQLite this is 100.

If this is unintended, I can look toward implementing a patch.

### Environment

- Optuna version: 3.1.0.dev
- Python version: 3.9.13
- OS: Linux-3.10.0-1160.53.1.el7.x86_64-x86_64-with-glibc2.17
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
sqlalchemy.exc.DatabaseError: (cx_Oracle.DatabaseError) ORA-01795: maximum number of expressions in a list is 1000
[SQL: SELECT trials.trial_id AS trials_trial_id, trials."number" AS trials_number, trials.study_id AS trials_study_id, trials.state AS trials_state, trials.d
atetime_start AS trials_datetime_start, trials.datetime_complete AS trials_datetime_complete
FROM trials
WHERE trials.trial_id IN (:trial_id_1_1, :trial_id_1_2, :trial_id_1_3, ..., :trial_id_1_1880, :trial_id_1_1881, :trial_id_1_1882) AND trials.study_i
d = :study_id_1 ORDER BY trials.trial_id]
[parameters: {'study_id_1': 19, 'trial_id_1_1': 35, 'trial_id_1_2': ..., 'trial_id_1_1880': 1914, 'trial_
id_1_1881': 1915, 'trial_id_1_1882': 1916}]
(Background on this error at: https://sqlalche.me/e/14/4xp6)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/wande/.pyenv/versions/3.9.13/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/wande/.pyenv/versions/3.9.13/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/wande/work/stat-arb-signals/src/stat_arb/signals/components/axioma_optuna.py", line 478, in <module>
    main()
[...]

PR #4158, opened 2022-11-10
Chunk the queries so `sqlalchemy` does not breach the #params limit

dycw:
## Motivation
As per https://github.com/optuna/optuna/issues/4149, `optuna` is no longer able to retrieve trial information as the `in_` clause passes into many parameters.

## Description of the changes
This PR chunks the queries using [`chunked`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.chunked), a new dependency, so that no subquery exceeds the parameter limit. The limits are hard coded, for the 5 major dialects supported by `sqlalchemy`, and we take 95% of that so other parameters can pass through (really, it's only `models.TrialModel.study_id == study_id`). We default to 100 for non-supported dialects.

The `try` block is dropped, since the failover previously was designed to handle this. However, it is not general enough. @c-bata 's suggestion [here](https://github.com/optuna/optuna/issues/4149#issuecomment-1309662423) probably would work. In short, this PR drops the "query everything fallback" for "query multiple times for exactly what you need".

Note: as noted in [another](https://github.com/optuna/optuna/pull/4157) PR of mine, I cannot get testing working. However, this is the solution I implemented in a private fork I'm using at work.

Note 2: the private fork I implemented imports these dialects as `from sqlalchemy.dialects.mssql import dialect as mssql_dialect`. I'm not sure what the `_LazyImport` is so I'm just mimicking what I see here.

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 from contextlib import contextmanager
 import copy
 from datetime import datetime
+from itertools import chain
 import json
 import logging
 import os
@@ -16,6 +17,8 @@ from typing import Sequence
 from typing import Set
 from typing import TYPE_CHECKING
 import uuid
+
+import more_itertools
 
 import optuna
 from optuna import distributions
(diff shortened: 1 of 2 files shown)


optuna/storages/_rdb/storage.py
```diff
@@ -863,45 +876,38 @@ def _get_trials(
...
+            elif isinstance(dialect, sqlalchemy_dialects_mysql.dialect):
+                max_params = 65535
+            elif isinstance(dialect, sqlalchemy_dialects_oracle.dialect):
+                max_params = 1000
+            elif isinstance(dialect, sqlalchemy_dialects_postgresql.dialect):
+                max_params = 32767
+            elif isinstance(dialect, sqlalchemy_dialects_sqlite.dialect):
+                max_params = 100
+            else:
+                max_params = 100  # conservative estimate
```
gen740, line 905: Where do these params come from? Are these params solid?
I guess we should catch the exceptions.

PR #4159, opened 2022-11-10
Disable scheduled workflow runs in forks

xadrianzetx:
## Motivation

A proposition to disable scheduled workflow runs in forks by default, as failures could be too noisy. If accepted, I will also apply these changes to optuna-examples repo.

## Description of the changes

* Disabled scheduled workflow runs outside of `optuna/optuna`


HideakiImamura (approved):
LGTM.

PR #3628, opened 2022-06-04
Remove useless object inheritance

harupy:
Signed-off-by: harupy <hkawamura0130@gmail.com>

## Motivation

Inheritance from `object` is unnecessary in Python >= 3.

## Description of the changes

Remove useless object inheritance.

PR #4161, opened 2022-11-11
Remove object inheritance from base classes

xadrianzetx:
## Motivation

A small refresh of base class definitions, to follow the standards. Related to #3628.

## Description of the changes

* Removes explicit inheritance of `object` in base classes
* Avoids `metaclass` usage as [doc](https://docs.python.org/3/library/abc.html#abc.ABC) suggests.


c-bata (approved):
Good catch 👍

PR #4181, opened 2022-11-16
Trigger `checks-integration` action on PR creation and run mypy only.

c-bata:
## Motivation

Fix https://github.com/optuna/optuna/actions/runs/3477321803/jobs/5813363554

### DIFF
--- a/.github/workflows/checks-integration.yml
+++ b/.github/workflows/checks-integration.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+  pull_request: {}
   schedule:
     - cron: '0 23 * * SUN-THU'
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/checks-integration.yml
```diff
@@ -39,13 +40,6 @@ jobs:
         pip install --progress-bar off -U bayesmark
         pip install --progress-bar off -U kurobako
 
-    - name: black
-      run: black . --check --diff
-    - name: flake8
-      run: flake8 .
-    - name: isort
-      run: isort . --check --diff
```
c-bata, line 47: I removed these scenarios since they are coverred at `checks` action.

.github/workflows/checks-integration.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+  pull_request: {}
```
c-bata, line 7: I'm not sure why `checks-integration` has not been triggered on pull request. Perhaps it was just forgotten at https://github.com/optuna/optuna/pull/3580

not522: This is slow to run for PR and hard to maintain `Successful` status, so we decided to eliminate it at that time.

c-bata: I see. That sounds reasonable. Thank you for letting me know.

c-bata, line 7: TODO: Remove this line after fixing failures.

```suggestion
```

PR #4183, opened 2022-11-17
Refactor ParzenEstimator

contramundum53:
## Motivation
`optuna.samplers._tpe.parzen_estimator` had been difficult to read, and we try to make it simpler by splitting it into "probability distributions" part supporting sampling and log pdf function, and "Parzen estimator" part that returns the mixture distribution given the trials.

## Description of the changes
* Extract "probability distribution" part into `optuna.samplers._tpe.probability_distributions`.
* Refactor `optuna.samplers._tpe.parzen_estimator` to be simpler.

The tests are also simplified by splitting those two parts.


optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -54,424 +45,229 @@ def __init__(
...
-            if isinstance(dist, distributions.CategoricalDistribution):
-                mus = sigmas = None
-                categorical_weights = self._calculate_categorical_params(
-                    param_observations, param_name
+                assert parameters.prior_weight is not None
+                prior_weight = parameters.prior_weight
+            weights = np.append(
+                weights,
+                [1.0 if len(transformed_observations) == 0 else prior_weight],
+            )
```
HideakiImamura, line 75: If `parameters.prior_weight` must be provided when `parameters.consider_prior = True`, the following simplification is possible.
```python
prior_weight = 1.0 if len(transformed_observations) == 0 else parameters.prior_weight
if prior_weight is not None:
    weights = np.apend(weights, prior_weight)
```

optuna/samplers/_tpe/probability_distributions.py
```diff
@@ -0,0 +1,136 @@
...
+EPS = 1e-12
+
+
+def _normal_cdf(
+    x: Union[float, np.ndarray], mu: Union[float, np.ndarray], sigma: Union[float, np.ndarray]
+) -> np.ndarray:
+    return 0.5 * (1 + special.erf((x - mu) / np.maximum(np.sqrt(2) * sigma, EPS)))
+
+
+class _MixtureOfProductDistribution(NamedTuple):
```
HideakiImamura, line 53: This class has 2 methods, so instead of making this a `NamedTuple`, it would be better to make it a regular class.

HideakiImamura (commented):
Thanks for the great refactoring. Basically, LGTM. I have two comments. PTAL.

By the way, I concern the decrease of the test coverage. Could you try to improve it?

HideakiImamura:
I think we can try the followings, but it is beyond of the scope of this PR. They can be follow-up.
- I think we can use `_SearchSpaceTransform` if we fix it to handle the batch of distributions. 
- It would be great to hide the internal `***Batched***Distributions` from the parzen estimator, but that is debatable.

optuna/samplers/_tpe/parzen_estimator.py
contramundum53: When `parameters.prior_weight is not None` and `parameters.consider_prior == False`, this code wrongly adds prior.

contramundum53: Personally, I think we could just set `prior_weight == 0` when `consider_prior == False`. However, that will be another medium-sized refactoring, so let me tackle that in a follow-up.

optuna/samplers/_tpe/probability_distributions.py
contramundum53: @HideakiImamura Could you explain the reason why a `NamedTuple` with two methods would be bad?
@c-bata and I agreed that using `NamedTuple` would simplify the perceptual burden of reading the code. 
(There is another side effect of using `NamedTuple` that it's easier to compare two `_MixtureOfProductDistribution`.)

contramundum53:
> By the way, I concern the decrease of the test coverage. Could you try to improve it?

It looks like the coverage is decreasing in places other than `parzen_estimator.py`. I'm not sure why this is happening, could you tell me how to fix it?

contramundum53:
I don't think I have removed any of the tests involving those files.

optuna/samplers/_tpe/parzen_estimator.py
HideakiImamura: So, how about the following?

```python
prior_weight = 1.0 if len(transformed_observations) == 0 else parameters.prior_weight
if prior_weight is not None and parameters.consider_prior:
    weights = np.apend(weights, prior_weight)
```

optuna/samplers/_tpe/probability_distributions.py
HideakiImamura: I see. It is important that the `_MixtureOfProductDistribution` is immutable. Thanks. 

At first, a `NamedTuple` with a method seemed unnatural because it is only a named `Tuple`. I thought `NamedTuple` should behave as a container like `Dict`, but it is immutable. However, I think it is good because there is no inconvenience in using this form due to its great practical advantages.

HideakiImamura (commented):
Thanks for the explanation. I have another comment on tests. Could you take a look?

optuna/samplers/_tpe/parzen_estimator.py
contramundum53: This is also wrong because when `len(transformed_observations) == 0` and `parameters.consider_prior == False`, we *should* add prior.

contramundum53: I could have just written 
```
weights = np.append(
                weights,
                [1.0 if len(transformed_observations) == 0 else parameters.prior_weight],
            )
```
but the type checker didn't let me.

contramundum53: I simplified the logic here. PTAL.

tests/samplers_tests/tpe_tests/test_probability_distributions.py
```diff
@@ -0,0 +1,90 @@
...
+
+import numpy as np
+
+from optuna.samplers._tpe.probability_distributions import _BatchedCategoricalDistributions
+from optuna.samplers._tpe.probability_distributions import _BatchedDiscreteTruncNormDistributions
+from optuna.samplers._tpe.probability_distributions import _BatchedTruncNormDistributions
+from optuna.samplers._tpe.probability_distributions import _MixtureOfProductDistribution
+
+
+def test_mixture_of_product_distribution() -> None:
```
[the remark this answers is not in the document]
contramundum53: Currently we do not have `log=True` in `probability_distributions.py` layer.

contramundum53:
I pulled the latest master and the test coverage was fixed. The test coverage for `study.py` still drops but I don't think this is relevant to this PR.

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -54,424 +45,230 @@ def __init__(
...
-                    )
-                if not np.all(np.isfinite(w)):
-                    raise ValueError(
-                        "The `weights`function is not allowed to return infinite or NaN values "
-                        + f"{w}. The argument of the `weights` function is {n_observations}."
-                    )
-        else:
-            w = predetermined_weights[:n_observations]
+        if prior_weight is not None:
+            weights = np.append(weights, [prior_weight])
```
not522, line 76: How about simplify it like this?

```python
if len(transformed_observations) == 0:
    weights = np.ones(1)
elif parameters.consider_prior:
    weights = np.append(weights, [parameters.consider_prior])
```

not522 (commented):
Sorry for my late response. It's almost LGTM. I'm checking this PR does not break compatibility.

optuna/samplers/_tpe/parzen_estimator.py
contramundum53: Thanks for your suggestion. Fixed.

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -54,424 +45,225 @@ def __init__(
...
+            if parameters.multivariate:
+                SIGMA0_MAGNITUDE = 0.2
+                sigma = (
+                    SIGMA0_MAGNITUDE
+                    * max(len(observations), 1) ** (-1.0 / (len(self._search_space) + 4))
+                    * (high - low + step_or_0)
+                )
+                sigmas = np.full(shape=(len(observations),), fill_value=sigma)
+            else:
+                # Why include prior_mu???
```
HideakiImamura, line 221: It would be good to remove this comment or replace with a TODO comment.

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -54,424 +45,225 @@ def __init__(
...
-            if not consider_endpoints and sorted_mus_with_endpoints.shape[0] >= 4:
-                sorted_sigmas[0] = sorted_mus_with_endpoints[2] - sorted_mus_with_endpoints[1]
-                sorted_sigmas[-1] = sorted_mus_with_endpoints[-2] - sorted_mus_with_endpoints[-3]
+                sigmas = sorted_sigmas[np.argsort(sorted_indices)][: len(observations)]
 
-            sigmas[:] = sorted_sigmas[np.argsort(sorted_indices)]
+            # We adjust the range of the 'sigmas' according to the 'consider_magic_clip' flag.
+            maxsigma = 1.0 * (high - low + step_or_0)
+            if parameters.consider_magic_clip:
+                # Why change minsigma depending on consider_prior???
```
HideakiImamura, line 248: Ditto.

optuna/samplers/_tpe/probability_distributions.py
```diff
@@ -0,0 +1,136 @@
...
+
+class _BatchedDiscreteTruncNormDistributions(NamedTuple):
+    mu: np.ndarray
+    sigma: np.ndarray
+    low: float  # Currently, low, high and step do not change per trial.
+    high: float
+    step: float
+
+
+_BatchedDistributionUnion = Union[
```
HideakiImamura, line 38: This would be simple name.
```suggestion
_BatchedDistributions = Union[
```

HideakiImamura (commented):
Thanks for the update and answering my questions. Almost LGTM with several minor comments. PTAL.

optuna/samplers/_tpe/parzen_estimator.py
contramundum53, line 221:
```suggestion
                # TODO(contramundum53): Remove dependency on prior_mu
```

contramundum53: I applied this change. PTAL.

optuna/samplers/_tpe/parzen_estimator.py
contramundum53:
```suggestion
                # TODO(contramundum53): Remove dependency of minsigma on consider_prior.
```

contramundum53: I applied this change. PTAL.

HideakiImamura:
Thanks for the update. LGTM without CI errors.

HideakiImamura (approved):
LGTM.

not522 (approved):
LGTM. Sorry for my late response.

HideakiImamura:
@contramundum53 Could you resolve the conflicts? I will merge this PR after your resolving it.

c-bata:
I'm sorry for the delay in my review.

Since this PR has blocked #4326, let me unassign myself and create create follow-up PRs if I have find additional suggestions :bow:

c-bata:
Let me merge this PR since this PR has two approvals.

PR #4188, opened 2022-11-18
Deprecate `set_system_attr` in Study and Trial

cross32768:
## Motivation
As in [document of `set_system_attr`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html#optuna.study.Study.set_system_attr), this method is mainly used internally.
Therefore, I propose to hide from normal users.

## Description of the changes
* Change method name `set_system_attr` -> `_set_system_attr` in Study and Trial.


optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -272,10 +272,10 @@ def _postprocess(
...
             self.pbar.update(1)
 
-        trial.set_system_attr(_ELAPSED_SECS_KEY, elapsed_secs)
-        trial.set_system_attr(_AVERAGE_ITERATION_TIME_KEY, average_iteration_time)
-        trial.set_system_attr(_STEP_NAME_KEY, self.step_name)
-        trial.set_system_attr(_LGBM_PARAMS_KEY, json.dumps(self.lgbm_params))
+        trial._set_system_attr(_ELAPSED_SECS_KEY, elapsed_secs)
+        trial._set_system_attr(_AVERAGE_ITERATION_TIME_KEY, average_iteration_time)
+        trial._set_system_attr(_STEP_NAME_KEY, self.step_name)
+        trial._set_system_attr(_LGBM_PARAMS_KEY, json.dumps(self.lgbm_params))
```
c-bata, line 278: Could you call `storage.set_trial_system_attrs` instead?

optuna/trial/_trial.py
```diff
@@ -557,7 +557,7 @@ def objective(trial):
 
         self.storage.set_trial_user_attr(self._trial_id, key, value)
 
-    def set_system_attr(self, key: str, value: Any) -> None:
+    def _set_system_attr(self, key: str, value: Any) -> None:
```
c-bata, line 560: Could you avoid renaming `set_system_attr` to `_set_system_attr` since it's a breaking change. We should just deprecate it (by wrapping with a `@deprecated` decorator) and then update the docstring.

tests/study_tests/test_study.py
```diff
@@ -300,7 +300,7 @@ def test_study_set_and_get_system_attrs(storage_mode: str) -> None:
     with StorageSupplier(storage_mode) as storage:
         study = create_study(storage=storage)
 
-        study.set_system_attr("system_message", "test")
+        study._set_system_attr("system_message", "test")
```
c-bata, line 303: It might be a bit controversial but I'd say we can just remove test scenarios that checks the behavior of `set_system_attr`. What do you think?

c-bata (commented):
@cross32768  Thank you for your pull request!
I understand it's WIP but left some suggestions since I noticed while quickly taking a look.

tests/study_tests/test_study.py
cross32768: Will we continue to use `set_system_attr` internally for the future? If so, I think it's better to test.

optuna/trial/_trial.py
cross32768: I added @deprecated_func("3.1.0", "6.0.0") to `set_system_attr` method.

cross32768: For now, I set the timing of deprecation `"6.0.0"`. Is it reasonable choice? Or too far?

cross32768:
I wil add document about deprecation to Docstrings later.

optuna/integration/_lightgbm_tuner/optimize.py
c-bata: I'd say we can get the `storage` object via `trial._storage`. Although it's a private attribute, it's acceptable to use inside the optuna's implementation.

HideakiImamura:

tests/study_tests/test_study.py
c-bata: No, we will internally use `storage.set_trial_system_attr` instead of `trial.se_system_attr`. However, as you said, it might be better to leave tests for `trial.set_system_attr`.

@gen740 (cc: @HideakiImamura ) What do you think?

optuna/trial/_trial.py
c-bata: Sounds reasonable 👍 It follows our deprecation policy.
https://github.com/optuna/optuna/wiki/Deprecation-policy

HideakiImamura:
Let me convert this PR to draft. 

@cross32768 If you make this PR passing CI and remaining TODOs, please convert this ready for review.

optuna/integration/_lightgbm_tuner/optimize.py
cross32768: Thank you for you advice! (and sorry for late reply)

I update my code to use `trial._storage.set_trial_system_attrs(trial._trial_id, key, value)` instead of `trial.set_system_attrs(key, value)`.

I also use `study._storage.set_study_system_attrs(study._study_id, key, value)` instead of `study.set_system_attrs(key, value)`.

tests/study_tests/test_study.py
cross32768: I heard that deprecated function tests are also deprecated customarily, so I deleted them.

cross32768:
* Change point
  * I replaced `trial.set_system_attrs(key, value)` by `trial.storage.set_trial_system_attrs(trial._trial_id, key, value)`
  * I also replaced `study.set_system_attrs(key, value)` by `study._storage.set_study_system_attrs(study._study_id, key, value)`
  * I removed some tests about `system_attrs`

cross32768:
* Current problem & discusion
  * Current code cannot pass doc test. But I cannot read which change is the problem from the error. What should I do?
  * `FrozenTrial` class does not have `storage`. Thus, I cannot replace `FrozenTrial.set_system_attr` with other way. 
  * Should I add the test for correct deprecation warning?
  * Should I also make`set_system_attrs` deprecated? This change make big effect to a lot of files (over 50 files).

optuna/multi_objective/study.py
```diff
@@ -373,6 +375,7 @@ def set_user_attr(self, key: str, value: Any) -> None:
 
         self._study.set_user_attr(key, value)
 
+    @deprecated_func("3.1.0", "6.0.0")
```
c-bata, line 378: nit: We can omit the deprecation message here since multi_objective module itself is deprecated.
```suggestion
```

optuna/multi_objective/trial.py
```diff
@@ -162,14 +163,15 @@ def set_user_attr(self, key: str, value: Any) -> None:
 
         self._trial.set_user_attr(key, value)
 
+    @deprecated_func("3.1.0", "6.0.0")
```
c-bata, line 166: ditto

```suggestion
```

c-bata (commented):
LGTM with nits.

c-bata (approved):

c-bata:
Let me change the reviewers from @gen740 to @HideakiImamura since Fujimoto-san is busy this week.

c-bata:

optuna/trial/_trial.py
not522: By default, we will remove it in v5.0.0 (two major versions after v3.1.0). But I agree to defer the version because users seem to use it, and the maintenance cost is low.

not522 (approved):
Thank you for your PR. LGTM.

abraaoz:
I use this method as follows:
```py
trial_note: str = "markdown here"
study_id: int = getattr(study, "_study_id")
trial_id = trial.storage.get_trial_id_from_study_id_trial_number(study_id, trial.number)
study.set_system_attr(f"dashboard:{trial_id}:note_ver", 1)
study.set_system_attr(f"dashboard:{trial_id}:note_str:0", trial_note)
```
How can I achieve the same result without using `set_system_attr`?
User attributes don’t appear as markdown in Optuna Dashboard.

c-bata:
@abraaoz You can use the `optuna_dashboard.save_note()` function.
https://optuna-dashboard.readthedocs.io/en/latest/_generated/optuna_dashboard.save_note.html#optuna_dashboard.save_note

For further questions, please use [GitHub Discussions](https://github.com/optuna/optuna/discussions/categories/q-a) instead.

PR #4189, opened 2022-11-18
Add a `directions` arg to `storage.create_new_study`

gen740:
## Motivation

Now, optuna's storage has a member function `set_study_directions`. But it is not necessary, because the study's `directions` is not changeable after creating the study.

So, I create a `directions` arg on `storage.create_new_study` and delete `set_study_directions`.

## Description of the changes

- Add `directions` arg to `storage.create_new_study` for all storages.
- Delete `set_study_directions` members from all storages including `dask`.
- Delete `SET_STUDY_DIRECTIONS` from `JournalOperation`.
- Rewrite `test_set_and_get_study_directions`.

### DIFF
--- a/optuna/integration/dask.py
+++ b/optuna/integration/dask.py
@@ -67,7 +67,6 @@ class _OptunaSchedulerExtension:
             "delete_study",
             "set_study_user_attr",
             "set_study_system_attr",
-            "set_study_directions",
             "get_study_id_from_name",
             "get_study_name_from_id",
             "get_study_directions",
@@ -99,9 +98,12 @@ class _OptunaSchedulerExtension:
         self,
         comm: "distributed.comm.tcp.TCP",
         storage_name: str,
+        directions: Sequence[StudyDirection],
         study_name: Optional[str] = None,
     ) -> int:
-        return self.get_storage(storage_name).create_new_study(study_name=study_name)
+        return self.get_storage(storage_name).create_new_study(
+            directions=directions, study_name=study_name
+        )
 
     def delete_study(
         self,
@@ -135,18 +137,6 @@ class _OptunaSchedulerExtension:
             study_id=study_id,
             key=key,
             value=value,
-        )
-
-    def set_study_directions(
-        self,
-        comm: "distributed.comm.tcp.TCP",
-        storage_name: str,
-        study_id: int,
-        directions: List[str],
-    ) -> None:
-        return self.get_storage(storage_name).set_study_directions(
-            study_id=study_id,
-            directions=[StudyDirection[direction] for direction in directions],
         )
 
     def get_study_id_from_name(
@@ -445,11 +435,14 @@ class DaskStorage(BaseStorage):
 
         return self.client.run_on_scheduler(_get_base_storage, name=self.name)
 
-    def create_new_study(self, study_name: Optional[str] = None) -> int:
+    def create_new_study(
+        self, directions: Sequence[StudyDirection], study_name: Optional[str] = None
+    ) -> int:
         return self.client.sync(
             self.client.scheduler.optuna_create_new_study,
             storage_name=self.name,
             study_name=study_name,
+            directions=directions,
         )
 
     def delete_study(self, study_id: int) -> None:
@@ -475,14 +468,6 @@ class DaskStorage(BaseStorage):
             study_id=study_id,
             key=key,
             value=value,
-        )
-
-    def set_study_directions(self, study_id: int, directions: Sequence[StudyDirection]) -> None:
-        return self.client.sync(
-            self.client.scheduler.optuna_set_study_directions,
-            storage_name=self.name,
-            study_id=study_id,
-            directions=[direction.name for direction in directions],
         )
 
     # Basic study access
--- a/optuna/integration/dask.py
+++ b/optuna/integration/dask.py
@@ -98,11 +98,12 @@ class _OptunaSchedulerExtension:
         self,
         comm: "distributed.comm.tcp.TCP",
         storage_name: str,
-        directions: Sequence[StudyDirection],
+        directions: List[str],
         study_name: Optional[str] = None,
     ) -> int:
         return self.get_storage(storage_name).create_new_study(
-            directions=directions, study_name=study_name
+            directions=[StudyDirection[direction] for direction in directions],
+            study_name=study_name,
         )
 
     def delete_study(
@@ -442,7 +443,7 @@ class DaskStorage(BaseStorage):
             self.client.scheduler.optuna_create_new_study,
             storage_name=self.name,
             study_name=study_name,
-            directions=directions,
+            directions=[direction.name for direction in directions],
         )
 
     def delete_study(self, study_id: int) -> None:
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -50,7 +50,9 @@ class BaseStorage(abc.ABC):
     # Basic study manipulation
 
     @abc.abstractmethod
-    def create_new_study(self, study_name: Optional[str] = None) -> int:
+    def create_new_study(
+        self, directions: Sequence[StudyDirection], study_name: Optional[str] = None
+    ) -> int:
         """Create a new study from a name.
 
         If no name is specified, the storage class generates a name.
@@ -121,27 +123,6 @@ class BaseStorage(abc.ABC):
         Raises:
             :exc:`KeyError`:
                 If no study with the matching ``study_id`` exists.
-        """
-        raise NotImplementedError
-
-    @abc.abstractmethod
-    def set_study_directions(self, study_id: int, directions: Sequence[StudyDirection]) -> None:
-        """Register optimization problem directions to a study.
-
-        Args:
-            study_id:
-                ID of the study.
-            directions:
-                A sequence of direction whose element is either
-                :obj:`~optuna.study.StudyDirection.MAXIMIZE` or
-                :obj:`~optuna.study.StudyDirection.MINIMIZE`.
-
-        Raises:
-            :exc:`KeyError`:
-                If no study with the matching ``study_id`` exists.
-            :exc:`ValueError`:
-                If the directions are already set and the each coordinate of passed ``directions``
-                is the opposite direction or :obj:`~optuna.study.StudyDirection.NOT_SET`.
         """
         raise NotImplementedError
 
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -59,6 +59,10 @@ class BaseStorage(abc.ABC):
         The returned study ID is unique among all current and deleted studies.
 
         Args:
+            directions:
+                 A sequence of direction whose element is either
+                 :obj:`~optuna.study.StudyDirection.MAXIMIZE` or
+                 :obj:`~optuna.study.StudyDirection.MINIMIZE`.
             study_name:
                 Name of the new study to create.
 
--- a/optuna/storages/_cached_storage.py
+++ b/optuna/storages/_cached_storage.py
@@ -30,7 +30,7 @@ class _StudyInfo:
         self.owned_or_finished_trial_ids: Set[int] = set()
         # Cache distributions to avoid storage access on distribution consistency check.
         self.param_distribution: Dict[str, distributions.BaseDistribution] = {}
-        self.directions: List[StudyDirection] = [StudyDirection.NOT_SET]
+        self.directions: List[StudyDirection] = []
         self.name: Optional[str] = None
 
 
@@ -154,7 +154,7 @@ class _CachedStorage(BaseStorage, BaseHeartbeat):
         with self._lock:
             if study_id in self._studies:
                 directions = self._studies[study_id].directions
-                if len(directions) > 1 or directions[0] != StudyDirection.NOT_SET:
+                if len(directions) > 1:
                     return directions
 
         directions = self._backend.get_study_directions(study_id)
--- a/optuna/storages/_cached_storage.py
+++ b/optuna/storages/_cached_storage.py
@@ -30,7 +30,7 @@ class _StudyInfo:
         self.owned_or_finished_trial_ids: Set[int] = set()
         # Cache distributions to avoid storage access on distribution consistency check.
         self.param_distribution: Dict[str, distributions.BaseDistribution] = {}
-        self.directions: List[StudyDirection] = []
+        self.directions: Optional[List[StudyDirection]] = None
         self.name: Optional[str] = None
 
 
@@ -104,6 +104,7 @@ class _CachedStorage(BaseStorage, BaseHeartbeat):
         with self._lock:
             study = _StudyInfo()
             study.name = study_name
+            study.directions = list(directions)
             self._studies[study_id] = study
 
         return study_id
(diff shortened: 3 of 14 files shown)


optuna/storages/_rdb/storage.py
```diff
@@ -255,7 +255,9 @@ def __setstate__(self, state: Dict[Any, Any]) -> None:
         if not self.skip_compatibility_check:
             self._version_manager.check_table_schema_compatibility()
 
-    def create_new_study(self, study_name: Optional[str] = None) -> int:
+    def create_new_study(
+        self, directions: Sequence[StudyDirection], study_name: Optional[str] = None
```
c-bata, line 259: Could you leave a todo comment to make this operation atomic?

optuna/storages/_base.py
```diff
@@ -50,7 +50,9 @@ class BaseStorage(abc.ABC):
     # Basic study manipulation
 
     @abc.abstractmethod
-    def create_new_study(self, study_name: Optional[str] = None) -> int:
+    def create_new_study(
+        self, directions: Sequence[StudyDirection], study_name: Optional[str] = None
```
c-bata, line 54: Could you update the docstring since `directions` argument is added?

optuna/storages/_journal/storage.py
```diff
@@ -135,11 +134,15 @@ def _sync_with_backend(self) -> None:
...
-    def create_new_study(self, study_name: Optional[str] = None) -> int:
+    def create_new_study(
+        self, directions: Sequence[StudyDirection], study_name: Optional[str] = None
+    ) -> int:
         study_name = study_name or DEFAULT_STUDY_NAME_PREFIX + str(uuid.uuid4())
 
         with self._thread_lock:
-            self._write_log(JournalOperation.CREATE_STUDY, {"study_name": study_name})
+            self._write_log(
+                JournalOperation.CREATE_STUDY, {"study_name": study_name, "directions": directions}
```
c-bata, line 144: Here you added `directions` attribute in the `CREATE_STUDY` log, so that you must also update the implementation of `_apply_create_new_study`.

c-bata (changes requested):
Thank you for the swift PR. I left two suggestions.

optuna/storages/_rdb/storage.py
gen740: I will make this operator atomic.

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM with several comments. Could you take a look?

optuna/integration/dask.py
```diff
@@ -99,9 +98,13 @@ def create_new_study(
         self,
         comm: "distributed.comm.tcp.TCP",
         storage_name: str,
+        directions: List[str],
```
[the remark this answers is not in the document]
gen740: I think so, but the original version of dask's `_OptunaSchedulerExtension.set_study_directions` takes `List[str]`, and if I change it will be inconsistent with `get_study_directions` which returns `List[str]`.

Would it be better to change these types?

optuna/storages/_cached_storage.py
```diff
@@ -30,7 +30,7 @@ def __init__(self) -> None:
         self.owned_or_finished_trial_ids: Set[int] = set()
         # Cache distributions to avoid storage access on distribution consistency check.
         self.param_distribution: Dict[str, distributions.BaseDistribution] = {}
-        self.directions: List[StudyDirection] = [StudyDirection.NOT_SET]
+        self.directions: List[StudyDirection] = []
```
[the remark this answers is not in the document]
gen740: I guess this value will be overwritten in the `create_new_study` function. So, I think it is better to use `Optional[List[StudyDirection]]` as `name` does.
I realize that I forgot to set `study.direction` in `create_new_study`.

optuna/storages/_cached_storage.py
```diff
@@ -168,7 +154,7 @@ def get_study_directions(self, study_id: int) -> List[StudyDirection]:
         with self._lock:
             if study_id in self._studies:
                 directions = self._studies[study_id].directions
-                if len(directions) > 1 or directions[0] != StudyDirection.NOT_SET:
+                if len(directions) > 1:
```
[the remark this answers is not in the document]
gen740: I want to `None` check as I mentioned.

optuna/storages/_journal/storage.py
```diff
@@ -156,6 +159,9 @@ def create_new_study(self, study_name: Optional[str] = None) -> int:
                     and study_id % SNAPSHOT_INTERVAL == 0
                 ):
                     self._backend.save_snapshot(pickle.dumps(self._replay_result))
+
+                self._sync_with_backend()
```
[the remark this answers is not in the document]
gen740: Oh, I see. It seems to do nothing.

optuna/integration/dask.py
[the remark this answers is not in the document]
c-bata: I would say it's better to use `List[str]` here since each storage implementation would be a bit simpler by restricting the type.

optuna/study/_study_direction.py
```diff
@@ -5,14 +5,11 @@ class StudyDirection(enum.IntEnum):
...
     Attributes:
-        NOT_SET:
-            Direction has not been set.
         MINIMIZE:
             :class:`~optuna.study.Study` minimizes the objective function.
         MAXIMIZE:
             :class:`~optuna.study.Study` maximizes the objective function.
     """
 
-    NOT_SET = 0
```
c-bata, line 16: How about keeping `NOT_SET` since it exists until we provide the data migration for RDBStorages?

gen740: I see.

gen740:
@c-bata @HideakiImamura A friendly reminder :)

optuna/storages/_cached_storage.py
```diff
@@ -96,13 +96,17 @@ def __setstate__(self, state: Dict[Any, Any]) -> None:
         self.__dict__.update(state)
         self._lock = threading.Lock()
 
-    def create_new_study(self, study_name: Optional[str] = None) -> int:
+    def create_new_study(
+        self, directions: Sequence[StudyDirection], study_name: Optional[str] = None
+    ) -> int:
 
-        study_id = self._backend.create_new_study(study_name)
+        study_id = self._backend.create_new_study(directions=directions, study_name=study_name)
```
c-bata, line 103: How about adding the following test case in `test_with_server.py`?

```diff
$ git diff
diff --git a/tests/storages_tests/test_with_server.py b/tests/storages_tests/test_with_server.py
index 18b4ae481..fecb2fd31 100644
--- a/tests/storages_tests/test_with_server.py
+++ b/tests/storages_tests/test_with_server.py
@@ -1,4 +1,4 @@
-from concurrent.futures import ProcessPoolExecutor
+from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
 import os
 import pickle
 from typing import Sequence
@@ -148,7 +148,15 @@ def test_store_nan_intermediate_values(storage: BaseStorage) -> None:
     assert np.isnan(got_value)


-def test_multiprocess(storage: BaseStorage) -> None:
+def test_multithread_create_study(storage: BaseStorage) -> None:
+    with ThreadPoolExecutor(10) as pool:
+        for _ in range(10):
+            pool.submit(optuna.create_study, storage=storage, study_name="test-multithread-create-study", load_if_exists=True)
+
+    assert len(storage.get_all_studies()) == 1
+
+
+def test_multiprocess_run_optimize(storage: BaseStorage) -> None:
     n_workers = 8
     n_trials = 20
     study_name = _STUDY_NAME
```

optuna/storages/_in_memory.py
```diff
@@ -129,7 +118,9 @@ def get_study_directions(self, study_id: int) -> List[StudyDirection]:
 
         with self._lock:
             self._check_study_id(study_id)
-            return self._studies[study_id].directions
+            directions = self._studies[study_id].directions
+            assert directions is not None  # Study directions are never `None`.
```
c-bata, line 122: Do we need this assertion? At least this PR does not seem to increase the possibility of `study.directions` being None here.

optuna/storages/_in_memory.py
```diff
@@ -258,9 +249,13 @@ def get_best_trial(self, study_id: int) -> FrozenTrial:
...
 
             best_trial_id = self._studies[study_id].best_trial_id
+
+            directions = self._studies[study_id].directions
+            assert directions is not None  # Study directions are never `None`.
+
             if best_trial_id is None:
                 raise ValueError("No trials are completed yet.")
-            elif len(self._studies[study_id].directions) > 1:
+            elif len(directions) > 1:
```
c-bata, line 258: ditto.

c-bata (changes requested):
Thanks for the update! Changes almost good to me. I left some suggestions though.

optuna/storages/_cached_storage.py
gen740: Sounds good! I will add this test.

optuna/storages/_in_memory.py
gen740: It's a type-checking thing, `pyright` LSP will cause an issue `directions are possibly None` so I add it. Of course, it's a bit redundant.

optuna/storages/_cached_storage.py
gen740: This test [failed](https://github.com/optuna/optuna/actions/runs/3704495678/jobs/6277185743#step:9:33)...

optuna/storages/_in_memory.py
```diff
@@ -63,7 +65,10 @@ def create_new_study(self, study_name: Optional[str] = None) -> int:
             else:
                 study_uuid = str(uuid.uuid4())
                 study_name = DEFAULT_STUDY_NAME_PREFIX + study_uuid
-            self._studies[study_id] = _StudyInfo(study_name)
+            study = _StudyInfo(study_name)
+            study.directions = list(directions)
+
+            self._studies[study_id] = study
```
HideakiImamura, line 71: How about giving `directions` to `_StudyInfo` directly? We need to change the definition of `_StudyInfo`.
```suggestion
            self._studies[study_id] =  _StudyInfo(study_name, list(directions))
```

optuna/storages/_in_memory.py
HideakiImamura: You can avoid this assertion by re-defining the `_StudyInfo` as follow.
```python
class _StudyInfo:
    def __init__(self, name: str, directions: List[StudyDirection]) -> None:
        self.trials: List[FrozenTrial] = []
        self.param_distribution: Dict[str, distributions.BaseDistribution] = {}
        self.user_attrs: Dict[str, Any] = {}
        self.system_attrs: Dict[str, Any] = {}
        self.name: str = name
        self.directions: List[StudyDirection] = directions
        self.best_trial_id: Optional[int] = None
```

tests/storages_tests/test_with_server.py
```diff
@@ -147,7 +149,20 @@ def test_store_nan_intermediate_values(storage: BaseStorage) -> None:
...
+    with ThreadPoolExecutor(10) as pool:
+        for _ in range(10):
+            pool.submit(
+                optuna.create_study,
+                storage=storage,
+                study_name="test-multithread-create-study",
+                load_if_exists=True,
+            )
+
+    assert len(storage.get_all_studies()) == 1
```
HideakiImamura, line 162: The above tests have created some storages, so we need to refresh DB before this test case or just remove the last assertion. I prefer the latter.

HideakiImamura (commented):
Almost LGTM. I have several comments. Could you take a look?

tests/storages_tests/test_with_server.py
gen740: Hmmm, instead of removing the last assertion. How about `assert len(storage.get_all_studies()) >= 1`?

optuna/storages/_in_memory.py
gen740: Thanks for the solution!

tests/storages_tests/test_with_server.py
c-bata: I'm sorry that my suggested test case is broken.

I agree with Imamura-san's comment. It may cause confusion to readers because the intention of the last assertion is not clear.

optuna/storages/_in_memory.py
```diff
@@ -63,7 +65,9 @@ def create_new_study(self, study_name: Optional[str] = None) -> int:
             else:
                 study_uuid = str(uuid.uuid4())
                 study_name = DEFAULT_STUDY_NAME_PREFIX + study_uuid
-            self._studies[study_id] = _StudyInfo(study_name)
+            study = _StudyInfo(study_name, list(directions))
+
+            self._studies[study_id] = study
```
c-bata, line 70: [nit] You can write it like:
```suggestion
            self._studies[study_id] = _StudyInfo(study_name, list(directions))
```

c-bata (commented):
Looks almost good to me. I left some nit-picking comments. I will approve this PR after https://github.com/optuna/optuna/pull/4189#discussion_r1050435247 will be resolved.

tests/storages_tests/test_storages.py
```diff
@@ -161,70 +164,44 @@ def test_get_study_id_from_name_and_get_study_name_from_id(storage_mode: str) ->
             storage.get_study_name_from_id(study_id + 1)
 
 
+# TODO(gen740): write a test create_study directions and get_study_directions
```
HideakiImamura, line 167: Is this TODO comment addressed, or left as a follow-up?

HideakiImamura (commented):
Thanks for the update. Almost, LGTM with a minor comment.

tests/storages_tests/test_with_server.py
HideakiImamura: The number of studies is already more than 1 before starting this test case, so that assertion is a bit strange. Can we just remove the assertion?

gen740: Ok, I see.

tests/storages_tests/test_storages.py
gen740: Oh..., I forgot to delete this comment.

tests/storages_tests/test_storages.py
```diff
@@ -166,65 +169,38 @@ def test_set_and_get_study_directions(storage_mode: str) -> None:
...
+            (StudyDirection.MAXIMIZE, StudyDirection.MINIMIZE),
+            [StudyDirection.MINIMIZE, StudyDirection.MAXIMIZE],
+            [StudyDirection.MAXIMIZE, StudyDirection.MINIMIZE],
         ]:
-            study_id = storage.create_new_study()
+            study_id = storage.create_new_study(directions=target)
 
-            def check_set_and_get(directions: Sequence[StudyDirection]) -> None:
-                storage.set_study_directions(study_id, directions)
+            def check_get() -> None:
```
c-bata, line 182: [nit] This is perhaps an out-of-scope comment, but I think we don't need to define an inner function here.
`check_get()` function cannot reduce the number of lines and cannot make this test readable.

We can work it in the following pull request.

c-bata (approved):

HideakiImamura (approved):
Thanks for the update. LGTM! 

This can be done as a follow-up. https://github.com/optuna/optuna/pull/4189/files#r1054107362

Issue #3511, opened 2022-04-25
Fix test error in `tests/integration_tests/test_catboost.py::test_catboost_pruning_callback_errors`


nzw0301:
### Expected behavior

No error when running pytest:

```sh
pytest tests/integration_tests/test_catboost.py::test_catboost_pruning_callback_errors
```

### Environment

- Optuna version:3.0.0b1.dev
- Python version: 3.8.13
- OS:macOS-10.16-x86_64-i386-64bit
- (Optional) Other libraries and their versions:
    - catboost: 1.0.5

### Error messages, stack traces, or logs

```shell
opt/homebrew/Caskroom/miniconda/base/envs/py38/lib/python3.8/traceback.py:481: in __init__
    cause = TracebackException(
/opt/homebrew/Caskroom/miniconda/base/envs/py38/lib/python3.8/traceback.py:481: in __init__
    cause = TracebackException(
/opt/homebrew/Caskroom/miniconda/base/envs/py38/lib/python3.8/traceback.py:481: in __init__
    cause = TracebackException(
/opt/homebrew/Caskroom/miniconda/base/envs/py38/lib/python3.8/traceback.py:481: in __init__
    cause = TracebackException(
/opt/homebrew/Caskroom/miniconda/base/envs/py38/lib/python3.8/traceback.py:481: in __init__
    cause = TracebackException(
/opt/homebrew/Caskroom/miniconda/base/envs/py38/lib/python3.8/traceback.py:481: in __init__
    cause = TracebackException(
/opt/homebrew/Caskroom/miniconda/base/envs/py38/lib/python3.8/traceback.py:481: in __init__
    cause = TracebackException(
/opt/homebrew/Caskroom/miniconda/base/envs/py38/lib/python3.8/traceback.py:481: in __init__
    cause = TracebackException(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <traceback.TracebackException object at 0x7fdb61bd5e80>, exc_type = <class 'SystemError'>
exc_value = SystemError("<class 'types.SimpleNamespace'> returned a result with an error set")
exc_traceback = <traceback object at 0x7fdb61cb6c00>

[...]

PR #4190, opened 2022-11-18
Fix a test error in `test_catboost.py`

Alnusjaponica:
## Motivation
Resolve [#3511](https://github.com/optuna/optuna/issues/3511)

## Description of the changes
Remove `pytest.mark.skip` of `tests/integration_tests/test_catboost.py::test_catboost_pruning_callback_errors`.  
When python version is lower than 3.9 and  pytest version is lower than 7.2.0, the resulting error type in `tests/integration_tests/test_catboost.py::test_catboost_pruning_callback_errors` is different from other cases. This is mainly because the _Py_CheckRecursionLimit variable used in from limited C API was removed after python 3.9 and that affect when calling function _CallbackAfterIteration in Catboost, which is written in Cython. Therefore, I added a block to deal these cases separately.


Alnusjaponica:
## Alternative Approach

Currently we the test function treat RecursionError and SystemError differently according to the version of Python and pytest. However, what the test case actually covers is errors in user's code and it seems that just terminating by an error is enough  for this case. As an alternative approach we can just catch error without distinguishing the error type.

tests/integration_tests/test_catboost.py
```diff
@@ -123,6 +123,17 @@ def objective(trial: optuna.trial.Trial) -> float:
 
     # Unknown validation name or metric.
     study = optuna.create_study(pruner=DeterministicPruner(False))
-    # Catboost terminates with a SystemError.
-    with pytest.raises(SystemError):
-        study.optimize(objective, n_trials=1)
+    # Catboost terminates with a SystemError when python>=3.9 or pytest>=7.2.0, otherwise terminates with RecursionError.
+    # This is mainly because the _Py_CheckRecursionLimit variable used in from limited C API was removed after python 3.9.
+    # That affect when calling function _CallbackAfterIteration in Catboost, which is written in Cython.
```
contramundum53, line 128:
```suggestion
    # Catboost terminates with a SystemError when python>=3.9 or pytest>=7.2.0,
    # otherwise terminates with RecursionError.
    # This is mainly because the _Py_CheckRecursionLimit variable used in from
    # limited C API was removed after python 3.9.
    # That affect when calling function _CallbackAfterIteration in Catboost,
    # which is written in Cython.
```
It seems that the CI is failing due to the long comment.

contramundum53:
@not522 I'm not sure whether introducing this kind of test that relies on undocumented/unstable behavior of the depending library is desirable. Could you also review this?

Alnusjaponica:
@contramundum53 Thank you for the suggestion. I fixed the comments.

not522:
I'm not sure what the intended behavior is. How about testing that some error is raised instead of `SystemError`?

Alnusjaponica:
CatBoostPruningCallback.after_iteration() is called in a [Cython function](https://github.com/catboost/catboost/blob/b8492fa14a13952b8a67ce14b9a09aa2b3a00e07/catboost/python-package/catboost/_catboost.pyx#L1167) and the error occurs here. That might make the resulting error type different as expected.

Alnusjaponica:
@not522 @contramundum53 I added the TODO comments according to the discussion.

not522 (approved):
Thank you for your investigation. LGTM.

Alnusjaponica:
@not522 Thank you for your approval. 
@contramundum53 Could you please review the latest version of this PR again?

PR #4205, opened 2022-11-25
Show best trial number and value in progress bar

### LINKED ISSUE #3661: Add best score reporting into progress bar

nzw0301:
## Motivation

Resolve #3661 and takeover #4093, which introduces a new progress bar stat info: best trial value.

## Description of the changes

When the objective is single, show best trial number and value in tqdm's progress bar.

Difference from the original PR
1. Apply reviewers' suggestions on the original PR
2. Add tests

### DIFF
--- a/optuna/progress_bar.py
+++ b/optuna/progress_bar.py
@@ -89,14 +89,14 @@ class _ProgressBar:
             elapsed_seconds:
                 The time past since :func:`~optuna.study.Study.optimize` started.
             study:
-                The current study object
+                The current study object.
         """
 
-        if not study._is_multi_objective():
-            msg = f"Best trial: {study.best_trial.number}. Best value: {study.best_value}"
-            self._progress_bar.set_description(msg)
-
         if self._is_valid:
+            if not study._is_multi_objective():
+                msg = f"Best trial: {study.best_trial.number}. Best value: {study.best_value}"
+                self._progress_bar.set_description(msg)
+
             if self._n_trials is not None:
                 self._progress_bar.update(1)
                 if self._timeout is not None:
--- a/optuna/progress_bar.py
+++ b/optuna/progress_bar.py
@@ -7,6 +7,7 @@ from tqdm.auto import tqdm
 
 from optuna import logging as optuna_logging
 from optuna._experimental import experimental_func
+from optuna.trial import TrialState
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


nzw0301:
I cannot reproduce the error on my local machine...

contramundum53:

optuna/progress_bar.py
```diff
@@ -75,15 +77,21 @@ def _init_valid(self) -> None:
...
 
         Args:
             elapsed_seconds:
                 The time past since :func:`~optuna.study.Study.optimize` started.
+            study:
+                The current study object.
         """
 
         if self._is_valid:
+            if not study._is_multi_objective() and study.get_trials(states=[TrialState.COMPLETE]):
```
not522, line 91: This line seems to be slowing down the process.

not522:
This PR seems to be several times slower than master.

optuna/progress_bar.py
not522: This PR seems to be several times slower than master.

not522:
I rerun the CI, and it passed.

optuna/progress_bar.py
nzw0301: Thank you for your comments! Indeed, the original code was severely slower than the master branch. I've updated this part.

optuna/progress_bar.py
```diff
@@ -75,15 +76,27 @@ def _init_valid(self) -> None:
...
+            study:
+                The current study object.
         """
 
         if self._is_valid:
+            if not study._is_multi_objective():
+                # Not updating the progress bar when there are no complete trial.
+                try:
+                    msg = f"Best trial: {study.best_trial.number}. \
+                            Best value: {study.best_value:.6g}"
```
not522, line 94: It seems unexpected spaces.

```suggestion
                    msg = (
                        f"Best trial: {study.best_trial.number}. "
                        f"Best value: {study.best_value:.6g}"
                    )
```

not522 (commented):
Thank you for your update. It's almost LGTM. Please check my suggestion.

optuna/progress_bar.py
nzw0301: Thank you for your careful check! I've accepted your suggestion.

not522 (approved):
LGTM!

contramundum53 (approved):
LGTM!

Issue #3964, opened 2022-09-06
Support Python 3.11


not522:
### Motivation

Python 3.11 will be released on 2022-10-24. https://peps.python.org/pep-0664/
This is a tracking issue to support the new Python.

### Description

* Add Python 3.11 to the CI of the unit test and [optuna-examples](https://github.com/optuna/optuna-examples/).
* Use Python 3.11 in the speed benchmark because of its improved speed.

### Additional context (optional)

See also the 3.10 tracking issue. #2981

not522:
Python 3.11 is now supported! (except integration module)

teneon:
Auch, i have upgraded to python3.11 and it looks like Optuna installation using Poetry fails. Could you guys make this a priority, it is very important?

kind regards,
Jim

```
poetry add optuna
Using version ^3.0.3 for optuna

Updating dependencies
Resolving dependencies... (1.5s)

The current project's Python requirement (>=3.11,<4.0) is not compatible with some of the required packages Python requirement:
  - scipy requires Python >=3.8,<3.11, so it will not be satisfied for Python >=3.11,<4.0
  - scipy requires Python >=3.7,<3.10, so it will not be satisfied for Python >=3.11,<4.0
  - scipy requires Python >=3.7,<3.10, so it will not be satisfied for Python >=3.11,<4.0
  - scipy requires Python >=3.7,<3.11, so it will not be satisfied for Python >=3.11,<4.0
  - scipy requires Python >=3.7,<3.11, so it will not be satisfied for Python >=3.11,<4.0
  - scipy requires Python >=3.8,<3.11, so it will not be satisfied for Python >=3.11,<4.0

Because no versions of scipy match >1.7.0,<1.7.1 || >1.7.1,<1.7.2 || >1.7.2,<1.7.3 || >1.7.3,<1.8.0 || >1.8.0,<1.8.1 || >1.8.1,<1.9.0
 and scipy (1.7.0) requires Python >=3.7,<3.10, scipy is forbidden.
And because scipy (1.7.1) requires Python >=3.7,<3.10, scipy is forbidden.
And because scipy (1.7.2) requires Python >=3.7,<3.11
 and scipy (1.7.3) requires Python >=3.7,<3.11, scipy is forbidden.
And because scipy (1.8.0) requires Python >=3.8,<3.11
 and scipy (1.8.1) requires Python >=3.8,<3.11, scipy is forbidden.
Because no versions of optuna match >3.0.3,<4.0.0
[...]

PR #4222, opened 2022-12-02
Support for 3.11

mertbozkir:
## Motivation

I'm Mert from PyData Global 2022 Optuna Sprint, I would love to contribute optuna which is a efficient solution to the Hyperparameter Tuning

## Description of the changes

For the integration testing, I introduced the Python 3.11 version. I'll look at which integrations fail and which succeed, and then figure out how to make Python 3.11 version support a success:)

### DIFF
--- a/.github/workflows/tests-integration.yml
+++ b/.github/workflows/tests-integration.yml
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
 
     steps:
     - name: Checkout
--- a/setup.py
+++ b/setup.py
@@ -111,7 +111,9 @@ def get_extras_require() -> Dict[str, List[str]]:
             # > https://github.com/tensorflow/tensorflow/issues/58032
             "tensorflow; python_version<'3.11'",
             "tensorflow-datasets",
-            "torch==1.11.0",
+            #TODO(mertbozkir): remove the python condition after torch supports 3.11
+            # > https://github.com/pytorch/pytorch/issues/86566
+            "torch==1.11.0; python_version<'3.11'",
             "torchaudio==0.11.0",
             "torchvision==0.12.0",
             "wandb",
--- a/setup.py
+++ b/setup.py
@@ -111,10 +111,12 @@ def get_extras_require() -> Dict[str, List[str]]:
             # > https://github.com/tensorflow/tensorflow/issues/58032
             "tensorflow; python_version<'3.11'",
             "tensorflow-datasets",
-            #TODO(mertbozkir): remove the python condition after torch supports 3.11
+            # TODO(mertbozkir): remove the python condition after torch supports 3.11
             # > https://github.com/pytorch/pytorch/issues/86566
             "torch==1.11.0; python_version<'3.11'",
-            "torchaudio==0.11.0",
+            # TODO (mertbozkir): remove the python condition after torchaudio supports 3.11
+            # > https://github.com/pytorch/audio/issues/2837
+            "torchaudio==0.11.0; python_version<'3.11'",
             "torchvision==0.12.0",
             "wandb",
             "xgboost",
--- a/setup.py
+++ b/setup.py
@@ -89,7 +89,9 @@ def get_extras_require() -> Dict[str, List[str]]:
             "cached-path<=1.1.2",
             "botorch>=0.4.0",
             "catalyst>=21.3",
-            "catboost>=0.26",
+            # TODO(mertbozkir): Remove the python condition after catboost supports python 3.11
+            # > https://github.com/catboost/catboost/issues/2213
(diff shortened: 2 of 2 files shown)


mertbozkir:
This pull request belongs to the issue about Python 3.11 Support:  #3964

hvy:
Seems like we also need to skip `tensorflow` 

```
ERROR: Ignored the following versions that require a different python version: 0.2.0 Requires-Python ==3.6
ERROR: Could not find a version that satisfies the requirement tensorflow; extra == "integration" (from optuna[integration]) (from versions: none)
ERROR: No matching distribution found for tensorflow; extra == "integration"
Error: Process completed with exit code 1.
```

After that, we probably need to skip (`--ignore`) certain tests for Python 3.11. We might be able to do that using `if` with `{{ matrix.python-version }}` in https://github.com/optuna/optuna/blob/1b5a5fb167c6d067034ddc2ff7778473073e892e/.github/workflows/tests-integration.yml

mertbozkir:
@hvy it gave another error from black -reformatting tool- ⚫️

In my opinion, I shoul
1. Git pull all codes 
2. Reformat setup.py with black
3. Then commit and find out if something wrongs. 

What do you think? 🤔

In addition to the black, I put some conditions for other libraries as you can see above: tensorflow, pytorch, torchaudio. I will continue till the integration tests succeed 😅

mertbozkir:
``` Collecting allennlp>=2.2.0
  Downloading allennlp-2.10.0-py3-none-any.whl (729 kB)
  Downloading allennlp-2.9.3-py3-none-any.whl (719 kB)
  Downloading allennlp-2.9.2-py3-none-any.whl (719 kB)
  Downloading allennlp-2.9.1-py3-none-any.whl (718 kB)
  Downloading allennlp-2.9.0-py3-none-any.whl (716 kB)
  Downloading allennlp-2.8.0-py3-none-any.whl (738 kB)
  Downloading allennlp-2.7.0-py3-none-any.whl (738 kB)
  Downloading allennlp-2.6.0-py3-none-any.whl (689 kB)
  Downloading allennlp-2.5.0-py3-none-any.whl (681 kB)
  Downloading allennlp-2.4.0-py3-none-any.whl (625 kB)
  Downloading allennlp-2.3.1-py3-none-any.whl (607 kB)
  Downloading allennlp-2.3.0-py3-none-any.whl (598 kB)
  Downloading allennlp-2.2.0-py3-none-any.whl (595 kB)
INFO: pip is looking at multiple versions of alembic to determine which version is compatible with other requirements. This could take a while.
Collecting alembic>=1.5.0
  Using cached alembic-1.8.1-py3-none-any.whl (209 kB)
INFO: pip is looking at multiple versions of optuna[integration] to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install optuna because these package versions have conflicting dependencies.

The conflict is caused by:
    allennlp 2.10.1 depends on torch<1.13.0 and >=1.10.0
    allennlp 2.10.0 depends on torch<1.12.0 and >=1.10.0
    allennlp 2.9.3 depends on torch<1.12.0 and >=1.6.0
    allennlp 2.9.2 depends on torch<1.12.0 and >=1.6.0
    allennlp 2.9.1 depends on torch<1.11.0 and >=1.6.0
    allennlp 2.9.0 depends on torch<1.11.0 and >=1.6.0
    allennlp 2.8.0 depends on torch<1.11.0 and >=1.6.0
    allennlp 2.7.0 depends on torch<1.10.0 and >=1.6.0
    allennlp 2.6.0 depends on torch<1.10.0 and >=1.6.0
    allennlp 2.5.0 depends on torch<1.9.0 and >=1.6.0
    allennlp 2.4.0 depends on torch<1.9.0 and >=1.6.0
    allennlp 2.3.1 depends on torch<1.9.0 and >=1.6.0
    allennlp 2.3.0 depends on torch<1.9.0 and >=1.6.0
    allennlp 2.2.0 depends on torch<1.9.0 and >=1.6.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```


**Should I put condition to the allennlp as we did for other integration or we should do something else?**

I think the pip cannot solve this dependency conflict, we should put another condition 😅

hvy:
Sorry for the delayed reply, and thanks for the updates!

As for https://github.com/optuna/optuna/pull/4222#issuecomment-1336159657, running black locally would be preferred allowing you to verify changes locally. However, the errors only involve lines being too long so it might not be that difficult to fix by inserting line breaks at seemingly appropriate locations.

As for https://github.com/optuna/optuna/pull/4222#issuecomment-1336161266, you're right we need to skip allennlp as well since it requires torch, which does not yet support 3.11. Could you do that too? 🙇

mertbozkir:
**Yes, of course!** 

I am a little busy till wednesday, then I can work on it and finish this contribution hopefully :)

PR #4225, opened 2022-12-05
Cache all trial objects in Study

amylase:
I was handed over #4141 by @semiexp. 

## Motivation
See #4141 

## Diff from #4141
* Hide `use_cache` option from external users
* Removed AllTrialsCache to simplify the code.

### DIFF
--- a/optuna/integration/_lightgbm_tuner/optimize.py
+++ b/optuna/integration/_lightgbm_tuner/optimize.py
@@ -698,10 +698,9 @@ class _LightGBMBaseTuner(_BaseTuner):
                 self,
                 deepcopy: bool = True,
                 states: Optional[Container[TrialState]] = None,
-                use_cache: bool = False,
             ) -> List[optuna.trial.FrozenTrial]:
 
-                trials = super()._get_trials(deepcopy=deepcopy, states=states, use_cache=use_cache)
+                trials = super()._get_trials(deepcopy=deepcopy, states=states)
                 return [t for t in trials if t.system_attrs.get(_STEP_NAME_KEY) == self._step_name]
 
             @property
--- a/optuna/integration/_lightgbm_tuner/optimize.py
+++ b/optuna/integration/_lightgbm_tuner/optimize.py
@@ -701,7 +701,7 @@ class _LightGBMBaseTuner(_BaseTuner):
                 use_cache: bool = False,
             ) -> List[optuna.trial.FrozenTrial]:
 
-                trials = super().get_trials(deepcopy=deepcopy, states=states, use_cache=use_cache)
+                trials = super()._get_trials(deepcopy=deepcopy, states=states, use_cache=use_cache)
                 return [t for t in trials if t.system_attrs.get(_STEP_NAME_KEY) == self._step_name]
 
             @property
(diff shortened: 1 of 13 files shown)


optuna/study/study.py
```diff
@@ -54,7 +54,12 @@
 
 
 class _ThreadLocalStudyAttribute(threading.local):
-    in_optimize_loop: bool = False
```
c-bata, line 57: Is there something disadvantage to use the default value here? It looks like a common pattern as used in the following pep.
https://peps.python.org/pep-0567/#converting-code-that-uses-threading-local

optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -698,9 +698,10 @@ def get_trials(
                 self,
                 deepcopy: bool = True,
                 states: Optional[Container[TrialState]] = None,
+                use_cache: bool = False,
```
c-bata, line 701: As you commented https://github.com/optuna/optuna/pull/4141#discussion_r1021103628, `get_trials` is a public method, so might be good to keep the API compatibility.

optuna/pruners/_hyperband.py
```diff
@@ -297,8 +297,9 @@ def get_trials(
                 self,
                 deepcopy: bool = True,
                 states: Optional[Container[TrialState]] = None,
+                use_cache: bool = False,
             ) -> List["optuna.trial.FrozenTrial"]:
-                trials = super().get_trials(deepcopy=deepcopy, states=states)
+                trials = super()._get_trials(deepcopy=deepcopy, states=states, use_cache=use_cache)
```
c-bata, line 302: ditto

c-bata (changes requested):
Thank you for your pull request! Let me leave some early comments since I noticed while quickly taking a look.

optuna/study/study.py
amylase: not really. will use default value implementation here.

optuna/integration/_lightgbm_tuner/optimize.py
amylase: good catch!

c-bata:

optuna/study/study.py
```diff
@@ -263,6 +264,31 @@ def objective(trial):
...
+                filtered_trials = trials
+            return copy.deepcopy(filtered_trials) if deepcopy else filtered_trials
+
+        return self._fetch_trials(deepcopy=deepcopy, states=states)
+
+    def _fetch_trials(
+        self,
+        deepcopy: bool = True,
+        states: Optional[Container[TrialState]] = None,
+    ) -> List[FrozenTrial]:
```
c-bata, line 291: Although this is a private method, it's a bit confusing because `_get_trials` and `_fetch_trials` looks quite similar in terms of the interface and the method names. How about removing this method and inlining into `_get_trials`? It doesn't increase the code size.

c-bata (commented):
Looks almost good to me. I have one suggestions though.

optuna/study/study.py
amylase: makes sense. updated the code.

c-bata (approved):
Thanks for the update! LGTM.

not522:
Sorry for my late response.
I discussed this feature with @c-bata, and we concluded that performance degradation could occur. For example, when we run parallel optimization using `TPESampler` with `constant_liar`, the option does not work as expected. https://colab.research.google.com/drive/1UDovCcqIGnGGBB3eu0a0XK0Q4gTQW49j
Let me close this PR. I think there is another way to use the trials cache effectively. I want to continue the discussion with other issues or PRs.

Issue #4029, opened 2022-09-29
Move `_solve_hssp` from `TPESampler` to `optuna/_hypervolume/`


not522:
### Motivation

`TPESampler` includes the solver of the hypervolume subset selection problem. Moving it to `optuna/_hypervolume/` is preferred because it is the general algorithm for multi-objective optimization.

https://github.com/optuna/optuna/blob/e8abeec7d58b64b207bd7606056311d20de0241f/optuna/samplers/_tpe/sampler.py#L791-L830

### Description

- Move `_solve_hssp` from `optuna/samplers/_tpe/sampler.py` to `optuna/_hypervolume/`
- Move tests from `tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py` to `tests/hypervolume_tests/`

RohanHBTU:
Respected Sir/Ma'am,

I am Rohan Kumar Singh. I have just entered my second year at HBTU. I am somewhat familiar to open source contributions and I am well aware of Python, SQL, MongoDB, ML algorithms (still learning Deep learning). I would love to contribute to your organisation Optuna but could you please tell me how to get started?
Hoping to hear from you soon.

Regards 
Rohan

nzw0301:
@RohanHBTU Thank you for sharing your interest in Optuna. Please read https://github.com/optuna/optuna/blob/master/CONTRIBUTING.md before sending a PR.

IMvision12:
Hello, if no one is working on it, may I take this issue?

nzw0301:
@RohanHBTU Are you working on the issue? If not, can @IMvision12 send a PR to resolve this issue?

IMvision12:
Do we need to create new file in _hypervolume and copy `_solve_hssp` in it or we need to move `sampler.py` to _hypervolume, also for `test_multi_objective_sampler.py`?

RohanHBTU:
> @RohanHBTU Are you working on the issue? If not, can @IMvision12 send a PR to resolve this issue?
> 

Yeah sure, @IMvision12  can send a PR.

halucinor:
if no one working on it, can i resolve this issue?

jpbianchi:
OK let me have a shot at this one. (Pydata code sprint)

PR #4223, opened 2022-12-02
WIP moved _solve_hssp to _hypervolume/utils.py - tests not done

jpbianchi:
I moved _solve_hssp to _hypervolume/utils.py and fixed all the broken links

PR #4227, opened 2022-12-07
WIP moved _solve_hssp to _hypervolume/utils.py

jpbianchi:
I created a fresh pull request from a branch (instead of master previously)
This PR is related to https://github.com/optuna/optuna/issues/4029.


jpbianchi:
@hvy Can you take a look at this? It's hard for me without being able to test the code locally.

hvy (changes requested):
Thanks for the updates, looks good. So you fixed the environment.

At the bottom of the PR page, you can see the results of the CI jobs. There are two failing.
One is the style check, https://github.com/optuna/optuna/actions/runs/3638489837/jobs/6140750112. Could you follow and push the suggestions presented in the log there. That should fix this job. The second failing jobs is https://github.com/optuna/optuna/actions/runs/3638489830/jobs/6140748923, which is unlikely to be related to your changes, please ignore this one for now.

hvy:
Just so that we don't forget, https://github.com/optuna/optuna/pull/4223#discussion_r1039331484 would be great to address not necessarily as a part of this PR but in a follow up.

hvy (approved):
Thanks for the swift fixes. LGTM!

Did you also not mind creating a separate file for `_solve_hssp` , e.g. `optuna/_hypervolume/hssp.py` ? That'd be great. I'm fine with either doing that in this PR, or in a separate one.

tests/hypervolume_tests/test_utils.py
```diff
@@ -23,3 +28,55 @@ def test_compute_2d() -> None:
...
+    test_case = np.vstack([test_case, [-float("inf") for _ in range(2)]])
+    truth, approx = _compute_hssp_truth_and_approx(test_case, subset_size)
+    assert np.isinf(truth)
+    assert np.isinf(approx)
+
+    test_case = np.asarray([[random.random() for _ in range(3)] for _ in range(8)])
+    test_case = np.vstack([test_case, [-float("inf") for _ in range(3)]])
+    truth, approx = _compute_hssp_truth_and_approx(test_case, subset_size)
+    assert np.isinf(truth)
+    assert np.isinf(approx)
```
hvy, line 82: Probably just forgot but you can remove the additions in this file now that you've added them in `test_hssp.py`.

jpbianchi: Oops! Fixed! 
Btw, I saw this TODO comment in utils.py
>def _get_observation_pairs(
    study: Study,
    param_names: List[str],
    constant_liar: bool = False,  # TODO(hvy): Remove default value and fix unit tests.

hvy (approved):
Thanks for the additional changes, still LGTM!

tests/hypervolume_tests/test_utils.py
hvy: You mean in `optuna/samplers/_tpe/sampler.py` ?
If I remember correctly, I wanted to remove the default argument since it's a parameter of a private function, and it made little good (to developers, and none to users). Might actually not be that relevant now that it has an additional parameter with a default value.

> def _get_observation_pairs(
    study: Study,
    param_names: List[str],
    constant_liar: bool = False,  # TODO(hvy): Remove default value and fix unit tests.
    constraints_enabled: bool = False,

https://github.com/optuna/optuna/blob/master/optuna/samplers/_tpe/sampler.py#L586-L590

You don't happen to be looking for another issue to work on? 🙂

not522 (approved):
Thank you for your PR. LGTM.

PR #4229, opened 2022-12-08
suggest_categorical: Add overloads for type narrowing

contramundum53:
Taking over #4143, since we'd like to merge this PR with current master and add a test for minimum version of `typing_extensions`.

We also added pull-request trigger in `test-with-minimum-versions.yaml` to run the test.


contramundum53:
@toshihikoyanase Could you review this PR or assign someone else?

toshihikoyanase:
Since we reopen #4143, I guess we can close this PR. What do you think?

PR #4231, opened 2022-12-09
[WIP] Deprecate study system attrs

cross32768:
## Motivation

By the same motivation with https://github.com/optuna/optuna/pull/4188 , I deprecate `system_attrs` property especially in `Study` class.


cross32768:
I re-create this PR since to start over is maybe faster.

PR #262, opened 2018-12-08
Make TYPE_CHECKING work before Python 3.5.2

okapies:
`typing.TYPE_CHECKING` is introduced in Python 3.5.2 and it doesn't work before. You can check like the following:

```python
$ pyenv install 3.5.1
$ pyenv local 3.5.1
$ python
Python 3.5.1 (default, Nov 15 2018, 17:50:51) 
[GCC 5.5.0 20171010] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from typing import TYPE_CHECKING
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'TYPE_CHECKING'
```

You need to handle `ImportError` to make it work:

```python
try:
    from typing import TYPE_CHECKING  # NOQA
except ImportError:
    # typing.TYPE_CHECKING doesn't exist before Python 3.5.2
    TYPE_CHECKING = False

if TYPE_CHECKING:
    from typing import Sequence

def double(a):
    # type: (Sequence[int]) -> Sequence[int]
    return [i * 2 for i in a]

print("TYPE_CHECKING: " + str(TYPE_CHECKING))
double("foo")  # type mismatch
```

The important thing is that mypy *does* work on the above snippet because it always defines `TYPE_CHECKING` as `True` even if `typing` module doesn't support it.

```bash
$ pyenv local 3.5.1
$ python import.py
TYPE_CHECKING: False
$ mypy import.py 
import.py:15: error: Argument 1 to "double" has incompatible type "str"; expected "Sequence[int]"
```

PR #3235, opened 2022-01-17
Remove `type_checking.py`

nzw0301:
## Motivation

In my understanding, `type_checking.py` is for python 3.5, but Optuna dropped Python 3.5 at the end of 2020. Moreover, optuna will drop 3.6 in the near future https://github.com/optuna/optuna/issues/3021. So I suppose this file is not necessary anymore.

## Description of the changes

Remove `optuna.type_checking`

---

Thank @himkt for discussing it.

PR #4238, opened 2022-12-12
Remove `optuna.TYPE_CHECKING`

not522:
## Motivation

`optuna.TYPE_CHECKING` was introduced in #262 for Python 3.5.1 or older versions.
We already dropped support for these versions, and `optuna.TYPE_CHECKING` is not used anywhere.

## Description of the changes
- Remove `optuna.TYPE_CHECKING`

### DIFF
--- a/optuna/__init__.py
+++ b/optuna/__init__.py
@@ -1,5 +1,3 @@
-from typing import TYPE_CHECKING
-
 from optuna import distributions
 from optuna import exceptions
 from optuna import integration
@@ -26,7 +24,6 @@ from optuna.version import __version__
 
 __all__ = [
     "Study",
-    "TYPE_CHECKING",
     "Trial",
     "TrialPruned",
     "__version__",


c-bata (approved):
Good catch 👍

c-bata:

toshihikoyanase (approved):
LGTM. Thank you!

[Notes]
- This seems to be a follow-up for #3235. We decided to use `typing.TYPE_CHECKING` directly in Optuna v3.
  - > It can be labeled as compat breaking, but it's not documented in the API references so the users are hopefully limited. Breaking this in v3 seems reasonable. 
- I couldn't find any users of `oputna.TYPE_CHECKING` in GitHub when I searched [`from optuna import TYPE_CHECKING`](https://github.com/search?q=%22from+optuna+import+TYPE_CHECKING%22&type=code) and [`optuna.TYPE_CHECKING`](https://github.com/search?l=Python&p=1&q=%22optuna.TYPE_CHECKING%22&type=Code).

PR #4240, opened 2022-12-12
Add an in-memory cache in `Trial` class

eukaryo:
## Motivation

make Optuna storage faster.

cf. [https://gist.github.com/c-bata/16d654471393f59e064482ffb8b52bf8](https://gist.github.com/c-bata/16d654471393f59e064482ffb8b52bf8)

> Optuna uses `CachedStorage`, a wrapper class of `BaseStorage` interface, since the API calls of `BaseStorage` interface tends to be expensive. However, the implementation of `CachedStorage` is too complex and it's actually not a wrapper of `BaseStorage` since we've introduced some private storage APIs for `CachedStorage` like [storage._check_and_set_param_distribution()](https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L238-L240) and [storage._get_trials()](https://github.com/optuna/optuna/blob/master/optuna/storages/_cached_storage.py#L387-L389).
So I implemented a prototype of a new simple caching mechanism to remove `CachedStorage` from Optuna. The change is only about 50 lines, but it is more efficient in many situations than CachedStorage.

## Description of the changes

cf. [https://gist.github.com/c-bata/16d654471393f59e064482ffb8b52bf8](https://gist.github.com/c-bata/16d654471393f59e064482ffb8b52bf8)

PR #4242, opened 2022-12-14
 Deprecate `set_system_attr` in Trial

eukaryo:
## Motivation

cf. PR https://github.com/optuna/optuna/pull/4188 and PR https://github.com/optuna/optuna/pull/4240 .

PR#4240 introduced `_get_latest_trial` function. It can be removed after `set_system_attr` is deprecated and removed from the internal codes; the removal of `_get_latest_trial` will increase speed. PR#4188 will remove them from Trial and Study. The present PR will preempt PR#4188 only for Trial, and remove `_get_latest_trial` function.

## Description of the changes

`set_system_attr`s in trial are removed.

`_get_latest_trial` function is removed.

### DIFF
--- a/optuna/integration/_lightgbm_tuner/optimize.py
+++ b/optuna/integration/_lightgbm_tuner/optimize.py
@@ -272,10 +272,14 @@ class _OptunaObjective(_BaseTuner):
             self.pbar.set_description(self.pbar_fmt.format(self.step_name, self.best_score))
             self.pbar.update(1)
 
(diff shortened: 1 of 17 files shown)


c-bata:
@eukaryo Let me close this PR since it is duplicated with #4188.

As the following comment implies, we need to deprecate `system_attrs` property, not the `set_system_attr` method.
https://github.com/optuna/optuna/blob/7cb75a060833e7d934fa7164c834d103288603d5/optuna/trial/_trial.py#L665

PR #4250, opened 2022-12-16
Deprecate `system_attrs` in `Study` class

cross32768:
## Motivation
By the same motivation as https://github.com/optuna/optuna/pull/4188 , I propose to deprecate `system_attrs` property in `Study` class.

## Description of the changes

* Add `deprecatd_func("3.1.0", "6.0.0")` to `system_attrs` property in `optuna.study.Study` class and `optuna.multi_objective.study.MultiObjectiveStudy` class.
* Following the above deprecation, I replace reference of `system_attrs` used from other file by `study._storage.get_study_system_attrs(study._study_id)`.
* I also deprecate `system_attrs` property in `optuna.study.StudySummary` class.

NOTE:  I left `system_attrs` property in `FrozenStudy` because `FrozenStudy` is internal use only.

### DIFF
--- a/optuna/multi_objective/study.py
+++ b/optuna/multi_objective/study.py
@@ -358,7 +358,6 @@ class MultiObjectiveStudy:
         return self._study.user_attrs
 
     @property
-    @deprecated_func("3.1.0", "6.0.0")
     def system_attrs(self) -> Dict[str, Any]:
         """Return system attributes.
 
--- a/optuna/multi_objective/study.py
+++ b/optuna/multi_objective/study.py
@@ -227,7 +227,9 @@ class MultiObjectiveStudy:
         self._study = study
 
         self._directions = []
-        for d in study.system_attrs["multi_objective:study:directions"]:
+        for d in study._storage.get_study_system_attrs(study._study_id)[
+            "multi_objective:study:directions"
+        ]:
             if d == "minimize":
                 self._directions.append(StudyDirection.MINIMIZE)
             elif d == "maximize":
@@ -356,6 +358,7 @@ class MultiObjectiveStudy:
         return self._study.user_attrs
 
     @property
+    @deprecated_func("3.1.0", "6.0.0")
     def system_attrs(self) -> Dict[str, Any]:
         """Return system attributes.
 
@@ -363,7 +366,7 @@ class MultiObjectiveStudy:
             A dictionary containing all system attributes.
         """
 
-        return self._study.system_attrs
+        return self._study._storage.get_study_system_attrs(self._study._study_id)
 
     def set_user_attr(self, key: str, value: Any) -> None:
         """Set a user attribute to the study.
--- a/optuna/study/_study_summary.py
+++ b/optuna/study/_study_summary.py
@@ -124,7 +124,6 @@ class StudySummary:
             "but this schedule is subject to change. "
             "See https://github.com/optuna/optuna/releases/tag/v3.1.0.",
             FutureWarning,
-            stacklevel=2,
         )
 
         return self._system_attrs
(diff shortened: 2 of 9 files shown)


optuna/multi_objective/study.py
```diff
@@ -356,14 +358,15 @@ def user_attrs(self) -> Dict[str, Any]:
         return self._study.user_attrs
 
     @property
+    @deprecated_func("3.1.0", "6.0.0")
```
c-bata, line 361: MultiObjectiveStudy` itself is deprecated and will be removed in v4.0.0 release. How about removing `deprecated_func` here?

cross32768: Oh, I forgot it. I removed this `deprecated_func`.

cross32768:
I cannot use `deprecated_func` decorator for `system_attrs` property of `StudySummary` class because it causes doctest error as below.

`/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/optuna/study/_study_summary.py:docstring of optuna.study.StudySummary.system_attrs:1: WARNING: duplicate object description of optuna.study.StudySummary.system_attrs, other instance in reference/generated/optuna.study.StudySummary, use :noindex: for one of them`

I think this error is caused from redundant document in docstring of `StudySummary` class and `system_attrs` property. Therefore, I added warning for deprecation to docstring of `StudySummary` class manually.

cross32768:
I update `StudySummary` because some manual code is required to deprecate `system_attrs` property in this class. Could you take another look? @c-bata

tests/samplers_tests/test_nsgaii.py
```diff
@@ -504,13 +504,14 @@ def test_crowding_distance_sort(values: List[List[float]]) -> None:
 def test_study_system_attr_for_population_cache() -> None:
     sampler = NSGAIISampler(population_size=10)
     study = optuna.create_study(directions=["minimize"], sampler=sampler)
+    study_system_attrs = study._storage.get_study_system_attrs(study._study_id)
```
c-bata, line 507: Currently, this test case depends on the implementation detail of `InMemoryStorage` (e.g. This test will be broken when using `RDBStorage`). How about moving this line into `get_cached_entries` function?

tests/multi_objective_tests/samplers_tests/test_nsga2.py
```diff
@@ -167,9 +167,10 @@ def test_study_system_attr_for_population_cache() -> None:
     def get_cached_entries(
         study: multi_objective.study.MultiObjectiveStudy,
     ) -> List[Tuple[int, List[int]]]:
+        study_system_attrs = study._storage.get_study_system_attrs(study._study_id)
```
c-bata, line 170: ditto.

optuna/study/_study_summary.py
```diff
@@ -108,3 +115,16 @@ def direction(self) -> StudyDirection:
...
+
+    @property
+    def system_attrs(self) -> Dict[str, Any]:
+        warnings.warn(
+            "`system_attrs` has been deprecated in v3.1.0. "
+            "The removal of this feature is currently scheduled for v6.0.0, "
+            "but this schedule is subject to change. "
+            "See https://github.com/optuna/optuna/releases/tag/v3.1.0.",
+            FutureWarning,
+            stacklevel=2,
```
c-bata, line 127:
```suggestion
```

You don't need to set `stacklevel` here.

c-bata (changes requested):
Changes almost looks good to me. I left some suggestions though.

contramundum53 (approved):
LGTM after @c-bata's comments.

cross32768:
Thank you for your reviews. I update my code according to reviews.

c-bata (approved):

PR #4268, opened 2022-12-22
Create a new process group if `group` of `TorchDistributedTrial` is `None`.

toshihikoyanase:
## Motivation
This is a follow-up PR for #4106.

### Always create a new process group

Currently, behavior of `TorchDistributedTrial` with `group=None` is inconsistent depending on a default process group (i.e., process groups for neural network training). The process group is reused if the backend of the default process group is `gloo`, while a new process group is created if the backend is `nccl`.
I think we can separate the communication of optuna and neural network training if we create a new process group for `gloo`.

And also, the current implementation does not take care of `mpi` backend. 
https://pytorch.org/docs/stable/distributed.html

### Create a process group for Optuna locally

The current master creates a process group as a (package) global variable  (i.e., `_g_pg`) silently. Users may have difficulty to release such hidden variables. So, I'd like to change it to a member of the `TorchDistributedTrial` class, so that it can be released when a `TorchDistributedTrial` object is removed.
This change may take some time to recreate gloo process groups trial by trial, but users can reuse the process group by the `group` argument.

## Description of the changes

- Create a new process group if `group=None`
- Create a new process group as a member of `TorchDistributedTrial`
- Update docstring and fix the annotations

### DIFF
--- a/optuna/integration/pytorch_distributed.py
+++ b/optuna/integration/pytorch_distributed.py
@@ -35,8 +35,6 @@ _suggest_deprecated_msg = (
     "Use :func:`~optuna.integration.TorchDistributedTrial.suggest_float` instead."
 )
 
-_g_pg: Optional["ProcessGroup"] = None
-
 
 def broadcast_properties(f: "Callable[_P, _T]") -> "Callable[_P, _T]":
     """Method decorator to fetch updated trial properties from rank 0 after ``f`` is run.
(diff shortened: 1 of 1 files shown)


contramundum53:

HideakiImamura (approved):
LGTM.

not522:
I want to merge #4301 first because the PyTorch distributed integration is not tested.

toshihikoyanase:
@not522 https://github.com/optuna/optuna/pull/4301 was merged today. So, could you resume your review, please?

not522:
Could you merge the master branch? In my local environment, `test_pytorch_distributed.py` failed.

toshihikoyanase:
I rebased `master`, and the `tests-mpi` jobs succeeded. Could you share the error message, please?

BTW, I'm concerned about the removal of process groups. In the DDP tutorial, the process group was removed explicitly using `dist.destroy_process_group`. 
https://pytorch.org/tutorials/intermediate/ddp_tutorial.html#basic-use-case
This PR will create process groups trial by trial, and the process groups of previous trials may be harmful if they are not released automatically.

c.f., https://github.com/pytorch/pytorch/issues/48203

toshihikoyanase:
> In the DDP tutorial, the process group was removed explicitly using dist.destroy_process_group.

If I add the following code to `TorchDIstributedTrial`, I guess we can remove unnecessary process groups when distributed trials are destroyed. What do you think of it?

```python
    def __del__(self):
        if self._group:
            dist.destroy_process_group(self._group)
```

HideakiImamura:

c-bata:
I'm really sorry for the delay. I'll review this PR today :bow:

toshihikoyanase:
I think we need to remove the process groups properly, so let me mark this PR as draft.

PR #4274, opened 2022-12-22
Add type annotation to `dask.py`

Alnusjaponica:
## Motivation
Solve scheduled Checks (Integration) failure

## Description of the changes

Currently, scheduled Checks (Integration) fails die to mypy check with --warn-unused-ignores option. This PR removes a part of the cause by adding proper `type: ignore`s to dask.py

### DIFF
--- a/optuna/integration/dask.py
+++ b/optuna/integration/dask.py
@@ -29,7 +29,7 @@ with try_import() as _imports:
     import distributed
     from distributed.protocol.pickle import dumps
     from distributed.protocol.pickle import loads
-    from distributed.utils import thread_state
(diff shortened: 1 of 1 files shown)


optuna/integration/dask.py
```diff
@@ -29,7 +29,7 @@
     import distributed
     from distributed.protocol.pickle import dumps
     from distributed.protocol.pickle import loads
-    from distributed.utils import thread_state
+    from distributed.utils import thread_state  # type: ignore[attr-defined]
```
gen740, line 32: I guess this should be `from dask.context import thread_state`.

PR #4279, opened 2022-12-23
Fix mypy error abut `PyTorch Lightning` in Checks (integration)

Alnusjaponica:
## Motivation
Solve scheduled Checks (Integration) failure

## Description of the changes
Currently, scheduled Checks (Integration) fails die to mypy check with `--warn-unused-ignores` option. This PR removes a part of the cause related to `PyTorch Lightning`.
Although test for  Checks (Integration) does't run when PR in opened, I confirmed this PR together with other "Fix mypy error" series reduces mypy error at [CI on my fork](https://github.com/Alnusjaponica/optuna/pull/3).

### DIFF
--- a/tests/integration_tests/test_pytorch_lightning.py
+++ b/tests/integration_tests/test_pytorch_lightning.py
@@ -22,7 +22,7 @@ with try_import() as _imports:
     import torch.nn.functional as F
 
 if not _imports.is_successful():
-    LightningModule = object  # type: ignore # NOQA
+    LightningModule = object  # type: ignore[assignment, misc]  # NOQA
 
 pytestmark = pytest.mark.integration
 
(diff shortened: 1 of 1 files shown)


HideakiImamura:
@gen740 @toshihikoyanase Could you review this PR?

gen740 (approved):
LGTM!

Alnusjaponica:
Thank you for your approval. @gen740 
Could you assign another reviewer to this PR? @toshihikoyanase

tests/integration_tests/test_pytorch_lightning.py
```diff
@@ -111,6 +110,7 @@ def validation_step(  # type: ignore
             accuracy = torch.tensor(0.6)
 
         self.log("accuracy", accuracy, sync_dist=True)
+        return {"validation_accuracy": accuracy}
```
toshihikoyanase, line 113: [Note] The return value of `validation_step` can be None, but the current change seems reasonable. `Dict[str, Any]` is also expected as can be seen in the [`STEP_OUTPUT` definition](https://github.com/Lightning-AI/lightning/blob/a008801e25a6e93214b410b68ca1edf2e3ddae14/src/pytorch_lightning/utilities/types.py#L35).

https://github.com/Lightning-AI/lightning/blob/a008801e25a6e93214b410b68ca1edf2e3ddae14/src/pytorch_lightning/core/module.py#L764

toshihikoyanase (approved):
I'm sorry for the delayed response. I checked the change in my local PC, and the errors were resolved.

PR #4318, opened 2023-01-11
Exclude benchmark directories from the sdist package

c-bata:
## Motivation

@himkt found that the sdist packages from Optuna v3.0.0 unintentionally contain `benchmarks` directory.
https://pypi.org/project/optuna/#files

The cause of the problem is that some Python packages were created in `benchmarks` directory in v3.0.0.

* https://github.com/optuna/optuna/tree/v2.10.1/benchmarks
* https://github.com/optuna/optuna/tree/v3.0.0/benchmarks

## Description of the changes

Like `tests` packages, I added `benchmarks.*` to the exclude argument of `find_packages` for now. Perhaps, adding `recursive-exclude benchmarks` in `MANIFEST.in` would be safer.
https://packaging.python.org/en/latest/guides/using-manifest-in/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -146,7 +146,7 @@ setup(
         "Documentation": "https://optuna.readthedocs.io",
         "Bug Tracker": "https://github.com/optuna/optuna/issues",
     },
-    packages=find_packages(exclude=("tests", "tests.*", "benchmarks")),
+    packages=find_packages(exclude=("tests", "tests.*", "benchmarks", "benchmarks.*")),
     package_data={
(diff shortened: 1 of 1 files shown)


c-bata:
I confirmed that this change works as expected.

Screenshot 2023-01-11 11 51 20

himkt (approved):
LGTM, thank you for the quick patch!

c-bata:

not522 (approved):
LGTM

Issue #4248, opened 2022-12-16
Avoid to use `verbose_eval` argument in LightGBM tuner.


c-bata:
### Motivation

`verbose_eval` argument is deprecated in LightGBM.

<img width="1549" alt="Screenshot 2022-12-16 9 57 44" src="https://user-images.githubusercontent.com/5564044/207998007-b7aa8898-7abe-4a72-87c5-a4ea67d16734.png">

https://optuna.readthedocs.io/en/stable/tutorial/10_key_features/005_visualization.html#sphx-glr-tutorial-10-key-features-005-visualization-py

### Suggestion

> 'verbose_eval' argument is deprecated and will be removed in a future release of LightGBM. Pass 'log_evaluation()' callback via 'callbacks' argument instead.

As the above deprecation message shows, we need to use `log_evaluation()` callback instead.

eromoe:
+1, same problem

nzw0301:
@eromoe Thank you for sharing the link! To be clear, the issue described in https://github.com/microsoft/LightGBM/issues/5241 has been resolved; you can disable the evaluation log correctly. The corresponding PR to this issue would affect only docstring.

PR #4335, opened 2023-01-14
Remove `verbose_eval` argument from lightgbm callback in tutorial pages

nzw0301:
## Motivation

Resolve https://github.com/optuna/optuna/issues/4248

## Description of the changes

Apply similar changes to https://github.com/optuna/optuna/pull/3086. 

---

One of the problem pages would be 

Screenshot 2023-01-14 at 22 14 55

https://optuna--4335.org.readthedocs.build/en/4335/tutorial/10_key_features/005_visualization.html#sphx-glr-tutorial-10-key-features-005-visualization-py

### DIFF
--- a/tutorial/10_key_features/005_visualization.py
+++ b/tutorial/10_key_features/005_visualization.py
@@ -77,7 +77,7 @@ def objective(trial):
     # Add a callback for pruning.
     pruning_callback = optuna.integration.LightGBMPruningCallback(trial, "auc")
     gbm = lgb.train(
-        param, dtrain, valid_sets=[dvalid], verbose_eval=False, callbacks=[pruning_callback]
+        param, dtrain, valid_sets=[dvalid], callbacks=[pruning_callback, lgb.log_evaluation(0)]
     )
 
     preds = gbm.predict(valid_x)
--- a/tutorial/10_key_features/005_visualization.py
+++ b/tutorial/10_key_features/005_visualization.py
@@ -77,7 +77,7 @@ def objective(trial):
     # Add a callback for pruning.
     pruning_callback = optuna.integration.LightGBMPruningCallback(trial, "auc")
     gbm = lgb.train(
-        param, dtrain, valid_sets=[dvalid], callbacks=[pruning_callback, lgb.log_evaluation(0)]
+        param, dtrain, valid_sets=[dvalid], callbacks=[pruning_callback]
     )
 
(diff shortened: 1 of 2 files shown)


c-bata:

tutorial/10_key_features/005_visualization.py
```diff
@@ -77,7 +77,7 @@ def objective(trial):
     # Add a callback for pruning.
     pruning_callback = optuna.integration.LightGBMPruningCallback(trial, "auc")
     gbm = lgb.train(
-        param, dtrain, valid_sets=[dvalid], verbose_eval=False, callbacks=[pruning_callback]
+        param, dtrain, valid_sets=[dvalid], callbacks=[pruning_callback, lgb.log_evaluation(0)]
```
contramundum53, line 80: @nzw0301 Could you explain what will happen if you pass 0 to `period` argument of `log_evaluation` function? I couldn't find that in [the documentation](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.log_evaluation.html)🙇‍♂️

contramundum53:
Otherwise, LGTM.

tutorial/10_key_features/005_visualization.py
nzw0301: if `period` is non-positive, `log_evaluation` disable the logging because the first condition is false in 
https://github.com/microsoft/LightGBM/blob/5df7d516f3e64e8431aecfea892a850afeb787bb/python-package/lightgbm/callback.py#L75

contramundum53: Hmm, if we do not want logging, why do we register this callback?

nzw0301: Ah, good point. I'll update the PR by removing it. Should I apply a similar change to the test code too? We might be able to remove `log_evaluation(-1)` in https://github.com/optuna/optuna/pull/3086/files.

contramundum53: 👍 
Thank you!

nzw0301: I tried LightGBMTuner without `log_evaluation`, but it didn't work like this example. So I don't update the this PR to remove `log_evaluation` for LightGBMTuner tests.

c-bata (approved):
I confirmed that #4248 is resolved.

Screenshot 2023-01-16 13 51 55

Issue #3521, opened 2022-04-27
Unify the arguments of `OptunaSearchCV` and scikit learn's `GridSearchCV`


nzw0301:
### Motivation

In my understanding, `OptunaSearchCV` is inspired by `GridSearchCV`'s interface to replace grid search in scikit-learn with Optuna's parameter search.

I realised that the current arguments of `OptunaSearchCV` were not the same as `GridSearchCV`:

- the order of arguments differ
- `GridSearchCV` uses keyword-only arguments, but `OptunaSearchCV` does not.

### Description

Update the arguments of `OptunaSearchCV` to follow `GridSearchCV` for consistency.

nzw0301:
#4336 partially addresses this issue except for argument ordering inconsistency. 

In addition, `scoring` and `refit` require a lot of work, according to [the previous attempt](https://github.com/optuna/optuna/pull/2944).

nzw0301:
My attempt for `scoring` and `refit` is https://github.com/nzw0301/optuna/tree/prototypo-scoring. Still WIP, but it might not be too hard as I expected.

PR #4336, opened 2023-01-14
Partially follow `sklearn.model_selection.GridSearchCV`'s arguments

nzw0301:
## Motivation

A part of https://github.com/optuna/optuna/issues/3521

## Description of the changes

1. `n_jobs`'s default value has been `None` since scikit-learn v0.20. This change isn't backwards-breaking change.
2. Follow keyword only argument rather than positional one. This convention has been introduced since scikit-learn 1.0 See https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_0_0.html#keyword-and-positional-arguments for more details.
3. update `cv` type hint, default value, and docstring. This change also isn't backwards-breaking change

### DIFF
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -376,8 +376,9 @@ class OptunaSearchCV(BaseEstimator):
         cv:
             Cross-validation strategy. Possible inputs for cv are:
 
+            - :obj:`None`, to use the default 5-fold cross validation,
             - integer to specify the number of folds in a CV splitter,
-            - a CV splitter,
+            - `CV splitter <https://scikit-learn.org/stable/glossary.html#term-CV-splitter>`_,
             - an iterable yielding (train, validation) splits as arrays of indices.
 
             For integer, if ``estimator`` is a classifier and ``y`` is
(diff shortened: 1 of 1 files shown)


c-bata:
@contramundum53 @gen740 Could you review this PR if you have time?

optuna/integration/sklearn.py
```diff
@@ -730,7 +732,7 @@ def __init__(
         self.estimator = estimator
         self.max_iter = max_iter
         self.n_trials = n_trials
-        self.n_jobs = n_jobs
+        self.n_jobs = n_jobs if n_jobs else 1
```
contramundum53, line 735: image
Maybe we should also be consistent when used in a [joblib.parallel_backend](https://joblib.readthedocs.io/en/latest/parallel.html#joblib.parallel_backend) context?

contramundum53 (commented):
Otherwise, LGTM.

optuna/integration/sklearn.py
nzw0301: Yeah, this part is confusing. In `OptunaSearchCV`, `n_jobs` is for `study.optimize` so this argument is independent of `joblib.parallel_backend`.

contramundum53: I see, thanks for explanation!

contramundum53 (approved):
LGTM.

gen740 (approved):
LGTM!

PR #4337, opened 2023-01-15
Add test to check `error_score` is stored

nzw0301:
## Motivation

The current test of OptunaSearchCV doesn't check `error_score` is used when `error_score` is a numerical value.

## Description of the changes

Add assertions to check this. By using `error_score`, `intermediate_values` and several values in `trial.user_attrs` are set `error_score` value

### DIFF
--- a/tests/integration_tests/test_sklearn.py
+++ b/tests/integration_tests/test_sklearn.py
@@ -289,6 +289,14 @@ def test_objective_error_score_nan() -> None:
     ):
         optuna_search.fit(X)
 
+    for trial in optuna_search.study_.get_trials():
+        assert np.all(np.isnan(list(trial.intermediate_values.values())))
+
+        # "_score" stores every score value for train and test validation holds.
+        for name, value in trial.user_attrs.items():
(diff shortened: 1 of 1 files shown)


tests/integration_tests/test_sklearn.py
```diff
@@ -289,6 +289,14 @@ def test_objective_error_score_nan() -> None:
     ):
         optuna_search.fit(X)
 
+    for trial in optuna_search.study_.get_trials():
```
nzw0301, line 292: `optuna_search.trials_` is unavailable because the property requires at least one complete trial.

c-bata:
@toshihikoyanase @HideakiImamura Could you review this PR if you have time?

HideakiImamura (approved):
LGTM.

toshihikoyanase (approved):
Thank you for your PR. Sounds reasonable.
https://github.com/optuna/optuna/blob/c4514c198768f5187c93a7920934fe588dec3860/optuna/integration/sklearn.py#L161-L165

PR #4341, opened 2023-01-16
Use Sphinx 5 until `sphinx_rtd_theme` supports Sphinx 6

not522:
## Motivation
`sphinx_rtd_theme` does not support Sphinx 6 yet.

At least, the following two problems should be fixed before updating to Sphinx 6.

- The upper left logo does not appear.
- The version selector does not work.

## Description of the changes
- Add upper limit of Sphinx

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,8 @@ def get_extras_require() -> Dict[str, List[str]]:
             "plotly>=4.9.0",  # optuna/visualization.
             "scikit-learn",
             "scikit-optimize",
-            "sphinx",
+            # TODO(not522): Remove the constraint after sphinx_rtd_theme supports Sphinx 6
+            "sphinx<6",
             "sphinx-copybutton",
             "sphinx-gallery",
             "sphinx-plotly-directive",


c-bata:

toshihikoyanase:
In addition to GitHub action's config, I think we need to update RTD's config (i.e., [`.readthedocs.yml`](https://github.com/optuna/optuna/blob/c4514c198768f5187c93a7920934fe588dec3860/.readthedocs.yml#L20-L23)). But, I couldn't find the way to restrict individual packages in it. Do you think we should edit `setup.py` to restrict `sphinx` version?
https://docs.readthedocs.io/en/stable/config-file/v2.html#packages

.github/workflows/sphinx-build.yml
```diff
@@ -84,6 +84,8 @@ jobs:
       run: |
         python -m pip install -U pip
         pip install --progress-bar off -U .[document]
+        # TODO(not522): Remove this line after sphinx_rtd_theme supports Sphinx 6
+        pip install -U 'sphinx<6'
```
toshihikoyanase, line 88: Unfortunately, ReadTheDocs does not refer to settings in `.github`. 
We can confirm it by checking the pre-built document:
https://optuna--4341.org.readthedocs.build/en/4341/

c-bata:
As @toshihikoyanase said, it seems that we need to edit `setup.py`.

> If you want to pin Sphinx to a specific version, use a requirements.txt or environment.yml file (see [Requirements file](https://docs.readthedocs.io/en/stable/config-file/v2.html#requirements-file) and [conda.environment](https://docs.readthedocs.io/en/stable/config-file/v2.html#conda-environment)). If you are using a metadata file to describe code dependencies like setup.py, pyproject.toml, or similar, you can use the extra_requirements option (see [Packages](https://docs.readthedocs.io/en/stable/config-file/v2.html#packages)). This also allows you to override [the default pinning done by Read the Docs if your project was created before October 2020](https://docs.readthedocs.io/en/stable/build-default-versions.html#external-dependencies).
> https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx

not522:
I see. I updated it. The document seems to be fixed. https://optuna--4341.org.readthedocs.build/en/4341/

c-bata (approved):

toshihikoyanase (approved):
LGTM! Thank you!
I confirmed that the document was rendered as expected.
image
image

Issue #2021, opened 2020-11-14
Optuna Suggests the Same Parameter Values in a lot of Trials (Duplicate Trials that Waste Time and Budget)


deathes:
Optuna TPESampler and RandomSampler try the same suggested integer values (possible floats and loguniforms as well) for any parameter more than once for some reason. I couldn't find a way to stop it from suggesting same values over over again. Out of 100 trials quite a few of them are just duplicates. Unique suggested value count ends up around 80-90 out of 100 trials. If I include more parameters for tuning, say 3, I even see all 3 of them getting the same values a few times in 100 trials.

It's like this:

[I 2020-11-14 14:44:05,320] Trial 8 finished with value: 45910.54012028659 and parameters: {'min_data_in_leaf': 75}. Best is trial 4 with value: 45805.19030897498.
[I 2020-11-14 14:44:07,876] Trial 9 finished with value: 45910.54012028659 and parameters: {'min_data_in_leaf': 75}. Best is trial 4 with value: 45805.19030897498.
[I 2020-11-14 14:44:10,447] Trial 10 finished with value: 45831.75933279074 and parameters: {'min_data_in_leaf': 43}. Best is trial 4 with value: 45805.19030897498.
[I 2020-11-14 14:44:13,502] Trial 11 finished with value: 46125.39810101329 and parameters: {'min_data_in_leaf': 4}. Best is trial 4 with value: 45805.19030897498.
[I 2020-11-14 14:44:16,547] Trial 12 finished with value: 45910.54012028659 and parameters: {'min_data_in_leaf': 75}. Best is trial 4 with value: 45805.19030897498.

Example code below: 

    def lgb_optuna(trial):

        rmse = []

        params = {
            "seed": 42,
            "objective": "regression",
            "metric": "rmse",
            "verbosity": -1,
            "boosting": "gbdt",
            "num_iterations":  1000,
            'min_data_in_leaf':  trial.suggest_int('min_data_in_leaf', 1, 100)
        }

        cv = StratifiedKFold(n_splits=5, random_state=42, shuffle=False)
        for train_index, test_index in cv.split(tfd_train, tfd_train[:,-1]):
            X_train, X_test = tfd_train[train_index], tfd_train[test_index]
            y_train = X_train[:,-2].copy()
            y_test = X_test[:,-2].copy()
            
            dtrain = lgb.Dataset(X_train[:,:-2], label=y_train)
            dtest = lgb.Dataset(X_test[:,:-2], label=y_test)
        
[...]

PR #4352, opened 2023-01-19
Add `last_value_if_duplicate`

contramundum53:
## Motivation
Provide workaround for #2021.

## Description of the changes
This PR adds a function `Trial.last_value_if_duplicate` that returns the values of the last trial that has exactly the same params as suggested in this trial, if exists.

Example: 
```python3
def objective(trial):

    past_value = trial.last_value_if_duplicate()
    if past_value is not None: 
        return past_value

    ...
```
This utility function makes it easy to add a simple memoization mechanism in the objective function.

## Alternative options
Add a freestanding function `optuna.last_value_if_duplicate(optuna.Trial)` that does the same.

### DIFF
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -5,6 +5,7 @@ from typing import Dict
 from typing import Optional
 from typing import overload
 from typing import Sequence
+from typing import List
 import warnings
 
 import optuna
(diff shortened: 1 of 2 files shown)


contramundum53:
Because this PR is very simple to implement for users, we decided not to introduce this functionality in optuna API, but rather put this in FAQ.

okaikov:
can't find the relevant section in FAQ, can you please provide a link?

contramundum53:
I'm very sorry, I haven't had time to write one. I will write one when I have time.

contramundum53:
A simple example would be the following.
```
    def objective(trial):
        # Sample parameters
        x = trial.suggest_int("x", 0, 5)
        ...
        
        # When you finish sampling, check whether there are trials with same values.
        complete_trials = trial.study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))
        for t in complete_trials[::-1]:
            if trial.params == t.params:
                return trial.values
        
        # Do actual computation
        return func(x)
```

PR #4357, opened 2023-01-21
Remove `cached-path` from `setup.py`

nzw0301:
## Motivation

Allennlp 2.10.1 specifies `cached-path` version: `cached-path>=1.1.3,<1.2.0` in https://github.com/allenai/allennlp/blob/0dc554f975a0130a79a9b43a2cad7a52ed579d7f/requirements.txt#L6. So we remove `cached-path` in `setup.py` by following TODO comment. This version specification also causes installation error in https://github.com/optuna/optuna/actions/runs/3964442009/jobs/6793305565#logs.

## Description of the changes

Remove `cached-path` from `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,8 +85,6 @@ def get_extras_require() -> Dict[str, List[str]]:
         ],
         "integration": [
             "allennlp>=2.2.0; python_version<'3.11'",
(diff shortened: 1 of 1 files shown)


c-bata (approved):
Good catch!

PR #4360, opened 2023-01-23
Minor fix of documents

toshihikoyanase:
## Motivation

During PDF build, some errors occurred as follows:
- Overflow of `study.trials` output
- Unescaped `_` in Hyperband formula.

This PR is related to #4349. Removed changes in CI settings.

## Description of the changes

- Fixed above two errors


c-bata:

gen740 (approved):
LGTM!
But, I have a little question. Why optuna doesn't use a raw string literal as a docstring to avoid `\\` escape sequence?

toshihikoyanase:
Thank you for your review.
> Why optuna doesn't use a raw string literal as a docstring to avoid `\\` escape sequence?

Good catch! I don't think we have a preference to avoid the raw string literal, and we can simplify the docstring, as you mentioned. Maybe we can work on it in a follow-up PR.

not522 (approved):
LGTM

PR #4366, opened 2023-01-27
Cache all trials in Study with TTL

amylase:
This PR is another try on #4141, with TTL.
## Motivation

See #4141 for the cache itself.

#4141 implementation may result in less optimization capability because some independent samplings may not be able to use the latest trials history. 

```python
def objective(trial):
  huge_preprocess()
  trial.suggest()
  mode.fit()
```

In the objective function above, `trial.suggest()` may use the stale cache because there is `huge_preprocess()` between the trial creation, where multivairate sampling occurs, and `trial.suggest()`. The cache is only updated in the multivariate sampling in #4141. 

## Description of the changes

See #4141 for the cache itself.
My change adds the TTL to all trials cahce on `Study`. The cache becomes invalid after a minute.

### DIFF
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -189,6 +189,7 @@ def _run_trial(
         optuna.storages.fail_stale_trials(study)
 
     trial = study.ask()
+    study._thread_local.cache_all_trials.activate(lambda: study.get_trials(deepcopy=False))
 
     state: Optional[TrialState] = None
     value_or_values: Optional[Union[float, Sequence[float]]] = None
@@ -220,6 +221,7 @@ def _run_trial(
         frozen_trial = study._storage.get_trial(trial._trial_id)
         raise
     finally:
+        study._thread_local.cache_all_trials.deactivate()
         if frozen_trial.state == TrialState.COMPLETE:
             study._log_completed_trial(frozen_trial)
         elif frozen_trial.state == TrialState.PRUNED:
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1,4 +1,5 @@
 import copy
+from datetime import datetime, timedelta
 from numbers import Real
 import threading
 from typing import Any
@@ -54,14 +55,18 @@ _logger = logging.get_logger(__name__)
 
 
 class AllTrialsCache:
+    CACHE_TTL = timedelta(minutes=1)
+
     _is_active: bool
     _cache: Optional[List["FrozenTrial"]]
     _getter: Optional[Callable[[], List["FrozenTrial"]]]
+    _last_updated: datetime
 
     def __init__(self) -> None:
         self._is_active = False
         self._cache = None
         self._getter = None
+        self._last_updated = datetime.now()
 
     def is_active(self) -> bool:
         return self._is_active
@@ -77,9 +82,10 @@ class AllTrialsCache:
 
     def get(self) -> List["FrozenTrial"]:
         assert self._is_active
-        if self._cache is None:
+        if self._cache is None or (datetime.now() - self._last_updated > AllTrialsCache.CACHE_TTL):
             assert self._getter is not None
             self._cache = self._getter()
+            self._last_updated = datetime.now()
         return self._cache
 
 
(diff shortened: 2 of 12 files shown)


optuna/study/_optimize.py
```diff
@@ -189,6 +189,7 @@ def _run_trial(
         optuna.storages.fail_stale_trials(study)
 
     trial = study.ask()
+    study._thread_local.cache_all_trials.activate(lambda: study.get_trials(deepcopy=False))
```
hvy, line 192: Is there an intention behind the design of making the argument to `activate` a lambda/configurable?

hvy, line 192: This cache will not do anything when optimizing `Study.ask` and `Study.tell` . I think that was one benefit with the current `_CachedStorage`. Is this something that you've thought about?

hvy, line 192: Just to clarify, is this functionality meant to replace `_CachedStorage`?
With `_CachedStorage` we only update non-cached (non-finished) trials. Here, we do that too if these two features co-exists, otherwise if it's to be replaced, we'll query the entire history in each cache update (every TTL interval).

hvy (commented):
Thanks a lot for the PR!

Left some initial comments. Do you mind taking a look?

HideakiImamura:
@hvy @c-bata Could you review this PR?

optuna/study/study.py
```diff
@@ -227,6 +270,7 @@ def get_trials(
         self,
         deepcopy: bool = True,
         states: Optional[Container[TrialState]] = None,
+        use_cache: bool = False,
```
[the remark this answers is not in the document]
amylase: Good point. We had similar discussion in previous PR (#4141) and introduced a separate private method for the case using cache. will update.

optuna/study/study.py
```diff
@@ -53,8 +54,50 @@
...
+
+    def get(self) -> List["FrozenTrial"]:
+        assert self._is_active
+        if self._cache is None or (
+            datetime.datetime.now() - self._last_updated > AllTrialsCache.CACHE_TTL
+        ):
+            assert self._getter is not None
+            self._cache = self._getter()
+            self._last_updated = datetime.datetime.now()
+        return self._cache
```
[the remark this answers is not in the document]
amylase: as described in the motivation of this PR, the cache may be stale if there is a big gap between the objective function call and `suggest` call that triggers independent sampling. I think this scenario is pretty common in machine learning (the gap is introduced by preprocess/dataset load)

optuna/study/study.py
```diff
@@ -259,10 +303,21 @@ def objective(trial):
...
             states:
                 Trial states to filter on. If :obj:`None`, include all states.
+            use_cache:
+                If True, returns trials from an in-memory cache, which is cached
+                before evaluating the objective function.
 
         Returns:
             A list of :class:`~optuna.trial.FrozenTrial` objects.
         """
+        if use_cache and self._thread_local.cache_all_trials.is_active():
```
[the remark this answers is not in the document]
amylase: as you mentioned, this cache will not be available in ask-and-tell. we don't want to distract ask-and-tell users by intended behavior.

optuna/study/_optimize.py
amylase: not really. I will remove callable argument to simplify the code

amylase: right, this cache will not be available for ask-and-tell users. my (and @c-bata -san)'s plan is to move all cache logic in  `_CachedStorage` to `RDBStorage` because `_CachedStorage` is only used by `RDBStorage`.
Note that this cache is still necessary for users of the other storage such as `JournalRedisStorage`.

amylase: we don't meat to replace `_CachedStorage`. please have a look at this comment for our plan
https://github.com/optuna/optuna/pull/4366#discussion_r1100976416

amylase:
This issue will be resolved in https://github.com/optuna/optuna/pull/4468 with the different approach.

amylase:
Measured the performance of this change under the same settings as [this experiment](https://gist.github.com/c-bata/16d654471393f59e064482ffb8b52bf8) in my computer.

| branch / storage | RDBStorage with CachedStorage | RDBStorage |
|-|-|-|
| master | 68.78 sec | 194.87 sec |
|-|-|-|
| this PR | 50.12 sec (1.37x faster) | 70.96 sec (2.75x faster) |

PR #4376, opened 2023-01-31
Disable colored log when `NO_COLOR` env or not tty.

gen740:
## Motivation

When piping optuna's log output to the file, escape sequences are also captured.

Given the following code.
```
python3 -c "import optuna;study = optuna.create_study()" &> log.txt 
```

log.txt would be like
```
[32m[I 2023-01-31 10:25:25,449][0m A new study created in memory with name: no-name-7c296450-89b2-45d3-9f6c-c363edf688f3[0m
```

Many python tools including `pip` would disable coloring when stdout is not a tty, so it is more natural to do this in optuna.

And also natural to disable coloring when `NO_COLOR` environment variable is set, (ref. https://no-color.org).

## Description of the changes

- Created a function `_color_supported` to check the environment variable and tty state.

### DIFF
--- a/optuna/logging.py
+++ b/optuna/logging.py
@@ -7,7 +7,6 @@ from logging import INFO
 from logging import WARN
 from logging import WARNING
 import os
-import platform
 import sys
 import threading
 from typing import Optional
--- a/optuna/logging.py
+++ b/optuna/logging.py
@@ -6,6 +6,7 @@ from logging import FATAL
 from logging import INFO
 from logging import WARN
 from logging import WARNING
+import sys
 import threading
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


HideakiImamura:
@Alnusjaponica @keisuke-umezawa Could you review this PR?

Alnusjaponica (approved):

Alnusjaponica:
LGTM

gen740:
Do I need to add the test? Because it's hard to test and `test/test_logger.py` doesn't test a private function (that starts with `_`).

contramundum53:
This breaks compatibility with old `colorlog`. #4403

PR #4322, opened 2023-01-12
Fix checks integration about pytorch lightning

Alnusjaponica:
## Motivation
Resolve #3418 and #4116 

### Refactor deprecated features 
- `trainer.training_type_plugin` is deleted since v1.8 ([PR#11239](https://github.com/Lightning-AI/lightning/pull/11239)). The attribute `training_type_plugin` is just renamed to `strategy`, so refactored as suggested.

- An optional argument of `trainer`, `accelerator` stopped to accept `ddp_cpu`. Instead, we can pass `ddp` to `strategy` and `cpu` to `accelerator`. Also, `num_processes` will be removed, so we give the number of processes to `devices` instead.

-  `AcceleratorConnector.distributed_backend` is deleted, but now they have `AcceleratorConnector.is_distributed` instead, so refactored as suggested.

- `callback.on_init_start()` is deleted since v1.8 ([Issue#10894](https://github.com/Lightning-AI/lightning/issues/10894), [PR#10940](https://github.com/Lightning-AI/lightning/pull/10940), [PR#14867](https://github.com/Lightning-AI/lightning/pull/14867)).
Although they don't provide exactly equivalent alternative, it would be possible to move this confirmation to somewhere else. Although it seems [`strategy.setup_environment`](https://github.com/Lightning-AI/lightning/blob/3e7c014845f939c20774486ee3798dbcef4db7f9/src/pytorch_lightning/trainer/trainer.py#L1034) is the right place to implement this check, implementing as a method of `Strategy` will affect user's code. Therefore it will be reasonable to implement as `callback.setup` or `callback.on_fit_start()`.

### Stop supporting DDP temporary
When you use DDP and `optuna.TrialPruned()` is raised from the child process, Pytorch Lightning tries to resynchronize to fix the "error" and finally deal it as a `DeadlockDetectedException`, which terminates the whole process. For more details, see [reconciliate_processes](https://github.com/Lightning-AI/lightning/blob/15ef52bc732d1f907de4de58683f131652c0d68c/src/pytorch_lightning/strategies/ddp.py#L427). To fix this problem, we need to change environment variable and private variable.

It might be possible to solve this problem by manually raising `optuna.TrialPruned()` from objective function as in `CatBoostPruningCallback` (see this [comment](https://github.com/optuna/optuna/pull/4327#issuecomment-1381498269)). If it is possible, I am going to apply the change as another PR.

[...]

PR #4384, opened 2023-02-02
Support DDP in `PyTorch-Lightning`

Alnusjaponica:
## Motivation
Follow up #4322. 
Temporary, DDP is not supported in `PyTorchLightningPruningCallback` because of the problem described in #4322.
This PR make `PyTorchLightningPruningCallback` support DDP again.

## Description of the changes
This PR
- **Require users to call `callback.check_pruned()` in objective functions when they use DDP**
- Activates `test_pytorch_lightning_pruning_callback_ddp_monitor` and `test_pytorch_lightning_pruning_callback_ddp_unsupported_storage`
- Store intermediate values, pruning state and message directly in the storage

### DIFF
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -135,7 +135,7 @@ class PyTorchLightningPruningCallback(Callback):
     def check_pruned(self) -> None:
         """Raise :class:`optuna.TrialPruned` manually if pruned.
 
-        Currently, intermediate_values are not properly propagated between processes due to
+        Currently, ``intermediate_values`` are not properly propagated between processes due to
         _CachedStorage. Therefore, necessary information is kept in trial_system_attrs when the
         trial runs in a distributed situation.
         Please do not call this method when you are not using DDP.
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -43,69 +43,16 @@ class PyTorchLightningPruningCallback(Callback):
             how this dictionary is formatted.
 
 
-    Example:
-
-        .. testcode::
-
-            import optuna
-            from optuna.integration import PyTorchLightningPruningCallback
-            import pytorch_lightning as pl
-
-
-            def objective(trial: optuna.trial.Trial) -> float:
-                callback = PyTorchLightningPruningCallback(trial, monitor="accuracy")
-                trainer = pl.Trainer(max_epochs=10, callbacks=[callback])
-
-                model = MyModel()
-                trainer.fit(model)
-
-                return trainer.callback_metrics["val_acc"].item()
-
-
-            pruner = optuna.pruners.MedianPruner()
-            study = optuna.create_study(direction="maximize", pruner=pruner)
-            study.optimize(objective, n_trials=100, timeout=600)
-
-    Note:
-        If you would like to use PyTorchLightningPruningCallback in a distributed training
-        environment, you need to evoke `PyTorchLightningPruningCallback.check_pruned()`
-        manually so that :class:`~optuna.exceptions.TrialPruned` is properly handled.
-
-        .. testcode::
-
-            import optuna
-            from optuna.integration import PyTorchLightningPruningCallback
-            import pytorch_lightning as pl
-
-
-            def objective(trial: optuna.trial.Trial) -> float:
-                callback = PyTorchLightningPruningCallback(trial, monitor="accuracy")
-                trainer = pl.Trainer(
-                    max_epochs=10,
-                    accelerator="auto",
-                    devices=2,
-                    enable_checkpointing=False,
-                    callbacks=[callback],
-                    strategy=DDPSpawnStrategy(find_unused_parameters=False),
-                )
-
-                model = MyModel()
-                trainer.fit(model)
-
-                callback.check_pruned()
-
-                return trainer.callback_metrics["val_acc"].item()
-
-
-            pruner = optuna.pruners.MedianPruner()
-            study = optuna.create_study(direction="maximize", pruner=pruner)
-            study.optimize(objective, n_trials=100, timeout=600)
-
-
     .. note::
         For the distributed data parallel training, the version of PyTorchLightning needs to be
         higher than or equal to v1.6.0. In addition, :class:`~optuna.study.Study` should be
         instantiated with RDB storage.
+
+
+    .. note::
+        If you would like to use PyTorchLightningPruningCallback in a distributed training
+        environment, you need to evoke `PyTorchLightningPruningCallback.check_pruned()`
+        manually so that :class:`~optuna.exceptions.TrialPruned` is properly handled.
     """
 
     def __init__(self, trial: optuna.trial.Trial, monitor: str) -> None:
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -59,7 +59,9 @@ class PyTorchLightningPruningCallback(Callback):
     def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> None:
         self.is_ddp_backend = trainer._accelerator_connector.is_distributed
         if self.is_ddp_backend:
-            if version.parse(pl.__version__) < version.parse("1.6.0"):  # type: ignore
+            if version.parse(pl.__version__) < version.parse(  # type: ignore[attr-defined]
+                "1.6.0"
+            ):
                 raise ValueError("PyTorch Lightning>=1.6.0 is required in DDP.")
             # If it were not for this block, fitting is started even if unsupported storage
             # is used. Note that the ValueError is transformed into ProcessRaisedException inside
@@ -74,6 +76,7 @@ class PyTorchLightningPruningCallback(Callback):
                 )
             # It is necessary to store intermediate values directly in the backend storage because
             # they are not properly propagated to main process due to cached storage.
+            # TODO(Shinichi) Remove intermediate_values from system_attr after PR #4431 is merged.
             self._trial.storage.set_trial_system_attr(
                 self._trial._trial_id,
                 _INTERMEDIATE_VALUE,
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -202,8 +202,10 @@ class PyTorchLightningPruningCallback(Callback):
         storage cache. Therefore, necessary information is kept in trial_system_attrs when the
         trial runs in a distributed situation. Please call this method right after calling
         ``pytorch_lightning.Trainer.fit()``.
-        Please do not call this method when you are not using DDP.
+        If a callback doesn't have any backend storage for DDP, this method does nothing.
         """
+        if self.is_ddp_backend is False:
+            return
 
         _trial_id = self._trial._trial_id
         _study = self._trial.study
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -56,13 +56,13 @@ class PyTorchLightningPruningCallback(Callback):
         self._trial = trial
         self.monitor = monitor
         self.is_ddp_backend = False
+        # It is necessary to store intermediate values directly in the backend storage because they
+        # are not properly propagated to main process due to cached storage.
         self._trial.storage.set_trial_system_attr(
             self._trial._trial_id,
             _INTERMEDIATE_VALUE,
             dict(),
         )
-        # It is necessary to store intermediate values directly in the backend storage because they
-        # are not properly propagated to main process due to cached storage.
 
     def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> None:
         self.is_ddp_backend = trainer._accelerator_connector.is_distributed
--- a/optuna/integration/pytorch_lightning.py
+++ b/optuna/integration/pytorch_lightning.py
@@ -9,9 +9,8 @@ from optuna.storages._rdb.storage import RDBStorage
 
 # Define key names of `Trial.system_attrs`.
 _EPOCH_KEY = "ddp_pl:epoch"
-_PRUNED_KEY = "ddp_pl:pruned"
-_MESSAGE_KEY = "ddp_pl:message"
 _INTERMEDIATE_VALUE = "ddp_pl:intermediate_value"
+_PRUNED_KEY = "ddp_pl:pruned"
 
 with optuna._imports.try_import() as _imports:
     import pytorch_lightning as pl
@@ -56,13 +55,6 @@ class PyTorchLightningPruningCallback(Callback):
         self._trial = trial
         self.monitor = monitor
         self.is_ddp_backend = False
-        # It is necessary to store intermediate values directly in the backend storage because they
-        # are not properly propagated to main process due to cached storage.
-        self._trial.storage.set_trial_system_attr(
-            self._trial._trial_id,
-            _INTERMEDIATE_VALUE,
-            dict(),
-        )
 
     def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> None:
         self.is_ddp_backend = trainer._accelerator_connector.is_distributed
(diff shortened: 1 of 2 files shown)


Alnusjaponica:
I made this PR review ready. Note that this PR is based on #4322 and should be merged after it.

toshihikoyanase:
@HideakiImamura This is a follow-up for #4322. Could you join the review, please?

toshihikoyanase:
@Alnusjaponica  #4322 was merged into `master`. Could you rebase `master`, please?

toshihikoyanase (commented):
Let me share my early comments.
I guess the change uses system attrs to store intermediate values even if users execute only a single process. I'm thinking of keeping them to simplify the logic. What do you think of it?

Alnusjaponica:
@toshihikoyanase Thank you for your indication. I fixed the comment and am going to make change to use system_attr only under distributed situation.

optuna/integration/pytorch_lightning.py
```diff
@@ -8,9 +8,10 @@
 
 
 # Define key names of `Trial.system_attrs`.
-_PRUNED_KEY = "ddp_pl:pruned"
 _EPOCH_KEY = "ddp_pl:epoch"
-
```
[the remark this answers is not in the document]
Alnusjaponica: Thank you for your comment. I don't have any particular reason for this. Is it better to arrange them in a particular order (e.g. alphabetically)?

optuna/integration/pytorch_lightning.py
```diff
@@ -55,12 +56,22 @@ def __init__(self, trial: optuna.trial.Trial, monitor: str) -> None:
...
+        )
+        # It is necessary to store intermediate values directly in the backend storage because they
+        # are not properly propagated to main process due to cached storage.
 
     def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> None:
         self.is_ddp_backend = trainer._accelerator_connector.is_distributed
         if self.is_ddp_backend:
             if version.parse(pl.__version__) < version.parse("1.6.0"):  # type: ignore
                 raise ValueError("PyTorch Lightning>=1.6.0 is required in DDP.")
+            # If it were not for this block, fitting is started even if unsupported storage
```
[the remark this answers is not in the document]
Alnusjaponica: I meant the latter one.

optuna/integration/pytorch_lightning.py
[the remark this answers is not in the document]
toshihikoyanase: > Is it better to arrange them in a particular order (e.g. alphabetically)

That makes sense. Could you arrange the keys in alphabetical order, please?

toshihikoyanase:
@Alnusjaponica Thank you for your update. As we pair-programmed the code, we may simplify the code in terms of the following points:
- The empty intermediate values in system_attrs are only required for the DDP. We can skip it by moving the logic to `on_fit_start`
- The logic in `on_validation_end` is a bit complicated. We may be separate the logic for single-process optimization and DDP optimization
- `check_pruned` is only used for DDP optimization. It should be documented.
- `check_pruned` is assumed to be used with `_cachedStorage`. We may remove `#type ignored` if we assert it.
- The message can be removed from the system_attrs since the pruned epoch can generate it.

optuna/integration/pytorch_lightning.py
```diff
@@ -69,60 +72,86 @@ def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> Non
...
-            return
+        if trainer.is_global_zero:
+            # Update system_attr from global zero process.
+            self._trial.storage.set_trial_system_attr(self._trial._trial_id, _PRUNED_KEY, True)
+            self._trial.storage.set_trial_system_attr(self._trial._trial_id, _EPOCH_KEY, epoch)
+
+    def check_pruned(self) -> None:
+        """Raise :class:`optuna.TrialPruned` manually if pruned.
+
+        Currently, intermediate_values are not properly propagated between processes due to
```
toshihikoyanase, line 140: A nit,
```suggestion
        Currently, ``intermediate_values`` are not properly propagated between processes due to
```

optuna/integration/pytorch_lightning.py
```diff
@@ -69,60 +72,86 @@ def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> Non
...
+        _CachedStorage. Therefore, necessary information is kept in trial_system_attrs when the
```
toshihikoyanase, line 141: `_CachedStorage` is a private class, so how about mentioning the storage cache without showing the class name?

```suggestion
       storage cache. Therefore, necessary information is kept in trial_system_attrs when the
```

optuna/integration/pytorch_lightning.py
```diff
@@ -69,60 +72,86 @@ def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> Non
...
+        assert isinstance(_study._storage, _CachedStorage)
+        _trial_system_attrs = _study._storage._backend.get_trial_system_attrs(_trial_id)
         is_pruned = _trial_system_attrs.get(_PRUNED_KEY)
-        epoch = _trial_system_attrs.get(_EPOCH_KEY)
-        intermediate_values = _trial.intermediate_values
-        for step, value in intermediate_values.items():
-            self._trial.report(value, step=step)
+        intermediate_values = _trial_system_attrs.get(_INTERMEDIATE_VALUE)
 
+        for epoch, score in intermediate_values.items():  # type: ignore[union-attr]
```
toshihikoyanase, line 153: As talked offline, `intermediate_values` cannot be `None`.
So, how about adding an assertion instead of `type ignore` as follows?

```suggestion
        assert intermediate_values is not None
        for epoch, score in intermediate_values.items():
```

optuna/integration/pytorch_lightning.py
```diff
@@ -69,60 +72,86 @@ def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> Non
...
         should_stop = False
+
+        # Determine if the trial should be terminated in a single process.
+        if not self.is_ddp_backend:
+            self._trial.report(current_score.item(), step=epoch)
+            should_stop = self._trial.should_prune()
+            if not should_stop:
+                return
+            message = f"Trial was pruned at epoch {epoch}."
+            raise optuna.TrialPruned(message)
```
toshihikoyanase, line 109: Maybe we can simplify the code by removing the local variables.
```suggestion
        if not self.is_ddp_backend:
            self._trial.report(current_score.item(), step=epoch)
            if not self._trial.should_prune():
                return
            raise optuna.TrialPruned(f"Trial was pruned at epoch {epoch}.")
```

toshihikoyanase (commented):
Thank you for your update. Let me add some small comments.

tests/integration_tests/test_pytorch_lightning.py
```diff
@@ -94,22 +95,31 @@ def validation_step(
...
 
         self.log("accuracy", accuracy, sync_dist=True)
         return {"validation_accuracy": accuracy}
 
+    def validation_epoch_end(
+        self,
+        output: Union[
+            List[Union["torch.Tensor", Dict[str, Any]]],
+            List[List[Union["torch.Tensor", Dict[str, Any]]]],
+        ],
```
toshihikoyanase, line 111: In general, the official Python document recommends `Sequence` and `Mapping` for arguments instead of `List` and `Dict`. Can we use them here?

> To annotate arguments it is preferred to use an abstract collection type such as [Sequence](https://docs.python.org/3/library/typing.html#typing.Sequence) or [Iterable](https://docs.python.org/3/library/typing.html#typing.Iterable).

https://docs.python.org/3/library/typing.html#typing.List
https://docs.python.org/3/library/typing.html#typing.Dict

tests/integration_tests/test_pytorch_lightning.py
```diff
@@ -143,24 +153,27 @@ def test_pytorch_lightning_pruning_callback_monitor_is_invalid() -> None:
...
             enable_checkpointing=False,
-            callbacks=[PyTorchLightningPruningCallback(trial, monitor="accuracy")],
+            callbacks=[callback],
+            strategy=DDPSpawnStrategy(find_unused_parameters=False),
         )
 
         model = ModelDDP()
         trainer.fit(model)
 
+        # evoke pruning manually.
```
toshihikoyanase, line 174: A nit.
```suggestion
        # Evoke pruning manually.
``````suggestion
        # Evoke pruning manually.
```

tests/integration_tests/test_pytorch_lightning.py
```diff
@@ -179,26 +192,29 @@ def objective(trial: optuna.trial.Trial) -> float:
...
```
toshihikoyanase, line 212: ditto.
```suggestion
        # Evoke pruning manually.
```

optuna/integration/pytorch_lightning.py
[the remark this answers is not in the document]
toshihikoyanase: Thanks!

optuna/integration/pytorch_lightning.py
```diff
@@ -69,60 +72,85 @@ def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> Non
...
         is_pruned = _trial_system_attrs.get(_PRUNED_KEY)
-        epoch = _trial_system_attrs.get(_EPOCH_KEY)
-        intermediate_values = _trial.intermediate_values
-        for step, value in intermediate_values.items():
-            self._trial.report(value, step=step)
+        intermediate_values = _trial_system_attrs.get(_INTERMEDIATE_VALUE)
 
+        assert intermediate_values is not None
+        for epoch, score in intermediate_values.items():
+            self._trial.report(score, step=int(epoch))
```
toshihikoyanase, line 153: Maybe we can remove this logic and just sync the owned trial after #4431 is merged. How about adding TODO comment about it?

toshihikoyanase (commented):
I confirmed that the updated callback worked with the PyTorch Lightning DDP example in optuna-examples. I have a small comment, but the change almost looks good to me.

<details>
<summary>Diff of pytorch_lightning_ddp.py</summary>
```diff
diff --git a/pytorch/pytorch_lightning_ddp.py b/pytorch/pytorch_lightning_ddp.py
index 030834d..6d609cb 100644
--- a/pytorch/pytorch_lightning_ddp.py
+++ b/pytorch/pytorch_lightning_ddp.py
@@ -78,6 +78,9 @@ class LightningNet(pl.LightningModule):
         self.log("val_acc", accuracy, sync_dist=True)
         self.log("hp_metric", accuracy, on_step=False, on_epoch=True, sync_dist=True)
 
+    def validation_epoch_end(self, output) -> None:
+        return
+
     def configure_optimizers(self) -> optim.Optimizer:
         return optim.Adam(self.model.parameters())
 
@@ -124,19 +127,22 @@ def objective(trial: optuna.trial.Trial) -> float:
     model = LightningNet(dropout, output_dims)
     datamodule = FashionMNISTDataModule(data_dir=DIR, batch_size=BATCHSIZE)
 
+    callback = PyTorchLightningPruningCallback(trial, monitor="val_acc")
     trainer = pl.Trainer(
         logger=True,
         limit_val_batches=PERCENT_VALID_EXAMPLES,
         enable_checkpointing=False,
         max_epochs=EPOCHS,
         gpus=-1 if torch.cuda.is_available() else None,
-        accelerator="ddp_cpu" if not torch.cuda.is_available() else None,
+        accelerator="cpu" if not torch.cuda.is_available() else None,
+        strategy="ddp_spawn",
         num_processes=os.cpu_count() if not torch.cuda.is_available() else None,
-        callbacks=[PyTorchLightningPruningCallback(trial, monitor="val_acc")],
+        callbacks=[callback],
     )
     hyperparameters = dict(n_layers=n_layers, dropout=dropout, output_dims=output_dims)
     trainer.logger.log_hyperparams(hyperparameters)
     trainer.fit(model, datamodule=datamodule)
+    callback.check_pruned()
 
     return trainer.callback_metrics["val_acc"].item()
```
</details>

optuna/integration/pytorch_lightning.py
Alnusjaponica: Thank you for your suggestion. I added TODO.

toshihikoyanase: Thanks! I'll approve this PR after all CI jobs are complete.

toshihikoyanase (approved):
We have some follow-up tasks as described in TODO comments, but I think we can work on them in a new PR.

LGTM. Thank you!

optuna/integration/pytorch_lightning.py
```diff
@@ -69,60 +74,86 @@ def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> Non
...
                     " supports only optuna.storages.RDBStorage in DDP."
                 )
+            # It is necessary to store intermediate values directly in the backend storage because
+            # they are not properly propagated to main process due to cached storage.
+            # TODO(Shinichi) Remove intermediate_values from system_attr after PR #4431 is merged.
+            self._trial.storage.set_trial_system_attr(
+                self._trial._trial_id,
+                _INTERMEDIATE_VALUE,
+                dict(),
+            )
```
HideakiImamura, line 84: This setup only should be done in the rank zero process? In the following scenario, the value of the `_INTERMEDIATE_VALUE` is wrong.

```
------------------------> time
rank 0: (A) -> (B)
rank 1:                (A)
```
(A): call `on_fit_start` and init the `_INTERMEDIATE_VALUE`
(B): call `on_validation_end` and report the `_INTERMEDIATE_VALUE`
Actually, thanks to the specification of `on_fit_start` and `on_validation_end`, this problem might not occur, but I think it is safe to call this init only in the rank 0 process for future specification change of Pytorch Lightning.

optuna/integration/pytorch_lightning.py
```diff
@@ -69,60 +74,86 @@ def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> Non
...
-
-    def on_fit_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
-        if not self.is_ddp_backend:
-            return
+        if trainer.is_global_zero:
+            # Update system_attr from global zero process.
+            self._trial.storage.set_trial_system_attr(self._trial._trial_id, _PRUNED_KEY, True)
+            self._trial.storage.set_trial_system_attr(self._trial._trial_id, _EPOCH_KEY, epoch)
+
+    def check_pruned(self) -> None:
```
HideakiImamura, line 138: This function is intended to be called by users by hand after the `Trainer.fit`. How about adding the concrete instruction on the document about where this function should be called?

optuna/integration/pytorch_lightning.py
```diff
@@ -69,60 +74,86 @@ def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> Non
...
+        assert isinstance(_study._storage, _CachedStorage)
+        _trial_system_attrs = _study._storage._backend.get_trial_system_attrs(_trial_id)
         is_pruned = _trial_system_attrs.get(_PRUNED_KEY)
-        epoch = _trial_system_attrs.get(_EPOCH_KEY)
-        intermediate_values = _trial.intermediate_values
-        for step, value in intermediate_values.items():
-            self._trial.report(value, step=step)
+        intermediate_values = _trial_system_attrs.get(_INTERMEDIATE_VALUE)
 
+        assert intermediate_values is not None
```
HideakiImamura, line 154: If `self.is_ddp_backend = False`, this function raises an `AssertionError` here. How about early returning this function if `self.is_ddp_backend = False` to allow users to use same codes w/ or w/o DDP?

HideakiImamura (commented):
Thanks for the PR and sorry for the late reply. I checked the overall codes and basically looks good to me. Could you check my several comment?

optuna/integration/pytorch_lightning.py
Alnusjaponica: I added some explanation and example codes in docstrings.

optuna/integration/pytorch_lightning.py
Alnusjaponica: Thank you for your advice. I updated this method accordingly.

optuna/integration/pytorch_lightning.py
Alnusjaponica: Thank you for your careful review. I added an if-block so that `storage.set_trial_system_attr._INTERMEDIATE_VALUE` is initialized only from global zero.

optuna/integration/pytorch_lightning.py
```diff
@@ -69,60 +81,90 @@ def on_fit_start(self, trainer: Trainer, pl_module: "pl.LightningModule") -> Non
...
+        If a callback doesn't have any backend storage for DDP, this method does nothing.
+        """
 
-        # Because on_validation_end is executed in spawned processes,
-        # _trial.report is necessary to update the memory in main process, not to update the RDB.
         _trial_id = self._trial._trial_id
         _study = self._trial.study
-        _trial = _study._storage._backend.get_trial(_trial_id)  # type: ignore
-        _trial_system_attrs = _study._storage.get_trial_system_attrs(_trial_id)
+        if not isinstance(_study._storage, _CachedStorage):
```
Alnusjaponica, line 158:
```suggestion

assert isinstance(_study._storage, _CachedStorage)
```

HideakiImamura (approved):
Thanks for the update. LGTM.

PR #4400, opened 2023-02-07
Use `ubuntu-latest` in PyPI publish CI

not522:
## Motivation
The PyPI publish CI has used ubuntu-18.04, which is deprecated.

## Description of the changes
- Use `ubuntu-latest` in PyPI publish CI

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   build-n-publish:
     name: Build and publish Python distributions to PyPI or TestPyPI
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
 
     # Not intended for forks.
     if: github.repository == 'optuna/optuna'


toshihikoyanase:
@cross32768 Could you review this PR, please?

@not522 The pypi publish jobs are not triggered by PR. Do you think we should check the behavior before wer merge the change? Or, we can fix it if the daily CI failed after merging this PR?

not522:
> Do you think we should check the behavior before wer merge the change?

It's reasonable. I'll check that we can upload it to TestPyPI.

toshihikoyanase (approved):
I confirmed that optuna packages could be built on `ubuntu:22.04` using docker. They passed `twine check`.

And I also found that `twine` is tested on `ubuntu-latest`.
https://github.com/pypa/twine/blob/9c2c0a1c535155931c3d879359330cb836950c6a/.github/workflows/main.yml#L46

So, I think testing this PR on testPyPI may be over-engineering, and I'd like to approve this PR.

cross32768 (approved):
I also confirmed that `twine check` returns "passed" by the process shown below (maybe the same process as above reviewer). LGTM!

* `docker run -it ubuntu:22.04`
* install required dependencies (`apt install git`, `pip install twine`, etc)
* `git clone https://github.com/optuna/optuna`
* `python3 -m build`
* `twine check dist/optuna-3.2.0.dev0-py3-none-any.whl`

It returns:
`Checking dist/optuna-3.2.0.dev0-py3-none-any.whl: PASSED
`

toshihikoyanase:
Let me merge this PR since it got two approvals. Thank you.

PR #4401, opened 2023-02-07
Implement the evaluator of regret bounds and its GP backend for Optuna Terminator 🤖

g-votte:
This PR is part of #4398, adding the evaluator class for regret bounds and its Gaussian process backend.

## Motivation
See #4398 for details. As part of Optuna Terminator, this PR implements the components to evaluate regret bounds.

## Description of the changes
The `RegretBoundEvaluator` class calculates the latest regret bound of the input history. The following snippet shows an example of evaluating the regret bound.

```python
from optuna import create_study
from optuna.terminator.regret.evaluator import RegretBoundEvaluator

def objective(trial):
    x = trial.suggest_float("x", -100, 100)
    return x ** 2

study = create_study()
study.optimize(objective, n_trials=100)

evaluator = RegretBoundEvaluator()
rb = evaluator.evaluate(trials=study.trials, study_direction=study.direction)
print(rb)
```

As described in [1], regret bounds are estimated by a Gaussian process (GP). GP is abstracted as `BaseGaussianProcess` that provides methods to estimate minimum values of UCB and LCB. In this PR, the concrete class of GP is implemented using [GPyTorch](https://gpytorch.ai/).

## References
[1] Makarova et al. "Automatic termination for hyperparameter optimization." International Conference on Automated Machine Learning (2022).

Issue #4398, opened 2023-02-07
[RFC] Introduce automatic terminator for hyperparameter optimization (Optuna Terminator 🤖)


g-votte:
### Motivation

This issue proposes a set of new features, referred to as "Optuna Terminator," for automatically stopping an Optuna study. Optuna Terminator is based on the method proposed by Makarova et al. in "[Automatic Termination for Hyperparameter Optimization](https://www.amazon.science/publications/automatic-termination-for-hyperparameter-optimization)" [1].

As the optimization progresses, the objective function often becomes noisy and the room left for the optimization decreases in many use cases. However, it can be difficult for the user to detect the noise and to determine when to terminate the optimization. The user tends to continue the study for a long time, even when the optimization is no longer beneficial, resulting in wasted computation resources and increased risk of overfitting the hyperparameters.

Optuna Terminator addresses this problem by visualizing the room left for optimization and automatically terminating the optimization at the appropriate time.

![image](https://user-images.githubusercontent.com/1061922/217133233-269e7d36-0fb6-4b8e-911e-9a51e3419fa8.png)

### Description

Optuna Terminator provides the following features:
- evaluating and visualizing the room left for the optimization as a metric called "regret bounds," and
- specifically with cross validation (CV), a callback to automatically stops the study when the CV variance exceeds the regret bound.

### Visualizing regret bounds
The following snippet provides an example of evaluating the regret bounds, i.e., the room left for the optimization.

```python
from optuna.visualization import plot_regret_bounds
...
study.optimize(objective, n_trials=50)
plot_regret_bounds(study).show()
```
Then, the room left for the optimization is visualized as shown in the following figure.

![image](https://user-images.githubusercontent.com/1061922/217133315-ae4944a1-d089-47d1-bfe7-23893c08cb0c.png)

### Automatic termination
The following snippet provides an example of automatically terminating the optimization.

```python
from optuna.terminator.callback import TerminatorCallback
...

def objective(trial):
    ...
    trial.set_user_attr("cv_scores", scores)
    ...

terminator = TerminatorCallback()
...
study.optimize(objective, callbacks=[terminator])
```

Then, the `study.stop` method will be automatically invoked when the CV variance exceeds the regret bound. Note that the number of trials, `n_trials`, is not specified in this example.

### References
[1] Makarova et al. "Automatic termination for hyperparameter optimization." International Conference on Automated Machine Learning (2022).

g-votte:
[...]

PR #4405, opened 2023-02-08
Implement the termination logic and APIs of Optuna Terminator 🤖

g-votte:
Note: please merge this PR after #4401

This PR is a part of #4398 and adds the termination logic and APIs of Optuna Terminator.

## Motivation
Please refer to #4398 for the whole picture of Optuna Terminator. As part of that, this PR implements the logic and APIs for automatic termination.

The following examples demonstrate how to apply automatic termination with those APIs.

You can plug `TerminatorCallback` into the `study.optimize` method.
```python
import optuna
from sklearn.ensemble import RandomForestClassifier

from optuna.terminator.serror import report_cross_validation_scores


    clf = RandomForestClassifier(
        max_depth=trial.suggest_int("max_depth", 2, 32),
        min_samples_split=trial.suggest_float("min_samples_split", 0, 1),
        criterion=trial.suggest_categorical("criterion", ("gini", "entropy")),
    )

    return scores.mean()


terminator = TerminatorCallback()
study.optimize(objective, callbacks=[terminator])
```

For the ask-and-tell API, you can directly use the `Terminator` class and call the `should_terminate` method after each trial.
```python
from sklearn.ensemble import RandomForestClassifier

import optuna
from optuna.terminator.serror import report_cross_validation_scores


terminator = Terminator()
min_n_trials = 20

while True:
    trial = study.ask()

    clf = RandomForestClassifier(
        max_depth=trial.suggest_int("max_depth", 2, 32),
        min_samples_split=trial.suggest_float("min_samples_split", 0, 1),
        criterion=trial.suggest_categorical("criterion", ("gini", "entropy")),
    )

    value = scores.mean()
    print(f"Trial #{trial.number} finished with value {value}.")
    study.tell(trial, value)

    if trial.number > min_n_trials and terminator.should_terminate(study):
        print("Terminated by Optuna Terminator!")
        break
```

## Description of the changes
The following figure illustrates the relationship between the components of Optuna Terminator. Optuna Terminator decides when to end optimization based on two metrics: "regret bound" and "statistical error." Both metrics are estimated using the optimization history stored in the storage.

image

This PR introduces the following components.
- The `StatisticalErrorEvaluator` classes evaluate the statistical error coming from the dataset.
    - For cross-validation (CV), the `CrossValidationStatisticalErrorEvaluator` calculate this metric using the CV scores at the best trial.
    - The `StaticStatisticalErrorEvaluator` returns a constant value, supporting use-cases where the user specifies a fixed regret bound threshold. 
- The `Terminator` class compares the regret bound and the statistical error, and terminates the optimization when the regret bound is small enough. 
- The `TerminatorCallback` class is a wrapper of the `Terminator` that can be plugged into the `study.optimize` method. 

The `RegretBoundEvaluator` and its GP backend are planned to be introduced in a separated PR (#4401).


g-votte:
The comparison diff may appear very large because #4401 has not been merged yet. To view the actual changes made by this PR, please see: https://github.com/g-votte/optuna/compare/terminator-gp...g-votte:optuna:terminator-terminator

optuna/terminator/serror.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def evaluate(
+        self,
+        study: Study,
+    ) -> float:
+        if len(study.get_trials(states=(TrialState.COMPLETE,))) == 0:
+            raise ValueError(
+                "Statistical error cannot be calculated because no trial has been completed."
+            )
+
```
hvy, line 53: I didn't understand why this error was needed, given the logic of simply returning `self._constant` but does it make sense in a broader context? A comment might be helpful for future maintainers.

optuna/terminator/callback.py
```diff
@@ -0,0 +1,34 @@
...
+from optuna.terminator.terminator import BaseTerminator
+from optuna.terminator.terminator import Terminator
+from optuna.trial import FrozenTrial
+from optuna.trial import TrialState
+
+
+_logger = get_logger(__name__)
+
+
+class TerminatorCallback:
```
hvy, line 15: Is it time to start considering `optuna.callbacks` ? We'll have `optuna.terminator.callback.TerminatorCallback` and `optuna.study.MaxTrialsCallback` with similar behaviors and APIs.

hvy (commented):
Looks great in my opinion.
Left some comments and questions though.

Please note that I'm not confident on the algorithm. I understand that the algorithmic part is very small in this PR.

optuna/terminator/callback.py
```diff
@@ -0,0 +1,34 @@
...
+    def __call__(self, study: Study, trial: FrozenTrial) -> None:
+        trials = study.get_trials(states=[TrialState.COMPLETE])
+
+        if len(trials) < self._min_n_trials:
+            return
+
+        should_terminate = self._terminator.should_terminate(study=study)
+
+        if should_terminate:
+            _logger.info("The terminator stopped the study because noise exceeds regret bound.")
```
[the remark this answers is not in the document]
g-votte: That's right. How about simply saying `The study has been stopped by the terminator.`?

optuna/terminator/callback.py
g-votte: Hmm, I think there are pros and cons of putting all callback classes in `optuna.callbacks`. The module structure may be simple for the user, but it also accumulates dependencies of the callback functionalities. In particular, the GP-related dependencies of `TerminatorCallback` may make the dependency puzzle even more complex.

optuna/terminator/serror.py
```diff
@@ -0,0 +1,54 @@
...
+        cv_scores = study.best_trial.user_attrs[self._user_attr_key]
+
+        k = len(cv_scores)
+        scale = 1 / k + 1 / (k - 1)
+
+        var = scale * np.var(cv_scores)
+
+        return float(var)
```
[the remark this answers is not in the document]
g-votte: > As an alternative, we could avoid using user attributes completely. If we introduce some helper function/method that internally sets system attributes (instead of user attributes), we could do validation when setting.

I prefer this option. I changed the interface so that the user calls the `report_cross_validation_scores` function instead of setting the scores as a user attr. PTAL. https://github.com/optuna/optuna/pull/4405/commits/7ab54308033810383aff9b6b4cab138635b41bf5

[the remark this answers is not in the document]
g-votte: Also, the length of the list is now checked in `report_cross_validation_scores`. https://github.com/optuna/optuna/pull/4405/commits/d23413ce8ab831e534149acff16066cf32ad156d

optuna/terminator/callback.py
c-bata: I agree with @g-votte.

In terms of simplicity of module structure, It would be clearer to users if the all terminator-related APIs (e.g. `TerminatorCallback`, `report_cross_validation_scores`, etc.) are placed under the `optuna.terminator` module.

optuna/terminator/serror.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+@experimental_class("3.2.0")
+def report_cross_validation_scores(trial: Trial, scores: List[float]) -> None:
```
c-bata, line 54: [nit] How about using `list[float]` with `from __future__ import annotations` since we've already dropped Python 3.6 support? `typing.List` was actually being deprecated from Python 3.9.
https://peps.python.org/pep-0585/#implementation

optuna/terminator/serror.py
```diff
@@ -0,0 +1,74 @@
...
+@experimental_class("3.2.0")
+class CrossValidationStatisticalErrorEvaluator(BaseStatisticalErrorEvaluator):
+    def evaluate(
+        self,
+        study: Study,
+    ) -> float:
+        if len(study.get_trials(states=(TrialState.COMPLETE,))) == 0:
+            raise ValueError(
+                "Statistical error cannot be calculated because no trial has been completed."
+            )
```
c-bata, line 33: Do you think it is enough to check that the `min_n_trials` argument of `TerminatorCallback` is greater than 0? Not a strong opinion, but it can reduce a storage API call.

c-bata (commented):
Thank you for your pull request. Changes basically look good to me. I left some questions.

Just curious, do you plan to work on adding docstrings or updating documentation as a follow-up?

optuna/terminator/serror.py
```diff
@@ -0,0 +1,74 @@
...
+
+        var = scale * np.var(cv_scores)
+
+        return float(var)
+
+
+@experimental_class("3.2.0")
+def report_cross_validation_scores(trial: Trial, scores: List[float]) -> None:
+    if len(scores) <= 1:
+        raise ValueError("The length of `scores` are expected to be greater than one.")
```
hvy, line 56: Nitpicking.
```suggestion
        raise ValueError("The length of `scores` is expected to be greater than one.")
```

optuna/terminator/serror.py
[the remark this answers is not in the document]
hvy: Thanks for the update, let me resolve this conversation.

optuna/terminator/callback.py
hvy: Sounds fair, at least for now, given that there also are multiple public components to the terminator besides the callback class itself. Thank you for the discussion.

hvy:
A detail regarding the API but how about sticking to being explicit with re-exported? C.f. https://github.com/optuna/optuna/pull/2013

optuna/terminator/terminator.py
```diff
@@ -0,0 +1,37 @@
...
+from optuna.terminator.serror import CrossValidationStatisticalErrorEvaluator
+
+
+class BaseTerminator(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def should_terminate(self, study: Study) -> bool:
+        pass
+
+
+class Terminator(BaseTerminator):
```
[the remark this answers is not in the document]
c-bata: I'm also concerned about the same point.

For example, instead of terminating based on cross-validation scores, we could add methods to see if the objective value is improved at a specific interval. In that case, this class name may not be appropriate and `CrossValidationTerminator` may be better.

optuna/terminator/serror.py
g-votte: Thanks. Let me keep the assertion because this is a public API and the user technically calls it with an empty study.
Meanwhile, I removed the `ValueError` and just put `assert`. https://github.com/optuna/optuna/pull/4405/commits/e93a62571272d4dfc32e3435aa0742d4e87abd61

optuna/terminator/serror.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class CrossValidationStatisticalErrorEvaluator(BaseStatisticalErrorEvaluator):
+    def __init__(self, user_attr_key: str = "cv_scores") -> None:
+        self._user_attr_key = user_attr_key
+
+    def evaluate(
+        self,
+        study: Study,
+    ) -> float:
```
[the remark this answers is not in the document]
g-votte: Thanks. I reflected your suggestion with this commit: https://github.com/optuna/optuna/pull/4405/commits/412c1eede16eb1905952f856ad7d6a5ba391ae77

optuna/terminator/serror.py
g-votte: Cool. Replaced `typing.List` with `list`: https://github.com/optuna/optuna/pull/4405/commits/76c745be11edfa944c435b67789a2ce545615791

optuna/terminator/serror.py
g-votte: Thanks. I removed this check: https://github.com/optuna/optuna/pull/4405/commits/e93a62571272d4dfc32e3435aa0742d4e87abd61

optuna/terminator/terminator.py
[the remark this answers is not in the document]
g-votte: Though the name, `Terminator`, is too general, I'd like to adopt it for the following reasons.
- this would be easier for the user to remember
- we can rename it later, since this API is still experimental

optuna/terminator/terminator.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(
+        self,
+        improvement_evaluator: Optional[BaseImprovementEvaluator] = None,
+        error_evaluator: Optional[BaseErrorEvaluator] = None,
+        min_n_trials: int = DEFAULT_MIN_N_TRIALS,
+    ) -> None:
+        self._improvement_evaluator = improvement_evaluator or RegretBoundEvaluator()
+        self._error_evaluator = error_evaluator or CrossValidationErrorEvaluator()
+        self._min_n_trials = min_n_trials
```
c-bata, line 30: How about raising a ValueError exception when `min_n_trials` is less than or equal to 0? Otherwise, `CrossValidationErrorEvaluator` may raise an AssertionError exception I guess.

optuna/terminator/erroreval.py
```diff
@@ -0,0 +1,59 @@
...
+_CROSS_VALIDATION_SCORES_KEY = "terminator:cv_scores"
+
+
+class BaseErrorEvaluator(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def evaluate(self, study: Study) -> float:
+        pass
+
+
+@experimental_class("3.2.0")
```
c-bata, line 22: [nit] Multiple experimental warnings are displayed when using TerminatorCallback.

```python

terminator = TerminatorCallback()
# 
# /.../example.py:29: ExperimentalWarning: TerminatorCallback is experimental (supported from v3.2.0). The interface can change in the future.
#   terminator = TerminatorCallback()
# /.../callback.py:20: ExperimentalWarning: Terminator is experimental (supported from v3.2.0). The interface can change in the future.
#   self._terminator = terminator or Terminator()
# /.../terminator.py:29: ExperimentalWarning: CrossValidationErrorEvaluator is experimental (supported from v3.2.0). The interface can change in the future.
#   self._error_evaluator = error_evaluator or CrossValidationErrorEvaluator()
```

Do you think these warning messages should be reduced? Perhaps we could only display the experimental warning when the Terminator class is initialized or when the `optuna.terminator` module is imported.

c-bata (commented):
The changes almost look good to me. I left some minor comments though.

optuna/terminator/terminator.py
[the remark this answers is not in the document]
hvy: I think that's reasonable, also given we cannot come up with better alternatives that are not too verbose. Let me resolve this comment for now although it's good for record purpose.

hvy:
I'm now wondering if don't want to re-export after all. Open for comments.

```python
>>> import optuna
>>> optuna.study  # OK
>>> optuna.trial  # OK
>>> optuna.terminator  # Fails, as expected?
AttributeError: module 'optuna' has no attribute 'terminator'
>>> import optuna.terminator  # OK
```

C.f. https://github.com/optuna/optuna/pull/4405#issuecomment-1447645746

optuna/terminator/erroreval.py
```diff
@@ -0,0 +1,59 @@
...
+            cv_scores = best_trial_attrs[_CROSS_VALIDATION_SCORES_KEY]
+        else:
+            raise ValueError(
+                "Cross-validation scores have not been reported. Please call "
+                "`report_cross_validation_scores(trial, scores)` during a trial and pass the "
+                "list of scores as `scores`."
+            )
+
+        k = len(cv_scores)
+        scale = 1 / k + 1 / (k - 1)
```
hvy, line 38: Might improve code readability addressing zero-division.
```suggestion
        k = len(cv_scores)
        assert k > 1, "Should be guaranteed by `report_cross_validation_scores`."
        scale = 1 / k + 1 / (k - 1)
```

optuna/terminator/terminator.py
```diff
@@ -0,0 +1,45 @@
...
+        pass
+
+
+@experimental_class("3.2.0")
+class Terminator(BaseTerminator):
+    def __init__(
+        self,
+        improvement_evaluator: Optional[BaseImprovementEvaluator] = None,
+        error_evaluator: Optional[BaseErrorEvaluator] = None,
+        min_n_trials: int = DEFAULT_MIN_N_TRIALS,
```
hvy, line 26: (optional) This might be more consistent with other parameters in Optuna `n_startup_trials`, `n_ei_candidates`, etc. We could also change the constant's name.
```suggestion
        n_min_trials: int = DEFAULT_MIN_N_TRIALS,
```

optuna/terminator/terminator.py
hvy: (Noticed we actually get a `ValueError` from the `RegretBoundEvaluator` if `min_n_trials` is `0`.)

optuna/terminator/erroreval.py
```diff
@@ -0,0 +1,59 @@
...
+class BaseErrorEvaluator(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def evaluate(self, study: Study) -> float:
+        pass
+
+
+@experimental_class("3.2.0")
+class CrossValidationErrorEvaluator(BaseErrorEvaluator):
+    def evaluate(self, study: Study) -> float:
+        assert len(study.get_trials(states=(TrialState.COMPLETE,))) != 0
```
hvy, line 25: (optional) This method seems like a public API in which case a `ValueError` would be preferred over an assertion in my opinion. Maybe this was discussed before, it's not a strong opinion.

hvy, line 25: Personally find this more readable. Length is never negative, etc.
```suggestion
        assert len(study.get_trials(states=(TrialState.COMPLETE,))) > 0
```

optuna/terminator/erroreval.py
```diff
@@ -0,0 +1,59 @@
...
+
+        best_trial_attrs = study.best_trial.system_attrs
+        if _CROSS_VALIDATION_SCORES_KEY in best_trial_attrs:
+            cv_scores = best_trial_attrs[_CROSS_VALIDATION_SCORES_KEY]
+        else:
+            raise ValueError(
+                "Cross-validation scores have not been reported. Please call "
+                "`report_cross_validation_scores(trial, scores)` during a trial and pass the "
+                "list of scores as `scores`."
+            )
```
hvy, line 35: For the record, could we leave a comment why we made this a strict error instead of a perhaps more relaxed version to not stop the study? (Is the reason technical or based on user experience?)

optuna/terminator/erroreval.py
```diff
@@ -0,0 +1,59 @@
...
+        scale = 1 / k + 1 / (k - 1)
+
+        var = scale * np.var(cv_scores)
+        std = np.sqrt(var)
+
+        return float(std)
+
+
+@experimental_class("3.2.0")
+def report_cross_validation_scores(trial: Trial, scores: list[float]) -> None:
```
hvy, line 47: Sorry for not having mentioned this sooner but I'm finding it just slightly confusing to have this function under the same sub-module as `StaticErrorEvalator` . I think we can leave it as is though as those are experimental APIs. (We could make it a static method of `CrossValidationErrorEvaluator` but it's a matter of preference).

tests/terminator_tests/test_erroreval.py
```diff
@@ -0,0 +1,84 @@
...
+        system_attrs={_CROSS_VALIDATION_SCORES_KEY: cv_scores},
+    )
+
+
+def test_cross_validation_evaluator() -> None:
+    study = create_study(direction="minimize")
+    study.add_trials(
+        [
+            _create_trial(value=2.0, cv_scores=[1.0, -1.0]),  # second best trial with 1.0 var
+            _create_trial(value=1.0, cv_scores=[2.0, -2.0]),  # best trial with 4.0 var
```
hvy, line 30: Nit. Sorry I haven't checked the line length.
```suggestion
            _create_trial(value=2.0, cv_scores=[1.0, -1.0]),  # Second best trial with 1.0 var.
            _create_trial(value=1.0, cv_scores=[2.0, -2.0]),  # Best trial with 4.0 var.
```

hvy (changes requested):

optuna/terminator/terminator.py
g-votte: Thanks for your comments. A non-positive `min_n_trials` now raises `ValueError`.
https://github.com/optuna/optuna/pull/4405/commits/ce89ef0f72bf1265b08ff864ef3cae8b35ec488f

optuna/terminator/erroreval.py
g-votte: Good catch. This may cause a little negative impact for the user experience. Though, let us discuss as a follow-up task.

optuna/terminator/terminator.py
g-votte: Hmm, I'd like to adopt `min_n_trials` because this is the minimum number of the number of trials.

optuna/terminator/erroreval.py
g-votte: Actually, this was changed from `ValueError` to `assert`. https://github.com/optuna/optuna/pull/4405#discussion_r1121193853

Let us adopt `assert` in this PR.

optuna/terminator/erroreval.py
g-votte: To make the API call short and simple, let us keep `report_cross_validation_scorers` an isolated function.

optuna/terminator/terminator.py
hvy: Got it, let's leave it as is.

c-bata (approved):
LGTM!

hvy (approved):
LGTM!

PR #4406, opened 2023-02-08
fix colorlog compatibility problem

### LINKED ISSUE #4403: #4376 breaks compatibility with old `colorlog`
### Expected behavior

`import optuna` should not fail.

### Environment

- Optuna version:
- Python version:
- OS:
- (Optional) Other libraries and their versions: **colorlog v4.8**

```shell
File ~/optuna/optuna/logging.py:36, in create_default_formatter()
     31 def create_default_formatter() -> colorlog.ColoredFormatter:
     32     """Create a default formatter of log messages.
     33 
     34     This function is not supposed to be directly accessed by library users.
     35     """
---> 36     return colorlog.ColoredFormatter(
     37         "%(log_color)s[%(levelname)1.1s %(asctime)s]%(reset)s %(message)s",
     38         no_color=False if _color_supported() else True,
     39     )

TypeError: ColoredFormatter.__init__() got an unexpected keyword argument 'no_color'
```

### Steps to reproduce

1.
```
$ pip install colorlog==4.8.0
```
2.
```
import optuna
```

Updating to the latest `colorlog` fixes this.

gen740:
## Motivation

Fix #4403.

## Description of the changes

Use logging.Formatter if `_color_supported` returns True.

### DIFF
--- a/optuna/logging.py
+++ b/optuna/logging.py
@@ -3,7 +3,6 @@ from logging import CRITICAL
 from logging import DEBUG
 from logging import ERROR
 from logging import FATAL
-from logging import Formatter
 from logging import INFO
 from logging import WARN
 from logging import WARNING
@@ -11,7 +10,6 @@ import os
 import sys
 import threading
 from typing import Optional
-from typing import Union
 
 import colorlog
 
@@ -30,7 +28,7 @@ _lock: threading.Lock = threading.Lock()
 _default_handler: Optional[logging.Handler] = None
 
 
-def create_default_formatter() -> Union[Formatter, colorlog.ColoredFormatter]:
+def create_default_formatter() -> logging.Formatter:
     """Create a default formatter of log messages.
 
     This function is not supposed to be directly accessed by library users.
@@ -41,7 +39,7 @@ def create_default_formatter() -> Union[Formatter, colorlog.ColoredFormatter]:
         return colorlog.ColoredFormatter(
             f"%(log_color)s{header}%(reset)s {message}",
         )
-    return Formatter(f"{header} {message}")
+    return logging.Formatter("[%(levelname)1.1s %(asctime)s] %(message)s")
 
 
 def _color_supported() -> bool:
--- a/optuna/logging.py
+++ b/optuna/logging.py
@@ -35,11 +35,13 @@ def create_default_formatter() -> Union[Formatter, colorlog.ColoredFormatter]:
 
     This function is not supposed to be directly accessed by library users.
     """
+    header = "[%(levelname)1.1s %(asctime)s]"
+    message = "%(message)s"
     if _color_supported():
         return colorlog.ColoredFormatter(
-            "%(log_color)s[%(levelname)1.1s %(asctime)s]%(reset)s %(message)s",
+            f"%(log_color)s{header}%(reset)s {message}",
         )
-    return Formatter("[%(levelname)1.1s %(asctime)s] %(message)s")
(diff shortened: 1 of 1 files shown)


optuna/logging.py
```diff
@@ -28,15 +30,16 @@
 _default_handler: Optional[logging.Handler] = None
 
 
-def create_default_formatter() -> colorlog.ColoredFormatter:
+def create_default_formatter() -> Union[Formatter, colorlog.ColoredFormatter]:
```
toshihikoyanase, line 33: `colorlog.ColoredFormatter` extends `logging.Formatter`, and I guess the return type can be `Formatter`.

```suggestion
def create_default_formatter() -> Formatter:
```

And, the returned value is pass to `logging.Handler.setFormatter`, and it expects `logging.Formatter`. So, I don't think it does not depend on the `colorlog` version.

https://github.com/optuna/optuna/blob/ea55b34f9205d3e52fd4d825d4bf89c9daaf6696/optuna/logging.py#L70

optuna/logging.py
```diff
@@ -28,15 +30,16 @@
...
     """
-    return colorlog.ColoredFormatter(
-        "%(log_color)s[%(levelname)1.1s %(asctime)s]%(reset)s %(message)s",
-        no_color=False if _color_supported() else True,
-    )
+    if _color_supported():
+        return colorlog.ColoredFormatter(
+            "%(log_color)s[%(levelname)1.1s %(asctime)s]%(reset)s %(message)s",
+        )
+    return Formatter("[%(levelname)1.1s %(asctime)s] %(message)s")
```
toshihikoyanase, line 42: I'm not sure, but do you think we can simplify the format strings by extracting the common parts?

```suggestion
    header = "[%(levelname)1.1s %(asctime)s]"
    message = "%(message)s"
    if _color_supported():
        return colorlog.ColoredFormatter(
            f"%(log_color)s{header}%(reset)s {message}",
        )
    return Formatter(f"{header} {message}")
```

toshihikoyanase (commented):
Thank you for your PR. I have two comments.

toshihikoyanase:
@contramundum53 Could you review this PR as an author of https://github.com/optuna/optuna/issues/4403?

contramundum53:
I confirmed that it works with colorlog 4.8. Thanks for the PR!

contramundum53 (approved):

contramundum53:

optuna/logging.py
```diff
@@ -28,15 +28,18 @@
...
-        "%(log_color)s[%(levelname)1.1s %(asctime)s]%(reset)s %(message)s",
-        no_color=False if _color_supported() else True,
-    )
+    header = "[%(levelname)1.1s %(asctime)s]"
+    message = "%(message)s"
+    if _color_supported():
+        return colorlog.ColoredFormatter(
+            f"%(log_color)s{header}%(reset)s {message}",
+        )
+    return logging.Formatter("[%(levelname)1.1s %(asctime)s] %(message)s")
```
toshihikoyanase, line 42: Thank you for your update. I think we can use `header` and `message` here.
```suggestion
    return logging.Formatter(f"{header} {message}")
```

gen740: Oops, I forgot to change it. Thank you!

keisuke-umezawa (approved):
LGTM! I confirmed it works on my local env for both situations. Thank you for your quick fix.

PR #4281, opened 2022-12-23
Fix mypy error abut `pytorch_distributed.py` in Checks (integration)

Alnusjaponica:
## Motivation
Solve scheduled Checks (Integration) failure

## Description of the changes
Currently, scheduled Checks (Integration) fails die to mypy check with `--warn-unused-ignores` option. This PR removes a part of the cause related to `pytorch_distributed.py`.
Although test for  Checks (Integration) does't run when PR in opened, I confirmed this PR together with other "Fix mypy error" series reduces mypy error at [CI on my fork](https://github.com/Alnusjaponica/optuna/pull/3).

PR #4410, opened 2023-02-09
Fix mypy error about `Chainer`

Alnusjaponica:
## Motivation
Resolve #4116

## Description of the changes
Ignore mypy error which occurs at the 23rd line of `tests/integration_tests/test_chainer.py`.
Note that mypy error of the line was used to be ignored before PR #4276 but I removed `type: ignore` to reduce unused warnings. Mypy recognize this as error again from 1.0.0.

PR #4413, opened 2023-02-09
Fix mypy error about `PyTorch Distributed`

### LINKED ISSUE #4116: Resolve mypy error on master branch
### Motivation

Due to the recent release of pytorch-lightning, `Checks (Integration)` fails on the master branch 
https://github.com/optuna/optuna/actions/runs/3381750055. 

### Suggestion

[The changelog](https://github.com/Lightning-AI/lightning/blob/e52d6c5b35e5fd42810dca949ee408e2c831047a/src/pytorch_lightning/CHANGELOG.md
) says 

> Renamed `training_type_plugin` file to `strategy` 

So we can replace `training_type_plugin` with `strategy` in the integration file.

The pytorch-lightning integration is not tested while in optuna https://github.com/optuna/optuna/issues/3418.

Alnusjaponica:
## Motivation
Resolve #4116

## Description of the changes
This PR fixes mypy error in `integration/pytorch_distributed.py`, by making it explicit that `__init__()` of `TorchDistributedTrial` changes a global variable. Note that mypy error of the line was used to be ignored before PR #4281 but I removed `type: ignore` to reduce unused warnings. Mypy recognize this as error again from 1.0.0.

### DIFF
--- a/optuna/integration/pytorch_distributed.py
+++ b/optuna/integration/pytorch_distributed.py
@@ -36,7 +36,7 @@ _suggest_deprecated_msg = (
     "Use :func:`~optuna.integration.TorchDistributedTrial.suggest_float` instead."
 )
 
-_g_pg: List[Optional["ProcessGroup"]] = [None]
+_g_pg: Optional["ProcessGroup"] = None
 
 
 def broadcast_properties(f: "Callable[_P, _T]") -> "Callable[_P, _T]":
@@ -123,7 +123,7 @@ class TorchDistributedTrial(optuna.trial.BaseTrial):
         if group is not None:
             self._group: "ProcessGroup" = group
         else:
-            if _g_pg[0] is None:
+            if _g_pg is None:
                 if dist.group.WORLD is None:
                     raise RuntimeError("torch distributed is not initialized.")
                 default_pg: "ProcessGroup" = dist.group.WORLD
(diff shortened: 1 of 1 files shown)


Alnusjaponica:
I confirmed that this PR fixes mypy error in `Checks (integration)` together with #4410.  
See [here](https://github.com/Alnusjaponica/optuna/pull/3) to confirm.

toshihikoyanase:
@cross32768 Could you review this PR since it follows up #4281.

It will fix the error in https://github.com/optuna/optuna/actions/runs/4139175958/jobs/7156458918.
> optuna/integration/pytorch_distributed.py:137: error: Incompatible types in assignment (expression has type "Optional[ProcessGroup]", variable has type "ProcessGroup")  [assignment]

optuna/integration/pytorch_distributed.py
```diff
@@ -36,7 +35,7 @@
     "Use :func:`~optuna.integration.TorchDistributedTrial.suggest_float` instead."
 )
 
-_g_pg: List[Optional["ProcessGroup"]] = [None]
+_g_pg: Optional["ProcessGroup"] = None
```
toshihikoyanase, line 38: This change seems reasonable to me since only the first element is used in the `master` code.
I'd like to use a local variable to manage the process group for Optuna in https://github.com/optuna/optuna/pull/4268, but I don't think we should work on it in this PR.

cross32768 (approved):
I checked this change in my local environment and confirmed that this change fix the mypy error. It looks reasonable fix. LGTM!

optuna/integration/pytorch_distributed.py
Alnusjaponica: Thank you for your follow-up.

toshihikoyanase (approved):
I also confirmed that the change worked with the pytorch distributed example in `oputna/optuna-examples`. LGTM!

PR #4414, opened 2023-02-09
Sort studies by study_id

c-bata:
## Motivation

@okaikov reported that optuna-dashboard shows studies in inconsistent orders.
https://github.com/optuna/optuna-dashboard/discussions/332#discussioncomment-4666534

I can also confirm  when using Optuna v3.1.0 with MySQL.

## Description of the changes

Sort studies by study_id.


contramundum53:
Should the trials be sorted by trial id or trial number?
I thought trial ids could be anything (for example a hash), as long as they are unique.

c-bata:
Thank you for your review!

> Should the trials be sorted by trial id or trial number? I thought trial ids could be anything (for example a hash), as long as they are unique.

I think trials have been sorted by trial_id at the following line:

https://github.com/optuna/optuna/blob/e68a89890c1e388066ea30a63533a68056fbfba3/optuna/storages/_rdb/storage.py#L823

contramundum53 (approved):
LGTM!

c-bata:
@amylase Could you review this PR if you have time?

amylase (commented):
The code change looks okay. 
By the way, why do we want to sort this in `RDBStorage`? Seems like `BaseStorage.get_all_studies` does not guarantee the order of return value (afaik from its documentation). 

If there is an important usage of this method that depends on this behavior, we should ensure it in all storages and update the documentation of `BaseStorage.get_all_studies`.

amylase (approved):
LGTM

c-bata:
@contramundum53 Could you merge this PR since it has two approvals.

PR #4416, opened 2023-02-10
Validate length of values in `add_trial`

not522:
## Motivation
Currently, `add_trial` does not check the length of trial's `values`. So, the study accepts an invalid trial.

```python
import optuna
study = optuna.create_study()
study.add_trial(optuna.create_trial(values=[0, 1]))  # No error!
```

## Description of the changes
- Raise `ValueError` if the length of `values` is invalid

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -898,8 +898,7 @@ class Study:
 
         if trial.values is not None and len(self.directions) != len(trial.values):
             raise ValueError(
-                f"Although the number of objectives is {len(self.directions)}, "
-                f"the trial has {len(trial.values)} length values."
+                f"The added trial has {len(trial.values)} values, which is different from the number of objectives {len(self.directions)} in the study (determined by Study.directions)."
             )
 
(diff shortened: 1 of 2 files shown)


Alnusjaponica (approved):
LGTM

optuna/study/study.py
```diff
@@ -896,6 +896,12 @@ def objective(trial):
 
         trial._validate()
 
+        if trial.values is not None and len(self.directions) != len(trial.values):
+            raise ValueError(
+                f"Although the number of objectives is {len(self.directions)}, "
+                f"the trial has {len(trial.values)} length values."
```
contramundum53, line 902:
```suggestion
                f"The added trial has {len(trial.values)} values, which is different from the number of objectives {len(self.directions)} in the study (determined by Study.directions)."
```
Nit: How about making the error message like this?

contramundum53 (approved):
LGTM.

contramundum53 (approved):

PR #4402, opened 2023-02-07
Fix test_with_server.py

contramundum53:
Currently, `test_multiprocess_run_optimize` function copies `RDBStorage` object (and thus its underlying sqlalchemy object) to subprocesses. This causes probabilistic and platform-dependent hangs in CI. We fix this by calling `get_storage` function in each subprocess.

PR #4422, opened 2023-02-15
Fix RDBStorage.get_best_trial when there are `inf`s

### LINKED ISSUE #4420: Objective values of +/-inf not handled correctly with PostgreSQL
### Expected behavior

Hi Optuna devs, thanks for all the hardwork on an amazing library! We're trying to put Optuna to good use at a big corp to optimise our industrial processes. Recently we've hit an issue with distributed optimisation jobs.

When performing an optimisation which attempts to maximise/minimise an objective function which returns -/+inf in some cases, I expect the optimisation process to be robust against the presence of possible inf values. In the case that I choose to maximise and some trials return -inf whilst others return finite values, I expect the optimum trial to be that which returned the largest finite value; instead, the first trial returning -inf is found to be the optimum. The same occurs when minimising and the objective sometimes returns +inf.

### Environment

- Optuna version: 3.1.0
- Python version: 3.9.16
- OS: Ubuntu 20.04
  - PostgreSQL 11

```shell
[I 2023-02-14 16:20:48,737] A new study created in RDB with name: optuna-inf-bug-87376ba1-5ca6-4da4-bf7f-7050ea2d6df6
[I 2023-02-14 16:20:51,302] Trial 0 finished with value: 1.0 and parameters: {}. Best is trial 0 with value: 1.0.
[I 2023-02-14 16:20:53,541] Trial 1 finished with value: 2.0 and parameters: {}. Best is trial 1 with value: 2.0.
[I 2023-02-14 16:20:56,011] Trial 2 finished with value: 3.0 and parameters: {}. Best is trial 2 with value: 3.0.
[I 2023-02-14 16:20:58,538] Trial 3 finished with value: -inf and parameters: {}. Best is trial 3 with value: -inf.
[I 2023-02-14 16:21:00,780] Trial 4 finished with value: 5.0 and parameters: {}. Best is trial 3 with value: -inf.
```

### Steps to reproduce

1. Set `DATABASE_URL` in the environment to securely connect to a PostgreSQL database configured for use with Optuna.
2. Run the below script.
3. Observe spurious reported best trial info in the Optuna output.
```python
#!/usr/bin/env python3
"""Minimal reproducible example for Optuna PostgreSQL storage inf values bug."""

from collections import deque
import os
import typing as _t
from uuid import uuid4

import numpy as np
import optuna


# Values of np.inf are stored with TrialValueModel.value=null and value_type=INF_NEG causing incorrect min/max values.
DATA: _t.Deque[float] = deque([1.0, 2.0, 3.0, -np.inf, 5.0])
# Using largest system float instead is a workaround.
# DATA: _t.Deque[float] = deque([1.0, 2.0, 3.0, np.finfo(np.float64).min, 5.0])


    return DATA.popleft()


def main() -> None:
    study: optuna.Study = optuna.create_study(
        study_name=f"optuna-inf-bug-{uuid4()}",
        sampler=optuna.samplers.TPESampler(),
        storage=os.getenv("DATABASE_URL"),
        direction="maximize",
    )
    study.optimize(objective, n_trials=len(DATA), n_jobs=1)


if __name__ == "__main__":
    main()
```

From poking around in the debugger and looking at the data in my PostgreSQL instance I have found the root cause. When attempting to retrieve the best trial from the database in [`optuna.storages._rdb.models.TrialModel.find_max_value_trial`](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/models.py#L182), there is an ORDER BY DESC TrialValueModel.value followed by LIMIT 1. This query does not account for TrialValueModel.value being stored as null in the database as it is for +/-inf. These null values appear at the top in the ordered query results and so get returned as optimal by the LIMIT statement.

Some thoughts... `order_by(nullslast(desc(TrialValueModel.value)))` could be used but this would preclude the possibility that +/-inf values constitute part of an optimal group. Results could be loaded into local memory without the LIMIT statement, converted to their python value with `TrialValueModel.stored_repr_to_value`, then passed to `max/min`; however, this could involve very large query sets being transferred to the host which is not desireable. I'd be happy to put in a PR for a fix but I'm not sure what the best approach would be here..

contramundum53:
## Motivation
Fix #4420.

## Description of the changes
Add `ORDER BY` clause based on the value of `value_type`.

### DIFF
--- a/tests/storages_tests/test_storages.py
+++ b/tests/storages_tests/test_storages.py
@@ -871,10 +871,10 @@ def test_get_n_trials_state_option(storage_mode: str) -> None:
     [
         [0.0, 1.0, 2.0],
         [0.0, float("inf"), 1.0],
-        [0.0, -float("inf"), 1.0],
-        [float("inf"), 0.0, 1.0, -float("inf")],
+        [0.0, float("-inf"), 1.0],
+        [float("inf"), 0.0, 1.0, float("-inf")],
         [float("inf")],
-        [-float("inf")],
+        [float("-inf")],
     ],
 )
 def test_get_best_trial(storage_mode: str, direction: StudyDirection, values: List[float]) -> None:
--- a/tests/storages_tests/test_storages.py
+++ b/tests/storages_tests/test_storages.py
@@ -865,9 +865,21 @@ def test_get_n_trials_state_option(storage_mode: str) -> None:
 
 
 @pytest.mark.parametrize("storage_mode", STORAGE_MODES)
-def test_get_best_trial(storage_mode: str) -> None:
+@pytest.mark.parametrize("direction", [StudyDirection.MAXIMIZE, StudyDirection.MINIMIZE])
+@pytest.mark.parametrize(
+    "values",
+    [
+        [0.0, 1.0, 2.0],
+        [0.0, float("inf"), 1.0],
+        [0.0, -float("inf"), 1.0],
+        [float("inf"), 0.0, 1.0, -float("inf")],
+        [float("inf")],
+        [-float("inf")],
+    ],
+)
+def test_get_best_trial(storage_mode: str, direction: StudyDirection, values: List[float]) -> None:
     with StorageSupplier(storage_mode) as storage:
-        study_id = storage.create_new_study(directions=[StudyDirection.MAXIMIZE])
+        study_id = storage.create_new_study(directions=[direction])
         with pytest.raises(ValueError):
             storage.get_best_trial(study_id)
 
@@ -875,12 +887,14 @@ def test_get_best_trial(storage_mode: str) -> None:
             storage.get_best_trial(study_id + 1)
 
         generator = random.Random(51)
-        for i in range(3):
+
+        for v in values:
             template_trial = _generate_trial(generator)
             template_trial.state = TrialState.COMPLETE
-            template_trial.value = float(i)
+            template_trial.value = v
             storage.create_new_trial(study_id, template_trial=template_trial)
-        assert storage.get_best_trial(study_id).number == i
+        expected_value = max(values) if direction == StudyDirection.MAXIMIZE else min(values)
+        assert storage.get_best_trial(study_id).value == expected_value
 
 
 def test_get_trials_excluded_trial_ids() -> None:
--- a/optuna/storages/_rdb/models.py
+++ b/optuna/storages/_rdb/models.py
@@ -6,6 +6,7 @@ from typing import Optional
 from typing import Tuple
 
 from sqlalchemy import asc
+from sqlalchemy import case
 from sqlalchemy import CheckConstraint
 from sqlalchemy import Column
 from sqlalchemy import DateTime
@@ -19,7 +20,6 @@ from sqlalchemy import orm
 from sqlalchemy import String
 from sqlalchemy import Text
 from sqlalchemy import UniqueConstraint
-from sqlalchemy import case
 
 from optuna import distributions
 from optuna.study._study_direction import StudyDirection
@@ -192,8 +192,8 @@ class TrialModel(BaseModel):
             .order_by(
                 desc(
                     case(
+                        {"INF_NEG": -1, "FINITE": 0, "INF_POS": 1},
                         value=TrialValueModel.value_type,
-                        whens={"INF_NEG": -1, "FINITE": 0, "INF_POS": 1},
                     )
                 ),
                 desc(TrialValueModel.value),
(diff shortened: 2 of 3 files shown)


c-bata:

chrisk314:
Might make sense if the values of [`TrialValueType`](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/models.py#L387-L390) enum were ordered on the real line, then these values could be reused directly in the proposed `case` statement. I'm not familiar with the code so don't know what the consequences of that change would be elsewhere. Possibly out of scope of this issue/PR. Just a thought.

contramundum53:
@chrisk314 Actually the enum seems to be treated like a string in the database, so I think `CASE` clause would be needed anyways.

chrisk314:
@contramundum53 yes, these are indeed stored as strings in the database. I can confirm that from investigating yesterday. The `CASE` statement is definitely needed. What I had in mind was something like the below. The benefit of this is that the `TrialValueModel` is the sole owner of the definition of the enum values, whereas now they're technically appearing as magic strings.

```python
class TrialValueType(Enum):
    INF_NEG = -1
    FINITE = 0
    INF_POS = 1
    
# This mapping can then be used in the `case` statement.
{e.name: e.value for e in TrialValueType}
```

c-bata (commented):
@contramundum53 Thank you for the pull request. Changes almost look good to me. Can we also add a test case in `test_with_server.py` to check this change with PostgreSQL and MySQL?

contramundum53:
That would produce a conflict with #4402. I'll work on it after #4402 is merged.

c-bata:
I've just merged #4402 👍

tests/storages_tests/test_storages.py
```diff
@@ -865,22 +865,36 @@ def test_get_n_trials_state_option(storage_mode: str) -> None:
 
 
 @pytest.mark.parametrize("storage_mode", STORAGE_MODES)
-def test_get_best_trial(storage_mode: str) -> None:
+@pytest.mark.parametrize("direction", [StudyDirection.MAXIMIZE, StudyDirection.MINIMIZE])
+@pytest.mark.parametrize(
+    "values",
+    [
+        [0.0, 1.0, 2.0],
+        [0.0, float("inf"), 1.0],
```
amylase, line 873: nit: could you write `float("-inf")` for better readability? the current notation requires extra attention because there is a gap between `-` and `inf`.

amylase (approved):
Looks good. I confirmed that all necessary code change is made. Added a small comment.

c-bata:
@contramundum53 Let me kindly remind you.

contramundum53:
Sorry for being late. I just updated the tests.

c-bata (approved):
I confirmed that the valid queries are emitted in MySQL, PostgreSQL, and SQLite3. LGTM.

**MySQL**

```
(Pdb) !from sqlalchemy.dialects import mysql
(Pdb) p print(query.statement.compile(dialect=mysql.dialect(),  compile_kwargs={"literal_binds": True}))
SELECT trials.trial_id, trials.number, trials.study_id, trials.state, trials.datetime_start, trials.datetime_complete
FROM trials INNER JOIN trial_values ON trials.trial_id = trial_values.trial_id
WHERE trials.study_id = 7 AND trials.state = 'COMPLETE' AND trial_values.objective = 0 ORDER BY CASE trial_values.value_type WHEN 'INF_NEG' THEN -1 WHEN 'FINITE' THEN 0 WHEN 'INF_POS' THEN 1 END ASC, trial_values.value ASC
 LIMIT 1
None
```

https://dev.mysql.com/doc/refman/5.7/en/case.html

**SQLite3**

```
(Pdb) !from sqlalchemy.dialects import sqlite
(Pdb) p print(query.statement.compile(dialect=sqlite.dialect(),  compile_kwargs={"literal_binds": True}))
SELECT trials.trial_id, trials.number, trials.study_id, trials.state, trials.datetime_start, trials.datetime_complete
FROM trials JOIN trial_values ON trials.trial_id = trial_values.trial_id
WHERE trials.study_id = 7 AND trials.state = 'COMPLETE' AND trial_values.objective = 0 ORDER BY CASE trial_values.value_type WHEN 'INF_NEG' THEN -1 WHEN 'FINITE' THEN 0 WHEN 'INF_POS' THEN 1 END ASC, trial_values.value ASC
 LIMIT 1 OFFSET 0
None
```

https://www.sqlite.org/lang_expr.html


**PostgreSQL**

```
(Pdb) !from sqlalchemy.dialects import postgresql
(Pdb) p print(query.statement.compile(dialect=postgresql.dialect(),  compile_kwargs={"literal_binds": True}))
SELECT trials.trial_id, trials.number, trials.study_id, trials.state, trials.datetime_start, trials.datetime_complete
FROM trials JOIN trial_values ON trials.trial_id = trial_values.trial_id
WHERE trials.study_id = 7 AND trials.state = 'COMPLETE' AND trial_values.objective = 0 ORDER BY CASE trial_values.value_type WHEN 'INF_NEG' THEN -1 WHEN 'FINITE' THEN 0 WHEN 'INF_POS' THEN 1 END ASC, trial_values.value ASC
 LIMIT 1
None
```

https://www.postgresql.org/docs/current/functions-conditional.html

PR #4425, opened 2023-02-16
Fix mac-test cache path

gen740:
## Motivation

mac-test's cache [did not work properly](https://github.com/optuna/optuna/actions/runs/4190237625/jobs/7263447625#step:12:2) due to setting the wrong path.
(ref. https://github.com/actions/cache/blob/main/examples.md#python---pip)

## Description of the changes

Fix it.


c-bata (approved):
LGTM!

PR #4427, opened 2023-02-16
Add rank plot

contramundum53:
## Motivation
https://github.com/optuna/optuna-dashboard/issues/384

## Description of the changes
Add `plot_rank` function.
This plots the parameter relationship as pairwise scatter plots with colors indicating the (normalized) rank of the objective value within the entire study.

image


c-bata:
@cross32768 Could you review this PR?

c-bata:
@hvy Could you review this PR?

hvy (changes requested):
Thanks for the PR.
I've mostly only went over the data preparation part but left some comments if you could take a look.
Also, could you add unit tests similar to our other visualization functions and the logic indeed has several conditionals, and add this function to the API reference under `docs/`?

contramundum53:
> did you consider add a thin border around the scatter dots as in plot_slice? Applying it here as well might be consistent and perhaps make it easier to comprehend? 🤔
I tried it but it produced no visible difference to me.

contramundum53:
@hvy Thanks for your comments. I fixed those points, could you take another look?

hvy:
>> did you consider add a thin border around the scatter dots as in plot_slice? Applying it here as well might be consistent and perhaps make it easier to comprehend? 🤔

> I tried it but it produced no visible difference to me. I can add it to be consistent.

Hm, I noticed a visual difference with ` "line":{"width": 0.5, "color": "Grey"}`. What do you say, should we apply it?

.github/workflows/tests-with-minimum-versions.yml
```diff
@@ -71,7 +71,7 @@ jobs:
         # Install dependencies with minimum versions.
         pip uninstall -y alembic cmaes packaging sqlalchemy plotly scikit-learn
         pip install alembic==1.5.0 cmaes==0.9.1 packaging==20.0 sqlalchemy==1.3.0 numpy==1.20.3 tqdm==4.27.0 colorlog==0.3 PyYAML==5.1
-        pip install plotly==4.9.0 scikit-learn==0.24.2  # optional extras
+        pip install plotly==5.0.0 scikit-learn==0.24.2  # optional extras
```
hvy, line 74: should we check this with the other maintainers? also, should we not update `pyproject.toml` ?

contramundum53: Updating `pyproject.toml` would prevent `plotly-4.9.0` users from using other visualization tools or accessing other optuna updates, so I don't think that's desirable.
For whether to depend on the newer version of plotly or reimplment the logic ourselves, I think we could have a discussion. What do you think, @toshihikoyanase?

contramundum53: I had a discussion with @toshihikoyanase . We agreed that
* Updating the required dependency would be undesirable.
* Since this functionality is experimental, we could add a nice assertion message on `plotly` version and reimplement the logic if users really need it.

contramundum53:
> Hm, I noticed a visual difference with  "line":{"width": 0.5, "color": "Grey"}. What do you say, should we apply it?
Could you show me some images? I'm not sure if I'm reproducing it right.

.github/workflows/tests-with-minimum-versions.yml
toshihikoyanase: I think we may want to memorize why plotly versions are inconsistent between `pyproject.toml` and this configuration. Someone may try to fix the inconsistency. How about leaving a comment on it?

contramundum53:

optuna/visualization/_rank.py
```diff
@@ -0,0 +1,393 @@
...
+    return _AxisInfo(
+        name=param_name,
+        range=(min_value, max_value),
+        is_log=is_log,
+        is_cat=is_cat,
+    )
+
+
+def _get_rank_subplot(
+    info: _RankSubplotInfo, target_name: str, print_all_objectives: bool
```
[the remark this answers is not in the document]
contramundum53: I changed the name from `print_all_objectives` to `print_raw_objectives`. This describes what the argument does.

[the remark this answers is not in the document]
cross32768: I got it. It looks better

cross32768:
Thank you for responding to the review. LGTM!

cross32768 (approved):

.github/workflows/tests-with-minimum-versions.yml
contramundum53: I added a comment.

optuna/visualization/_rank.py
```diff
@@ -0,0 +1,396 @@
...
+        range=(min_value, max_value),
+        is_log=is_log,
+        is_cat=is_cat,
+    )
+
+
+def _get_rank_subplot(
+    info: _RankSubplotInfo, target_name: str, print_raw_objectives: bool
+) -> "Scatter":
+    colormap = "RdYlBu_r"
```
not522, line 276: How about using `RdYlBu` for maximizing?

optuna/visualization/_rank.py
```diff
@@ -0,0 +1,396 @@
...
+        x=[None],
+        y=[None],
+        mode="markers",
+        marker=dict(
+            line={"width": 0.5, "color": "Grey"},
+            colorscale=colormap,
+            showscale=True,
+            cmin=0,
+            cmax=1,
+            colorbar=dict(thickness=10, tickvals=tick_coloridxs, ticktext=ticktext),
```
not522, line 390: How about adding the title?

```suggestion
            colorbar=dict(thickness=10, tickvals=tick_coloridxs, ticktext=ticktext, title=f"Rank ({info.target_name})"),
```

optuna/visualization/_rank.py
contramundum53: I think using `RdYlBu_r` for both maximizing and minimizing would be more intuitive, because higher values correspond to red and lower values correspond to blue (which meets our intuition as temperature).
In my opinion, it would be somewhat counterintuitive if it is reversed.
Also, we don't know whether the target is maximized or minimized when a custom `target` is specified.

optuna/visualization/_rank.py
contramundum53: image

In my opinion, this looks a bit too redundant and noisy.

not522: Hmmm, but without that, isn't it hard to understand what the numbers mean?

contramundum53: What about changing the title of the plot to `Rank (Objective Value)`?

not522: That seems reasonable!

optuna/visualization/_rank.py
not522: OK, let's leave it in this PR and discuss it later.

optuna/visualization/_rank.py
contramundum53: @not522 PTAL.

not522 (approved):
LGTM.
@hvy Have your concerns been resolved? If so, please merge it.

hvy (approved):
LGTM!

PR #4431, opened 2023-02-17
Sync owned trials when calling `study.ask` and `study.get_trials`

c-bata:
## Motivation

`read_trials_from_remote_storage` method in `_CachedStorage` does not sync trials owned by itself. This behavior causes problems when another worker (e.g. Optuna Dashboard) make the trial completed as follows:
https://gist.github.com/c-bata/69e81a34c395d3ea368e24b0d5e0d367

## Description of the changes

This PR changes `excluded_trial_ids` from `owned_or_finished_trial_ids` to `finished_trial_ids` only.


tests/study_tests/test_study.py
```diff
@@ -1502,7 +1502,7 @@ def test_tell_from_another_process() -> None:
         assert study.best_value == 1.2
 
         # Should fail because the trial0 is already finished.
-        with pytest.raises(RuntimeError):
+        with pytest.raises(ValueError):
```
c-bata, line 1505: FYI: Previously, `check_trial_is_updatable()` method in `optuna/storages/_rdb/storage.py` raises `RuntimeError`. However, from this change, `optuna/study/_tell.py"` raises `ValueError` like the following CI output.
https://github.com/optuna/optuna/actions/runs/4200669694/jobs/7286925478

c-bata:
@amylase @contramundum53 Could you review this PR?

optuna/storages/_cached_storage.py
```diff
@@ -344,19 +349,21 @@ def get_all_trials(
             trials = list(sorted(trials.values(), key=lambda t: t.number))
             return copy.deepcopy(trials) if deepcopy else trials
 
-    def read_trials_from_remote_storage(self, study_id: int) -> None:
+    def read_trials_from_remote_storage(self, study_id: int, sync_owned_trials: bool) -> None:
```
amylase, line 352: Did you consider to implement two separate methods instead of having a boolean parameter?

optuna/storages/_cached_storage.py
```diff
@@ -328,7 +333,7 @@ def get_all_trials(
         states: Optional[Container[TrialState]] = None,
     ) -> List[FrozenTrial]:
         if study_id not in self._studies:
-            self.read_trials_from_remote_storage(study_id)
+            self.read_trials_from_remote_storage(study_id, sync_owned_trials=False)
```
amylase, line 336: Why can we skip syncing owned trials here? If external processes can access & update them, I think we should always sync them.

amylase (commented):
Also, could you update the section of `_CachedStorage`'s docstring that mentions `read_trials_from_remote_storage`?

https://github.com/optuna/optuna/pull/4431/files#diff-8ce5c3176c6b5fa3a21ed11da3f19d07c7c0291b7b7b6b67221b057855b7ac50R55

optuna/storages/_cached_storage.py
```diff
@@ -183,7 +188,7 @@ def create_new_trial(self, study_id: int, template_trial: Optional[FrozenTrial]
             # WAITING trials are exception and they can be modified from arbitral worker.
             # Thus, we cannot add them to a list of cached trials.
             if frozen_trial.state != TrialState.WAITING:
-                study.owned_or_finished_trial_ids.add(frozen_trial._trial_id)
+                study.owned_trial_ids.add(frozen_trial._trial_id)
```
c-bata, line 191: @contramundum53 pointed out that we need to add following conditions.

```python
if frozen_trial.state.is_finished():
    study.finished_trial_ids.add(frozen_trial._trial_id)
```

optuna/storages/_cached_storage.py
```diff
@@ -194,39 +184,6 @@ def set_trial_param(
...
-
-                dists = copy.copy(cached_trial.distributions)
-                dists[param_name] = distribution
-                cached_trial.distributions = dists
-
-                if cached_dist:  # Already persisted in case of cache miss so no need to update.
-                    self._backend.set_trial_param(
-                        trial_id, param_name, param_value_internal, distribution
-                    )
-                return
```
c-bata, line 228: This logic updates the cached owned trials. However, from this PR, owned trials are always synced, so this logic is no longer needed.

c-bata:
FYI: I ran the following benchmark to compare how CachedStorage slows down with this change.

https://gist.github.com/c-bata/16d654471393f59e064482ffb8b52bf8

|  | master | This PR |
| --- | --- | --- |
| RDBStorage | 47.9438s | 51.0639s |
| CachedStorage(RDBStorage) | 36.6144s | 54.9168s |

c-bata:
Let me unassign reviewers since the performance become slower than I expected. I will mark this ready for review after fixing the performance issue.

PR #4436, opened 2023-02-17
Implement NSGA-III

Alnusjaponica:
## Motivation
Implement [NSGA-III](https://ieeexplore.ieee.org/document/6600851) algorithm, which is a modified version of [NSGA-II](https://ieeexplore.ieee.org/document/996017) for many-objective optimization problem. This algorithm is also desired at  https://github.com/optuna/optuna/discussions/4075.

## Description of the changes
Implement `NSGAIIISampler` which realizes NSGA-III algorithm. The main change in the algorithms is the strategy used to determine who survives from the boundary front, where NSGA-II uses crowding distance while NSGA-III uses reference points.
The difference between NSGA-II and NSGA-III in implementation is the following two points:

- `NSGAIIISampler` has attribute `reference_points` used in `_select_elite_population()`
- `_crowding_distance_sort()` called from `_select_elite_population()` is replaced with new procedure proposed in NSGA-III algorithm.

Other part of `nsgaiii.py` is almost the same as `nsgaii/_sampler.py`.

As confirmed in the following benchmark results on test problems, NSGA-III is competitive enough compared to NSGA-II in terms of hyper volume metric.
WFG6_M_3
WFG6_M_5
WFG6_M_8

 In addition, the results of NSGA-III tends to be more diverse as observed in the following plots. 

Figure_1

Figure_2

Figure_3

This is because the crowding distance is set to infinite when an individual has the largest or smallest value on any of the objective axes, and such individuals always survive. Therefore, the resulting points of NSGA-II tend to concentrate near the axes (i.e., take extreme values) in situations where the dimension is large and the first front often becomes the boundary rank front.


Alnusjaponica:
The detailed description of the experiment is as follows:

1. How hypervolume is improved as number of trials is increased
    This experiment 
    - compares: NSGA-II, NSGA-III with dividing parameter p=5,8
    - runs 5000 trials 5 times for each setting
    - uses WFG6 with objective dimension 3, 5, 8
    - determines reference point for hypervolume as the nadir point of best_trials at each step (number of trials = 500,1000,...5000)

2. What the geometrical feature of the results is like
This experiment    
    - compares: NSGA-II, NSGA-III with dividing parameter p=5
    - run 3000 trials and extract best trials among all 6000 trials
    - uses DTLZ1 with objective dimension 9
    - plot the objective values by 3 dimension

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,542 @@
...
+from optuna.trial import TrialState
+
+
+# Define key names of `Trial.system_attrs`.
+_GENERATION_KEY = "nsga3:generation"
+_POPULATION_CACHE_KEY_PREFIX = "nsga3:population"
+
+
+class NSGAIIISampler(BaseSampler):
+    """Multi-objective sampler using the NSGA-III algorithm.
```
toshihikoyanase, line 41: The reference for `NSGAIIISampler` was not rendered. I think this is because it is not listed in https://github.com/optuna/optuna/blob/master/docs/source/reference/samplers/index.rst. Could you add `NSGAIIISampler` to the target of `autosummary `?

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,542 @@
...
+            After non-dominated sort, who out of borderline front are going to survived is
+            determined according to how sparse the closest reference point of each person is.
+            In the default setting the algorithm uses `uniformly` spread points to diversify the
+            result.
+
+            .. note::
+                Other parameters are the same as :class:`~optuna.samplers.nsgaii.NSGAIISampler`.
+
+
+    """
```
toshihikoyanase, line 68: Maybe we have too many blank lines.
```suggestion
    """
```

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,543 @@
...
+    def _select_elite_population(
+        self, study: Study, population: List[FrozenTrial]
+    ) -> List[FrozenTrial]:
+        elite_population: List[FrozenTrial] = []
+        population_per_rank = self._fast_non_dominated_sort(population, study.directions)
+        for population in population_per_rank:
+            if len(elite_population) + len(population) < self._population_size:
+                elite_population.extend(population)
+            else:
+                objective_matrix = _normalize(elite_population, population)
```
HideakiImamura, line 297: I think the name of the functions `_normalize`, `_associate`, and `_niching` are too general. Could you elaborate the name of the functions for code readability?

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,543 @@
...
+        if self._constraints_func is not None:
+            _process_constraints_after_trial(self._constraints_func, study, trial, state)
+        self._random_sampler.after_trial(study, trial, state, values)
+
+
+def multi_choose(n: int, k: int) -> int:
+    return math.factorial(n + k - 1) // math.factorial(k) // math.factorial(n - 1)
+
+
+def generate_default_reference_point(
```
HideakiImamura, line 382: It seems that this function is only used in the unit tests. Do you intend to call this function in the `__init__` of the `NSGAIIISampler`?

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,543 @@
...
+        (
+            multi_choose(objective_dimension, dividing_parameter),
+            objective_dimension,
+        )
+    )
+    for i, comb in enumerate(
+        itertools.combinations_with_replacement(range(objective_dimension), dividing_parameter)
+    ):
+        for j in comb:
+            reference_points[i, j] += 1.0
```
HideakiImamura, line 396: This is just a question. Does the behavior of the algorithm depend on the scale of the reference points, i.e., does the behavior of the algorithm change if we change this value of `1.0` to `10000.`?

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,543 @@
...
+    elite_population: List[FrozenTrial],
+    population: List[FrozenTrial],
+    weights: Optional[np.ndarray] = None,
+    method: str = "original",
+) -> np.ndarray:
+    """Normalizes objective values of population
+
+    An ideal point z* consists of minimums in each axis. Each objective value of population is
+    then subtracted by the ideal point.
+    An extreme point of each axis is (originally) defined as a minimum solution of ASF from the
```
HideakiImamura, line 410: What is "ASF"?

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,543 @@
...
+
+    # Collect objective values
+    objective_dimension = len(population[0].values)
+    objective_matrix = np.zeros((len(elite_population + population), objective_dimension))
+    for i, trial in enumerate(elite_population + population):
+        objective_matrix[i] = np.array(trial.values, dtype=float)
+
+    # Subtract ideal point
+    objective_matrix -= np.min(objective_matrix, axis=0)
+    objective_matrix[objective_matrix < 10e-3] = 0.0
```
HideakiImamura, line 424: Why this mask is needed?

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,543 @@
...
+        for j in comb:
+            reference_points[i, j] += 1.0
+    return reference_points
+
+
+def _normalize(
+    elite_population: List[FrozenTrial],
+    population: List[FrozenTrial],
+    weights: Optional[np.ndarray] = None,
+    method: str = "original",
```
HideakiImamura, line 404: Logics for this `weights` and `method` are dead codes (not used and not tested). Should we leave these arguments on this PR? (I think it's OK to leave them as they are, as long as we work on propagating them to the constructor ASAP after this PR.)

HideakiImamura (commented):
Thanks for the PR. I have some early comments and questions. I will review the details of `_normalize`,`_associate`, and `_niching` in the next week.

optuna/samplers/nsgaiii.py
Alnusjaponica: Users can use arbitrary point set as a reference point set according to their preference and give them `NSGAIIISampler`. Since `generate_default_reference_point` only provides reference point sets in default setting, I thought it will be easier to separate from `__init__`. Is it better to move?

optuna/samplers/nsgaiii.py
Alnusjaponica: My understanding is that it doesn't. This is because the algorithm considers the distance between objective values and the reference lines, which　connect the origin and each reference point.

optuna/samplers/nsgaiii.py
Alnusjaponica: This is short for "achievement scalarizing function." The definition is in the original paper and is basically the maximum entry of weighted objective value.

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,543 @@
...
+    Nondominated Sorting Approach, Part II: Handling Constraints and Extending to an Adaptive
+    Approach <https://ieeexplore.ieee.org/document/6595567>`_
+
+    Args:
+        reference_points:
+            A 2 dimension ``numpy.ndarray`` with objective dimension columns. Represents
+            a list of reference points which is used to determine who to survive.
+            After non-dominated sort, who out of borderline front are going to survived is
+            determined according to how sparse the closest reference point of each person is.
+            In the default setting the algorithm uses `uniformly` spread points to diversify the
```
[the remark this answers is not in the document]
Alnusjaponica: It is also possible to reflect user's preferences by giving custamized set of target points since the algorithm prioritizes individuals around reference points.

optuna/samplers/nsgaiii.py
Alnusjaponica: This is a kind of stabilization adopted from pymoo. I'll removed it until we realize why this mask is necesssary.

optuna/samplers/nsgaiii.py
Alnusjaponica: I romoved the options for now.

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,554 @@
...
+        individual_idx = rng.choice(len(nearest_points_count_to_reference_point[count]))
+        (
+            nearest_points_count_to_reference_point[count][individual_idx],
+            nearest_points_count_to_reference_point[count][-1],
+        ) = (
+            nearest_points_count_to_reference_point[count][-1],
+            nearest_points_count_to_reference_point[count][individual_idx],
+        )
+
+        reference_point_id = nearest_points_count_to_reference_point[count].pop()
```
HideakiImamura, line 533: I think this section is for randomly selecting one element from nearest_points_count_to_reference_point[count]. Since the code is a bit long and hard to read, how about shuffling nearest_points_count_to_reference_point[count] randomly at first and then popping from the end?

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,554 @@
...
+                reference_point_to_population[reference_point_id][-1],
+            ) = (
+                reference_point_to_population[reference_point_id][-1],
+                reference_point_to_population[reference_point_id][reference_point_idx],
+            )
+        else:
+            # TODO(Shinichi) avoid sort
+            reference_point_to_population[reference_point_id].sort(reverse=True)
+
+        _, selected_individual_id = reference_point_to_population[reference_point_id].pop()
```
HideakiImamura, line 549: Ditto.

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,554 @@
...
+            nearest_points_count_to_reference_point[count][-1],
+        ) = (
+            nearest_points_count_to_reference_point[count][-1],
+            nearest_points_count_to_reference_point[count][individual_idx],
+        )
+
+        reference_point_id = nearest_points_count_to_reference_point[count].pop()
+        if count:
+            reference_point_idx = rng.choice(
+                len(reference_point_to_population[reference_point_id])
```
HideakiImamura, line 536: What happen if `reference_point_to_population[reference_point_id]` is an empty set?

HideakiImamura (commented):
Thanks for the update. I have some minor comments. PTAL.

optuna/samplers/nsgaiii.py
Alnusjaponica: Thank you for your comment. I am going to update the code accordingly.

optuna/samplers/nsgaiii.py
Alnusjaponica: `reference_point_to_population[reference_point_id]` won't be empty because
- `nearest_points_count_to_reference_point` only has reference points with non-empty neighbors list at first.
-  `reference_point_id` is pushed back to `nearest_point_count_to_reference_point` again after one of the neighbor is selected as a member elite population and deleted from the neighbors list only when the list is non-empty.

optuna/samplers/nsgaiii.py
Alnusjaponica: I updated the code accordingly.

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,621 @@
...
+
+                # TODO(Shinichi) reconsider alternative value for infs.
+                # Replace -inf with zero.
+                objective_matrix[objective_matrix < 0] = 0
+                # Replace inf with nadir value * 2.
+                mask_inf = np.isinf(objective_matrix)
+                objective_matrix[mask_inf] = np.nan
+                nadir_point = np.nanmax(objective_matrix, axis=0)
+                rows, cols = np.where(mask_inf)
+                objective_matrix[rows, cols] = nadir_point[cols] * 2
```
HideakiImamura, line 323: These logic should be in the `_normalize_objective_values`, I think.

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,621 @@
...
+        state: TrialState,
+        values: Sequence[float] | None,
+    ) -> None:
+        assert state in [TrialState.COMPLETE, TrialState.FAIL, TrialState.PRUNED]
+        if self._constraints_func is not None:
+            _process_constraints_after_trial(self._constraints_func, study, trial, state)
+        self._random_sampler.after_trial(study, trial, state, values)
+
+
+def multi_choose(n: int, k: int) -> int:
```
HideakiImamura, line 409: This function name should be started with `_`.

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,621 @@
...
+    return math.factorial(n + k - 1) // math.factorial(k) // math.factorial(n - 1)
+
+
+def generate_default_reference_point(n_objectives: int, dividing_parameter: int = 3) -> np.ndarray:
```
HideakiImamura, line 413: Ditto.

Alnusjaponica:
As we discussed offline, we've decided that `inf` and `-inf` are replaced with some large large number before normalization. For now, `COEF` times the length of the interval of the objective values in the axis is added to max for `inf` and subtracted from min for `-inf` in each objective dimension. 
@toshihikoyanase @HideakiImamura Could you take another look?

HideakiImamura:
Thanks for the update! I''' have another look.

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,634 @@
...
+            `target` points since the algorithm prioritizes individuals around reference points.
+
+        dividing_parameter:
+            An parameter to determine the density of default reference points. This parameter
+            determines how many divisions are made between reference points on each axis. The
+            smaller this value is, the less reference points you have. The default value is 3.
+            Note that this parameter is not used when reference_points is not None.
+
+            .. note::
+                Other parameters are the same as :class:`~optuna.samplers.nsgaii.NSGAIISampler`.
```
HideakiImamura, line 72: Might be good to unindent these lines.

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,634 @@
...
+            )
+
+        if population_size < crossover.n_parents:
+            raise ValueError(
+                f"Using {crossover},"
+                f" the population size should be greater than or equal to {crossover.n_parents}."
+                f" The specified `population_size` is {population_size}."
+            )
+
+        self.reference_points = reference_points
```
HideakiImamura, line 130: Is this attribute intended to be referenced by users? If not, how about starting the name from `_`?

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,634 @@
...
+        for population in population_per_rank:
+            if len(elite_population) + len(population) < self._population_size:
+                elite_population.extend(population)
+            else:
+                n_objectives = len(study.directions)
+                # Construct reference points in the first run.
+                if self.reference_points is None:
+                    self.reference_points = _generate_default_reference_point(
+                        n_objectives, self._dividing_parameter
+                    )
```
HideakiImamura, line 310: It would be great to construct the reference point in the `__init__`.

HideakiImamura (commented):
Almost, LGTM! I have some very minor comments. PTAL.

optuna/samplers/nsgaiii.py
Alnusjaponica: IIRC, we came to the conclusion that `n_objectives` is necessary to construct default reference points and it is generally unknown when constructing sampler.

optuna/samplers/nsgaiii.py
HideakiImamura: I see. My personal opinion is that if you want the user to be able to change this attribute, and they can, you should implement a setter function and a getter function. setter/getter implementation is not that much work, so you can use this PR Would you do it? By the way, in any case, I think it is better to start the attribute name itself with `_`.

optuna/samplers/nsgaiii.py
HideakiImamura: Thanks for the reply. I understood the reason. Let me resolve this comment.

optuna/samplers/nsgaiii.py
Alnusjaponica: Thank you for your advice. I privatized `reference_points` and defined a getter and a setter.

optuna/samplers/nsgaiii.py
Alnusjaponica: I updated the indent. Could you confirm the change is as what you expect by checking the complied document?

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,643 @@
...
+from optuna.study._multi_objective import _dominates
+from optuna.trial import FrozenTrial
+from optuna.trial import TrialState
+
+
+# Define key names of `Trial.system_attrs`.
+_GENERATION_KEY = "nsga3:generation"
+_POPULATION_CACHE_KEY_PREFIX = "nsga3:population"
+# Define a coefficient for scaling intervals, used in _filter_inf() to replace +-inf
+_COEF = 3
```
toshihikoyanase, line 36: [Notes] The variable name of `_COEF` might be too general, but the code comment explains the usage clearly. So, it's not mandatory, but we might rename it in a follow-up PR.

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,643 @@
...
+from optuna.study import Study
+from optuna.study._multi_objective import _dominates
+from optuna.trial import FrozenTrial
+from optuna.trial import TrialState
+
+
+# Define key names of `Trial.system_attrs`.
+_GENERATION_KEY = "nsga3:generation"
+_POPULATION_CACHE_KEY_PREFIX = "nsga3:population"
+# Define a coefficient for scaling intervals, used in _filter_inf() to replace +-inf
```
toshihikoyanase, line 35: A nit.
```suggestion
# Define a coefficient for scaling intervals, used in _filter_inf() to replace +-inf.
```

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,643 @@
...
+            `target` points since the algorithm prioritizes individuals around reference points.
+
+        dividing_parameter:
+            An parameter to determine the density of default reference points. This parameter
+            determines how many divisions are made between reference points on each axis. The
+            smaller this value is, the less reference points you have. The default value is 3.
+            Note that this parameter is not used when reference_points is not None.
+
+    .. note::
+        Other parameters than reference_points and dividing_parameter are the same as
```
toshihikoyanase, line 72: A nit, but `reference_points` and `dividing_parameter` should be shown as code.

image


```suggestion
        Other parameters than ``reference_points`` and ``dividing_parameter`` are the same as
```

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,643 @@
...
+            determined according to how sparse the closest reference point of each individual is.
+            In the default setting the algorithm uses `uniformly` spread points to diversify the
+            result. It is also possible to reflect your `preferences` by giving an arbitrary set of
+            `target` points since the algorithm prioritizes individuals around reference points.
+
+        dividing_parameter:
+            An parameter to determine the density of default reference points. This parameter
+            determines how many divisions are made between reference points on each axis. The
+            smaller this value is, the less reference points you have. The default value is 3.
+            Note that this parameter is not used when reference_points is not None.
```
toshihikoyanase, line 69:
```suggestion
            Note that this parameter is not used when ``reference_points`` is not :obj:`None`.
```

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,643 @@
...
+
+    .. note::
+        Other parameters than reference_points and dividing_parameter are the same as
+        :class:`~optuna.samplers.nsgaii.NSGAIISampler`.
+
+    """
```
toshihikoyanase, line 75: I'm not sure, but the `reference_points` is shown as a public attribute of `NSGAIIISampler`. I don't think it is necessary for users to read/write it.

image

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,643 @@
...
+        self,
+        population_size: int = 50,
+        reference_points: np.ndarray | None = None,
+        dividing_parameter: int = 3,
+        mutation_prob: float | None = None,
+        crossover: BaseCrossover | None = None,
+        crossover_prob: float = 0.9,
+        swapping_prob: float = 0.5,
+        seed: int | None = None,
+        constraints_func: Callable[[FrozenTrial], Sequence[float]] | None = None,
```
toshihikoyanase, line 87: How about following the argument order of `NSGAIISampler.__init__`?

optuna/samplers/nsgaiii.py
```diff
@@ -0,0 +1,643 @@
...
+
+
+# Define key names of `Trial.system_attrs`.
+_GENERATION_KEY = "nsga3:generation"
+_POPULATION_CACHE_KEY_PREFIX = "nsga3:population"
+# Define a coefficient for scaling intervals, used in _filter_inf() to replace +-inf
+_COEF = 3
+
+
+class NSGAIIISampler(BaseSampler):
```
toshihikoyanase, line 39: Could you add `experimental_class` decorator to `NSGAIIISampler`?

toshihikoyanase (commented):
Thank you for your update. Let me share some small comments.

optuna/samplers/nsgaiii.py
Alnusjaponica: I see. I sorted the argument as in `NSGAIISampler.__init__()`

optuna/samplers/nsgaiii.py
Alnusjaponica: I added the decorator. Please confirm that I noted the correct version.

optuna/samplers/nsgaiii.py
Alnusjaponica: That is because I added getter and setter for `_reference_points` as in the discussion here. https://github.com/optuna/optuna/pull/4436#discussion_r1173189869

Alnusjaponica: As we discussed offline, I made `_reference_points` completely private.

HideakiImamura (approved):
Thanks for your long running. LGTM!

toshihikoyanase:
I found that the module of `optuna.samplers.nsgaii` was public. I think users import `NSGAIIISampler` like `from optuna.samplers import NSGAIIISampler` like the other samplers, and `optuna.samplers.nsgaii` does not provide any public classes/functions excepting for `NSGAIIISampler`. So, I guess we can rename `nsgaiii.py` to `_nsgaiii.py`.

Alnusjaponica:
Thanks for your comment. I renamed the module and changed the import paths.

toshihikoyanase (commented):
Thank you for your update. I found two functions which are not used in the test cases. Could you check them?

tests/samplers_tests/test_nsgaiii.py
```diff
@@ -0,0 +1,613 @@
...
+    trial = optuna.trial.create_trial(
+        state=optuna.trial.TrialState.COMPLETE,
+        values=list(values),
+        system_attrs={} if constraints is None else {_CONSTRAINTS_KEY: list(constraints)},
+    )
+    trial.number = number
+    trial._trial_id = number
+    return trial
+
+
```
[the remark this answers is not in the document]
Alnusjaponica: Thank you for pointing that out. I removed those unused utility functions.

optuna/samplers/_nsgaiii.py
```diff
@@ -0,0 +1,637 @@
...
+            A ``numpy.ndarray`` with rows of generation size. Each row is the index of the closest
+            reference point to the corresponding individual.
+        rng:
+            Random number generator.
+
+    Returns:
+        A list of trials which are selected as the next generation.
+    """
+    if len(population) < target_population_size:
+        raise IndexError(
```
toshihikoyanase, line 586: Could you explain why you choose `IndexError`? `ValueError` may be suitable since input values are invalid.

optuna/samplers/_nsgaiii.py
```diff
@@ -0,0 +1,637 @@
...
+        state: TrialState,
+        values: Sequence[float] | None,
+    ) -> None:
+        assert state in [TrialState.COMPLETE, TrialState.FAIL, TrialState.PRUNED]
+        if self._constraints_func is not None:
+            _process_constraints_after_trial(self._constraints_func, study, trial, state)
+        self._random_sampler.after_trial(study, trial, state, values)
+
+
+# TODO(Shinichi) Replace with math.comb after support for python3.7 is deprecated.
```
toshihikoyanase, line 408: The definition of `math.comb` is different from `_multi_choose`.
`math.comb`: `n! / (k! * (n - k)!)`
`_multi_choose`: `(n + k - 1) / (k! * (n-1)!)`

So, we might elaborate this code comment regarding the difference.

toshihikoyanase (commented):
Thanks for your update. I have two minor comments.

optuna/samplers/_nsgaiii.py
```diff
@@ -0,0 +1,637 @@
...
+        elite_population_count = reference_point_to_elite_population_count[reference_point_idx]
+        nearest_points_count_to_reference_points[elite_population_count].append(
+            reference_point_idx
+        )
+
+    count = -1
+    additional_elite_population: list[FrozenTrial] = []
+    is_shuffled: defaultdict[int, bool] = defaultdict(bool)
+    while len(additional_elite_population) < target_population_size:
+        if not nearest_points_count_to_reference_points[count]:
```
toshihikoyanase, line 618: `len([]) == 0` might be easier for readers than `not []`.

```suggestion
        if len(nearest_points_count_to_reference_points[count]) == 0:
```

optuna/samplers/_nsgaiii.py
```diff
@@ -0,0 +1,637 @@
...
+    # Initialize weights.
+    weights = np.eye(n_objectives)
+    weights[weights == 0] = 1e6
+
+    # Calculate extreme points to normalize objective values.
+    # TODO(Shinichi) Reimplement to reduce time complexity.
+    asf_value = np.max(
+        np.einsum("nm,dm->dnm", objective_matrix, weights),
+        axis=2,
+    )
```
toshihikoyanase, line 484: I guess this corresponds to eq. 4 of https://www.egr.msu.edu/~kdeb/papers/k2012009.pdf. We might mention it in the code comment, but we can work on it in a follow-up PR.

optuna/samplers/_nsgaiii.py
```diff
@@ -0,0 +1,637 @@
...
+        "ni,pi,p,pm->npm",
+        objective_matrix,
+        reference_points,
+        1 / reference_point_norm_squared,
+        reference_points,
+    )
+    distance_from_reference_lines = np.linalg.norm(
+        objective_matrix[:, np.newaxis, :] - perpendicular_vectors_to_reference_lines,
+        axis=2,
+    )
```
toshihikoyanase, line 544: ditto. This seems to correspond to L5 to 7 of Algorithm 3 in https://www.egr.msu.edu/~kdeb/papers/k2012009.pdf. We'd like to improve the readability in a follow-up PR.

toshihikoyanase (commented):
The change basically looks good to me. Let me add some minor comments.

optuna/samplers/_nsgaiii.py
Alnusjaponica: I fixed it. Thank you for the indication.

optuna/samplers/_nsgaiii.py
toshihikoyanase: I'm sorry, but I got confused. It should be `if len(nearest_points_count_to_reference_points[count]) > 0:`.

toshihikoyanase (approved):
LGTM. Thank you for completing such a long-running PR!

We have possible follow-up tasks, but I don't think they are critical:
- https://github.com/optuna/optuna/pull/4436#discussion_r1178603627
- https://github.com/optuna/optuna/pull/4436#discussion_r1190666906
- https://github.com/optuna/optuna/pull/4436#discussion_r1190661542

PR #4441, opened 2023-02-20
Add unit tests for the _BoTorchGaussianProcess class

g-votte:
This is a follow-up PR of #4401.

The test cases for the `_BoTorchGaussianProcess` are added in `test_botorch.py` with the following scenarios:
- there are multiple trials and the Gram matrix is regular
- there are multiple trials and the Gram matrix is non-regular
- there is a single trial
- there is a single parameter

### DIFF
--- a/tests/terminator_tests/improvement_tests/gp_tests/test_botorch.py
+++ b/tests/terminator_tests/improvement_tests/gp_tests/test_botorch.py
@@ -4,7 +4,7 @@ from optuna.trial import create_trial
 
 
 def test_fit_predict() -> None:
-    # A typical fit-predict scenario is being tested here, where there are more than one trials 
+    # A typical fit-predict scenario is being tested here, where there are more than one trials
     # and the Gram matrix is a regular one.
     trials = [
         create_trial(
--- a/tests/terminator_tests/improvement_tests/gp_tests/test_botorch.py
+++ b/tests/terminator_tests/improvement_tests/gp_tests/test_botorch.py
@@ -1,0 +1,108 @@
+from optuna.distributions import FloatDistribution
+from optuna.terminator.improvement.gp.botorch import _BoTorchGaussianProcess
+from optuna.trial import create_trial
+
+
+def test_fit_predict() -> None:
+    # A typical fit-predict scenario is being tested here, where there are more than one trials 
+    # and the Gram matrix is a regular one.
+    trials = [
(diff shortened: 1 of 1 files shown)


tests/terminator_tests/improvement_tests/gp_tests/test_botorch.py
```diff
@@ -0,0 +1,108 @@
...
+        ),
+    ]
+
+    gp = _BoTorchGaussianProcess()
+    gp.fit(trials)
+    gp.predict_mean_std(trials)
+
+
+def test_fit_predict_non_regular_gram_matrix() -> None:
+    # This test case validates that the GP class works even when the Gram matrix is non-regular,
```
eukaryo, line 78:
```suggestion
    # This test case validates that the GP class works even when the Gram matrix is not positive definite,
```

tests/terminator_tests/improvement_tests/gp_tests/test_botorch.py
```diff
@@ -0,0 +1,108 @@
...
+            },
+        ),
+    ]
+
+    gp = _BoTorchGaussianProcess()
+    gp.fit(trials)
+    gp.predict_mean_std(trials)
+
+
+def test_fit_predict_non_regular_gram_matrix() -> None:
```
eukaryo, line 77: I think here is typo. In this case, the Gram matrix seems to be quadratic square matrix such that all elements are 1. It is regular, and not positive definite.

eukaryo (approved):
My comments above are due to misunderstandings. On second thought, the current code is LGTM.

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM.

How about adding test cases for extreme values? That is, 
- An infinite trial value.
- A distribution concentrated on one point.
- Two distributions with very different scales

We can work on them as a follow-up.

g-votte:
@HideakiImamura 
Good catch. The following cases actually raise an error due to `NaN` calculation.

- An infinite trial value.
- A distribution concentrated on one point.

Let us work on them on follow-up PRs, though. While we'd need some more additional discussions to determine the behaviors, let's just target adding trivial unit tests in this PR.

HideakiImamura:
Thanks for the reply. It's OK to add those tests as follow-ups. LGTM.

HideakiImamura (approved):

PR #4442, opened 2023-02-20
Simplify an argument's name of RegretBoundEvaluator

g-votte:
This is a follow-up PR of #4401, renaming an argument of the `RegretBoundEvaluator` to a simpler one.

### DIFF
--- a/optuna/terminator/improvement/evaluator.py
+++ b/optuna/terminator/improvement/evaluator.py
@@ -41,12 +41,12 @@ class RegretBoundEvaluator(BaseImprovementEvaluator):
         gp: Optional[BaseGaussianProcess] = None,
         top_trials_ratio: float = DEFAULT_TOP_TRIALS_RATIO,
         min_n_trials: int = DEFAULT_MIN_N_TRIALS,
-        min_lcb_n_samples: int = 2000,
+        min_lcb_n_additional_samples: int = 2000,
     ) -> None:
         self._gp = gp or _BoTorchGaussianProcess()
         self._top_trials_ratio = top_trials_ratio
         self._min_n_trials = min_n_trials
-        self._min_lcb_n_samples = min_lcb_n_samples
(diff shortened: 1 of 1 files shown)


optuna/terminator/improvement/evaluator.py
```diff
@@ -59,7 +59,7 @@ def get_preprocessing(self, add_random_inputs: bool = False) -> BasePreprocessin
         ]
 
         if add_random_inputs:
-            processes += [AddRandomInputs(self._min_lcb_n_additional_trials)]
+            processes += [AddRandomInputs(self._min_lcb_n_samples)]
```
toshihikoyanase, line 62: Do you think `min_lcb_n_random_samples` (or `min_lcb_n_additional_samples`) is too long?
`min_lcb_n_samples` sounds like the total number of samples used to calculate minimum LCB.

```suggestion
            processes += [AddRandomInputs(self._min_lcb_n_random_samples)]
```

toshihikoyanase (commented):
Thank you for your update. I have a question about the lost information in the new name.

optuna/terminator/improvement/evaluator.py
g-votte: I actually think using a long name is also fine. I rather prefer `min_lcb_n_additional_samples` to `min_lcb_n_random_samples` because a non-random sampling method may be used in the future.

g-votte: I changed the name to `min_lcb_n_additional_samples`. Please let me know if you have any other idea or concern.

toshihikoyanase: Thank you for your update. Sounds good to me.

toshihikoyanase (approved):
LGTM. Thank you!

toshihikoyanase:
Let me merge this PR with a single approval since it only renames variables.

PR #4458, opened 2023-02-22
Make mac-test "required" on PRs.

gen740:
## Motivation

Make mac-test "required" on PRs.

## Description of the changes

- Add scheduled test to the mac-test.


gen740:
Make sure integration tests don't run on PRs.

.github/workflows/mac-tests.yml
```diff
@@ -57,15 +55,22 @@ jobs:
...
         pip freeze --all
 
-    - name: Tests
+    - name: Scheduled tests
+      if:  ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
       run: |
         pytest tests -m "not integration"
 
+    - name: Tests
+      if:  ${{ github.event_name != 'schedule' }}
```
eukaryo, line 64:
```suggestion
      if:  ${{ github.event_name != 'schedule' && github.event_name != 'workflow_dispatch' }}
```

eukaryo (commented):
Almost LGTM; there seems to be a case which pytest will be executed in duplicate.

.github/workflows/mac-tests.yml
```diff
@@ -7,11 +7,10 @@ on:
   push:
     branches:
       - master
-  pull_request:
-    paths:
-      - '.github/workflows/mac-tests.yml'
+  pull_request: {}
   schedule:
     - cron: '0 23 * * SUN-THU'
+  workflow_dispatch:
```
eukaryo, line 13: Good catch! Thank you.

eukaryo (approved):
LGTM!

.github/workflows/mac-tests.yml
```diff
@@ -57,15 +56,22 @@ jobs:
...
+      run: |
+        pytest tests -m "not integration and not slow"
+
+
   tests-integration-mac:
     runs-on: macos-latest
 
     # Not intended for forks.
-    if: github.repository == 'optuna/optuna'
+    if: github.repository == 'optuna/optuna' && ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
```
HideakiImamura, line 74: This type of condition in [`tests-integration.yml`](https://github.com/optuna/optuna/blob/master/.github/workflows/tests-integration.yml#L18) is `(A && B) || B`. Should we use the same condition?

HideakiImamura: And, this should be applied to the `tests-mac` job.

HideakiImamura (commented):
Thanks for the PR. I have some comments on the conditions that limit the execution of tests on forked repos.

.github/workflows/mac-tests.yml
```diff
@@ -22,7 +21,7 @@ jobs:
     runs-on: macos-latest
 
     # Not intended for forks.
-    if: github.repository == 'optuna/optuna'
+    if: (github.event_name == 'schedule' && github.repository == 'optuna/optuna') || (github.event_name != 'schedule')
```
toshihikoyanase, line 24: I think the above comment is not consistent with this condition since non-scheduled workflows are expected to be executed in fork repositories. We can see a similar change in L74.

gen740: I update the comment.

gen740: The same conditions appear on Tests.yml (and other tests). Is it better to add the comment to them?

toshihikoyanase: Thank you for your update. I think we should work on `Tests.yml` in a follow-up PR to minimize the changes in this PR.

HideakiImamura (approved):
Looks very good.

HideakiImamura:
@toshihikoyanase Please merge this PR after your approval.

toshihikoyanase (approved):
Thank you for letting me know. LGTM!

PR #4461, opened 2023-02-24
Fix `Checks(integration)` about `terminator/.../botorch.py`

Alnusjaponica:
## Motivation
Fix current [`Checks(integration)` failure](https://github.com/optuna/optuna/actions/runs/4230036371/jobs/7346954668)
## Description of the changes
Currently, mypy fails at `torch.no_grad()` when torch is installed. This PR adds a `type:ignore` to prevent this.

### DIFF
--- a/optuna/terminator/improvement/gp/botorch.py
+++ b/optuna/terminator/improvement/gp/botorch.py
@@ -75,7 +75,7 @@ class _BoTorchGaussianProcess(BaseGaussianProcess):
 
         x, _ = _convert_trials_to_tensors(trials)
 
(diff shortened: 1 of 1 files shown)


toshihikoyanase:
Good catch!
The fix sounds reasonable to me according to the following error message:

> optuna/terminator/improvement/gp/botorch.py:78: error: Call to untyped function "no_grad" in typed context  [no-untyped-call]

Alnusjaponica:
@toshihikoyanase Thank you for your review. 
@g-votte Could you take another look?

toshihikoyanase (approved):
LGTM!

g-votte (approved):
LGTM

PR #4468, opened 2023-02-26
Cache all trials in Study with delayed relative sampling

amylase:
This is yet anothe try on #4141, with delayed relative sampling.
Context: #4141, #4366

## Motivation

See #4141, #4366

## Description of the changes

Delayed relative sampling until it is necessary. GridSampler is also updated because it uses own relative sampling logic.

### DIFF
--- a/optuna/samplers/_grid.py
+++ b/optuna/samplers/_grid.py
@@ -136,6 +136,10 @@ class GridSampler(BaseSampler):
         if "grid_id" in trial.system_attrs or "fixed_params" in trial.system_attrs:
             return {}
 
+        # self._sample_grid_id(study, trial)
+        return {}
+
+    def _sample_grid_id(self, study: Study, trial: FrozenTrial) -> int:
         target_grids = self._get_unvisited_grid_ids(study)
 
         if len(target_grids) == 0:
@@ -158,8 +162,7 @@ class GridSampler(BaseSampler):
 
         study._storage.set_trial_system_attr(trial._trial_id, "search_space", self._search_space)
         study._storage.set_trial_system_attr(trial._trial_id, "grid_id", grid_id)
-
-        return {}
+        return grid_id
 
     def sample_independent(
         self,
@@ -168,7 +171,11 @@ class GridSampler(BaseSampler):
         param_name: str,
         param_distribution: BaseDistribution,
     ) -> Any:
-        if "grid_id" not in trial.system_attrs:
+        if "grid_id" in trial.system_attrs:
+            grid_id = trial.system_attrs["grid_id"]
+        elif "fixed_params" not in trial.system_attrs:
+            grid_id = self._sample_grid_id(study, trial)
+        else:
             message = "All parameters must be specified when using GridSampler with enqueue_trial."
             raise ValueError(message)
 
@@ -179,7 +186,6 @@ class GridSampler(BaseSampler):
         # TODO(c-bata): Reduce the number of duplicated evaluations on multiple workers.
         # Current selection logic may evaluate the same parameters multiple times.
         # See https://gist.github.com/c-bata/f759f64becb24eea2040f4b2e3afce8f for details.
-        grid_id = trial.system_attrs["grid_id"]
         param_value = self._all_grids[grid_id][self._param_names.index(param_name)]
         contains = param_distribution._contains(param_distribution.to_internal_repr(param_value))
         if not contains:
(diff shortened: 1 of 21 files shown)


optuna/samplers/_tpe/sampler.py
```diff
@@ -622,7 +622,7 @@ def _get_observation_pairs(
     scores = []
     values: Dict[str, List[Optional[float]]] = {param_name: [] for param_name in param_names}
     violations: Optional[List[float]] = [] if constraints_enabled else None
-    for trial in study.get_trials(deepcopy=False, states=states):
+    for trial in study._get_trials(deepcopy=False, states=states, use_cache=True):
```
c-bata, line 625: Does it sound reasonable to make `use_cache=False` when `constant_liar` option is true?

c-bata (commented):
Thank you for your pull request! I left some suggestions.

amylase:
Measured the performance of this change under the same settings as [this experiment](https://gist.github.com/c-bata/16d654471393f59e064482ffb8b52bf8) in my computer.

| branch / storage | RDBStorage with CachedStorage | RDBStorage |
|-|-|-|
| master | 68.78 sec | 194.87 sec |
| this PR | 50.12 sec (1.37x faster) | 70.96 sec (2.75x faster) |

HideakiImamura:

optuna/samplers/_tpe/sampler.py
amylase: This is a good point. I discussed this with @c-bata -san in person and concluded that we should do this.
When `constant_liar` is `True`, parameters of latest trials is necessary to avoid evaluating the similar parameters in distributed environment.

optuna/samplers/_grid.py
```diff
@@ -133,9 +133,9 @@ def sample_relative(
 
         # When the trial is created by RetryFailedTrialCallback or enqueue_trial, we should not
         # assign a new grid_id.
-        if "grid_id" in trial.system_attrs or "fixed_params" in trial.system_attrs:
-            return {}
+        return {}
```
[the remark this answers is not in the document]
amylase: I am a bit against this, because "supporting the delayed relative sampling" includes suporting it in `GridSampler`.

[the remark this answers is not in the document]
amylase: I also tested the code without the changes on `_grid.py`. In this case, the code always throws exception on `sample_independent` if `enqueue_trial` is used. I think it is easier to fix `_grid.py` as well.

optuna/study/_optimize.py
```diff
@@ -220,6 +220,8 @@ def _run_trial(
         frozen_trial = study._storage.get_trial(trial._trial_id)
         raise
     finally:
+        # We must invalidate all trials cache here as it is only valid within a trial.
+        study._thread_local.cached_all_trials = None
```
amylase, line 224: This is not the right place to invalidate the cache. The stale cache will be used forever if optuna is used from ask and tell interface.

optuna/samplers/_grid.py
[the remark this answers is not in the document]
c-bata: I see. How about just moving this logic into `infer_relative_search_space()` method instead? There is essentially no difference whether the grid_id is assigned in sample_relative or infer_relative_search_space, and the changes can be kept small.

Screenshot 2023-03-16 12 19 30

https://github.com/optuna/optuna/compare/master...c-bata:optuna:review-delayed-sample-relative?expand=1

The unit tests seemed to pass.
https://github.com/c-bata/optuna/actions/runs/4427465420/jobs/7765092867

c-bata:
Could you add the following test scenario?

```python
@pytest.mark.parametrize("n_jobs", [1, 2])
def test_cache_is_invalidated(n_jobs) -> None:
    study = create_study()

        assert trial._relative_params is None
        assert study._thread_local.cached_all_trials is None

        trial.suggest_float("x", -10, 10)
        assert trial._relative_params is not None
        assert study._thread_local.cached_all_trials is not None
        return -1

    study.optimize(objective, n_trials=10, n_jobs=n_jobs)
```

c-bata (commented):
Almost LGTM.

@amylase Could you resolve a file conflict?

gen740 (approved):
LGTM!

gen740:
There seems to be something wrong with the CI workflows. @amylase Could you re-run the workflows?

optuna/trial/_trial.py
```diff
@@ -61,9 +61,16 @@ def _init_relative_params(self) -> None:
         self.relative_search_space = self.study.sampler.infer_relative_search_space(
             study, self._cached_frozen_trial
         )
-        self.relative_params = self.study.sampler.sample_relative(
-            study, self._cached_frozen_trial, self.relative_search_space
-        )
+        self._relative_params: Optional[Dict[str, Any]] = None
```
c-bata, line 64: Since `_init_relative_params()` does not actually initialize relative_params due to this change, how about removing this method and writing these logics directly in `__init__()` method?

c-bata (approved):
Thanks for the update! LGTM with a nit.

c-bata (approved):

PR #4469, opened 2023-02-27
Add an experimental decorator to RegretBoundEvaluator

g-votte:
This is a follow-up PR of #4401 and adds the experimental decorator to `RegretBoundEvaluator`.

### DIFF
--- a/optuna/terminator/improvement/evaluator.py
+++ b/optuna/terminator/improvement/evaluator.py
@@ -14,9 +14,9 @@ from optuna.terminator.improvement._preprocessing import PreprocessingPipeline
 from optuna.terminator.improvement._preprocessing import SelectTopTrials
 from optuna.terminator.improvement._preprocessing import ToMinimize
 from optuna.terminator.improvement._preprocessing import UnscaleLog
+from optuna.terminator.improvement.gp.base import _min_lcb
+from optuna.terminator.improvement.gp.base import _min_ucb
 from optuna.terminator.improvement.gp.base import BaseGaussianProcess
-from optuna.terminator.improvement.gp.base import min_lcb
-from optuna.terminator.improvement.gp.base import min_ucb
 from optuna.terminator.improvement.gp.botorch import _BoTorchGaussianProcess
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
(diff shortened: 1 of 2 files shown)


contramundum53 (approved):
LGTM.

contramundum53:
Could you review this PR?

optuna/terminator/improvement/evaluator.py
```diff
@@ -35,6 +36,7 @@ def evaluate(
         pass
 
 
+@experimental_class("3.2.0")
 class RegretBoundEvaluator(BaseImprovementEvaluator):
```
toshihikoyanase, line 40: `BaseImprovementEvaluator` is not marked as experimental. Do we keep the compatibility from now?

If we add the experimental decorator to both `RegretBoundEvaluator` and `BaseImprovementEvaluator`, only the warning message of `RegretBoundEvaluator` is shown. So, the duplicated warning messages do not suffer users.

toshihikoyanase (commented):
The change looks solid to me, but I have a question about `BaseImpromvenetEvaluator`.

optuna/terminator/improvement/evaluator.py
g-votte: Good catch. As long as the message is not duplicated, let's make the base classes experimental as well.

I also changed related class and functions.
- `BaseGaussianProcess` is now an experimental class, since it is exposed to the user and guaranteeing the compatibility is still risky.
- `min_ucb` and `min_lcb` is now private functions, since their exposure is not required.

toshihikoyanase: Thank you for your update. Sounds reasonable to me!

toshihikoyanase (approved):
I confirmed that the experimental warnings were shown as expected:

```console
>>> optuna.terminator.improvement.evaluator.RegretBoundEvaluator()
<stdin>:1: ExperimentalWarning: RegretBoundEvaluator is experimental (supported from v3.2.0). The interface can change in the future.
<optuna.terminator.improvement.evaluator.RegretBoundEvaluator object at 0x1034255b0>

>>> class T(optuna.terminator.improvement.gp.base.BaseGaussianProcess):
...     def fit(self, _): pass
...     def predict_mean_std(self, _): pass
... 
>>> T()
<stdin>:1: ExperimentalWarning: BaseGaussianProcess is experimental (supported from v3.2.0). The interface can change in the future.
<__main__.T object at 0x109482760>
```

And I also confirmed that `_BoTorchGaussianProcess` did not raise experimental warnings:

```console
>>> optuna.terminator.improvement.gp.botorch._BoTorchGaussianProcess()
<optuna.terminator.improvement.gp.botorch._BoTorchGaussianProcess object at 0x1029415b0>
```

LGTM!

toshihikoyanase:
Let me update the title for the release note.

Issue #4472, opened 2023-02-27
QMSSampler incorrectly discards `seed=0`


Yura52:
### Expected behavior

I have not used this sampler, but there is a bug which comes into play when `seed=0`
https://github.com/optuna/optuna/blob/e774c0be70188ae3e62e9d66ad6949bbccc0e149/optuna/samplers/_qmc.py#L155

It should be this instead:

```
self._seed = np.random.PCG64().random_raw() if seed is None else seed
```

### Environment

N/A

### Error messages, stack traces, or logs

```shell
N/A
```

### Steps to reproduce

N/A

contramundum53:
Nice catch!

PR #4480, opened 2023-03-02
Make `QMSSampler` samplers reproducible with `seed=0`

nzw0301:
## Motivation

Resolve https://github.com/optuna/optuna/issues/4472.

## Description of the changes

- Fix the seeding part as suggested by #4472 (thanks!)
- Fix docs

### DIFF
--- a/optuna/samplers/_qmc.py
+++ b/optuna/samplers/_qmc.py
@@ -152,7 +152,7 @@ class QMCSampler(BaseSampler):
         warn_independent_sampling: bool = True,
     ) -> None:
         self._scramble = scramble
-        self._seed = seed or np.random.PCG64().random_raw()
+        self._seed = np.random.PCG64().random_raw() if seed is None else seed
         self._independent_sampler = independent_sampler or optuna.samplers.RandomSampler(seed=seed)
         self._initial_search_space: Optional[Dict[str, BaseDistribution]] = None
         self._warn_independent_sampling = warn_independent_sampling
--- a/tests/samplers_tests/test_qmc.py
+++ b/tests/samplers_tests/test_qmc.py
@@ -251,9 +251,7 @@ def test_sample_relative_seeding(seed: int, scramble: bool, qmc_type: str) -> No
     objective: Callable[[Trial], float] = lambda t: t.suggest_float("x", 0, 1)
 
     # Base case.
-    sampler = _init_QMCSampler_without_exp_warning(
-        scramble=scramble, qmc_type=qmc_type, seed=seed
-    )
+    sampler = _init_QMCSampler_without_exp_warning(scramble=scramble, qmc_type=qmc_type, seed=seed)
     study = optuna.create_study(sampler=sampler)
     study.optimize(objective, n_trials=10, n_jobs=1)
     past_trials = study._storage.get_all_trials(study._study_id, states=(TrialState.COMPLETE,))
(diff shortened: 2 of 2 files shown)


nzw0301:
In my mind, possible approaches to moving forward this PR are as follows:

1. Revert cf2ebf7
    - a) Add a warning that PyCMASampler's seed should be non-zero for reproducibility to its docs or code.
2. Wait for pycma

nzw0301:
The test, `test_reproducible` is wrapped by `@pytest.mark.slow`, so the CI cannot show the failure case, doesn't it?

not522:
In my understanding, the behavior of pycma with `seed=0` is expected. https://github.com/CMA-ES/pycma/issues/111

nzw0301:
@not522 Thank you for your comment!

nzw0301:
I realised that pycma's `seed=0` issue should be addressed by another PR to focus on the original issue. So the current PR is ready for review. Thanks.

HideakiImamura:
@toshihikoyanase @gen740 Could you review this PR?

optuna/samplers/_qmc.py
```diff
@@ -152,7 +152,7 @@ def __init__(
         warn_independent_sampling: bool = True,
     ) -> None:
         self._scramble = scramble
-        self._seed = seed or np.random.PCG64().random_raw()
+        self._seed = np.random.PCG64().random_raw() if seed is None else seed
```
toshihikoyanase, line 155: The change looks solid to me.
How about adding a test case to detect the bug in the corresponding test case?

The confirmed that `test_sample_relative_seeding` failed if I added the case `seed=0`.
```python
@pytest.mark.parametrize("seed", [0, 12345])
@pytest.mark.parametrize("scramble", [True, False])
@pytest.mark.parametrize("qmc_type", ["sobol", "halton"])
def test_sample_relative_seeding(seed: int, scramble: bool, qmc_type: str) -> None:
```

toshihikoyanase (commented):
Thank you for your PR. The code fix looks good, but I have two comments about the tests.

optuna/samplers/_qmc.py
toshihikoyanase: Thanks!

tests/samplers_tests/test_qmc.py
```diff
@@ -240,29 +240,26 @@ def test_sample_relative_sobol() -> None:
...
-    # If empty search_space, return {}
+    # If empty search_space, return {}.
     np.testing.assert_allclose(samples, ref_samples, rtol=1e-6)
 
 
+@pytest.mark.parametrize("seed", [0, 12345])
 @pytest.mark.parametrize("scramble", [True, False])
 @pytest.mark.parametrize("qmc_type", ["sobol", "halton"])
-def test_sample_relative_seeding(scramble: bool, qmc_type: str) -> None:
+def test_sample_relative_seeding(seed: int, scramble: bool, qmc_type: str) -> None:
```
toshihikoyanase, line 250: (This comment is optional.)
I guess this code follows my suggestion, but it may be more natural if the order of arguments is consistent with `_init_QMCSampler_without_exp_warning`.

How about reordering them like `scramble`, `qmc_type` and `seed`?

Sorry for bothering you.

toshihikoyanase (approved):
LGTM! I have an optional comment, but I can work on it in a follow-up PR.

gen740 (approved):
LGTM!

tests/samplers_tests/test_qmc.py
nzw0301: No worries. I've fixed it too.

toshihikoyanase: Thank you!

toshihikoyanase (approved):
Thank you for your update. LGTM again!

PR #4483, opened 2023-03-03
Avoid error at import time for `optuna.terminator.improvement.gp.botorch`

hvy:
## Motivation

Delaying import error when `torch` is missing to when the component is actually used and not at submodule import time. 
It's made similar to e.g. https://github.com/optuna/optuna/blob/release-v3.1.0/optuna/integration/botorch.py#L55-L58.

## Description of the changes

Changes to use string type type hinting for optional dependencies.

### DIFF
--- a/optuna/terminator/improvement/gp/botorch.py
+++ b/optuna/terminator/improvement/gp/botorch.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import List
 from typing import Optional
 from typing import Tuple
--- a/optuna/terminator/improvement/gp/botorch.py
+++ b/optuna/terminator/improvement/gp/botorch.py
@@ -1,8 +1,6 @@
 from __future__ import annotations
 
-from typing import List
 from typing import Optional
-from typing import Tuple
 
 import numpy as np
 
@@ -45,7 +43,7 @@ class _BoTorchGaussianProcess(BaseGaussianProcess):
 
     def fit(
         self,
-        trials: List[FrozenTrial],
+        trials: list[FrozenTrial],
     ) -> None:
         self._trials = trials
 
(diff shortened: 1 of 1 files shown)


optuna/terminator/improvement/gp/botorch.py
```diff
@@ -84,7 +84,7 @@ def predict_mean_std(
         return mean.detach().numpy(), std.detach().numpy()
 
 
-def _convert_trials_to_tensors(trials: List[FrozenTrial]) -> Tuple[torch.Tensor, torch.Tensor]:
+def _convert_trials_to_tensors(trials: List[FrozenTrial]) -> Tuple["torch.Tensor", "torch.Tensor"]:
```
c-bata, line 87: Does it sound reasonable to use `from __future__ import annotations` instead of making it string? Then we can also replace `typing.List` and `typing.Tuple` with `list` and `tuple`.

```suggestion
def _convert_trials_to_tensors(trials: list[FrozenTrial]) -> tuple[torch.Tensor, torch.Tensor]:
```

hvy: Sounds good, let me adopt it!

hvy: PTAL.

optuna/terminator/improvement/gp/botorch.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import List
 from typing import Optional
 from typing import Tuple
```
c-bata, line 5: [nit] Might be good to replace all `typing.List` and `typing.Tuple` in this file with `list` and `tuple`. Do you want to work on it as a follow-up?

c-bata (approved):
LGTM with a nit.

optuna/terminator/improvement/gp/botorch.py
hvy: Thanks fixed this in this PR.

c-bata (approved):
LGTM!

HideakiImamura (approved):
LGTM.

PR #4488, opened 2023-03-06
Avoid standardizing `Yvar` in `_BoTorchGaussianProcess`

g-votte:
This is a follow-up PR of #4401 and fixes the standardization process of `_BoTorchGaussianProcess`.

## Motivation
In the GP fitting logic of `_BoTorchGaussianProcess`, originally, the scale of the noise for `FixedNoiseGP` was not expected to be impacted by the scale of `Y`. However, in the current implementation, the noise value is scaled depending on `Y`, because [both `y` and `Yvar` are processed by the `Standardize` module](https://github.com/pytorch/botorch/blob/ba90c9b5069f85d3f071560e95f02984c8941f77/botorch/models/gp_regression.py#L246-L247) inside `FixedNoiseGP.__init__`.

## Description of the changes
This PR fixes the unintended behavior by introducing `_StandardizeIgnoringYvar`, which overrides `Standardize` and ignores `YVar` when standardizing.

### DIFF
--- a/optuna/terminator/improvement/gp/botorch.py
+++ b/optuna/terminator/improvement/gp/botorch.py
@@ -33,6 +33,14 @@ __all__ = [
 ]
 
 
+class _StandadizeIgnoringYvar(Standardize):
+    def forward(
+        self, Y: torch.Tensor, Yvar: Optional[torch.Tensor] = None
+    ) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
+        Y, _ = super().forward(Y, None)
+        return Y, Yvar
+
+
 class _BoTorchGaussianProcess(BaseGaussianProcess):
     def __init__(self) -> None:
         _imports.check()
@@ -60,7 +68,7 @@ class _BoTorchGaussianProcess(BaseGaussianProcess):
             y,
             torch.full_like(y, 1e-8),
             input_transform=Normalize(d=self._n_params, bounds=bounds),
-            outcome_transform=Standardize(m=1),
+            outcome_transform=_StandadizeIgnoringYvar(m=1),
         )
 
         mll = gpytorch.mlls.ExactMarginalLogLikelihood(self._gp.likelihood, self._gp)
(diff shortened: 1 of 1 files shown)


HideakiImamura:
@contramundum53 @gen740 Could you review this PR?

HideakiImamura (commented):
The noise for `FixedNoiseGP` is set by estimating the scale of the noise that would actually objective function has, so it behaves as expected to be normalized to include noise.

However, as you say, it is difficult to actually determine the scale of the noise that would actually objective function has, so what we often do is the following, where the noise is a constant multiple of the standard deviation calculated from the samples.
```python
self._gp = FixedNoiseGP(
            x,
            y,
            torch.full_like(y, 1e-8 * y.std()),
            ...
        )
```

This would solve the problem with a one-line fix. It is OK for this problem?

optuna/terminator/improvement/gp/botorch.py
```diff
@@ -55,10 +55,11 @@ def fit(
         y = torch.tensor([trial.value for trial in trials], dtype=torch.float64)
         y = torch.unsqueeze(y, 1)
 
+        noise = 1e-8 * y.std().item()
         self._gp = FixedNoiseGP(
             x,
             y,
-            torch.full_like(y, 1e-8),
+            torch.full_like(y, noise),
```
gen740, line 62: Almost LGTM. But how about writing in one line?
```suggestion
            torch.full_like(y, 1e-8 * y.std().item()),
```

g-votte: This actually takes [the follow-up PR](https://github.com/optuna/optuna/pull/4490/files) into consideration, where there would be too many things in one line if we [get the min noise value from the gpytorch setting](https://github.com/optuna/optuna/pull/4490/files#diff-82cb56a8e78506746061bbf40a9f6b1852cdd6a93406348a3bb1b3b4c6f8a05eR63). 

Anyway, I think writing one line is much simpler so far in this PR. Thanks for your suggestion!

contramundum53 (approved):
LGTM.

gen740 (approved):
LGTM!

PR #4491, opened 2023-03-07
Move `optuna.samplers._search_space.group_decomposed.py` to `optuna.search_space.group_decomposed.py`

cross32768:
## Motivation
`optuna.samplers._search_space` module should not be in `samplers/` since some codes outside of `samplers/` use this module (i.e. `optuna.samplers._search_space.IntersectionSearchSpace` is used from `optuna/terminator/improvement/evaluator.py`) and this module is essentially independent from samplers.

To resolve this problem, I try to move `optuna.samplers._search_space` module to `optuna.search_space`.
To make the PR size small, first I move the place of `optuna.samplers._search_space.group_decomposed.py`.
I will create the following PR to move the place of `optuna.samplers._search_space.intersection.py` later.

## Description of the changes
* Move the place of `optuna.samplers._search_space.group_decomposed.py` to `optuna.search_space.group_decomposed.py`
* Move the place of `tests.sampler_tests.search_space_tests.test_group_decomposed.py` to `tests.search_space_tests.test_group_decomposed.py`
* Fix some import path for the above change
* Change the directory name `_search_space` to `search_space` since current `_search_space` directory has some public API in `intersetion.py` (this file will be moved in the following PR).


HideakiImamura:

tests/search_space_tests/test_group_decomposed.py
```diff
@@ -1,18 +1,17 @@
 import pytest
 
+import optuna
```
HideakiImamura, line 3: This is not a strong opinion, but how about importing `_GroupDecomposedSearchSpace` and `_SearchSpaceGroup` directly?
```suggestion
from optuna.search_space import _GroupDecomposedSearchSpace
from optuna.search_space import _SearchSpaceGroup
```

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM. I have a minor suggestion. Please take a look.

tests/search_space_tests/test_group_decomposed.py
cross32768: I fixed it. Thank you pointing out! (I changed these imports to resolve path problem in the middle of change and forgot to put back.)

gen740 (approved):
LGTM!

cross32768:
@HideakiImamura Cloud you take another look?

PR #4496, opened 2023-03-08
fix ci test conditions

gen740:
## Motivation

Currently, when trigger test and test-integration using `workflow_dispatch`,
both Tests and Scheduled Tests would be executed.

It is enough to check only Scheduled Tests on `workflow_dispatch`.

## Description of the changes

Fix workflow conditions


HideakiImamura (approved):
LGTM.

HideakiImamura:
Let me merge this PR with my one approval since this is a small change.

PR #4505, opened 2023-03-10
Move `optuna.samplers._search_space.intersection.py` to `optuna.search_space.intersection.py` 

cross32768:
## Motivation

This PR is the follow-up PR of #4491 .

## Description of the changes

* Copy `optuna.samplers._search_space.intersection.py` to `optuna.search_space.intersection.py`
* Deprecate class and functions in `optuna.samplers._search_space.intersection.py`
* Fix some import path, comments, and docs for the above change
* Change the module name `_search_space` to `search_space` since current `_search_space` module has some public API in `intersetion.py` 

**NOTE**: This PR **does not fix** import paths in `optuna.terminator` module. I will create the follow-up PR to do this.
  * Reason: `optuna.terminator` module has their original `search_space` module. This is almost the same as original `optuna.samplers._search_space`, but including some small differences. Therefore, I consider fix relating `optuna.terminator` should be tried later.

### DIFF
--- a/docs/source/reference/index.rst
+++ b/docs/source/reference/index.rst
@@ -13,6 +13,7 @@ API Reference
     logging
     pruners
     samplers/index
+    search_space
     storages
     study
     trial
--- a/docs/source/reference/search_space.rst
+++ b/docs/source/reference/search_space.rst
@@ -1,0 +1,14 @@
+.. module:: optuna.search_space
+
+optuna.search_space
+=================
+
+The :mod:`~optuna.search_space` module provides functionality for controlling search space of parameters.
+
+
+.. autosummary::
+   :toctree: generated/
+   :nosignatures:
+
+   optuna.search_space.IntersectionSearchSpace
+   optuna.search_space.intersection_search_space
--- a/docs/source/reference/search_space.rst
+++ b/docs/source/reference/search_space.rst
@@ -1,7 +1,7 @@
 .. module:: optuna.search_space
 
 optuna.search_space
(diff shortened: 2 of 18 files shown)


HideakiImamura:
@contramundum53 @eukaryo Could you review this PR?

contramundum53:
I think `IntersectionSearchSpace` and `intersection_search_space` are both private interfaces because they are inside `_search_space` module and are not exported in any public module. 
Given that, I think we should keep the module name `_search_space`.
What do you think, @toshihikoyanase ?

toshihikoyanase:
> `IntersectionSearchSpace` and `intersection_search_space` are both private interface

Hmm, they are listed in the stable version of the references as follows:

https://optuna.readthedocs.io/en/stable/reference/samplers/generated/optuna.samplers.IntersectionSearchSpace.html
https://optuna.readthedocs.io/en/stable/reference/samplers/generated/optuna.samplers.intersection_search_space.html

So, I guess we treat them as public interfaces for now. In my understanding, `intersection_search_space` is provided to users to create user-defined samplers as explained in the tutorial page:
https://optuna.readthedocs.io/en/stable/tutorial/20_recipes/005_user_defined_sampler.html

contramundum53:
I see, thank you  @toshihikoyanase !
In this case, maybe we need to add underscores to all other private members in the future...

contramundum53 (approved):
LGTM!

eukaryo:
LGTM too!

PR #4507, opened 2023-03-10
Deploy results of visual regression test on Netlify

eukaryo:
## Motivation

I want to have Netlify host the results of a visual regression test built with GitHub Actions so that we can check the results. In addition, I want the URL of the site on Netlify to be automatically commented in the PRs.

## Description of the changes

Added a yml file to workflow.

We will be able to see the results like the PR example below.
https://github.com/eukaryo/optuna/pull/2


toshihikoyanase:
[Notes] We cannot confirm the result of the deployment in this PR because the change in the feature branch cannot access the secrets (e.g., `NETLIFY_AUTH_TOKEN` and `NETLIFY_SITE_ID`) of this repository.

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,58 @@
+name: Visual regression with Netlify
```
toshihikoyanase, line 1: I think this is a kind of regression tests to detect the degradation of visualization functions.
So, how about adding `tests` to the name?
```suggestion
name: Visual regression tests with Netlify
```

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,58 @@
...
+
+on:
+  push:
+    branches:
+      - master
```
HideakiImamura, line 6: - Visual regression test to help review visualization features? If so, I think it may not need to be triggered by master push. 
- Also, I was wondering if it would be ok to only start it when there is a change in the visualization module.
- Furthermore, I think there are cases where we may want to start it manually, so I think it would be better to add `workflow_dispatch`.

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM. Could you check one comment?

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,58 @@
...
+  pull_request: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  documentation:
```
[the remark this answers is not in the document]
eukaryo: Agreed. I renamed it.

.github/workflows/visual-regression.yml
eukaryo: Agreed. I revised the code.

.github/workflows/visual-regression.yml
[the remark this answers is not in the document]
toshihikoyanase: Thanks!

.github/workflows/visual-regression.yml
toshihikoyanase: Thank you.

.github/workflows/visual-regression.yml
toshihikoyanase: Sounds reasonable. In additioin, we can save the usage of netlify by restricting the trigger targets.
I agree with @HideakiImamura.

HideakiImamura (approved):
LGTM.

toshihikoyanase (approved):

toshihikoyanase:
@eukaryo Could you rebase master, please? #4522 resolved the CI failures.

PR #4490, opened 2023-03-06
Change the noise value in `_BoTorchGaussianProcess` to suppress warning messages

g-votte:
Please marge this PR after #4488.

This PR sets the noise value of `FixedNoiseGP` in `_BoTorchGaussianProcess` according to the minimum setting of `gpytorch`.

## Motivation
`gpytorch` defines the minimum noise value [here](https://github.com/cornellius-gp/gpytorch/blob/eb5066ca508dcf5630bd7e6b569591ba9cbe6200/gpytorch/settings.py#L281-L293). 

Currently, the noise value of `_BoTorchGaussianProcess` is set to `1e-8`, but it is actually overwritten by the minimum value setting of `gpytorch`. This frequently causes the following warning message in the Optuna Terminator.

```
.../python3.9/site-packages/gpytorch/likelihoods/noise_models.py:144: NumericalWarning: Very small noise values detected. This will likely lead to numerical instabilities. Rounding small noise values up to 1e-06.
  warnings.warn(
```

## Description of the changes
This PR suppresses the warning message by using the minimum setting of `gpytorch` as the noise value instead of hardcoding it. The specific noise value is changed from `1e-8` to `1e-6`.

PR #4510, opened 2023-03-14
Change the noise value in `_BoTorchGaussianProcess` to suppress warning messages

g-votte:
Resubmitting #4490.

This PR sets the noise value of `FixedNoiseGP` in `_BoTorchGaussianProcess` according to the minimum setting of `gpytorch`.

## Motivation
`gpytorch` defines the minimum noise value [here](https://github.com/cornellius-gp/gpytorch/blob/eb5066ca508dcf5630bd7e6b569591ba9cbe6200/gpytorch/settings.py#L281-L293). 

Currently, the noise value of `_BoTorchGaussianProcess` is set to `1e-8`, but it is actually overwritten by the minimum value setting of `gpytorch`. This frequently causes the following warning message in the Optuna Terminator.

```
.../python3.9/site-packages/gpytorch/likelihoods/noise_models.py:144: NumericalWarning: Very small noise values detected. This will likely lead to numerical instabilities. Rounding small noise values up to 1e-06.
  warnings.warn(
```

## Description of the changes
This PR suppresses the warning message by using the minimum setting of `gpytorch` as the noise value instead of hardcoding it. The specific noise value is changed from `1e-8` to `1e-6`.


toshihikoyanase:
This is a re-submission of #4490.
@contramundum53 Could you review this PR, please? If you are busy, please feel free to remove the assignment.

optuna/terminator/improvement/gp/botorch.py
```diff
@@ -56,8 +56,9 @@ def fit(
         y = torch.unsqueeze(y, 1)
 
         assert self._n_trials is not None
+        noise_scale = gpytorch.settings.min_fixed_noise.value(torch.float64)
         noise = (
-            torch.full_like(y, 1e-8 * y.std().item())
+            torch.full_like(y, noise_scale * y.std().item())
```
toshihikoyanase, line 61: If `y.std().item()` returns a smaller value than `1`, I think elements of `noise` will be smaller than `noise_scale`. If so, we may be able to use `torch.Tensor.clamp_min` (or `torch.clamp`) as can be seen in `FixedGaussianNoise`.

https://github.com/cornellius-gp/gpytorch/blob/58b049b60da07ac5c33682d2bddf3fffabc4e0a0/gpytorch/likelihoods/noise_models.py#LL154C13-L154C47

toshihikoyanase (commented):
I'm sorry but I don't fully understand the behavior of the `train_Yvar` argument of `FixedNoiseGP`. If it is standardized in `FixedNoiseGP` as mentioned in #4488, we may not need to scale the noise.

> However, in the current implementation, the noise value is scaled depending on Y, because [both y and Yvar are processed by the Standardize module](https://github.com/pytorch/botorch/blob/ba90c9b5069f85d3f071560e95f02984c8941f77/botorch/models/gp_regression.py#L246-L247) inside FixedNoiseGP.__init__.

In addition, could you give me some reproducible code of the problem?
I failed to reproduce it and got different errors during the trial-and-error:

> python3.9/site-packages/gpytorch/distributions/multivariate_normal.py:319: NumericalWarning: Negative variance values detected. This is likely due to numerical instabilities. Rounding negative variances up to 1e-10.

g-votte:
@toshihikoyanase 
> In addition, could you give me some reproducible code of the problem?

Though it's not a minimum repro, the following code should show the warning after 20 trials.

```python
from sklearn.ensemble import RandomForestClassifier

import optuna
from optuna.terminator.erroreval import report_cross_validation_scores

terminator = Terminator()

while True:
    trial = study.ask()

    clf = RandomForestClassifier(
        max_depth=trial.suggest_int("max_depth", 2, 32),
        min_samples_split=trial.suggest_float("min_samples_split", 0, 1),
        criterion=trial.suggest_categorical("criterion", ("gini", "entropy")),
    )

    value = scores.mean()
    print(f"Trial #{trial.number} finished with value {value}.")
    study.tell(trial, value)

    if terminator.should_terminate(study):
        print("Terminated by Optuna Terminator!")
        break
```

optuna/terminator/improvement/gp/botorch.py
g-votte: Ahhh, correct. The problem still happens when the std is less than 1.0.
Added `clamp_min` as you suggested. PTAL. https://github.com/optuna/optuna/pull/4510/commits/5a22f986e72156857a2b32d7d29966fb665c40f8

g-votte:
I found a bug in the noise scaling. We should not have scaled `y.std` but `y.var`.

Though this is an independent problem from the warning itself to be precise, let me include its fix into this PR. https://github.com/optuna/optuna/pull/4510/commits/a57911692c1cb17953337af0b819ae1333bfbf9e

optuna/terminator/improvement/gp/botorch.py
toshihikoyanase: Thank you for your confirmation. Looks solid.

toshihikoyanase (approved):
Thank you for your update. Looks good to me.

contramundum53 (approved):
LGTM.

PR #4514, opened 2023-03-15
Change the argument of `intersection_search_space` from `study` to `trials`

cross32768:
## Motivation
The current `calculate()` method of the `IntersectionSearchSpace` class takes `study` as input, but `study.trials` is sufficient for actual processing in this method.

Also, in the design of `optuna.terminator`, it is easier to implement if this method takes `trials` as input (see https://github.com/optuna/optuna/blob/master/optuna/terminator/_search_space/intersection.py and call point of this function). 

For these reasons, I would like to change the argument of the `calculate()` method from `study` to `trials`.

## Description of the changes
* Change the argument of `IntersectionSearchSpace.calculate` and `intersection_search_space` from `Study` to `List[FrozenTrial]`
* Change the call of these functions from `calculate(study)` to `calculate(study.trials)`

### DIFF
--- a/optuna/importance/_base.py
+++ b/optuna/importance/_base.py
@@ -72,7 +72,7 @@ def _get_distributions(study: Study, params: Optional[List[str]]) -> Dict[str, B
     _check_evaluate_args(completed_trials, params)
 
     if params is None:
-        return intersection_search_space(study, ordered_dict=True)
+        return intersection_search_space(study.get_trials(deepcopy=False), ordered_dict=True)
 
     # New temporary required to pass mypy. Seems like a bug.
     params_not_none = params
--- a/optuna/search_space/intersection.py
+++ b/optuna/search_space/intersection.py
@@ -14,7 +14,7 @@ def _calculate(
     trials: list[optuna.trial.FrozenTrial],
     include_pruned: bool = False,
     search_space: Dict[str, BaseDistribution] | None = None,
-    cursor: int = -1,
+    cached_trial_number: int = -1,
 ) -> Tuple[Dict[str, BaseDistribution] | None, int]:
     states_of_interest = [
         optuna.trial.TrialState.COMPLETE,
@@ -27,13 +27,15 @@ def _calculate(
 
     trials_of_interest = [trial for trial in trials if trial.state in states_of_interest]
 
-    next_cursor = trials_of_interest[-1].number + 1 if len(trials_of_interest) > 0 else -1
+    next_cached_trial_number = (
+        trials_of_interest[-1].number + 1 if len(trials_of_interest) > 0 else -1
+    )
     for trial in reversed(trials_of_interest):
-        if cursor > trial.number:
+        if cached_trial_number > trial.number:
             break
 
         if not trial.state.is_finished():
-            next_cursor = trial.number
+            next_cached_trial_number = trial.number
             continue
 
         if search_space is None:
@@ -46,7 +48,7 @@ def _calculate(
             if trial.distributions.get(name) == distribution
         }
 
-    return search_space, next_cursor
+    return search_space, next_cached_trial_number
 
 
 class IntersectionSearchSpace:
@@ -69,7 +71,7 @@ class IntersectionSearchSpace:
     """
 
     def __init__(self, include_pruned: bool = False) -> None:
-        self._cursor: int = -1
+        self._cached_trial_number: int = -1
         self._search_space: Dict[str, BaseDistribution] | None = None
         self._study_id: int | None = None
 
@@ -100,11 +102,11 @@ class IntersectionSearchSpace:
             if self._study_id != study._study_id:
                 raise ValueError("`IntersectionSearchSpace` cannot handle multiple studies.")
 
-        self._search_space, self._cursor = _calculate(
+        self._search_space, self._cached_trial_number = _calculate(
             study.get_trials(deepcopy=False),
             self._include_pruned,
             self._search_space,
-            self._cursor,
+            self._cached_trial_number,
         )
         search_space = self._search_space or {}
 
--- a/optuna/search_space/intersection.py
+++ b/optuna/search_space/intersection.py
@@ -27,7 +27,7 @@ def _calculate(
 
     trials_of_interest = [trial for trial in trials if trial.state in states_of_interest]
 
-    next_cursor = trials[-1].number + 1 if len(trials) > 0 else -1
+    next_cursor = trials_of_interest[-1].number + 1 if len(trials_of_interest) > 0 else -1
     for trial in reversed(trials_of_interest):
         if cursor > trial.number:
             break
--- a/optuna/search_space/intersection.py
+++ b/optuna/search_space/intersection.py
@@ -3,10 +3,50 @@ import copy
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import Tuple
 
 import optuna
 from optuna.distributions import BaseDistribution
 from optuna.study import Study
+
+
+def _calculate(
+    trials: List[optuna.trial.FrozenTrial],
+    include_pruned: bool = False,
+    search_space: Optional[Dict[str, BaseDistribution]] = None,
+    cursor: int = -1,
+) -> Tuple[Optional[Dict[str, BaseDistribution]], int]:
+    states_of_interest = [
+        optuna.trial.TrialState.COMPLETE,
+        optuna.trial.TrialState.WAITING,
+        optuna.trial.TrialState.RUNNING,
+    ]
+
+    if include_pruned:
+        states_of_interest.append(optuna.trial.TrialState.PRUNED)
+
+    trials_of_interest = [trial for trial in trials if trial.state in states_of_interest]
+
+    next_cursor = trials[-1].number + 1 if len(trials) > 0 else -1
+    for trial in reversed(trials_of_interest):
+        if cursor > trial.number:
+            break
+
+        if not trial.state.is_finished():
+            next_cursor = trial.number
+            continue
+
+        if search_space is None:
+            search_space = copy.copy(trial.distributions)
+            continue
+
+        search_space = {
+            name: distribution
+            for name, distribution in search_space.items()
+            if trial.distributions.get(name) == distribution
+        }
+
+    return search_space, next_cursor
 
 
 class IntersectionSearchSpace:
@@ -50,7 +90,6 @@ class IntersectionSearchSpace:
 
         Returns:
             A dictionary containing the parameter names and parameter's distributions.
-
         """
 
         if self._study_id is None:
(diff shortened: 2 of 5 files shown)


toshihikoyanase:
＠hvy Could you review this PR, please? If you are busy, please feel free to remove the assignment.

optuna/importance/_base.py
```diff
@@ -72,7 +72,7 @@ def _get_distributions(study: Study, params: Optional[List[str]]) -> Dict[str, B
     _check_evaluate_args(completed_trials, params)
 
     if params is None:
-        return intersection_search_space(study, ordered_dict=True)
+        return intersection_search_space(study.trials, ordered_dict=True)
```
hvy, line 75: `study.trials` internally calls `study.get_trials` with `deepcopy=True`. This has some performance implications given that the earlier `IntersectionSearchSpace` called this method with `deepcopy=False`. Could you replace this`study.trials` with `study.get_trials(deepcopy=False)` , and similarly for other invocations?

optuna/search_space/intersection.py
```diff
@@ -17,11 +17,6 @@ class IntersectionSearchSpace:
     neither is included in the resulting search space
     (i.e., the parameters with dynamic value ranges are excluded).
 
-    Note that an instance of this class is supposed to be used for only one study.
```
hvy, line 20: I think this is still true since `self._cursor` and `self._search_space` are still associated with a particular study. I suggest we keep this line, and if to keep this PR no compat breaking, reintroduce the `raise ValueError("IntersectionSearchSpace cannot handle multiple studies.")` at the callers, e.g. in the samplers instances, store the study ID, and raise this error. This behavior itself is not so pretty in my opinion, and can be discussed, though. What do you think?

hvy (changes requested):
Thanks for the PR. I left some comments. Could you take a look?

optuna/search_space/intersection.py
toshihikoyanase: I agree. Regarding the current design, developers of user-defined samplers need to manage relationship between `IntersectionSearchSpace` instances and studies. Otherwise, `IntersectionSearchSpace.calculate` will return unexpected values silintly.

How about sticking to the `study` as an argument of `IntersectionSearchSpace.calculate`?
I roughly implemented a prototype in the commit https://github.com/optuna/optuna/commit/c8b8e0db5917d3860eee39a9555729d75d0b3f51.

cross32768: Thank you for your comments and insightful prototype! I fixed my PR to keep the study as an argument of `IntersectionSearchSpace.calculate` and change the argument of `intersection_search_space`.

 As a difference to prototype by @toshihikoyanase , I does not extract common interface from `IntersectionSearchSpace.calculate` and `intersection_search_space`. Instead, I implement some required part of of `IntersectionSearchSpace.calculate` to `intersection_search_space`. This violates DRY principle, but I think this change makes some code simpler. What do you think about this change?

optuna/importance/_base.py
cross32768: Thank you for your comment. I fixed this point.

optuna/search_space/intersection.py
toshihikoyanase: Thank you for your update. I'm a bit afraid of the maintenance of the two similar logics. Developers need to synchronize the change in both search space calculation logics, and the review cost may be doubled since they are similar, but not exactly the same.

What do you think of extracting a part of the calculation logic into a function?
The return type of `_calculate` in https://github.com/optuna/optuna/commit/c8b8e0db5917d3860eee39a9555729d75d0b3f51 was complicated since it created both external and internal search space. If we return the internal search space, code may be more readable.

https://github.com/optuna/optuna/compare/master...toshihikoyanase:optuna:ya-4514?expand=1

cross32768: Thank you for your comment. I update my code according to direction of extracting a part of the calculation logic. I implemented myself, but eventually, it becomes almost same as your prototype.

I have 2 points to consult.
1. This update changes the interface of `intersection_search_space` function. Now, the interface is different from `IntersectionSearchSpace.calculate`. So, should we change the name of this function? For example, my idea is `intersection_search_space_from_trials`.
2. (nits) When annotating a variable that can take None or other types, which is better to use Optional or |?

toshihikoyanase: > So, should we change the name of this function?

Good point! I don't think it is necessary to rename the function. This is because 
- we provide type hints, and modern development environments can notify developers
- I expect `intersection_search_space` if `intersection_search_space_from_trials` exists, but it may a bit overengineering

> (nits) When annotating a variable that can take None or other types, which is better to use Optional or |?

I we apply `from __future__ import annotations` to code, `|` seems to be selected as can be seen in #4526. But, I think we can work on it in a follow-up PR.

cross32768: Ok, I leave the function name as now.
And I updated my PR to use `|` instead of `Optional`.

toshihikoyanase: Thanks!

optuna/search_space/intersection.py
```diff
@@ -1,13 +1,55 @@
+from __future__ import annotations
+
 from collections import OrderedDict
 import copy
 from typing import Dict
-from typing import Optional
+from typing import List
```
toshihikoyanase, line 6: Since we introduce `__futuer__.annotations`, we can use `list` instead of  `typing.List`.
```suggestion
```

toshihikoyanase: `typing.Tuple` is still required according to https://github.com/optuna/optuna/pull/4526/files#r1148474747.

optuna/search_space/intersection.py
```diff
@@ -1,13 +1,55 @@
...
+from typing import Tuple
 
 import optuna
 from optuna.distributions import BaseDistribution
 from optuna.study import Study
 
 
+def _calculate(
+    trials: List[optuna.trial.FrozenTrial],
```
toshihikoyanase, line 15: ditto.
```suggestion
    trials: list[optuna.trial.FrozenTrial],
```

optuna/search_space/intersection.py
```diff
@@ -100,9 +116,11 @@ def calculate(self, study: Study, ordered_dict: bool = False) -> Dict[str, BaseD
 
 
 def intersection_search_space(
-    study: Study, ordered_dict: bool = False, include_pruned: bool = False
+    trials: List[optuna.trial.FrozenTrial],
```
toshihikoyanase, line 119: ditto.
```suggestion
    trials: list[optuna.trial.FrozenTrial],
```

optuna/search_space/intersection.py
```diff
@@ -1,13 +1,55 @@
...
+        optuna.trial.TrialState.WAITING,
+        optuna.trial.TrialState.RUNNING,
+    ]
+
+    if include_pruned:
+        states_of_interest.append(optuna.trial.TrialState.PRUNED)
+
+    trials_of_interest = [trial for trial in trials if trial.state in states_of_interest]
+
+    next_cursor = trials[-1].number + 1 if len(trials) > 0 else -1
```
toshihikoyanase, line 31: The original code checks the length of filtered trials.

```suggestion
    next_cursor = trials[-1].number + 1 if len(trials_of_interest) > 0 else -1
```

toshihikoyanase (commented):
The calculation of the next cursor seems different from the original code. Could you take a look?

optuna/search_space/intersection.py
```diff
@@ -1,13 +1,55 @@
...
 import optuna
 from optuna.distributions import BaseDistribution
 from optuna.study import Study
 
 
+def _calculate(
+    trials: List[optuna.trial.FrozenTrial],
+    include_pruned: bool = False,
+    search_space: Dict[str, BaseDistribution] | None = None,
+    cursor: int = -1,
```
hvy, line 18: `cursor` is not very descriptive in the context of optuna. Not necessarily addressed in this PR.
```suggestion
    last_seen_(largest_)trial_number: int = -1,
```

hvy (approved):
Thanks, personally LGTM after @toshihikoyanase 's comments.

I wasn't aware of the `|` type hint. 🙇 
This is outside the scope of this PR and just a comment but `_get_distributions` is sort of duplicating logic and might be open for improvement/DRY too.

optuna/search_space/intersection.py
```diff
@@ -1,13 +1,55 @@
...
 
 import optuna
 from optuna.distributions import BaseDistribution
 from optuna.study import Study
 
 
+def _calculate(
+    trials: List[optuna.trial.FrozenTrial],
+    include_pruned: bool = False,
+    search_space: Dict[str, BaseDistribution] | None = None,
```
hvy, line 17: If we address https://github.com/optuna/optuna/pull/4514/files#r1154026385
```suggestion
    last_seen_search_space: Dict[str, BaseDistribution] | None = None,
```

or something along `cached_search_space`.

optuna/search_space/intersection.py
gen740: Actually, it doesn't.

optuna/search_space/intersection.py
hvy: Should we also get the number from this the last filtered trials? Maybe we should fix tests.

optuna/search_space/intersection.py
gen740: What is unsupported on < python3.10 is not the `Tuple` but the `TypeAlias` feature. TypeAlian cannot be used with the `|` operator.

optuna/search_space/intersection.py
cross32768: I think there are room of discussion for this change. Because as shown below, we simplify implementation by using same variable name for search_space and I think it is not in line with reality to use the name `last_seen_search_space` or `cached_search_space` consistently in this function. https://github.com/optuna/optuna/blob/85312225325e1f690ec58a096d343a249479afcf/optuna/search_space/intersection.py#L32-L48

optuna/search_space/intersection.py
cross32768: Good catch! I changed the variable name `cursor` to `cached_trial_number` and `next_cursor` to `next_cached_trial_number` because we use also the variable name `next_cursor` and `next_last_seen_trial_number` is a bit confusing name. What do you think?

optuna/search_space/intersection.py
cross32768: Thank you for your suggestions, But, can I leave this topic to a follow-up PR?

optuna/search_space/intersection.py
cross32768: Ideally, I should add some tests, but it seems like it would be quite complicated, so can I leave this problem open in this PR?

hvy: You're right. Actually, current tests seem sufficient, and getting the number from `trials` seems fine in terms of correctness.

optuna/search_space/intersection.py
hvy: Looks good, thank you very much.

hvy (approved):
LGTM!

optuna/search_space/intersection.py
cross32768: Thank you. Let me leave it at this PR.

cross32768:
@toshihikoyanase Could you take another look?

toshihikoyanase (approved):

@hvy Thank you for creating a issue to track the following task about `OrderedDict`.

PR #4523, opened 2023-03-16
Increase time precision to check heart beat interval

toshihikoyanase:
## Motivation

`tests-mac` is promoted to be a required test in #[4458](https://github.com/optuna/optuna/pull/4458), but it sometimes failed. More specifically, the `test_record_heartbeat` failed at the following assertion:

https://github.com/optuna/optuna/blob/51d159fae12de3f5c143c4bd9e5d255e270d29fb/tests/storages_tests/rdb_tests/test_storage.py#L335-L337

https://github.com/optuna/optuna/actions/runs/4434158826/jobs/7779875352?pr=4522
> FAILED tests/storages_tests/rdb_tests/test_storage.py::test_record_heartbeat - assert ((4 - 2) - 0) <= 1

I think `trial_prep` and `heartbeats_interval` lose the subsecond information since they are `datetime.timedelta.second`. So, I'd like to increase their time precision.

## Description of the changes

- Use [`datetime.timedelta.total_seconds()`](https://docs.python.org/3/library/datetime.html#datetime.timedelta.total_seconds) instead of [`datetime.timedelta.second`](https://docs.python.org/3/library/datetime.html#datetime.timedelta)

### DIFF
--- a/tests/storages_tests/rdb_tests/test_storage.py
+++ b/tests/storages_tests/rdb_tests/test_storage.py
@@ -332,6 +332,6 @@ def test_record_heartbeat() -> None:
             datetime_start = trials[i + 1].datetime_start
             prev_datetime_complete = trials[i].datetime_complete
             assert datetime_start is not None and prev_datetime_complete is not None
-            trial_prep = (datetime_start - prev_datetime_complete).seconds
-            heartbeats_interval = (trial_heartbeats[i + 1] - trial_heartbeats[i]).seconds
+            trial_prep = (datetime_start - prev_datetime_complete).total_seconds()
+            heartbeats_interval = (trial_heartbeats[i + 1] - trial_heartbeats[i]).total_seconds()
             assert heartbeats_interval - sleep_sec - trial_prep <= 1


toshihikoyanase:
@eukaryo @gen740 Could you review this PR, please?

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -332,6 +332,6 @@ def objective(trial: Trial) -> float:
             datetime_start = trials[i + 1].datetime_start
             prev_datetime_complete = trials[i].datetime_complete
             assert datetime_start is not None and prev_datetime_complete is not None
-            trial_prep = (datetime_start - prev_datetime_complete).seconds
-            heartbeats_interval = (trial_heartbeats[i + 1] - trial_heartbeats[i]).seconds
+            trial_prep = (datetime_start - prev_datetime_complete).total_seconds()
+            heartbeats_interval = (trial_heartbeats[i + 1] - trial_heartbeats[i]).total_seconds()
             assert heartbeats_interval - sleep_sec - trial_prep <= 1
```
gen740, line 337: I realize that this value becomes negative.
What is this assertion intended to confirm?

toshihikoyanase: Thank you for your confirmation. I think negative values are not expected.
Maybe the scenario of this test should be revised.
I don't think this change resolves the error, so let me close it.

Issue #4508, opened 2023-03-10
Use `__future__.annotations` everywhere in the Optuna code base


HideakiImamura:
### Motivation

Optuna drops Python 3.6 from v3.1, so we can use `__future__.annotations`, which simplifies the code base. See [PEP 563](https://peps.python.org/pep-0563/), [PEP584](https://peps.python.org/pep-0584/), [PEP 585](https://peps.python.org/pep-0585/), and [PEP 604](https://peps.python.org/pep-0604/) for more details. This issue suggests to use the module and simplifies the code base.

### Suggestion

Use `__future__.annotations` for each file and simplify the type annotations. The list of classes whose type annotations can be simplified is [here](https://peps.python.org/pep-0585/#implementation). The list of files where the `__future__.annotations` can be used is as follows. In order to reduce review costs and to encourage more contributors to work on it, please, as a rule, fix one file per PR.

> [!NOTE]
> All the files were resolved, thank you everyone for the contributions!
> We are now waiting for Python 3.8 drop to address the files below.

### Files that cannot be updated now

> [!IMPORTANT]
> Currently (23 Dec 2024), Optuna supports Python 3.8, making it impossible to reduce some typing.
> For example, `|` cannot be used in a type declaration and we need to use `typing.Union` instead.
> These typings can only be eliminated after we drop Python 3.8 support.
> Please note that any other type annotations can be replaced as long as we have the line `from __future__ import annotations`.

The following files cannot be updated as long as we support Python 3.8.

- [ ] optuna/samplers/_grid.py
    - Cannot reduce `Union` in the type definition.
- [ ] optuna/samplers/_cmaes.py
    - Cannot reduce `Union` in the type definition.
- [ ] optuna/samplers/_tpe/probability_distributions.py
    - Cannot reduce `Union` in the type definition.
- [ ] optuna/study/study.py
    - Cannot reduce `Callable` and `Sequence` in the type definition.
- [ ] optuna/_typing.py
    - Cannot reduce `Union` in the type definition.
- [ ] optuna/trial/_frozen.py
    - Cannot reduce `Dict` in the casting.
- [ ] optuna/distributions.py
    - Cannot reduce `Union` in the type definition.
- [ ] tests/study_tests/test_study.py
    - Cannot reduce `Callable` in the type definition.

### Additional context (optional)

The above list is generated by the following script.

<details>
<summary>script</summary>

```python
import os
import pathlib


PATTERNS = [
    "from typing import Union",
[...]

PR #4525, opened 2023-03-18
Simplify annotations in `optuna.visualization`

harupy:
## Motivation

https://github.com/optuna/optuna/issues/4508

PR #4526, opened 2023-03-18
Simplify annotations in `tests.visualization_tests`

harupy:
## Motivation

#4508

### DIFF
--- a/tests/visualization_tests/test_contour.py
+++ b/tests/visualization_tests/test_contour.py
@@ -30,8 +30,6 @@ from optuna.visualization.matplotlib._matplotlib_imports import Axes
 from optuna.visualization.matplotlib._matplotlib_imports import plt
 
 
-RANGE_TYPE = Union[Tuple[str, str], Tuple[float, float]]
-
 parametrize_plot_contour = pytest.mark.parametrize(
     "plot_contour", [plotly_plot_contour, plt_plot_contour]
 )
--- a/tests/visualization_tests/test_contour.py
+++ b/tests/visualization_tests/test_contour.py
@@ -4,8 +4,6 @@ from io import BytesIO
 import math
 from typing import Any
 from typing import Callable
-from typing import Tuple
-from typing import Union
 
 import numpy as np
 import pytest
--- a/tests/visualization_tests/test_contour.py
+++ b/tests/visualization_tests/test_contour.py
@@ -1,10 +1,9 @@
+from __future__ import annotations
+
 from io import BytesIO
 import math
 from typing import Any
 from typing import Callable
-from typing import Dict
-from typing import List
-from typing import Optional
 from typing import Tuple
 from typing import Union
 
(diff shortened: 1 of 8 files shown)


c-bata (approved):
LGTM. I have the same question as #4525 regarding `typing.Optional` though.

c-bata:

tests/visualization_tests/test_contour.py
```diff
@@ -1,10 +1,9 @@
+from __future__ import annotations
+
 from io import BytesIO
 import math
 from typing import Any
 from typing import Callable
-from typing import Dict
-from typing import List
-from typing import Optional
 from typing import Tuple
```
eukaryo, line 7: I think it can be deleted.
```suggestion
```
In the case, line 33 should be changed as `RANGE_TYPE = Union[tuple[str, str], tuple[float, float]]`

tests/visualization_tests/test_param_importances.py
```diff
@@ -1,9 +1,8 @@
+from __future__ import annotations
+
 from io import BytesIO
 from typing import Any
 from typing import Callable
-from typing import List
-from typing import Optional
-from typing import Tuple
```
eukaryo, line 6: I think it can be deleted, too.
`Tuple` was used only in Line 193, and it has already been properly modified.

eukaryo (commented):
Thank you for your pull request! It is almost LGTM, and I have two comments.

tests/visualization_tests/test_param_importances.py
eukaryo: Sorry, it is already resolved. This comment was my mistake.

tests/visualization_tests/test_contour.py
harupy: @eukaryo Unfortunately, it can't be changed in Python.

```
> git diff
diff --git a/tests/visualization_tests/test_contour.py b/tests/visualization_tests/test_contour.py
index 4bcfc6d46..644bf6885 100644
--- a/tests/visualization_tests/test_contour.py
+++ b/tests/visualization_tests/test_contour.py
@@ -31,7 +31,7 @@ from optuna.visualization.matplotlib._matplotlib_imports import Axes
 from optuna.visualization.matplotlib._matplotlib_imports import plt
 
 
-RANGE_TYPE = Union[Tuple[str, str], Tuple[float, float]]
+RANGE_TYPE = Tuple[str, str] | Tuple[float, float]
 
 parametrize_plot_contour = pytest.mark.parametrize(
     "plot_contour", [plotly_plot_contour, plt_plot_contour]

> python tests/visualization_tests/test_contour.py
  File "tests/visualization_tests/test_contour.py", line 34, in <module>
    RANGE_TYPE = Tuple[str, str] | Tuple[float, float]
TypeError: unsupported operand type(s) for |: '_GenericAlias' and '_GenericAlias'
```

gen740: I confirmed that, `Union[tuple[str, str], tuple[float, float]]` is OK, not `tuple[str, str] | tuple[float, float]`.

gen740: What is unsupported on < python3.10 is the `TypeAlias` feature, so wrapping types into `Union` goes well.
And you can use `RANGE_TYPE = "tuple[str, str] | tuple[float, float]"` for type alias on < python3.10.

gen740: I think `RANGE_TYPE` does not use anywhere. Can't delete it?

harupy: @gen740 Great catch! Removed it

eukaryo: LGTM!

eukaryo (approved):

PR #4532, opened 2023-03-22
Securely deploy results of visual regression test on Netlify

eukaryo:
## Motivation

[PR#4507](https://github.com/optuna/optuna/pull/4507) does not work properly because a workflow triggered by `pull_request` cannot access secrets. I will use `pull_request_target` instead of it.

## Description of the changes

Use `pull_request_target` as a trigger instead of `pull_request`.
Implemented so that it is triggered only at the moment a PR is labeled `safe-to-run-visual-regression-tests`. This allows that the workflow can only be triggered by PRs that have been identified as safe.

### DIFF
--- a/.github/workflows/visual-regression.yml
+++ b/.github/workflows/visual-regression.yml
@@ -1,7 +1,8 @@
 name: Visual regression tests with Netlify
 
 on:
-  pull_request:
+  pull_request_target:
+    types: labeled
     paths:
       - 'optuna/visualization/**'
   workflow_dispatch:
@@ -15,8 +16,12 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    # Not intended for forks.
-    if: github.repository == 'optuna/optuna'
+    if: |
+      ( github.repository == 'optuna/optuna' ) && # Not intended for forks.
+      (
+         contains(github.event.pull_request.labels.*.name, 'safe-to-exec-visual-regression-test') ||
+         ( github.event_name == 'workflow_dispatch' )
+      )
 
     steps:
 
--- a/.github/workflows/visual-regression.yml
+++ b/.github/workflows/visual-regression.yml
@@ -15,11 +15,11 @@ jobs:
     runs-on: ubuntu-latest
 
     # Not intended for forks.
-    if: github.repository == 'optuna/optuna'
-    if:  |
+    if: |
+      ( github.repository == 'optuna/optuna' ) && # Not intended for forks.
       (
-         contains(github.event.pull_request.labels.*.name, 'safe-to-run-visual-regression-tests') ||
-         (github.event_name == 'workflow_dispatch')
+         contains(github.event.pull_request.labels.*.name, 'safe-to-exec-visual-regression-test') ||
+         ( github.event_name == 'workflow_dispatch' )
       )
 
     steps:
--- a/.github/workflows/visual-regression.yml
+++ b/.github/workflows/visual-regression.yml
@@ -14,11 +14,12 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    if: |
-      ( github.repository == 'optuna/optuna' ) && # Not intended for forks.
+    # Not intended for forks.
+    if: github.repository == 'optuna/optuna'
+    if:  |
       (
          contains(github.event.pull_request.labels.*.name, 'safe-to-run-visual-regression-tests') ||
-         ( github.event_name == 'workflow_dispatch' )
+         (github.event_name == 'workflow_dispatch')
       )
 
     steps:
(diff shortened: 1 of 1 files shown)


c-bata:
@toshihikoyanase @Alnusjaponica Could you review this PR if you have time?

.github/workflows/visual-regression.yml
```diff
@@ -15,8 +14,12 @@ jobs:
...
     runs-on: ubuntu-latest
 
-    # Not intended for forks.
-    if: github.repository == 'optuna/optuna'
+    if: |
+      ( github.repository == 'optuna/optuna' ) && # Not intended for forks.
+      (
+         contains(github.event.pull_request.labels.*.name, 'safe-to-run-visual-regression-tests') ||
+         ( github.event_name == 'workflow_dispatch' )
+      )
```
toshihikoyanase, line 22: Do you think it is possible to use an early return pattern here?

```suggestion
    # Not intended for forks.
    if: github.repository == 'optuna/optuna'
    if:  |
      (
         contains(github.event.pull_request.labels.*.name, 'safe-to-run-visual-regression-tests') ||
         (github.event_name == 'workflow_dispatch')
      )
```

.github/workflows/visual-regression.yml
```diff
@@ -1,9 +1,8 @@
 name: Visual regression tests with Netlify
 
 on:
-  pull_request:
-    paths:
-      - 'optuna/visualization/**'
+  pull_request_target:
+    types: [labeled]
```
toshihikoyanase, line 5: [Notes] The limitation of the path was removed, but it sounds reasonable to me since the workflows are manually triggered.

toshihikoyanase: [Notes] The document of `pull_request_target` is [this one](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target). I think the change follows the usage instruction.

toshihikoyanase (commented):
It basically looks good to me. The condition of this workflow seems to be a bit complicated and I'm thinking if we can refactor it.

toshihikoyanase:
If you have tested `pull_request_target` on your fork repository, could you share the URL of it, please?

eukaryo:
I see. Here is an example. https://github.com/eukaryo/optuna/pull/5
Since `github.repository == 'optuna/optuna'` prevents the tests in fork, I made another branch, comment-outed it, and made PR from yet another dummy branch.

.github/workflows/visual-regression.yml
eukaryo: Thank you for the additional clarifications.

toshihikoyanase:
Thank you for your reply. I guess you tested on https://github.com/eukaryo/optuna/pull/4, not https://github.com/eukaryo/optuna/pull/5, as far as I checked the diff of the CI settings.

In https://github.com/eukaryo/optuna/pull/4, I couldn't find the CI job entry about the visual regression test.

toshihikoyanase:
Ah, that makes sense. Thank you for your explanation.
I found that `visual-regression.yml` of https://github.com/eukaryo/optuna/pull/5 actually used `pull_request_target`.

https://github.com/eukaryo/optuna/blob/f4d63c2669bf5dba9d3baf126f05ada0d55be0bb/.github/workflows/visual-regression.yml#L18-L23

BTW, could you change the following line to `if: github.repository == 'eukaryo/optuna'` to confirm that the workflow setting accepts multiple if statement, please?

https://github.com/eukaryo/optuna/blob/f4d63c2669bf5dba9d3baf126f05ada0d55be0bb/.github/workflows/visual-regression.yml#L18

toshihikoyanase:
Thank you for your confirmation. I confirmed that we could not divide the if statement as mentioned in https://github.com/optuna/optuna/pull/4532#discussion_r1145565278. I confirmed that https://github.com/optuna/optuna/pull/4532/commits/e1834d4d632a951e00fd588cf533a4be159db805 and https://github.com/optuna/optuna/pull/4532/commits/a2d1146924848f5acc0069348f81f70fe95da328 reverted the change.


https://github.com/eukaryo/optuna/actions/runs/4508379960

toshihikoyanase:
Thanks for your update. We cannot comment by prefixing `#` in the if-statement. c.f., https://github.com/optuna/optuna/pull/4532/commits/1a5eb5157feeccc4e0639876498be6946c70bc01

toshihikoyanase (approved):
I confirmed that this workflow file could upload the test results as expected, as can be seen in https://github.com/eukaryo/optuna/pull/5#issuecomment-1480602306. LGTM!

Alnusjaponica (approved):

toshihikoyanase:
Let me merge this PR since it got two approvals.

PR #4556, opened 2023-03-30
Replace `hacking` with `flake8`

gen740:
## Motivation
"hacking" flake8 extension depends on `flake8 ~= 4.0.0` while latest version is
`6.0.0`. Due to this fixed version of flake8, latest flake8 linting fails on optuna
master branch.

## Description of the changes
- Drop `hacking`.
- Fix lint errors.

### DIFF
--- a/formats.sh
+++ b/formats.sh
@@ -15,7 +15,7 @@ if [ $? -eq 1 ] ; then
 fi
 command -v flake8 &> /dev/null
 if [ $? -eq 1 ] ; then
-  missing_dependencies+=(hacking)
+  missing_dependencies+=(flake8)
 fi
 command -v isort &> /dev/null
 if [ $? -eq 1 ] ; then
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -21,8 +21,8 @@ def _solve_hssp(
     - `Greedy Hypervolume Subset Selection in Low Dimensions
        <https://ieeexplore.ieee.org/document/7570501>`_
     """
-    selected_vecs = []  # type: List[np.ndarray]
-    selected_indices = []  # type: List[int]
+    selected_vecs: List[np.ndarray] = []
+    selected_indices: List[int] = []
     contributions = [
         optuna._hypervolume.WFG().compute(np.asarray([v]), reference_point)
         for v in rank_i_loss_vals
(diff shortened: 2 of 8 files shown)


cross32768:

toshihikoyanase:
Let me rename this PR since it does not simply removes `hacking`, but adds `flake8` instead.

toshihikoyanase:
Before diving into the details of the changes, I'd like to find out why optuna uses `hacking`  instead of `flake8`. If you have any thoughts or suggestions, please feel free to add them in the form of comments.

not522:
`hacking` was introduced very early in Optuna (#22). I remember that `hacking` was a popular lint at the time.

toshihikoyanase:
Thanks. `hacking` is also found at https://github.com/optuna/optuna/pull/2. The motivation was to follow PEP8, and the description did not mention OpenStack StyleGuide. So, I guess we don't need to stick to `hacking`.

toshihikoyanase:
I'm not sure which lint rules are removed after merging this PR.
@gen740 If you have time, could you list the actual changes in the lint rules, please?

toshihikoyanase:
> list the actual changes in the lint rules

I investigated the change.

The following packages were installed along with `hacking` and `flake8`:

| name | `pip install hacking` | `pip install flake8` | comments |
| --- | --- | --- | --- |
| [`flake8`](https://pypi.org/project/flake8/) | 4.0.1 | 6.0.0 | |
| [`mccabe`](https://pypi.org/project/mccabe/) | 0.6.1 | 0.7.0 |   |
| [`pycodestyle`](https://pypi.org/project/pycodestyle/) | 2.8.0 | 2.10.0 |  formally known as `pep8`. |
| [`pyflakes`](https://pypi.org/project/pyflakes/) | 2.4.0 | 3.0.1 | |

So, no flake8 plugins excepting `hacking` itself are removed by this PR.

The [`hacking` document](https://opendev.org/openstack/hacking#adding-additional-checks) suggests that additional checks comes from `pep8` plugin which is currently known as `pycodestyle`. After replacing `hacking` with `flake8`, the newer version of `pycodestyle` still exists.

I think this change will be beneficial to keep the coding style up-to-date. Please correct me if I'm wrong.

cross32768:
I also checked the difference of dependency between `flake8` and `hacking` and got the same results as the table above.
I have not been able to confirm if there are any checks included in `hacking` itself that cannot be covered by others and will fail silently, but I considering that `hacking` updates have been stopped for a long time, I think this update is basically good idea.

cross32768 (approved):

toshihikoyanase (approved):
LGTM!

PR #4591, opened 2023-04-09
Support batched sampling with BoTorch

kstoneriv3:
## Motivation

Currently, botorch integration of Optuna does not take the running trials into account when calculating the acquisition function. (By the way, in contrast, TPE has the `constant_lier` option that tries to leverage the information of running trials.) However, the original botorch API can leverage this information, and that seems to substantially improve parallel optimization performance. 

On a simple benchmark using 10-dimensional quadratic function (y = sum_{i=0, ..., 9} |x_i|^2 with -5 <= x_i <= 5), with n_trials = 40 and n_workers=10, the average objective values were 4.55 ± 0.22 (w/ batching) vs 10.30 ± 0.59 (w/o batching). The script for the benchmark can be found at [https://gist.github.com/kstoneriv3/1e877830e535d9ebea0f9d55ed50734c](https://gist.github.com/kstoneriv3/1e877830e535d9ebea0f9d55ed50734c).

## Description of the changes

I just need to change several lines of code at `optuna.integration.botorch` for passing the information about running trials to the BoTorch API. ~The test is WIP.~


contramundum53:
Changing the signature of `candidates_func` is considered a breaking change. 

I personally think that it is OK because 1. `BoTorchSampler` itself is experimental, 2. not many people seem to be using customized `candidates_func`, and 3. for those people who do customize `candidates_func`, it is very easy to fix.

But it needs to be discussed.

It is also possible to avoid breaking change by adding a try/catch or a condition on calling `candidates_func`. (Although I think it's a bit of over-engineering.)

tests/integration_tests/test_botorch.py
```diff
@@ -480,3 +488,56 @@ def constraints_func(trial: FrozenTrial) -> Sequence[float]:
...
+        (integration.botorch.qnehvi_candidates_func, 2),
+        (integration.botorch.qnehvi_candidates_func, 3),  # alpha > 0
+    ],
+)
+def test_botorch_consider_running_trials(candidates_func: Any, n_objectives: int) -> None:
+    sampler = BoTorchSampler(
+        candidates_func=candidates_func,
+        n_startup_trials=1,
+        consider_running_trials=True,
+    )
```
kstoneriv3, line 508: An issue with this test is that it alone takes about 75 sec in total, whereas all the other tests seem to take about 60 sec in total. Thus, I'd appreciate suggestions for trimming cases or different ways of testing.

kstoneriv3:
Thank you for the preliminary review! I added a test so the PR is ready for review.

kstoneriv3:
Following the comment by @contramundum53, I added a commit that would ensure backward compatibility. As discussed above, it seems to be a bit of overengineering and makes `sample_relative` even bigger. Please let me know whether you would like to have it or not.

optuna/integration/botorch.py
```diff
@@ -670,9 +670,20 @@ def sample_relative(
...
+            except TypeError:
+                if self._consider_running_trials:
+                    warnings.warn(
+                        f"The provided candidates_func {self._candidates_func.__name__} must take "
+                        "an argument pending_x to support option consider_running_trials=True. "
+                        "consider_running_trials option will be ignored."
+                    )
+                candidates = self._candidates_func(
+                    completed_params, completed_values, con, bounds
+                )  # type: ignore
```
kstoneriv3, line 686: Getting the mypy check right here would require even more lines of code, so I'd use `# type: ignore`.

optuna/integration/botorch.py
```diff
@@ -609,11 +658,32 @@ def sample_relative(
...
+                candidates = self._candidates_func(
+                    completed_params, completed_values, con, bounds, running_params
+                )
+            except TypeError:
+                if self._consider_running_trials:
+                    warnings.warn(
+                        f"The provided candidates_func {self._candidates_func.__name__} must take "
+                        "an argument pending_x to support option consider_running_trials=True. "
+                        "consider_running_trials option will be ignored."
+                    )
```
kstoneriv3, line 683: I am not sure which of `warnings.warn(...)` or `_logger.warning(...)` is the preferable way of warning here. Please let me know if you prefer one over the other.

optuna/integration/botorch.py
```diff
@@ -466,6 +488,10 @@ class BoTorchSampler(BaseSampler):
         n_startup_trials:
             Number of initial trials, that is the number of trials to resort to independent
             sampling.
+        consider_running_trials:
+            If True, the acquisition function takes into consideration the running parameters
+            whose evaluation has not completed. Enabling this option is considered to improve the
+            performance of parallel optimization.
```
HideakiImamura, line 494: It would be great to show this argument is experimental in the docstring.

optuna/integration/botorch.py
```diff
@@ -486,12 +512,14 @@ def __init__(
...
                     Optional["torch.Tensor"],
                     "torch.Tensor",
+                    Optional["torch.Tensor"],
                 ],
                 "torch.Tensor",
             ]
         ] = None,
         constraints_func: Optional[Callable[[FrozenTrial], Sequence[float]]] = None,
         n_startup_trials: int = 10,
+        consider_running_trials: bool = False,
```
contramundum53, line 522: My personal opinion is that making `consider_running_trials=True` as default is okay. What do you think, @HideakiImamura ?

optuna/integration/botorch.py
```diff
@@ -508,6 +537,13 @@ def __init__(
         self._search_space = IntersectionSearchSpace()
         self._device = device or torch.device("cpu")
 
+        if consider_running_trials:
+            warnings.warn(
+                "``consider_running_trials`` option is an experimental feature."
+                " The interface can change in the future.",
+                ExperimentalWarning,
+            )
+
```
contramundum53, line 546: `BoTorchSampler` is an experimental class already. I don't think we need this warning.

optuna/integration/botorch.py
```diff
@@ -542,10 +578,13 @@ def sample_relative(
...
 
-        trials = study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))
+        completed_trials = study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))
+        running_trials = study.get_trials(deepcopy=False, states=(TrialState.RUNNING,))
+        trials = completed_trials + running_trials
 
         n_trials = len(trials)
-        if n_trials < self._n_startup_trials:
+        n_completed_trials = len(completed_trials)
+        if n_completed_trials < self._n_startup_trials:
```
contramundum53, line 587: Do you think it's more consistent to use `n_trials < self._n_startup_trials` when `consider_running_trials=True`?

optuna/integration/botorch.py
```diff
@@ -609,11 +658,33 @@ def sample_relative(
...
+                if self._consider_running_trials:
+                    warnings.warn(
+                        f"The provided candidates_func {self._candidates_func.__name__} must take "
+                        "a positional argument pending_x to support option "
+                        "consider_running_trials=True of the BoTorchSampler. "
+                        "consider_running_trials option will be ignored."
+                    )
+                candidates = self._candidates_func(
+                    completed_params, completed_values, con, bounds
+                )  # type: ignore
```
contramundum53, line 687: After some discussions, we agreed that keeping the interface for `candidates_func` would be overengineering, so it's okay to simply write as such. 
```suggestion
            candidates = self._candidates_func(
                completed_params, completed_values, con, bounds, running_params
            )
```

optuna/integration/botorch.py
contramundum53:
```suggestion
        consider_running_trials: bool = True,
```

tests/integration_tests/test_botorch.py
HideakiImamura: Hmm.. Why this test take so long time? 

By the way, we does not execute the integration tests per PR. The integration tests are only executed as scheduled tests. See [here](https://github.com/optuna/optuna/blob/master/.github/workflows/tests-integration.yml#L121) for more details. So, I think we can accept this type of long test.

HideakiImamura (commented):

optuna/integration/botorch.py
contramundum53: @HideakiImamura I'd also like your opinion.

contramundum53 (commented):
Sorry for my late reply 🙇
I added some comments.

optuna/integration/botorch.py
contramundum53: Thank you for your attempt!

optuna/integration/botorch.py
kstoneriv3: I do not have a strong preference for this but it may slow down the parameter suggestion if the number of running trials is large(, though I did not observe any significant slowdown while testing it with the number of running trials up to 10). I will leave it to @HideakiImamura and @contramundum53.

optuna/integration/botorch.py
```diff
@@ -542,9 +573,14 @@ def sample_relative(
         if len(search_space) == 0:
             return {}
 
-        trials = study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))
+        completed_trials = study.get_trials(deepcopy=False, states=(TrialState.COMPLETE,))
+        running_trials = [
+            t for t in study.get_trials(deepcopy=False, states=(TrialState.RUNNING,)) if t != trial
+        ]
```
kstoneriv3, line 579: Just for your information, I was missing a detail here in the previous commit; the `running_trials` include the `trial` itself so I need to filter it out.

tests/integration_tests/test_botorch.py
kstoneriv3: OK, I will leave it as it is. 

As for the reason for the slowness, I do not know for sure but this is possibly because this test has a large number of actual BoTorch optimization for the suggested hyperparameters compared to other tests (5 test parameters * 8 iterations is relatively large).

optuna/integration/botorch.py
kstoneriv3: I added a note section saying that this is experimental.

kstoneriv3:
Thank you for the review! I should have addressed all of your reviews. PTAL

optuna/integration/botorch.py
HideakiImamura: I agree with this idea.

kstoneriv3: Fixed as suggested.

optuna/integration/botorch.py
HideakiImamura: I think it's OK to change the default behavior of this sampler since this is still an experimental.

contramundum53 (approved):
LGTM! Thanks for the PR!

HideakiImamura (approved):
LGTM.

PR #4594, opened 2023-04-11
Fix `metric_names` on `_log_completed_trial()` function

c-bata:
## Motivation

Follow-up #4383 

## Description of the changes

After merging #4383, I found a bug in `_log_completed_trial()` function. This PR fixes it.


HideakiImamura (approved):
Good catch! LGTM!

PR #4603, opened 2023-04-12
Get `system_attrs` lazily for samplers and pruners

not522:
## Motivation
This is part of the work to deprecate `_CachedStorage`.

When directly using `RDBStorage`, `_get_latest_trial` is a bottleneck because it reads `system_attrs` from the remote storage.

## Description of the changes
Getting `system_attrs` lazily. This change eliminates access to remote storage for `system_attrs` by samplers and pruners except `GridSampler`, `SuccessiveHalvingPruner`, and `HyperbandPruner`.

### DIFF
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -1,3 +1,6 @@
+from __future__ import annotations
+
+from collections import UserDict
 import copy
 import datetime
 from typing import Any
@@ -59,6 +62,7 @@ class Trial(BaseTrial):
             study, self._cached_frozen_trial
         )
         self._relative_params: Optional[Dict[str, Any]] = None
+        self._fixed_params = self._cached_frozen_trial.system_attrs.get("fixed_params", [])
 
     @property
     def relative_params(self) -> Dict[str, Any]:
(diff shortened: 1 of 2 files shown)


optuna/trial/_trial.py
```diff
@@ -752,3 +754,18 @@ def number(self) -> int:
...
+    def __init__(self, trial_id: int, storage: optuna.storages.BaseStorage) -> None:
+        super().__init__()
+        self._trial_id = trial_id
+        self._storage = storage
+        self._initialized = False
+
+    def __getattribute__(self, key: str) -> Any:
+        if key == "data":
+            if not self._initialized:
+                self._initialized = True
```
c-bata, line 769: Just wondering that do we need to make this thread-safe?

c-bata (commented):
My initial impression is that changes look almost perfect to me. Let me check the performance of _CachedStorage with this change.

toshihikoyanase:
@gen740 Could you review this PR, please?

optuna/trial/_trial.py
not522: I think it is not necessary because `_LazyTrialSystemAttrs` are deleted after returning from the sampler's or pruner's method, and the operation is idempotence.

c-bata: Sounds make sense. Thank you for clarification!

not522:
This PR was taken over by #4631.

PR #4609, opened 2023-04-14
Add `plot_terminator_improvement` as visualization of `optuna.terminator`

cross32768:
## Motivation
`optuna.terminator` needs the function to visualize terminator improvement. This PR adds a visualization function `plot_terminator_improvement` using `plotly`.

## Description of the changes
I add `plot_terminator_improvement` function.
It visualizes terminator improvement computed by `optuna.terminator`.
This function has 2 important features.
1. It can visualize cross validation error on the same figure. This allows to compare terminator improvement and cross validation error and help to understand termination by `optuna.terminator` (if line of cross validation error cross line of terminator improvement, `optuna.terminator` teminate its optimization).
2. It declare the area of `min_n_trials`. It is important because in this area, terminator improvement is not so reliable.


optuna/visualization/_regret_bound.py
```diff
@@ -0,0 +1,158 @@
...
+        return _RegretBoundInfo(
+            trial_numbers=trial_numbers, regret_bounds=regret_bounds, errors=None
+        )
+    else:
+        return _RegretBoundInfo(
+            trial_numbers=trial_numbers, regret_bounds=regret_bounds, errors=errors
+        )
+
+
+def _get_regret_bound_plot(info: _RegretBoundInfo, min_n_trials: int) -> "go.Figure":
```
g-votte, line 88: From a readability perspective, I think that there are too many tasks performed in this method, making it difficult to understand its logic at first glance. A single method should focus on a single responsibility (see also: [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)). How about defining a small method with clear and descriptive names for each individual sub-logic, like:
```python
def _get_regret_bound_plot(info: _RegretBoundInfo, min_n_trials: int) -> "go.Figure":
    ...
    
    add_regret_bounds_trace(
        fig, info.trial_numbers[:min_n_trials+1], info.regret_bounds[: min_n_trials + 1],
        n_trials, min_n_trials, plotly_blue_opacity,
    )

    add_regret_bounds_trace(
        fig, info.trial_numbers[min_n_trials:], info.regret_bounds[min_n_trials:],
        n_trials, min_n_trials, plotly_blue,
    )

    add_error_trace(fig, info, plotly_red)
    
    update_y_range(fig, info, min_n_trials)
    
    return fig
```

c-bata:

optuna/visualization/_regret_bound.py
cross32768: Thank you for your comment. I updated my code by introducing small function extracting some sub-logic. Could you take another look?

optuna/visualization/_regret_bound.py
```diff
@@ -0,0 +1,176 @@
...
+from optuna.visualization._plotly_imports import _imports
+
+
+if _imports.is_successful():
+    from optuna.visualization._plotly_imports import go
+
+_logger = get_logger(__name__)
+
+
+PADDING_RATIO = 0.05
```
g-votte, line 25: [nit] I think the name, `PADDING_RATIO`, is too general to be put in this scope. How about `PADDING_RATIO_Y`?

optuna/visualization/_regret_bound.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class _RegretBoundInfo(NamedTuple):
+    trial_numbers: list[int]
+    regret_bounds: list[float]
+    errors: list[float] | None
+
+
+@experimental_func("3.2.0")
+def plot_regret_bound(
```
g-votte, line 36: Shall we rename this function to `plot_improvement` or `plot_terminator_improvement`?

Indeed, the argument is not necessarily a `RegretBoundEvaluator` but an `ImprovementEvaluator`.

If we rename, almost all occurrences of "regret bound" in this file should be renamed as well.

optuna/visualization/_regret_bound.py
```diff
@@ -0,0 +1,176 @@
...
+        trials = study.trials[: i + 1]
+        trials = [t for t in trials if t.state == optuna.trial.TrialState.COMPLETE]
+
+        regret_bound = improvement_evaluator.evaluate(
+            trials=trials, study_direction=study.direction
+        )
+        regret_bounds.append(regret_bound)
+
+        if get_error:
+            error = error_evaluator.evaluate(trials=trials, study_direction=study.direction)
```
g-votte, line 75: [Question] Currently, the error handling seems to be delegated to the `ErrorEvaluator` when `plot_error` is `True` and `error_evaluator` is `None`. Do you think it is better to handle errors in this function?

optuna/visualization/_regret_bound.py
```diff
@@ -0,0 +1,176 @@
...
+        x=trial_numbers,
+        y=errors,
+        mode="markers+lines",
+        name="Error",
+        marker=dict(color=plotly_red),
+        line=dict(color=plotly_red),
+    )
+
+
+def _get_reasonable_y_range(info: _RegretBoundInfo, min_n_trials: int) -> tuple[float, float]:
```
g-votte, line 125: I prefer `_get_y_range` for the function name, because "reasonable" is redundant and ambiguous. We'll expect a "reasonable" range even if the function name is `_get_y_range`.

What do you think?

g-votte (changes requested):
@cross32768 
Sorry for a late response. I put a few additional comments. Could you please check them?

Also, would you mind giving us the code snippet or unit tests you used to verify this PR?

optuna/visualization/_regret_bound.py
g-votte: Thanks for your changes! Now, the function is much more readable for me.

optuna/visualization/_regret_bound.py
cross32768: I see. I think `_get_y_range` is better by your reason, and also it is shorter. I'll fix it.

optuna/visualization/_regret_bound.py
cross32768: In my opinion, current implementation is reasonable because `ErrorEvaluator` already handle such error appropriately and to handle errors for various `ErrorEvaluator` in this function is not so clear approach. What do you think?

cross32768:
@g-votte  Here is the code snippet that I use to test my implementation.
I have no unit test now, but I will add some unit tests by the follow-up PR (if you required, it is also capable to add unit test in this PR.)

```

from lightgbm import LGBMClassifier


import optuna
from optuna.terminator.erroreval import report_cross_validation_scores
from optuna.visualization import plot_regret_bound


def objective(trial):
    
    clf = LGBMClassifier(
        reg_alpha=trial.suggest_float("reg_alpha", 1e-8, 10.0, log=True),
        reg_lambda=trial.suggest_float("reg_lambda", 1e-8, 10.0, log=True),
        num_leaves=trial.suggest_int("num_leaves", 2, 256),
        colsample_bytree=trial.suggest_float("colsample_bytree", 0.4, 1.0),
        subsample=trial.suggest_float("subsample", 0.4, 1.0),
        subsample_freq=trial.suggest_int("subsample_freq", 1, 7),
        min_child_samples=trial.suggest_int("min_child_samples", 5, 100),
    )
    
    return scores.mean()


terminator = TerminatorCallback()

study = optuna.create_study()
study.optimize(objective, n_trials=50, )

fig = plot_regret_bound(study, plot_error=True)
fig.show()
```

cross32768:
This is example of the result of the above snippet.
スクリーンショット 2023-05-10 15 44 14

optuna/visualization/_regret_bound.py
cross32768: Thank you for your suggestion. I changed the function name to `plot_terminator_improvement`, and also updated all occurrences of "regret bound" to "improvement". Could you take another look?

optuna/visualization/_regret_bound.py
g-votte: Being nitpicky, the `BaseErrorEvaluator` class seems not to be responsible for the error handling when the error scores are not reported. The child class, `CrossValidationErrorEvaluator`, coincidentally handles it in the current implementation.

However, I think it's OK to ignore those details so far. We can revisit it when we implement another subclass of `BaseErrorEvaluator`.

g-votte (approved):
LGTM.

I'm not fully confident about edge cases, to be honest. Let's increase precision when we implement unit tests in a following pull request.

Alnusjaponica:
This is not directly related to the changes in this PR but it seems `terminator` is not used in the example from https://github.com/optuna/optuna/pull/4609#issuecomment-1541203793. The fourth line from the bottom should be 
```
study.optimize(objective, n_trials=50, callbacks=[terminator])
```

cross32768:
> Comment: This is not directly related to the changes in this PR but it seems `terminator` is not used in the example from [#4609 (comment)](https://github.com/optuna/optuna/pull/4609#issuecomment-1541203793). The fourth line from the bottom should be as follows:
> 
> ```
> study.optimize(objective, n_trials=50, callbacks=[terminator])
> ```

Yes, we can specify `terminator` object as callbacks. But I think it is not a mistake of sample snippet code because this visualization function can be used without specifying callback.

optuna/visualization/_terminator_improvement.py
```diff
@@ -0,0 +1,178 @@
...
+        error_evaluator = CrossValidationErrorEvaluator()
+
+    trial_numbers = []
+    improvements = []
+    errors = []
+
+    for i, trial in enumerate(tqdm.tqdm(study.trials)):
+        trial_numbers.append(trial.number)
+        trials = study.trials[: i + 1]
+        trials = [t for t in trials if t.state == optuna.trial.TrialState.COMPLETE]
```
Alnusjaponica, line 67: Does trial state change during this for loop? if not, the following change seems to make it faster and to provide the same results.
```suggestion
    trials = []
    for i, trial in enumerate(tqdm.tqdm(study.trials)):
        if trial.state == optuna.trial.TrialState.COMPLETE:
            trial_numbers.append(trial.number)
            trials.append(trial)
```

optuna/visualization/_terminator_improvement.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+    if n_trials == 0:
+        _logger.warning("There are no complete trials.")
+        return fig
+
+    fig.add_trace(
+        _get_improvement_scatter(
+            info.trial_numbers[: min_n_trials + 1],
+            info.improvements[: min_n_trials + 1],
+            OPACITY,  # Plot line for values below min_n_trials by light color.
```
Alnusjaponica, line 162: "values below min_n_trials" might be misleading. Please ignore this comment when you don't agree with.
```suggestion
            # Plot a line with a lighter color before the number of trials reaches min_n_trials.
            OPACITY,
```

Alnusjaponica:
LGTM except [this suggenstion](https://github.com/optuna/optuna/pull/4609#discussion_r1190650762). 

This is just a comment, but it seems that `_get_improvement_scatter()` and `_get_error_scatter()` can be merged into a single function introducing proper class like `_ImprovemrntPlotInfo(NamedTuple)`.

optuna/visualization/_terminator_improvement.py
Alnusjaponica: I realized that the length of `trial_numbers` and `improvements` can be different in the original code, when study contains trials whose state is not complete. Is that on purpose?

cross32768: I think in the original code,  the length of `trial_numbers` and `improvements` is always same. Even if study contains trials whose state is not complete, the length of `improvements` is increased as the same timing as `trial_numbers`. The computation of `improvements` by `improvement_evaluator` is conducted only by completed trials, but it is on purpose.

cross32768: The first point you said is right. Thank you for your good catch! I updated my code according to your pointing.

optuna/visualization/_terminator_improvement.py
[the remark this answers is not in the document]
cross32768: Thank you. I applied this suggestion.

Alnusjaponica:
I just left a small [suggestion](https://github.com/optuna/optuna/pull/4609#discussion_r1191456797) and that would be the last one.

By the way, I still don't understand why there's no problem even though
```
improvement_evaluator.evaluate(trials=completed_trials, study_direction=study.direction)
```
returns different value with the same input, such as observed in the following example. Is it easy to explain?
```
import numpy as np
from sklearn.datasets import load_iris
from sklearn.linear_model import SGDClassifier
from sklearn.model_selection import train_test_split

import optuna
from optuna.terminator.erroreval import report_cross_validation_scores
from optuna.visualization import plot_terminator_improvement

X, y = load_iris(return_X_y=True)
X_train, X_valid, y_train, y_valid = train_test_split(X, y)
classes = np.unique(y)
n_train_iter = 100

def objective(trial):
    alpha = trial.suggest_float("alpha", 0.0, 1.0)
    clf = SGDClassifier(alpha=alpha)
    for step in range(n_train_iter):
        clf.partial_fit(X_train, y_train, classes=classes)

        intermediate_value = clf.score(X_valid, y_valid)
        trial.report(intermediate_value, step)

        if trial.should_prune():
            raise optuna.TrialPruned()

    return scores.mean()

study = optuna.create_study(
    direction="maximize",
    pruner=optuna.pruners.HyperbandPruner(
        min_resource=1, max_resource=n_train_iter, reduction_factor=3
    ),
)
study.optimize(
    objective,
    n_trials=30,
)
fig = plot_terminator_improvement(study, plot_error=True)
fig.show()
```

optuna/visualization/_terminator_improvement.py
```diff
@@ -0,0 +1,182 @@
...
+        improvement_evaluator = RegretBoundEvaluator()
+    if error_evaluator is None:
+        error_evaluator = CrossValidationErrorEvaluator()
+
+    trial_numbers = []
+    completed_trials = []
+    improvements = []
+    errors = []
+
+    for i, trial in enumerate(tqdm.tqdm(study.trials)):
```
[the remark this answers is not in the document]
cross32768: Oh, I overlooked it. Thank you. I fixed it

cross32768:
I think `RegretBoundEvaluator` may return different evaluated value even inputs are same because random input is added ([ref](https://github.com/optuna/optuna/blob/master/optuna/terminator/improvement/evaluator.py#LL80C37-L80C37)).

Alnusjaponica:
I get it. Thanks! LGTM.

Alnusjaponica (approved):
I get it. Thanks! LGTM.

Alnusjaponica:
@toshihikoyanase Could you merge this PR?

toshihikoyanase:
Thank you for your review. Let me merge this PR.

PR #4610, opened 2023-04-14
Fix targets of `black` and `isort` in `formats.sh`

not522:
## Motivation
`black` and `isort` in `formats.sh` wrongly reformat all files.

## Description of the changes
Specify the targets of `black` and `isort`.


toshihikoyanase (approved):
Good catch! LGTM. Thank you.

toshihikoyanase:
Let me approve this PR with a single approval since it does not update the Optuna module itself and the changes are obvious.

PR #4615, opened 2023-04-16
Add `sphinxcontrib.jquery` extension to `conf.py`

Alnusjaponica:
## Motivation
In the latest optuna document, the search result page shows "Searching..." but that it never ends. In addition, the menu tab in the lower left part doesn't pull up. 
This PR resolves these problems.

## Description of the changes

As written in https://github.com/readthedocs/sphinx_rtd_theme/issues/1451, the combination of the latest versions of  `sphinx` and `sphinx_rtd_theme` causes error in search result page. This PR adds `sphinxcontrib.jquery` extension in `conf.py` to make `jQuery` loaded.


Alnusjaponica:
This workaround might be possible to remove after https://github.com/readthedocs/sphinx_rtd_theme/issues/1452 is resolved.

docs/source/conf.py
```diff
@@ -55,6 +55,7 @@
     "sphinx.ext.napoleon",
     "sphinx.ext.viewcode",
     "sphinx.ext.githubpages",
+    "sphinxcontrib.jquery",
```
toshihikoyanase, line 58: As you mentioned in https://github.com/optuna/optuna/pull/4615#issuecomment-1510024686, this fix is a workaround for https://github.com/readthedocs/sphinx_rtd_theme/issues/1452. How about adding a TODO comment to remove it after resolving the issue?

toshihikoyanase (commented):
I confirmed that the rendered reference worked as expected, including the search window and the version selection. I have a small suggestion.

not522 (approved):
LGTM except for @toshihikoyanase's comment.

docs/source/conf.py
Alnusjaponica: Thank you for your indication. I added a TODO comment.

toshihikoyanase (approved):

PR #4620, opened 2023-04-18
Delete `import Rectangle` in `visualization/matplotlib`

eukaryo:
## Motivation

In `optuna/visualization/matplotlib/_matplotlib_imports.py`, `Rectangle` is imported but not used.

## Description of the changes

Delete codes to import `Rectangle` in `visualization/matplotlib`.


toshihikoyanase (approved):
Good catch!

I confirmed that `Rectangle` is not used in Optuna code base. I don't think it is used by library users because `_matplotlib_imports.py` is private. So, I'd like to approve this change. Thank you.

toshihikoyanase:
Let me merge this PR with a single approval since the change is obvious.

PR #4622, opened 2023-04-19
Replace Duplicate Code in `PyTorchLightningPruningCallback`

YJWon99:
## Motivation

Current implementation of `PyTorchLightningPruningCallback` involves a few duplicate code when dealing with a `trial`'s `system_attr`. Specifically, access and modification of a `trial`'s `system_attr` is done by explicitly accessing the attributes of the `trial`. However, the same could be done by simply accessing `trial.system_attrs` and `trial.set_system_attr`. This pull request resolves this duplicate implementation, and consequently hides the inner implementation of `trial`, while also improving readability.

## Description of the changes

1. Replace `self._trial.storage.set_trial_system_attr` calls with `self._trial.set_system_attr` calls.
2. Replace current `_trial_system_attrs` access with direct access to `self._trial.system_attrs`.


YJWon99:
Thank you for the detailed explanation. I misunderstood the context behind the current implementation. Thank you!

PR #4626, opened 2023-04-20
Fix `ImportError` for `botorch<=0.4.0`

not522:
## Motivation
`botorch.version.version_tuple` was introduced in BoTorch v0.5.0. If BoTorch v0.4.0 is installed, `from optuna.integration import BoTorchSampler` raises an incomprehensible exception.

```
$ python -c "from optuna.integration import BoTorchSampler"
  File "<string>", line 1, in <module>
ImportError: cannot import name 'BoTorchSampler' from 'optuna.integration' (unknown location)
```

## Description of the changes
Use ~`pkg_resources.parse_version`~ `packaging.version` instead of `version_tuple`

### DIFF
--- a/optuna/integration/botorch.py
+++ b/optuna/integration/botorch.py
@@ -8,7 +8,7 @@ from typing import Union
 import warnings
 
 import numpy
-from pkg_resources import parse_version
+from packaging import version
 
 from optuna import logging
 from optuna._experimental import experimental_class
@@ -39,7 +39,7 @@ with try_import() as _imports:
     from botorch.sampling import SobolQMCNormalSampler
     import botorch.version
 
-    if parse_version(botorch.version.version) < parse_version("0.8.0"):
+    if version.parse(botorch.version.version) < version.parse("0.8.0"):
         from botorch.fit import fit_gpytorch_model as fit_gpytorch_mll
 
         def _get_sobol_qmc_normal_sampler(num_samples: int) -> SobolQMCNormalSampler:
--- a/optuna/integration/botorch.py
+++ b/optuna/integration/botorch.py
@@ -8,6 +8,7 @@ from typing import Union
 import warnings
 
 import numpy
+from pkg_resources import parse_version
 
 from optuna import logging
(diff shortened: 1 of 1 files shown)


optuna/integration/botorch.py
```diff
@@ -8,6 +8,7 @@
 import warnings
 
 import numpy
+from pkg_resources import parse_version
```
toshihikoyanase, line 11: Just a curiosity, but some other modules use `packaging.version` instead of `pkg_resouses.parse_version`. I

https://github.com/optuna/optuna/blob/906f6a6f2bf8d0e0700ece482c1b666c636076c7/optuna/visualization/matplotlib/_matplotlib_imports.py#L1
https://github.com/optuna/optuna/blob/906f6a6f2bf8d0e0700ece482c1b666c636076c7/optuna/visualization/_plotly_imports.py#L1

Could you tell me why you chose `pkg_resource` in this case?

toshihikoyanase (commented):
Thank you for your PR. The change basically looks good to me, but I have a small question.

toshihikoyanase:
@gen740 Could you join the review, please?

optuna/integration/botorch.py
not522: I have no particular reason. I think both are OK.

gen740: [pkg_resource](https://github.com/pypa/pkg_resources/tree/main) package is no more active while [packaging](https://github.com/pypa/packaging) is active. I'm not sure, but it would be better to use `packaging`.

not522: `pkg_resources` seems to be developing in `setuptools`. https://github.com/pypa/setuptools/tree/main/pkg_resources
By the way, unifying with `packaging` could lower maintenance costs later on. I'll update this PR.

gen740 (approved):
LGTM!

toshihikoyanase (approved):
Sorry for the delayed response. LGTM!

PR #4547, opened 2023-03-27
Use BoTorch's `optimize_acqf` for min-LCB evaluation in optuna terminator

contramundum53:
## Motivation
The original sampling scheme for min-LCB evaluation results in noisy and underestimated regrets, especially when the dimension is high. This causes premature termination and is highly undesirable.

We address this issue by using BoTorch's `optimize_acqf` function.

## Description of the changes
* Change `min_lcb` evaluation algorithm.
* Since BoTorch's `optimize_acqf` function assumes BoTorch's GP model, we entirely remove the abstraction of `BaseGaussianProcess` and simply write all code in `RegretBoundEvaluator`. This simplifies much of the abstraction.

PR #4633, opened 2023-04-25
Better categorical support for Optuna Terminator

contramundum53:
# Motivation
#4547 does not support CategoricalDistributions well. In particular, it does not enforce the one-hot constraint on `x` and the `optimize_acqf` function can give a solution like `[1, 1, ..., 1]`. This is highly undesirable because it results in unnecessarily high regret bound. 

Enforcing one-hot constraint is hard, because the optimization becomes a mixed-integer optimization. We instead relax the one-hot constraint and enforce `sum=1` (simplex constraint).


contramundum53:
Unfortunately, NASBench reveals that this results in a bad optimization result.

image

The original feasible region is the vertices and the optimal solution should be the vertex labeled "4", but relaxation finds the (infeasible) solution between "2" and "3".

g-votte:
For the same reason as stated in https://github.com/optuna/optuna/pull/4547#issuecomment-1524274571, let's close this PR for now and revisit it later after devising solutions. Thank you @contramundum53  for organizing the remaining problems!

PR #4635, opened 2023-04-25
Ensure that the plotly version of timeline plot draws a legend even if all TrialStates are the same.

eukaryo:
## Motivation

I found that if all `TrialStates` are `COMPLETED`, the matplotlib version of timeline plot displays the legend, but the plotly version does not unless explicitly indicated.  I want to make the behavior of both versions the same.

## Description of the changes

Added explicit indication to always display legend(s).

||plotly|matplotlib|
|---|---|---|
|before|  test-fig-plotly-old-1state | test-fig-matplotlib-old-1state |
|after|test-fig-plotly-new-1state|test-fig-matplotlib-new-1state|
|after|test-fig-plotly-new-3states|test-fig-matplotlib-new-3states|

The last row is provided to confirm that this PR does not cause degradation.

### DIFF
--- a/optuna/visualization/_timeline.py
+++ b/optuna/visualization/_timeline.py
@@ -140,4 +140,5 @@ def _get_timeline_plot(info: _TimelineInfo) -> "go.Figure":
             yaxis={"title": "Trial"},
         )
     )
+    fig.update_layout(showlegend=True)  # Draw a legend even if all TrialStates are the same.
     return fig


toshihikoyanase (approved):
Good catch! I confirmed that the legend was shown as expected.

Before:
image

After:
image

contramundum53 (approved):
LGTM!

Issue #4557, opened 2023-03-30
OptunaSearchCV.error_score not working


jckkvs:
### Expected behavior

In my understanding, OptunaSearchCV's error_score is a setting for ignoring errors during fit.

sklearn's GridSearchCV can also set error_score.

For example, the n_components of PLSRegression need to be changed according to the dataset, so I think it is necessary to carefully set the maximum value of param_range or param_distributions.

Here, I verified the operation of GridSearchCV and OptunaSearchCV.

Is this by design, or is it a bug that behaves differently than expected?

I would have expected the best_params_ to be sought even if param_distributions were specified in numbers greater than n_features and some trials were not possible.

### Environment

- Optuna version:3.1.0
- Python version:3.10
- OS:windows 10
- sklearn version:1.2.1

### Error messages, stack traces, or logs

```shell
[I 2023-03-30 11:47:51,179] A new study created in memory with name: no-name-537777eb-4267-4ec4-90a0-c178cbbe369e
[W 2023-03-30 11:47:51,194] Trial 0 failed with parameters: {'n_components': 189} because of the following error: ValueError('\nAll the 10 fits failed.\nIt is very likely that your model is misconfigured.\nYou can try to debug the error by setting error_score=\'raise\'.\n\nBelow are more details about the failures:\n--------------------------------------------------------------------------------\n10 fits failed with the following error:\nTraceback (most recent call last):\n  File "C:\\Users\\11665307\\Anaconda3\\envs\\py310\\lib\\site-packages\\sklearn\\model_selection\\_validation.py", line 686, in _fit_and_score\n    estimator.fit(X_train, y_train, **fit_params)\n  File "C:\\Users\\11665307\\Anaconda3\\envs\\py310\\lib\\site-packages\\sklearn\\cross_decomposition\\_pls.py", line 660, in fit\n    super().fit(X, Y)\n  File "C:\\Users\\11665307\\Anaconda3\\envs\\py310\\lib\\site-packages\\sklearn\\cross_decomposition\\_pls.py", line 251, in fit\n    raise ValueError(\nValueError: `n_components` upper bound is 100. Got 189 instead. Reduce `n_components`.\n').
Traceback (most recent call last):
  File "C:\Users\11665307\Anaconda3\envs\py310\lib\site-packages\optuna\study\_optimize.py", line 200, in _run_trial
    value_or_values = func(trial)
[...]

PR #4642, opened 2023-04-27
Solve issue #4557 - error_score

jckkvs:
## Motivation

solve https://github.com/optuna/optuna/issues/4557#issue-1646828230
## Description of the changes

The current implementation passes self.error_score to cross_validate.
When error_score is np.nan, sklearn's _warn_or_raise_about_fit_failures determines whether all fits succeed. This is a judgment whether **all fits in each trial** of OptunaSearchCV were successful.
However, it should be judged whether **all fits in all trials** of OptunaSearchCV were successful, 
that's why the current implementation is wrong.

Therefore, by setting np.nan in scores when cross_validate fails, we can realized the desired function.

### DIFF
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -897,8 +897,6 @@ class OptunaSearchCV(BaseEstimator):
             callbacks=self.callbacks,
         )
 
-        self.study_.best_params
-
         _logger.info("Finished hyperparemeter search!")
 
         if self.refit:
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -231,7 +231,9 @@ class _Objective:
                 n_splits = self.cv.get_n_splits(self.X, self.y, self.groups)
                 fit_time = np.array([np.nan] * n_splits)
                 score_time = np.array([np.nan] * n_splits)
-                test_score = np.array([self.error_score] * n_splits)
+                test_score = np.array(
+                    [self.error_score if self.error_score is not None else np.nan] * n_splits
+                )
 
                 scores = {
                     "fit_time": fit_time,
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -227,7 +227,7 @@ class _Objective:
                     return_train_score=self.return_train_score,
                     scoring=self.scoring,
                 )
-            except:
+            except ValueError:
                 n_splits = self.cv.get_n_splits(self.X, self.y, self.groups)
                 fit_time = np.array([np.nan] * n_splits)
                 score_time = np.array([np.nan] * n_splits)
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -215,17 +215,29 @@ class _Objective:
         if self.enable_pruning:
             scores = self._cross_validate_with_pruning(trial, estimator)
         else:
-            scores = cross_validate(
-                estimator,
-                self.X,
-                self.y,
-                cv=self.cv,
-                error_score=self.error_score,
-                fit_params=self.fit_params,
-                groups=self.groups,
-                return_train_score=self.return_train_score,
-                scoring=self.scoring,
-            )
+            try:
+                scores = cross_validate(
+                    estimator,
+                    self.X,
+                    self.y,
+                    cv=self.cv,
+                    error_score=self.error_score,
+                    fit_params=self.fit_params,
+                    groups=self.groups,
+                    return_train_score=self.return_train_score,
+                    scoring=self.scoring,
+                )
+            except:
+                n_splits = self.cv.get_n_splits(self.X, self.y, self.groups)
+                fit_time = np.array([np.nan] * n_splits)
+                score_time = np.array([np.nan] * n_splits)
+                test_score = np.array([np.nan] * n_splits)
+
+                scores = {
+                    "fit_time": fit_time,
+                    "score_time": score_time,
+                    "test_score": test_score,
+                }
 
         self._store_scores(trial, scores)
 
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -888,7 +888,7 @@ class OptunaSearchCV(BaseEstimator):
             "Searching the best hyperparameters using {} "
             "samples...".format(_num_samples(self.sample_indices_))
         )
-
+        
         self.study_.optimize(
             objective,
             n_jobs=self.n_jobs,
@@ -897,6 +897,8 @@ class OptunaSearchCV(BaseEstimator):
             callbacks=self.callbacks,
         )
 
+        _logger.info(f"best_params:{self.study_.best_params}")
+        
         _logger.info("Finished hyperparemeter search!")
 
         if self.refit:
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -888,7 +888,7 @@ class OptunaSearchCV(BaseEstimator):
             "Searching the best hyperparameters using {} "
             "samples...".format(_num_samples(self.sample_indices_))
         )
-        
+
         self.study_.optimize(
             objective,
             n_jobs=self.n_jobs,
@@ -900,7 +900,6 @@ class OptunaSearchCV(BaseEstimator):
         _logger.info(f"best_params:{self.study_.best_params}")
 
         _logger.info("Finished hyperparemeter search!")
-
 
         if self.refit:
             self._refit(X, y, **fit_params)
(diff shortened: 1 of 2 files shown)


contramundum53:

contramundum53:

HideakiImamura (commented):
Thanks for the PR! Changes looks good to me. By the way, Can we validate the changes in unitetests? It would be great to add a test case.

jckkvs:
Thank you for reviewing this PR.
I agree to add a test case. However, I am not familiar with the test case method used in optuna, so it takes time.

Is the test case necessary to merge the PR or not?

jckkvs:
Add tests.

HideakiImamura:
Sorry for the late. It would be great to add tests in this PR since it makes clear to validate the changes.

jckkvs:
@HideakiImamura 
Thanks. 
I try to know wht the test failed in python 3.7. Please let me know if you have any insight

HideakiImamura:
The specific versions of `scikit-learn` does not cause the error reported in #4557. So, you need to fix the test more simpler or skip testing when Python is 3.7. I prefer the former. 

How about creating the dummy estimator which raises an ValueError anytime? It simplifies the test.

jckkvs:
@HideakiImamura 
Thanks. I will try later.

But..
> How about creating the dummy estimator which raises an ValueError anytime? It simplifies the test.

I don't think that's enough.
As I mentioned in my first comment, it should be a ValueError when all fits of all trials in OptunaSearchcv fail.
On the other hand, we should not raise a ValueError if all fits in a trial fail.


For example, if n_features is 10,
"n_components": should raise ValueError if low=15, high=20, but
If "n_components": low=5, high=20 then n_components between 5 and 10 should not raise an Error.

The dummy estimator you proposed can test the former case, but not the latter.

Of course we could consider a dummy estimator for the latter case, but I doubt that such a complex dummy estimator would outperform sklearn's estimator.

optuna/integration/sklearn.py
```diff
@@ -885,6 +897,8 @@ def fit(
             callbacks=self.callbacks,
         )
 
+        self.study_.best_params
+
```
Alnusjaponica, line 901: nit

```suggestion
```

tests/integration_tests/test_sklearn.py
```diff
@@ -302,6 +312,34 @@ def test_objective_error_score_invalid() -> None:
...
+        param_dist,
+        cv=3,
+        error_score=np.nan,
+        random_state=0,
+    )
+
+    if expect:
+        with pytest.raises(ValueError):
+            optuna_search.fit(X, y)
+
```
Alnusjaponica, line 338: nit

```suggestion
```

tests/integration_tests/test_sklearn.py
```diff
@@ -302,6 +312,34 @@ def test_objective_error_score_invalid() -> None:
...
+def test_objective_error_score_nan_regression(param_dist: dict, expect: bool) -> None:
+    X, y = make_regression(n_samples=100, n_features=10)
+    est = DecisionTreeRegressor()
+    optuna_search = integration.OptunaSearchCV(
+        est,
+        param_dist,
+        cv=3,
+        error_score=np.nan,
+        random_state=0,
+    )
```
Alnusjaponica, line 333: I guess what you want to do is as follows:

```suggestion
        ({"max_depth": distributions.IntDistribution(0, 1)}, False),
        ({"max_depth": distributions.IntDistribution(0, 0)}, True),
    ],
)
@pytest.mark.filterwarnings("ignore::RuntimeWarning")
@pytest.mark.filterwarnings("ignore::UserWarning")
def test_objective_error_score_nan_regression(param_dist: dict, expect: bool) -> None:
    X, y = make_regression(n_samples=100, n_features=10)
    est = DecisionTreeRegressor()
    study = optuna.create_study()
    optuna_search = integration.OptunaSearchCV(
        est,
        param_dist,
        cv=3,
        error_score=np.nan,
        random_state=0,
        study=create_study(sampler=BruteForceSampler(), direction="maximize"),
    )
```

tests/integration_tests/test_sklearn.py
```diff
@@ -5,12 +5,14 @@
...
 from sklearn.decomposition import PCA
 from sklearn.exceptions import ConvergenceWarning
 from sklearn.exceptions import NotFittedError
 from sklearn.linear_model import LogisticRegression
 from sklearn.linear_model import SGDClassifier
 from sklearn.neighbors import KernelDensity
+from sklearn.tree import DecisionTreeRegressor
 
 from optuna import distributions
 from optuna import integration
```
Alnusjaponica, line 18: Please remember to import `BruteForceSampler` if you accept the suggestion in `test_objective_error_score_nan_regression`.

```suggestion
from optuna import integration
from optuna.samplers import BruteForceSampler
```

optuna/integration/sklearn.py
```diff
@@ -215,17 +215,29 @@ def __call__(self, trial: Trial) -> float:
...
+                    fit_params=self.fit_params,
+                    groups=self.groups,
+                    return_train_score=self.return_train_score,
+                    scoring=self.scoring,
+                )
+            except ValueError:
+                n_splits = self.cv.get_n_splits(self.X, self.y, self.groups)
+                fit_time = np.array([np.nan] * n_splits)
+                score_time = np.array([np.nan] * n_splits)
+                test_score = np.array([np.nan] * n_splits)
```
Alnusjaponica, line 234: The element of `test_score` should be `self.error_score` if it is not `None`, isn't it?

Alnusjaponica (commented):
Sorry for my delayed reply. I added some comment on the suggestion and I hope these helps you.

Alnusjaponica (commented):
Sorry for my delayed reply. I added some comment on the suggestion and I hope these helps you.

Alnusjaponica (commented):
Sorry for my delayed reply. I added some comment on the suggestion and I hope these helps you.

optuna/integration/sklearn.py
Alnusjaponica: @jckkvs 
You need to consider the case that `self.error_score` is `None`.

tests/integration_tests/test_sklearn.py
```diff
@@ -302,6 +313,31 @@ def test_objective_error_score_invalid() -> None:
         optuna_search.fit(X)
 
 
+@pytest.mark.parametrize(
+    "param_dist, expect",
+    [
+        ({"max_depth": distributions.IntDistribution(0, 1)}, False),
+        ({"max_depth": distributions.FloatDistribution(0, 1)}, True),
```
Alnusjaponica, line 320: Are there any reason to use `FloatDistribution`? `IntDistribution(0, 0)` seems enough to cause error in all trials.
BTW, `FloatDistribution` without `step` should be not acceptable for `BruteForceSampler` (see [docs](https://optuna.readthedocs.io/en/stable/reference/samplers/generated/optuna.samplers.BruteForceSampler.html#optuna.samplers.BruteForceSampler:~:text=(objective)-,Note,-The%20defined%20search)), so what this test catch as `ValueError` seem the one from `BruteForceSampler`, not from `OptunaSearchCV`.

Alnusjaponica: @HideakiImamura 
Is it better to implement something like `DeterministicSampler` instead of using `BruteForceSampler`?

HideakiImamura:
@jckkvs Thanks for your work. Since it appears to be very costly at the point of adding unit tests, how about we proceed to add them without unit tests once in this PR, or only add very simple ones, and then add tests as a follow-up by committers (us)?

HideakiImamura:
@jckkvs Thanks for the reply. So, could you remove the unit-tests which you added in this PR? Sorry for the confusion.

HideakiImamura:
It would be better to undo all changes in the test file.

optuna/integration/sklearn.py
```diff
@@ -215,17 +215,29 @@ def __call__(self, trial: Trial) -> float:
...
+                    fit_params=self.fit_params,
+                    groups=self.groups,
+                    return_train_score=self.return_train_score,
+                    scoring=self.scoring,
+                )
+            except ValueError:
+                n_splits = self.cv.get_n_splits(self.X, self.y, self.groups)
+                fit_time = np.array([np.nan] * n_splits)
+                score_time = np.array([np.nan] * n_splits)
+                test_score = np.array([self.error_score] * n_splits)
```
Alnusjaponica, line 234: https://github.com/optuna/optuna/pull/4642#discussion_r1223835320

```suggestion
                test_score = np.array(
                    [self.error_score if self.error_score is not None else np.nan] * n_splits
                )
```

Alnusjaponica (changes requested):

Alnusjaponica:
Other than the suggestion above, LGTM.

Alnusjaponica (approved):
LGTM.

HideakiImamura (approved):
Thank you for the long running! LGTM!

Issue #3450, opened 2022-04-06
[RFC] Enable trial report multiple intermediate values by `trial.report`


nzw0301:
### Motivation

Requested by https://github.com/optuna/optuna/issues/2792. With this change, users will be able to implement multi-objective pruning algorithms easily. In addition, storing intermediate values for multi-objective optimisation is supposed to be useful to analyse the trials after optimisation.

### Description

Accept multiple intermediate values: `trial.report(value=[1.0, 2.0] step=1)` or `trial.report(values=[1.0, 2.0] step=1)`

HideakiImamura:
I think it would be nice to allow the multiple intermediate values. I would like to know other opinions.

josecannete:
I also think this is an important feature.

In my case, I want to use Optuna for multi-objective optimization and it seems that pruning is one of the keys for a good search. One simple approach would be to use pruning just based on one objective, I think that should be pretty easy to implement but it would require the interface to accept multiple intermediate values. More complex pruning methods can come later.

Hope this feature gets discussed or into the roadmap.

julian-berger:
I can attest that this would be a much appreciated feature! Same reasons as above, also would agree that pruning based on one objective would already be a big step forward

mariajmolina:
Will this feature be implemented soon? I agree that it would be extremely useful (multi-objective pruning enabled for trial.report).

nzw0301:
@mariajmolina I'm afraid I don't think so because no core-dev team has updated this issue.

nzw0301:
@josecannete  @julian-berger  @mariajmolina @gitpeblo @raquelhortab 
Hi, I've created a rough implementation to address this issue #4652. By this, we can report multiple intermediate values like 
 `trial.report(value=[1.0, 2.0] step=1)`, and you can prune unpromising trials for multi-objective optimisation by defining your favourite pruner.

The limitations are 

1. You cannot report intermediate values at the same step, the same as the current Optuna's `trial.report`.
2. The reported intermediate values should have the same length as their objectives.

I really appreciate it if you leave any comments or reviews.

cordeirojoao:
Hello,
I would also appreciate both having report in objective optimisation, and having prunning.
That would a very welcome feature

cross32768:
Hello.

For this issue, I was working on PR #4822 to make the `intermediate_value` compatible with `Sequence[float]`, but I have noticed that it is not easy as expected by the following reasons:

[...]

PR #4652, opened 2023-05-04
[RFC] Allow `trial.report` to take multiple intermediate values

nzw0301:
## Motivation

Make the discussion of https://github.com/optuna/optuna/issues/3450, which is requested by multiple users forward.

## Description of the changes

Modify the codebase to take multiple intermediate values at the same step mainly as follows:

1. the RDB storage scheme
2. `Trial.report`
3. pruners that do nothing for multi-objective values.

Meanwhile, I try to keep the backward compatibility as much as possible.

---

I'm wondering if the current approach could work for the dev team. 

Note that I'm not familiar with writing migration code in `optuna/storages/_rdb/alembic/versions/`, which is not described in [`CONTRIBUTING.md`](https://github.com/optuna/optuna/blob/master/CONTRIBUTING.md), so I really appreciate if you tell me how to write it.


optuna/samplers/_tpe/sampler.py
```diff
@@ -629,10 +629,12 @@ def _get_observation_pairs(
             score = (-float("inf"), [sign * v for sign, v in zip(signs, trial.values)])
         elif trial.state is TrialState.PRUNED:
             if study._is_multi_objective():
+                # TODO(nzw0301): Can we use intermediate_values values?
                 continue
 
             if len(trial.intermediate_values) > 0:
                 step, intermediate_value = max(trial.intermediate_values.items())
+                assert isinstance(intermediate_value, float)
```
nzw0301, line 637: for mypy

optuna/storages/_rdb/storage.py
```diff
@@ -863,6 +879,28 @@ def _build_frozen_trial_from_trial_model(self, trial: "models.TrialModel") -> Fr
...
+                    intermediate_values[step] = tuple(
+                        intermediate_value
+                        for _, intermediate_value in sorted(
+                            intermediate_indices_and_values.items()
+                        )
+                    )
+            else:
+                for step, intermediate_indices_and_values in unsorted_intermediates.items():
+                    intermediate_values[step] = intermediate_indices_and_values[0]
+
```
nzw0301, line 903: I'm not sure `intermediate_values` is sorted by `intermediate_value_index` in case, but if not, we can simplify this part.

HideakiImamura:
Thank you for submitting an important PR. We discussed this PR within the committer. First of all, as a basic premise, we committers were not fully up to speed with the discussion of #3450. Our apologies. We are positive about moving this PR forward, as we think it is important to note that there have been several requests from users for the ability to receive multi-objective intermediate values.

However, we do have some concerns. The first is a scheduling concern. We are currently in the process of developing v3.2 for release and are currently pushing for a deadline for next month's release. So, would you be willing to discuss the details of this PR after the v3.2 release?

My second concern is about use cases. Do you have any ideas for an algorithm to prune trials using multi-objective intermediate values? I would like to have some idea of how specific multi-objective intermediate values are used. Also, does the intermediate values we report need to be the same as the dimension of the multi-objective function? I suppose this could change depending on the use case.

nzw0301:
Thank you for letting me know that. I'm glad to hear that.

### Scheduling

No problem. 

### Use cases

> Do you have any ideas for an algorithm to prune trials using multi-objective intermediate values? 

No, I don't. But one idea proposed in https://github.com/optuna/optuna/issues/3450 is just using one of the intermediate values, see https://github.com/optuna/optuna/issues/3450#issuecomment-1274956712 and they want to save the intermediate values at least.

Personally, even if we don't have such a pruning algorithm for multi-objective functions, accepting intermediate values is useful for those who (let's say an intern student at PFN or researchers who use Optuna) want to implement such algorithms since we can't report multiple intermediate values at the same step.

> Also, does the intermediate values we report need to be the same as the dimension of the multi-objective function? I suppose this could change depending on the use case.

Definitely, this point should be discussed as mentioned in https://github.com/optuna/optuna/pull/4652#discussion_r1185198407.

nzw0301:
I'll resolve the conflicts in 8 weeks...

HideakiImamura:
@nzw0301 Let's restart this work. Do you have any concern about this specifications?

HideakiImamura:
As we discussed in offline, if pruning is used, the specification that the last intermediate value is used as the trial value may have some adverse effect. This specification is defined in `Study.tell`. 

On the other hand, in the various Optuna samplers, too, how pruned trials are handled, for example, `TPESampler` similarly uses the value of the last intermediate value as a value independently of `Study.tell`. We are concerned that such specifications scattered within Optuna may cause some inconsistency with the features introduced in this PR.

If possible, we would appreciate it if you could investigate whether or not this specification would cause any inconsistencies, as it may be a bit of a time-consuming task as we need to look at a wide range of `Study.tell`, each sampler, etc. If you are too busy to afford this, I would be happy to wind up this conversation with our committer.

HideakiImamura:
@nzw0301 How about this one? We can do a handover if you like.

nzw0301:
As implied by https://github.com/optuna/optuna/pull/4652#issuecomment-1589315093, I don't have time to update this PR soon because I'm a bit busy nowadays. Of course, I don't care that someone kindly takes over this PR.

cross32768:
If you don't mind, I will take over this PR and proceed.

nzw0301:
@cross32768 Brilliant. Thank you!

optuna/trial/_trial.py
```diff
@@ -472,19 +470,35 @@ def objective(trial):
...
+                raise TypeError(message) from None
+
+        # TODO(nzw0301): # This same length constraint is unnecessary and the inconsistency can be
+        # detected in pruners. But it would make pruner more complicated; need discussions on this.
+        if isinstance(values, float):
+            num_intermediate_values = 1
+        else:
+            num_intermediate_values = len(values)
+        num_objectives = len(self.study.directions)
+        if num_objectives != num_intermediate_values:
```
cross32768, line 497: This check may be not needed

cross32768:
I discussed with the other committers for this PR. The size of the PR is getting a bit large, so I'm thinking of closing this PR and creating a separate PR that rewrites the some part first while reusing the code, is that OK @nzw0301 ?

nzw0301:
@cross32768 Of course. Thanks again.

PR #4443, opened 2023-02-20
Fix CMA-ES Sampler

contramundum53:
## Motivation
Currently, `sigma0` of `CmaEsSampler` is determined by the minimum width of search space. (https://github.com/optuna/optuna/blob/master/optuna/samplers/_cmaes.py#L516)

This becomes a problem when there are two variables with very different scales. 

## Description of the changes
We rescale all variables to [0, 1]. To do this, we add an option `transform_0_1` to `_SearchSpaceTransform`.

PR #4661, opened 2023-05-10
Fix CMA-ES with margin bug

contramundum53:
## Motivation
A bug is introduced for CMA-ES with margin in #4443 . It normalizes the search space of all variables to 0-1 segment, but it does not transform `step` value, thus giving incorrect optimization result.

## Description of the changes
Also transform `step`.


optuna/samplers/_cmaes.py
```diff
@@ -560,7 +560,12 @@ def _init_optimizer(
             for i, dist in enumerate(trans._search_space.values()):
                 assert isinstance(dist, (IntDistribution, FloatDistribution))
                 # Set step 0.0 for continuous search space.
-                steps[i] = dist.step or 0.0
+                if dist.step is None:
+                    steps[i] = 0.0
+                elif dist.low == dist.high:
+                    steps[i] = 1.0
+                else:
+                    steps[i] = dist.step / (dist.low - dist.high)
```
c-bata, line 568: When given `FloatDistribution(0, 10, step=2)`, the `steps[i]` should be `0.2`. However, in this code, it looks that `steps[i]` is set to `-0.2`. What do you think?

```suggestion
                    steps[i] = dist.step / (dist.high - dist.low)
```

optuna/samplers/_cmaes.py
```diff
@@ -560,7 +560,12 @@ def _init_optimizer(
             for i, dist in enumerate(trans._search_space.values()):
                 assert isinstance(dist, (IntDistribution, FloatDistribution))
                 # Set step 0.0 for continuous search space.
-                steps[i] = dist.step or 0.0
+                if dist.step is None:
+                    steps[i] = 0.0
+                elif dist.low == dist.high:
+                    steps[i] = 1.0
+                else:
+                    steps[i] = dist.step / (dist.high - dist.low)
```
knshnb, line 568: As we discussed privately, this ad-hoc transform of `step` might easily cause a bug (e.g., by changing `_SearchSpaceTransform` options). How about adding `step` property to `_SearchSpaceTransform`?

c-bata (commented):
I checked the performance with HPOBench with the similar settings as https://github.com/CyberAgentAILab/cmaes/pull/121#issuecomment-1296691448 and confirmed that the performance regression bug has been fixed in this PR.

| Problem Name | Before (master) | After (PR) |
| --- | --- | --- |
| Slice | hpo-bench-slice-fbe0ab9a7615a98973700fda6fd263493d8fe07ad14bd51a00add56b88088bbc | hpo-bench-slice-fbe0ab9a7615a98973700fda6fd263493d8fe07ad14bd51a00add56b88088bbc |
| Protein | hpo-bench-protein-ba1fe778428f2231dec3b556bb1a56fab4fc773a94b4d148f22a4b8dcd27bfe4 | hpo-bench-protein-ba1fe778428f2231dec3b556bb1a56fab4fc773a94b4d148f22a4b8dcd27bfe4 |
| Parkinson | hpo-bench-parkinson-569fd93441367839d07bdcffacf92a08451dcbc8da343d577839268955cb4044 | hpo-bench-parkinson-569fd93441367839d07bdcffacf92a08451dcbc8da343d577839268955cb4044 |
| Naval | hpo-bench-naval-a3678e022edad9a53645d616da69183b67b2f5d0eab398eefd9bbded4d8b016a | hpo-bench-naval-a3678e022edad9a53645d616da69183b67b2f5d0eab398eefd9bbded4d8b016a |

Code: https://gist.github.com/c-bata/c257feffa35ad5a35f5df2ca100e73e9

c-bata (approved):
LGTM

eukaryo (approved):
LGTM. I used simple two-dimensional examples and confirmed the followings:

- If the scales of variables x and y are very different, both vanilla-cmaes and cmaes-with-margin converge if both variables are float.
- If at least one variable is int, cmaes-with-margin converges with high probability; vanilla-cmaes also converges (slightly lower than the probability of with-margin, as expected).

optuna/samplers/_cmaes.py
knshnb: Memo: This PR is a hotfix and we will work on this as a follow-up work.

knshnb (approved):
LGTM!

knshnb:
Let me list up follow-up tasks as a reminder.
- Move the logic of transforming `step` inside `_SearchSpaceTransform` after refactoring it.
- Consider how `IntDistribution` with `log=True` can be handled by `CMAwM` (currently, it is regarded as a continuous space).

PR #4668, opened 2023-05-11
Move the function `_distribution_is_log` to `optuna.distributions` from `optuna/terminator/__init__.py`

cross32768:
## Motivation
The function `_distribution_is_log` is in `optuna/terminator/__init__.py`, but it is not so appropriate position. 
For example, this can cause circular import if we import something in optuna.terminator in `optuna/terminator/__init__.py` because `_distribution_is_log` is used in `optuna.terminator`.
In this PR, I move `_distribution_is_log` to `optuna.distributions`.

## Description of the changes
* Move the function `_distribution_is_log` to `optuna.distributions` from `optuna/terminator/__init__.py`
* Change the name to`_is_distribution_log` because I think it is better name
* Add tests


contramundum53 (approved):
LGTM.

contramundum53:
@eukaryo Could you reivew this PR?

eukaryo (approved):
LGTM too.

PR #4457, opened 2023-02-22
Add windows ci check

gen740:
## Motivation

In order to make the windows-ci "required". Add Windows-test.yaml to workflows to run tests
on Windows OS. And run these tests on PR, master merge, and schedule.

## Description of the changes

- Add sleep(0.001) before and after creating a trial. (This change would fix #4453)
- Be sure to `plt.close()` after creating `plt.Figure` (This change is related to #3815)
- Add `engine.storage.dispose()` to make sure to detach the process from the database before deleting the temporary database.
- Add `dtype=np.int64` argument to `randint` because on Windows int is 32bit.
- `tempfile.NamedTemporaryFile(delete=False)` to not delete the temporary file because Windows cannot delete the file occupied by a process.
- Skip the tests which would fail on Windows OS.
- Use QtAgg as a matplotlib backend because on Windows `plt.figure()` sometimes fails when using a tkinter backend.

PR #4696, opened 2023-05-19
Support `JournalFileStorage` and `JournalRedisStorage` on CLI

c-bata:
## Motivation

It's useful to support JournalFileStorage and JournalRedisStorage as requested at https://github.com/optuna/optuna-dashboard/issues/379.

## Description of the changes

Currently Optuna's command line interface only allows specifying RDBStorage, but this PR will also allow specifying JournalFileStorage and JournalRedisStorage as follows.

* `JournalFileStorage` : `optuna create-study --storage ./journal.log`
* `JournalRedisStorage` : `optuna create-study --storage redis://localhost:6379`
* `RDBStorage` : `optuna create-study --storage sqlite://db.sqlite3`

However, such a design may cause following two problems.

1. if there are multiple FileStorage or RedisStorage of the same storage_url format, it is not possible to identify which storage class is used from the `storage_url`.
1. As RedisStorage was removed in Optuna v3.1 and replaced by JournalRedisStorage, it may cause breaking changes to storage classes in the future.

Therefore, this PR introduce the `--storage-class` argument. It is possible to explicitly specify which storage class is to be used, thus solving the above problems.

* `JournalFileStorage` : `optuna create-study --storage ./journal.log --storage-class JournalFileStorage`
* `JournalRedisStorage` : `optuna create-study --storage redis://localhost:6379 --storage-class JournalRedisStorage`
* `RDBStorage` : `optuna create-study --storage sqlite://db.sqlite3 --storage-class RDBStorage`

### DIFF
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -21,8 +21,8 @@ from typing import Type
 from typing import Union
 import warnings
 
-import yaml
 import sqlalchemy.exc
+import yaml
 
 import optuna
 from optuna._imports import _LazyImport
@@ -685,7 +685,9 @@ class _StorageUpgrade(_BaseCommand):
     def take_action(self, parsed_args: Namespace) -> int:
         storage_url = _check_storage_url(parsed_args.storage)
         try:
-            storage = RDBStorage(storage_url, skip_compatibility_check=True, skip_table_creation=True)
+            storage = RDBStorage(
+                storage_url, skip_compatibility_check=True, skip_table_creation=True
+            )
         except sqlalchemy.exc.ArgumentError:
             self.logger.info("Invalid RDBStorage URL.")
             return 1
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -689,7 +689,7 @@ class _StorageUpgrade(_BaseCommand):
                 storage_url, skip_compatibility_check=True, skip_table_creation=True
             )
         except sqlalchemy.exc.ArgumentError:
-            self.logger.info("Invalid RDBStorage URL.")
+            self.logger.error("Invalid RDBStorage URL.")
             return 1
         current_version = storage.get_current_version()
         head_version = storage.get_head_version()
@@ -940,8 +940,12 @@ def _add_common_arguments(parser: ArgumentParser) -> ArgumentParser:
     parser.add_argument(
         "--storage-class",
         help="Storage class hint (e.g. JournalFileStorage)",
-        type=str,
         default=None,
+        choices=[
+            RDBStorage.__name__,
+            JournalFileStorage.__name__,
+            JournalRedisStorage.__name__,
+        ],
     )
     verbose_group = parser.add_mutually_exclusive_group()
     verbose_group.add_argument(
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -22,11 +22,16 @@ from typing import Union
 import warnings
 
 import yaml
+import sqlalchemy.exc
 
 import optuna
 from optuna._imports import _LazyImport
 from optuna.exceptions import CLIUsageError
 from optuna.exceptions import ExperimentalWarning
+from optuna.storages import BaseStorage
+from optuna.storages import JournalFileStorage
+from optuna.storages import JournalRedisStorage
+from optuna.storages import JournalStorage
 from optuna.storages import RDBStorage
 from optuna.trial import TrialState
 
@@ -49,6 +54,27 @@ def _check_storage_url(storage_url: Optional[str]) -> str:
         )
         return env_storage
     raise CLIUsageError("Storage URL is not specified.")
+
+
+def _get_storage(storage_url: Optional[str], storage_class: Optional[str]) -> BaseStorage:
+    storage_url = _check_storage_url(storage_url)
+    if storage_class:
+        if storage_class == JournalRedisStorage.__name__:
+            return JournalStorage(JournalRedisStorage(storage_url))
+        if storage_class == JournalFileStorage.__name__:
+            return JournalStorage(JournalFileStorage(storage_url))
+        if storage_class == RDBStorage.__name__:
+            return RDBStorage(storage_url)
+        raise CLIUsageError("Unsupported storage class")
+
+    if storage_url.startswith("redis"):
+        return JournalStorage(JournalRedisStorage(storage_url))
+    if os.path.isfile(storage_url):
+        return JournalStorage(JournalFileStorage(storage_url))
+    try:
+        return RDBStorage(storage_url)
+    except sqlalchemy.exc.ArgumentError:
+        raise CLIUsageError("Failed to guess storage class from storage_url")
 
 
 def _format_value(value: Any) -> Any:
@@ -293,8 +319,7 @@ class _CreateStudy(_BaseCommand):
         )
 
     def take_action(self, parsed_args: Namespace) -> int:
-        storage_url = _check_storage_url(parsed_args.storage)
-        storage = optuna.storages.get_storage(storage_url)
+        storage = _get_storage(parsed_args.storage, parsed_args.storage_class)
         study_name = optuna.create_study(
             storage=storage,
             study_name=parsed_args.study_name,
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -1049,6 +1056,51 @@ def test_check_storage_url() -> None:
         optuna.cli._check_storage_url(None)
 
 
+@pytest.mark.skipif(platform.system() == "Windows", reason="Skip on Windows")
```
c-bata, line 1059: These tests issues following errors on Windows.

> FAILED tests/test_cli.py::test_get_storage_without_storage_class - sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) unable to open database file
> (Background on this error at: https://sqlalche.me/e/20/e3q8)
> FAILED tests/test_cli.py::test_get_storage_with_storage_class - sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) unable to open database file
> (Background on this error at: https://sqlalche.me/e/20/e3q8)
> = 2 failed, 4103 passed, 88 skipped, 1003 deselected, 164 warnings in 1067.88s (0:17:47) =
> Error in atexit._run_exitfuncs:
> https://github.com/optuna/optuna/actions/runs/5020758022/jobs/9002514278?pr=4696

I checked `tests/storage_tests/rdb_tests/test_storage.py` to resolve above errors, however tests are skipped on Windows as follows. So let me use `pytest.mark.skipif` decorator here.

https://github.com/optuna/optuna/pull/4457/files#diff-42166922cf20058a0456b29a4eccb65307689d7aefed170b335539b5e3ed11eb

optuna/cli.py
```diff
@@ -35,6 +40,28 @@
 
 _DATETIME_FORMAT = "%Y-%m-%d %H:%M:%S"
 
+# The following code is quoted from the source code of SQLAlchemy project (MIT License).
+# See LICENSE file for detials.
+# https://github.com/zzzeek/sqlalchemy/blob/c6554ac52/lib/sqlalchemy/engine/url.py#L234-L292
```
toshihikoyanase, line 45: The description of the repository recommends 
 https://github.com/sqlalchemy/sqlalchemy. Do we stick to `https://github.com/zzzeek/sqlalchemy`?

image

optuna/cli.py
```diff
@@ -35,6 +40,28 @@
 
 _DATETIME_FORMAT = "%Y-%m-%d %H:%M:%S"
 
+# The following code is quoted from the source code of SQLAlchemy project (MIT License).
+# See LICENSE file for detials.
```
toshihikoyanase, line 44: `The following code` might be a bit clear. It can include all the line after L44, so we might want to specify the implementation of `_rfc1738_pattern`.

optuna/cli.py
c-bata: Thank you for pointing it out. Good catch. We should refer the `sqlalchemy/sqlalchemy` repo.

```suggestion
# https://github.com/sqlalchemy/sqlalchemy/blob/c6554ac52/lib/sqlalchemy/engine/url.py#L234-L252
```

optuna/cli.py
c-bata: Thank you for your review. Does the following change look good?

```suggestion
# `_rfc1738_pattern` is quoted from the source code of SQLAlchemy project (MIT License).
# See LICENSE file for details.
```

c-bata:
@toshihikoyanase Thank you for your review! I updated the implementation and squashed the commits. PTAL.

optuna/cli.py
```diff
@@ -910,6 +937,12 @@ def _add_common_arguments(parser: ArgumentParser) -> ArgumentParser:
             "Also can be specified via OPTUNA_STORAGE environment variable."
         ),
     )
+    parser.add_argument(
+        "--storage-class",
+        help="Storage class hint (e.g. JournalFileStorage)",
+        type=str,
```
toshihikoyanase, line 943: This is not a strong opinion, but what do you think of the `choices` argument to enumerate the acceptable storage classes?  It could potentially help users understand its usage more readily. On the other hand,  we would then need to maintain the list along with the implementation of `_get_storage`.

optuna/cli.py
```diff
@@ -656,14 +680,17 @@ def take_action(self, parsed_args: Namespace) -> int:
...
     def take_action(self, parsed_args: Namespace) -> int:
         storage_url = _check_storage_url(parsed_args.storage)
-        if storage_url.startswith("redis"):
-            self.logger.info("This storage does not support upgrade yet.")
+        try:
+            storage = RDBStorage(
+                storage_url, skip_compatibility_check=True, skip_table_creation=True
+            )
+        except sqlalchemy.exc.ArgumentError:
+            self.logger.info("Invalid RDBStorage URL.")
```
toshihikoyanase, line 692: It might be an error message, but the original implementation uses `self.logger.info`. Just a question, but do we have any reason not to use `self.logger.error` here?

toshihikoyanase (commented):
Thank you for your update. The change overall looks good to me, but I have some questions.

toshihikoyanase (commented):
Thank you for your update. The change overall looks good to me, but I have some questions.

optuna/cli.py
c-bata: Thank you for your comment.

The reason I used `logger.info(...)` here is because the previous implementation used `logger.info(...)` when given a Redis storage url, so I followed that. However, as you pointed out, using `logger.error(...)` sounds reasonable.

```suggestion
            self.logger.error("Invalid RDBStorage URL.")
```

optuna/cli.py
c-bata: Sounds reasonable! I will update the PR.

c-bata:
Thank you for your comments. I updated the PR. PTAL.

optuna/cli.py
```diff
@@ -51,6 +56,27 @@ def _check_storage_url(storage_url: Optional[str]) -> str:
...
+        if storage_class == JournalFileStorage.__name__:
+            return JournalStorage(JournalFileStorage(storage_url))
+        if storage_class == RDBStorage.__name__:
+            return RDBStorage(storage_url)
+        raise CLIUsageError("Unsupported storage class")
+
+    if storage_url.startswith("redis"):
+        return JournalStorage(JournalRedisStorage(storage_url))
+    if os.path.isfile(storage_url):
+        return JournalStorage(JournalFileStorage(storage_url))
```
toshihikoyanase, line 73: This implementation expects an existing file, so the following command fails if `journal.log` does not exist.

```console
$ optuna create-study --storage ./journal.log
[E 2023-06-13 14:10:56,836] Failed to guess storage class from storage_url
```

How about accepting newly created files?

toshihikoyanase (commented):
I'm sorry for the delayed response. I found an issue regarding `JournalFileStorage` with new files. 


In addition, I found that error messages might be somewhat unclear when an incorrect sqlite3 database file was provided to the `JournalFileStorage`.

Here's the console output:

```console
$ optuna create-study --storage sqlite:///db.sqlite3
[I 2023-06-13 14:14:35,798] A new study created in RDB with name: no-name-c733a028-03cd-4105-82af-791623edfe44
no-name-c733a028-03cd-4105-82af-791623edfe44
$ optuna create-study --storage ./db.sqlite3
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd0 in position 99: invalid continuation byte
```
It seems the error message could be made more user-friendly by explicitly mentioning the cause of the error. It could, for example, indicate to users that they have incorrectly specified `JournalFileStorage` when `RDBStorage` was the expected input. 
 But, it should be resolved in a follow-up PR.

optuna/cli.py
toshihikoyanase, line 73: More concretely, how about checking the existence of the files?
```suggestion
    if os.path.isfile(storage_url) or not os.path.exists(storage_url):
        return JournalStorage(JournalFileStorage(storage_url))
```

eukaryo: In this suggested change, we are trying to make sure that the directory does not exist.
But won't it cause a bug if we specify a directory that doesn't exist?

contramundum53: [to be discussed in a follow-up] Maybe we could check whether `os.path.dirname(storage_url)` is an existing directory?

toshihikoyanase: The usage of `os.path.dirname(storage_url)` might not provide a straightforward solution because in some cases, `os.path.dirname` treats a part of the DB_URL as a parent directory. Therefore, the issue may not be resolved with this approach alone.

```python
>>> import os
>>> os.path.dirname("sqlite:///foo/bar.db")
'sqlite:///foo'
```

Let us accept the current implementation and discuss it in a follow-up PR.

toshihikoyanase (approved):
LGTM!

c-bata:
Let me reassign the reviewer. @not522 Could you review this PR?

not522 (approved):
LGTM!

PR #4709, opened 2023-05-26
Fix docstring of terminator improvement for `min_n_trials`

cross32768:
## Motivation
Docstring of terminator improvement has an error, so this PR fix it


HideakiImamura (approved):
LGTM.

PR #4712, opened 2023-05-26
Speed up `logpdf` in `_truncnorm.py`

not522:
## Motivation
The shapes of `a` and `b` in `logpdf` is `(1, n_below or n_above)`, but broadcasted them are `(n_ei_candidates, n_below or n_above)`. Calculating `_log_gauss_mass` before broadcast makes it faster.

----

```python
import optuna

def objective(trial):
    return sum([trial.suggest_float(f"x_{i}", -100, 100) ** 2 for i in range(10)])

study = optuna.create_study()
```

- master
```
$ time python bench.py
real	1m11.063s
user	1m9.636s
sys	0m1.247s
```

- PR
```
$ time python bench.py
real	1m2.653s
user	1m1.911s
sys	0m0.533s
```

## Description of the changes
- Call `_log_gauss_mass` before broadcast

### DIFF
--- a/optuna/samplers/_tpe/_truncnorm.py
+++ b/optuna/samplers/_tpe/_truncnorm.py
@@ -225,10 +225,10 @@ def logpdf(
     x = (x - loc) / scale
 
     x, a, b = np.atleast_1d(x, a, b)
-    x, a, b = np.broadcast_arrays(x, a, b)
 
     out = _norm_logpdf(x) - _log_gauss_mass(a, b)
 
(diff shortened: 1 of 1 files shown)


c-bata:
@HideakiImamura @cross32768 Could you review this PR if you have time?

cross32768 (approved):
I also measured the running time of proposed example code in my environment and observe the speedup. LGTM!
```
master: 78.099 s
this PR: 69.646 s
```

HideakiImamura (approved):
LGTM.

PR #4716, opened 2023-05-30
Speed up `get_all_trials` in `InMemoryStorage`

not522:
## Motivation
Filtering trials with `filter` and `list` is slower than list comprehension. It is a bottleneck when the number of trials is large.

----

```python
import optuna

def objective(trial):
    return sum([trial.suggest_float(f"x_{i}", -100, 100) ** 2 for i in range(10)])

study = optuna.create_study(sampler=optuna.samplers.RandomSampler())
```

- master

```
$ time python bench.py
real	0m33.892s
user	0m33.397s
sys	0m0.530s
```

- PR

```
$ time python bench.py
real	0m29.847s
user	0m29.413s
sys	0m0.497s
```

## Description of the changes
- Use a list comprehension to speed up `get_all_trials`.
- Simplify the implementation.


c-bata:

optuna/storages/_in_memory.py
```diff
@@ -362,14 +360,12 @@ def get_all_trials(
...
+            trials = self._studies[study_id].trials
             if states is not None:
-                trials = filter(lambda t: t.state in states, trials)
+                trials = [t for t in trials if t.state in states]
 
             if deepcopy:
-                trials = copy.deepcopy(list(trials))
+                trials = copy.deepcopy(trials)
             else:
                 trials = list(trials)
```
c-bata, line 370: [nit] Can we also remove the call of `list()` here since `self._studies[...].trials` is `List[FrozenTrial]`?

```suggestion
```

c-bata (commented):
LGTM with a nit.

optuna/storages/_in_memory.py
not522: This shallow copy is necessary to ensure that returned trials are not modified by `set_trial_xxx`. I added commit for clarity.

Alnusjaponica:
The change itself LGTM. I am going to approve this PR after @not522 added the test for the commented part.

optuna/storages/_in_memory.py
c-bata: I see. Thank you for the update!

c-bata:
> The change itself LGTM. I am going to approve this PR after @not522 added the test for the commented part.

I think that we might not need to add a test case if the existing test cases would fail without the shallow copy.

@not522 Do you think it's necessary to add more test cases?

not522:
I think the tests are insufficient. There is no test for this feature and only the `SuccessiveHalvingPruner`'s test fails. https://github.com/optuna/optuna/actions/runs/5116276586/jobs/9198283871

not522:
I added the test that returned `FrozenTrials` are not modified by the subsequent `set_trial_xxx`. PTAL.

Alnusjaponica (approved):
LGTM.

Alnusjaponica (approved):
LGTM.

c-bata (approved):

PR #4720, opened 2023-05-31
Fix `BruteForceSampler` for pruned trials

not522:
## Motivation
`BruteForceSampler` should consider that trials may be pruned.

## Description of the changes
- Add `PRUNED` for building search tree.

### DIFF
--- a/tests/samplers_tests/test_brute_force.py
+++ b/tests/samplers_tests/test_brute_force.py
@@ -133,10 +133,15 @@ def test_study_optimize_with_pruned_trials() -> None:
 
         if a == 0:
             trial.suggest_float("b", -1.0, 1.0, step=0.5)
+            raise optuna.TrialPruned
         elif a == 1:
-            trial.suggest_categorical("c", ["x", "y", None])
-
-        raise optuna.TrialPruned
+            c = trial.suggest_categorical("c", ["x", "y", None])
+            if c == "x":
+                return a + 1
+            else:
+                return a - 1
+        else:
+            return a * 2
 
     study = optuna.create_study(sampler=samplers.BruteForceSampler())
     study.optimize(objective)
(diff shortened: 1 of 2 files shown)


c-bata:
@cross32768 @gen740 Could you review this PR if you have time?

tests/samplers_tests/test_brute_force.py
```diff
@@ -127,6 +127,37 @@ def objective(trial: Trial) -> float:
...
+def test_study_optimize_with_pruned_trials() -> None:
+    def objective(trial: Trial) -> float:
+        a = trial.suggest_int("a", 0, 2)
+
+        if a == 0:
+            trial.suggest_float("b", -1.0, 1.0, step=0.5)
+        elif a == 1:
+            trial.suggest_categorical("c", ["x", "y", None])
+
+        raise optuna.TrialPruned
```
cross32768, line 139: :nit: Wouldn't we need a `return` statement?

cross32768 (approved):
Otherwise, LGTM!

tests/samplers_tests/test_brute_force.py
gen740: I think it's okay because the pruned state is set in the `catch TrialState.Pruned` line of the [`_optimize` function.](https://github.com/optuna/optuna/blob/master/optuna/study/_optimize.py#L199-L204).

gen740: I think this objective function will always be `Pruned`, so it would be better to make the pruning stochastic.

not522: I have updated the test to include `COMPLETE` trials. PTAL.

cross32768: LGTM! 
@gen740 Could you take another look? If you approve, I will merge this PR.

gen740 (approved):
LGTM!

PR #4724, opened 2023-06-01
Fix `plot_slice` bug when some of the choices are numeric

### LINKED ISSUE #4680: X-axis of plotly's `plot_slice` is broken with categorical values
### Expected behavior

matplotlib:
image

plotly:image

The categorical value "a" should be displayed on the x-axis like the matplotlib version.

### Environment

- Optuna version: current master
- Python version: 3.9.6
- OS: macOS

```shell
The categorical value "a" is not shown on the x-axis (See above figure).
```

### Steps to reproduce

```python
import optuna


def objective(trial):
    x = trial.suggest_categorical("x", (0, "a"))
    if x == 0:
        return 0
    else:
        return 1


study = optuna.create_study()
optuna.visualization.plot_slice(study).show()
```

_No response_

not522:
## Motivation
Fix #4680.

## Description of the changes
This PR adds `update_xaxes(type="category")`. It is documented here. https://plotly.com/python/categorical-axes/

### DIFF
--- a/optuna/visualization/matplotlib/_slice.py
+++ b/optuna/visualization/matplotlib/_slice.py
@@ -127,19 +127,29 @@ def _generate_slice_subplot(
     target_name: str,
 ) -> "PathCollection":
     ax.set(xlabel=subplot_info.param_name, ylabel=target_name)
-    x_values = subplot_info.x
     scale = None
     if subplot_info.is_log:
         ax.set_xscale("log")
         scale = "log"
-    elif not subplot_info.is_numerical:
-        x_values = [str(x) for x in subplot_info.x]
+    if subplot_info.is_numerical:
+        x_values = subplot_info.x
+        y_values = subplot_info.y
+        c_values = subplot_info.trial_numbers
+    else:
+        x_values = []
+        y_values = []
+        c_values = []
+        assert subplot_info.x_labels is not None
+        for x_label in subplot_info.x_labels:
+            for x, y, number in zip(subplot_info.x, subplot_info.y, subplot_info.trial_numbers):
+                if x == x_label:
+                    x_values.append(str(x))
+                    y_values.append(y)
+                    c_values.append(number)
         scale = "categorical"
     xlim = _calc_lim_with_padding(x_values, padding_ratio, scale)
     ax.set_xlim(xlim[0], xlim[1])
-    sc = ax.scatter(
-        x_values, subplot_info.y, c=subplot_info.trial_numbers, cmap=cmap, edgecolors="grey"
-    )
+    sc = ax.scatter(x_values, y_values, c=c_values, cmap=cmap, edgecolors="grey")
     ax.label_outer()
 
     return sc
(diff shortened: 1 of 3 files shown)


c-bata:

gen740 (commented):
Considering a following objective funcion.
```python
def objective(trial):
    version = trial.suggest_categorical("version", (0.1, 0.2, 1.1))
    if version == 0:
        return 0
    else:
        return 1
```
The result of plot is
newplot-3

What are your thoughts on this case? Writing the version as a string literal would indeed prevent this issue. But a radical solution would be to switch the plot type according to the distribution type, such as CategoricalDistribution.

not522:
I think it is dependent on the case. However, numeric axes can be hard to see like follows.

```python
import optuna

def objective(trial):
    return trial.suggest_categorical("x", (0.0001, 0.001, 0.01, 0.1, 1))

study = optuna.create_study()
optuna.visualization.plot_slice(study).show()
```

I have updated to use the categorical axis when the distribution is categorical. Also, I make sure that the labels are in the same order as the choices.

not522:
@c-bata Let me unassign you as you may be busy.

optuna/visualization/matplotlib/_slice.py
```diff
@@ -127,19 +127,29 @@ def _generate_slice_subplot(
...
+        x_values = []
+        y_values = []
+        c_values = []
+        assert subplot_info.x_labels is not None
+        for x_label in subplot_info.x_labels:
+            for x, y, number in zip(subplot_info.x, subplot_info.y, subplot_info.trial_numbers):
+                if x == x_label:
+                    x_values.append(str(x))
+                    y_values.append(y)
+                    c_values.append(number)
```
Alnusjaponica, line 148: I think this double loop can be separated with `defaultdict(list)` or whatever. Are there any reason to keep this double loop?

Alnusjaponica (commented):
I have a minor comment and other than that LGTM.

optuna/visualization/matplotlib/_slice.py
not522: I couldn't understand your comment, could you please write me a suggested code?

Alnusjaponica: What I imegined was like this. 
```suggestion
        y_values_dict = defaultdict(list)
        c_values_dict = defaultdict(list)
        assert subplot_info.x_labels is not None
        for x, y, number in zip(subplot_info.x, subplot_info.y, subplot_info.trial_numbers):
            y_values_dict[x].append(y)
            c_values_dict[x].append(number)
        x_values_count = Counter(subplot_info.x)

        for x_label in subplot_info.x_labels:
            x_values.extend([str(x_label)] * x_values_count[x_label])
            y_values.extend(y_values_dict[x_label])
            c_values.extend(c_values_dict[x_label])
```

not522: Thank you. I have updated to eliminate unnecessary searching. How about it?

Alnusjaponica: Thank for the update. LGTM.

Alnusjaponica (approved):
LGTM. 
@gen740 Could you take another look?

gen740 (approved):
LGTM! @not522 Can you rerun the CI check? There seems to be something wrong with it.

c-bata (approved):

Issue #4679, opened 2023-05-17
Add a warning for a progress bar not being displayed


knshnb:
### Motivation

When we call `study.optimize(..., n_trials=None, timeout=None, show_progress_bar=True)`, a progress bar is not displayed.

### Description

We should raise a warning for not displaying the progress bar when `n_trials=None` and `timeout=None`.

rishabsinghh:
Hey, i would like to start working on this issue. Could you please assign this to me?

knshnb:
Sure, please!

HideakiImamura:
This is resolved in #4728 by @rishabsinghh.

PR #4728, opened 2023-06-04
Add a warning for a progress bar not being displayed #4679

rishabsinghh:
## Description of the changes
 When n_trials == 0 and timeout == None and the show_progress_bar is set to True, It will output a warning.

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -445,9 +445,9 @@ class Study:
             RuntimeError:
                 If nested invocation of this method occurs.
         """
-        if ((n_trials == 0) and (timeout is None)) & (show_progress_bar==True):
+        if show_progress_bar and n_trials is None and timeout is None:
             warnings.warn("n_trials and timeout is set to None.Progress bar won't be displayed")
-            
+       
         _optimize(
             study=self,
             func=func,
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -446,6 +446,7 @@ class Study:
                 If nested invocation of this method occurs.
         """
 
+
         _optimize(
             study=self,
             func=func,
@@ -526,6 +527,7 @@ class Study:
 
         if not self._thread_local.in_optimize_loop and is_heartbeat_enabled(self._storage):
             warnings.warn("Heartbeat of storage is supposed to be used with Study.optimize.")
+
 
         fixed_distributions = fixed_distributions or {}
         fixed_distributions = {
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -445,9 +445,6 @@ class Study:
             RuntimeError:
                 If nested invocation of this method occurs.
         """
-        if show_progress_bar and n_trials is None and timeout is None:
-            warnings.warn("n_trials and timeout is set to None.Progress bar won't be displayed")
-       
         _optimize(
             study=self,
             func=func,
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -445,8 +445,9 @@ class Study:
             RuntimeError:
                 If nested invocation of this method occurs.
         """
-
-
+        if ((n_trials == 0) and (timeout is None)) & (show_progress_bar==True):
+            warnings.warn("n_trials and timeout is set to None.Progress bar won't be displayed")
+            
         _optimize(
             study=self,
             func=func,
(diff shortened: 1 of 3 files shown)


optuna/study/study.py
```diff
@@ -445,7 +445,9 @@ def objective(trial):
             RuntimeError:
                 If nested invocation of this method occurs.
         """
-
+        if ((n_trials == 0) and (timeout is None)) & (show_progress_bar==True):
+            warnings.warn("n_trials and timeout is set to None.Progress bar won't be displayed")
+
```
nzw0301, line 450: By following https://github.com/optuna/optuna/issues/4679, `n_trials` should be `None`. In addition, the condition can be simplified as follows:
```suggestion
        if show_progress_bar and n_trials is None and timeout is None:
            warnings.warn("n_trials and timeout is set to None.Progress bar won't be displayed")
       
```

tests/study_tests/test_study.py
```diff
@@ -14,6 +14,7 @@
 from typing import Union
 from unittest.mock import Mock
 from unittest.mock import patch
+from unittest import TestCase
```
nzw0301, line 17: Could you remove this?
```suggestion
```

tests/study_tests/test_study.py
```diff
@@ -1577,3 +1578,24 @@ def test_set_invalid_metric_names() -> None:
     study = create_study(directions=["minimize", "minimize"])
     with pytest.raises(ValueError):
         study.set_metric_names(metric_names)
+
+def objective(trial):
```
nzw0301, line 1582: We can remove this function because we can use `func` defined in this file as an objective.

tests/study_tests/test_study.py
```diff
@@ -1577,3 +1578,24 @@ def test_set_invalid_metric_names() -> None:
...
+
+    return result
+def test_warning_condition():
+    study = create_study()
+    n_trials = 0
+    timeout = None
+
+    # Trigger the condition
+    with pytest.warns(UserWarning):
+        Study.optimize(study,objective, n_trials=n_trials, timeout=timeout,show_progress_bar=True)
```
nzw0301, line 1598: Again, `n_trials` should be `None` in this case. 
I think the test would be as follows:

```python
def test_optimize_infinite_budget_progbar() -> None:
    def terminate_study(study: Study, trial: FrozenTrial) -> None:
        study.stop()

    study = create_study()

        study.optimize(
            func, n_trials=None, timeout=None, show_progress_bar=True, callbacks=[terminate_study]
        )
```

Note that `callbacks` is necessary to stop the `optimize`, otherwise, the test doesn't finish forever.

tests/study_tests/test_study.py
```diff
@@ -1577,3 +1578,24 @@ def test_set_invalid_metric_names() -> None:
     study = create_study(directions=["minimize", "minimize"])
     with pytest.raises(ValueError):
         study.set_metric_names(metric_names)
+
+def objective(trial):
+    x = trial.suggest_uniform('x', -10, 10)
+    y = trial.suggest_uniform('y', -5, 5)
+    z = trial.suggest_uniform('z', 0, 1)
```
nzw0301, line 1585: Optuna repo has introduced a few packages to unify the coding style as explained in https://github.com/optuna/optuna/blob/master/CONTRIBUTING.md. Part of these errors would be removed by running `./formats.sh`.

optuna/study/study.py
rishabsinghh: Hey. One question, If n_trials and timeout is None, to test this, we would need a triggering condition to stop it. Otherwise, won't it run indefinitely?

tests/study_tests/test_study.py
rishabsinghh: Ah, Got the answer i was looking for in the first comment!

nzw0301 (approved):
LGTM. Thanks!

tests/study_tests/test_study.py
rishabsinghh: Hi, Do you have a Linkedin? Wanted to connect on knowing more about the library so that i can contribute better

nzw0301: Mine or Optuna?

rishabsinghh: Yours too. And if Optuna has a slack channel or something, that would be helpful too

HideakiImamura (approved):
LGTM.

tests/study_tests/test_study.py
HideakiImamura: Optuna does not have any open slack channel. Sorry for the thin communication, I hope we can communicate via GitHub. If you have any concerns, please create an issue or discussion!

PR #4735, opened 2023-06-08
Add `plot_rank` and `plot_timeline` plots to visualisation tutorial

nzw0301:
## Motivation

In the same as the other visualisation functions, add examples of newly implemented plot functions to the tutorial page.

## Description of the changes

As the title added two functions except for the Terminator plot since it needs terminator logic that would be out-of-scope for the tutorial.


contramundum53 (approved):
LGTM! Thanks for your PR!

PR #4742, opened 2023-06-14
Get values after `_get_observation_pairs`

not522:
## Motivation
`_get_observation_pairs` computes the params and values information simultaneously, which makes the process difficult to follow. This PR changes it so that the params are obtained after `_get_observation_pairs`.

The name and documentation for `_get_observation_pairs` will not correspond to the implementation, but this will be corrected in a later PR.

## Description of the changes
- Get values after `_get_observation_pairs`

### DIFF
--- a/tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
+++ b/tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
@@ -311,7 +311,8 @@ def test_multi_objective_sample_independent_ignored_states() -> None:
 )
 @pytest.mark.parametrize("objective_value", [-5.0, 5.0, 0.0, -float("inf"), float("inf")])
 @pytest.mark.parametrize("multivariate", [True, False])
-@pytest.mark.parametrize("constant_liar", [True, False])
+# TODO(not522): Test constant_liar=True
+@pytest.mark.parametrize("constant_liar", [False])
 @pytest.mark.filterwarnings("ignore::optuna.exceptions.ExperimentalWarning")
 def test_multi_objective_get_observation_pairs(
     int_value: int,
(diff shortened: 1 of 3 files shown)


tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
```diff
@@ -311,7 +311,8 @@ def test_multi_objective_sample_independent_ignored_states() -> None:
 )
 @pytest.mark.parametrize("objective_value", [-5.0, 5.0, 0.0, -float("inf"), float("inf")])
 @pytest.mark.parametrize("multivariate", [True, False])
-@pytest.mark.parametrize("constant_liar", [True, False])
+# TODO(not522): Test constant_liar=True
```
HideakiImamura, line 314: Why does not this test `constant_liar=True`?

HideakiImamura (commented):
Thanks for the PR. I have a question.

tests/samplers_tests/tpe_tests/test_multi_objective_sampler.py
not522: I fixed the test as I realized we should ignore the `constant_liar` option as documented.

HideakiImamura: I see. Thanks for the explanation.

HideakiImamura (approved):
Thanks for the update. LGTM.

contramundum53 (approved):
LGTM!
A nit: Maybe the name `_get_observation_pairs` is obsolete, because we don't get the parameter values anymore?

not522:
Yes, I will work on a later PR.

> The name and documentation for _get_observation_pairs will not correspond to the implementation, but this will be corrected in a later PR.

PR #4747, opened 2023-06-15
Make BruteForceSampler consider failed trials

Alnusjaponica:
## Motivation
Ensure consistent behaviors for failed trials of both`BruteForceSampler` and `GridSampler`.
The following is minimum reproducible code to make `BruteForceSampler` sample the same parameters multiple times.
```
import optuna


def objective(trial):
    x = trial.suggest_int("x", 0, 1)
    return [0, 1] if x else "a"


study = optuna.create_study(sampler=optuna.samplers.BruteForceSampler(seed=0))
study.optimize(objective, n_trials=2)

```

## Description of the changes
Consider trials in `TrialState.FAIL` as finished trials.

### DIFF
--- a/tests/samplers_tests/test_brute_force.py
+++ b/tests/samplers_tests/test_brute_force.py
@@ -287,5 +287,6 @@ def test_study_optimize_with_failed_trials() -> None:
 
     expected_suggested_values = [{"x": i} for i in range(100)]
     all_suggested_values = [t.params for t in study.trials]
+    assert len(all_suggested_values) == len(expected_suggested_values)
     for a in expected_suggested_values:
         assert a in all_suggested_values
--- a/tests/samplers_tests/test_brute_force.py
+++ b/tests/samplers_tests/test_brute_force.py
@@ -275,3 +275,18 @@ def test_study_optimize_with_nonconstant_search_space() -> None:
     )
     with pytest.raises(ValueError):
         study.optimize(objective_decreasing_variable, n_trials=10)
+
+
+def test_study_optimize_with_failed_trials() -> None:
+    def objective(trial: Trial) -> float:
+        x = trial.suggest_int("x", 0, 99)
+        return np.nan
+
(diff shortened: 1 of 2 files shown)


contramundum53:
Shouldn't `BruteForceSampler` redo failed trials (by ignoring them)?

Alnusjaponica:
I believe it should not redo failed trials, as long as `GridSampler` does not, for consistency. In addition, it is possible that the failure is due to misconfiguration of parameters and redoing those trials instead of unseen parameter setting is inefficient.

toshihikoyanase:
@contramundum53 Could you continue the PR, please?
@not522 Could you join the review of this PR, please?

contramundum53:
Hmm, a failed trial may result from a trial being preempted or the cloud computation resources becoming unstable. What about adding an option to specify whether failed trials should be skipped?

not522:
> What about adding an option to specify whether failed trials should be skipped?

It sounds reasonable but may need discussion about the design. How about doing it with another PR?

tests/samplers_tests/test_brute_force.py
```diff
@@ -275,3 +275,17 @@ def objective_decreasing_variable(trial: Trial) -> float:
...
+def test_study_optimize_with_failed_trials() -> None:
+    def objective(trial: Trial) -> float:
+        x = trial.suggest_int("x", 0, 99)  # NOQA[F811]
+        return np.nan
+
+    study = optuna.create_study(sampler=samplers.BruteForceSampler())
+    study.optimize(objective, n_trials=100)
+
+    expected_suggested_values = [{"x": i} for i in range(100)]
+    all_suggested_values = [t.params for t in study.trials]
```
not522, line 289: Could you test the length of `all_suggested_values`?

contramundum53 (commented):
The code LGTM except for @not522 's comment.

contramundum53 (approved):

not522 (approved):
LGTM!

PR #4751, opened 2023-06-16
Add cv_results_ to OptunaSearchCV

jckkvs:
## Motivation
This is to enhance compatibility with [sklearn.GridSearchCV](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html).

The user_attrs is already implemented in Optuna, to obtain the score of each trial, but cv_results_enable to obtain the scores of all trials at once.

In addition,
we can easily calculate the bias_correction proposed by Tibshirani and Tibshirani by cv_results_　(https://tibshirani.su.domains/ftp/AOAS224.pdf).

## Description of the changes

Add cv_results_ to OptunaSearchCV

The variable names  ```cv_results_dict_in_list``` , ```cv_results_list_in_dict``` are very long, so we can change them according to optuna's conventions.

### DIFF
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -565,7 +565,8 @@ class OptunaSearchCV(BaseEstimator):
 
     @property
     def cv_results_(self) -> Dict[str, Any]:
-        """A dictionary mapping a metric name to a list of Cross-Validation results of all trials."""
+        """A dictionary mapping a metric name to a list 
+            of Cross-Validation results of all trials."""
         cv_results_dict_in_list = [trial_.user_attrs for trial_ in self.trials_]
         if len(cv_results_dict_in_list) == 0:
             cv_results_list_in_dict = {}
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -570,10 +570,13 @@ class OptunaSearchCV(BaseEstimator):
             A list of CV results of all trials.
         """
         cv_results_dict_in_list = [trial_.user_attrs for trial_ in self.trials_]
-        cv_results_list_in_dict = {
-            key: [dict_[key] for dict_ in cv_results_dict_in_list]
-            for key in cv_results_dict_in_list[0]
-        }
+        if len(cv_results_dict_in_list) == 0:
+            cv_results_list_in_dict = {}
+        else:
+            cv_results_list_in_dict = {
+                key: [dict_[key] for dict_ in cv_results_dict_in_list]
+                for key in cv_results_dict_in_list[0]
+            }
         return cv_results_list_in_dict
 
     @property
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -565,9 +565,9 @@ class OptunaSearchCV(BaseEstimator):
 
     @property
     def cv_results_(self) -> Dict[str, Any]:
-        """Return the dictionary has list of all trials.
+        """Return the dictionary has list of CV results of all trials.
         Returns:
-            A list of results of all trials.
+            A list of CV results of all trials.
         """
         cv_results_dict_in_list = [trial_.user_attrs for trial_ in self.trials_]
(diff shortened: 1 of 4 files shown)


optuna/study/study.py
```diff
@@ -174,6 +174,20 @@ def best_trials(self) -> List[FrozenTrial]:
...
+
+        Returns:
+            A list of results of all trials.
+        """
+        cv_results_dict_in_list = [trial_.user_attrs for trial_ in self.trials]
+        cv_results_list_in_dict = {
+            key: [dict_[key] for dict_ in cv_results_dict_in_list]
+            for key in cv_results_dict_in_list[0]
+        }
+        return cv_results_list_in_dict
```
toshihikoyanase, line 189: Could you consider moving this implementation to `optuna/integration/sklearn.py`?
`Trial.user_attrs` can hold a variety of data, not just CV results, so the current implementation might be specific to `OptunaSearchCV`.

toshihikoyanase (changes requested):
Thank you for your PR. I'm a bit hesitant about introducing a new method to the `Study` class, as this functionality seems to be specific to `OptunaSearchCV`. Please check the following comment.

toshihikoyanase:
@Alnusjaponica Could you join the review, please?

Alnusjaponica (changes requested):
@jckkvs You need to test the behavior of `OptunaSearchCV.cv_results_` too.

optuna/integration/sklearn.py
```diff
@@ -563,6 +563,19 @@ def classes_(self) -> OneDimArrayLikeType:
 
         return self.best_estimator_.classes_
 
+    @property
+    def cv_results_(self) -> Dict[str, Any]:
+        """Return the dictionary has list of all trials.
+        Returns:
+            A list of results of all trials.
+        """
```
toshihikoyanase, line 571: Sorry for the delayed response. Could you update the docstrings? I don't think this attribute returns all trials (i.e., `list[FrozenTrial]`). In my understanding, it returns a list of CV results corresponding to all the trials.

optuna/integration/sklearn.py
```diff
@@ -563,6 +563,19 @@ def classes_(self) -> OneDimArrayLikeType:
...
+        cv_results_dict_in_list = [trial_.user_attrs for trial_ in self.trials_]
+        cv_results_list_in_dict = {
+            key: [dict_[key] for dict_ in cv_results_dict_in_list]
+            for key in cv_results_dict_in_list[0]
```
Alnusjaponica, line 575: Is this safe when `len(cv_results_dict_in_list)==0`?

optuna/integration/sklearn.py
```diff
@@ -563,6 +563,22 @@ def classes_(self) -> OneDimArrayLikeType:
 
         return self.best_estimator_.classes_
 
+    @property
+    def cv_results_(self) -> Dict[str, Any]:
+        """Return the dictionary has list of CV results of all trials.
+        Returns:
+            A list of CV results of all trials.
+        """
```
toshihikoyanase, line 571: These lines seem to be duplicated. How about unifying it like the other properties?

```suggestion
        """A dictionary mapping a metric name to a list of Cross-Validation results of all trials."""
```

optuna/integration/sklearn.py
toshihikoyanase: Thanks!

optuna/study/study.py
toshihikoyanase: > your proposal contains typo

Oops! Sorry for bothering you. Please ignore my comment and overwrite it with your change.

toshihikoyanase (approved):
@Alnusjaponica Please merge after your review.

Alnusjaponica (approved):
Apologies for the delay. LGTM too.
@toshihikoyanase Could you merge this PR?

toshihikoyanase:
@Alnusjaponica Thank you for your review. Let me merge this PR.

Issue #4739, opened 2023-06-10
Add `qnevi_candidates_func` to the candidates_func of `optuna.integration.botorch`


kstoneriv3:
### Motivation

I would like to have yet another acquisition function implemented in the `optuna.integration.botorch`. So I'd like to discuss if it makes sense to support it, as we already have several candidatea functions implemented there.candidates

### Description

Right now, optuna supports `botorch.acquisitions.monte_carlo.qNoisyExpectedHypervolumeImprovement` but it does not support `botorch.acuiqision.monte_carlo.qNoisyExpectedImprovement` (the difference is about the absence of `Hypervolume`, which makes it multi-objective). This leads to an inconsistent offering of functionality as we do not have an API for noisy single-objective cases (ideally with batched optimization support). Additionally, though one may think that `qNoisyExpectedHypervolumeImprovement` should be able to handle a single-objective as its special case but it is not the case: https://github.com/pytorch/botorch/issues/1880

Therefore, I would like to implement `optuna.integration.botorch.qnevi_candidates_func` which wraps `botorch.acuiqision.monte_carlo.qNoisyExpectedImprovement`.

As `botorch.acuiqision.monte_carlo.LogNoisyExpectedImprovement` provides a more accurate acquisition function if we do not care about batched optimization, I can also implement this if there is an appetite for it as in #4667.

### Alternatives (optional)

There are a few options available:

1. Do not implement anything
2. Implement `qnevi_candidates_func` only
3. Implement `lognevi_candidates_func` only
4. Implement both `qnevi_candidates_func` and `lognevi_candidates_func`
5. Provide a more general API that can take a wider subset of `botorch.acquisition.*`

The last option would require a significant amount of change in the codebase so it would definitely require another standalone issue to discuss, but now that most of the candidates_funcs share very similar boilerplate code it may be time to consider such an option. If you are interested in the last option, I am happy to open another issue for it.

contramundum53:
Hmm, I'm not sure whether I understand the functionality correctly by reading https://botorch.org/api/acquisition.html.
Am I correct to say that the `NoisyExpectedImprovement` is better than `ExpectedImprovement` in that it tries to estimate the right baseline, rather than using the "best trial so far" which may happen to be too good?

contramundum53:
`lognevi_candidates_func` needs `FixedNoiseGP`, which is not available in Optuna (because we don't have a way to specify the noise).

contramundum53:
[...]

PR #4753, opened 2023-06-16
add `optuna.integration.botorch.qnei_candidates_func`

kstoneriv3:
## Motivation

#4739.

## Description of the changes

Add a new acquisition function `qnei_candidates_func` that complements the existing `qnehvi_candidates_func`, which is a multi-objective version of it (but it is strictly multi-objective and therefore does not cover the single objective case.) 

This PR will let users to take advantage of the noisy surrogate model in BO for a single objective case. For more details, 
please refer to #4739.

### DIFF
--- a/docs/source/reference/integration.rst
+++ b/docs/source/reference/integration.rst
@@ -27,6 +27,7 @@ BoTorch
 
    optuna.integration.BoTorchSampler
    optuna.integration.botorch.qei_candidates_func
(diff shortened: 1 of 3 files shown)


HideakiImamura (approved):
Thanks for the PR. LGTM.

contramundum53 (approved):
LGTM!

PR #4748, opened 2023-06-15
Implement hypervolume history plot for matplotlib backend

y0z:
## Motivation
This PR implements a hypervolume history plot function to visualize the achieved hypervolume at each trial to evaluate the performance of a sampler in multiobjective optimization for the matplotlib backend.

### Demo
```python
import optuna

import matplotlib.pyplot as plt


def objective(trial):
    x = trial.suggest_float("x", 0, 5)
    y = trial.suggest_float("y", 0, 3)

    v0 = 4 * x ** 2 + 4 * y ** 2
    v1 = (x - 5) ** 2 + (y - 5) ** 2
    return v0, -v1


study = optuna.create_study(directions=["minimize", "maximize"])
study.optimize(objective, n_trials=50)

reference_point=[140., -50.]
optuna.visualization.matplotlib.plot_pareto_front(study)
plt.scatter(x=reference_point[0], y=reference_point[1], color='k', marker='*')
plt.tight_layout()
plt.show()

optuna.visualization.matplotlib.plot_hypervolume_history(study, reference_point)
plt.tight_layout()
plt.show()
```

## Description of the changes
- Add _hypervolume_history.py

PR #4757, opened 2023-06-22
Add hypervolume history plot for plotly backend.

y0z:
## Motivation
This PR implements a hypervolume history plot function for the plotly backend.
This is a follow-up for #4748 that introduced the matplotlib version.

## Description of the changes
- Add _hypervolume_history.py to optuna.visualization

### DIFF
--- a/optuna/visualization/_hypervolume_history.py
+++ b/optuna/visualization/_hypervolume_history.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import NamedTuple
 from typing import Sequence
 
--- a/optuna/visualization/_hypervolume_history.py
+++ b/optuna/visualization/_hypervolume_history.py
@@ -1,0 +1,92 @@
+from typing import Sequence
+
+import numpy as np
+
+from optuna._experimental import experimental_func
+from optuna.study import Study
+from optuna.visualization._plotly_imports import _imports
+from optuna.visualization.matplotlib._hypervolume_history import _get_hypervolume_history_info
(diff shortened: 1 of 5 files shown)


optuna/visualization/_hypervolume_history.py
```diff
@@ -0,0 +1,92 @@
+from typing import Sequence
+
+import numpy as np
+
+from optuna._experimental import experimental_func
+from optuna.study import Study
+from optuna.visualization._plotly_imports import _imports
+from optuna.visualization.matplotlib._hypervolume_history import _get_hypervolume_history_info
+from optuna.visualization.matplotlib._hypervolume_history import _HypervolumeHistoryInfo
```
not522, line 9: Could you move these to this file?

not522 (commented):
Thank you for your PR! Please check my comment to align with other visualization function implementations.

optuna/visualization/_hypervolume_history.py
y0z: I've moved these implementations to this file.
The corresponding test file has also moved from matplotlib_tests to visualization_tests.

toshihikoyanase:
@HideakiImamura Could you review this PR, please?

not522 (approved):
LGTM!

HideakiImamura (approved):
LGTM!

Note that it would be great to add an unit-tests only test that it is possible to plot the figure in the regular setting like other visualization functions, but it is OK to work on as a follow-up.

PR #4758, opened 2023-06-22
Follow up PR #4642 by adding a unit test to confirm `OptunaSearchCV` behavior

Alnusjaponica:
## Motivation
Follow up #4642 by adding a unit test.

## Description of the changes
Add a unit test to confirm if `OptunaSearchCV` keeps optimizing even when some trial fails due to exception.


toshihikoyanase:
@not522 Could you review this PR, please?
I think we can merge this PR with a single approval since it only changes a unit test.

not522 (commented):
Could you add a comment that sklearn raises `ValueError` for `max_depth==0`?

Alnusjaponica:
I added code comments. PTAL.

not522 (approved):
LGTM!

PR #4763, opened 2023-06-23
Add FileSystemArtifactStore

gen740:
## Motivation

I am submitting this PR to implement the artifact feature.
As part of this, I have introduced the FileSystemArtifactStore from [optuna-dashboard](https://github.com/optuna/optuna-dashboard/blob/main/optuna_dashboard/artifact).

## Description of the changes

- Create artifact submodule.
- Add ArtifactStore Protocol
- Add FileSytemArtifactStore.


not522:
@c-bata @contramundum53 Could you review this PR?

optuna/artifact/exceptions.py
```diff
@@ -0,0 +1,9 @@
+class ArtifactNotFound(Exception):
```
c-bata, line 1: [nit] I'm wondering if the `ArtifactNotFound` exception should inherit the `OptunaError` exception. Any opinions on this?

optuna/artifact/_protocol.py
```diff
@@ -0,0 +1,59 @@
...
+
+class ArtifactStore(Protocol):
+    """A protocol defining the interface for an artifact backend.
+
+    An artifact backend is responsible for managing the storage and retrieval
+    of artifact data. The backend should provide methods for opening, writing
+    and removing artifacts.
+    """
+
+    def open(self, artifact_id: str) -> BinaryIO:
```
c-bata, line 24: I understand you took the implementation from Optuna Dashboard, which is added by me. But the name `open()` is a bit confusing since it looks as if it can open a file in write mode.

`get_reader()` or `open_reader()` might be more specific and clearer. @contramundum53 @gen740 What do you think?

c-bata (commented):
Thank you for your pull request.

I have a suggestion that how about making the module name plural, like `optuna.artifacts`, to make it consistent with other module names.

optuna/artifact/_protocol.py
gen740: I agree.

I favor the name `open_reader()` as it's more natural when used within a `with` statement's scope.

optuna/artifact/exceptions.py
gen740: As `StorageInternalError` inherits the `OptunaError`, it seems more natural to inherit `OptunaError`.

optuna/artifact/_protocol.py
c-bata: Sounds reasonable 👍

c-bata (approved):
LGTM with a nit!

contramundum53 (commented):
[Nit, can be follow-up] I think we should document what `artifact_id` is okay and what is not. For example, can the letter "/" be used in artifact_id? It helps organizing artifacts, but may result in non-existent directory error.

contramundum53 (approved):
LGTM.

PR #4769, opened 2023-06-26
Remove unused attributes in `TPESampler`

not522:
## Motivation
`self._prior_weight` and `self._weights` are not used.

## Description of the changes
- Remove `self._prior_weight` and `self._weights`


not522:
@HideakiImamura Could you review this PR? I think that one approval is enough to merge since the change is obvious.

HideakiImamura (approved):
LGTM. Good catch!

Issue #4771, opened 2023-06-27
BruteForceSampler crashes with named hyperparameters out of alphabetical order


dkalsan:
### Expected behavior

The `BruteForceSampler` should run regardless of the alphabetical order of `name`s on repeated invocations of `suggest_float(...)`.

### Environment

- Optuna version: 3.2.0
- Python version: 3.10.6
- OS: Linux-5.15.0-71-generic-x86_64-with-glibc2.35
- (Optional) Other libraries and their versions: /

### Error messages, stack traces, or logs

```shell
Traceback (most recent call last):
  File "/home/user/Desktop/minimal-example.py", line 14, in <module>
    study.optimize(objective, n_trials=100)
  File "/home/user/.virtualenvs/tmp/lib/python3.10/site-packages/optuna/study/study.py", line 443, in optimize
    _optimize(
  File "/home/user/.virtualenvs/tmp/lib/python3.10/site-packages/optuna/study/_optimize.py", line 66, in _optimize
    _optimize_sequential(
  File "/home/user/.virtualenvs/tmp/lib/python3.10/site-packages/optuna/study/_optimize.py", line 163, in _optimize_sequential
    frozen_trial = _run_trial(study, func, catch)
  File "/home/user/.virtualenvs/tmp/lib/python3.10/site-packages/optuna/study/_optimize.py", line 212, in _run_trial
    frozen_trial = _tell_with_warning(
  File "/home/user/.virtualenvs/tmp/lib/python3.10/site-packages/optuna/study/_tell.py", line 174, in _tell_with_warning
    study.sampler.after_trial(study, frozen_trial, state, values)
  File "/home/user/.virtualenvs/tmp/lib/python3.10/site-packages/optuna/samplers/_brute_force.py", line 213, in after_trial
    tree = self._build_tree(
  File "/home/user/.virtualenvs/tmp/lib/python3.10/site-packages/optuna/samplers/_brute_force.py", line 167, in _build_tree
    leaf.set_leaf()
  File "/home/user/.virtualenvs/tmp/lib/python3.10/site-packages/optuna/samplers/_brute_force.py", line 53, in set_leaf
    self.expand(None, [])
  File "/home/user/.virtualenvs/tmp/lib/python3.10/site-packages/optuna/samplers/_brute_force.py", line 46, in expand
    raise ValueError(f"param_name mismatch: {self.param_name} != {param_name}")
```

### Steps to reproduce

```
import optuna
from optuna.samplers import BruteForceSampler


[...]

PR #4775, opened 2023-06-29
Sort params on fetch

contramundum53:
## Motivation
On `RDBStorage`, param order is not conserved in user API, although the database stores the order information in the form of `param_id`. This creates issues (e.g. #4771)  for certain samplers that use the order of parameters to analyze the parameter dependency relation in dynamic search space. We ensure that `RDBStorage` returns the parameters in the order they are suggested.

## Description of the changes
* Ensure that `trial.params` is in the order they are suggested, in `RDBStorage`.
* Test that for all storages, `trial.params` are in the right order.


not522:

contramundum53:
After a discussion, we decided not to require all storage classes to preserve the order of parameters. However, preserving the order of parameters is still helpful when it's possible, so I'll leave this PR open.

not522 (approved):
LGTM

gen740 (approved):
LGTM!

PR #4782, opened 2023-07-04
Fix BruteForceSampler

### LINKED ISSUE #4771: BruteForceSampler crashes with named hyperparameters out of alphabetical order

contramundum53:
## Motivation
Fix #4771.

BruteForceSampler currently assumes that the storage retains the order of suggested parameters. This is currently true for all storages except for RDBStorage, but we decided not to add a constraint in storage classes, because it would then be hard to implement a new storage based on key-value stores. Instead, we rewrote BruteForceSampler not to rely on the order of parameters.

The new BruteForceSampler builds a tree from the trial history by successively choosing one "pivot" parameter that has constant distribution in the current branch. Since the candidates for the "pivot" is tried in the order of parameters of a previous trial, the behavior (the probability for choosing each branch) remains exactly the same if the storage actually retains the order.

## Description of the changes
* Change BruteForceSampler to support the cases where the order of parameters are not retained in the storage.
* Add a test that changes the order of parameters during optimization.


c-bata:
@not522 @knshnb Could you review this PR if you have time?

knshnb:
I understand the algorithm. Let me take a detailed look next week 🙏

not522:
Sorry for my late response. I'm checking the correctness of the algorithm. I worry that I'm missing edge cases.
The stateless implementation has good points, but I think it's hard to maintain. How about reverting to the queue-based algorithm like the original implementation?

contramundum53:
I don't think the algorithm itself (whose complexity contributes to the difficulty of maintenance) is that more complex than the previous version, and using enqueue_trial to do DFS is also tricky in another way (especially when combined with parallel optimization). How about discussing the concerns offline?

contramundum53:
Would your concern be resolved if I put a comment describing the proof that this algorithm is correct?

contramundum53:
If you think this algorithm is significantly more complex than the previous (without this PR) algorithm, you can also keep it as it is, since we already have #4775 merged.

contramundum53:
After a discussion, we concluded that we can close this PR for now, since #4775 already fixes the critical problem.

not522:
Discussion log:

Considering the shortcomings of the queue-based algorithm, the tree-based algorithm is preferred.

- The queue needs to handle trials other than the one currently being sampled.
- It does not work with other samplers at the same time.

The PR's implementation can build a wrong search tree (even then, the sampling is correct), which makes the algorithm complex. The original issue #4771 is already fixed in the currently implemented storages, then we keep the current code and will come back here when this PR is necessary.

PR #4793, opened 2023-07-06
Add constraints support to `_optimization_history_plot`

hrntsm:
## Motivation

Currently, only the pareto front can visualize the results considering the constraints.
To make it easier to understand the results, we have added an optimized_history_plot to allow plots that consider the constraints.

## Description of the changes

Modified _optimization_history_plot.py to allow plots that reflect constraints.
Tests have been modified accordingly.

An example plot of the modified results is shown below.

```py
    v0 = 4 * x**2 + 4 * y**2
    trial.set_user_attr("constraint", [1000 - v0])
    return v0
```

r8MhiXGk

When calculating error bars(`plot_optimization_history(study, error_bar=True)`), only feasible trials are included in the calculation. 
Therefore, if an infeasible trial is included, output the following as warnings.
"Your study contains infeasible trials. In optimization history plot,error bars are calculated for only feasible trial values." 

If the constraints are strict, the error bar may not be displayed from about trial number 0 because the first feasible trial is not easily obtained.
The following example uses `trial.set_user_attr("constraint", [1000 - v0, x-10, y-10])`
Screenshot 2023-07-06 at 23 28 29

### DIFF
--- a/optuna/visualization/_optimization_history.py
+++ b/optuna/visualization/_optimization_history.py
@@ -143,7 +143,10 @@ def _get_optimization_history_info_list(
                 values_info = best_values_info
             for n, v, s in zip(trial_numbers, values_info.values, values_info.states):
                 if not math.isinf(v):
-                    values[n].append(v)
+                    if not use_best_value and s == _ValueState.Feasible:
+                        values[n].append(v)
+                    elif use_best_value:
+                        values[n].append(v)
                 states[n].append(s)
         trial_numbers_union: list[int] = []
         value_states: list[_ValueState] = []
--- a/optuna/visualization/_optimization_history.py
+++ b/optuna/visualization/_optimization_history.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from enum import Enum
 import math
 from typing import Callable
 from typing import cast
@@ -24,11 +25,17 @@ if _imports.is_successful():
 _logger = get_logger(__name__)
 
 
+class _ValueState(Enum):
+    Feasible = 0
+    Infeasible = 1
+    Incomplete = 2
+
+
 class _ValuesInfo(NamedTuple):
     values: list[float]
     stds: list[float] | None
     label_name: str
-    states: list[str]
+    states: list[_ValueState]
 
 
 class _OptimizationHistoryInfo(NamedTuple):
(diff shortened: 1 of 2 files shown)


not522 (commented):
Thank you for your PR! It's a desirable feature for supporting constrained optimization in Optuna.
I took a quick look. Please check my comment.

optuna/visualization/_optimization_history.py
```diff
@@ -26,6 +28,7 @@ class _ValuesInfo(NamedTuple):
     values: list[float]
     stds: list[float] | None
     label_name: str
+    states: list[str]
```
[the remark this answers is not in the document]
hrntsm: Thank you for your comment!
It is better that way. Will fix.
I'm checking other optuna implementations and see Enum using both `Enum` and `IntEnum`, are there any rules?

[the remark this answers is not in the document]
not522: As far as I know, there are no strict rules for `Enum` or `IntEnum`. I think `Enum` is enough here.

optuna/visualization/_optimization_history.py
```diff
@@ -22,10 +25,17 @@
 _logger = get_logger(__name__)
 
 
+class _ValueState(Enum):
```
hrntsm, line 28: I created an enum representing the state of the value

not522 (approved):
Thank you for your update! It looks good to me!
The test has room for improvement but can be followed up with other PRs. (It's OK adding tests in this PR if you have time.)
- Test creating `_OptimizationHistoryInfo` with infeasible trials
- Test warnings

optuna/visualization/_optimization_history.py
```diff
@@ -48,35 +58,66 @@ def _get_optimization_history_info_list(
...
+                values.append(float("nan"))
+                value_states.append(_ValueState.Incomplete)
+                continue
+            has_constraints = _CONSTRAINTS_KEY in trial.system_attrs
+            constraints_value = trial.system_attrs[_CONSTRAINTS_KEY] if has_constraints else [0]
+            value_states.append(
+                _ValueState.Feasible
+                if all(map(lambda x: x <= 0.0, constraints_value))
+                else _ValueState.Infeasible
+            )
```
c-bata, line 76: Can we simplify this logic like below?

```suggestion
            constraints = trial.system_attrs.get(_CONSTRAINTS_KEY)
            if constraints is None or all([x <= 0.0 x for x in constraints]):
                value_states.append(_ValueState.Feasible)
            elsel:
                value_states.append(_ValueState.Infeasible)
```

c-bata (commented):
@hrntsm Thank you for your pull request. I left one suggestion.

On the whole, it looks good, but I'm considering whether we can refactor the `_aggregate` method. We could tackle it as a follow-up, but let me think the way and send the suggestion if I will have an idea.

optuna/visualization/_optimization_history.py
hrntsm: This one is simpler and better. I will fix it.

hrntsm:
Simplified _aggregate. I think it is better than before.

c-bata (commented):
Looks good to me! I will approve after fixing the CI failures. Please wait until #4809 will be merged.

c-bata (approved):
LGTM!

PR #4760, opened 2023-06-22
Separate after_trial strategy

Alnusjaponica:
## Motivation
Progress v3.3 task to make `NSGAII Sampler` loose coupled.

## Description of the changes
Add `after_trial_strategy` argument to the constructor if `NSGAIISampler` and `NSGAIIISampler` and separate the after_trial process so that users can customize `after_trial()`.

PR #4803, opened 2023-07-11
Follow up separation of after trial strategy

Alnusjaponica:
## Motivation
Follow up https://github.com/optuna/optuna/pull/4760
## Description of the changes
- Rearrange the order of the arguments of `NSGAIIISampler` constructor.
- Make `NSGAIISampler` and `NSGAIIISampler` return `ExperimentalWarning` when `after_trial_strategy` is given in initialization.
- Add test for `after_trial_strategy` of `NSGAIIISampler`.

### DIFF
--- a/optuna/samplers/_nsgaiii.py
+++ b/optuna/samplers/_nsgaiii.py
@@ -118,6 +118,12 @@ class NSGAIIISampler(BaseSampler):
                 " The interface can change in the future.",
                 ExperimentalWarning,
             )
+        if after_trial_strategy is not None:
+            warnings.warn(
+                "The after_trial_strategy option is an experimental feature."
+                " The interface can change in the future.",
+                ExperimentalWarning,
+            )
 
         if crossover is None:
             crossover = UniformCrossover(swapping_prob)
--- a/optuna/samplers/_nsgaiii.py
+++ b/optuna/samplers/_nsgaiii.py
@@ -77,6 +77,7 @@ class NSGAIIISampler(BaseSampler):
 
     def __init__(
         self,
+        *,
         population_size: int = 50,
         mutation_prob: float | None = None,
         crossover: BaseCrossover | None = None,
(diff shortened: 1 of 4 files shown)


optuna/samplers/_nsgaiii.py
```diff
@@ -86,12 +86,12 @@ def __init__(
         swapping_prob: float = 0.5,
         seed: int | None = None,
         constraints_func: Callable[[FrozenTrial], Sequence[float]] | None = None,
+        reference_points: np.ndarray | None = None,
+        dividing_parameter: int = 3,
```
contramundum53, line 90: Nit: we have this many arguments, so how about making these parameters keyword-only?

contramundum53:

optuna/samplers/_nsgaiii.py
```diff
@@ -118,6 +118,12 @@ def __init__(
                 " The interface can change in the future.",
                 ExperimentalWarning,
             )
+        if after_trial_strategy is not None:
+            warnings.warn(
+                "The after_trial_strategy option is an experimental feature."
+                " The interface can change in the future.",
+                ExperimentalWarning,
+            )
```
not522, line 126: I think this warning is not necessary because all of `NSGAIIISampler` is experimental.

optuna/samplers/_nsgaiii.py
contramundum53: We had a discussion and we decided that it is okay to break the existing interface and make the arguments keyword-only.

not522 (approved):
LGTM except for @contramundum53's comment.

optuna/samplers/_nsgaiii.py
Alnusjaponica: I updated the arguments in the same way as  https://github.com/optuna/optuna/pull/4806/commits/fd761a7dcfb45d280baea1873e54923c01f0fdd8. 
Could you take another look?

contramundum53 (approved):
LGTM!

PR #4816, opened 2023-07-18
Add constraints support to `matplotlib._optimization_history_plot`

hrntsm:
## Motivation

Reflect #4793  updates to matplotlib part

## Description of the changes

Constraints are applied in the same way as plotly

Screenshot 2023-07-18 at 22 00 41

Screenshot 2023-07-18 at 22 00 58

### DIFF
--- a/optuna/visualization/matplotlib/_optimization_history.py
+++ b/optuna/visualization/matplotlib/_optimization_history.py
@@ -120,8 +120,8 @@ def _get_optimization_history_plot(
                 fmt="o",
                 color="tab:blue",
             )
-            infeasible_trial_numbers = []
-            infeasible_trial_values = []
+            infeasible_trial_numbers: list[int] = []
+            infeasible_trial_values: list[float] = []
         else:
             feasible_trial_numbers = [
                 n for n, s in zip(trial_numbers, values_info.states) if s == _ValueState.Feasible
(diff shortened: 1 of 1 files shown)


hrntsm:
I noticed in the PR below that the matplotlib side also has markers on the best value, but the plotly side only has lines. Should I match one to the other?
https://github.com/optuna/optuna-visualization-regression-tests/pull/13

c-bata:
> I noticed in the PR below that the matplotlib side also has markers on the best value, but the plotly side only has lines. Should I match one to the other? [optuna/optuna-visualization-regression-tests#13](https://github.com/optuna/optuna-visualization-regression-tests/pull/13)

Good catch 👍 I don't think it's a high priority, but we can align the implementation between plotly and matplotlib. In Optuna Dashboard, the history plot do not render all makers on the best value line since it increases the calculation cost when a lot of trials exist. So if we have to choose one, I prefer plotly's approach.

Screenshot 2023-07-19 11 36 47

c-bata (approved):
LGTM!


Here is an output image of https://github.com/optuna/optuna-visualization-regression-tests/pull/13.

Screenshot 2023-07-19 12 07 29

hrntsm:
Fixed best value does not include markers.
Screenshot 2023-07-19 at 12 44 36
Screenshot 2023-07-19 at 12 44 49

HideakiImamura (approved):
Thanks for the PR. LGTM!

Issue #2792, opened 2021-07-08
Is there any way to prune a trial with multi-objective study?


rsek147:
Not implemented: https://github.com/optuna/optuna/blob/master/optuna/trial/_trial.py#L569-L572

Do you have a plan to do this?

nzw0301:
Hi, thank you for asking the question! Such functionality is definitely interesting but a well-known algorithm might not exist yet in the machine learning community as far as I know. Do you know any pruning algorithm for multi-objective?

jeremycochoy:
Just ran into the `NotImplementedError` of `should_prune`. Even if there is no "pruning algorithm for multi-objective" in the literature, I would be really pleased by the possibility to prune according to a single objective (PercentilePruner) or even both objective (PercentilePruner condition met on all objectives).

Also, having an exception on the should prune is a bit disturbing, as I had to go to the source code of should_prune and simple guard my call with a copy past of the `len(optuna_trial.study.directions)` check... 😅 

I am really tempted by hacking a very naive implementation that could suit my simple usage, and I think it make more sense if I make it available for all optuna users. How would you write this? What about a pruner wrapper that would project on the `n` coordinate of the objective function so that it enable access to all existing pruners?

rsek147:
I think if determining whether the trial is pareto front before the end of the trial, it is possible to prune it. But I don't know how to implement the algorithm...

jeremycochoy:
You can say that if it is bellow (for the partial order on n-uplets) existing trials, then it is not pareto front (at the given iteration you are comparing scores, of course) and you could prune it.

y0z:
The following paper proposes a pruning algorithm for multiobjective optimization.
https://ml.informatik.uni-freiburg.de/papers/21-ARXIV-BBMoNASHPO.pdf

simpsus:
I also just ran into this error. In Single objective optimization, the trial.should_prune() is the main reason a lot of unnecessary trials are skipped. With multi-objective, that is simply not available.

Even something very trivial that would say for the median pruner:
If both values are below their respective median ==> prune
Would help a great deal

PfeifferMicha:
I'd like to re-open this discussion. MO pruning seems like an important feature...

I'd also like to add another vote for any very simple pruning strategy ("all values meet the single-value pruning criterion individually" or "one of the values meets the pruning criterion").

simpsus:
It would even be a start if trial.report would support multiple values. What I have resorted to is reimplementing the storing of trial values ...

[...]

PR #4822, opened 2023-07-21
Update type of `intermediate_value` from `float` to `Union[float, Sequence[float]]`

cross32768:
## Motivation
This PR is the splitted version of https://github.com/optuna/optuna/pull/4652.
The original motivation of https://github.com/optuna/optuna/pull/4652 is to allow `trial.report` to take multiple intermediate values, which is required by multiple users in https://github.com/optuna/optuna/issues/3450 and https://github.com/optuna/optuna/issues/2792.

To keep the PR size small, first this PR updates type of `intermediate_value` from `float` to `Union[float, Sequence[float]]`.
I will create the follow-up PR to update interface of `trial.report` to take multiple intermediate values.
 

## Description of the changes
Modify the codebase to update type of `intermediate_value` from `float` to `Union[float, Sequence[float]]`.

Almost all code is from https://github.com/optuna/optuna/pull/4652 by @nzw0301 .


not522:

optuna/samplers/_cmaes.py
```diff
@@ -736,6 +736,9 @@ def _get_trials(self, study: "optuna.Study") -> List[FrozenTrial]:
                 _, value = max(t.intermediate_values.items())
                 if value is None:
                     continue
+
+                assert isinstance(value, float)
```
HideakiImamura, line 740: If the multi-dimensional intermediate values are given, this line causes an AssertionError. This means the sampler implicitly assumes that intermediate values are one-dimensional. For instance, the pruner may issue a warning in such cases, but is it okay for the sampler not to explicitly produce error messages or warnings?

HideakiImamura: This comment can be applied to other samplers.

optuna/integration/dask.py
```diff
@@ -319,7 +319,7 @@ def set_trial_intermediate_value(
         storage_name: str,
         trial_id: int,
         step: int,
-        intermediate_value: float,
+        intermediate_value: Union[float, Sequence[float]],
```
HideakiImamura, line 322: Instead of using `Union`, it is more modern to use `|`, and Optuna is gradually adopting it. Could you use `|` in this PR as well? Additionally, instead of the `typing` version, could you use `collections.abc` for `Sequence`?

```suggestion
        intermediate_value: float | Sequence[float],
```

HideakiImamura: This comment will be applied to other files.

optuna/storages/_rdb/storage.py
```diff
@@ -866,6 +882,27 @@ def _build_frozen_trial_from_trial_model(self, trial: "models.TrialModel") -> Fr
             values = None
 
         params = sorted(trial.params, key=lambda p: p.param_id)
+        unsorted_intermediates: Dict[int, Dict[int, float]] = defaultdict(dict)
```
HideakiImamura, line 885: Please use `dict` instead of `Dict`.

optuna/storages/_rdb/storage.py
```diff
@@ -866,6 +882,27 @@ def _build_frozen_trial_from_trial_model(self, trial: "models.TrialModel") -> Fr
...
+                for step, intermediate_indices_and_values in unsorted_intermediates.items():
+                    intermediate_values[step] = tuple(
+                        intermediate_value
+                        for _, intermediate_value in sorted(
+                            intermediate_indices_and_values.items()
+                        )
+                    )
+            else:
+                for step, intermediate_indices_and_values in unsorted_intermediates.items():
+                    intermediate_values[step] = intermediate_indices_and_values[0]
```
HideakiImamura, line 906: How about adding tests cases for these logics?

HideakiImamura (commented):

optuna/multi_objective/trial.py
```diff
@@ -299,8 +299,9 @@ def __init__(self, n_objectives: int, trial: FrozenTrial):
         self.n_objectives = n_objectives
         self._trial = trial
 
-        self.values = tuple(trial.intermediate_values.get(i) for i in range(n_objectives))
-
+        self.values: Sequence[Optional[float]] = tuple(
```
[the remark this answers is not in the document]
cross32768: Yes, this module is deprecated. But I think simply reverting this change causes mypy error. Thus, I consider to add "#type: ignore".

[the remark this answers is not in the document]
HideakiImamura: Sounds good to me. Let me resolve this comment.

optuna/samplers/_cmaes.py
```diff
@@ -767,6 +767,14 @@ def _get_trials(self, study: "optuna.Study") -> List[FrozenTrial]:
                 _, value = max(t.intermediate_values.items())
                 if value is None:
                     continue
+
+                if not isinstance(value, float):
+                    warnings.warn(
+                        "This sampler doesn't support multiple intermediate values. "
+                        "Skip this trial."
+                    )
```
HideakiImamura, line 775: I noticed that this warning message is repeatedly being displayed. Would it be possible to limit the output of the message to at most once per trial, or alternatively, consider not displaying the message at all? I apologize for any confusion caused.

tests/storages_tests/rdb_tests/test_storage.py
HideakiImamura, line 1: How about adding `v3.4.0.a` to all upgrade test?

HideakiImamura (commented):

optuna/samplers/_cmaes.py
cross32768: I see. For now, I deleted sampler-related warnings once and for all.

tests/storages_tests/rdb_tests/test_storage.py
cross32768: I added test and db for test,
The new db "3.4.0.a.db" is generated by `optuna storage upgrade --storage sqlite:///3.2.0.a.db`. Is it the correct procedure?

optuna/storages/_rdb/storage.py
cross32768: Thank you for your comment. I added test cases for these logics in tests/storages_tests/rdb_tests/test_storage.py.

not522:
Sorry for my late response.
I'm wondering how about aligning the handling `value` and `intermediate_values` in `FrozenTrial`, i.e., `FrozenTrial` has `self._intermediate_values: dict[int, Sequence[float]]` and convert it to `dict[int, float | Sequence[float]]` in the property method. This approach may reduce the complexity of storages.

cross32768:
Ok, I'm going to try it.
But sorry maybe it takes some time because it is required to radically change the code.

cross32768:
@not522 
I have tried your idea, but I found that some code substitute `frozen_trial.intermediate_values` by code as below:
 `frozen_trial.intermediate_values[step] = value`
 
 Because of indexing by `step`, I think it is difficult to convert type of this variable by property method.
 
 As you said, it is better to consider simpler way to hold `intermediate_values`, but could we consider this problem as the follow-up PR? I think to do this with consistent code, we need some breaking change (i.e. the interface of `set_trial_intermediate_values`)

cross32768:
@nzw0301 
I tried to introduce new feature to deal with multiple values in `intermediate_value`, but I found that it is not so easy as I wrote below.
https://github.com/optuna/optuna/issues/3450#issuecomment-1706109413

I'm going to close this PR at this time, and we consider the different implementation policy for this feature in future.

nzw0301:
@cross32768 thank you for your hard work and for letting me know that.

PR #4826, opened 2023-07-21
Fix checks integration

Alnusjaponica:
## Motivation
Fix current failure of `Checks(Integration)` and progress https://github.com/optuna/optuna/issues/4508.

## Description of the changes
- `integration/_lightgbm_tuner/optimize.py`
  Properly add assert sentence to pass mypy check
- `integration_tests/lightgbm_tuner_tests/test_optimize.py`
  Replace dummy object with mock object
- Both
  Apply type annotation update

See https://github.com/Alnusjaponica/optuna/pull/21 to check the result.

### DIFF
--- a/optuna/integration/_lightgbm_tuner/optimize.py
+++ b/optuna/integration/_lightgbm_tuner/optimize.py
@@ -1,16 +1,17 @@
+from __future__ import annotations
+
 import abc
+from collections.abc import Callable
+from collections.abc import Container
+from collections.abc import Generator
+from collections.abc import Iterator
 import copy
 import json
 import os
 import pickle
 import time
 from typing import Any
-from typing import Callable
 from typing import cast
-from typing import Container
-from typing import Dict
-from typing import Generator
-from typing import Iterator
 from typing import List
 from typing import Optional
 from typing import Tuple
@@ -64,8 +65,8 @@ _logger = optuna.logging.get_logger(__name__)
 class _BaseTuner:
     def __init__(
         self,
-        lgbm_params: Optional[Dict[str, Any]] = None,
-        lgbm_kwargs: Optional[Dict[str, Any]] = None,
+        lgbm_params: dict[str, Any] | None = None,
+        lgbm_kwargs: dict[str, Any] | None = None,
     ) -> None:
         # Handling alias metrics.
         if lgbm_params is not None:
@@ -92,7 +93,7 @@ class _BaseTuner:
 
     def _get_booster_best_score(self, booster: "lgb.Booster") -> float:
         metric = self._get_metric_for_objective()
-        valid_sets: Optional[VALID_SET_TYPE] = self.lgbm_kwargs.get("valid_sets")
+        valid_sets: VALID_SET_TYPE | None = self.lgbm_kwargs.get("valid_sets")
 
         if self.lgbm_kwargs.get("valid_names") is not None:
             if type(self.lgbm_kwargs["valid_names"]) is str:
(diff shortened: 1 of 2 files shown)


toshihikoyanase:
@eukaryo Could you review this PR, please?

optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -1,18 +1,18 @@
+from __future__ import annotations
+
 import abc
+from collections.abc import Callable
+from collections.abc import Container
+from collections.abc import Generator
+from collections.abc import Iterator
```
toshihikoyanase, line 7: I think most of Optuna codebase uses `typing.Callable`. Do you think these changes are necessary to fix the CI?

optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -479,7 +482,7 @@ def best_params(self) -> Dict[str, Any]:
             params.update(self.lgbm_params)
             return params
 
-    def get_best_booster(self) -> "lgb.Booster":
+    def get_best_booster(self) -> "lgb.Booster" | "lgb.CVBooster":
```
toshihikoyanase, line 485: [Question] Does this method returns `lgb.CVBooster` instances? Or, do you add `lgb.CVBooster` class only to fix the CI?

toshihikoyanase (commented):
Thank you for your PR. I have two questions.

optuna/integration/_lightgbm_tuner/optimize.py
Alnusjaponica: This is not necessary to fix CI but just followed [PEP584](https://peps.python.org/pep-0585/) as written in https://github.com/optuna/optuna/issues/4508. If it is better to revert this part, let me know.

optuna/integration/_lightgbm_tuner/optimize.py
Alnusjaponica: In my understanding, this method does not return `lgb.CVBooster` when called from subclass `LightGBMTuner`. However, `get_best_booster` is overwritten in subclass `LightGBMTunerCV` and without this change causes mypy error. 
Shall I move the method from `_LightGBMBaseTuner` to `LightGBMTuner`?

optuna/integration/_lightgbm_tuner/optimize.py
toshihikoyanase: Thank you for letting me know. As mentioned in [PEP585](https://peps.python.org/pep-0585/), we can keep the current changes since the corresponding classes in the `typing` are deprecated.

> collections.abc.Callable
> collections.abc.Container
> collections.abc.Generator
> collections.abc.Iterator
> Importing those from typing is deprecated.

optuna/integration/_lightgbm_tuner/optimize.py
Alnusjaponica: Sure. I fixed it.

optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -897,6 +862,45 @@ def _create_objective(
...
+        """Return the best booster.
+
+        If the best booster cannot be found, :class:`ValueError` will be raised. To prevent the
+        errors, please save boosters by specifying the ``model_dir`` argument of
+        :meth:`~optuna.integration.lightgbm.LightGBMTuner.__init__`,
+        when you resume tuning or you run tuning in parallel.
+        """
+        if self._best_booster_with_trial_number is not None:
+            if self._best_booster_with_trial_number[1] == self.study.best_trial.number:
+                assert isinstance(self._best_booster_with_trial_number[0], lgb.Booster)
```
toshihikoyanase, line 875: This assertion might not be necessary anymore.
```suggestion
```

toshihikoyanase (approved):
Thank you for your update. LGTM, except for a redundant line.

optuna/integration/_lightgbm_tuner/optimize.py
Alnusjaponica: You're right. I remove the sentence.

eukaryo (approved):
LGTM too.

PR #4838, opened 2023-07-26
Remove OrderedDict

### LINKED ISSUE #4592: Remove `OrderedDict` after Python 3.7
### Motivation

From the Python documentation

```
Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6.
```

https://docs.python.org/3.7/library/stdtypes.html#dict

Python `dict` now preserve their order and we could reduce the usage of `OrderedDict`.
However, we don't want to break any compatibilities.

### Description

It may be reasonable to at least follow up https://github.com/optuna/optuna/pull/4514 with a PR to always return a dictionary, which is of type `dict` and the order is always sorted. This is new not yet released API and does not have any compatibility issues at the time or writing. Performance regression should perhaps be checked for in this case.

_`contribution-welcome` when #4514 is merged._

### Alternatives (optional)

_No response_

_No response_

taniokay:
## Motivation

Resolve #4592

## Description of the changes

- Replace existing `OrderedDict` with `dict`
- Fix related tests and docstrings.

### DIFF
--- a/optuna/importance/__init__.py
+++ b/optuna/importance/__init__.py
@@ -1,4 +1,3 @@
-from collections import OrderedDict
 from typing import Callable
 from typing import Dict
 from typing import List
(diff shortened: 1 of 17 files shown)


toshihikoyanase:
@Alnusjaponica @not522 Could you review this PR, please?

Alnusjaponica (approved):
LGTM. Thanks for your contribution.

not522 (approved):
Thank you! LGTM!

PR #4839, opened 2023-07-27
Rename test directory

Alnusjaponica:
## Motivation
Apply name rule.

## Description of the changes
Rename directory name from `tests/articrafts` to `tests/articrafts_tests`


toshihikoyanase (approved):
Good catch! This PR only moves the test file, so let me merge this PR with a single approval.

[Notes] I confirmed that the test was executed as expected.
> tests/artifacts_tests/test_filesystem.py ..                              [  8%]

PR #4823, opened 2023-07-21
Add `upload_artifact` api

gen740:
## Motivation

This PR is a follow-up to #4763. It introduces the `upload_artifact` function which uploads an arbitary file to the Artifact Backend.

## Description of the changes

- Add the `upload_artifact` user API.
- Add the `upload_artifact` tests.

PR #4840, opened 2023-07-27
Add `Boto3ArtifactStore`

Alnusjaponica:
## Motivation
Migrate `Boto3ArtifactStore` from `optuna-dashboard` to `optuna`.

## Description of the changes
Move `Boto3Backend`, related dependency and tests from `optuna-dashboar` to `optuna`


optuna/artifacts/_filesystem.py
```diff
@@ -12,24 +12,7 @@
...
 class FileSystemArtifactStore:
-    """An artifact backend for file systems.
-
-    Example:
-       .. code-block:: python
-
-           import optuna
-           from optuna_dashboard.artifact import upload_artifact
-           from optuna_dashboard.artifact.file_system import FileSystemBackend
-
```
c-bata, line 23: Just for the record, here we removed an example from the docstring since this example uses optuna-dashboard APIs. As a follow-up, we should add an example again after merging #4823.

optuna/artifacts/_boto3.py
```diff
@@ -0,0 +1,71 @@
...
+        # This flag is added to avoid that upload_fileobj() method of Boto3 client may close the
+        # source file object. See https://github.com/boto/boto3/issues/929.
+        self._avoid_buf_copy = avoid_buf_copy
+
+    def open_reader(self, artifact_id: str) -> BinaryIO:
+        try:
+            obj = self.client.get_object(Bucket=self.bucket, Key=artifact_id)
+        except ClientError as e:
+            if _is_not_found_error(e):
+                raise ArtifactNotFound("not found") from e
```
c-bata, line 36: [nit] I've received feedback from Optuna Dashboard users suggesting that it's unclear which artifact wasn't found. It might be out of the scope of this PR, but what do you think about including the artifact_id in the error message?

```suggestion
                raise ArtifactNotFound(f"artifact={artifact_id} is not found") from e
```

optuna/artifacts/_boto3.py
```diff
@@ -0,0 +1,71 @@
...
+            buf.seek(0)
+            fsrc = buf
+        self.client.upload_fileobj(fsrc, self.bucket, artifact_id)
+
+    def remove(self, artifact_id: str) -> None:
+        try:
+            self.client.delete_object(Bucket=self.bucket, Key=artifact_id)
+        except ClientError as e:
+            if _is_not_found_error(e):
+                raise ArtifactNotFound("not found") from e
```
c-bata, line 56: ditto.

```suggestion
                raise ArtifactNotFound(f"artifact={artifact_id} is not found") from e
```

optuna/artifacts/_boto3.py
```diff
@@ -0,0 +1,71 @@
...
+
+def _is_not_found_error(e: ClientError) -> bool:
+    error_code = e.response.get("Error", {}).get("Code")
+    http_status_code = e.response.get("ResponseMetadata", {}).get("HTTPStatusCode")
+    return error_code == "NoSuchKey" or http_status_code == 404
+
+
+if TYPE_CHECKING:
+    # A mypy-runtime assertion to ensure that BackendLocalArtifactBackend implements all abstract
+    # methods in ArtifactBackendProtocol.
```
c-bata, line 68:
```suggestion
    # A mypy-runtime assertion to ensure that Boto3ArtifactStore implements all abstract
    # methods in ArtifactStore.
```

tests/artifacts_tests/test_boto3.py
```diff
@@ -0,0 +1,64 @@
...
+
+import boto3
+from moto import mock_s3
+import pytest
+
+from optuna.artifacts import Boto3ArtifactStore
+from optuna.artifacts.exceptions import ArtifactNotFound
+
+
+class TestBoto3ArtifactStore:
```
c-bata, line 14: What do you think about writing these tests in a function-based manner since Optuna uses pytest runner?

c-bata (commented):
Thank you for your pull request. Looks almost good to me, but I left some suggestions.

optuna/artifacts/_filesystem.py
Alnusjaponica: Thanks for the comment. I opened https://github.com/optuna/optuna/issues/4843 to track this.

optuna/artifacts/_boto3.py
Alnusjaponica: Fair enough. Is it better to add bucket name as well?
```suggestion
                raise ArtifactNotFound(f"bucket: {self.bucket}, artifact: {artifact_id} was not found") from e
```

c-bata: Sounds great to me 👍

toshihikoyanase:
@HideakiImamura Could you review this PR, please?

tests/artifacts_tests/test_boto3.py
Alnusjaponica: I reformat them.

optuna/artifacts/_boto3.py
Alnusjaponica: I added some more explanation. Please take another look to confirm if they sounds natural.

c-bata (approved):
LGTM after passing CI.

optuna/artifacts/_boto3.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+if TYPE_CHECKING:
+    from typing import BinaryIO
+
+    from mypy_boto3_s3 import S3Client
+
+
+class Boto3ArtifactStore:
+    """An artifact backend for Boto3."""
```
HideakiImamura, line 20: How about adding the explanations for arguments?

optuna/artifacts/_filesystem.py
```diff
@@ -12,24 +12,7 @@
 
 
 class FileSystemArtifactStore:
-    """An artifact backend for file systems.
```
HideakiImamura, line 15: ditto.

tests/artifacts_tests/test_boto3.py
```diff
@@ -0,0 +1,85 @@
...
+    objects = s3_client.list_objects(Bucket=bucket_name).get("Contents", [])
+    assert len([obj for obj in objects if obj["Key"] == artifact_id]) == 0
+
+
+def test_file_not_found_exception(
+    init_mock_client: Annotated[tuple[str, S3Client], pytest.fixture]
+) -> None:
+    bucket_name, _ = init_mock_client
+    backend = Boto3ArtifactStore(bucket_name)
+    with pytest.raises(ArtifactNotFound):
```
HideakiImamura, line 84: How about testing the case for `remove`?

tests/artifacts_tests/test_boto3.py
```diff
@@ -0,0 +1,85 @@
...
+        objects = s3_client.list_objects(Bucket=bucket_name).get("Contents", [])
+        if objects:
+            s3_client.delete_objects(
+                Bucket=bucket_name,
+                Delete={"Objects": [{"Key": obj["Key"] for obj in objects}], "Quiet": True},
+            )
+        s3_client.delete_bucket(Bucket=bucket_name)
+
+
+def test_upload_download(
```
HideakiImamura, line 43: How about adding the test case for `avoid_buf_copy` using `pytest.mark.parametrize`?

Alnusjaponica: I added test cases.

tests/artifacts_tests/test_boto3.py
Alnusjaponica: I added test to check `remove`

optuna/artifacts/_boto3.py
Alnusjaponica: Thanks for the comments. I added more explanations.

optuna/artifacts/_filesystem.py
Alnusjaponica: ditto

optuna/artifacts/_boto3.py
```diff
@@ -0,0 +1,91 @@
...
+    def remove(self, artifact_id: str) -> None:
+        try:
+            self.client.delete_object(Bucket=self.bucket, Key=artifact_id)
+        except ClientError as e:
+            if _is_not_found_error(e):
+                raise ArtifactNotFound(
+                    f"Artifact storage with bucket: {self.bucket}, artifact_id: {artifact_id} was"
+                    " not found"
+                ) from e
+            raise
```
Alnusjaponica, line 77: @c-bata 
It seems `client.delete_object` does not raise error when nothing found (See [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/delete_object.html#:~:text=Removes%20the%20null%20version%20(if%20there%20is%20one)%20of%20an%20object%20and%20inserts%20a%20delete%20marker%2C%20which%20becomes%20the%20latest%20version%20of%20the%20object.%20If%20there%20isn%E2%80%99t%20a%20null%20version%2C%20Amazon%20S3%20does%20not%20remove%20any%20objects%20but%20will%20still%20respond%20that%20the%20command%20was%20successful.)). Shall I just simply refactor like this?
```suggestion
        self.client.delete_object(Bucket=self.bucket, Key=artifact_id)
```

c-bata: @Alnusjaponica Thank you for the delayed response. Good catch. It sounds good 👍

HideakiImamura:
Ops, could you try to resolve CI errors?

optuna/artifacts/_boto3.py
c-bata: @Alnusjaponica It's just an implementation detail that `Boto3Backend.remove()` does not raise an `ArtifactNotFound` exception, so there is no need to make it consistent with other ArtifactStores. If the `ArtifactNotFound` exception is not beneficial for the users, we might need to consider revising the following specification, but let's address it in a follow-up PR.

https://github.com/optuna/optuna/blob/master/optuna/artifacts/_protocol.py#L49-L59

Alnusjaponica: I see. Thanks for the comments. I mark this conversation as resesolved.

HideakiImamura (approved):
LGTM.

PR #4846, opened 2023-07-28
Remove `ordered_dict` argument from `IntersectionSearchSpace`

not522:
## Motivation
By #4838, `OrderedDict` is removed from Optuna's code. Therefore the `ordered_dict` argument is now inappropriate.

## Description of the changes
This PR removes the `ordered_dict` argument and makes the returned dict always sorted.

### DIFF
--- a/optuna/importance/_base.py
+++ b/optuna/importance/_base.py
@@ -71,7 +71,7 @@ def _get_distributions(study: Study, params: Optional[List[str]]) -> Dict[str, B
     _check_evaluate_args(completed_trials, params)
 
     if params is None:
-        return intersection_search_space(study.get_trials(deepcopy=False), ordered_dict=True)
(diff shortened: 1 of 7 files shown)


not522:
@Alnusjaponica @contramundum53 Could you review this PR?

Alnusjaponica:
LGTM.
This is just a question but I just want to double check
- whether you intentionally left `samplers/_search_space/intersection.py` unchanged:
- and why it is necessary to sort items.

contramundum53 (approved):
LGTM.

not522:
> whether you intentionally left samplers/_search_space/intersection.py unchanged

Yes, it's already deprecated, so no need to update.

> why it is necessary to sort items in the first place.

`importance` features assumes that the order of parameters is fixed. The current implementation sorts `trial.distributions` but it is not the specification. (ref #4775)

PR #4848, opened 2023-07-31
Display best objective value in contour plot for a given param pair, not the value from the most recent trial

### LINKED ISSUE #4733: Display best objective value in contour plot for a given param pair, not the value from the most recent trial
### Motivation

I am optimizing 'learning rate', 'encoder', and 'dropout'. I happen to have 4 trails that all have encoder:resnet34 & dropout:0.1, each with a different learning rate:

image

The contour plot of 'encoder' vs. 'dropout' is displaying the objective value from the most recent trial (71) for resnet34 & dropout 0.1, not the best value (from trail 21). Ironically, trial 21 is my best overall trial, which is why I'm frustrated this isn't being reflected in the contour plot

image

This issue was mentioned in a comment here, but never followed up on https://github.com/optuna/optuna/issues/2524#issuecomment-970780116

### Description

When running a study with >2 parameters that are categorical and/or have step values, it is likely that there will be multiple trials with same combo of values for some of those parameters - i.e. in my example, there are 4 trials with resnet34 and dropout 0.1, but with different learning rates. Currently, the contour plot displays the objective value from the most recent trial.

Feature request: when generating contour plots, display the **_best_** objective value for a given combo of parameters, not the objective value from the most recent trial.

### Alternatives (optional)

It seems the only real work around at the moment is filtering by the best N trials, such as seen here https://github.com/optuna/optuna/issues/1402

This does work for my current issue, but the problem with this is that there's no way to ensure that I'm seeing the best performance across the full parameter space unless I do this independently for each parameter pair

image

_No response_

xadrianzetx:
## Motivation

This patch changes behavior of contour plot (both backends) when two or more trials params are overlapping. Closes #4733.

## Description of the changes

Currently, the objective value always comes from the last completed trial. This implementation selects the best objective value w.r.t. optimization direction (with reasonable default in multi-objective cases).

### DIFF
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -327,8 +327,8 @@ def _get_contour_subplot_info(
 
         existing = z_values.get((x_i, y_i))
         if existing is None or target is not None:
-            # In multi-objective case we don't know which objective was selected in target
-            # function therefore we don't know how to select the best objective value.
+            # When target function is present, we can't be sure what the z-value
+            # represents and therefore we don't know how to select the best one.
             z_values[(x_i, y_i)] = value
         else:
             z_values[(x_i, y_i)] = (
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -6,6 +6,7 @@ from typing import NamedTuple
 
 from optuna.logging import get_logger
 from optuna.study import Study
+from optuna.study import StudyDirection
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
 from optuna.visualization._plotly_imports import _imports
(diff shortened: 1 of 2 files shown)


not522:

optuna/visualization/_contour.py
```diff
@@ -323,7 +325,17 @@ def _get_contour_subplot_info(
             value = target(trial)
         assert value is not None
 
-        z_values[(x_i, y_i)] = value
+        existing = z_values.get((x_i, y_i))
+        if existing is None or target is not None:
+            # In multi-objective case we don't know which objective was selected in target
+            # function therefore we don't know how to select the best objective value.
```
not522, line 331: Since `target` is also used in single-objective optimization, could you please modify this comment? For example, other metrics or elapsed time may be used as `target`.

not522 (commented):
Thank you for you PR! It's almost LGTM. Please check my comment.

optuna/visualization/_contour.py
xadrianzetx: Good point, updated in 4d7f4077e99faeb2234868e450f8b1e1793c510f.

xadrianzetx:
I see the sphinx issue already fixed on main, so I'll leave it at that since this check is not required.

not522 (approved):
Thank you! LGTM!

Alnusjaponica (approved):
LGTM! 
@not522 Could you merge this PR?

Issue #4832, opened 2023-07-25
Support constraint in `plot_rank`


hrntsm:
### Motivation

Make  `plot_rank` plots support optimization with constraints

### Description

The values of the constraints are taken from _CONSTRAINTS_KEY in trial.system_attrs and applied to the results of the plot.
_CONSTRAINTS_KEY is located below, import and use it.
https://github.com/optuna/optuna/blob/master/optuna/samplers/_base.py#L202

Infeasible trials to be displayed in `#CCCCCC` color.

ryota717:
i'd like to do this. but found a bug in test cases and code.
firstly i'll fix it.

HideakiImamura:
@ryota717 is working on this issue, so let me remove the contribution-welcome.

not522:
This feature has been implemented by #4899. Thank you for your contribution!

PR #4874, opened 2023-08-19
Fix numpy indexing bugs and named tuple comparing.

ryota717:
## Motivation

In association with #4832, i found a bug in 'visualization/_rank'.
So i firstly fix this in this PR. 

## Description of the changes

- `_named_tuple_equal` https://github.com/optuna/optuna/blob/247093640034f7313b4ee3dce18903105bc2a7fd/tests/visualization_tests/test_rank.py#L100-L109
  - python `zip` of different length arrays adjusts to array of shorter length. e.g. `for a,b in zip([], [1]): do_something()` do nothing
- after fixed above, i found another problem in https://github.com/optuna/optuna/blob/247093640034f7313b4ee3dce18903105bc2a7fd/optuna/visualization/_rank.py#L198-L204
  - `np.array([])` doesn't work in this code, so we should use python array.
```
Python 3.7.10 (default, Apr 16 2023, 07:20:07)
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> a = np.array([1,2,3])
>>> empty_arr = []
>>> empty_ndarr = np.array([])
>>> a[empty_arr]
array([], dtype=int64)
>>> a[empty_ndarr]
IndexError: arrays used as indices must be of integer (or boolean) type
```


HideakiImamura:
@contramundum53 @Alnusjaponica Could you review this PR?

Alnusjaponica (approved):
Looks good to me! We appreciate your contribution.

One thing to note is that the test input has been changed in this PR instead of the expected output. The alternative would be keeping original trial setting and changing L316-L338 as follows:
```
assert _named_tuple_equal(
        info,
        _RankPlotInfo(
            params=params,
            sub_plot_infos=[
                [
                    _RankSubplotInfo(
                        xaxis=axis_infos[params[0]],
                        yaxis=axis_infos[params[1]],
                        xs=[study.get_trials()[0].params[params[0]]],
                        ys=[study.get_trials()[0].params[params[1]]],
                        trials=[study.get_trials()[0]],
                        zs=np.array([0.0]),
                        color_idxs=np.array([0.0]),
                    )
                ]
            ],
            target_name="Objective Value",
            zs=np.array([0.0, 2.0]),
            color_idxs=np.array([0.0, 1.0]),
            has_custom_target=False,
        ),
    )
```
@contramundum53 Which one do you think is better?

contramundum53:
@ryota717 Thank you for your PR!
I like @Alnusjaponica 's test because it covers more code paths. Could you change your test as @Alnusjaponica 's?

ryota717:
@Alnusjaponica @contramundum53 Thank you for your reviews! I added the test case you suggested in [d6738aa](https://github.com/optuna/optuna/pull/4874/commits/d6738aa41c058be52048340437be1831a2a44450).

Alnusjaponica:
@contramundum53 
LGTM. Could you merge this PR?

contramundum53 (approved):
LGTM! Thanks for the PR!

PR #4877, opened 2023-08-21
CHG constrained param displayed in #cccccc

louis-she:
## Motivation

https://github.com/optuna/optuna/issues/4834 

## Description of the changes

Hey guys, it's the first time I open the PR, I'm new to optuna and want to learn by doing.

The code is current for demonstration, and the abstraction needs to be discussed. Adding a new state like `constrained` to `TrialState` could be more comfortable to implement but it could have much more side effects that I do not know yet. So instead I doing some hacks only in the visualization codes.

SCR-20230821-nhwx

### DIFF
--- a/optuna/visualization/_timeline.py
+++ b/optuna/visualization/_timeline.py
@@ -85,7 +85,11 @@ def _get_timeline_info(study: Study) -> _TimelineInfo:
     for t in study.get_trials(deepcopy=False):
         date_complete = t.datetime_complete or datetime.datetime.now()
         date_start = t.datetime_start or date_complete
-        infeasible = any([x > 0 for x in t.system_attrs[_CONSTRAINTS_KEY]])
+        infeasible = (
+            False
+            if _CONSTRAINTS_KEY not in t.system_attrs
+            else any([x > 0 for x in t.system_attrs[_CONSTRAINTS_KEY]])
+        )
         if date_complete < date_start:
             _logger.warning(
                 (
--- a/optuna/visualization/_timeline.py
+++ b/optuna/visualization/_timeline.py
@@ -116,7 +116,7 @@ def _get_timeline_info(study: Study) -> _TimelineInfo:
 
 def _get_timeline_plot(info: _TimelineInfo) -> "go.Figure":
     _cm = {
-        "COMPLETE": "#cccccc",
+        "COMPLETE": "blue",
         "FAIL": "red",
         "PRUNED": "orange",
         "RUNNING": "green",
@@ -125,7 +125,7 @@ def _get_timeline_plot(info: _TimelineInfo) -> "go.Figure":
 
     fig = go.Figure()
     for s, infeasible in product(sorted(TrialState, key=lambda x: x.name), [False, True]):
-        color = "#cccccc" if infeasible else _cm[s.name]
+        color = "#cccccc" if infeasible and s.name == "COMPLETE" else _cm[s.name]
         bars = [b for b in info.bars if b.state == s and b.infeasible == infeasible]
         if len(bars) == 0:
             continue
--- a/optuna/visualization/_timeline.py
+++ b/optuna/visualization/_timeline.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import datetime
-from itertools import product
 from typing import NamedTuple
 
 from optuna._experimental import experimental_func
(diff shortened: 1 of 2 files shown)


HideakiImamura:
@louis-she Thanks for the PR. We discussed this among the committers and reached the conclusion that graying out should be applied to trials that are both infeasible and completed, as it might be clearer. Could you make this change accordingly? Additionally, it seems that the continuous integration (CI) is failing. Once the CI passes, the reviewer will be assigned, so could you make the necessary adjustments appropriately?

louis-she:
image

toshihikoyanase:
@eukaryo Could you review this PR, please?

optuna/visualization/_timeline.py
```diff
@@ -107,28 +116,29 @@ def _get_timeline_info(study: Study) -> _TimelineInfo:
 
 def _get_timeline_plot(info: _TimelineInfo) -> "go.Figure":
     _cm = {
-        "COMPLETE": "blue",
+        "COMPLETE": "#cccccc",
```
toshihikoyanase, line 119: I think feasible trials should be blue while infeasible trials will be `#cccccc`.

>  We discussed this among the committers and reached the conclusion that graying out should be applied to trials that are both infeasible and completed, as it might be clearer.

optuna/visualization/_timeline.py
```diff
@@ -107,28 +116,29 @@ def _get_timeline_info(study: Study) -> _TimelineInfo:
...
         "PRUNED": "orange",
         "RUNNING": "green",
         "WAITING": "gray",
     }
 
     fig = go.Figure()
-    for s in sorted(TrialState, key=lambda x: x.name):
-        bars = [b for b in info.bars if b.state == s]
+    for s, infeasible in product(sorted(TrialState, key=lambda x: x.name), [False, True]):
+        color = "#cccccc" if infeasible else _cm[s.name]
```
toshihikoyanase, line 128: @mamu suggested that only infeasible and complete trials will be gray, but the current implementation changes the color of all infeasible trials to gray, regardless of their state.

toshihikoyanase (commented):
Sorry for the delayed response. I have some comments on the rule to change the color.

eukaryo:
Thank you for your PR! Apologies for the delayed response.
I agree the previous comments. Only trials that are Infeasible and have been computed (=completed) should be painted `#cccccc`.
Trials which are completed and feasible should be painted `blue`.
Trials which are running and infeasible should be painted `green`, because we want to visualize anomaly trials that stay in the RUNNING state for long time.

louis-she:
I used to think that infeasible trials would not actually be run so the state of it is undefined, looks like i have some misunderstanding here.

eukaryo:
All infeasible trials should be painted `#cccccc`, but the current code does not. For example, pruned and infeasible trials are painted `orange`. And, the legend on the right should be combined into one.
```
import time

import optuna


def objective(trial):
    time.sleep(0.1)

    c0 = 3 - (x**2 + y**2)
    trial.set_user_attr("constraint", (c0,))

    if int(x * 10000) % 2 == 0:
        raise optuna.TrialPruned()

    v0 = x**2 + y**2
    return v0


def constraints(trial):


sampler = optuna.samplers.TPESampler(constraints_func=constraints, seed=12345)
study = optuna.create_study(
    direction="minimize",
    sampler=sampler,
)
optuna.visualization.plot_timeline(study).write_image("test.png")

```
test

louis-she:
fixed
image

eukaryo:
Almost LGTM! 
Could you please modify the test code, [test_timeline.py](https://github.com/optuna/optuna/blob/master/tests/visualization_tests/test_timeline.py) ? (to add assertions about constraints)

eukaryo (approved):
LGTM!

eukaryo:
@toshihikoyanase Could you take another look?

toshihikoyanase:
Thank you for letting me know. I'll check it.

tests/visualization_tests/test_timeline.py
```diff
@@ -58,9 +66,16 @@ def test_get_timeline_info_empty() -> None:
...
 
-def test_get_timeline_info() -> None:
+@pytest.mark.parametrize(
+    "sys_attrs",
+    [
+        None,
+        {_CONSTRAINTS_KEY: [1.0]},
+    ],
+)
+def test_get_timeline_info(sys_attrs: Optional[Dict[str, Any]]) -> None:
```
toshihikoyanase, line 76: I noticed that the current test case does not cover the scenario where all constraint violations are negative. Could you verify it, for example, as follows?

```suggestion
@pytest.mark.parametrize(
    "trial_sys_attrs, infeasible",
    [
        (None, False),
        ({_CONSTRAINTS_KEY: [1.0]}, True),
        ({_CONSTRAINTS_KEY: [-1.0]}, False),
    ],
)
def test_get_timeline_info(trial_sys_attrs: Optional[Dict[str, Any]], infeasible: bool) -> None:
```

In addition, `sys_attrs` should be `trial_sys_attrs` to align with the argument of `_create_study`.

toshihikoyanase (commented):
Thank you for your update. The changes in the visualization module look solid to me.
I found a missing test scenario, so could you update the test, please?

tests/visualization_tests/test_timeline.py
```diff
@@ -2,19 +2,26 @@
...
 from optuna.trial import TrialState
 from optuna.visualization._plotly_imports import go
 from optuna.visualization._timeline import _get_timeline_info
 from optuna.visualization._timeline import plot_timeline
 
 
-def _create_study(trial_states_list: list[TrialState]) -> Study:
+def _create_study(
+    trial_states_list: list[TrialState],
+    trial_sys_attrs: Optional[Dict[str, Any]] = None,
```
toshihikoyanase, line 23: A nit, but we use `dict` instead of `typing.Dict` as can be seen in #4508.
```suggestion
    trial_sys_attrs: dict[str, Any] | None = None,
```

tests/visualization_tests/test_timeline.py
```diff
@@ -2,19 +2,26 @@
 
 import datetime
 from io import BytesIO
+from typing import Any
+from typing import Dict
+from typing import Optional
```
toshihikoyanase, line 7:
```suggestion
```

toshihikoyanase (commented):
Thank you for your update. The change almost looks good to me. I have minor comments on the type annotations.

toshihikoyanase:
I confirmed that the infeasible trials are shown as expected.

```python
import optuna
from optuna.samplers._base import _CONSTRAINTS_KEY


    if x < 0.1:
        raise ValueError()
    if x > 0.9:
        raise optuna.exceptions.TrialPruned

    u = x - 0.3
    trial.set_system_attr(_CONSTRAINTS_KEY, [u])

    return x


study = optuna.create_study()
study.optimize(objective, n_trials=15, catch=(ValueError,))

optuna.visualization.plot_timeline(study).show()
```

image

tests/visualization_tests/test_timeline.py
louis-she: Fixed

toshihikoyanase (approved):
Looks good to me. Your contribution is highly appreciated. It's great to see this being your first contribution to the project.

toshihikoyanase:
Possible follow-up will be to add the matplotlib implementation

louis-she:
@toshihikoyanase thanks, and I'm willing to contribute more : )

PR #4882, opened 2023-08-24
Support `asv` 0.6.0

not522:
## Motivation
Speed benchmark CI has been failed because asv 0.6.0 was released. Change log is here. https://github.com/airspeed-velocity/asv/blob/master/CHANGES.rst#060-2023-08-20

## Description of the changes
- Remove `--strict` option
- Add `pip+` prefix (I'm not sure this is the intended behavior, but it seems necessary in asv 0.6.0.)

I have confirmed the change on my fork repository. https://github.com/not522/optuna/actions/runs/5961145531/job/16169802756

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -76,8 +76,8 @@
     // followed by the pip installed packages).
     //
     "matrix": {
-        "fakeredis": [""],
-        "pytest": [""],
+        "pip+fakeredis": [""],
+        "pip+pytest": [""],
     },
 
     // Combinations of libraries/python versions can be excluded/included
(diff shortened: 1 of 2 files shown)


asv.conf.json
```diff
@@ -76,8 +76,8 @@
     // followed by the pip installed packages).
     //
     "matrix": {
-        "fakeredis": [""],
-        "pytest": [""],
+        "pip+fakeredis": [""],
+        "pip+pytest": [""],
```
toshihikoyanase, line 80: Notes: I've confirmed that adding `pip+` is required in this configuration. Without it, the following error occurs:
>    ModuleNotFoundError: No module named 'fakeredis'

toshihikoyanase (approved):

Let me merge this PR since it does not update code base but only fixes a CI job for maintainers.

PR #4885, opened 2023-08-24
update parent-child relations to cascade deletes

cody-mar10:
- see this answer for the details: https://stackoverflow.com/questions/5033547/sqlalchemy-cascade-delete/38770040?r=Saves_UserSavesList#38770040
  -  these updates force the database backend to cascade deletes properly instead of sqlalchemy
  - without these changes, foreign key constraint errors are raised, presumably due to trying to null nonnullable fields

## Motivation

I've had the need to merge independently run studies using a sqlite storage. During merging the databases, I opted to remove any failed trials, but the current cascade settings raise `sqlite3.IntegrityError`s once certain trials are removed from the `trials` table.

## Description of the changes

1. All instances of `ForeignKey(...)` were changed to `ForeignKey(..., ondelete="CASCADE")`.
    - `ondelete="CASCADE"` adds this constraint to the table schema, allowing the db engine to handle cascades instead of SQLalchemy
2. All instances of `orm.relationship(..., backref=orm.backref(...))` were changed to `orm.relationship(..., backref=orm.backref(..., passive_deletes=True))`
    - `passive_deletes` allows the db engine to remove records where a foreign key is `NULL` before an error occurs
3. `black` also made formatting changes

### DIFF
--- a/optuna/storages/_rdb/models.py
+++ b/optuna/storages/_rdb/models.py
@@ -1,30 +1,28 @@
 import enum
 import math
-from typing import Any
-from typing import List
-from typing import Optional
-from typing import Tuple
+from typing import Any, List, Optional, Tuple
 
-from sqlalchemy import asc
-from sqlalchemy import case
-from sqlalchemy import CheckConstraint
-from sqlalchemy import Column
-from sqlalchemy import DateTime
-from sqlalchemy import desc
-from sqlalchemy import Enum
-from sqlalchemy import Float
-from sqlalchemy import ForeignKey
-from sqlalchemy import func
-from sqlalchemy import Integer
-from sqlalchemy import orm
-from sqlalchemy import String
-from sqlalchemy import Text
-from sqlalchemy import UniqueConstraint
+from sqlalchemy import (
+    CheckConstraint,
+    Column,
+    DateTime,
+    Enum,
+    Float,
+    ForeignKey,
+    Integer,
+    String,
+    Text,
+    UniqueConstraint,
+    asc,
+    case,
+    desc,
+    func,
+    orm,
+)
 
 from optuna import distributions
 from optuna.study._study_direction import StudyDirection
 from optuna.trial import TrialState
-
 
 try:
     from sqlalchemy.orm import declarative_base
@@ -87,11 +85,12 @@ class StudyDirectionModel(BaseModel):
     __table_args__: Any = (UniqueConstraint("study_id", "objective"),)
     study_direction_id = Column(Integer, primary_key=True)
     direction = Column(Enum(StudyDirection), nullable=False)
-    study_id = Column(Integer, ForeignKey("studies.study_id"), nullable=False)
+    study_id = Column(Integer, ForeignKey("studies.study_id", ondelete="CASCADE"), nullable=False)
     objective = Column(Integer, nullable=False)
 
     study = orm.relationship(
-        StudyModel, backref=orm.backref("directions", cascade="all, delete-orphan")
+        StudyModel,
+        backref=orm.backref("directions", cascade="all, delete-orphan", passive_deletes=True),
     )
 
     @classmethod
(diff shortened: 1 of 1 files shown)


toshihikoyanase:
Thank you so much for your contribution! We really appreciate the time and effort you've put into this PR.

I guess your changes were motivated to directly modify the database and delete individual trials. It's worth mentioning that the current design of Optuna expects the deletion of studies as a whole, rather than individual trials. Direct modifications to the database also pose a risk of data integrity issues, which is why it's not supported.

Additionally, @c-bata mentioned SQLite3 does not fully support CASCADE on the schema level, according to SQLAlchemy documentation.

> SQLite supports FOREIGN KEY syntax when emitting CREATE statements for tables, however by default these constraints have no effect on the operation of the table.
> https://docs.sqlalchemy.org/en/13/dialects/sqlite.html#foreign-key-support 

Given these reasons, it's challenging for us to merge your current PR.

However, if your original goal is to merge studies, Optuna's public APIs like `optuna.create_study` and [`Study.add_trials`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html#optuna.study.Study.add_trials) are the recommended way to go about it.


```python
storage = ...

# study1 and study2 are existing studies.
study1 = optuna.load_study(storage=storage, ...)
study2 = optuna.load_study(storage=storage, ...)

# create a new study to merge study1 and study2.
merged_study = optuna.create_study(storage=storage, ...)

# add trials in study1 and study2
merged_study.add_trials(study1.trials)
merged_study.add_trials(study2.trials)

# delete study1 and study2 if needed
optuna.delete_study(study_name=study1.study_name, storage=storage)
optuna.delete_study(study_name=study2.study_name, storage=storage)
```

cody-mar10:
Thank you for your response! The `add_trials` API basically handles my use case, so greatly appreciated for simplifying my workflow.

The only issue I encountered was merging in the same trials over and over again. I am working in a high throughput cluster compute environment, so I am trying to leverage parallelism over a number of independent trials that effectively "share" the same study session, even if these are actually broken among multiple storages. I know a single remote RDB like MySQL/PostgreSQL is probably better for this, but there are hurdles to make this work for the cluster I work in, so I figured I'd try merging independent studies as a simpler workaround.

My workflow starts with merging independent runs, but as this continues to happen, trials that occurred at the beginning of this get repeatedly merged. My simple workaround was to just store unique trial fields (ie datetime start/complete) and not add the trial if it has already been added. Something like:
```python
seen = set()
for study in studies:
    for trial in study.trials:
        start_stop = (trial.datetime_start, trial.datetime_complete)
        if start_stop not in seen:
            merged_study.add_trial(trial)
        seen.add(start_stop)
```
But is there any optuna API for this that I should use instead?

toshihikoyanase:
Thank you for elaborating on your use case. Your workaround does seem reasonable.

> I know a single remote RDB like MySQL/PostgreSQL is probably better for this, but there are hurdles to make this work for the cluster I work in

I understand that managing SQL servers in your cluster could be challenging. Have you considered using `JournalFileStorage`? Introduced in Optuna v3.1.0, it allows for parallel optimization using shared file systems like NFS.
c.f., https://medium.com/optuna/distributed-optimization-via-nfs-using-optunas-new-operation-based-logging-storage-9815f9c3f932

Please note that `JournalStorage` is still an experimental feature, so it may have breaking changes in future versions of Optuna.

cody-mar10:
> I understand that managing SQL servers in your cluster could be challenging. Have you considered using `JournalFileStorage`? Introduced in Optuna v3.1.0, it allows for parallel optimization using shared file systems like NFS. c.f., https://medium.com/optuna/distributed-optimization-via-nfs-using-optunas-new-operation-based-logging-storage-9815f9c3f932
 
 I haven't tried this yet since I am using `PyTorch Lightning`, and it look likes the [optuna integration](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.PyTorchLightningPruningCallback.html) for PyTorch lightning requires `RDBStorage` currently. 

Anyway thanks for your help!

toshihikoyanase:
> it look likes the [optuna integration](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.PyTorchLightningPruningCallback.html) for PyTorch lightning requires `RDBStorage` currently.

Thank you for pointing it out. @c-bata has suggested that the documentation may not be up-to-date, and the current PyTorch lightning integration should support `JournalStorage`.

Please note that some environments, such as Windows, may require additional options (e.g., `lock_obj`). You can find the latest information on `JournalStorage` at
https://optuna.readthedocs.io/en/latest/reference/generated/optuna.storages.JournalStorage.html.

We'd appreciate your feedback if you encounter any issues with it.

PR #4892, opened 2023-08-29
DRAFT: Add carbs

dan-jacobson:
## Motivation
Recently, [Generally Intelligent](https://generallyintelligent.com/) released a paper called [Tune As You Scale: hyperparameter optimization for compute efficient training](https://arxiv.org/pdf/2306.08055.pdf), introducing an algorithm they call Cost-Aware Pareto Region Bayesian Search, or CARBS.

This is a quite interesting approach to a specific multi-objective optimization problem: that of maximizing performance while minimizing some notion of "cost", usually FLOPS or some other measure of compute.

Basically, they calculate a pareto front from trials, then model both performance and cost via GP surrogate models, which they then use to sample new hyperparameter combinations along the existing cost-performance pareto front.

Seems cool, fun idea -- I hadn't seen it implemented anywhere so I gave it a shot.

Anyways, here's my rough sketch of what the Generally Intelligent team calls their "simplified" algorithm. There's still a bunch of things missing, this is not finished. Also, this is my first time getting into the weeds of the Optuna API, so I know I'm certainly not using it correctly.

However, I wanted to toss this out there and get some feedback -- even just to see if other folks were interested or thought it would be worth working on more. This is not yet tested, and it doesn't really work yet.

## Description of the changes
I've added a `CARBSSampler` class to `multi_objective/samplers` which implements my understanding of the CARBS algorithm in its `sample_relative()` method.

I've added an (as of now empty) test file, and i've adapted the `002_multi_objective` tutorial to use CARBS as an integration test.

A couple of things I'd be interested in feedback on:

1. This requires adding a dependency on `sklearn.gaussian_process` because I use the scikit-learn implementation of GaussianProcessRegressor. Is the Optuna team fundamentally against adding additional dependencies? Seems reasonable that they might, but this would be a big blocker because I'm not going to roll my own GP implementation most likely.
2. Unless I'm mistaken (and I very much might be) Optuna's `distributions` play poorly with CARBS: the algorithm assumes users want to define (for each hyperparam) a `search_center` and `sigma` and then model gaussians around them restricted by some `min` and `max`. If I squint I can kiiinda see how to twist Optuna distributions to match that but it makes me nervous.
3. CARBS is not _truly_ multi-objective. It assumes "cost" and "performance" are the outcomes of interest. I think it's plausible to try and generalize to any two-objective study, but I haven't thought all the way through it and there may be some edge cases. Is it still useful if it's not general?
4. The GP's need rectangular inputs so right now we can't model situations where trials contain variable numbers of hyperparameters. This seems like it cuts against some of Optuna's design philosophy. Probably we could just pad-out to match the length of the trial with the most hyperparameters and make it work anyways, but I haven't confirmed that yet.
5. I don't really understand the `seach_space` API so I'm almost certainly using it wrong.

Anyways, hopefully this is interesting to someone. I'm open to any and all feedback, comments, suggestions, and criticism. Unless one of the core devs says this is existentially doomed, I'll probably keep working on it. Thanks for reading!

Also, I'm not sure exactly who to tag, so I'm just tagging some folks that have had stuff merged recently -- my apologies if that's incorrect.
@c-bata @HideakiImamura @Alnusjaponica


wbrenton:
Would love to see this get some attention. CARBS would be a worthwhile addition to Optuna. @dan-jacobson are you continuing to work on this?

contramundum53:
@dan-jacobson 
Sorry for my late response and thank you very much for writing this draft PR!
I had been somewhat busy and I finally got some time to read the paper. Multi-fidelity (cost-aware) optimization is indeed an important feature that is missing out in Optuna (except for `Pruner`s, that is), and CARBS seems one interesting practical approach to it.   

However, I'm afraid that maintaining this algorithm as one of the samplers that optuna ships _by default_ is somewhat "too heavy", considering the following points:
* As you mentioned, for the algorithm to work properly, users need to specify `sigma` appropriately for each parameter and, more importantly, a lot more hyper-hyperparameters if users want to use the sophisticated version of the CARBS algorithm for best performance (instead of the simplified one you implemented). Reading the paper, I fear that this amount of hyper-hyperparameters might be beyond the capability of optuna to encapsulate, and instead users might be better off implementing their own version of CARBS to have full control over all those optimization techniques.
* Samplers shipped in optuna value generality very much, and as you mentioned, borrowing the interface of multi-objective optimization to do multi-fidelity optimization would not be appropriate. This is not to say that the problem setting is too nitch - I personally would love to see a multi-fidelity optimization algorithm coming in optuna - but for now, there isn't a nice interface for multi-fidelity optimization in optuna yet. Of course we are willing to invent a new interface if there is a satisfactory multi-fidelity optimization algorithm.

(These two are the main points; Other points you mentioned could also be problematic but are not as hard to overcome:
1. Indeed we need to be very careful when adding new dependency, and `GPy` seems unmaintained for a while so we would not want to use it. Fortunately, for gaussian process, we use `gpytorch` and `botorch` (as in `BoTorchSampler`), so it wouldn't be a problem if we could reimplement CARBS with these libraries. 
4. Needing rectangular inputs are admissible, as in `CmaEsSampler` and `QMCSampler`. Needing all parameters to be continuous can be more problematic in practice, but `BoTorchSampler` has similar problem and we are constantly looking for solutions.)

So here are my suggestions - how about creating a new repository and put your optuna sampler there? I think it will attract many people's interest, and it will be as useful if you can make it an optuna sampler. We can also add your sampler as an example user-defined sampler in [optuna-examples](https://github.com/optuna/optuna-examples) and/or introduce your sampler in tutorial/blog, etc.

Again, thanks for all your work and nice try!

contramundum53:
A short story aside: we have actually tried multi-fidelity optimization before, based on "knowledge gradient" approach. It was theoretically nice, but the BoTorch implementation we tried was, as we concluded, not mature enough: evaluation of that acquisition function would need an optimization process and the optimization would get trapped in local optima in significantly higher probability than more classical acquisition functions (such as EI).

PR #4896, opened 2023-09-01
Use `isinstance` instead of `if type() is ...`

Alnusjaponica:
## Motivation
Replace `if type() is ...` with `if isinstance` as recommended in [PEP 8](https://peps.python.org/pep-0008/#programming-recommendations).

## Description of the changes
- Replace `if type() is ...` with `if isinstance` sentence.

### DIFF
--- a/optuna/integration/_lightgbm_tuner/optimize.py
+++ b/optuna/integration/_lightgbm_tuner/optimize.py
@@ -99,7 +99,7 @@ class _BaseTuner:
         if self.lgbm_kwargs.get("valid_names") is not None:
             if isinstance(self.lgbm_kwargs["valid_names"], str):
                 valid_name = self.lgbm_kwargs["valid_names"]
-            elif isinstance(self.lgbm_kwargs["valid_names"], Iterable):
+            elif isinstance(self.lgbm_kwargs["valid_names"], Sequence):
                 valid_name = self.lgbm_kwargs["valid_names"][-1]
             else:
                 raise NotImplementedError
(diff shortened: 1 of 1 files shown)


Alnusjaponica:
I am going to resolve `codecov` warning after (or when) https://github.com/optuna/optuna/issues/4959 is resolved.

optuna/integration/_lightgbm_tuner/optimize.py
```diff
@@ -116,27 +118,27 @@ def _get_booster_best_score(self, booster: "lgb.Booster") -> float:
...
-            eval_at = [1, 2, 3, 4, 5]
+            or [1, 2, 3, 4, 5]
+        )
 
         # Optuna can handle only a single metric. Choose first one.
-        if type(eval_at) in [list, tuple]:
-            return "{}@{}".format(metric, eval_at[0])
-        if type(eval_at) is int:
-            return "{}@{}".format(metric, eval_at)
+        if isinstance(eval_at, (list, tuple)):
```
not522, line 135:
```suggestion
        if isinstance(eval_at, Sequence):
```

not522 (commented):
Thank you, it's almost LGTM. Could you check my suggested change?

optuna/integration/_lightgbm_tuner/optimize.py
not522: We discussed it offline and understood that it should not include string.

not522 (approved):
LGTM!

Issue #4868, opened 2023-08-10
Imporve colormap of optuna.visualization.plot_parallel_coordinate()


LiShaoyu5:
### Motivation

When I use optuna for multi-objective optimization and draw plots with optuna.visualization.plot_parallel_coordinate(), I found that the larger the target value, the lighter the color of the corresponding line, which is oppsite with single-objective optimization. Then I looked for [documentation](https://optuna.readthedocs.io/en/stable/reference/visualization/generated/optuna.visualization.plot_parallel_coordinate.html#optuna.visualization.plot_parallel_coordinate) and found this:

> The colormap is reversed when the target argument isn’t [None](https://docs.python.org/3/library/constants.html#None) or direction of [Study](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html#optuna.study.Study) is minimize.

It's confusing that the colormap rules vary by single and multiple objectives and are independent of the optimization direction.

Besides, current colormap makes some of the lines difficult to recognize even in the example image of documentation because the background color is also white, especially when numbers of trials become large:
![image](https://github.com/optuna/optuna/assets/37066594/952ac33d-0511-4dea-b827-cfe47beba2ad)

### Description

Suggest to use colormap similiar to the function plot_rank(), which uses red and blue to differentiate data instead of white and blue:
![image](https://github.com/optuna/optuna/assets/37066594/89348aa8-8a4c-4e4f-8253-abf6e72aefff)

contramundum53:
Thank you for your suggestion!

You can change your plot to your favorite color scale by the following code:
```
fig = optuna.visualization.plot_parallel_coordinate(study)
fig.data[0].line["colorscale"] = "RdYlBu" # Reverse color scale by "RdYlBu_r"
fig.show()
```

If more people want this, we can consider making it the default color scale. 
Let's have a vote here, if you think the color scale of `plot_rank` is better than the current `plot_parallel_coordinate`, please react by ❤️ . If you feel the opposite, please react by ☹️ .

contramundum53:
The story for the curent behavior of the color is a bit long:

When we only had single-objective optimization, we first wanted to show good lines in dark colors and bad ones in light colors. It is by design that some lines (bad ones) can be difficult to see, since most of the times we would like to know what kind of parameters are good. 

[...]

PR #4899, opened 2023-09-04
Support constraints plot_rank

ryota717:
## Motivation

#4832 

Supports `plot_rank` visualization for Infeasible trials to be displayed in `#CCCCCC` color.

## Description of the changes

- [f10a978](https://github.com/optuna/optuna/pull/4899/commits/f10a97875735d99a9d905497b701888608fc6c54): refactor to pass each color as rgb(numpy array with 0 ~ 1).
- [07092ba](https://github.com/optuna/optuna/pull/4899/commits/07092ba7c24619081c426e67e4310d5173d99c74): infeasible trials visualization support.

### DIFF
--- a/optuna/visualization/_rank.py
+++ b/optuna/visualization/_rank.py
@@ -294,7 +294,7 @@ def _get_rank_subplot(
         x=info.xs,
         y=info.ys,
         marker={
-            "color": info.colors,
+            "color": list(map(plotly.colors.label_rgb, info.colors)),
             "cmin": 0,
             "cmax": 1,
             "line": {"width": 0.5, "color": "Grey"},
--- a/optuna/visualization/_rank.py
+++ b/optuna/visualization/_rank.py
@@ -82,10 +82,18 @@ def plot_rank(
             def objective(trial):
                 x = trial.suggest_float("x", -100, 100)
                 y = trial.suggest_categorical("y", [-1, 0, 1])
+
+                c0 = 400 - (x + y)**2
+                trial.set_user_attr("constraint", [c0])
+
                 return x ** 2 + y
 
 
-            sampler = optuna.samplers.TPESampler(seed=10)
+            def constraints(trial):
+                return trial.user_attrs["constraint"]
+
+
+            sampler = optuna.samplers.TPESampler(seed=10, constraints_func=constraints)
             study = optuna.create_study(sampler=sampler)
             study.optimize(objective, n_trials=30)
 
--- a/optuna/visualization/_rank.py
+++ b/optuna/visualization/_rank.py
@@ -204,9 +204,7 @@ def _get_rank_subplot_info(
         if constraints is not None and any([x > 0.0 for x in constraints]):
             infeasible_trial_ids.append(i)
 
-    colors[infeasible_trial_ids] = plotly.colors.unconvert_from_RGB_255(
-        (plotly.colors.hex_to_rgb("#cccccc"))
-    )
+    colors[infeasible_trial_ids] = plotly.colors.hex_to_rgb("#cccccc")
 
     filtered_ids = [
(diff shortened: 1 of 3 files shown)


not522:
@contramundum53 @Alnusjaponica Could you review this PR?

Alnusjaponica (changes requested):
We really appreciate your continuous contribution. 
I made some comments, mainly on how to handle color values. PTAL.

Alnusjaponica:
This is totally optional but could you use the newly implemented feature in the docstring example? 
You can just leave this as follow-up too.

ryota717 (commented):
@Alnusjaponica Thanks for your review! Reflect each of your comments. Also i add constraints to `plot_rank` docstring example on  [60ad37d](https://github.com/optuna/optuna/pull/4899/commits/60ad37d8d45c349dc1f7f01d8e2efa0262b94ecd).
Would you check them?

optuna/visualization/_rank.py
```diff
@@ -403,3 +414,17 @@ def _get_rank_plot(
...
+    # sample_colorscale requires plotly >= 5.0.0.
+    labeled_colors = plotly.colors.sample_colorscale(colormap, color_idxs)
+    scaled_rgb_colors = np.array(
+        [
+            plotly.colors.unconvert_from_RGB_255(plotly.colors.unlabel_rgb(cl))
+            for cl in labeled_colors
+        ]
+    )
+
+    return scaled_rgb_colors.reshape(-1, 3)
```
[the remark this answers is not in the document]
contramundum53: In that case, it would be difficult to customize the color scale later. (cf. https://github.com/optuna/optuna/issues/4868#issuecomment-1679984127)

Alnusjaponica (changes requested):
@ryota717 
I apologize for the delayed response. Thank you for updating it up to the example. I have provided a comment with a single modification, but other than that, it looks good to me. Thanks a lot!

Alnusjaponica (approved):
@ryota717 Thanks for your prompt response. LGTM.
@contramundum53 Could you take another look?

contramundum53:
Sorry for late reply. I'll take a look.

optuna/visualization/_rank.py
```diff
@@ -160,16 +169,18 @@ def target(trial: FrozenTrial) -> float:
     target_values = np.array([target(trial) for trial in trials])
     raw_ranks = _get_order_with_same_order_averaging(target_values)
     color_idxs = raw_ranks / (len(trials) - 1) if len(trials) >= 2 else np.array([0.5])
```
contramundum53, line 171: Currently, the rank does not take constraints into consideration, so when there are many many "good" points that do not satisfy constraints, points will be colored in either gray or "bad" colors.
A more intuitive alternative is to color points based on ranks of feasible points only. What do you think?

ryota717: @contramundum53 Sounds great, but there were two things that concerned me.

1. I think it's important that whether user can notice constraints are appropriately effective. I believe `plot_rank` is one of such ways. For example, below cases result same visualization if we make colors based on only feasible points.

- N infeasible points with target value smaller than $val$, M feasible points with target value larger than $val$
- N infeasible points with target value larger than $val$, M feasible points with target value smaller than $val$

2. How to treat `tick_info` is very difficult. It's so strange for me that there may be some infeasible points with out of `tick_info` range.

If aboves are no problems for you, i would change base colors(But is it really necessary to plot infeasible trials in such case?).

contramundum53: Hmm, I see there could be controversy. Let's merge this PR and see how users feel 👍

contramundum53 (approved):
LGTM!

PR #4904, opened 2023-09-08
Fix Tests (MPI) CI

not522:
## Motivation
CI fails because `torchmetrics` cannot be installed from `https://download.pytorch.org/whl/cpu`.

## Description of the changes
Install `torchmetrics` from PyPI before other integration libraries.


not522:

contramundum53 (approved):
CI Passes. LGTM.

PR #4905, opened 2023-09-08
Support Study Artifacts

c-bata:
## Motivation

Currently, artifacts (output files) can only be associated with trials. This PR also allows users to upload an artifact, which is associated with studies.

## Description of the changes

Change `upload_artifact()` function to accept `optuna.Study` as well as trial objects. There is a minor breaking change that the code `upload_artifact(trial=trial, ...)` will be broken. However we discussed it internally and reached the consensus to accept it since this feature is still experimental and has not been promoted at Optuna 3.3 release.


not522:

gen740 (approved):
LGTM!

not522 (approved):
LGTM!

PR #4910, opened 2023-09-10
Add mimetype input

hrntsm:
## Motivation

Support for displaying 3d models in https://github.com/optuna/optuna-dashboard/pull/552
I am considering increasing the number of supported 3D file types.

Added an option to specify directly the mimetype of the 3d model file, since it is not always expected to be appropriate using the guess function.

For example, ".obj" file, which is one of the extensions of a 3D model, will be `application/x-tgif` in one environment, but `model/obj` in another environment.

This is because the function mimetypes.guess_type gets mimetype information from OS-dependent files. (I suspect this is due to the use of [knownfiles](https://github.com/python/cpython/blob/63a7f7765c6e9c1c9b93b7692e828ecf7bbd3bb9/Lib/mimetypes.py#L48) .)

## Description of the changes

The implementation was made the same as the [optuna dashboard implementation](https://github.com/optuna/optuna-dashboard/blob/941f926abdcff1fa7f9377cb8ce07937161f998f/optuna_dashboard/artifact/_backend.py#L142).

### DIFF
--- a/optuna/artifacts/_upload.py
+++ b/optuna/artifacts/_upload.py
@@ -5,6 +5,7 @@ from dataclasses import dataclass
 import json
 import mimetypes
 import os
+from typing import Optional
 import uuid
 
 from optuna._experimental import experimental_func
@@ -33,6 +34,8 @@ def upload_artifact(
     artifact_store: ArtifactStore,
     *,
     storage: BaseStorage | None = None,
+    mimetype: Optional[str] = None,
+    encoding: Optional[str] = None,
 ) -> str:
     """Upload an artifact to the artifact store.
 
@@ -67,8 +70,8 @@ def upload_artifact(
     artifact = ArtifactMeta(
         artifact_id=artifact_id,
         filename=filename,
-        mimetype=guess_mimetype or DEFAULT_MIME_TYPE,
-        encoding=guess_encoding,
+        mimetype=mimetype or guess_mimetype or DEFAULT_MIME_TYPE,
+        encoding=encoding or guess_encoding,
     )
     attr_key = ARTIFACTS_ATTR_PREFIX + artifact_id
     storage.set_trial_system_attr(trial_id, attr_key, json.dumps(asdict(artifact)))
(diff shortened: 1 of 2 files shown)


c-bata:

optuna/artifacts/_upload.py
```diff
@@ -33,6 +34,8 @@ def upload_artifact(
     artifact_store: ArtifactStore,
     *,
     storage: BaseStorage | None = None,
+    mimetype: Optional[str] = None,
+    encoding: Optional[str] = None,
```
c-bata, line 38: Could you update the `Args` section in the docstring?

c-bata (commented):
@hrntsm Thank you for your pull request! Looks almost good to me. I left one comment though.

optuna/artifacts/_upload.py
```diff
@@ -48,6 +51,12 @@ def upload_artifact(
         storage:
             A storage object. If trial is not a :class:`~optuna.trial.Trial` object, this argument
             is required.
+        mimetype:
+            A MIME type of the artifact. If not specified, the MIME type is guessed from the file
+            extension.
+        encoding:
+            An encoding of the artifact. If not specified, the encoding is guessed from the file
+            extension.
```
c-bata, line 59: A nit.

```suggestion
            An encoding of the artifact, which is suitable for use as a ``Content-Encoding``
            header (e.g. gzip). If not specified, the encoding is guessed from the file extension.
```

c-bata (approved):
LGTM with a nit! Thank you for your contribution.

optuna/artifacts/_upload.py
```diff
@@ -34,6 +35,8 @@ def upload_artifact(
     artifact_store: ArtifactStore,
     *,
     storage: BaseStorage | None = None,
+    mimetype: Optional[str] = None,
+    encoding: Optional[str] = None,
```
gen740, line 39: Could you use `str | None` instead of `Optional`?

gen740 (approved):
Almost LGTM! I've left a minor comment. It seems something is wrong with the CI; could you please rerun it?

c-bata:
Currently, integration tests seems to be broken since XGBoost 2.0.0 was released about 2 hours ago. I've just created a hotfix https://github.com/optuna/optuna/pull/4919, so please wait until #4919 will be merged :bow:

c-bata:
@hrntsm Could you merge the main branch since #4921 was merged?

PR #4927, opened 2023-09-15
Mention a pruner instance is not stored in a storage in resuming tutorial

nzw0301:
## Motivation

The current tutorial only mentions a sampler instance is not stored in the storage. In my understanding, a pruner instance has the same situation. It would be nice to clarify this in the tutorial. 

A user actually confused this behaviour as mentioned in https://github.com/optuna/optuna/discussions/4883.

## Description of the changes

Add a few words to mention pruner instance is not saved in the storage.


contramundum53 (approved):
LGTM! Thanks for the PR!

Issue #4895, opened 2023-08-31
Allow user-defined objective names in hyperparameter importance plots


xadrianzetx:
### Motivation

Support for multi-objective optimization has been added to hyperparameter importance plots in #4871. By default, objective names displayed in legend follow the pattern `Objective Value {n}`. To improve readability of the plots, it would be nice to let users define their own objective names to display. 

### Description

As discussed in https://github.com/optuna/optuna/pull/4871#discussion_r1292456529, both `optuna.visualization.plot_param_importances` and `optuna.visualization.matplotlib.plot_param_importances` APIs should be modified to allow passing custom objective names. APIs should accept:
* Strings like `"Custom objective name"`, which for multi-objective optimization will be suffixed with objective number.
* Sequences of strings like `["Objective foo", "Objective bar"]`, where each objective will be assigned a name from the collection.

not522:
Hi! Thank you for your great suggestion!
Recently, we introduced the `metric_names` API for objective names. I think that it is suitable for `plot_param_importances`. (ref: #4383)

xadrianzetx:
Oh, that's very cool. No need to modify existing `plot_param_importances` APIs then.

xadrianzetx:
BTW @not522, do you think it's worth introducing a getter for metric names as well? Currently you have to [traverse quite far and know magic literals to retrieve them](https://github.com/optuna/optuna/blob/de870c0bdff01a0c188dedc688cd30f04b05a9b9/optuna/visualization/_pareto_front.py#L346-L348).

not522:
Yes, I think it would be useful to introduce the getter. The reason we were cautious about implementing it at that time was that there were several API design options. (e.g., `create_study(..., directions={"loss": MINIMIZE, "duration": MINIMIZE})` or return dict in the objective function)
In my opinion, it would be good to add an attribute like `study.metric_names` as an experimental feature.

Issue #4586, opened 2023-04-07
Drop Python 3.7 Support


nzw0301:
### Motivation

The end of life of python 3.7 is coming soon. Specifically, the date is `27 Jun 2023` according to https://endoflife.date/python. 

A few critical dependencies, such as Numpy and PyTorch, have already dropped support python 3.7, so at least the integration module requiring PyTorch would not work soon. 

### Description

I'd like to know the opinions from core-team's decision, so please replace this field to clarify the tasks to resolve this issue.

### Alternatives (optional)

- https://github.com/optuna/optuna/issues/3021
- https://github.com/optuna/optuna/issues/4484

not522:
Thank you for your notification. For now, we continue to support Python 3.7 and do not force users to upgrade their Python. When it becomes a barrier to development, we will consider stopping support.

c-bata:
We internally discussed this topic again and decided to run integration tests with Python 3.7 (refs #4784).

PR #4930, opened 2023-09-17
Add `metric_names` getter to study

xadrianzetx:
## Motivation

This PR complements `study.set_metric_names` with getter attribute. Ref https://github.com/optuna/optuna/issues/4895#issuecomment-1711283435

Initially I've implemented this as a cached property (#4928), but due to amount of workarounds needed to make it work with Python 3.7, I've ultimately decided to go with simple uncached  version instead. This should still be fine, as I expect this is read-once kind of attribute for most of the users. Adding the cache back (in nice `cached_property` form from 093a4c6e298bf991b280f8d2e6a209e8417d2e0c) can be left as a possible enhancement after Python 3.7 is dropped.  

## Description of the changes

* Added metric_names attribute to study.
* Library code that was previously querying storage for metric names is now using the new attribute.
* Added tests

### DIFF
--- a/optuna/study/_dataframe.py
+++ b/optuna/study/_dataframe.py
@@ -8,7 +8,6 @@ from typing import Tuple
 
 import optuna
 from optuna._imports import try_import
-from optuna.study.study import _SYSTEM_ATTR_METRIC_NAMES
 from optuna.trial._state import TrialState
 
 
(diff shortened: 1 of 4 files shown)


contramundum53:
@HideakiImamura @Alnusjaponica Could you review this PR?

Alnusjaponica (approved):
LGTM.
If you have a concise description of todos for cache property, I appreciate it if you could link the task with #4586 (e.g. make a short comment on #4586 or #4928).

@HideakiImamura Could you take another look?

HideakiImamura (approved):
Looks very good.

PR #4941, opened 2023-09-24
Fix type annotation

Alnusjaponica:
## Motivation
Pass `spmatrix` as a string, which is used only when type checking.

## Description of the changes
Currently, if you use `OptunaSearchCV` with default dependencies and `scikit-learn`, you will get following error:
```
NameError: name 'spmatrix' is not defined
```
This would be because `import pandas as pd` under `try_import()` context fails and `spmatrix` is left uimported. The error message will not let users know that they need to install `pandas`. With this patch, the error message is going to be as follows:
```
ModuleNotFoundError: No module named 'pandas'
```

### DIFF
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -52,7 +52,7 @@ ArrayLikeType = Union[List, np.ndarray, "pd.Series", "spmatrix"]
 OneDimArrayLikeType = Union[List[float], np.ndarray, "pd.Series"]
 TwoDimArrayLikeType = Union[List[List[float]], np.ndarray, "pd.DataFrame", "spmatrix"]
 IterableType = Union[List, "pd.DataFrame", np.ndarray, "pd.Series", "spmatrix", None]
-IndexableType = Union[Iterable, None]
+IndexableType = Optional[Iterable]
 
 _logger = logging.get_logger(__name__)
 
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -48,10 +48,10 @@ with try_import() as _imports:
 if not _imports.is_successful():
     BaseEstimator = object  # NOQA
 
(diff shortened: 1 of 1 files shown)


HideakiImamura:
@eukaryo @gen740 Could you review this PR?

optuna/integration/sklearn.py
```diff
@@ -48,10 +48,10 @@
...
     BaseEstimator = object  # NOQA
 
-ArrayLikeType = Union[List, np.ndarray, "pd.Series", spmatrix]
+ArrayLikeType = Union[List, np.ndarray, "pd.Series", "spmatrix"]
 OneDimArrayLikeType = Union[List[float], np.ndarray, "pd.Series"]
-TwoDimArrayLikeType = Union[List[List[float]], np.ndarray, "pd.DataFrame", spmatrix]
-IterableType = Union[List, "pd.DataFrame", np.ndarray, "pd.Series", spmatrix, None]
+TwoDimArrayLikeType = Union[List[List[float]], np.ndarray, "pd.DataFrame", "spmatrix"]
+IterableType = Union[List, "pd.DataFrame", np.ndarray, "pd.Series", "spmatrix", None]
 IndexableType = Union[Iterable, None]
```
gen740, line 55: This suggestion isn't directly related to this PR, but how about using `Optional`.
```suggestion
IndexableType = Optional[Iterable]
```

gen740 (approved):
I guess, these type aliases do not so much serve as helpful type annotations, rather they add unnecessary complexity to the code.

For example.
The `_num_samples` function takes `ArrayLikeType` but in l72 `TwoDimArrayLikeType` is passed, which is not compatible with `ArrayLikeType`.

For now, it might be best to proceed with the current changes and consider making corrections later on.

eukaryo (approved):
LGTM, and I agree with the comment above. Let me merge this PR.

PR #4972, opened 2023-09-30
Update study.py for sphinx

eukaryo:
## Motivation

Fix [issue#4917](https://github.com/optuna/optuna/issues/4917).

`sphinx` could not recognize `pd.DataFrame`, which is the return type of `trials_dataframe` function, and exerts an warning in `nitpicky` mode.

## Description of the changes

I used `sys.modules` in order to check whether `sphinx` was imported,. If so, let it to import`pd`.

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -5,6 +5,7 @@ from collections.abc import Iterable
 from collections.abc import Mapping
 import copy
 from numbers import Real
+import sys
 import threading
 from typing import Any
 from typing import Callable
@@ -42,7 +43,7 @@ from optuna.trial import TrialState
 
 _dataframe = _LazyImport("optuna.study._dataframe")
 
-if TYPE_CHECKING:
+if TYPE_CHECKING or "sphinx" in sys.modules:
     from optuna.study._dataframe import pd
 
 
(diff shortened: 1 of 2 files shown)


optuna/study/study.py
```diff
@@ -42,7 +43,7 @@
 
 _dataframe = _LazyImport("optuna.study._dataframe")
 
-if TYPE_CHECKING:
+if TYPE_CHECKING or "sphinx" in sys.modules:
     from optuna.study._dataframe import pd
```
toshihikoyanase, line 47: I'm not fully sure, but if users have Sphinx installed without Pandas, they might encounter an import error. Could you please look into this?

eukaryo: When I uninstalled `pandas` and then performed `made html`, I got a warning from `study.py`, but `plotly` gave import errors regardless of this PR.
However, 
```
>>> import optuna
>>> import sys
>>> 'sphinx' in sys.modules
False
>>> import sphinx
>>> 'sphinx' in sys.modules
True
```
It seems to be not a good behavior.

eukaryo: On second thought, I thought it would not be a problem. I think lazy-import will prevent import error.

toshihikoyanase: Thank you for your investigation.
BTW, this PR seems to be still work-in-progress, so let me remove the assignment for now.

Issue #4979, opened 2023-09-30
`metric_names is experimental` warning everywhere


semiexp:
### Expected behavior

We don't want to see warnings like

```
metric_names is experimental (supported from v3.4.0). The interface can change in the future.
```

but it appears too frequently (even if we don't explicitly use `metric_names`). For example, we can find a lot of this kind of warnings in [latest tutorial](https://optuna.readthedocs.io/en/latest/tutorial/20_recipes/index.html).

### Environment

- Optuna version: latest
- Python version:
- OS:
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
metric_names is experimental (supported from v3.4.0). The interface can change in the future.
```

### Steps to reproduce

We will be able to reproduce this warning by running a tutorial code https://optuna.readthedocs.io/en/latest/tutorial/20_recipes/001_rdb.html#sphx-glr-tutorial-20-recipes-001-rdb-py

semiexp:
Resolved in #4983

PR #4983, opened 2023-09-30
Remove experimental_func from metric_names property

semiexp:
## Motivation

As described in #4979, we see too many `metric_names is experimental` warning. We don't want to see this warning when we don't use `metric_names`

## Description of the changes

To reduce this kind of warnings, we just remove `@experimental_class` from the `metric_names` property. `set_metric_names` is still marked as experimental and the warning will be shown when we attempt to set `metric_names`.

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -334,7 +334,6 @@ class Study:
         return copy.deepcopy(self._storage.get_study_system_attrs(self._study_id))
 
     @property
-    @experimental_func("3.4.0")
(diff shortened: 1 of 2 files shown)


c-bata (approved):
LGTM!

semiexp:
Looks like this warning is no longer apparent

https://optuna--4983.org.readthedocs.build/en/4983/tutorial/20_recipes/006_user_defined_pruner.html

PR #4995, opened 2023-10-03
Used-future-annotation-in-optuna/study/_optimize.py-issue#4508

jot-s-bindra:
## Motivation

## Description of the changes

I have imported the `__future__.annotations` feature and have adjusted the type hints in my code by removing the quotes from type annotations. This change is necessary because, with `__future__.annotations`, quotes are no longer needed in type annotations, making the code more aligned with modern Python conventions for type hinting.


not522:
Thank you for your PR. Could you fix the CI fails?

PR #4997, opened 2023-10-03
Used  __future__.annotations in optuna/study/_optimize.py Issue#4508

jot-s-bindra:
## Motivation

## Description of the changes

I have imported the __future__.annotations feature and have adjusted the type hints in my code by removing the quotes from type annotations. This change is necessary because, with __future__.annotations, quotes are no longer needed in type annotations, making the code more aligned with modern Python conventions for type hinting.

### DIFF
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -17,6 +17,7 @@ from typing import Tuple
 from typing import Type
 from typing import Union
 import warnings
+from __future__ import annotations
 
 import optuna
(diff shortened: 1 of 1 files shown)


nabenabe0928 (commented):
Could you also do the following?

```python
Union[foo, bar] --> foo | bar
Tuple[foo] --> tuple[foo]
List[foo] --> list[foo]
Optioinal[foo] --> foo | None
```

PR #4998, opened 2023-10-04
Drop Python 3.7 on `tests-mpi`

not522:
## Motivation
Python 3.7 support for integrations is already dropped. #4586
Tests (MPI) failed in Python 3.7 and it only tests PyTorch Distributed integration. So this PR removes the support.

## Description of the changes
Drop Python 3.7 support on `tests-mpi`.

### DIFF
--- a/.github/workflows/tests-mpi.yml
+++ b/.github/workflows/tests-mpi.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


not522:
@contramundum53 Could you review this PR? Since this PR is just a small change to the CI, I think one approval is sufficient for merge.

contramundum53 (approved):
LGTM.

PR #4999, opened 2023-10-04
Make test order deterministic for `pytest-xdist`

not522:
## Motivation
`pytest-xdist` is useful for parallelizing tests and reducing execution time, but Optuna's tests result in errors because the order is not deterministic. This PR fixes this error. After merging this PR, `pytest -m "not integration" -n auto` will succeed.

## Description of the changes
Make test order deterministic.

### DIFF
--- a/tests/storages_tests/test_journal.py
+++ b/tests/storages_tests/test_journal.py
@@ -23,19 +23,14 @@ from optuna.testing.storages import StorageSupplier
 from optuna.testing.tempfile_pool import NamedTemporaryFilePool
 
 
-LOG_STORAGE = {
+LOG_STORAGE = [
     "file_with_open_lock",
     "file_with_link_lock",
     "redis_default",
     "redis_with_use_cluster",
-}
+]
 
-LOG_STORAGE_SUPPORTING_SNAPSHOT = {
-    "redis_default",
-    "redis_with_use_cluster",
-}
-
-JOURNAL_STORAGE_SUPPORTING_SNAPSHOT = {"journal_redis"}
+JOURNAL_STORAGE_SUPPORTING_SNAPSHOT = ["journal_redis"]
 
 
 class JournalLogStorageSupplier:


not522:
@HideakiImamura @Alnusjaponica Could you review this PR?

HideakiImamura:
Should we add `pytest-xdist` to `pyproject.toml` in `test`?

HideakiImamura:
Thanks for the PR. I confirmed that we can execute `pytest -m "not integration" -n auto` with this PR.

Alnusjaponica (approved):
LGTM. It might be better to add some instruction in `optuna/blob/master/CONTRIBUTING.md`

tests/storages_tests/test_journal.py
```diff
@@ -23,19 +23,14 @@
...
     "file_with_link_lock",
     "redis_default",
     "redis_with_use_cluster",
-}
+]
 
-LOG_STORAGE_SUPPORTING_SNAPSHOT = {
-    "redis_default",
-    "redis_with_use_cluster",
-}
```
not522, line 36: This variable has not been used.

tests/storages_tests/test_journal.py
```diff
@@ -23,19 +23,14 @@
...
 
 
-LOG_STORAGE = {
+LOG_STORAGE = [
     "file_with_open_lock",
     "file_with_link_lock",
     "redis_default",
     "redis_with_use_cluster",
-}
+]
```
not522, line 31: The order of `set` items is not deterministic. As a result, `pytest-xdist` recognizes the collected tests are different between workers.

not522:
> Should we add `pytest-xdist` to `pyproject.toml` in test?

`pytest-xdist` may give confusing errors, so I don't recommend adding it in CI. I think it is useful in local environments.

HideakiImamura (approved):
Thanks for the reply. I understood. LGTM.

PR #5001, opened 2023-10-04
Used __future__.annotations in optuna/study/_optimize.py Issue#4508 

jot-s-bindra:
## Motivation

## Description of the changes

I have imported the future.annotations feature and have adjusted the type hints in my code by removing the quotes from type annotations. This change is necessary because, with future.annotations, quotes are no longer needed in type annotations, making the code more aligned with modern Python conventions for type hinting.


not522:
Could you add commits to an existing PR instead of making another PR?

Issue #5000, opened 2023-10-04
Support Python 3.12


not522:
### Motivation

Python 3.12 released on 2023-10-02. https://www.python.org/downloads/release/python-3120/
This is a tracking issue to support the new Python.

### Description

- [x] Add Python 3.12 to the CI of the unit tests.
- [x] Add Python 3.12 to the CI of the unit tests for integrations.
- [ ] Add Python 3.12 to the CI of the Docker build.
- [x] Add Python 3.12 to the CI of [optuna-examples](https://github.com/optuna/optuna-examples/).
- [x] Use Python 3.12 in the speed benchmark.
- [x] Add branch protection rules.
- [x] Add badge.
- [x] Update classifiers in `pyproject.toml`.

### Additional context (optional)

See also the 3.11 tracking issue. #3964

not522:
Current status:
Optuna already supports Python 3.12, but some CIs need to be updated to use it.
Many integration modules on `optuna-integration` also work with Python 3.12, but some are not yet supported.

PR #5018, opened 2023-10-11
Support Python 3.12 on tests for core modules

not522:
## Motivation
Adding CI to support Python 3.12. #5000
Since BoTorch does not yet support Python 3.12, CI ignores it.

## Description of the changes
- Add Python 3.12 in `tests.yml`
- Add badge in `README.md`
- Update classifiers in `pyproject.toml`


HideakiImamura (commented):
Let me clarify the scope of this PR. Is this PR work on the first TODO in https://github.com/optuna/optuna/issues/5000? Is so, I think we should modify the title and undo the change in the integration tests.

.github/workflows/tests.yml
```diff
@@ -77,13 +77,27 @@ jobs:
...
       run: |
-        pytest tests -m "not integration and not slow"
+        if [ ${{ matrix.python-version }} == 3.12 ]; then
+          # TODO(not522): Remove ignores when BoTorch supports Python 3.12
+          pytest tests -m "not integration and not slow" \
+            --ignore tests/terminator_tests/ \
+            --ignore tests/visualization_tests/test_terminator_improvement.py
+        else
+          pytest tests -m "not integration and not slow"
+        fi
```
nabenabe0928, line 101: What about this?
I bundled "Tests" and "Scheduled tests", and then the maintenance cost will lower.

```yaml
    - name: Tests
      run: |
        if [ "${{ github.event_name }}" = "schedule" ] || \
           [ "${{ github.event_name }}" = "workflow_dispatch" ] ; then
          target="not integration"
        else
          target="not integration and not slow"
        fi

        if [ "${{ matrix.python-version }}" = "3.12" ] ; then
          # TODO(not522): Remove ignores when BoTorch supports Python 3.12
          ignore_botorch_option="--ignore tests/terminator_tests/ --ignore \
          tests/visualization_tests/test_terminator_improvement.py"
        else
          ignore_botorch_option=""
        fi

        pytest tests -m "$target" $ignore_botorch_option
```

not522:
> Let me clarify the scope of this PR. Is this PR work on the first TODO in https://github.com/optuna/optuna/issues/5000? Is so, I think we should modify the title and undo the change in the integration tests.

Thank you for your review. I updated the title. This PR handles first and last two TODOs.
The change for `test_botorch.py` is necessary to avoid `ImportError`.

nabenabe0928 (approved):
LGTM:)

HideakiImamura (approved):
LGTM.

Issue #4984, opened 2023-09-30
`formats.sh` does not handle `tutorial/`


semiexp:
### Expected behavior

`formats.sh` should format files under `tutorial/`

### Environment

- Optuna version:
- Python version:
- OS:
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
(no formatting occurred but CI failed)
```

### Steps to reproduce

1.
2.
3.
```python
# python code
```

### Additional context (optional)

https://github.com/optuna/optuna/pull/4980

sousu4:
I will work on this issue.

HideakiImamura:
Let me close this issue since this was fixed in #5023.

PR #5023, opened 2023-10-11
Fix the issue where formats.sh does not handle tutorial/

sousu4:
## Motivation

I made the changes to address issue #4984.

## Description of the changes

Fixed the issue where optuna/formats.sh does not handle tutorial/
The configuration for mypy was set not to inspect 'tutorial/', which is why it is excluded from this round of fixes.

### DIFF
--- a/formats.sh
+++ b/formats.sh
@@ -43,7 +43,7 @@ do
 done
 
 target="optuna tests benchmarks tutorial"
-mypy_target="optuna tests benchmarks tutorial"
+mypy_target="optuna tests benchmarks"
 res_all=0
 
 res_black=$(black $target --check --diff 2>&1)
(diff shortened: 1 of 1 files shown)


formats.sh
```diff
@@ -42,8 +42,8 @@ do
   esac
 done
 
-target="optuna tests benchmarks"
-mypy_target="optuna tests benchmarks"
+target="optuna tests benchmarks tutorial"
+mypy_target="optuna tests benchmarks tutorial"
```
HideakiImamura, line 46: I think you don't have to fix this line.

HideakiImamura (approved):
LGTM.

contramundum53 (approved):
LGTM!

PR #4898, opened 2023-09-02
Implement `get_all_study_names()`

Alnusjaponica:
## Motivation
Currently, we only have `get_study_summaries()` to retrieve all study names from storage, and it is performing too slowly for the intended use. This PR introduces `get_study_names()`, specifically designed for faster retrieval of study names compared to `get_study_summaries()`.

## Description of the changes
- Implements `get_study_names()` based on the existing `get_study_summaries()` function.
- Adds tests for `get_study_names()`.
- Updates type annotations in `optuna/study/study.py` and `tests/study_tests/test_study.py`.

PR #5029, opened 2023-10-12
Add `optuna study-names` cli

Alnusjaponica:
## Motivation
Follow up https://github.com/optuna/optuna/pull/4898 to add cli of `get_all_study_names()`

## Description of the changes
Add `optuna study-names` with `--format` option

### DIFF
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -193,7 +193,7 @@ def _dump_value(records: List[Dict[str, Any]], header: List[str]) -> str:
         for column_name in header:
             row.append(str(record.get(column_name, "")))
         values.append(" ".join(row))
-    return os.linesep.join(values)
+    return "\n".join(values)
 
 
 def _dump_table(records: List[Dict[str, Any]], header: List[str]) -> str:
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -186,6 +186,16 @@ class CellValue:
             return f"{value:<{width}}"
 
 
+def _dump_value(records: List[Dict[str, Any]], header: List[str]) -> str:
+    values = []
+    for record in records:
+        row = []
+        for column_name in header:
+            row.append(str(record.get(column_name, "")))
+        values.append(" ".join(row))
+    return os.linesep.join(values)
+
+
 def _dump_table(records: List[Dict[str, Any]], header: List[str]) -> str:
     rows = []
     for record in records:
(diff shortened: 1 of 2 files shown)


optuna/cli.py
```diff
@@ -382,6 +382,30 @@ def take_action(self, parsed_args: Namespace) -> int:
...
+
+    def add_arguments(self, parser: ArgumentParser) -> None:
+        parser.add_argument(
+            "-f",
+            "--format",
+            type=str,
+            choices=("json", "table", "yaml"),
+            default="table",
+            help="Output format.",
+        )
```
c-bata, line 396: How about removing format option and just printing study names per line?

c-bata:

nabenabe0928:
> @nabenabe0928 Could you review this PR?

yes

nabenabe0928:
looks good to me:)
I also checked the behavior on my Linux machine and it was perfect!

nabenabe0928 (approved):
LGTM

contramundum53:
@c-bata ping

optuna/cli.py
c-bata: We discussed internally and decided to provide format option. Let me resolve this.

optuna/cli.py
```diff
@@ -382,6 +382,30 @@ def take_action(self, parsed_args: Namespace) -> int:
...
+class _StudyNames(_BaseCommand):
+    """Get all study names stored in a specified storage"""
+
+    def add_arguments(self, parser: ArgumentParser) -> None:
+        parser.add_argument(
+            "-f",
+            "--format",
+            type=str,
+            choices=("json", "table", "yaml"),
+            default="table",
```
c-bata, line 394: We discussed the default format and decided to output study names per line like below. Could you update the PR?

```
$ optuna study-names --storage ...
study_1
study_2
```

c-bata (dismissed):
Sorry for the delay. I left a comment.

optuna/cli.py
Alnusjaponica: I updated so that `study-names` output is default to `value` format. PTAL

HideakiImamura:

tests/test_cli.py
```diff
@@ -71,9 +71,12 @@ def _parse_output(output: str, output_format: str) -> Any:
         For table format, a list of dict formatted rows.
         For JSON or YAML format, a list or a dict corresponding to ``output``.
     """
-
-    if output_format == "table":
-        rows = output.split(os.linesep)
+    if output_format == "value":
+        # Currently, _parse_output with output_format="value" is used only for
+        # `study-names` command.
+        return [{"name": values} for values in output.split("\n")]
```
not522, line 77: Could you use `os.linesep` instead of `\n`? I think that Windows uses `\r\n` as the line separator in the output of CLI command.

Alnusjaponica: It seems that just replacing `os.linesep` does not resolve the error since the original implementation used them as in https://github.com/optuna/optuna/pull/5029/commits/e95c13a15e98675dc1545340b994227922876d7d. I'll look for the cause.

not522 (approved):
LGTM!
Let me merge it since @c-bata's concern has been resolved.

Issue #5036, opened 2023-10-13
Migrate from `threading.local` to `contextvars`


contramundum53:
### Motivation

https://github.com/optuna/optuna/blob/3fcbf8839c760916cc0cf38fb545f526d5efe85d/optuna/study/study.py#L56 uses `threading.local`, but `contextvars` can make the code more sophisticated (https://peps.python.org/pep-0567/#converting-code-that-uses-threading-local).

### Suggestion

Use `contextvars` instead of `threading.local`.

sousu4:
I will work on this issue.

SG5:
Hey @sousu4 I just saw your comment. I started working on this issue 3 days ago. Let me know if you want to pass it to me

sousu4:
HEY @SG5 OK,  Please proceed.

contramundum53:
After a discussion, we found a potential bug (discrepancy from the original design) in the original Study class. The original intention of the issue was to use contextvars to simplify the caching mechanism based on the original design (which looks impossible to do now without some breaking change in the current Study class), and we might need to investigate the potential bug first.

contramundum53:
In the original design, we didn't allow the `Study` class to cache trials when the user directly used `ask` and `tell`. The cache is only for `optimize` function. In that case, we could create a new context for each optimize call.

PR #5038, opened 2023-10-15
WIP: Migrate from threading.local to contextvars

SG5:
## Motivation

Fix issue #5036

## Description of the changes
Remove `_ThreadLocalStudyAttribute` class and move both properties to variables


HideakiImamura:
Thank you for your contribution. I'm sorry, but it seems that @sausu4 has already started working on the issue #5038. Could you close this PR? In the case you want to work on an issue, I would recommend first commenting on the issue before sending the solution PR to avoid conflicts and make the process smoother. I'm very sorry for the confusion.

PR #5041, opened 2023-10-15
Migrate from threading.local to contextvars

### LINKED ISSUE #5036: Migrate from `threading.local` to `contextvars`

SG5:
## Motivation

Migrate from threading.local to contextvars, fix #5036

## Description of the changes

Replace `_ThreadLocalStudyAttribute` class with `ContextVar` properties

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -96,7 +96,7 @@ class Study:
     def __setstate__(self, state: dict[Any, Any]) -> None:
         self.__dict__.update(state)
         self._in_optimize_loop = ContextVar("_in_optimize_loop", default=False)
-        self._reset_cached_trials()
+        self._cached_all_trials = ContextVar("_cached_all_trials", default=None)
 
     @property
     def best_params(self) -> dict[str, Any]:
@@ -280,9 +280,6 @@ class Study:
             return copy.deepcopy(filtered_trials) if deepcopy else filtered_trials
 
         return self._storage.get_all_trials(self._study_id, deepcopy=deepcopy, states=states)
-
-    def _reset_cached_trials(self) -> None:
-        self._cached_all_trials = ContextVar("_cached_all_trials", default=None)
 
     @property
     def user_attrs(self) -> dict[str, Any]:
@@ -537,7 +534,7 @@ class Study:
         }
 
         # Sync storage once every trial.
-        self._reset_cached_trials()
+        self._cached_all_trials.set(None)
 
         trial_id = self._pop_waiting_trial_id()
         if trial_id is None:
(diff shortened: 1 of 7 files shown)


contramundum53:
@c-bata @HideakiImamura Could you review this PR?

nabenabe0928:
> @c-bata @nabenabe0928 Could you review this PR?

Yes

optuna/study/study.py
```diff
@@ -536,7 +534,7 @@ def ask(
         }
 
         # Sync storage once every trial.
-        self._thread_local.cached_all_trials = None
+        self._cached_all_trials.set(None)
```
c-bata, line 537: Could you please create a new context for each `ask()` method call instead of clearing the cache?

c-bata (changes requested):

optuna/study/study.py
SG5: Hey @c-bata,

It won't work. Object `Study` is shared between threads. Once we create a new `_cached_all_trials` in one thread, all threads lose data in context.
Test `test_cache_is_invalidated` works fine for n_jobs=1 and fails for n_jobs>1 at `assert study._cached_all_trials.get() is not None`

Another possible solution - clone `Study` for each thread

HideakiImamura:
@SG5 Please pass the CI first. We, the reviewers, will review the PRs that have passed the CI. If you have difficulty passing the CI, it's okay to ask us for help by summarizing the error you're having trouble with and the way to reproduce it.

SG5:
@HideakiImamura rolled back changes with recreating context, CI passed

HideakiImamura:
@gen740 could you review this PR?

gen740 (approved):
LGTM!

gen740:
It would be better to merge this PR for now. I think that creating a new context, as suggested by @c-bata, should be addressed in a follow-up PR.

HideakiImamura:
@SG5 Please wait the merge since we, committers, will discuss the use of `ContextVar`. It will take several days.

contramundum53:
@SG5 We are sorry, after a discussion, we found a potential bug (discrepancy from the original design) in the original `Study` class. The original intention of the issue was to use `contextvars` to simplify the caching mechanism based on the original design (which looks impossible to do now without some breaking change in the current `Study` class), and we might need to investigate the potential bug first. 
Let us close this PR for now. Thanks for working on this issue!

contramundum53:
In the original design, we didn't allow the `Study` class to cache trials when the user directly used `ask` and `tell`. The cache is only for `optimize` function. In that case, we could create a new context for each `optimize` call.

PR #5042, opened 2023-10-16
Install PyTorch for CPU on CIs

not522:
## Motivation
PyPI's PyTorch installs CUDA-related libraries, which waste disk space. It may result CI fails. `--extra-index-url https://download.pytorch.org/whl/cpu` prevents it.

## Description of the changes
Install PyTorch for CPU on CIs.


contramundum53 (approved):
LGTM.

PR #5043, opened 2023-10-16
Remove non oldest and latest Python versions from tests

nabenabe0928:
## Motivation

As the CI is slow with tests on all the Python versions, I made a change so that the tests will run only on the oldest and the latest supported Python versions.
Note that we still maintain the tests on all the versions for scheduled tests to make sure that Optuna still works on all the versions.

## Description of the changes

More specifically, the changes are:
1. Remove the Python version 3.8, 3.9, 3.10 from tests, and
2. Maintain these versions at least for scheduled tests.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -27,8 +27,8 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    # TODO (nzw0301): Remove `exclude` once integration works with Python 3.11.
     # NOTE (nabenabe0928): 3.8 exists for the check of the oldest version with dev.
+    # TODO (nabenabe0928): Remove `exclude` of Python3.8 x non-dev once 3.7 is removed.
     strategy:
       matrix:
         python_version: ['3.7', '3.8', '3.11']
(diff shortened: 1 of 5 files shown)


contramundum53:

.github/workflows/dockerimage.yml
```diff
@@ -28,15 +28,14 @@ jobs:
...
 
     # TODO (nzw0301): Remove `exclude` once integration works with Python 3.11.
+    # NOTE (nabenabe0928): 3.8 exists for the check of the oldest version with dev.
     strategy:
       matrix:
-        python_version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python_version: ['3.7', '3.8', '3.11']
         build_type: ['', 'dev']  # "dev" installs all the dependencies including pytest.
         exclude:
         - python_version: '3.7'
```
contramundum53, line 37: I think you can also exclude {python_version: '3.8', build_type: ''}.

nabenabe0928: I did not remove 3.8 because 3.7x'dev' does not pass and that is why it is skipped right now.
I could exclude 3.8x'' and let 3.8x'dev' stay as it did.
What do @contramundum53  think?

nabenabe0928:
I will open another PR for the same purpose after [PR#5067](https://github.com/optuna/optuna/pull/5067) is merged.

Issue #3324, opened 2022-02-19
Migrate to use keyword-only arguments


higucheese:
### Motivation

Current function calls using positional arguments limit changes to function signatures by developers in order to minimize code changes on the user side. In contrast, keyword arguments allow them to change the signature freely, with exceptions such as reducing the number of arguments.

### Description

By making all arguments of major functions such as `create_study` / `load_study` keyword-only arguments, developers can change the signatures without changing on the userside. Conventional calls with positional arguments will be changed to keyword-only arguments internally, so that user code will not be affected. At least in v3, it would be better to only issue a FutureWarning for the use of positional arguments, and depending on the response, consider whether we should unify the use of keyword-only arguments.

### Additional context (optional)

- Based on #2911.
  - Changing the order of keyword arguments does not affect the caller.
- Implemented in #3270.

not522:
Are there any remaining tasks for this issue?

higucheese:
Since this issue was created as an explanation for FutureWarning of using positional arguments, this should be closed after the FutureWarning is no more needed. It seems a bit difficult to decide remaining tasks to close this issue by myself. Do you have any ideas? @himkt

himkt:
This issue played the role as the place where users send feedbacks if there are. Since it would take a long time until we can discuss deleting `FutureWarning`, I think it's OK to close this issue now.

himkt:
I asked other developers for the opinion and decided to close. Thank you @higucheese for your great contribution! 🚀

PR #5044, opened 2023-10-16
Make positional args to kwargs in suggest_int

nabenabe0928:
## Motivation

As `suggest_float` has `*` before `steps` and `log` so that users need to pass them as `kwargs`, I applied the same to `suggest_int`.

## Description of the changes

I made the following changes:
1. Change `suggest_int(name, low, high, step, log)` into `suggest_int(name, low, high, *, step, log)` to strengthen the future safety, and
2. Wrap `suggest_int` with `convert_positional_args` for now to avoid breaking changes.

### DIFF
--- a/optuna/_convert_positional_args.py
+++ b/optuna/_convert_positional_args.py
@@ -28,9 +28,9 @@ def _get_positional_arg_names(func: "Callable[_P, _T]") -> list[str]:
     return positional_arg_names
 
 
-def _infer_given_args(previous_positional_arg_names: Sequence[str], *args: Any) -> dict[str, Any]:
-    inferred_args = {arg_name: val for val, arg_name in zip(args, previous_positional_arg_names)}
-    return inferred_args
+def _infer_kwargs(previous_positional_arg_names: Sequence[str], *args: Any) -> dict[str, Any]:
+    inferred_kwargs = {arg_name: val for val, arg_name in zip(args, previous_positional_arg_names)}
+    return inferred_kwargs
 
 
 def convert_positional_args(
@@ -54,11 +54,11 @@ def convert_positional_args(
         @wraps(func)
         def converter_wrapper(*args: Any, **kwargs: Any) -> "_T":
             positional_arg_names = _get_positional_arg_names(func)
-            inferred_args = _infer_given_args(previous_positional_arg_names, *args)
-            if len(inferred_args) > len(positional_arg_names):
-                kwargs_expected = set(inferred_args) - set(positional_arg_names)
+            inferred_kwargs = _infer_kwargs(previous_positional_arg_names, *args)
+            if len(inferred_kwargs) > len(positional_arg_names):
+                expected_kwds = set(inferred_kwargs) - set(positional_arg_names)
                 warnings.warn(
-                    f"{func.__name__}() got {kwargs_expected} as positional arguments "
+                    f"{func.__name__}() got {expected_kwds} as positional arguments "
                     "but they were expected to be given as keyword arguments.",
                     FutureWarning,
                     stacklevel=warning_stacklevel,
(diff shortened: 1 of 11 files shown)


nabenabe0928:
I have a question.
This PR does not improve the safety of the software by itself, but is it still fine?

contramundum53:

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM. I have two minor comments. PTAL.

By the way, as you mentioned, this PR does not force anything on the users as it is a change aimed at continuing support for the older API. However, this change is a milestone for the future. It is expected that the older API will no longer be supported in a future major version, and by making these changes in the code, it becomes easier for us to keep motivation.

tests/multi_objective_tests/test_trial.py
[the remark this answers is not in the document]
nabenabe0928: Just to make sure, but did you mean "do not have to" or "discouraging" (In other words, should I remove it)?

optuna/trial/_base.py
```diff
@@ -11,6 +11,9 @@
 from optuna.distributions import CategoricalChoiceType
 
 
+_SUGGEST_INT_POSITIONAL_ARGS = ["self", "name", "low", "high", "step", "log"]
```
not522, line 14: Could you make only `step` and `log` positional? Currently `convert_positional_args` does not support partial conversion, but you can use `inspect.signature` to achieve this.

nabenabe0928: Hi sorry for the late response:(
The variable name was actually misleading because these are "previous" positional arguments, but not the current positional arguments.
Plus, since we do not have any non-positional arguments in `suggest_int` before this PR, I would like to simply change the variable name and let the `convert_positional_args` stay as it does.
If necessary, I can also make another PR for this.
How do you like it? (@not522)

not522: Oh, I'm sorry, it's my mistake. I wanted to say "make only `step` and `log` keyword-only".

nabenabe0928: Certainly!
Then I will change the code so that inappropriate args usages will be warned!

HideakiImamura:
@nabenabe0928 The checks CI fails. Could you fix them?

not522 (commented):
I found unnecessary warnings, for example, `suggest_int("x", 0, 9)` outputs `FutureWarning`. I think that the check for the length in `converter_wrapper` is not correct.

nabenabe0928:
> I found unnecessary warnings, for example, `suggest_int("x", 0, 9)` outputs `FutureWarning`. I think that the check for the length in `converter_wrapper` is not correct.

Now I changed so that the `FutureWarning` happens only if keyword-only is specified as arguments.
Also the new implementation warns only once rather than raising warnings for each argument.

optuna/_convert_positional_args.py
```diff
@@ -37,9 +53,13 @@ def converter_decorator(func: "Callable[_P, _T]") -> "Callable[_P, _T]":
...
-            if len(args) >= 1:
+            positional_arg_names = _get_positional_arg_names(func)
+            inferred_args = _infer_given_args(previous_positional_arg_names, *args)
+            if len(inferred_args) > len(positional_arg_names):
+                kwargs_expected = set(inferred_args) - set(positional_arg_names)
                 warnings.warn(
-                    f"{func.__name__}(): Please give all values as keyword arguments."
+                    f"{func.__name__}() got {kwargs_expected} as positional arguments "
+                    "but they were expected to be given as keyword arguments."
                     " See https://github.com/optuna/optuna/issues/3324 for details.",
```
not522, line 63: It's almost LGTM. How about removing this line because this error is not only related to #3324?

nabenabe0928: Right, because of the last change, #3324 does not relate to this line anymore. (while it still makes sense to use kwargs only, the future warning for it is not the default Python design, so I removed it and stick to your suggestion.)

not522 (approved):
LGTM!

HideakiImamura (approved):
LGTM.

PR #5053, opened 2023-10-19
add hint how to customize plotly figures generated by optuna

gituser789:
## Motivation
When using a plotly graph generated by optuna, it is not possible to change things according to the layout. E.g. it is not possible to choose a title for your graph. Some people like to use this functionality and might search a long time how to do this.

## Description of the changes
I have added a small hintbox to the `plot_pareto_front()` function documentation which links to the plotly `update_layout()` method.

### DIFF
--- a/optuna/visualization/_pareto_front.py
+++ b/optuna/visualization/_pareto_front.py
@@ -72,6 +72,9 @@ def plot_pareto_front(
 
             fig = optuna.visualization.plot_pareto_front(study)
             fig.show()
+            
+        .. seealso::
+	        To customize the generated diagram afterwards (change title, ...), use the function 'update_layout()' of plotly. See https://plotly.com/python/reference/layout/.
 
     Example:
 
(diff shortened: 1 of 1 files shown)


contramundum53:
image
Thank you for adding explanation. LGTM!

optuna/visualization/_pareto_front.py
```diff
@@ -73,6 +73,10 @@ def objective(trial):
             fig = optuna.visualization.plot_pareto_front(study)
             fig.show()
 
+        .. seealso::
+            To customize the generated diagram afterwards (change title, ...), use the
+            function 'update_layout()' of plotly. See https://plotly.com/python/reference/layout/.
```
contramundum53, line 78:
```suggestion
            function ``update_layout()`` of plotly. See https://plotly.com/python/reference/layout/.
```
@gituser789 How about using a code block? It might look better than simply quoting it.

gituser789:
I have updated the code block as suggested

HideakiImamura:
I believe this advice is more common to all visualization features, rather than specific to `plot_pareto_front`. For instance, how about documenting this content in `docs/source/reference/visualization/index.rst`?

optuna/visualization/_pareto_front.py
contramundum53: image
Thank you, now it looks better.

Issue #5061, opened 2023-10-20
`GridSampler` currently does not support Sequences as a type of categorical feature


Koen1999:
### Motivation

I would like to perform grid search, where one of the parameters is categorical. The type of the categorical feature is given by `Sequence[str]`.

More specifically, I want to supply the following to the `GridSampler`as `search_space`:
`{'groupby': (('src',), ('dst',), ('src', 'dst',), ('src_dst',), ('src', 'src_dst',), ('dst', 'src_dst',), ('src', 'dst', 'src_dst',)),}`

### Description

I think `GridSampler` should support this feature, since a `Sequence` can be considered as a slightly more complex type of categorical parameter.

I do think one constraint to be put on the Sequence is that it should be hashable. Otherwise, comparison of grid search spaces will fail.

### Additional context (optional)

This also triggered a downstream error in `optuna-dashboard`. Please see: https://github.com/optuna/optuna-dashboard/issues/665

This also related to the `CategoricalChoiceType` and `CategoricalDistribution`, which do not support this behavior either at the moment.

Koen1999:
I just discovered there are several related older issues that were never addressed: https://github.com/optuna/optuna/issues/3292 https://github.com/optuna/optuna/issues/2341 https://github.com/optuna/optuna/issues/905 https://github.com/optuna/optuna/issues/3093#issuecomment-968075749
Tagged to keep all the issues somewhat together.

PR #5063, opened 2023-10-20
Added support for `tuple` to `CategoricalChoiceType` and `CategoricalDistribution`

Koen1999:
## Motivation
Optuna currently does not support sequences as a type for categorical parameters. This was a feature suggested to improve Optuna's flexibility in https://github.com/optuna/optuna/issues/5061 and caused problems with the dashboard as well, as indicated in https://github.com/optuna/optuna-dashboard/issues/665.

## Description of the changes
`tuple` was added as an allowed type for `CategoricalChoiceType`, with the additional requirement that the elements of the `tuple` are also of type `CategoricalChoiceType`.
The initialization of `CategoricalDistribution` was changed accordingly, as well as its DocStrings.
A new test case (or a parameter to an already existing test case) was added to also test this functionality.

How this works internally with JSON serialization, is that tuples are formatted as lists in JSON. When reading, they hence are lists, but they are converted to tuples in `json_to_distribution`.

This issue solves https://github.com/optuna/optuna/issues/5061 and https://github.com/optuna/optuna-dashboard/issues/665

Since you might wonder how this works downstream in the dashboard, it works excellent as visualized below!
image

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -7,6 +7,7 @@ from typing import Any
 from typing import cast
 from typing import Dict
 from typing import Sequence
+from typing import Tuple
 from typing import Union
 import warnings
 
@@ -15,7 +16,7 @@ import numpy as np
 from optuna._deprecated import deprecated_class
 
 
-CategoricalChoiceType = Union[None, bool, int, float, str]
+CategoricalChoiceType = Union[None, bool, int, float, str, Tuple["CategoricalChoiceType", ...]]
 
 
 _float_distribution_deprecated_msg = (
@@ -503,7 +504,7 @@ class CategoricalDistribution(BaseDistribution):
 
         Not all types are guaranteed to be compatible with all storages. It is recommended to
         restrict the types of the choices to :obj:`None`, :class:`bool`, :class:`int`,
-        :class:`float` and :class:`str`.
+        :class:`float`, :class:`str`, and :class`tuple`.
 
     Attributes:
         choices:
@@ -515,11 +516,11 @@ class CategoricalDistribution(BaseDistribution):
         if len(choices) == 0:
             raise ValueError("The `choices` must contain one or more elements.")
         for choice in choices:
-            if choice is not None and not isinstance(choice, (bool, int, float, str)):
+            if choice is not None and not isinstance(choice, (bool, int, float, str, tuple)):
                 message = (
                     "Choices for a categorical distribution should be a tuple of None, bool, "
-                    "int, float and str for persistent storage but contains {} which is of type "
-                    "{}.".format(choice, type(choice).__name__)
+                    "int, float, str, and tuple for persistent storage but contains {} which "
+                    "is of type {}.".format(choice, type(choice).__name__)
                 )
                 warnings.warn(message)
 
@@ -591,7 +592,10 @@ def json_to_distribution(json_str: str) -> BaseDistribution:
 
     if "name" in json_dict:
         if json_dict["name"] == CategoricalDistribution.__name__:
-            json_dict["attributes"]["choices"] = tuple(json_dict["attributes"]["choices"])
+            choices = json_dict["attributes"]["choices"]
+            # Cast list to tuple to ensure hashability
+            choices = [tuple(choice) if isinstance(choice, list) else choice for choice in choices]
+            json_dict["attributes"]["choices"] = tuple(choices)
 
         for cls in DISTRIBUTION_CLASSES:
             if json_dict["name"] == cls.__name__:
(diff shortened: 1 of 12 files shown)


Koen1999:
I can see that a code formatting check is failing because the line is too long. I am not sure how to fix it since the string needs to be broken up. Suggestions are welcome.

HideakiImamura:
You can fix the line as follows.

```python
"c4": (
    '{"name": "CategoricalDistribution", '
    '"attributes": {"choices": [["src"], ["dst"], ["src", "dst"]]}}'
),
```

HideakiImamura:
Thank you for your contribution. We have discussed this change among committers and have concluded not to include such changes in Optuna. The reasons are as follows:

- We want to avoid ad-hoc responses such as adding only tuples when various sequences exist.
- In the current implementation, type conversion does not work well in a nested search space where a tuple exists within a tuple. We want to avoid recursively converting, which leads to complexity in the implementation.
- Most importantly, there is a simple workaround to achieve the same thing. Specifically, it's okay to define a categorical distribution using a string that corresponds to each tuple, as shown below:

```python
import optuna


def objective(trial):
    categories = [('src',), ('dst',), ('src', 'dst')]
    c = trial.suggest_categorical('c', [str(cat) for cat in categories])
    cat = eval(c)
    assert cat in categories
    return 0


if __name__ == '__main__':
    study = optuna.create_study()
    print(study.best_params)

```

We apologize for not incorporating your contribution after you've put in so much effort. We sincerely appreciate you creating numerous issues, not just this PR. We look forward to further collaborations.

Koen1999:
> Thank you for your contribution. We have discussed this change among committers and have concluded not to include such changes in Optuna. The reasons are as follows:
> 
>     * We want to avoid ad-hoc responses such as adding only tuples when various sequences exist.
> 
>     * In the current implementation, type conversion does not work well in a nested search space where a tuple exists within a tuple. We want to avoid recursively converting, which leads to complexity in the implementation.
> 
>     * Most importantly, there is a simple workaround to achieve the same thing. Specifically, it's okay to define a categorical distribution using a string that corresponds to each tuple, as shown below:
> 
> 
> ```python
> import optuna
> 
> 
> def objective(trial):
>     categories = [('src',), ('dst',), ('src', 'dst')]
>     c = trial.suggest_categorical('c', [str(cat) for cat in categories])
>     cat = eval(c)
>     assert cat in categories
>     return 0
> 
> 
> if __name__ == '__main__':
>     study = optuna.create_study()
>     study.optimize(objective, n_trials=10)
>     print(study.best_params)
> ```
> 
> We apologize for not incorporating your contribution after you've put in so much effort. We sincerely appreciate you creating numerous issues, not just this PR. We look forward to further collaborations.

Thanks for discussing the PR with the other collaborators and giving your thoughts.
While I understand the hesitation, I am convinced it should be implemented. I will give reasons below, but understand the maintainers have the final say regarding this issue.

1. Different types of sequences indeed exist, but they all represent the same thing: an ordered collection. The reason different types exist is that they support different operations or are more effective in certain operations. For the purpose of hyper parameter optimization, however, they behave exactly the same. The hashable and immutable type of a tuple is therefore fit for representing all kinds of sequences.
2. That's indeed a valid point. I changed the implementation and the tests to apply (de)-serialization recursively. As can be seen in the code, this is not a huge complexity. The tests indicate that recursive de-serialization is not a problem.
3. I guess this point boils down to what is expected of optuna is a framework. When skimming through past issues, there have been numerous people who asked about this feature or a similar feature as indicated in https://github.com/optuna/optuna/issues/5061#issuecomment-1772911658. I think the library should adapt to match user expectations more closely. While the workaround you present would work, it is not user friendly.

I would appreciate it if you take these comments into consideration, but I understand if you do not want to merge.

not522:
Thank you for your proposal. I agree supporting `tuple` may be a nice-to-have feature, and some users have faced this problem. However, we should refine the design of the API to avoid patchwork.
We have had many discussions about `choices`, including supporting other types and dynamic search space. But still, we are not confident that those changes are worth abandoning the current simplicity.
In my personal opinion, it is non-intuitive why only `tuple` is supported instead of `Any` or JSON serializable values. Sometimes, users have requested them, and I may want to use them as a user. Of course, there is a case for incrementally adding functionality, but if it is unclear what goal we are trying to achieve, Optuna will end up with a collection of half-baked code that is not easy to maintain.
I appreciate your opinion, and thank you for writing PR. My point is not that there is anything wrong with this PR. But unless we carefully discuss other options, it is difficult to maintain the simplicity and stability of Optuna.

HideakiImamura:
We, the committers, have discussed this change. There are various options for revisiting the current CategoricalChoiceType, all of which would have a significant impact on the core part of Optuna, greatly affecting its simplicity and maintainability. We have decided to take some time to consider this further. We, the committers, are well aware that there is a demand for such features from the community, but we have decided to not merge these changes from this PR at this time. We will, however, continue to consider this points. Thank you for your contribution.

PR #5067, opened 2023-10-23
Reduce `tests-mpi` to the oldest and latest Python versions

nabenabe0928:
## Motivation

As the CI is slow with tests on all the Python versions, I made a change so that the tests will run only on the oldest and the latest supported Python versions.
For the first step, I applied the changes to tests mpi, which is not one of the required tests.
Note that we still maintain the tests on all the versions for scheduled tests to make sure that Optuna still works on all the versions.

## Description of the changes

More specifically, the changes are:
1. Remove the Python version 3.9, 3.10 from tests, and
2. Maintain these versions at least for scheduled tests.

### DIFF
--- a/.github/workflows/tests-integration.yml
+++ b/.github/workflows/tests-integration.yml
@@ -15,7 +15,7 @@ concurrency:
 
 jobs:
   tests-integration:
-    if: false && (github.event_name == 'schedule' && github.repository == 'optuna/optuna') || (github.event_name != 'schedule')
+    if: (github.event_name == 'schedule' && github.repository == 'optuna/optuna') || (github.event_name != 'schedule')
     runs-on: ubuntu-latest
 
     strategy:
--- a/.github/workflows/tests-integration.yml
+++ b/.github/workflows/tests-integration.yml
@@ -15,7 +15,7 @@ concurrency:
 
 jobs:
   tests-integration:
-    if: (github.event_name == 'schedule' && github.repository == 'optuna/optuna') || (github.event_name != 'schedule')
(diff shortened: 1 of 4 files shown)


HideakiImamura:

HideakiImamura (approved):
LGTM.

While a method to terminate jobs early would make scripting more straightforward, it may not be supported according to [this discussion](https://github.com/actions/runner/issues/662). 

In this approach, jobs that are not actually executed on a per-PR basis are still displayed in the CI list on the PR screen. I believe this is one of the disadvantages.

Another alternative could be to split the workflow definition files into ones that are run per PR and those that are scheduled. However, this doesn't seem to be a particularly appropriate measure as the number of yml files to manage would simply double.

not522:
@HideakiImamura The matrix can be dynamically changed. How about this approach? https://stackoverflow.com/questions/59977364/github-actions-how-use-strategy-matrix-with-script

nabenabe0928:
> @HideakiImamura The matrix can be dynamically changed. How about this approach? https://stackoverflow.com/questions/59977364/github-actions-how-use-strategy-matrix-with-script

I updated the yaml file with the dynamic matrix:)

nabenabe0928:
> @HideakiImamura The matrix can be dynamically changed. How about this approach? https://stackoverflow.com/questions/59977364/github-actions-how-use-strategy-matrix-with-script

@not522 
I updated the yaml file with the dynamic matrix:)

not522:
Could you resolve this warning?
image

nabenabe0928:
> Could you resolve this warning? image

Sure!

not522 (approved):
LGTM!
@HideakiImamura If you approve the change, could you merge this PR?

HideakiImamura (approved):
LGTM.

PR #5075, opened 2023-10-26
Update the FAQ on reproducible optimization results to remove note on HyperbandPruner

felix-cw:
## Motivation

The FAQ contains a note on HyperbandPruner advising setting the python hash seed to ensure reproducible results.
It's in the paragraph after the code block.
It seems like this this is no longer applicable since #4131 and a similar note on in the API docs for HyperbandPruner was removed at that time

## Description of the changes

Remove paragraph
```rst
To make the pruning by :class:`~optuna.pruners.HyperbandPruner` reproducible, you can specify ``study_name`` of :class:`~optuna.study.Study` and `hash seed <https://docs.python.org/3/using/cmdline.html#envvar-PYTHONHASHSEED>`_.
```

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -182,8 +182,6 @@ To make the parameters suggested by Optuna reproducible, you can specify a fixed
     study = optuna.create_study(sampler=sampler)
     study.optimize(objective)
 
-To make the pruning by :class:`~optuna.pruners.HyperbandPruner` reproducible, you can specify ``study_name`` of :class:`~optuna.study.Study` and `hash seed <https://docs.python.org/3/using/cmdline.html#envvar-PYTHONHASHSEED>`_.
-
 However, there are two caveats.
 
 First, when optimizing a study in distributed or parallel mode, there is inherent non-determinism.


HideakiImamura:
@eukaryo @nabenabe0928 Could you review this PR?

nabenabe0928 (commented):
I tried the following code with different setups for `study_name`, `n_jobs`, and `seed`.

```python
import optuna

study_name = ...
n_jobs = ...
seed = ...

def objective(trial):
    x = trial.suggest_float("x", -5.0, 5.0)
    for i in range(100):
        trial.report(x + (100 - i) * (6 - x), i + 1)

    return x

study = optuna.create_study(
    sampler=optuna.samplers.TPESampler(seed=seed),
    pruner=optuna.pruners.HyperbandPruner(min_resource=1, max_resource=101, reduction_factor=3),
    study_name=study_name,
)
study.optimize(objective, n_trials=30, n_jobs=n_jobs)
```

So basically, the result is reproducible only if `n_jobs=1` and `seed is not None`.
Probably, we could replace the statement here like this as `HyperbandPruner` by itself cannot guarantee reproducible optimizations.

felix-cw:
> So basically, the result is reproducible only if `n_jobs=1` and `seed is not None`. Probably, we could mention this as `HyperbandPruner` by itself cannot guarantee reproducible optimizations.

Thanks for having a look!
Just above in the FAQ it says to set a seed in the sampler for reproducibility and below it says that reproducibility is only feasible when studies are run sequentially.  I think the latter point corresponds to setting `n_jobs=1`.

nabenabe0928 (approved):
> > So basically, the result is reproducible only if `n_jobs=1` and `seed is not None`. Probably, we could mention this as `HyperbandPruner` by itself cannot guarantee reproducible optimizations.
> 
> Thanks for having a look! Just above in the FAQ it says to set a seed in the sampler for reproducibility and below it says that reproducibility is only feasible when studies are run sequentially. I think the latter point corresponds to setting `n_jobs=1`.

You are right! I approved the change:)

eukaryo (approved):
I read the conversation, LGTM.

nzw0301:
Actually, we still need to `study_name` for reproducibility. I will send a PR to partially revert this and the parent PR.

PR #5079, opened 2023-10-27
Return `trials` for above in MO split when n_below=0

nabenabe0928:
* The last one was returning an empty list, which was incorrect

## Motivation

As the last implementation returns an empty list when `n_below=0` and the union of above and below should be `trials`, I simply return `trials` instead of an empty list. 

## Description of the changes

I replaced `[]` with `trials`.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -646,7 +646,7 @@ def _split_complete_trials_multi_objective(
     n_below: int,
 ) -> tuple[list[FrozenTrial], list[FrozenTrial]]:
     if n_below == 0:
-        return [], []
+        return [], trials
 
     lvals = np.asarray([trial.values for trial in trials])
     for i, direction in enumerate(study.directions):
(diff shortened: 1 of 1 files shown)


optuna/samplers/_tpe/sampler.py
```diff
@@ -646,7 +646,8 @@ def _split_complete_trials_multi_objective(
     n_below: int,
 ) -> tuple[list[FrozenTrial], list[FrozenTrial]]:
     if n_below == 0:
-        return [], []
+        # The type of trials must be `list`, but not `Sequence`.
+        return [], [t for t in trials]
```
not522, line 650: How about this?

```suggestion
        return [], list(trials)
```

nabenabe0928: I changed the code:)

optuna/samplers/_tpe/sampler.py
```diff
@@ -646,7 +646,8 @@ def _split_complete_trials_multi_objective(
     n_below: int,
 ) -> tuple[list[FrozenTrial], list[FrozenTrial]]:
     if n_below == 0:
-        return [], []
+        # The type of trials must be `list`, but not `Sequence`.
```
knshnb, line 649: [Nit] I feel this comment is a bit obvious and not necessary.
```suggestion
```

knshnb (approved):
Thanks for the PR! LGTM.

not522 (approved):
LGTM

PR #5084, opened 2023-10-27
Remove `MOTPESampler` from `index.rst` file

Ashhar-24:
## Motivation

Resolved Issue #5072 

## Description of the changes

As the issue #5072 says that `optuna.samplers.MOTPESampler` is deprecated from v2.9.0, therefore removal of it from the documentation is needed to avoid misunderstanding.
I have removed the line `optuna.samplers.MOTPESampler` from `index.rst` file present in `doc\source\reference\samplers` directive.

PR #4167, opened 2022-11-14
Remove the document of the `multi_objective` module

not522:
## Motivation
The `multi_objective` module has been deprecated. But it is still in the document, which makes users confused.
This PR removes all the documents of the `multi_objective` module.

## Description of the changes
- Remove the document of the `multi_objective` module

PR #5086, opened 2023-11-01
Remove MOTPESampler from the doc because it is deprecated

nabenabe0928:
## Motivation

Since `MOTPESampler` has been already deprecated and `TPESampler` directly supports multi-objective optimization after v2.9.0, we decided to remove the documentation of `MOTPESampler` and add supplementary information for it to `TPESampler`.

## Description of the changes

- Add a guide to `TPESamper` doc and its alternative coding using `TPESampler` in the `MOTPESampler` doc,
- Divide the references for `MOTPE` so that the new reference self-explains the information source for both, and
- Add a code example of multi-objective optimization using `TPESampler` to the `TPESampler` doc.

### DIFF
--- a/optuna/multi_objective/samplers/_motpe.py
+++ b/optuna/multi_objective/samplers/_motpe.py
@@ -28,6 +28,28 @@ class MOTPEMultiObjectiveSampler(BaseMultiObjectiveSampler):
 
     This sampler is a multiobjective version of :class:`~optuna.samplers.TPESampler`.
 
+    .. note::
+        For `v2.9.0 <https://github.com/optuna/optuna/releases/tag/v2.9.0>`_ or later,
+        MOTPEMultiObjectiveSampler is deprecated and :class:`~optuna.samplers.TPESampler` should be
+        used instead. The following code shows how you run TPESampler on a multi-objective task:
+
+        .. testcode::
+
+            import optuna
+
+            def objective(trial):
+                x = trial.suggest_float("x", -100, 100)
+                y = trial.suggest_categorical("y", [-1, 0, 1])
+                f1 = x**2 + y
+                f2 = -((x - 2) ** 2 + y)
+                return f1, f2
+
+
+            # We minimize the first objective and the second objective.
+            sampler = optuna.samplers.TPESampler()
+            study = optuna.create_study(directions=["minimize", "maximize"], sampler=sampler)
+            study.optimize(objective, n_trials=100)
+
     For further information about MOTPE algorithm, please refer to the following paper:
 
     - `Multiobjective tree-structured parzen estimator for computationally expensive optimization
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -74,11 +74,16 @@ class TPESampler(BaseSampler):
       <https://papers.nips.cc/paper/4443-algorithms-for-hyper-parameter-optimization.pdf>`_
     - `Making a Science of Model Search: Hyperparameter Optimization in Hundreds of
       Dimensions for Vision Architectures <http://proceedings.mlr.press/v28/bergstra13.pdf>`_
-    - `Multiobjective tree-structured parzen estimator for computationally expensive optimization
-      problems <https://dl.acm.org/doi/10.1145/3377930.3389817>`_
+    - `Tree-Structured Parzen Estimator: Understanding Its Algorithm Components and Their Roles for
+      Better Empirical Performance <https://arxiv.org/abs/2304.11127>`_
+
+    For multi-objective TPE (MOTPE), please refer to the following papers:
+    - `Multiobjective Tree-Structured Parzen Estimator for Computationally Expensive Optimization
+      Problems <https://dl.acm.org/doi/10.1145/3377930.3389817>`_
     - `Multiobjective Tree-Structured Parzen Estimator <https://doi.org/10.1613/jair.1.13188>`_
 
     Example:
+        An example of a single-objective optimization is as follows:
 
         .. testcode::
 
@@ -93,6 +98,28 @@ class TPESampler(BaseSampler):
 
             study = optuna.create_study(sampler=TPESampler())
             study.optimize(objective, n_trials=10)
+
+    .. note::
+        For `v2.9.0 <https://github.com/optuna/optuna/releases/tag/v2.9.0>`_ or later,
+        :class:`~optuna.samplers.MOTPESampler` is deprecated and TPESampler should be used instead.
+        The following code shows how you run TPESampler on a multi-objective task:
+
+        .. testcode::
+
+            import optuna
+
+            def objective(trial):
+                x = trial.suggest_float("x", -100, 100)
+                y = trial.suggest_categorical("y", [-1, 0, 1])
+                f1 = x**2 + y
+                f2 = -((x - 2) ** 2 + y)
+                return f1, f2
+
+
+            # We minimize the first objective and the second objective.
+            sampler = optuna.samplers.TPESampler()
+            study = optuna.create_study(directions=["minimize", "maximize"], sampler=sampler)
+            study.optimize(objective, n_trials=100)
 
     Args:
         consider_prior:


not522:
@knshnb @Alnusjaponica Could you review this PR?

optuna/samplers/_tpe/sampler.py
```diff
@@ -94,6 +100,29 @@ def objective(trial):
             study = optuna.create_study(sampler=TPESampler())
             study.optimize(objective, n_trials=10)
 
+    .. note::
+        For `v2.9.0 <https://github.com/optuna/optuna/releases/tag/v2.9.0>`_ or later,
+        :class:`~optuna.samplers.MOTPESampler` is deprecated and this sampler should be used
+        instead. The following code shows how you apply this sampler to a multi-objective task:
```
Alnusjaponica, line 106: It might be confusing to use remark as "this sampler". How about writing down  the name.
```suggestion
        :class:`~optuna.samplers.MOTPESampler` is deprecated and
        :class:`~optuna.samplers.TPESampler` should be used instead. The following code shows
        how you apply :class:`~optuna.samplers.TPESampler` to a multi-objective task:
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -94,6 +100,29 @@ def objective(trial):
...
+
+            def objective(trial):
+                x = trial.suggest_float("x", -100, 100)
+                y = trial.suggest_categorical("y", [-1, 0, 1])
+                f1 = x**2 + y
+                f2 = -((x - 2) ** 2 + y)
+                return f1, f2
+
+
+            # We minimize the first objective and the second objective.
```
Alnusjaponica, line 121: Nit
```suggestion
            # We minimize the first objective and maximize the second objective.
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -74,11 +74,17 @@ class TPESampler(BaseSampler):
       <https://papers.nips.cc/paper/4443-algorithms-for-hyper-parameter-optimization.pdf>`_
     - `Making a Science of Model Search: Hyperparameter Optimization in Hundreds of
       Dimensions for Vision Architectures <http://proceedings.mlr.press/v28/bergstra13.pdf>`_
-    - `Multiobjective tree-structured parzen estimator for computationally expensive optimization
-      problems <https://dl.acm.org/doi/10.1145/3377930.3389817>`_
+    - `Tree-Structured Parzen Estimator: Understanding Its Algorithm Components and Their Roles for
+      Better Empirical Performance <https://arxiv.org/abs/2304.11127>`_
```
Alnusjaponica, line 78: I wonder where in the codebase this paper is from.

optuna/multi_objective/samplers/_motpe.py
```diff
@@ -26,9 +26,33 @@
...
+
+            def objective(trial):
+                x = trial.suggest_float("x", -100, 100)
+                y = trial.suggest_categorical("y", [-1, 0, 1])
+                f1 = x**2 + y
+                f2 = -((x - 2) ** 2 + y)
+                return f1, f2
+
+
+            # We minimize the first objective and the second objective.
```
Alnusjaponica, line 50: Ditto.
```suggestion
            # We minimize the first objective and maximize the second objective.
```

Alnusjaponica (commented):
In my understanding, `multi_objective` module has already been removed from docs by https://github.com/optuna/optuna/pull/4167 and https://github.com/optuna/optuna/pull/5084 ~and wonder if it is really necessary to add notes to `optuna/multi_objective/samplers/_motpe.py`~.
I also made some comments. PTAL.

optuna/samplers/_tpe/sampler.py
nabenabe0928: I can also discuss whether the reference should include this paper (, which did not exist before) or not, but it actually makes sense to add this paper as long as the documentation says "For further information about TPE algorithm, please refer to the following papers:".

Just as a comment, I add the short summary of the paper.
The paper describes the control parameters of TPE and their effects on search in detail.
Importantly, each control parameter refers to the exact Optuna argument names in the paper.
The other four papers do not discuss each control parameter and its meaning.

optuna/samplers/_tpe/sampler.py
nabenabe0928: Thank you, I confirmed that self-referencing is common in the documentation, so I will take your suggestion.
https://optuna.readthedocs.io/en/stable/reference/optuna.html

optuna/samplers/_tpe/sampler.py
Alnusjaponica: Thanks for the explanation. I see your point but it sound better to split the PR. 
Could you create another PR for adding the reference?

optuna/samplers/_tpe/multi_objective_sampler.py
```diff
@@ -22,9 +22,33 @@ def _default_weights_above(x: int) -> np.ndarray:
...
+
+            def objective(trial):
+                x = trial.suggest_float("x", -100, 100)
+                y = trial.suggest_categorical("y", [-1, 0, 1])
+                f1 = x**2 + y
+                f2 = -((x - 2) ** 2 + y)
+                return f1, f2
+
+
+            # We minimize the first objective and the second objective.
```
Alnusjaponica, line 46: I found one more.
```suggestion
            # We minimize the first objective and maximize the second objective.
```

nabenabe0928:
> In my understanding, `multi_objective` module has already been removed from docs by #4167 and wonder if it is really necessary to add notes to `optuna/multi_objective/samplers/_motpe.py`. I also made some comments. PTAL.

As users may still look at the doc-string of `MOTPESampler` via VSCode or any kind of IDEs, I will leave the note about the `MOTPESampler` deprecation in the corresponding parts.

Alnusjaponica (approved):
LGTM. 
@knshnb Could you take another look?

optuna/multi_objective/samplers/_motpe.py
```diff
@@ -26,9 +26,33 @@
 class MOTPEMultiObjectiveSampler(BaseMultiObjectiveSampler):
     """Multi-objective sampler using the MOTPE algorithm.
 
-    This sampler is a multiobjective version of :class:`~optuna.samplers.TPESampler`.
+    This sampler is a multi-objective version of :class:`~optuna.samplers.TPESampler`.
 
-    For further information about MOTPE algorithm, please refer to the following paper:
+    .. note::
+        For `v2.9.0 <https://github.com/optuna/optuna/releases/tag/v2.9.0>`_ or later,
```
knshnb, line 32: It seems v2.4.0?
```suggestion
        For `v2.4.0 <https://github.com/optuna/optuna/releases/tag/v2.4.0>`_ or later,
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -94,6 +98,30 @@ def objective(trial):
             study = optuna.create_study(sampler=TPESampler())
             study.optimize(objective, n_trials=10)
 
+    .. note::
+        For `v2.9.0 <https://github.com/optuna/optuna/releases/tag/v2.9.0>`_ or later,
+        :class:`~optuna.samplers.MOTPESampler` is deprecated and
+        :class:`~optuna.samplers.TPESampler` should be used instead. The following code shows
```
knshnb, line 104: I feel including deprecation information of `MOTPESampler` in the doc of `TPESampler` is not necessary and rather misleading, but what do you think?

knshnb (commented):
Thanks for the PR. This could definitely improve the document! I left two comments.

optuna/samplers/_tpe/sampler.py
nabenabe0928: The background of this PR was actually when I looked for the usage of MOTPE, `MOTPESampler` showed up on the google search and I was wondering how I should use `multivariate` there.

I was wondering if it has not just been supported yet because `TPESampler`, which has the `multivariate` feature, does not mention anything about how it should be applied to multi-objective optimization.

So I thought that it would be nice to have a note saying "you can use this sampler for multi-objective optimization in this way!" ALTHOUGH I now agree that it is not necessary to have this note given the fact that we deprecated `optuna/multi_objective` from the documentation.

nabenabe0928: I changed here accordingly.

knshnb (approved):
Thanks for the update. LGTM.

Issue #4958, opened 2023-09-29
Support `TerminatorCallback` in `OptunaSearchCV`


Alnusjaponica:
### Motivation

OptunaSearchCV performs cross-validation, implying that it should theoretically be feasible to employ the TerminatorCallback with it.

### Description

Currently, it is possible to pass `TerminatorCallback` though `calback` argument of `OptunaSearchCV`. Therefore it suffice to properly report cv scores within the `fit()` function, i.e., call `report_cross_validation_scores()` after every time cv scores are obtained.

adjeiv:
I'd like to work on this issue!

Alnusjaponica:
This issue was resolved by #5098. Thanks @adjeiv

Alnusjaponica:
This issue was resolved by https://github.com/optuna/optuna/pull/5098. Thanks @adjeiv

PR #5098, opened 2023-11-06
Report CV scores from within OptunaSearchCV

adjeiv:
## Motivation
Refs #4958 

## Description of the changes
Call report_cross_validation_scores from within OptunaSearchCV, so that if we pass in a terminator callback (e.g. using CrossValidationErrorEvaluator), we can use the scores to determine whether to terminate the study.

Tested with the following adapted sample code:

```
import optuna
from sklearn.datasets import load_iris
from sklearn.svm import SVC
from optuna.terminator import TerminatorCallback, Terminator, CrossValidationErrorEvaluator

clf = SVC(gamma="auto")
param_distributions = {
    "C": optuna.distributions.FloatDistribution(1e-10, 1e10, log=True)
}
terminator = TerminatorCallback(terminator=Terminator(error_evaluator=CrossValidationErrorEvaluator(), min_n_trials=4))
optuna_search = optuna.integration.OptunaSearchCV(clf, param_distributions, callbacks=[terminator])
X, y = load_iris(return_X_y=True)
optuna_search.fit(X, y)
```

### DIFF
--- a/optuna/integration/sklearn.py
+++ b/optuna/integration/sklearn.py
@@ -24,6 +24,7 @@ from optuna._experimental import experimental_class
 from optuna._imports import try_import
 from optuna.distributions import _convert_old_distribution_to_new_distribution
 from optuna.study import StudyDirection
+from optuna.terminator import report_cross_validation_scores
 from optuna.trial import FrozenTrial
 from optuna.trial import Trial
 
@@ -242,6 +243,7 @@ class _Objective:
                 }
 
         self._store_scores(trial, scores)
+        report_cross_validation_scores(trial, scores["test_score"])
 
         return trial.user_attrs["mean_test_score"]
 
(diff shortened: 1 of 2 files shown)


HideakiImamura:
@Alnusjaponica @nabenabe0928 Could you review this PR?

nabenabe0928 (approved):
Thank you for the PR, LGTM!

optuna/integration/sklearn.py
```diff
@@ -243,6 +244,10 @@ def __call__(self, trial: Trial) -> float:
 
         self._store_scores(trial, scores)
 
+        test_scores = scores["test_score"]
```
Alnusjaponica, line 247: The `cross_validate()` function does not necessarily return a dict with "test_score" key, but it is only when multiple metrics are given though `scoring` parameter, which does not happen in this case.

Alnusjaponica (approved):
LGTM. We really appreciate your contribution.

Alnusjaponica:
Let me merge this PR after I confirm that `windows-tests` passes.

PR #5103, opened 2023-11-10
FileStorage: enable loading of read-only files

Guillaume227:
## Motivation

In FileStorage initialization, current logic attempts to create a file (in append mode) even if it exists already. When trying to access a read-only file that results in a permission error even though no modification of the file is intended.
I came upon that problem when trying to load/display an existing study to which I had only read-only access.

## Description of the changes

If the file exists, do not attempt to create it.


c-bata:

gen740 (approved):
LGTM!

PR #5105, opened 2023-11-11
Fix type annotation

Alnusjaponica:
## Motivation
Correct type annotation of initializer of `OptunaSearchCV`.

## Description of the changes
The parameter `n_traials` accepts `None` as documented, but annotated as if it only does `int`. This PR
- Fixes type annotation of `n_traials` properly and 
- Updates type annotations as mentioned by https://github.com/optuna/optuna/issues/4508 while I'm at it


not522 (approved):
LGTM!

PR #5117, opened 2023-11-16
Refactor TPESampler for more clarity before c-TPE integration

nabenabe0928:
## Motivation

This PR aims for:
1. making the processing of each part clearer,
2. making the sample method more concise, and
3. making it easier to integrate [c-TPE](https://arxiv.org/abs/2211.14411) afterward.

Note that this PR does not intend to break any behaviors. In other words, this commit should not change any outputs.

## Description of the changes

In this PR, I made the following changes:
1. Separate the process for building a Parzen estimator as a method,
2. Create a method to compute the acquisition function values,
3. Make the role of `_compare` clearer by changing its arguments from log-likelihoods to the acquisition function values.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -538,7 +538,7 @@ class TPESampler(BaseSampler):
 
         if sample_size != acq_fn_vals.size:
             raise ValueError(
-                "The sizes of `samples` and `acq_fn_vals` must be same , but got "
+                "The sizes of `samples` and `acq_fn_vals` must be same, but got "
                 f"(samples.size, acq_fn_vals.size) = ({sample_size}, {acq_fn_vals.size})."
             )
 
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -7,7 +7,6 @@ from typing import cast
 from typing import Dict
 from typing import Optional
 from typing import Sequence
-from typing import Union
 import warnings
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


c-bata:
@knshnb @HideakiImamura Could you review this PR?

HideakiImamura (commented):
Thanks for the PR. I think this PR itself is not related to c-TPE. Is this correct?

In addition, I think it would be great to make the title of this PR more concrete.

optuna/samplers/_tpe/sampler.py
```diff
@@ -477,59 +476,75 @@ def _sample(
...
 
-        below = self._get_internal_repr(below_trials, search_space)
-        above = self._get_internal_repr(above_trials, search_space)
+        mpe_below = self._build_parzen_estimator(
+            study, search_space, below_trials, handle_below=True
+        )
+        mpe_above = self._build_parzen_estimator(study, search_space, above_trials)
+
+        samples_below = mpe_below.sample(self._rng.rng, self._n_ei_candidates)
+        acq_fn_vals = self._compute_acquisition_function(samples_below, mpe_below, mpe_above)
```
HideakiImamura, line 485: The variable name `acq_fn_vals` is not clear. How about `acquisition_function_values`?

optuna/samplers/_tpe/sampler.py
```diff
@@ -477,59 +476,75 @@ def _sample(
             self._constraints_func is not None,
         )
 
-        below = self._get_internal_repr(below_trials, search_space)
-        above = self._get_internal_repr(above_trials, search_space)
+        mpe_below = self._build_parzen_estimator(
```
HideakiImamura, line 479: Is there a need to divide the part that builds the Parzen estimator into a function? Simply increasing the depth of function calling will, in a sense, increase the complexity of the code. The function _build_parzen_estimator does not seem to obscure processes too complex to separate. Personally, it seems that the existence of the handle_below argument makes this method confusing.

HideakiImamura (commented):
Thanks for the PR. I have two comments.

optuna/samplers/_tpe/sampler.py
```diff
@@ -477,59 +476,75 @@ def _sample(
...
+        return ret
 
-        # We then sample by maximizing log likelihood ratio.
-        if study._is_multi_objective():
+    def _build_parzen_estimator(
+        self,
+        study: Study,
+        search_space: dict[str, BaseDistribution],
+        trials: list[FrozenTrial],
+        handle_below: bool = False,
```
knshnb, line 498: I feel using the default argument, which could cause bugs, here is unnecessary. How about avoiding removing the default argument?

optuna/samplers/_tpe/sampler.py
```diff
@@ -477,59 +476,75 @@ def _sample(
...
+    def _compute_acquisition_function(
+        self,
+        samples: dict[str, np.ndarray],
+        mpe_below: _ParzenEstimator,
+        mpe_above: _ParzenEstimator,
+    ) -> np.ndarray:
+        log_likelihoods_below = mpe_below.log_pdf(samples)
+        log_likelihoods_above = mpe_above.log_pdf(samples)
+        acq_fn_vals = log_likelihoods_below - log_likelihoods_above
+        return acq_fn_vals
```
knshnb, line 527: A nit.
```suggestion
        return log_likelihoods_below - log_likelihoods_above
```

knshnb:
Sorry for the delayed response. Almost LGTM. What do you think about these comments? https://github.com/optuna/optuna/pull/5117#discussion_r1409363922
https://github.com/optuna/optuna/pull/5117#discussion_r1416644266

nabenabe0928:
@HideakiImamura @knshnb 
I added a comment [here](https://github.com/optuna/optuna/pull/5117#discussion_r1409363922).
I thought that the comment was already sent out, but it was still pending when I looked at it today.

optuna/samplers/_tpe/sampler.py
knshnb: test

knshnb:
Oh, I think you need to push "Finish your review" button to make the comments visible to others.

knshnb (approved):
LGTM!

HideakiImamura (approved):
LGTM.

PR #5123, opened 2023-11-18
Pin mypy version to 1.6.*

Alnusjaponica:
## Motivation
Temporary fix `Checks(integration)` by adding version specification on `mypy`

## Description of the changes
It appears that `mypy v1.7.0` is causing significant regressions when used in conjunction with `pybind11`, upon which mypy's type inference for torch modules depends. This issue is highlighted in https://github.com/python/mypy/issues/16486. 
This pull request addresses the problem by setting the `mypy` version to `v1.6.*` until the associated bugs are resolved, as `Pytorch` does [here](https://github.com/pytorch/pytorch/blob/fb3bc3949a64fd976f84806077bcc36167618505/.ci/docker/requirements-ci.txt#L78-L79).

## Alternatives 
Add `type:ignore`s and wait for the `warn-unused-ignores` option to detect that the bug has been resolved.

### DIFF
--- a/.github/workflows/checks-integration.yml
+++ b/.github/workflows/checks-integration.yml
@@ -29,7 +29,7 @@ jobs:
         sudo apt-get update
         sudo apt-get -y install openmpi-bin libopenmpi-dev libopenblas-dev
 
-    # TODO(Shinichi): Remove the version constraint on mypy
+    # TODO(Shinichi): Remove the version constraint on mypy.
     - name: Install
       run: |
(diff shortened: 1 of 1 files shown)


not522 (approved):
LGTM!

Alnusjaponica:
@HideakiImamura I resolved the conflict and revised the comment a little. Please take anther look.

HideakiImamura (approved):
LGTM.

PR #5137, opened 2023-12-03
Clarify that `Recommended budgets` include `n_startup_trials`

### LINKED ISSUE #4989: BoTorchSampler `n_startup_trials` and minimum recommended budget
In the Optuna [sampler comparation table](https://optuna.readthedocs.io/en/stable/reference/samplers/index.html) it states that `BoTorchSampler` recommended budget is between 10 and 100 trials, however in the `BoTorchSampler` [documentation page](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.BoTorchSampler.html) the `n_startup_trials` parameter is set to `10`. This means that if someone with a very expensive optimization problem uses BoTorch with only 10 trials, it is equivalent to performing a random search.

Should the default parameter `n_startup_trials` be changed to zero? What is the recommended value?

nzw0301:
## Motivation

Close #4989.

## Description of the changes

Add a sentence in the note `Recommended budgets`.


contramundum53 (approved):
LGTM!

PR #5135, opened 2023-12-01
Pin the version of PyQt6-Qt6

not522:
## Motivation
The windows-tests CI has been failed because of the PyQt6-Qt6 v6.6.1 release. (The same problem was reported in matplotlib. https://github.com/matplotlib/matplotlib/pull/27412)

## Description of the changes
- Pin the version of PyQt6-Qt6 below 6.6.1

PR #5140, opened 2023-12-06
Revert "Pin the version of PyQt6-Qt6"

not522:
## Motivation
The `ImportError` should be fixed since PyQt v6.6.1 has been released.

## Description of the changes
Reverts optuna/optuna#5135


contramundum53 (approved):
The test works now. Thanks.

PR #5154, opened 2023-12-20
Bundle the common processes for importance evaluators

nabenabe0928:
* Add target_values transformation to before_evaluate
* Separte before/after_evaluate as functions
* Remove after_evaluate

## Motivation

As some processes for importance evaluators are in common, I bundled them as a function so that we do not have to repeat the same process in each evaluator.

## Description of the changes

I simply bundled the common processes and replaced the current hard-coded parts with the function call.

### DIFF
--- a/optuna/importance/_base.py
+++ b/optuna/importance/_base.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import abc
 from typing import Callable
 from typing import cast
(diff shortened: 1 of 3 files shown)


nabenabe0928:

nabenabe0928:
@c-bata I removed your assignee and I would like to ask @contramundum53 instead.
Could @contramundum53 review this PR?

contramundum53 (commented):
It is easily confirmed that the code fix doesn't change semantics.
I'm a little concerned that the new function `before_evaluate` actually does a bunch of things. This may obfuscate the code (along with somewhat opaque function name), and future fixes in the function may cause unexpected effects in different call sites.

nabenabe0928:
I will close the PR for now as we can come back here whenever we would like to fix this issue again!

Issue #5089, opened 2023-11-02
Unify the implementations of non-dominated sort


Alnusjaponica:
### Motivation

Currently, the `Optuna` codebase has implementations of fast non-dominated sort in three different places:
- `optuna.study._multi_objective._get_pareto_front_trials_2d()` and `optuna.study._multi_objective._get_pareto_front_trials_nd()`
- `optuna.samplers._tpe.sampler._calculate_nondomination_rank()`
- `optuna.samplers.nsgaii._elite_population_selection_strategy._fast_non_dominated_sort()`

This issue aims to unify those implementation to make the codebase simpler.

### Suggestion
One of the possible steps are
1. Unify `_fast_non_dominated_sort()` and `_calculate_nondomination_rank()` into `optuna.study._multi_objective._fast_non_dominated_sort()`. The new function will
   -  take either of a `trial` list or a `np.ndarray`,
   - use numpy matrix operation to compare each trials,
   -  equip with early termination when top $k$ individuals (for TPE) or enough ranks which includes $k$ individuals (for NSGA-II) are obtained,
   - and acccept `_constrained_dominates`.
2. (Optional) Merge `_get_pareto_front_trials_nd()` into `optuna.study._multi_objective._fast_non_dominated_sort()`. This might makes pareto front calculation slower, so it will be necessary to run some benchmarks.
3. (Optional) It is also possible to accelerate `_get_pareto_front_trials_nd()` by using  [Kung's algorithm](https://dl.acm.org/doi/pdf/10.1145/321906.321910) or eliminating unnecessary comparison between new item and dominated items. Note that this will also needs to confirm benchmarks before merging.
4. Resolve https://github.com/optuna/optuna/pull/5160#discussion_r1491890749

### Additional context (optional)

Non dominated sorts
- https://github.com/optuna/optuna/blob/e54ae1f54979e1b916b801b8993ec075be4b9757/optuna/study/_multi_objective.py#L11-L57
- https://github.com/optuna/optuna/blob/e54ae1f54979e1b916b801b8993ec075be4b9757/optuna/samplers/_tpe/sampler.py#L564-L576
- https://github.com/optuna/optuna/blob/e54ae1f54979e1b916b801b8993ec075be4b9757/optuna/samplers/nsgaii/_elite_population_selection_strategy.py#L112-L150

The above functions have some difference:
- `_get_pareto_front_trials_2d()` and `_get_pareto_front_trials_nd()`
[...]

PR #5157, opened 2023-12-21
Unify the implementation of `_create_frozen_trial()` under `testing` module

Alnusjaponica:
## Motivation
Resolve this todo comment and unify the implementation of the same name function.
https://github.com/optuna/optuna/blob/d430a23d36489e19c0d8306be80486b922da4807/tests/samplers_tests/test_nsgaii.py#L578

## Description of the changes
Implement `testing._create_frozen_trial()`, which behaves exactly the same as `test_cma._create_frozen_trial()` and `test_nsgaii._create_frozen_trial()` except the order of the positional arguments, and use it instead of `_create_frozen_trial()` in each test module.

PR #5160, opened 2023-12-22
Unify implementation of fast non-dominated sort

Alnusjaponica:
## Motivation
(Partially) resolve https://github.com/optuna/optuna/issues/5089

## Description of the changes

- Unify the backend of non-dominated sort in `sampler/_tpe/sampler.py` and `sampler/nsgaii/_elite_population_selection_strategy.py` into `study._multi_objective._fast_non_dominated_sort()`, which
  - (theoretically) speeds up the non dominated sort step used in `TPESampler`, and
  - speeds up fast non dominated sort step in `NSGAIISampler` and `NSGAIIISampler` by replacing for loop with ndarray operation,
- Add helper function `_constraints_evaluation._evaluate_penalty()` to handle constraints. This eliminates `_fast_non_dominated_sort()`’s dependence on `_constrained_dominates()` and prevents duplicate penalty value calculations,
- Add helper function `_rank_population()` for NSGA-II, -III. This bridges the gap between I/O type of non-dominated sort implementations,
- Update argument for `_validate_constraints()` and move constraints value validation step in `_constrained_dominates()` to `_validate_constraints()`, which reduces duplicated validation steps,
- Rename a module name from `optuna.samplers.nsgaii._dominates` to `optuna.samplers.nsgaii._constraints_evaluation`,
- Move `test_calculate_nondomination_rank()`, which now tests `study._multi_objective._fast_non_dominated_sort()` from `samplers_tests/tpe_tests/test_multi_objective_sampler.py` to `study_tests/test_multi_objective.py`.

Please note that this PR is based on https://github.com/optuna/optuna/pull/5157 and merge it first.


optuna/study/_multi_objective.py
```diff
@@ -69,6 +74,64 @@ def _get_pareto_front_trials(study: "optuna.study.Study") -> List[FrozenTrial]:
     return _get_pareto_front_trials_by_trials(study.trials, study.directions)
 
 
+def _fast_non_dominated_sort(
+    objective_values: np.ndarray,
+    *,
+    penalty: np.ndarray | None = None,
+    n_below: int | None = None,
+) -> np.ndarray:
```
nabenabe0928, line 82:
```python

import time

import numpy as np


def run_alnusjaponica(objective_values: np.ndarray) -> np.ndarray:
    domination_mat = np.all(
        objective_values[:, np.newaxis, :] >= objective_values[np.newaxis, :, :], axis=2
    ) & np.any(
        objective_values[:, np.newaxis, :] > objective_values[np.newaxis, :, :], axis=2
    )

    domination_list = np.nonzero(domination_mat)
    domination_map = defaultdict(list)
    for dominated_idx, dominating_idx in zip(*domination_list):
        domination_map[dominating_idx].append(dominated_idx)

    ranks = np.full(len(objective_values), -1)
    dominated_count = np.sum(domination_mat, axis=1)

    rank = -1
    ranked_idx_num = 0
    while ranked_idx_num < len(objective_values):
        (non_dominated_idxs,) = np.nonzero(dominated_count == 0)
        ranked_idx_num += len(non_dominated_idxs)
        rank += 1
        ranks[non_dominated_idxs] = rank

        dominated_count[non_dominated_idxs] = -1
        for non_dominated_idx in non_dominated_idxs:
            dominated_count[domination_map[non_dominated_idx]] -= 1
    return ranks


def is_pareto_front_nd(ordered_objective_values: np.ndarray) -> np.ndarray:
    loss_values = ordered_objective_values.copy()
    n_trials = loss_values.shape[0]
    on_front_indices = np.arange(n_trials)
    next_index = 0
    while next_index < len(loss_values):
        nondominated_mask = np.any(loss_values < loss_values[next_index], axis=1)
        nondominated_mask[~nondominated_mask] |= np.all(
            loss_values[~nondominated_mask] == loss_values[next_index], axis=1
        )
        nondominated_mask[next_index] = True
        # Remove dominated trials.
        on_front_indices = on_front_indices[nondominated_mask]
        loss_values = loss_values[nondominated_mask]
        next_index = np.sum(nondominated_mask[:next_index]) + 1

    mask = np.zeros(n_trials, dtype=bool)
    mask[on_front_indices] = True
    return mask


def is_pareto_front_2d(ordered_objective_values: np.ndarray) -> np.ndarray:
    n_trials = ordered_objective_values.shape[0]
    cummin_value1 = np.minimum.accumulate(ordered_objective_values[:, 1])
    is_value1_min = cummin_value1 == ordered_objective_values[:, 1]
    is_value1_new_min = cummin_value1[1:] < cummin_value1[:-1]

    on_front = np.ones(n_trials, dtype=bool)
    is_value0_same = ordered_objective_values[1:, 0] == ordered_objective_values[:-1, 0]
    on_front[1:] = is_value1_min[1:] & (is_value0_same | is_value1_new_min)
    return on_front


def is_pareto_front(ordered_objective_values: np.ndarray) -> np.ndarray:
    (n_trials, n_objectives) = ordered_objective_values.shape
    if n_objectives == 1:
        return ordered_objective_values[:, 0] == ordered_objective_values[0]
    elif n_objectives == 2:
        return is_pareto_front_2d(ordered_objective_values)
    else:
        return is_pareto_front_nd(ordered_objective_values)


def run_nabenabe(objective_values: np.ndarray) -> np.ndarray:
    (n_trials, n_objectives) = objective_values.shape
    ranks = np.zeros(n_trials, dtype=int)
    rank = 0
    indices = np.arange(n_trials)

    if n_objectives == 1:
        order = np.argsort(objective_values[:, 0])
    else:
        order = np.lexsort([objective_values[:, 1], objective_values[:, 0]])

    order_inv = np.zeros_like(order)
    order_inv[order] = np.arange(n_trials)
    ordered_objective_values = objective_values[order]
    while indices.size > 0:
        on_front = is_pareto_front(ordered_objective_values)
        ranks[indices[on_front]] = rank
        # Remove pareto front points.
        indices = indices[~on_front]
        ordered_objective_values = ordered_objective_values[~on_front]
        rank += 1

    return ranks[order_inv]


def measure_time(target, objective_values: np.ndarray) -> tuple[np.ndarray, float]:
    start = time.time()
    results = target(objective_values.copy())
    elapsed_time = (time.time() - start) * 1000
    # print(f"{elapsed_time:.2f}[ms]")
    return results, elapsed_time


if __name__ == "__main__":
    n_trials = 2000
    n_objectives = 2
    n_seeds = 5
    results = {"nabenabe": [], "alnusjaponica": []}
    for seed in range(n_seeds):
        rng = np.random.RandomState(seed)
        objective_values = rng.normal(size=(n_trials, n_objectives))
        ans_nabenabe, t = measure_time(run_nabenabe, objective_values)
        results["nabenabe"].append(t)
        ans_alnusjaponica, t = measure_time(run_alnusjaponica, objective_values)
        results["alnusjaponica"].append(t)
        print(np.all(ans_alnusjaponica == ans_nabenabe))

    print({k: f"{np.mean(v):.2f} +/- {np.std(v) / np.sqrt(n_seeds):.2f} [ms]" for k, v in results.items()})

```

For simplicity, I removed `penalty` and `n_below`, but this implementation maintains the same results by your program yet much quicker.

nabenabe0928: The unit is milliseconds.
||Mine|Yours|
|:--:|:--:|:--:|
|`n_trials=1000, n_objectives=2`|1.04 $\pm$ 0.1|84.5 $\pm$ 3.1|
|`n_trials=10000, n_objectives=2`|16.4 $\pm$ 1.4|9858.0 $\pm$ 126.8|
|`n_trials=1000, n_objectives=3`|17.54 $\pm$ 2.1|61.5 $\pm$ 1.0|
|`n_trials=10000, n_objectives=3`|469.0 $\pm$ 10.1|7025.0 $\pm$ 181.4|

nabenabe0928: I think your implementation of dominance with `penalty` includes the unnecessary consideration of `penalty` for feasible cases.
Namely, we do not consider the penalty amount once each trial satisfies the penalty, but yours are considering the penalty amount even for feasible cases.

Please check the following definition by Deb et al. [1]:

image


Plus, the implementation below is much quicker.

```python
def calculate_nondomination_rank_with_penalty(objective_values: np.ndarray, penalty: np.ndarray | None = None) -> np.ndarray:
    if penalty is None:
        return calculate_nondomination_rank(objective_values)

    # If values[i] constrained-dominates values[j] given penalty[i] and penalty[j],
    # one of the following must be satisfied:
    # 1. penalty[i] <= 0 and penalty[j] > 0,
    # 2. penalty[i] > 0 and penalty[j] > 0 and penalty[i] < penalty[j], or
    # 3. penalty[i] <= 0 and penalty[j] <= 0 and values[i] dominates values[j].
    # Therefore, if trials[i] is feasible and trials[j] is infeasible,
    # nondomination_rank[i] <= nondomination_rank[j] always holds and we can separate the sortings
    # for feasible trials and infeasible trials.
    # Ref: Definition 1 by K. Deb et al. in `A Fast and Elitist Multiobjective Genetic Algorithm: NSGA-II`
    if len(penalty) != len(objective_values):
        raise ValueError(
            f"The length of penalty and objective_values must be same, but got "
            f"len(penalty)={len(penalty)} and len(objective_values)={len(objective_values)}."
        )

    penalty[np.isnan(penalty)] = np.inf
    is_feasible = penalty <= 0
    nondomination_rank = np.zeros(len(objective_values), dtype=int)
    nondomination_rank[is_feasible] += calculate_nondomination_rank(objective_values[is_feasible])
    nondomination_rank[~is_feasible] += np.max(nondomination_rank[is_feasible], initial=-1) + 1
    nondomination_rank[~is_feasible] += calculate_nondomination_rank(
        penalty[~is_feasible, np.newaxis],
    )
    return nondomination_rank
```

[1] K. Deb et al. A Fast and Elitist Multiobjective Genetic Algorithm: NSGA-II

tests/study_tests/test_multi_objective.py
```diff
@@ -110,3 +112,41 @@ def test_dominates_complete_vs_incomplete(t1_state: TrialState) -> None:
 
     assert _dominates(t2, t1, list(directions))
     assert not _dominates(t1, t2, list(directions))
+
+
+def test_calculate_nondomination_rank() -> None:
```
nabenabe0928, line 117: Tests for duplicated trials are missing, so I think it is better adding them.
```python
test_case = np.array([[1, 1], [1, 1], [1, 2], [2, 1], [0, 1.5], [1.5, 0], [0, 1.5]])
ans = [0, 0, 1, 1, 0, 0, 0]

test_case = np.array([[1, 1], [1, 1], [1, 2], [2, 1], [1, 1], [0, 1.5], [0, 1.5]])
ans = [0, 0, 1, 1, 0, 0, 0]
```

Also, could you add something similar for the case of `n_objectives=3`?

HideakiImamura:
@Alnusjaponica It seems to me that if we include the current PR, it would not be in the form of passing the dominates function to the function that does the fast non-dominated sort. What do you think about this?

optuna/study/_multi_objective.py
Alnusjaponica: > this implementation maintains the same results as your program yet much quicker.

Thanks for the suggestion. I just move what's implemented in `_tpe/sampler.py` and no need to stick to the first implementation. I'll consider how to reconcile current `_get_pareto_front_trials_by_trials` function with your suggestion.
> I think your implementation of dominance with penalty includes the unnecessary consideration of penalty for feasible cases.

In my understanding, the penalty is set to 0 when the trial is feasible, thus having no influence on the result. Anyway, I'll use the faster implementation in the suggestion.

nabenabe0928:
Just for the future comment, I found a Python implementation of non-dominated sort algorithm with the time complexity of $O(N \log^{M - 1} N)$.
https://github.com/DEAP/deap/blob/master/deap/tools/emo.py

optuna/study/_multi_objective.py
nabenabe0928:
```python

import itertools

import numpy as np


def is_pareto_front(trials: list[Trial]) -> np.ndarray:
    n_trials = len(trials)
    is_front = np.zeros(n_trials, dtype=bool)
    next_index = 0
    nondominated_indices = np.arange(n_trials)
    while next_index < len(trials):
        nondominated_mask = np.array([not dominates(t, trials[next_index]) for t in trials])
        trials = list(itertools.compress(trials, nondominated_mask))
        nondominated_indices = nondominated_indices[nondominated_mask]
        next_index = np.sum(nondominated_mask[:next_index]) + 1

    is_front[nondominated_indices] = True
    return is_front


def calculate_nondomination_rank(trials: list[Trial]) -> np.ndarray:
    n_trials = len(trials)
    ranks = np.zeros(n_trials, dtype=int)
    rank = 0
    indices = np.arange(n_trials)
    while indices.size > 0:
        on_front = is_pareto_front(trials)
        ranks[indices[on_front]] = rank
        indices = indices[~on_front]
        trials = list(itertools.compress(trials, ~on_front))
        rank += 1

    return ranks
```

This is Just a memo for future discussion, please ignore it for now.

When using `dominates`, the runtime will be much much longer compared to the vectorization version, but still it runs quicker than creating the dominance matrix.

optuna/samplers/nsgaii/_constraints_evaluation.py
```diff
@@ -86,15 +85,45 @@ def _constrained_dominates(
...
+        feasible/infeasible and None means that the trial does not have constraint values.
+    """
+
+    penalty: list[float] = []
+    for trial in population:
+        constraints = trial.system_attrs.get(_CONSTRAINTS_KEY)
+        if constraints is None:
+            penalty.append(np.nan)
+        else:
+            assert isinstance(constraints, (list, tuple))
```
HideakiImamura, line 101: The type of the constraints is `Sequence[float]` by [definition,](https://github.com/optuna/optuna/blob/master/optuna/samplers/nsgaii/_sampler.py#L142) which includes not only list and tuple but also range and concrete classes inheriting from `collections.abc.Sequence`. Can we avoid this assertion?

optuna/samplers/nsgaii/_constraints_evaluation.py
```diff
@@ -86,15 +85,45 @@ def _constrained_dominates(
...
 def _validate_constraints(
     population: list[FrozenTrial],
-    constraints_func: Callable[[FrozenTrial], Sequence[float]] | None = None,
+    *,
+    is_constrained: bool = False,
 ) -> None:
-    if constraints_func is None:
+    if not is_constrained:
         return
+    assert len(population) > 0
```
HideakiImamura, line 113: Where does this assertion come from?

optuna/samplers/nsgaii/_constraints_evaluation.py
```diff
@@ -86,15 +85,45 @@ def _constrained_dominates(
...
+                f"Trial {_trial.number} does not have constraint values."
+                " It will be dominated by the other trials."
+            )
             continue
+        # Initialize num_constraints with the number of constraints of the first trial with values.
+        num_constraints = len(_constraints) if num_constraints is None else num_constraints
         if np.any(np.isnan(np.array(_constraints))):
             raise ValueError("NaN is not acceptable as constraint value.")
+        elif len(_constraints) != num_constraints:
+            raise ValueError("Trials with different numbers of constraints cannot be compared.")
```
HideakiImamura, line 129: It would be great to add an test to cover this line as suggested by codecov.

HideakiImamura (commented):
Basically, LGTM except several minor comments. PTAL.

optuna/samplers/nsgaii/_constraints_evaluation.py
```diff
@@ -86,15 +85,45 @@ def _constrained_dominates(
...
-    constraints_func: Callable[[FrozenTrial], Sequence[float]] | None = None,
+    *,
+    is_constrained: bool = False,
 ) -> None:
-    if constraints_func is None:
+    if not is_constrained:
         return
+    assert len(population) > 0
+
+    num_constraints = None
```
nabenabe0928, line 115: Can we compute it in advance?

```suggestion
    num_constraints = max([t.system_attrs.get(_CONSTRAINTS_KEY, []) for t in population], default=0)
```

optuna/samplers/nsgaii/_elite_population_selection_strategy.py
```diff
@@ -109,42 +112,30 @@ def _crowding_distance_sort(population: list[FrozenTrial]) -> None:
...
+        return []
+
+    objective_values = np.array(
+        [
+            trial.values if trial.values else [float("inf")] * len(directions)
+            for trial in population
+        ],
+        dtype=np.float64,
+    )
+    objective_values *= np.array(
```
nabenabe0928, line 131: Can we use `loss_values` to be clear to us in the future that each objective is better when it is lower in this array?
Please check out other functions as well.

optuna/samplers/nsgaii/_constraints_evaluation.py
Alnusjaponica: This assertion came from here:
https://github.com/optuna/optuna/blob/53d85542ae8e291101e929ceb4bbbdb146609613/optuna/samplers/nsgaii/_dominates.py#L56-L57
but seems unnecessary. I'll remove it.

optuna/samplers/nsgaii/_constraints_evaluation.py
Alnusjaponica: It seems unnecessary. Removed.

optuna/samplers/nsgaii/_constraints_evaluation.py
Alnusjaponica: Thanks for the suggestion. I added a test.

Alnusjaponica:
As we discussed offline, we will not include speeding up changes and will instead implement them in succeeding PRs. Now that I have fixed almost all the comments, I would appreciate it if you could take another look.

By the way, since this PR has become quite large, should I split some independent changes? For example, adding test cases and updating `_validate_constraints` can be separated from this PR.

optuna/samplers/_nsgaiii/_elite_population_selection_strategy.py
```diff
@@ -52,10 +50,13 @@ def __call__(self, study: Study, population: list[FrozenTrial]) -> list[FrozenTr
         Returns:
             A list of trials that are selected as elite population.
         """
-        _validate_constraints(population, self._constraints_func)
+        _validate_constraints(population, is_constrained=self._constraints_func is not None)
+        population_per_rank = _rank_population(
+            population,
+            study.directions,
+            is_constrained=self._constraints_func is not None,
```
nabenabe0928, line 57:
```suggestion
            population, study.directions, is_constrained=self._constraints_func is not None
```

Alnusjaponica:
Regarding reconciliation with what dominates, I concluded that it would be more efficient to focus on expediting changes, as it ultimately alters the design.

optuna/study/_multi_objective.py
```diff
@@ -69,6 +74,89 @@ def _get_pareto_front_trials(study: "optuna.study.Study") -> List[FrozenTrial]:
...
+    is_infeasible = np.logical_and(~np.isnan(penalty), penalty > 0)
+    ranks, infeasible_bottom_rank = _calculate_nondomination_rank(
+        penalty[is_infeasible, np.newaxis],
+    )
+    nondomination_rank[is_infeasible] += ranks + (feasible_bottom_rank + 1)
+
+    # Third, we calculate the domination rank for trials with no penalty information.
+    is_nan = np.isnan(penalty)
+    ranks, _ = _calculate_nondomination_rank(objective_values[is_nan], n_below=n_below)
+    nondomination_rank[is_nan] += ranks + (feasible_bottom_rank + 1) + (infeasible_bottom_rank + 1)
```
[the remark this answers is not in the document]
Alnusjaponica: > The code could be re-written as below if we consider penalty is nan equivalent, which is how we should handle infeasible solutions according to what we can see from the definition of constrained dominance.

In the previous implementation, trials with nan constraints are actually compared separately, and your suggestion alters the behavior. Please refer to [this PR](https://github.com/Alnusjaponica/optuna/pull/30) for the outcome. Therefore, we have to discuss wether we should change it or not on a different PR or issue.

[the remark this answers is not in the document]
Alnusjaponica: > Btw, what happens to the trials that have nondomination_rank of -1, which is obtained for the trials that were not ranked during _calculate_nondomination_rank?

Thanks for the comment. Although the tests pass, `nondomination_rank` was not properly updated. I correct the implementation and properly ignore rank -1 trials in NSGA-X samplers.

Alnusjaponica:
@HideakiImamura @nabenabe0928 
I resolved all the review comments. Could you take another look?

optuna/study/_multi_objective.py
```diff
@@ -69,6 +74,94 @@ def _get_pareto_front_trials(study: "optuna.study.Study") -> List[FrozenTrial]:
...
+    *,
+    penalty: np.ndarray | None = None,
+    n_below: int | None = None,
+) -> np.ndarray:
+    if penalty is None:
+        ranks, _ = _calculate_nondomination_rank(objective_values, n_below=n_below)
+        return ranks
+
+    if len(penalty) != len(objective_values):
+        raise ValueError(
```
HideakiImamura, line 88: Could you add the test case for this line?

optuna/study/_multi_objective.py
```diff
@@ -69,6 +74,94 @@ def _get_pareto_front_trials(study: "optuna.study.Study") -> List[FrozenTrial]:
     return _get_pareto_front_trials_by_trials(study.trials, study.directions)
 
 
+def _fast_non_dominated_sort(
+    objective_values: np.ndarray,
+    *,
+    penalty: np.ndarray | None = None,
+    n_below: int | None = None,
+) -> np.ndarray:
```
HideakiImamura, line 82: It would be great to leave the shape and its meaning of the returned ndarray. Could you add the explanation?

optuna/study/_multi_objective.py
```diff
@@ -69,6 +74,94 @@ def _get_pareto_front_trials(study: "optuna.study.Study") -> List[FrozenTrial]:
...
+    n_below: int | None = None,
+    base_rank: int = 0,
+) -> tuple[np.ndarray, int]:
+    # Normalize n_below.
+    n_below = n_below or len(objective_values)
+    n_below = min(n_below, len(objective_values))
+
+    # The ndarray `domination_matrix` is a boolean 2d matrix where
+    # `domination_matrix[i, j] == True` means that the j-th trial dominates the i-th trial in the
+    # given multi objective minimization problem.
```
HideakiImamura, line 135: The name of variable is inconsistent.
```suggestion
    # The ndarray `domination_mat` is a boolean 2d matrix where
    # `domination_mat[i, j] == True` means that the j-th trial dominates the i-th trial in the
    # given multi objective minimization problem.
```

HideakiImamura (commented):
Thanks for the update. Almost, LGTM. I have several minor comments. PTAL.

nabenabe0928:
Followup
https://github.com/optuna/optuna/pull/5160/files#r1435351380

optuna/samplers/nsgaii/_elite_population_selection_strategy.py
nabenabe0928: Just to give @HideakiImamura the context:
`loss_vals` (or another name is `lvals`) is often used in `TPESampler` (please check [here](https://github.com/optuna/optuna/blob/master/optuna/samplers/_tpe/sampler.py#L610-L622) and [here](https://github.com/optuna/optuna/blob/master/optuna/samplers/_tpe/sampler.py#L687-L735))

So the problem here is that I would need to take a look at the source of the function calls whether `objective_values` is always better when each objective is lower.
loss does not necessarily mean the machine-learning loss functions, but just loss, whichwe already have a universal consensus such that it is better when loss is lower even in normal conversations.

I do not mind using `objective_values`, but I strongly encourage you to specify whether each objective is better when it is lower.
Again, the reason is simple because I would need to refer to the function call origin to see if `objective_values` is always better when it is lower.

optuna/study/_multi_objective.py
```diff
@@ -91,26 +91,31 @@ def _fast_non_dominated_sort(
                 len(penalty), len(objective_values)
             )
         )
-    nondomination_rank = np.zeros(len(objective_values), dtype=int)
+    nondomination_rank = np.full(len(objective_values), -1)
```
nabenabe0928, line 94: As we can already bound `max(nondomination_rank)` by `n_below` and `nondomination_rank` of `n_below + 1` will not be used, so what about using `n_below + 1`?
Another reason why we should probably avoid `-1` is that it might cause unexpected bugs in the future when some developers use `nondomination_rank` being always better when it is lower.
Plus, this implementation requires an ad-hoc handling of `nondomination_rank=-1` in each place where the function is used.

optuna/study/_multi_objective.py
```diff
@@ -91,26 +91,31 @@ def _fast_non_dominated_sort(
...
     is_nan = np.isnan(penalty)
+    n_below = n_below or len(objective_values)
 
     # First, we calculate the domination rank for feasible trials.
     is_feasible = np.logical_and(~is_nan, penalty <= 0)
-    ranks, feasible_bottom_rank = _calculate_nondomination_rank(
+    ranks, bottom_rank = _calculate_nondomination_rank(
```
nabenabe0928, line 100: If we define `nondomination_rank` as:
```python
nondomination_rank = np.full(len(objective_values), n_below + 1)
```

`bottom_rank` becomes `bottom_rank = np.max(ranks)`.
Note that if `np.max(bottom_rank) = n_below + 1`, the processes hereafter simply define each `nondomination_rank` as `n_below + <positive_integer>`, so they will be ignored.

optuna/study/_multi_objective.py
```diff
@@ -91,26 +91,31 @@ def _fast_non_dominated_sort(
...
         objective_values[is_feasible], n_below=n_below
     )
-    nondomination_rank[is_feasible] += ranks
+    nondomination_rank[is_feasible] += 1 + ranks
```
nabenabe0928, line 103: Do we start the rank from 1 now?
Probably, it should be documented somewhere.

optuna/study/_multi_objective.py
```diff
@@ -119,12 +124,15 @@ def _calculate_nondomination_rank(
     objective_values: np.ndarray,
     *,
     n_below: int | None = None,
+    base_rank: int = 0,
 ) -> tuple[np.ndarray, int]:
-    # Calculate the domination matrix.
-    # The resulting matrix `domination_matrix` is a boolean matrix where
+    # Normalize n_below.
+    n_below = n_below or len(objective_values)
+    n_below = min(n_below, len(objective_values))
```
nabenabe0928, line 131: if `n_below` is non-positive, we can already return the results using `base_rank`.

optuna/study/_multi_objective.py
```diff
@@ -91,26 +91,31 @@ def _fast_non_dominated_sort(
...
+    nondomination_rank[is_feasible] += 1 + ranks
+    n_below -= np.count_nonzero(is_feasible)
 
     # Second, we calculate the domination rank for infeasible trials.
     is_infeasible = np.logical_and(~is_nan, penalty > 0)
-    ranks, infeasible_bottom_rank = _calculate_nondomination_rank(
-        penalty[is_infeasible, np.newaxis], n_below=n_below
+    ranks, bottom_rank = _calculate_nondomination_rank(
+        penalty[is_infeasible, np.newaxis], n_below=n_below, base_rank=bottom_rank + 1
     )
```
nabenabe0928, line 110: For 1d array, `np.unique` is better.

```suggestion
    _, ranks = np.unique(penalty[is_infeasible], return_inverse=True)
    ranks += bottom_rank + 2
    bottom_rank = np.max(ranks)
```

optuna/study/_multi_objective.py
```diff
@@ -91,26 +91,31 @@ def _fast_non_dominated_sort(
                 len(penalty), len(objective_values)
             )
         )
-    nondomination_rank = np.zeros(len(objective_values), dtype=int)
+    nondomination_rank = np.full(len(objective_values), -1)
     is_nan = np.isnan(penalty)
```
nabenabe0928, line 95:
```suggestion
    is_penalty_nan = np.isnan(penalty)
```

optuna/study/_multi_objective.py
Alnusjaponica: Thanks for your comment. I added a test.

optuna/study/_multi_objective.py
Alnusjaponica: Thanks for your comment. I added an eraly return black.

optuna/study/_multi_objective.py
Alnusjaponica: I totally agree what you say but it makes this PR even larger. Can I split the task as a follow-up and resolve your comment in another PR?

optuna/study/_multi_objective.py
Alnusjaponica: I replaced `_calculate_nondomination_rank` with `np.unique`.

optuna/study/_multi_objective.py
Alnusjaponica: I added a description of current implementation to the docstrings.

Alnusjaponica:
Thanks for your review. I've addressed nearly all comments, so please take another look.
As for whether the rank should be 0-indexed or not, I believe it's better to discuss in another PR. Do you have any opinions on this?

optuna/study/_multi_objective.py
nabenabe0928: Sure, then could you please add an inline comment for this?

Alnusjaponica: The suggestion is a little bit complicated, so I remarked the comment on https://github.com/optuna/optuna/issues/5089

nabenabe0928:
Followup
- [ ] https://github.com/optuna/optuna/pull/5160/files#r1435351380
- [ ] https://github.com/optuna/optuna/pull/5160#discussion_r1491890749
- [ ] https://github.com/optuna/optuna/pull/5160#issuecomment-1958851517
- [ ] https://github.com/optuna/optuna/pull/5160/files/4bfe8714ead38f3a1e232846ba044d7f39d031b9..542f0111b7b2da1fc662d486f91bf9252a163d72#r1491887538

nabenabe0928 (approved):
Thank you for your work and sorry for taking long!
I left followups, so let's try checking them in (an)other PR(s)!

Alnusjaponica:
Thanks for your reviews. I remarked follow-up comment from the issue.
@HideakiImamura Could you take another look?

HideakiImamura (approved):
LGTM.

Issue #5113, opened 2023-11-14
 No exception when passing invalid data to .enqueue_trial


qnbhd:
### Expected behavior

It is expected that some exception will be raised when passing obviously incorrect data (not matching the `Dict[str, Any]` type) to `study.enqueue_trial`. 

### Environment

- Optuna version:3.4.0
- Python version:3.9.6
- OS:macOS-13.0.1-arm64-arm-64bit

### Error messages, stack traces, or logs

```shell
[I 2023-11-14 12:09:55,858] A new study created in memory with name: no-name-7cf052d2-a932-4b47-bdae-130b5491e38b
0 5
1 4
2 2
3 0
4 0
```

### Steps to reproduce

```python
import optuna

study = optuna.create_study(direction="maximize")
study.enqueue_trial({"foo": 5})  # OK
study.enqueue_trial([{"foo": 15}]) # Exception expected
study.enqueue_trial([1, 1, 3]) # Exception expected

for _ in range(5):
    trial = study.ask()
    foo = trial.suggest_int("foo", 0, 10)
    print(trial._trial_id, foo)  # noqa
    study.tell(trial, 0.1)
```

nzw0301:
Thank for your your input. As we can see the example, `study` cannot know the parameter space defined in the objective when `enqueue_trial` is called. In addition, I do not expect `study.enqueue_trial([{"foo": 15}])` shows some warning message because Optuna supports dynamic search space, which contradicts with causing some exception.

I agree that showing some warning message for the third case `study.enqueue_trial([1, 1, 3])` seems useful.

Thus I remove `bug` label from this issue.

PR #5164, opened 2023-12-24
Raise TypeError if params is not a dict in enqueue_trial

adjeiv:
## Motivation
Refs #5113, as well as https://github.com/optuna/optuna-dashboard/issues/679

## Description of the changes
The `params` parameter of `enqueue_trial` is expected to be a dictionary, so raise a `TypeError` if this is not the case.
Test case added for coverage

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -854,6 +854,9 @@ class Study:
             hyperparameters manually.
         """
 
+        if not isinstance(params, dict):
+            raise TypeError("params must be a dictionary.")
+
         if skip_if_exists and self._should_skip_enqueue(params):
             _logger.info(f"Trial with params {params} already exists. Skipping enqueue.")
             return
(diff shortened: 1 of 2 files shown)


HideakiImamura:
In my opinion, there is no need for this kind of error handling regarding the type of a variable, because the source code or documentation clearly specifies the type of the variable. Optuna does not provide this type of error handling.

HideakiImamura:
We committers are now on vacation for the New Year's vacations. Therefore, the review may be delayed until the New Year. We apologize for any inconvenience.

not522:

not522 (approved):
Thank you for your PR! I think that the validations for the arguments are useful when the types are not trivial. For example, `tell` and `add_trial` check their arguments because users often make mistakes. This PR deserves for merge.

HideakiImamura (approved):
It is indeed confusing that the argument of enqueue_trial must be dict, so checking the type is kind. LGTM.

PR #5181, opened 2024-01-12
Wilcoxon pruner

contramundum53:
## Motivation
Currently, all pruners implemented in Optuna only assume that the intermediate values eventually converge to the final value. This is useful for applications such as machine learning, where the intermediate values can be the loss values on the learning curve.
However, there is another large area of applications in pruning, where the objective function involves evaluating a function (e.g., the performance score of a heuristic algorithm) with many inputs and aggregating (e.g., taking mean/median of) the results: such objective functions include mean performance of a heuristic algorithm, the k-fold cross validation score of a machine learning model, and even mean performance of a prompt given to a large language model.
In these cases, the evaluation score for each input can be considered independent, and pruning can be done more aggressively by using statistical tests.

We implement a pruner that uses [Wilcoxon signed-rank test](https://en.wikipedia.org/wiki/Wilcoxon_signed-rank_test) for this purpose.

## Description of the changes
Implements `WilcoxonPruner`.

This pruner performs the Wilcoxon signed-rank test between the current trial and the current best trial, and stops whenever the pruner is sure up to a given p-value that the current trial is worse than the best one.

There can be "easy" or "hard" inputs (the pruner handles correspondence of the inputs between different trials), but it is recommended to shuffle the order of inputs once before the optimization.

When you use this pruner, you must call `Trial.report(value, step)` function **for each `step = 1, 2, ..., N`** with the **past average values of evaluations** (`value=np.mean(evaluation_values[:step])`), regardless of the actual objective function. This interface is designed so that other pruners can be used interchangeably.


contramundum53:
@y0z @knshnb Could you review this PR? Please feel free to unassign yourself if you are busy!

y0z:
I think a documentation for Wilcoxon Pruner is required.
https://optuna--5181.org.readthedocs.build/en/5181/reference/pruners.html

contramundum53:
Nice catch!
I'll add reference in the docs.
For tutorial, I think I will write one in the follow-up.

optuna/pruners/_wilcoxon.py
```diff
@@ -0,0 +1,189 @@
...
+
+        diffs = step_values - best_step_values
+        diffs[np.isnan(diffs)] = 0  # inf - inf or (-inf) - (-inf)
+        if study.direction == StudyDirection.MAXIMIZE:
+            diffs *= -1
+
+        if len(diffs) == 0:
+            return False
+
+        p = ss.wilcoxon(diffs, alternative="greater", zero_method="zsplit").pvalue
```
y0z, line 188: Can we use the `alternative="less"` option instead of `diff *= -1` for maximization?
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.wilcoxon.html

y0z (commented):
I left some comments.

optuna/pruners/_wilcoxon.py
```diff
@@ -0,0 +1,189 @@
...
+        if n_startup_steps < 0:
+            raise ValueError(f"n_startup_steps must be nonnegative but got {n_startup_steps}.")
+        if not 0.0 <= p_threshold <= 1.0:
+            raise ValueError(f"p_threshold must be between 0 and 1 but got {p_threshold}.")
+
+        self._n_startup_steps = n_startup_steps
+        self._p_threshold = p_threshold
+
+    def prune(self, study: "optuna.study.Study", trial: FrozenTrial) -> bool:
+        ss = _LazyImport("scipy.stats")
```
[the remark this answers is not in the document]
contramundum53: `try_import` would import scipy regardless of whether you use `WilcoxonPruner`. This would slow down `import optuna`. `_LazyImport` does that lazily, so you import `scipy` only when you use `scipy`. (`ImportError` is emitted at that time, too.)

contramundum53:
@y0z @eukaryo I have changed the interface so that we need to report the raw evaluation value at step `i`, rather than the historical average.
@eukaryo I added a small note that we must report different values when comparing with other pruners. Do you think this is enough to prevent people from misusing?

contramundum53:
@y0z I have applied most of your comments. PTAL!

optuna/pruners/_wilcoxon.py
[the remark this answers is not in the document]
contramundum53:
```suggestion
        import scipy.stats as ss
```

eukaryo:
Added note seems to be good.

optuna/pruners/_wilcoxon.py
```diff
@@ -0,0 +1,160 @@
...
+import optuna
+from optuna._experimental import experimental_class
+from optuna.pruners import BasePruner
+from optuna.study._study_direction import StudyDirection
+from optuna.trial import FrozenTrial
+
+
+@experimental_class("3.6.0")
+class WilcoxonPruner(BasePruner):
+    """Pruner based on the Wilcoxon signed-rank test <https://en.wikipedia.org/wiki/Wilcoxon_signed-rank_test>`_.
```
y0z, line 16: This link is not rendered as intended.
image

tutorial/10_key_features/003_efficient_optimization_algorithms.py
```diff
@@ -80,6 +80,8 @@
 #
 # - Threshold pruning algorithm implemented in :class:`~optuna.pruners.ThresholdPruner`
 #
+# - A pruning algorithm based on `Wilcoxon signed-rank test <https://en.wikipedia.org/wiki/Wilcoxon_signed-rank_test>`_ implemented in :class:`~optuna.pruners.WilcoxonPruner`
```
y0z, line 83: This link is also not rendered as intended.
image

optuna/pruners/_wilcoxon.py
```diff
@@ -0,0 +1,160 @@
...
+            input_data = np.linspace(-1, 1, 100)
+
+            # It is recommended to shuffle the input data once before optimization.
+            np.random.shuffle(input_data)
+
+
+            def objective(trial):
+                s = 0.0
+                for i in range(len(input_data)):
+                    param = trial.suggest_uniform("param", -1, 1)
```
y0z, line 60: According to the following warning, please use `suggest_float` instead of `suggest_uniform`.
```
wp.py:20: FutureWarning: suggest_uniform has been deprecated in v3.0.0. This feature will be removed in v6.0.0. See https://github.com/optuna/optuna/releases/tag/v3.0.0. Use suggest_float instead. 
```

optuna/pruners/_wilcoxon.py
```diff
@@ -0,0 +1,160 @@
...
+                "WilcoxonPruner finds steps existing in the current trial "
+                "but does not exist in the best trial. "
+                "Those values are ignored."
+            )
+
+        diff_values = step_values[idx1] - best_step_values[idx2]
+        # Overwrite (inf - inf) or (-inf) - (-inf) with 0.
+        diff_values[step_values[idx1] == best_step_values[idx2]] = 0
+
+        if len(diff_values) < self._n_startup_steps:
```
y0z, line 153:
```
n_startup_steps:
            The number of steps before which no trials are pruned.
```
Based on the documentation, I feel whether `len(steps)` or `len(diff_values)` is used for this condition is ambiguous.

y0z (commented):
I leave some comments.

optuna/pruners/_wilcoxon.py
contramundum53:
```suggestion
                    param = trial.suggest_float("param", -1, 1)
```

optuna/pruners/_wilcoxon.py
```diff
@@ -0,0 +1,163 @@
...
+            raise ValueError(f"n_startup_steps must be nonnegative but got {n_startup_steps}.")
+        if not 0.0 <= p_threshold <= 1.0:
+            raise ValueError(f"p_threshold must be between 0 and 1 but got {p_threshold}.")
+
+        self._n_startup_steps = n_startup_steps
+        self._p_threshold = p_threshold
+
+    def prune(self, study: "optuna.study.Study", trial: FrozenTrial) -> bool:
+        # Import scipy in the method because it is an optional dependency.
+        import scipy.stats as ss
```
contramundum53, line 115: Use `_LazyImport`

optuna/pruners/_wilcoxon.py
```diff
@@ -0,0 +1,160 @@
...
```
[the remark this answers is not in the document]
eukaryo: Sorry, I had pending comment

optuna/pruners/_wilcoxon.py
```diff
@@ -0,0 +1,163 @@
...
+        if n_startup_steps < 0:
+            raise ValueError(f"n_startup_steps must be nonnegative but got {n_startup_steps}.")
+        if not 0.0 <= p_threshold <= 1.0:
+            raise ValueError(f"p_threshold must be between 0 and 1 but got {p_threshold}.")
+
+        self._n_startup_steps = n_startup_steps
+        self._p_threshold = p_threshold
+
+    def prune(self, study: "optuna.study.Study", trial: FrozenTrial) -> bool:
+        # Import scipy in the method because it is an optional dependency.
```
contramundum53, line 114: Assert single-objective optimization

optuna/pruners/_wilcoxon.py
```diff
@@ -0,0 +1,163 @@
...
+        # Overwrite (inf - inf) or (-inf) - (-inf) with 0.
+        diff_values[step_values[idx1] == best_step_values[idx2]] = 0
+
+        if len(diff_values) < self._n_startup_steps:
+            return False
+
+        alt = "less" if study.direction == StudyDirection.MAXIMIZE else "greater"
+
+        # We use zsplit to avoid the problem when all values are zero.
+        p = ss.wilcoxon(diff_values, alternative=alt, zero_method="zsplit").pvalue
```
contramundum53, line 162: Test when diff_value contains infinite values.

eukaryo:
Permalink could be used to cite a wikipedia's article. For example, it for the latest version is:
https://en.wikipedia.org/w/index.php?title=Wilcoxon_signed-rank_test&oldid=1195011212

y0z (approved):
LGTM

eukaryo:
LGTM.

eukaryo (approved):

Issue #4801, opened 2023-07-09
Bayesian Gaussian process sampler


david-thrower:
### Motivation

 
TPE sampler is not the most effective for Deep learning and black box situations having a mix of continuous, discrete, and categorical hyperparameters to tune. There are other packages that support Bayesian Gaussian process, but they are not set up to parallelize like Optuna.

### Description

Described feature would be a sampler that samples via Gaussian process, once n initial trials are completed via random search.

### Alternatives (optional)

There are none that I am aware of.  

### Additional context (optional)

Most samplers that are effective for black box other than Bayesian Gaussian process are not able to optimize parameters of all 3 common data types: integers, floating point numbers, and categorical (e.g. which activation to use, or which optimizer to use in the objective to optimize its weights).

contramundum53:
We have `optuna.integration.BoTorchSampler` that does Gaussian Process-based Bayesian optimization. This sampler is effective when all parameters are continuous, but we are having difficulty optimizing the acquisition function when the search space is discrete or mixed.
If you have any ideas for optimizing acquisition functions in a mixed search space, we appreciate your contribution!

contramundum53:
Actually I personally also want to have a powerful GP implementation in Optuna, and we are using GP not only in sampler but also in `optuna.terminator` and the upcoming preferential optimization (currently under development). They all currently suffer from the same situation when the search space is mixed.

contramundum53:
Now we have a `GPSampler` (Yay!) so we close this issue.

carefree0910:
> Now we have a `GPSampler` capable for all parameter types (Yay!) so we close this issue. Note that our GPSampler is not currently designed to work well in a parallelized environment (You may find that different processes are evaluating almost the same parameters). Something like `constant_liar` in TPE won't come in v3.6, but If that becomes problematic, I think we might need to work on that.

Hi! First of all, thanks for the `GPSampler`!!

And now I exactly encountered exactly the same problem - I want to use `GPSampler` with `n_worker > 1` and found that duplicated params occurred here and there. 😢

Is something like `constant_liar` in TPE on the way? Or is there any other suggestions that I can follow? (Currently I'm using TPE with `multivariate=True` and `constant_liar=True` as a fallback)

Thanks in advance!

PR #5185, opened 2024-01-15
Add GPSampler

contramundum53:
## Motivation
Gaussian process-based bayesian optimization has been a famous and standard bayesian optimization algorithm for long, yet Optuna has only supported it through `optuna.integration.BoTorchSampler`. (cf. #4801)

BoTorch is a pretty complex library, has a long list of dependency that easily conflicts with the users' environment, and its dependency is not stable across updates. This has resulted in occasional failure of our CI and has been annoying for our development.
Also, BoTorch is unnecessarily slow without batching (especially the evaluation of acquisition functions), possibly due to its abstraction. As a result, a simple optimization for 50 trials would take `BoTorchSampler` around 30 seconds to finish. Although this is not a big problem for some usages, there are cases where this becomes a problem, including CI. `BoTorchSampler` only uses the simplest features of BoTorch, so it might not be worth paying for its abstraction.
Lastly, it was not easy to add new features in GP due to BoTorch's heavy abstraction. Such features include support for mixed search spaces (with continuous _and_ discrete parameters), which is important for real-world use cases.

We intend to implement a simple GP-BO for ourselves and provide a natively supported `GPSampler`.

The current implementation status is preliminary: the acquisition function optimizer is a trivial Quasi-Monte Carlo sampler. It handles small search spaces well, but once the search space gets larger, it starts to behave poorly. In the future, we'll consider implementing a stronger optimizer, and also integrate it with `optuna.terminator`. Once it's done, we can safely move `BoTorchSampler` to `optuna_integration` repository.

## Description of the changes
* Implement a basic `GPSampler`. The current implementation uses:
    - Matern kernel with nu=2.5 (differentiable twice)
    - Automatic relevance determination (ARD) for a lengthscale of each parameter
    - Gamma prior for lengthscales, kernel scale, and noise scale
    - Log Expected Improvement (logEI) as the acquisition function
    - Quasi-Monte Carlo (QMC) sampling to optimize the acquisition function

* Update test


contramundum53:
@not522 @nabenabe0928 Could you review this PR? If you are busy, please feel free to unassign yourself!

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,155 @@
...
+            xs[:, i] = round_one_transformed_param(
+                xs[:, i], param_types[i], (bounds[i, 0], bounds[i, 1]), steps[i]
+            )
+    return xs
+
+
+def get_search_space_and_transformed_params(
+    trials: list[optuna.trial.FrozenTrial],
+    optuna_search_space: dict[str, optuna.distributions.BaseDistribution],
+) -> tuple[SearchSpace, np.ndarray]:
```
nabenabe0928, line 84: Could you check if we can use `_SearchSpaceTransform` [here](https://github.com/optuna/optuna/blob/master/optuna/_transform.py#L16)?

nabenabe0928 (commented):
I have not checked everything, but I left some comments.

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,155 @@
...
+
+import optuna
+
+
+LINEAR = 0
+LOG = 1
+CATEGORICAL = 2
+
+
+class SearchSpace(NamedTuple):
```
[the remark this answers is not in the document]
contramundum53: Sorry, it is an side effect of trying to use numba (numba only supports `NamedTuple` and not `dataclass`, cf. https://github.com/numba/numba/issues/4037). I'll change to `dataclass` for now since we don't need `numba` now.

optuna/_gp/_search_space.py
contramundum53: No, `_SearchSpaceTransform` in its current form cannot be used here because it converts categorical parameters to one-hot vectors. How about leaving it a follow-up to merge this file with `_SearchSpaceTransform`?

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,155 @@
...
+
+import numpy as np
+import scipy.stats.qmc
+
+import optuna
+
+
+LINEAR = 0
+LOG = 1
+CATEGORICAL = 2
```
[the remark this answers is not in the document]
contramundum53: This was also a side effective of trying to use numba. (Numba does not support IntEnum https://github.com/numba/numba/issues/6631.)

optuna/samplers/_gp/_sampler.py
```diff
@@ -0,0 +1,169 @@
...
+
+    This sampler fits a Gaussian process (GP) to the objective function and optimizes
+    the acquisition function to suggest the next parameters.
+
+    The current implementation uses:
+    - Matern kernel with nu=2.5 (differentiable twice)
+    - Automatic relevance determination (ARD) for a lengthscale of each parameter
+    - Gamma prior for lengthscales, kernel scale, and noise scale
+    - Log Expected Improvement (logEI) as the acquisition function
+    - Quasi-Monte Carlo (QMC) sampling to optimize the acquisition function
```
nabenabe0928, line 59: [nit] I changed it with the Oxford comma.
```suggestion
    The current implementation uses:
        - Matern kernel with nu=2.5 (twice differentiable),
        - Automatic relevance determination (ARD) for the length scale of each parameter,
        - Gamma prior for lengthscales, kernel scale, and noise scale,
        - Log Expected Improvement (logEI) as the acquisition function, and
        - Quasi-Monte Carlo (QMC) sampling to optimize the acquisition function.
```

optuna/samplers/_gp/_sampler.py
```diff
@@ -0,0 +1,169 @@
...
+
+    .. note::
+        This sampler requires `scipy` and `pytorch` to be installed.
```
nabenabe0928, line 62: nit

```suggestion
        This sampler requires `scipy` and `pytorch`.
```

optuna/samplers/_gp/_sampler.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def reseed_rng(self) -> None:
+        self._independent_sampler.reseed_rng()
+
+    def infer_relative_search_space(
+        self, study: Study, trial: FrozenTrial
+    ) -> dict[str, BaseDistribution]:
+        return self._intersection_search_space.calculate(study)
+
+    def sample_relative(
```
nabenabe0928, line 101: [suggestion] Shall we raise an error for `len(directions) > 1` as your implementation seems to not support multi-objective optimization?

optuna/samplers/_gp/_sampler.py
```diff
@@ -0,0 +1,169 @@
...
+            internal_search_space,
+            transformed_params,
+        ) = _search_space.get_search_space_and_transformed_params(trials, search_space)
+        values = np.array([trial.value for trial in trials])
+        if study.direction == StudyDirection.MINIMIZE:
+            values = -values
+        values -= values.mean()
+
+        EPS = 1e-10
+        values /= max(EPS, values.std())
```
nabenabe0928, line 122: [nit] I would like to name it `loss_vals` instead of `values` because `values` typically serves as a list of different objectives and we transform the values so that the value will be better when it is lower.

```suggestion
        minimize = study.direction == StudyDirection.MINIMIZE
        loss_vals = np.array([trial.value if minimize else -trial.value for trial in trials])
        loss_vals = (loss_vals - np.mean(loss_vals)) / max(1e-10, np.std(values))
```

optuna/samplers/_gp/_sampler.py
```diff
@@ -0,0 +1,169 @@
...
+        seed: int | None = None,
+        independent_sampler: BaseSampler | None = None,
+        n_startup_trials: int = 10,
+    ) -> None:
+        self._independent_sampler = independent_sampler or optuna.samplers.RandomSampler(seed=seed)
+        self._intersection_search_space = optuna.search_space.IntersectionSearchSpace()
+        self._n_startup_trials = n_startup_trials
+        self._log_prior: "Callable[[_gp.KernelParams], torch.Tensor]" = default_log_prior
+        self._minimum_noise: float = 1e-6
+        self._last_kernel_params: "_gp.KernelParams | None" = None
```
nabenabe0928, line 90: [nit] Do you think `cache` is better? (At least, I needed to read the corresponding lines to understand what `last` means)

```suggestion
        self._kernel_params_cache: "_gp.KernelParams | None" = None
```

nabenabe0928 (commented):
I left some minor comments about variable namings.

HideakiImamura:

optuna/samplers/_gp/_sampler.py
contramundum53: We actually transform it so that higher is better (which is the default notation of GP, cf. acquisition function)

contramundum53:
@nabenabe0928 I applied your comments. PTAL!

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+import math
+from typing import Any
+
+import numpy as np
+import scipy.stats.qmc
+
+import optuna
+
+
+class ParamType(IntEnum):
```
nabenabe0928, line 14: [nit] Probably, scale is more appropriate.
```suggestion
class ScaleType(IntEnum):
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+    if param_type == ParamType.CATEGORICAL:
+        return param_value
+    else:
+        bounds2 = (bounds[0] - 0.5 * step, bounds[1] + 0.5 * step)
+        if param_type == ParamType.LOG:
+            bounds2 = (math.log(bounds2[0]), math.log(bounds2[1]))
+        param_value = param_value * (bounds2[1] - bounds2[0]) + bounds2[0]
+        if param_type == ParamType.LOG:
+            param_value = np.exp(param_value)
+        return param_value
```
nabenabe0928, line 39:
```suggestion
    low, high = (bounds[0] - 0.5 * step, bounds[1] + 0.5 * step)
    if param_type == ParamType.LOG:
        low, high = (math.log(low), math.log(high))
    param_value = param_value * (high - low) + low
    if param_type == ParamType.LOG:
        param_value = np.exp(param_value)
    return param_value
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def transform_one_param(
```
nabenabe0928, line 42: I think normalize is more specific.
```suggestion
def normalize_one_param(
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+@dataclass(frozen=True)
+class SearchSpace:
+    param_type: np.ndarray
+    bounds: np.ndarray
+    step: np.ndarray
+
+
+def untransform_one_param(
```
nabenabe0928, line 27:
```suggestion
def denormalize_one_param(
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+) -> np.ndarray:
+    if param_type == ParamType.CATEGORICAL:
+        return param_value
+    else:
+        bounds2 = (bounds[0] - 0.5 * step, bounds[1] + 0.5 * step)
+        if param_type == ParamType.LOG:
+            bounds2 = (math.log(bounds2[0]), math.log(bounds2[1]))
+            param_value = np.log(param_value)
+        param_value = (param_value - bounds2[0]) / (bounds2[1] - bounds2[0])
+        return param_value
```
nabenabe0928, line 53:
```suggestion
    low, high = (bounds[0] - 0.5 * step, bounds[1] + 0.5 * step)
    if param_type == ParamType.LOG:
        low, high = (math.log(bounds2[0]), math.log(bounds2[1]))
        param_value = np.log(param_value)
    param_value = (param_value - low) / (high - low)
    return param_value
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+                param_values[:, i], param_types[i], (bounds[i, 0], bounds[i, 1]), steps[i]
+            )
+    return param_values
+
+
+def get_search_space_and_transformed_params(
+    trials: list[optuna.trial.FrozenTrial],
+    optuna_search_space: dict[str, optuna.distributions.BaseDistribution],
+) -> tuple[SearchSpace, np.ndarray]:
+    param_type = np.zeros(len(optuna_search_space), dtype=np.int64)
```
nabenabe0928, line 90: I think list is better because we do not take an integer inside the array anymore.

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+    bounds = np.zeros((len(optuna_search_space), 2), dtype=np.float64)
+    step = np.zeros(len(optuna_search_space), dtype=np.float64)
+    values = np.zeros((len(trials), len(optuna_search_space)), dtype=np.float64)
+    for i, (param, distribution) in enumerate(optuna_search_space.items()):
+        if isinstance(distribution, optuna.distributions.CategoricalDistribution):
+            param_type[i] = ParamType.CATEGORICAL
+            bounds[i, 0] = 0.0
+            bounds[i, 1] = len(distribution.choices)
```
nabenabe0928, line 98:
```suggestion
            bounds[i, 0] = [0.0, len(distribution.choices)]
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+            step[i] = 1.0
+            for ti, trial in enumerate(trials):
+                values[ti, i] = distribution.to_internal_repr(trial.params[param])
```
nabenabe0928, line 101:
```suggestion
            values[:, i] = [distribution.to_internal_repr(t.params[param]) for t in trials]
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+            if distribution.log:
+                param_type[i] = ParamType.LOG
+            else:
+                param_type[i] = ParamType.LINEAR
+            if distribution.step is None:
+                step[i] = 0.0
+            else:
+                step[i] = distribution.step
+            bounds[i, 0] = distribution.low
+            bounds[i, 1] = distribution.high
```
nabenabe0928, line 119:
```suggestion
            param_type[i] = ParamType.LOG if distribution.log else ParamType.LINEAR
            step[i] = 0.0 if distribution.step is None else distribution.step
            bounds[i, :] = [distribution.low, distribution.high]
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+
+            external_values = np.zeros((len(trials),))
+            for ti, trial in enumerate(trials):
+                external_values[ti] = trial.params[param]
```
nabenabe0928, line 123:
```suggestion
            external_values = np.asarray([trial.params[param] for t in trials])
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+
+from dataclasses import dataclass
+from enum import IntEnum
+import math
+from typing import Any
+
+import numpy as np
+import scipy.stats.qmc
+
+import optuna
```
nabenabe0928, line 11: What about importing them to reduce lines?
```suggestion
from optuna.distributions import BaseDistribution
from optuna.distributions import CategoricalDistribution
from optuna.distributions import FloatDistribution
from optuna.distributions import IntDistribution
from optuna.trial import FrozenTrial
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+                ),
+            )
+            if distribution.log:
+                param_type = ParamType.LOG
+            else:
+                param_type = ParamType.LINEAR
+            if distribution.step is None:
+                step = 0.0
+            else:
+                step = distribution.step
```
nabenabe0928, line 153:
```suggestion
            param_type = ParamType.LOG if distribution.log else ParamType.LINEAR
            step = 0.0 if distribution.step is None else distribution.step
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+            bounds = (distribution.low, distribution.high)
+            param_value = float(
+                untransform_one_param(transformed_param[i], param_type, bounds, step)
```
nabenabe0928, line 156: The typing is incorrect based on `untransform_one_param`.
```suggestion
                untransform_one_param(np.asarray([transformed_param[i]]), param_type, bounds, step)[0]
```

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,162 @@
...
+            )
+            param_value = min(max(param_value, distribution.low), distribution.high)
```
nabenabe0928, line 158:
```suggestion
            param_value = np.clip(param_value, distribution.low, distribution.high)
```

nabenabe0928 (commented):
Mostly minor nit picking!

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,160 @@
...
+
+class ParamType(IntEnum):
+    LINEAR = 0
+    LOG = 1
+    CATEGORICAL = 2
+
+
+@dataclass(frozen=True)
+class SearchSpace:
+    param_type: np.ndarray
```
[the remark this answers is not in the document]
nabenabe0928: Sorry, please ignore it

optuna/_gp/_search_space.py
contramundum53: Hmm...I don't think this would be more readable. I thought `float` type was semantically convertible to `np.ndarray` (as 0-dimensional array, and that can also be done explicitly by `np.array(transformed_param[i])`). Also mypy does not complain about this. How about leaving the original line as it is?

optuna/_gp/_search_space.py
```diff
@@ -0,0 +1,160 @@
...
+    else:
+        bounds2 = (bounds[0] - 0.5 * step, bounds[1] + 0.5 * step)
+        if param_type == ParamType.LOG:
+            bounds2 = (math.log(bounds2[0]), math.log(bounds2[1]))
+            x = np.log(x)
+        x = (x - bounds2[0]) / (bounds2[1] - bounds2[0])
+        return x
+
+
+def round_one_transformed_param(
```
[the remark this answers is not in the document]
contramundum53: I feel that "round" is more appropriate here, because in most cases this actually means to cut off numbers after the decimal point (but done in the transformed parameter space).

"Normalize" is too ambiguous, and it could mean a reversible process of transforming the parameter from the "unnormalized" space to a "normalized" one (which is incorrect here), or it could mean a process of translating  input (like a file path) into something equivalent but more canonical (which is again not a very accurate description of what this function does, because the input of this function is not equivalent to the output).

optuna/_gp/_search_space.py
contramundum53: Sorry, this is also an artifact of trying to use `numba`.
`numba` is very slow to process Python lists, but it is fast to process `numpy.ndarray`. We want to use `numpy.ndarray` as much as possible.

contramundum53: The `param_type` field used in `SearchSpace` class was passed to a `numba` compiled function, I'd like to stick to the `np.ndarray` implementation if this does not change readability very much.

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,177 @@
...
+def kernel(
+    is_categorical: torch.Tensor, kernel_params: KernelParams, X1: torch.Tensor, X2: torch.Tensor
+) -> torch.Tensor:
+    # kernel(x1, x2) = kernel_scale * matern52_kernel_from_sqdist(d2(x1, x2) * inv_sq_lengthscales)
+    # d2(x1, x2) = sum_i d2_i(x1_i, x2_i)
+    # d2_i(x1_i, x2_i) = (x1_i - x2_i) ** 2  # if x_i is continuous
+    # d2_i(x1_i, x2_i) = 1 if x1_i != x2_i else 0  # if x_i is categorical
+
+    d2 = (X1[..., :, None, :] - X2[..., None, :, :]) ** 2
+    d2[..., is_categorical] = torch.where(
```
nabenabe0928, line 72:
```suggestion
    # Use the Hamming distance for categorical parameters.
    d2[..., is_categorical] = torch.where(
```

optuna/_gp/_optim.py
```diff
@@ -0,0 +1,13 @@
...
+
+from . import _acqf
+from ._search_space import sample_transformed_params
+
+
+def optimize_acqf_sample(acqf: _acqf.Acqf, n_samples: int = 2048) -> tuple[np.ndarray, float]:
+    xs = sample_transformed_params(n_samples, acqf.search_space)
+    res = _acqf.eval_acqf_no_grad(acqf, xs)
+    best_i = np.argmax(res)
+    return xs[best_i, :], res[best_i]
```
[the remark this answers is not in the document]
contramundum53: This is not used in the sampler, but will be used in the `terminator` module later on.

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,177 @@
...
+# cov_Y_Y_inv_Y[len(trials)]: cov_Y_Y_inv @ Y
+# max_Y: maximum of Y (Note that we transform the objective values such that it is maximized.)
+
+
+class Matern52KernelFromSqdist(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx: typing.Any, sqdist: torch.Tensor) -> torch.Tensor:  # type: ignore
+        sqrt5d = torch.sqrt(5 * sqdist)
+        exp_part = torch.exp(-sqrt5d)
+        val = exp_part * ((1 / 3) * sqrt5d * sqrt5d + sqrt5d + 1)
```
nabenabe0928, line 31:
```suggestion
        val = exp_part * (5 * sqdist / 3 + sqrt5d + 1)
```

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,177 @@
...
+        deriv = (-5 / 6) * (sqrt5d + 1) * exp_part
```
nabenabe0928, line 32: [memo] I will check here later

nabenabe0928 (commented):
I still have some stuff to check, but I will leave some comments for now.

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,177 @@
...
+    K_x_X: torch.Tensor,
+    K_x_x: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    # mean = K_x_X @ inv(K_X_X + noise * I) @ Y
+    # var = K_x_x - K_x_X @ inv(K_X_X + noise * I) @ K_x_X.T
+
+    mean = K_x_X @ cov_Y_Y_inv_Y  # [batch]
+    kalman_gain = K_x_X @ cov_Y_Y_inv  # [batch, N]
+
+    var = K_x_x - (K_x_X * kalman_gain).sum(dim=-1)  # [batch]
```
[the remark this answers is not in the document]
contramundum53: If you do that, you would be calculating `K_x_X @ cov_Y_Y_inv` twice. This is `O(N^2)` and is potentially the bottleneck, so I'd like to compute it as fast as possible.

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def posterior(
+    cov_Y_Y_inv: torch.Tensor,
+    cov_Y_Y_inv_Y: torch.Tensor,
+    K_x_X: torch.Tensor,
+    K_x_x: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    # mean = K_x_X @ inv(K_X_X + noise * I) @ Y
+    # var = K_x_x - K_x_X @ inv(K_X_X + noise * I) @ K_x_X.T
```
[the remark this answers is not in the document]
contramundum53: No, the `posterior` function returns the posterior distribution of the true value at `x` (denoted by `y_{n+1}` in the textbook), not the observed value at `x` (denoted by `t_{n+1}`). We don't need to add the observation noise here.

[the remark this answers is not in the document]
nabenabe0928: I meant here:

```python
    # My intention
    # var = K_x_x + noise - ....
```

image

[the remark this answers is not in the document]
contramundum53: Yes, this formula is to calculate `p(t_{N+1} | t)`. What we calculate here is `p(y_{N+1} | t)` instead.

contramundum53:
@nabenabe0928 I applied most of your comments, PTAL!

optuna/_gp/_search_space.py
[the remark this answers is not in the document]
nabenabe0928: Sorry, I do not remember why I left this comment in this line...
I was probably looking at the normalize function, but somehow commented here to rename it. Anyways, please ignore it!

optuna/_gp/_search_space.py
nabenabe0928: Actually, I did not know the spec.
Then could you add an inline comment above about this spec?

optuna/_gp/_gp.py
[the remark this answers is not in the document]
nabenabe0928: Thank you for the clarification!
if we are calculating the variance of the mean of an objective function, it makes sense")

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,184 @@
...
+    cov_fX_fX = kernel(is_categorical, kernel_params, X, X)
+
+    cov_Y_Y_chol = torch.linalg.cholesky(
+        cov_fX_fX + kernel_params.noise * torch.eye(X.shape[0], dtype=torch.float64)
+    )
+    logdet = torch.log(torch.diag(cov_Y_Y_chol)).sum()
+    cov_Y_Y_chol_inv_Y = torch.linalg.solve_triangular(cov_Y_Y_chol, Y[:, None], upper=False)[:, 0]
+    return -0.5 * (
+        logdet + math.log(2 * math.pi) + torch.vdot(cov_Y_Y_chol_inv_Y, cov_Y_Y_chol_inv_Y)
+    )
```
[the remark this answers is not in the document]
contramundum53: Good catch! thanks!

contramundum53:
@y0z  I applied your review comments. PTAL.

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,186 @@
...
+        sqrt5d = torch.sqrt(5 * squared_distance)
+        exp_part = torch.exp(-sqrt5d)
+        val = exp_part * ((5 / 3) * squared_distance + sqrt5d + 1)
+        deriv = (-5 / 6) * (sqrt5d + 1) * exp_part
+        ctx.save_for_backward(deriv)
+        return val
+
+    @staticmethod
+    def backward(ctx: typing.Any, grad: torch.Tensor) -> torch.Tensor:  # type: ignore
+        (deriv,) = ctx.saved_tensors
```
nabenabe0928, line 39:
```suggestion
        # Let x be squared_distance, f(x) be forward(ctx, x), and g(f) be a provided function,
        # then deriv := df/dx, grad := dg/df, and deriv * grad = df/dx * dg/df = dg/dx.
```

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,186 @@
...
+# d2: squared distance between two points
+
+
+class Matern52Kernel(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx: typing.Any, squared_distance: torch.Tensor) -> torch.Tensor:  # type: ignore
+        sqrt5d = torch.sqrt(5 * squared_distance)
+        exp_part = torch.exp(-sqrt5d)
+        val = exp_part * ((5 / 3) * squared_distance + sqrt5d + 1)
+        deriv = (-5 / 6) * (sqrt5d + 1) * exp_part
```
nabenabe0928, line 33:
```suggestion
        # Notice that the derivative is taken w.r.t. d^2, but not w.r.t. d.
        deriv = (-5 / 6) * (sqrt5d + 1) * exp_part
```

nabenabe0928 (commented):
I verified the Matern5/2 formulation.

optuna/_gp/_acqf.py
```diff
@@ -0,0 +1,126 @@
...
+    cov_Y_Y_inv: torch.Tensor,
+    cov_Y_Y_inv_Y: torch.Tensor,
+    max_Y: torch.Tensor,
+    x: torch.Tensor,
+    # Additional noise to prevent numerical instability.
+    # Usually this is set to a very small value.
+    stabilizing_noise: float,
+) -> torch.Tensor:
+    cov_fx_fX = kernel(is_categorical, kernel_params, x[..., None, :], X)[..., 0, :]
+    cov_fx_fx = MATERN_KERNEL0 * kernel_params.kernel_scale
```
nabenabe0928, line 61: Is it possible to include `MATERN_KERNEL0` in `KernelParams`?

optuna/_gp/_acqf.py
```diff
@@ -0,0 +1,126 @@
...
+    (mean, var) = posterior(cov_Y_Y_inv, cov_Y_Y_inv_Y, cov_fx_fX, cov_fx_fx)
+    val = logei(mean, var + stabilizing_noise, max_Y)
+
+    return val
+
+
+@dataclass(frozen=True)
+class Acqf:
```
nabenabe0928, line 69: Could this name be more informative?

```suggestion
class AcquisitionFuncParams:
```

optuna/_gp/_acqf.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def eval_acqf(acqf: Acqf, x: torch.Tensor) -> torch.Tensor:
+    return eval_logei(
+        kernel_params=acqf.kernel_params,
+        X=torch.from_numpy(acqf.X),
+        is_categorical=torch.from_numpy(acqf.search_space.scale_types == ScaleType.CATEGORICAL),
+        cov_Y_Y_inv=torch.from_numpy(acqf.cov_Y_Y_inv),
+        cov_Y_Y_inv_Y=torch.from_numpy(acqf.cov_Y_Y_inv_Y),
+        max_Y=torch.tensor(acqf.max_Y, dtype=torch.float64),
```
nabenabe0928, line 118: Can we simply pass `acqf` instead of each param?

optuna/_gp/_acqf.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+@dataclass(frozen=True)
+class Acqf:
+    kernel_params: KernelParams
+    X: np.ndarray
+    search_space: SearchSpace
+    cov_Y_Y_inv: np.ndarray
+    cov_Y_Y_inv_Y: np.ndarray
+    max_Y: np.ndarray
```
nabenabe0928, line 75: Do they have to be np.ndarray or could they also be torch.Tensor?
I think the latter requires fewer communication between torch and numpy.

nabenabe0928 (commented):
I checked till marginal log likelihood.

optuna/_gp/_acqf.py
contramundum53: `KernelParams` is parameters that need to be fit when using GP. I didn't think that putting `MATERN_KERNEL0` in `KernelParams` made sense.

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,189 @@
...
+    cov_fx_fX: torch.Tensor,  # [(batch,) len(trials)]
+    cov_fx_fx: torch.Tensor,  # Scalar or [(batch,)]
+) -> tuple[torch.Tensor, torch.Tensor]:  # [(batch,)], [(batch,)]
+    # mean = cov_fx_fX @ inv(cov_fX_fX + noise * I) @ Y
+    # var = cov_fx_fx - cov_fx_fX @ inv(cov_fX_fX + noise * I) @ cov_fx_fX.T
+
+    mean = cov_fx_fX @ cov_Y_Y_inv_Y  # [batch]
+    var = cov_fx_fx - (cov_fx_fX * (cov_fx_fX @ cov_Y_Y_inv)).sum(dim=-1)  # [batch]
+    # We need to clamp the variance to avoid negative values due to numerical errors.
+    return (mean, torch.clamp(var, min=0.0))
```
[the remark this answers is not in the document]
contramundum53: No, this happens a lot when the precision is lost due to numerical errors. It would be annoying for user to see the warning in that case.

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,191 @@
...
+        loss = -marginal_log_likelihood(
+            torch.from_numpy(X), torch.from_numpy(Y), torch.from_numpy(is_categorical), params
+        ) - log_prior(params)
+        loss.backward()  # type: ignore
+        return loss.item(), np.concatenate(
+            [
+                log_inv_sq_lengthscales.grad.detach().numpy(),  # type: ignore
+                [log_kernel_scale.grad.item(), log_noise.grad.item()],  # type: ignore
+            ]
+        )
```
nabenabe0928, line 181:
```suggestion
        return loss.item(), log_kernel_params_tensor.grad.detach().numpy()
```

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,191 @@
...
+
+    res = scipy.optimize.minimize(loss_func, repr0, jac=True)
```
nabenabe0928, line 183:
```suggestion
    # jac=True means loss_func returns the Jacobian matrix for gradient descent.
    res = scipy.optimize.minimize(loss_func, repr0, jac=True)
```

optuna/samplers/_gp/_sampler.py
```diff
@@ -0,0 +1,178 @@
...
+    _gp = _LazyImport("optuna._gp._gp")
+    _optim = _LazyImport("optuna._gp._optim")
+    _acqf = _LazyImport("optuna._gp._acqf")
+
+
+def log_prior(kernel_params: "_gp.KernelParams") -> "torch.Tensor":
+    # Log of prior distribution of kernel parameters.
+
+    def gamma_log_prior(x: "torch.Tensor", concentration: float, rate: float) -> "torch.Tensor":
+        return (concentration - 1) * torch.log(x) - rate * x
```
nabenabe0928, line 40:
```suggestion
        # We omit the constant factor `rate ** concentration / factorial(concentration)`.
        return (concentration - 1) * torch.log(x) - rate * x
```

optuna/samplers/_gp/_sampler.py
```diff
@@ -0,0 +1,178 @@
...
+
+    return (
+        gamma_log_prior(kernel_params.inv_sq_lengthscales, 2, 0.5).sum()
+        + gamma_log_prior(kernel_params.kernel_scale, 2, 1)
+        + gamma_log_prior(kernel_params.noise, 1.1, 20)
+    )
```
nabenabe0928, line 46: Do you have any reasons why you picked these values?

```suggestion
    return (
        gamma_log_prior(kernel_params.inv_sq_lengthscales, 2, 0.5).sum()
        + gamma_log_prior(kernel_params.kernel_scale, 2, 1)
        + gamma_log_prior(kernel_params.noise, 1.1, 20)
    )
```

optuna/samplers/_gp/_sampler.py
```diff
@@ -0,0 +1,178 @@
...
+            internal_search_space,
+            normalized_params,
+        ) = _search_space.get_search_space_and_normalized_params(trials, search_space)
+        score_vals = np.array([trial.value for trial in trials])
+        if study.direction == StudyDirection.MINIMIZE:
+            score_vals = -score_vals
+        score_vals -= score_vals.mean()
+
+        EPS = 1e-10
+        score_vals /= max(EPS, score_vals.std())
```
nabenabe0928, line 130:
```suggestion
        _sign = -1.0 if study.direction == StudyDirection.MINIMIZE else 1.0
        score_vals = np.array([_sign * trial.value for trial in trials])
        standarized_score_vals = (score_vals - score_vals.mean()) / max(1e-10, score_vals.std())
```

optuna/samplers/_gp/_sampler.py
```diff
@@ -0,0 +1,178 @@
...
+        states = (TrialState.COMPLETE,)
+        trials = study._get_trials(deepcopy=False, states=states, use_cache=True)
+        if len(trials) < self._n_startup_trials:
+            return {}
+
+        (
+            internal_search_space,
+            normalized_params,
+        ) = _search_space.get_search_space_and_normalized_params(trials, search_space)
+        score_vals = np.array([trial.value for trial in trials])
```
nabenabe0928, line 124: Btw, I did not pay attention, but how do we handle `inf`, `-inf`, `None`, and `nan`?

optuna/_gp/_acqf.py
```diff
@@ -0,0 +1,126 @@
...
+
+    return Acqf(
+        kernel_params=kernel_params,
+        X=X,
+        search_space=search_space,
+        cov_Y_Y_inv=cov_Y_Y_inv,
+        cov_Y_Y_inv_Y=cov_Y_Y_inv_Y,
+        max_Y=np.max(Y),
+        acqf_stabilizing_noise=acqf_stabilizing_noise,
+    )
```
nabenabe0928, line 108:
```suggestion

    return Acqf(
        kernel_params=kernel_params,
        X=X,
        search_space=search_space,
        cov_Y_Y_inv=cov_Y_Y_inv,
        cov_Y_Y_inv_Y=cov_Y_Y_inv @ Y,
        max_Y=np.max(Y),
        acqf_stabilizing_noise=acqf_stabilizing_noise,
    )
```

optuna/_gp/_acqf.py
```diff
@@ -0,0 +1,126 @@
...
+            torch.from_numpy(search_space.scale_types == ScaleType.CATEGORICAL),
+            kernel_params,
+            torch.from_numpy(X),
+            torch.from_numpy(X),
+        )
+        cov_Y_Y_inv = (
+            torch.linalg.inv(K + kernel_params.noise * torch.eye(X.shape[0], dtype=torch.float64))
+            .detach()
+            .numpy()
+        )
```
nabenabe0928, line 97:
```suggestion
    with torch.no_grad():
        X_tensor = torch.from_numpy(X)
        is_categorical = torch.from_numpy(search_space.scale_types == ScaleType.CATEGORICAL)
        cov_Y_Y = kernel(is_categorical, kernel_params, X_tensor, X_tensor).detach().numpy()
        cov_Y_Y[np.diag_indices(X.shape[0])] += kernel_params.noise.item()
        cov_Y_Y_inv = np.linalg.inv(cov_Y_Y)
```

optuna/samplers/_gp/_sampler.py
contramundum53: We don't and cannot handle them, I think. Neither does `BoTorchSampler`.

optuna/samplers/_gp/_sampler.py
contramundum53: No, I just visualized these priors and picked ones that looked sensible.
I checked the BoTorch default priors, but BoTorch defaults didn't look very sensible. (Possibly BoTorch default doesn't assume the inputs (X and Y) are normalized/standardized, or the implementer possibly mistook GPyTorch's definition of `Gamma` distribution I think.)

contramundum53: Also since we are using different prior distribution than BoTorch (we use gamma prior on `inv_sq_lengthscales` while BoTorch uses gamma prior on `lengthscales`), we cannot use the same parameters anyway. I feel using gamma prior on `lengthscales` would place too much prior on very small lengthscales, which is very bad (I confirm that in many cases `BoTorch` does indeed falls back to very small lengthscales. Small lengthscales mean no information, so that would render optimization useless.)

optuna/samplers/_gp/_sampler.py
nabenabe0928: So shall we filter them above?

optuna/samplers/_gp/_sampler.py
nabenabe0928: Thank you for the clarification!
Then let's just make a note about it:)

optuna/samplers/_gp/_sampler.py
```diff
@@ -0,0 +1,178 @@
...
+    _acqf = _LazyImport("optuna._gp._acqf")
+
+
+def log_prior(kernel_params: "_gp.KernelParams") -> "torch.Tensor":
+    # Log of prior distribution of kernel parameters.
+
+    def gamma_log_prior(x: "torch.Tensor", concentration: float, rate: float) -> "torch.Tensor":
+        return (concentration - 1) * torch.log(x) - rate * x
+
+    return (
```
nabenabe0928, line 42:
```suggestion

    # NOTE(contramundum53): The parameters below were picked qualitatively.
    return (
```

optuna/_gp/_acqf.py
contramundum53: I wanted to cut the dependency on `Acqf` class here, because there could be more acquisition functions (e.g. UCB) in the near future and we would then be branching here.

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,189 @@
...
+    cov_fX_fX = kernel(is_categorical, kernel_params, X, X)
+
+    cov_Y_Y_chol = torch.linalg.cholesky(
+        cov_fX_fX + kernel_params.noise * torch.eye(X.shape[0], dtype=torch.float64)
+    )
+    logdet = torch.log(torch.diag(cov_Y_Y_chol)).sum()
+    cov_Y_Y_chol_inv_Y = torch.linalg.solve_triangular(cov_Y_Y_chol, Y[:, None], upper=False)[:, 0]
+    return -0.5 * (
+        logdet
+        + X.shape[0] * math.log(2 * math.pi)
```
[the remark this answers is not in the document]
contramundum53: This wouldn't be burdensome to compute, so I would leave it here.

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,191 @@
...
+    # sqrt5d = sqrt(5 * squared_distance)
+    # exp(sqrt5d) * (1/3 * sqrt5d ** 2 + sqrt5d + 1)
+    #
+    # We cannot let PyTorch differentiate the above expression because
+    # the gradient runs into 0/0 at squared_distance=0.
+    return Matern52Kernel.apply(squared_distance)  # type: ignore
+
+
+@dataclass(frozen=True)
+class KernelParams:
```
nabenabe0928, line 60: Can we add a method like this and use it instead of `MATERN_KERNEL0`?

```python
    @classmethod
    def kernel_value_for_zero_distance(cls) -> float:
        return 1.0
```

optuna/_gp/_gp.py
```diff
@@ -0,0 +1,191 @@
...
+    # d2_i(x1_i, x2_i) = 1 if x1_i != x2_i else 0  # if x_i is categorical
+
+    d2 = (X1[..., :, None, :] - X2[..., None, :, :]) ** 2
+
+    # Use the Hamming distance for categorical parameters.
+    d2[..., is_categorical] = torch.where(
+        d2[..., is_categorical] > 0.0,
+        torch.tensor(1.0, dtype=torch.float64),
+        torch.tensor(0.0, dtype=torch.float64),
+    )
```
nabenabe0928, line 86:
```suggestion
    d2[..., is_categorical] = (d2[..., is_categorical] > 0.0).type(torch.float64)
```

optuna/samplers/_gp/_sampler.py
contramundum53: I clamped inf values to current worst/best values.

optuna/_gp/_gp.py
nabenabe0928: Sorry, please ignore it!

contramundum53:
@nabenabe0928 I applied most of your comments. PTAL!

contramundum53:
Toy Benchmark:
```
import optuna
study = optuna.create_study(sampler=optuna.samplers.GPSampler(seed=0, n_startup_trials=0))
def objective(trial):
    x = trial.suggest_float("x", -2, 1)
    y = trial.suggest_float("y", 0.5, 1.5, log=True)
    z = trial.suggest_int("z", -3, 3)
    w = trial.suggest_int("w", 1, 10, log=True)
    u = trial.suggest_categorical("u", [-1, 0, 1])
    return x ** 2 + (y-1) ** 2 + ((z-0.3) / 3) ** 2 + ((w-4.6) / 5) ** 2 + (u - 0.2) ** 2

optuna.visualization.plot_optimization_history(study)
```
image

nabenabe0928:
@contramundum53 I will merge this PR to the master branch once the errors in the windows tests are resolved!

y0z (approved):
LGTM

optuna/samplers/_gp/sampler.py
```diff
@@ -0,0 +1,209 @@
...
+            kernel_params=kernel_params,
+            search_space=internal_search_space,
+            X=normalized_params,
+            Y=standarized_score_vals,
+        )
+
+        normalized_param, _ = optim.optimize_acqf_sample(
+            acqf_params,
+            n_samples=self._optimize_n_samples,
+            seed=self._rng.rng.random_integers(0, np.iinfo(np.int32).max),
```
nabenabe0928, line 184: `random_integers` is already deprecated in the latest version, so let's use `randint`.
```suggestion
            seed=self._rng.rng.random_int(np.iinfo(np.int32).max),
```

nabenabe0928 (changes requested):

optuna/samplers/_gp/sampler.py
contramundum53:
```suggestion
            seed=self._rng.rng.randint(np.iinfo(np.int32).max),
```

nabenabe0928 (approved):

Issue #5138, opened 2023-12-05
`JSONDecodeError` in `JournalStorage`


not522:
### Expected behavior

The all CIs should be passed but `test_pop_waiting_trial_multiprocess_safe` in `tests/storages_tests/test_journal.py` sometimes fails. https://github.com/optuna/optuna/actions/runs/7079914046/job/19267103786

### Environment

- Optuna version: master
- Python version: (at least) 3.7 & 3.8
- OS: Ubuntu
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
___________________ test_pop_waiting_trial_multiprocess_safe ___________________
concurrent.futures.process._RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/concurrent/futures/process.py", line 239, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/home/runner/work/optuna/optuna/tests/storages_tests/test_journal.py", line 103, in pop_waiting_trial
    study = optuna.load_study(storage=storage, study_name=study_name)
  File "/home/runner/work/optuna/optuna/optuna/_convert_positional_args.py", line 83, in converter_wrapper
    return func(**kwargs)
  File "/home/runner/work/optuna/optuna/optuna/study/study.py", line 1350, in load_study
    return Study(study_name=study_name, storage=storage, sampler=sampler, pruner=pruner)
  File "/home/runner/work/optuna/optuna/optuna/study/study.py", line 85, in __init__
    self._directions = storage.get_study_directions(study_id)
  File "/home/runner/work/optuna/optuna/optuna/storages/_journal/storage.py", line 213, in get_study_directions
    self._sync_with_backend()
  File "/home/runner/work/optuna/optuna/optuna/storages/_journal/storage.py", line 149, in _sync_with_backend
    logs = self._backend.read_logs(self._replay_result.log_number_read)
  File "/home/runner/work/optuna/optuna/optuna/storages/_journal/file.py", line 175, in read_logs
    raise last_decode_error
  File "/home/runner/work/optuna/optuna/optuna/storages/_journal/file.py", line 184, in read_logs
    logs.append(json.loads(line))
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/json/__init__.py", line 357, in loads
    return _default_decoder.decode(s)
[...]

PR #5195, opened 2024-01-22
Fix JSONDecodeError in JournalStorage

y0z:
## Motivation
- Fix JSONDecodeError in JournalStorage #5138

## Description of the changes
- Maintain `remaining_log_size` to allow writing by another process while reading the log.

### DIFF
--- a/optuna/storages/_journal/file.py
+++ b/optuna/storages/_journal/file.py
@@ -11,6 +11,7 @@ from typing import List
 from typing import Optional
 import uuid
 
+from optuna.exceptions import StorageInternalError
 from optuna.storages._journal.base import BaseJournalLogStorage
 
 
@@ -164,6 +165,7 @@ class JournalFileStorage(BaseJournalLogStorage):
     def read_logs(self, log_number_from: int) -> List[Dict[str, Any]]:
         logs = []
         with open(self._file_path, "rb") as f:
+            # Maintain remaining_log_size to allow writing by another process while reading the log.
             remaining_log_size = os.stat(self._file_path).st_size
             log_number_start = 0
             if log_number_from in self._log_number_offset:
@@ -172,31 +174,31 @@ class JournalFileStorage(BaseJournalLogStorage):
                 remaining_log_size -= self._log_number_offset[log_number_from]
 
             last_decode_error = None
-            log_number = log_number_start
-            for line in f:
-                if last_decode_error is not None:
-                    raise last_decode_error
+            for log_number, line in enumerate(f, start=log_number_start):
+                if log_number < log_number_from:
+                    continue
+
                 byte_len = len(line)
                 remaining_log_size -= byte_len
-                if line in b"\r\n":  # to avoid json.loads(<line separator>)
-                    self._log_number_offset[log_number] += byte_len
-                    continue
                 if remaining_log_size < 0:
                     break
+                if last_decode_error is not None:
+                    raise last_decode_error
                 if log_number + 1 not in self._log_number_offset:
                     self._log_number_offset[log_number + 1] = (
                         self._log_number_offset[log_number] + byte_len
                     )
-                if log_number < log_number_from:
+
+                # Ensure that each line ends with line separators (\n, \r\n).
+                if not line.endswith(b'\n') and not line.endswith(b'\r\n'):
+                    last_decode_error = StorageInternalError("Invalid log format.")
+                    del self._log_number_offset[log_number + 1]
                     continue
                 try:
                     logs.append(json.loads(line))
                 except json.JSONDecodeError as err:
                     last_decode_error = err
                     del self._log_number_offset[log_number + 1]
-                if remaining_log_size == 0:
-                    break
-                log_number += 1
             return logs
 
     def append_logs(self, logs: List[Dict[str, Any]]) -> None:
--- a/optuna/storages/_journal/file.py
+++ b/optuna/storages/_journal/file.py
@@ -190,7 +190,6 @@ class JournalFileStorage(BaseJournalLogStorage):
                     "\r",
                     "\r\n",
                 ]:  # to avoid json.loads(<line separator>)
-                    print('bang!')
                     continue
                 try:
                     logs.append(json.loads(line))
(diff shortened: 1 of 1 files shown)


ernestum:
Hi thanks for fixing this. I just did a test run of this branch on the cluster with 100 parallel workers and observed no more crashes :-)

optuna/storages/_journal/file.py
```diff
@@ -164,28 +164,40 @@ def __init__(self, file_path: str, lock_obj: Optional[JournalFileBaseLock] = Non
...
                     self._log_number_offset[log_number + 1] = (
                         self._log_number_offset[log_number] + byte_len
                     )
                 if log_number < log_number_from:
                     continue
+                if line.decode("utf-8") in [
+                    "\n",
+                    "\r",
+                    "\r\n",
+                ]:  # to avoid json.loads(<line separator>)
```
c-bata, line 192: Can we just write this like below?

```suggestion
                if line in b"\r\n":  # to avoid json.loads(<line separator>)
```

c-bata (changes requested):
Thank you for your pull request. I left two comments.

optuna/storages/_journal/file.py
y0z: I found that this implementation and your suggestion did not work correctly and have fixed the implementation to handle the following corner case.

## Corner case

**Process 1 (writer)**
time step1-a: write '{"operation", "op1"}'
time step1-c: write '\n'
time step2: write '{"operation", "op2"}\n'
 
**Process 2 (reader)**
time step 1-b: (read) log_number = 0, line = '{"operation", "op1"}', offset[1] = len(line)
time step 3-a: (read) line = '\n'

**logfile** (after time step 2)
line 0: {"operation", "op1"}\n
line 1: {"operation", "op2"}\n


Based on my understanding, at time step 3-a, `log_number` should be 0 and `offset[1]` should be updated (`offset[1] += len('\n')`).
Then, reading should `continue` rather than `break` when finding a `line` only contains \<line separator\>), otherwise, an empty log will be returned nevertheless line 1 exists (the following time step 3-b).

**Process 2 (reader)**
time step 3-b: (read) log_number = 1, line = '{"operation", "op2"}\n', offset[1] = len('{"operation", "op1"}') + len('\n')

----

The following json.loads' behavior causes this corner case.
We need to avoid to evaluate `json.loads('\n')` but maintain the log_number and offset properly according to `'\n'`.

```python
>>> json.loads('{"hoge": "piyo"}')
{'hoge': 'piyo'}

>>> json.loads('{"hoge": "piyo"}\n')
{'hoge': 'piyo'}

>>> json.loads('\n')
json.decoder.JSONDecodeError: Expecting value: line 2 column 1 (char 1)
```

y0z: Thank you for your feedback.

Your updated code's idea that ensures each line ends with linesep seems to work fine.
I think a line ends with '\r' should not be allowed because it will make a split like '{"op": "op1"}\r' and "\n".
And, platforms use '\r' (e.g. MacOS ver. < 9) as a linesep are quite rare today.

c-bata (approved):
The changes look good to me, but a second thorough review would be helpful.

optuna/storages/_journal/file.py
```diff
@@ -164,28 +164,40 @@ def __init__(self, file_path: str, lock_obj: Optional[JournalFileBaseLock] = Non
...
                     raise last_decode_error
                 if log_number + 1 not in self._log_number_offset:
-                    byte_len = len(line)
                     self._log_number_offset[log_number + 1] = (
                         self._log_number_offset[log_number] + byte_len
                     )
-                if log_number < log_number_from:
+
+                # Ensure that each line ends with line separators (\n, \r\n)
+                if not line.endswith(b"\n") and not line.endswith(b"\r\n"):
```
c-bata, line 192: A nit. The code looks equivalent to the following.

```suggestion
                if not line.endswith(b"\n"):
```

optuna/storages/_journal/file.py
```diff
@@ -164,28 +164,40 @@ def __init__(self, file_path: str, lock_obj: Optional[JournalFileBaseLock] = Non
...
             log_number_start = 0
             if log_number_from in self._log_number_offset:
                 f.seek(self._log_number_offset[log_number_from])
                 log_number_start = log_number_from
+                remaining_log_size -= self._log_number_offset[log_number_from]
 
             last_decode_error = None
             for log_number, line in enumerate(f, start=log_number_start):
+                if log_number < log_number_from:
+                    continue
```
c-bata, line 178: Sorry, I've noticed that we need to put this if statement under L189. Otherwise, remaining_log_size will be wrong.

```python
for log_number, line in ...:
    ...
    if log_number + 1 not in ...:
         self._log_number_offset[log_number + 1] = ...
    if log_number < log_number_from:
         continue
     # Ensure that ...
```

c-bata (approved):
LGTM. Thanks for the update!

optuna/storages/_journal/file.py
```diff
@@ -164,28 +164,39 @@ def __init__(self, file_path: str, lock_obj: Optional[JournalFileBaseLock] = Non
...
                 if log_number + 1 not in self._log_number_offset:
-                    byte_len = len(line)
                     self._log_number_offset[log_number + 1] = (
                         self._log_number_offset[log_number] + byte_len
                     )
                 if log_number < log_number_from:
                     continue
+
+                # Ensure that each line ends with line separators (\n, \r\n)
+                if not line.endswith(b"\n"):
```
not522, line 191: How about using the line separator specified by `os.linesp`?

```suggestion
                if not line.endswith(os.linesep.encode()):
```

y0z:
I applied all the comments. PTAL.

not522:
Sorry, I realize that my previous comment (using `os.linesep`) doesn't work on Windows. It is because the line separator for `what_to_write.encode("utf-8")` is `\n` instead of `\r\n`. Could you revert the last commit?

y0z:
I see. I've reverted the change.

not522 (approved):
Thank you for the update. LGTM.

Issue #5184, opened 2024-01-14
Support best value with constraint


hrntsm:
### Motivation

Study's best_value (or best_xxx) will output a value that does not consider the constraint, even if they exist.
I want to output the best value that takes into account the constraints.

### Description

For example, best_trials gets COMPLETE trials and creates pareto front trials, but by making it trials that is COMPLETE and feasible, the best_trials including the constraint condition is obtained.
https://github.com/optuna/optuna/blob/b8ff10bae792d826898500f0e375e6cb2dcecc98/optuna/study/_multi_objective.py#L14

However, since there are multiple codes related to Study.best_xxx, is it possible to implement them?
Let me know if you have any concerns.

not522:
Thanks for the suggestion. I think it is an important feature, but there are a couple of points that need to be discussed.

I have two concerns: first, the loss of backward compatibility with this improvement, and second, the difficulty of efficient implementation at the storage layer.

It is debatable, but since constrained optimization is an experimental feature, backward compatibility may not be too much of a concern. Also, efficiency can be ignored, at least in the current situation.

In my opinion, I am in favor of implementing this feature in the `Study`'s methods with `UserWarning`.

hrntsm:
Thanks for the reply.
I will try to implement it and ask again if there are any questions.

PR #5201, opened 2024-01-26
Add support best value with constraint

hrntsm:
## Motivation

implement #5184 

## Description of the changes

Until now, when finding the value of best in Study, it was calculated from the COMPLETE trial. In addition to that, it was changed so that the trial would satisfy the constraint.

Also added a warn output to show the changes made.


optuna/storages/_base.py
```diff
@@ -17,8 +18,11 @@
 from optuna.trial import TrialState
 
 
+_CONSTRAINTS_KEY = "constraints"
```
hrntsm, line 21: I would like to use `_CONSTRAINT_KEY` in `optuna.sampler._base`, but it would be a circular import, so it is a direct value. Please let me know if there is a better way.

optuna/study/_multi_objective.py
```diff
@@ -3,15 +3,32 @@
 from typing import Sequence
 
 import optuna
+from optuna import logging
 from optuna.study._study_direction import StudyDirection
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
 
 
+_CONSTRAINTS_KEY = "constraints"
```
hrntsm, line 12: ditto

nabenabe0928:
Could @not522 and @eukaryo review this PR?

not522:
@nabenabe0928 I've been busy, so could you please reassign the reviewer?

nabenabe0928:
Could you review this PR? @HideakiImamura

HideakiImamura:
I would like to discuss the design of the implementation. As @not522 suggested [here](https://github.com/optuna/optuna/issues/5184#issuecomment-1891212936), I think we should not change the internal implementation of `BaseStorage` on the current situation because it changes the existing behavior for all users and downgrade the speed performance. How about implementing the experimental method `get_best_feasible_trial` or attribute `best_feasible_trial` of the `Study` and emitting the `Userwarning` to explain that the `best_feasible_trial` is slower than `best_trial`?

hrntsm:
Thanks for confirming.
I will modify it with that information.

hrntsm:
@HideakiImamura 
I implemented this by eliminating the changes to strorage._base and adding attributes to Study.
How about this one?
CI is failing in the mypy test section, but I will add a test if this direction is acceptable.

HideakiImamura:
@hrntsm Sorry for the late reply. The current approach looks great to me. Please proceed this direction.

hrntsm:
@HideakiImamura 
I fixed fails in CI. Could you please review?

optuna/study/study.py
```diff
@@ -171,6 +172,81 @@ def best_trials(self) -> list[FrozenTrial]:
...
+        .. seealso::
+            The :ref:`reuse_best_trial` tutorial provides a detailed example of how to use this
+            method.
+
+        """
+        warnings.warn(
+            "This attribute is slower than best_trial"
+            "because it checks for feasibility for all complete trials.",
+            UserWarning,
+        )
```
HideakiImamura, line 199: Is this warning necessary? It's already pointed out in docstring, so I don't think there's a need to display it everytime it runs. From a different perspective, following [the coding style convention](https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#logging-howto), if you just want to notify the user with a warning, I think you should use `Logger.warning()`.

optuna/study/study.py
```diff
@@ -171,6 +172,81 @@ def best_trials(self) -> list[FrozenTrial]:
...
+
+        if self._is_multi_objective():
+            raise RuntimeError(
```
HideakiImamura, line 202: Please add the test case for this error.

optuna/study/study.py
```diff
@@ -171,6 +172,81 @@ def best_trials(self) -> list[FrozenTrial]:
...
+                "A single best trial cannot be retrieved from a multi-objective study. Consider "
+                "using Study.best_trials to retrieve a list containing the best trials."
+            )
+        trials = self._storage.get_all_trials(self._study_id, states=[TrialState.COMPLETE])
+        if len(trials) == 0:
+            raise ValueError("No trials are completed yet.")
```
HideakiImamura, line 208: Ditto.

optuna/study/study.py
```diff
@@ -171,6 +172,81 @@ def best_trials(self) -> list[FrozenTrial]:
...
+
+        feasible_trials = []
+        for trial in trials:
+            constraints = trial.system_attrs.get(_CONSTRAINTS_KEY)
+            if constraints is None or all([x <= 0.0 for x in constraints]):
+                feasible_trials.append(trial)
+
+        if len(feasible_trials) == 0:
+            raise ValueError("There are no feasible trials.")
```
HideakiImamura, line 217: Ditto.

optuna/study/study.py
```diff
@@ -171,6 +172,81 @@ def best_trials(self) -> list[FrozenTrial]:
...
+
+        if len(self.directions) > 1:
```
HideakiImamura, line 219: This check is already covered in L201. Can be omitted.

optuna/study/study.py
```diff
@@ -171,6 +172,81 @@ def best_trials(self) -> list[FrozenTrial]:
...
+            )
+        trials = self._storage.get_all_trials(self._study_id, states=[TrialState.COMPLETE])
+        if len(trials) == 0:
+            raise ValueError("No trials are completed yet.")
+
+        feasible_trials = []
+        for trial in trials:
+            constraints = trial.system_attrs.get(_CONSTRAINTS_KEY)
+            if constraints is None or all([x <= 0.0 for x in constraints]):
+                feasible_trials.append(trial)
```
HideakiImamura, line 214: How about using `_get_feasible_trials` in `study/_multi_objective.py`?

optuna/study/study.py
HideakiImamura: This check is covered in L216, so can be omitted.

optuna/study/study.py
```diff
@@ -171,6 +172,81 @@ def best_trials(self) -> list[FrozenTrial]:
...
+            If your study is multi-objective,
+            use :attr:`~optuna.study.Study.best_feasible_trials` instead.
+
+        Returns:
+            A :class:`~optuna.trial.FrozenTrial` object of the best trial.
+
+        .. seealso::
+            The :ref:`reuse_best_trial` tutorial provides a detailed example of how to use this
+            method.
+
```
HideakiImamura, line 193: I think we should add the `Error` condition in the docstring.

optuna/study/study.py
```diff
@@ -171,6 +172,81 @@ def best_trials(self) -> list[FrozenTrial]:
...
+            raise ValueError("There are no feasible trials.")
+
+        if len(self.directions) > 1:
+            raise RuntimeError(
+                "Best trial can be obtained only for single-objective optimization."
+            )
+        direction = self.directions[0]
+
+        if direction == StudyDirection.MAXIMIZE:
+            best_feasible_trial = max(feasible_trials, key=lambda t: cast(float, t.value))
```
HideakiImamura, line 226: Could you add the test case for this?

optuna/study/study.py
```diff
@@ -171,6 +172,81 @@ def best_trials(self) -> list[FrozenTrial]:
...
+        because it checks for feasibility for all complete trials.
+
+        Returns:
+            A list of :class:`~optuna.trial.FrozenTrial` objects.
+        """
+        warnings.warn(
+            "This attribute is slower than best_trials"
+            "because it checks for feasibility for all complete trials.",
+            UserWarning,
+        )
```
HideakiImamura, line 246: This warning can be removed.

HideakiImamura (commented):

HideakiImamura:
@hrntsm I've discussed with other commiters about the direction of this PR, but haven't reached a conclusion as the opinions diverged. However, since we've received opinions from many that we should be cautious about introducing new public APIs, there's a high chance that the current policy might change. Could you please hold off on making amendments to the PR for now?

HideakiImamura:
We've had a discussion amongst the commiters regarding the direction of this PR. Given the significant impact of introducing a public API, we've concluded that it may be best to conduct it not in the upcoming minor version v3.6, but in the next planned major version v4.0. Would it be acceptable for this PR to only include changes in the private API, i.e., adding an argument to consider constraints only to `_get_pareto_front_trials`? Although we won't officially offer the API to users, users who really want to use this feature at this point will use the private API, knowing that the feature might change in the future. As for `best_feasible_trial`/`best_feasible_trials`, please allow us to reconsider it in v4.

HiroakiNatsume:
@HideakiImamura 
Thank you for the confirmation within the committer.
I agree with this conclusion as I too had concerns about the impact.
As for the specific modifications, would it be correct to eliminate the changes in Study.py and only modify _multi_objective.py with the comments I received?

hrntsm:
@HideakiImamura
Thank you for the confirmation within the committer.
I agree with this conclusion as I too had concerns about the impact.

As for the specific modifications, would it be correct to eliminate the changes in Study.py and only modify _multi_objective.py with the comments I received?

HideakiImamura (approved):
Thanks for the update! LGTM.

eukaryo:
LGTM.

hrntsm:
@eukaryo 

Thanks for the review. I have fixed the conflicts, please merge.

PR #5207, opened 2024-01-27
Adapt tests to backward-incompatible changes in moto 5.0.0

pkoz:
## Motivation

I forked Optuna and ran tests about 4 hours ago, right after `moto` [issued their new backward-incompatible release](https://github.com/getmoto/moto/releases/tag/5.0.0). 

As a consequence, one of the tests failed. Additionally, `pyproject.toml` has no version constraint for `moto` therefore, tests will fail for all new repo clones or package upgrades (as well as CI/CD unless dependencies are cached between runs). 

Edit: As mentioned by @nzw0301 `moto=5.0.0` does not support python `3.7` anymore; therefore, I've changed the approach and do not require any specific version of `moto` but instead handle the `ImportError`.

## Description of the changes

Hopefully, the fix is trivial: 
- I've replaced the imported decorator name `mock_s3` with `mock_aws`. When the import fails, retreat to `mock_s3`
- ~Added version constraint to `pyproject.toml`~

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,7 +122,7 @@ test = [
     "coverage",
     "fakeredis[lua]",
     "kaleido",
-    "moto>=5.0.0",
+    "moto",
     "pytest",
     "scipy>=1.9.2; python_version>='3.8'",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,7 +122,7 @@ test = [
     "coverage",
     "fakeredis[lua]",
     "kaleido",
-    "moto",
+    "moto>=5.0.0",
     "pytest",
     "scipy>=1.9.2; python_version>='3.8'",
 ]
--- a/tests/artifacts_tests/test_boto3.py
+++ b/tests/artifacts_tests/test_boto3.py
@@ -4,17 +4,16 @@ import io
 from typing import TYPE_CHECKING
 
 import boto3
-
-try:
-    from moto import mock_aws
-except ImportError:
-    from moto import mock_s3 as mock_aws
-
 import pytest
 
 from optuna.artifacts import Boto3ArtifactStore
 from optuna.artifacts.exceptions import ArtifactNotFound
 
+
+try:
+    from moto import mock_aws
+except ImportError:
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -122,7 +122,7 @@ test = [
     "coverage",
     "fakeredis[lua]",
     "kaleido",
-    "moto",
+    "moto>=5.0.0",
```
nzw0301, line 125: I'm not familiar with this test and library, but I think there are two ways to resolve this issue for the current CI error.

The first one is to add a version restriction:

```
    "moto<5.0.0",
```

and adding some TOOD comment to remove it after dropping python 3.7 by Optuna.

Alternatively, add a version constraint more specifically: 
```suggestion
    "moto>=5.0.0; python_version>='3.8'",
```
and adding python version dependent import for `tests/artifacts_tests/test_boto3.py`.

pkoz: Thank you for the heads-up! I was tricked because I tried to run tests on Python 3.7 and got a bunch of other errors.
Thanks for the suggestion of using the environment marker. It looked very promising, but I double-checked [PEP-508](https://peps.python.org/pep-0508) and realized that the false environment condition ignores the dependency specification.

I think there is an even more straightforward way that won't require specifying a `moto` version at all. Let me re-evaluate tests and come back with a solution :)

nabenabe0928:
Could @y0z review this PR?

nzw0301 (approved):
Thank you for updating the PR! The change looks great to me.

y0z:
Sorry for my late response. LGTM.

y0z (approved):

nabenabe0928:
@pkoz 
Thank you for the PR, as mentioned in [PR#5214](https://github.com/optuna/optuna/pull/5214), you will be the joint author of PR#5214 and this PR has already been merged with it and I will close this PR.
This was to reduce our operation cost (e.g., removing branch protection rules temporarily and whatsoever).
Sorry for the inconvenience, but your name will be surely on the new PR, so your effort will not be in vain!

PR #5211, opened 2024-01-29
Upgrade `FrozenTrial._validate()`

eukaryo:
## Motivation

More clearly define specifications of trial and increase the strictness of the software.

## Description of the changes

Added codes to check:
- _values is None if state is PRUNED or FAIL
- _values does not include NaN if state is COMPLETE


HideakiImamura (dismissed):
LGTM.

optuna/trial/_frozen.py
```diff
@@ -321,8 +322,15 @@ def _validate(self) -> None:
                     "`datetime_complete` is supposed to be None for an unfinished trial."
                 )
 
+        if self.state == TrialState.PRUNED and self._values is not None:
+            raise ValueError("`value` is supposed not to be set for a pruned trial.")
```
contramundum53, line 326:
```suggestion
            raise ValueError("`value` is not supposed to be set for a pruned trial.")
```
[nit] Fix grammatical error

optuna/trial/_frozen.py
```diff
@@ -321,8 +322,15 @@ def _validate(self) -> None:
...
+        if self.state == TrialState.FAIL and self._values is not None:
+            raise ValueError("`value` is supposed not to be set for a failed trial.")
```
contramundum53, line 328:
```suggestion
        if self.state in (TrialState.PRUNED, TrialState.FAIL) and self._values is not None:
            raise ValueError(f"values should be None for a trial with state {self.state}, but got {self._values}.")
```
[nit] Make the code readable and easier to debug for users.

optuna/trial/_frozen.py
```diff
@@ -321,8 +322,15 @@ def _validate(self) -> None:
...
         if self.state == TrialState.COMPLETE and self._values is None:
             raise ValueError("`value` is supposed to be set for a complete trial.")
```
contramundum53, line 330:
```suggestion
            raise ValueError("values should be set for a complete trial.")
```
[nit] Just making the format the same.

optuna/trial/_frozen.py
```diff
@@ -321,8 +322,15 @@ def _validate(self) -> None:
...
+        self_values: List[float] = self._values if self._values is not None else []  # for mypy
+        if self.state == TrialState.COMPLETE and any([math.isnan(x) for x in self_values]):
+            raise ValueError("`value` is supposed not to be NaN for a complete trial.")
```
contramundum53, line 333:
```suggestion
        if self.state == TrialState.COMPLETE:
            if self._values is None:
                raise ValueError("values should be set for a complete trial.")
            elif any(math.isnan(x) for x in self._values):
                raise ValueError("values should not contain NaN.")
```
A better way to silence mypy

contramundum53 (approved):
LGTM!

PR #4807, opened 2023-07-14
[hotfix] Add version constraints on LightGBM

c-bata:
## Motivation

To resolve https://github.com/optuna/optuna/actions/runs/5549532139/jobs/10133727103?pr=4793

## Description of the changes

LightGBM v4.0.0 was released and the CI became broken. This PR adds the version constraints on LigtGBM.

PR #5217, opened 2024-01-31
Fix the errors in mypy for the `Checks (Integration)` CI

nabenabe0928:
## Motivation

This is a hotfix for the current mypy errors.

## Description of the changes

I added:
1. version constraint for moto when using Python 3.7, and
2. Modification of the type-ignore in Pytorch distributed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,7 +83,7 @@ document = [
     "sphinx-gallery",
     "sphinx-plotly-directive",
     "sphinx_rtd_theme>=1.2.0",
-    "torch<=2.1.2",
+    "torch",
     "torchaudio",
     "torchvision",
 ]
@@ -100,7 +100,7 @@ integration = [
     "scikit-optimize",
     "shap",
     "tensorflow",
-    "torch<=2.1.2",
+    "torch",
     "torchaudio",
     "torchvision",
     "wandb",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,7 +83,7 @@ document = [
     "sphinx-gallery",
     "sphinx-plotly-directive",
     "sphinx_rtd_theme>=1.2.0",
-    "torch",
+    "torch<=2.1.2",
     "torchaudio",
     "torchvision",
 ]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -100,7 +100,7 @@ integration = [
     "scikit-optimize",
     "shap",
     "tensorflow",
-    "torch",
+    "torch<=2.1.2",
```
toshihikoyanase, line 103: It may be more beneficial to update the CI configuration instead of `pyproject.toml`, aligning with the strategy adopted in  https://github.com/optuna/optuna/pull/4807. By taking this approach, we can proactively avoid unforeseen version conflicts.

nabenabe0928: Thank you for the feedback:)
I will apply your suggestion!

HideakiImamura:
@nabenabe0928 Please assign one more reviewer.

HideakiImamura:
Let me confirm that this PR intends to fix the CI error of `Checks (Integration)` like [this](https://github.com/optuna/optuna/actions/runs/7748773286/job/21132045196). Then, should we change the yml fie of that CI, i.e., `checks-integration.yml`?

By the way, other CI (windows-tests) also fails today like [this](https://github.com/optuna/optuna/actions/runs/7748781618/job/21132069525). We should fix it in another PR.

HideakiImamura:
Now, I'm confirming that the current change fixes the `Checks (Integration)` in https://github.com/HideakiImamura/optuna/pull/7. See also https://github.com/HideakiImamura/optuna/pull/6 for comparison.

HideakiImamura:
Oh, I see! So the changes to the file under optuna fixes the CI in Checks (Integration), and you are doing the fixes to windows-tests together? If so, I think it would be a good idea to change the title of the PR to something appropriate.

HideakiImamura (approved):
I confirmed that the Checks (Integration) passes with the change of this PR [here](https://github.com/HideakiImamura/optuna/pull/7). LGTM.

nabenabe0928:

y0z (approved):
LGTM

PR #5220, opened 2024-02-01
Correctly fork seed for samplers with underlying random sampler

contramundum53:
## Motivation
Currently seeds are passed directly to internal samplers of many samplers (including TPESampler). This may cause problems because the two pseudo random sequence would have correlations.

## Description of the changes
Implement `_fork_seed` in `BaseSampler`, and use the function to correctly fork the seeds.

### DIFF
--- a/optuna/samplers/_base.py
+++ b/optuna/samplers/_base.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import abc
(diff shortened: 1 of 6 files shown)


contramundum53:
We decided that this is practically not a big problem and is not worth complicating the codebase.

PR #5229, opened 2024-02-06
Fix trial validation

contramundum53:
## Motivation

#5211 introduced a bug that enforced all `PRUNED` trials to have `None` value. In https://github.com/optuna/optuna/blob/f831ee677db1b32289825e5c13577ae969705879/optuna/study/_tell.py#L136-L146, pruned trials have the value of the last intermediate value (if it exists and it is acceptable as an objective value). 

## Description of the changes
Remove validation of `value` on `PRUNED` trials.

### DIFF
--- a/optuna/trial/_frozen.py
+++ b/optuna/trial/_frozen.py
@@ -322,11 +322,8 @@ class FrozenTrial(BaseTrial):
                     "`datetime_complete` is supposed to be None for an unfinished trial."
                 )
 
-        if self.state in (TrialState.PRUNED, TrialState.FAIL) and self._values is not None:
-            raise ValueError(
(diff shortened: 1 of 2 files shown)


contramundum53:
@knshnb @eukaryo Could you reivew this PR?

eukaryo:
LGTM. Thank you for your hotfix.
My understanding of Optuna's specification is that `intermediate_values` allow the presence of `NaN` (for historical reason), but `value(s)` does not allow `NaN` at any point. These are inconsistent with the specification that `value(s)` must equal `intermediate_value[-1]`. Resolving this inconsistency is outside the scope of #5211 .

eukaryo (approved):

knshnb (approved):
LGTM!

PR #5247, opened 2024-02-13
Make `gp.fit_kernel_params` more robust

contramundum53:
## Motivation
`gp.fit_kernel_params` currently does not check the status of `scipy.optimize`. `scipy.optimize` may fail due to irregularities, but currently this function silently ignores it. Also, `torch.linalg.cholesky` might raise `RuntimeError` due to numerical issues, and in this case the optimization accidentally stops due to the error.
I try to make this function more robust against such errors.

## Description of the changes
If `scipy.optimize` fails from the given initial parameters (which is possible when the new observation reveals that the old kernel params is wildly wrong), this function retries the optimization process from the default initial parameters. If the optimization still fails, the function gives out a warning and returns the given initial parameters without optimization.


optuna/_gp/gp.py
```diff
@@ -186,11 +166,54 @@ def loss_func(raw_params: np.ndarray) -> tuple[float, np.ndarray]:
...
+        # with the default initial_kernel_params.
+        # This increases the robustness of the optimization.
+        raw_params_opt = optimize(initial_kernel_params)
+    except RuntimeError:
+        try:
+            raw_params_opt = optimize(default_initial_kernel_params)
+        except RuntimeError as e:
+            warnings.warn(
+                f"The optimization of the kernel parameters failed. ({e})"
+                "The kernel parameters are set to the initial values."
```
y0z, line 212:
```suggestion
                "The kernel parameters are set to the default values."
```

nabenabe0928 (commented):
I think it is clearer to separate the repeating part and easier to check the optimization part later if necessary.

https://github.com/nabenabe0928/optuna/compare/master...nabenabe0928:optuna:code-fix/refactor-fit-kernel-params?expand=1

HideakiImamura:

optuna/_gp/gp.py
```diff
@@ -186,11 +166,54 @@ def loss_func(raw_params: np.ndarray) -> tuple[float, np.ndarray]:
...
+        inverse_squared_lengthscales=torch.ones(n_params, dtype=torch.float64),
+        kernel_scale=torch.tensor(1.0, dtype=torch.float64),
+        noise_var=torch.tensor(1.0, dtype=torch.float64),
+    )
+    if initial_kernel_params is None:
+        initial_kernel_params = default_initial_kernel_params
+
     # jac=True means loss_func returns the gradient for gradient descent.
-    res = so.minimize(loss_func, initial_raw_params, jac=True)
+    def optimize(initial_kernel_params: KernelParamsTensor) -> np.ndarray:
```
nabenabe0928, line 178: By renaming `fit_kernel_params` to `_fit_kernel_params` and introducing the following thin wrapper, any future modifications in `_fit_kernel_params` will be easier:

```python
def fit_kernel_params(...) -> KernelParamsTensor:
    default_initial_kernel_params = KernelParamsTensor(...)
    if initial_kernel_params is None:
        initial_kernel_params = default_initial_kernel_params

    for init_kernel_params in [initial_kernel_params, default_initial_kernel_params]:
        try:
            return _fit_kernel_params(..., init_kernel_params)
        except RuntimeError:
            pass

    return default_initial_kernel_params
```

In this way, it is also easier to modify the part, which makes the optimization robust, in the future as well.

Please take a look at [here](https://github.com/nabenabe0928/optuna/compare/master...nabenabe0928:optuna:code-fix/refactor-fit-kernel-params?expand=1) for more details.

contramundum53:
@nabenabe0928 Thank you for your comment! I used your code now. PTAL.

nabenabe0928 (commented):
Thank you for the change, LGTM!

y0z (approved):
LGTM

nabenabe0928 (approved):

PR #5256, opened 2024-02-17
Add a safety guard to Wilcoxon pruner, and modify the docstring

eukaryo:
## Motivation

I want to make Wilcoxon pruner more safe and effective.

## Description of the changes

- If the value of the best_trial was not better than the average of the current trial's intermediate_values, the pruner returns False.
- Docstring is modified. In each trial, the modified example code shuffles the order in which data is processed. It never shuffles the dataset itself.


nabenabe0928:
@y0z @contramundum53 Could you review this PR?

optuna/pruners/_wilcoxon.py
```diff
@@ -58,22 +58,22 @@ def eval_func(param, input_):
...
+                ordering = rng.permutation(range(len(input_data)))
+                s = []
+                for i in ordering:
                     param = trial.suggest_float("param", -1, 1)
                     loss = eval_func(param, input_data[i])
                     trial.report(loss, i)
-                    s += loss
+                    s.append(loss)
                     if trial.should_prune():
                         raise optuna.TrialPruned()
```
contramundum53, line 74:
```suggestion
                        return sum(s) / len(s)
```
Could you also change the docstring to explain that this is necessary, because this is different from other pruners?

optuna/pruners/_wilcoxon.py
```diff
@@ -167,6 +169,21 @@ def prune(self, study: "optuna.study.Study", trial: FrozenTrial) -> bool:
 
         alt = "less" if study.direction == StudyDirection.MAXIMIZE else "greater"
 
+        if alt == "less":
+            average_is_best = best_trial.value <= sum(step_values) / len(step_values)
+        else:
+            average_is_best = best_trial.value >= sum(step_values) / len(step_values)
```
contramundum53, line 175:
```suggestion
        if study.direction == StudyDirection.MAXIMIZE:
            alt = "less"
            average_is_best = best_trial.value <= sum(step_values) / len(step_values)
        else:
            alt = "greater"
            average_is_best = best_trial.value >= sum(step_values) / len(step_values)
```
A nit.

contramundum53 (approved):
Otherwise, LGTM.

optuna/pruners/_wilcoxon.py
```diff
@@ -167,6 +169,21 @@ def prune(self, study: "optuna.study.Study", trial: FrozenTrial) -> bool:
...
+
         # We use zsplit to avoid the problem when all values are zero.
         p = ss.wilcoxon(diff_values, alternative=alt, zero_method="zsplit").pvalue
+
+        if p < self._p_threshold and average_is_best:
+            # WilcoxonPruner found the current trial should be pruned,
+            # but the value of the best trial was not better than
+            # the average of the current trial's intermediate values.
+            # For safety, WilcoxonPruner concludes it should not be pruned.
+            # If this if-statement is frequently satisfied,
```
y0z, line 185: Although this is a simple question, is there a reasonable way that users can detect `this if-statement is frequently satisfied`?

optuna/pruners/_wilcoxon.py
eukaryo: I agree that the suggested technique is a beneficial workaround. However, I believe it might be too advanced for use in a docstring example. In addition, it is not necessary; for example, it has no effect with GridSampler.
I added an comment in the code and wrote a separate note.

contramundum53: If you add a note in the docstring, why not demonstrate that in the example?
This workaround is actually necessary for all samplers except for `GridSampler`, `BruteForceSampler`, `QMCSampler` and `RandomSampler` I think.

eukaryo: I see; I modified the example so that `return sum(s) / len(s)`.

y0z (approved):
LGTM

contramundum53 (approved):
LGTM.

PR #5266, opened 2024-02-22
Add a tutorial of Wilcoxon pruner

eukaryo:
## Motivation

I want to add a tutorial of wilcoxon pruner.

## Description of the changes

Added a 013_wilcoxon_pruner.py file.

### DIFF
--- a/tutorial/20_recipes/013_wilcoxon_pruner.py
+++ b/tutorial/20_recipes/013_wilcoxon_pruner.py
@@ -1,0 +1,158 @@
+"""
+.. _wilcoxon_pruner:
+
+Wilcoxon Pruner
+===============
+
+This tutorial showcases Optuna's wilcoxon pruner.
+This pruner is effective for objective functions that averages multiple evaluations.
+
+We solve Traveling Salesman Problem(TSP) by Simulated Annealing(SA) accelerated by `numba`.
+"""
+
+import math
+from typing import NamedTuple
+
+import numba
+import numpy as np
+import optuna
+from numpy.linalg import norm
+
+
+class SAOptions(NamedTuple):
+    max_iter: int = 1000
+    T0: float = 1.0
+    alpha: float = 1.0
+    patience: int = 300
+
+
+@numba.njit
+def simulated_annealing(vertices, initial_idxs, options: SAOptions):
+
+    def temperature(t: float):
+        # t: 0 ... 1
+        return options.T0 * (1 - t) ** options.alpha
+
+    idxs = initial_idxs.copy()
+    N = len(vertices)
+    assert len(idxs) == N
+
+    cost = sum(
+        [norm(vertices[idxs[i]] - vertices[idxs[(i + 1) % N]]) for i in range(N)]
+    )
+    best_idxs = idxs.copy()
+    best_cost = cost
+
+    remaining_patience = options.patience
+    np.random.seed(11111)
+
+    for iter in range(options.max_iter):
+        i = np.random.randint(0, N)
+        j = (i + 2 + np.random.randint(0, N - 3)) % N
+        i, j = min(i, j), max(i, j)
+        delta_cost = (
+            -norm(vertices[idxs[(i + 1) % N]] - vertices[idxs[i]])
+            - norm(vertices[idxs[j]] - vertices[idxs[(j + 1) % N]])
+            + norm(vertices[idxs[i]] - vertices[idxs[j]])
+            + norm(vertices[idxs[(i + 1) % N]] - vertices[idxs[(j + 1) % N]])
+        )
+        temp = temperature(iter / options.max_iter)
+
+        if delta_cost <= 0.0 or np.random.rand() < math.exp(-delta_cost / temp):
+            cost += delta_cost
+            idxs[i + 1 : j + 1] = idxs[i + 1 : j + 1][::-1]
+            if cost < best_cost:
+                best_idxs[:] = idxs
+                best_cost = cost
+
+        if cost >= best_cost:
+            remaining_patience -= 1
+            if remaining_patience == 0:
+                idxs[:] = best_idxs
+                cost = best_cost
+                remaining_patience = options.patience
+
+    return best_idxs
+
+
+###################################################################################################
+# We make a random dataset of TSP.
+
+
+def make_dataset(num_vertex, num_problem, seed):
+    rng = np.random.default_rng(seed=seed)
+    dataset = []
+    for _ in range(num_problem):
+        dataset.append(
+            {
+                "vertices": rng.random((num_vertex, 2)),
+                "idxs": rng.permutation(num_vertex),
+            }
+        )
+    return dataset
+
+
+NUM_PROBLEM = 20
+dataset = make_dataset(200, NUM_PROBLEM, seed=33333)
+
+
+###################################################################################################
+# In each trial, it is recommended to shuffle the order in which data is processed.
+# We make pseudo random number generator here.
+
+
+rng = np.random.default_rng(seed=44444)
+
+
+###################################################################################################
+# In each trial, it is recommended to shuffle the order in which data is processed.
+# We counts the number of evaluation to know how many problems is pruned.
+
+
+num_evaluation = 0
+def objective(trial):
+    global num_evaluation
+    patience = trial.suggest_int("patience", 10, 1000, log=True)
+    T0 = trial.suggest_float("T0", 0.1, 10.0, log=True)
+    alpha = trial.suggest_float("alpha", 1.1, 10.0, log=True)
+    options = SAOptions(max_iter=1000000, patience=patience, T0=T0, alpha=alpha)
+    ordering = rng.permutation(range(len(dataset)))
+    results = []
+    for i in ordering:
+        num_evaluation += 1
+        d = dataset[i]
+        result_idxs = simulated_annealing(d["vertices"], d["idxs"], options)
+        result_cost = 0.0
+        n = len(d["vertices"])
+        for j in range(n):
+            result_cost += norm(
+                d["vertices"][result_idxs[j]] - d["vertices"][result_idxs[(j + 1) % n]]
+            )
+        results.append(result_cost)
+
+        trial.report(result_cost, i)
+        if trial.should_prune():
+            return sum(results) / len(results)
+
+    return sum(results) / len(results)
+
+
+###################################################################################################
+# We use `TPESampler`` with `WilcoxonPruner``.
+
+
+NUM_TRIAL = 100
+sampler = optuna.samplers.TPESampler(seed=55555)
+pruner = optuna.pruners.WilcoxonPruner(p_threshold=0.05)
+study = optuna.create_study(direction="minimize", sampler=sampler, pruner=pruner)
+study.enqueue_trial({"patience": 300, "T0": 1.0, "alpha": 1.8})  # default params
+study.optimize(objective, n_trials=NUM_TRIAL)
+
+
+###################################################################################################
+# We can show the optimization results as:
+
+
+print(f"The number of trials: {len(study.trials)}")
+print(f"Best value: {study.best_value} (params: {study.best_params})")
+print(f"Number of evaluation: {num_evaluation} / {NUM_PROBLEM * NUM_TRIAL}")
--- a/tutorial/20_recipes/013_wilcoxon_pruner.py
+++ b/tutorial/20_recipes/013_wilcoxon_pruner.py
@@ -112,7 +112,7 @@ def objective(trial):
     patience = trial.suggest_int("patience", 10, 1000, log=True)
     T0 = trial.suggest_float("T0", 0.1, 10.0, log=True)
     alpha = trial.suggest_float("alpha", 1.1, 10.0, log=True)
-    options = SAOptions(max_iter=1000000, patience=patience, T0=T0, alpha=alpha)
+    options = SAOptions(max_iter=10000, patience=patience, T0=T0, alpha=alpha)
     ordering = rng.permutation(range(len(dataset)))
     results = []
     for i in ordering:
--- a/tutorial/20_recipes/013_wilcoxon_pruner.py
+++ b/tutorial/20_recipes/013_wilcoxon_pruner.py
@@ -37,9 +37,7 @@ def simulated_annealing(vertices, initial_idxs, options: SAOptions):
     N = len(vertices)
     assert len(idxs) == N
 
-    cost = sum(
-        [norm(vertices[idxs[i]] - vertices[idxs[(i + 1) % N]]) for i in range(N)]
-    )
+    cost = sum([norm(vertices[idxs[i]] - vertices[idxs[(i + 1) % N]]) for i in range(N)])
     best_idxs = idxs.copy()
     best_cost = cost
 
@@ -105,11 +103,12 @@ rng = np.random.default_rng(seed=44444)
 
 
 ###################################################################################################
-# In each trial, it is recommended to shuffle the order in which data is processed.
 # We counts the number of evaluation to know how many problems is pruned.
 
 
 num_evaluation = 0
+
+
 def objective(trial):
     global num_evaluation
     patience = trial.suggest_int("patience", 10, 1000, log=True)
--- a/tutorial/20_recipes/013_wilcoxon_pruner.py
+++ b/tutorial/20_recipes/013_wilcoxon_pruner.py
@@ -7,13 +7,12 @@ Wilcoxon Pruner
 This tutorial showcases Optuna's wilcoxon pruner.
 This pruner is effective for objective functions that averages multiple evaluations.
 
-We solve Traveling Salesman Problem(TSP) by Simulated Annealing(SA) accelerated by `numba`.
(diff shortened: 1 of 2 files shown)


tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,257 @@
...
+
+We solve Traveling Salesman Problem (TSP) by Simulated Annealing (SA).
+
+Overview of Traveling Salesman Problem
+--------------------------------------
+
+Traveling Salesman Problem (TSP) is a classic problem in combinatorial optimization
+that involves finding the shortest possible route for a salesman
+who needs to visit a set of cities, each exactly once, and return to the starting city.
+TThe problem is classified as NP-hard, indicating that it is extremely challenging
```
eukaryo, line 18:
```suggestion
The problem is classified as NP-hard, indicating that it is extremely challenging
```

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,350 @@
...
+
+###################################################################################################
+# .. note::
+#     The following `tsp_simulated_annealing` function can be acceralated by `numba`.
+# .. note::
+#     For simplicity, we will use the 2-opt neighborhood search to solve TSP.
+#     The 2-opt method is valued for its fundamental approach, straightforward implementation,
+#     and broad applicability. However, it's important to note that for more complex TSP instances,
+#     there are significantly more advanced methods than the 2-opt neighborhood search,
+#     such as Iterated Local Search (ILS) combined with Lin-Kernighan heuristics.
```
contramundum53, line 104:
```suggestion
#     For simplicity of implementation, we use SA with the 2-opt neighborhood to solve TSP,
#     but note that this is far from the "best" way to solve TSP. There are significantly more
#     advanced methods than this method.
```

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,350 @@
...
+# ultimately aiming for convergence towards an optimal solution.
+#
+# `Alpha` (Cooling Rate)
+# -----------------------------
+#
+# `Alpha` is a parameter that dictates the rate at which the temperature is decreased in each
+# iteration of the algorithm. It is typically a value between 0 and 1, and
+# it's used to multiply the current temperature to obtain the temperature for the next iteration,
+# following the formula `T = Alpha * T`. A smaller `Alpha` value results in a quicker decrease
+# in temperature, making the algorithm converge faster but potentially missing broader exploration.
```
contramundum53, line 209:
```suggestion
# `T0` and `alpha` defining the temperature schedule
# -----------------------------
#
# In simulated annealing, it is important to determine a good temperature scheduling, but
# there is no "silver schedule" that is good for all problems, so we must tune the schedule
# for this problem.
# This code parametrizes the temperature as a monomial function `T0 * (1 - t) ** alpha`, where 
# `t` progresses from 0 to 1. We try to optimize the two parameters `T0` and `alpha`.
```
I think we should put the explanation of temperature itself in the description of simulated annealing.
I think the explanation of the temperature should go

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,350 @@
...
+# -----------------------------
+#
+# In this specific context, `patience` refers to the mechanism of reverting to the best solution
+# found so far after a certain number of iterations without improvement. This concept is
+# somewhat akin to a "reset" or "rollback" function, where if the algorithm hasn't found a
+# better solution within the defined `patience` threshold (a set number of iterations),
+# it will revert to the best solution it has encountered. This strategy can prevent the
+# algorithm from wandering too far from promising regions of the solution space and can
+# help in maintaining a focus on refining the best solutions found, rather than continuing
+# to explore less promising paths.
```
contramundum53, line 221:
```suggestion
# This parameter specifies a threshold of how many iterations we allow the annealing process continue without updating the best value. Practically, simulated annealing often drives the solution far away from the current best solution, and rolling back to the best solution periodically often improves optimization efficiency a lot. However, if the rollback happens too often, the optimization may get stuck in a local optimum, so we must tune the threshold to a sensible amount.
```

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,350 @@
...
+#
+# .. note::
+#     As an advanced workaround, if `trial.should_prune()` returns `True`,
+#     you can return an estimation of the final value (e.g., the average of all evaluated values)
+#     instead of `raise optuna.TrialPruned()`.
+#     Some algorithms including `TPESampler` internally split trials into below (good) and above (bad),
+#     and pruned trial will always be classified as above.
+#     However, there are some trials that are slightly worse than the best trial and will be pruned,
+#     but they should be classified as below (e.g., top 10%).
+#     This workaround provides beneficial information about such trials to these algorithms.
```
contramundum53, line 231:
```suggestion
#     Some algorithms, including the default `TPESampler`, currently cannot utilize the
#     information of pruned trials effectively (especially when the last intermediate value
#     is not the best approximation to the final objective function).
#     As a workaround for this issue, you can return an estimation of the final value
#     (e.g., the average of all evaluated values) when `trial.should_prune()` returns `True`,
#     instead of `raise optuna.TrialPruned()`.
#     This will improve the sampler performance.
```

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,350 @@
...
+the algorithm becomes more conservative, accepting only solutions
+that improve the objective function or those that do not significantly worsen it.
+
+This method allows the SA algorithm to balance exploration and exploitation,
+making it effective for solving complex optimization problems where
+the solution space is large and potentially rugged with many local optima.
+However, determining an effective cooling schedule is problem-dependent and
+can be challenging. There is no one-size-fits-all approach, and
+the optimal schedule may vary significantly between different types of problems.
+Finding a good cooling schedule is an integral part of successfully applying SA.
```
contramundum53, line 56:
```suggestion
Overview of Traveling Salesman Problem
--------------------------------------

Traveling Salesman Problem (TSP) is a classic problem in combinatorial optimization
that involves finding the shortest possible route for a salesman
who needs to visit a set of cities, each exactly once, and return to the starting city.
TSP has been extensively studied in fields such as mathematics, computer science,
and operations research, and has numerous practical applications in logistics,
manufacturing, and DNA sequencing, among others.
The problem is classified as NP-hard, so approximation algorithms or
heuristic methods are commonly employed for larger instances.

One simple heuristic method applicable to TSP is simulated annealing (SA). SA starts with an initial solution (it can be constructed by a simpler heuristic like greedy method), and it randomly checks the neighborhood (defined later) of the solution. If a neighbor is better, the solution is updated to the neighbor. If the neighbor is worse, SA still updates the solution to the neighbor with probability $e^{-\Delta c / T}$, where $\Delta c (> 0)$ is the difference of the cost (sum of the distance) between the new solution and the old one and $T$ is a parameter called "temperature". The temperature controls how much worsening of the solution is tolerated to escape from the local minimum (high means more tolerant). If the temperature is too low, SA will quickly fall into a local minimum; if the temperature is too high, SA will be like a random walk and the optimization will be inefficient. Typically, we set a "temperature schedule" that starts from a high temperature and gradually decreases to zero.

There are several ways to define neighborhood for TSP, but we use a simple neighborhood called 2-opt. 2-opt neighbor chooses a path in the current solution and reverses the visiting order in the path. For example, if the initial solution is `a→b→c→d→e→a`, `a→d→c→c→e→a` is a 2-opt neighbor (the path from `b` to `d` is reversed). This neighborhood is good because computing the difference of the cost can be done in constant time (we only need to care about the start and the end of the chosen path). 

Overview of Simulated Annealing
-------------------------------

Simulated Annealing (SA) is a probabilistic optimization algorithm used to find
the global optimum of a given function.
Inspired by the physical process of annealing in metallurgy,
where materials such as metal or glass are heated to a high temperature
and then cooled slowly to remove defects and reduce energy states,
the algorithm mimics this process to search for solutions in the problem space.

The algorithm starts with an initial solution and then moves to
a neighboring solution with a certain probability that depends on the
difference in the energy states (or costs) of the solutions and
a global parameter called "temperature". At high temperatures,
the algorithm is more likely to accept worse solutions,
allowing it to explore the solution space more freely and
avoid getting stuck in local optima.
As the temperature decreases according to a cooling schedule,
the algorithm becomes more conservative, accepting only solutions
that improve the objective function or those that do not significantly worsen it.

This method allows the SA algorithm to balance exploration and exploitation,
making it effective for solving complex optimization problems where
the solution space is large and potentially rugged with many local optima.
However, determining an effective cooling schedule is problem-dependent and
can be challenging. There is no one-size-fits-all approach, and
the optimal schedule may vary significantly between different types of problems.
Finding a good cooling schedule is an integral part of successfully applying SA.
```

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+num_evaluation = 0
+
+
+###################################################################################################
+# In this tutorial, we optimize three parameters: `T0`, `alpha`, and `patience`.
+#
+# `T0` and `alpha` defining the temperature schedule
+# -----------------------------
```
eukaryo, line 192:
```suggestion
# ---------------------------------------------------------------------------------------
```

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,350 @@
...
+
+This method allows the SA algorithm to balance exploration and exploitation,
+making it effective for solving complex optimization problems where
+the solution space is large and potentially rugged with many local optima.
+However, determining an effective cooling schedule is problem-dependent and
+can be challenging. There is no one-size-fits-all approach, and
+the optimal schedule may vary significantly between different types of problems.
+Finding a good cooling schedule is an integral part of successfully applying SA.
+
+Main Tutorial: Tuning SA Parameters for Solving TSP
```
contramundum53, line 58:
```suggestion
Main Tutorial: Tuning SA Parameters for TSP
```

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,324 @@
...
+(high means more tolerant). If the temperature is too low, SA will quickly
+fall into a local minimum; if the temperature is too high, SA will be like
+a random walk and the optimization will be inefficient. Typically, we set a
+"temperature schedule" that starts from a high temperature and gradually
+decreases to zero.
+
+There are several ways to define neighborhood for TSP, but we use a
+simple neighborhood called 2-opt. 2-opt neighbor chooses a path in
+the current solution and reverses the visiting order in the path.
+For example, if the initial solution is `a→b→c→d→e→a`, `a→d→c→c→e→a` is
```
contramundum53, line 42:
```suggestion
For example, if the initial solution is `a→b→c→d→e→a`, `a→d→c→b→e→a` is
```

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,324 @@
...
+    for _ in range(len(vertices) - 1):
+        dists_from_last = norm(vertices[idxs[-1], None] - vertices, axis=-1)
+        dists_from_last[idxs] = np.inf
+        idxs.append(np.argmin(dists_from_last))
+    return np.array(idxs)
+
+
+###################################################################################################
+# .. note::
+#     The following `tsp_simulated_annealing` function can be acceralated by `numba`.
```
contramundum53, line 88:
```suggestion
```
How about removing this line because simply adding `@numba.njit` will result in error?

contramundum53:

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,324 @@
+"""
+.. _wilcoxon_pruner:
+
+Early-stopping independent evaluations by WilcoxonPruner
+============================================================
+
+This tutorial showcases Optuna's WilcoxonPruner.
```
contramundum53, line 7: WilcoxonPruner can be made a link.

eukaryo: I added link of the latest document because stable document do not have WilcoxonPruner yet.
https://optuna.readthedocs.io/en/latest/reference/generated/optuna.pruners.WilcoxonPruner.html

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,324 @@
...
+    num_vertex=100,
+    num_problem=50,
+)
+
+N_TRIALS = 50
+
+
+###################################################################################################
+# We set a very small number of SA iterations for demonstration purpose.
+# In practice, you should set a larger number of iterations.
```
HideakiImamura, line 164: It would be helpful to note the appropriate range of the number of iterations.

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,324 @@
...
+#     Some samplers, including the default ``TPESampler``, currently cannot utilize the
+#     information of pruned trials effectively (especially when the last intermediate value
+#     is not the best approximation to the final objective function).
+#     As a workaround for this issue, you can return an estimation of the final value
+#     (e.g., the average of all evaluated values) when ``trial.should_prune()`` returns ``True``,
+#     instead of `raise optuna.TrialPruned()`.
+#     This will improve the sampler performance.
+
+
+def objective(trial: optuna.Trial) -> float:
```
HideakiImamura, line 210: It would be great to explain this part. For example, `We define the objective function to be optimized as follows. We early stop the evaluation by using the pruner.`.

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,324 @@
...
+    ordering = np.random.permutation(len(dataset))
+    for i in ordering:
+        num_evaluation += 1
+        result_idxs = tsp_simulated_annealing(vertices=dataset[i], options=options)
+        result_cost = tsp_cost(dataset[i], result_idxs)
+        results.append(result_cost)
+
+        trial.report(result_cost, i)
+        if trial.should_prune():
+            # raise optuna.TrialPruned()
```
HideakiImamura, line 230:
```suggestion
            # This is a standard logic of pruning in Optuna, but this leads to degradation with TPE sampler.
            # raise optuna.TrialPruned()
```

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,324 @@
...
+# We can show the optimization results as:
+
+
+print(f"The number of trials: {len(study.trials)}")
+print(f"Best value: {study.best_value} (params: {study.best_params})")
+print(f"Number of evaluation: {num_evaluation} / {len(dataset) * N_TRIALS}")
+
+
+###################################################################################################
+# Visualize the optimization history.
```
HideakiImamura, line 262: It would be helpful to note that this history plot shows both completed and pruned trials in same ways.

HideakiImamura (commented):
Thanks for the PR. The code looks great to me. I have several comments about explanation. Please take a look.

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,324 @@
...
+:math:`T` is a parameter called "temperature". The temperature controls
+how much worsening of the solution is tolerated to escape from the local minimum
+(high means more tolerant). If the temperature is too low, SA will quickly
+fall into a local minimum; if the temperature is too high, SA will be like
+a random walk and the optimization will be inefficient. Typically, we set a
+"temperature schedule" that starts from a high temperature and gradually
+decreases to zero.
+
+There are several ways to define neighborhood for TSP, but we use a
+simple neighborhood called 2-opt. 2-opt neighbor chooses a path in
```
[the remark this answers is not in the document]
contramundum53: cf. https://en.wikipedia.org/wiki/2-opt

tutorial/20_recipes/013_wilcoxon_pruner.py
contramundum53: It not only degrades TPE Sampler, but also other samplers, so maybe we do not have to say TPESampler.

tutorial/20_recipes/013_wilcoxon_pruner.py
```diff
@@ -0,0 +1,324 @@
...
+        dists_from_last[idxs] = np.inf
+        idxs.append(np.argmin(dists_from_last))
+    return np.array(idxs)
+
+
+###################################################################################################
+# .. note::
+#     For simplicity of implementation, we use SA with the 2-opt neighborhood to solve TSP,
+#     but note that this is far from the "best" way to solve TSP. There are significantly more
+#     advanced methods than this method.
```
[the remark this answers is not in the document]
eukaryo: There is a link to wikipedia. I think it is sufficient.

HideakiImamura (approved):
LGTM!

PR #5267, opened 2024-02-22
Rename variables in `plot_timeline` files

nzw0301:
## Motivation

When I read `plot_timeline`, I'm a bit confused, so I'd like to suggest renaming a few variables.
## Description of the changes

- Avoid single character variable names:  `s` and `t`, appearing multiple lines.
    - `s` is used for `TrialState` but `timeline` plot uses `start` or `trial.datetime_start`, so it is a little bit confusing as a shortform of start related variable.
    - Similarly `t` is used for `Trial`, but these functions often use time and datetime.
- Minor suggestion on variable names


y0z:
@eukaryo Could you review this PR? Because this PR simply renames variables, this PR will be merged after getting one approval.

eukaryo (approved):
LGTM!

Issue #5175, opened 2024-01-04
Passing float-castable str to `study.tell` raises an error


knshnb:
### Expected behavior

We should catch it correctly and make the trial `FAIL` instead of raising an error.

### Environment

- Optuna version: 3.5.0
- Python version: 3.9.6
- OS: macOS
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
[I 2024-01-04 14:01:53,511] A new study created in memory with name: no-name-2af3429a-171e-4d5c-ab32-32ead2065d79
Traceback (most recent call last):
  File "/Users/knshnb/pfn/optuna/issue.py", line 6, in <module>
    study.tell(trial, "2.2")
  File "/Users/knshnb/pfn/optuna/optuna/study/study.py", line 635, in tell
    return _tell_with_warning(
  File "/Users/knshnb/pfn/optuna/optuna/study/_tell.py", line 151, in _tell_with_warning
    values_conversion_failure_message = _check_values_are_feasible(study, values)
  File "/Users/knshnb/pfn/optuna/optuna/study/_tell.py", line 70, in _check_values_are_feasible
    if math.isnan(v):
TypeError: must be real number, not str
```

### Steps to reproduce

```python
import optuna


study = optuna.create_study()
trial = study.ask()
study.tell(trial, "2.2")
```

### Additional context (optional)

Tasks
- Check if the value is an instance of `numbers.Real` instead of checking it is castable to float
- Add test cases

ryota717:
@knshnb Hi, i'm tackling this issue and is there any idea for testing?
a test like https://github.com/optuna/optuna/pull/5269/commits/f550d90b50f995992a1861589229459908bbea07#diff-2eae7c178876c67d218547f9b3b4fc142b361f02f776bcbad410df0e58fdf4d7R1311-R1313 occers mypy error.
https://github.com/optuna/optuna/actions/runs/8020379337/job/21910113006?pr=5269

knshnb:
Hi, thanks for the contribution! I think you can ignore mypy for this test case like this:
https://github.com/optuna/optuna/blob/11721cc081ec290a5c16c09de57b5be06818a51b/tests/trial_tests/test_trial.py#L585

ryota717:
I didn't know the workaround, great thanks for your reply!

PR #5269, opened 2024-02-23
Fix checking value in `study.tell` .

ryota717:
## Motivation

#5175 

## Description of the changes

- 6a082706992a4c5013a2592e27ff545529d3e29e: small fix to treat the single string value as list of single string.
- e52a96a1d6ea7f8dd159351769b7d7fa83b0fb01: fix the problem mentioned in the issue.

### DIFF
--- a/optuna/study/_tell.py
+++ b/optuna/study/_tell.py
@@ -118,7 +118,7 @@ def _tell_with_warning(
     values: Optional[Sequence[float]]
     if value_or_values is None:
         values = None
-    elif isinstance(value_or_values, Sequence):
+    elif not isinstance(value_or_values, str) and isinstance(value_or_values, Sequence):
         values = value_or_values
     else:
         values = [value_or_values]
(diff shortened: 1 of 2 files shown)


optuna/study/_tell.py
```diff
@@ -118,7 +117,7 @@ def _tell_with_warning(
     values: Optional[Sequence[float]]
     if value_or_values is None:
         values = None
-    elif isinstance(value_or_values, Sequence):
+    elif not isinstance(value_or_values, str) and isinstance(value_or_values, Sequence):
```
contramundum53, line 120: Is this change necessary?

contramundum53: I see, you are trying to make the error readable 👍

knshnb (approved):
Thanks for the PR! LGTM.

contramundum53 (approved):
Sorry for late reply, LGTM!

PR #5274, opened 2024-02-24
Enhance performance of GPSampler

contramundum53:
## Motivation
This PR enhances the performance of `GPSampler` to the same level of (if not above) BoTorch.

## Description of the changes
This PR introduces the following changes:
* Add `optimize_acqf_mixed` function that optimizes acquisition function in a mixed search space by combination of L-BFGS-B and coordinate descent and a restart strategy.
* Add `deterministic` flag to notify the sampler that the objective function is a deterministic function. This promotes the sampler to exploit variation of the objective functions in narrower ranges. Without this flag, the sampler might suppose that the noise is the dominant factor in the objective function, when the function value changes largely by slight changes in the parameters.
* Change prior to penalize both very large & very small lengthscales, to stabilize `fit_kernel_params`. Gamma prior can only penalize very large values enough but not very small values.
* Specify a tolerant `gtol` in `fit_kernel_params` and `optimize_acqf`, because otherwise the numerical instability hinders optimization.

Benchmark suggests that by introducing this change, the performance improves a lot in many problems.


contramundum53:
Some benchmark results:

sigopt benchmarks (deterministic function):
image
image
image
image

HPOBench (stochastic function, discrete space; Note that the objective function is very ill-scaled, so taking log will help for the performance.)
image
image

NASBench (stochastic function, discrete space)
image

contramundum53:
@nabenabe0928 @y0z Could you review this PR?

contramundum53:
Note that our sampler is designed for real-world problems, and real-world hyperparameter optimization problems have different characteristics from benchmark functions like `sigopt` functions. Most notably, real-world problems tend to be smoother in global landscape but is often noisy in heterogeneous way and sometimes ill-scaled.
Changing the prior of kernel parameters is important to fit in a particular domain (although we try to find a good default prior). We do not expose a public interface for it, but currently you can do that by overwriting `gp_sampler._log_prior` member.

contramundum53:
Also we do not publicly expose the interface for getting posterior distribution of the surrogate model, but you can mimic the logic in `GPSampler` (fitting the model is deterministic) and use `optuna._gp.gp.posterior` function.

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,267 @@
...
+            x0=normalized_params[continuous_params] / continuous_param_scale,
+            bounds=continuous_scaled_bounds,
+            pgtol=math.sqrt(tol),
+            maxiter=200,
+        )
+
+        if info["warnflag"] == 2:
+            fval_opt = negfun_continuous_with_grad(x_opt)[0]
+
+        if -fval_opt < fval:
```
nabenabe0928, line 94: Opposite?
```suggestion
        if -fval_opt >= fval:
```

contramundum53: No, this means "worse" because fval is maximized.

nabenabe0928: Oh sorry, I forgot that this is minimization.

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,267 @@
...
+
+        current_choice = normalized_params[param_idx]
+        choices_except_current = choices[choices != normalized_params[param_idx]]
+
+        all_params = np.repeat(normalized_params[None, :], len(choices_except_current), axis=0)
+        all_params[:, param_idx] = choices_except_current
+        fvals = eval_acqf_no_grad(acqf_params, all_params)
+        best_idx = np.argmax(fvals)
+
+        if fvals[best_idx] >= fval:
```
nabenabe0928, line 114: Why is it inequality with the equal? (Do we count tie as improvement?)
```suggestion
        if fvals[best_idx] > fval:
```

contramundum53: Sorry, that was a mistake (it might fall into infinite loop).

optuna/_gp/gp.py
```diff
@@ -64,7 +64,7 @@ def matern52_kernel_from_squared_distance(squared_distance: torch.Tensor) -> tor
     return Matern52Kernel.apply(squared_distance)  # type: ignore
 
 
-@dataclass(frozen=True)
+@dataclass
```
contramundum53, line 67:
```suggestion
@dataclass(frozen=True)
```

optuna/_gp/gp.py
```diff
@@ -184,6 +184,8 @@ def loss_func(raw_params: np.ndarray) -> tuple[float, np.ndarray]:
                 else torch.exp(raw_params_tensor[n_params + 1]) + minimum_noise
             ),
         )
+        if deterministic_objective:
+            params.noise_var = torch.tensor(minimum_noise, dtype=torch.float64)
```
contramundum53, line 188:
```suggestion
```

optuna/terminator/improvement/evaluator.py
```diff
@@ -17,8 +17,6 @@
 
 
 if TYPE_CHECKING:
-    import torch
-
```
contramundum53, line 21: I noticed that `torch` was unnecessary in this file.

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+    import scipy.optimize as so
+else:
+    from optuna import _LazyImport
+
+    so = _LazyImport("scipy.optimize")
+
+_logger = get_logger(__name__)
+
+
+def _local_search_continuous(
```
nabenabe0928, line 27: I think `_gradient_ascent` is simpler.

```suggestion
def _gradient_ascent(
```

nabenabe0928:
Memo

https://github.com/optuna/optuna/compare/master...nabenabe0928:refactor-optim-mixed-clean-with-class?expand=1

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+    acqf_params: AcquisitionFunctionParams,
+    initial_params: np.ndarray,
+    initial_fval: float,
+    continuous_params: np.ndarray,
+    continuous_param_scale: np.ndarray,
```
nabenabe0928, line 32: As `ScaleType` is log, Categorical, or linear, it might be confusing, so can we use `param_precond_scale` or something similar?

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+    tol: float,
+) -> tuple[np.ndarray, float] | None:
+    if len(continuous_params) == 0:
+        return None
+    normalized_params = initial_params.copy()
+
+    def negfun_continuous_with_grad(x: np.ndarray) -> tuple[float, np.ndarray]:
```
nabenabe0928, line 39:
```suggestion
    def negative_acqf_with_grad(scaled_cont_params: np.ndarray) -> tuple[float, np.ndarray]:
```

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+    so = _LazyImport("scipy.optimize")
+
+_logger = get_logger(__name__)
+
+
+def _local_search_continuous(
+    acqf_params: AcquisitionFunctionParams,
+    initial_params: np.ndarray,
+    initial_fval: float,
+    continuous_params: np.ndarray,
```
nabenabe0928, line 31: `continuous_indices` or `cont_indices` is less confusing.
(`continuous_params` sounds `params[cont_indices]` to me)
```suggestion
    continuous_indices: np.ndarray,
```

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+    continuous_param_scale: np.ndarray,
+    tol: float,
+) -> tuple[np.ndarray, float] | None:
+    if len(continuous_params) == 0:
+        return None
+    normalized_params = initial_params.copy()
+
+    def negfun_continuous_with_grad(x: np.ndarray) -> tuple[float, np.ndarray]:
+        normalized_params[continuous_params] = x * continuous_param_scale
```
nabenabe0928, line 40: Isn't it actually the opposite?

```suggestion
        normalized_params[continuous_params] = x / continuous_param_scale
```

I guessed in this way because, e.g., Gaussian Kernel is:
$k(x, x') = \exp(- 1/2 * (x - x')^2 / l^2)$.
Then what we want after the transformation is:
$k(lx, lx') = \exp(- 1/2 * (lx - lx')^2 / l^2) = \exp(- 1/2 * (x - x')^2)$,
if my guess is correct.

Then the argument of this function should be $lx$ and `continuous_param_scale` takes $l$ if my check is correct.
In this case, we need to divide $x$ by $l$, i.e., `continuous_param_scale`.
If my statement is incorrect, please tell me!

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+        (fval, grad) = eval_acqf_with_grad(acqf_params, normalized_params)
+        # Flip sign because scipy minimizes functions.
```
nabenabe0928, line 42:
```suggestion
        # Flip sign because scipy minimizes functions.
        # Let the scaled acqf be g(x) and the acqf be f(sx), then dg/dx = df/dx * s.
```

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+        return (-fval, -grad[continuous_params] * continuous_param_scale)
+
+    x_opt, fval_opt, info = so.fmin_l_bfgs_b(
+        func=negfun_continuous_with_grad,
+        x0=normalized_params[continuous_params] / continuous_param_scale,
+        bounds=[(0, 1 / s) for s in continuous_param_scale],
```
nabenabe0928, line 48: If the earlier statement is correct, here is also the opposite.

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+        pgtol=math.sqrt(tol),
+        maxiter=200,
+    )
+
+    if info["warnflag"] == 2:
+        fval_opt = negfun_continuous_with_grad(x_opt)[0]
```
nabenabe0928, line 54: [Question] What is the difference between `fval_opt` above and the `fval_opt` here?

```suggestion
        # Maximum number of iterations reached.
        fval_opt = negfun_continuous_with_grad(x_opt)[0]
```

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+        return None
+    normalized_params = initial_params.copy()
+
+    def negfun_continuous_with_grad(x: np.ndarray) -> tuple[float, np.ndarray]:
+        normalized_params[continuous_params] = x * continuous_param_scale
+        (fval, grad) = eval_acqf_with_grad(acqf_params, normalized_params)
+        # Flip sign because scipy minimizes functions.
+        return (-fval, -grad[continuous_params] * continuous_param_scale)
+
+    x_opt, fval_opt, info = so.fmin_l_bfgs_b(
```
nabenabe0928, line 45:
```suggestion
    scaled_cont_x_opt, neg_fval_opt, info = so.fmin_l_bfgs_b(
```

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+        bounds=[(0, 1 / s) for s in continuous_param_scale],
+        pgtol=math.sqrt(tol),
+        maxiter=200,
+    )
+
+    if info["warnflag"] == 2:
+        fval_opt = negfun_continuous_with_grad(x_opt)[0]
+
+    if -fval_opt < initial_fval or info["nit"] == 0:
+        return None  # Return None if the optimization did not improve the value.
```
nabenabe0928, line 57:
```suggestion
        # Return None if the optimization did not improve the value.
        # `nit` is the number of iterations.
        return None  # 
```

nabenabe0928 (commented):
I checked only the continuous optimization part.
I would like @contramundum53 to check if the scaling is correct.
More specifically, I think the division and the multiplication of the preconditioning scaling is the opposite.

optuna/_gp/optim_mixed.py
contramundum53: Your math is correct, and we want to make $x$ the parameter passed to scipy.optimize rather than $lx$ (which is the `normalized_params`).  So, we are multiplicating $x$ (the parameter given from scipy.optimize) with $l$ (lengthscale).

optuna/_gp/optim_mixed.py
contramundum53: It's just ensuring that `fval_opt` is consistent with `x_opt`, because `l-bfgs-b` may not return a correct `fval_opt` or `x_opt` when an error (info["warnflag"] == 2) happens.

optuna/_gp/optim_mixed.py
nabenabe0928: Thank you for the comment!
Memo:
What we want to optimize is $h(x | k)$ and what `scipy.optimize` should optimize is $h(lz | k)$ w.r.t. $z$.

optuna/_gp/optim_mixed.py
contramundum53: It seems that this line is unnecessary. I'll remove it.

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,312 @@
...
+    x_opt, fval_opt, info = so.fmin_l_bfgs_b(
+        func=negfun_continuous_with_grad,
+        x0=normalized_params[continuous_params] / continuous_param_scale,
+        bounds=[(0, 1 / s) for s in continuous_param_scale],
+        pgtol=math.sqrt(tol),
+        maxiter=200,
+    )
+
+    if info["warnflag"] == 2:
+        fval_opt = negfun_continuous_with_grad(x_opt)[0]
```
contramundum53, line 55:
```suggestion
```

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+
+    NEVER = -1
+    CONTINUOUS = -2
+    last_changed_param = NEVER
+
+    for _ in range(max_iter):
+        if last_changed_param == CONTINUOUS:
+            # Parameters not changed since last time.
+            return (normalized_params, fval)
+        res = _local_search_continuous(
```
nabenabe0928, line 229: Can we modify it to something like this as a mixed return type increases the ambiguity of codes?

```suggestion
        normalized_params, fval, updated = _local_search_continuous(
```

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+            acqf_params,
+            normalized_params,
+            fval,
+            continuous_params,
+            continuous_param_scale,
+            tol,
+        )
+        if res is not None:
+            normalized_params, fval = res
+            last_changed_param = CONTINUOUS
```
nabenabe0928, line 239: If we update the return type, it becomes:
```suggestion
        last_changed_param = CONTINUOUS if updated else last_changed_param
```

I think it looks clearer as well.

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+
+    so = _LazyImport("scipy.optimize")
+
+_logger = get_logger(__name__)
+
+
+def _local_search_continuous(
+    acqf_params: AcquisitionFunctionParams,
+    initial_params: np.ndarray,
+    initial_fval: float,
```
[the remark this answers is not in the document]
contramundum53: I think `initial_fval` is better because it is clearer that it is the function value of `initial_params`.

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+def _local_search_discrete(
+    acqf_params: AcquisitionFunctionParams,
+    initial_params: np.ndarray,
+    initial_fval: float,
+    param_idx: int,
+    choices: np.ndarray,
+    xtol: float,
+) -> tuple[np.ndarray, float] | None:
```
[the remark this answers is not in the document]
contramundum53: I think I could just put the constant definition inside the function.

contramundum53:
@nabenabe0928 I applied all comments. PTAL.

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,320 @@
...
+        pgtol=math.sqrt(tol),
+        maxiter=200,
+    )
+
+    if -neg_fval_opt < initial_fval or info["nit"] == 0:
+        # `nit` is the number of iterations.
+        return (initial_params, initial_fval, False)
+    else:
+        normalized_params[continuous_indices] = scaled_cont_x_opt * lengthscale
+        return (normalized_params, -neg_fval_opt, True)
```
nabenabe0928, line 75: Shall we remove zero improvement? (The last one included zero improvement in updated.)
And just to reduce the same mistake in the equality stuff, I just follow this format for both exhaustive search and line search as well.

```suggestion
    if -neg_fval_opt > initial_fval and info["nit"] > 0:  # Improved.
        # `nit` is the number of iterations.
        normalized_params[continuous_indices] = scaled_cont_x_opt * lengthscale
        return (normalized_params, -neg_fval_opt, True)

    return (initial_params, initial_fval, False)  # No improvement.
```

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,320 @@
...
+
+    all_params = np.repeat(initial_params[None, :], len(choices_except_current), axis=0)
+    all_params[:, param_idx] = choices_except_current
+    fvals = eval_acqf_no_grad(acqf_params, all_params)
+    best_idx = np.argmax(fvals)
+
+    if fvals[best_idx] > initial_fval:
+        return (all_params[best_idx, :], fvals[best_idx], True)
+    else:
+        return (initial_params, initial_fval, False)
```
nabenabe0928, line 95: I just followed the gradient ascent where we early-return no improvement.
```suggestion
    if fvals[best_idx] > initial_fval:  # Improved.
        return (all_params[best_idx, :], fvals[best_idx], True)

    return (initial_params, initial_fval, False)  # No improvement.
```

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,320 @@
...
+    )
+    opt_idx = find_nearest_index(res.x)
+    fval_opt = -negative_acqf_with_cache(opt_idx)
+
+    # We check both conditions because of numerical errors.
+    if opt_idx != current_choice_i and fval_opt > initial_fval:
+        normalized_params[param_idx] = grids[opt_idx]
+        return (normalized_params, fval_opt, True)
+    else:
+        return (initial_params, initial_fval, False)
```
nabenabe0928, line 162:
```suggestion
    if opt_idx != current_choice_i and fval_opt > initial_fval:
        normalized_params[param_idx] = grids[opt_idx]
        return (normalized_params, fval_opt, True)

    return (initial_params, initial_fval, False)  # No improvement.
```

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,320 @@
...
+
+    continuous_indices = np.where(steps == 0.0)[0]
+
+    inverse_squared_lengthscales = (
+        acqf_params.kernel_params.inverse_squared_lengthscales.detach().numpy()
+    )
+    # This is a technique for speeding up optimization.
+    # We use an isotropic kernel, so scaling the gradient will make
+    # the hessian better-conditioned.
+    lengthscale = 1 / np.sqrt(inverse_squared_lengthscales[continuous_indices])
```
nabenabe0928, line 206: `inverse_squared_lengthscales` is plural, so I also changed it to the plural form. 
```suggestion
    lengthscales = 1 / np.sqrt(inverse_squared_lengthscales[continuous_indices])
```

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,320 @@
...
+                param_value=np.arange(bounds[i, 0], bounds[i, 1] + 0.5 * steps[i], steps[i]),
+                scale_type=ScaleType(scale_types[i]),
+                bounds=(bounds[i, 0], bounds[i, 1]),
+                step=steps[i],
+            )
+        )
+        for i in discrete_indices
+    ]
+
+    noncontinuous_paramwise_xtol = [
```
nabenabe0928, line 223:
```suggestion
    discrete_xtols = [
```

nabenabe0928 (commented):
I left some minor comments for consistency.

contramundum53:
@nabenabe0928 I applied your comments, PTAL!

optuna/samplers/_gp/sampler.py
```diff
@@ -116,13 +116,17 @@ def infer_relative_search_space(
     def _optimize_acqf(
         self,
         acqf_params: "acqf.AcquisitionFunctionParams",
+        best_params: np.ndarray,
     ) -> np.ndarray:
         # Advanced users can override this method to change the optimization algorithm.
         # However, we do not make any effort to keep backward compatibility between versions.
         # Particularly, we may remove this function in future refactoring.
-        normalized_params, _ = optim_sample.optimize_acqf_sample(
+        normalized_params, _acqf_val = optim_mixed.optimize_acqf_mixed(
```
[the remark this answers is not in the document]
contramundum53: No, it's just for clarity. (It's never used so it starts with an underscore.)

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,320 @@
...
+
+    _logger.warn("local_search_mixed: Local search did not converge.")
+    return (best_normalized_params, best_fval)
+
+
+def optimize_acqf_mixed(
+    acqf_params: AcquisitionFunctionParams,
+    *,
+    given_initial_xs: np.ndarray | None = None,
+    n_additional_samples: int = 2048,
```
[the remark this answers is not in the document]
contramundum53: This is not the number of initial parameters. How about just saying `n_samples`?

nabenabe0928 (approved):
Thanks for the work! LGTM!

optuna/_gp/optim_mixed.py
```diff
@@ -0,0 +1,319 @@
...
+        p=probs,
+    )
+
+    best_x = sampled_xs[max_i, :]
+    best_f = float(f_vals[max_i])
+
+    for x_guess in np.vstack(
+        [sampled_xs[max_i, :], sampled_xs[chosen_idxs, :], warmstart_normalized_params_array]
+    ):
+        x, f = local_search_mixed(acqf_params, x_guess, tol=tol)
```
nabenabe0928, line 314:
```suggestion
    )
    chosen_idxs = np.append(max_i, chosen_idxs)

    best_x = sampled_xs[max_i, :]
    best_f = float(f_vals[max_i])

    for x_warmstart in np.vstack([sampled_xs[chosen_idxs, :], warmstart_normalized_params_array]):
        x, f = local_search_mixed(acqf_params, x_warmstart, tol=tol)
```

nabenabe0928 (commented):
I left one last comment!

y0z (approved):
LGTM

Issue #5073, opened 2023-10-25
Unify the behavior of the terminator and its visualization function


HideakiImamura:
### Motivation

We are developing `optuna.terminator` as an automatic termination module, and have prepared `optuna.visualization.plot_terminator_improvement` for its visualization. However, there are inconsistencies in their behavior, and I propose that we align them.

### Description

The `optuna.terminator.Terminator` class is initialized with `improvement_evaluator`, `error_evaluator`, and `min_n_trials`. If `improvement_evaluator` is an instance of `BestValueStagnationEvaluator` and `error_evaluator` is `None`, `error_evaluator` is initialized with `StaticErrorEvaluator(constant=0)` ([ref](https://github.com/optuna/optuna/blob/master/optuna/terminator/terminator.py#L115)). `plot_terminator_improvement` is not set up like this. The latter should be made to match the former.

PR #5276, opened 2024-02-25
Fix inconsistencies between terminator and terminator visualization i…

SimonPop:
Hi!

The goal of this pull request is to answer issue #5073 

## Motivation
As mentioned in the issue, the objective is to harmonize the `plot_terminator_improvement` function and `Terminator` class when initializing the `error_evaluator` object. 

## Description of the changes
It adds a similar logic to the initialization of the `error_evaluator` for `plot_terminator_improvement` so that it reflects what the `Terminator` class does.

## Suggestions
Currently, I replicated the logic in `plot_terminator_improvement`. Should I factorize that code with that of the `Terminator` class so that when updated, we do not risk creating the same discrepancy in the future? 

I could add a static method to Terminator to initialize `error_evaluator`. What do you think? That way, I could also add a unit test more easily.

### DIFF
--- a/optuna/visualization/_terminator_improvement.py
+++ b/optuna/visualization/_terminator_improvement.py
@@ -12,6 +12,8 @@ from optuna.terminator import BaseErrorEvaluator
 from optuna.terminator import BaseImprovementEvaluator
 from optuna.terminator import CrossValidationErrorEvaluator
 from optuna.terminator import RegretBoundEvaluator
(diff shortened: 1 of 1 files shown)


y0z:

contramundum53:

contramundum53 (approved):
LGTM!

contramundum53:
@HideakiImamura Could you review this PR as the author of the issue?

HideakiImamura (approved):
LGTM.

PR #5298, opened 2024-03-07
Fix `_split_trials` of `TPESampler` for constrained optimization with constant liar

### LINKED ISSUE #5291: TPESampler constant_liar=True with constrained optimization spams "Trial does not have constraint values. It will be treated as a lower priority than other trials."
### Expected behavior

When I don't pass in `constant_liar=True` this warning doesn't show up, and I don't think having `constant_liar=True` should be warning either. Also when I rerun the same code it no longer warns.

### Environment

- Optuna version:3.5.0
- Python version:3.10.13
- OS:Linux-5.15.102

```shell
/usr/.../lib/python3.10/site-packages/optuna/samplers/_tpe/sampler.py:746: UserWarning:
Trial 10 does not have constraint values. It will be treated as a lower priority than other trials.
...
Trial 199 does not have constraint values. It will be treated as a lower priority than other trials.
```

Strangely, Trials 0-9 do not have any warnings.

### Steps to reproduce

```python
import optuna

def objective(trial):
    # Binh and Korn function with constraints.

    # Constraints which are considered feasible if less than or equal to zero.
    # The feasible region is basically the intersection of a circle centered at (x=5, y=0)
    # and the complement to a circle centered at (x=8, y=-3).
    c0 = (x - 5) ** 2 + y**2 - 25
    c1 = -((x - 8) ** 2) - (y + 3) ** 2 + 7.7

    # Store the constraints as user attributes so that they can be restored after optimization.
    trial.set_user_attr("constraint", (c0, c1))

    v0 = 4 * x**2 + 4 * y**2

    return v0, v1


def constraints(trial):

sampler = optuna.samplers.TPESampler(constraints_func=constraints, constant_liar=True)
study = optuna.create_study(
    sampler=sampler,
)

# distributed optimization
batch_size = 10
rounds = 20
for round_idx in range(rounds):
    objs = []
    trials = []
    for batch_idx in range(batch_size):
        trials.append((trial := study.ask()))
        objs.append(objective(trial))

    for obj, trial in zip(objs, trials):
        study.tell(trial, obj)
```

not522:
## Motivation
Fix #5291.
The constraint values are set in `after_trial`, then `RUNNING` trials don't have them. It causes unexpected warnings (#5291).

## Description of the changes
Fix  `_split_trials` to check if the state of the trial is `RUNNING` before the feasibility check.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -620,14 +620,16 @@ def _split_trials(
     infeasible_trials = []
 
     for trial in trials:
-        if constraints_enabled and _get_infeasible_trial_score(trial) > 0:
+        if trial.state == TrialState.RUNNING:
+            # We should check if the trial is RUNNING before the feasibility check
+            # because its constraint values have not yet been set.
+            running_trials.append(trial)
+        elif constraints_enabled and _get_infeasible_trial_score(trial) > 0:
             infeasible_trials.append(trial)
         elif trial.state == TrialState.COMPLETE:
             complete_trials.append(trial)
         elif trial.state == TrialState.PRUNED:
             pruned_trials.append(trial)
-        elif trial.state == TrialState.RUNNING:
-            running_trials.append(trial)
         else:
             assert False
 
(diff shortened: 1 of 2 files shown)


HideakiImamura:

HideakiImamura (approved):
LGTM.

eukaryo (approved):
LGTM!

Issue #5203, opened 2024-01-26
Speed up `_hypervolume.WFG`


HideakiImamura:
### What is an issue?

Optuna has implemented the _hypervolume.WFG class for calculating hypervolume, but this implementation is naive and slow. Speeding this up would be of great benefit in multi-objective optimization. For example, https://ieeexplore.ieee.org/document/1583625 could be a candidate for the algorithm.

nabenabe0928:
Just in case, I am already working on a faster implementation of WFG.
https://github.com/nabenabe0928/optuna/compare/enhance/speed-up-non-dominated-sort...nabenabe0928:optuna:enhance/speed-up-wfg

nabenabe0928:
As mentioned [here](https://github.com/optuna/optuna/pull/5504#issuecomment-2196548251),  hypervolume computation for 3D or more requires a special data structure.

Here is an example of HV3D.

https://github.com/optuna/optuna/compare/master...nabenabe0928:add-hv3d?expand=1

nabenabe0928:
Once we introduce the box decomposition, we can replace the WFG with it for a small number of objectives (up to 4 or 5 objectives)

- https://github.com/optuna/optuna/compare/master...nabenabe0928:temp/add-log-ehvi?expand=1

PR #5303, opened 2024-03-11
Make 2d HV computation twice faster

nabenabe0928:
# Motivation

As MOTPE is still not quick, I replaced the current Hypervolume 2d computation with numpy vectorization.
By using vectorization, we get a significant speedup.
My implementation halves the runtime for the 2d MOTPE experiment below.

This PR partially relates to:
- https://github.com/optuna/optuna/issues/5203

# Description of the changes

I simply replaced the current implementation with the equivalent vectorization implementation.

# Benchmark

<details>

```python

import time

import optuna


def objective2d(trial: optuna.Trial) -> tuple[float, float]:
    return x ** 2 + y ** 2, (x - 2) ** 2 + (y - 2) ** 2


def objective3d(trial: optuna.Trial) -> tuple[float, float]:


if __name__ == "__main__":
    start = time.time()
    n_trials = 1000
    study = optuna.create_study(sampler=sampler, directions=["minimize"]*2)
    study.optimize(objective2d, n_trials=n_trials)
    print(f"n_obj=2, {study.sampler}", time.time() - start)
```

</details>

This PR finishes it in about 80 seconds while the original one takes 170 seconds in my local environment.

# Verification

To verify the new implementation, I compared the numerical outputs with the original implementation and my implementation passed the pure comparison with the original implementation.

<details>

```python
import time

import numpy as np


def _compute_2points_volume(point1: np.ndarray, point2: np.ndarray) -> float:
    """Compute the hypervolume of the hypercube, whose diagonal endpoints are given 2 points.

    Args:
        point1:
            The first endpoint of the hypercube's diagonal.
        point2:
            The second endpoint of the hypercube's diagonal.
    """

    return float(np.abs(np.prod(point1 - point2)))


def original(solution_set: np.ndarray, reference_point: np.ndarray) -> float:
    """Compute the hypervolume for the two-dimensional space.

    This algorithm divides a hypervolume into
    smaller rectangles and sum these areas.

    Args:
        solution_set:
            The solution set which we want to compute the hypervolume.
        reference_point:
            The reference point to compute the hypervolume.
    """

    sorted_solution_set = solution_set[np.lexsort((-solution_set[:, 1], solution_set[:, 0]))]

    hypervolume = 0.0
    for solution in sorted_solution_set:
        if reference_point[1] < solution[1]:
            continue

        # Compute an area of a rectangle with reference_point
        # and one point in solution_sets as diagonals.
        hypervolume += _compute_2points_volume(reference_point, solution)

        # Update a reference_point to create a new hypervolume that
        # exclude a rectangle from the current hypervolume
        reference_point[1] = solution[1]

    return hypervolume


def this_pr(solution_set: np.ndarray, reference_point: np.ndarray) -> float:
    """Compute the hypervolume for the two-dimensional space.

    This algorithm divides a hypervolume into
    smaller rectangles and sum these areas.

    Args:
        solution_set:
            The solution set which we want to compute the hypervolume.
        reference_point:
            The reference point to compute the hypervolume.
    """
    # Descending order in the 1st objective, and ascending order in the 2nd objective.
    sorted_solution_set = solution_set[np.lexsort((-solution_set[:, 1], solution_set[:, 0]))]
    reference_points_y = np.append(reference_point[1], sorted_solution_set[:-1, 1])
    reference_points_y_cummin = np.minimum.accumulate(reference_points_y)
    mask = sorted_solution_set[:, 1] <= reference_points_y_cummin

    used_solution = sorted_solution_set[mask]
    edge_length_x = reference_point[0] - used_solution[:, 0]
    edge_length_y = reference_points_y_cummin[mask] - used_solution[:, 1]
    return edge_length_x @ edge_length_y


if __name__ == "__main__":
    rng = np.random.RandomState(42)
    for _ in range(30):
        x = rng.normal(size=(100000, 2))
        reference_point = np.max(x, axis=0) * 1.1
        start = time.time()
        ans = original(x.copy(), reference_point.copy())
        print("Original", time.time() - start)
        start = time.time()
        output = this_pr(x.copy(), reference_point.copy())
        print("This PR", time.time() - start)
        assert np.isclose(ans, output), f"{ans=}, {output=}"
```
</details>

### DIFF
--- a/optuna/_hypervolume/utils.py
+++ b/optuna/_hypervolume/utils.py
@@ -27,19 +27,13 @@ def _compute_2d(solution_set: np.ndarray, reference_point: np.ndarray) -> float:
             The reference point to compute the hypervolume.
     """
 
+    # Descending order in the 1st objective, and ascending order in the 2nd objective.
     sorted_solution_set = solution_set[np.lexsort((-solution_set[:, 1], solution_set[:, 0]))]
+    reference_points_y = np.append(reference_point[1], sorted_solution_set[:-1, 1])
+    reference_points_y_cummin = np.minimum.accumulate(reference_points_y)
+    mask = sorted_solution_set[:, 1] <= reference_points_y_cummin
 
-    hypervolume = 0.0
-    for solution in sorted_solution_set:
-        if reference_point[1] < solution[1]:
-            continue
-
-        # Compute an area of a rectangle with reference_point
-        # and one point in solution_sets as diagonals.
-        hypervolume += _compute_2points_volume(reference_point, solution)
-
-        # Update a reference_point to create a new hypervolume that
-        # exclude a rectangle from the current hypervolume
-        reference_point[1] = solution[1]
-
-    return hypervolume
+    used_solution = sorted_solution_set[mask]
+    edge_length_x = reference_point[0] - used_solution[:, 0]
+    edge_length_y = reference_points_y_cummin[mask] - used_solution[:, 1]
+    return edge_length_x @ edge_length_y
--- a/optuna/_hypervolume/utils.py
+++ b/optuna/_hypervolume/utils.py
@@ -34,6 +34,6 @@ def _compute_2d(solution_set: np.ndarray, reference_point: np.ndarray) -> float:
     mask = sorted_solution_set[:, 1] <= reference_points_y_cummin
 
     used_solution = sorted_solution_set[mask]
-    edge_length_x = np.abs(reference_point[0] - used_solution[:, 0])
-    edge_length_y = np.abs(reference_points_y_cummin[mask] - used_solution[:, 1])
+    edge_length_x = reference_point[0] - used_solution[:, 0]
+    edge_length_y = reference_points_y_cummin[mask] - used_solution[:, 1]
     return edge_length_x @ edge_length_y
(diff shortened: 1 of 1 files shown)


nabenabe0928:
@y0z @HideakiImamura Could you review this PR?
Note that the CI is failing because of `fakeredis`, which will be fixed in #5307 .

HideakiImamura (approved):
I confirmed that the logic is valid. Thank you for the speed up! LGTM!

optuna/_hypervolume/utils.py
```diff
@@ -27,19 +27,13 @@ def _compute_2d(solution_set: np.ndarray, reference_point: np.ndarray) -> float:
             The reference point to compute the hypervolume.
     """
```
y0z, line 29: Why don't you add assertions like the ones below?
```python
    assert solution_set.shape[1] == 2
    assert reference_point.shape[0] == 2
```

y0z (commented):
NIT

I find that the current implementation of `_compute_2d` updates the provided `reference_point` (i.e. calling `_compute_2d` twice results in different values) whereas this PR's one does not.
I think the latter's behavior is generally preferable unless there is no side effect to other functionalities in Optuna.
https://github.com/optuna/optuna/blob/master/optuna/_hypervolume/utils.py#L43

After checking the implementation of `WFG.compute` that utilizes `_compute_2d`, I conclude that this behavior change does not affect the result of hypervolume computation.
https://github.com/optuna/optuna/blob/master/optuna/_hypervolume/wfg.py#L23-L31

nabenabe0928:
@y0z I addressed your comment:)

y0z (approved):
LGTM

PR #5306, opened 2024-03-11
Add `is_exhausted()` function in the `GridSampler` class

### LINKED ISSUE #5052: Feature to determine whether `GridSampler` has exhausted the grid
### Motivation

Currently, there is no good way in which we can determine whether the grid of `GridSampler` is exhausted when resuming a study.
While the method `_get_unvisited_grid_ids()` exists, it is a protected method and should not be used outside of the library.

### Description

I suggest adding a method called something like `is_exhausted()` that is public and simply returns `len(self._get_unvisited_grid_ids()) == 0`

### Alternatives (optional)

Alternatively, it may be useful to add keyword arguments to be passed to `GridSampler` that give more control over what happens if the grid is already exhausted. For example, one could add a keyword like `reevaluate_when_exhausted: bool`.

I encountered this suggestion when receiving the following warning: `GridSampler` is re-evaluating a configuration because the grid has been exhausted. This may happen due to a timing issue during distributed optimization or when re-running optimizations on already finished studies."

This issue further builds upon the discussion of https://github.com/optuna/optuna/issues/3256

aaravm:
## Motivation

fixes #5052

## Description of the changes

Added the is_exhausted() function, where it returns the bool value of len(self._get_unvisited_grid_ids()) == 0, as suggested by the issue.
<br>
Additional question: Should I also add reevaluate_when_exhausted argument?


nabenabe0928:

> Additional question: Should I also add reevaluate_when_exhausted argument?


@aaravm 
Thank you for the contribution:)
Let's keep this PR small and simple for now!

nabenabe0928:
Probably because you merged your repository's master, but not the upstream master.
You need to fetch the latest master head in optuna/optuna, but not aaravm/optuna.

HideakiImamura (commented):
Basically, LGTM. Could you add a document as a docstring and a test case in `tests/samplers_tests/test_grid.py`?

aaravm:
I have added the following test in test_grid.py:
```
def test_is_exhausted() -> None:
        a = trial.suggest_int("a", 0, 100)
        b = trial.suggest_float("b", -100, 100)
        return a * b
    search_space_0 = {"a": [0, 50], "b": [-50, 0, 50]}
    sampler_0 = samplers.GridSampler(search_space_0)
    study = optuna.create_study(sampler=sampler_0)
    assert not sampler_0.is_exhausted(study)

```    
   However, when I run pytest test_grid.py, I get the following error:
   

> TypeError: GridSampler.is_exhausted() takes 1 positional argument but 2 were given

   
   Can anyone suggest what the issue might be?

nabenabe0928:
@aaravm 
How can I reproduce your issue?
Are you really sure that everything was coded as provided?

aaravm:
I changed the function from before and added the tests and the docstring.

optuna/samplers/_grid.py
```diff
@@ -277,3 +277,9 @@ def _same_search_space(self, search_space: Mapping[str, Sequence[GridValueType]]
                     return False
 
         return True
+
+    def is_exhausted(self, study: Study) -> bool:
+        """
+        Returns True if the grid of GridSampler is exhausted, else returns False
```
nabenabe0928, line 283: [nit] It is better to describe the function more clearly.
```suggestion
        Return True if all the possible params are evaluated, otherwise return False.
```

nabenabe0928 (commented):
Thank you for the changes! I left a minor comment:)

nabenabe0928 (approved):
LGTM!

eukaryo (approved):
LGTM. I'll proceed with merging it.

PR #5311, opened 2024-03-13
Make each importance evaluator compatible with doc

nabenabe0928:
## Motivation

In Optuna v3.0, `normalize` option was added to the importance evaluator (see [here](https://github.com/optuna/optuna/blob/master/optuna/importance/__init__.py#L90-L93)):
```
 A boolean option to specify whether the sum of the importance values should be normalized to 1.0. Defaults to True.
```

However, this option does not do anything for f-ANOVA and PED-ANOVA, because they normalize in their `evaluate` method.
For this reason, I removed the normalization for each implementation.

## Description of the changes

I removed the normalization parts from `FanovaImportanceEvaluator` and `PedAnovaImportanceEvaluator` and modified the corresponding tests.

### DIFF
--- a/optuna/importance/_fanova/_evaluator.py
+++ b/optuna/importance/_fanova/_evaluator.py
@@ -126,7 +126,6 @@ class FanovaImportanceEvaluator(BaseImportanceEvaluator):
         param_importances = numpy.array(
             [evaluator.get_importance(i)[0] for i in range(len(non_single_distributions))]
         )
-        param_importances /= numpy.sum(param_importances)
 
         return _sort_dict_by_importance(
             {
--- a/optuna/importance/_fanova/_evaluator.py
+++ b/optuna/importance/_fanova/_evaluator.py
@@ -126,7 +126,7 @@ class FanovaImportanceEvaluator(BaseImportanceEvaluator):
         param_importances = numpy.array(
             [evaluator.get_importance(i)[0] for i in range(len(non_single_distributions))]
         )
-        #Normalize is here to keep the backward compatibility.
+        # We normalize here to keep the backward compatibility.
         param_importances /= numpy.sum(param_importances)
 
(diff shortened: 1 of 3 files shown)


nabenabe0928:
@HideakiImamura @eukaryo Could you review this PR?

optuna/importance/_fanova/_evaluator.py
```diff
@@ -126,6 +126,8 @@ def evaluate(
         param_importances = numpy.array(
             [evaluator.get_importance(i)[0] for i in range(len(non_single_distributions))]
         )
+        # NOTE(nabenabe0928): Normalize is here to keep the backward compatibility.
+        # TODO(nabenabe0928): Add the `normalize` argument optionally.
```
HideakiImamura, line 130: I agree that there doesn't seem to be a particular consensus on this, shall we delete this comment?

HideakiImamura (commented):
Thanks for the PR. Basically, LGTM. I have minor comments. PTAL.

optuna/importance/_fanova/_evaluator.py
nabenabe0928:
```suggestion
```

nabenabe0928:
@HideakiImamura thank your for your comment! I applied your suggestion:)

optuna/importance/_fanova/_evaluator.py
```diff
@@ -126,6 +126,7 @@ def evaluate(
         param_importances = numpy.array(
             [evaluator.get_importance(i)[0] for i in range(len(non_single_distributions))]
         )
+        #Normalize is here to keep the backward compatibility.
```
nabenabe0928, line 129:
```suggestion
        # We normalize here to keep the backward compatibility.
```

HideakiImamura (approved):
LGTM.

PR #5315, opened 2024-03-15
Fix doc example in WilcoxonPruner

contramundum53:
This PR fixes doc example in `WilcoxonPruner`.


eukaryo (approved):
LGTM. I will merge after CI passes.

PR #5320, opened 2024-03-15
Replaced `import List`, `import Union`  with `__future__.annotations`

keita-sa:
## Motivation
I want to contribute to Optuna's OSS development and decided to work on the `contribution-welcome` tasks on a trial basis.

Among them, #4508 looks a simple task to just add `__future__.annotations` and change the notation of `List`, `Tuple`, `Dict`, `Optional`, and `Union`, so I've committed it.

## Description of the changes
Added:
`from __future__ import annotations`

Deleted:
`from typing import List`
`from typing import Union`

Changed:
`List`; 3 places
`Union`; 1 place
`Optional`; 1 place

*To avoid **Checks / checks** failure, which was not successful in the previous pull request, I've installed a formatter black with `pip install black`.

### DIFF
--- a/optuna/pruners/_hyperband.py
+++ b/optuna/pruners/_hyperband.py
@@ -1,9 +1,9 @@
+from __future__ import annotations
+
 import binascii
 import math
 from typing import Container
-from typing import List
 from typing import Optional
-from typing import Union
 
 import optuna
 from optuna import logging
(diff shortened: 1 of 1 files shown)


aaravm:
I think that you can run `./formats.sh` and then commit it. I think then all your checks will pass

keita-sa:
Thanks for your direction, I'll commit it again. @aaravm

PR #5324, opened 2024-03-15
Updated checks.yml

keita-sa:
## Motivation
I encountered a failure with the **black** formatter in **Checks / checks** when I replaced `import List`, etc. with `__future__.annotations`. According to the other's blog post, it seems to be caused by the descriptions in `checks.yml`, so can I make a slight addition?

I'm not sure if it will work.

スクリーンショット 2024-03-15 19 33 51

## Description of the changes
```
 - uses: psf/black@stable
   with:
     src: "./"
```

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -24,6 +24,10 @@ jobs:
       with:
         python-version: 3.8
 
+    - uses: psf/black@stable
+      with:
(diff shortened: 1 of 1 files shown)


nzw0301:
I suppose you misunderstand his suggestion. This PR is not necessary.

PR #5323, opened 2024-03-15
Replaced `import List`, `import Union`  with `__future__.annotations`

keita-sa:
## Motivation
I want to contribute to Optuna's OSS development and decided to work on the `contribution-welcome` tasks on a trial basis.

Among them, #4508 looks a simple task to just add `__future__.annotations` and change the notation of `List`, `Tuple`, `Dict`, `Optional`, and `Union`, so I've committed it.

## Description of the changes
Added:
`from __future__ import annotations`

Deleted:
`from typing import List`
`from typing import Union`

Changed:
`List`; 3 places
`Union`; 1 place
`Optional`; 1 place
 
*To avoid **Checks / checks** failure, I've run `./formats.sh` as @aaravm informed.

PR #5326, opened 2024-03-16
Replaced `import List, Union` with `__future__.annotations`

keita-sa:
## Motivation
Since the previous pull request for #5323 failed with **Checks / checks**, so I would introduced `__future__.annotations` in another py.file as a trial.

## Description of the changes
Added:
`from __future__ import annotations`

Deleted:
`from typing import List`
`from typing import Union`

Changed:
List; 5 places
Union; 1 place
Optional; 2 places

### DIFF
--- a/optuna/pruners/_hyperband.py
+++ b/optuna/pruners/_hyperband.py
@@ -1,9 +1,9 @@
+from __future__ import annotations
+
 import binascii
 import math
 from typing import Container
-from typing import List
 from typing import Optional
-from typing import Union
 
 import optuna
 from optuna import logging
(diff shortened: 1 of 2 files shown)


optuna/pruners/_hyperband.py
```diff
@@ -1,9 +1,9 @@
+from __future__ import annotations
+
 import binascii
 import math
 from typing import Container
```
nzw0301, line 5: I think this should be `collections.abc.Container` rather than `typing.Container` as explained by the original issue.

optuna/pruners/_hyperband.py
```diff
@@ -1,9 +1,9 @@
...
-from typing import List
 from typing import Optional
```
nzw0301, line 6: Could you use `ANOTER_TYPE | None` instead of `optinal[ANOTHER_TYPE]` to address the original issue?
https://peps.python.org/pep-0604/#proposal

```suggestion
```

optuna/pruners/_successive_halving.py
```diff
@@ -1,7 +1,7 @@
+from __future__ import annotations
+
 import math
-from typing import List
 from typing import Optional
```
nzw0301, line 4: ditto

nzw0301 (changes requested):

PR #5330, opened 2024-03-17
Replaced `typing.Optional`, added `collections.abc.Container`, applied `isort`

keita-sa:
## Motivation
I've updated the notation of `Optional`, also added `collections.abc.Container`. Applied `isort`.

## Description of the changes
Added:
`__future__.annotations`
`collections.abc.Container`

Deleted:
`typing.List`
`typing.Union`
`typing.Optional`
`typing.Container`

### DIFF
--- a/optuna/pruners/_hyperband.py
+++ b/optuna/pruners/_hyperband.py
@@ -1,9 +1,9 @@
+from __future__ import annotations
+
 import binascii
 import math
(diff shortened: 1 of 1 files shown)


keita-sa:
If we use `black` and `isort` together, we may need to edit `.toml`, so I will do that later on.

PR #5339, opened 2024-03-18
Add the explanation of how to sample the best expected params to FAQ

nabenabe0928:
## Motivation

As some people would like to get the best parameters that are not affected by the random noise effect, I added the guide on how to sample the expected best parameters using the `GPSampler` introduced in v3.6.0.

## Description of the changes

I added the description of the way to sample the best expected parameters.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -715,11 +715,12 @@ Optuna may sometimes suggest parameters evaluated in the past and if you would l
     study = optuna.create_study()
     study.optimize(objective, n_trials=100)
 
-How can I sample the expected best parameters based on the evaluated trials?
+How can I sample the best expected parameters based on the evaluated trials?
 ----------------------------------------------------------------------------
 
 As the optimization often overtunes the parameters of an objective function with an observational noise, it is convenient to sample the best parameters based on the debiased objective function.
-:class:`~optuna.samplers.GPSampler` enables us to sample the best-expected parameters using the mean estimation of parameters.
+:class:`~optuna.samplers.GPSampler` enables us to sample the best expected parameters using the mean estimation of parameters.
+Note that the best expected parameters are defined as :math:`\mathrm{arg}\min_{x \in X} \mathbb{E}_{f}[f(x)]` where :math:`f` follows the Gaussian process.
 However, as we made the interface of :class:`~optuna.samplers.GPSampler` as simple as possible, we need to tweak it as follows:
 
 .. code-block:: python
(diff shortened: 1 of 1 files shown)


eukaryo:
@contramundum53 @HideakiImamura Could you review this PR?

docs/source/faq.rst
```diff
@@ -714,3 +714,66 @@ Optuna may sometimes suggest parameters evaluated in the past and if you would l
 
     study = optuna.create_study()
     study.optimize(objective, n_trials=100)
+
+How can I sample the best expected parameters based on the evaluated trials?
```
contramundum53, line 718: I think some people need the parameters without sampling a new trial, so maybe it's beneficial to introduce the non-sampler way that directly uses `_gp` classes?

contramundum53: Also it should be pointed out that this is a "hack" and it is not officially supported. It is possible that the interface changes in the future.

docs/source/faq.rst
```diff
@@ -714,3 +714,66 @@ Optuna may sometimes suggest parameters evaluated in the past and if you would l
...
+----------------------------------------------------------------------------
+
+As the optimization often overtunes the parameters of an objective function with an observational noise, it is convenient to sample the best parameters based on the debiased objective function.
+:class:`~optuna.samplers.GPSampler` enables us to sample the best expected parameters using the mean estimation of parameters.
+Note that the best expected parameters are defined as :math:`\mathrm{arg}\min_{x \in X} \mathbb{E}_{f}[f(x)]` where :math:`f` follows the Gaussian process.
```
nzw0301, line 723: Is $X$ is a set of suggested params in a trial and $x$ is a one param in the suggested params as in the example below? Can we clarify $X$ and $x$ here?

docs/source/faq.rst
```diff
@@ -714,3 +714,66 @@ Optuna may sometimes suggest parameters evaluated in the past and if you would l
...
+            acqf_params = acqf.AcquisitionFunctionParams(**acqf_params_kwargs)
+            # Optimize the mean estimator. (Use self._rng.rng optionally for reproducibility.)
+            best_normalized_params, _ = optimize_acqf_mixed(acqf_params)
+            return best_normalized_params
+
+
+    def objective(trial, evaluate_func=True):
+        X = np.array([trial.suggest_float(f"x{i}", -5, 5) for i in range(2)])
+        if evaluate_func:
+            # Could be a very expensive evaluation in practice such as a training of DNN.
```
nzw0301, line 753: I agree that DNN is a widely used abbreviation, but `deep neural networks` sounds more reader-friendly.
```suggestion
            # Could be a very expensive evaluation in practice such as a training of deep neural networks.
```

nzw0301 (commented):
Let me leave a minor comments. I also think some references to gaussian process would be helpful to understand the technical terms, such as UCB and acquisition function, especially for users who are not familiar with gaussian process.

HideakiImamura:

eukaryo (commented):
LGTM if the current comments are resolved.

HideakiImamura:
@nabenabe0928 Please address @nzw0301's comments.

Issue #5285, opened 2024-02-29
Introduce lazy greedy algorithm to `_solve_hssp`


Alnusjaponica:
### Motivation

Practically speed-up `_solve_hssp()` using lazy greedy algorithm.

### Description

The current implementation of [`_solve_hssp()`](https://github.com/optuna/optuna/blob/2f279f321664bf8793914f442ca13b1b97495b2b/optuna/_hypervolume/hssp.py#L8) relies on the greedy algorithm for submodular maximization problems with a cardinality constraint. the greedy algorithm guarantees the best approximation ratio of $1-1/e$ (refer to https://doi.org/10.1007/BFb0121195) for monotone cases and obtains practically good solution for non-monotone cases.

To further enhance performance, I propose introducing the lazy greedy algorithm (see https://doi.org/10.1007/BFb0006528) to the hypervolume subset selection problems. This extension allows the algorithm to strategically delay oracle calls until necessary, leveraging the submodular property for practical speed improvements over the naive greedy algorithm. Notably, the number of oracle calls is maintained at most equivalent to the original greedy algorithm. 

This straightforward combination of contribution calculation and lazy strategy has been experimentally validated in https://doi.org/10.1109/CEC48606.2020.9185878.

### Alternatives (optional)

We need to examine benchmark results to judge whether the lazy strategy really accelerates `_solve_hssp()`. This is because lazy greedy has the overhead of heap operations and can be slow depending on the input.

abant07:
Hey would I be able to work on this. Just out of curiosity however, would this deal with backend more?

Also, is there anything I need to know when reproducing the environment other than just following the ReadME?

Thanks

Alnusjaponica:
Thank you very much for expressing your interest in contributing to our repository.

To get started, I recommend reviewing [CONTRIBUTING.md](https://github.com/optuna/optuna/blob/master/CONTRIBUTING.md), which outlines guidelines and instructions for initiating contributions.

Regarding this issue, it involves understanding the algorithm's functionality and comparing benchmark results before and after. We are going to decide whether to use a lazy greedy algorithm based on these comparisons. Consequently, it might seem overwhelming for those who haven't contributed to our repositories before, unless this particular issue aligns well with your preferences.

If you're enthusiastic about contributing, we have several issues tagged with "contribution-welcome." These are specifically designed for newcomers, offering a smooth entry into the project. Feel free to choose any of these issues to begin your journey.

abant07:
[...]

PR #5355, opened 2024-03-23
Make HV calculations in HSSP lazy to speed up

nabenabe0928:
## Motivation

As the HSSP for more than two objectives is still very slow, I introduce a lazy evaluation of HV calculations.
This PR relates to the following issue:
- https://github.com/optuna/optuna/issues/5285

## Description of the changes

This PR speeds up HSSP by using the following property:
- contributions for each point can always be calculated by an HV calculation,
- it is unnecessary to calculate HVs if we already know they will not be the maximum contribution in the next iteration, and
- we update the hypervolume from the ones with higher upper bound based on the submodularity.

The new algorithm does not change the output of this function, so the results obtained by TPESampler will also be preserved after this PR is merged. 

## Mathematical Background for Submodularity

When we define $S \subseteq T \subseteq [N] \coloneqq \\{1,2,...,N\\}$ and the hypervolume measure as $H(S)$, the hypervolume measure is known to have the submodularity, i.e. $H(S \cup \\{i\\}) - H(S) \geq H(T \cup \\{i\\}) - H(T)$ holds for $i \in [N] \setminus T$.

At each iteration, we pick an index of a point $i \in \max_{i^\prime \in [N] \setminus S} H(S \cup \\{i^\prime\\}) - H(S)$ and increment $S$ to $T \coloneqq S \cup \\{i\\}$, the problem here is that the calculation of $H(S \cup \\{i^\prime\\})$ is expensive and we would like to avoid unnecessary calculations as much as possible.

Once we start to calculate, every single $H(S \cup \\{i\\})$ for each $i$ will be recorded every time they are calculated for some $S^\prime \subseteq S \subseteq [N]$.
However, if there exists an index $i \in [N] \setminus S$ such that $H(S \cup \\{i\\}) - H(S) \geq H(S^\prime \cup \\{j\\}) - H(S^\prime) \geq H(S \cup \\{j\\}) - H(S)$ for $j \in [N] \setminus S$, it is unnecessary to calculate $H(S \cup \\{j\\})$ for now.
In principle, we can skip some calculations based on the submodularity as discussed above.

Furthermore, we start to calculate $H(S \cup \\{i\\})$ with $i$ that has a larger upper bound $H(S^\prime \cup \\{i\\})$ because we are likely to end up skipping more calculations.

### DIFF
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -30,13 +30,14 @@ def _solve_hssp(
         max_index = int(np.argmax(contribs))
         selected_indices[k] = indices[max_index]
         selected_vecs[k] = rank_i_loss_vals[indices[max_index]].copy()
-        keep = contribs >= 0.0
+        keep = np.ones(contribs.size, dtype=bool)
+        keep[max_index] = False
         contribs = contribs[keep]
         indices = indices[keep]
         if k == subset_size - 1:
             break
 
-        hv_selected = optuna._hypervolume.WFG().compute(selected_vecs[:k + 1], reference_point)
+        hv_selected = optuna._hypervolume.WFG().compute(selected_vecs[: k + 1], reference_point)
         max_contrib = 0.0
         index_from_larger_contrib = np.argsort(-contribs)
         for i in index_from_larger_contrib:
@@ -46,7 +47,7 @@ def _solve_hssp(
                 continue
 
             selected_vecs[k + 1] = rank_i_loss_vals[indices[i]].copy()
-            hv_plus = optuna._hypervolume.WFG().compute(selected_vecs[:k + 2], reference_point)
+            hv_plus = optuna._hypervolume.WFG().compute(selected_vecs[: k + 2], reference_point)
             contribs[i] = hv_plus - hv_selected
             max_contrib = max(contribs[i], max_contrib)
 
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -58,26 +58,30 @@ def _solve_hssp(
         return rank_i_indices
 
     assert not np.any(reference_point - rank_i_loss_vals <= 0)
+    # Take unique to avoid a duplication error in case rank_i_loss_vals has `inf`or `-inf`.
+    rank_i_unique_loss_vals, indices_of_unique_loss_vals = np.unique(
+        rank_i_loss_vals, return_index=True, axis=0
+    )
     n_objectives = reference_point.size
-    contribs = np.prod(reference_point - rank_i_loss_vals, axis=-1)
+    contribs = np.prod(reference_point - rank_i_unique_loss_vals, axis=-1)
     selected_indices = np.zeros(subset_size, dtype=int)
     selected_vecs = np.empty((subset_size, n_objectives))
-    indices = np.arange(rank_i_indices.size, dtype=int)
+    indices = np.arange(rank_i_unique_loss_vals.shape[0], dtype=int)
     for k in range(subset_size):
         max_index = int(np.argmax(contribs))
         selected_indices[k] = indices[max_index]
-        selected_vecs[k] = rank_i_loss_vals[max_index].copy()
+        selected_vecs[k] = rank_i_unique_loss_vals[max_index].copy()
         keep = np.ones(contribs.size, dtype=bool)
         keep[max_index] = False
         contribs = contribs[keep]
         indices = indices[keep]
-        rank_i_loss_vals = rank_i_loss_vals[keep]
+        rank_i_unique_loss_vals = rank_i_unique_loss_vals[keep]
         if k == subset_size - 1:
             # We do not need to update contribs at the last iteration.
             break
 
         contribs = _lazy_contribs_update(
-            contribs, rank_i_loss_vals, selected_vecs[: k + 2], reference_point
+            contribs, rank_i_unique_loss_vals, selected_vecs[: k + 2], reference_point
         )
 
-    return rank_i_indices[selected_indices]
+    return rank_i_indices[indices_of_unique_loss_vals[selected_indices]]
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -5,6 +5,39 @@ import numpy as np
 import optuna
 
 
+def _lazy_contribs_update(
+    contribs: np.ndarray,
+    pareto_loss_values: np.ndarray,
+    selected_vecs: np.ndarray,
+    reference_point: np.ndarray,
+) -> np.ndarray:
+    """Lazy update the hypervolume contributions.
+
+    S=selected_indices \ {indices[max_index]}, T=selected_indices, and S' is a subset of S.
+    As we would like to know argmax H(T v {i}) in the next iteration, we can skip HV
+    calculations for j if H(T v {i}) - H(T) > H(S' v {j}) - H(S') >= H(T v {j}) - H(T).
+    We used the submodularity for the inequality above. As the upper bound of contribs[i] is
+    H(S' v {j}) - H(S'), we start to update from i with a higher upper bound so that we can
+    skip more HV calculations.
+    """
+    hv_selected = optuna._hypervolume.WFG().compute(selected_vecs[:-1], reference_point)
+    max_contrib = 0.0
+    index_from_larger_upper_bound_contrib = np.argsort(-contribs)
+    for i in index_from_larger_upper_bound_contrib:
+        if contribs[i] < max_contrib:
+            # Lazy evaluation to reduce HV calculations.
+            # If contribs[i] will not be the maximum next, it is unnecessary to compute it.
+            continue
+
+        selected_vecs[-1] = pareto_loss_values[i].copy()
+        hv_plus = optuna._hypervolume.WFG().compute(selected_vecs, reference_point)
+        # inf - inf in the contribution calculation is always inf.
+        contribs[i] = hv_plus - hv_selected if not np.isinf(hv_plus) else np.inf
+        max_contrib = max(contribs[i], max_contrib)
+
+    return contribs
+
+
 def _solve_hssp(
     rank_i_loss_vals: np.ndarray,
     rank_i_indices: np.ndarray,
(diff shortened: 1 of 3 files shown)


optuna/_hypervolume/hssp.py
```diff
@@ -1,10 +1,43 @@
...
 
+def _lazy_contribs_update(
+    contribs: np.ndarray,
+    pareto_loss_values: np.ndarray,
+    selected_vecs: np.ndarray,
+    reference_point: np.ndarray,
+) -> np.ndarray:
+    """Lazy update the hypervolume contributions.
+
+    S=selected_indices \ {indices[max_index]}, T=selected_indices, and S' is a subset of S.
```
nabenabe0928, line 16:
```suggestion
    S=selected_indices - {indices[max_index]}, T=selected_indices, and S' is a subset of S.
```

gen740 (commented):
I tested with the Viennet function, which has three objective functions. (ref. https://en.wikipedia.org/wiki/Test_functions_for_optimization#Test_functions_for_multi-objective_optimization)

$$
\begin{align*}
f_1(x) &= 0.5 \left( x_1^2 + x_2^2 \right) + \sin \left( x_1^2 + x_2^2 \right) \\
f_2(x) &= \frac{(3x_1 - 2x_2 + 4)^2}{8} + \frac{(x_1 - x_2 + 1)^2}{27} + 15 \\
f_3(x) &= \frac{1}{x_1^2 + x_2^2 + 1} - 1.1 \exp \left( -x_1^2 - x_2^2 \right)
\end{align*}
$$

## Test code
```python
import optuna
import numpy as np

def objective(trial):
    x = trial.suggest_float("x", -3, 3)
    y = trial.suggest_float("y", -3, 3)
    f1 = 0.5 * (x**2 + y**2) + np.sin(x**2 + y**2)
    f2 = (3 * x - 2 * y + 4) ** 2 / 8 + (x - y + 1) ** 2 / 27 + 15
    f3 = 1 / (x**2 + y**2 + 1) - 1.1 * np.exp(-(x**2) - y**2)
    return f1, f2, f3


sampler = optuna.samplers.TPESampler(seed=0)
study = optuna.create_study(
    directions=["minimize", "minimize", "minimize"], sampler=sampler
)
study.optimize(objective, n_trials=200)

for trial in study.best_trials:
    print(f"{trial.values} {trial.params}")
```

I confirmed that the answer is identical.
For reference, I ran some benchmarks with trials = 20, 100, 200 and took the average and standard deviation.

| number of trials | old | new |
| --- | --- | --- |
| 20 (24runs) | 114.0 ms ±   0.8 ms | 113.7 ms ±   1.3 ms |
| 100 (10runs) | 1.437 s ±  0.024 s | 700.9 ms ±   8.5 ms |
| 200 (10runs) | 22.914 s ±  0.091 s | 5.190 s ±  0.023 s |

nabenabe0928:
```python
import optuna
import numpy as np

def objective(trial):
    x = trial.suggest_float("x", -3, 3)
    y = trial.suggest_float("y", -3, 3)
    f1 = 0.5 * (x**2 + y**2) + np.sin(x**2 + y**2)
    f4 = np.inf if -1 < x < 1 else x * y
    return f1, f4


sampler = optuna.samplers.TPESampler(seed=0)
study = optuna.create_study(directions=["minimize"]*2, sampler=sampler)
study.optimize(objective, n_trials=200)

for trial in study.best_trials:
    print(f"{trial.values} {trial.params}")
```

nabenabe0928:
```python
def objective3d(trial: optuna.Trial) -> tuple[float, float]:
```

This benchmark function slows down much more and hence it is easier to see the improvement.

gen740 (approved):
Thank you for updating! LGTM!

eukaryo (approved):
LGTM. I reviewed the algorithm and code, and confirmed the new implementation returns same value as one of old impl if imputs are all finite.

eukaryo:
<details>

```python

import numpy as np

import optuna


def _lazy_contribs_update(
    contribs: np.ndarray,
    pareto_loss_values: np.ndarray,
    selected_vecs: np.ndarray,
    reference_point: np.ndarray,
) -> np.ndarray:
    """Lazy update the hypervolume contributions.

    S=selected_indices - {indices[max_index]}, T=selected_indices, and S' is a subset of S.
    As we would like to know argmax H(T v {i}) in the next iteration, we can skip HV
    calculations for j if H(T v {i}) - H(T) > H(S' v {j}) - H(S') >= H(T v {j}) - H(T).
    We used the submodularity for the inequality above. As the upper bound of contribs[i] is
    H(S' v {j}) - H(S'), we start to update from i with a higher upper bound so that we can
    skip more HV calculations.
    """
    hv_selected = optuna._hypervolume.WFG().compute(selected_vecs[:-1], reference_point)
    max_contrib = 0.0
    index_from_larger_upper_bound_contrib = np.argsort(-contribs)
    for i in index_from_larger_upper_bound_contrib:
        if contribs[i] < max_contrib:
            # Lazy evaluation to reduce HV calculations.
            # If contribs[i] will not be the maximum next, it is unnecessary to compute it.
            continue

        selected_vecs[-1] = pareto_loss_values[i].copy()
        hv_plus = optuna._hypervolume.WFG().compute(selected_vecs, reference_point)
        # inf - inf in the contribution calculation is always inf.
        contribs[i] = hv_plus - hv_selected if not np.isinf(hv_plus) else np.inf
        max_contrib = max(contribs[i], max_contrib)

    return contribs


def _solve_hssp_on_unique_loss_vals(
    rank_i_loss_vals: np.ndarray,
    rank_i_indices: np.ndarray,
    subset_size: int,
    reference_point: np.ndarray,
) -> np.ndarray:
    assert not np.any(reference_point - rank_i_loss_vals <= 0)
    assert subset_size <= rank_i_indices.size
    n_objectives = reference_point.size
    contribs = np.prod(reference_point - rank_i_loss_vals, axis=-1)
    selected_indices = np.zeros(subset_size, dtype=int)
    selected_vecs = np.empty((subset_size, n_objectives))
    indices = np.arange(rank_i_loss_vals.shape[0], dtype=int)
    for k in range(subset_size):
        max_index = int(np.argmax(contribs))
        selected_indices[k] = indices[max_index]
        selected_vecs[k] = rank_i_loss_vals[max_index].copy()
        keep = np.ones(contribs.size, dtype=bool)
        keep[max_index] = False
        contribs = contribs[keep]
        indices = indices[keep]
        rank_i_loss_vals = rank_i_loss_vals[keep]
        if k == subset_size - 1:
            # We do not need to update contribs at the last iteration.
            break

        contribs = _lazy_contribs_update(
            contribs, rank_i_loss_vals, selected_vecs[: k + 2], reference_point
        )

    return rank_i_indices[selected_indices]


def _solve_hssp(
    rank_i_loss_vals: np.ndarray,
    rank_i_indices: np.ndarray,
    subset_size: int,
    reference_point: np.ndarray,
) -> np.ndarray:
    """Solve a hypervolume subset selection problem (HSSP) via a greedy algorithm.

    This method is a 1-1/e approximation algorithm to solve HSSP.

    For further information about algorithms to solve HSSP, please refer to the following
    paper:

    - `Greedy Hypervolume Subset Selection in Low Dimensions
       <https://doi.org/10.1162/EVCO_a_00188>`_
    """
    if subset_size == rank_i_indices.size:
        return rank_i_indices

    rank_i_unique_loss_vals, indices_of_unique_loss_vals = np.unique(
        rank_i_loss_vals, return_index=True, axis=0
    )
    n_unique = indices_of_unique_loss_vals.size
    if n_unique < subset_size:
        chosen = np.zeros(rank_i_indices.size, dtype=bool)
        chosen[indices_of_unique_loss_vals] = True
        duplicated_indices = np.arange(rank_i_indices.size)[~chosen]
        chosen[duplicated_indices[: subset_size - n_unique]] = True
        return rank_i_indices[chosen]

    subset_indices_of_unique_loss_vals = _solve_hssp_on_unique_loss_vals(
        rank_i_unique_loss_vals, indices_of_unique_loss_vals, subset_size, reference_point
    )
    return rank_i_indices[subset_indices_of_unique_loss_vals]


def _solve_hssp_simple(
    rank_i_loss_vals: np.ndarray,
    rank_i_indices: np.ndarray,
    subset_size: int,
    reference_point: np.ndarray,
) -> np.ndarray:
    """Solve a hypervolume subset selection problem (HSSP) via a greedy algorithm.

    This method is a 1-1/e approximation algorithm to solve HSSP.

    For further information about algorithms to solve HSSP, please refer to the following
    paper:

    - `Greedy Hypervolume Subset Selection in Low Dimensions
       <https://doi.org/10.1162/EVCO_a_00188>`_
    """
    selected_vecs: List[np.ndarray] = []
    selected_indices: List[int] = []
    contributions = [
        optuna._hypervolume.WFG().compute(np.asarray([v]), reference_point)
        for v in rank_i_loss_vals
    ]
    hv_selected = 0.0
    while len(selected_indices) < subset_size:
        max_index = int(np.argmax(contributions))
        contributions[max_index] = -1  # mark as selected
        selected_index = rank_i_indices[max_index]
        selected_vec = rank_i_loss_vals[max_index]
        for j, v in enumerate(rank_i_loss_vals):
            if contributions[j] == -1:
                continue
            p = np.max([selected_vec, v], axis=0)
            contributions[j] -= (
                optuna._hypervolume.WFG().compute(np.asarray(selected_vecs + [p]), reference_point)
                - hv_selected
            )
        selected_vecs += [selected_vec]
        selected_indices += [selected_index]
        hv_selected = optuna._hypervolume.WFG().compute(np.asarray(selected_vecs), reference_point)

    return np.asarray(selected_indices, dtype=int)


def test_solve_hssp_duplicated_infinite_loss() -> None:
    test_case = np.array([[np.inf, 0, 0], [np.inf, 0, 0], [0, np.inf, 0], [0, 0, np.inf]])
    r = np.full(3, np.inf)
    res = optuna._hypervolume._solve_hssp(
        rank_i_loss_vals=test_case, rank_i_indices=np.arange(4), subset_size=2, reference_point=r
    )
    assert (0 not in res) or (1 not in res)


def test_random(num_point = 8, dim = 4):
    rng = np.random.default_rng(0)
    for iter in range(100):
        print(f"start: {iter=}")
        test_case = np.zeros([num_point, dim])
        for i in range(num_point):
            for j in range(dim):
                test_case[i, j] = rng.uniform(0.0, 100.0)
        reference_point = np.full(dim, 101.0)
        a1 = _solve_hssp_simple(test_case, np.arange(num_point), num_point // 2, reference_point)
        v1 = optuna._hypervolume.WFG().compute(np.asarray(test_case[a1]), reference_point)
        a2 = _solve_hssp(test_case, np.arange(num_point), num_point // 2, reference_point)
        v2 = optuna._hypervolume.WFG().compute(np.asarray(test_case[a2]), reference_point)
        if abs(v1 - v2) > 0.0000001:
            print(f"{a1=}")
            print(f"{v1=}")
            print(f"{a2=}")
            print(f"{v2=}")
            assert False


test_random()

```

</details>

Issue #5358, opened 2024-03-26
GP Sampler throw `ValueError: Fewer non-zero entries in p than size`


hrntsm:
### Expected behavior

When optimizing with GPSampler, it stops in the middle of the process.

### Environment

- Optuna version:3.6.0
- Python version:3.9.6
- OS:macOS-14.4-arm64-arm-64bit

### Error messages, stack traces, or logs

```shell
Traceback (most recent call last):
  File "/Users/hrntsm/Desktop/blog/tpe_check copy.py", line 20, in <module>
    study.optimize(objective, n_trials=100)
  File "/Users/hrntsm/Desktop/blog/.venv/lib/python3.9/site-packages/optuna/study/study.py", line 451, in optimize
    _optimize(
  File "/Users/hrntsm/Desktop/blog/.venv/lib/python3.9/site-packages/optuna/study/_optimize.py", line 62, in _optimize
    _optimize_sequential(
  File "/Users/hrntsm/Desktop/blog/.venv/lib/python3.9/site-packages/optuna/study/_optimize.py", line 159, in _optimize_sequential
    frozen_trial = _run_trial(study, func, catch)
  File "/Users/hrntsm/Desktop/blog/.venv/lib/python3.9/site-packages/optuna/study/_optimize.py", line 247, in _run_trial
    raise func_err
  File "/Users/hrntsm/Desktop/blog/.venv/lib/python3.9/site-packages/optuna/study/_optimize.py", line 196, in _run_trial
    value_or_values = func(trial)
  File "/Users/hrntsm/Desktop/blog/tpe_check copy.py", line 8, in objective
    return trial.suggest_float("x", 0, 1)
  File "/Users/hrntsm/Desktop/blog/.venv/lib/python3.9/site-packages/optuna/trial/_trial.py", line 163, in suggest_float
    suggested_value = self._suggest(name, distribution)
  File "/Users/hrntsm/Desktop/blog/.venv/lib/python3.9/site-packages/optuna/trial/_trial.py", line 625, in _suggest
    elif self._is_relative_param(name, distribution):
  File "/Users/hrntsm/Desktop/blog/.venv/lib/python3.9/site-packages/optuna/trial/_trial.py", line 657, in _is_relative_param
    if name not in self.relative_params:
  File "/Users/hrntsm/Desktop/blog/.venv/lib/python3.9/site-packages/optuna/trial/_trial.py", line 73, in relative_params
    self._relative_params = self.study.sampler.sample_relative(
[...]

PR #5359, opened 2024-03-26
Address issue#5358

nabenabe0928:
## Motivation

This PR addresses the issue reported in:
- https://github.com/optuna/optuna/issues/5358

Basically, this bug appears when the optimization already converges but `n_trials` still remains.
Although this is not critical in terms of optimization, I made a patch fix in this PR.

## Description of the changes

Add a patch fix for the issue caused in the acquisition function optimization.


toshihikoyanase:
If possible, could we add some test cases to verify that the bug has been resolved?

nabenabe0928:
I created a unit test in another PR.
- https://github.com/optuna/optuna/pull/5365

PR #5366, opened 2024-03-28
Fix `average_is_best` implementation in `WilcoxonPruner`

contramundum53:
## Motivation
Since we cannot assume that the objective value is the mean of all steps when using WilcoxonPruner, we must compare average with explicitly computed average.


HideakiImamura (approved):
LGTM.

eukaryo (approved):
LGTM

PR #5378, opened 2024-03-31
Prefer last trials for best trial

R-N:
## Motivation
Currently, when there are trials with same best value, Optuna selects the first trial to be best trial. This can happen when the value space is quite discrete and the hyperparameter change rarely affects the value.

In my experience though, the first best trials could have been a fluke specially when it came from the first few trials, when Optuna should still be exploring or warming up. Meanwhile the actual best trials are among the last trials (with same best value) after Optuna has exploited enough. So I think it's better to use the last best trial.

## Description of the changes
When selecting best trial, order the trials by trial id descending after the value.

### DIFF
--- a/optuna/storages/_rdb/models.py
+++ b/optuna/storages/_rdb/models.py
@@ -206,6 +206,7 @@ class TrialModel(BaseModel):
                     )
                 ),
                 desc(TrialValueModel.value),
+                desc(cls.trial_id),
             )
             .limit(1)
             .one_or_none()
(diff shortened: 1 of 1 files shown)


not522:
Thank you for your proposal. We discussed this PR and concluded that merging it would be unacceptable as it breaks backward compatibility. I think what you really want is a method that suggests the best parameters based on the optimization history. We are working on this issue, for example in #5339.

Issue #5099, opened 2023-11-07
The Windows CI fails due to timeout.


HideakiImamura:
### Expected behavior

I have been observing an issue for the past few days where the Windows CI randomly fails due to timeout, regardless of whether it's a PR or on schedule. It looks something like the following CI:　https://github.com/optuna/optuna/actions/runs/6777581250/job/18421473256#step:8:15

All of the CIs appear to fail on the first test case of the visualization features using matplotlib. I have no windows machine to reproduce the error on my local environment.

### Environment

- Optuna version: the current master https://github.com/optuna/optuna/commit/523dcc276a64e0d22ff8c92c17b88509b94bea88
- Python version: 3.9
- OS: Windows Server 2022
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs
Here is the CI's output.

```shell

Run pytest -m "not integration"
============================= test session starts =============================
platform win32 -- Python 3.9.13, pytest-7.4.3, pluggy-1.3.0
rootdir: D:\a\optuna\optuna
configfile: pyproject.toml
plugins: jaxtyping-0.2.23, typeguard-2.13.3
collected 4850 items / 839 deselected / 4011 selected

tests\test_callbacks.py .                                                [  0%]
tests\test_cli.py ...................................................... [  1%]
...
tests\trial_tests\test_trial.py ........................................ [ 80%]
........................................................................ [ 81%]
..............................                                           [ 82%]
tests\trial_tests\test_trials.py ....................................... [ 83%]
...                                                                      [ 83%]
tests\visualization_tests\test_contour.py ..
Error: The operation was canceled.
```

### Steps to reproduce

Trigger the windows CI by workflow dispatcher on your fork.

not522:
CI seems to be stuck on the third test in `test_contour.py`. This is the first test in `test_plot_contour`, which outputs an image using plotly. I am not sure, but perhaps this issue https://github.com/plotly/Kaleido/issues/110 is related.

HideakiImamura:
Downgrading kaleido to 0.1.0 will resolve issue according to the comments like https://github.com/plotly/Kaleido/issues/110#issuecomment-1520317283. I will hotfix it.

not522:
According to the issue, 0.1.0post1 may be better.

not522:
Unfortunately, kaleido 0.1.0post1 did not resolve this issue. https://github.com/optuna/optuna/actions/runs/6818449794

not522:
[...]

PR #5389, opened 2024-04-12
Create an unique renaming filename for each release operation in lock systems of `JournalStorage`

eukaryo:
## Motivation

I want to refactor the code, and hopefully resovling issue https://github.com/optuna/optuna/issues/5099.

https://github.com/optuna/optuna/blob/v3.6.1/tests/storages_tests/test_journal.py#L70 This test fails with time-out at low frequency.

In the current implementation of the lock system for `JournalFileStorage`, renaming filename is created at the beginning and all processes use the same filename.
The NFS v2 specification guarantees that symlink creation and `rename` are atomic operations, but the atomicity of `unlink` operation is not guaranteed. Perhaps it may cause a trouble to perform `unlink` and `rename` operations to the same filename simultaneously on Windows OS.

## Description of the changes

Generate a one-time uuid for each lock-releasing operation.

### DIFF
--- a/optuna/storages/_journal/file.py
+++ b/optuna/storages/_journal/file.py
@@ -43,7 +43,6 @@ class JournalFileSymlinkLock(JournalFileBaseLock):
     def __init__(self, filepath: str) -> None:
         self._lock_target_file = filepath
         self._lock_file = filepath + LOCK_FILE_SUFFIX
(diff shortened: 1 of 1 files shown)


nabenabe0928:

c-bata (approved):
Good catch! LGTM.

not522 (approved):
LGTM!

PR #5391, opened 2024-04-12
Use `TYPE_CHECKING` for `Study` in `samplers`

nabenabe0928:
## Motivation

As circular references happen in `samplers` when removing `multi_objective` due to `Study`, I replaced `Study` in `samplers`  with `"Study"` using `TYPE_CHECKING`.

## Description of the changes

Replace all `Study` in `samplers` with `"Study"` using `TYPE_CHECKING`.

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -46,8 +46,9 @@ if TYPE_CHECKING:
     from optuna.trial import FrozenTrial
     from optuna.trial import Trial
 
+    ObjectiveFuncType = Callable[[Trial], Union[float, Sequence[float]]]
 
-ObjectiveFuncType = Callable[["Trial"], Union[float, Sequence[float]]]
+
 _SYSTEM_ATTR_METRIC_NAMES = "study:metric_names"
 
 
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -42,9 +42,9 @@ from optuna.trial import TrialState
 _dataframe = _LazyImport("optuna.study._dataframe")
 
 if TYPE_CHECKING:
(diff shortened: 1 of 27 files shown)


nabenabe0928:
Although I made everything `"Study"`, we could also replace it with `optuna.study.Study`.

The former is more concise while the latter allows developers to use the direct link to `Study` in IDE.

I would like to ask which one is better.

nabenabe0928:
@eukaryo @not522 Could you review this PR?

not522:
I recommend `from __future__ import annotations` instead of `"Study"` or `optuna.study.Study`.

optuna/study/study.py
```diff
@@ -36,17 +36,18 @@
...
 _dataframe = _LazyImport("optuna.study._dataframe")
 
 if TYPE_CHECKING:
     from optuna.study._dataframe import pd
+    from optuna.trial import FrozenTrial
+    from optuna.trial import Trial
 
 
-ObjectiveFuncType = Callable[[trial_module.Trial], Union[float, Sequence[float]]]
+ObjectiveFuncType = Callable[["Trial"], Union[float, Sequence[float]]]
```
not522, line 50: How about moving it under `if TYPE_CHECKING:`?

```suggestion
    ObjectiveFuncType = Callable[[Trial], Union[float, Sequence[float]]]
```

nabenabe0928: Oh, I did not know that it works in this way!
I reflected your suggestion!

eukaryo (commented):
LGTM.

eukaryo (approved):

PR #4684, opened 2023-05-17
Remove experimental label from _ProgressBar

tungbq:
## Motivation

Remove experimental label from `_ProgressBar`
Closes #4683 
## Description of the changes

 Some changes in` _ProgressBar `
- remove experimental_func
- remove _init_valid method
- remove related comments

PR #5393, opened 2024-04-15
Update docs about `show_progress_bar`

eukaryo:
## Motivation

`show_progress_bar` is not experimental now, but docs is not updated.

## Description of the changes

Update the description.

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -439,10 +439,9 @@ class Study:
                     :ref:`out-of-memory-gc-collect`
 
             show_progress_bar:
-                Flag to show progress bars or not. To disable progress bar, set this :obj:`False`.
-                Currently, progress bar is experimental feature and disabled
-                when ``n_trials`` is :obj:`None`, ``timeout`` is not :obj:`None`, and
-                ``n_jobs`` :math:`\\ne 1`.
+                Flag to show progress bars or not. To show progress bar, set this :obj:`True`.
+                Note that it is disabled when ``n_trials`` is :obj:`None`,
+                ``timeout`` is not :obj:`None`, and ``n_jobs`` :math:`\\ne 1`.
 
         Raises:
             RuntimeError:


nabenabe0928:

optuna/study/study.py
```diff
@@ -439,10 +439,9 @@ def objective(trial):
                     :ref:`out-of-memory-gc-collect`
 
             show_progress_bar:
-                Flag to show progress bars or not. To disable progress bar, set this :obj:`False`.
-                Currently, progress bar is experimental feature and disabled
-                when ``n_trials`` is :obj:`None`, ``timeout`` is not :obj:`None`, and
-                ``n_jobs`` :math:`\\ne 1`.
+                Flag to show progress bars or not. To show progress bar, set this :obj:`True`.
+                Note that it is disabled when ``n_trials`` is :obj:`None`,
+                ``timeout`` is not :obj:`None`, and ``n_jobs`` :math:`\\ne 1`.
```
nabenabe0928, line 444: Isn't it `or`, meaning we cannot use the progress bar if any one of them is `True`? 
```suggestion
                ``timeout`` is not :obj:`None`, or ``n_jobs`` :math:`\\ne 1`.
```

nabenabe0928 (commented):
I left a minor comment, but otherwise, LGTM!

optuna/study/study.py
eukaryo: https://github.com/optuna/optuna/blob/master/optuna/study/_optimize.py#L52
According to the current implementation, `and`  is correct. Iff we allow unlimited trial, set a time-limit, and make multiple jobs, then `progress_bar` is disabled.

nabenabe0928: Thank you for the confirmation!

nabenabe0928 (approved):
LGTM!

not522 (approved):
Thank you. LGTM.
It missed in #4684.

PR #5400, opened 2024-04-17
Speed up `to_internal_repr` in `CategoricalDistribution`

nabenabe0928:
## Motivation

As the current implementation of `to_internal_repr` in `CategoricalDistribution` significantly slows down the optimization when the number of choices is large, I made an improvement.

```python
from argparse import ArgumentParser
import time

import optuna


def objective(trial):
    return trial.suggest_categorical("c", list(range(10000)))


start = time.time()
parser = ArgumentParser()
parser.add_argument("--n_trials", required=True, type=int)
args = parser.parse_args()
study = optuna.create_study()
study.optimize(objective, n_trials=args.n_trials)
print(time.time() - start)

``` 

By using the code above, I got the following results:

|n_trials|Original| This PR|
|:--:|:--:|:--:|
|100|25.6|2.5|
|200|104|5.2|
|300|235|8.2|
|400|419|11.6|
|500|668|15.4|

The unit of the table is second.

## Description of the changes

Use `dict` that maps from a choice to the corresponding index except for float.
By doing so, the time complexity improves from $O(N)$ to $O(1)$ where $N$ is the number of trials.

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -532,6 +532,7 @@ class CategoricalDistribution(BaseDistribution):
         try:
             return self.choices.index(param_value_in_external_repr)
         except ValueError:  # ValueError: param_value_in_external_repr is not in choices.
+            # ValueError also happens if external_repr is nan or includes precision error in float.
             for index, choice in enumerate(self.choices):
                 if _categorical_choice_equal(param_value_in_external_repr, choice):
                     return index
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -6,6 +6,7 @@ from numbers import Real
 from typing import Any
 from typing import cast
 from typing import Dict
+from typing import Hashable
 from typing import Sequence
 from typing import Union
 import warnings
@@ -85,7 +86,7 @@ class BaseDistribution(abc.ABC):
         raise NotImplementedError
 
     def _asdict(self) -> Dict:
-        return self.__dict__
+        return {k: v for k, v in self.__dict__.items() if not k.startswith("_")}
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, BaseDistribution):
@@ -95,7 +96,7 @@ class BaseDistribution(abc.ABC):
         return self.__dict__ == other.__dict__
 
     def __hash__(self) -> int:
-        return hash((self.__class__,) + tuple(sorted(self.__dict__.items())))
+        return hash((self.__class__,) + tuple(sorted(self._asdict().items())))
 
     def __repr__(self) -> str:
         kwargs = ", ".join("{}={}".format(k, v) for k, v in sorted(self._asdict().items()))
@@ -524,7 +525,7 @@ class CategoricalDistribution(BaseDistribution):
                 warnings.warn(message)
 
         self.choices = tuple(choices)
-        self._choice_to_index = {c: i for i, c in enumerate(choices)}
+        self._choice_to_index = {c: i for i, c in enumerate(choices) if isinstance(c, Hashable)}
 
     def to_external_repr(self, param_value_in_internal_repr: float) -> CategoricalChoiceType:
         return self.choices[int(param_value_in_internal_repr)]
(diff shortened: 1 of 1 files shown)


nabenabe0928:
Seems like the change is not really compatible with some storages, so I will close this PR for now...

optuna/distributions.py
```diff
@@ -550,15 +555,27 @@ def __eq__(self, other: Any) -> bool:
...
             return False
         for key, value in self.__dict__.items():
+            cat_system_attr_key = "_choice_to_index"
             if key == "choices":
                 if len(value) != len(getattr(other, key)):
                     return False
                 for choice, other_choice in zip(value, getattr(other, key)):
                     if not _categorical_choice_equal(choice, other_choice):
                         return False
+            elif key == cat_system_attr_key:
```
nabenabe0928, line 565: I would like reviewers to check if we need these lines because `_choice_to_index` should be aligned with `choices` and if `choices` are identical, we should get the identical results in all the existing methods although we can easily make an edge case using machine delta, which is not expected to be use cases anyways.

nabenabe0928:
@eukaryo @not522 Could you review this PR?

optuna/distributions.py
```diff
@@ -524,11 +525,15 @@ def __init__(self, choices: Sequence[CategoricalChoiceType]) -> None:
                 warnings.warn(message)
 
         self.choices = tuple(choices)
+        self._choice_to_index = {c: i for i, c in enumerate(choices) if isinstance(c, Hashable)}
 
     def to_external_repr(self, param_value_in_internal_repr: float) -> CategoricalChoiceType:
         return self.choices[int(param_value_in_internal_repr)]
 
     def to_internal_repr(self, param_value_in_external_repr: CategoricalChoiceType) -> float:
+        if param_value_in_external_repr in self._choice_to_index:
```
nabenabe0928, line 534:
```python
import pickle


class C:
    def __init__(self, a: int):
        self.a = a


with open("foo.pkl", mode="wb") as f:
    pickle.dump(C(1), f)


class C:
    def __init__(self, a: int):
        self.a = a
        self.b = 1 / a


with open("foo.pkl", mode="rb") as f:
    c = pickle.load(f)
    assert isinstance(c, C)
    print(f"Successfully loaded with attributes {c.__dict__}")

```

According to the code above, we can keep the backward compatibility and the old version will also be regarded as instantiated from the latest `CategoricalDistribution`, but any older versions will not have `self._choice_to_index` in the attributes of `CategoricalDistribution` if users store Optuna information in the pickle format.

Shall I change the condition as follows?

```suggestion
        if hasattr(self, "_choice_to_index") and param_value_in_external_repr in self._choice_to_index:
```

not522:
How about the following? I think it will satisfy your needs.
```
diff --git a/optuna/distributions.py b/optuna/distributions.py
index cf370e41b..02488a5d3 100644
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -529,9 +529,13 @@ class CategoricalDistribution(BaseDistribution):
         return self.choices[int(param_value_in_internal_repr)]
 
     def to_internal_repr(self, param_value_in_external_repr: CategoricalChoiceType) -> float:
-        for index, choice in enumerate(self.choices):
-            if _categorical_choice_equal(param_value_in_external_repr, choice):
-                return index
+        try:
+            return self.choices.index(param_value_in_external_repr)
+        except:
+            # self.choices.index raises a ValueError when the choice is NaN.
+            for index, choice in enumerate(self.choices):
+                if _categorical_choice_equal(param_value_in_external_repr, choice):
+                    return index
 
         raise ValueError(f"'{param_value_in_external_repr}' not in {self.choices}.")
 
```

nabenabe0928:
> ```python
> from argparse import ArgumentParser
> import time
> 
> import optuna
> 
> 
> def objective(trial):
>     return trial.suggest_categorical("c", list(range(10000)))
> 
> 
> start = time.time()
> parser = ArgumentParser()
> parser.add_argument("--n_trials", required=True, type=int)
> args = parser.parse_args()
> study = optuna.create_study()
> study.optimize(objective, n_trials=args.n_trials)
> print(time.time() - start)
> ```

@not522 Thank you for the suggestion!
I tried it with 3000 trials, and it took 257 seconds for mine and took 276 seconds for yours.
Overall, the difference is more or less negligible for most cases, so I will stick to your suggestion!

optuna/distributions.py
```diff
@@ -529,9 +529,12 @@ def to_external_repr(self, param_value_in_internal_repr: float) -> CategoricalCh
         return self.choices[int(param_value_in_internal_repr)]
 
     def to_internal_repr(self, param_value_in_external_repr: CategoricalChoiceType) -> float:
-        for index, choice in enumerate(self.choices):
-            if _categorical_choice_equal(param_value_in_external_repr, choice):
-                return index
+        try:
+            return self.choices.index(param_value_in_external_repr)
+        except ValueError:
```
not522, line 534: Could you add a comment here? It is difficult to recognize why `ValueError` is raised.

not522: Thank you for your update https://github.com/optuna/optuna/pull/5400/commits/9e65e7b54c068be9610bc451a7778592f12722a8. But, could you please make clarify more? Normally, `index` is the same as for-if in `except`, but the behavior is different only when the elements contain NaN.

FYI:
```python
(0, 1, float("nan")).index("nan")  # raise ValueError
```

nabenabe0928: Ok, I added another line!

not522 (approved):

eukaryo (approved):
LGTM.

PR #5411, opened 2024-04-23
Do not normalize importance in f-ANOVA

nabenabe0928:
## Motivation

As `get_param_importances` provides an option to not normalize importances and raw importance values in f-ANOVA have meaning by themselves, we will stop normalizing the importances by f-ANOVA if not specified.

## Description of the changes

Simply stop normalizing impotances.


not522 (approved):
LGTM!
It is a breaking change but necessary for the user to choose whether to normalize or not.

eukaryo (approved):
LGTM.

PR #5412, opened 2024-04-23
Replace `numpy` with `np`

nabenabe0928:
## Motivation

As `np` is used instead of `numpy` in the Optuna project, I replace `numpy` with `np`.

## Description of the changes

Replace `numpy` with `np`.

I enumerated the files with `git grep -l 'import numpy' | xargs grep -L 'import numpy as np'`.

### DIFF
--- a/optuna/_transform.py
+++ b/optuna/_transform.py
@@ -5,7 +5,7 @@ from typing import List
 from typing import Tuple
 from typing import Union
 
-import numpy
+import numpy as np
 
 from optuna.distributions import BaseDistribution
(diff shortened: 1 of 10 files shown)


eukaryo (commented):
LGTM. I confirmed the old and changed codes do not use the name `np` in another meaning.

eukaryo (approved):

not522 (approved):
LGTM.
Having a consistent coding style will make development smoother, and [numpy.org](https://numpy.org/) says that `import numpy as np` is "The standard way to import NumPy."

PR #5413, opened 2024-04-23
Remove experimental from plot

nabenabe0928:
## Motivation

As `plot_rank`, and `plot_timeline` are stable, we remove the experimental label from them.

We do not remove the label from the matplotlib version because the handling of the directory structure for matplotlib could be still under discussion.

For `plot_rank` and `plot_timeline`, I will summarize some discussion points:
- they are often used, so we decided to remove the experimental label,
- the feature of `plot_rank` is similar to that of `plot_contour`, but we decided to keep `plot_rank` as well and we have not seen any problem for a while,
- Although `plot_timeline` may need some additional arguments such as `n_trials`, `head`, and `tail`, we can add such arguments in the future without breaking the compatibility.

For this reason, we decide to remove the experimental label as we will not break the compatibility in the future.

## Description of the changes

Remove the experimental label from the two plot functions mentioned above.

### DIFF
--- a/optuna/visualization/_hypervolume_history.py
+++ b/optuna/visualization/_hypervolume_history.py
@@ -5,6 +5,7 @@ from typing import NamedTuple
 
 import numpy as np
 
+from optuna._experimental import experimental_func
 from optuna._hypervolume import WFG
 from optuna.logging import get_logger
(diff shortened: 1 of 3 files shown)


eukaryo (approved):
LGTM!

not522 (approved):
LGTM!

PR #5414, opened 2024-04-23
Remove `samplers.intersection`

nabenabe0928:
## Motivation

This PR removes `samplers.intersection`, which has been deprecated for a while.

## Description of the changes

Remove `samplers.intersection` and reflect the changes to the corresponding documentation.


eukaryo (approved):
LGTM. I am not familiar with the ancient historical background very much, but this could be a good catch...

not522 (approved):
LGTM!

PR #5415, opened 2024-04-23
Drop implicit create-study in `ask` command

eukaryo:
## Motivation

I want `ask` command to stop implicit create-study.

## Description of the changes

- I replaced code of create-study in `ask` command with a custom error message.
- I modified related tests.
- I added a test for this change.

## Note

I believe that the usage of the CLI should be similar to the usage of the python library. When using optuna in python, user must call `optuna.create_study` method explicitly. Therefore, I want to introduce this change.


optuna/cli.py
```diff
@@ -831,7 +831,10 @@ def take_action(self, parsed_args: Namespace) -> int:
             )
 
         except KeyError:
-            study = optuna.create_study(**create_study_kwargs)
+            raise KeyError(
+                "Implicit study creation within the 'ask' command was dropped in Optuna v4.0."
```
not522, line 835:
```suggestion
                "Implicit study creation within the 'ask' command was dropped in Optuna v4.0.0. "
```

not522 (commented):
Could you add tests to confirm that that `ask` command fails on non-exist studies?

not522 (approved):
Thank you for your update.LGTM.

tests/test_cli.py
```diff
@@ -1349,6 +1364,36 @@ def test_ask_sampler_kwargs_without_sampler() -> None:
...
+            db_url,
+            "--study-name",
+            study_name,
+            "--search-space",
+            search_space,
+        ]
+
+        result = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        error_message = result.stderr.decode()
+        assert (
```
HideakiImamura, line 1391: I think that it is fragile to check the entire error message. How about avoiding to check the message?

HideakiImamura (commented):
Thanks for the PR. I have a comment. PTAL.

tests/test_cli.py
eukaryo: This testing method is commonly used in `test_cli.py`. I agree that your concern is valid, however, I think a refactoring addressing the concern is outside the scope of this PR.

HideakiImamura (approved):
LGTM.

PR #5422, opened 2024-05-01
Fix `_normalize_value` for incomplete trials

not522:
## Motivation

`_normalize_value` should return infinity value for incomplete trials. It seems unused in the current implementation, but I want to fix it to eliminate the possibility of bugs occurring in the future.

## Description of the changes

- Return `float("inf")` for `None` value.
- Add a test.

### DIFF
--- a/tests/study_tests/test_multi_objective.py
+++ b/tests/study_tests/test_multi_objective.py
@@ -6,6 +6,7 @@ import pytest
 from optuna.study import StudyDirection
 from optuna.study._multi_objective import _dominates
 from optuna.study._multi_objective import _fast_non_dominated_sort
+from optuna.study._multi_objective import _normalize_value
 from optuna.trial import create_trial
 from optuna.trial import TrialState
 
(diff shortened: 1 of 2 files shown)


not522:
@c-bata @nabenabe0928 Could you review this PR?

tests/study_tests/test_multi_objective.py
```diff
@@ -161,3 +162,10 @@ def test_fast_non_dominated_sort_invalid() -> None:
         _fast_non_dominated_sort(
             np.array([[1.0, 2.0], [3.0, 4.0]]), penalty=np.array([1.0, 2.0, 3.0])
         )
+
+
+def test_normalize_value() -> None:
+    assert _normalize_value(1.0, StudyDirection.MINIMIZE) == 1.0
+    assert _normalize_value(1.0, StudyDirection.MAXIMIZE) == -1.0
+    assert _normalize_value(None, StudyDirection.MINIMIZE) == float("inf")
+    assert _normalize_value(None, StudyDirection.MAXIMIZE) == float("inf")
```
nabenabe0928, line 171: I did not fully understand this part, because shouldn't `None` be the worst value in the given direction, i.e. `-inf`, in this case?
I would like to know what bugs you expect when we return `-inf` instead of `inf` for maximization problems with `value=None`.

nabenabe0928 (commented):
Thank you for your PR!
I left a question!

tests/study_tests/test_multi_objective.py
c-bata: As `_normalize_value(1.0, StudyDirection.MAXIMIZE)` returns `-1.0`, this function transforms the maximization problem to the minimization problem. So that this is an expected behavior.

c-bata (approved):
LGTM.

tests/study_tests/test_multi_objective.py
nabenabe0928: Oh yeah, point taken!

nabenabe0928 (approved):

PR #5423, opened 2024-05-01
Add test for `best_trial`

not522:
## Motivation

`best_trial` does not have the unit test. Especially, the test for the same objective values is lacking (e.g. #5378).

## Description of the changes

Add test for `best_trial`.

### DIFF
--- a/tests/study_tests/test_study.py
+++ b/tests/study_tests/test_study.py
@@ -1166,6 +1166,21 @@ def test_optimize_with_multi_objectives(n_objectives: int) -> None:
         assert len(trial.values) == n_objectives
 
 
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES)
+def test_best_trial(storage_mode: str) -> None:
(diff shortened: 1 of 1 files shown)


not522:
@c-bata @eukaryo Could you review this PR?

tests/study_tests/test_study.py
```diff
@@ -1166,6 +1166,21 @@ def objective(trial: Trial) -> list[float]:
...
+        study = create_study(storage=storage)
+        with pytest.raises(ValueError):
+            study.best_trial
+        study.tell(study.ask(), 1)
+        assert study.best_trial.number == 0
+        study.tell(study.ask(), 0)
+        assert study.best_trial.number == 1
+        study.tell(study.ask(), 0)
+        # If the objective values are the same, the best trial will be the first trial executed.
+        assert study.best_trial.number == 1
```
c-bata, line 1181: [Question] I believe this is merely an implementation detail. Do you think there are any advantages to making this behavior a specification?

PR #5425, opened 2024-05-01
Replace `abc.ABC` with `typing.Protocol` in `BaseSampler`

porink0424:
## Motivation

In Python, static type checkers are now getting popular, and there are fewer and fewer times to use dynamic ones such as `abc.ABC`. This PR suggests replacing `abc.ABC` with `Protocol` in `BaseSampler`.

## Description of the changes

- Replaced `abc.ABC` with `typing.Protocol` in `BaseSampler`.
- Adjusted implementations of each sampler to align with modifications in `BaseSampler`.
- Added `typing_extensions` dependency when python version is less than 3.8.
- Refactored `test_samplers`.

## Note

By merging this PR, dynamic type checks would be no longer available such as `isinstance(obj, BaseSampler)`, `issubclass(cls, BaseSampler)`. The migration guide will be needed for users who rely on dynamic type checks.

### DIFF
--- a/optuna/samplers/_base.py
+++ b/optuna/samplers/_base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from abc import abstractmethod
 from typing import Any
 from typing import Callable
 from typing import Dict
@@ -62,6 +63,7 @@ class BaseSampler(Protocol):
     def __str__(self) -> str:
         return self.__class__.__name__
 
+    @abstractmethod
     def infer_relative_search_space(
         self, study: Study, trial: FrozenTrial
     ) -> Dict[str, BaseDistribution]:
@@ -88,6 +90,7 @@ class BaseSampler(Protocol):
         """
         ...
 
+    @abstractmethod
     def sample_relative(
         self, study: Study, trial: FrozenTrial, search_space: Dict[str, BaseDistribution]
     ) -> Dict[str, Any]:
@@ -118,6 +121,7 @@ class BaseSampler(Protocol):
         """
         ...
 
+    @abstractmethod
     def sample_independent(
         self,
         study: Study,
@@ -173,7 +177,7 @@ class BaseSampler(Protocol):
             trial:
                 Target trial object.
         """
-        ...
+        return
 
     def after_trial(
         self,
@@ -203,7 +207,7 @@ class BaseSampler(Protocol):
                 Resulting trial values. Guaranteed to not be :obj:`None` if trial succeeded.
 
         """
-        ...
+        return
 
     def reseed_rng(self) -> None:
         """Reseed sampler's random number generator.
@@ -214,7 +218,7 @@ class BaseSampler(Protocol):
         same values. To prevent this issue, this method assigns a different seed to each random
         number generator.
         """
-        ...
+        return
 
 
 def _raise_error_if_multi_objective(sampler: BaseSampler, study: Study) -> None:
--- a/optuna/samplers/_base.py
+++ b/optuna/samplers/_base.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+<<<<<<< fix/replace-abc-with-protocol
 from abc import abstractmethod
+=======
+import abc
+from collections.abc import Callable
+from collections.abc import Sequence
+>>>>>>> master
 from typing import Any
-from typing import Callable
-from typing import Dict
-from typing import Optional
-from typing import Sequence
 from typing import TYPE_CHECKING
 
 
@@ -66,7 +68,7 @@ class BaseSampler(Protocol):
     @abstractmethod
     def infer_relative_search_space(
         self, study: Study, trial: FrozenTrial
-    ) -> Dict[str, BaseDistribution]:
+    ) -> dict[str, BaseDistribution]:
         """Infer the search space that will be used by relative sampling in the target trial.
 
         This method is called right before :func:`~optuna.samplers.BaseSampler.sample_relative`
@@ -92,8 +94,8 @@ class BaseSampler(Protocol):
 
     @abstractmethod
     def sample_relative(
-        self, study: Study, trial: FrozenTrial, search_space: Dict[str, BaseDistribution]
-    ) -> Dict[str, Any]:
+        self, study: Study, trial: FrozenTrial, search_space: dict[str, BaseDistribution]
+    ) -> dict[str, Any]:
         """Sample parameters in a given search space.
 
         This method is called once at the beginning of each trial, i.e., right before the
@@ -184,7 +186,7 @@ class BaseSampler(Protocol):
         study: Study,
         trial: FrozenTrial,
         state: TrialState,
-        values: Optional[Sequence[float]],
+        values: Sequence[float] | None,
     ) -> None:
         """Trial post-processing.
 
--- a/optuna/samplers/_base.py
+++ b/optuna/samplers/_base.py
@@ -4,9 +4,15 @@ from typing import Any
 from typing import Callable
 from typing import Dict
 from typing import Optional
-from typing import Protocol
 from typing import Sequence
 from typing import TYPE_CHECKING
+
+
+try:
+    from typing import Protocol
+except ImportError:
+    from typing_extensions import Protocol  # type: ignore
+
 import warnings
 
 import numpy as np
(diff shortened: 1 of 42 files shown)


pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
   "sqlalchemy>=1.3.0",
   "tqdm",
   "PyYAML",  # Only used in `optuna/cli.py`.
+  "typing_extensions; python_version<'3.8'",
```
toshihikoyanase, line 42: The `typing_extensions` package may be subject to version restrictions. See https://github.com/optuna/optuna/blob/ab958d42e44cad2e0517907278d55ca7432ea32c/pyproject.toml#L63.

optuna/samplers/_base.py
```diff
@@ -216,15 +218,15 @@ def reseed_rng(self) -> None:
...
 
-        pass
 
-    def _raise_error_if_multi_objective(self, study: Study) -> None:
-        if study._is_multi_objective():
-            raise ValueError(
-                "If the study is being used for multi-objective optimization, "
-                f"{self.__class__.__name__} cannot be used."
-            )
+def _raise_error_if_multi_objective(sampler: BaseSampler, study: Study) -> None:
```
toshihikoyanase, line 224: While `_rase_error_if_multi_objective` is private, some user-defined samplers use it, and these will not be compatible with Optuna v4.
https://github.com/Yard1/autopipeline/blob/b3c8c325adb969ddf9585d313f89dc53e0f6c5eb/automl/search/samplers/optuna_rf_sampler.py#L694

optuna/samplers/_base.py
```diff
@@ -20,7 +27,7 @@
     from optuna.study import Study
 
 
-class BaseSampler(abc.ABC):
+class BaseSampler(Protocol):
```
toshihikoyanase, line 30: Following this update, expressions like `isinstance(foo, BaseSampler)` and `issubclass(bar, BaseSampler)` will not work. We can see such usage in some OSS code, and it might be helpful to provide a migration guide for those affected.
https://github.com/volico/mlexp/blob/c07c96de964ea181571663a710bb2fc9a1559b25/mlexp/trainers/_base_trainer.py#L78-L80

toshihikoyanase: Although `typing.runtime_checkable` allows us to use `isinstance` and `issubclass` with `Protocol`, it comes with some drawbacks, including impacts on execution speed and code complexity. Therefore, it may not be a good idea to employ it here. https://docs.python.org/3/library/typing.html#typing.runtime_checkable

y0z (commented):
The implementation looks OK.

I think It is required to update some documentations.
For example, the following sentence is no longer exact because duck typing allows you create a sampler without inheriting BaseSampler.
https://optuna.readthedocs.io/en/stable/tutorial/20_recipes/005_user_defined_sampler.html
> To create a new sampler, you need to define a class that inherits [BaseSampler](https://optuna.readthedocs.io/en/stable/reference/samplers/generated/optuna.samplers.BaseSampler.html#optuna.samplers.BaseSampler).

porink0424:
@y0z 
Thank you for your comment, I updated the docs on the part you mentioned.
I think other parts are outside the scope of this PR, so I decided not to update them at this time.

y0z (approved):
LGTM

tutorial/20_recipes/005_user_defined_sampler.py
```diff
@@ -19,11 +19,20 @@
...
+2. Define a class that inplements all six methods of :class:`~optuna.samplers.BaseSampler`; 
 :meth:`~optuna.samplers.BaseSampler.infer_relative_search_space`,
-:meth:`~optuna.samplers.BaseSampler.sample_relative`, and
-:meth:`~optuna.samplers.BaseSampler.sample_independent`.
+:meth:`~optuna.samplers.BaseSampler.sample_relative`,
+:meth:`~optuna.samplers.BaseSampler.sample_independent`,
+:meth:`~optuna.samplers.BaseSampler.before_trial`,
+:meth:`~optuna.samplers.BaseSampler.after_trial`,
+and :meth:`~optuna.samplers.BaseSampler.reseed_rng`.
+    - In this case, you do not need to inherit :class:`~optuna.samplers.BaseSampler`.
```
nabenabe0928, line 35: Could you try adding a blank line before the itemization?
```suggestion

    - In this case, you do not need to inherit :class:`~optuna.samplers.BaseSampler`.
```

nabenabe0928 (commented):
Probably, doc generation fails because of a syntax error.

tutorial/20_recipes/005_user_defined_sampler.py
```diff
@@ -19,11 +19,20 @@
...
 
-To create a new sampler, you need to define a class that inherits :class:`~optuna.samplers.BaseSampler`.
-The base class has three abstract methods;
+To create a new sampler, you have two options:
+
+1. Define a class that inherits :class:`~optuna.samplers.BaseSampler` and implements three abstract methods; 
+:meth:`~optuna.samplers.BaseSampler.infer_relative_search_space`,
+:meth:`~optuna.samplers.BaseSampler.sample_relative`,
+and :meth:`~optuna.samplers.BaseSampler.sample_independent`.
+2. Define a class that inplements all six methods of :class:`~optuna.samplers.BaseSampler`;
```
nabenabe0928, line 28: Typo
```suggestion
2. Define a class that implements all six methods of :class:`~optuna.samplers.BaseSampler`; 
```

nabenabe0928 (commented):
I found a typo.

porink0424:
@nabenabe0928 
Thank you for your catch!

toshihikoyanase (changes requested):
@porink0424 You may already be aware, but could you please address this comment? https://github.com/optuna/optuna/pull/5425#discussion_r1593485903

optuna/samplers/_base.py
porink0424: @toshihikoyanase 
Thank you for your pointing out. Where should I leave a migration guide for this change?

toshihikoyanase: @porink0424 Maybe we can publish the guide in a GitHub discussion along with other breaking changes. Could you outline the guide here, please?

Honestly, I haven't come up with straightforward ways to migrate the code.  In [this case](https://github.com/volico/mlexp/blob/c07c96de964ea181571663a710bb2fc9a1559b25/mlexp/trainers/_base_trainer.py#L78-L80), we might simply remove the `isinstance` assertion and rely on the [type checking](https://github.com/volico/mlexp/blob/c07c96de964ea181571663a710bb2fc9a1559b25/mlexp/trainers/_base_trainer.py#L40) instead.

So, we might recommend users migrate runtime checks such as `isinstance` and `issubclass` to type checking such, for example by using `mypy`. 

What do you think?

porink0424: @toshihikoyanase 

Here is my suggestion for the guide:

----------------

## Runtime check for `BaseSampler` is not available

In Optuna v4, the `BaseSampler` class now inherits `typing.Protocol` instead of `abc.ABC`. This change makes it possible to type-check samplers using static type checkers. However, this change makes it no longer possible to dynamically check whether a sampler is an instance of `BaseSampler` or whether a sampler class is a subclass of `BaseSampler`.

```python
# Optuna v3.x
isinstance(sampler, optuna.samplers.BaseSampler)
issubclass(MySampler, optuna.samplers.BaseSampler)

# These checks are no longer available in Optuna v4
```

Please use static type checking such as `mypy` instead of runtime checks in Optuna v4.

----------------

toshihikoyanase: Sorry for the delayed response. Seems good to me. Thank you!

toshihikoyanase:
@porink0424 Thank you for your response. I believe it will also be helpful for developers.

@not522 I think my comments have been resolved. Please merge after your review.

nabenabe0928:
We discussed this PR internally and had the following concerns:
- `BaseSampler` has `__str__` in the protocol interface,
- on the other hand, a new `Sampler` does not have to explicitly define `__str__` because a class object always has `__str__` by default and it cannot be detected by protocol, and
- as `sampler.__class__.__name__` itself may be referred by some users, it is not a good idea to change the behavior of `sampler.__class__.__name__` if undefined.

For these reasons, we decided to stop and close this PR.

PR #5428, opened 2024-05-02
Check whether the study is multi-objective in `sample_independent` of `GPSampler`

eukaryo:
## Motivation

`GPSampler` does not currently support multi-objective optimization as well as `CmaEsSampler`.

`CmaEsSampler` checks if the study is multi-objective in both `sample_relative` ([code](https://github.com/optuna/optuna/blob/39f83f7eda0979af27d5e9dd6acc7a08fa1c63ec/optuna/samplers/_cmaes.py#L388)) and `sample_independent` ([code](https://github.com/optuna/optuna/blob/39f83f7eda0979af27d5e9dd6acc7a08fa1c63ec/optuna/samplers/_cmaes.py#L733)).

On the other hand, `GPSampler` checks it only in `sample_relative` ([code](https://github.com/optuna/optuna/blob/39f83f7eda0979af27d5e9dd6acc7a08fa1c63ec/optuna/samplers/_gp/sampler.py#L137)).

## Description of the changes

I added code to check if the study is multi-objective in `sample_independent` of `GPSampler`.

### DIFF
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -206,6 +206,7 @@ class GPSampler(BaseSampler):
         param_name: str,
         param_distribution: BaseDistribution,
     ) -> Any:
+        self._raise_error_if_multi_objective(study)
         return self._independent_sampler.sample_independent(
             study, trial, param_name, param_distribution
         )


HideakiImamura (approved):
LGTM.

PR #5431, opened 2024-05-07
Fix heartbeat for race condition

not522:
## Motivation

The current heartbeat implementation has a bug when it comes to race conditions. If the heartbeat value is updated between getting the current time and getting the heartbeat value, the time difference will be negative and the `.seconds` value will be 86399. (ref: https://docs.python.org/3/library/datetime.html#datetime.timedelta)

I have checked this bug using the following code and modification.

```python
import random
import time
import optuna

def objective(trial):
    time.sleep(random.randrange(5))
    return 0

storage = optuna.storages.RDBStorage(url="sqlite:///tmp.db", heartbeat_interval=1, grace_period=2)
study.optimize(objective, n_trials=100, n_jobs=2)
```

```
$ git diff
diff --git a/optuna/storages/_rdb/storage.py b/optuna/storages/_rdb/storage.py
index 5a225aa10..497b7f6de 100644
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -998,6 +998,9 @@ class RDBStorage(BaseStorage, BaseHeartbeat):
             # https://github.com/optuna/optuna/pull/2190#issuecomment-766605088 for details
             current_heartbeat = current_heartbeat.replace(tzinfo=None)
 
+            import random
+            import time
+            time.sleep(random.randrange(5))
             running_trials = (
                 session.query(models.TrialModel)
                 .options(sqlalchemy_orm.selectinload(models.TrialModel.heartbeats))
```

## Description of the changes

Comparison between timedelta objects instead of seconds.

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 from contextlib import contextmanager
 import copy
 from datetime import datetime
+from datetime import timedelta
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


eukaryo (approved):
LGTM. I agree with using tiemdelta.

nabenabe0928 (approved):
Thank you for the PR, LGTM!

Issue #5374, opened 2024-03-29
Returning `float("inf")` failed with MOTPE


gen740:
### Expected behavior

No errors occur.

### Environment

- Optuna version:4.0.0.dev
- Python version:3.12.2
- OS:macOS-14.4.1-arm64-arm-64bit

### Error messages, stack traces, or logs

```shell
[I 2024-03-29 17:20:44,217] A new study created in memory with name: no-name-3289a723-a36c-4729-a37e-6b440b830668
[I 2024-03-29 17:20:44,218] Trial 0 finished with values: [-0.497867971784556, inf] and parameters: {'x': -0.497867971784556}. 
[I 2024-03-29 17:20:44,218] Trial 1 finished with values: [1.3219469606529486, 0.0] and parameters: {'x': 1.3219469606529486}. 
[I 2024-03-29 17:20:44,218] Trial 2 finished with values: [-2.9993137510959307, inf] and parameters: {'x': -2.9993137510959307}. 
[I 2024-03-29 17:20:44,218] Trial 3 finished with values: [-1.1860045642089614, inf] and parameters: {'x': -1.1860045642089614}. 
[I 2024-03-29 17:20:44,218] Trial 4 finished with values: [-2.119464655097322, inf] and parameters: {'x': -2.119464655097322}. 
[I 2024-03-29 17:20:44,218] Trial 5 finished with values: [-2.445968431387213, inf] and parameters: {'x': -2.445968431387213}. 
[I 2024-03-29 17:20:44,218] Trial 6 finished with values: [-1.8824387317339746, inf] and parameters: {'x': -1.8824387317339746}. 
[I 2024-03-29 17:20:44,218] Trial 7 finished with values: [-0.9266356377417135, inf] and parameters: {'x': -0.9266356377417135}. 
[I 2024-03-29 17:20:44,218] Trial 8 finished with values: [-0.6193951546159804, inf] and parameters: {'x': -0.6193951546159804}. 
[I 2024-03-29 17:20:44,218] Trial 9 finished with values: [0.23290040402014167, 0.0] and parameters: {'x': 0.23290040402014167}. 
/Users/gen740/home/PFN/optuna/optuna/_hypervolume/hssp.py:62: RuntimeWarning: invalid value encountered in subtract
  assert not np.any(reference_point - rank_i_loss_vals < 0)
/Users/gen740/home/PFN/optuna/optuna/_hypervolume/hssp.py:64: RuntimeWarning: invalid value encountered in subtract
[...]

PR #5432, opened 2024-05-08
Suppress warnings from `numpy` in hypervolume computation

eukaryo:
## Motivation

I want to reduce warning `RuntimeWarning: invalid value encountered in subtract` in unittests.
Related to issue https://github.com/optuna/optuna/issues/3815 .

## Description of the changes

I applied `with np.errstate(invalid='ignore'):` to minimum extent necessary and suppressed warnings.

### DIFF
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -45,8 +45,7 @@ def _solve_hssp_on_unique_loss_vals(
     reference_point: np.ndarray,
 ) -> np.ndarray:
     with np.errstate(invalid="ignore"):
-        distance_from_reference_point = reference_point - rank_i_loss_vals
-    assert not np.any(distance_from_reference_point <= 0)
+        diff_of_loss_vals_and_ref_point = reference_point - rank_i_loss_vals
     assert subset_size <= rank_i_indices.size
     n_objectives = reference_point.size
     contribs = np.prod(distance_from_reference_point, axis=-1)
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -44,6 +44,8 @@ def _solve_hssp_on_unique_loss_vals(
     subset_size: int,
     reference_point: np.ndarray,
 ) -> np.ndarray:
+    if not np.isfinite(reference_point).all():
+        return rank_i_indices[list(range(subset_size))]
     with np.errstate(invalid="ignore"):
         diff_of_loss_vals_and_ref_point = reference_point - rank_i_loss_vals
     assert subset_size <= rank_i_indices.size
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -48,7 +48,7 @@ def _solve_hssp_on_unique_loss_vals(
         diff_of_loss_vals_and_ref_point = reference_point - rank_i_loss_vals
     assert subset_size <= rank_i_indices.size
     n_objectives = reference_point.size
-    contribs = np.prod(distance_from_reference_point, axis=-1)
+    contribs = np.prod(diff_of_loss_vals_and_ref_point, axis=-1)
     selected_indices = np.zeros(subset_size, dtype=int)
     selected_vecs = np.empty((subset_size, n_objectives))
     indices = np.arange(rank_i_loss_vals.shape[0], dtype=int)
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -44,7 +44,7 @@ def _solve_hssp_on_unique_loss_vals(
     subset_size: int,
     reference_point: np.ndarray,
 ) -> np.ndarray:
-    with np.errstate(invalid='ignore'):
+    with np.errstate(invalid="ignore"):
         distance_from_reference_point = reference_point - rank_i_loss_vals
     assert not np.any(distance_from_reference_point <= 0)
     assert subset_size <= rank_i_indices.size
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -44,10 +44,12 @@ def _solve_hssp_on_unique_loss_vals(
     subset_size: int,
     reference_point: np.ndarray,
 ) -> np.ndarray:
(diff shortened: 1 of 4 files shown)


optuna/_hypervolume/hssp.py
```diff
@@ -44,10 +44,12 @@ def _solve_hssp_on_unique_loss_vals(
     subset_size: int,
     reference_point: np.ndarray,
 ) -> np.ndarray:
-    assert not np.any(reference_point - rank_i_loss_vals <= 0)
+    with np.errstate(invalid="ignore"):
+        distance_from_reference_point = reference_point - rank_i_loss_vals
+    assert not np.any(distance_from_reference_point <= 0)
```
nabenabe0928, line 49: I found out that `BaseHypervolume` actually tests `assert not np.any(diff_of_loss_vals_and_ref_point <= 0)`, so we do not need it.
```suggestion
    with np.errstate(invalid="ignore"):
        diff_of_loss_vals_and_ref_point = reference_point - rank_i_loss_vals
```

optuna/_hypervolume/hssp.py
```diff
@@ -44,10 +44,12 @@ def _solve_hssp_on_unique_loss_vals(
...
     assert subset_size <= rank_i_indices.size
     n_objectives = reference_point.size
-    contribs = np.prod(reference_point - rank_i_loss_vals, axis=-1)
+    contribs = np.prod(distance_from_reference_point, axis=-1)
```
nabenabe0928, line 52:
```suggestion
    contribs = np.prod(diff_of_loss_vals_and_ref_point, axis=-1)
```

nabenabe0928 (commented):
Thank you for the PR, I left a minor comment!

nabenabe0928:
In fact,  this PR relates to the following issue's comment:
- https://github.com/optuna/optuna/issues/5374#issuecomment-2028249506

Namely, we should probably return `inf` instead of `nan` for `inf - inf`.

eukaryo:
I added guards that immediately returns if reference_points has non-finite value(s). I kept `with np.errstate(invalid='ignore'):` clauses because trials can have non-finite value(s).

Anyway, I think issue https://github.com/optuna/optuna/issues/5374 will resolved by PR https://github.com/optuna/optuna/pull/5435 .

optuna/_hypervolume/utils.py
```diff
@@ -35,6 +37,7 @@ def _compute_2d(solution_set: np.ndarray, reference_point: np.ndarray) -> float:
     mask = sorted_solution_set[:, 1] <= reference_points_y_cummin
 
     used_solution = sorted_solution_set[mask]
-    edge_length_x = reference_point[0] - used_solution[:, 0]
-    edge_length_y = reference_points_y_cummin[mask] - used_solution[:, 1]
+    with np.errstate(invalid="ignore"):
+        edge_length_x = reference_point[0] - used_solution[:, 0]
+        edge_length_y = reference_points_y_cummin[mask] - used_solution[:, 1]
```
nabenabe0928, line 42: Do we still need it after we added the guard? Shouldn't the warning happen at all because `inf - inf` will not happen?

optuna/_hypervolume/hssp.py
```diff
@@ -44,10 +44,13 @@ def _solve_hssp_on_unique_loss_vals(
     subset_size: int,
     reference_point: np.ndarray,
 ) -> np.ndarray:
-    assert not np.any(reference_point - rank_i_loss_vals <= 0)
+    if not np.isfinite(reference_point).all():
+        return rank_i_indices[list(range(subset_size))]
+    with np.errstate(invalid="ignore"):
```
nabenabe0928, line 49: Do we still need it after we added the guard? Shouldn't the warning happen at all because `inf - inf` will not happen?

optuna/_hypervolume/utils.py
eukaryo: Sorry, you are right.

optuna/_hypervolume/hssp.py
```diff
@@ -44,10 +44,12 @@ def _solve_hssp_on_unique_loss_vals(
     subset_size: int,
     reference_point: np.ndarray,
 ) -> np.ndarray:
-    assert not np.any(reference_point - rank_i_loss_vals <= 0)
+    if not np.isfinite(reference_point).all():
+        return rank_i_indices[list(range(subset_size))]
```
nabenabe0928, line 48: In this function, the length of `rank_i_indices` is guaranteed to be larger than or equal to `subset_size`, so the following should be fine.
```suggestion
        return rank_i_indices[:subset_size]
```

nabenabe0928 (commented):
Thank you for the changes, I left a minor comment!

optuna/_hypervolume/hssp.py
eukaryo: Good catch! Thank you

nabenabe0928 (commented):
Thank you for the change and the update in the unittest, LGTM!

nabenabe0928 (approved):

optuna/_hypervolume/hssp.py
```diff
@@ -44,10 +44,12 @@ def _solve_hssp_on_unique_loss_vals(
     subset_size: int,
     reference_point: np.ndarray,
 ) -> np.ndarray:
-    assert not np.any(reference_point - rank_i_loss_vals <= 0)
+    if not np.isfinite(reference_point).all():
+        return rank_i_indices[:subset_size]
+    diff_of_loss_vals_and_ref_point = reference_point - rank_i_loss_vals
```
nabenabe0928, line 49: Btw, can we revert here?

eukaryo: I think the current defensive code is better for when `_solve_hssp_on_unique_loss_vals` is called from other code in the future.

nabenabe0928: Ok, then let's keep the current code!

nabenabe0928:
@not522 
Could you review this PR?

not522:
Could you remove `filterwarnings` from `test_hssp.py`? I think that they are no longer needed.

not522 (approved):
LGTM

PR #5435, opened 2024-05-08
Guarantee `weights_below` to be finite in MOTPE

eukaryo:
## Motivation

Resolve https://github.com/optuna/optuna/issues/5374

## Description of the changes

I added code to guarantee `weights_below` to be finite in MOTPE. Weights fall-back to uniform if it contains inf, nan, or a value large enough to be inf when summed.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -516,8 +516,7 @@ class TPESampler(BaseSampler):
             weights_below = _calculate_weights_below_for_multi_objective(
                 study, trials, self._constraints_func
             )[param_mask_below]
-            if not np.isfinite(weights_below).all():
-                weights_below = np.clip(1.0 - np.isfinite(weights_below), EPS, 1.0)
+            assert np.isfinite(weights_below).all()
             mpe = self._parzen_estimator_cls(
                 observations, search_space, self._parzen_estimator_parameters, weights_below
             )
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -516,6 +516,8 @@ class TPESampler(BaseSampler):
             weights_below = _calculate_weights_below_for_multi_objective(
                 study, trials, self._constraints_func
             )[param_mask_below]
+            if not np.isfinite(weights_below.sum()):
+                weights_below = np.ones_like(weights_below)
             mpe = self._parzen_estimator_cls(
                 observations, search_space, self._parzen_estimator_parameters, weights_below
             )
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -516,8 +516,8 @@ class TPESampler(BaseSampler):
             weights_below = _calculate_weights_below_for_multi_objective(
                 study, trials, self._constraints_func
             )[param_mask_below]
-            if not np.isfinite(weights_below.sum()):
-                weights_below = np.ones_like(weights_below)
+            if not np.isfinite(weights_below).all():
+                weights_below = 1.0 - np.isfinite(weights_below)
             mpe = self._parzen_estimator_cls(
                 observations, search_space, self._parzen_estimator_parameters, weights_below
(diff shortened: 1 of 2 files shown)


nabenabe0928:
Probably, what we should do is as follows:

```python
def _calculate_leave_one_out_hypervolume_contributions(
    loss_vals: np.ndarray, ref_point: np.ndarray
) -> np.ndarray:
    n_below = len(loss_vals)
    on_front = _is_pareto_front(loss_vals, assume_unique_lexsorted=False)
    pareto_sols = loss_vals[on_front]
    contribs = np.zeros(n_below, dtype=float)
    hv = WFG().compute(pareto_sols, reference_point)
    if np.isfinite(hv):
        leave_one_out_masks = ~np.eye(pareto_sols.shape[0]).astype(bool)
        hvs_leave_one_out = np.asarray(
            [WFG().compute(pareto_sols[loo], ref_point) for loo in leave_one_out_masks]
        )
        contribs[on_front] = hv - hvs_leave_one_out
    else:
        _, inv, pareto_sol_counts = np.unique(
            pareto_sols, axis=0, return_counts=True, return_inverse=True
        )
        on_front_and_not_duplicated = np.zeros(n_below, dtype=bool)
        # pareto_sol_counts[inv, i] is how many times we see pareto_sols[i] in pareto_sols.
        on_front_and_not_duplicated[on_front] = pareto_sol_counts[inv] > 1
        contribs[on_front_and_not_duplicated] = 1.0

    return contribs
```

eukaryo:
Thank you for your detailed advice! However, some of them are beyond the scope of this PR because it is just intended to fix the issue.
Currently I believe that:
1. if the `reference_point` contains non-finite value, it is reasonable to fallback all `weights_below` to 1.0 because all of them should be non-finite values. Specifically, [this line](https://github.com/optuna/optuna/blob/ab958d42e44cad2e0517907278d55ca7432ea32c/optuna/samplers/_tpe/sampler.py#L815) results nan under such condition. (This is not numpy subtraction, so it does not cause `RuntimeWarning`.)
1. Elseif the trial exists that contains non-finite value (`-inf`), the initial PR code will fallback all `weights_below` to 1.0. However, I have reconsidered that the `weights_below` for trials that do not contain `-inf` should be zero.
1. Mathematically, trials with non-finite values are incomparable, so it is justified to assume that all of them are on the pareto front and set the `weights_below` to 1.0.

optuna/samplers/_tpe/sampler.py
```diff
@@ -814,8 +816,8 @@ def _calculate_weights_below_for_multi_objective(
         contributions = np.asarray(
             [hv - WFG().compute(lvals[indices_mat[i]], reference_point) for i in range(n_below)]
         )
-        contributions += EPS
-        weights_below = np.clip(contributions / np.max(contributions), 0, 1)
+        contributions = np.clip(contributions, EPS, None)
+        weights_below = np.clip(contributions / np.max(contributions), EPS, 1.0)
```
nabenabe0928, line 820: If the debugging is the purpose, what about the following and removing the line in `_build_parzen_estimator`?
```suggestion
        contributions[np.isnan(contributions)] = np.inf
        max_contribution = np.max(contributions)
        if not np.isfinite(max_contribution):
            weights_below = np.ones_like(contributions, dtype=float)
        else:
            weights_below = np.clip(contributions / max_contribution, EPS, 1)
```

eukaryo: I thought `contributions = np.clip(contributions, EPS, None)` is slightly better than `contributions += EPS` just because `contributions` can be close to `sys.float_info.max`; in that case, adding eps results inf.
However, since that rarely happens, I reconsidered and agree that your suggestion is better.

nabenabe0928 (approved):
LGTM

nabenabe0928:

eukaryo:
@y0z Could you review this PR? (I am aware that you are very busy with another higher priority project; this re-assignment could be a formality.)

y0z (approved):
LGTM

PR #5438, opened 2024-05-09
Enable single pruner in multi-objective optimization

eukaryo:
## Motivation

- I want to enable pruning in multi-objective optimization.
- I want not to change DB Schema in this PR; I think that updating DB and enabling multiple pruners are in future work.

## Description of the changes

- Enabled reporting scalar intermediate value in multi objective optimization.
- Added an optional argument to some pruners, which determines the direction of pruner. The argument must be set by user for multi-objective optimization, and should not be set for single-objective optimization.
- Added unittests.
- Added codes and comments to define the extended specification of values in FrozenTrial of which TrialState is PRUNED.


eukaryo:
At this time, I am aware that the docstring has not yet been added. I will work on it soon.
All other code is review-ready.

PR #5448, opened 2024-05-15
Implement `download_artifact`

porink0424:
## Motivation

Although `upload_artifact` existed, its counterpart did not exist. This PR introduces a new API `download_artifact` as the counterpart. In order to have symmetry with `upload_artifact`, `download_artifact` has an interface that specifies `file_path` as an argument.

## Description of the changes

- Added `download_artifact`
- Added test for `download_artifact`

### DIFF
--- a/optuna/artifacts/_download.py
+++ b/optuna/artifacts/_download.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import shutil
 
 from optuna._experimental import experimental_func
@@ -18,5 +19,8 @@ def download_artifact(artifact_store: ArtifactStore, artifact_id: str, file_path
         file_path:
             A path to save the downloaded artifact.
     """
+    if os.path.exists(file_path):
+        raise FileExistsError(f"File already exists: {file_path}")
+
     with artifact_store.open_reader(artifact_id) as reader, open(file_path, "wb") as writer:
         shutil.copyfileobj(reader, writer)
--- a/optuna/artifacts/_download.py
+++ b/optuna/artifacts/_download.py
@@ -1,10 +1,9 @@
 from __future__ import annotations
 
+import shutil
+
 from optuna._experimental import experimental_func
 from optuna.artifacts._protocol import ArtifactStore
-
-
-BUFFER_SIZE = 1024 * 1024  # 1MB
 
 
 @experimental_func("4.0.0")
@@ -20,8 +19,4 @@ def download_artifact(artifact_store: ArtifactStore, artifact_id: str, file_path
             A path to save the downloaded artifact.
     """
     with artifact_store.open_reader(artifact_id) as reader, open(file_path, "wb") as writer:
-        while True:
-            chunk = reader.read(BUFFER_SIZE)
-            if not chunk:  # `chunk` becomes an empty string at EOL.
-                break
-            writer.write(chunk)
+        shutil.copyfileobj(reader, writer)
--- a/optuna/artifacts/_download.py
+++ b/optuna/artifacts/_download.py
@@ -1,0 +1,26 @@
+from __future__ import annotations
+
+from optuna._experimental import experimental_func
+from optuna.artifacts._protocol import ArtifactStore
+
+
+@experimental_func("4.0.0")
+def download_artifact(
+    artifact_store: ArtifactStore,
+    artifact_id: str,
+    file_path: str,
+) -> None:
+    """Download an artifact from the artifact store.
+
+    Args:
+        artifact_store:
+            An artifact store.
+        artifact_id:
+            The identifier of the artifact to download.
+        file_path:
+            A path to download the artifact to.
+    """
+    with artifact_store.open_reader(artifact_id) as f:
+        content = f.read()
+    with open(file_path, "wb") as f:
+        f.write(content)
--- a/optuna/artifacts/_download.py
+++ b/optuna/artifacts/_download.py
@@ -4,6 +4,9 @@ from optuna._experimental import experimental_func
 from optuna.artifacts._protocol import ArtifactStore
 
 
+BUFFER_SIZE = 1024 * 1024  # 1MB
+
+
 @experimental_func("4.0.0")
 def download_artifact(
     artifact_store: ArtifactStore,
@@ -20,7 +23,9 @@ def download_artifact(
         file_path:
             A path to download the artifact to.
     """
-    with artifact_store.open_reader(artifact_id) as f:
-        content = f.read()
-    with open(file_path, "wb") as f:
-        f.write(content)
+    with artifact_store.open_reader(artifact_id) as reader, open(file_path, "wb") as writer:
+        while True:
+            chunk = reader.read(BUFFER_SIZE)
+            if not chunk:
+                break
+            writer.write(chunk)
--- a/optuna/artifacts/_download.py
+++ b/optuna/artifacts/_download.py
@@ -22,6 +22,6 @@ def download_artifact(artifact_store: ArtifactStore, artifact_id: str, file_path
     with artifact_store.open_reader(artifact_id) as reader, open(file_path, "wb") as writer:
         while True:
             chunk = reader.read(BUFFER_SIZE)
-            if not chunk:
+            if not chunk:  # `chunk` becomes an empty string at EOL.
                 break
             writer.write(chunk)
--- a/optuna/artifacts/_download.py
+++ b/optuna/artifacts/_download.py
@@ -8,11 +8,7 @@ BUFFER_SIZE = 1024 * 1024  # 1MB
 
 
 @experimental_func("4.0.0")
-def download_artifact(
-    artifact_store: ArtifactStore,
-    artifact_id: str,
-    file_path: str,
-) -> None:
+def download_artifact(artifact_store: ArtifactStore, artifact_id: str, file_path: str) -> None:
     """Download an artifact from the artifact store.
 
     Args:
@@ -21,7 +17,7 @@ def download_artifact(
         artifact_id:
             The identifier of the artifact to download.
         file_path:
-            A path to download the artifact to.
+            A path to save the downloaded artifact.
     """
     with artifact_store.open_reader(artifact_id) as reader, open(file_path, "wb") as writer:
         while True:
--- a/tests/artifacts_tests/test_download_artifact.py
+++ b/tests/artifacts_tests/test_download_artifact.py
@@ -1,0 +1,40 @@
+from __future__ import annotations
+
+import pathlib
+import pytest
+
+import optuna
+from optuna.artifacts._protocol import ArtifactStore
+from optuna.artifacts import FileSystemArtifactStore
+from optuna.artifacts._download import download_artifact
+from optuna.artifacts._upload import upload_artifact
+
+
+@pytest.fixture(params=["FileSystem"])
+def artifact_store(tmp_path: pathlib.PurePath, request: pytest.FixtureRequest) -> ArtifactStore:
+    if request.param == "FileSystem":
+        return FileSystemArtifactStore(str(tmp_path))
+    assert False, f"Unknown artifact store: {request.param}"
+
+
+def test_download_artifact(tmp_path: pathlib.PurePath, artifact_store: ArtifactStore) -> None:
+    storage = optuna.storages.InMemoryStorage()
+    study = optuna.create_study(storage=storage)
+    artifact_ids: list[str] = []
+
+    def objective(trial: optuna.Trial) -> float:
+        x = trial.suggest_int("x", 0, 100)
+        y = trial.suggest_int("y", 0, 100)
+        dummy_file = str(tmp_path / f"dummy_{trial.number}.txt")
+        with open(dummy_file, "w") as f:
+            f.write(f"{x} {y}")
+        artifact_ids.append(upload_artifact(trial, dummy_file, artifact_store))
+        return x**2 + y**2
+
+    study.optimize(objective, n_trials=5)
+
+    for i, artifact_id in enumerate(artifact_ids):
+        dummy_downloaded_file = str(tmp_path / f"dummy_downloaded_{i}.txt")
+        download_artifact(artifact_store, artifact_id, dummy_downloaded_file)
+        with open(dummy_downloaded_file, "r") as f:
+            assert f.read() == f"{study.trials[i].params['x']} {study.trials[i].params['y']}"
--- a/tests/artifacts_tests/test_download_artifact.py
+++ b/tests/artifacts_tests/test_download_artifact.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
 import pathlib
+
 import pytest
 
 import optuna
-from optuna.artifacts._protocol import ArtifactStore
 from optuna.artifacts import FileSystemArtifactStore
 from optuna.artifacts._download import download_artifact
+from optuna.artifacts._protocol import ArtifactStore
 from optuna.artifacts._upload import upload_artifact
 
 
--- a/tests/artifacts_tests/test_download_artifact.py
+++ b/tests/artifacts_tests/test_download_artifact.py
@@ -19,8 +19,7 @@ def artifact_store(tmp_path: pathlib.PurePath, request: pytest.FixtureRequest) -
 
 
 def test_download_artifact(tmp_path: pathlib.PurePath, artifact_store: ArtifactStore) -> None:
-    storage = optuna.storages.InMemoryStorage()
-    study = optuna.create_study(storage=storage)
+    study = optuna.create_study()
     artifact_ids: list[str] = []
 
     def objective(trial: optuna.Trial) -> float:
--- a/tests/artifacts_tests/test_download_artifact.py
+++ b/tests/artifacts_tests/test_download_artifact.py
@@ -5,10 +5,10 @@ import pathlib
 import pytest
 
 import optuna
+from optuna.artifacts import download_artifact
 from optuna.artifacts import FileSystemArtifactStore
-from optuna.artifacts._download import download_artifact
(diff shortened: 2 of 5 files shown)


tests/artifacts_tests/test_download_artifact.py
```diff
@@ -0,0 +1,41 @@
...
+
+import pathlib
+
+import pytest
+
+import optuna
+from optuna.artifacts import FileSystemArtifactStore
+from optuna.artifacts._download import download_artifact
+from optuna.artifacts._protocol import ArtifactStore
+from optuna.artifacts._upload import upload_artifact
```
toshihikoyanase, line 11: Generally, we test public functions/classes instead of private ones if possible (c.f., [Optuna's test policy](https://github.com/optuna/optuna/wiki/Test-Policy#code-quality)). How about importing these from `optuna.artifacts`, excluding the private ones?

toshihikoyanase (changes requested):
Thank you for your great proposal. I've actually been waiting for this feature.

Let me share my early comment.

porink0424:
@toshihikoyanase 
Thank you for your comments! I have fixed the points,

optuna/artifacts/_download.py
```diff
@@ -0,0 +1,26 @@
...
+    """Download an artifact from the artifact store.
+
+    Args:
+        artifact_store:
+            An artifact store.
+        artifact_id:
+            The identifier of the artifact to download.
+        file_path:
+            A path to download the artifact to.
+    """
```
toshihikoyanase, line 22: Could you add an entry for this function to the Artiface reference?
I'd like to check the rendered document.

https://github.com/optuna/optuna/blob/f954e3529d81ae156a9e91d31b50c89b390a3f67/docs/source/reference/artifacts.rst?plain=1#L20

optuna/artifacts/_download.py
```diff
@@ -0,0 +1,26 @@
...
+    with artifact_store.open_reader(artifact_id) as f:
+        content = f.read()
+    with open(file_path, "wb") as f:
+        f.write(content)
```
toshihikoyanase, line 26: [Notes] I guess we can use the streaming to keep the memory size small, but it seems to be over-engineering.

```python
    # Set buffer size to 1MB.
    buffer_size = 1024 * 1024
    with artifact_store.open_reader(artifact_id) as reader, open(file_path, "wb") as writer:
        while True:
            chunk = reader.read(buffer_size)
            if not chunk:
                break
            writer.write(chunk)
```

tests/artifacts_tests/test_download_artifact.py
```diff
@@ -0,0 +1,41 @@
...
+        artifact_ids.append(upload_artifact(trial, dummy_file, artifact_store))
+        return x**2 + y**2
+
+    study.optimize(objective, n_trials=5)
+
+    for i, artifact_id in enumerate(artifact_ids):
+        dummy_downloaded_file = str(tmp_path / f"dummy_downloaded_{i}.txt")
+        download_artifact(artifact_store, artifact_id, dummy_downloaded_file)
+        with open(dummy_downloaded_file, "r") as f:
+            assert f.read() == f"{study.trials[i].params['x']} {study.trials[i].params['y']}"
```
toshihikoyanase, line 41: I'm not sure, but can we use temporary files/directories to keep the environment clean?

toshihikoyanase (commented):
Thanks for your update!
I added two suggestions and a note.

tests/artifacts_tests/test_download_artifact.py
porink0424: Dummy files actually remain on `tmp_path`, so the machine where this test is run would not be kept completely clean.
Since I implemented this test referring to `test_upload_artifact`, which does not keep the environment clean as well, we may need to modify both tests if we do not want this situation.

nabenabe0928:
Could you add an explanation of `download_artifact` to `tutorial/20_recipes/012_artifact_tutorial.py`?

optuna/artifacts/_download.py
```diff
@@ -0,0 +1,31 @@
...
+
+BUFFER_SIZE = 1024 * 1024  # 1MB
+
+
+@experimental_func("4.0.0")
+def download_artifact(
+    artifact_store: ArtifactStore,
+    artifact_id: str,
+    file_path: str,
+) -> None:
```
nabenabe0928, line 15:
```suggestion
def download_artifact(artifact_store: ArtifactStore, artifact_id: str, file_path: str) -> None:
```

optuna/artifacts/_download.py
```diff
@@ -0,0 +1,31 @@
...
+    """Download an artifact from the artifact store.
+
+    Args:
+        artifact_store:
+            An artifact store.
+        artifact_id:
+            The identifier of the artifact to download.
+        file_path:
+            A path to download the artifact to.
```
nabenabe0928, line 24: What does this mean?
Did you mean `A path to save the downloaded artifact.`?
```suggestion
            A path to save the downloaded artifact.
```

optuna/artifacts/_download.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    with artifact_store.open_reader(artifact_id) as reader, open(file_path, "wb") as writer:
+        while True:
+            chunk = reader.read(BUFFER_SIZE)
+            if not chunk:
```
nabenabe0928, line 29: As `chunk` becomes `None` at end of line, it is better to use `chunk is None` because `chunk is None` is a stricter condition.
https://docs.python.org/ja/3/library/io.html#io.RawIOBase.read
```suggestion
            if chunk is None:  # `chunk` becomes `None` at EOL.
```

optuna/artifacts/_download.py
```diff
@@ -0,0 +1,31 @@
...
+                break
+            writer.write(chunk)
```
nabenabe0928, line 31: Although it is not important to guarantee the file safety here because the downloaded file will not be used by Optuna, I have a concern about the file lock here.
Could we, first of all, check whether the `file_path` already exists at `open`?

nabenabe0928 (changes requested):
Thank you for the PR, I left some comments!
Could you please take another look?

optuna/artifacts/_download.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import annotations
+
+from optuna._experimental import experimental_func
+from optuna.artifacts._protocol import ArtifactStore
+
+
+BUFFER_SIZE = 1024 * 1024  # 1MB
```
nabenabe0928, line 7: [Question] How did you decide this number?
It is just my opinion, but I think it is better to set this number as high as possible considering loop counts and the overhead we incur.

optuna/artifacts/_download.py
toshihikoyanase: @nabenabe0928 Thank you for your careful review.

Let me clarify my understanding of the following sentence.
Does this change always read the file with non-blocking mode?

> If 0 bytes are returned, and size was not 0, this indicates end of file. If the object is in non-blocking mode and no bytes are available, None is returned.

tests/artifacts_tests/test_download_artifact.py
toshihikoyanase: Sorry for the delayed response.
Temporary files can sometimes complicate the execution of git commands. So, we might not want to leave them to minimize operational mistakes.

> Since I implemented this test referring to `test_upload_artifact`, which does not keep the environment clean as well, we may need to modify both tests if we do not want this situation.

Thank you for pointing it out. For the scope of this PR, let us solely concentrate on the tests of downloading, and we can work on the other files in follow-up PRs.

optuna/artifacts/_download.py
toshihikoyanase: That makes sense. Could you give us some recommendations?

tests/artifacts_tests/test_download_artifact.py
porink0424: According to [this](https://github.com/pytest-dev/pytest/issues/8141#issuecomment-1278960826), pytest keeps 3 temporary directories by default, and we can prevent pytest from saving temporary directories by setting `tmp_path_retention_policy`.
So, adding just one line `tmp_path_retention_policy = "none"` to `pyproject.toml` would resolve this problem. Should I include this fix in this PR?

optuna/artifacts/_download.py
porink0424: @nabenabe0928 
Thank you for your comments!
Do you mean that we should check if a file exists at `file_path`, and raise error if it does?

optuna/artifacts/_download.py
porink0424: I don't have a strong reason here. 
How much do you think is a good value to set? How about 1GB?

optuna/artifacts/_download.py
nabenabe0928: @toshihikoyanase 
You are right, and sorry for the confusion.
I modified the suggestion as follows:

```suggestion
            if not chunk:  # `chunk` becomes an empty string at EOL.
```

optuna/artifacts/_download.py
nabenabe0928: I think that is kinder.
But warning or error should be discussed more closely.
At least, `os.makedirs` checks whether the target already exists, for example.
https://docs.python.org/ja/3/library/os.html#os.makedirs

tests/artifacts_tests/test_download_artifact.py
```diff
@@ -0,0 +1,41 @@
...
+@pytest.fixture(params=["FileSystem"])
+def artifact_store(tmp_path: pathlib.PurePath, request: pytest.FixtureRequest) -> ArtifactStore:
+    if request.param == "FileSystem":
+        return FileSystemArtifactStore(str(tmp_path))
+    assert False, f"Unknown artifact store: {request.param}"
+
+
+def test_download_artifact(tmp_path: pathlib.PurePath, artifact_store: ArtifactStore) -> None:
+    storage = optuna.storages.InMemoryStorage()
+    study = optuna.create_study(storage=storage)
```
not522, line 23: `InMemoryStorage` is the default storage, and it can be omitted.

```suggestion
    study = optuna.create_study()
```

tests/artifacts_tests/test_download_artifact.py
not522: My understanding is that there are two concerns, neither of which is a problem, so I think this PR is fine as is.
The first is that temporary files complicate git management, but since `tmp_path` is created in a system temporary directory, it does not affect git. (https://docs.pytest.org/en/7.3.x/how-to/tmp_path.html#the-default-base-temporary-directory)
The second is that the temporary files affect other test runs, but since `tmp_path` is unique for each function invocation, this is not a problem. (https://docs.pytest.org/en/7.3.x/reference/reference.html#tmp-path)

optuna/artifacts/_download.py
```diff
@@ -0,0 +1,27 @@
...
+            The identifier of the artifact to download.
+        file_path:
+            A path to save the downloaded artifact.
+    """
+    with artifact_store.open_reader(artifact_id) as reader, open(file_path, "wb") as writer:
+        while True:
+            chunk = reader.read(BUFFER_SIZE)
+            if not chunk:
+                break
+            writer.write(chunk)
```
c-bata, line 27: How about using `shutil.copyfileobj()` here?

```suggestion
    with artifact_store.open_reader(artifact_id) as reader, open(file_path, "wb") as writer:
        shutil.copyfileobj(reader, writer)
```

https://docs.python.org/3/library/shutil.html#shutil.copyfileobj

optuna/artifacts/_download.py
porink0424: Instead of looping over manually, I decided to use `shutil.copyfileobj()` and let the default values of this function take care of this problem. Thank you for the discussion.

optuna/artifacts/_download.py
porink0424: I have added the check whether a file exists.

tutorial/20_recipes/012_artifact_tutorial.py
```diff
@@ -139,6 +139,13 @@ def objective(trial: optuna.Trial) -> float:
 
     print(content)
 
+Also, you can easily download the artifact as a file using `download_artifact` function, instead of using the artifact module:
```
c-bata, line 142: I have two suggestions.

1. Could you remove the code that uses `artifact_store.open_reader()` since it is not a public api? It was just a workaround for `download_artifact()`.
2. Could you add a link on the `download_artifact` to the api doc like below?

```suggestion
Also, you can easily download the artifact as a file using :func:`~optuna.artifacts.download_artifact` function, instead of using the artifact module:
```

porink0424: Thank you for your review! I have removed `open_reader()` from the tutorial.

tests/artifacts_tests/test_download_artifact.py
toshihikoyanase: Ah, I misunderstood. These test cases employ the `tmp_path` fixture instead of creating temporary files in the test cases. Thanks for your explanation.

not522 (approved):
LGTM!

toshihikoyanase (commented):
Sorry for the delayed response. My comments have been resolved.
@nabenabe0928 @c-bata Please merge after your review.

tutorial/20_recipes/012_artifact_tutorial.py
```diff
@@ -132,12 +133,11 @@ def objective(trial: optuna.Trial) -> float:
...
-    # Loading and displaying artifacts associated with the best trial.
-    best_artifact_id = study.best_trial.user_attrs.get("artifact_id")
-    with artifact_store.open_reader(best_artifact_id) as f:
-        content = f.read().decode("utf-8")
 
-    print(content)
+    # Downloading artifacts associated with the best trial.
+    best_artifact_id = study.best_trial.user_attrs.get("artifact_id")
+    download_file_path = ...  # Set the path to save the downloaded artifact.
+    download_artifact(artifact_store, best_artifact_id, download_file_path)
```
nabenabe0928, line 140:
```suggestion
    download_artifact(artifact_store, best_artifact_id, download_file_path)
    with open(download_file_path, "r") as f:
        content = f.read().decode("utf-8")
    print(content)
```

tutorial/20_recipes/012_artifact_tutorial.py
```diff
@@ -213,12 +214,11 @@ def objective(trial: optuna.Trial) -> float:
...
```
nabenabe0928, line 221:
```suggestion
    download_artifact(artifact_store, best_artifact_id, download_file_path)
    with open(download_file_path, "r") as f:
        content = f.read().decode("utf-8")
    print(content)
```

tutorial/20_recipes/012_artifact_tutorial.py
```diff
@@ -384,11 +385,16 @@ def main():
...
 
     best_artifact_id = study.best_trial.user_attrs["structure"]
-    with artifact_store.open_reader(best_artifact_id) as f:
-        content = f.read().decode("utf-8")
-    best_atoms = json_to_atoms(content)
-    print(best_atoms)
-    write("best_atoms.png", best_atoms, rotation=("315x,0y,0z"))
+
+    with tempfile.TemporaryDirectory() as tmpdir_name:
+        download_file_path = f"{tmpdir_name}/{best_artifact_id}.json"
```
nabenabe0928, line 390:
```suggestion
        download_file_path = os.path.join(tmpdir_name, f"{best_artifact_id}.json")
```

tutorial/20_recipes/012_artifact_tutorial.py
```diff
@@ -384,11 +385,16 @@ def main():
...
+        download_artifact(artifact_store, best_artifact_id, download_file_path)
+
+        with open(download_file_path) as f:
+            content = f.read()
```
nabenabe0928, line 394:
```suggestion
        with open(download_file_path, "r") as f:
            content = f.read().decode("utf-8")
```

nabenabe0928 (commented):
Thank you for the update!
I left some minor comments!

tutorial/20_recipes/012_artifact_tutorial.py
c-bata: Your proposed code seems to open `download_file_path` in text mode. Could you explain why `.decode("utf-8")` is necessary?

tutorial/20_recipes/012_artifact_tutorial.py
c-bata: ditto.

tutorial/20_recipes/012_artifact_tutorial.py
nabenabe0928: Oh sorry, you are right.
`mode` was supposed to be `rb`, but not `r`. 
I modified the suggestions.

c-bata: I see. I have another question. Do you think it's necessary to call `print(content)` in this tutorial? In my opinion, we don't need to stick to keeping the previous code. Avoiding the call to `print(content)` makes the tutorial simpler and more readable.

What do you think?

nabenabe0928: In my honest opinion, it is better to have these lines for the sanity check as tutorials are, in general, like a hello-world for the target feature.
Yet I agree that we do not need these lines if the readers completely understand what they are doing in each line.

not522: I agree with @nabenabe0928's comment to add the code to open the file. I think that if the tutorial code ends with `download_artifact`, users need to read the API reference to understand how to use the downloaded artifact.

nabenabe0928 (approved):

PR #5449, opened 2024-05-17
Replace `abc.ABC` with `typing.Protocol` in `BaseStorage`

porink0424:
## Motivation

Same motivation as #5425. 

## Description of the changes

- Replace `abc.ABC` with `typing.Protocol` in `BaseStorage`.
- Added `typing_extensions` dependency when python version is less than 3.8.

## Note

Similar concerns as in #5425 apply here. Dynamic type checking would be no longer available such as `isinstance(storage, BaseStorage)`, `issubclass(cls, BaseStorage)`. The migration guide will be needed.

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import abc
+from abc import abstractmethod
 from collections.abc import Container
 from collections.abc import Sequence
 from typing import Any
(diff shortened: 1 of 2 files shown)


nabenabe0928:
We discussed whether we should introduce protocol to `BaseStorage`, but we decided to close this PR because `BaseStorage` is too complex for giving its interface as a duck type, which is explicitly discouraged by [PEP544](https://peps.python.org/pep-0544/).

PR #5462, opened 2024-05-27
Fix `_log_complete_trial` for constrained optimization

not522:
## Motivation

`_log_complete_trial` assumes that `study.best_trial` does not raise `ValueError` because the latest trial is `COMPLETE` (Note: if no trial is `COMPLETE`, `best_trial` raises `ValueError`). But, after #5426 was merged, it raises `ValueError` if no trial is **feasible**.

## Description of the changes

This PR added try-except for the constrained optimization and fixed the logging message.

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1116,12 +1116,6 @@ class Study:
                 )
             )
         elif len(trial.values) == 1:
-            try:
-                best_trial = self.best_trial
-            except ValueError:
-                # If no feasible trials are completed yet, study.best_trial raises ValueError.
-                best_trial = None
-
             trial_value: float | dict[str, float]
             if metric_names is None:
                 trial_value = trial.values[0]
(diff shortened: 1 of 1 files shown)


not522:
@c-bata @HideakiImamura Could you review this PR?

optuna/study/study.py
```diff
@@ -1111,27 +1111,30 @@ def _log_completed_trial(self, trial: FrozenTrial) -> None:
...
-                    best_trial.number,
-                    best_trial.value,
-                )
+
+            message = (
+                f"Trial {trial.number} finished with value: {trial_value} and parameters: "
+                f"{trial.params}."
             )
+            if best_trial is not None:
+                message += f" Best is trial {best_trial.number} with value: {best_trial.value}."
```
c-bata, line 1136: A nit: I would say we can simplify the code a little as follows:

```suggestion
            try:
                best_trial = self.best_trial
                message += f" Best is trial {best_trial.number} with value: {best_trial.value}."
            except ValueError:
                # If no feasible trials are completed yet, study.best_trial raises ValueError.
                pass
```

c-bata (approved):
LGTM with a nit.

c-bata:
Should we add a bug label, by the way? The problem was introduced in #5426, which is not released yet.

not522:
Thank you for your review. I updated it as you suggested.

> Should we add a bug label, by the way? The problem was introduced in #5426, which is not released yet.

Not a strong opinion, but as a type of PR it is a bug fix, so I think it can stay as is.

HideakiImamura (approved):
LGTM.

PR #5464, opened 2024-05-29
Cache split in TPE for high-dimensional optimization

nabenabe0928:
## Motivation

As TPE significantly slows down for high-dimensional optimization, this PR introduces caching mechanism for the TPE split.

## Description of the changes

- Cache the split information in sampler so that we can re-use the result

For `n_trials=1000` and `dim=10`, the runtimes are the following:

|This PR |Master|
|:--:|:--:|
|29|47|

### DIFF
--- a/optuna/_hypervolume/utils.py
+++ b/optuna/_hypervolume/utils.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import numpy as np
 
 
--- a/optuna/_hypervolume/utils.py
+++ b/optuna/_hypervolume/utils.py
@@ -1,48 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
-
 import numpy as np
-
-from optuna.study._multi_objective import _is_pareto_front
-
-
-if TYPE_CHECKING:
-    import sortedcontainers
-else:
-    from optuna._imports import _LazyImport
-
-    sortedcontainers = _LazyImport("sortedcontainers")
-
-
-def _compute_3d(solution_set: np.ndarray, reference_point: np.ndarray) -> float:
-    hv = 0.0
-    # NOTE(nabenabe0928): The indices of Y and Z in the sorted lists are the reverse of each other.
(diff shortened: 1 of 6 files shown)


not522:
What is the relationship between this PR and #5454? Should we review it after #5454?

nabenabe0928:
@not522 
These two PRs are orthogonal works, so we can separately work!
This PR aims to share the split information in a trial.
The other PR aims to share the information over multiple trials giving the same set of arguments to the hssp solver.

not522:
@eukaryo @gen740 Could you review this PR?

optuna/_hypervolume/utils.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
gen740, line 1: I believe this change is not relevant to this PR.
This change would be good to include when using features from ﻿`__future__`.

eukaryo:
Sorry, I am temporarily busy because my computer is not working, and I suppose @HideakiImamura -san is the appropriate reviewer. @HideakiImamura, could you review this PR?

optuna/_hypervolume/utils.py
nabenabe0928: Done!

HideakiImamura (commented):
Holding cache on the sampler instance causes the data inconsistency across different processes. Could you give me an opinion that?

nabenabe0928:
We discussed internally and decided to close this PR.

nabenabe0928:
Just for future reminder, I will leave some comments:

### This PR does not cause any issues between processes

As each trial is sampled in a thread and a sampling of a specific trial will not be scattered on multiple processes or threads, we do not have to be concerned about the missing cached data in another  process.

However, when using multiple threads, we need to take care of another thread overwriting the cache before one trial completes its sampling.
For this reason, I added a buffer to store the split data up to the latest 64 trials in a thread.

Anyways, missing cache does not cause any issue because we simply need to re-calculate the split, which incurs some more computational burden.

PR #5466, opened 2024-05-29
Fix rdb heartbeat

eukaryo:
## Motivation

I want to handover PR https://github.com/optuna/optuna/pull/5447 

## Description of the changes

- Fixed retrying mechanism in heartbeat
- Added with_for_update to the Heartbeat table only if the query writes something to the database

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -100,7 +100,7 @@ def _create_scoped_session(
             message = (
                 "An exception is raised during the commit. "
                 "This typically happens due to invalid data in the commit, "
-                "e.g. exceeding max length. "
+                "e.g., exceeding max length. "
             )
             raise optuna.exceptions.StorageInternalError(message) from e
     except sqlalchemy_exc.SQLAlchemyError as e:
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -464,7 +464,7 @@ class RDBStorage(BaseStorage, BaseHeartbeat):
                 except sqlalchemy_exc.OperationalError:
                     if n_retries >= MAX_RETRIES:
                         raise
-                    time.sleep((2 ** n_retries) + random.random())
+                    time.sleep((2**n_retries) + random.random())
 
                 n_retries += 1
 
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -448,8 +448,10 @@ class RDBStorage(BaseStorage, BaseHeartbeat):
 
         # Retry a couple of times. Deadlocks may occur in distributed environments.
         n_retries = 0
-        MAX_RETRIES = 3
+        MAX_RETRIES = 5
         while n_retries <= MAX_RETRIES:
+            if n_retries != 0:
+                time.sleep((2**n_retries) * (1.0 + random.random()))
             with _create_scoped_session(self.scoped_session) as session:
                 try:
                     # Ensure that that study exists.
@@ -458,40 +460,35 @@ class RDBStorage(BaseStorage, BaseHeartbeat):
                     # atomic operation. More precisely, the trial number computed in
                     # `_get_prepared_new_trial` is prone to race conditions without this lock.
                     models.StudyModel.find_or_raise_by_id(study_id, session, for_update=True)
-
                     trial = self._get_prepared_new_trial(study_id, template_trial, session)
+                    if template_trial:
+                        frozen = copy.deepcopy(template_trial)
+                        frozen.number = trial.number
+                        frozen.datetime_start = trial.datetime_start
+                        frozen._trial_id = trial.trial_id
+                    else:
+                        frozen = FrozenTrial(
+                            number=trial.number,
+                            state=trial.state,
+                            value=None,
+                            values=None,
+                            datetime_start=trial.datetime_start,
+                            datetime_complete=None,
+                            params={},
+                            distributions={},
+                            user_attrs={},
+                            system_attrs={},
+                            intermediate_values={},
+                            trial_id=trial.trial_id,
+                        )
                     break  # Successfully created trial.
                 except sqlalchemy_exc.OperationalError:
-                    if n_retries >= MAX_RETRIES:
-                        raise
-                    time.sleep((2**n_retries) + random.random())
+                    n_retries += 1
+                    raise
 
-                n_retries += 1
+        assert n_retries <= MAX_RETRIES, "The _create_new_trial was repeatedly failed."
 
-            if template_trial:
-                frozen = copy.deepcopy(template_trial)
-                frozen.number = trial.number
-                frozen.datetime_start = trial.datetime_start
-                frozen._trial_id = trial.trial_id
-            else:
-                frozen = FrozenTrial(
-                    number=trial.number,
-                    state=trial.state,
-                    value=None,
-                    values=None,
-                    datetime_start=trial.datetime_start,
-                    datetime_complete=None,
-                    params={},
-                    distributions={},
-                    user_attrs={},
-                    system_attrs={},
-                    intermediate_values={},
-                    trial_id=trial.trial_id,
-                )
-
-            return frozen
-        assert False
-        return FrozenTrial()
+        return frozen
 
     def _get_prepared_new_trial(
         self,
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -468,7 +468,6 @@ class RDBStorage(BaseStorage, BaseHeartbeat):
             # TODO(takizawa): Ponder if this exponential backoff is useful.
             if n_retries != 0:
                 time.sleep((2**n_retries) * (1.0 + random.random()))
-            n_retries += 1
             with _create_scoped_session(self.scoped_session, ignore_timeout_error=True) as session:
                 try:
                     # Ensure that that study exists.
@@ -501,6 +500,7 @@ class RDBStorage(BaseStorage, BaseHeartbeat):
                     break  # Successfully created trial.
                 except sqlalchemy_exc.OperationalError:
                     raise
+            n_retries += 1
 
         assert n_retries <= MAX_RETRIES, "The _create_new_trial was repeatedly failed."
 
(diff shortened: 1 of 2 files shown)


HideakiImamura:

eukaryo:
I have measured performance on my PC.

result: (bold means median)

with heartbeat, v3.6.1
speedtest No.0 finished. HEAVINESS=1000000. elapsed time = 0:00:42.551121
speedtest No.1 finished. HEAVINESS=1000000. elapsed time = 0:00:42.582342
speedtest No.2 finished. HEAVINESS=1000000. elapsed time = 0:00:42.835726
speedtest No.3 finished. HEAVINESS=1000000. elapsed time = 0:00:42.795765
**speedtest No.4 finished. HEAVINESS=1000000. elapsed time = 0:00:42.768239**

without heartbeat, v3.6.1
speedtest No.0 finished. HEAVINESS=1000000. elapsed time = 0:00:42.734669
speedtest No.1 finished. HEAVINESS=1000000. elapsed time = 0:00:43.343668
speedtest No.2 finished. HEAVINESS=1000000. elapsed time = 0:00:42.655627
**speedtest No.3 finished. HEAVINESS=1000000. elapsed time = 0:00:42.728745**
speedtest No.4 finished. HEAVINESS=1000000. elapsed time = 0:00:42.671716

with heartbeat, v3.6.1+PR https://github.com/optuna/optuna/pull/5466 (using FOR UPDATE only if write)
speedtest No.0 finished. HEAVINESS=1000000. elapsed time = 0:00:43.476506
speedtest No.1 finished. HEAVINESS=1000000. elapsed time = 0:00:43.552098
speedtest No.2 finished. HEAVINESS=1000000. elapsed time = 0:00:43.386705
speedtest No.3 finished. HEAVINESS=1000000. elapsed time = 0:00:43.612788
**speedtest No.4 finished. HEAVINESS=1000000. elapsed time = 0:00:43.490778**

with heartbeat, v3.6.1+PR https://github.com/optuna/optuna/pull/5447 (using FOR UPDATE everytime)
speedtest No.0 finished. HEAVINESS=1000000. elapsed time = 0:00:43.919765
speedtest No.1 finished. HEAVINESS=1000000. elapsed time = 0:00:44.192499
**speedtest No.2 finished. HEAVINESS=1000000. elapsed time = 0:00:44.270614**
speedtest No.3 finished. HEAVINESS=1000000. elapsed time = 0:00:44.628502
speedtest No.4 finished. HEAVINESS=1000000. elapsed time = 0:00:44.827820


setup mysql (before each experiment):
```
DROP DATABASE IF EXISTS optuna;
CREATE DATABASE optuna;
CREATE USER 'optuna'@'localhost' IDENTIFIED BY 'your_password';
GRANT ALL PRIVILEGES ON optuna.* TO 'optuna'@'localhost';
FLUSH PRIVILEGES;
exit;
```
python code:
```
import datetime
import hashlib
import uuid

import optuna

HEAVINESS = 1000000


def objective(trial):
    x = hashlib.sha256(str(x).encode()).hexdigest()
    for _ in range(HEAVINESS):
        x = hashlib.sha256(x.encode()).hexdigest()
    return float(int(x, 16))


result = []
for _ in range(5):
        url="mysql+pymysql://optuna:your_password@localhost/optuna",
        heartbeat_interval=60,  # Comment out this line for experiments w/o heartbeat.
    )
    study = optuna.create_study(study_name=str(uuid.uuid4()), storage=storage)

    start = datetime.datetime.now()
    study.optimize(objective, n_trials=100, n_jobs=12)
    end = datetime.datetime.now()
    result.append(end - start)

for i in range(5):
    print(f"speedtest No.{i} finished. {HEAVINESS=}. elapsed time = {result[i]}")
```

optuna/storages/_rdb/storage.py
```diff
@@ -81,6 +84,25 @@ def _create_scoped_session(
...
+        if ignore_timeout_error and all([s in str(e.orig) for s in error_keystrings]):
+            _logger.debug(
+                "Ignoring {}. This happens due to a timeout. No exception is propagated here"
+                "because expecting a retry to be made by the calling function.".format(repr(e))
+            )
+        else:
+            message = (
+                "An exception is raised during the commit. "
+                "This typically happens due to invalid data in the commit, "
+                "e.g. exceeding max length. "
```
nabenabe0928, line 103:
```suggestion
                "e.g., exceeding max length. "
```

optuna/storages/_rdb/storage.py
```diff
@@ -81,6 +84,25 @@ def _create_scoped_session(
...
+        session.rollback()
+
+        # The following keystrings depend on the backend DB.
+        # If another issue about another DB is raised, the keystrings will be updated.
+        error_keystrings = ("1205", "Lock wait timeout exceeded; try restarting transaction")
+
+        if ignore_timeout_error and all([s in str(e.orig) for s in error_keystrings]):
+            _logger.debug(
+                "Ignoring {}. This happens due to a timeout. No exception is propagated here"
+                "because expecting a retry to be made by the calling function.".format(repr(e))
```
nabenabe0928, line 97:
```suggestion
                f"Ignoring {repr(e)}. This happens due to a timeout. No exception is propagated "
                "here because expecting a retry to be made by the calling function."
```

optuna/storages/_rdb/storage.py
```diff
@@ -81,6 +84,25 @@ def _create_scoped_session(
...
         else:
             raise
+    except sqlalchemy_exc.OperationalError as e:
+        session.rollback()
+
+        # The following keystrings depend on the backend DB.
+        # If another issue about another DB is raised, the keystrings will be updated.
+        error_keystrings = ("1205", "Lock wait timeout exceeded; try restarting transaction")
+
+        if ignore_timeout_error and all([s in str(e.orig) for s in error_keystrings]):
```
nabenabe0928, line 94:
```suggestion
        if ignore_timeout_error and all(s in str(e.orig) for s in error_keystrings):
```

optuna/storages/_rdb/storage.py
```diff
@@ -81,6 +84,25 @@ def _create_scoped_session(
             )
         else:
             raise
+    except sqlalchemy_exc.OperationalError as e:
+        session.rollback()
+
+        # The following keystrings depend on the backend DB.
+        # If another issue about another DB is raised, the keystrings will be updated.
+        error_keystrings = ("1205", "Lock wait timeout exceeded; try restarting transaction")
```
nabenabe0928, line 92: This is very fragile to version differences as well, so I think we need to discuss a bit more.

optuna/storages/_rdb/storage.py
```diff
@@ -446,46 +468,50 @@ def _create_new_trial(
...
-                    datetime_start=trial.datetime_start,
-                    datetime_complete=None,
-                    params={},
-                    distributions={},
-                    user_attrs={},
-                    system_attrs={},
-                    intermediate_values={},
-                    trial_id=trial.trial_id,
-                )
+        assert n_retries <= MAX_RETRIES, "The _create_new_trial was repeatedly failed."
```
nabenabe0928, line 512: I think it is more appropriate to use `raise RuntimeError` or something similar, but not `assert`.

optuna/storages/_rdb/storage.py
```diff
@@ -446,46 +468,50 @@ def _create_new_trial(
 
         # Retry a couple of times. Deadlocks may occur in distributed environments.
         n_retries = 0
-        with _create_scoped_session(self.scoped_session) as session:
-            while True:
+        MAX_RETRIES = 5
+        while n_retries <= MAX_RETRIES:
+            # TODO(takizawa): Ponder if this exponential backoff is useful.
+            if n_retries != 0:
+                time.sleep((2**n_retries) * (1.0 + random.random()))
```
nabenabe0928, line 475: TBD

optuna/storages/_rdb/storage.py
```diff
@@ -446,46 +468,50 @@ def _create_new_trial(
 
         # Retry a couple of times. Deadlocks may occur in distributed environments.
         n_retries = 0
-        with _create_scoped_session(self.scoped_session) as session:
-            while True:
+        MAX_RETRIES = 5
+        while n_retries <= MAX_RETRIES:
```
nabenabe0928, line 472: We can also code like this:

```python
for _ in range(MAX_RETRIES):
    with ... as session:
        try:
            trial = ...
            return _create_frozen_trial_from_trial(...)
        except ...:
            raise
else:
    raise RuntimeError(...)
```
Instead of the current while loop.

nabenabe0928 (commented):
Thank you for the PR, I left some comments!

HideakiImamura:
I have a question about the performance measurement. It seems that the test script run the 100 trials per each study, take a heartbeat by 60 seconds, and take 42 ~ 45 seconds over one run. It means that the each heartbeat only took 1 time at the beginning of the trials. Is this intended? If so, could this performance measurement validate the implementation of this PR? (If I miss the motivation of your experiment, please tell me the details.)

optuna/storages/_rdb/storage.py
eukaryo: Agreed. I will remove this condition. Arbitrary `OperationalError` will be catched and retried. In other words, I will assume that errors that retry has no effect will not occur.

optuna/storages/_rdb/storage.py
eukaryo: I choiced this backoff timescale to mitigate congestion, but I rethinked and it is outside the scope of Optuna's responsibility. (User should scale-up the DB server) I will chance the timescale to 1/100.

optuna/storages/_rdb/storage.py
eukaryo: Agreed.

optuna/storages/_rdb/storage.py
```diff
@@ -446,46 +468,50 @@ def _create_new_trial(
...
+                            values=None,
+                            datetime_start=trial.datetime_start,
+                            datetime_complete=None,
+                            params={},
+                            distributions={},
+                            user_attrs={},
+                            system_attrs={},
+                            intermediate_values={},
+                            trial_id=trial.trial_id,
+                        )
```
HideakiImamura, line 504: I think these lines should be placed outside of the `while` loop. It would be better to use another scoped session, since these are read only steps.

optuna/storages/_rdb/storage.py
eukaryo: Agreed. I will revise. Fow now, I think that `raise sqlalchemy_exc.OperationalError` is appropriate.

PR #5474, opened 2024-06-04
Revise docs to specify ArtifactStore's methods as non-public

c-bata:
## Motivation

Like `optuna.samplers.BaseStorage`, the methods defined in `ArtifactStore` protocol are not intended to call directly by library users. Source code that calls ArtifactStore's method directly may be broken in the future release.

## Description of the changes

I updated the documentation and the docstring to let users know methods in ArtifactStore are private.

### DIFF
--- a/optuna/artifacts/_protocol.py
+++ b/optuna/artifacts/_protocol.py
@@ -16,6 +16,8 @@ if TYPE_CHECKING:
 class ArtifactStore(Protocol):
     """A protocol defining the interface for an artifact backend.
 
+    Methods defined in this protocol are not supposed to be directly intended by library users.
+
     An artifact backend is responsible for managing the storage and retrieval
     of artifact data. The backend should provide methods for opening, writing
     and removing artifacts.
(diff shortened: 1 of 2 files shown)


HideakiImamura (approved):
LGTM.

optuna/artifacts/_protocol.py
```diff
@@ -16,6 +16,8 @@
 class ArtifactStore(Protocol):
     """A protocol defining the interface for an artifact backend.
 
+    Methods defined in this protocol are not supposed to be directly intended by library users.
```
nabenabe0928, line 19:
```suggestion
    The methods defined in this protocol are not supposed to be directly called by library users.
```

nabenabe0928 (commented):
Thank you for the PR!
I left a minor comment!

optuna/artifacts/_protocol.py
c-bata: Thank you for your suggestion!

c-bata:
Thank you for your review. PTAL.

nabenabe0928 (approved):

PR #5482, opened 2024-06-06
1205Fix rdb heartbeat

takuya4880:


takuya4880:
Sorry. Just a mistake.

Issue #5489, opened 2024-06-11
GridSampler: Duplicates


ldv1:
### Expected behavior

I use the GridSampler.
In 3.2.0 it runs fine. I upgraded to 3.6.1, and now I see duplicates. If I run the code below 3 times, I might get x=0, then x=1, and then x=0 again. I hadn't such an issue with Optuna 3.2.0

```
import time
import optuna
import random
print(f"optuna: {optuna.__version__}")

def objective(trial: optuna.Trial):
    x = trial.suggest_int("x", -100, 100)
    print("Using parameters:", trial.params)
    time.sleep(20)
    return x**2

if __name__ == '__main__':
    search_space = { "x": [0, 1, 2] }
    sampler = optuna.samplers.GridSampler(search_space)
    storage = optuna.storages.RDBStorage(url="sqlite:////home/me/a.db", engine_kwargs={"connect_args": {"timeout": 100}})
    study = optuna.create_study(storage=storage,
                                sampler=sampler,
                                direction="minimize",
                                pruner=optuna.pruners.NopPruner(),
                                study_name="mystudy",
                                load_if_exists=True)
    study.optimize(objective, n_trials=3, n_jobs=1, callbacks=[optuna.study.MaxTrialsCallback(3, states=None)])
```

### Environment

- Optuna version:3.6.1
- Python version:3.11.4
- OS:Linux-5.14.21-150500.55.39-default-x86_64-with-glibc2.31

### Error messages, stack traces, or logs

```shell
python3 a.py&
[1] 4927
[I 2024-06-11 22:22:22,097] A new study created in RDB with name: mystudy
Using parameters: {'x': 0}

python3 a.py&
[2] 5026
[I 2024-06-11 22:22:26,466] Using an existing study with name 'mystudy' instead of creating a new one.
Using parameters: {'x': 1}

python3 a.py&
[3] 5117
[I 2024-06-11 22:22:31,327] Using an existing study with name 'mystudy' instead of creating a new one.
Using parameters: {'x': 0}
```

### Steps to reproduce

Run the code above.

nzw0301:
Thank you for sharing the reproducible code! I think this is not a bug. As explained in https://optuna.readthedocs.io/en/stable/tutorial/20_recipes/001_rdb.html#resume-study, the sampler state is not stored in the storage.

nzw0301:
Sorry, I misunderstood the issue description. I think the distributed optimisation is performed here. In this case,

[...]

PR #4918, opened 2023-09-11
Refactor _grid.py

eukaryo:
## Motivation

I want to refactor and speed-up the grid-sampler.

## Description of the changes

- I utilized `before_trial`. Codes in `infer_relative_search_space` were moved into `before_trial`.
- For the first N trials, the pre-shuffled `_all_grids` were to be executed in order, based on ~`_trial_id`~`trial.number`.
  - This change reduces the number of calls of `_get_unvisited_grid_ids` function, which is heavy and tends to output the same parameters multiple times.
  - This change may resolve or mitigate https://github.com/optuna/optuna/issues/4627 .

PR #5490, opened 2024-06-13
Use a fixed seed value when `seed=None` in `GridSampler`

nzw0301:
## Motivation

Resolve https://github.com/optuna/optuna/issues/5489. I'm happy to hear the opinion on this change because I'm not sure why this shuffle was introduced. 

## Description of the changes

~Remove the shuffle.~

For distributed optimisation, optuna docs suggest not setting `seed` for samplers or set different values. In this PR, call `shuffle` when `seed=0`.

### DIFF
--- a/optuna/samplers/_grid.py
+++ b/optuna/samplers/_grid.py
@@ -125,9 +125,10 @@ class GridSampler(BaseSampler):
         self._all_grids = list(itertools.product(*self._search_space.values()))
         self._param_names = sorted(search_space.keys())
         self._n_min_trials = len(self._all_grids)
-        self._rng = LazyRandomState(seed or 0)
-        if seed is not None:
-            self._rng.rng.shuffle(self._all_grids)  # type: ignore[arg-type]
+        if seed is None:
+            seed = 0
+        self._rng = LazyRandomState(seed)
+        self._rng.rng.shuffle(self._all_grids)  # type: ignore[arg-type]
 
     def reseed_rng(self) -> None:
         self._rng.rng.seed()
--- a/optuna/samplers/_grid.py
+++ b/optuna/samplers/_grid.py
@@ -198,9 +198,6 @@ class GridSampler(BaseSampler):
             message = "The parameter name, {}, is not found in the given grid.".format(param_name)
             raise ValueError(message)
 
-        # TODO(c-bata): Reduce the number of duplicated evaluations on multiple workers.
-        # Current selection logic may evaluate the same parameters multiple times.
-        # See https://gist.github.com/c-bata/f759f64becb24eea2040f4b2e3afce8f for details.
         grid_id = trial.system_attrs["grid_id"]
         param_value = self._all_grids[grid_id][self._param_names.index(param_name)]
         contains = param_distribution._contains(param_distribution.to_internal_repr(param_value))
--- a/optuna/samplers/_grid.py
+++ b/optuna/samplers/_grid.py
@@ -104,11 +104,10 @@ class GridSampler(BaseSampler):
         seed:
             A seed to fix the order of trials as the grid is randomly shuffled. This shuffle is
             beneficial when the number of grids is larger than ``n_trials`` in
-            :meth:`~optuna.Study.optimize` not to suggest similar grids. Please note that it is
-            recommended using the same ``seed`` or :obj:`None` in distributed optimization
-            settings because internally each grid is assigned to an integer after shuffling grids.
-            Otherwise, this sampler may increase the number of duplicate suggestions during
-            distributed optimization.
+            :meth:`~optuna.Study.optimize` to suppress suggesting similar grids. If ``seed`` is
+            :obj:`None`, the shuffle does not happen. Please note that fixing ``seed`` for each
+            process is strongly recommended in distributed optimization to avoid duplicated
+            suggestions.
     """
 
     def __init__(
--- a/optuna/samplers/_grid.py
+++ b/optuna/samplers/_grid.py
@@ -122,7 +122,9 @@ class GridSampler(BaseSampler):
         self._all_grids = list(itertools.product(*self._search_space.values()))
         self._param_names = sorted(search_space.keys())
         self._n_min_trials = len(self._all_grids)
-        self._rng = LazyRandomState(seed)
+        self._rng = LazyRandomState(seed or 0)
+        if seed is not None:
+            self._rng.rng.shuffle(self._all_grids) # type: ignore[arg-type]
 
     def reseed_rng(self) -> None:
         self._rng.rng.seed()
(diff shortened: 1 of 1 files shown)


eukaryo:
FYI: This shuffling is introduced in PR https://github.com/optuna/optuna/pull/4918.

eukaryo:
When I introduced this PR, I believed that `GridSampler` should search in a random order. This is because when `n_trials < len(self._all_grids)`, there is no guarantee that the regions searched earlier will have good values. In such cases, `RandomSampler` tends to evaluate the same point multiple times, whereas `GridSampler` (with shuffling) is expected to evaluate each point once without duplication. Of course, I agree that the issue should be resolved. However, I believe that the lack of a shuffling option results in a performance loss.

optuna/samplers/_grid.py
```diff
@@ -123,7 +123,6 @@ def __init__(
         self._param_names = sorted(search_space.keys())
         self._n_min_trials = len(self._all_grids)
         self._rng = LazyRandomState(seed)
```
eukaryo, line 125:
```suggestion
        self._rng = LazyRandomState(seed or 0)
        if seed is not None:
            self._rng.rng.shuffle(self._all_grids) # type: ignore[arg-type]
```

eukaryo (commented):
I have suggested a change. PTAL.

eukaryo:
@not522 could you review this PR? I suppose you are the appropriate reviewer because you were also a reviewer of PR https://github.com/optuna/optuna/pull/4918.

nzw0301:
Thank you for your explanation! I agree with you to shuffle the grid!

nzw0301:
Should I add a note on this? This is still issue when user specify different random seeds among processes.

eukaryo:
Yes, it would be a great help if you could do that.

nzw0301:
Thanks. Sure!

eukaryo (approved):
LGTM!

not522:
This PR makes the order of trials fixed for `seed=None`. It will be a breaking change, so I'm thinking how to resolve #5489 without this problem...

eukaryo:
In my current understanding, the requirements can be organized and enumerated as follows:

1. If `seed=None`, search order should be shuffled. (From the historical background; initial implementation bahaves that way.)
1. If `seed=None`, search order should be shuffled differently each time. (Because someone may have set up an experimental system that assumes this (expecially in a non-distributed environment).)
1. If `seed=None`, duplicates should not occur in a distributed environment (issue#5489).
1. `_get_unvisited_grid_ids` calls `get_all_trials` internally, which is heavy and I wanted to avoid calling it (PR#4918). Additionally, the process including `_get_unvisited_grid_ids` is not atomic, so it cannot prevent duplication perfectly.

The second request means that the `self._rng = LazyRandomState(seed)` should be resulted identical across all processes of any distibuted environment when `seed=None`.

If we want all processes to have an unique value in a distributed environment, the only way is to share the value using storage and appropriate transaction process. The only time when this is possible is not during `__init__`, but during `before_trial`, because it needs `study` received as an argument. Therefore, it is simplest to perform transaction processing and shuffling as necessary at the time of `before_trial`.

This discussion is somewhat complicated, but since it involves historical circumstances and a distributed environment, I believe such complexity is inevitable and unavoidable.

To implement the above plan, we need to enable `set_study_system_attr` to do "write only if it does not exist, otherwise do nothing". It is necessary to add an optional argument or a new method to `BaseStorage` and all derived storages.

eukaryo:
After discussion with @not522 -san and others, I have reconsidered the second of the above requirements is not necessary. (It is a little strange that only GridSampler behaves deterministic with `seed=None` while other Samplers etc. behave stochastic, but I think it is acceptable if explained in the docstring.) Adding functionality to the storage module should be done with the most caution and should not be tied to this PR, nor is it necessary.

optuna/samplers/_grid.py
```diff
@@ -102,10 +102,13 @@ def objective(trial):
...
-            it is not recommended using this option in distributed optimization settings since
-            this option cannot ensure the order of trials and may increase the number of duplicate
-            suggestions during distributed optimization.
+            A seed to fix the order of trials as the grid is randomly shuffled. This shuffle is
+            beneficial when the number of grids is larger than ``n_trials`` in
+            :meth:`~optuna.Study.optimize` not to suggest similar grids. Please note that it is
+            recommended using the same ``seed`` or :obj:`None` in distributed optimization
+            settings because internally each grid is assigned to an integer after shuffling grids.
+            Otherwise, this sampler may increase the number of duplicate suggestions during
+            distributed optimization.
```
nabenabe0928, line 111: I modified the expressions.
```suggestion
            A seed to fix the order of trials as the grid is randomly shuffled. This shuffle is
            beneficial when the number of grids is larger than ``n_trials`` in
            :meth:`~optuna.Study.optimize` to suppress suggesting similar grids. If ``seed`` is
            :obj:`None`, the shuffle does not happen. Please note that fixing ``seed`` for each
            process is strongly recommended in distributed optimization to avoid duplicated
            suggestions.
```

nabenabe0928 (commented):
Thank you for the PR!
I added a suggestion to the doc-string.

optuna/samplers/_grid.py
```diff
@@ -122,8 +125,9 @@ def __init__(
         self._all_grids = list(itertools.product(*self._search_space.values()))
         self._param_names = sorted(search_space.keys())
         self._n_min_trials = len(self._all_grids)
-        self._rng = LazyRandomState(seed)
-        self._rng.rng.shuffle(self._all_grids)  # type: ignore[arg-type]
+        self._rng = LazyRandomState(seed or 0)
+        if seed is not None:
+            self._rng.rng.shuffle(self._all_grids)  # type: ignore[arg-type]
```
not522, line 130: How about always shuffling? I think it is not proper to use the grid in order only when `seed=None`.

```suggestion
        if seed is None:
            seed = 0
        self._rng = LazyRandomState(seed)
        self._rng.rng.shuffle(self._all_grids)  # type: ignore[arg-type]
```

nabenabe0928 (commented):
@eukaryo @not522 
Do you think we can remove this TODO comment?
- https://github.com/optuna/optuna/pull/5490/files#diff-87630c6e088d5b9840e0ec55c5c4b4b813f653d780f41d8ef889646578d5ba0aL201-L203

nzw0301:
Thank you for your comments and discussions. I've updated the PR by following the suggestions.

optuna/samplers/_grid.py
```diff
@@ -122,6 +123,8 @@ def __init__(
         self._all_grids = list(itertools.product(*self._search_space.values()))
         self._param_names = sorted(search_space.keys())
         self._n_min_trials = len(self._all_grids)
+        if seed is None:
+            seed = 0
         self._rng = LazyRandomState(seed)
```
nabenabe0928, line 128:
```suggestion
        self._rng = LazyRandomState(seed or 0)
```

nabenabe0928 (commented):
Thank you for the changes!
I added another nit comment.

not522 (approved):
Thank you for your update. It looks good to me.
@nabenabe0928 If you approve this PR, please merge it.

nabenabe0928 (approved):
LGTM!

PR #5496, opened 2024-06-17
Lock the trial heartbeat only if writes

eukaryo:
## Motivation

I want to handover PR https://github.com/optuna/optuna/pull/5447.
I also want to clarify by separating single issue in PR https://github.com/optuna/optuna/pull/5466 , which involves multiple issues and becomes complicated.

## Description of the changes

I added `with_for_update` to the query about heartbeat mechanism, only if the query writes something to the database.

PR #5497, opened 2024-06-17
Bugfix and refactor `_create_new_trial`

eukaryo:
## Motivation

I want to refactor `_create_new_trial`
I want `_create_new_trial` to generate no empty trial_id when transaction processing fails.

(note: I want to clarify by separating single issue in PR https://github.com/optuna/optuna/pull/5466 , which involves multiple issues and becomes complicated.)

## Description of the changes

In the current implementation, `with _create_scoped_session` is outside and `while` loop is inside, so if the transaction fails in `_get_prepared_new_trial`, Optuna retries without rollback of the preceding `find_or_raise_by_id`. As a result, empty `trial_id` can be generated. (I confirmed the behaviour by a simple experiment that inserts a line raising an exception at `random` in `_get_prepared_new_trial`.)
In the proposed implementation, `with _create_scoped_session` is inside the for loop, so a rollback occurs against every fail of transaction.

## Note
In the current inplementation, only `sqlalchemy_exc.OperationalError:` will be caught, and other exceptions are re-raised as is. I believe the proposed implementation does not change the behaviour.

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -500,10 +500,6 @@ class RDBStorage(BaseStorage, BaseHeartbeat):
                 # Should unavoidable exceptions be raised, the last exception is propagated
                 # after five retries.
                 error_obj = e
-                pass
-            except Exception as e:
-                error_obj = e
-                raise
         assert error_obj is not None
         raise error_obj
 
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -8,6 +8,8 @@ from datetime import timedelta
 import json
 import logging
 import os
+import random
+import time
 from typing import Any
 from typing import Callable
 from typing import Container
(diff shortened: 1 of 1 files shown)


eukaryo:
PR https://github.com/optuna/optuna/pull/5466 proposed to change `_create_scoped_session`, whereas the present PR does not require any change of `_create_scoped_session`, and the present PR and PR https://github.com/optuna/optuna/pull/5496 will cover the all features in PR https://github.com/optuna/optuna/pull/5466.

eukaryo:
I will add unittest(s) if possible; I assume it is not difficult and will be done today. The unittest(s) will confirm that the `try-except`s works as I intended, and will improve the code coverage.

optuna/storages/_rdb/storage.py
```diff
@@ -444,48 +446,68 @@ def _create_new_trial(
...
-                    # Locking within a study is necessary since the creation of a trial is not an
-                    # atomic operation. More precisely, the trial number computed in
-                    # `_get_prepared_new_trial` is prone to race conditions without this lock.
-                    models.StudyModel.find_or_raise_by_id(study_id, session, for_update=True)
-
-                    trial = self._get_prepared_new_trial(study_id, template_trial, session)
-                    break  # Successfully created trial.
-                except sqlalchemy_exc.OperationalError:
-                    if n_retries > 2:
+        # Retry maximum five times. Deadlocks may occur in distributed environments.
```
nabenabe0928, line 449: I think it is better to make the nested part a bit clearer by shortening the line counts.
```python
        def _create_frozen_trial(
            trial: "models.TrialModel", template_trial: FrozenTrial | None
        ) -> FrozenTrial:
            if template_trial:
                frozen = copy.deepcopy(template_trial)
                frozen.number = trial.number
                frozen.datetime_start = trial.datetime_start
                frozen._trial_id = trial.trial_id
            else:
                frozen = FrozenTrial(
                    number=trial.number,
                    state=trial.state,
                    value=None,
                    values=None,
                    datetime_start=trial.datetime_start,
                    datetime_complete=None,
                    params={},
                    distributions={},
                    user_attrs={},
                    system_attrs={},
                    intermediate_values={},
                    trial_id=trial.trial_id,
                )

        # Retry maximum five times. Deadlocks may occur in distributed environments.
        error_obj: Exception | None = None
        trial: "models.TrialModel" | None = None
        for n_retries in range(5):
            try:
                with _create_scoped_session(self.scoped_session) as session:
                    try:
                        # Ensure that that study exists.
                        #
                        # Locking within a study is necessary since the creation of a trial is not
                        # an atomic operation. More precisely, the trial number computed in
                        # `_get_prepared_new_trial` is prone to race conditions without this lock.
                        models.StudyModel.find_or_raise_by_id(study_id, session, for_update=True)
                        trial = self._get_prepared_new_trial(study_id, template_trial, session)
                    except Exception as e:
                        # `sqlalchemy_exc.OperationalError` will be caught within the function.
                        # The others will be immediately propagated to the caller.
                        error_obj = e
                        raise
            except sqlalchemy_exc.OperationalError:
                # Note: According to SQLAlchemy specifications,
                # `sqlalchemy_exc.OperationalError` can be raised in situations where
                # retries are not effective (e.g., input string is too long).
                #
                # It is assumed here that such bugs do not exist and retries are effective.
                # Should such bugs be present, the content of the `OperationalError` will be
                # propagated at the end of this function, and details will be recorded in
                # the error log.
                # This backoff is to solve retries caused by deadlock.
                # This is not an EXPONENTIAL backoff that reduces DB server congestion,
                # but it is intentional.
                # Optuna's philosophy is that it is better for the DB administrator to detect
                # such saturation and augment the server than for Optuna itself to detect
                # DB congestion and slow it down.
                time.sleep(random.random() * 2.0)
            else:
                assert trial is not None  # MyPy redefinition.
                return _create_frozen_trial(trial, template_trial)

        assert error_obj is not None
        raise error_obj
```

nabenabe0928 (commented):
Thank you for the PR, I left a comment!

optuna/storages/_rdb/storage.py
nabenabe0928: Btw, what happens if we code like this?

```python
        error_obj: Exception | None = None
        trial: "models.TrialModel" | None = None
        for n_retries in range(5):
            try:
                with _create_scoped_session(self.scoped_session) as session:
                    models.StudyModel.find_or_raise_by_id(study_id, session, for_update=True)
                    trial = self._get_prepared_new_trial(study_id, template_trial, session)
            except sqlalchemy_exc.OperationalError:
                time.sleep(random.random() * 2.0)
            except Exception as e:
                error_obj = e
            else:
                assert trial is not None  # MyPy redefinition.
                return _create_frozen_trial(trial, template_trial)

        assert error_obj is not None
        raise error_obj
```

More specifically, what is the difference between the two below?

<details>
<summary>Code 1</summary>

```python
            try:
                with _create_scoped_session(self.scoped_session) as session:
                    ...
            except sqlalchemy_exc.OperationalError:
                ...
            except Exception as e:
                error_obj = e
            else:
                ...
                return frozen

            raise error_obj
```

</details>

<details>
<summary>Code 2</summary>

```python
            try:
                with _create_scoped_session(self.scoped_session) as session:
                    try:
                        ...
                    except Exception as e:
                        error_obj = e
                        raise e
            except sqlalchemy_exc.OperationalError:
                ...
            else:
                ...
                return frozen

            raise error_obj
```
</details>

eukaryo: Good catch! Indeed, since we have replaced `while` outside of `with`, it is sufficient to do `try` only outside of `with`.
(nit) We must do `error_obj = e` inside `except sqlalchemy_exc.OperationalError` too. This is because if `except sqlalchemy_exc.OperationalError` fails five times in a row, the error log of the last failure must be returned.

eukaryo: In the current (v3.6.1) code, "`_create_frozen_trial`" is inside the `with` scope. I had followed the implementation. Currently I am not sure that the processing does not require the session. Specifically, `_create_scoped_session` uses `session.close()`, but I believe this is a mistake (should really use `session.remove()`). In my understanding, `session.close()` does not release resources until garbage-collector works. So even if unittests passed, it is not clear that we can safely move "`_create_frozen_trial`" outside the `with` scope.

optuna/storages/_rdb/storage.py
```diff
@@ -444,48 +446,68 @@ def _create_new_trial(
...
-            return frozen
+            except sqlalchemy_exc.OperationalError:
+                # Note: According to SQLAlchemy specifications,
+                # `sqlalchemy_exc.OperationalError` can be raised in situations where
+                # retries are not effective (e.g., input string is too long).
+                #
+                # It is assumed here that such bugs do not exist and retries are effective.
+                # Should such bugs be present, the content of the `OperationalError` will be
+                # propagated at the end of this function, and details will be recorded in
+                # the error log.
```
HideakiImamura, line 507: I think what we want to say is that the unavoidable `OperationalError` would be raised after 5 retries, isn't it?

eukaryo: Yes. I revised the comments.

optuna/storages/_rdb/storage.py
eukaryo: I refactored the code with your suggestion.

optuna/storages/_rdb/storage.py
```diff
@@ -444,48 +446,66 @@ def _create_new_trial(
...
+                # `sqlalchemy_exc.OperationalError` can be raised in situations where
+                # retries are not effective (e.g., input string is too long).
+                #
+                # It is assumed here that exceptions are avoidable and retries are effective.
+                # Should unavoidable exceptions be raised, the last exception is propagated
+                # after five retries.
+                error_obj = e
+                pass
+            except Exception as e:
+                error_obj = e
```
nabenabe0928, line 505:
```suggestion
            except Exception:
```

optuna/storages/_rdb/storage.py
```diff
@@ -444,48 +446,66 @@ def _create_new_trial(
...
+                raise
```
nabenabe0928, line 506:
```suggestion
```

optuna/storages/_rdb/storage.py
```diff
@@ -444,48 +446,66 @@ def _create_new_trial(
...
+        # Retry maximum five times. Deadlocks may occur in distributed environments.
+        error_obj: Exception | None = None
+        for n_retries in range(5):
+            if n_retries != 0:
+                # This backoff is to solve retries caused by deadlock.
+                # This is not an EXPONENTIAL backoff that reduces DB server congestion,
+                # but it is intentional.
+                # Optuna's philosophy is that it is better for the DB administrator to detect
+                # such saturation and augment the server than for Optuna itself to detect
+                # DB congestion and slow it down.
```
nabenabe0928, line 482: Enhancements of comments are Followup

eukaryo: Agreed.

nabenabe0928 (approved):
LGTM!

HideakiImamura (approved):
LGTM.

PR #5504, opened 2024-06-21
Make `WFG` a function

nabenabe0928:
## Motivation

As the hypervolume class looks unnecessary, I replaced it with a function.

## Description of the changes

- Replace `WFG` with a function


HideakiImamura:
The `_compute_2d` is already private. It this change necessary?

HideakiImamura:
There is as possibility that `_compute_2d` is used by another hypervolume computation algorithms.

nabenabe0928:
> There is a possibility that `_compute_2d` is used by another hypervolume computation algorithms.

WFG is inarguably known to be the best algorithm for `n_objectives >= 7` [1], so there is no need to change the algorithm for `n_objectives >= 7`.

For `3 <= n_objectives <= 6`, there are some better algorithms compared to `WFG`.
But they require a special data structure.

For example, 3D and 4D require balanced binary tree [1] ([sortedcontainers](https://pypi.org/project/sortedcontainers/)).
5D and 6D require kd tree, quadtree or segment tree [2].

IMO, it is a bit harder for me to read the code, so I honestly feel we do not even need `WFG` class.

But probably, we can still separate `_compute_2d`.

[1] https://dl.acm.org/doi/pdf/10.1145/3453474
[2] https://arxiv.org/pdf/1510.01963

eukaryo:
Thank you for your PR! I think the recent comment and the current PR implementation looks generally reasonable.

However, I would prefer not to use nested functions in the implementation of intricate algorithms such as this PR, because nested functions are inherently confusing. In a concrete example, `_compute_hv` can refer to `solution_set` as if it were a global variable, but it does not actually refer to it, because it has `sorted_sols`. (As a minor suggestion, `sorted_sols` should have a clearer name, such as `sorted_solutions`.)

Regarding the future, if an easy-to-implement algorithm for `3 <= n_objectives <= 6` emerges, there is a concern that the `_compute_hv` function will become even larger if it is also implemented as another nested function.

optuna/_hypervolume/wfg.py
```diff
@@ -2,61 +2,75 @@
...
-            intersec = np.prod(self._reference_point - np.maximum(sorted_sols[0], sorted_sols[1]))
+            intersec = np.prod(reference_point - np.maximum(sorted_sols[0], sorted_sols[1]))
             return np.sum(inclusive_hvs) - intersec
 
         limited_sols_array = np.maximum(sorted_sols[:, np.newaxis], sorted_sols)
         return sum(
-            self._compute_exclusive_hv(limited_sols_array[i, i + 1 :], inclusive_hv)
+            _compute_exclusive_hv(limited_sols_array[i, i + 1 :], inclusive_hv)
             for i, inclusive_hv in enumerate(inclusive_hvs)
         )
```
HideakiImamura, line 58: How about moving this function (and others placed below) outside of `compute_hypervolume`?

nabenabe0928:
@eukaryo 
> I would prefer not to use nested functions in the implementation of intricate algorithms such as this PR

I made these functions nested just to reduce the line differences in GitHub and I can work on the refactoring as a followup.

eukaryo (approved):
I see. LGTM, at least except for nested-function-related issues which seem not critical or urgent. This PR makes the code better.
I think resolving function nesting can be done in this PR, instead of separate follow-up PR. Either is fine.

optuna/_hypervolume/wfg.py
nabenabe0928: @HideakiImamura 
I addressed your suggestion!

https://github.com/optuna/optuna/pull/5504/commits/718b56e979aa7638d17a90ec67804350d8f220fd

@eukaryo 
As I transferred the nested functions outside, we do not need the followup anymore:)

HideakiImamura (approved):
Looks very good.

PR #5507, opened 2024-06-22
Add atomic operations that write an attribute only if the key is absent

eukaryo:
## Motivation

I want to solve PR https://github.com/optuna/optuna/pull/5490.
I does not want to cause any breaking change or affect any public interface.
I want to empower OptunaHub contributors from the features to be introduced in this PR.

## Description of the changes

This PR will add the atomic operations that write value only if key does not exist, to all tha `storage`s. Supporting destinations are all `(study|trial)_(system|user)_attrs`.

### DIFF
--- a/optuna/storages/_base.py
+++ b/optuna/storages/_base.py
@@ -89,10 +89,10 @@ class BaseStorage(abc.ABC):
         raise NotImplementedError
 
     @abc.abstractmethod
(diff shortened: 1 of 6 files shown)


eukaryo:
Please let this PR close as the team is very cautious about changing functionality to storage.

PR #5508, opened 2024-06-24
Add version constraint `< 3.0.0` to NumPy

nabenabe0928:
## Motivation

This PR is to prevent Optuna from failing without any patch release when another major version of NumPy is released.

## Description of the changes

- Add a version constraint to NumPy


nabenabe0928:
As the version constraint can be avoided by a patch release and this PR unnecessarily narrow down the usability, I will close this PR.

PR #5519, opened 2024-06-25
Remove skipped tests for BoTorch with Python 3.12

nabenabe0928:
## Motivation

As `BoTorch` was removed from the dependencies, I also remove the skip set for BoTorch with Python 3.12. 

## Description of the changes

- Simply remove the if statement for the skip


not522 (approved):
LGTM!

PR #5523, opened 2024-06-26
Fix the version conflicts in the workflow for the minimum version constraints

### LINKED ISSUE #5499: Wrong testing in tests with minimum versions
### Expected behavior

In [the recent tests](https://github.com/optuna/optuna/pull/5492/files), it seems numpy=1.26.4 was used, but it is apparently not the expected behavior as we should use the oldest numpy version.

The detail of the current issue is as follows:
1. the latest `matplotlib` is installed in the CI
2. the latest `matplotlib` conflicts with NumPy v1.20.3
3. hence `pip install numpy==1.20.3` is skipped
4. Meanwhile, `numpy` is necessary, so `numpy` will be installed anyways without the version constraint
5.  Since `numpy` installed without a version constraint, the latest `numpy`, i.e., v2.0.0, is installed
6. the latest `numpy` causes the failure

In principle, we need to have a version constraint for `matplotlib` so that `numpy==1.20.3` can be installed.
On the other hand, we are not sure if a version constraint to `matplotlib` is sufficient for an expected behavior.
We need to investigate necessary version constraints to circumvent this problem.

### Environment

The versioning follows [this workflow file](https://github.com/optuna/optuna/blob/master/.github/workflows/tests-with-minimum-versions.yml).

```shell
The error message is here:


ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts. matplotlib 3.9.0 requires numpy>=1.23, but you have numpy 1.20.3 which is incompatible.
```

### Steps to reproduce

This is an error in CI.
The workflow file is here.
https://github.com/optuna/optuna/blob/master/.github/workflows/tests-with-minimum-versions.yml

_No response_

porink0424:
Fixes #5499 

## Motivation

This PR fixes the issue #5499 where the version of numpy in the `tests-with-minimum-versions.yml` workflow was not as intended.

## Description of the changes

The packages that conflicted with numpy were `matplotlib`, `pandas`, and `scipy`. By explicitly specifying the same versions of these packages as used in the following workflows, the specified version of numpy is now being correctly used:

- https://github.com/optuna/optuna/actions/runs/8656988956/job/23738403280
- https://github.com/optuna/optuna/actions/runs/8656988956/job/23738403643
- https://github.com/optuna/optuna/actions/runs/8656988956/job/23738403940

Note that python v3.9 has not been using  numpy `1.20.3` in this workflow for some time and instead has been using numpy `1.26.4`. Considering the purpose of this workflow, it might be reasonable to remove the tests for python v3.9, but as I think it is beyond the scope of this PR, it will be left as is for now.

### DIFF
--- a/.github/workflows/tests-with-minimum-versions.yml
+++ b/.github/workflows/tests-with-minimum-versions.yml
@@ -70,14 +70,14 @@ jobs:
       run: |
         # Install dependencies with minimum versions.
         pip uninstall -y alembic cmaes packaging sqlalchemy plotly scikit-learn
-        pip install alembic==1.5.0 cmaes==0.10.0 packaging==20.0 sqlalchemy==1.3.0 numpy==1.20.3 tqdm==4.27.0 colorlog==0.3 PyYAML==5.1
+        pip install alembic==1.5.0 cmaes==0.10.0 packaging==20.0 sqlalchemy==1.3.0 tqdm==4.27.0 colorlog==0.3 PyYAML==5.1
         pip uninstall -y matplotlib pandas scipy
         if [ "${{ matrix.python-version }}" = "3.7" ]; then
-          pip install matplotlib==3.5.3 pandas==1.3.5 scipy==1.7.3
+          pip install matplotlib==3.5.3 pandas==1.3.5 scipy==1.7.3 numpy==1.20.3
         elif [ "${{ matrix.python-version }}" = "3.8" ]; then
-          pip install matplotlib==3.7.5 pandas==2.0.3 scipy==1.10.1
+          pip install matplotlib==3.7.5 pandas==2.0.3 scipy==1.10.1 numpy==1.20.3
         elif [ "${{ matrix.python-version }}" = "3.9" ]; then
-          pip install matplotlib==3.8.3 pandas==2.2.1 scipy==1.12.0
+          pip install matplotlib==3.8.3 pandas==2.2.1 scipy==1.12.0 numpy==1.26.4
         fi
         pip install plotly==5.0.0 scikit-learn==0.24.2  # optional extras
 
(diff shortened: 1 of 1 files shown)


nabenabe0928:
Could you add this issue to the description?
- https://github.com/optuna/optuna/issues/5499

HideakiImamura (approved):
LGTM.

nabenabe0928 (commented):
Thank you for the PR!
At least, I confirmed that  the specified NumPy version is installed correctly.
Could you tell me how you found the versions specified here?

porink0424:
@nabenabe0928 
Based on the command in `.github/workflows/tests-with-minimum-versions.yml`, I built the environment and ran `pip install numpy==1.20.3`. This resulted in the following error, revealing that `matplotlib`, `pandas`, and `scipy` were conflictling with numpy:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
matplotlib 3.9.0 requires numpy>=1.23, but you have numpy 1.20.3 which is incompatible.
pandas 2.2.2 requires numpy>=1.22.4; python_version < "3.11", but you have numpy 1.20.3 which is incompatible.
scipy 1.13.1 requires numpy<2.3,>=1.22.4, but you have numpy 1.20.3 which is incompatible.
```

To determine the versions needed to resolve the conflict, I referred to the versions previously install in CI as I described in the description of this PR.

porink0424:
Additionally, due to the very recent release of `pillow` 10.4.0, which caused errors:
https://github.com/optuna/optuna/actions/runs/9771534901/job/26974420231
I have added version constraints for `pillow` as well.

nabenabe0928 (approved):
Thank you for the change, LGTM!

PR #5529, opened 2024-06-27
Match the argument order of `download_artifact` to that of `upload_artifact`

nabenabe0928:
## Motivation

I united the argument order of `download_artifact` to that of `upload_artifact` to avoid confusion.

### DIFF
--- a/optuna/artifacts/_download.py
+++ b/optuna/artifacts/_download.py
@@ -8,14 +8,14 @@ from optuna.artifacts._protocol import ArtifactStore
 
 
 @experimental_func("4.0.0")
-def download_artifact(artifact_store: ArtifactStore, artifact_id: str, file_path: str) -> None:
+def download_artifact(file_path: str, artifact_store: ArtifactStore, artifact_id: str) -> None:
     """Download an artifact from the artifact store.
 
     Args:
-        artifact_store:
-            An artifact store.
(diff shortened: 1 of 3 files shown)


y0z:
@c-bata, @porink0424 Could you review this PR?

toshihikoyanase:
Could you show us the correspondence of the arguments between `download_artifact` and `upload_artifact `, please?

nabenabe0928:
@toshihikoyanase 
The current API for each function looks like this:

```python
def upload_artifact(study_or_trial, file_path, artifact_store, ...) -> str:
    ...

def download_artifact(artifact_store, artifact_id, file_path) -> None:
    ...
```

I updated `download_artifact` to:

```python
def download_artifact(file_path, artifact_store, artifact_id) -> None:
    ...
```

so that `file_path` comes before `artifact_store` for both functions.
Another possibility is to make them:

```python
def upload_artifact(artifact_store, file_path, study_or_trial,  ...) -> str:
    ...

def download_artifact(artifact_store, file_path, artifact_id) -> None:
    ...
```

Although this change makes the order clearer, it will be a breaking change.
I avoided the breaking change in this PR, but what do you think, @c-bata ?

porink0424 (approved):
I think this change is enough, and further breaking change is a bit much.

PR #5530, opened 2024-06-27
Add a note saying that artifact store information must be saved individually

nabenabe0928:
## Motivation

As artifact store information for each artifact store is not provided, I added a note to warn this. 

## Description of the changes

- Add a note in the doc-string

### DIFF
--- a/optuna/artifacts/_download.py
+++ b/optuna/artifacts/_download.py
@@ -18,6 +18,11 @@ def download_artifact(artifact_store: ArtifactStore, artifact_id: str, file_path
             The identifier of the artifact to download.
         file_path:
             A path to save the downloaded artifact.
+    
+    .. note:
+        Optuna does not provide any API to save artifact store information linked to each artifact,
+        so we kindly ask users to save the artifact store information individually.
+
     """
     if os.path.exists(file_path):
         raise FileExistsError(f"File already exists: {file_path}")
--- a/optuna/artifacts/_download.py
+++ b/optuna/artifacts/_download.py
@@ -19,7 +19,7 @@ def download_artifact(artifact_store: ArtifactStore, artifact_id: str, file_path
         file_path:
             A path to save the downloaded artifact.
 
-    .. note:
+    .. note::
         Optuna does not provide any API to save artifact store information linked to each artifact,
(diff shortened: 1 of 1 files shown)


y0z:
@c-bata, @porink0424 Could you review this PR?

c-bata (changes requested):
@nabenabe0928 The change seems not to be reflected on the documentation. Could you check the following page?
https://optuna--5530.org.readthedocs.build/en/5530/reference/artifacts.html#optuna.artifacts.download_artifact

optuna/artifacts/_download.py
```diff
@@ -18,6 +18,11 @@ def download_artifact(artifact_store: ArtifactStore, artifact_id: str, file_path
             The identifier of the artifact to download.
         file_path:
             A path to save the downloaded artifact.
+
+    .. note:
```
nabenabe0928, line 22:
```suggestion
    .. note::
```

nabenabe0928 (commented):
I found an error based on the notice by c-bata.

nabenabe0928:
@c-bata 
The problem has been fixed:)
image

optuna/artifacts/_download.py
```diff
@@ -18,6 +18,11 @@ def download_artifact(artifact_store: ArtifactStore, artifact_id: str, file_path
             The identifier of the artifact to download.
         file_path:
             A path to save the downloaded artifact.
+
+    .. note::
+        Optuna does not provide any API to save artifact store information linked to each artifact,
```
porink0424, line 23: The phrase `artifact store information` might be a bit ambiguous. How about providing a specific example to clarify?

c-bata (commented):
Thank you for the update. I should have pointed this out at the beginning, but I don't quite understand the motivation for this change. Generally, information about the ArtifactStore is specified in the source code and I don't see the necessity to store it separately. There are similar issues with the way Optuna Storage is specified, but I think it's obvious, so that it's not specified in the documentation.

What do you think?

nabenabe0928:
@c-bata 
You are right, the artifact store information is obvious for users, so I will close this PR.

PR #5498, opened 2024-06-17
Enhance the error message for integration installation

nabenabe0928:
## Motivation

This PR is a followup of this PR:
- https://github.com/optuna/optuna-integration/pull/130

## Description of the changes

Enhance the error message for `from optuna.integration import X` and `from optuna.integration.Y import X`.

> [!NOTE]
> We do not have to use `try_import` in each integration module file because `__init__.py` does the lazy import for each of them.
> However, `ImportError` in each integration module file is necessary as we would like to raise an error when `optuna.integration.X` path is specified explicitly.

Issue #5197, opened 2024-01-23
BoTorch Dependency Not Installed with Optuna via pip


resnant:
### Expected behavior

When installing Optuna using the standard `pip install optuna` command, BoTorch, one of its dependencies, should be automatically installed along with Optuna. Currently, this is not the case, and users need to install BoTorch separately. 
The expected behavior is that Optuna's installation process includes the installation of all necessary dependencies, including BoTorch.

### Environment

- Optuna version:3.5.0
- Python version:3.11.5
- OS:Linux-4.19.128-microsoft-standard-x86_64-with-glibc2.31 (Docker `continuumio/miniconda3:latest`)

### Error messages, stack traces, or logs

```shell
(base) root@aaf69946013a:/# python
Python 3.11.5 (main, Sep 11 2023, 13:54:46) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from optuna_dashboard.preferential.samplers.gp import PreferentialGPSampler
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/conda/lib/python3.11/site-packages/optuna_dashboard/preferential/samplers/gp.py", line 9, in <module>
    import botorch.acquisition.analytic
ModuleNotFoundError: No module named 'botorch'
```

### Steps to reproduce

1. `docker run --rm -it continuumio/miniconda3 bash`
2. `pip install optuna optuna-dashboard`
3. 
```python
from optuna_dashboard.preferential.samplers.gp import PreferentialGPSampler
```

contramundum53:
@resnant 
Sorry, `botorch` is in the _optional_ dependency of `optuna`. Please install all optional dependencies with 
```
pip install optuna[optional]
```
or install dependency separately as it requires. (`BoTorchSampler` only requires `botorch`.)

We cannot make `botorch>=0.8.1` a mandatory dependency because `botorch` has large and strict child dependencies and many `optuna` users would fail to install `optuna` at all otherwise.

contramundum53:
Maybe we should make the error message more readable, though.

nabenabe0928:
Additional dependencies for `optuna-integration` can now installed via pip and I modified the error message so that users can easily notice.
With that, let me close this issue.

PR #5563, opened 2024-07-10
Enhance the error message for integration import error

nabenabe0928:
## Motivation

Although the `ModuleNotFoundError` for integration has been improved in
- https://github.com/optuna/optuna/pull/5498

the import error message has not been improved as the error message above happens only if `optuna-integration` has not been installed yet.

So we would like to enhance the error message even after `optuna-integration` is installed.
For example, 

```python
from optuna.integration import BoTorchSampler

BoTorchSampler()
```

will raise

```
ImportError: Tried to import 'botorch' but failed. Please make sure that the package is installed correctly to use this feature. Actual error: No module named 'botorch'.
```

However, dependency for `optuna_integration.XXX` is not necessarily one package, so it is nicer to raise like:

```
ImportError:
Tried to import optuna-integration for 'botorch' but failed.
Please install the dependencies via:
        $ pip install --upgrade optuna-integration[botorch]
to use this feature. Actual error: No module named 'botorch'.
```

## Description of the changes

- Enhance the error message when `ImportError` is propagated from `optuna_integration`.

### DIFF
--- a/optuna/_imports.py
+++ b/optuna/_imports.py
@@ -15,21 +15,6 @@ _INTEGRATION_IMPORT_ERROR_TEMPLATE = (
 )
 
 
-def _get_exception_source_module(traceback: TracebackType | None) -> str | None:
-    if traceback is None:
-        return None
-
-    if not hasattr(traceback, "tb_frame"):
-        return None
-
-    traceback_frame = traceback.tb_frame
-    if not hasattr(traceback_frame, "f_locals"):
-        return None
-
-    traceback_frame_local_name = traceback.tb_frame.f_locals
-    return traceback_frame_local_name.get("__name__")
-
-
 class _DeferredImportExceptionContextManager:
     """Context manager to defer exceptions from imports.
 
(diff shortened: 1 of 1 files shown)


nabenabe0928:
Related issue:
- https://github.com/optuna/optuna/issues/5197

eukaryo:
@HideakiImamura @y0z Could you review this PR?

nabenabe0928:
This PR should be merged after
- https://github.com/optuna/optuna/pull/5565
considering the priority.

HideakiImamura:
While I believe the changes in this PR would make the error messages more understandable, couldn't this be covered in the documentation? The complexity that would be added to the codebase by these changes seems relatively large, so I want to be cautious about introducing this PR.

nabenabe0928:
> While I believe the changes in this PR would make the error messages more understandable, couldn't this be covered in the documentation? The complexity that would be added to the codebase by these changes seems relatively large, so I want to be cautious about introducing this PR.

That sounds actually fair.
I will change the error message to include the readthedocs for optuna-integration instead.

The latest mentions how we install the dependencies of a submodule.

https://optuna-integration.readthedocs.io/en/latest/installation.html

nabenabe0928:
@HideakiImamura 
I simply included the integration installation URL so that the complexity will not increase.

optuna/_imports.py
```diff
@@ -56,16 +58,24 @@ def __exit__(
...
+                )
                 message = (
-                    "Tried to import '{}' but failed. Please make sure that the package is "
-                    "installed correctly to use this feature. Actual error: {}."
-                ).format(exc_value.name, exc_value)
+                    f"\nTried to import '{exc_value.name}' but failed."
+                    "\nPlease make sure that the package is installed correctly."
+                    "\nIf the error relates to optuna-integration, please refer to:"
+                    f"\n\t{integration_installation_doc_url}"
+                    f"\nActual error: {exc_value}."
```
HideakiImamura, line 70: I think this is confusing for users who don't use integration modules. I would like to just close this PR.

PR #5576, opened 2024-07-12
Speed up `mypy` by stopping the checks in `torch`

eukaryo:
## Motivation

I want to speed up `$ mypy --no-incremental . ` by stopping the checks inside `torch`.
I want `checks.yml` in GitHub Workflow not to ignore `torch`, because it can utilize mypy_cache and therefore ignoring brings no benefit.

## Description of the changes

Update `setup.cfg`. This change will make `mypy` stop checking `torch`, resulting in about half the time required (in my local environment).

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -44,6 +44,6 @@ jobs:
     - name: isort
       run: isort . --check --diff
     - name: mypy
-      run: mypy --follow-imports=normal .
+      run: awk '/\[mypy-torch\.\*\]/{exit} {print}' setup.cfg | mypy --config-file=/dev/stdin .
     - name: blackdoc
       run: blackdoc . --check --diff
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -44,6 +44,6 @@ jobs:
     - name: isort
       run: isort . --check --diff
     - name: mypy
-      run: mypy .
+      run: mypy --follow-imports=normal .
     - name: blackdoc
       run: blackdoc . --check --diff
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -44,6 +44,6 @@ jobs:
     - name: isort
       run: isort . --check --diff
     - name: mypy
-      run: awk '/\[mypy-torch\.\*\]/{exit} {print}' setup.cfg | mypy --config-file=/dev/stdin .
+      run: sed '/# IGNORE MYPY TORCH START/,/# IGNORE MYPY TORCH END/d' setup.cfg | mypy --config-file=/dev/stdin .
     - name: blackdoc
       run: blackdoc . --check --diff
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,6 +25,8 @@ no_implicit_reexport = True
 ignore_missing_imports = True
 exclude = .venv|venv|build|docs|tutorial|optuna/storages/_rdb/alembic
 
+# IGNORE MYPY TORCH START (This line is used in checks.yml)
 [mypy-torch.*]
 # Any `torch`-related modules will be replaced with `typing.Any` at type checking.
 follow_imports = skip
+# IGNORE MYPY TORCH END (This line is used in checks.yml)
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,4 +26,5 @@ ignore_missing_imports = True
 exclude = .venv|venv|build|docs|tutorial|optuna/storages/_rdb/alembic
 
 [mypy-torch.*]
+# Any `torch`-related modules will be replaced with `typing.Any` at type checking.
 follow_imports = skip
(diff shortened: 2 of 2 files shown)


eukaryo:
```
[mypy-torch.*]
follow_imports = skip
```
We found that with this option, mypy passes without even installing `torch`.

y0z:

nabenabe0928:
Master branch at 18 July, 2024, 10:00

```shell
$ rm -r .mypy_cache/; time mypy optuna tests

real    0m32,012s
user    0m30,992s
sys     0m0,724s
```

This PR

```shell
$ rm -r .mypy_cache/; time mypy optuna tests

real    0m20,068s
user    0m19,283s
sys     0m0,418s
```

setup.cfg
```diff
@@ -24,3 +24,6 @@ no_implicit_reexport = True
 
 ignore_missing_imports = True
 exclude = .venv|venv|build|docs|tutorial|optuna/storages/_rdb/alembic
+
+[mypy-torch.*]
+follow_imports = skip
```
nabenabe0928, line 29: According to [the MyPy official documentation](https://mypy.readthedocs.io/en/stable/running_mypy.html#following-imports), this line has an effect to simply replace any `torch`-related modules with `Any` in type checking.
Probably, it should be mentioned.

```suggestion
# Any `torch`-related modules will be replaced with `typing.Any` at type checking.
follow_imports = skip
```

nabenabe0928 (commented):
Thank you for the PR, I left a comment!

.github/workflows/checks.yml
```diff
@@ -44,6 +44,6 @@ jobs:
     - name: isort
       run: isort . --check --diff
     - name: mypy
-      run: mypy .
+      run: mypy --follow-imports=normal .
```
nabenabe0928, line 47: In my environment, the following does not import `torch` as intended:

```shell
rm -r .mypy_cache/; time mypy optuna tests --follow-imports=normal
```

nabenabe0928: @eukaryo Could you confirm it?

setup.cfg
eukaryo: Agreed

.github/workflows/checks.yml
eukaryo: I realized that I had misunderstood about precedence order of configurations.
`--follow-imports=normal` in command-line option cannot override `[mypy-torch.*]` section in the config file.
(cf. https://mypy.readthedocs.io/en/stable/config_file.html )

eukaryo: I have revised the code.
`rm -r .mypy_cache/; awk '/\[mypy-torch\.\*\]/{exit} {print}' setup.cfg | (time mypy --config-file=/dev/stdin .)`
This command will work as intended. Could you confirm it?

.github/workflows/checks.yml
```diff
@@ -44,6 +44,6 @@ jobs:
     - name: isort
       run: isort . --check --diff
     - name: mypy
-      run: mypy .
+      run: awk '/\[mypy-torch\.\*\]/{exit} {print}' setup.cfg | mypy --config-file=/dev/stdin .
```
nabenabe0928, line 47: I think this is a bit buggy because people may not notice that `setup.cfg` is cut off at the line after `[mypy-torch.*]`.

With this command, if we add any additional lines after `[mypy-torch.*]`, they will be ignored at the test time.

I would recommend adding inline markers to the start and the end of the lines that should be ignored.
Then it is not gonna be a complicated command and easier to circumvent the issue.

For example, if we change `setup.cfg` as follows:

```cfg
# IGNORE MYPY TORCH START (This line is used in checks.yml)
[mypy-torch.*]
# Any `torch`-related modules will be replaced with `typing.Any` at type checking.
follow_imports = skip
# IGNORE MYPY TORCH END (This line is used in checks.yml)
```

the line would look like this:

```shell
$ sed '/# IGNORE MYPY TORCH START/,/# IGNORE MYPY TORCH END/d' setup.cfg
```

By doing so, the shell command itself is clearer and more understandable by a glance.

nabenabe0928 (commented):
Thank you for the fix!
I proposed another solution to the issue we faced:)

.github/workflows/checks.yml
eukaryo: Agreed, and I confirmed your suggestion also works as intended. I changed the code.

nabenabe0928: Thank you for the update!
Could you also change formats.sh?

not522:
How about making `mypy-ci.ini` and using `mypy . --config-file mypy-ci.ini` in the CI? It seems more straightforward than a single config file.

.github/workflows/checks.yml
eukaryo: I think it is not needed to change formats.sh, because the behavior of `$ mypy .` and `$ bash formats.sh` should match in the user's local environment.

nabenabe0928: It's probably right that this change might be confusing.
If necessary, we can come back to here when we modify `GPSampler`-related modules.

nabenabe0928:
@eukaryo Could you mention me once you push another commit addressing @not522 's comment?

PR #5579, opened 2024-07-13
Simplify annotation in `trial`

RektPunk:
## Motivation
Related to https://github.com/optuna/optuna/issues/4508

## Description of the changes
I removed the calls to the typing module except for `Any`, `cast` and `overload`.

### DIFF
--- a/optuna/trial/_frozen.py
+++ b/optuna/trial/_frozen.py
@@ -6,6 +6,7 @@ import datetime
 import math
 from typing import Any
 from typing import cast
+from typing import Dict
 from typing import overload
 import warnings
 
@@ -445,7 +446,7 @@ class FrozenTrial(BaseTrial):
 
     @system_attrs.setter
     def system_attrs(self, value: Mapping[str, JSONSerializable]) -> None:
-        self._system_attrs = cast(dict, value)
+        self._system_attrs = cast(Dict[str, Any], value)
 
     @property
     def last_step(self) -> int | None:
--- a/optuna/trial/_frozen.py
+++ b/optuna/trial/_frozen.py
@@ -6,7 +6,6 @@ import datetime
 import math
 from typing import Any
 from typing import cast
-from typing import Dict
 from typing import overload
 import warnings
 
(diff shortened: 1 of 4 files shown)


RektPunk:
I encountered the following error: TypeError: 'type' object is not subscriptable  when modify the following part using cast and annotations:
https://github.com/optuna/optuna/blob/9b5c8635f3b1ef23c80a12cdafda4055490b72db/optuna/trial/_frozen.py#L447-L449
to
```python
self._system_attrs = cast(dict[str, Any], value) 
```
I reverted to the original state because I wasn't sure which method is preferred.

y0z (approved):
Thank you for your contribution.
LGTM

optuna/trial/_frozen.py
```diff
@@ -409,47 +408,47 @@ def _set_values(self, v: Optional[Sequence[float]]) -> None:
...
 
     @property
-    def system_attrs(self) -> Dict[str, Any]:
+    def system_attrs(self) -> dict[str, Any]:
         return self._system_attrs
 
     @system_attrs.setter
     def system_attrs(self, value: Mapping[str, JSONSerializable]) -> None:
-        self._system_attrs = cast(Dict[str, Any], value)
+        self._system_attrs = cast(dict, value)
```
nabenabe0928, line 448: This line slightly changes the type checking behavior.

The previous code was stricter than the change made by you.
For example, the last one did not allow `Dict[int, Any]`, because `cast` ensures that `value` is `Dict[str, Any]`.
However, the current code allows `Dict[Any, Any]`.

```suggestion
        self._system_attrs = cast(Dict[str, Any], value)
```

nabenabe0928 (changes requested):
Thank you for the PR, I left a request!

optuna/trial/_frozen.py
RektPunk: Thanks, as you said, it is fixed in 9300fb5

nabenabe0928 (approved):
Thank you for the prompt response, LGTM!

PR #5580, opened 2024-07-14
Fix mypy warnings in `matplotlib/_contour.py` with Python3.11

eukaryo:
## Motivation

I want to improve type annotations. Currently, CI has confirmed that no issues found by mypy with Python 3.8. I checked with Python3.11, and some errors were found. I want to fix them without causing any issues with 3.8. I believe this activity improves quality of code. 

## Description of the changes

Fixed `matplotlib/_contour.py`. I confirmed that `_generate_contour_subplot` is used only in `matplotlib/_contour.py` by `$ git grep -l "_generate_contour_subplot"`, and the code-fix does not change any behaviour.

### DIFF
--- a/optuna/visualization/matplotlib/_contour.py
+++ b/optuna/visualization/matplotlib/_contour.py
@@ -282,10 +282,12 @@ def _calculate_griddata(
     )
 
 
-def _generate_contour_subplot(info: _SubContourInfo, ax: "Axes", cmap: "Colormap") -> "ContourSet":
+def _generate_contour_subplot(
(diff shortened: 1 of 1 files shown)


y0z (approved):
LGTM

nabenabe0928 (approved):
I confirmed the change with the following code:

```python
import optuna

study = optuna.create_study()
study.optimize(lambda t: t.suggest_float("x", 0, 1), n_trials=30)
optuna.visualization.matplotlib.plot_contour(study)
plt.show()
```

Issue #5409, opened 2024-04-22
Enhance `plot_parallel_coordinate()` by eliminating redundant for-loops


c-bata:
### Motivation

In parallel coordinate plot, `_is_log_scale()`, `_is_categorical()`, and `_is_numerical()` functions are invoked.

https://github.com/optuna/optuna/blob/fcffbeab86859c32c3c646947f4fd764860c922b/optuna/visualization/_parallel_coordinate.py#L213-L237

Each one of these functions iterates over the trials, which is clearly redundant.

https://github.com/optuna/optuna/blob/fcffbeab86859c32c3c646947f4fd764860c922b/optuna/visualization/_utils.py#L72-L98

### Suggestion

Making `plot_parallel_coorinate()` faster by removing wasted for-loop.

kAIto47802:
I will take over [the PR](https://github.com/optuna/optuna/pull/5547) (I will checkout another branch from the branch created by @nzw301) and would like to solve this issue!

kAIto47802:
This issue was resolved by

- #5587.

However, the additional issue

- #5571

related to this is still unresolved. (currently, I'm not working on this.)

PR #5581, opened 2024-07-14
Remove redundant loop in `_get_rank_subplot_info`

RektPunk:
## Motivation
Making faster by removing wasted for-loop.

## Description of the changes
Combine the loops into one loop for trial.

### DIFF
--- a/optuna/visualization/_rank.py
+++ b/optuna/visualization/_rank.py
@@ -210,24 +210,24 @@ def _get_rank_subplot_info(
     xaxis = _get_axis_info(trials, x_param)
     yaxis = _get_axis_info(trials, y_param)
 
-    infeasible_trial_ids = []
-    for i in range(len(trials)):
-        constraints = trials[i].system_attrs.get(_CONSTRAINTS_KEY)
+    xs: list[Any] = []
+    ys: list[Any] = []
+    zs: list[np.ndarray] = []
+    filtered_trials: list[FrozenTrial] = []
+    filtered_colors: list[np.ndarray] = []
+
+    for idx, trial in enumerate(trials):
+        constraints = trial.system_attrs.get(_CONSTRAINTS_KEY)
         if constraints is not None and any([x > 0.0 for x in constraints]):
-            infeasible_trial_ids.append(i)
+            colors[idx] = (204, 204, 204)  # equal to "#CCCCCC"
 
-    colors[infeasible_trial_ids] = (204, 204, 204)  # equal to "#CCCCCC"
+        if x_param in trial.params and y_param in trial.params:
+            xs.append(trial.params[x_param])
+            ys.append(trial.params[y_param])
+            zs.append(target_values[idx])
+            filtered_trials.append(trial)
+            filtered_colors.append(colors[idx])
 
-    filtered_ids = [
-        i
-        for i in range(len(trials))
-        if x_param in trials[i].params and y_param in trials[i].params
-    ]
-    filtered_trials = [trials[i] for i in filtered_ids]
-    xs = [trial.params[x_param] for trial in filtered_trials]
-    ys = [trial.params[y_param] for trial in filtered_trials]
-    zs = target_values[filtered_ids]
-    colors = colors[filtered_ids]
     return _RankSubplotInfo(
         xaxis=xaxis,
         yaxis=yaxis,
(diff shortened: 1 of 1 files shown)


c-bata:
Let me update the PR description since this PR is not related to #5409.

c-bata:
@RektPunk Since the motivation of this PR is to improve speed, could you prepare a microbenchmark script and share the results with us? Please note that the microbenchmark script should not be pushed to the Optuna repository, but instead, please share it through the comments of this PR or via Gist.

RektPunk:
You're right @c-bata, it's not a significant difference though, but you can see the difference in the script below.
<details>

```python

import numpy as np
from datetime import datetime
import optuna

def sphere(trial: optuna.Trial) -> float:
    X = np.asarray([trial.suggest_float(f"x{i}", -5, 5) for i in range(10)])
    return np.sum(X ** 2)


study = optuna.create_study()
study.optimize(sphere, n_trials=1000)

trials = study.get_trials(deepcopy=False, states=(1,))
params = {"x0", "x1"}
target_values = np.array([trial.value for trial in trials])
colors = np.array([(204, 204, 204) for _ in range(1000)])

before = datetime.now()
x_param = "x1"
y_param = "x2"

infeasible_trial_ids = []
for i in range(len(trials)):
    constraints = trials[i].system_attrs.get("constraints")
    if constraints is not None and any([x > 0.0 for x in constraints]):
        infeasible_trial_ids.append(i)

colors[infeasible_trial_ids] = (204, 204, 204)  # equal to "#CCCCCC"

filtered_ids = [
    i
    for i in range(len(trials))
    if x_param in trials[i].params and y_param in trials[i].params
]
filtered_trials = [trials[i] for i in filtered_ids]
xs = [trial.params[x_param] for trial in filtered_trials]
ys = [trial.params[y_param] for trial in filtered_trials]
zs = target_values[filtered_ids]
colors = colors[filtered_ids]
after = datetime.now()
print((after - before).total_seconds())

before = datetime.now()
x_param = "x1"
y_param = "x2"
infeasible_trial_ids = []
filtered_ids = []
for idx, trial in enumerate(trials):
    constraints = trial.system_attrs.get("constraints")
    if constraints is not None and any([x > 0.0 for x in constraints]):
        infeasible_trial_ids.append(idx)

    if x_param in trial.params and y_param in trial.params:
        filtered_ids.append(idx)

filtered_trials = [trials[i] for i in filtered_ids]
xs = [trial.params[x_param] for trial in filtered_trials]
ys = [trial.params[y_param] for trial in filtered_trials]
zs = target_values[filtered_ids]
colors[infeasible_trial_ids] = (204, 204, 204)
colors = colors[filtered_ids]
after = datetime.now()
print((after - before).total_seconds())
```
result >
```cmd
0.002047
0.001205
```
</details>

eukaryo (approved):
LGTM.

c-bata (approved):
Thank you for sharing the microbenchmark results! I modified your script to compare the speed of `_get_rank_info()` functions.

```python
import numpy as np
from datetime import datetime
import optuna
import time
from optuna.visualization._rank import _get_rank_info

def sphere(trial: optuna.Trial) -> float:
    X = np.asarray([trial.suggest_float(f"x{i}", -5, 5) for i in range(10)])
    return np.sum(X ** 2)

study = optuna.create_study(sampler=optuna.samplers.RandomSampler(seed=1))
study.optimize(sphere, n_trials=10000)

elapsed = []
for i in range(10):
    start = time.time()
    _get_rank_info(study, None, None, "Objective Value")
    elapsed.append(time.time() - start)

print(f"Elapsed: {np.mean(elapsed):.4f} seconds (+/- {np.std(elapsed):.4f})")
```

`master`

- 1000 trials: 0.2344 seconds (+/- 0.0037)
- 10000 trials: 3.0675 seconds (+/- 0.0230)

`RektPunk:refactor-rank-subplot-info`

- 1000 trials: 0.2258 seconds (+/- 0.0016)
- 10000 trials: 3.0951 seconds (+/- 0.0426)

From the above results, I cannot see the performance improvements on this PR since it takes longer than `master` branch when I increased trials to 10000. According to the output of cProfile, we need to make `_get_axis_info` function (including`_is_logscale` and `is_numerical` functions) faster as reported at #5409.

https://gist.github.com/c-bata/6e6d829aa2062feda761a1d66fe308a5#file-cprofile_get_rank_info-py

But anyway, I think this PR is acceptable as a refactoring. It's slightly reduces the number of lines of code and code looks easier to read than using list comprehensions.

LGTM.

PR #5583, opened 2024-07-16
Simplify annotations in `_pertial_fixed.py`

eukaryo:
## Motivation

Related to Issue https://github.com/optuna/optuna/issues/4508.

I have considering about `PartialFixedSampler`, and wanted to resolve the issue in the module as a preliminary action.

## Description of the changes

I removed `Dict` and `Optional` to typing from `_pertial_fixed.py`, and refactored the code.

### DIFF
--- a/optuna/samplers/_partial_fixed.py
+++ b/optuna/samplers/_partial_fixed.py
@@ -1,8 +1,6 @@
 from __future__ import annotations
 
 from typing import Any
-from typing import Dict
-from typing import Optional
 from typing import Sequence
 from typing import TYPE_CHECKING
 import warnings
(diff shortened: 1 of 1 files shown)


optuna/samplers/_partial_fixed.py
```diff
@@ -1,8 +1,6 @@
 from __future__ import annotations
 
 from typing import Any
-from typing import Dict
-from typing import Optional
 from typing import Sequence
```
not522, line 4: `typing.Sequence` is deprecated. Could you replace it with `collections.abc.Sequence`?

```suggestion
from collections.abc import Sequence
```

not522 (approved):
LGTM

y0z (approved):
LGTM

PR #5587, opened 2024-07-16
Refactor `_is_categorical()` in `optuna/optuna/visualization`

kAIto47802:
## Motivation

This PR is related to:
- #5409 

## Description of the changes

For the function `_is_categorical()`, the problem of iterating over trials multiple times, which is indicated in #5409, is already resolved. This is because the function `_is_categorical()` was originally designed to return as soon as it encountered a categorical trial, and in Optuna, the categorical distribution should be consistent across all trials.

Also, the function `_get_parallel_coordinate_info()` originally contains a section that iterated over trials:
https://github.com/optuna/optuna/blob/fcffbeab86859c32c3c646947f4fd764860c922b/optuna/visualization/_parallel_coordinate.py#L205-L211
Therefore, by combining these parts, we can address the visually redundant iteration. Additionally, since the function `_is_categorical()` is used only in the function `_get_parallel_coordinate_info()`, there is no need to separate this into `_utils.py`.

### DIFF
--- a/optuna/visualization/_utils.py
+++ b/optuna/visualization/_utils.py
@@ -86,7 +86,9 @@ def _is_numerical(trials: list[FrozenTrial], param: str) -> bool:
         if isinstance(dist, (IntDistribution, FloatDistribution)):
             return True
         elif isinstance(dist, CategoricalDistribution):
-            return all(isinstance(v, (int, float)) and not isinstance(v, bool) for v in dist.choices)
+            return all(
+                isinstance(v, (int, float)) and not isinstance(v, bool) for v in dist.choices
+            )
         else:
             assert False, "Should not reach."
     return True
--- a/optuna/visualization/_utils.py
+++ b/optuna/visualization/_utils.py
@@ -81,13 +81,6 @@ def _is_log_scale(trials: list[FrozenTrial], param: str) -> bool:
     return False
 
 
-def _is_categorical(trials: list[FrozenTrial], param: str) -> bool:
-    for trial in trials:
-        if param in trial.params:
-            return isinstance(trial.distributions[param], CategoricalDistribution)
-    return False
-
-
 def _is_numerical(trials: list[FrozenTrial], param: str) -> bool:
     for trial in trials:
         if param in trial.params:
--- a/optuna/visualization/_utils.py
+++ b/optuna/visualization/_utils.py
@@ -71,29 +71,25 @@ def _check_plot_args(
 
 def _is_log_scale(trials: list[FrozenTrial], param: str) -> bool:
     for trial in trials:
-        if param in trial.params:
-            dist = trial.distributions[param]
-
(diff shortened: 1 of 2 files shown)


nabenabe0928 (commented):
Thank you for the PR!
I left a comment and will check utils.py later as well!

optuna/visualization/_utils.py
```diff
@@ -73,25 +73,27 @@ def _is_log_scale(trials: list[FrozenTrial], param: str) -> bool:
...
 
-def _is_categorical(trials: list[FrozenTrial], param: str) -> bool:
-    return any(
-        param in t.params and isinstance(t.distributions[param], CategoricalDistribution)
-        for t in trials
-    )
+            if isinstance(dist, (FloatDistribution, IntDistribution)):
+                return dist.log
+            else:
+                return False
```
nabenabe0928, line 80: What about this?
```python
if param not in trial.params:
    continue

dist = trial.distributions[param]
return isinstance(dist, (FloatDistribution, IntDistribution)) and dist.log
```

nabenabe0928 (commented):
I left some additional comments! Please take another look:)

> [!IMPORTANT]
> I noticed that the behavior of `_is_numerical` has changed with this PR.
> More specifically, if `study` observes only a part of `choices` in a categorical parameter and a categorical parameter includes (a) numerical parameter(s), the result may change.
> For example, if `choices` is `[1, 2, "3"]` and `study` observes only `1` or `2`, `_is_numerical` returns `True` for the master branch, but this PR returns `False`.

However, as I think the behavior of this PR is more appropriate, I would like to proceed with this behavior.

nabenabe0928:
@kAIto47802 Could you refactor the code based on my comments?

kAIto47802:
Thank you for the detailed review. I was just working on the fixes, and I've completed them.

nabenabe0928:
@kAIto47802 I will check this PR asap, but I cannot do it this week probably.

toshihikoyanase:
@nabenabe0928 It appears this is still under review. Could you consider reassigning the reviewers if needed?

nabenabe0928 (approved):

nabenabe0928:

optuna/visualization/_utils.py
```diff
@@ -71,26 +71,26 @@ def _check_plot_args(
...
-            value = trial.params[param]
-            if not isinstance(value, (int, float)) or isinstance(value, bool):
-                return False
+        if param not in trial.params:
+            continue
+        dist = trial.distributions[param]
+        if isinstance(dist, (IntDistribution, FloatDistribution)):
+            return True
+        elif isinstance(dist, CategoricalDistribution):
+            return all(
```
nabenabe0928, line 89:
```suggestion
            # NOTE: Although it is a bit odd to do so, we keep it as is only for visualization.
            return all(
```

y0z (approved):
LGTM

PR #5591, opened 2024-07-20
Speed up WFG by using a fact that hypervolume calculation does not need (second or later) duplicated Pareto solutions

nabenabe0928:
## Motivation

The biggest motivation is to speed up MOTPE.
By this change, the routine of WFG will be quicker in 3D Hypervolume calculation using balanced-binary tree introduced in
- https://github.com/optuna/optuna/issues/5203

## Description of the changes

- Stop `np.unique` operation for `limited_sols`
- Add the explanations why we can make the change above. 

Basically, `limited_sols` is sorted by the first objective, so `is_pareto_front` works as intended except for duplicated solutions after the first found ones will be eliminated from Pareto solutions.
The duplicated elimination does not affect the result of hypervolume because duplicated solutions do not contribute to the hypervolume at all.

### DIFF
--- a/optuna/study/_multi_objective.py
+++ b/optuna/study/_multi_objective.py
@@ -162,7 +162,7 @@ def _is_pareto_front_for_unique_sorted(unique_lexsorted_loss_values: np.ndarray)
 def _is_pareto_front(loss_values: np.ndarray, assume_unique_lexsorted: bool) -> np.ndarray:
     # NOTE(nabenabe): If assume_unique_lexsorted=True, but loss_values is not a unique array,
     # Duplicated Pareto solutions will be filtered out except for the earliest occurrences.
-    # If assume_unique_lexsorted=True and loss_vals[:, 0] is not sorted, then the result will be
+    # If assume_unique_lexsorted=True and loss_values[:, 0] is not sorted, then the result will be
     # incorrect.
     if assume_unique_lexsorted:
         return _is_pareto_front_for_unique_sorted(loss_values)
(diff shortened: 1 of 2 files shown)


nabenabe0928:
<details>
<summary>Benchmarking Code</summary>

```python

import time

import numpy as np

import optuna


def objective3d(trial: optuna.Trial) -> tuple[float, float, float]:
    if (trial.number + 1) % 50 == 0:
        print(trial.number)


if __name__ == "__main__":

    start = time.time()
    n_obj = 3
    sampler = optuna.samplers.TPESampler(seed=42, multivariate=True)
    study = optuna.create_study(sampler=sampler, directions=["minimize"]*n_obj)
    study.optimize(objective3d, n_trials=300)

    print(time.time() - start, study.trials[-1].params)

```

</details>

nabenabe0928:
@eukaryo @y0z 
Could you review this PR?

eukaryo (approved):
LGTM.

optuna/study/_multi_objective.py
```diff
@@ -159,6 +160,10 @@ def _is_pareto_front_for_unique_sorted(unique_lexsorted_loss_values: np.ndarray)
 
 
 def _is_pareto_front(loss_values: np.ndarray, assume_unique_lexsorted: bool) -> np.ndarray:
+    # NOTE(nabenabe): If assume_unique_lexsorted=True, but loss_values is not a unique array,
+    # Duplicated Pareto solutions will be filtered out except for the earliest occurrences.
+    # If assume_unique_lexsorted=True and loss_vals[:, 0] is not sorted, then the result will be
```
y0z, line 165:
```suggestion
    # If assume_unique_lexsorted=True and loss_values[:, 0] is not sorted, then the result will be
```

y0z (approved):
LGTM

PR #5594, opened 2024-07-24
Reorganize names related to `search_space` in `terminator/improvement/evaluator.py`

eukaryo:
## Motivation

I want to avoid name collisions which exist in `Terminator` module.

## Description of the changes

- Renamed colliding name of variables, `gp_search_space`,  to ~`considered_search_space`~`search_space`. The name `gp_search_space` is used in following: https://github.com/optuna/optuna/blob/39f83f7eda0979af27d5e9dd6acc7a08fa1c63ec/optuna/samplers/_gp/sampler.py#L30
- Replaced `from optuna._gp import search_space` with `from optuna._gp import search_space as gp_search_space`.
  - Fixed affected codes.

### DIFF
--- a/optuna/terminator/improvement/evaluator.py
+++ b/optuna/terminator/improvement/evaluator.py
@@ -166,10 +166,8 @@ class RegretBoundEvaluator(BaseImprovementEvaluator):
         # _gp module assumes that optimization direction is maximization
         sign = -1 if study_direction == StudyDirection.MINIMIZE else 1
         values = np.array([t.value for t in complete_trials]) * sign
-        search_space, normalized_params = (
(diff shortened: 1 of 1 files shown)


HideakiImamura:
How about just using `search_space` instead of `considered_search_space`? The `considered_search_space` is not clear name for me.

eukaryo:
Thank you for your comments! I have changed the code to reflect them.
What I really wanted was just to stop using `gp_search_space` as a variable name and use it as a module name.

not522 (approved):
LGTM. It sounds more natural than the previous naming.

HideakiImamura (commented):
LGTM!

PR #5598, opened 2024-07-27
Refactor `_bisect` in `_truncnorm.py`

eukaryo:
## Motivation

I want to refactor code, resolve a TODO-comment, and enhance `TPESampler`.

## Description of the changes

- Changed the stop condition of the binary search in `_bisect` function. Since the target function uses `lru_cache`, there is only a small speedup (about 2% in my local environment).


eukaryo:
Thank you for your comments! Your suggested code seems excellent, but it is your own achievement rather than revision of this PR. So I think the suggestion should be considered in another PR after this one...

eukaryo:
FYI: 
PR https://github.com/optuna/optuna/pull/4105
Related comment about using Newton's method: https://github.com/optuna/optuna/pull/4105/files#r1065455150

optuna/samplers/_tpe/_truncnorm.py
```diff
@@ -151,14 +151,15 @@ def mass_case_central(a: np.ndarray, b: np.ndarray) -> np.ndarray:
 def _bisect(f: Callable[[float], float], a: float, b: float, c: float) -> float:
     if f(a) > c:
         a, b = b, a
-    # TODO(amylase): Justify this constant
-    for _ in range(100):
+    # In the algorithm, it is assumed that all of (a + b), (a * 2), and (b * 2) are finite.
+    while True:
         m = (a + b) / 2
+        if a == m or b == m:
```
eukaryo, line 157: It is possible that `a != m and b == m` holds in a converged state.

optuna/samplers/_tpe/_truncnorm.py
```diff
@@ -151,14 +151,15 @@ def mass_case_central(a: np.ndarray, b: np.ndarray) -> np.ndarray:
 def _bisect(f: Callable[[float], float], a: float, b: float, c: float) -> float:
     if f(a) > c:
         a, b = b, a
-    # TODO(amylase): Justify this constant
-    for _ in range(100):
+    # In the algorithm, it is assumed that all of (a + b), (a * 2), and (b * 2) are finite.
+    while True:
```
not522, line 155: How about keeping the upper bound of the loop count? I am not confident that the condition `a == m or b == m` is satisfied in the convergent state in all environments.

eukaryo: I see. I have revised the code.

not522 (approved):
LGTM

nabenabe0928 (approved):
LGTM!

PR #5626, opened 2024-08-16
Update Trial doc comment

bionicles:
## Motivation
The doc comment indicates users do not care about Trial.__init__, but I am a user and I care about it because I do not use `optimize`. 

Therefore, this doc comment implies users feel a certain way, which isn't necessarily true, and may lead maintainers to feel justified in lack of support for custom training loops.

## Description of the changes
Update the Trial doc comment to reflect usage in custom training loops

### DIFF
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -32,9 +32,14 @@ class Trial(BaseTrial):
     """A trial is a process to evaluate an objective function.
 
     This mutable Trial object provides interfaces to:
-        - suggest parameters
-        - manage trial state
-        - set/get user-defined attributes
+        - suggest_float(name: str, low: float, high: float, *, step: float | None = None)
+        - suggest_int(name: str, low: int, high: int, *, step: int = 1) -> int
+        - suggest_categorical(name: str, choices: Sequence[Choice]) -> Choice
+            - where Choice = None | bool | int | float | str
+        - report objective values
+        - should_prune to suggest pruning this trial or not
+        - set_user_attr to set user-defined trial attributes
+        - user_attrs is a dict of user-defined attributes
 
     Automatic Optimization
     ----------------------
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -33,8 +33,8 @@ class Trial(BaseTrial):
 
     This mutable Trial object provides interfaces to:
         - suggest_float(name: str, low: float, high: float, *, step: float | None = None)
-        - suggest_int(name: str, low: int, high: int, *, step: int = 1) -> int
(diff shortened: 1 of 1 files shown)


nzw0301:
Thank you for sending the PR. I'm unsure why you're not using `study.ask` for such a custom optimisation loop because I've never ever used `Trial.__init__` for a few years of optuna development. Could you use `ask` instead of directly calling `trial.__init__`?

bionicles:
TLDR - Updated doc:
image

@nzw0301 sorry for delay, I just saw this, you read my mind. **Yes, you're right, including `ask` in the doc to mention the preferred way to create new trials in custom optimization loops would guide users to create trials in this way.** To be clear, this doc comment is for the Trial class, not for the `Trial.__init__` constructor method. Should we include the names of the `suggest` methods? 

For use cases with long-running studies with long-running trials, where we interrupt and restart our optimization process with trials in progress, then **we need to re-make trials already created and lost from our local cache because we restarted the process or rebooted the machine, in order to get the parameters (not to suggest new ones) and to know the id**.

 - Study.optimize: automatically creates / uses Trials
 - Study.ask: manually creates Trials (custom optimization loop)
 - **Storage.get_trial: manually gets a frozen trial**
 - Study.tell: report values by trial id

So my issue was, "trial" on tell took an integer trial_id but I was constructing the Trial to pass that instead of just passing the trial_id. I think Storage.get_trial is the move, however that's on the Storage and not on Study, do you think it would make sense to add a `Study.get_trial` for consistent interface?

bionicles:
image

ok, here's how it looks now, please let me know if you think this is alright, it's 
a minor docs improvement (I wrote the `suggest` method interfaces so we can see how to use Trial when we mouse over the class)

nzw0301:
Thank you for your explanations. I think this PR discussed different topics at the same time, so let me leave comments to each.

### PR for Docs

I don't think we need to update `Trial` for the custom loop usage because we have https://optuna.readthedocs.io/en/stable/tutorial/20_recipes/009_ask_and_tell.html, and this modified section by the PR does not look directly related. Let me close the issue.

### resume optimisation

In my mind, there are three approaches for such a case:

1. https://optuna.readthedocs.io/en/stable/faq.html#can-i-monitor-trials-and-make-them-failed-automatically-when-they-are-killed-unexpectedly,
2. https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html#optuna.study.Study.enqueue_trial
3. ask / tell

I suggest using https://github.com/optuna/optuna/discussions for additional discussion on the later topic. 

Best regards,

bionicles:
that's a tutorial, not API reference

PR #5632, opened 2024-08-23
Remove the use of `evaluator.evaluate` function in the example of `PedAnovaImportanceEvaluator`

porink0424:
## Motivation

In `PedAnovaImportanceEvaluator`,  `evaluator.evaluate` function is not meant to be called by library users. However, it was used in the examples within the documentation, which could cause confusion. This PR resolves the issue by replacing it with an implementation that uses `optuna.importance.get_param_importances`.

## Description of the changes

Replaced the implementation with one that uses `optuna.importance.get_param_importances`.


nabenabe0928 (approved):
@porink0424 Thank you for the catch:) I did not notice it...!

HideakiImamura (approved):

PR #5390, opened 2024-04-12
Remove `multi_objective`

nabenabe0928:
## Motivation

As a milestone to v4.0, we decided to remove `multi_objective`, which was deprecated for a while.

## Description of the changes

I removed `multi_objective` and updated every single documents regarding `multi_objective` if exist.

PR #5640, opened 2024-08-27
Remove `optuna.samplers.MOTPESampler`

c-bata:
## Motivation

Follow-up https://github.com/optuna/optuna/pull/5390

While the implementation of `MOTPESampler` was removed in #5390, the `optuna.samplers.MOTPESampler` class itself still exists. In my opinion, the existence of `MOTPESampler` is a bit confusing as it may make it difficult to detect errors with mypy or flake8. It would be better to remove it at v4 major release, rather than leaving it in its current state.

## Description of the changes

Remove `MOTPESampler` from `optuna.samplers` module.

### DIFF
--- a/optuna/samplers/__init__.py
+++ b/optuna/samplers/__init__.py
@@ -26,13 +26,3 @@ __all__ = [
     "GPSampler",
     "nsgaii",
 ]
-
-
-class MOTPESampler:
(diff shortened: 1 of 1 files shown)


nabenabe0928 (approved):
LGTM!

not522 (approved):
LGTM!

PR #5645, opened 2024-08-28
Remove duplicated definition of license in `pyproject.toml`

toshihikoyanase:
## Motivation

In the metadata section of [pypi](https://pypi.org/project/optuna/), the license of optuna is not fully shown.
I think this is because `pyproject.toml` imports the contents of the `LICENSE` file.
Instead, we can use 'classifiers' starting with `License ::` according to [Python packaging user guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license).

image

The related PR of the OptunaHub library is https://github.com/optuna/optunahub/pull/58.

## Description of the changes

This PR stops importing the license file and keeps the [definition in the classifiers](https://github.com/optuna/optuna/blob/569191d1a16d4b5f1e54f59381e1293cbe80af0a/pyproject.toml#L17).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,6 @@ build-backend = "setuptools.build_meta"
 name = "optuna"
 description = "A hyperparameter optimization framework"
 readme = "README.md"
-license = {file = "LICENSE"}
 authors = [
   {name = "Takuya Akiba"}
 ]


nabenabe0928 (commented):
> If you are using a standard, well-known license, it is not necessary to use this field. Instead, you should use one of the [classifiers](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers) starting with License ::. (As a general rule, it is a good idea to use a standard, well-known license, both to avoid confusion and because some organizations avoid software whose license is unapproved.)

I also confirmed [the official statement](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license)!

nabenabe0928 (approved):
LGTM!

not522 (approved):
LGTM!

PR #5650, opened 2024-08-29
Add tips to the news section

nabenabe0928:
## Description of the changes

Add tips to the news section to let people notice the past articles.


toshihikoyanase (commented):
Hmm, I noticed we already have a link to Medium in the README.md.
Adding another link in this section seems out of place since it's intended to be a news section rather than a blog.

PR #5675, opened 2024-09-18
Remove performance benchmarks

not522:
## Motivation
The performance benchmark scripts are no longer used and are outdated, so this PR removes them.

## Description of the changes
- Remove workflow files of GitHub actions
- Remove benchmark scripts
- Remove benchmarks/README.md
  - This file does not contain the information of the speed benchmark.


eukaryo:

HideakiImamura (commented):
Basically LGTM. I think we can remove several dependencies from pyproject.toml. For example, we can remove `botorch` from the `benchmark` optional dependencies.

not522:
Thank you. I've updated this PR according to your comment.

HideakiImamura (approved):

eukaryo (approved):
LGTM.

PR #5676, opened 2024-09-18
Update PostgreSQL version in CI to latest

not522:
## Motivation
The current CI uses PostgreSQL 10.1, but it has already reached the end of life.
https://www.postgresql.org/support/versioning/

## Description of the changes
This PR updates the PostgreSQL version in CI to the latest.

### DIFF
--- a/.github/workflows/tests-storage.yml
+++ b/.github/workflows/tests-storage.yml
@@ -50,7 +50,7 @@ jobs:
           --health-timeout 5s
           --health-retries 5
       postgres:
-        image: postgres:10.1-alpine
+        image: postgres:latest
         ports:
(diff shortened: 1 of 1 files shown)


eukaryo:

eukaryo (approved):
LGTM; btw, how about specifying a version? For example, “current minor” of just one major version before “latest” (15.8 for now) also seems moderate.

not522:
Since the major version of PostgreSQL is updated every year, I think that CI tends to be stale if the version is fixed.

eukaryo:
Thank you, I understand.

y0z (approved):
LGTM

PR #5683, opened 2024-09-25
Add a comment for an unexpected bug in `CategoricalDistribution`

nabenabe0928:
## Motivation

Here is the unexpected error that has existed since the Optuna v0.0.1.

```python
import optuna

d = optuna.distributions.CategoricalDistribution([1, 1.0, True])

print(d.to_internal_repr(1))  # Return: 0, Expected: 0.
print(d.to_internal_repr(1.0))  # Return: 0, Expected: 1.
print(d.to_internal_repr(True))  # Return: 0, Expected: 2.
```

As this bug has existed such a long time and has not really been a problem, we decided to accept this bug for less complexity and better processing speed.

However, since this statement may be raised by some other developers or users again, I added a comment about this decision. 

## Description of the changes

- Add a comment line about our decision

### DIFF
--- a/optuna/distributions.py
+++ b/optuna/distributions.py
@@ -530,6 +530,10 @@ class CategoricalDistribution(BaseDistribution):
 
     def to_internal_repr(self, param_value_in_external_repr: CategoricalChoiceType) -> float:
         try:
+            # NOTE(nabenabe): With this implementation, we cannot distinguish some values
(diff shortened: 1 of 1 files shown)


nabenabe0928:
@not522 @eukaryo 
Could you review this PR?

not522 (approved):
LGTM

eukaryo (approved):
LGTM!

PR #5687, opened 2024-09-27
Add more information about the hack in `WFG`

nabenabe0928:
## Motivation

As the information about the hack in WFG was not sufficient, I added more information to prevent any future developers remove this line.

> [!NOTE]
> Without the hack, i.e. `assume_unique_lexsorted=True` although `limited_sols` could be non-unique, `_compute_hv` becomes an exponential algorithm with the time complexity of $O(2^K)$ where $K$ is the maximum duplicated number of a Pareto solution.
> Let's define the time complexity of `_compute_hv` with $K$ as $f(K)$ and $f(1) = 1$, then $f(K) = \sum_{k=1}^{K - 1} f(k)$ because of the recursive algorithm without the hack. This leads to the time complexity given above.

## Description of the changes

- Add some comment lines


y0z (approved):

HideakiImamura:

not522 (approved):
LGTM

PR #5688, opened 2024-10-01
Update document and notice interoperability between NSGA-II and Pruners

eukaryo:
## Motivation

NSGA-II sampler and a pruner can be used together if the objective function is single-objective. I want to update ReadTheDocs to notice this fact.

The following code works:
```python
import numpy as np
import optuna


def objective(trial):
    seed = trial.suggest_int("seed", 0, 10000)
    rng = np.random.default_rng(seed)
    total_sum = 0.0
    for i in range(10):
        x = rng.normal()
        total_sum += x
        trial.report(value=total_sum, step=i)
        if trial.should_prune():
            raise optuna.TrialPruned()
    return total_sum


study = optuna.create_study(
    direction="maximize",
    sampler=optuna.samplers.NSGAIISampler(),
    pruner=optuna.pruners.MedianPruner(
        n_startup_trials=5, n_warmup_steps=1, interval_steps=1
    ),
)
```

## Description of the changes

Update `index.rst`.


eukaryo:
The new apperance:

not522:
Since `PRUNED` trials are simply ignored, wouldn't it be more accurate to say "Works, but inefficiently." rather than "Supports this feature."? For example, if all trials are `PRUNED`, all generations of trials are 0.

```python
import optuna

def objective(trial):
    trial.suggest_float("x", 0, 10)
    raise optuna.TrialPruned()

study = optuna.create_study(sampler=optuna.samplers.NSGAIISampler())
for trial in study.trials:
    print(trial.system_attrs)
```

eukaryo:
Sorry for my late reply. I agree; I have modified the code.

not522 (approved):
Thank you for your update. It's LGTM!
@HideakiImamura  Could you take another look for the change?

HideakiImamura (approved):

PR #5691, opened 2024-10-02
Run tests with Python 3.13

gen740:
## Motivation

Python 3.13 is coming. So I added the python3.13 to the test to support Python 3.13.

## Description of the changes

- Add `3.13.0-rc.2` to the python matrix
- Do not install torch if python >= 3.13
- Ignore tests that import `torch`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,6 +77,11 @@ jobs:
         pip install --progress-bar off .[test] --extra-index-url https://download.pytorch.org/whl/cpu
         pip install --progress-bar off .[optional] --extra-index-url https://download.pytorch.org/whl/cpu
 
+        # TODO(gen740): Remove this after pytorch supports Python 3.13
+        if [ "${{ matrix.python-version }}" = "3.13" ] ; then
+          pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cpu
+        fi
+
     - name: Output installed packages
       run: |
         pip freeze --all
@@ -94,19 +99,6 @@ jobs:
           target="not slow"
         fi
 
-        if [ "${{ matrix.python-version }}" = "3.13" ] ; then
-          pytest tests -m "$target" \
-            --ignore=tests/gp_tests/ \
-            --ignore=tests/samplers_tests/test_gp.py \
-            --ignore=tests/terminator_tests/improvement_tests/test_evaluator.py \
-            --ignore=tests/terminator_tests/improvement_tests/test_evaluator.py \
-            --ignore=tests/samplers_tests/test_samplers.py \
-            --ignore=tests/terminator_tests/improvement_tests/test_emmr_evaluator.py \
-            --ignore=tests/terminator_tests/test_median_erroreval.py \
-            --ignore=tests/terminator_tests/test_terminator.py \
-            --ignore=tests/visualization_tests/test_terminator_improvement.py
-        else
-          pytest tests -m "$target"
-        fi
+        pytest tests -m "$target"
       env:
         SQLALCHEMY_WARN_20: 1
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -89,7 +89,7 @@ optional = [
   "scikit-learn>=0.24.2",
   # optuna/visualization/param_importances.py.
   "scipy",  # optuna/samplers/_gp
-  "torch",  # optuna/samplers/_gp
+  "torch; python_version<='3.12'", # TODO(gen740): Remove this line after torch supports Python 3.13
```
toshihikoyanase, line 92: The information that `torch` is required by `optuna/samplers/_gp` is missing in this PR.
Including a comment such as '# scipy and torch are required for optuna/samplers/_gp' would ensure that other contributors know the necessary modules' functionality packages.

(I'm not sure if writing dependencies in comments is a good practice. For example, I don't track whether those dependencies are updated when used in other modules.)

toshihikoyanase:
I think this is out of the scope of this PR, but the nightly build of `torch` seems to support Python 3.13, according to https://github.com/pytorch/pytorch/issues/130249. You might be able to try the skipped test cases with it.

> pip3 install --pre torch  --index-url https://download.pytorch.org/whl/nightly/cpu

I tried it, but the macOS binary is currently unavailable. For now, only the Linux binary seems to be provided.

pyproject.toml
gen740: In fact, also `optuna/_gp` has `torch` as a dependency (and also terminator, because it depends on `optuna/_gp`). So I think it is not realistic to list them all.

I realize that the above comment about 'scipy' is incorrect. It is not 'optuna/samplers/_gp' that uses 'scipy,' but rather 'optuna/_gp.'

I would comment like this `# TODO(gen740): Remove this line when 'torch', a dependency of 'optuna/_gp', supports Python 3.13` What do you think about this @toshihikoyanase?

toshihikoyanase: > I would comment like this # TODO(gen740): Remove this line when 'torch', a dependency of 'optuna/_gp', supports Python 3.13

Sounds reasonable. It will minimize the change while keeping the dependency information. Thanks!
(If you have a line-length issue, please keep the comment you proposed originally.)

> In fact, also optuna/_gp has torch as a dependency (and also terminator, because it depends on optuna/_gp). So I think it is not realistic to list them all.
> I realize that the above comment about 'scipy' is incorrect. It is not 'optuna/samplers/_gp' that uses 'scipy,' but rather 'optuna/_gp.'

Hmm, we may want to discuss the policy of comments to manage dependencies. But it is definitely out of scope of this PR.

not522:
Let me change it to Draft PR, as we are waiting for the Python 3.13 environment to be ready.

gen740:
I have just updated the test.yml as setup-python has been updated to support Python 3.13.

y0z:
@HideakiImamura 

Could you review this PR?

.github/workflows/tests.yml
```diff
@@ -94,6 +94,19 @@ jobs:
...
         fi
 
-        pytest tests -m "$target"
+        if [ "${{ matrix.python-version }}" = "3.13" ] ; then
+          pytest tests -m "$target" \
+            --ignore=tests/gp_tests/ \
+            --ignore=tests/samplers_tests/test_gp.py \
+            --ignore=tests/terminator_tests/improvement_tests/test_evaluator.py \
+            --ignore=tests/terminator_tests/improvement_tests/test_evaluator.py \
+            --ignore=tests/samplers_tests/test_samplers.py \
```
HideakiImamura, line 103: The `test_samplers.py` includes a lot of tests for other samplers. How about installing the nightly version of `torch` in the CI, and do not skip all tests?

HideakiImamura (approved):
LGTM.

y0z (approved):
LGTM

PR #5693, opened 2024-10-04
Reduce `SELECT` statements by using outer join

porink0424:
## Motivation 

This PR is motivated by the desire to improve performance by reducing the number of unnecessary `SELECT` statements being executed in the basic optimization flow.

## Description of the changes

- In `RDBStorage.get_trial`, individual queries were being executed due to lazy loading when creating `frozen_trial` right after retrieving `trial`. I modified this to use an outer join, which eliminates the need for these separate queries.

## Verification of reduced `SELECT` statements

```python
import optuna
import logging

study = optuna.create_study(storage="mysql+pymysql://optuna:password@127.0.0.1:3306/optuna")


    s = 0.0
    for i in range(8):
        s += trial.suggest_float(f"x{i}", 0.0, 1.0)
    return s


logging.basicConfig()
```

```bash
❯ python mysql_optimize.py 2>! tmp.txt
❯ grep -o SELECT tmp.txt | wc -l
```

### Number of `SELECT` statements

Before the change: 66
After the change: 51

### DIFF
--- a/optuna/storages/_rdb/models.py
+++ b/optuna/storages/_rdb/models.py
@@ -239,17 +239,28 @@ class TrialModel(BaseModel):
 
     @classmethod
     def find_or_raise_by_id(
-        cls, trial_id: int, session: orm.Session, for_update: bool = False
+        cls,
+        trial_id: int,
+        session: orm.Session,
+        for_update: bool = False,
+        include_trial_values: bool = False,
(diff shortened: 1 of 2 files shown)


porink0424:
I ran a quick benchmark test, but it didn't show any improvement, so I am closing this PR.

PR #5706, opened 2024-10-15
Fix for mypy.

y0z:
## Motivation & Description of the changes

Checks CI fails because of an upgrade to the latest version of mypy (v 1.12.0).
This PR modifies some code slightly to resolve the corresponding errors.

- For `study.py`, an error is caused by the return type of [`numpy.isnan`](https://numpy.org/doc/2.0/reference/generated/numpy.isnan.html), which is `ndarray or bool` rather than `bool`.  In our case, adding an explicit conversion to `bool` resolves this error.
- For `_convert_positional_args.py`, I add `# type: ignore` since I can't find the cause of this strange mypy error.

### DIFF
--- a/optuna/_convert_positional_args.py
+++ b/optuna/_convert_positional_args.py
@@ -80,7 +80,7 @@ def convert_positional_args(
 
             kwargs.update(inferred_kwargs)
 
-            return func(**kwargs)
+            return func(**kwargs)  # type: ignore
 
         return converter_wrapper
 
--- a/optuna/_convert_positional_args.py
+++ b/optuna/_convert_positional_args.py
@@ -80,7 +80,7 @@ def convert_positional_args(
 
             kwargs.update(inferred_kwargs)
 
-            return func(**kwargs)  # type: ignore
+            return func(**kwargs)  # type: ignore[call-arg]
 
         return converter_wrapper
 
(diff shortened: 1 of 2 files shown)


optuna/study/study.py
```diff
@@ -1091,6 +1091,7 @@ def _should_skip_enqueue(self, params: Mapping[str, JSONSerializable]) -> bool:
                     if isinstance(param_value, Real)
                     else param_value == existing_param
                 )
+                assert isinstance(is_repeated, bool)
```
nzw0301, line 1094: This can be `False`.  How about `repeated_params: list[bool | np.bool] = []` or `bool(is_repeated)`?

y0z: Thank you. I applied your approach.

nzw0301 (approved):
Thanks!

optuna/_convert_positional_args.py
```diff
@@ -80,7 +80,7 @@ def converter_wrapper(*args: Any, **kwargs: Any) -> "_T":
 
             kwargs.update(inferred_kwargs)
 
-            return func(**kwargs)
+            return func(**kwargs)  # type: ignore
```
not522, line 83:
```suggestion
            return func(**kwargs)  # type: ignore[call-arg]
```

not522: I'm not sure, but it looks similar to this bug report https://github.com/python/mypy/issues/17960, so I think it's OK to ignore it for now.

y0z: Thank you for your suggestion.

HideakiImamura (approved):
LGTM!

PR #5709, opened 2024-10-16
Reduce `SELECT` statements by removing unnecessary distribution compatibility check in `set_trial_param()`

porink0424:
## Motivation

The `set_trial_param()` process includes a flow that checks if the same `param_name` has already been set, issuing `UPDATE` if it exists or `INSERT` if it does not. However, the need to issue `UPDATE` instead of `INSERT` is limited to few cases: when storage is used directly or when parameters are suggested from multiple processes. By removing this check, we can reduce the number of `SELECT` statements.

## Description of the changes

- Modified the storage logic so that it always issues `INSERT`, regardless of the case.
- Removed test cases where this check was necessary.

## Verification of reduced `SELECT` statements

```python
import optuna
import logging

study = optuna.create_study(storage="mysql+pymysql://optuna:password@127.0.0.1:3306/optuna")


    s = 0.0
    for i in range(8):
        s += trial.suggest_float(f"x{i}", 0.0, 1.0)
    return s


logging.basicConfig()
```

```
❯ python mysql_optimize.py 2>! tmp.txt
❯ grep -o SELECT tmp.txt | wc -l
```

I confirmed by running the above code that 8 `SELECT` statements were reduced.  Note that this is  because `suggest_float()` is called 8 times.

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -588,27 +588,14 @@ class RDBStorage(BaseStorage, BaseHeartbeat):
         trial = models.TrialModel.find_or_raise_by_id(trial_id, session)
         self.check_trial_is_updatable(trial_id, trial.state)
 
-        trial_param = models.TrialParamModel.find_by_trial_and_param_name(
-            trial, param_name, session
+        trial_param = models.TrialParamModel(
+            trial_id=trial_id,
+            param_name=param_name,
+            param_value=param_value_internal,
+            distribution_json=distributions.distribution_to_json(distribution),
         )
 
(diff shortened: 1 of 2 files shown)


y0z:
@eukaryo @HideakiImamura 

Could you review this PR?

HideakiImamura (approved):
Looks very good. I confirmed that the distribution compatibility check is conducted in the Trial. https://github.com/optuna/optuna/blob/74e36182e7234b4205aa8e044ba217222fc4ab96/optuna/trial/_trial.py#L620

eukaryo (approved):
Thank you for your PR! LGTM.

PR #5720, opened 2024-10-23
Include all test files in `sdist` by updating `MANIFEST.in`

porink0424:
## Motivation

Previously, only files under `tests` directory that matched the pattern `test*.py` were included in the Optuna `sdist` package, resulting in an incomplete state. This PR resolves this issue by modifying `MANIFEST.in` to include all files under `tests` directory in the `sdist`.

## Description of the changes

- All files under `tests` are now included in the `sdist` package, excluding files like `*.pyc`.
- When the `license-files` parameter is not specified in `pyproject.toml`, files matching `LICENSE*` are automatically included in `sdist` [(reference)](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html#:~:text=All%20files%20specified%20by%20the%20license,COPYING*%2C%20NOTICE*%2C%20AUTHORS**%3B). As a result, the two files that were explicitly included in the original `MANIFEST.in` (i.e., `LICENSE` and `LICENSE_THIRD_PARTY`) no longer need to be specified, so this PR removes those explicit entries.

## Reference

- [Controlling files in the distribution](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html)

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,2 @@
-include LICENSE
-include LICENSE_THIRD_PARTY
+graft tests
+global-exclude *~ *.py[cod] *.so


c-bata:
@HideakiImamura @not522 Could you review this PR?

not522 (approved):
LGTM!
I checked that the distribution includes test and license files.

HideakiImamura (approved):
LGTM. I confirmed that the generated package includes `test` and `LICENSE`

PR #5729, opened 2024-10-25
Eliminated redundant for loops in `plot_parallel_coordinate()`

Param1304:
## Motivation

In parallel coordinate plot, _is_log_scale(), _is_categorical(), and _is_numerical() functions are invoked. Each one of these functions iterates over the trials, which is clearly redundant. plot_parallel_coorinate() can be made faster by removing wasted for-loop.

## Description of the changes

Added a new method _preprocess_trial_params(). This function performs a single pass through trials to populate is_log_scale and is_numerical dictionaries. These dictionaries store whether each parameter is on a log scale or is numerical, eliminating the need for repeated checks. Updated _is_log_scale() and _is_numerical(). They now take the precomputed dictionarieds as arguments and perform a simple dictionary lookup, removing redundant loops. This change should enhance performance when dealing with large datasets by reducing the computational overhead in plot_parallel_coordinate thus solving issue #5409 .

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,60 @@
+# pre-commit package installation is necessary to use pre-commit.
+# $ pip install pre-commit
+# $ pre-commit install
+
+default_language_version:
+  python: python3
+
+repos:
+  # Args are based on setup.cfg.
+  - repo: https://github.com/psf/black
+    rev: 24.8.0
+    hooks:
+      - id: black
+  - repo: https://github.com/pycqa/flake8
+    rev: 7.1.1
+    hooks:
+      - id: flake8
+        exclude: tutorial|docs/visualization_examples|docs/visualization_matplotlib_examples
+        args: [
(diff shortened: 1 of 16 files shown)


nzw0301 (changes requested):
Hi thank you for sending a PR. I've just skimmed the PR and let me leave a few comments.

1. Could you remove pandas? It looks like unnecessary dependency.
2. Could you fix the test errors?
3. Could you report execution time before and after this PR with minimal reproducible script as a comment? I think that this PR should shorten the execution time by removing unnecessary loops.

Param1304:
Yes I will do that

nabenabe0928:
@Param1304 
Thank you for the PR and I am sorry to tell you, but the issue below
- https://github.com/optuna/optuna/issues/5409
was already resolved as mentioned in 
- https://github.com/optuna/optuna/issues/5409#issuecomment-2406568771

I should have closed the issue.

PR #5733, opened 2024-10-30
Rename `test_cache_is_invalidated` and remove `assert study._thread_local.cached_all_trials is None`

nabenabe0928:
## Motivation

The original version test checks the cache clear after before_trial, so if we call `_get_all_trials` with `use_cache=True` at `before_trial`, this unit test fails.
To avoid this problem, we actually need to check the cache clear before `before_trial`.
To this end, I introduced a mock function to do so.

## Description of the changes

- Add a mock function so that we can check the cache clear before each trial

> [!NOTE]
> I thought through if I should add any tests for this regard to `study_tests`, but as this check should happen before `before_trial`, which is a method of each sampler, I simply modified the existing test in `test_samplers.py` instead of `study_tests`.

### DIFF
--- a/tests/samplers_tests/test_samplers.py
+++ b/tests/samplers_tests/test_samplers.py
@@ -1043,22 +1043,17 @@ def test_reproducible_in_other_process(sampler_name: str, unset_seed_in_test: No
 
 @pytest.mark.parametrize("n_jobs", [1, 2])
 @parametrize_relative_sampler
-def test_cache_is_invalidated(
+def test_trial_relative_params(
     n_jobs: int, relative_sampler_class: Callable[[], BaseSampler]
 ) -> None:
+    # TODO(nabenabe): Consider moving this test to study.
     sampler = relative_sampler_class()
-    original_before_trial = sampler.before_trial
-
-    def mock_before_trial(study: Study, trial: FrozenTrial) -> None:
-        assert study._thread_local.cached_all_trials is None
-        original_before_trial(study, trial)
 
     with patch.object(sampler, "before_trial", side_effect=mock_before_trial):
         study = optuna.study.create_study(sampler=sampler)
 
         def objective(trial: Trial) -> float:
             assert trial._relative_params is None
-
             trial.suggest_float("x", -10, 10)
             trial.suggest_float("y", -10, 10)
             assert trial._relative_params is not None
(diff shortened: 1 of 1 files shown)


nabenabe0928:
@y0z 
Could you review this PR?

tests/samplers_tests/test_samplers.py
```diff
@@ -1047,11 +1048,17 @@ def test_cache_is_invalidated(
     n_jobs: int, relative_sampler_class: Callable[[], BaseSampler]
 ) -> None:
     sampler = relative_sampler_class()
+    original_before_trial = sampler.before_trial
+
+    def mock_before_trial(study: Study, trial: FrozenTrial) -> None:
+        assert study._thread_local.cached_all_trials is None
+        original_before_trial(study, trial)
+
+    sampler.before_trial = Mock(side_effect=mock_before_trial)
```
y0z, line 1057: This line causes the "tests/samplers_tests/test_samplers.py:1057: error: Cannot assign to a method  [method-assign]" mypy error.

I think [`mock.patch.object`](https://docs.python.org/3.13/library/unittest.mock.html) is a better way to replace a class method.

Example:
```python
import unittest.mock


class C:
    def method(self):
        print("hello")

def mock_method():
    print("hi")

c = C()

c.method()
with unittest.mock.patch.object(C, 'method', unittest.mock.Mock(side_effect=mock_method)):
    c.method()
```
Output:
```
hello
hi
```

y0z (approved):
LGTM

nabenabe0928:
@HideakiImamura 
Could you review this PR?

HideakiImamura:
Thanks for the PR. IMO, the line of `assert study._thread_local.cached_all_trials is None` should be removed from the current test function, because the it is not a specification that the trials are not cached before the call of suggest methods at now. 

If we want to test the behavior of the trial cache between sampler's before trial method and the cache creation, it would be better to add the additional test cases. However, I think we don't have to specify this behavior, so removing the assertion is sufficient, IMO.

HideakiImamura (approved):
Thanks for the update. Looks very good.

HideakiImamura:
@y0z Please merge this PR after your approval, since the direction of this PR is completely changed.

y0z:
The change looks OK. I rename the PR title since the first title is no longer proper.

PR #5727, opened 2024-10-25
Drop the support for Python 3.7 and update package metadata for Python 3.13 support

c-bata:
## Motivation
Closes https://github.com/optuna/optuna/issues/4586

## Description of the changes
This PR updates followings to drop the support for Python 3.7.

* CI settings
* README and documentation
* Package metadata

For any remaining TODO comments in the implementation and tests, I will open a follow-up PR after this one is merged: https://github.com/optuna/optuna/compare/master...c-bata:optuna:resolve-todo-comments-for-python3.7?expand=1

PR #5736, opened 2024-10-30
Remove dependency specifier for installing SciPy

not522:
## Motivation

Follow-up #5727

`requires-python` has been updated for `>=3.8`. So, the dependency specifier for installing SciPy is no longer required.

## Description of the changes

- Remove dependency specifier for installing SciPy

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,7 +97,7 @@ test = [
   "kaleido",
   "moto",
   "pytest",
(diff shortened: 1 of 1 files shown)


c-bata (approved):
Good catch! LGTM.

HideakiImamura (approved):
LGTM.

PR #5737, opened 2024-10-31
Fix a concurrency issue in `GPSampler`

c-bata:
## Motivation

As discussed in https://github.com/optuna/optunahub-registry/pull/168#pullrequestreview-2404054969, there is a concurrency issue in `GPSampler`.

## Description of the changes

Through investigation by @not522, it was found that SciPy's Sobol engine can cause concurrency-related errors, which can be reproduced with the following code:

```python
import scipy.stats.qmc as qmc
from concurrent.futures import ThreadPoolExecutor

futures = set()
with ThreadPoolExecutor(max_workers=2) as executor:
    for i in range(2):
        futures.add(executor.submit(qmc.Sobol, 2))
```

This PR addresses the issue by using `threading.Lock` to prevent these exceptions.

### DIFF
--- a/optuna/_gp/search_space.py
+++ b/optuna/_gp/search_space.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from enum import IntEnum
 import math
+import threading
 from typing import Any
 from typing import TYPE_CHECKING
 
@@ -21,6 +22,9 @@ else:
     from optuna._imports import _LazyImport
 
     qmc = _LazyImport("scipy.stats.qmc")
+
+
+_threading_lock = threading.Lock()
 
 
 class ScaleType(IntEnum):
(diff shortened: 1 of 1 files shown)


optuna/_gp/search_space.py
```diff
@@ -92,8 +96,14 @@ def sample_normalized_params(
...
     steps = search_space.steps
-    qmc_engine = qmc.Sobol(dim, scramble=True, seed=rng.randint(np.iinfo(np.int32).max))
-    param_values = qmc_engine.random(n)
+
+    # Sobol engine likely shares its internal state among threads.
+    # Without threading.Lock, ValueError exceptions are raised in Sobol engine as discussed in
+    # https://github.com/optuna/optunahub-registry/pull/168#pullrequestreview-2404054969
+    with _threading_lock:
+        qmc_engine = qmc.Sobol(dim, scramble=True, seed=rng.randint(np.iinfo(np.int32).max))
+        param_values = qmc_engine.random(n)
```
not522, line 105: I'm not sure about the behavior of `Sobol`, but it works without the lock for the sampling.

```suggestion
    param_values = qmc_engine.random(n)
```

c-bata: Thank you for your review.

I'm also unsure about the behavior. Considering the Sobol's sampling works faster, it would be safer to lock both two lines for now. What do you think?

c-bata:
@nabenabe0928 Could you assign one additional reviewer?

y0z:
I empirically confirmed that this error does not occur after applying this PR with the following toy example.

```python
import optuna

optuna.create_study(
    sampler=optuna.samplers.GPSampler()
).optimize(
    lambda t: t.suggest_float("x", 0, 1)**2 + t.suggest_float("y", 0, 1)**2,
    n_trials=20,
    n_jobs=2
)
```

y0z (approved):
As a hotfix, the fix seems good, but I'm not so confident since I'm not familiar with the internal implementation of Sobol.

optuna/_gp/search_space.py
c-bata: We discussed offline and decided to apply this suggestion.

not522 (approved):
LGTM!
I investigated the cause and confirmed that the thread lock for the initialization of `Sobol` is enough to fix this error. ref: https://github.com/scipy/scipy/pull/21794

Issue #5732, opened 2024-10-29
Use Python 3.12 to build the documentation in CI


c-bata:
## Motivation

The Optuna documentation is currently built on GitHub Actions using Python 3.8, which has officially reached end-of-life (EOL). The build environment should be updated to a non-EOL Python version, specifically Python 3.9 or later.

## Description

An attempt to upgrade the Python version from 3.8 to 3.12 was made in [PR #5728](https://github.com/optuna/optuna/pull/5728), but it resulted in CI failures ([error logs](https://github.com/optuna/optuna/actions/runs/11549638785/job/32143038062)). This issue aims to update the Python version used in the GitHub Actions workflow to a non-EOL version (preferably Python 3.9 or newer).

PR #5738, opened 2024-10-31
Avoid using functions in `sphinx_gallery_conf` to remove document build error with python 3.12

nzw0301:
## Motivation

Reduce one warning reported in https://github.com/optuna/optuna/issues/5732. Concretely, 

```shell
WARNING: cannot cache unpickable configuration value: 'sphinx_gallery_conf' (because it contains a function, class, or module object) [config.cache]
```

## Description of the changes

Instead of callable as a value of `sphinx_gallery_conf`, use the string alias of `FileNameSortKey` as implemented in https://github.com/sphinx-gallery/sphinx-gallery/pull/1289 and string expression for plotly by following https://sphinx-gallery.github.io/stable/configuration.html#importing-callables.

PR #5742, opened 2024-11-01
Use Python 3.12 for docs build CI

nzw0301:
## Motivation

Resolve https://github.com/optuna/optuna/issues/5732.

## Description of the changes

- Update python version 3.12 in `.github/workflows/sphinx-build.yml`
- Remove `'WARNING: Summarised items should not include the current module'`, which are the majority of changes in this PR. This change is suggested by the warning message like `WARNING: Summarised items should not include the current module. Replace 'optuna.delete_study' with 'delete_study'. [autosummary.import_cycle]` in https://github.com/optuna/optuna/actions/runs/11640162393/job/32417223006.
    - This change removes a parent module prefix of each class/method in a table on a page. for example

before: 
Screenshot 2024-11-03 at 0 53 33

after:
Screenshot 2024-11-03 at 1 04 09

- Remove "WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]" for `TrialState` and `StudyDirection`. As I commented in https://github.com/optuna/optuna/pull/5742#discussion_r1826508957, I'm not sure how this change affects the docs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -118,9 +118,9 @@ Whether you edit any tutorial or not doesn't matter.
 
 To avoid having to run the tutorials, you may download executed tutorial artifacts named "tutorial" from our CI (see the capture below) and put them in `docs/build` before
 extracting the files in the zip to `docs/source/tutorial` directory.
-Note that the CI runs with Python 3.8 and the generated artifacts contain pickle files.
+Note that the CI runs with Python 3.12 and the generated artifacts contain pickle files.
 The pickle files are serialized with [the protocol version 5](https://docs.python.org/3/library/pickle.html#data-stream-format) so you will see the error with Python 3.7 or older.
-Please use Python 3.8 or later if you build the documentation with artifacts.
+Please use Python 3.12 or later if you build the documentation with artifacts.
 
 ![image](https://user-images.githubusercontent.com/16191443/107472296-0b211400-6bb2-11eb-9203-e2c42ce499ad.png)
 
@@ -205,4 +205,3 @@ Once you get a good understanding of how Minituna is designed, it will not be to
 We encourage you to practice reading the Minituna code with the following article.
 
 [An Introduction to the Implementation of Optuna, a Hyperparameter Optimization Framework](https://medium.com/optuna/an-introduction-to-the-implementation-of-optuna-a-hyperparameter-optimization-framework-33995d9ec354)
-
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,8 +119,7 @@ Whether you edit any tutorial or not doesn't matter.
 To avoid having to run the tutorials, you may download executed tutorial artifacts named "tutorial" from our CI (see the capture below) and put them in `docs/build` before
 extracting the files in the zip to `docs/source/tutorial` directory.
 Note that the CI runs with Python 3.12 and the generated artifacts contain pickle files.
-The pickle files are serialized with [the protocol version 5](https://docs.python.org/3/library/pickle.html#data-stream-format) so you will see the error with Python 3.7 or older.
-Please use Python 3.12 or later if you build the documentation with artifacts.
+Please use the same Python version as in the CI if you build the documentation with artifacts to avoid unexpected errors due to the python version difference.
 
 ![image](https://user-images.githubusercontent.com/16191443/107472296-0b211400-6bb2-11eb-9203-e2c42ce499ad.png)
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -186,7 +186,7 @@ autosummary_generate = True
 autodoc_typehints = "description"
 autodoc_default_options = {
     "members": True,
-    "inherited-members": True,
+    "inherited-members": "int",
     "exclude-members": "with_traceback",
 }
 
--- a/docs/source/reference/storages.rst
+++ b/docs/source/reference/storages.rst
@@ -9,11 +9,11 @@ The :mod:`~optuna.storages` module defines a :class:`~optuna.storages.BaseStorag
    :toctree: generated/
    :nosignatures:
 
-   optuna.storages.RDBStorage
-   optuna.storages.RetryFailedTrialCallback
-   optuna.storages.fail_stale_trials
-   optuna.storages.JournalStorage
-   optuna.storages.InMemoryStorage
+   RDBStorage
+   RetryFailedTrialCallback
+   fail_stale_trials
+   JournalStorage
+   InMemoryStorage
 
 optuna.storages.journal
 -----------------------
@@ -27,8 +27,8 @@ optuna.storages.journal
    :toctree: generated/
    :nosignatures:
 
-   optuna.storages.journal.JournalFileBackend
-   optuna.storages.journal.JournalRedisBackend
+   journal.JournalFileBackend
+   journal.JournalRedisBackend
 
 Users can flexibly choose a lock object for :class:`~optuna.storages.journal.JournalFileBackend` and here is the list of supported lock objects:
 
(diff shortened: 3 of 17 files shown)


docs/source/conf.py
```diff
@@ -188,7 +186,7 @@
 autodoc_typehints = "description"
 autodoc_default_options = {
     "members": True,
-    "inherited-members": True,
+    "inherited-members": "int",
```
nzw0301, line 189: I'm not sure about this attribute, but when this value is `True` with python 3.12, we got the following warning messages:

```shell
optuna/study/_study_direction.py:docstring of optuna.study.StudyDirection.from_bytes:9: WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]
optuna/study/_study_direction.py:docstring of optuna.study.StudyDirection.to_bytes:8: WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]
optuna/trial/_state.py:docstring of optuna.trial.TrialState.from_bytes:9: WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]
optuna/trial/_state.py:docstring of optuna.trial.TrialState.to_bytes:8: WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]
```

As explained by https://stackoverflow.com/questions/77914856/sphinx-raises-warnings-for-class-derived-from-intenum-can-i-do-something-about, which was shared by @himkt privately, the workaround is that wait for python 3.13 or use `replace `True` with int` as in this line. 

I appreciate any comments on this.

## Master

Screenshot 2024-11-02 at 15 15 40

## This PR

Screenshot 2024-11-02 at 15 15 22

HideakiImamura:
Is this PR ready for review? It would be great to discuss several changes from now, IMO. How about that?

nzw0301:
Thank you for your comments. Yes it is. I will merge the master branch into this PR since this PR depended on https://github.com/optuna/optuna/pull/5738 in a few hours.

HideakiImamura:
@y0z @kAIto47802 Could you review this PR?

docs/source/conf.py
HideakiImamura: It seems that the page of `optuna.trial.TrialState` as you mentioned with the current PR is completely same as that with the master branch.

https://optuna--5742.org.readthedocs.build/en/5742/reference/generated/optuna.trial.TrialState.html#optuna.trial.TrialState

Screenshot 2024-11-06 at 14 47 27

Do I misunderstand something?

nzw0301: Interesting. It sounds like my local Sphinx build issue. Never mind.

not522: I think that ReadTheDocs did not use Python 3.12 because `.readthedocs.yml` was not updated.
https://github.com/optuna/optuna/blob/9def06d79213526d565eada4612eeeedbc2db060/.readthedocs.yml#L11

nzw0301: Thank you for letting us know that. Should we update this too?

nzw0301: I've updated python version in `.readthedocs.yml`. Then `optuna.trial.TrialStat` page is different from the master's docs as I written above.

https://optuna--5742.org.readthedocs.build/en/5742/reference/generated/optuna.trial.TrialState.html#optuna.trial.TrialState

HideakiImamura:
@nzw0301 Thanks for the update and discussions. I discussed the direction of this PR with committers, and completely agreed with the current direction. The change of looking of documents is acceptable. Could you proceed this PR to merge?

nzw0301:
@HideakiImamura thank you for sharing the updates. Sure but what should I do next for this PR? I suppose this PR is ready for reviews and the CI error was the python 3.13 test that this PR should not affect.

y0z (approved):
LGTM

CONTRIBUTING.md
```diff
@@ -118,9 +118,9 @@ Whether you edit any tutorial or not doesn't matter.
 
 To avoid having to run the tutorials, you may download executed tutorial artifacts named "tutorial" from our CI (see the capture below) and put them in `docs/build` before
 extracting the files in the zip to `docs/source/tutorial` directory.
-Note that the CI runs with Python 3.8 and the generated artifacts contain pickle files.
+Note that the CI runs with Python 3.12 and the generated artifacts contain pickle files.
 The pickle files are serialized with [the protocol version 5](https://docs.python.org/3/library/pickle.html#data-stream-format) so you will see the error with Python 3.7 or older.
-Please use Python 3.8 or later if you build the documentation with artifacts.
+Please use Python 3.12 or later if you build the documentation with artifacts.
```
c-bata, line 123: nit: According to the documentation, the Pickle protocol version has not been incremented since Python 3.8, so it may be fine to keep it as is.

nzw0301: Thanks. Absolutely right. Alternatively, how about removing the previous line 

> The pickle files are serialized with [the protocol version 5](https://docs.python.org/3/library/pickle.html#data-stream-format) so you will see the error with Python 3.7 or older.
? 
Since a different python version from Optuna sphinx CI causes different behaviours of sphinx build when a contributor builds the docs locally.

c-bata: Agree. We should simply recommend building with the same Python version.

docs/source/reference/storages.rst
```diff
@@ -27,17 +27,17 @@ optuna.storages.journal
...
 Users can flexibly choose a lock object for :class:`~optuna.storages.journal.JournalFileBackend` and here is the list of supported lock objects:
 
 .. autosummary::
    :toctree: generated/
    :nosignatures:
 
-   optuna.storages.journal.JournalFileSymlinkLock
-   optuna.storages.journal.JournalFileOpenLock
+   journal.JournalFileSymlinkLock
+   journal.JournalFileOpenLock
```
HideakiImamura, line 40: I think it would be better to remove the prefix of `journal.` here. It is obvious that these are placed inside `optuna.storages.journal`.

HideakiImamura (commented):
Thanks for the update. I have a minor comment. PTAL.

docs/source/reference/storages.rst
nzw0301: I'm unsure why these changes were not suggested in sphinx's build message https://github.com/optuna/optuna/actions/runs/11640162393/job/32417223006. Anyway, I'll take a look at them. Thanks.

nzw0301: When I removed `journal.` from the lines, the sphinx shows the following warning messages:
```
WARNING: [autosummary] failed to import optuna.storages.JournalFileBackend.
Possible hints:
* ModuleNotFoundError: No module named 'optuna.storages.JournalFileBackend'
* ImportError:
* AttributeError: module 'optuna.storages' has no attribute 'JournalFileBackend'
WARNING: [autosummary] failed to import optuna.storages.JournalRedisBackend.
Possible hints:
* ImportError:
* ModuleNotFoundError: No module named 'optuna.storages.JournalRedisBackend'
* AttributeError: module 'optuna.storages' has no attribute 'JournalRedisBackend'
```

nzw0301: Maybe alias causes this issue. in my understanding, `DeprecatedJournalFileOpenLock` and `DeprecatedJournalFileSymlinkLock` are actual class names.

nzw0301: Let me clarify this a little bit more.
To remove `journal` prefix here, the definition, they should be inhttps://github.com/optuna/optuna/blob/master/optuna/storages/__init__.py. However, 
`optuna.storages.JournalFileSymlinkLock` and `JournalFileOpenLock` are aliases of `DeprecatedJournalFileSymlinkLock` and `DeprecatedJournalFileOpenLock`, respectively.

nzw0301: In my mind importing the following classes 

```python
JournalFileBackend
JournalRedisBackend
JournalFileSymlinkLock  # deprecatedのaliasではない方
JournalFileOpenLock  # deprecatedのaliasではない方
``` 
in `optuna/storages/__init__.py` is an approach to remove the prefix.

HideakiImamura: Thanks for the investigation. According to your survey, I wouldn't like to add deprecated class into `__init__.py`. So, current direction is fine. Let me resolve this conversation. Thank you so much.

HideakiImamura (approved):
LGTM!

Issue #4850, opened 2023-08-01
Add Log Scale Option for Objective Value in optuna.visualization Plots


corentinlger:
### Motivation

Add an option to optuna.visualization functions to enable log_scale plotting of objective values.

### Description

Currently, when using optuna.visualization to plot optimization results, it is possible to use log-scaled axes for parameters suggested by Optuna using the log=True option. However, there is no built-in option to apply the same log scale for the objective function values. This can lead to difficulties in visualizing results when the objective values span a wide range, making it challenging to distinguish values that are close to zero.

### Feature Request:

I propose adding a new argument (let's call it "objective_log_scale" for example) to the plotting functions in optuna.visualization. This new argument would allow users to choose whether to display the objective function values using a log scale or not. By enabling the log scale option for the objective value, users can gain a more insightful understanding of the optimization process, especially when dealing with objective values that have a wide range of values.

### Example Use Case:

Consider a scenario where the objective values range from approximately 10¹ to 10⁻⁴. In the current implementation, it becomes challenging to distinguish the values that are close to zero in the plots. However, when using a custom plot with a log scale applied to the objective values, users can see the differences even with a lr going from 0.1 to 1. This can provide valuable insights into the optimization process.

Here is an illustration with an optuna slice_plot and another custom plot (the values are not exactly the same but come from a random hyper parameter search on the same problem).

![image](https://github.com/optuna/optuna/assets/111868204/defd6768-04c1-49fc-b02b-f08e034f0daa)

![image](https://github.com/optuna/optuna/assets/111868204/600023c1-132a-451e-b55c-0a1a4f13668b)

Do you think it would be useful to add this kind of feature to the visualization methods ?

nzw0301:
Since Optuna's visualisation methods return an object that users can edit, so changing the scale can be done easily.

corentinlger:
Ok thanks for your response !

Pablo1990:
> Personally, I don't think this is necessary. Since Optuna's visualisation methods return an object that users can edit, changing the scale can be done easily without introducing such a new argument, which makes the code and test more complicated.

Hi! I was trying to figure this out, and it was not as easy for me. So, I am providing the code to do it in case someone needs it (e.g.):
```
fig = optuna.visualization.plot_optimization_history(study)
fig.update_yaxes(type="log")
```

FrankwaP:
Just in case someone wants an example with the matplotlib back-end:

```python
# case where an Axes
ax = plot_optimization_history(study)
[...]

PR #5749, opened 2024-11-06
Adding a `log_scale` option for `plot_contour`.

FrankwaP:
## Motivation
As discussed in [4850](https://github.com/optuna/optuna/issues/4850#issuecomment-2444757889), using a log scale for contour plots is not straight forward. So I added a `log_scale` option for both plotly and matplotlib backends. 

For plotly, I've followed the recommendations found [on their forum](https://community.plotly.com/t/how-to-set-log-scale-for-z-axis-on-a-heatmap/292/8) which plot the log-transformed value then makes use of a custom [ColorBar](https://plotly.github.io/plotly.py-docs/generated/plotly.graph_objects.contour.html#plotly.graph_objects.contour.ColorBar) (defined in a dictionary form) to display the original values.

For matplotlib it was quite easy since they propose a [LogLocator](https://matplotlib.org/stable/api/ticker_api.html#matplotlib.ticker.LogLocator).

### DIFF
--- a/optuna/visualization/_contour.py
+++ b/optuna/visualization/_contour.py
@@ -56,6 +56,7 @@ class _ContourInfo(NamedTuple):
     sub_plot_infos: list[list[_SubContourInfo]]
     reverse_scale: bool
     target_name: str
+    log_scale: bool
 
 
 class _PlotValues(NamedTuple):
@@ -257,6 +258,7 @@ def _get_contour_info(
     params: list[str] | None = None,
     target: Callable[[FrozenTrial], float] | None = None,
     target_name: str = "Objective Value",
+    log_scale: bool = False,
 ) -> _ContourInfo:
     _check_plot_args(study, target, target_name)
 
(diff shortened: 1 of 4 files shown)


FrankwaP:
OK… so the `pre-commit` hooks do not analyze the tests…

not522:
Thank you for your PR, but how about `plot_contour(study, target=lambda trial: np.log10(trial.value))`? I'd like to keep the number of arguments to a minimum if possible.

FrankwaP:
In this case the color bar is less informative since it uses a linear scale of the log 10 values, instead of a log10 scale of the values.

not522:
Thanks. I discussed this PR with other maintainers, and we concluded that [OptunaHub](https://hub.optuna.org/) is suitable for this feature. It gathers customized or specialized features.
Would you please make a pull request introducing a package that implements the feature to [optunahub-registry](https://github.com/optuna/optunahub-registry)?

FrankwaP:
Thank you so much for this detailed answer!

PR #5768, opened 2024-11-15
Update requirements for OptunaHub in README

y0z:
## Motivation & Description of the changes

Add required dependencies for AutoSampler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,7 @@ You can use the registered features and publish your packages.
 
 ```shell
 $ pip install optunahub
+$ pip install cmaes scipy torch # install AutoSampler dependencies
 ```
 
 You can load registered module with `optunahub.load_module`.
--- a/README.md
+++ b/README.md
@@ -206,7 +206,8 @@ You can use the registered features and publish your packages.
 
 ```shell
 $ pip install optunahub
-$ pip install cmaes scipy torch # install AutoSampler dependencies
+# Install AutoSampler dependencies (CPU only is sufficient for PyTorch)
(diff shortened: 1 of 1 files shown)


y0z:
@porink0424 @kAIto47802 Could you review this PR?

README.md
```diff
@@ -206,6 +206,7 @@ You can use the registered features and publish your packages.
 
 ```shell
 $ pip install optunahub
+$ pip install cmaes scipy torch # install AutoSampler dependencies
```
nabenabe0928, line 209:
```suggestion
# Install AutoSampler dependencies (CPU only is sufficient for PyTorch)
$ pip install cmaes scipy torch --extra-index-url https://download.pytorch.org/whl/cpu
```

nabenabe0928 (commented):
I think it would be more friendly if the example installs the CPU-only version.

README.md
y0z: Applied.

nabenabe0928 (approved):
LGTM!

kAIto47802:
As stated in [the pip documentation](https://pip.pypa.io/en/stable/cli/pip_install/#finding-packages), there is no prioritization between PyPI and locations added with `--extra-index-url`. This might makes it non-obvious that the CPU version is prioritized over the default GPU version when installed on Linux. However, experimental confirmation shows that the CPU version is indeed installed in such cases.

kAIto47802 (approved):
LGTM

porink0424 (approved):
LGTM!

PR #5778, opened 2024-11-20
Fix ticks in Matplotlib contour plot

### LINKED ISSUE #5774: Axis tick scale in `plot_contour` with Matplotlib backend
### Expected behavior

When I plot with `optuna.visualization.matplotlib.plot_contour`, the horizontal and vertical ticks are set for each subfigure.

The subfigures on the diagonal are empty (as it should be), but the ticks are always scaled linearly. This is inconsistent with the off-diagonal subfigures in the corresponding columns/rows, where the axis ticks are logarithmic for logarithmic parameters.

This is only a problem (annoyance) in the first row and the last column, since the tick labels (which are rendered on the left and the bottom of the figure) do not correspond to the actual subfigures they should be describing.

### Environment

- Optuna version: 4.0.0
- Python version: 3.12.7
- OS: Linux-6.11.7-arch1-1-x86_64-with-glibc2.40
    - Matplotlib: 3.9.2

```shell
-
```

### Steps to reproduce

1. Create a study with at least 3 parameters that has logarithmic scale
2. Load and plot:

```python
import optuna
study = optuna.load_study(...)
import optuna.visualization.matplotlib as vis
vis.plot_contour(study)
plt.show()
```

This bug does not appear with the Plotly backend.

sulan:
## Motivation
The diagonal subfigures in the contour plot visualisation had wrong axis ticks when the parameter was log scale float or when it was categorical.

Fixes #5774.

## Description of the changes

In the `optuna.visualization.matplotlib._contour._generate_contour_subplot`, when there is nothing to plot, i.e., when it is a subplot on the diagonal, there is an early return. Axis tick settings (`set_scale` for numerical and manual positioning for categorical parameters) were moved before this eary return.

### DIFF
--- a/optuna/visualization/matplotlib/_contour.py
+++ b/optuna/visualization/matplotlib/_contour.py
@@ -153,6 +153,43 @@ class _LabelEncoder:
         return list(range(len(self.labels)))
 
 
+def _filter_missing_values(
+    xaxis: _AxisInfo, yaxis: _AxisInfo
+) -> tuple[list[str | float], list[str | float]]:
+    x_values = []
+    y_values = []
+    for x_value, y_value in zip(xaxis.values, yaxis.values):
+        if x_value is not None and y_value is not None:
+            x_values.append(x_value)
+            y_values.append(y_value)
+    return x_values, y_values
+
+
+def _calculate_axis_data(
+    axis: _AxisInfo,
+    values: Sequence[str | float],
+) -> tuple[np.ndarray, list[str], list[int], list[int | float]]:
+    # Convert categorical values to int.
+    cat_param_labels: list[str] = []
+    cat_param_pos: list[int] = []
+    returned_values: Sequence[int | float]
+    if axis.is_cat:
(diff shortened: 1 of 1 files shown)


y0z:

y0z (approved):
I confirmed that the change works fine.

LGTM.

not522 (approved):
LGTM. I checked this PR using the following example.

```python
import optuna
import optuna.visualization.matplotlib as vis

def f(trial):
    trial.suggest_float('a', 1, 10, log=True)
    trial.suggest_float('b', 100, 1000, log=True)
    trial.suggest_float('c', 1000, 10_000, log=True)
    return 1.0

def g(trial):
    trial.suggest_categorical('a', ["a", "b", "c"])
    trial.suggest_categorical('b', ["x", "y", "z"])
    trial.suggest_categorical('c', ["s", "t", "u"])
    return 1.0

study = optuna.create_study()
study.optimize(f, n_trials=100)
vis.plot_contour(study)
plt.show()
```

Issue #5797, opened 2024-11-28
Fixes a fragile unit test, `test_check_distribution_suggest_loguniform[sqlite3]`


c-bata:
### Motivation

In https://github.com/optuna/optuna/actions/runs/12058983729/job/33626763864?pr=5795, `test_check_distribution_suggest_loguniform[sqlite]` was once failed and succeeded at the second attempt.

### Suggestion

Fix the `test_check_distribution_suggest_loguniform` test function to make unit tests non-fragile.

sawa3030:
Hello @c-bata, I think it might be more robust to avoid counting warnings in the unit test. Just to check if there is warning comes up would be enough, I think. This is because diagnosing the root cause of such failures can be time-consuming and challenging.

not522:
The following CI run encountered the error. https://github.com/optuna/optuna/actions/runs/18447504761/job/52556302035

```
=================================== FAILURES ===================================
____________ test_check_distribution_suggest_uniform[journal_redis] ____________
[gw5] linux -- Python 3.13.7 /opt/hostedtoolcache/Python/3.13.7/x64/bin/python

storage_mode = 'journal_redis'

    @pytest.mark.filterwarnings("ignore::FutureWarning")
    @pytest.mark.parametrize("storage_mode", STORAGE_MODES)
    def test_check_distribution_suggest_uniform(storage_mode: str) -> None:
        sampler = samplers.RandomSampler()
        with StorageSupplier(storage_mode) as storage:
            study = create_study(storage=storage, sampler=sampler)
            trial = study.ask()
    
            with pytest.warns() as record:
                trial.suggest_uniform("x", 10, 20)
                trial.suggest_uniform("x", 10, 20)
                trial.suggest_uniform("x", 10, 30)
    
            # we expect exactly one warning (not counting ones caused by deprecation)
            warning_records = [r for r in record if r.category is not FutureWarning]
>           assert len(warning_records) == 1, [warning.message for warning in warning_records]
E           AssertionError: [ResourceWarning('unclosed database in <sqlite3.Connection object at 0x7f1cb6e15b70>'), ResourceWarning('unclosed data... call and ignores all following. Using these values: {\'step\': None, \'low\': 10.0, \'high\': 20.0, \'log\': False}')]
E           assert 3 == 1
[...]

PR #5800, opened 2024-11-28
Make the assertion messages in `test_trial.py` readable

not522:
## Motivation

ref: #5797
To investigate the cause of the fragile tests, this PR makes the assertion messages readable.

## Description of the changes

* Make the assertion messages in `test_trial.py` readable

### DIFF
--- a/tests/trial_tests/test_trial.py
+++ b/tests/trial_tests/test_trial.py
@@ -79,8 +79,8 @@ def test_check_distribution_suggest_uniform(storage_mode: str) -> None:
             trial.suggest_uniform("x", 10, 30)
 
         # we expect exactly one warning (not counting ones caused by deprecation)
-        warnings = [r for r in record if r.category is not FutureWarning]
-        assert len(warnings) == 1, [warning.message for warning in warnings]
+        warning_records = [r for r in record if r.category is not FutureWarning]
+        assert len(warning_records) == 1, [warning.message for warning in warning_records]
 
         with pytest.raises(ValueError):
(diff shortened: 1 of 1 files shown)


c-bata:

porink0424 (approved):
LGTM!
I confirmed that multiple warning messages are properly displayed in a readable array format.

tests/trial_tests/test_trial.py
```diff
@@ -79,7 +79,8 @@ def test_check_distribution_suggest_uniform(storage_mode: str) -> None:
             trial.suggest_uniform("x", 10, 30)
 
         # we expect exactly one warning (not counting ones caused by deprecation)
-        assert len([r for r in record if r.category is not FutureWarning]) == 1
+        warnings = [r for r in record if r.category is not FutureWarning]
```
c-bata, line 82: `test_trial.py` imports `warnings` package at L12. How about renaming the variable like below?

```suggestion
        warning_msgs = [r for r in record if r.category is not FutureWarning]
```

c-bata (changes requested):
Thank you for your pull request. I left one suggestion.

tests/trial_tests/test_trial.py
not522: How about `warning_records`? It is not the message itself.

c-bata: Sounds better to me 👍

not522:
I renamed `warnings` to `warning_records`. PTAL.

c-bata (approved):
LGTM.

PR #5813, opened 2024-12-01
Refactor the MOTPE split

nabenabe0928:
## Motivation

This PR is motivated by some contributions to OptunaHub.
As we may get some TPE-related modules there, I wanted to refactor some functions to make functions for MOTPE more reusable.

## Description of the changes

I refactored `_split_complete_trials_multi_objective` and `_calculate_weights_below_for_multi_objective`.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -683,7 +683,7 @@ def _split_complete_trials_multi_objective(
         )
         indices_below = np.append(indices_below, selected_indices)
 
-    below_indices_set = set(indices_below.tolist())
+    below_indices_set = set(cast(list[int], indices_below.tolist()))
     below_trials = [trials[i] for i in range(len(trials)) if i in below_indices_set]
     above_trials = [trials[i] for i in range(len(trials)) if i not in below_indices_set]
     return below_trials, above_trials
@@ -739,7 +739,10 @@ def _calculate_weights_below_for_multi_objective(
     if constraints_func is None:
         is_feasible = np.ones(len(below_trials), dtype=bool)
     else:
-        is_feasible = np.asarray([all(c <= 0 for c in constraints_func(t)) for t in below_trials])
+        is_feasible = cast(
+            np.ndarray[tuple[int], np.dtype[np.bool_]],
+            np.asarray([all(c <= 0 for c in constraints_func(t)) for t in below_trials]),
+        )
 
     n_below_feasible = np.count_nonzero(is_feasible)
     # For now, EPS weight is assigned to unpromising (e.g., infeasible) trials.
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -668,8 +668,9 @@ def _split_complete_trials_multi_objective(
     lvals = np.array([trial.values for trial in trials])
     lvals *= np.array([-1.0 if d == StudyDirection.MAXIMIZE else 1.0 for d in study.directions])
     nondomination_ranks = _fast_non_domination_rank(lvals, n_below=n_below)
-    _, rank_counts = np.unique(nondomination_ranks, return_counts=True)
+    ranks, rank_counts = np.unique(nondomination_ranks, return_counts=True)
     last_rank_before_tiebreak = np.searchsorted(np.cumsum(rank_counts), n_below, side="right") - 1
+    assert all(ranks[: last_rank_before_tiebreak + 1] == np.arange(last_rank_before_tiebreak + 1))
     indices = np.arange(len(trials))
     indices_below = indices[nondomination_ranks <= last_rank_before_tiebreak]
 
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -666,9 +666,8 @@ def _split_complete_trials_multi_objective(
     lvals = np.array([trial.values for trial in trials])
     lvals *= np.array([-1.0 if d == StudyDirection.MAXIMIZE else 1.0 for d in study.directions])
     nondomination_ranks = _fast_non_domination_rank(lvals, n_below=n_below)
-    sorted_ranks, counts_of_each_rank = np.unique(nondomination_ranks, return_counts=True)
-    before_tiebreak = np.cumsum(counts_of_each_rank) <= n_below
-    last_rank_before_tiebreak = sorted_ranks[before_tiebreak][-1] if any(before_tiebreak) else -1
+    _, rank_counts = np.unique(nondomination_ranks, return_counts=True)
+    last_rank_before_tiebreak = np.searchsorted(np.cumsum(rank_counts), n_below, side="right") - 1
     indices = np.arange(len(trials))
     indices_below = indices[nondomination_ranks <= last_rank_before_tiebreak]
 
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -27,6 +27,7 @@ from optuna.search_space import IntersectionSearchSpace
 from optuna.search_space.group_decomposed import _GroupDecomposedSearchSpace
 from optuna.search_space.group_decomposed import _SearchSpaceGroup
 from optuna.study._multi_objective import _fast_non_domination_rank
+from optuna.study._multi_objective import _is_pareto_front
 from optuna.study._study_direction import StudyDirection
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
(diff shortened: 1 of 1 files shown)


optuna/samplers/_tpe/sampler.py
```diff
@@ -742,52 +734,34 @@ def _calculate_weights_below_for_multi_objective(
...
-        indices_mat = ~np.eye(n_below).astype(bool)
-        contributions = np.asarray(
-            [
-                hv - compute_hypervolume(lvals[indices_mat[i]], reference_point)
-                for i in range(n_below)
-            ]
-        )
-        contributions[np.isnan(contributions)] = np.inf
-        max_contribution = np.maximum(np.max(contributions), EPS)
-        if not np.isfinite(max_contribution):
```
nabenabe0928, line 783: From the monotonicity of the hypervolume (submodularity), `hv` is always larger than the leave-one-out hypervolume, i.e. `compute_hypervolume(lvals[indices_mat[i]], reference_point)`, meaning that each element of `contributions` becomes `inf` only if `hv` is `inf`.

Since we also define the line `contributions[np.isnan(contributions)] = np.inf`, which takes care of `inf - inf = inf` in the hypervolume contribution calculation, if any element in `contributions` takes `inf`, all the elements in `contributions` become `inf`.

That is why we can separate this exception handling as the processing only for `np.isinf(hv) == True`.

optuna/samplers/_tpe/sampler.py
```diff
@@ -742,52 +734,34 @@ def _calculate_weights_below_for_multi_objective(
...
-    return weights_below_all
+    lvals = np.asarray([t.values for t in below_trials])[is_feasible]
+    lvals *= np.array([-1.0 if d == StudyDirection.MAXIMIZE else 1.0 for d in study.directions])
+    ref_point = _get_reference_point(lvals)
+    on_front = _is_pareto_front(lvals, assume_unique_lexsorted=False)
+    hv = compute_hypervolume(lvals[on_front], ref_point, assume_pareto=True)
+    if np.isinf(hv) or n_below_feasible == 1:
+        weights_below[is_feasible] = 1.0
+        # TODO(nabenabe): Assign EPS to non-Pareto solutions, and
+        # solutions with finite contrib if hv is inf.
```
nabenabe0928: This PR is refactoring, so I will not do it in this PR, but in principle, the current calculation of the hypervolume contribution is actually incorrect when `np.isinf(hv) == True`.

There are several cases to consider:
1. non-Pareto solutions must get zero weight, but we now get 1.0 for non-Pareto solutions
2. duplicated solutions should get zero weight (at least for now, but it might change in the future), but we get 1.0 for duplicated solutions,
3. solutions with `np.prod(ref_point - values) != inf` should get 0.0, but we get 1.0 for such solutions.

optuna/samplers/_tpe/sampler.py
```diff
@@ -742,52 +734,34 @@ def _calculate_weights_below_for_multi_objective(
...
-    return weights_below_all
+    lvals = np.asarray([t.values for t in below_trials])[is_feasible]
+    lvals *= np.array([-1.0 if d == StudyDirection.MAXIMIZE else 1.0 for d in study.directions])
+    ref_point = _get_reference_point(lvals)
+    on_front = _is_pareto_front(lvals, assume_unique_lexsorted=False)
+    hv = compute_hypervolume(lvals[on_front], ref_point, assume_pareto=True)
+    if np.isinf(hv) or n_below_feasible == 1:
+        weights_below[is_feasible] = 1.0
+        # TODO(nabenabe): Assign EPS to non-Pareto solutions, and
+        # solutions with finite contrib if hv is inf. Ref: PR#5813.
```
nabenabe0928, line 756: This PR is refactoring, so I will not do it in this PR, but in principle, the current calculation of the hypervolume contribution is actually incorrect when `np.isinf(hv) == True`.

There are several cases to consider:
1. non-Pareto solutions must get zero weight, but we now get 1.0 for non-Pareto solutions
2. duplicated solutions should get zero weight (at least for now, but it might change in the future), but we get 1.0 for duplicated solutions,
3. solutions with `np.prod(ref_point - values) != inf` should get 0.0, but we get 1.0 for such solutions.

optuna/samplers/_tpe/sampler.py
```diff
@@ -652,46 +660,29 @@ def _split_complete_trials_multi_objective(
...
-
-    below_trials = []
-    above_trials = []
-    for index in range(len(trials)):
-        if index in indices_below:
-            below_trials.append(trials[index])
-        else:
-            above_trials.append(trials[index])
+    _, rank_counts = np.unique(nondomination_ranks, return_counts=True)
+    last_rank_before_tiebreak = np.searchsorted(np.cumsum(rank_counts), n_below, side="right") - 1
```
nabenabe0928, line 670: if `crc = np.cumsum(rank_counts)`, `crc[i]` is the number of solutions that achieve the non-domination rank of `i` or lower.
`last_rank_before_tiebreak = np.searchsorted(crc, n_below, side="right") - 1` is the integer such that `crc[last_rank_before_tiebreak] <= n_below < crc[last_rank_before_tiebreak]`.
Note that if `last_rank_before_tiebreak == -1`, here we assume that `crc[-1] = 0` instead of the access to the last element of `crc`.

not522:

optuna/samplers/_tpe/sampler.py
```diff
@@ -742,52 +736,34 @@ def _calculate_weights_below_for_multi_objective(
...
+        return weights_below
+
+    contribs = np.zeros(n_below_feasible, dtype=float)
+    # NOTE(nabenabe): Hypervolume contribution of a non-Pareto solution is zero.
+    pareto_indices = ~np.eye(n_below_feasible, dtype=bool)[on_front]
+    contribs[on_front] = hv - np.asarray(
+        [compute_hypervolume(lvals[indices], ref_point) for indices in pareto_indices]
+    )
+    # Calculate weights based on hypervolume contributions.
+    weights_below[is_feasible] = np.maximum(contribs / max(np.max(contribs), EPS), EPS)
```
y0z, line 768: Please confirm this line.
The following might be the correct implementation:
```suggestion
    weights_below[is_feasible] = np.maximum(contribs / hv), EPS)
```

nabenabe0928: Let $H(S, r)$ be a hypervolume of a set $S$ given a reference point $r$.
Then did you mean each weight $w$ should be $1 - \frac{H(S \setminus \\{s_i\\}, r)}{H(S, r)}$ rather than $\frac{H(S \setminus \\{s_i\\}, r)}{H(S, r) - \min_{i \in [N]} H(S \setminus \\{s_i\\}, r)}$?

nabenabe0928:

nabenabe0928:
@y0z 
I made this PR much shorter and separated one of the refactorings to another PR, so PTAL!
- https://github.com/optuna/optuna/pull/5871

kAIto47802 (changes requested):
Thank you for your PR! Basically LGTM, but I left a minor comment :octocat:

optuna/samplers/_tpe/sampler.py
```diff
@@ -652,46 +659,33 @@ def _split_complete_trials_multi_objective(
...
+        assert ranks[last_rank_before_tiebreak + 1] == last_rank_before_tiebreak + 1
+        need_tiebreak = nondomination_ranks == last_rank_before_tiebreak + 1
+        rank_i_lvals = lvals[need_tiebreak]
+        subset_size = n_below - indices_below.size
+        selected_indices = _solve_hssp(
+            rank_i_lvals, indices[need_tiebreak], subset_size, _get_reference_point(rank_i_lvals)
+        )
+        indices_below = np.append(indices_below, selected_indices)
+
+    below_indices_set = set(cast(list, indices_below.tolist()))
```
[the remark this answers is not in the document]
nabenabe0928: I addressed your comment!

kAIto47802 (approved):

optuna/samplers/_tpe/sampler.py
```diff
@@ -652,46 +659,33 @@ def _split_complete_trials_multi_objective(
...
+        assert ranks[last_rank_before_tiebreak + 1] == last_rank_before_tiebreak + 1
+        need_tiebreak = nondomination_ranks == last_rank_before_tiebreak + 1
+        rank_i_lvals = lvals[need_tiebreak]
+        subset_size = n_below - indices_below.size
+        selected_indices = _solve_hssp(
+            rank_i_lvals, indices[need_tiebreak], subset_size, _get_reference_point(rank_i_lvals)
+        )
+        indices_below = np.append(indices_below, selected_indices)
+
+    below_indices_set = set(indices_below.tolist())
```
y0z, line 686:
```suggestion
    below_indices_set = set(cast(list, indices_below.tolist()))
```

y0z (approved):

PR #5828, opened 2024-12-05
Adding check in `cli.py` to handle an empty database

### LINKED ISSUE #5816: CLI for empty DB raises `ValueError`
### Expected behavior

CLI for empty DB should output empty result, but the current implementation raises `ValueError`.

### Environment

- Optuna version:4.2.0.dev
- Python version:3.13.0
- OS:macOS-15.1-x86_64-i386-64bit-Mach-O

```shell
See below.
```

### Steps to reproduce

For empty DB (`tmp.db` does not exist before the command), the `optuna studies` command raises `ValueError`.

```bash
$ optuna --storage sqlite:///tmp.db studies
  File "/Users/naotomizuno/.pyenv/versions/optuna-3.13.0/bin/optuna", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/naotomizuno/optuna/optuna/cli.py", line 991, in main
    return args.handler(args)
           ~~~~~~~~~~~~^^^^^^
  File "/Users/naotomizuno/optuna/optuna/cli.py", line 463, in take_action
    _format_output(
    ~~~~~~~~~~~~~~^
        records, self._study_list_header, parsed_args.format, parsed_args.flatten
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/naotomizuno/optuna/optuna/cli.py", line 258, in _format_output
    return _dump_table(values, header).strip()
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/Users/naotomizuno/optuna/optuna/cli.py", line 222, in _dump_table
    max_width = max(len(header[column]), max(row[column].width() for row in rows))
                                         ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: max() iterable argument is empty
```

_No response_

willdavidson05:
## Motivation

This PR aims to close #5816 

## Description of the changes

- Added a check in `cli.py` in `_dump_table` to handle cases where no rows exist (empty database).
- If `rows` is empty, the column width is determined based only on the header length, so that `max()` isn't called on an empty sequence.
- Also created a conditional block to append row strings (`ret += "\n".join(rows_string) + "\n"`) only if rows exist, to avoid unnecessary empty lines in the output.

### DIFF
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -199,6 +199,7 @@ def _dump_value(records: list[dict[str, Any]], header: list[str]) -> str:
 
 
 def _dump_table(records: list[dict[str, Any]], header: list[str]) -> str:
+
     rows = []
     for record in records:
         row = []
@@ -215,7 +216,10 @@ def _dump_table(records: list[dict[str, Any]], header: list[str]) -> str:
         for t in value_types:
             if t == ValueType.STRING:
                 value_type = ValueType.STRING
-        max_width = max(len(header[column]), max(row[column].width() for row in rows))
+        if not rows:
+            max_width = len(header[column])
+        else:
+            max_width = max(len(header[column]), max(row[column].width() for row in rows))
         separator += "-" * (max_width + 2) + "+"
         if value_type == ValueType.NUMERIC:
             header_string += f" {header[column]:>{max_width}} |"
@@ -228,7 +232,8 @@ def _dump_table(records: list[dict[str, Any]], header: list[str]) -> str:
     ret += separator + "\n"
     ret += header_string + "\n"
     ret += separator + "\n"
-    ret += "\n".join(rows_string) + "\n"
+    if rows:
+        ret += "\n".join(rows_string) + "\n"
     ret += separator + "\n"
 
     return ret
--- a/optuna/cli.py
+++ b/optuna/cli.py
@@ -231,8 +231,8 @@ def _dump_table(records: list[dict[str, Any]], header: list[str]) -> str:
     ret += separator + "\n"
     ret += header_string + "\n"
     ret += separator + "\n"
-    if len(rows) == 0:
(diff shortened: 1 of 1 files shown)


not522:
@gen740 @HideakiImamura Could you review this PR?

optuna/cli.py
```diff
@@ -215,7 +215,10 @@ def _dump_table(records: list[dict[str, Any]], header: list[str]) -> str:
         for t in value_types:
             if t == ValueType.STRING:
                 value_type = ValueType.STRING
-        max_width = max(len(header[column]), max(row[column].width() for row in rows))
+        if not rows:
```
gen740, line 218:
```suggestion
        if len(rows) == 0:
```
Please use `len(rows) == 0` for checking if the list is empty, rather than relying on `not rows`.

optuna/cli.py
```diff
@@ -228,7 +231,8 @@ def _dump_table(records: list[dict[str, Any]], header: list[str]) -> str:
     ret += separator + "\n"
     ret += header_string + "\n"
     ret += separator + "\n"
-    ret += "\n".join(rows_string) + "\n"
+    if rows:
+        ret += "\n".join(rows_string) + "\n"
```
gen740, line 235:
```suggestion
    for row_string in rows_string:
        ret += row_string + "\n"
```
It might be more straightforward to simply iterate over rows_string.

gen740 (commented):
Thank you for your PR!

I've left a few comments.

willdavidson05:
@gen740 Thank you for the feedback, I requested another review!

gen740 (approved):
LGTM!

HideakiImamura (approved):

Issue #5806, opened 2024-11-29
Make error message for `create_study`'s `direction` easier to understand


not522:
### Motivation

Users sometimes make the mistake of passing the list of directions to the `direction` argument. The error message for this mistake is hard to understand what is the problem.

* Correct

```python
optuna.create_study(directions=["minimize", "minimize"])
```

* Wrong

Using `direction` instead of `directions`

```python
optuna.create_study(direction=["minimize", "minimize"])
```

```
ValueError: Please set either 'minimize' or 'maximize' to direction. You can also set the corresponding `StudyDirection` member.
```

### Suggestion

We can make the error message more helpful for the mistake. It can be something like "for multi-objective optimization, using directions instead of direction".

not522:
Thanks! It's `study/study.py`.

https://github.com/optuna/optuna/blob/b29fdff87e46fd1c1e025e23af005c9a9ad80a11/optuna/study/study.py#L1254-L1257

PR #5839, opened 2024-12-07
Made error message easier to understand for create_study's direction #5806

unKnownNG:
## Motivation

Users sometimes pass a list of directions to the direction argument, expecting multi-objective optimization, when they should use directions instead. 
The previous error message: 
`ValueError: Please set either 'minimize' or 'maximize' to direction. You can also set the corresponding 'StudyDirection' member.`

This message doesn't clearly explain that multi-objective optimization requires using directions. The updated message provides clearer guidance. The issue was raised in the #5806

## Description of the changes

The error message is changed to `Use 'directions' instead of 'direction' for multi-objective optimization.`

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1252,8 +1252,7 @@ def create_study(
         for d in directions
     ):
         raise ValueError(
-            "Please set either 'minimize' or 'maximize' to direction. You can also set the "
-            "corresponding `StudyDirection` member."
+            "Use 'directions' instead of 'direction' for multi-objective optimization."
         )
 
     direction_objects = [


HideakiImamura:

y0z (commented):
I think it will be mergeable once https://github.com/optuna/optuna/pull/5839#issuecomment-2527087926 have been addressed.

nabenabe0928:
I think the original comment should stay as is, but what we should check is whether the typing of `direction` is correct, meaning that we should check whether `direction` is a list or not.
If `direction` is a list, we should raise the error suggested by @unKnownNG .

What do you, @y0z @not522 , think?

PR #5843, opened 2024-12-08
Fix logger.warn() deprecation warning in GP module

iamarunbrahma:
## Motivation
The code was using the deprecated `logger.warn()` method which triggers deprecation warnings. According to Python's logging documentation, `warn()` is deprecated in favor of `warning()`. This PR fixes the deprecation warning to ensure clean logs and follow Python best practices.

## Description of the changes
- Replace deprecated `logger.warn()` with `logger.warning()` in `optuna/_gp/gp.py`
- The change is a straightforward method name update with no functional changes to the logging behavior

### DIFF
--- a/optuna/_gp/gp.py
+++ b/optuna/_gp/gp.py
@@ -257,7 +257,7 @@ def fit_kernel_params(
         except RuntimeError as e:
             error = e
 
-    logger.warn(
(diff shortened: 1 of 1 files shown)


not522 (approved):
Thank you! LGTM!

PR #5853, opened 2024-12-10
Refactor the fix for MyPy errors due to NumPy v2.2.0

nabenabe0928:
## Motivation

This PR is to fix misleading typing to hotfix CI checking.
- https://github.com/optuna/optuna/pull/5848

## Description of the changes

- Eliminate typings that do not align with the actual behaviors

> [!NOTE]
> These changes can probably be removed after the following issue is resolved. 
> https://github.com/numpy/numpy/issues/27957

### DIFF
--- a/optuna/_gp/optim_mixed.py
+++ b/optuna/_gp/optim_mixed.py
@@ -208,7 +208,8 @@ def local_search_mixed(
     # TODO(kAIto47802): Think of a better way to handle this.
     lengthscales = 1 / np.sqrt(inverse_squared_lengthscales[continuous_indices])
 
-    discrete_indices = np.where(steps > 0)[0]
+    # NOTE(nabenabe): MyPy Redefinition for NumPy v2.2.0. (Cast signed int to int)
+    discrete_indices = np.where(steps > 0)[0].astype(int)
     choices_of_discrete_params = [
         (
             np.arange(bounds[i, 1])
@@ -251,10 +252,7 @@ def local_search_mixed(
         if updated:
             last_changed_param = CONTINUOUS
 
-        for ind, choices, xtol in zip(
-            discrete_indices, choices_of_discrete_params, discrete_xtols
-        ):
-            i = int(ind)
+        for i, choices, xtol in zip(discrete_indices, choices_of_discrete_params, discrete_xtols):
             if last_changed_param == i:
                 # Parameters not changed since last time.
                 return (best_normalized_params, best_fval)
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from typing import Any
-
 import numpy as np
 
 import optuna
@@ -94,19 +92,18 @@ def _solve_hssp_on_unique_loss_vals(
 
     # The following logic can be used for non-unique rank_i_loss_vals as well.
     diff_of_loss_vals_and_ref_point = reference_point - rank_i_loss_vals
-    assert subset_size <= rank_i_indices.size
-    n_objectives = reference_point.size
+    (n_solutions, n_objectives) = rank_i_loss_vals.shape
+    assert isinstance(n_solutions, int), "MyPy Redefinition for NumPy v2.2.0."
     contribs = np.prod(diff_of_loss_vals_and_ref_point, axis=-1)
     selected_indices = np.zeros(subset_size, dtype=int)
     selected_vecs = np.empty((subset_size, n_objectives))
-    indices: np.ndarray[tuple[int, ...], np.dtype[Any]] = np.arange(
-        rank_i_loss_vals.shape[0], dtype=int
-    )
+    indices = np.arange(n_solutions, dtype=int)
     for k in range(subset_size):
         max_index = int(np.argmax(contribs))
         selected_indices[k] = indices[max_index]
         selected_vecs[k] = rank_i_loss_vals[max_index].copy()
-        keep = np.ones(contribs.size, dtype=bool)
+        assert n_solutions - k > 0
+        keep = np.ones(n_solutions - k, dtype=bool)
         keep[max_index] = False
         contribs = contribs[keep]
         indices = indices[keep]
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -93,7 +93,7 @@ def _solve_hssp_on_unique_loss_vals(
     # The following logic can be used for non-unique rank_i_loss_vals as well.
     diff_of_loss_vals_and_ref_point = reference_point - rank_i_loss_vals
     (n_solutions, n_objectives) = rank_i_loss_vals.shape
-    assert isinstance(n_solutions, int), "MyPy Redefinition for NumPy v2.2.0."
+    n_solutions = int(n_solutions)  # MyPy Redefinition for NumPy v2.2.0.
     contribs = np.prod(diff_of_loss_vals_and_ref_point, axis=-1)
     selected_indices = np.zeros(subset_size, dtype=int)
     selected_vecs = np.empty((subset_size, n_objectives))
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -90,10 +90,10 @@ def _solve_hssp_on_unique_loss_vals(
     if rank_i_loss_vals.shape[-1] == 2:
         return _solve_hssp_2d(rank_i_loss_vals, rank_i_indices, subset_size, reference_point)
 
+    assert subset_size < rank_i_indices.size
     # The following logic can be used for non-unique rank_i_loss_vals as well.
     diff_of_loss_vals_and_ref_point = reference_point - rank_i_loss_vals
     (n_solutions, n_objectives) = rank_i_loss_vals.shape
-    n_solutions = int(n_solutions)  # MyPy Redefinition for NumPy v2.2.0.
     contribs = np.prod(diff_of_loss_vals_and_ref_point, axis=-1)
     selected_indices = np.zeros(subset_size, dtype=int)
     selected_vecs = np.empty((subset_size, n_objectives))
@@ -102,8 +102,7 @@ def _solve_hssp_on_unique_loss_vals(
         max_index = int(np.argmax(contribs))
         selected_indices[k] = indices[max_index]
         selected_vecs[k] = rank_i_loss_vals[max_index].copy()
-        assert n_solutions - k > 0
-        keep = np.ones(n_solutions - k, dtype=bool)
+        keep = np.ones(contribs.size, dtype=bool)
         keep[max_index] = False
         contribs = contribs[keep]
         indices = indices[keep]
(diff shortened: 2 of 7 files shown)


optuna/samplers/_nsgaiii/_elite_population_selection_strategy.py
```diff
@@ -103,7 +103,7 @@ def __call__(self, study: Study, population: list[FrozenTrial]) -> list[FrozenTr
 
 def _generate_default_reference_point(
     n_objectives: int, dividing_parameter: int = 3
-) -> np.ndarray[tuple[int, int], np.dtype[np.float64]]:
```
nabenabe0928, line 106: Technically speaking, this typing is stricter and correct, but for consistency, I removed this typing.

optuna/samplers/_gp/sampler.py
```diff
@@ -329,14 +329,15 @@ def _warn_and_convert_inf(
 
 def _get_constraint_vals_and_feasibility(
     study: Study, trials: list[FrozenTrial]
-) -> tuple[np.ndarray, np.bool | np.ndarray]:
```
nabenabe0928, line 332: The second return is definitely not boolean, so I removed this to avoid any confusion.

optuna/_hypervolume/hssp.py
```diff
@@ -94,19 +92,18 @@ def _solve_hssp_on_unique_loss_vals(
...
     # The following logic can be used for non-unique rank_i_loss_vals as well.
     diff_of_loss_vals_and_ref_point = reference_point - rank_i_loss_vals
-    assert subset_size <= rank_i_indices.size
-    n_objectives = reference_point.size
+    (n_solutions, n_objectives) = rank_i_loss_vals.shape
+    assert isinstance(n_solutions, int), "MyPy Redefinition for NumPy v2.2.0."
     contribs = np.prod(diff_of_loss_vals_and_ref_point, axis=-1)
     selected_indices = np.zeros(subset_size, dtype=int)
     selected_vecs = np.empty((subset_size, n_objectives))
-    indices: np.ndarray[tuple[int, ...], np.dtype[Any]] = np.arange(
```
nabenabe0928, line 102: This typing is also not correct, because `indices` is always a 1d array.
As this typing says `indices` can be a ND array, which is not correct, I removed this.

optuna/_gp/optim_mixed.py
```diff
@@ -208,7 +208,8 @@ def local_search_mixed(
     # TODO(kAIto47802): Think of a better way to handle this.
     lengthscales = 1 / np.sqrt(inverse_squared_lengthscales[continuous_indices])
 
-    discrete_indices = np.where(steps > 0)[0]
+    # NOTE(nabenabe): MyPy Redefinition for NumPy v2.2.0. (Cast signed int to int)
+    discrete_indices = np.where(steps > 0)[0].astype(int)
```
nabenabe0928, line 212: By casting from unsinged int to int, we can avoid the issue although this loosens the typing.

optuna/study/_multi_objective.py
```diff
@@ -127,17 +126,18 @@ def _is_pareto_front_nd(unique_lexsorted_loss_values: np.ndarray) -> np.ndarray:
     loss_values = unique_lexsorted_loss_values[:, 1:]
     n_trials = loss_values.shape[0]
     on_front = np.zeros(n_trials, dtype=bool)
-    nondominated_indices: np.ndarray[tuple[int, ...], np.dtype[np.signedinteger[Any]]] = np.arange(
-        n_trials
-    )
```
nabenabe0928, line 132: The same reason
- https://github.com/optuna/optuna/pull/5853#discussion_r1877207250

optuna/study/_multi_objective.py
```diff
@@ -127,17 +126,18 @@ def _is_pareto_front_nd(unique_lexsorted_loss_values: np.ndarray) -> np.ndarray:
...
         nondominated_and_not_top = np.any(loss_values < loss_values[0], axis=1)
         # NOTE: trials[j] cannot dominate trials[i] for i < j because of lexsort.
         # Therefore, nondominated_indices[0] is always non-dominated.
         on_front[nondominated_indices[0]] = True
         loss_values = loss_values[nondominated_and_not_top]
-        nondominated_indices = nondominated_indices[nondominated_and_not_top]
+        # NOTE(nabenabe0928): Ignore typing for a temporal solution to NumPy v2.2.0 weird typing.
+        nondominated_indices = nondominated_indices[
+            nondominated_and_not_top
+        ]  # type: ignore[assignment]
```
nabenabe0928, line 140: I could not find any good ways to tell NumPy that the return of this vectorization is gonna be a 1d integer array, so I ignore the typing here for now.

HideakiImamura:
@porink0424 @y0z Could you review this PR?

optuna/study/_multi_objective.py
nabenabe0928: I found this option, so I replaced the code.
- https://github.com/numpy/numpy/issues/27957#issuecomment-2531536212

optuna/_hypervolume/hssp.py
```diff
@@ -94,19 +92,18 @@ def _solve_hssp_on_unique_loss_vals(
...
-        rank_i_loss_vals.shape[0], dtype=int
-    )
+    indices = np.arange(n_solutions)
     for k in range(subset_size):
         max_index = int(np.argmax(contribs))
         selected_indices[k] = indices[max_index]
         selected_vecs[k] = rank_i_loss_vals[max_index].copy()
-        keep = np.ones(contribs.size, dtype=bool)
+        assert n_solutions - k > 0
+        keep = np.ones(n_solutions - k, dtype=bool)
```
porink0424, line 106: The implementation code itself has changed, and I'm concerned it might result in different behavior.

This seems to go beyond the scope of the PR, which is meant to fix static analysis issues with mypy, so  would it be possible to adjust it to stay within the scope?

optuna/study/_multi_objective.py
```diff
@@ -127,17 +126,22 @@ def _is_pareto_front_nd(unique_lexsorted_loss_values: np.ndarray) -> np.ndarray:
     loss_values = unique_lexsorted_loss_values[:, 1:]
     n_trials = loss_values.shape[0]
     on_front = np.zeros(n_trials, dtype=bool)
-    nondominated_indices: np.ndarray[tuple[int, ...], np.dtype[np.signedinteger[Any]]] = np.arange(
-        n_trials
-    )
+    # TODO(nabenabe): Replace with the following once Python 3.8 is dropped.
+    # nondominated_indices: np.ndarray[tuple[int], np.dtype[np.signedinteger]] = ...
+    nondominated_indices = np.arange(n_trials)
```
porink0424, line 131: The original type `np.ndarray[tuple[int, ...], np.dtype[np.signedinteger[Any]]]` is not incorrect, just loosely typed, so I think that leaving a TODO comment would be better than using a type ignore statement.

porink0424 (changes requested):
I left a few comments, PTAL.

optuna/study/_multi_objective.py
nabenabe0928: Done!

optuna/_hypervolume/hssp.py
nabenabe0928: Done!

y0z (approved):
LGTM

optuna/_hypervolume/hssp.py
```diff
@@ -94,19 +92,18 @@ def _solve_hssp_on_unique_loss_vals(
 
     # The following logic can be used for non-unique rank_i_loss_vals as well.
     diff_of_loss_vals_and_ref_point = reference_point - rank_i_loss_vals
-    assert subset_size <= rank_i_indices.size
-    n_objectives = reference_point.size
+    (n_solutions, n_objectives) = rank_i_loss_vals.shape
+    n_solutions = int(n_solutions)  # MyPy Redefinition for NumPy v2.2.0.
```
porink0424, line 96: As I mentioned in the previous review, I still have some concerns due to my limited understanding of the logic:
- Is it safe to remove the assertion comparing `subset_size` and `rank_i_indices.size`?
- Are `rank_i_loss_vals.shape[1]` and `reference_point.size` guaranteed to be the same value?

optuna/_hypervolume/hssp.py
```diff
@@ -94,19 +92,18 @@ def _solve_hssp_on_unique_loss_vals(
...
-        rank_i_loss_vals.shape[0], dtype=int
-    )
+    indices = np.arange(n_solutions)
     for k in range(subset_size):
         max_index = int(np.argmax(contribs))
         selected_indices[k] = indices[max_index]
         selected_vecs[k] = rank_i_loss_vals[max_index].copy()
-        keep = np.ones(contribs.size, dtype=bool)
+        assert n_solutions - k > 0
+        keep = np.ones(n_solutions - k, dtype=bool)
```
porink0424, line 106: Is it possible to use `contribs.size` as it is in the original logic?

porink0424 (commented):
It seems that the refactor goes beyond addressing the errors by mypy.

optuna/_hypervolume/hssp.py
nabenabe0928: I tweaked the code and I made it.
But interestingly, the numpy sizing is now very sensitive to the cast information.

For example, if we replace `indices = np.arange(n_solutions)` with `indices = np.arange(int(n_solutions))`, `contribs.size` does not work anymore.

Anyways, your concern was resolved.

y0z (approved):
LGTM

optuna/_hypervolume/hssp.py
nabenabe0928: > Is it safe to remove the assertion comparing subset_size and rank_i_indices.size?

Yes, it is, because we had `assert n_solutions - k > 0` in the for loop, but as we removed this line, I reverted.

The original assertion was `assert subset_size < n_solutions`.
What we did was `assert k < n_solutions`, which we, in turn, check till whether `subset_size - 1 < n_solutions`.
The only difference is that we do not check the case of `subset_size < n_solutions`, but because of the following line:

https://github.com/optuna/optuna/blob/298653ed09df8bd6350f424dcbe55855355987fa/optuna/_hypervolume/hssp.py#L90-L91 

we were already checking `subset_size != n_solutions` and `subset_size - 1 < n_solutions`, leading to `subset_size <= n_solutions and subset_size != n_solutions`, i.e. `subset_size < n_solutions`.

This completed the original assertion.

> Are rank_i_loss_vals.shape[1] and reference_point.size guaranteed to be the same value?

Yes, it is guaranteed.

First of all, the code-level wise, it is guaranteed as `_solve_hssp` is only used here.

https://github.com/optuna/optuna/blob/298653ed09df8bd6350f424dcbe55855355987fa/optuna/samplers/_tpe/sampler.py#L682-L683

But most importantly, `_solve_hssp` is, roughly speaking, a sorting algorithm for an array with a shape of `(n_solutions, n_objectives)` given a reference point with the shape of `(n_objectives, )`.

porink0424 (approved):
Thank you for the detailed explanation and prompt fix.
LGTM 👍

PR #5864, opened 2024-12-13
Add `BaseGASampler`

gen740:
## Motivation
In optuna GA (Genetic Algorithm) samplers are not easy to implement, this is because each sampler should handle generation id and caches correctly.

So for easy implementation. I implemented the abstract class of the GASampler `BaseGASampler` for general GA.

There are also some bugs in the current implementation.
The nsgaii sampler stopped with a random multiprocess error at [this line.](https://github.com/optuna/optuna/blob/e6e0cb4c8cd48da6f8f82d45d3a09d5299d947ef/optuna/samplers/nsgaii/_sampler.py#L349)

## Performance
I created a simple benchmark with the following simple objective function.
```python
def objective(trial):
    return (x - 2) ** 2 + (y - 3) ** 2, (x + 2) ** 2 + (y + 3) ** 2
``` 

Run with in-memory storage and 3000 trials. Repeated 20 times. And the results are as follows

| Current implementation | This implementation |
| ---------------------- | ------------------- |
| 1.476 ± 0.016s         | 1.477 ± 0.008s      |

## Description of the changes
- Introduction of `BaseGASampler`.
- Rewrite the NSGA-II sampler with `BaseGASampler`.
- Add tests.


HideakiImamura:

HideakiImamura:

optuna/samplers/_ga/_base.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @classmethod
+    def _get_generation_key(cls) -> str:
+        return cls._GENERATION_KEY
+
+    @classmethod
+    def _get_parent_cache_key_prefix(cls) -> str:
+        return cls._PARENT_CACHE_KEY_PREFIX
+
+    def __init__(self, population_size: int):
```
y0z, line 49: How about allowing `population_size` to be `None` in the constructor?
`population_size` is often automatically determined based on the dimensionality of the search space.
For instance, `CmaEsSampler` adopts such a mechanism as follows.
https://github.com/optuna/optuna/blob/46d7dab172ecb62ea3a0bad87e771ab25905f0ff/optuna/samplers/_cmaes.py#L388-L389

optuna/samplers/_ga/_base.py
```diff
@@ -0,0 +1,180 @@
...
+        Returns:
+            Generation number of the given trial.
+        """
+        generation = trial.system_attrs.get(self._get_generation_key(), None)
+        if generation is not None:
+            return generation
+
+        trials = study._get_trials(deepcopy=False, states=[TrialState.COMPLETE], use_cache=True)
+
+        max_generation, max_generation_number = 0, 0
```
y0z, line 107: I prefer `max_generation_individuals_count` or something easy to understand to `max_generation_number` based on my understanding that this variable indicates the number of individuals belonging to the maximum generation.

optuna/samplers/_ga/_base.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def get_generation(self, study: optuna.Study, trial: FrozenTrial) -> int:
+        """Get the current generation number of the given trial.
+
+        This method returns the generation number of the specified trial. If the generation number
+        is not set in the trial's system attributes, it will calculate and set the generation
+        number.
+
+        The current generation number is defined as the maximum generation number among all
+        completed trials.
```
y0z, line 90: I think that the generation number can be `max_generation + 1`.

optuna/samplers/_ga/_base.py
```diff
@@ -0,0 +1,180 @@
...
+                Trial object to get the generation number.
+
+        Returns:
+            Generation number of the given trial.
+        """
+        generation = trial.system_attrs.get(self._get_generation_key(), None)
+        if generation is not None:
+            return generation
+
+        trials = study._get_trials(deepcopy=False, states=[TrialState.COMPLETE], use_cache=True)
```
y0z, line 105: I'm wondering if this method works in multi-threading.
Will a scenario like the following never occur, or is it not problematic?

Thread1: call `get_generation(study, trial)`
Thread2: call `get_generation(study, trial)`
Thread1: call `study._get_trials(deepcopy=False, states=[TrialState.COMPLETE], use_cache=True)`
Thread3: Increase the number of completed trials by finishing another trial
Thread2: call `study._get_trials(deepcopy=False, states=[TrialState.COMPLETE], use_cache=True)`
Thread1: `study._storage.set_trial_system_attr(trial._trial_id, self._get_generation_key(), generation)`
Thread2: `study._storage.set_trial_system_attr(trial._trial_id, self._get_generation_key(), generation)`

optuna/samplers/_ga/_base.py
```diff
@@ -0,0 +1,180 @@
...
+        Returns:
+            List of frozen trials in the given generation.
+        """
+        return [
+            trial
+            for trial in study._get_trials(
+                deepcopy=False, states=[TrialState.COMPLETE], use_cache=True
+            )
+            if trial.system_attrs.get(self._get_generation_key(), None) == generation
+        ]
```
y0z, line 147: Although this is a mere question, is it possible for a trial without a generation attribute to be complete?

optuna/samplers/_ga/_base.py
```diff
@@ -0,0 +1,180 @@
...
+        This method caches the parent population in the study's system attributes.
+
+        Args:
+            study:
+                Target study object.
+            generation:
+                Target generation number.
+
+        Returns:
+            List of parent frozen trials.
```
y0z, line 161: It is better to describe the behavior that if `generation == 0`, an empty list is returned.

optuna/samplers/_ga/_base.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        if generation == 0:
+            return []
+
+        study_system_attrs = study._storage.get_study_system_attrs(study._study_id)
+        cached_parent_population = study_system_attrs.get(
```
y0z, line 167: It seems that `cached_parent_population_ids` is more precise than `cached_parent_population`.

y0z (commented):
I leave some comments on `_base.py`.

optuna/samplers/_ga/_base.py
```diff
@@ -0,0 +1,180 @@
...
+            self._get_parent_cache_key_prefix() + str(generation), None
+        )
+
+        if cached_parent_population is not None:
+            return [study._storage.get_trial(trial_id) for trial_id in cached_parent_population]
+        else:
+            parent_population = self.select_parent(study, generation)
+            study._storage.set_study_system_attr(
```
y0z, line 175: TODO: Check race conditions

optuna/samplers/_ga/_base.py
gen740: I think there is some possiblility to change samplers while running like `auto_sampler`. In this case some trials have no generation_keys in their `system_attrs` so I think this is reasonable to return None.

porink0424 (commented):
I've left a few comments, please take a look.

porink0424 (approved):
LGTM!

y0z (approved):
LGTM

PR #5872, opened 2024-12-20
Introduce client-side cache in GrpcStorageProxy

c-bata:
## Motivation

Follow-up #5852 
To make `GrpcStorageProxy` faster by introducing a client-side cache on GrpcStorageProxy.

## Description of the changes

This PR makes `GrpcStorageProxy` faster by cache `storage.get_all_trials` calls. Here is the performance comparison results and the benchmark script I used.

| `master` branch | This PR |
| --- | --- |
| 87.392s (+/- 0.6052104) | 48.293s (+/- 0.223478) |

```bash
# Start a MySQL server
$ docker run -d --rm -p 3306:3306 -v ./mysql-conf:/etc/mysql/conf.d -e MYSQL_USER=optuna -e MYSQL_DATABASE=optuna -e MYSQL_PASSWORD=password -e MYSQL_ALLOW_EMPTY_PASSWORD=yes --name optuna-mysql mysql:8.0

# Start a gRPC storage proxy server
$ python3 -c "import optuna; from concurrent.futures import ThreadPoolExecutor; from optuna.storages import run_grpc_proxy_server; run_grpc_proxy_server(storage=optuna.storages.get_storage('mysql+pymysql://optuna:password@127.0.0.1:3306/optuna'), host='localhost', port=13000, thread_pool=ThreadPoolExecutor(10))"
```

```python

from concurrent.futures import as_completed
import threading
import time
from typing import Literal

import numpy as np

import optuna
from optuna.storages import BaseStorage
from optuna.storages import GrpcStorageProxy


study_name = "grpc-example"
n_workers = 10
n_repeat = 5
n_trials = 1000

STORAGE_KIND = Literal["grpc", "rdb"]
storage_kinds: list[STORAGE_KIND] = ["grpc"]


def get_storage(storage_kind: STORAGE_KIND) -> BaseStorage:
    if storage_kind == "grpc":
        return GrpcStorageProxy(host="localhost", port=13000)
    elif storage_kind == "rdb":
        return optuna.storages.get_storage("mysql+pymysql://optuna:password@127.0.0.1:3306/optuna")


    s = 0.0
    for i in range(10):
        s += trial.suggest_float(f"x{i}", -10, 10) ** 2
    return s


def study_optimize(storage_kind: STORAGE_KIND, n_trials: int) -> int:
    start = time.time()
    storage = get_storage(storage_kind)
    study = optuna.load_study(
        storage=storage, study_name=study_name, sampler=optuna.samplers.RandomSampler()
    )
    study.optimize(objective, n_trials=n_trials)
    elapsed = time.time() - start
    return elapsed


def bench_study_optimize(storage_kind: STORAGE_KIND) -> list[int]:
    elapsed = []
    study = optuna.create_study(storage=get_storage("rdb"), study_name=study_name)
    with ProcessPoolExecutor(n_workers) as pool:
        futures = [
            pool.submit(study_optimize, storage_kind, n_trials / n_workers)
            for _ in range(n_workers)
        ]
        for fut in as_completed(futures):
            elapsed.append(fut.result())
    return elapsed


def main():
    try:
        optuna.delete_study(storage=get_storage("rdb"), study_name=study_name)
    except Exception:
        pass

    elapsed_optimize = {}
    elapsed_get_all_trials = {}
    for storage_kind in storage_kinds:
        elapsed_optimize[storage_kind] = bench_study_optimize(storage_kind)
        optuna.delete_study(storage=get_storage("rdb"), study_name=study_name)

    for key in elapsed_optimize:
        print(f"{key=} {n_trials=} {n_repeat=} {n_workers=}")
        elapsed = elapsed_optimize[key]
        print(f"  study.optimize: {np.mean(elapsed)=} {np.std(elapsed)=}")


if __name__ == "__main__":
    main()
```


HideakiImamura (dismissed):
Thanks for the follow-up. Looks very good. i confirmed that all logics are consistent with the cache logic of `_CachedStorage` and match with the current implementation.

optuna/storages/_grpc/client.py
```diff
@@ -102,6 +105,7 @@ def delete_study(self, study_id: int) -> None:
             if e.code() == grpc.StatusCode.NOT_FOUND:
                 raise KeyError from e
             raise
+        self._cache.delete_study_cache(study_id)
```
c-bata, line 108: I've found an issue regarding the cache invalidation.

With the current schema of Optuna's SQLite3 database, SQLite3 reuses `study_id`, as shown below.

```sql
sqlite> .schema studies
CREATE TABLE studies (
        study_id INTEGER NOT NULL, 
        study_name VARCHAR(512) NOT NULL, 
        PRIMARY KEY (study_id)
);
```

```python
>>> storage = optuna.storages.RDBStorage("sqlite:///foo.db")
>>> storage.create_new_study(directions=[optuna.study.StudyDirection.MINIMIZE])
[I 2024-12-20 17:10:54,500] A new study created in RDB with name: no-name-2a8665b6-fb78-48d5-937b-cd2c5a71f9ac
1
>>> storage.delete_study(study_id=1)
>>> storage.create_new_study(directions=[optuna.study.StudyDirection.MINIMIZE])
[I 2024-12-20 17:11:04,641] A new study created in RDB with name: no-name-0b71c4ee-123f-4894-9903-4014c6581942
```

Consequently, when multiple gRPC storage proxy servers are running and `storage.delete_study()` is called, the cache may not be invalidated. This issue does not occur with MySQL, as shown below.

```python
>>> import optuna
>>> storage = optuna.storages.get_storage("mysql+pymysql://optuna:password@127.0.0.1:3306/optuna")
>>> storage.create_new_study(directions=[optuna.study.StudyDirection.MINIMIZE])
[I 2024-12-20 18:37:24,719] A new study created in RDB with name: no-name-95110374-00cc-4136-a1a3-58d8e5e88cba
132
>>> storage.delete_study(study_id=132)
>>> storage.create_new_study(directions=[optuna.study.StudyDirection.MINIMIZE])
[I 2024-12-20 18:37:30,603] A new study created in RDB with name: no-name-feac51b7-be94-407c-8b83-2e0e8e2d0b9d
133
```

This issue is relatively limited, as it only arises when multiple gRPC storage proxies are running with a single SQLite3 database, and o`ptuna.delete_study()` are executed. So, my suggestion is to proceed with this PR and address this issue in the future by adding `AUTOINCREMENT` (please refer to https://www.sqlite.org/autoinc.html for more details).

c-bata:
@y0z @nabenabe0928 Could you review this PR?

optuna/storages/_grpc/api.proto
```diff
@@ -378,17 +378,18 @@ message GetTrialReply {
...
 
 /**
- * Request to get all trials in a study.
+ * Request to get trials in a study.
  */
-message GetAllTrialsRequest {
+message GetTrialsRequest {
     int64 study_id = 1;
-    repeated TrialState states = 2;
+    repeated int64 included_trial_ids = 2;
```
nabenabe0928, line 385: Probably, `including` is better if I understand the context of this request.
```suggestion
    repeated int64 including_trial_ids = 2;
```

nabenabe0928 (commented):
I feel like the following methods can also use cache, but do you wanna include them in this PR, make another PR, or do not wanna do it at all? 

- [ ] get_study_directions
- [ ] get_trial_id_from_study_id_trial_number
- [ ] get_trial
- [ ] get_study_name_from_id
- [ ] get_study_id_from_name

nabenabe0928 (commented):
I will approve this PR once I confirm the performance.

y0z:
The changes look good to me.

However, I sometimes encountered the following error when I ran the benchmarking script.
This problem happened for **both the master and this PR**.
Moreover, the problem occurs stochastically and is not deterministic.
This makes it difficult to reproduce.
This problem should be resolved in another PR.

```python
[I 2024-12-24 15:26:34,943] Trial 924 finished with value: 290.2282271538501 and parameters: {'x7': [I 2024-12-24 15:40:03,132] Trial 972 finished with value: 225.24752071871094 and parameters: {'x7': -6.2522912173605505, 'x2': -9.919431600233237, 'x4': 0.044403182081824255, 'x1': 4.557349591146185, 'x6': 1.7976231513999519, 'x0': 0.3857780539269626, 'x3': 6.8879029045391285, 'x9': 2.375230714328289, 'x5': -2.495022132598197, 'x8': 2.0735256279292553}. Best is trial 625 with value: 80.69101882439877.
[I 2024-12-24 15:40:03,648] Trial 973 finished with value: 289.3464461810266 and parameters: {'x7': 2.381450328633779, 'x2': 1.9933529378663124, 'x4': 2.258098349460056, 'x1': 3.975028662239829, 'x6': 1.9883090130037946, 'x0': 7.296691363420539, 'x3': -4.43790924552073, 'x9': 9.69317044115908, 'x5': -6.608645512992846, 'x8': 6.6543188068702435}. Best is trial 625 with value: 80.69101882439877.
[I 2024-12-24 15:40:04,151] Trial 974 finished with value: 206.050847349017 and parameters: {'x7': -3.851493335810887, 'x2': -3.5448847331385664, 'x4': -2.0483589374393807, 'x1': 5.714988193665674, 'x6': -8.106049930368089, 'x0': 3.675040384660633, 'x3': -3.2218795371186637, 'x9': -1.1367502366625644, 'x5': -7.13484941915987, 'x8': 0.03196690174535988}. Best is trial 625 with value: 80.69101882439877.
[I 2024-12-24 15:40:04,611] Trial 975 finished with value: 282.71786380668885 and parameters: {'x7': -5.74931015351887, 'x2': -4.28596562642862, 'x4': 4.491680141165432, 'x1': 8.65879309760668, 'x6': -8.477996393827132, 'x0': 4.373272358617948, 'x3': 1.7027574571316748, 'x9': 0.46300610792595265, 'x5': -5.839393434437474, 'x8': 2.8159722200976454}. Best is trial 625 with value: 80.69101882439877.
concurrent.futures.process._RemoteTraceback: 
"""
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
        status = StatusCode.UNKNOWN
        details = "Exception calling application: An exception is raised during the commit. This typically happens due to invalid data in the commit, e.g. exceeding max length. "
        debug_error_string = "UNKNOWN:Error received from peer  {grpc_message:"Exception calling application: An exception is raised during the commit. This typically happens due to invalid data in the commit, e.g. exceeding max length. ", grpc_status:2, created_time:"2024-12-24T15:38:04.407866+09:00"}"
>

  File "/Users/yozaki/.asdf/installs/python/3.11.7/lib/python3.11/concurrent/futures/process.py", line 217, in _sendback_result
    result_queue.put(_ResultItem(work_id, result=result,
  File "/Users/yozaki/.asdf/installs/python/3.11.7/lib/python3.11/multiprocessing/queues.py", line 371, in put
    obj = _ForkingPickler.dumps(obj)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yozaki/.asdf/installs/python/3.11.7/lib/python3.11/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: cannot pickle '_thread.RLock' object
"""

The above exception was the direct cause of the following exception:

  File "/Users/yozaki/work/review/optuna/benchmark.py", line 81, in <module>
    main()
  File "/Users/yozaki/work/review/optuna/benchmark.py", line 71, in main
    elapsed_optimize[storage_kind] = bench_study_optimize(storage_kind)
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yozaki/work/review/optuna/benchmark.py", line 59, in bench_study_optimize
    elapsed.append(fut.result())
                   ^^^^^^^^^^^^
  File "/Users/yozaki/.asdf/installs/python/3.11.7/lib/python3.11/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/yozaki/.asdf/installs/python/3.11.7/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
TypeError: cannot pickle '_thread.RLock' object
```

y0z (approved):

y0z (commented):
Please apply flake8.
https://github.com/optuna/optuna/actions/runs/12429266158/job/34702406029?pr=5872

nabenabe0928:
Hmmm, I tried with `n_trials=300`, but I could only observe degradation...

Plus, it took me so long to finish `n_trials=1000` (300+ seconds) in my environment.

I used MySQL as a database and Python 3.11 on Ubuntu 20.04.

```
$ pip freeze
# Output
alembic==1.14.0
colorlog==6.9.0
greenlet==3.1.1
grpcio==1.68.1
Mako==1.3.8
MarkupSafe==3.0.2
mysqlclient==2.2.6
numpy==2.2.1
packaging==24.2
protobuf==5.29.2
PyYAML==6.0.2
SQLAlchemy==2.0.36
tqdm==4.67.1
typing_extensions==4.12.2
```

nabenabe0928:
I measured the runtime of pure RDBStorage in my env and it took me about a minute for the code below.

<details>

```python
import os
import time

import optuna


    s = 0.0
    for i in range(10):
        s += trial.suggest_float(f"x{i}", -10, 10) ** 2
    
    return s


study = optuna.create_study(sampler=optuna.samplers.RandomSampler(), storage=os.environ["OPTUNA_STORAGE"])
start = time.time()
study.optimize(objective, n_trials=1000, n_jobs=10)
print(time.time() - start)

```

</detials>

optuna/storages/_grpc/api.proto
HideakiImamura: I think the current naming and implementation are consistent with the `RDBStorage` ([ref](https://github.com/optuna/optuna/blob/master/optuna/storages/_rdb/storage.py#L812)).

optuna/storages/_grpc/client.py
```diff
@@ -62,12 +67,14 @@ def __init__(self, *, host: str = "localhost", port: int = 13000) -> None:
...
         self._host = host
         self._port = port
 
     def __getstate__(self) -> dict[Any, Any]:
         state = self.__dict__.copy()
         del state["_stub"]
+        del state["_cache"]
         return state
 
     def __setstate__(self, state: dict[Any, Any]) -> None:
```
HideakiImamura, line 80: We need to create the deleted `_cache` here.

```python
    def __setstate__(self, state: dict[Any, Any]) -> None:
        self.__dict__.update(state)
        self._stub = api_pb2_grpc.StorageServiceStub(
            grpc.insecure_channel(f"{self._host}:{self._port}")
        )  # type: ignore
        self._cache = GrpcClientCache(self._stub)
```

HideakiImamura:
I confirmed that this PR is faster than the current master by using the following benchmark code. It looks like the `get_all_trials` is completely speeded up.

- Server side code
```python

import os
from optuna.storages import run_grpc_proxy_server
import optuna
import signal


STORAGE_URL = "YOUR_MYSQL_URL"

def handler(signum, frame):
    print("Received signal", signum)
    exit(0)

if __name__ == "__main__":
    signal.signal(signal.SIGINT, handler)
    signal.signal(signal.SIGTERM, handler)

    run_grpc_proxy_server(
        optuna.storages.get_storage(STORAGE_URL),
        host="localhost", 
        port=13000,
    )

```

- Client side code
```python

import argparse
from concurrent.futures import ProcessPoolExecutor, as_completed
from collections.abc import Sequence
from collections.abc import Container
import cProfile
import time
import numpy as np
import optuna
import threading
from typing import Literal
from optuna.storages import GrpcStorageProxy, BaseStorage
import os
import uuid
from time import sleep


study_name = "grpc-example"
n_workers = 10
n_repeat = 5
n_trials = 1000

STORAGE_KIND = Literal["grpc", "rdb", "middleware"]
STORAGE_URL = "YOUR_MYSQL_URL"


def get_storage(storage_kind: STORAGE_KIND) -> BaseStorage:
    if storage_kind == "grpc":
        return GrpcStorageProxy(host="localhost", port=13000)
    elif storage_kind == "rdb":
        return optuna.storages.get_storage(STORAGE_URL)


    s = 0.0
    for i in range(10):
        trial.set_user_attr(f"attr{i}", "attr value")
    for i in range(10):
        s += trial.suggest_float(f"x{i}", -10, 10) ** 2
    return s


def study_optimize(storage_kind: STORAGE_KIND, n_trials: int) -> int:
    start = time.time()
    storage = get_storage(storage_kind)
    while True:
        try:
            study = optuna.load_study(storage=storage, study_name=study_name, sampler=optuna.samplers.RandomSampler())
            break
        except:
            sleep(10)
    study.optimize(objective, callbacks=[optuna.study.MaxTrialsCallback(n_trials)])
    elapsed = time.time() - start
    return elapsed


def bench_study_optimize(storage_kind: STORAGE_KIND) -> list[int]:
    elapsed = []
    study = optuna.create_study(
        storage=get_storage("rdb"), 
        study_name=study_name, 
        load_if_exists=True,
        sampler=optuna.samplers.RandomSampler(),
    )
    study_optimize(storage_kind, n_trials)
    return elapsed


def main(storage: STORAGE_KIND, benchmark_file: str):
    try:
        optuna.delete_study(storage=get_storage("rdb"), study_name=study_name)
    except:
        pass
    
    cProfile.runctx(
        f"bench_study_optimize('{storage}')", 
        globals(), 
        locals(), 
        filename=benchmark_file,
    )
    optuna.delete_study(storage=get_storage(storage), study_name=study_name)


if __name__ == "__main__":
    parser.add_argument("--storage", type=str, required=True)
    parser.add_argument("--benchmark_file", type=str, required=True)
    args = parser.parse_args()
    main(args.storage, args.benchmark_file)

```

- Results

| Current master | This PR |
| --- | --- |
| 604 sec | 203 sec |
| Screenshot 2024-12-26 at 10 35 09 | Screenshot 2024-12-26 at 10 35 29 |

optuna/storages/_grpc/client.py
```diff
@@ -53,6 +54,9 @@ class GrpcStorageProxy(BaseStorage):
         host: The hostname of the gRPC server.
         port: The port of the gRPC server.
 
+    .. warning::
+
+       Currently, gRPC storage proxy does not work as expected when using a SQLite3 database as the backend and calling :func:`optuna.delete_study`.
```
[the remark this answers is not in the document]
c-bata: The gRPC storage proxy does accept `optuna.delete_study()` without raising exceptions. However, it may result in an inconsistent state when using an SQLite3 database as the backend and multiple storage proxies are running.

c-bata:
> I feel like the following methods can also use cache, but do you wanna include them in this PR, make another PR, or do not wanna do it at all?
> 
> * [ ]  get_study_directions
> * [ ]  get_trial_id_from_study_id_trial_number
> * [ ]  get_trial
> * [ ]  get_study_name_from_id
> * [ ]  get_study_id_from_name

Thank you for your review. I think we don't need to use cache for these methods unless they become a bottleneck. What do you think?

optuna/storages/_grpc/api.proto
c-bata: Thank you for your review! I'll address this with RDBStorage's `_get_trials()` method in the following pull request, as this is a private api.

optuna/storages/_grpc/client.py
c-bata: Thank you for your review. I addressed this in https://github.com/optuna/optuna/pull/5872/commits/baffa73f37b00a43c7cfea92605c490d80117bb4

c-bata:
Let me unassign @y0z and assign @HideakiImamura instead.

nabenabe0928:
- https://github.com/optuna/optuna/pull/5872#issuecomment-2560920622

Ok, I figured out why I did not get the speedup.

I originally used the following code, leading to the use of RDBStorage without cache in local process: 

```python
import os

from optuna.storages import run_grpc_proxy_server
from optuna.storages import RDBStorage


storage = RDBStorage(os.environ["OPTUNA_STORAGE"])
run_grpc_proxy_server(storage, host="localhost", port=13000)

```

However, when I used `get_storage`, which returns `_CachedStorage` wrapping `RDBStorage`, I observed the speedup.

```python
import os

from optuna.storages import run_grpc_proxy_server
from optuna.storages import get_storage

storage = get_storage(os.environ["OPTUNA_STORAGE"])
run_grpc_proxy_server(storage, host="localhost", port=13000)

```

Master: 112 seconds, This PR: 61 seconds.

Probably, the doc-string should be changed [here](https://github.com/optuna/optuna/blob/ea9ed5dbde5da466729d5c29c02f3440ada1011b/optuna/storages/_grpc/server.py#L450-L474).

nabenabe0928:
Followups:
- Replace doc-string [here](https://github.com/optuna/optuna/blob/ea9ed5dbde5da466729d5c29c02f3440ada1011b/optuna/storages/_grpc/server.py#L450-L474)
- Replace `try_import` with `_LazyImport`

HideakiImamura (approved):
LGTM.

nabenabe0928 (approved):

PR #5892, opened 2024-12-27
Fix contour plot of matplotlib

fusawa-yugo:
## Motivation

Outputs of coutour plot are different between plotly and matplotlib. 
Visualization of matplotlib is not working properly.

スクリーンショット 2024-12-27 17 27 32

- how to reproduce
Use these and visualize through `optuna.visualization.matplotlib.plot_contour`
https://github.com/optuna/optuna-visualization-regression-tests/blob/6b98f408f82a613fc5271cfb12308649e3d55211/studies.py#L45-L53
https://github.com/optuna/optuna-visualization-regression-tests/blob/6b98f408f82a613fc5271cfb12308649e3d55211/studies.py#L24-L29

In this case,  if `category` is `bar`, `x1` is missing as shown below.
```
Params: {'category': 'bar', 'x2': -6.85817929203591}
Params: {'category': 'foo', 'x1': 9.955852432587085}
Params: {'category': 'foo', 'x1': 8.553363427670403}
```
Now, when plotting a gragh related to `x1`, `bar` is dropped from `category` axis.
This is the same for `foo` and `x1` .

## Description of the changes

I solved this by including all values on the axis with a categorical distribution, regardless of whether the value on the other axis is None.

contour_mat_4 2dev


optuna/visualization/matplotlib/_contour.py
```diff
@@ -175,7 +175,9 @@ def _calculate_axis_data(
     returned_values: Sequence[int | float]
     if axis.is_cat:
         enc = _LabelEncoder()
-        returned_values = enc.fit_transform(list(map(str, values)))
+        # prepare all values in categorical distribution
```
toshihikoyanase, line 178: A nit, but please follow the coding convention: "Block and inline comments are supposed to be sentence(s)."
https://github.com/optuna/optuna/wiki/Coding-Style-Conventions#block-and-inline-comments-are-supposed-to-be-sentences

```suggestion
        # Prepare all values in categorical distributions.
```

HideakiImamura:

HideakiImamura (commented):
Thanks for the PR. The change looks good to me. Is it possible to introduce the test cases to validate this change?

nabenabe0928:
The PR description could be better by the following revisions:


### Describe how they look differently

In my eyes, the difference was not that obvious.
If I understand the difference correctly, each plot includes only a part of categories, making the dots only on the left side or the bottom of each plot in matplotlib.

### Claim which figure is for matplotlib

In the first figure, it is not obvious which figure shows the result obtained by matplotlib, so it would be much more friendly with the guide.

### Add a code block for smoother PR reviews

It is not that hard to add a code block, but it significantly makes reviews easier.

<details>
<summary>Demo Code</summary>

```python
import optuna


    category = trial.suggest_categorical("category", ["foo", "bar"])
    if category == "foo":
        return (trial.suggest_float("x1", 0, 10) - 2) ** 2
    else:
        return -((trial.suggest_float("x2", -10, 0) + 5) ** 2)


study = optuna.create_study()
fig = optuna.visualization.plot_contour(study)
fig.show()
optuna.visualization.matplotlib.plot_contour(study)
plt.show()

```

</details>

optuna/visualization/matplotlib/_contour.py
```diff
@@ -175,7 +175,9 @@ def _calculate_axis_data(
...
+        enc.fit(list(map(str, filter(lambda value: value is not None, axis.values))))
```
nabenabe0928, line 179: Please note the following comment as well:
- https://github.com/optuna/optuna/pull/5892/files#r1898380046

```suggestion
        # Fit LabelEncoder with all the categories in categorical distribution.
        enc.fit(list(map(str, filter(lambda value: value is not None, axis.values))))
        # Then transform the values using the fitted label encoder.
        # Note that `values` may not include all the categories, so we use `axis.values` for fitting.
```

nabenabe0928:
Seems the interpolation algorithm is a bit different from plotly, but that is not a range of this PR, so I will leave it as is.

fusawa-yugo:
I fixed comments.
Could you please review it?

nabenabe0928 (approved):
LGTM!

HideakiImamura:
@fusawa-yugo Could you add a unittest for this change?

fusawa-yugo:
@HideakiImamura 
I discussed with @c-bata, and I think that translating the expected inputs and outputs of this private function into unit tests is not straightforward in this case. 
Therefore, I feel if it might be reasonable to omit adding unit tests this time.

HideakiImamura (approved):
Thanks for the discussion. I agree with you. It would be great to work on the test of matplotlib contour tests as a follow-up.

PR #5893, opened 2024-12-27
Resolve the issue where `BruteForceSampler` fails to suggest all combinations

### LINKED ISSUE #5780: `BruteForceSampler` stops before suggesting all combinations in parallel optimization
### Expected behavior

`BruteForceSampler` should suggest all parameter combinations, but it stops unexpectedly before that in parallel optimization.

### Environment

- Optuna version:4.2.0.dev
- Python version:3.9.4
- OS:macOS-15.1-x86_64-i386-64bit

```shell
See the following MRE.
```

### Steps to reproduce

```python
import time
import optuna

def objective(trial):
    x = trial.suggest_int("x", 0, 1)
    time.sleep(x)
    y = trial.suggest_int("y", 0, 1)
    return x + y

sampler = optuna.samplers.BruteForceSampler(seed=42)
study.optimize(objective, n_jobs=2)
```

```
[I 2024-11-22 17:43:20,854] A new study created in memory with name: no-name-86ca63d9-e334-4237-a5a5-854dd2483f9f
[I 2024-11-22 17:43:20,855] Trial 0 finished with value: 1.0 and parameters: {'x': 0, 'y': 1}. Best is trial 0 with value: 1.0.
[I 2024-11-22 17:43:20,856] Trial 2 finished with value: 0.0 and parameters: {'x': 0, 'y': 0}. Best is trial 2 with value: 0.0.
[I 2024-11-22 17:43:21,861] Trial 1 finished with value: 1.0 and parameters: {'x': 1, 'y': 0}. Best is trial 2 with value: 0.0.
```

_No response_

porink0424:
## Motivation

Fixes #5780 

## Description of the changes

When running `BruteForceSampler` in parallel, there was a bug where a node being processed by one thread was forcibly changed to a leaf-state. This caused other threads to mistakenly assume there were no further nodes to explore beyond that point, leading to premature termination before all combinations were explored.

To address this issue, this PR introduces an `is_running` flag to the nodes. This flag allows distinguishing between nodes that are simply unexpanded and those that are unexpanded because they are currently being processed.


HideakiImamura:
@gen740 @contramundum53 Could you review this PR?

optuna/samplers/_brute_force.py
```diff
@@ -130,10 +149,14 @@ def objective(trial):
             that it is not recommended using this option in distributed optimization settings since
             this option cannot ensure the order of trials and may increase the number of duplicate
             suggestions during distributed optimization.
+        enforce_unique_samples:
+            If this is :obj:`True`, the sampler strictly avoids duplicate samples. Please note that
+            the sampler may stop before trying the entire search space when this option is enabled.
     """
 
-    def __init__(self, seed: int | None = None) -> None:
+    def __init__(self, seed: int | None = None, enforce_unique_samples: bool = False) -> None:
```
contramundum53, line 157: Setting `exclude_running = True` does not strictly avoid duplicate samples because we do not hold any lock on the storage (two workers may sample the same parameter at the same time and we have no way to strictly prevent that). So, a different name is preferred. How about "avoid_premature_stop"?

optuna/samplers/_brute_force.py
```diff
@@ -68,23 +72,38 @@ def add_path(
...
-            else sum(child.count_unexpanded() for child in self.children.values())
+            if self.children is None and (not exclude_running or not self.is_running)
+            else (
+                0
+                if self.children is None
+                else sum(
+                    child.count_unexpanded(exclude_running) for child in self.children.values()
+                )
+            )
         )
```
contramundum53, line 87: A better way to implement this function is 
```
        return max(0, (
            1
            if self.children is None
            else sum(child.count_unexpanded(exclude_running) for child in self.children.values())
        ) - int(self.is_running and exclude_running))
```
This would be advantageous when there is a running trial on already-expanded node.

contramundum53: Sorry, I noticed that this may cause premature stop in additional cases.

porink0424:
@nabenabe0928 
Thank you for your review, I updated the code. PTAL.

gen740 (approved):
LGTM!

Issue #5862, opened 2024-12-13
Tracking Issue: free-threaded Python support


porink0424:
### Motivation

This issue aims to track and summarize Optuna’s support for the experimental free-threaded mode (Python v3.13t) introduced in CPython 3.13.

### Description

The current status of Optuna’s compatibility with v3.13t is as follows:

### Sampler

| Sampler           | Support for 3.13t | Remarks                                                                                                            |
|-------------------|-------------------|--------------------------------------------------------------------------------------------------------------------|
| RandomSampler     | Yes               | -                                                                                                                  |
| NSGAIISampler     | Yes               | -                                                                                                                  |
| TPESampler        | Yes               | Dependent on NumPy, but NumPy already supports 3.13t                                                               |
| GridSampler       | Yes               | Same as above                                                                                                      |
| NSGAIIISampler    | Yes               | Same as above                                                                                                      |
| BruteForceSampler | No                | Despite the pure Python implementation, a `RuntimeError: dictionary changed size during iteration` was confirmed to occur when running in multi-threaded mode on Python 3.13t. The cause is currently under investigation ([Link](https://gist.github.com/porink0424/e43886aa85f7962cf4814703e1db99af) to reproduce the error without pytest-freethreaded). |
| CmaEsSampler      | Yes               | Dependent on NumPy and cmaes, but both already support 3.13t                                                       |
| GPSampler         | No                | Dependent on PyTorch, which is not yet supported on PyPI                                                           |
| QMCSampler        | No                | Dependent on SciPy, which is not yet supported on PyPI                                                             |

### Storage

| Storage              | Support for 3.13t | Remarks                                                                                                                                                                                                                                                                                                                              |
|-----------------------|-------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| InMemoryStorage      | Yes               | -                                                                                                                                                                                                                                                                                                                                    |
[...]

PR #5922, opened 2025-01-10
fix(storage): Fix threading lock logic

gen740:
## Motivation

As shown in #5862, journal storage sometimes fails on free-threaded Python. This is because the journal store accesses the shared data without getting the lock.

## Description of the changes

Increase the lock range in the journal storage.

### DIFF
--- a/optuna/storages/journal/_storage.py
+++ b/optuna/storages/journal/_storage.py
@@ -266,13 +266,13 @@ class JournalStorage(BaseStorage):
             self._sync_with_backend()
             trial_id = self._replay_result._last_created_trial_id_by_this_process
 
-        # Dump snapshot here.
-        if (
-            isinstance(self._backend, BaseJournalSnapshot)
-            and trial_id != 0
-            and trial_id % SNAPSHOT_INTERVAL == 0
-        ):
-            self._backend.save_snapshot(pickle.dumps(self._replay_result))
+            # Dump snapshot here.
+            if (
+                isinstance(self._backend, BaseJournalSnapshot)
+                and trial_id != 0
+                and trial_id % SNAPSHOT_INTERVAL == 0
+            ):
+                self._backend.save_snapshot(pickle.dumps(self._replay_result))
         return trial_id
 
     def set_trial_param(


optuna/storages/journal/_storage.py
```diff
@@ -266,13 +266,13 @@ def create_new_trial(self, study_id: int, template_trial: FrozenTrial | None = N
...
-            and trial_id % SNAPSHOT_INTERVAL == 0
-        ):
-            self._backend.save_snapshot(pickle.dumps(self._replay_result))
+            # Dump snapshot here.
+            if (
+                isinstance(self._backend, BaseJournalSnapshot)
+                and trial_id != 0
+                and trial_id % SNAPSHOT_INTERVAL == 0
+            ):
+                self._backend.save_snapshot(pickle.dumps(self._replay_result))
```
c-bata, line 275: `backend.save_snapshot` takes a relatively long time. How about acquiring the lock object while only calling `pickle.dumps`?

```suggestion
        # Dump snapshot here.
        if (
            isinstance(self._backend, BaseJournalSnapshot)
            and trial_id != 0
            and trial_id % SNAPSHOT_INTERVAL == 0
        ):
            with self._thread_lock:
                replay_result_pkl = pickle.dumps(self._replay_result)
            self._backend.save_snapshot(replay_result_pkl)
```

gen740: I think it's reasonable, but other parts of the code are pessimistic lock.

So I think it's better to follow the other code in this PR.

nabenabe0928:
@c-bata @not522 Could you review this PR?

c-bata (approved):
LGTM.

not522 (commented):
The change is LGTM, but I'm not confident it relates to the free-threaded Python. I confirmed that [the reported error](https://gist.github.com/porink0424/8b974b2d638cffc176ab38212bf0e391) in #5862 occurs in `3.13.1` and `3.13.1t`.
Is it okay to merge this PR as a ordinary bug fix?

gen740:
@not522 
In my environment, the reported error in #5862 does not occur after this change in both `3.13.1` and `3.13.1t`. Could you give me more details?

not522:
Yes, this PR fixes the error in the my both environments. So, the motivation described above is misleading?

> As shown in https://github.com/optuna/optuna/issues/5862, journal storage sometimes fails on free-threaded Python.

I think this PR is not limited to the free-threaded Python.

gen740:
I see, that sounds reasonable.

not522:
Thanks! My question is resolved, so I merge this PR.

not522 (approved):

PR #5937, opened 2025-01-23
Adjust test conditions to adress the problem of fragile tests

sawa3030:
## Motivation
ref: [Fix a fragile unit test, `test_check_distribution_suggest_loguniform[sqlite3]` #5797](https://github.com/optuna/optuna/issues/5797)
Instead of counting the number of the warnings, the test now verifies the presence of any warnings. This approach reduces fragility and makes the test more reliable.

## Description of the changes
- Update the test to verify whether any warnings are raised instead of counting the exact number of warnings, and
- Add TODO comments to investigate the root cause of the warnings in the future.

### DIFF
--- a/tests/trial_tests/test_trial.py
+++ b/tests/trial_tests/test_trial.py
@@ -74,9 +74,11 @@ def test_check_distribution_suggest_uniform(storage_mode: str) -> None:
             trial.suggest_uniform("x", 10, 20)
             trial.suggest_uniform("x", 10, 30)
 
+        # TODO(sawa3030): Investigate and resolve the issue causing multiple warnings
+        # to appear sporadically in the CI pipeline.
         # we expect exactly one warning (not counting ones caused by deprecation)
         warning_records = [r for r in record if r.category is not FutureWarning]
-        assert len(warning_records) == 1, [warning.message for warning in warning_records]
+        assert len(warning_records) >= 1, [warning.message for warning in warning_records]
 
         with pytest.raises(ValueError):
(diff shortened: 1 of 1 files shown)


nabenabe0928:
Conventionally, we use the imperative form for each verb in descriptions.

https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13

nabenabe0928:
I revised your description, so please take a look:)

## Motivation

Ref:
- https://github.com/optuna/optuna/issues/5797

Instead of counting the number of the warnings, the test now verifies the presence of any warnings.
This approach reduces fragility and makes the test more reliable.

> [!NOTE]
> The itemization of PR/issue URLs will look more stylish.

## Description of the changes

- Update the test to verify whether any warnings are raised instead of counting the exact number of warnings, and
- Add TODO comments to investigate the root cause of the warnings in the future.

> [!NOTE]
> I changed past tense to present.
> Plus, I changed the enumeration to the Oxford comma style.

nabenabe0928 (approved):
The changes look good to me, btw:)

sawa3030:
@nabenabe0928 Thank you for your comments. They are very helpful!

not522:
Thank you for your PR. But, wouldn't this change make it harder to investigate the cause? I think it is OK to keep the tests as is, what do you think?

not522:
I'm going to close this PR due to keep the tests as is, while we investigate the cause of error. Thank you for your understanding.

Issue #5921, opened 2025-01-09
`JournalStorage` with `JournalFileSymlinkLock` may fail to resume


not522:
### Expected behavior

`JournalStorage` should be able to resume even if the process terminates unexpectedly. However, if the process terminates suddenly, a lock file may remain and it may not be able to resume.

### Environment

- Optuna version:4.2.0.dev
- Python version:3.11.0
- OS:macOS-15.1-x86_64-i386-64bit
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
If we encounter the above bug, the resumed process will sleep forever.
```

### Steps to reproduce

1. Terminate the process suddenly (not Ctrl-C; it deletes the lock file correctly.)
2. Resume the optimization process

PR #5971, opened 2025-02-13
Prevent Lock Blocking by Adding Timeout to `JournalStorage`

sawa3030:
## Motivation
- Currently, if a process is terminated without releasing the lock, the next process fails to resume. This change introduces a mechanism to release the existing lock if acquiring it takes too long.
- ref: [JournalStorage with JournalFileSymlinkLock may fail to resume #5921](https://github.com/optuna/optuna/issues/5921)

## Description of the changes
- Introduced a grace period for acquiring locks in `JournalFileSymlinkLock` and `JournalFileOpenLock`.
- If the lock cannot be acquired within the grace period, it is considered stale and released to prevent blocking new processes.

### DIFF
--- a/optuna/storages/journal/_file.py
+++ b/optuna/storages/journal/_file.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import abc
 from collections.abc import Iterator
 from contextlib import contextmanager
+import datetime
 import errno
 import json
 import os
@@ -128,20 +129,25 @@ class JournalFileSymlinkLock(BaseJournalFileLock):
     Args:
         filepath:
             The path of the file whose race condition must be protected.
+        grace_period:
+            Grace period before an existing lock is forcibly released.
     """
 
-    def __init__(self, filepath: str) -> None:
+    def __init__(self, filepath: str, grace_period: int = 30) -> None:
         self._lock_target_file = filepath
         self._lock_file = filepath + LOCK_FILE_SUFFIX
+        if grace_period is not None and grace_period <= 0:
+            raise ValueError("The value of `grace_period` should be a positive integer.")
+        self.grace_period = grace_period
 
     def acquire(self) -> bool:
         """Acquire a lock in a blocking way by creating a symbolic link of a file.
 
         Returns:
             :obj:`True` if it succeeded in creating a symbolic link of ``self._lock_target_file``.
-
         """
         sleep_secs = 0.001
+        start_datetime = datetime.datetime.now()
         while True:
             try:
                 os.symlink(self._lock_target_file, self._lock_file)
@@ -150,6 +156,10 @@ class JournalFileSymlinkLock(BaseJournalFileLock):
                 if err.errno == errno.EEXIST:
                     time.sleep(sleep_secs)
                     sleep_secs = min(sleep_secs * 2, 1)
+                    if datetime.datetime.now() - start_datetime > datetime.timedelta(
+                        seconds=self.grace_period
+                    ):
+                        self.release()
                     continue
                 raise err
             except BaseException:
@@ -181,10 +191,15 @@ class JournalFileOpenLock(BaseJournalFileLock):
     Args:
         filepath:
             The path of the file whose race condition must be protected.
+        grace_period:
+            Grace period before an existing lock is forcibly released.
     """
 
-    def __init__(self, filepath: str) -> None:
+    def __init__(self, filepath: str, grace_period: int = 30) -> None:
         self._lock_file = filepath + LOCK_FILE_SUFFIX
+        if grace_period is not None and grace_period <= 0:
+            raise ValueError("The value of `grace_period` should be a positive integer.")
+        self.grace_period = grace_period
 
     def acquire(self) -> bool:
         """Acquire a lock in a blocking way by creating a lock file.
@@ -194,6 +209,7 @@ class JournalFileOpenLock(BaseJournalFileLock):
 
         """
         sleep_secs = 0.001
+        start_datetime = datetime.datetime.now()
         while True:
             try:
                 open_flags = os.O_CREAT | os.O_EXCL | os.O_WRONLY
@@ -203,6 +219,10 @@ class JournalFileOpenLock(BaseJournalFileLock):
                 if err.errno == errno.EEXIST:
                     time.sleep(sleep_secs)
                     sleep_secs = min(sleep_secs * 2, 1)
+                    if datetime.datetime.now() - start_datetime > datetime.timedelta(
+                        seconds=self.grace_period
+                    ):
+                        self.release()
                     continue
                 raise err
             except BaseException:
--- a/optuna/storages/journal/_file.py
+++ b/optuna/storages/journal/_file.py
@@ -136,7 +136,7 @@ class JournalFileSymlinkLock(BaseJournalFileLock):
     def __init__(self, filepath: str, grace_period: int = 30) -> None:
         self._lock_target_file = filepath
         self._lock_file = filepath + LOCK_FILE_SUFFIX
-        if grace_period is not None and grace_period <= 0:
+        if grace_period <= 0:
             raise ValueError("The value of `grace_period` should be a positive integer.")
         if grace_period < 3:
             warnings.warn("The value of `grace_period` might be too small. ")
@@ -162,10 +162,14 @@ class JournalFileSymlinkLock(BaseJournalFileLock):
                         if current_mtime != mtime:
                             mtime = current_mtime
                             last_update_time = time.monotonic()
-                        if time.monotonic() - last_update_time > self.grace_period:
-                            self.release()
                     except Exception:
                         pass
+                    if time.monotonic() - last_update_time > self.grace_period:
+                        warnings.warn(
+                            "The existing lock file has not been released for an extended period. "
+                            "Forcibly releasing the lock file."
+                        )
+                        self.release()
                     time.sleep(sleep_secs)
                     sleep_secs = min(sleep_secs * 2, 1)
                     continue
@@ -205,7 +209,7 @@ class JournalFileOpenLock(BaseJournalFileLock):
 
     def __init__(self, filepath: str, grace_period: int = 30) -> None:
         self._lock_file = filepath + LOCK_FILE_SUFFIX
-        if grace_period is not None and grace_period <= 0:
+        if grace_period <= 0:
             raise ValueError("The value of `grace_period` should be a positive integer.")
         if grace_period < 3:
             warnings.warn("The value of `grace_period` might be too small. ")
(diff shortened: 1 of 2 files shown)


y0z:

not522 (commented):
Thank you for your PR! Could you check the following points?
- How about using `time.monotonic()`? This differs from `datetime.now()` in that it is not affected by system clock updates.
- How about checking the last modified time of the lock? As the number of processes increases, it may continue to fail to acquire a lock.

sawa3030:
@not522 Thank you for your suggestions. There are some changes I made.
- I updated the logic to check the last modified time. The previous approach had issues when the grace period was too short, but with your suggested logic, it is now more robust regardless of the grace period length.
- I have confirmed that this PR successfully resolves the issue: [JournalStorage with JournalFileSymlinkLock may fail to resume #5921](https://github.com/optuna/optuna/issues/5921). After an unexpected process termination, the optimization process can now successfully resume even if the lock file still exists. It successfully resumes the optimization process whether a new process starts before or after the grace period has elapsed.
- I added tests to ensure that a customized grace period functions correctly without issues.

And here are some concerns
- As far as I know, `time.monotonic()` cannot be used for checking the modified time of a file. Do you think it would be better to keep the previous logic, as shown in the code below? Or would that make the implementation unnecessarily complex?
```
if min(datetime.datetime.now() - mtime, time.monotonic() - start_time) > datetime.timedelta(seconds=self.grace_period):
    self.release()
```
- [The documentation for st_mtime](https://docs.python.org/ja/3.13/library/os.html#os.stat_result.st_birthtime_ns) states that its resolution depends on the file system, and some file systems have a low resolution (e.g., 2-seconds). If the grace period is set too short, this could lead to unexpected behavior. I’ve added a warning if the grace period is set to less than 3 seconds, but I’m unsure whether this warning is necessary or if it just adds unnecessary complexity.

c-bata:

not522:
In my understanding, comparing `st_mtime` and the local time of the machine is unreliable (for example, NFS records the time of the server, not the machine where the process is running). But, we can trust whether it has been modified or not. So, how about the following logic?

```python
start_time = time.monotonic()
mtime = datetime.datetime.fromtimestamp(os.stat(self._lock_file).st_mtime)
while True:
    ...
    current_mtime = datetime.datetime.fromtimestamp(os.stat(self._lock_file).st_mtime)
    if current_mtime != mtime:
        start_time = time.monotonic()
        mtime = current_mtime
        continue
    if time.monotonic() - start_time > self.grace_period:
        self.release()
```

sawa3030:
Thank you for your suggestion! I've updated the logic accordingly.

sawa3030:
@not522 @y0z I’ve made some updates. Could you review the PR when you have time? Thank you!

tests/storages_tests/journal_tests/test_journal.py
```diff
@@ -28,33 +28,48 @@
...
+                        self.tempfile.name, self.grace_period
+                    )
+                else:
+                    lock = optuna.storages.journal.JournalFileSymlinkLock(self.tempfile.name)
             else:
                 raise Exception("Must not reach here")
             return optuna.storages.journal.JournalFileBackend(self.tempfile.name, lock)
         elif self.storage_type.startswith("redis"):
+            if self.grace_period is not None:
+                raise RuntimeError("Grace period is not supported for Redis storage.")
```
not522, line 72: How about just asserting `self.grace_period is None`?

not522 (commented):
Thank you for your update! Could you check my suggestions?

sawa3030:
@not522 Thank you for your advice! I've made some updates based on your feedback.

optuna/storages/journal/_file.py
```diff
@@ -128,26 +129,43 @@ class JournalFileSymlinkLock(BaseJournalFileLock):
...
         sleep_secs = 0.001
+        last_update_time = time.monotonic()
+        mtime = None
         while True:
             try:
                 os.symlink(self._lock_target_file, self._lock_file)
                 return True
             except OSError as err:
                 if err.errno == errno.EEXIST:
+                    try:
```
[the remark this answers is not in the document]
not522: I think that it is just for `self.release`, isn't it?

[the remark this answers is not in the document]
sawa3030: I overlooked that release() could also raise an error. To answer your quesetion, in my understanding, os.stat() can generate an [OSError](https://docs.python.org/ja/3.13/library/os.html#os.error) if the lock file is removed by another process and no longer exists.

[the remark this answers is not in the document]
not522: I see. How about clarifying the target of the try as follows:

```python
try:
    current_mtime = os.stat(self._lock_file).st_mtime
except OSError:
    continue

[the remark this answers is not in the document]
Comment (sawa3030) [reply optuna/storages/journal/_file.py:160]: I wasn’t aware that this code works without encountering a scope error. Thank you for your step by step suggestion🙇
```
try:
    current_mtime = os.stat(self._lock_file).st_mtime
except OSError:
    continue
if current_mtime != mtime:
    mtime = current_mtime
    last_update_time = time.monotonic()
```

File: optuna/storages/journal/_file.py [hunk 3]
```diff
@@ -194,13 +228,32 @@ def acquire(self) -> bool:
...
+                        current_mtime = os.stat(self._lock_file).st_mtime
+                    except OSError:
+                        continue
+                    if current_mtime != mtime:
+                        mtime = current_mtime
+                        last_update_time = time.monotonic()
```
y0z, line 246: I am confused about why we need these three variables: `last_update_time`, `mtime`, and `current_mtime`. At first glance, it seems that `last_update_time = current_mtime` since `current_mtime != mtime` means the lock file has been modified. Could you please provide additional explanation as to the difference between their roles?

optuna/storages/journal/_file.py
```diff
@@ -128,26 +129,52 @@ class JournalFileSymlinkLock(BaseJournalFileLock):
...
```
y0z, line 166: Ditto: https://github.com/optuna/optuna/pull/5971/files#r1973051800

optuna/storages/journal/_file.py
```diff
@@ -128,26 +129,52 @@ class JournalFileSymlinkLock(BaseJournalFileLock):
     Args:
         filepath:
             The path of the file whose race condition must be protected.
+        grace_period:
+            Grace period before an existing lock is forcibly released.
     """
 
-    def __init__(self, filepath: str) -> None:
+    def __init__(self, filepath: str, grace_period: int = 30) -> None:
```
y0z, line 136: What about allowing `grace_period = None` (i.e., unlimited)? Forcing the release of long-held locks seems like a good practice in many cases, but I'm not sure it's always OK.

optuna/storages/journal/_file.py
```diff
@@ -128,26 +129,52 @@ class JournalFileSymlinkLock(BaseJournalFileLock):
...
+                        mtime = current_mtime
+                        last_update_time = time.monotonic()
+
+                    if time.monotonic() - last_update_time > self.grace_period:
+                        warnings.warn(
+                            "The existing lock file has not been released "
+                            "for an extended period. Forcibly releasing the lock file."
+                        )
+                        try:
+                            self.release()
```
y0z, line 174: Would it be a good idea to reset `sleep_secs` to `0.001` after calling `release()`? Another thread may acquire the lock between your `release()` and `os.symlink()`.

optuna/storages/journal/_file.py
```diff
@@ -181,10 +208,17 @@ class JournalFileOpenLock(BaseJournalFileLock):
     Args:
         filepath:
             The path of the file whose race condition must be protected.
+        grace_period:
+            Grace period before an existing lock is forcibly released.
     """
 
-    def __init__(self, filepath: str) -> None:
+    def __init__(self, filepath: str, grace_period: int = 30) -> None:
```
y0z, line 215: Ditto: https://github.com/optuna/optuna/pull/5971/files#r1973119843

optuna/storages/journal/_file.py
sawa3030: `last_update_time` is based on monotonic clock time, whereas `current_mtime` and `m_time` are Epoch timestamps. I chose not to compare `mtime` directly with grace_period to avoid inconsistencies.

sawa3030: I feel like the name `last_update_time` is a little confusing, but I couldn't come up with a better variable name.

sawa3030:
@y0z I made the update. Thank you for your feedback!

optuna/storages/journal/_file.py
sawa3030: I change the variable name from `last_update_time` to `last_update_monotonic_time`!

sawa3030:
@y0z @not522 I think all updates have been applied. Thank you for your suggestions! Please let me know if there's anything else that needs to be addressed.

optuna/storages/journal/_file.py
```diff
@@ -128,26 +129,55 @@ class JournalFileSymlinkLock(BaseJournalFileLock):
...
+
+                        if time.monotonic() - last_update_monotonic_time > self.grace_period:
+                            warnings.warn(
+                                "The existing lock file has not been released "
+                                "for an extended period. Forcibly releasing the lock file."
+                            )
+                            try:
+                                self.release()
+                                sleep_secs = 0.001
+                            except OSError:
```
not522, line 178: Maybe it should be `RuntimeError`.

```suggestion
                            except RuntimeError:
```

not522 (commented):
Thank you. It's almost LGTM. Could you checking the following points?

- The error type
- Adding test for the invalid argument (i.e., zero and negative integer)

I think the warning message can be improved, but I think we can address in a separate PR.

y0z (approved):
LGTM

tests/storages_tests/journal_tests/test_journal.py
```diff
@@ -28,33 +28,43 @@
...
-    "file_with_link_lock",
-    "redis_default",
-    "redis_with_use_cluster",
+LOG_STORAGE_WITH_PARAMETER = [
+    ("file_with_open_lock", 3),
+    ("file_with_open_lock", 30),
+    ("file_with_open_lock", None),
+    ("file_with_link_lock", 3),
+    ("file_with_link_lock", 30),
+    ("file_with_link_lock", None),
```
not522, line 37: I think `grace_period=3` is redundant.

```suggestion
    ("file_with_open_lock", 30),
    ("file_with_open_lock", None),
    ("file_with_link_lock", 30),
    ("file_with_link_lock", None),
```

tests/storages_tests/journal_tests/test_journal.py
```diff
@@ -244,3 +258,18 @@ def test_raise_error_for_deprecated_class_import_from_journal() -> None:
...
+
+@pytest.mark.parametrize(
+    "log_storage_type,grace_period",
+    [
+        ("file_with_open_lock", 0),
+        ("file_with_open_lock", -1),
+        ("file_with_link_lock", 0),
+        ("file_with_link_lock", -1),
+    ],
+)
```
not522, line 271: It can be written more simply.

```suggestion
@pytest.mark.parametrize("log_storage_type", ("file_with_open_lock", "file_with_link_lock"))
@pytest.mark.parametrize("grace_period", (0, -1))
```

sawa3030:
@not522 Thank you for letting me know that. I updated the code as you suggested.

not522 (approved):
LGTM!

PR #5981, opened 2025-02-17
improve time complexity of IntersectionSearchSpace #5949

GittyHarsha:
## Motivation
Fix issue #5949 

## Description of the changes
Removed this line
https://github.com/optuna/optuna/blob/229fe58389a649ca90e7ee2205aeb3d648182235/optuna/search_space/intersection.py#L29

and adding filtering logic inside the for-loop

PR #5982, opened 2025-02-17
Improve time complexity of IntersectionSearchSpace #5949

### LINKED ISSUE #5949: `IntersectionSearchSpace` is slow
### Motivation

The time complexity of `IntersectionSearchSpace` is O(n_trials) for each trial. We can improve it to the constant time.

We filter trials here, but we can do it in the following for-loop.
https://github.com/optuna/optuna/blob/229fe58389a649ca90e7ee2205aeb3d648182235/optuna/search_space/intersection.py#L29

### Suggestion

We can move the filtering to the inside of the for-loop. The loop will finish immediately using `cached_trial_number`.

_No response_

GittyHarsha:
## Motivation
Fix #5949 

## Description of the changes
Removed this line 
https://github.com/optuna/optuna/blob/5a7760c8ee640fb95d3f7626056c0d71a282b202/optuna/search_space/intersection.py#L29

Added filtering logic inside the for-loop so that the loop breaks to avoid putting filter checks on the trials with `trial.number` less than `cached_trial_number`

### DIFF
--- a/optuna/search_space/intersection.py
+++ b/optuna/search_space/intersection.py
@@ -26,12 +26,15 @@ def _calculate(
     if include_pruned:
         states_of_interest.append(optuna.trial.TrialState.PRUNED)
 
-    trials_of_interest = [trial for trial in trials if trial.state in states_of_interest]
+    next_cached_trial_number = -1
 
-    next_cached_trial_number = (
-        trials_of_interest[-1].number + 1 if len(trials_of_interest) > 0 else -1
-    )
-    for trial in reversed(trials_of_interest):
+    for trial in reversed(trials):
+        if trial.state not in states_of_interest:
+            continue
+
+        if next_cached_trial_number == -1:
+            next_cached_trial_number = trial.number + 1
+
         if cached_trial_number > trial.number:
             break
 


not522:
Thank you for your PR!
This is just a comment unrelated to this PR, but updating your branch (GittyHarsha:master) will update the PR, so there's no need to close the previous PR #5981! (but let's keep working on it here!)

c-bata:
@sawa3030 @not522 Could you review this PR?

GittyHarsha:
Thanks. I'll keep that in mind.

optuna/search_space/intersection.py
```diff
@@ -26,12 +26,18 @@ def _calculate(
...
-    trials_of_interest = [trial for trial in trials if trial.state in states_of_interest]
+    next_cached_trial_number = -1
+
+    found_trial_of_interest = False
+
+    for trial in reversed(trials):
+        if trial.state not in states_of_interest:
+            continue
+
+        elif not found_trial_of_interest:
```
not522, line 37: How about checking `next_cached_trial_number` is equal to `-1`? It will be simpler than using `found_trial_of_interest`.

```suggestion
        if next_cached_trial_number == -1:
```

GittyHarsha: Using a Boolean flag is faster especially when repeatedly applied in a loop but reduces readability?

not522 (approved):
LGTM! I confirmed that this PR will speed up the following code.

```python
import optuna

def objective(trial):
    return sum(trial.suggest_float(f"x{i}", 0, 1) ** 2 for i in range(10))

study = optuna.create_study(sampler=optuna.samplers.CmaEsSampler(seed=42))
```

### master

```
real	1m15.972s
user	1m14.434s
sys	0m1.037s
```

### PR

```
real	1m8.455s
user	1m7.028s
sys	0m1.076s
```

sawa3030:
LGTM! I tested this in my local environment and confirmed that the PR will speed up the code @not522 provided above.

**master**
real    3m39.997s
user    3m39.561s
sys     0m2.473s

**PR**
real    3m23.284s
user    3m19.046s
sys     0m5.473s

c-bata:
Let me merge this PR since this PR got two approvals. @GittyHarsha  Thank you for your contribution!

PR #5983, opened 2025-02-18
code fix: #5806

mehakmander11:
changed error statement in study.py

Ref:
- https://github.com/optuna/optuna/issues/5806

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1252,8 +1252,7 @@ def create_study(
         for d in directions
     ):
         raise ValueError(
-            "Please set either 'minimize' or 'maximize' to direction. You can also set the "
-            "corresponding `StudyDirection` member."
+            "It looks like you passed a list to 'direction'. For multi-objective optimization, use 'directions' instead of 'direction'."
         )
 
     direction_objects = [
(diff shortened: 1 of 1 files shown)


mehakmander11:
alright , i will update code(error) it in more proficient way

mehakmander11:
i have done please do check , if still not , i will do it again,
no worries, i am here to contribute as much i can😊

c-bata:
@nabenabe0928 @gen740 Could you review this PR?

optuna/study/study.py
```diff
@@ -1252,8 +1252,8 @@ def objective(trial):
         for d in directions
     ):
         raise ValueError(
-            "Please set either 'minimize' or 'maximize' to direction. You can also set the "
-            "corresponding `StudyDirection` member."
+            "'direction' must be 'minimize' or 'maximize'."
+            " For multiple objectives, use 'directions'."
```
nabenabe0928, line 1256: The problem here happens when users pass `directions` instead of `direction`.
This statement is triggered when `directions` is passed anyways, so it would be better to say that users used `directions` in a wrong way instead of just saying `use directions`.
 
```suggestion
            f"`directions` must be a list of `minimize` or `maximize`, but got {directions}. "
            "For single-objective optimization, please use `direction` instead of `directions`."
```

Issue #5948, opened 2025-01-28
`_pop_waiting_trial_id` for `InMemoryStorage` is slow


not522:
### Motivation

`_pop_waiting_trial_id` for `InMemoryStorage` becomes slow with a large number of trials. We can increase the speed with caching the number of the previous trial with waiting state.

The main cause of this issue is below.
https://github.com/optuna/optuna/blob/229fe58389a649ca90e7ee2205aeb3d648182235/optuna/storages/_in_memory.py#L380

ref: #4716

### Suggestion

We can introduce something like `previous_waiting_trial_number` to cache the number of the previous waiting trial.

PR #5993, opened 2025-02-25
Add `_prev_waiting_trial_number` in `InMemoryStorage` to improve the efficiency of `_pop_waiting_trial_id`

sawa3030:
## Motivation
- Fix [_pop_waiting_trials for InMemoryStorage is slow #5948](https://github.com/optuna/optuna/issues/5948)

## Description of the changes
- Introduced `_prev_waiting_trial_id` in InMemoryStorage to improve efficiency of `get_all_trials(states = (TrialState.WAITING))`.

### DIFF
--- a/optuna/storages/_in_memory.py
+++ b/optuna/storages/_in_memory.py
@@ -380,15 +380,12 @@ class InMemoryStorage(BaseStorage):
 
             if states == (TrialState.WAITING,):
                 trials: list[FrozenTrial] = []
-                for i in range(
-                    self._prev_waiting_trial_number[study_id], len(self._studies[study_id].trials)
-                ):
-                    trial = self.get_trial(
-                        self.get_trial_id_from_study_id_trial_number(study_id, i)
-                    )
+                for trial in self._studies[study_id].trials[
+                    self._prev_waiting_trial_number[study_id] :
+                ]:
                     if trial.state == TrialState.WAITING:
                         if not trials:
-                            self._prev_waiting_trial_number[study_id] = i
+                            self._prev_waiting_trial_number[study_id] = trial.number
                         trials.append(trial)
                 if not trials:
                     self._prev_waiting_trial_number[study_id] = len(self._studies[study_id].trials)
--- a/optuna/storages/_in_memory.py
+++ b/optuna/storages/_in_memory.py
@@ -55,7 +55,7 @@ class InMemoryStorage(BaseStorage):
         self._max_trial_id = -1
 
         self._lock = threading.RLock()
-        self._prev_waiting_trial_id = 0
+        self._prev_waiting_trial_number: dict[int, int] = {}
 
     def __getstate__(self) -> dict[Any, Any]:
         state = self.__dict__.copy()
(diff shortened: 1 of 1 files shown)


sawa3030:
## Performance Improvement
This change improves performance by reducing the overhead of fetching all trials. Below are benchmark results comparing the current master branch.

**Before (master)**
image

**After (PR)**
image

```
import optuna
import cProfile

study = optuna.create_study()

def objective(trial):
    y = trial.suggest_uniform("y", -10, 10)
    return x**2 + y**2

for i in range(-5, 5):
    study.enqueue_trial({"x": i, "y": i})

def profile_objective():

cProfile.run('profile_objective()', 'journal_storage_master_100.prof')
# cProfile.run('profile_objective()', 'journal_storage_fix_100.prof')
```

HideakiImamura:

optuna/storages/_in_memory.py
```diff
@@ -55,6 +55,7 @@ def __init__(self) -> None:
         self._max_trial_id = -1
 
         self._lock = threading.RLock()
+        self._prev_waiting_trial_id = 0
```
not522, line 58: Could you make it so that we can handle multiple studies using a `dict`?

optuna/storages/_in_memory.py
```diff
@@ -375,9 +376,24 @@ def get_all_trials(
         with self._lock:
             self._check_study_id(study_id)
 
-            trials = self._studies[study_id].trials
-            if states is not None:
-                trials = [t for t in trials if t.state in states]
+            if states == (TrialState.WAITING,):
+                trials: list[FrozenTrial] = []
+                for i in range(self._prev_waiting_trial_id, self._max_trial_id + 1):
```
not522, line 381: I prefer caching `trial.number` over the `trial_id` because the trial numbers are sequential on trials.

sawa3030: Thank you for your advice! If I understand the logic correctly, using `trial.number` also eliminates the need to catch a KeyError.

not522: `self._studies[study_id].trials` has trials numbered from `0` to `len(trials)-1`. So, we can write it like this: `self._studies[study_id].trials[min_number:]` (without catching `KeyError`).

sawa3030:
Let me show you the performance improvement with various examples.

not522:
Could you share benchmarks with larger `n_trials` (e.g., 10,000 or 50,000)? When it is small, `_pop_waiting_trial_id` is fast enough to be ignored.

sawa3030:
Sure. But let me do it with n_trials = 10,000 and for a single iteration since my PC isn't powerful enough and doing optimization with larger number of n_trials needs significant amount of time

sawa3030:
@not522 I added the data to [the comment above](https://github.com/optuna/optuna/pull/5993#issuecomment-2687361249). The improvement is significant—total execution time for `pop_waiting_trials` (10,000 runs) dropped from 24.929083 seconds to just 0.391722 seconds.

optuna/storages/_in_memory.py
```diff
@@ -375,9 +378,22 @@ def get_all_trials(
         with self._lock:
             self._check_study_id(study_id)
 
-            trials = self._studies[study_id].trials
-            if states is not None:
-                trials = [t for t in trials if t.state in states]
+            if states == (TrialState.WAITING,):
```
not522, line 381: Could you adding a comment that this is for speeding up?

not522 (approved):

HideakiImamura (approved):
LGTM.

PR #5995, opened 2025-02-27
Fix a minor bug in GPSampler for objective that returns `inf`

nabenabe0928:
## Motivation

This PR fixes a minor bug in `GPSampler`.
The bug of my interest was embedded here:
- https://github.com/optuna/optuna/commit/84502dcdf5747cc128a4caa1573c4bac9efb48be#diff-fe893dd86f2d66f2d847ab31510f07079718c70f104067a431616cd4117214b8

image

In principle, this change does not affect most users because this statement is triggered only if a user-defined objective function returns non-finite values.

The bug was originally introduced because of the following misunderstanding:
- whether to give the `initial` argument in `np.max` and `np.min` does not affect the result as long as the array of interest is not empty.

However, this is not actually true.
For example, `np.max([1], initial=10)` gives us `10` instead of `1`, which we would like to yield in fact.

To this end, I address this issue by modifying the corresponding part.
Note that `emmr` also needs to be fixed due to the same issue.

## Description of the changes

- Address the issue stated above

### DIFF
--- a/tests/gp_tests/test_gp.py
+++ b/tests/gp_tests/test_gp.py
@@ -13,8 +13,11 @@ import optuna._gp.prior as prior
 @pytest.mark.parametrize(
     "values,ans",
     [
-        (np.array([-1, 0, 1]), np.array([-1, 0, 1])),
-        (np.array([-1, -np.inf, 0, np.inf, 1]), np.array([-1, -1, 0, 1, 1])),
+        (np.array([-1, 0, 1])[:, np.newaxis], np.array([-1, 0, 1])[:, np.newaxis]),
+        (
+            np.array([-1, -np.inf, 0, np.inf, 1])[:, np.newaxis],
+            np.array([-1, -1, 0, 1, 1])[:, np.newaxis]
+        ),
         (np.array([[-1, 2], [0, -2], [1, 0]]), np.array([[-1, 2], [0, -2], [1, 0]])),
         (
             np.array([[-1, 2], [-np.inf, np.inf], [0, -np.inf], [np.inf, -2], [1, 0]]),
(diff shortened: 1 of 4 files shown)


HideakiImamura:
@c-bata @gen740 @sawa3030 Could you review this PR?

sawa3030:
I believe the previous logic computed the maximum over all elements in `values`, whereas the new implementation takes the maximum along `axis=0`. When `values.shape = (n, m)` with `m >= 2`, this change affects the return values. For example, if `values = [[inf, 10], [inf, 10]]`, the previous implementation would return `[[10, 10], [10, 10]]`, while the new implementation returns `[[0, 10], [0, 10]]`. [Here](https://github.com/optuna/optuna/blob/master/optuna/samplers/_gp/sampler.py#L154), `consstraint_vals` can have multiple columns when multiple constraints are provided, which I think may lead to unintended changes in behavior.

nabenabe0928:
<details>

```python
import numpy as np


def original(values: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
    finite_vals = values[np.isfinite(values)]
    best_finite_val = np.max(finite_vals, axis=0, initial=0.0)
    worst_finite_val = np.min(finite_vals, axis=0, initial=0.0)
    return best_finite_val, worst_finite_val


def this_pr(values: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
    finite_vals_with_nan = np.where(np.isfinite(values), values, np.nan)
    is_any_finite = np.any(np.isfinite(finite_vals_with_nan), axis=0)
    best_finite_vals = np.where(is_any_finite, np.nanmax(finite_vals_with_nan, axis=0), 0.0)
    worst_finite_vals = np.where(is_any_finite, np.nanmin(finite_vals_with_nan, axis=0), 0.0)
    return best_finite_vals, worst_finite_vals
```
</details>

nabenabe0928:
> I believe the previous logic computed the maximum over all elements in `values`, whereas the new implementation takes the maximum along `axis=0`. When `values.shape = (n, m)` with `m >= 2`, this change affects the return values. For example, if `values = [[inf, 10], [inf, 10]]`, the previous implementation would return `[[10, 10], [10, 10]]`, while the new implementation returns `[[0, 10], [0, 10]]`. [Here](https://github.com/optuna/optuna/blob/master/optuna/samplers/_gp/sampler.py#L154), `consstraint_vals` can have multiple columns when multiple constraints are provided, which I think may lead to unintended changes in behavior.

@sawa3030 
In fact, this is a good point, and it seems the original implementation embedded the bug you pointed out.

First of all, we expect the shapes of `best_finite_vals` and `worst_finite_vals` to be `(m, )` while we actually get the shape of `(1, )`, which is already incorrect based on the verification code:
- https://github.com/optuna/optuna/pull/5995#issuecomment-2702514716

So we should probably add some unit tests for this part.

nabenabe0928:
@sawa3030 
I added some unit tests to ensure the code behavior, PTAL")

sawa3030:
Thank you for the explanation and for adding the unit tests. LGTM

tests/gp_tests/test_gp.py
```diff
@@ -6,9 +6,42 @@
...
+        ),
+        (np.array([-np.inf, np.inf]), np.array([0, 0])),
+        (np.array([]), np.array([])),
+    ],
+)
+def test_warn_and_convert_inf(values: np.ndarray, ans: np.ndarray) -> None:
+    assert np.allclose(warn_and_convert_inf(values), ans)
+    if len(values.shape) == 1:
+        # Test also with the shape of (n, 1) to ensure the batched version.
+        assert np.allclose(warn_and_convert_inf(values[:, np.newaxis]), ans[:, np.newaxis])
```
c-bata, line 42: This test code would be more readable if it were split into two test cases: one for single-dimensional arrays and another for two-dimensional arrays. What do you think?

c-bata (commented):
@nabenabe0928 Changes look almost good to me. I left one comment though.

tests/gp_tests/test_gp.py
nabenabe0928: I addressed your comment, PTAL!

c-bata (approved):
LGTM after CI passes!

gen740 (approved):
LGTM!

PR #5997, opened 2025-02-28
Fix `TPESampler._ParzenEstimator._calculate_categorical_distributions` when `consider_prior is False`

sawa3030:
## Motivation
- Fix `TPESampler._ParzenEstimator._calculate_categorical_distribution` to ensure the correct behavior when `consider_prior is False`

## Description of the changes
- Set the initial weights to 0 when the `consider_prior is False`
- Update the corresponding tests to reflect this change.

### DIFF
--- a/optuna/samplers/_tpe/parzen_estimator.py
+++ b/optuna/samplers/_tpe/parzen_estimator.py
@@ -191,13 +191,22 @@ class _ParzenEstimator:
             )
 
         n_kernels = len(observations) + parameters.consider_prior
-        assert parameters.prior_weight is not None
-        weights = np.full(
-            shape=(n_kernels, n_choices),
-            fill_value=parameters.prior_weight / n_kernels,
-        )
+        if parameters.consider_prior is False:
+            weights = np.full(
+                shape=(n_kernels, n_choices),
+                fill_value=0.0,
+            )
+        else:
+            assert parameters.prior_weight is not None
+            weights = np.full(
+                shape=(n_kernels, n_choices),
+                fill_value=parameters.prior_weight / n_kernels,
+            )
         observed_indices = observations.astype(int)
-        if param_name in parameters.categorical_distance_func:
+        if (
+            param_name in parameters.categorical_distance_func
+            and parameters.prior_weight is not None
+        ):
             # TODO(nabenabe0928): Think about how to handle combinatorial explosion.
             # The time complexity is O(n_choices * used_indices.size), so n_choices cannot be huge.
             used_indices, rev_indices = np.unique(observed_indices, return_inverse=True)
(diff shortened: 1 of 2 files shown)


HideakiImamura:

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -191,13 +191,22 @@ def _calculate_categorical_distributions(
...
+        else:
+            weights = np.full(
+                shape=(n_kernels, n_choices),
+                fill_value=0.0,
+            )
         observed_indices = observations.astype(int)
-        if param_name in parameters.categorical_distance_func:
+        if (
+            param_name in parameters.categorical_distance_func
+            and parameters.prior_weight is not None
```
nabenabe0928, line 208: [Question] Why do we need this condition?
More specifically, isn't it better to raise an error when we get `prior_weight = None` and `categorical_distance_func` simultaneously?

optuna/samplers/_tpe/parzen_estimator.py
```diff
@@ -191,13 +191,22 @@ def _calculate_categorical_distributions(
...
+            assert parameters.prior_weight is not None
+            weights = np.full(
+                shape=(n_kernels, n_choices),
+                fill_value=parameters.prior_weight / n_kernels,
+            )
+        else:
+            weights = np.full(
+                shape=(n_kernels, n_choices),
+                fill_value=0.0,
+            )
```
nabenabe0928, line 204: I think we can merge these two by defining a variable for `fill_value`.

nabenabe0928 (commented):
Thank you for the PR, I left two comments:)

optuna/samplers/_tpe/parzen_estimator.py
sawa3030: I realized that my previous implementation might not be entirely correct. There could be cases where `categorical_distance_func` is used even when `consider_prior` is False, which means `prior_weight` could be None. To address this, I change the logic to set `coef` to 1.0 when `consider_prior` is False. I've updated the logic accordingly—could you take a look and let me know if there's anything I misunderstood?

optuna/samplers/_tpe/parzen_estimator.py
sawa3030: I made the update accordingly!

sawa3030:
@nabenabe0928 When using the categorical distribution with `consider_prior = True`, some weights are set to zero. This might be problematic because categories that were not selected initially will have a very low probability of being chosen in subsequent selections. What could be the solution to this? Would it be appropriate to raise a warning when `consider_prior` is set to True?

sawa3030:
Because of the issues discussed above, `consider_prior` option will be deprecated. The logic will be fixed to always assume consider_prior is True. Therefore, this PR will be closed.

Issue #6002, opened 2025-03-04
S3-based Journal Backend


amitani:
### Motivation

Our use case doesn't require high-frequency read/writes, and some latency when obtaining suggestions is acceptable. We can use RDS, Redis or NFS, but they all come with some infrastructure management burden. We already use S3 for data and artifacts, so it will be much simpler if we can store the study on AWS S3.

### Description

With conditional writes and read-after-write consistency, AWS S3 provides enough mutual exclusion mechanisms to implement a `BaseJournalBackend` subclass.

https://aws.amazon.com/about-aws/whats-new/2024/08/amazon-s3-conditional-writes/
https://aws.amazon.com/blogs/aws/amazon-s3-update-strong-read-after-write-consistency/

c-bata:
Thank you again for your huge effort! Let me close this issue due to the reason I commented at https://github.com/optuna/optuna/pull/6003#issuecomment-2713208652.

PR #6003, opened 2025-03-05
Add `JournalS3Backend`

amitani:
## Motivation

#6002 

## Description of the changes

Add `JournalS3Backend`, and add a serialized test using `moto`.
Unfortunately `moto` doesn't support conditional writes, so the tests fail with multi-threading and multi-processing. I confirmed the same tests pass with an actual AWS S3 bucket.

### DIFF
--- a/docs/source/reference/storages.rst
+++ b/docs/source/reference/storages.rst
@@ -31,6 +31,7 @@ optuna.storages.journal
 
    journal.JournalFileBackend
    journal.JournalRedisBackend
+   journal.JournalS3Backend
 
 Users can flexibly choose a lock object for :class:`~optuna.storages.journal.JournalFileBackend` and here is the list of supported lock objects:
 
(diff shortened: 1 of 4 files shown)


nabenabe0928:

c-bata:
@amitani Thank you for your contribution and for taking the time to create this PR! We truly appreciate your efforts in adding AWS S3 as a backend for Optuna storage. However, there are a few concerns that need to be addressed:

1. As mentioned in the description, moto lacks sufficient feature support to test these scenarios. This means we cannot ensure the stability of this feature, as potential issues may go undetected in CI if any degradation occurs in the future.
2. JournalStorage frequently performs read and write operations, which may not be ideal for object storage like AWS S3. Generally, we recommend using JournalFileBackend and then upload the generated file to AWS S3.

Given these factors, I discussed this with other Optuna maintainers, and we decided to close this PR at this time. However, we would be delighted if you consider publishing this as a third-party library for Optuna instead. Thanks again for your understanding and contributions! :100:

amitani:
@c-bata Thank you for the feedback! The two concerns are valid and I agree it's reasonable not to merge this to optuna for now. Thanks to the modular design of the journal storages, having this as a third-party library should work fine. I'll further test this approach, and will consider publishing it later.

whatnick:
Is there an S3JournalBackend Plugin in the works or public on GitHub / PyPI. Would be interested in trying it and maintaining it.

PR #6004, opened 2025-03-07
Fix a bug that a gRPC server doesn't work with JournalStorage

fusawa-yugo:
## Motivation

Currently, the gRPC server fails to function properly when using JournalStorage.

## Description of the changes

The `RepeatedScalarContainer` class is used to convey arrays for protobuf. 
However, an instance of `RepeatedScalarContainer` cannot be converted to JSON. 
To address this, I made the following change:
```
values = list(request.values)
```

Additionally, to prevent issues like this in the future, I added `grpc_journal_file` to the `STORAGE_MODES` for pytest:
```
STORAGE_MODES: list[Any] = [
    "inmemory",
    "sqlite",
    "cached_sqlite",
    "journal",
    "journal_redis",
    "grpc_rdb",
    "grpc_journal_file",
]
```

### DIFF
--- a/tests/study_tests/test_study.py
+++ b/tests/study_tests/test_study.py
@@ -1036,7 +1036,7 @@ def test_optimize_infinite_budget_progbar() -> None:
 
 @pytest.mark.parametrize("storage_mode", STORAGE_MODES)
 def test_get_trials(storage_mode: str) -> None:
-    if storage_mode == "grpc":
+    if storage_mode == "grpc_rdb" or storage_mode == "grpc_journal_file":
         pytest.skip("gRPC storage doesn't use `copy.deepcopy`.")
 
     with StorageSupplier(storage_mode) as storage:
--- a/tests/study_tests/test_study.py
+++ b/tests/study_tests/test_study.py
@@ -1651,7 +1651,7 @@ def test_tell_from_another_process() -> None:
 
 @pytest.mark.parametrize("storage_mode", STORAGE_MODES)
 def test_pop_waiting_trial_thread_safe(storage_mode: str) -> None:
-    if "sqlite" == storage_mode or "cached_sqlite" == storage_mode or "grpc" == storage_mode:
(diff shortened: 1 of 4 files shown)


y0z:
@c-bata @nabenabe0928 Could you review this PR?

c-bata (approved):
LGTM! I confirmed that this PR allows the following code to work correctly.

```
$ python3 -c "import optuna; from optuna.storages import run_grpc_proxy_server; run_grpc_proxy_server(storage=optuna.storages.JournalStorage(optuna.storages.journal.JournalFileBackend('./tmp/journal.log')), host='localhost', port=13000)"
```

```python
import optuna


def objective(trial):
    return x ** 2


if __name__ == "__main__":
    study = optuna.create_study(
        study_name="distributed-example",
        storage=optuna.storages.GrpcStorageProxy(host="localhost", port=13000),
        load_if_exists=True,
    )
```

tests/study_tests/test_study.py
```diff
@@ -1036,7 +1036,7 @@ def terminate_study(study: Study, trial: FrozenTrial) -> None:
 
 @pytest.mark.parametrize("storage_mode", STORAGE_MODES)
 def test_get_trials(storage_mode: str) -> None:
-    if storage_mode == "grpc":
+    if storage_mode == "grpc_rdb" or storage_mode == "grpc_journal_file":
```
nabenabe0928, line 1039:
```suggestion
    if storage_mode in ("grpc_rdb", "grpc_journal_file"):
```

nabenabe0928 (commented):
Thank you for the PR!
I confirmed that the modification works as intended!
Please apply my comments and then I will approve this PR!

fusawa-yugo:
@nabenabe0928 
Thank you for your review!!
I revised them. Plus, I have corrected the parts I previously overlooked.
Could you check it again?

nabenabe0928 (approved):
Thank you for the updates, LGTM!

PR #6005, opened 2025-03-10
Deprecate `consider_prior` in `TPESampler`

sawa3030:
## Motivation
In `TPESampler`, when `consider_prior` is set to `False`, samples that were not selected during the initial exploration are unlikely to be chosen in subsequent samplings. This behavior is undesirable, so `consider_prior` should be deprecated and always set to `True`.  

## Description of the changes
- Deprecation warning is added to the code and the docstring of `TPESampler`.
    - `consider_prior` constructs `_ParzenEstimatorParameters` in the init function of `TPESampler`, so ideally, deprecation warning is generated in the `_ParzenEstimatorParameters`. However, [`NamedTuple` doesn’t support overriding __new__](https://github.com/python/typing/issues/526), and the adding a deprecation warning there would require extensive refactoring. Therefore, the warning is instead placed in `TPESampler`.  
    - In addition to that, field names in NamedTuple cannot start with an underscore, and changing the field name to `_consider_prior` is not allowed in `_ParzenEstimatorParameters`.
- `consider_prior` will be fixed to `True` in future PRs.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -10,6 +10,7 @@ import warnings
 
 import numpy as np
 
+from optuna import _deprecated
 from optuna._experimental import warn_experimental_argument
 from optuna._hypervolume import compute_hypervolume
 from optuna._hypervolume.hssp import _solve_hssp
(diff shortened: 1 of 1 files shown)


y0z:

HideakiImamura (approved):
Looks very good.

y0z (approved):
LGTM

PR #6008, opened 2025-03-12
Enhance SBXCrossover

hrntsm:
## Motivation

I would like to reduce the difference in the distribution of children between the reference paper and optuna's SBX as shown in [this discussion](https://github.com/optuna/optuna/discussions/6000)

## Description of the changes

The fixed values of “establishment” and “probability” are used as arguments to reproduce the same distribution as in the paper.

I have two questions.

1. Would this change be better for Optunahub?
2. Although `establishment` and `probability` are names that originally appeared in comments in the code, I think that expressions such as `variable_crossover_prob`, for example, would be easier to understand.

### DIFF
--- a/optuna/samplers/nsgaii/_crossovers/_sbx.py
+++ b/optuna/samplers/nsgaii/_crossovers/_sbx.py
@@ -111,7 +111,7 @@ class SBXCrossover(BaseCrossover):
         c1 = 0.5 * ((xs_min + xs_max) - betaq1 * xs_diff)  # Equation (4).
         c2 = 0.5 * ((xs_min + xs_max) + betaq2 * xs_diff)  # Equation (5).
 
-        # SBX applies crossover with use_chile_gene_probability and uniform_crossover_probability.,
+        # SBX applies crossover with use_child_gene_prob and uniform_crossover_prob.
         # the gene of the parent individual is the gene of the child individual.
         # The original SBX creates two child individuals,
         # but optuna's implementation creates only one child individual.
--- a/optuna/samplers/nsgaii/_crossovers/_sbx.py
+++ b/optuna/samplers/nsgaii/_crossovers/_sbx.py
@@ -19,6 +19,12 @@ class SBXCrossover(BaseCrossover):
     Generates a child from two parent individuals
     according to the polynomial probability distribution.
 
+    In the paper, SBX has only one argument, ``eta``,
+    and generate two child individuals.
+    However, Optuna can only return one child individual in one crossover operation,
+    so it uses the ``uniform_crossover_prob`` and ``use_child_gene_prob`` arguments
+    to make two individuals into one.
+
     - `Deb, K. and R. Agrawal.
       “Simulated Binary Crossover for Continuous Search Space.”
       Complex Syst. 9 (1995): n. pag.
@@ -29,18 +35,19 @@ class SBXCrossover(BaseCrossover):
             Distribution index. A small value of ``eta`` allows distant solutions
             to be selected as children solutions. If not specified, takes default
             value of ``2`` for single objective functions and ``20`` for multi objective.
-        establishment:
-            ``establishment`` is the probability of uniform crossover
+        uniform_crossover_prob:
+            ``uniform_crossover_prob`` is the probability of uniform crossover
             between two individuals selected as candidate child individuals.
-            Optuna returns only one child individual at a time, while SBX crossover generates two.
             This argument is whether or not two individuals are
             crossover to make one child individual.
             If not specified, takes default value of ``0.5``.
             The range of values is ``[0.0, 1.0]``.
-        probability:
-            ``probability`` is the probability of using the value of the generated
+        use_child_gene_prob:
+            ``use_child_gene_prob`` is the probability of using the value of the generated
             child variable rather than the value of the parent.
-            where ``1-probability`` is the probability of using the parent's values as it is.
+            This probability is applied to each variable individually.
+            where ``1-use_chile_gene_prob`` is the probability of
+            using the parent's values as it is.
             If not specified, takes default value of ``0.5``.
             The range of values is ``(0.0, 1.0]``.
     """
@@ -48,16 +55,21 @@ class SBXCrossover(BaseCrossover):
     n_parents = 2
 
     def __init__(
-        self, eta: float | None = None, establishment: float = 0.5, probability: float = 0.5
+        self,
+        eta: float | None = None,
+        uniform_crossover_prob: float = 0.5,
+        use_child_gene_prob: float = 0.5,
     ) -> None:
         self._eta = eta
 
-        if establishment < 0.0 or establishment > 1.0:
-            raise ValueError("The value of `establishment` must be in the range [0.0, 1.0].")
-        if probability <= 0.0 or probability > 1.0:
-            raise ValueError("The value of `probability` must be in the range (0.0, 1.0].")
-        self._establishment = establishment
-        self._probability = probability
+        if uniform_crossover_prob < 0.0 or uniform_crossover_prob > 1.0:
+            raise ValueError(
+                "The value of `uniform_crossover_prob` must be in the range [0.0, 1.0]."
+            )
+        if use_child_gene_prob <= 0.0 or use_child_gene_prob > 1.0:
+            raise ValueError("The value of `use_child_gene_prob` must be in the range (0.0, 1.0].")
+        self._uniform_crossover_prob = uniform_crossover_prob
+        self._use_child_gene_prob = use_child_gene_prob
 
     def crossover(
         self,
@@ -99,39 +111,33 @@ class SBXCrossover(BaseCrossover):
         c1 = 0.5 * ((xs_min + xs_max) - betaq1 * xs_diff)  # Equation (4).
         c2 = 0.5 * ((xs_min + xs_max) + betaq2 * xs_diff)  # Equation (5).
 
-        # SBX applies crossover with establishment, and with probability,
+        # SBX applies crossover with use_chile_gene_probability and uniform_crossover_probability.,
         # the gene of the parent individual is the gene of the child individual.
         # The original SBX creates two child individuals,
         # but optuna's implementation creates only one child individual.
         # Therefore, when there is no crossover,
         # the gene is selected with equal probability from the parent individuals x1 and x2.
 
-        index_prob = rng.rand()
-        child_params_list = []
+        child1_params_list = []
+        child2_params_list = []
 
         for c1_i, c2_i, x1_i, x2_i in zip(c1, c2, parents_params[0], parents_params[1]):
-            if rng.rand() < self._probability:
-                if rng.rand() < self._establishment:
-                    if index_prob >= 0.5:
-                        child_params_list.append(c1_i)
-                    else:
-                        child_params_list.append(c2_i)
+            if rng.rand() < self._use_child_gene_prob:
+                if rng.rand() >= self._uniform_crossover_prob:
+                    child1_params_list.append(c1_i)
+                    child2_params_list.append(c2_i)
                 else:
-                    if index_prob >= 0.5:
-                        child_params_list.append(c2_i)
-                    else:
-                        child_params_list.append(c1_i)
+                    child1_params_list.append(c2_i)
+                    child2_params_list.append(c1_i)
             else:
-                if rng.rand() < self._establishment:
-                    if index_prob >= 0.5:
-                        child_params_list.append(x1_i)
-                    else:
-                        child_params_list.append(x2_i)
+                if rng.rand() >= self._uniform_crossover_prob:
+                    child1_params_list.append(x1_i)
+                    child2_params_list.append(x2_i)
                 else:
-                    if index_prob >= 0.5:
-                        child_params_list.append(x2_i)
-                    else:
-                        child_params_list.append(x1_i)
+                    child1_params_list.append(x2_i)
+                    child2_params_list.append(x1_i)
+
+        child_params_list = child1_params_list if rng.rand() < 0.5 else child2_params_list
         child_params = np.array(child_params_list)
 
         return child_params
--- a/optuna/samplers/nsgaii/_crossovers/_sbx.py
+++ b/optuna/samplers/nsgaii/_crossovers/_sbx.py
@@ -29,6 +29,20 @@ class SBXCrossover(BaseCrossover):
             Distribution index. A small value of ``eta`` allows distant solutions
             to be selected as children solutions. If not specified, takes default
             value of ``2`` for single objective functions and ``20`` for multi objective.
+        establishment:
+            ``establishment`` is the probability of uniform crossover
+            between two individuals selected as candidate child individuals.
+            Optuna returns only one child individual at a time, while SBX crossover generates two.
+            This argument is whether or not two individual are
+            crossover to make one child individual.
+            If not specified, takes default value of ``0.5``.
+            The range of values is ``[0.0, 1.0]``.
+        probability:
+            ``probability`` is the probability of using the value of the generated
+            child variable rather than the value of the parent.
+            where ``1-probability`` is the probability of using the parent's values as it is.
+            If not specified, takes default value of ``0.5``.
+            The range of values is ``[0.0, 1.0]``.
     """
 
     n_parents = 2
@@ -96,16 +110,28 @@ class SBXCrossover(BaseCrossover):
         child_params_list = []
 
         for c1_i, c2_i, x1_i, x2_i in zip(c1, c2, parents_params[0], parents_params[1]):
-            if rng.rand() < self._establishment:
-                if index_prob < self._probability:
-                    child_params_list.append(c1_i)
+            if rng.rand() < self._probability:
+                if rng.rand() < self._establishment:
+                    if index_prob >= 0.5:
+                        child_params_list.append(c1_i)
+                    else:
+                        child_params_list.append(c2_i)
                 else:
-                    child_params_list.append(c2_i)
+                    if index_prob >= 0.5:
+                        child_params_list.append(c2_i)
+                    else:
+                        child_params_list.append(c1_i)
             else:
-                if index_prob < self._probability:
-                    child_params_list.append(x1_i)
+                if rng.rand() < self._establishment:
+                    if index_prob >= 0.5:
+                        child_params_list.append(x1_i)
+                    else:
+                        child_params_list.append(x2_i)
                 else:
-                    child_params_list.append(x2_i)
+                    if index_prob >= 0.5:
+                        child_params_list.append(x2_i)
+                    else:
+                        child_params_list.append(x1_i)
         child_params = np.array(child_params_list)
 
         return child_params
--- a/optuna/samplers/nsgaii/_crossovers/_sbx.py
+++ b/optuna/samplers/nsgaii/_crossovers/_sbx.py
@@ -40,6 +40,9 @@ class SBXCrossover(BaseCrossover):
             between two individuals selected as candidate child individuals.
             This argument is whether or not two individuals are
             crossover to make one child individual.
+            If the ``uniform_crossover_prob`` exceeds 0.5,
+            the result is equivalent to ``1-uniform_crossover_prob``,
+            because it returns one of the two individuals of the crossover result.
             If not specified, takes default value of ``0.5``.
             The range of values is ``[0.0, 1.0]``.
         use_child_gene_prob:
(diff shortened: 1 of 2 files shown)


c-bata:
@sawa3030 @y0z @HideakiImamura Could you review this PR?

sawa3030:
@hrntsm Thank you for your PR! I’ve left some comments—please take a look when you have time!

hrntsm:
@sawa3030 
Thanks for the review.
There does not appear to be any comments specifically attached, where can I check?

sawa3030:
@hrntsm Sorry about that! Can you check them now?

optuna/samplers/nsgaii/_crossovers/_sbx.py
```diff
@@ -76,25 +80,25 @@ def crossover(
...
+            return a if index_prob < 0.5 else b
+
         for c1_i, c2_i, x1_i, x2_i in zip(c1, c2, parents_params[0], parents_params[1]):
-            if rng.rand() < 0.5:
-                if rng.rand() < 0.5:
-                    child_params_list.append(c1_i)
-                else:
-                    child_params_list.append(c2_i)
+            if rng.rand() < self._establishment:
+                options = (c1_i, c2_i) if rng.rand() < self._probability else (c2_i, c1_i)
```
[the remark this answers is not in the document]
hrntsm: @sawa3030 

Once I accepted the suggestion and committed. But it did not work, so I am commenting on it.

The proposed method would not result in the combination of individuals I thought it would unless I branched out with each of the following probabilities.

```py
        for c1_i, c2_i, x1_i, x2_i in zip(c1, c2, parents_params[0], parents_params[1]):
            if rng.rand() < self._establishment:
                if rng.rand() < self._probability:
                    if index_prob < 0.5:
                        child_params_list.append(c1_i)
                    else:
                        child_params_list.append(c2_i)
                else:
                    if index_prob < 0.5:
                        child_params_list.append(c2_i)
                    else:
                        child_params_list.append(c1_i)
            else:
                if rng.rand() < self._probability:
                    if index_prob < 0.5:
                        child_params_list.append(x1_i)
                    else:
                        child_params_list.append(x2_i)
                else:
                    if index_prob < 0.5:
                        child_params_list.append(x2_i)
                    else:
                        child_params_list.append(x1_i)
```


Instead, I believe that it would be better to create the combination as it is now and then determine whether to select one or the other created after the combination is created, which would reduce duplication of code.

optuna/samplers/nsgaii/_crossovers/_sbx.py
```diff
@@ -33,9 +33,18 @@ class SBXCrossover(BaseCrossover):
 
     n_parents = 2
 
-    def __init__(self, eta: float | None = None) -> None:
+    def __init__(
+        self, eta: float | None = None, establishment: float = 0.5, probability: float = 0.5
```
HideakiImamura, line 37: Could you add the document of the arguments?

tests/samplers_tests/test_nsgaii.py
```diff
@@ -900,8 +900,8 @@ def test_crossover_duplicated_param_values(crossover: BaseCrossover) -> None:
         (SPXCrossover(), 0.5, np.array([2.75735931, 3.75735931])),  # rs = [0.5, 0.25].
         (SPXCrossover(), 1.0, np.array([-1.0, 0.0])),  # rs = [1, 1], xks[0].
         (SBXCrossover(), 0.0, np.array([2.0, 3.0])),  # c1 = (p1 + p2) / 2.
-        (SBXCrossover(), 0.5, np.array([3.0, 4.0])),  # p2.
-        (SBXCrossover(), 1.0, np.array([3.0, 4.0])),  # p2.
+        (SBXCrossover(), 0.5, np.array([1.0, 2.0])),  # p2.
+        (SBXCrossover(), 1.0, np.array([1.0, 2.0])),  # p2.
```
HideakiImamura, line 904: According the CI errors https://github.com/optuna/optuna/actions/runs/13983115513/job/39151996886?pr=6008, we don't have to change the test cases.

HideakiImamura (commented):
Thanks for the PR. I have two comments. PTAL.

y0z:
@hrntsm Thank you for your contribution.

I would like to share my opinion.

`SBXCrossOver` generates two children in the original. However, Optuna's Operator must generate only one child. Therefore, the existing implementation generates a child that is a mix of the two children. To get closer to the original implementation, it seems to be better to randomly generate one of the two children.

Also, how about changing the default value of eta to `1` for single-objective optimization? Currently, the setting is `2`, but I couldn't find a reason for it.
https://github.com/optuna/optuna/blob/master/optuna/samplers/nsgaii/_crossovers/_sbx.py#L57

I tried such an implementation and it seems to work well.
```python
    def crossover(
        self,
        parents_params: np.ndarray,
        rng: np.random.RandomState,
        study: Study,
        search_space_bounds: np.ndarray,
    ) -> np.ndarray:
        # https://www.researchgate.net/profile/M-M-Raghuwanshi/publication/267198495_Simulated_Binary_Crossover_with_Lognormal_Distribution/links/5576c78408ae7536375205d7/Simulated-Binary-Crossover-with-Lognormal-Distribution.pdf
        # Section 2 Simulated Binary Crossover (SBX)

        # To avoid generating solutions that violate the box constraints,
        # alpha1, alpha2, xls and xus are introduced, unlike the reference.
        xls = search_space_bounds[..., 0]
        xus = search_space_bounds[..., 1]

        xs_min = np.min(parents_params, axis=0)
        xs_max = np.max(parents_params, axis=0)
        if self._eta is None:
            eta = 20.0 if study._is_multi_objective() else 1.0  # Suggestion 2: change default eta to 1.0
        else:
            eta = self._eta

        xs_diff = np.clip(xs_max - xs_min, 1e-10, None)
        beta1 = 1 + 2 * (xs_min - xls) / xs_diff
        beta2 = 1 + 2 * (xus - xs_max) / xs_diff
        alpha1 = 2 - np.power(beta1, -(eta + 1))
        alpha2 = 2 - np.power(beta2, -(eta + 1))

        us = rng.rand(len(search_space_bounds))
        mask1 = us > 1 / alpha1  # Equation (3).
        betaq1 = np.power(us * alpha1, 1 / (eta + 1))  # Equation (3).
        betaq1[mask1] = np.power((1 / (2 - us * alpha1)), 1 / (eta + 1))[mask1]  # Equation (3).

        mask2 = us > 1 / alpha2  # Equation (3).
        betaq2 = np.power(us * alpha2, 1 / (eta + 1))  # Equation (3)
        betaq2[mask2] = np.power((1 / (2 - us * alpha2)), 1 / (eta + 1))[mask2]  # Equation (3).

        c1 = 0.5 * ((xs_min + xs_max) - betaq1 * xs_diff)  # Equation (4).
        c2 = 0.5 * ((xs_min + xs_max) + betaq2 * xs_diff)  # Equation (5).

        # SBX applies crossover with establishment 0.5, and with probability 0.5,
        # the gene of the parent individual is the gene of the child individual.
        # The original SBX creates two child individuals,
        # but optuna's implementation creates only one child individual.
        # Therefore, when there is no crossover,
        # the gene is selected with equal probability from the parent individuals x1 and x2.

        # child_params_list = []
        # for c1_i, c2_i, x1_i, x2_i in zip(c1, c2, parents_params[0], parents_params[1]):
        #     if rng.rand() < 0.5:
        #         if rng.rand() < 0.5:
        #             child_params_list.append(c1_i)
        #         else:
        #             child_params_list.append(c2_i)
        #     else:
        #         if rng.rand() < 0.5:
        #             child_params_list.append(x1_i)
        #         else:
        #             child_params_list.append(x2_i)
        # child_params = np.array(child_params_list)

        return c1 if rng.rand() < 0.5 else c2  # Suggestion 1: return one of the two children randomly
```
Figure_1

hrntsm:
> Also, how about changing the default value of eta to `1` for single-objective optimization? Currently, the setting is `2`, but I couldn't find a reason for it. https://github.com/optuna/optuna/blob/master/optuna/samplers/nsgaii/_crossovers/_sbx.py#L57

Here, in K.Deb's SBX paper below, it is mentioned that n=2~5 is close to a single point crossing in equation 19.
There is no direct mention of good or bad, but I am wondering if it is safe to leave it at 2 for continuity with previous implementations.

[Simulated Binary Crossover for Continuous Search Space](https://www.complex-systems.com/abstracts/v09_i02_a02/)

This point is subject to the judgment of OptunaTeam.

HideakiImamura:
This is not strong opinion, but I think it is reasonable to leave the value of `eta` as it is (i.e., 2) since it is suggested to set the value in the range of `[2, 5]` according to the [Simulated Binary Crossover for Continuous Search Space](https://www.complex-systems.com/abstracts/v09_i02_a02/).

y0z:
Thanks for providing [the reference](https://www.complex-systems.com/abstracts/v09_i02_a02/).

> I think it is reasonable to leave the value of eta as it is (i.e., 2)

+1

hrntsm:
I have been looking into which individuals to return. I have looked at several papers, but I feel that they only mention returning two generated individuals c1 and c2, as y0z mentioned.

On the other hand, I checked the implementation of some optimization libraries.

|lib name| comment|
|---|---|
|[DEAP](https://github.com/DEAP/deap/blob/a321a568d68c56295652030ab0947cd377026ec1/deap/tools/crossover.py#L262)|The only variable is eta, and the two individuals generated are returned as is.|
|[pymoo](https://www.pymoo.org/operators/crossover.html#Simulated-Binary-Crossover-(SBX))|The variables are eta, prob_bin(establishment), prob_var(probability), similar to the implementation as in this PR.|
|[jMetal](https://github.com/jMetal/jMetal/blob/fd0dac56a08fd50289153e6ab556696064f181a3/jmetal-core/src/main/java/org/uma/jmetal/operator/crossover/impl/SBXCrossover.java#L177)|Same as the current implementation of Optuna, with establishment and probability fixed at 0.5|
|[PlatEMO](https://github.com/BIMK/PlatEMO/blob/43c5f3b3775d0ed57ebf14735924f786d39c5f46/PlatEMO/Algorithms/Multi-objective%20optimization/S-NSGA-II/sbx.m#L26)|The only variable is eta, and the two individuals generated are returned.|

After all, the two arguments other than eta were not specified in any of the papers cited in any of the implementations.

`establishment` is the probability of binary crossover (uniform crossover) among 2 parents and 2 children, `probability` is the probability of applying the SBX crossover (1 - probability of using the parent values as it is)

This is my opinion, but considering the current implementation of Optuna and the reproduction of the paper, I think the arguments should be eta, establishment, and probability.
On the other hand, I thought we need to discuss what the default values should be.

I am thinking either of the following two.

1. The value as it is in the current situation (eta=2, establishment=0.5, probability=0.5)
2. A value that reproduces the distribution of the paper (eta=2, establishment=0.0, probability=1.0)

y0z:
@hrntsm 

Thank you for your survey.

Based on the discussions, introducing `establishment` and `probability` parameters with default values that are the same as in the current implementation, i.e., not changing the current Optuna behavior, seems good.

hrntsm:
@sawa3030 @HideakiImamura 
Thank you for all your comments!

The all checks passed, please confirm.
The implementation is as follows. The default has been the same result so far.

|default(eta=2,establishment=probability=0.5)|reproduct ref paper(eta=1,establishment=0.0,probability=1)|
|---|---|
|Screenshot 2025-03-31 at 14 58 26|Screenshot 2025-03-31 at 14 57 58|

optuna/samplers/nsgaii/_crossovers/_sbx.py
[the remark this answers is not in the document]
y0z: Based on my understanding, the code presented in https://github.com/optuna/optuna/pull/6008#discussion_r2006872347 is equivalent to the code below.
Then, `index_prob` is converted to `1 - index_prob` with probability `0.5` (i.e., `index_prob <= 0.5`).
How can I interpret this behavior? It is not intuitive to understand it from the description in the docstring " ``establishment`` is the probability of uniform crossover between two individuals selected as candidate child individuals."

```python
        index_prob = rng.rand()
        child_params_list = []
        establishment = self._establishment if index_prob >= 0.5 else 1 - self._establishment
        for c1_i, c2_i, x1_i, x2_i in zip(c1, c2, parents_params[0], parents_params[1]):
            if rng.rand() < self._probability:
                if rng.rand() < establishment:
                    child_params_list.append(c1_i)
                else:
                    child_params_list.append(c2_i)
            else:
                if rng.rand() < establishment:
                    child_params_list.append(x1_i)
                else:
                    child_params_list.append(x2_i)
        child_params = np.array(child_params_list)
```

optuna/samplers/nsgaii/_crossovers/_sbx.py
```diff
@@ -76,25 +99,39 @@ def crossover(
...
+                    if index_prob >= 0.5:
+                        child_params_list.append(c1_i)
+                    else:
+                        child_params_list.append(c2_i)
                 else:
-                    child_params_list.append(c2_i)
+                    if index_prob >= 0.5:
+                        child_params_list.append(c2_i)
+                    else:
+                        child_params_list.append(c1_i)
```
HideakiImamura, line 123: This is mathematically equal to the following. Is this intended?
```python
if rng.rand() <= 0.5:
    child_params_list.append(c1_i)
else:
    child_params_list.append(c2_i)
```

optuna/samplers/nsgaii/_crossovers/_sbx.py
```diff
@@ -29,13 +29,36 @@ class SBXCrossover(BaseCrossover):
             Distribution index. A small value of ``eta`` allows distant solutions
             to be selected as children solutions. If not specified, takes default
             value of ``2`` for single objective functions and ``20`` for multi objective.
+        establishment:
+            ``establishment`` is the probability of uniform crossover
+            between two individuals selected as candidate child individuals.
+            Optuna returns only one child individual at a time, while SBX crossover generates two.
+            This argument is whether or not two individual are
```
HideakiImamura, line 36: A nit.
```suggestion
            This argument is whether or not two individuals are
```

optuna/samplers/nsgaii/_crossovers/_sbx.py
HideakiImamura: Sorry, this is my misunderstanding.

sawa3030:
@hrntsm Thank you for your updates! I have a few suggestions and clarifications I'd like to share:

1. As discussed in [this comment](https://github.com/optuna/optuna/pull/6008#discussion_r2022154513), the implementation of this PR behaves the same when `_establishment` is set to 0 or 1, which suggests that the `index_prob` variable may be unnecessary. I suggest simplifying the logic as shown below. This approach preserves the current Optuna behavior, as noted [here](https://github.com/optuna/optuna/pull/6008#issuecomment-2765158084), and could also align with the distribution described in the original paper, as discussed [here](https://github.com/optuna/optuna/discussions/6000#discussioncomment-12472369).

    ```
    for c1_i, c2_i, x1_i, x2_i in zip(c1, c2, parents_params[0], parents_params[1]):
        if rng.rand() < self._probability:
            if rng.rand() < self._establishment:
                child_params_list.append(c1_i)
            else:
                child_params_list.append(c2_i)
        else:
            if rng.rand() < self._establishment:
                child_params_list.append(x1_i)
            else:
                child_params_list.append(x2_i)
    ```

2. In both the current implementation and the code above, `c1` is implicitly associated with `x1`, and `c2` with `x2`. To be more precise, it might be better to decouple these associations and consider introducing separate control variable for selecting between the two child candidates (`c1` and `c2`) and between the two parent values (`x1` and `x2`).

3. As pointed out [in the description of this PR](https://github.com/optuna/optuna/pull/6008#issue-2913367648), the variable name `establishment` doesn't clearly reflect its actual role in the crossover operation. I would suggest renaming it to `uniform_crossover_prob`, and likewise renaming `probability` to `use_child_gene_prob` to better communicate their intent. Of course, these are just suggestions — I'm happy to discuss other naming ideas if anyone has alternatives.

y0z:
@sawa3030 

The code in this PR and the one you suggested seem to have different behavior.
With `establishment=0.1` and `probability=0.5`, we find the difference.

> As discussed in https://github.com/optuna/optuna/pull/6008#discussion_r2022154513, the implementation of this PR behaves the same when _establishment is set to 0 or 1, which suggests that the index_prob variable may be unnecessary. I suggest simplifying the logic as shown below. This approach preserves the current Optuna behavior, as noted https://github.com/optuna/optuna/pull/6008#issuecomment-2765158084, and could also align with the distribution described in the original paper, as discussed https://github.com/optuna/optuna/discussions/6000#discussioncomment-12472369.
>
> ```python
> for c1_i, c2_i, x1_i, x2_i in zip(c1, c2, parents_params[0], parents_params[1]):
>    if rng.rand() < self._probability:
>        if rng.rand() < self._establishment:
>            child_params_list.append(c1_i)
>        else:
>            child_params_list.append(c2_i)
>    else:
>        if rng.rand() < self._establishment:
>            child_params_list.append(x1_i)
>        else:
>            child_params_list.append(x2_i)
>```

Figure_1
Figure_2

optuna/samplers/nsgaii/_crossovers/_sbx.py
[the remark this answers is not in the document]
hrntsm: @y0z 
As you point out, this implementation is confusing because it is simultaneously doing the change for each variable and choosing between the two candidate child individuals.
The final behavior is the same as pointed out, but it would be easier to understand the intent if it were reimplemented as follows.

1. Generate two children `c1`, `c2` based on equation(4) & (5) in SBX paper.
2. Check that the generated children will be used.(`_probability`)
3. Check that uniform crossover applies.(`_establishment`)
4. This is repeated for each variable to create two candidate child individuals.(`child_params_list`, `child2_params_list`)
5. Finally, Optuna can only return one individual, so choose one of them.
   - Equivalent to `index_prob` but used only once, so omitted. 

```py
        child1_params_list = []
        child2_params_list = []

        for c1_i, c2_i, x1_i, x2_i in zip(c1, c2, parents_params[0], parents_params[1]):
            if rng.rand() < self._probability:
                # If the probability of "applying" uniform crossover is `_establishment`, then the reverse of the equal sign may be more appropriate.
                if rng.rand() < self._establishment:
                    child1_params_list.append(c1_i)
                    child2_params_list.append(c2_i)
                else:
                    child1_params_list.append(c2_i)
                    child2_params_list.append(c1_i)
            else:
                if rng.rand() < self._establishment:
                    child1_params_list.append(x1_i)
                    child2_params_list.append(x2_i)
                else:
                    child1_params_list.append(x2_i)
                    child2_params_list.append(x1_i)

        child_params_list = child1_params_list if rng.rand() < 0.5 else child2_params_list
        child_params = np.array(child_params_list)
```

hrntsm:
@sawa3030 

As for 1 & 2, please check https://github.com/optuna/optuna/pull/6008#discussion_r2026322423 for a summary of the intent of the implementation.

As for 3(the variable name), it more accurately describes the behavior than the one I gave at the beginning, and I agree with that suggestion.

optuna/samplers/nsgaii/_crossovers/_sbx.py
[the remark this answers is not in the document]
y0z: Thank you! It is more understandable!
It also sounds good to rename the `establishment` `uniform_crossover_prob` as suggested.

optuna/samplers/nsgaii/_crossovers/_sbx.py
```diff
@@ -76,25 +111,33 @@ def crossover(
         c1 = 0.5 * ((xs_min + xs_max) - betaq1 * xs_diff)  # Equation (4).
         c2 = 0.5 * ((xs_min + xs_max) + betaq2 * xs_diff)  # Equation (5).
 
-        # SBX applies crossover with establishment 0.5, and with probability 0.5,
+        # SBX applies crossover with use_chile_gene_probability and uniform_crossover_probability.,
```
y0z, line 114:
```suggestion
        # SBX applies crossover with use_child_gene_prob and uniform_crossover_prob.,
```

y0z (approved):
Thank you for your careful discussion.
LGTM.

sawa3030:
Thank you so much for your clear classification and explanation — it really helped me understand the logic.
I have one more question: since the behavior is essentially the same when `uniform_crossover_prob` is set to 0 or 1, and it’s a bit unclear how users should choose the value, would it make sense to restrict the range to `[0.0, 0.5]`? This isn’t a strong preference, as values greater than 0.5 don’t cause any issues in practice.

hrntsm:
@sawa3030 
> would it make sense to restrict the range to `[0.0, 0.5]`?

I once thought the same thing, but the “probability” range of [0, 0.5] seemed counter-intuitive, so I decided to use [0,1].
I will add followings to the docstring to make it more understandable.

> If the `uniform_crossover_prob` exceeds 0.5, the result is equivalent to `1-uniform_crossover_prob`, because it returns one of the two individuals of the crossover result.

sawa3030 (approved):
Thank you for all the discussions and clarifications. LGTM!

Issue #5485, opened 2024-06-07
Specify the removing version on `convert_positional_args`


not522:
### Motivation

`convert_positional_args` does not output the information for `deprecated_version` and `removed_version`. They are necessary to let users know when features will be removed.

### Description

Adding `deprecated_version` and `removed_version` arguments to `convert_positional_args`. They should be aligned with `deprecated_func` and `deprecated_class`.

fusawa-yugo:
I would like to take on it.

PR #6009, opened 2025-03-13
Add arguments of versions to `convert_positional_args`

fusawa-yugo:
## Motivation

This PR is from https://github.com/optuna/optuna/issues/5485
## Description of the changes

I added arguments and warning for `deprecated_version` and `removed_version`, referring to `deprecated_func`
The format of warning is:
```
"Positional arguments in {name} have been deprecated since v{d_ver}. "
"They will be removed in v{r_ver}. "
"Please use keyword arguments instead. "
"See https://github.com/optuna/optuna/releases/tag/v{d_ver} for details."
```

### DIFF
--- a/optuna/_convert_positional_args.py
+++ b/optuna/_convert_positional_args.py
@@ -25,7 +25,7 @@ _DEPRECATION_WARNING_TEMPLATE = (
     "Positional arguments {deprecated_positional_arg_names} in {func_name}() "
     "have been deprecated since v{d_ver}. "
     "They will be replaced with the corresponding keyword arguments in v{r_ver}, "
-    "so please use keyword specification instead. "
+    "so please use the keyword specification instead. "
     "See https://github.com/optuna/optuna/releases/tag/v{d_ver} for details."
 )
 
--- a/optuna/_convert_positional_args.py
+++ b/optuna/_convert_positional_args.py
@@ -6,7 +6,6 @@ from functools import wraps
 from inspect import Parameter
 from inspect import signature
 from typing import Any
-from typing import Optional
 from typing import TYPE_CHECKING
 from typing import TypeVar
 import warnings
@@ -50,8 +49,8 @@ def convert_positional_args(
     *,
     previous_positional_arg_names: Sequence[str],
     warning_stacklevel: int = 2,
-    deprecated_version: Optional[str] = None,
-    removed_version: Optional[str] = None,
+    deprecated_version: str | None = None,
+    removed_version: str | None = None,
 ) -> "Callable[[Callable[_P, _T]], Callable[_P, _T]]":
     """Convert positional arguments to keyword arguments.
 
--- a/optuna/_convert_positional_args.py
+++ b/optuna/_convert_positional_args.py
@@ -93,7 +93,7 @@ def convert_positional_args(
             positional_arg_names = _get_positional_arg_names(func)
             inferred_kwargs = _infer_kwargs(previous_positional_arg_names, *args)
 
-            if args and (deprecated_version or removed_version):
+            if inferred_kwargs and (deprecated_version or removed_version):
                 warning_messages.append(
                     _DEPRECATION_WARNING_TEMPLATE.format(
                         positional_arg=previous_positional_arg_names,
--- a/optuna/_convert_positional_args.py
+++ b/optuna/_convert_positional_args.py
@@ -5,7 +5,6 @@ from collections.abc import Sequence
 from functools import wraps
 from inspect import Parameter
 from inspect import signature
-import textwrap
 from typing import Any
 from typing import TYPE_CHECKING
 from typing import TypeVar
@@ -21,25 +20,13 @@ if TYPE_CHECKING:
     _P = ParamSpec("_P")
     _T = TypeVar("_T")
 
-_DEPRECATION_NOTE_TEMPLATE = """
-
-.. warning::
-    Using positional arguments is deprecated since v{d_ver}. Support for positional arguments
-    will be removed in the future. The removal is currently scheduled for v{r_ver}, but this
-    schedule is subject to change. Please use keyword arguments instead.
-    See https://github.com/optuna/optuna/releases/tag/v{d_ver}.
-"""
 
 _DEPRECATION_WARNING_TEMPLATE = (
-    "Positional arguments in {name} have been deprecated since v{d_ver}. "
+    "Positional arguments {positional_arg} in {func_name}() have been deprecated since v{d_ver}. "
     "They will be removed in v{r_ver}. "
     "Please use keyword arguments instead. "
     "See https://github.com/optuna/optuna/releases/tag/v{d_ver} for details."
 )
-
-
-def _get_docstring_indent(docstring: str) -> str:
-    return docstring.split("\n")[-1] if "\n" in docstring else ""
 
 
 def _get_positional_arg_names(func: "Callable[_P, _T]") -> list[str]:
@@ -72,9 +59,9 @@ def convert_positional_args(
         warning_stacklevel:
             Level of the stack trace where decorated function locates.
         deprecated_version:
-            Version number in which the use of positional arguments is deprecated.
+            The version in which the use of positional arguments is deprecated.
         removed_version:
-            Version number in which the use of positional arguments was completely removed.
+            The version in which the use of positional arguments will be removed.
     """
 
     if deprecated_version is None and removed_version is None:
@@ -82,11 +69,11 @@ def convert_positional_args(
     else:
         if deprecated_version is None:
             raise ValueError(
-                "deprecated_version must not be None when removed_version is specified"
+                "deprecated_version must not be None when removed_version is specified."
             )
         if removed_version is None:
             raise ValueError(
-                "removed_version must not be None when deprecated_version is specified"
+                "removed_version must not be None when deprecated_version is specified."
             )
 
         _validate_version(deprecated_version)
@@ -94,15 +81,6 @@ def convert_positional_args(
         _validate_two_version(deprecated_version, removed_version)
 
     def converter_decorator(func: "Callable[_P, _T]") -> "Callable[_P, _T]":
-        if deprecated_version or removed_version:
-            if func.__doc__ is None:
-                func.__doc__ = ""
-
-            note = _DEPRECATION_NOTE_TEMPLATE.format(
-                d_ver=deprecated_version, r_ver=removed_version
-            )
-            indent = _get_docstring_indent(func.__doc__)
-            func.__doc__ = func.__doc__.strip() + textwrap.indent(note, indent) + indent
 
         assert set(previous_positional_arg_names).issubset(set(signature(func).parameters)), (
             f"{set(previous_positional_arg_names)} is not a subset of"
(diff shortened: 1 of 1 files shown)


c-bata:
@not522 @porink0424 Could you review this PR?

optuna/_convert_positional_args.py
```diff
@@ -37,22 +61,66 @@ def convert_positional_args(
     *,
     previous_positional_arg_names: Sequence[str],
     warning_stacklevel: int = 2,
+    deprecated_version: str | None = None,
+    removed_version: str | None = None,
```
not522, line 65: Let me confirm whether we will make the arguments optional. I recommend it is temporarily optional, and we will make it required on follow-up. How about that?

optuna/_convert_positional_args.py
```diff
@@ -37,22 +61,66 @@ def convert_positional_args(
...
+        if deprecated_version or removed_version:
+            if func.__doc__ is None:
+                func.__doc__ = ""
+
+            note = _DEPRECATION_NOTE_TEMPLATE.format(
+                d_ver=deprecated_version, r_ver=removed_version
+            )
+            indent = _get_docstring_indent(func.__doc__)
+            func.__doc__ = func.__doc__.strip() + textwrap.indent(note, indent) + indent
+
```
not522, line 106: I think there is no need to add the deprecation note to the document because it is written in the recommended way.

optuna/_convert_positional_args.py
```diff
@@ -37,22 +61,66 @@ def convert_positional_args(
...
+            if deprecated_version or removed_version:
+                warnings.warn(
+                    _DEPRECATION_WARNING_TEMPLATE.format(
+                        name=func.__name__,
+                        d_ver=deprecated_version,
+                        r_ver=removed_version,
+                    ),
+                    FutureWarning,
+                    stacklevel=warning_stacklevel,
+                )
```
not522, line 123: With this implementation, the warning will always be shown, so please change it so that it only appears for positional arguments.

not522 (commented):
Thank you for your PR! Could you check my comments?

optuna/_convert_positional_args.py
```diff
@@ -5,18 +5,42 @@
...
+
+.. warning::
+    Using positional arguments is deprecated since v{d_ver}. Support for positional arguments
+    will be removed in the future. The removal is currently scheduled for v{r_ver}, but this
+    schedule is subject to change. Please use keyword arguments instead.
+    See https://github.com/optuna/optuna/releases/tag/v{d_ver}.
+"""
+
+_DEPRECATION_WARNING_TEMPLATE = (
+    "Positional arguments in {name} have been deprecated since v{d_ver}. "
```
porink0424, line 34: The warning message currently only shows the function name, but it would be more helpful to also display which positional arguments have been deprecated.

optuna/_convert_positional_args.py
```diff
@@ -37,22 +61,66 @@ def convert_positional_args(
...
-        previous_positional_arg_names: List of names previously given as positional arguments.
-        warning_stacklevel: Level of the stack trace where decorated function locates.
+        previous_positional_arg_names:
+            List of names previously given as positional arguments.
+        warning_stacklevel:
+            Level of the stack trace where decorated function locates.
+        deprecated_version:
+            Version number in which the use of positional arguments is deprecated.
+        removed_version:
+            Version number in which the use of positional arguments was completely removed.
```
porink0424, line 77: It would be better to align these explanations with `deprecated_func`:
```suggestion
        deprecated_version:
            The version in which the use of positional arguments is deprecated.
        removed_version:
            The version in which the use of positional arguments will be removed.
```

optuna/_convert_positional_args.py
```diff
@@ -37,22 +61,66 @@ def convert_positional_args(
...
     """
 
+    if deprecated_version is None and removed_version is None:
+        pass
+    else:
+        if deprecated_version is None:
+            raise ValueError(
+                "deprecated_version must not be None when removed_version is specified"
```
porink0424, line 85: [nits] 
```suggestion
                "deprecated_version must not be None when removed_version is specified."
```

optuna/_convert_positional_args.py
```diff
@@ -37,22 +61,66 @@ def convert_positional_args(
...
+            )
+        if removed_version is None:
+            raise ValueError(
+                "removed_version must not be None when deprecated_version is specified"
```
porink0424, line 89: Ditto.

optuna/_convert_positional_args.py
porink0424, line 123: Currently the implementation results in two warnings being issued, including the following:
```
                warnings.warn(
                    f"{func.__name__}() got {expected_kwds} as positional arguments "
                    "but they were expected to be given as keyword arguments.",
                    ...
                )
```
This may be a bit annoying for users, so it would be better to consolidate them into a single warning.

porink0424 (commented):
Thanks for your PR, I've left some comments.

optuna/_convert_positional_args.py
fusawa-yugo: I agree with you.

optuna/_convert_positional_args.py
fusawa-yugo: I see.
I will delete this.

fusawa-yugo:
Thank you for reviews!!
I revised the warnings.
Warning is displayed like below.

### Test function
```
@convert_positional_args(
    previous_positional_arg_names=["a", "b"],
    deprecated_version="3.0.0",
    removed_version="4.0.0",
)
def example_function(a: int, b: int, c: int = 3, d: int = 4) -> None:
    print(f"a: {a}, b: {b}, c: {c}, d: {d}")

example_function(1, 2) 
```

### Warning
```
FutureWarning: Positional arguments ['a', 'b'] in example_function() have been deprecated since v3.0.0. They will be removed in v4.0.0. Please use keyword arguments instead. See https://github.com/optuna/optuna/releases/tag/v3.0.0 for details.
  example_function(1, 2)
a: 1, b: 2, c: 3, d: 4
```

optuna/_convert_positional_args.py
```diff
@@ -37,32 +48,73 @@ def convert_positional_args(
...
             f" {set(signature(func).parameters)}"
         )
 
         @wraps(func)
         def converter_wrapper(*args: Any, **kwargs: Any) -> "_T":
+            warning_messages = []
             positional_arg_names = _get_positional_arg_names(func)
             inferred_kwargs = _infer_kwargs(previous_positional_arg_names, *args)
+
+            if inferred_kwargs and (deprecated_version or removed_version):
```
not522, line 96: `inferred_kwargs` include ordinal positional arguments. So we should check the lengths of them. 

```suggestion
            if len(inferred_kwargs) > len(positional_arg_names) and (deprecated_version or removed_version):
```

not522:
Thank you for your update! It's almost OK.

How about swap the order of warning messages? For example, the following outputs the versions first, but I think It's better to first output what users should do (i.e., `example_function() got {'b', 'a'} as ...`).

```python
from optuna._convert_positional_args import convert_positional_args

@convert_positional_args(
    previous_positional_arg_names=["a", "b", "c", "d"],
    deprecated_version="3.0.0",
    removed_version="4.0.0",
)
def example_function(*, a: int, b: int, c: int = 3, d: int = 4) -> None:
    print(f"a: {a}, b: {b}, c: {c}, d: {d}")

example_function(1, 2)
```

```
/Users/naotomizuno/optuna/tmp_convert.py:11: FutureWarning: Positional arguments ['a', 'b', 'c', 'd'] in example_function() have been deprecated since v3.0.0. They will be removed in v4.0.0. Please use keyword arguments instead. See https://github.com/optuna/optuna/releases/tag/v3.0.0 for details.
example_function() got {'b', 'a'} as positional arguments but they were expected to be given as keyword arguments.
```

fusawa-yugo:
Thank you for your comments again!!
I updated the code.
Please review it again!

not522 (approved):
LGTM!

c-bata:
@HideakiImamura I unassigned @porink0424 since he is busy for a while. Could you assign an additional reviewer?

optuna/_convert_positional_args.py
```diff
@@ -18,6 +21,14 @@
     _T = TypeVar("_T")
 
 
+_DEPRECATION_WARNING_TEMPLATE = (
+    "Positional arguments {positional_arg} in {func_name}() have been deprecated since v{d_ver}. "
```
nabenabe0928, line 25:
```suggestion
    "Positional arguments {deprecated_positional_arg_names} in {func_name}() have been deprecated since v{d_ver}. "
```

optuna/_convert_positional_args.py
```diff
@@ -18,6 +21,14 @@
...
+    "They will be removed in v{r_ver}. "
+    "Please use keyword arguments instead. "
```
nabenabe0928, line 27: [nit] They will NOT be removed, but simply updated so that only keyword argument specification works.
```suggestion
    "They will be replaced with the corresponding keyword arguments in v{r_ver}, so please use keyword specification instead. "
```

optuna/_convert_positional_args.py
nabenabe0928: Although I have no preference, I think it would be better to make the input typings stricter based on [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

The reason behind this comes from the fact that looser input typings implicitly assume that the function have a routine where we handle the possible input typings.
Namely, we have the processing for `removed_version=None` in this case and everywhere this function is called may use `removed_version=None`, which is a bit annoying to update later.
As we have the guidelines for deprecation, I think it would be better to not have optional for now.

nabenabe0928: Btw, is `removed_version` actually `replaced_version`?

optuna/_convert_positional_args.py
```diff
@@ -37,32 +48,73 @@ def convert_positional_args(
...
+
             if len(inferred_kwargs) > len(positional_arg_names):
                 expected_kwds = set(inferred_kwargs) - set(positional_arg_names)
-                warnings.warn(
+                warning_messages.append(
                     f"{func.__name__}() got {expected_kwds} as positional arguments "
-                    "but they were expected to be given as keyword arguments.",
-                    FutureWarning,
-                    stacklevel=warning_stacklevel,
+                    "but they were expected to be given as keyword arguments."
```
nabenabe0928, line 100: [nit] Do you think we need a space here?
```suggestion
                    "but they were expected to be given as keyword arguments. "
```

optuna/_convert_positional_args.py
nabenabe0928: Sorry, now I noticed that most of them have not specified version, so we need to make it optional for now, probably.

optuna/_convert_positional_args.py
```diff
@@ -37,32 +48,73 @@ def convert_positional_args(
...
+            Level of the stack trace where decorated function locates.
+        deprecated_version:
+            The version in which the use of positional arguments is deprecated.
+        removed_version:
+            The version in which the use of positional arguments will be removed.
     """
 
+    if deprecated_version is None and removed_version is None:
+        pass
+    else:
```
nabenabe0928, line 69:
```suggestion
    if deprecated_version is not None or removed_version is not None:
```

nabenabe0928 (commented):
Thank you for the PR, i left some comments:)

fusawa-yugo:
Thank you for review!!
Could you check it again?

optuna/_convert_positional_args.py
```diff
@@ -22,9 +23,10 @@
 
 
 _DEPRECATION_WARNING_TEMPLATE = (
-    "Positional arguments {positional_arg} in {func_name}() have been deprecated since v{d_ver}. "
-    "They will be removed in v{r_ver}. "
-    "Please use keyword arguments instead. "
+    "Positional arguments {deprecated_positional_arg_names} in {func_name}() "
+    "have been deprecated since v{d_ver}. "
+    "They will be replaced with the corresponding keyword arguments in v{r_ver}, "
+    "so please use keyword specification instead. "
```
nabenabe0928, line 29:
```suggestion
    "so please use the keyword specification instead. "
```

optuna/_convert_positional_args.py
```diff
@@ -48,8 +50,8 @@ def convert_positional_args(
     *,
     previous_positional_arg_names: Sequence[str],
     warning_stacklevel: int = 2,
-    deprecated_version: str | None = None,
-    removed_version: str | None = None,
+    deprecated_version: Optional[str] = None,
+    removed_version: Optional[str] = None,
```
nabenabe0928, line 54: Could you revert here?
It is recommended to use `str | None` instead of `Optional[str]` according to [PEP604](https://peps.python.org/pep-0604/).

nabenabe0928 (commented):
Thank you for the changes, almost LGTM!

nabenabe0928 (approved):

PR #6010, opened 2025-03-13
Add `wait_server_ready` method in GrpcStorageProxy

hitsgub:
## Motivation

The GrpcStorageProxy class introduced in v4.2.0 does not have any mechanism to wait for the server to start, so users need to write code to wait for the server to start. In this PR, we change the GrpcStorageProxy class to wait for the server to start when initializing. 

## Description of the changes

Use grpc.channel_ready_future function for wait to server.

### DIFF
--- a/optuna/storages/_grpc/client.py
+++ b/optuna/storages/_grpc/client.py
@@ -90,6 +90,7 @@ class GrpcStorageProxy(BaseStorage):
         self._cache = GrpcClientCache(self._stub)
 
     def __del__(self) -> None:
+        del self._cache
         self._channel.close()
 
     def __getstate__(self) -> dict[Any, Any]:
--- a/optuna/storages/_grpc/client.py
+++ b/optuna/storages/_grpc/client.py
@@ -78,11 +78,17 @@ class GrpcStorageProxy(BaseStorage):
         self._setup()
 
     def _setup(self) -> None:
+        """Set up the gRPC channel and stub."""
         self._channel = create_insecure_channel(self._host, self._port)
         self._stub = api_pb2_grpc.StorageServiceStub(self._channel)
         self._cache = GrpcClientCache(self._stub)
 
     def wait_server_ready(self, timeout: float | None = None) -> None:
+        """Wait until the gRPC server is ready.
+
+        Args:
+            timeout: The maximum time to wait in seconds. If :obj:`None`, wait indefinitely.
+        """
         try:
             with create_insecure_channel(self._host, self._port) as channel:
                 grpc.channel_ready_future(channel).result(timeout=timeout)
@@ -90,6 +96,7 @@ class GrpcStorageProxy(BaseStorage):
             raise ConnectionError("GRPC connection timeout") from e
 
     def close(self) -> None:
+        """Close the gRPC channel."""
         self._channel.close()
 
     def __getstate__(self) -> dict[Any, Any]:
--- a/optuna/storages/_grpc/client.py
+++ b/optuna/storages/_grpc/client.py
@@ -85,10 +85,17 @@ class GrpcStorageProxy(BaseStorage):
                 grpc.channel_ready_future(channel).result(timeout=self._timeout)
         except grpc.FutureTimeoutError as e:
             raise ConnectionError("GRPC connection timeout") from e
-        self._cache = GrpcClientCache(self._host, self._port)
+        self._channel = create_insecure_channel(self._host, self._port)
+        self._stub = api_pb2_grpc.StorageServiceStub(self._channel)
+        self._cache = GrpcClientCache(self._stub)
+
+    def __del__(self) -> None:
+        self._channel.close()
 
     def __getstate__(self) -> dict[Any, Any]:
         state = self.__dict__.copy()
+        del state["_channel"]
+        del state["_stub"]
         del state["_cache"]
         return state
 
@@ -106,26 +113,22 @@ class GrpcStorageProxy(BaseStorage):
             ],
             study_name=study_name or DEFAULT_STUDY_NAME_PREFIX + str(uuid.uuid4()),
         )
-        with create_insecure_channel(self._host, self._port) as channel:
-            stub = api_pb2_grpc.StorageServiceStub(channel)
-            try:
-                response = stub.CreateNewStudy(request)
-            except grpc.RpcError as e:
-                if e.code() == grpc.StatusCode.ALREADY_EXISTS:
-                    raise DuplicatedStudyError from e
-                raise
+        try:
+            response = self._stub.CreateNewStudy(request)
+        except grpc.RpcError as e:
+            if e.code() == grpc.StatusCode.ALREADY_EXISTS:
+                raise DuplicatedStudyError from e
+            raise
         return response.study_id
 
     def delete_study(self, study_id: int) -> None:
         request = api_pb2.DeleteStudyRequest(study_id=study_id)
-        with create_insecure_channel(self._host, self._port) as channel:
-            stub = api_pb2_grpc.StorageServiceStub(channel)
-            try:
-                stub.DeleteStudy(request)
-            except grpc.RpcError as e:
-                if e.code() == grpc.StatusCode.NOT_FOUND:
-                    raise KeyError from e
-                raise
+        try:
+            self._stub.DeleteStudy(request)
+        except grpc.RpcError as e:
+            if e.code() == grpc.StatusCode.NOT_FOUND:
+                raise KeyError from e
+            raise
         # TODO(c-bata): Fix a cache invalidation issue when using SQLite3
         # Please see https://github.com/optuna/optuna/pull/5872/files#r1893708995 for details.
         self._cache.delete_study_cache(study_id)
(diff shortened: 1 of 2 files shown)


y0z:
@HideakiImamura @c-bata 
Could you review this PR?

HideakiImamura:
I confirmed that the failed tests cases pass in my local environment (Apple M2 chip, macOS 14.6.1). I used same python version with the CI (Python 3.12). The CI uses Apple M1 chip and macOS 14 according to https://github.com/actions/runner-images/tree/main. I would like to investigate the reason of the CI failure.

toshihikoyanase:
I reproduced the issue on my M1 MacBook. When running tests, the gRPC Storage Proxy is repeatedly started on the same port, which leads to a port unavailability error. This seems to be due to the server not shutting down gracefully. 

I verified that implementing a graceful shutdown (by waiting for the shutdown event) resolves the problem on my machine. Please review the changes and let me know your thoughts.

```diff
diff --git a/optuna/testing/storages.py b/optuna/testing/storages.py
index 1893994f2..819339e61 100644
--- a/optuna/testing/storages.py
+++ b/optuna/testing/storages.py
@@ -121,7 +121,8 @@ class StorageSupplier:

         if self.server:
             assert self.thread is not None
-            self.server.stop(None)
+            shutdown_event = self.server.stop(0.1)
+            shutdown_event.wait()
             self.thread.join()
             self.server = None
             self.thread = None
```

toshihikoyanase:
I reconsidered my approach and examined the gRPC clients rather than the gRPC servers.
I tested the change in my forked repository and found that the mactest ran as expected.
https://github.com/toshihikoyanase/optuna/actions/runs/13943424624/job/39025049106?pr=3

The main change is that it appears `GrpcStorageProxy._cache` and `GrpcStorageProxy._stub` were retaining a gRPC client. I removed them because I suspect that the timing of their deletion was causing issues—possibly by keeping connections open. Consequently, a graceful shutdown is not necessary, and port 13000 will be continuously reused.

https://github.com/toshihikoyanase/optuna/pull/3/commits/819737b66105b26c46dffe0d815b144723c1f348

toshihikoyanase:
It seems a bit strange that only the test failed, so I reviewed the main code and noticed that the channel might not be closing properly.

In the changes of this PR, the insecure_channel is used with the with statement, so it is closed when the scope is exited:

```python
class GrpcStorageProxy(BaseStorage):
    def __init__(self, *, host: str = "localhost", port: int = 13000, timeout: int = 900) -> None:
        def wait_server(host: str, port: int, timeout: int) -> None:
            try:
                with grpc.insecure_channel(
                    f"{host}:{port}", options=[("grpc.max_receive_message_length", -1)]
                ) as channel:
                    grpc.channel_ready_future(channel).result(timeout=timeout)
            except grpc.FutureTimeoutError as e:
                raise ConnectionError("GRPC connection timeout") from e
```

However, the channel specified for _stub below does not seem to be closed:

```python
self._stub = api_pb2_grpc.StorageServiceStub(
            grpc.insecure_channel(
                f"{host}:{port}",
                options=[("grpc.max_receive_message_length", -1)],
            )
        )  # type: ignore
```
Typically, there might not be issues since we don't reconnect many times to the same port in a short period, and it might be acceptable to leave it to garbage collection. However, it might be problematic depending on the implementation, for example, in the case of mac-test CI jobs.

hitsgub:
Thank you.
In order to more reliably close a channel, I have changed to use a `with` statement at each point where stub is used.

optuna/storages/_grpc/client.py
```diff
@@ -59,36 +69,39 @@ class GrpcStorageProxy(BaseStorage):
...
+
+    def setup(self) -> None:
+        def wait_server(host: str, port: int, timeout: int) -> None:
+            try:
+                with create_insecure_channel(host, port) as channel:
+                    grpc.channel_ready_future(channel).result(timeout=timeout)
+            except grpc.FutureTimeoutError as e:
+                raise ConnectionError("GRPC connection timeout") from e
+
+        wait_server(self._host, self._port, self._timeout)
```
HideakiImamura, line 94: It seems that it's possible to avoid to define the function in the function.

HideakiImamura (commented):
Thanks for the PR and great efforts. I have a minor comment. PTAL.

HideakiImamura (approved):

hitsgub:
trying to re-use stubs and channels

https://grpc.io/docs/guides/performance/

optuna/storages/_grpc/client.py
```diff
@@ -59,36 +65,43 @@ class GrpcStorageProxy(BaseStorage):
     Args:
         host: The hostname of the gRPC server.
         port: The port of the gRPC server.
+        timeout: The timeout for the gRPC connection in seconds.
```
c-bata, line 68: This `timeout` argument is only used to wait for the server to become ready and is not applied to individual grpc requests, such as `stub.CreateNewStudy()`. I believe most users would assume that the `timeout` argument is passed to each request, similar to the following gRPC example:

```python
# https://github.com/grpc/grpc/blob/master/examples/python/timeout/greeter_client.py
response = stub.SayHello(
    helloworld_pb2.HelloRequest(name=message), timeout=3
)
```
 
Would it be clearer to rename the argument to something more descriptive, such as `startup_timeout`? This change would avoid breaking compatibility if we decide to introduce a `timeout` argument for individual requests in the future.

c-bata: But on second thought, this concern will be addressed by the suggestion I commented on `setup` method.

optuna/storages/_grpc/client.py
```diff
@@ -59,36 +65,43 @@ class GrpcStorageProxy(BaseStorage):
...
+
+    def setup(self) -> None:
+        try:
+            with create_insecure_channel(self._host, self._port) as channel:
+                grpc.channel_ready_future(channel).result(timeout=self._timeout)
+        except grpc.FutureTimeoutError as e:
+            raise ConnectionError("GRPC connection timeout") from e
+        self._channel = create_insecure_channel(self._host, self._port)
+        self._stub = api_pb2_grpc.StorageServiceStub(self._channel)
+        self._cache = GrpcClientCache(self._stub)
```
c-bata, line 90: In my opinion, it's better to separate this method into the following two, allowing more flexible usage. This way, users can instantiate the `GrpcStorageProxy` class even if the server isn't ready yet.

```python
    def _setup(self) -> None:
        self._channel = create_insecure_channel(self._host, self._port)
        self._stub = api_pb2_grpc.StorageServiceStub(self._channel)
        self._cache = GrpcClientCache(self._stub)

    def wait_server_ready(self, timeout: int) -> None:
        try:
            with create_insecure_channel(self._host, self._port) as channel:
                grpc.channel_ready_future(channel).result(timeout=self._timeout)
        except grpc.FutureTimeoutError as e:
            raise ConnectionError("GRPC connection timeout") from e
```

Accordingly, you can update `optuna/testing/storages.py` as follows:

```python
    def _create_proxy(self, storage: BaseStorage) -> GrpcStorageProxy:
        port = _find_free_port()
        self.server = optuna.storages._grpc.server.make_server(storage, "localhost", port)
        self.proxy = GrpcStorageProxy(host="localhost", port=port)

        self.thread = threading.Thread(target=self.server.start)
        self.thread.start()

        self.proxy.wait_server_ready(timeout=60)
        return self.proxy
```

What do you think?

c-bata (commented):
Thank you for your pull request. I left some early feedback comments.

optuna/storages/_grpc/client.py
hitsgub: So the change is that the user explicitly waits for the server to start.
I think it is good.

optuna/storages/_grpc/client.py
```diff
@@ -59,36 +65,43 @@ class GrpcStorageProxy(BaseStorage):
...
 
-    def __init__(self, *, host: str = "localhost", port: int = 13000) -> None:
-        self._stub = api_pb2_grpc.StorageServiceStub(
-            grpc.insecure_channel(
-                f"{host}:{port}",
-                options=[("grpc.max_receive_message_length", -1)],
-            )
-        )  # type: ignore
-        self._cache = GrpcClientCache(self._stub)
+    def __init__(self, *, host: str = "localhost", port: int = 13000, timeout: int = 900) -> None:
```
hitsgub, line 76: I will fix int -> float
fyi:
https://github.com/grpc/grpc/blob/ecdeae1f13cb68b8e6d0d2c0a596238206c5710d/src/python/grpcio/grpc/_utilities.py#L161

optuna/storages/_grpc/client.py
```diff
@@ -67,28 +73,35 @@ class GrpcStorageProxy(BaseStorage):
...
         self._host = host
         self._port = port
+        self.setup()
+
+    def setup(self) -> None:
+        self._channel = create_insecure_channel(self._host, self._port)
+        self._stub = api_pb2_grpc.StorageServiceStub(self._channel)
+        self._cache = GrpcClientCache(self._stub)
+
+    def wait_server_ready(self, timeout: float) -> None:
```
c-bata, line 85: It might be good to change the default value to None, which means wait forever until ready. This behavior is the same with `.result()` method.
https://grpc.github.io/grpc/python/grpc.html#grpc.Future.result

```suggestion
    def wait_server_ready(self, timeout: float | None = None) -> None:
```

optuna/storages/_grpc/client.py
```diff
@@ -67,28 +73,35 @@ class GrpcStorageProxy(BaseStorage):
...
-                f"{host}:{port}",
-                options=[("grpc.max_receive_message_length", -1)],
-            )
-        )  # type: ignore
-        self._cache = GrpcClientCache(self._stub)
         self._host = host
         self._port = port
+        self.setup()
+
+    def setup(self) -> None:
```
c-bata, line 80: [nit] How about making this method private? Unlike other methods such as `wait_server_ready` and `close` methods, I have no idea how users can utilize this method for now.

```suggestion
    def _setup(self) -> None:
```

optuna/storages/_grpc/client.py
```diff
@@ -67,28 +73,35 @@ class GrpcStorageProxy(BaseStorage):
...
+        self._cache = GrpcClientCache(self._stub)
+
+    def wait_server_ready(self, timeout: float) -> None:
+        try:
+            with create_insecure_channel(self._host, self._port) as channel:
+                grpc.channel_ready_future(channel).result(timeout=timeout)
+        except grpc.FutureTimeoutError as e:
+            raise ConnectionError("GRPC connection timeout") from e
+
+    def close(self) -> None:
```
c-bata, line 92: Currently, the docstrings for the `close`, `setup`, `wait_server_ready` methods are empty as shown in the link below. Could you add docstrings?

https://optuna--6010.org.readthedocs.build/en/6010/reference/generated/optuna.storages.GrpcStorageProxy.html#optuna.storages.GrpcStorageProxy

c-bata (commented):
Changes almost look great to me 💯  I left some minor suggestions though.

c-bata (approved):
LGTM after CI passes!

@hitsgub Could you update the PR title as well? "Add `wait_server_ready` method in GrpcStorageProxy" might be clearer.

HideakiImamura (approved):
Thanks for the update. LGTM!

PR #6016, opened 2025-03-21
Set default `study_name` of `load_study` to `None`

not522:
## Motivation

`study_name` can be `None` after #2608, but we need to specify it, i.e., `load_study(study_name=None, storage=...)`, because it is a positional argument. #3270 converted it to a positional argument, so we can set `None` to the default of `study_name`. This PR will allow us to write `load_study(storage=...)`.

## Description of the changes

- Set default `study_name` of `load_study` to `None`
- Add tests

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1294,7 +1294,7 @@ def create_study(
 )
 def load_study(
     *,
-    study_name: str | None,
+    study_name: str | None = None,
     storage: str | storages.BaseStorage,
     sampler: "samplers.BaseSampler" | None = None,
     pruner: pruners.BasePruner | None = None,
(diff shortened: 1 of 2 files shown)


c-bata:
@not522 Thank you for your pull request. I don't have a strong opinion, but I have a slight concern about the change.

When using the `optuna.load_study()` function, I believe users generally need to be aware of the existence of the `study_name` argument. If someone who is not very familiar with Optuna loads a study using `optuna.load_study(storage=...)`, they may be confused if they create an additional study, causing their code to stop working. If users explicitly specify `study_name=None`, it might make it easier for them to understand what is happening.

c-bata:
We discussed internally with other Optuna committers and decided to close this pull request due to the reason I commented above. Thank you again for your proposal!

PR #6021, opened 2025-03-27
Made error message for `create_study`'s direction easier to understand `optuna.study`

sinano1107:
issue: #5806

## Motivation

Users sometimes makes the mistake of passing the list of directions to the direction argument. The error message for this mistake is hard to understand what is the problem.

- Correct
```python
optuna.create_study(directions=["minimize", "minimize"])
```
- Wrong

Using direction instead of directions

```python
optuna.create_study(direction=["minimize", "minimize"])
```
```
ValueError: Please set either 'minimize' or 'maximize' to direction. You can also set the corresponding `StudyDirection` member.
```

## Description of the changes
The error message displayed when a list object, is assigned to the direction parameter of create_study has been changed as follows.

- previous
```
ValueError: Please set either 'minimize' or 'maximize' to direction. You can also set the corresponding `StudyDirection` member.
```

- new
```
ValueError: For multi-objective optimization, using `directions` instead of `direction`.
```

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1256,8 +1256,8 @@ def create_study(
         for d in directions
     ):
         raise ValueError(
-            "Please set either 'minimize' or 'maximize' to direction. You can also set the "
-            "corresponding `StudyDirection` member."
+            f"`directions` must be a list of `minimize` or `maximize`, but got {directions}. "
+            "For single-objective optimization, please use `direction` instead of `directions`."
         )
 
     direction_objects = [
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1239,6 +1239,10 @@ def create_study(
     elif direction is not None and directions is not None:
         raise ValueError("Specify only one of `direction` and `directions`.")
     elif direction is not None:
+        if isinstance(direction, Sequence) and not isinstance(direction, str):
+            raise ValueError(
+                "For multi-objective optimization, using `directions` instead of `direction`."
+            )
         directions = [direction]
     elif directions is not None:
         directions = list(directions)
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -1241,7 +1241,7 @@ def create_study(
     elif direction is not None:
         if isinstance(direction, Sequence) and not isinstance(direction, str):
             raise ValueError(
-                "For multi-objective optimization, using `directions` instead of `direction`."
+                "Use `directions` instead of `direction` for multi-objective optimization."
             )
         directions = [direction]
     elif directions is not None:
--- a/tests/study_tests/test_study.py
+++ b/tests/study_tests/test_study.py
@@ -154,6 +154,9 @@ def test_optimize_with_direction() -> None:
     with pytest.raises(ValueError):
         create_study(direction=["maximize", "minimize"])  # type: ignore [arg-type]
 
+    with pytest.raises(ValueError):
+        create_study(directions="minimize")
+
 
 @pytest.mark.parametrize("n_trials", (0, 1, 20))
 @pytest.mark.parametrize("n_jobs", (1, 2, -1))
--- a/tests/study_tests/test_study.py
+++ b/tests/study_tests/test_study.py
@@ -152,8 +152,6 @@ def test_optimize_with_direction() -> None:
         create_study(direction="test")
 
     with pytest.raises(ValueError):
-        # Type hintingly wrong input.
-        # Make mypy detection be ignored.
         create_study(direction=["maximize", "minimize"])  # type: ignore [arg-type]
 
 
--- a/tests/study_tests/test_study.py
+++ b/tests/study_tests/test_study.py
@@ -148,8 +148,19 @@ def test_optimize_with_direction() -> None:
     assert study.direction == StudyDirection.MAXIMIZE
     check_study(study)
 
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError) as e:
         create_study(direction="test")
+    assert str(e.value) == (
+        "Please set either 'minimize' or 'maximize' to direction. You can also set the "
+        + "corresponding `StudyDirection` member."
+    )
+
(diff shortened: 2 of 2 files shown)


optuna/study/study.py
```diff
@@ -1239,6 +1239,10 @@ def objective(trial):
     elif direction is not None and directions is not None:
         raise ValueError("Specify only one of `direction` and `directions`.")
     elif direction is not None:
+        if isinstance(direction, list):
```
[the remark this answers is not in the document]
sinano1107: I agree with you!

sinano1107:
For some reason docs/readthedocs.org:optuna now fails.
If you know anything about this, I would appreciate it.

sinano1107:
I thought the docs/readthedocs.org:optuna might be failing because I used `git push -force-with-lease`, so I created a new #6022, it also has the same error.

Sorry for the mess in the history and PR section.

In the end I don't know the cause either.

sinano1107:
I don't know why, but it didn't seem to pinch this time!

nzw0301 (commented):
Yeah, CI jobs sometimes fail due to an external service. Sorry for confusing you. 

In terms of PR, could you add a test?

sinano1107:
I added a test in the form of an extension to `test_optimize_with_direction` `test_create_study_with_multi_objectives`, but I feel it is a little less prospective.
Let's change it to create a new function like `test_create_study_with_invalid_args` to test it if necessary. Which would you prefer?

HideakiImamura:
@nzw0301 Could you continue to review this PR?

tests/study_tests/test_study.py
```diff
@@ -148,8 +148,19 @@ def test_optimize_with_direction() -> None:
     assert study.direction == StudyDirection.MAXIMIZE
     check_study(study)
 
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError) as e:
         create_study(direction="test")
+    assert str(e.value) == (
+        "Please set either 'minimize' or 'maximize' to direction. You can also set the "
+        + "corresponding `StudyDirection` member."
+    )
```
HideakiImamura, line 156: IMO, we should not test the exact contents of the error message since it is not a specification, so might be changed later. It would be great to confirm that the error is emitted.

tests/study_tests/test_study.py
```diff
@@ -148,8 +148,19 @@ def test_optimize_with_direction() -> None:
...
+
+    with pytest.raises(ValueError) as e:
+        create_study(direction=["maximize", "minimize"])
+    assert (
+        str(e.value)
+        == "For multi-objective optimization, using `directions` instead of `direction`."
+    )
```
HideakiImamura, line 163: Ditto.

tests/study_tests/test_study.py
```diff
@@ -1152,15 +1163,18 @@ def test_create_study_with_multi_objectives() -> None:
     assert study.directions == [StudyDirection.MAXIMIZE, StudyDirection.MINIMIZE]
     assert study._is_multi_objective()
 
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError) as e:
         # Empty `direction` isn't allowed.
         _ = create_study(directions=[])
+    assert str(e.value) == "The number of objectives must be greater than 0."
```
HideakiImamura, line 1169: Ditto.

tests/study_tests/test_study.py
```diff
@@ -1152,15 +1163,18 @@ def test_create_study_with_multi_objectives() -> None:
...
 
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError) as e:
         _ = create_study(direction="minimize", directions=["maximize"])
+    assert str(e.value) == "Specify only one of `direction` and `directions`."
```
HideakiImamura, line 1173: Ditto.

tests/study_tests/test_study.py
```diff
@@ -1152,15 +1163,18 @@ def test_create_study_with_multi_objectives() -> None:
...
 
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError) as e:
         _ = create_study(direction="minimize", directions=[])
+    assert str(e.value) == "Specify only one of `direction` and `directions`."
```
HideakiImamura, line 1177: Ditto.

HideakiImamura (commented):

optuna/study/study.py
```diff
@@ -1239,6 +1239,10 @@ def objective(trial):
     elif direction is not None and directions is not None:
         raise ValueError("Specify only one of `direction` and `directions`.")
     elif direction is not None:
+        if isinstance(direction, Sequence) and not isinstance(direction, str):
+            raise ValueError(
+                "For multi-objective optimization, using `directions` instead of `direction`."
```
nabenabe0928, line 1244: [nit]
```suggestion
                "Use `directions` instead of `direction` for multi-objective optimization."
```

sinano1107:
@HideakiImamura 
Yes, I agree, but such a test would not detect the changes made in this PR. The code before the changes are made will still pass the test.

Do you think it is ok for us to ignore it?

nabenabe0928 (commented):
I think this part should also be modified.
https://github.com/optuna/optuna/blob/896e0ca8e35419a3b6d2de9ef6e6652f6329cf5b/optuna/study/study.py#L1258-L1261

Could you, @sinano1107 , please check [the comment here](https://github.com/optuna/optuna/pull/5983#discussion_r1960955322) as well?

sinano1107:
@nabenabe0928 
Does this edit captured your intentions?

nabenabe0928:
@sinano1107 
Yes, the change looks good to me:)
Could you add the unit test for this change as well?
Namely, could you add a test using `directions="minimize"`, which is an invalid input.

sinano1107:
@nabenabe0928 
I'm happy to hear that from you.
I have added a test, please check it out.

nabenabe0928 (approved):
LGTM!

PR #6023, opened 2025-03-28
Introduce server-side TTL cache in `GrpcStorageProxy`

hitsgub:
## Motivation
To make `GrpcStorageProxy` faster by introducing server-side TTL cache.
In the gRPC storage proxy, if a client reads get_all_trials continuously for a short period of time (e.g. 1 sec), the state of the RDB is expected to not change.
TTL cache for trials can be introduced on the server side to speed up the process. 

## Description of the changes
This PR makes `GrpcStorageProxy` faster by cache `OptunaStorageProxyService.GetTrials` calls. Here is the performance comparison results and the benchmark script I used.

### Results
|  | `master` branch | This PR (ttl_cache_seconds = 10) |
| --- | ------------- | ------------- |
| total | 859.9sec | 233.1sec |
| `client.py:368(get_all_trials)` | 705.3sec | 16.83sec |
| stats | image | image |

<details>
<summary>client.py</summary>

```python

import argparse
import cProfile
import time
import optuna
from typing import Literal
from optuna.storages import GrpcStorageProxy, BaseStorage
from time import sleep
from uuid import uuid4


study_name = "ttl" + str(uuid4())[:8]
n_trials = 1000

STORAGE_KIND = Literal["grpc", "rdb", "middleware"]
STORAGE_URL = "YOUR_STORAGE_URL"


def get_storage(storage_kind: STORAGE_KIND) -> BaseStorage:
    if storage_kind == "grpc":
        return GrpcStorageProxy(host="localhost", port=13000)
    elif storage_kind == "rdb":
        return optuna.storages.get_storage(STORAGE_URL)


    s = 0.0
    for i in range(10):
        trial.set_user_attr(f"attr{i}", "attr value")
    s += trial.suggest_float("x", -10, 10) ** 2
    return s


def study_optimize(storage_kind: STORAGE_KIND, n_trials: int) -> int:
    start = time.time()
    storage = get_storage(storage_kind)
    while True:
        try:
            study = optuna.load_study(storage=storage, study_name=study_name, sampler=optuna.samplers.RandomSampler())
            break
        except:
            sleep(10)
    study.optimize(objective, callbacks=[optuna.study.MaxTrialsCallback(n_trials)])
    elapsed = time.time() - start
    return elapsed


def bench_study_optimize(storage_kind: STORAGE_KIND) -> list[int]:
    elapsed = []
    _study = optuna.create_study(
        storage=get_storage("rdb"),
        study_name=study_name,
        load_if_exists=True,
        sampler=optuna.samplers.RandomSampler(),
    )
    study_optimize(storage_kind, n_trials)
    return elapsed


def main(storage: STORAGE_KIND, benchmark_file: str):
    try:
        optuna.delete_study(storage=get_storage("rdb"), study_name=study_name)
    except:
        pass
    cProfile.runctx(
        f"bench_study_optimize('{storage}')",
        globals(),
        locals(),
        filename=benchmark_file,
    )


if __name__ == "__main__":
    parser.add_argument("--storage", type=str, required=True)
    parser.add_argument("--benchmark_file", type=str, required=True)
    args = parser.parse_args()
    main(args.storage, args.benchmark_file)
```

</details>

<details>
<summary>server.py</summary>

```

import argparse
from optuna.storages import run_grpc_proxy_server
import optuna
import signal


STORAGE_URL = "YOUR_STORAGE_URL"

def handler(signum, frame):
    print("Received signal", signum)
    exit(0)

if __name__ == "__main__":
    parser.add_argument("--ttl-cache-seconds", type=float, default=None)
    args = parser.parse_args()
    signal.signal(signal.SIGINT, handler)
    signal.signal(signal.SIGTERM, handler)

    run_grpc_proxy_server(
        optuna.storages.RDBStorage(STORAGE_URL),
        host="localhost",
        port=13000,
        ttl_cache_seconds=args.ttl_cache_seconds,
    )
```

</details>

### DIFF
--- a/optuna/storages/_grpc/servicer.py
+++ b/optuna/storages/_grpc/servicer.py
@@ -350,11 +350,12 @@ class OptunaStorageProxyService(api_pb2_grpc.StorageServiceServicer):
         if self._ttl_cache_seconds is None:
             trials = self._get_all_trials(study_id, context)
         else:
-            now = time.time()
-            if now >= self._cached_trials[study_id].last_sync_time + self._ttl_cache_seconds:
-                _logger.info("Syncing trials for study_id=%d", study_id)
-                self._cached_trials[study_id].trials = self._get_all_trials(study_id, context)
-                self._cached_trials[study_id].last_sync_time = now
+            with self._lock:
+                now = time.time()
+                if now >= self._cached_trials[study_id].last_sync_time + self._ttl_cache_seconds:
+                    _logger.info("Syncing trials for study_id=%d", study_id)
+                    self._cached_trials[study_id].trials = self._get_all_trials(study_id, context)
+                    self._cached_trials[study_id].last_sync_time = now
             trials = self._cached_trials[study_id].trials
         filtered_trials = [
             _to_proto_trial(t)
--- a/optuna/storages/_grpc/servicer.py
+++ b/optuna/storages/_grpc/servicer.py
@@ -42,7 +42,7 @@ class CachedTrials:
 
 
 class OptunaStorageProxyService(api_pb2_grpc.StorageServiceServicer):
-    def __init__(self, storage: BaseStorage, ttl_seconds: float = 0.0) -> None:
+    def __init__(self, storage: BaseStorage, ttl_seconds: float = 10.0) -> None:
         self._backend = storage
         self._lock = threading.Lock()
         self._ttl_seconds = ttl_seconds
--- a/optuna/storages/_grpc/servicer.py
+++ b/optuna/storages/_grpc/servicer.py
@@ -42,10 +42,10 @@ class CachedTrials:
 
 
 class OptunaStorageProxyService(api_pb2_grpc.StorageServiceServicer):
-    def __init__(self, storage: BaseStorage, ttl_seconds: float = 10.0) -> None:
+    def __init__(self, storage: BaseStorage, ttl_cache_seconds: float | None = None) -> None:
         self._backend = storage
         self._lock = threading.Lock()
-        self._ttl_seconds = ttl_seconds
+        self._ttl_cache_seconds = ttl_cache_seconds
         self._cached_trials: defaultdict[int, CachedTrials] = defaultdict(CachedTrials)
 
     def CreateNewStudy(
@@ -332,6 +332,13 @@ class OptunaStorageProxyService(api_pb2_grpc.StorageServiceServicer):
 
         return api_pb2.GetTrialReply(trial=_to_proto_trial(trial))
 
+    def _get_all_trials(self, study_id: int, context: grpc.ServicerContext) -> list[FrozenTrial]:
+        try:
+            trials = self._backend.get_all_trials(study_id, deepcopy=False)
+        except KeyError as e:
+            context.abort(code=grpc.StatusCode.NOT_FOUND, details=str(e))
+        return trials
+
     def GetTrials(
         self,
         request: api_pb2.GetTrialsRequest,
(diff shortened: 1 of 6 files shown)


HideakiImamura:
Thanks for the PR! The change basically looks good to me. Could you tell me the value of `ttl_cache_seconds` in the benchmark comparison?

HideakiImamura:

HideakiImamura (changes requested):
Could you add a test case to validate the added argument?

tests/storages_tests/test_storages.py
```diff
@@ -768,6 +771,44 @@ def test_get_all_trials(storage_mode: str) -> None:
             storage.get_all_trials(non_existent_study_id)
 
 
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES_GRPC)
+@pytest.mark.parametrize("ttl_cache_seconds", [None, 1.0])
+def test_get_all_trials_grpc_proxy_server_ttl_cache(
+    storage_mode: str, ttl_cache_seconds: float | None
+) -> None:
```
HideakiImamura, line 778: I am concerned that there are two if statements inside the test depending on whether ttl_cache_seconds is None or not, so why not separate the test functions for cases where ttl_cache_seconds is None and 1.0?

hitsgub: thanks, I will separate them

HideakiImamura (dismissed):
Thanks! LGTM!

optuna/storages/_grpc/servicer.py
```diff
@@ -329,11 +347,15 @@ def GetTrials(
...
-
+        if self._ttl_cache_seconds is None:
+            trials = self._get_all_trials(study_id, context)
+        else:
+            now = time.time()
+            if now >= self._cached_trials[study_id].last_sync_time + self._ttl_cache_seconds:
+                _logger.info("Syncing trials for study_id=%d", study_id)
+                self._cached_trials[study_id].trials = self._get_all_trials(study_id, context)
+                self._cached_trials[study_id].last_sync_time = now
+            trials = self._cached_trials[study_id].trials
```
c-bata, line 358: In real-world use cases, it's possible for multiple Kubernetes pods to send `GetAllTrials` requests simultaneously. Since `self._get_all_trials(study_id, context)` is a time-consuming operation, the cache becomes ineffective in such situations. For example, if 10 requests arrive at the same time, `self._get_all_trials(study_id, context)` will be called 10 times.

To prevent this, you could wrap the relevant section with a `threading.Lock` to ensure that only one thread performs the `self._get_all_trials` call while others wait for the result to be cached.

hitsgub: thanks, I fixed it.

hitsgub:
The file for test added in this PR will also be changed according to the following PR.

https://github.com/optuna/optuna/pull/6044#discussion_r2046187267

optuna/storages/_grpc/servicer.py
```diff
@@ -329,11 +347,16 @@ def GetTrials(
...
-        except KeyError as e:
-            context.abort(code=grpc.StatusCode.NOT_FOUND, details=str(e))
-
+        if self._ttl_cache_seconds is None:
+            trials = self._get_all_trials(study_id, context)
+        else:
+            with self._lock:
+                now = time.time()
+                if now >= self._cached_trials[study_id].last_sync_time + self._ttl_cache_seconds:
+                    _logger.info("Syncing trials for study_id=%d", study_id)
```
c-bata, line 356: Nit: Users don't really need to be aware of the cache here, so I think logger.debug might be more appropriate. What do you think?

```suggestion
                    _logger.debug("Syncing trials for study_id=%d", study_id)
```

c-bata (commented):
The changes look almost good to me. I'm now running the additional benchmark scenario. I'll approve once it finishes.

optuna/testing/storages.py
```diff
@@ -40,6 +40,11 @@
     "cached_sqlite",
 ]
 
+STORAGE_MODES_GRPC = [
+    "grpc_rdb",
+    "grpc_journal_file",
+]
+
```
c-bata, line 47: Nit: how about moving this variable to `test_grpc.py`?

hitsgub:
完了したstudyがちゃんと消えることを確認
なんらかの理由で途中でworkflowを潰した場合は…放置するか手動でなんとかするしかないですね

hitsgub:
When ttl cache is enabled and parallel processing is used, it has been observed that multiple processes may generate trials with the same trial_id.

The fundamental solution is:
• Stop using sequential trial_ids and instead use UUIDs, then rewrite all trial_id ordering-dependent logic to use timestamps.

As a provisional solution:
• When calling get_all_trials within create_new_trial, force synchronization with the server by ignoring the ttl cache.

HideakiImamura:
@hitsgub Thanks for the update and comments. I would like to discuss that problem (sampling same trial ids with ttl cache) inside committers. Please wait a moment.

PR #6026, opened 2025-03-31
Fix CI

nabenabe0928:
## Motivation

This PR fixes the CI by removing unused global referencing.

### DIFF
--- a/tests/artifacts_tests/test_gcs.py
+++ b/tests/artifacts_tests/test_gcs.py
@@ -31,7 +31,6 @@ class MockBucket:
         return MockBlob(blob_name)
 
     def delete_blob(self, blob_name: str) -> None:
-        global _MOCK_BUCKET_CONTENT
         del _MOCK_BUCKET_CONTENT[blob_name]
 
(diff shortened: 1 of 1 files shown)


nabenabe0928:
@y0z @c-bata 
Could you review this PR?

tests/artifacts_tests/test_gcs.py
```diff
@@ -31,7 +31,6 @@ def blob(self, blob_name: str) -> "MockBlob":
         return MockBlob(blob_name)
 
     def delete_blob(self, blob_name: str) -> None:
-        global _MOCK_BUCKET_CONTENT
```
nabenabe0928, line 34: Although the dict itself is not a global variable, the contents are globally editable.

```python
D = {}

def add():
    D["a"] = 1

def delete():
    del D["a"]
```

c-bata (approved):
I confirmed that this PR fixes the following error.
https://github.com/optuna/optuna/actions/runs/14160851530/job/39666074487

LGTM!

PR #6033, opened 2025-04-07
Update vsbx

hrntsm:
## Motivation

Improve vSBX based on #6000, #6008

## Description of the changes

I have changed the implementation to make it possible to create a distribution of child individuals as in the reference paper.
See #6000 for details.

In order to align the paper with the results, the implementation has been changed and the crossover results are no longer compatible with the previous results. Therefore, the test is being rewritten.
I understand that this is not desirable for Optuna.
On the other hand, the fact that the implementation does not match the reference paper is considered a bug, so I create this PR.

Any comments on what form it should take would be appreciated.

### DIFF
--- a/optuna/samplers/nsgaii/_crossovers/_vsbx.py
+++ b/optuna/samplers/nsgaii/_crossovers/_vsbx.py
@@ -64,6 +64,8 @@ class VSBXCrossover(BaseCrossover):
         uniform_crossover_prob: float = 0.5,
         use_child_gene_prob: float = 0.5,
     ) -> None:
+        if (eta is not None) and (eta < 0.0):
+            raise ValueError("The value of `eta` must be greater than or equal to 0.0.")
         self._eta = eta
 
         if uniform_crossover_prob < 0.0 or uniform_crossover_prob > 1.0:
@@ -89,9 +91,10 @@ class VSBXCrossover(BaseCrossover):
         else:
             eta = self._eta
 
+        eps = 1e-10
         us = rng.rand(len(search_space_bounds))
-        beta_1 = np.power(1 / (2 * us), 1 / (eta + 1))
-        beta_2 = np.power(1 / (2 * (1 - us)), 1 / (eta + 1))
+        beta_1 = np.power(1 / np.maximum((2 * us), eps), 1 / (eta + 1))
+        beta_2 = np.power(1 / np.maximum((2 * (1 - us)), eps), 1 / (eta + 1))
 
         u_1 = rng.rand()
         if u_1 <= 0.5:
--- a/optuna/samplers/nsgaii/_crossovers/_vsbx.py
+++ b/optuna/samplers/nsgaii/_crossovers/_vsbx.py
@@ -104,9 +104,9 @@ class VSBXCrossover(BaseCrossover):
         else:
             c2 = 0.5 * (-(1 - beta_1) * parents_params[0] + (3 - beta_2) * parents_params[1])
 
-        # vSBX applies crossover with establishment 0.5, and with probability 0.5,
+        # vSBX applies crossover with use_child_gene_prob and uniform_crossover_prob.
         # the gene of the parent individual is the gene of the child individual.
-        # The original SBX creates two child individuals,
+        # The original vSBX creates two child individuals,
         # but optuna's implementation creates only one child individual.
         # Therefore, when there is no crossover,
         # the gene is selected with equal probability from the parent individuals x1 and x2.
--- a/optuna/samplers/nsgaii/_crossovers/_vsbx.py
+++ b/optuna/samplers/nsgaii/_crossovers/_vsbx.py
@@ -20,6 +20,12 @@ class VSBXCrossover(BaseCrossover):
     vSBX generates child individuals without excluding any region of the parameter space,
     while maintaining the excellent properties of SBX.
 
+    In the paper, vSBX has only one argument, ``eta``,
+    and generate two child individuals.
+    However, Optuna can only return one child individual in one crossover operation,
+    so it uses the ``uniform_crossover_prob`` and ``use_child_gene_prob`` arguments
+    to make two individuals into one.
+
     - `Pedro J. Ballester, Jonathan N. Carter.
       Real-Parameter Genetic Algorithms for Finding Multiple Optimal Solutions
       in Multi-modal Optimization. GECCO 2003: 706-717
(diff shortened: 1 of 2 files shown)


tests/samplers_tests/test_nsgaii.py
```diff
@@ -865,7 +866,6 @@ def test_crossover_inlined_categorical_distribution() -> None:
         BLXAlphaCrossover(),
         SPXCrossover(),
         SBXCrossover(),
-        VSBXCrossover(),
```
hrntsm, line 868: Removed because even if the parent has the same score, the child is created in its neighborhood and the purpose of this test, to be the same individual as the parent, is not achieved.

tests/samplers_tests/test_nsgaii.py
```diff
@@ -902,7 +902,7 @@ def test_crossover_duplicated_param_values(crossover: BaseCrossover) -> None:
         (SBXCrossover(), 0.0, np.array([2.0, 3.0])),  # c1 = (p1 + p2) / 2.
         (SBXCrossover(), 0.5, np.array([3.0, 4.0])),  # p2.
         (SBXCrossover(), 1.0, np.array([3.0, 4.0])),  # p2.
-        (VSBXCrossover(), 0.0, np.array([2.0, 3.0])),  # c1 = (p1 + p2) / 2.
+        (VSBXCrossover(), 0.0, np.array([-inf, -inf])),  # c1 = (p1 + p2) / 2.
```
hrntsm, line 905: Random is used in `beta_1`, the variable handled in the crossover calculation. If the value of this random is 0, the zero division occurs, so it is inf.
The random variable will be the one corresponding to `u` in the paper and `us` in the implementation. The range of `u` is [0,1] in the reference paper, and this problem occurs.

y0z:
@sawa3030 Could you review this PR? (cc @HideakiImamura )

optuna/samplers/nsgaii/_crossovers/_vsbx.py
```diff
@@ -30,13 +36,45 @@ class VSBXCrossover(BaseCrossover):
...
     n_parents = 2
 
-    def __init__(self, eta: float | None = None) -> None:
+    def __init__(
+        self,
+        eta: float | None = None,
+        uniform_crossover_prob: float = 0.5,
+        use_child_gene_prob: float = 0.5,
+    ) -> None:
         self._eta = eta
```
y0z, line 67: Is the valid range for $\eta$ is $\geq 0$? I want to validate the value here since I find that setting $\eta = -1$ causes ZeroDivisionError. This may also happen in SBXCrossOver.
```python
    beta_1 = np.power(1 / (2 * us), 1 / (eta + 1))
                                    ~~^~~~~~~~~~~
ZeroDivisionError: division by zero
```

optuna/samplers/nsgaii/_crossovers/_vsbx.py
```diff
@@ -52,37 +90,47 @@ def crossover(
...
-        c1[mask] = (
-            0.5 * ((1 - beta_1) * parents_params[0] + (1 + beta_1) * parents_params[1])[mask]
-        )
-        c2 = 0.5 * ((3 - beta_2) * parents_params[0] - (1 - beta_2) * parents_params[1])
-        c2[mask] = (
-            0.5 * (-(1 - beta_2) * parents_params[0] + (3 - beta_2) * parents_params[1])[mask]
-        )
-
-        # vSBX applies crossover with establishment 0.5, and with probability 0.5,
+        beta_1 = np.power(1 / (2 * us), 1 / (eta + 1))
```
y0z, line 93: Although it is a very small probability, there is a chance of becoming the denominator in `1 / (2 * us)` can be zero since `rng.rand()` $\in [0, 1)$ ([ref](https://numpy.org/doc/stable/reference/random/generated/numpy.random.rand.html)).

The below might be safer. (This pseudocode might require modification since us is vector whereas eps is scalar.)
```
        beta_1 = np.power(1 / max((2 * us), eps), 1 / (eta + 1))
```

tests/samplers_tests/test_nsgaii.py
y0z: I consider that avoiding division by zero is better in the implementation since it fails optimization.
```python
  File "/Users/yozaki/work/review/optuna/optuna/samplers/nsgaii/_crossovers/_vsbx.py", line 94, in crossover
    beta_1 = np.power(1 / (2 * us), 1 / (eta + 1))
                      ~~^~~~~~~~~~
ZeroDivisionError: division by zero
```

optuna/samplers/nsgaii/_crossovers/_vsbx.py
hrntsm: I could not find a description of the range of `eta` in the references listed in the document, but the following ones mention that it is a non-negative value.
Add a check for non-negative values, since there is also the issue of zero divisions.

[Self-adaptive simulated binary crossover for real-parameter optimization](https://dl.acm.org/doi/10.1145/1276958.1277190)

I will implement the same for the SBX as a another PR.

optuna/samplers/nsgaii/_crossovers/_vsbx.py
hrntsm: There is no strong reason for the `eps`, but I have chosen to use the same values used in the SBX.

https://github.com/optuna/optuna/blob/19296c387c5376c137f9c3afc4d18159e70bdc9c/optuna/samplers/nsgaii/_crossovers/_sbx.py#L99

Although out of the range of values actually generated, 1 is set as us value in the test, and since zero divisions occur even in beta_2, a similar correction has been made.

```
eta_2 = np.power(1 / (2 * (1 - us)), 1 / (eta + 1))
```

tests/samplers_tests/test_nsgaii.py
```diff
@@ -902,7 +901,11 @@ def test_crossover_duplicated_param_values(crossover: BaseCrossover) -> None:
         (SBXCrossover(), 0.0, np.array([2.0, 3.0])),  # c1 = (p1 + p2) / 2.
         (SBXCrossover(), 0.5, np.array([3.0, 4.0])),  # p2.
         (SBXCrossover(), 1.0, np.array([3.0, 4.0])),  # p2.
-        (VSBXCrossover(), 0.0, np.array([2.0, 3.0])),  # c1 = (p1 + p2) / 2.
+        (
+            VSBXCrossover(),
+            0.0,
+            np.array([-1076.02679423, -2151.84728898]),
```
y0z, line 907: Question: Where do these values come from? I have no idea about what is tested here.

hrntsm: Based on https://github.com/optuna/optuna/pull/6033#discussion_r2032211800, I modified the following so that zero divisions do not occur.

```
beta_1 = np.power(1 / np.maximum((2 * us), eps), 1 / (eta + 1))
```

The original test was [-inf, -inf] because of the zero division, but has been corrected as above.

However, beta_1 still be a very large value, so vsbx generate a child on the -inf side. The value tested here is considered to be a result that depends on the eps value.

hrntsm: I followed the original test with a random value of 0 and adjusted the test results, but as you pointed out, this is not a very meaningful test, so it may be more appropriate to specify a different meaningful random value.

sawa3030: I think it's important to keep the test with `us = 0` since it covers an edge case. That said, I agree that the test is ambiguous. How about adding a comment to clarify the behavior, as you mentioned [here](https://github.com/optuna/optuna/pull/6033#discussion_r2044246979)?

hrntsm: I added a comment on the test content.

sawa3030 (approved):

PR #6039, opened 2025-04-10
Add a module to preprocess solutions for hypervolume improvement calculation

nabenabe0928:
## Motivation

This PR adds a module to preprocess for hypervolume improvement based on the papers `A Box Decomposition Algorithm to Compute the Hypervolume Indicator` and `Towards Efficient Multiobjective Optimization: Multiobjective statistical criterions`.

### `_get_upper_bound_set`

Basically, the main algorithm (`_get_upper_bound_set`) is based on the following pseudocode quoted from Algorithm 2 of [1]:

image

Figure 1 of [1] visualizes `upper_bound_set` as the black dots and the Pareto solutions (lower is better) as a cross.

image

### `_get_hyper_rectangle_bounds`

This function calculates a set of diagonal points in the gray rectangles in the figure above, which is based the equation (2) in [1]:

image

One of `B(u)` corresponds to a gray rectangle.

### `_get_non_dominated_hyper_rectangle_bounds`

This calculates a set of diagonal points in the white rectangles in Figure 1 of [2].
To do so, we perform `_get_upper_bound_set` twice: once for $l_k$ below and the other for $u_k$ below.

image

> [!NOTE]
> [2] considers the maximization problem unlike Optuna and [1].

[1] [A Box Decomposition Algorithm to Compute the Hypervolume Indicator](https://arxiv.org/abs/1510.01963)
[2] [Differentiable Expected Hypervolume Improvement for Parallel Multi-Objective Bayesian Optimization](https://arxiv.org/abs/2006.05078)

PR #6052, opened 2025-04-16
Add `AcquisitionFuncParams` for LogEHVI

nabenabe0928:
## Motivation

This PR adds an interface for LogEHVI, a multi-objective acquisition function.

Although this PR passes the CI standalone, this PR should be merged after
- https://github.com/optuna/optuna/pull/6039

### DIFF
--- a/optuna/_gp/acqf.py
+++ b/optuna/_gp/acqf.py
@@ -186,16 +186,16 @@ class MultiObjectiveAcquisitionFunctionParams(AcquisitionFunctionParams):
         non_dominated_box_lower_bounds, non_dominated_box_upper_bounds = (
             _get_non_dominated_box_bounds()
         )
-        # The inverse_squared_lengthscales below is used only in optim_mixed.
-        inverse_squared_lengthscales_array = np.array(
-            [
-                acqf_params.kernel_params.inverse_squared_lengthscales.detach().numpy()
-                for acqf_params in acqf_params_for_objectives
-            ]
-        )
         # Since all the objectives are equally important, we simply use the mean of
         # inverse of squared mean lengthscales over all the objectives.
-        mean_lengthscales = np.mean(1 / np.sqrt(inverse_squared_lengthscales_array), axis=0)
+        mean_lengthscales = np.mean(
+            [
+                1
+                / np.sqrt(acqf_params.kernel_params.inverse_squared_lengthscales.detach().numpy())
+                for acqf_params in acqf_params_for_objectives
+            ],
+            axis=0,
+        )
         dummy_kernel_params = KernelParamsTensor(
             # inverse_squared_lengthscales is used in optim_mixed.py.
             # cf. https://github.com/optuna/optuna/blob/v4.3.0/optuna/_gp/optim_mixed.py#L200-L209
--- a/optuna/_gp/acqf.py
+++ b/optuna/_gp/acqf.py
@@ -54,8 +54,9 @@ def logehvi(
         torch.minimum(Y_post[..., torch.newaxis, :], non_dominated_box_upper_bounds)
         - non_dominated_box_lower_bounds,
     )
-    log_hvi_vals = torch.special.logsumexp(diff.log().sum(dim=-1), dim=-1)
-    return -log_n_qmc_samples + torch.special.logsumexp(log_hvi_vals, dim=-1)
+    # NOTE(nabenabe): logsumexp with dim=-1 is for the HVI calculation and that with dim=-2 is for
+    # expectation of the HVIs over the fixed_samples.
+    return torch.special.logsumexp(diff.log().sum(dim=-1), dim=(-2, -1)) - log_n_qmc_samples
 
 
 def standard_logei(z: torch.Tensor) -> torch.Tensor:
--- a/optuna/_gp/acqf.py
+++ b/optuna/_gp/acqf.py
@@ -25,13 +25,14 @@ else:
 
 
 def _sample_from_normal_sobol(dim: int, n_samples: int, seed: int | None) -> torch.Tensor:
-    # cf. https://github.com/pytorch/botorch/blob/v0.13.0/botorch/sampling/qmc.py#L26
-    sobol_engine = torch.quasirandom.SobolEngine(
-        dimension=dim, scramble=True, seed=seed
-    )  # type: ignore
-    # The Sobol sequence in [-1, 1].
-    samples = 2.0 * (sobol_engine.draw(n_samples, dtype=torch.float64) - 0.5)
-    # Inverse transform to standard normal (values to close to 0/1 result in inf values).
+    # NOTE(nabenabe): Normal Sobol sampling based on BoTorch.
+    # https://github.com/pytorch/botorch/blob/v0.13.0/botorch/sampling/qmc.py#L26-L97
+    # https://github.com/pytorch/botorch/blob/v0.13.0/botorch/utils/sampling.py#L109-L138
+    sobol_samples = torch.quasirandom.SobolEngine(dimension=dim, scramble=True, seed=seed).draw(  # type: ignore[no-untyped-call]
+        n_samples, dtype=torch.float64
+    )
+    samples = 2.0 * (sobol_samples - 0.5)  # The Sobol sequence in [-1, 1].
+    # Inverse transform to standard normal (values to close to -1 or 1 result in infinity).
     return torch.erfinv(samples) * float(np.sqrt(2))
 
 
--- a/optuna/_gp/acqf.py
+++ b/optuna/_gp/acqf.py
@@ -165,7 +165,7 @@ class MultiObjectiveAcquisitionFunctionParams(AcquisitionFunctionParams):
             pareto_sols: np.ndarray, ref_point: np.ndarray
         ) -> tuple[np.ndarray, np.ndarray]:
             # NOTE(nabenabe): This is a dummy implementation for formatter.
-            return np.zeros((1, Y.shape[-1])), np.zeros((1, Y.shape[-1]))
+            return np.ones((5, Y.shape[-1])) * 5.0, np.ones((5, Y.shape[-1])) * 10.0
 
         def _get_non_dominated_box_bounds() -> tuple[torch.Tensor, torch.Tensor]:
             loss_vals = -Y  # NOTE(nabenabe): Y is to be maximized, loss_vals is to be minimized.
@@ -252,7 +252,7 @@ def _eval_ehvi(
     )
     Y_post = []
     fixed_samples = evhi_acqf_params.fixed_samples
-    for acqf_params in evhi_acqf_params.acqf_params_for_objectives:
+    for i, acqf_params in enumerate(evhi_acqf_params.acqf_params_for_objectives):
         mean, var = posterior(
             kernel_params=acqf_params.kernel_params,
             X=X,
@@ -264,14 +264,11 @@ def _eval_ehvi(
         stdev = torch.sqrt(var + evhi_acqf_params.acqf_stabilizing_noise)
         # NOTE(nabenabe): By using fixed samples from the Sobol sequence, EHVI becomes
         # deterministic, making it possible to optimize the acqf by l-BFGS.
-        Y_post.append(
-            mean[..., torch.newaxis, :]
-            + stdev[..., torch.newaxis, :] * fixed_samples[..., torch.newaxis, :, :]
-        )
+        Y_post.append(mean[..., torch.newaxis] + stdev[..., torch.newaxis] * fixed_samples[..., i])
 
     # NOTE(nabenabe): Use the following once multi-task GP is supported.
     # L = torch.linalg.cholesky(cov)
-    # Y_post = means[..., torch.newaxis, :] + torch.einsum("...MM,NM->...NM", L, fixed_samples)
+    # Y_post = means[..., torch.newaxis, :] + torch.einsum("...MM,SM->...SM", L, fixed_samples)
     return logehvi(
         Y_post=torch.stack(Y_post, dim=-1),
         non_dominated_box_lower_bounds=evhi_acqf_params.non_dominated_box_lower_bounds,
--- a/optuna/_gp/acqf.py
+++ b/optuna/_gp/acqf.py
@@ -36,16 +36,16 @@ def _sample_from_normal_sobol(dim: int, n_samples: int, seed: int | None) -> tor
 
 def logehvi(
     Y_post: torch.Tensor,  # (..., n_qmc_samples, n_objectives)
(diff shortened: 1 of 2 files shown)


optuna/_gp/acqf.py
```diff
@@ -117,6 +151,72 @@ def from_acqf_params(
...
+        def get_non_dominated_box_bounds(
+            pareto_sols: np.ndarray, ref_point: np.ndarray
+        ) -> tuple[np.ndarray, np.ndarray]:
+            # NOTE(nabenabe): This is a dummy implementation for formatter.
+            return np.ones((5, Y.shape[-1])) * 5.0, np.ones((5, Y.shape[-1])) * 10.0
+
+        def _get_non_dominated_box_bounds() -> tuple[torch.Tensor, torch.Tensor]:
+            loss_vals = -Y  # NOTE(nabenabe): Y is to be maximized, loss_vals is to be minimized.
+            pareto_sols = loss_vals[_is_pareto_front(loss_vals, assume_unique_lexsorted=False)]
+            ref_point = np.max(loss_vals, axis=0) * 1.1
```
nabenabe0928, line 178:
```suggestion
            ref_point = np.max(loss_vals, axis=0)
            ref_point = np.maximum(ref_point * 1.1, ref_point * 0.9)
```

nabenabe0928:
@HideakiImamura @kAIto47802 
Could you review this PR?

optuna/_gp/acqf.py
```diff
@@ -148,7 +246,44 @@ def create_acqf_params(
     )
 
 
+def _eval_ehvi(
```
kAIto47802, line 249: I've reviewed this function, and it looks sound in principle, though I still need to check it precisely.

optuna/_gp/acqf.py
```diff
@@ -22,6 +24,40 @@
...
+    # https://github.com/pytorch/botorch/blob/v0.13.0/botorch/utils/safe_math.py
+    # https://github.com/pytorch/botorch/blob/v0.13.0/botorch/acquisition/multi_objective/logei.py#L146-L266
+    _EPS = torch.tensor(1e-12, dtype=torch.float64)  # NOTE(nabenabe): grad becomes nan when EPS=0.
+    diff = torch.maximum(
+        _EPS,
+        torch.minimum(Y_post[..., torch.newaxis, :], non_dominated_box_upper_bounds)
+        - non_dominated_box_lower_bounds,
+    )
+    log_hvi_vals = torch.special.logsumexp(diff.log().sum(dim=-1), dim=-1)
+    return -log_n_qmc_samples + torch.special.logsumexp(log_hvi_vals, dim=-1)
```
kAIto47802, line 58: This can be simply written as follows:
```suggestion
    return -log_n_qmc_samples + torch.special.logsumexp(diff.log().sum(dim=-1), dim=(-2, -1))
```

optuna/_gp/acqf.py
```diff
@@ -22,6 +24,40 @@
...
+    sobol_engine = torch.quasirandom.SobolEngine(
+        dimension=dim, scramble=True, seed=seed
+    )  # type: ignore
+    # The Sobol sequence in [-1, 1].
+    samples = 2.0 * (sobol_engine.draw(n_samples, dtype=torch.float64) - 0.5)
+    # Inverse transform to standard normal (values to close to 0/1 result in inf values).
+    return torch.erfinv(samples) * float(np.sqrt(2))
+
+
+def logehvi(
```
kAIto47802, line 38: I've reviewed this function and almost LGTM.

kAIto47802 (changes requested):

optuna/_gp/acqf.py
```diff
@@ -22,6 +24,41 @@
     torch = _LazyImport("torch")
 
 
+def _sample_from_normal_sobol(dim: int, n_samples: int, seed: int | None) -> torch.Tensor:
```
kAIto47802, line 27: I've also reviewed this function, and it appears LGTM.

optuna/_gp/acqf.py
```diff
@@ -117,6 +155,67 @@ def from_acqf_params(
         )
 
 
+@dataclass(frozen=True)
+class MultiObjectiveAcquisitionFunctionParams(AcquisitionFunctionParams):
```
kAIto47802, line 159: I've also reviewed this function, and it looks sound in principle.

optuna/_gp/acqf.py
```diff
@@ -117,6 +155,71 @@ def from_acqf_params(
...
+        inverse_squared_lengthscales = np.mean(
+            [
+                acqf_params.kernel_params.inverse_squared_lengthscales.detach().numpy()
+                for acqf_params in acqf_params_for_objectives
+            ],
+            axis=0,
+        )
+        dummy_kernel_params = KernelParamsTensor(
+            # inverse_squared_lengthscales is used in optim_mixed.py.
+            inverse_squared_lengthscales=torch.from_numpy(inverse_squared_lengthscales),
```
kAIto47802, line 201: How about keeping it in `torch.Tensor` without temporarily converting it to `np.ndarray`? It would make the implementation simpler.
However, in this case, no copy is performed in both conversions, i.e., both from `torch.Tensor` to `np.ndarray` and from `np.ndarray` to `torch.Tensor`, so it does not incur any overhead. Thus, the current implementation is also acceptable if you prefer.
```suggestion
        inverse_squared_lengthscales = sum(
            acqf_params.kernel_params.inverse_squared_lengthscales.detach().numpy()
            for acqf_params in acqf_params_for_objectives
        ) / len(acqf_params_for_objectives)
        dummy_kernel_params = KernelParamsTensor(
            # inverse_squared_lengthscales is used in optim_mixed.py.
            inverse_squared_lengthscales=inverse_squared_lengthscales,
```

optuna/_gp/acqf.py
```diff
@@ -117,6 +155,71 @@ def from_acqf_params(
...
+    fixed_samples: torch.Tensor
+
+    @classmethod
+    def from_acqf_params(
+        cls,
+        acqf_params_for_objectives: list[AcquisitionFunctionParams],
+        Y: np.ndarray,
+        n_qmc_samples: int,
+        qmc_seed: int | None,
+    ) -> MultiObjectiveAcquisitionFunctionParams:
```
kAIto47802, line 172: Although the interface of this dataclass and the naming of this classmethod are open to debate, this section needs refactoring for abstraction anyway. Thus, there is little point in discussing it now, and we can leave it as it is.

kAIto47802 (changes requested):
I have finished reviewing `acqf.py`, leaving some comments. PTAL :octocat:

optuna/_gp/acqf.py
```diff
@@ -117,6 +155,71 @@ def from_acqf_params(
...
+        def _get_non_dominated_box_bounds() -> tuple[torch.Tensor, torch.Tensor]:
+            loss_vals = -Y  # NOTE(nabenabe): Y is to be maximized, loss_vals is to be minimized.
+            pareto_sols = loss_vals[_is_pareto_front(loss_vals, assume_unique_lexsorted=False)]
+            ref_point = np.max(loss_vals, axis=0)
+            ref_point = np.maximum(1.1 * ref_point, 0.9 * ref_point)
```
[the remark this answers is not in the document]
nabenabe0928: Let me modify like this because 1.1/0.9 multiplication is simply the convention (to avoid the hypervolume of zero). 
```suggestion
            ref_point = np.nextafter(np.maximum(1.1 * ref_point, 0.9 * ref_point), np.inf)
```

optuna/_gp/acqf.py
HideakiImamura, line 27: Uniform sampling based on Sobol sequence and the inverse cumulative distribution function is a good method with fast convergence, especially effective in low-dimensional problems. However, compared to simply sampling normally distributed random variables using methods like `torch.randn`, is this approach really suitable for the current purpose? Are there any references or experimental results that support the use of this method?

optuna/_gp/acqf.py
```diff
@@ -148,7 +251,44 @@ def create_acqf_params(
     )
 
 
+def _eval_ehvi(
+    evhi_acqf_params: MultiObjectiveAcquisitionFunctionParams, x: torch.Tensor
```
HideakiImamura, line 255: Typo
```suggestion
    ehvi_acqf_params: MultiObjectiveAcquisitionFunctionParams, x: torch.Tensor
```

optuna/_gp/acqf.py
```diff
@@ -117,6 +155,71 @@ def from_acqf_params(
...
+        # Since all the objectives are equally important, we simply use the mean of
+        # inverse_squared_lengthscales over all the objectives.
+        # cf. https://github.com/optuna/optuna/blob/v4.3.0/optuna/_gp/optim_mixed.py#L200-L209
+        inverse_squared_lengthscales = np.mean(
+            [
+                acqf_params.kernel_params.inverse_squared_lengthscales.detach().numpy()
+                for acqf_params in acqf_params_for_objectives
+            ],
+            axis=0,
+        )
```
HideakiImamura, line 198: From what I can see in `_eval_ehvi`, the `kernel_params` in `MultiObjectiveAcquisitionFunctionParams` appears to be a dummy and is not actually used. Therefore, the `inverse_squared_lengthscales` does not need to hold any meaningful values either - would it make sense to simply set it to a dummy tensor, such as `torch.empty(len(params))`?

optuna/_gp/acqf.py
```diff
@@ -117,6 +155,71 @@ def from_acqf_params(
...
+        repr_acqf_params = acqf_params_for_objectives[0]
+        return cls(
+            acqf_type=AcquisitionFunctionType.LOG_EHVI,
+            kernel_params=dummy_kernel_params,
+            X=repr_acqf_params.X,
+            search_space=repr_acqf_params.search_space,
+            cov_Y_Y_inv=np.empty(0),
+            cov_Y_Y_inv_Y=np.empty(0),
+            max_Y=np.nan,  # This is not used anywhere.
+            beta=None,
```
HideakiImamura, line 214: All of them are not used anywhere, aren't they? How about adding the excuse comments for all of them.

Also, for `kernel_params`.

HideakiImamura (commented):

optuna/_gp/acqf.py
nabenabe0928: Good catch, I did not realize it!

optuna/_gp/acqf.py
nabenabe0928: It is a bit tricky because only `inverse_squared_lengthscales` is explicitly accessed externally; please see
- https://github.com/optuna/optuna/blob/v4.3.0/optuna/_gp/optim_mixed.py#L200-L209

So we cannot actually make this tensor a dummy version.

optuna/_gp/acqf.py
HideakiImamura:
```
        # NOTE(nabenabe): By using fixed samples from the Sobol sequence, EHVI becomes
        # deterministic, making it possible to optimize the acqf by l-BFGS.
```

This comment helped me understand the necessity of this function. Since the SobolEngine is re-initialized with the same random seed each time, it makes sense that exactly the same points are sampled every time a new instance of MultiObjectiveAcquisitionFunctionParams is created. On the other hand, I'm curious whether it's better to use this approach or to use torch.randn with a newly initialized generator as shown below. What do you think?

```python
g = torch.Generator()
g.manual_seed(seed)
samples = torch.randn(n_samples, dim, dtype=torch.float64, generator=g)
```

optuna/_gp/acqf.py
nabenabe0928: I will do so!

tests/gp_tests/test_acqf.py
```diff
@@ -4,15 +4,62 @@
...
+        inverse_squared_lengthscales=torch.tensor([2.0, 3.0], dtype=torch.float64),
+        kernel_scale=torch.tensor(4.0, dtype=torch.float64),
+        noise_var=torch.tensor(0.1, dtype=torch.float64),
+    )
+    search_space = SearchSpace(
+        scale_types=np.full(n_dims, ScaleType.LINEAR),
+        bounds=np.array([[0.0, 1.0] * n_dims]),
+        steps=np.zeros(n_dims),
+    )
+    return kernel_params, search_space, X
```
HideakiImamura, line 45: How about defining fixture of them?

optuna/_gp/acqf.py
nabenabe0928: The reference is available at Appendix D of [this paper](https://arxiv.org/pdf/2006.05078), but basically, the authors checked four scenarios:

image

tests/gp_tests/test_acqf.py
nabenabe0928: I updated the unit tests accordingly!

optuna/_gp/acqf.py
HideakiImamura: I understood. Sounds very good to use QMC fixed samples.

optuna/_gp/acqf.py
HideakiImamura: I understood. Thanks for the explanation. BTW, how about using harmonic mean after taking square root? It ensures that the averaged lengthscale is used.

nabenabe0928: Like this?

```python
        inverse_squared_lengthscales_array = np.array(
            [
                acqf_params.kernel_params.inverse_squared_lengthscales.detach().numpy()
                for acqf_params in acqf_params_for_objectives
            ]
        )
        mean_lengthscales = np.mean(1 / np.sqrt(inverse_squared_lengthscales_array))
        dummy_kernel_params = KernelParamsTensor(
            # inverse_squared_lengthscales is used in optim_mixed.py.
            inverse_squared_lengthscales=torch.from_numpy(1.0 / mean_lengthscales**2),
            # These parameters will not be used anywhere.
            kernel_scale=torch.empty(0),
            noise_var=torch.empty(0),
        )
```

nabenabe0928: I applied your suggestion with the co-authorship:)

HideakiImamura (approved):
Thanks for the update! LGTM!

optuna/_gp/acqf.py
nabenabe0928: As `np.mean` is more intuitive compared to `sum(...) / len(...)` and `mean(stack(...))`, I decided to stick to the current implementation, but thank you for the suggestion:)

optuna/_gp/acqf.py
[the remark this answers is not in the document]
kAIto47802: Thank you for your reply. I didn't know about the convention.

kAIto47802 (approved):
Thank you for the update! LGTM

Issue #6065, opened 2025-04-27
Record has changed since last read in table 'study_system_attributes [mariadb]


jvrodrisan:
### Expected behavior

When running an optimization using NSGAII and VSBXCrossover, I expect it to run without database errors. The NSGAII initialization is as follows:

```python
sampler = optuna.samplers.NSGAIISampler(
                population_size=150,
                mutation_prob=0.08, 
                crossover_prob=0.85,
                swapping_prob=0.5,
                seed=_seed,
                crossover=optuna.samplers.nsgaii.VSBXCrossover(eta=15),
                constraints_func=self.constraints_opti if self.constraints else None
)
```
and database configuration below:
```python
storage = optuna.storages.RDBStorage(url=self.db_url, heartbeat_interval=80, grace_period=160,
                   failed_trial_callback=RetryFailedTrialCallback(max_retry=5))
```

### Environment

- Optuna version: 4.2.1
- Python version: 3.13.3
- OS: Linux-6.12.21-amd64-x86_64-with-glibc2.41 (Debian 12)

### Error messages, stack traces, or logs

```shell
ile "/home/jyck/.pyenv/versions/miniforge3-latest/envs/py313/lib/python3.13/site-packages/pymysql/connections.py", line 1199, in read
    first_packet = self.connection._read_packet()
  File "/home/jyck/.pyenv/versions/miniforge3-latest/envs/py313/lib/python3.13/site-packages/pymysql/connections.py", line 775, in _read_packet
    packet.raise_for_error()
    ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/jyck/.pyenv/versions/miniforge3-latest/envs/py313/lib/python3.13/site-packages/pymysql/protocol.py", line 219, in raise_for_error
    err.raise_mysql_exception(self._data)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/home/jyck/.pyenv/versions/miniforge3-latest/envs/py313/lib/python3.13/site-packages/pymysql/err.py", line 150, in raise_mysql_exception
    raise errorclass(errno, errval)
sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (1020, "Record has changed since last read in table 'study_system_attributes'")
[SQL: UPDATE study_system_attributes SET value_json=%(value_json)s WHERE study_system_attributes.study_system_attribute_id = %(study_system_attributes_study_system_attribute_id)s]
[...]

PR #6066, opened 2025-04-29
Fix a bug in distributed optimization using NSGA-II/III

leevers:
The cached parent stores as trial id but retrieves by index, created a quick work around

## Motivation
I am using optuna with the NSGAIISampler and this is causing a crash

## Description of the changes
Dirty hack so that it can continue

### DIFF
--- a/optuna/samplers/_ga/_base.py
+++ b/optuna/samplers/_ga/_base.py
@@ -171,7 +171,10 @@ class BaseGASampler(BaseSampler, abc.ABC):
 
         if cached_parent_population_ids is not None:
             trials = study._get_trials(deepcopy=False)
-            return [trials[trial_id] for trial_id in cached_parent_population_ids]
+            trials_by_id = {}
+            for t in trials:
+                trials_by_id[t._trial_id] = t
+            return [trials_by_id[trial_id] for trial_id in cached_parent_population_ids]
         else:
             parent_population = self.select_parent(study, generation)
             study._storage.set_study_system_attr(
--- a/optuna/samplers/_ga/_base.py
+++ b/optuna/samplers/_ga/_base.py
@@ -171,10 +171,8 @@ class BaseGASampler(BaseSampler, abc.ABC):
 
         if cached_parent_population_ids is not None:
             trials = study._get_trials(deepcopy=False)
-            trials_by_id = {}
-            for t in trials:
-                trials_by_id[t._trial_id] = t
-            return [trials_by_id[trial_id] for trial_id in cached_parent_population_ids]
+            parent_population_ids = set(cached_parent_population_ids)
+            return [trial for trial in trials if trial._trial_id in parent_population_ids]
         else:
             parent_population = self.select_parent(study, generation)
             study._storage.set_study_system_attr(
--- a/tests/samplers_tests/test_base_gasampler.py
+++ b/tests/samplers_tests/test_base_gasampler.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from typing import Any
-from unittest.mock import call
 from unittest.mock import MagicMock
 from unittest.mock import Mock
 from unittest.mock import patch
--- a/tests/samplers_tests/test_base_gasampler.py
+++ b/tests/samplers_tests/test_base_gasampler.py
@@ -219,11 +219,10 @@ def test_get_parent_population(args: dict[str, Any]) -> None:
         class StrictTrialMock(MagicMock):
             def __getattr__(self, name):
                 if name != "_trial_id":
-                    raise AttributeError(
-                        f"Access to attribute '{name}' is not allowed."
-                    )
+                    raise AttributeError(f"Access to attribute '{name}' is not allowed.")
                 return super().__getattr__(name)
-    
+
+
         mock_study._get_trials.return_value = [
             StrictTrialMock(_trial_id=i)
             for i in args["study_system_attrs"][
--- a/tests/samplers_tests/test_base_gasampler.py
+++ b/tests/samplers_tests/test_base_gasampler.py
@@ -219,7 +219,7 @@ def test_get_parent_population(args: dict[str, Any]) -> None:
         mock_study._get_trials.return_value = [
             optuna.trial.FrozenTrial(
                 number=i,
-                trial_id=i+10,
+                trial_id=i + 10,
                 state=optuna.trial.TrialState.WAITING,
                 value=None,
                 datetime_start=None,
--- a/tests/samplers_tests/test_base_gasampler.py
+++ b/tests/samplers_tests/test_base_gasampler.py
@@ -217,7 +217,7 @@ def test_get_parent_population(args: dict[str, Any]) -> None:
     if args["cache"]:
 
         class StrictTrialMock(MagicMock):
-            def __getattr__(self, name: str):
+            def __getattr__(self, name: str) -> Any:
                 if name != "_trial_id":
                     raise AttributeError(f"Access to attribute '{name}' is not allowed.")
                 return super().__getattr__(name)
--- a/tests/samplers_tests/test_base_gasampler.py
+++ b/tests/samplers_tests/test_base_gasampler.py
@@ -215,6 +215,23 @@ def test_get_parent_population(args: dict[str, Any]) -> None:
 
     mock_study = MagicMock()
     mock_study._storage.get_study_system_attrs.return_value = args["study_system_attrs"]
+    if args["cache"]:
+        mock_study._get_trials.return_value = [
+            optuna.trial.FrozenTrial(
+                number=i,
+                trial_id=i,
+                state=optuna.trial.TrialState.WAITING,
+                value=None,
+                datetime_start=None,
+                datetime_complete=None,
+                params={},
+                distributions={},
+                user_attrs={},
+                system_attrs={},
+                intermediate_values={},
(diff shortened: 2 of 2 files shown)


nabenabe0928:
@leevers 
Could you give us the code to reproduce your error?

leevers:
import optuna
from optuna.samplers import NSGAIISampler

storage_path = 'sqlite:///optuna_test_6066.db'

def objective(trial):
    test_param_1 = trial.suggest_categorical(
        "test_param_1",
        [False, True]
    )
    
    value = 100 * (1 if test_param_1 else 0)
       
    return value

# Do a study, this will fill the first 10 initial trial_ids in the database
study_1 = optuna.create_study(
        direction='maximize',
        load_if_exists=True,
        storage=storage_path,
        study_name='optuna_test_6066_study_1')

study_1.optimize(objective, n_trials=10, catch=(RuntimeError, ValueError, AssertionError))

# Run another study, this way the trial_id does not match the index
study_2 = optuna.create_study(
        direction='maximize',
        sampler=NSGAIISampler(
            population_size=4,
            mutation_prob=0.1,
            crossover_prob=0.9,
            swapping_prob=0.5),
        load_if_exists=True,
        storage=storage_path,
        study_name='optuna_test_6066_study_2')

study_2.optimize(objective, n_trials=10, catch=(RuntimeError, ValueError, AssertionError))

leevers:
This could be related to https://github.com/optuna/optuna/issues/6065 as this doesn't occur when RDBStorage is not used

nabenabe0928:
@y0z @gen740 Could you review this PR?

y0z (commented):
A CI error exists.
https://github.com/optuna/optuna/actions/runs/14723994781/job/41322869127?pr=6066
> FAILED tests/samplers_tests/test_base_gasampler.py::test_get_parent_population[args0] - KeyError: 0

Please update the corresponding test as follows.


```python
def test_get_parent_population(args: dict[str, Any]) -> None:
    test_sampler = BaseGASamplerTestSampler(population_size=3)

    mock_study = MagicMock()
    mock_study._storage.get_study_system_attrs.return_value = args["study_system_attrs"]
    if args["cache"]:
        mock_study._get_trials.return_value = [
            optuna.trial.FrozenTrial(
                number=i,
                trial_id=i,
                state=optuna.trial.TrialState.WAITING,
                value=None,
                datetime_start=None,
                datetime_complete=None,
                params={},
                distributions={},
                user_attrs={},
                system_attrs={},
                intermediate_values={},
                values=None,
            ) for i in args["study_system_attrs"][BaseGASamplerTestSampler._get_parent_cache_key_prefix() + "1"]
        ]

    with patch.object(
        BaseGASamplerTestSampler,
        "select_parent",
        return_value=[Mock(_trial_id=i) for i in args["parent_population"]],
    ) as mock_select_parent:
        return_value = test_sampler.get_parent_population(mock_study, args["generation"])

    if args["generation"] == 0:
        assert mock_select_parent.call_count == 0
        assert mock_study._storage.get_study_system_attrs.call_count == 0
        assert mock_study._get_trials.call_count == 0
        assert return_value == []
        return

    mock_study._storage.get_study_system_attrs.assert_called_once_with(mock_study._study_id)

    if not args["cache"]:
        mock_select_parent.assert_called_once_with(mock_study, args["generation"])
        mock_study._storage.set_study_system_attr.assert_called_once_with(
            mock_study._study_id,
            BaseGASamplerTestSampler._get_parent_cache_key_prefix() + str(args["generation"]),
            [i._trial_id for i in mock_select_parent.return_value],
        )
```

leevers:
I have updated the test and got it to a passing state

tests/samplers_tests/test_base_gasampler.py
```diff
@@ -215,6 +214,26 @@ def test_get_parent_population(args: dict[str, Any]) -> None:
...
+                value=None,
+                datetime_start=None,
+                datetime_complete=None,
+                params={},
+                distributions={},
+                user_attrs={},
+                system_attrs={},
+                intermediate_values={},
+                values=None,
+            )
```
gen740, line 232: [NIT]
There’s a test utility function for creating FrozenTrial objects:
https://github.com/optuna/optuna/blob/022f97dbcb9be635ba2987662a33f684ad9811f0/optuna/testing/trials.py#L13-L34

y0z: Could you add a new test case to test this condition?

leevers: i don't know enough about mocking etc to know how

HideakiImamura:

tests/samplers_tests/test_base_gasampler.py
gen740: Since `get_parent_population()` only accesses the `_trial_id` attribute, creating full `FrozenTrial` instances is unnecessary. Simple mocks with just `_trial_id` are sufficient for this test:
```python
if args["cache"]:
    mock_study._get_trials.return_value = [
        Mock(_trial_id=i)
        for i in args["study_system_attrs"][
            BaseGASamplerTestSampler._get_parent_cache_key_prefix() + "1"
        ]
    ]
```
It’s a bit more complicated, but for additional safety, it’s useful to enforce that only `_trial_id` is accessed. This can be done with a strict mock:
```python
if args["cache"]:

    class StrictTrialMock(MagicMock):
        def __getattr__(self, name):
            if name != "_trial_id":
                raise AttributeError(
                    f"Access to attribute '{name}' is not allowed."
                )
            return super().__getattr__(name)

    mock_study._get_trials.return_value = [
        StrictTrialMock(_trial_id=i)
        for i in args["study_system_attrs"][
            BaseGASamplerTestSampler._get_parent_cache_key_prefix() + "1"
        ]
    ]
```
This setup ensures that `get_parent_population()` strictly depends only on `_trial_id`, making the test independent of the trial number or any other attributes.

optuna/samplers/_ga/_base.py
```diff
@@ -171,7 +171,10 @@ def get_parent_population(self, study: optuna.Study, generation: int) -> list[Fr
 
         if cached_parent_population_ids is not None:
             trials = study._get_trials(deepcopy=False)
-            return [trials[trial_id] for trial_id in cached_parent_population_ids]
+            trials_by_id = {}
+            for t in trials:
+                trials_by_id[t._trial_id] = t
+            return [trials_by_id[trial_id] for trial_id in cached_parent_population_ids]
```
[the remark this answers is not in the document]
nabenabe0928: Using a set is further better w.r.t. the time complexity.
```suggestion
            parent_population_ids = set(cached_parent_population_ids)
            return [trial for trial in trials if trial._trial_id in parent_population_ids]
```

tests/samplers_tests/test_base_gasampler.py
```diff
@@ -233,9 +246,8 @@ def test_get_parent_population(args: dict[str, Any]) -> None:
     mock_study._storage.get_study_system_attrs.assert_called_once_with(mock_study._study_id)
 
     if args["cache"]:
-        mock_study._get_trials.assert_has_calls(
-            [call(deepcopy=False)] + [call().__getitem__(i) for i in args["parent_population"]]
-        )
+        assert return_value == mock_study._get_trials.return_value
```
y0z, line 249: Adding a test to confirm that the cache is used instead of calling `select_parent` is helpful.
```suggestion
        assert mock_select_parent.call_count == 0
        assert return_value == mock_study._get_trials.return_value
```

tests/samplers_tests/test_base_gasampler.py
```diff
@@ -215,6 +214,20 @@ def test_get_parent_population(args: dict[str, Any]) -> None:
 
     mock_study = MagicMock()
     mock_study._storage.get_study_system_attrs.return_value = args["study_system_attrs"]
+    if args["cache"]:
+
+        class StrictTrialMock(MagicMock):
+            def __getattr__(self, name: str) -> Any:
+                if name != "_trial_id":
+                    raise AttributeError(f"Access to attribute '{name}' is not allowed.")
+                return super().__getattr__(name)
```
y0z, line 223: @gen740 @leevers 

Perhaps, [Mock does not support overriding `__getattr__`.](https://docs.python.org/3.13/library/unittest.mock.html#mocking-magic-methods)

So, how about simply using MagicMock here?

Actually, I've confirmed a curious behavior, i.e., the name passed to the `__getattr__` is replaced with `_mock_methods `, in the code below, and the current implementation does not work as expected.
```python
        class StrictTrialMock(MagicMock):
            def __getattr__(self, name: str) -> Any:
                print(name)
                if name != "_trial_id":
                    raise AttributeError(f"Access to attribute '{name}' is not allowed.")
                return super().__getattr__(name)
```

```sh
pytest -s tests/samplers_tests/test_base_gasampler.py::test_get_parent_population
=================================================================================== test session starts ====================================================================================
platform darwin -- Python 3.11.7, pytest-7.4.4, pluggy-1.3.0
rootdir: /Users/yozaki/work/review/optuna
configfile: pyproject.toml
plugins: jaxtyping-0.2.30, typeguard-2.13.3
collected 4 items                                                                                                                                                                          

tests/samplers_tests/test_base_gasampler.py _mock_methods
_mock_methods
_mock_methods
```

sawa3030:
Let me work on a follow-up PR to update the tests.

PR #6025, opened 2025-03-28
Remove `restart_strategy` and `inc_popsize` to simplify `CmaEsSampler`

sawa3030:
## Motivation
Simplify the implementation of `CmaEsSampler` by removing support for restart strategies. The current implementation with `restart_strategy` will be put into OptunaHub

## Summary of Changes
- Removed support for `restart_strategy` and `inc_popsize` options in `CmaEsSampler`
- Cleaned up related logic:
  - Removed associated attributes from `_CmaEsAttrKeys`, including `popsize`, `n_restarts`, `n_restarts_with_large`, `poptype`, `small_n_eval`, and `large_n_eval`
  - Eliminated usage of `set_trial_system_attr` since state tracking is no longer needed
- Updated unit tests accordingly

PR #6068, opened 2025-05-02
Remove `CmaEsAttrKeys` and `_attr_keys` for Simplification

sawa3030:
## Motivation  
- The callable functions in `CmaEsAttrKeys` were removed in [#6025](https://github.com/optuna/optuna/pull/6025).
- This cleanup further simplifies the CMA-ES implementation.  
- ref: [#6025 (comment)](https://github.com/optuna/optuna/pull/6025#discussion_r2070100306)

## Description of the Changes  
- Removed the `CmaEsAttrKeys` class.  
- Removed the `_attr_keys` method and replaced its usage with inline logic where necessary.

### DIFF
--- a/optuna/samplers/_cmaes.py
+++ b/optuna/samplers/_cmaes.py
@@ -434,7 +434,7 @@ class CmaEsSampler(BaseSampler):
         else:
             params = optimizer.ask()
 
-        generation_attr_key = self._attr_prefix + "generation"
+        generation_attr_key = self._attr_key_generation
         study._storage.set_trial_system_attr(
             trial._trial_id, generation_attr_key, optimizer.generation
         )
@@ -443,9 +443,17 @@ class CmaEsSampler(BaseSampler):
 
         return external_values
 
+    @property
+    def _attr_key_generation(self) -> str:
+        return self._attr_prefix + "generation"
+
+    @property
+    def _attr_key_optimizer(self) -> str:
+        return self._attr_prefix + "optimizer"
+
     def _concat_optimizer_attrs(self, optimizer_attrs: dict[str, str]) -> str:
         return "".join(
-            optimizer_attrs["{}:{}".format(self._attr_prefix + "optimizer", i)]
+            optimizer_attrs["{}:{}".format(self._attr_key_optimizer, i)]
             for i in range(len(optimizer_attrs))
         )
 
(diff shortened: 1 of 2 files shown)


nabenabe0928:

optuna/samplers/_cmaes.py
```diff
@@ -433,7 +434,7 @@ def sample_relative(
         else:
             params = optimizer.ask()
 
-        generation_attr_key = self._attr_keys.generation
+        generation_attr_key = self._attr_prefix + "generation"
```
c-bata, line 437: Nit: It might be better to introduce the following two property methods to avoid inconsistencies in the key definitions.

```python
@property
def _attr_key_generation(self) -> str:
    return self._attr_prefix + "generation"

@property
def _attr_key_optimizer(self, index: int) -> str:
    return f"{self._attr_prefix}optimizer:{index}"
```

c-bata (approved):
LGTM with a nit.

optuna/samplers/_cmaes.py
```diff
@@ -480,7 +475,7 @@ def _restore_optimizer(
             optimizer_attrs = {
                 key: value
                 for key, value in trial.system_attrs.items()
-                if key.startswith(self._attr_keys.optimizer)
+                if key.startswith(self._attr_key_optimizer)
```
sawa3030, line 478: Due to this implementation, the index is not passed as an argument to `_attr_key_optimizer`.

sawa3030:
Thank you for the review. I made the update accordingly!

optuna/samplers/_cmaes.py
c-bata: I think we don't need to use `self._attr_key_optimizer` here.  Instead, you can simply hardcode `"cma:optimizer"` in this line. Let me explain the background.

In older versions of Optuna, the CMA-ES optimizer was stored as a single system_attr without being split. However, this approach caused issues when using MySQL as the storage backend, so it was revised in the following PR to store the optimizer across multiple string entries:
https://github.com/optuna/optuna/pull/1833

To keep the backward compatibility, PR #1833 introduced these lines. But since more than four years have passed, it's likely that very few users are still affected by this logic, and we could even consider removing it (though it would be better to do so in a separate PR).

HideakiImamura (approved):
Thanks for the follow-up! LGTM!

PR #6069, opened 2025-05-02
Support Multi-Objective Optimization `GPSampler`

nabenabe0928:
## Motivation

This PR adds the multi-objective GPSampler as a followup of these PRs:
- https://github.com/optuna/optuna/pull/6039
- https://github.com/optuna/optuna/pull/6052

## Description of the changes

- Modify GPSampler so that the multi-objective optimization is supported

### DIFF
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -297,9 +297,7 @@ class GPSampler(BaseSampler):
                 n_pareto_sols = len(pareto_params)
                 # TODO(nabenabe): Verify the validity of this choice.
                 size = min(self._n_local_search // 2, n_pareto_sols)
-                chosen_indices = self._rng.rng.choice(
-                    n_pareto_sols, size=size, replace=False
-                )
+                chosen_indices = self._rng.rng.choice(n_pareto_sols, size=size, replace=False)
                 best_params = pareto_params[chosen_indices]
         else:
             assert (
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -119,6 +119,7 @@ class GPSampler(BaseSampler):
 
         # Control parameters of the acquisition function optimization.
         self._n_preliminary_samples: int = 2048
+        # NOTE(nabenabe): ehvi in BoTorchSampler uses 20.
         self._n_local_search = 10
         self._tol = 1e-4
 
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -109,6 +109,7 @@ class GPSampler(BaseSampler):
         self._log_prior: Callable[[gp.KernelParamsTensor], torch.Tensor] = prior.default_log_prior
         self._minimum_noise: float = prior.DEFAULT_MINIMUM_NOISE_VAR
         # We cache the kernel parameters for initial values of fitting the next time.
+        # TODO(nabenabe): Make the cache lists system_attrs to make GPSampler stateless.
         self._kernel_params_cache_list: list[gp.KernelParamsTensor] | None = None
         self._constraints_kernel_params_cache: list[gp.KernelParamsTensor] | None = None
         self._deterministic = deterministic_objective
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from collections.abc import Callable
+from collections.abc import Sequence
 from typing import Any
 from typing import TYPE_CHECKING
 import warnings
@@ -9,6 +11,7 @@ import numpy as np
 import optuna
 from optuna._experimental import experimental_class
 from optuna._experimental import warn_experimental_argument
+from optuna.distributions import BaseDistribution
 from optuna.samplers._base import _CONSTRAINTS_KEY
 from optuna.samplers._base import _process_constraints_after_trial
 from optuna.samplers._base import BaseSampler
@@ -20,9 +23,6 @@ from optuna.trial import TrialState
 
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
-    from collections.abc import Sequence
-
     import torch
 
     import optuna._gp.acqf as acqf
@@ -30,7 +30,6 @@ if TYPE_CHECKING:
     import optuna._gp.optim_mixed as optim_mixed
     import optuna._gp.prior as prior
     import optuna._gp.search_space as gp_search_space
-    from optuna.distributions import BaseDistribution
     from optuna.study import Study
 else:
     from optuna._imports import _LazyImport
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -208,7 +208,7 @@ class GPSampler(BaseSampler):
     def sample_relative(
         self, study: Study, trial: FrozenTrial, search_space: dict[str, BaseDistribution]
     ) -> dict[str, Any]:
-        if len(study.directions) > 1:
+        if study._is_multi_objective():
             if self._constraints_func is not None:
                 raise ValueError(
                     f"{self.__class__.__name__} has not supported multi-objective optimization "
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -302,7 +302,7 @@ class GPSampler(BaseSampler):
                 )
                 best_params = pareto_params[chosen_indices]
         else:
-            assert len(kernel_params_list) == 1, "Multi-objective has not been supported."
+            assert n_objectives == len(kernel_params_list) == 1, "Multi-objective has not been supported."
             constraint_vals, is_feasible = _get_constraint_vals_and_feasibility(study, trials)
             is_all_infeasible = not np.any(is_feasible)
 
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from typing import Any
 from typing import TYPE_CHECKING
-import warnings
 
 import numpy as np
 
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -337,10 +337,6 @@ class GPSampler(BaseSampler):
         param_distribution: BaseDistribution,
     ) -> Any:
         if len(study.directions) > 1:
-            warnings.warn(
-                f"Multi-objective optimization by {self.__class__.__name__} is an experimental "
-                "feature. The interface can change in the future."
-            )
             if self._constraints_func is not None:
                 raise ValueError(
                     f"{self.__class__.__name__} has not supported multi-objective optimization "
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -1,27 +1,28 @@
 from __future__ import annotations
 
-from collections.abc import Callable
-from collections.abc import Sequence
 from typing import Any
-from typing import cast
 from typing import TYPE_CHECKING
+import warnings
 
 import numpy as np
 
 import optuna
 from optuna._experimental import experimental_class
 from optuna._experimental import warn_experimental_argument
-from optuna.distributions import BaseDistribution
 from optuna.samplers._base import _CONSTRAINTS_KEY
 from optuna.samplers._base import _process_constraints_after_trial
 from optuna.samplers._base import BaseSampler
 from optuna.samplers._lazy_random_state import LazyRandomState
 from optuna.study import StudyDirection
+from optuna.study._multi_objective import _is_pareto_front
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
 
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
+    from collections.abc import Sequence
+
     import torch
 
     import optuna._gp.acqf as acqf
@@ -29,6 +30,7 @@ if TYPE_CHECKING:
     import optuna._gp.optim_mixed as optim_mixed
     import optuna._gp.prior as prior
     import optuna._gp.search_space as gp_search_space
+    from optuna.distributions import BaseDistribution
     from optuna.study import Study
 else:
     from optuna._imports import _LazyImport
@@ -42,6 +44,14 @@ else:
 
 
 EPS = 1e-10
+
+
+def _standardize_values(values: np.ndarray) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+    clipped_values = gp.warn_and_convert_inf(values)
+    means = np.mean(clipped_values, axis=0)
+    stds = np.std(clipped_values, axis=0)
+    standardized_values = (clipped_values - means) / np.maximum(EPS, stds)
+    return standardized_values, means, stds
 
 
 @experimental_class("3.6.0")
@@ -79,7 +89,8 @@ class GPSampler(BaseSampler):
             Whether the objective function is deterministic or not.
             If :obj:`True`, the sampler will fix the noise variance of the surrogate model to
             the minimum value (slightly above 0 to ensure numerical stability).
-            Defaults to :obj:`False`.
+            Defaults to :obj:`False`. Currently, all the objectives will be assume to be
+            deterministic if :obj:`True`.
     """
 
     def __init__(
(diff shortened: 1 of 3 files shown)


c-bata:
@kAIto47802 @HideakiImamura Could you review this PR?

nabenabe0928:
Performance Assessment:

hh-wfg-9d-2objectives
hh-wfg-9d-4objectives

Runtime Assessment:
runtime-wfg-9d-4objectives

runtime-wfg-9d-2objectives

optuna/samplers/_gp/sampler.py
```diff
@@ -214,42 +223,77 @@ def sample_relative(
...
+                    qmc_seed=self._rng.rng.randint(1 << 30),
+                )
+                pareto_params = normalized_params[
+                    _is_pareto_front(standardized_score_vals, assume_unique_lexsorted=False)
+                ]
+                n_pareto_sols = len(pareto_params)
+                size = min(self._n_local_search // 2, n_pareto_sols)
+                chosen_indices = self._rng.rng.choice(
+                    np.arange(n_pareto_sols), size=size, replace=False
+                )
```
HideakiImamura, line 293: If the solution on the Pareto-front is large enough, you randomly select `n_local_search // 2` solutions from it as `best_params`. Is there any reference for this selection?

optuna/samplers/_gp/sampler.py
```diff
@@ -214,42 +223,77 @@ def sample_relative(
...
+            if n_objectives == 1:
+                assert len(kernel_params_list) == 1
+                acqf_params = acqf.create_acqf_params(
+                    acqf_type=acqf.AcquisitionFunctionType.LOG_EI,
+                    kernel_params=kernel_params_list[0],
+                    search_space=internal_search_space,
+                    X=normalized_params,
+                    Y=standardized_score_vals[:, 0],
+                )
+                best_params = normalized_params[np.argmax(standardized_score_vals), np.newaxis]
```
HideakiImamura, line 267: The shape of `best_params` is changed. Is `self._optimize_acqf` designed to work with the new shape?

optuna/samplers/_gp/sampler.py
```diff
@@ -214,42 +223,77 @@ def sample_relative(
...
+                    _is_pareto_front(standardized_score_vals, assume_unique_lexsorted=False)
+                ]
+                n_pareto_sols = len(pareto_params)
+                size = min(self._n_local_search // 2, n_pareto_sols)
+                chosen_indices = self._rng.rng.choice(
+                    np.arange(n_pareto_sols), size=size, replace=False
+                )
+                best_params = pareto_params[chosen_indices]
         else:
+            assert len(kernel_params_list) == 1, "Multi-objective has not been supported."
```
HideakiImamura, line 296: The error message is not user-friendly. How about the following?
```suggestion
            assert len(kernel_params_list) == 1, "Multi-objective and constrained optimization is not supported yet in the `GPSampler`."
```

HideakiImamura: In addition, it would be great to add some documents on the `GPSampler` (maybe on its `constraints_func` argument).

optuna/samplers/_gp/sampler.py
```diff
@@ -289,7 +333,17 @@ def sample_independent(
...
+        if len(study.directions) > 1:
+            warnings.warn(
+                f"Multi-objective optimization by {self.__class__.__name__} is an experimental "
+                "feature. The interface can change in the future."
+            )
+            if self._constraints_func is not None:
+                raise ValueError(
+                    f"{self.__class__.__name__} has not supported multi-objective optimization "
+                    "with constraints."
+                )
```
HideakiImamura, line 345: This error should be raised more earlier. How about raising in the `sample_relative`?

optuna/samplers/_gp/sampler.py
```diff
@@ -289,7 +333,17 @@ def sample_independent(
         param_name: str,
         param_distribution: BaseDistribution,
     ) -> Any:
-        self._raise_error_if_multi_objective(study)
+        if len(study.directions) > 1:
+            warnings.warn(
+                f"Multi-objective optimization by {self.__class__.__name__} is an experimental "
+                "feature. The interface can change in the future."
+            )
```
HideakiImamura, line 340: `GPSampler` itself is experimental. So, I don't think we need to add an optional experimental warning for the multi-objective optimization.

tests/samplers_tests/test_samplers.py
HideakiImamura, line 1: It seems that there is no test for `GPSampler` in the multi-objective and constrained optimization setting. Could you add the testcase to validate the error?

HideakiImamura (commented):
Thank you for the PR. I have several comments. Could you take a look?

HideakiImamura:
Thank you for the performance comparison experiments. The runtime speed performance looks very good. `GPSampler` is significantly faster than `BoTorchSampler`. I'm concerned about the hypervolume performance of `GPSampler`. The `GPSampler` is not necessarily superior to the `BoTorchSampler` in terms of hypervolume. In fact, it seems to lose in more cases. How should this be interpreted?

optuna/samplers/_gp/sampler.py
nabenabe0928: No, this is a heuristic inspired from the single-objective optimization by `GPSampler`.
We can also follow the BoTorch convention if necessary.

optuna/samplers/_gp/sampler.py
nabenabe0928: Btw, I noticed that we do not have the documentation for `constraints_func`, so I made another PR to add it.
- https://github.com/optuna/optuna/pull/6081

optuna/samplers/_gp/sampler.py
nabenabe0928: Yes, it is.

image

optuna/samplers/_gp/sampler.py
HideakiImamura: I see. This raises a performance concern. In multi-objective optimization, unlike single-objective optimization, there are multiple Pareto-optimal solutions to reach (and in many cases, there are infinitely many). Therefore, we should ideally start from multiple points for each of these infinitely many Pareto solutions. Of course, that is not feasible, but at the very least, shouldn’t we be optimizing the acquisition function toward Pareto solutions using a larger number of initial points compared to single-objective optimization? I suspect that this might be the reason for the performance difference compared to BoTorch. How does BoTorch handle this?

optuna/samplers/_gp/sampler.py
HideakiImamura: Sounds reasonable.

nabenabe0928:
I document the differences between `BoTorchSampler` and `GPSampler` at the time of this comment:
- Number of local search: 20 ([BoTorchSampler](https://github.com/optuna/optuna-integration/blob/v4.3.0/optuna_integration/botorch/botorch.py#L454)) vs 10 ([GPSampler](https://github.com/optuna/optuna/blob/v4.3.0/optuna/samplers/_gp/sampler.py#L141))
- Number of warm start sampling (QMC): 1024 ([BoTorchSampler](https://github.com/optuna/optuna-integration/blob/v4.3.0/optuna_integration/botorch/botorch.py#L455)) vs 2048 ([GPSampler](https://github.com/optuna/optuna/blob/v4.3.0/optuna/samplers/_gp/sampler.py#L140))
- Warm start sampling selection: Roulette selection with standardization ([BoTorchSampler](https://github.com/pytorch/botorch/blob/v0.14.0/botorch/optim/initializers.py#L942)) vs Roulette selection by the ratios with the maximum acquisition function value ([GPSampler](https://github.com/optuna/optuna/blob/v4.3.0/optuna/_gp/optim_mixed.py#L301-L304))
- Reference point: minimum minus small value ([BoTorchSampler](https://github.com/optuna/optuna-integration/blob/v4.3.0/optuna_integration/botorch/botorch.py#L433)) vs Same as `TPESampler` ([GPSampler](https://github.com/nabenabe0928/optuna/blob/5efa6342bd9cb2b04053e8bbbc37880ca5b3f960/optuna/_gp/acqf.py#L177-L178))
- EVHI calculation: Monte-Carlo expectation of HVI values based on [`log_fatplus`](https://github.com/pytorch/botorch/blob/v0.14.0/botorch/utils/safe_math.py#L288)  ([BoTorchSampler](https://github.com/pytorch/botorch/blob/v0.13.0/botorch/acquisition/multi_objective/logei.py#L146-L266)) vs Monte-Carlo expectation of the exact HVI values ([GPSampler](https://github.com/optuna/optuna/blob/5efa6342bd9cb2b04053e8bbbc37880ca5b3f960/optuna/_gp/acqf.py#L53-L57))

> [!NOTE]
> The default configuration of the logEHVI in BoTorch is available [here](https://github.com/pytorch/botorch/blob/v0.14.0/botorch/acquisition/input_constructors.py#L984).

optuna/samplers/_gp/sampler.py
```diff
@@ -214,42 +229,80 @@ def sample_relative(
...
         ) = gp_search_space.get_search_space_and_normalized_params(trials, search_space)
 
-        _sign = -1.0 if study.direction == StudyDirection.MINIMIZE else 1.0
-        score_vals = np.array([_sign * cast(float, trial.value) for trial in trials])
-
-        score_vals = gp.warn_and_convert_inf(score_vals)
-        standardized_score_vals = (score_vals - np.mean(score_vals)) / max(EPS, np.std(score_vals))
+        _sign = np.array([-1.0 if d == StudyDirection.MINIMIZE else 1.0 for d in study.directions])
+        standardized_score_vals, _, _ = _standardize_values(
+            values=gp.warn_and_convert_inf(_sign * np.array([trial.values for trial in trials]))
```
kAIto47802, line 234: Since `gp.warn_and_convert_inf` is called at the beginning of `_standardize_values`, this call is duplicated.
```suggestion
            _sign * np.array([trial.values for trial in trials])
```

kAIto47802 (changes requested):
I've just started my review, but for now, I'll leave some comments on what I've noticed so far. PTAL :octocat:

kAIto47802 (changes requested):

kAIto47802 (changes requested):

optuna/samplers/_gp/sampler.py
HideakiImamura: Could you check this comment?

HideakiImamura (commented):
Thanks for the update. Basically, LGTM. I have some additional comments. PTAL.

optuna/samplers/_gp/sampler.py
nabenabe0928: Although BoTorch itself does not have a fixed parameter for the number of local searches, `BoTorchSampler` does have it.
- Number of local search: 20 ([BoTorchSampler](https://github.com/optuna/optuna-integration/blob/v4.3.0/optuna_integration/botorch/botorch.py#L454)) vs 10 ([GPSampler](https://github.com/optuna/optuna/blob/v4.3.0/optuna/samplers/_gp/sampler.py#L141))

I would like you to refer to this comment for more details:
- https://github.com/optuna/optuna/pull/6069#issuecomment-2875949475

nabenabe0928:
@HideakiImamura 

I responded to your comment:
- https://github.com/optuna/optuna/pull/6069#discussion_r2092223696

For now, I left a comment here:

- https://github.com/optuna/optuna/pull/6069/files#diff-fe893dd86f2d66f2d847ab31510f07079718c70f104067a431616cd4117214b8R298

We can surely discuss what parameters to use in another PR!

optuna/samplers/_gp/sampler.py
```diff
@@ -99,15 +108,20 @@ def __init__(
...
-        self._optimize_n_samples: int = 2048
         self._deterministic = deterministic_objective
         self._constraints_func = constraints_func
 
         if constraints_func is not None:
             warn_experimental_argument("constraints_func")
 
+        # Control parameters of the acquisition function optimization.
+        self._n_preliminary_samples: int = 2048
+        self._n_local_search = 10
```
nabenabe0928, line 122:
```suggestion
        # NOTE(nabenabe): ehvi in BoTorchSampler uses 20.
        self._n_local_search = 10
```

optuna/samplers/_gp/sampler.py
```diff
@@ -198,7 +208,11 @@ def _get_constraints_acqf_params(
     def sample_relative(
         self, study: Study, trial: FrozenTrial, search_space: dict[str, BaseDistribution]
     ) -> dict[str, Any]:
-        self._raise_error_if_multi_objective(study)
+        if study._is_multi_objective() and self._constraints_func is not None:
+            sampler_name = self.__class__.__name__
+            raise ValueError(
+                f"{sampler_name} has not supported constrained multi-objective optimization "
+            )
```
nabenabe0928, line 215:
```suggestion
            raise ValueError(
                "GPSampler has not supported constrained multi-objective optimization "
            )
```

optuna/samplers/_gp/sampler.py
```diff
@@ -214,42 +228,83 @@ def sample_relative(
...
+                        )
+                    )
+                acqf_params = acqf.MultiObjectiveAcquisitionFunctionParams.from_acqf_params(
+                    acqf_params_for_objectives=acqf_params_for_objectives,
+                    Y=standardized_score_vals,
+                    n_qmc_samples=128,  # NOTE(nabenabe): The BoTorch default value.
+                    qmc_seed=self._rng.rng.randint(1 << 30),
+                )
+                pareto_params = normalized_params[
+                    _is_pareto_front(standardized_score_vals, assume_unique_lexsorted=False)
```
nabenabe0928, line 295:
```suggestion
                    _is_pareto_front(-standardized_score_vals, assume_unique_lexsorted=False)
```

optuna/samplers/_gp/sampler.py
```diff
@@ -198,7 +209,10 @@ def _get_constraints_acqf_params(
     def sample_relative(
         self, study: Study, trial: FrozenTrial, search_space: dict[str, BaseDistribution]
     ) -> dict[str, Any]:
-        self._raise_error_if_multi_objective(study)
+        if study._is_multi_objective() and self._constraints_func is not None:
+            raise ValueError(
+                "GPSampler has not supported constrained multi-objective optimization "
```
kAIto47802, line 214:
```suggestion
                "GPSampler has not yet supported constrained multi-objective optimization."
```

kAIto47802 (changes requested):

HideakiImamura (approved):
Thanks! LGTM!

optuna/samplers/_gp/sampler.py
```diff
@@ -214,42 +228,83 @@ def sample_relative(
...
+                        )
+                    )
+                acqf_params = acqf.MultiObjectiveAcquisitionFunctionParams.from_acqf_params(
+                    acqf_params_for_objectives=acqf_params_for_objectives,
+                    Y=standardized_score_vals,
+                    n_qmc_samples=128,  # NOTE(nabenabe): The BoTorch default value.
+                    qmc_seed=self._rng.rng.randint(1 << 30),
+                )
+                pareto_params = normalized_params[
+                    _is_pareto_front(-standardized_score_vals, assume_unique_lexsorted=False)
```
kAIto47802, line 295: Since the sign has changed in https://github.com/optuna/optuna/pull/6069/commits/81e7e8e4b4ef3f98b676c50183e96bf38ced2eba and the choice has changed from the worst one to the best one, the benchmarking results might become different.

nabenabe0928: Yeap, you are right. I will work on it in another timing!

kAIto47802 (approved):

optuna/samplers/_gp/sampler.py
```diff
@@ -214,42 +228,83 @@ def sample_relative(
...
+                ]
+                n_pareto_sols = len(pareto_params)
+                # TODO(nabenabe): Verify the validity of this choice.
+                size = min(self._n_local_search // 2, n_pareto_sols)
+                chosen_indices = self._rng.rng.choice(
+                    np.arange(n_pareto_sols), size=size, replace=False
```
kAIto47802, line 301: This would be sufficient:
```suggestion
                    n_pareto_sols, size=size, replace=False
```

nabenabe0928:
Other than these,
- https://github.com/optuna/optuna/pull/6069#issuecomment-2875949475

The prior for the kernel hyperparameters is different from BoTorch, whose prior follows [this paper](https://arxiv.org/abs/2402.02229).

PR #6078, opened 2025-05-09
Refactor `_tell_with_warning`

sawa3030:
## Motivation
Improve the performance of `_tell_with_warning` by reducing unnecessary deep copies and simplifying how warning messages are handled.

## Description of the Changes
- Replaced `deepcopy` with a shallow copy in `_tell_with_warning` to reduce overhead when returning a `FrozenTrial`.
- A `deepcopy` is performed only in contexts where the `FrozenTrial` might be modified, such as when it is passed into callbacks or used in `study.tell()`.
- Refactored warning handling by returning the `warning_message` directly from `_tell_with_warning`, rather than storing it in `system_attrs`.
- Added tests to ensure that warning messages are correctly returned and handled in affected scenarios.

### DIFF
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -7,6 +7,7 @@ from concurrent.futures import FIRST_COMPLETED
 from concurrent.futures import Future
 from concurrent.futures import ThreadPoolExecutor
 from concurrent.futures import wait
+import copy
 import datetime
 import gc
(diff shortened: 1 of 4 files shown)


sawa3030:
Let me close this PR for now and split it into two separate ones: one for avoiding `deepcopy`, and another for handling the warning message.

PR #6082, opened 2025-05-14
Refactor warning handling of `_tell_with_warning`

sawa3030:
## Motivation
The `warning_message` does not need to be stored in `frozen_trial.system_attrs`. This PR refactors the implementation to return the warning message directly from `_tell_with_warning`.

**Note:** After this change, the warning message will no longer be accessible via `frozen_trial.system_attrs[STUDY_TELL_WARNING_KEY]`. However, warning messages can still be captured via `logging.getLogger`. In addition, I checked public GitHub repositories, and it looks like `STUDY_TELL_WARNING_KEY` is not used directly by users, so the impact should be minimal.

## Description of the changes
- Refactored the warning handling logic to return the warning message from `_tell_with_warning` instead of storing it in `system_attrs`.
- Updated affected call sites to handle the returned warning message explicitly.
- Added tests to ensure the warning message is correctly returned and can be logged or captured as expected.

### DIFF
--- a/optuna/study/_optimize.py
+++ b/optuna/study/_optimize.py
@@ -23,7 +23,6 @@ from optuna import trial as trial_module
 from optuna.storages._heartbeat import get_heartbeat_thread
 from optuna.storages._heartbeat import is_heartbeat_enabled
 from optuna.study._tell import _tell_with_warning
-from optuna.study._tell import STUDY_TELL_WARNING_KEY
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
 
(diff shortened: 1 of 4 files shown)


HideakiImamura:

sawa3030:
@gen740 Thank you for the review. I've made the update accordingly. PTAL!

gen740 (approved):
LGTM!

not522 (commented):
Could you remove tests for warning messages? They are implementation detail and generally are not tested.

sawa3030:
@not522 Thank you for the advice. I've made the update. PTAL!

not522 (approved):
LGTM!

PR #6087, opened 2025-05-23
Optimize `tell_with_warning` by Avoiding `storage.get_trial` Calls

sawa3030:
## Motivation  
This PR improves the efficiency of the `tell_with_warning` function by eliminating unnecessary calls to `storage.get_trial`

## Description of the Changes  
- Use `trial._cached_frozen_trial` when available to avoid redundant retrieval from storage.

### DIFF
--- a/optuna/study/_tell.py
+++ b/optuna/study/_tell.py
@@ -9,6 +9,7 @@ import warnings
 import optuna
 from optuna import logging
 from optuna import pruners
+from optuna.exceptions import UpdateFinishedTrialError
 from optuna.trial import FrozenTrial
 from optuna.trial import TrialState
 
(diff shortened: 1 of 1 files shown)


sawa3030:
## Considerations for This PR

- `after_trial()` may now receive a `FrozenTrial` that is not in the `RUNNING` state, which differs from the behavior in the master branch.
- `trial._cached_frozen_trial` may be passed to both `_pruners.filter_study()` and `after_trial()`. If the cached trial is stale (e.g., due to manual updates to the storage), this could lead to inconsistencies or unexpected behavior.
- Contributors working with `_tell_with_warning()` should be aware that this implementation relies on `trial._cached_frozen_trial`. A note has been added to clarify this requirement.

sawa3030:
In the optimization process where `ask` and `tell` are executed in different pods, the `tell` function may be interrupted unexpectedly after the `FrozenTrial` in storage has been updated. This can lead to inconsistencies between the cached version and the stored `FrozenTrial`. To avoid introducing potential inconsistencies or bugs, we have decided to close this PR.

Issue #6084, opened 2025-05-20
`JournalStorage` fails frequently in distributed optimization setups in combination with `GrpcProxyStorage`


nabenabe0928:
### Expected behavior

Some processes are killed stochastically when using `JournalStorage` with `GrpcProxyStorage`.
It seems this problem happens due to the design assumption of `JournalStorage`, i.e., only one thread in a process takes care of a trial from the beginning to the end.
However, `GrpcProxyStorage`, in principle, breaks this rule because the process for sampling and that for evaluating a trial are separated.

![Image](https://github.com/user-attachments/assets/f772b9c5-cdd5-4115-8338-b07d7ce5fbaf)

> [!NOTE]
> This issue typically happens when using `enqueue_trial`, but I cannot deny the possibility that we face the same issue with other operations as well.

### Environment

- Optuna version: 4.3
- Python version: 3.11
- OS: Ubuntu 20.04

### Error messages, stack traces, or logs

```shell
[I 2025-05-20 07:15:00,216] Using an existing study with name 'b99b661c-4229-4a54-92bf-f29b0ed7db25' instead of creating a new one.
[I 2025-05-20 07:15:00,222] Using an existing study with name 'b99b661c-4229-4a54-92bf-f29b0ed7db25' instead of creating a new one.
[I 2025-05-20 07:15:00,227] Using an existing study with name 'b99b661c-4229-4a54-92bf-f29b0ed7db25' instead of creating a new one.
[I 2025-05-20 07:15:00,231] Using an existing study with name 'b99b661c-4229-4a54-92bf-f29b0ed7db25' instead of creating a new one.
[I 2025-05-20 07:15:00,242] Using an existing study with name 'b99b661c-4229-4a54-92bf-f29b0ed7db25' instead of creating a new one.
[I 2025-05-20 07:15:00,245] Using an existing study with name 'b99b661c-4229-4a54-92bf-f29b0ed7db25' instead of creating a new one.
[I 2025-05-20 07:15:00,246] Using an existing study with name 'b99b661c-4229-4a54-92bf-f29b0ed7db25' instead of creating a new one.
[I 2025-05-20 07:15:00,254] Using an existing study with name 'b99b661c-4229-4a54-92bf-f29b0ed7db25' instead of creating a new one.
[I 2025-05-20 07:15:00,294] Trial 0 finished with value: 114.0 and parameters: {'y': 3.8116090154495232, 'x': 0.08687552782907204}. Best is trial 0 with value: 114.0.
[...]

PR #6097, opened 2025-05-27
Add a warning about the combination of gRPC Proxy and Journal Storage

toshihikoyanase:
## Motivation

This PR adds a warning to the `GrpcStorageProxy` class documentation to clarify a known limitation regarding unsupported storage types. The corresponding issue is #6084.

## Description of the changes

- Add a warning section about combining the gRPC proxy and journal storage.

## Design choice

1. I'm unsure if we can add a link to the tracking issue (i.e., https://github.com/optuna/optuna/issues/6084).
2. IMHO, we can move this warning as well as the warning about SQLite3 to the reference of [`run_grpc_proxy_server`](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.storages.run_grpc_proxy_server.html#optuna.storages.run_grpc_proxy_server) since users specify the storage there. Regarding consistency, I placed this warning in the `GrpcStorageProxy` reference.

### DIFF
--- a/optuna/storages/_grpc/client.py
+++ b/optuna/storages/_grpc/client.py
@@ -70,6 +70,13 @@ class GrpcStorageProxy(BaseStorage):
 
         Currently, gRPC storage proxy in combination with an SQLite3 database may cause unexpected
         behaviors when calling :func:`optuna.delete_study` due to non-invalidated cache.
+
+    .. warning::
+
+        Currently, gRPC storage proxy does not support the
+        :class:`~optuna.storages.JournalStorage`. This issue is tracked in
+        https://github.com/optuna/optuna/issues/6084. Please use
+        :class:`~optuna.storages.RDBStorage` instead.
     """
 
     def __init__(self, *, host: str = "localhost", port: int = 13000) -> None:


toshihikoyanase:
We can see the rendering results in https://optuna--6097.org.readthedocs.build/en/6097/reference/generated/optuna.storages.GrpcStorageProxy.html#optuna.storages.GrpcStorageProxy.

image

sawa3030:
The updated documentation looks good to me. By the way, I was wondering if it would be better to add a warning message in `run_grpc_proxy_server` when the storage type passed to it is `JournalStorage`.

toshihikoyanase:
Thank you very much for your feedback.
Regarding runtime warnings, since an experimental warning for the gPRC Proxy is already in place, we would like to avoid introducing too many additional runtime warnings. For this PR, we may prefer to focus on modifying the reference for now. Would this approach be acceptable to you?

sawa3030 (approved):
Thank you for the explanation. LGTM

y0z (approved):
LGTM

Issue #6096, opened 2025-05-27
`operator does not exist: trialvaluetype = character varying` in CASE statement on PostgreSQL 17.3 with Optuna 4.3.0


vcovo:
### Expected behavior

**Expected:**
Able to run a study with numerous trials and be able to access the best trial(s) so far without any database errors. PostgreSQL should be able to compare the `trialvaluetype` ENUM with string literals in the `CASE` statement generated by Optuna/SQLAlchemy for finding the best trial.

**Actual:**
The script fails when `study.best_trial` (or any operation that calls `_storage.get_best_trial()`) is accessed. Able to run some trials until this occurs. The root cause is a `psycopg.errors.UndefinedFunction` error from PostgreSQL, indicating it cannot compare the ENUM type `trialvaluetype` (used for `trial_values.value_type`) with a `character varying` type within a `CASE` statement.

### Environment

    *   Optuna version: 4.3.0
    *   Python version: 3.13.x
    *   SQLAlchemy version: ~2.0.35 (e.g., 2.0.35)
    *   Psycopg version: 3.2.4 (psycopg3)
    *   PostgreSQL version: 17.3 (Homebrew)
    *   OS: macOS

### Error messages, stack traces, or logs

```shell
[I 2025-05-27 10:23:14,481] A new study created in RDB with name: study
Study created. ID: 4. Exiting to check schema.
Traceback (most recent call last):
  File "/Users/vcovo/code/venvs/coins/lib/python3.13/site-packages/sqlalchemy/engine/base.py", line 1967, in _exec_single_context
    self.dialect.do_execute(
    ~~~~~~~~~~~~~~~~~~~~~~~^
        cursor, str_statement, effective_parameters, context
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/vcovo/code/venvs/coins/lib/python3.13/site-packages/sqlalchemy/engine/default.py", line 941, in do_execute
    cursor.execute(statement, parameters)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vcovo/code/venvs/coins/lib/python3.13/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
psycopg.errors.UndefinedFunction: operator does not exist: trialvaluetype = character varying
[...]

PR #6099, opened 2025-05-27
Fix: Resolve PG17 incompatibility for ENUMS in CASE statements 

vcovo:
## Motivation
I have a postgresql 17.3 DB that I also wanted to use for the results of the trials ran through Optuna - potentially linking them with my own tables through a common id. Unfortunately this lead to an error: `operator does not exist: trialvaluetype = character varying` error when using PostgreSQL 17.3 as explained in #6096.

## Description of the changes
This changes the SQLAlchemy case() structure in TrialModel.find_max/min_value_trial_id to directly compare the ENUM column with Python ENUM members, avoiding problematic VARCHAR casts on string literals.

### DIFF
--- a/optuna/storages/_rdb/models.py
+++ b/optuna/storages/_rdb/models.py
@@ -198,8 +198,18 @@ class TrialModel(BaseModel):
             .order_by(
                 desc(
                     case(
-                        {"INF_NEG": -1, "FINITE": 0, "INF_POS": 1},
-                        value=TrialValueModel.value_type,
+                        (
+                            TrialValueModel.value_type == TrialValueModel.TrialValueType.INF_NEG,
+                            -1,
+                        ),
+                        (
+                            TrialValueModel.value_type == TrialValueModel.TrialValueType.FINITE,
+                            0,
+                        ),
+                        (
+                            TrialValueModel.value_type == TrialValueModel.TrialValueType.INF_POS,
+                            1,
+                        ),
                     )
                 ),
                 desc(TrialValueModel.value),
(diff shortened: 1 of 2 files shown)


HideakiImamura:
@c-bata @gen740 @sawa3030 Could you review this PR?

vcovo:
@gen740 in case it was missed I added the requested tests. Let me know if anything is still missing for this to be merged.

optuna/storages/_rdb/models.py
```diff
@@ -198,8 +198,18 @@ def find_max_value_trial_id(cls, study_id: int, objective: int, session: orm.Ses
...
+                            -1,
+                        ),
+                        (
+                            TrialValueModel.value_type == TrialValueModel.TrialValueType.FINITE,
+                            0,
+                        ),
+                        (
+                            TrialValueModel.value_type == TrialValueModel.TrialValueType.INF_POS,
+                            1,
+                        ),
```
c-bata, line 212: [question] I didn't tested yet, but based on the error message, can we simply write these lines like this?

```suggestion
                        {"INF_NEG": -1, "FINITE": 0, "INF_POS": 1},
                        value=TrialValueModel.value_type.value,
```

c-bata:
Let me change the reviewer since I'll be on leave until the week after next.

optuna/storages/_rdb/models.py
sawa3030: I tried several alternatives, but they didn’t work as expected:

1. Passing `.value` or `.name` causes errors, likely because `value=` expects a column expression rather than an attribute:

```
case({"INF_NEG": -1, "FINITE": 0, "INF_POS": 1},
     value=TrialValueModel.value_type.value)  # Fails

case({"INF_NEG": -1, "FINITE": 0, "INF_POS": 1},
     value=TrialValueModel.value_type.name)   # Also fails
```

These cases failed in:
[run 15626347340/job 44021240261](https://github.com/sawa3030/optuna/actions/runs/15626347340/job/44021240261)
[run 15624687460/job 44016565482](https://github.com/sawa3030/optuna/actions/runs/15624687460/job/44016565482)


2. Using Enum members as dictionary keys also failed:

```
case(
    {
        TrialValueModel.TrialValueType.INF_NEG: -1,
        TrialValueModel.TrialValueType.FINITE: 0,
        TrialValueModel.TrialValueType.INF_POS: 1,
    },
    value=TrialValueModel.value_type,
)
```
This failed in:
[run 15625311560/job 44018289101](https://github.com/sawa3030/optuna/actions/runs/15625311560/job/44018289101)

Given these issues, I feel that the current comparison approach is more robust.

y0z (approved):
LGTM

BarryHart:
Good day!

FYI:
I am using Optuna 4.4.0, postgresql 17.5, and psycopg3 and I got the same issue as above.

In order to fix it, I used psycopg to connect to optuna's database after creating the storage object using optuna.storages.RDBStorage(), then I executed 
"cur.execute(
sql.SQL("ALTER TABLE trial_values ALTER COLUMN value_type TYPE VARCHAR;")
 )"
on the db before starting the Optuna optimisation.

Hope this helps!

PR #6103, opened 2025-05-30
Fix to prevent daily `checks-optional` CI on the fork repositories

not522:
## Motivation

Daily `checks-optional` CI is not necessary for the fork repositories, so this PR makes it skip.

## Description of the changes

- Fix workflow setting

### DIFF
--- a/.github/workflows/checks-optional.yaml
+++ b/.github/workflows/checks-optional.yaml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   checks:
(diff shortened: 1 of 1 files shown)


toshihikoyanase (approved):
I confirmed that the change is identical to https://github.com/optuna/optuna/blob/29f19f1acb9f8ac55799ba7278a4ec105222f3ff/.github/workflows/checks.yml#L18. SGTM.

PR #6106, opened 2025-05-31
Emit ExperimentalWarning if heartbeat is enabled

lan496:
## Motivation

When we use the heartbeat mechanism of RDBStorage, only `fail_stale_trials` emits ExperimentalWarning currently.
Because the heartbeat mechanism itself is an experimental feature at this moment, it is more helpful to emit an ExperimentalWarning more holistically. 

## Description of the changes

This PR emits ExperimentalWarning at initializing RDBStorage with `heartbeat_interval`. 

This PR does not address the existing ExperimentalWarning from `fail_stale_trials`. Thus, we will see two distinct ExperimentalWarning if we use `fail_stale_trials`.

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -23,6 +23,7 @@ import uuid
 import optuna
 from optuna import distributions
 from optuna import version
+from optuna._experimental import warn_experimental_argument
 from optuna._imports import _LazyImport
 from optuna._typing import JSONSerializable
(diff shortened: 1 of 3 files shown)


contramundum53:
LGTM. 
Maybe we should document in the Storage class that heartbeat is experimental.

contramundum53 (approved):

contramundum53:
Since this PR is small, I'll merge it with one approval.

PR #6112, opened 2025-05-31
Add `_compute_3d` for hypervolume computation

shmurai:
## Motivation
`_compute_hv` has room of improvement on 3D points.

## Description of the changes
I implemented a dedicated function for 3D points (`_compute_3d`)

### DIFF
--- a/optuna/_hypervolume/wfg.py
+++ b/optuna/_hypervolume/wfg.py
@@ -13,6 +13,59 @@ def _compute_2d(sorted_pareto_sols: np.ndarray, reference_point: np.ndarray) -> 
     return edge_length_x @ edge_length_y
 
 
+def _compress_coordinate(coords: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
+    # Returns a permutation of 0, 1, ..., N-1 preserving the original order
+    # as well as the sorted sequence.
+    # Example:
+    #    _compress_coordinate([20.0, 40.0, 30.0, 10.0])
+    # == ([1, 3, 2, 0], [10.0, 20.0, 30.0, 40.0])
+    # NOTE: we do not care about tie-breaks but it is sufficent for the use case in _compute_3d.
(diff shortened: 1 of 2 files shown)


not522:
Thank you for your PR! I confirmed this PR makes the following scripts faster.

```python
import optuna

def objective(trial):
    return x**2 + y**2, (x - 2) ** 2 + (y - 2) ** 2, (x + 2) ** 2 + (y + 2) ** 2

```

```python
import optuna

def objective(trial):
    z = trial.suggest_float("z", 0, 1)
    trial.set_user_attr("constraints", [x ** 2 + y ** 2 + z ** 2 - 1])
    return x, y, z

def constraints_func(trial):
    return trial.user_attrs["constraints"]

sampler = optuna.samplers.NSGAIISampler(seed=42, constraints_func=constraints_func)
study = optuna.create_study(directions=["maximize", "maximize", "maximize"], sampler=sampler)

reference_point=[0, 0, 0]
optuna.visualization.plot_hypervolume_history(study, reference_point)
```

not522:
Could you check the following points?

1. Could you add a test to check `_compute_3d` and `_compute_hv` return the same results?
1. I think this pre-computation is not necessary for `_compute_3d`.
https://github.com/optuna/optuna/blob/d2442957739b370373486328cda0f56225cc2e89/optuna/_hypervolume/wfg.py#L116-L118

shmurai:
Precomputation (`_is_pareto_front`) is also O(N^2) and it will not do much harm even in the worst case.
Also, it can improve space complexity for the case in which the number of points on the pareto front is much smaller than that of all points:
- `_is_pareto_front` requires O(N) space
- `_compute_3d` uses Θ(N^2) space

Thus I think we may keep the precomputation.

not522 (approved):
LGTM!

HideakiImamura (approved):
Looks very good.

PR #6114, opened 2025-05-31
[WIP] Run tests grouped by files in parallel 

lan496:
## Motivation

Currently, CI takes ~15 minutes to execute tests with a single process. It would be nice to make it faster.

## Description of the changes

This PR introduces pytest-xdist to execute tests in multiple processes.
Tests are grouped by files (`--dist=loadfile`) because I have found the test become flaky when I use multiple processes...


lan496:
I'll close this PR since even n=2 processes causes flaky fails :(

PR #6110, opened 2025-05-31
Add `n_recent_trials` to `plot_timeline`

msdsm:
## Description of the changes
Added `n_recent_trials` parameter to `plot_timeline()` to limit the number of trials displayed.
### Changes
- Added `n_recent_trials: int | None = None` parameter
- When specified, only the most recent N trials are plotted
- Raises `ValueError` if `n_recent_trials <= 0`
- Updated docstring and added comprehensive tests

### Usage
```python
# Plot all trials (default)
plot_timeline(study)

# Plot only the 10 most recent trials
plot_timeline(study, n_recent_trials=10)
```

PR #6116, opened 2025-05-31
Improve test cases for `n_recent_trials` of `plot_timeline` (follow-up #6110).

toshihikoyanase:
## Motivation

This is a follow-up PR for #6110. Some test cases for the `plot_timeline` function can be improved, such as
- the `n_recent_trials` argument is not used in `test_get_timeline_plot`,
- the case `n_recent_trials=0` will raise `ValueError`, and it should be handled by `test_plot_timeline_n_recent_trials_invalid`, not  `test_get_timeline_plot`,
- and we'd like to avoid exact matching of the error message since it can be fragile

## Description of the changes

- Update test cases for `optuna.visualization.plot_timeline`


HideakiImamura (approved):
LGTM.

y0z:

PR #6119, opened 2025-05-31
Optimize `Study.best_trial` performance by avoiding unnecessary deep copy

msdsm:
## Description of the changes
- Added private method `_get_best_trial(deepcopy: bool = True)` 
- Modified `best_trial` property to call `_get_best_trial(deepcopy=True)`
- Updated `_log_completed_trial` to use `_get_best_trial(deepcopy=False)`

## Performance
Example code execution time improved from **1.5650s → 1.2588s**

```python
import optuna
def objective(trial):
    y = trial.suggest_int("y", -1, 1)
    return x**2 + y

study = optuna.create_study(sampler=optuna.samplers.RandomSampler())
```


tests/study_tests/test_study.py
```diff
@@ -1725,3 +1725,29 @@ def test_get_metric_names() -> None:
     assert study.metric_names == ["v0"]
     study.set_metric_names(["v1"])
     assert study.metric_names == ["v1"]
+
+
+@pytest.mark.parametrize("storage_mode", STORAGE_MODES)
+def test_get_best_trial_deepcopy_control(storage_mode: str) -> None:
```
toshihikoyanase, line 1731: I'm not fully sure, but we can use a simple function name here.
This is partly because this test file does not have `test_get_best_trial`, and partly because we will not add it since this test case covers full functionality of `_get_best_trial`.

```suggestion
def test_get_best_trial(storage_mode: str) -> None:
```

optuna/study/study.py
```diff
@@ -308,6 +285,42 @@ def _get_trials(
 
         return self._storage.get_all_trials(self._study_id, deepcopy=deepcopy, states=states)
 
+    def _get_best_trial(self, deepcopy: bool = True) -> FrozenTrial:
```
toshihikoyanase, line 288: How about making the argument mandatory? All the lines using this function specify the argument, and the Optuna developers should be aware of the copy.
```suggestion
    def _get_best_trial(self, deepcopy: bool) -> FrozenTrial:
```

toshihikoyanase (changes requested):
Thank you for your PR! I have two comments.

tests/study_tests/test_study.py
```diff
@@ -1725,3 +1725,29 @@ def test_get_metric_names() -> None:
...
+    with StorageSupplier(storage_mode) as storage:
+        study = create_study(storage=storage)
+        study.optimize(lambda t: t.suggest_float("x", -10, 10), n_trials=5)
+
+        with patch("copy.deepcopy", wraps=copy.deepcopy) as mock_deepcopy:
+            # Test _get_best_trial with deepcopy=False.
+            best_trial_no_copy = study._get_best_trial(deepcopy=False)
+            assert mock_deepcopy.call_count == 0
+
+            # Test _get_best_trial with deepcopy=True (default).
```
toshihikoyanase, line 1745: A nit.
```suggestion
            # Test _get_best_trial with deepcopy=True.
```

not522 (approved):
LGTM! I have confirmed the speed improvement.

optuna/study/study.py
```diff
@@ -308,6 +285,42 @@ def _get_trials(
...
         return self._storage.get_all_trials(self._study_id, deepcopy=deepcopy, states=states)
 
+    def _get_best_trial(self, deepcopy: bool) -> FrozenTrial:
+        """Return the best trial in the study.
+
+        Args:
+            deepcopy: Flag to control whether to apply copy.deepcopy() to the trial.
+                     If False, returns the trial without deep copying for better performance.
+                     Note that if you set this to False, you shouldn't mutate any fields
+                     of the returned trial.
```
toshihikoyanase, line 295: We may be able to use inline code blocks. I guess we can work on it in a follow-up PR since this is not a public interface.

optuna/study/study.py
```diff
@@ -308,6 +285,42 @@ def _get_trials(
...
+
+        Returns:
+            A FrozenTrial object of the best trial.
```
toshihikoyanase, line 298: ditto.

toshihikoyanase (approved):
Thanks for your update.
I confirmed a 15.6% speedup in my simple example. LGTM!

We may have room for improvement, but we can address it in a follow-up PR.

PR #6128, opened 2025-06-04
Cache split infromation for TPE sampler.

fusawa-yugo:
## Motivation

This is a revival of https://github.com/optuna/optuna/pull/5464.
In the current implementation,  `_split_cache` is unnecessarily repeated even when the results would be identical.
This PR aims to eliminate such redundant computations and improve efficiency.

## Description of the changes

The content is same as https://github.com/optuna/optuna/pull/5464 for now.
The cache uses `trial.number` as the key, and the value is the result of the split operation.

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -341,6 +341,7 @@ class TPESampler(BaseSampler):
         self._group_decomposed_search_space: _GroupDecomposedSearchSpace | None = None
         self._search_space_group: _SearchSpaceGroup | None = None
         self._search_space = IntersectionSearchSpace(include_pruned=True)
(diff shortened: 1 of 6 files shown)


fusawa-yugo:
I have a question.
Are there any issues with storing the cache in `system_attrs` of `Study`?
I feel that storing cache as an attribute of the Study would be more semantically appropriate.

nabenabe0928:
I noticed a concern, so let me leave a note here:

In principle, this PR should speed up TPESampler in many cases, but it could also have a negative impact.
One example happens in combination with RDBStorage due to the exceed max string length error.
We probably need to consider an option to deactivate this caching feature if the RDBStorage does not allow us to cache a long string.

PR #6130, opened 2025-06-05
Calculate hypervolume in HSSP using sum of contributions

not522:
## Motivation

The hypervolume determined by the selected points can be calculated as the sum of their contributions. This allows skipping the initial hypervolume calculation for each iteration in HSSP.

## Description of the changes

- Change to calculate hypervolume in HSSP using sum of contributions

## Benchmark

- master: 26.196007 sec
- PR: 25.605701 sec

```python
import optuna


def objective(trial: optuna.Trial) -> tuple[float, float, float]:
    return x**2 + y**2, (x - 2)**2 + (y - 2)**2, (x + 2)**2 + (y + 2)**2


trials = study.trials
print((trials[-1].datetime_complete - trials[0].datetime_start).total_seconds())
```

### DIFF
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -45,6 +45,7 @@ def _lazy_contribs_update(
     pareto_loss_values: np.ndarray,
     selected_vecs: np.ndarray,
     reference_point: np.ndarray,
+    hv_selected: float,
 ) -> np.ndarray:
     """Lazy update the hypervolume contributions.
 
@@ -55,9 +56,6 @@ def _lazy_contribs_update(
     H(S' v {j}) - H(S'), we start to update from i with a higher upper bound so that we can
     skip more HV calculations.
     """
-    hv_selected = optuna._hypervolume.compute_hypervolume(
-        selected_vecs[:-1], reference_point, assume_pareto=True
-    )
     max_contrib = 0.0
     index_from_larger_upper_bound_contrib = np.argsort(-contribs)
     for i in index_from_larger_upper_bound_contrib:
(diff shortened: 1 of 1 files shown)


nabenabe0928 (commented):
Could you change here to `np.maximum(0.0, contribs)` so that any `contribs` will not be lower than `max_contrib` and add an inline note about this?

https://github.com/optuna/optuna/blob/5b6348dca33c7d53f282514703047043a65620ce/optuna/_hypervolume/hssp.py#L75

nabenabe0928:
@sawa3030 Could you check this PR?

not522:
> Could you change here to `np.maximum(0.0, contribs)` so that any `contribs` will not be lower than `max_contrib` and add an inline note about this? Such a situation can happen due to numerical instability.
> 
> https://github.com/optuna/optuna/blob/5b6348dca33c7d53f282514703047043a65620ce/optuna/_hypervolume/hssp.py#L75

It's OK, but does that situation cause any errors?

nabenabe0928:
> > Could you change here to `np.maximum(0.0, contribs)` so that any `contribs` will not be lower than `max_contrib` and add an inline note about this? Such a situation can happen due to numerical instability.
> > https://github.com/optuna/optuna/blob/5b6348dca33c7d53f282514703047043a65620ce/optuna/_hypervolume/hssp.py#L75
> 
> It's OK, but does that situation cause any errors?

I am not sure how likely it is, but it may trigger some unwanted behavior.
For example, if `all(contribs < 0)` becomes True at some point, `contribs` will not be updated anymore due to the check `if contribs[i] < max_contrib: continue`.
Yet, there should (or can) still be rankings among `contribs`.
By removing `np.maximum`, we may lose such information, resulting in picking up indices based on the `contribs` right before `all(contribs < 0)` holds.

not522:
In such cases, all contributions become almost zero, so doesn't it matter which one we choose?
By the way, starting `max_contrib` from negative infinity rather than 0 seems like a simple solution to me.

nabenabe0928:
@not522 
Let me leave this comment as a sidenote.
> all contributions become almost zero, so doesn't it matter which one we choose?

It depends on the scale of each objective, but the exact reason why we use the logarithm in the single-objective optimization is to capture such a nuanced difference.
However, I am open to the change:)

The detailed discussion is available in [this paper](https://arxiv.org/abs/2310.20708).

nabenabe0928:
@sawa3030 
Could you please review this PR?
Once you think this PR looks sound, we can merge this PR:D

nabenabe0928 (approved):
Thank you for the PR, LGTM!

PR #6131, opened 2025-06-05
Use hypervolume difference as upperbound of contribs in HSSP

not522:
## Motivation

Since `contribs` in HSSP manages the upper bound of contribution for each point, it can be updated by the difference HV with the last selected point and the next candidate. This reduces the number of points that actually require hypervolume calculation.

## Description of the changes

- Use hypervolume difference as upperbound of contribs in HSSP

## Benchmark

- master: 26.196007 sec
- PR: 21.065669 sec

```python
import optuna


def objective(trial: optuna.Trial) -> tuple[float, float, float]:
    return x**2 + y**2, (x - 2)**2 + (y - 2)**2, (x + 2)**2 + (y + 2)**2


trials = study.trials
print((trials[-1].datetime_complete - trials[0].datetime_start).total_seconds())
```

### DIFF
--- a/optuna/_hypervolume/hssp.py
+++ b/optuna/_hypervolume/hssp.py
@@ -58,6 +58,17 @@ def _lazy_contribs_update(
     hv_selected = optuna._hypervolume.compute_hypervolume(
         selected_vecs[:-1], reference_point, assume_pareto=True
     )
+
+    # Since contribs manages the upper bound of contribution for each point,
+    # it can be updated by the difference HV with the selected point and the next candidate.
+    # Note: H(T v {j}) - H(T) <= H({t} v {j}) - H({t}) = H({j}) - H({t} ^ {j}).
+    # Here, t is the last selected point and j is the candidate. The inequality comes from
+    # submodularity and the equality comes from the inclusion-exclusion principle.
+    single_volume = np.abs(np.prod(pareto_loss_values - reference_point[np.newaxis, :], axis=1))
+    intersection = np.maximum(selected_vecs[-2:-1, :], pareto_loss_values)
+    intersection_volume = np.abs(np.prod(intersection - reference_point[np.newaxis, :], axis=1))
+    contribs = np.minimum(contribs, single_volume - intersection_volume)
+
     max_contrib = 0.0
     index_from_larger_upper_bound_contrib = np.argsort(-contribs)
     for i in index_from_larger_upper_bound_contrib:


optuna/_hypervolume/hssp.py
```diff
@@ -58,6 +58,17 @@ def _lazy_contribs_update(
     hv_selected = optuna._hypervolume.compute_hypervolume(
         selected_vecs[:-1], reference_point, assume_pareto=True
     )
+
+    # Since contribs manages the upper bound of contribution for each point,
+    # it can be updated by the difference HV with the selected point and the next candidate.
+    # Note: H(T v {j}) - H(T) <= H({t} v {j}) - H({t}) = H({j}) - H({t} ^ {j}).
+    # Here, t is the last selected point and j is the candidate. The inequality comes from
+    # submodularity and the equality comes from the inclusion-exclusion principle.
+    single_volume = np.abs(np.prod(pareto_loss_values - reference_point[np.newaxis, :], axis=1))
```
nabenabe0928: We could precompute `single_volume` if necessary.
It is an implicit assumption, but the `compute_hypervolume` function internally checks `np.all(pareto_loss_values <= reference_point[np.newaxis, :])`, so `np.prod(reference_point[np.newaxis, :] - pareto_loss_values, axis=1)` is equivalent.

> [!NOTE]
> `r - a` is quicker than `r[np.newaxis, :] - a`.
```python
In [1]: import numpy as np
   ...: 
   ...: 
   ...: a = np.random.random((30, 3))
   ...: r = np.ones(3)
   ...: %timeit r - a
   ...: %timeit r[np.newaxis, :] - a
   ...: assert np.allclose(r - a, r[np.newaxis, :] - a)
708 ns ± 8.28 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
887 ns ± 10.7 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

```

optuna/_hypervolume/hssp.py
```diff
@@ -58,6 +58,17 @@ def _lazy_contribs_update(
...
+    intersection = np.maximum(selected_vecs[-2:-1, :], pareto_loss_values)
```
nabenabe0928: `a[-2]` is quicker than `a[-2:-1, :]`.

```suggestion
    intersection = np.maximum(selected_vecs[-2, :], pareto_loss_values)
```

Benchmarking:

```python
In [8]: a = np.random.random((30, 3))

In [9]: %timeit b = a[-2:-1, :]
102 ns ± 2.22 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [10]: %timeit b = a[-2, :]
81.2 ns ± 0.731 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [11]: %timeit b = a[-2]
52.7 ns ± 0.392 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

optuna/_hypervolume/hssp.py
```diff
@@ -58,6 +58,17 @@ def _lazy_contribs_update(
...
+    intersection_volume = np.abs(np.prod(intersection - reference_point[np.newaxis, :], axis=1))
```
nabenabe0928: Same here.
```suggestion
    intersection_volume = np.prod(reference_point - intersection, axis=1)
```

Or we can alternatively join the processing above here:
```suggestion
    intersection_volume = np.prod(
        reference_point - np.maximum(selected_vecs[-2], pareto_loss_values), axis=1
    )
```

optuna/_hypervolume/hssp.py
```diff
@@ -58,6 +58,17 @@ def _lazy_contribs_update(
     hv_selected = optuna._hypervolume.compute_hypervolume(
         selected_vecs[:-1], reference_point, assume_pareto=True
     )
+
+    # Since contribs manages the upper bound of contribution for each point,
+    # it can be updated by the difference HV with the selected point and the next candidate.
+    # Note: H(T v {j}) - H(T) <= H({t} v {j}) - H({t}) = H({j}) - H({t} ^ {j}).
+    # Here, t is the last selected point and j is the candidate. The inequality comes from
+    # submodularity and the equality comes from the inclusion-exclusion principle.
```
nabenabe0928:
```suggestion
    # The HV difference only using the latest selected point and a candidate is a simple, yet
    # obvious, contribution upper bound. Denote t as the latest selected index and j as an
    # unselected index. Then, H(T v {j}) - H(T) <= H({t} v {j}) - H({t}) holds where the inequality
    # comes from submodularity. We use the inclusion-exclusion principle to calculate the RHS.
```

Btw, as `t` and `j` are indices, the notation `H({t} ^ {j})` is not well-defined.
If we would like to use `^`, i.e., the maximum operator here, we need to rewrite the docstring above.

nabenabe0928 (commented):
Thank you for the PR, I left some comments:)

nabenabe0928:
I confirmed the reproducibility with the following!

```python
    contribs = np.minimum(
        contribs,
        (
            np.prod(reference_point - pareto_loss_values, axis=1)
            - np.prod(reference_point - np.maximum(selected_vecs[-2], pareto_loss_values), axis=1)
        ),
    )
```

optuna/_hypervolume/hssp.py
```diff
@@ -58,6 +58,16 @@ def _lazy_contribs_update(
     hv_selected = optuna._hypervolume.compute_hypervolume(
         selected_vecs[:-1], reference_point, assume_pareto=True
     )
+
+    # The HV difference only using the latest selected point and a candidate is a simple, yet
+    # obvious, contribution upper bound. Denote t as the latest selected index and j as an
+    # unselected index. Then, H(T v {j}) - H(T) <= H({t} v {j}) - H({t}) holds where the inequality
+    # comes from submodularity. We use the inclusion-exclusion principle to calculate the RHS.
+    single_volume = np.prod(reference_point - pareto_loss_values, axis=1)
+    intersection = np.maximum(selected_vecs[-2, :], pareto_loss_values)
```
nabenabe0928:
```suggestion
    intersection = np.maximum(selected_vecs[-2], pareto_loss_values)
```

PR #6133, opened 2025-06-06
Refactor `tell_with_warning` to avoid unnecessary `get_trial` call

sawa3030:
## Motivation  
Remove unnecessary calls to `storage.get_trial` to improve efficiency.

## Description of the changes  
- Removed the final `storage.get_trial` call in `tell_with_warning`.  
- Instead of returning the `FrozenTrial`, `tell_with_warning` now returns `state`, `values`, `params`, and `values_conversion_failure_message`, which are required for subsequent log output.

### DIFF
--- a/optuna/study/_tell.py
+++ b/optuna/study/_tell.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from collections.abc import Sequence
 import math
-from typing import Any
 from typing import TYPE_CHECKING
 import warnings
 
@@ -87,7 +86,7 @@ def _tell_with_warning(
     state: TrialState | None = None,
     skip_if_finished: bool = False,
     suppress_warning: bool = False,
-) -> tuple[TrialState, list[float] | None, dict[str, Any], str | None]:
+) -> tuple[TrialState, list[float] | None, str | None]:
     """Internal method of :func:`~optuna.study.Study.tell`.
 
     Refer to the document for :func:`~optuna.study.Study.tell` for the reference.
@@ -111,7 +110,7 @@ def _tell_with_warning(
             f"{value_or_values} and state {state} since trial was already finished. "
             f"Finished trial has values {frozen_trial.values} and state {frozen_trial.state}."
         )
-        return frozen_trial.state, frozen_trial.values, frozen_trial.params, None
+        return frozen_trial.state, frozen_trial.values, None
     elif frozen_trial.state != TrialState.RUNNING:
         raise ValueError(f"Cannot tell a {frozen_trial.state.name} trial.")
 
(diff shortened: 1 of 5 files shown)


sawa3030:
## Performance Improvement
The runtime of `_run_trial` was compared between this PR and the `main` branch using the following code. The results shows a noticeable improvement in the cumulative time (`cumtime`) of `_run_trial`, as expected.

| Branch    | Average Cumtime |
|-----------|------------------|
| master   | 40.8345 ± 0.8265     | 
| this PR    | 37.3786 ± 0.5339      |

```
import optuna
import optunahub
import cProfile

    url="sqlite:///:memory:",
    engine_kwargs={"pool_size": 20, "connect_args": {"timeout": 10}},
)

    y = trial.suggest_int("y", -1, 1)
    return x**2 + y

study = optuna.create_study(sampler=optuna.samplers.RandomSampler(), storage=storage)

profiller = cProfile.Profile()
study.optimize(objective, n_trials=500)
```

not522:
> The runtime of _run_trial was compared between this PR and the main branch using the following code. The results shows a noticeable improvement in the cumulative time (cumtime) of _run_trial, as expected.

Are the results written on the table reversed?

optuna/study/_tell.py
```diff
@@ -174,6 +175,4 @@ def _tell_with_warning(
     finally:
         study._storage.set_trial_state_values(frozen_trial._trial_id, state, values)
 
-    frozen_trial = study._storage.get_trial(frozen_trial._trial_id)
-
-    return frozen_trial, values_conversion_failure_message
+    return state, values, frozen_trial.params, values_conversion_failure_message
```
not522: Do we need to return `params`? `_run_trial` stores them in the trial.

c-bata:

sawa3030:
> > The runtime of _run_trial was compared between this PR and the main branch using the following code. The results shows a noticeable improvement in the cumulative time (cumtime) of _run_trial, as expected.
> 
> Are the results written on the table reversed?

Thank you for pointing that out — the results were indeed reversed. I've corrected the comment accordingly.

optuna/study/study.py
```diff
@@ -668,14 +669,14 @@ def df(x):
...
-        frozen_trial, _ = _tell_with_warning(
+        _tell_with_warning(
             study=self,
             trial=trial,
             value_or_values=values,
             state=state,
             skip_if_finished=skip_if_finished,
         )
-        return copy.deepcopy(frozen_trial)
+        return copy.deepcopy(_get_frozen_trial(self, trial))
```
not522, line 679: How about use `self._storage.get_trial`?

sawa3030: I've made the updates. PTAL!

not522: Since the argument is the trial number, we should use `get_trial_id_from_study_id_trial_number`. After reconsideration, this solution turns out to be more complex than I expected, so reverting to using `_get_frozen_trial` may be good. :bow:

not522 (approved):
LGTM!

kAIto47802 (approved):
Thank you for the PR. LGTM!

Issue #1080, opened 2020-04-02
Is it possible to specify a starting point for the values suggested?


vmgustavo:
Instead of starting from a random point I would like to define a specific set of parameters to start the search. Lets say I have prior knowledge about the problem and I know a set of parameters that would perform well and it should try to improve from there, is it possible to specify this initial set? Or is it always random?

hvy:
Hi @vmgustavo. You can "queue" parameter configurations before starting the optimizations using [`Study.enqueue_trial`](https://optuna.readthedocs.io/en/latest/reference/study.html?#optuna.study.Study.enqueue_trial). Would that work for you? Let me leave a reference to #417 as it sounds like a duplicate.

vmgustavo:
It seems to be enough for me. Thanks, I will try to use it.

hvy:
Let me close this ticket for now as resolved. Feel free to reopen as needed.

turbotimon:
@hvy I had the same question and took my a while to find this answer. I would love to see this in the FAQ, would you agree or do you think it's too specific?

hvy:
Makes sense to me. Thanks for the suggestion.
I just created a PR to start a discussion at https://github.com/optuna/optuna/pull/6157 . Would this have addressed your issue?

turbotimon:
Perfect, thanks!

PR #6157, opened 2025-06-15
Add an FAQ entry about specifying optimization parameters

hvy:
## Motivation

Addresses https://github.com/optuna/optuna/issues/1080#issuecomment-2967031231 given that the issue has several reactions and users seemingly end up here instead of in the tutorial https://optuna.readthedocs.io/en/stable/tutorial/20_recipes/008_specify_params.html . 

## Description of the changes

Adds an entry to the FAQ answering _if_ it's possible to specify parameter starting points before the optimization.
_How_ this is done is addressed in the tutorial, so the entry is merely a pointer to the tutorial. Improving the searchability of the tutorial may be a less straight forward but alternative solution.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -775,3 +775,10 @@ However, if it is necessary to remove artifacts from a Python script, users can 
         for artifact_meta in get_all_artifact_meta(study):
             # Remove the artifacts uploaded to ``base_path``.
             artifact_store.remove(artifact_meta.artifact_id)
+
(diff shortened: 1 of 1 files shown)


nabenabe0928 (approved):
Thank you for the PR, LGTM!!

nabenabe0928:

turbotimon:
Looks good to me, thanks! (I was looking in the tutorial section but somehow missed that too)

y0z (approved):
LGTM

PR #6170, opened 2025-06-20
Add pytest-xdist to speedup the ci test

gen740:
## Motivation

Current ci test takes around 15 minutes. To improve the speed of ci tests, I added the `pytest-xdist` to parallelize.

## Description of the changes

- Add `pytest-xdist` to `pyproject.toml`
- Add `-n 8` to every pytest execution.
- Add lock file for getting free port to avoid race-condition of GRPC server.

## Additional context
To determine best worker number for the ci test, I ran a lot of test on [my repository](https://github.com/gen740/optuna/actions/runs/15773540575), and I found it seems the 8 workers are enough.

Following figures are the mean and stddev of 4 ci-tests varying the worker number from 1 to 10. The measurement was done by my repository. The elapsed time is just pytest execution. So this graph **DOES NOT** contain the pip install or other setup time.
Figure_1

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -97,6 +97,6 @@ jobs:
           target="not slow"
         fi
 
-        pytest tests -m "$target"
+        pytest tests -m "$target" -n 8
       env:
         SQLALCHEMY_WARN_20: 1
--- a/optuna/testing/storages.py
+++ b/optuna/testing/storages.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import fcntl
 import socket
 import threading
 from types import TracebackType
@@ -41,6 +42,8 @@ STORAGE_MODES_HEARTBEAT = [
 ]
 
 SQLITE3_TIMEOUT = 300
+
+FIND_FREE_PORT_LOCK_FILE = "/tmp/optuna_find_free_port.lock"
 
 
 class StorageSupplier:
@@ -109,16 +112,24 @@ class StorageSupplier:
             assert False
 
     def _create_proxy(self, storage: BaseStorage) -> GrpcStorageProxy:
-        port = _find_free_port()
-        self.server = optuna.storages._grpc.server.make_server(storage, "localhost", port)
-        self.thread = threading.Thread(target=self.server.start)
-        self.thread.start()
-        self.proxy = GrpcStorageProxy(host="localhost", port=port)
-        self.proxy.wait_server_ready(timeout=60)
-        return self.proxy
+        with open(FIND_FREE_PORT_LOCK_FILE, "w") as lockfile:
+            fcntl.flock(lockfile, fcntl.LOCK_EX)
+            try:
+                port = _find_free_port()
+                self.server = optuna.storages._grpc.server.make_server(storage, "localhost", port)
+                self.thread = threading.Thread(target=self.server.start)
+                self.thread.start()
+                self.proxy = GrpcStorageProxy(host="localhost", port=port)
+                self.proxy.wait_server_ready(timeout=60)
+                return self.proxy
+            finally:
+                fcntl.flock(lockfile, fcntl.LOCK_UN)
 
     def __exit__(
-        self, exc_type: type[BaseException], exc_val: BaseException, exc_tb: TracebackType
+        self,
+        exc_type: type[BaseException],
+        exc_val: BaseException,
+        exc_tb: TracebackType,
     ) -> None:
         if self.tempfile:
             self.tempfile.close()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,6 +99,7 @@ test = [
   "kaleido<0.4",  # TODO(nzw0301): Remove the version constraint by installing browser separately.
   "moto",
   "pytest",
+  "pytest-xdist",
   "scipy>=1.9.2",
   "torch; python_version<='3.12'",  # TODO(gen740): Remove this line when 'torch', a dependency of 'optuna/_gp', supports Python 3.13
   "grpcio",  # optuna/storages/_grpc.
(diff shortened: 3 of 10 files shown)


nabenabe0928:
@kAIto47802 
Could you review this PR?

nabenabe0928:
How about using `-n 4` instead of `-n 8` because the default CPU count in GitHub Actions is 4?

- https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job?utm_source=chatgpt.com#standard-github-hosted-runners-for-public-repositories

pyproject.toml
```diff
@@ -171,6 +172,7 @@ filterwarnings = 'ignore::optuna.exceptions.ExperimentalWarning'
 markers = [
   "skip_coverage: marks tests are skipped when calculating the coverage",
   "slow: marks tests as slow (deselect with '-m \"not slow\"')",
+  "serial: marks tests as fragile when executed in parallel.",
```
nabenabe0928: Do we still need this line?

optuna/testing/storages.py
```diff
@@ -109,16 +154,20 @@ def __enter__(
...
+            self.proxy = GrpcStorageProxy(host="localhost", port=port)
+            self.proxy.wait_server_ready(timeout=60)
+            return self.proxy
 
     def __exit__(
-        self, exc_type: type[BaseException], exc_val: BaseException, exc_tb: TracebackType
+        self,
+        exc_type: type[BaseException],
+        exc_val: BaseException,
+        exc_tb: TracebackType,
```
nabenabe0928: Reverted here.
```suggestion
        self, exc_type: type[BaseException], exc_val: BaseException, exc_tb: TracebackType,
```

tests/study_tests/test_study.py
```diff
@@ -1660,6 +1660,8 @@ def test_tell_from_another_process() -> None:
 def test_pop_waiting_trial_thread_safe(storage_mode: str) -> None:
     if storage_mode in ("sqlite", "cached_sqlite", "grpc_rdb"):
         pytest.skip("study._pop_waiting_trial is not thread-safe on SQLite3")
+    if storage_mode == "grpc_journal_file":
+        pytest.skip("gRPC journal file storage does not support multi-threading")
```
nabenabe0928, line 1664: I am not confident if it is fine to skip this test.

optuna/testing/storages.py
```diff
@@ -43,6 +45,49 @@
...
+        if sys.platform == "win32":
+            import msvcrt
+
+            msvcrt.locking(self._lockfile.fileno(), msvcrt.LK_UNLCK, 1)
+        else:
+            import fcntl
+
+            fcntl.flock(self._lockfile, fcntl.LOCK_UN)
+
+        self._lockfile.close()
```
nabenabe0928: The context manager can be simpler with `contextmanager`.

```suggestion
from contextlib import contextmanager

@contextmanager
def _lock_to_search_for_free_port():
    on_windows = sys.platform == "win32"
    if on_windows:
        lock_path: str = os.path.join(
            os.environ.get("PROGRAMDATA", "C:\\ProgramData"),
            "optuna",
            "optuna_find_free_port.lock",
        )
    else:
        lock_path: str = f"/tmp/optuna_find_free_port.lock"

    os.makedirs(os.path.dirname(lock_path), exist_ok=True)
    lockfile = open(lock_path, "w")
    if on_windows:
        import msvcrt

        msvcrt.locking(lockfile.fileno(), msvcrt.LK_LOCK, 1)
        yield
        msvcrt.locking(lockfile.fileno(), msvcrt.LK_UNLCK, 1)
    else:
        import fcntl

        fcntl.flock(lockfile, fcntl.LOCK_EX)
        yield
        fcntl.flock(lockfile, fcntl.LOCK_UN)

    lockfile.close()
```

optuna/testing/storages.py
```diff
@@ -43,6 +45,49 @@
...
+
+
 class StorageSupplier:
```
kAIto47802: This is not a problem with your PR itself, but subclassing from `AbstractContextManager` might be better[^1]:
```suggestion
class StorageSupplier(AbstractContextManager):
```

Also, as @nabenabe0928 mentions, this class can be rewritten by using the `contextmanager` decorator. However, for this class, the current approach might be better, since the `__enter__` method is much more complicated than `FindFreePortLockFile`.

[^1]: https://docs.python.org/3/library/contextlib.html

kAIto47802:
> How about using `-n 4` instead of `-n 8` because the default CPU count in GitHub Actions is 4?

Regarding the `-n` option, how about using `-n auto`, which uses as many processes as the number of physical CPU cores[^1].

[^1]: https://pytest-xdist.readthedocs.io/en/stable/distribution.html

.github/workflows/tests.yml
```diff
@@ -97,6 +97,6 @@ jobs:
           target="not slow"
         fi
 
-        pytest tests -m "$target"
+        pytest tests -m "$target" -n 8
```
nabenabe0928, line 100: Probably, you wanted to do something like this?

```shell
pytest tests -m "not serial" -n auto
pytest tests -m "serial"
```

tests/study_tests/test_study.py
gen740: As I referred to in #6172, this test with `grpc_journal_file` is potentially fragile, so I think it's better to skip it.
This test with thread sizes more than 20 tends to fail on `grpc_journal_file`, and with xdist exceeding this thread size.

pyproject.toml
gen740: I forgot to delete this line.

gen740:
> Regarding the -n option, how about using -n auto, which uses as many processes as the number of physical CPU cores[1](https://github.com/optuna/optuna/pull/6170#user-content-fn-1-0fea44e5b468792a2f0e1c480f2c3810).

I think the reason the CI test is so slow isn’t the computational load but the tests that include sleep (such as heartbeat tests). In this case, increasing parallelism beyond the number of CPU cores can be effective.

nabenabe0928 (approved):
LGTM!

Issue #6172, opened 2025-06-20
Suspicious `JournalFileStorage` behavior with `GrpcProxyStorage` in a multi-threading setup


gen740:
### Expected behavior

The enqueued trial number and trial id number should be the same.

### Environment

- Optuna version: 4.4.0.dev
- Python version: 3.12.10
- OS: Linux-6.14.10-orbstack-00291-g1b252bd3edea-aarch64-with-glibc2.40
- (Optional) Other libraries and their versions:

### Error messages, stack traces, or logs

```shell
num_enqueued = 50,  len(trial_id_set) = 47
Traceback (most recent call last):
  File "/Users/gen/Projects/Optuna/optuna/workdir/test.py", line 27, in <module>
    assert len(trial_id_set) == num_enqueued
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

### Steps to reproduce

Following is the modified version of the [study test]( https://github.com/optuna/optuna/blob/f80e7202f5c2e2a703238bacfdcac6f0d9524012/tests/study_tests/test_study.py#L1671-L1691).

```python
from optuna import create_study
from optuna.storages import GrpcStorageProxy
from concurrent.futures import ThreadPoolExecutor
from concurrent.futures import as_completed

num_enqueued = 50
study = create_study(
    storage=GrpcStorageProxy(
        host="localhost",
        port=13000,
    )
)
for i in range(num_enqueued):
    study.enqueue_trial({"i": i})

trial_id_set = set()

with ThreadPoolExecutor(100) as pool:
    futures = []
    for i in range(num_enqueued):
        future = pool.submit(study._pop_waiting_trial_id)
        futures.append(future)

    for future in as_completed(futures):
        trial_id_set.add(future.result())
print(f"{num_enqueued = },  {len(trial_id_set) = }")
assert len(trial_id_set) == num_enqueued
```

This test seems fine on the CI, but increasing the ThreadPool size, it fails.
This happened only on Linux. I cannot reproduce this on Mac.

### Additional context (optional)

This issue emerged when I tried to parallelize the CI test in #6170.
For now, I have skipped this test on grpc_journal_file in #6170.

I think this issue would be related to #6084.

nabenabe0928:
@gen740 
Thank you for the report.
I think this issue relates to:
- https://github.com/nabenabe0928/optuna/pull/23

Could you try if the change fixes your issue?

PR #6175, opened 2025-06-23
Fix ill-combination of journal and gRPC

nabenabe0928:
## Motivation

This PR resolves the following issues:
- https://github.com/optuna/optuna/issues/6084
- https://github.com/optuna/optuna/issues/6172

## Description of the changes

- Fix the logic for update check in `set_trial_state_values`
- Modify an existing unit test so that the current master branch fails much more often
- Add a unit test to verify whether JournalStorage works with gRPC on multi-processing setups


nabenabe0928:
@c-bata @gen740 
Could you review this PR?

tests/study_tests/test_study.py
```diff
@@ -1673,8 +1673,13 @@ def test_pop_waiting_trial_thread_safe(storage_mode: str) -> None:
...
 
-    num_enqueued = 10
-    with StorageSupplier(storage_mode) as storage:
+    num_enqueued = 30
+    # NOTE(nabenabe): Fewer threads in gRPC increases the probability of thread collision on the
+    # proxy side. See https://github.com/optuna/optuna/issues/6084
+    storage_kwargs = (
+        {"thread_pool": ThreadPoolExecutor(2)} if storage_mode == "grpc_journal_file" else {}
+    )
+    with StorageSupplier(storage_mode, **storage_kwargs) as storage:
```
nabenabe0928: The master branch with this change yielded 80 failures out of 100 runs on Ubuntu 20.04.

```shell
$ (for _ in `seq 0 99`; do python -m pytest tests/study_tests/test_study.py::test_pop_waiting_trial_thread_safe[grpc_journal_file] | grep "1 f
ailed"; done) | wc -l

>>> 80
```

> [!NOTE]
> Meanwhile, I got no failures with the changes in this PR.

tests/study_tests/test_study.py
```diff
@@ -1673,8 +1673,16 @@ def test_pop_waiting_trial_thread_safe(storage_mode: str) -> None:
     if storage_mode in ("sqlite", "cached_sqlite", "grpc_rdb"):
         pytest.skip("study._pop_waiting_trial is not thread-safe on SQLite3")
 
-    num_enqueued = 10
-    with StorageSupplier(storage_mode) as storage:
+    num_enqueued = 30
```
nabenabe0928: https://github.com/optuna/optuna/pull/6175#discussion_r2176749742

nabenabe0928:
@c-bata @gen740 
I added a unit test and modified an existing one accordingly:)

tests/storages_tests/journal_tests/test_combination_with_grpc.py
nabenabe0928, line 1: This unit test also induces quite frequent failures in the master branch.

tests/study_tests/test_study.py
```diff
@@ -1673,8 +1673,16 @@ def test_pop_waiting_trial_thread_safe(storage_mode: str) -> None:
...
+    # NOTE(nabenabe): Fewer threads in gRPC increases the probability of thread collision on the
+    # proxy side. See https://github.com/optuna/optuna/issues/6084
+    # However, only one thread guarantees no failure because each get_all_trials call in
+    # _pop_waiting_trial_id happens sequentially. This is why, theoretically speaking, the failure
+    # is likely to happen with two threads the most.
+    storage_kwargs = (
+        {"thread_pool": ThreadPoolExecutor(2)} if storage_mode == "grpc_journal_file" else {}
```
c-bata: How about using the same thread pool size as in L1691, which is 10 threads? Otherwise, this test case is only executed in parallel with 2.

nabenabe0928:
I tested the new unit tests on the master branch:

```shell
$ (for _ in `seq 0 99`; do python -m pytest tests/storages_tests/journal_tests/test_combination_with_grpc.py::test_pop_waiting_trial_multiprocess_safe | grep "1 failed"; done) | wc -l
>>> 45

$ (for _ in `seq 0 99`; do python -m pytest tests/storages_tests/journal_tests/test_combination_with_grpc.py::test_pop_waiting_trial_thread_safe | grep "1 failed"; done) | wc -l
>>> 8
```

nabenabe0928:
This PR addresses the concern here:
- https://github.com/optuna/optuna/pull/6170/files#r2176367263

nabenabe0928:
@gen740 @c-bata 

It seems the thread problem in the unit test also went away:
- https://github.com/optuna/optuna/pull/6170

optuna/storages/journal/_storage.py
```diff
@@ -320,13 +321,25 @@ def set_trial_state_values(
...
+                # return statement of trial_id == _replay_result.owned_trial_id. To eliminate false
+                # positives, we verify whether another process is already evaluating the trial with
+                # trial_id. If True, it means this query does not update the trial state.
+                existing_trial = self._replay_result._trials.get(trial_id)
+                if existing_trial is not None and existing_trial.state != TrialState.WAITING:
+                    if existing_trial.state.is_finished():
+                        raise UpdateFinishedTrialError(
+                            UNUPDATABLE_MSG.format(trial_number=existing_trial.number)
+                        )
+                    return False
```
c-bata: I have two questions.

1. Can we use an `assert` statement to check that `existing_trial is not None`, so we simplify the condition in the if statement?
2. Since `existing_trial.state.is_finished()` implies that `existing_trial.state != TrialState.WAITING` is always true, can we reduce the nesting like this?

```python
                existing_trial = self._replay_result._trials.get(trial_id)
                assert existing_trial is not None, (
                    "This must be True. Please file a bug report on GitHub if this line raises AssertionError."
                )
                if existing_trial.state.is_finished():
                    raise UpdateFinishedTrialError(
                        UNUPDATABLE_MSG.format(trial_number=existing_trial.number)
                    )
                if existing_trial.state != TrialState.WAITING:  # this line is equivalent to `existing_trial.state == TrialState.RUNNING`.
                    return False
```

tests/study_tests/test_study.py
nabenabe0928: It's been resolved.

optuna/storages/journal/_storage.py
nabenabe0928: I confirmed that this PR works nicely on a toy problem even with your change!
Let me confirm with a bigger task:)
I am gonna get back to you as soon as the experiment finishes!

c-bata:
I also checked `JournalRedisBackend` works as expected with this code, so just leaving it here for reference.

```sh
docker run -d --name redis -p 127.0.0.1:6379:6379 redis
```

```python

import optuna
from optuna.storages import JournalStorage
from optuna.storages.journal import JournalFileBackend
from optuna.storages.journal import JournalRedisBackend

storage = JournalStorage(JournalRedisBackend(url="redis://127.0.0.1:6379", prefix="pr-6175"))


    return x ** 2


if __name__ == "__main__":
    for i in range(1000):
        study.enqueue_trial({"x": -5.0 + float(i) / 100})

    with ProcessPoolExecutor(max_workers=20) as pool:
        for i in range(100):
            pool.submit(study.optimize, objective, n_trials=10)

    print(f"{len(study.trials)=}")    
    print(f"{len({trial.number for trial in study.trials})=}")    
    print(f"{len({trial._trial_id for trial in study.trials})=}")
```

c-bata (commented):
Changes look good to me. this PR can be merged after:

* Confirming it works as expected with the larger task by @nabenabe0928
* Receiving approval from the second reviewer

c-bata (approved):

optuna/storages/journal/_storage.py
```diff
@@ -320,13 +321,29 @@ def set_trial_state_values(
...
+            if state == TrialState.RUNNING:
+                # NOTE(nabenabe): This section is triggered only when we are using `enqueue_trial`
+                # and `GrpcProxyStorage` in distributed optimization setups and solves the issue
+                # https://github.com/optuna/optuna/issues/6084.
+                # When using gRPC, the current thread may already have popped the trial with
+                # trial_id for another process, potentially leading to a false positive in the
+                # return statement of trial_id == _replay_result.owned_trial_id. To eliminate false
+                # positives, we verify whether another process is already evaluating the trial with
+                # trial_id. If True, it means this query does not update the trial state.
+                existing_trial = self._replay_result._trials.get(trial_id)
```
gen740, line 338: Isn't it necessary to call `self._sync_with_backend()`? Since the thin line uses self._reply_result.

nabenabe0928: Thanks, will do it:)

gen740 (approved):
LGTM!

PR #6179, opened 2025-06-25
Warn when `GPSampler` falls back to `RandomSampler`

sisird864:
## Motivation

Add a runtime warning to GPSampler when it falls back to RandomSampler, per the community Q&A suggestion. This informs users whenever the Gaussian-process sampler degrades to independent sampling (e.g. due to dynamic search-space changes or GP fitting errors) and keeps its behavior consistent with CmaEsSampler.

## Description of the changes

Imports & logger
- In optuna/samplers/_gp/sampler.py, imported the standard warnings module and initialized a module-level logger via get_logger.

Warning helper
- Added a private method _log_independent_sampling(trial, param_name) that emits a warning describing the fallback, mirroring CmaEsSampler.

Guarded fallback
Modified the except Exception block in sample_independent to:
- Check self._warn_independent_sampling
- Call _log_independent_sampling(...) once
- Disable further warnings
- Delegate to the independent sampler

Testing
No new unit tests were added for the log message itself (logging was manually smoke-tested per maintainer guidance).

### DIFF
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -41,6 +41,10 @@ else:
     acqf = _LazyImport("optuna._gp.acqf")
     prior = _LazyImport("optuna._gp.prior")
 
+import logging
+
+
+_logger = logging.getLogger(__name__)
 
 EPS = 1e-10
 
@@ -142,6 +146,12 @@ class GPSampler(BaseSampler):
             subject to change in future releases.
             Currently, the ``constraints_func`` option is not supported for multi-objective
             optimization.
+        warn_independent_sampling:
+            If this is :obj:`True`, a warning message is emitted when
+            the value of a parameter is sampled by using an independent sampler,
+            meaning that no GP model is used in the sampling.
+            Note that the parameters of the first trial in a study are always sampled
+            via an independent sampler, so no warning messages are emitted in this case.
     """
 
     def __init__(
@@ -152,6 +162,7 @@ class GPSampler(BaseSampler):
         n_startup_trials: int = 10,
         deterministic_objective: bool = False,
         constraints_func: Callable[[FrozenTrial], Sequence[float]] | None = None,
+        warn_independent_sampling: bool = True,
     ) -> None:
         self._rng = LazyRandomState(seed)
         self._independent_sampler = independent_sampler or optuna.samplers.RandomSampler(seed=seed)
@@ -165,6 +176,7 @@ class GPSampler(BaseSampler):
         self._constraints_gprs_cache_list: list[gp.GPRegressor] | None = None
         self._deterministic = deterministic_objective
         self._constraints_func = constraints_func
+        self._warn_independent_sampling = warn_independent_sampling
 
         if constraints_func is not None:
             warn_experimental_argument("constraints_func")
@@ -174,6 +186,20 @@ class GPSampler(BaseSampler):
         # NOTE(nabenabe): ehvi in BoTorchSampler uses 20.
         self._n_local_search = 10
         self._tol = 1e-4
+
+    def _log_independent_sampling(self, trial: FrozenTrial, param_name: str) -> None:
+        """Log a warning when GPSampler falls back to independent sampling."""
+        _logger.warning(
+            "The parameter '{}' in trial#{} is sampled independently "
+            "by using `{}` instead of `GPSampler` "
+            "(optimization performance may be degraded). "
+            "`GPSampler` does not support dynamic search space. "
+            "You can suppress this warning by setting `warn_independent_sampling` "
+            "to `False` in the constructor of `GPSampler`, "
+            "if this independent sampling is intended behavior.".format(
+                param_name, trial.number, self._independent_sampler.__class__.__name__
+            )
+        )
 
     def reseed_rng(self) -> None:
         self._rng.rng.seed()
@@ -388,6 +414,11 @@ class GPSampler(BaseSampler):
         param_name: str,
         param_distribution: BaseDistribution,
     ) -> Any:
+        if self._warn_independent_sampling:
+            states = (TrialState.COMPLETE,)
+            complete_trials = study._get_trials(deepcopy=False, states=states, use_cache=True)
+            if len(complete_trials) >= self._n_startup_trials:
+                self._log_independent_sampling(trial, param_name)
         return self._independent_sampler.sample_independent(
             study, trial, param_name, param_distribution
         )


nabenabe0928:

nabenabe0928:
@fusawa-yugo
Could you also review this PR?

sisird864:
Thanks for the feedback. I added these changes:
- Emit only one warning per trial in `sample_relative`
- Disable further warnings after the first occurrence

Pushed the update. Please have another look.

nabenabe0928:
@fusawa-yugo 

Could you please review this PR?

optuna/samplers/_gp/sampler.py
```diff
@@ -273,6 +296,12 @@ def sample_relative(
         trials = study._get_trials(deepcopy=False, states=states, use_cache=True)
 
         if len(trials) < self._n_startup_trials:
+            if self._warn_independent_sampling:
+                # Log warning once for the trial indicating independent sampling
+                self._log_independent_sampling(
+                    trial, "GPSampler fell back to independent sampling for all parameters"
+                )
+                self._warn_independent_sampling = False
```
fusawa-yugo, line 304: Could you revert this?
I think a warning here is unnecessary, since it’s natural to sample independently when the number of samples is insufficient.

optuna/samplers/_gp/sampler.py
```diff
@@ -388,6 +417,9 @@ def sample_independent(
         param_name: str,
         param_distribution: BaseDistribution,
     ) -> Any:
+        if self._warn_independent_sampling:
+            self._log_independent_sampling(trial, param_name)
```
fusawa-yugo: A warning should be displayed when sampling independently and the number of trials are sufficient for the same reason  as previous comment
```suggestion
        if self._warn_independent_sampling:
            states = (TrialState.COMPLETE,)
            complete_trials = study._get_trials(deepcopy=False, states=states, use_cache=True)
            if len(complete_trials) >= self._n_startup_trials:
                self._log_independent_sampling(trial, param_name)
```

optuna/samplers/_gp/sampler.py
```diff
@@ -388,6 +415,8 @@ def sample_independent(
...
```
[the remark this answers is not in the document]
fusawa-yugo: Just a comment: this implementation doesn't match the [counterpart](https://github.com/optuna/optuna/blob/master/optuna/samplers/_cmaes.py#L584-L587) in `_cmaes.py`.

optuna/samplers/_gp/sampler.py
fusawa-yugo: Could you revert this?
I think a warning here is unnecessary, since it’s natural to sample independently when the number of samples is insufficient.

fusawa-yugo (commented):
Thank you for your PR!!
I left some comments. Could you check them?

optuna/samplers/_gp/sampler.py
```diff
@@ -388,6 +417,9 @@ def sample_independent(
...
+            self._warn_independent_sampling = False
```
fusawa-yugo, line 422: Just a comment: this implementation doesn't match the [counterpart](https://github.com/optuna/optuna/blob/master/optuna/samplers/_cmaes.py#L584-L587) in `_cmaes.py.`

fusawa-yugo (commented):
Sorry, please ignore this comment.

optuna/samplers/_gp/sampler.py
1kastner: @fusawa-yugo I agree that this does not need to be a warning. However, could it still be noted in the optimization results in the database? For my research, keeping track of which sampler was used at which instance would be super helpful. I don't mean to do this in this PR but maybe more as a long-term goal of optuna?

sisird864:
Thanks for the review, I’ve updated the PR to:

1. **Revert** the warning in `sample_relative`.
2. **Add** a startup‐threshold guard in `sample_independent` so that we only log the fallback warning once we’ve passed `n_startup_trials`.  

Please have another look. Thank you.

optuna/samplers/_gp/sampler.py
1kastner: I moved this to #6185

optuna/samplers/_gp/sampler.py
```diff
@@ -142,6 +146,9 @@ class GPSampler(BaseSampler):
             subject to change in future releases.
             Currently, the ``constraints_func`` option is not supported for multi-objective
             optimization.
+        warn_independent_sampling:
+            Whether to warn when parameters are sampled independently instead of using
+            the GP model. Defaults to :obj:`True`.
```
nabenabe0928: [nit] I followed the convention.

https://github.com/optuna/optuna/blob/632ab66c6b6c2b2ad09f51e58554d37c2ee3e792/optuna/samplers/_cmaes.py#L145-L150

```suggestion
            If this is :obj:`True`, a warning message is emitted when
            the value of a parameter is sampled by using an independent sampler,
            meaning that no GP model is used in the sampling.

            Note that the parameters of the first trial in a study are always sampled
            via an independent sampler, so no warning messages are emitted in this case.
```

optuna/samplers/_gp/sampler.py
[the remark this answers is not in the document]
nabenabe0928: I do not think we need to disable the warning after the first emission since it is not so obvious for users whether these parameters are repeatedly sampled independently or not.
What do you guys, @sisird864 @fusawa-yugo @y0z @1kastner, think?

nabenabe0928 (commented):
Hey thank you for the PR, I left some comments:)

1kastner:
@nabenabe0928 I agree that repeated logging does not harm. Especially when the code regarding the search space gets more complex, debugging it becomes much more challenging and some support to the users could be great. I wouldn't mind the extra message in my log files, either.

sisird864:
Thanks for the discussion. @1kastner makes a great point. Repeated warnings can help users debug unexpected independent sampling. I’ll remove the one-time disable in `sample_independent` so that every fallback logs a warning. Pushing that now.

nabenabe0928:
This PR relates to:
- https://github.com/optuna/optuna/pull/6188/

optuna/samplers/_gp/sampler.py
nabenabe0928: @sisird864 
Could you please apply the suggestion?
Once the change is applied, this PR looks good to me:)

sisird864: Hello, I’ve added the startup-threshold guard to sample_independent as requested pushed the update. CI is green now. Please take another look. Thanks.

nabenabe0928 (approved):
Thank you for the changes, LGTM")

fusawa-yugo:
LGTM!!
Thank you for your changes.

y0z (approved):
LGTM

PR #6181, opened 2025-06-27
Fix a bug in constrained `GPSampler`

nabenabe0928:
## Motivation

The original implementation takes `best_params` as the `x` fetched by  `argmax(Y[is_feasible])`.
However, what we really want is `x` with the best feasible `y`.

See the example below,

```python
X = [[1.0], [4.0], [2.0]]
Y = [2.0, 3.0, 5.0]
C = [1.0, -1.0, -1.0]
```

In this case, `feasible = [False, True, True]`, meaning that `Y_feasible = [3.0, 5.0]` and `X_feasible = [[4.0], [2.0]]`.
So `X[argmax(Y_feasible)]` becomes `[4.0]` instead of `[2.0]`, which is the actual best feasible params.
I fixed this bug in this PR.

> [!NOTE]
> Here, we aim to maximize `Y`.

## Description of the changes

- See the motivation

### DIFF
--- a/optuna/samplers/_gp/sampler.py
+++ b/optuna/samplers/_gp/sampler.py
@@ -325,27 +325,24 @@ class GPSampler(BaseSampler):
         else:
             assert n_objectives == len(gprs_list) == 1, "Multi-objective has not been supported."
             constraint_vals, is_feasible = _get_constraint_vals_and_feasibility(study, trials)
-            is_all_infeasible = not np.any(is_feasible)
-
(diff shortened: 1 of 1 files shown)


nabenabe0928:
@sawa3030 @kAIto47802 
Could you review this PR?

kAIto47802 (approved):
Thank you for the PR. LGTM!

sawa3030 (approved):
LGTM

y0z (approved):
LGTM

PR #6189, opened 2025-07-02
Fix `TPESampler` with `multivariate` and `constant_liar`

not522:
## Motivation

Combining `TPESampler`'s `multivariate` and `constant_liar` can sometimes cause the `constant_liar` to function improperly during batch optimization.

Running the following script demonstrates that the result remains unchanged whether `constant_liar` is `True` or `False` in the current master branch.

```python
import optuna

N_TRIAL = 50
N_BATCH = 10

for multivariate in (False, True):
    for constant_liar in (False, True):
        sampler = optuna.samplers.TPESampler(
            seed=42,
            multivariate=multivariate,
            constant_liar=constant_liar,
        )

        for i in range(0, N_TRIAL, N_BATCH):
            trials = []
            for j in range(N_BATCH):
                trials.append(study.ask())
            X = [trial.suggest_float("x", -10, 10) for trial in trials]
            Y = [trial.suggest_float("y", -10, 10) for trial in trials]
            for j in range(N_BATCH):
                study.tell(trials[j], X[j] ** 2 + Y[j] ** 2)

            # Skip first random sampling.
            if i > 0:
                plt.plot(X, Y, ".")
        plt.xlim(-10, 10)
        plt.ylim(-10, 10)
        plt.savefig(f"{multivariate}-{constant_liar}.png")
        plt.clf()
```

After applying this PR, we can see that sampling becomes more dispersed when `constant_liar` is set to `True`. (Same colors indicate samples in the same batch.)

- master
  - `multivariate=True`, `constant_liar=False`
True-False
  - `multivariate=True`, `constant_liar=True`
True-True

- PR
  - `multivariate=True`, `constant_liar=False`
True-False
  - `multivariate=True`, `constant_liar=True`
True-True

## Description of the changes

- Store the results of relative sampling in `system_attrs` to make them available to other processes.
- Use them before `RUNNING` trials do actual sampling.

### Speed

It takes slightly longer to save parameters in `system_attrs`, but the difference isn't significant.

<details>

```python
import time
import optuna
from tqdm import tqdm

N_TRIAL = 1000
N_BATCH = 10

for multivariate in (False, True):
    for constant_liar in (False, True):
        start = time.time()
        sampler = optuna.samplers.TPESampler(
            seed=42,
            multivariate=multivariate,
            constant_liar=constant_liar,
        )

        for i in range(0, N_TRIAL, N_BATCH):
            trials = []
            for j in range(N_BATCH):
                trials.append(study.ask())
            X = [trial.suggest_float("x", -10, 10) for trial in trials]
            Y = [trial.suggest_float("y", -10, 10) for trial in trials]
            for j in range(N_BATCH):
                study.tell(trials[j], X[j] ** 2 + Y[j] ** 2)
        print(f"{multivariate=} {constant_liar=} {time.time()-start}")
```

- master

```
multivariate=False constant_liar=False 9.37236499786377
multivariate=False constant_liar=True 11.942036867141724
multivariate=True constant_liar=False 10.05676007270813
multivariate=True constant_liar=True 11.465260982513428
```

- PR

```
multivariate=False constant_liar=False 9.201738119125366
multivariate=False constant_liar=True 12.185273170471191
multivariate=True constant_liar=False 9.935747146606445
multivariate=True constant_liar=True 12.39444088935852
```

</details>

### Optimization performance

The difference in optimization performance becomes identical to simply having or not having `constant_liar`.

<details>

```python
from multiprocessing import Pool
import optuna
import optunahub
from tqdm import tqdm
import numpy as np
from scipy.stats import mannwhitneyu

N_STUDY = 100
N_TRIAL = 1000
N_BATCH = 50
function_id = 3

bbob = optunahub.load_module("benchmarks/bbob")


def run_study(args):
    n_dim, constant_liar, seed = args
    objective = bbob.Problem(function_id=function_id, dimension=n_dim, instance_id=1)
    assert objective.directions == [optuna.study.StudyDirection.MINIMIZE]

    study_name = f"{n_dim}_{constant_liar}_{seed:02d}"
    sampler = optuna.samplers.TPESampler(
        seed=seed,
        multivariate=True,
        constant_liar=constant_liar,
    )
    study = optuna.create_study(study_name=study_name, sampler=sampler)

    for i in range(0, N_TRIAL, N_BATCH):
        trials = [study.ask() for _ in range(N_BATCH)]
        for d in range(n_dim):
            for trial in trials:
                trial.suggest_float(f"x{d}", -5, 5)
        values = [objective(trial) for trial in trials]
        for trial, value in zip(trials, values):
            study.tell(trial, value)

    storage = optuna.storages.JournalStorage(
        optuna.storages.journal.JournalFileBackend(f"./benchmark_{function_id:02d}/{study_name}.log")
    )
    optuna.copy_study(
        from_study_name=study_name,
        from_storage=study._storage,
        to_storage=storage,
    )


def main():
    for n_dim in [2, 3, 5, 10, 20, 40]:
        args = []
        for constant_liar in (True, False):
            for seed in range(N_STUDY):
                args.append((n_dim, constant_liar, seed))
        with Pool(processes=10) as pool:
            with tqdm(total=len(args)) as t:
                for _ in pool.imap_unordered(run_study, args):
                    t.update(1)
        all_best_values = [[], []]
        for constant_liar in (True, False):
            for seed in range(N_STUDY):
                study_name = f"{n_dim}_{constant_liar}_{seed:02d}"
                storage = optuna.storages.JournalStorage(
                    optuna.storages.journal.JournalFileBackend(f"./benchmark_{function_id:02d}/{study_name}.log")
                )
                study = optuna.load_study(storage=storage, study_name=study_name)
                best_values = []
                best_value = float("inf")
                for trial in study.trials:
                    if best_value > trial.value:
                        best_value = trial.value
                    best_values.append(best_value)
                all_best_values[int(constant_liar)].append(best_values)
        all_best_values = np.asarray(all_best_values)
        p_values = []
        for i in range(N_TRIAL):
            p_value = mannwhitneyu(
                all_best_values[1, :, i],
                all_best_values[0, :, i],
                alternative='less',
            ).pvalue
            p_values.append(p_value)
        plt.plot(range(N_TRIAL), p_values, label=f"D={n_dim}")
    plt.ylim(0, 1)
    plt.legend(loc=1)
    plt.savefig(f"benchmark_{function_id:02}.png")


if __name__ == "__main__":
    main()
```

The following figures show the Mann-Whitney U tests. If the line is close to 0, `constant_liar` would result in better performance. Conversely, if it's close to 1, the effect would be the opposite. Overall, the effect isn't clear, but some settings clearly show an advantage with `constant_liar`.

- function_id=3 (Rastrigin Function)

benchmark_03

- function_id=19 (Composite Griewank-Rosenbrock Function F8F2)

benchmark_19

- function_id=22 (Gallagher’s Gaussian 21-hi Peaks Function)

benchmark_22

</details>

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -550,7 +550,9 @@ class TPESampler(BaseSampler):
     ) -> _ParzenEstimator:
         observations = self._get_internal_repr(trials, search_space)
         if handle_below and study._is_multi_objective():
-            param_mask_below = [search_space.keys() <= self._get_params(trial).keys() for trial in trials]
+            param_mask_below = [
+                search_space.keys() <= self._get_params(trial).keys() for trial in trials
+            ]
             weights_below = _calculate_weights_below_for_multi_objective(
                 study, trials, self._constraints_func
             )[param_mask_below]
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -474,7 +474,8 @@ class TPESampler(BaseSampler):
 
     def _get_params(self, trial: FrozenTrial) -> dict[str, Any]:
         if trial.state.is_finished() or not self._multivariate:
-            # NOTE(not522): If not multivariate, `relative_params` does not exist and `system_attrs` access will be unnecessary, so we skip it.
+            # NOTE(not522): If not multivariate, `relative_params` does not exist and
+            # `system_attrs` access will be unnecessary, so we skip it.
             return trial.params
 
         params_strs = []
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -516,6 +516,10 @@ class TPESampler(BaseSampler):
         use_cache = not self._constant_liar
         trials = study._get_trials(deepcopy=False, states=states, use_cache=use_cache)
 
+        if self._constant_liar:
+            # For constant_liar, filter out the current trial.
+            trials = [t for t in trials if trial.number != t.number]
+
         # We divide data into below and above.
         n = sum(trial.state != TrialState.RUNNING for trial in trials)  # Ignore running trials.
         below_trials, above_trials = _split_trials(
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -485,11 +485,9 @@ class TPESampler(BaseSampler):
 
         if len(params_strs) == 0:
             return trial.params
-        else:
-            params_str = "".join(params_strs)
-            params = json.loads(params_str)
-            params.update(trial.params)
-            return params
+        params = json.loads("".join(params_strs))
+        params.update(trial.params)
+        return params
 
     def _get_internal_repr(
         self, trials: list[FrozenTrial], search_space: dict[str, BaseDistribution]
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -474,6 +474,7 @@ class TPESampler(BaseSampler):
 
     def _get_params(self, trial: FrozenTrial) -> dict[str, Any]:
         if trial.state.is_finished() or not self._multivariate:
+            # NOTE(not522): If not multivariate, `relative_params` does not exist and `system_attrs` access will be unnecessary, so we skip it.
             return trial.params
 
         params_strs = []
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -42,6 +42,8 @@ if TYPE_CHECKING:
 EPS = 1e-12
 _logger = get_logger(__name__)
 
+_RELATIVE_PARAMS_KEY = "tpe:relative_params"
+
 
 def default_gamma(x: int) -> int:
     return min(int(np.ceil(0.1 * x)), 25)
@@ -409,9 +411,13 @@ class TPESampler(BaseSampler):
                     if not distribution.single():
                         search_space[name] = distribution
                 params.update(self._sample_relative(study, trial, search_space))
-            return params
         else:
-            return self._sample_relative(study, trial, search_space)
+            params = self._sample_relative(study, trial, search_space)
+
+        if search_space != {} and self._constant_liar:
+            # Store the results of relative sampling to make them available to other processes.
+            study._storage.set_trial_system_attr(trial._trial_id, _RELATIVE_PARAMS_KEY, params)
+        return params
 
     def _sample_relative(
         self, study: Study, trial: FrozenTrial, search_space: dict[str, BaseDistribution]
@@ -457,15 +463,23 @@ class TPESampler(BaseSampler):
 
         return self._sample(study, trial, {param_name: param_distribution})[param_name]
 
+    def _get_params(self, trial: FrozenTrial) -> dict[str, Any]:
+        if trial.state.is_finished():
+            return trial.params
+
+        params = trial.system_attrs.get(_RELATIVE_PARAMS_KEY, {})
+        params.update(trial.params)
+        return params
+
     def _get_internal_repr(
         self, trials: list[FrozenTrial], search_space: dict[str, BaseDistribution]
     ) -> dict[str, np.ndarray]:
         values: dict[str, list[float]] = {param_name: [] for param_name in search_space}
         for trial in trials:
-            if all((param_name in trial.params) for param_name in search_space):
-                for param_name in search_space:
-                    param = trial.params[param_name]
-                    distribution = trial.distributions[param_name]
+            params = self._get_params(trial)
+            if all((param_name in params) for param_name in search_space):
+                for param_name, distribution in search_space.items():
+                    param = params[param_name]
                     values[param_name].append(distribution.to_internal_repr(param))
         return {k: np.asarray(v) for k, v in values.items()}
 
@@ -515,9 +529,8 @@ class TPESampler(BaseSampler):
         if handle_below and study._is_multi_objective():
             param_mask_below = []
             for trial in trials:
-                param_mask_below.append(
-                    all((param_name in trial.params) for param_name in search_space)
-                )
+                params = self._get_params(trial)
+                param_mask_below.append(all((param_name in params) for param_name in search_space))
             weights_below = _calculate_weights_below_for_multi_objective(
                 study, trials, self._constraints_func
             )[param_mask_below]
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -418,7 +418,7 @@ class TPESampler(BaseSampler):
             params = self._sample_relative(study, trial, search_space)
 
         if search_space != {} and self._constant_liar:
-            # Store the results of relative sampling to make them available to other processes.
+            # Share the params obtained by the relative sampling with the other processes.
             params_str = json.dumps(params)
             for i in range(0, len(params_str), _SYSTEM_ATTR_MAX_LENGTH):
                 study._storage.set_trial_system_attr(
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -473,7 +473,7 @@ class TPESampler(BaseSampler):
         return self._sample(study, trial, {param_name: param_distribution})[param_name]
 
     def _get_params(self, trial: FrozenTrial) -> dict[str, Any]:
-        if trial.state.is_finished():
+        if trial.state.is_finished() or not self._multivariate:
             return trial.params
 
         params_strs = []
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from collections.abc import Callable
 from collections.abc import Sequence
+import json
 import math
 from typing import Any
 from typing import cast
@@ -43,6 +44,8 @@ EPS = 1e-12
 _logger = get_logger(__name__)
 
 _RELATIVE_PARAMS_KEY = "tpe:relative_params"
+# The value of system_attrs must be less than 2046 characters on RDBStorage.
+_SYSTEM_ATTR_MAX_LENGTH = 2045
 
 
 def default_gamma(x: int) -> int:
@@ -416,7 +419,13 @@ class TPESampler(BaseSampler):
 
         if search_space != {} and self._constant_liar:
             # Store the results of relative sampling to make them available to other processes.
-            study._storage.set_trial_system_attr(trial._trial_id, _RELATIVE_PARAMS_KEY, params)
+            params_str = json.dumps(params)
+            for i in range(0, len(params_str), _SYSTEM_ATTR_MAX_LENGTH):
+                study._storage.set_trial_system_attr(
+                    trial._trial_id,
+                    f"{_RELATIVE_PARAMS_KEY}:{i // _SYSTEM_ATTR_MAX_LENGTH}",
+                    params_str[i : i + _SYSTEM_ATTR_MAX_LENGTH],
+                )
         return params
 
     def _sample_relative(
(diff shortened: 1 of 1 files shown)


y0z:
@nabenabe0928 @sawa3030 Could you review this PR?

optuna/samplers/_tpe/sampler.py
```diff
@@ -457,15 +472,32 @@ def sample_independent(
...
     ) -> dict[str, np.ndarray]:
         values: dict[str, list[float]] = {param_name: [] for param_name in search_space}
         for trial in trials:
-            if all((param_name in trial.params) for param_name in search_space):
-                for param_name in search_space:
-                    param = trial.params[param_name]
-                    distribution = trial.distributions[param_name]
+            params = self._get_params(trial)
+            if all((param_name in params) for param_name in search_space):
+                for param_name, distribution in search_space.items():
```
nabenabe0928, line 504: > [!NOTE]
> `distribution` is used only for `to_internal_repr` and the return of `to_internal_repr` is not affected by any dynamic search space.
> Please also note that `CategoricalDistribution` cannot be dynamic.

optuna/samplers/_tpe/sampler.py
```diff
@@ -409,9 +414,19 @@ def sample_relative(
                     if not distribution.single():
                         search_space[name] = distribution
                 params.update(self._sample_relative(study, trial, search_space))
-            return params
         else:
-            return self._sample_relative(study, trial, search_space)
+            params = self._sample_relative(study, trial, search_space)
+
+        if search_space != {} and self._constant_liar:
+            # Store the results of relative sampling to make them available to other processes.
```
nabenabe0928:
```suggestion
            # Share the params obtained by the relative sampling with the other processes.
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -457,15 +472,32 @@ def sample_independent(
...
+        i = 0
+        while params_str_i := trial.system_attrs.get(f"{_RELATIVE_PARAMS_KEY}:{i}"):
+            params_str += params_str_i
+            i += 1
+
+        if params_str == "":
+            return trial.params
+        else:
+            params = json.loads(params_str)
+            params.update(trial.params)
```
nabenabe0928: In my understanding, `trial.params` is a subset of `params`, but are there any scenarios where this is not the case?
I mean, doesn't `params.update(trial.params)` do nothing here?

optuna/samplers/_tpe/sampler.py
```diff
@@ -457,15 +472,32 @@ def sample_independent(
...
 
+    def _get_params(self, trial: FrozenTrial) -> dict[str, Any]:
+        if trial.state.is_finished():
+            return trial.params
+
+        params_str = ""
+        i = 0
+        while params_str_i := trial.system_attrs.get(f"{_RELATIVE_PARAMS_KEY}:{i}"):
+            params_str += params_str_i
+            i += 1
```
nabenabe0928: What about using `"".join` since it is quicker and we can expect that string lengths are typically very long when we need to iterate the loop?

nabenabe0928:
The PR looks mostly good to me:)
Let me benchmark this PR once I find a time!

optuna/samplers/_tpe/sampler.py
not522: > trial.params is a subset of params

This is incorrect when the objective function changes.
By the way, using `trial.params` might be more appropriate in this case. 🤔 

```python
import optuna

def objective1(trial):
    x = trial.suggest_float("x", 0, 10)
    y = trial.suggest_float("y", 0, 10)
    return x ** 2 + y ** 2

def objective2(trial):
    x = trial.suggest_float("x", 10, 20)
    y = trial.suggest_float("y", 10, 20)
    return x ** 2 + y ** 2

sampler = optuna.samplers.TPESampler(multivariate=True, constant_liar=True)
study.optimize(objective1, n_trials=10)
study.optimize(objective2, n_trials=1)
```

nabenabe0928: Thank you for the correction, I confirmed it:

```
[I 2025-07-08 05:37:34,237] Trial 9 finished with value: 65.10574953954168 and parameters: {'x': 2.3991078898053453, 'y': 7.703897122405998}. Best is trial 1 with value: 13.028020613050282.
self._relative_params={'x': 1.391003910457944, 'y': 1.5331091061768467}
self._relative_params={'x': 1.391003910457944, 'y': 1.5331091061768467}
[W 2025-07-08 05:37:34,239] The parameter `x` in Trial#10 is sampled independently using `RandomSampler` instead of `TPESampler`, potentially degrading the optimization performance. This fallback happend because dynamic search space is not supported for `multivariate=True`. You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `TPESampler` if this independent sampling is intended behavior.
self._relative_params={'x': 1.391003910457944, 'y': 1.5331091061768467}
self._relative_params={'x': 1.391003910457944, 'y': 1.5331091061768467}
[W 2025-07-08 05:37:34,241] The parameter `y` in Trial#10 is sampled independently using `RandomSampler` instead of `TPESampler`, potentially degrading the optimization performance. This fallback happend because dynamic search space is not supported for `multivariate=True`. You can suppress this warning by setting `warn_independent_sampling` to `False` in the constructor of `TPESampler` if this independent sampling is intended behavior.
trial.params={'x': 13.801593636509187, 'y': 13.520713187030058}
[I 2025-07-08 05:37:34,242] Trial 10 finished with value: 373.2936719932594 and parameters: {'x': 13.801593636509187, 'y': 13.520713187030058}. Best is trial 1 with value: 13.028020613050282.
```

nabenabe0928: > By the way, using trial.params might be more appropriate in this case. 🤔

You mean something like this?

```python
relative_params = ...
trial_params = ...

trial_params.update(relative_params)
```

If so, we should probably use `search_space` to limit the parameters to update:

```python
trial_params.update([relative_params[param_name] for param_name in search_space if param_name in relative_params])
```

For example, I am not sure the exact and appropriate behavior for the following case:

```python
relative_params = {"a": 11.0, "b": 12.0, "c": 100}
trial_params = {"a": 1.0, "b": 2.0}  # Assume `c` is suggested only if a > 10.
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -409,9 +414,19 @@ def sample_relative(
                     if not distribution.single():
                         search_space[name] = distribution
                 params.update(self._sample_relative(study, trial, search_space))
-            return params
         else:
-            return self._sample_relative(study, trial, search_space)
+            params = self._sample_relative(study, trial, search_space)
+
+        if search_space != {} and self._constant_liar:
```
sawa3030: In my understanding, the condition `search_space != {}` is used because when the search space is empty, `params` is also empty, and there's no need to store empty `params`. If that's correct, I was wondering if checking `params != {}` might be a more straightforward approach and could help avoid unnecessary access to the storage.

optuna/samplers/_tpe/sampler.py
```diff
@@ -457,15 +472,32 @@ def sample_independent(
 
         return self._sample(study, trial, {param_name: param_distribution})[param_name]
 
+    def _get_params(self, trial: FrozenTrial) -> dict[str, Any]:
+        if trial.state.is_finished():
```
sawa3030:
```suggestion
        if trial.state.is_finished() or not self._constant_liar:
```
I was wondering if this version might help avoid an unnecessary call to `trial.system_attrs.get`.

not522: When `self._constant_liar` is `False`, the trial is always finished. Meanwhile, when `self._multivariate` is `False`, there was an unnecessary access. I fixed it.

optuna/samplers/_tpe/sampler.py
not522: I mean that we just use `trial.params` when we detect the search space is changed.

optuna/samplers/_tpe/sampler.py
sawa3030: Thank you for the correction!

sawa3030 (approved):
Thank you for all the explanation. LGTM

optuna/samplers/_tpe/sampler.py
```diff
@@ -457,15 +472,33 @@ def sample_independent(
 
         return self._sample(study, trial, {param_name: param_distribution})[param_name]
 
+    def _get_params(self, trial: FrozenTrial) -> dict[str, Any]:
+        if trial.state.is_finished() or not self._multivariate:
```
nabenabe0928, line 480: Please correct me if I'm wrong, but probably either of the following is better?

```suggestion
        if trial.state.is_finished() or not self._constant_liar:
```
or 

```suggestion
        if trial.state.is_finished() or not self._multivariate or not self._constant_liar:
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -457,15 +472,33 @@ def sample_independent(
...
+            params_strs.append(params_str_i)
+            i += 1
+
+        if len(params_strs) == 0:
+            return trial.params
+        else:
+            params_str = "".join(params_strs)
+            params = json.loads(params_str)
+            params.update(trial.params)
+            return params
```
nabenabe0928: [nit] We can reduce the indents here.
```suggestion
        params = json.loads("".join(params_strs))
        params.update(trial.params)
        return params
```

optuna/samplers/_tpe/sampler.py
nabenabe0928: Hm, that is indeed a very good point.
What about adding an inline comment about this?

nabenabe0928 (commented):
This PR looks almost good to me:)
Let me approve this PR once I get the response from @not522 to each of my comments!

optuna/samplers/_tpe/sampler.py
nabenabe0928: I missed the following conversation:
- https://github.com/optuna/optuna/pull/6189#discussion_r2196416364

I got the gist, but could you add an inline comment here?
For example, something like this:
```suggestion
        if trial.state.is_finished() or not self._multivariate:
            # NOTE(not522): If not multivariate, `relative_params` does not exist and `system_attrs` access will be unnecessary, so we skip it.
```

optuna/samplers/_tpe/sampler.py
```diff
@@ -457,15 +472,32 @@ def sample_independent(
 
         return self._sample(study, trial, {param_name: param_distribution})[param_name]
 
+    def _get_params(self, trial: FrozenTrial) -> dict[str, Any]:
+        if trial.state.is_finished() or not self._multivariate:
+            # NOTE(not522): If not multivariate, `relative_params` does not exist and `system_attrs` access will be unnecessary, so we skip it.
```
nabenabe0928: I will merge this PR once this flake8 fix makes the CI pass.
```suggestion
            # NOTE(not522): If not multivariate, `relative_params` does not exist and
            # `system_attrs` access will be unnecessary, so we skip it.
```

nabenabe0928 (approved):
LGTM!

not522:
I noticed `_get_params` isn't working correctly in edge cases. I'm working on figuring out a fix.

nabenabe0928:
@not522 
Sure, thank you for the report... :sob:

not522:
> I noticed _get_params isn't working correctly in edge cases. I'm working on figuring out a fix.

I found that the previous code may use the stale sampling result for `sample_independent` (See trial#10 on the following example).
I fixed it to filter out the current trial. PTAL.

```python
import optuna

def objective1(trial):
    x = trial.suggest_float("x", 0, 10)
    y = trial.suggest_float("y", 0, 10)
    return x ** 2 + y ** 2

def objective2(trial):
    x = trial.suggest_float("x", 10, 20)
    y = trial.suggest_float("y", 10, 20)
    return x ** 2 + y ** 2

sampler = optuna.samplers.TPESampler(multivariate=True, constant_liar=True)
study.optimize(objective1, n_trials=10)
study.optimize(objective2, n_trials=1)
```

nabenabe0928 (approved):
Sorry for the delay, LGTM!

nabenabe0928:
I will merge this PR once @sawa3030 confirms the change!

sawa3030 (approved):
Thank you for the correction. LGTM

PR #6198, opened 2025-07-09
Add `ConstrainedLogEHVI`

kAIto47802:

### DIFF
--- a/optuna/_gp/acqf.py
+++ b/optuna/_gp/acqf.py
@@ -265,3 +265,34 @@ class LogEHVI(BaseAcquisitionFunc):
             non_dominated_box_lower_bounds=self._non_dominated_box_lower_bounds,
             non_dominated_box_upper_bounds=self._non_dominated_box_upper_bounds,
         )
+
+
+class ConstrainedLogEHVI(BaseAcquisitionFunc):
+    def __init__(
+        self,
+        gpr_list: list[GPRegressor],
+        search_space: SearchSpace,
+        Y_train: torch.Tensor,
+        n_qmc_samples: int,
+        qmc_seed: int | None,
+        constraints_gpr_list: list[GPRegressor],
+        constraints_threshold_list: list[float],
+        stabilizing_noise: float = 1e-12,
+    ) -> None:
+        self._acqf = LogEHVI(
+            gpr_list, search_space, Y_train, n_qmc_samples, qmc_seed, stabilizing_noise
+        )
+        self._constraints_acqf_list = [
+            LogPI(_gpr, search_space, _threshold, stabilizing_noise)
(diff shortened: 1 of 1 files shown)


optuna/_gp/acqf.py
```diff
@@ -265,3 +267,40 @@ def eval_acqf(self, x: torch.Tensor) -> torch.Tensor:
...
+        # inverse_squared_lengthscales is used in optim_mixed.py.
+        # cf. https://github.com/optuna/optuna/blob/v4.3.0/optuna/_gp/optim_mixed.py#L200-L209
+        super().__init__(np.mean([gpr.length_scales for gpr in gpr_list], axis=0), search_space)
+
+    def eval_acqf(self, x: torch.Tensor) -> torch.Tensor:
+        return (
+            self._acqf.eval_acqf(x)
+            if self._acqf is not None
+            else torch.zeros(x.shape[:-1], dtype=torch.float64)
+        ) + sum(acqf.eval_acqf(x) for acqf in self._constraints_acqf_list)
```
nabenabe0928: What about it?
We can avoid unnecessary zero sum with the change:)
```suggestion
        log_feas_prob = sum(acqf.eval_acqf(x) for acqf in self._constraints_acqf_list)
        if self._acqf is None:
            return log_feas_prob
        return log_feas_prob + self._acqf.eval_acqf(x)
```

optuna/_gp/acqf.py
```diff
@@ -265,3 +267,40 @@ def eval_acqf(self, x: torch.Tensor) -> torch.Tensor:
...
+        gpr_list: list[GPRegressor],
+        search_space: SearchSpace,
+        Y_feasible: torch.Tensor | None,
+        n_qmc_samples: int,
+        qmc_seed: int | None,
+        constraints_gpr_list: list[GPRegressor],
+        constraints_threshold_list: list[float],
+        stabilizing_noise: float = 1e-12,
+    ) -> None:
+        # NOTE(kAIto47802): It is sufficient to only passing the feasible objective values
```
nabenabe0928:
```suggestion
        assert len(constraints_gpr_list) == len(constraints_threshold_list) and len(constraints_gpr_list) > 0
        # NOTE(kAIto47802): It is sufficient to only passing the feasible objective values
```

nabenabe0928:
This PR looks almost good to me:)
Could you, @kAIto47802 , please add an inline comment saying that "TODO: Rename Y_train" in `LogEHVI` as `Y_train` would not be accurate anymore?

sawa3030:
This PR looks almost good to me. I’ll take another look once the part pointed out by @nabenabe0928 is addressed.

sawa3030 (approved):
LGTM

Issue #6029, opened 2025-04-01
Use `TYPE_CHECKING` if necessary


nabenabe0928:
> [!NOTE]
> An example is available here:
> - https://github.com/optuna/optuna/pull/6030

### Motivation

Some modules are imported only for type checking, but this may cause circular import, leading to an error.
Although such an error is not a big deal for most cases, it may require annoying changes sometimes as in:
- https://github.com/optuna/optuna/pull/5391

The issue above motivates using `TYPE_CHECKING`.

### Suggestion

First, we can check which files to modify by the following:

```shell
$ ruff check . --select TCH
```

In order to reduce review costs and to encourage more contributors to work on it, please, as a rule, fix one file per PR.

nabenabe0928:
```shell
$ ruff check . --select TCH --output-format=concise | cut -d: -f1 | sort -u
```

The files to work on (the list is updated time to time):

- [ ] `optuna/integration/__init__.py`
- [ ] `optuna/integration/lightgbm.py`
- [ ] `optuna/samplers/_tpe/parzen_estimator.py`
- [ ] `optuna/storages/_base.py`
- [ ] `optuna/storages/_cached_storage.py`
- [ ] `optuna/storages/_callbacks.py`
- [ ] `optuna/storages/_grpc/client.py`
- [ ] `optuna/storages/_grpc/server.py`
- [ ] `optuna/storages/_grpc/servicer.py`
- [ ] `optuna/storages/_heartbeat.py`
- [ ] `optuna/storages/_in_memory.py`
- [ ] `optuna/storages/_rdb/storage.py`
- [ ] `optuna/storages/journal/_base.py`
- [ ] `optuna/storages/journal/_file.py`
- [ ] `optuna/storages/journal/_redis.py`
- [ ] `optuna/storages/journal/_storage.py`
- [ ] `optuna/study/_constrained_optimization.py`
- [ ] `optuna/study/_dataframe.py`
- [ ] `optuna/study/_multi_objective.py`
- [ ] `optuna/study/_study_summary.py`
- [ ] `optuna/terminator/improvement/evaluator.py`
- [ ] `optuna/terminator/terminator.py`
- [ ] `optuna/visualization/_parallel_coordinate.py`
- [ ] `optuna/visualization/_utils.py`
- [ ] `optuna/visualization/matplotlib/_hypervolume_history.py`
- [ ] `optuna/visualization/matplotlib/_intermediate_values.py`
[...]

PR #6204, opened 2025-07-16
Use `TYPE_CHECKING` in `optuna/_gp/acqf.py` to avoid circular imports

CarvedCoder:
## Motivation
Currently, two modules (GPRegressor and SearchSpace) are imported unconditionally in optuna/_gp/acqf.py, which can trigger circular‐import errors at runtime. By moving these imports into a if TYPE_CHECKING: block, we ensure that they’re only used for static type hints and never executed during normal imports.

## Description of the changes
This fixes issue  #6029 by wrapping the two imports in optuna/_gp/acqf.py with if TYPE_CHECKING: so they’re only for type hints.
Also ran python3 -m flake8 optuna/_gp/acqf.py —> no more TC001 warnings in that file.

Fixes  #6029

### DIFF
--- a/optuna/_gp/acqf.py
+++ b/optuna/_gp/acqf.py
@@ -7,14 +7,15 @@ from typing import TYPE_CHECKING
 
 import numpy as np
 
-from optuna._gp.gp import GPRegressor
-from optuna._gp.search_space import SearchSpace
 from optuna._hypervolume import get_non_dominated_box_bounds
 from optuna.study._multi_objective import _is_pareto_front
 
 
 if TYPE_CHECKING:
     import torch
+
+    from optuna._gp.gp import GPRegressor
+    from optuna._gp.search_space import SearchSpace
 else:
     from optuna._imports import _LazyImport
 


optuna/_gp/acqf.py
```diff
@@ -7,8 +7,11 @@
 
 import numpy as np
 
-from optuna._gp.gp import GPRegressor
-from optuna._gp.search_space import SearchSpace
+
+if TYPE_CHECKING:
+    from optuna._gp.gp import GPRegressor
+    from optuna._gp.search_space import SearchSpace
+
```
nabenabe0928: The `TYPE_CHECKING` section is already prepared below, so could you please move these module imports to the section?

CarvedCoder: Sure! Thanks for pointing that out , I’ll move the imports into the existing TYPE_CHECKING section and update the PR shortly

nabenabe0928 (approved):
Thank you for the change, LGTM:)

PR #6205, opened 2025-07-16
Use TYPE_CHECKING in optuna/_gp/optim_mixed.py to avoid circular imports

Subodh-12:
## Motivation
 Currently, a module (optuna._gp.acqf.BaseAcquisitionFunc) is imported unconditionally in optuna/_gp/opim_mixed.py, which can trigger circular‐import errors at runtime. By moving these imports into a if TYPE_CHECKING: block, we ensure they’re only used for static type hints and never executed during normal imports. This improves import safety and prevents spurious CI failures or runtime import errors when using the GP acquisition‐function code.

## Description of the changes
 python3 -m flake8 optuna/_gp/opim_mixed.py
emits no TC001 warnings

Fixed #6029

### DIFF
--- a/optuna/_gp/optim_mixed.py
+++ b/optuna/_gp/optim_mixed.py
@@ -4,8 +4,7 @@ import math
 from typing import TYPE_CHECKING
 
 import numpy as np
-if TYPE_CHECKING:
-    from optuna._gp.acqf import BaseAcquisitionFunc
+
 from optuna._gp.search_space import normalize_one_param
 from optuna._gp.search_space import sample_normalized_params
 from optuna._gp.search_space import ScaleType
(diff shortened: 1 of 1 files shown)


optuna/_gp/optim_mixed.py
```diff
@@ -4,8 +4,8 @@
 from typing import TYPE_CHECKING
 
 import numpy as np
-
-from optuna._gp.acqf import BaseAcquisitionFunc
+if TYPE_CHECKING:
+    from optuna._gp.acqf import BaseAcquisitionFunc
```
nabenabe0928: Could you merge the `TYPE_CHECKING` section?
It is already defined below.

Subodh-12:
issue has been fixed

nabenabe0928 (approved):

Issue #6206, opened 2025-07-17
CI fails frequently after introducing `pytest-xdist`


nabenabe0928:
### Expected behavior

I will update the issue description whenever I find another failure.

In principle, the CI sometimes fails due to the parallel run by the following PR: 
- https://github.com/optuna/optuna/pull/6170

### Environment

CI environment

### Error messages, stack traces, or logs

```shell
- https://github.com/optuna/optuna/actions/runs/16332163798/job/46137014991#step:10:426


_________________ test_log_completed_trial_skip_storage_access _________________
[gw4] linux -- Python 3.9.23 /opt/hostedtoolcache/Python/3.9.23/x64/bin/python

    def test_log_completed_trial_skip_storage_access() -> None:
        study = create_study()
    
        # Create a trial to retrieve it as the `study.best_trial`.
        study.optimize(lambda _: 0.0, n_trials=1)
        frozen_trial = study.best_trial
    
        storage = study._storage
    
        with patch.object(storage, "get_best_trial", wraps=storage.get_best_trial) as mock_object:
            study._log_completed_trial(frozen_trial.values, frozen_trial.number, frozen_trial.params)
>           assert mock_object.call_count == 1
E           AssertionError: assert 0 == 1
E            +  where 0 = <MagicMock name='get_best_trial' id='140017043141632'>.call_count

tests/study_tests/test_study.py:1154: AssertionError


- https://github.com/optuna/optuna/actions/runs/16327000266/job/46119643469#step:12:301


____ test_get_timeline_plot_with_killed_running_trials[0.0-plot_timeline0] _____
[gw7] linux -- Python 3.8.18 /opt/hostedtoolcache/Python/3.8.18/x64/bin/python

plot_timeline = <function plot_timeline at 0x7ff526abbb80>, waiting_time = 0.0

    @parametrize_plot_timeline
    @pytest.mark.parametrize("waiting_time", [0.0, 1.5])
    def test_get_timeline_plot_with_killed_running_trials(
        plot_timeline: Callable[..., Any], waiting_time: float
    ) -> None:
        def _objective_with_sleep(trial: optuna.Trial) -> float:
            sleep_start_datetime = datetime.datetime.now()
            # Spin waiting is used here because high accuracy is necessary even in weak VM.
            # Please check the motivation of the bugfix in https://github.com/optuna/optuna/pull/5549/
            while datetime.datetime.now() - sleep_start_datetime < datetime.timedelta(seconds=0.1):
[...]

PR #6210, opened 2025-07-18
Delete `test_get_timeline_plot_with_killed_running_trials`

gen740:
## Motivation

Related to #6206.

`test_get_timeline_plot_with_killed_running_trials` has become brittle since we introduced `pytest-xdist`. Under heavy CI load, execution is considerably slower, and the test now fails occasionally.

In addition, the test couples tightly to Plotly’s internal structure:
```
bar_length_in_milliseconds = figure["data"][1]["x"][0]
```
Because these internals may change with any Plotly update, the test is easily broken.
Given the maintenance cost and limited benefit, I believe we should remove it.

## Description of the changes

- Delete test_get_timeline_plot_with_killed_running_trials


c-bata (approved):
LGTM!

PR #6211, opened 2025-07-18
Fix fragile test `test_experimental`

gen740:
## Motivation

Related to #6206

According to #6206, `test_experimental` occasionally fail.
```
_______________________ test_experimental_func_decorator _______________________
[gw0] linux -- Python 3.13.5 /opt/hostedtoolcache/Python/3.13.5/x64/bin/python

    def test_experimental_func_decorator() -> None:
        version = "1.1.0"
        decorator_experimental = _experimental.experimental_func(version)
        assert callable(decorator_experimental)
    
        decorated_func = decorator_experimental(_sample_func)
        assert decorated_func.__name__ == _sample_func.__name__
        assert decorated_func.__doc__ == _experimental._EXPERIMENTAL_NOTE_TEMPLATE.format(ver=version)
    
        with pytest.warns(ExperimentalWarning) as warnings:
            decorated_func()
    
>       (warning,) = warnings
        ^^^^^^^^^^
E       ValueError: too many values to unpack (expected 1)

tests/test_experimental.py:61: ValueError
```
However, this error appears to be deeply rooted in pytest-xdist, and it’s very hard to reproduce in a local environment. It seems the issue was a race condition with pytest.warns, since pytest is not thread-safe, but xdist is a multi-process plugin and does not create any additional threads.

I believe checking the warning message is excessive.

## Description of the changes

- Remove the logic that checks for ExperimentalWarning output.

### DIFF
--- a/tests/test_experimental.py
+++ b/tests/test_experimental.py
@@ -55,14 +55,8 @@ def test_experimental_func_decorator() -> None:
     assert decorated_func.__name__ == _sample_func.__name__
     assert decorated_func.__doc__ == _experimental._EXPERIMENTAL_NOTE_TEMPLATE.format(ver=version)
 
-    with pytest.warns(ExperimentalWarning) as warnings:
+    with pytest.warns(ExperimentalWarning):
         decorated_func()
-
-    (warning,) = warnings
-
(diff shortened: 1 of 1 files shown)


c-bata (approved):
LGTM.

PR #6215, opened 2025-07-24
Remove unnecessary NumPy usage

nabenabe0928:
## Motivation

`math` is usually quicker than `numpy` for a scalar, so I replaced `numpy` with `math`.

## Description of the changes

- Use `math` instead of `numpy` for a scalar call

> [!NOTE]
> `math.ceil` returns an integer value instead of a floating number. [ref](https://docs.python.org/3/library/math.html#math.ceil)

## Benchmarking

```python
In [1]: import math
   ...: import numpy as np
   ...: 
   ...: 
   ...: %timeit math.isinf(100)
   ...: %timeit np.isinf(100)    
16 ns ± 0.776 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
390 ns ± 9.09 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [2]: %timeit 1000**0.5
   ...: %timeit math.sqrt(1000)
   ...: %timeit np.sqrt(1000)
3.37 ns ± 0.0315 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
21.1 ns ± 0.662 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
573 ns ± 1.84 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [3]: %timeit math.ceil(0.1 * 777)
   ...: %timeit np.ceil(0.1 * 777)
14.6 ns ± 0.269 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
394 ns ± 2.01 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -49,7 +49,7 @@ def default_gamma(x: int) -> int:
 
 
 def hyperopt_default_gamma(x: int) -> int:
-    return min(int(np.ceil(0.25 * x**0.5)), 25)
+    return min(math.ceil(0.25 * x**0.5), 25)
 
 
 def default_weights(x: int) -> np.ndarray:
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -45,11 +45,11 @@ _logger = get_logger(__name__)
 
 
 def default_gamma(x: int) -> int:
-    return min(int(np.ceil(0.1 * x)), 25)
+    return min(math.ceil(0.1 * x), 25)
 
 
 def hyperopt_default_gamma(x: int) -> int:
-    return min(int(np.ceil(0.25 * np.sqrt(x))), 25)
+    return min(int(np.ceil(0.25 * x**0.5)), 25)
 
 
 def default_weights(x: int) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


y0z:
@sawa3030 
Could you review this PR?

optuna/samplers/_tpe/sampler.py
```diff
@@ -45,11 +45,11 @@
 
 
 def default_gamma(x: int) -> int:
-    return min(int(np.ceil(0.1 * x)), 25)
+    return min(math.ceil(0.1 * x), 25)
 
 
 def hyperopt_default_gamma(x: int) -> int:
-    return min(int(np.ceil(0.25 * np.sqrt(x))), 25)
+    return min(int(np.ceil(0.25 * x**0.5)), 25)
```
sawa3030:
```suggestion
    return min(math.ceil(0.25 * x**0.5), 25)
```
Would it be possible to use `math.ceil` here as well?

nabenabe0928: Good catch, thanks:)

y0z (approved):
LGTM

sawa3030 (approved):
LGTM

PR #6218, opened 2025-07-25
Use subset comparator to judge if trials are included in search space

nabenabe0928:
## Motivation

`to_internal_repr` in `TPESampler` is very slow, so I would like to speed it up with the following change:

```python
In [1]: d = {f"x{i}": i for i in range(50)}

In [2]: s = {f"x{i}": i for i in range(50)}

In [3]: %timeit s.keys() <= d.keys()
569 ns ± 4.71 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: %timeit all(p in d.keys() for p in s)
1.93 μs ± 32.8 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

```

## Description of the changes

- Use the subset comparator to speed up the internal repr judge

### DIFF
--- a/optuna/samplers/_tpe/sampler.py
+++ b/optuna/samplers/_tpe/sampler.py
@@ -466,7 +466,7 @@ class TPESampler(BaseSampler):
     ) -> dict[str, np.ndarray]:
         values: dict[str, list[float]] = {param_name: [] for param_name in search_space}
         for trial in trials:
-            if all((param_name in trial.params) for param_name in search_space):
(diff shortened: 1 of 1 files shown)


nabenabe0928:
@kAIto47802 @y0z Could you review this PR?

y0z (approved):
LGTM

kAIto47802 (approved):
LGTM!

PR #6228, opened 2025-08-01
Support 1D Search Spaces in `CmaEsSampler`

sawa3030:
## Motivation
Thanks to [CMA-ES PR #190](https://github.com/CyberAgentAILab/cmaes/pull/190), the CMA-ES library now supports 1-dimensional search spaces. This PR updates CmaEsSampler to allow the use of 1D search spaces accordingly.

## Description of the changes
- Removed the logic that falls back to the independent sampler for 1D search spaces.
- Since SepCMA currently does not support 1D search spaces (see [source](https://github.com/CyberAgentAILab/cmaes/blob/main/cmaes/_sepcma.py#L82)), SepCMA will be bypassed when `use_separable_cma=True` and the search space is 1D.
- Updated the `cmaes` version in `pyproject.toml`.

### DIFF
--- a/optuna/samplers/_cmaes.py
+++ b/optuna/samplers/_cmaes.py
@@ -377,7 +377,7 @@ class CmaEsSampler(BaseSampler):
 
         optimizer = self._restore_optimizer(completed_trials)
         if optimizer is None:
-            optimizer = self._init_optimizer(trans, study.direction, len(search_space) == 1)
+            optimizer = self._init_optimizer(trans, study.direction)
 
         if optimizer.dim != len(trans.bounds):
             if self._warn_independent_sampling:
@@ -478,7 +478,6 @@ class CmaEsSampler(BaseSampler):
         self,
         trans: _SearchSpaceTransform,
         direction: StudyDirection,
-        is_single_dimension: bool,
     ) -> "CmaClass":
         lower_bounds = trans.bounds[:, 0]
         upper_bounds = trans.bounds[:, 1]
@@ -520,7 +519,7 @@ class CmaEsSampler(BaseSampler):
         sigma0 = max(sigma0, _EPS)
 
         if self._use_separable_cma:
-            if is_single_dimension:
+            if len(trans.bounds) == 1:
                 warnings.warn(
                     "Separable CMA-ES does not operate meaningfully on single-dimensional "
                     "search spaces. The setting `use_separable_cma=True` will be ignored.",
--- a/optuna/samplers/_cmaes.py
+++ b/optuna/samplers/_cmaes.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from collections.abc import Sequence
 import copy
-from importlib.metadata import version
 import math
 import pickle
 from typing import Any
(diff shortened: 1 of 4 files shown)


sawa3030:
Other options of CmaEsSampler worked without errors even when using a 1D search space.

optuna/samplers/_cmaes.py
```diff
@@ -529,7 +519,7 @@ def _init_optimizer(
         # Avoid ZeroDivisionError in cmaes.
         sigma0 = max(sigma0, _EPS)
 
-        if self._use_separable_cma:
+        if self._use_separable_cma and not is_single_dimension:
```
y0z: It would be helpful to issue a warning when attempting to use sepCMA in 1D, as it is meaningless.

c-bata:

y0z (approved):
LGTM.

I also confirmed that this change works `with_margin=True`.
```python
optuna.samplers.CmaEsSampler(seed=42, with_margin=True)
```

nabenabe0928:
Let me re-assign from @kAIto47802 to @y0z  !

optuna/samplers/_cmaes.py
```diff
@@ -489,6 +478,7 @@ def _init_optimizer(
         self,
         trans: _SearchSpaceTransform,
         direction: StudyDirection,
+        is_single_dimension: bool,
```
c-bata: How about checking `trans.bounds` instead of adding a new argument here?

tests/samplers_tests/test_cmaes.py
```diff
@@ -252,19 +252,6 @@ def test_infer_relative_search_space_1d() -> None:
     assert sampler.infer_relative_search_space(study, study.best_trial) == {}
 
 
-def test_sample_relative_1d() -> None:
```
c-bata, line 255: Could you add test scenarios for vanilla CMA-ES, CMA-ES with margin, and warm-starting CMA-ES on a one-dimensional search space?

Regarding the warm-starting CMA-ES, you can run it like below:

```python
import optuna


def source_task(trial):
    return trial.suggest_float("x", -10, 10) ** 2


def target_task(trial):
    return (trial.suggest_float("x", -10, 10) - 5) ** 2


source_study = optuna.create_study(sampler=optuna.samplers.CmaEsSampler())
source_study.optimize(source_task, n_trials=100)

target_study = optuna.create_study(sampler=optuna.samplers.CmaEsSampler(
    source_trials=source_study.trials
))
target_study.optimize(target_task, n_trials=100)
```

sawa3030: Thank you for the feedback! I’ve added the following tests:

- `test_1d_search_space_with_margin` for both vanilla CMA-ES and CMA-ES with margin.
- `test_warm_starting_cmaes_with_1d_search_space`, which is essentially a 1D adaptation of the existing `test_warm_starting_cmaes`.

c-bata: Tests looks good to me. Thank you for the update.

PR #6243, opened 2025-08-13
Move `fit_kernel_params` to `GPRegressor`

nabenabe0928:
## Motivation

This PR is necessary to cache the pair-wise distances of `X_train`, which speeds up `GPSampler`, done in:
- https://github.com/optuna/optuna/pull/6244

## Description of the changes

- Move `fit_kernel_params` to `GPRegressor`

### DIFF
--- a/optuna/_gp/gp.py
+++ b/optuna/_gp/gp.py
@@ -287,6 +287,7 @@ def fit_kernel_params(
     gtol: float = 1e-2,
 ) -> GPRegressor:
     default_kernel_params = torch.ones(X.shape[1] + 2, dtype=torch.float64)
+    # TODO: Move this function into a method of `GPRegressor`
 
     def _default_gpr() -> GPRegressor:
         return GPRegressor(
--- a/optuna/_gp/gp.py
+++ b/optuna/_gp/gp.py
@@ -117,7 +117,7 @@ class GPRegressor:
     def _cache_matrix(self) -> None:
         assert (
             self._cov_Y_Y_inv is None and self._cov_Y_Y_inv_Y is None
-        ), "Cannot cache_matrix more than once."
+        ), "Cannot call cache_matrix more than once."
         with torch.no_grad():
             cov_Y_Y = self.kernel(self._X_train, self._X_train).detach().numpy()
 
--- a/optuna/_gp/gp.py
+++ b/optuna/_gp/gp.py
@@ -127,6 +127,12 @@ class GPRegressor:
         # NOTE(nabenabe): Here we use NumPy to guarantee the reproducibility from the past.
         self._cov_Y_Y_inv = torch.from_numpy(cov_Y_Y_inv)
         self._cov_Y_Y_inv_Y = torch.from_numpy(cov_Y_Y_inv_Y)
+        self.inverse_squared_lengthscales = self.inverse_squared_lengthscales.detach()
+        self.inverse_squared_lengthscales.grad = None
+        self.kernel_scale = self.kernel_scale.detach()
+        self.kernel_scale.grad = None
+        self.noise_var = self.noise_var.detach()
+        self.noise_var.grad = None
 
     def kernel(self, X1: torch.Tensor, X2: torch.Tensor) -> torch.Tensor:
         """
--- a/optuna/_gp/gp.py
+++ b/optuna/_gp/gp.py
@@ -215,7 +215,7 @@ class GPRegressor:
         minimum_noise: float,
         deterministic_objective: bool,
         gtol: float,
-    ) -> GPRegressor:
+    ) -> Self:
         n_params = self._X_train.shape[1]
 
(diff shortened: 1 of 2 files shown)


y0z:
@gen740 @kAIto47802 Could you review this PR?

optuna/_gp/gp.py
```diff
@@ -200,87 +209,71 @@ def marginal_log_likelihood(self) -> torch.Tensor:  # Scalar
...
-    initial_raw_params = np.concatenate(
-        [
-            np.log(gpr_cache.inverse_squared_lengthscales.detach().numpy()),
+    def _fit_kernel_params(
+        self,
+        log_prior: Callable[[GPRegressor], torch.Tensor],
+        minimum_noise: float,
+        deterministic_objective: bool,
+        gtol: float,
+    ) -> GPRegressor:
```
kAIto47802, line 218: How about using `Self` as noted in PEP 673[^1]? We can use it by importing it from `typing`.
```suggestion
    ) -> Self:
```

[^1]: https://peps.python.org/pep-0673/

nabenabe0928: Let me revert this change because `from typing import Self` is available only on Python 3.11+.
We can address here after we drop Python 3.10.
But please note that we can import it from `typing_extensions`.
To avoid the branching at the import section, I will simply revert.

optuna/_gp/gp.py
```diff
@@ -200,87 +209,71 @@ def marginal_log_likelihood(self) -> torch.Tensor:  # Scalar
...
-    initial_raw_params = np.concatenate(
-        [
-            np.log(gpr_cache.inverse_squared_lengthscales.detach().numpy()),
+    def _fit_kernel_params(
+        self,
+        log_prior: Callable[[GPRegressor], torch.Tensor],
+        minimum_noise: float,
+        deterministic_objective: bool,
+        gtol: float,
+    ) -> Self:
```
nabenabe0928:
```suggestion
    ) -> GPRegressor:
```

gen740:
I used the following benchmark code (in #6244):
```python
import optuna


    return sum(trial.suggest_float(f"x{i}", -5, 5)**2 for i in range(5))


sampler = optuna.samplers.GPSampler(seed=0)
study.optimize(objective, n_trials=300)
print((study.trials[-1].datetime_complete - study.trials[0].datetime_start).total_seconds())
```
And I found that the master branch is much faster:

- Master: 39.261927s
- This PR: 55.45938s

I tested several times, but the results were consistent.

nabenabe0928:
@gen740 
Is it probably because I haven't merged the upstream master?
I pushed the latest version, so could you please benchmark again?

gen740:
@nabenabe0928 
Thank you! I reran the benchmark and confirmed that the performance looks identical.
I also used wfg4 and ran the multi-objective optimization benchmark (see the code below), but the performance is almost the same as on the master branch.

- master: 143.247278
- this pr: 142.377885

```python
import optuna
import optunahub

wfg = optunahub.load_module("benchmarks/wfg")
wfg4 = wfg.Problem(function_id=4, n_objectives=4, dimension=8, k=6)

sampler = optuna.samplers.GPSampler(seed=0)
study = optuna.create_study(
    sampler=sampler,
    directions=wfg4.directions,
)

study.optimize(wfg4, n_trials=100)

print(
    (
        study.trials[-1].datetime_complete - study.trials[0].datetime_start
    ).total_seconds()
)
```

gen740 (approved):
LGTM!

optuna/_gp/gp.py
```diff
@@ -294,30 +287,37 @@ def fit_kernel_params(
...
     default_kernel_params = torch.ones(X.shape[1] + 2, dtype=torch.float64)
-    default_gpr_cache = GPRegressor(
-        is_categorical=torch.from_numpy(is_categorical),
-        X_train=torch.from_numpy(X),
-        y_train=torch.from_numpy(Y),
-        inverse_squared_lengthscales=default_kernel_params[:-2].clone(),
-        kernel_scale=default_kernel_params[-2].clone(),
-        noise_var=default_kernel_params[-1].clone(),
-    )
+
```
kAIto47802, line 291: I'd like to move this function into a method of `GPRegressor` as well, since it seems inappropriate for the `fit_kernel_params` function to create an instance of `GPRegressor`. Currently, the `fit_kernel_params` function has two responsibilities: creating a `GPRegressor` instance and fitting its kernel parameters, violating the single-responsibility principle.

```suggestion
    # TODO: Move this function into a method of `GPRegressor`

```

If we insist on leaving this function as is without further refactoring, we should make the `_fit_kernel_params` method public (i.e., remove the leading underscore), since it is not internal and is used outside of other `GPRegressor` methods.

kAIto47802 (approved):
Thank you for the update! LGTM!

PR #6244, opened 2025-08-13
Cache pair-wise distances to speed up `GPSampler`

nabenabe0928:
## Motivation

This PR speeds up `GPSampler` by skipping the pair-wise distance calculations in the kernel parameter fitting.
Please note that this PR depends on:
- https://github.com/optuna/optuna/pull/6243

> [!IMPORTANT]
> 7% speedup by this PR (240 sec to 223 sec) in the benchmark below :tada: 

## Description of the changes

- Cache the pair-wise distances

```python
import optuna


    return sum(trial.suggest_float(f"x{i}", -5, 5)**2 for i in range(5))


sampler = optuna.samplers.GPSampler(seed=0)
study.optimize(objective, n_trials=300)
print((study.trials[-1].datetime_complete - study.trials[0].datetime_start).total_seconds())
```

### DIFF
--- a/optuna/_gp/gp.py
+++ b/optuna/_gp/gp.py
@@ -103,6 +103,10 @@ class GPRegressor:
         self._is_categorical = is_categorical
         self._X_train = X_train
         self._y_train = y_train
+        self._squared_X_diff = (X_train[..., None, :] - X_train[..., None, :, :]).square()
+        self._squared_X_diff[..., self._is_categorical] = (
+            self._squared_X_diff[..., self._is_categorical] > 0.0
+        ).type(torch.float64)
         self._cov_Y_Y_inv: torch.Tensor | None = None
         self._cov_Y_Y_inv_Y: torch.Tensor | None = None
         # TODO(nabenabe): Rename the attributes to private with `_`.
@@ -119,7 +123,7 @@ class GPRegressor:
             self._cov_Y_Y_inv is None and self._cov_Y_Y_inv_Y is None
         ), "Cannot call cache_matrix more than once."
         with torch.no_grad():
-            cov_Y_Y = self.kernel(self._X_train, self._X_train).detach().numpy()
+            cov_Y_Y = self.kernel().detach().numpy()
 
         cov_Y_Y[np.diag_indices(self._X_train.shape[0])] += self.noise_var.item()
         cov_Y_Y_inv = np.linalg.inv(cov_Y_Y)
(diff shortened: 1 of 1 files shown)


nabenabe0928:
@sawa3030 @kAIto47802 
Could you review this PR?

optuna/_gp/gp.py
```diff
@@ -147,8 +153,17 @@ def kernel(self, X1: torch.Tensor, X2: torch.Tensor) -> torch.Tensor:
...
         """
-        d2 = (X1[..., :, None, :] - X2[..., None, :, :]) ** 2
-        d2[..., self._is_categorical] = (d2[..., self._is_categorical] > 0.0).type(torch.float64)
+        if X1 is None:
+            assert X2 is None
+            d2 = self._squared_X_diff
+        else:
+            if X2 is None:
+                X2 = self._X_train
+            d2 = (X1[..., None, :] - X2[..., None, :, :]) ** 2
```
sawa3030: I'm a bit confused about this part. Would it be more straightforward to write it as `d2 = (X1[..., :, None, :] - X2[..., None, :, :]) ** 2`?

nabenabe0928: Thank you for pointing it out:)
`X1[..., :, None, :] - X2[..., None, :, :]` first requires `X1.shape[-1] == X2.shape[-1]` and `len(X1.shape) >= 2` and `len(X2.shape) >= 2`.

The newer version loosens the requirements to `X1.shape[-1] == X2.shape[-1]` and `len(X1.shape) >= 1` and `len(X2.shape) >= 2`.

It essentially makes the following line look cleaner:

- https://github.com/optuna/optuna/pull/6244/files#diff-ece0405c7ca361941c87c91dd91c9199fb904968c172f5b0c8d8283f9b4f60b4R184


```python
# Original (x must be 2 dimensions or more, requiring [..., None, :].)
cov_fx_fX = self.kernel(x[..., None, :])[..., 0, :]
# New (x can be 1 dimension, eliminating the need of [..., None, :] and flitering afterwards.)
cov_fx_fX = self.kernel(x)
```

sawa3030 (approved):
Thank you for the explanation. LGTM

optuna/_gp/gp.py
```diff
@@ -103,6 +103,10 @@ def __init__(
         self._is_categorical = is_categorical
         self._X_train = X_train
         self._y_train = y_train
+        self._squared_X_diff = (X_train[..., None, :] - X_train[..., None, :, :]).square()
+        self._squared_X_diff[..., self._is_categorical] = (
+            self._squared_X_diff[..., self._is_categorical] > 0.0
+        ).type(torch.float64)
```
kAIto47802: How about limiting this only when `if self._is_categorical.any():`, to maintain consistency with the line 164:
https://github.com/nabenabe0928/optuna/blob/enhance/cache-pair-wise-distance/optuna/_gp/gp.py#L164
```suggestion
        if self._is_categorical.any():
            self._squared_X_diff[..., self._is_categorical] = (
                self._squared_X_diff[..., self._is_categorical] > 0.0
            ).type(torch.float64)
```

kAIto47802 (approved):
Thank you for the PR! Basically LGTM, leaving a minor comment.

optuna/_gp/gp.py
nabenabe0928: I was thinking of it for another PR, but we can do it now:)

y0z (approved):
LGTM

PR #6254, opened 2025-08-18
Include legend within image bounds for optimization history plot

fusawa-yugo:
## Motivation

Legend gets cut off when using `optuna.visualization.matplotlib.plot_optimization_history`.

## Description of the changes

The issue was that the legend fell outside the figure boundaries. I fixed this by adding `tight_layout=True`.
This doesn't appear to exist in other plot functions.
I also edited docs related to the change.

### DIFF
--- a/optuna/visualization/matplotlib/_optimization_history.py
+++ b/optuna/visualization/matplotlib/_optimization_history.py
@@ -72,7 +72,7 @@ def _get_optimization_history_plot(
 ) -> "Axes":
     # Set up the graph style.
     plt.style.use("ggplot")  # Use ggplot style sheet for similar outputs to plotly.
-    _, ax = plt.subplots()
+    _, ax = plt.subplots(tight_layout=True)
     ax.set_title("Optimization History Plot")
     ax.set_xlabel("Trial")
     ax.set_ylabel(target_name)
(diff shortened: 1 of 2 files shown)


fusawa-yugo:
<details>
<summary>Test Code</summary>

```python
import optuna

def objective(trial):
    y = trial.suggest_categorical("y", [-1, 0, 1])
    return x**2 + y

optuna.visualization.matplotlib.plot_optimization_history(study)

**Comment by y0z** [comment]
@gen740 @nabenabe0928 Could you review this PR?

File: optuna/visualization/matplotlib/_optimization_history.py [hunk 1]
```diff
@@ -72,7 +68,7 @@ def _get_optimization_history_plot(
 ) -> "Axes":
     # Set up the graph style.
     plt.style.use("ggplot")  # Use ggplot style sheet for similar outputs to plotly.
-    _, ax = plt.subplots()
+    _, ax = plt.subplots(tight_layout=True)
```
nabenabe0928, line 71: Only providing `tight_layout=True` is not sufficient for Ubuntu to make the figure look prettier.
To do so, you need to calibrate the legend location.

this_pr


For example, see below:

example

gen740: I think it’s fine since the output of the figure looks the same as the Plotly version.

newplot

nabenabe0928: That's right. (IMO, the plotly version should be modified though)

nabenabe0928 (approved):
LGTM!

PR #6257, opened 2025-08-21
Speed up EHVI by optimizing tensor operation order

nabenabe0928:
## Motivation

This PR enhances the constant factor of EHVI.
By this PR, the operation on `(B, S, N, K, D)` will be reduced.

## Description of the changes

- Change the order of the torch tensor operations

### DIFF
--- a/optuna/_gp/acqf.py
+++ b/optuna/_gp/acqf.py
@@ -56,7 +56,7 @@ def logehvi(
     # https://github.com/pytorch/botorch/blob/v0.13.0/botorch/utils/safe_math.py
     # https://github.com/pytorch/botorch/blob/v0.13.0/botorch/acquisition/multi_objective/logei.py#L146-L266
     diff = Y_post.unsqueeze(-2) - non_dominated_box_lower_bounds
-    diff.clamp_(min=torch.tensor(_EPS).double(), max=non_dominated_box_intervals)
+    diff.clamp_(min=torch.tensor(_EPS, dtype=torch.float64), max=non_dominated_box_intervals)
     # NOTE(nabenabe): logsumexp with dim=-1 is for the HVI calculation and that with dim=-2 is for
     # expectation of the HVIs over the fixed_samples.
     return torch.special.logsumexp(diff.log().sum(dim=-1), dim=(-2, -1)) - log_n_qmc_samples


c-bata:
@gen740 @not522 Could you review this PR?

optuna/_gp/acqf.py
```diff
@@ -49,11 +49,8 @@ def logehvi(
     # https://github.com/pytorch/botorch/blob/v0.13.0/botorch/utils/safe_math.py
     # https://github.com/pytorch/botorch/blob/v0.13.0/botorch/acquisition/multi_objective/logei.py#L146-L266
     _EPS = torch.tensor(1e-12, dtype=torch.float64)  # NOTE(nabenabe): grad becomes nan when EPS=0.
-    diff = torch.maximum(
-        _EPS,
-        torch.minimum(Y_post[..., None, :], non_dominated_box_upper_bounds)
-        - non_dominated_box_lower_bounds,
-    )
+    diff = Y_post.unsqueeze(-2) - non_dominated_box_lower_bounds
+    diff.clamp_(min=_EPS, max=non_dominated_box_upper_bounds - non_dominated_box_lower_bounds)
```
not522: When `non_dominated_box_upper_bounds - non_dominated_box_lower_bounds` is smaller than `_EPS`, the result will differ from master.

https://docs.pytorch.org/docs/stable/generated/torch.clamp.html

```suggestion
    diff.clamp_(min=_EPS, max=torch.maximum(non_dominated_box_upper_bounds - non_dominated_box_lower_bounds, _EPS))
```

nabenabe0928: Thank you for the comment, I fixed this part accordingly!

optuna/_gp/acqf.py
```diff
@@ -7,26 +7,22 @@
 from typing import TYPE_CHECKING
 
 import numpy as np
+import torch
```
not522: Optuna doesn't install PyTorch by default. How about using `_LazyImport` or `try_import`?

optuna/_gp/acqf.py
```diff
@@ -54,12 +55,8 @@ def logehvi(
...
     # https://github.com/pytorch/botorch/blob/v0.13.0/botorch/utils/safe_math.py
     # https://github.com/pytorch/botorch/blob/v0.13.0/botorch/acquisition/multi_objective/logei.py#L146-L266
-    _EPS = torch.tensor(1e-12, dtype=torch.float64)  # NOTE(nabenabe): grad becomes nan when EPS=0.
-    diff = torch.maximum(
-        _EPS,
-        torch.minimum(Y_post[..., None, :], non_dominated_box_upper_bounds)
-        - non_dominated_box_lower_bounds,
-    )
+    diff = Y_post.unsqueeze(-2) - non_dominated_box_lower_bounds
+    diff.clamp_(min=torch.tensor(_EPS).double(), max=non_dominated_box_intervals)
```
not522: There is a small error.

```suggestion
    diff.clamp_(min=torch.tensor(_EPS, dtype=torch.float64), max=non_dominated_box_intervals)
```

nabenabe0928: Isn't it indeed the same thing?
I am very flexible regarding here, but I asked the question out of my curiosity!

not522: They are different. It's a very trivial difference, but since we can maintain the original behavior, I prefer `dtype=torch.float64`.
```
>>> torch.tensor(1e-12).double().numpy()
array(9.99999996e-13)
>>> torch.tensor(1e-12, dtype=torch.float64).numpy()
array(1.e-12)
```

nabenabe0928: You are right.
I observed that `torch.tensor` first casts `1e-12` to `float32`, which halves the original precision.
Thank you for pointing it out!

not522 (approved):
LGTM!

nabenabe0928:

kAIto47802 (approved):
LGTM!

PR #6258, opened 2025-08-25
Fix log PDF of discrete trunc log-norm distribution for `TPESampler`

not522:
## Motivation
`TPESampler` exhibits inconsistent handling of `suggest_int(..., log=True)`. While the Parzen estimator uses the log-normal distribution's interval during sampling, it employs the log-normal distribution's value at the sampled point for the log PDF. This PR fixes the behavior to use the log-normal distribution's interval for the log PDF as well.

## Description of the changes
- The first commit (https://github.com/optuna/optuna/pull/6258/commits/4ede7000c897772afb618667f7bf6f00d83cb785) introduced `_BatchedTruncLogNormDistributions` and `_BatchedDiscreteTruncLogNormDistributions` to clarify the behavior of each distribution. It is a refactoring and does not change the behavior.
- The second commit (https://github.com/optuna/optuna/pull/6258/commits/a5ab97c7ba93387bfd7831e6e50a52380a111879) fixed the log PDF of `_BatchedDiscreteTruncLogNormDistributions`.

Using the code below, I confirmed that the optimization performance remained nearly unchanged.

```python
import numpy as np
import optuna

def objective(trial):
    params = np.empty(5)
    for i in range(5):
        params[i] = trial.suggest_int(f"x{i}", 1, 100, log=True)
    return np.sum((np.log(params) - np.log(10)) ** 2)

N_STUDIES = 100
N_TRIALS = 100

y = [[] for _ in  range(N_STUDIES)]
for seed in range(N_STUDIES):
    sampler = optuna.samplers.TPESampler(seed=seed)
    study.optimize(objective, n_trials=N_TRIALS)
    y[seed].append(study.trials[0].value)
    for i in range(1, N_TRIALS):
        y[seed].append(min(y[seed][-1], study.trials[i].value))

y = np.asarray(y)
np.save("master.npy", y)
# np.save("pr.npy", y)
```
```python
y = np.load("master.npy")
mean = np.mean(y, axis=0)
std = np.std(y, axis=0)
plt.fill_between(range(N_TRIALS), mean + std, mean - std, alpha=0.1)
plt.plot(range(N_TRIALS), mean, label="master")

y = np.load("pr.npy")
mean = np.mean(y, axis=0)
std = np.std(y, axis=0)
plt.fill_between(range(N_TRIALS), mean + std, mean - std, alpha=0.1)
plt.plot(range(N_TRIALS), mean, label="fixed")

plt.xlabel("trial number")
plt.ylabel("loss")
plt.legend()
plt.savefig("logint.png")
```

logint

### DIFF
--- a/optuna/samplers/_tpe/probability_distributions.py
+++ b/optuna/samplers/_tpe/probability_distributions.py
@@ -93,8 +93,7 @@ class _MixtureOfProductDistribution(NamedTuple):
             | _BatchedDiscreteTruncNormDistributions
             | _BatchedDiscreteTruncLogNormDistributions
         ] = []
-        lows_cont = []
-        highs_cont = []
+        lows_numeric, highs_numeric = [], []
         for i, d in enumerate(self.distributions):
             if isinstance(d, _BatchedCategoricalDistributions):
                 active_weights = d.weights[active_indices, :]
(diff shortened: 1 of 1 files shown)


not522:
After examining the behavior of `TPESampler`, I discovered that `suggest_int(..., log=True)` exhibits a bias toward sampling smaller values. To properly address this issue, I will temporarily convert this PR to draft status.

not522:
This PR introduces the following behavioral changes, so I've confirmed that results may vary depending on the objective function.

- The original implementation tends to sample larger values.
- This PR makes sigmas greater.

While there may be room for discussion about how probability distributions should be determined, I think this PR can be treated as bug fixes.

====

```python
def objective(trial):
    params = np.empty(5)
    for i in range(5):
        params[i] = trial.suggest_int(f"x{i}", 1, 100, log=True)
    return np.sum((np.log(params) - np.log(1)) ** 2)
```
logint1

```python
def objective(trial):
    params = np.empty(20)
    for i in range(20):
        params[i] = trial.suggest_int(f"x{i}", 1, 100, log=True)
    return np.sum((np.log(params) - np.log(10)) ** 2)
```
logint20

optuna/samplers/_tpe/probability_distributions.py
```diff
@@ -69,10 +86,15 @@ class _MixtureOfProductDistribution(NamedTuple):
...
+        disc_inds, numerical_inds, log_inds = [], [], []
         numerical_dists: list[
-            _BatchedTruncNormDistributions | _BatchedDiscreteTruncNormDistributions
+            _BatchedTruncNormDistributions
+            | _BatchedTruncLogNormDistributions
+            | _BatchedDiscreteTruncNormDistributions
+            | _BatchedDiscreteTruncLogNormDistributions
         ] = []
+        lows_cont = []
+        highs_cont = []
```
kAIto47802:
```suggestion
        lows_num, highs_num = [], []
```
How about renaming it to "num", since naming "cont" for both continuous and discrete distributions is a bit misleading.

kAIto47802 (commented):
Thank you for the PR!
It's almost LGTM, leaving only a minor comment.

optuna/samplers/_tpe/probability_distributions.py
not522: How about `numeric`? (`num` might be confusing with `number`)

kAIto47802: Thank you for your response.
That's right. `numeric` seems to be better.
```suggestion
        lows_number, highs_number = [], []
```

not522: Thank you. I've updated it.

not522:

y0z (approved):
LGTM

PR #6261, opened 2025-08-28
Remove `prior_mu` from `compute_sigmas`

not522:
## Motivation
The calculation of sigmas in `ParzenEstimator` is influenced by the prior's mu, but this is inappropriate for the window computation. This behavior stems from referencing the implementation of hyperopt (https://github.com/hyperopt/hyperopt/commit/5d8b88e65713314bdba788064c3be04cc4401189).

## Description of the changes
This PR removes `prior_mu` from `compute_sigmas`.

Benchmark results vary depending on the objective function, and when the optimal values are in the middle of the suggestion ranges, performance slightly deteriorated.

```python
import numpy as np
import optuna

def objective(trial):
    params = np.empty(20)
    for i in range(20):
        params[i] = trial.suggest_float(f"x{i}", -10, 10)
    return np.sum(params ** 2)
    # return np.sum((params - 5) ** 2)

N_STUDIES = 100
N_TRIALS = 100

y = [[] for _ in  range(N_STUDIES)]
for seed in range(N_STUDIES):
    sampler = optuna.samplers.TPESampler(seed=seed)
    study.optimize(objective, n_trials=N_TRIALS)
    trials = study.trials
    y[seed].append(trials[0].value)
    for i in range(1, N_TRIALS):
        y[seed].append(min(y[seed][-1], trials[i].value))

y = np.asarray(y)
np.save("master.npy", y)
# np.save("pr.npy", y)
```
```python
y = np.load("master.npy")
mean = np.mean(y, axis=0)
std = np.std(y, axis=0)
plt.fill_between(range(N_TRIALS), mean + std, mean - std, alpha=0.1)
plt.plot(range(N_TRIALS), mean, label="master")

y = np.load("pr.npy")
mean = np.mean(y, axis=0)
std = np.std(y, axis=0)
plt.fill_between(range(N_TRIALS), mean + std, mean - std, alpha=0.1)
plt.plot(range(N_TRIALS), mean, label="fixed")

plt.xlabel("trial number")
plt.ylabel("loss")
plt.legend()
plt.savefig("sigma.png")
```

Σx^2 / Σ(x-5)^2

sigma0

sigma5

The internal state of the `ParzenEstimator` was verified using the following code:

```python
import numpy as np
import optuna
import optunahub

def objective(trial):
    params = np.empty(20)
    for i in range(20):
        params[i] = trial.suggest_float(f"x{i}", -10, 10)
    return np.sum(params ** 2)

module = optunahub.load_module(
    package="visualization/tpe_acquisition_visualizer",
)
tpe_acquisition_visualizer = module.TPEAcquisitionVisualizer()
sampler = optuna.samplers.TPESampler(seed=0)
study.optimize(objective, n_trials=11, callbacks=[tpe_acquisition_visualizer])
fig = tpe_acquisition_visualizer.plot(study, 10, "x0")
fig.savefig(f"x0_10.png")
plt.close(fig)
```

master / PR

x0_10_master

x0_10_pr

### DIFF
--- a/tests/samplers_tests/tpe_tests/test_parzen_estimator.py
+++ b/tests/samplers_tests/tpe_tests/test_parzen_estimator.py
@@ -68,33 +68,33 @@ def test_init_parzen_estimator(multivariate: bool) -> None:
         distributions=[
             _BatchedTruncNormDistributions(
                 mu=np.array([1.0, 50.5]),
-                sigma=np.array([49.5, 99.0]),
+                sigma=np.array([99.0, 99.0]),
                 low=1.0,
                 high=100.0,
             ),
             _BatchedTruncNormDistributions(
                 mu=np.array([np.log(1.0), np.log(100) / 2.0]),
-                sigma=np.array([np.log(100) / 2, np.log(100)]),
+                sigma=np.array([np.log(100), np.log(100)]),
                 low=np.log(1.0),
                 high=np.log(100.0),
             ),
             _BatchedDiscreteTruncNormDistributions(
                 mu=np.array([1.0, 50.5]),
-                sigma=np.array([49.5, 102.0]),
+                sigma=np.array([100.5, 102.0]),
                 low=1.0,
                 high=100.0,
                 step=3.0,
             ),
             _BatchedDiscreteTruncNormDistributions(
                 mu=np.array([1.0, 50.5]),
-                sigma=np.array([49.5, 100.0]),
+                sigma=np.array([99.5, 100.0]),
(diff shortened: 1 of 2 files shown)


y0z:
@nabenabe0928 @fusawa-yugo 
Could you review this PR?

nabenabe0928:
I'd recommend revisiting this PR once the bandwidth selection becomes stable.

not522:
I discussed with @y0z as well, we concluded that performance concerns warrant postponing this change until the next major release.

PR #6268, opened 2025-09-08
Speed up `GPSampler` by Batching Acquisition Function Evaluations

Kaichi-Irie:
## Motivation

The multi-start optimization of the acquisition function in the `GPSampler` is a significant performance bottleneck, accounting for 50-60% of its total execution time. The current implementation uses 10 initial points (`n_local_search=10`) and performs optimization from each point sequentially.

This pull request aims to significantly speed up this multi-start optimization by batching the evaluation of the acquisition function and its gradient. Benchmark results show that this change can reduce the execution time by **approximately 40-50%** without compromising the optimization accuracy of the objective function.

### Previous Sequential Approach

The previous implementation runs an optimization loop for each starting point individually.

```python
# Previous pseudo-code
def multistart_optimize(acqf, x0_list, ...):
    ...
    min_fvals = ...
    min_xs = ...
    for i, x0 in enumerate(x0_list):
        # optimize() performs iterative optimization for a single initial point x0
        min_fval, min_x = optimize(acqf, x0, ...)
        min_fvals[i] = min_fval
        min_xs[i] = min_x
    best_idx = np.argmin(min_fvals)
    return min_xs[best_idx], min_fvals[best_idx]
```

### Proposed Batched Approach

The proposed approach executes each optimization step for all initial points in a single batch.

Evaluations of the acquisition function (`acqf`) and its gradient (`acqf_grad`) are performed at once using vectorized operations on `torch.Tensor`. The iterative optimization from each starting point is managed cooperatively using `greenlet`. This enables a cooperative, iterative workflow. In each iteration, the function values for all starting points are evaluated as a single batch. Subsequently, each optimization is advanced by one step to determine the next set of points to be evaluated, which then forms the input for the following batch evaluation cycle.

```python
# New pseudo-code
def multistart_optimize(acqf, x0_list, ...):
    ...
    # optimize_batched manages optimization for multiple initial points x0_list
    min_xs, min_fvals = optimize_batched(acqf, x0_list, ...)
    best_idx = np.argmin(min_fvals)
    return min_xs[best_idx], min_fvals[best_idx]
```

For backward compatibility, a fallback mechanism has been implemented. If `greenlet` fails to import, the logic will revert to the original sequential execution.

### Dependencies

This change utilizes `greenlet`, which is already included as a dependency of `sqlalchemy`, a core Optuna dependency. Therefore, this PR does not introduce any new library dependencies.

### Performance Evaluation (Benchmark)

To validate the effectiveness of the proposed method, we compared the performance in the following three modes:

1. `This PR (Batched)`: The implementation in this PR with batch evaluation enabled.
2. `original`: The current implementation on the `master` branch.
3. `This PR (Fallback)`: The implementation in this PR with batch evaluation intentionally disabled to activate the sequential fallback logic. This helps confirm that the new implementation's overhead is acceptable and that the implementation is functioning correctly.

#### Experimental Setup

  * **CPU:** Apple M4 (10-core)
  * **Trials:** 300 trials per study
  * **Seed:** 42, 43, 44 (results are averaged over three runs)
  * **Objective Functions:**
      * `rotated_ellipsoid` (10 dimensions)
      * `f6` from `optunahub/benchmarks/bbob` (20 dimensions)
      * `wfg` (2 objectives, 3 dimensions, k=1) from `optunahub/benchmarks/wfg`

#### Results

| Objective Function (Dimension) | Mode                  | Avg. Execution Time (s) | Avg. Best Objective Value |
| :----------------------------- | :-------------------- | ----------------------: | ------------------------: |
| `f6` (D=20)                    | **This PR (Batched)** |                **19.0** |                     133.2 |
| `f6` (D=20)                    | `original`            |                    35.7 |                     155.9 |
| `f6` (D=20)                    | `This PR (Fallback)`  |                    31.9 |                     159.2 |
| `rotated_ellipsoid` (D=10)     | **This PR (Batched)** |                **19.4** |                   3.64e-4 |
| `rotated_ellipsoid` (D=10)     | `original`            |                    39.6 |                   3.75e-4 |
| `rotated_ellipsoid` (D=10)     | `This PR (Fallback)`  |                    34.0 |                   2.28e-4 |
| `wfg` (D=3)                    | **This PR (Batched)** |                **35.3** |                         - |
| `wfg` (D=3)                    | `original`            |                    74.4 |                         - |
| `wfg` (D=3)                    | `This PR (Fallback)`  |                    66.9 |                         - |

speed_comparison_plot

*The following plot show that the hypervolume values remain consistent across all modes, confirming that the change does not negatively impact search performance.*

hypervolume_history

#### Analysis

  * The implementation with batched evaluation is **approximately 1.8x to 2.0x faster** than the current implementation.
  * The best objective values obtained are comparable across all modes. The observed variations are likely due to changes in the order of floating-point operations and do not indicate a degradation in the sampler's search performance.
  * The fallback mode is as fast as the original implementation, suggesting that the structural overhead of the proposed changes is minimal.

##### `f6` (D=20, seed=42)

*The following tables show that the objective values remain consistent across all modes, confirming that the change does not negatively impact search performance.*

| mode \ Trial           | Trial 9            |          Trial 10 |           Trial 11 | Trial 12           | Trial 13          | Trial 14          |
| :--------------------- | :----------------- | ----------------: | -----------------: | ------------------ | ----------------- | ----------------- |
| **This PR (Batched)**  | 1193964.5576430012 | 435258.7193614113 |  429479.3338986842 | 328199.63425374887 | 541828.1312516633 | 571560.8492535071 |
| **This PR (Fallback)** | 1193964.5576430012 | 435258.7193614113 | 429479.33389868634 | 328199.63425374887 | 541828.1312516204 | 571560.8492535099 |
| **original**           | 1193964.5576430012 | 435258.7193614113 |  429479.3338986842 | 328199.6342537497  | 541828.1312516493 | 571560.8492535143 |

##### `rotated_ellipsoid` (D=10, seed=42)

| mode \ Trial           | Trial 9            |          Trial 10 |          Trial 11 | Trial 12           | Trial 13           | Trial 14           |
| :--------------------- | :----------------- | ----------------: | ----------------: | ------------------ | ------------------ | ------------------ |
| **This PR (Batched)**  | 12.130935341373624 | 6.497476295603841 | 9.159600880097619 | 11.184451388408986 | 13.400053021276653 | 10.290407711547488 |
| **This PR (Fallback)** | 12.130935341373624 | 6.497476295603834 | 9.159600880097601 | 11.184451388408993 | 13.400053021279028 | 10.290407711541484 |
| **original**           | 12.130935341373624 | 6.497476295603843 | 9.159600880097617 | 11.18445138840904  | 13.40005302128083  | 10.290407711537135 |

<details>

```python

import json
import os
from itertools import product

import numpy as np
import optunahub

import optuna

np.random.seed(42)

NUM_CONTINUOUS_PARAMS = 10
ROTATION_MATRIX, _ = np.linalg.qr(np.random.rand(NUM_CONTINUOUS_PARAMS, NUM_CONTINUOUS_PARAMS))

CONDITIONING = 10
WEIGHTS = np.array([CONDITIONING**(i / (NUM_CONTINUOUS_PARAMS - 1)) for i in range(NUM_CONTINUOUS_PARAMS)])

def rotated_ellipsoid(trial):
    xs = np.array([trial.suggest_float(f"x_{i}", -1, 1) for i in range(NUM_CONTINUOUS_PARAMS)])
    rotated_xs = ROTATION_MATRIX @ xs
    return np.sum(WEIGHTS * (rotated_xs**2))

bbob = optunahub.load_module("benchmarks/bbob")
f6 = bbob.Problem(function_id=6, dimension=20, instance_id=1)

def execute_benchmark(
    mode: str,
    objective_type: str,
    n_trials: int,
    seed: int,
    results_file="results.jsonl",
    output_dir="output",
):
    sampler = optuna.samplers.GPSampler(seed=seed)
    name = f"{objective_type}_{seed}_{mode}_{n_trials}"
    log_file = f"{name}.jsonl"
    study = optuna.create_study(study_name=name, sampler=sampler)

    study.optimize(func=f6 if objective_type == "f6" else rotated_ellipsoid, n_trials=n_trials)
    print(study.best_trial.params, study.best_trial.value)
    elapsed = (study.trials[-1].datetime_complete - study.trials[0].datetime_start).total_seconds() # type: ignore
    print(f"{mode} took {elapsed:f} seconds. ")

    result = {
        "objective_type": objective_type,
        "seed": seed,
        "mode": mode,
        "elapsed": round(elapsed, 2),
        "n_trials": n_trials,
        "best_value": study.best_trial.value,
    }

    with open(os.path.join(output_dir, results_file), "a") as f:
        f.write(json.dumps(result) + "\n")

seeds = [42,43,44]
n_trials = 300

# To test different modes, you need to edit the source code of the GPSampler directly.
# This script runs with the batched evaluation enabled by default.
modes: list[str] = [
    "This PR (Batched)",
    # "This PR (Fallback)",
    # "original",
]

objective_types = [
    "rotated_ellipsoid",
    "f6",
]

for seed, mode, objective_type in product(seeds, modes, objective_types):
    execute_benchmark(
        mode=mode,
        objective_type=objective_type,
        n_trials=n_trials,
        seed=seed,
    )

```

</details>

### Tests
Unit tests must be added.

### DIFF
--- a/optuna/_gp/batched_lbfgsb.py
+++ b/optuna/_gp/batched_lbfgsb.py
@@ -13,6 +13,7 @@ with try_import() as _greenlet_imports:
 
 if TYPE_CHECKING:
     from collections.abc import Callable
+
     import scipy.optimize as so
 
 else:
@@ -30,7 +31,7 @@ if not _greenlet_imports.is_successful():
 
 
 def _batched_lbfgsb(
-    func_and_grad: Callable[[np.ndarray], tuple[np.ndarray, np.ndarray]],
+    func_and_grad: Callable[[np.ndarray, np.ndarray], tuple[np.ndarray, np.ndarray]],
     x0_batched: np.ndarray,
     bounds: list[tuple[float, float]] | None,
     m: int,
@@ -47,6 +48,7 @@ def _batched_lbfgsb(
     xs_opt = np.empty_like(x0_batched)
     fvals_opt = np.empty(batch_size, dtype=float)
     n_iterations = np.empty(batch_size, dtype=int)
+    is_remaining_batch = np.ones(batch_size, dtype=bool)
 
     def run(i: int) -> None:
         def _func_and_grad(x: np.ndarray) -> tuple[float, np.ndarray]:
@@ -68,19 +70,29 @@ def _batched_lbfgsb(
         xs_opt[i] = x_opt
         fvals_opt[i] = fval_opt
         n_iterations[i] = info["nit"]
+        is_remaining_batch[i] = False
 
     greenlets = [greenlet(run) for _ in range(batch_size)]
     x_batched = [gl.switch(i) for i, gl in enumerate(greenlets)]
-    while len(x_batched := [x for x in x_batched if x is not None]) > 0:
-        fvals, grads = func_and_grad(np.asarray(x_batched))
-        x_batched = [gl.switch((fvals[i], grads[i])) for i, gl in enumerate(greenlets)]
-        greenlets = [gl for x, gl in zip(x_batched, greenlets) if x is not None]
 
+    while np.any(is_remaining_batch):
+        remaining_batch_indices = np.where(is_remaining_batch)[0]
+        fvals, grads = func_and_grad(np.asarray(x_batched), np.asarray(remaining_batch_indices))
+
+        x_batched = []
+        next_greenlets = []
+        for i, gl in enumerate(greenlets):
+            x = gl.switch((fvals[i], grads[i]))
+            if x is None:
+                continue
+            x_batched.append(x)
+            next_greenlets.append(gl)
+        greenlets = next_greenlets
     return xs_opt, fvals_opt, n_iterations
 
 
 def batched_lbfgsb(
-    func_and_grad: Callable[[np.ndarray], tuple[np.ndarray, np.ndarray]],
+    func_and_grad: Callable[[np.ndarray, np.ndarray], tuple[np.ndarray, np.ndarray]],
     x0_batched: np.ndarray,
     bounds: list[tuple[float, float]] | None = None,
     m: int = 10,
@@ -111,8 +123,10 @@ def batched_lbfgsb(
         fvals_opt = np.empty(x0_batched.shape[0], dtype=float)
         n_iterations = np.empty(x0_batched.shape[0], dtype=int)
         for i, x0 in enumerate(x0_batched):
+            batch_indices = np.array([i])
             xs_opt[i], fvals_opt[i], info = so.fmin_l_bfgs_b(
                 func=func_and_grad,  # type: ignore[arg-type]
+                args=(batch_indices,),  # type: ignore[arg-type]
                 x0=x0,
                 bounds=bounds,
                 m=m,
--- a/optuna/_gp/optim_mixed.py
+++ b/optuna/_gp/optim_mixed.py
@@ -53,19 +53,24 @@ def _gradient_ascent_batched(
         return initial_params_batched, initial_fvals, np.zeros(len(initial_fvals), dtype=bool)
     normalized_params = initial_params_batched.copy()
 
-    def negative_acqf_with_grad(scaled_x: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
+    def negative_acqf_with_grad(
+        scaled_x: np.ndarray, batch_indices: np.ndarray
+    ) -> tuple[np.ndarray, np.ndarray]:
         # Scale back to the original domain, i.e. [0, 1], from [0, 1/s].
         if scaled_x.ndim == 1:
-            scaled_x = scaled_x[None, :]
-        normalized_params[: len(scaled_x), continuous_indices] = scaled_x * lengthscales
-        x_tensor = torch.from_numpy(normalized_params[: len(scaled_x), :]).requires_grad_(True)
+            # NOTE(Kaichi-Irie): When scaled_x is 1D, regard it as a single batch.
+            scaled_x = scaled_x[None]
+        assert scaled_x.ndim == 2
+        normalized_params[np.ix_(batch_indices, continuous_indices)] = scaled_x * lengthscales
+        # NOTE(Kaichi-Irie): If fvals.numel() > 1, backward() cannot be computed, so we sum up.
+        x_tensor = torch.from_numpy(normalized_params[batch_indices]).requires_grad_(True)
         neg_fvals = -acqf.eval_acqf(x_tensor)
         neg_fvals.sum().backward()
         grads = x_tensor.grad.detach().numpy()  # type: ignore
         neg_fvals = neg_fvals.detach().numpy()
         # Flip sign because scipy minimizes functions.
         # Let the scaled acqf be g(x) and the acqf be f(sx), then dg/dx = df/dx * s.
-        return neg_fvals.reshape(-1), grads[:, continuous_indices] * lengthscales
+        return np.atleast_1d(neg_fvals), grads[:, continuous_indices] * lengthscales
 
     with single_blas_thread_if_scipy_v1_15_or_newer():
         scaled_cont_xs_opt, neg_fvals_opt, n_iterations = batched_lbfgsb.batched_lbfgsb(


Kaichi-Irie:
@not522 @contramundum53 
May I request your review of this PR?

optuna/_gp/batched_lbfgsb.py
nabenabe0928, line 1: The implementation (and typing, naming) could look better:

<details>
<summary>Refactored Version</summary>

```python

import numpy as np

from optuna._imports import try_import
from optuna.logging import get_logger


    from greenlet import greenlet

if TYPE_CHECKING:
    from typing import Protocol

    import scipy.optimize as so

    class FuncAndGrad(Protocol):
        def __call__(self, x: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
            raise NotImplementedError

else:
    from optuna import _LazyImport

    so = _LazyImport("scipy.optimize")


if _imports.is_successful():
    _logger.warning(
        "The 'greenlet' package is unavailable, falling back to sequential L-BFGS-B optimization. "
        "This may lead to slower suggestions."
    )


def _batched_lbfgsb(
    func_and_grad: FuncAndGrad,
    x0_batched: np.ndarray,
    bounds: list[tuple[float, float]] | None,
    m: int,
    factr: float,
    pgtol: float,
    max_evals: int,
    max_iters: int,
    max_line_search: int,
) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
    assert (
        x0_batched.ndim == 2
    ), f"The shape of x0 must be (batch_size, dim), but got {x0_batched.shape}."
    batch_size = len(x0_batched)
    xs_opt = np.empty_like(x0_batched)
    fvals_opt = np.empty(batch_size, dtype=float)
    n_iterations = np.empty(batch_size, dtype=int)

    def run(i: int) -> None:
        def _func_and_grad(x: np.ndarray) -> tuple[float, np.ndarray]:
            fval, grad = greenlet.getcurrent().parent.switch(x)
            # NOTE(nabenabe): copy is necessary to convert grad to writable.
            return float(fval), grad.copy()

        x_opt, fval_opt, info = so.fmin_l_bfgs_b(
            func=_func_and_grad,
            x0=x0_batched[i],
            bounds=bounds,
            m=m,
            factr=factr,
            pgtol=pgtol,
            maxfun=max_evals,
            maxiter=max_iters,
            maxls=max_line_search,
        )
        xs_opt[i] = x_opt
        fvals_opt[i] = fval_opt
        n_iterations[i] = info["nit"]

    greenlets = [greenlet(run) for _ in range(batch_size)]
    x_batched = [gl.switch(i) for i, gl in enumerate(greenlets)]
    while len(x_batched := [x for x in x_batched if x is not None]) > 0:
        fvals, grads = func_and_grad(np.asarray(x_batched))
        x_batched = [gl.switch((fvals[i], grads[i])) for i, gl in enumerate(greenlets)]
        greenlets = [gl for x, gl in zip(x_batched, greenlets) if x is not None]

    return xs_opt, fvals_opt, n_iterations


def batched_lbfgsb(
    func_and_grad: FuncAndGrad,
    x0_batched: np.ndarray,
    bounds: list[tuple[float, float]] | None = None,
    m: int = 10,
    factr: float = 1e7,
    pgtol: float = 1e-5,
    max_evals: int = 15000,
    max_iters: int = 15000,
    max_line_search: int = 20,
) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
    def func_and_grad_1D_wrapper(scaled_x: np.ndarray) -> tuple[float, np.ndarray]:
        """A wrapper for `func_and_grad` to handle 1D inputs.

        This is used as a fallback to sequential optimization when the batched
        L-BFGS-B is not available. It adapts the batched `func_and_grad` for
        use with optimizers like `scipy.optimize.fmin_l_bfgs_b` that expect
        a 1D input array.
        """
        assert scaled_x.ndim == 1
        fval, grad = func_and_grad(scaled_x[None])
        return fval.item(), grad.ravel()

    if _imports.is_successful() and len(x0_batched) > 1:
        xs_opt, fvals_opt, n_iterations = _batched_lbfgsb(
            func_and_grad=func_and_grad,
            x0_batched=x0_batched,
            bounds=bounds,
            m=m,
            factr=factr,
            pgtol=pgtol,
            max_evals=max_evals,
            max_iters=max_iters,
            max_line_search=max_line_search,
        )

    # fall back to sequential optimization if greenlet is not available.
    else:
        xs_opt = np.empty_like(x0_batched)
        fvals_opt = np.empty(x0_batched.shape[0])
        n_iterations = np.empty(x0_batched.shape[0], dtype=int)
        for i, x0 in enumerate(x0_batched):
            xs_opt[i], fvals_opt[i], info = so.fmin_l_bfgs_b(
                func=func_and_grad_1D_wrapper,
                x0=x0,
                bounds=bounds,
                m=m,
                factr=factr,
                pgtol=pgtol,
                maxfun=max_evals,
                maxiter=max_iters,
                maxls=max_line_search,
            )
            n_iterations[i] = info["nit"]

    return xs_opt, fvals_opt, n_iterations

```

</details>

nabenabe0928: IMO, the function name should be `lbfgsb` and accept both batched and non-batched functions, meaning that `func_and_grad` should return func and grads with the shapes of `x0.shape[:-1]` and `x0.shape`, respectively.
Namely, I prefer to not see `func_and_grad_1D_wrapper` here.


```python
def batched_lbfgsb(
    func_and_grad: FuncAndGrad,
    x0_batched: np.ndarray,
    bounds: list[tuple[float, float]] | None = None,
    m: int = 10,
    factr: float = 1e7,
    pgtol: float = 1e-5,
    max_evals: int = 15000,
    max_iters: int = 15000,
    max_line_search: int = 20,
) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
    def func_and_grad_1D_wrapper(scaled_x: np.ndarray) -> tuple[float, np.ndarray]:
        """A wrapper for `func_and_grad` to handle 1D inputs.

        This is used as a fallback to sequential optimization when the batched
        L-BFGS-B is not available. It adapts the batched `func_and_grad` for
        use with optimizers like `scipy.optimize.fmin_l_bfgs_b` that expect
        a 1D input array.
        """
        assert scaled_x.ndim == 1
        fval, grad = func_and_grad(scaled_x[None])
        return fval.item(), grad.ravel()

```

optuna/_gp/optim_mixed.py
```diff
@@ -45,32 +50,41 @@ def _gradient_ascent(
...
-    def negative_acqf_with_grad(scaled_x: np.ndarray) -> tuple[float, np.ndarray]:
+    def negative_acqf_with_grad(scaled_x: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
         # Scale back to the original domain, i.e. [0, 1], from [0, 1/s].
-        normalized_params[continuous_indices] = scaled_x * lengthscales
-        (fval, grad) = acqf.eval_acqf_with_grad(normalized_params)
+        assert scaled_x.ndim == 2
+        normalized_params[: len(scaled_x), continuous_indices] = scaled_x * lengthscales
+        x_tensor = torch.from_numpy(normalized_params[: len(scaled_x), :]).requires_grad_(True)
+        neg_fvals = -acqf.eval_acqf(x_tensor)
+        neg_fvals.sum().backward()
```
nabenabe0928: Do we have to limit `x.ndim` to 2?

```suggestion
        assert scaled_x.ndim <= 2
        normalized_params[: len(scaled_x), continuous_indices] = scaled_x * lengthscales
        x_tensor = torch.from_numpy(normalized_params[: len(scaled_x), :]).requires_grad_(True)
        neg_fvals = -acqf.eval_acqf(x_tensor)
        neg_fvals.backward() if scaled_x.ndim == 1 else neg_fvals.sum().backward()
```

optuna/_gp/optim_mixed.py
```diff
@@ -45,32 +50,41 @@ def _gradient_ascent(
...
         # Let the scaled acqf be g(x) and the acqf be f(sx), then dg/dx = df/dx * s.
-        return -fval, -grad[continuous_indices] * lengthscales
+        return neg_fvals, grads[:, continuous_indices] * lengthscales
 
     with single_blas_thread_if_scipy_v1_15_or_newer():
-        scaled_cont_x_opt, neg_fval_opt, info = so.fmin_l_bfgs_b(
-            func=negative_acqf_with_grad,
-            x0=normalized_params[continuous_indices] / lengthscales,
+        scaled_cont_x_opts, neg_fval_opts, n_iterations = batched_lbfgsb.batched_lbfgsb(
+            func_and_grad=negative_acqf_with_grad,  # type: ignore
```
nabenabe0928: [nit] Can we modify the code so that we can eliminate the ignore?
```suggestion
            func_and_grad=negative_acqf_with_grad,
```

optuna/_gp/optim_mixed.py
```diff
@@ -45,32 +50,41 @@ def _gradient_ascent(
...
 
-    return initial_params, initial_fval, False  # No improvement.
+    # If any parameter is updated, return the updated parameters and values.
+    # Otherwise, return the initial ones.
+    is_updated_batch = (-neg_fval_opts > initial_fvals) & (n_iterations > 0)
+    final_params = initial_params_batched.copy()
+    final_params[is_updated_batch, :] = normalized_params[is_updated_batch, :]
+    final_fvals = initial_fvals.copy()
+    final_fvals[is_updated_batch] = -neg_fval_opts[is_updated_batch]
+    return final_params, final_fvals, is_updated_batch
```
nabenabe0928: [nit]
```suggestion
    return (
        np.where(is_updated_batch, normalized_params, initial_params_batched),
        np.where(is_updated_batch, -neg_fval_opts, initial_fvals),
        is_updated_batch,
    )
```

optuna/_gp/optim_mixed.py
```diff
@@ -45,32 +50,41 @@ def _gradient_ascent(
...
-    if -neg_fval_opt > initial_fval and info["nit"] > 0:  # Improved.
-        # `nit` is the number of iterations.
-        normalized_params[continuous_indices] = scaled_cont_x_opt * lengthscales
-        return normalized_params, -neg_fval_opt, True
+    normalized_params[:, continuous_indices] = scaled_cont_x_opts * lengthscales
 
-    return initial_params, initial_fval, False  # No improvement.
+    # If any parameter is updated, return the updated parameters and values.
+    # Otherwise, return the initial ones.
+    is_updated_batch = (-neg_fval_opts > initial_fvals) & (n_iterations > 0)
```
nabenabe0928: [nit] Shall we define `fvals_opt := -neg_fval_opts` here?

> [!NOTE]
> `neg_fval_opts` should be renamed to `neg_fvals_opt` as multiple optimal function values exist and `opt` here refers to `optimal`, which is an adjective.

optuna/_gp/optim_mixed.py
```diff
@@ -149,7 +163,7 @@ def interpolated_negative_acqf(x: float) -> float:
         method="brent",
         tol=xtol,
     )
-    opt_idx = find_nearest_index(res.x)
+    opt_idx = find_nearest_index(res.x)  # type: ignore
```
nabenabe0928: [Question] Is the ignore here necessary?

optuna/_gp/optim_mixed.py
```diff
@@ -180,13 +194,36 @@ def _local_search_discrete(
...
+    param_idx: int,
+    choices: np.ndarray,
+    xtol: float,
+) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+    best_normalized_params_batched = initial_params_batched.copy()
+    best_fvals = initial_fvals.copy()
+
+    is_updated_batch = np.zeros(len(initial_fvals), dtype=bool)
+    for batch, normalized_params in enumerate(initial_params_batched):
+        (best_normalized_params, best_fval, updated) = _local_search_discrete(
```
nabenabe0928: [nit]
```suggestion
        best_normalized_params, best_fval, updated = _local_search_discrete(
```

optuna/_gp/optim_mixed.py
```diff
@@ -180,13 +194,36 @@ def _local_search_discrete(
...
+            acqf, normalized_params, best_fvals[batch], param_idx, choices, xtol
+        )
+        best_normalized_params_batched[batch] = best_normalized_params
+        best_fvals[batch] = best_fval
+        is_updated_batch[batch] = updated
+
+    return best_normalized_params_batched, best_fvals, is_updated_batch
+
+
+def local_search_mixed_batched(
```
nabenabe0928, line 229: [nit] Does this function have to be named `batched`?

optuna/_gp/optim_mixed.py
```diff
@@ -206,45 +243,77 @@ def local_search_mixed(
...
 
-    best_normalized_params = initial_normalized_params.copy()
-    best_fval = float(acqf.eval_acqf_no_grad(best_normalized_params))
-
-    CONTINUOUS = -1
-    last_changed_param: int | None = None
+    best_normalized_params_batched = initial_normalized_params_batched.copy()
+    best_fvals = np.array(
+        [float(acqf.eval_acqf_no_grad(p)) for p in best_normalized_params_batched]
+    )
```
nabenabe0928: Can we simply call it once?
```suggestion
    best_fvals = acqf.eval_acqf_no_grad(best_normalized_params_batched)
```

optuna/_gp/optim_mixed.py
nabenabe0928: This function could be much more concise:

```python
def local_search_mixed(
    acqf: BaseAcquisitionFunc, xs0: np.ndarray, *, tol: float = 1e-4, max_iter: int = 100
) -> tuple[np.ndarray, np.ndarray]:
    # This is a technique for speeding up optimization. We use an isotropic kernel, so scaling the
    # gradient will make the hessian better-conditioned.
    # NOTE: Ideally, separating lengthscales should be used for the constraint functions,
    # but for simplicity, the ones from the objective function are being reused.
    # TODO(kAIto47802): Think of a better way to handle this.
    lengthscales = acqf.length_scales[cont_inds := acqf.search_space.continuous_indices]
    discrete_indices = acqf.search_space.discrete_indices
    choices_of_discrete_params = acqf.search_space.get_choices_of_discrete_params()
    discrete_xtols = [
        # Terminate discrete optimizations once the change in x becomes smaller than this.
        # Basically, if the change is smaller than min(dx) / 4, it is useless to see more details.
        np.min(np.diff(choices), initial=np.inf) / 4
        for choices in choices_of_discrete_params
    ]
    best_fvals = acqf.eval_acqf_no_grad(best_xs := xs0.copy())
    CONTINUOUS = -1
    last_changed_dims = np.full(len(best_xs), CONTINUOUS, dtype=int)
    remaining_inds = np.arange(len(best_xs))
    for _ in range(max_iter):
        xs0, fvals0 = best_xs[remaining_inds], best_fvals[remaining_inds]
        best_xs[remaining_inds], best_fvals[remaining_inds], updated = _gradient_ascent(
            acqf, best_xs[remaining_inds], best_fvals[remaining_inds], cont_inds, lengthscales, tol
        )
        last_changed_dims = np.where(updated, CONTINUOUS, last_changed_dims)
        for i, choices, xtol in zip(discrete_indices, choices_of_discrete_params, discrete_xtols):
            last_changed_dims = last_changed_dims[~(is_converged := last_changed_dims == i)]
            remaining_inds = remaining_inds[~is_converged]
            if remaining_inds.size == 0:
                return best_xs, best_fvals
            best_xs[remaining_inds], best_fvals[remaining_inds], updated = _local_search_discrete(
                acqf, best_xs[remaining_inds], best_fvals[remaining_inds], i, choices, xtol
            )
            last_changed_dims = np.where(updated, i, last_changed_dims)

        # Parameters not changed from the beginning or last changed dimension is continuous.
        remaining_inds = remaining_inds[~(is_converged := last_changed_dims == CONTINUOUS)]
        last_changed_dims = last_changed_dims[~is_converged]
        if remaining_inds.size == 0:
            return best_xs, best_fvals
    else:
        _logger.warning("local_search_mixed: Local search did not converge.")
    return best_xs, best_fvals
```

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: Thanks for the review and for improving the code. It looks much clearer now.

I have a quick question about this part:

```python
if _imports.is_successful():
    _logger.warning(
        "The 'greenlet' package is unavailable, falling back to sequential L-BFGS-B optimization. "
        "This may lead to slower suggestions."
    )
```

I'm wondering if the intention here is to log a message when the import fails. If so, shouldn't the condition be `if not _imports.is_successful(): ...` ?

Could you please confirm the logic?

nabenabe0928: Oh right, thank you for the catch:)

nabenabe0928:
Followup:
- Add unit tests
- Benchmark the multi-objective optimization since the maximum tensor size would be very huge

optuna/_gp/optim_mixed.py
Kaichi-Irie: Thanks for the feedback. Using a positional-only argument would fix the mypy error.

```python
class FuncAndGrad(Protocol):
 	def __call__(self, x: np.ndarray, /) -> tuple[np.ndarray, np.ndarray]:
 		...
```

However, I think a Callable type alias is a simpler and more direct approach than defining a Protocol here. I'll update the PR to use this instead:

```python
from typing import Callable
...
func_and_grad: Callable[[np.ndarray], tuple[np.ndarray, np.ndarray]]
...
```

optuna/_gp/optim_mixed.py
contramundum53: I would prefer ndim to be either fixed to 2 or accept any dimension (the dimensions except last are considered as the batch).
Making it accept either 1 or 2 and nothing else creates additional branching, which needs to be checked separately by tests.

contramundum53:
Could you explicitly add greenlet inside project.toml, as an optional dependency (like torch)? 
Putting assumption on transitive dependency is not good.

contramundum53:
Also, I think we should have a followup discussion as to whether create some branches in optional dependencies, because GPSampler now gets multiple necessary dependencies (torch and greenlet).
`greenlet` is currently depended on by sqlalchemy so users should have them already, but we don't want the silent slow down when sqlalchemy doesn't depend on greenlet anymore.

optuna/_gp/batched_lbfgsb.py
```diff
@@ -0,0 +1,138 @@
...
+            m=m,
+            factr=factr,
+            pgtol=pgtol,
+            max_evals=max_evals,
+            max_iters=max_iters,
+            max_line_search=max_line_search,
+        )
+
+    # fall back to sequential optimization if greenlet is not available.
+    else:
```
contramundum53, line 121: I think we should have some kind of warning when execution gets fallen back.
Silent slow down due to missing dependency is almost always bad experience IMO.

optuna/_gp/optim_mixed.py
```diff
@@ -180,13 +196,36 @@ def _local_search_discrete(
...
+def _local_search_discrete_batched(
     acqf: BaseAcquisitionFunc,
-    initial_normalized_params: np.ndarray,
+    initial_params_batched: np.ndarray,
+    initial_fvals: np.ndarray,
+    param_idx: int,
+    choices: np.ndarray,
+    xtol: float,
+) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+    best_normalized_params_batched = initial_params_batched.copy()
```
contramundum53, line 214:
```suggestion
    # TODO(Kaichi-Irie): Actually make it batched.
    best_normalized_params_batched = initial_params_batched.copy()
```
Nit

contramundum53 (approved):
Otherwise LGTM

optuna/_gp/batched_lbfgsb.py
contramundum53: warning.warn should be appropriate, because it's expected to be fixable by user.

nabenabe0928: +1, Same opinion
- https://github.com/optuna/optuna/pull/6268#discussion_r2330123819

optuna/_gp/optim_mixed.py
Kaichi-Irie: Thanks for the feedback. I'll update `negative_acqf_with_grad` to accept multi-dimensional tensors. As pointed out in [this comment](https://github.com/optuna/optuna/pull/6268#discussion_r2330178473), this change will also allow me to remove the conditional branching and `func_and_grad_1D_wrapper` from `batched_lbfgsb.py`.

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: Thank you for your feedback. I've now added a warning when a fall back occurs.

optuna/_gp/optim_mixed.py
Kaichi-Irie: Thanks for this fix! I agree this implementation is clearer.

I had a few thoughts on the naming conventions:
- `_local_search_discrete_batched` and `_gradient_ascent_batched` should probably include "batched" in their names.
- `optimize_acqf_mixed` should not be named "batched".
- I think `local_search_mixed` or `local_search_mixed_batched`  are both acceptable, as you pointed out.

(Minor note: To use the walrus operator with indexing in Python 3.8/3.9, you need to wrap the expression in parentheses, e.g., `arr[(x:=1)]`.)

optuna/_gp/optim_mixed.py
Kaichi-Irie: It was unnecessary, so I removed it, thank you.

optuna/_gp/optim_mixed.py
```diff
@@ -45,32 +50,44 @@ def _gradient_ascent(
     As the domain of `x` is [0, 1], that of `z` becomes [0, 1/l].
     """
     if len(continuous_indices) == 0:
-        return initial_params, initial_fval, False
-    normalized_params = initial_params.copy()
+        return initial_params_batched, initial_fvals, np.array([False] * len(initial_fvals))
```
not522:
```suggestion
        return initial_params_batched, initial_fvals, np.zeros(len(initial_fvals), dtype=bool)
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -0,0 +1,126 @@
+from __future__ import annotations
+
+from typing import Callable
```
not522: `typing.Callable` is deprecated since Python 3.9. Please use `collections.abc.Callable` instead.

```suggestion
from collections.abc import Callable
```

optuna/_gp/optim_mixed.py
Kaichi-Irie: Thank you for the review. This change improves the code's readability.

pyproject.toml
```diff
@@ -91,6 +91,7 @@ optional = [
   "scikit-learn>=0.24.2",  # optuna/importance.
   "scipy",  # optuna/_gp.
   "torch",  # optuna/_gp.
+  "greenlet",  # optuna/_gp. necessary for faster gp. Currently greenlet is included as a dependency of SQLAlchemy.
```
not522: Some environments don't include greenlet in their SQLAlchemy dependencies, so how about removing this comment? https://docs.sqlalchemy.org/en/20/faq/installation.html

```suggestion
  "greenlet",  # optuna/_gp. necessary for faster gp.
```

Kaichi-Irie: I didn't realize there were versions of sqlalchemy that don't include greenlet - thank you for pointing that out.
However, I'd like to emphasize that this batch evaluation doesn't introduce a completely new dependency,  greenlet; rather, it's already included in the sqlalchemy dependency.

not522: Source code comments are written to help future readers and modifers understand the code, so describing this particular change could be misleading. (Such comments should remain on the PR instead.) Here, simply commenting where greenlet is used is sufficient.

Kaichi-Irie: OK, I understand. I'll remove the comment. Thank you for the feedback!

not522 (approved):
I confirmed the speed improvement in my environment.

optuna/_gp/optim_mixed.py
```diff
@@ -45,32 +50,44 @@ def _gradient_ascent(
...
 
-    def negative_acqf_with_grad(scaled_x: np.ndarray) -> tuple[float, np.ndarray]:
+    def negative_acqf_with_grad(scaled_x: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
         # Scale back to the original domain, i.e. [0, 1], from [0, 1/s].
-        normalized_params[continuous_indices] = scaled_x * lengthscales
-        (fval, grad) = acqf.eval_acqf_with_grad(normalized_params)
+        if scaled_x.ndim == 1:
+            scaled_x = scaled_x[None, :]
+        normalized_params[: len(scaled_x), continuous_indices] = scaled_x * lengthscales
+        x_tensor = torch.from_numpy(normalized_params[: len(scaled_x), :]).requires_grad_(True)
```
nabenabe0928: This slicing is indeed incorrect when we have (a) discrete parameter(s).
For example, think about the case where `normalized_params = np.array([1.0, 0.0], [0.0, 1.0])` and `normalized_params[:, 1]` is the discrete dimension.
Let's say the first batch converged and the second batch is still in progress, then the current slicing uses the discrete parameter value of 0.0, which is indeed from the first batch.
But we would like to use the discrete parameter from the second batch, meaning that we need to have the unconverged batch indices from the `batched_l_bfgs_b` so that we can correctly assign the discrete parameter values.

Kaichi-Irie: Thanks for pointing out the bug! You are right.
I added batch indices as an argument to the `fun_and_grad` function, as well as the corresponding code. This change appears to be working; the runtime and accuracy seem fine.

PR #6273, opened 2025-09-12
Make the interface of `batched_lbfgsb` module compatible with `scipy.optimize`

Kaichi-Irie:
## Motivation
This PR introduces a refactoring of batched_lbfgsb and its related functions, making the interface of the `batched_lbfgsb` compatible with the `scipy.optimize` module. The goal is to address two pieces of technical debt that could lead to future bugs and make the API less intuitive.

1. Resolve Variable Capture to Prevent Bugs
The core issue is that the nested function `negative_acqf_with_grad` was implicitly using the `normalized_params` variable from its enclosing scope (`_gradient_ascent_batched`). While this may not be causing an active bug, it is a form of technical debt that makes the code harder to reason about and could easily lead to errors in the future. This PR resolves the capture.

2. Improve API Flexibility for `func_and_grad`
The current implementation of `batched_lbfgsb` implicitly requires that the `func_and_grad` callable accepts a second argument (`batch_indices`), even if the function itself does not use it. This design is inflexible and counter-intuitive.

For example, a simple single-argument function is currently incompatible:

```python
# This function is not used, but is required as a placeholder argument
def myfunc1(x: np.ndarray, batched_indices: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
    value = -np.sum(x**2)
    grad = -2 * x
    return value, grad

# This more natural, single-argument function should work, but currently throws an error
def myfunc2(x: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
    value = -np.sum(x**2)
    grad = -2 * x
    return value, grad

# Works correctly
batched_lbfgsb(myfunc1, ...)

# Fails with a TypeError because the second argument is missing
batched_lbfgsb(myfunc2, ...)
```

This change allows `batched_lbfgsb` to correctly handle callables with and without the second argument, making the API more flexible and user-friendly.

## Description of the changes
- The function signature for `batched_lbfgsb` has been updated.
- The `batch_indices` argument has been removed from`func_and_grad`. This information is now passed via `args_list` instead.

### DIFF
--- a/tests/gp_tests/test_batched_lbfgsb.py
+++ b/tests/gp_tests/test_batched_lbfgsb.py
@@ -135,7 +135,7 @@ def test_batched_lbfgsb_invalid_input() -> None:
     def dummy_func_and_grad(x: np.ndarray, _arg: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
         return np.sum(x, axis=1), np.ones_like(x)
 
-    with pytest.raises(ValueError):
+    with pytest.raises(AssertionError):
         batched_lbfgsb(
             func_and_grad=dummy_func_and_grad,
             x0_batched=x0_batched,


optuna/_gp/batched_lbfgsb.py
```diff
@@ -48,17 +50,18 @@ def _batched_lbfgsb(
...
     fvals_opt = np.empty(batch_size, dtype=float)
     n_iterations = np.empty(batch_size, dtype=int)
-    is_remaining_batch = np.ones(batch_size, dtype=bool)
 
     def run(i: int) -> None:
-        def _func_and_grad(x: np.ndarray) -> tuple[float, np.ndarray]:
-            fval, grad = greenlet.getcurrent().parent.switch(x)
+        def _func_and_grad(x: np.ndarray, *args: Any) -> tuple[float, np.ndarray]:
+            args_ = None if len(args) == 0 else args
+            fval, grad = greenlet.getcurrent().parent.switch(x, args_)
```
nabenabe0928:
```suggestion
            fval, grad = greenlet.getcurrent().parent.switch(x, args if len(args) else None)
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -119,15 +126,19 @@ def batched_lbfgsb(
 
     # fall back to sequential optimization if greenlet is not available.
     else:
+
+        def _func_and_grad_wrapper(x: np.ndarray, *args: Any) -> tuple[float, np.ndarray]:
+            fval, grad = func_and_grad(x, *(list(args),) if len(args) else ())
```
nabenabe0928:
```suggestion
            fval, grad = func_and_grad(x, *([args],) if len(args) else ())
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -119,15 +125,19 @@ def batched_lbfgsb(
...
         fvals_opt = np.empty(x0_batched.shape[0], dtype=float)
         n_iterations = np.empty(x0_batched.shape[0], dtype=int)
         for i, x0 in enumerate(x0_batched):
-            batch_indices = np.array([i])
             xs_opt[i], fvals_opt[i], info = so.fmin_l_bfgs_b(
-                func=func_and_grad,  # type: ignore[arg-type]
-                args=(batch_indices,),  # type: ignore[arg-type]
+                func=_func_and_grad_wrapper,
                 x0=x0,
+                args=(args_list[i],) if args_list is not None else (),  # type: ignore[arg-type]
```
nabenabe0928:
```suggestion
                args=args_list[i] if args_list is not None else (),
```

nabenabe0928:
@not522 
Could you review this PR?

optuna/_gp/batched_lbfgsb.py
```diff
@@ -119,15 +125,19 @@ def batched_lbfgsb(
 
     # fall back to sequential optimization if greenlet is not available.
     else:
+
+        def _func_and_grad_wrapper(x: np.ndarray, *args: Any) -> tuple[float, np.ndarray]:
+            fval, grad = func_and_grad(x, *([args],) if len(args) else ())
```
not522: Is the implementation correct? It's changing types depending on `len(args)`.

nabenabe0928: The SciPy implementation also accepts variable length in the function arguments.

References:
- https://github.com/scipy/scipy/blob/v1.16.2/scipy/optimize/_lbfgsb_py.py#L413
- https://github.com/scipy/scipy/blob/main/scipy/optimize/_optimize.py#L310
- https://github.com/scipy/scipy/blob/main/scipy/optimize/_differentiable_functions.py#L236
- https://github.com/scipy/scipy/blob/main/scipy/_lib/_util.py#L584

nabenabe0928:
```python

import numpy as np
from scipy.optimize import fmin_l_bfgs_b


if TYPE_CHECKING:
    from typing import Any
    from typing import Concatenate
    from typing import ParamSpec

    _Args = ParamSpec("_Args")


def test(func: Callable[Concatenate[np.ndarray, _Args], tuple[float, np.ndarray]]) -> None:
    x0 = np.random.random(5)
    fmin_l_bfgs_b(func=func, x0=x0)


def sphere1(x: np.ndarray) -> tuple[float, np.ndarray]:
    value = float(np.sum(x**2))
    grad = 2.0 * x
    return value, grad


def sphere2(x: np.ndarray, *args: *tuple[Any]) -> tuple[float, np.ndarray]:
    value = float(np.sum(x**2))
    grad = 2.0 * x
    return value, grad


test(sphere1)
test(sphere2)

```

But it seems `Callable[Concatenate[np.ndarray, _Args]` is better.

not522: What I want to point out is that when `args_list` is `None`, the arguments being passed to `negative_acqf_with_grad` might be incorrect.

nabenabe0928: Do you, @not522, think the following looks better?
```suggestion
            if len(args):
                fval, grad = func_and_grad(x, [args])
            else:
                fval, grad = func_and_grad(x)
```

not522: My concern is about the bug, not the code format. For example, the following code raises an error.

```python
batched_lbfgsb.batched_lbfgsb(
    func_and_grad=negative_acqf_with_grad,
    args_list=None,
    ...
)
```

nabenabe0928: I did not see any errors:(

```python
import numpy as np

from optuna._gp.batched_lbfgsb import batched_lbfgsb


def sphere_with_args(x: np.ndarray, *args) -> tuple[np.ndarray, np.ndarray]:
    fX = np.sum(x**2, axis=-1)
    gX = 2 * x
    return fX, gX


def sphere_without_args(x: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
    fX = np.sum(x**2, axis=-1)
    gX = 2 * x
    return fX, gX


print(batched_lbfgsb(sphere_with_args, x0_batched=np.random.random((5, 10))*10 - 5, args_list=None))
print(batched_lbfgsb(sphere_without_args, x0_batched=np.random.random((5, 10))*10 - 5, args_list=None))

```

not522: I'm commenting about `negative_acqf_with_grad`, not about other functions.

Kaichi-Irie: Thanks for FB. This looks like expected behavior, not the bug.
`negative_acqf_with_grad` has to get two positional arguments, and if not, it raises an error and does not work.

Kaichi-Irie: Of course, it is possible to handle errors of the type None in the function `negative_acqf_with_grad`, but I don't think this behavior should be changed.`negative_acqf_with_grad`should get exactly two positional arguments.

not522: Hmm, will `batched_lbfgsb` with `args_list=None` be used? (I'm concerned that the implementation might be too complex.)

nabenabe0928: Yes, it will be used in the future for kernel hyperparameter optimization for multi-objective (or/and constrained) optimization.

not522: Why is it not an empty tuple?

Kaichi-Irie: > Why is it not an empty tuple?

I am sorry, I cannot understand your point. Could you explain in detail? Do you mean `args_list = [()]` or `args_list = ()`?

Kaichi-Irie: BTW, this is a small comment on [my comment](https://github.com/optuna/optuna/pull/6273#discussion_r2361752755).
> `negative_acqf_with_grad` should get exactly two positional arguments.

This behavior is the same as the original SciPy optimization function `fmin_l_bfgs_b`.
```python
import numpy as np
from scipy.optimize import fmin_l_bfgs_b

def sphere1(x: np.ndarray) -> tuple[float, np.ndarray]:
    value = float(np.sum(x**2))
    grad = 2.0 * x
    return value, grad

def sphere2(x: np.ndarray, dummy_param) -> tuple[float, np.ndarray]:
    value = float(np.sum(x**2))
    grad = 2.0 * x
    return value, grad

x0 = np.random.random(5)
fmin_l_bfgs_b(func=sphere1, x0=x0) # OK
fmin_l_bfgs_b(func=sphere2, x0=x0, args=("dummy",)) # OK

fmin_l_bfgs_b(func=sphere2, x0=x0) # TypeError: sphere2() missing 1 required positional argument: 'dummy_param'
```

not522: The default value of `fmin_l_bfgs_b`'s `args` is `()`. So, to make the interface compatible, `[()]` would be preferable.

nabenabe0928:
@not522 Thank you for the review, but as we discussed internally, I will re-assign to another reviewer next week...!

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: I've refactored the logic for handling the function arguments, specifically changing how `args_list` (now `args_tuple`) is structured.

The core idea is to **transpose the argument structure**.

**Previous Implementation:**
Arguments were grouped per batch item, like `args_list = [(alpha_1, beta_1), ...]`. The shape was `(B, D)`, where `B` is the batch size and `D` is the number of arguments.

```python
def f(x, alpha, beta):
	...

def f_b(xs, args_list: list):
	"""
	args_list = [(alpha_1, beta_1), ...] # Shape: (B, D)
	"""
	...
	for x, args in zip(xs, args_list):
		f(x, *args)
```

**New Implementation:**
Arguments are now grouped by type, like `args_tuple = ([alpha_1, ...], [beta_1, ...])`. The shape is `(D, B)`.

```python
def f(x, alpha, beta):
	...

def f_b(xs, alphas, betas):
	...
	# Internally, zip transposes the args for iteration
	for x, *args in zip(xs, alphas, betas):
		f(x, *tuple(args))

def opt_b(func_b, args_tuple: tuple):
	"""
	args_tuple = ([alpha_1,...], [beta_1, ...]) # Shape: (D, B)
	"""
	func_b(xs, *args_tuple)
```

I made this change for two main reasons:

1.  **More Natural API:** As you pointed out, the new implementation is a more natural extension of the SciPy `args` interface. The batched function `f_b` should accept concrete parameters (e.g., `alphas`, `betas`) directly, rather than an abstract list of argument tuples.
2.  **Removes Implicit Assumptions:** The previous approach implicitly assumed `args_list` contained a set of fixed parameters, which is no longer the case.

For these reasons, I have reshaped `args_tuple` from `(B, D)` to `(D, B)`.

nabenabe0928 (commented):
@Kaichi-Irie 
Could you merge the master branch and modify the corresponding unit tests accordingly?

Kaichi-Irie:
I merged the master and modified the tests.

optuna/_gp/batched_lbfgsb.py
```diff
@@ -70,30 +72,39 @@ def _func_and_grad(x: np.ndarray) -> tuple[float, np.ndarray]:
...
+        x_and_argsT_pairs, greenlets = map(list, zip(*live_pairs_and_greenlets))
+
     return xs_opt, fvals_opt, n_iterations
 
 
 def batched_lbfgsb(
-    func_and_grad: Callable[[np.ndarray, np.ndarray], tuple[np.ndarray, np.ndarray]],
+    func_and_grad: Callable[..., tuple[np.ndarray, np.ndarray]],
     x0_batched: np.ndarray,
+    args_tuple: tuple[Any] | None = None,
```
not522: How about use an empty tuple as a default value?

```suggestion
    args_tuple: tuple[Any] = (),
```

Kaichi-Irie: Also looks better for me, so I fixed it. Thank you for your feedback!

optuna/_gp/batched_lbfgsb.py
```diff
@@ -48,17 +50,17 @@ def _batched_lbfgsb(
     xs_opt = np.empty_like(x0_batched)
     fvals_opt = np.empty(batch_size, dtype=float)
     n_iterations = np.empty(batch_size, dtype=int)
-    is_remaining_batch = np.ones(batch_size, dtype=bool)
 
     def run(i: int) -> None:
-        def _func_and_grad(x: np.ndarray) -> tuple[float, np.ndarray]:
-            fval, grad = greenlet.getcurrent().parent.switch(x)
+        def _func_and_grad(x: np.ndarray, *args: Any) -> tuple[float, np.ndarray]:
+            fval, grad = greenlet.getcurrent().parent.switch(x, args if args else None)
```
not522:
```suggestion
            fval, grad = greenlet.getcurrent().parent.switch(x, args)
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -48,17 +50,17 @@ def _batched_lbfgsb(
...
             # NOTE(nabenabe): copy is necessary to convert grad to writable.
             return float(fval), grad.copy()
 
         x_opt, fval_opt, info = so.fmin_l_bfgs_b(
             func=_func_and_grad,
             x0=x0_batched[i],
+            args=tuple(arg[i] for arg in args_tuple) if args_tuple else (),
```
not522:
```suggestion
            args=tuple(arg[i] for arg in args_tuple),
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -70,30 +72,39 @@ def _func_and_grad(x: np.ndarray) -> tuple[float, np.ndarray]:
...
+    while x_and_argsT_pairs:
+        x_batched_list = [pair[0] for pair in x_and_argsT_pairs if pair is not None]
+        args_tuple_transposed = [pair[1] for pair in x_and_argsT_pairs if pair is not None]
+        x_batched = np.array(x_batched_list)
+
+        if args_tuple_transposed and all(a is not None for a in args_tuple_transposed):
+            current_args_tuple = tuple(zip(*args_tuple_transposed))
+            fvals, grads = func_and_grad(x_batched, *current_args_tuple)
+        else:
+            fvals, grads = func_and_grad(x_batched)
```
not522:
```suggestion
        current_args_tuple = tuple(zip(*args_tuple_transposed))
        fvals, grads = func_and_grad(x_batched, *current_args_tuple)
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -119,15 +133,23 @@ def batched_lbfgsb(
 
     # fall back to sequential optimization if greenlet is not available.
     else:
+        # args: (P,dim) -> (P,1,dim)
+        def _func_and_grad_wrapper(x: np.ndarray, *args: Any) -> tuple[float, np.ndarray]:
+            if args:
+                args_ = ([arg] for arg in args)  # (P,1,dim)
+                fval, grad = func_and_grad(x, *args_)
+            else:
+                fval, grad = func_and_grad(x)
```
not522:
```suggestion
            args_ = ([arg] for arg in args)  # (P,1,dim)
            fval, grad = func_and_grad(x, *args_)
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -119,15 +133,23 @@ def batched_lbfgsb(
...
         fvals_opt = np.empty(x0_batched.shape[0], dtype=float)
         n_iterations = np.empty(x0_batched.shape[0], dtype=int)
         for i, x0 in enumerate(x0_batched):
-            batch_indices = np.array([i])
             xs_opt[i], fvals_opt[i], info = so.fmin_l_bfgs_b(
-                func=func_and_grad,  # type: ignore[arg-type]
-                args=(batch_indices,),  # type: ignore[arg-type]
+                func=_func_and_grad_wrapper,
                 x0=x0,
+                args=tuple(arg[i] for arg in args_tuple) if args_tuple else (),
```
not522:
```suggestion
                args=tuple(arg[i] for arg in args_tuple),
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +113,15 @@ def batched_lbfgsb(
     max_iters: int = 15000,
     max_line_search: int = 20,
 ) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
-
+    if args_tuple and x0_batched.ndim > 2:
+        raise ValueError("x0_batched must be 2D when args_tuple is provided.")
```
not522: How about limit to 2D even if `args_tuple` is not provided?

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: Thanks for your comment. It looks simpler and working good for me, too. So, I modified and pushed them.

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: Do you mean to restrict the input array strictly to 2D, or allow up to 2D (≤2D)?
I agree with **requiring exactly 2D input** (not 1D, not 3D+) for simplicity and readability. For example:

```python
def batched_lbfgsb(...):
    ...
    assert x0_batched.ndim == 2, "x0_batched must be 2D."
    ...
```

The current implementation allows 1D input, which makes the shape of `args_tuple` and the signature of `func_and_grad` ambiguous. By requiring all input arrays (`batched_lbfgsb`, `func_and_grad`, etc.) to be 2D, the shape of `args_tuple` is always `(B, arg_dims)`, and the implementation around it becomes more consistent and easier to follow.

If 1D input support is required, introducing a dedicated wrapper function would be a clearer solution (rather than branching inside `batched_lbfgsb` and `func_and_grad`).
For a 1D input array, the shape of `args` would naturally be `(args_dim,)` instead of `(1, args_dim)`.

not522: I also think it's better to limit it to 2D. 👍

Kaichi-Irie: Thanks for your advice!
I modified the implementation to restrict the input to a 2D array and added some shape verification code.

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +104,44 @@ def batched_lbfgsb(
...
+      `args_tuple` is `([alpha1, ..., alphaB], [beta1, ..., betaB])`, then
+      `func_and_grad` is called as
+      `func_and_grad(x0_batched, [alpha1, ..., alphaB], [beta1, ..., betaB])`. Note that each
+      argument in `args_tuple` is expected to be a list of length `B` (batch size).
+    """
+    if x0_batched.ndim != 2:
+        raise ValueError("x0_batched must be 2D array.")
+
+    batch_size, dim = x0_batched.shape
+    # Validate args_tuple shapes: each arg must be a sequence of length B
```
not522:
```suggestion
    # Validate args_tuple shapes: each arg must be a sequence of length B.
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +104,44 @@ def batched_lbfgsb(
...
+    for j, arg in enumerate(args_tuple):
+        try:
+            if len(arg) != batch_size:
+                raise ValueError(f"args_tuple[{j}] must have length {batch_size}, got {len(arg)}.")
+        except TypeError as e:
+            raise TypeError(
+                f"Each element of args_tuple must be a sequence with length {batch_size}."
+            ) from e
+
+    # Validate bounds
```
not522:
```suggestion
    # Validate bounds.
```

not522 (commented):
Thank you for your update! The refactoring is LGTM!
This is somewhat trivial, but could you add tests to check for errors when the argument's shape is incorrect?

Kaichi-Irie:
I added the corresponding unit tests to check if the intended errors occur when the shape of args_tuple is mismatched.

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +104,44 @@ def batched_lbfgsb(
...
+    batch_size, dim = x0_batched.shape
+    # Validate args_tuple shapes: each arg must be a sequence of length B.
+    for j, arg in enumerate(args_tuple):
+        try:
+            if len(arg) != batch_size:
+                raise ValueError(f"args_tuple[{j}] must have length {batch_size}, got {len(arg)}.")
+        except TypeError as e:
+            raise TypeError(
+                f"Each element of args_tuple must be a sequence with length {batch_size}."
+            ) from e
```
not522: Could you test these validations?

tests/gp_tests/test_batched_lbfgsb.py
```diff
@@ -120,3 +117,28 @@ def test_batched_lbfgsb(
...
+    batch_size = 10
+    dimension = 2
+    x0_batched = np.random.rand(batch_size, dimension)
+
+    def dummy_func_and_grad(x: np.ndarray, _arg: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
+        return np.sum(x, axis=1), np.ones_like(x)
+
+    with pytest.raises(
+        ValueError, match=rf"args_tuple\[0\] must have length {batch_size}, got 5."
+    ):
```
not522: Our policy is to not test error messages because they are fragile.

```suggestion
    with pytest.raises(ValueError):
```

tests/gp_tests/test_batched_lbfgsb.py
```diff
@@ -120,3 +117,28 @@ def test_batched_lbfgsb(
...
+        batched_lbfgsb(
+            func_and_grad=dummy_func_and_grad,
+            x0_batched=x0_batched,
+            args_tuple=(np.random.rand(5),),
+        )
+
+    with pytest.raises(TypeError, match="Each element of args_tuple must be a sequence"):
```
not522: Ditto.

```suggestion
    with pytest.raises(TypeError):
```

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: I added tests corresponding to each error. Should I add more?

optuna/_gp/optim_mixed.py
```diff
@@ -49,21 +49,21 @@ def _gradient_ascent_batched(
...
     ) -> tuple[np.ndarray, np.ndarray]:
+        next_params = np.array(fixed_params)  # (B, dim)
+        assert next_params.ndim == 2
         # Scale back to the original domain, i.e. [0, 1], from [0, 1/s].
-        if scaled_x.ndim == 1:
-            # NOTE(Kaichi-Irie): When scaled_x is 1D, regard it as a single batch.
-            scaled_x = scaled_x[None]
         assert scaled_x.ndim == 2
-        normalized_params[np.ix_(batch_indices, continuous_indices)] = scaled_x * lengthscales
+        assert next_params[:, continuous_indices].shape == scaled_x.shape
```
nabenabe0928: Isn't the following assertion sufficient?
```python
        assert scaled_x.ndim == 2 and next_params.ndim == 2
```

Although `continuous_indices` is implicitly assumed to be a 1d integer array, this should be tested in the outer scope if really necessary.
As long as `continuous_indices` is 1d and the assertion above passes, it automatically passes the original assertion.
I basically do not want to repeat the unnecessary index filtering.

nabenabe0928: Indeed, we already checked `assert next_params.ndim == 2` above, so this line literally does nothing as long as `continuous_indices` is a 1d integer array.

optuna/_gp/batched_lbfgsb.py
not522: Sorry, I pointed the wrong place.

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +104,44 @@ def batched_lbfgsb(
...
+      a 1D array of function values and a 2D array of gradients.
+    - `x0_batched` is a 2D array where each row is an initial point for optimization.
+    - `args_tuple` is a tuple of additional arguments to pass to `func_and_grad`. e.g., if
+      `args_tuple` is `([alpha1, ..., alphaB], [beta1, ..., betaB])`, then
+      `func_and_grad` is called as
+      `func_and_grad(x0_batched, [alpha1, ..., alphaB], [beta1, ..., betaB])`. Note that each
+      argument in `args_tuple` is expected to be a list of length `B` (batch size).
+    """
+    if x0_batched.ndim != 2:
+        raise ValueError("x0_batched must be 2D array.")
```
not522: Could you test this validation?

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +104,44 @@ def batched_lbfgsb(
...
+                f"Each element of args_tuple must be a sequence with length {batch_size}."
+            ) from e
+
+    # Validate bounds.
+    if bounds is not None:
+        if len(bounds) != dim:
+            raise ValueError(f"bounds must have length {dim}, got {len(bounds)}.")
+        for k, (lb, ub) in enumerate(bounds):
+            if lb > ub:
+                raise ValueError(f"Lower bound must be <= upper bound at index {k}.")
```
not522: Ditto.

optuna/_gp/batched_lbfgsb.py
```diff
@@ -41,24 +43,23 @@ def _batched_lbfgsb(
     max_iters: int,
     max_line_search: int,
 ) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
-    assert (
-        x0_batched.ndim == 2
-    ), f"The shape of x0 must be (batch_size, dim), but got {x0_batched.shape}."
+    if x0_batched.ndim != 2:
+        raise ValueError("x0_batched must be 2D array.")
```
not522: This error never actually occurs, so an assert is sufficient.

```suggestion
    assert x0_batched.ndim == 2
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -119,15 +153,23 @@ def batched_lbfgsb(
 
     # fall back to sequential optimization if greenlet is not available.
     else:
+
+        def _func_and_grad_wrapper(x_1d: np.ndarray, *args_1d: Any) -> tuple[float, np.ndarray]:
+            if x_1d.ndim != 1:
+                raise ValueError("x must be 1D array.")
```
not522:
```suggestion
            assert x_1d.ndim == 1
```

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: Let me check my understanding. 
Is `assert` preferred to `raise SomeError` when the function is not intended for user use and the variable will not be changed by users? If so, should I use `assert` instead of `raise SomeError` around the error handlers of `bounds` and `args_tuple`?

tests/gp_tests/test_batched_lbfgsb.py
Kaichi-Irie: Sorry for the last-minute change, but I changed the type of `args_tuple` from `tuple[Any, ...]` to `tuple[list[Any], ...]` and removed the `TypeError` catcher.

optuna/_gp/optim_mixed.py
Kaichi-Irie: Thanks for your comment. As you pointed out, this assertion looks verbose, so I remove it.

optuna/_gp/batched_lbfgsb.py
not522: The definition of "user" seems problematic. From my perspective, rather than referring to Optuna users, we should consider the caller of `batched_lbfgsb` as the "user." Therefore, `ValueError` would be more appropriate for shape errors in `bounds` or `args_tuple`.

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: I added the corresponding tests!

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: I see, thank you for your clear explanation!

not522:
Thank you for update! It's LGTM except this comment. https://github.com/optuna/optuna/pull/6273#pullrequestreview-3321157392

nabenabe0928:
@Kaichi-Irie 
I will start the review except for the one @not522 mentioned above.
Please address his comment!

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: I replaced `ValueError` with assertion.

not522 (approved):

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +103,38 @@ def batched_lbfgsb(
...
+        if len(arg) != batch_size:
+            raise ValueError(f"args_tuple[{j}] must have length {batch_size}, got {len(arg)}.")
+
+    # Validate bounds.
+    if bounds is not None:
+        if len(bounds) != dim:
+            raise ValueError(f"bounds must have length {dim}, got {len(bounds)}.")
+        for k, (lb, ub) in enumerate(bounds):
+            if lb > ub:
+                raise ValueError(f"Lower bound must be <= upper bound at index {k}.")
```
nabenabe0928: Do we need it?
It is already tested in SciPy.

nabenabe0928:
```suggestion
```

Kaichi-Irie: Thank you for your comment. As you pointed out, it's unnecessary to validate whether the lower bound is smaller than the upper bound here. I'll remove it.
However, validating the length of the bounds is necessary because the API specification is unclear. The user might think the shape of `bounds` is `(B, D)`, not `(D,)`.

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +103,38 @@ def batched_lbfgsb(
...
+      `func_and_grad(x0_batched, [alpha1, ..., alphaB], [beta1, ..., betaB])`. Note that each
+      argument in `args_tuple` is expected to be a list of length `B` (batch size).
+    """
+    assert x0_batched.ndim == 2
+
+    batch_size, dim = x0_batched.shape
+    # Validate args_tuple shapes: each arg must be a sequence of length B.
+    for j, arg in enumerate(args_tuple):
+        if len(arg) != batch_size:
+            raise ValueError(f"args_tuple[{j}] must have length {batch_size}, got {len(arg)}.")
```
nabenabe0928: `batch_size, dim = x0_batched.shape` checks whether `x0_batched` is 2D.

`X must be .... Got Y` or `X must be ..., but got Y.` is more common.
Optuna mostly uses the latter.

```suggestion
    batch_size, dim = x0_batched.shape
    # Validate args_tuple shapes: each arg must be a sequence of length B.
    for j, arg in enumerate(args_tuple):
        if len(arg) != batch_size:
            raise ValueError(f"args_tuple[{j}] must have length {batch_size}, but got {len(arg)}.")
```

Kaichi-Irie: Thanks for your comment. I'll fix the error comment.

> batch_size, dim = x0_batched.shape checks whether x0_batched is 2D.

As you point out, this unpacking raises `ValueError: too many values to unpack (expected 2)` when the input dimension mismatches.
However, `AssertionError` more clearly shows that the input dimension is limited to 2D, don't you think?

nabenabe0928: > However, AssertionError more clearly shows that the input dimension is limited to 2D

assertion in Optuna is, in principle, a sanity check, so as long as the code fails for any invalid inputs, it is sufficient.

optuna/_gp/batched_lbfgsb.py
```diff
@@ -70,30 +70,31 @@ def _func_and_grad(x: np.ndarray) -> tuple[float, np.ndarray]:
...
-        x_batched = []
-        next_greenlets = []
-        for i, gl in enumerate(greenlets):
-            x = gl.switch((fvals[i], grads[i]))
-            if x is None:
-                continue
-            x_batched.append(x)
-            next_greenlets.append(gl)
-        greenlets = next_greenlets
+    x_and_argsT_pairs = [gl.switch(i) for i, gl in enumerate(greenlets)]
```
nabenabe0928: IMHO, I prefer another name for this variable.
Reasons:
- `args` here is not transposed, but it is simply args. We could probably say each argument is batched, but each argument is not transposed. In this sense, I prefer `batched_args` short for batched arguments.
- `pair` is obvious when we use `_and_`, so it is unnecessary.

My personal recommendation would be `x_and_args`, or `x_and_args_list` if you prefer keeping the data type in the name.

optuna/_gp/batched_lbfgsb.py
```diff
@@ -70,30 +70,31 @@ def _func_and_grad(x: np.ndarray) -> tuple[float, np.ndarray]:
...
+        current_args_tuple = tuple(zip(*args_tuple_transposed))
+        fvals, grads = func_and_grad(x_batched, *current_args_tuple)
+        results = [gl.switch((fvals[i], grads[i])) for i, gl in enumerate(greenlets)]
+        live_pairs_and_greenlets = [
+            (pair, gl) for pair, gl in zip(results, greenlets) if pair is not None
+        ]
+        if not live_pairs_and_greenlets:
+            break
+        x_and_argsT_pairs, greenlets = map(list, zip(*live_pairs_and_greenlets))
+
```
nabenabe0928: Much shorter, but the identical behavior with the following.

> [!NOTE]
> You need to modify the last line of `run`, which corresponds to `greenlet.getcurrent().parent.switch(None, None)`.

```suggestion
        greenlet.getcurrent().parent.switch(None, None)

    greenlets = [greenlet(run) for _ in range(batch_size)]
    x_and_args_list = [gl.switch(i) for i, gl in enumerate(greenlets)]
    batched_args = tuple(zip(*[args for _, args in x_and_args_list]))
    while (x_batched := np.array([x for x, _ in x_and_args_list if x is not None])).size:
        fvals, grads = func_and_grad(x_batched, *batched_args)
        x_and_args_list = [gl.switch(fvals[i], grads[i]) for i, gl in enumerate(greenlets)]
        greenlets = [gl for (x, _), gl in zip(x_and_args_list, greenlets) if x is not None]
        batched_args = tuple(zip(*[args for _, args in x_and_args_list if args is not None]))
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +103,38 @@ def batched_lbfgsb(
...
+    """
+    Batched L-BFGS-B optimization with/without greenlet.
+    - `func_and_grad` is expected to take a 2D array as the first argument and return a tuple of
+      a 1D array of function values and a 2D array of gradients.
+    - `x0_batched` is a 2D array where each row is an initial point for optimization.
+    - `args_tuple` is a tuple of additional arguments to pass to `func_and_grad`. e.g., if
+      `args_tuple` is `([alpha1, ..., alphaB], [beta1, ..., betaB])`, then
+      `func_and_grad` is called as
+      `func_and_grad(x0_batched, [alpha1, ..., alphaB], [beta1, ..., betaB])`. Note that each
+      argument in `args_tuple` is expected to be a list of length `B` (batch size).
```
nabenabe0928: Same statement, but the argument here is not arguments of tuple, but batched arguments.
Importantly, `args` is basically a tuple, so saying tuple does not change anything in the context.
The most important information is that each argument is batched.
Please replace `args_tuple` with `batched_args` for correctness.

```suggestion
    - `batched_args` is additional arguments to pass to `func_and_grad`. e.g., if
      `batched_args` is `([alpha1, ..., alphaB], [beta1, ..., betaB])`, then
      `func_and_grad` is called by
      `func_and_grad(x0_batched, [alpha1, ..., alphaB], [beta1, ..., betaB])`. Note that each
      argument in `batched_args` is expected to be a list of length `B` (batch size).
```

nabenabe0928:
I left some comments, but otherwise, the PR looks good to me.

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: I understand. Thank you!

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +103,33 @@ def batched_lbfgsb(
...
+    """
+    batch_size, dim = x0_batched.shape
+    # Validate args_tuple shapes: each arg must be a sequence of length B.
+    for j, arg in enumerate(args_tuple):
+        if len(arg) != batch_size:
+            raise ValueError(f"args_tuple[{j}] must have length {batch_size}, but got {len(arg)}.")
+
+    # Validate bounds.
+    if bounds is not None and len(bounds) != dim:
+        raise ValueError(f"bounds must have length {dim}, but got {len(bounds)}.")
```
nabenabe0928: Probably, this is better because `dim` can be `2`.
```suggestion
    assert bounds is None or np.shape(bounds) == (dim, 2), (
        f"The shape of bounds must be ({dim=}, 2), but got {np.shape(bounds)}."
    )
```

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: Thank you for your feedback! I agree with renaming `args_tuple` as `batched_args`, so I fixed it.

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: Thank you for your comment, this looks better!

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: Thanks for your suggestion! This looks much clearer, so I will modify it.
In my opinion, the code below is almost identical to yours, but it looks simpler because the same processing is repeated in the same order. (However, it might appear more verbose.)
I have already committed this change.
```
...
greenlets = [greenlet(run) for _ in range(batch_size)]
x_and_args_list = [gl.switch(i) for i, gl in enumerate(greenlets)]
x_batched = np.array([x for x, _ in x_and_args_list if x is not None])
batched_args = tuple(zip(*[args for _, args in x_and_args_list if args is not None]))
while x_batched.size:
    fvals, grads = func_and_grad(x_batched, *batched_args)
    x_and_args_list = [gl.switch(fvals[i], grads[i]) for i, gl in enumerate(greenlets)]
    x_batched = np.array([x for x, _ in x_and_args_list if x is not None])
    batched_args = tuple(zip(*[args for _, args in x_and_args_list if args is not None]))
    greenlets = [gl for (x, _), gl in zip(x_and_args_list, greenlets) if x is not None]
...
```

optuna/_gp/batched_lbfgsb.py
Kaichi-Irie: Thank you for your feedback! I agree with your suggestion, and I modified it.

optuna/_gp/batched_lbfgsb.py
```diff
@@ -108,28 +103,32 @@ def batched_lbfgsb(
...
+            raise ValueError(
+                f"batched_args[{j}] must have length {batch_size}, but got {len(arg)}."
+            )
 
     # Validate bounds.
-    if bounds is not None and len(bounds) != dim:
-        raise ValueError(f"bounds must have length {dim}, but got {len(bounds)}.")
+    assert bounds is None or np.shape(bounds) == (
+        dim,
+        2,
```
nabenabe0928, line 122:
```suggestion
        dim, 2
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +98,37 @@ def batched_lbfgsb(
...
+      `func_and_grad(x0_batched, [alpha1, ..., alphaB], [beta1, ..., betaB])`. Note that each
+      argument in `batched_args` is expected to be a list of length `B` (batch size).
+    """
+    batch_size, dim = x0_batched.shape
+    # Validate batched_args shapes: each arg must be a sequence of length B.
+    for j, arg in enumerate(batched_args):
+        if len(arg) != batch_size:
+            raise ValueError(
+                f"batched_args[{j}] must have length {batch_size}, but got {len(arg)}."
+            )
```
nabenabe0928:
```suggestion
        assert len(arg) == batch_size, (
            f"batched_args[{j}] must have length {batch_size}, but got {len(arg)}."
        )
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +98,35 @@ def batched_lbfgsb(
...
+      `func_and_grad` is called by
+      `func_and_grad(x0_batched, [alpha1, ..., alphaB], [beta1, ..., betaB])`. Note that each
+      argument in `batched_args` is expected to be a list of length `B` (batch size).
+    """
+    batch_size, dim = x0_batched.shape
+    # Validate batched_args shapes: each arg must be a sequence of length B.
+    for j, arg in enumerate(batched_args):
+        assert len(arg) == batch_size, (
+            f"batched_args[{j}] must have length {batch_size}, but got {len(arg)}."
+        )
```
nabenabe0928:
```suggestion
        assert (
            len(arg) == batch_size
        ), f"batched_args[{j}] must have length {batch_size}, but got {len(arg)}."
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +98,35 @@ def batched_lbfgsb(
...
+    # Validate batched_args shapes: each arg must be a sequence of length B.
+    for j, arg in enumerate(batched_args):
+        assert (
+            len(arg) == batch_size
+        ), f"batched_args[{j}] must have length {batch_size}, but got {len(arg)}."
+
+    # Validate bounds.
+    assert bounds is None or np.shape(bounds) == (
+        dim, 2
+    ), f"The shape of bounds must be ({dim=}, 2), but got {np.shape(bounds)}."
```
nabenabe0928, line 123:
```suggestion
    assert (
        bounds is None or np.shape(bounds) == (dim, 2)
    ), f"The shape of bounds must be ({dim=}, 2), but got {np.shape(bounds)}."
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +98,35 @@ def batched_lbfgsb(
...
+    # Validate batched_args shapes: each arg must be a sequence of length B.
+    for j, arg in enumerate(batched_args):
+        assert (
+            len(arg) == batch_size
+        ), f"batched_args[{j}] must have length {batch_size}, but got {len(arg)}."
+
+    # Validate bounds.
+    assert (
+        bounds is None or np.shape(bounds) == (dim, 2)
+    ), f"The shape of bounds must be ({dim=}, 2), but got {np.shape(bounds)}."
```
nabenabe0928:
```suggestion
    assert bounds is None or np.shape(bounds) == (
        dim,
        2,
    ), f"The shape of bounds must be ({dim=}, 2), but got {np.shape(bounds)}."
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +98,36 @@ def batched_lbfgsb(
...
+      `batched_args` is `([alpha1, ..., alphaB], [beta1, ..., betaB])`, then
+      `func_and_grad` is called by
+      `func_and_grad(x0_batched, [alpha1, ..., alphaB], [beta1, ..., betaB])`. Note that each
+      argument in `batched_args` is expected to be a list of length `B` (batch size).
+    """
+    batch_size, dim = x0_batched.shape
+    # Validate batched_args shapes: each arg must be a sequence of length B.
+    for j, arg in enumerate(batched_args):
+        assert (
+            len(arg) == batch_size
```
nabenabe0928, line 116:
```suggestion
            len(arg) != batch_size
```

optuna/_gp/batched_lbfgsb.py
```diff
@@ -102,12 +98,36 @@ def batched_lbfgsb(
...
+      `batched_args` is `([alpha1, ..., alphaB], [beta1, ..., betaB])`, then
+      `func_and_grad` is called by
+      `func_and_grad(x0_batched, [alpha1, ..., alphaB], [beta1, ..., betaB])`. Note that each
+      argument in `batched_args` is expected to be a list of length `B` (batch size).
+    """
+    batch_size, dim = x0_batched.shape
+    # Validate batched_args shapes: each arg must be a sequence of length B.
+    for j, arg in enumerate(batched_args):
+        assert (
+            len(arg) != batch_size
```
nabenabe0928:
```suggestion
            len(arg) == batch_size
```

tests/gp_tests/test_batched_lbfgsb.py
```diff
@@ -120,3 +117,37 @@ def test_batched_lbfgsb(
...
+        batched_lbfgsb(
+            func_and_grad=lambda x: (np.sum(x, axis=1), np.ones_like(x)),
+            x0_batched=x0_batched[0],  # not 2D
+        )
+
+    # batched_args validation
+    def dummy_func_and_grad(x: np.ndarray, _arg: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
+        return np.sum(x, axis=1), np.ones_like(x)
+
+    with pytest.raises(ValueError):
```
nabenabe0928:
```suggestion
    with pytest.raises(AssertionError):
```

PR #6274, opened 2025-09-12
Add unit tests for batched L-BFGS-B

Kaichi-Irie:
## Motivation
Add unit tests for `batched_lbfgs` module. This is the followup PR of this PR:
- https://github.com/optuna/optuna/pull/6268/

## Description of the changes
This change add the unit tests for `batched_lbfgsb` with/without `greenlet` and an end-to-end test where `greenlet` is unavailable and `batched_lbfgsb` should be fallen back.

### DIFF
--- a/tests/gp_tests/test_batched_lbfgsb.py
+++ b/tests/gp_tests/test_batched_lbfgsb.py
@@ -9,7 +9,8 @@ import numpy as np
 import pytest
 from scipy.optimize import fmin_l_bfgs_b
 
-from optuna._gp.batched_lbfgsb import batched_lbfgsb
+
+RADIUS = 5.12
 
 
 def rastrigin_and_grad(x: np.ndarray, batch_indices: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
@@ -35,11 +36,12 @@ def styblinski_tang_and_grad(
     return fval, grad
 
 
-def X0_and_bounds(dim: int, n_localopts: int) -> tuple[np.ndarray, np.ndarray]:
-    R = 5.12
+def X0_and_bounds(
+    dim: int, n_localopts: int, lower_bound: float, upper_bound: float
+) -> tuple[np.ndarray, np.ndarray]:
     rng = np.random.RandomState(0)
-    X0 = rng.random((n_localopts, dim)) * 2 * R - R
-    bounds = np.array([[-R, R]] * dim)
+    X0 = rng.random((n_localopts, dim)) * 2 * RADIUS - RADIUS
+    bounds = np.array([[lower_bound, upper_bound]] * dim)
     return X0, bounds
 
 
@@ -48,10 +50,8 @@ def _verify_results(
     func_and_grad: Callable,
     kwargs_ours: Any,
     kwargs_scipy: Any,
-    batched_lbfgsb_func: Callable | None = None,
+    batched_lbfgsb_func: Callable,
 ) -> None:
-    if batched_lbfgsb_func is None:
-        batched_lbfgsb_func = batched_lbfgsb
     xs_opt1, fvals_opt1, n_iters1 = batched_lbfgsb_func(
         func_and_grad=func_and_grad, x0_batched=X0, **kwargs_ours
     )
@@ -84,6 +84,10 @@ test_params = [
 
 @pytest.mark.parametrize("func_and_grad,kwargs_ours,kwargs_scipy", test_params)
 @pytest.mark.parametrize("use_greenlet", [True, False])
+@pytest.mark.parametrize(
+    "lower_bound,upper_bound",
+    [(-RADIUS, RADIUS), (-np.inf, RADIUS), (-RADIUS, np.inf), (-np.inf, np.inf)],
+)
 @pytest.mark.parametrize("dim, n_localopts", [(10, 10), (1, 10), (10, 1), (1, 1)])
 def test_batched_lbfgsb(
     monkeypatch: pytest.MonkeyPatch,
@@ -91,6 +95,8 @@ def test_batched_lbfgsb(
     kwargs_ours: Any,
     kwargs_scipy: Any,
     use_greenlet: bool,
+    lower_bound: float,
+    upper_bound: float,
     dim: int,
     n_localopts: int,
 ) -> None:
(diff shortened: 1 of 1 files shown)


tests/gp_tests/test_batched_lbfgsb.py
nabenabe0928, line 1: Could you make this PR only for this test file?

tests/samplers_tests/test_gp.py
```diff
@@ -110,3 +112,18 @@ def objective(
...
     assert trials[0].values == list(objective(trials[0]))  # The values are normal.
     assert trials[0].system_attrs[_CONSTRAINTS_KEY] is None  # None is set for constraints.
+
+
+def test_behavior_without_greenlet(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setitem(sys.modules, "greenlet", None)
+    import optuna._gp.batched_lbfgsb as my_module
+
+    importlib.reload(my_module)
+    assert my_module._greenlet_imports.is_successful() is False
```
nabenabe0928:
```suggestion
    assert not my_module._greenlet_imports.is_successful()
```

tests/gp_tests/test_batched_lbfgsb.py
```diff
@@ -0,0 +1,149 @@
...
+    if lower_bound is not None:
+        bounds[:, 0] = lower_bound
+    if upper_bound is not None:
+        bounds[:, 1] = upper_bound
+    kwargs_ours.update(bounds=bounds)
+    kwargs_scipy.update(bounds=bounds)
+    _verify_results(X0, func_and_grad, kwargs_ours, kwargs_scipy)
+
+
+def test_greenlet_import_behavior(monkeypatch: pytest.MonkeyPatch) -> None:
```
nabenabe0928: Do we need this test? I feel like it's tested in the other tests. It would be better avoid duplications to not doubly manage similar tests.

tests/gp_tests/test_batched_lbfgsb.py
```diff
@@ -0,0 +1,149 @@
...
+    dim: int,
+    n_localopts: int,
+) -> None:
+    if not use_greenlet:
+        monkeypatch.setitem(sys.modules, "greenlet", None)
+
+    import optuna._gp.batched_lbfgsb as my_module
+
+    importlib.reload(my_module)
+    assert my_module._greenlet_imports.is_successful() is use_greenlet
```
nabenabe0928: Please use `==` to check the value validity.

```suggestion
    assert my_module._greenlet_imports.is_successful() == use_greenlet
```

tests/gp_tests/test_batched_lbfgsb.py
```diff
@@ -0,0 +1,149 @@
...
+        X0, func_and_grad, kwargs_ours, kwargs_scipy, batched_lbfgsb_func=my_module.batched_lbfgsb
+    )
+
+
+@pytest.mark.parametrize("func_and_grad,kwargs_ours,kwargs_scipy", test_params)
+@pytest.mark.parametrize(
+    "lower_bound,upper_bound", [(-np.inf, None), (None, np.inf), (-np.inf, np.inf), (None, None)]
+)
+@pytest.mark.parametrize("dim, n_localopts", [(10, 10), (1, 10), (10, 1), (1, 1)])
+def test_batched_lbfgsb_without_bounds(
```
nabenabe0928: Why not using `use_greenlet` in this function?

tests/gp_tests/test_batched_lbfgsb.py
Kaichi-Irie: Thanks for your comment. As you pointed out, this test is verbose, so I will remove it.

Kaichi-Irie:
Thank you for your review! I addressed the issues you mentioned.

nabenabe0928 (approved):

tests/gp_tests/test_batched_lbfgsb.py
```diff
@@ -0,0 +1,145 @@
...
+    assert my_module._greenlet_imports.is_successful() == use_greenlet
+
+    X0, bounds = X0_and_bounds(dim=dim, n_localopts=n_localopts)
+    if lower_bound is not None:
+        bounds[:, 0] = lower_bound
+    if upper_bound is not None:
+        bounds[:, 1] = upper_bound
+    kwargs_ours.update(bounds=bounds)
+    kwargs_scipy.update(bounds=bounds)
+    _verify_results(X0, func_and_grad, kwargs_ours, kwargs_scipy)
```
y0z: Is `batched_lbfgsb_func=my_module.batched_lbfgsb` not required here?

tests/gp_tests/test_batched_lbfgsb.py
```diff
@@ -0,0 +1,145 @@
...
+    (styblinski_tang_and_grad, {"max_evals": 3}, {"maxfun": 3}),
+    (rastrigin_and_grad, {"max_iters": 3}, {"maxiter": 3}),
+    (styblinski_tang_and_grad, {"max_iters": 3}, {"maxiter": 3}),
+]
+
+
+@pytest.mark.parametrize("func_and_grad,kwargs_ours,kwargs_scipy", test_params)
+@pytest.mark.parametrize("use_greenlet", [True, False])
+@pytest.mark.parametrize("dim, n_localopts", [(10, 10), (1, 10), (10, 1), (1, 1)])
+def test_batched_lbfgsb(
```
y0z, line 92: What's the difference between (1)`test_batched_lbfgsb` and (2)`test_batched_lbfgsb_without_bounds` with `"lower_bound,upper_bound"=(None, None)`?

nabenabe0928: Basically, L-BFGS-B falls back to L-BFGS when bounds are not specified.
It will be used for kernel hyperparameter optimization.

tests/gp_tests/test_batched_lbfgsb.py
nabenabe0928: No, we do not need it.
The fallback automatically happens in batched_l_bfgs_b.

Kaichi-Irie: I modified the implementation to explicitly pass `batched_lbfgsb_func` to `_verify_results`.

tests/gp_tests/test_batched_lbfgsb.py
Kaichi-Irie: Unbounded batched L-BFGS-B will be used to optimize the kernel hyperparameters. 
However, passing the bound as `None` could be misleading, so the implementation was modified to pass the bounds directly.

PR #663, opened 2019-11-04
Make `{FrozenTrial,Trial}.trial_id` private

hvy:
This PR makes `trial_id` a private interface by prefixing it with an underscore. The previous interface is kept but yields a `DeprecationWarning` (already true in master for `Trial` but not for `FrozenTrial`).

`FrozenTrial` is now mutable (subclass of  `object` instead of `namedtuple`) as a `namedtuple` cannot have private attributes prefixed with underscores. It would yield an error unless ([mangled using `rename`](https://docs.python.org/3/library/collections.html#collections.namedtuple)), which makes it more or less pointless as it basically cannot be accessed. I'd like to discuss whether this may cause any issues to existing code.

[docs/source/tutorial/first.rst](https://github.com/pfnet/optuna/blob/master/docs/source/tutorial/first.rst) should probably be updated as well to correctly render the `FrozenTrial`. I'm thinking it could be a separate PR unless it requires by tests as this PR is already quite large.

PR #6281, opened 2025-09-17
Change `TrialState.__repr__` and `TrialState.__str__`

ktns:
## Motivation
According to [this comment](https://github.com/optuna/optuna/pull/663/files#r342478310), `TrialState.__repr__` was intended to return a string that can be passed to `eval` and evaluated as itself. At that time `TrialState` was a subclass of `enum.Enum` but now `enum.IntEnum.__str__` returns just a numeric string.

## Description of the changes
Use `enum.Enum.__str__` in `TrialState.__repr__` and ensure it's evaluated as itself via a unit test.


nabenabe0928:
@c-bata 
Could you review this PR?

not522:
Thank you for your PR and sorry for the late response.
`IntEnum`'s `__str__` and `__repr__` were changed in Python 3.11 (https://docs.python.org/3.13/library/enum.html#enum.IntEnum). So, we need to discuss what behavior is desirable.
I think using `Enum.__str__` for `__str__` and keeping `__repr__` as default is appropriate (https://docs.python.org/3.13/library/enum.html#notes).

```python
class TrialState(enum.IntEnum):
    ...
    __str__ = enum.Enum.__str__
    ...
```
```
>>> import optuna
>>> str(optuna.trial.TrialState.COMPLETE)
'TrialState.COMPLETE'
>>> repr(optuna.trial.TrialState.COMPLETE)
'<TrialState.COMPLETE: 1>'
```

It does not return a valid Python expression for `__repr__`, but I think what number is used as the value is important for developers (See the reference [https://docs.python.org/3.13/reference/datamodel.html#object.\_\_repr\_\_](https://docs.python.org/3.13/reference/datamodel.html#object.__repr__)).
What do you think?

ktns:
I also prefer to use the default `__repr__`. When returning a valid Python expression isn't strictly required, let's use the default.

not522:
Thank you! Could you update this PR as my comment?

not522:
1. How about adding `__str__ = enum.Enum.__str__`? In the current implementation, `str(optuna.trial.TrialState.COMPLETE)` just shows `'1'`.
2. `test_repr` in `test_frozen.py` failed, but I think this test can be deleted.

ktns:
strange, I pushed changes to the test but failing CI job seems to have executed the old test with `eval`...

ktns:
Ah... sorry I misunderstood. I didn't notice there was another test using eval.

not522 (approved):
Thank you! LGTM!

ktns:
I just noticed the PR title was obsolete. Is the new title OK?

gen740 (commented):
I think `tests/trial_tests/test_state.py` is unnecessary since it only tests Python’s built-in functionality, not Optuna.

ktns:
I think it's advisable to be there to notify people when stdlib changed behavior.

ktns:
and also when the baseclass is changed again.

gen740:
> I think it’s advisable to be there to notify people when stdlib changed behavior.

If so, the test should compare with the full output of the __repr__ like,

```python
@pytest.mark.parametrize("state", TrialState)
def test_trial_state_repr(state: TrialState) -> None:
    assert f"<TrialState.{state.name}: {state.value}>" in repr(state)
```

but I think even this is not necessary, because Optuna does not depend on any `__repr__` output as far as I know.

Would there be any issue if the base class of `TrialState` changed and its `__repr__` output became different?

not522:
While I'm neutral about this test, Optuna's policy is to not test visual elements like visualization or error messages, so this test might be unnecessary.

ktns:
I'd rather insist to keep the test because it describes the intention to use `Enum.__repr__`. Before I write this PR, I had totally no idea why `TrialState.__repr__` is returning just a numeric string until finding [this comment](https://github.com/optuna/optuna/pull/663/files#r342478310). PR comments are not easy to find.

not522:
I'm neutral with having tests, but aren't your points about documentation or comments in the code rather than about the tests themselves?

gen740:
In the first place, why does `TrialState` use [IntEnum](https://docs.python.org/3/library/enum.html#enum.IntEnum)?
`IntEnum` is intended for cases where the state needs to be interpreted as an integer—so that values can be added or subtracted.
However, `TrialState` is neither additive nor subtractive, so using a plain `Enum` would be more appropriate.

If `Enum` were used instead of `IntEnum`, the `__repr__` method would not need to be overridden, and the corresponding test would be unnecessary.

not522:
To explain the background: `TrialState` was originally `Enum`, but was changed to `IntEnum` during the implementation of `JournalStorage` (#3854). Inheriting from int makes the JSON conversion implementation within JournalStorage simpler.

ktns:
Of course documentation or comments will do. I won't insist to have the test if they are written but I prefer test personally because documentation or comments tend to get ignored.

 Do you want me to write comments?

c-bata:
@ktns Thank you for your continuous contribution!
Sorry for the extra work, but due to the recent update in the branch protection rule after #6302, could you please merge the latest master branch into yours? :pray:

Regarding the ongoing discussion, I also think leaving a code comment instead of unit tests would be more appropriate in this case, since it would be difficult to find the background from the current test code. I don’t think we should spend too much time for a few lines of test code, but my impression is closer to @gen740’s and @not522’s opinions.

ktns:
Sorry, I totallly forgot that I added a comment on `__str__` definition. Does this look good?

PR #6289, opened 2025-09-30
Limit the blackdoc version

not522:
## Motivation
blackdoc versions from v0.4.2 onward are broken, so setting a version upper limit.

## Description of the changes
Limit the blackdoc version to less than 0.4.2.


y0z (approved):
LGTM

PR #6287, opened 2025-09-26
Update the table of Samplers in docs

fusawa-yugo:
## Motivation

Changing the layout of the Samplers table would help users.

## Description of the changes

Changes I made are:
1. Reorder of columns
   - popular ones are placed on the left
2. Add `NSGAIIISampler`
   - (next to NSGAIISampler)
3. Embed links for each Sampler 

Image:
スクリーンショット 2025-09-26 16 02 59

スクリーンショット 2025-09-26 16 03 22

PR #6290, opened 2025-10-01
Fix the table of samplers in the docs

kAIto47802:
## Motivation

This PR is a follow-up PR for this PR:
- #6287.

## Description of the changes

- Add a link to `AutoSampler` and format it as a code block.
- Fix the link of `BoTorchSampler`, since it is located in `optuna_integration`, not in `optuna`.
- Remove the white spaces before and after the en dash (U+2013)as they should not be inserted according to several style guides, including Chicago Manual of Style [^1], APA Style, IEEE Editorial Style Manual, Oxford Style Manual, New Oxford Style Manual.
- (nit) Fixed the spacing around BruteForceSampler, which was slightly misaligned.

To add a link to the `BoTorchSampler` document in `optuna_integration`, I add `optuna_integration` in `intersphinx_mapping` in `conf.py`. Note that this approach cannot be used for `AutoSampler`, since its document is not based on sphinx.

[^1]: https://www.chicagomanualofstyle.org/qanda/data/faq/topics/HyphensEnDashesEmDashes/faq0108.html


y0z:
@sawa3030, could you review this PR?

sawa3030:
I found that the text `AutoSampler` is displayed in red, even though the link functions correctly and redirects to the corresponding documentation. This behavior appears unintended and undesirable. Could you please update this?

image

kAIto47802:
I’m aware that the text for AutoSampler is showing up in red. If you want to update it, could you please suggest a concrete solution?
As far as I know, the only way to make it exactly the same as the other samplers is to use custom CSS. As I mentioned above, AutoSampler cannot use cross-references, since its documentation is not based on Sphinx but on its own system.

cf. In the original state, it had no link at all and wasn’t even shown as inline code.

kAIto47802:
I also made the following two update:
- Update the order in the autosummary section to match the table, whose order was updated in the previous PR (https://github.com/optuna/optuna/pull/6287).
- Add NSGAIIISampler to the note, since it has the same exception as NSGAIISampler but was missed in the previous PR (https://github.com/optuna/optuna/pull/6287), which added NSGAIIISampler to this document.

docs/source/reference/samplers/index.rst
```diff
@@ -14,34 +14,36 @@ The :mod:`~optuna.samplers` module defines a base class for parameter sampling a
...
++------------------------------------+---------------------------------+-----------------------------------------+--------------------------------------+-------------------------------------+----------------------------------------+-------------------------------------------------------------------------------+-------------------------------------------------------------------------------+---------------------------------------+--------------------------------------+------------------------------------------------+---------------------------------------------------------------------------------+
+
+.. |AutoSampler| replace:: ``AutoSampler``
```
y0z: It's a simple solution, but it appears to work (I've confirmed that the hyperlink also works as expected).
```suggestion
.. |AutoSampler| replace:: :class:`AutoSampler`
```

Image

sawa3030:
> I’m aware that the text for AutoSampler is showing up in red. If you want to update it, could you please suggest a concrete solution? As far as I know, the only way to make it exactly the same as the other samplers is to use custom CSS. As I mentioned above, AutoSampler cannot use cross-references, since its documentation is not based on Sphinx but on its own system.
> 
> cf. In the original state, it had no link at all and wasn’t even shown as inline code.

I’m sorry for leaving everything up to you, and thank you for the clear explanation.
As far as I have investigated, I believe there are four possible options here:

1. Apply custom CSS to AutoSampler, aligning it with the other samplers.
2. Apply custom CSS to the other samplers (excluding AutoSampler), aligning them with AutoSampler.
3. Avoid using any custom CSS (the current implementation in this PR).
4. Remove the link altogether.

I think maintaining a consistent text style across all sampler names is desirable. To achieve this, the second option seems more suitable to me, as it ensures consistent formatting while also making the text bold, which works well for a column header. I’ve added comments based on this approach.
That said, I’m not entirely sure about the potential side effects of this method. I’ve tested it locally but haven’t yet fully examined its broader impact. Any feedback or insights you could share on this approach would be greatly appreciated.

By the way, this is not a strong opinion — if this point has already been discussed and a decision has been made, I’m completely fine with keeping the current implementation of this PR!

docs/source/reference/samplers/index.rst
kAIto47802: I see. Thank you for the information!

PR #6291, opened 2025-10-01
Use cached trials for intersection search space calculation

not522:
## Motivation
The current implementation of `TPESampler` does not utilize cached trials when computing the search space. This PR addresses the issue by delaying the execution of `infer_relative_search_space` until absolutely necessary and performing it concurrently with `sample_relative`, thus enabling the use of cached trials. Additionally, the `TPESampler`'s `infer_relative_search_space` method now incorporates caching functionality.

(While the same performance optimizations are possible for other sampler implementations, for PR brevity, I focus on `TPESampler` here.)

## Description of the changes
This PR introduces a `use_cache` argument for `IntersectionSearchSpace` and `_GroupDecomposedSearchSpace` and makes `TPESampler` using cached trials for intersection search space calculation.

## Benchmark

```python
import optuna

    return x**2 + y**2

sampler = optuna.samplers.TPESampler(seed=42, multivariate=True, constant_liar=True)
# sampler = optuna.samplers.TPESampler(seed=42, multivariate=True, constant_liar=True, group=True)

```
- `group=False`

| master | PR |
| - | - |
| 11.287s | 10.150s |

- `group=True`

| master | PR |
| - | - |
| 11.771s | 10.581s |

### DIFF
--- a/optuna/trial/_trial.py
+++ b/optuna/trial/_trial.py
@@ -58,6 +58,7 @@ class Trial(BaseTrial):
 
         self.study.sampler.before_trial(study, self._cached_frozen_trial)
 
+        # NOTE(not522): Evaluate it lazily to get as latest search as possible.
         self.relative_search_space: dict[str, BaseDistribution] | None = None
         self._relative_params: dict[str, Any] | None = None
         self._fixed_params = self._cached_frozen_trial.system_attrs.get("fixed_params", {})


y0z:
@nabenabe0928, could you review this PR?

optuna/trial/_trial.py
```diff
@@ -58,16 +58,17 @@ def __init__(self, study: "optuna.study.Study", trial_id: int) -> None:
...
         self._relative_params: dict[str, Any] | None = None
         self._fixed_params = self._cached_frozen_trial.system_attrs.get("fixed_params", {})
 
     @property
     def relative_params(self) -> dict[str, Any]:
         if self._relative_params is None:
             study = pruners._filter_study(self.study, self._cached_frozen_trial)
+            self.relative_search_space = self.study.sampler.infer_relative_search_space(
+                study, self._cached_frozen_trial
+            )
```
nabenabe0928, line 72: Could you explain why we need to move the call here?

not522: `_get_trials(..., use_cache=use_cache)` creates the cache during the first call in that trial. Therefore, this call should be delayed until it's actually needed. For example, in the following objective function, calling `infer_relative_search_space` in `__init__` would be premature:

```python
def objective(trial):
    # Very time-consuming initialization process.
    ...
    # Then perform the suggestion.
    trial.suggest_xxx(...)
```

nabenabe0928: Thanks, this makes a lot of sense.

optuna/trial/_trial.py
```diff
@@ -58,16 +58,17 @@ def __init__(self, study: "optuna.study.Study", trial_id: int) -> None:
 
         self.study.sampler.before_trial(study, self._cached_frozen_trial)
 
-        self.relative_search_space = self.study.sampler.infer_relative_search_space(
-            study, self._cached_frozen_trial
-        )
+        self.relative_search_space: dict[str, BaseDistribution] | None = None
```
nabenabe0928, line 62:
```suggestion
        # NOTE(not522): Evaluate it lazily to get as latest search as possible.
        self.relative_search_space: dict[str, BaseDistribution] | None = None
```

PR #6227, opened 2025-07-31
Pass stacklevel to some warnings.warn

ktns:
## Motivation
While testing some code, I encountered the following warning message:
```
 /usr/local/lib/python3.12/site-packages/optuna/visualization/_utils.py:67: UserWarning: `target` is specified, but `target_name` is the default value, 'Objective Value'.
```

In this case, the issue lies not with the called function itself but with the arguments provided by the caller. Therefore, displaying the caller information would be more helpful for users I think.

## Description of the changes
I have added `stacklevel` to `warnings.warn` calls in the above line and some other relevant locations.

I would be grateful if you could review this change.

PR #6293, opened 2025-10-03
Introduce stacklevel-aware custom warnings

gen740:
## Motivation

For better user experience, warnings should point to the user’s code, not Optuna’s internal code.

For example, with the following code in the current master branch, the warning points to non-user code. In such cases, it is difficult for users to identify the actual source in their own code.

```python
import optuna


study = optuna.create_study()


trial = study.ask()

trial.suggest_float("x", 0, 1)
trial.suggest_float("x", 0, 10)
```

output
```
[I 2025-10-03 15:50:51,639] A new study created in memory with name: no-name-e267bc38-d5bf-45f8-b97a-4514fa0d9f67
/Users/gen/work/Optuna/optuna/optuna/trial/_trial.py:682: RuntimeWarning: Inconsistent parameter values for distribution with name "x"! This might be a configuration mistake. Optuna allows to call the same distribution with the same name more than once in a trial. When the parameter values are inconsistent optuna only uses the values of the first call and ignores all following. Using these values: {'step': None, 'low': 0.0, 'high': 1.0, 'log': False}
  warnings.warn(
```

As discussed in #6227, `warnings.warn` accepts a second argument stacklevel, which allows skipping a certain number of frames in the stack. However, passing the correct stack level everywhere is difficult, since different call paths may require different values.

To address this, I implemented a wrapper that determines the stack level based on the module path. This wrapper calculates the appropriate stack level relative to the user’s code and passes it to `warnings.warn`.

Starting from Python 3.12, `warnings.warn` provides the `skip_file_prefixes` argument, which can skip stack frames belonging to files with specific prefixes. However, this argument is not available in versions earlier than 3.12. Therefore, I implemented a custom wrapper inspired by the `skip_file_prefixes` mechanism.

## Description of the changes

- Add `_warnings.py`.
- Replace `import warnings` to `from optuna import _warnings as warnings`.


c-bata:

optuna/_warnings.py
```diff
@@ -0,0 +1,42 @@
...
+        while True:
+            if (
+                not getattr(sys._getframe(level).f_code, "co_filename", "")
+                .casefold()
+                .startswith(_OPTUNA_MODULE_ROOT)
+            ):
+                return level
+            level += 1
+    except ValueError:
+        return level
```
kAIto47802: How about using `inspect` instead of `sys._getframe`, since `sys._getframe` is internal function and not officially supported?
This function can be rewritten using `inspect` as follows:
```suggestion
def find_stacklevel() -> int:
    level = 1
    frame = inspect.currentframe()
    while frame is not None:
        frame = frame.f_back
        filename = frame and frame.f_code and frame.f_code.co_filename
        if not filename or not filename.casefold().startswith(_OPTUNA_MODULE_ROOT):
            return level
        level += 1
    return level
```

kAIto47802 (changes requested):

optuna/_warnings.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def warn(
```
kAIto47802: > Starting from Python 3.12, warnings.warn provides the skip_file_prefixes argument, which can skip stack frames belonging to files with specific prefixes. However, this argument is not available in versions earlier than 3.12. Therefore, I implemented a custom wrapper inspired by the skip_file_prefixes mechanism.

How about adding a note in this file for future developments?

kAIto47802 (commented):
Thank you for the PR! I left some comments. PTRL :octocat:

Also, the stack level is currently specified in the decorators to point to the user code, which I think is no longer necessary with this PR. Therefore, how about removing these specifications?
In the current implementation, the stack level has increased by one, so it no longer points to the user code but instead points inside the Optuna internal decorator. If we don’t remove these specifications, we need to increase their stack level by one.

These specifications can be found in the following files:
- optuna/_convert_positional_args.py
- optuna/_deprecated.py
- optuna/_experimental.py

kAIto47802:
Also, the change to `from optuna import _warnings as warnings` is missing in optuna/samplers/_base.py.

optuna/_warnings.py
c-bata: I agree with @kAIto47802 that `sys._getframe()` is a private function that we should avoid using here.

However, I’m also concerned about relying on `inspect.currentframe()`, since the official Python documentation states that this function is a CPython implementation detail. Furthermore, we should also be careful about ensuring that the while loop does not end up in an infinite loop.

> CPython implementation detail: This function relies on Python stack frame support in the interpreter, which isn’t guaranteed to exist in all implementations of Python. If running in an implementation without Python stack frame support, this function returns None.
> https://docs.python.org/3.13/library/inspect.html#inspect.currentframe

As an alternative, how about simply using the `skip_file_prefixes` option as shown below?

```python
if sys.version_info >= (3, 12):
    warnings.warn(..., skip_file_prefixes= _OPTUNA_MODULE_ROOT)
else:
    warnings.warn(...)
```

kAIto47802: Thank you for the reply!
If the stack frame is not supported and `inspect.currentframe()` returns `None`, this function returns 1, which corresponds to the default stack level setting used in `warnings.warn`. Thus, when the stack frame is unavailable, the behavior is identical to the original `warnings.warn`, while otherwise it points to the user code.

Also, I don't think this while loop can result in an infinite loop, since it moves up the call stack every iterations via `frame = frame.f_back`, and `frame.f_back` eventually returns `None` at the top of the stack. However, do you know any corner  cases where this could lead to an infinite loop? Also, as a alternative, if you're still concerned about the corner some cases, we could use a bounded loop, such as `for _ in range(_MAX_STACK_SEARCH_DEPTH).

> As an alternative, how about simply using the skip_file_prefixes option as shown below?

Thank you for the comment. I agree with the idea of using the `skip_file_prefixes` option if it's available.
However, does it means only is would only points to the user code when running on Python 3.12 or greater, not supporting this functionality on earlier version?

c-bata: > However, does it means only is would only points to the user code when running on Python 3.12 or greater, not supporting this functionality on earlier version?

Yes, users on Python 3.11 or earlier won’t benefit from this change, but it won’t make things any less convenient than before. Considering that the number of users on older versions will gradually decrease, I think it’s better to prioritize code maintainability. What do you think?

kAIto47802: Okay, if we don't pursue this functionality on Python 3.11 or earlier, it would also be acceptable.

gen740:
I changed the name of the wrapper function of `warnings.warn` because I think the name `warn` is so confusing.

kAIto47802 (approved):
Thank you for the update.
I confirmed that it behaves as intended for both Python 3.12 and earlier versions, and for both the decorator and non-decorator cases.
I also confirmed that there are no remaining instances of code that should have been updated to use `optuna_warn`.

One minor concern is that for Python < 3.12, the user message displays `optuna_warn(` instead of `warnings.warn(`, which might be surprising for users who are accustomed to seeing `warnings.warn(` in standard warning messages.
However, since no information is lost compared to the original behavior, and this issue will disappear once support for versions earlier than 3.12 is dropped, it’s acceptable.

Therefore, it LGTM!

c-bata (approved):
LGTM!

Issue #6230, opened 2025-08-01
`GPSampler` suddenly slows down once `n_trials` hits exactly 100


nabenabe0928:
### Expected behavior

`GPSampler` should not slow down even after 100 trials.

### Environment

- Optuna version: latest
- Python version: 3.12
- OS: Ubuntu 20.04

### Error messages, stack traces, or logs

```shell
None
```

### Steps to reproduce

Even the following single objective problem slows down significantly, but I do not know why.

```python
import optuna


def objective(trial: optuna.Trial) -> tuple[float, float]:
    x = trial.suggest_float("x", -5, 5)
    return x**2


sampler = optuna.samplers.GPSampler(seed=0)
study = optuna.create_study(sampler=sampler)
study.optimize(objective, n_trials=110)

```

Kaichi-Irie:
In my environment, this problem does not occur. 
- Optuna version: 4.5.0.dev
- Python version: 3.12.8
- OS: macOS Sequoia Version 15.6
- CPU: M4, 2025

```python
import optuna

def objective(trial: optuna.Trial) -> float:
    x = trial.suggest_float("x", -5, 5)
    y = trial.suggest_float("y", -5, 5)
    return (x - 2.0) ** 2 + (y - 3.0) ** 2

sampler = optuna.samplers.GPSampler(seed=0)
study = optuna.create_study(sampler=sampler)
study.optimize(objective, n_trials=200)
```

nabenabe0928:
Seems the slowdown happens only after numpy v2.0.0.
I did not observe any slowdown when using `numpy==1.26.4`.

nabenabe0928:
The bottleneck indeed comes from `np.linalg.inv`. 
- https://github.com/numpy/numpy/issues/29884#issuecomment-3381534222

We should probably think of switching to `np.linalg.cholesky` instead because the Cholesky-decomposed matrix is sufficient in our case.

nabenabe0928:
It's been resolved in:
- https://github.com/optuna/optuna/pull/6296

PR #6296, opened 2025-10-09
Replace `np.linalg.inv` with `np.linalg.cholesky` to speed up `GPSampler` for `numpy>=2.0.0`

nabenabe0928:
## Motivation

This PR resolves the following issue:
- https://github.com/optuna/optuna/issues/6230

See also the issue in NumPy:
- https://github.com/numpy/numpy/issues/29884

Basically, `np.linalg.inv` suddenly slows down after `cov_Y_Y.shape` exceeds `(100, 100)`.
We avoid this issue by using `np.linalg.cholesky`.

Here are some backgrounds.
Let's denote $C \in \mathbb{R}^{N \times N}$ as the kernel matrix with the addition of a noise at the diagonal elements and $k = [k(x^\star, x_1), \dots, k(x^\star, x_N)]\in \mathbb{N}$ as the kernel vector at a new point $x^\star$.
The Gaussian process infers the posterior variance using $k^\top C^{-1} k$, necessitating the matrix inversion in our current implementation.
However, as $C$ is a positive definite matrix owing to the kernel matrix nature, we can indeed rewrite $v = C^{-1} k$ as $k = C v = L L^\top v$ where $L \in \mathbb{R}^{N \times N}$ is the lower triangular matrix yielded by the Cholesky decomposition of $C$.
We first obtain $u = L^\top v$ by solving  $L u = k$ and then obtain $v$ by solving $L^\top v = u$.
Please note that each linear system can be solved with the time complexity of $O(N^2)$ because $L$ is a triangular matrix.

By doing so, we can avoid the matrix inversion, leading to a significant speedup.
The by-product of this modification is the numerical stability.
The speedup effect can be found below:

bench

## Description of the changes

- Replace `np.linalg.inv` with `np.linalg.cholesky`

### DIFF
--- a/optuna/_gp/gp.py
+++ b/optuna/_gp/gp.py
@@ -196,13 +196,13 @@ class GPRegressor:
         cov_fx_fx = self.kernel_scale  # kernel(x, x) = kernel_scale
         mean = cov_fx_fX @ self._cov_Y_Y_inv_Y
         # K @ inv(C) = V --> K = V @ C --> K = V @ L @ L.T
-        cov_fx_fx_cov_Y_Y_inv = torch.linalg.solve_triangular(
+        cov_fx_fX_cov_Y_Y_inv = torch.linalg.solve_triangular(
             self._cov_Y_Y_chol,
             torch.linalg.solve_triangular(self._cov_Y_Y_chol.T, cov_fx_fX, upper=True, left=False),
             upper=False,
             left=False,
         )
-        var_ = (cov_fx_fx - torch.linalg.vecdot(cov_fx_fX, cov_fx_fx_cov_Y_Y_inv)).clamp_min_(0.0)
+        var_ = (cov_fx_fx - torch.linalg.vecdot(cov_fx_fX, cov_fx_fX_cov_Y_Y_inv)).clamp_min_(0.0)
         return (mean.squeeze(0), var_.squeeze(0)) if is_single_point else (mean, var_)
 
     def marginal_log_likelihood(self) -> torch.Tensor:  # Scalar
--- a/optuna/_gp/gp.py
+++ b/optuna/_gp/gp.py
@@ -129,6 +129,8 @@ class GPRegressor:
         cov_Y_Y[np.diag_indices(self._X_train.shape[0])] += self.noise_var.item()
         cov_Y_Y_chol = np.linalg.cholesky(cov_Y_Y)
         # cov_Y_Y_inv @ y = v --> y = cov_Y_Y @ v --> y = cov_Y_Y_chol @ cov_Y_Y_chol.T @ v
+        # NOTE(nabenabe): Don't use np.linalg.inv because it is too slow und unstable.
+        # cf. https://github.com/optuna/optuna/issues/6230
         cov_Y_Y_inv_Y = scipy.linalg.solve_triangular(
             cov_Y_Y_chol.T,
             scipy.linalg.solve_triangular(cov_Y_Y_chol, self._y_train.numpy(), lower=True),
--- a/optuna/_gp/gp.py
+++ b/optuna/_gp/gp.py
@@ -33,12 +33,12 @@ from optuna.logging import get_logger
 if TYPE_CHECKING:
     from collections.abc import Callable
 
-    import scipy.optimize as so
+    import scipy
     import torch
 else:
     from optuna._imports import _LazyImport
 
-    so = _LazyImport("scipy.optimize")
+    scipy = _LazyImport("scipy")
     torch = _LazyImport("torch")
 
 logger = get_logger(__name__)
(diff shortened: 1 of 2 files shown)


optuna/_gp/gp.py
```diff
@@ -181,13 +188,21 @@ def posterior(self, x: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
         Please note that we clamp the variance to avoid negative values due to numerical errors.
         """
         assert (
-            self._cov_Y_Y_inv is not None and self._cov_Y_Y_inv_Y is not None
+            self._cov_Y_Y_chol is not None and self._cov_Y_Y_inv_Y is not None
         ), "Call cache_matrix before calling posterior."
-        cov_fx_fX = self.kernel(x)
+        is_single_point = x.ndim == 1
+        cov_fx_fX = self.kernel(x) if not is_single_point else self.kernel(x.unsqueeze(0))
```
not522: This conditional branch seems to be equivalent to `atleast_2d`.

```suggestion
        cov_fx_fX = self.kernel(torch.atleast_2d(x))
```

optuna/_gp/gp.py
```diff
@@ -181,13 +188,21 @@ def posterior(self, x: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
...
-        return mean, torch.clamp(var, min=0.0)
+        # K @ inv(C) = V --> K = V @ C --> K = V @ L @ L.T
+        cov_fx_fx_cov_Y_Y_inv = torch.linalg.solve_triangular(
+            self._cov_Y_Y_chol,
+            torch.linalg.solve_triangular(self._cov_Y_Y_chol.T, cov_fx_fX, upper=True, left=False),
+            upper=False,
+            left=False,
+        )
+        var_ = (cov_fx_fx - torch.linalg.vecdot(cov_fx_fX, cov_fx_fx_cov_Y_Y_inv)).clamp_min_(0.0)
+        return (mean.squeeze(0), var_.squeeze(0)) if is_single_point else (mean, var_)
```
not522, line 206:
```suggestion
        return torch.atleast_2d(mean), torch.atleast_2d(var_)
```

c-bata:
@kAIto47802 @not522 Could you review this PR?

optuna/_gp/gp.py
nabenabe0928: This part reduces the dimension augmented above, so the original is inrended.

not522: I see. Thanks.

optuna/terminator/improvement/emmr.py
```diff
@@ -312,7 +312,13 @@ def _posterior_of_batched_theta(
...
     mean = cov_ftheta_fX @ cov_Y_Y_inv_Y
     assert mean.shape == (len_batch,)
-    var = cov_ftheta_ftheta - cov_ftheta_fX @ cov_Y_Y_inv @ cov_ftheta_fX.T
+    V = torch.linalg.solve_triangular(
+        cov_Y_Y_chol,
+        torch.linalg.solve_triangular(cov_Y_Y_chol.T, cov_ftheta_fX, upper=True, left=False),
+        upper=False,
+        left=False,
+    )
+    var = cov_ftheta_ftheta - torch.linalg.vecdot(cov_ftheta_fX, V)
```
not522: The calculation results seem to be changed.

```suggestion
    var = cov_ftheta_ftheta - cov_ftheta_fX @ V.T
```

nabenabe0928: Oh, right.
It seems this part infers the joint posterior of batched points, but not the independent posterior of batched points.

kAIto47802: Isn't it more straightforward, since V correspond to `cov_ftheta_fX @ cov_Y_Y_inv`?
```suggestion
    var = cov_ftheta_ftheta - V @ cov_ftheta_fX.T
```

> [!NOTE]
> Since the output (`cov_ftheta_fX @ cov_Y_Y_inv @ cov_ftheta_fX.T`) is a symmetric matrix, `V @ cov_ftheta_fX.T` and `cov_ftheta_fX @ V.T` is the same and `cov_ftheta_fX @ V.T` is also correct.

nabenabe0928: Thank you two for the comments:)
Just as a future note, I will write it down.
In principle, the master version computes the joint posterior covariance, meaning that it considers the covariance of the posterior between each sampled point.
The `vecdot` version computed the diagonal elements of the covariance matrix.
Hence, `torch.allclose(orig, out.diag())` will be identical.
Btw, since we created the GP class, I will move the joint posterior calculation to the GP class in the follow-up PR.

```python
import torch


torch.manual_seed(0)
for _ in range(5):
    N = 30
    B = 10
    A = torch.randn((N, N), dtype=torch.float64)
    X = A @ A.T + 1.0
    X_inv = torch.linalg.inv(X)
    L = torch.linalg.cholesky(X)
    k = torch.rand((B, N), dtype=torch.float64)
    V = torch.linalg.solve_triangular(
        L,
        torch.linalg.solve_triangular(L.T, k, upper=True, left=False),
        upper=False,
        left=False,
    )
    orig = torch.linalg.vecdot(k, V)
    out = k @ V.T
    out2 = V @ k.T
    ans = k @ X_inv @ k.T
    print(torch.allclose(out, out2), torch.allclose(out, ans), torch.allclose(out2, ans), torch.allclose(orig, out.diag()))
```

nabenabe0928:
Followup
- https://github.com/nabenabe0928/optuna/compare/replace-linalg-inv-with-linalg-cholesky...nabenabe0928:optuna:refactor-emmr?expand=1

not522 (approved):
LGTM!

kAIto47802 (approved):
Thank you for the PR and update!
Indeed, even without any issues in NumPy (BLAS), it is numerically recommended to use Cholesky decomposition instead of explicitly computing the inverse matrix.
LGTM!

PR #5978, opened 2025-02-15
Remove `_StudyInfo`'s `param_distribution` in `_cached_storage.py`

tarunprabhu11:
## Motivation
Hello! I'm new to contributing and looking forward to contribute more.
closes #5968

Removed `_StudyInfo`'s `param_distribution` in `_cached_storage.py`

PR #6301, opened 2025-10-14
Cache distributions to skip consistency check

not522:
## Motivation
`RDBStorage` performs consistency checks with past distributions in `set_trial_param`, but it can be skipped by caching past distributions in `_CachedStorage`.
In the original `_CachedStorage` (#1140), it was implemented, but lost that feature in #4631 and removed in #5978. This PR reintroduces that functionality.

## Description of the changes
- Cache distributions in `_CachedStorage` to skip consistency check

## Benchmark

```python
import optuna

    return x**2 + y**2

sampler = optuna.samplers.TPESampler(seed=42, multivariate=True, constant_liar=True)

```

| master | PR |
| - | - |
| 10.164s | 9.806s |

### DIFF
--- a/optuna/storages/_rdb/storage.py
+++ b/optuna/storages/_rdb/storage.py
@@ -582,10 +582,7 @@ class RDBStorage(BaseStorage, BaseHeartbeat):
         param_value_internal: float,
         distribution: distributions.BaseDistribution,
     ) -> None:
-        with _create_scoped_session(self.scoped_session, True) as session:
-            self._set_trial_param_without_commit(
-                session, trial_id, param_name, param_value_internal, distribution
-            )
+        self._set_trial_param(trial_id, param_name, param_value_internal, distribution, None)
 
     def _set_trial_param(
         self,


tests/storages_tests/test_storages.py
```diff
@@ -579,9 +579,6 @@ def test_set_trial_param(storage_mode: str) -> None:
         # Check the previous call does not change the params.
         with pytest.raises(KeyError):
             storage.get_trial_param(trial_id_2, "y")
-        # State should be checked prior to distribution compatibility.
-        with pytest.raises(UpdateFinishedTrialError):
-            storage.set_trial_param(trial_id_2, "y", 0.4, distribution_z)
```
not522, line 584: This PR changes the order of checks for state and distribution compatibility, causing `ValueError` exceptions to be thrown. Since it's not critical whether `UpdateFinishedTrialError` or `ValueError` is raised in this test case, I'll remove the test.

not522:
The improvement from this PR is proportional to the number of times `set_trial_param` is called, so when the number of parameters is large, the difference becomes even more pronounced as follows:

```python
import optuna

    return sum(trial.suggest_float(f"x{i}", -100, 100) ** 2 for i in range(10))

sampler = optuna.samplers.RandomSampler(seed=42)
```

| master | PR |
| - | - |
| 11.885s | 10.554s |

y0z:
@sawa3030 @kAIto47802 Could you review this PR?

optuna/storages/_rdb/storage.py
```diff
@@ -587,25 +587,60 @@ def set_trial_param(
                 session, trial_id, param_name, param_value_internal, distribution
             )
```
kAIto47802: How about using `_set_trial_params` here, since `set_trial_params` performs the same operation while just hiding the `previous_distribution` argument from the user?
```python
        return self._set_trial_param(trial_id, param_name, param_value_internal, distribution)
```

kAIto47802 (commented):
Thank you for the PR!
It basically LGTM, with only a minor comment :octocat:

optuna/storages/_rdb/storage.py
not522: Thank you. I've updated as suggested. PTAL.

kAIto47802 (approved):
Thank you for the update! LGTM

PR #6303, opened 2025-10-16
Explicitly close DB connections when discarding SQLAlchemy's `Engine`

not522:
## Motivation
Recently, random failures in CI have been occurring frequently (e.g., https://github.com/optuna/optuna/issues/5797#issuecomment-3399807512). This may be caused by connection objects created by SQLAlchemy not being properly released (see https://docs.sqlalchemy.org/en/20/core/connections.html#engine-disposal).
To call `Engine.dispose()` after each unit test will resolve this issue. However, modifying all relevant tests would be extensive work, so this PR limits its scope to modifications to the `StorageSupplier`. This should result in most tests being fixed, thereby reducing the likelihood of CI failures.

From my understanding, this issue only occurs when creating `Engine` objects frequently—like in CI scenarios. It doesn't happen in normal user usage.

The master branch ran the CI 10 times, with one failure (attempt 7 at https://github.com/not522/optuna/actions/runs/18522577119). Meanwhile, this PR never failed for this error (https://github.com/not522/optuna/actions/runs/18548728266; attempt 5 failed for a different reason).

## Description of the changes

- Explicitly close DB connections in `StorageSupplier`

### DIFF
--- a/optuna/testing/storages.py
+++ b/optuna/testing/storages.py
@@ -86,6 +86,8 @@ class StorageSupplier(AbstractContextManager):
         self.server: grpc.Server | None = None
         self.thread: threading.Thread | None = None
         self.proxy: GrpcStorageProxy | None = None
+        self.storage: BaseStorage | None = None
+        self.backend_storage: BaseStorage | None = None
 
     def __enter__(
         self,
@@ -99,7 +101,7 @@ class StorageSupplier(AbstractContextManager):
         if self.storage_specifier == "inmemory":
             if len(self.extra_args) > 0:
                 raise ValueError("InMemoryStorage does not accept any arguments!")
-            return optuna.storages.InMemoryStorage()
+            self.storage = optuna.storages.InMemoryStorage()
         elif "sqlite" in self.storage_specifier:
             self.tempfile = NamedTemporaryFilePool().tempfile()
             url = "sqlite:///{}".format(self.tempfile.name)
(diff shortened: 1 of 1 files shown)


y0z:
@c-bata @sawa3030 Could you review this PR?

c-bata (approved):
Changes themselves look good to me.

According to the Python 3.13 release note, the sqlite3 module started emitting ResourceWarning from version 3.13. Some failed tests such as [this workflow run](https://github.com/not522/optuna/actions/runs/18522577119/attempts/7) and [this one](https://github.com/optuna/optuna/actions/runs/18483881520/job/52663519328) are using Python 3.13.

> A [ResourceWarning](https://docs.python.org/3/library/exceptions.html#ResourceWarning) is now emitted if a [Connection](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection) object is not [closed](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.close) explicitly. (Contributed by Erlend E. Aasland in [gh-105539](https://github.com/python/cpython/issues/105539).)
> https://docs.python.org/3/whatsnew/3.13.html#sqlite3

However, following recent workflow runs uses Python 3.8 or 3.12, so I would say we still need to investigate those errors.

* https://github.com/optuna/optuna/actions/runs/18488888348/job/52677711048?pr=6300
* https://github.com/optuna/optuna/actions/runs/18512431531/job/52756077902

Anyway, thank you for your pull request. LGTM.

PR #6312, opened 2025-10-22
Changed the old `.format` code to the new f string format in the `test_journal.py`

Zrahay:
**Motivation**

Now that Optuna has dropped the support for Python 3.8 we can use modern f string syntax in the codebase which replaces the old .format method.

**Description Of The Changes**

Replaced the occurence of the old '.format()' formatting technique with the new f string technique in the test_journal.py file. The file is present at the location tests/storage_tests/journal_tests/test_journal.py.

No other functional changes were made in this.

### DIFF
--- a/tests/storages_tests/journal_tests/test_journal.py
+++ b/tests/storages_tests/journal_tests/test_journal.py
@@ -70,7 +70,7 @@ class JournalLogStorageSupplier:
             journal_redis_storage._redis = FakeStrictRedis()
             return journal_redis_storage
         else:
-            raise RuntimeError(f"Unknown Log Storage Type: {self.storage_type}")
+            raise RuntimeError(f"Unknown log storage type: {self.storage_type}")
 
     def __exit__(
         self, exc_type: type[BaseException], exc_val: BaseException, exc_tb: TracebackType


Zrahay:
Hey there! If there's any sort of ambiguity or errors in my code or maybe if I missed some important step, please do let me know. I'll fix that as soon as possible.

Cheers!

tests/storages_tests/journal_tests/test_journal.py
```diff
@@ -70,7 +70,7 @@ def __enter__(self) -> optuna.storages.journal.BaseJournalBackend:
             journal_redis_storage._redis = FakeStrictRedis()
             return journal_redis_storage
         else:
-            raise RuntimeError("Unknown log storage type: {}".format(self.storage_type))
+            raise RuntimeError(f"Unknown Log Storage Type: {self.storage_type}")
```
not522: Thank you for your PR! Could you keep the behavior as it was before?

```suggestion
            raise RuntimeError(f"Unknown log storage type: {self.storage_type}")
```

Zrahay: Hey there! Kinda confused as to what do you mean by keeping the behaviour as it was before? Do you want me to change the technique back to the .format() method?

I'm sorry if it seems a pretty bad question, I'm new to open source so can't understand what did you mean by this.

Thanks!

not522: Could you leave the error message as is? That is, stop using capitalize. Focus this PR's changes solely on changing from `format` to f-string.

c-bata (approved):
LGTM.

Issue #6305, opened 2025-10-17
Use f-string and `{var_name=}` instead of `.format` and `var_name={var_name}`


nabenabe0928:
> [!NOTE]
> Please also take a look at the other good-first issues as well:
> - https://github.com/optuna/optuna/issues/4508
> - https://github.com/optuna/optuna/issues/6029

### Motivation

Now that Optuna does not support Python 3.8 anymore, we can use the `{var_name}=` syntax.
Since it is cleaner, I recommend using it instead of the good old style.

> [!IMPORTANT]
> To promote more contributions from a wide range of developers, please address only one file per PR:)

### Suggestion

Please fix one of the files below:
- https://github.com/search?q=repo%3Aoptuna%2Foptuna+%22.format&type=code&p=2

Example:

```python
## Current
message = "The `value` argument is of type '{}' but supposed to be a float.".format(
    type(value).__name__
)

## Preferred
message = f"The `value` argument is of type '{type(value).__name__}' but supposed to be a float."

## Current
if reduction_factor < 2:
    raise ValueError("`reduction_factor >= 2` must hold but got reduction_factor={}.".format(reduction_factor))

## Preferred
if reduction_factor < 2:
    raise ValueError(f"`reduction_factor >= 2` must hold but got {reduction_factor=}.")
```

nabenabe0928:
More Concrete Examples: 
- https://github.com/optuna/optuna/pull/6306
- https://github.com/optuna/optuna/pull/6307

unKnownNG:
Hi! Thank you for merging my previous PR. Since I’ve already contributed one file as requested, I just wanted to mention that if no one else contributes the remaining files and the issue deadline is approaching, I’d be happy to help by working on the other files as well. Please let me know if that would be helpful.

Thank you!

Zrahay:
Hey there! I'm working on this issue, specifically on the test_journal.py file.😊

Cheers!

SoJaoAnant:
Hey! I'm new here, I don't know if i should get a permission to contribute to this issue but I will work on this issue, it sounds fun 😋

dross20:
I can take `optuna/visualization/_pareto_front.py`!

satyarth7srivastava:
Hello this is my first time contributing to a project so I'm starting with the small file 'optuna/tutorial/10_key_features/001_first.py'

nihalsiddiqui7:
[...]

PR #6320, opened 2025-10-29
Refactor: Replace .format() with f-strings in _frozen.py

Jongwan93:
## Motivation

Following the recommendation in #6305, I'm replacing the old-style `.format()` string formatting with the cleaner f-string syntax. 

## Description of the changes

- Replaced `.format()` calls with f-strings in the `__repr__` method
- Used `!r` specifier for `repr()` calls to follow f-string conventions
- Maintained the exact same functionality and output format


nabenabe0928:
@Jongwan93 
Hey, I found it less readable because of the embedded string in each `{}`, so would you mind trying out another file and closing this PR?

Jongwan93:
@nabenabe0928 okay Thank you for the review

PR #6330, opened 2025-11-07
Dispose SQLAlchemy's engine in `storage_tests/test_with_server.py`

kAIto47802:
## Motivation

This PR is the follow-up on:

and runs in parallel with::

- https://github.com/optuna/optuna/pull/6337, and
- https://github.com/optuna/optuna/pull/6338.

The original PR did not cover all test cases. This PR addresses the remaining ones in `storage_tests/test_with_server.py`.

> [!IMPORTANT]

## Description of the changes

- Make the `get_storage` funciton contextmanager and use it for all tests in `storage_tests/test_with_server.py`.

> [!NOTE]

> [!IMPORTANT]

### DIFF
--- a/tests/storages_tests/test_with_server.py
+++ b/tests/storages_tests/test_with_server.py
@@ -58,10 +58,11 @@ def get_storage() -> Generator[BaseStorage, None, None]:
     else:
         assert False, f"The mode {storage_mode} is not supported."
 
-    yield storage
-
-    if isinstance(storage, optuna.storages.RDBStorage):
-        storage.engine.dispose()
+    try:
+        yield storage
+    finally:
+        if isinstance(storage, optuna.storages.RDBStorage):
+            storage.engine.dispose()
 
 
 def run_optimize(study_name: str, n_trials: int) -> None:


not522:
@Alnusjaponica @sawa3030 Could you review this PR?

tests/storages_tests/test_with_server.py
```diff
@@ -55,15 +58,18 @@ def get_storage() -> BaseStorage:
     else:
         assert False, f"The mode {storage_mode} is not supported."
 
-    return storage
+    yield storage
+
+    if isinstance(storage, optuna.storages.RDBStorage):
+        storage.engine.dispose()
```
sawa3030:
```suggestion
    try:
        yield storage
    finally:
        if isinstance(storage, optuna.storages.RDBStorage):
            storage.engine.dispose()
```

How about using a `try`-`finally` block here to ensure that the connection is properly closed, as recommended in [the official documentation](https://docs.python.org/3.13/library/contextlib.html#contextlib.contextmanager)?.

kAIto47802:
Thank you for the comment. I’ve updated the code accordingly.
(I noticed a typo in the branch name and tried to fix it, but the PR was automatically closed during the process, so I have reopened it.)

Alnusjaponica (approved):
Thanks for your contribution. LGTM.

PR #6337, opened 2025-11-12
Dispose SQLAlchemy's engine in `storage_tests/test_cached_storage.py`

kAIto47802:
## Motivation

This PR is the follow-up on:

and runs in parallel with::

- https://github.com/optuna/optuna/pull/6330, and
- https://github.com/optuna/optuna/pull/6338.

The original PR did not cover all test cases. This PR addresses the remaining ones in `storage_tests/test_cached_storage.py`.

> [!IMPORTANT]

## Description of the changes

- Use `RDBStorage` within a context manager so that `storage.engine.dispose()` is executed every time.

> [!NOTE]

> [!IMPORTANT]

### DIFF
--- a/tests/storages_tests/test_cached_storage.py
+++ b/tests/storages_tests/test_cached_storage.py
@@ -172,7 +172,6 @@ def test_unfinished_trial_ids() -> None:
             create_rdb_storage(storage_url) as base_storage1,
             create_rdb_storage(storage_url) as base_storage2,
         ):
-
             storage1 = _CachedStorage(base_storage1)
             study1 = optuna.create_study(
                 study_name=study_name,


y0z:
@fusawa-yugo Could you review this PR?

tests/storages_tests/test_cached_storage.py
```diff
@@ -10,43 +14,54 @@
...
 
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+
+@contextmanager
+def create_rdb_storage(url: str = "sqlite:///:memory:") -> Generator[RDBStorage, None, None]:
+    storage = RDBStorage(url)
+    yield storage
+    storage.engine.dispose()
```
fusawa-yugo:
```suggestion
    try:
        yield storage
    finally:
        storage.engine.dispose()
```

it would be better to change this here for readability and consistency as mentioned in https://github.com/optuna/optuna/pull/6330#discussion_r2521966118.

c-bata:

sawa3030:
It’s basically LGTM. I’d appreciate it if you could update the part pointed out [here](https://github.com/optuna/optuna/pull/6337#discussion_r2526595837).

kAIto47802:
Thank you for the comment. I've update the code accordingly. PTAL :octocat:

sawa3030 (approved):

fusawa-yugo:
LGTM.
Thank you!

PR #6338, opened 2025-11-12
Dispose SQLAlchemy's engine in `storage_tests/rdb_tests/test_storage.py`

kAIto47802:
## Motivation

This PR is the follow-up on:

and runs in parallel with::

- https://github.com/optuna/optuna/pull/6330, and
- https://github.com/optuna/optuna/pull/6337.

The original PR did not cover all test cases. This PR addresses the remaining ones in `storage_tests/rdb_tests/test_storage.py`.

> [!IMPORTANT]

## Description of the changes

- Make existing `create_test_storage` context manager so that `storage.engine.dispose()` is executed every time.
- Make sure to use `create_test_storage` for all tests.

> [!NOTE]

> [!IMPORTANT]

### DIFF
--- a/tests/storages_tests/rdb_tests/test_storage.py
+++ b/tests/storages_tests/rdb_tests/test_storage.py
@@ -52,8 +52,10 @@ def create_test_storage(
         skip_compatibility_check=skip_compatibility_check,
         skip_table_creation=skip_table_creation,
     )
-    yield storage
-    storage.engine.dispose()
+    try:
+        yield storage
+    finally:
+        storage.engine.dispose()
 
 
 def test_init() -> None:


y0z:
@fusawa-yugo Could you review this PR?

tests/storages_tests/rdb_tests/test_storage.py
```diff
@@ -123,50 +129,62 @@ def test_set_default_engine_kwargs_for_mysql_with_other_rdb() -> None:
...
+    skip_table_creation: bool = False,
+) -> Generator[RDBStorage, None, None]:
+    storage = RDBStorage(
+        url,
+        engine_kwargs=engine_kwargs,
+        skip_compatibility_check=skip_compatibility_check,
+        skip_table_creation=skip_table_creation,
+    )
+    yield storage
+    storage.engine.dispose()
```
fusawa-yugo:
```suggestion
    try:
        yield storage
    finally:
        storage.engine.dispose()
```
it would be better to add `try` and `finally` for readability. (ref: https://github.com/optuna/optuna/pull/6330#discussion_r2521966118)

fusawa-yugo:
could you please use `contextmanager` here?
https://github.com/optuna/optuna/blob/55e08db1683fd79fd770017d867159da51ec82ba/tests/storages_tests/rdb_tests/test_storage.py#L102

c-bata:

tests/storages_tests/rdb_tests/test_storage.py
sawa3030: This might be slightly outside the scope of this PR, but could you move the function `create_test_storage` to the top of the file, before `test_init`? Since it’s used in the very first test (`test_init`), I think it would make the file a bit easier to follow.

kAIto47802: Thank you for the comment.
Indeed, I had been thinking about this as well, but I avoided including it in this PR to keep the diff minimal.
However, I completely agree that this function should be moved to the top of the file, so I will apply this change as you suggested. I also think it is unlikely that a standalone PR whose sole purpose is to move this function will be created in the future.

kAIto47802:
Thank you for the comment. I've update the code accordingly. PTAL :octocat:

fusawa-yugo:
LGTM.
Thank you!

PR #6341, opened 2025-11-14
Migrate to ruff from black/blackdoc/isort/flake8

sawa3030:
## Motivation
Adopt Ruff and drop Black / Blackdoc / isort / Flake8.

## Summary of Changes
- Add Ruff configuration to `pyproject.toml` and remove settings for Black, Blackdoc, isort, and Flake8.
- Update `format.sh`, `.pre-commit-config.yaml`, and `CONTRIBUTING.md` to use Ruff.
- Update CI jobs (check and check-optional) to run Ruff.
- Apply minimal code changes where Ruff’s rules differ from the previous toolchain.

## Notes
- Some minor formatter differences are temporarily ignored via `[tool.ruff.format]`. This is because Ruff’s behavior is not 100% identical to Black/Blackdoc.
- To keep this PR focused, we’re not addressing all remaining nits. We’ll track and clean these up in a follow-up issue/PR.

### DIFF
--- a/optuna/samplers/_brute_force.py
+++ b/optuna/samplers/_brute_force.py
@@ -111,7 +111,7 @@ class BruteForceSampler(BaseSampler):
     Example:
 
         .. testcode::
-            ``` python
+
             import optuna
 
 
@@ -127,7 +127,7 @@ class BruteForceSampler(BaseSampler):
 
             study = optuna.create_study(sampler=optuna.samplers.BruteForceSampler())
             study.optimize(objective)
-            ```
+
     Note:
         The defined search space must be finite. Therefore, when using
         :class:`~optuna.distributions.FloatDistribution` or


gen740 (commented):
Thank you for the PR!

I have a question.

Is that OK to remove `blackdoc`? `ruff` cannot substitute `blackdoc`.

c-bata:
> Is that OK to remove blackdoc?

Yes, that’s ok. We previously encountered some issues with blackdoc (though I don’t remeber the exact details), and we had discussed removing it back then.

not522:
> Is that OK to remove blackdoc? ruff cannot substitute blackdoc.

In my understanding, `ruff` checks docstrings if `docstring-code-format = true`.

sawa3030:
Thank you for the discussion. From my testing, Ruff’s docstring formatter does not recognize ``.. testcode::`` directives as Python code to format (ref: the official documentation: https://docs.astral.sh/ruff/formatter/#docstring-formatting). If we use code blocks inside the `.. testcode::` directive like [here](https://github.com/optuna/optuna/pull/6341/commits/ca7d43cf35c1ad9e5a446b1e869de1240a0aa8fb), Ruff does format them, but then the Sphinx doctests fail (example: https://github.com/optuna/optuna/actions/runs/19525035471/job/55896143443?pr=6341).

Given these findings, I think it might be reasonable to drop blackdoc for now and wait until Ruff can better support Sphinx-style doctest blocks. If there is any better ideas, I would really appreciate your suggestions.

c-bata:

not522 (approved):
LGTM!

PR #6346, opened 2025-11-18
Fix type checking and GP integration for MDI importance

davidonthethird:
## Motivation

This PR is a partial contribution toward #6029. The goal of that issue is to move type-only imports behind `typing.TYPE_CHECKING` so we avoid circular imports and unnecessary runtime dependencies, while keeping static type checking working correctly.

## Description of the changes

This PR focuses on the following modules:

- `optuna/_gp/acqf.py`
- `optuna/_gp/optim_sample.py`
- `optuna/importance/_mean_decrease_impurity.py`
- `optuna/importance/_ped_anova/evaluator.py`
- `optuna/importance/_ped_anova/scott_parzen_estimator.py`

For these files I:

- Moved imports that are only used for type annotations into `if TYPE_CHECKING:` blocks.
- Updated type annotations to use forward references / fully qualified names where needed so they still work with the moved imports.
- Kept runtime behavior unchanged (imports that are actually needed at runtime stay at the top level).

These changes are intended to reduce the risk of circular imports and bring these modules in line with the pattern suggested in #6029.

## Testing

Locally I ran:

```bash
pytest -q tests/gp_tests/test_gp.py::test_fit_kernel_params -k "not slow" --maxfail=1

### DIFF
--- a/optuna/_gp/acqf.py
+++ b/optuna/_gp/acqf.py
@@ -311,5 +311,5 @@ class ConstrainedLogEHVI(BaseAcquisitionFunc):
     def eval_acqf(self, x: torch.Tensor) -> torch.Tensor:
         constraints_acqf_values = sum(acqf.eval_acqf(x) for acqf in self._constraints_acqf_list)
         if self._acqf is None:
-            return cast(torch.Tensor, constraints_acqf_values)
+            return cast("torch.Tensor", constraints_acqf_values)
         return constraints_acqf_values + self._acqf.eval_acqf(x)
--- a/optuna/_gp/optim_sample.py
+++ b/optuna/_gp/optim_sample.py
@@ -1,12 +1,14 @@
 from __future__ import annotations
 
 import numpy as np
+from typing import TYPE_CHECKING
 
-from optuna._gp import acqf as acqf_module
+if TYPE_CHECKING:
+    from optuna._gp.acqf import BaseAcquisitionFunc
 
 
 def optimize_acqf_sample(
-    acqf: acqf_module.BaseAcquisitionFunc,
+    acqf: "BaseAcquisitionFunc",
     *,
     n_samples: int = 2048,
     rng: np.random.RandomState | None = None,
--- a/optuna/importance/_ped_anova/evaluator.py
+++ b/optuna/importance/_ped_anova/evaluator.py
@@ -1,21 +1,26 @@
 from __future__ import annotations
 
-from collections.abc import Callable
+from typing import TYPE_CHECKING
 
 import numpy as np
 
 from optuna._experimental import experimental_class
 from optuna._warnings import optuna_warn
-from optuna.distributions import BaseDistribution
-from optuna.importance._base import _get_distributions
-from optuna.importance._base import _get_filtered_trials
-from optuna.importance._base import _sort_dict_by_importance
-from optuna.importance._base import BaseImportanceEvaluator
+from optuna.importance._base import (
+    _get_distributions,
+    _get_filtered_trials,
+    _sort_dict_by_importance,
+    BaseImportanceEvaluator,
+)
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: optuna/importance/_ped_anova/scott_parzen_estimator.py [hunk 1]
```diff
@@ -10,7 +11,9 @@
 from optuna.samplers._tpe.parzen_estimator import _ParzenEstimatorParameters
 from optuna.samplers._tpe.probability_distributions import _BatchedDiscreteTruncNormDistributions
 from optuna.samplers._tpe.probability_distributions import _BatchedDistributions
-from optuna.trial import FrozenTrial
+
+if TYPE_CHECKING:
+    from optuna.trial import FrozenTrial
```
kAIto47802, line 16: We can move `BaseDistribution` and `_BatchedDistributions` here, since they are also only used for type annotations.
```suggestion
    from optuna.trial import FrozenTrial
    from optuna.distributions import BaseDistribution
    from optuna.samplers._tpe.probability_distributions import _BatchedDistributions
```

c-bata:
@kAIto47802 @Alnusjaponica Could you review this PR?

kAIto47802:
Thank you for the PR. I left a comment. PTAL.
Also, could you please apply the formatter to fix the import section and fix the CI issue?
You can do this by running `./formats.sh` at the root of the Optuna directory.

optuna/_gp/optim_sample.py
```diff
@@ -1,12 +1,14 @@
 from __future__ import annotations
 
 import numpy as np
+from typing import TYPE_CHECKING
 
-from optuna._gp import acqf as acqf_module
+if TYPE_CHECKING:
+    from optuna._gp.acqf import BaseAcquisitionFunc
```
nabenabe0928, line 7:
```suggestion
    from optuna._gp import acqf as acqf_module
```

optuna/_gp/optim_sample.py
```diff
@@ -1,12 +1,14 @@
...
 
 
 def optimize_acqf_sample(
-    acqf: acqf_module.BaseAcquisitionFunc,
+    acqf: "BaseAcquisitionFunc",
```
nabenabe0928, line 11: Could you revert here?

```suggestion
    acqf: acqf_module.BaseAcquisitionFunc,
```

optuna/importance/_ped_anova/evaluator.py
```diff
@@ -1,21 +1,26 @@
...
-from optuna.importance._base import _get_distributions
-from optuna.importance._base import _get_filtered_trials
-from optuna.importance._base import _sort_dict_by_importance
-from optuna.importance._base import BaseImportanceEvaluator
+from optuna.importance._base import (
+    _get_distributions,
+    _get_filtered_trials,
+    _sort_dict_by_importance,
+    BaseImportanceEvaluator,
+)
```
nabenabe0928, line 14: This import syntax is prohibited by the formatter.
Please import them one by one as in the original code.

optuna/_gp/optim_sample.py
Alnusjaponica: Should be enclosed in double quotation marks?
```suggestion
    acqf: "acqf_module.BaseAcquisitionFunc",
```

optuna/_gp/acqf.py
```diff
@@ -311,5 +311,5 @@ def __init__(
     def eval_acqf(self, x: torch.Tensor) -> torch.Tensor:
         constraints_acqf_values = sum(acqf.eval_acqf(x) for acqf in self._constraints_acqf_list)
         if self._acqf is None:
-            return cast(torch.Tensor, constraints_acqf_values)
+            return cast("torch.Tensor", constraints_acqf_values)
```
Alnusjaponica, line 314: May I ask what caused this change?

c-bata:

PR #6362, opened 2025-11-28
Add support for local HPI in PED-ANOVA

kAIto47802:
## Motivation

While [the original PED-ANOVA paper](https://arxiv.org/abs/2304.10255) introduces a local HPI that enables computing HPI in top-performing regions, the current implementation does not support this feature and only provides global HPI.

## Description of the changes

- Introduce the `target_trials` and `region_quantile` arguments, which correspond to $\gamma'$ and $\gamma$ in the original paper, respectively.
- Deprecate the `baseline_quantile` argument, which was originally used for $\gamma$.

### DIFF
--- a/optuna/importance/_ped_anova/evaluator.py
+++ b/optuna/importance/_ped_anova/evaluator.py
@@ -18,7 +18,6 @@ from optuna.study import StudyDirection
 from optuna.trial import FrozenTrial
 
 
-
 class _QuantileFilter:
     def __init__(
         self,


c-bata:
@y0z @sawa3030 Could you review this PR?

optuna/importance/_ped_anova/evaluator.py
```diff
@@ -121,11 +145,25 @@ def objective(trial):
...
+        target_quantile: float = 0.1,  # gamma' in the original paper
+        region_quantile: float = 1.0,  # gamma in the original paper
+        baseline_quantile: float | None = None,
         evaluate_on_local: bool = True,
-    ):
-        assert 0.0 <= baseline_quantile <= 1.0, "baseline_quantile must be in [0, 1]."
-        self._baseline_quantile = baseline_quantile
+    ) -> None:
+        assert 0.0 <= target_quantile <= 1.0, "`target_quantile` must be in [0, 1]."
+        assert 0.0 < region_quantile <= 1.0, "`region_quantile` must be in (0, 1]."
```
y0z:
```suggestion
        assert 0.0 < target_quantile < region_quantile <= 1.0, "condition 0.0 < `target_quantile` < `region_quantile` <= 1.0 must be satisfied"
```

optuna/importance/_ped_anova/evaluator.py
```diff
@@ -65,30 +66,53 @@ class PedAnovaImportanceEvaluator(BaseImportanceEvaluator):
...
+            trials achieve top-50% performance. If ``region_quantile=1.0``, the importance is
+            computed in the whole search space.
+
         baseline_quantile:
             Compute the importance of achieving top-``baseline_quantile`` quantile objective value.
             For example, ``baseline_quantile=0.1`` means that the importances give the information
             of which parameters were important to achieve the top-10% performance during
             optimization.
+
+            .. warning::
```
y0z, line 106: Since this feature is experimental, we can remove this argument without a transition period. At least, we can remove this in Optuna v5.

optuna/importance/_ped_anova/evaluator.py
```diff
@@ -65,30 +66,53 @@ class PedAnovaImportanceEvaluator(BaseImportanceEvaluator):
...
+                Deprecated in v4.7.0. This feature will be removed in the future. The removal of
+                this feature is currently scheduled for v0.6.0, but this schedule is subject to
```
y0z: Nit.
```suggestion
                this feature is currently scheduled for v6.0.0, but this schedule is subject to
```

optuna/importance/_ped_anova/evaluator.py
y0z: Based on my understanding, allowing `target_quantile = 0.0` does not yield any reasonable result since this setting means there is no sample in the target region.

optuna/importance/_ped_anova/evaluator.py
y0z, line 45: When `q = 0`, `cuttoff_index` becomes `-1`, and this function will return the maximum of `loss_vals`. Is there a bug?

optuna/importance/_ped_anova/evaluator.py
```diff
@@ -153,7 +191,7 @@ def _get_top_trials(
             is_lower_better = True
 
         top_trials = _QuantileFilter(
-            self._baseline_quantile, is_lower_better, self._min_n_top_trials, target
+            quantile, is_lower_better, self._min_n_top_trials, target
         ).filter(trials)
 
         if len(trials) == len(top_trials):
```
y0z: Could this situation happen even with proper setting assertions and L259 (i.e., `trials if self._region_quantile == 1.0 else ...`)?

optuna/importance/_ped_anova/evaluator.py
y0z: `_get_top_quantile_trials` may be better.

tests/importance_tests/pedanova_tests/test_evaluator.py
```diff
@@ -72,7 +72,7 @@ def test_n_trials_equal_to_min_n_top_trials() -> None:
 def test_baseline_quantile_is_1() -> None:
```
y0z: Please rename test properly.

tests/importance_tests/pedanova_tests/test_evaluator.py
```diff
@@ -90,8 +90,8 @@ def test_direction() -> None:
 
 def test_baseline_quantile() -> None:
```
y0z: Please rename test properly. And, it is better to add some tests with `region_quantile < 1.0`.

optuna/importance/_ped_anova/evaluator.py
kAIto47802: Thank you for the comment.
Indeed, no, this condition no longer happens after we add the assertion `assert 0.0 < target_quantile < region_quantile <= 1.0`, which guarantees that `target_quantile` is strictly less than 1.0.

I also noticed that this warning message is no longer appropriate for the updated implementation, since it would still be raised even when `region_quantile` is 1.0, which is the default setting for global HPI.

optuna/importance/_ped_anova/evaluator.py
kAIto47802: Okay, then I'll rename it.

tests/importance_tests/pedanova_tests/test_evaluator.py
kAIto47802: Oh, i didn't notice that. Thank you for the nice catch!

tests/importance_tests/pedanova_tests/test_evaluator.py
kAIto47802: Thank you for the comment.
I'll add some tests for `region_quantile`

optuna/importance/_ped_anova/evaluator.py
kAIto47802: The original implementation allows the condition `baseline_quantile == 1` (which corresponds to `target_quantile`), also including a test for this condition.
Do we remove this condition and restrict it to be strictly less than 1?
(This condition is not so meaningful, so I think restricting it would be a reasonable choice.)

optuna/importance/_ped_anova/evaluator.py
kAIto47802: It seems to be a bug in the original implementation.
After we restrict the target quantile to be strictly greater than 0, this would not happen.

kAIto47802: Maybe we can update the assertion for the `quantile` argument in the constructor of `_QuantileFilter` .

optuna/importance/_ped_anova/evaluator.py
kAIto47802: Oh, thank you for the nice catch!

optuna/importance/_ped_anova/evaluator.py
```diff
@@ -213,13 +251,18 @@ def evaluate(
...
 
-        top_trials = self._get_top_trials(study, trials, params, target)
-        quantile = len(top_trials) / len(trials)
+        target_trials = self._get_top_quantile_trials(study, trials, self._target_quantile, target)
+        region_trials = (
+            trials
+            if self._region_quantile == 1.0
+            else self._get_top_quantile_trials(study, trials, self._region_quantile, target)
+        )
+        quantile = len(target_trials) / len(region_trials)
```
sawa3030, line 258:
```suggestion
        if len(target_trials) == len(region_trials):
            _logger.warning(
                "Target and region quantiles select the same set of trials. "
                "Parameter importances will be equal."
            )
        quantile = len(target_trials) / len(region_trials)
```

Even if `target_quantile` < `region_quantile` holds, it is still possible that `len(target_trials) == len(region_trials)` when the two quantiles are very close, especially with a small number of trials. So how about implementing the warning, which used to be available in `_get_top_trials`?

optuna/importance/_ped_anova/evaluator.py
```diff
@@ -28,8 +29,8 @@ def __init__(
         is_lower_better: bool,
         min_n_top_trials: int,
         target: Callable[[FrozenTrial], float] | None,
-    ):
-        assert 0 <= quantile <= 1, "quantile must be in [0, 1]."
+    ) -> None:
+        assert 0 < quantile <= 1, "quantile must be in [0, 1]."
```
sawa3030:
```suggestion
        assert 0 < quantile <= 1, "quantile must be in (0, 1]."
```
nit

sawa3030 (commented):
I've left some minor comments. PTAL!

nabenabe0928:
If I remember the context correctly, we, including @HideakiImamura , intentionally removed this functionality to reduce the complexity.
The discussion at the time came to the agreement such that we add this feature on request.

optuna/importance/_ped_anova/evaluator.py
kAIto47802: Thank you for the comment! I'll incorporate this suggestion

optuna/importance/_ped_anova/evaluator.py
kAIto47802: Does it mean that we are removing this argument immediately in this PR or setting the removal version to v5.0.0?

kAIto47802:
Thank you for the comment. I've updated the code accordingly.

sawa3030 (approved):
LGTM

PR #6363, opened 2025-11-28
Remove redundant `_color_supported()` check.

gen740:
## Motivation

The logging module `colorlog` already supports the `NO_COLOR` environment variable since https://github.com/borntyping/python-colorlog/releases/tag/v6.0.0-alpha.2. Therefore, checking and switching the Formatter in Optuna code is unnecessary.

Additionally, `colorlog` supports the `FORCE_COLOR` option, which forces colored outputs regardless of the environment. This is particularly useful when running Optuna in GitHub CI or other environments that don’t have a tty device but support colored outputs. However, in the current code, when the user provides the `FORCE_COLOR` environment variable, Optuna ignores it, resulting in colorless output.

## Description of the changes

- Remove the `_color_supported()` function since the `colorlog` module itself can handle it.

### DIFF
--- a/.github/workflows/tests-with-minimum-versions.yml
+++ b/.github/workflows/tests-with-minimum-versions.yml
@@ -70,7 +70,7 @@ jobs:
       run: |
         # Install dependencies with minimum versions.
         pip uninstall -y alembic cmaes packaging sqlalchemy plotly scikit-learn pillow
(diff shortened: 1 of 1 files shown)


gen740:
I previously submitted the pull request #4376 to support non-tty devices with no color. I also introduced the `NO_COLOR` and `isatty` logic. However, the correct approach is to specify the `stream = sys.stderr` to the `colorlog.ColoredFormatter`.

c-bata:

c-bata (approved):
LGTM!

Issue #6365, opened 2025-11-30
Get trial by id


PabloRuizCuevas:
### Motivation

for getting a trial by id the only public API is something like:

```python

trials = study.get_trials()
trial = next(t for t in trials if t.number == trial_id)

```

Which is excruciatingly slow, of course you can pass the argument to avoid deep copying but still it seems that there is a better way using private methods:

``` python
internal_id = study._storage.get_trial_id_from_study_id_trial_number(study._study_id,trial_id)
trial = study._storage.get_trial(internal_id)
```

My proposal is add to study get_trials_by_id that will return exactly this

### Description

My proposal is add to study get_trials_by_id that will return the trial by the public id.

### Alternatives (optional)

leaving it as is

not522:
Let me close this issue for the same reason as #6366.
As a side note, I think you're confusing `trial_id` and `trial_number`. `trial_id` is an ID used internally by Optuna, while `trial_number` is the ID intended for user use.

PR #6366, opened 2025-11-30
[FEAT] Add new method to retrieve trials by number and include its unit tests

WenhaoYou1:
## Motivation

Currently, retrieving a trial by its number requires either fetching all trials and iterating through them, or using private storage methods. The public API approach is inefficient for studies with many trials, while the private method approach is not recommended for users.

This PR adds a public, efficient method `get_trial_by_number()` to address this gap in the API.

## Description of Changes

Added a new public method `Study.get_trial_by_number()` that efficiently retrieves a trial by its number without fetching all trials from storage. We also commented with the same style of current codebase. This addresses the performance issue where users previously had to either:
- Call `get_trials()` and iterate through all trials, or
- Use private storage methods

The new method:
- Uses existing storage backend methods (`get_trial_id_from_study_id_trial_number()` and `get_trial()`)
- Supports `deepcopy` parameter for consistency with other Study methods
- Raises `KeyError` when trial number doesn't exist
- Works across all storage backends (in-memory, RDB, Journal, Redis)

**Modified files:**
- `optuna/study/study.py`: Added `get_trial_by_number()` method
- `tests/study_tests/test_study.py`: Added comprehensive test coverage

## Description of Tests

Added `test_get_trial_by_number()` parameterized across all storage modes:
- Validates correct trial retrieval by number
- Verifies `deepcopy` parameter behavior (using mocks)
- Confirms `KeyError` is raised for non-existent trial numbers
- Tests compatibility with all storage backends

**Test results:** 5 passed, 2 skipped (gRPC storage backends that don't use `copy.deepcopy`)

All existing tests pass without any regression.

## Performance Comparison

| Method | Time per call | Speedup |
|--------|--------------|---------|
| `get_trials()` + iteration | 13.20 ms | baseline |
| `get_trial_by_number()` | 0.01 ms | **~1000x faster** |

**Hardware:** Apple M3 Pro, 18 GB RAM

The dramatic improvement comes from avoiding the overhead of fetching and deserializing all trials when only one is needed.

## Related Issue

#6365

### DIFF
--- a/optuna/study/study.py
+++ b/optuna/study/study.py
@@ -266,6 +266,52 @@ class Study:
         """
         return self._get_trials(deepcopy, states, use_cache=False)
 
+    def get_trial_by_number(self, trial_number: int, deepcopy: bool = True) -> FrozenTrial:
+        """Return a trial by its number.
+
+        This method efficiently retrieves a specific trial by its number without
+        fetching all trials in the study.
+
+        Example:
+            .. testcode::
+
+                import optuna
+
+
+                def objective(trial):
+                    x = trial.suggest_float("x", -1, 1)
+                    return x**2
+
+
+                study = optuna.create_study()
+                study.optimize(objective, n_trials=3)
+
+                # Get trial number 1
+                trial = study.get_trial_by_number(1)
+                assert trial.number == 1
+
+        Args:
+            trial_number:
+                The number of the trial to retrieve.
+            deepcopy:
+                Flag to control whether to apply ``copy.deepcopy()`` to the trial.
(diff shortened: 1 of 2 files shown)


c-bata (changes requested):
Thank you for your suggestion.
After considering it, we decided not to add get_trial_by_number to the Study class. We generally try to avoid expanding the Study API, since it can lead to a larger and harder-to-maintain codebase. Also, the same functionality can already be achieved with `study.trials[trial_number]`.

For these reasons, we will close this PR. Thank you for your understanding.

PR #6369, opened 2025-12-05
Add `StorageTestCase` class in `optuna.testing` package

c-bata:
## Motivation

Third-party storage implementations currently have to copy Optuna’s storage tests (which is written in `tests/storage_tests/test_storage.py`). By exposing the shared `StorageTestCase` in `optuna.testing`, third-party Optuna storages can run the same test suites, improving consistency and fixing compatibility issues without duplicating test logic.

## Description of the changes

This PR suggests to exposing `StorageTestCase`, so that third-party Optuna storages can write tests like below:

```python
import pytest
from optuna.testing.pytest_storages import StorageTestCase
from optuna_integration.dask import DaskStorage


@pytest.fixture
def storage() -> BaseStorage:
    return DaskStorage(...)

class TestDaskStorageCase(StorageTestCase):
     @pytest.mark.skip("DaskStorage does not support this feature")
     def test_pickle(self, storage: BaseStorage) -> None:
        ...
```

### DIFF
--- a/optuna/testing/pytest_storages.py
+++ b/optuna/testing/pytest_storages.py
@@ -17,13 +17,8 @@ from optuna._typing import JSONSerializable
 from optuna.distributions import CategoricalDistribution
 from optuna.distributions import FloatDistribution
 from optuna.exceptions import UpdateFinishedTrialError
-from optuna.storages import _CachedStorage
 from optuna.storages import BaseStorage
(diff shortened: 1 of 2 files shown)


kAIto47802 (approved):
Thank you for the PR. LGTM
I've checked the all lines and confirmed that there are no missing tests and all tests are identical.

PR #6386, opened 2025-12-12
Move `FrozenTrial` import under `TYPE_CHECKING` for `_study_summary.py` file

Sip4818:
### Description

This PR addresses part of issue #6029: "Use TYPE_CHECKING if necessary" by moving the import of `FrozenTrial` in `study_summary.py` under a `TYPE_CHECKING` block.

`FrozenTrial` is only used for type annotations in this module and is not required at runtime. Moving this import under `TYPE_CHECKING` reduces unnecessary runtime imports and aligns with Optuna's ongoing cleanup of typing-only imports.

### Changes

- This PR moves the `FrozenTrial` import so that it is only loaded during type checking, since it is not needed at runtime.
- There are no runtime behavior changes.

### DIFF
--- a/optuna/study/_study_summary.py
+++ b/optuna/study/_study_summary.py
@@ -3,12 +3,15 @@ from __future__ import annotations
 from collections.abc import Sequence
 import datetime
 from typing import Any
+from typing import TYPE_CHECKING
 
 from optuna import logging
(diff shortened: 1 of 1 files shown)


not522 (approved):
Thank you! LGTM!

PR #6390, opened 2025-12-17
Use f-strings in `optuna/_experimental.py`

Rohan0497:
## Motivation

This PR addresses #6305. Since Optuna no longer supports Python 3.8, we can use f-strings instead of `.format()`.

## Description of the changes

Replaced `.format()` usage with f-strings in `optuna/_experimental.py` (one file only).

## Testing
- `pre-commit run --all-files`
- `pytest tests/test_experimental.py -vv`

Refs #6305


not522 (approved):
LGTM!

PR #6391, opened 2025-12-18
Fix invalid `StorageTestCase` scenarios involving trial state and values

c-bata:
## Motivation
Follow-up #6369

## Description of the changes

`StorageTestCase` includes several scenarios that do not align with Optuna’s expected trial lifecycle. These patterns are not valid in normal usage and can cause inconsistencies with the current validation logic.

This PR removes or updates the following cases:
* Marking a trial as COMPLETE without providing values.
* Providing values when marking a trial as FAIL, which violates the constraints enforced by [`FrozenTrial._validate()`](https://github.com/optuna/optuna/blob/a8f7033ea22d6443ebb1825f6fb4807861f28aac/optuna/trial/_frozen.py#L322-L323).
* Providing values for a WAITING trial, which is obviously unexpected scenario.

### DIFF
--- a/optuna/testing/pytest_storages.py
+++ b/optuna/testing/pytest_storages.py
@@ -744,8 +744,7 @@ class StorageTestCase:
         for state in states:
             t = _generate_trial(generator)
             t.state = state
-            if t.state == TrialState.COMPLETE:
-                t.values = [0.0]
+            t.values = [0.0] if t.state == TrialState.COMPLETE else None
             storage.create_new_trial(study_id, template_trial=t)
 
         trials = storage.get_all_trials(study_id, states=None)
@@ -817,8 +816,7 @@ class StorageTestCase:
         for s in states:
             t = _generate_trial(generator)
             t.state = s
-            if t.state == TrialState.COMPLETE:
-                t.values = [0.0]
+            t.values = [0.0] if t.state == TrialState.COMPLETE else None
             storage.create_new_trial(study_id, template_trial=t)
 
         assert storage.get_n_trials(study_id, TrialState.COMPLETE) == 2


c-bata:
@y0z @kAIto47802 Could you review this PR?

optuna/testing/pytest_storages.py
```diff
@@ -741,6 +744,8 @@ def test_get_all_trials_state_option(self, storage: BaseStorage) -> None:
         for state in states:
             t = _generate_trial(generator)
             t.state = state
+            if t.state == TrialState.COMPLETE:
+                t.values = [0.0]
```
kAIto47802:
```suggestion
            if t.state == TrialState.COMPLETE:
                t.values = [0.0]
                t.value = None
```

How about setting `t.value` to `None`, since the `value` and `values` attributes must not be specified in the same time? Currently, the `value` attribute is settled in `_generate_trial`, when the state is already `TrialState.COMPLETE`.

https://github.com/optuna/optuna/blob/21ac00e3ed4a51adf2fd5942fa8adc909b82f19c/optuna/trial/_frozen.py#L159-L160

kAIto47802:
```suggestion
            t.values = [0.0] if t.state == TrialState.COMPLETE else None
```
Currently, when the `t.state` is already `TrialState.COMPLETE` but the newly-set `state` is not `TrialState.COMPLETE`, the trial still has a non-None value in the `values` attribute, which is invalid.

optuna/testing/pytest_storages.py
```diff
@@ -812,6 +817,8 @@ def test_get_n_trials_state_option(self, storage: BaseStorage) -> None:
         for s in states:
             t = _generate_trial(generator)
             t.state = s
+            if t.state == TrialState.COMPLETE:
+                t.values = [0.0]
```
kAIto47802:
```suggestion
            t.values = [0.0] if t.state == TrialState.COMPLETE else None
```
Same as above.

kAIto47802 (changes requested):
Thank you for the PR. I left some comments. PTAL :octocat:

c-bata:
@kAIto47802 Thank you for your review. I applied your suggestions. PTAL.

PR #6393, opened 2025-12-19
Remove `asv` and the speed benchmark workflow

c-bata:
## Motivation
The `speed-benchmarks.yml` workflow has not been regularly monitored by Optuna maintainers for some time. After discussion among maintainers, we decided to remove both the workflow file and the asv related files to reduce their maintenance cost.

## Description of the changes
This PR removes `speed-benchmarks.yml` and `asv` related files.


pyproject.toml
```diff
@@ -43,7 +43,6 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 benchmark = [
-  "asv>=0.5.0",
   "cma",
   "virtualenv"
 ]
```
not522: Since this PR removes all optional `benchmark` installations, they should be completely removed.

not522:

pyproject.toml
c-bata: Good catch. I addressed this comment in 35f1e07de.

not522:
Could you also remove `benchmarks/__init__.py` and these `benchmark`/`benchmarks`? https://github.com/optuna/optuna/blob/21ac00e3ed4a51adf2fd5942fa8adc909b82f19c/pyproject.toml#L160 https://github.com/optuna/optuna/blob/21ac00e3ed4a51adf2fd5942fa8adc909b82f19c/Dockerfile#L19

c-bata:
Thank you for pointing it out :bow: I updated two files you pointed it out and `checks-optional.yaml` file.

not522:
Let's completely remove the `benchmarks` directory as well.

c-bata:
Addressed in https://github.com/optuna/optuna/pull/6393/commits/a7dbfb669b1d7fad428c5111ffd56c7fd00f6402 :bow: :bow: :bow:

PR #6394, opened 2025-12-19
Remove python 3.9 due to EOL Status

ParagEkbote:
## Motivation
Since python 3.9 has reached [EOL](https://devguide.python.org/versions/) for some time, we can remove it for security and development purpose. Could you please review?

## Description of the changes
Removed python 3.9 from every major file.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -28,7 +28,7 @@ jobs:
 
     strategy:
       matrix:
-        python_version: ['3.9', '3.10', '3.11', '3.12']
+        python_version: ['3.10', '3.11', '3.12']
         build_type: ['', 'dev']  # "dev" installs all the dependencies including pytest.
 
     # This action cannot be executed in the forked repository.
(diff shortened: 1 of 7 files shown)


ParagEkbote:
I have some queries about a couple of files:

1.  Can the minimum python value in the dockerfile be also updated: https://github.com/optuna/optuna/blob/master/Dockerfile

2.  Can the minimum python version in the following gh-action: https://github.com/optuna/optuna/blob/master/.github/workflows/tests-with-minimum-versions.yml#L23

The scikit-learn dependency in the singular failing test has to be updated to `scikit-learn>=1.0.2` since the versions above have wheels with py3.10 support. What is the min version to update it?

c-bata:
Thank you for your pull request.

Regarding https://github.com/optuna/optuna-examples/pull/344, we have decided to proceed with removing Python 3.9 at this time. For Optuna itself, however, we would like to be a bit more cautious and align on the timing among the maintainers, taking downstream users into account. We will therefore close this PR for now and revisit the change once we have reached a consensus on the schedule.

Thanks again for your understanding and contribution.

Issue #6085, opened 2025-05-22
RetryFailedTrialCallback not working?


Sebastian-0:
### Expected behavior

When using the `RetryFailedTrialCallback` I expect failed trials to retry and not terminate the optimizer immediately.

According to [the docs](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.storages.RetryFailedTrialCallback.html):
> When a trial fails, this callback can be used with a class in [optuna.storages](https://optuna.readthedocs.io/en/stable/reference/storages.html#module-optuna.storages) to recreate the trial in TrialState.WAITING to queue up the trial to be run again.

So, for any errors like exceptions, wrong return type, etc... it should retry up to specified amount of times.

### Environment

- Optuna version: 4.3.0
- Python version: 3.12.3
- OS: Linux-6.11.0-112021-tuxedo-x86_64-with-glibc2.39

### Error messages, stack traces, or logs

```shell
$ python3 test.py 
/home/sebastian/IntuiCell/intui/experiments/suitability/tmp/test.py:8: ExperimentalWarning: RetryFailedTrialCallback is experimental (supported from v2.8.0). The interface can change in the future.
  failed_trial_callback=RetryFailedTrialCallback(max_retry=5),
[I 2025-05-22 10:10:00,274] A new study created in RDB with name: test-2025-05-22_10:10:00
/home/sebastian/IntuiCell/intui/experiments/suitability/tmp/venv/lib/python3.12/site-packages/optuna/study/_optimize.py:186: ExperimentalWarning: fail_stale_trials is experimental (supported from v2.9.0). The interface can change in the future.
  optuna.storages.fail_stale_trials(study)
[W 2025-05-22 10:10:05,323] Trial 0 failed with parameters: {'val': 0.4} because of the following error: RuntimeError('Throw some error').
Traceback (most recent call last):
  File "/home/sebastian/IntuiCell/intui/experiments/suitability/tmp/venv/lib/python3.12/site-packages/optuna/study/_optimize.py", line 197, in _run_trial
    value_or_values = func(trial)
                      ^^^^^^^^^^^
  File "/home/sebastian/IntuiCell/intui/experiments/suitability/tmp/test.py", line 17, in obj
    raise RuntimeError("Throw some error")
RuntimeError: Throw some error
[W 2025-05-22 10:10:05,323] Trial 0 failed with value None.
Traceback (most recent call last):
[...]

PR #6398, opened 2025-12-22
Clarify catch requirement for RetryFailedTrialCallback

Quant-Quasar:
## Motivation
Users reported confusion around `RetryFailedTrialCallback` where retries do not occur if
the optimization terminates on an uncaught exception (see #6085).

The current documentation does not explicitly mention that exceptions must be caught in
`Study.optimize` for the retry mechanism to take effect.

## Description of the changes
This PR clarifies the usage contract of `RetryFailedTrialCallback` by adding a note to the
documentation explaining that retries require the optimization loop to continue running.

It explicitly states that users must specify exceptions to catch in `Study.optimize` and
provides a minimal working example using the `catch` argument. No runtime behavior is changed.

### DIFF
--- a/optuna/storages/_callbacks.py
+++ b/optuna/storages/_callbacks.py
@@ -14,6 +14,9 @@ class RetryFailedTrialCallback:
 
     When a trial fails, this callback can be used with a class in :mod:`optuna.storages` to
     recreate the trial in ``TrialState.WAITING`` to queue up the trial to be run again.
+    Note that the optimization loop must continue running after a failure for retries to
+    occur; therefore, exceptions must be caught in
+    :func:`~optuna.study.Study.optimize` (e.g., via the ``catch`` argument).
 
     The failed trial can be identified by the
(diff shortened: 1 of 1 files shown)


not522:
Sorry, but the original documentation and the comment in #6085 are incorrect as explanations for `RetryFailedTrialCallback`, and this PR hasn't properly corrected them. This feature is designed to rerun trials marked as stale by `RDBStorage`'s heartbeat mechanism, and it won't affect other "failed trials." Since I think significant changes are needed to the API and documentation, I'm sorry but I'll have to close this PR. I'll also discuss this with other core developers and add comments to #6085. Thank you for your contribution.

Quant-Quasar:
Thank you for the detailed explanation and for taking the time to clarify the intended design. I appreciate the feedback and understand that this requires broader API and documentation changes. Happy to follow the discussion on #6085 and contribute further if helpful.

PR #6400, opened 2025-12-25
Use `logger.warning` instead of `optuna_warn` for lock-acquisition delay notifications

sawa3030:
## Motivation
Follow-up to #6361. This message is not something users can fix by changing their code, but it’s still useful to note as an event. [Per Python’s logging guide](https://docs.python.org/3/howto/logging.html#when-to-use-logging), this should be logged via `logger.warning()` rather than raised with `warnings.warn()`.

## Description of the changes
- Replace `warnings.warn()` with `logger.warning()` for the lock-acquisition delay notification.


not522:

sawa3030:
@c-bata Thanks for the suggestion. I’ve replaced `warnings.warn()` with `_logger.warning()` in those two places as well.

PR #6409, opened 2026-01-11
Replace `.format()` with f-string in `_rank.py`

jwalith:
- Changed line 131 from .format() to f-string
- Related to issue #6305

## Motivation
Addressing (https://github.com/optuna/optuna/issues/6305 for compatibility.)

## Description of the changes
Replaced the .format() call with f-string on line 131 in optuna/visualization/_rank.py with an f-string using the preferred {var_name=} syntax.


jwalith:
Thank you, Really appreciate the thought of keeping one file per contributor. It allows new contributors like me to start our open source journey.

PR #6412, opened 2026-01-16
Replace .format() with f-string in _fixed.py

VedantMadane:
## Summary

Part of #6305 - Replace .format() with f-string in optuna/trial/_fixed.py.

## Changes

Converted string formatting in FixedTrial._suggest() method from .format() to f-strings:
- Error message when parameter value is not found
- Warning message when parameter value is out of distribution range

## Test Plan

- Existing tests should pass
- No functional change, only formatting style


VedantMadane:
There are still ~45 more files with .format() calls in the optuna codebase if you'd like me to continue with more?

PR #6416, opened 2026-01-16
Replace .format() with f-string in _imports.py

VedantMadane:
## Summary

Part of #6305 - Replace `.format()` with f-string in `optuna/_imports.py`.

## Changes

Converted error messages in `_DeferredImportExceptionContextManager.__exit__()` to use f-strings for both ImportError and SyntaxError cases.

## Test Plan

- Existing tests should pass
- No functional change, only formatting style


VedantMadane:
There are still ~45 more files with .format() calls in the optuna codebase if you'd like me to continue with more?

PR #6424, opened 2026-01-27
Add `SamplerTestCase` class in `optuna.testing package`

y0z:
## Motivation

Similar to #6369, this PR makes it easier for third-party sampler implementations to use the samplers' tests in Optuna.

## Description of the changes

- Introduce several fixtures to categorize samplers, such as those supporting relative sampling, multi-objective optimization, etc.
- Introduce the `SamplerTestCase` class in `optuna.testing.pytest_samplers.py` that implements general tests for samplers using the fixtures above.
- Refactoring `test_samplers.py` with `SamplerTestCase`.

### DIFF
--- a/optuna/testing/tempfile_pool.py
+++ b/optuna/testing/tempfile_pool.py
@@ -4,38 +4,40 @@
 from __future__ import annotations
 
 import atexit
-import gc
+import copy
 import os
 import tempfile
+import threading
 from types import TracebackType
 from typing import Any
+from typing import cast
+from typing import ClassVar
 from typing import IO
 
 
 class NamedTemporaryFilePool:
-    tempfile_pool: list[IO[Any]] = []
-
-    def __new__(cls, **kwargs: Any) -> "NamedTemporaryFilePool":
-        if not hasattr(cls, "_instance"):
-            cls._instance = super(NamedTemporaryFilePool, cls).__new__(cls)
-            atexit.register(cls._instance.cleanup)
-        return cls._instance
+    _lock: ClassVar[threading.Lock] = threading.Lock()
+    _path: ClassVar[list[str]] = []
+    _registered: ClassVar[bool] = False
 
     def __init__(self, **kwargs: Any) -> None:
         self.kwargs = kwargs
+        self._file: IO[bytes] | IO[str] | None = None
 
-    def tempfile(self) -> IO[Any]:
-        self._tempfile = tempfile.NamedTemporaryFile(delete=False, **self.kwargs)
-        self.tempfile_pool.append(self._tempfile)
-        return self._tempfile
+        with self.__class__._lock:
+            if not self.__class__._registered:
+                _ = atexit.register(self.__class__.cleanup)
+                self.__class__._registered = True
 
-    def cleanup(self) -> None:
-        gc.collect()
-        for i in self.tempfile_pool:
-            os.unlink(i.name)
-
-    def __enter__(self) -> IO[Any]:
+    def __enter__(self) -> IO[bytes] | IO[str]:
         return self.tempfile()
+
+    def tempfile(self) -> IO[bytes] | IO[str]:
+        f = cast("IO[bytes] | IO[str]", tempfile.NamedTemporaryFile(delete=False, **self.kwargs))
+        self._file = f
+        with self.__class__._lock:
+            self.__class__._path.append(f.name)
+        return self._file
 
     def __exit__(
         self,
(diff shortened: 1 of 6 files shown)


c-bata:

tests/samplers_tests/test_samplers.py
```diff
@@ -999,65 +92,445 @@ def objective(trial: Trial) -> float:
...
+        # But the sequences are expected to be the same.
+        assert sequence_dict[0] == sequence_dict[1] == sequence_dict[2]
+
+    @pytest.mark.parametrize("seed", [0, 169208])
+    def test_pickle_random_sampler(self, seed: int) -> None:
+        sampler = optuna.samplers.RandomSampler(seed)
+        restored_sampler = pickle.loads(pickle.dumps(sampler))
+        assert sampler._rng.rng.bytes(10) == restored_sampler._rng.rng.bytes(10)
+
+    def test_before_trial(self) -> None:
```
sawa3030: I noticed that `test_before_trial`, `test_after_trial`, `test_after_trial_pruning`,
`test_after_trial_failing`, `test_after_trial_failing_in_after_trial`,
`test_after_trial_with_study_tell`, and `test_sample_relative` aren’t really testing sampler
implementations themselves. I think they’re more about verifying the study’s behavior, so it feels a bit inconsistent to treat them as peers of the other
sampler-parameterized tests in `TestSampler` / `SamplerTestCase`.

Would it make sense to move these tests out of `TestSampler`?

sawa3030:
If my understanding is correct, testing a third-party sampler currently requires implementing fixtures like this:

``` python
import pytest
from optuna.testing.pytest_samplers import SamplerTestCase

class TestTurboSampler(SamplerTestCase):
    @pytest.fixture
    def sampler_class(self):
        return TuRBOSampler()

    @pytest.fixture
    def relative_sampler_class(self):
        pytest.skip("Relative Sampler cannot be tested since TuRBOSampler requires startup trials.")
    
    @pytest.fixture
    def multi_objective_sampler_class(self):
        pytest.skip("TurboSampler does not support multi-objective optimization.")

    @pytest.fixture
    def single_only_sampler_class(self):
        return TuRBOSampler()
    
```

Would it be possible (or reasonable) to make this more straightforward, for example:

``` python
import pytest
from optuna.testing.pytest_samplers import SamplerTestCase

@pytest.fixture
def sampler():
    return TurboSampler(...)

class TestTurboSampler(SamplerTestCase):
    SUPPORTS_RELATIVE = False 
    SUPPORTS_MULTI_OBJECTIVE = False
    SUPPORTS_SINGLE_ONLY = True
```

I’m not entirely sure if this would fit the current design, so I’d like to take a bit of time to think through the implementation.

tests/samplers_tests/test_samplers.py
y0z: Thank you for your suggestion.
I moved the tests as follows.
- `test_before_trial` -> `test_trial.py`
  - Since `before_trial` is called in [`Trial`](https://github.com/optuna/optuna/blob/master/optuna/trial/_trial.py#L59).
- `test_after_trial_*` -> `test_study.py`
  - Since `after_trial` is called in [`study/_tell.py`](https://github.com/optuna/optuna/blob/master/optuna/study/_tell.py#L172).

y0z:
I divided the test cases into four scenarios: `BasicSamplerTestCase`, `MultiObjectiveSamplerTestCase`, `RelativeSamplerTestCase`, and `SingleOnlySamplerTestCase`. Third-party developers can easily use the test cases as follows by defining a sampler fixture and inheriting the scenarios they need.

```python
import pytest

import optuna
from optuna.samplers import BaseSampler
from optuna.testing.pytest_samplers import BasicSamplerTestCase
from optuna.testing.pytest_samplers import MultiObjectiveSamplerTestCase
from optuna.testing.pytest_samplers import RelativeSamplerTestCase
from optuna.testing.pytest_samplers import SingleOnlySamplerTestCase

# Example 1
class TestSampler1(BasicSamplerTestCase, MultiObjectiveSamplerTestCase):
    @pytest.fixture
    def sampler(self) -> Callable[[], BaseSampler]:
        return optuna.samplers.TPESampler

# Example 2
class TestSampler2(RelativeSamplerTestCase):
    @pytest.fixture
    def sampler(self) -> Callable[[], BaseSampler]:
        return lambda: optuna.samplers.TPESampler(multivariate=True, n_startup_trials=0)

# Example 3
class TestSampler3(SingleOnlySamplerTestCase):
    @pytest.fixture
    def sampler(self) -> Callable[[], BaseSampler]:
        return optuna.samplers.CmaEsSampler

    def test_user_defined(self, sampler) -> None:
        # Additional test can be defined if necessary.
        assert sampler.__name__ == "CmaEsSampler"
```

What do you think?

y0z:
@c-bata @sawa3030 

I've split the changes into two PRs. Please review and merge the first one (#6429) before reviewing this PR.

c-bata (approved):
LGTM. Thank you for your pull requests!

sawa3030 (approved):
LGTM!

